PR #22275, opened 2025-01-07
CI: add a GPU CI job

### LINKED ISSUE #21740: CI: adding a GPU-enabled CI job
_We've touched on this in a few places, in a [community meeting](https://hackmd.io/pyhudrC5TgSwdJtpPldHgQ#Agenda-for-July-24th-2024), on a PR I can't find back right now, and in in-person discussions. So here is a new issue with a concrete proposal._

We have experimental support for CuPy, PyTorch and JAX - all of which have CUDA (and in some cases ROCm and/or Intel XPUs) support, which we are not exercising in CI. As a result, we get issues like gh-21486 where someone finds locally that the test suite has regressed for a GPU-enabled array library. This is obviously not ideal, and we'd like a low-overhead and low-cost way to test PRs that a reviewer thinks may impact GPU support in CI.

Scikit-learn has such a CI job, see [the GHA workflow file for it](https://github.com/scikit-learn/scikit-learn/blob/9d39f57399d6f1f7d8e8d4351dbc3e9244b98d28/.github/workflows/cuda-ci.yml) and @betatim's [blog post describing the design](https://betatim.github.io/posts/github-action-with-gpu/). 

The other thing we needed for it was a credit card from NumFOCUS that we can attach to this repo (with a limit on spending level, as described in the blog post) - we have that as well now. So we're all set in principle to add this.

Suggested logistics and design:
- The SciPy core team should sign off on the spending limit (just like we did for Cirrus CI usage). I'd suggest either $50/month or $100/month as the limit, and adjust if needed once we have some experience with it.
- The label-based triggering that scikit-learn has, where adding the label to a PR does a single run and then the label gets removed again so new commits don't auto-run the job again, seems nice - so let's copy it.
- We can start with building SciPy in the GPU-enabled CI runnner; if it becomes expensive we can move the SciPy wheel build to a separate (CPU-only, free) job, upload the wheel to the GHA cache, and then trigger a second job based on the first one to run the CUDA tests.
- The one change from the scikit-learn setup we should make is to use `pixi` rather than `conda-lock`. The latter makes sense for scikit-learn since they were already using it, but `pixi` will work better for us and we've mostly already got the setup worked out in https://github.com/rgommers/pixi-dev-scipystack. I don't want to propose adding that setup in the root of the repo, but only the single environment + lock file that we need to run this GPU job.

Thoughts?

rgommers:
_draft for now because not all tests pass yet. these are recent regressions in `main`, and I'd prefer for those to be fixed in another PR first_

- Runs tests for all array API-enabled functionality with PyTorch, JAX and CuPy in a single CI job (for now)
- Builds on the GPU runner, because with `ccache` builds are very fast and the setup time is dominated by the ~3 GB worth of packages that are needed for GPU installs of PyTorch, JAX and CuPy in a single conda environment.
- Uses [Pixi](https://pixi.sh/latest/) to ensure we get a maintainable and robust environment.  
    - Keeps `pixi.toml` inside `.github/workflows/` for now, because I don't (yet) want to cross the bridge of enabling Pixi for all SciPy contributors (there's too much logistics to work out still there for how to do lock file updates).
    - Failures reproduce quite well locally. Should be a matter of `cd .github/workflows` and then running the same `pixi run test-cuda ...` command as the failing CI job step.

Closes gh-21740

### DIFF
--- a/.github/workflows/gpu-ci.yml
+++ b/.github/workflows/gpu-ci.yml
@@ -42,13 +42,23 @@ jobs:
         uses: cirruslabs/cache@v4  #caa3ad0624c6c2acd8ba50ad452d1f44bba078bb # v4
         with:
           path: ${{ env.PIXI_CACHE_DIR }}
-          key: ${{ runner.os }}-gpu-pixi
+          # Cache hit if lock file did not change. If it did, still restore the cache,
+          # since most packages will still be the same - the cache save will
+          # then happen at the end (in case the lock file didn't change,
+          # nothing is saved at the end of a job).
+          key: ${{ runner.os }}-gpu-pixi-${{ hashFiles('.github/workflows/pixi.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-gpu-pixi
 
       - name: Setup compiler cache
         uses: cirruslabs/cache@v4  #caa3ad0624c6c2acd8ba50ad452d1f44bba078bb # v4
         with:
           path: ${{ env.CCACHE_DIR }}
-          key: ${{ runner.os }}-gpu-ccache
+          # Make primary key unique by using `run_id`, this ensures the cache
+          # is always saved at the end.
+          key: ${{ runner.os }}-gpu-ccache-${{ github.run_id }}
+          restore-keys: |
+            ${{ runner.os }}-gpu-ccache
 
       - name: run nvidia-smi
         run: nvidia-smi
@@ -60,6 +70,7 @@ jobs:
         with:
           pixi-version: v0.39.2
           manifest-path: .github/workflows/pixi.toml
+          cache: false
 
       - name: Build SciPy
         working-directory: .github/workflows
--- a/.github/workflows/gpu-ci.yml
+++ b/.github/workflows/gpu-ci.yml
@@ -1,5 +1,31 @@
 name: GPU jobs
 
+# Updating the Pixi lock file:
+#
+# 1. Please try to avoid this unless it's actually necessary. Updating the lock
+#    file generates a large diff, and triggers a lot of large package downloads
+#    (also for local testing by maintainers).
+# 2. Consider committing the `pixi.lock` change in one commit, and the
+#    `pixi.lock` change in a second commit. This allows the reviewer to redo
+#    the lock file changes locally, or discard them in case of conflicts while
+#    still keeping the `pixi.toml` changes.
+#
+#
+# In order to run a CI job step locally on a Linux machine with an NVIDIA GPU:
+#
+#     $ cd .github/workflows
+#     $ pixi run build
+#     $ pixi run -e cupy test-cupy -v -m "array_api_backends and not slow"
+#
+# Or run different test commands in the same environments, e.g.:
+#
+#     $ pixi run test-torch-cuda -s special
+#     $ pixi run test-torch -s special
+#
+# To see available tasks per environment:
+#
+#     $ pixi task ls -s
+
 on:
   push:
     branches:
--- a/.github/workflows/gpu-ci.yml
+++ b/.github/workflows/gpu-ci.yml
@@ -5,7 +5,7 @@ name: GPU jobs
 # 1. Please try to avoid this unless it's actually necessary. Updating the lock
 #    file generates a large diff, and triggers a lot of large package downloads
 #    (also for local testing by maintainers).
-# 2. Consider committing the `pixi.lock` change in one commit, and the
+# 2. Consider committing the `pixi.toml` change in one commit, and the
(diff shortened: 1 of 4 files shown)


rgommers:
Note that the job itself won't run on this PR before merging. Here's a [CI log](https://github.com/scipy/scipy/actions/runs/12658348129) that's less than an hour old.

.github/workflows/gpu-ci.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: run nvidia-smi
+        run: nvidia-smi
+
+      - name: run nvidia-smi --query
+        run: nvidia-smi --query
+
+      - uses: prefix-dev/setup-pixi@ba3bb36eb2066252b2363392b7739741bb777659 # v0.8.1
+        with:
+          pixi-version: v0.39.2
+          manifest-path: .github/workflows/pixi.toml
```
lucascolley, line 62: It sounds like this is needed to ensure caching of the non-default environments: https://github.com/prefix-dev/setup-pixi?tab=readme-ov-file#install-multiple-environments-in-one-job

```suggestion
      - uses: prefix-dev/setup-pixi@ba3bb36eb2066252b2363392b7739741bb777659 # v0.8.1
        with:
          pixi-version: v0.39.2
          manifest-path: .github/workflows/pixi.toml
          environments: >-
            default
            torch-cuda
            jax-cuda
            cupy
```

lucascolley: unless this was deliberate to avoid hitting the cache limit?

ev-br:
> Note that the job itself won't run on this PR before merging.

How about merge and iterate then? Unless you want to fix regression before merging

rgommers:
> How about merge and iterate then? Unless you want to fix regression before merging

I do want the job to pass before merging, otherwise it will make CI on all PRs red.

Help is very welcome - most issues are recent regressions for PRs that weren't tested on GPU. I won't be able to do more before Friday at least. It may be a matter of some skips in `signal` and `stats`, at least to fix PyTorch GPU.

ev-br:
https://github.com/scipy/scipy/pull/22279 fixes all torch and jax.numpy failures I see locally.

.github/workflows/gpu-ci.yml
rgommers: It's caching everything it looks like - there's 3 GB worth of packages cached says the CI log, and there's no downloads on using the other envs. 

This doesn't quite work the way you think - `setup-pixi` does no caching, it's done higher up by `cirruslabs/cache@v4`. Pixi already has a local cache of all packages that were downloaded in a previous successful run, so no envs need specifying.

lucascolley: Ah okay, so setup-pixi *can* do caching, but we're using cirruslabs/cache instead, so it would be redundant - is that right?

rgommers: Yes, that is correct. The README at https://github.com/cirruslabs/cache explains the benefit over regular caching (tl;dr it's more local and hence faster, no need to cycle between Cirrus and GitHub infra)

lucascolley: should we use `cache: false` with setup-pixi, then?

rgommers:
All tests pass after gh-22279, thanks again @ev-br and @lucascolley.

Test suite runtime is reasonable at the moment:

image

If we enable much more functionality on GPU then we may end up splitting it, but for now it's easier to have it in a single job.

Still optimizing cache strategies a bit, then this should be ready.

.github/workflows/gpu-ci.yml
rgommers: Yes, good point - adding that.

rgommers:
Caching performance after the changes I'm about to push, that's about as good as it's going to get:

image

rgommers:
Okay, ready for final review.

rgommers:
See https://github.com/scipy/scipy/actions/workflows/gpu-ci.yml for more logs if you want to see what the changes do.

.github/workflows/gpu-ci.yml
```diff
@@ -0,0 +1,115 @@
+name: GPU jobs
+
+# Updating the Pixi lock file:
+#
+# 1. Please try to avoid this unless it's actually necessary. Updating the lock
+#    file generates a large diff, and triggers a lot of large package downloads
+#    (also for local testing by maintainers).
+# 2. Consider committing the `pixi.lock` change in one commit, and the
```
lucascolley, line 8:
```suggestion
# 2. Consider committing the `pixi.toml` change in one commit, and the
```

nit

lucascolley (approved):
let's give this a go and follow up if anything goes red - thanks Ralf!

PR #22222, opened 2025-01-01
ENH: `special`/`stats`: implement xp-compatible `stdtrit` and use in `stats`

mdhaber:
#### Reference issue
gh-20544
Follow-up to gh-20884

#### What does this implement/fix?
This adds an Array API implementation of `stdtrit` for backends that define `betainc`.

#### Additional information
Came up in the context of mdhaber/marray#71

PR #22279, opened 2025-01-08
TST: Add skips for GPU CI failures

ev-br:
Add test skips for `SCIPY_DEVICE=cuda`, as seen locally (on qgpu3 in fact, but that's no matter).

AFAICS, the bulk of failures are genuinely cpu-only:
- ndimage is compiled C code
-  signal.order_filter is compiled C code
- stats ttest-related failures are all hitting a special function from either specfun or cephes, so it's a skip until xsf is a thing

### DIFF
--- a/scipy/ndimage/tests/test_interpolation.py
+++ b/scipy/ndimage/tests/test_interpolation.py
@@ -603,6 +603,7 @@ class TestMapCoordinates:
         assert out.dtype is np.dtype('f')
         assert_array_almost_equal(out, xp.asarray([[1]]))
 
+    @pytest.mark.skip_xp_backends(cpu_only=True)
     @pytest.mark.skipif('win32' in sys.platform or np.intp(0).itemsize < 8,
                         reason='do not run on 32 bit or windows '
                                '(no sparse memory)')
--- a/scipy/signal/tests/test_signaltools.py
+++ b/scipy/signal/tests/test_signaltools.py
@@ -1621,6 +1621,7 @@ class TestCSpline1DEval:
         assert ynew.dtype == y.dtype
 
 
+@skip_xp_backends(cpu_only=True, exceptions=['cupy'])
 class TestOrderFilt:
 
     def test_basic(self, xp):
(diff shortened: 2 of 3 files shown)


lucascolley:
@mdhaber some of this is fallout from gh-22222, I should have ran the GPU tests. Could you investigate whether it was wrong to remove the GPU skips, or whether something isn't working in the implementation?

lucascolley:
https://github.com/scipy/scipy/blob/9166b6fc4714fa55d4e829c8a76a278b8ecaf647/scipy/special/tests/test_support_alternative_backends.py#L61-L69

Looks like a skip should be added for torch in these stats tests, but we don't need full cpu_only skips.

mdhaber:
> Could you investigate whether it was wrong to remove the GPU skips, 

Did you already answer this question in the next comment? I think it works for CuPy. I did test that locally. I can see why it would fail for torch on GPU: no `betainc`, and no fallback to NumPy/SciPy.

lucascolley:
yep

rgommers (approved):
Thanks! Seems to fix all the problems indeed.

rgommers:
Not merging because I think @lucascolley's comment above asks for a tweak.

ev-br:
Have to admit I don't understand https://github.com/scipy/scipy/pull/22279#issuecomment-2577482688. Lucas, could you either push an update to this PR or give some more specifics of what your suggestion is, please?

lucascolley:
For the stats tests here, gh-22222 went from `cpu_only` to no skips. But the non-cpu skips should only have been removed on libraries for which we can delegate to `betainc` (CuPy, JAX).

This PR currently restores the full `cpu_only` skips. I believe the correct fix is to replace them with `cpu_only=True, exceptions=["cupy", "jax.numpy"]` skips.

Is that clear @ev-br ?

ev-br:
Thanks, done.

lucascolley (approved):

lucascolley:
thanks!

rgommers:
The job isn't actually triggering, I'm looking into it.

Issue #22246, opened 2025-01-04
tracking issue for JAX support in SciPy


rgommers:
_This is part of all the work on array types support (xref gh-18286). JAX support, including `jax.jit` and GPU support, is a multi-step process that is slightly more involved than CuPy/PyTorch support because JAX's programming model diverges more from NumPy, and after the discussion in gh-22070 a separate tracking issue seems warranted._

The full set of changes that are currently on our radar as needed for solid JAX support, from beginning to end, are (not necessarily exactly in order):

- [x] The initial work in gh-20085 that introduced eager JAX support
- [x] The changes to test skip syntax that that helped trigger
- [x] A change to deal with in-place operations with integer indexing (done in [gh-22070](https://github.com/scipy/scipy/pull/22070))
- [ ] A similar change for dealing with in-place operations with boolean masks
- [ ] A way to avoid `jax.jit` choking on input argument validation (xref [array-api-compat#225](https://github.com/data-apis/array-api-compat/issues/225))
- [ ] Making code that calls into compiled code in SciPy work with `jax.jit` (doable as long as shape and dtype of the returned array from the compiled function is predictable, probably with `jax.pure_callback`)
- [ ] Adding a CI job that verifies that `jax.jit` works on SciPy functionality supporting it.

What I think we should do is make these changes as quickly as we can on *a representative subset of SciPy only*, to avoid having a large amount of churn for one of the steps 3-6, across the whole code base and then figuring out that the end goal is somehow blocked by an unforeseen issue. After discussion with @crusaderky (who is doing the heavy lifting at the moment), I think `cluster` plus the part of `special` that already delegates to JAX for matching functions is suitable for this.

We need an assessment at the end of getting `jax.jit` to work on `cluster` and `special`, to decide that we're happy with what JAX support looks like. We expect that the benefits of supporting JAX will be worth the costs - before making a large amount of changes to all other relevant SciPy submodules, let's validate that. The benefits should include:

1. significant enough performance gains of `jax.jit` vs. `numpy` on CPU and GPU,
2. enough of SciPy's functionality/submodules able to have support for `jax.jit` (and `vmap`, ...),

The potential costs are:

1. impact on support for other libraries because of having to rewrite some code constructs in a different style - ensure that there is no significant impact.

    - Expected outcome: No degradation in NumPy (and CuPy and eager PyTorch) performance, aside from perhaps a small O(1) overhead for extra checks - but certainly no extra memory copies or worse asymptotic behavior.
 
[...]

PR #22281, opened 2025-01-08
TST: special: better skip message for stdtrit on JAX

crusaderky:
`scipy.special.stdtrit` depends on scipy.optimize.
As explained in https://github.com/scipy/scipy/issues/22246, I'd rather not go down the rabbit hole of fixing non-jitted JAX on scipy.optimize before I've finished the work on scipy.special and scipy.cluster.


lucascolley:
thanks!

PR #22285, opened 2025-01-09
WIP/MAINT: interpolate/BSpline: lock mutations of self.

ev-br:
Try adapting the BSpline code to work in the nogil context. 

Not knowning much about free-threading, I *assume* that 
- mutating attributes of `self` is not safe in the free-threaded context
- mutating attributes of `self` in `__init__(self, ...)` does not need manual locking though 
- a module-level `Lock` is enough and we do not need to carry a lock instance on each instance of a class
- python code which mutates function-scope python objects does not need manual locking

### DIFF
--- a/scipy/interpolate/tests/test_bsplines.py
+++ b/scipy/interpolate/tests/test_bsplines.py
@@ -460,7 +460,6 @@ class TestBSpline:
         spl1 = BSpline(t, c[1], k)
         xp_assert_equal(spl(2.5), [spl0(2.5), spl1(2.5)])
 
-    @pytest.mark.thread_unsafe
     def test_design_matrix_bc_types(self):
         '''
         Splines with different boundary conditions are built on different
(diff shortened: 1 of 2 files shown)


scipy/interpolate/tests/test_bsplines.py
```diff
@@ -635,7 +634,6 @@ def test_read_only(self):
         b = BSpline(t=t, c=c, k=0)
         xp_assert_close(b(xx), np.ones_like(xx) * 3.0)
 
-    @pytest.mark.thread_unsafe
```
andfoy, line 638: I think this marker was put in place to prevent threads from spawning more threads, given that the test already runs in parallel

andfoy:
> a module-level Lock is enough and we do not need to carry a lock instance on each instance of a class

Given that the class access can be concurrent, and there are no module-level instances (i.e., singleton object), it would be much better to define this lock as an attribute.

andfoy:
> mutating attributes of self is not safe in the free-threaded context

It depends if the concurrent access occurs in a single object instance; my guess is that for paralellism gains, each thread would instantiate a separate instance of `BSpline`, rather than accessing the same instance, thus ensuring no concurrent access takes place

ev-br:
Huge thanks @andfoy for the review! 

Since the original premise of this PR was to remove `@thread_unsafe` decorators from the test suite, and it appears that those decorators were deliberate and do not actually signal real thread-safety issues, let's reject and close this PR.

PR #22286, opened 2025-01-09
BUG: sparse.csgraph.floyd_warshall: do not set to infinity for masked array

### LINKED ISSUE #12424: Different algorithms giving different results in `shortest_path`
It seems that, at least when masked arrays are passed to `shortest_path` and some entries are `0`, the output is algorithm-dependent.

In my application, the meaning of `0` in a dense graph to be passed to `shortest_path` should be that of an edge with no cost, not one with an infinite cost (absent edge). I understand that this is not the view adopted by `shortest_path`, and have hence resorted to using masked arrays instead. However, this leads to unexpected discrepancies in the results. 

```
import numpy as np
from scipy.sparse.csgraph import shortest_path

csgraph = np.array(
    [[0, 1, 0], 
     [1, 0, 0],
     [0, 0, 0]]
)
csgraph_masked = np.ma.masked_invalid(csgraph, directed=False)

shortest_FW = shortest_path(csgraph_masked, method='FW')
shortest_J = shortest_path(csgraph_masked, method='J')
shortest_D = shortest_path(csgraph_masked, method='D')
shortest_BF = shortest_path(csgraph_masked, method='BF')
```

#### Expected result:
The documentation does not suggest that, when the graphs are undirected and there are no negative values, outputs according to different methods should be different (and since the `'auto'` option is available, as I user one would rather like a guarantee that this is not the case). However, in the above example, while the last three results are all equal to
```
array([[0., 0., 0.],
       [0., 0., 0.],
       [0., 0., 0.]])
```
(which, incidentally, is the one I want), `shortest_FW` equals
```
array([[ 0.,  1., inf],
       [ 1.,  0., inf],
       [inf, inf,  0.]])
```
What I guess is happening is that `'FW'` is converting to a dense representation and forgetting about the mask completely.

```
1.4.1 1.18.4 sys.version_info(major=3, minor=8, micro=2, releaselevel='final', serial=0)
```

czgdp1807:
#### Reference issue

Closes https://github.com/scipy/scipy/issues/12424

Simple fix - just added an extra condition in floyd warhsall algorithm. It avoids setting infinity for masked arrays as well. Then its output is exactly same as other algorithms.

### DIFF
--- a/scipy/sparse/csgraph/tests/test_shortest_path.py
+++ b/scipy/sparse/csgraph/tests/test_shortest_path.py
@@ -389,22 +389,21 @@ def test_masked_input():
     for method in methods:
         check(method)
 
+
 @pytest.mark.parametrize("method", ['FW', 'J', 'BF'])
 def test_masked_invalid_input(method):
     # Reference - https://github.com/scipy/scipy/pull/22286
     csgraph = np.array(
         [[0, 1, 0],
-        [1, 0, 0],
-        [0, 0, 0]]
+         [1, 0, 0],
+         [0, 0, 0]]
     )
     csgraph_masked = np.ma.masked_invalid(csgraph)
-
     zeros = np.zeros((3, 3))
 
     SP = shortest_path(csgraph_masked, method=method, directed=True,
-                        overwrite=False)
+                       overwrite=False)
     assert_equal(SP, zeros)
-
 
 
 def test_overwrite():
--- a/scipy/sparse/csgraph/tests/test_shortest_path.py
+++ b/scipy/sparse/csgraph/tests/test_shortest_path.py
@@ -386,6 +386,24 @@ def test_masked_input():
     for method in methods:
         check(method)
 
+def test_masked_invalid_input():
+    csgraph = np.array(
+        [[0, 1, 0],
+        [1, 0, 0],
+        [0, 0, 0]]
+    )
+    csgraph_masked = np.ma.masked_invalid(csgraph)
+
+    zeros = np.zeros((3, 3))
+
+    def check(method):
+        SP = shortest_path(csgraph_masked, method=method, directed=True,
+                           overwrite=False)
+        assert_array_equal(SP, zeros)
+
+    for method in methods:
+        check(method)
+
 
 def test_overwrite():
     G = np.array([[0, 3, 3, 1, 2],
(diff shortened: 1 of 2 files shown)


scipy/sparse/csgraph/tests/test_shortest_path.py
```diff
@@ -386,6 +386,24 @@ def check(method):
...
+    csgraph_masked = np.ma.masked_invalid(csgraph)
+
+    zeros = np.zeros((3, 3))
+
+    def check(method):
+        SP = shortest_path(csgraph_masked, method=method, directed=True,
+                           overwrite=False)
+        assert_array_equal(SP, zeros)
+
+    for method in methods:
```
j-bowhay, line 404: Can we use `@pytest.mark.parametrize("method", ['FW', 'J', 'BF'])` instead

scipy/sparse/csgraph/tests/test_shortest_path.py
```diff
@@ -386,6 +386,23 @@ def check(method):
...
+        [1, 0, 0],
+        [0, 0, 0]]
+    )
+    csgraph_masked = np.ma.masked_invalid(csgraph)
+
+    zeros = np.zeros((3, 3))
+
+    SP = shortest_path(csgraph_masked, method=method, directed=True,
+                        overwrite=False)
+    assert_array_equal(SP, zeros)
```
j-bowhay, line 403: Sorry one final nit, I think new tests should use `assert_equal` instead as per the numpy docs

czgdp1807:
Resolved everything. Ready from my end.

scipy/sparse/csgraph/tests/test_shortest_path.py
```diff
@@ -387,6 +390,22 @@ def check(method):
         check(method)
 
 
+@pytest.mark.parametrize("method", ['FW', 'J', 'BF'])
+def test_masked_invalid_input(method):
+    # Reference - https://github.com/scipy/scipy/pull/22286
```
j-bowhay, line 395:
```suggestion
    # Reference - https://github.com/scipy/scipy/issues/12424
```

j-bowhay (commented):
While it's not immediately obvious that this is documented behaviour with masked arrays it seems reasonable that the methods in `shortest_path` should behave the same. However I wasn't sure if this will be a breaking change for those using the `floyd_warshall` directly.

I wonder if @jakevdp has an opinion as he was active on the original issue

lucascolley:
gentle ping @jakevdp

PR #22267, opened 2025-01-06
ENH: spatial.transform: baseline implementation of `RigidTransform`

scottshambaugh:
#### Reference issue
Closes https://github.com/scipy/scipy/issues/19254

Forum post: https://discuss.scientific-python.org/t/enh-scipy-spatial-transform-rigidtransformation-for-representing-3d-coordinate-frames/1673

#### What does this implement/fix?
This adds a new `RigidTransform` class to the `scipy.spatial.transform` module, which allows for representing and manipulating rotations and translations in 3D space. Proper rigid transformations are equivalent to the Special Euclidean Group SE(3).

The end goal here is to allow for the representation of arbitrary _coordinate frames_, which is fundamental functionality in the fields of robotics, aerospace, mechanical engineering, computer graphics, and anything that has to do with observing, modeling, analyzing, or controlling the physical world in 3D space.

The class is designed to have a similar interface to the existing `Rotation` class.

#### Additional information

TODO:

- [x] Implement `from_expcoords`, `as_expcoords`, with tests (necessary for baseline?)
- [x] Implement `from_dualquat`, `as_dualquat`, with tests (necessary for baseline?)
- [x] Is current behavior of `random()` what we want? Can defer this if not
- [x] Flesh out documentation for all methods with plenty of examples
- [x] Get concurrence on the naming of the methods and object
- [x] Post on forum for review
- ~~[ ] This is a large change, do we want to make this a beta feature? Target a feature branch? Not sure on how this is usually handled.~~

Out of scope for now, can cut in later:

* Composition with arrays or rotations via operator overloading
* Fractional powers of transformations
* Transformation slerps, interpolations, splines
* `rotate_about` method
* Other representations (screw axis, screw matrix, etc.)
* `approx_equal` method (for rotations this checks an angular distance rather than elements of the rotation matrix, should think about this more)
* cython functions for potential performance improvements

PR #22293, opened 2025-01-09
ENH: Add __repr__ method to scipy.spatial.transform.Rotation

davidkun:
Very small & simple PR, but hopefully provides some QoL improvements to analysts/devs. 😄 

---

This PR implements the `__repr__` method for `Rotation` (in `scipy.spatial.transform`).

#### Additional information

The interactive console and debugger call `repr` to inspect objects. Without a custom `__repr__` implementation, we see something like this:

```python
>>> from scipy.spatial.transform import Rotation as R
>> R.from_quat([0, 1, 0, 1])
<scipy.spatial.transform._rotation.Rotation object at 0x000002281258FD20>
```

As a best practice, the string returned by `__repr__` should match the source code necessary to re-create the represented object. I implemented by falling back on Numpy's `array` representation, so now we have:

````python
>> R.from_quat([0, 1, 0, 1])
Rotation(array([0.        , 0.70710678, 0.        , 0.70710678]))

>> R.from_quat([[0, 1, 0, 1], [0, 0, 1, 1]])
Rotation(array([[0.        , 0.70710678, 0.        , 0.70710678],
        [0.        , 0.        , 0.70710678, 0.70710678]]))

# or, using f-strings:
>>> print(f"{R.from_quat([0, 1, 0, 1])!r}")
Rotation(array([0.        , 0.70710678, 0.        , 0.70710678]))
````

`__repr__` is also a fallback for `str` (`__str__`), so this works too:

```python
>>> print(R.from_quat([0, 1, 0, 1]))
Rotation(array([0.        , 0.70710678, 0.        , 0.70710678]))
````

---

I added a couple of tests and ran the tests + linter, so things should be good. 👍

### DIFF
--- a/scipy/spatial/transform/_rotation.pyx
+++ b/scipy/spatial/transform/_rotation.pyx
@@ -3616,6 +3616,10 @@ cdef class Rotation:
         else:
             return cls.from_matrix(C), rssd
 
+    def __repr__(Rotation self):
+        q = np.asarray(self._quat).squeeze()
+        return f"Rotation({q!r})"
+
 class Slerp:
     """Spherical Linear Interpolation of Rotations.
 
--- a/scipy/spatial/transform/_rotation.pyx
+++ b/scipy/spatial/transform/_rotation.pyx
@@ -3617,8 +3617,7 @@ cdef class Rotation:
             return cls.from_matrix(C), rssd
 
     def __repr__(Rotation self):
-        q = np.asarray(self._quat).squeeze()
-        return f"Rotation({q!r})"
+        return f"Rotation.from_quat({self.as_quat()!r})"
 
 class Slerp:
     """Spherical Linear Interpolation of Rotations.
--- a/scipy/spatial/transform/_rotation.pyx
+++ b/scipy/spatial/transform/_rotation.pyx
@@ -3617,7 +3617,10 @@ cdef class Rotation:
             return cls.from_matrix(C), rssd
 
     def __repr__(Rotation self):
-        return f"Rotation.from_quat({self.as_quat()!r})"
+        m = f"{self.as_matrix()!r}".splitlines()
+        # bump indent (+21 characters)
+        m[1:] = [" " * 21 + m[i] for i in range(1, len(m))]
+        return "Rotation.from_matrix(" + "\n".join(m) + ")"
 
 class Slerp:
     """Spherical Linear Interpolation of Rotations.
--- a/scipy/spatial/transform/tests/test_rotation.py
+++ b/scipy/spatial/transform/tests/test_rotation.py
@@ -1524,6 +1524,22 @@ def test_align_vectors_primary_only():
         assert np.isclose(rssd, 0, atol=atol)
 
 
+def test_repr_single_rotation():
+    q = np.array([[0, 1, 0, 1]]) / np.sqrt(2)
+    actual = repr(Rotation.from_quat(q))
+    # drop whitespace for comparison
+    assert actual.replace(" ", "") == "Rotation(array([0.,0.70710678,0.,0.70710678]))"
+
+
+def test_repr_rotation_sequence():
+    q = np.array([[0, 1, 0, 1], [0, 0, 1, 1]]) / np.sqrt(2)
+    actual = f"{Rotation.from_quat(q)!r}"
+    # drop whitespace for comparison
+    expected = """Rotation(array([[0.,0.70710678,0.,0.70710678],
+                                  [0.,0.,0.70710678,0.70710678]]))""".replace(" ", "")
+    assert actual.replace(" ", "") == expected
+
+
 def test_slerp():
     rnd = np.random.RandomState(0)
 
--- a/scipy/spatial/transform/tests/test_rotation.py
+++ b/scipy/spatial/transform/tests/test_rotation.py
@@ -1527,17 +1527,25 @@ def test_align_vectors_primary_only():
 def test_repr_single_rotation():
     q = np.array([0, 0, 0, 1])
     actual = repr(Rotation.from_quat(q))
-    # drop whitespace for comparison
-    assert actual.replace(" ", "") == "Rotation.from_quat(array([0.,0.,0.,1.]))"
+    expected = """\
+Rotation.from_matrix(array([[1., 0., 0.],
+                            [0., 1., 0.],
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: scipy/spatial/transform/_rotation.pyx [hunk 1]
```diff
@@ -3616,6 +3616,10 @@ cdef class Rotation:
         else:
             return cls.from_matrix(C), rssd
 
+    def __repr__(Rotation self):
+        q = np.asarray(self._quat).squeeze()
```
scottshambaugh, line 3620: Should take out the squeeze, [0,0,0,1] and [[0,0,0,1]] are different.

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1524,6 +1524,22 @@ def test_align_vectors_primary_only():
...
+    # drop whitespace for comparison
+    assert actual.replace(" ", "") == "Rotation(array([0.,0.70710678,0.,0.70710678]))"
+
+
+def test_repr_rotation_sequence():
+    q = np.array([[0, 1, 0, 1], [0, 0, 1, 1]]) / np.sqrt(2)
+    actual = f"{Rotation.from_quat(q)!r}"
+    # drop whitespace for comparison
+    expected = """Rotation(array([[0.,0.70710678,0.,0.70710678],
+                                  [0.,0.,0.70710678,0.70710678]]))""".replace(" ", "")
```
scottshambaugh, line 1539: Can remove the replace here since it's a hardcoded string.

scottshambaugh (commented):
This is nice change. My only thought is whether we want to show quaternions or some other representation. Rotation matrices or rotation vectors might be easier for most people to understand. Thoughts?

scottshambaugh:
I think I would give a preference for the rotation matrix representation, since even though it's more text it's probably what we'll use for the rigid transformations that are getting developed right now.

scipy/spatial/transform/_rotation.pyx
davidkun: Interestingly, when I created it from a single quaternion (`R.from_quat([0,0,0,1])`), it still came out as `Rotation(array([[0.,0.,0.,1.]]))`. (I didn't have the `squeeze()` there until after my test/expectation failed.)

davidkun:
> This is nice change. My only thought is whether we want to show quaternions or some other representation. Rotation matrices or rotation vectors might be easier for most people to understand. Thoughts?

@scottshambaugh, thanks for the quick review and thoughtful comments! 

I had two reasons for using quaternion(s) in `repr`:
1. it's more compact, and
2. `quat` is the first/main parameter in the constructor, so you can literally evaluate the `__repr__` output to build a new instance, i.e., `eval(repr(R.from_quat([0,0,0,1])))` is a clean roundtrip.

However, it sounds like the rotation matrix version is more convenient for you. I'm happy to change it, but perhaps there is middle ground? What if I implement `__str__` too, and have `__str__` be the rotation matrix view, such that:

```python
# repr is true object representation:

>> r = R.from_quat([0, 0, 0, 1])
>> r
Rotation(array([0., 0., 0., 1.]))

# str is user-friendly:

>> print(r)
Rotation(
 [[1. 0. 0.]
  [0. 1. 0.]
  [0. 0. 1.]]
)
```

This way, the string returned by `repr` is a valid Python expression (good practice), but the `str`/`print` outputs are user-friendly and human readable (also good practice). 

What do you think?

scipy/spatial/transform/tests/test_rotation.py
davidkun: that's for the leading space on the second line. the line was too long, so I broke it up. I can also string concat them together with no spaces to avoid the `replace`.

scipy/spatial/transform/_rotation.pyx
scottshambaugh: Ah, yeah you're right the internal representation is the same. `as_quat` changes the output based on the `single` flag, probably best to call `as_quat` in the repr if we keep it as quaternions.

scottshambaugh:
It won't raise an error, but Rotations aren't meant to be created from the constructor directly - we call out in the top level docs that users should make them using the `from_*` methods. That said, point taken that `__repr__` is traditionally meant to be a valid input to eval. Thoughts on something like the below that threads the needle? I like that it also indicates how to interpret the array values.

```
>> r
Rotation.from_quat(array([0., 0., 0., 1.]))
```

@rkern might also have some opinions on what representation we want to show here. I'm hesitant to put something in `__str__` that isn't just prettier than `__repr__` but is a different representation altogether, but tbh don't have too much depth of insight into the best practice conventions here.

davidkun:
> Thoughts on something like the below that threads the needle? I like that it also indicates how to interpret the array values.
> 
> ```python
> >> r
> Rotation.from_quat(array([0., 0., 0., 1.]))
> ```

I like it. I updated the PR with this feedback (plus feedback on the test method, and calling `.as_quat()` internally).

scottshambaugh:
Thanks! I want to get some feedback from other people on what representation (quaternion or rotation matrix) we should show here before merging. Pinging also @hubernikus and @AlexanderFabisch for their opinions bikeshedding this :)

AlexanderFabisch:
> I want to get some feedback from other people on what representation (quaternion or rotation matrix) we should show here before merging. Pinging also @hubernikus and @AlexanderFabisch for their opinions bikeshedding this :)

If you find the from_* form acceptable, I would argue that rotation matrices are more interpretable: each column represents a basis vector. That's often quite easy to visualize / imagine. Quaternions are ... complicated 😅. (Multiple conventions like scalar first vs. scalar last don't make this easier.)

davidkun:
Ok, it sounds like 2 votes for `from_matrix` representation. There was a bit of extra work for the indentation to line up, but I think it's worth it:

```python
>> r = Rotation.from_quat([0., 0., 0., 1.])
>> r
Rotation.from_matrix(array([[1., 0., 0.],
                            [0., 1., 0.],
                            [0., 0., 1.]]))
```

scottshambaugh (approved):
Looks great, thanks for the contribution!

j-bowhay:
Thanks for your first contribution to SciPy @davidkun ! It would be great to have your input on #22267 if that's a feature you'd be interested in

PR #22295, opened 2025-01-10
DOC: signal.ShortTimeFFT.nearest_k_p: fix typo

Gautzilla:
Fixed a grammatical error in the scipy.signal.ShortTimeFFT.nearest_k_p method docstring.

Replace `if [...] than [...] else [...]` wordings with `if [...] then [...] else [...]`

[docs only]


lucascolley (approved):
thanks @Gautzilla !

PR #22292, opened 2025-01-09
MAINT: Update highs subproject commit

HaoZeke:
#### Reference issue

Closes #22257.

#### What does this implement/fix?

As discussed elsewhere, localizes the binding code to prevent conflicts.

#### Additional information

See also:

- https://github.com/scipy/HiGHS/pull/71
- https://github.com/ERGO-Code/HiGHS/pull/2116

PR #22276, opened 2025-01-08
TST: stats.Normal: bump tolerance on test of logcdf

mdhaber:
#### Reference issue
e.g. gh-22274

#### What does this implement/fix?
There is a test failing in main with the latest release of `hypothesis`. It looks like it is able to find a little region where quadrature is less accurate than usual:
```python3
import numpy as np
from scipy import stats
X = stats.Normal(mu=-0.044622861356271804, sigma=1.234783885368829)
X.logcdf(-0.84)       - X.logcdf(-0.84, method='quadrature')        # np.float64(0.0)
X.logcdf(-0.84693408) - X.logcdf(-0.84693408, method='quadrature')  # np.float64(1.1598066906870486e-07)
X.logcdf(-0.85)       - X.logcdf(-0.85, method='quadrature')        # np.float64(0.0)
```

So this bumps the tolerance slightly.

PR #22272, opened 2025-01-07
TST: fix thread safety issue in interpolate.bsplines memmap test

rgommers:
This test was recently introduced in gh-22158. It causes a crash when running the test suite with `pytest-run-parallel`, because of writing to the same tempfile from multiple threads.

Note that the `get_native_id` pattern to make filenames unique is present in multiple other places in the test suite.

PR #22270, opened 2025-01-07
BLD: fix some issues with undeclared internal build dependencies

rgommers:
This is a follow-up to gh-21983, and fixes some more issues reported on gh-18909. It's possible to run into these when building with a high level of parallelism, or trying to build just a single target like `spatial/_qhull.so` with Ninja.

Should be backported to 1.15.1.

Hat tip to @mathstuf for finding these.

PR #22296, opened 2025-01-10
MAINT: backports for 1.15.1

tylerjereddy:
A fairly serious bug (gh-22257) was discovered and patched, warranting a prompt release of `1.15.1`.

Backports included so far:

1. gh-22245
2. gh-22270
3. gh-22272
4. gh-22292
5. gh-22276

TODO:

- [x] regular CI green
- [ ] wheels CI green as precaution before starting release proper (could skip maybe, backport count is small-ish)

- I had a few doc build warnings locally, but we have an issue open for that and it is "ok" in the release env usually I think.
- Locally, `TestCurveFit.test_curvefit_omitnan` had a small tolerance issue; `test_quad_vec_pool` failed because it took too long; I'll probably ignore both if CI is happy though

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -24,7 +24,7 @@
 	url = https://github.com/cobyqa/cobyqa.git
 [submodule "scipy/_lib/array_api_extra"]
 	path = scipy/_lib/array_api_extra
-	url = https://github.com/lucascolley/array-api-extra.git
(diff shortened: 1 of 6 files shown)


tylerjereddy:
The one test failure is gh-22297, let me check that real quick.

tylerjereddy:
Ok, Matt already solved that one, so I've backported that in now.

PR #22301, opened 2025-01-10
ENH: linalg.sqrtm: add structure detection and `assume_a` argument

### LINKED ISSUE #12994: Expose "_sqrtm_triu" as a function
I've encountered a problem where I'd like to find the square root of an upper triangular matrix. (I can provide more specifics if helpful, but this seems like a relatively generic problem users might encounter.)

I was pleasantly surprised to find that `scipy.linalg.sqrtm` calls a helper function [`_sqrtm_triu`](https://github.com/scipy/scipy/blob/v1.5.3/scipy/linalg/_matfuncs_sqrtm.py#L24) which I think accomplishes exactly what I want! I think it would be great to expose this as a function for users -- maybe call it something like `sqrtm_triangular(...)`. This would save the computation time of an unecessary Schur decomposition.

j-bowhay:
#### Reference issue

Closes #12994

In #12994 it was requested that we expose the private function `_sqrtm_triu`. As pointed out by @ilayn, we probably don't want to expose _another_ function to the public API. Instead, this pr does two things:

1. Adds an `assume_a` argument so the user can specify the matrix structure 
2. If no structure is provided then it is automatically determined

This is useful as if the matrix is upper or lower triangular it avoids unnecessarily computing the schur decomposition.

### DIFF
--- a/scipy/linalg/_matfuncs_sqrtm.py
+++ b/scipy/linalg/_matfuncs_sqrtm.py
@@ -201,12 +201,11 @@ def sqrtm(A, disp=True, blocksize=64, *, assume_a='general'):
         X.fill(np.nan)
     else:
         if assume_a == "general":
-            ZH = np.conjugate(Z).T
-            X = Z.dot(R).dot(ZH)
+            X = Z @ R @ Z.conj().T
         elif assume_a == "upper triangular":
-            X = Z
+            X = R
         elif assume_a == "lower triangular":
-            X = Z.T
+            X = R.T
         dtype = np.result_type(A.dtype, 1j if np.iscomplexobj(X) else 1)
         X = X.astype(dtype, copy=False) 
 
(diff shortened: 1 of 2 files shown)


ilayn:
I'm sending a PR soon about this in C which also does the batching internally. Unfortunately it is a bit more involved than this for the real valued case which was a bit of an annoyance for some time as it returns unnecessary complex result for some real square roots.

But the tests are nice so can I ask some time to finish that up?

ilayn:
Apologies, I got lost in NumPy C-API while trying to loop for slices, I'm done now so I can get back to writing the actual sqrtm code. Should not be more than a couple of days.

ilayn:
Some sitrep FYI : I did not forget this and sparing an hour here and there between work. But I nerd-sniped myself with the improvement of the rsf2csf step that was taking way too much time to massage it to work with LAPACK. I hope it will be faster in the end.

PR #22300, opened 2025-01-10
MAINT: remove end year from copyright

SL3333PY:
#### Reference issue

gh-22224

#### What does this implement/fix?

Removed the end year from copyright declarations. 
Updated `LICENSE.txt`, `doc/source/conf.py`.

#### Additional information

N/A

PR #22302, opened 2025-01-11
MAINT: remove unused library import

SL3333PY:
#### Reference issue

Follow-up to #22300

Removed unused library import missed in the previous PR.

#### Additional information

N/A


tupui (approved):
Thanks

dschmitz89:
Unused imports should be picked up by the linter. Have we deactivated that rule?

PR #22304, opened 2025-01-11
ENH: stats.special_ortho_group: speed up, allow 1x1 and 0x0 ortho and unitary groups

### LINKED ISSUE #22303: ENH: improve and simplify special_ortho_group

Unlike `ortho_group` and `unitary_group`, `scipy.stats.special_ortho_group` is implemented using a Python loop over matrix dimensions.

A simpler option would be to generate a Haar-distributed matrix from the orthogonal group, and multiply its first line by the matrix determinant. This ensures that the determinant of the resulting matrix is always $+1$, and because the original matrix is Haar-distributed, so stays the result.

As the timing analysis below shows, this option is also faster.

N/A

Here is the timing script and the resulting timing comparison using scipy 1.15. We see that the proposed simplification is significantly faster for most sizes except for 129, where it is closely comparable. The difference is especially pronounced for small matrix sizes, where I feel that the function is likely more useful.

```py
from timeit import Timer
import numpy as np

setup = """
from scipy.stats import ortho_group, special_ortho_group
import numpy as np
"""

special_ortho = "special_ortho_group.rvs({})".format
ortho = """a = ortho_group.rvs({})
a[0] *= np.linalg.det(a)
""".format

times = []
for dim in np.hstack([np.arange(2, 10), np.logspace(1, 3, 10, dtype=int)]):
    timer = Timer(special_ortho(dim), setup=setup)
    num_calls = timer.autorange()[0] * 5 # 1 second per call
    t_so = min(timer.repeat(repeat=5, number=num_calls)) / num_calls
    timer = Timer(ortho(dim), setup=setup)
    num_calls = timer.autorange()[0] * 5 # 1 second per call
    t_o = min(timer.repeat(repeat=3, number=num_calls)) / num_calls
    times.append((dim, t_so, t_o))

times = np.array(times)

plt.scatter(times[:, 0], times[:, 1], label="current algorithm")
plt.plot(times[:, 0], times[:, 1])
plt.scatter(times[:, 0], times[:, 2], label="proposed alternative")
plt.plot(times[:, 0], times[:, 2])
plt.xscale("log")
plt.yscale("log")
plt.xlabel("Dimension")
plt.ylabel("Time (s)")
plt.legend()
```

image

I will provide a PR shortly, here I am documenting the timing analysis.

akhmerov:

- Closes #22303
- Allows to use symmetric group distributions with `dim=0` and `dim=1`

#### Additional information

The motivation for the second change is that the existing code actually already works with `dim=1` and `dim=0`. Not special-casing small sizes is a general design principle of linear algebra in numpy and scipy, and therefore this simplifies the interface for the users.

### DIFF
--- a/scipy/stats/_multivariate.py
+++ b/scipy/stats/_multivariate.py
@@ -3644,55 +3644,11 @@ class special_ortho_group_gen(multi_rv_generic):
         """
         random_state = self._get_random_state(random_state)
 
-        size = int(size)
-        size = (size,) if size > 1 else ()
-
-        dim = self._process_parameters(dim)
-
-        # H represents a (dim, dim) matrix, while D represents the diagonal of
-        # a (dim, dim) diagonal matrix. The algorithm that follows is
-        # broadcasted on the leading shape in `size` to vectorize along
-        # samples.
-        H = np.empty(size + (dim, dim))
-        H[..., :, :] = np.eye(dim)
-        D = np.empty(size + (dim,))
-
-        for n in range(dim-1):
-
-            # x is a vector with length dim-n, xrow and xcol are views of it as
-            # a row vector and column vector respectively. It's important they
-            # are views and not copies because we are going to modify x
-            # in-place.
-            x = random_state.normal(size=size + (dim-n,))
-            xrow = x[..., None, :]
-            xcol = x[..., :, None]
-
-            # This is the squared norm of x, without vectorization it would be
-            # dot(x, x), to have proper broadcasting we use matmul and squeeze
-            # out (convert to scalar) the resulting 1x1 matrix
-            norm2 = np.matmul(xrow, xcol).squeeze((-2, -1))
-
-            x0 = x[..., 0].copy()
-            D[..., n] = np.where(x0 != 0, np.sign(x0), 1)
-            x[..., 0] += D[..., n]*np.sqrt(norm2)
-
-            # In renormalizing x we have to append an additional axis with
-            # [..., None] to broadcast the scalar against the vector x
-            x /= np.sqrt((norm2 - x0**2 + x[..., 0]**2) / 2.)[..., None]
-
-            # Householder transformation, without vectorization the RHS can be
-            # written as outer(H @ x, x) (apart from the slicing)
-            H[..., :, n:] -= np.matmul(H[..., :, n:], xcol) * xrow
-
-        D[..., -1] = (-1)**(dim-1)*D[..., :-1].prod(axis=-1)
-
-        # Without vectorization this could be written as H = diag(D) @ H,
-        # left-multiplication by a diagonal matrix amounts to multiplying each
-        # row of H by an element of the diagonal, so we add a dummy axis for
-        # the column index
-        H *= D[..., :, None]
-        return H
-
+        q = ortho_group.rvs(dim, size, random_state)
+        dets = np.linalg.det(q)
+        if dim:
+            q[..., 0, :] *= dets[..., np.newaxis]
+        return q
 
 special_ortho_group = special_ortho_group_gen()
 
--- a/scipy/stats/_multivariate.py
+++ b/scipy/stats/_multivariate.py
@@ -3620,9 +3620,9 @@ class special_ortho_group_gen(multi_rv_generic):
 
     def _process_parameters(self, dim):
         """Dimension N must be specified; it cannot be inferred."""
-        if dim is None or not np.isscalar(dim) or dim <= 1 or dim != int(dim):
+        if dim is None or not np.isscalar(dim) or dim < 0 or dim != int(dim):
             raise ValueError("""Dimension of rotation must be specified,
-                                and must be a scalar greater than 1.""")
+                                and must be a scalar nonnegative integer.""")
 
         return dim
 
@@ -3763,9 +3763,9 @@ class ortho_group_gen(multi_rv_generic):
 
     def _process_parameters(self, dim):
         """Dimension N must be specified; it cannot be inferred."""
-        if dim is None or not np.isscalar(dim) or dim <= 1 or dim != int(dim):
+        if dim is None or not np.isscalar(dim) or dim < 0 or dim != int(dim):
             raise ValueError("Dimension of rotation must be specified,"
-                             "and must be a scalar greater than 1.")
+                             "and must be a scalar nonnegative integer.")
 
         return dim
 
@@ -4118,7 +4118,7 @@ class unitary_group_gen(multi_rv_generic):
     Parameters
     ----------
     dim : scalar
-        Dimension of matrices, must be greater than 1.
+        Dimension of matrices.
     seed : {None, int, np.random.RandomState, np.random.Generator}, optional
         Used for drawing random variates.
         If `seed` is `None`, the `~np.random.RandomState` singleton is used.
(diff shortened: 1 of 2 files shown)


akhmerov:
An alternative implementation of `special_ortho_group.rvs` would look like this:

```py
def rvs(self, dim, size=1, random_state=None):
    """Draw random samples from SO(N).

    Parameters
    ----------
    dim : integer
        Dimension of rotation space (N).
    size : integer, optional
        Number of samples to draw (default 1).

    Returns
    -------
    rvs : ndarray or scalar
        Random size N-dimensional matrices, dimension (size, dim, dim)

    """
    random_state = self._get_random_state(random_state)

    size = int(size)

    dim = self._process_parameters(dim)

    size = (size,) if size > 1 else ()
    z = random_state.normal(size=size + (dim, dim))
    q, r = np.linalg.qr(z)
    # The last two dimensions are the rows and columns of R matrices.
    # Extract the diagonals. Note that this eliminates a dimension.
    d = r.diagonal(offset=0, axis1=-2, axis2=-1)
    # Add back a dimension for proper broadcasting: we're dividing
    # each row of each R matrix by the diagonal of the R matrix.
    q *= (d/abs(d))[..., np.newaxis, :]  # to broadcast properly
    # The last column of q may need adjustment to ensure that det(q) = 1.
    if dim:
        q[..., 0, :] *= np.prod(np.sign(d), axis=-1)[..., np.newaxis] * (-1)**(dim - 1)
    return q
```

This relies on `qr` in LAPACK using Householder rotations, each with determinant `-1` to correct the determinant of `Q` without spending `N^3` extra operations on computing it.

scipy/stats/tests/test_multivariate.py
```diff
@@ -1979,7 +1979,7 @@ def test_haar(self):
         dim = 5
         samples = 1000  # Not too many, or the test takes too long
         ks_prob = .05
-        np.random.seed(514)
+        np.random.seed(513)
```
mdhaber, line 1982: The preivious seed happened to fail?

scipy/stats/tests/test_multivariate.py
```diff
@@ -2085,6 +2093,21 @@ def test_haar(self):
         ks_tests = [ks_2samp(proj[p0], proj[p1])[1] for (p0, p1) in pairs]
         assert_array_less([ks_prob]*len(pairs), ks_tests)
 
+    def test_one_by_one(self):
+        # Test that the 1x1 distribution gives ±1 with equal probability.
+        dim = 1
+        xs = ortho_group.rvs(dim, size=5000)
+        np.random.seed(514)
```
mdhaber, line 2100: We're in the process of phasing out use of `np.random.seed` throughout the codebase. Please use `rng = np.random.default_rng(4825489765835639)` and pass `rng` to `rvs`.

scipy/stats/_multivariate.py
```diff
@@ -3644,55 +3644,11 @@ def rvs(self, dim, size=1, random_state=None):
...
-
-        # Without vectorization this could be written as H = diag(D) @ H,
-        # left-multiplication by a diagonal matrix amounts to multiplying each
-        # row of H by an element of the diagonal, so we add a dummy axis for
-        # the column index
-        H *= D[..., :, None]
-        return H
-
+        q = ortho_group.rvs(dim, size, random_state)
+        dets = np.linalg.det(q)
```
mdhaber, line 3648: Would it be worth taking the `sign` to avoid possibly scaling?

scipy/stats/_multivariate.py
```diff
@@ -3644,55 +3644,11 @@ def rvs(self, dim, size=1, random_state=None):
...
+        if dim:
+            q[..., 0, :] *= dets[..., np.newaxis]
```
mdhaber, line 3650: @ilayn could you confirm that this looks OK? My original suggestion was to multiply the whole matrix, but I suppose if one row does the trick and doesn't mess with statistical properties, it's probably faster.

mdhaber (approved):
This looks pretty good to me; just a few things to take a look at.

scipy/stats/tests/test_multivariate.py
mdhaber: Right - does it still appear to pass approximately `1-ks_prob` of the time? 
Please check (roughly) when switching to use of a Generator (https://github.com/scipy/scipy/pull/22304/files/4938d3dcb73fc5f84399edd39db90adf90691fed#diff-a998a313f078eba79aeb6347a65117e0a0c4542a4d778a4cfd398f1737380a71)

scipy/stats/_multivariate.py
akhmerov: Multiplying the whole matrix is wrong: $\det(x A) = x^d \det A$ for a $d$-dimensional matrix $A$.

Multiplying a single row is correct because any matrix with rows `[x1, x2, x3, ...]` is equally likely to be produced by `ortho_group` as matrix `[-x1, -x2, x3, ...]` (and same for an arbitrary other position of the minus sign). Applying the function to each of those equally likely matrices we observed that it does not matter which row we multiply by $-1$.

scipy/stats/tests/test_multivariate.py
akhmerov: I switched to division by determinant. This should result in compensating the error from `ortho_group`, if that error occurs.

scipy/stats/_multivariate.py
mdhaber: I take it https://github.com/scipy/scipy/pull/22304#discussion_r1912487107 was meant to go here. That sounds great!

scipy/stats/tests/test_multivariate.py
```diff
@@ -1930,11 +1930,10 @@ def test_logpdf_4x4(self):
 
 class TestSpecialOrthoGroup:
     def test_reproducibility(self):
-        np.random.seed(514)
-        x = special_ortho_group.rvs(3)
-        expected = np.array([[-0.99394515, -0.04527879, 0.10011432],
-                             [0.04821555, -0.99846897, 0.02711042],
-                             [0.09873351, 0.03177334, 0.99460653]])
+        x = special_ortho_group.rvs(3, random_state=np.random.default_rng(514))
```
mdhaber, line 1933: Oops, `Generator`s don't produce the same results as the global `RandomState` given the same seed.

akhmerov: Thanks, done. It seems that the second half of the test uses legacy numpy random API (`np.random.RandomState`), so I deleted it and adjusted the stored value.

scipy/stats/tests/test_multivariate.py
akhmerov: The test runs for 6 different comparisons, so I believe the success chance is `(1 - ks_prob)**6`. At a glance, however, it seems that everything is fine.

mdhaber: Ah, I didn't notice; I meant individual tests, but that explains why it's not unlikely to fail. Sounds good.

mdhaber (approved):
I'll go ahead and merge when tests pass.

mdhaber:
Thanks @akhmerov! Please develop future PRs from feature branches since it makes it easier for reviewers to work with them, but it was no problem here.

PR #22305, opened 2025-01-11
MAINT, DOC: forward port 1.15.1 relnotes

tylerjereddy:
* Forward port the SciPy `1.15.1` release notes and update the version switcher.

[docs only]

I seem to be having doc build issues locally so make sure the doc build actually passes first. These kind of issues seem more common with the increasing complexity of the interactive/notebook-related infrastructure; hard to avoid some tradeoff I guess.

### DIFF
--- a/doc/source/_static/version_switcher.json
+++ b/doc/source/_static/version_switcher.json
@@ -5,9 +5,14 @@
         "url": "https://scipy.github.io/devdocs/"
     },
     {
-        "name": "1.15.0 (stable)",
-        "version":"1.15.0",
+        "name": "1.15.1 (stable)",
+        "version":"1.15.1",
         "preferred": true,
+        "url": "https://docs.scipy.org/doc/scipy-1.15.1/"
+    },
+    {
+        "name": "1.15.0",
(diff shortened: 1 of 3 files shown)


rgommers (commented):
Something is still off with the version switcher, both in the rendered preview of this PR and with the deployed 1.15.1 docs. Both think that 1.15.0 is stable. E.g., the "Switch to stable version" button:

image

links to https://docs.scipy.org/doc/scipy-1.15.0/index.html.

tylerjereddy:
make upload hasn't happened yet for docs.scipy.org for 1.15.1; I can do that in a bit

tylerjereddy:
FWIW, I did `make upload` within `docs.scipy.org` repo proper just now, in case that helps.

tylerjereddy:
Let's see if merging helps, I don't think anything has changed in my procedure.

tylerjereddy:
Yes, the problem went away after the doc build that happens *after* merging this PR in the new `main` branch.

This is a perpetual source of confusion; nonetheless, things seems to work out in the end every time.

PR #22240, opened 2025-01-04
ENH: array types: add dask.array support

lucascolley:
Continues gh-20956.

---

#### Reference issue
Towards gh-18867

#### What does this implement/fix?
Test with `dask.array` via array-api-compat.

#### Problems so far

Dask:
- dask/dask#11287
- https://github.com/dask/dask/pull/11288
- https://github.com/dask/dask/issues/4368
- https://github.com/dask/dask/issues/3245
- https://github.com/dask/dask/issues/11398

array-api-compat:
- data-apis/array-api-compat#176
- data-apis/array-api-compat#139
- No `sort` or `argsort`

PR #22308, opened 2025-01-12
TST: `_lib`: run tests with `@jax.jit`

crusaderky:
Compile tested functions with jax.jit.
This same infrastructure will be reused after we introduce Dask support (https://github.com/scipy/scipy/pull/22240) to ensure that Dask graphs aren't computed implicitly.

This design is the least intrusive I could come up with.
The obvious drawback is that there is nothing ensuring that we didn't forget to tag a function after importing it. There is also nothing preventing calls directly to the module namespace itself, which would not be jitted.

This PR introduces a new module, `scipy._lib._lazy_testing`. I plan to move this whole module and its tests to `array_api_extra`. The chief motivation is that the parameterised tests for `at` and more functions there need it too, and it is generically very useful.
We can merge this PR as-is now and refactor after we have a release of array-api-extra that includes this functionality, or we can leave this PR unmerged until that happens. Anyway, I wanted to have a clean, comprehensive PR that demonstrates the whole machinery end-to-end.

CC @rgommers

### DIFF
--- a/scipy/_lib/_lazy_testing.py
+++ b/scipy/_lib/_lazy_testing.py
@@ -1,0 +1,111 @@
+from collections.abc import Callable, Iterable, Sequence
+from types import ModuleType
+import pytest
+from scipy._lib._array_api import is_jax
+
+
+def lazy_xp_function(
+    func: Callable,
+    *,
+    jax_jit: bool = True,
+    static_argnums: int | Sequence[int] | None=None,
+    static_argnames: str | Iterable[str] | None=None,
+) -> None:
+    """Tag a function, which must be imported in the test module globals,
+    so that when any tests defined in the same module are executed with
+    xp=jax.numpy the function is replaced with a jitted version of itself.
+
+    This will be later expanded to provide test coverage for other lazy backends,
+    e.g. Dask.
+
+    Example::
+
+      # test_mymodule.py:
+      from scipy._lib._lazy_testing import lazy_xp_function
+      from scipy.mymodule import myfunc
+
+      lazy_xp_function(myfunc)
+
+      def test_myfunc(xp):
+          a = xp.asarray([1, 2])
+          # When xp=jax.numpy, this is the same as
+          # b = jax.jit(myfunc)(a)
+          b = myfunc(a)
+
+    Parameters
+    ----------
+    func : callable
+        Function to be tested
+    jax_jit : bool, optional
+        Set to True to replace `func` with `jax.jit(func)` when calling the
+        `patch_lazy_xp_functions` test helper with `xp=jax.numpy`.
+        Set to False if `func` is only compatible with eager (non-jitted) JAX.
+        Default: True.
+    static_argnums : int | Sequence[int], optional
+        Passed to jax.jit.
+        Positional arguments to treat as static (trace- and compile-time constant).
+        Default: infer from static_argnames using `inspect.signature(func)`.
+    static_argnames : str | Iterable[str], optional
+        Passed to jax.jit.
+        Named arguments to treat as static (compile-time constant).
+        Default: infer from static_argnums using `inspect.signature(func)`.
+
+    Notes
+    -----
+    A test function can circumvent this monkey-patching system by calling `func` an
+    attribute of the original module. You need to sanitize your code to
+    make sure this does not happen.
+
+    Example::
+
+      import mymodule
+      from mymodule import myfunc
+
+      lazy_xp_function(myfunc)
+
+      def test_myfunc(xp):
+          a = xp.asarray([1, 2])
+          b = myfunc(a)  # This is jitted when xp=jax.numpy
+          c = mymodule.myfunc(a)  # This is not
+
+    See Also
+    --------
+    patch_lazy_xp_functions
+    jax.jit: https://jax.readthedocs.io/en/latest/_autosummary/jax.jit.html
+    """
+    if jax_jit:
+        func._lazy_jax_jit_kwargs = {  # type: ignore[attr-defined]
+            "static_argnums": static_argnums,
+            "static_argnames": static_argnames,
+        }
+
+
+def patch_lazy_xp_functions(
+    xp: ModuleType, request: pytest.FixtureRequest, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """If xp==jax.numpy, search for all functions which have been tagged by
+    `lazy_xp_function` in the globals of the module that defines the current test
+    and wrap them with `jax.jit`. Unwrap them at the end of the test.
+
+    Parameters
+    ----------
+    xp: module
+        Array namespace to be tested
+    request: pytest.FixtureRequest
+        Pytest fixture, as acquired by the test itself or by one of its fixtures.
+    monkeypatch: pytest.MonkeyPatch
+        Pytest fixture, as acquired by the test itself or by one of its fixtures.
+
+    See Also
+    --------
+    lazy_xp_function
+    https://docs.pytest.org/en/stable/reference/reference.html#std-fixture-request
+    """
+    if is_jax(xp):
+        import jax
+
+        globals_ = request.module.__dict__
+        for name, func in globals_.items():
+            kwargs = getattr(func, "_lazy_jax_jit_kwargs", None)
+            if kwargs is not None:
+                monkeypatch.setitem(globals_, name, jax.jit(func, **kwargs))
--- a/scipy/_lib/tests/test__util.py
+++ b/scipy/_lib/tests/test__util.py
@@ -382,6 +382,8 @@ class TestContainsNaN:
         xp_assert_equal(_contains_nan(x, "propagate"), xp.asarray(False))
         xp_assert_equal(_contains_nan(x, "omit", xp_omit_okay=True), xp.asarray(False))
         # Lazy arrays don't support "omit" and "raise" policies
+        # TODO test that we're emitting a user-friendly error message.
+        #      Blocked by https://github.com/data-apis/array-api-compat/pull/228
         with pytest.raises(TypeError):
             _contains_nan(x, "omit")
         with pytest.raises(TypeError):
--- a/scipy/_lib/tests/test__util.py
+++ b/scipy/_lib/tests/test__util.py
@@ -14,12 +14,18 @@ from scipy.conftest import skip_xp_invalid_arg
 
 from scipy._lib._array_api import (xp_assert_equal, xp_assert_close, is_numpy,
                                    is_array_api_strict)
+from scipy._lib._lazy_testing import lazy_xp_function
 from scipy._lib._util import (_aligned_zeros, check_random_state, MapWrapper,
                               getfullargspec_no_self, FullArgSpec,
                               rng_integers, _validate_int, _rename_parameter,
                               _contains_nan, _rng_html_rewrite, _lazywhere)
 import scipy._lib.array_api_extra as xpx
 from scipy import cluster, interpolate, linalg, optimize, sparse, spatial, stats
+
+
+lazy_xp_function(_contains_nan, static_argnames=("nan_policy", "xp_omit_okay", "xp"))
+# FIXME @jax.jit fails: complex bool mask
+lazy_xp_function(_lazywhere, jax_jit=False, static_argnames=("f", "f2"))
 
 
 @pytest.mark.slow
@@ -343,8 +349,9 @@ class TestContainsNaN:
         data4 = np.array([["1", 2], [3, np.nan]], dtype='object')
         assert _contains_nan(data4)
 
+    @pytest.mark.skip_xp_backends("jax.numpy", reason="lazy backends tested separately")
     @pytest.mark.parametrize("nan_policy", ['propagate', 'omit', 'raise'])
-    def test_array_api(self, xp, nan_policy):
+    def test_array_api_eager(self, xp, nan_policy):
         rng = np.random.default_rng(932347235892482)
         x0 = rng.random(size=(2, 3, 4))
         x = xp.asarray(x0)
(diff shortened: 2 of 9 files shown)


scipy/_lib/tests/test__util.py
```diff
@@ -19,9 +19,17 @@
...
 import scipy._lib.array_api_extra as xpx
+from scipy.conftest import lazy_jax_jit
 from scipy import cluster, interpolate, linalg, optimize, sparse, spatial, stats
 
 
+lazy_jax_jit(_contains_nan,
+             static_argnums=(1, ),
+             static_argnames=("nan_policy", "xp_omit_okay", "xp"))
+# FIXME @jax.jit fails: complex bool mask
+# lazy_jax_jit(_lazywhere, static_argnums=(2, 4), static_argnames=("f", "f2"))
```
crusaderky, line 30: If we apply enough rigor in our patterns, we could search for `# lazy_jax_jit` (with initial hash) to spot all cases where the jit doesn't work yet.

scipy/_lib/tests/test__util.py
```diff
@@ -343,8 +351,9 @@ def test_contains_nan_with_strings(self):
         data4 = np.array([["1", 2], [3, np.nan]], dtype='object')
         assert _contains_nan(data4)
 
+    @pytest.mark.skip_xp_backends("jax.numpy", reason="lazy backends tested separately")
```
crusaderky, line 354: TODO add dask

scipy/_lib/tests/test__util.py
```diff
@@ -358,9 +367,38 @@ def test_array_api(self, xp, nan_policy):
...
+    def test_array_api_lazy(self, xp):
+        rng = np.random.default_rng(932347235892482)
+        x0 = rng.random(size=(2, 3, 4))
+        x = xp.asarray(x0)
+
+        xp_assert_equal(_contains_nan(x), xp.asarray(False))
+        xp_assert_equal(_contains_nan(x, "propagate"), xp.asarray(False))
+        xp_assert_equal(_contains_nan(x, "omit", xp_omit_okay=True), xp.asarray(False))
+        # Lazy arrays don't support "omit" and "raise" policies
+        with pytest.raises(TypeError):
```
crusaderky, line 387: TODO get an error message that's comprehensible to the final user. Blocked by https://github.com/data-apis/array-api-compat/pull/228

scipy/_lib/_lazy_testing.py
```diff
@@ -0,0 +1,111 @@
...
+    --------
+    patch_lazy_xp_functions
+    jax.jit: https://jax.readthedocs.io/en/latest/_autosummary/jax.jit.html
+    """
+    if jax_jit:
+        func._lazy_jax_jit_kwargs = {  # type: ignore[attr-defined]
+            "static_argnums": static_argnums,
+            "static_argnames": static_argnames,
+        }
+
```
crusaderky, line 81: When we add Dask (https://github.com/scipy/scipy/pull/22240), here we'll add:
```python
if disable_dask_compute:
    func._disable_dask_compute = True
```

scipy/_lib/_lazy_testing.py
```diff
@@ -0,0 +1,111 @@
...
+    https://docs.pytest.org/en/stable/reference/reference.html#std-fixture-request
+    """
+    if is_jax(xp):
+        import jax
+
+        globals_ = request.module.__dict__
+        for name, func in globals_.items():
+            kwargs = getattr(func, "_lazy_jax_jit_kwargs", None)
+            if kwargs is not None:
+                monkeypatch.setitem(globals_, name, jax.jit(func, **kwargs))
```
crusaderky, line 111: When we'll add dask, here we'll replace the function with a wrapped version of itself that runs inside the context manager

```python
@wraps(func)
def wrapper(*arg, **kwargs):
    with dask.config({"dask.scheduler": get_raise}):
        return func(*args, **kwargs)
```
Where
```python
def get_raise(dsk, keys, **kwargs):
    raise AssertionError("Detected implicit call to dask.compute() or dask.persist()")
```

In other words: while you are calling the function, any call to `bool()`, `np.asarray()`, etc. will raise.
However, when you are calling `xp_assert_close` on the function output, the graph will be materialized as normal.

scipy/_lib/tests/test__util.py
```diff
@@ -22,6 +23,11 @@
 from scipy import cluster, interpolate, linalg, optimize, sparse, spatial, stats
 
 
+lazy_xp_function(_contains_nan, static_argnames=("nan_policy", "xp_omit_okay", "xp"))
+# FIXME @jax.jit fails: complex bool mask
+lazy_xp_function(_lazywhere, jax_jit=False, static_argnames=("f", "f2"))
```
crusaderky, line 28: It should be easy enough to search for all instances of `jax_jit=False` in the codebase to figure out what is still not working

scipy/_lib/tests/test_array_api.py
```diff
@@ -62,6 +67,8 @@ def test_array_api_extra_hook(self):
     def test_copy(self, xp):
         for _xp in [xp, None]:
             x = xp.asarray([1, 2, 3])
+            # When _xp=None, on JAX this also tests calling
+            # array_namespace() inside @jax.jit
```
crusaderky, line 71:
```suggestion
```

scipy/_lib/tests/test__util.py
crusaderky: The purpose of this syntax, as opposed to either
- commenting out the whole line, or
- not writing the line at all

is for forward compatibility with Dask, so that we can choose to disable jitting on jax while keeping the compute guard on dask, or vice versa.

scipy/conftest.py
```diff
@@ -209,7 +210,7 @@ def num_parallel_threads():
     pytest.param(v, id=k, marks=pytest.mark.array_api_backends)
     for k, v in xp_available_backends.items()
 ])
-def xp(request):
+def xp(request, monkeypatch):
```
rgommers, line 213: This is going to interfere with [pytest-run-parallel](https://github.com/Quansight-Labs/pytest-run-parallel). Just noting for now. We don't really need to test jitting and free-threading safety at the same time, but test markers may start interfering with each other.

scipy/_lib/tests/test__util.py
```diff
@@ -343,8 +349,9 @@ def test_contains_nan_with_strings(self):
         data4 = np.array([["1", 2], [3, np.nan]], dtype='object')
         assert _contains_nan(data4)
 
+    @pytest.mark.skip_xp_backends("jax.numpy", reason="lazy backends tested separately")
     @pytest.mark.parametrize("nan_policy", ['propagate', 'omit', 'raise'])
-    def test_array_api(self, xp, nan_policy):
+    def test_array_api_eager(self, xp, nan_policy):
```
rgommers, line 354: We should probably avoid this change, since "eager" is the default and should remain implicit - otherwise we're going to be renaming many thousands of tests, probably inconsistently.

scipy/_lib/tests/test__util.py
rgommers: I'm wondering if eager `jax.numpy` still has value to test, since it works and the JIT tests are probably going to end up being very slow. Not sure how much slower - what are you seeing for this test?

scipy/_lib/tests/test_array_api.py
```diff
@@ -7,6 +7,11 @@
 )
 from scipy._lib import array_api_extra as xpx
 from scipy._lib._array_api_no_0d import xp_assert_equal as xp_assert_equal_no_0d
+from scipy._lib._lazy_testing import lazy_xp_function
+
+lazy_xp_function(_asarray, static_argnames=(
+                 "dtype", "order", "copy", "xp", "check_finite", "subok"))
+lazy_xp_function(xp_copy, static_argnames=("xp", ))
```
rgommers, line 14: This looks nice as a pattern - 1-2 lines per function if jitting works without any other changes.

rgommers (commented):
Thanks @crusaderky, looks pretty good at first glance. A few initial questions and comments.

scipy/conftest.py
crusaderky: We can disable all monkey-patching when we detect parallel running I suppose?

crusaderky:
> I'm wondering if eager jax.numpy still has value to test, since it works and the JIT tests are probably going to end up being very slow. Not sure how much slower - what are you seeing for this test?

It looks like there is no performance difference at all.

```
$ python dev.py test -b jax.numpy -t scipy._lib.tests.test_array_api -t scipy._lib.tests.test__util -- -k jax

main (no jit)
15 passed, 2 skipped, 51 deselected in 7.71s
This PR (jit)
13 passed, 5 skipped, 51 deselected in 7.89s 
```

I'm _fairly_ sure that I'm not caching on disk; `jax.config.jax_compilation_cache_dir` returns None.

scipy/_lib/tests/test__util.py
crusaderky: It looks like there is no measurable performance difference.

```
$ python dev.py test -b jax.numpy -t scipy._lib.tests.test_array_api -t scipy._lib.tests.test__util -- -k jax

main (no jit)
15 passed, 2 skipped, 51 deselected in 7.71s
This PR (jit)
13 passed, 5 skipped, 51 deselected in 7.89s 
```

I'm _fairly_ sure that I'm not caching on disk; `jax.config.jax_compilation_cache_dir` returns None.

scipy/conftest.py
rgommers: Yes, that should work I hope.

I just tested, and can't make it fail locally. I think it'd need both free-threading and JAX, and that combo isn't yet available. And then testing like:
```
$ pixi r test-nogil -b jax.numpy -s _lib -- --parallel-threads=8
```

scipy/_lib/tests/test__util.py
rgommers: Seems reasonable. If the line wasn't written at all, then the next person looking at this function would be rediscovering the problem probably.

Is it a FIXME for this PR, or a future one?

scipy/conftest.py
crusaderky: You should definitely be able to make it fail with the GIL too. Probably your test data is small enough that the GIL is never released long enough for a context switch between two tests.

scipy/_lib/tests/test__util.py
crusaderky: Future. I intend to move `_lazywhere` to array-api-extra and fix it for good. A fairly sizeable change by itself.

scipy/_lib/tests/test__util.py
rgommers: Great, let's give it a try as is then.

rgommers:
I just tested a bit locally - not yet done, but I had no trouble understanding what is happening and could make a test for `cluster.whiten` pass under `jax.jit` with the following change:
```diff
diff --git a/scipy/cluster/tests/test_vq.py b/scipy/cluster/tests/test_vq.py
index 055143ae57..93e3de5806 100644
--- a/scipy/cluster/tests/test_vq.py
+++ b/scipy/cluster/tests/test_vq.py
@@ -19,6 +19,12 @@ from scipy._lib import array_api_extra as xpx
 from scipy._lib._array_api import (
     SCIPY_ARRAY_API, array_namespace, xp_copy, xp_assert_close, xp_assert_equal
 )
+from scipy._lib._lazy_testing import lazy_xp_function
+
+
+lazy_xp_function(whiten, static_argnames=("check_finite"))
+
+
 
 skip_xp_backends = pytest.mark.skip_xp_backends
 
@@ -92,7 +98,7 @@ class TestWhiten:
                           [0.87545741, 0.00735733],
                           [0.85124403, 0.26499712],
                           [0.45067590, 0.45464607]])
-        xp_assert_close(whiten(obs), desired, rtol=1e-5)
+        xp_assert_close(whiten(obs, check_finite=False), desired, rtol=1e-5)
 
     @pytest.fixture
     def whiten_lock(self):
diff --git a/scipy/cluster/vq.py b/scipy/cluster/vq.py
index ddf4f0a666..5f703d70ea 100644
--- a/scipy/cluster/vq.py
+++ b/scipy/cluster/vq.py
@@ -137,7 +137,7 @@ def whiten(obs, check_finite=True):
     obs = _asarray(obs, check_finite=check_finite, xp=xp)
     std_dev = xp.std(obs, axis=0)
     zero_std_mask = std_dev == 0
-    std_dev = xpx.at(std_dev, zero_std_mask).set(1.0)
+    std_dev = xp.where(zero_std_mask, 1.0, std_dev)
     if check_finite and xp.any(zero_std_mask):
         warnings.warn("Some columns have standard deviation zero. "
                       "The values of these columns will not change.",

File: scipy/conftest.py (hunk 8 above)
Comment (rgommers) [reply scipy/conftest.py:214]: Yeah, if I try really hard with heavy enough functions, it should be possible. Too much trouble though; this will be way easier when `cp313t` JAX arrives (soon hopefully).

**Review by rgommers — APPROVED** [review]
One question on `array-api-compat` pulling in yes or no, otherwise this looks quite good to me!

**Comment by rgommers** [comment]
> The obvious drawback is that there is nothing ensuring that we didn't forget to tag a function after importing it.

That seems okay for now. Once we start using this at scale for a while, we may find that we need to test for completeness here - that may then intersect with a way of generating reference docs that indicate what array types a function supports.

**Review by lucascolley — APPROVED** [review]
LGTM modulo https://github.com/scipy/scipy/pull/22308/files#r1915449009!

File: scipy/_lib/tests/test__util.py (hunk 3 above)
Comment (crusaderky) [reply scipy/_lib/tests/test__util.py:385]: No, there is much more widespread work that needs to happen downstream of that PR

Comment (crusaderky) [reply scipy/_lib/tests/test__util.py:385]: Added TODO to the code

File: scipy/conftest.py (hunk 8 above)
Comment (crusaderky) [reply scipy/conftest.py:214]: it should be as simple as

```python
def pytest.mark.parametrize("repeat", range(100))
def test1(xp, repeat):
    sleep(np.random.random() / 10)   # 0~100ms
    x = xp.asarray(...)
    y = jitted_function(x)
    xp_assert_equal(y, xp.asarray(...))
```

if [pytest-run-parallel](https://github.com/Quansight-Labs/pytest-run-parallel) doesn't parallelize different iterations of `parametrize`, you'll need two identical tests running in parallel.

xp_assert_equal should crash on xp!=jax.numpy saying that y has namespace jax.numpy.

crusaderky:
I'd rather get this merged without the bump in array-api-compat for now

scipy/conftest.py
rgommers: It works a bit differently, it runs a single pytest test, rewritten to run the same code in multiple threads within that one test (see https://github.com/Quansight-Labs/pytest-run-parallel/pull/24). It maximizes the chance of collisions - only in the absence of the GIL.

Issue #22309, opened 2025-01-12
DOC: optimize.elementwise.find_minimum: harmonize documented/implemented interval termination condition


mdhaber:
### Issue with current documentation:

The documentation of [`optimize.elementwise.find_minimum`](https://docs.scipy.org/doc/scipy-1.15.0/reference/generated/scipy.optimize.elementwise.find_minimum.html) states:
<img width="983" alt="image" src="https://github.com/user-attachments/assets/f2d59163-9a4f-448e-9276-a2d95036deff" />

However, the x-interval criterion in the code is:
https://github.com/scipy/scipy/blob/37ea990aecf3e498d4dd0b23def0477d885ec6c5/scipy/optimize/_chandrupatla.py#L519-L524

which is consistent with the [reference](https://www.sciencedirect.com/science/article/pii/S0045782597001904?via%3Dihub):
<img width="1030" alt="image" src="https://github.com/user-attachments/assets/392e5176-52e2-4edd-8cc2-64a9123ae814" />

Presumably, I documented what I thought the criterion *should* be, perhaps with the intent of making it so, but ultimately I couldn't bring myself to diverge from what was written in the paper.

### Idea or request for content:

The documentation and implementation should be consistent. 

I still think what is documented would be more useful (because then `xatol` would directly control the tolerance on the location of the minimizer), but I suppose it would be easier to change the documentation, and we couldn't really be faulted for implementing what the paper says. (Any idea why it was written that way?)

steppi:
> Any idea why it was written that way?)

I think it's because, by the nature of the algorithm, the minimizer $x^{*}$ should be closer to $x_2$ than it is to $x_3$, so if
$\left|x_2 - x_3\right| < 2\mathrm{xtol}$ then it follows $\left|x_2 - x^{*}\right| < \mathrm{xtol}$.

So as it turns out, I think it's written this way so that `xatol` will actually directly control the tolerance on the minimizer. We should test it out, but I think we can trust `xatol` as is and update the documentation to match what's in Chandrupatla's paper.

mdhaber:
Hmm so you mean that we are bounding the true error in our approximation directly rather than the width of the bracket. If I'm thinking through the implications correctly, this means that the algorithm may be able to terminate earlier than if we were to control the width of the bracket with no difference in the uncertainty of the estimate?

[...]

PR #15932, opened 2022-04-03
MAINT: stats: fix and thoroughly test `rv_sample` at non-integer xk

mdhaber:
#### Reference issue
Closes gh-3758

#### What does this implement/fix?
gh-3758 noted that some methods of `rv_discrete` distributions defined by `values` (`xk`/`pk` pairs, not formulas) work for non-integral `xk`, but not all do. The situation was improved by gh-5672, but more work was needed. This PR aims to fix the methods that were not working as expected and test all methods thoroughly.

There are two main problems:
* in `rv_discrete`, the `pmf` was defined to be zero for non-integral `xk`. The solution is simply to add this condition only for `rv_discrete` that are not instances of `rv_sample` (that is, defined by `values`)
* `rvs` seems to have problems, too, but I haven't looked into it closely yet. Should be easy enough to fix, because passing uniformly distributed random variates into the `ppf` works just fine.

PR #22312, opened 2025-01-12
ENH: stats.Binomial: add binomial distribution with new infrastructure

### LINKED ISSUE #17891: BUG: inconsistent checks for integrality in several distributions
In the infrastructure for discrete distributions, we have a [check](https://github.com/scipy/scipy/blob/v1.10.0/scipy/stats/_discrete_distns.py#L25-L26) whether certain parameters are integral, which gets used in 6 distributions:
* binom
* betabinom
* boltzmann
* hypergeom
* nhypergeom
* randint

This check explicitly lets floats pass that are just integers in disguise, i.e. `_isintegral(1.0) == True`. However, those values are then propagated to underlying and _typed_ methods that barf on floats, for (the respective parameters of) all the above distributions, leading to very unintuitive error-messages.

```
>>> import scipy.stats
>>> scipy.stats.binom.rvs(1.0, 0.5)
  File "<stdin>", line 1, in <module>
  File "/home/[xxx]/.conda/envs/scipy_env/lib/python3.9/site-packages/scipy/stats/_distn_infrastructure.py", line 3357, in rvs
    return super().rvs(*args, **kwargs)
  File "/home/[xxx]/.conda/envs/scipy_env/lib/python3.9/site-packages/scipy/stats/_distn_infrastructure.py", line 1048, in rvs
    vals = self._rvs(*args, size=size, random_state=random_state)
  File "/home/[xxx]/.conda/envs/scipy_env/lib/python3.9/site-packages/scipy/stats/_discrete_distns.py", line 62, in _rvs
    return random_state.binomial(n, p, size)
  File "mtrand.pyx", line 3439, in numpy.random.mtrand.RandomState.binomial
TypeError: Cannot cast scalar from dtype('float64') to dtype('int64') according to the rule 'safe'
>>> scipy.__version__
'1.10.0'
>>> numpy.__version__
'1.24.1'
```

I think we should keep the integrality checks as-is, but cast those parameters to integers as appropriate. It's not 100% obvious to me how this should be handled in the context of the dist infrastructure, which calls `_argcheck` in several places, but doesn't have capabilities to signal that certain parameters now should be cast to integers.

Still, I could imagine a similar private method `_cast_params(*args)`, which would get called _after_ each invocation to `_argcheck` in [`_distn_infrastructure.py`](https://github.com/scipy/scipy/blob/main/scipy/stats/_distn_infrastructure.py), and that could be overridden by each distribution as necessary (just as they're already overriding `_argcheck`).

mdhaber:
#### Reference issue
Closes gh-17891[^1]
Addresses gh-18564
Addresses gh-11026[^2]

This adds the binomial distribution in the new infrastructure as `scipy.stats.Binomial`. To support discrete RVs, I renamed `ContinuousDistribution` to `_UnivariateDistribution` and created subclasses `ContinuousDistribution` and `DiscreteDistribution`. The `DiscreteDistribution` subclass doesn't require too much modification:

- Wherever "quadrature" is done, use `nsum` on the PMF instead of `tanhsinh` on the PDF. ([comment](https://github.com/scipy/scipy/pull/22312#discussion_r1912521674), [comment]( https://github.com/scipy/scipy/pull/22312#discussion_r1912522530))
- When using "quadrature" to evaluate a distribution function, replace the argument with `floor(x)` (CDF) or `floor(x+1)` (CCDF).  ([comment](https://github.com/scipy/scipy/pull/22312#discussion_r1912524834))
- When inverting a distribution function numerically, set `xatol=1` and use the floor of the right end of the resulting bracket (and handle edge cases). ([comment](https://github.com/scipy/scipy/pull/22312#discussion_r1912522132), [comment](https://github.com/scipy/scipy/pull/22312#discussion_r1912525124))
- When finding the mode with numerical optimization, set a coarse[^3] `xatol` and check the `floor` and `ceil` of the result to determine which is the mode[^3]. ([comment](https://github.com/scipy/scipy/pull/22312#discussion_r1912522746))
- Adjust the `_set_invalid_nan` decorator so PMF at non-integral args is 0, and make sure the distribution functions at the left end of the support are NOT replaced (as they can be for continuous distributions). ([comment](https://github.com/scipy/scipy/pull/22312#discussion_r1912519492), [comment](https://github.com/scipy/scipy/pull/22312#discussion_r1912519877))

It also updates `make_distribution` for use with discrete distributions. All transformations are explicitly disabled (for now); thoughts on follow-ups below.

#### Additional information
The PR is in a draft state.
- Formulas for `Binomial` should be added where possible. (Relying on the infrastructure to compute everything from the PMF has been useful during informal testing.)
- We need to add tests. It would be good for someone else to take a close look at the testing strategy to see if we missed anything for `ContinuousDistribution`. @steppi, would you be willing to write them, perhaps adding formulas for the binomial distribution at the same time (so the tests can easily compare the formulas against the generic implementations)?
- I need to check which transformations work with `DiscreteDistribution`. Transformations should probably only accept instances of `ContinuousDistribution` by default (and raise otherwise). (*See https://github.com/scipy/scipy/pull/22312#issuecomment-2588114565.*)
- The "correct" behavior of two-arg distribution functions is not obvious when `y < x`. Currently, the input validation of the public functions swaps the arguments and flips the sign, which may not be what we want. Other behavior that might be more natural include (e.g. for `cdf`):
  - `cdf(y) - cdf(x) + pmf(x)`
  - 0
  - NaN

One shortcoming I'd rather address in a follow-up: currently, methods that work with numerical inversion are slow to recognize the special case when there is no valid bracket within the support. They get the right result (e.g. `Binomial(n=10, p=0.5).icdf(1e-10)` returns `0`) and it's not too slow when the distribution has compact support, but it can be slow when the support is infinite. It would be good to recognize that the support is not a valid bracket and special-case accordingly.

[^1]: In the sense that the *new* infrastructure replaces non-integral values of integral shape parameters with NaN, and consequently, methods will produce NaNs. This is consistent with the policy established in gh-21050: an error due to one bad element of an array should not cause the rest to fail.
[^2]: In the old infrastructure, `ppf(0)` and `isf(1)` are defined to be the left end of the support minus 1 (e.g. `stats.binom(n=10, p=0.5).ppf(0)  # 0`). Since `cdf` and `ccdf` are not invertible for discrete distributions, there are multiple ways `icdf` and `iccdf` can be defined, but I don't think the old definition was as useful as the new one, which returns the left end of the support. This is what R discrete distributions (e.g. `qbinom(0, 10, 0.5)`) and Mathematica (`Quantile[BinomialDistribution[10, 0.5], 0]`) seem to do. See the revised documentation of CDF/CCDF functions.
[^3]: See inline comment.
[^4]: Assumes unimodality, but that's an underlying assumption of the bracket finder / minimizer. The [documentation of `mode`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.Normal.mode.html#scipy.stats.Normal.mode) addresses this.


scipy/integrate/_tanhsinh.py
```diff
@@ -1308,7 +1308,7 @@ def _integral_bound(f, a, b, step, args, constants, xp):
         tol = special.logsumexp(xp.stack((tol, rtol + lb.integral)), axis=0)
     else:
         tol = tol + rtol*lb.integral
-    i_skip = lb.status < 0  # avoid unnecessary f_evals if integral is divergent
+    i_skip = lb.status == -3  # avoid unnecessary f_evals if integral is divergent
```
mdhaber, line 1311: For infinite sums, `nsum` estimate the magnitude of the sum with an integral before doing anything else. This allows us to turn the relative tolerance into an absolute tolerance, which is needed when figuring out how many terms of the sum to evaluate.

This line was supposed to provide an early-exit for divergent integrals, but it (incorrectly) assumes that the only reason for failure status is that the integral is divergent. When testing with the discrete Laplace distribution, I saw that this integral failed to meet the desired tolerance when integrating across the centerline of the distribution. This is reasonable, since the function is not smooth. Consequently, the result of the whole operation was NaN, but if we remove the early-exit, we get an accurate result. The fix is to make the condition stricter - only fail here if the result of the integral is non-finite. It's a bit slower in the exceptional case, but I think that's a good trade.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -366,44 +368,6 @@ def contains(self, item, parameter_values=None):
         in_right = item <= b if right_inclusive else item < b
         return in_left & in_right
 
-
-class _RealDomain(_SimpleDomain):
```
mdhaber, line 370: This is mostly just *moved* to below, and I moved the `draw` method to `_SimpleDomain`.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -448,8 +415,8 @@ def draw(self, n, type_, min, max, squeezed_base_shape, rng=None):
 
         elif type_ == 'out':
             # make this work for infinite bounds
-            z = min_nn - rng.uniform(size=shape)
-            zr = max_nn + rng.uniform(size=shape)
+            z = min_nn - uniform(1, 5, size=shape)
+            zr = max_nn + uniform(1, 5, size=shape)
```
mdhaber, line 419: Maybe we should comment that `5` is arbitrary; and finite number is fine.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -866,6 +897,9 @@ def filtered(self, x, *args, **kwargs):
         x = np.asarray(x)
         dtype = self._dtype
         shape = self._shape
+        discrete = isinstance(self, DiscreteDistribution)
+        keep_low_endpoint = discrete and method_name in {'_cdf1', '_logcdf1',
+                                                         '_ccdf1', '_logccdf1'}
```
mdhaber, line 902: Part of change 5. For continuous distributions, `cdf(0)` is replaced with `0` and `ccdf(0)` is replaced with `1`. We don't want to do this for discrete distributions.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -911,6 +945,13 @@ def filtered(self, x, *args, **kwargs):
             any_endpoint = (mask_endpoint if mask_endpoint.shape == ()
                             else np.any(mask_endpoint))
 
+        # Check for non-integral arguments to PMF method
```
mdhaber, line 948: Part of change 5: "Adjust the `_set_invalid_nan` decorator so PMF at non-integral args is 0".

We might want to floor the argument of other distribution methods so public methods would only call private methods with integer arguments?

scipy/stats/_distribution_infrastructure.py
```diff
@@ -1307,9 +1355,9 @@ def _generate_example(dist_family):
         instantiation = f"{name}()"
         X = dist
 
-    p = 0.32
-    x = round(X.icdf(p), 2)
-    y = round(X.icdf(2 * p), 2)
+    prob = 0.32
```
mdhaber, line 1358: I thought this was interfering with the shape parameter `p`, but in retrospect I don't think this is necessary.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -1344,19 +1392,22 @@ def _generate_example(dist_family):
...
+    {X.pdf(x), X.pmf(x)}
+    
     The cumulative distribution function, its complement, and the logarithm
     of these functions are evaluated similarly.
 
     >>> np.allclose(np.exp(X.logccdf(x)), 1 - X.cdf(x))
     True
+    """
 
+    if issubclass(dist_family, ContinuousDistribution):
```
mdhaber, line 1409: The first example here fails for discrete distributions because the the inverse distribution functions are very sensitive to numerical error of the input (e.g. subtractive cancellation of `1 - X.ccdf(x)`) near the edges.

The second example is two-arg CDF, which is not implemented for discrete distributions right now.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -1372,22 +1423,39 @@ def _generate_example(dist_family):
...
+
     >>> X.skewness(), X.kurtosis()
     {X.skewness(), X.kurtosis()}
+
     >>> np.allclose(X.moment(order=6, kind='standardized'),
     ...             X.moment(order=6, kind='central') / X.variance()**3)
     True
+    """
+
+    if issubclass(dist_family, ContinuousDistribution):
```
mdhaber, line 1447: `DiscreteDistribution` doesn't implement `logentropy` yet.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -1985,12 +2053,18 @@ def _quadrature(self, integrand, limits=None, args=None,
...
         # For now, we ignore the status, but I want to return the error
         # estimate - see question 5 at the top.
-        return res.integral
+        if isinstance(self, ContinuousDistribution):
+            res = _tanhsinh(f, a, b, args=args, log=log, rtol=rtol)
+            return res.integral
+        else:
+            res = np.asarray(nsum(f, a, b, args=args, log=log, tolerances=dict(rtol=rtol)).sum)
+            zero = -np.inf if log else 0
+            res[a > b] = zero  # nsum should probably return 0 instead of nan
```
mdhaber, line 2064: Should we change `nsum`?

scipy/stats/_distribution_infrastructure.py
```diff
@@ -1985,12 +2053,18 @@ def _quadrature(self, integrand, limits=None, args=None,
         args = np.broadcast_arrays(*args)
         # If we know the median or mean, consider breaking up the interval
         rtol = None if _isnull(self.tol) else self.tol
-        res = _tanhsinh(f, a, b, args=args, log=log, rtol=rtol)
         # For now, we ignore the status, but I want to return the error
         # estimate - see question 5 at the top.
-        return res.integral
+        if isinstance(self, ContinuousDistribution):
```
mdhaber, line 2058: Change 1: use `nsum` instead of `tanhsinh` for discrete distributions.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -1999,7 +2073,10 @@ def _solve_bounded(self, f, p, *, bounds=None, params=None):
         p, xmin, xmax = np.broadcast_arrays(p, xmin, xmax)
         if not p.size:
             # might need to figure out result type based on p
-            return np.empty(p.shape, dtype=self._dtype)
+            res = _RichResult()
+            empty = np.empty(p.shape, dtype=self._dtype)
+            res.xl, res.x, res.xr = empty, empty, empty
+            res.fl, res.fr = empty, empty
```
mdhaber, line 2079: Kind of clumsy/annoying, but it allows the methods that call `_solve_bounded` to treat this as any other case. I figure it's better to special-case here rather than everwhere that calls this function.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -2034,8 +2111,10 @@ def f2(x, _p, **kwargs):  # named `_p` to avoid conflict with shape `p`
         res = _bracket_root(f3, xl0=a, xr0=b, xmin=xmin, xmax=xmax, args=args)
         # For now, we ignore the status, but I want to use the bracket width
         # as an error estimate - see question 5 at the top.
+
         xrtol = None if _isnull(self.tol) else self.tol
-        return _chandrupatla(f3, a=res.xl, b=res.xr, args=args, xrtol=xrtol).x
+        xatol = None if xatol is None else xatol
+        return _chandrupatla(f3, a=res.xl, b=res.xr, args=args, xrtol=xrtol, xatol=xatol)
```
mdhaber, line 2117: Change 3: when inverting a discrete distribution function numerically, set `xatol=1` to avoid unnecessary iteration. The integer root is the only integer within the bracket. (Special cases - if the bracket was not valid or if the final bracket happens to have integer endpoints - are handled in the functions that call `_solve_bounded`. The right thing to do depends on whether it's the CDF or CCDF.)

Note that this now returns the whole result object so the caller method can see the result bracket and function values. This is why we had to change the `if not p.size` case above, and in the continuous inverse methods, we now have to extract attribute `x`.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -2182,8 +2261,8 @@ def _logentropy_logexp_safe(self, **params):
 
     def _logentropy_quadrature(self, **params):
         def logintegrand(x, **params):
-            logpdf = self._logpdf_dispatch(x, **params)
-            return logpdf + np.log(0j+logpdf)
+            logpxf = self._logpxf_dispatch(x, **params)
```
mdhaber, line 2264: Part of change 1: referring to "PXF" instead of "PDF"/"PMF" does the right thing for both continuous and discrete distributions. Note that `self._quadrature` below uses `tanhsinh` for continuous and `nsum` for discrete.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -2209,9 +2288,12 @@ def _entropy_logexp(self, **params):
 
     def _entropy_quadrature(self, **params):
         def integrand(x, **params):
-            pdf = self._pdf_dispatch(x, **params)
-            logpdf = self._logpdf_dispatch(x, **params)
-            return logpdf * pdf
+            pxf = self._pxf_dispatch(x, **params)
+            logpxf = self._logpxf_dispatch(x, **params)
+            temp = np.asarray(pxf)
+            i = (pxf != 0)  # 0 * inf -> nan; should be 0
```
mdhaber, line 2294: Surprised this hadn't come up before.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -2249,17 +2331,18 @@ def _mode_dispatch(self, method=None, **params):
...
         a, b = self._support(**params)
         m = self._median_dispatch(**params)
 
-        f, args = _kwargs2args(lambda x, **params: -self._pdf_dispatch(x, **params),
+        f, args = _kwargs2args(lambda x, **params: -self._pxf_dispatch(x, **params),
                                args=(), kwargs=params)
         res_b = _bracket_minimum(f, m, xmin=a, xmax=b, args=args)
-        res = _chandrupatla_minimize(f, res_b.xl, res_b.xm, res_b.xr, args=args)
+        res = _chandrupatla_minimize(f, res_b.xl, res_b.xm, res_b.xr,
+                                     args=args, xatol=xatol)
```
mdhaber, line 2345: Part of Change 4: using "coarse" `xatol`.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -2361,6 +2444,43 @@ def _pdf_formula(self, x, **params):
...
 
+    @_set_invalid_nan
+    def logpmf(self, x, /, *, method=None):
+        return self._logpmf_dispatch(x, method=method, **self._parameters)
+
+    @_dispatch
+    def _logpmf_dispatch(self, x, *, method=None, **params):
+        if self._overrides('_logpmf_formula'):
+            method = self._logpmf_formula
+        elif _isnull(self.tol):  # ensure that developers override _logpmf
```
mdhaber, line 2455: Just mirroring `_pdf_dispatch` logic here. This was always a placeholder for better logic. The concern is that the PMF can underflow before the log is taken, and the hope is that annoying failures here will encourage distributions authors to override `_logpmf_formula`.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3226,7 +3346,7 @@ def logintegrand(x, order, logcenter, **params):
 
     ### Convenience
 
-    def plot(self, x='x', y='pdf', *, t=('cdf', 0.0005, 0.9995), ax=None):
+    def plot(self, x='x', y=None, *, t=None, ax=None):
```
mdhaber, line 3349: Lots of changes here to make pretty plots of discrete distribution PMFs / CDFs. This also improves the handling of distributions with finite support endpoints.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3417,6 +3567,117 @@ def plot(self, x='x', y='pdf', *, t=('cdf', 0.0005, 0.9995), ax=None):
...
+        return self._logpdf_dispatch(x, method=method, **params)
+
+
+class DiscreteDistribution(_UnivariateDistribution):
+    def _overrides(self, method_name):
+        if method_name in {'_logpdf_formula', '_pdf_formula'}:
+            return True
+        return super()._overrides(method_name)
+
+    # These should probably check whether pmf = 0
```
mdhaber, line 3595: Is it worth the time to check whether `np.isneginf(logpmf)`? This is typically not going to be true within the theoretical support. Also, for distributions that only define PMF, underflows would cause PDF to be zero at points that theoretically have nonzero probability density. I figure it's better for developers to override this where necessary.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3417,6 +3567,117 @@ def plot(self, x='x', y='pdf', *, t=('cdf', 0.0005, 0.9995), ax=None):
...
+        return super()._cdf_quadrature(np.floor(x), **params)
+
+    def _logcdf_quadrature(self, x, **params):
+        return super()._logcdf_quadrature(np.floor(x), **params)
+
+    def _ccdf_quadrature(self, x, **params):
+        return super()._ccdf_quadrature(np.floor(x + 1), **params)
+
+    def _logccdf_quadrature(self, x, **params):
+        return super()._logccdf_quadrature(np.floor(x + 1), **params)
```
mdhaber, line 3618: Part of change 2: When using "quadrature" to evaluate a distribution function, replace the argument with `floor(x)` (CDF) or `floor(x+1)` (CCDF).

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3417,6 +3567,117 @@ def plot(self, x='x', y='pdf', *, t=('cdf', 0.0005, 0.9995), ax=None):
...
+        res = self._solve_bounded(self._ccdf_dispatch, x, params=params, xatol=1)
+        res = np.where(res.fl <= 0, np.floor(res.xl), np.floor(res.xr))
+        res[np.isnan(x)] = np.nan
+        return res
+
+    def _ilogccdf_inversion(self, x, **params):
+        res = self._solve_bounded(self._logccdf_dispatch, x, params=params, xatol=1)
+        res = np.where(res.fl <= 0, np.floor(res.xl), np.floor(res.xr))
+        res[np.isnan(x)] = np.nan
+        return res
```
mdhaber, line 3642: Here's the special case handling in change 3. We might be able to eliminate it if we handle the case in which there is no valid bracket.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3417,6 +3567,117 @@ def plot(self, x='x', y='pdf', *, t=('cdf', 0.0005, 0.9995), ax=None):
...
+
+    def _mode_optimization(self, **params):
+        # If `x` is the true mode of a unimodal continuous function, we can find
+        # the mode among the integers by rounding in each direction and checking
+        # which is better. I think `xatol` should be able to be 1 if the documented
+        # convergence criterion were implemented, but the actual convergence
+        # criterion is different, and I'm not sure if it controls the total width
+        # of the bracket. `xatol=0.1` might not be safe as-implemented.
+        x = super()._mode_optimization(xatol=0.1, **params)
```
mdhaber, line 3651: Part of change 4. `xatol=0.1` is arbitrary. Ideally we could set it to `xatol = 1` to guarantee that the minimum has been narrowed down to an interval of unit width, but see gh-22309.

scipy/stats/_probability_distribution.py
```diff
@@ -30,8 +30,8 @@ def support(self):
         Notes
         -----
         Suppose a continuous probability distribution has support ``(l, r)``.
-        The following table summarizes the value returned by methods
-        of ``ContinuousDistribution`` for arguments outside the support.
+        The following table summarizes the value returned by several
+        methods when the argument is outside the support.
```
mdhaber, line 34: Please review doc updates carefully.

scipy/stats/tests/test_continuous.py
```diff
@@ -1031,19 +1031,25 @@ def test_shapes(self):
...
-    @pytest.mark.parametrize('i, distdata', enumerate(distcont))
+    @pytest.mark.parametrize('i, distdata', enumerate(distcont + distdiscrete))
     def test_make_distribution(self, i, distdata):
         distname = distdata[0]
 
         slow = {'argus', 'exponpow', 'exponweib', 'genexpon', 'gompertz', 'halfgennorm',
                 'johnsonsb', 'kappa4', 'ksone', 'kstwo', 'kstwobign', 'powerlognorm',
-                'powernorm', 'recipinvgauss', 'studentized_range', 'vonmises_line'}
+                'powernorm', 'recipinvgauss', 'studentized_range', 'vonmises_line',
+                'betanbinom', 'zipf', 'logser', 'skellam'}  # discrete
```
mdhaber, line 1041: `betabinom`, `zipf`, and `logser` are all slow because for the parameters tested, the median is at the left end of the support, so we run into the last shortcoming of the [top post](https://github.com/scipy/scipy/pull/22312#issue-2782644701).

scipy/stats/tests/test_continuous.py
```diff
@@ -1031,19 +1031,25 @@ def test_shapes(self):
...
             pytest.skip('Skipping as XSLOW')
 
-        if distname in {  # skip these distributions
-            'levy_stable',  # private methods seem to require >= 1d args
-            'vonmises',  # circular distribution; shouldn't work
+        if distname in {              # skip these distributions
+            'levy_stable',            # private methods seem to require >= 1d args
+            'vonmises',               # circular distribution; shouldn't work
+            'poisson_binom',          # vector shape parameter
+            'hypergeom',              # distribution functions need interpolation
```
mdhaber, line 1049:
```python3
from scipy.special import _ufuncs as scu
M, n, N = 20, 7, 12
scu._hypergeom_pmf(3, n, N, M)  # np.float64(0.19865841073271417)
scu._hypergeom_pmf(4, n, N, M)  # np.float64(0.3575851393188855)
scu._hypergeom_pmf(3.5, n, N, M)  # np.float64(nan)
```
It would be helpful for the infinite sums if this would interpolate for non-integral arguments. We can add that, of course, but often this is something that happens naturally in the underlying implementation but is manually disabled. Any chance that's the case and we can enable it?

scipy/stats/tests/test_continuous.py
```diff
@@ -1031,19 +1031,25 @@ def test_shapes(self):
...
+            'zipfian',                # distribution functions need interpolation
```
mdhaber, line 1050: Actually I can fix this one just by adding `k = np.floor(k)` to the CDF. We could also add that to the `_dispatch` method to make sure that no `_cdf_formula` is ever called with non-integral `k`. Not sure if we want to pull the trigger on that.

scipy/stats/tests/test_continuous.py
```diff
@@ -1031,19 +1031,25 @@ def test_shapes(self):
...
+            'nchypergeom_fisher',     # distribution functions don't accept NaN
```
mdhaber, line 1051: I think the arguments have to be *integers*, but we use `nan`s to when parameters are invalid in hope that functions fast-track them.
```python3
import numpy as np
from scipy import stats
M, n, N, odds = 140, 80, 60, 0.5
stats.nchypergeom_fisher._pmf(np.nan, M, n, N, odds)
# ValueError: cannot convert float NaN to integer
```
Can the wrapper just return NaN whenever any argument is a NaN (before converting to integer)?

After that, we'd face the problem that the PMF needs to interpolate... maybe doing that in the wrapper is a tall order?

mdhaber:
Other follow-ups that should be relatively straightforward:
- I'm not sure how useful it would be, but it doesn't look like shifting and scaling (*multiplying*) by integers would be a problem. I tested this briefly by:
  - giving `ContinuousDistribution` and `DiscreteDistribution` an attribute (e.g. `self._type`) that identifies the instance as continuous or discrete.
  - making `TransformedDistribution` a subclass of `UnivariateDistribution`, and making sure that `TransformedDistribution` copies the `_type` attribute in the initializer. 
  - Replacing some `isinstance` checks with checking the `_type` attribute.
  - (detail) Making the `p` argument of `_shift_scale_inverse_function.wrapped` positional-only to avoid name collisions with shape parameters. (This will be needed in some other places, too.)
 
  After that, a shifted/scaled `Binomial` seemed to work just fine.
  For more general shifting/scaling, some of the code assumes that the supported points are integers. Even if we generalize that, the user experience might not be great due to inherent limitations of floating point arithmetic: in general, the PMF would be nonzero and the CDF would jump only at very specific floating point values; this might be problematic if there is round-off error (either when the user calculates the arguments or we unshift/unscale them). I think there is some relevant discussion at https://github.com/scipy/scipy/pull/15932#issuecomment-1086949899 and nearby.

I'll extend this post with thoughts on `truncate` and `Mixture` soon.

I don't think I'll study other transformations in detail right now.
- `abs` and `**` (with integers on the left or right) would probably work. If they would be useful, I'll leave them to others.
- `log` and `exp` are not in the cards any time soon because the supported points would no longer be integral or even arithmetically spaced.
- I don't really want to think about order statistics in the case of discrete distributions, but this paper [discusses it](https://www.math.wm.edu/~leemis/2005informsjoc.pdf).

scipy/stats/_distribution_infrastructure.py
steppi: I had this same thought when looking through this code off GitHub, and came here to comment to find you already mentioned it.

I think the way it should go is that the pdf should be infinite for all points in the support and zero for all points out of the support. Even if the` `pmf` is zero for a point in the support, I think it's safe to act under the assumption that this is just underflow. If the probability at a point is truly zero, I think that point should be taken out of the support. Do you think that make sense?

scipy/stats/tests/test_continuous.py
steppi: The Boost Math implementation branches based on the size of `N`, and the large `N` implementation could be adapted to handle non-integral arguments, though it is currently constrained to unsigned values. 

https://github.com/boostorg/math/blob/a5c062542afb2037054c4e1b1be44db7b0859c38/include/boost/math/distributions/detail/hypergeometric_pdf.hpp#L467-L472

For the smaller `N` case, one coud evaluate the definition in terms of `binom`.

This is completely doable. I guess it's probably best to ask the Boost devs if they'd consider it.

mdhaber: @mborland Some of the discrete distribution PMFs from Boost (e.g. binomial and negative binomial) are defined for non-integral values of the argument `k` whereas the hypergeometric distribution PMF returns NaN. There are some situations in which it is useful to approximate infinite sums involving the PDF with an integral, so if it is natural to define the hypergeometric distribution PMF for non-integral arguments, we'd appreciate it. Would you be willing/able to do that? If not, we can do our own interpolation; it would just be a bit slower. Thanks for considering it!

scipy/stats/_distribution_infrastructure.py
mdhaber: Yup, that's what I had in mind. I'll adjust the comment.

mdhaber: How does this look?
```suggestion
    # We assume the PDF is infinite at all integers within the support.
    # Even if `_pmf_formula` is numerically zero within the support, this could easily
    # be the result of underflow, so we ignore it. If a `_logpmf_formula` is provided,
    # we *could* trust it to be -inf only where the PMF is exactly zero, but we do not.
    # Consequently, developers should override `_pdf_formula`/`_logpmf_formula` to
    # return the correct result (0 / -inf) wherever the PMF is theoretically zero within
    # the support. (Or, if there is sufficient demand, we could provide for the
    # definition of disconnected support.)
```

steppi: Looks good to me.

scipy/stats/tests/test_continuous.py
mborland: I opened an [issue in math](https://github.com/boostorg/math/issues/1240) since this seems like a reasonable ask. I tagged you both so that you know when it's closed out.

steppi: Thank you

scipy/stats/_new_distributions.py
```diff
@@ -367,6 +367,34 @@ def _pdf_formula(self, x, *, a, **kwargs):
...
+
+    _parameterizations = [_Parameterization(_n_param, _p_param)]
+    _variable = _x_param
+
+    def __init__(self, *, n, p, **kwargs):
+        super().__init__(n=n, p=p, **kwargs)
+
+    def _pmf_formula(self, x, *, n, p, **kwargs):
+        return special.binom(n, x) * p**x * (1 - p)**(n - x)
+
```
mdhaber, line 396: (obviously add implementations of the other methods before merge)

mdhaber:
An MRE for one of the hypothesis failures is:
```python3
import numpy as np
from scipy import stats
x = np.asarray([np.nan, np.nan])
y = np.array([ 6., 10.])
X = stats.Binomial(n=np.array(15.), p=np.array([0.65663512, 0.25826382]))
X.logccdf(x, y, method='addition')
# array([-0.03805847, -8.74909714])
```
Also I saw that the return type of `logcdf` is still complex, but it can never by negative now.

mdhaber:
The refguide failure hints at a mistake in merge conflict resolution at some point. Some of the new `_make_distribution_custom` code didn't make it, apparently. We'll have to skim the diff here for accidental regressions. (Presumably there aren't many if tests are passing, though.)

mdhaber:
One failure is still:
```python3
from scipy import stats
X = stats.Binomial(n=18, p=1)
X.moment(3, kind='central', method='normalize')
```
Not raising on all platforms.

The one on the mypy job is:
```python3
import numpy as np
from scipy import stats
X = stats.Binomial(n=np.asarray([10., 16.]), p=np.asarray([0., 0.]))
X.iccdf(0.99156979), X.iccdf(0.99156979, method='inversion') 
# (array([0., 0.]), array([10., 16.]))
```

steppi:
The failure is unrelated; I also see it here https://github.com/scipy/scipy/pull/22939. Today I learned that Sphinx uses the [Snowball Stemmer](https://pypi.org/project/snowballstemmer/) to reduce words to a root form for use in search. CI was unlucky and pulled a broken release of the snowball stemmer before it was yanked, see https://github.com/snowballstem/snowball/issues/229. This has since been resolved upstream.

steppi (approved):
I'm happy with the state this is in. There are some important things to follow-up on. Offhand,

1. Find and stick with a consistent two argument cdf definition for both continuous and discrete distributions. 
2. Get rid of hypothesis, and reorganize the test suite so that tests are determinstic, with failures in one assert not blocking remaining asserts from running, and skip logic can be done at the pytest mark level, not in conditionals spread throughout the code.
3. Decide on what to do about functions currently needing to be responsible for knowing when the output should be nan, and whether we really want to require nan checks strewn throughout various function implementations to make things work. We may want the NaN checks in the input validation instead.
4. A possible improvement @mdhaber had suggested to the mode calculation for discrete distributions, https://github.com/mdhaber/scipy/pull/121#discussion_r2059267607.

Very nice work @mdhaber. Sorry I couldn't get the probability definition for two argument cdf working, it seems it will require some thought. I'm going to regular merge this so that our separate work doesn't get blended together in the git blame.

mdhaber:
A few other items for follow-up:

- investigate/fix test skips (e.g. discrete distributions that don't work with `make_distribution`
- fix slow inversion / edge case issues when bracket search fails; this should fix degenerate `Binomial` distributions
- https://github.com/scipy/scipy/pull/22312#issuecomment-2588114565

PR #22320, opened 2025-01-13
TST: array types: wrap namespaces centrally

crusaderky:
Prevent headaches such as https://github.com/scipy/scipy/pull/22240/files#r1913475509

### DIFF
--- a/scipy/differentiate/tests/test_differentiate.py
+++ b/scipy/differentiate/tests/test_differentiate.py
@@ -581,8 +581,10 @@ class TestJacobian(JacobianHessianTest):
             return xp.sin(2*x) * y**2
 
         res = jacobian(df1, z, initial_step=10)
-        assert xp_size(xp.unique_values(res.nit)) == 4
-        assert xp_size(xp.unique_values(res.nfev)) == 4
+        # FIXME https://github.com/scipy/scipy/pull/22320#discussion_r1914898175
+        if not is_torch(xp):
+            assert xp_size(xp.unique_values(res.nit)) == 4
+            assert xp_size(xp.unique_values(res.nfev)) == 4
 
         res00 = jacobian(lambda x: df1_0xy(x, z[1]), z[0:1], initial_step=10)
         res01 = jacobian(lambda y: df1_0xy(z[0], y), z[1:2], initial_step=10)
--- a/scipy/differentiate/tests/test_differentiate.py
+++ b/scipy/differentiate/tests/test_differentiate.py
@@ -5,7 +5,7 @@ import numpy as np
 
 import scipy._lib._elementwise_iterative_method as eim
 from scipy._lib._array_api_no_0d import xp_assert_close, xp_assert_equal, xp_assert_less
-from scipy._lib._array_api import is_numpy, is_torch, array_namespace
+from scipy._lib._array_api import is_numpy, is_torch
 
 from scipy import stats, optimize, special
 from scipy.differentiate import derivative, jacobian, hessian
@@ -376,8 +376,7 @@ class TestDerivative:
         # Test that integers are not passed to `f`
         # (otherwise this would overflow)
         def f(x):
-            xp_test = array_namespace(x)  # needs `isdtype`
-            assert xp_test.isdtype(x.dtype, 'real floating')
+            assert xp.isdtype(x.dtype, 'real floating')
             return x ** 99 - 1
 
         if not is_torch(xp):  # torch defaults to float32
@@ -659,10 +658,9 @@ class TestHessian(JacobianHessianTest):
 
     def test_nfev(self, xp):
         z = xp.asarray([0.5, 0.25])
-        xp_test = array_namespace(z)
 
         def f1(z):
-            x, y = xp_test.broadcast_arrays(*z)
+            x, y = xp.broadcast_arrays(*z)
             f1.nfev = f1.nfev + (math.prod(x.shape[2:]) if x.ndim > 2 else 1)
             return xp.sin(x) * y ** 3
         f1.nfev = 0
--- a/scipy/differentiate/tests/test_differentiate.py
+++ b/scipy/differentiate/tests/test_differentiate.py
@@ -5,7 +5,7 @@ import numpy as np
 
 import scipy._lib._elementwise_iterative_method as eim
 from scipy._lib._array_api_no_0d import xp_assert_close, xp_assert_equal, xp_assert_less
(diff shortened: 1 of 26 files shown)


mdhaber:
I had assumed the purpose of using un-wrapped namespaces in the tests was for the tests to better reflect the conditions in practice, assuming end-users do not use `array_api_compat`. 

That said, since `array_api_compat` does not wrap array objects themselves (IIRC), I don't know if the SciPy function being tested will behave any differently whether the array arguments were generated using a wrapped or unwrapped namespace. 

From that perspective, providing a wrapped `xp` to the unit tests really shouldn't affect the results , but would make writing the tests more convenient. So if that's right, seems good to me.

crusaderky:
> That said, since `array_api_compat` does not wrap array objects themselves (IIRC)

Correct.

> I don't know if the SciPy function being tested will behave any differently whether the array arguments were generated using a wrapped or unwrapped namespace.

Unless a user-facing function accepts xp as an optional parameter and the test explicitly passes `xp=xp` to it, it can't behave any different. I haven't witnessed this specific condition throughout the scipy codebase.

mdhaber:
> Unless a user-facing function accepts xp as an optional parameter 

I am not aware of any. 
So +1 for the idea.

lucascolley:
I think this works, yes. I suppose if array-api-compat ever did modify the array objects then we would have problems.

> Unless a user-facing function accepts xp as an optional parameter

https://docs.scipy.org/doc/scipy/reference/generated/scipy.fft.fftfreq.html

crusaderky:
> I think this works, yes. I suppose if array-api-compat ever did modify the array objects then we would have problems.
> 
> > Unless a user-facing function accepts xp as an optional parameter
> 
> https://docs.scipy.org/doc/scipy/reference/generated/scipy.fft.fftfreq.html

You also need a test that explicitly passes the unwrapped xp, otherwise it will default internally to `array_namespace(args)`

scipy/ndimage/tests/test_interpolation.py
```diff
@@ -1059,8 +1052,7 @@ def test_shift_with_nonzero_cval(self, order, mode, dtype, xp):
                                [0, 1, 1, 1],
                                [0, 1, 1, 1]], dtype=dtype)
 
-        isdtype = array_namespace(data).isdtype
-        if isdtype(data.dtype, 'complex floating'):
+        if np.isdtype(data.dtype, 'complex floating'):
```
crusaderky, line 1055: not a typo

lucascolley: Does that work for `np<2`?

scipy/stats/tests/test_stats.py
```diff
@@ -8160,12 +8140,12 @@ def test_monotonicity(self, variant, method, xp):
         # column (all), or independently down each column (random).
         if variant == "single":
             pvaluess = xp.broadcast_to(xp.asarray(rng.random(n)), (m, n))
-            pvaluess = xp_test.concat([xp.reshape(xp.linspace(0.1, 0.9, m), (-1, 1)),
+            pvaluess = xp.concat([xp.reshape(xp.linspace(0.1, 0.9, m), (-1, 1)),
                                        pvaluess[:, 1:]], axis=1)
```
crusaderky, line 8144:
```suggestion
                                  pvaluess[:, 1:]], axis=1)
```

scipy/ndimage/tests/test_interpolation.py
crusaderky: it does, because np is array_api_compat.numpy

crusaderky: Doesn't. Thanks for spotting this.

scipy/differentiate/tests/test_differentiate.py
```diff
@@ -582,9 +581,8 @@ def df1_1xy(x, y):
             return xp.sin(2*x) * y**2
 
         res = jacobian(df1, z, initial_step=10)
-        if is_numpy(xp):
-            assert len(np.unique(res.nit)) == 4
-            assert len(np.unique(res.nfev)) == 4
+        assert xp_size(xp.unique_values(res.nit)) == 4
+        assert xp_size(xp.unique_values(res.nfev)) == 4
```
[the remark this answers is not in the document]
mdhaber: Not a bug. I'll fix the test in a bit . You can skip that line if it's holding you up.

scipy/ndimage/tests/test_interpolation.py
```diff
@@ -879,9 +875,8 @@ def test_affine_transform26(self, order, xp):
         tform_original = xp.eye(2)
         offset_original = -xp.ones((2, 1))
 
-        concat = array_namespace(tform_original, offset_original).concat
-        tform_h1 = concat((tform_original, offset_original), axis=1)  # hstack
-        tform_h2 = concat( (tform_h1, xp.asarray([[0.0, 0, 1]])), axis=0)  # vstack
+        tform_h1 = xp.concat((tform_original, offset_original), axis=1)  # hstack
+        tform_h2 = xp.concat( (tform_h1, xp.asarray([[0.0, 0, 1]])), axis=0)  # vstack
```
lucascolley, line 879: nit

```suggestion
        tform_h2 = xp.concat((tform_h1, xp.asarray([[0.0, 0, 1]])), axis=0)  # vstack
```

lucascolley (approved):
thanks @crusaderky, very nice clean-up

scipy/differentiate/tests/test_differentiate.py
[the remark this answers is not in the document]
crusaderky: skipped, ty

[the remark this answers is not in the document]
mdhaber: Looking more carefully, this part of the test was NumPy-only before the PR:
https://github.com/scipy/scipy/blob/df134eab5a500c2146ed4552c8674a78d8154ee9/scipy/differentiate/tests/test_differentiate.py#L589-L591

The purpose is to check that it's *possible* for all four numbers to be different. It was challenging to find an example for which that was the case, and it turned out to be sensitive to arithmetic, so we made it NumPy only.

It must have started failing only because the `if is_numpy` line was removed?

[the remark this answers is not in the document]
crusaderky: > It must have started failing only because the if is_numpy line was removed?

Correct. I find surprising that different backends can yield different results.

PR #22322, opened 2025-01-14
MAINT: stats.Mixture: make return type consistent when `shape == ()`

mdhaber:
#### Reference issue
-

While working on mixtures involving discrete random variables, I noticed that the return type (scalar or array) depended on which method was called. This makes the type self-consistent and consistent with the rest of the new infrastructure.


tupui (approved):

tupui:
Makes sense to me 👍 thanks Matt.

Issue #6052, opened 2016-04-10
savemat does not save keys starting with underscore


daskol:
If one tries to save key with the underscore prefix, mat-file will not contain that key. Mat-file has not ASCII string for the key in hex editor. Try this code
```python
In [1]: savemat('test.mat', {'_key': 0})
In [2]: whosmat('test.mat')
Out[2]: []
```
So the issue is in `savemat` method of `scipy.io` and I probably have solution.

Is there any reason to ignore key with the underscore prefix?

ev-br:
Duplicate of https://github.com/scipy/scipy/issues/5435

PR #22324, opened 2025-01-14
ENH: io: add invalid field name warning for `savemat`

### LINKED ISSUE #5435: savemat silently drops entries starting with "_"
Currently, savemat silently drops entries whose name start with `_` (mio5.py line 832 `if name[0] == '_':`) and sub-dict entries whose name start with `_` or a digit (mio5.py line 439 `field[0] not in '_0123456789')`).  I'd suggest dropping this restriction: these are far from being the only "invalid" strings (really, this should be testing for whether `name` or `field` is a valid MATLAB indentifier, which also includes issues such as non-ASCII characters and so on), their presence doesn't prevent MATLAB from reading the file, and e.g. invalid struct entries can be accessed on MATLAB's side by `x = <invalid-key>; struct.(x)` (à la `getattr`).

If the restriction is kept, at least a warning should be given when an entry is dropped.

### LINKED ISSUE #16203: BUG: scipy.io.savemat discards nested names with a leading digit
### Describe your issue.

Matlab/Octave require field names for structs constructed with dot notation to [begin with a letter](https://www.mathworks.com/help/matlab/ref/struct.html). However, structure arrays created with the `struct` function have no such restriction. For example:

    mystruct = struct('d', struct('1a', {1,2,3}))
    save -binary -v6 octave.mat mystruct

This can be loaded in Python with something like this:

    import scipy.io as sio

    load_dict = sio.loadmat('octave.mat')
    mystruct = load_dict['mystruct']
    d = mystruct['d']
    myvar = d[0][0]
    print(myvar)
    print(myvar['1a'])

However, when attempting to generate the same with `savemat`, the data is silently discarded. For example:

    import scipy.io as sio
    import numpy as np
    import io

    save_dict = {}
    save_dict['d'] = {}
    save_dict['d']['1a'] = {}
    save_dict['d']['1a'] = np.array([1.,2.,3.])

    fp = io.BytesIO()

    sio.savemat(fp, save_dict, format='5')

    load_dict = sio.loadmat(fp)
    print(load_dict['d'])
    print(load_dict['d']['1a'])

Simply replacing `'1a'` with `'a1'` results in a functional script:

    import scipy.io as sio
    import numpy as np
    import io

    save_dict = {}
    save_dict['a1'] = {}
    save_dict['a1'] = np.array([1.,2.,3.])

    fp = io.BytesIO()

    sio.savemat(fp, save_dict, format='5')

    load_dict = sio.loadmat(fp)
    print(load_dict['a1'])

To make matters more confusing, this only applies to nested names, not top-level ones, i.e. this will run without issue:

    import scipy.io as sio
    import numpy as np
    import io

    save_dict = {}
    save_dict['1a'] = {}
    save_dict['1a'] = np.array([1.,2.,3.])

    fp = io.BytesIO()

    sio.savemat(fp, save_dict, format='5')

    load_dict = sio.loadmat(fp)
    print(load_dict['1a'])

Edit: fixed the key here, accidentally wrote it as `'a1'` instead of `'1a'`.

The reason I noticed this was because of attempting to use it for data that was keyed by date strings like e.g. `example["id"]["2022-02-17_my_name"]`.

### Reproducing Code Example

```python
#! /usr/bin/env python3
import scipy.io as sio
import numpy as np
import io

save_dict = {}
save_dict['d'] = {}
save_dict['d']['1a'] = {}
save_dict['d']['1a'] = np.array([1.,2.,3.])

fp = io.BytesIO()

sio.savemat(fp, save_dict, format='5')

load_dict = sio.loadmat(fp)
load_dict['d']['1a']
```

### Error message

```shell
  File "./reproduce.py", line 16, in <module>
    load_dict['d']['1a']
IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices
```

1.4.1 1.18.1 sys.version_info(major=3, minor=6, micro=9, releaselevel='final', serial=0)

AtsushiSakai:
#### Reference issue

Fix #5435, #6052, #16203

As repeatedly reported in the above issues, the current [savemat](https://docs.scipy.org/doc/scipy/reference/generated/scipy.io.savemat.html) function ignores elements 
1. where the top-level keys start with an underscore 
or
2. where sub-dictionary keys start with an underscore or a number when saving to a .mat file. 

However, since these elements are ignored without any logs, many users seem to be confused.

To address this, as suggested in #5435, I have made it so that a warning log is output when such invalid field names are ignored. 
Additionally, while there was a [MatReadWarning](https://docs.scipy.org/doc/scipy/reference/generated/scipy.io.matlab.MatReadWarning.html#scipy.io.matlab.MatReadWarning), there was no `MatWriteWarning`, so I have newly defined it. 
And I updated the savemat function’s docstring to explain this behavior.
I have also made modifications to existing unit tests where warnings are issued and added new unit tests.

### DIFF
--- a/scipy/io/matlab/_mio.py
+++ b/scipy/io/matlab/_mio.py
@@ -267,7 +267,10 @@ def savemat(file_name, mdict,
         True``).
         Can also pass open file_like object.
     mdict : dict
-        Dictionary from which to save matfile variables.
+        Dictionary from which to save matfile variables. Note that this dict
+        has a key starting with `_` or sub dict has a key starting with `_`
+        or a digit, these key's items will not be saved in the mat file and
+        `MatWriteWarning` will be issued.
     appendmat : bool, optional
         True (the default) to append the .mat extension to the end of the
         given filename, if not already present.
--- a/scipy/io/matlab/_mio.py
+++ b/scipy/io/matlab/_mio.py
@@ -267,8 +267,8 @@ def savemat(file_name, mdict,
         True``).
         Can also pass open file_like object.
     mdict : dict
-        Dictionary from which to save matfile variables. Note that this dict
-        has a key starting with `_` or sub dict has a key starting with `_`
+        Dictionary from which to save matfile variables. Note that if this dict
+        has a key starting with ``_`` or a sub-dict has a key starting with ``_``
         or a digit, these key's items will not be saved in the mat file and
         `MatWriteWarning` will be issued.
     appendmat : bool, optional
--- a/scipy/io/matlab/tests/test_mio.py
+++ b/scipy/io/matlab/tests/test_mio.py
@@ -715,12 +715,12 @@ def test_to_writeable():
     assert_any_equal(to_writeable({'a':1,'b':2}), alternatives)
     # Fields with underscores discarded with a warning message.
     with pytest.warns(MatWriteWarning, match='Starting field name with'):
-        assert_any_equal(to_writeable({'a':1,'b':2, '_c':3}), alternatives)
+        assert_any_equal(to_writeable({'a':1, 'b':2, '_c':3}), alternatives)
     # Not-string fields discarded
     assert_any_equal(to_writeable({'a':1,'b':2, 100:3}), alternatives)
     # String fields that are valid Python identifiers discarded
     with pytest.warns(MatWriteWarning, match='Starting field name with'):
-        assert_any_equal(to_writeable({'a':1,'b':2, '99':3}), alternatives)
+        assert_any_equal(to_writeable({'a':1, 'b':2, '99':3}), alternatives)
     # Object with field names is equivalent
 
     class klass:
@@ -1374,8 +1374,7 @@ def test_invalid_field_name_warning():
         ('mystr', mlarr('a string')))
     check_mat_write_warning(names_vars)
 
-    names_vars = (
-        ('mymap', {"a": 1, "_b": 2}),)
+    names_vars = (('mymap', {"a": 1, "_b": 2}),)
     check_mat_write_warning(names_vars)
 
     names_vars = (('mymap', {"a": 1, "1a": 2}),)
--- a/scipy/io/matlab/tests/test_mio.py
+++ b/scipy/io/matlab/tests/test_mio.py
@@ -22,8 +22,8 @@ from scipy.sparse import issparse, eye_array, coo_array, csc_array
 import scipy.io
 from scipy.io.matlab import MatlabOpaque, MatlabFunction, MatlabObject
 import scipy.io.matlab._byteordercodes as boc
-from scipy.io.matlab._miobase import (
-    matdims, MatWriteError, MatReadError, matfile_version)
+from scipy.io.matlab._miobase import (matdims, MatWriteError, MatReadError,
+                                      matfile_version, MatWriteWarning)
 from scipy.io.matlab._mio import mat_reader_factory, loadmat, savemat, whosmat
 from scipy.io.matlab._mio5 import (
     MatFile5Writer, MatFile5Reader, varmats_from_mat, to_writeable,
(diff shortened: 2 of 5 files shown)


matthew-brett:
You caught me!  Yes, I will take a look later today.

j-bowhay:
Thanks for the investigation. I think my preference would be to side on simplicity and just preserve the current behaviour. Then if an octave user requests what you have proposed we can always make the change

scipy/io/matlab/tests/test_mio.py
```diff
@@ -1358,3 +1366,28 @@ def test_large_m4():
...
+
+
+def test_invalid_field_name_warning():
+    names_vars = (
+        ('_1', mlarr(np.arange(10))),
+        ('mystr', mlarr('a string')))
+    check_mat_write_warning(names_vars)
+
+    names_vars = (
+        ('mymap', {"a": 1, "_b": 2}),)
```
j-bowhay, line 1378:
```suggestion
    names_vars = (('mymap', {"a": 1, "_b": 2}),)
```

scipy/io/matlab/tests/test_mio.py
```diff
@@ -713,12 +713,14 @@ def test_to_writeable():
     expected2 = np.array([(2, 1)], dtype=[('b', '|O8'), ('a', '|O8')])
     alternatives = (expected1, expected2)
     assert_any_equal(to_writeable({'a':1,'b':2}), alternatives)
-    # Fields with underscores discarded
-    assert_any_equal(to_writeable({'a':1,'b':2, '_c':3}), alternatives)
+    # Fields with underscores discarded with a warning message.
+    with pytest.warns(MatWriteWarning, match='Starting field name with'):
+        assert_any_equal(to_writeable({'a':1,'b':2, '_c':3}), alternatives)
```
j-bowhay, line 718:
```suggestion
        assert_any_equal(to_writeable({'a':1, 'b':2, '_c':3}), alternatives)
```

scipy/io/matlab/tests/test_mio.py
```diff
@@ -713,12 +713,14 @@ def test_to_writeable():
...
     # Not-string fields discarded
     assert_any_equal(to_writeable({'a':1,'b':2, 100:3}), alternatives)
     # String fields that are valid Python identifiers discarded
-    assert_any_equal(to_writeable({'a':1,'b':2, '99':3}), alternatives)
+    with pytest.warns(MatWriteWarning, match='Starting field name with'):
+        assert_any_equal(to_writeable({'a':1,'b':2, '99':3}), alternatives)
```
j-bowhay, line 723:
```suggestion
        assert_any_equal(to_writeable({'a':1, 'b':2, '99':3}), alternatives)
```

scipy/io/matlab/_mio.py
```diff
@@ -267,7 +267,10 @@ def savemat(file_name, mdict,
         True``).
         Can also pass open file_like object.
     mdict : dict
-        Dictionary from which to save matfile variables.
+        Dictionary from which to save matfile variables. Note that this dict
+        has a key starting with `_` or sub dict has a key starting with `_`
```
j-bowhay, line 271:
```suggestion
        has a key starting with ``_`` or a sub-dict has a key starting with ``_``
```

j-bowhay (approved):
Thanks both!

AtsushiSakai:
@matthew-brett  Thank you for your comment. I habe the same opinion as @j-bowhay, I think it’s best to keep the current behavior as it is and only display a warning for backward compatibility.

@j-bowhay Thank you for reviewing and merging!! Sorry, I didn't have time to check GitHub in the past few days...

matthew-brett:
Exploring a little more - Matlab will `load` Matfiles with structs having underscores and digits as first character of their names, and silently skips variables beginning with digits or underscores.  However, the`matfile` function errors out when opening a file with a variable name prefixed by a digit or an underscore.   It allows struct names beginning with digits or underscores.

Comparing Matlab code searches on Github, "matfile(" comes up 6.1k times, compared to "load(" 461k times.   So the use of `matfile` is pretty niche.

The behavior allowing digits at the start of variable names at the top level is odd though (I bet I made it sometime).

You can read struct fields with digits or underscores with `getfield`, but you cannot create them or change values with `setfield`.

This makes me think it's reasonable to continue the current behavior of dropping the struct fields (sub-dict keys) beginning with digits and underscores.   It also seems generally sensible to drop fields prefixed by underscores given the Python 'private' convention for underscores and attributes.

However, it does not seem sensible to continue saving variables with digits as first characters, as these break `matfile`, and Matlab discards them in any case.   So I do think it's worth changing this behavior by default, at some point, to a mode that drops variables with digits as first character.

I think the best way to do this is plan to make this mode the default mode, and:

* Make a new `write_compat` flag to `savemat`, defaulting to NoValue.
* Warn about pending change of default from NoValue to True in due course.  Recommend setting explicit value to True or False for future compatibility.
* With `write_compat=True` we drop variables beginning with digit.
* With `write_compat=NoValue we warn that future `write_compat=True` default will result in variables beginning with digits being dropped in the future, but we do save them.
* `write_compat=False gives the current behavior.

Thoughts?

j-bowhay:
No objections:)

AtsushiSakai:
I agree.

PR #22337, opened 2025-01-16
MAINT: stats.Mixture: fix inverse functions when mean is undefined

mdhaber:
#### Reference issue
https://github.com/scipy/scipy/issues/22334#issuecomment-2594224949

The inverse distribution functions of `Mixture` started bracket search with `self.mean()`, but this didn't work for distributions with undefined mean. This fixes the problem by factoring the bracket search initialization out of `ContinuousDistribution._solve_bounded` and using it in `Mixture._invert`.

#### Additional information
We still might want to special-case `p=0` and `p=1`, but I'd like to see what we do with related special cases in gh-22312 before pulling the trigger on that.

### DIFF
--- a/scipy/stats/_distribution_infrastructure.py
+++ b/scipy/stats/_distribution_infrastructure.py
@@ -1213,7 +1213,7 @@ def _logexpxmexpy(x, y):
 
 
 def _guess_bracket(xmin, xmax):
-    a = -np.ones_like(xmin)
+    a = np.full_like(xmin, -1.0)
     b = np.ones_like(xmax)
 
     i = np.isfinite(xmin) & np.isfinite(xmax)
--- a/scipy/stats/_distribution_infrastructure.py
+++ b/scipy/stats/_distribution_infrastructure.py
@@ -1212,6 +1212,25 @@ def _logexpxmexpy(x, y):
     return res
 
 
+def _guess_bracket(xmin, xmax):
+    a = np.asarray(-np.ones_like(xmin))
+    b = np.asarray(np.ones_like(xmax))
+
+    i = np.isfinite(xmin) & np.isfinite(xmax)
+    a[i] = xmin[i]
+    b[i] = xmax[i]
+
+    i = np.isfinite(xmin) & ~np.isfinite(xmax)
+    a[i] = xmin[i]
+    b[i] = xmin[i] + 1
+
+    i = np.isfinite(xmax) & ~np.isfinite(xmin)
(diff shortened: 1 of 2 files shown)


dschmitz89 (approved):

scipy/stats/_distribution_infrastructure.py
```diff
@@ -1212,6 +1212,25 @@ def _logexpxmexpy(x, y):
     return res
 
 
+def _guess_bracket(xmin, xmax):
+    a = np.asarray(-np.ones_like(xmin))
```
j-bowhay, line 1216: Can `np.ones_like` not be an array?

mdhaber: Huh. Something happened during testing to make me add that. But now when I test it, it does seem that `ones_like` always returns an array. Let's find out.

mdhaber, line 1216:
```suggestion
    a = -np.ones_like(xmin)
```

scipy/stats/_distribution_infrastructure.py
```diff
@@ -1212,6 +1212,25 @@ def _logexpxmexpy(x, y):
...
+    b = np.asarray(np.ones_like(xmax))
```
mdhaber, line 1217:
```suggestion
    b = np.ones_like(xmax)
```

scipy/stats/_distribution_infrastructure.py
mdhaber: Well I guess that answers our immediate question, but it does leave another.

mdhaber: Oh, I remember. It's the negation. Because, you know, obviously a 0-d array shouldn't survive a negation.
LMK if you'd prefer it like it was or if you want this to use `np.full_like` with argument `-1`.

j-bowhay: `full_like` sounds good

scipy/stats/_distribution_infrastructure.py
```diff
@@ -1212,6 +1212,25 @@ def _logexpxmexpy(x, y):
     return res
 
 
+def _guess_bracket(xmin, xmax):
+    a = -np.ones_like(xmin)
```
mdhaber, line 1216:
```suggestion
    a = np.full_like(xmin, -1.0)
```

j-bowhay (approved):

PR #22080, opened 2024-12-14
BUG: special: Fix unchecked memory allocations in `specfun.h`

WarrenWeckesser:
A fix for https://github.com/scipy/scipy/issues/21417.

Summary of the changes:

* Add another SF error code, "SF_ERROR_NOMEM", that is used by ufuncs that require internally allocated memory.  Just like the other possible errors, how this error is handled can be queried and set using `geterr()` and `seterr()`.  Unlike the other error conditions, the default behavior is to raise an exception if a memory allocation fails.

* Modify the C++ functions in `specfun.h` that allocate heap memory to return a status code that is used to indicate when the allocation fails.  Modify all callers of such functions to handle the possible error status.

* Fix names passed as the first argument of `set_error()`.  This argument should be the name of the public ufunc, as it will be shown to the user as part of the error message.

Ping @izaid, @ilayn, @steppi.

PR #22339, opened 2025-01-16
BUG: special: Fix unchecked malloc in stirling2.h

### LINKED ISSUE #22336: BUG: special: Unchecked malloc in stirling2.h
`malloc()` is called here, but the possible NULL return value is not handled:

https://github.com/scipy/scipy/blob/171d671bbc4bf3ad30bf349f8121e6b5090dce33/scipy/special/stirling2.h#L43

If `malloc()` fails, dereferencing `curr` will crash Python.

j-bowhay:
#### Reference issue

Closes #22336

Replaces use of `malloc` with `std::vector`.


WarrenWeckesser:
Thanks, @j-bowhay, but this doesn't fix the problem.  `std::vector` will raise `std::bad_alloc` is there is not sufficient memory for the vector.

j-bowhay:
Ah ok I wasn't quite sure about this, should I catch this and throw a `PyErr_NoMemory()` error?

j-bowhay:
Updated, hopefully this is more along the right lines @WarrenWeckesser. I guess this should use `SF_ERROR_NOMEM` when #22080 is merged

j-bowhay:
Thanks done

scipy/special/stirling2.h
```diff
@@ -40,7 +43,11 @@ double _stirling2_dp(double n, double k){
         return 0.;
     }
     int arraySize = k <= n - k + 1 ? k : n - k + 1;
-    double *curr = (double *) malloc(arraySize * sizeof(double));
+    auto curr = std::unique_ptr<double[]>{new (std::nothrow) double[arraySize]()};
```
WarrenWeckesser, line 46:
```suggestion
    auto curr = std::unique_ptr<double[]>{new (std::nothrow) double[arraySize]};
```
Those parentheses cause the array to be initialized with zeros.  This is not needed, since `curr` will be initialized to 1 below.

scipy/special/stirling2.h
```diff
@@ -40,7 +43,11 @@ double _stirling2_dp(double n, double k){
...
+    if (curr == nullptr){
```
WarrenWeckesser, line 47:
```suggestion
    if (curr == nullptr) {
```
Whitespace nit.

scipy/special/stirling2.h
```diff
@@ -49,7 +56,6 @@ double _stirling2_dp(double n, double k){
             for (int j = 1; j < k; j++){
                 curr[j] = (j + 1) * curr[j] + curr[j - 1];
                 if (isinf(curr[j])){
-                    free(curr);
```
WarrenWeckesser, line 52: Might as well fix this while touching this file: here and below, on overflow, call `sf_error` with the appropriate arguments:

```
                    sf_error("stirling2", SF_ERROR_OVERFLOW, NULL);
```

WarrenWeckesser (changes requested):
I suggested a few tweaks to the code.

scipy/special/stirling2.h
j-bowhay: I considered it this but it seemed inconsistent with the rest of the file?

WarrenWeckesser: Actually, the file is not completely consistent. I can't imagine any C/C++ coding standard that would not put a space here. And barring the adoption of some other style guide, my fallback is [PEP 7](https://peps.python.org/pep-0007/), which at least has an example of an `if` statement that has a space before the opening bracket of the code block. (See also [Google's style guide](https://google.github.io/styleguide/cppguide.html#Formatting_Looping_Branching), which includes the space.)  So let's use the more common convention in new code here.

And if you want, go ahead and fix the other missing spaces (but that's not a blocker).

steppi (approved):
This looks good to me. Thanks @j-bowhay. Thanks for the good suggestions @WarrenWeckesser. I'll wait for you  to approve or dismiss the requested changes before merging or I guess you can just merge it.

WarrenWeckesser (approved):

scipy/special/stirling2.h
WarrenWeckesser: It turns out that this condition can't happen, because currently `_stirling2_dp(n, k)` is called only if `n <= 50`, and for values of `n` that small, there is no overflow.  The change was still worthwhile, since how the code is called could change in the future.

PR #22342, opened 2025-01-16
ENH: cluster: `linkage` support for jax.jit and dask

crusaderky:
Apply a wealth of new paradigms to `scipy.cluster.linkage` and all its trivial aliases (`single`, `complete`, `average`, `weighted`, `centroid`, `median`, and `ward`).
It makes the above functions compatible with `jax.jit` as well as with dask (with the big caveat that the whole computation must fit on a single dask worker).

This PR also makes `scipy.cluster.leaders` compatible with Dask, but not with `jax.jit`, as the function has data-dependent output shape.

This PR wants to be a look & feel showcase of the paradigm that should eventually be applied widely to scipy.

- This PR is blocked by https://github.com/data-apis/array-api-extra/pull/86.

## New functions
- [`array_api_compat.is_lazy_array(x)`](https://data-apis.org/array-api-compat/helper-functions.html#array_api_compat.is_lazy_array)
Return True for Dask and JAX; False otherwise.
- [`xpx.lazy_apply(func, *args: Array, shape, dtype) -> Array`](https://github.com/data-apis/array-api-extra/pull/86)
Convert all inputs to numpy, apply the function, and re-wrap the output(s) with the same namespace as the inputs. On jax and dask, delay the execution of the applied function. 
**Caveats:**
  - on `jax.jit` and Dask, you need to know the output shape in advance.
  - on `jax.jit`, if the wrapped function raises it leads to undefined behaviour.
  - on Dask, the whole inputs are squashed into a single chunk and need to fit in memory on a single worker.


scipy/cluster/hierarchy.py
```diff
@@ -1015,35 +1017,53 @@ def linkage(y, method='single', metric='euclidean', optimal_ordering=False):
...
         distance.is_valid_y(y, throw=True, name='y')
     elif y.ndim == 2:
-        if (y.shape[0] == y.shape[1] and np.allclose(np.diag(y), 0) and
-                xp.all(y >= 0) and np.allclose(y, y.T)):
+        if (
+            y.shape[0] == y.shape[1]
+            and not is_lazy_array(y) 
+            and xp.allclose(xp.diag(y), 0) 
+            and xp.all(y >= 0)
+            and xp.allclose(y, y.T)
```
crusaderky, line 1025: as the outcome of this is just a warning, we simply skip it on jax and dask.

scipy/spatial/distance.py
```diff
@@ -2295,14 +2295,28 @@ def pdist(X, metric='euclidean', *, out=None, **kwargs):
...
+    if X.ndim != 2:
         raise ValueError('A 2-dimensional array must be passed.')
 
-    m, n = s
+    m = X.shape[0]
+    return xpx.apply_lazy(
+        _np_pdist, 
+        X,
+        out,
+        shape=((m * (m - 1)) // 2, ),
```
crusaderky, line 2307: Hoping this is always correct. Tests *should* fail once I apply `lazy_jax_jit` to them if it isn't.

crusaderky: One thing that frustrated me out of this exercise is scope creep.
The exercise started with the intention to exclusively work on `cluster.linkage` and blindly wrap with `xpx.lazy_apply` any non-array-API-aware function. Which is what the current approach to the migration to array api compatibility has been with the pattern `y = xp.asarray(numpy_only_func(np.sarray(x))`.

However, with `lazy_apply` there's a major issue arising, which is that the output shape and dtype can be non-obvious and not clearly documented in the docstring. Here I had to read deep into the implementation of `pdist` to figure it out, and I'm still not confident about it. Moving the knowledge of the output shape higher in the call stack, to `scipy.cluster`, would be very unwise, so I found that the lesser evil was to at least partially migrate this function.

scipy/cluster/hierarchy.py
```diff
@@ -1513,31 +1532,39 @@ def optimal_leaf_ordering(Z, y, metric='euclidean'):
...
+    )
 
-    Z = np.asarray(Z)
-    y = np.asarray(y)
-    return xp.asarray(_optimal_leaf_ordering.optimal_leaf_ordering(Z, y))
+    return xpx.lazy_apply(
+        _optimal_leaf_ordering.optimal_leaf_ordering,
+        Z, 
+        y,
+        shape=Z.shape,
```
lucascolley, line 1564: very minor point, but I wonder whether this pattern is more readable in general (when there is more than one argument to the callable), to separate the callable's params from those of `xpx.lazy_apply`
```suggestion
    return xpx.lazy_apply(
        _optimal_leaf_ordering.optimal_leaf_ordering,
        *(Z, y),
        shape=Z.shape,
```

rgommers (commented):
Thanks @crusaderky, and sorry it took a while to wrap my head around all this. I'd say that overall I'm positive about this PR, and moderately positive about the changes needed for `jax.jit` support.  The changes to `spatial/distance.py` are very nice and clean - that shows that `lazy_apply` with the callback can work as intended.

The main issue seems to be related to error and warning handling, that seems a bit tedious to deal with and end up with code that is a bit harder to understand. `scipy.cluster` has more of this than most other submodules, so I hope that overall it won't be too bad. In the end, this is a usability issue that lazy implementations have. The exceptions aren't used for control flow though, they're invariably to deal with input data that is ill-conditioned in some way. So probably it's reasonable to tell users that they'll have an easier time prototyping with NumPy before migrating to JAX et al. The same applies for direct usage of `jax.numpy` compared to `numpy`, I don't think this is specific to the array API standard or how we implement support for array types in SciPy.

* * *

One meta comment: with PRs like these it does seem quite important to keep the diff as small as possible. This makes things easier to understand, and also for future readers and maintainers who do not follow along closely, it will look less intimidating. The occasional 1-2 line tweak to improve things that one may want to fold in is okay, but from this comment it follows that one should avoid or factor out larger refactorings, adding static typing, and structural reformatting of code.

scipy/cluster/hierarchy.py
```diff
@@ -1015,35 +1018,50 @@ def linkage(y, method='single', metric='euclidean', optimal_ordering=False):
...
         distance.is_valid_y(y, throw=True, name='y')
     elif y.ndim == 2:
-        if (y.shape[0] == y.shape[1] and np.allclose(np.diag(y), 0) and
-                xp.all(y >= 0) and np.allclose(y, y.T)):
-            warnings.warn('The symmetric non-negative hollow observation '
-                          'matrix looks suspiciously like an uncondensed '
-                          'distance matrix',
-                          ClusterWarning, stacklevel=2)
+        if y.shape[0] == y.shape[1]:
+            zero = xp.zeros((), dtype=y.dtype, device=xp_device(y))
```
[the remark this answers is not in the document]
crusaderky: `0` fails on 
- pytorch 2.5.1,
- array-api-strict 2.2, and
- array-api-strict git tip as of today.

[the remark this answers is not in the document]
rgommers: Ah yes, the changes are still being rolled out - thanks for checking. I think once updates to `-strict` and `-compat` are made, using `0` can work.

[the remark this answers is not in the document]
crusaderky: Added a FIXME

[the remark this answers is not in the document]
ev-br: What is the `-strict` failure here?

scipy/cluster/hierarchy.py
```diff
@@ -2388,7 +2442,7 @@ def correspond(Z, Y):
     True
 
     """
-    is_valid_linkage(Z, throw=True)
+    is_valid_linkage(Z, throw=True)  # Don't validate lazy arrays
```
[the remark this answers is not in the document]
crusaderky: For lazy arrays you need to deal with the return value of `is_valid_linkage(throw=True)`, otherwise there is no validation whatsoever. I've updated the comment to clarify.

scipy/cluster/hierarchy.py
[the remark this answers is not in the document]
crusaderky: @ev-br on git tip, `AttributeError: 'int' object has no attribute 'dtype'`

scipy/cluster/hierarchy.py
```diff
@@ -1015,35 +1018,50 @@ def linkage(y, method='single', metric='euclidean', optimal_ordering=False):
...
+        else:
+            return _hierarchy.fast_linkage(y, n, method_code)
+
+    result = xpx.lazy_apply(
+        cy_linkage,
+        y,
+        shape=(n - 1, 4),
+        dtype=xp.float64,
+        as_numpy=True,
+    )
```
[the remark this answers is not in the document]
crusaderky: I've compacted all changes.

scipy/cluster/hierarchy.py
```diff
@@ -2240,64 +2295,63 @@ def is_valid_linkage(Z, warning=False, throw=False, name=None):
...
-
-
-def _check_hierarchy_not_all_clusters_used(Z):
-    n = Z.shape[0] + 1
-    chosen = set()
-    for i in range(0, n - 1):
-        chosen.add(int(Z[i, 0]))
-        chosen.add(int(Z[i, 1]))
-    must_chosen = set(range(0, 2 * n - 2))
-    return len(must_chosen.difference(chosen)) > 0
```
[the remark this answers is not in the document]
crusaderky: https://github.com/scipy/scipy/pull/22430

scipy/cluster/hierarchy.py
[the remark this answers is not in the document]
crusaderky: > Ah yes, the changes are still being rolled out - thanks for checking. I think once updates to `-strict` and `-compat` are made, using `0` can work.

torch is not wrapped by compat though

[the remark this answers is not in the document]
crusaderky: I was mistaken. For array-api-strict, this is a array-api-extra bug.

crusaderky:
Following discussion on JAX being incapable of dealing with errors inside `jax.pure_callback`, I've heavily reworked this PR. It is now ready for a second round of review.

CC @ev-br; could you please ping other relevant stakeholders parties?

- When a final user explicitly calls `is_valid_linkage` or `is_valid_im` with either `throw=True` or `warning=True`, they still materialise the graph in JAX and Dask. This means that they can't be invoked inside `jax.jit`. When in Dask, they have been improved to materialise the graph only once instead of once per check.
- When the same functions are invoked with neither `throw=True` nor `warning=True`, they now return a 0-dimensional Array with bool dtype in JAX and Dask, thus avoiding graph materialisation, and a python bool on all other backends.
- When the same functions are invoked with `throw=True` as an internal health check by many other `scipy.cluster` functions, they skip all checks on the array contents. When the function later calls a Cython core, content checks are then executed inside it, thus avoiding an early graph materialization. If these tests fail in JAX, it will lead to undefined behaviour, but at least doesn't carry the risk of triggering an infinite loop within the Cython code (which, when run on hardware that's billed by the hour, could prove extremely expensive).

scipy/_lib/array_api_extra
crusaderky, line 1: DNM. Points to https://github.com/data-apis/array-api-extra/pull/86

scipy/cluster/hierarchy.py
```diff
@@ -1016,35 +1019,42 @@ def linkage(y, method='single', metric='euclidean', optimal_ordering=False):
...
-        result = _hierarchy.mst_single_linkage(y, n)
-    elif method in ['complete', 'average', 'weighted', 'ward']:
-        result = _hierarchy.nn_chain(y, n, method_code)
-    else:
-        result = _hierarchy.fast_linkage(y, n, method_code)
-    result = xp.asarray(result)
+    def cy_linkage(y, validate):
+        if validate and not np.all(np.isfinite(y)):
+            raise ValueError("The condensed distance matrix must contain only "
+                            "finite values.")
```
crusaderky, line 1045: It's not very useful to run the check that would result in a warning here. In Dask, it would appear on the Dask worker log, very hard to spot.

scipy/cluster/tests/test_hierarchy.py
```diff
@@ -69,24 +70,61 @@
...
-
-
+xfail_xp_backends = pytest.mark.xfail_xp_backends
+use_linkage = skip_xp_backends(cpu_only=True, reason="linkage() invokes Cython code")
+
+lazy_xp_function(single)
+lazy_xp_function(ward)
+lazy_xp_function(linkage, static_argnames=('method', 'metric', 'optimal_ordering'))
+lazy_xp_function(cut_tree, jax_jit=False, allow_dask_compute=999, 
+                 static_argnames=('n_clusters', 'height'))
```
crusaderky, line 80: This and the rest are all TODOs for later PRs.

scipy/cluster/tests/test_hierarchy.py
```diff
@@ -436,32 +468,34 @@ def check_is_valid_linkage_various_size(self, nrow, ncol, valid, xp):
         Z = xp.asarray([[0, 1, 3.0, 2, 5],
                         [3, 2, 4.0, 3, 3]], dtype=xp.float64)
         Z = Z[:nrow, :ncol]
-        assert_(is_valid_linkage(Z) == valid)
+        xp_assert_equal(is_valid_linkage(Z), valid, check_namespace=False)
         if not valid:
             assert_raises(ValueError, is_valid_linkage, Z, throw=True)
```
crusaderky, line 473: Note that this works in jax.jit because it's just a shape test.

scipy/cluster/tests/test_hierarchy.py
```diff
@@ -595,23 +649,32 @@ def test_num_obs_linkage_empty(self, xp):
...
         # 4 and 15 (step size 3).
         for i in range(4, 15, 3):
             y = np.random.rand(i*(i-1)//2)
             y = xp.asarray(y)
             Z = linkage(y)
-            assert_equal(num_obs_linkage(Z), i)
+            xp_assert_equal(num_obs_linkage(Z), i, check_namespace=False)
+
+    @use_linkage
+    def test_num_obs_linkage_multi_matrix(self, xp):
```
crusaderky, line 671: moved from below

scipy/spatial/distance.py
```diff
@@ -2648,34 +2655,25 @@ def is_valid_y(y, warning=False, throw=False, name=None):
     False
 
     """
-    y = np.asarray(y, order='c')
-    valid = True
+    y = _asarray(y)
```
crusaderky, line 2658: This is a regression as y can legitimately be a list of arbitrary python objects, which causes `_asarray` to raise.
Working on a fix.

scipy/spatial/tests/test_distance.py
```diff
@@ -63,6 +63,7 @@
                                     russellrao, seuclidean, sokalmichener,  # noqa: F401
                                     sokalsneath, sqeuclidean, yule)
 from scipy._lib._util import np_long, np_ulong
+from scipy.conftest import skip_xp_invalid_arg
```
crusaderky, line 66: Note: while `pdist` is now array API compatible, adding unit tests for it would be vastly enlarge the scope of this PR, so I would rather leave it to a later PR.

scipy/cluster/tests/test_hierarchy.py
```diff
@@ -423,45 +456,43 @@ def help_is_isomorphic_randperm(self, nobs, nclusters, noniso=False, nerrors=0,
...
+        xp_assert_equal(is_valid_linkage(Z), valid, check_namespace=False)
         if not valid:
             assert_raises(ValueError, is_valid_linkage, Z, throw=True)
 
     def test_is_valid_linkage_int_type(self, xp):
         # Tests is_valid_linkage(Z) with integer type.
         Z = xp.asarray([[0, 1, 3.0, 2],
                         [3, 2, 4.0, 3]], dtype=xp.int64)
-        assert_(is_valid_linkage(Z) is False)
+        xp_assert_equal(is_valid_linkage(Z), False, check_namespace=False)
```
crusaderky, line 476: This is an artefact of `lazy_xp_function`: `is_valid_linkage` always returns (python bool) False, but `jax.jit(is_valid_linkage)` returns `jnp.Array(False)`. This is not the case when jax.jit is applied higher in the call stack, as it almost always will in real life.

crusaderky:
CI failures are unrelated. This is ready for final review. Merge blocked by https://github.com/data-apis/array-api-extra/pull/86.

scipy/_lib/array_api_extra
crusaderky, line 1: https://github.com/data-apis/array-api-extra/pull/165

crusaderky:
Test failures are unrelated. @rgommers gentle ping to review

scipy/spatial/distance.py
```diff
@@ -2648,34 +2666,25 @@ def is_valid_y(y, warning=False, throw=False, name=None):
...
+    name_str = f"'{name}' " if name else ""
     try:
         if len(y.shape) != 1:
-            if name:
-                raise ValueError(f"Condensed distance matrix '{name}' must "
-                                 "have shape=1 (i.e. be one-dimensional).")
-            else:
-                raise ValueError('Condensed distance matrix must have shape=1 '
-                                 '(i.e. be one-dimensional).')
+            raise ValueError(f"Condensed distance matrix {name_str}must "
```
rgommers, line 2673: typo: needs space before "must"

scipy/spatial/distance.py
```diff
@@ -2648,34 +2666,25 @@ def is_valid_y(y, warning=False, throw=False, name=None):
...
         if (d * (d - 1) / 2) != n:
-            if name:
-                raise ValueError(f"Length n of condensed distance matrix '{name}' "
-                                 "must be a binomial coefficient, i.e."
-                                 "there must be a k such that (k \\choose 2)=n)!")
-            else:
-                raise ValueError('Length n of condensed distance matrix must '
-                                 'be a binomial coefficient, i.e. there must '
-                                 'be a k such that (k \\choose 2)=n)!')
+            raise ValueError(f"Length n of condensed distance matrix {name_str}"
```
rgommers, line 2678: typo: space before closing `"` (also on next line)

rgommers (commented):
It looks like the only reason that many functions have `jax_jit=False` in tests is that `is_valid_linkage`/`is_valid_im` force evaluation - functions that don't call those validation utilities work. That's fine for this PR; good to know though because potentially the checks can be bypassed. 

Such bypassing is actually done in two functions with `@wraps` and an extra `validate=lazy` keyword in this PR. That does make the code a bit harder to understand - that's the tradeoff we're evaluating. Code was:
```python
    is_valid_linkage(Z, throw=True, name='Z')
    if not xp.all(xp.isfinite(y)):
        raise ValueError("The condensed distance matrix must contain only "
                         "finite values.")

    Z = np.asarray(Z)
    y = np.asarray(y)
    return xp.asarray(_optimal_leaf_ordering.optimal_leaf_ordering(Z, y))
```
and becomes:
```python
    # @wraps causes the Dask key to have prefix "optimal_leaf_ordering-*"
    @wraps(_optimal_leaf_ordering.optimal_leaf_ordering)
    def func(Z, y, validate):
        if validate:
            is_valid_linkage(Z, throw=True, name='Z')
            if not np.all(np.isfinite(y)):
                raise ValueError("The condensed distance matrix must contain only "
                                 "finite values.")
        return _optimal_leaf_ordering.optimal_leaf_ordering(Z, y)

    return xpx.lazy_apply(func, Z, y, validate=lazy,
                          shape=Z.shape, dtype=Z.dtype, as_numpy=True)
```
Is it actually necessary to use `@wraps` here? The docs say that its intended purpose is to preserve function name and docstring - but do we care, since this is not user-facing?

rgommers:
I think the `@wraps` usage was my main design-level question. Overall this looks pretty nice, and works as advertised in my local testing.

scipy/spatial/distance.py
lucascolley: I don't think this is a typo, space is deliberately added on line line 2670 to accommodate for the error making sense when `name` is not provided

rgommers: ah yes, never mind - resolving this one

scipy/spatial/distance.py
rgommers: this isn't a typo indeed, as Lucas pointed out - but the one on the next line still is

scipy/cluster/tests/test_hierarchy.py
```diff
@@ -69,24 +70,62 @@
...
+
+lazy_xp_function(single)
+lazy_xp_function(ward)
+lazy_xp_function(linkage, static_argnames=('method', 'metric', 'optimal_ordering'))
+lazy_xp_function(cut_tree, jax_jit=False, allow_dask_compute=999, 
+                 static_argnames=('n_clusters', 'height'))
+lazy_xp_function(to_tree, jax_jit=False, allow_dask_compute=999,
+                 static_argnames=('rd', ))
+lazy_xp_function(optimal_leaf_ordering, static_argnames=('metric',))
+lazy_xp_function(cophenet, jax_jit=False, allow_dask_compute=999)
```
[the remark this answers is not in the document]
crusaderky: It's the number of allowed computations. It lets us distinguish between functions that exclusively feature a non-lazified cython kernel, which can pass with `allow_dask_compute=1`, and functions that also feature many passes of eager validation. The former will be noticeably painful for Dask users but borderline useful; the latter won't be usable at all unless the user explicitly calls `persist()` just ahead of it. I could have scouted the exact number of compute calls but, pragmatically, beyond 2 nobody cares.

scipy/cluster/hierarchy.py
```diff
@@ -2241,32 +2283,85 @@ def is_valid_linkage(Z, warning=False, throw=False, name=None):
...
+         f'Linkage {name_str}contains excessive observations in a cluster'),
+        (xp.any(xp.max(Z[:, :2], axis=1) >= xp.arange(n + 1, 2 * n + 1, dtype=Z.dtype)),
+         f'Linkage {name_str}uses non-singleton cluster before it is formed.'),
+        (xpx.nunique(Z[:, :2]) < n * 2,
+         f'Linkage {name_str}uses the same cluster more than once.'),
+        throw=throw, warning=warning, materialize=materialize, xp=xp
+    )
+
+
+def _lazy_valid_checks(*args, throw=False, warning=False, materialize=False, xp):
```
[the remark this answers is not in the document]
crusaderky: When the user directly calls `is_valid_linkage`, it uses `materialize=True`  so it will behave exactly like in numpy, but at the cost of not being able to run inside jax.jit.
When another function, such as `linkage`, calls this function internally, it sets `materialize=False` and skips the raises.

To recap:

| backend | eager backends | Dask / JAX |
| --- | --- | --- |
| user calls `is_valid_linkage` directly | raise on invalid linkage | raise on invalid linkage; materialize graph |
| user calls `linkage` | raise on invalid linkage | skip content checks; do not materialize graph |

crusaderky:
> Is it actually necessary to use @wraps here? The docs say that its intended purpose is to preserve function name and docstring - but do we care, since this is not user-facing?

It is user-facing for Dask users. The function name will show up prominently on their real-time dashboard as well as on any metrics dashboards they may have hooked to Prometheus (Coiled offers one out of the box).
I've renamed these internal functions to `optimal_leaf_ordering_` and `leaders_` respectively without `@wraps` for the sake of simplicity.

scipy/cluster/tests/test_hierarchy.py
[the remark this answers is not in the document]
crusaderky: I've tweaked a few and set them to 1~2.

rgommers:
> It is user-facing for Dask users.

Ouch. Bubbling up internals like that is not great design - it seems to me that the only sane default  is to stop at the public API level.

> I've renamed these internal functions to `optimal_leaf_ordering_` and `leaders_` respectively without `@wraps` for the sake of simplicity.

Nice, that looks quite a bit better to me, and I think also keeps it readable on Dask logging.

scipy/cluster/hierarchy.py
```diff
@@ -4233,9 +4232,9 @@ def leaders(Z, T):
 
     n_obs = Z.shape[0] + 1
 
-    # @wraps causes the Dask key to have prefix "leaders-*"
-    @wraps(_hierarchy.leaders)
-    def func(Z, T, validate):
+    # The function name is prominently visible on the user-facing Dask dashboard;
+    # make sure it is meaningful.
```
rgommers, line 4236: Would it make sense to keep this comment in the source code here once for explanatory purposes, but rather than repeat it with every `lazy_apply` invocation, make it a prominent note in the [`lazy_apply` docs](https://data-apis.org/array-api-extra/generated/array_api_extra.lazy_apply.html)?

scipy/cluster/tests/test_hierarchy.py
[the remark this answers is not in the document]
rgommers: Thanks, looks good now - resolving.

scipy/cluster/hierarchy.py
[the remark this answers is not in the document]
rgommers: I'm not sure the right bottom corner is ~correct~ relevant - I tried changing the tests for a number of functions to `jax_jit=True`, and they do try to materialize but fail (not `linkage`, because it doesn't end up calling `_lazy_valid_checks`, but all the functions that do do that):

```
scipy/cluster/hierarchy.py:2356: in _lazy_valid_checks
    conds = [bool(cond) for cond in conds]
E   jax.errors.TracerBoolConversionError: Attempted boolean conversion of traced array with shape bool[].
E   The error occurred while tracing the function fcluster at /home/rgommers/code/pixi-dev-scipystack/scipy/scipy/build-install/lib/python3.12/site-packages/scipy/cluster/hierarchy.py:2463 for jit. This concrete value was not available in Python because it depends on the value of the argument Z.
E   See https://jax.readthedocs.io/en/latest/errors.html#jax.errors.TracerBoolConversionError
E   --------------------
E   For simplicity, JAX has removed its internal frames from the traceback of the following exception. Set JAX_TRACEBACK_FILTERING=off to include these.
```

scipy/cluster/hierarchy.py
rgommers: It's an error rather than a warning. That said, I think it's reasonable to allow bypassing it here. This is similar to why linalg functions have a `check_finite` keyword - it's a costly check that the user may not want if they've already tested their code and can trust their input data.

rgommers:
This looks good to me now. Module final tweaks and possible fresh insights when I wake up, I plan to hit the green button tomorrow. If anyone still wants to look at this, now would be a good time. It's very much okay if no one has more comments though - we'll revisit the overall design and tradeoffs of JAX (and Dask) support soon on gh-22246.

scipy/cluster/hierarchy.py
crusaderky: https://github.com/data-apis/array-api-extra/pull/189

crusaderky:
> > It is user-facing for Dask users.
> 
> Ouch. Bubbling up internals like that is not great design - it seems to me that the only sane default is to stop at the public API level.

Depends. It is very much desirable when the function name is captured from something that the final user has decorated with `@dask.delayed`. Here, the decorator is applied deep into library calls, so it becomes confusing.
But yes, even with a plain `dask.array` workflow it gets very confusing very fast. That's why I designed [spans](https://distributed.dask.org/en/stable/spans.html) to help with it.

rgommers (approved):
This LGTM now, the diff is quite clean and reading it gives a good sense of what it takes to support graph-based libraries I think. The pattern to add testing for `jax.jit` also seems understandable - once we've done the evaluation of pros and cons and it's clear we're keeping this in its current form, we should add docs for that patterns to the dev guide. In it goes! Nice work @crusaderky 🚀

Issue #22194, opened 2024-12-27
RFC: stats: sunsetting `scipy.stats.mstats`


mdhaber:
`scipy.stats.mstats` is a mostly-separate re-implementation of `scipy.stats` with support for NumPy masked arrays. Masked values are treated as missing: for 1-D slices, the result is typically the same as if the masked value were not present.

While there seems to be demand for statistical functions to support missing values, I'd suggest that having two separate implementations of these functions is not the best way to satisfy the need. 

- Maintaining two implementations is approximately twice the work of maintaing a single implementation that supports masked and non-masked arrays.
- The two implementations have fallen out of sync and will inevitably continue to do so. This would seem to introduce an unfortunate choice between the more capable `scipy.stats` function or the masked capabilities of its `scipy.stats.mstats` counterpart[^1].

[^1]: If the `scipy.stats` version did not already support masked arrays - but many do. (Addressed below.)

I have seen the opinion that we can combine the implementations but must maintain a separate `scipy.stats.mstats` namespace. While this does not double the workload, maintaining two interfaces is more work than maintaining one. For instance, many `scipy.stats.mstats` are missing "Returns" (https://github.com/scipy/scipy/issues/22065#issuecomment-2541955142) and "Examples" (gh-7168) sections of their documentation. Also, having separate interfaces for essentially identical functionality is unnecessarily complicated for users.

I see two other reasons why a namespace should not be devoted to NumPy masked arrays.
- While not actually deprecated, NumPy masked arrays themselves are problematic and most unmaintained.
- NumPy masked arrays are not compatible with the Python Array API and are explicitly rejected by our `array_namespace` function.

Fortunately, many `scipy.stats` functions already offer the same functionality as their `scipy.stats.mstats` counterparts, making the separate namespace redundant. There are actually *two* obvious ways [^2] to ignore missing values in most `scipy.stats` functions with a `scipy.stats.mstats` counterpart:
- Replace the masked values with `nan` and use `nan_policy='omit'`.
- Simply pass the masked array to the `scipy.stats` function. This behavior has been handled by the `_axis_nan_policy` decorator for several years.

Both of these avoid a common pitfall of NumPy masked arrays, which mask non-finite values that arise during calculations. This behavior is problematic because NaNs and infinities should not always be treated the same as missing data.

[...]

PR #22352, opened 2025-01-17
ENH: stats.quantile: add array API compatible quantile function

mdhaber:
#### Reference issue
data-apis/array-api#795
gh-22194

- Some SciPy functions will need an array API compatible quantile or median function if they are to be translated to the array API.
- Looking toward gh-22194, we'll need a function to replace `mquantiles`, which is one of the most-used functions in `stats.mstats`. `scoreatpercentile` is not be a great substitute for a few reasons, including the unfamiliar name.
- Based on data-apis/array-api#795, its not looking like the array API standard will converge on a `quantile` function that would do everything we need it to do.

This PR adds an array API compatible `quantile` function with native support for `axis`, `nan_policy` (for the data array), and `keepdims`. Rather than perform a cartesian product operation like `numpy.quantile` (compute all quantiles for all slices), it follows more familiar and flexible broadcasting rules. NaNs in the probability array produce NaNs in the output rather than causing the entire operation to fail.

#### Additional information
- I imagine that support for `weights` will be requested. If so, we can discuss that in a follow-up issue. 
- In the first commit, this does not support masked arrays, but it would be trivial to support [marray](https://github.com/mdhaber/marray/)s natively, and we *could* support NumPy masked arrays by either converting them to `marray`s or setting masked elements to NaN and using `nan_policy='omit'`. (The latter, of course, would not work if the desired `nan_policy='propagate'`.)
- Initially, I've only implemented methods 4-9. It will be easy to add 1-3; I just didn't. We could also allow the method to be specified as a tuple of floats, which could be interpreted as the `alphap` and `betap` of `mquantiles`.


scipy/stats/_quantile.py
```diff
@@ -0,0 +1,268 @@
...
+
+    contains_nans = _contains_nan(x, nan_policy, xp_omit_okay=True, xp=xp)
+
+    if keepdims not in {None, True, False}:
+        message = "If specified, `keepdims` must be True or False."
+        raise ValueError(message)
+
+    dtype = xp.result_type(p, x)
+
+    if x.shape[axis] == 0:
```
mdhaber, line 48: Should add a comment about what's going on here. If data has length zero along axis, the result will be an array of NaNs just as if the data had length 1 along axis and were filled with NaNs.

scipy/stats/_quantile.py
```diff
@@ -0,0 +1,268 @@
...
+
+    if (keepdims is False) and (p.shape[axis] != 1):
+        message = "`keepdims` may be False only if the length of `p` along `axis` is 1."
+        raise ValueError(message)
+    keepdims = (p.shape[axis] != 1) if keepdims is None else keepdims
+
+    y = xp.moveaxis(y, axis, -1)
+    p = xp.moveaxis(p, axis, -1)
+
+    if contains_nans:
```
mdhaber, line 64: Probably should only do this with `nan_policy='omit'`. `propagate` will just return NaNs, so it doesn't affect the result, but conceptually it's not right, and it's slower.

mdhaber:
Harwell-Davis quantiles can easily be incorporated as `method='harwell-davis'` as shown in https://github.com/mdhaber/scipy/compare/quantile...hdquantile.

scipy/stats/_quantile.py
```diff
@@ -0,0 +1,268 @@
...
+        shape = (1,)*(ndim - 1) + res.shape
+        res = xp.reshape(res, shape)
+        axis = -1
+
+    res = xp.moveaxis(res, -1, axis)
+
+    if not keepdims:
+        res = xp.squeeze(res, axis=axis)
+
+    return res[()]
```
ev-br, line 268: This won't work on array-api-strict, correct? In general, this is not compliant unless res is strictly 1D

ev-br (commented):
Looks all reasonable.

One thing might be worth doing upfront is to explicitly decide (and lightly test) the policy on float32  inputs: for f32 inputs, is the output f32 for f64. I personally don't have a strong preference either way, as long as there's a decision.

Historically we were "f64 is enough for everybody", now that there are array libraries that default to f32, new functions better make a decision. As long as this one is private, I guess whichever works better for our internal usage is fine.

scipy/stats/tests/test_quantile.py
```diff
@@ -0,0 +1,128 @@
...
+def quantile_reference(x, p, *, axis, nan_policy, keepdims):
+    x, p = np.moveaxis(x, axis, -1), np.moveaxis(p, axis, -1)
+    res = quantile_reference_last_axis(x, p, nan_policy)
+    res = np.moveaxis(res, -1, axis)
+    if not keepdims:
+        res = np.squeeze(res, axis=axis)
+    return res
+
+
+@skip_xp_backends('array_api_strict', reason="No take_along_axis yet.")
```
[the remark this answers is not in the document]
mdhaber: I am happy waiting until it's released. An update when it's released is no problem.

scipy/stats/tests/test_quantile.py
```diff
@@ -0,0 +1,128 @@
...
+
+        message = "`x` must have real dtype."
+        with pytest.raises(ValueError, match=message):
+            stats.quantile(xp.asarray([True, False]), p)
+        with pytest.raises(ValueError):
+            stats.quantile(xp.asarray([1+1j, 2]), p)
+
+        message = "`p` must have real floating dtype."
+        with pytest.raises(ValueError, match=message):
+            stats.quantile(x, xp.asarray([0, 1]))
```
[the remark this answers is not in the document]
mdhaber: `x` can be integral, but `p` is a probability, and 0 and 1 are the only integer probabilities.

scipy/stats/_quantile.py
```diff
@@ -0,0 +1,268 @@
...
+    if not xp.isdtype(p.dtype, 'real floating'):
+        raise ValueError("`p` must have real floating dtype.")
+
+    axis_none = axis is None
+    ndim = max(x.ndim, p.ndim)
+    if axis_none:
+        x = xp_ravel(x)
+        p = xp_ravel(p)
+        axis = 0
+    elif np.iterable(axis) or int(axis) != axis:
```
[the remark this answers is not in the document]
mdhaber: Well, currently this will currently raise an informative error for non-tuple iterables.

scipy/stats/_quantile.py
mdhaber: Then we will do the usual `res[()] if res is None else res`. Surely there will be other unfortunate verbosities I forgot, and the test suite will let us know.

scipy/stats/tests/test_quantile.py
```diff
@@ -0,0 +1,144 @@
+import pytest
+import numpy as np
+from matplotlib.pyplot import axis
```
mdhaber, line 3: Not sure how this got there.
```suggestion
```

mdhaber:
I think only one test failed for a good reason, and then the [GPU job reported failure for all subsequent tests](https://github.com/scipy/scipy/actions/runs/13079947634/job/36501025634). Is that supposed to happen?

rgommers:
> I think only one test failed for an understandable reason, and then the [GPU job reported failure for all subsequent tests](https://github.com/scipy/scipy/actions/runs/13079947634/job/36501025634). Is that supposed to happen?

That can happen when a GPU gets into a bad state. If the log gets eaten then it's a problem, because you cannot see where the problem is - this happened during setting up the GPU job and turned out to be a Cirrus virtualization bug. If the tests complete and failures are reported normally by `pytest`, then it is usually normal.

mdhaber:
> That can happen when a GPU gets into a bad state

Sure, but I didn't see how the failed assertion should cause the GPU to get into a bad state. I think it was just an incorrect assumption about dtypes in the test, so the result dtype was different than expected; I don't think anything else went wrong. Would a failed assertion cause a GPU to get into a bad state?

rgommers:
> Sure, but I didn't see how the failed assertion should cause the GPU to get into a bad state

Well, the first failure contains a pretty good clue:
```
../../../../.github/workflows/.pixi/envs/torch-cuda/lib/python3.12/site-packages/torch/testing/_comparison.py:711: in compare
    self._compare_values(actual, expected)
       actual     = <[RuntimeError('CUDA error: device-side assert triggered\nCUDA kernel errors might be asynchronously reported at some ...pile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.\n') raised in repr()] Tensor object at 0x7f37083759a0>
```

So it's not a `assert xxx` executed by `pytest` that failed, but an assert executed on the GPU. That tends to result in this kind of behavior.

mdhaber:
I was surprised by the idea that failure of a `pytorch` test assertions would cause this sort of error state when executed on the GPU, so I dug a bit deeper. It turns out that the problem actually occured earlier due to pytorch/pytorch#146211, and the stack trace was just incorrect.

ev-br (approved):
This now LGTM, and the discourse thread is silent for about a week. Let's give it at least one more weekend, I'd say, and then interpret it as a lazy consensus.

rgommers:
That makes sense, thanks for digging in deeper @mdhaber.

mdhaber:
https://github.com/mdhaber/scipy/compare/quantile...hdquantile now adds methods 1-3, a Harwell-Davis method, and `marray` support. I'll open that up as soon as this lands.

ev-br:
Okay, the discourse thread is still silent, the code LGTM, let's merge and keep the ball rolling. Thanks Matt, Ralf.

PR #22353, opened 2025-01-18
BUG: sparse: fix selecting wrong dtype for coo coords

PTNobel:
This fixes a bug triggered by CVXPY where reshaping a (1045507, 523266) matrix to (547078265862, 1) causes a negative index to be formed because the index type is being selected based on the old coordinates and shape and not the new one.

This seems to be a regression from 1.14.1

### DIFF
--- a/scipy/sparse/_coo.py
+++ b/scipy/sparse/_coo.py
@@ -148,7 +148,7 @@ class _coo_base(_data_matrix, _minmax_mixin):
         else:
             new_coords = np.unravel_index(flat_coords, shape, order=order)
 
-        idx_dtype = self._get_index_dtype(new_coords, maxval=max(shape))
+        idx_dtype = self._get_index_dtype(self.coords, maxval=max(shape))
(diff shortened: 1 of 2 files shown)


dschult:
Thanks for catching this!  
self.shape is one arg of the `check_shape` function that creates `shape` . But so is `args` and thus we need to use “shape” instead of “self.shape” here.   I will look elsewhere for this kind of issue, but this should definitely be fixed.

And I agree that this is a back port candidate…

dschult (approved):
I added a test so this doesn't come back again.

Also, the suggested change was too much. :)  In the case when the flatten coords need to be int64, the computation might reshape the values back into a shape that fits int32. So we will leave the `self.coords` in the `_get_index_dtype` call (to pick up any input array larger than needed and maintain it) and rely on `shape` to pick up what's needed.  Note that using `new_coords` will pick up 'int64' if flat_coords needs to be `int64`, so that's too much.

This PR and tests are ready.  I approve this PR. 
Thanks @PTNobel

j-bowhay (approved):

PR #22357, opened 2025-01-18
MAINT: Stop installing `libhighs`

### LINKED ISSUE #22349: BLD: libhighs.a static library is installed
At
https://github.com/scipy/scipy/blob/2299d44490067ec50ca144987e16bb2b7597f462/scipy/optimize/_highspy/meson.build#L49
we see that the `libhighs.a` static library for the `_highspy` component of scipy.optimize is installed alongside the python code (`_highs_wrapper.py`) and compiled python extensions (`_core.so` and `_highs_options.so`).

But it seems to me that libhighs.a is redundant for the installation. If I understood correctly, it is already compiled into _core.so and _highs_options.so, and doesn't seem to serve any purpose as an installed file.

The presence of  `libhighs.a` confuses the Debian build system, for instance, which unifies the build for different python versions (by adding a version-specific suffix to the .so extensions).  The debian build system doesn't know whether to treat this `libhighs.a` file as specific to a given python version or not.

Is it necessary to install the libhighs.a file?

HaoZeke:
Fix suggested by @drew-parsons , and seems to work alright, `highs` symbols are present in `_core.so` anyway.

Closes #22349

### DIFF
--- a/scipy/optimize/_highspy/meson.build
+++ b/scipy/optimize/_highspy/meson.build
@@ -46,8 +46,7 @@ _highslib = static_library(
     # private symbols don't get re-exported.
     gnu_symbol_visibility: 'inlineshidden',
     pic: true,
-    install: true,
-    install_dir: py3.get_install_dir() / 'scipy/optimize/_highspy',
+    install: false,
 )
 
 _highs_dep = declare_dependency(


lucascolley (approved):
CI is happy, thanks @HaoZeke!

PR #6722, opened 2016-10-25
ENH: cython_special: GIL free error handling

person142:
This changes the error handling in `cython_special` to be based on a global errno-style variable. With these changes error handling looks like:

``` cython
import scipy.special.cython_special as sc

cdef double res
cdef sc.error_t err

res = sc.spence(-1.0)
err = sc.get_errno()
if err != sc.OK:
    # Do something
```

To achieve that effect with the old error handling system one would have to do something like:

``` cython
cimport scipy.special.cython_special as sc
import warnings

cdef double res

warnings.simplefilter('raise')
try:
    res = sc.spence(-1.0)
except:
    # Do something
```

Note that with the old error handling changing the flow of the program based on errors requires grabbing the GIL. 

If SciPy is compiled by a C compiler with OpenMP support, then the new error handling is threadsafe for OpenMP threads. In particular this covers uses of `cython.parallel`; e.g. something like the following code:

``` cython
from cython cimport parallel
cimport scipy.special.cython_special as sc

cdef int n
for n in parallel.prange(10):
    sc.loggamma(<double>n)
    if sc.get_errno() != sc.OK:
        # Do something
```

is fine.

See https://github.com/scipy/scipy/issues/6681 for more discussion.

Some aspects of this PR are ugly and could perhaps be done in a nicer way; in particular it currently implements a custom `build_ext` in the top-level `setup.py` that tries to detect whether the C compiler supports OpenMP. A better way of doing this would be welcome. (Presumably changing `numpy.distutils` would be a better way of doing this, but ISTM it is expedient to get this figured out before `cython_special` is in a release.)

PR #22358, opened 2025-01-18
MAINT: `special.nctdtrit`: migrate to boost

dschmitz89:
#### Reference issue
Part of #21966

#### Additional information
Originally, I wanted to use the same test data as for `nctdtr` from [here](https://github.com/scipy/scipy/blob/528c900c07378f4b0e5cda02f1fcaa0fcf741d8f/scipy/special/tests/test_cdflib.py#L663) but the quantile has issues especially at `q=1`. Instead, I took a hand full of them into a new test dedicated to `nctdtrit`.


dschmitz89:
In the last commit I marked `nctdtrit` as a known failure for the cython API tests. The issue is that the testing parameters include `1.0`, landing in the extreme right tail for CDF related functions. `nctdtrit` yields an OverflowError in that regime. Testing out other values resulted in errors for other functions from `special` unfortunately.

We should probably overhaul that cython testing code in a way that it is possible to pass testing parameters to individual functions.

scipy/special/boost_special_functions.h
```diff
@@ -1018,22 +1018,47 @@ nct_pdf_double(double x, double v, double l)
...
+    }
+    catch (const std::domain_error& e) {
+        sf_error("nctdtrit", SF_ERROR_DOMAIN, NULL);
+        y = NAN;
+    } catch (const std::overflow_error& e) {
+        sf_error("nctdtrit", SF_ERROR_OVERFLOW, NULL);
+        y = INFINITY;
+    } catch (const std::underflow_error& e) {
+        sf_error("nctdtrit", SF_ERROR_UNDERFLOW, NULL);
+        y = -INFINITY;
```
steppi, line 1043:
```suggestion
    } catch (const std::underflow_error& e) {
        sf_error("nctdtrit", SF_ERROR_UNDERFLOW, NULL);
        y = 0; 
```

typo?

scipy/special/boost_special_functions.h
```diff
@@ -957,11 +957,11 @@ nct_cdf_wrap(const Real v, const Real l, const Real x)
...
     }
     if (v <= 0) {
-	sf_error("nctdtr", SF_ERROR_DOMAIN, NULL);
-	return NAN;
+	    sf_error("nctdtr", SF_ERROR_DOMAIN, NULL);
+	    return NAN;
     }
     if (std::isinf(x)) {
-	return  (x > 0) ? 1.0 : 0.0;
+	    return  (x > 0) ? 1.0 : 0.0;
```
steppi, line 964: some kind of editor setting maybe?

steppi (commented):
Thanks @dschmitz89. Can you explain further what's going on with `test_cython_api.py`? That doesn't seem like something that should be skipped, if calling the the `cython_special` `nctdtrit` is causing an overflow error, that suggests something is wrong.

scipy/special/boost_special_functions.h
dschmitz89: 🙈

dschmitz89:
> Thanks @dschmitz89. Can you explain further what's going on with `test_cython_api.py`? That doesn't seem like something that should be skipped, if calling the the `cython_special` `nctdtrit` is causing an overflow error, that suggests something is wrong.

The overflow error occurs in the Python API for `nctdtrit` for some of the points that are autogenerated in the cython api testing machinery from [here](https://github.com/scipy/scipy/blob/1a3c042112d7d7a1e0959ac5039c6bb409db3dad/scipy/special/tests/test_cython_special.py#L12). `cython_special` results in a `SystemError` then, the `sf_error` mechanism does not seem to work there. In principle though, `nctdtrit` works also from cython, the issue is the choice of points where it is tested. That's why I suggested to potentially update the `cython_special` testing mechanism to accept individual testing parameters by function. Suppressing `OverflowError` in the tests is an anti pattern imo.

steppi:
> > Thanks @dschmitz89. Can you explain further what's going on with `test_cython_api.py`? That doesn't seem like something that should be skipped, if calling the the `cython_special` `nctdtrit` is causing an overflow error, that suggests something is wrong.
> 
> The overflow error occurs in the Python API for `nctdtrit` for some of the points that are autogenerated in the cython api testing machinery from [here](https://github.com/scipy/scipy/blob/1a3c042112d7d7a1e0959ac5039c6bb409db3dad/scipy/special/tests/test_cython_special.py#L12). `cython_special` results in a `SystemError` then, the `sf_error` mechanism does not seem to work there. In principle though, `nctdtrit` works also from cython, the issue is the choice of points where it is tested. That's why I suggested to potentially update the `cython_special` testing mechanism to accept individual testing parameters by function. Suppressing `OverflowError` in the tests is an anti pattern imo.

I'll look into it when I have a chance. By the Python API for `nctdtrit` do you mean the regular ufunc, or the Python API from `cython_special` that was added for testing purposes?   `sf_error` is disabled for `cython_special` due to issues surrounding sharing error handling state between extension modules. See https://github.com/scipy/scipy/pull/6722#issuecomment-256539142 and this blog post, https://labs.quansight.org/blog/libsf-error-state, for more context. When an error occurs, `cython_special` functions should just return `NaN` I think. I don't think there should be a SystemError being raised.

dschmitz89:
> > > Thanks @dschmitz89. Can you explain further what's going on with `test_cython_api.py`? That doesn't seem like something that should be skipped, if calling the the `cython_special` `nctdtrit` is causing an overflow error, that suggests something is wrong.
> > 
> > 
> > The overflow error occurs in the Python API for `nctdtrit` for some of the points that are autogenerated in the cython api testing machinery from [here](https://github.com/scipy/scipy/blob/1a3c042112d7d7a1e0959ac5039c6bb409db3dad/scipy/special/tests/test_cython_special.py#L12). `cython_special` results in a `SystemError` then, the `sf_error` mechanism does not seem to work there. In principle though, `nctdtrit` works also from cython, the issue is the choice of points where it is tested. That's why I suggested to potentially update the `cython_special` testing mechanism to accept individual testing parameters by function. Suppressing `OverflowError` in the tests is an anti pattern imo.
> 
> I'll look into it when I have a chance. By the Python API for `nctdtrit` do you mean the regular ufunc, or the Python API from `cython_special` that was added for testing purposes? `sf_error` is disabled for `cython_special` due to issues surrounding sharing error handling state between extension modules. See [#6722 (comment)](https://github.com/scipy/scipy/pull/6722#issuecomment-256539142) and this blog post, https://labs.quansight.org/blog/libsf-error-state, for more context. When an error occurs, `cython_special` functions should just return `NaN` I think. I don't think there should be a `SystemError` being raised.

The `OverflowError` occurs in the regular ufunc: `nctdtrit(1, 1, 1)` is an example.

steppi:
> The `OverflowError` occurs in the regular ufunc: `nctdtrit(1, 1, 1)` is an example.

Yeah, that's a bug. I'm seeing this on your branch

```
In [2]: nctdtrit(1, 1, 1)
OverflowError                             Traceback (most recent call last)
Cell In[2], line 1
----> 1 nctdtrit(1, 1, 1)

OverflowError: Error in function quantile(const non_central_t_distribution<d>, %1%): Overflow Error
```

but the behavior on main is

```
In [2]: nctdtrit(1, 1, 1)
Out[2]: np.float64(nan)
```

```
In [4]: with special.errstate(all="warn"):
   ...:     nctdtrit(1, 1, 1)
   ...:
<ipython-input-4-da5ed05adc18>:2: SpecialFunctionWarning: scipy.special/nctdtrit: (invalid input argument) Input parameter p is out of range
  nctdtrit(1, 1, 1)
```

special function ufuncs by default shouldn't raise errors like this due to a single bad input. Error handling is instead controlled by `special.seterr` and `special.errstate`. This needs fixing. I think you just need to do some input validation.

dschmitz89:
@steppi : you are right. I had made a copy paste error, with the right boost policy it works.

scipy/special/boost_special_functions.h
steppi: @dschmitz89.  We don't use 8 space indents in C++ code in special. Can you address this?

scipy/special/boost_special_functions.h
```diff
@@ -1023,11 +1023,11 @@ nct_cdf_wrap(const Real v, const Real l, const Real x)
...
```
dschmitz89, line 1030:
```suggestion
	return  (x > 0) ? 1.0 : 0.0;
```

scipy/special/boost_special_functions.h
dschmitz89: Done in last commit.

steppi (approved):
Thanks @dschmitz89. This looks good now.

PR #21707, opened 2024-10-15
ENH: `stats.make_distribution`: generate a `ContinuousDistribution` from existing `rv_continuous`

mdhaber:
#### What does this implement/fix?
This adds a function that accepts an instance of `stats.rv_continuous` and returns a subclass of `stats.ContinuousDistribution`. The `ContinuousDistribution` uses overridden methods of the `rv_continuous` class where possible (e.g. `pdf`, `logpdf`) but uses its own generic method implementations (which are often faster and more accurate).

As an example inspired by gh-18799:
```python3
from scipy import stats
GH = stats.make_distribution(stats.gausshyper)
X = stats.gausshyper(a=1.5, b=2.5, c=2, z=0)
Y = GH(a=1.5, b=2.5, c=2, z=0)
```
`X` is a "frozen" distribution object representing the Gauss hypergeometric distribution and `Y` is a corresponding instance of a `ContinuousDistribution`. Since `gausshyper_gen` defines `_pdf`, the `pdf` methods of the two objects behave identically:
```python3
X.pdf(0.3), Y.pdf(0.3)
# (np.float64(1.6337206456367985), np.float64(1.6337206456367985))
```
but because `gausshyper_gen` does not define `_cdf`, `Y` will use the `ContinuousDistribution` framework to evaluate the CDF, which tends to be much faster and more accurate than the generic implementation in the `rv_continuous` framework.
<details>

```python3
import numpy as np
import matplotlib.pyplot as plt
from scipy import stats

import numpy as np
from scipy import stats
from scipy import special as sc
from time import perf_counter
from mpmath import mp
mp.dps = 50

a, b, c, z = 1.5, 2.5, 2, 0

def _pdf_mpmath(x):
    a, b, c, z = 1.5, 2.5, 2, 0
    a, b, c, z = mp.mpf(a), mp.mpf(b), mp.mpf(c), mp.mpf(z)
    Cinv = mp.gamma(a) * mp.gamma(b) / mp.gamma(a + b) * mp.hyp2f1(c, a, a + b, -z)
    return mp.one / Cinv * x ** (a - mp.one) * (mp.one - x) ** (b - mp.one) / (mp.one + z * x) ** c

X = stats.gausshyper(a=a, b=b, c=c, z=z)
GH = stats.make_distribution(stats.gausshyper)
Y = GH(a=a, b=b, c=c, z=z)

x = np.linspace(0, 1, 1000)[1:]
tic = perf_counter()
cdf_vals = X.cdf(x)
[...]

PR #22371, opened 2025-01-20
ENH: stats.make_distribution: allow definition of custom distributions

### LINKED ISSUE #12133: How to define new distributions?
I would like to define new distributions using `scipy.stats`. Such a distribution would have a set of parameters, and one could evaluate its PDF, CDF, create random variates etc.

For continuous distributions, there is a short description under "Subclassing" in the documentation of `rv_continuous`. It says one should subclass `rv_continuous`, and as a minimum redefine `_pdf` and `_cdf`, plus possibly a list of others. The description alludes to a lot of things that I don't understand, or understand only very superficially (e.g. "frozen distributions").

I guess one could understand this description if there were some kind of general documentation of the design of `scipy.stats`, but I wasn't able to find it.

Questions:

- Can I redefine `__init__`, and if yes, how should I call `super().__init__`?

- In case I am not supposed to redefine `__init__`, how do I pass parameters to my subclass?

- What are the constraints on the signatures of the methods to be redefined? For example, do I have to provide for distribution parameters to be passed, even though I just want to define them upon instantiation?

- Do I have to support `loc` and `scale`, though I don't want to translate and scale my distribution? The phrase "re-defining at least the _pdf or the _cdf method (normalized to location 0 and scale 1)" seems to suggest that these parameters are automatically taken care of?

- If I implement `_rvs`, can I assume that `self.random_state` is defined? Should I implement a parameter `random_state` myself, or is that taken care of by `rv_continuous.rvs`?

- How does `rvs` pass the number of samples (`size`) to `_rvs`?

- How are the interface methods (e.g. `ppf`) based on the methods I redefine? If I have performance problems with some interface method, which internal method should I additionally redefine? A dependency graph with some indication of performance bottlenecks and loss of precision would be useful.

- > A note on shapes: subclasses need not specify them explicitly. In this case, shapes will be automatically deduced from the signatures of the overridden methods (pdf, cdf etc). If, for some reason, you prefer to avoid relying on introspection, you can specify shapes explicitly as an argument to the instance constructor.

    I have only the faintest idea what this means. How exactly are shapes "automatically deduced from the signatures of the overridden methods"? How can I "specify shapes explicitly as an argument to the instance constructor" if `__init__` is not in the list of methods to be redefined?

These are just the questions I came up with at my current state of understanding, I'm sure if I understood more I had much more to ask.

### LINKED ISSUE #21606: ENH: stats: generic power law with negative index

I see SciPy has a very specific implementation of the Powerlaw. It is restricted in unit intervals and has a positive index. In astrophysics, we use Powerlaw with negative indexes too with broader support intervals (the common thing it is truncated on both sides). [NumPyro](https://github.com/pyro-ppl/numpyro/) recently provided this [distribution](https://num.pyro.ai/en/stable/distributions.html#doublytruncatedpowerlaw). It would be great to have this in `scipy.stats` module.

Either add these changes to the existing `scipy.stats.powerlaw` or provide a new one.

_No response_

_No response_

mdhaber:
#### Reference issue
gh-21707
Closes gh-21606
Closes gh-12133

Updates `make_distribution` to allow users to easly define custom distributions. For example, a more general power law distribution was requested in gh-21606. This might be defined like:

```python3
import numpy as np
from scipy import stats

class PowerLaw:
    @property
    def __make_distribution_version__(self):
        return "1.16.0"

    @property
    def parameters(self):
        return {'a': {'endpoints': (0, np.inf), 'inclusive': (False, False)},
                'b': {'endpoints': ('a', np.inf), 'inclusive': (False, False)},
                'c': {'endpoints': (-np.inf, np.inf), 'inclusive': (False, False)}}

    @property
    def support(self):
        return 'a', 'b'

    def pdf(self, x, a, b, c):
        return (c+1) * x**c / (b**(c+1) - a**(c+1))

```

Compare against our truncated Pareto distribution:
```python3

X = stats.make_distribution(PowerLaw())(a=1.0, b=3.0, c=-2.0)
sample = X.sample(shape=10000)
Y = stats.make_distribution(stats.truncpareto)(b=1, c=3)

ax = plt.gca()
X.plot(ax=ax)
x = np.linspace(1, 3, 300)
ax.plot(x, Y.pdf(x), '--')
ax.hist(sample, bins=30, density=True, alpha=0.5)
ax.legend(['PowerLaw (pdf)', 'TruncatedPareto', 'PowerLaw (histogram)'])
plt.show()
```

image

#### Additional information
This could be a bit better about sharing code with that used for instances of `rv_continuous`, but I copy-pasted for now to minimize conflicts with gh-22312.
Currently, I don't document how the user's class can override the behavior of `sample` because the interface of `_sample_formula` requires arguments other than those provided to the public `sample` method + the distribution parameters. This can be addressed in a follow-up.
Currently, this doesn't allow the `support` to be an arbitrary function of the shape parameters. ~~That would be a very easy follow-up.~~ Done in gh-22560. See https://github.com/scipy/scipy/pull/22560#issuecomment-2676153771 for additional follow-up work that remains.
Input validation could be improved if someone were so inclined. (E.g., a motivated contributor could make this emit more helpful errors if the user-provided class does not actually satisfy the required interface.)

### DIFF
--- a/scipy/stats/_distribution_infrastructure.py
+++ b/scipy/stats/_distribution_infrastructure.py
@@ -3496,7 +3496,7 @@ _distribution_names = {
 
 # beta, genextreme, gengamma, t, tukeylambda need work for 1D arrays
 def make_distribution(dist):
-    """Generate a `ContinuousDistribution` from a distribution-like object
+    """Generate a `ContinuousDistribution` class from a compatible object
 
     The argument may be an instance of `rv_continuous` or an instance of
     another class that satisfies the interface described below.
@@ -3547,13 +3547,14 @@ def make_distribution(dist):
         The class **must** also define a ``pdf`` method and **may** define methods
         ``logentropy``, ``entropy``, ``median``, ``mode``, ``logpdf``,
         ``logcdf``, ``cdf``, ``logccdf``, ``ccdf``,
-        ``ilogcdf``, ``icdf``, ``ilogccdf``, and ``iccdf````.
-        All methods must accept the parameters of the distributions as keyword arguments
-        and any positional-only arguments accepted by the corresponding method of
-        `ContinuousDistribution`. Methods ``moment_raw``, ``moment_central``,
-        ``moment_standardized`` may be defined to accept the ``order`` of the moment
-        by position, acccept all distribution parameters by keyword, and return the raw,
-        central, and standardized moments of the distribution, respectively.
+        ``ilogcdf``, ``icdf``, ``ilogccdf``, and ``iccdf``.
+        If defined, these methods must accept the parameters of the distributions as
+        keyword arguments and also accept any positional-only arguments accepted by
+        the corresponding method of `ContinuousDistribution`. Methods ``moment_raw``,
+        ``moment_central``, ``moment_standardized`` may also be defined; if so,
+        they must accept the ``order`` of the moment by position, accept all
+        distribution parameters by keyword, and return the raw, central, and
+        standardized moments of the distribution, respectively.
 
     Returns
     -------
@@ -3791,9 +3792,8 @@ def _make_distribution_custom(dist):
 
     support_etc = _combine_docs(CustomDistribution, include_examples=False).lstrip()
     docs = [
-        f"This class represents `{dist.__class__.__name__}` as a subclass of "
+        f"This class represents `{repr_str}` as a subclass of "
         "`ContinuousDistribution`.",
-        f"The `repr`/`str` of class instances is `{repr_str}`.",
         f"The PDF of the distribution is defined {support_etc}"
     ]
     CustomDistribution.__doc__ = ("\n".join(docs))
--- a/scipy/stats/_distribution_infrastructure.py
+++ b/scipy/stats/_distribution_infrastructure.py
@@ -3496,7 +3496,10 @@ _distribution_names = {
 
 # beta, genextreme, gengamma, t, tukeylambda need work for 1D arrays
 def make_distribution(dist):
-    """Generate a `ContinuousDistribution` from an instance of `rv_continuous`
+    """Generate a `ContinuousDistribution` from a distribution-like object
+
+    The argument may be an instance of `rv_continuous` or an instance of
+    another class that satisfies the interface described below.
 
     The returned value is a `ContinuousDistribution` subclass. Like any subclass
     of `ContinuousDistribution`, it must be instantiated (i.e. by passing all shape
@@ -3514,7 +3517,43 @@ def make_distribution(dist):
     Parameters
     ----------
     dist : `rv_continuous`
-        Instance of `rv_continuous`.
+        Instance of `rv_continuous` OR an instance of any class with the following
+        attributes:
+
+        __make_distribution_version__ : str
+            A string containing the version number of SciPy in which this interface
+            is defined. The preferred interface may change in future SciPy versions,
+            in which case support for an old interface version may be deprecated
+            and eventually removed.
+        parameters : dict
+            A dictionary describing the parameters of the distribution.
+            Each key is the name of a parameter,
+            and the corresponding value is itself a dictionary with the following items.
+
+            endpoints : tuple
+                A tuple defining the lower and upper endpoints of the domain of the
+                parameter. Allowable values are floats or the name (string) of another
+                parameter.
+
+            inclusive : tuple of bool
+                A tuple specifying whether the endpoints are included within the domain
+                of the parameter.
+
+        support : tuple
+            A tuple defining the lower and upper endpoints of the support of the
+            distribution. Allowable values are floats or the name (string) of a
+            parameter.
+
+        The class **must** also define a ``pdf`` method and **may** define methods
+        ``logentropy``, ``entropy``, ``median``, ``mode``, ``logpdf``,
+        ``logcdf``, ``cdf``, ``logccdf``, ``ccdf``,
+        ``ilogcdf``, ``icdf``, ``ilogccdf``, and ``iccdf````.
+        All methods must accept the parameters of the distributions as keyword arguments
+        and any positional-only arguments accepted by the corresponding method of
+        `ContinuousDistribution`. Methods ``moment_raw``, ``moment_central``,
+        ``moment_standardized`` may be defined to accept the ``order`` of the moment
+        by position, acccept all distribution parameters by keyword, and return the raw,
+        central, and standardized moments of the distribution, respectively.
 
     Returns
     -------
@@ -3536,8 +3575,11 @@ def make_distribution(dist):
     >>> import numpy as np
     >>> import matplotlib.pyplot as plt
     >>> from scipy import stats
-    >>> LogU = stats.make_distribution(stats.loguniform)
-    >>> X = LogU(a=1.0, b=3.0)
+
+    Create a `ContinuousDistribution` from `scipy.stats.loguniform`.
+
+    >>> LogUniform = stats.make_distribution(stats.loguniform)
+    >>> X = LogUniform(a=1.0, b=3.0)
     >>> np.isclose((X + 0.25).median(), stats.loguniform.ppf(0.5, 1, 3, loc=0.25))
     np.True_
     >>> X.plot()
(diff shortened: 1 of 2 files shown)


scipy/stats/_distribution_infrastructure.py
```diff
@@ -3546,14 +3588,47 @@ class or its methods for more information.
...
+    ...
+    ...     @property
+    ...     def support(self):
+    ...         return 'log_a', 'log_b'
+    ...
+    ...     def pdf(self, x, log_a, log_b):
+    ...         return 1 / (x * (log_b - log_a))
+    >>>
+    >>> MyLogUniform = stats.make_distribution(MyLogUniform())
+    >>> Y = MyLogUniform(a=np.log(1.0), b=np.log(3.0))
```
mdhaber, line 3613:
```suggestion
    >>> Y = MyLogUniform(log_a=np.log(1.0), log_b=np.log(3.0))
```

mdhaber, line 3613:
```suggestion
    >>> Y = MyLogUniform(a=1.0, b=3.0)
```

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3674,6 +3749,58 @@ def _overrides(method_name):
     return CustomDistribution
 
 
+def _make_distribution_custom(dist):
+    parameters = []
+    support = getattr(dist, 'support')
+
+    for name, info in dist.parameters.items():
+        domain = _RealDomain(endpoints=info['endpoints'],
```
mdhaber, line 3757: Let's give these defaults and make them optional.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3496,7 +3496,10 @@ def plot(self, x='x', y='pdf', *, t=('cdf', 0.0005, 0.9995), ax=None):
 
 # beta, genextreme, gengamma, t, tukeylambda need work for 1D arrays
 def make_distribution(dist):
-    """Generate a `ContinuousDistribution` from an instance of `rv_continuous`
+    """Generate a `ContinuousDistribution` from a distribution-like object
```
mdhaber, line 3499:
```suggestion
    """Generate a `ContinuousDistribution` class
```

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3514,7 +3517,43 @@ def make_distribution(dist):
...
+            inclusive : tuple of bool
+                A tuple specifying whether the endpoints are included within the domain
+                of the parameter.
+
+        support : tuple
+            A tuple defining the lower and upper endpoints of the support of the
+            distribution. Allowable values are floats or the name (string) of a
+            parameter.
+
+        The class **must** also define a ``pdf`` method and **may** define methods
```
mdhaber, line 3547:
```suggestion
        The class must also define a ``pdf`` method and may define methods```

File: scipy/stats/_distribution_infrastructure.py [hunk 5]
```diff
@@ -3514,7 +3517,43 @@ def make_distribution(dist):
...
+        ``logentropy``, ``entropy``, ``median``, ``mode``, ``logpdf``,
+        ``logcdf``, ``cdf``, ``logccdf``, ``ccdf``,
+        ``ilogcdf``, ``icdf``, ``ilogccdf``, and ``iccdf````.
```
mdhaber, line 3550:
```suggestion
        ``ilogcdf``, ``icdf``, ``ilogccdf``, and ``iccdf``.
```

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3674,6 +3749,58 @@ def _overrides(method_name):
...
+        if hasattr(dist, method):
+            # Make it an attribute of the new object with the new name
+            new_method = f"_{method}_formula"
+            setattr(CustomDistribution, new_method, getattr(dist, method))
+
+    support_etc = _combine_docs(CustomDistribution, include_examples=False).lstrip()
+    docs = [
+        f"This class represents `{dist.__class__.__name__}` as a subclass of "
+        "`ContinuousDistribution`.",
+        f"The `repr`/`str` of class instances is `{repr_str}`.",
```
mdhaber, line 3796:
```suggestion
        f"This class represents `{repr_str}` as a subclass of "
        "`ContinuousDistribution`.",
```

scipy/stats/_distribution_infrastructure.py
mdhaber: @melissawm This is another instance of a problem we've had for several years: we need a way to render documentation of classes that aren't publicly importable.

In this case, `ContinuousDistribution` is a private base class from which public distributions are defined (e.g. [Normal](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.Normal.html#scipy.stats.Normal), [Uniform](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.Uniform.html#scipy.stats.Uniform)). All the subclasses will have different details that we want to render, e.g.:

image

vs

image

But their method documentation is identical; we don't need separate copies for each subclass. Instead, it would be nice to have the `ContinuousDistribution` documentation render, and subclass links to method documentation like:

image

would go to the method documentation of `ContinuousDistribution`.

Other times we have result objects that we don't intend for users to import and instantiate on their own, but we want to be able to document them so they can be used. Examples include all of [these](https://docs.scipy.org/doc/scipy/reference/stats.html#result-classes-used-in-scipy-stats).

So I think I have two separate questions here:
- How can we get `ContinuousDistribution` and result class documentation to render without making them public?
- Separately (and less relevant to *this* PR), how can we get subclass method links to go to the method documentation of base class `ContinuousDistribution`?

melissawm: Interesting! I can investigate.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3514,7 +3517,43 @@ def make_distribution(dist):
...
+        The class **must** also define a ``pdf`` method and **may** define methods
+        ``logentropy``, ``entropy``, ``median``, ``mode``, ``logpdf``,
+        ``logcdf``, ``cdf``, ``logccdf``, ``ccdf``,
+        ``ilogcdf``, ``icdf``, ``ilogccdf``, and ``iccdf````.
+        All methods must accept the parameters of the distributions as keyword arguments
+        and any positional-only arguments accepted by the corresponding method of
+        `ContinuousDistribution`. Methods ``moment_raw``, ``moment_central``,
+        ``moment_standardized`` may be defined to accept the ``order`` of the moment
+        by position, acccept all distribution parameters by keyword, and return the raw,
+        central, and standardized moments of the distribution, respectively.
```
steppi, line 3556:
```suggestion
        If defined, these methods must accept the parameters of the distributions as
        keyword arguments and also accept any positional-only arguments accepted by
        the corresponding method of `ContinuousDistribution`. Methods ``moment_raw``,
        ``moment_central``, ``moment_standardized`` may also be optionally defined, and
         if so, must accept the ``order`` of the moment by position, accept all distribution
        parameters by keyword, and return the raw, central, and standardized moments
        of the distribution, respectively.
```
I'm not sure if it's post-flu brain fog, but I had some trouble parsing this out. The suggestion is my attempt to disambiguate based on my understanding of what was meant. If you think this looks good, I'd advise against directly committing the suggestion, because I typed it directly into the GitHub comment box, and it may not be formatted correctly.

scipy/stats/_distribution_infrastructure.py
steppi: Can you elaborate? Do you mean the endpoints?

steppi (approved):
This looks good to me. I have some minor comments about the docs.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3514,7 +3517,43 @@ def make_distribution(dist):
...
     dist : `rv_continuous`
-        Instance of `rv_continuous`.
+        Instance of `rv_continuous` OR an instance of any class with the following
+        attributes:
+
+        __make_distribution_version__ : str
+            A string containing the version number of SciPy in which this interface
+            is defined. The preferred interface may change in future SciPy versions,
+            in which case support for an old interface version may be deprecated
+            and eventually removed.
```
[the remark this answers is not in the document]
mdhaber: Not necessarily, but I wanted to give ourselves the flexibility to do so.

scipy/stats/_distribution_infrastructure.py
mdhaber: I think I was thinking about both `endpoints` and `inclusive`. How about `-inf, inf`, non-inclusive?

scipy/stats/_distribution_infrastructure.py
mdhaber: This was a lot of information that I was trying to express compactly, and perhaps I went a bit too far. Sure, we can decompress it a bit.

mdhaber:
```suggestion
        If defined, these methods must accept the parameters of the distributions as
        keyword arguments and also accept any positional-only arguments accepted by
        the corresponding method of `ContinuousDistribution`. Methods ``moment_raw``,
        ``moment_central``, ``moment_standardized`` may also be defined; if so,
        they must accept the ``order`` of the moment by position, accept all
        distribution parameters by keyword, and return the raw, central, and
        standardized moments of the distribution, respectively.
```

scipy/stats/_distribution_infrastructure.py
steppi: Sounds good.

steppi (approved):
Looks good to me. Thanks @mdhaber.

PR #21799, opened 2024-11-03
ENH: sparse.linalg: add `is_sptriangular` and `spbandwidth` functions

dschult:
This provides a performant upper/lower triangular detector for CSR/CSC sparse formats.

Addresses part of #21679 (check for triangular in `spsolve` and dispatch to `is_sptriangular`)

**Questions:** 
- names... `is_sptriangular` vs `is_triangular` and `spbandwidth` vs `bandwidth`. It seemed that a `sp` prefix is commonly used in `sparse.linalg`.  What is the preference?

- Should we even bother including a bandwidth checking function yet. There is no solver to take advantage of the structure. But it was easy to write it while writing the triangular check, so I included it. I'll remove it if that's better. 

Note: Since `spsolve` converts any other formats (and even dense input), these functions do too.

**Results:**

**`is_sptriangular(A) -> lower, upper`** indicating True/False for lower or upper triangular. Diagonal returns `(True, True)`. Unstructured returns (False, False). This implementation provides a shortcut check of the ~middle column and then 1st and last columns. These should catch most matrices of interest. (middle is "typical" and 1st and last are "often different" so can rule out almost-triangular. After the shortcut, we just check `cols >= rows` and `cols <= rows` using `numpy` tools to construct these arrays.

  The timing results below show checking is much faster than solving, as desired. Letting $t_g$ be the general solver time, $t_T$ be the special triangular solver time and $t_c$ be the checking time. Then $t_c \ll t_g$ for full matrices and $t_T + t_c \ll t_g$ and $T_c \ll t_T$ for triangular case. Note that $t_c$ can be much faster for full than for triangular matrices. But both are much faster than solving time. All the times depend on the sparse `nnz`, and scaling arguments show that the "much faster than" will only get even more-so as `nnz` increases. Timing reveal the theoretical $O(nnz^3)$ for general solver, $O(nnz^2)$ for triangular solver and $O(nnz)$ for checking.

  I include the timing for a checker assuming a canonical CSC structure. But because the loops are in Python, for the triangular case, it is actually slower than using the general case numpy tools. Since all these checking times are so much smaller than solving I don't think we should use canonical methods or implement in cython. But we obviously could.
```
For full matrix:
=================== ======== ======== ======== ========
task \ nnz           1000     2000     4000     8000
=================== ======== ======== ======== ========
spsolve (t_g)        0.028s   0.195s   1.71s    12.4s
check of tri (t_c)   2.48µs   2.54µs   2.48µs   2.53µs
[...]

PR #22374, opened 2025-01-21
DOC: sparse.linalg: add two recent functions to namespace and fix doctests

dschult:
Adds `spbandwidth` and `is_sptriangular` to the `scipy/sparse/linalg` namespace as was originally intended.

While trying to help with smoke-doc failures in `scipy/sparse/linalg` I found that the two recently provided functions `spbandwidth` and `is_sptriangular` were not added to the `sparse.linalg` namespace in `__init__.py`.

This PR adds them to the namespace and then  updates the doc_strings to use those imports. It helps to have valid imports to get your doctests to pass. :)

[Edited:]
The PR to add these functions was #21799 and was merged in November.  The 1.15 release notes announce these functions and they do exist there, but they involve e.g. `scipy.sparse.linalg._dsolve.is_sptriangular` which has a private part of the access path.  This makes it `scipy.sparse.linalg.is_sptriangular`. I'm not sure if that means this should be backported. But I'm adding that label so people think about whether it should be backported. I'm fine either way.

### DIFF
--- a/scipy/sparse/linalg/__init__.py
+++ b/scipy/sparse/linalg/__init__.py
@@ -44,7 +44,7 @@ Direct methods for linear equation systems:
    spsolve -- Solve the sparse linear system Ax=b
    spsolve_triangular -- Solve sparse linear system Ax=b for a triangular A.
    is_sptriangular -- Check if sparse A is triangular.
-   spbandwidth -- find the bandwidth of a sparse matrix.
+   spbandwidth -- Find the bandwidth of a sparse matrix.
    factorized -- Pre-factorize matrix to a function solving a linear system
    MatrixRankWarning -- Warning on exactly singular matrices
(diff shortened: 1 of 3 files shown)


scipy/sparse/linalg/__init__.py
```diff
@@ -43,6 +43,8 @@
 
    spsolve -- Solve the sparse linear system Ax=b
    spsolve_triangular -- Solve sparse linear system Ax=b for a triangular A.
+   is_sptriangular -- Check if sparse A is triangular.
+   spbandwidth -- find the bandwidth of a sparse matrix.
```
dschmitz89, line 47:
```suggestion
   spbandwidth -- Find the bandwidth of a sparse matrix.
```

perimosocordiae (approved):

PR #21870, opened 2024-11-12
MAINT: spatial/special: follow-up on `std::isnan`, `std::isinf`

barracuda156:
A follow-up for https://github.com/scipy/scipy/pull/21861

@rgommers @steppi

PR #21861, opened 2024-11-10
MAINT: special: Use consistent std::isnan and std::isinf in a few C++ headers

barracuda156:
@rgommers Could you please review this? Referring to https://github.com/scipy/scipy/issues/21826#issuecomment-2466445083

It is a half of what I had to patch, but these looks rather straightforward, since the same files currently mix C99 style with C++ style randomly in C++ sources, apparently because in most cases compilers do not complain much and different committers chose different code style. 

I.e.,
https://github.com/scipy/scipy/blob/c45b3ba3df5ea9ebb517e98d62d138d0f9164724/scipy/special/boost_special_functions.h#L765
https://github.com/scipy/scipy/blob/c45b3ba3df5ea9ebb517e98d62d138d0f9164724/scipy/special/boost_special_functions.h#L813
But:
https://github.com/scipy/scipy/blob/c45b3ba3df5ea9ebb517e98d62d138d0f9164724/scipy/special/boost_special_functions.h#L76
Etc.

I am not including the rest, but if those look reasonable, I can add them too. See: https://github.com/barracuda156/scipy/commit/c8ea21327dbb48103bfa8d6349c6ee587f3b4a43
(Please ignore the old version of the patch which is currently in MacPorts, it was done for the old version of `scipy` and not in the best way possible.)

Issue #21826, opened 2024-11-07
[BUG] Build error: `error: no match for 'operator*' (operand types are 'double' and 'std::complex<float>')`


barracuda156:
### Describe your issue.

There are quite a number of errors due to `isnan` and `isinf` not being consistently used with `std::`, but other than that, there is this failure:
```
In file included from ../scipy/special/special/specfun.h:4,
                 from ../scipy/special/special.h:6,
                 from ../scipy/special/_special_ufuncs.cpp:7:
../scipy/special/special/specfun/specfun.h: In instantiation of 'std::complex<_Tp> special::specfun::e1z(std::complex<_Tp>) [with T = float]':
../scipy/special/special/expint.h:17:40:   required from 'std::complex<_Tp> special::exp1(std::complex<_Tp>) [with T = float]'
   17 |     std::complex<T> outz = specfun::e1z(z);
      |                            ~~~~~~~~~~~~^~~
../scipy/special/_special_ufuncs.cpp:304:47:   required from here
  304 |          static_cast<func_F_F_t>(special::exp1), static_cast<func_D_D_t>(special::exp1)},
      |                                               ^
../scipy/special/special/specfun/specfun.h:2099:70: error: no match for 'operator*' (operand types are 'double' and 'std::complex<float>')
 2099 |             ce1 = -el - std::log(-z) + z*ce1 - copysign(pi, z.imag())*std::complex<T>(0.0, 1.0);
      |                                                ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~
/opt/local/include/gcc14/c++/complex:400:5: note: candidate: 'template<class _Tp> std::complex<_Tp> std::operator*(const complex<_Tp>&, const complex<_Tp>&)'
  400 |     operator*(const complex<_Tp>& __x, const complex<_Tp>& __y)
      |     ^~~~~~~~
/opt/local/include/gcc14/c++/complex:400:5: note:   template argument deduction/substitution failed:
../scipy/special/special/specfun/specfun.h:2099:70: note:   mismatched types 'const std::complex<_Tp>' and 'double'
 2099 |             ce1 = -el - std::log(-z) + z*ce1 - copysign(pi, z.imag())*std::complex<T>(0.0, 1.0);
      |                                                ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #22381, opened 2025-01-21
MAINT: special: A bit of clean up in stirling2.h

WarrenWeckesser:
* We've been compiling this as C++ for a while, and we now use std::unique_ptr and new, so there is no need for the use of `#if defined(__cplusplus)`.
* Remove some unused includes.
* Use a C++-style cast.
* Make a few minor style and whitespace corrections.

### DIFF
--- a/scipy/special/stirling2.h
+++ b/scipy/special/stirling2.h
@@ -1,18 +1,10 @@
 #ifndef STIRLING_H
 #define STIRLING_H
 
-#if defined(__cplusplus)
 #include <cmath>
-using std::isinf;
-#endif
-
-#include <complex.h>
-#include <math.h>
-#include <stdio.h>
-#include <stdint.h>
-#include <stdlib.h>
-#include <memory>
+#include <complex>
 #include <limits>
+#include <memory>
 
 #include "xsf/binom.h"
 #include "xsf/lambertw.h"
@@ -72,8 +64,7 @@ double _stirling2_dp(double n, double k) {
             }
         }
     }
-    double output = curr[arraySize - 1];
-    return output;
+    return curr[arraySize - 1];
 }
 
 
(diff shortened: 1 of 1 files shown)


j-bowhay (approved):
Thanks!

steppi:
See https://github.com/scipy/scipy/issues/21826#issuecomment-2466445083, https://github.com/scipy/scipy/pull/21861 and https://github.com/scipy/scipy/pull/21870 for some context on the use of

```c++
if defined(__cplusplus)
#include <cmath>
#include <cmath>
using std::isinf;
#include <complex>
#endif
```

here. 

There were errors appearing on MacPorts due to unqualified use of `isinf` and `isnan` instead of `std::isinf` and `std::isnan`. I think the author of the PRs above just added that boilerplate to each C++ file where `isinf` and `isnan` appeared unqualified, rather than going through and changing individual lines.

Before this is merged, `isinf` on lines 50 and 60 should be replaced with `std::isinf`.

WarrenWeckesser:
That sounds good, but I'd like to understand better just what went wrong.

There is a stackoverflow question [here](https://stackoverflow.com/questions/10694255/cmath-vs-math-h-and-similar-c-prefixed-vs-h-extension-headers) with an answer that says 

> [cmath] defines symbols in the std namespace, and may also define symbols in the global namespace. [math.h] defines symbols in the global namespace, and may also define symbols in the std namespace. if you include the former and use an unqualified symbol, it may compile with one compiler but not with another. therefore it's a good idea to use [math.h]. and in general, for such header pairs, to use the [.h] version.

Is that accurate? I don't know--it is just some person on the internet saying it!  But if it is, then the problem in gh-21826 appears to be that the compiler on that platform did not populate the global namespace with `isinf` and `isnan` when `cmath` was included.  That implies that when writing C++ code that uses `cmath`, we should always qualify all the math functions with `std::` (or expose them with a `using` statement).

Is that, in fact, what was happening in gh-21826?

Regardless, I'll go ahead and qualify all the standard math functions in `stirling2.h` with `std::`.

steppi:
> That sounds good, but I'd like to understand better just what went wrong.

Me too actually. I get the `cmath` wasn't defining `isinf` and `isnan` in the global namespace on that particular platform, but am not sure why `exp` for instance was being defined. I just put it down to platform-specific weirdness. When I write C++ code for special I try to qualify all standard library functions with `std`, but I've been more lax when reviewing others code. It seems like some codestyle guidelines would be useful.

steppi (approved):
Looks good to me. Thanks @WarrenWeckesser

PR #22385, opened 2025-01-21
DOC: Update documentation on interactive examples and notebooks

melissawm:
#### Reference issue
---

Updates and expands the guide on adding interactive docs and notebooks to the documentation.

Pinging also @agriyakhetarpal, and I will leave a question below.


doc/source/dev/contributor/rendering_documentation.rst
```diff
@@ -197,6 +210,34 @@ with a ``.ipynb`` file.
...
 .. _MyST-NB: https://myst-nb.readthedocs.io/
 
+.. _interactive-docs:
+
+Interactive examples in docstrings
+----------------------------------
+
+Examples within docstrings can be made interactive using the [``jupyterlite-sphinx`` extension](https://jupyterlite-sphinx.readthedocs.io/en/latest/).
+The buttons for converting examples sections into embedded interactive
+notebooks are hidden by default on clean docs builds. To enable interactive
```
melissawm, line 220: This does not seem to be true. I built locally and I do see the buttons, they are just not functional. Should we remove this, or is it an error on jupyterlite-sphinx?

Note that the try_examples.json file is as described below, so theoretically the buttons _should not_ be enabled.

doc/source/dev/contributor/rendering_documentation.rst
```diff
@@ -197,6 +210,34 @@ with a ``.ipynb`` file.
 
 .. _MyST-NB: https://myst-nb.readthedocs.io/
 
+.. _interactive-docs:
+
+Interactive examples in docstrings
+----------------------------------
+
+Examples within docstrings can be made interactive using the [``jupyterlite-sphinx`` extension](https://jupyterlite-sphinx.readthedocs.io/en/latest/).
```
mdhaber, line 218: This isn't rendering.
image

doc/source/dev/contributor/rendering_documentation.rst
```diff
@@ -197,6 +210,34 @@ with a ``.ipynb`` file.
 
 .. _MyST-NB: https://myst-nb.readthedocs.io/
 
+.. _interactive-docs:
+
+Interactive examples in docstrings
```
mdhaber, line 215: Why move this? Do contributors have to think about this unless they're rendering locally?

doc/source/dev/contributor/rendering_documentation.rst
```diff
@@ -188,6 +169,38 @@ preamble similar to the one below, indicating that this is an executable file:
 
 You don't need to edit this preamble, as it is autogenerated.
```
mdhaber, line 170: Clearly this does not need to have a full markdown tutorial, but there are a few points I think might be worth calling out here:
- Best practices for adding images. (E.g. where should the files go?)
- Best practices for adding links to RST documentation from notebook tutorials. Should I just add a regular hyperlink [like this](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.gmean.html#scipy.stats.gmean) or is there a fancy way like RST references / intersphinx links?
- How to link to notebook tutorials from RST documentation.

doc/source/dev/contributor/rendering_documentation.rst
melissawm: OOps, my bad.

doc/source/dev/contributor/rendering_documentation.rst
melissawm: I think having an explanation of these directives is important. Even if just updating the docstrings they are going to bump into the directives and may be confused since they are not yet widely used in other projects etc. But I think it's a good point that I didn't include an explanation of the directive here, so I'll do that. I'm also happy to move it back if you feel it's more appropriate.

mdhaber: Sorry, maybe I am missing something. Can you point to examples of these directives? Do we have to use them manually to make individual examples interactive? I thought orginally the idea was that all docstring examples would be rendered as interactive online (and presumably it is just turned off right now)?

melissawm: I am clarifying with the jupyterlite-sphinx folks and will update here accordingly.

doc/source/dev/contributor/rendering_documentation.rst
melissawm: Let me know if this is reasonable. If you want to go through these instructions to translate a notebook before merging, that could be a good way to catch anything missing from the instructions.

doc/source/dev/contributor/rendering_documentation.rst
```diff
@@ -188,6 +163,72 @@ preamble similar to the one below, indicating that this is an executable file:
...
+- **Formatting:** MyST Markdown supports standard Markdown formatting, such as
+  bold text, italic text, and code blocks. For example, to make text bold, you
+  can use double asterisks: ``**bold text**``. To make text monospace/formatted
+  as code, you can use *single_ backticks: ```code```.
+- **Images:** If your notebook contains images, you can include them in the MyST
+  Markdown file by using the following syntax:
+  ``![alt text](path/to/image.png)``. Images are usually stored in the same
+  folder as the MyST Markdown file, but you can also use relative paths to
+  reference images in other folders. Note that outputs to executed cells should
+  not be included in version control, as they will be automatically generated
```
mdhaber, line 191: How might that happen? I seem to recall a setting that prevented this, and I think in your instructions you mentioned that the output should be cleared before saving the notebook.

doc/source/dev/contributor/rendering_documentation.rst
```diff
@@ -188,6 +163,72 @@ preamble similar to the one below, indicating that this is an executable file:
...
+- **Internal links:** For internal cross-references such as links to SciPy
+  classes or functions, as well as intersphinx links, you can use the following
+  syntax: ``{role}`link text <reference>```, where ``role`` can be ``ref``,
+  ``class``, ``func`` or any other role you would use with rST. For example, to
+  link to the ``scipy.stats.bartlett`` function, use
+  ``{func}`scipy.stats.bartlett```.
+- **Formatting:** MyST Markdown supports standard Markdown formatting, such as
+  bold text, italic text, and code blocks. For example, to make text bold, you
+  can use double asterisks: ``**bold text**``. To make text monospace/formatted
+  as code, you can use *single_ backticks: ```code```.
```
mdhaber, line 185:
```suggestion
  as code, you can use *single* backticks: ```code```.
```

doc/source/dev/contributor/rendering_documentation.rst
```diff
@@ -188,6 +163,72 @@ preamble similar to the one below, indicating that this is an executable file:
...
 
+Links, formatting and images
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When converting a notebook to a MyST Markdown file, you may need to adjust some
+of the content to match the MyST Markdown syntax.
+
+- **External links:** MyST Markdown uses the standard Markdown syntax for links.
+  For example, ``[link text](https://www.example.com)``. If you have a link in a
+  notebook cell, you can convert it to the MyST Markdown format by enclosing the
```
mdhaber, line 174: What format would it start in such that a conversion would be necessary?

doc/source/dev/contributor/rendering_documentation.rst
```diff
@@ -188,6 +163,72 @@ preamble similar to the one below, indicating that this is an executable file:
...
+  link text in square brackets and the URL in parentheses.
+- **Internal links:** For internal cross-references such as links to SciPy
+  classes or functions, as well as intersphinx links, you can use the following
+  syntax: ``{role}`link text <reference>```, where ``role`` can be ``ref``,
+  ``class``, ``func`` or any other role you would use with rST. For example, to
```
mdhaber, line 179:
```suggestion
  ``class``, ``func`` or any other role you would use with reST. For example, to
```

doc/source/dev/contributor/rendering_documentation.rst
mdhaber: Thanks. I think the assumption here should be that the contributor wants all docstring examples to have the button to make it interactive (if that is the SciPy default). They probably don't need to do anything manually to make that happen in the real, rendered documentation, but if there's something they need to do to make it happen locally, that's worth mentioning here.

I think other information, like how to *disable* the button ("To hide the buttons that turn docstring examples into embedded notebooks") might belong in the core developer guide, but I don't know that all contributors need it.

mdhaber (commented):
Thanks for those additions!

doc/source/dev/contributor/rendering_documentation.rst
```diff
@@ -188,6 +163,72 @@ preamble similar to the one below, indicating that this is an executable file:
...
+- **Images:** If your notebook contains images, you can include them in the MyST
+  Markdown file by using the following syntax:
+  ``![alt text](path/to/image.png)``. Images are usually stored in the same
+  folder as the MyST Markdown file, but you can also use relative paths to
+  reference images in other folders. Note that outputs to executed cells should
+  not be included in version control, as they will be automatically generated
+  when the SciPy documentation is built.
+- **Linking to the MyST markdown pages:** MyST markdown supports adding link
+  anchors, which can be used to link to specific pages from other documents.
+  To add an anchor to a page, add ``(anchor_name)=`` to the top of the document.
```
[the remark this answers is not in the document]
melissawm: Yes at any point - good catch, I'll clarify.

doc/source/dev/contributor/rendering_documentation.rst
melissawm: Exactly - maybe I should rephrase. What I mean is you should NOT manually add any of the output images as they will be automatically generated when the docs are built. When you run `python dev.py doc` all notebooks with a jupytext preamble get executed, so the outputs are generated and appended to the notebook.

doc/source/dev/contributor/rendering_documentation.rst
melissawm: It could be just a plain URL that gets highlighted as a link in a notebook.

doc/source/dev/contributor/rendering_documentation.rst
melissawm: My fear with moving some of this documentation to the core devs guide is that is will create discoverability problems - do you think this would be findable by searching the docs?

doc/source/dev/contributor/rendering_documentation.rst
```diff
@@ -188,6 +163,72 @@ preamble similar to the one below, indicating that this is an executable file:
...
+  From other markdown pages, you can link to this anchor using the following
+  syntax: ``{ref}`anchor_name```. From other rST pages, you can link to this
```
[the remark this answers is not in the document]
melissawm: I always get those confused... 😅

doc/source/dev/contributor/rendering_documentation.rst
mdhaber: Who needs that information, though?

doc/source/dev/contributor/rendering_documentation.rst
mdhaber: So in MyST markdown, https://www.example.com will render differently than it does here?

doc/source/dev/contributor/rendering_documentation.rst
melissawm: Ok, that makes sense. I will try to find a place for that in the core devs guide. To be clear this information is also kind of in flux right now as changes are made to jupyterlite-sphinx, so they are not permanent.

doc/source/dev/contributor/rendering_documentation.rst
melissawm: I am not sure these links are automatically converted to a `<a href` element, I should check that.

doc/source/dev/contributor/rendering_documentation.rst
mdhaber: I guess what I'm getting at is whether the user needs to double check whether output is sneaking into the MyST markdown notebooks. It's mentioned twice, suggesting that the user needs to be vigilant, but in my experience with Jupyter Lab, I don't manually need to clear the notebook *or* double check; output of code cells is never saved. If this is because of a particular setting I have or a feature of Jupyter Lab that is uncommon among notebook editors, I can understand the need to mention it. However, if is something that should be transparent to the user if they have followed the rest of the instructions, I would document it differently. (E.g. I would maybe add a warning to the relevant instruction that deviations could result in output being saved to the markdown notebooks.)

melissawm: My fear is if I add an item tellling people how to add images, they will assume they have to add images from the output as well, which is not the case. But I now added that the instructions are for "static images" so maybe that's enough and I can remove that sentence.

mdhaber:
@melissawm I got the impression that we are waiting for some things here, e.g.  https://github.com/scipy/scipy/pull/22385#discussion_r1924436954. Please LMK when you'd like me to review again and possibly merge.

doc/source/dev/contributor/rendering_documentation.rst
```diff
@@ -109,14 +84,10 @@ for more information.
...
-
-     .. code-block:: bash
-
-        python -m http.server --directory doc/build/html
-
-     The documentation pages should then be available at `http://localhost:8000/`.
+   - The SciPy documentation contains interactive examples rendered with the
+     help of the `jupyterlite-sphinx` extension. This functionality may not be
+     enabled for local builds of the documentation, or in docs CI previews. For
+     more details, see :ref:`interactive-docs`.
```
agriyakhetarpal, line 90: As long as both `jupyterlite-sphinx` and a kernel via `jupyterlite-pyodide-kernel` are installed, the interactive docs functionality will always be enabled in local builds and docs CI previews (unless there's something I'm missing?).

doc/source/dev/contributor/rendering_documentation.rst
```diff
@@ -137,10 +108,13 @@ Adding or editing tutorials as Jupyter notebooks
 ------------------------------------------------
 
 Under the ``doc/source/`` folder of the SciPy tree you can find a few
-documents written in MyST-NB_ format. These files are executable, meaning that
-their content is executed when the SciPy documentation is built (locally or on
-CI) and any outputs generated by the execution are rendered in the final HTML
-files.
+documents written in MyST-NB_ markdown format. These files are executable,
```
agriyakhetarpal, line 111: Nitpicking: the name of the format is MyST Markdown, and MyST-NB is the Sphinx extension that allows adding documents bearing this format to websites built with Sphinx. Isn't it?

doc/source/dev/contributor/rendering_documentation.rst
```diff
@@ -188,6 +163,71 @@ preamble similar to the one below, indicating that this is an executable file:
 
 You don't need to edit this preamble, as it is autogenerated.
 
+Links, formatting and images
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When converting a notebook to a MyST Markdown file, you may need to adjust some
```
agriyakhetarpal, line 169: Should we mention at the end of this section that one may use the `rst2myst` utility to convert `.rst` files to MyST Markdown first, before converting them to MyST Markdown notebooks?

doc/source/dev/contributor/rendering_documentation.rst
```diff
@@ -188,6 +163,71 @@ preamble similar to the one below, indicating that this is an executable file:
...
+If you want to enable interactivity for your notebook when it is rendered in
+the SciPy documentation website, you need to add the following snippet at the
+beginning of the markdown file:
+
+.. code-block:: markdown
+
+    ```{eval-rst}
+    .. notebooklite:: <filename>.md
+       :new_tab: True
+    ```
```
agriyakhetarpal, line 211: Here, I think we should 1. link to the `NotebookLite` directive's documentation (in case users want to use other options), and 2. also add a link to the `JupyterLite` directive's documentation (even if we don't recommend using it).

doc/source/dev/contributor/rendering_documentation.rst
```diff
@@ -197,6 +237,40 @@ with a ``.ipynb`` file.
 
 .. _MyST-NB: https://myst-nb.readthedocs.io/
 
+.. _interactive-docs:
+
+Interactive examples in docstrings
+----------------------------------
+
+Examples within docstrings can be made interactive using the `jupyterlite-sphinx extension <https://jupyterlite-sphinx.readthedocs.io/en/latest/>`__,
```
agriyakhetarpal, line 245:
```suggestion
Examples within docstrings can be made interactive using the `jupyterlite-sphinx extension <https://jupyterlite-sphinx.readthedocs.io/en/stable/>`__,
```

Nit: probably linking to the stable version is better

doc/source/dev/contributor/rendering_documentation.rst
```diff
@@ -197,6 +237,40 @@ with a ``.ipynb`` file.
...
+which implements the ``.. try_examples::`` directive. When used in the
+"Examples" section of a docstring, this directive will create a button that
+allows users to open the example in an interactive JupyterLite window, or in a
+new tab as a Jupyter notebook. For SciPy, this directive is automatically added
+to all docstrings with an "Examples" section.
```
agriyakhetarpal, line 250:
```suggestion
new tab as a Jupyter notebook. For SciPy, this directive is automatically added
to all docstrings with an "Examples" section through the
``global_enable_try_examples`` configuration option in ``doc/conf.py``.
```

agriyakhetarpal (commented):
Hi @melissawm – this looks great overall! I just have a bunch of comments.

doc/source/dev/contributor/rendering_documentation.rst
agriyakhetarpal: If the `try_examples.json` config file is present and is set up  like you've described here through the regex pattern, all the examples should get disabled. If the config file is absent, all examples are enabled everywhere. I checked and the config file is present in the `main` branch, and the interactive examples are disabled on both the stable version https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.expm.html#scipy.sparse.linalg.expm and on the devdocs https://scipy.github.io/devdocs/reference/generated/scipy.sparse.linalg.expm.html, so this is working, isn't it?

steppi: @melissawm. Did you mean you see the buttons for the interactive examples in the API reference, or for the notebooks. The config file only impacts the `try_examples` directive and the tutorial markdown notebooks don't use that directive. That could explain things.

melissawm: Yeah - I mean in a clean local docs build, with try_examples.json as 
```
{
    "global_min_height": "400px",
    "ignore_patterns": [".*"]
}
```

I see the buttons. I usually watch the static html files, so my understanding is that the buttons are hidden by JS that doesn't get loaded in the static pages, maybe?

melissawm:
Thanks all- I'll update with these suggestions.

doc/source/dev/contributor/rendering_documentation.rst
melissawm: Ok - I just added linkify-it-py to the requirements list and we don't need to worry about this anymore.

doc/source/dev/contributor/rendering_documentation.rst
melissawm: You are right, I think this is just outdated. Removed

doc/source/dev/contributor/rendering_documentation.rst
melissawm: Fair points! I reworded this.

doc/source/dev/contributor/rendering_documentation.rst
melissawm: Honestly, I've never felt the need to use this tool, but I will add a note for completeness.

doc/source/dev/contributor/rendering_documentation.rst
melissawm: The JupyterLite-sphinx docs are not that large, I think adding a link to the notebooklite directive docs is enough to point folks in the right direction. Wdyt?

doc/source/dev/contributor/rendering_documentation.rst
melissawm: In that case it should also probably be better to point the jupyterlite docs page default to be stable instead of latest

melissawm:
I think this is ready for another look.

mdhaber (commented):
I think the new content looks fine and is helpful. If people run into trouble with it, we can reword as needed. The last thing I'd ask for here is an adjustment to the links from the TOC and to have all the MyST Notebook stuff on its own page.

doc/source/dev/contributor/rendering_documentation.rst
```diff
@@ -161,15 +135,16 @@ If you don't need to keep the ``.ipynb`` file, and want to work with MyST
 Markdown only, follow the steps below.
 
 1. Install the jupytext_ tool, using ``pip install jupytext`` or
-   ``conda install jupytext -c conda-forge``
-2. Clear all outputs from your ``.ipynb`` file
+   ``conda install jupytext -c conda-forge``;
+2. Clear all outputs from your ``.ipynb`` file;
```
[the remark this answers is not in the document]
melissawm: Oh that's right, we can skip this as conversion to markdown format excludes the outputs. Great catch

doc/source/dev/contributor/rendering_documentation.rst
agriyakhetarpal: Sounds good!

doc/source/dev/contributor/rendering_documentation.rst
agriyakhetarpal: Thanks @melissawm, that's a good point. I have admin permissions there, so I modified that in the RTD settings just now. I think the change should reflect within a few hours/days.

agriyakhetarpal:
Thanksz Melissa. Please let me know when this is ready for another look!

mdhaber:
> Thanks, Melissa. Please let me know when this is ready for another look!

I was going to merge, but I'll let you take another look @agriyakhetarpal. I think it's ready for that. There are still some things I  might change in the future, but this is the improvement I was looking for.

doc/source/dev/contributor/adding_notebooks.rst
```diff
@@ -0,0 +1,146 @@
...
+------------------------------------------------
+
+Under the ``doc/source/`` folder of the SciPy tree you can find a few
+documents written in MyST markdown format. The content in these files is
+executed when the SciPy documentation is built (locally or on CI) and any
+outputs generated by the execution are rendered in the final HTML files. In
+addition, they can be made interactive in the SciPy documentation website by
+using the
+`jupyterlite-sphinx <https://jupyterlite-sphinx.readthedocs.io/en/latest/>`__
+extension.
```
agriyakhetarpal, line 13:
```suggestion
using the
`jupyterlite-sphinx <https://jupyterlite-sphinx.readthedocs.io/en/stable/>`__
extension.
```

doc/source/dev/contributor/adding_notebooks.rst
```diff
@@ -0,0 +1,146 @@
+.. _adding-notebooks:
+
+Adding or editing tutorials as Jupyter notebooks
+------------------------------------------------
+
+Under the ``doc/source/`` folder of the SciPy tree you can find a few
+documents written in MyST markdown format. The content in these files is
```
agriyakhetarpal, line 7:
```suggestion
documents written in MyST Markdown format. The content in these files is
```
For consistency with how it is mentioned slightly later in the file and in other places in the docs

doc/source/dev/contributor/adding_notebooks.rst
```diff
@@ -0,0 +1,146 @@
...
+
+If you have a document written in Jupyter notebook format (an ``.ipynb`` file)
+and would like to submit it as part of the SciPy documentation, there are two
+options: you can convert it into a MyST Markdown file, and work with a ``.md``
+file only, or you can pair your ``.ipynb`` file with a ``.md`` file and work
+with both. Note that ``.ipynb`` files *should not* be submitted to the SciPy
+documentation.
+
+For more details, please consult the
+`MyST-NB documentation <https://myst-nb.readthedocs.io/en/latest/authoring/text-notebooks.html>`__.
```
agriyakhetarpal, line 23:
```suggestion
`MyST-NB documentation <https://myst-nb.readthedocs.io/en/stable/authoring/text-notebooks.html>`__.
```

doc/source/dev/contributor/adding_notebooks.rst
```diff
@@ -0,0 +1,146 @@
...
+You can also consult the
+`Pairing tutorial on NumPy Tutorials <https://numpy.org/numpy-tutorials/content/pairing.html>`__
+for more information on MyST-NB, Jupytext and pairing notebooks.
+
+How to convert a ``.ipynb`` file to executable ``.md`` file
```
agriyakhetarpal, line 28:
```suggestion
How to convert a ``.ipynb`` file to a ``.md`` file
```

This is a similar suggestion to my previous one about removing the "executable" word around notebooks

doc/source/dev/contributor/adding_notebooks.rst
```diff
@@ -0,0 +1,146 @@
...
+  can use double asterisks: ``**bold text**``. To make text monospace/formatted
+  as code, you can use *single* backticks: ```code```.
+- **Images:** If your notebook contains static images, you can include them in
+  the MyST Markdown file by using the following syntax:
+  ``![alt text](path/to/image.png)``. Images are usually stored in the same
+  folder as the MyST Markdown file, but you can also use relative paths to
+  reference images in other folders. Note that outputs of executed cells should
+  not be included in version control, as they will be automatically generated
+  when the notebook is executed during the SciPy documentation build.
+- **Linking to the MyST markdown pages:** MyST markdown supports adding link
```
agriyakhetarpal, line 87:
```suggestion
- **Linking to the MyST Markdown pages:** MyST Markdown supports adding link
```

doc/source/dev/contributor/adding_notebooks.rst
```diff
@@ -0,0 +1,146 @@
...
+.. note::
+
+    For simplicity, we don't want certain notebook cells to be visible in the
+    rendered documentation. We can hide them by adding ``"jupyterlite_sphinx_strip"``
+    to the tags section of the corresponding cell metadata. To do that in the
+    markdown file, add the following snippet right before cell you want to hide:
+
+    .. code-block:: markdown
+
+        +++ {"tags": ["jupyterlite_sphinx_strip"]}
```
agriyakhetarpal, line 125: As the Markdown cells are not fenced the same way as other cells, we should mention here that the `+++` notation should come in pairs (otherwise, the entire content up to the following code cell would get hidden).

doc/source/dev/contributor/adding_notebooks.rst
```diff
@@ -0,0 +1,146 @@
...
+`Bartlett's test for equal variances <https://scipy.github.io/devdocs/tutorial/stats/hypothesis_bartlett.html>`__
+(and `its source <https://github.com/scipy/scipy/blob/main/doc/source/tutorial/stats/hypothesis_bartlett.md?plain=1>`__)
+for an example.
+
+.. note::
+
+    For simplicity, we don't want certain notebook cells to be visible in the
+    rendered documentation. We can hide them by adding ``"jupyterlite_sphinx_strip"``
+    to the tags section of the corresponding cell metadata. To do that in the
+    markdown file, add the following snippet right before cell you want to hide:
```
agriyakhetarpal, line 121:
```suggestion
    Markdown file, add the following snippet right before the cell you want to hide:
```

agriyakhetarpal (changes requested):
Thanks for waiting for me, @mdhaber. I have just a bunch of some more small nits. This is looking good to merge already – I don't intend to block it in any way :)

doc/source/dev/contributor/adding_notebooks.rst
```diff
@@ -0,0 +1,146 @@
...
+for more details and options.
+
+Opening MyST Markdown files in the Jupyter Notebook application
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you have the ``jupytext`` tool installed, you can open MyST Markdown ``.md``
+files in the Jupyter Notebook application and execute them, just as you would
+with a ``.ipynb`` file.
+
+Converting ``.rst`` files to MyST Markdown
```
mdhaber, line 137: @agriyakhetarpal I think you mentioned `rst-to-myst`, initially? Would it be OK to add a note about it at the end of the "Links, formatting and images" section, which is mostly about converting to markdown, instead of having its own section?

doc/source/dev/contributor/adding_notebooks.rst
```diff
@@ -0,0 +1,146 @@
...
+outputs generated by the execution are rendered in the final HTML files. In
+addition, they can be made interactive in the SciPy documentation website by
+using the
+`jupyterlite-sphinx <https://jupyterlite-sphinx.readthedocs.io/en/latest/>`__
+extension.
+
+If you have a document written in Jupyter notebook format (an ``.ipynb`` file)
+and would like to submit it as part of the SciPy documentation, there are two
+options: you can convert it into a MyST Markdown file, and work with a ``.md``
+file only, or you can pair your ``.ipynb`` file with a ``.md`` file and work
```
mdhaber, line 18: It might be worth mentioning:

```
In this document, we address conversion to a MyST Markdown file. For more information about MyST-NB, Jupytext and pairing notebooks, you can also consult the `Pairing tutorial on NumPy Tutorials <https://numpy.org/numpy-tutorials/content/pairing.html>`__. For even more information, please consult the
`MyST-NB documentation <https://myst-nb.readthedocs.io/en/latest/authoring/text-notebooks.html>`__.
```

doc/source/dev/contributor/adding_notebooks.rst
```diff
@@ -0,0 +1,146 @@
...
+      name: python3
+   ---
+
+You don't need to edit this preamble, as it is autogenerated.
+
+Links, formatting and images
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When converting a notebook to a MyST Markdown file, you may need to adjust some
+of the content to match the MyST Markdown syntax.
```
mdhaber, line 66: Mentioning these items is useful, but some of these are important mostly because they are different from reST (which is the overwhelming majority of the documentation), not because they need to be adjusted when converting between `ipynb` and `md`, right?

For example, external links and standard formatting won't change, IIUC. And do you know if existing internal links within a notebook are automatically converted to MyST format? Are anchors automatically generated to sections?

It might be worth highlighting that if internal hyperlinks are currently formatted as external links, those should be manually converted to internal links - if that is important. (Is it?) 

Definitely it's helpful to point out how to include images.  That is one thing that I know I had to _convert_ manually after using `jupytext`.

mdhaber (commented):
Since this is already looking at another round of review, I'll go ahead and add the remaining thoughts I had.

doc/source/dev/contributor/adding_notebooks.rst
melissawm: Maybe @agriyakhetarpal can chime in here - my experience with conversion tools is that there were still quite a few cases when I had to adjust for formatting, but more importantly it may be useful to check for custom directives or things like the plot directives, which should not be just translated directly but removed (since plot directive output would be autogenerated with executed code-blocks).

doc/source/dev/contributor/rendering_documentation.rst
agriyakhetarpal: I think the JS should get loaded – as the HTML pages generated by Sphinx contain `<script src="_static/jupyterlite_sphinx.js?v=<some kind of ID here>"></script>`, which will attempt to load the `try_examples.json` config file. @steppi is right, the notebook buttons can't be hidden at the moment, this only impacts the API reference examples.

I don't know if it makes sense to add a global configuration for notebook buttons, as they are added in each notebook and are currently not global.

doc/source/dev/contributor/adding_notebooks.rst
agriyakhetarpal: > my experience with conversion tools is that there were still quite a few cases when I had to adjust for formatting, but more importantly it may be useful to check for custom directives or things like the plot directives

Same here. I've seen that `pandoc` as a tool can handle it quite well, but it doesn't know how to change to MyST Markdown-specific syntax that's not a part of Markdown in general, and `rst2myst` works better, but it still leaves out incompatible ones.

I think we can mention one sentence to address the fact that when making the notebook interactive (which is then described in the following section), some directives might still need to be wrapped with "eval-rst", such as:

````markdown

Some Markdown here

```{eval-rst}
.. some_rest_directive_here::
  :some_option: some_value
```

Some more Markdown here
````

and the notebook should be verified after conversion.

melissawm: What do you think now?

agriyakhetarpal (approved):
Thanks, @melissawm! This looks great now :)

melissawm:
@mdhaber I think this is ready for another look!

mdhaber:
Thanks @melissawm @agriyakhetarpal!

PR #22387, opened 2025-01-22
DOC: Add example to show usage of ``predecessors`` matrix returned from `sparse.csgraph.shortest_path`

### LINKED ISSUE #13914: DOC: sparse.csgraph.shortest_path: predecessors array contains self references
I have a system where row N of the predecessors array returned by `shortest_path(... return_predecessors=True)` contains the number N (not on the diagonal).  From what I understand from the docs, this would mean that the shortest path from N to some other node is to go back to N, which means there is no valid path information.  If this is correct, I think it's a bug in `shortest_path`, and I'll try to come up with a simple example (the connectivity matrix right now is fairly large, about 1100).  If not, maybe I'm misunderstanding the docs, so I'd call it a documentation issue.

The distance matrix is being calculated with `sklearn.neighbors.NearestNeighbors.kneighbors_graph()`, and then symmetrized with `nns += nns.T`.  I suppose the bug could be in a faulty distance matrix, but I'd have hoped that `shortest_path` would notice if there's a problem.

```
dist_mat, pred = shortest_path(nn_matrix, directed=False, return_predecessors=True)
print(sum(pred[918, :] == 918))
```

#### Error message:
I'd expect the above code to print 0, indicating that nowhere in row 918 (of my example, which I haven't included yet because it's so big) is there a self reference as a predecessor (or perhaps 1, on the diagonal), but instead it returns 12, indicating many self references.

```
1.5.4 1.19.4 sys.version_info(major=3, minor=6, micro=8, releaselevel='final', serial=0)
```

czgdp1807:
#### Reference issue

Closes https://github.com/scipy/scipy/issues/13914

In this PR I have added one example to show how to use ``predecessors`` matrix returned from `sparse.csgraph.shortest_path`. This example will help in clearing up how to read the said matrix and use it to re-construct the shortest paths from sources to nodes. https://github.com/scipy/scipy/issues/13914 was opened due to the confusion in the format of ``predecessors`` matrix. The example I have added helps in clearing that up because it uses ``predecessors`` matrix to construct shortest paths (which is what it is meant for).

@bernstei Is this good? 

The documentation of `predecessors` is good enough, the example just adds on it and concretely shows how to read it.

### DIFF
--- a/scipy/sparse/csgraph/_shortest_path.pyx
+++ b/scipy/sparse/csgraph/_shortest_path.pyx
@@ -170,6 +170,60 @@ def shortest_path(csgraph, method='auto',
     >>> predecessors
     array([-9999,     0,     0,     1], dtype=int32)
 
+    >>> graph = [
+    ... [0, 0, 7, 0],
+    ... [0, 0, 8, 5],
+    ... [7, 8, 0, 0],
+    ... [0, 5, 0, 0]
+    ... ]
+    >>> graph = csr_array(graph)
+    >>> print(graph)
+    <Compressed Sparse Row sparse array of dtype 'int64'
+    	with 6 stored elements and shape (4, 4)>
+    	Coords	Values
+    	(0, 2)	7
+    	(1, 2)	8
+    	(1, 3)	5
+    	(2, 0)	7
+    	(2, 1)	8
+    	(3, 1)	5
+
+    >>> sources = [0, 2]
+    >>> dist_matrix, predecessors = shortest_path(csgraph=graph, directed=False, indices=sources, return_predecessors=True)
+    >>> dist_matrix
+    array([[ 0., 15.,  7., 20.],
+           [ 7.,  8.,  0., 13.]])
+    >>> predecessors
+    array([[-9999,     2,     0,     1],
+           [    2,     2, -9999,     1]], dtype=int32)
+
+    >>> shortest_paths = {}
+    >>> for idx in range(len(sources)):
+    ...     for node in range(4):
+    ...         curr_node = node
+    ...         path = []
+    ...         while curr_node != -9999:
+    ...             path = [curr_node] + path
+    ...             curr_node = int(predecessors[idx][curr_node])
+    ...         shortest_paths[(sources[idx], node)] = path
+    ...
+
+    >>> shortest_paths[(0, 3)]
+    [0, 2, 1, 3]
+    >>> path03 = shortest_paths[(0, 3)]
+    >>> sum([graph[path03[0], path03[1]], graph[path03[1], path03[2]], graph[path03[2], path03[3]]])
+    np.int64(20)
+    >>> dist_matrix[0][3]
+    np.float64(20.0)
+
+    >>> shortest_paths[(2, 3)]
+    [2, 1, 3]
+    >>> path23 = shortest_paths[(2, 3)]
+    >>> sum([graph[path23[0], path23[1]], graph[path23[1], path23[2]]])
+    np.int64(13)
+    >>> dist_matrix[1][3]
+    np.float64(13.0)
+
     """
     csgraph = convert_pydata_sparse_to_scipy(csgraph, accept_fv=[0, np.inf, np.nan])
 
--- a/scipy/sparse/csgraph/_shortest_path.pyx
+++ b/scipy/sparse/csgraph/_shortest_path.pyx
@@ -1733,7 +1733,7 @@ cdef void _yen(
             if (
                 total_distance != INFINITY
                 and _yen_is_path_in_candidates(candidate_predecessors,
-                                               shortest_paths_predecessors[k-1], 
+                                               shortest_paths_predecessors[k-1],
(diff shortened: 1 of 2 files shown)


scipy/sparse/csgraph/_shortest_path.pyx
```diff
@@ -168,6 +168,60 @@ def shortest_path(csgraph, method='auto',
...
     array([-9999,     0,     0,     1], dtype=int32)
 
+    >>> graph = [
+    ... [0, 0, 7, 0],
+    ... [0, 0, 8, 5],
+    ... [7, 8, 0, 0],
+    ... [0, 5, 0, 0]
+    ... ]
+    >>> graph = csr_array(graph)
+    >>> print(graph)
```
j-bowhay, line 178: Let's remove this print, the previous example shows what csr array looks like when printed

scipy/sparse/csgraph/_shortest_path.pyx
```diff
@@ -168,6 +168,60 @@ def shortest_path(csgraph, method='auto',
...
+    	(2, 0)	7
+    	(2, 1)	8
+    	(3, 1)	5
+
+    >>> sources = [0, 2]
+    >>> dist_matrix, predecessors = shortest_path(csgraph=graph, directed=False, indices=sources, return_predecessors=True)
+    >>> dist_matrix
+    array([[ 0., 15.,  7., 20.],
+           [ 7.,  8.,  0., 13.]])
+    >>> predecessors
```
j-bowhay, line 194: Up until this point the example is exactly the same as the old one, we could probably just remove the old one for conciseness?

scipy/sparse/csgraph/_shortest_path.pyx
```diff
@@ -168,6 +168,60 @@ def shortest_path(csgraph, method='auto',
     >>> predecessors
     array([-9999,     0,     0,     1], dtype=int32)
 
+    >>> graph = [
```
j-bowhay, line 171: It would be great to start the example with a sentence on what I'll show

scipy/sparse/csgraph/_shortest_path.pyx
```diff
@@ -168,6 +168,60 @@ def shortest_path(csgraph, method='auto',
...
+    >>> sources = [0, 2]
+    >>> dist_matrix, predecessors = shortest_path(csgraph=graph, directed=False, indices=sources, return_predecessors=True)
+    >>> dist_matrix
+    array([[ 0., 15.,  7., 20.],
+           [ 7.,  8.,  0., 13.]])
+    >>> predecessors
+    array([[-9999,     2,     0,     1],
+           [    2,     2, -9999,     1]], dtype=int32)
+
+    >>> shortest_paths = {}
```
j-bowhay, line 198: It would be great to have some text here explaining what this is about to show

scipy/sparse/csgraph/_shortest_path.pyx
```diff
@@ -168,6 +168,60 @@ def shortest_path(csgraph, method='auto',
...
+    ...     for node in range(4):
+    ...         curr_node = node # start from the destination node
+    ...         path = []
+    ...         while curr_node != -9999: # no previous node available, exit the loop
+    ...             path = [curr_node] + path # prefix the previous node obtained from the last iteration
+    ...             curr_node = int(predecessors[idx][curr_node]) # set current node to previous node
+    ...         shortest_paths[(sources[idx], node)] = path
+    ...
+
+    >>> shortest_paths[(0, 3)]
```
j-bowhay, line 209: Again some commentary here would be great

j-bowhay (changes requested):
Thanks @czgdp1807, a few comments. In general it would be good to include more commentary in the example

scipy/sparse/csgraph/_shortest_path.pyx
czgdp1807: > we could probably just remove the old one for conciseness?

Makes sense. Would prefer removing. The new example is better IMO.

jakevdp (commented):
The example here would be more useful with further explanation of what the code is intended to do. I think this may be difficult without an example where the graph has more meaning; for instance there's already one example of using the predecessors array in https://docs.scipy.org/doc/scipy/tutorial/csgraph.html#example-word-ladders Maybe linking to that would be sufficient?

czgdp1807:
> Maybe linking to that would be sufficient?

The link you shared is very nice. Linking it is a nice idea and I would do it. However, in my opinion it's one full page to read. The example I added here is very short. With more commentary it will quickly help the users in understanding the semantics of `predecessors`. Let me know if you still want to just have https://docs.scipy.org/doc/scipy/tutorial/csgraph.html#example-word-ladders linked and remove the example I added.

Meanwhile I wait for your response I will work on linking https://docs.scipy.org/doc/scipy/tutorial/csgraph.html#example-word-ladders. Will remove my example once you say so.l

j-bowhay (approved):

PR #22393, opened 2025-01-22
TST: stats: test support for array API compatible masked arrays

mdhaber:
#### Reference issue
gh-22194

gh-22194 suggested that functions that have been translated to the array API would (almost) "just work" with array API compatible masked arrays such as [`marray`](https://github.com/mdhaber/marray/)s.

This PR gives evidence by testing:
- `stats._stats_py._xp_mean`
- `stats._stats_py._xp_var`
- `stats.gmean`
- `stats.pmean`
- `stats.hmean`
- `stats.moment`
- `stats.skewness`
- `stats.kurtosis`
- `stats.describe`
- `stats.sem`
- `stats.kstat`
- `stats.kstatvar`
- `stats.circmean`
- `stats.circvar`
- `stats.circstd`
- `stats.tmean`
- `stats.tvar`
- `stats.tmin`
- `stats.tmax`
- `stats.tstd`,
- `stats.tsem`
- `stats.zmap`
- `stats.zscore`
- `stats.gzscore`
 
with `marray`s. 

Some of these functions have something like `n = x.shape[axis]`, which I updated to count the number of non-masked elements along axis (for masked arrays only). Other than that, very little needed modification.

~~In a follow-up PR, I can make a generic test that works with lots of functions.~~ Now this works with lots of functions.


scipy/stats/_stats_py.py
```diff
@@ -10824,7 +10824,8 @@ def _xp_mean(x, /, *, axis=None, weights=None, keepdims=False, nan_policy='propa
     if weights is None:
         return xp.mean(x, axis=axis, keepdims=keepdims)
 
-    norm = xp.sum(weights, axis=axis)
+    # ones_like ensures that the mask of `x` is considered
+    norm = xp.sum(xp.ones_like(x) * weights, axis=axis)
```
mdhaber, line 10828: I know that multiplying by an array of ones is wasteful in the non-masked case.

One thought is that during explicit broadcasting of `x` and `weights` (above), we might want to logical OR the two masks. I could add that behavior to `_broadcast_arrays` or even `marray` if we think it's typically the desired thing. That would eliminate the need for the change here.

Alternatively, we can add a special case here for when the input is a masked array; I was just trying to avoid that.

scipy/stats/tests/test_stats.py
```diff
@@ -9443,6 +9464,40 @@ def test_integer(self, xp):
...
         xp_assert_equal(_xp_mean(y, weights=x), _xp_mean(y, weights=y))
 
+    @skip_xp_backends('jax.numpy', reason="JAX doesn't allow item assignment.")
+    @pytest.mark.parametrize('axis', [0, 1])
+    @pytest.mark.parametrize('keepdims', [False, True])
+    def test_marray(self, axis, keepdims, xp):
+        try:
+            import marray
+        except ModuleNotFoundError:
+            pytest.mark.skip('marray not available')
```
mdhaber, line 9474: This probably isn't the preferred way of adding coverage. Suggestions welcome. I don't think it makes sense to add `marray`s to the list of backends that get passed as `xp`, though. It's pretty wasteful to run all tests (with no masks) with `marray`s. I'd think we only really want to use `marray`s in new tests with masks.

mdhaber, line 9474:
```suggestion
            pytest.skip('marray not available')
```

scipy/stats/tests/test_stats.py
```diff
@@ -7087,6 +7087,27 @@ def test_weights_masked_1d_array(self, xp):
...
 
+    #@skip_xp_backends('jax.numpy', reason="JAX doesn't allow item assignment.")
+    @pytest.mark.parametrize('axis', [0, 1])
+    def test_marray(self, axis):
+        xp = np  # can't use array-api-compat xp because data-apis/array-api-compat#226
+        try:
+            import marray
+            mxp = marray._get_namespace(xp)
+        except ModuleNotFoundError:
+            pytest.mark.skip('marray not available')
```
mdhaber, line 7098:
```suggestion
            pytest.skip('marray not available')
```

scipy/stats/tests/test_stats.py
```diff
@@ -7087,6 +7087,27 @@ def test_weights_masked_1d_array(self, xp):
...
                           dtype=np.float64, xp=xp)
 
+    #@skip_xp_backends('jax.numpy', reason="JAX doesn't allow item assignment.")
+    @pytest.mark.parametrize('axis', [0, 1])
+    def test_marray(self, axis):
+        xp = np  # can't use array-api-compat xp because data-apis/array-api-compat#226
+        try:
+            import marray
+            mxp = marray._get_namespace(xp)
+        except ModuleNotFoundError:
```
j-bowhay, line 7097: I think this can be tidied up a little bit with `pytest.importorskip`

mdhaber: Thanks. Didn't know about that!

mdhaber: So with that, you think this pattern is OK? Add new test to exercise masked arrays, and run for all `xp` backends?

scipy/stats/tests/test_stats.py
lucascolley: That sounds right to me.

mdhaber: Continuing this discussion at https://github.com/scipy/scipy/pull/22393#discussion_r1926176277.

mdhaber:
Failure is unrelated (gh-22395).

lucascolley (approved):
overall, this approach looks good to me!

I'm not sure whether it is worth merging this right now though. We might be able to make a more informed decision about https://github.com/scipy/scipy/pull/22393/files#r1925931554 after trying this with some more functions, and it would be nice to be able to test this on other backends once that array-api-compat issue is resolved.

scipy/stats/tests/test_stats.py
j-bowhay: No objections, we can always reconsider if it becomes too much for a burden

mdhaber:
> We might be able to make a more informed decision about https://github.com/scipy/scipy/pull/22393/files#r1925931554 after trying this with some more functions, 

I added tests for several other functions. I'll just keep adding until you tell me it's enough.

> and it would be nice to be able to test this on other backends once that array-api-compat issue is resolved.

Easy to update the tests when that is resolved, though. I doubt that will give us any information about issues in `stats`; it seems more likely that any new failures would be due to issues in `marray`. So maybe that doesn't need to hold things up?

scipy/stats/_stats_py.py
```diff
@@ -1221,11 +1221,19 @@ def _var(x, axis=0, ddof=0, mean=None, xp=None):
...
+        n = _shape_nonmasked(x, axis)
         var *= np.divide(n, n-ddof)  # to avoid error on division by zero
     return var
 
 
+def _shape_nonmasked(x, axis, keepdims=False):
+    if hasattr(x, 'mask'):
+        mxp = array_namespace(x)
+        n = mxp.astype(mxp.count(x, axis=axis, keepdims=keepdims), x.dtype)
+        return n[()] if n.ndim == 0 else n
```
mdhaber, line 1233: We don't need the ternary here.

scipy/stats/_stats_py.py
```diff
@@ -1529,7 +1537,8 @@ def describe(a, axis=0, ddof=1, bias=True, nan_policy='propagate'):
     if xp_size(a) == 0:
         raise ValueError("The input must not be empty.")
 
-    n = a.shape[axis]
+    n = xp.asarray(_shape_nonmasked(a, axis), dtype=xp.int64)
+    n = n[()] if n.ndim == 0 else n
```
mdhaber, line 1541: We can use `xp.astype` and eliminate the ternary once data-apis/array-api-compat#226 is resolved.

scipy/stats/tests/test_marray.py
```diff
@@ -0,0 +1,88 @@
...
+from scipy.stats._stats_py import _xp_mean
+skip_backend = pytest.mark.skip_xp_backends
+
+
+def get_arrays(n_arrays, *, dtype='float64', xp=np, shape=(7, 8), seed=84912165484321):
+    marray = pytest.importorskip('marray')
+    xpm = marray._get_namespace(xp)
+
+    entropy = np.random.SeedSequence(seed).entropy
+    rng = np.random.default_rng(entropy)
```
mdhaber, line 15: Oops. Borrowed from `marray`.
```suggestion
    rng = np.random.default_rng(seed)
```

scipy/stats/tests/test_marray.py
```diff
@@ -0,0 +1,88 @@
...
+    [(stats.moment, {'order': 2}),
+     (stats.skew, {}),
+     (stats.skew, {'bias': False}),
+     (stats.kurtosis, {}),
+     (stats.kurtosis, {'bias': False}),
+     ])
+@pytest.mark.parametrize('axis', [0, 1, None])
+def test_several(fun, kwargs, axis):
+    mxp, marrays, narrays = get_arrays(1)
+    kwargs.update(dict(axis=axis))
```
mdhaber, line 70: I think these dictionaries are only used one time, so should be OK to update? I can avoid it if preferred.

scipy/stats/_stats_py.py
```diff
@@ -1145,13 +1145,12 @@ def _demean(a, mean, axis, *, xp, precision_warning=True):
...
         rel_diff = xp.max(xp.abs(a_zero_mean), axis=axis,
                           keepdims=True) / xp.abs(mean)
+
+    n = _length_nonmasked(a, axis)
     with np.errstate(invalid='ignore'):
-        precision_loss = xp.any(rel_diff < eps)
-    n = (xp_size(a) if axis is None
-         # compact way to deal with axis tuples or ints
-         else np.prod(np.asarray(a.shape)[np.asarray(axis)]))
+        precision_loss = xp.any(xp.asarray((rel_diff < eps)) & xp.asarray(n > 1))
```
mdhaber, line 1151:
```suggestion
        precision_loss = xp.any(xp.asarray(rel_diff < eps) & xp.asarray(n > 1))
```

mdhaber:
I can probably add a few more functions in this PR if desired, but then I'll be at about two dozen, which I think is a fair number for one PR.

Some that I skipped are:
- `variation` which has a few special cases involving `n`. I wonder if we can avoid the special cases; if not, we might have to use `_lazywhere` because `n` is no longer a scalar. I'd rather tackle this in a follow-up.
- I couldn't immediately tell whether `entropy` would be easy or not, so I didn't try. Again, would be better to take a look separately in case it is more challenging than the rest.
- `directional_stats` looks easy, but it doesn't have `nan_policy`, so no easy way to calculate ground truth.
- I stopped when I got to hypothesis tests because they will require evaluating special functions. No problem - it just means i have to modify `special._support_alternative_backends` to accept and return masked arrays. But for thtat reason, I'd rather do all those in a separate PR.

Failure is now with regular masked arrays; looks easy to fix.

mdhaber:
I updated `array-api-compat` so we could get the fix from data-apis/array-api-compat#240, and I added multiple backends to the tests. This did not turn up any bugs in the SciPy code or `marray`, but it did reveal data-apis/array-api-compat#242.

scipy/stats/tests/test_marray.py
```diff
@@ -0,0 +1,121 @@
...
+    xp_assert_close(res.data, xp.asarray(ref))
+
+
+@skip_backend('jax.numpy', reason="JAX doesn't allow item assignment.")
+@skip_backend('torch', reason="array-api-compat#242")
+@pytest.mark.parametrize('axis', [0, 1])
+@pytest.mark.parametrize('kwargs', [{}])
+def test_describe(axis, kwargs, xp):
+    mxp, marrays, narrays = get_arrays(1, xp=xp)
+    kwargs.update(dict(axis=axis))
```
mdhaber, line 100:
```suggestion
    kwargs = dict(axis=axis) | kwargs
```

mdhaber:
I prototyped the changes that would need to be made for hypothesis tests: https://github.com/mdhaber/scipy/compare/marray_gmean...mdhaber:scipy:marray_hypotests?expand=1.

1. `_support_alternative_backends` needs to deal with the masks as mentioned in https://github.com/scipy/scipy/pull/22393#issuecomment-2611422442
2. For paired-sample statistics/hypothesis tests, we need a function that takes the `logical_or` of the masks among the array arguments[^1].We could definitely use it in `_xp_mean` to avoid the need for https://github.com/scipy/scipy/pull/22393#discussion_r1925931554. LMK if you want that to go in here or a separate PR. If here, should I move `_share_masks` and `_length_nonmasked` to `_lib`, maybe `_array_api`?

[^1]: I actually haven't thought of a situation in which we'd be using `broadcast_arrays` and *don't* want to do this, so I'm leaning more toward moving this behavior to `marray`'s `broadcast_arrays` function, but I think I'd wait to pull the trigger on that after testing more hypothesis tests.

mdhaber:
@lucascolley should we skip tests with Dask? Apparently Dask arrays don't have a `device` attribute. 

Anything else to be done here?

lucascolley:
Can you use https://data-apis.org/array-api-compat/helper-functions.html#array_api_compat.device in MArray?

mdhaber:
No, I don't plan to add any dependencies.

scipy/stats/_stats_py.py
```diff
@@ -967,7 +967,8 @@ def tsem(a, limits=None, inclusive=(True, True), axis=0, ddof=1):
         # by the axis_nan_policy decorator shortly.
         sd = _xp_var(a, correction=ddof, axis=axis, nan_policy='omit', xp=xp)**0.5
 
-    n_obs = xp.sum(~xp.isnan(a), axis=axis, dtype=sd.dtype)
+    nan_nan = xp.astype(~xp.isnan(a), a.dtype)
```
lucascolley, line 970: for my understanding, what does "nan_nan" mean here? Is it basically "not_nan"?

scipy/stats/_stats_py.py
lucascolley: any further thoughts here?

scipy/stats/tests/test_marray.py
```diff
@@ -0,0 +1,126 @@
...
+from scipy import stats
+
+from scipy._lib._array_api import xp_assert_close, xp_assert_equal
+from scipy.stats._stats_py import _xp_mean, _xp_var
+skip_backend = pytest.mark.skip_xp_backends
+
+
+def get_arrays(n_arrays, *, dtype='float64', xp=np, shape=(7, 8), seed=84912165484321):
+    marray = pytest.importorskip('marray')
+    xpm = marray._get_namespace(xp)
```
lucascolley, line 12: nit for consistency (here and below)
```suggestion
    mxp = marray._get_namespace(xp)
```

lucascolley (approved):
this looks great to me, thanks Matt! Nice to see this coming together :)

scipy/stats/_stats_py.py
mdhaber: That was probably supposed to be `not_nan`, yeah. I'll change it.

scipy/stats/tests/test_marray.py
mdhaber: Ok. I did find myself going back and forth. I think I prefer `xpm` but found myself doing `mxp` sometimes. You have a preference?

scipy/stats/_stats_py.py
mdhaber: Yeah the follow-up with hypothesis test will change this back. There is some code in there that shares the masks.

scipy/stats/tests/test_marray.py
lucascolley: I have a preference for `mxp`. I think we've discussed this before, but my mental model of library-wrapping like this goes left-to-right as it goes outside-in, so e.g. `marray(dask(numpy))` rather than `((numpy)dask)marray)`.

mdhaber: I don't think one could disagree with the mental model of namespace composition, of course. But that is not the only factor one might want to represent with the reference name. I think it's more important to emphasize `xp`, which means to me that it's standing in for *any* *Py*thon array API compatible backend, than the fact that it's a masked version of the namespace. `m` might as well be a subscript of `xp` to denoting the fact that it's a masked version. (That doesn't mean I'd advocate for `xp_m`, though.)

I think it's similar to:
```python3
import numpy as np
import cupy as cp
x = [1, 2, 3, 4, 5]
x_np = np.asarray(x)
x_cp = cp.asarray(x)
```
One could argue that a NumPy/CuPy array was created from `x` and therefore it should be `np_x`/`cp_x`, but I think the fact that the variable still represents `x` is the important thing and therefore belongs first. Would you advocate for `np_x`/`cp_x`?

Anyway, it's not super important, so I'll change it to `mxp` here. I'm just explaining that my preference isn't arbitrary, either, and I've thought about changing it in the `marray` documentation.

lucascolley: Well, I'm not sure. I think somebody might have used `npm` before, though 😅

mdhaber:
OK, hopefully we're good to go after those updates. I'll be working on hypothesis tests with `marray`s this weekend (e.g. https://github.com/mdhaber/scipy/pull/118), so it would be great to get this in to avoid merge conflicts.

scipy/stats/_stats_py.py
```diff
@@ -1222,11 +1222,23 @@ def _var(x, axis=0, ddof=0, mean=None, xp=None):
...
     var = _moment(x, 2, axis, mean=mean, xp=xp)
     if ddof != 0:
-        n = x.shape[axis] if axis is not None else xp_size(x)
+        n = _length_nonmasked(x, axis)
         var *= np.divide(n, n-ddof)  # to avoid error on division by zero
     return var
 
 
+def _length_nonmasked(x, axis, keepdims=False):
+    if hasattr(x, 'mask') and not isinstance(x, np.ndarray):
```
rgommers, line 1231: This isn't going to be robust, it assumes that no other array type has a `mask` attribute. Can you make it more specific to Marray?

scipy/stats/tests/test_marray.py
```diff
@@ -0,0 +1,126 @@
...
+import numpy as np
+from scipy import stats
+
+from scipy._lib._array_api import xp_assert_close, xp_assert_equal
+from scipy.stats._stats_py import _xp_mean, _xp_var
+skip_backend = pytest.mark.skip_xp_backends
+
+
+def get_arrays(n_arrays, *, dtype='float64', xp=np, shape=(7, 8), seed=84912165484321):
+    marray = pytest.importorskip('marray')
```
rgommers, line 11: Can this be a module-level skip rather than within any test? Faster and easier to understand for readers new to this file.

.github/workflows/array_api.yml
```diff
@@ -74,6 +74,10 @@ jobs:
       run: |
         python -m pip install dask[array]
 
+    - name: Install MArray
+      run: |
+        python -m pip install marray
```
rgommers, line 79: Can the tests for this run in a separate job step? There's only some handfuls of `stats` functions being tested, so folding it in with all other backends seems unnecessary - it'll lead to many more skips for all other submodules I think.

rgommers:
I had a quick browse, it looks pretty good to me overall.

scipy/stats/_stats_py.py
lucascolley: IMO it would be fine to punt on this until we hit a real problem. But perhaps adding `is_marray` to array-api-compat would be an acceptable way forward?

.github/workflows/array_api.yml
lucascolley: I think we should leave it as is. We aren't registering an marray namespace as a new backend value for `xp` to take - rather all tests continue to be parametrised by the same backends, but tests specific to masked arrays test `marray`-wrapped versions of each backend.

Could you clarify where you expect the large amount of skips to come from?

scipy/stats/_stats_py.py
mdhaber: Actually, I made it nonspecific on purpose. I didn't want to assume this little project would deserve special treatment in SciPy. For example, perhaps an project would still want to complete their own array API implementation for performance. It does assume that real array API compatible libraries with a `mask` attribute behave like what we call "masked arrays". I thought that was a fair assumption until there is a counterexample. It looks like its going to be a little while before there are too many array API backends for us to keep track of.

scipy/stats/tests/test_marray.py
mdhaber: Definitely. Somehow I just learned about`importorskip` and didn't see that existed.

mdhaber: If we don't end up changing anything else, I'll do this in the follow-up that adds hypothesis tests.

.github/workflows/array_api.yml
rgommers: Ah I misread that, this should be fine. And with only handfuls of tests, running them with all array API backends seems fine.

rgommers: resolving, because no change needed

scipy/stats/_stats_py.py
rgommers: > It does assume that real array API compatible libraries with a `mask` attribute behave like what we call a "masked array". I thought that was a fair assumption until there is a counterexample.

That shouldn't be hard to construct a counterexample for? A numpy.ndarray subclass with a `.mask` attribute should trigger it. There was also PyTorch NestedTensor that had a mask attribute. There may be more. Other libraries could also support subclassing, and then any user-defined class with a mask attribute will likely do the wrong thing.

I'd say the assumption is the wrong way around - there is nothing that specifies or asks for these semantics, so don't assume that they must be implemented that way. It's fine to check specifically for `marray`.

lucascolley: how could we make this more robust to alternative Masked array implementations going forwards? Could we add a `"masked arrays"` entry to https://data-apis.org/array-api/draft/API_specification/generated/array_api.info.capabilities.html#array_api.info.capabilities?

rgommers: We can't in a reasonable way I think. It should just be specific, at least until there's a known alternative implementation of a masked array library/object in development.

mdhaber: > I'd say the assumption is the wrong way around

Well, we have the same information and come to different conclusions. I didn't really think either conclusion was "wrong", though; I can see it from both perspectives. In fact, originally I wrote an `is_marray` function and used it - and I'm willing to go back. It's just that upon further consideration, I decided to go a different way.

I thought that it was more likely for the semantics of `mask` and `count` to become somewhat standardized before there would be an array library that met all the conditions required for a real user to get an incorrect result without noticing. The latter is not as unlikely as a hash collision, of course, but there are only so many array libraries out there.

Perhaps with the standard that will change. And _if_ we plan to guarantee that SciPy will work seamlessly with any array API compatible library and all possible subclasses, including ones we haven't tested with or even heard of, it is certainly safer to be more specific. So I'll go back to looking for `marray` specifically.

>  There was also PyTorch NestedTensor that had a mask attribute

Perhaps there were, but NestedTensors don't seem to have a `mask` attribute or now, and there is no `count` function.

```python3
import torch
a, b = torch.arange(3), torch.arange(5) + 3
nt = torch.nested.nested_tensor([a, b])
nt.mask  # AttributeError: 'Tensor' object has no attribute 'mask'
torch.count  # AttributeError: module 'torch' has no attribute 'count'
torch.nested.count  # AttributeError: module 'torch.nested' has no attribute 'count'
```
Neither do Torch's masked tensors, incidentally.

> We can't in a reasonable way I think. 

The standard does say:
> The following topics are out of scope: I/O, polynomials, ..., subclassing of an array class, masked arrays, and missing data.

but this experiment makes me think it might be worth revisiting this point sometime. That said, there are many other things I'd prioritize (more general fancy indexing, standardizing *something* about integer (array) to float promotion, `special` extension, etc.).

scipy/stats/tests/test_marray.py
```diff
@@ -0,0 +1,127 @@
+import pytest
+marray = pytest.importorskip('marray')
+
```
mdhaber, line 3:
```suggestion
```

scipy/stats/tests/test_marray.py
```diff
@@ -0,0 +1,127 @@
...
+import numpy as np
+from scipy import stats
+
+from scipy._lib._array_api import xp_assert_close, xp_assert_equal
+from scipy.stats._stats_py import _xp_mean, _xp_var
+skip_backend = pytest.mark.skip_xp_backends
```
mdhaber, line 9:
```suggestion

marray = pytest.importorskip('marray')
skip_backend = pytest.mark.skip_xp_backends
```

scipy/stats/_stats_py.py
mdhaber: Done; now we check for `is_marray`.

rgommers: > Perhaps with the standard that will change. And _if_ we plan to guarantee that SciPy will work seamlessly with any array API compatible library and all possible subclasses [...] but this experiment makes me think it might be worth revisiting this point sometime.

I'd say that while the standard itself is unlikely to start saying anything about subclassing, it is an aim to support subclasses of libraries (e.g., numpy, pytorch) that themselves support it. And that's something I explicitly documented in gh-18286 (see the table of behavior for different array types).

> Done; now we check for `is_marray(xp)`.

Thanks!

> Of course, we could be a little more restrictive 

I think the `is_marray` check is fine. Any string name-based check can be subverted, but one has to really try. You own the name on PyPI after all.

mdhaber:
Explicit approval + implicit approval, so I'll go ahead and merge so I can continue with the hypothesis tests. Thanks!

PR #22396, opened 2025-01-23
DOC: signal: Use where='post' when plotting discrete response #21020

### LINKED ISSUE #21020: DOC: signal: Use `where='post'` when plotting discrete response

By definition a discrete time system `x[n]` is related to the corresponding continuous time system `x(t)` by `x[n] = x(nT)`. Hence, `x[0] = x(0)` in the interval `[0, T)`. So, if `x(0) = 0`, `x[0]` should remain zero for the first time period.

However, when plotting with the step function of pyplot the `pre` option is used by default. As per the [documentation](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.step.html) 

> 'pre': The y value is continued constantly to the left from every x position, i.e. the interval (x[i-1], x[i]] has the value y[i].
> 'post': The y value is continued constantly to the right from every x position, i.e. the interval [x[i], x[i+1]) has the value y[i].

Here is a comparison between the two options.
plot

As you can see, `x` becomes negative at `t = 0` with the `pre` option. It should remain zero until the second sample at `t = 0.1 s`. I've used the transfer function from this [MATLAB](https://www.mathworks.com/help/control/ug/time-response-of-discrete-time-model.html) example.
TimeDomainResponsesofDiscreteTimeModelExample_01

I'm not sure about the entire list of the affected pages. At least [dstep](https://scipy.github.io/devdocs/reference/generated/scipy.signal.dstep.html#dstep) and [dimpulse](https://scipy.github.io/devdocs/reference/generated/scipy.signal.dimpulse.html#dimpulse).

_No response_

_No response_

yahya-darman:
#### Reference issue
Closes #21020.

This pull request updates the documentation for the `dstep` and `dimpulse` functions in the `scipy.signal` module to use `where='post'` when plotting discrete responses. This change ensures that the plots accurately represent the behavior of discrete-time systems, where the value at each sample point should hold until the next sample.

#### Additional information
This change addresses the issue raised in [#21020](https://github.com/scipy/scipy/issues/21020), where the default behavior of using `where='pre'` was found to be incorrect for discrete-time signals.

### DIFF
--- a/scipy/signal/_ltisys.py
+++ b/scipy/signal/_ltisys.py
@@ -3192,7 +3192,7 @@ def dimpulse(system, x0=None, t=None, n=None):
 
     >>> butter = signal.dlti(*signal.butter(3, 0.5))
     >>> t, y = signal.dimpulse(butter, n=25)
-    >>> plt.step(t, np.squeeze(y))
+    >>> plt.step(t, np.squeeze(y), where='post', label='(where="post")')
     >>> plt.grid()
     >>> plt.xlabel('n [samples]')
     >>> plt.ylabel('Amplitude')
@@ -3279,7 +3279,7 @@ def dstep(system, x0=None, t=None, n=None):
 
     >>> butter = signal.dlti(*signal.butter(3, 0.5))
     >>> t, y = signal.dstep(butter, n=25)
-    >>> plt.step(t, np.squeeze(y))
+    >>> plt.step(t, np.squeeze(y), where='post', label='(where="post")')
     >>> plt.grid()
     >>> plt.xlabel('n [samples]')
     >>> plt.ylabel('Amplitude')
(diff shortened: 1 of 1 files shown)


scipy/signal/_ltisys.py
```diff
@@ -3192,7 +3192,7 @@ def dimpulse(system, x0=None, t=None, n=None):
 
     >>> butter = signal.dlti(*signal.butter(3, 0.5))
     >>> t, y = signal.dimpulse(butter, n=25)
-    >>> plt.step(t, np.squeeze(y))
+    >>> plt.step(t, np.squeeze(y), where='post', label='(where="post")')
```
j-bowhay, line 3195: In both cases I'd prefer not to add the new label

j-bowhay (changes requested):
Thanks for working on this

scipy/signal/_ltisys.py
yahya-darman: Thanks! I updated the two lines to remove label='(where="post")'.

j-bowhay (approved):
Thanks! Will merge if CI is green

PR #22399, opened 2025-01-24
TST: remove thread-unsafe skips for a now fixed Cython fused types issue

rgommers:
This was Cython issue 6506, fixed a while back. The Cython nightly wheels and the latest conda-forge package (3.1.0a2.dev0) contain the fix.

[skip ci]

### DIFF
--- a/scipy/linalg/tests/test_basic.py
+++ b/scipy/linalg/tests/test_basic.py
@@ -183,7 +183,6 @@ class TestSolveBanded:
         x = solve_banded((l, u), ab, b)
         assert_array_almost_equal(dot(a, x), b)
 
-    @pytest.mark.thread_unsafe  # due to Cython fused types, see cython#6506
     @pytest.mark.parametrize('dt_ab', [int, float, np.float32, complex, np.complex64])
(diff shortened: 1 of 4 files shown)


ev-br:
Do we need to bump the cython minumum version, too?

rgommers:
> Do we need to bump the cython minumum version, too?

We can't do that before 3.1.0 is released. We require alpha/nightly versions for free-threading support.

tylerjereddy (approved):
As a sniff test I did two trial test runs (`Python 3.13.0` free threaded, ARM mac) on this branch for these adjusted test modules with Cython `3.1.0a1` vs. latest Cython `master` branch (`4f079c8`) with `export PYTHON_GIL=0`, `main` branch of `pytest-run-parallel`, and this specific incantation:

`python dev.py test -t scipy/linalg/tests/test_basic.py -t scipy/ndimage/tests/test_measurements.py -t scipy/signal/tests/test_cont2discrete.py -t scipy/signal/tests/test_signaltools.py -- --parallel-threads=10 --iterations=10`

There were failures in the older Cython version in both trials, and no failures in either of the test trials with latest Cython.

That seems good enough here for now given the rest of the discussion here.

Issue #22382, opened 2025-01-21
ENH: sparse.spmatrix: allow fast import


ricardoV94:
### Is your feature request related to a problem? Please describe.

Trying to import `spmatrix` for dispatch registering triggers a whole load of imports.

Plotting the output of `python -X importtime -c "from scipy.sparse._matrix import spmatrix" 2> import.log && python -m tuna import.log`

![Image](https://github.com/user-attachments/assets/4340b4f8-8324-49d9-a2f4-a79c27f6a0a3)

### Describe the solution you'd like.

Could we have a way to import spmatrix directly without triggering everything in between?

perimosocordiae:
The issue here is that while the `_matrix.py` file is quite trivial, Python will import all parent modules (`scipy.sparse` and `scipy`) automatically.

It may be possible to remove some of the cost by making some imports lazy: https://scientific-python.org/specs/spec-0001/
For example, doing this for the `csgraph` import would save ~18% based on your plot.

lucascolley:
> It may be possible to remove some of the cost by making some imports lazy

If I understand correctly from https://github.com/scipy/scipy/pull/15230/files#r771091534, we currently have lazy loading for the top-level submodules, but not for nested submodules, like `sparse.csgraph` - is that correct?

dschult:
> If I understand correctly from https://github.com/scipy/scipy/pull/15230/files#r771091534, we currently have lazy loading for the top-level submodules, but not for nested submodules, like sparse.csgraph - is that correct?


That is correct. We have lazy loading for the top level submodules. It turns out `sparse.csgraph` gets imported with `sparse`, but `sparse.linalg` does not. The `sparse.linalg` subpackage requires direct importing rather than lazy loading or relying on sparse being imported. I don't know much about other non-top-level-subpackages. But it makes sense to me that `csgraph` and `linalg` both be lazy-loaded when `sparse` is loaded.  I haven't seen any issues appearing with the currently top-level lazy-loading. And it has been a few years since that went in.

It might be a good time to add lazy-loading for lower level subpackages.  At least for `sparse.linalg` and `sparse.csgraph`.

@ricardoV94, the biggest chunk (44%) of import time is from importing `numpy`, which scipy does at the top level to verify that a valid numpy version is available. So we may be able to save you the 0.035s of csgraph, but not the 0.083s of numpy (unless we convince numpy to use lazy-load methods or we implement a way to make numpy load lazily within scipy.) 

[...]

PR #22407, opened 2025-01-26
MAINT: remove `_lib`->`sparse` dependency

lucascolley:
#### Reference issue
x-ref gh-22382

With your config @rgommers:
```
scipy on 🎋 issparse [$!+?] is 📦 v1.16.0.dev0 via 🐍 v3.13.1 via 🐚 scipy:tach took 2s 
❯ tach check --exact
❌: Found unused dependencies: 
	'scipy._lib' does not depend on: ['scipy.sparse']

Remove the unused dependencies from tach.toml, or consider running 'tach sync' to update module configuration and remove all unused dependencies.
```

- new ABC `SparseBase` for sparse arrays/matrices which can live in `_lib`
- `_spbase` is registered as a virtual subclass in `sparse`

#### Additional information
FYI @jorenham not sure if this interacts with the new `sparray` you added at all.

This doesn't address the `fft`->`sparse` dependency: see https://github.com/scipy/scipy/issues/22412

### DIFF
--- a/scipy/sparse/_base.py
+++ b/scipy/sparse/_base.py
@@ -5,7 +5,7 @@ import numpy as np
 from ._sputils import (asmatrix, check_reshape_kwargs, check_shape,
                        get_sum_dtype, isdense, isscalarlike,
                        matrix, validateaxis, getdtype)
-from scipy._lib._sparse import SparseBase, issparse
+from scipy._lib._sparse import SparseABC, issparse
 
 from ._matrix import spmatrix
 
@@ -59,7 +59,11 @@ _ufuncs_with_fixed_point_at_zero = frozenset([
 MAXPRINT = 50
 
 
-class _spbase:
+# `_spbase` is a subclass of `SparseABC`.
+# This allows other submodules to check for instances of sparse subclasses
+# via `scipy._lib._sparse.issparse`, without introducing
+# an import dependency on `scipy.sparse`.
+class _spbase(SparseABC):
     """ This class provides a base class for all sparse arrays.  It
     cannot be instantiated.  Most of the work is provided by subclasses.
     """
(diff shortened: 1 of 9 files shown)


jorenham:
> FYI @jorenham not sure if this interacts with the new `sparray` you added at all.

Shouldn't be a problem

scipy/sparse/_base.py
```diff
@@ -1391,6 +1392,13 @@ def _get_index_dtype(self, arrays=(), maxval=None, check_contents=False):
                                (check_contents and not isinstance(self, sparray)))
 
 
+# `_spbase`` is a "virtual subclass" of `SparseBase`.
+# This allows other submodules to check for sparse subclasses
+# via `scipy._lib._sparse.issparse`, without introducing
+# an import dependency on `scipy.sparse`.
+SparseBase.register(_spbase)
```
perimosocordiae, line 1399: Why register here, instead of regular inheritance?

perimosocordiae (approved):

scipy/sparse/_base.py
lucascolley: I suppose I was being cautious to avoid inadvertently changing `_spbase`, but maybe concrete subclassing would be fine here.

rgommers: Subclassing seems more natural to me as well.

Other than that, this PR looks like a very good idea.

jorenham: I'm also on team "subclass" (`stubtest` probably won't understand it otherwise).

Also, `_spbase` isn't public API, so it seems like a low-risk change to me 🤷🏻

lucascolley:
thanks all!

PR #22418, opened 2025-01-28
BUG: scipy.spatial: Fix inaccurate orthonormalization in `Rotation.from_matrix()`

### LINKED ISSUE #22417: BUG: spatial.transform: `Rotation.from_matrix()` uses inaccurate orthonormalization algorithm
### Describe your issue.

The algorithm for orthonormalizing rotation matrices in `Rotation.from_matrix()` uses the method in the paper 
[Markley, F. Landis. "Unit quaternion from rotation matrix." Journal of guidance, control, and dynamics 31.2 (2008): 440-442.](https://arc.aiaa.org/doi/abs/10.2514/1.31730?journalCode=jgcd). However, the algorithm is only valid for matrices that are already close to orthonormal, and was designed to be used to correct for roundoff error during repeated multiplications. When given inputs that are not already near orthonormal, the result is far off from the correct solution to the [orthogonal Procrustes problem](https://en.wikipedia.org/wiki/Orthogonal_Procrustes_problem).

I believe this is inappropriate to use for a method which is intended to orthonormalize any user matrix input, and should be changed to use an exact solution such as can be obtained with singular value decomposition.

Discovered in this comment thread: https://github.com/scipy/scipy/pull/22267#discussion_r1927740092

The example below is taken from the current docs: https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.from_matrix.html

### Reproducing Code Example

```python
import numpy as np

a = np.array([
    [0, -0.5, 0],
    [0.5, 0, 0],
    [0, 0, 0.5]])

print('Exact SVD method')
U, _, Vt = np.linalg.svd(a)
print(U @ Vt)

print('Inaccurate existing method')
r = R.from_matrix(a)
print(r.as_matrix())
```
Outputs:
```
Exact SVD method
[[ 0. -1.  0.]
 [ 1.  0.  0.]
 [ 0.  0.  1.]]
Inaccurate existing method
[[-0.38461538 -0.92307692  0.        ]
 [ 0.92307692 -0.38461538  0.        ]
 [ 0.          0.          1.        ]]
```

These results are 22.6 degrees apart!

```shell
scipy 1.15.1
```

scottshambaugh:
#### Reference issue
Closes https://github.com/scipy/scipy/issues/22417

The existing algorithm to orthonormalize rotation matrices only works for matrices that are already near orthonormal, and was designed to correct for round-off error. This can lead to large errors when handling arbitrary user inputs - indeed the existing example in the [`Rotation.from_matrix()`](https://docs.scipy.org/doc/scipy-1.15.1/reference/generated/scipy.spatial.transform.Rotation.from_matrix.html) docs shows an extremely poor result which is 22.6 degrees (!!) off the exact solution.

This updates the algorithm to correctly solve the orthogonal Procrustes problem using a singular value decomposition method. The existing algorithm is still used to convert the orthonormal result to a unit quaternion.

This also disallows left-handed rotation matrices, which return even worse results:

```python
import numpy as np

# left-handed rotation matrix
m = np.eye(3)
m[0, 0] = -1
print(m)

r = R.from_matrix(m)
print(r.as_matrix())

"""
[[-1.  0.  0.]
 [ 0.  1.  0.]
 [ 0.  0.  1.]]

[[-1.  0.  0.]
 [ 0.  1.  0.]
 [ 0.  0. -1.]]  # note negative sign
"""
```

#### Additional information
We may want to use Davenport's q method or Shuster's QUEST instead, but these seem less readable and not necessarily faster to me:  https://ahrs.readthedocs.io/en/latest/filters/davenport.html https://ahrs.readthedocs.io/en/latest/filters/quest.html

@nmayorov FYI

### DIFF
--- a/scipy/spatial/transform/_rotation.pyx
+++ b/scipy/spatial/transform/_rotation.pyx
@@ -1008,7 +1008,10 @@ cdef class Rotation:
 
         Rotations in 3 dimensions can be represented with 3 x 3 proper
         orthogonal matrices [1]_. If the input is not proper orthogonal,
-        an approximation is created using the method described in [2]_.
+        an approximation is created by orthonormalizing the input matrix
+        using the method described in [2]_, and then converting the
+        orthonormal rotation matrices to quaternions using the algorithm
+        described in [3]_. Matrices must be right-handed.
 
         Parameters
         ----------
@@ -1025,7 +1028,8 @@ cdef class Rotation:
         References
         ----------
         .. [1] https://en.wikipedia.org/wiki/Rotation_matrix#In_three_dimensions
-        .. [2] F. Landis Markley, "Unit Quaternion from Rotation Matrix",
+        .. [2] https://en.wikipedia.org/wiki/Orthogonal_Procrustes_problem
+        .. [3] F. Landis Markley, "Unit Quaternion from Rotation Matrix",
                Journal of guidance, control, and dynamics vol. 31.2, pp.
                440-442, 2008.
 
@@ -1040,6 +1044,8 @@ cdef class Rotation:
         ... [0, -1, 0],
         ... [1, 0, 0],
         ... [0, 0, 1]])
+        >>> r.single
+        True
         >>> r.as_matrix().shape
         (3, 3)
 
@@ -1058,6 +1064,10 @@ cdef class Rotation:
         ... ]])
         >>> r.as_matrix().shape
         (2, 3, 3)
+        >>> r.single
+        False
+        >>> len(r)
+        2
 
         If input matrices are not special orthogonal (orthogonal with
         determinant equal to +1), then a special orthogonal estimate is stored:
@@ -1067,15 +1077,15 @@ cdef class Rotation:
         ... [0.5, 0, 0],
         ... [0, 0, 0.5]])
         >>> np.linalg.det(a)
-        0.12500000000000003
+        0.125
         >>> r = R.from_matrix(a)
         >>> matrix = r.as_matrix()
         >>> matrix
-        array([[-0.38461538, -0.92307692,  0.        ],
-               [ 0.92307692, -0.38461538,  0.        ],
-               [ 0.        ,  0.        ,  1.        ]])
+        array([[ 0., -1.,  0.],
+               [ 1.,  0.,  0.],
+               [ 0.,  0.,  1.]])
         >>> np.linalg.det(matrix)
-        1.0000000000000002
+        1.0
 
         It is also possible to have a stack containing a single rotation:
 
@@ -1097,7 +1107,7 @@ cdef class Rotation:
         .. versionadded:: 1.4.0
         """
         is_single = False
-        matrix = np.asarray(matrix, dtype=float)
+        matrix = np.array(matrix, dtype=float)
 
         if (matrix.ndim not in [2, 3] or
             matrix.shape[len(matrix.shape)-2:] != (3, 3)):
@@ -1107,13 +1117,35 @@ cdef class Rotation:
         # If a single matrix is given, convert it to 3D 1 x 3 x 3 matrix but
         # set self._single to True so that we can return appropriate objects in
         # the `to_...` methods
-        cdef double[:, :, :] cmatrix
         if matrix.shape == (3, 3):
-            cmatrix = matrix[None, :, :]
+            matrix = matrix[np.newaxis, :, :]
             is_single = True
-        else:
-            cmatrix = matrix
 
+        # Calculate the determinant of the rotation matrix
+        # (should be positive for right-handed rotations)
+        dets = np.linalg.det(matrix)
+        if np.any(dets <= 0):
+            ind = np.where(dets <= 0)[0][0]
+            raise ValueError("Non-positive determinant (left-handed or null "
+                             f"coordinate frame) in rotation matrix {ind}: "
+                             f"{matrix[ind]}.")
+
+        # Gramian orthonormality check
+        # (should be the identity matrix for orthonormal matrices)
+        grams = matrix @ np.swapaxes(matrix, 1, 2)
+        orthonormal_ok = np.all(np.isclose(grams, np.eye(3), atol=1e-12), axis=(1, 2))
+        idxs = np.where(~orthonormal_ok | ~np.isclose(dets, 1, atol=1e-12))[0]
+
+        # Orthonormalize the rotation matrices where necessary
+        if len(idxs) > 0:
+            # Exact solution to the orthogonal Procrustes problem using SVD
+            U, _, Vt = np.linalg.svd(matrix[idxs, :, :])
+            matrix[idxs, :, :] = U @ Vt
+
+        # Convert the orthonormal rotation matrices to quaternions using the
+        # algorithm described in [3]_.
+        cdef double[:, :, :] cmatrix
+        cmatrix = matrix
         cdef Py_ssize_t num_rotations = cmatrix.shape[0]
         cdef Py_ssize_t i, j, k
         cdef double[:] decision = _empty1(4)
--- a/scipy/spatial/transform/_rotation.pyx
+++ b/scipy/spatial/transform/_rotation.pyx
@@ -1006,12 +1006,12 @@ cdef class Rotation:
     def from_matrix(cls, matrix):
         """Initialize from rotation matrix.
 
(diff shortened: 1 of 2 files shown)


scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -320,6 +325,37 @@ def test_from_matrix_ortho_output():
...
 
+def test_from_matrix_normalize():
+    mat = np.array([
+        [1, 1, 0],
+        [0, 1, 0],
+        [0, 0, 1]])
+    expected = np.array([[ 0.894427, 0.447214, 0.0],
+                         [-0.447214, 0.894427, 0.0],
+                         [ 0.0,      0.0,      1.0]])
+    assert_array_almost_equal(Rotation.from_matrix(mat).as_matrix(), expected)
```
j-bowhay, line 336: Nit, new tests should use `assert_allclose`

scipy/spatial/transform/_rotation.pyx
```diff
@@ -1107,13 +1117,35 @@ cdef class Rotation:
...
+            ind = np.where(dets <= 0)[0][0]
+            raise ValueError("Non-positive determinant (left-handed or null "
+                             f"coordinate frame) in rotation matrix {ind}: "
+                             f"{matrix[ind]}.")
+
+        # Gramian orthonormality check
+        # (should be the identity matrix for orthonormal matrices)
+        grams = matrix @ np.swapaxes(matrix, 1, 2)
+        orthonormal_ok = np.all(np.isclose(grams, np.eye(3), atol=1e-12), axis=(1, 2))
+        idxs = np.where(~orthonormal_ok | ~np.isclose(dets, 1, atol=1e-12))[0]
```
nmayorov, line 1137: What if we omit this block and will do orthogonalization regardless?

nmayorov (approved):
I agree with your findings and that SVD orthgonalization makes more sense in general.

Let's merge it before `RigidTransformation` pull request.

scipy/spatial/transform/_rotation.pyx
scottshambaugh: I think it would be quite a bit slower for rotation matrices that are already well formed, as the expectation is that SVD is expensive relative to calculating these diagnostics. 

For the matrices where SVD is skipped, they will still be orthogonalized in the conversion to quaternion, and should be close enough that the original method is valid.

scottshambaugh: Sounds good, I'll switch it over once this bugfix is in main.

scipy/spatial/transform/_rotation.pyx
```diff
@@ -1107,13 +1117,35 @@ cdef class Rotation:
...
+        dets = np.linalg.det(matrix)
+        if np.any(dets <= 0):
+            ind = np.where(dets <= 0)[0][0]
+            raise ValueError("Non-positive determinant (left-handed or null "
+                             f"coordinate frame) in rotation matrix {ind}: "
+                             f"{matrix[ind]}.")
+
+        # Gramian orthonormality check
+        # (should be the identity matrix for orthonormal matrices)
+        grams = matrix @ np.swapaxes(matrix, 1, 2)
```
nmayorov, line 1135: `grams` to something more readable? Or simply move into `np.isclose`.

scipy/spatial/transform/_rotation.pyx
nmayorov, line 1137: Rename into `bad_indices` or something like that.

scipy/spatial/transform/_rotation.pyx
scottshambaugh: `gramian` ok? https://en.wikipedia.org/wiki/Gram_matrix

scipy/spatial/transform/_rotation.pyx
scottshambaugh: It turns out the determinant check here is redundant, since matrices which pass the gramian check will have `det in {-1, 1}`, and we ruled out the negative case above. Removed that bit

scipy/spatial/transform/_rotation.pyx
```diff
@@ -1107,13 +1117,40 @@ cdef class Rotation:
...
+        if np.any(dets <= 0):
+            ind = np.where(dets <= 0)[0][0]
+            raise ValueError("Non-positive determinant (left-handed or null "
+                             f"coordinate frame) in rotation matrix {ind}: "
+                             f"{matrix[ind]}.")
+
+        # Gramian orthonormality check
+        # (should be the identity matrix for orthonormal matrices)
+        # Note that we have already ruled out left-handed cases above
+        gramians = matrix @ np.swapaxes(matrix, 1, 2)
```
nmayorov, line 1136: (minor) As far as I know `np.transpose` can also swap required axes, perhaps it will communicate the idea better.

scipy/spatial/transform/_rotation.pyx
```diff
@@ -1107,13 +1117,40 @@ cdef class Rotation:
...
+        is_orthonormal = np.all(np.isclose(gramians, np.eye(3), atol=1e-12),
+                                axis=(1, 2))
```
nmayorov, line 1138: (minor) I think "orthogonal matrix" is more standard and well understood term, I don't know exact reasons, but "orthonormal" is rarely used as far as I'm aware.

scottshambaugh: Orthogonal doesn't capture the unit norm quality. I could switch to unitary?

ilayn: We can use unitary here as we use Hermitian for symmetric real anyways. Unitary is also easier to link in your head to something being 1.0 though you can't immediately remember what it would be, slightly easier to type too. But not a blocker, I don't have strong opinions.

nmayorov: > Orthogonal doesn't capture the unit norm quality.

Again I don't known deep reasons, but SO3 matrices are almost always called orthogonal (from my experience).

WarrenWeckesser: > Orthogonal doesn't capture the unit norm quality. 

FYI: The definition of *orthogonal matrix* includes orthonormality.

scottshambaugh: > The definition of orthogonal matrix includes orthonormality.

Whelp, I looked it up and looks like I wasn't remembering correctly. -2 linear algebra points for me :) 

Updated the wording since orthogonal is a more common term.

ilayn: It should be the other way around. `a = np.array([[1, 2], [-2, -1]])` is only orthogonal but not **ortho**gonal and **normal**ized. I know this is a convention and definition issue but I know many fields that orthogonal/orthonormal difference is a real thing.

ilayn: [Orthogonal matrix](https://en.wikipedia.org/wiki/Orthogonal_matrix) entry does not make a distinction but [Unitary Matrix](https://en.wikipedia.org/wiki/Unitary_matrix#Equivalent_conditions) does and mentions orthonormality. So I guess this is one of those "it depends who you talk to" issues.

scottshambaugh: I'm also finding some sources that call any matrix with perpendicular basis vectors an orthogonal matrix. But others saying that the unit norm is an additional constraint. I don't think it matters too much, the user input is normalized and the code should be clear for anyone making future modifications.

ilayn: Indeed, and also in the rotation context it is pretty much given that rotation matrices are unitary, or should be.

scottshambaugh:
I tagged this as 1.15.2 but will leave it to the maintainers on whether that or 1.16.0 is more appropriate.

ilayn:
Cython code looks good to me. We can tweak it here and there but with diminishing returns hence let's tackle if it becomes an issue in the future.

nmayorov:
I have merged it. Let's get back to RigidTransformation.

nh2:
Thanks for this, this original bug caused me multiple days of figuring out why my algorithm was slow, until I found that my rotations were all messed up and `scipy` hadn't pointed this out; it took me a long time to find it. Had I had this PR, I would have noticed immediately. Good improvement!

PR #21674, opened 2024-10-07
DEV: use `spin`

czgdp1807:
#### Reference issue

Closes gh-16452
Closes gh-18046
Closes gh-18951
Closes gh-18998

#### What does this implement/fix?

Linked PR from `spin` - https://github.com/scientific-python/spin/pull/256
~https://github.com/numpy/numpy/pull/27606~

Regarding the context, we are moving away from `dev.py` and shifting to `spin` for building, testing SciPy. Right now it's a work in progress. I have completed `spin build`, `spin test` and `spin docs`. Apologies for the delay in providing background for this PR. Its in a good shape now. Testing on CI still remains to be done.

Sub commands check list,

- [x] `build`
- [x] `test`
- [x] `lint`
- [x] `mypy`
- [x] `shell`
- [x] `python`
- [x] `ipython`
- [x] `docs`
- [x] `refguide-check`
- [x] `smoke-docs`
- [x] `bench`
- [x] `smoke-tutorials`

What needs to be done to merge,

- [x] Implement as many of the sub-commands without touching `spin`. Only reviews will be addressed in https://github.com/scientific-python/spin/pull/256 from now onwards.
- [x] Test everything on CI here once https://github.com/scientific-python/spin/pull/256 is in and a new release is available.
- [x] Address all the reviews and questions.
- [x] Address https://github.com/scientific-python/spin/pull/256#issuecomment-2505857378
- [x] Take care of https://github.com/scipy/scipy/issues/16452
- [x] Drop the workaround in https://github.com/scipy/scipy/pull/21674/commits/c1cab968f94156e2e8a006ca724bc81409412565

#### Additional information

cc: @melissawm @rgommers 

Steps to reproduce the change here and in https://github.com/scientific-python/spin/pull/238,

1. Clone `spin` locally into your system. `git clone https://github.com/scientific-python/spin`.
2. Change your working directory to `spin` (cloned above).
3. Activate your `scipy` environment. For example I use `scipy-dev` from `conda`. So I do, `conda activate scipy-dev`.
4. Fetch my branch from https://github.com/scientific-python/spin/pull/256. I basically do, `git remote add czgdp1807 https://github.com/czgdp1807/spin`. `get fetch czgdp1807 meson_kwargs`. `czgdp1807` is the remote pointing to my fork.
5. Checkout to `meson_kwargs` branch (just fetched above).
6. Do an editable install of `spin`. I do, `pip install -e .`.
7. Now switch to SciPy directory in your system.
[...]

PR #22419, opened 2025-01-28
MAINT: special: Remove `libsf_error_state` shared library in favor of coordinated update of separate static libraries

steppi:
#### Reference issue

I think this PR will likely close some issues or unblock the closing of others, but I'm not sure which at the moment.

Following up on https://github.com/scipy/scipy/pull/21956#issuecomment-2608025506, this PR replaces the use of a shared library for managing special function error handling state between extension modules, with simpler code which just exposes a function `_set_action` in each extension module separately, which each get called in turn when the error state is being updated from Python using `seterr` or `errstate`. This was actually already happening for the extension modules `ufuncs` and `ufuncs_cxx`. See:

https://github.com/scipy/scipy/blob/c1c3dfc83f27a2750c8e86bde50b62abb773efe4/scipy/special/_ufuncs_extra_code.pxi#L163-L164

I've just updated things so it now happens for each of the extension modules. All of the additional complexity which was introduced just to support this shared library has also been removed.

It was an interesting experiment to try to ship a shared library in SciPy to allow sharing state between extension modules, but it's caused too many build trouble to be worth it.

#### Additional information

`libsf_error_state` will be gone but not forgotten. I wrote a blog post about it which became a bit of a cautionary tale: https://labs.quansight.org/blog/libsf-error-state.


rgommers:
> Following up on [#21956 (comment)](https://github.com/scipy/scipy/pull/21956#issuecomment-2608025506),

That's only one of a handful of recent issues. Let me list them to document the rationale here clearly. The shared library is currently working well in wheels and downstream packaging, however it turns up new blockers:

1. We don't have a good way of making it thread-safe under free-threading on Windows, because MinGW-w64 does not support thread local variables (see [gh-21956](gh-21956)).
2. It was reported that on a Kitware Paraview build pipeline, use of rpath's is forbidden by default as a policy and hence the SciPy build on macOS was failing (see [gh-22270#comment](https://github.com/scipy/scipy/pull/22270#issuecomment-2575747846)).
3. The migration from `dev.py` to `spin` turned up a new problem in `spin`, and fixing that problem is delayed because of shared library handling (see [gh-21674](gh-21674) and [spin#257](https://github.com/scientific-python/spin/pull/257)).
4. My PR for building OpenBLAS from source as part of the SciPy build is blocked on an issue in meson-python with rpath handling (see [gh-22047](gh-22047) and [meson-python#771](https://github.com/mesonbuild/meson-python/issues/711)).

I'd summarize these as: things can and should work in principle, but since having an internal shared library is so unusual for Python packages, we are running into a lot of unexpected bugs and corner cases.

rgommers:
From a quick look: things should still be made thread-safe with thread-local variables it looks like. I'm fine with leaving that to a follow-up PR, since it's probably less mechanical than this change. What do you prefer?

j-bowhay:
> The `special/tests/test_mpmath.py::TestSystematic::test_besselj_complex` failure on the pre-release job is possibly real, worth looking into.

 #22395

steppi:
> From a quick look: things should still be made thread-safe with thread-local variables it looks like. I'm fine with leaving that to a follow-up PR, since it's probably less mechanical than this change. What do you prefer?

Thanks Ralf. I'd prefer to leave it to a follow-up.

rgommers:
Rebased to get a fresh test run.

rgommers (commented):
This is looking pretty good, and CI is happy. Just one comment: I think the `install_rpath` usages in `meson.build` can be cleaned up. Can you confirm that and update @steppi?

steppi:
> This is looking pretty good, and CI is happy. Just one comment: I think the `install_rpath` usages in `meson.build` can be cleaned up. Can you confirm that and update @steppi?

Thanks @rgommers! Good catch.

rgommers (approved):
All happy, in it goes. Thanks Albert!

drew-parsons:
postscriptum to this PR:

I patched special/meson.build in the debian scipy package to add an ABI to libsf_error_state,
 see https://salsa.debian.org/python-team/packages/scipy/-/blob/8b4e7a06ebbb86487d2452b1cf0322ba6006e691/debian/patches/special_libsf_error_SOABI.patch
 It helped with cross-platform handling of the shared library.
 
With this PR removing  libsf_error_state, I'll be able to remove this debian patch for scipy 1.16.0

rgommers:
Thanks for sharing @drew-parsons, it's really helpful to have a complete overview of all the sharp edges of using a shared library for whomever tries this next.

PR #22421, opened 2025-01-28
BUG: ndimage: `binary_erosion` vs. broadcasted input

crusaderky:
binary_erosion and binary_dilation return incorrect results when the input is a broadcasted zero-dimensional array of ones.

### DIFF
--- a/scipy/ndimage/tests/test_morphology.py
+++ b/scipy/ndimage/tests/test_morphology.py
@@ -782,6 +782,15 @@ class TestNdimageMorphology:
         assert_array_almost_equal(out, xp.asarray([0, 1, 0]))
 
     @pytest.mark.parametrize('dtype', types)
+    @xfail_xp_backends("cupy", reason="https://github.com/cupy/cupy/issues/8912")
+    def test_binary_erosion05_broadcasted(self, dtype, xp):
+        dtype = getattr(xp, dtype)
+        data = xp.ones((1, ), dtype=dtype)
+        data = xp.broadcast_to(data, (3, ))
+        out = ndimage.binary_erosion(data)
+        assert_array_almost_equal(out, xp.asarray([0, 1, 0]))
+
+    @pytest.mark.parametrize('dtype', types)
     def test_binary_erosion06(self, dtype, xp):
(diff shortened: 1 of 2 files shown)


scipy/ndimage/tests/test_morphology.py
```diff
@@ -1501,6 +1510,14 @@ def test_binary_dilation05(self, dtype, xp):
         out = ndimage.binary_dilation(data)
         assert_array_almost_equal(out, xp.asarray([1, 1, 1]))
 
+    @pytest.mark.parametrize('dtype', types)
+    def test_binary_dilation05_broadcasted(self, dtype, xp):
+        dtype = getattr(xp, dtype)
+        data = xp.ones((), dtype=dtype)
```
crusaderky, line 1516:
```suggestion
        data = xp.ones((1, ), dtype=dtype)
```

crusaderky:
Test failure is unrelated.

lucascolley (approved):
LGTM, thanks!

rgommers (commented):
Thanks for addressing this @crusaderky! The tests LGTM, however using `order='C'` will make a full copy for Fortran-ordered input when that doesn't look necessary. The problem is specific to strides with a 0 element it looks like:

```python
>>> x = np.ones((1,), dtype=np.float64)
>>> y = np.broadcast_to(x, (3,))
>>> y.strides
(0,)
>>> y.flags.c_contiguous
False
>>> y.flags.f_contiguous
False
```

A nicer fix may be something like this:
```python
input = np.asarray(input)
if not (input.flags.c_contiguous or input.flags.f_contiguous):
    # The Cython code can't cope with inputs with strides containing 0
    input = np.asarray(input, order='C')
```

For completeness, showing there is a copy:
```python
>>> x = np.arange(6).reshape((2,3), order='F')
>>> y = np.asarray(x, order='C')
>>> y.data is x.data
False
>>> np.may_share_memory(x, y)
False
>>> x.strides
(8, 16)
>>> y.strides
(24, 8)
```

rgommers (approved):
LGTM now, in it goes. Thanks @crusaderky

PR #22422, opened 2025-01-28
MAINT: `_lib`: adapt `array_namespace` to accept scalars with arrays

crusaderky:
Fix bug, that was introduced by scipy's override of `array_namespace`, where a mixture of Array API objects and array-likes - including scalars - in a function's parameters would raise.

### DIFF
--- a/scipy/_lib/_array_api.py
+++ b/scipy/_lib/_array_api.py
@@ -8,7 +8,7 @@ https://data-apis.org/array-api/latest/use_cases.html#use-case-scipy
 """
 import os
 
-from collections.abc import Generator
+from collections.abc import Generator, Iterable, Iterator
 from contextlib import contextmanager
 from contextvars import ContextVar
 from types import ModuleType
@@ -58,8 +58,8 @@ Array: TypeAlias = Any  # To be changed to a Protocol later (see array-api#589)
 ArrayLike: TypeAlias = Array | npt.ArrayLike
 
 
-def _compliance_scipy(arrays):
-    """Raise exceptions on known-bad subclasses.
+def _compliance_scipy(arrays: Iterable[ArrayLike]) -> Iterator[Array]:
+    """Raise exceptions on known-bad subclasses and discard ArrayLike objects.
 
     The following subclasses are not supported and raise and error:
     - `numpy.ma.MaskedArray`
@@ -68,8 +68,9 @@ def _compliance_scipy(arrays):
     - Any array-like which is neither array API compatible nor coercible by NumPy
     - Any array-like which is coerced by NumPy to an unsupported dtype
     """
-    for i in range(len(arrays)):
-        array = arrays[i]
+    for array in arrays:
+        if array is None:
+            continue
 
         # this comes from `_util._asarray_validated`
         if issparse(array):
@@ -80,14 +81,19 @@ def _compliance_scipy(arrays):
 
         if isinstance(array, np.ma.MaskedArray):
             raise TypeError("Inputs of type `numpy.ma.MaskedArray` are not supported.")
-        elif isinstance(array, np.matrix):
+
+        if isinstance(array, np.matrix):
             raise TypeError("Inputs of type `numpy.matrix` are not supported.")
+
         if isinstance(array, np.ndarray | np.generic):
             dtype = array.dtype
             if not (np.issubdtype(dtype, np.number) or np.issubdtype(dtype, np.bool_)):
                 raise TypeError(f"An argument has dtype `{dtype!r}`; "
                                 f"only boolean and numerical dtypes are supported.")
-        elif not is_array_api_obj(array):
+
+        if is_array_api_obj(array):
+            yield array
+        else:
             try:
                 array = np.asanyarray(array)
             except TypeError:
@@ -100,8 +106,8 @@ def _compliance_scipy(arrays):
                     f"only boolean and numerical dtypes are supported."
                 )
                 raise TypeError(message)
-            arrays[i] = array
-    return arrays
+            # Discard array. If there are no actual arrays in the input, the namespace
+            # will default to NumPy.
 
 
 def _check_finite(array: Array, xp: ModuleType) -> None:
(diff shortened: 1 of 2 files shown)


rgommers:
All CI was passing, so I doubt there was a bug. `array_namespace(x, [1, 2]) is xp` is not supposed to work by design unless `x` is a numpy array.

crusaderky:
> All CI was passing, so I doubt there was a bug. `array_namespace(x, [1, 2]) is xp` is not supposed to work by design unless `x` is a numpy array.

I'm flexible about how `array_namespace(x, [1, 2])` should behave - that's a special case introduced by scipy which is not in CI.
The problem here is that `array_namespace(x, 1)` raises when it is overridden by scipy, and that breaks array-api-extra which also uses the overridden version.

crusaderky:
Zooming out to the version in array-api-compat: why is `array_namespace(x, 1)` meant to work while `array_namespace(x, [1, 2])` isn't? What's different?

rgommers:
For more context:
- Array-likes (lists, tuples, generators, custom objects, etc.) are pretty much an anti-feature specific to NumPy, and are not supported by most other array libraries. They expect only instances of their own array types.
- Scalars are similar, but a bit separate - there are usability benefits to allowing that, and the array API standard is introducing support for scalars in many functions (constraint: there has to be >=1 array input). We have not yet discussed SciPy-specific changes that we may want to make as a result of that.





> The problem here is that `array_namespace(x, 1)` raises when it is overridden by scipy, and that breaks array-api-extra which also uses the overridden version.

Can you elaborate on when that is needed? I think the design is for it to raise.

crusaderky:
> I think the design is for it to raise.

Nope, it is explicitly allowed:
https://github.com/data-apis/array-api-compat/blob/fbf66e0f668d2aa0c91d3fb41730ccac93ab9eb7/array_api_compat/common/_helpers.py#L577-L578

> Can you elaborate on when that is needed?

The problem is that I have functionality in array-api-extra that is all green in array-api-extra CI, which fails once you vendor array-api-extra and array-api-compat in scipy, because scipy alters the behaviour of `array_namespace` for everyone.

It is not "needed" - I can change the code in array-api-extra to explicitly wrap scalars - but unless array-api-compat becomes as strict as scipy (which in turn would break everything that doesn't vendor it or pin it), the need to explicitly wrap all function parameters is going to remain untested and untestable.

ev-br:
cross-ref https://github.com/data-apis/array-api-compat/pull/147  for a previous discussion.

rgommers:
So it looks like we should decide what to do with scalars here. I think it's probably an improvement to allow _Python_ scalars only (not NumPy scalars), and reject everything else. As long as there is >=1 array instance. And that is also what `array-api-compat`/`array-api-extra` do?

crusaderky:
Updated PR to allow scalars and disallow non-scalars

ev-br:
> So it looks like we should decide what to do with scalars here. I think it's probably an improvement to allow Python scalars only (not NumPy scalars), and reject everything else. As long as there is >=1 array instance. And that is also what array-api-compat/array-api-extra do?

Ideally, `array_namespace` in scipy is the same as the one in `-compat`. Historically the difference was IIUC to make better error messages for weird inputs (masked arrays, `scipy.sparse` arrays etc).
So how about we keep the relation: filter out explicitly blacklisted objects and forward the rest to `array_api_compat.array_namespace`

crusaderky:
> Ideally, `array_namespace` in scipy is the same as the one in `-compat`. Historically the difference was IIUC to make better error messages for weird inputs (masked arrays, `scipy.sparse` arrays etc). So how about we keep the relation: filter out explicitly blacklisted objects and forward the rest to `array_api_compat.array_namespace`

That would break UI when the users pass lists etc.

lucascolley:
indeed, the biggest customisation of array_namespace is defaulting to NumPy for array-likes. That is here to stay for a long time, I imagine.

lucascolley (approved):
Thanks @crusaderky, LGTM!

lucascolley (commented):
one mypy error

```
💻  mypy.api.run --config-file /home/runner/work/scipy/scipy/mypy.ini scipy
scipy/_lib/_array_api.py:154: error: Incompatible types in assignment (expression has type "list[Any]", variable has type "tuple[Any, ...]")  [assignment]
```

rgommers:
`-compat` does not allow lists and other array-like input, right? The SciPy behavior should be a superset of the `-compat` behavior.

lucascolley:
> The SciPy behavior should be a superset of the -compat behavior.

I'm not sure exactly what you mean by this, but is more relaxed in the sense of giving NumPy back instead of raising for array-likes, while rejecting known bad classes instead of trying to coerce to NumPy.

rgommers:
Yes that is what I mean - more relaxed. It should be possible to pre-filter for some things, and then defer to `-compat` for the rest. There should not be any conflicting behavior AFAIK.

crusaderky:
> one mypy error
> 
> ```
> 💻  mypy.api.run --config-file /home/runner/work/scipy/scipy/mypy.ini scipy
> scipy/_lib/_array_api.py:154: error: Incompatible types in assignment (expression has type "list[Any]", variable has type "tuple[Any, ...]")  [assignment]
> ```

should be fixed now

lucascolley:
> I don't see if coercion happens, and if so then where.

https://github.com/scipy/scipy/blob/b396430f57b17c16bd280ce518ed1ac3adf6aaf2/scipy/_lib/_array_api.py#L98-L99

> If there's a function `def func(x1, x2)` and `x2` is a scalar, how does it get converted to a 0-D array of the same array type as `x1`?

- Scalars are coerced to 0-D NumPy arrays on the aforementioned line.
- These 0-D arrays are then dropped (not yielded) here ⬇️

https://github.com/scipy/scipy/blob/b396430f57b17c16bd280ce518ed1ac3adf6aaf2/scipy/_lib/_array_api.py#L110-L114

- Finally, with only `x1` yielded, it is passed to array-api-compat to determine the namespace here ⬇️

https://github.com/scipy/scipy/blob/b396430f57b17c16bd280ce518ed1ac3adf6aaf2/scipy/_lib/_array_api.py#L158-L159

The idea is that we drop the scalars, rather than passing them through to array-api-compat.

rgommers:
> Scalars are coerced to 0-D NumPy arrays on the aforementioned line.

That sounds wrong, they should be coerced with `xp.asarray` to the correct array type.

lucascolley:
> > Scalars are coerced to 0-D NumPy arrays on the aforementioned line.
>
> That sounds wrong, they should be coerced with xp.asarray to the correct array type.

Why is that? Remember that `array_namespace` does not return any arrays, it returns a namespace. Coercing scalars to any namespace seems irrelevant for determining the return namespace.

Unless I am missing something?

crusaderky:
> That sounds wrong, they should be coerced with `xp.asarray` to the correct array type.

the only reason we're coercing them to numpy is to run health checks on them, e.g. disallow object arrays. They never feed into `array_api_compat.array_namespace`.

rgommers:
Oh wait, you are right - we are still doing the manual coercion afterwards with `_asarray`, e.g.:

https://github.com/scipy/scipy/blob/b396430f57b17c16bd280ce518ed1ac3adf6aaf2/scipy/cluster/hierarchy.py#L2393-L2395

Issue #22202, opened 2024-12-27
BUG: special/xsf: Unchecked use of `new` in a few places.


WarrenWeckesser:
### Describe your issue.

The following are unchecked uses of the C++ `new` operator that should be fixed:

https://github.com/scipy/scipy/blob/2b8f9466f5f7862e9c24c5e51793eb5b26ba2dec/scipy/special/xsf/numpy.h#L879

https://github.com/scipy/scipy/blob/2b8f9466f5f7862e9c24c5e51793eb5b26ba2dec/scipy/special/xsf/numpy.h#L902-L904

https://github.com/scipy/scipy/blob/2b8f9466f5f7862e9c24c5e51793eb5b26ba2dec/scipy/special/xsf/amos/amos.h#L2453

The `new` operation can raise the exception `bad_alloc`.  If the exception is not caught, the Python interpreter will crash.

### Reproducing Code Example

```python
n/a
```

### Error message

```shell
n/a
```

### SciPy/NumPy/Python version and system information

```shell
n/a
```

j-bowhay:
Closing as I have open an issue upstream

PR #22423, opened 2025-01-28
BUG: special: Fix a memory leak in the AMOS function besy().

### LINKED ISSUE #22314: BUG: special: Potential memory leak  in the function `besy()` in `amos.h`
The function [`besy()`](https://github.com/scipy/scipy/blob/f2e38dbe5698320b1a9acbfc6174b3cd928ef6f5/scipy/special/xsf/amos/amos.h#L2294) in [`amos.h`](https://github.com/scipy/scipy/blob/main/scipy/special/xsf/amos/amos.h) allocates an array using `new` at line 2453:

https://github.com/scipy/scipy/blob/f2e38dbe5698320b1a9acbfc6174b3cd928ef6f5/scipy/special/xsf/amos/amos.h#L2453

After the array is allocated, there are several returns that can occur without freeing that array:

https://github.com/scipy/scipy/blob/f2e38dbe5698320b1a9acbfc6174b3cd928ef6f5/scipy/special/xsf/amos/amos.h#L2460-L2480

If any of those return statements are executed, the memory pointed to by `cwrk` is lost.

WarrenWeckesser:
Use a unique_ptr to manage the heap-allocated array.

Closes gh-22314.

This also addresses one of the problems reported in gh-22202.

### DIFF
--- a/scipy/special/xsf/amos.h
+++ b/scipy/special/xsf/amos.h
@@ -5,9 +5,12 @@
 
 namespace xsf {
 
+//
+// Return sf_error equivalents for AMOS ierr values.
+// 'ierr' refers to the last parameter of the AMOS functions
+// airy(), besh(), besi(), besj(), besk(), besy(), and biry().
+//
 inline sf_error_t ierr_to_sferr(int nz, int ierr) {
-    /* Return sf_error equivalents for amos ierr values */
-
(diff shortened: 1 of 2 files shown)


steppi (approved):
This looks good to me. Thanks @WarrenWeckesser!

PR #11788, opened 2020-04-04
MAINT: fix domain edge error for betainc a or b ==0 case is 1.

rlucas7:
#### Reference issue
Closes gh-8411

#### What does this implement/fix?
the `betainc` function was not providing correct values at boundary values for `a` and `b`, when either are `0.0`. In the described boundary case the values returned should be `1.0`. This PR implements the change. Also adds a test module to confirm the boundary values are as expected. 

#### Additional information
Previously (taken from gh-8411):
```
import numpy as np
from scipy.special import betainc

X = 0.5
Z = np.array(range(0,11))
W = 3

print(betainc(Z,W,X))
[       nan 0.875      0.6875     0.5        0.34375    0.2265625
 0.14453125 0.08984375 0.0546875  0.03271484 0.01928711]
```

Now: 
```
(scipy-dev) Lucass-MacBook:scipy rlucas$ python 
Python 3.6.7 |Anaconda, Inc.| (default, Oct 23 2018, 14:01:38) 
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy as np
>>> from scipy.special import betainc
>>> X = 0.5
>>> Z = np.array(range(0,11))
>>> W = 3
>>> print(betainc(Z,W,X))
[1.         0.875      0.6875     0.5        0.34375    0.2265625
 0.14453125 0.08984375 0.0546875  0.03271484 0.01928711]
>>> 
(scipy-dev) Lucass-MacBook:scipy rlucas$
```

PR #22425, opened 2025-01-28
MAINT: special: Update handling of `betainc` and `betaincc` edge cases

### LINKED ISSUE #8411: nan with betainc for a=0, b=3 and x=0.5
`scipy.special.betainc(a, b, x) ` returns `nan` when `a=0`, `b=3`, and `x=0.5`.

### Reproducing code example:
```
# Script ported from Matlab documentation:
# https://www.mathworks.com/help/matlab/ref/betainc.html

import numpy as np
from scipy.special import betainc

X = 0.5
Z = np.array(range(0,11))
W = 3

print(betainc(Z,W,X))
```

### Output:
```
[       nan 0.875      0.6875     0.5        0.34375    0.2265625
 0.14453125 0.08984375 0.0546875  0.03271484 0.01928711]
```

### Scipy/Numpy/Python version information:
```
1.0.0 1.14.0 sys.version_info(major=3, minor=5, micro=4, releaselevel='final', serial=0)
```

EDIT: edited for clarity

steppi:
#### Reference issue

Closes #8411
Supercedes #11788

This PR updates the edge case behavior for the functions `betainc` and `betaincc`, particularly for the cases where `a = 0` or `b = 0`. I left a comment here, https://github.com/scipy/scipy/pull/11788#discussion_r1932747665, explaining my rationale for the choices made here. I'm starting a new PR because `betainc` and `betaincc` were updated to use Boost Math for their scalar kernels in time since  #11788 last had activity.

#### Additional information

cc @mdhaber


scipy/special/_add_newdocs.py
```diff
@@ -792,6 +792,21 @@ def add_newdoc(name, doc):
     function by multiplying the result of ``betainc(a, b, x)`` by
     ``beta(a, b)``.
 
+    ``betainc(a, b, x)`` is treated as a two parameter family of functions
+    of a single variable `x`, rather than as a function of three variables.
+    This impacts only the limiting cases ``a = 0``, ``b = 0``, ``a = inf``,
+    ``b = inf``.
```
mdhaber, line 798: Want to add `inf` cases to the tests?

scipy/special/_add_newdocs.py
```diff
@@ -792,6 +792,21 @@ def add_newdoc(name, doc):
...
+
+    In general
+
+    .. math::
+
+        \lim_{(a, b) \rightarrow (a_0, b_0)} \mathrm{betainc}(a, b, x)
+
+    is treated as a pointwise limit in ``x``. Thus for example,
+    ``betainc(0, b, 0)`` equals ``0`` for ``b > 0``, although it should be
```
mdhaber, line 807:
```suggestion
    ``betainc(0, b, 0)`` equals ``0`` for ``b > 0``, although it would be
```
Maybe to avoid the appearance of suggesting that the behavior *should* be otherwise?

scipy/special/_add_newdocs.py
```diff
@@ -883,6 +898,11 @@ def add_newdoc(name, doc):
     -----
     .. versionadded:: 1.11.0
 
+    Like ``betainc``, ``betaincc(a, b, x)`` is treated as a two parameter
+    family of functions of a single variable `x`, rather than as a function of
+    three variables. See the ``betainc`` docstring for more info on how this
```
mdhaber, line 903: Maybe if we want to link?
```suggestion
    Like `betainc`, ``betaincc(a, b, x)`` is treated as a two parameter
    family of functions of a single variable `x`, rather than as a function of
    three variables. See the `betainc` docstring for more info on how this
```

scipy/special/boost_special_functions.h
```diff
@@ -120,10 +153,43 @@ Real ibetac_wrap(Real a, Real b, Real x)
...
+
+    if ((a < 0) || (b < 0) || (x < 0) || (x > 1)) {
         sf_error("betaincc", SF_ERROR_DOMAIN, NULL);
         return NAN;
     }
+
+    /* In limiting cases, SciPy treats `betaincc` as a two parameter family
+     * of functions of a single variable `x`, rather than as a function of
+     * three variables `a`, `b`, `x`. The limit ``(a, b) -> (a0, b0)`` of
+     * ``betainc(a, b, x)`` is treated as the pointwise limit in `x`.
```
mdhaber, line 165:
```suggestion
     * ``betaincc(a, b, x)`` is treated as the pointwise limit in `x`.
```

mdhaber (approved):
Thanks @steppi!

scipy/special/_add_newdocs.py
steppi: Yes, that's an oversight. Good suggestion.

scipy/special/_add_newdocs.py
steppi: Good point.

PR #22433, opened 2025-01-29
BUG: interpolate.RectBivariateSpline: fix `NaN` output when smoothing large arrays

### LINKED ISSUE #17787: BUG: Erratic results from RectBivariateSpline when smoothing large arrays
### Describe your issue.

According to the documentation, RectBivariateSpline can be used for smoothing by setting `s` to a non zero value. This works on small grids, but on larger grids, using any `s` value other than zero produces a spline that evaluates to NaN everywhere. To be clear, this occurs on data that do not contain any NaN. The problem occurs without any warning or error message. The problem occurs irregularly (running the code sometimes work, but most of the time it does not).

The problem occurs on grid with dtype int as well as float.

The problem occurs on Windows 10 and mac OS.

Since the problem does not alway occur, I suspect it may be due to a failure to allocate memory. If enough memory is available, everything works fine. However, it the memory allocation fails, RectBivariateSpline returns a nonsensical result without informing the user about the memory error.

### Reproducing Code Example

```python
import numpy as np
from scipy import interpolate

s = 1  # any value greater than zero causes the problem
nx, ny = 1000, 1700    # Set to larger values to reproduce the problem more consistently
img = np.random.randint(0, 100, size=(nx, ny))
x = np.arange(nx)
y = np.arange(ny)
spl = interpolate.RectBivariateSpline(x, y, img, s=s)
print(np.isnan(spl(x, y)).any())
```

### Error message

```shell
No error message
```

1.9.3 1.23.5 sys.version_info(major=3, minor=10, micro=8, releaselevel='final', serial=0)

1.10.0 1.24.1 sys.version_info(major=3, minor=10, micro=8, releaselevel='final', serial=0)

czgdp1807:
#### Reference issue

Closes https://github.com/scipy/scipy/issues/17787

The fix contains two parts,

1. When knots are maximised in a direction then recomputing of knots as per interpolation in that direction is done.

This is implemented in b278464a85c1aa8e69fc54b5029ea3c1c790b878 and the test is added in b93ba4085af8d096fec17ba94b85a52128ede9d3. The tests check for two things, a) the output is non-NaN, and b) the output is within a tolerance value (for larger `s`, the tolerance increases, which is expected from my point of view).

2. Increasing the maximal iterations for large arrays - In case of large arrays, for some values of `s`, we need higher iterations to converge to an acceptable solution.

This is implemented by exposing `maxit` parameter as optional keyword argument and then passing higher values for `maxit`. This way `maxit` becomes configurable and users can increase it according to their needs. It's better than hardcoding `maxit` to pass the tests I added.

1dfcd52e1113db19546378720d5fd1ed598cf712 exposes the `maxit` as optional keyword argument. 708519d3ca1f7eb8b3ce2dbdbd79b75d72e3c34f updates the documentation for `maxit` and the error message is updated to suggest increasing `maxit`. 48dc2e746c14ead1a6a04e07933bea366e0fdfc2 adds test to make sure the fix works (the pattern for this test is same as above).

### DIFF
--- a/scipy/interpolate/fitpack/fpregr.f
+++ b/scipy/interpolate/fitpack/fpregr.f
@@ -21,6 +21,8 @@ c  ..local scalars
 c  ..function references..
       real*8 abs,fprati
       integer max0,min0
+      logical onlyxknots
+      onlyxknots = .false.
 c  ..subroutine references..
 c    fpgrre,fpknot
 c  ..
@@ -88,6 +90,10 @@ c  test whether the required storage space exceeds the available one.
       if(ny.gt.nyest .or. nx.gt.nxest) go to 420
 c  find the position of the interior knots in case of interpolation.
 c  the knots in the x-direction.
+      if(onlyxknots) then
+        onlyxknots = .false.
+        goto 120
+      end if
   10  mk1 = mx-kx1
       if(mk1.eq.0) go to 60
       k3 = kx/2
@@ -267,7 +273,10 @@ c  add a new knot in the x-direction
           call fpknot(x,mx,tx,nx,fpintx,nrdatx,nrintx,nxest,1)
 c  test whether we cannot further increase the number of knots in the
 c  x-direction.
-          if(nx.eq.nmaxx) go to 10
+          if(nx.eq.nmaxx) then
+            onlyxknots = .true.
+            go to 10
+          end if
           if(nx.eq.nxe) go to 250
  220    continue
         go to 250
--- a/scipy/interpolate/fitpack/fpregr.f
+++ b/scipy/interpolate/fitpack/fpregr.f
@@ -90,10 +90,6 @@ c  test whether the required storage space exceeds the available one.
       if(ny.gt.nyest .or. nx.gt.nxest) go to 420
 c  find the position of the interior knots in case of interpolation.
 c  the knots in the x-direction.
-      if(onlyxknots) then
-        onlyxknots = .false.
-        goto 120
-      end if
   10  mk1 = mx-kx1
       if(mk1.eq.0) go to 60
       k3 = kx/2
@@ -111,6 +107,10 @@ c  the knots in the x-direction.
         i = i+1
         j = j+1
   50  continue
+      if(onlyxknots) then
+        onlyxknots = .false.
+        goto 120
+      end if
 c  the knots in the y-direction.
   60  mk1 = my-ky1
       if(mk1.eq.0) go to 120
@@ -376,4 +376,3 @@ c  error codes and messages.
       fp = 0.
  440  return
       end
-
--- a/scipy/interpolate/fitpack/fpregr.f
+++ b/scipy/interpolate/fitpack/fpregr.f
@@ -88,7 +88,7 @@ c  test whether the required storage space exceeds the available one.
       if(ny.gt.nyest .or. nx.gt.nxest) go to 420
 c  find the position of the interior knots in case of interpolation.
 c  the knots in the x-direction.
-      mk1 = mx-kx1
+  10  mk1 = mx-kx1
       if(mk1.eq.0) go to 60
       k3 = kx/2
       i = kx1+1
@@ -267,6 +267,7 @@ c  add a new knot in the x-direction
           call fpknot(x,mx,tx,nx,fpintx,nrdatx,nrintx,nxest,1)
 c  test whether we cannot further increase the number of knots in the
 c  x-direction.
+          if(nx.eq.nmaxx) go to 10
           if(nx.eq.nxe) go to 250
  220    continue
         go to 250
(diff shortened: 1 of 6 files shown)


czgdp1807:
@ev-br Please let me know how we can make this PR better. It's always fun working with you. Looking forward to your reviews.

scipy/interpolate/fitpack/fpregr.f
```diff
@@ -267,6 +267,7 @@ recursive subroutine fpregr(iopt,x,mx,y,my,z,mz,xb,xe,yb,ye,
           call fpknot(x,mx,tx,nx,fpintx,nrdatx,nrintx,nxest,1)
 c  test whether we cannot further increase the number of knots in the
 c  x-direction.
+          if(nx.eq.nmaxx) go to 10
```
ev-br, line 270: So this line is: if we have nmaxx knots, we need to make sure that the x-knots, tx, satisfies the not-a-knot boundary condition. But the y-knots, ty, can be whatever they are.
However IIUC  the `goto 10` makes both x- and y- knots to be not-a-knot.

So I believe you need to make sure that `tx` is indeed not-a-knot and `ty` is what it was at line 269.

scipy/interpolate/tests/test_rgi.py
```diff
@@ -792,6 +792,38 @@ def test_spline_2d_outofbounds(self):
...
+    def test_spline_large_2d_maxit(self):
+        nx, ny = 1000, 1700
+        s, atol, rtol = 2, 2e-2, 1e-12
+        x, y, z = self._sample_large_2d_data(nx, ny)
+
+        spl = RectBivariateSpline(x, y, z, s=s, maxit=25)
+        z_spl = spl(x, y)
+        assert_(not np.isnan(z_spl).any())
+        xp_assert_close(z_spl, z, atol=atol, rtol=rtol)
+
```
ev-br, line 826: 1. Please add a code comment with a reference to https://github.com/scipy/scipy/issues/17787 where this test originates.

2. Please move this test to `test_fitpack2.py`: let's not mix RGI tests with Fitpack tests.

ev-br (approved):
Looks close! This is a thorny issue which likely was there forever (tm) i.e. from the original Fortran IV implementation from mid-1970s.

IIUC, the main issue you've found and you're fixing is this:

- the algorithm reaches the state where `len(tx) == m + kx + 1` and `len(ty) < m + ky + 1`; 
- no more x knots can be added, and the machinery proceeds to computing the coefficients of the spline
- the `tx` knots are "wrong": they do not guarantee the boundary condition and the linear algebra problem for the coefficients becomes rank-deficient.

What needs to happen is that the `tx` array is corrected to be essentially this: https://github.com/scipy/scipy/blob/main/scipy/interpolate/_bsplines.py#L1013
and _`ty` knots need to stay what they were_.

Again IIUC, your change correctly adjusts `tx` but also adjusts `ty` where it shouldn't.

ev-br (changes requested):
Sorry, wrong button: "request changes" not "approve", not just yet.

scipy/interpolate/tests/test_rgi.py
```diff
@@ -792,6 +792,38 @@ def test_spline_2d_outofbounds(self):
...
+                                        (1, 7e-3, 1e-4),
+                                        (3, 2e-2, 1e-4)])
+    def test_spline_large_2d(self, shape, s_tols):
+        nx, ny = shape
+        s, atol, rtol = s_tols
+        x, y, z = self._sample_large_2d_data(nx, ny)
+
+        spl = RectBivariateSpline(x, y, z, s=s)
+        z_spl = spl(x, y)
+        assert_(not np.isnan(z_spl).any())
```
j-bowhay, line 814: Nit since there are other changes to be made, please use plain `assert` for new tests

scipy/interpolate/fitpack/fpregr.f
czgdp1807: Makes sense. 100%. Let me first address this review and then other ones.

czgdp1807: Made the changes. Please verify the logic. Basically we just note that if we are coming to `10` label from `if(nx.eq.nmaxx) go to 10` (by checking `onlyxknots`) and skip `ty` re-computation in this case.

ev-br:
cc @loicseguin ,  @nepomnyi : could you test your workloads with this PR?

scipy/interpolate/fitpack/fpregr.f
ev-br: Please double-check the `onlyxknots` logic. After `goto 10`, what is there to avoid recomputing `ty`?

ev-br: looks better now, thanks

ev-br:
One of CI timeouts is relevant. Looks like the "large" test needs to be decorated with `@slow`

ev-br (approved):

ev-br:
LGTM on a re-read and the CI is green. Let's land this, thank you @czgdp1807 !

PR #22435, opened 2025-01-29
MAINT: bump to array-api-extra git tip

crusaderky:
- Move xp_lazy_function to array-api-extra
- All functions tagged by `xp_lazy_function` are now tested against accidental Dask graph materializations
- `xpx.at()` now supports bool masks in Dask and JAX as long as the RHS value is a scalar
- New function `xpx.isclose()`
- New function `xpx.nunique()`


scipy/cluster/hierarchy.py
```diff
@@ -2256,7 +2257,7 @@ def is_valid_linkage(Z, warning=False, throw=False, name=None):
         ):
             raise ValueError(f'Linkage {name_str}uses non-singleton cluster before'
                              ' it is formed.')
-        if xp.unique_values(Z[:, :2]).shape[0] < n * 2:
+        if xpx.nunique(Z[:, :2]) < n * 2:
```
crusaderky, line 2260: This helps with nan-shaped Dask arrays but doesn't otherwise prevent the issue of having to materialize the graph.
Will iterate more on it.

scipy/cluster/hierarchy.py
```diff
@@ -4132,7 +4133,7 @@ def leaders(Z, T):
     if T.shape[0] != Z.shape[0] + 1:
         raise ValueError('Mismatch: len(T)!=Z.shape[0] + 1.')
 
-    n_clusters = int(np.asarray(xp.unique_values(T)).shape[0])
+    n_clusters = int(xpx.nunique(T))
```
crusaderky, line 4136: Same here

scipy/optimize/tests/test_chandrupatla.py
```diff
@@ -482,7 +483,7 @@ def test_bracket_order(self, xp):
         loc = xp.linspace(-1, 1, 6)[:, xp.newaxis]
         brackets = xp.asarray(list(permutations([-5, 0, 5]))).T
         res = _chandrupatla_minimize(self.f, *brackets, args=(loc,))
-        assert xp.all(xp.isclose(res.x, loc) | (res.fun == self.f(loc, loc)))
+        assert xp.all(xpx.isclose(res.x, loc) | (res.fun == self.f(loc, loc)))
```
crusaderky, line 486: This accidentally worked before because the test is skipping array-api-strict for unrelated reasons

lucascolley (approved):
thanks @crusaderky !

PR #22439, opened 2025-01-29
MAINT: special: Add `log1mexp` for ``log(1 - exp(x))``

### LINKED ISSUE #19021: ENH: special: add log1mexp function?

I have $\log(x)$, $x \in [0, 1]$. I want to know $\log(1 - \exp(x))$. 

Computed directly, we lose precision if $\exp(x)$ is very close to either 0 or 1, in which case $1 - \exp(x)$ or $\exp(x)$ can be indistinguishable from 1 in finite precision.

I don't think this is a one-liner for all $\exp(x) \in [0, 1]$. `logsumexp` doesn't help because it's not a problem with overflow or underflow.

See https://github.com/scipy/scipy/pull/18138#issuecomment-1470958845

```python3
def log1mexp(x):
    r"""Compute the log of the complement of the exponential

    This function is equivalent to::

        log1mexp(x) = np.log(1-np.exp(x))

    but avoids loss of precision when ``np.exp(x)`` is nearly 0 or 1.

    Parameters
    ----------
    x : array_like
        Input array.

    Returns
    -------
    y : ndarray
        An array of the same shape as `x`.

    Examples
    --------
    >>> import numpy as np
    >>> from scipy.special import log1mexp
    >>> x = 1e-300  # log of a number very close to 1
    >>> log1mexp(x)  # log of the complement of a number very close to 1
    -690.7755278982137
    >>> # p.log(1 - np.exp(x))  # -inf; emits warning

    """
    def f1(x):
        # good for exp(x) close to 0
        return np.log1p(-np.exp(x))

    def f2(x):
        # good for exp(x) close to 1
        return np.real(np.log(-special.expm1(x + 0j)))

    x = _asarray_validated(x, check_finite=False)
    # additional input validation (e.g. input real in [0, 1])
    return _lazywhere(x < -1, (x,), f=f1, f2=f2)[()]
```

Not adding it. This might be too specialized. I've only considered it for $x$ in the domain $[0, 1]$, so I'm not sure whether this implementation is accurate outside of that range. It's fine to make it a private function in `stats` (it allows accurate calculation of `logsf` from accurate `logcdf` and vice versa), but I thought it might be useful to someone else.

Comparison of this solution, direct evaluation, and use of the logsumexp trick against `mpmath` reference.

```python3
import numpy as np
from mpmath import mp
from scipy.special import log1mexp, logsumexp
mp.dps = 1000

# generate logs of numbers very close to 0 and very close to 1
x = np.concatenate((np.linspace(-690, -1), -10**np.linspace(-1, -300)))
xmp = np.asarray([mp.mpf(xi) for xi in x])

# Compute log(1-exp(x))
y = log1mexp(x)  # proposed solution
y_naive = np.log(1 - np.exp(x))  # naive expression
y_lse = logsumexp(np.vstack((np.zeros(len(x)), x)),  # logsumexp trick
                  b=[[1], [-1]], axis=0)

# Compute reference with mpmath
@np.vectorize
def log1mexp_mp(x):
    return mp.log(mp.one - mp.exp(x))
y_mp = log1mexp_mp(x).astype(np.float64)

# Check relative errors
rtol = np.abs((y - y_mp)/y_mp)
np.max(np.abs(rtol))  # 1.8398691820785418e-16; all are ~as accurate as possible

rtol_naive = np.abs((y_naive - y_mp)/y_mp)
np.sum(rtol_naive >= 1) / len(x)  # 94% have lost all precision

rtol_lse = np.abs((y_lse - y_mp)/y_mp)
np.sum(rtol_lse >= 1) / len(x)  # same as direct evaluation
```

steppi:
#### Reference issue

Closes #19021 

This adds a private ufunc with a numerically sound implementation of `log(1 - exp(x))` for use in `stats` following the suggestion laid out by @mdhaber in #19021.


scipy/special/xsf/log_exp.h
```diff
@@ -65,4 +66,26 @@ T log_expit(T x) {
...
 
+
+/* Compute log(1 - exp(x)). */
+template <typename T>
+T log1mexp(T x) {
+    if (x > 1) {
+	set_error("_log1mexp", SF_ERROR_DOMAIN, NULL);
+	return std::numeric_limits<T>::quiet_NaN();
+    }
+    if (x == 1) {
```
mdhaber, line 77: Should the threshold here be `0` instead of `1`?

mdhaber:
@steppi I'll go ahead and merge, but please see addition to `_ufuncs.pyi` in https://github.com/scipy/scipy/pull/22439/commits/fedd0bb8a111afc2e852d22a54b821fd891c62a2 in response to [previous mypy failure](https://github.com/scipy/scipy/actions/runs/13044225384/job/36391979592?pr=22439). Hopefully that was right?

steppi:
> @steppi I'll go ahead and merge, but please see addition to `_ufuncs.pyi` in [fedd0bb](https://github.com/scipy/scipy/commit/fedd0bb8a111afc2e852d22a54b821fd891c62a2) in response to [previous mypy failure](https://github.com/scipy/scipy/actions/runs/13044225384/job/36391979592?pr=22439). Hopefully that was right?

Yes, that looks right to me.

PR #22445, opened 2025-01-30
CI: avoid mpmath pre-release version that's failing in CI

rgommers:
xref gh-22395


steppi (approved):
Looks good to me. Thanks @rgommers.

PR #22448, opened 2025-01-31
DOC: optimize.elementwise.find_minimum: fix documented termination condition

### LINKED ISSUE #22309: DOC: optimize.elementwise.find_minimum: harmonize documented/implemented interval termination condition

mdhaber:
#### Reference issue
Closes gh-22309

Makes the documented termination condition consistent with the reference and implementation.

### DIFF
--- a/scipy/optimize/_elementwise.py
+++ b/scipy/optimize/_elementwise.py
@@ -342,11 +342,8 @@ def find_minimum(f, init, /, *, args=(), tolerances=None, maxiter=100, callback=
     (where one of the inequalities is strict) are the values of `f` evaluated
     at those points, then the algorithm is considered to have converged when:
 
-    - ``xr - xl <= abs(xm)*xrtol + xatol`` or
+    - ``abs(xr - xm)/2 <= abs(xm)*xrtol + xatol`` or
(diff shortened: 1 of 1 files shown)


steppi (approved):
Looks good to me. Thanks @mdhaber

PR #20610, opened 2024-04-29
ENH: signal.ShortTimeFFT: determine arbitrary dual windows

DietBru:
The [ShortTimeFFT] can only calculate the canonical dual window (which is utilized in the [istft]). 

This PR provides a new function [closest_STFT_dual_window] to calculate the closest dual window in a least-squares sense for a given desired window as well as a new method [ShortTimeFFT.from_win_equals_dual] to determine windows which are equal to their dual window.  

The Constant OverLap Add (COLA) constraint is equivalent of having a constant dual window, i.e., ``all(ShortTimeFFT.dual_win == ShortTimeFFT.dual_win[0])``. Hence, `closest_STFT_dual_window()` is a generalization of [check_COLA], which is deprecated in this PR.

Closes #20608 by adding `'math:numref'` to the white list in [refguide_check.py].

Further minor changes:
* In  `_calc_dual_canonical_window(Fixed edgde case i)` fixed edge case for catching invalid zero-valued `win`. 
* Allow `np.int` for `hop` parameters.
* Removed minor typos and reference for STFT applications  
* Fixed a linter error in`_spectral_helper()`

Update: The rendered documentation can be found here:
* [closest_STFT_dual_window]
* [ShortTimeFFT.from_win_equals_dual]
* [ShortTimeFFT.spectrogram]
* [Updates]
* [check_COLA]

~~Update: Resolves  #20622 by adding additional documentation.~~
Update: Though there is an example added utilizing a cross-spectrogram, it is probably not expressive enough to close the Issue #20622.

[ShortTimeFFT]: https://output.circle-artifacts.com/output/job/2fbf1eab-d6ef-45a5-97b8-0163de6e680b/artifacts/0/html/reference/generated/scipy.signal.ShortTimeFFT.html

[istft]: https://output.circle-artifacts.com/output/job/2fbf1eab-d6ef-45a5-97b8-0163de6e680b/artifacts/0/html/reference/generated/scipy.signal.ShortTimeFFT.istft.html#scipy.signal.ShortTimeFFT.istft

[refguide_check.py]: https://github.com/scipy/scipy/blob/main/tools/refguide_check.py#L339

[closest_STFT_dual_window]: https://output.circle-artifacts.com/output/job/2fbf1eab-d6ef-45a5-97b8-0163de6e680b/artifacts/0/html/reference/generated/scipy.signal.closest_STFT_dual_window.html

[...]

PR #22454, opened 2025-02-01
BUG: signal.ShortTimeFFT: make attributes `win` and `dual_win` immutable

DietBru:
This a follow-up to this [suggestion](https://github.com/scipy/scipy/pull/20610#discussion_r1925984787) from @larsoner:

To prevent the following, the WRITABLE flag of the respective arrays need to be set to ``False``:
```python
SFT = ShortTimeFFT(np.ones(6), 3, fs=1)  
SFT.win[3] = 5
```

Furthermore, this PR contains some minor docstr corrections.

### DIFF
--- a/scipy/signal/_short_time_fft.py
+++ b/scipy/signal/_short_time_fft.py
@@ -301,7 +301,7 @@ class ShortTimeFFT:
     A typical STFT application is the creation of various types of time-frequency
     plots, often subsumed under the term "spectrogram". Note that this term is also
     used to explecitly refer to the absolute square of a STFT [11]_, as done in
-    `~ShortTimeFFT.spectrogram`.
+    :meth:`spectrogram`.
 
     The STFT can also be used for filtering and filter banks as discussed in [12]_.
 
@@ -432,8 +432,9 @@ class ShortTimeFFT:
             raise ValueError("Parameter win must have finite entries!")
         if not (hop >= 1 and isinstance(hop, int | np.integer)):
             raise ValueError(f"Parameter {hop=} is not an integer >= 1!")
-        self._win, self._hop, self.fs = win, hop, fs
 
+        self._win, self._hop, self.fs = win, hop, fs
+        self.win.setflags(write=False)
         self.mfft = len(win) if mfft is None else mfft
 
         if dual_win is not None:
@@ -441,6 +442,7 @@ class ShortTimeFFT:
                 raise ValueError(f"{dual_win.shape=} must equal {win.shape=}!")
             if not all(np.isfinite(dual_win)):
                 raise ValueError("Parameter dual_win must be a finite array!")
+            dual_win.setflags(write=False)
         self._dual_win = dual_win  # needs to be set before scaling
 
         if scale_to is not None:  # needs to be set before fft_mode
@@ -658,7 +660,7 @@ class ShortTimeFFT:
         Notes
         -----
         The set of all possible windows with identical dual is defined by the set of
-        linear constraints of Eq. :math:numref:`eq_STFT_EqualWindDualCond` in the
+        linear constraints of Eq. :math:numref:`eq_STFT_AllDualWinsCond` in the
         :ref:`tutorial_stft` section of the :ref:`user_guide`. There it is also
         derived that ``ShortTimeFFT.dual_win == ShortTimeFFT.m_pts * ShortTimeFFT.win``
         needs to hold for an STFT to be a unitary mapping.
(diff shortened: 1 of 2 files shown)


scipy/signal/_short_time_fft.py
```diff
@@ -773,6 +776,23 @@ def from_win_equals_dual(
...
+        setting the flag WRITEABLE of `a` to ``False``. This prevents the following:
+
+        >>> SFT = ShortTimeFFT(np.ones(6), 3, fs=1)  # doctest: +SKIP
+        ... SFT.win[3] = 5  # this raises a ValueError
+
+        Currently, this only works with NumPy arrays. For all other Array-API types,
+        parameter `a` is not modified.
+        """
+        xp = array_namespace(a)
+        if xp.is_numpy_array(a):
```
j-bowhay, line 793: Does this need to be array agnostic since the class isn't?

DietBru: No,it is not—it is supposed to be an optimistic start for making the class array agnostic. The main stumbling block for me is [np.pad]. The other things, besides `fft` functions, seem quite straightforward.

Do you think the array agnosticisms should be added in go?

[np.pad]: https://github.com/scipy/scipy/blob/36291dcfd82165242792e7f28560d6a6f56ee6ea/scipy/signal/_short_time_fft.py#L1088

j-bowhay: I guess my thought was that setting the read only flag only works for numpy but since these look like properties could we not add a setter that raises an error?

Re `np.pad` see https://data-apis.org/array-api-extra/generated/array_api_extra.pad.html#array_api_extra.pad

DietBru: >  I guess my thought was that setting the read only flag only works for numpy but since these look like properties could we not add a setter that raises an error?

If you are referring to adding setters to `ShortTimeFFT`—this will not have any effect.

After giving it some more thought, I removed the Array-API sensible behavior and added a note to the docstrs of `win` and `dual_win`. 

PS: The `array_api_extra.pad` seems useful, thanks. I did not find when I looked for it a while back.

DietBru: ... though `array_api_extra.pad` does not have all the `modes` implemented yet compared to [np.pad].

[np.pad]: https://numpy.org/doc/stable/reference/generated/numpy.pad.html

scipy/signal/_short_time_fft.py
```diff
@@ -773,12 +775,29 @@ def from_win_equals_dual(
...
+        immutable by setting the flag WRITEABLE of `a` to ``False``.
+        This prevents the following:
+
+        >>> SFT = ShortTimeFFT(np.ones(6), 3, fs=1)  # doctest: +SKIP
+        ... SFT.win[3] = 5  # this raises a ValueError
+
+        This one-line helper exists to make future porting to Array-API compatbile
+        behavior easier.
+        """
+        a.setflags(write=False)
```
j-bowhay, line 792: Up to you but IMO probably not worth a static method for a one liner

scipy/signal/_short_time_fft.py
j-bowhay: I imagine at the time the other `modes` weren't immediately needed, I'm sure addictions would be welcome:)

scipy/signal/_short_time_fft.py
ev-br: FWIW, I'm +1 to what Jake said. 

If you *really* want to make array attributes immutable, the only future-proof way is to make them properties where getters return copies:

```
@property
def win(self):
    return self._win.copy()
```

Whether this is worth it vs documenting that users should not modify them willy-nilly... My knee-jerk reaction is towards the latter but maybe it is really justified in this particular case, I don't know.

DietBru: * Using `.copy()` is not too attractive, since `win` and `dual_win` are used internally quite a bit amd `dual_win` is created lazily. Furthermore, a design idea of the `ShortTimeFFT` class was to provide all STFT variables (like `win` or `f`) with little or no overhead.
* You're right, a one-line helper is probably not that useful → removed.
* I do think that un-setting the WRITABLE flag is marginally more useful than not doing it. Furthermore, this can be reversed in the future without breaking any existing code.
* I squashed the commits, because the back and forth is not useful without the context of this discussion.

scipy/signal/_short_time_fft.py
```diff
@@ -774,11 +776,13 @@ def from_win_equals_dual(
         return SFT
 
 
+
```
j-bowhay, line 779:
```suggestion
```

scipy/signal/tests/test_short_time_fft.py
```diff
@@ -278,7 +278,6 @@ def test_exceptions_properties_methods():
         SFT.fft_mode = 'twosided'
         SFT.extent(n=100)
```
j-bowhay, line 280:
```suggestion


```

j-bowhay (approved):
Thanks!

DietBru:
Thanks @j-bowhay  and  @ev-br!

PR #22456, opened 2025-02-02
ENH: stats: add nan_policy support to power_divergence, chisquare

mdhaber:
#### Reference issue
Toward gh-14651

Adds `nan_policy` support to `stats.power_divergence` and `stats.chisquare`.

### DIFF
--- a/scipy/stats/_stats_py.py
+++ b/scipy/stats/_stats_py.py
@@ -7059,6 +7059,12 @@ Power_divergenceResult = namedtuple('Power_divergenceResult',
                                     ('statistic', 'pvalue'))
 
 
+def _pd_nsamples(kwargs):
+    return 2 if kwargs.get('f_exp', None) is not None else 1
+
+
+@_axis_nan_policy_factory(Power_divergenceResult, paired=True, n_samples=_pd_nsamples,
+                          too_small=-1)
 def power_divergence(f_obs, f_exp=None, ddof=0, axis=0, lambda_=None):
     """Cressie-Read power divergence statistic and goodness of fit test.
 
@@ -7266,9 +7272,9 @@ def _power_divergence(f_obs, f_exp, ddof, axis, lambda_, sum_check=True):
                 raise ValueError(msg)
 
     else:
-        # Ignore 'invalid' errors so the edge case of a data set with length 0
-        # is handled without spurious warnings.
-        with np.errstate(invalid='ignore'):
+        # Avoid warnings with the edge case of a data set with length 0
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
             f_exp = xp.mean(f_obs, axis=axis, keepdims=True)
 
     # `terms` is the array of terms that are summed along `axis` to create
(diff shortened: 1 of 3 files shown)


scipy/stats/_stats_py.py
```diff
@@ -7059,6 +7059,12 @@ def ttest_rel(a, b, axis=0, nan_policy='propagate', alternative="two-sided"):
                                     ('statistic', 'pvalue'))
 
 
+def _pd_nsamples(kwargs):
+    return 2 if kwargs.get('f_exp', None) is not None else 1
+
+
+@_axis_nan_policy_factory(Power_divergenceResult, paired=True, n_samples=_pd_nsamples,
+                          too_small=-1)
```
mdhaber, line 7067: Technically, the statistic is `0` for samples of zero length.

scipy/stats/_stats_py.py
```diff
@@ -7266,9 +7272,9 @@ def _power_divergence(f_obs, f_exp, ddof, axis, lambda_, sum_check=True):
                 raise ValueError(msg)
 
     else:
-        # Ignore 'invalid' errors so the edge case of a data set with length 0
-        # is handled without spurious warnings.
-        with np.errstate(invalid='ignore'):
+        # Avoid warnings with the edge case of a data set with length 0
+        with warnings.catch_warnings():
```
mdhaber, line 7276: I think we're already using `catch_warnings` in this file for a similar purpose.

scipy/stats/tests/test_axis_nan_policy.py
```diff
@@ -866,15 +868,16 @@ def small_sample_generator(n_dims):
                     sup.filter(RuntimeWarning, "Mean of empty slice.")
                     sup.filter(RuntimeWarning, "invalid value encountered")
                     expected = np.mean(concat, axis=axis) * np.nan
+                    mask = np.isnan(expected)
+                    expected = [np.asarray(expected.copy()) for i in range(n_outputs)]
```
mdhaber, line 872: The purpose of this part of `assert_equal(res[i], expected[i])` is to check that empty slices produce the intended statistic and p-value. Previously, the only function that didn't produce NaN with empty input was `entropy`, and that has only one output. (For example, see below that `expected[mask] = empty_val`, but `empty_val` is the output of `hypotest`, so if it were a tuple, we'd have issues.) So this needs to be generalized for multiple outputs.

This line assumes that all expected values are NaN; we adjust that for special case functions below.

scipy/stats/tests/test_stats.py
```diff
@@ -4228,13 +4228,13 @@ def check_power_divergence(self, f_obs, f_exp, ddof, axis, lambda_,
         with suppress_warnings() as sup:
             sup.filter(RuntimeWarning, "Mean of empty slice")
             stat, p = stats.power_divergence(
-                                f_obs=f_obs, f_exp=f_exp, ddof=ddof,
+                                f_obs, f_exp=f_exp, ddof=ddof,
```
mdhaber, line 4231: Compensates for a shortcoming of `axis_nan_policy_decorator` when non-NumPy backends are used and no arrays are passed as keyword arguments.

scipy/stats/tests/test_stats.py
```diff
@@ -4413,7 +4413,7 @@ def test_chisquare_12282a(self, xp):
         with assert_raises(ValueError, match='For each axis slice...'):
             f_obs = xp.asarray([10., 20.])
             f_exp = xp.asarray([30., 60.])
-            stats.chisquare(f_obs=f_obs, f_exp=f_exp)
+            stats.chisquare(f_obs, f_exp=f_exp)
```
mdhaber, line 4416: I missed this in the first CI run, and it was the cause of the only failure, so I went ahead and skipped CI on this commit.

j-bowhay (approved):

PR #22457, opened 2025-02-02
TST: sparse: add tests for subscriptable types

zitongzhoueric:
This PR adds two tests that add coverage for runtime-subscriptable ```sparray``` and ```spmatrix``` types that were introduced in #22401.

### DIFF
--- a/scipy/sparse/tests/test_base.py
+++ b/scipy/sparse/tests/test_base.py
@@ -29,6 +29,8 @@ from numpy.testing import (assert_equal, assert_array_equal,
         assert_array_almost_equal, assert_almost_equal, assert_,
         assert_allclose, suppress_warnings)
 
+from types import GenericAlias
+
 import scipy.linalg
 
 import scipy.sparse as sparse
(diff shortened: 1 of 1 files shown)


j-bowhay:
Cc @jorenham

jorenham (commented):
Perhaps it's also worth checking whether the concrete subclasses, e.g. `coo_array` and `coo_matrix`, are also (still) subscriptable?

And in case of `sparray`, it should also be allowed to pass a single type parameter, which you could also consider checking here (although it's rather trivial if you consider the implementation, but regressions might occur either way)

anonymouspenguin1024:
> Perhaps it's also worth checking whether the concrete subclasses, e.g. `coo_array` and `coo_matrix`, are also (still) subscriptable?

The commit already checks that `coo_array` is subscriptable. For `spmatrix` I didn't find a proper test class so ended up checking `csr_matrix`. Let me know if there is better way of doing it. 

> And in case of `sparray`, it should also be allowed to pass a single type parameter, which you could also consider checking here (although it's rather trivial if you consider the implementation, but regressions might occur either way)

Added

jorenham (approved):

anonymouspenguin1024:
Oh didn't notice that you just approved it. NVM

jorenham (approved):

anonymouspenguin1024:
test failure seems unrelated?

tylerjereddy (approved):
Yes, failures unrelated. Merging based on Joren's approval on the typing side, and LGTM as well (I'm not a typing expert, but this does look pretty straightforward).

tylerjereddy:
Thanks for you first contribution to SciPy.

PR #22462, opened 2025-02-03
ENH: stats.pearsonr: two simple (but substantial) efficiency improvements

mdhaber:
This makes two simple efficiency improvements to `stats.pearsonr` that are important for the use cases of pairwise calculations.
- Avoid broadcasting the arrays until necessary. A lot of work is done on the arrays separately before they are brought together, so don't potentially increase their sizes prematurely.
- Use `linalg.vecdot`. It is available in the standard, and it's a lot faster than `*` and `sum`.

#### Additional information
Consider pairwise correlation of 500 samples, each with 500 observations:
```python3
import numpy as np
from scipy import stats
rng = np.random.default_rng(2785487287265)

m, n = 500, 500
x = rng.random((m, n))
stats.pearsonr(x, x[:, np.newaxis, :], axis=-1)
```
Timing and profiling:
```
# %timeit stats.pearsonr(x, x[:, np.newaxis, :], axis=-1)
# before: 4.3 s ± 263 ms per loop (mean ± std. dev. of 7 runs, 1 loop each), ~5% of time getting pvalue
# after: 238 ms ± 2.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each), ~85% of time getting pvalue
```

There is one problem as-written: like other `stats` functions (and unlike NumPy gufuncs, unfortunately) `axis` is interpreted after matching the number of dimensions in the two arrays. (This is something we might consider changing for SciPy 2.0.) So if the arrays are of different dimensionality and positive indices are used, this would change the behavior. To be backward compatible and consistent with other stats functions, I would need to add something like:
```python3
ndim = max(x.ndim, y.ndim)
x = x.reshape((1,)*(ndim - x.ndim) + x.shape)
y = y.reshape((1,)*(ndim - y.ndim) + y.shape)
```
This would make the PR look a little more complex than it is, so I wanted to wait to add that until it looks good otherwise.

### DIFF
--- a/scipy/stats/_stats_py.py
+++ b/scipy/stats/_stats_py.py
@@ -4564,7 +4564,7 @@ def pearsonr(x, y, *, alternative='two-sided', method=None, axis=0):
         raise ValueError('`x` and `y` must have length at least 2.')
 
     try:
-        x, y = xp.broadcast_arrays(x, y)
+        np.broadcast_shapes(x.shape, y.shape)
     except (ValueError, RuntimeError) as e:
         message = '`x` and `y` must be broadcastable.'
         raise ValueError(message) from e
@@ -4658,7 +4658,7 @@ def pearsonr(x, y, *, alternative='two-sided', method=None, axis=0):
         warnings.warn(stats.NearConstantInputWarning(msg), stacklevel=2)
 
     with np.errstate(invalid='ignore', divide='ignore'):
-        r = xp.sum(xm/normxm * ym/normym, axis=axis)
+        r = xp.linalg.vecdot(xm / normxm, ym / normym, axis=axis)
 
     # Presumably, if abs(r) > 1, then it is only some small artifact of
     # floating point arithmetic.
--- a/scipy/stats/_stats_py.py
+++ b/scipy/stats/_stats_py.py
@@ -4556,18 +4556,26 @@ def pearsonr(x, y, *, alternative='two-sided', method=None, axis=0):
         raise ValueError('`axis` must be an integer.')
     axis = axis_int
 
+    try:
+        np.broadcast_shapes(x.shape, y.shape)
+        # For consistency with other `stats` functions, we need to
+        # match the dimensionalities before looking at `axis`.
+        # (Note: this is not the NEP 5 / gufunc order of operations;
+        #  see TestPearsonr::test_different_dimensionality for more information.)
+        ndim = max(x.ndim, y.ndim)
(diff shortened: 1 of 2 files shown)


scipy/stats/_stats_py.py
```diff
@@ -4658,7 +4658,7 @@ def statistic(x, y, axis):
         warnings.warn(stats.NearConstantInputWarning(msg), stacklevel=2)
 
     with np.errstate(invalid='ignore', divide='ignore'):
-        r = xp.sum(xm/normxm * ym/normym, axis=axis)
+        r = xp.linalg.vecdot(xm / normxm, ym / normym, axis=axis)
```
j-bowhay, line 4661: Do we need to worry about checking if the `linalg` extension is implemented? https://data-apis.org/array-api/latest/extensions/linear_algebra_functions.html#extension-name-and-usage

mdhaber: I suppose I assumed that `array_api_compat` already would. If not, and If we don't already have an `xp_vecdot` that does that, sounds like something for `_array_api.py` or `array-api-extra`.

mdhaber: But it might also be valid to error out of it's not implemented. If we're not going to provide a generic implementationsl, I don't see a big difference between checking and raising vs just raising when it fails.

mdhaber: Oops, I should have just called `vecdot`, not `linalg.vecdot`.

j-bowhay: Sounds good, the rest looks good too

scipy/stats/tests/test_stats.py
```diff
@@ -764,6 +770,24 @@ def test_nd_special_cases(self, xp):
         xp_assert_close(ci.low, -ones)
         xp_assert_close(ci.high, ones)
 
+    def test_different_dimensionality(self, xp):
```
j-bowhay, line 773: I tried this test with 1.14.1 and it failed, is this expected?

mdhaber: Yup. I noticed while I was addressing this that the code was inconsistent. That's why I swapped the order of things. This opens up behavior that failed before (that shouldn't have), but it doesn't prevent anything that did work before. Try the test with basically any other two-arg reducing function.

j-bowhay (approved):
Nice speed up!

PR #22463, opened 2025-02-04
DOC: update Halton docs

tupui:
To account for some comments from Art Owen.

1. `n` could be misinterpreted as any integer while it is to be a prime number. (Not something user have control over as part of the VDC construction.)
2. The mention to Owen scrambling could be confused with Owen's nested uniform scrambling from 1995. We do reference the paper from 2017 but this would clarify a bit more in the params.

### DIFF
--- a/scipy/stats/_qmc.py
+++ b/scipy/stats/_qmc.py
@@ -1117,7 +1117,7 @@ class Halton(QMCEngine):
     Pseudo-random number generator that generalize the Van der Corput sequence
     for multiple dimensions. The Halton sequence uses the base-two Van der
     Corput sequence for the first dimension, base-three for its second and
-    base-:math:`n` for its :math:`n`-dimension, with :math:`n` the
+    base-:math:`p` for its :math:`n`-dimension, with :math:`p` the
     :math:`n`'th prime.
 
     Parameters
--- a/scipy/stats/_qmc.py
+++ b/scipy/stats/_qmc.py
@@ -1117,14 +1117,16 @@ class Halton(QMCEngine):
     Pseudo-random number generator that generalize the Van der Corput sequence
     for multiple dimensions. The Halton sequence uses the base-two Van der
     Corput sequence for the first dimension, base-three for its second and
-    base-:math:`n` for its n-dimension.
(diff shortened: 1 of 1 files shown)


scipy/stats/_qmc.py
```diff
@@ -1117,14 +1117,16 @@ class Halton(QMCEngine):
     Pseudo-random number generator that generalize the Van der Corput sequence
     for multiple dimensions. The Halton sequence uses the base-two Van der
     Corput sequence for the first dimension, base-three for its second and
-    base-:math:`n` for its n-dimension.
+    base-:math:`n` for its :math:`n`-dimension, with :math:`n` the :math:`n`'th
+    prime.
```
tupui, line 1121:
```suggestion
    base-:math:`n` for its :math:`n`-dimension, with :math:`n` the
    :math:`n`'th prime.
```

scipy/stats/_qmc.py
```diff
@@ -1117,14 +1117,16 @@ class Halton(QMCEngine):
     Pseudo-random number generator that generalize the Van der Corput sequence
     for multiple dimensions. The Halton sequence uses the base-two Van der
     Corput sequence for the first dimension, base-three for its second and
-    base-:math:`n` for its n-dimension.
+    base-:math:`n` for its :math:`n`-dimension, with :math:`n` the
+    :math:`n`'th prime.
```
rkern, line 1121: I think something like this is needed to disambiguate.

```suggestion
    base-:math:`p` for its :math:`n`-dimension, with :math:`p` the
    :math:`n`'th prime.
```

scipy/stats/_qmc.py
```diff
@@ -1117,14 +1117,16 @@ class Halton(QMCEngine):
...
 
     Parameters
     ----------
     d : int
         Dimension of the parameter space.
     scramble : bool, optional
-        If True, use Owen scrambling. Otherwise no scrambling is done.
+        If True, use random scrambling from Owen. Otherwise no scrambling
```
mdhaber, line 1128:
```suggestion
        If True, use random scrambling from [2]_. Otherwise no scrambling
```
Is that what "Owen" refers to here?

rkern (approved):
LGTM!

PR #22464, opened 2025-02-04
DOC: Prevent A@x=b from becoming a URL

star1327p:

Change the linear system `A@x=b` to `A @ x = b`.
This prevents A@x=b from becoming a URL.

### DIFF
--- a/scipy/sparse/linalg/_interface.py
+++ b/scipy/sparse/linalg/_interface.py
@@ -56,7 +56,7 @@ class LinearOperator:
     """Common interface for performing matrix vector products
 
     Many iterative methods (e.g. cg, gmres) do not need to know the
-    individual entries of a matrix to solve a linear system A @ x = b.
+    individual entries of a matrix to solve a linear system ``A@x = b``.
     Such solvers only require the computation of matrix vector
     products, A @ v where v is a dense vector.  This class serves as
     an abstract interface between iterative solvers and matrix-like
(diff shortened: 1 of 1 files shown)


scipy/sparse/linalg/_interface.py
```diff
@@ -56,9 +56,9 @@ class LinearOperator:
     """Common interface for performing matrix vector products
 
     Many iterative methods (e.g. cg, gmres) do not need to know the
-    individual entries of a matrix to solve a linear system A@x=b.
+    individual entries of a matrix to solve a linear system A @ x = b.
```
j-bowhay, line 59:
```suggestion
    individual entries of a matrix to solve a linear system ``A@x = b``.
```

scipy/sparse/linalg/_interface.py
```diff
@@ -56,9 +56,9 @@ class LinearOperator:
...
     Such solvers only require the computation of matrix vector
-    products, A@v where v is a dense vector.  This class serves as
+    products, A @ v where v is a dense vector.  This class serves as
```
j-bowhay, line 61:
```suggestion
    products, ``A@v`` where ``v`` is a dense vector.  This class serves as
```

j-bowhay (changes requested):
Thanks for spotting this! These should actually be surrounded in back ticks

Also please use `[docs only]` in the commit message for doc changes to avoid unnecessary ci usage

star1327p:
@j-bowhay Thank you for your suggestions! I have added your edits into this PR.

ilayn:
Thank you @star1327p . Do you mind double back-ticking a few things in the "Parameters" section similarly while you are at it?

j-bowhay (approved):
Thanks @star1327p ! I took the liberty of adding the required back ticks

ilayn (approved):

PR #22469, opened 2025-02-04
TST: stats: improve JAX test coverage

mdhaber:
Removes most remaining JAX test skips in `scipy.stats`. Most were needed before due to JAX's lack of mutability, but this is mostly resolved with `array_api_extra.at.set`.

*Update: will reopen when https://github.com/data-apis/array-api-extra/issues/134 is addressed*


scipy/stats/_entropy.py
```diff
@@ -403,9 +404,9 @@ def _ebrahimi_entropy(X, m, *, xp):
     differences = X[..., 2 * m:] - X[..., : -2 * m:]
 
     i = xp.arange(1, n+1, dtype=X.dtype)
-    ci = xp.ones_like(i)*2
-    ci[i <= m] = 1 + (i[i <= m] - 1)/m
-    ci[i >= n - m + 1] = 1 + (n - i[i >= n-m+1])/m
+    ci = xp.where(i <= m, 1 + (i - 1)/m, xp.ones_like(i)*2)
```
mdhaber, line 407: I suppose this could be `xp.asarray(2., dtype=X.dtype)`, but let's leave it alone until 2024.12 of the standard comes out. (Then it can just be, you know, `2.`).

scipy/stats/_stats_py.py
```diff
@@ -4672,7 +4672,7 @@ def statistic(x, y, axis):
     # floating point arithmetic.
     one = xp.asarray(1, dtype=dtype)
     r = xp.asarray(xp.clip(r, -one, one))
-    r[const_xy] = xp.nan
+    r = xpx.at(r, const_xy).set(xp.nan)
```
mdhaber, line 4675: @lucascolley I don't have Dask locally, but this looks like the only thing that could be causing the failures. Can you tell why this would change anything?

lucascolley: The docs for `xpx.at` say:

> Boolean masks are supported on Dask and jitted JAX arrays exclusively when idx has the same shape as x and y is 0-dimensional.

Are both of those conditions true? Maybe try `xp.asarray(xp.nan))` instead?

mdhaber: > Are both of those conditions true?

Yes, that's the problem. `r[const_xy] = xp.nan` worked, but because the conditions are met for `xpx.at.set` to do something different, it fails.

It looks like the problem doesn't occur if I use `array_api_extra` directly; just SciPy's current version.  Here is a minimal reproducer.
<details>

```python3
import numpy as np
import dask.array as xp
from scipy._lib._array_api import array_namespace

# import array_api_extra as xpx  # this works
import scipy._lib.array_api_extra as xpx  # this doesn't

x = xp.asarray([1., 2., 3.])
y = xp.asarray([1., 2.5, 3.])
xp = array_namespace(x, y)

r = xp.vecdot(x, y)
mask = xp.asarray(False)
y = xp.asarray(xp.nan)

# This works
# r[mask] = xp.nan

# This doesn't
r = xpx.at(r, xp.asarray(False)).set(xp.nan)

np.asarray(r)
```

</details>

But this brings up a question - why does `array_api_extra.at.set` get involved here at all?  The documentation of Dask array says:
image

Doing the regular indexing assignment (when `y` is a scalar, at least) seems to be something Dask can do.

lucascolley: I can probably look into this and give an answer, but @crusaderky may be able to do the same with less effort

mdhaber: I'll open up a report at array-api-extra since this seems to be new since 0.6.0.

mdhaber:
```suggestion
```
I think we can just get rid of this. If it were replacing using `nconst_xy` as the mask, it might be useful, but right now, it's using `const_xy`, and where that is `True`, the result will already be NaN.

scipy/stats/tests/test_stats.py
```diff
@@ -9648,10 +9610,11 @@ def test_nan_policy(self, xp):
         mask = (x == 3)
         x = xp.where(mask, xp.asarray(xp.nan), x)
 
-        # nan_policy='raise' raises an error
-        message = 'The input contains nan values'
-        with pytest.raises(ValueError, match=message):
-            _xp_var(x, nan_policy='raise')
+        if not is_jax(xp):  # skip jax.numpy (lazy array -> no `nan_policy='raise'`)
```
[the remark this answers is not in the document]
crusaderky: As a general approach I much prefer `skip_xp_backends` / `xfail_xp_backends` to inline `is_<backend>` tests, whenever reasonable, as it makes text search for TODOs much easier.

scipy/stats/tests/test_stats.py
```diff
@@ -380,6 +383,7 @@ def test_pROUNDROUND(self):
 
 
 @skip_xp_backends(cpu_only=True)
+@skip_xp_backends("jax.numpy", reason="data-apis/array-api-extra#134")
```
crusaderky, line 386: Should be fixed with dask 2025.2.0
```suggestion
```

crusaderky, line 386: This is a skip for jax, but the reason is a dask-specific issue (that has been resolved in dask 2025.2.0)?

mdhaber: Right, I just happened to noticed the mistake, too.

scipy/stats/tests/test_stats.py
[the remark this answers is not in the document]
mdhaber: Yeah, I'm familiar with the typical pattern. I just don't think it is the best thing to do *here* if it is temporary. If this is a permanent thing and lazy arrays will never raise with `nan_policy="raise"`, I'll change it, but I'm not a fan of breaking tests apart just for the sake of a temporary test skip. I'll include a comment with that text in the code for searchability, if that's the concern.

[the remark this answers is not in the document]
crusaderky: This is permanent. As a general rule lazy backends _cannot_ materialize the graph, short of implementing very fancy auto-materialization heuristics (which I understand torch should have, but I haven't tested them myself). 

Dask materializes the graph when you asks it to, and that is exceptionally bad for performance as everything until that moment gets computed twice. 
`jax.jit` and ndonnx flat out raise an exception.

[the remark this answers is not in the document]
mdhaber: That said, I wanted to check on whether there was an explicit decision somewhere for lazy arrays to ignore `nan_policy='raise'`? I get that it is inefficient, but `nan_policy='raise'` inherently does unnecessary work, and perhaps the user just needs to be familiar with the consequences. 

`nan_policy` for non-NumPy arrays is not in a great state right now, so it doesn't really matter yet, but I was hoping to make the behavior consistent eventually.

scipy/stats/_stats_py.py
```diff
@@ -4754,7 +4756,6 @@ def statistic(x, y, axis):
     # floating point arithmetic.
     one = xp.asarray(1, dtype=dtype)
     r = xp.asarray(xp.clip(r, -one, one))
```
mdhaber, line 4758:
```suggestion
    r = xp.asarray(xp.clip(r, -one, one))
    r = xpx.at(r, const_xy).set(xp.nan)
```

scipy/stats/tests/test_stats.py
mdhaber:
```suggestion
```

scipy/stats/tests/test_stats.py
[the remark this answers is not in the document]
crusaderky: https://github.com/scipy/scipy/pull/22566

scipy/stats/_entropy.py
crusaderky: 2024.12 is out now for numpy, cupy, array-api-strict, dask, and torch.
JAX is still at 2023.12 but it may already support this bit.

mdhaber:
Oops, didn't mean to delete the comment. Anyway, I also made the change requested that involved `xp.trim`.

mdhaber:
@crusaderky I broke up the test. I think that was the last comment. Look OK?

scipy/stats/_stats_py.py
```diff
@@ -4676,6 +4676,8 @@ def pearsonr(x, y, *, alternative='two-sided', method=None, axis=0):
         msg = ("An input array is constant; the correlation coefficient "
                "is not defined.")
         warnings.warn(stats.ConstantInputWarning(msg), stacklevel=2)
+        x = xp.where(const_x[..., np.newaxis], xp.asarray(np.nan, dtype=dtype), x)
+        y = xp.where(const_y[..., np.newaxis], xp.asarray(np.nan, dtype=dtype), y)
```
lucascolley, line 4680: this might just be cosmetic, but `xp.newaxis` and `xp.nan` rather than `np`?

lucascolley (approved):
LGTM, thanks Matt! Just one comment

scipy/stats/_stats_py.py
mdhaber: Oops, sure.

scipy/stats/tests/test_stats.py
```diff
@@ -9649,6 +9600,13 @@ def test_nan_policy(self, xp):
             ref = xp.asarray(xp.nan)
             xp_assert_equal(res, ref)
 
+    @skip_xp_backends(eager_only=True)
+    def test_nan_policy_raise(self, xp):
+        # nan_policy='raise' raises an error when NaNs are present
+        message = 'The input contains nan values'
+        with pytest.raises(ValueError, match=message):
+            _xp_var(xp.asarray([1, 2, np.nan]), nan_policy='raise')
```
mdhaber, line 9608:
```suggestion
            _xp_var(xp.asarray([1, 2, xp.nan]), nan_policy='raise')
```

scipy/stats/_stats_py.py
mdhaber:
```suggestion
        x = xp.where(const_x[..., xp.newaxis], xp.asarray(xp.nan, dtype=dtype), x)
        y = xp.where(const_y[..., xp.newaxis], xp.asarray(xp.nan, dtype=dtype), y)
```

mdhaber:
Looked safe enough to skip CI, but please confirm. Thanks!

PR #22473, opened 2025-02-05
DOC: slight change to an error message

Kakalinn:
This is just from personal experience, but I think it's clearer.

### DIFF
--- a/scipy/spatial/_spherical_voronoi.py
+++ b/scipy/spatial/_spherical_voronoi.py
@@ -186,7 +186,7 @@ class SphericalVoronoi:
             raise ValueError(f"Rank of input points must be at least {self._dim}")
 
         if cKDTree(self.points).query_pairs(threshold * self.radius):
-            raise ValueError("Duplicate generators present.")
+            raise ValueError("Duplicate points present.")
 
         radii = np.linalg.norm(self.points - self.center, axis=1)
         max_discrepancy = np.abs(radii - self.radius).max()


scipy/spatial/_spherical_voronoi.py
```diff
@@ -186,7 +186,7 @@ def __init__(self, points, radius=1, center=None, threshold=1e-06):
             raise ValueError(f"Rank of input points must be at least {self._dim}")
 
         if cKDTree(self.points).query_pairs(threshold * self.radius):
-            raise ValueError("Duplicate generators present.")
+            raise ValueError("Duplicate points present.")
```
tylerjereddy, line 189: For context, I use the term `generator` to represent the input points that are the basis of generating computational geometry diagrams based on a reading of the "Discrete and Computational Geometry"  textbook by Devadoss and O'Rourke. I think part of the motivation is to separate the terminology from the points that are produced after the algorithm runs (i.e., the Voronoi vertices are separated from the initial generators).

That said, perhaps it is best if I just mention my original motivation and kind of step out of the way and let other SciPy devs who may not work in the computational geometry space decide if this is effectively non-productive jargon.

I would ask that whatever decision is made gets applied consistently. There is another mention of generators elsewhere in the docs here, with:

> The Convex Hull of the input points (generators) is calculated ...

if the decision is that "generators" is basically pointless jargon, feel free to clean that up as well.

Kakalinn: This is reasonable. I should give my context for the change.

I got this error, using a third party program, as an unhandled exception. I was confused at first, but when I read the line above I understood what it meant. My confusion was maybe caused by the fact that I 'knew' there weren't repeat points in my input. The error was caused by a lack of precision.

I wouldn't have pushed this if not for the fact that the [docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.SphericalVoronoi.html) use "duplicate points". An alternative would be to use "generators" everywhere in the docs, namely in the "Parameters" description.

PR #22141, opened 2024-12-20
ENH: parallelisation of `optimize._numdiff.approx_derivative`

andyfaff:
#22119 prompted me to think a little more about parallelisation with various `minimize` methods. One of the most basic things to achieve would be to parallelise the numerical differentiation, probably using the `workers` pattern. The purpose of this PR is to get people's feedback about whether it's worth doing as part of a step on the journey.

PR #22478, opened 2025-02-06
ENH: optimize: add `workers` kwarg to BFGS, SLSQP, trust-constr

andyfaff:
Builds on #22141.
Covers: BFGS, Newton-CG, CG, SLSQP, TNC, trust-constr.

Only parallelises the gradient of the objective function, doesn't cover the gradient of any constraints (SLSQP, trust-constr) yet.

### DIFF
--- a/scipy/optimize/tests/test_optimize.py
+++ b/scipy/optimize/tests/test_optimize.py
@@ -3257,7 +3257,15 @@ def test_sparse_hessian(method, sparse_type):
 @pytest.mark.parametrize('workers', [None, 2])
 @pytest.mark.parametrize(
     'method',
-    ['l-bfgs-b', 'bfgs', 'slsqp', 'trust-constr', 'Newton-CG', 'CG', 'tnc'])
+    ['l-bfgs-b',
+     'bfgs',
+     'slsqp',
+     'trust-constr',
+     'Newton-CG',
+     'CG',
+     'tnc',
+     'trust-ncg',
+     'trust-krylov'])
 class TestWorkers:
 
     def setup_method(self):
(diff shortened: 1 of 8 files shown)


andyfaff:
No constraint parallelisation yet. I don't think the PR necessarily needs it.
It's troublesome pickling old constraints that are converted to new, and vice versa

scipy/optimize/tests/test_optimize.py
```diff
@@ -3264,19 +3266,24 @@ def setup_method(self):
     def test_smoke(self, workers, method):
         # checks parallelised optimization output is same as serial
         workers = workers or map
+
+        jac = None
+        if method in ['Newton-CG']:
+            jac = rosen_der
```
dschmitz89, line 3272: Why the special case for `neweton-cg`? If we use the analytical derivative, the new parallel numerical differentiation is not used from what I understand.

andyfaff: For Newton-CG if `hess` is going to be estimated via finite differences then `jac` has to be analytical. Therefore the only parallelisation is happening with the numerical differentiation of `hess`

andyfaff: `jac` has to be analytical for Newton-CG

dschmitz89: Ah alright, I was not aware of this. I am curios though: why exactly does it need an analytical gradient? Is the numerically approximated Hessian too imprecise for this particular algorithm otherwise? The code contains no comment, maybe lost to history?

andyfaff: I think it'll be due to inaccuracy. A second order partial derivative for the Hessian will be a lot more inaccurate than a first order partial derivative of the gradient.

andyfaff:
This PR is ready for review now.

dschmitz89 (approved):

andyfaff:
Have added release note.

andyfaff:
Merging, approval was given by other maintainer.

PR #22481, opened 2025-02-06
BUG: fix `make_lsq_spline` with a non-default axis

ev-br:

Add smoke testing of `make_interp_spline` and `make_lsq` spline with batched `y` arrays and non-default `axis`. The behavior is mostly derived from that of `BSpline`, which is why it was not explicitly tested before.

Actually testing it smokes out a small bug: compiled code expects C contiguous arrays, so need to actually ensure that they are.

### DIFF
--- a/scipy/interpolate/_bsplines.py
+++ b/scipy/interpolate/_bsplines.py
@@ -1765,6 +1765,9 @@ def make_lsq_spline(x, y, t, k=3, w=None, axis=0, check_finite=True, *, method="
     axis = normalize_axis_index(axis, y.ndim)
 
     y = np.moveaxis(y, axis, 0)    # now internally interp axis is zero
+    if not y.flags.c_contiguous:
+        # C routines in _dierckx currently require C contiguity
+        y = y.copy()
 
     if x.ndim != 1:
         raise ValueError("Expect x to be a 1-D sequence.")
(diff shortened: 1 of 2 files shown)


scipy/interpolate/_bsplines.py
```diff
@@ -1765,6 +1765,9 @@ def make_lsq_spline(x, y, t, k=3, w=None, axis=0, check_finite=True, *, method="
     axis = normalize_axis_index(axis, y.ndim)
 
     y = np.moveaxis(y, axis, 0)    # now internally interp axis is zero
+    if not y.flags.c_contiguous:
+        # C routines in _dierckx currently require C contiguity
+        y = y.copy()
```
mdhaber, line 1770: I see it's the default, but might be best to make it explicit since it's _not_ the default of `np.copy`.
```suggestion
        y = y.copy(order='C')
```

scipy/interpolate/tests/test_bsplines.py
```diff
@@ -1525,6 +1525,26 @@ def test_shapes(self):
         b = make_interp_spline(x, y, k, bc_type=(d_l, d_r))
         assert b.c.shape == (n + k - 1, 5, 6, 7)
 
+    @pytest.mark.parametrize("axis", range(1, 4))
+    def test_shapes_axis(self, axis):
```
mdhaber, line 1529: This is OK for now, but I can imagine a C vs F contiguity bug that gives correct shapes but incorrect numbers. In the course of reviewing new `axis` arguments, I'd like to write a common test `@parametrize`d over the various spline functions to confirm that values match those you'd get by looping over 1D slices.

mdhaber (approved):
Please consider making `order` explicit, but in any case, I can merge this when you're ready.

scipy/interpolate/_bsplines.py
ev-br: Agreed, done

PR #22499, opened 2025-02-09
DOC: remove links to the reference guide in the tutorials page

### LINKED ISSUE #22485: DOC: remove links to the reference guide in the tutorials page
Looking for a tutorial / userguide for a submodule in scipy, my search engine brings me to https://docs.scipy.org/doc/scipy/tutorial/index.html#user-guide, which has two columns : a module and a link to its tutorial.
While I am looking for the latter, every single time I click on the former. Not sure if it's the layout or that font is bold or someting else.

Image

karthik-ganti-2025:
…ex.rst

Since doc/source/conf.py makes text marked up by single backticks into auto-hyperlinks, I've modified the subpackages column to use double backticks.

#### Reference issue
Closes #22485

### DIFF
--- a/doc/source/tutorial/index.rst
+++ b/doc/source/tutorial/index.rst
@@ -25,23 +25,23 @@ linked in the Description column:
 ==================  ========================================
 Subpackage          Description and User Guide
 ==================  ========================================
-`cluster`           Clustering algorithms
-`constants`         Physical and mathematical constants
-`differentiate`     Finite difference differentiation tools
-`fft`               :doc:`./fft`
-`fftpack`           Fast Fourier Transform routines (legacy)
-`integrate`         :doc:`./integrate`
-`interpolate`       :doc:`./interpolate`
-`io`                :doc:`./io`
-`linalg`            :doc:`./linalg`
-`ndimage`           :doc:`./ndimage`
-`odr`               Orthogonal distance regression
-`optimize`          :doc:`./optimize`
-`signal`            :doc:`./signal`
-`sparse`            :doc:`./sparse`
-`spatial`           :doc:`./spatial`
-`special`           :doc:`./special`
-`stats`             :doc:`./stats`
+``cluster``           Clustering algorithms
+``constants``         Physical and mathematical constants
+``differentiate``     Finite difference differentiation tools
+``fft``               :doc:`./fft`
+``fftpack``           Fast Fourier Transform routines (legacy)
+``integrate``         :doc:`./integrate`
+``interpolate``       :doc:`./interpolate`
+``io``                :doc:`./io`
+``linalg``            :doc:`./linalg`
+``ndimage``           :doc:`./ndimage`
+``odr``               Orthogonal distance regression
+``optimize``          :doc:`./optimize`
+``signal``            :doc:`./signal`
+``sparse``            :doc:`./sparse`
+``spatial``           :doc:`./spatial`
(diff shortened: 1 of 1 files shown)


mdhaber:
Exactly. Could you `git reset --soft HEAD~1` then redo the commit with `[docs only]` in the commit message? That way, only the documentation CI will run to save resources.

doc/source/tutorial/index.rst
```diff
@@ -25,23 +25,23 @@ linked in the Description column:
 ==================  ========================================
 Subpackage          Description and User Guide
 ==================  ========================================
```
j-bowhay, line 27: I'm not sure if we need to tweak the length of these lines too?

karthik-ganti-2025: I'm not sure. The rule is 77 characters or something, right? This is 60 at the maximum.

karthik-ganti-2025:
> Exactly. Could you `git reset --soft HEAD~1` then redo the commit with `[docs only]` in the commit message? You'll need to force push (and that's OK in this case). That way, only the documentation CI will run to save resources.

I could do that. But it looks like the tests did pass eventually - not sure if this was before or after you took a look. Should I do this anyway - it sounds like they'd have to run all over it again.

Either way, thank you, I'll definitely look to implement this feedback for future PRs etc. 

Just let me know if you want me to do it for this one as well.

mdhaber:
No need here. It so happens that for first time contributors, the documentation jobs run without maintainer approval but the rest don't.

doc/source/tutorial/index.rst
j-bowhay: Sorry I wasn't quite clear, since the content in the table got longer does the number  of equals signs etc need to be changed?

mdhaber, line 27:
```suggestion
==================    ========================================
Subpackage            Description and User Guide
==================    ========================================
```

doc/source/tutorial/index.rst
```diff
@@ -25,23 +25,23 @@ linked in the Description column:
...
+``linalg``            :doc:`./linalg`
+``ndimage``           :doc:`./ndimage`
+``odr``               Orthogonal distance regression
+``optimize``          :doc:`./optimize`
+``signal``            :doc:`./signal`
+``sparse``            :doc:`./sparse`
+``spatial``           :doc:`./spatial`
+``special``           :doc:`./special`
+``stats``             :doc:`./stats`
 ==================  ========================================
```
mdhaber, line 45:
```suggestion
==================    ========================================
```

mdhaber (commented):
@j-bowhay actually because there were so many `=` to begin with, I think this is all that's needed to satisfy the rules. It is already rendering OK, though.

doc/source/tutorial/index.rst
```diff
@@ -22,27 +22,27 @@ computing domains. These are summarized in the following table, with
 their API reference linked in the Subpackage column, and user guide (if available)
 linked in the Description column:
```
mdhaber, line 23:
```suggestion
their user guide linked in the Description and User Guide column (if available):
```

mdhaber:
This renders as expected:
image

And I think alignment issue in the source code (https://github.com/scipy/scipy/pull/22499#discussion_r1948334505) has been addressed. Thanks @karthik-ganti-2025!

PR #22503, opened 2025-02-10
ENH: optimize.minimize_scalar: add `linewalker` method

chuck-a-knight58:
Linewalker is an optimization method particularly well-suited for nonconvex (multimodal)
functions. The attached paper will show that it is significantly more successful in finding global optima over the existing methods provided.

#### Reference issue
None

Linewalker constructs a smooth surrogate on a set of equally-spaced
grid points by evaluating the true function at a sparse set of
judiciously chosen grid points. At each iteration, the surrogate's
non-tabu local minima and maxima are identified as candidates for
sampling. Tabu search constructs are also used to promote
diversification. If no non-tabu extrema are identified, a simple
exploration step is taken by sampling the midpoint of the largest
unexplored interval. The algorithm continues until a user-defined
function evaluation limit is reached.

Linewalker is particularly well-suited for nonconvex (multimodal)
functions. If the underlying function is known to to be convex,
use a local solver.

#### Additional information
See attached paper providing further discussion and numerical argument for the addition
[linewalker_for_scipy_summary.pdf](https://github.com/user-attachments/files/18735176/linewalker_for_scipy_summary.pdf)


mdhaber:
Thanks for your interest in contributing. Please compare this against the existing [global optimizers](https://docs.scipy.org/doc/scipy/reference/optimize.html#global-optimization). The existing minimizers can be removed from the comparison as they are all for local minimization. Also, many global optimizers are guaranteed to *eventually* find the global optimum on a bounded interval with infinite function evaluations. If it does not already, please ensure that the methodology puts the optimizers on a level playing field in terms of a function evaluation or time budget. Thanks!

andyfaff:
@chuck-a-knight58, thank you for the potential contribution. For contributions of functions of this type we recommend that the contributors ask on the scipy forum first, https://discuss.scientific-python.org/c/contributor/scipy. In this way a discussion can be had before detailed engineering work is carried out (see https://docs.scipy.org/doc/scipy/dev/core-dev/index.html#deciding-on-new-features).

An additional step for adding new global optimizers includes running the proposed solver against all the [global optimizer benchmarks](https://github.com/scipy/scipy/tree/main/benchmarks/benchmarks/go_benchmark_functions). To extend @mdhaber reply, can I ask that you run the solver, with default options, against all those functions?

- the functions will need to be filtered to select those that have `change_dimensionality=True`. The benchmark suite was originally designed to minimize scalar functions that have several variables.
- when you set up instances of the classes you'll need to set the dimensions to 1.
- can you run the linewalker as well as the other [global minimizers]?(https://docs.scipy.org/doc/scipy/reference/optimize.html#global-optimization).  Both with **default** solver options.
- For each class/minimizer combination can you record the percentage of successful minimisations, as well as the average number of function evaluations (for success and fail)?

At that point the strengths of the proposed addition should become clear.

I apologise in advance for the work that this will require. It's more straightforward to run the benchmark suite for functions that have several variables, it wasn't designed for functions with scalar variables.

chuck-a-knight58:
We have completed the requested computational comparison and shown that Linewalker is, indeed, superior for the setting of interest.  There are a couple of important points that apparently were not clear in our original submission:

1) Budget matters. Linewalker is best suited for black-box optimization governed by a strict function evaluation “budget.” Black-box optimization, also known as derivative-free optimization, is a rich field in which one often assumes function evaluations are computationally expensive, e.g., require running an intricate simulator to obtain a value (see, e.g., [Derivative-free optimization methods](https://www.cambridge.org/core/journals/acta-numerica/article/derivativefree-optimization-methods/84479E2B03A9BFFE0F9CD46CF9FCD289)). Thus, the user specifies a strict limit on the number of function evaluations that can be made because the time to perform these evaluations will dominate the overall computation time. We have already demonstrated in previous papers that Linewalker is competitive with state-of-the-art methods.

2) Why were local solvers included? In our original submission, we compared against local solvers because we assumed that non-experts are likely to invoke a scipy.optimize.minimize_scalar method, unaware that a global optimizer is available.  Moreover, SciPy’s global solvers do not treat the maximum function evaluations limit as a strict bound as described below. 

3) Global solver comparison. [Optimization and root finding (scipy.optimize) — SciPy v1.15.1 Manual](https://docs.scipy.org/doc/scipy/reference/optimize.html#global-optimization) lists six global solvers.  Only three of these solvers claim to have a parameter to limit the budget of function evaluations: direct, dual_annealing, and shgo.  Since it makes no sense to compare against solvers that do not enforce a budget limit, we did not consider the remaining three.  For good measure, we also compared against [bayesian-optimization · PyPI](https://pypi.org/project/bayesian-optimization/), which has proven to be quite effective for this problem class.

4)  Scipy’s global solvers enforce a “soft” budget. It became immediately clear that direct, dual_annealing, and shgo do not strictly enforce a function evaluation limit/budget.  For example, when given a limit of 11 function evaluations on the ackley function, dual_annealing and shgo would use well over 150 function evaluations instead.   (To be precise, we set ‘maxfun’ = 11 for dual_annealing and (‘maxev’ = 11 and ‘maxfev’ =11) for shgo.)  This lack of strict control over the budget was frustrating, but we performed what we believe is a worthwhile analysis nonetheless.

5) Linewalker performs best given a restricted function evaluation budget. The attached figure shows the fraction of instances (the same 51 instances shown in our first submission) solved to global optimality given an increasing budget of function evaluations.  Only linewalker and bayesopt strictly adhered to the budget limit.  For the three SciPy global solvers, we accepted the instance to be solved if they returned an optimal solution within a budget of 2*B, where B is the desired budget that linewalker and bayesopt strictly followed.  Thus, for a budget of B=11, the “direct” solver achieves slightly better overall performance than linewalker because it consistently used 11-20 budget evaluations.  This is the only solver-budget pair where one of the three SciPy global solvers outperforms linewalker, which we believe is quite possibly because it used more function evaluations that linewalker was allotted. 

6) Linewalker also returns a surrogate.  Whereas most global solvers only return the best function evaluation and the corresponding minimizer, we also provide a surrogate of the underlying ground truth function.  This additional information can be quite useful.

scipy_global_solver_comparison

scipy/optimize/_optimize.py
```diff
@@ -4168,3 +4171,444 @@ def show_options(solver=None, method=None, disp=True):
...
+    >>> b = np.mean(res['ix_sorted'])
+    >>> print(f"res['ix_sorted'] len, mean = {a}, {b}")
+    >>> a = len(res['x_coord'])
+    >>> b = np.mean(res['x_coord'])
+    >>> print(f"res['x_coord'] len, mean = {a}, {b}")
+    >>> print(f"res['sp'] len, mean = {len(res['sp'])}, {np.mean(res['sp'])}")
+
+    | res['fun'] = np.float64(0.0)
+    | res['x'] = np.float64(1.0)
+    | res['nit'] = 19
```
lucascolley, line 4561: I am not familiar with this syntax but it seems to be failing CI. Have you tried just including the expected outputs after each line?

chuck-a-knight58:
You are correct. Being gathered together as an artifact of a single print statement that was originally there. It just didn’t work well in the documentation. Fixing it now.ThanksOn Apr 4, 2025, at 11:54 AM, Lucas Colley ***@***.***> wrote:﻿
@lucascolley commented on this pull request.



In scipy/optimize/_optimize.py:
> +    >>> res = linewalker(f, brack=(-1, 2))
+    >>> print(f"{res['fun'] = }")
+    >>> print(f"{res['x'] = }")
+    >>> print(f"{res['nit'] = }")
+    >>> print(f"{res['nfev'] = }")
+    >>> print(f"{res['success'] = }")
+    >>> print(f"{res['message'] = }")
+    >>> print(f"{res['f_min_evaluated'] = }")
+    >>> print(f"{res['minimizer_evaluated'] = }")
+    >>> print(f"{res['num_major_iterations'] = }")
+    >>> print(f"{res['num_function_evaluations'] = }")
+    >>> print(f"{res['f_min_predicted'] = }")
+    >>> print(f"{res['ix_arg_min_evaluated'] = }")
+    >>> print(f"{res['grid_size'] = }")
+    >>> print(f"res['fit'] len, mean = {len(res['fit'])}, {np.mean(res['fit'])}")
+    >>> print(f"{res['max_fit'] = }")
+    >>> print(f"{res['min_fit'] = }")
+    >>> print(f"{res['range_fit'] = }")
+    >>> print(f"{res['ix_arg_max_fit'] = }")
+    >>> print(f"{res['ix_arg_min_fit'] = }")
+    >>> print(f"res['ix'] len, mean = {len(res['ix'])}, {np.mean(res['ix'])}")
+    >>> a = len(res['ix_sorted'])
+    >>> b = np.mean(res['ix_sorted'])
+    >>> print(f"res['ix_sorted'] len, mean = {a}, {b}")
+    >>> a = len(res['x_coord'])
+    >>> b = np.mean(res['x_coord'])
+    >>> print(f"res['x_coord'] len, mean = {a}, {b}")
+    >>> print(f"res['sp'] len, mean = {len(res['sp'])}, {np.mean(res['sp'])}")
+
+    | res['fun'] = np.float64(0.0)
+    | res['x'] = np.float64(1.0)
+    | res['nit'] = 19

I am not familiar with this syntax but it seems to be failing CI. Have you tried just including the expected outputs after each line?

chuck-a-knight58:
Looks like I finally passed. Let me know what the next steps are. Thanks

nickodell:
> Looks like I finally passed. Let me know what the next steps are. Thanks

Hello, I made a [brief search of the SciPy discussion forum](https://discuss.scientific-python.org/search?q=linewalker). I don't see a post about linewalker there. Have you made a post there to engage the community and see if they are interested in this type of solver, as @andyfaff suggested?

> Thank you for the potential contribution. For contributions of functions of this type we recommend that the contributors ask on the scipy forum first, https://discuss.scientific-python.org/c/contributor/scipy. In this way a discussion can be had before detailed engineering work is carried out (see https://docs.scipy.org/doc/scipy/dev/core-dev/index.html#deciding-on-new-features).

---

>Since it makes no sense to compare against solvers that do not enforce a budget limit, we did not consider the remaining three.

This seems like a really weird justification for dropping those solvers from your benchmark.

Suppose you set a global function budget of 10 evalutions. The solver ignores this budget and ends up actually using 13. This means that this solution doesn't qualify for having solved the function in 10 evalutions. You could plot a point at x=13, or if you want to be really restrictive, round it up to x=20.

I don't see 'solvers that strictly respect function budgets' as being such a distinct category such that they don't have to compete against the existing solvers.

---

> 6\. Linewalker also returns a surrogate.  Whereas most global solvers only return the best function evaluation and the corresponding minimizer, we also provide a surrogate of the underlying ground truth function.  This additional information can be quite useful.
> 
> 
> scipy_global_solver_comparison

How did you run this benchmark?

I'm going to second @mdhaber's request:

>Also, can you provide the code for running your benchmark set? I'd like to run some tests myself.

fancidev:
I took a look at the referenced paper https://doi.org/10.1007/s11081-023-09879-9 that described the algorithm. Personally I'm interested to find out (1) what accuracy guarantee the optimizer provides, and (2) what sort of function is the optimizer most suitable for.

For (1), Theorem 3 of the reference paper states that if the true function is Lipschitz continuous, then the number of points at which the function needs to be evaluated to guarantee a given accuracy is bounded in the obvious manner. I think this is fine, though there may exist alternative methods that provide better bounds by making further assumptions about the smoothness of the function. Anyway, these bounds may be far too large to be practically useful for a black-box function. So for this point, I reckon the proposed method is on par with alternatives.

For (2), I didn't find a discussion in the referenced paper about that, though the paper does provide extensive comparison on benchmarks. From Figure 5 of that paper, it seems that NOMAD is quite competitive among the candidates. I have not personally encountered a case where I needed to optimize a totally black box function. And I think some idea about the behavior of the function is essential for trusting the output of the optimizer. That's why I'm curious to know what kind of functions the proposed method works best with.

nickodell:
I feel conflicted about this PR. chuck-a-knight58 has clearly put quite a lot of effort here into writing an innovative solver, and making it performant. He has also made an effort to continuously solve merge conflicts on the PR. On the other hand, I feel that there are three concerns that we have not gotten any closer to solving.

* The methods improves on certain benchmarks. However, SciPy developers do not have the ability to independently reproduce these benchmarks, as they rely on source code that has not been shared.
* Is there community interest in using the solver? For example, papers that cite the method or community engagement.
* Is it better than existing global methods?

I propose that we close this PR on Feb 20th if no progress is made here. I would like to avoid leaving this PR in a limbo state - we should make a decision either that it is a good addition to the existing global solvers, or that we should encourage chuck-a-knight58 to make his own package, to allow interested users to try this optimization method.

chuck-a-knight58:
Sorry for the lack of response but was forced into retirement and really don’t have the cycles to mess with it. I don’t have access tot he code anymore. It’s probably best to kill it.

nickodell:
> Sorry for the lack of response but was forced into retirement and really don’t have the cycles to mess with it. I don’t have access tot he code anymore. It’s probably best to kill it.

Sorry to hear about that. Thank you for the PR - we appreciate the contribution.

Issue #22487, opened 2025-02-07
BUG: wheel build failure on MacOS due to recent updates in `shortest_path`


dschmitz89:
### Describe your issue.

A recent wheel build run un #22350 unveiled a compilation failure likely introduced in #20366 : see [here](https://github.com/scipy/scipy/actions/runs/13182828693/job/36797727498?pr=22350#step:8:926). The issue seems to be something related to `std::lower_bound`. See the screenshot below.

![Image](https://github.com/user-attachments/assets/66b2c387-5b6d-45ea-aa55-e527e5d0d853)

CC @tsery-ns @Kai-Striega 

### Reproducing Code Example

```python
n/a
```

### Error message

```shell
n/a
```

### SciPy/NumPy/Python version and system information

```shell
main
```

Kai-Striega:
Thanks I'll look into that over the weekend

Kai-Striega:
I've spent a couple of hours digging into this, the bug is pretty obvious from the error message, ``__lower_bound`` wants ``int (&)(...)``  and we're giving it ``int (*)(...)``. I suspect that the suggested fix (``dereference the argument with *``) will work, but isn't a great solution. What's interesting is that this only occurs for mac 10.13, not for 14.0 or 12.3. I suspect that these include newer compilers which are smart enough to figure out that you could dereference it automagically.

I'm not sure if that's the right solution. Or what to do about it if it is.  

One option would be to remove ``lower_bound`` and replace it with our own implementation. I don't agree with this at all. But I'm not sure what the alternatives are.

tsery-ns:
> I suspect that the suggested fix (dereference the argument with *) will work, but isn't a great solution.

Why not? We can use the Cython [derefernce](https://cython.readthedocs.io/en/latest/src/userguide/wrapping_CPlusPlus.html#c-operators-not-compatible-with-python-syntax) operator.

Let me know if there any way I can help, I don't work on a mac so my ability to verify the problem is solved would be limited.

rgommers:
I'll open a PR with a bump to the minimum macOS and Clang version to fix the problem with wheel builds.

@tsery-ns it would be useful to verify that the code Cython generates is valid in C++17. It looks like there were some changes to `std::lower_bound` in C++20, and if that is what's being generated by Cython, it may require a fix in Cython or we should avoid using `libcpp.algorithm.lower_bound` in our Cython code.

rgommers:
[...]

PR #22504, opened 2025-02-10
BLD: bump min version of Clang to 15.0, and macOS min version to 10.14

rgommers:
Addresses the wheel build failures reported for macOS x86-64 in gh-22487.

macOS 10.14 is from 2018, so is still a very loose lower bound, we don't expect issues as a result.

The Clang >=15 is probably also reasonable, since it's rare to see a version below 16 being used. Targeting macOS 10.13 causes cibuildwheel to pull in XCode 14.1 with Clang 14.0, while macOS 10.14 pulls in XCode 15.2 and Clang 15.0.

I verified the fix on my fork: https://github.com/rgommers/scipy/actions/runs/13241617566.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -83,7 +83,7 @@ jobs:
         - [ubuntu-22.04, musllinux, x86_64, "", ""]
         - [ubuntu-24.04-arm, manylinux, aarch64, "", ""]
         - [ubuntu-24.04-arm, musllinux, aarch64, "", ""]
-        - [macos-13, macosx, x86_64, openblas, "10.13"]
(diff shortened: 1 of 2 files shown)


andyfaff:
This looks ok to me. Is it worth running on the entire macOS wheel matrix? Otherwise I think this is mergeable.

rgommers:
> Is it worth running on the entire macOS wheel matrix? 

The behavior is orthogonal to Python version, and I tested with minimum and maximum versions. I can push a `[wheel build]` commit though if you prefer.

andyfaff:
Let's just merge and see if problems bubble up.

PR #22505, opened 2025-02-10
ENH: stats.quantile: add discontinuous (HF 1-3) and Harrell-Davis methods; add `marray` support

mdhaber:
#### Reference issue
Follow-up to gh-feature

This adds the following capabilities to the new `scipy.stats.quantile`:
- the discontinuous estimators (1-3) from [1]: `inverted_cdf`, `averaged_inverted_cdf`, and `closest_observation`
- the Harrell-Davis estimator (as `method='harrell-davis') to replace `mstats.hdmedian` and `mstats.hdquantile`
- support for `marray`s + associated tests

### DIFF
--- a/scipy/stats/_quantile.py
+++ b/scipy/stats/_quantile.py
@@ -3,6 +3,7 @@ from scipy.special import betainc
 from scipy._lib._array_api import (xp_take_along_axis, xp_default_dtype,
                                    xp_ravel, array_namespace)
 from scipy.stats._axis_nan_policy import _broadcast_arrays, _contains_nan
+from scipy.stats._stats_py import _length_nonmasked
 
 
 def _quantile_iv(x, p, method, axis, nan_policy, keepdims):
@@ -68,7 +69,8 @@ def _quantile_iv(x, p, method, axis, nan_policy, keepdims):
     y = xp.moveaxis(y, axis, -1)
     p = xp.moveaxis(p, axis, -1)
 
-    n = xp.astype(xp.asarray(y.shape[-1]), dtype)
+    n = _length_nonmasked(y, -1, xp=xp, keepdims=True)
+    n = xp.asarray(n, dtype=dtype)
     if contains_nans:
         nans = xp.isnan(y)
 
--- a/scipy/stats/_quantile.py
+++ b/scipy/stats/_quantile.py
@@ -33,7 +33,8 @@ def _quantile_iv(x, p, method, axis, nan_policy, keepdims):
         raise ValueError(message)
     axis = int(axis)
 
-    methods = {'hazen', 'interpolated_inverted_cdf', 'linear',
+    methods = {'inverted_cdf', 'averaged_inverted_cdf', 'closest_observation',
+               'hazen', 'interpolated_inverted_cdf', 'linear',
                'median_unbiased', 'normal_unbiased', 'weibull',
                'harrell-davis'}
     if method not in methods:
@@ -117,6 +118,9 @@ def quantile(x, p, *, method='linear', axis=0, nan_policy='propagate', keepdims=
         The method to use for estimating the quantile.
         The available options, numbered as they appear in [1]_, are:
 
+        1. 'inverted_cdf'
+        2. 'averaged_inverted_cdf'
+        3. 'closest_observation'
         4. 'interpolated_inverted_cdf'
         5. 'hazen'
         6. 'weibull'
@@ -201,14 +205,14 @@ def quantile(x, p, *, method='linear', axis=0, nan_policy='propagate', keepdims=
 
     Note that indices ``j`` and ``j + 1`` are clipped to the range ``0`` to
     ``n - 1`` when the results of the formula would be outside the allowed
-    range of non-negative indices. The ``- 1`` in the formulas for ``j`` and
+    range of non-negative indices. The ``-1`` in the formulas for ``j`` and
     ``g`` accounts for Python's 0-based indexing.
 
     The table above includes only the estimators from H&F that are continuous
-    functions of probability `p` (estimators 4-9). NumPy also provides the
+    functions of probability `p` (estimators 4-9). SciPy also provides the
     three discontinuous estimators from H&F (estimators 1-3), where ``j`` is
-    defined as above, ``m`` is defined as follows, and ``g`` is a function
-    of the real-valued ``index = p*n + m - 1`` and ``j``.
+    defined as above, ``m`` is defined as follows, and ``g`` is ``0`` when
+    ``index = p*n + m - 1`` is less than ``0`` and otherwise is defined below.
 
     1. ``inverted_cdf``: ``m = 0`` and ``g = int(index - j > 0)``
     2. ``averaged_inverted_cdf``: ``m = 0`` and
@@ -268,7 +272,8 @@ def quantile(x, p, *, method='linear', axis=0, nan_policy='propagate', keepdims=
     temp = _quantile_iv(x, p, method, axis, nan_policy, keepdims)
     y, p, method, axis, nan_policy, keepdims, n, axis_none, ndim, p_mask, xp = temp
 
-    if method in {'hazen', 'interpolated_inverted_cdf', 'linear',
+    if method in {'inverted_cdf', 'averaged_inverted_cdf', 'closest_observation',
+                  'hazen', 'interpolated_inverted_cdf', 'linear',
                   'median_unbiased', 'normal_unbiased', 'weibull'}:
         res = _quantile_hf(y, p, n, method, xp)
     elif method in {'harrell-davis'}:
@@ -293,12 +298,22 @@ def quantile(x, p, *, method='linear', axis=0, nan_policy='propagate', keepdims=
 
 
 def _quantile_hf(y, p, n, method, xp):
-    methods = dict(interpolated_inverted_cdf=0, hazen=0.5, weibull=p, linear=1 - p,
-                   median_unbiased=p / 3 + 1 / 3, normal_unbiased=p / 4 + 3 / 8)
-    m = methods[method]
-    jg = p * n + m - 1
+    ms = dict(inverted_cdf=0, averaged_inverted_cdf=0, closest_observation=-0.5,
+              interpolated_inverted_cdf=0, hazen=0.5, weibull=p, linear=1 - p,
+              median_unbiased=p / 3 + 1 / 3, normal_unbiased=p / 4 + 3 / 8)
+    m = ms[method]
+    jg = p*n + m - 1
     j = jg // 1
     g = jg % 1
+    if method == 'inverted_cdf':
+        g = xp.astype((g > 0), jg.dtype)
+    elif method == 'averaged_inverted_cdf':
+        g = (1 + xp.astype((g > 0), jg.dtype)) / 2
+    elif method == 'closest_observation':
+        g = (1 - xp.astype((g == 0) & (j % 2 == 1), jg.dtype))
+    if method in {'inverted_cdf', 'averaged_inverted_cdf', 'closest_observation'}:
+        g = xp.asarray(g)
+        g[jg < 0] = 0
     j = xp.clip(j, 0., n - 1)
     jp1 = xp.clip(j + 1, 0., n - 1)
     j = xp.astype(j, xp.int64)
--- a/scipy/stats/_quantile.py
+++ b/scipy/stats/_quantile.py
@@ -1,4 +1,5 @@
 import numpy as np
+from scipy.special import betainc
 from scipy._lib._array_api import (xp_take_along_axis, xp_default_dtype,
                                    xp_ravel, array_namespace)
 from scipy.stats._axis_nan_policy import _broadcast_arrays, _contains_nan
@@ -32,7 +33,8 @@ def _quantile_iv(x, p, method, axis, nan_policy, keepdims):
     axis = int(axis)
 
     methods = {'hazen', 'interpolated_inverted_cdf', 'linear',
-               'median_unbiased', 'normal_unbiased', 'weibull'}
+               'median_unbiased', 'normal_unbiased', 'weibull',
+               'harrell-davis'}
     if method not in methods:
         message = f"`method` must be one of {methods}"
         raise ValueError(message)
@@ -44,6 +46,8 @@ def _quantile_iv(x, p, method, axis, nan_policy, keepdims):
         raise ValueError(message)
 
     dtype = xp.result_type(p, x)
+    x = xp.astype(x, dtype, copy=False)
+    p = xp.astype(p, dtype, copy=False)
 
     # If data has length zero along `axis`, the result will be an array of NaNs just
     # as if the data had length 1 along axis and were filled with NaNs. This is treated
@@ -83,6 +87,9 @@ def _quantile_iv(x, p, method, axis, nan_policy, keepdims):
         if xp.any(nan_out):
             y = xp.asarray(y, copy=True)  # ensure writable
             y[nan_out] = np.nan
+        elif xp.any(nans) and method == 'harrell-davis':
+            y = xp.asarray(y, copy=True)  # ensure writable
+            y[nans] = 0
 
     p_mask = (p > 1) | (p < 0) | xp.isnan(p)
     if xp.any(p_mask):
@@ -115,6 +122,8 @@ def quantile(x, p, *, method='linear', axis=0, nan_policy='propagate', keepdims=
         8. 'median_unbiased'
         9. 'normal_unbiased'
 
+        'harrell-davis' is also available to compute the quantile estimate
+        according to [2]_.
         See Notes for details.
     axis : int or None, default: 0
         Axis along which the quantiles are computed.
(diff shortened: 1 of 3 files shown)


mdhaber:
GPU failures for `torch` with Harrell-Davis method are expected because Torch doesn't have `betainc` and currently we are not allowed to transfer between GPU and CPU; we'll have to skip this case.

To address platform-dependent failures in `test_transition`, maybe I should test with `n=8`. Small rounding issues when multiplying `p * n` can affect which side of the integer the result is on.

mdhaber:
@JohannesBuchner one thing to note about [`scipy.stats.quantile`](https://scipy.github.io/devdocs/reference/generated/scipy.stats.quantile.html#scipy.stats.quantile), the `mquantiles` replacement here, is that the `method` argument currently only supports named interpolation options (like NumPy does) as opposed to having `alphap` and `betap` arguments. Is that going to work for you? I've considered allowing `method` to accept a tuple of floats, but that makes the interface a bit messier.

Another thing is that we haven't added *NumPy* masked array support right now, but support for [`marray`](https://github.com/mdhaber/marray)s. There are a few workarounds that would be possible in that case:
- Convert NumPy masked arrays to `marray`s before passing them in, and convert the result back (if need be) 
- replace masked elements with NaNs and use `nan_policy='omit'` with this function (or `np.nanquantile`, which is probably much faster)

Would either of those work for you, or do we need to consider accepting NumPy masked arrays? In that case, would it be ok for the output to be a regular NumPy array (since the output always has an all-`False` mask unless `p` is masked or slices have no elements)?

JohannesBuchner:
For my code, you can see https://github.com/search?q=owner%3AJohannesBuchner%20mquantiles&type=code&p=1 I only use quantiles (single number, list or array) and axis parameters in addition to the array of values.

mdhaber:
Are there ever NaNs in either argument? If so, what do you want to happen?

JohannesBuchner:
Yes, there may be a tiny fraction <1% so that when I compute e.g. 10% quantiles I expect they should be not considered.

Regular array.

mdhaber:
> Regular array

Then it sounds like you don't need `mquantiles` anyway. You can use `np.quantile` if you want NaNs to propagate and `np.nanquantile` if you want them ignored.

JohannesBuchner:
From the documentation it's not clear which np.quantile method is equivalent to the default mquantile behaviour, which seemed better when I chose it. A mapping of the two in either or both documentation pages would help.

Am I supposed to write `np.nanquantiles(method='normal_unbiased')` every time now, that seems quite long for a reasonable default behaviour?

The np.quantiles documentation page does not describe how nans are treated at all?

My code and me starting to use this function is in part from 2009, so I am not sure if `np.quantile` from then was already as powerful as it is now.

In general, in the python ecosystem it is quite difficult to build long-term projects (10+ years) and as a maintainer keep attention on scientific improvements, because every 6 months something breaks or needs to be updated (python2->3, python<3.5 deprecation, pip / packaging changes, Travis CI went out of service, numpy >2 and <2 is a mess right now where each pip install either flips backs or forth; and my cython packages broke).

I know it is annoying to keep legacy code, but please consider this.

mdhaber:
> Am I supposed to write `np.nanquantiles(method='normal_unbiased')` every time now, that seems quite long for a reasonable default behaviour?

Well, NumPy has reasonable defaults for some use cases. `method='normal_unbiased'` gets you closer to the settings you preferred (although it is not exactly the same). And whether you use SciPy or NumPy, you have to specify that you want to ignore NaNs. Note that `mquantiles` doesn't ignore NaNs, and doesn't have an option to do so:

```python3
import numpy as np
from scipy import stats
x = np.asarray([1, 2, 3, 4, 5, np.nan])
stats.mstats.mquantiles(x)  # array([1.95, 3.5 ,  nan])
```


> My code and me starting to use this function is in part from 2009

I see. You're right, `np.quantile` did not have these options back then.

Yes, I understand there have been a lot of transitions, so we'll be sure to give plenty of notice and a suitable replacement. Thanks for the insight into your use case.

scipy/stats/_quantile.py
```diff
@@ -257,3 +295,43 @@ def quantile(x, p, *, method='linear', axis=0, nan_policy='propagate', keepdims=
...
+        g = (1 + xp.astype((g > 0), jg.dtype)) / 2
+    elif method == 'closest_observation':
+        g = (1 - xp.astype((g == 0) & (j % 2 == 1), jg.dtype))
+    if method in {'inverted_cdf', 'averaged_inverted_cdf', 'closest_observation'}:
+        g = xp.asarray(g)
+        g[jg < 0] = 0
+    j = xp.clip(j, 0., n - 1)
+    jp1 = xp.clip(j + 1, 0., n - 1)
+    j = xp.astype(j, xp.int64)
+    jp1 = xp.astype(jp1, xp.int64)
```
ev-br, line 320: why is this needed?

In general, it's not very easy to follow which variables are integer and which are floating-point. Consider adding some code comments to this effect.

scipy/stats/_quantile.py
```diff
@@ -257,3 +295,43 @@ def quantile(x, p, *, method='linear', axis=0, nan_policy='propagate', keepdims=
...
+
+    return ((1 - g) * xp_take_along_axis(y, j, axis=-1)
+            + g * xp_take_along_axis(y, jp1, axis=-1))
+
+
+def _quantile_hd(y, p, n, xp):
+    p = xp.moveaxis(p, -1, 0)[..., np.newaxis]
```
ev-br, line 327: Nit: xp.newaxis or just None

scipy/stats/_quantile.py
```diff
@@ -257,3 +295,43 @@ def quantile(x, p, *, method='linear', axis=0, nan_policy='propagate', keepdims=
...
+    a = p * (n + 1)
+    b = (1 - p) * (n + 1)
+    i = xp.arange(y.shape[-1] + 1)
+    i = xp.astype(i, y.dtype)
```
ev-br, line 331: Nit: `xp.arange(y.shape[-1] + 1, dtype=y.dtype)`

scipy/stats/_quantile.py
```diff
@@ -257,3 +295,43 @@ def quantile(x, p, *, method='linear', axis=0, nan_policy='propagate', keepdims=
...
+    # There will be edge case bugs with a == 0 or b == 0 due to gh-8411
+    w = betainc(a, b, i / n)
+    w = w[..., 1:] - w[..., :-1]
+    w[xp.isnan(w)] = 0
```
ev-br, line 335: `xp.where` to avoid boolean indexing?

scipy/stats/_quantile.py
```diff
@@ -257,3 +295,43 @@ def quantile(x, p, *, method='linear', axis=0, nan_policy='propagate', keepdims=
...
+    res = xp.vecdot(w, y, axis=-1)
+    return xp.moveaxis(res, 0, -1)
```
ev-br, line 337: consider adding some explanatory comments on the axis handling

scipy/stats/tests/test_quantile.py
```diff
@@ -95,21 +103,46 @@ def test_against_numpy(self, method, shape_x, shape_p, axis, xp):
...
         x[mask] = np.nan
         if not keepdims:
             p = np.mean(p, axis=axis, keepdims=True)
 
+        # inject p = 0 and p = 1 to test edge cases
+        # Currently would fail with CuPy/JAX (cupy/cupy#8934, jax-ml/jax#21900)
+        if is_numpy(xp):
+            p0 = p.ravel()
+            p0[1] = 0.
+            p0[-2] = 1.
```
ev-br, line 125: consider splitting into a separate test with corresponding `@xfail_xp_backends` decorators.

scipy/stats/tests/test_quantile.py
```diff
@@ -95,21 +103,46 @@ def test_against_numpy(self, method, shape_x, shape_p, axis, xp):
...
         dtype = getattr(xp, dtype)
-        kwargs = dict(axis=axis, keepdims=keepdims, nan_policy=nan_policy)
+
+        if nan_policy == 'marray':
+            if method == 'harrell-davis':
+                pytest.skip("Needs gh-22490")
+            if is_torch(xp):
+                pytest.skip("sum_cpu not implemented for UInt64, see "
+                            "data-apis/array-api-compat#242")
+            marray = pytest.importorskip('marray')
```
ev-br, line 135: Had to google it. So this either imports `marray` or skips the test if `marray` is not available, correct? It'd be way more readable IMO to use a "straitghforward" way with a top-level import and a module-level variable.
```
try:
   import marray
   HAVE_MARRAY = True
except ModuleNotFoundError:
   HAVE_MARRAY=False
```

ev-br (commented):
Looks good modulo several minor comments.

scipy/stats/_quantile.py
```diff
@@ -44,6 +48,8 @@ def _quantile_iv(x, p, method, axis, nan_policy, keepdims):
         raise ValueError(message)
 
     dtype = xp.result_type(p, x)
+    x = xp.astype(x, dtype, copy=False)
```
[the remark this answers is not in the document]
mdhaber: Sorry? Line 60 does not modify `x`; it replaces it. After line 62, which copies `x` with the `sort`, we don't touch `x` again, right?

scipy/stats/_quantile.py
```diff
@@ -190,9 +205,34 @@ def quantile(x, p, *, method='linear', axis=0, nan_policy='propagate', keepdims=
 
     Note that indices ``j`` and ``j + 1`` are clipped to the range ``0`` to
     ``n - 1`` when the results of the formula would be outside the allowed
-    range of non-negative indices. The ``- 1`` in the formulas for ``j`` and
+    range of non-negative indices. The ``-1`` in the formulas for ``j`` and
     ``g`` accounts for Python's 0-based indexing.
 
+    The table above includes only the estimators from H&F that are continuous
```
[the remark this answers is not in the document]
mdhaber: I think I was just following what I saw in NumPy before I rewrote the NumPy docstring. We should just use `[1]_` instead of H&F throughout, except maybe in the table because Sphinx probably can't handle something as terribly complex as a reference inside a table.

scipy/stats/_quantile.py
```diff
@@ -222,27 +262,25 @@ def quantile(x, p, *, method='linear', axis=0, nan_policy='propagate', keepdims=
     .. [1] R. J. Hyndman and Y. Fan,
        "Sample quantiles in statistical packages,"
        The American Statistician, 50(4), pp. 361-365, 1996
+    .. [2] Harrell, Frank E., and C. E. Davis.
```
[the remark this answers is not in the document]
mdhaber: You mean the `(1, 2)`? Those are links to the places in the document that the reference appears. It is just what Sphinx does, I think. If you don't see it in other places, it may be because they don't cite the references in the text.

scipy/stats/_quantile.py
```diff
@@ -257,3 +295,43 @@ def quantile(x, p, *, method='linear', axis=0, nan_policy='propagate', keepdims=
...
+
+
+def _quantile_hf(y, p, n, method, xp):
+    ms = dict(inverted_cdf=0, averaged_inverted_cdf=0, closest_observation=-0.5,
+              interpolated_inverted_cdf=0, hazen=0.5, weibull=p, linear=1 - p,
+              median_unbiased=p / 3 + 1 / 3, normal_unbiased=p / 4 + 3 / 8)
+    m = ms[method]
+    jg = p*n + m - 1
+    j = jg // 1
+    g = jg % 1
```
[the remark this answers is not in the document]
mdhaber: Oh, I didn't even know about it. But it's not in the array api standard, so I can't use.

scipy/stats/_quantile.py
mdhaber: All variables are floating point until this line. (Even `n` because the array API standard doesn't support operations between float and integer arrays. And yeah `n` is an array in general because we have different numbers of non-NaN, non-masked elements in each slice.)

Right below, we use these in `take_along_axis`, so we have to ensure integer dtype. I could try to put these conversions into those lines, if that would help? Then there's no question that everything is float throughout.

scipy/stats/_quantile.py
mdhaber: I'll use `xpx.at.set`. I thought I did...

scipy/stats/tests/test_quantile.py
mdhaber: Yup, I did consider it and decided to go this way. When those linked issues are resolved, we'll just take this part out of the `if is_numpy` rather than deduplicating tests.

scipy/stats/tests/test_quantile.py
mdhaber: Yes, that's what it does.

In fact I had a `try`-`except` for this purpose in the original `marray` PR, but it was recommended I use `importorskip` instead.

I didn't know about it either, but it seems extremely handy to me. Why have all that at the top, then two lines of `if not HAVE_MARRAY: pytest.skip()` here when it can all just be one line?

My guess is that the second or third time you encounter it, you'll find it more readable than the alternative.

mdhaber:
Thanks @ev-br! Adjusted most of what I could; please see comments about the rest. Diff is [here](https://github.com/scipy/scipy/pull/22505/commits/ea61aa20be216c0e9253ca39c1214d95db0a9934).

scipy/stats/tests/test_quantile.py
```diff
@@ -144,3 +178,14 @@ def test_size_0(self, axis, keepdims, xp):
...
         res = stats.quantile(xp.zeros(tuple(shape)), 0.5, axis=axis, keepdims=keepdims)
         assert res.shape == tuple(out_shape)
+
+    @pytest.mark.parametrize('method',
+        ['inverted_cdf', 'averaged_inverted_cdf', 'closest_observation'])
+    def test_transition(self, method, xp):
+        # test that values of discontinuous estimators are correct when
+        # p*n + m - 1 is integral.
+        x = np.arange(8., dtype=np.float64)
+        p = np.arange(0, 1.125, 0.125)
```
mdhaber, line 188: Oops.
```suggestion
        p = np.arange(0, 1.0625, 0.0625)
```
But let's let test run.

scipy/stats/tests/test_quantile.py
```diff
@@ -95,21 +103,47 @@ def test_against_numpy(self, method, shape_x, shape_p, axis, xp):
 
     @pytest.mark.parametrize('axis', [0, 1])
     @pytest.mark.parametrize('keepdims', [False, True])
-    @pytest.mark.parametrize('nan_policy', ['omit', 'propagate'])
+    @pytest.mark.parametrize('nan_policy', ['omit', 'propagate', 'marray'])
     @pytest.mark.parametrize('dtype', ['float32', 'float64'])
-    def test_against_reference(self, axis, keepdims, nan_policy, dtype, xp):
+    @pytest.mark.parametrize('method', ['linear', 'harrell-davis'])
+    def test_against_reference(self, axis, keepdims, nan_policy, dtype, method, xp):
```
mdhaber, line 109: Oops. Make this CPU-only because torch doesn't have `betainc` and we can't move data from GPU to CPU.

mdhaber:
Tests are failing due to a bug in NumPy's `closest_observation` method, only recently fixed in numpy/numpy#26959. I'll just skip the test for old NumPy.

mdhaber:
OK, remaining CI failures are just gh-22534. I don't have anything to add right now.

scipy/stats/_quantile.py
```diff
@@ -190,9 +206,34 @@ def quantile(x, p, *, method='linear', axis=0, nan_policy='propagate', keepdims=
...
+
+    .. math::
+
+        w_{n, i} = I_{i/n}(a, b) - I_{(i - 1)/n}(a, b)
+
+    where :math:`n` is the number of elements in the sample,
+    :math:`i` are the indices :math:`1, 2, ..., n-1, n` of the sorted elements,
+    :math:`a = p (n + 1)`, :math:`b = (1 - p)(n + 1)`,
+    :math:`p` is the probability of the quantile, and
+    :math:`I` is the regularized, lower incomplete beta function (`special.betainc`).
```
ev-br, line 235: `special.betainc` does not unfortunately resolve into a hyperlink. Probably needs a fully qualified name, `scipy.special.betainc` or a leading `~` symbol (am never sure how taht one works in sphinx).

scipy/stats/tests/test_quantile.py
ev-br: Well, my concern is obviously not about readability. I'm concerned about hiding optional dependencies  deep in test functions. 
That said, if you as a stats maintainer find it good enough, great,  it's not worth splitting hairs over.

ev-br (approved):
LGTM now.

scipy/stats/_quantile.py
mdhaber, line 235:
```suggestion
    :math:`I` is the regularized, lower incomplete beta function
    (`scipy.special.betainc`).
```

mdhaber:
[Rendering](https://output.circle-artifacts.com/output/job/54d86b68-82eb-4e88-935d-680b98359caa/artifacts/0/html/reference/generated/scipy.stats.quantile.html#scipy.stats.quantile) correctly now:
image

ev-br:
Merged to get the ball rolling. Thanks @mdhaber

PR #22506, opened 2025-02-10
ENH: special.btdtria: migrate to boost

dschmitz89:
#### Reference issue
Towards #21966 

Migrates `btdtria` from cdflib to boost's `ibeta_inva`.

**Benchmark results**
btdtria_relative_error_histogram

<p>

```python

from scipy.special import btdtria, exp10
import numpy as np
from mpmath import mp

BRANCH = "main"

mp.dps = 50

@np.vectorize
def btdtr_mpmath(a, b, x):
    a, b, x = map(mp.mpf, (a, b, x, ))
    try:
        res = mp.betainc(a, b, x2=x, regularized=True)
        return float(res)
    except Exception:
        return np.nan

    """
    """
    return rel_error

"""
rng = np.random.default_rng(123456789)

a_test = exp10(rng.uniform(-4, 3, 30))
b_test = exp10(rng.uniform(-2, 5, 30))
x_test = exp10(rng.uniform(-5, 7, 30))

a, b, x = np.meshgrid(a_test, b_test, x_test)
a, b, x = a.ravel(), b.ravel(), x.ravel()

betainc_ref = btdtr_mpmath(a, b, x) #np.load("betainc_mpmath_ref.npy")# btdtr_mpmath(a, b, x)

# filter out 0 and 1 values for betainc_ref
mask = (betainc_ref > 0) & (betainc_ref < 1)
betainc_ref = betainc_ref[mask]
b = b[mask]
x = x[mask]
a = a[mask]
btdtria_result = btdtria(betainc_ref, b, x)

"""
#np.save(f"btdtria_{BRANCH}_test_results", btdtria_result)
#np.save("btdtria_a_ref.npy", a)
#np.save("betainc_mpmath_ref.npy", betainc_ref)
btdtria_result_pr = np.load(f"btdtria_PR_test_results.npy")
btdtria_result_main = np.load(f"btdtria_main_test_results.npy")
a = np.load("btdtria_a_ref.npy")

rel_error_pr = calculate_rel_error(btdtria_result_pr, a)
rel_error_main = calculate_rel_error(btdtria_result_main, a)

plt.figure(figsize=(10, 6))
plt.hist(np.log10(rel_error_pr), bins=int(np.sqrt(len(rel_error_pr))), alpha=0.5, label='btdtria PR')
plt.hist(np.log10(rel_error_main), bins=int(np.sqrt(len(rel_error_main))), alpha=0.5, label='btdtria main')
plt.ylabel('Frequency')
plt.title('Histogram of Relative Errors for btdtria')
plt.legend()
plt.grid()
plt.savefig("btdtria_relative_error_histogram.png")

``` 

</p>
</details>


scipy/special/tests/test_basic.py
```diff
@@ -1384,11 +1384,13 @@ def test_a1_b1(self):
...
          # gh-12796:
          (4, 99997, 0.0001947841578892121, 0.999995)])
-    def test_betainc_betaincinv(self, a, b, x, p):
+    def test_betainc_betaincinv_betaincinva(self, a, b, x, p):
         p1 = special.betainc(a, b, x)
         assert_allclose(p1, p, rtol=1e-15)
         x1 = special.betaincinv(a, b, p)
         assert_allclose(x1, x, rtol=5e-13)
+        a1 = special.betaincinva(p, b, x)
+        assert_allclose(a1, a, rtol=1e-13)
```
dschmitz89, line 1393: Two of these test cases fail with `btdtria`.

dschmitz89:
I am not familiar with mypy, any pointers to a fix for the failure would be appreciated.

dschmitz89:
> Are you OK with decoupling the rename from the improvements? I'd like to see a rename of opaquely named cdflib functions too, but would prefer that a naming scheme be thought of and discussed for all of the functions together, rather than renaming in piecemeal fashion one function at a time.

A thought out naming scheme is music in my ears. Definitely OK with that. Will update this PR accordingly.

scipy/special/tests/test_basic.py
```diff
@@ -178,9 +178,6 @@ def test_berp(self):
     def test_besselpoly(self):
         assert_equal(cephes.besselpoly(0,0,0),1.0)
 
-    def test_btdtria(self):
-        assert_equal(cephes.btdtria(1,1,1),5.0)
-
```
dschmitz89, line 183: I do not think that the expected value of 5 makes sense, so I deleted this test.

dschmitz89:
Sorry about the noise here, something went wrong trying to rewrite commit history.

dschmitz89:
Added a benchmark to the opening post. Again boost outperforms cdflib by a large margin.

ilayn:
I guess, we should have just switched to boost directly instead of translating them.

dschmitz89:
> I guess, we should have just switched to boost directly instead of translating them. 

Well, the migration to boost is also a major undertaking that will likely take until the end of the year at least. Without the rewrite we would still have the messy build system, so your translation was not in vain in my opinion.

scipy/special/boost_special_functions.h
```diff
@@ -264,6 +264,48 @@ ibeta_inv_double(double a, double b, double p)
...
+        y = 0;
+    } catch (...) {
+        sf_error("btdtria", SF_ERROR_OTHER, NULL);
+        y = NAN;
+    }
+    return y;
+}
+
+float
+ibeta_inva_float(float p, float b, float x)
```
ilayn, line 299: Does this templating mechanism handle `float`, `double`, `float` input combos?

dschmitz89: Puh, to be honest, I am doing a lot of monkey see, monkey do programming here. Not sure if our ufunc generation mechanism would supports this. More important question: should we support mixed precisions? I have never encountered something like this at least.

ilayn: Not that I investigated but in case someone sends accidentally numpy scalars of different type what is supposed to happen in the wrapper, fail or upcast?

dschmitz89:
I will merge this within a week if there are no further comments. Boost is clearly better and this is a rather niche function.

steppi:
> I will merge this within a week if there are no further comments. Boost is clearly better and this is a rather niche function.

I finally have a chance to look at this today. Sorry for the delay.

scipy/special/boost_special_functions.h
steppi: The ufunc machinery handles this. If there is no implementation for the given types, it searches for types the inputs can be safely cast to. See https://numpy.org/doc/stable/user/basics.ufuncs.html#type-casting-rules.

scipy/special/tests/test_basic.py
steppi: Right. All positive finite values of `a` are valid results here, since the cdf at the right endpoint of the support will always be 1. The cdflib implementation was returning the default initial guess for the root basically. It looks like Boost returns the smallest normal number here

```
In [3]: btdtria(1, 1, 1)
Out[3]: np.float64(2.2250738585072014e-308)
```

which seems reasonable enough. I was thinking indeterminate cases like this should either return NaN, or there should be a convention for which solution to choose. If we find Boost does have an intended convention, or decide to impose a convention, then it makes sense to have this test case (with appropriate desired result), but for now, deleting seems fine.

steppi (approved):
This looks good @dschmitz89. Everything seems in order. Many apologies for the delay. I'll try to be more timely getting to your PRs from now on.

dschmitz89:
> This looks good @dschmitz89. Everything seems in order. Many apologies for the delay. I'll try to be more timely getting to your PRs from now on.

No worries and thanks for merging. Hopefully we can also find more reviewers for `special`. For PRs that do not implement special function math but simply wrap boost we should be able to rely on benchmarks and tests.

PR #22507, opened 2025-02-11
BENCH: make Benchmark.change_dimensionality a class variable

andyfaff:


andyfaff:
Making this attribute into a class variable makes it easier to filter which problems are suitable for different dimensions.

PR #22493, opened 2025-02-08
DOC: Add a docstring to OptimizeWarning

star1327p:
Add a docstring to OptimizeWarning:    
"""General warning class for ``scipy.optimize``. It indicates that   
the optimization algorithm did not successfully converge."""

In the `scipy.optimize` page, the OptimizeWarning explanation is currently blank.
https://scipy.github.io/devdocs/reference/optimize.html

PR #22509, opened 2025-02-11
DOC: sparse.linalg: add explanation for `MatrixRankWarning`

star1327p:

Add explanation for MatrixRankWarning. The definition "Warning on exactly singular matrices." came from `scipy/scipy/sparse/linalg/__init__.py`.

#### Additional information

Related to PR #22493.


scipy/sparse/linalg/_dsolve/linsolve.py
```diff
@@ -26,6 +26,7 @@
 
 
 class MatrixRankWarning(UserWarning):
+    """Warning on exactly singular matrices."""
```
j-bowhay, line 29: Can you change both instances to the following 
```suggestion
    """Warning for exactly singular matrices."""
```

j-bowhay (approved):

PR #22511, opened 2025-02-11
BUG: sparse.csgraph: Added support for casting coo array to csc/csr array in ``min_weight_full_bipartite_matching``

### LINKED ISSUE #22501: BUG: `min_weight_full_bipartite_matching` fails for `coo_matrix` input with `AttributeError: 'coo_matrix' object has no attribute 'indices'`
### Describe your issue.

I noticed calling `scipy.sparse.csgraph.min_weight_full_bipartite_matching` with its argument of `scipy.sparse.coo_matrix` fails while it seems to be allowed in the [documentation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csgraph.min_weight_full_bipartite_matching.html). This happens only for `scipy>=1.15.0`.

### Reproducing Code Example

```python
import numpy as np
from scipy.sparse import coo_matrix
from scipy.sparse.csgraph import min_weight_full_bipartite_matching
test_mat = np.array([[1,2,3],[4,5,6],[7,8,9]])

min_weight_full_bipartite_matching(coo_matrix(test_mat))
```

### Error message

```shell
  File "/Users/fukai/scipy-bug.py", line 6, in <module>
    min_weight_full_bipartite_matching(coo_matrix(test_mat))
  File "_matching.pyx", line 486, in scipy.sparse.csgraph._matching.min_weight_full_bipartite_matching
AttributeError: 'coo_matrix' object has no attribute 'indices'
```

```shell
{
  "Compilers": {
    "c": {
      "name": "clang",
      "linker": "ld64",
      "version": "15.0.0",
      "commands": "cc"
    },
    "cython": {
      "name": "cython",
      "linker": "cython",
      "version": "3.0.11",
      "commands": "cython"
    },
    "c++": {
      "name": "clang",
      "linker": "ld64",
      "version": "15.0.0",
      "commands": "c++"
    },
    "fortran": {
      "name": "gcc",
      "linker": "ld64",
      "version": "13.3.0",
      "commands": "gfortran"
    },
    "pythran": {
      "version": "0.16.1",
      "include directory": "../../../../../../private/var/folders/zq/1mkv5ktx0_5cz1rlynrmdmlh0000gn/T/pip-build-env-pw6uu9w1/overlay/lib/python3.12/site-packages/pythran"
    }
  },
  "Machine Information": {
    "host": {
      "cpu": "x86_64",
      "family": "x86_64",
      "endian": "little",
      "system": "darwin"
    },
    "build": {
      "cpu": "x86_64",
      "family": "x86_64",
      "endian": "little",
      "system": "darwin"
    },
    "cross-compiled": false
  },
  "Build Dependencies": {
    "blas": {
      "name": "accelerate",
      "found": true,
      "version": "unknown",
      "detection method": "extraframeworks",
      "include directory": "unknown",
      "lib directory": "unknown",
      "pc file directory": "unknown"
    },
    "lapack": {
      "name": "accelerate",
      "found": true,
      "version": "unknown",
      "detection method": "extraframeworks",
      "include directory": "unknown",
      "lib directory": "unknown",
      "pc file directory": "unknown"
    },
    "pybind11": {
      "name": "pybind11",
      "version": "2.12.0",
    }
  },
  "Python Information": {
    "path": "/private/var/folders/zq/1mkv5ktx0_5cz1rlynrmdmlh0000gn/T/cibw-run-mpfy6fi_/cp312-macosx_x86_64/build/venv/bin/python",
    "version": "3.12"
  }
}
```

czgdp1807:
#### Reference issue

Closes https://github.com/scipy/scipy/issues/22501

The fix involves casting ``row`` and ``col`` for ``coo_array``. The original code in `main` branch assumes that the input to ``min_weight_full_bipartite_matching`` is in ``csr`` format. However, ``coo`` format is also allowed. I am handling ``coo`` format in my fix.

Added tests for ``coo`` format. They are not there in ``main``.

### DIFF
--- a/scipy/sparse/__init__.py
+++ b/scipy/sparse/__init__.py
@@ -32,7 +32,7 @@ SciPy 2-D sparse array package for numeric data.
    - Sparse arrays use array style *slicing* operations, returning scalars,
      1D, or 2D sparse arrays. If you need 2D results, use an appropriate index.
      E.g. ``A[:, i, None]`` or ``A[:, [i]]``.
-   - All index arrays for a given sparse array should be of same type.
+   - All index arrays for a given sparse array should be of same dtype.
      For example, for CSR format, ``indices`` and ``indptr`` should have
      the same dtype. For COO, each array in `coords` should have same dtype.
 
--- a/scipy/sparse/__init__.py
+++ b/scipy/sparse/__init__.py
@@ -32,6 +32,9 @@ SciPy 2-D sparse array package for numeric data.
    - Sparse arrays use array style *slicing* operations, returning scalars,
      1D, or 2D sparse arrays. If you need 2D results, use an appropriate index.
      E.g. ``A[:, i, None]`` or ``A[:, [i]]``.
+   - All index arrays for a given sparse array should be of same type.
+     For example, for CSR format, ``indices`` and ``indptr`` should have
+     the same dtype. For COO, each array in `coords` should have same dtype.
 
    The construction utilities (`eye`, `kron`, `random`, `diags`, etc.)
    have appropriate replacements (see :ref:`sparse-construction-functions`).
--- a/scipy/sparse/csgraph/_matching.pyx
+++ b/scipy/sparse/csgraph/_matching.pyx
@@ -6,7 +6,7 @@ cimport numpy as np
 from libc.math cimport INFINITY
 
 
-from scipy.sparse import issparse, csr_array
+from scipy.sparse import issparse, csr_array, coo_array
 from scipy.sparse._sputils import (convert_pydata_sparse_to_scipy,
                                    safely_cast_index_arrays)
 
@@ -482,11 +482,15 @@ def min_weight_full_bipartite_matching(biadjacency, maximize=False):
 
     a = np.arange(np.min(biadjacency.shape))
 
-    indices, indptr = safely_cast_index_arrays(biadjacency, ITYPE, msg="csgraph")
-    if indices is not biadjacency.indices:
-        # create a new object without copying data
-        biadjacency = csr_array((biadjacency.data, indices, indptr),
-                                shape=biadjacency.shape, dtype=biadjacency.dtype)
+    if biadjacency.format != "coo":
+        indices, indptr = safely_cast_index_arrays(biadjacency, ITYPE, msg="csgraph")
+        if indices is not biadjacency.indices:
+            # create a new object without copying data
+            biadjacency = csr_array((biadjacency.data, indices, indptr),
+                                    shape=biadjacency.shape, dtype=biadjacency.dtype)
+    else:
+        row, col = safely_cast_index_arrays(biadjacency, ITYPE, msg="csgraph")
+        biadjacency = coo_array((biadjacency.data, (row, col)), shape=biadjacency.shape)
 
     # The algorithm expects more columns than rows in the graph, so
     # we use the transpose if that is not already the case. We also
--- a/scipy/sparse/csgraph/_matching.pyx
+++ b/scipy/sparse/csgraph/_matching.pyx
@@ -6,7 +6,7 @@ cimport numpy as np
 from libc.math cimport INFINITY
 
 
-from scipy.sparse import issparse, csr_array, coo_array
+from scipy.sparse import issparse, csr_array
 from scipy.sparse._sputils import (convert_pydata_sparse_to_scipy,
                                    safely_cast_index_arrays)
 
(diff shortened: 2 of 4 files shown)


dschult (dismissed):
This looks good to me!
Thanks @czgdp1807 

~~I think the failures are unrelated -- but there are many of them. So maybe merge with main once the actions are updated there.~~

czgdp1807:
@perimosocordiae Could you please take a look at this PR as well? Thank you.

scipy/sparse/csgraph/_matching.pyx
```diff
@@ -482,11 +482,15 @@ def min_weight_full_bipartite_matching(biadjacency, maximize=False):
...
-                                shape=biadjacency.shape, dtype=biadjacency.dtype)
+    if biadjacency.format != "coo":
+        indices, indptr = safely_cast_index_arrays(biadjacency, ITYPE, msg="csgraph")
+        if indices is not biadjacency.indices:
+            # create a new object without copying data
+            biadjacency = csr_array((biadjacency.data, indices, indptr),
+                                    shape=biadjacency.shape, dtype=biadjacency.dtype)
+    else:
+        row, col = safely_cast_index_arrays(biadjacency, ITYPE, msg="csgraph")
+        biadjacency = coo_array((biadjacency.data, (row, col)), shape=biadjacency.shape)
```
czgdp1807, line 493: I think we can add a condition like, `if row is not biadjacency.row or col is not biadjacency.col:`

What do you say @dschult @perimosocordiae ?

dschult: Adding a check of whether the coords arrays have changed can save a constructor call to `coo_array`. So +1 for making a check before reconstructing.

Your suggestion includes checking both coord arrays instead of checking only one (we only check `indices` in the code for CSR rather than `indices` and `indptr`).  This is a subtle distinction that we should talk about and maybe update. The current code assumes index arrays have the same dtype. That is done in many places in sparse -- we don't support e.g. indptr of type int32 while indptr is int64. Some code-paths might work with such an arrangement, but most won't and any sparsetools calls would definitely not work.  The question this suggestions brings up is: should enforce/check both index dtypes? That is, can we assume the index dtypes are the same when we call `safely_cast_index_arrays`? Or should we check that they are the same? Either way we handle it we should probably put something in the docs about requiring uniform index dtypes with a single sparse array.

czgdp1807: My check is a bit over-safe. `safely_cast_index_arrays` indeed casts all the index arrays to same `dtype`. I can do one check only as well, `if row is not biadjacency.row`. Doing for both is also fine by me. Its harmless.

> Either way we handle it we should probably put something in the docs about requiring uniform index dtypes with a single sparse array.

Yes. We can put a generic statement in the `Note` section of https://docs.scipy.org/doc/scipy/reference/sparse.html like, "All index arrays for sparse arrays should be of same type. For example, for `csr_array`, `index` and `indptr` should have same type. For `coo_array`, `row` and `col` should have same type."

What do you say?

dschult:
I have revisited my review above from the larger context of this function and it's suuport for COO format.
Very soon after this PR's changes, the sparse input (or its transpose) gets converted to CSR.  That means the COO input will get safely cast here and then converted almost immediately afterwards. We should be able to avoid doing both conversions where posible.

So I am now thinking that instead of converting the dtype of COO, we should convert the input (if COO) to either CSR or CSC  before checking/casting the index dtype.  
```python
        if biadjacency.format not in ("csc", "csr"):
            biadjacency = biadjacency.to_csc() if j < i else biadjacency.to_csr()
```
Then the previous call to `safely_cast....` would remain. This seems easier to read, and hopefully avoids multiple conversions. What do you think?  Am I missing something?

And yes, I think `doc/scipy/reference/sparse.rst` is the right place to put a sentence like: (I changed wording slightly, and index -> indices and row/col -> coords.) 
```
All index arrays for a given sparse array should be of same type.
For example, for CSR format, ``indices`` and ``indptr`` should have
the same dtype. For COO, each array in `coords` should have same dtype.
```

czgdp1807:
I agree with you 100%. The reason why I didn't do it is to keep the programming paradigm intact. `tocsc` and `tocsr` both have `copy=False` by default. So performance wise it should be good. 

Regarding the documentation change, I think since its in the scope of this PR, so I will add a commit for it here only.

czgdp1807:
CI failures seem irrelevant to this PR.

scipy/sparse/__init__.py
```diff
@@ -32,6 +32,9 @@
    - Sparse arrays use array style *slicing* operations, returning scalars,
      1D, or 2D sparse arrays. If you need 2D results, use an appropriate index.
      E.g. ``A[:, i, None]`` or ``A[:, [i]]``.
+   - All index arrays for a given sparse array should be of same type.
```
dschult, line 35:
```suggestion
   - All index arrays for a given sparse array should be of same dtype.
```

dschult (approved):
This looks good to me -- with a typo suggestion below (`type` to `dtype`)
I approve this PR

czgdp1807:
CI failures seem irrelevant. Circle CI say, it received exit code 2. The Array API failure is in a different test (not relevant to my changes).

@dschult Fixed the typo. Everything is ready here.

@j-bowhay @perimosocordiae

j-bowhay (approved):

Issue #19473, opened 2023-11-04
ENH: speicial: add ufunc for logarithm of the binomial coefficient


dschmitz89:
### Is your feature request related to a problem? Please describe.

In many discrete distributions the logarithm of the binomial coefficients is used for the logpmf calculation. As the binomial coefficient is fundamentally important elsewhere as well, I think it is worth adding a dedicated function for that in `special`.

### Describe the solution you'd like.

A function `special.logbinom(x, y)`.

aadya940:
Hey, If we're implementing this I have this small implementation:
`
cimport cython
from cython.parallel import prange

cdef extern from "math.h":
    double log(double x)

'''
Use Stirling's approximation if approximate = True
i.e log(n!) = n*log(n) - n
'''

cdef double _factorial(double x) nogil:
    cdef double i = 1
    cdef double _fact = 1
    for i in prange(1, x + 1):
        _fact *= i
    return _fact

cdef double _stirlings_approximation(double x):
    return x * log(x) - x

@cython.ufunc
cdef double _logbinomial(double x, double y, bint approximate=True):
    cdef double result
    if approximate:
        result = _stirlings_approximation(x) - _stirlings_approximation(y) - _stirlings_approximation(x - y)
    else:
        result = _factorial(x) / (_factorial(y) * _factorial(x - y))
    return result
`

aadya940:
Hey, If we're implementing this, I have a small implementation here: 
```
cimport cython
from cython.parallel import prange

cdef extern from "math.h":
    double log(double x)

'''
Use Stirling's approximation if approximate = True
i.e log(n!) = n*log(n) - n
'''

cdef double _factorial(double x) nogil:
    cdef double i = 1
    cdef double _fact = 1
    for i in prange(1, x + 1):
        _fact *= i
    return _fact

cdef double _stirlings_approximation(double x):
    return x * log(x) - x

@cython.ufunc
cdef double _logbinomial(double x, double y, bint approximate=True):
    cdef double result
    if approximate:
        result = _stirlings_approximation(x) - _stirlings_approximation(y) - _stirlings_approximation(x - y)
    else:
        result = _factorial(x) / (_factorial(y) * _factorial(x - y))
    return result
```

steppi:
Using stirlings isn't a great approach. I think for the general case you'd want to rely on

$${n \choose k} = \frac{1}{(n+1)\mathrm{B}(n - k + 1, k + 1)}$$

and use `betaln`.

[...]

PR #22512, opened 2025-02-11
MAINT: special: Add private ufunc for log of absolute value of binomial coefficient

steppi:
#### Reference issue

This comes most of the way to closing #19473. The only remaining thing is that the ufunc added here is private, but I interpret that issue as asking for a public implementation.

This adds a private ufunc `scipy.special._ufuncs._binomln` for the log of the absolute value of the binomial coefficient. It relies primarily on the formula

$$\log\left(\{n \choose k}\right)\ = -\log\left(n + 1\right) - \log\left(\mathrm{B}(n - k + 1, k + 1)\right)$$

with some special handling for the small $k$ case to deal with the difficulties @mdhaber mentioned in https://github.com/scipy/scipy/issues/19473#issuecomment-2336839165.

#### Additional information

For $k$ small it in turn relies on

$$-\log\left(\mathrm{B}(n - k + 1, k + 1)\right) = \log\left(\Gamma(n + 2)\right) - \log\left(\Gamma(k+1)\right) - \log\left(\Gamma(n - k + 1)\right)$$
 
$\log\left(\Gamma(n+2)\right)$ can be rewritten as $\log(n + 1) + \log\left(\Gamma(n + 1)\right)$ using the classic functional relation for the Gamma function https://dlmf.nist.gov/5.5#E1. The $\log(n+1)$ terms in the expression for `binomln` then cancel, leaving us with

$$\log\left(\{n \choose k}\right) = - \log\left(\Gamma(k+1)\right) + \log\left(\Gamma(n + 1)\right) - \log\left(\Gamma(n - k + 1)\right)$$

$\log\left(\Gamma(k+1)\right)$ can be computed accurately with `scipy.special._ufuncs._lgam1p`. 

$\log\left(\Gamma(n + 1)\right) - \log\left(\Gamma(n - k + 1)\right)$ can be computed accurately using the taylor series

$$\log\left(\Gamma(n + 1 + \epsilon)\right) = \log\left(\Gamma(n+1)\right) + \sum_{j=1}^{\infty}\frac{\psi^{j}(n + 1)\epsilon^j}{j!}$$

 where $\psi^{j}(x)$ is the [Polygamma function](https://dlmf.nist.gov/5.15).

To simplify things, I only took the expansion to first order, yielding the formula

$$\log\left(\{n \choose k}\right) \approx -\log\left(\Gamma(k + 1)\right)  + k\psi(n + 1)$$, where $psi$ is the Digamma function.

When $n$ is small, taking the expansion to only first order fails because the Polygamma terms grow large in this limit. We can just use `lgam1p` again in this case though:

```C++
    /* n and k are both small. */
    return cephes::lgam1p(n) - cephes::lgam1p(k) - cephes::lgam1p(n - k);
```

I didn't put too much effort into identifying where the cutoff should be for $k$ and $n$ to be considered small. I picked a threshold of `1e-7` for each based on some limited trial and error. Before we make this public we might want to choose these in a more principled way.

In https://github.com/scipy/scipy/issues/19473#issuecomment-2336839165, @mdhaber had also pointed out that there are cases where this implementation suffers in accuracy purely due to limitations in `scipy.special.betaln`. I haven't done anything to address that here, but we should look into improving `betaln` at some point. As @mdhaber had surmised, these difficult cases are just at the cusp of where `betaln` switches over to an asymptotic expansion.


mdhaber:
Just to simplify the derivation: numbering from the top, equation 3 is just the log of the usual definition of $n$ choose $k$ but using the gamma function instead of factorial.

scipy/special/xsf/binom.h
```diff
@@ -86,4 +88,40 @@ XSF_HOST_DEVICE inline float binom(float n, float k) {
...
+	double result = n >= -1 ? -std::log1p(n) : -std::log(-n - 1);
+        result -= cephes::lbeta(1 + n - k, 1 + k);
+	return result;
+    }
+    if (std::abs(n) > 1e-7) {
+	/* Take log_binom(n, k) = -log(n + 1) - betaln(n - k + 1, k + 1)
+	 * and apply betaln(n - k + 1, k + 1) =
+	 * gammaln(n - k + 1) + gammaln(k + 1) - gammaln(n + 2).
+	 * Using gammaln(n + 2) = log(n + 1) + gammaln(n + 1) the log(n + 1)
+	 * terms cancel, yielding
```
mdhaber, line 114:
```suggestion
```
I think the next line is obvious enough (e.g. from the [docstring of `binom`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.binom.html)) that these are not needed.

scipy/special/xsf/binom.h
```diff
@@ -86,4 +88,40 @@ XSF_HOST_DEVICE inline float binom(float n, float k) {
...
+	 * log_binom(n, k) = gammaln(n + 1) - gammaln(n + 1 - k) - gammaln(1 + k)
+	 * gammaln(1 + k) can be computed accurately for small k using lgam1p.
+	 * We can then use the taylor series expansion
+	 * gammaln(n + 1 - k) = gammaln(n + 1) - k*digamma(n + 1) +
+	 * k^2 * polygamma(3, n + 1) + ... + (-k)^j * polygamma(j, n + 1) + ...
```
mdhaber, line 119:
```suggestion
	 * We can then use the Taylor series expansion
	 * gammaln(n + 1 - k) = gammaln(n + 1) - k*digamma(n + 1) +
	 * k^2 * polygamma(3, n + 1) + ... + (-k)^j * polygamma(j, n + 1) + ...
```
From the `(-k)^j * polygamma(j, n + 1)` term, I would have expected the integers in `k^2 * polygamma(3, n + 1)` to match?

scipy/special/xsf/binom.h
```diff
@@ -86,4 +88,40 @@ XSF_HOST_DEVICE inline float binom(float n, float k) {
...
+    bool denominator_pole = (k <= - 1 && k == std::floor(k)) || (a <= 0 && a == std::floor(a));
+    if (n <= -1 && n == std::floor(n)) {
+	return denominator_pole ? std::numeric_limits<double>::quiet_NaN() : std::numeric_limits<double>::infinity();
+    }
+    if (denominator_pole) {
+	return -std::numeric_limits<double>::infinity();
+    }
+    if (n == k) {
+	return 0;
+    }
```
mdhaber, line 103: Is there similar logic in `binom`? That is, can we handle these cases with with `log(binom(n, k))`?

mdhaber (commented):
Notes to self.

scipy/special/xsf/binom.h
```diff
@@ -86,4 +88,40 @@ XSF_HOST_DEVICE inline float binom(float n, float k) {
...
+    if (std::abs(k) > 1e-7) {
+	double result = n >= -1 ? -std::log1p(n) : -std::log(-n - 1);
+        result -= cephes::lbeta(1 + n - k, 1 + k);
```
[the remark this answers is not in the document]
steppi: Right. It's written this way because I wanted to still be able to use `log1p` to maintain precision for smaller `n`. It may be a bit of a premature optimization, but I didn't use `abs` to avoid an extra branch.

scipy/special/xsf/binom.h
```diff
@@ -86,4 +88,40 @@ XSF_HOST_DEVICE inline float binom(float n, float k) {
...
+	 * log_binom(n, k) = gammaln(n + 1) - gammaln(n + 1 - k) - gammaln(1 + k)
+	 * gammaln(1 + k) can be computed accurately for small k using lgam1p.
+	 * We can then use the taylor series expansion
+	 * gammaln(n + 1 - k) = gammaln(n + 1) - k*digamma(n + 1) +
+	 * k^2 * polygamma(3, n + 1) + ... + (-k)^j * polygamma(j, n + 1) + ...
+	 */
+	return -cephes::lgam1p(k) + k*digamma(n + 1);
+    }
+    /* n and k are both small. */
+    return cephes::lgam1p(n) - cephes::lgam1p(k) - cephes::lgam1p(n - k);
```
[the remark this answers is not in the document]
steppi: Ha, right. That's what happens when you sit down with a notebook and just single mindedly try to work something out as fast as possible without reflecting.

mdhaber:
Just want to confirm that all sufficiently negative values of `n` are effectively integers, so we expect NaNs for all sufficiently negative `n`?

scipy/special/xsf/binom.h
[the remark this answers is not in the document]
mdhaber: The code makes sense to me. I was just confused by `(-n -1)` initially.

scipy/special/xsf/binom.h
mdhaber: K, the thing is that I need to review it. What should I use as a reference?
Or would the expression in terms of `betaln` do the right thing?

steppi:
> > 
> 
> No, I didn't expect this. I have a fix though. Just pushed.

I made it fall back to `gammaln(n + 1) - gammaln(k + 1) - gammaln(n - k + 1)` if the first formula in terms of `betaln` produces `NaN`. This still isn't accurate in the very large `n`, very large `k`, `n >> k` regime, though it won't return `NaN`. I just made it so it signals `SF_ERROR_LOSS` if this fallback occurs and returns a finite result.

This fallback also leads to the correct behavior at poles without special handling, so I dropped the special cases for poles.

mdhaber:
I think we have a pretty good grasp of when this disagrees with `mpmath` and why.

- When `n < -1e15`, `n` is a negative integer, and we return NaN. 
- When `n > 1e50` and `abs(n) >> abs(k)`, loss of precision in `n - k` causes accuracy issues.
- When `k > 1e15`, `k` is integral, so `n - k` is a negative integer, and we return `-inf`.

We can also run into cases where `n - k` is integral by chance when the magnitude of one the two is close to the limit.

Filtering out those cases, we agree with `mpmath` about which results should be negative infinity vs finite, and there are no NaNs or positive infinities. Then relative tolerances are typically better than `1e-12` *unless* the result is nearly `0`. I think the trouble in that case falls into a few categories:

- When `n` and `k` are both tiny, we just don't have a good implementation.
- When `n` and `k` are both large but very close to one another, `n - k` loses precision.
- I think there is another case, and I'm finding it late: `n` is negative with large magnitude, `k` is positive with small magnitude, presumable `n - k` loses precision. 

Here is some code I've been playing around with.

<details>

```
import numpy as np
from mpmath import mp
from scipy.special._ufuncs import _binomln

mp.dps = 400
rng = np.random.default_rng(43598768734563456)

@np.vectorize
def binomln(n, k):
    return float(mp.log(abs(mp.binomial(n, k))))

def get_data(rng):
    x = np.concat((np.logspace(-300, -10, 30),
                   np.logspace(-10, 10, 50),
                   np.logspace(10, 300, 30)))
    x = np.concat((-x[::-1], x))
    factor = 1 + 0.1*rng.normal(size=len(x))
    return factor * x

n = get_data(rng)
k = get_data(rng)
n = n[n > -1e10]  # because too large values are integers and we return NaN
n = n[n < 1e50]  # when abs(n) is large and much larger than abs(k)
k = k[k < 1e10]  # e.g. (-9874128620.24379, 1.1425364482510619e+20): k is integral and therefore n-k+1 is integral and negative

n = n[:, np.newaxis]
n, k = np.broadcast_arrays(n, k)

res = _binomln(n, k)
ref = binomln(n, k)

np.testing.assert_equal(np.isfinite(res), np.isfinite(ref))
np.testing.assert_equal(np.isneginf(res), np.isneginf(ref))
assert not np.any(np.isnan(res))
assert not np.any(np.isposinf(res))

np.testing.assert_allclose(res, ref, atol=1e-3, rtol=1e-12)  # only one failure here

mask = np.isfinite(ref) & (np.abs(ref) > 0)
n = n[mask]
k = k[mask]
res = res[mask]
ref = ref[mask]
abs_err = np.abs(res - ref)
rel_err = np.abs(abs_err / ref)

mask = (abs_err > 1e-14) & (rel_err > 1e-9) & (np.abs(res) > 0)
n = n[mask]
k = k[mask]
res = res[mask]
ref = ref[mask]
abs_err = np.abs(res - ref)
rel_err = np.abs(abs_err / ref)
print(np.max(np.abs((n - k)/k)))

```

</details>

scipy/special/tests/test_binomln.py
```diff
@@ -0,0 +1,152 @@
...
+    #         cases * np.asarray([[-1, -1]]),
+    #     ]
+    # )
+
+    # test_cases = []
+    # for n, k in cases:
+    #     desired = mp_binomln(n, k)
+    #     actual = _binomln(n, k)
+    #     relative_error = abs(observed_result - expected_result) / abs(expected_result)
+    #     rtol = f"1e{int(max(-15, np.ceil(np.log10(relative_error))))}"
```
mdhaber, line 60: I don't think these lines match the actual code used to generate this. One difference is the name of `desired`/`actual`, the other is that you wouldn't get `rtol=5e-15` with this code.

steppi: Oh, good point. I changed the actual script but not the comment. Also the script in the comment is a nonworking version while I was in the middle of updating things. The choice of test cases is probably not that good. I threw this together quickly, the intent was just to get something that looked reasonable without thinking too much about it. And `rtol=5e-15` came from a manual bump to get that test to pass on all platforms in CI. We can definitely do better than this and use fewer magic numbers.

fbourgey:
@steppi can you please open a PR over at XSF for this first?

steppi:
> @steppi can you please open a PR over at XSF for this first?

I'd forgotten about this PR. It actually predates the split of `xsf` into a separate repo, still needs a lot of work, and can be safely closed and moved there when I get a chance.

PR #22441, opened 2025-01-30
BUG: special: Fix incorrect handling of `nan` input in `gammainc` and `gammaincc`

steppi:
#### Reference issue

Closes #21317

#### What does this implement/fix?

This PR adds `nan` checks to the scalar kernels for `gammainc` and `gammaincc`. As observed by @mdhaber in #21317, there were some cases where non-`nan` output was being returned despite `nan` values of `a` or `x`. These were not cases like `pow(1, nan)` where a constant partial function remains constant for `nan` input. The output should be `nan`.

#### Additional information

This will also fix the same problem for `chdtr` and `chdtrc` also pointed out in #21317, since these wrap the same `cephes` function which has been fixed here. I haven't added test cases confirming the fix for `chdtr` and `chdtrc` but can if desired.

PR #22514, opened 2025-02-11
TST: special: Add edgecase tests for gammainc and friends

### LINKED ISSUE #20972: BUG: special.chdtrc: returns 1.0 when both degrees of freedom and argument are negative
### Describe your issue.

As expected, the CDF of the chi-squared distribution is undefined for negative degrees of freedom.
```python3
from scipy import special
special.chdtr(-1, [-1, 1])  # array([nan, nan])
```

Yet:
```python3
special.chdtrc(-1, [-1, 1])  # array([1., nan])
```

### Reproducing Code Example

```python
from scipy import special
special.chdtrc(-1, -1)  # produces 1.0 instead of NaN
```

### Error message

```shell
NA
```

```shell
1.15.0.dev0+935.e984c95 1.26.4 sys.version_info(major=3, minor=12, micro=3, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    detection method: pkgconfig
    found: true
    include directory: C:/Users/matth/miniforge3/envs/scipy-dev/Library/include/openblas
    lib directory: C:/Users/matth/miniforge3/envs/scipy-dev/Library/lib
    name: openblas
    openblas configuration: USE_64BITINT= NO_CBLAS= NO_LAPACK= NO_LAPACKE= DYNAMIC_ARCH=ON
      DYNAMIC_OLDER=OFF NO_AFFINITY=1 USE_OPENMP= HASWELL MAX_THREADS=128
    pc file directory: C:\Users\matth\miniforge3\envs\scipy-dev\Library\lib\pkgconfig
    version: 0.3.27
  lapack:
    detection method: pkgconfig
    found: true
    include directory: C:/Users/matth/miniforge3/envs/scipy-dev/Library/include/openblas
    lib directory: C:/Users/matth/miniforge3/envs/scipy-dev/Library/lib
    name: openblas
    openblas configuration: USE_64BITINT= NO_CBLAS= NO_LAPACK= NO_LAPACKE= DYNAMIC_ARCH=ON
      DYNAMIC_OLDER=OFF NO_AFFINITY=1 USE_OPENMP= HASWELL MAX_THREADS=128
    pc file directory: C:\Users\matth\miniforge3\envs\scipy-dev\Library\lib\pkgconfig
    version: 0.3.27
  pybind11:
    detection method: pkgconfig
    include directory: C:/Users/matth/miniforge3/envs/scipy-dev/Library/include
    name: pybind11
    version: 2.12.0
Compilers:
  c:
    commands: cc
    linker: ld.bfd
    name: gcc
    version: 10.3.0
  c++:
    commands: c++
    linker: ld.bfd
    name: gcc
    version: 10.3.0
  cython:
    commands: cython
    linker: cython
    name: cython
    version: 3.0.10
  fortran:
    commands: gfortran
    linker: ld.bfd
    name: gcc
    version: 10.3.0
  pythran:
    include directory: ..\..\miniforge3\envs\scipy-dev\Lib\site-packages\pythran
    version: 0.15.0
Machine Information:
  build:
    cpu: x86_64
    endian: little
    family: x86_64
    system: windows
  cross-compiled: false
  host:
    cpu: x86_64
    endian: little
    family: x86_64
    system: windows
Python Information:
  path: C:\Users\matth\miniforge3\envs\scipy-dev\python.exe
  version: '3.12'
```

steppi:
#### Reference issue

Closes #20972

As confirmed in https://github.com/scipy/scipy/issues/20972#issuecomment-2649494122, #20972 was incidentally fixed in #22441. This PR adds tests confirming this. The tests uses all entries of the table in https://github.com/scipy/scipy/issues/20972#issuecomment-2649494122, checking that NaN is returned when it should be, and something other than NaN is returned when the result should not be NaN.

### DIFF
--- a/scipy/special/tests/test_basic.py
+++ b/scipy/special/tests/test_basic.py
@@ -4336,6 +4336,33 @@ def test_chi2c_x_nan(v):
     assert np.isnan(special.chdtrc(v, np.nan))
 
 
+@pytest.mark.parametrize("v", [-0.01, 0, 0.01, 1, np.inf])
+@pytest.mark.parametrize("x", [-np.inf, -0.01, 0, 0.01, np.inf])
+@pytest.mark.parametrize(
+    "func", [special.chdtr, special.chdtrc, special.gammainc, special.gammaincc,
+             special.gdtr, special.gdtrc]
+)
+def test_chi2_edgecase_table(v, x, func):
+    # Tests that a variety of edgecases for chi square distribution functions
+    # correctly return NaN when and only when they are supposed to, when
+    # computed through different related ufuncs.
+    if func in [special.chdtr, special.chdtrc]:
+        result = func(v, x)
+    elif func in [special.gdtr, special.gdtrc]:
+        result = func(1, v/2, x/2)
+    else:
+        result = func(v/2, x/2)
+    if (
+            x < 0
+            or v < 0
+            or x == 0 and v == 0
+            or np.isinf(v) and np.isinf(x)
+    ):
+        assert_equal(result, np.nan)
+    else:
+        assert not np.isnan(result)
+
+
 def test_chi2c_smalldf():
     assert_almost_equal(special.chdtrc(0.6,3), 1-0.957890536704110)
 
--- a/scipy/special/tests/test_basic.py
+++ b/scipy/special/tests/test_basic.py
@@ -4336,31 +4336,25 @@ def test_chi2c_x_nan(v):
     assert np.isnan(special.chdtrc(v, np.nan))
 
 
-@pytest.mark.parametrize("v", [-0.01, 0, 0.01, 1, np.inf])
-@pytest.mark.parametrize("x", [-np.inf, -0.01, 0, 0.01, np.inf])
-@pytest.mark.parametrize(
-    "func", [special.chdtr, special.chdtrc, special.gammainc, special.gammaincc,
-             special.gdtr, special.gdtrc]
-)
-def test_chi2_edgecase_table(v, x, func):
+def test_chi2_edgecases_gh20972():
     # Tests that a variety of edgecases for chi square distribution functions
     # correctly return NaN when and only when they are supposed to, when
-    # computed through different related ufuncs.
-    if func in [special.chdtr, special.chdtrc]:
-        result = func(v, x)
-    elif func in [special.gdtr, special.gdtrc]:
-        result = func(1, v/2, x/2)
-    else:
-        result = func(v/2, x/2)
-    if (
-            x < 0
-            or v < 0
-            or x == 0 and v == 0
-            or np.isinf(v) and np.isinf(x)
-    ):
-        assert_equal(result, np.nan)
-    else:
-        assert not np.isnan(result)
+    # computed through different related ufuncs. See gh-20972.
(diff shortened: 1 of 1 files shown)


mdhaber (approved):
LGTM.

One thought is that the test would run a bit faster by including all the combinations in arrays.
```python3
v = np.asarray([-0.01, 0, 0.01, 1, np.inf])[:, np.newaxis]
x = np.asarray([-np.inf, -0.01, 0, 0.01, np.inf])
mask = (x < 0) | (v < 0) | (x == 0) & (v == 0) | np.isinf(v) & np.isinf(x)
```
I agree that it is easier to see what is wrong when it's parametrized - but that only matters if there's a regression.

I mention that mostly because the test would also be slightly stronger if we used one of these as as a reference and ensured that the numerical results were not only finite but consistent.
```python3
ref = special.gammainc(v/2, x/2)
assert np.all(np.isnan(gammainc[mask]))
assert np.all(np.isfinite(gammainc[~mask]))

assert_allclose(special.chdtr(v, x), ref)
assert_allclose(special.gdtr(1, v/2, x/2), ref)
assert_allclose(1 - special.gammaincc(v/2, x/2), ref)
assert_allclose(1 - special.chdtrc(v, x), ref)
assert_allclose(1 - special.gdtrc(1, v/2, x/2), ref)
```

If you prefer for this to be targeted toward NaN vs non-NaN, it's fine as-is. LMK either way.

steppi:
OK, yeah. The test runs in around 0.5 seconds now, which is a bit on the long side. I'll change it to follow what you just wrote.

mdhaber:
Looks like CI should be fixed, so I went ahead and changed it.

steppi:
> OK, yeah. The test runs in around 0.5 seconds now, which is a bit on the long side. I'll change it to follow what you just wrote.

Ok,

```
 python dev.py test -t scipy.special.tests.test_basic -m full -- -k "test_chi2_edgecase_table"
```

averaged `0.48s` over 5 runs on my machine using the parametrized version from this PR. Changing the test to the following dropped this to an average of `0.42s`.

```python
def test_chi2_edgecase_table():
    # Tests that a variety of edgecases for chi square distribution functions
    # correctly return NaN when and only when they are supposed to, when
    # computed through different related ufuncs.
    v = np.asarray([-0.01, 0, 0.01, 1, np.inf])[:, np.newaxis]
    x = np.asarray([-np.inf, -0.01, 0, 0.01, np.inf])
    mask = (x < 0) | (v < 0) | (x == 0) & (v == 0) | np.isinf(v) & np.isinf(x)
    ref = special.gammainc(v/2, x/2)
    assert np.all(np.isnan(ref[mask]))
    assert np.all(np.isfinite(ref[~mask]))

    assert_allclose(special.chdtr(v, x), ref)
    assert_allclose(special.gdtr(1, v/2, x/2), ref)
    assert_allclose(1 - special.gammaincc(v/2, x/2), ref)
    assert_allclose(1 - special.chdtrc(v, x), ref)
    assert_allclose(1 - special.gdtrc(1, v/2, x/2), ref)
```

The later is testing more things though. 




> Looks like CI should be fixed, so I went ahead and changed it. LMK if this look good, or feel free to merge.

Thanks @mdhaber. Looks good to me. I'll go ahead and merge.

PR #22517, opened 2025-02-11
CI: Use actions/cache 4.2.0

### LINKED ISSUE #22515: CI: Some GitHub workflows failing due to check on `actions/cache` version.
I noticed several CI jobs failing in https://github.com/scipy/scipy/pull/22512, purportedly due to the use of a deprecated version of `actions/cache`. Here's a picture of the error message.

Image

and here's the relevant GitHub link to gain more information: https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down.

The version of `actions/cache` being used isn't actually deprecated though

https://github.com/scipy/scipy/blob/324a8c0477d016cdb4abd3b3a0e4dcd75b252a6c/.github/workflows/linux_intel_oneAPI.yml#L70

it's just pinned to a particular commit instead of to a version number. It appears that GitHub's logic for determining if the version is deprecated did not account for the possibility of pinning to a particular commit.

Update: Actually, checking here, https://github.com/actions/cache/discussions/1510, it appears that the particular version of `actions/cache` being used *is* deprecated. This link says this:

*If you are using pinned SHAs, please use the SHAs of versions v4.2.0 or v3.4.0*

I guess we should just update to use the SHA for `v4.2.0`. I don't know enough to know why we are using SHA's instead of version numbers though.

nbelakovski:
4.1.2 is deprecated, see https://github.com/scipy/scipy/issues/22515

#### Reference issue
Closes gh-22515

See above

#### Additional information
N/A

### DIFF
--- a/.github/workflows/array_api.yml
+++ b/.github/workflows/array_api.yml
@@ -88,7 +88,7 @@ jobs:
         echo "timestamp=${NOW}" >> $GITHUB_OUTPUT
 
     - name: Setup compiler cache
-      uses:  actions/cache@6849a6489940f00c2f30c0fb92c6274307ccb58a # v4.1.2
+      uses:  actions/cache@1bd1e32a3bdc45362d1e726936510720a7c30a57 # v4.2.0
       id:    cache-ccache
       with:
         path: ${{ steps.prep-ccache.outputs.dir }}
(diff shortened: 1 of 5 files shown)


andyfaff:
I checked the hashes, they're all correct.

tylerjereddy:
marked for backport and pulled it in for `1.15.3` given CI failures

PR #22519, opened 2025-02-12
ENH: add workers to least_squares

andyfaff:
Adds a `workers` keyword to `least_squares`, to enable parallelisation of numerical differentiation during jacobian estimation.

This keyword will only have an effect for `'dogbox'` or `'trf'`, as `'lm'` carries out numerical differentiation inside `minpack->lmdif`.

It might be possible to apply to `'lm'`, but this would require that the numerical differentiation be handled by Python. I think that means `minpack-->lmder` is used instead. This might lead to a slowdown in the serial case (flow of numerical differentiation is handled by Python instead of C), but would hopefully prove faster in the parallel case. 

In addition I am not sure if that change would be substantially identical in behaviour to what we currently have. 
i.e. is `lmder` with a numerical estimation of the jacobian (2-point forward differences) equivalent to `lmdif` with 2-point forward differences?  If `lmder` assumes that it's Jacobian is analytic it might work differently.

### DIFF
--- a/scipy/optimize/_lsq/least_squares.py
+++ b/scipy/optimize/_lsq/least_squares.py
@@ -9,6 +9,7 @@ from scipy.optimize import _minpack, OptimizeResult
 from scipy.optimize._numdiff import approx_derivative, group_columns
 from scipy.optimize._minimize import Bounds
 from scipy._lib._sparse import issparse
+from scipy._lib._util import _workers_wrapper
 
 from .trf import trf
 from .dogbox import dogbox
@@ -241,12 +242,13 @@ def construct_loss_function(m, loss, f_scale):
     return loss_function
 
 
+@_workers_wrapper
 def least_squares(
         fun, x0, jac='2-point', bounds=(-np.inf, np.inf), method='trf',
         ftol=1e-8, xtol=1e-8, gtol=1e-8, x_scale=1.0, loss='linear',
         f_scale=1.0, diff_step=None, tr_solver=None, tr_options=None,
         jac_sparsity=None, max_nfev=None, verbose=0, args=(), kwargs=None,
-        callback=None
+        callback=None, workers=None
 ):
     """Solve a nonlinear least-squares problem with bounds on the variables.
 
@@ -467,6 +469,12 @@ def least_squares(
 
         If the `callback` function raises `StopIteration` the optimization algorithm
         will stop and return with status code -2.
+
+        .. versionadded:: 1.16.0
+    workers : map-like callable, optional
+        A map-like callable, such as `multiprocessing.Pool.map` for evaluating
+        any numerical differentiation in parallel.
+        This evaluation is carried out as ``workers(fun, iterable)``.
 
         .. versionadded:: 1.16.0
 
@@ -929,7 +937,8 @@ def least_squares(
             def jac_wrapped(x, f):
                 J = approx_derivative(fun, x, rel_step=diff_step, method=jac,
                                       f0=f, bounds=bounds, args=args,
-                                      kwargs=kwargs, sparsity=jac_sparsity)
+                                      kwargs=kwargs, sparsity=jac_sparsity,
+                                      workers=workers)
                 if J.ndim != 2:  # J is guaranteed not sparse.
                     J = np.atleast_2d(J)
 
--- a/scipy/_lib/_util.py
+++ b/scipy/_lib/_util.py
@@ -750,6 +750,29 @@ class MapWrapper:
                             " form f(func, iterable)") from e
 
 
+def _workers_wrapper(func):
+    """
+    Wrapper to deal with setup-cleanup of workers outside a user function via a
+    ContextManager. It saves having to do the setup/tear down with within that
+    function, which can be messy.
+    """
+    @functools.wraps(func)
+    def inner(*args, **kwds):
+        kwargs = kwds.copy()
+        if 'workers' not in kwargs:
+            _workers = map
+        elif 'workers' in kwargs and kwargs['workers'] is None:
+            _workers = map
+        else:
+            _workers = kwargs['workers']
+
+        with MapWrapper(_workers) as mf:
+            kwargs['workers'] = mf
+            return func(*args, **kwargs)
+
+    return inner
+
+
 def rng_integers(gen, low, high=None, size=None, dtype='int64',
                  endpoint=False):
     """
--- a/scipy/_lib/tests/test__util.py
+++ b/scipy/_lib/tests/test__util.py
@@ -2,6 +2,7 @@ from multiprocessing import Pool
 from multiprocessing.pool import Pool as PWL
 import re
 import math
+import functools
 from fractions import Fraction
 
 import numpy as np
@@ -17,7 +18,8 @@ from scipy._lib._array_api import (xp_assert_equal, xp_assert_close, is_numpy,
 from scipy._lib._util import (_aligned_zeros, check_random_state, MapWrapper,
                               getfullargspec_no_self, FullArgSpec,
                               rng_integers, _validate_int, _rename_parameter,
-                              _contains_nan, _rng_html_rewrite, _lazywhere)
+                              _contains_nan, _rng_html_rewrite, _lazywhere,
+                              _workers_wrapper)
 import scipy._lib.array_api_extra as xpx
 from scipy._lib.array_api_extra.testing import lazy_xp_function
 from scipy import cluster, interpolate, linalg, optimize, sparse, spatial, stats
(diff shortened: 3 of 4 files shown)


andyfaff:
A quick example showing the potential speedup for a residual function of 100 parameters that takes 2 ms to evaluate. Benefit depends on the number of parameters and the speed of the residual function. The estimation of the Jacobian is O(N) where N is the number of parameters.

```python
import numpy as np
from multiprocessing import Pool
from slow_func import fun_trivial

rng = np.random.default_rng()
x0 = rng.uniform(low=0, high=10, size=(100,))

# slow_func.py
# import time
# def fun_trivial(x, a=0):
#     time.sleep(0.002)
#     return (x - a)**2 + 5.0

%timeit least_squares(fun_trivial, x0, method='trf', workers=map)

with Pool() as pwl:
    %timeit least_squares(fun_trivial, x0, method='trf', workers=pwl.map)
```
gives:
```
2.61 s ± 4.39 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
503 ms ± 2.27 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

scipy/optimize/_lsq/least_squares.py
```diff
@@ -241,12 +242,13 @@ def loss_function(f, cost_only=False):
     return loss_function
 
 
+@_workers_wrapper
```
ilayn, line 245: Just for my own understanding as it's been a while since I used the workers, are we cleaning up the user's workers in case they want to do something else with the pool ? Not sure if that is a problem or not.

andyfaff: If the user supplies their own workers then Mapwrapper doesn't do any cleanup.
It only cleans up any resources (Pools) that it creates.
The purpose of the decorator is so that we don't need to do all that wrangling inside least squares.

andyfaff: The decorator is so that we don't need to setup Mapwrapper within least squares

ilayn: Ah my bad, I've read the decorator backwards. Then apologies for the noise.

andyfaff: No, I appreciate the eyes. The main thing I'd like to know is about lmdif and lmder. At the moment `method='lm'` uses two point forward differences with lmdif. That's not immediately accessible to parallelise because it's done in C. I was wondering if we could change the lm method to use approx_derivative with lmder, and whether the two would be equivalent.

ilayn: This is what the authors wrote half a century ago 

image

ilayn: And this is the explanation of the Jacobian detail 

image

ilayn: As far as I can tell, they are equivalent, contingent upon the correctness of the Jacobian calculator. So I'd be willing to bet on yes you can do that. The authors have no reservations either way.

andyfaff:
[059843b](https://github.com/scipy/scipy/pull/22519/commits/059843bed0e7dd8c085ffb701d697f168d09bc8b) does not result in exactly equivalent behaviour for `method='lm'` with *no* jacobian (i.e. 2-point forward difference), but it's very close.

Consider this test case:

```python
import numpy as np
from numpy.testing import assert_allclose

class ExponentialFittingProblem:
    """Provide data and function for exponential fitting in the form
    y = a + exp(b * x) + noise."""

    def __init__(self, a, b, noise, n_outliers=1, x_range=(-1, 1),
                 n_points=11, random_seed=None):
        rng = np.random.RandomState(random_seed)
        self.m = n_points
        self.n = 2

        self.p0 = np.zeros(2)
        self.x = np.linspace(x_range[0], x_range[1], n_points)

        self.y = a + np.exp(b * self.x)
        self.y += noise * rng.randn(self.m)

        outliers = rng.randint(0, self.m, n_outliers)
        self.y[outliers] += 50 * noise * rng.rand(n_outliers)

        self.p_opt = np.array([a, b])

    def fun(self, p):
        return p[0] + np.exp(p[1] * self.x) - self.y

    def jac(self, p):
        J = np.empty((self.m, self.n))
        J[:, 0] = 1
        J[:, 1] = self.x * np.exp(p[1] * self.x)
        return J

random_seed = 1298098
efp = ExponentialFittingProblem(1, 0.1, noise=1.0, random_seed=random_seed)

res = least_squares(efp.fun, efp.p0, method='lm')
assert_allclose(res.optimality, 0, atol=1e-2)
res
```

with this PR the output is:

```
    message: `ftol` termination condition is satisfied.
     success: True
      status: 2
         fun: [ 1.327e-01  1.370e+00  1.238e+00  1.497e+00  1.203e+00
                1.729e+00  5.201e-01 -1.336e+01  1.178e+00  1.593e+00
                2.894e+00]
           x: [ 2.165e+00  6.459e-01]
        cost: 100.52182749946
         jac: [[ 1.000e+00 -5.242e-01]
               [ 1.000e+00 -4.772e-01]
               ...
               [ 1.000e+00  1.341e+00]
               [ 1.000e+00  1.908e+00]]
        grad: [ 1.593e-07  2.993e-04]
  optimality: 0.0002992568060026235
 active_mask: [0 0]
        nfev: 79
        njev: None
```

with current main the output is:

```
     message: `ftol` termination condition is satisfied.
     success: True
      status: 2
         fun: [ 1.327e-01  1.370e+00  1.238e+00  1.497e+00  1.203e+00
                1.729e+00  5.201e-01 -1.336e+01  1.178e+00  1.593e+00
                2.894e+00]
           x: [ 2.165e+00  6.459e-01]
        cost: 100.52182749927002
         jac: [[ 1.000e+00 -5.242e-01]
               [ 1.000e+00 -4.772e-01]
               ...
               [ 1.000e+00  1.341e+00]
               [ 1.000e+00  1.908e+00]]
        grad: [ 2.930e-07  2.911e-04]
  optimality: 0.0002911031111078044
 active_mask: [0 0]
        nfev: 56
        njev: None
```

Please ignore `nfev` between the two, I believe that main is counting function evaluations incorrectly. You'll see from `res.cost` that main is better than the PR, but only in the 10th decimal place. The optimality is slightly worse, 2.911e-4 (main) vs 2.992e-4 (this PR).

I am unsure at what level these discrepancies become important. If they are important I can drop the commit.

ilayn:
Is there any such problems that the optimal value/cost is known theoretically? These look very close but not sure which one should be the reference. While approximating Jacobian, it can very well be hitting a lucky final step, that the exact Jacobian does not reach after many attempts. Hence I would not just cross out the validity of the approach yet. 

> I believe that main is counting function evaluations incorrectly, and it should be 63

This can very well be true. MINPACK does not really count but adds a prescribed number everytime it enters a function and I might be missing one of those counts. I might have introduced a bug there.

andyfaff:
To be absolutely clear, both those attempts are with `lm`, and they're both runs with finite difference approximation for the jacobian.
The difference is that with this PR `_minpack._lmder` is used, with a jacobian estimated from `approx_derivative`, compared to main using `_minpack._lmdif`, with the jacobian being estimated inside minpack.
I'm fairly sure that `approx_derivative` works well, the test suite examines how closely it reproduces analytic derivatives.

andyfaff:
I modified the code slightly to permit `3-point` estimation of the Jacobian for `_minpack._lmder`. It doesn't really change the course of the minimisation, it still requires the same amount of function evaluations compared to `2-point`. To me this indicates that it's not the accuracy of the jacobian that's the cause of any difference.

ilayn:
Then probably the only remaining part is to `printf` the `lmdif` approximation and the external approximation values. I think if there was a significant bug we would have heard by now. Although that's never a guarantee at least the happy flow should be OK. Hence we can rely on the approximation values to spot any significant differences.

andyfaff:
Can one printf from c, where does the output turn up? I thought I might have to learn to use gdb.

ilayn:
No just use printf whereever you want and it will show up on your console, ipython or notebook. `gdb` is always better but `printf` is unbeatable :) Especially for these one-off things. 

You have to build and install it everytime you make a change though. So working inside your build-install dir is better, wherever it is.

ilayn:
If I am not setting up a full `gdb` this is basically what I do. 

image

The command at the bottom goes back to root folder runs the build and reenters the installed dir and opens ipython. It's extremely vulgar and neanderthal but works great with zero tooling :) I run both the fortran print and C version printf's and watch bugs drive me crazy.

ev-br:
> Can one printf from c, where does the output turn up?

A possible gotcha (am stumbling on this every single time) is to add `-s` to `python dev.py test ... -s` to see the printed output from debug prints (python or C) when running via the test suite.

`<digress>` The main reason I compile things as C++ not C is `std::cout << variable << "\n";` instead of having to use `printf` and its format specifiers.
`</digress>`

ilayn:
> A possible gotcha (am stumbling on this every single time) is to add -s to python dev.py test ... -s to see the printed output from debug prints (python or C) when running via the test suite.

That's true but in this neanderthal way, you don't test anything but actually run different code and see how things move internally. Testing is a different monster to discover how terrible my translations are. Then bug hunting starts

ev-br:
> That's true but in this neanderthal way, you don't test anything but actually run different code and see how things move internally. Testing is a different monster to discover how terrible my translations are. Then bug hunting starts

My debugging often involves running things via tests instead of one-off scripts. There are 11 workflows for each 10 devs I guess :-).

ilayn:
Unfortunately, I'm not a dev so ignorance is bliss :)

andyfaff:
I think it might make sense to remove the commit that modifies the `lm` method, get this PR merged, then work on the `lm` method in another PR.

andyfaff:
Ok, I backed out the changes to the `'lm'` method, so this should be good to go.

andyfaff:
I went back a couple of years to the oldest version I could run on my macOS arm64 computer (1.10/1.11 from 2023). According to the git history this was before minpack was converted to c (July last year). I found that the difference in behaviour -  more user function evaluations in lmder compared to lmdif for the above example - is still observed there. So if there is a problem in the code it was there before the translation occurred.

andyfaff:
I figured out what was going on. There's not an issue with minpack, the numerical differentiation I was doing wasn't taking advantage of f0,  leading to more function evaluations than necessary. I know how to do a python fix. As mentioned above I'll do it in another PR.

dschmitz89 (approved):
Thanks @andyfaff

PR #22253, opened 2025-01-06
MAINT: Fix a dtype check in `scipy.signal._waveforms.py`

pavyamsiri:
#### Reference issue

Closes #22252.

#### What does this implement/fix?

The old code used to call `dtype.char` and then check if it was contained with a list of a single string `['fdFD']`. This of course would never pass because a `'f'` is not a member of `['fdFD']` but it is a member of `'fdFD'`.

This is likely a mistake in the code and so this commit fixes it.

PR #22520, opened 2025-02-12
MAINT: Remove an extraneous dtype check in `scipy/signal/_waveforms.py`

### LINKED ISSUE #22252: MAINT: Fix a dtype check in `scipy.signal._waveforms.py`
## Describe your issue.

This is not a high priority bug and may potentially be breaking<sup>1</sup>? I discovered after working on [scipy-stubs](https://github.com/jorenham/scipy-stubs), some code in `scipy.signal._waveforms.py` that seem to be buggy (see jorenham/scipy-stubs#195)<sup>2</sup>.

## Relevant Section of Code

See this code under the functions `sawtooth` and `square`.

```python
if t.dtype.char in ['fFdD']:
    ytype = t.dtype.char
else:
    ytype = 'd'
```

The intent of the code is to check that if the array `t` is a `float` or `double` however because the string is within a list this check can never succeed.

## The Fix

The fix is to simply remove the brackets to not put it in a list
```python
if t.dtype.char in 'fFdD':
    ytype = t.dtype.char
else:
    ytype = 'd'
```

I have made an associated PR that does this #22253.

1. Perhaps someone expects the dtype to always be `'d'`?
2. I didn't file an issue then because I didn't want to have to add the overloads to `scipy-stubs`.

pavyamsiri:
#### Reference issue

Closes #22252.

Instead of the PR #22253 which fixes the dtype check so that float32 arrays result in float32 output arrays, this PR just removes the extraneous check entirely and now returns only float64 arrays.

This is a backwards compatible change in contrast to #22253 which does change behaviour.

#### Additional information

I have additionally added tests to make sure that float32 and float64 array inputs result only in float64 array outputs.

### DIFF
--- a/scipy/signal/_waveforms.py
+++ b/scipy/signal/_waveforms.py
@@ -55,11 +55,7 @@ def sawtooth(t, width=1):
     t, w = asarray(t), asarray(width)
     w = asarray(w + (t - t))
     t = asarray(t + (w - w))
-    if t.dtype.char in ['fFdD']:
-        ytype = t.dtype.char
-    else:
-        ytype = 'd'
-    y = zeros(t.shape, ytype)
+    y = zeros(t.shape, dtype="d")
 
     # width must be between 0 and 1 inclusive
     mask1 = (w > 1) | (w < 0)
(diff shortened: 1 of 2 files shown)


ev-br (approved):
LGTM, thank you @pavyamsiri !

Issue #21852, opened 2024-11-09
ENH: bring back nnls.f in v1.15.0


jd-edp:
### Is your feature request related to a problem? Please describe.

I tested the new nnls version written in cython and must say its still slower than the fortran version and is quiet cpu heavy. If you want to split up multiple trainings in many processes you get a complete chaos and nothing is getting finished. I don't know why the nnls.f version is still the best in these things, so it would be easy to bring back the fortran version until there is a new equal version of it.

### Describe the solution you'd like.

bring back nnls.f or a equal fast and low consuming cpu version of it, so I don't have to stick to v1.11.4

j-bowhay:
Cc @ilayn

lucascolley:
could you provide some code and benchmarks which display the slow/heavy behaviour @jd-edp ?

h-vetinari:
> I tested the new nnls version written in cython 

Presumably that refers to #21021, which addressed the very similar #20984.

Can you provide more quantifiable performance metrics? Or better: a reproducer of some code that is fast on 1.11.4 and (still) slow on main.

ilayn:
I checked similar benchmarks on my machine at the time and things were very close to each other. If you are seeing big differences, then might be a cyton thing. 

Or there is a bug remaining that makes the loop spin more.

Either way the code does the same steps so it's probably the second but we need a clean reproducer.

ilayn:
Forgot to mention that this is going to be ported to C anyways since SLSQP is also using it. So it will probably resolve itself in terms of remaining performance differences.

maindlt:
Just in case it is still of interest: we have been using nnls in our galaxy dynamics code for a while and got the following solve times by release:
1.11.4 (Fortran): 1h36m
1.14.1 (Python): 7h02m
1.15.3 (Cython): 7h08m (strange that it's not faster...)
1.16.2 (C): 1h34m

So we are very happy with the new, fast C implementation, keeping the Fortran version available would have saved us much headache after 1.12 came out, though ;-)

ilayn:
Apologies for the screw up. It took quite some time to fix it indeed.

jd-edp:
Yeah i'm also running on the latest version of scipy and its as fast as fortran was. Thank you :)

maindlt:
> Apologies for the screw up. It took quite some time to fix it indeed.

No problem, we figured it out at the end. I think SciPy is terrific package - many thanks for the great job you are doing!!

PR #19121, opened 2023-08-23
WIP:ENH:optimize:Rewrite SLSQP solver

ilayn:
#### Reference issue
Part of #18566 

#### What does this implement/fix?
In the current SQP solver code, there are multiple serious bugs that makes this code unsuitable for production code which I found out completely by accident. A few of them are

1. Probably the obvious one is that the function `linmin` is fully broken because of uninitialized variables due to this initial jump 
https://github.com/scipy/scipy/blob/maintenance/1.11.x/scipy/optimize/slsqp/slsqp_optmz.f#L1660
returning junk depending on the compiler. But since exact line search is typically not used, it stayed under the radar (or that would be my guess). It turns out NLOPT also turned this branch off. Hence that's dead code but pythonized it anyways. When GOTO's are unrolled it just reads quite bad however there is no point losing time on that part.
2. The linear algebra is completely hand-woven. No blas functions are used and not much care is given to arithmetic ops. Related to that it is not difficult to choke the code with large arrays. 
3. The helper functions are mostly trivial given scipy.linalg and numpy already has everything needed at a high level. Depending on the slowdown amount I'll Cythonize/Pythranize the parts if necessary. This is all regular matrix ops hence performance is likely to stay the same as with previous rewrites. I'll rewrite LDL update in a bit more contemporary fashion. 
4. Through random data examples, the accuracy of `lsei` and hence `lsq` is already increased probably due to <strike>blas/lapack</strike> better `nnls` implementation. Thus probably things will get slightly better in terms of solvability.

function conversion progress

-  [ ] slsqp (trivial wrapper)
-  [ ] slsqpb
-  [x] lsq (another not trivial but a bogus wrapper that converts `xl, xu` bounds to `Gx >= h` format)
-  [x] lsei
-  [x] lsi
-  [x] ldp
- [x]  linmin

I'll add docstrings and other details as things shape up.I just wanted to put it here in case someone is interested and have comments in the early stage. 

#### Additional information
I'd appreciate if there are large problems that you know of bigger than a couple of thousand variables to test things out properly.

Issue #18566, opened 2023-05-27
META: FORTRAN Code inventory


ilayn:
This issue is meant for a central place for our Fortran (almost exclusively F77) codebase to track per module and if possible remove depending on availability of contributor affinity of the subject matter and fortran-fu. 

List has projected effort required just based on code length hence involves prejudice.

- [ ] scipy.integrate:
  -  [ ] dop Convert: Easy
  - [ ] mach Convert: Easy
  - [ ] odepack : Convert : Hard
  - [ ] quadpack : Convert : Moderate-Hard
- [ ] scipy.interpolate
  - [ ] fitpack : Convert : Moderate-Hard
- [ ] scipy.io
  - [ ] Test Fortran : ? 
- [x] scipy.linalg
  - [x] lu/det : Replaced in #18225, #18358
  - [x] id_dist : Convert: Moderate (replacement in progress)
- [ ] scipy.odr
  - [ ] odrpack : Convert : Moderate
- [ ] scipy.optimize
  - [ ] nnls: Convert : Easy/Moderate
  - [ ] cobyla : Convert : Hard (The new implementation PRIMA is going to replace cobyla)
  - [ ] L-BFGS-B: Convert : Hard
  - [ ] minpack: Convert : Hard (reimplemented by LFortran team)
  - [ ] minpack2: Convert : Easy
  - [ ] slsqp : Convert: Hard
- [ ] scipy.sparse.linalg
  - [ ] ARPACK Convert : Hard
  - [x] iterative Replaced : #18488
  - [ ] PROPACK : Convert : Hard
- [ ] scipy.special
  - [ ] AMOS: Convert : Hard
  - [ ] cdflib: Convert : Hard
  - [ ] mach: Convert : Easy
  - [ ] specfun : Convert: Very Hard

steppi:
`specfun`actually isn't so bad because each function is mostly self contained, and it can be converted bit by bit. I've already made good headway with `hyp2f1` and that's arguably the most difficult function to replace. I'm determined to banish the F77 from special and have time to get back to it now.

I recently took a close look at ODEPACK to review gh-14552. Converting that will be a bear since it consists of large pieces of machinery with sophisticated control logic. I'd be up for helping to replace it with a C, C++, or Cython implementation if we can find someone else with sufficient, time, knowledge, and interest to work on it together with me. I think it would have to wait until the all of the deliverables are met for the SDG @tirthasheshpatel and I have to work on `special` though.

[...]

PR #13009, opened 2020-10-28
MAINT: clip SLSQP step within bounds

andyfaff:
Addresses a subset of the issues in #11403, see [here](https://github.com/scipy/scipy/issues/11403#issuecomment-717676023) for why this PR is necessary.

PR #22524, opened 2025-02-13
ENH:MAINT:optimize: Rewrite SLSQP and NNLS in C

### LINKED ISSUE #3816: wrap_function seems not to be working when wrapper_args is a one element list
optimize.optimize.py line 274-283

```
x0
Out[179]: array([-1800.])

[d]
Out[180]: [Timestamp('2014-07-23 00:00:00+0200', tz='Europe/Berlin')]

optimal_flows = (opt.fmin_slsqp(func=objective_function, x0=x0,
     ...:                  f_ieqcons=f_ieqcons, bounds=bounds, args=[d],
     ...:                  iter=iter, iprint=0, full_output=False, epsilon=100))

  File "<ipython-input-181-5ade67dafaf1>", line 3, in <module>
    iter=iter, iprint=0, full_output=False, epsilon=100))

  File "C:\Users\uk40291\AppData\Local\Continuum\Anaconda\lib\site-packages\scipy\optimize\slsqp.py", line 200, in fmin_slsqp
    constraints=cons, **opts)

  File "C:\Users\uk40291\AppData\Local\Continuum\Anaconda\lib\site-packages\scipy\optimize\slsqp.py", line 354, in _minimize_slsqp
    fx = func(x)

  File "C:\Users\uk40291\AppData\Local\Continuum\Anaconda\lib\site-packages\scipy\optimize\optimize.py", line 281, in function_wrapper
    return function(*(wrapper_args + args))

TypeError: can only concatenate tuple (not "list") to tuple


debug
> c:\users\uk40291\appdata\local\continuum\anaconda\lib\site-packages\scipy\optimize\optimize.py(281)function_wrapper()
280         ncalls[0] += 1
```

--> 281         return function(*(wrapper_args + args))
    282 

```
ipdb> wrapper_args
(array([-1800.]),)

ipdb> print args
[Timestamp('2014-07-23 00:00:00+0200', tz='Europe/Berlin')]

ipdb> exit
```

### LINKED ISSUE #6689: Use NLopt's version of `slsqp`
NLopt's version of `slsqp`

https://github.com/stevengj/nlopt/tree/master/slsqp

has some improvements over the original Fortran version--there are some bugfixes and the code is reentrant. I had thought we couldn't use it because NLopt was LGPL:

http://ab-initio.mit.edu/wiki/index.php/NLopt

but the Github repo appears to be MIT:

https://github.com/stevengj/nlopt/blob/master/COPYRIGHT

and the `slsqp` part retains its BSD license (inherited from SciPy) anyway:

https://github.com/stevengj/nlopt/blob/master/slsqp/COPYRIGHT

Might be worth replacing SciPy's version? (For concreteness: I volunteer to do it; just want to make sure there's not a reason we shouldn't.)

cc @stevengj

### LINKED ISSUE #7518: DOC: optimize: meaning of accuracy in `fmin_slsqp `undocumented
In the following docstring:

https://docs.scipy.org/doc/scipy-0.19.1/reference/generated/scipy.optimize.fmin_slsqp.html

One of the parameters is ``acc``, which is listed as ``Requested accuracy``. I think this should make clearer whether it is an absolute or relative accuracy.

### LINKED ISSUE #7519: fmin_slsqp not converging to sensible solution
I am trying to use the ``fmin_slsqp`` function while fitting a Gaussian to fake data:

```python
import numpy as np
from scipy.optimize import fmin_slsqp

w = np.linspace(1, 2, 100)
f = 1.2e-8 * np.exp(-0.5 * (w-1.4)**2 / 0.11**2)

def gaussian(p, x):
    return p[0] * np.exp(-0.5 * (x - p[1])**2 / p[2]**2)

def errfunc(p, x, y):
    return np.sum((gaussian(p, x) - y)**2)

p0 = [1.5e-8, 1.6, 0.2]
p_new = fmin_slsqp(errfunc, x0=p0, args=(w, f))

print('p0   ', p0)
print('p_new', p_new)
```

However this does not converge to the expected solution despite the reasonable initial guess:

```
Optimization terminated successfully.    (Exit mode 0)
            Current function value: 4.793585605039844e-15
            Iterations: 1
            Function evaluations: 5
            Gradient evaluations: 1
p0    [1.5e-08, 1.6, 0.2]
p_new [  1.50000000e-08   1.60000000e+00   2.00000000e-01]
```

I suspected that this was due to the ``acc`` tolerance which may be absolute, so I changed it to:

```python
p_new = fmin_slsqp(errfunc, x0=p0, args=(w, f), iter=10000, acc=1.e-40)
```

However while this now does many iterations instead of one, it still does not converge:

```
Iteration limit exceeded    (Exit mode 9)
            Current function value: 2.7710800492819493e-15
            Iterations: 10001
            Function evaluations: 149997
            Gradient evaluations: 10001
p0    [1.5e-08, 1.6, 0.2]
p_new [  2.15505669e-11   1.60000000e+00   2.00000000e-01]
```

It seems like this could be a bug?

I'm using Python 3.6.1 with Scipy 0.19.0 and Numpy 1.11.3.

(cc @ibusko - just so you are in the loop)

### LINKED ISSUE #8065: SLSQP finds minima that aren't minima
In one of my tests of optimize.minimize(), it found minima that are not actually minima.

figure_1-19

figure_1-9

### Reproducing code example:
```
import numpy as np
from numpy import sin
from scipy import optimize

x = np.linspace(10, 30, 1000000)
plt.plot(x, sin(x))

plt.xlim(x.min(), x.max())

for x0 in np.linspace(20.41, 20.43, 100):
    minres = optimize.minimize(sin, x0, method='SLSQP')
    plt.plot(minres.x, minres.fun, 'go', markersize=3.5)
    print(x0, minres.x, minres.success, '\t', minres.message)
```

### Scipy/Numpy/Python version information:
```
0.19.1 1.13.1 sys.version_info(major=3, minor=6, micro=2, releaselevel='final', serial=0)
```

### LINKED ISSUE #10416: ENH: optimize.minimize: slsqp: give better error when work array overflows
My issue is about ```mode``` in _minimize_slsqp  in slsqp.py where when minimize() is called, the key for which exit message is to be used is somehow generated and is not within the exit_modes dictionary and I'm not sure how that's possible. Specifically, -200448291 is somehow defined somewhere, which is the value causing the KeyError

### Reproducing code example:

```
# Where this seems to be failing
res = minimize(opt_funct, x0, method='SLSQP',
               constraints=[equality_constraints,
                            inequality_constraints],
               bounds=bounds,
               options={'ftol': 1e-3, 'eps': 10.0, 'maxiter': 5},
               args=(coef_df,))

# CODE FOR REPRODUCING ERROR

import numpy as np
import pandas as pd
from scipy.optimize import Bounds

bounds = Bounds(np.zeros(66619), np.full(66619, fill_value=1e12))

coef_df = pd.DataFrame(np.random.randint(0,200, size=(66619, 8)))
coef_df = coef_df.rename(columns={7:'coef'})

eq_matrix = np.array(np.random.randint(0,1000, size=(197, 66620)))
ineq_matrix = np.array(np.random.randint(0,1000, size=(42, 66620)))

def opt_funct(x, coef_df):
    """returns the optimization value"""
    return (-100000.0) * np.dot(
        coef_df['coef'], np.sqrt(x))

def _write_constraint_matrix(constr_matrix, x):
    """Create a constraint matrix that can be put into scipy.optimize.minimize

    Parameters
    ----------
    constr : 2-d np.array
        matrix of constraints
    x : np.array
        vector of optimization output values

    Returns
    -------
    vector of values for the inequality or equality constraint
    """
    return np.dot(constr_matrix[:, :-1], x) + constr_matrix[:, -1]


def initialize_x0(eq_matrix, ineq_matrix, coef_df):
    """Create a valid x0 to initialize scipy.optimize.minimize

    Parameters
    ----------
    eq_matrix : 2-d np.array
        matrix of equality constraints
    ineq_matrix : 2-d np.array
        matrix of inequality constraints
    coef_df : pd.DataFrame
        DataFrame that contains model coefficients

    Returns
    -------
    x0 that meets initialization constraints
    """
    # include only minimum constraints which must be met by x0 and
    # convert them to max constraints for initializing x0
    x0_ineq = -1 * ineq_matrix[ineq_matrix[:, -1] < 0, :]
    x0_ineq = np.vstack([eq_matrix, x0_ineq])
    bounds = Bounds(
            np.zeros(coef_df.shape[0]),
            np.full((coef_df.shape[0]), fill_value=1e12))

    inequality_constraints = {'type': 'ineq',
                              'fun': lambda x: np.array(
                                _write_constraint_matrix(
                                  x0_ineq, x
                                  )
                                  ),
                              'jac': lambda x: [y[:-1] for y in x0_ineq]
                              }
    x0 = np.zeros(coef_df.shape[0])
    # TODO replace/remove maxiter and update eps
    res = minimize(fun=opt_funct, x0=x0, method='SLSQP',
                   constraints=inequality_constraints,
                   bounds=bounds,
                   options={'ftol': 1e-3, 'eps': 10000.0,
                            'disp': True, 'maxiter': 1000000},
                   args=(coef_df, ))
    new_x0 = res['x']
    new_x0[new_x0 >= 1e11] = 0
    return new_x0

if __name__ == '__main__:
     x0 = initialize_x0(eq_matrix, ineq_matrix, coef_df)
```

### Error message:

```
  File "<input>", line 6, in <module>
  File "/Users/anaconda/envs/my_env/lib/python3.6/site-packages/scipy/optimize/_minimize.py", line 611, in minimize
    constraints, callback=callback, **options)
  File "/Users/anaconda/envs/my_env/lib/python3.6/site-packages/scipy/optimize/slsqp.py", line 454, in _minimize_slsqp
    message=exit_modes[int(mode)], success=(mode == 0))
KeyError: -200448291
```

### Scipy/Numpy/Python version information:

First discovered on scipy==1.1.0 but also tried on 1.2.0 and 1.3.0
```
numpy==1.16.3
pandas==0.23.4
scipy==1.1.0
python==3.6
```

### LINKED ISSUE #14394: ENH: optmize.slsqp: return Lagrange multipliers
Hello,

I was wondering whether anyone had tested the accuracy of the Lagrange multipliers returned by the minimizer method SLSQP? 

I created a very simple problem of minimizing (x - c)^2 with bounds 1 <= x <= 2. The solution returned by the solver for x is accurate to within the tolerance set. The multipliers I have plotted in the attachment and show a consistent and noticeable offset. I know that the multipliers are not officially returned by the minimizer method but I would expect that accurate multipliers are needed to ensure that the gradient of the Lagrangian is zero at the solution?

Thanks.

multipliers

```
import numpy as np
import scipy.optimize as opt

ineq_cons = {'type' : 'ineq',
             'fun' : lambda x: np.array([x[0] - 1.0, 2.0 - x[0]]),
             'jac' : lambda x: np.array([[1.0], [-1.0]])}

c = np.linspace(0, 3, num = 101)

x0 = np.array([1.5])

x = np.zeros(len(c))
m = np.zeros(len(c))
l = np.zeros((2, len(c)))

for i in range(len(c)):
    fun = lambda x: np.array([(x[0] - c[i]) * (x[0] - c[i])])
    jac = lambda x: np.array([2 * (x[0] - c[i])])
    
    res, w = opt.minimize(fun, x0, method = 'SLSQP', jac = jac, constraints = ineq_cons)
    
    x[i] = res.x
    l[0, i] = w[0]
    l[1, i] = w[1]

    if c[i] < 1:
        m[i] = 2 - 2 * c[i]
    elif c[i] > 2:
        m[i] = 2 * c[i] - 4
    else:
        m[i] = 0.0

plt.title("Min (x - c)^2 s.t. 1 <= x <= 2")
plt.xlabel("c")
plt.ylabel("lambda(c)")
plt.plot(c, l[0,:], 's', label="x >= 1")
plt.plot(c, l[1,:], 'o', label="x <= 2")
plt.plot(c, m, 'k', label="Analytical")
plt.grid()
plt.legend(loc="upper center")
plt.show()
```

```
1.7.0 1.21.0 sys.version_info(major=3, minor=7, micro=7, releaselevel='final', serial=0)
```

### LINKED ISSUE #14915: BUG: optimize.minimize: corruption/segfault with constraints
### Describe your issue.

When minimizing with `scipy.optimize.minimize` it crashes python (kernel), with a corruption error.
This only happens when constraints are active with `SLSQP`.

I also tried scipy 1.7-1.6, but it has the same issue. Also found this issue which might be related: https://github.com/scipy/scipy/issues/14159

### Reproducing Code Example

```python
import numpy as np
import scipy.optimize as sco

x = np.random.rand(22,365)
target = np.linspace(0.9, 4.0, 50)

def metric(v, weights):
    return [[0, 0],[1, 1]]

def efficient_metric(v, target):
    def metric_a(weights):
        return metric(v, weights)[1][0]
    
    def metric_b(weights, v):
        return metric(v, weights)[0][0]

    constraints = ({'type': 'eq', 'fun': lambda x: metric_a(x) - target},
                   {'type': 'eq', 'fun': lambda x: np.sum(x) - 1})
    weights = np.array([len(v)*[1./len(v)]])
    result = sco.minimize(metric_b, weights, args=(v,),
                          method='SLSQP', constraints=constraints)
    return result

efficient_metric(x, target)
```

### Error message

```shell
corrupted size vs. prev_size
OR
corrupted double-linked list
OR
double free or corruption (!prev)
```

1.7.1 1.21.3 sys.version_info(major=3, minor=9, micro=2, releaselevel='final', serial=0)

### LINKED ISSUE #15179: ENH: Adding a gradient tolerance to SLSQP

Various minimization approaches in scipy.optimize.minimize do have the `gtol` option, which generally refers to the length of the gradient vector. One of the most efficient methods, 'SLSQP' does however not. Adding this feature could however be useful as `gtol` is a preferred convergence criteria for some methods.

Currently, I have it hacked in, I pass an option called `gtol` to `_minimize_slsqp` ([line 214](https://github.com/scipy/scipy/blob/bd4b7323acc3837dfa650c140348071a6919614b/scipy/optimize/_slsqp_py.py#L214)) and then I replace [Lines 431 and 433](https://github.com/scipy/scipy/blob/bd4b7323acc3837dfa650c140348071a6919614b/scipy/optimize/_slsqp_py.py#L431-L433) with the following piece of code:

```python
        if mode == -1:  # gradient evaluation required
            g = append(wrapped_grad(x), 0.0)
            a = _eval_con_normals(x, cons, la, n, m, meq, mieq)
            if linalg.norm(g) < gtol:
                print("Converge based upon gradient.")
                break
```

However, there is a drawback to this solution, this is a computation that is performed every iteration. Implementing it in the underlying compiled Fortran code and giving it an exit flag would be more efficient.

In my usecase I can also set `ftol` which is supported by the method. But for the function that I'm minimizing, there is little physical meaning to the function value itself, whereas the gradient has meaning related to stability.

_No response_

### LINKED ISSUE #19130: META: `slsqp` optimizer
Hi all,

after looking at the long list of `scipy.optimize` issues. I thougt it was a good idea to generate a separate tracking issue for the `slsqp` optimizer. Why? There are more than a dozen related issues and `slsqp` is currently being rewritten in Python inhttps://github.com/scipy/scipy/pull/19121. So we could see if the new implementation potentially closes some of them.

#3824
#6689
#7519
https://github.com/scipy/scipy/issues/3816
https://github.com/scipy/scipy/issues/15179
https://github.com/scipy/scipy/issues/14915
https://github.com/scipy/scipy/issues/14394
https://github.com/scipy/scipy/issues/10416
https://github.com/scipy/scipy/issues/8065
https://github.com/scipy/scipy/issues/7518

### LINKED ISSUE #19362: BUG: optimize: warning generated by SLSQP is useless
https://github.com/scipy/scipy/blob/166e1f2b1ea0a1a2c3d7b030bd829549f8a5844a/scipy/optimize/_optimize.py#L422-L429

The above code generates a warning for some calls to SLSQP (see https://github.com/scipy/scipy/issues/11403#issuecomment-717676023). 

Users cannot do anything with this warning, as the reason for the warning is due to the implementation of SLSQP. The warning is also not informative as the result is correct always, so the message is not warning about any potential problem.

Perhaps scipy developers want to keep the warning during development in case it may catch some unexpected effects of a patch, but it should not be given to users.

ilayn:
This PR introduces new C code both for F77 SLSQP and the previously translated NNLS. (Superseeds #19121)

#### NNLS changes
 - Rewritten to match F77 code closer so that SLSQP does not diverge from the F77 implementation
 - Removed the Cython code
 - Added some more comments
 - Nothing too different than the Cython code but it has dramatically increased in speed back to old F77 speed so implicitly fixing #21852
 - Tidied up the error codes to match the SLSQP code

#### SLSQP changes
 - Added wrappers for the internal functions and tested them offline for various test cases, in case I haven't missed a particular edge case, all pathological cases agree
 - Provided dummy wrappers for them in case we need them later, for the release, the compiler will optimize them out (they are removed from the merge but available in the previous commits).
 - Internal functions have the same speed with F77 code only winning slightly
 - Also tested with the previous Python code #19121 
 - Recounted all the buffer sizes and corrected a few places such that segfaults don't happen anymore 
 - Internal functions are used only as SLSQP uses them. Original F77 functions have a few fundamental limitations and minor bugs that SLSQP code guarantees not hitting them.
 - Placed some comments where I can decrypt and recognize the steps at the same time
 - F77 code is removed

#### Reference issue
Towards #18566 

I'll just copy/paste the list @dschmitz89 collated in #19130 

Closes #6689 
Closes #7518 
Closes #10416
Closes #14915 
Closes #19362

- [ ] #19130 : META issue for the SLSQP issues
- [ ] #3816  : list for the `args` argument needs to be coerced into tuple
- [ ] #3824  : undocumented acc 
- [ ] #7519  : Unscaled data does not converge
- [ ] #8065  : `slsqp` converges to maximum, not minimum
- [ ] #14394 : asks to return Lagrange multipliers
- [ ] #15179 : Gradient tolerance parameter request

#### Additional information
I did not mean to open the PR prematurely but I think there are some design issues that we have been suffering, for example https://github.com/scipy/scipy/pull/13009#issuecomment-717703198 from @andyfaff  hence if anything needed to be done, please let me know so I can incorporate.


mdhaber:
I'll add check boxes above so we don't forget that this PR may not actually close all those issues. Some of them are feature requests, and without testing, we won't know whether some of the bug fixes are fixed by the translation or if they are inherent to the algorithm.

scipy/optimize/__slsqp.c
```diff
@@ -0,0 +1,527 @@
...
+
+    // Copy multipliers from scratch to gmults
+    for (int i = 0; i < mg; i++) { gmults[i] = lsi_scratch[i]; }
+
+    // If no equality constraints this was an LSI problem all along.
+    if (me == 0) { return; }
+
+    t = dnrm2_(&me, &x[nvars], &one);
+    // Modify the norm by adding the equality solution.
+    *xnorm = sqrt((*xnorm)*(*xnorm) + t*t);
```
dschmitz89, line 359: Could this be replaced with `hypot`?

dschmitz89 (commented):
Impressive! Still, would it make sense to do `nnls` first and then a second PR for `slsqp`? The previous `nnls` rewrites caused quite some churn.

ilayn:
> Some of them are feature requests, and without testing, we won't know whether some of the bug fixes are fixed by the translation or if they are inherent to the algorithm.

That's the idea. Obviously there will be tests later once I get some feedback. If we don't have any changes to be done then it closes all of them since I have looked into them. So not sure if I am getting the point. 

> The previous nnls rewrites caused quite some churn.

Yes and this is still the corrected version. I just changed the language. If there is a change, SLSQP also goes down under spectacularly.

scipy/optimize/__slsqp.c
ilayn: Yes I think so.

scipy/optimize/__nnls.c
```diff
@@ -0,0 +1,267 @@
...
+            pivot = a[indz + j*m];
+            tmpint = m - indz;
+            dlarfgp_(&tmpint, &pivot, &a[indz + 1 + j*m], &one, &tau);
+
+            // Compute the norm of a[0:indz, j] to check for linear dependence.
+            unorm = (indz > 0 ? dnrm2_(&indz, &a[j*m], &one) : 0.0);
+
+            // Test for independence by checking the pivot for zero.
+            // Note, e.g., (100. + 1e-15) - 100. == 0.0 in floating point arithmetic.
+            if (((unorm + fabs(pivot)*0.01) - unorm) > 0.0)
```
dschmitz89, line 82: I guess this is a translation of the F77 version? Is it supposed to be a check for `pivot > eps`?

ilayn: That's actually a very sneaky tactic when you don't have `np.spacing()` function at your disposal. Basically what they are trying to do is a relative tolerance with the powers of 2. 

The classical example you can test yourself too is the following 

```python
>>> import numpy as np
>>> zzz = np.float64(1234567891011.123456789)

>>> zzz + 1e-6 == zzz
True

>>> zzz + 1e-5 == zzz
True

>>> zzz + 1e-1 == zzz
False
```

If we just use a fixed tolerance then we would miss these. But if we used just the fixed relative tolerance we will miss it when things are small. This is a quite old tactic to get spacing of the numbers at that particular power of 2. So in other words

```python
>>> s = np.spacing(zzz)
>>> s
np.float64(0.000244140625)

>>> s / zzz
np.float64(1.9775390788759812e-16)  # Note this is smaller than np.finfo(dtype).eps ~2.2e-16

>>> np.log2(s)
np.float64(-12.0)
```

Though, `np.spacing` is <strike>stolen</strike>lifted from Fortran since it also has `SPACING` function somehow the old foxes did not use it in their code. This is another reason why I want to write unit tests using spacing. It has a more uniform tolerance value and still relative to the magnitude of the scalar.

mdhaber:
> If we don't have any changes to be done then it closes all of them since I have looked into them

Perhaps I'm missing them, but does this add the Lagrange multipliers to the output to address gh-14394 or the `gtol` option for gh-15179? I didn't see those two or any tests, which made me wonder about the rest. That's great if this addresses all that, though.

ilayn:
Returning the multipliers at the low level is quite easy since they are always computed. The more important part is if the API wants to expose it. I think we can have that conversation once everything is in place. I'm more interested in whether you want to do something with the bounds or API changes if things are going to be much easier if we change the calling structure etc. I did not follow the `optimize` discussions lately hence my question. 

If you think there are ways in which you don't need to form big arrays or whatever I think I can accomodate for it now that I'm a bit more familiar with its inner workings.

scipy/optimize/__nnls.c
dschmitz89: Whoa, quite interesting. We should gather such tips and tricks somewhere. Still is there a more modern alternative to achieve the same thing in C?

ilayn: If you hold your nose, yes but it is not very different than this just more edge cases 

https://github.com/numpy/numpy/blob/a651643fc9b5699a6dc6f3c85ba499c1f52ce3aa/numpy/_core/src/npymath/ieee754.cpp#L29

ilayn:
@j-bowhay This is getting a test failure of 

```
FAILED scipy/_lib/tests/test_public_api.py::test_private_but_present_deprecation[scipy.optimize.slsqp-None] - AttributeError: module 'scipy.optimize._slsqp_py' has no attribute 'zeros'
```

But I only removed `numpy.zeros` usage from `slsqp` however I can't find where this is actually defined. Could you help me take a look why it is confusing with `scipy.optimize.zeros` ?

ilayn:
Also weirdly enough some tests can't find `numpy.concat`

scipy/optimize/slsqp.py
```diff
@@ -9,7 +9,6 @@
     'OptimizeResult',
     'fmin_slsqp',
     'slsqp',
-    'zeros',
```
j-bowhay, line 12: Yep this looks like the right fix :+1:

ilayn: Ah thanks, I found it after the fact. But good to have confirmation.

ilayn:
Other than the tolerance issue it seems like, this is ready for review. We can start addressing the linked issues.

ilayn:
Something happened to mypy jobs as these errors are not emitted from this PR but from others too. Other than that they are now all passing.

ilayn:
@anugrahjo In case you are interested, SLSQP code is undergoing some facelift hence the gentle ping in case you were touching the internals for pySLSQP.

scipy/optimize/_nnls.py
```diff
@@ -72,11 +72,13 @@ def nnls(A, b, *, maxiter=None, atol=_NoValue):
...
-        raise ValueError("Expected a two-dimensional array (matrix)" +
-                         f", but the shape of A is {A.shape}")
-    if len(b.shape) != 1:
-        raise ValueError("Expected a one-dimensional array (vector)" +
-                         f", but the shape of b is {b.shape}")
+        raise ValueError(f"Expected a 2D array, but the shape of A is {A.shape}")
+
+    if (b.ndim > 2) or ((b.ndim == 2) and (b.shape[1] != 1)):
+        raise ValueError("Expected a 1D array,(or 2D with one column), but the,"
+                         f" shape of b is {b.shape}")
```
dschmitz89, line 79: Before , `b` was allowed to be only 1D, this would allow also `(n, 1)`. Why should we change this?

scipy/optimize/_slsqp_py.py
```diff
@@ -393,125 +384,161 @@ def cjac(x, *args):
...
+    #               the problem is augmented with a regularizing dummy variable.
+    # reset: holds the count of resetting bfgs to identity matrix.
+    # iter  : the current and itermax is the maximum number of iterations.
+    # line  : the current line search iteration.
+    # mode  : the exit mode of the solver.
+    # alpha, f0, gs, h1, h2, h3, h4, t, t0 : internal variables used by the solver.
+    #
+    # The dict holds the intermediate state of the solver. The keys are the same
+    # as the C struct members and will be modified in-place.
+    Vars = {
```
dschmitz89, line 408: Nitpick: why capitalized?

dschmitz89 (commented):
I cannot comment on the intrinsics of the C code. We will probably review this based on benchmarks and tests mostly. All existing tests pass from what I see, benchmarks we should have a look at. I assume, that for https://github.com/scipy/scipy/issues/14915 we can add a regression test?

Besides that, one nitpick about capitalization and one question about an in my opinion unnecessary change to the `nnls` part.

scipy/optimize/_nnls.py
ilayn: Just consistency in terms of problem definition. Since we do this everywhere else, I did not think of any reason to reject it because `Ax = b` is consistent for such `b`. Sometimes it might be coming from another piece of code that slices a column and it's just these extra unnecessary disruptions that annoy users the most. I don't have strong opinions though either way is fine for me.

scipy/optimize/_slsqp_py.py
ilayn: Too much C coding. On the C side it is a struct (poor man's Class object). I'll fix.

scipy/optimize/_nnls.py
dschmitz89: Ah, in `linalg`, this shape is typically allowed? Then I agree, let's do so also for `nnls`. Just let's leave it for a followup and add tests if that is alright for you?

ilayn:
I moved the bounds check from #13009 to C code so that it can never violate bounds when leaving the C level. Hence removed the extra guardrails from the Python side. Not sure if this grants us the freedom to remove the bounds check from the `lsq` internals.

ilayn:
In terms of performance, I spot checked a few places, they are always compatible except when the problem size grows to 100 for some reason C code starts to win. Might be that the Python side modifications are becoming dominant. 

I don't have much steam left how to do a proper performance check for this. I tried to use the CUTest suite or https://github.com/GrattonToint/S2MPJ but could not make it work so if anyone has any appetite, it would be great.

ewu63:
@ilayn No promises but I have CUTEst set up here so I'll try to run some comparisons in the next few days.

dschmitz89:
Our own benchmarks should suffice. Looking at the [last run](https://app.circleci.com/pipelines/github/scipy/scipy/33754/workflows/96ae996b-9542-4995-9312-d7e7a41a6915/jobs/114440/timing), I guess they were cancelled?

ilayn:
Thanks @ewu63 regardless of success, much appreciated.

> Looking at the [last run](https://app.circleci.com/pipelines/github/scipy/scipy/33754/workflows/96ae996b-9542-4995-9312-d7e7a41a6915/jobs/114440/timing), I guess they were cancelled?

Last one is a linter commit clicking on the previous commit cross for the full run.

dschmitz89:
> Thanks @ewu63 regardless of success, much appreciated.
> 
> > Looking at the [last run](https://app.circleci.com/pipelines/github/scipy/scipy/33754/workflows/96ae996b-9542-4995-9312-d7e7a41a6915/jobs/114440/timing), I guess they were cancelled?
> 
> Last one is a linter commit, click on the previous commit cross for the full run.

Ah, great. The section we need to look at are the DFO benchmarks ([URL](https://circleci.com/api/v1.1/project/github/scipy/scipy/114431/output/104/0?file=true&allocation-id=67dee0ce30e94f45e1d58e50-0-build%2FABCDEFGH)):

image

Will compare the results against main this week unless someone beats me to it.

ilayn:
OK flexing some regex muscle 

<details>

```
DFO    solver      result type           THIS PR                     MAIN
--------------- --------------      ----------------       ---------------------
 0     SLSQP        mean_nfev             21.0                      21.0          
 0     SLSQP         min_obj       35.99999999999999         35.99999999999999
 1     SLSQP        mean_nfev             21.0                      21.0          
 1     SLSQP         min_obj              36.0                      36.0      
 2     SLSQP        mean_nfev             23.0                      23.0          
 2     SLSQP         min_obj       8.380281690140846         8.380281690140846    
 3     SLSQP        mean_nfev             23.0                      23.0          
 3     SLSQP         min_obj       8.380284602260867         8.380284602260867    
 4     SLSQP        mean_nfev             23.0                      23.0          
 4     SLSQP         min_obj       9.880597014925373         9.880597014925373    
 5     SLSQP        mean_nfev             23.0                      23.0          
 5     SLSQP         min_obj       9.880597015430164         9.880597015430164    
 6     SLSQP        mean_nfev            109.0                     109.0          
 6     SLSQP         min_obj     3.536662171120014e-05     3.534301293924672e-05  
 7     SLSQP        mean_nfev            302.0                     295.0          
 7     SLSQP         min_obj     3.8304842753951935e-05    2.4772818584112955e-05 
 8     SLSQP        mean_nfev            112.0                     112.0          
 8     SLSQP         min_obj     2.4611393473318765e-06    2.468294162292619e-06  
 9     SLSQP        mean_nfev            114.0                     114.0          
 9     SLSQP         min_obj     3.152591755101908e-06     3.1523978560448592e-06 
 10    SLSQP        mean_nfev             99.0                      99.0          
 10    SLSQP         min_obj     9.231469568590042e-05     9.231453061284653e-05  
 11    SLSQP        mean_nfev            179.0                     179.0          
 11    SLSQP         min_obj     3.4343375187151474e-05     3.44391715908274e-05  
 12    SLSQP        mean_nfev             28.0                      28.0          
 12    SLSQP         min_obj       48.984254140285884         48.9842541402506    
 13    SLSQP        mean_nfev             47.0                      47.0          
 13    SLSQP         min_obj       48.984285105690184        48.984285104912686   
 14    SLSQP        mean_nfev             36.0                      36.0          
 14    SLSQP         min_obj      0.010926876667685267      0.010926876593972496  
 15    SLSQP        mean_nfev             80.0                      80.0          
 15    SLSQP         min_obj      0.01085244432931276       0.010852444332488186  
 16    SLSQP        mean_nfev             22.0                      22.0          
 16    SLSQP         min_obj     0.0005011638819800904     0.0005011638818990982  
 17    SLSQP        mean_nfev            102.0                     150.0          
 17    SLSQP         min_obj       1417868770.8854303        1417869003.358761    
 18    SLSQP        mean_nfev            145.0                     145.0          
 18    SLSQP         min_obj     0.0031019726400799996     0.0031019739183846718  
 19    SLSQP        mean_nfev            334.0                     334.0          
 19    SLSQP         min_obj     0.0023889121209729348      0.002387635461912692  
 20    SLSQP        mean_nfev            435.0                     435.0          
 20    SLSQP         min_obj     3.7457280155552515e-05    3.7403380991915257e-05 
 21    SLSQP        mean_nfev            1023.0                    975.0          
 21    SLSQP         min_obj      0.005610846479511704      0.005234314850263207  
 22    SLSQP        mean_nfev            627.0                     613.0          
 22    SLSQP         min_obj      0.007587777492846376      0.007622482850835413  
 23    SLSQP        mean_nfev            1344.0                    1331.0         
 23    SLSQP         min_obj      0.08092297298795201       0.10106796888625674   
 24    SLSQP        mean_nfev             85.0                      85.0          
 24    SLSQP         min_obj      0.03902392303645375       0.03902328171745464   
 25    SLSQP        mean_nfev             16.0                      16.0          
 25    SLSQP         min_obj             2020.0                    2020.0         
 26    SLSQP        mean_nfev            130.0                     130.0          
 26    SLSQP         min_obj       85822.20163045886         85822.20163050466    
 27    SLSQP        mean_nfev            252.0                     252.0          
 27    SLSQP         min_obj       85822.20162647945         85822.20162647964    
 28    SLSQP        mean_nfev             73.0                      73.0          
 28    SLSQP         min_obj     0.00012798947628289288    0.0001275291143854763  
 29    SLSQP        mean_nfev             58.0                      58.0          
 29    SLSQP         min_obj     0.0001779226297279068     0.00017792262972978623 
 30    SLSQP        mean_nfev             72.0                      72.0          
 30    SLSQP         min_obj      0.006287595198804419      0.006287589283329275  
 31    SLSQP        mean_nfev            113.0                     113.0          
 31    SLSQP         min_obj     7.559378770788214e-05     7.559376195674753e-05  
 32    SLSQP        mean_nfev             76.0                      76.0          
 32    SLSQP         min_obj      0.007482970142625719      0.00748298302440139   
 33    SLSQP        mean_nfev             80.0                      80.0          
 33    SLSQP         min_obj      0.006136907945763587      0.006136919804702198  
 34    SLSQP        mean_nfev             47.0                      47.0          
 34    SLSQP         min_obj     1.820222566350904e-05     1.8202225661529484e-05 
 35    SLSQP        mean_nfev             74.0                      74.0          
 35    SLSQP         min_obj       1.105662922842976         1.1056629223458196   
 36    SLSQP        mean_nfev            426.0                     426.0          
 36    SLSQP         min_obj       0.041404387368612        0.04140438887707551   
 37    SLSQP        mean_nfev            228.0                     228.0          
 37    SLSQP         min_obj       1.7898252567696078        1.7898252426543697   
 38    SLSQP        mean_nfev            216.0                     216.0          
 38    SLSQP         min_obj       10.23899469997763         10.238994699907447   
 39    SLSQP        mean_nfev            276.0                     276.0          
 39    SLSQP         min_obj       18.281172707850697        18.281172707363083   
 40    SLSQP        mean_nfev            398.0                     398.0          
 40    SLSQP         min_obj        22.2606981670772          22.2606981682437    
 41    SLSQP        mean_nfev            367.0                     367.0          
 41    SLSQP         min_obj       26.27277447070526         26.272774470672505   
 42    SLSQP        mean_nfev            331.0                     337.0          
 42    SLSQP         min_obj     0.0031827413798409215      0.002869140620444522  
 43    SLSQP        mean_nfev            401.0                     395.0          
 43    SLSQP         min_obj     0.0026572704164352107     0.0027446414753164675  
 44    SLSQP        mean_nfev            535.0                     535.0          
 44    SLSQP         min_obj     0.0031750456086234386     0.0032067490009361457  
 45    SLSQP        mean_nfev             89.0                      89.0          
 45    SLSQP         min_obj     1.191477661292291e-07     1.191747264184552e-07  
 46    SLSQP        mean_nfev             84.0                      84.0          
 46    SLSQP         min_obj     9.955078050077315e-10     9.996542865405878e-10  
 47    SLSQP        mean_nfev            156.0                     156.0          
 47    SLSQP         min_obj     8.585763340052324e-10     8.593657918793688e-10  
 48    SLSQP        mean_nfev            139.0                     139.0          
 48    SLSQP         min_obj      0.003432184872607524     0.0034359354793408884  
 49    SLSQP        mean_nfev            304.0                     304.0          
 49    SLSQP         min_obj     2.468593000190951e-06     2.6786053785224307e-06 
 50    SLSQP        mean_nfev            174.0                     174.0          
 50    SLSQP         min_obj       16411.760641336634        16090.611617512228   
 51    SLSQP        mean_nfev            321.0                     321.0          
 51    SLSQP         min_obj      0.003924544260296388      0.003924596822050221  
 52    SLSQP        mean_nfev            998.0                     993.0          
 52    SLSQP         min_obj       331.9612969603753         306.9740542671161    
```

</details>

Problem 52 of DFO seems a bit weird and problem 42 looks like an early exit, but other than that things look OK. The other smooth unbounded benchmark


```
                 test function          solver      result type           THIS PR                     MAIN
             ---------------------- -------------- ------------- ------------------------   ------------------------

                rosenbrock_slow         SLSQP        mean_nfev            154.9                      155.1          
                rosenbrock_slow         SLSQP        mean_time     0.023609375953674315       0.026223421096801758  
               rosenbrock_nograd        SLSQP        mean_nfev            154.9                      155.1          
               rosenbrock_nograd        SLSQP        mean_time     0.007560014724731445       0.008600091934204102  
                   rosenbrock           SLSQP        mean_nfev            102.2                      102.3          
                   rosenbrock           SLSQP        mean_time     0.005167090892791748      0.0065847992897033695  
                rosenbrock_tight        SLSQP        mean_nfev            108.55                     108.15         
                rosenbrock_tight        SLSQP        mean_time     0.005497491359710694       0.008199214935302734  
                simple_quadratic        SLSQP        mean_nfev             6.0                        6.0           
                simple_quadratic        SLSQP        mean_time    0.00040684938430786134     0.00048006772994995116 
              asymmetric_quadratic      SLSQP        mean_nfev             6.0                        6.0           
              asymmetric_quadratic      SLSQP        mean_time    0.0004140615463256836      0.0004865169525146484  
                     sin_1d             SLSQP        mean_nfev             6.95                       6.95          
                     sin_1d             SLSQP        mean_time    0.0005295634269714356      0.0006420612335205078  
                     booth              SLSQP        mean_nfev             12.3                       12.3          
                     booth              SLSQP        mean_time    0.0006984472274780273       0.000834810733795166  
                     beale              SLSQP        mean_nfev             25.9                       25.9          
                     beale              SLSQP        mean_time    0.0010523676872253418       0.001302945613861084  
                       LJ               SLSQP        mean_nfev            179.45                     206.4          
                       LJ               SLSQP        mean_time     0.009283041954040528       0.012150728702545166  
```

dschmitz89:
I asked chatGPT to do the comparison for me. Overall, differences are indeed very minor @ilayn . Based on benchmarks, the C translation should be ok imo. Last blocker for me is https://github.com/scipy/scipy/pull/22524#discussion_r2005540498.

<p>

```

DFO    solver      result type           THIS PR                     MAIN          Comparison
--------------- --------------      ----------------       ---------------------  -------------
 0     SLSQP        mean_nfev             21.0                      21.0          Equal
 0     SLSQP         min_obj       35.99999999999999         35.99999999999999   Equal
 1     SLSQP        mean_nfev             21.0                      21.0          Equal
 1     SLSQP         min_obj              36.0                      36.0          Equal
 2     SLSQP        mean_nfev             23.0                      23.0          Equal
 2     SLSQP         min_obj       8.380281690140846         8.380281690140846   Equal
 3     SLSQP        mean_nfev             23.0                      23.0          Equal
 3     SLSQP         min_obj       8.380284602260867         8.380284602260867   Equal
 4     SLSQP        mean_nfev             23.0                      23.0          Equal
 4     SLSQP         min_obj       9.880597014925373         9.880597014925373   Equal
 5     SLSQP        mean_nfev             23.0                      23.0          Equal
 5     SLSQP         min_obj       9.880597015430164         9.880597015430164   Equal
 6     SLSQP        mean_nfev            109.0                     109.0          Equal
 6     SLSQP         min_obj     3.536662171120014e-05     3.534301293924672e-05  Higher in PR
 7     SLSQP        mean_nfev            302.0                     295.0          Higher in PR
 7     SLSQP         min_obj     3.8304842753951935e-05    2.4772818584112955e-05 Higher in PR
 8     SLSQP        mean_nfev            112.0                     112.0          Equal
 8     SLSQP         min_obj     2.4611393473318765e-06    2.468294162292619e-06  Lower in PR
 9     SLSQP        mean_nfev            114.0                     114.0          Equal
 9     SLSQP         min_obj     3.152591755101908e-06     3.1523978560448592e-06 Higher in PR
10     SLSQP        mean_nfev             99.0                      99.0          Equal
10     SLSQP         min_obj     9.231469568590042e-05     9.231453061284653e-05  Higher in PR
11     SLSQP        mean_nfev            179.0                     179.0          Equal
11     SLSQP         min_obj     3.4343375187151474e-05     3.44391715908274e-05  Lower in PR
12     SLSQP        mean_nfev             28.0                      28.0          Equal
12     SLSQP         min_obj       48.984254140285884         48.9842541402506    Higher in PR
13     SLSQP        mean_nfev             47.0                      47.0          Equal
13     SLSQP         min_obj       48.984285105690184        48.984285104912686   Higher in PR
14     SLSQP        mean_nfev             36.0                      36.0          Equal
14     SLSQP         min_obj      0.010926876667685267      0.010926876593972496  Higher in PR
15     SLSQP        mean_nfev             80.0                      80.0          Equal
15     SLSQP         min_obj      0.01085244432931276       0.010852444332488186  Lower in PR
16     SLSQP        mean_nfev             22.0                      22.0          Equal
16     SLSQP         min_obj     0.0005011638819800904     0.0005011638818990982  Higher in PR
17     SLSQP        mean_nfev            102.0                     150.0          Lower in PR
17     SLSQP         min_obj       1417868770.8854303        1417869003.358761    Lower in PR
18     SLSQP        mean_nfev            145.0                     145.0          Equal
18     SLSQP         min_obj     0.0031019726400799996     0.0031019739183846718  Lower in PR
19     SLSQP        mean_nfev            334.0                     334.0          Equal
19     SLSQP         min_obj     0.0023889121209729348      0.002387635461912692  Higher in PR
20     SLSQP        mean_nfev            435.0                     435.0          Equal
20     SLSQP         min_obj     3.7457280155552515e-05    3.7403380991915257e-05 Higher in PR
21     SLSQP        mean_nfev            1023.0                    975.0          Higher in PR
21     SLSQP         min_obj      0.005610846479511704      0.005234314850263207  Higher in PR
22     SLSQP        mean_nfev            627.0                     613.0          Higher in PR
22     SLSQP         min_obj      0.007587777492846376      0.007622482850835413  Lower in PR
23     SLSQP        mean_nfev            1344.0                    1331.0         Higher in PR
23     SLSQP         min_obj      0.08092297298795201       0.10106796888625674   Lower in PR
24     SLSQP        mean_nfev             85.0                      85.0          Equal
24     SLSQP         min_obj      0.03902392303645375       0.03902328171745464   Higher in PR
25     SLSQP        mean_nfev             16.0                      16.0          Equal
25     SLSQP         min_obj             2020.0                    2020.0         Equal
26     SLSQP        mean_nfev            130.0                     130.0          Equal
26     SLSQP         min_obj       85822.20163045886         85822.20163050466    Lower in PR
27     SLSQP        mean_nfev            252.0                     252.0          Equal
27     SLSQP         min_obj       85822.20162647945         85822.20162647964    Lower in PR
28     SLSQP        mean_nfev             73.0                      73.0          Equal
28     SLSQP         min_obj     0.00012798947628289288    0.0001275291143854763  Higher in PR
29     SLSQP        mean_nfev             58.0                      58.0          Equal
29     SLSQP         min_obj     0.0001779226297279068     0.00017792262972978623 Lower in PR

``` 
</p>
</details>

dschmitz89:
> Thanks for the check.
> 
> We don't break anything here just allowing 2D singletons so I think that's pretty safe operation because it doesn't let any 2D in and we have quite a stable prior art in linalg. Are you OK if handle it here since it's already here?

Ok, let's handle it here, could you just add a simple test that 2D singletons and 1D arrays produce the same result with `nnls`?

ewu63:
Here are some CUTEst results, primarily because I wasn't sure whether the benchmarks cover a wide enough range of optimization problems. I ran a total of 677 problems from the benchmark, comparing this branch vs. v1.15.2. All the problems ran with explicit Jacobians computed by pyCUTEst, without any finite differencing by scipy, and I capped it at 500 iterations.

```
Number of problems with the same status: 657/677
Number of problems with the same fun value (with tol=1E-3): 605/677
Number of problems with the same nfev: 571/677
```
Each row is a subset of the previous set. This means that 571 problems behaved pretty much exactly like the previous implementation, which I think is pretty great news. 

Out of those that converged to the same answer but took different number of iterations, there is a large variance. Some took much fewer, while others took substantially more, and some were only off by a few. Perhaps this result is expected.

For those problems that converged to a different function value (but same exit code), many of the results were actually quite different (often >100%). It's not clear to me whether these are benign issues, e.g. maybe the problems were multimodal and SLSQP converged to a different optimum.

Finally, we have the set of problems that resulted in a different exit code. The breakdown of exit codes (ordered `old_status -> new_status: #occurances`) are as follows:
```
0->5: 8
4->0: 5
8->4: 2
0->4: 2
4->8: 1
6->0: 1
6->4: 1
```
Given that 10 problems used to converge but now fails, vs. 6 vice versa, it is pretty even here. Overall I meant this as a general sanity check against a more representative and large set of problems, and the success rate of 84% with perfect match and 89% reaching the same optimum is honestly quite good. Hopefully this result is somewhat useful here :)

ilayn:
@ewu63 This is fantastic thank you for your efforts. It's **very** helpful. 

For the `0->5`, I can't if it is all of them but some are due to an obvious facepalming bug which I now see thanks to your analysis. I don't know how it got in there but I'll fix it. The returning code is wrong 🤦‍♂️ 

For the `0->4` SLSQP had its own old version of NNLS and it is a bit more relaxed than the more recent one (still decades old). So probably that is playing a role there.

dschmitz89:
The 32bit CI failure is related to the new NNLS test.

@rgommers @andyfaff : would you have time to look over this massive PR also? It is the last missing piece to remove F77 from `optimize`. To aid review: an extensive benchmark was carried out in https://github.com/scipy/scipy/pull/22524#issuecomment-2753322913 which unearthed one last bug from what I understood.

scipy/optimize/__slsqp.h
```diff
@@ -0,0 +1,405 @@
...
+    // Copy the data from the numpy array
+    for (int j = 0; j < n; j++) {
+        for (int i = 0; i < m; i++) {
+            a[i + j*m] = data_A[(j*strides[1] + i*strides[0])/sizeof(double)];
+        }
+    }
+    for (int i = 0; i < m; i++)
+    {
+        b[i] = data_b[(i * rc_stride)/sizeof(double)];
+    }
```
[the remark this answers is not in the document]
ilayn: If b is noncontiguous it will not work properly. This ensures a contiguous copy.

ilayn:
I don't know what happened to the stats test in the meantime. But optimize parts are OK.

lucascolley:
(by the way, I realised that we don't need `[skip cirrus]` any more, since gh-22446 :) )

lucascolley:
suspiciously off by almost exactly a factor of 10?
```
________________________ test_weightedtau_vs_quadratic _________________________
scipy/stats/tests/test_stats.py:2129: in test_weightedtau_vs_quadratic
    assert_approx_equal(expected, actual)
E   AssertionError: 
E   Items are not equal to 7 significant digits:
E    ACTUAL: 5.836994531322873e-17
E    DESIRED: 5.836994531322869e-16
        _          = 1
        a          = [1, 2, 3, 3, 2, 3]
        actual     = np.float64(5.836994531322869e-16)
        add        = False
        b          = [2, 3, 3, 3, 2, 1]
        expected   = np.float64(5.836994531322873e-17)
        i          = 3
        rank       = array([0, 2, 4, 5, 3, 1])
        s          = 4
        weigher    = <function test_weightedtau_vs_quadratic.<locals>.weigher at 0x3cad6060d20>
        wkq        = <function test_weightedtau_vs_quadratic.<locals>.wkq at 0x3cad6061c20>
```

ilayn:
Free threading job is a bit funky, occasionally tripping up on different tests. Let me rerun it.

> (by the way, I realised that we don't need `[skip cirrus]` any more, since https://github.com/scipy/scipy/pull/22446 :) )

Ha! cargo cult programming on my side, thanks.

ilayn:
And now it is gone. I have the feeling that something is going on in this free-threaded job that we have yet to uncover.

mdhaber:
I saw that failure in gh-22695, too, btw.

scipy/optimize/__slsqp.h
```diff
@@ -0,0 +1,405 @@
...
+    #define STRUCT_FIELD_NAMES STRUCT_INT_FIELD_NAMES STRUCT_DOUBLE_FIELD_NAMES
+
+    // Parse the dictionary, if the field is not found, raise an error.
+    // Do it separately for doubles and ints.
+    // Initialize the struct that will be populated from dict with zeros
+    #define X(name) Vars.name = 0;
+    STRUCT_FIELD_NAMES
+    #undef X
+
+    // PyDict_GetItemString returns a borrowed reference.
```
rgommers, line 246: Is this cargo culted from somewhere, or newly written? It looks a little funky. Probably because `SLSQP_static_vars` is a pretty unusual struct. Also, why the `static` in the name?

scipy/optimize/__slsqp.h
```diff
@@ -0,0 +1,405 @@
...
+    int exact, inconsistent, reset, iter, itermax, line, m, meq, mode, n;
+};
+
+void __slsqp_body(struct SLSQP_static_vars* S, double* funx, double* gradx, double* C, double* d, double* sol, double* mult, double* xl, double* xu, double* buffer, int* indices);
+
+
+static PyObject*
+nnls(PyObject* Py_UNUSED(dummy), PyObject* args) {
+
+    int maxiter, info = 0;
```
rgommers, line 84: Should implementations be moved to `__slsqp.c`? That's where I would expect them, and this header is only included in that one C file. `_nnls.h` looks more pleasant to read.

scipy/optimize/meson.build
```diff
@@ -145,10 +135,14 @@ py3.extension_module('_bglu_dense',
...
-py3.extension_module('_cython_nnls',
-  opt_gen.process('_cython_nnls.pyx'),
-  c_args: cython_c_args,
-  dependencies: np_dep,
+py3.extension_module('_slsqplib',
+  [
+    '__slsqp.h',
+    '__slsqp.c',
+    '__nnls.h',
+    '__nnls.c'
```
rgommers, line 143: Are the double underscores on the functions needed? Why not a single underscore?

rgommers (commented):
I only skimmed the code really (too much to be able to review in detail). This looks like another good step forward (🚀). The C code isn't the prettiest with `goto`'s etc., but I'm sure you already knew that - it's certainly much easier to work on it in the future compared to the Fortran code.

I added only a couple of comments. The test suite being pretty much unchanged and extra benchmarking passing give good confidence. PRs like these are hard to review in detail, but that's also not always needed (or even feasible) - I'd be happy to see this go in soon.

scipy/optimize/__slsqp.h
ilayn: I wrote it for maintenance reasons actually. But nice that it sounded like AI Slop hahah. 

Specifically, we are communicating with the Python side through a dict which holds the current state to make things thread-safe. So each time C code sends the control back to Python side we have to pack the struct into a Python dict. But CPython API is a bit fun in the sense that sometimes you have to ref count sometimes you don't. They call it strong and weak(borrowed) reference. So `GetItemString` is weak but `SetItemString` is strong so you need to involve ref counting in the X-macro.

So this is supposed to be a note about that particular detail.

scipy/optimize/__slsqp.h
ilayn: This is actually the input parsing that is coming from the user. So these values are populated by the `PyArg_ParseTuple` function. Then we call the actual low level function after a bit of sanity checks etc. Am I understanding your comment correctly?

scipy/optimize/__slsqp.h
ilayn: `static` in the sense that original code had these `SAVE`d hence they were not thread-safe in C `SAVE` is done via declaring them `static`. Now that I moved them into a dict that's what I meant but might be confusing indeed.

scipy/optimize/meson.build
ilayn: Not really any reason other than visually separating them from the `_XXX.py` files. It looked a bit unstructured when they were all garbled together. Maybe we should throw them all into a SRC dir later.

scipy/optimize/__slsqp.h
rgommers: What I meant is that normally only signatures, includes, header guards go in header files, and the actual code goes in the source files. It's not always consistent in our code base, but if you grep for `PyInit_` for example, the vast majority is in `.c`/`.cc` files and not in the header. Input argument parsing is similar I think, usually in the sources not the header.

`__nnls.h` is quite minimal, and that seems more desirable in principle. 

Anyway, if you want to keep it like this, please don't let me stop you. It just seemed worth pointing out.

scipy/optimize/__slsqp.h
rgommers: So they're not actually declared `static`. Are you saying that the current version is still not thread-safe? I thought that it is thread-safe now. If so, it'd be good to remove `_static` from the name. If not, I think `pytest-run-parallel` should have caught it.

scipy/optimize/meson.build
rgommers: `src/` sounds good to me.

scipy/optimize/__slsqp.h
ilayn: Ah OK, I understand now. So the previous practice was to create a `<name>module.c` to handle all this and indeed the header was just to deal with the declarations. I could definitely have done that way, now that you put it like that. 

Hmm then I made all of them wrong so far. Let me get these in as is, then I'll make a clean up session for all of them.

scipy/optimize/__slsqp.h
ilayn: No they are now thread-safe now that we do the dictionary ping-pong. The name is supposed to be saying `these are the variables that were previously static`. 

But yeah I already confused you meaning I'll confuse others too. Let me rename it.

ilayn:
OK free-thread is coughing again but the rest is OK. 

> The C code isn't the prettiest with goto's etc., but I'm sure you already knew that

Yes I thought about removing gotos (which is only in the actual SLSQP code) but they are not just conditional jump instructions to unroll but they are also the reentry points to the code. So did not want to entangle things a bit more. That part needs a proper rewrite from scratch. As I mentioned above, the unicorn move would be to detach the LSQ part out and put Highs QP solver in.

ilayn:
Thank you all for the reviews and benchmarks, much appreciated.

andyfaff:
Well done @ilayn

rgommers:
Awesome work Ilhan!

Question about the list of issues with (unchecked) check boxes: is the idea that those weren't actually closed by this PR yet, but are now more tractable to work on?

ilayn:
They are more of algorithm issues and not F77 problems. I wanted to do something about them but did not attempt as I don't have a full grasp of what other functions are doing in `scipy.optimize`, but dropped some comments. I think judging by the silence we can close them but I wasn't sure.

PR #22525, opened 2025-02-13
ENH: Improve `cdist` Minkowski performance for `p != 0, 1, 2, np.inf`…

schiffinor:
#### Reference issue
N/A

Optimized the `cdist` computation for Minkowski distance when `p` is not 0, 1, 2, or `np.inf`, improving performance by approximately 3x. 

Changes include:
- Introduced `compute_abs_difference` in `_distance_pybind` using OpenMP for parallel computation.
- - Implemented `arrayPrepper` to offload absolute difference computation to a C++ extension via `compute_abs_difference` (`_distance_pybind`).
- Created `lp_norm_v2_pList` to call `arrayPrepper` and leverage  efficient NumPy operations (einsum).
- Added a shortcut in `_cdist_callable` to call `lp_norm_v2_pList` when applicable.

#### Additional information
First time making a contribution and I don't have too much formal background in CS. I have a degree in math and just really like CS. I apologize if there's any formality or things I missed. This passed all my tests though. I can also provide an external tester file I made if necessary/desired.

### DIFF
--- a/scipy/spatial/src/distance_pybind.cpp
+++ b/scipy/spatial/src/distance_pybind.cpp
@@ -547,6 +547,37 @@ py::array cdist(const py::object& out_obj, const py::object& x_obj,
     return out;
 }
 
+py::array_t<double> compute_abs_difference(const py::array_t<double>& a, const py::array_t<double>& b) {
+    const auto buf_a = a.request(), buf_b = b.request();
+    if (buf_a.ndim != 2 || buf_b.ndim != 2) {
+        throw std::runtime_error("Input should be 2-D NumPy arrays");
+    }
+    if (buf_a.shape[1] != buf_b.shape[1]) {
+        throw std::runtime_error("The inner dimensions must match");
+    }
+
+    Py_ssize_t rows_a = buf_a.shape[0], cols_a = buf_a.shape[1];
+    Py_ssize_t rows_b = buf_b.shape[0];
+
+    auto result = py::array_t<double>({rows_a, rows_b, cols_a});
+    const auto buf_result = result.request();
+
+    const auto *a_ptr = static_cast<double *>(buf_a.ptr);
+    const auto *b_ptr = static_cast<double *>(buf_b.ptr);
+    const auto result_ptr = static_cast<double *>(buf_result.ptr);
+
+#pragma omp parallel for
+    for (Py_ssize_t i = 0; i < rows_a; ++i) {
+        for (Py_ssize_t j = 0; j < rows_b; ++j) {
+            for (Py_ssize_t k = 0; k < cols_a; ++k) {
+                result_ptr[i * rows_b * cols_a + j * cols_a + k] = std::abs(a_ptr[i * cols_a + k] - b_ptr[j * cols_a + k]);
+            }
+        }
+    }
+
+    return result;
+}
+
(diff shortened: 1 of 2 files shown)


scipy/spatial/src/distance_pybind.cpp
```diff
@@ -547,6 +547,37 @@ py::array cdist(const py::object& out_obj, const py::object& x_obj,
...
+    Py_ssize_t rows_b = buf_b.shape[0];
+
+    auto result = py::array_t<double>({rows_a, rows_b, cols_a});
+    const auto buf_result = result.request();
+
+    const auto *a_ptr = static_cast<double *>(buf_a.ptr);
+    const auto *b_ptr = static_cast<double *>(buf_b.ptr);
+    const auto result_ptr = static_cast<double *>(buf_result.ptr);
+
+#pragma omp parallel for
```
j-bowhay, line 569: Other more knowledgeable will confirm but SciPy doesn't use openmp

schiffinor: Oh whoopsies it should do just fine without it then, a holdover from my initial local implementation.

tylerjereddy (commented):
I added some initial review comments. Is there a citation for this approach or just some improvement that was intuitively obvious to you?

At the moment, on x86_64 Linux, I don't see evidence of a performance improvement using our standard `asv` benchmark suite though (on a branch where I purged out your OMP directive):

`asv continuous -e -b "Xdist" main d48616698bc`

`BENCHMARKS NOT SIGNIFICANTLY CHANGED.`

Those benchmarks check `minkowski` with `p=3`, so if the performance improvement is around 3x for any and all data sizes/inputs, it should be showing up there. If the performance changes are actually more nuanced, it may be helpful to clarify and specify which `p` value we should be additionally checking to reflect the ~3x speedup.

schiffinor:
Yeah no citation just something that was intuitive for me. I was taking a data analysis class and I came up with this method.

schiffinor:
As for the benchmarking I'll now test it without the OMP and as for the docstring style I'll adjust that as well as the naming

schiffinor:
I just checked it without the OMP directive and I'm still getting the same speed increases. I should specify that my code only runs if there's no weight array and if minkowski is specifically called. Again, if desired I can upload my test file somewhere. But to be precise I'm using mp.random.uniform to generate random arrays of size 5,000 and 3,000 points of dimension 1000, each coordinate between -1 million and 1 million and I tested p-values 3, 0.25, 4.5, and π.

Edit:
I just discovered the problem. I was accidentally running the wrong file. I named my test files numerically and made a typo. It would seem that although, my custom implementation works considerably faster standalone. It's the same speed when run through c-dist. I assume it's because of the data validation step or something. I'll look into this more. 

But, for reference my initial implementation (here):
https://github.com/schiffinor/dataVisualization/blob/afb10d54d57d00d7cb155212edbb62c46ac2be88/Project7/analysis.py#L222

https://github.com/schiffinor/dataVisualization/blob/main/Project7/numpyProj/numpyProj/numpyProj.cpp

Ran three times faster. However, it would seem it doesn't when built into scipy.

tylerjereddy:
> However, it would seem it doesn't when built into scipy.

Right, that's the purpose of the `asv` benchmarks. I'm not trying to discourage you though, it is good to see activity around performance in `spatial` internals, let us know if we can clarify anything.

Issue #18583, opened 2023-05-30
DEP: replace `scipy.interpolate.interp2d` by a stub


h-vetinari:
Follow-up to #17180 & #17363

The intention here is not (yet) full removal. As suggested on the [mailing list](https://mail.python.org/archives/list/scipy-dev@python.org/thread/LU72CAGYFH37PZWCANMUF6UWFNSPZRHU), the plan is:
- [x] in release 1.10, `interp2d` starts emitting DeprecationWarnings;
- [x] in release 1.14, `interp2d` implementation is replaced by a stub which gives a hard error with an explanation and a pointer to a porting guide;
 - [ ] in release N+X (X ≫ 2), the stub is finally gone, and an attempt to `from scipy.interpolate import interp2d` gives a usual ImportError. This last point may even turn out to be a SciPy 2.0 event.

h-vetinari:
@j-bowhay, if you change the plan or the milestone, could you leave a comment please?

I found the [why](https://github.com/scipy/scipy/pull/18022#issuecomment-1439615471) through #18618, and I don't mind. It would just be helpful to not have to find out by accident and then have to hunt for context. ;-)

j-bowhay:
Milestone bumped to 1.13 for the second task in the todo list due to a doc issue that made the transition guide difficult to access

ev-br:
If 1.13 is out-of order, we should hold on to this to 1.14

j-bowhay:
Removing milestone and bumped removal to 1.14 due to out-of-order release

Issue #15765, opened 2022-03-11
META: deprecation tracker


h-vetinari:
### _Please edit this tracker as necessary_

Apologies for bombing the issue tracker with lots of outstanding deprecations, but it felt like some necessary housekeeping. 🙃 

Below a list of deprecations by kind and scipy release. Ideally - in my opinion at least (and this is what I propose to use this issue for) - all _**deprecations should come with an announced behaviour change, and with an announced timeframe**_ (i.e. scipy version where things will be changed/removed), and other items migrated there as appropriate.

As the list below shows, this is a trend that started only relatively recently, with removal/behaviour change in `1.(N+2)` if a given deprecation is introduced in `1.N`. Still, I think that's a reasonable policy.

### Deprecations in documentation only

Need to add proper deprecation warnings. This list is likely incomplete

* 0.10
  * [ ] #15748
* 0.19
  * [ ] #15735

### Deprecated without announced removal

Sharpen warning to announce removal in 1.10/1.11, or remove directly in 1.9, depending.

* 0.13
  * [ ] #15764
* 0.15
  * [ ] #15733
* 0.16 
  * [ ] #15755
* 0.17 
  * [ ] #15758
* 0.18
  * [ ] #15751
* 0.19
  * [ ] #15734
  * [ ] #15736
  * [ ] #15752
  * [ ] #15754
  * [ ] #15759
* 1.0
  * [ ] #15741
  * [ ] #15742
  * [ ] #15753
  * [ ] #15760
* 1.1
  * [ ] #15732
  * [ ] #15738
  * [ ] #15745
  * [ ] #15761
* 1.3 
  * [ ] #15757
* 1.4 
  * [ ] #14889
  * [ ] #15743
* 1.5
  * [ ] #15747
  * [ ] #15762
* 1.7
  * [ ] #15737
* 1.8
  * [ ] #15744
  * [ ] #15749
* 1.9
  * [ ] #15739

### Removal announced, missed

* 1.5 (announced removal for 1.7)
  * [ ] #15731
* 1.6 (announced removal for 1.8)
  * [ ] #15746
  * [ ] #15750

### Removal announced in `1.N` for `1.(N+2)`

* 1.7 (announced removal for 1.9)
  * [ ] #15740
* 1.9 (announced removal for 1.11)
  * [ ] #15722
  * [ ] #15707
  * [ ] #15706
  * [ ] #15756

### Outstanding deprecations

List most likely incomplete

* [ ] #5676
[...]

PR #22526, opened 2025-02-13
DOC: interpolate: reorganize the API listing

ev-br:
Restructure for clarity, tone down legacy routines.

[docs only]

### DIFF
--- a/scipy/interpolate/__init__.py
+++ b/scipy/interpolate/__init__.py
@@ -49,7 +49,6 @@ Multivariate interpolation
    CloughTocher2DInterpolator
    griddata - a convenience wrapper for the above three interpolators
    RBFInterpolator
-   interp2d
 
 **For data on a grid:**
 
@@ -57,7 +56,6 @@ Multivariate interpolation
    :toctree: generated/
 
    RegularGridInterpolator
-   interpn - a convenience wrapper for `RegularGridInterpolator`
 
 .. seealso::
 
@@ -95,7 +93,6 @@ Rational Approximation
    :toctree: generated/
 
    AAA
-   pade
 
 
 Interfaces to FITPACK routines for 1D and 2D spline fitting
@@ -177,12 +174,15 @@ Additional tools
 
    lagrange
    approximate_taylor_polynomial
+   pade
 
+   interpn - a convenience wrapper for `RegularGridInterpolator`
    barycentric_interpolate - A convenience wrapper for `BarycentricInterpolator`
    krogh_interpolate - A convenience wrapper for `KroghInterpolator`
    pchip_interpolate - A convenience wrapper for `PchipInterpolator`
-   interp1d - A legacy interface to 1D interpolation (not recommended for new code)
    Rbf - a legacy RBF interpolator class (prefer ``RBFInterpolator`` in new code)
+   interp1d - A legacy interface to 1D interpolation (not recommended for new code)
+   interp2d - Removed in version 1.14. See the :ref:`replacement guide <interp-transition-guide>`
 
 
 .. seealso::
@@ -190,7 +190,7 @@ Additional tools
    `scipy.ndimage.map_coordinates`,
    `scipy.ndimage.spline_filter`,
 
-"""
+"""  # noqa: E501
 from ._interpolate import *
 from ._fitpack_py import *
 
--- a/scipy/interpolate/__init__.py
+++ b/scipy/interpolate/__init__.py
@@ -5,13 +5,11 @@ Interpolation (:mod:`scipy.interpolate`)
 
 .. currentmodule:: scipy.interpolate
 
-Sub-package for objects used in interpolation.
+Sub-package for functions and objects used in interpolation.
 
-As listed below, this sub-package contains spline functions and classes,
-1-D and multidimensional (univariate and multivariate)
-interpolation classes, Lagrange and Taylor polynomial interpolators, and
-wrappers for `FITPACK <http://www.netlib.org/dierckx/>`__
-and DFITPACK functions.
+See the :ref:`user guide <tutorial-interpolate>` for recommendations on choosing a
+routine, and other usage details.
+
 
 Univariate interpolation
 ========================
@@ -19,51 +17,56 @@ Univariate interpolation
 .. autosummary::
    :toctree: generated/
 
-   interp1d
-   BarycentricInterpolator
-   KroghInterpolator
-   barycentric_interpolate
-   krogh_interpolate
-   pchip_interpolate
-   CubicHermiteSpline
+   make_interp_spline
+   CubicSpline
    PchipInterpolator
    Akima1DInterpolator
-   CubicSpline
+   FloaterHormannInterpolator
+   BarycentricInterpolator
+   KroghInterpolator
+   CubicHermiteSpline
+
+**Low-level data structures for univariate interpolation:**
+
+.. autosummary::
+   :toctree: generated/
+
    PPoly
    BPoly
-   FloaterHormannInterpolator
+   BSpline
 
 
 Multivariate interpolation
 ==========================
 
-Unstructured data:
+**Unstructured data**
 
 .. autosummary::
    :toctree: generated/
 
-   griddata
    LinearNDInterpolator
    NearestNDInterpolator
    CloughTocher2DInterpolator
+   griddata - a convenience wrapper for the above three interpolators
    RBFInterpolator
-   Rbf
    interp2d
 
-For data on a grid:
+**For data on a grid:**
 
 .. autosummary::
    :toctree: generated/
 
-   interpn
    RegularGridInterpolator
-   RectBivariateSpline
+   interpn - a convenience wrapper for `RegularGridInterpolator`
 
 .. seealso::
 
-    `scipy.ndimage.map_coordinates`
+    `scipy.ndimage.map_coordinates`,
+    :ref:`An example wrapper for map_coordinates <tutorial-interpolate_cartesian-grids>`
 
-Tensor product polynomials:
+
+**Low-level data structures for tensor product polynomials and splines:**
+
 
 .. autosummary::
    :toctree: generated/
@@ -71,21 +74,46 @@ Tensor product polynomials:
    NdPPoly
    NdBSpline
 
-1-D Splines
-===========
+
+1-D spline smoothing and approximation
+======================================
 
 .. autosummary::
    :toctree: generated/
 
-   BSpline
-   make_interp_spline
    make_lsq_spline
    make_smoothing_spline
    generate_knots
    make_splrep
    make_splprep
 
-Functional interface to FITPACK routines:
+
+Rational Approximation
+======================
+
+.. autosummary::
+   :toctree: generated/
+
+   AAA
+   pade
+
+
+Interfaces to FITPACK routines for 1D and 2D spline fitting
+===========================================================
+
+This section lists wrappers for `FITPACK <http://www.netlib.org/dierckx/>`__
+functionality for 1D and 2D smoothing splines. In most cases, users are better off
+using higher-level routines listed in previous sections.
+
+
+1D FITPACK splines
+------------------
+
+This package provides two sets of functionally equivalent wrappers: object-oriented and
+functional.
+
+**Functional FITPACK interface:**
+
 
 .. autosummary::
    :toctree: generated/
@@ -100,7 +128,7 @@ Functional interface to FITPACK routines:
    splantider
    insert
 
-Object-oriented FITPACK interface:
+**Object-oriented FITPACK interface:**
 
 .. autosummary::
    :toctree: generated/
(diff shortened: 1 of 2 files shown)


scipy/interpolate/__init__.py
```diff
@@ -5,87 +5,117 @@
...
+------------------------------------------------------
+
+.. autosummary::
+   :toctree: generated/
+
    PPoly
    BPoly
-   FloaterHormannInterpolator
+   BSpline
+   CubicHermiteSpline
```
mdhaber, line 37: Can we put this at the top? Although this happens to be the base class of some of the others, it's pretty thin on top of `PPoly` and quite usable on its own. I'd say the criterion for "low-level" is that these others don't accept `x`, `y`.

scipy/interpolate/__init__.py
```diff
@@ -5,87 +5,115 @@
...
    :toctree: generated/
 
-   griddata
    LinearNDInterpolator
    NearestNDInterpolator
    CloughTocher2DInterpolator
+   griddata - a convenience wrapper for the above three interpolators
    RBFInterpolator
-   Rbf
    interp2d
```
mdhaber, line 52: The docs suggest that this is removed?
https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interp2d.html#scipy.interpolate.interp2d

scipy/interpolate/__init__.py
ev-br: Yup, this can go either way. Done and force-pushed (+a couple of fix-ups for the layout and references).

scipy/interpolate/__init__.py
ev-br: Not quite. It follows a special-cased staged removal procedure: https://github.com/scipy/scipy/issues/18583
AFAIR it used to be listed in the deprecation tracker https://github.com/scipy/scipy/issues/15765, cannot find it in it anymore though.

scipy/interpolate/__init__.py
```diff
@@ -5,87 +5,115 @@
...
-For data on a grid:
+**For data on a grid:**
 
 .. autosummary::
    :toctree: generated/
 
-   interpn
    RegularGridInterpolator
-   RectBivariateSpline
+   interpn - a convenience wrapper for `RegularGridInterpolator`
```
mdhaber, line 60: We seem to be going away from convenience wrappers. How about moving this to additional tools? 

Also, I'm not sure if the stuff after the name of the function renders. I think that will be replaced with the one line summary, so if that's the case (let's check), let's leave it out.

scipy/interpolate/__init__.py
```diff
@@ -5,87 +5,115 @@
...
-Functional interface to FITPACK routines:
+
+Rational Approximation
+======================
+
+.. autosummary::
+   :toctree: generated/
+
+   AAA
+   pade
```
mdhaber, line 98: Should we de-emphasize this by putting it in additional tools? I think it would belong up here if it had an interface something like the others. If the method has advantages over AAA, it might be worth creating a `Pade` with a similar interface and moving `pade` to the bottom. If not, even more reason to just move `pade` to the bottom.

j-bowhay: The current implementation of `pade` is flawed so no objections to moving to the bottom. I have a rewrite on my laptop back home that I am planning to contribute but I haven't decided whether to write a new interface or just make it fit the old one . 

`pade` and `AAA` serve different purposes: `AAA` is a bit like fitting a polynomial to data but for rational functions and `pade` is like a rational Taylor series

ev-br: +1 for moving `pade` to additional tools, done. It's going to be an odd one out no matter what, so it's a right place for it indeed. Done.

scipy/interpolate/__init__.py
ev-br: Ok, additional tools it is then, in the last commit.

scipy/interpolate/__init__.py
ev-br: Huh. It definitely used to render, but does not seem to anymore. Which part of the doc infra is responsible, I've no idea.

mdhaber:
Thanks. One more thing for here or a separate PR that would really make this nice: the one line summaries could do a better job of distinguishing between the methods. Personally, I'm only really interested in the things above "Interfaces to FITPACK...". I would submit a PR to your branch, but I don't have enough background, so I thought I'd get some feedback first.

---

In the the univariate interpolation section, I think we should decide what information (besides the name of the method) should be included in the description. 

| Feature | Description |
| -------- | ------- |
| `make_interp_spline` | Compute the (coefficients of) interpolating B-spline.| 
| `CubicSpline` | Cubic spline data interpolator.| 
| `PchipInterpolator` | PCHIP 1-D monotonic cubic interpolation.| 
| `Akima1DInterpolator` | Akima interpolator.| 
| `FloaterHormannInterpolator` | Floater-Hormann barycentric rational interpolator.| 
| `BarycentricInterpolator` | Interpolating polynomial for a set of points.| 
| `KroghInterpolator` | Interpolating polynomial for a set of points.| 
| `CubicHermiteSpline` | Piecewise-cubic interpolator matching values and first derivatives.| 

IIUC, the words "spline", "interpolating polynomial", and "piecewise interpolator" found in these descriptions does not help the user distinguish between them. All of these fit "a set of points" or "data" (although `CubicHermiteSpline` is unique in that it fits first derivatives). All of these (except `make_interp_spline` are classes, so if we are to pick between "interpolation" and "interpolator", then "interpolator" seems more appropriate. With that in mind:

| Feature | Description |
| -------- | ------- |
| `make_interp_spline` | Create an interpolating B-spline of given degree and boundary conditions. | 
| `CubicSpline` | Piecewise cubic interpolator.| 
| `PchipInterpolator` | PCHIP (Piecewise Cubic Hermite Interpolating Polynomial) interpolator.| 
| `Akima1DInterpolator` | Akima interpolator.| 
| `FloaterHormannInterpolator` | Floater-Hormann barycentric rational interpolator.| 
| `BarycentricInterpolator` | Barycentric interpolator.| 
| `KroghInterpolator` | Krogh interpolator.| 
| `CubicHermiteSpline` | Piecewise-cubic interpolator matching values and first derivatives.| 

Can we say more about the Akima, barycentric, and Krogh interpolators to distinguish among them?

---


| Feature | Description |
| -------- | ------- |
| `PPoly` | Piecewise polynomial in terms of coefficients and breakpoints
| `BPoly` | Piecewise polynomial in terms of coefficients and breakpoints.
| `BSpline` | Univariate spline in the B-spline basis.

I don't remember from the tutorial - is there something fundamentally different from these besides the basis? That is, could all of these be described as a "univariate spline"? Can we align the descriptions accordingly? I don't know if it's accurate, but this is what it would look like.

| Feature | Description |
| -------- | ------- |
| `PPoly` | Univariate spline in the power basis.
| `BPoly` | Univariate spline in the Bernstein basis.
| `BSpline` | Univariate spline in the B-spline basis.

---

Likewise for multivariate interpolation:
image


`NearestNDInterpolator` needs a description ("Nearest neighbor interpolation"?), and `griddata` presumably is able to "Interpolate unstructed D-D data" given that it's in the "multivariate interpolation, unstructured data" section. Is there an algorithm name we can mention, and should we use $N$ throughout to represent an arbitrary integer?

Please edit feel free to edit my post with changes, and I can make a PR to this branch from it.

ev-br:
Thanks a bunch, it's great input!
I'll offer just a handful of small suggestions, most of which are best decided on by a reader who did not spend time and time with all this, so you :-).

`Univariate interpolation` section: +1 to your wordings. 
Possible tweaks (up to you): 
- PCHIP and Akima are both C1 smooth, and _monotone interpolation_ is a term AFAIK. May or may not want to include both bits of info. Strictly speaking, Akima1D is not guaranteed to be strictly monotonic (while PCHIP is, by construction), but it is still typically classed as a kind of a monotone interpolator.
- "Floater-Hormann barycentric interpolator" I don't think we need to elaborate IMO. This wording is enough to google it up (which I did when I first saw the PR introducing them). Unless @j-bowhay has a better suggestion of course.
- KroghInterpolator : I don't know if we want to say much about it. First half dozen google hits are to scipy anyway, followed by links to an original paper which introduced this particular scheme of organizing computations. So the literature here basically follows what an original author implemented, a decade and a half ago. IIUC one feature of KroghInterpolator is that new points can be added in O(1) time; the price is that it's incredibly slow to evaluate.  From where I stand, it can even move to "Additional tools". Possibly together with the BarycentricInterpolator (the decision is up to Matt and Jake; I'm on a fence, and agree with whatever you two agree on). 


`Low-level data structures` section: TL;DR: let's keep the wording as is.

- No, please don't call PPoly and BPoly splines. The terminology is messy enough (it's the literature, not us), so let's keep calling `PPoly` and `BPoly` _piecewise polynomials_. Typically "spline" implies some continuity, and these two can represent discontinuous polynomials. `BSpline`, OTOH, is C^k. So the current wording might be confusing, but at least it's precise and googlable.

`Multivariate interpolation` section:

- +1 to your edits, with a small tweak: RBF, as implemented, works in `N >= 1` dimensions.

--------------------

Re: convenience functions:

- `interpn` and `griddata` I'd keep close to the classes they wrap. The reasoning is 1) the naming is confusing. E.g. `griddata` is not for data on the grid! 2) they are actually used in the wild, so probably we cannot deprecate them inasmuch I would personally support just removing them. (I'm on record for disliking this breed of convenience functions; the dislike is not strong enough to justify the churn IMO; call it a Stockholm syndrome if you will).
So the suggestion is to keep them, keep them close to the classes they wrap, and forget about them.  

- `krogh_interpolate`, `pchip_interpolate` etc : I don't think these are used much or at all, so I think we can remove them relatively easily. That'd be some churn in downstream libraries (not end users!), so my mode of operation so far was to carry on as if they did not exist.

mdhaber:
"monotone interpolation" I've seen that, and considered keeping it, but to the unfamiliar, it sounds like it is limited to *monotonic* functions. That is not what it means, and I thought "shape preserving" is better for the one-line description. The rest can be explained in the detailed summary, etc.

> No, please don't call PPoly and BPoly splines.

Would it be fair to say that "*all* splines are piecewise polynomials" (where *all* really means *typically but with occasional exceptions in terminology*) but "not all piecewise polynomials are splines"? That sort of relationship?

> `interpn` and `griddata` I'd keep close to the classes they wrap.

I take the part you wrote about "so probably we cannot deprecate them inasmuch I would personally support just removing them" as support for _de-emphasizing_ them : ) I'm not talking about _deprecating_ them, just moving where they appear in the top-level documentation. If we don't *recommend* their use (and actually would support removing them if it were not for the painful churn), I say keep the documentation simpler for new users, and *existing* users can still find what they need wherever they appear. Location in the documentation really can't be expected to cover for entirely misleading function names...

@j-bowhay I'd appreciate your thoughts  on this and the rest.

j-bowhay:
Ok a couple of thoughts:
The summary for `BarycentricInterpolator` should make the link with Lagrange interpolation since this is probably the phase you learn about in any intro to numerical analysis class. From the [original paper](https://people.maths.ox.ac.uk/trefethen/barycentric.pdf) 
> Barycentric interpolation is a variant of Lagrange polynomial interpolation that is fast and
 stable. It deserves to be known as the standard method of polynomial interpolation.

So how about a summary along the lines of `Fast and stable variant of Lagrange interpolation `. I would be -1 on moving out of the univariate interpolation section, it's a really good method if your function is sufficiently nice and your data is not evenly spaced.

For `FloaterHormannInterpolator` I'd probably stick with what you have in the table.

I'm still making my mind up about moving `interpn` and `griddata`, both arguments are compelling!

mdhaber:
OK. Can we say anything about the continuity of barycentric, FH, and Krogh?

j-bowhay:
Barycentric interpolation is polynomial interpolation so is $\mathcal{C}^\infty$. FH is $\mathcal{C}^\infty(\mathbb{R})$ although I'd want to triple check that

ev-br:
> So how about a summary along the lines of Fast and stable variant of Lagrange interpolation

Stable yes, fast please no. I know what you mean and do not dispute this in principle---but the actual implementation of the BarycentricInterpolator is not fast compared to anything PPoly based. And since in the API listing we only have one line, it's not a place to mix in implementation details.

mdhaber:
OK, I've updated the entry for `BarycentricInterpolator`

> Barycentric (stable Lagrange) interpolator (C∞ smooth).

mdhaber:
Wait - this doesn't necessary affect the continuity, though, right? `CubicHermiteSpline` is piecewise cubic, so you can only match the values and first derivatives between pieces. But this Krogh thing could still be a _single_ (C∞) polynomial, and it just increases the order of the single polynomial to fit derivatives as well as values. No?

ev-br:
> No?

Yes :-).
My comment is more about utility, not continuity.

mdhaber:
OK, I'll merge this one and open a separate PR with the rest.

PR #22527, opened 2025-02-13
DOC: sparse: add returns sections to some `_construct.py` functions

### LINKED ISSUE #22065: DOC: sparse: Several functions are missing the 'Returns' section of their docstring

The docstrings of the following functions in `scipy.sparse` are missing the 'Returns' section:

* [`diags`](https://scipy.github.io/devdocs/reference/generated/scipy.sparse.diags.html)
* [`diags_array`](https://scipy.github.io/devdocs/reference/generated/scipy.sparse.diags_array.html)
* [`eye`](https://scipy.github.io/devdocs/reference/generated/scipy.sparse.eye.html)
* [`eye_array`](https://scipy.github.io/devdocs/reference/generated/scipy.sparse.eye_array.html)
* [`identity`](https://scipy.github.io/devdocs/reference/generated/scipy.sparse.identity.html)
* [`save_npz`](https://scipy.github.io/devdocs/reference/generated/scipy.sparse.save_npz.html)
* [`spdiags`](https://scipy.github.io/devdocs/reference/generated/scipy.sparse.spdiags.html)

Nothing urgent here--what each function returns is pretty obvious in all cases--but it would be nice to fix these to conform to the docstring standard.

`save_npz` is pretty clearly a "side effects only" function and returns `None`, so it doesn't necessarily need a 'Returns' section.

dschult:
This adds `Returns` sections to the docs as noted in #22065.  The discussion there branched into how to monitor and gradually implement `Returns` sections in all functions across SciPy. But this PR sticks to the original post in #22065 which noted that a number of functions in `sparse._construct.py` don't have `Returns` sections.

I did not add that section for `save_npz` because it returns None (doesn't have a return statement).

I also cleaned up some backticks and changed a few words:
- added See Also section to `identity()`
- changed the first line of `eye` and `eye_array` to remove the word Identity and also state that the diagonal of ones could be controlled.  Now reads: `Sparse array with ones on (and zeros off) the specified diagonal`

I'm not sure that this closes #22065 because of the larger discussion there. But maybe @WarrenWeckesser can make a judgement for whether to close that when this is merged.

### DIFF
--- a/scipy/sparse/_construct.py
+++ b/scipy/sparse/_construct.py
@@ -28,6 +28,12 @@ def spdiags(data, diags, m=None, n=None, format=None):
     """
     Return a sparse matrix from diagonals.
 
+    .. warning::
+
+        This function returns a sparse matrix -- not a sparse array.
+        You are encouraged to use ``dia_array`` to take advantage
+        of the sparse array functionality.
+
     Parameters
     ----------
     data : array_like
@@ -50,12 +56,6 @@ def spdiags(data, diags, m=None, n=None, format=None):
     -------
     new_matrix : sparse matrix
         dia_matrix format with values in ``data`` on diagonals from ``diags``.
-
-    .. warning::
-
-        This function returns a sparse matrix -- not a sparse array.
-        You are encouraged to use ``dia_array`` to take advantage
-        of the sparse array functionality.
 
     Notes
     -----
@@ -137,6 +137,10 @@ def diags_array(diagonals, /, *, offsets=0, shape=None, format=None, dtype=None)
     Each value in the input ``diagonals`` is used.
 
     .. versionadded:: 1.11
+
+    See Also
+    --------
+    dia_array : constructor for the sparse DIAgonal format.
 
     Examples
     --------
@@ -258,11 +262,6 @@ def diags(diagonals, offsets=0, shape=None, format=None, dtype=None):
         dia_matrix holding the values in `diagonals` offset from the main diagonal
         as indicated in `offsets`.
 
-    See Also
-    --------
-    spdiags : construct matrix from diagonals
-    diags_array : construct sparse array instead of sparse matrix
-
     Notes
     -----
     Repeated diagonal offsets are disallowed.
@@ -280,6 +279,11 @@ def diags(diagonals, offsets=0, shape=None, format=None, dtype=None):
     Each value in the input ``diagonals`` is used.
 
     .. versionadded:: 0.11
+
+    See Also
+    --------
+    spdiags : construct matrix from diagonals
+    diags_array : construct sparse array instead of sparse matrix
 
     Examples
     --------
@@ -452,6 +456,12 @@ def eye(m, n=None, k=0, dtype=float, format=None):
     Returns a sparse matrix (m x n) where the kth diagonal
     is all ones and everything else is zeros.
 
+    .. warning::
+
+        This function returns a sparse matrix -- not a sparse array.
+        You are encouraged to use ``eye_array`` to take advantage
+        of the sparse array functionality.
+
     Parameters
     ----------
     m : int
@@ -465,16 +475,14 @@ def eye(m, n=None, k=0, dtype=float, format=None):
     format : str, optional
         Sparse format of the result, e.g., format="csr", etc.
 
-    .. warning::
-
-        This function returns a sparse matrix -- not a sparse array.
-        You are encouraged to use ``eye_array`` to take advantage
-        of the sparse array functionality.
-
     Returns
     -------
     new_matrix : sparse matrix
         Sparse matrix of chosen shape with ones on the kth diagonaland zeros elsewhere.
+
+    See Also
+    --------
+    eye_array : Sparse array of chosen shape with ones on a specified diagonal.
 
     Examples
     --------
--- a/scipy/sparse/_construct.py
+++ b/scipy/sparse/_construct.py
@@ -124,13 +124,13 @@ def diags_array(diagonals, /, *, offsets=0, shape=None, format=None, dtype=None)
     -----
     Repeated diagonal offsets are disallowed.
 
-    The result from `diags_array` is the sparse equivalent of::
+    The result from ``diags_array`` is the sparse equivalent of::
 
         np.diag(diagonals[0], offsets[0])
         + ...
         + np.diag(diagonals[k], offsets[k])
 
-    ``diags_array`` differs from ``dia_array`` in the way it handles off-diagonals.
+    ``diags_array`` differs from `dia_array` in the way it handles off-diagonals.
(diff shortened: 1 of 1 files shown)


scipy/sparse/_construct.py
```diff
@@ -241,6 +252,12 @@ def diags(diagonals, offsets=0, shape=None, format=None, dtype=None):
     dtype : dtype, optional
         Data type of the matrix.
 
+    Returns
+    -------
+    new_matrix : dia_matrix
+        dia_matrix holding the values in `diagonals` offset from the main diagonal
```
perimosocordiae, line 258: Double backticks here and the next line

scipy/sparse/_construct.py
```diff
@@ -301,16 +318,16 @@ def diags(diagonals, offsets=0, shape=None, format=None, dtype=None):
...
     that it has a square shape with ones only on the main diagonal.
-    It is thus the multiplicative identity. `eye_array` allows
+    It is thus the multiplicative identity. ``eye_array`` allows
     rectangular shapes and the diagonal can be offset from the main one.
 
     .. warning::
 
         This function returns a sparse matrix -- not a sparse array.
-        You are encouraged to use ``eye_array`` to take advantage
+        You are encouraged to use `eye_array` to take advantage
```
perimosocordiae, line 330: Double backticks here?

scipy/sparse/_construct.py
```diff
@@ -322,6 +339,16 @@ def identity(n, dtype='d', format=None):
     format : str, optional
         Sparse format of the result, e.g., format="csr", etc.
 
+    Returns
+    -------
+    new_matrix : sparse matrix
+        A square sparse matrix with ones on (and zeros off) the main diagonal.
```
perimosocordiae, line 345: A square sparse matrix with ones on the main diagonal, and zeroes elsewhere.

scipy/sparse/_construct.py
```diff
@@ -322,6 +339,16 @@ def identity(n, dtype='d', format=None):
...
+
+    See Also
+    --------
+    eye_array : Sparse array with ones on (and zeros off) the specified diagonal
+    eye : Sparse matrix with ones on (and zeros off) the specified diagonal
```
perimosocordiae, line 350: I think these would be more useful if we mentioned how they differ from the main function that we're documenting. For example, the see also section for identity() could mention that the eye() function produces ones on a specified diagonal, not just the main diagonal.

perimosocordiae (approved):

scipy/sparse/_construct.py
j-bowhay: I think single backticks here are correct since these are arguments to the function

dschult: Yes, it looks like single ticks should be used for arguments so they get italics. Thanks!

scipy/sparse/_construct.py
```diff
@@ -439,6 +471,11 @@ def eye(m, n=None, k=0, dtype=float, format=None):
         You are encouraged to use ``eye_array`` to take advantage
         of the sparse array functionality.
 
+    Returns
```
perimosocordiae, line 474: Let's move this up before the warning section

dschult: I've moved the warning sections, rechecked the single/double backticks and tried to expend those backtick suggestions to the rest of this module. 
@perimosocordiae this should be ready for a re-review. :)

dschult:
@WarrenWeckesser does this address your initial concern with #22065?
Is there more to do here (without going repo-wide)?

scipy/sparse/_construct.py
```diff
@@ -46,15 +52,14 @@ def spdiags(data, diags, m=None, n=None, format=None):
...
 
-    .. warning::
-
-        This function returns a sparse matrix -- not a sparse array.
-        You are encouraged to use ``dia_array`` to take advantage
-        of the sparse array functionality.
+    Returns
+    -------
+    new_matrix : sparse matrix
+        dia_matrix format with values in ``data`` on diagonals from ``diags``.
```
j-bowhay, line 58: Single back ticks since one is a class and the other are arguments
```suggestion
        `dia_matrix` format with values in `data` on diagonals from `diags`.
```

scipy/sparse/_construct.py
```diff
@@ -109,11 +114,17 @@ def diags_array(diagonals, /, *, offsets=0, shape=None, format=None, dtype=None)
     dtype : dtype, optional
         Data type of the array.
 
+    Returns
+    -------
+    new_array : dia_array
+        dia_array holding the values in `diagonals` offset from the main diagonal
```
j-bowhay, line 120:
```suggestion
        `dia_array` holding the values in `diagonals` offset from the main diagonal
```

scipy/sparse/_construct.py
```diff
@@ -241,29 +256,35 @@ def diags(diagonals, offsets=0, shape=None, format=None, dtype=None):
...
         Data type of the matrix.
 
-    See Also
-    --------
-    spdiags : construct matrix from diagonals
-    diags_array : construct sparse array instead of sparse matrix
+    Returns
+    -------
+    new_matrix : dia_matrix
+        dia_matrix holding the values in `diagonals` offset from the main diagonal
```
j-bowhay, line 262:
```suggestion
        `dia_matrix` holding the values in `diagonals` offset from the main diagonal
```

scipy/sparse/_construct.py
mdhaber: Rules for backticks are in the [numpydoc style guide](https://numpydoc.readthedocs.io/en/latest/format.html#common-rest-concepts).

scipy/sparse/_construct.py
mdhaber: Single should be correct, since `eye_array` is a function. Something to check is whether the link to the function documentation renders.

scipy/sparse/_construct.py
j-bowhay: It's not immediately obvious to me why this isn't working since `data` is an argument...

dschult: The numpydoc stype guide only says that single backticks is preferred because numpydocs will add a feature for it. But in the meantime it may give warnings.  So that doesn't really help us figure out when to use backticks.

In this case, there are two things apparently going on: both `data` and  `diags` are args, not kwargs, so no link is found. At least if I make the function signature `data=None, diags=None` it doesn't have trouble finding a link to the params.  Could it be that links are only found to parameters that are keywords?

Secondly, `diags` is a function in this module. So the docs link to that function rather than to the parameter.

So, I have reverted to using double backticks for both `data` and `diags`. I also put double backticks around a piece of code a few lines up since I noticed that need.

Hoping for some green checks! :)

dschult:
The failing tests look unrelated.

Issue #16191, opened 2022-05-16
SciPy for Biomedicine


mdhaber:
This is a meta-issue recording action items from a review of biomedical articles that cite SciPy and the results of a survey sent to authors of over 50 of these papers.

- [ ] [Vaccine optimization for COVID-19: Who to vaccinate first?](https://www.science.org/doi/full/10.1126/sciadv.abf1374) used `optimize.minimize` to study how COVID vaccination priority should be assigned. The researchers reported having to “repair” solutions returned by SciPy’s Nelder-Mead solver, but had the researchers used a solver capable of enforcing problem constraints, this might have been avoided. To resolve this, we will improve the documentation of the optimizers to ensure that researchers find the appropriate solver for their problem, and we will enhance the solvers’ support for constraints.
- [ ] [Modeling and forecasting the early evolution of the Covid-19 pandemic in Brazil](https://www.nature.com/articles/s41598-020-76257-1) used the `solve_ivp` method to numerically integrate a system of ordinary differential equations. Leveraging Boost, the C++ library added to SciPy as part of EOSS-0000000432, we will add more modern integrators to `solve_ivp` to improve `solve_ivp`’s performance.
- [ ] [Multi-Omics Resolves a Sharp Disease-State Shift between Mild and Moderate COVID-19](https://www.sciencedirect.com/science/article/pii/S0092867420314446) used the `hierarchy.fcluster` function to perform T cell receptor (TCR) clustering analysis. We will resolve the [two open bug reports](https://github.com/scipy/scipy/issues?q=is%3Aissue+is%3Aopen+fcluster+) about `fcluster` to facilitate future use.
- [ ] An author of [The impact of reducing signal acquisition specifications on neuronal spike sorting](https://ieeexplore.ieee.org/document/9630669) suggested "having helper functions within scipy.signal.fft that 1) converts the fft coefficients to power (V**2/Hz) and 2) converts power and phase to fft coefficients would be nice. I always forget how to do the conversion and a function that does this would be nice to have". 
- [ ] The same author suggested improving the speed of [scipy.signal.resample](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.resample.html) and "having an option to resample via a non-Fourier method in the case the signal is not periodic"
[...]

PR #22538, opened 2025-02-15
RFC: spatial.distance: add periodic boundary conditions

mdhaber:
#### Reference issue
gh-16191

In gh-16191, a user noted that "awareness of periodic boundary conditions in `scipy.spatial` would be helpful". As a step in that direction, this PR proposes a decorator that adds support for periodic boundary conditions to distance metrics that depend on `|u_i - v_i|`.

#### Additional information
No tests at the moment; just thought I'd prototype this and get some feedback as to whether this feature is something to work on. 

Certainly there are other places in `spatial` where periodic BCs would be more useful; this is only intended to be a small step. If this goes well, perhaps `pdist` is on the horizon.


scipy/spatial/distance.py
```diff
@@ -307,6 +311,76 @@ def _validate_weights(w, dtype=np.float64):
...
+    doc = str(doc).split("\n", 1)[1]  # remove signature
+    wrapped.__doc__ = str(doc)
+
+    sig = inspect.signature(wrapped)
+    parameters = sig.parameters
+    parameter_list = list(parameters.values())
+    parameter_list.append(_periodic_bounds_param)
+    sig = sig.replace(parameters=parameter_list)
+    wrapped.__signature__ = sig
+    return wrapped
```
tylerjereddy, line 381: This construct is pretty hard to read IMO -- I might prefer just adding optional argument(s) to each function directly to be honest, if there really is demand/motivation for it. Common infrastructure could still be abstracted if it really is that common and really handles PBC in a general way for different container/unit cell space geometries.

It looks like a lot like the nan handling and rng handling stuff, which I've always found a bit hard to follow. In `spatial` I'd at least want to be a voice for potentially having a few more lines of code/docs vs. abstraction/overriding that gets a bit hard to follow, but happy to be outvoted as well.

tylerjereddy (commented):
As a downstream developer of MDAnalysis, I do think there is some value in PBC support, though probably the main place it would be useful at the moment is in `KDTree`.

At least in the molecular simulation world, handling the full breadth of PBC possibilities is pretty tricky--there are some classes/geometries of container type that are particularly challenging to handle without basically replicating the coordinates of the unit cell IIRC. In any case, I'd probably want to see a clear discussion of scope on PBC handling if we're going there.

scipy/spatial/distance.py
mdhaber: Sure. In `spatial`, I would defer to your preferences if this would be accepted otherwise. So before merging, I could eliminate the decorator and copy-paste everything to make the code as verbose as desired, but I'd rather keep it concise before then to avoid duplicate work.

mdhaber:
> In any case, I'd probably want to see a clear discussion of scope on PBC handling if we're going there.

I am only proposing to add PBC to these distance metrics, which seem pretty clear cut. If these would not be added without a more complete discussion of the trajectory of `spatial`, I can open an issue to start the conversation, but I don't have much to say; this is not my area.

tylerjereddy:
Let me ping @orbeckst and maybe @hmacdope and @richardjgowers re: PBC distance calculation handling downstream. I think Oli knows more than I about triclinic, orthorhombic, and so on, though I'm not sure how much of that applies vs. doesn't for the various metrics here.

It may be the case that downstream could just process to a common format before calling whatever SciPy does here, or it may be the case that they wouldn't care much about it/use it either way, but good to confirm/get their input.

scipy/spatial/distance.py
tylerjereddy: I think I was also a bit skeptical that we'd get away with things being that common across the PBC handling scenarios that are possible (i.e., the implementations might fork across metrics depending on how general the PBC handling is). Anyway, I've pinged some folks whom I think are interested in PBC to get some feedback. Maybe it really is that simple but good to ask..

scipy/spatial/distance.py
```diff
@@ -307,6 +311,76 @@ def _validate_weights(w, dtype=np.float64):
...
+        _name, inspect.Parameter.KEYWORD_ONLY, default=None)
+
+    def wrap_to_bounds(x, lb, period):
+        return _lazywhere(np.isfinite(period),
+                          (x, lb, period),
+                          f=lambda x, lb, period: (x - lb) % period + lb,
+                          f2=lambda x, lb, period: x)
+
+    @functools.wraps(fun)
+    def wrapped(*args, periodic_bounds=None, **kwargs):
```
mdhaber, line 336: This can probably simplified. For these metrics, I think all we need is the periodicity.

scipy/spatial/distance.py
mdhaber: I don't think they will be common across all. I applied this only to the metrics for which $d(u, v) = \sum_i f(|u_i - v_i|)$ and $f$ is increasing. It seemed that if there is period $\lambda_i$ in the coordinate $i$, we can take 

```math
  |u_i - v_i| = \min \left( \mod(u_i - v_i, \lambda_i), \quad \lambda_i - \mod(u_i - v_i, \lambda_i) \right)
```

Because $d$ is just a sum of some increasing functions of these terms, it is safe to find the minimum distance by doing the minimization above for each coordinate separately.

tylerjereddy: Yeah, I've just had bad experiences depending on PBC-handling code doing the right thing when I work with different geometries. For example, with cubic-style unit cell, "ok" it usually works, but with splitting over a rhombic dodecahedron I usually can't depend on the scientific Python ecosystem to do the right thing, even in downstream libs that are more specialized.

If you've already thought about all those geometries/scenarios then you're way ahead of me--that's my main point of caution.

mdhaber: > If you've already thought about all those geometries

No, I think this makes an assumption about the geometry (which we would document) and hopes that is useful in some applications. This seemed natural enough that it would be useful across domains. I didn't imagine we'd offer options for more exotic geometries in SciPy.

mdhaber:
@orbeckst since drafting this, I've read up on PBCs a bit, and I see that generalization to triclinic cells might be useful. If you think so, what would be the prefered way of specifying the relative orientations of the axes, and in what coordinates would you want to specify the inputs?

orbeckst:
In MD we are using the three unit cell ("box") vectors in the form of a 3x3 matrix where the columns are the first, second, and third vectors in a (hopefully) right handed coordinate system.

We keep coordinates in the standard Cartesian coordinate system (axes (1,0,0), (0,1,0), (0,0,1)). In materials science, people often use relative/fractional coordinates in terms of the unit cell vectors, though. 

### See Also
As a "See Also": We call this representation "[triclinic_dimensions](https://docs.mdanalysis.org/stable/documentation_pages/coordinates/timestep.html#MDAnalysis.coordinates.timestep.Timestep.triclinic_dimensions)" but that's more of a local jargon than an official name, I think. [MDAnalysis.lib.mdamath.triclinic_vectors()](https://docs.mdanalysis.org/stable/documentation_pages/lib/mdamath.html#MDAnalysis.lib.mdamath.triclinic_vectors) converts from `[Lx, Ly, Lz, alpha, beta, gamma]` (where the first axis is parallel to (1,0,0) and has length Lx, the second axis is in the x-y plane (i.e. the plane's normal is (0,0,1)) with length Ly and angle alpha to (0,1,0), and the third axis makes angles beta and gamma to the other two — this has been the traditional representation in MD but it cannot represent rotations of the box.)  

More see also: We had some long discussions on PBC in https://github.com/MDAnalysis/mdanalysis/issues/1475 and found @Tsjerk's dissertation quite helpful (Ch 3, Appendix C in particular):

* T. A. Wassenaar. Molecular dynamics of sense and sensibility in processing and analysis of data. PhD thesis, University of Groningen, Groningen, The Netherlands, 2006. https://hdl.handle.net/11370/b0c3a19b-9f60-4911-ab23-d9725a2d45a2

mdhaber:
> In materials science, people often use relative/fractional coordinates in terms of the unit cell vectors, though.

Great! That's exactly why I was asking.

So let me rephase to check my understanding:
- You express points in normal, unbounded Cartesian coordinates instead of, say, fractional coordinates in terms of unit cell vectors.
- You define the unit cell using your unit cell vectors in a Cartesian coordinate system instead of, say, `[Lx, Ly, Lz, alpha, beta, gamma]`. These vectors happen to be collected in a single matrix, but the columns of the matrix are just vectors that define the length and orientation of sides of the unit cell.

One other question I have, in case you know the answer. 

With an orthorhombic cell, I can find the (minimum) distance considering the PBCs by looking at each coordinate individually. Conceptually, after wrapping points so they are in the same unit cell, I can check whether the difference in _each coordinate_ between the two points is smaller if I stay within the cell or go around the boundary associated with that coordinate. Then, I can use those three differences in coordinates to compute the distance. I wrote this out somewhat mathemetically in https://github.com/scipy/scipy/pull/22538#discussion_r1957214399.

Besides the complexities associated with having basis vectors that aren't orthogonal, can I still do something *conceptually* similar with triclinic unit cells? Like, if I have already wrapped my points into the same unit cell, and I've expressed these points in terms of the non-orthogonal unit vectors that define the unit cell, can I still find the minimum differences in each coordinate separately (considering the possibility of wrapping), and then (correctly compensating for the fact that the basis vectors are not orthogonal) compute the distance from these coordinate differences?

Tsjerk:
Funny. Just as I had turned to wrapping my head around triclinic lattice Voronoi regions again, I'm being poked with a related thread. A small remark to Oliver's post: the handedness doesn't matter, as any non-parallel set of lattice vector works, and any combination of them does. It doesn't affect the space. That being said, for finding the nearest neighbours, the choice of lattice vectors matters, and you want to first find the **reduced lattice**, which is the one in which no two vectors have a projection over half onto each other. This corresponds to the most regular cell. Unfortunately, then getting the nearest neighbours is not a matter of finding the minimal distance over the dimensions. With skewing, there may be a combination of lattice vectors that gives a shift with a smaller distance. This is illustrated in the appendix of Chapter 3 of my thesis. However, you can already set a bound on the distance: if the distance is shorter than half the minimal diagonal value of the (reduced) box matrix, it's the minimal distance, period. This may be particularly useful when determining distances between point sets. If the distance is larger than the bound, there may be a shorter distance. You should then check combinations of shifts.

mdhaber:
Thanks @Tsjerk! Could you link to your thesis? There's not much to go on in your GH profile : )

Tsjerk:
Oh, Oliver had already done that a few posts back:

T. A. Wassenaar. Molecular dynamics of sense and sensibility in processing and analysis of data. PhD thesis, University of Groningen, Groningen, The Netherlands, 2006. https://hdl.handle.net/11370/b0c3a19b-9f60-4911-ab23-d9725a2d45a2

I did a quick check, and it appears that for a rhombic dodecahedron with a square base 30% of the contents needs to be shifted to get the shortest distance after putting everything in the triclinic cell. If the rhombic dodecahedron is defined with a hexagonal base, then this appplies to 25% of the contents, but in that case there are only single vector shifts to be checked, and it is a priori clear which direction, so only three shifts need to be considered. If the lattice with square base is defined by the lattice vectors u, v, w, you may simply take the lattice L = (u, w-v, w), put all of space P in there: Q = P L^-1, Q -= np.floor(Q), then for the volume where Q.sum(axis=0) < 1 check positive shifts, and where Q.sum(axis=0) > 2 check negative shifts. This is not generic for triclinic lattices, but specific for orthorhombic ones. If the lattice is almost that, the volume of error will be small.

orbeckst:
To explicitly state that you understood me correctly @mdhaber :
> So let me rephase to check my understanding:
> *   You express points in normal, unbounded Cartesian coordinates instead of, say, fractional coordinates in terms of unit cell vectors.

Yes.

> *   You define the unit cell using your unit cell vectors in a Cartesian coordinate system instead of, say, `[Lx, Ly, Lz, alpha, beta, gamma]`. 

Some MD codes natively define the unit cell with `[Lx, Ly, Lz, alpha, beta, gamma]`, other with the "box matrix",i.e., the unit cell vectors. I would recommend adopting the unit cell vectors as the more natural (and more information-rich) representation.

> * These vectors happen to be collected in a single matrix, but the columns of the matrix are just vectors that define the length and orientation of sides of the unit cell.

Correct.

mdhaber:
Thanks @orbeckst! The other part of the question was whether it be useful to start with orthorhombic cells - that is, a diagonal matrix, or at least orthogonal columns?

orbeckst:
Even orthorhombic would be useful as a start.

I am not sure about all the metrics — in my work I only use Euclidean distances when it comes to molecules.

mdhaber:
Ok @tylerjereddy, I think we have some more clarity now.

This PR is limited to orthorombic unit cells, but it is still a useful start. It would be modified to accept (initially orthogonal) matrices that define the unit cells. This would establish a very general convention that could be extended to other functions and triclinic unit cells (non-orthogonal matrices) in the future. 

I think maintainers need to decide whether this is a direction that SciPy should go down. If so, does the proposed convention sounds good? Is it is OK to start with a PR with this scope, even if I may not be available to extend the capabilities further? Let me know your thoughts and who else should weigh in. Thanks!

tylerjereddy:
> I think maintainers need to decide whether this is a direction that SciPy should go down.

I'm not 100 % sure yet, but I'm just chiming in to say that I saw the ping and I'm pleased that Tsjerk and Oli have weighed in. I'll make a note to circle back..

Issue #7168, opened 2017-03-14
DOC: Add "Examples" to docstrings


WarrenWeckesser:
Many of the functions in scipy do not include examples in their docstrings.  Even a simple example can be helpful for a new user, so we should try to include examples wherever possible.

Here's a script that searches for scipy functions that are missing an "Examples" section:

```
"""
Find scipy functions whose docstrings do not contain "Examples".

This looks for *functions* only, not all callables.
"""

from __future__ import print_function

import importlib
import types
import numpy as np
import scipy

modules = ['cluster.hierarchy', 'cluster.vq', 'constants', 'fftpack',
           'integrate', 'interpolate', 'io', 'linalg', 'misc', 'ndimage',
           'odr', 'optimize', 'signal', 'sparse', 'sparse.linalg',
           'sparse.csgraph', 'spatial', 'special', 'stats',
           'stats.contingency', 'stats.mstats']

skip = ['misc.lena']

print("scipy version %s" % scipy.__version__)
print()

total = 0
for module_name in modules:
    mod = importlib.import_module('.' + module_name, package='scipy')
    objects = [(name, getattr(mod, name))
                   for name in getattr(mod, '__all__', dir(mod))
                       if not name.startswith('_')]
    funcs = [item for item in objects
                      if isinstance(item[1], (types.FunctionType,
                                              types.BuiltinFunctionType,
                                              np.ufunc))]
    noex = [item for item in funcs
                     if ((module_name + '.' + item[0]) not in skip and
                         (item[1].__doc__ is None or
                          ("is deprecated" not in item[1].__doc__) and
                          ("Examples" not in item[1].__doc__)))]
    if len(noex) > 0:
        total += len(noex)
        print(module_name, "(%d)" % len(noex))
        for name, func in noex:
            print("   ", name, end='')
            if func.__doc__ is None:
                print(" \t[no docstring]")
            else:
                print()

print()
print('Found %d functions' % total)

```

[...]

PR #21454, opened 2024-08-26
ENH: special: Add root finding tools to xsf and implement gdtrib for GPU while making substantial improvements

steppi:
#### Reference issue

This will tick a box for https://github.com/scipy/scipy/issues/20223

##### Root finding

This PR implements scalar root-finding tools in xsf that we can use to bring functions from cdflib from the GPU. Rather than trying to translate the root finding tools from the current cdflib C translation, I rewrote them from scratch, following the Algol code in the paper [Two Efficient Algorithms with Guaranteed Convergence for Finding a Zero of a Function](https://dl.acm.org/doi/10.1145/355656.355659) by Bus and Dekker, for the implementation of the root finding algorithm used in cdflib, and writing an initial bracket finder based on the general idea from @mdhaber's [bracket_root](https://github.com/scipy/scipy/blob/4d2c9e050663efd54396a3d8749792fad0b55dde/scipy/optimize/_elementwise.py#L462-L624), which itself uses well understood bracket finding techniques. I made some adjustments to accommodate searching over the entire range of double values. Both the root finding and bracket finding routines assume monotonicity.

The implementation of root finding tools in cdflib is hopelessly tangled, due its use of a reverse communication style which is a workaround for Fortran 77's lack of callbacks. I think the new implementation is much easier to follow, and because I'm able to understand, I was able to make some improvements. I had pointed out [here](https://github.com/scipy/scipy/pull/19824#issuecomment-1902098317) that the cdflib implementation doesn't do a good job of letting you control convergence. The rate of convergence outspeeds the reported tolerance, so tolerance was set at 1e-8 originally, because setting the tolerance too low often caused additional jumps that lead to a worse solution. The new implementation uses a criterion which allows us to search to the limits of machine precision, which will be seen in the improvement in the quality of results.

##### gdtrib

I have implemented `gdtrib` from cdflib as a proof of concept. The work in this PR is only a small fraction of the work I've done relating to cdflib. I've translated much of it to C++, and benchmarked it's routines against those from cephes. Rather than using the incomplete gamma function routine from cdflib, I've opted to use `igam` and `igamc` from cephes, which are comparable in accuracy, and have the benefit of not allocating arrays internally; which means they can be used on the GPU. Cephes is also what is used in `gdtr`, so by using cephes, `gdtrib` becomes more of a proper inverse with respect to `b` of our `gdtr`. I think it would be good to follow a general principle that our inverse cdf functions invert the actual implementations we use for the `cdf`.

##### Unable to use std::function

[...]

PR #19492, opened 2023-11-08
MAINT: special: use faddeeva for real-valued erf and ndtr

ev-br:
#### Reference issue

inspired by https://github.com/scipy/scipy/issues/19486

#### What does this implement/fix?

Route `erf` and `ndtr` calculations through faddeeva for real-valued arguments. Previously real-valued versions were from cephes and complex-valued versions were from faddeeva.

PR #22539, opened 2025-02-16
ENH/MAINT: migrate `special.chndtr` and its inverses to boost

dschmitz89:
#### Reference issue
Towards #21966  and #7168

This migrates `special.chndtr` and its inverses `chndtrix`, `chndtrinc` and `chndtridf` to use boost as backend.

#### Additional information
I had to juggle some 20 year old tests. Where possible I took them over in a new testing class for the noncentral chi squared special functions.

#### Benchmark results
I set up a benchmark for a few hundred cases using an a mpmath reference for the CDF. Boost is also not perfect but clearly better. Note that the cases of very high error could also be a consequence of wrong reference values. Infinite series require a high number of digits in mpmath and I increased it to 400 but one can never know.

ncx2_benchmark_results_PR_vs_main

<details>
from mpmath import mp
import numpy as np
import os

"""
mp.dps = 400

BRANCH = "pr"

os.mkdir("./ncx2_benchmark_results") if not os.path.exists("./ncx2_benchmark_results") else None

    return rel_error

@np.vectorize
def noncentral_chi_squared_cdf(x, df, nc):
    print(f"Calculating noncentral chi-squared CDF for x={x}, df={df}, nc={nc}")
    x, df, nc = map(mp.mpf, (x, df, nc))
    def term(i):
        return (mp.exp(-nc/2) * (nc/2)**i / mp.factorial(i) *
                mp.gammainc(df/2 + i, 0, x/2, regularized=True))
    try:
        return float(mp.nsum(term, [0, mp.inf]))
    except Exception:
        return np.nan

number_of_samples = 8
rng = np.random.default_rng(12345673423235)
df = sc.exp10(rng.uniform(-4, 4, number_of_samples))
x = sc.exp10(rng.uniform(-3, 2, number_of_samples))
nc = sc.exp10(rng.uniform(-2, 3, number_of_samples))

df, x, nc = np.meshgrid(df, x, nc)
df, x, nc = df.ravel(), x.ravel(), nc.ravel()

p_ref = np.load("ncx2_benchmark_results/ncx2_p_mpmath.npy") #noncentral_chi_squared_cdf(x, df, nc)#

#np.save(f"ncx2_benchmark_results/ncx2_p_mpmath.npy", p_ref)

#print(f"p_ref: {p_ref.shape}, p: {p.shape}")
mask = (p_ref > 0) & (p_ref < 1)
#print(f"mask: {mask.shape}, {np.sum(mask)}")
p_ref = p_ref[mask]
x = x[mask]
df = df[mask]
nc = nc[mask]

print(f"Number of valid samples: {len(p_ref)}")

p = sc.chndtr(x, df, nc)
chndtrinc = sc.chndtrinc(x, df, p_ref)
chndtridf = sc.chndtridf(x, p_ref, nc)
chndtrix = sc.chndtrix(p_ref, df, nc)
rel_error_chndtr = calculate_rel_error(p, p_ref)
rel_error_chndtrix = calculate_rel_error(chndtrix, x)
rel_error_chndtrinc = calculate_rel_error(chndtrinc, nc)
rel_error_chndtridf = calculate_rel_error(chndtridf, df)

np.save(f"ncx2_benchmark_results/chndtr_rel_error_{BRANCH}.npy", rel_error_chndtr)
np.save(f"ncx2_benchmark_results/chndtrix_rel_error_{BRANCH}.npy", rel_error_chndtrix)
np.save(f"ncx2_benchmark_results/chndtrinc_rel_error_{BRANCH}.npy", rel_error_chndtrinc)
np.save(f"ncx2_benchmark_results/chndtridf_rel_error_{BRANCH}.npy", rel_error_chndtridf)

"""
rel_error_chndtr_main = np.load("ncx2_benchmark_results/chndtr_rel_error_main.npy")
rel_error_chndtrix_main = np.load("ncx2_benchmark_results/chndtrix_rel_error_main.npy")
rel_error_chndtrinc_main = np.load("ncx2_benchmark_results/chndtrinc_rel_error_main.npy")
rel_error_chndtridf_main = np.load("ncx2_benchmark_results/chndtridf_rel_error_main.npy")

rel_error_chndtr_pr = np.load("ncx2_benchmark_results/chndtr_rel_error_pr.npy")
rel_error_chndtrix_pr = np.load("ncx2_benchmark_results/chndtrix_rel_error_pr.npy")
rel_error_chndtrinc_pr = np.load("ncx2_benchmark_results/chndtrinc_rel_error_pr.npy")
rel_error_chndtridf_pr = np.load("ncx2_benchmark_results/chndtridf_rel_error_pr.npy")
nbins = int(np.sqrt(len(rel_error_chndtr_main)))

print(rel_error_chndtr_main.max())
print(rel_error_chndtr_main.min())
plt.figure(figsize=(20, 12))
plt.subplot(2, 2, 1)
plt.hist(np.log10(rel_error_chndtr_main), bins=nbins, alpha=0.5, label='main', density=True, cumulative=True)
plt.hist(np.log10(rel_error_chndtr_pr), bins=nbins, alpha=0.5, label='pr', density=True, cumulative=True)
plt.xlabel("log10(Relative Error)")
plt.grid()
plt.title("Relative Error of chndtr")
plt.legend()

plt.subplot(2, 2, 2)
plt.hist(np.log10(rel_error_chndtrix_main), bins=nbins, alpha=0.5, label='main', density=True, cumulative=True)
plt.hist(np.log10(rel_error_chndtrix_pr), bins=nbins, alpha=0.5, label='pr', density=True, cumulative=True)
plt.xlabel("log10(Relative Error)")
plt.title("Relative Error of chndtrix")
plt.grid()
plt.legend()

plt.subplot(2, 2, 3)
plt.hist(np.log10(rel_error_chndtrinc_main), bins=nbins, alpha=0.5, label='main', density=True, cumulative=True)
plt.hist(np.log10(rel_error_chndtrinc_pr), bins=nbins, alpha=0.5, label='pr', density=True, cumulative=True)
plt.grid()
plt.xlabel("log10(Relative Error)")
plt.title("Relative error of chndtrinc")
plt.legend()

plt.subplot(2, 2, 4)
plt.hist(np.log10(rel_error_chndtridf_main), bins=nbins, alpha=0.5, label='main', density=True, cumulative=True)
plt.hist(np.log10(rel_error_chndtridf_pr), bins=nbins, alpha=0.5, label='pr', density=True, cumulative=True)
plt.grid()
plt.xlabel("log10(Relative Error)")
plt.legend()
plt.title("Relative error of chndtridf")
plt.savefig(f"/mnt/c/Development/ncx2_benchmark_results_PR_vs_main.png")#
</details>


scipy/special/tests/test_basic.py
```diff
@@ -199,36 +199,6 @@ def test_chdtri(self):
...
-            [350.0, 300., 10., 0.93880128006276407710],
-            [100.0, 13.5, 10., 0.99999999650104210949],
-            [700.0, 20.0, 400, 0.99999999925680650105],
-            [150.0, 13.5, 10., 0.99999999999999983046],
-            [160.0, 13.5, 10., 0.99999999999999999518],  # 1.0
-        ])
-        cdf = cephes.chndtr(values[:, 0], values[:, 1], values[:, 2])
-        assert_allclose(cdf, values[:, 3], rtol=1e-12)
-
-        assert_almost_equal(cephes.chndtr(np.inf, np.inf, 0), 2.0)
```
dschmitz89, line 226: I do not understand this test. Why should a CDF be 2? I simply removed it.

scipy/special/tests/test_basic.py
```diff
@@ -199,36 +199,6 @@ def test_chdtri(self):
...
-        assert_almost_equal(cephes.chndtr(2, 1, np.inf), 0.0)
```
dschmitz89, line 227: With boost, `nc=inf` results in `nan`, not 0. This test was removed for the time being. Not sure what the better behaviour is. If requested, this could still be handled as an edge case in the wrapper.

scipy/special/tests/test_cdflib.py
```diff
@@ -695,3 +695,83 @@ def test_nctdtr_accuracy(self, df, nc, x, expected, rtol):
...
+
+    @pytest.mark.parametrize(
+        "x, df, nc, expected_cdf, rtol",
+        [(0.1, 200, 50, 1.1311224867205481e-299, 1e-13),
+         (1e-12, 20, 50, 3.737446313006551e-141, 1e-13),
+         (1, 200, 50, 8.09760974833666e-200, 1e-13),
+         (9e4, 1e5, 1e3, 1.6895533704217566e-141, 5e-12)]
+    )
+    def test_chndtr_left_tail(self, x, df, nc, expected_cdf, rtol):
+        assert_allclose(sp.chndtr(x, df, nc), expected_cdf, rtol=rtol)
```
dschmitz89, line 748: These only pass with boost but fail with cdflib.

scipy/special/tests/test_basic.py
lucascolley: looks like this came from https://github.com/scipy/scipy/pull/3153 — any idea @ev-br ?

mdhaber: Fine to remove it.

scipy/special/tests/test_basic.py
mdhaber: I'd think 0 is right, actually, unless there are zero degrees of freedom.

scipy/special/tests/test_cdflib.py
```diff
@@ -695,3 +695,83 @@ def test_nctdtr_accuracy(self, df, nc, x, expected, rtol):
...
+
+    @pytest.mark.parametrize("args",
+        [(np.inf, 1, 1), (np.inf, np.inf, 1), (np.inf, np.inf, np.inf)]
+    )
+    def test_chndtr_x_inf_equals_1(self, args):
+        assert_allclose(sp.chndtr(*args), 1.0, rtol=0.0)
```
mdhaber, line 754: I'm not sure that this is right when either degrees of freedom or noncentrality is infinite.

scipy/special/_add_newdocs.py
```diff
@@ -1470,7 +1470,43 @@ def add_newdoc(name, doc):
...
-    chndtrix, chndtridf, chndtrinc
+    chndtrix: Noncentral Chi Squared distribution quantile
+    chndtridf: Inverse of `chndtr` with respect to `df`
+    chndtrinc: Inverse of `chndtr` with respect to `nc`
+    scipy.stats.ncx2: Non-central chi-squared distribution
+
+    Notes
+    -----
+    The noncentral chi squared distribution is also available in
+    `scipy.stats.ncx2`. `scipy.stats.ncx2.cdf` is equivalent to `chndtr`.
```
dschmitz89, line 1481:
```suggestion
    `scipy.stats.ncx2`. ``scipy.stats.ncx2.cdf`` is equivalent to `chndtr`.
```

scipy/special/tests/test_basic.py
steppi: That test using a reference value of `2.0` for a cdf is something. I assume it just escaped noticed when that PR was getting reviewed. I think @mdhaber is right, `chndtr(np.inf, np.inf, 0)` should be `NaN`.

dschmitz89:
@steppi @mdhaber : let me know what the edge case behaviour should be in your opinion for infinite df and nc, I do not have a strong opinion there. Or we could leave the edge case handling for a followup.

scipy/special/boost_special_functions.h
```diff
@@ -721,12 +721,32 @@ template<typename Real>
 Real
 ncx2_cdf_wrap(const Real x, const Real k, const Real l)
```
dschmitz89, line 722: For long maintainability, it might be actually better to rename all these variables to x, df and nc throughout stats and special.

mdhaber:
Below, I'm using `1` to stand in for any positive, finite number. Based on the definition of a noncentral chi-squared random variable - not mathematical limits - I think:

When `nc = 0`, the distribution is just chi-squared, so we should just follow `chdtrc`.
- `df=0`: distribution is point mass at `x = 0`. Seems to me that the CDF should be 1 when evaluated at any non-negative argument because there is 100% probability that the random variable will assume a value <= 0. But `chdtrc`, which we recently worked on, returns NaN at `x = 0`, so I guess we should stick with that.
- `df=1`: I think this works as-is.
- `df=np.inf`: CDF should be 0 for any finite `x` and NaN when `x = np.inf`

When `nc = 1`:
- `df=0`: This doesn't make sense to me because the sum of zero squared means can't be 1. NaN across the board (as it is now).
- `df=1`: works as-is
- `df=np.inf`: this means the sum of infinite squared means is 1... Let's stick with NaNs across the board.

When `nc = np.inf`:
- `df=0`: same as `nc = 1` with zero degrees of freedom; doesn't make sense. NaNs across the board.
- `df=1`: Seems to me that the CDF should be zero when evaluated at any finite, non-negative argment, and NaN when evaluated at `x=np.inf`.
- `df=np.inf`: I think this is the same for any positive degrees of freedom.

dschmitz89:
> Below, I'm using `1` to stand in for any positive, finite number. Based on the definition of a noncentral chi-squared random variable - not mathematical limits - I think:
> 
> When `nc = 0`, the distribution is just chi-squared, so we should just follow `chdtrc`.
> 
> * `df=0`: distribution is point mass at `x = 0`. Seems to me that the CDF should be 1 when evaluated at any non-negative argument because there is 100% probability that the random variable will assume a value <= 0. But `chdtrc`, which we recently worked on, returns NaN at `x = 0`, so I guess we should stick with that.
> * `df=1`: I think this works as-is.
> * `df=np.inf`: CDF should be 0 for any finite `x` and NaN when `x = np.inf`
> 
> When `nc = 1`:
> 
> * `df=0`: This doesn't make sense to me because the sum of zero squared means can't be 1. NaN across the board (as it is now).
> * `df=1`: works as-is
> * `df=np.inf`: this means the sum of infinite squared means is 1... Let's stick with NaNs across the board.
> 
> When `nc = np.inf`:
> 
> * `df=0`: same as `nc = 1` with zero degrees of freedom; doesn't make sense. NaNs across the board.
> * `df=1`: Seems to me that the CDF should be zero when evaluated at any finite, non-negative argment, and NaN when evaluated at `x=np.inf`.
> * `df=np.inf`: I think this is the same for any positive degrees of freedom.
> 
> @steppi It's pretty late, so I might not be thinking straight, but I think most of these suggestions are quite safe except for the last two - what do you think about those (infinite noncentrality, positive degrees of freedom)?

@mdhaber : just to double check, is the table below correct? I would like to be sure before coding the branching madness.

| x | df | nc | target |
|--------|--------|--------|--------|
| 1 | 0 | 0 | nan |
| 1| inf | 0 | 0 |
| inf | inf | 0 | nan |
| 1 | 0 | 1 | nan |
| 1 | inf | 1 | nan |
| inf | inf | 1 | nan
| 1 | 0 | inf | nan |
| 1 | 1 | inf | 0 |
| inf | 1 | inf | nan |
| 1 | inf | inf | nan |
| inf | inf | inf | nan |

mdhaber:
Not quite. 

The first row should have a target of `1.0`. For the `nc=0` cases, can you just delegate to `chdtr` so we don't have to worry about them? There are three rows in the table, but many more cases.

The three rows you have for `nc=1` look right, but again, there are many more cases to be covered.

For `nc = inf`, I wrote that the case should be the same for any positive degrees of freedom, so the target of the second to last row and fourth to last row should match: 0.

I think the logic is:
```
if nc == 0:
    # use chdtr
elif (0 < nc < inf and df == inf) or (df == 0):  # or if any argument is negative, I assume
    return NaN
elif nc == inf:
    return 0 if x < inf else NaN
else:
    # use Boost
```
Please check what that looks like it agrees with what I wrote.

dschmitz89:
@mdhaber : thanks for the clarification, those cases make sense. One issue I see is that `chdtr` only works with doubles as it comes from cephes while `chndtr` works for single and double precision via boost. So I added special handling for `nc=0`. Many edge cases are now covered by the tests. If CI passes, I would prefer to leave any further edge case cleanup for a follow up.

bashtage:
It is hard to say.  I think it is probably better to merge and then we can check nightly.

bashtage:
It is hard to say.  I think it is probably better to merge and then we can check nightly.

dschmitz89:
@bashtage: thanks for the reply. I just noticed that it might be better to actually migrate all functions in one go if you do round trip checks with tight tolerances. WDYT?

bashtage:
It's probably something easier for us if the functions are migrated in coherent sets, w.g., all chi2, so we can trace it back to a PR that would have changed functions.

dschmitz89:
Converted to draft until the inverse functions are also done.

scipy/special/tests/test_cdflib.py
```diff
@@ -710,3 +711,100 @@ def test_nctdtr_accuracy(self, df, nc, x, expected, rtol):
...
+        cdf = sp.chndtr(values[:, 0], values[:, 1], values[:, 2])
+        assert_allclose(cdf, values[:, 3], rtol=1e-13)
+        # the last two values are very close to 1.0, so we do not
+        # test the inverses for them
+        x = sp.chndtrix(values[:, 3], values[:, 1], values[:, 2])
+        assert_allclose(x[:-2], values[:-2, 0], rtol=1e-8)
+        df = sp.chndtridf(values[:, 0], values[:, 3], values[:, 2])
+        assert_allclose(df[:-2], values[:-2, 1], rtol=1e-8)
+        nc = sp.chndtrinc(values[:, 0], values[:, 1], values[:, 3])
+        assert_allclose(nc[:-2], values[:-2, 2], rtol=1e-8)
```
dschmitz89, line 746: I added these test cases to provide at least some tests against arbitrary precision reference values also for the inverses. The tests using the `assert_inverts` machinery fail for `chndtridf` and `chndtrinc` also with boost but I could not find a quick way to debug that beast.

ev-br:
> Converted to draft until the inverse functions are also done.

Reading the intention, have just actually converted to draft and bumped the milestone.

dschmitz89:
> > Converted to draft until the inverse functions are also done.
> 
> Reading the intention, have just actually converted to draft and bumped the milestone.

Bumping te milestone works for me but this PR is actually done, ready for review.

scipy/special/boost_special_functions.h
```diff
@@ -735,12 +735,47 @@ template<typename Real>
...
-            boost::math::non_central_chi_squared_distribution<Real, StatsPolicy>(k, l), x);
+    if (std::isnan(x) || std::isnan(k) || std::isnan(l)) {
+	return NAN;
     }
-    // -inf => 0, inf => 1
-    return 1 - std::signbit(x);
+    if (k < 0 || x < 0 || l < 0) {
+	sf_error("chndtr", SF_ERROR_DOMAIN, NULL);
+	return NAN;
+    }
```
ev-br, line 744: Minor, trivial, optional: the indentation in this part looks a bit weird.

ev-br (commented):
Thanks, and sorry for misreading the intention then!

From a quick look, all looks reasonable to me. A repeated pattern of converting boost exceptions into `sf_error` codes can probably be refactored into a less verbose solution, but that's certainly not critical.

Ran out of steam for today for a closer review though, and I'm not that familiar on how things are done in `scipy.special` these days. So ideally it's for Albert to review, unless Tyler is willing to YOLO it into 1.16 so that it can get tested by downstream consumers.

dschmitz89:
stats models relies on these functions, so we should do a proper review. If required I can do more extensive testing like you did with `gdtrib` for example @steppi .

steppi:
> stats models relies on these functions, so we should do a proper review. If required I can do more extensive testing like you did with `gdtrib` for example @steppi .

I'm not sure it's necessary to add more to the test suite. What I'd like to see though is a script to compare Boost and Cephes to an arbitrary precision reference over a wide range of points like I did for `gdtrib`.  I remember in this one, https://github.com/scipy/scipy/pull/22358, there was a bug that was only caught because `test_cython_special` happened to have a test case which triggered an `OverFlowError` that didn't get handled by `sf_error`. There were also problems in the past where a special function migration to Boost broke something for `statsmodels` due to less accuracy in particular regions. So I'd be concerned about merging this without a proper study.

Unlike specfun, Cephes is a production grade library and on par with Boost in many cases, so I think it makes sense to try to measure the improvement. For this PR, https://github.com/scipy/scipy/pull/19492, more extensive testing found that Cephes was much more accurate than a potential replacement from Faddeeva for `ndtr`.

The script I wrote here, https://github.com/scipy/scipy/pull/21454, to check for accuracy for `gdtrib` should be a good template to get started from. I think I should write a tutorial on extensive testing for special function implementations like this though.

dschmitz89:
> > stats models relies on these functions, so we should do a proper review. If required I can do more extensive testing like you did with `gdtrib` for example @steppi .
> 
> I'm not sure it's necessary to add more to the test suite. What I'd like to see though is a script to compare Boost and Cephes to an arbitrary precision reference over a wide range of points like I did for `gdtrib`. I remember in this one, #22358, there was a bug that was only caught because `test_cython_special` happened to have a test case which triggered an `OverFlowError` that didn't get handled by `sf_error`. There were also problems in the past where a special function migration to Boost broke something for `statsmodels` due to less accuracy in particular regions. So I'd be concerned about merging this without a proper study.
> 
> Unlike specfun, Cephes is a production grade library and on par with Boost in many cases, so I think it makes sense to try to measure the improvement. For this PR, #19492, more extensive testing found that Cephes was much more accurate than a potential replacement from Faddeeva for `ndtr`.
> 
> The script I wrote here, #21454, to check for accuracy for `gdtrib` should be a good template to get started from. I think I should write a tutorial on extensive testing for special function implementations like this though.

I agree that we should carry out very thorough benchmarks of cephes against boost. This PR replaces cdflib functions by boost though.

In principle, I can also perform more benchmarks, but the issue for the noncentral chi squared distribution is the generation of reference values. We need to perform numerical integration with mpmath with high `dps`. That plus root finding for the inverses will exhaust my private laptop for quite some time.

ev-br:
> That plus root finding for the inverses will exhaust my private laptop for quite some time.

If there's a script to run, I can run it on a server. Place it somewhere like like gist.github, I can run it and drop the results somewhere. ideally, think of a way you'll be able to verify the results though.

dschmitz89:
> > That plus root finding for the inverses will exhaust my private laptop for quite some time.
> 
> If there's a script to run, I can run it on a server. Place it somewhere like like gist.github, I can run it and drop the results somewhere. ideally, think of a way you'll be able to verify the results though.

Thanks for the offer, will come back to you if necessary. The CDF can also be computed by an infinite series, still takes very long unfortunately.

dschmitz89:
I added some more tests for the inverses here. Since boost is already used in stats for the noncentral chi squared distribution, `chndtr` and `chndtrix` should be good. On top, this PR moves the inverses for the degrees of freedom (`chndtridf`) and the noncentrality parameter (`chndtrinc`) from cdflib to boost. I do not know of any instance where cdflib was superior to boost and  am hesitating to perform large scale testing here due to the necessary computing resources. Can we agree to give this one green light if CI comes back green?

dschmitz89:
I added a benchmark in the opening post that demonstrates the improvements here. Unless there are more comments, I will merge this next week then. There might be edge cases left but they can be handled in a follow up. CI failures are unrelated.

dschmitz89:
@ilayn There are two fresh CI failures here: A segfault in MINPACK and a ARPACK precision issue on Windows on ARM. Thoughts?

PR #21613, opened 2024-09-23
ENH: sparse: add nD support for element-wise and reduction operations

anushkasuyal:
_Context: This is the third PR (follow-up to https://github.com/scipy/scipy/pull/21197 and #21435) aiming to extend support for n-D COO arrays in SciPy (refer to https://github.com/scipy/scipy/issues/21169). I am undertaking this work as part of my summer internship with Quansight Labs till the end of September, under the mentorship of @rgommers and @izaid._

This PR continues the extension of COO sparse arrays to n-dimensions, by implementing support for element-wise operations such as boolean comparators, `multiply`, `maximum`, `minimum` etc. between n-D COO arrays. Previously, these operations were possible for arrays of the same shape, but now, as we move towards aligning the sparse array behaviour with that of the array API standard, we allow for broadcasting (which was introduced in the PR #21435) of arrays of different shapes so that the above operations can be performed on them.

After CSR `_broadcast_to` is added in #21412, we can now broadcast CSR arrays too. Since this PR implements broadcasting for COO arrays, it would introduce inconsistencies if COO arrays differed in behaviour from other formats. Therefore, this PR also adds broadcasting to these operations for the CSR format.

To keep any addition of compiled code to a minimum, the basic approach followed by most of the methods implemented in this PR is to utilise NumPy's `ravel`/`unravel` and `reshape` tools to map n-D arrays to 2-D and use the pre-existing CSR sparsetools. Since COO allows for quick conversion to CSR, and CSR allows for faster arithmetic operations, this approach would prevent introducing more C++ functions while also utilising the CSR methods that have already been implemented and rigorously tested in the codebase.

We also implement reduction operations such as `max`, `min`, `argmax`, `argmin`, `nanmax`, `nanmin`, `sum`, and `mean` in this PR. Although these are not element-wise operations, they use an approach similar to that of element-wise operations - where we map to 2-D, route via CSR, and map back to n-D.

(Note- diff may be easier to view once #21435 and #21412 have been merged to main.)

Issue #21169, opened 2024-07-12
Extending SciPy COO Sparse Arrays to N Dimensions


anushkasuyal:
### Summary
This project is undertaken as part of my summer internship at Quansight Labs, mentored by @izaid and @rgommers.
I would like to propose extending the current COO sparse array support in SciPy to n-dimensions. This enhancement targets users dealing with high-dimensional data, providing efficient storage and computation, reducing memory overhead, and avoiding unnecessary dense format computations. It will also be helpful throughout many other parts of SciPy that rely on sparse arrays. For instance, there are many methods that use sparse arrays to compute a result, and extending them to n dimensions will enable broadcasting via stacked sets of sparse arrays.

### Motivation
Currently, SciPy's sparse arrays are limited to 1D and 2D, with high-dimensional operations relying on dense NumPy arrays, leading to significant memory and computational inefficiencies. Extending COO sparse arrays to n dimensions will:

1. **Improve Efficiency:** Avoid dense format storage for high-dimensional sparse tensors.
2. **Enhance Performance:** Reduce unnecessary computations in matrix and tensor operations.
3. **Expand Functionality:** Enable direct manipulation of n-dimensional sparse data.

### Proposed Changes
The implementation can be approached in a phased manner, focusing on one chunk at a time. Community feedback will be sought at each stage to ensure alignment with user needs.

**1.	Data Structure and Basic Methods**

- Extend COO format to support creation of n-dimensional sparse arrays (`toarray`), and facilitate conversions between COO and dense formats.
- Implement/extend basic methods to n-dimensions.
- Implement element wise arithmetic operations (addition, subtraction etc. of COO arrays)

**2.	Matrix Multiplication and Tensor Operations**
- Extend matrix multiplication for n-dimensional COO arrays (i.e. sparse array alternative for `numpy.matmul` (may require modifying compiled code for efficiency).
- Implement basic tensor operations for sparse arrays like `tensordot` and `tensorsolve`.

**3.	Other operations**
- Adapt other existing linear algebra functions for n-dimensional COO arrays.
- Implement Boolean comparators for COO arrays (currently, most comparisons are routed via the CSR format, which wouldn’t support sparse arrays of dim>2)
- Ensure seamless integration with other SciPy modules.

### Implementation Plan
[...]

PR #22540, opened 2025-02-16
ENH: sparse: Add nD support for elementwise and reduction operations without broadcasting

dschult:
This PR is a split of #21613 where I have removed the feature addition of broadcasting.
So this PR is only the elementwise and reduction operations for nD COO format.

Hopefully this will make it easier to review and get these nice features in.
Here is a recap (without broadcasting) of the description from #21613 

> This PR continues the extension of COO sparse arrays to n-dimensions, by implementing support for element-wise operations such as boolean comparators, multiply, maximum, minimum etc. between n-D COO arrays. 

> To keep any addition of compiled code to a minimum, the basic approach followed by most of the methods implemented in this PR is to utilise NumPy's ravel/unravel and reshape tools to map n-D arrays to 2-D and use the pre-existing CSR sparsetools. Since COO allows for quick conversion to CSR, and CSR allows for faster arithmetic operations, this approach would prevent introducing more C++ functions while also utilising the CSR methods that have already been implemented and rigorously tested in the codebase.

> We also implement reduction operations such as max, min, argmax, argmin, nanmax, nanmin, sum, and mean in this PR. Although these are not element-wise operations, they use an approach similar to that of element-wise operations - where we map to 2-D, route via CSR, and map back to n-D.

This is a continuation of the work done by @anushkasuyal as described in #21169

### DIFF
--- a/scipy/sparse/_coo.py
+++ b/scipy/sparse/_coo.py
@@ -18,7 +18,8 @@ from ._base import issparse, SparseEfficiencyWarning, _spbase, sparray
 from ._data import _data_matrix, _minmax_mixin
 from ._sputils import (upcast_char, to_native, isshape, getdtype,
                        getdata, downcast_intp_index, get_index_dtype,
-                       check_shape, check_reshape_kwargs, isscalarlike, isdense)
+                       check_shape, check_reshape_kwargs, isscalarlike, isdense,
+                       is_pydata_spmatrix)
 
 import operator
 
@@ -298,15 +299,11 @@ class _coo_base(_data_matrix, _minmax_mixin):
             M, N = self.shape
             coo_todense(M, N, self.nnz, self.row, self.col, self.data,
                         B.ravel('A'), fortran)
-        else:
-            if fortran:
-                strides = np.append(1, np.cumprod(self.shape[:-1]))
-            else:
-                strides = np.append(np.cumprod(self.shape[1:][::-1])[::-1], 1)
-            coords = np.concatenate(self.coords)
-            coo_todense_nd(strides, self.nnz, self.ndim,
-                           coords, self.data, B.ravel('A'), fortran)
-        # Note: reshape() doesn't copy here, but does return a new array (view).
+
+        else: # dim>2
+            flat_indices = np.ravel_multi_index(self.coords, self.shape)
+            B = np.zeros(self.shape, dtype=self.dtype)
+            np.add.at(B.ravel(), flat_indices, self.data)
         return B.reshape(self.shape)
 
     toarray.__doc__ = _spbase.toarray.__doc__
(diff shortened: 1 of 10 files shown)


scipy/sparse/_coo.py
```diff
@@ -298,15 +299,11 @@ def toarray(self, order=None, out=None):
...
-        else:
-            if fortran:
-                strides = np.append(1, np.cumprod(self.shape[:-1]))
-            else:
-                strides = np.append(np.cumprod(self.shape[1:][::-1])[::-1], 1)
-            coords = np.concatenate(self.coords)
-            coo_todense_nd(strides, self.nnz, self.ndim,
-                           coords, self.data, B.ravel('A'), fortran)
-        # Note: reshape() doesn't copy here, but does return a new array (view).
+
```
dschult, line 302: Just a note here that this removes use of `coo_todense_nd` in favor of the faster `ravel_multi_index/np.add.at`. We should do explore whether to do the same for the 1D case (~6 lines above) by using `coo_todense`... that would avoid including the `coo_todense_nd` function in sparsetools. But I think that could be a separate PR which also explores whether speedups in the code for `coo_todense_nd` are possible. (I recall the speed-up idea being to construct the strides in c++ instead of python.)  Let's look into that separately.

scipy/sparse/_coo.py
```diff
@@ -1275,6 +1262,460 @@ def _broadcast_to(self, new_shape, copy=False):
...
+            # reshaping back to n-D if output is 2-D boolean array
+            if isinstance(result, np.ndarray | sparray):
+                result = result.reshape(result_shape)
+
+            return result
+        else:
+            return NotImplemented
+
+    def __eq__(self, other):
+        return self._process_arrays_for_comparison(other, '__eq__')
```
perimosocordiae, line 1319: Let's change this to use operator functions instead, as that way we can avoid getattr on the base class.

scipy/sparse/_coo.py
```diff
@@ -1275,6 +1262,460 @@ def _broadcast_to(self, new_shape, copy=False):
...
+            if other_a.ndim == 0 and other_a.dtype == np.object_:
+                return NotImplemented
+
+            try:
+                other.shape
+            except AttributeError:
+                other = other_a
+
+        if self.ndim < 3 and (isscalarlike(other) or other.ndim < 3):
+            return getattr(_data_matrix, op_name)(self.tocsr(), other)
```
perimosocordiae, line 1282: We shouldn't be calling `self.tocsr()` here, we can let the base class method handle conversion if needed.

scipy/sparse/_coo.py
```diff
@@ -1367,6 +1808,34 @@ def _ravel_non_reduced_axes(coords, shape, axes):
     return np.ravel_multi_index(non_reduced_coords, non_reduced_shape)
 
 
+def _validateaxes(axis, ndim, shape):
```
perimosocordiae, line 1811: Let's see if we can merge this with the original validateaxes in sputils, maybe using a kwarg like `allow_nd` to gate the new behavior.

scipy/sparse/_data.py
```diff
@@ -210,7 +210,20 @@ def _min_or_max(self, axis, out, min_or_max, explicit):
         if out is not None:
             raise ValueError("Sparse arrays do not support an 'out' parameter.")
 
+        if axis == ():
+            return self.copy()
```
perimosocordiae, line 214: Let's raise an error in this case, as it's a no-op and unlikely to be what the user expected.

scipy/sparse/_sputils.py
```diff
@@ -408,21 +408,16 @@ def validateaxis(axis) -> None:
...
+    if not isinstance(axis, tuple):
+        if not np.issubdtype(np.dtype(axis_type), np.integer):
+            raise TypeError('axis must be an integer/tuple of ints,'
+                            f'not {axis_type.__name__}')
+        axis = [axis]
+
+    if len(axis)>2:
+        raise ValueError("axis tuple has too many elements")
+
+    if any(ax >= 2 or ax < -2 for ax in axis):
```
perimosocordiae, line 420: Let's normalize the `axis` into a tuple first, then we can loop over the elements and check that they're integers within the correct range (would require passing ndim here)

perimosocordiae (changes requested):

dschult:
This is superceded/replaced by 2 PRs #22873 and #22897 
The part of this not in those two PRs is removal of use of the sparsetools `coo_todense_nd`. That will be taken up in another PR that is in the works.

So I'm closing this PR.

Issue #22513, opened 2025-02-11
ENH: add `axis` parameter to FloaterHormannInterpolator and AAA


ev-br:
### Is your feature request related to a problem? Please describe.

Following up on https://github.com/scipy/scipy/pull/22484, the majority of `scipy.interpolate` allows batching: `y.ndim > 1`,  and the `axis` parameter selects which axis in `y` is the interpolation axis and the rest of axes are batching dimensions.

The newer `FloaterHormannInterpolator` and `AAA`  barycentric interpolator/approximator however lack the `axis` parameter. 

### Describe the solution you'd like.

Add batching/axis support to `FloaterHormannInterpolator` and `AAA` to make them walk and quack more like other interpolators.

mdhaber:
We have discussed documenting a recipe for adding batch support to `BSpline`s when different slices can have different numbers of knots:
https://github.com/scipy/scipy/blob/324a8c0477d016cdb4abd3b3a0e4dcd75b252a6c/scipy/interpolate/tests/test_bsplines.py#L3662-L3687

If it's not easy to support batching for these functions, we could do something similar here. What do you think @j-bowhay?

j-bowhay:
It should be straight forward to add an `axis` argument to `FloaterHormannInterpolator` as it already supports `y.dim > 1` it just doesn't have an `axis` argument.

Off the top of my head, `AAA` would require more thought as the barycentric weights and the support points depend on both `x` and `y` so would be different across slices so would be ragged. Something like the above would work as a low effort solution.

mdhaber:
Can I generalize this issue to be about consistency of univariate interpolator interfaces, or should I open a new one? 

I'd like to also track:
- Adding notes to the other functions/classes that don't support `axis` about why it is not supported, and point to the `BatchSpline` recipe (e.g. in the tutorial). Incidentally there was just a [request about this on S.O.](https://stackoverflow.com/questions/79428677/batch-make-smoothing-spline-in-scipy/79428992#79428992) the other day.
- `BSpline` has a `from_power_basis` method (that accepts only `CubicSpline` instances for now. How hard would it be to generalize? I ask because `PPoly` has `from_spline` and `from_bernstein_basis`, and `BPoly` has `from_power_basis`. This suggests that there could be methods that (from the user's perspective) convert from any one representation to any other representation (although we might actually implement these with `PPoly` as an intermediate representation).
[...]

PR #22541, opened 2025-02-16
Interpolate tutorial: discuss the bases and interconversions

ev-br:
Following up on https://github.com/scipy/scipy/issues/22513#issuecomment-2660517663, update the interpolate user guide:

- remove statements which turn out to be more confusing than helpful
- add a discussion of BPoly objects
- add a minimal discussion of conversions between different bases of piecewise polynomials. @mdhaber you wanted to expand this part?


doc/source/tutorial/interpolate/splines_and_polynomials.rst
```diff
@@ -203,10 +203,11 @@ B-splines are piecewise polynomials, represented as linear combinations of
...
 of usual monomials, :math:`x^m` with :math:`m=0, 1, \dots, k`.
 
-The b-spline basis is generally more computationally stable than the power basis
-and is useful for a variety of applications which include interpolation, regression
-and curve representation. The main feature is that these basis elements are
-*localized* and equal to zero outside of an interval defined by the *knot array*.
+The b-spline basis is used in a variety of applications which include interpolation,
+regression and curve representation. The main feature is that each basis element is
+*localized* and equal to zero outside of a certain interval (see below for an illustration).
+This is in contrast to usual monomials, :math:`x^m`, which are non-zero on the whole
```
j-bowhay, line 209: I get that this is trying to avoid introducing readers to what it means for a function to have compact support but this is maybe slightly confusing as $x^m$ is clearly zero at one point!

j-bowhay: It is also slightly confusing as we are told
> *b-spline basis elements* --- which themselves are certain linear combinations
of usual monomials
But then we're told b-spline basis elements are localised but monomials aren't which if you were reading for the first time could sound a bit contradictory

ev-br: Yeah, I can see how this is confusing. Any suggestions on how to make it less confusing?

ilayn: I don't think we need math-speak for compact support. _"These bases can take nonzero values only on a finite interval and do not extend to infinity like polynomial basis functions do. This makes them attractive for local approximations etc."_ or something along those lines maybe mentioning wavelets or whatever.

mdhaber: I think there are two distinct points of confusion above (which I actually understood, so they didn't both me) and two others that I *didn't* understand before. I will keep it short, but just say that there is an important distinction between a basis that is "the usual monomials $x^i$", $i \in {0, 1, \dots, n}$ and a basis of "piecewise (power basis) polynomials", each *segment* $i$ of which _can be thought of_ as the sum of $k + 1$ functions $f_{i, j} = x^j$, $j \in {0, 1, \dots, k}$. I think this is subtle, and easily missed unless the point is made very explicitly. If we want to avoid this section exploding, is there any way to avoid mentioning the fact that B-spline basis functions have compact support (regardless of whether we use "localized", "compact support", or something else)? Any readers like me would not appreciate that fact (without a lot of outside study), anyway.

ev-br: Okay, let's shorten the discussion a bit more, and refer the reader to the literature.

ev-br:
Close/reopen to flush CircleCI. The doc build errored out, and I wonder if it was a fluke, or it's Sphinx 8.2.0 released on Feb 18.

ev-br:
The doc build is green now after capping Sphinx to < 8.2.0 and adding a missing `linkify` requirement to the doc build environment.
The benchmark error is unrelated.

mdhaber:
A few final suggestions in ev-br/scipy#18.

doc/source/tutorial/interpolate/splines_and_polynomials.rst
```diff
@@ -199,14 +199,17 @@ of :math:`1, x, \cdots, x^k`.
...
 of usual monomials, :math:`x^m` with :math:`m=0, 1, \dots, k`.
 
-The b-spline basis is generally more computationally stable than the power basis
-and is useful for a variety of applications which include interpolation, regression
-and curve representation. The main feature is that these basis elements are
-*localized* and equal to zero outside of an interval defined by the *knot array*.
+The properties of b-splines are well described in the literature (see, for examples,
+references listed in the `BSpline` docstring). For our purposes, it is enough to know
+that a b-spline function is uniquely defined by an array of coefficients and
+and array of the so-called *knots*, which may or may not coincide with the data points,
```
j-bowhay, line 211:
```suggestion
an array of the so-called *knots*, which may or may not coincide with the data points,
```

doc/source/tutorial/interpolate/splines_and_polynomials.rst
```diff
@@ -286,3 +289,49 @@ contains basis elements evaluated at ``xnew[j]``:
...
 
+
+Bernstein polynomials, ``BPoly``
+================================
+
+For :math:`t \in [0, 1]`, Bernstein basis polynomials of degree :math:`k` are defined via
+
+.. math::
+
+    b(t; k, a) = C_k^a t^k (1-t)^{k - a}
```
j-bowhay, line 300:
```suggestion
    b(t; k, a) = C_k^a t^a (1-t)^{k - a}
```
Possibly typo?

doc/source/tutorial/interpolate/splines_and_polynomials.rst
```diff
@@ -286,3 +289,49 @@ contains basis elements evaluated at ``xnew[j]``:
...
+
+
+Conversion between bases
+========================
+
+In principle, all three bases for piecewise polynomials (the power basis, the Bernstein
+basis, and b-splines) are equivalent, and a polynomial in one basis can be converted
+into a different basis. One reason for converting between bases is that not all bases
+implement all operations. For instance, root-finding is only implemented for `PPoly`,
+and therefore to find roots of a `BSpline` object, you need convert to `PPoly` first.
```
j-bowhay, line 331:
```suggestion
and therefore to find roots of a `BSpline` object, you need to convert to `PPoly` first.
```

doc/source/tutorial/interpolate/splines_and_polynomials.rst
```diff
@@ -199,14 +199,17 @@ of :math:`1, x, \cdots, x^k`.
...
+regression and curve representation.
 B-splines are piecewise polynomials, represented as linear combinations of
 *b-spline basis elements* --- which themselves are certain linear combinations
 of usual monomials, :math:`x^m` with :math:`m=0, 1, \dots, k`.
 
-The b-spline basis is generally more computationally stable than the power basis
-and is useful for a variety of applications which include interpolation, regression
-and curve representation. The main feature is that these basis elements are
-*localized* and equal to zero outside of an interval defined by the *knot array*.
+The properties of b-splines are well described in the literature (see, for examples,
```
j-bowhay, line 208:
```suggestion
The properties of b-splines are well described in the literature (see, for example,
```

j-bowhay (changes requested):

ev-br:
Thanks Jake, took your edits along.

j-bowhay (approved):
LGTM!

rgommers:
> The doc build is green now after capping Sphinx to < 8.2.0 and adding a missing `linkify` requirement to the doc build environment.

What's going on with `linkify-it-py` there? We don't use it, so we shouldn't be adding a dependency without understanding why. If it's needed, it has to be a bug in one of our other doc dependencies.

ev-br:
> What's going on with linkify-it-py there? 

I was wondering about this, too. At least locally, the doc build fails without it.

rgommers (dismissed):
> I was wondering about this, too. At least locally, the doc build fails without it.

Let's investigate then - this PR shouldn't be merged with this dependency added, unless (a) we understand why, and (b) it's temporary and we know how/when to revert it.

ev-br:
The extension is present in https://github.com/scipy/scipy/blob/main/doc/source/conf.py#L411, and was added in https://github.com/scipy/scipy/blob/main/doc/source/conf.py#L411

In that PR it was manually added to `doc/requirements.txt` and `environment.yml`, and here in this PR I did as directed by the top comment in the former file: https://github.com/scipy/scipy/pull/22541/files#diff-9ba32e29b4b0a2ef5581fa10c8a8a79f01605f541f722957c098fc05a4addb22R2

@melissawm could you advise?

melissawm:
Hi all - linkify-it-py helps create hyperlinks automatically when writing syntax like "https://example.com" in markdown files. This was added as part of https://github.com/scipy/scipy/pull/22385 to match what rst already does. If you feel this is not appropriate, feel free to remove. It will not break existing docs builds, it will only disable those links.

ev-br:
Okay, so the addition to pyproject.toml is just maintenance within the status quo; if we want to remove the dependency, we can do it separately.
One other unrelated doc change is the Sphinx version pin to unblock the problem which I believe also is also on main. Am happy to submit it as a separate PR, if desired.

Other than that, the PR is ready from my side.

PR #22549, opened 2025-02-18
ENH: use `workers` keyword in `optimize._differentiable_functions.VectorFunction`

andyfaff:
This PR adds the `workers` keyword to `VectorFunction`. The workers aid parallel evaluation of the jacobian when using finite differences. The keyword has no effect on the finite difference estimation of the hessian as this is carried out with a `LinearOperator` that only uses a small number of function evaluations.

Parallelising `VectorFunction` is a precursor step to using it in `optimize.least_squares` where its memoisation will reduce the number of function evaluations, as well as allowing the parallelisation of the `lm` method.

### DIFF
--- a/scipy/optimize/_differentiable_functions.py
+++ b/scipy/optimize/_differentiable_functions.py
@@ -481,7 +481,10 @@ class VectorFunction:
             finite_diff_options["method"] = hess
             finite_diff_options["rel_step"] = finite_diff_rel_step
             finite_diff_options["as_linear_operator"] = True
-
+            # workers is not useful for evaluation of the LinearOperator
+            # produced by approx_derivative. Only two/three function
+            # evaluations are used, and the LinearOperator may persist
+            # outside the scope that workers is valid in.
             self.x_diff = np.copy(self.x)
         if jac in FD_METHODS and hess in FD_METHODS:
             raise ValueError("Whenever the Jacobian is estimated via "
@@ -602,9 +605,9 @@ class VectorFunction:
             def update_hess():
                 self._update_jac()
                 self.H = approx_derivative(jac_dot_v, self.x,
-                                          f0=self.J.T.dot(self.v),
-                                          args=(self.v,),
-                                          **finite_diff_options)
+                                           f0=self.J.T.dot(self.v),
+                                           args=(self.v,),
+                                           **finite_diff_options)
 
             update_hess()
             self.H_updated = True
--- a/scipy/optimize/_differentiable_functions.py
+++ b/scipy/optimize/_differentiable_functions.py
@@ -12,7 +12,7 @@ from scipy._lib._util import _ScalarFunctionWrapper
 FD_METHODS = ('2-point', '3-point', 'cs')
 
 
-class _GradWrapper:
+class _ScalarGradWrapper:
     """
     Wrapper class for gradient calculation
     """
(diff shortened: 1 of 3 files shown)


scipy/optimize/_differentiable_functions.py
```diff
@@ -587,10 +603,11 @@ def jac_dot_v(x, v):
 
             def update_hess():
                 self._update_jac()
-                self.H = approx_derivative(jac_dot_v, self.x,
-                                           f0=self.J.T.dot(self.v),
-                                           args=(self.v,),
-                                           **finite_diff_options)
+                self.H, dct = approx_derivative(jac_dot_v, self.x,
```
andyfaff, line 606: `jac_dot_v` doesn't need to be pickleable because `finite_diff_options['as_linear_operator'] is True`. 

The creation of the `LinearOperator` by `approx_derivative` doesn't consume any function evaluations, they are only consumed when the `LinearOperator` is used. Each call of the `LinearOperator` will use two or three evaluations, which didn't seem worth parallelising to me. If it's not parallelised, then `jac_dot_v` doesn't need to be pickleable.

scipy/optimize/_numdiff.py
```diff
@@ -555,8 +558,8 @@ def approx_derivative(fun, x0, method='3-point', rel_step=None, abs_step=None,
         if rel_step is None:
             rel_step = _eps_for_method(x0.dtype, f0.dtype, method)
 
-        J, _nfev = _linear_operator_difference(fun_wrapped, x0,
-                                               f0, rel_step, method)
+        J, _ = _linear_operator_difference(fun_wrapped, x0,
+                                           f0, rel_step, method)
```
andyfaff, line 562: We don't use `workers` with `as_linear_operator` because:

- the `LinearOperator` may still be used outside the scope in which `workers` is being used.
- there are only two or three function evaluations, which isn't a high priority for being parallelised.

scipy/optimize/_differentiable_functions.py
```diff
@@ -523,34 +536,38 @@ def update_jac():
...
                 self.sparse_jacobian = True
 
             elif sps.issparse(self.J):
                 def update_jac():
                     self._update_fun()
-                    self.J = approx_derivative(fun_wrapped, self.x, f0=self.f,
-                                               **finite_diff_options).toarray()
+                    self.J, dct = approx_derivative(fun_wrapped, self.x, f0=self.f,
+                                                    **finite_diff_options).toarray()
+                    self.nfev += dct['nfev']
```
dschmitz89, line 560: Could you elaborate a bit on this? I might not understand fully but was `nfev` actually counted correctly before this PR without parallelization?

andyfaff: It was counted correctly. The main change is that previously there were wrappers of the user's objective function that tracked the number of calls by storing it in an attribute. These wrappers (converted to classes) can now be evaluated in parallel in different processes in `approx_derivative`. This parallel evaluation means that the attribute can't be updated properly. Therefore `approx_derivative` calculates what the number of function evaluations would've been and returns it in the dictionary.

andyfaff: The code calling `approx_derivative` then has to keep a total tally of func evaluations. The tests for total number of function evaluations are quite strict for `_differentiable_functions`.

dschmitz89 (approved):

dschmitz89:
Thanks @andyfaff .

PR #22551, opened 2025-02-18
MAINT: optimize: remove unnecessary `not_first_iteration` boolean variable in `scalar_search_wolfe2`

### LINKED ISSUE #22550: unnecessary `not_first_iteration` boolean variable
The following check and the associated variable `not_first_iteration` are unnecessary:
https://github.com/scipy/scipy/blob/4732f799d0019d3cd763fc10bb75dd0b62268f78/scipy/optimize/_linesearch.py#L435

In the original algorithm by Wright and Nocedal (Numerical Optimization, 1999, pp. 59–61), this check ensures that the corresponding function value exists. However, in the current implementation https://github.com/scipy/scipy/blob/4732f799d0019d3cd763fc10bb75dd0b62268f78/scipy/optimize/_linesearch.py#L408
it is set with the default value equal to `phi0`. Furthermore, since `derphi0` represents a descent direction (`derphi0 < 0`), on the first iteration, if the condition (`phi_a1 >= phi_a0`) is true that would imply that the previous Armijo condition would've been satisfied first. Hence, the check for `not_first_iteration` is redundant and can be removed.

serhii-havrylov:
remove unnecessary `not_first_iteration` boolean variable in scalar_search_wolfe2

#### Reference issue
https://github.com/scipy/scipy/issues/22550

fixes https://github.com/scipy/scipy/issues/22550

### DIFF
--- a/scipy/optimize/_linesearch.py
+++ b/scipy/optimize/_linesearch.py
@@ -390,10 +390,10 @@ def scalar_search_wolfe2(phi, derphi, phi0=None,
     if derphi0 is None:
         derphi0 = derphi(0.)
     if derphi0 == 0.0:
-       alpha_star = 0.0
-       phi_star = phi0
-       derphi_star = derphi0
-       return alpha_star, phi_star, phi0, derphi_star
+        alpha_star = 0.0
+        phi_star = phi0
+        derphi_star = derphi0
+        return alpha_star, phi_star, phi0, derphi_star
 
     alpha0 = 0
     if old_phi0 is not None and derphi0 != 0:
--- a/scipy/optimize/_linesearch.py
+++ b/scipy/optimize/_linesearch.py
@@ -450,7 +450,7 @@ def scalar_search_wolfe2(phi, derphi, phi0=None,
                 derphi_star = derphi_a1
                 break
 
-        if (derphi_a1 >= 0):
+        if derphi_a1 > 0:
             alpha_star, phi_star, derphi_star = \
                         _zoom(alpha1, alpha0, phi_a1,
                               phi_a0, derphi_a1, phi, derphi,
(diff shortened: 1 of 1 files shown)


j-bowhay:
CI failures look related please could you take a look

serhii-havrylov:
I've just checked the Newton-CG test:
https://github.com/scipy/scipy/blob/f7da7b74bcf297aa812e14f284b49e6be21e3490/scipy/optimize/tests/test_optimize.py#L549
The problem is that Newton-CG violates the assumption of the line-search algorithm:
https://github.com/scipy/scipy/blob/f7da7b74bcf297aa812e14f284b49e6be21e3490/scipy/optimize/_optimize.py#L2172
Namely, it is supposed to provide a descent direction. However, in this case, it provides a zero vector, so the Fortran procedure correctly fails to perform the line search:
https://github.com/scipy/scipy/blob/f7da7b74bcf297aa812e14f284b49e6be21e3490/scipy/optimize/_optimize.py#L1185
and the python fallback is called. However, the assumption is still violated and now derphi0 is zero instead of <0 as I previously assumed.
Previously, the first check  failed  due to the presence of `not_first_iteration` variable, and some step length was returned.
https://github.com/scipy/scipy/blob/f7da7b74bcf297aa812e14f284b49e6be21e3490/scipy/optimize/_linesearch.py#L434
But now due to the assumptions being violated it wrongly called _zoom function which obviously can't converge, thus producing the current error.
With the direction vector being zero, any step length would work. So, the previous behaviuor was rather a happy coincidence that it worked at all, since no meaningful step size can be found if the derivative of the line search scalar function is zero. 

There are several ways to fix this, such as adding an additional check for derphi0 == 0 and returning a zero step size. However, this feels like a hack, since no one should call the line search algorithm with a zero direction vector in the first place.

Can anyone please advise on how to proceed?

PR #22553, opened 2025-02-19
ENH: sparse: add `__binsparse__` protocol

hameerabbasi:
#### Reference issue
https://github.com/data-apis/array-api/issues/840

Implements the `__binsparse__` protocol as described in the above issue. For context, this is the equivalent of `__dlpack__` for sparse arrays

#### Additional information
~~Binsparse currently doesn't support the structure-of-array format for COO; hence COO support is missing for the moment.~~

It doesn't have array-of-structures support for COO. A discussion was carried out, and support for that is coming. This turns out to be important for PyTorch and PyData/Sparse.

### DIFF
--- a/scipy/sparse/_compressed.py
+++ b/scipy/sparse/_compressed.py
@@ -1464,6 +1464,42 @@ class _cs_matrix(_data_matrix, _minmax_mixin, IndexMixin):
             indices = np.tile(np.arange(N), len(self.data))
             indptr = self.indptr * N
         return self.__class__((data, indices, indptr), shape=shape, copy=False)
+    
+    def __binsparse_descriptor__(self) -> dict:
+        """Return a `dict` equivalent to a parsed JSON [`binsparse` descriptor](https://graphblas.org/binsparse-specification/#descriptor)
+        of this array.
+
+        Returns
+        -------
+        dict
+            Parsed `binsparse` descriptor.
+        """
+        from scipy import __version__
+
+        data_dt = str(self.data.dtype)
+        if np.issubdtype(data_dt, np.complexfloating):
+            data_dt = f"complex[float{self.data.dtype.itemsize * 4}]"
+        return {
+            "binsparse": {
+                "version": "0.1",
+                "format": self.format.upper(),
+                "shape": list(self.shape),
+                "number_of_stored_values": self.nnz,
+                "data_types": {
+                    "pointers_to_1": str(self.indices.dtype),
+                    "indices_1": str(self.indptr.dtype),
+                    "values": data_dt,
+                },
+            },
+            "original_source": f"`scipy.sparse`, version {__version__}",
+        }
+
+    def __binsparse__(self) -> dict[str, np.ndarray]:
+        return {
+            "pointers_to_1": self.indptr,
+            "indices_1": self.indices,
+            "values": self.data,
+        }
 
 
 def _make_diagonal_csr(data, is_array=False):
--- a/scipy/sparse/_matrix_io.py
+++ b/scipy/sparse/_matrix_io.py
@@ -1,7 +1,7 @@
 import numpy as np
 import scipy as sp
 
-__all__ = ['save_npz', 'load_npz']
+__all__ = ['save_npz', 'load_npz', 'from_binsparse']
 
 
 # Make loading safe vs. malicious input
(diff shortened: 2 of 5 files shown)


lucascolley:
cc @ilan-gold

scipy/sparse/_compressed.py
```diff
@@ -1464,6 +1464,42 @@ def _broadcast_to(self, shape, copy=False):
...
+                "format": self.format.upper(),
+                "shape": list(self.shape),
+                "number_of_stored_values": self.nnz,
+                "data_types": {
+                    "pointers_to_1": str(self.indices.dtype),
+                    "indices_1": str(self.indptr.dtype),
+                    "values": data_dt,
+                },
+            },
+            "original_source": f"`scipy.sparse`, version {__version__}",
```
rgommers, line 1494: This is a bit questionable, it makes more sense for a stored file on disk rather than an in-memory format. If it's left out then you'd not have to do the `desc = desc["binsparse"]` unpacking.

scipy/sparse/_compressed.py
```diff
@@ -1464,6 +1464,42 @@ def _broadcast_to(self, shape, copy=False):
...
+        from scipy import __version__
+
+        data_dt = str(self.data.dtype)
+        if np.issubdtype(data_dt, np.complexfloating):
+            data_dt = f"complex[float{self.data.dtype.itemsize * 4}]"
+        return {
+            "binsparse": {
+                "version": "0.1",
+                "format": self.format.upper(),
+                "shape": list(self.shape),
```
rgommers, line 1486: Tuple would be better here.

scipy/sparse/_matrix_io.py
```diff
@@ -165,3 +165,111 @@ def load_npz(file):
...
+
+    desc = desc["binsparse"]
+    version_tuple: tuple[int, ...] = tuple(int(v) for v in desc["version"].split("."))
+    if version_tuple != (0, 1):
+        raise RuntimeError("Unsupported `__binsparse__` protocol version.")
+
+    format = desc["format"]
+    format_err_str = f"Unsupported format: `{format!r}`."
+
+    if isinstance(format, str):
```
rgommers, line 182: Is this check needed? I think `format` has to always be a string? Things go wrong here if it isn't.

rgommers (commented):
Thanks @hameerabbasi. I realize that this PR is still draft so this may be coming, but a few questions:

- Has the discussion on the in-memory format upstream in `binsparse` stabilized? And where will the specification live?
- Is there a PR with an implementation for PyData Sparse, or any other library? 
- If so, does it make sense to test against PyData Sparse rather than only roundtrip tests?

scipy/sparse/_compressed.py
hameerabbasi: I was following the structure in the [specification](https://graphblas.org/binsparse-specification/#descriptor) as closely as possible. Of course, one could change https://github.com/data-apis/array-api/issues/840 to say that we only return the `binsparse` key of the descriptor, and then this (and the unpacking) would go away.

scipy/sparse/_compressed.py
hameerabbasi: JSON doesn't have tuples, and this, being a parsed JSON descriptor, would use a list. One can change https://github.com/data-apis/array-api/issues/840 and do some word-gymnastics saying "shapes always have to be tuples", but I'd rather leave this as a list to be 100% compliant.

hameerabbasi:
> Thanks @hameerabbasi. I realize that this PR is still draft so this may be coming, but a few questions:

Thanks for the review, @rgommers!

> * Has the discussion on the in-memory format upstream in `binsparse` stabilized? And where will the specification live?

Binsparse doesn't specify a storage back-end at the moment, and is fairly restrictive about what primitives such a back-end would need to have. This is an advantage for us because we can create an in-memory storage back-end without any changes to the spec itself.

Of course, features are coming (such as blocked-format support and AoS COO support) but support for those can be deferred, as the spec is versioned.

The spec lives at [this repo](https://github.com/GraphBLAS/binsparse-specification), with the rendered version living [here](https://graphblas.org/binsparse-specification/).

> * Is there a PR with an implementation for PyData Sparse, or any other library?

There is https://github.com/pydata/sparse/pull/764, but that has to be merged pending the final discussions here and elsewhere.

> * If so, does it make sense to test against PyData Sparse rather than only roundtrip tests?

It definitely does! However, I'd prefer to do a follow-up PR to both libraries once the initial support is merged.

rgommers:
How about hashing out with the Binsparse team on https://github.com/GraphBLAS/binsparse-specification/issues/4 whether the authoritative spec will live in the array API standard or in Binsparse itself? If the former (perhaps a little more authoritative to the libraries that are expected to implement these methods), then let's agree on signatures in https://github.com/data-apis/array-api/issues/840. That doesn't have to be merged, but it should be a central place to look. Right now the draft PRs to SciPy, PyData Sparse and CuPy are unconnected.

hameerabbasi:
> How about hashing out with the Binsparse team on [GraphBLAS/binsparse-specification#4](https://github.com/GraphBLAS/binsparse-specification/issues/4) whether the authoritative spec will live in the array API standard or in Binsparse itself?

Done in https://github.com/GraphBLAS/binsparse-specification/issues/4#issuecomment-2677815815

PR #22554, opened 2025-02-19
BUG: `lfiltic`'s handling of `a[0] != 1` differs from `lfilter`'s 

### LINKED ISSUE #22470: BUG: `lfiltic`'s handling of `a[0] != 1` differs from `lfilter`'s
### Describe your issue.

`lfiltic` is supposed to construct initial conditions for `lfilter` given input and output vectors.
It works fine when `a[0] == 1` but the behaviour is unexpected (buggy?) when this is not the case.
Unlike the documentation for `lfilter` (which clearly states how `a[0] != 1` is handled), nothing is said about `lfiltic`. However, one could reasonably assume it to be the same, i.e. that the following MRE gives the same result in both cases.

Running the MRE below gives no error message but the state `zi2` computed by `lfiltic` (using `lfilter`'s output) differs from the state `zi1` returned by `lfilter`. Consequently, if one uses the state computed by `lfiltic` as argument `zi` to a subsequent call to `lfilter` with the same `a` and `b`, the result is wrong.
The fix is pretty straightforward: it suffices to divide the resulting state by `a[0]`, i.e. replacing [line 255 of _signaltools.py](https://github.com/scipy/scipy/blob/df134eab5a500c2146ed4552c8674a78d8154ee9/scipy/signal/_signaltools.py#L2255) with:
`return zi / a[0]`
This new behaviour would also be consistent with Matlab's and [Octave's `filtic`](https://sourceforge.net/p/octave/signal/ci/default/tree/inst/filtic.m).

### Reproducing Code Example

```python
import numpy as np
from scipy.signal import lfilter, lfiltic

a = [2., 1., 3.]
b = [.5, 1., .2]

x = np.sin(np.arange(10))

y1, zi1 = lfilter(b, a, x, zi=[0., 0.])
zi2 = lfiltic(b, a, y1[::-1], x[::-1])

print(zi1, zi2, zi2 / a[0], sep='\n')
```

### Error message

(No error message, the output is as follows)
```shell
[ 0.37972536 -1.67552273]
[ 0.75945072 -3.35104546]
[ 0.37972536 -1.67552273]
```

```shell
scipy.__version__ = 1.12.0
numpy.__version__ = 1.26.4
```

lciti:
#### Reference issue
Closes #22470.

Fixes 22470 by dividing the resulting `zi` by `a[0]`.
Also added checks that `a` is non-empty and its first element is non-zero.
Added tests to test the issue before and after the fix (also added additional test cases for `lfiltic`).

#### Additional information

None

### DIFF
--- a/scipy/signal/_signaltools.py
+++ b/scipy/signal/_signaltools.py
@@ -2359,11 +2359,14 @@ def lfiltic(b, a, y, x=None):
         if x is not None:
             _reject_objects(x, 'lfiltic')
 
-    a = xp.asarray(a)
-    b = xp.asarray(b)
-
-    N = xp_size(a) - 1
-    M = xp_size(b) - 1
+    a = xpx.atleast_nd(xp.asarray(a), ndim=1, xp=xp)
+    b = xpx.atleast_nd(xp.asarray(b), ndim=1, xp=xp)
+    if a.ndim > 1:
+        raise ValueError('Filter coefficients `a` must be 1-D.')
+    if b.ndim > 1:
+        raise ValueError('Filter coefficients `b` must be 1-D.')
+    N = a.shape[0] - 1
+    M = b.shape[0] - 1
     K = max(M, N)
     y = xp.asarray(y)
 
@@ -2402,11 +2405,10 @@ def lfiltic(b, a, y, x=None):
     for m in range(N):
         zi[m] -= xp.sum(a[m + 1:] * y[:N - m], axis=0)
 
-    a0 = a[0] if N > 0 else a
-    if a0 != 1.:
-        if a0 == 0.:
+    if a[0] != 1.:
+        if a[0] == 0.:
             raise ValueError("First `a` filter coefficient must be non-zero.")
-        zi /= a0
+        zi /= a[0]
 
     return zi
 
--- a/scipy/signal/_signaltools.py
+++ b/scipy/signal/_signaltools.py
@@ -2386,11 +2386,9 @@ def lfiltic(b, a, y, x=None):
             result_type = xp.float64
         x = xp.astype(x, result_type)
 
-        concat = array_namespace(a).concat
-
         L = xp_size(x)
         if L < M:
-            x = concat((x, xp.zeros(M - L)))
+            x = xp.concat((x, xp.zeros(M - L)))
 
     y = xp.astype(y, result_type)
     zi = xp.zeros(K, dtype=result_type)
--- a/scipy/signal/_signaltools.py
+++ b/scipy/signal/_signaltools.py
@@ -2367,6 +2367,9 @@ def lfiltic(b, a, y, x=None):
     K = max(M, N)
     y = xp.asarray(y)
 
+    if N < 0:
+        raise ValueError("There must be at least one `a` coefficient.")
+
     if x is None:
         result_type = xp.result_type(b, a, y)
         if xp.isdtype(result_type, ('bool', 'integral')):  #'bui':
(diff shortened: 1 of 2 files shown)


scipy/signal/_signaltools.py
```diff
@@ -2367,6 +2367,9 @@ def lfiltic(b, a, y, x=None):
     K = max(M, N)
     y = xp.asarray(y)
 
+    if N < 0:
+        raise ValueError("There must be at least one `a` coefficient.")
```
j-bowhay, line 2371: Please can you add test coverage for this

scipy/signal/_signaltools.py
```diff
@@ -2391,14 +2394,20 @@ def lfiltic(b, a, y, x=None):
...
     for m in range(M):
         zi[m] = xp.sum(b[m + 1:] * x[:M - m], axis=0)
 
     for m in range(N):
         zi[m] -= xp.sum(a[m + 1:] * y[:N - m], axis=0)
 
+    a0 = a[0] if N > 0 else a
+    if a0 != 1.:
+        if a0 == 0.:
+            raise ValueError("First `a` filter coefficient must be non-zero.")
```
j-bowhay, line 2408: And this

lciti:
@ev-br I am getting some errors in `lfiltic` (see [here](https://github.com/scipy/scipy/actions/runs/13415830954/job/37476370519#step:13:3604)) around the "concat = array_namespace(a).concat" line. I was wondering if you could help me understand why that line is needed in the first place. Can't we just use "x = **xp.**concat((x, xp.zeros(M - L)))" as it's done a couple of lines below for "y"?

ev-br:
> I am getting some errors in lfiltic (see here) around the "concat = array_namespace(a).concat" line. I was wondering if you could help me understand why that line is needed in the first place.

This pattern is needed when `xp` is not fully array API compatible. For instance, `numpy.concat` is new in numpy 2.0 So to run with numpy == 1.x, you go through `array_api_compat.numpy` which wraps numpy proper and adds Array API compatible functions. 

You can add these kinds of wrappers in a reactive way: if the CI is OK on numpy 2.0 and fails on 1.x, you just add this stanza to where it fails. 

The errors on CI though are different though:

```
 a          = array([1.0, 0.5, 1.5], dtype=object)
```

is wrong. Object arrays should not be created, so your best bet is to trace where this comes from and fix it to be a normal numerical array.

lciti:
@ev-br Thank you for your reply. The "object" type is due to the way the test is already set up (it tries different data types, including objects) and has nothing to do with my edit. I have set that line to "xp.concat" for now since this is similar to a change made to "y" after your initial commit. That test seems to work now.

lciti:
@j-bowhay I have taken your comments onboard and I have done a few minor changes to address the CI issues with non-numpy array APIs. It seems to pass all the tests now.

lciti:
Hi, just a friendly ping on this PR. I've addressed all comments and would appreciate it if you could let me know if there's anything else needed or if it's ready to be merged. Thanks!

j-bowhay:
Thanks for making the changes, it would be good if one of the signal regulars could take a look

larsoner (approved):
I am not knowledgeable about the `xp`-related changes but the signal changes look good to me!

j-bowhay (approved):
Looks good from an array api perspective and there is approval from a signal regular, thanks @lciti, keep them coming!

tylerjereddy:
I considered for backporting but it seemed too messy according to my note in gh-22840, so I'll remove the label.

Ping me if there is a concern about that, but I need to stop dragging my feet on the `1.15.3` release so getting organized for that.

PR #22556, opened 2025-02-19
ENH: optimize: speed up `LbfgsInvHessProduct.todense` on large matrices

### LINKED ISSUE #18105: ENH: optimize `LbfgsInvHessProduct.todense()`, 10x speed improvement for size 1000

Using `optimize.minimize(..., method='l-bfgs-b').inv_hess.todense()` ([code](https://github.com/scipy/scipy/blob/c1ed5ece8ffbf05356a22a8106affcd11bd3aee0/scipy/optimize/_lbfgsb_py.py#L474-L494)) I noticed its implementation does $O(n^3)$ operations when it could be $O(n^2)$. I made a few tries at optimizing it. The methods I considered are:

1) modifying `LbfgsInvHessProduct.todense` to avoid matrix by matrix multiplications
2) implementing `LbfgsInvHessProduct._matmat` and doing `obj.matmat(eye(*obj.shape))`
3) using the stock implementation of `LbfgsInvHessProduct._matmat`

Code and benchmark:

```python
from scipy import optimize
import numpy as np

sk, yk = np.random.randn(2, 10, 1000)
invh = optimize.LbfgsInvHessProduct(sk, yk)

def todense(self):
    """ Rewrite LbfgsInvHessProduct.todense avoiding matrix @ matrix """
    s, y, n_corrs, rho = self.sk, self.yk, self.n_corrs, self.rho
    Hk = np.eye(*self.shape, dtype=self.dtype)

    for i in range(n_corrs):
        # H  <--  (I - rho sy') H (I - rho ys') + rho ss' =
        #       = H - rho Hys' - rho sy'H + rho^2 sy'Hys' + rho ss' =
        #       = H - rho Hys' - rho sy'H + rho(1 + rho y'Hy) ss'
        rhoHys = -rho[i] * np.outer(Hk @ y[i], s[i])
        Hk += rho[i] * (1 + rho[i] * (y[i] @ Hk @ y[i])) * np.outer(s[i], s[i])
        Hk += rhoHys
        Hk += rhoHys.T

    return Hk

def _matmat(self, X):
    """ Rewrite LbfgsInvHessProduct._matvec broadcasting on columns of X """
    s, y, n_corrs, rho = self.sk, self.yk, self.n_corrs, self.rho
    Q = np.array(X, dtype=self.dtype, copy=True)
    assert Q.ndim == 2

    alpha = np.empty((n_corrs, Q.shape[1]))

    for i in range(n_corrs-1, -1, -1):
        alpha[i] = rho[i] * np.dot(s[i], Q)
        Q = Q - alpha[i]*y[i][:, np.newaxis]

    R = Q
    for i in range(n_corrs):
        beta = rho[i] * np.dot(y[i], R)
        R = R + s[i][:, np.newaxis] * (alpha[i] - beta)

    return R

inv1 = invh.todense()
inv2 = todense(invh)
inv3 = _matmat(invh, np.eye(*invh.shape))
np.testing.assert_allclose(inv1, inv2, atol=0, rtol=1e-5)
np.testing.assert_allclose(inv1, inv3, atol=0, rtol=1e-5)

%timeit -r1 invh.todense()
%timeit -r1 todense(invh)
%timeit -r1 _matmat(invh, np.eye(*invh.shape))
%timeit -r1 invh.matmat(np.eye(*invh.shape))
```

Benchmark result with n = 1000:

```
583 ms ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
109 ms ± 0 ns per loop (mean ± std. dev. of 1 run, 10 loops each)
48.4 ms ± 0 ns per loop (mean ± std. dev. of 1 run, 10 loops each)
52.5 ms ± 0 ns per loop (mean ± std. dev. of 1 run, 10 loops each)
```

With n = 10:

```
76.5 µs ± 0 ns per loop (mean ± std. dev. of 1 run, 10,000 loops each)
77.2 µs ± 0 ns per loop (mean ± std. dev. of 1 run, 10,000 loops each)
56.1 µs ± 0 ns per loop (mean ± std. dev. of 1 run, 10,000 loops each)
316 µs ± 0 ns per loop (mean ± std. dev. of 1 run, 1,000 loops each)
```

With n = 10000:

```
3min 10s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
8.08 s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
3.99 s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
3.82 s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
```

Conclusions: the direct rewrite of `todense` is slower than using the matvec operation, and the existing matmat is slower than the reimplementation for small sizes because it involves a python loop ([code](https://github.com/scipy/scipy/blob/c1ed5ece8ffbf05356a22a8106affcd11bd3aee0/scipy/sparse/linalg/_interface.py#L180-L187)).

I can do a PR for this if interested.

_No response_

_No response_

_No response_

Gattocrucco:
#### Reference issue
Closes gh-18105

An `LbfgsInvHessProduct` (subclass of `LinearOperator`) object represents the running inverse Hessian estimate used by LBFGS. The method `todense()` converts it to the numpy array representation of the matrix (internally, no actual matrix is stored). I re-implemented `todense()` to make it faster.

#### Additional information
The current implementation has computational complexity $O(n^3)$, where the matrix is $n\times n$, while the new version is $O(n^2)$, so it's much faster on large matrices. I also took care to check it's efficient on very small matrices.

### DIFF
--- a/scipy/optimize/_lbfgsb_py.py
+++ b/scipy/optimize/_lbfgsb_py.py
@@ -586,18 +586,17 @@ class LbfgsInvHessProduct(LinearOperator):
         """
         s, y, n_corrs, rho = self.sk, self.yk, self.n_corrs, self.rho
         Q = np.array(X, dtype=self.dtype, copy=True)
-        assert Q.ndim == 2
 
         alpha = np.empty((n_corrs, Q.shape[1]))
 
         for i in range(n_corrs-1, -1, -1):
             alpha[i] = rho[i] * np.dot(s[i], Q)
-            Q = Q - alpha[i]*y[i][:, np.newaxis]
+            Q -= alpha[i]*y[i][:, np.newaxis]
 
         R = Q
         for i in range(n_corrs):
             beta = rho[i] * np.dot(y[i], R)
-            R = R + s[i][:, np.newaxis] * (alpha[i] - beta)
+            R += s[i][:, np.newaxis] * (alpha[i] - beta)
 
         return R
 
--- a/scipy/optimize/_lbfgsb_py.py
+++ b/scipy/optimize/_lbfgsb_py.py
@@ -563,6 +563,44 @@ class LbfgsInvHessProduct(LinearOperator):
 
         return r
 
+    def _matmat(self, X):
+        """Efficient matrix-matrix multiply with the BFGS matrices.
+
+        This calculation is described in Section (4) of [1].
+
+        Parameters
+        ----------
+        X : ndarray
+            An array with shape (n,m)
+
+        Returns
+        -------
+        Y : ndarray
+            The matrix-matrix product
+
+        Notes
+        -----
+        This implementation is written starting from _matvec and broadcasting
+        all expressions along the second axis of X.
+
+        """
+        s, y, n_corrs, rho = self.sk, self.yk, self.n_corrs, self.rho
+        Q = np.array(X, dtype=self.dtype, copy=True)
+        assert Q.ndim == 2
+
+        alpha = np.empty((n_corrs, Q.shape[1]))
+
+        for i in range(n_corrs-1, -1, -1):
+            alpha[i] = rho[i] * np.dot(s[i], Q)
+            Q = Q - alpha[i]*y[i][:, np.newaxis]
+
+        R = Q
+        for i in range(n_corrs):
+            beta = rho[i] * np.dot(y[i], R)
+            R = R + s[i][:, np.newaxis] * (alpha[i] - beta)
+
+        return R
+
     def todense(self):
         """Return a dense array representation of this operator.
 
@@ -573,14 +611,5 @@ class LbfgsInvHessProduct(LinearOperator):
             the same data represented by this `LinearOperator`.
 
         """
-        s, y, n_corrs, rho = self.sk, self.yk, self.n_corrs, self.rho
         I_arr = np.eye(*self.shape, dtype=self.dtype)
-        Hk = I_arr
-
-        for i in range(n_corrs):
-            A1 = I_arr - s[i][:, np.newaxis] * y[i][np.newaxis, :] * rho[i]
-            A2 = I_arr - y[i][:, np.newaxis] * s[i][np.newaxis, :] * rho[i]
-
-            Hk = np.dot(A1, np.dot(Hk, A2)) + (rho[i] * s[i][:, np.newaxis] *
-                                                        s[i][np.newaxis, :])
-        return Hk
+        return self._matmat(I_arr)
--- a/scipy/optimize/tests/test_lbfgsb_hessinv.py
+++ b/scipy/optimize/tests/test_lbfgsb_hessinv.py
@@ -41,3 +41,25 @@ def test_2():
     assert_allclose(H1, result2.hess_inv, rtol=1e-2, atol=0.03)
 
 
+def test_3():
+
+    def todense_old_impl(self):
+        s, y, n_corrs, rho = self.sk, self.yk, self.n_corrs, self.rho
+        I_arr = np.eye(*self.shape, dtype=self.dtype)
+        Hk = I_arr
+
+        for i in range(n_corrs):
+            A1 = I_arr - s[i][:, np.newaxis] * y[i][np.newaxis, :] * rho[i]
+            A2 = I_arr - y[i][:, np.newaxis] * s[i][np.newaxis, :] * rho[i]
+
+            Hk = np.dot(A1, np.dot(Hk, A2)) + (rho[i] * s[i][:, np.newaxis] *
+                                                        s[i][np.newaxis, :])
+        return Hk
+
+    H0 = [[3, 0], [1, 2]]
+
+    def f(x):
+        return np.dot(x, np.dot(scipy.linalg.inv(H0), x))
+
+    result1 = minimize(fun=f, method='L-BFGS-B', x0=[10, 20])
+    assert_allclose(result1.hess_inv.todense(), todense_old_impl(result1.hess_inv))


j-bowhay:
It would be good to add a test that the new method produces the same result as the old one

Gattocrucco:
Do you mean an identical result, or just close?

Case 1: I'm pretty sure it won't be identical, that would break

Case 2: The current tests check `.todense()` against the matrix-vector product routine, but to be extra-sure yeah I can add a test that hardcodes the original implementation

j-bowhay:
> Do you mean an identical result, or just close?
> 
> Case 1: I'm pretty sure it won't be identical, that would break
> 
> Case 2: The current tests check `.todense()` against the matrix-vector product routine, but to be extra-sure yeah I can add a test that hardcodes the original implementation

Ok I will double check but sounds like the existing tests are sufficient

dschmitz89:
Thanks for the PR @Gattocrucco . Test failures are unrelated.

Could you provide a simple illustrative example or point to one of the benchmarks so that we can evaluate the performance improvement? Something like a high dimensional Rosenbrock function might suffice.

Gattocrucco:
You can see a benchmark of the individual method in gh-18105. Summary: for small-medium sizes it's a bit faster; for large sizes it's day and night. However IIUC the minimizer does not ever call that method, so it should have no effect on the minimization time.

As soon as I have some time, I will try to cook up a more realistic usage example. Back when I opened that issue, I used it to estimate a precision matrix.

scipy/optimize/_lbfgsb_py.py
```diff
@@ -563,6 +563,44 @@ def _matvec(self, x):
...
+
+        Notes
+        -----
+        This implementation is written starting from _matvec and broadcasting
+        all expressions along the second axis of X.
+
+        """
+        s, y, n_corrs, rho = self.sk, self.yk, self.n_corrs, self.rho
+        Q = np.array(X, dtype=self.dtype, copy=True)
+        assert Q.ndim == 2
```
j-bowhay, line 589: `assert` can be disabled by running python in optimized mode. If this check is important please you rewrite it as an of statement + exception

scipy/optimize/_lbfgsb_py.py
```diff
@@ -563,6 +563,44 @@ def _matvec(self, x):
...
+
+        alpha = np.empty((n_corrs, Q.shape[1]))
+
+        for i in range(n_corrs-1, -1, -1):
+            alpha[i] = rho[i] * np.dot(s[i], Q)
```
j-bowhay, line 594:
```suggestion
            alpha[i] = rho[i] * s[i] @ Q
```

scipy/optimize/_lbfgsb_py.py
```diff
@@ -563,6 +563,44 @@ def _matvec(self, x):
...
+            Q = Q - alpha[i]*y[i][:, np.newaxis]
```
j-bowhay, line 595:
```suggestion
            Q -= alpha[i]*y[i][:, np.newaxis]
```

scipy/optimize/_lbfgsb_py.py
```diff
@@ -563,6 +563,44 @@ def _matvec(self, x):
...
+
+        R = Q
```
j-bowhay, line 597: Is this extra assignment required?

scipy/optimize/_lbfgsb_py.py
```diff
@@ -563,6 +563,44 @@ def _matvec(self, x):
...
+        for i in range(n_corrs):
+            beta = rho[i] * np.dot(y[i], R)
```
j-bowhay, line 599:
```suggestion
            beta = rho[i] * y[i] @ R
```

scipy/optimize/_lbfgsb_py.py
```diff
@@ -563,6 +563,44 @@ def _matvec(self, x):
...
+            R = R + s[i][:, np.newaxis] * (alpha[i] - beta)
```
j-bowhay, line 600:
```suggestion
            R += s[i][:, np.newaxis] * (alpha[i] - beta)
```

scipy/optimize/_lbfgsb_py.py
j-bowhay: Did the code previously work for `X.dim != 2`? It's important we don't break backwards compatibility

j-bowhay (commented):
Thanks @Gattocrucco some initial comments, I have yet to check the reference

scipy/optimize/_lbfgsb_py.py
Gattocrucco: The `_matmat` method is not meant to be called directly, it is invoked by `LinearOperator.(r)matmat`, which raises if the array is not 2d. I checked that on the main branch indeed calling `LbfgsInvHessProduct.(r)matmat` with anything other than a 2d array triggers an exception in the `LinearOperator` implementation.

The assert is for code readability, so I think it is fine to leave it as an assert.

scipy/optimize/_lbfgsb_py.py
Gattocrucco: This assignment has the only effect of renaming `Q` to `R`. It's in the original implementation I'm following (`_matvec`, just above my code). I guess it follows the notation of the source where the algorithm was copied from. I think it's better to leave it there.

scipy/optimize/_lbfgsb_py.py
Gattocrucco: This potentially changes the dtype of `Q`: inplace subtraction keeps the initial type, while standard subtraction does type promotion. Since `LbfgsInvHessProduct` unconditionally self-initializes with dtype `float64`, which is inherited by `Q`, this should never matter, and I'm in favor of switching to inplace assignment; however since you show a conservative attitude I'd like to ear your opinion.

scipy/optimize/_lbfgsb_py.py
Gattocrucco: See other comment above.

scipy/optimize/_lbfgsb_py.py
ilayn: We prefer not having naked asserts, one because they are not good at user experience when they are hit and two if it is supposed to guard something optimizations remove it and it becomes a silent grenade down the line. For local uses it is OK but not in distributed code. Either this needs a meaningful exception or should not be there if it is not going to be hit because in a few versions noone will remember why it was put in.

scipy/optimize/_lbfgsb_py.py
ilayn: You already converted X toQ so there is no foreign dtype left no?

scipy/optimize/_lbfgsb_py.py
Gattocrucco: Ok, removed.

scipy/optimize/_lbfgsb_py.py
Gattocrucco: Yes. You seem to agree as well so I did the inplace assignment.

j-bowhay (approved):
The benchmark in the original issue looks like this is a nice speed-up. Thanks for the contribution @Gattocrucco, keep them coming:)

PR #22557, opened 2025-02-19
ENH: Replace `_lazywhere` with `xpx.apply_where`

crusaderky:
- Adds support for `jax.jit`
- Adds support for Dask
- Adds support for non-scalar fill_value
- Slightly reworked, hopefully improved UI
- Reverts workaround introduced by https://github.com/scipy/scipy/pull/22393 to mdhaber/marray#89. Unit tests remain all green. @mdhaber


mdhaber:
> Note that marray tests do not run in CI

They should be running in the array API job because we install `marray` there. Is that not happening?

mdhaber:
To reduce the diff and potential for regressions, can we move the existing `_lazywhere` to `stats._distn_infrastructure` and use it there (with a note to use it only in the old distribution infrastructure and old distributions)? I don't foresee that code being updated for array API. I might give old distribution private methods array API support so they can be used with the new infrastructure, and at that time, I would swap out `_lazywhere` with `apply_where`.

scipy/stats/_stats_py.py
```diff
@@ -670,7 +668,7 @@ def tmean(a, limits=None, inclusive=(True, True), axis=None):
     # explicit dtype specification required due to data-apis/array-api-compat#152
     sum = xp.sum(a, axis=axis, dtype=a.dtype)
     n = xp.sum(xp.asarray(~mask, dtype=a.dtype), axis=axis, dtype=a.dtype)
-    mean = _lazywhere(n != 0, (sum, n), xp.divide, xp.nan)
+    mean = xpx.apply_where(n != 0, operator.truediv, (sum, n), fill_value=xp.nan)
```
crusaderky, line 671: Explanation here: https://github.com/data-apis/array-api-extra/pull/141/files#r1961964801

crusaderky:
> > Note that marray tests do not run in CI
> 
> They should be running in the array API job because we install `marray` there. Is that not happening?

You're correct, my bad.

crusaderky:
> To reduce the diff and potential for regressions, can we move the existing `_lazywhere` to `stats._distn_infrastructure` and use it there

To clarify, are you asking me to revert to `_lazywhere` in
- `scipy.stats._continuous_distns`
- `scipy.stats._discrete_distns`
- `scipy.stats._distn_infrastructure`

and leave `xpx.apply_where` in the rest of `scipy.stats`?
- `scipy.stats._levy_stable`
- `scipy.stats._multivariate`
- `scipy.stats._stats_py`
- `scipy.stats._wilcoxon`

> (with a note to use it only in the old distribution infrastructure and old distributions)? 

Sorry I'm not familiar with what you mean with "old" versus "new" infrastructure - could you explain, give me a pointer to the documentation, or give me the exact verbiage?

mdhaber:
Thanks for asking. Yeah - very close. `scipy.stats._continuous_distns`, `scipy.stats._discrete_distns`, `scipy.stats._distn_infrastructure`, and `scipy.stats._levy_stable`. and `scipy.stats._multivariate` are "old" in contrast to `scipy.stats._distribution_infrastructure`, `scipy.stats_probability_distribution`, and `scipy.stats._new_distributions` (which need to be renamed at some point). So originally I would have asked for those to be left alone. 

But you know what - since you've already done the work, let's just rever the changes in `scipy.stats._distn_infrastructure`, move `_lazywhere` there, and restrict use of `_lazywhere` to that.

scipy/stats/_multivariate.py
```diff
@@ -4625,7 +4626,7 @@ def asymptotic(dim, df):
 
         # preserves ~12 digits accuracy up to at least `dim=1e5`. See gh-18465.
         threshold = dim * 100 * 4 / (np.log(dim) + 1)
-        return _lazywhere(df >= threshold, (dim, df), f=asymptotic, f2=regular)
+        return xpx.apply_where(df < threshold, regular, asymptotic, (dim, df))
```
mdhaber, line 4629: I'm not sure if it matters here, but reversing the order can change what happens with NaNs. Was they idea that the "regular" case should come first?

scipy/stats/_stats_py.py
mdhaber: Haven't studied the reason yet, but change looks fine.

scipy/stats/_discrete_distns.py
```diff
@@ -867,14 +868,13 @@ def _rvs1(M, n, r, size, random_state):
...
     def _logpmf(self, k, M, n, r):
-        cond = ((r == 0) & (k == 0))
-        result = _lazywhere(~cond, (k, M, n, r),
-                            lambda k, M, n, r:
-                                (-betaln(k+1, r) + betaln(k+r, 1) -
-                                 betaln(n-k+1, M-r-n+1) + betaln(M-r-k+1, 1) +
-                                 betaln(n+1, M-n+1) - betaln(M+1, 1)),
-                            fillvalue=0.0)
-        return result
+        return xpx.apply_where(
```
mdhaber, line 871: Maybe I'm confusing myself, but did this change the order? I thought the intent was to simplify
```python3
cond = ((r == 0) & (k == 0))
~cond
```
to 
```python3
(r != 0) | (k != 0)
```
but I see that the order of the arguments also changed.

scipy/_lib/tests/test__util.py
```diff
@@ -647,65 +641,3 @@ def test_rng_deterministic(self, method, arg_name):
         res3 = method(self, **{arg_name: None})
         assert_equal(res2, res1)
         assert_equal(res3, res1)
-
-
-class TestLazywhere:
```
mdhaber, line 652: Does `apply_where` still pass this test

mdhaber (commented):
I reviewed most files and it all looks pretty good, but I'd like to address https://github.com/scipy/scipy/pull/22557#discussion_r1962042105/https://github.com/data-apis/array-api-extra/pull/141/files#r1962054546 before finishing up. Also, I'd feel more comfortable with the change if the test with `hypothesis` were run on `apply_where` (e.g. if it were brought over to `array-api-extra`). That was written pretty carefully IIRC.

scipy/stats/_multivariate.py
crusaderky: I didn't consider the use case of NaNs. Should I revert all these swaps just to be safe?

scipy/stats/_stats_py.py
```diff
@@ -10959,7 +10957,8 @@ def _xp_var(x, /, *, axis=None, correction=0, keepdims=False, nan_policy='propag
             n = n - xp.sum(nan_mask, axis=axis, keepdims=keepdims)
 
         # Produce NaNs silently when n - correction <= 0
-        factor = _lazywhere(n-correction > 0, (n, n-correction), xp.divide, xp.nan)
+        nc = n - correction
+        factor = xpx.apply_where(nc > 0, operator.truediv, (n, nc), fill_value=xp.nan)
```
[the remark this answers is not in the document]
crusaderky: Readability and speed, as it's no longer doing the subtraction twice.
Happy to revert if you feel strongly about it.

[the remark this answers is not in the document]
mdhaber: Oops, no, I didn't see it was used twice.

scipy/stats/_multivariate.py
mdhaber: It would be safer, but I think whether to swap depends on whether there is going to be any asymmetry between the arguments. For instance, in `np.testing.assert_close`, the first argument should be the actual result, and second should be the (desired) reference value. Is there any such concept with `apply_where`? If not, fine to change it back.

crusaderky: > Is there any such concept with apply_where?

`apply_where(cond, f1, f2, args)` will always produce identical results to `apply_where(~cond, f2, f1, args)`.
There isn't any performance difference either.

scipy/_lib/_util.py
```diff
@@ -72,88 +71,6 @@ class Generator:  # type: ignore[no-redef]
...
-        if type(fillvalue) in {bool, int, float, complex}:
-            with np.errstate(invalid='ignore'):
-                dtype = (temp1 * fillvalue).dtype
-        else:
-           dtype = xp.result_type(temp1.dtype, fillvalue)
-        # whenever mdhaber/marray#89 is resolved, `data` won't need to be extracted here
-        # whenever PAAPIS 2024.12 is released, no need for fill_value to be an array
-        fill_value = xp.asarray(fillvalue, dtype=dtype)
-        fill_value = fill_value.data if is_marray(xp) else fill_value
-        out = xp.full(cond.shape, dtype=dtype, fill_value=fill_value)
```
mdhaber, line 144: In retrospect, I think this should not have converted the argument to an array. I was thinking of `where` (and similar), which currently requires the arguments to be arrays.

mdhaber: I've been burned by that sort of thing so many times it's just a reflex.

scipy/_lib/tests/test__util.py
crusaderky: I've added them to `array_api_extra`. all green.

scipy/stats/_multivariate.py
mdhaber: K, then yeah, I suppose it's safer to change it back.

scipy/_lib/tests/test__util.py
mdhaber: Ah, great!

scipy/stats/_multivariate.py
crusaderky: Reverted all swaps

crusaderky:
@mdhaber I've changed the parameters as requested. Both this PR and https://github.com/data-apis/array-api-extra/pull/141 are ready for final review.
(merge blocked by https://github.com/data-apis/array-api-extra/pull/141).

scipy/stats/_continuous_distns.py
```diff
@@ -7806,39 +7813,33 @@ def _rvs(self, df, nc, size=None, random_state=None):
         return random_state.noncentral_chisquare(df, nc, size)
 
     def _logpdf(self, x, df, nc):
-        cond = np.ones_like(x, dtype=bool) & (nc != 0)
```
mdhaber, line 7809: Do you think this was effectively for broadcasting? I'm assuming it passes tests without it, but did you dig any deeper to see if it was needed?

scipy/stats/_continuous_distns.py
```diff
@@ -10372,45 +10370,50 @@ def n_th_moment(n, a, b):
             Returns n-th moment. Defined only if n >= 0.
             Function cannot broadcast due to the loop over n
             """
-            pA, pB = self._pdf(np.asarray([a, b]), a, b)
-            probs = [pA, -pB]
+            ab = np.asarray([a, b])
+            pA, pB = self._pdf(ab, a, b)
+            probs = np.asarray([pA, -pB])
+            cond = probs.astype(bool)
```
mdhaber, line 10376: If you were tempted to write this as `cond = probs != 0`, I would agree with it.

mdhaber (approved):
Pretty much everything looks straightforward, especially without any comparison direction swaps. Looks good to me after http://data-apis/array-api-extra#141 is merged. (Marking as draft until then to prevent merge.)

You know, SciPy also has a `_lazyselect` (generalization of `_lazywhere` to arbitrary number of callables) in case you haven't had enough fun yet.

scipy/stats/tests/test_stats.py
```diff
@@ -3074,6 +3077,7 @@ def test_zscore_constant_input_2d(self, xp):
             z = stats.zscore(y, axis=None)
         xp_assert_equal(z, xp.full(y.shape, xp.asarray(xp.nan)))
 
+    @pytest.mark.filterwarnings("ignore:invalid value encountered:RuntimeWarning:dask")
```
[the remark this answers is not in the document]
crusaderky: You'd need to call xp_assert_equal, which materializes the graph and actually runs the underlying numpy code, inside the block that catches the warning. You'd also probably need to use the `sync` scheduler so that all warnings are raised in the main thread (this can be set easily from conftest.py)

scipy/stats/_continuous_distns.py
crusaderky: I couldn't figure out what the idea behind it was.

crusaderky:
> You know, SciPy also has a `_lazyselect` (generalization of `_lazywhere` to arbitrary number of callables) in case you haven't had enough fun yet.

I'm aware, but it's used exclusively within `_continuous_distns.py`, so until that needs to get migrated that to the Array API it's pretty low in my bucket list.

scipy/stats/tests/test_stats.py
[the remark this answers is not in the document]
lucascolley: FYI @mdhaber https://github.com/dask/dask/issues/3245

rgommers (approved):
This does look quite good to me. One minor question inline. I agree with @mdhaber's review - can be merged once `xpx.apply_where` is in.

> Slightly reworked, hopefully improved UI

Yep I think so!

scipy/stats/_stats_py.py
```diff
@@ -692,11 +689,11 @@ def tmean(a, limits=None, inclusive=(True, True), axis=None):
     # explicit dtype specification required due to data-apis/array-api-compat#152
     sum = xp.sum(a, axis=axis, dtype=a.dtype)
     n = xp.sum(xp.asarray(~mask, dtype=a.dtype), axis=axis, dtype=a.dtype)
-    mean = _lazywhere(n != 0, (sum, n), xp.divide, xp.nan)
+    mean = xpx.apply_where(n != 0, (sum, n), operator.truediv, fill_value=xp.nan)
```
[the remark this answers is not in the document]
lucascolley: See https://github.com/data-apis/array-api-extra/issues/160:

> If you forget about the meta-namespace and just use xp in the lambdas, at the moment most things will keep working.
> This is because accidentally several functions in the dask.array, numpy, and cupy namespaces are interoperable or even the same function. However you will find cases where this doesn't hold true and you need the correct namespace.
> 
> This will become a much bigger source of headaches in the future when dask around generic Array API compatible namespaces will become commonplace (note: Dask does NOT support them today).
> 
> This pattern repeats itself many, many times in scipy. At the moment there are only a handful of cases that are array API-aware, and they all use xp.divide, so the problem can be worked around by replacing it with operator.truediv. But if you look at scipy.stats in https://github.com/scipy/scipy/pull/22557 you'll find a miriad of calls to np. functions inside the lambdas.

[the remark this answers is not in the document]
rgommers: Ah yes, I remember that now. Given that the code now looks a little harder to understand and this trick won't actually work in other places, I'd probably keep it unchanged and have an issue for solving this problem - and just fail the Dask test in the meantime.

If that's too much churn and this one-line change is more pragmatic, then fine with me as well of course. Dask just has a bunch of issues with namespacing, and this feels like a hack that happens to work.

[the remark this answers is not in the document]
crusaderky: AFAIK `operator.truediv` and `xp.divide` are identical though? It doesn't feel harder to read to me before or after the change?

scipy/stats/tests/test_stats.py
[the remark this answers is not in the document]
crusaderky: There can be two reasons why you're getting a warning in Dask but not in numpy:
1. your production code contains
```python
with np.errstate(...='ignore'):
    (critical section)
```
This can be fixed by using `xpx.apply_where`.

2. your test code contains
```python
with np.errstate(...='ignore'):
    actual = tested_func()
xp_assert_equal(actual, expect)
```
This is fixed by wrapping tested_func in `lazy_xp_function`.

crusaderky:
Test failures are unrelated. @mdhaber this is ready to merge!

mdhaber (approved):

PR #22560, opened 2025-02-19
ENH:  Allow endpoints of custom distributions created with `stats.make_distribution` to be defined with callables

steppi:
#### Reference issue

This is a follow-up to the PR #22371. There's no reference issue, but this was mentioned by @mdhaber in the PR comment https://github.com/scipy/scipy/pull/22371#issue-2800372373.

> Currently, this doesn't allow the `support` to be an arbitrary function of the shape parameters. That would be a very easy follow-up.

~This updates `stats.make_distribution` to allow users to create custom distributions with variable support. It updates to the internals of `stats.make_distribution` when called on existing SciPy distributions with variable support from the old infrastructure, to rely on the same codepaths that allow variable support for custom distributions.~

~The general idea is for instances realizing the abstract base class `_Domain` to accept callables for the `endpoints` parameter, and for `generate_numerical_endpoints` to be updated to handle the case where `endpoints` is a callable. This was a straightforward change, aligning with @mdhaber's comment that this would be an easy follow-up.~

This PR updates `stats.make_distribution` to allow the endpoints of the support to be defined with using callables which map the parameters to numerical endpoints. It does this by updating `get_numerical_endpoints` in domains to handle the case of callable's for endpoints. The code paths `stats.make_distribution` takes when called on distributions from the old infrastructure have been updated to use this new feature.

I've added tests which use `make_distribution` to create a custom version of the [Generalized Extreme Value Distribution](https://en.wikipedia.org/wiki/Generalized_extreme_value_distribution), which is available in SciPy as [scipy.stats.genextreme](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.genextreme.html).
#### Additional information

~This is still a work in progress. I haven't documented things completely yet and there is still a test that is failing due to the `cdf` formula in the custom distribution not being defined on the endpoint of the support in some cases.~


scipy/stats/_distribution_infrastructure.py
```diff
@@ -395,13 +406,22 @@ class _RealDomain(_SimpleDomain):
...
-        a, b = self.endpoints
-        left_inclusive, right_inclusive = self.inclusive
+        if callable(self.endpoints):
+            if self.endpoints.__doc__ is not None:
+                return self.endpoints.__doc__
+            # If endpoints callable has no docstring, fall back to a generic
+            # str which only expresses that the endpoints equal some function
+            # of the parameters.
+            params = inspect.getfullargspec(self.endpoints).kwonlyargs
+            a, b = f"f({','.join(params)})", f"g({','.join(params)})"
```
steppi, line 416: Is it overkill to bring in `inspect` just to have a better default `__str__` for domains? I'm happy to get rid of this.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3497,17 +3517,21 @@ def make_distribution(dist):
 
             endpoints : tuple
                 A tuple defining the lower and upper endpoints of the domain of the
-                parameter. Allowable values are floats or the name (string) of another
+                parameter; allowable values are floats or the name (string) of another
```
steppi, line 3520: Oops. I didn't mean to change this. This happened because I'd accidentally documented `endpoints` as taking a method here instead of `support`, and the semicolon seemed to make things flow better after adding an additional sentence on `method` values of `endpoints`, since it make the linking of the first two sentence more clear.

```suggestion
                parameter. Allowable values are floats or the name (string) of another
```

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3497,17 +3517,21 @@ def make_distribution(dist):
...
+        support : tuple or method.
             A tuple defining the lower and upper endpoints of the support of the
-            distribution. Allowable values are floats or the name (string) of a
-            parameter.
+            distribution or a callable which calculates such a tuple based on the
+            parameter values. If a tuple, values can be floats or the name (string)
+            of a parameter. If ``support`` is defined as a method, it must take the
+            parameter names as keyword arguments and return a 2-tuple giving the
+            numerical lower and upper endpoints of the support for fixed values of
+            the parameters.
```
steppi, line 3534: I need to document that if a `method` is passed for  `support`, it must accept array arguments for the parameters, and in that case return a tuple of arrays.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3717,15 +3739,26 @@ def _overrides(method_name):
...
-    _x_support = _RealDomain(endpoints=support, inclusive=(True, True))
+    if callable(dist.support):
+        def support(**parameter_values):
+            a, b = dist.support(**parameter_values)
+            return np.asarray(a)[()], np.asarray(b)[()]
+    else:
+        support = dist.support
+
+    if hasattr(dist, "support_inclusive"):
+        inclusive = dist.support_inclusive
```
steppi, line 3757: Do we actually want to allow letting users decide if the support is `inclusive`?. If so, is `support_inclusive` the best attribute name to use here? I added this because the `cdf` and `pdf` formulas for the generalized extreme value distribution used in the tests isn't defined on a finite endpoint of the support in some cases.

scipy/stats/tests/test_continuous.py
```diff
@@ -1148,6 +1148,76 @@ def pdf(self, x, a, b):
...
+                right[c > 0] = mu[c > 0] + sigma[c > 0] / c[c > 0]
+
+                return left[()], right[()]
+
+            @property
+            def support_inclusive(self):
+                return (False, False)
+
+            def pdf(self, x, *, c, mu, sigma):
+                x, c, mu, sigma = np.broadcast_arrays(x, c, mu, sigma)
```
steppi, line 1184: I saw an error in the tests if I didn't manually broadcast the arrays like this in these methods.

scipy/stats/tests/test_continuous.py
```diff
@@ -1148,6 +1148,76 @@ def pdf(self, x, a, b):
...
+                t[~mask] = (1  - c[~mask]*(x[~mask] - mu[~mask])/sigma[~mask])**(1/c[~mask])
+                result = 1/sigma * t**(1 - c)*np.exp(-t)
+                return result[()]
+
+            def cdf(self, x, *, c, mu, sigma):
+                x, c, mu, sigma = np.broadcast_arrays(x, c, mu, sigma)
+                t = np.empty_like(x)
+                mask = (c == 0)
+                t[mask] = np.exp(-(x[mask] - mu[mask])/sigma[mask])
+                t[~mask] = (1  - c[~mask]*(x[~mask] - mu[~mask])/sigma[~mask])**(1/c[~mask])
```
steppi, line 1197: The tests currently fail due to this formula not being defined on one endpoint of the support when `c` is negative. Do I just need to ensure the `cdf` formula is defined on the endpoints.

mdhaber: Depends.

I think I was very careful to make sure that the public methods respect `inclusive` and do the right thing when a parameter or argument value is an endpoint. But it is possible that there are private methods that are called with an *argument* (not parameter, I think - those would just become NaN) at an endpoint even if it is not supposed to be "inclusive". I guess I figured that if this caused a problem, we'd find out in tests and adjust whichever method needed to be adjusted.

But I'll look into it.

mdhaber (commented):
This is looking pretty good. The main thing to think about is whether `endpoint` should *always* be a 2-tuple, and the elements can either be a float, string, or callable. Then, the natural extension to that idea is for custom distribution `support` to be defined just like the parameters - in a dictionary with `endpoints` and `inclusive`.

scipy/stats/tests/test_continuous.py
mdhaber: I see. What I was thinking is that the public methods ensure that the arguments are broadcasted before calling the private methods. However, the integrator and root finder might not actually do the broadcasting and just pass arguments with broadcastable shapes. For example, if you try to evaluate the `cdf` with `x = [1, 2, 3]` and `c`, `mu`, and `sigma` as scalars, the private `_cdf_dispatch` gets called with `c`, `mu`, and `sigma` broadcasted with `x`. However, then the integrator might want to evaluate at an array `x` with shape `(3, 66)` and the parameters of shape `(3, 1)`.

There are pros and cons to pre-emptive broadcasting. In principle, you can save work if the broadcasting is not done before passing the arguments in; e.g., if there is setup that depends only on the parameter values, it could be re-used for each value of the argument. But we're already broadcasting the parameters and argument in the public method, so if that were the goal, we have partially eliminated the advantage.

Let's consider this to be fine for now, and think of it as a micro-optimization we might want to take on later - either avoid broadcasting in the public methods before calling the private methods. or ensure that the integrator/root finder *also* broadcast before calling the callable.

scipy/stats/_distribution_infrastructure.py
steppi: OK, yeah. I thought of that too but was worried it might be too cumbersome. I'm happy with that.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -262,15 +266,18 @@ class _SimpleDomain(_Domain):
...
     """
     def __init__(self, endpoints=(-inf, inf), inclusive=(False, False)):
         self.symbols = super().symbols.copy()
-        a, b = endpoints
-        self.endpoints = np.asarray(a)[()], np.asarray(b)[()]
+        if callable(endpoints):
+            self.endpoints = endpoints
+        else:
+            a, b = endpoints
+            self.endpoints = np.asarray(a)[()], np.asarray(b)[()]
```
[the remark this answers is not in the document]
steppi: OK, yeah. I think (premature) optimization of the case where computing the left and right endpoints required duplicate work is why I did it this way. I'm happy with having separate callables.

scipy/stats/tests/test_continuous.py
mdhaber: The infrastructure mostly uses `inclusive` to determine whether *parameters* are in the valid range. For support, I think `inclusive` only makes a difference in the `pdf` and `logpdf` - if `inclusive`, it will rely on the private PDF method at the endpoint, otherwise, it will return 0 because the endpoint is not in the support. For public methods like `cdf` and such, it doesn't really matter whether the developer considers a support endpoint to be in the support or not. For instance, the CDF at the left endpoint is always going to be zero, so it doesn't rely on the `cdf` method for that sort of thing, regardless of `inclusive`.

Since the integrator uses tanh-sinh quadrature, the endpoints have zero weight. The way it's implemented, it might evaluate the `pdf` or other methods at the endpoints of the support, but the value is ignored - even if it's NaN, I think.  So `inclusive` is not used here, and probably doesn't need to be.

The root finder is what is evaluating your class' `cdf` at the left end of the support when the CDF is being inverted. Your implementation emits a warning, but it actually produces the correct result, I think, so all that would be required to make the test pass is to ignore the warning with `np.errstate`.

However, I think there is an improvement to be made here. It might actually be related - or maybe solved in conjunction with - the issue with slow convergence of the root finder when the correct result is the left end of the support (see end of top post of gh-22312. I think you were slated to look into that, too, so perhaps both can be tackled at once.) Details below.

<details>

I investigated by having the `cdf` return NaN when evaluated at the left end of the support. I found that `icdf` still worked, but it started taking a really long time. What happens is that even though the bracket finder sees NaN when evaluated at `x_min`, it continues looking for a zero-crossing toward the right. It never finds one, of course, and returns with an invalid bracket with the right end at an absurdly large float. That doesn't stop `_chandrupatla` from being executed, and apparently, it even converges to the correct root even though it got a NaN at the left endpoint. (I thought it would have errored out.) It just takes a really long time because the right end of the bracket is so absurdly high.

I'm not sure what the best solution is. I don't want the root finder to go through the full public method just to avoid this - there's lots of reasons that would be problematic as written. Two obvious things we could do if `inclusive=False`:
- wrap the function passed to `_solve_bounded`; for instance, if it's trying to invert the CDF, we wrap the function to return 0 at the left end of the support. 
- increment the endpoint with `np.nextafter`

Whatever happens, we have a problem in `find_root`:
```python3
import numpy as np
from scipy.optimize.elementwise import find_root

def f(x):
    return np.where(x != 0, 1-x, np.nan)

res = find_root(f, (0, 2.5))  # this shouldn't work
res = find_root(f, (0, 1.1))  # this returns nonsense
```
I'm really not sure how we missed that.

</details>

scipy/optimize/_bracket.py
```diff
@@ -289,7 +289,9 @@ def check_termination(work):
         # not there.
         j = j[j < work.active.shape[0]]
         # Check whether they are still there.
-        j = j[also_stop == work.active[j]]
+        if j.size:
+            # If j is empty, there's nothing to check.
+            j = j[also_stop == work.active[j]]
```
mdhaber, line 294: MRE:
```python3
import numpy as np
from scipy.optimize._bracket import _bracket_root

def f(x, p):
    return np.exp(x) - p

p = np.asarray([0.29, 0.62, 0.41, 0.35])
res = _bracket_root(f, xl0=-1, xmin=-np.inf, xmax=0, args=(p,))
```

mdhaber: After playing with this for far too long, I've come to the conclusion that this was an obvious logical gap that we would have caught at the time if we only had the right test case.

The code says:
```python3
        # If the location exceeds the length of the `work.active`, they are
        # not there.
        j = j[j < work.active.shape[0]]
```
and after this line `j` is empty. And before your change, it said:

```python3
        # Check whether they are still there.
        j = j[also_stop == work.active[j]]
```
But as your comment indicates, there just isn't anything to check. We know they're not there.

However, it seems that it would also be possible for `j.size` to be nonzero yet `also_stop` and `j` still have different lengths, so I'm not sure this is a complete fix.

mdhaber: Well here's another issue.
```python3
_bracket_root(lambda x: x + 0.25, xl0=-0.5, xmin=-np.inf, xmax=0)
#      success: False
#       status: -5
#           xl: -0.5
#           xr: 0.5
#          nit: 0
#         nfev: 2
#           fl: -0.25
#           fr: 0.75
```
I'll have to look back at whether we lost commits. I am surprised this happens; I thought we fixed this stuff.

mdhaber: OK, not an MRE, but here's evidence that we can still hit the same problem:
```python3
import numpy as np
from scipy.optimize._bracket import _bracket_root

rng = np.random.default_rng(2385492345)

for i in range(100):
    print(i)
    n = rng.integers(4, 10)
    c = np.sign(rng.normal(size=n))

    def f(x, p, c):
        return np.exp(x*c) - p

    p = rng.random(n)
    xl0 = rng.random(size=n) * 10 * (-c)
    xmin = np.where(c < 0, 0, -np.inf)
    xmax = np.where(c < 0, np.inf, 0)
    xr0 = np.clip(xl0 + 1, xmin, xmax)

    res = _bracket_root(f, xl0=xl0, xr0=xr0, xmin=xmin, xmax=xmax, args=(p, c))
#   File "C:\Users\matth\Desktop\scipy\scipy\optimize\_bracket.py", line 294, in check_termination
#     j = j[also_stop == work.active[j]]
#           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
# ValueError: operands could not be broadcast together with shapes (3,) (2,) 
```

mdhaber: I think I've figured out what this should have been in the first place. Just to make sure we get this right, let's see if we come to the same conclusion.

scipy/stats/tests/test_continuous.py
steppi: Cool. I used `pytest.mark.filterwarnings` to make the tests pass. I'll take a look at the underlying issue when I look at gh-22312.

scipy/stats/tests/test_continuous.py
```diff
@@ -1148,6 +1148,76 @@ def pdf(self, x, a, b):
                 assert_allclose(X.moment(order, kind=kind),
                                 Y.moment(order, kind=kind))
 
+    @pytest.mark.parametrize("c", [-1, 0, 1])
```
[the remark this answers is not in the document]
steppi: Done. I added an array case.

scipy/stats/_distribution_infrastructure.py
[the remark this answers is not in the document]
steppi: Done. `endpoints` should always be a `tuple` again. I also realized that it was specifically having two calls to `_get_support` in `make_distribution_rv_generic` like below that made me make it so `endpoints` could be a `callable` that returns a `tuple`.

```python
    if _overrides("_get_support"):
        def left(**parameter_values):
            a, _ = dist._get_support(**parameter_values)
            return np.asarray(a)[()]

        def right(**parameter_values):
            _, b = dist._get_support(**parameter_values)
            return np.asarray(b)[()]

        endpoints = (left, right)
    else:
        endpoints = support
```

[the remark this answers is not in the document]
mdhaber: Yeah, I thought about that, but I think it's OK. I don't think we should let the old design stop us from doing something that's more convenient here, anyway. If it were really a problem, we could probably memoize

scipy/stats/tests/test_continuous.py
mdhaber: OK. Were you getting more warnings than `np.errstate` could handle? `errstate` worked for me locally.

scipy/stats/tests/test_continuous.py
```diff
@@ -1148,6 +1148,80 @@ def pdf(self, x, a, b):
...
                                 Y.moment(order, kind=kind))
 
+    # pdf and cdf formulas below can warn on boundary of support in some cases.
+    # See https://github.com/scipy/scipy/pull/22560#discussion_r1962763840.
+    @pytest.mark.filterwarnings("ignore::RuntimeWarning")
+    @pytest.mark.parametrize("c", [-1, 0, 1, np.asarray([-2.1, -1., 0., 1., 2.1])])
+    def test_custom_variable_support(self, c):
+        rng = np.random.default_rng(7548723590230982)
+
+        class MyGenExtreme:
```
mdhaber, line 1158: I didn't check the formulas line by line. I figured the fact that the results agree with the existing implementation (without *using* the existing implementation) is sufficient, but LMK if you want me to look more carefully.

scipy/stats/tests/test_continuous.py
```diff
@@ -1148,6 +1148,80 @@ def pdf(self, x, a, b):
...
+                    return result[()]
+
+                return {"endpoints": (left, right), "inclusive": (False, False)}
+
+            def pdf(self, x, *, c, mu, sigma):
+                x, c, mu, sigma = np.broadcast_arrays(x, c, mu, sigma)
+                t = np.empty_like(x)
+                mask = (c == 0)
+                t[mask] = np.exp(-(x[mask] - mu[mask])/sigma[mask])
+                t[~mask] = (1  - c[~mask]*(x[~mask] - mu[~mask])/sigma[~mask])**(1/c[~mask])
```
mdhaber, line 1192: Not needed now because `_lazywhere` is being transitioned to `xpx.apply_where`, but just to show what it would look like:
```suggestion
                t = _lazywhere(c == 0, (x, c, mu, sigma),
                               lambda x, c, mu, sigma: np.exp(-(x - mu)/sigma),
                               lambda x, c, mu, sigma: (1 - c*(x - mu)/sigma)**(1 / c))
```
A bit easier on the eyes, and I think it does the broadcasting for you.
But maybe it's because I was looking at `apply_where` PR for an hour and got used to it.

scipy/stats/tests/test_continuous.py
```diff
@@ -1148,6 +1148,80 @@ def pdf(self, x, a, b):
                 assert_allclose(X.moment(order, kind=kind),
                                 Y.moment(order, kind=kind))
 
+    # pdf and cdf formulas below can warn on boundary of support in some cases.
+    # See https://github.com/scipy/scipy/pull/22560#discussion_r1962763840.
+    @pytest.mark.filterwarnings("ignore::RuntimeWarning")
+    @pytest.mark.parametrize("c", [-1, 0, 1, np.asarray([-2.1, -1., 0., 1., 2.1])])
+    def test_custom_variable_support(self, c):
```
mdhaber, line 1155: Does this run reasonably fast locally?

scipy/stats/_distribution_infrastructure.py
```diff
@@ -393,15 +402,23 @@ class _RealDomain(_SimpleDomain):
...
+            if endpoint.__doc__ is not None:
+                return endpoint.__doc__
+            params = inspect.signature(endpoint).parameters.values()
+            params = [
+                p.name for p in params if p.kind == inspect.Parameter.KEYWORD_ONLY
+            ]
+            return f"{funcname}({','.join(params)})"
+        return self.symbols.get(endpoint, f"{endpoint}")
 
     def __str__(self):
```
mdhaber, line 416: There are some tests of the `__str__` method of `_RealDomain` that maybe could be updated:
https://github.com/scipy/scipy/blob/0b91033ba51a8756de2c4b3ce4c1e6dfeb3884eb/scipy/stats/tests/test_continuous.py#L108

Could also add oddball things to
https://github.com/scipy/scipy/blob/0b91033ba51a8756de2c4b3ce4c1e6dfeb3884eb/scipy/stats/tests/test_continuous.py#L1713

if they don't fit elsewhere.

mdhaber (commented):
Looking good! A few thoughts, but this looks quite close.

scipy/stats/tests/test_continuous.py
steppi:
```
python dev.py test -t scipy.stats.tests.test_continuous -- -k custom_variable_support
```

runs in 0.67 seconds locally for me. I guess that may be a little too slow?

scipy/stats/tests/test_continuous.py
steppi: I like `_lazywhere` and have used it before. I was thinking since this code  is supposed to be an example of a user created custom distribution, it wouldn't be good to rely on something outside SciPy's public API. I was thinking it might be good to put this example in the documentation.

scipy/stats/tests/test_continuous.py
steppi: I think the formulas should be fine. The agreement should be enough.

scipy/stats/tests/test_continuous.py
steppi: `np.errstate` works too. Maybe this isn't the best reason, but my thought was ideally the example here shouldn't be raising a runtime warning due to behavior on the support boundary if inclusive is `False` for the relevant endpoint, but it just happens that this isn't working as well as it could. I saw putting the ignore at the `pytest` level as stating that a runtime warning shouldn't be raised here, and we are just temporarily ignoring itin the test until the problem is fixed.

mdhaber: Ok, that makes sense - it shouldn't be called at the endpoint, so it shouldn't need to be patched in the code, right? Ok.

scipy/stats/tests/test_continuous.py
mdhaber: Mmm. yes, I was thinking we need an example of arbitrary function `support` in he documentation, too, and you're right that we wouldn't want to use a private function there.

mdhaber:
Let's get this in, since there are a few things for a follow-up: 
- example in documentation
- I think `parameters` could be allowed to return a tuple of dictionaries like the one it returns now. Each would simply be a separate `_Parameterization`.
- probably something for an even later follow-up: a way to specify the `_process_parameters` method.
- tests of some of the new `_SimpleDomain` behavior
- default values of `endpoints` and `inclusive`

Could you just update the title and top post to reflect that this allows the support endpoints to be specified by an arbitrary callable? (Variable support of a very common form was already allowed.)

scipy/stats/tests/test_continuous.py
```diff
@@ -1120,7 +1120,7 @@ def parameters(self):
 
             @property
             def support(self):
-                return 'a', 'b'
+                return {'endpoints': ('a', 'b'), 'inclusive': (True, True)}
```
mdhaber, line 1123:
```suggestion
                return {'endpoints': ('a', 'b')}
```
We could test the default here.

mdhaber:
Thanks @steppi!

PR #22566, opened 2025-02-20
TST: `skip_xp_backends(eager_only=True)`

crusaderky:
xref https://github.com/scipy/scipy/pull/22469#discussion_r1963936773

- Add `skip_xp_backends(eager_only=True, reason=...)` and `xfail_xp_backends(eager_only=True, reason=...)`
- general cleanup of the backends skip mechanics

### DIFF
--- a/scipy/cluster/tests/test_hierarchy.py
+++ b/scipy/cluster/tests/test_hierarchy.py
@@ -336,8 +336,6 @@ class TestLeaders:
                   reason='`is_isomorphic` only supports NumPy backend')
 class TestIsIsomorphic:
 
-    @skip_xp_backends(np_only=True,
-                      reason='array-likes only supported for NumPy backend')
     def test_array_like(self, xp):
         assert is_isomorphic([1, 1, 1], [2, 2, 2])
         assert is_isomorphic([], [])
@@ -417,8 +415,6 @@ class TestIsIsomorphic:
                 Q = np.random.permutation(nobs)
                 b[Q[0:nerrors]] += 1
                 b[Q[0:nerrors]] %= nclusters
-            a = xp.asarray(a)
-            b = xp.asarray(b)
             assert is_isomorphic(a, b) == (not noniso)
             assert is_isomorphic(b, a) == (not noniso)
 
--- a/scipy/conftest.py
+++ b/scipy/conftest.py
@@ -13,7 +13,7 @@ import hypothesis
 
 from scipy._lib._fpumode import get_fpu_mode
 from scipy._lib._array_api import (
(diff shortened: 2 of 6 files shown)


scipy/cluster/tests/test_hierarchy.py
```diff
@@ -336,8 +336,6 @@ def test_leaders_single(self, xp):
                   reason='`is_isomorphic` only supports NumPy backend')
 class TestIsIsomorphic:
 
-    @skip_xp_backends(np_only=True,
-                      reason='array-likes only supported for NumPy backend')
```
crusaderky, line 340: duplicated

scipy/conftest.py
```diff
@@ -167,7 +171,16 @@ def num_parallel_threads():
...
     try:
         import cupy  # type: ignore[import-not-found]
+        # Note: cupy disregards SCIPY_DEVICE and always runs on cuda
         xp_available_backends.update({'cupy': cupy})
+        xp_skip_cpu_only_backends.add('cupy')
+
+        # this is annoying in CuPy 13.x
+        warnings.filterwarnings(
+            'ignore', 'cupyx.jit.rawkernel is experimental', category=FutureWarning
+        )
```
crusaderky, line 181: Moved from below

scipy/conftest.py
```diff
@@ -167,7 +171,16 @@ def num_parallel_threads():
...
+        from cupyx.scipy import signal
+        del signal
```
crusaderky, line 183: Moved from below

crusaderky:
CI failures are unrelated

scipy/conftest.py
```diff
@@ -262,43 +284,68 @@ def xp(request, monkeypatch):
 
 
 def _backends_kwargs_from_request(request, skip_or_xfail):
-    """A helper for {skip,xfail}_xp_backends"""
-    # do not allow multiple backends
-    args_ = request.keywords[f'{skip_or_xfail}_xp_backends'].args
-    if len(args_) > 1:
-        # np_only / cpu_only has args=(), otherwise it's ('numpy',)
-        # and we do not allow ('numpy', 'cupy')
-        raise ValueError(f"multiple backends: {args_}")
```
lucascolley, line 271: If I understand correctly, this is no longer required since `request.node.iter_markers` was introduced—is that right?

lucascolley (commented):
thanks @crusaderky ! One question for my understanding.

scipy/conftest.py
crusaderky: This is now checked by line 341:344. Before that, 341:344 was unreachable.

lucascolley (approved):
thanks @crusaderky !

PR #22567, opened 2025-02-20
BUG: optimize: Fix `bracket_root` termination check and default initial bracket.

steppi:
#### Reference issue

This fixes I bug I discovered while working on #22560.

This PR is fixes a~n edgecase~ bug in the elementwise root bracket finder `scipy.optimize.bracket_root`. For context, this bracket finder starts with an initial bracket and simultaneously searches for valid brackets by expanding this initial bracket to the left and right. If multiple searches in one direction terminate at the same time, in an iteration after all corresponding searches in the other direction have already terminated unsuccessfully (by reaching a boundary point in the search space), currently a `ValueError` will occur due to `==` comparision between two arrays, one of which is empty and one of which is not.

Update:
I've also fixed a bug @mdhaber identified here, https://github.com/scipy/scipy/pull/22560#discussion_r1962947434, involving the choice of the default initial bracket.

#### Additional information

I'd originally fixed this directly in #22560, but am moving to a separate PR since this is the principled thing to do and @mdhaber helped identify a minimal reproducible example for use in a test case.

### DIFF
--- a/scipy/optimize/_bracket.py
+++ b/scipy/optimize/_bracket.py
@@ -55,7 +55,8 @@ def _bracket_root_iv(func, xl0, xr0, xmin, xmax, factor, args, maxiter):
     # Calculate the default value of xr0 if a value has not been supplied.
     # Be careful to ensure xr0 is not larger than xmax.
     if xr0_not_supplied:
-        xr0 = xl0 + xp.minimum((xmax - xl0)/ 8, xp.asarray(1.0, dtype=xl0.dtype))
+        xr0 = xl0 + xp.minimum((xmax - xl0)/ 8, xp.asarray(1.0))
+        xr0 = xp.astype(xr0, xl0.dtype, copy=False)
 
     maxiter = xp.asarray(maxiter)
     message = '`maxiter` must be a non-negative integer.'
@@ -476,9 +477,11 @@ def _bracket_minimum_iv(func, xm0, xl0, xr0, xmin, xmax, factor, args, maxiter):
     # by the user. We need to be careful to ensure xl0 and xr0 are not outside
     # of (xmin, xmax).
     if xl0_not_supplied:
-        xl0 = xm0 - xp.minimum((xm0 - xmin)/16, xp.asarray(0.5, dtype=xm0.dtype))
+        xl0 = xm0 - xp.minimum((xm0 - xmin)/16, xp.asarray(0.5))
+        xl0 = xp.astype(xl0, xm0.dtype, copy=False)
     if xr0_not_supplied:
-        xr0 = xm0 + xp.minimum((xmax - xm0)/16, xp.asarray(0.5, dtype=xm0.dtype))
+        xr0 = xm0 + xp.minimum((xmax - xm0)/16, xp.asarray(0.5))
+        xr0 = xp.astype(xr0, xm0.dtype, copy=False)
 
     maxiter = xp.asarray(maxiter)
     message = '`maxiter` must be a non-negative integer.'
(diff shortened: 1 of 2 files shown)


steppi:
I missed the comments here:  https://github.com/scipy/scipy/pull/22560/files#r1962862935

This needs more work.

scipy/optimize/tests/test_bracket.py
```diff
@@ -345,6 +345,35 @@ def f(x):
                                 xmin=1)
         assert not res.success
 
+        # 5. Edge case. Multiple searches in same direction terminate
```
steppi, line 348: I wrote these tests before I understood what was going on. I guess these aren't even edge cases, just straight up bugs we missed. I'll separate these out into a different test since they're not special cases.

steppi: Actually, nevermind. I guess these new tests are fine here. I'm not sure where to put them or how to name them otherwise.

scipy/optimize/_bracket.py
```diff
@@ -45,6 +52,11 @@ def _bracket_root_iv(func, xl0, xr0, xmin, xmax, factor, args, maxiter):
     if not xp.all(factor > 1):
         raise ValueError('All elements of `factor` must be greater than 1.')
 
+    # Calculate the default value of xr0 if a value has not been supplied.
+    # Be careful to ensure xr0 is not larger than xmax.
+    if xr0_not_supplied:
+        xr0 = xl0 + xp.minimum((xmax - xl0)/ 8, xp.asarray(1.0))
```
mdhaber, line 58: Looks like this would not preserve lower-precision dtypes in libraries for which `float64` is the default.
```suggestion
        xr0 = xl0 + xp.minimum((xmax - xl0)/ 8, xp.asarray(1.0, dtype=xl0.dtype))
```
Could you confirm locally?

Eagerly waiting the next version of the standard, which will accept Python scalars where it makes sense.

scipy/optimize/tests/test_bracket.py
```diff
@@ -345,6 +345,40 @@ def f(x):
...
+            return np.exp(x) - p
+
+        p = np.asarray([0.29, 0.35])
+        res = _bracket_root(f, xl0=-1, xmin=-np.inf, xmax=0, args=(p, ))
+
+        # 6. https://github.com/scipy/scipy/pull/22560/files#r1962952517
+        def f(x, p, c):
+            return np.exp(x*c) - p
+
+        p = [0.32061201, 0.39175242, 0.40047535, 0.50527218, 0.55654373,
```
mdhaber, line 364: subnormal nit: I think this will still fail (but can look a little more compact) if we chop some digits off.

mdhaber (approved):
Other than the dtype issue, this looks good to me.

scipy/optimize/tests/test_bracket.py
mdhaber: Could create a new bug fix test. That would be fine with me.

scipy/optimize/_bracket.py
steppi: Yep, I noticed independently and pushed a fix just now. The same bug was already there in `bracket_minimum` so I fixed it too.

scipy/optimize/tests/test_bracket.py
steppi: True. I'm off for the rest of the day but could do it tomorrow if you think it's important enough.

mdhaber:
Thanks @steppi!

mdhaber:
Marking as a backport candidate in case there is a 1.5.3.

Issue #22534, opened 2025-02-14
CI: failures in`ndimage/tests/test_morphology.py` (Dask) and `linalg/tests/test_extending` 


mdhaber:
### Describe your issue.

Two CI failures have started appearing recently, e.g. gh-22532.

### Reproducing Code Example

```python
-
```

### Error message

```shell
=========================== short test summary info ============================
FAILED scipy/ndimage/tests/test_filters.py::TestNdimageFilters::test_minimum_filter05_overlap[dask.array] - AssertionError: 
Not equal to tolerance rtol=0, atol=1.5e-06

Mismatched elements: 10 / 15 (66.7%)
Max absolute difference among violations: 8
Max relative difference among violations: 0.88888889
 ACTUAL: array([[2, 2, 1, 1, 1],
       [2, 2, 1, 1, 1],
       [5, 3, 3, 1, 1]])
 DESIRED: array([[3, 2, 5, 1, 4],
       [7, 6, 9, 3, 5],
       [5, 8, 3, 7, 1]])
FAILED scipy/ndimage/tests/test_morphology.py::TestNdimageMorphology::test_grey_erosion01_overlap[dask.array] - AssertionError: 
Not equal to tolerance rtol=0, atol=1.5e-06

Mismatched elements: 9 / 15 (60%)
Max absolute difference among violations: 8
Max relative difference among violations: 8.
 ACTUAL: array([[3, 2, 5, 1, 4],
       [7, 6, 9, 3, 5],
       [5, 8, 3, 7, 1]])
 DESIRED: array([[2, 2, 1, 1, 1],
       [2, 3, 1, 3, 1],
       [5, 5, 3, 3, 1]])
= 2 failed, 75981 passed, 29913 skipped, 56621 deselected, 13347 xfailed, 2 xpassed in 1067.28s (0:17:47) =
Error: Process completed with exit code 1.

------

=========================== short test summary info ============================
FAILED scipy/linalg/tests/test_extending.py::test_cython - subprocess.CalledProcessError: Command '['meson', 'setup', '--native-file', '/tmp/pytest-of-runner/pytest-0/popen-gw1/test_cython0/140434317613952/linalg/tests/_cython_examples/interpreter-native-file.ini', '/tmp/pytest-of-runner/pytest-0/popen-gw1/test_cython0/140434317613952/linalg/tests/_cython_examples']' returned non-zero exit status 1.
[...]

PR #22471, opened 2025-02-05
MAINT: backports for 1.15.2

tylerjereddy:
Merge conflicts seem to be causing many (236) test failures--I may back out PRs that had merge conflicts and that don't look critical if I don't see a straightforward path forward for them. Some cases had too much drift relative to `main` and didn't really seem like serious things to fix anyway.

Backports included so far (merge conflicts noted):

1. gh-22038
2. gh-22080
3. gh-22187
4. gh-22322
5. gh-22337
6. gh-22339
7. gh-22353
8. gh-22356
9. gh-22357
10. gh-22372
11. gh-22373
12. gh-22374
13. gh-22405 (merge conflicts)
14. gh-22418
15. gh-22423
16. gh-22446 (there may be a flaky 3.10 CI job?)
17. gh-22447 (merge conflicts)
18. gh-22472
19. gh-22482
20. gh-22402
21. gh-22408
22. gh-22345

TODO:

- [x] fix or back out patches that had merge conflicts and are causing problems
- [x] test failure `test_quadrature.py::TestCumulativeSimpson::test_cumulative_simpson_against_simpson_with_default_dx`
- [x] start restoring the CI and get it passing
- [x] pull in more desired backports (gh-22402, others?)
- [x] update the release notes and probably mention new 3.13t wheels on Linux ARM
- [x] wheel builds succeeding?
- [ ] is the CuPy team happy re: https://github.com/cupy/cupy/issues/8866 ?

PR #22569, opened 2025-02-20
CI: fix dev-deps job by not testing Meson master

rgommers:
The failures happened in `test_extending.py::test_cython`, as reported in gh-22534.

Tyler already fixed this in the 1.15.x branch in gh-22471, but the job has been failing on `main` for the last few days.


tylerjereddy (approved):
Makes sense, I guess I'll let the job pass to be safe, but seems clearly correct for now.

tylerjereddy:
thanks Ralf, the remaining CI failure is already mentioned in the linked issue above

PR #22582, opened 2025-02-24
BUG: ``integrate.solve_ivp``: Avoid duplicate time stamps in `ts` when `dense_output=True`

### LINKED ISSUE #19418: BUG: integrate.solve_ivp fails for some step sizes if dense_output=True with events but no t_eval
### Describe your issue.

Calling scipy.integrate.solve_ivp under the following combination of conditions results in failure:

- Small step size `max_step`.
- `dense_output = True`
- `events` provided.
- `t_eval` not provided. 

The error appears to be after scipy/integrate/_ivp/ivp.py here: https://github.com/scipy/scipy/blob/43bb524ecf1544743493ae69053e38a93c5aee7c/scipy/integrate/_ivp/ivp.py#L689-L691

If an event is found in the current timestep, that when solved gives a time value _exactly equal_ to the start of the interval (i.e. the end of the last interval), appending that time value means a duplicate entry is added to the `ts` array.  When `dense_output = True` and `t_eval` is `None`, this call to `OdeSolution` fails:  https://github.com/scipy/scipy/blob/43bb524ecf1544743493ae69053e38a93c5aee7c/scipy/integrate/_ivp/ivp.py#L727-L731

This is because a monotonically increasing / decreasing series of time values is required.  It seems to only occur if the maximum step size is controlled to be a small value.  A simple fix might be to simply check for a duplicate value before appending the event time mark, but I haven't thought about this too deeply. 

_Added:_ This error is very similar to https://github.com/scipy/scipy/issues/17066#issue-1381164720  The main issue appears to be when the derivative term happens to be an exact multiple of the timestep, so that when the terminating event is reached the solution rounds to exactly the timestep value again - this giving a repeated value.  My workaround is to slightly offset the timestep by a tiny value to prevent this, but this is undesirable.

### Reproducing Code Example

```python
# Example based on the documentation.
import numpy as np
from scipy.integrate import solve_ivp

def upward_cannon(t, y):
    return [y[1], -9.80665]

def hit_ground(t, y):
    return y[0]

hit_ground.terminal = True
hit_ground.direction = -1

sol = solve_ivp(upward_cannon, [0, np.inf],[0, 0.001],
                max_step=0.05 * 0.001 / 9.80665,
                events=hit_ground, dense_output=True)
print(sol.t_events)
```

### Error message

```shell
  File "C:.........\scratch_4.py", line 17, in <module>
    sol = solve_ivp(upward_cannon, [0, np.inf],[0, 0.001],
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:.........\venv\Lib\site-packages\scipy\integrate\_ivp\ivp.py", line 672, in solve_ivp
    sol = OdeSolution(ts, interpolants)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:.........\venv\Lib\site-packages\scipy\integrate\_ivp\common.py", line 159, in __init__
    raise ValueError("`ts` must be strictly increasing or decreasing.")
ValueError: `ts` must be strictly increasing or decreasing.
```

```shell
1.10.1 1.24.2 sys.version_info(major=3, minor=11, micro=3, releaselevel='final', serial=0)
```

czgdp1807:
#### Reference issue

Closes https://github.com/scipy/scipy/issues/19418

I verified that indeed `ts` has duplicate values as pointed in https://github.com/scipy/scipy/issues/19418. I added a check to avoid it. Last `interpolant` is also removed because if `ts[-1] == ts[-2]` then there shouldn't be a point in keeping it. It should be noted that for `len(ts) == 2` duplicates are allowed (so that condition is kept preserved here). 

I also added a regression test. 

Let me know if the fix is making sense. :-)

@ericjwhitney

### DIFF
--- a/scipy/integrate/_ivp/tests/test_ivp.py
+++ b/scipy/integrate/_ivp/tests/test_ivp.py
@@ -167,8 +167,8 @@ def test_duplicate_timestamps():
     assert_allclose(sol.sol(0.01), np.asarray([-0.00039033, -0.08806632]),
                     rtol=1e-5, atol=1e-8)
     assert_allclose(sol.t_events, np.asarray([[0.00203943]]), rtol=1e-5, atol=1e-8)
-    assert_allclose(sol.y_events, [np.asarray([[ 8.5808408e-20, -1.0000000e-02 ]])])
-    assert_(sol.success)
+    assert_allclose(sol.y_events, [np.asarray([[ 0.0, -0.01 ]])], atol=1e-9)
+    assert sol.success
     assert_equal(sol.status, 1)
 
 @pytest.mark.thread_unsafe
--- a/scipy/integrate/_ivp/tests/test_ivp.py
+++ b/scipy/integrate/_ivp/tests/test_ivp.py
@@ -151,6 +151,25 @@ def compute_error(y, y_true, rtol, atol):
     e = (y - y_true) / (atol + rtol * np.abs(y_true))
     return np.linalg.norm(e, axis=0) / np.sqrt(e.shape[0])
 
+def test_duplicate_timestamps():
+    def upward_cannon(t, y):
+        return [y[1], -9.80665]
+
+    def hit_ground(t, y):
+        return y[0]
+
+    hit_ground.terminal = True
+    hit_ground.direction = -1
+
+    sol = solve_ivp(upward_cannon, [0, np.inf], [0, 0.01],
+                    max_step=0.05 * 0.001 / 9.80665,
+                    events=hit_ground, dense_output=True)
+    assert_allclose(sol.sol(0.01), np.asarray([-0.00039033, -0.08806632]),
+                    rtol=1e-5, atol=1e-8)
+    assert_allclose(sol.t_events, np.asarray([[0.00203943]]), rtol=1e-5, atol=1e-8)
+    assert_allclose(sol.y_events, [np.asarray([[ 8.5808408e-20, -1.0000000e-02 ]])])
+    assert_(sol.success)
+    assert_equal(sol.status, 1)
 
 @pytest.mark.thread_unsafe
 def test_integration():
--- a/scipy/integrate/_ivp/common.py
+++ b/scipy/integrate/_ivp/common.py
@@ -65,7 +65,7 @@ def norm(x):
     return np.linalg.norm(x) / x.size ** 0.5
 
 
-def select_initial_step(fun, t0, y0, t_bound, 
+def select_initial_step(fun, t0, y0, t_bound,
                         max_step, f0, direction, order, rtol, atol):
     """Empirically select a good initial step.
 
@@ -80,7 +80,7 @@ def select_initial_step(fun, t0, y0, t_bound,
     y0 : ndarray, shape (n,)
         Initial value of the dependent variable.
     t_bound : float
-        End-point of integration interval; used to ensure that t0+step<=tbound 
+        End-point of integration interval; used to ensure that t0+step<=tbound
         and that fun is only evaluated in the interval [t0,tbound]
     max_step : float
         Maximum allowable step size.
@@ -112,7 +112,7 @@ def select_initial_step(fun, t0, y0, t_bound,
     interval_length = abs(t_bound - t0)
     if interval_length == 0.0:
         return 0.0
-    
+
     scale = atol + np.abs(y0) * rtol
     d0 = norm(y0 / scale)
     d1 = norm(f0 / scale)
--- a/scipy/integrate/_ivp/ivp.py
+++ b/scipy/integrate/_ivp/ivp.py
@@ -694,8 +694,15 @@ def solve_ivp(fun, t_span, y0, method='RK45', t_eval=None, dense_output=False,
             g = g_new
 
         if t_eval is None:
-            ts.append(t)
-            ys.append(y)
+            donot_append = (len(ts) > 1 and
+                            ts[-1] == t and
+                            dense_output)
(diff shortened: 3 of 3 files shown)


czgdp1807:
If the fix makes sense then I will enable the CI (by dropping my last commit here). Marking it as ready for review.

ericjwhitney:
> Let me know if the fix is making sense. :-)
> 
> @ericjwhitney

@czgdp1807  seems good to me. 👍 I had initially thought using `donot_append` might be bit wordy and the code could be simplified by just inverting the `if not` conditions.  But on second thought this version has a clear intent to anyone reading the code later on.  Cheers 👌

scipy/integrate/_ivp/tests/test_ivp.py
```diff
@@ -151,6 +151,25 @@ def compute_error(y, y_true, rtol, atol):
...
+    def hit_ground(t, y):
+        return y[0]
+
+    hit_ground.terminal = True
+    hit_ground.direction = -1
+
+    sol = solve_ivp(upward_cannon, [0, np.inf], [0, 0.01],
+                    max_step=0.05 * 0.001 / 9.80665,
+                    events=hit_ground, dense_output=True)
+    assert_allclose(sol.sol(0.01), np.asarray([-0.00039033, -0.08806632]),
```
j-bowhay, line 167: It would be good to know where the reference values are obtained from? Are they computed analytically?

ericjwhitney: For straight up-down ballistic motion without air resistance, the time to return to the starting point is given by $t=2u/g$ where $u$ is the initial velocity and $g$ is the gravitational acceleration.  For the two cases the analytical result is:

- $u=0.01$ gives $t=2 \times 0.01 / 9.80665 = 0.002039432$
- $u=0.001$ gives $t=2 \times 0.001 / 9.80665 = 0.000203943$

So the check on $t$ is correct for the given $u$.  At impact the $y$ conditions will be:

$$ \mathbf{y} = \begin{pmatrix} 
0 
\\ 
-u 
\end{pmatrix} $$

So the test seems pretty much correct here too, although the 'tiny' value of `8.5808408e-20` could probably just be converted to `0.0`, but it's not critical.  I'm reminded of course that generally tests should be designed to reproduce the error rather than check the results - but in this case we can do both. 🤷‍♂️

scipy/integrate/_ivp/tests/test_ivp.py
```diff
@@ -151,6 +151,25 @@ def compute_error(y, y_true, rtol, atol):
...
+    def upward_cannon(t, y):
+        return [y[1], -9.80665]
+
+    def hit_ground(t, y):
+        return y[0]
+
+    hit_ground.terminal = True
+    hit_ground.direction = -1
+
+    sol = solve_ivp(upward_cannon, [0, np.inf], [0, 0.01],
```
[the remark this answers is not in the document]
ericjwhitney: Yes I saw that too - but I don't think it really matters either way.  The only reason I used the same value in my example was to make the maximum timestep a direct multiple of the velocity.  Provided the error occurs either way its fine.

scipy/integrate/_ivp/tests/test_ivp.py
```diff
@@ -151,6 +151,25 @@ def compute_error(y, y_true, rtol, atol):
...
+                    max_step=0.05 * 0.001 / 9.80665,
+                    events=hit_ground, dense_output=True)
+    assert_allclose(sol.sol(0.01), np.asarray([-0.00039033, -0.08806632]),
+                    rtol=1e-5, atol=1e-8)
+    assert_allclose(sol.t_events, np.asarray([[0.00203943]]), rtol=1e-5, atol=1e-8)
+    assert_allclose(sol.y_events, [np.asarray([[ 8.5808408e-20, -1.0000000e-02 ]])])
```
ericjwhitney, line 170: Actually now that I tried the patch locally and ran the example given, it didn't pass... so yes I would suggest changing this line to simply `assert_allclose(sol.y_events, [np.asarray([[0.0, -0.01]])], atol=1e-9)`  👍

scipy/integrate/_ivp/tests/test_ivp.py
ericjwhitney: On re-reading, I'm not sure what the purpose of this line is as it checks 'after the event': `assert_allclose(sol.sol(0.01), np.asarray([-0.00039033, -0.08806632]), rtol=1e-5, atol=1e-8)`

scipy/integrate/_ivp/tests/test_ivp.py
```diff
@@ -151,6 +151,25 @@ def compute_error(y, y_true, rtol, atol):
...
+    assert_(sol.success)
```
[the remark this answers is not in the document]
czgdp1807: Done and resolving.

scipy/integrate/_ivp/tests/test_ivp.py
czgdp1807: Thank you @ericjwhitney for showing the calculations. <3.

scipy/integrate/_ivp/tests/test_ivp.py
czgdp1807: Done and therefore resolving.

czgdp1807:
I have enabled the CI now. Let me know if more changes need to be made. Thank you.

j-bowhay (approved):
Thanks both!

PR #12575, opened 2020-07-18
BUG: stats: fix multinomial.pmf NaNs when params sum > 1

mdhaber:
Closes gh-11860

This is one way to fix this bug ~~in a backwards compatible way~~: maintain original behavior when row sum < 1; normalize by sum of row when row sums > 1. The caveats are that:

- According to the old behavior, rows that sum > 1 would result in NaNs; now they don't. If a user were (for some reason) relying on the code to return NaNs (instead of a correct answer), this would break that code. I don't think this is a credible concern.
- The behavior is different depending on whether the row sums < 1 or > 1.

Another option would be to add a parameter, e.g. `normalization`, via which the user could opt-in to the new behavior.

I figure that most of the time this bug occurs because the sum is very slightly < 1 or > 1, in which case the two options will result in the same pmf values to reasonable precision. With that in mind, I started with the first option because it fixes the bug without requiring any user action.

Issue #11860, opened 2020-04-15
Occurrence of nan values when using multinomial.pmf from scipy.stats?


bunkerj:
Running a small experiment, I noticed that if the sum of the parameters that are passed to multinomial.pmf are even slightly above 1, then the return value is nan.

See the example below:
```python3
import numpy as np
from scipy.stats import multinomial as multi_s


def safe_multi(x, params):
    params_sum = params.sum()
    safe_params = params / params_sum if params_sum > 1 else params
    return multi_s.pmf(x, sum(x), safe_params)


params1 = np.array(
    [0.21310660657549002, 0.21310660657549002, 0.21310660657549002,
     2.8699968847179538e-06, 0.0023286820110742764, 2.8699968847179538e-06,
     0.0023286820110742764, 2.8699968847179538e-06, 2.8699968847179538e-06,
     0.0023258120141895593, 0.0016006555371205703, 0.0023258120141895593,
     0.0016006555371205703, 0.04333851102588555, 0.04333851102588555,
     0.04333851102588555, 0.04333851102588555, 0.04333851102588555,
     0.04333851102588555, 0.04333851102588555, 0.04333851102588555,
     0.0007251564770689873, 0.0007251564770689873, 7.377915317967555e-27,
     7.377915317967555e-27])

params2 = np.array(
    [0.3333333333333332, 0.3333333333333332, 0.3333333333333332,
     2.931077467598623e-93, 6.532951191692606e-25, 1.4080539652716124e-224,
     6.532951191692606e-25, 1.4080539652716124e-224, 1.4080539652716124e-224,
     6.532951191692606e-25, 6.532951191692606e-25, 6.532951191692606e-25,
     6.532951191692606e-25, 3.5127398105835854e-17, 3.5127398105835854e-17,
     3.5127398105835854e-17, 3.5127398105835854e-17, 3.5127398105835854e-17,
     3.5127398105835854e-17, 3.5127398105835854e-17, 3.5127398105835854e-17,
     7.860388790608641e-191, 7.860388790608641e-191, 2.931077467598623e-93,
     2.931077467598623e-93])

samples = np.array(
    [1.0, 1.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0,
     0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0])

[...]

PR #22585, opened 2025-02-24
MAINT: stats.multinomial: `FutureWarning` about normalization of `p`

### LINKED ISSUE #22565: BUG: stats.multinomial.pmf: inconsistent results?
### Describe your issue.

I came across something odd. With the same counts but in different order the multinomial.pmf gives a slightly different result. Since this could be used to determine a cdf and looping over quite a lot of different combinations, these minor differences do add up.

### Reproducing Code Example

```python
multinomial.pmf(x=[1, 2, 5, 7, 4], n=19, p=[0.2, 0.2, 0.2, 0.2, 0.2])
0.00021969062225510555

multinomial.pmf(x=[1, 2, 4, 5, 7], n=19, p=[0.2, 0.2, 0.2, 0.2, 0.2])
0.000219690622255104
```

### Error message

```shell
there's no error, just a difference
```

```shell
1.13.1 1.26.4 sys.version_info(major=3, minor=12, micro=7, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    detection method: pkgconfig
    found: true
    include directory: C:/Users/Peter/anaconda3/Library/include
    lib directory: C:/Users/Peter/anaconda3/Library/lib
    name: mkl-sdl
    pc file directory: C:\b\abs_efv75hqhju\croot\scipy_1717521501389\_h_env\Library\lib\pkgconfig
    version: '2023.1'
  lapack:
    detection method: pkgconfig
    found: true
    include directory: C:/Users/Peter/anaconda3/Library/include
    lib directory: C:/Users/Peter/anaconda3/Library/lib
    name: mkl-sdl
    pc file directory: C:\b\abs_efv75hqhju\croot\scipy_1717521501389\_h_env\Library\lib\pkgconfig
    version: '2023.1'
  pybind11:
    detection method: pkgconfig
    include directory: C:/Users/Peter/anaconda3/Library/include
    name: pybind11
    version: 2.12.0
Compilers:
  c:
    commands: clang-cl.exe
    linker: lld-link
    name: clang-cl
    version: 14.0.6
  c++:
    commands: clang-cl.exe
    linker: lld-link
    name: clang-cl
    version: 14.0.6
  cython:
    commands: cython
    linker: cython
    name: cython
    version: 3.0.10
  fortran:
    commands: ifort
    linker: xilink
    name: intel-cl
    version: 2021.6.0
  pythran:
    include directory: ..\..\_h_env\Lib\site-packages\pythran
    version: 0.15.0
Machine Information:
  build:
    cpu: x86_64
    endian: little
    family: x86_64
    system: windows
  cross-compiled: false
  host:
    cpu: x86_64
    endian: little
    family: x86_64
    system: windows
Python Information:
  path: C:\b\abs_efv75hqhju\croot\scipy_1717521501389\_h_env\python.exe
  version: '3.12'
```

mdhaber:
#### Reference issue
Closes gh-22565

The `multinomial` distribution currently changes the last element in rows of the user-provided `p` without warning to ensure that rows sum to 1.0.
```python3
import numpy as np
from scipy import stats
res = stats.multinomial.entropy(10, [0.25, 75])
ref = stats.multinomial.entropy(10, [0.25, 0.75])
np.testing.assert_equal(res, ref)  # passes, no warning
```
This is documented, but has led to issues like gh-12575 and gh-11860. This PR proposes to warn:
```python3
res = stats.multinomial.entropy(10, [0.25, 75])
# <ipython-input-2-5540b560d502>:3: FutureWarning: Some rows of `p` do not sum to 1.0 within tolerance of eps=1e-15. Currently, the last element of these rows is adjusted to compensate, but this condition will produce NaNs beginning in SciPy 1.18.0. Please ensure that rows of `p` sum to 1.0 to avoid futher disruption.
#   res = stats.multinomial.entropy(10, [0.25, 75])
```
and produce NaNs beginning in SciPy 1.18.0. This is consistent with the behavior of other distributions in which invalid shape parameters produce a NaN, e.g.
```python3
stats.binom(n=10, p=-1).entropy()  # nan
```

#### Additional information
Forum post [here](https://discuss.scientific-python.org/t/maint-stats-multinomial-futurewarning-about-normalization-of-p/1748).

The documentation currently says:
> Each element of `p` should be in the interval $[0, 1]$ and the elements should sum to `1`. If they do not sum to `1`, the last element of the `p` array is not used and is replaced with the remaining probability left over from the earlier elements.

Presumably "should" will change to "must" in 1.18.0 and the second sentence would be removed. How should we change it here?

In such cases:
- Mathematica emits a warning that the probabilities must sum to 1 and does not produce a numerical result.
- R normalizes `p` to sum to 1.0 by dividing by `p /= p.sum()`. This behavior was proposed in gh-12575, but not accepted.
- Matlab raises.

### DIFF
--- a/scipy/stats/_multivariate.py
+++ b/scipy/stats/_multivariate.py
@@ -3282,7 +3282,8 @@ class multinomial_gen(multi_rv_generic):
         """
         p = np.array(p, dtype=np.float64, copy=True)
         p_adjusted = 1. - p[..., :-1].sum(axis=-1)
-        i_adjusted = np.abs(p_adjusted) > eps
+        # only make adjustment when it's significant
+        i_adjusted = np.abs(1 - p.sum(axis=-1)) > eps
         p[i_adjusted, -1] = p_adjusted[i_adjusted]
 
         # true for bad p
--- a/scipy/stats/_multivariate.py
+++ b/scipy/stats/_multivariate.py
@@ -2,6 +2,7 @@
 # Author: Joris Vankerschaver 2013
 #
 import math
+import warnings
 import threading
 import numpy as np
 import scipy.linalg
@@ -3285,6 +3286,14 @@ class multinomial_gen(multi_rv_generic):
         # only make adjustment when it's significant
         i_adjusted = np.abs(1 - p.sum(axis=-1)) > eps
         p[i_adjusted, -1] = p_adjusted[i_adjusted]
+
+        if np.any(i_adjusted):
+            message = ("Some rows of `p` do not sum to 1.0 within tolerance of "
+                       f"{eps=}. Currently, the last element of these rows is adjusted "
+                       "to compensate, but this condition will produce NaNs "
+                       "beginning in SciPy 1.18.0. Please ensure that rows of `p` sum "
+                       "to 1.0 to avoid futher disruption.")
+            warnings.warn(message, FutureWarning, stacklevel=3)
 
         # true for bad p
         pcond = np.any(p < 0, axis=-1)
--- a/scipy/stats/tests/test_multivariate.py
+++ b/scipy/stats/tests/test_multivariate.py
@@ -2056,6 +2056,16 @@ class TestMultinomial:
         logpmf = multinomial.logpmf(x, n, p)
         assert np.isfinite(logpmf)
 
+    def test_gh_22565(self):
+        # Same issue as gh-11860 above, essentially, but the original
+        # fix didn't completely solve the problem.
+        n = 19
+        p = [0.2, 0.2, 0.2, 0.2, 0.2]
+        res1 = multinomial.pmf(x=[1, 2, 5, 7, 4], n=n, p=p)
+        res2 = multinomial.pmf(x=[1, 2, 4, 5, 7], n=n, p=p)
+        np.testing.assert_allclose(res1, res2, rtol=1e-15)
+
+
 class TestInvwishart:
     def test_frozen(self):
         # Test that the frozen and non-frozen inverse Wishart gives the same
--- a/scipy/stats/tests/test_multivariate.py
+++ b/scipy/stats/tests/test_multivariate.py
@@ -1929,10 +1929,17 @@ class TestMultinomial:
     @pytest.mark.parametrize("n", [0, 3])
     def test_rvs_np(self, n):
         # test that .rvs agrees w/numpy
-        sc_rvs = multinomial.rvs(n, [1/4.]*3, size=7, random_state=123)
-        rndm = np.random.RandomState(123)
-        np_rvs = rndm.multinomial(n, [1/4.]*3, size=7)
-        assert_equal(sc_rvs, np_rvs)
+        message = "Some rows of `p` do not sum to 1.0 within..."
+        with pytest.warns(FutureWarning, match=message):
+            rndm = np.random.RandomState(123)
+            sc_rvs = multinomial.rvs(n, [1/4.]*3, size=7, random_state=123)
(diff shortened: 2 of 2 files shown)


ev-br (approved):
LGTM.
The whole story with compensating the last element is made to match numpy's `random.multinomial` treatment of its `pvals` argument, https://numpy.org/doc/2.0/reference/random/generated/numpy.random.Generator.multinomial.html#numpy.random.Generator.multinomial

Emitting a warning---with all caveats of python warnings---is OK if it helps clearing user confusion.

jakevdp:
We're finding that this change makes it impossible to use `multinomial.pdf` with float32 inputs. For example:
```python
import numpy as np
from scipy.stats import multinomial

rng = np.random.default_rng()
x = rng.normal(size=(100,)).astype('float32')
n = x.sum(dtype=x.dtype)
p = rng.normal(size=(100,)).astype('float32')
p / p.sum()  # explicitly normalized

multinomial.logpmf(x, n, p) 
# FutureWarning: Some rows of `p` do not sum to 1.0 within tolerance of eps=1e-15...
```
Despite explicitly normalizing the input `p` values, we get the warning because float32 precision does not allow for a tolerance of 1e-15.

Would it be possible to adjust this tolerance depending on the input dtype? If not, we adjust the documentation to mention that non-float64 inputs are essentially not allowed any more.

mdhaber:
@jakevdp I opened gh-22632 to address this, but could you answer @ev-br's question https://github.com/scipy/scipy/pull/22585#issuecomment-2700924730?

jakevdp:
> How did it work previously? https://github.com/scipy/scipy/blob/main/scipy/stats/_multivariate.py#L3284 is
> `p = np.array(p, dtype=np.float64, copy=True)`, does it happen to be float32 instead?

It worked previously because before this PR, the correction was done silently and no `FutureWarning` was issued.

scipy/stats/_multivariate.py
```diff
@@ -3282,9 +3283,18 @@ def _process_parameters(self, n, p, eps=1e-15):
...
+        # only make adjustment when it's significant
+        i_adjusted = np.abs(1 - p.sum(axis=-1)) > eps
         p[i_adjusted, -1] = p_adjusted[i_adjusted]
 
+        if np.any(i_adjusted):
+            message = ("Some rows of `p` do not sum to 1.0 within tolerance of "
+                       f"{eps=}. Currently, the last element of these rows is adjusted "
+                       "to compensate, but this condition will produce NaNs "
+                       "beginning in SciPy 1.18.0. Please ensure that rows of `p` sum "
+                       "to 1.0 to avoid futher disruption.")
```
j-bowhay, line 3295: @mdhaber are you ok to tackle the follow up?

mdhaber: Sure I can do that today

mdhaber: Ugh.

For all methods other than `rvs`, this is very easy. Any one of several strategies would work, like `p[i_adjusted, -1] = np.nan`, or probably better, use the condition of `i_adjusted` in the `pcond` mask below. Those get set to NaN.

For NumPy, these don't work. NumPy *raises* if there is a NaN in `p` or if a negative value is required to adjust `p` to sum to 1.0. I thought "OK, we'll just continue to perform the adjustment as before, but also adjust `pcond` to NaN-out the results where appropriate".

But we can't NaN-out the result: `rvs` currently returns integers, so returning NaN would require changing the dtype.

Given that `rvs` is already raising where other methods would return NaN (e.g. if there is a NaN in `p` or `n < 0`), shall we make good on the warning for other methods but let `rvs` raise?

*Update: done in gh-24868.*

Issue #22241, opened 2025-01-04
DOC: Interactive notebooks in stats tutorials are not working in version 1.15 docs.


steppi:
### Issue with current documentation:

It appears that the interactive notebooks in the stats tutorials are not working in the new version 1.15 docs. The kernel selection menu does not list any available kernels. This was first pointed out to me by @mdhaber. The devdocs are still working as expected. See https://docs.scipy.org/doc/scipy/tutorial/stats/hypothesis_chi2_contingency.html for an example of a notebook that isn't working, and https://scipy.github.io/devdocs/tutorial/stats/hypothesis_chi2_contingency.html for the corresponding working notebook in the devdocs.

The notebooks work properly when I build them locally off the `v1.15.0` branch. In the dev console on Firefox, I see an error for the stable docs which doesn't appear for the dev docs, message pasted in the additional context section below.

### Idea or request for content:

@agriyakhetarpal, do you know what might be going on? I remember that you're working on documentation for how to change the pyodide kernel version for the built documentation, and I think that may be helpful here. As far as I know, I don't think there is anything special @tylerjereddy needs to do during the release process to make these work.

### Additional context (e.g. screenshots, GIFs)

No kernel found in the kernel selection menu:

![no_kernel](https://github.com/user-attachments/assets/bc126466-da15-4df2-bad5-663da7ec17ee)

The error I'm seeing in the dev console

```
Uncaught (in promise) TypeError: e is undefined
    Ce toc.js:513
    s toc.js:421
    s toc.js:419
    c toc.js:480
    d index.es6.js:555
    emit index.es6.js:513
    emit index.es6.js:112
    _onCellInViewportChanged widget.js:652
    d index.es6.js:555
    emit index.es6.js:513
    emit index.es6.js:112
    set inViewport widget.js:164
    attachWidget windowing.js:198
    insertWidget index.es6.js:2516
    _updateForDeferMode widget.js:758
    _runOnIdleTime widget.js:727
    _runOnIdleTime widget.js:743
```

Here's the relevant code from `toc.js`. Line 513 is the one with `if (heading.type === Cell.HeadingType.Markdown) {`

```javascript
/**
 * Get the element id for an heading
 * @param heading Heading
 * @param parser The markdownparser
 * @returns The element id
 */
export async function getIdForHeading(heading, parser, sanitizer) {
    let elementId = null;
    if (heading.type === Cell.HeadingType.Markdown) {
[...]

PR #22587, opened 2025-02-24
BUG: Pin jupyterlite-sphinx to >= 0.19.1

steppi:
#### Reference issue

Reference issue: https://github.com/scipy/scipy/issues/22241., but merging this is not enough to close that, since the docs will also need to be rebuilt and re-uploaded.

This PR pins `jupyterlite-sphinx` to greater than or equal to version `0.19.1` to incorporate the fix https://github.com/jupyterlite/jupyterlite-sphinx/pull/271, which should fix the problem that caused  #22241. The issue was that `jupyterlite-sphinx` called `jupyter` using `subprocess.run`, and in the `make dist` doc workflow, the relevant virtual environment was not active in this `subprocess`. 

#### Additional information

@tylerjereddy, I've marked this as a backport candidate. If there is a `0.15.3`, I think there's minimal risk in backporting the dependency bump here.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -48,7 +48,7 @@ dependencies:
   - jupytext
   - myst-nb>=1.2.0
   - linkify-it-py
-  - jupyterlite-sphinx>=0.17.1
+  - jupyterlite-sphinx>=0.19.1
   - jupyterlite-pyodide-kernel
   # Some optional test dependencies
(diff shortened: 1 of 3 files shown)


rgommers (approved):
Thanks for tracking that down @steppi!

PR #22588, opened 2025-02-24
BUG/BLD: xsf: force defining the mdspan parenthesis operator even with C++ >= 23

Romain-Geissler-1A:
This allows to keep the same code, independently of the compiler C++ dialect being used. The default behavior of kokkos mdspan is the following:
 - if __cpp_multidimensional_subscript is defined (ie C++ >= 23), then enable by default the square bracket operator to access elements, and disable by default the parenthesis operator.
 - otherwise (C++ < 23), disable the square bracket operator which cannot work, and replace it by the parenthesis operator instead.

The current xsf code always uses the parenthesis operator. To avoid having to always write the two possibilities, force the kokkos mdspan implementation to always define the parenthesis operator for now.

If this fix is accepted, I guess it is worth backporting it in the 1.15 release branch (as we do hit this issue on our side with release 1.15.2).

### DIFF
--- a/scipy/special/xsf/numpy.h
+++ b/scipy/special/xsf/numpy.h
@@ -19,6 +19,9 @@
 
 #include "dual.h"
 #include "error.h"
+// Force defining the parenthesis operator even when compiling with a compiler
+// defaulting to C++ >= 23.
+#define MDSPAN_USE_PAREN_OPERATOR 1
 #include "third_party/kokkos/mdspan.hpp"
 
 /* PyUFunc_getfperr gets bits for current floating point error (fpe) status codes so we


lucascolley:
cc @steppi

steppi:
Thanks @Romain-Geissler-1A.  SciPy's Meson build configuration sets the language standard to C++17, so I think we'd just assumed this wouldn't be built with later versions of C++ for the time being. This looks good to me, and will without question be needed when `xsf` becomes a separate library, although I wonder whether it makes more sense  to define `MDSPAN_USE_PAREN_OPERATOR` in `config.h` just so that it doesn't accidentally get left out if `mdspan` gets used in another header in `xsf` in the future. I think it's fine to leave it where you put it for now.

Out of curiosity, are you building SciPy with C++>=23 or are you using `xsf` outside of SciPy by any chance? If the latter, all that remains before we finish separating out https://github.com/scipy/xsf, is setting up proper tests and CI, which is nearly finished. You've helped me realize that we'll want to test with multiple language standards.

steppi (approved):
This looks good to me. I think the risk in backporting is minimal. I guess `C++17` is just the default in the Meson build

https://github.com/scipy/scipy/blob/f7527ee40607c8ccde2269e6b24a5bc6018549da/meson.build#L11

the documentation mentions the possibility of changing the language level here, https://docs.scipy.org/doc/scipy/building/compilers_and_options.html#adding-a-custom-compiler-or-linker-flag, and the [Toolchain Roadmap](https://docs.scipy.org/doc/scipy/dev/toolchain.html) only mentions minimum supported language standards, so I think it's fair to call this a bug.

Romain-Geissler-1A:
> although I wonder whether it makes more sense to define MDSPAN_USE_PAREN_OPERATOR in config.h

I also wondered (I should have mentioned it in my first message) about this. Actually I also wondered if you wanted to abandon the vendored kokkos library for people who do have already an `<mdspan>` implementation in their toolchain (it's not my case as I use gcc where even in the latest gcc 15 trunk this isn't implemented), with some logic like:

```
#ifdef __cpp_multidimensional_subscript
#include <mdspan>
#else
#include "third_party/kokkos/mdspan.hpp"
#endif
```

However doing that means that then inside XSF, all access to the index have to also handle the call by bracket vs call by parenthesis issue, since the C++23 mdspan implementation defines only the bracket operator. So you would end up defining an index access macro and use it at all call site, effectively "ugyfying" the code until C++23 becomes the minimum require standard, not sure you actually want this for now.

> Out of curiosity, are you building SciPy with C++>=23 or are you using xsf outside of SciPy by any chance? If the latter, all that remains before we finish separating out https://github.com/scipy/xsf, is setting up proper tests and CI, which is nearly finished. You've helped me realize that we'll want to test with multiple language standards.

We do build our own GNU toolchains and we do build ourselves all our C/C++/python stack with these gnu toolchains. Each toolchain we have is bound to a given C/C++ standard as when mixing .o/.a files coming from different libraries built with different dialects, especially with recent standards still considered experimental by gcc, you might see here and there some subtle source compability & ABI differences leading to problem when everything is compiledlinked together. We do rebuilt our python dependencies (so among other things, scipy) via `pip wheel`, but before calling pip we explicitly set `C/CXX/LDFLAGS` to include our common C/C++ dialects. It's quite uncommon I agree, that's why we were the only ones hitting this very issue, but this allows to find bugs maintainers of libraries would have hit 5 years in the future ;)

steppi:
> However doing that means that then inside XSF, all access to the index have to also handle the call by bracket vs call by parenthesis issue, since the C++23 mdspan implementation defines only the bracket operator. So you would end up defining an index access macro and use it at all call site, effectively "ugyfying" the code until C++23 becomes the minimum required standard, not sure you actually want this for now.

I think the way I'd want to do it is to change `xsf`'s vendored version of `mdspan` to not inject into `std` like this

https://github.com/scipy/scipy/blob/b9751d5d1ef0ef76abb19dbd5f7e919e84cf3e63/scipy/special/xsf/third_party/kokkos/mdspan.hpp#L25

which is undefined behavior anyway, but I thought it was OK for @izaid do it as a temporary expedient in SciPy. Then use conditional compilation so that the `mdspan` in `xsf::std` or whatever we want to call it would either use the vendored `kokkos`  `mdspan` or a derived class of the actual `std::mdspan` which adds `operator()`. This seems like it can wait though.

> We do build our own GNU toolchains and we do build ourselves all our C/C++/python stack with these gnu toolchains. Each toolchain we have is bound to a given C/C++ standard as when mixing .o/.a files coming from different libraries built with different dialects, especially with recent standards still considered experimental by gcc, you might see here and there some subtle source compability & ABI differences leading to problem when everything is compiled/linked together. We do rebuilt our python dependencies (so among other things, scipy) via `pip wheel`, but before calling pip we explicitly set `C/CXX/LDFLAGS` to include our common C/C++ dialects. It's quite uncommon I agree, that's why we were the only ones hitting this very issue, but this allows to find bugs maintainers of libraries would have hit 5 years in the future ;)

That's pretty cool. Thanks for the 5 year early fix!

PR #22564, opened 2025-02-20
TST: add missing custom markers to pytest.ini

rgommers:
This is needed to avoid warnings when running tests on an installed `scipy` with `pytest --pyargs scipy`.

PR #22593, opened 2025-02-25
TST: add one more missing custom marker (`fail_slow`) to pytest.ini

rgommers:
This is a follow-up to PR gh-22564. Avoids verbose warnings from Pytest.


tylerjereddy (approved):
thanks Ralf

PR #22598, opened 2025-02-26
MAINT: stats.bootstrap: broadcast like other stats functions

mdhaber:
#### Reference issue
gh-20858

This PR implements the change discussed in gh-20858. I see that the change didn't get highlighted explicitly in the release notes (although the PR is in the change log), but it has been in the rendered documentation, there was a `FutureWarning`, and the impact is probably small - it only makes a difference when a multi-sample statistic is bootstrapped in batch and the arrays have different numbers of dimensions.


tupui (approved):
LGTM 👍

PR #22602, opened 2025-02-28
DOC: stats.pearsonr: add tutorial

mdhaber:
#### Reference issue
gh-16191

Adds a tutorial for `scipy.stats.pearsonr` that demonstrates how the statistic and p-value can be interpreted.

For comparison, see tutorial for [`kendalltau`](https://docs.scipy.org/doc/scipy/tutorial/stats/hypothesis_kendalltau.html#hypothesis-kendalltau).


doc/source/tutorial/stats/hypothesis_pearsonr.md
```diff
@@ -0,0 +1,157 @@
...
+ax.set_xlim(0.85, 1)
+ax.set_ylim(0, 0.1)
+plt.show()
+```
+
+```{code-cell} ipython3
+res.pvalue  # one-sided p-value; half of the two-sided p-value
+```
+
+Note that the beta distribution is the exact null distribution for samples of any size under the null hypothesis that the samples were drawn from a normal distribution. We can convince ourselves of this by computing a Monte Carlo null distribution: explicitly drawing samples from independent normal distributions and computing Pearson's statistic for each pair.
```
mdhaber, line 122:
```suggestion
Note that the beta distribution is the exact null distribution for samples of any size under this null hypothesis. We can convince ourselves of this by computing a Monte Carlo null distribution: explicitly drawing samples from independent normal distributions and computing Pearson's statistic for each pair.
```

doc/source/tutorial/stats/hypothesis_pearsonr.md
```diff
@@ -0,0 +1,157 @@
...
+
+fig, ax = plt.subplots(figsize=(8, 5))
+plot(ax)
+ax.hist(ref.null_distribution, np.linspace(-1, 1, 26), density=True)
+ax.legend(['exact null distribution (independent, normally-distributed observations)',
+           f'Monte Carlo null distribution \n({len(ref.null_distribution)} permutations)'])
+plt.show()
+```
+
+This is often an reasonable null hypothesis to test, but in other cases, it may be more conservative to perform a permutation test: Under the null hypothesis that total collagen and free proline are independent (but not necessarily normally distributed), each of the free proline measurements were equally likely to have been observed with any of the total collagen measurements. Therefore, we can form an exact null distribution by calculating the statistic under each possible pairing of elements between `x` and `y`. This is the null distribution used when we provide `pearsonr` with `method=stats.PermutationMethod()`.
```
mdhaber, line 139:
```suggestion
This is often an reasonable null hypothesis to test, but in other cases, it may be more conservative to perform a permutation test: Under the null hypothesis that total collagen and free proline are independent (but not necessarily normally distributed), each of the free proline measurements are equally likely to be observed with any of the total collagen measurements. Therefore, we can form an exact null distribution by calculating the statistic under each possible pairing of elements between `x` and `y`. This is the null distribution used when we provide `pearsonr` with `method=stats.PermutationMethod()`.
```

mdhaber:
And looks like I need to add this to the ToC.

doc/source/tutorial/stats/hypothesis_pearsonr.md
```diff
@@ -0,0 +1,167 @@
...
+```{code-cell} ipython3
+import matplotlib.pyplot as plt
+fig, ax = plt.subplots(figsize=(8, 5))
+ax.plot(x, y, '.')
+ax.set_xlabel("total collagen (mg/g)")
+ax.set_ylabel("free proline (μ mole/g)")
+plt.show()
+```
+
+These data were analyzed in [^2] using Spearman’s correlation coefficient, a statistic sensitive to monotonic correlation between the samples. Here, we will analyze the data using Pearson's correlation coefficient, which is sensitive to linear correlation.
```
melissawm, line 48: Not mandatory, but just in case, if you want to cite the actual pearsonr function this is how you can do it:

```suggestion
These data were analyzed in [^2] using Spearman’s correlation coefficient, a statistic sensitive to monotonic correlation between the samples. Here, we will analyze the data using Pearson's correlation coefficient (`{class}`scipy.stats.pearsonr`) which is sensitive to linear correlation.
```

As a general rule I've been trying to have at least one link from tutorial to reference, and one from reference to tutorial (those are the See Also links). That makes both the tutorial and the reference docs more visible.

melissawm (approved):
Looks great, thanks @mdhaber !

melissawm:
I'm happy to merge when green, unless you are waiting for more reviews!

PR #22603, opened 2025-02-28
MAINT: _lib: bump version array_api_compat to 1.11

ev-br:
Bump vendored `array-api-compat` package to 1.11, so that we can use internally array api features from the 2024.12 revision.


lucascolley (approved):
thx!

Issue #22178, opened 2024-12-24
DOC: stats: 5 build warnings locally


lucascolley:
### Issue with current documentation:

```
reading sources... [100%] tutorial/stats/continuous_loguniform .. tutorial/thread_safety
/Users/lucascolley/programming/pixi-dev-scipystack/scipy/scipy/build-install/lib/python3.11/site-packages/scipy/stats/__init__.py:docstring of scipy.stats:47: WARNING: autosummary: stub file not found 'scipy.stats.uniform'. Check your autosummary_generate setting.
/Users/lucascolley/programming/pixi-dev-scipystack/scipy/scipy/build-install/lib/python3.11/site-packages/scipy/signal/_ltisys.py:docstring of scipy.signal._ltisys.place_poles:127: WARNING: Exception occurred in plotting scipy-signal-place_poles-1
 from /Users/lucascolley/programming/pixi-dev-scipystack/scipy/scipy/doc/source/reference/generated/scipy.signal.place_poles.rst:
Traceback (most recent call last):
  File "/Users/lucascolley/programming/pixi-dev-scipystack/scipy/.pixi/envs/doc/lib/python3.11/site-packages/matplotlib/sphinxext/plot_directive.py", line 552, in _run_code
    exec(code, ns)
  File "<string>", line 46, in <module>
  File "<__array_function__ internals>", line 200, in det
  File "/Users/lucascolley/programming/pixi-dev-scipystack/scipy/.pixi/envs/doc/lib/python3.11/site-packages/numpy/linalg/linalg.py", line 2139, in det
    r = _umath_linalg.det(a, signature=signature)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeWarning: divide by zero encountered in det
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
copying assets... 
copying static files... done
copying extra files... done
copying assets: done
/Users/lucascolley/programming/pixi-dev-scipystack/scipy/scipy/build-install/lib/python3.11/site-packages/scipy/stats/__init__.py:docstring of scipy.stats:479:<autosummary>:1: WARNING: py:obj reference target not found: scipy.stats.Uniform
/Users/lucascolley/programming/pixi-dev-scipystack/scipy/scipy/doc/source/release/1.15.0-notes.rst:254: WARNING: py:obj reference target not found: scipy.stats.Uniform
[...]

PR #22606, opened 2025-02-28
DOC: Ignore dict subclass docstring warning

melissawm:
#### Reference issue
Addresses https://github.com/scipy/scipy/issues/22178#issuecomment-2578136965

Some subclasses of dicts raise a Sphinx warning when the docs are built with Python 3.12. The warning is `WARNING: py:class reference target not found: None.  Update D from mapping/iterable E and F.`

#### Additional information
Doesn't affect CI since it's still running Python 3.11.


j-bowhay (approved):
Thanks!

PR #22607, opened 2025-02-28
MAINT: special.logsumexp: improve behavior with complex infinities

### LINKED ISSUE #22601: BUG: logsumexp is inconsistent with naiive numpy counterpart
### Describe your issue.

This is an edge case, but there is a difference between scipy and naiive numpy implementation of `logsumexp`. The following code produces different outputs:

### Reproducing Code Example

```python
import numpy as np
from numpy import inf
from scipy.special import logsumexp

def check(x):
    y_scipy = logsumexp(x, 1)
    y_numpy = np.log(np.sum(np.exp(x), 1))

    print("numpy:", y_numpy)
    print("scipy:", y_scipy)

x = np.array([[  inf + 0.0000j, -0.9116 + 0.0000j],
              [ -inf + 0.0000j,    -inf + 0.0000j]])
check(x)

x = np.array([[  inf + 0.0000j, -0.9116 + 0.0000j],
              [ -inf + 0.0000j,    -inf + 0.7533j]])
check(x)

x = np.array([[  inf + 0.0000j, -0.9116 + 0.7533j],
              [ -inf + 0.0000j,    -inf + 0.0000j]])
check(x)

x = np.array([[  inf + 0.0000j, -0.9116 + 0.7533j],
              [ -inf + 0.0000j,    -inf + 0.7533j]])
```

### Error message

```shell
Scipy version 1.5.0 and newer prints:

numpy: [ inf+0.j -inf+0.j]
scipy: [ inf+0.j -inf+0.j]
numpy: [ inf+0.j -inf+0.j]
scipy: [ inf+0.j     -inf+0.7533j]
numpy: [ inf+0.j -inf+0.j]
scipy: [ inf+0.20946886j -inf+0.j        ]


Scipy version 1.14.1 and older prints:

numpy: [ inf+0.j -inf+0.j]
scipy: [ inf+0.j -inf+0.j]
numpy: [ inf+0.j -inf+0.j]
scipy: [ inf+0.j -inf+0.j]
numpy: [ inf+0.j -inf+0.j]
scipy: [ inf+0.j -inf+0.j]
```

```shell
Scipy v1.5.0 and newer.
```

mdhaber:
#### Reference issue
Closes gh-22601

gh-22601 reported that `special.logsumexp` does not always return the correct complex infinity. (Although there are some differences between the old behavior and the new behavior, it was never correct in general.) This improves that behavior.

#### Additional information
Skipped a test with torch, ultimately because
```python3
import torch as xp
a = xp.asarray([1 + 0j])
b = xp.asarray([xp.inf + 0j])
print(a - b, -b + a)
# tensor([-inf+nanj]) tensor([-inf+0.j])
```
Should this be reported?

---

I'm not sure if I wrote the math out somewhere before, so here is what the current implementation does:

```math
\texttt{logsumexp}(x, b) = \log \left( b_1 \exp(x_1) + \cdots + b_n \exp(x_n) \right)
```

We want to rearrange things to avoid overflow, so we separate this as:

```math
\texttt{logsumexp}(x, b) = \log \left( m \exp(x') + \sum_{i : x_i \neq x'} \exp(x_i)  \right)
```

where $x'$ is the element of $x_i$ with maximum real component and 

```math
m = \sum_{i : x_i = x'} b_i
```

Now we can divide and multiply by $m \exp(x')$ to get:

```math
\texttt{logsumexp}(x, b) = \log \left( \left(  1 + \frac{1}{m} \sum_{i : x_i \neq x'} \frac{\exp(x_i)}{\exp(x')} \right) m \exp(x') \right)
```

Simplifying:
```math
\texttt{logsumexp}(x, b) = \texttt{log1p} \left( \frac{1}{m} \sum_{i : x_i \neq x'} \exp(x_i - x') \right) + \log(m) + x'
```

The issue with this that led to gh-22601 is that when $x'$ has infinite real part, $x_i - x'$ can break down. There are four cases:
- Only $x'$ has infinite (positive or negative) real part. No problem because $x_i - x'$ will just have negative infinite real part and the exponentiation will turn this to zero.
- $x'$ has infinite positive real part, and at least one of $x_i$ has negative real part. Also not a problem, because again, $x_i - x'$ will just have negative infinite real part; no NaNs.
- $x'$ and at least one of the $x_i$ has infinite negative real part. Really, this requires that *all* the $x_i$ also have negative infinite real part because $x'$ has the maximum real part. The correct result of `logsumexp` in this case is the logarithm of `0`. This will have negative real part, and I have trouble imagining a case in which the phase matters. In the past, I think we've always returned `-inf + 0j` in this case. What we do now is avoid subtracting $x_i - x'$. The sum of exponentials ends up as `0 + 0j` rather than NaN, and the final result is $\log(m) + x'$. This has the property that `logsumexp([-inf + y j]*m) = -inf + y j` rather than `-inf + 0j`, which doesn't make *less* sense to me than the old behavior. 
- $x'$ and at least one of the $x_i$ has infinite positive real part. Previously, we made an adjustment to prevent the subtraction from producing NaNs, figuring (correctly for real input) that the final $+ x'$ would make the result positive infinity as it should be. What this didn't account for was the complex case, for which we can have complex infinities that are aligned, anti-aligned, or somewhere in-between. The correct result depends on which is the case, and we have never correctly distinguished between them in general, so now we let the subtraction produce NaNs, and the result is `nan + nan j`. It *might* be better to produce `inf + nanj`, but even a light breeze will turn this unicorn into `nan + nanj`, so we don't bother.


mdhaber:
Failure is due to a warning emitted by the comparison:

```python3
import numpy as np
np.testing.assert_allclose(np.inf + 1j, np.inf + 1j)
# /usr/local/lib/python3.11/dist-packages/numpy/core/numeric.py:2358: RuntimeWarning: invalid value encountered in multiply
#   x = x * ones_like(cond)
# /usr/local/lib/python3.11/dist-packages/numpy/core/numeric.py:2359: RuntimeWarning: invalid value encountered in multiply
#   y = y * ones_like(cond)
```
Apparently this is fixed in NumPy 2.0+. Shall I skip the test pre NumPy 2.0?

steppi:
What I suggest is taking into consideration the relevant C99 standards for complex `log` and `exp` and trying to be consistent with them. It seems to be the path of minimal complaint.

For case 3) above these are:

```
log(0 + 0j) = (-inf + 0j)
log(0 - 0j) = (-inf - 0j)
log(-0 + 0j) = (-inf + πj)
log(-0 - 0j) = (-inf - πj)
```

```
exp(-inf + yj) = 0 * cis(y), y finite
exp(-inf + yj) = 0 (signs undetermined), otherwise
```

where `cis(y) = cos(y) + sin(y)*1j`

I think we get the most consistent behavior automatically by directly using `xp.log(xp.sum(b*xp.exp(a)))` when $x^'$ has negative infinite real part. Similarly for case 4) where the relevant parts of the standard say:

```
exp(inf + yj) = inf * cis(y), y finite
exp(inf + yj) = ±inf + nanj (sign of real part undetermined), otherwise
```

```
log(inf + infj) = (inf + π/4j)
log(inf - infj) = (inf - π/4j)
log(-inf + infj) = (inf + 3π/4j)
log(-inf - infj) = (inf - 3π/4j)
log(±inf + nanj) = (inf + nanj)
```

We'd lose a little efficiency, but I guess we could do something like the following:

```python
    cond = xp.logical_or(xp.isfinite(a_max), xp_real(a_max) > 0)
    shift = xp.where(cond, a_max, xp.asarray(0, dtype=a_max.dtype))
    undo_shift = xp.where(cond, xp.log(m) + a_max, xp.asarray(0, dtype=a_max.dtype))
```

I don't think anyone can complain if we do this, and it should only need a short comment to justify.

scipy/special/_logsumexp.py
```diff
@@ -203,9 +203,24 @@ def _logsumexp(a, b, axis, return_sign, xp):
...
+    # the specifics. The logic for this is not implemented, and previous implementations
+    # of `logsumexp` have returned garbage phase for multiple complex infinities, so
+    # returning `nan + nan*j` is an improvement over a result with incorrect phase.
+    # For negative infinities, this can only occur if *all* elements have negative
+    # infinite real part, so the result is the logarithm of `0`, and the phase is
+    # meaningless. The value returned by `logsumexp` will have the same phase as
+    # `a_max`, which should be fine, since this function has never had a principled
+    # phase convention for this case. Perhaps NaN phase would be best, but we save
+    # that for another day.
+    shift = xp.where(xp.logical_or(xp.isfinite(a_max), xp_real(a_max) > 0),
```
steppi, line 222:
```suggestion
    shift = xp.where(xp.logical_or(xp.isfinite(a_max), xp.real(a_max) > 0),
```

I missed this one when resolving the merge conflicts.

steppi (approved):
I'm happy with this. I'll merge if nothing unexpected happens in CI.

steppi:
I guess these shouldn't have been surprises. But the new failures are due to warnings triggered in the "unchosen" branch of `xp.where` in various places.

scipy/special/_logsumexp.py
```diff
@@ -110,8 +110,25 @@ def logsumexp(a, axis=None, b=None, keepdims=False, return_sign=False):
     axis = tuple(range(a.ndim)) if axis is None else axis
 
     if xp_size(a) != 0:
+        with np.errstate(over='ignore'):
```
steppi, line 113:
```suggestion
        with np.errstate(divide='ignore', invalid='ignore', over='ignore'):
```

I think this fix the test failures.

steppi:
I'm not sure why the free-threaded test is hanging indefinitely, but it's clear there were no related failures. In this goes

steppi:
Thanks @mdhaber!

PR #22608, opened 2025-02-28
BUG: ndimage.median_filter: fix segfault when using `mode='mirror'`

### LINKED ISSUE #22586: BUG: ndimage.median_filter: additional hard crashes
Following-up on https://github.com/scipy/scipy/issues/22250#issuecomment-2678376069, it isn't too surprising we still have problems given my residual doubts noted in gh-22402, though to be fair this is for a different `mode` value than before.

Given the increasing cost of crash reports around `median_filter` with 1D input (related to the new/recent functionality there), I went ahead and wrote a more thorough `hypothesis` property-based test (also on my fork feature branch `treddy_median_filter_property_crash_guard`) that indeed easily reproduces crashing behavior on x86_64 Linux (see below).

It may take me a while to patch this second round of crashing (I'm a little swamped, but should get to it), but I'd be tempted to go this route to avoid any additional surprises, at least as far as hard crashes go (I might even be tempted to increase the max number of examples `hypothesis` uses and mark the test `slow` or something like that).

```diff
--- a/scipy/ndimage/tests/test_filters.py
+++ b/scipy/ndimage/tests/test_filters.py
@@ -6,6 +6,9 @@ import re
 import numpy as np
 import pytest
 from numpy.testing import suppress_warnings, assert_allclose, assert_array_equal
+from hypothesis import strategies as st
+from hypothesis import given
+import hypothesis.extra.numpy as npst
 from pytest import raises as assert_raises
 from scipy import ndimage
 from scipy._lib._array_api import (
@@ -2858,3 +2861,16 @@ def test_gh_22333():
     expected = [58, 67, 87, 108, 163, 108, 108, 108, 87]
     actual = ndimage.median_filter(x, size=9, mode='constant')
     assert_array_equal(actual, expected)
+
+
+@given(x=npst.arrays(dtype=np.float64,
+                     shape=st.integers(min_value=1, max_value=1000)),
+       size=st.integers(min_value=1, max_value=50),
+       mode=st.sampled_from(["constant", "mirror", "wrap", "reflect",
+                             "nearest"]),
+      )
+def test_gh_22250_crash_property(x, size, mode):
+    # property-based test for median_filter resilience to hard crashing
+    # after additional report at:
+    # https://github.com/scipy/scipy/issues/22250#issuecomment-2678376069
+    ndimage.median_filter(x, size=size, mode=mode)
```

aiudirog:
#### Reference issue
Closes gh-22586

Fixed an off-by-one error in the 1D optimized implementation of `ndimage.median_filter()` when `mode='mirror'` is selected, which was resulting in segfaults down the line.

#### Additional information
- Added a test case that compares the 1D implementation to the general ND implementation
  - This test also reliably reproduced the segfault
- Updated the initialization of the working `data` array to use the C++ zerod syntax: `new T[len]();`

Please let me know if you need anything else!

### DIFF
--- a/scipy/ndimage/src/_rank_filter_1d.cpp
+++ b/scipy/ndimage/src/_rank_filter_1d.cpp
@@ -155,6 +155,7 @@ void _rank_filter(T *in_arr, int rank, int arr_len, int win_len, T *out_arr,
                   int mode, T cval, int origin) {
   int i, arr_len_thresh, lim = (win_len - 1) / 2 - origin;
   int lim2 = arr_len - lim;
+  if (lim2 < 0) return;
   int offset;
   Mediator *m = MediatorNew(win_len, rank);
   T *data = new T[win_len]();
@@ -200,7 +201,6 @@ void _rank_filter(T *in_arr, int rank, int arr_len, int win_len, T *out_arr,
     MediatorInsert(data, m, in_arr[i]);
     out_arr[i - lim] = data[m->heap[0]];
   }
-
   switch (mode) {
   case REFLECT:
     arr_len_thresh = arr_len - 1;
--- a/scipy/ndimage/src/_rank_filter_1d.cpp
+++ b/scipy/ndimage/src/_rank_filter_1d.cpp
@@ -157,11 +157,7 @@ void _rank_filter(T *in_arr, int rank, int arr_len, int win_len, T *out_arr,
   int lim2 = arr_len - lim;
   int offset;
   Mediator *m = MediatorNew(win_len, rank);
-  T *data = new T[win_len];
-  for (int i = 0; i < win_len; ++i) {
-    data[i] = 0;
-  }
-
+  T *data = new T[win_len]();
 
   switch (mode) {
   case REFLECT:
@@ -204,6 +200,7 @@ void _rank_filter(T *in_arr, int rank, int arr_len, int win_len, T *out_arr,
     MediatorInsert(data, m, in_arr[i]);
     out_arr[i - lim] = data[m->heap[0]];
   }
+
   switch (mode) {
   case REFLECT:
     arr_len_thresh = arr_len - 1;
@@ -227,7 +224,7 @@ void _rank_filter(T *in_arr, int rank, int arr_len, int win_len, T *out_arr,
     break;
   case MIRROR:
     arr_len_thresh = arr_len - 2;
-    for (i = 0; i < lim + 1; i++) {
+    for (i = 0; i < lim; i++) {
       MediatorInsert(data, m, in_arr[arr_len_thresh - i]);
       out_arr[lim2 + i] = data[m->heap[0]];
     }
--- a/scipy/ndimage/tests/test_filters.py
+++ b/scipy/ndimage/tests/test_filters.py
@@ -7,6 +7,10 @@ import contextlib
 import numpy as np
 import pytest
 from numpy.testing import suppress_warnings, assert_allclose, assert_array_equal
+from numpy.typing import NDArray
+from hypothesis import strategies as st
+from hypothesis import given
+import hypothesis.extra.numpy as npst
 from pytest import raises as assert_raises
 from scipy import ndimage
 from scipy._lib._array_api import (
@@ -2985,17 +2989,21 @@ class TestVectorizedFilter:
 
 
 @pytest.mark.slow
-@pytest.mark.parametrize('length', range(1, 1000, 50))
-@pytest.mark.parametrize('size', range(1, 25))
-@pytest.mark.parametrize('mode', ['reflect', 'constant', 'nearest', 'mirror', 'wrap'])
+@given(
+    arr=npst.arrays(
+        dtype=np.float64,
+        shape=st.integers(min_value=1, max_value=20).map(lambda x: x * 50),
+    ),
+    size=st.integers(min_value=1, max_value=25),
+    mode=st.sampled_from(['reflect', 'constant', 'nearest', 'mirror', 'wrap']),
+)
 def test_gh_22586_median_filter_no_segfault_and_1d_matches_nd(
-    length: int,
+    arr: NDArray[np.float64],
     size: int,
     mode: str,
     monkeypatch: pytest.MonkeyPatch,
 ):
-    size = min(size, length)
-    arr = np.random.random(length)
+    size = min(size, len(arr))
     mf = functools.partial(
         ndimage.median_filter,
         size=size,
--- a/scipy/ndimage/tests/test_filters.py
+++ b/scipy/ndimage/tests/test_filters.py
@@ -2982,3 +2982,40 @@ class TestVectorizedFilter:
         res = ndimage.vectorized_filter(input, function, size=21)
         ref = ndimage.vectorized_filter(input, function, size=21)
         xp_assert_close(res, ref)
+
+
+@pytest.mark.slow
+@pytest.mark.parametrize('length', range(1, 1000, 50))
+@pytest.mark.parametrize('size', range(1, 25))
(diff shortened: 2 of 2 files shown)


scipy/ndimage/tests/test_filters.py
```diff
@@ -2766,3 +2766,40 @@ def test_gh_22333():
...
+
+    def _1d_to_nd_rank_filter(i, r, s, *args):
+        from scipy.ndimage import _nd_image
+        return _nd_image.rank_filter(i, r, np.ones(s, dtype=bool), *args)
+
+    with monkeypatch.context():
+        monkeypatch.setattr(
+            'scipy.ndimage._rank_filter_1d.rank_filter',
+            _1d_to_nd_rank_filter,
+        )
```
tylerjereddy, line 2802: a bit convoluted--I guess there's no easy way to go through a public function directly to check?

tylerjereddy (commented):
I still need to test on my Linux box since the test passes even without the patch on my Mac.

scipy/ndimage/tests/test_filters.py
```diff
@@ -2766,3 +2766,40 @@ def test_gh_22333():
...
+        desired = mf(arr)
+
+    np.testing.assert_array_almost_equal(actual, desired)
```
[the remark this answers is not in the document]
aiudirog: Ah, good catch, thanks!

scipy/ndimage/tests/test_filters.py
aiudirog: Yeah there isn't great access to get there without patching or adding another parameter. The closest I got was using an Nx1 array, but that didn't work for `mode='constant'` where it would just output a column of zeros.

scipy/ndimage/tests/test_filters.py
```diff
@@ -2766,3 +2766,40 @@ def test_gh_22333():
     expected = [58, 67, 87, 108, 163, 108, 108, 108, 87]
     actual = ndimage.median_filter(x, size=9, mode='constant')
     assert_array_equal(actual, expected)
+
+
+@pytest.mark.slow
+@pytest.mark.parametrize('length', range(1, 1000, 50))
+@pytest.mark.parametrize('size', range(1, 25))
+@pytest.mark.parametrize('mode', ['reflect', 'constant', 'nearest', 'mirror', 'wrap'])
```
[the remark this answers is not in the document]
aiudirog: Hypothesis was making a bit more difficult to find the cases that worked vs didn't. I'll rewrite the test now that it seems to be all good.

scipy/ndimage/tests/test_filters.py
```diff
@@ -2766,3 +2766,40 @@ def test_gh_22333():
...
+def test_gh_22586_median_filter_no_segfault_and_1d_matches_nd(
+    length: int,
+    size: int,
+    mode: str,
+    monkeypatch: pytest.MonkeyPatch,
+):
+    size = min(size, length)
+    arr = np.random.random(length)
```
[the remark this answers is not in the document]
aiudirog: This disappears in the hypothesis rewrite, but good to know for any future PRs. Thanks!

tylerjereddy:
I rebased and resolved merge conflicts

scipy/ndimage/tests/test_filters.py
```diff
@@ -2982,3 +2986,44 @@ def test_edge_cases(self, xp):
...
+@pytest.mark.slow
+@given(
+    arr=npst.arrays(
+        dtype=np.float64,
+        shape=st.integers(min_value=1, max_value=20).map(lambda x: x * 50),
+    ),
+    size=st.integers(min_value=1, max_value=25),
+    mode=st.sampled_from(['reflect', 'constant', 'nearest', 'mirror', 'wrap']),
+)
+def test_gh_22586_median_filter_no_segfault_and_1d_matches_nd(
```
tylerjereddy, line 3000: I'm not seeing a fail before/pass after application of the patch with this test in its current state on x86_64 Linux, even if I manually increase the 
 number of scenarios probed via `@settings(max_examples=5000)`.

Is anyone seeing this test fail on Linux if the source patch is reverted back to `main`?

In my hands, the original reproducing test from gh-22586 still does the trick. I suppose we could add both tests to be safe perhaps, but I don't even see a failure let alone a segfault for this test on `main` at the moment.

tylerjereddy:
We also don't usually add type hints to test functions, though I suppose it isn't wrong. In the interests of moving this forward for `1.15.3`, I may consider just pushing in the test that segfaults before/passes after as well, if I don't hear back by the weekend or so.

scipy/ndimage/tests/test_filters.py
aiudirog: I see the same thing and after some digging, it appears I may have made a mistake when converting the test to hypothesis? If you drop that commit, the parametrized pytest version segfaults if the changes to `_rank_filter_1d.cpp` aren't applied.

tylerjereddy:
Yeah, I observed that too.

Also, now even with the patch applied I see a segfault for `spin test -t scipy/ndimage/tests/test_filters.py::test_gh_22586_crash_property` using `hypothesis` `6.131.9`. Sometimes it will pass but emit `double free or corruption (!prev)`. hmm...

aiudirog:
Doing the same on my system (also x86_64 Linux - AMD 9800X3D) tosses `corrupted size vs. prev_size` which is apparently a different heap corruption glibc message, so it would seem things are definitely still broken.

scipy/ndimage/src/_rank_filter_1d.cpp
```diff
@@ -157,11 +157,7 @@ void _rank_filter(T *in_arr, int rank, int arr_len, int win_len, T *out_arr,
   int lim2 = arr_len - lim;
```
aiudirog, line 157: I think the other issue might be that this isn't validated to not be negative. Digging into it currently.

aiudirog: Adding `if (lim2 < 0) return;` on the next line seems to stop the heap corruption. Do you see the same thing?

tylerjereddy:
I made a few changes here to see if we can have this ready for the `1.15.3` release:

1) I switched us back to the original property-based test used in the matching issue--it sounds like you were "ok" with that at https://github.com/scipy/scipy/pull/22608#issuecomment-2840537071.
2) I added your additional patch from https://github.com/scipy/scipy/pull/22608#discussion_r2067771870.
3) I removed an unrelated whitespace change in the diff.
4) I rebased on latest `main`.

With those changes, the following test incantation (50 repeats with `pytest-repeat`) reliably passes on this feature branch and fails (segfaults) when the source patch is reverted:

`spin test -t scipy/ndimage/tests/test_filters.py::test_gh_22586_crash_property -- --count=50`.

If the CI agrees/is happy, I'll probably go ahead and squash-merge for inclusion in `1.15.3`.

The main point of hesitation is that we maybe don't understand all the details of what is going on here, but that was also true of the previous patches applied on this code in `1.15.x` series, and preventing a segfault is good.

aiudirog:
Yeah this is good, definitely best to just make sure the segfault is gone.
My only concern is that it doesn't confirm that the 1D version gives the
same result as the ND version.

I did notice in cases where lim2 would be negative, the ND version does
produce an output different from the input, whereas this just does nothing.
Though I'm not sure if that result should be considered "undefined".

tylerjereddy:
Ok, CI is happy--let's give this a try for `1.15.3` segfault guard.

I suspect this is just another step toward the actually-correct approach rather than perfecting things just yet.

tylerjereddy:
Thanks for your first contribution to SciPy

PR #21933, opened 2024-11-24
DOC: add Array API Standard support directive

mdhaber:
#### Reference issue
gh-17597
gh-21255

#### What does this implement/fix?
This adds an admonition that advertises support for non-NumPy backends.

#### Additional information
Just an experiment; it's up for debate whether we want to do this yet.

Issue #21255, opened 2024-07-24
RFC:DOC: Better documentation of functions that support the array API


j-bowhay:
Currently, we list functions that support the array API on https://scipy.github.io/devdocs/dev/api-dev/array_api.html#currently-supported-functionality . Initially, while support was very limited, this was a good solution. However, I believe we are now outgrowing it as more functions gain support. I would like to propose we discuss a long term solution to how we document array API support. I think this is necessary for the following reasons:

- Visibility. The current documentation is not very visible and requires you to go back and forth between the API docs, which is annoying as a user. Whilst this is still an experimental feature, I believe we should start to draw more attention to array API support so that we can gain more feedback and users before flicking the switch.
- Partial support: Currently it is not clear which functions support all array backend vs CPU only. Also some functions don't work with JAX etc
- Delegation: I think it is important the documentation is clear about when we delegate to another library ie as we do with special functions. Whilst normally the user should not have to care about implementation details, I would argue scientific computing is different. Different implementations may have different stability and performance characteristics which users care about. For example, with special some implementations have subpar accuracy for certain functions/inputs so it is important the user knows where their function call is "going".
- Maintainability: The current list is a bit of a chore to maintain being in sentence format as if every array API PR added one function to this list there would be a lot of merge conflicts. This means in practice we wait and add them in bulk before a release. This is easy to forget (ie last release)!

I propose we maintain some sort of central record that allows us to record functions/modules that support the array API, which array types they support, and where/which libraries they delegate to. For this central record, we could then automatically create a table of support that supersedes our current documentation. We could also automatically inject into the API documentation pages to improve the visibility of functions that support the array API.

cc @rgommers

j-bowhay:
Also in the case of datasets etc which will always return NumPy arrays it should be made clear that it is the user's job to then do `xp.asarray` on the output

rgommers:
Thanks for this @j-bowhay! I agree that it's time to make this more visible in the docs, and also that we need a structured and maintainable way of recording this metadata and translating it into html docs.

One important point that we discussed in the last community meeting is that while we want both a single-page (or per-submodule) overview as well as info in the individual function documentation, we probably do not want to dynamically update all the docstrings at import time, to avoid a large import time hit (this would need checking, but feels expensive). 

[...]

PR #22609, opened 2025-03-01
ENH: stats: shared array api support information to generate docs and test skip decorators

mdhaber:
#### Reference issue
Supersedes gh-21933
Toward gh-21255

gh-21255 proposes that there should be a central source of information about the Array API capabilities of a given function, and this information should be used to inject documentation. I also thought it would be nice to use this information to control test skips, and it looks like that was also suggested in https://github.com/scipy/scipy/issues/21255#issuecomment-2328382749. This PR implements these ideas.

#### Additional information
Didn't think too carefully about function names, so feel free to suggest specific new ones.


lucascolley:
> I ran out of steam for using the information to skip tests, so that part is sloppy. We need to augment the way the information is entered and stored (e.g. so we can store skip reasons), and we'll probably want to use the information to produce a single decorator that itself applies a sequence of `pytest.mark.skip_xp_backend` decorators to a function/class - then this decorator gets applied to the test class/function.

any thoughts here @crusaderky ?

scipy/_lib/_array_api.py
```diff
@@ -652,3 +657,91 @@ def xp_default_dtype(xp):
...
+        for attr in ['cpu', 'gpu', 'mask']:
+            val = "✓" if getattr(backend, attr) else " "
+            val += " " * (len('{numpy.}') + len(attr) - 1)
+            setattr(backend, attr, val)
+
+    table = f"""                
+    +---------+-------------+-------------+--------------+
+    | Library | CPU         | GPU         | ``marray``   |
+    +=========+=============+=============+==============+
+    | NumPy   | {numpy.cpu} | {numpy.gpu} | {numpy.mask} |
```
mdhaber, line 709: Looks like Sphinx wants borders between lines to render this as intended

scipy/stats/_stats_py.py
```diff
@@ -4363,6 +4366,10 @@ def confidence_interval(self, confidence_level=0.95, method=None):
         return ci
 
 
+xp_capabilities_table['pearsonr'] = dict(cpu_only=True)
+
+
+@xp_capabilities(**xp_capabilities_table['pearsonr'])
```
mdhaber, line 4372: Assuming we want to leave `skip_xp_backends` right now (I don't want to touch it, anyway), I'm thinking this can accept a sequence of dictionaries, each of which defines valid keargs input to `skip_xp_backends`. We process these to produce the capabilities table here, and it would then be very easy to chain together a test decorator from them since each element is valid input to `skip_xp_backends`.

ev-br:
Tangentially related: I've a prototype of a small hack which records, for a test run, for each scipy function all array types (== backends) it was called with. So that after `python dev.py test -b all` we'll have a list of dicts scipy_function: list_of_backends.
It's not PR ready yet, will clean it up soon (tm).

scipy/stats/_stats_py.py
```diff
@@ -4363,6 +4366,10 @@ def confidence_interval(self, confidence_level=0.95, method=None):
         return ci
 
 
+xp_capabilities_table['pearsonr'] = dict(cpu_only=True)
```
mdhaber, line 4369: Currently, `xp_capabilities_table` is a dictionary defined in the module per Ralf's suggestion, but it can be defined anywhere.

The keys are function names, and the values are dictionaries that accept keyword input similar to the `skip_xp_backend` decorator, like `cpu_only`, `np_only`, and `exceptions`. The one difference is that all the backends to skip and corresponding reasons would be provided as list of tuples, e.g.
```python3
[('cupy', 'I said so'), 
 ('jax.numpy', 'Just because']
```
Or I suppose these tuple could be provided as `*args` (each tuple is an element), but currently it's a single argument, `skip_backends`.

I've only implemented `skip_backends` and `cpu_only` for now.

scipy/stats/_stats_py.py
```diff
@@ -4363,6 +4366,10 @@ def confidence_interval(self, confidence_level=0.95, method=None):
...
+
+
+@xp_capabilities(xp_capabilities_table)
```
mdhaber, line 4372: The job of the decorator that gets applied to the function is to add the table to the documentation. The decorator factory `xp_capabilities` accepts the module's `xp_capabilities_table` as an argument. The decorator returned by the factory gets passed the function as an argument `f`, so inside that we can look up `xp_capabilities_table[f.__name__]`.

scipy/stats/tests/test_stats.py
```diff
@@ -379,7 +381,7 @@ def test_pROUNDROUND(self):
         assert_approx_equal(r,1.0)
 
 
-@skip_xp_backends(cpu_only=True)
+@make_skip_xp_backends('pearsonr', xp_capabilities_table)
```
mdhaber, line 384: Now we only have to apply one decorator to test classes. The `make_skip_xp_backends` decorator factory looks up the information in `xp_capabilities_table['pearsonr']`, creates all the appropriate decorators with that information, then returns a single decorator that chains them all together.

If we were to name test classes in a standard way (which is usually the case right now, especially as we go back through and do arrray API translations), we wouldn't have to provide the function name as an argument.

mdhaber (commented):
OK, I went ahead and implemented what I had in mind for skipping tests based on the shared info.

scipy/_lib/_array_api.py
```diff
@@ -652,3 +657,115 @@ def xp_default_dtype(xp):
...
+def _make_capabilities_note(fun_name, capabilities):
+    table = _make_capabilities_table(capabilities)
+    note = f"""
+    `{fun_name}` has experimental support for Python Array API Standard compatible
+    backends in addition to NumPy. Please consider testing these features
+    by setting an environment variable ``SCIPY_ARRAY_API=1`` and providing
+    CuPy, PyTorch, JAX, or Dask arrays as array arguments. The following
+    combinations of backend and device (or other capability) are supported.
+    {table}
+     See :ref:`dev-arrayapi` for more information.
```
mdhaber, line 732:
```suggestion
    See :ref:`dev-arrayapi` for more information.
```
Other than this, looks pretty good to me:
image

scipy/_lib/_array_api.py
```diff
@@ -652,3 +657,115 @@ def xp_default_dtype(xp):
...
+        capabilities = _make_capabilities(skip_backends=skip_backends,
+                                          cpu_only=cpu_only)
+
+        @functools.wraps(f)
+        def wrapper(*args, **kwargs):
+            return f(*args, **kwargs)
+
+        note = _make_capabilities_note(fun_name, capabilities)
+        doc = FunctionDoc(wrapper)
+        doc['Notes'].append(note)
```
j-bowhay, line 751: Does `numpydoc` allow us define a new section like `Array API Compatibility` for better visibility rather than being at the end of the notes?

scipy/_lib/_array_api.py
```diff
@@ -652,3 +657,115 @@ def xp_default_dtype(xp):
...
+
+    for backend in [numpy, cupy, torch, jax, dask]:
+        for attr in ['cpu', 'gpu', 'mask']:
+            val = "✓" if getattr(backend, attr) else " "
+            val += " " * (len('{numpy.}') + len(attr) - 1)
+            setattr(backend, attr, val)
+
+    table = f"""                
+    +---------+-------------+-------------+--------------+
+    | Library | CPU         | GPU         | ``marray``   |
```
j-bowhay, line 707: Certainly not required for this pr but it could be nice in the future if the `marray` text and other libraries linked to their docs

scipy/stats/_stats_py.py
j-bowhay: There was some concern when I originally suggested the idea that using decorators to inject the docs might cause too much of a slowdown at import time if there are lots of functions. However it seems reasonable to me to start with something like this and then, if this becomes an issue, move to some sphinx magic

scipy/stats/tests/test_stats.py
j-bowhay: If we defined `xp_capabilities_table` centrally in `_array_api.py` then the decorators could just use it directly?

scipy/stats/_stats_py.py
j-bowhay: Looks like we'd need mark `jax.numpy` support as false since all tests are skipped due to item assignment so a good opportunity to test this out

j-bowhay (commented):
Thanks for getting the ball rolling on this, I think it's an important step in making the array API support more useable!

scipy/_lib/_array_api.py
mdhaber: I'm not sure, but if not, we could ask.

scipy/_lib/_array_api.py
mdhaber: Yeah I thought about that, but I'm not sure reST tables support links. I tried citations before, and that didn't work.

scipy/_lib/_array_api.py
j-bowhay: Not required for this pr though but it would be nice for the future

scipy/stats/_stats_py.py
mdhaber: Yeah. But if that comes up again, then it's basically saying that only rendered docs matter. If that's the case, then we should be "allowed" to use :math: liberally.

scipy/stats/tests/test_stats.py
mdhaber: They could, but didn't we want to keep it close to the functions?

mdhaber: > I mean just defining the empty dictionary centrally

Ah, yes, thought about that, but shied away given the new multithreading work. I would definitely like that better than separate dictionaries in each module, though. It would mostly eliminate the advantages of attaching the information to the function. Is it ok to have a global dictionary that's mutated all over SciPy (but only once, on import)?

doc/source/dev/api-dev/array_api.rst
```diff
@@ -121,6 +121,32 @@ Support is provided in `scipy.stats` for the following functions:
...
+| NumPy   | ✓           | ✗           | ✓            |
++---------+-------------+-------------+--------------+
+| CuPy    | ✗           | ✓           |              |
++---------+-------------+-------------+--------------+
+| PyTorch | ✓           |             |              |
++---------+-------------+-------------+--------------+
+| JAX     | ✓           | ✓           |              |
++---------+-------------+-------------+--------------+
+| Dask    |             |             |              |
++---------+-------------+-------------+--------------+
```
mdhaber, line 138: This looks like it's aligned locally.

<details>

image

</details>

mdhaber (commented):
I defined the dictionary in `_array_api.py` as suggested. If that's going to be a problem, we can change it to having one dictionary per module.

I added a bit of an explanation of the table to `array_api.rst`; that links to `marray`. We could add links to the other backend websites there, too.

I think the machinery now supports all the `skip_xp_backends` features that we use in the tests - besides skipping individual backends, that's `cpu_only`, possibly with `exceptions`, and `np_only`.  If `exceptions` is supposed to work with `np_only`, I'd suggest we trim things down a bit so that there is not so much variety in how that can be accomplished. 

I also applied this work to `zmap`/`zscore`. If this is looking good, I can apply it to the rest of `_stats_py.py` and tests; I leave the rest of SciPy to follow-up work.

doc/source/dev/api-dev/array_api.rst
```diff
@@ -121,6 +121,32 @@ Support is provided in `scipy.stats` for the following functions:
 `scipy.stats.jarque_bera`, `scipy.stats.bartlett`, `scipy.stats.power_divergence`,
 and `scipy.stats.monte_carlo_test`.
 
+Some features provide a capability table in the documentation like this:
+
++---------+-------------+-------------+--------------+
+| Library | CPU         | GPU         | ``marray``   |
++=========+=============+=============+==============+
+| NumPy   | ✓           | ✗           | ✓            |
```
rgommers, line 129: minor: probably better to use `x` to indicate support is missing when that's possible, and "n/a" when the cell can never be checked - NumPy on GPU or CuPy on CPU will never be a thing.

rgommers:
I had a quick look (in transit, only a bit of time now) and it looks promising to me. Thanks for working on this @mdhaber!

The `@make_skip_...` approach to reuse the compatibility info LGTM. There will still be cases where `@skip_...`will be needed, when the incompatibility is in the test itself instead of in the function under test, but that should be a minority of cases. Overall this should make things more consistent and a bit more concise inside the test files.

scipy/stats/_stats_py.py
```diff
@@ -4363,6 +4371,10 @@ def confidence_interval(self, confidence_level=0.95, method=None):
         return ci
 
 
+xp_capabilities_table['pearsonr'] = dict(cpu_only=True, exceptions=['cupy'])
```
j-bowhay, line 4374: I'm not sure I have a strong opinion but I wonder if we should should use the function handle as the key instead as, for example, `linalg` and `sparse.linalg` have functions with the same name

mdhaber: The other example shows why we don't necessarily just want one key per function, and in that case strings are appropriate. Collisions will be caught during testing, and the user is welcome to fix them by providing a fully qualifiied name as a string or function handle, if they prefer.

j-bowhay: Fair point. Another advantage of using the function handle is that it would allow for a module level comparability table in the docs but I guess we can always change things down the line if that's something we want to add

scipy/_lib/_array_api.py
```diff
@@ -652,3 +657,139 @@ def xp_default_dtype(xp):
...
+    +=========+=============+=============+==============+
+    | NumPy   | {numpy.cpu} | ✗           | {numpy.mask} |
+    +---------+-------------+-------------+--------------+
+    | CuPy    | ✗           | {cupy.gpu } | {cupy.mask } |
+    +---------+-------------+-------------+--------------+
+    | PyTorch | {torch.cpu} | {torch.gpu} | {torch.mask} |
+    +---------+-------------+-------------+--------------+
+    | JAX     | {jax.cpu  } | {jax.gpu  } | {jax.mask  } |
```
crusaderky, line 724: I think it's important to distinguish between JAX in eager mode vs. jax.jit

scipy/_lib/_array_api.py
```diff
@@ -652,3 +657,139 @@ def xp_default_dtype(xp):
...
+    +---------+-------------+-------------+--------------+
+    | Dask    | {dask.cpu } | {dask.gpu } | {dask.mask } |
```
crusaderky, line 726: Likewise, saying "it functionally reaches the end with dask" is a very different thing from "it is usable in production with Dask and you should expect OK performance at least in some cases" due to materializations.

This and the above information about JAX is currently stored in the `lazy_xp_function` helpers in the test modules.

scipy/stats/_stats_py.py
```diff
@@ -4363,6 +4371,10 @@ def confidence_interval(self, confidence_level=0.95, method=None):
...
+
+
+@xp_capabilities()
```
crusaderky, line 4377: UX nit: the `@xp_capabilities` decorator could inject into `xp_capabilities_table` itself
```suggestion
@xp_capabilities(cpu_only=True, exceptions=['cupy'])
```

scipy/stats/_stats_py.py
```diff
@@ -2660,6 +2660,13 @@ def _isconst(x):
         return (y[0] == y).all(keepdims=True)
 
 
+xp_capabilities_table['zmap_zscore'] = dict(skip_backends=[
```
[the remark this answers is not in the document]
ev-br: +1. `stats.zmap_score` would be better, really.

scipy/_lib/_array_api.py
mdhaber: Would there be a column or row for that? I was aware of the distinction, but I don't know much more.

scipy/stats/_stats_py.py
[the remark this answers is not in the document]
mdhaber: No. There is not a function with that name.

scipy/_lib/_array_api.py
crusaderky: jax.jit functionality is always a subset of eager jax. So maybe the layout could look something like this?
```
JAX | {green check mark} |
JAX | {yellow warning symbol} Eager only |
JAX | {red x} |
```

scipy/stats/_stats_py.py
mdhaber: I actually described that it would work that way before writing it. Not sure why I didn't do that, so I'll change it.

scipy/_lib/_array_api.py
mdhaber: Sure. I was hoping to start simple, though. Would this require a change to the information stored in the `xp_capabilities_table ` dictionary, or do you mean below that it be looked up for a given function with `lazy_xp_function`?

j-bowhay (approved):
A big improvement IMO for visibility and for figuring out backend support. I'm sure we will end up tweaking things and finalizing the design (eg. https://github.com/scipy/scipy/pull/22609#discussion_r1980342136) but I don't think we need to fully converge here so I would propose merging soon.

Looks like we have picked up a merge conflict in the meantime.

scipy/stats/tests/test_stats.py
```diff
@@ -6313,8 +6312,7 @@ def test_ttest_1samp_new_omit(xp):
     xp_assert_close(t, tr)
 
 
-@make_skip_xp_backends('ttests')
-@pytest.mark.skip_xp_backends('jax.numpy', reason='Generic impl mutates array.')
+@pytest.mark.skip_xp_backends('jax.numpy', reason='Generic stdtrit mutates array.')
```
mdhaber, line 6315:
```suggestion
@make_skip_xp_backends('ttests')
@pytest.mark.skip_xp_backends('jax.numpy', reason='Generic stdtrit mutates array.')
```

scipy/stats/tests/test_stats.py
```diff
@@ -75,7 +75,6 @@
 ROUND = array([0.5,1.5,2.5,3.5,4.5,5.5,6.5,7.5,8.5], float)
 
 
-@make_skip_xp_backends("trimmed statistics")
 class TestTrimmedStats:
```
mdhaber, line 78:
```suggestion
@make_skip_xp_backends("trimmed statistics")
class TestTrimmedStats:
````

mdhaber:
Merge conflicts resolved @j-bowhay. Thanks!

PR #22611, opened 2025-03-01
MAINT: stats: rewrite `gaussian_kde.integrate_box`, remove `_mvn` F77 extension

ev-br:
#### Reference issue

A follow-up to https://github.com/scipy/scipy/pull/22298

Use a new `_qmvn` multivariate normal integrator in `gaussian_kde`. Previously, gh-22298 replaced the F77 integrator, `_mvn.mvnun`, with a improved version. Now this PR replaces `_mvn.mvnun_weighted`, which completes the removal of the F77 original.

#### Additional information

The first commits only adds tests of the "old" implementation vs multivariate_normal; the second commit changes the implementation, and the third commit removes the Fortran code.


scipy/stats/_kde.py
```diff
@@ -390,19 +383,13 @@ def integrate_box(self, low_bounds, high_bounds, maxpts=None):
...
-            msg = f'An integral in _mvn.mvnun requires more points than {self.d * 1000}'
-            warnings.warn(msg, stacklevel=2)
+            kwds.update({'maxpts': maxpts})
 
-        return value
+        low, high = low_bounds - self.dataset.T, high_bounds - self.dataset.T
+        values = multivariate_normal.cdf(
+            high, lower_limit=low, cov=self.covariance, **kwds
+        )
+        return (values * self.weights).sum(axis=-1)
```
mdhaber, line 392: Probably not the bottleneck, but would be nice to use `np.vecdot` when it's available.

scipy/stats/tests/test_kdeoth.py
```diff
@@ -19,7 +19,15 @@ def test_kde_1d():
     gkde = stats.gaussian_kde(xn)
 
     # evaluate the density function for the kde for some points
-    xs = np.linspace(-7,7,501)
+    xx = np.asarray([0.1, 0.5, 0.9])
+    loc, scale = gkde.dataset, np.sqrt(gkde.covariance)
+    assert_allclose(
```
mdhaber, line 24: Just taking the opportunity to add PDF tests?

mdhaber (commented):
Didn't look at the nitty gritty of the tests yet. Before I do, just wanted to check what prompted the addition of PDF tests - is there nothing like them there already?

rkern (approved):
LGTM!

scipy/stats/tests/test_kdeoth.py
ev-br: Pretty much yes. Was checking that it works the way I think it does, added it as a test. Can remove if desired.

scipy/stats/_kde.py
```diff
@@ -371,7 +364,7 @@ def integrate_box_1d(self, low, high):
                         special.ndtr(normalized_low)))
         return value
 
-    def integrate_box(self, low_bounds, high_bounds, maxpts=None):
+    def integrate_box(self, low_bounds, high_bounds, maxpts=None, **kwds):
```
rgommers, line 367: `**kwds` isn't documented in this PR, and this is a public method. Is this actually necessary? It ties `integrate_box` to current and future implementation details of `multivariate_normal`.

rgommers (commented):
Looks good! One question on the API.

scipy/stats/_kde.py
ev-br: You're suggesting an explicit branching on the numpy version, or just a comment saying can use vecdot when minimum numpy is >= 2.0?

scipy/stats/_kde.py
ev-br: Yeah, not strictly necessary, no. Currently it is an escape hatch for when (if and when, actually) the "new" integrator is not exactly backwards compatible and then a user can fiddle with `rng` and `abserr` knobs of  `multivariate_normal.cdf`. Can definitely take it out, at least until there's an explicit user request.

rgommers: +1 for taking it out then

scipy/stats/_kde.py
mdhaber: I figure that if it's just a comment, it probably won't happen. I'll bet there are a lot of places that would benefit from vecdot, so maybe a helper in `_util` that does the branching. (It's available pre 2.0 in `numpy.array_api`.)

If you'd prefer, leave it like this, and I'll consider a separate PR to change it here and other places that could use it. It really made a big speed difference in `pearsonr`.

scipy/stats/tests/test_kdeoth.py
mdhaber: They are probably fine. I'll take a look today.

scipy/stats/_kde.py
ev-br: FTR, 

```
In [8]: a = np.arange(10000, dtype=float)

In [9]: %timeit (a * a).sum(axis=-1)
7.08 μs ± 35 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [10]: %timeit np.vecdot(a, a)
2.39 μs ± 18.9 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

Wow. In retrospect, it's not that surprising: probably `O(N)` copying to a temp array for the product takes longer than `O(N)` work. Still, did not expect this.

Let me add a comment to this effect so that we change it everywhere in the array api conversion.

scipy/stats/_kde.py
ev-br: OK, done.

scipy/stats/_kde.py
mdhaber: I imagined also that `vecdot` can use a single BLAS routine for the operation?

It's OK, I'll just open a PR now to change it in a few places.

scipy/stats/_kde.py
```diff
@@ -389,20 +382,12 @@ def integrate_box(self, low_bounds, high_bounds, maxpts=None):
...
-        with MVN_LOCK:
-            value, inform = _mvn.mvnun_weighted(low_bounds, high_bounds,
-                                                self.dataset, self.weights,
-                                                self.covariance, **extra_kwds)
-        if inform:
-            msg = f'An integral in _mvn.mvnun requires more points than {self.d * 1000}'
-            warnings.warn(msg, stacklevel=2)
-
-        return value
+        low, high = low_bounds - self.dataset.T, high_bounds - self.dataset.T
```
mdhaber, line 385: Confirmed that `dataset` is `atleast_2d`. Also, it looks like `low_bounds` and `high_bounds` are only supposed to work with a single point at a time, and that will always be aligned along the `-1` axis (which is why we transpose `dataset`).

scipy/stats/_kde.py
```diff
@@ -389,20 +382,12 @@ def integrate_box(self, low_bounds, high_bounds, maxpts=None):
...
+        values = multivariate_normal.cdf(
+            high, lower_limit=low, cov=self.covariance, maxpts=maxpts
+        )
+        # XXX: xp.linalg.vecdot is much faster than (a*b).sum(axis=-1)
+        return (values * self.weights).sum(axis=-1)
```
mdhaber, line 390: Also checked that weights is already normalized. 👍

scipy/stats/tests/test_kdeoth.py
```diff
@@ -19,7 +19,15 @@ def test_kde_1d():
...
+        gkde(xx), 
+        stats.norm.pdf(xx[:, None], loc=loc, scale=scale).sum(axis=-1) / gkde.n,
+        atol=1e-14
```
mdhaber, line 27: Is there a reason to use `atol` throughout? If not, I'd default to `rtol`. Especially when looking at multivariate normal later, it's not immediately obvious that assertions about the PDF with `atol=1e-14` really test anything.

mdhaber (approved):
LGTM, too; just one question about whether there's a reason to use `atol` instead of `rtol`. It may be fine here, but it's not immediately obvious because MVN pdf/cdf can easily get quite small, so I'd default to `rtol` unless there's a reason to use `atol`.

scipy/stats/tests/test_kdeoth.py
ev-br: `¯\_(ツ)_/¯` this is just my default to "nearly equals in double precision", with an order of magnitude of a wiggle room.

mdhaber: OK, then I'd suggest changing that default to `rtol=1e-14`, at least when working on stats and special functions.

scipy/stats/tests/test_kdeoth.py
```diff
@@ -85,6 +108,26 @@ def test_kde_2d():
     # get kde for original sample
     gkde = stats.gaussian_kde(xn)
 
+    # evaluate vs multivariate normal, using the KDE definition
+    xx = np.asarray([[1, 2], [3, 4], [5, 6]])
+    arg = xx[:, None, :] - gkde.dataset.T
+    pdf = stats.multivariate_normal.pdf
+    assert_allclose(
+        gkde(xx.T),
```
mdhaber, line 116: For instance,
```
gkde(xx.T)
[5.35496592e-02 3.93564062e-04 6.68839968e-14]
```
so with `atol=1e-14`, the test with `[5, 6]` is not doing much.

scipy/stats/tests/test_kdeoth.py
```diff
@@ -19,7 +19,15 @@ def test_kde_1d():
     gkde = stats.gaussian_kde(xn)
 
     # evaluate the density function for the kde for some points
-    xs = np.linspace(-7,7,501)
+    xx = np.asarray([0.1, 0.5, 0.9])
+    loc, scale = gkde.dataset, np.sqrt(gkde.covariance)
+    assert_allclose(
+        gkde(xx), 
+        stats.norm.pdf(xx[:, None], loc=loc, scale=scale).sum(axis=-1) / gkde.n,
+        rtol=1e-14
```
mdhaber, line 27:
```suggestion
        rtol=5e-14
```

scipy/stats/tests/test_kdeoth.py
```diff
@@ -51,7 +59,18 @@ def test_kde_1d_weighted():
...
-    xs = np.linspace(-7,7,501)
+    # evaluate the density function for the kde for some points
+    xx = np.asarray([0.1, 0.5, 0.9])
+    loc, scale = gkde.dataset, np.sqrt(gkde.covariance)
+
+    pdf = stats.norm.pdf
+    assert_allclose(
+        gkde(xx), 
+        np.sum(pdf(xx[:, None], loc=loc, scale=scale) * gkde.weights, axis=-1),
+        rtol=1e-14
```
mdhaber, line 70:
```suggestion
        rtol=5e-14
```

scipy/stats/tests/test_kdeoth.py
```diff
@@ -85,6 +108,26 @@ def test_kde_2d():
...
     gkde = stats.gaussian_kde(xn)
 
+    # evaluate vs multivariate normal, using the KDE definition
+    xx = np.asarray([[1, 2], [3, 4], [5, 6]])
+    arg = xx[:, None, :] - gkde.dataset.T
+    pdf = stats.multivariate_normal.pdf
+    assert_allclose(
+        gkde(xx.T),
+        pdf(arg, cov=gkde.covariance).sum(axis=-1) / gkde.n,
+        rtol=1e-14
```
mdhaber, line 118:
```suggestion
        rtol=5e-14
```

mdhaber:
Alright, thanks @ev-br and reviewers!

PR #22613, opened 2025-03-02
BUG: ndimage: fix memory management issues in `_rank_filter_1d.cpp`

### LINKED ISSUE #22365: BUG: ndimage: C++ memory management issues in `_rank_filter_1d.cpp`
The file [`_rank_filter_1d.cpp`](https://github.com/scipy/scipy/blob/main/scipy/ndimage/src/_rank_filter_1d.cpp) was added in https://github.com/scipy/scipy/pull/20543.  A quick review of that code reveals a few problems that should be fixed:

* In these lines, the possibility of a failed memory allocation is not handled correctly:

  https://github.com/scipy/scipy/blob/7372d710ef6e97a3a9f3027efcf6b5b77bb33777/scipy/ndimage/src/_rank_filter_1d.cpp#L95-L101

  There are three problems here.

  * First, if any of the `new` operations fails, the exception `std::bad_alloc` will be thrown, and because the exception is not caught, Python will crash.

    This can be fixed by changing `new ...` to `new (std::nothrow) ...`.

  * Then there is the problem that `m` is not checked for being `nullptr` until after it has already been dereferenced with `m->pos`.  The check for `m == nullptr` must be made before an attempt is made to dereference it.

  * Finally, if any of the allocations have failed, it is wrong to print a message and exit Python!  This code must provide a mechanism to pass the error condition back to Python so that a `MemoryError` exception is raised.

* There is another unchecked memory allocation at

  https://github.com/scipy/scipy/blob/7372d710ef6e97a3a9f3027efcf6b5b77bb33777/scipy/ndimage/src/_rank_filter_1d.cpp#L159

* In this code

  https://github.com/scipy/scipy/blob/7372d710ef6e97a3a9f3027efcf6b5b77bb33777/scipy/ndimage/src/_rank_filter_1d.cpp#L251-L258

  if only one of `in_arr` or `out_arr` is NULL, then the memory allocated for the other is leaked.

(I was looking at this file because of https://github.com/scipy/scipy/issues/22250, but I don't think any of the issues reported here are the cause of the crashes reported there.)

anazli:
Prevent operator new from throwing bad alloc. If allocation fails, report a memory error.
Closes #22365
Additional information:
This also fixes some formatting issues in this file.


ev-br:
There's now a git conflict on one of the lines changed by this PR. Would be great to resolve @anazli 

Other than that, the change looks reasonable to me, as it seems to add error checking which is otherwise missing.

scipy/ndimage/src/_rank_filter_1d.cpp
```diff
@@ -315,19 +319,17 @@ static struct PyModuleDef _rank_filter_1d = {
     myMethods};
 
 // init the module
-PyMODINIT_FUNC
-PyInit__rank_filter_1d(void)
-{
-    PyObject *module;
+PyMODINIT_FUNC PyInit__rank_filter_1d(void) {
```
rgommers, line 322: Please try not to mix in style changes with bug fixes in the same commit. This change actually makes the code less idiomatic (see, e.g., https://docs.python.org/3/c-api/intro.html#c.PyMODINIT_FUNC) so please revert this one.

Indentation changes to match the style in a single file are okay, but best done in a separate PR or at least a separate commit.

scipy/ndimage/src/_rank_filter_1d.cpp
```diff
@@ -315,19 +319,17 @@ static struct PyModuleDef _rank_filter_1d = {
...
+  PyObject *module;
 
-    import_array();
-    module = PyModule_Create(&_rank_filter_1d);
-    if (module == NULL) {
-        return module;
-    }
+  import_array();
+  module = PyModule_Create(&_rank_filter_1d);
+  if (!module) {
```
rgommers, line 327: Please undo this change as well, less idiomatic

rgommers (changes requested):
Thanks for working on this @anazli, and sorry you had to wait so long for reviews. A few comments; rewriting git history or removing style changes completely would be nice.

WarrenWeckesser:
@anazli, thank you for your interest in improving SciPy!

There are more substantial changes that would be needed to complete this PR, and we haven't heard from the author since March, so I'm closing it.  I have new PR with a fix over in https://github.com/scipy/scipy/pull/23616.

PR #22616, opened 2025-03-02
ENH: stats: use `vecdot` and `nonzero` where appropriate

mdhaber:
#### Reference issue
https://github.com/scipy/scipy/pull/22611#discussion_r1976530805
gh-22462 

There are pretty substantial performance arguments for using `vecdot(x, y)` instead of `sum(x, y)`, so I think it's worth using, especially when it's conceptually clear.

```python3
import numpy as np
rng = np.random.default_rng(325943759872345)
x = rng.random(1000)
y = rng.random(1000)

%timeit np.sum(x * y)
8.45 μs ± 3.87 μs per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

%timeit np.vecdot(x, y)
# 1.16 μs ± 9.92 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

While I was at it, I thought it would be worth using `count_nonzero` when we're counting booleans, especially for very long slices.
```python3
mask = x > 0.5

%timeit np.sum(mask)
# 4.21 μs ± 94.4 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

%timeit np.count_nonzero(mask)
# 432 ns ± 2.54 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

#### Additional information
There are currently some gotchas to using `vecdot` that we need to watch out for, especially when the backend is not NumPy; see data-apis/array-api#910.

This just covers `_stats_py.py` for now. 

What do we want to do for things like `np.sum(x * y * z)`, where we might get a performance improvement, but may not be an improvement conceptually?


scipy/stats/_stats_py.py
```diff
@@ -8852,9 +8850,9 @@ def brunnermunzel(x, y, alternative="two-sided", distribution="t",
     rankx_mean = np.mean(rankx)
     ranky_mean = np.mean(ranky)
 
-    Sx = np.sum(np.power(rankcx - rankx - rankcx_mean + rankx_mean, 2.0))
+    Sx = np.linalg.norm(rankcx - rankx - rankcx_mean + rankx_mean)**2
```
mdhaber, line 8853: Note that input to `brunnermunzel` is always 1d.

```suggestion
    temp = rankcx - rankx - rankcx_mean + rankx_mean
    Sx = np_vecdot(temp, temp)
```

scipy/stats/_stats_py.py
```diff
@@ -8852,9 +8850,9 @@ def brunnermunzel(x, y, alternative="two-sided", distribution="t",
...
     Sx /= nx - 1
-    Sy = np.sum(np.power(rankcy - ranky - rankcy_mean + ranky_mean, 2.0))
+    Sy = np.linalg.norm(rankcy - ranky - rankcy_mean + ranky_mean)**2
```
mdhaber, line 8855:
```suggestion
    temp = rankcy - ranky - rankcy_mean + ranky_mean
    Sy = np_vecdot(temp, temp)
```

scipy/stats/_stats_py.py
```diff
@@ -10115,7 +10111,8 @@ def _sum_of_squares(a, axis=0):
 
     """
     a, axis = _chk_asarray(a, axis)
-    return np.sum(a*a, axis)
+    # return np.linalg.norm(a, axis=axis)**2  # faster, but loses precision
```
mdhaber, line 10114:
```suggestion
```

Note that `_chk_asarray` ravels input with `axis=None`, and `axis` tuples aren't supported, so data-apis/array-api#910 doesn't bite us here.

ev-br:
One gotcha is that vecdot complex conjugates its first argument. So any change from `sum(a, b)` to `vecdot(a, b)` must imply that `a` is definitely real.

ilayn:
Did not know about `vecdot` until today. Is this an array API thing? But you can cut it half further if you use the `dot` method where it matters. 

```python
In [5]: import numpy as np
   ...: rng = np.random.default_rng(325943759872345)
   ...: x = rng.random(1000)
   ...: y = rng.random(1000)
   ...: 
   ...: %timeit np.sum(x * y)
   ...: %timeit np.vecdot(x, y)
   ...: %timeit x.dot(y)
3.17 μs ± 9.14 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
927 ns ± 4.01 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
549 ns ± 13.4 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

lucascolley:
> Strange that we are going away from @ even on the numpy side which is meant to be used as the array product infix operator to avoid all this.

On my machine `x @ y` is ever so slightly worse than `vecdot`

```python
In [12]: %timeit np.vecdot(x, y)
731 ns ± 2.03 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
In [13]: %timeit x @ y
751 ns ± 3.58 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
In [14]: %timeit x.dot(y)
464 ns ± 2.17 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

mdhaber:
> Is this an array API thing?

It is specified in the array API, and NumPy seems to have added it in the effort toward array API compatibility.

> Strange that we are going away from @ even on the numpy side...

I think the use case for `vecdot` is pretty clear. `@` is fine for 1D arrays and when you actually want something that behaves like matrix multiplication. `vecdot` is for batch vector-vector dot products, and it was not as obvious how to do that before.

> But you can cut it half further 

There are some places here we could use `.dot` but:
- usually not, because there is often an `axis` argument, and
- we won't be able to when we use the array API.

ilayn:
>  vecdot is for batch vector-vector dot products

That's still what matrix multiplication is (albeit without proper axis rotations it is inconvenient to express). I don't have any objections for `vecdot`. But NumPy apparently missed yet another train for making `@` cover all these cases in the most performant way in my opinion. Like Lucas showed, it should be the fastest but isn't. We are populating the namespaces with these separate redundant functions to do the most fundamental linalg operation. But anyways, better  not get into another discussion here. I was just surprised to see matvec and vecmat and others.

mdhaber:
OK. 
I'll wait for comments before committing the changes in self-review above and bumping the tolerance on the test. (I don't think we need to preserve the exact floating point result, even if that was the point of the test.)

ev-br (approved):
LGTM.

On a high level,

- I'm not keep on replacing `sum(x*x)` with `linalg.norm(x)**2`. It does lose precision indeed from taking the square root, and is less expressive IMO. Precision loss is probably irrelevant in most cases here, the perf difference between `linalg.norm(x)**2` and `vecdot(x, x)`  is not that large, is it. - So as a perf improvement I'd only do it where it _really_ matters, in tight loops or something. And each use probably needs a code comment justifying the choice.
- I respectfully disagree that `vectdot` is inferior than matmul, `x @ x`, somehow. "linalg" vs "batch" semantic is incompatible, so this is always confusing when batches are present. I personally would write `a @ b` only if `a` and `b` are both known to be matrices (2D matrices or batches of 2D matrices), and use explicit `vecdot`, `matvec` and `vecmat` otherwise. Ditto for any code I read or review. What `np.dot` does with non-trivial dimensions, I've no idea (meaning, I need to RTFM and an interactive session to figure out, every single time---maybe I'm just dense enough)

ev-br:
Optionally, can add the patch below, as discussed in https://github.com/scipy/scipy/pull/22611#discussion_r1976530805
I can send it as a PR to your branch, but that needs a rebase on main to apply.

```diff
diff --git a/scipy/stats/_kde.py b/scipy/stats/_kde.py
index 143e62e274..9af046ee8b 100644
--- a/scipy/stats/_kde.py
+++ b/scipy/stats/_kde.py
@@ -19,7 +19,7 @@
 
 # SciPy imports.
 from scipy import linalg, special
-from scipy._lib._util import check_random_state
+from scipy._lib._util import check_random_state, np_vecdot
 
 from numpy import (asarray, atleast_2d, reshape, zeros, newaxis, exp, pi,
                    sqrt, ravel, power, atleast_1d, squeeze, sum, transpose,
@@ -359,9 +359,8 @@ class gaussian_kde:
         normalized_low = ravel((low - self.dataset) / stdev)
         normalized_high = ravel((high - self.dataset) / stdev)
 
-        value = np.sum(self.weights*(
-                        special.ndtr(normalized_high) -
-                        special.ndtr(normalized_low)))
+        delta = special.ndtr(normalized_high) - special.ndtr(normalized_low))
+        value = np_vecdot(self.weights, delta)
         return value
 
     def integrate_box(self, low_bounds, high_bounds, maxpts=None):
@@ -386,8 +385,7 @@ class gaussian_kde:
         values = multivariate_normal.cdf(
             high, lower_limit=low, cov=self.covariance, maxpts=maxpts
         )
-        # XXX: xp.linalg.vecdot is much faster than (a*b).sum(axis=-1)
-        return (values * self.weights).sum(axis=-1)
+        return np_vecdot(values, self.weights, axis=-1)
 
     def integrate_kde(self, other):
         """
@@ -429,7 +427,7 @@ class gaussian_kde:
             diff = large.dataset - mean
             tdiff = linalg.cho_solve(sum_cov_chol, diff)
 
-            energies = sum(diff * tdiff, axis=0) / 2.0
+            energies = np_vecdot(diff, tdiff, axis=0) / 2.0
             result += sum(exp(-energies)*large.weights, axis=0)*small.weights[i]
 
         sqrt_det = np.prod(np.diagonal(sum_cov_chol[0]))
```

scipy/_lib/_util.py
```diff
@@ -1345,3 +1345,15 @@ def wrapper(*args, **kwargs):
 
         return wrapper
     return decorator
+
+
+def np_vecdot(x1, x2, /, *, axis=-1):
+    # `np.vecdot` has advantages (e.g. see gh-22462), so let's use it when
+    # available. As functions are translated to Array API, this can be replaced
+    # with `array-api-strict` NumPy's vecdot.
```
lucascolley, line 1353: Optional nit but I think this is clearer (if this is what you mean).

```suggestion
    # with `xp.vecdot`.
```

ilayn:
There are some details getting lost in here. `sum(x*x)` is a naive implementation of norm. The squareroot precision loss is not in par with what is possibly lost with an improper implementation of [`dnrm2`](https://netlib.org/lapack/explore-html/d1/d2a/group__nrm2_gab5393665c8f0e7d5de9bd1dd2ff0d9d0.html#gab5393665c8f0e7d5de9bd1dd2ff0d9d0) (in LAPACK lingo) when you need it. Because there is no scaling involved and no guard rails are placed in `sum(x*x)` for underflow or overflow. These are established algorithms by now and hence these linalg tricks won't match the robustness of dnrm2 if the vector data is susceptible to underflow or overflow. That's why Anderson's article is implemented as is in `nrm2`, `lartg` and a few more places in LAPACK. 

> I respectfully disagree that vectdot is inferior than matmul,

Disagreement is acknowledged :) Because there is no way you can match performance of matmul by separating the vectors (essentially replacing a BLAS3 op with BLAS2). That's the whole point of DGEMM vs DGEMV which are optimized to death by every BLAS implementation. Batch vectors vs. matrix multiplication is a semantic that NumPy is enforcing us to think about because the required flexibility is not there. And if `vecdot` is faster than `@` that means the implementation of `@` needs work not that we discovered a better way to multiply batch vectors (which are matrices). Anytime you need something else than `@` means we screwed up somewhere along the line. It's normal and can happen this is a collective effort and maybe didn't had enough resources to do it. But because it does not perform well should not justify the other operation is a better one. 

Because of the mental and memory costs involved, it's not fun to rotate arrays so that the parts you want are the inner dimensions. I fully acknowledge that and suffer similarly. Hence typically I resort back to `einsum` to handle the index gymnastics if I have to do such things. 

But implementing several functions to do matvec (and by the way why `vecmat` too because the transpose is broken? ) is definitely a nonperformant option. You can still choose it if it makes things more readable or you are more comfortable with it but inferiority stays constant 😉. Or go fight with the numerical computing mafia yourself.

ilayn:
After that wall of text, I think here it won't make any difference since probably the previous steps will dominate the performance numbers.

ev-br:
> sum(x*x) is a naive implementation of norm. The squareroot precision loss is not in par with what is possibly lost with an improper implementation of dnrm2 (in LAPACK lingo) when you need it. Because there is no scaling involved and no guard rails are placed in sum(x*x) for underflow or overflow. These are established algorithms by now and hence these linalg tricks won't match the robustness of dnrm2 if the vector data is susceptible to underflow or overflow. That's why Anderson's article is implemented as is in nrm2, lartg and a few more places in LAPACK. We might be implementing linalg.norm in a nonperformant fashion and that's we should fix if it is so.

We seem to be on the same page here. 
- if you're in a linalg context and compute a 2-norm of a vector, use `?nrm2`. (from lapack, not DIY), this way you get robustness from blas experts;
- if you're computing `sum(x * y)` where `x` and `y` are just some one-dimensional containers of numbers, use `sum(x * y)` (or `vecdot(x, y)` as a quick perf optimization -- which is being done here).

> Because there is no way you can match performance of matmul by separating the vectors (essentially replacing a BLAS3 op with BLAS2). That's the whole point of DGEMM vs DGEMV which are optimized to death by every BLAS implementation.

Sure. Inside tight loops or other contexts where you need to squeeze the last drop of performance, you'll need to rewrite the computation to be blas-level-3 rich, use BLAS directly etc. Most likely in a compiled code.

Where performance is not the prime concern, there's a question of being expressive. 
Which is where I believe `vec{doc, mat}` and `matvec` enter the stage.
And if I sound dismissive of performance, I am not. (My background is where a week-long simulation is "small", so believe me, all sorts of _readability be damned I need the flops_ tricks are near and dear).

> Batch vectors vs. matrix multiplication is a semantic that NumPy is enforcing us to think about because the required flexibility is not there. . And if vecdot is faster than @ that means the implementation of @ needs work not that we discovered a better way to multiply batch vectors (which are matrices). Anytime you need something else than @ means we screwed up somewhere along the line

I think numpy semantics is secondary to the math semantics here. 

>>> [vecdot(v1[i, :], v2[i, :]) for i in range(n) ]

is fundamentally different from `matmul(v1, v2)`. You can probably express both in the `np.einsum` notation (even though the former needs a Kroneker delta with the Einsten notation on the whileboard); and low-level both will be coded as some strided loops --- still there's a fundamental difference.

So it's not a matter of performance per se. 
And readablity wise, I think it's hard to beat `vecmat(v, a)` with einsum or anything. 

-------------------------------------------------------------------------------

I kept typing above the line but then stopped and removed 1/2 of the wall of text. Because ISTM i) on a high level we are on about the same page modulo matters of taste (?), and (ii) this all starts to be OT for this specific patch (?).

scipy/_lib/_util.py
mdhaber: I wasn't thinking. I think I meant array API compat vecdot. But yes, `xp.vecdot` would work.

ilayn:
My point is 

```
>>> [v1[i, :] @ v2[i, :] for i in range(n) ]
```

should be the same as `vecdot` so you don't need to use it as a function. 

Infix operators beat any function notation in terms of readability. This is the very reason `@` came in after a truck load of discussion which is not just a shorthand for `matmul` but the dot product everywhere it applies (in particular `.dot()`). But anyways, I agree that the point is getting lost in this PR so let's do it somewhere else. Sorry for the noise.

ev-br:
> My point is
> `>>> [v1[i, :] @ v2[i, :] for i in range(n) ]`
> should be the same as vecdot so you don't need to use it as a function.

Modulo performance. The O(n) loop in python vs compiled code, likely at BLAS level 2 speeds

mdhaber:
I implemented the changes in my self-review and used `np_vecdot` throughout `_kde.py` (https://github.com/scipy/scipy/pull/22616#issuecomment-2697160853 and then some). Looks like CI is happy! I leave the rest of SciPy for follow-up work.

ev-br (approved):
LGTM, thanks Matt.

The bulk of the discussion is on  a general topic, and does not seem to block this impovement specifically.

PR #22617, opened 2025-03-03
CI: minimise disk space usage for oneAPI jobs

### LINKED ISSUE #22615: CI: oneAPI job: `Not enough disk space.`
The oneAPI CI job is failing consistently with:

```
Not enough disk space.
Drive / has 20.4 GB available. 20.7 GB is required.
```

cc @czgdp1807

czgdp1807:
#### Reference issue

Closes https://github.com/scipy/scipy/issues/22615

### DIFF
--- a/.github/workflows/linux_intel_oneAPI.yml
+++ b/.github/workflows/linux_intel_oneAPI.yml
@@ -48,6 +48,29 @@ jobs:
         with:
           submodules: recursive
 
+      - name: cache install
+        id: cache-install
+        uses: actions/cache@1bd1e32a3bdc45362d1e726936510720a7c30a57 # v4.2.0
+        with:
+          path: |
+              /opt/intel/oneapi/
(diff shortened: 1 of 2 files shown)


czgdp1807:
Seems like the same error is happening on Linux as well. Windows approach will not work on Linux because there we need not only Fortran but also C++ components of intel oneAPI. Let me check on my work.

czgdp1807:
This is ready for merge from my end. Tests should pass now.

lucascolley (approved):
thanks for the quick fix here @czgdp1807 !

czgdp1807:
Great. Actually that's the lowest we can go on disk space. Anything below will require doing hacks to remove redundant packages. For now its working so I wouldn't do those.

tylerjereddy:
tentatively adding backport label since the disk space issue shows up in the maintenance branch

PR #22618, opened 2025-03-03
BUG: Fix dual quaternion normalization procedure

AlexanderFabisch:
The dual quaternion normalization procedure introduced with https://github.com/scipy/scipy/pull/22267 seems to be incorrect and slightly inefficient. This PR fixes these issues.

It is incorrect because it uses np.linalg.norm on a quaternion with only a scalar part where it should use np.sqrt for the scalar part, which means that the component of the normalization factor is $\sqrt{r^2} = r$, when it should actually be $\sqrt{r}$. In addition, there was an issue with the real part being updated and then used to compute the dual part while the old real part should be used in the update.

Furthermore, there were some inefficiencies because quaternion multiplication was used at places where a scalar multiplication was sufficient.

So far this bug did not cause any problems because dual quaternions are converted to valid transformation matrices internally.

Ping @scottshambaugh @nmayorov who were involved in the PR that added this module.

### DIFF
--- a/scipy/spatial/transform/_rigid_transform.pyx
+++ b/scipy/spatial/transform/_rigid_transform.pyx
@@ -62,19 +62,6 @@ cdef _compute_se3_log_translation_transform(rot_vec):
     return np.identity(3) - 0.5 * s + k[:, None, None] * s @ s
 
 
-cdef _quaternion_conjugate(quat):
-    conjugate = np.copy(quat)
-    conjugate[:, :3] = -conjugate[:, :3]
-    return conjugate
-
-
-cdef _compose_dual_quaternion(real_part1, dual_part1, real_part2, dual_part2):
-    prod_real = compose_quat(real_part1, real_part2)
-    prod_dual = (np.asarray(compose_quat(real_part1, dual_part2))
-                 + np.asarray(compose_quat(dual_part1, real_part2)))
-    return prod_real, prod_dual
-
-
 def normalize_dual_quaternion(dual_quat):
     """Normalize dual quaternion."""
     real, dual = _normalize_dual_quaternion(
@@ -83,40 +70,30 @@ def normalize_dual_quaternion(dual_quat):
 
 
 cdef _normalize_dual_quaternion(real_part, dual_part):
-    """Ensure that the norm is 1 and that real and dual part are orthogonal."""
-    # the normalization procedure is based on
-    # https://stackoverflow.com/a/76313524/915743
+    """Ensure that unit norm of the dual quaternion.
 
+    The norm is a dual number and must be 1 + 0 * epsilon, which means that
+    the real quaternion must have unit norm and the dual quaternion must be
+    orthogonal to the real quaternion.
+    """
     real_part = np.copy(real_part)
     dual_part = np.copy(dual_part)
 
-    # compute the dual quaternion product of the input and its
-    # component-wise quaternion conjugate
-    real_conjugate = _quaternion_conjugate(real_part)
-    dual_conjugate = _quaternion_conjugate(dual_part)
-    prod_real, prod_dual = _compose_dual_quaternion(
-        real_part, dual_part, real_conjugate, dual_conjugate)
+    real_norm = np.linalg.norm(real_part, axis=1)
 
-    # the squared norm of a dual quaternion is a dual number s + t * epsilon
-    s = prod_real[:, 3]
-    t = prod_dual[:, 3]
+    # special case: real quaternion is 0, we set it to identity
+    zero_real_mask = real_norm == 0.0
+    real_part[zero_real_mask, :4] = [0., 0., 0., 1.]
+    real_norm[zero_real_mask] = 1.0
 
-    prod_real_norm = np.sqrt(s)
-    # special case: invalid real quaternion
-    invalid_real_mask = prod_real_norm == 0.0
-    real_part[invalid_real_mask, :4] = [0., 0., 0., 1.]
-    prod_real_norm[invalid_real_mask] = 1.0
+    # 1. ensure unit real quaternion
+    real_part /= real_norm[:, np.newaxis]
+    dual_part /= real_norm[:, np.newaxis]
 
-    # compute normalization factor: the inverse square root of s + t * epsilon
-    # is a dual number u + v * epsilon
-    u = 1.0 / prod_real_norm[:, np.newaxis]
-    v = -0.5 * t[:, np.newaxis] * u ** 3
+    # 2. ensure orthogonality of real and dual quaternion
+    dual_part -= np.sum(real_part * dual_part, axis=1)[:, np.newaxis] * real_part
 
-    # normalize dual quaternion
-    new_real_part = u * real_part
-    dual_part = u * dual_part + v * real_part
-
-    return new_real_part, dual_part
+    return real_part, dual_part
 
 
 def _create_transformation_matrix(translations, rotation_matrices, single):
--- a/scipy/spatial/transform/_rigid_transform.pyx
+++ b/scipy/spatial/transform/_rigid_transform.pyx
@@ -64,10 +64,7 @@ cdef _compute_se3_log_translation_transform(rot_vec):
 
 cdef _quaternion_conjugate(quat):
     conjugate = np.copy(quat)
-    # we exploit the double cover property of unit rotation quaternions with
-    conjugate[:, 3] = -conjugate[:, 3]
-    # instead of using the mathematically correct conjugate implementation
-    # conjugate[:, :3] = -conjugate[:, :3]
+    conjugate[:, :3] = -conjugate[:, :3]
     return conjugate
 
 
@@ -78,8 +75,18 @@ cdef _compose_dual_quaternion(real_part1, dual_part1, real_part2, dual_part2):
     return prod_real, prod_dual
 
 
+def normalize_dual_quaternion(dual_quat):
+    """Normalize dual quaternion."""
+    real, dual = _normalize_dual_quaternion(
+        dual_quat[..., :4], dual_quat[..., 4:])
+    return np.concatenate((real, dual), axis=-1)
+
+
 cdef _normalize_dual_quaternion(real_part, dual_part):
     """Ensure that the norm is 1 and that real and dual part are orthogonal."""
+    # the normalization procedure is based on
+    # https://stackoverflow.com/a/76313524/915743
+
     real_part = np.copy(real_part)
     dual_part = np.copy(dual_part)
 
@@ -90,22 +97,26 @@ cdef _normalize_dual_quaternion(real_part, dual_part):
     prod_real, prod_dual = _compose_dual_quaternion(
         real_part, dual_part, real_conjugate, dual_conjugate)
 
+    # the squared norm of a dual quaternion is a dual number s + t * epsilon
+    s = prod_real[:, 3]
+    t = prod_dual[:, 3]
+
+    prod_real_norm = np.sqrt(s)
     # special case: invalid real quaternion
-    prod_real_norm = np.linalg.norm(prod_real, axis=1)
     invalid_real_mask = prod_real_norm == 0.0
     real_part[invalid_real_mask, :4] = [0., 0., 0., 1.]
     prod_real_norm[invalid_real_mask] = 1.0
 
-    # compute normalization factor
-    real_inv_sqrt = 1.0 / prod_real_norm
-    dual_inv_sqrt = -0.5 * prod_dual * real_inv_sqrt[:, np.newaxis] ** 3
+    # compute normalization factor: the inverse square root of s + t * epsilon
+    # is a dual number u + v * epsilon
+    u = 1.0 / prod_real_norm[:, np.newaxis]
+    v = -0.5 * t[:, np.newaxis] * u ** 3
 
     # normalize dual quaternion
-    real_part = real_inv_sqrt[:, np.newaxis] * real_part
-    dual_part = real_inv_sqrt[:, np.newaxis] * dual_part + np.asarray(
-        compose_quat(dual_inv_sqrt, real_part))
+    new_real_part = u * real_part
+    dual_part = u * dual_part + v * real_part
 
-    return real_part, dual_part
+    return new_real_part, dual_part
 
 
 def _create_transformation_matrix(translations, rotation_matrices, single):
--- a/scipy/spatial/transform/tests/test_rigid_transform.py
+++ b/scipy/spatial/transform/tests/test_rigid_transform.py
@@ -3,6 +3,7 @@ import pytest
 import numpy as np
 from numpy.testing import assert_allclose
 from scipy.spatial.transform import Rotation, RigidTransform
+from scipy.spatial.transform._rigid_transform import normalize_dual_quaternion
 
 
 def test_repr():
@@ -896,3 +897,12 @@ def test_copy_flag():
     tf = RigidTransform(matrix, normalize=False, copy=False)
     matrix[0, 0] = 2
     assert tf.as_matrix()[0, 0] == 2
+
+
+def test_normalize_dual_quaternion():
+    rng = np.random.default_rng(103213650)
+    dual_quat = rng.normal(size=(1000, 8))
+    dual_quat = normalize_dual_quaternion(dual_quat)
+    assert_allclose(np.linalg.norm(dual_quat[:, :4], axis=1), 1.0, atol=1e-12)
+    assert_allclose(np.einsum("ij,ij->i", dual_quat[:, :4], dual_quat[:, 4:]),
+                    0.0, atol=1e-12)
--- a/scipy/spatial/transform/tests/test_rigid_transform.py
+++ b/scipy/spatial/transform/tests/test_rigid_transform.py
@@ -900,6 +900,10 @@ def test_copy_flag():
 
 
 def test_normalize_dual_quaternion():
+    dual_quat = normalize_dual_quaternion(np.zeros((1, 8)))
+    assert_allclose(np.linalg.norm(dual_quat[0, :4]), 1.0, atol=1e-12)
+    assert_allclose(dual_quat[0, :4] @ dual_quat[0, 4:], 0.0, atol=1e-12)
+
     rng = np.random.default_rng(103213650)
     dual_quat = rng.normal(size=(1000, 8))
     dual_quat = normalize_dual_quaternion(dual_quat)


scottshambaugh:
I'm not practiced enough with dual quats to feel comfortable auditing the algorithm, but the tests do comprehensively prove the correct behavior.

scipy/spatial/transform/_rigid_transform.pyx
```diff
@@ -90,22 +97,26 @@ cdef _normalize_dual_quaternion(real_part, dual_part):
     prod_real, prod_dual = _compose_dual_quaternion(
         real_part, dual_part, real_conjugate, dual_conjugate)
 
+    # the squared norm of a dual quaternion is a dual number s + t * epsilon
+    s = prod_real[:, 3]
+    t = prod_dual[:, 3]
+
+    prod_real_norm = np.sqrt(np.maximum(s, 0.0))
```
nmayorov, line 104: This maximum looks suspicious a little bit.

As far I understand `s` is equal to the sum of squares of the real part coefficients.

Not sure whether it's justified,  but for purity maybe it's better to remove it. If we get a negative number - something clearly went wrong and unexpectedly and needs to be investigated.

scipy/spatial/transform/_rigid_transform.pyx
```diff
@@ -90,22 +97,26 @@ cdef _normalize_dual_quaternion(real_part, dual_part):
...
     real_part[invalid_real_mask, :4] = [0., 0., 0., 1.]
     prod_real_norm[invalid_real_mask] = 1.0
 
-    # compute normalization factor
-    real_inv_sqrt = 1.0 / prod_real_norm
-    dual_inv_sqrt = -0.5 * prod_dual * real_inv_sqrt[:, np.newaxis] ** 3
+    # compute normalization factor: the inverse square root of s + t * epsilon
+    # is a dual number u + v * epsilon
+    u = 1.0 / prod_real_norm[:, np.newaxis]
+    v = -0.5 * t[:, np.newaxis] * u ** 3
```
nmayorov, line 113: This dual number algebra is for sure not very easy to quickly understand.

In more simple terms the unit dual quaternion has unit norm of the real part and the dual part is orthogonal to the real part. My question is: is the implement procedure is equivalent to the normalization of the real part and removing the "along" component from the dual part?

I think I'll do such experiment, and if the result is the same maybe it's better to implement the normalization in these terms, without turning to dual number explicitly.

AlexanderFabisch: I checked the equations. Doing the normalization in two separate steps should be a bit more readable. In the current version, s is the squared norm of the real quaternion. When we divide the whole dual quaternion by sqrt(s) in a first step, this will simplify v to the dot product of the real and dual quaternion, which means in the second step we only subtract dot(real, dual) * real from the dual quaternion.

scipy/spatial/transform/_rigid_transform.pyx
AlexanderFabisch: I added it because I got an error in some of the test cases. Maybe this was caused by another error. We can try to remove it and see if the tests still work.

scipy/spatial/transform/_rigid_transform.pyx
nmayorov: This is a simple function I wrote:

```
def normalize_dual_quaternion(dq):
    p = dq[:, :4].copy()
    q = dq[:, 4:].copy()

    p /= np.linalg.norm(p, axis=1)[:, None]
    q -= np.sum(q * p, axis=1)[:, None] * p

    return np.hstack((p, q))
```

It does return a unit dual quaternion according to its definition. However it does not give the same dual quaternion, but the correction to the dual part is minimal. I'm not sure what is the correct way to normalize the dual quaternion, this is something to think about. Most likely some criteria of optimal normalization must be postulated. For example, normalization of an Euclidean vector - a unit vector closest to the given.

AlexanderFabisch: According to my paper math it should be exactly the same. At least I don't immediately see why the solution should be different. Give me some time to check this in detail tomorrow.

nmayorov: By an analogy, I'm inclined to think that the implement approach is correct, but I'm puzzled what does it mean geometrically.

scipy/spatial/transform/_rigid_transform.pyx
```diff
@@ -90,22 +97,26 @@ cdef _normalize_dual_quaternion(real_part, dual_part):
...
 
     # normalize dual quaternion
-    real_part = real_inv_sqrt[:, np.newaxis] * real_part
-    dual_part = real_inv_sqrt[:, np.newaxis] * dual_part + np.asarray(
-        compose_quat(dual_inv_sqrt, real_part))
+    new_real_part = u * real_part
+    dual_part = u * dual_part + v * real_part
```
nmayorov, line 117: Any intuition why these formulas apply? We multiply a dual number with a dual quaternion like 2 dual numbers basically, correct?

AlexanderFabisch: Yes, the difficulty here is that we have to multiply the dual quaternion with the inverse square root of a dual number (in this case of the squared norm of a dual quat). I have the derivation on paper or you can find some (imo not so intuitive) version of it in the stack overflow answer that I linked at the top of the function. The point is, it's not straightforward to compute the inverse square root of a dual number. I can write it down tomorrow.

AlexanderFabisch:
Here is the mathematical background required to understand the two solutions. I call the simpler one the "two-step solution" at the end.

## Background

Let's first establish some fundamentals. A dual number is a number $r + d \epsilon$ that consists of a real part $r$ and a dual part $d$ with $\epsilon^2 = 0$ and $\epsilon \neq 0$. A quaternion $r + \boldsymbol{v}$ consists of a real part $r$ and a vector part $\boldsymbol{v} \in \mathbb{R}$. A dual quaternion is a combination of the two. A dual quaternion $\sigma = p + q \epsilon$ consists of a real quaternion $p = r_p + \boldsymbol{v}_p$ and a dual quaternion $q = r_q + \boldsymbol{v}_q$. Normalizing a dual quaternion means the dual quaternion should have unit norm, that is, $\sigma \cdot \sigma^* = 1$ (with the dual quaternion product $\cdot$). It is often assumed that it is enough to ensure that the real quaternion has unit norm by dividing the whole dual quaternion by the norm of the real quaternion. This works for most cases, but it is an approximation. We see this immediately, when we write the unit norm property correctly: $\sigma \sigma^* = 1 + 0 \epsilon$. Ensuring that the second component is 0 is not trivial. There is a solution that I found on stackoverflow [here](https://stackoverflow.com/questions/23174899/properly-normalizing-a-dual-quaternion/76313524#76313524), but it lacks some details that I would like to write down here.

The following definitions and properties of quaternions, dual quaternions, and dual numbers are required to understand the solution.

### Quaternion Product

$$q_1 \cdot q_2 = r_1 r_2 - \boldsymbol{v}_1^T \boldsymbol{v}_2 + r_1 \boldsymbol{v}_2 + r_2 \boldsymbol{v}_1 + \boldsymbol{v}_1 \times \boldsymbol{v}_2$$

### Quaternion Conjugate

$$q^* = (r + \boldsymbol{v})^* = r - \boldsymbol{v}$$

### Quaternion Product with Conjugate

From the last two definitions, we can see easily that

$$q \cdot q^* = r^2 + \boldsymbol{v}^T\boldsymbol{v} - r \boldsymbol{v} + r \boldsymbol{v} - \boldsymbol{v} \times \boldsymbol{v} = r^2 + \boldsymbol{v}^T \boldsymbol{v},$$

which is just a real number without the vector part.

### Dual Quaternion Product

$$\sigma_1 \cdot \sigma_2 = (p_1 + q_1 \epsilon) \cdot (p_2 + q_2 \epsilon) = p_1 \cdot p_2 + (p_1 \cdot q_2 + p_2 \cdot q_1) \epsilon$$

### Dual Quaternion Conjugate

There are three different dual quaternion conjugate. We will use the quaternion conjugate of the two individual quaternions.

$$\sigma^* = (p + q \epsilon)^* = p^* + q^* \epsilon$$

## Norm of a Dual Quaternion

Now we can start computing the (squared) norm of a dual quaternion. We need the dual quaternion product and the dual quaternion conjugate:

$$\sigma \cdot \sigma^* = (p + q \epsilon) \cdot (p + q \epsilon)^* = p \cdot p^* + (p q^* + q p^*) \epsilon$$

We can further break this down with $p = r_p + \boldsymbol{v}_p$ and $q = r_q + \boldsymbol{v}_q$ to

$$p \cdot p^* = r_p^2 + \boldsymbol{v}_p^T \boldsymbol{v}_p$$

$$p q^* = r_p r_q + \boldsymbol{v}_p^T \boldsymbol{v}_q - r_p \boldsymbol{v}_q + r_q \boldsymbol{v}_p - \boldsymbol{v}_p \times \boldsymbol{v}_q$$
$$q p^* = r_q r_p + \boldsymbol{v}_q^T \boldsymbol{v}_p - r_q \boldsymbol{v}_p + r_p \boldsymbol{v}_q - \boldsymbol{v}_q \times \boldsymbol{v}_p$$

With $\boldsymbol{v}_p \times \boldsymbol{v}_q = - \boldsymbol{v}_q \times \boldsymbol{v}_p$, we can furthermore simplify
$$p q^* + q p^* = (r_p r_q + \boldsymbol{v}_p^T \boldsymbol{v}_q \underline{- r_p \boldsymbol{v}_q + r_q \boldsymbol{v}_p} - \underline{\boldsymbol{v}_p \times \boldsymbol{v}_q}) + (r_q r_p + \boldsymbol{v}_q^T \boldsymbol{v}_p \underline{- r_q \boldsymbol{v}_p + r_p \boldsymbol{v}_q} - \underline{\boldsymbol{v}_q \times \boldsymbol{v}_p}) = 2 (r_p r_q + \boldsymbol{v}_p^T \boldsymbol{v}_q),$$

which does not contain a vector part anymore.

To summarize, the squared norm of a dual quaternion is a dual number of the form

$$\sigma \cdot \sigma^* = r_p^2 + \boldsymbol{v}_p^T \boldsymbol{v}_p + 2 (r_p r_q + \boldsymbol{v}_p^T \boldsymbol{v}_q) \epsilon,$$

that is, the vector parts of the quaternions are $0$.

When we write $p = (r_p, v_{1p} v_{2p}, v_{3p})^T$ and $q = (r_q, v_{1q} v_{2q}, v_{3q})^T$ as 4d vectors, $r_p^2 + \boldsymbol{v}_p^T \boldsymbol{v}_p = p^T p = ||p||_2^2$ is the squared Euclidean norm of the real quaternion and $r_p r_q + \boldsymbol{v}_p^T \boldsymbol{v}_q = p^T q$ is the dot product of the real and dual quaternion, hence

$$\sigma \cdot \sigma^* = p^T p + 2 p^T q \epsilon.$$

### Example

The following function will compute the norm. It returns an array of shape `(2,)` with the real part in the first component and the dual part in the second component.


```python
import pytransform3d.transformations as pt


def dual_quaternion_norm(sigma):
    return pt.concatenate_dual_quaternions(
        sigma, pt.dq_q_conj(sigma, unit=False), unit=False
    )[[0, 4]]
```

We can also use our simplified expression to implement the norm:


```python
def dual_quaternion_norm_simplified(sigma):
    p = sigma[:4]
    q = sigma[4:]
    real = p @ p
    dual = 2 * p @ q
    return np.array([real, dual])
```

Let's try to generate an arbitrary 8d vector, interpret it as a dual quaternion and compute its norm.


```python
import numpy as np


rng = np.random.default_rng(42)
sigma = rng.normal(size=8)
sigma
```




    array([ 0.30471708, -1.03998411,  0.7504512 ,  0.94056472, -1.95103519,
           -1.30217951,  0.1278404 , -0.31624259])




```python
dual_quaternion_norm(sigma)
```




    array([2.62225842, 1.11644721])




```python
dual_quaternion_norm_simplified(sigma)
```




    array([2.62225842, 1.11644721])



## Normalization Procedure

To normalize a dual quaternion, we have to multiply it with the inverse square root of its squared norm. Computing the inverse square root of a dual number is not trivial. One derivation is outlined [here](https://stackoverflow.com/questions/23174899/properly-normalizing-a-dual-quaternion/76313524#76313524). However, I find the following more easy to understand.

We define the inverse square root of a dual number $s + t \epsilon$ as

$$\frac{1}{\sqrt{s + t \epsilon}} = u + v \epsilon$$

and square the expression on both sides to

$$\frac{1}{s + t \epsilon} = u^2 + 2uv \epsilon.$$

We multiply by $s + t \epsilon$ and get

$$1 + 0 \epsilon = (u^2 + 2uv \epsilon)(s + t \epsilon) = u^2s + (u^2t + 2uvs) \epsilon,$$

which we can split into

$$1 = u^2s \Leftrightarrow u = \pm\frac{1}{\sqrt{s}}$$

$$0 \epsilon = (u^2t + 2uvs) \epsilon \Leftrightarrow v \epsilon = - \frac{0.5 t}{(\pm \sqrt{s})^3} \epsilon$$

Hence, to normalize a dual quaternion, we multiply the dual quaternion with the dual number

$$\frac{1}{\sqrt{s + t \epsilon}} = \pm\left(\frac{1}{\sqrt{s}} - \frac{0.5 t}{(\sqrt{s})^3} \epsilon\right).$$

We will only use the positive solution here, multiplying this normalization factor with the dual quaternion gives us the solution

$$(p + q \epsilon) \left(\frac{1}{\sqrt{s}} - \frac{0.5 t}{(\sqrt{s})^3} \epsilon\right) = \frac{p}{\sqrt{s}} + \left(\frac{q}{\sqrt{s}} - \frac{0.5 t}{(\sqrt{s})^3} p \right) \epsilon.$$

Plugging in $s = p^T p$ and $t = 2 p^T q$ from the dual quaternion norm, we would get the normalization factor

$$\pm\left(\frac{1}{\sqrt{p^T p}} - \frac{p^Tq}{\left(\sqrt{p^Tp}\right)^3} \epsilon\right)$$

### Example


```python
def norm_dual_quaternion(sigma):
    sigma_prod = pt.concatenate_dual_quaternions(
        sigma, pt.dq_q_conj(sigma, unit=False), unit=False
    )
    s = sigma_prod[0]
    t = sigma_prod[4]
    u = 1.0 / np.sqrt(s)
    v = -0.5 * t / np.sqrt(s)**3

    p = sigma[:4] * u
    q = sigma[4:] * u + sigma[:4] * v

    return np.hstack((p, q))
```


```python
def norm_dual_quaternion_simplified(sigma):
    p = sigma[:4]
    q = sigma[4:]
    s = p @ p

    u = 1.0 / np.sqrt(s)
    v = -(p @ q) / np.sqrt(s)**3

    p = sigma[:4] * u
    q = sigma[4:] * u + sigma[:4] * v

    return np.hstack((p, q))
```


```python
sigma_unit = norm_dual_quaternion(sigma)
sigma_unit
```




    array([ 0.18817376, -0.64222759,  0.4634306 ,  0.58083254, -1.24489263,
           -0.66742595, -0.01970857, -0.31893819])




```python
dual_quaternion_norm(sigma_unit).round(3)
```




    array([1., 0.])




```python
sigma_unit = norm_dual_quaternion_simplified(sigma)
sigma_unit
```




    array([ 0.18817376, -0.64222759,  0.4634306 ,  0.58083254, -1.24489263,
           -0.66742595, -0.01970857, -0.31893819])




```python
dual_quaternion_norm(sigma_unit).round(3)
```




    array([ 1., -0.])



## Edge Case

It could also happen that the real quaternion $p = 0$. In this case, $u$ is not defined. We will omit this case here, but the solution would be to set $p = 1$ and then proceed with the normalization procedure.


```python
norm_dual_quaternion(np.zeros(8))
```

    /tmp/ipykernel_5178/3139420.py:7: RuntimeWarning: divide by zero encountered in scalar divide
      u = 1.0 / np.sqrt(s)
    /tmp/ipykernel_5178/3139420.py:8: RuntimeWarning: invalid value encountered in scalar divide
      v = -0.5 * t / np.sqrt(s)**3
    /tmp/ipykernel_5178/3139420.py:10: RuntimeWarning: invalid value encountered in multiply
      p = sigma[:4] * u
    /tmp/ipykernel_5178/3139420.py:11: RuntimeWarning: invalid value encountered in multiply
      q = sigma[4:] * u + sigma[:4] * v





    array([nan, nan, nan, nan, nan, nan, nan, nan])



## Normalization in Two Steps

To simplify the terms involved in the normalization procedure, we can split the normalization up into two steps:

1. Multiply $\sigma$ with the normalization factor $\frac{1}{\sqrt{s}} = \frac{1}{\sqrt{p^T p}}$ for the real quaternion to obtain $\sigma' = \frac{\sigma}{\sqrt{s}}$, which ensures that the norm is $1 + t' \epsilon$.
2. Multiply $\sigma' = p' + q' \epsilon$ with the normalization factor $(1 - \frac{t'}{2} \epsilon)$ for the dual quaternion (since s' = 1) to obtain $\sigma'' = p' + q' \epsilon - p'^T q' p' \epsilon$.


```python
def norm_dual_quaternion_two_steps(sigma):
    p = np.copy(sigma[:4])
    q = np.copy(sigma[4:])

    real_norm = np.sqrt(p @ p)

    p /= real_norm
    q /= real_norm

    q -= p @ q * p

    return np.hstack((p, q))
```


```python
sigma_unit = norm_dual_quaternion_two_steps(sigma)
sigma_unit
```




    array([ 0.18817376, -0.64222759,  0.4634306 ,  0.58083254, -1.24489263,
           -0.66742595, -0.01970857, -0.31893819])




```python
dual_quaternion_norm(sigma_unit).round(3)
```




    array([ 1., -0.])

scipy/spatial/transform/_rigid_transform.pyx
AlexanderFabisch: OK, I removed the maximum. Seems to work fine in the tests. Maybe this was needed with the old version of the quaternion conjugate.

scipy/spatial/transform/_rigid_transform.pyx
AlexanderFabisch: I implemented two alternative versions:

```python
def normalize_dual_quaternion2(dual_quat):
    """Normalize dual quaternion."""
    real, dual = _normalize_dual_quaternion2(
        dual_quat[..., :4], dual_quat[..., 4:])
    return np.concatenate((real, dual), axis=-1)


cdef _normalize_dual_quaternion2(real_part, dual_part):
    """Ensure that the norm is 1 and that real and dual part are orthogonal."""
    real_part = np.copy(real_part)
    dual_part = np.copy(dual_part)

    real_norm = np.linalg.norm(real_part, axis=1)
    # special case: invalid real quaternion
    invalid_real_mask = real_norm == 0.0
    real_part[invalid_real_mask, :4] = [0., 0., 0., 1.]
    real_norm[invalid_real_mask] = 1.0

    # 1. ensure unit real quaternion
    real_part /= real_norm[:, np.newaxis]
    dual_part /= real_norm[:, np.newaxis]

    # 2. ensure orthogonality of real and dual quaternion
    dual_part -= np.sum(real_part * dual_part, axis=1)[:, np.newaxis] * real_part

    return real_part, dual_part


def normalize_dual_quaternion3(dual_quat):
    """Normalize dual quaternion."""
    real, dual = _normalize_dual_quaternion3(
        dual_quat[..., :4], dual_quat[..., 4:])
    return np.concatenate((real, dual), axis=-1)


cdef _normalize_dual_quaternion3(real_part, dual_part):
    """Ensure that the norm is 1 and that real and dual part are orthogonal."""
    # the normalization procedure is based on
    # https://stackoverflow.com/a/76313524/915743

    real_part = np.copy(real_part)
    dual_part = np.copy(dual_part)

    # We compute the dual quaternion product of the input and its
    # component-wise quaternion conjugate, which is the squared norm of the
    # dual quaternion and a dual number s + t * epsilon (without the vector
    # parts of the dual quaternion).
    s = np.sum(real_part * real_part, axis=1)
    t = 2 * np.sum(real_part * dual_part, axis=1)

    real_norm = np.sqrt(s)
    # special case: invalid real quaternion
    invalid_real_mask = real_norm == 0.0
    real_part[invalid_real_mask, :4] = [0., 0., 0., 1.]
    real_norm[invalid_real_mask] = 1.0

    # compute normalization factor: the inverse square root of s + t * epsilon
    # is a dual number u + v * epsilon
    u = 1.0 / real_norm[:, np.newaxis]
    v = -0.5 * t[:, np.newaxis] * u ** 3

    # normalize dual quaternion
    new_real_part = u * real_part
    dual_part = u * dual_part + v * real_part

    return new_real_part, dual_part
```

Checked them with

```python
from scipy.spatial.transform._rigid_transform import normalize_dual_quaternion, normalize_dual_quaternion2, normalize_dual_quaternion3
import numpy as np
q = np.random.randn(10, 8)
(normalize_dual_quaternion(q) - normalize_dual_quaternion2(q)).round(8)
(normalize_dual_quaternion(q) - normalize_dual_quaternion3(q)).round(8)
```

They all seem to be identical. So which version do we want?

AlexanderFabisch: I also see now the difference in your code:

```python
def normalize_dual_quaternion(dq):
    p = dq[:, :4].copy()
    q = dq[:, 4:].copy()

    p /= np.linalg.norm(p, axis=1)[:, None]
    q -= np.sum(q * p, axis=1)[:, None] * p

    return np.hstack((p, q))
```

q has to be divided by the real norm as well.

nmayorov: I'm not arguing here, but why `q` has to be divided? There is some hidden geometrical meaning.

I think the version 2 is the most straightforward and should be the final implementation.

AlexanderFabisch: To construct a dual quaternion from a rotation $q \in \mathcal{S}^3$ and a translation $t \in \mathbb{R}^3$ we compute $$q + 0.5 q t \epsilon,$$
so the real quaternion is part of both individual quaternions.

Geometrically, your code ensures that both quaternions are orthogonal, but without the normalization factor, the dual quaternion will have a different length, which produces a different translation.

nmayorov: Basically if we changed `q` in the first part it at least *makes sense* to adjust its influence to the second part.

nmayorov:
Thanks for the comprehensive overview, very clear and informative.

```
def norm_dual_quaternion_two_steps(sigma):
    p = np.copy(sigma[:4])
    q = np.copy(sigma[4:])

    real_norm = np.sqrt(p @ p)

    p /= real_norm
    q /= real_norm

    q -= p @ q * p

    return np.hstack((p, q))
 ```

I think this should be the final version (with checks of zero `p`), very concise and relies on a familiar orthogonalization formula.

AlexanderFabisch:
> I think this should be the final version (with checks of zero p), very concise and relies on a familiar orthogonalization formula.

I updated the PR.

j-bowhay:
Thank you for your first contribution to SciPy @AlexanderFabisch, keep them coming!

PR #22619, opened 2025-03-03
DOC: stats.gaussian_kde: clarify the meaning of `factor`

### LINKED ISSUE #22581: DOC: stats.gaussian_kde: clarify the meaning of `factor`

The relationship between kde.factor, dataset's covariance and kde.covariance is not clear in the following pages.

https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.gaussian_kde.html
“factor float: The bandwidth factor, obtained from kde.covariance_factor. The square of kde.factor multiplies the covariance matrix of the data in the kde estimation.” needs to be clarified to be:
“factor float: The bandwidth factor, obtained from kde.covariance_factor. The covariance matrix of the data in the kde estimation multiplied by the square of kde.factor is the **kernel covariance matrix (which means the square of the bandwidth in 1d dataset.)**”

https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.gaussian_kde.covariance_factor.html#scipy.stats.gaussian_kde.covariance_factor
“Computes the coefficient (kde.factor) that multiplies the data covariance matrix to obtain the kernel covariance matrix. " needs to be clarified to be:

"Computes the coefficient (kde.factor). **The square of kde.factor** multiplies the data covariance matrix to obtain the kernel covariance matrix." 

It takes me a while to figure out the relationship. The "factor" is not easy to be understanded as "bandwidth". I also see some people confused on the same question at stack overflow: https://stackoverflow.com/questions/23630515/getting-bandwidth-used-by-scipys-gaussian-kde-function

_No response_

_No response_

fbourgey:
Closes #22581


mdhaber:
@rkern would you take a look at this doc suggestion for `gaussian_kde`? Is it correct? Helpful?

mdhaber (commented):
I've made a few additional changes. Once could argue that instead of removing `kde.`, I could have replaced `kde` with `gaussian_kde`. Another judgement call that could have gone the other way is describing the calculation of the kernel covariance matrix (product of data covariance and square of bandwidth factor) in different places rather than only under `covariance`. But at least this refers to `factor` as "bandwidth factor" more consistently and makes it unambiguous that the kernel covariance is the product of `factor**2` and the data covariance.

scipy/stats/_kde.py
```diff
@@ -517,10 +525,10 @@ def set_bandwidth(self, bw_method=None):
         bw_method : str, scalar or callable, optional
             The method used to calculate the estimator bandwidth.  This can be
```
mdhaber, line 526:
```suggestion
            The method used to calculate the bandwidth factor.  This can be
```

scipy/stats/_kde.py
```diff
@@ -500,11 +509,10 @@ def silverman_factor(self):
...
-        multiplies the data covariance matrix to obtain the kernel covariance
-        matrix. The default is `scotts_factor`.  A subclass can overwrite this
+    covariance_factor.__doc__ = """Computes the bandwidth factor `factor`.
+        The default is `scotts_factor`.  A subclass can overwrite this
         method to provide a different method, or set it through a call to
-        `kde.set_bandwidth`."""
+        `set_bandwidth`."""
 
     def set_bandwidth(self, bw_method=None):
         """Compute the estimator bandwidth with given method.
```
mdhaber, line 518:
```suggestion
        """Compute the bandwidth factor with given method.
```

mdhaber (commented):
This isn't all of them, but I think I'm going to call it good for now.

mdhaber:
Thanks for bringing this up @fbourgey! Feel free to submit a new PR with amendments.

PR #22628, opened 2025-03-05
MAINT: stats.tmin/tmax: ensure exact results with unreasonably large integers

### LINKED ISSUE #22626: BUG: scipy.stats: tmin/tmax: loss of precision for large integers
### Describe your issue.

tmin and tmax temporarily convert int to float64 and then back to int. This can lead to loss of precision, even when there are no NaNs in the output.

`array_api_strict` rightfully refuses to cooperate and crashes when the implicit int->float conversion is attempted (xref #22625).

CC @mdhaber 

### Reproducing Code Example

```python
>>> import numpy as np
>>> from scipy.stats import tmin
>>> a = np.arange(10)[None, :]
>>> tmin(a) - a
array([[0, 0, 0, 0, 0, 0, 0, 0, 0, 0]])
>>> b = np.arange(2**62, 2**62+10)[None, :]
>> tmin(b) - b
array([[ 0, -1, -2, -3, -4, -5, -6, -7, -8, -9]])
```

### SciPy version

```shell
1.16.0.dev0+git20250304.60e5d8e
```

mdhaber:
#### Reference issue
Closes gh-22626

gh-22626 noted that the results of `tmin`/`tmax` could produce results with relative errors as high as ~`1e-16` when the calculation could have been exact. This avoids the problem and ensures exact results for integer input by ensuring that the calculations remain integral throughout.

@crusaderky

### DIFF
--- a/scipy/stats/_stats_py.py
+++ b/scipy/stats/_stats_py.py
@@ -614,10 +614,7 @@ def _put_val_to_limits(a, limits, inclusive, val=np.nan, xp=None):
     if xp.all(mask):
         raise ValueError("No array values within given limits")
     if xp.any(mask):
-        # hopefully this (and many other instances of this idiom) are temporary when
-        # data-apis/array-api#807 is resolved
-        dtype = xp.asarray(1.).dtype if xp.isdtype(a.dtype, 'integral') else a.dtype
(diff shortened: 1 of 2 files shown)


j-bowhay (approved):
Changes look resonable and I verified locally that the provided test previously failed

PR #22632, opened 2025-03-05
MAINT: stats.multinomial: use dtype-dependent tolerance

mdhaber:
#### Reference issue
gh-22585

gh-22585 adjusted the condition under which `p` is normalized in `scipy.stats.multinomial`. `multinomial` has converted `p` to `float64` internally since it was introduced, but users have been allowed to pass in input with other dtypes, and the change did not take this into account. This compensates by making the tolerance dtype-dependent.

### DIFF
--- a/scipy/stats/_multivariate.py
+++ b/scipy/stats/_multivariate.py
@@ -3275,12 +3275,13 @@ class multinomial_gen(multi_rv_generic):
         """
         return multinomial_frozen(n, p, seed)
 
-    def _process_parameters(self, n, p, eps=1e-15):
+    def _process_parameters(self, n, p):
         """Returns: n_, p_, npcond.
 
         n_ and p_ are arrays of the correct shape; npcond is a boolean array
         flagging values out of the domain.
         """
+        eps = np.finfo(np.result_type(np.asarray(p), np.float32)).eps * 10
         p = np.array(p, dtype=np.float64, copy=True)
         p_adjusted = 1. - p[..., :-1].sum(axis=-1)
         # only make adjustment when it's significant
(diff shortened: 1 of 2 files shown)


scipy/stats/_multivariate.py
```diff
@@ -3275,12 +3275,13 @@ def __call__(self, n, p, seed=None):
...
         return multinomial_frozen(n, p, seed)
 
-    def _process_parameters(self, n, p, eps=1e-15):
+    def _process_parameters(self, n, p):
         """Returns: n_, p_, npcond.
 
         n_ and p_ are arrays of the correct shape; npcond is a boolean array
         flagging values out of the domain.
         """
+        eps = np.finfo(np.result_type(np.asarray(p), np.float32)).eps * 10
```
mdhaber, line 3284:
```suggestion
        eps = np.finfo(np.result_type(np.asarray(p), np.float16)).eps * 10
```
maybe?

ev-br: can this be just
`eps = xp.finfo(xp_default_dtype(xp)).eps * 10   # or other wiggle factor`

This will not work currently since `xp_default_dtype` only account for torch defaulting to f32; 
Assuming that the failure mode is that jax returns f32 array even when explicitly asked for an f64,  `xp_default_dtype` should grow some logic to account for JAX idiosyncrasy.

mdhaber: There are tests that are passing integers like `[0, 1]` or something.

ev-br: Ok then, that figures. Then indeed, there's no way around `result_type`. Sorry for the noise

mdhaber:
@ev-br marked this as ready given https://github.com/scipy/scipy/pull/22585#issuecomment-2701602064. I suppose we could also require that input be `float64` if that conversion is going to be done anyway, but I'd rather not.

ev-br:
LGTM, merged. Thanks Matt

PR #21505, opened 2024-09-04
BUG: special: Use Boost for ncfdtr, fixing accuracy issues

steppi:
#### Reference issue

Closes #15582, ticks a box for #20223 

#### What does this implement/fix?

This PR changes the scalar kernel for `ncfdtr` to use a wrapper of the more accurate Boost implementation which was already being used in stats for `ncf.cdf`. Previously there were two ufuncs a public `ncfdtr` using cdflib, and a private `_ncf_cdf` using Boost. I've changed it so there is no only one ufunc, `ncfdtr` which uses Boost. I had to change the argument order for `_ncf_cdf`, which differed from that of `ncfdtr`, but this was not difficult because `_ncf_cdf` is only used in one place in stats. I also had to add `NAN` handling in the Boost wrapper.

I had to `xfail` round trip tests for `ncfdtri`, `ncfdtridfn`, `ncfdtridfn`, and `ncfdtrinc`, because these functions invert the less accurate `cdflib` implementation of `ncfdtr`, and therefore now fail. We can fix this after https://github.com/scipy/scipy/pull/21454 lands and we are able to have these functions invert the Boost implementation now being used.

I added some tests comparing `ncfdtr` against an `mpmath` reference. 11 of the 12 test cases I added fail on main. Locally I tested a larger range consisting of 625 cases. On `main` the deciles for relative error over these cases are

```
array([0.00000000e+00, 1.15300845e-14, 4.37093510e-13, 5.72470839e-10,
       1.48813582e-08, 7.50889455e-08, 2.50213950e-07, 1.19707132e-06,
       1.48768369e-05, 3.80639979e-05, 2.12464729e-04])
```

using Boost, the deciles for relative error are

```
array([0.00000000e+00, 0.00000000e+00, 1.40500993e-16, 2.52732515e-16,
       4.34020841e-16, 6.33097261e-16, 1.09243091e-15, 1.53843495e-15,
       2.34037246e-15, 6.13809455e-15, 8.41339873e-14])
```

The max relative error for Boost is better than the 20th percentile for cdflib.

The script generating the test cases is in the comments for the test I added to `special/tests/test_cdflib.py`.
#### Additional information
As suggested by @dschmitz89 here, https://github.com/scipy/scipy/issues/15582#issuecomment-2177581310, we should follow up by using the Boost implementations instead of cdflib when stats is already using Boost, and investigate Boost or other alternatives to cdflib for other functions.)

PR #22636, opened 2025-03-05
MAINT: fix domain check for `ncfdtri`

dschmitz89:
#21505 introduced a small typo: if parameters were outside of the correct domain, `ncfdtri` would raise the error for `ncfdtr` instead. Noticed this while working on #22539 .


steppi (approved):
Thanks @dschmitz89. This looks good to me.

Issue #20678, opened 2024-05-09
ENH: signal: add array API support


lucascolley:
Towards gh-18867.

Test module checklist:

- [ ] test_result_type.py
- [ ] test_dltisys.py
- [ ] test_waveforms.py
- [ ] test_fir_filter_design.py
- [ ] test_windows.py gh-20668
- [ ] test_czt.py
- [ ] test_max_len_seq.py
- [ ] test_wavelets.py (deprecated?)
- [ ] test_spectral.py gh-19001
- [ ] test_ltisys.py
- [ ] test_upfirdn.py
- [ ] test_filter_design.py
- [ ] test_savitzky_golay.py
- [ ] test_array_tools.py
- [ ] test_signaltools.py
- [ ] test_bsplines.py
- [ ] test_peak_finding.py
- [ ] test_cont2discrete.py
- [ ] test_short_time_fft.py

cc @tylerjereddy @jakevdp

ev-br:
cross-ref https://github.com/scipy/scipy/pull/21512

lucascolley:
@ev-br is there anything blocking us from adding PyTorch re-wrapping to the delegation layer in `signal`, like we have for `ndimage`? The equivalent of https://github.com/scipy/scipy/blob/2c16b8bfab4d98a824d4723aebd133a402b53c83/scipy/ndimage/_support_alternative_backends.py#L49-L65 seems to be missing.

ev-br:
I don't think there is anything blocking, no. The two versions of `_support_alternative_backends` should be merged together in fact.

--------------------------------------------

Now that we're on the topic of delegation / `support_alternative_backends`, there are two more possible improvements:

1. In `signal`, we call `array_namespace` twice: once in the delegator, and the second time in the function itself. Instead, the delegator should not only figure the namespace; it should also convert the array arguments so that the delegated function receives the already-converted arguments and the namespace. This will eliminate a class of bugs where we forget to call `np.asarray` in the implementation.

2. Currently, the delegation logic is "delegate to cupy/jax if the matching function is available, convert to numpy otherwise". This fails for e.g. torch GPU tensors. We can extend the delegation logic to e.g. convert to CuPy if the device is compatible.

lucascolley:
Are we happy with routing PyTorch inputs through CuPy?

ev-br:
[...]

PR #22639, opened 2025-03-06
ENH: special: `support_alternative_backends` on Dask and jax.jit

crusaderky:
This is part of a set of three PRs, which can be merged in any order with minor conflicts:
- #22641
- #22639
- #22718

Branch that contains all three: https://github.com/crusaderky/scipy/tree/special_staging

# In this PR
- Add no-materialization Dask support for `scipy.special.log_ndtr`, `ndtr`, `ndtri`, `erf`, `erfc`, `i0`, `i0e`, `i1`, `i1e`, `gammaln`, `gammainc`, `gammaincc`, `logit`, `expit`, `entr`, `chdtr`, `chdtrc`, `betainc`, `betaincc`, `stdtr`, `stdtrit`
- Suppress warnings in Dask for `scipy.special.xlogy`
- Test that all the above functions are compatible with `jax.jit` too
- Dask and [marray](https://github.com/mdhaber/marray) arrays wrapping around non-numpy objects - e.g. cupy - should now work with these functions. This feature remains untested and shouldn't be advertised for now as it requires fundamental groundwork in order to be offered to the public (XREF https://github.com/data-apis/array-api-compat/pull/270) CC @mdhaber


scipy/special/_support_alternative_backends.py
```diff
@@ -40,15 +41,31 @@ def get_array_special_func(f_name, xp, n_array_args):
...
+                *array_args
+            )
+
         else:
+            assert array_args  # is_numpy(xp) is True
+            device = xp_device(array_args[0])
             array_args = [np.asarray(arg) for arg in array_args]
             out = _f(*array_args, *other_args, **kwargs)
-            return _xp.asarray(out)
+            return _xp.asarray(out, device=device)
```
crusaderky, line 68: As I expect wrapper arrays to proliferate in the medium term future (@lucascolley's https://github.com/quantity-dev/quantity-array will soon join the party), this pattern is not sustainable in the long run.

I'm unsure about how to solve it, as there is a fundamental problem with wrapper-specific metadata.
Namely, quantity-array would need to know what kind of operation each binop ufunc performs: a+b (need to align data and metadata before unwrapping)? a*b? a/b? (different operation on metadata and possibly data scaling).

crusaderky, line 68: This device is parameter is untested, as only torch CPU and array-api-strict arrive here. So it only benefits an unknown non-CPU array with no guard against implicit device<->cpu transfers.

scipy/special/tests/test_support_alternative_backends.py
```diff
@@ -1,11 +1,24 @@
...
 from scipy._lib._array_api_no_0d import xp_assert_close
-from scipy._lib._array_api import is_cupy, is_dask, is_jax, is_torch, SCIPY_DEVICE
+from scipy._lib._array_api import (is_cupy, is_dask, is_jax, is_torch,
+                                   SCIPY_DEVICE, SCIPY_ARRAY_API)
 from scipy._lib.array_api_compat import numpy as np
+from scipy._lib.array_api_extra.testing import lazy_xp_function
+
+
+# When SCIPY_ARRAY_API is False, the functions are just ufuncs
+if SCIPY_ARRAY_API:
```
crusaderky, line 19: FIXME https://github.com/data-apis/array-api-extra/pull/153

scipy/special/_support_alternative_backends.py
lucascolley: Yes, I've thought about this myself a bit. I think quantity-array should provide a function which takes a callable and some standardised representation of how the output units are a function of the input units, and returns a Quantity with the output of that callable as the value and the desired units. I think this sort of standardised representation is within the scope of the broader quantity-dev work.

lucascolley: Perhaps it would also be nice to `marray.apply_func(f, (a, b), mask="or")` or similar.

crusaderky: > Perhaps it would also be nice to have `marray.apply_func(f, (a, b), mask="or")` or similar.

I don't know the details of how mask propagation is supposed to work - could there be a ternary operation where the output mask is a more complex function, e.g. `a and b or c`?

crusaderky: Answering my own question by reading https://github.com/mdhaber/marray/blob/main/marray/__init__.py: every operation has its own arbitrary, point-by-point mask propagation rules.

mdhaber: Almost all special functions - and certainly all the ones we are treating with `support_alternative_backends` right now - are elementwise functions, so we can restrict our attention to the scalar case. If we have a function `f(a, b, c, d, ...)` that really depends on all the parameters, and we are missing one of them, then we cannot compute the value of the function. This is where the mask behavior comes from. 

It was not designed to cover *all* cases that might arise in `scipy.special`, and it wasn't meant to be extendable to other libraries; it just does what we've needed so far.

mdhaber: > every operation has its own arbitrary, point-by-point mask propagation rules.

I'm not sure if I'm understanding correctly, but that's not how I would express it. All the elementwise functions are handles with this block in a big loop.

https://github.com/mdhaber/marray/blob/500d1027a8d633fecf298fc29527daca7edb08eb/marray/__init__.py#L305-L337

crusaderky: I'm looking at `xp.where`:
```python
        mask = xp.where(condition.data,
                        condition.mask | x1.mask,
                        condition.mask | x2.mask)
```
Do I understand correctly that a generic, agnostic `condition.mask | x1.mask | x2.mask` would not be _wrong_, and that the above is just more precise? In other words, that having a false-positive on the mask is acceptable?

mdhaber: If there is a standards document for masked array behavior, i'd really like to know about it! But I'm not aware of one, so I don't think it's a matter of right and wrong, but what's useful.

I think `where` fails the premise from above: 

> If we have a function $f(a, b, c, d, ...)$ that really depends on all the parameters...

in that the output doesn't depend on all the parameters, even in principle. For instance, `np.where(True, 1, x)` doesn't depend on the value of `x` at all - even a NaN won't propagate. So right or wrong, I don't think sharing the mask there would be the most useful behavior, and I think users would eventually complain.

Certainly there are special function for which, at particular values of one of the arguments, the value of one of the other arguments almost doesn't matter. But I think we'd still want to propagate a NaN in such cases, and I'd guess that we'd still propagate a mask, although depending on the situation, users might have a case to argue that we shouldn't.

crusaderky:
All CI failures are unrelated. @mdhaber this is ready for review and merge.

lucascolley (approved):
LGTM!

crusaderky:
> `special` is relatively simple because essentially all arguments are arrays and everything works elementwise

Does this stand true for all `special` functions though? For example `logsumexp` (bad example, as it is implemented agnostically in array API) performs reductions.

mdhaber:
> Does this stand true for all special functions though?

No. By "essentially all", I mean almost all, and that was intended to cover the statement about working  elementwise. We discussed this in the context of `marray` support in special, but I was clarify the wording above to avoid confusion.

ev-br:
> The support_alternative_backends systems in signal and ndimage ended up looking pretty different. We had talked about a common system for all modules, so if the special one is being rewritten, it might be a good time to revisit that. (CC @ev-br in case you're still interested.)

I still am, yes. In the language of signal and ndimage, the system in special has just a few delegators: `def single_arg_signature(x, *args, **kwds): return array_namespace(x)` and `def two_arg_signature(x, y, *args, **kwds): return array_namespace(x, y)`. 

It might make sense though to consider https://github.com/scipy/scipy/issues/20678#issuecomment-2764563367, esp the output rewrapping. Does it make sense in scipy.special?

crusaderky:
> > The support_alternative_backends systems in signal and ndimage ended up looking pretty different. We had talked about a common system for all modules, so if the special one is being rewritten, it might be a good time to revisit that. (CC @ev-br in case you're still interested.)
> 
> I still am, yes. In the language of signal and ndimage, the system in special has just a few delegators: `def single_arg_signature(x, *args, **kwds): return array_namespace(x)` and `def two_arg_signature(x, y, *args, **kwds): return array_namespace(x, y)`.
> 
> It might make sense though to consider [#20678 (comment)](https://github.com/scipy/scipy/issues/20678#issuecomment-2764563367), esp the output rewrapping. Does it make sense in scipy.special?

As a rule of thumb it makes sense to use the same dispatch mechanism everywhere.
The important caveat is that (most of) `scipy.special` is elementwise. This means that a few blanket statements are possible:
- You know in advance the output shape will be the `broadcast_shapes` of the inputs (important for Dask and JAX)
- You know in advance that each point of the output is a function of the matching point of the input and nothing more (important for MArray and Dask)

I hadn't thought about dispatching between GPU backends - that's a pretty neat thing to do. I just would be careful about dispatch priority (namely, when inside jax.jit I expect an Array API agnostic implementation to be much faster than CuPy or PyTorch).

ev-br:
> As a rule of thumb it makes sense to use the same dispatch mechanism everywhere.

+1 to this, with some fine print. 

First, I rather strongly believe that scipy's delegation story must be as straightforward as possible. If there are multiple options with potential perf implications, we should use one strategy by default and give users a way to override this choice explicitly. No more, not until we gain more collective experience with dispatch/delegation across the ecosystem.

Second, with scipy.special specifically. ISTM we should factor xsf development into consideration. Last I heard, the plan was for essentially all elementwise kernels to migrate to `xsf`, which will be a dual CPU / GPU C++ codebase. Then if all relevant libraries grew an `xsf`-powered layer, all scipy needs doing is to delegate to that layer?

lucascolley:
> Last I heard, the plan was for essentially all elementwise kernels to migrate to xsf, which will be a dual CPU / GPU C++ codebase. Then if all relevant libraries grew an xsf-powered layer, all scipy needs doing is to delegate to that layer?

I imagine we will need to delegate to `jax.scipy.special` to have any hope of things working with `jax.jit` where a pure Python kernel would kill performance.

rgommers:
`xsf` isn't relevant here at all I think, it's a C++ library that will be internal to whatever array library uses it, one cannot delegate to that. It doesn't matter whether a `somelibrary.special.func` is backed by `xsf` or not, only whether `special.func` exists in `somelibrary`.

ev-br:
> xsf isn't relevant here at all I think, it's a C++ library that will be internal to whatever array library uses it, one cannot delegate to that. 

Of course. `cupyx.scipy.special`, and that it's powered by `xsf` is an implementation detail.
All this assuming the plans of xsf authors did not change, given that the last CuPy PR I see now was in October 2024, https://github.com/cupy/cupy/pull/8620.

crusaderky:
I think that the grand unified dispatch design that is being discussed above should be a topic for a follow-up, not this PR.
There are several big benefits that merging this PR as-is delivers today, listed [in the opening comment](https://github.com/scipy/scipy/pull/22639#issue-2900347763).

lucascolley:
> I think @lucascolley even mentioned that NetworkX and others were working on some sort of common system, and we should use that? (Do you know what I'm talking about, Lucas? Is that still relevant, or do we still need to roll our own system?)

I don't think I ever read this at the time of the comment, sorry @mdhaber. That's probably referring to https://github.com/scientific-python/spatch/issues/1, but it looks like that hasn't had any activity in a while. @betatim or @stefanv may be able to tell you more about https://github.com/scikit-image/scikit-image/pull/7520 and the prospects for generalising further.

stefanv:
We implemented whole-function dispatching in NetworkX and scikit-image, and I suspect at some point we'll try to find commonalities and repackage it as spatch.

PR #22641, opened 2025-03-06
ENH: special: add Dask support to `rel_entr` 

crusaderky:
This is part of a set of three PRs, which can be merged in any order with minor conflicts:
- #22641
- #22639
- #22718

Branch that contains all three: https://github.com/crusaderky/scipy/tree/special_staging

# In this PR
- Add support for `rel_entr` on Dask
- Fix `rel_entr` behaviour vs. nan and inf
- Fix `rel_entr` behaviour vs. mismatched int and float args

Tests in #22718.


scipy/special/tests/test_support_alternative_backends.py
```diff
@@ -17,28 +17,6 @@ def test_dispatch_to_unrecognized_library():
     xp_assert_close(res, ref)
 
 
-@pytest.mark.parametrize('dtype', ['float32', 'float64', 'int64'])
-def test_rel_entr_generic(dtype):
```
crusaderky, line 21: I understand that float32 and float64 is redundant with `test_support_alternative_backends` just below.
int64 is improper as discussed above.

scipy/_lib/array_api_extra
crusaderky, line 1: DNM Bump to https://github.com/data-apis/array-api-extra/pull/141

rgommers: This PR does look pretty clean. Over to the linked PR then:)

scipy/special/tests/test_support_alternative_backends.py
```diff
@@ -18,84 +22,100 @@ def test_dispatch_to_unrecognized_library():
...
+        # gh-20972
+        or ((is_cupy(xp) or is_jax(xp) or is_torch(xp)) and f_name == 'chdtrc')):
+        positive_only = True
+    else:
+        positive_only = False
+    return positive_only
 
+
+@pytest.mark.parametrize('f_name,n_args', array_special_func_map.items())
+@pytest.mark.filterwarnings("ignore:invalid value encountered:RuntimeWarning:dask")
```
crusaderky, line 56: this line will go away in #22639

scipy/special/_support_alternative_backends.py
```diff
@@ -53,25 +55,24 @@ def __f(*args, _f=_f, _xp=xp, **kwargs):
     return __f
 
 
-def _get_shape_dtype(*args, xp):
```
mdhaber, line 56: This was going to happen in gh-22695, of course. Would you take a look at that again now that your comments have been addressed?

scipy/special/_support_alternative_backends.py
```diff
@@ -53,25 +55,24 @@ def __f(*args, _f=_f, _xp=xp, **kwargs):
...
+        xy_inf = xp.isinf(x) & xp.isinf(y)
+        res = xpx.apply_where(
+            xy_pos & ~xy_inf,
+            (x, y),
+            # Note: for very large x, this can overflow.
+            lambda x, y: x * (mxp.log(x) - mxp.log(y)),
+            fill_value=xp.inf
+        )
+        res = xpx.at(res)[(x == 0) & (y >= 0)].set(0)
+        res = xpx.at(res)[(x != x) | (y != y) | (xy_pos & xy_inf)].set(xp.nan)
```
mdhaber, line 73: I'd like to avoid these. Will this work?
```suggestion
        res = xpx.at(res)[xp.isnan(x) | xp.isnan(y) | (xy_pos & xy_inf)].set(xp.nan)
```

mdhaber (approved):

crusaderky:
CI failure is unrelated.
@mdhaber ready for merge if you're happy

scipy/special/_support_alternative_backends.py
mdhaber: Right, I knew NumPy would go through the compiled code. I think the `np.errstate` context manager would work with the other backends, but not Dask IIRC. Is it possible for *a* context manager to suppress the Dask warning, even if it doesn't exist yet?

crusaderky: Dask would need to implement capturing the np.errstate context and replicate it inside the tasks. It's technically doable, but quite hacky.

mdhaber: I ask because warnings - even just from NumPy - have been a point of consternation for a long time. 

Suppose the context manager were to work with all backends. Would you still prefer `apply_where` over a context manager?

rgommers: I think there was an open Dask issue for this and it's a recurring issue, so my impression is that it seems worth doing inside Dask despite it being hacky.

mdhaber: dask/dask#3245

PR #22642, opened 2025-03-06
BUG: sparse: sparse sum/mean out parameter shape not enforced for 1D

dschult:
I found that for 1D sparse arrays, the `sum` and `mean` methods do not enforce their `out` parameters to be the right size. Specifically, `A.sum(out=out)` works for `out` with shape in `[(), (1,), (1,1)]`. The tests make sure it works for `(1,)` but don't test cases that shouldn't work. And the case it makes sure works, shouldn't work if we want to match numpy.

Fixed test and added new tests of invalid input and fixed sum/mean methods. While putting in the fix, I noticed that both 1D and 2D code attempts to check the `out` parameter, but we can let numpy do that instead (nice error message, ensures same behavior as numpy, probably faster, etc.). This also simplifies the code for our `sum` method in a few modules.

I doubt anyone is counting on the lack of enforcement of 1D out parameters for sum and mean, yet. But they eventually would. And this cleans up sum/mean code a little too.

@perimosocordiae this is a side-track on the way toward an improved `validateaxes` method.

### DIFF
--- a/scipy/sparse/_base.py
+++ b/scipy/sparse/_base.py
@@ -1150,13 +1150,8 @@ class _spbase(SparseABC):
         if self.ndim == 1:
             if axis not in (None, -1, 0):
                 raise ValueError("axis must be None, -1 or 0")
-            ret = (self @ np.ones(self.shape, dtype=res_dtype)).astype(dtype)
-
-            if out is not None:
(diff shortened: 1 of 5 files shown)


perimosocordiae (approved):
Very nice! Always happy to let numpy do the validation for us.

Issue #22589, opened 2025-02-25
BUG: spatial: `Rotation` no longer supports zero-length collections


matthewchilders77:
### Describe your issue.

Between 1.14.1 and 1.15, the `scipy.spatial.transform.Rotation` class dropped support for zero-length collections of rotations, i.e., rotations where `single=False` and the quaternions are represented by the 0x4 empty 2D array. It is explicitly excluded from supported instantiation [here](https://github.com/scipy/scipy/blob/0fb2a7fc3bb18e21bdcc36ea086a1ea69f0b0d6b/scipy/spatial/transform/_rotation.pyx#L841).

Support for zero-length collections is important so that `Rotation.__getitem__` is compatible with advanced indexing. For example, consider an N-element collection of rotations, e.g., `rotations = Rotation.identity(4)`. We can index on this with an N-element boolean mask, e.g.,  `mask=[False, False, False, False]` to retrieve only the rotations at the specified indices. Not supporting the zero-length collection breaks fully generalized indexing on rotation collections and is a breaking change relative to v1.14.1.

### Reproducing Code Example

```python
from scipy.spatial.rotation import Rotation

rotations = Rotation.identity(4)
mask = [False, False, False, False]

selected_rotations = rotations[mask]
```

### Error message

```shell
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "_rotation.pyx", line 3165, in scipy.spatial.transform._rotation.Rotation.__getitem__
  File "_rotation.pyx", line 837, in scipy.spatial.transform._rotation.Rotation.__init__
ValueError: Expected `quat` to have shape (4,) or (N, 4), got (0, 4).
```

### SciPy/NumPy/Python version and system information

```shell
1.15.2 2.2.3 sys.version_info(major=3, minor=11, micro=5, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    detection method: pkgconfig
    found: true
    include directory: C:/Users/runneradmin/AppData/Local/Temp/cibw-run-chm4s13i/cp311-win_amd64/build/venv/Lib/site-packages/scipy_openblas32/include
    lib directory: C:/Users/runneradmin/AppData/Local/Temp/cibw-run-chm4s13i/cp311-win_amd64/build/venv/Lib/site-packages/scipy_openblas32/lib
    name: scipy-openblas
    openblas configuration: OpenBLAS 0.3.28 DYNAMIC_ARCH NO_AFFINITY Haswell MAX_THREADS=24
[...]

PR #22643, opened 2025-03-06
BUG: spatial.transform.Rotation: support 0-length rotations

hubernikus:
#### Reference issue
https://github.com/scipy/scipy/issues/22589

Enable 0-length Rotation.
A `Rotation` with `self_single = False` should support any number of rotations from 0 ... infinity. 
Similarly to how numpy is able to handle arrays with dimension of size 0. 

#### Additional information
I created a single test to check all the zero-Rotation functionalities (but could be split among existing test, if that is preferred).
The tests for the creation methods include so far quaternion, euler and rotation matrix. I do not expect other methods to fail, since the modification was mostly on the cython.view.array creation.

Please point out if other methods need explicit checking!

### DIFF
--- a/scipy/spatial/transform/_rotation.pyx
+++ b/scipy/spatial/transform/_rotation.pyx
@@ -3762,7 +3762,7 @@ class Slerp:
         if not isinstance(rotations, Rotation):
             raise TypeError("`rotations` must be a `Rotation` instance.")
 
-        if rotations.single or len(rotations) == 1:
+        if rotations.single or len(rotations) <= 1:
             raise ValueError("`rotations` must be a sequence of at least 2 rotations.")
 
         times = np.asarray(times)
--- a/scipy/spatial/transform/_rotation.pyx
+++ b/scipy/spatial/transform/_rotation.pyx
@@ -456,7 +456,12 @@ cdef inline void _compose_quat_single( # calculate p * q into r
 cdef inline double[:, :] _compose_quat(
     const double[:, :] p, const double[:, :] q
 ) noexcept:
-    cdef Py_ssize_t n = max(p.shape[0], q.shape[0])
+    cdef Py_ssize_t n
+    if p.shape[0] == 1:
+        n = q.shape[0]
+    else:
+        n = p.shape[0]
+         
     cdef double[:, :] product = _empty2(n, 4)
 
     # dealing with broadcasting
--- a/scipy/spatial/transform/_rotation.pyx
+++ b/scipy/spatial/transform/_rotation.pyx
@@ -850,8 +850,7 @@ cdef class Rotation:
         quat = np.asarray(quat, dtype=float)
 
         if (quat.ndim not in [1, 2]
-            or quat.shape[len(quat.shape) - 1] != 4
-            or quat.shape[0] == 0):
+            or quat.shape[len(quat.shape) - 1] != 4):
             raise ValueError("Expected `quat` to have shape (4,) or (N, 4), "
                              f"got {quat.shape}.")
 
--- a/scipy/spatial/transform/_rotation.pyx
+++ b/scipy/spatial/transform/_rotation.pyx
@@ -456,11 +456,7 @@ cdef inline void _compose_quat_single( # calculate p * q into r
 cdef inline double[:, :] _compose_quat(
     const double[:, :] p, const double[:, :] q
 ) noexcept:
-    cdef Py_ssize_t n
-    if p.shape[0] == 1:
-        n = q.shape[0]
-    else:
-        n = p.shape[0]
+    cdef Py_ssize_t n = q.shape[0] if p.shape[0] == 1 else p.shape[0]
          
     cdef double[:, :] product = _empty2(n, 4)
 
--- a/scipy/spatial/transform/_rotation.pyx
+++ b/scipy/spatial/transform/_rotation.pyx
@@ -2958,7 +2958,7 @@ cdef class Rotation:
         array([0.24945696, 0.25054542, 0.24945696])
         """
         if self._quat.shape[0] == 0:
-            raise ValueError("Mean of empty rotation.")
+            raise ValueError("Mean of an empty rotation set is undefined.")
             
         if weights is None:
             weights = np.ones(len(self))
--- a/scipy/spatial/transform/_rotation.pyx
+++ b/scipy/spatial/transform/_rotation.pyx
@@ -2958,7 +2958,7 @@ cdef class Rotation:
         array([0.24945696, 0.25054542, 0.24945696])
         """
         if self._quat.shape[0] == 0:
-            raise RuntimeWarning("Mean of empty rotation.")
+            raise ValueError("Mean of empty rotation.")
             
         if weights is None:
             weights = np.ones(len(self))
--- a/scipy/spatial/transform/_rotation.pyx
+++ b/scipy/spatial/transform/_rotation.pyx
@@ -15,12 +15,20 @@ np.import_array()
 
 # utilities for empty array initialization
 cdef inline double[:] _empty1(int n) noexcept:
+    if n == 0:
+        return array(shape=(1,), itemsize=sizeof(double), format=b"d")[:0]
     return array(shape=(n,), itemsize=sizeof(double), format=b"d")
 cdef inline double[:, :] _empty2(int n1, int n2) noexcept :
+    if n1 == 0:
+        return array( shape=(1, n2), itemsize=sizeof(double), format=b"d" )[:0]
     return array(shape=(n1, n2), itemsize=sizeof(double), format=b"d")
 cdef inline double[:, :, :] _empty3(int n1, int n2, int n3) noexcept:
+    if n1 == 0:
+        return array(shape=(1, n2, n3), itemsize=sizeof(double), format=b"d" )[:0]
     return array(shape=(n1, n2, n3), itemsize=sizeof(double), format=b"d")
 cdef inline double[:, :] _zeros2(int n1, int n2) noexcept:
+    if n1 == 0:
+        return array(shape=(1, n2), itemsize=sizeof(double), format=b"d")[:0]
     cdef double[:, :] arr = array(shape=(n1, n2),
         itemsize=sizeof(double), format=b"d")
     arr[:, :] = 0
--- a/scipy/spatial/transform/_rotation.pyx
+++ b/scipy/spatial/transform/_rotation.pyx
@@ -20,11 +20,11 @@ cdef inline double[:] _empty1(int n) noexcept:
     return array(shape=(n,), itemsize=sizeof(double), format=b"d")
 cdef inline double[:, :] _empty2(int n1, int n2) noexcept :
     if n1 == 0:
-        return array( shape=(1, n2), itemsize=sizeof(double), format=b"d" )[:0]
+        return array( shape=(1, n2), itemsize=sizeof(double), format=b"d")[:0]
     return array(shape=(n1, n2), itemsize=sizeof(double), format=b"d")
 cdef inline double[:, :, :] _empty3(int n1, int n2, int n3) noexcept:
     if n1 == 0:
-        return array(shape=(1, n2, n3), itemsize=sizeof(double), format=b"d" )[:0]
+        return array(shape=(1, n2, n3), itemsize=sizeof(double), format=b"d")[:0]
     return array(shape=(n1, n2, n3), itemsize=sizeof(double), format=b"d")
 cdef inline double[:, :] _zeros2(int n1, int n2) noexcept:
     if n1 == 0:
--- a/scipy/spatial/transform/_rotation.pyx
+++ b/scipy/spatial/transform/_rotation.pyx
@@ -20,7 +20,7 @@ cdef inline double[:] _empty1(int n) noexcept:
     return array(shape=(n,), itemsize=sizeof(double), format=b"d")
 cdef inline double[:, :] _empty2(int n1, int n2) noexcept :
     if n1 == 0:
-        return array( shape=(1, n2), itemsize=sizeof(double), format=b"d")[:0]
+        return array(shape=(1, n2), itemsize=sizeof(double), format=b"d")[:0]
     return array(shape=(n1, n2), itemsize=sizeof(double), format=b"d")
 cdef inline double[:, :, :] _empty3(int n1, int n2, int n3) noexcept:
     if n1 == 0:
--- a/scipy/spatial/transform/_rotation.pyx
+++ b/scipy/spatial/transform/_rotation.pyx
@@ -2957,6 +2957,9 @@ cdef class Rotation:
         >>> r.mean().as_euler('zyx', degrees=True)
         array([0.24945696, 0.25054542, 0.24945696])
         """
+        if self._quat.shape[0] == 0:
+            raise RuntimeWarning("Mean of empty rotation.")
+            
         if weights is None:
             weights = np.ones(len(self))
         else:
--- a/scipy/spatial/transform/tests/test_rotation.py
+++ b/scipy/spatial/transform/tests/test_rotation.py
@@ -2133,7 +2133,7 @@ def test_zero_rotation_concatentation():
     assert len(r4) == 4
 
     for pp in [-1.5, -1, 0, 1, 1.5]:
-        pow0 = r0**0
+        pow0 = r**pp
         assert len(pow0) == 0
 
     # Methods
--- a/scipy/spatial/transform/tests/test_rotation.py
+++ b/scipy/spatial/transform/tests/test_rotation.py
@@ -1631,18 +1631,23 @@ def test_slerp_rot_is_rotation():
         t = np.array([0, 1])
         Slerp(t, r)
 
+SLERP_MESSAGE = "must be a sequence of at least 2 rotations"
 
 def test_slerp_single_rot():
-    msg = "must be a sequence of at least 2 rotations"
-    with pytest.raises(ValueError, match=msg):
-        r = Rotation.from_quat([1, 2, 3, 4])
+    r = Rotation.from_quat([1, 2, 3, 4])
+    with pytest.raises(ValueError, match=SLERP_MESSAGE):
         Slerp([1], r)
 
 
+def test_slerp_rot_len0():
+    r = Rotation.random()
+    with pytest.raises(ValueError, match=SLERP_MESSAGE):
+        Slerp([], r)
+
+
 def test_slerp_rot_len1():
-    msg = "must be a sequence of at least 2 rotations"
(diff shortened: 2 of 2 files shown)


scipy/spatial/transform/_rotation.pyx
```diff
@@ -15,12 +15,20 @@ np.import_array()
 
 # utilities for empty array initialization
 cdef inline double[:] _empty1(int n) noexcept:
+    if n == 0:
+        return array(shape=(1,), itemsize=sizeof(double), format=b"d")[:0]
```
nmayorov, line 19: `shape=(0,)` not going to work?

scipy/spatial/transform/_rotation.pyx
```diff
@@ -15,12 +15,20 @@ np.import_array()
...
     return array(shape=(n,), itemsize=sizeof(double), format=b"d")
 cdef inline double[:, :] _empty2(int n1, int n2) noexcept :
+    if n1 == 0:
+        return array( shape=(1, n2), itemsize=sizeof(double), format=b"d" )[:0]
```
nmayorov, line 23: Odd space.

scipy/spatial/transform/_rotation.pyx
```diff
@@ -2945,6 +2957,9 @@ cdef class Rotation:
         >>> r.mean().as_euler('zyx', degrees=True)
         array([0.24945696, 0.25054542, 0.24945696])
         """
+        if self._quat.shape[0] == 0:
+            raise RuntimeWarning("Mean of empty rotation.")
```
nmayorov, line 2961: Should it be an error?

nmayorov:
And that's it? Looks impressively short and a little bit hackish. I'll experiment with the code myself, perhaps suggest some more tests.

hubernikus:
Yes, most the computation are based on loops / array multiplication, shape=0 is not a special case.

The main limitation was that to my understanding `cython.view.array` has no way of generating a zero-dimension array.
But since helper functions are already in place it requires minimal change.

scipy/spatial/transform/_rotation.pyx
hubernikus: Using gives an error from the cython.view.array. I'm not sure if there is a more elegant way to solve that an return a placeholder for an empty array, but that's all I found.

scipy/spatial/transform/_rotation.pyx
hubernikus: Due to the empty-placeholder, this allowed to include 0-dimensional arrays with minimal change

scipy/spatial/transform/_rotation.pyx
hubernikus: I'll change that.

scottshambaugh (commented):
Agree it's rather nice that these are the only needed changes.

I'd like to see more tests fleshed out:
* Should test boolean indexing like the linked issue, with cases of all False, all True, and a mix.
* Should test `apply` for 2D vector shapes
* Should test rotation multiplication for a range of rotation shapes, on both sides of the operand (single, len 0, len N)
* Should test `from_rotvec`/`as_rotvec`, `from_davenport`/`as_davenport`, `from_mrp`/`as_mrp`
* Should test `__pow__` (with positive, negative, 1, 0, -1 powers, both int and float), `concatenate`
* What happens when used with `Slerp`?
* What happens with `identity(0)`?
* `random(0)` could use some more explicit tests.
* Should test that indexing (get/set) on a len 0 Rotation raises errors.

We will also need to do all this for `RigidTransform`. I'm ok with that being a separate PR once this one is ironed out as a template, but they should be part of the same release.

hubernikus:
I added the tests as you suggested @scottshambaugh 

>     * What happens when used with `Slerp`?

Raises an exception, as it does with Rotation.random(1). The fail-condition needed minor adaptation.

>     * What happens with `identity(0)`?

Creates a Rotation of len() = 0

>     * `random(0)` could use some more explicit tests.

I don't understand this input?


> We will also need to do all this for `RigidTransform`. I'm ok with that being a separate PR once this one is ironed out as a template, but they should be part of the same release.

Yes that would make sense, but in a separate PR.

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -2039,3 +2040,123 @@ def test_compare_as_davenport_as_euler():
...
+    v = np.array([1, 2, 3])
+    v_rotated = r.apply(v)
+    assert v_rotated.shape == (0, 3)
+
+    v0 = np.zeros((0, 3))
+    v0_rot = r.apply(v0)
+    assert v0_rot.shape == (0, 3)
+
+    v2 = np.ones((2, 3))
+    with pytest.raises(ValueError):
```
scottshambaugh, line 2096: For this and all the other ValueError checks, could you please add a string for at least a partial match on the expected error? Ensures that the error is being raised where you expect it to be.

```suggestion
    with pytest.raises(ValueError, match="operands could not be broadcast together with shapes (0,3) (2,3)"):
```

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -2039,3 +2040,123 @@ def test_compare_as_davenport_as_euler():
...
+
+    p_con = r.concatenate(Rotation.random(1))
+    assert len(p_con) == 1
+
+    p_con = r.concatenate(Rotation.random(3))
+    assert len(p_con) == 3
+
+    # Power
+    for pp in [-1.5, -1, 0, 1, 1.5]:
+        pow0 = r0**0
```
scottshambaugh, line 2130:
```suggestion
        pow0 = r0**pp
```

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -2039,3 +2040,123 @@ def test_compare_as_davenport_as_euler():
...
+
+    r0 = Rotation.random(0)
+    r0_mult = r * r0
+    assert len(r0_mult) == 0
+
+    r2 = Rotation.random(2)
+    with pytest.raises(ValueError):
+        r0 * r2
+
+    with pytest.raises(ValueError):
```
scottshambaugh, line 2115: Please also add a `r0 * r0` case

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -2039,3 +2040,123 @@ def test_compare_as_davenport_as_euler():
...
+
+    magnitude = r.magnitude()
+    assert magnitude.shape == (0,)
+
+    with pytest.raises(ValueError):
+        r.mean()
+
+    # Comparison
+    assert r.approx_equal(Rotation.random(0)).shape == (0,)
+    assert r.approx_equal(Rotation.random()).shape == (0,)
```
scottshambaugh, line 2145: Please also add `Rotation.random(0).approx_equal(r).shape`, as well as single rotation cases

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -2039,3 +2040,123 @@ def test_compare_as_davenport_as_euler():
...
+
+    # Get / set
+    r_get = r[[]]
+    assert len(r_get) == 0
+
+    with pytest.raises(IndexError):
+        r[[0]]
+
+    with pytest.raises(IndexError):
+        r[[True]]
```
scottshambaugh, line 2159: Could you add a standalone test that covers boolean indexing cases on normal non-0-len rotations? I don't think we have that explicitly tested anywhere.

scottshambaugh (commented):
Tests are a lot better, added some comments.

> `random(0)` could use some more explicit tests.

I was thinking that we should generally check the array directly rather than just the length, but `._quat` isn't actually accessible so can ignore this comment.

scipy/spatial/transform/_rotation.pyx
nmayorov: I meant why put additional spaces near the arguments in `array`? Just a minor formatting thing.

scipy/spatial/transform/_rotation.pyx
```diff
@@ -2945,6 +2957,9 @@ cdef class Rotation:
         >>> r.mean().as_euler('zyx', degrees=True)
         array([0.24945696, 0.25054542, 0.24945696])
         """
+        if self._quat.shape[0] == 0:
+            raise ValueError("Mean of empty rotation.")
```
nmayorov, line 2961: Perhaps make a more proper sentence like: "Mean of an empty rotation set is undefined" or "Can't compute mean of an empty rotation (set)", something like that.

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1635,18 +1631,23 @@ def test_slerp_rot_is_rotation():
         t = np.array([0, 1])
         Slerp(t, r)
 
+SLERP_MESSAGE = "must be a sequence of at least 2 rotations"
```
nmayorov, line 1634: Perhaps `SLERP_EXCEPTION_MESSAGE`.

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -2039,3 +2040,123 @@ def test_compare_as_davenport_as_euler():
...
+
+def test_zero_length_rotation_construction():
+    r = Rotation.random(num=0)
+    assert len(r) == 0
+
+    r_ide = Rotation.identity(num=0)
+    assert len(r_ide) == 0
+
+    r_get = Rotation.random(num=3)[[]]
+    assert len(r_get) == 0
```
nmayorov, line 2053: Add a similar test with `[False, False, False]` indexing.

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -2039,3 +2040,123 @@ def test_compare_as_davenport_as_euler():
...
+    # Concatenate
+    p_con = r.concatenate(Rotation.random(0))
+    assert len(p_con) == 0
+
+    p_con = r.concatenate(Rotation.random(1))
+    assert len(p_con) == 1
+
+    p_con = r.concatenate(Rotation.random(3))
+    assert len(p_con) == 3
+
```
nmayorov, line 2127: Also test concatenation of a zero-length to another non-zero length rotation.

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -2039,3 +2040,123 @@ def test_compare_as_davenport_as_euler():
...
+        assert len(pow0) == 0
+
+    # Methods
+    r_inv = r.inv()
+    assert len(r_inv) == 0
+
+    magnitude = r.magnitude()
+    assert magnitude.shape == (0,)
+
+    with pytest.raises(ValueError):
```
nmayorov, line 2140: Might want to check the message content.

scipy/spatial/transform/tests/test_rotation.py
nmayorov: Good point.

scipy/spatial/transform/tests/test_rotation.py
nmayorov: As @scottshambaugh this might go into a separate test for indexing.

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -2039,3 +2040,123 @@ def test_compare_as_davenport_as_euler():
             eul = rot.as_euler(seq)
             dav = rot.as_davenport(ax, order)
             assert_allclose(eul, dav, rtol=1e-12)
+
+
+def test_zero_length_rotation_construction():
```
nmayorov, line 2045: The name can be `test_zero_length_rotation_construction_and_representation`.

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -2039,3 +2040,123 @@ def test_compare_as_davenport_as_euler():
...
+    assert len(r_vec) == 0
+
+    r_dav = Rotation.from_davenport(np.eye(3), "extrinsic", np.zeros((0, 3)))
+    assert len(r_dav) == 0
+
+    r_mrp = Rotation.from_mrp(np.zeros((0, 3)))
+    assert len(r_mrp) == 0
+
+
+def test_zero_length_array():
```
nmayorov, line 2083: Maybe `test_zero_length_rotation_operations`.

scipy/spatial/transform/_rotation.pyx
nmayorov: I've checked that you can use `np.empty` to create memory buffers, but not sure how it affects performance, so better start with `array`.

nmayorov: By the way can someone explain me how the memory ownership works in this scenario? Some memory buffer is created, but then some view to it is returned. I hope it works correctly, but not super clear why.

scipy/spatial/transform/tests/test_rotation.py
hubernikus: Fixed in eda31b837ab41be604975dd3f9dd6cfb47aa3069

scipy/spatial/transform/tests/test_rotation.py
hubernikus: Added exception string matching to ValueError tetsts in 55fe2abf92682ae3be76fdb60090afbb2f20929e

scipy/spatial/transform/tests/test_rotation.py
hubernikus: Added exception string matching to ValueError tetsts in 55fe2abf92682ae3be76fdb60090afbb2f20929e

scipy/spatial/transform/tests/test_rotation.py
hubernikus: I updated the concatenation tests in 8a914434572d4d7b480f63994b0b0d6e2d76856f

scipy/spatial/transform/tests/test_rotation.py
hubernikus: I added an additional boolean-indexing test function in 6a76337fe3c74fbc86c57c628c695db3c359ff05

Let me know if it this is what you had in mind?

scipy/spatial/transform/tests/test_rotation.py
hubernikus: Renamed in c8a64228d69d13cb11315a00e7db32dbd0de686e

scipy/spatial/transform/tests/test_rotation.py
hubernikus: I added an additional boolean-indexing test function in https://github.com/scipy/scipy/commit/6a76337fe3c74fbc86c57c628c695db3c359ff05

Replied above, too.

scipy/spatial/transform/tests/test_rotation.py
hubernikus: This case is there now with
```
 r0 = Rotation.random(0)
    r_mult = r * r0
 ```

File: scipy/spatial/transform/tests/test_rotation.py (hunk 7 above)
Comment (hubernikus) [reply scipy/spatial/transform/tests/test_rotation.py:2167]: That's the same check as line 2144. As in both cases we compare two len=0 Rotations, no?

File: scipy/spatial/transform/tests/test_rotation.py (hunk 14 above)
Comment (hubernikus) [reply scipy/spatial/transform/tests/test_rotation.py]: I separate the 0-rotation tests, into more separate test function. I gathered assert statements which were similar. So all names are now different.

Let me know if you have suggestions.

File: scipy/spatial/transform/tests/test_rotation.py (hunk 7 above)
Comment (scottshambaugh) [reply scipy/spatial/transform/tests/test_rotation.py:2167]: Ah, yeah looks like you have the cases all covered

**Review by scottshambaugh — APPROVED** [review]
I think this looks good now!

**Comment by nmayorov** [comment]
Yes, it looks. I would personally not split the tests into too many functions, but it's a preference, no need to change.

Please consider my comments about slight formatting issues in `_rotation.pyx` and that'd be it.

File: scipy/spatial/transform/_rotation.pyx [hunk 16]
```diff
@@ -15,12 +15,20 @@ np.import_array()
...
+    if n == 0:
+        return array(shape=(1,), itemsize=sizeof(double), format=b"d")[:0]
     return array(shape=(n,), itemsize=sizeof(double), format=b"d")
 cdef inline double[:, :] _empty2(int n1, int n2) noexcept :
+    if n1 == 0:
+        return array( shape=(1, n2), itemsize=sizeof(double), format=b"d" )[:0]
     return array(shape=(n1, n2), itemsize=sizeof(double), format=b"d")
 cdef inline double[:, :, :] _empty3(int n1, int n2, int n3) noexcept:
+    if n1 == 0:
+        return array(shape=(1, n2, n3), itemsize=sizeof(double), format=b"d" )[:0]
```
nmayorov, line 27: Extra space after `format=b"d"`.

scipy/spatial/transform/_rotation.pyx
nmayorov: Extra spaces before `shape=(1, n2)` and after `format=b"d"`.

scipy/spatial/transform/_rotation.pyx
hubernikus: I removed the spaces in 
ed45e3b72698d8fa1cd4278851ca922cfe8826bd

hubernikus:
Thanks for the feedback.

If there is no other input on splitting the tests, then I'll leave them as they're now. 

A last change from my side, was creating a one-liner for the shape of the multiplication array in 4e67cc1b4fd95e5322ec5303da89deaaf7b9933c

scipy/spatial/transform/_rotation.pyx
```diff
@@ -15,12 +15,20 @@ np.import_array()
 
 # utilities for empty array initialization
 cdef inline double[:] _empty1(int n) noexcept:
+    if n == 0:
+        return array(shape=(1,), itemsize=sizeof(double), format=b"d")[:0]
     return array(shape=(n,), itemsize=sizeof(double), format=b"d")
 cdef inline double[:, :] _empty2(int n1, int n2) noexcept :
+    if n1 == 0:
+        return array( shape=(1, n2), itemsize=sizeof(double), format=b"d")[:0]
```
nmayorov, line 23: Here still hanging " " .

hubernikus: Fixed in 7d097440d4bf6703c357f8a34d7be862cd34d943

nmayorov:
Please fix the linter complaint: https://github.com/scipy/scipy/actions/runs/13748444348/job/38446408698?pr=22643

hubernikus:
Fixed the linter errors in 25a43eccc7b33dff6a9e15696ff18091ae31f06c

nmayorov:
@hubernikus great job, thanks!

PR #22644, opened 2025-03-07
ENH: stats.quantile: methods to support trimming/Winsorizing

mdhaber:
#### Reference issue
Toward gh-22194

TLDR: I propose that we add examples and three `method='winsor_'` options to `scipy.stats.quantiles` to facilitate trimming and Winsorizing data, which are two methods for dealing with outliers. For example, to replace data to be trimmed with NaN (e.g. for further processing with `nan_policy='omit'`:

```python3
import numpy as np
from scipy import stats
x = np.arange(1., 21.)
low, high = stats.quantile(x, [0.1, 0.9])  # use keepdims=True for n-d arrays
x[(x < low) | (x > high)] = np.nan
array([nan, nan,  3.,  4.,  5.,  6.,  7.,  8.,  9., 10., 11., 12., 13., 14., 15., 16., 17., 18., nan, nan])
```
or to remove the data:
```python3
x = np.arange(1., 21.)
low, high = stats.quantile(x, [0.1, 0.9])
x[(x > low) & (x < high)]
# array([ 3.,  4.,  5.,  6.,  7.,  8.,  9., 10., 11., 12., 13., 14., 15., 16., 17., 18.])
```
or to mask it:
```python3
from marray import numpy as mxp
print(mxp.asarray(x, mask = (x < low) | (x > high)))
[  _   _  3.  4.  5.  6.  7.  8.  9. 10. 11. 12. 13. 14. 15. 16. 17. 18.   _   _]
```
or to Winsorize it:
```python3
low, high = stats.quantile(x, [0.1, 0.9], method='winsor')  # need method='winsor'
np.clip(x, low, high)
# array([ 3.,  3.,  3.,  4.,  5.,  6.,  7.,  8.,  9., 10., 11., 12., 13., 14., 15., 16., 17., 18., 18., 18.])
```

---

gh-22194 proposes that we provide functionality in `scipy.stats` that supersedes `scipy.stats.mstats.trim` and `scipy.stats.mstats.winsorize`. An obvious solution is to add a `scipy.stats.trim` and a `scipy.stats.winsorize` with similar behavior. A few observations about that plan:

- Trimming and Winsorizing are closely related operations. 
  - Trimming *removes* a certain percentage of observations from each tail.
  - Winsorizing *replaces* a certain percentage of observations from each tail with the next most extreme observation. 
- It's not immediately obvious what a `scipy.stats.trim` should do for multidimensional, non-masked input.
  - For 1-D input, options include a) *removing* data entirely, b) replacing trimmed data with NaN, and c) masking trimmed data (but probably only for masked input).
  - For N-D input, we can't remove data from each slice because if there are NaNs or masked elements, each slice could have a different number of elements to be trimmed, so we could end up with a ragged array. We're left with options b) and c).
- There is a *lot* of variety in how outliers are to be dealt with, and there is some disagreement about whether trimming/Winsorizing are good things to do.

Perhaps the last point is why other languages don't really have standard functions for Winsorizing. For instance:
- If you search for a function for Winsorizing in R, you'll find `DescTools` [`winsorize`](https://www.rdocumentation.org/packages/DescTools/versions/0.99.58/topics/Winsorize). I tested it, and the documentation really does not seem to reflect the behavior. It might be usable, but I don't think it works as advertised.
- Matlab has nothing built-in; there are just files on the file exchange.
- Mathematica only seems to have a `WinsorizedMean` function.
- SAS has nothing built-in; see e.g. articles [here](https://blogs.sas.com/content/iml/2015/07/15/winsorize-data.html) and [here](https://blogs.sas.com/content/iml/2017/02/08/winsorization-good-bad-and-ugly.html).
- GraphPAD Prism does not seem to have features for these things, just a [description of them](https://www.graphpad.com/guides/prism/latest/statistics/stat_trimmed__windsorized_and_harmo.htm).

With all that in mind, let me suggest an alternative that adds much less maintenance burden to SciPy and allows the user to do exactly what they want: add an option to the new `quantile` function that allows users to find the tail values at which the data would be trimmed or Winsorized. From there, it is a one-liner to replace tail values with NaNs (with a boolean mask), mask them (with `MArray` or `np.ma`), or Winsorize tail values (with `clip`).

For trimming (say, replacing with NaNs), this is easy to do already.
```python3
import numpy as np
from scipy import stats
x = np.arange(1., 21.)
low, high = stats.quantile(x, [0.1, 0.9])
x[(x < low) | (x > high)] = np.nan
# array([nan, nan,  3.,  4.,  5.,  6.,  7.,  8.,  9., 10., 11., 12., 13., 14., 15., 16., 17., 18., nan, nan])
```
This *just works* with `nan_policy='omit'` or `MArray` input. It also works with N-d arrays with arbitrary `axis` as long as `keepdims=True` is passed to `quantile`. Instead of replacing the data to be trimmed with NaN, it's just as easy to remove it (1-D) only, or mask it (e.g. with `MArray` or `np.ma`).

For Winsorizing, the only problem is that none of the existing `method` options do exactly what we would want it to do. In the example above, for instance, we *want* quantile to return `3` and `18` - the "next most extreme" elements - and then we can `clip` to these values. However, these would not be consistent with any of the existing `method` options[^1]. This PR suggests the addition of one or more additional `method` options to `quantile`, e.g. `method='winsor'`. With that:

```python3
x = np.arange(1., 21.)
low, high = stats.quantile(x, [0.1, 0.9], method='winsor')
np.clip(x, low, high)
array([ 3.,  3.,  3.,  4.,  5.,  6.,  7.,  8.,  9., 10., 11., 12., 13., 14., 15., 16., 17., 18., 18., 18.])
```

The reason we might want to add more than one "Winsor" `method` is that it is not always possible to trim or Winsorize *exactly* the prescribed amount of data (i.e. when `p * n` is not integral), so we might want options for which direction to round.

In addition to adding these `method` options, I've added a tutorial to show how to use it to replicate results of the dozen or so `trim`-related functions in `stats` and `mstats` and more. I think this is the better way to go rather than adding bloated `trim` and `winsorize` functions that can never match what the user can do with `quantile` plus one or two lines of Array API.

[^1]: At first glance, this may seem counter-intuitive and wrong. To avoid making this initial post any longer, I'm not going to get into it, but if you'd like more information, LMK. I did a bit of a deep dive for this, going back to the original papers about Winsorization, and I'd be happy to rant about how essentially all descriptions of Winsorizing online are wrong or at least ambiguous, etc. [This](https://blogs.sas.com/content/iml/2017/02/08/winsorization-good-bad-and-ugly.html) supports what I'm saying here, but even that makes a bit of a mistake, I think.


steppi:
> The reason we might want to add more than one "Winsor" `method` is that it is not always possible to trim or Winsorize _exactly_ the prescribed amount of data (i.e. when `p * n` is not integral), so we might want options for which direction to round. `winsor` might not be a good name, either; it's not standard, so I'm open to suggestions.
> 
> If this sounds good, I'd be happy to write a few examples or even a tutorial in addition to adding this `method` option. I just think this is the better way to go rather than adding bloated `trim` and `winsorize` functions that can never match what the user could do with `quantile` plus one or two lines of Array API.

This sounds like a good plan to me. I'd be happy to see the examples. I'll try to think about naming too, there might be something more standard we could use for this operation.

mdhaber:
I added some additional `method`s and drafted a tutorial / transition guide. I think we can get by with fewer `method`s, and tutorial needs some work, but I think reviewing the draft will help things along.

doc/source/tutorial/stats/outliers.md
```diff
@@ -0,0 +1,521 @@
...
+
+```{code-cell} ipython3
+stats.mstats.trimmed_mean(x, limits=(2, 19), relative=False, inclusive=(True, True))
+```
+
+What if we want to remove data outside a "confidence interval" estimate. Do any of these functions - or perhaps `scipy.stats.trim` followed by `np.mean` - offer anything to facilitate that?
+
+No, not really.
+
+So at the time of writing, SciPy offers three or four ways of performing some operations related to trimming, but none are able to perform *all* operations we might reasonably expect them to. In the spirit of the Zen of Python, we suggest the new "one-- and preferably only one --obvious way to do it.": manually, in one or two lines of code, using more familiar, general-purpose functions. Whether calculating a trimmed or winsorized statistic, there will be three steps:
```
mdhaber, line 84: "three or four ways of performing some operations related to trimming" - my mistake. There are three functions for the trimmed mean alone, and many more options for trimming.

doc/source/tutorial/stats/outliers.md
```diff
@@ -0,0 +1,521 @@
...
+   :new_tab: True
+```
+
+(outliers)=
+
++++
+
+# Trimming and Winsorization Transition Guide
+
+An outlier is an observation that differs substantially from other observations in a sample. There are many references available about identifying outliers, whether to make adjustments for outliers, and if so, what adjustments to make for outliers. This guide does not make any recommendations on these topics. Instead, the focus is how to perform two of the most common procedures:
```
mdhaber, line 27:
```suggestion
An outlier is an observation that differs substantially from other observations in a sample. There are many references available about identifying outliers, whether to make adjustments for outliers, and if so, what adjustments to make. This guide does not make any recommendations on these topics. Instead, the focus is how to perform two of the most common procedures:
```

doc/source/tutorial/stats/outliers.md
```diff
@@ -0,0 +1,521 @@
...
+
+```{code-cell} ipython3
+import numpy as np
+from scipy import stats
+np.set_printoptions(linewidth=120)
+x = np.arange(20.)
+stats.trim_mean(x, proportiontocut=0.1)
+```
+
+Without carefully reading the documentation, it is not immediately obvious how the operation was performed. In this case, we would have gotten the same result whether `proportiontocut` refered to the proportion to cut from each tail or the total amount of data to remove.
```
mdhaber, line 45:
```suggestion
Without carefully reading the documentation, it is not immediately obvious how the operation was performed. In this case, we would have gotten the same result whether `proportiontocut` referred to the proportion to cut from each tail or the total amount of data to remove.
```

doc/source/tutorial/stats/outliers.md
```diff
@@ -0,0 +1,521 @@
...
+
+```{code-cell} ipython3
+# Remove 2/20 observations total before taking the `mean`, one in each tail
+np.mean(x[1:-1])
+```
+
+```{code-cell} ipython3
+# Remove 2/20 observations from *each* tail before taking the `mean`
+np.mean(x[2:-2])
+```
```
mdhaber, line 55: I'm trying to show that it's not obvious what's really going on under the hood when the operations of trimming and taking a statistic are combined. In retrospect, I'm not sure if the particular approach was effective.

doc/source/tutorial/stats/outliers.md
```diff
@@ -0,0 +1,521 @@
...
+
+```{code-cell} ipython3
+stats.tmean(x, limits=(2, 19), inclusive=(True, True))
+```
+
+```{code-cell} ipython3
+stats.mstats.trimmed_mean(x, limits=(2, 19), relative=False, inclusive=(True, True))
+```
+
+What if we want to remove data outside a "confidence interval" estimate. Do any of these functions - or perhaps `scipy.stats.trim` followed by `np.mean` - offer anything to facilitate that?
```
mdhaber, line 80: Does "outside a 'confidence interval' estimate make sense? What I am trying to refer to is outside specified percentiles of the data, but I don't want to use those terms because it is so similar to trimming a specified percentage of the data. (Apparently, the difference is so subtle, that the two are often treated interchangably when describing how to perform trimming/winsorization.)

doc/source/tutorial/stats/outliers.md
```diff
@@ -0,0 +1,521 @@
...
+p2d = [[p  , 1 - p  ],  # thresholds for zeroth row
+       [2*p, 1 - 2*p]]  # thresholds for next row
+stats.quantile(x2d, p2d, method='winsor_less', axis=-1)
+```
+
+## Missing Data
+
++++
+
+If there are values to be ignored, use an [MArray](https://mdhaber.github.io/marray/tutorial.html) to mask them out.
```
mdhaber, line 138: I should probably include a short section motivating the transition from NumPy masked arrays to MArrays here, too?

doc/source/tutorial/stats/outliers.md
```diff
@@ -0,0 +1,521 @@
...
+If there are values to be ignored, use an [MArray](https://mdhaber.github.io/marray/tutorial.html) to mask them out. 
+
+```{code-cell} ipython3
+from marray import numpy as mxp
+mask = ((x2d % 2) == 0)  # ignore odd elements
+y = mxp.asarray(x2d, mask=mask)
+print(y)
+```
+
+`quantile` and many other `scipy.stats` functions natively support `MArray`s.
```
mdhaber, line 147: I recently created a system for documenting backend support, but initially I left MArrays out of the [table](https://scipy.github.io/devdocs/dev/api-dev/array_api.html#currently-supported-functionality). Once MArray is in the table, I can explain how the user can determine whether a function support MArrays.

doc/source/tutorial/stats/outliers.md
```diff
@@ -0,0 +1,521 @@
...
+
+```{code-cell} ipython3
+stats.quantile(y, mxp.asarray(p2d), method='winsor_less', axis=-1)
+```
+
+If NaNs are used to represent missing values, the `nan_policy='omit'` option is also available
```
mdhaber, line 153: Later I motivate why NaNs are not ideal, but maybe I should bump that up.

doc/source/tutorial/stats/outliers.md
```diff
@@ -0,0 +1,521 @@
...
+
+# Trimming and Winsorizing
+
++++
+
+## Winsorizing is `clip`ing
+
++++
+
+Once a numerical threshold has been identified, winsorizing is identical to a routine so fundamental that it is part of the array API standard: `clip`.
```
mdhaber, line 169: Should probably link to the array API standard. Probably need to explain that `SCIPY_ARRAY_API=1` needs to be set to do the things in this tutorial.

doc/source/tutorial/stats/outliers.md
```diff
@@ -0,0 +1,521 @@
...
+np.nanmean(x2d_nan, axis=-1)
+```
+
+And many SciPy functions that accept a `nan_policy` argument.
+
+```{code-cell} ipython3
+stats.kurtosis(x2d_nan, nan_policy='omit', axis=-1)
+```
+
+# Recipes
```
mdhaber, line 297: Things sort of got sloppy here. I only realized what a jungle it was after starting to write. Not sure whether I should try to make this more comprehensive, remove it entirely, or something in-between. Ideally, I would be able to identify a correspondence between options like `proportiontocut`, `relative`, and `inclusive` and the new method, but I find the existing behavior somewhat confusing, so I'm not sure if it's useful to map them to the new options.

doc/source/tutorial/stats/outliers.md
```diff
@@ -0,0 +1,521 @@
...
+
+```python3
+stats.mstatats.trim(a, limits=limits, inclusive=inclusive, relative=relative).mean(axis=axis)
+```
+
+So the behavior is best understood in terms of `stats.mstats.trim`.
+
++++
+
+## `stats.mstats.trim`
```
mdhaber, line 393: Oops. Forgot to mention that `stats.mstats.trimboth` and `stats.mstats.trimtail` are two more convenience wrappers around `trimr`, and `stats.trimboth` and `stats.trim1` are additional options for 1-D data that actually remove elements rather than masking them.

scipy/stats/_quantile.py
```diff
@@ -335,3 +345,25 @@ def _quantile_hd(y, p, n, xp):
...
+               winsor_round=(xp.round, xp.round))
+    op_left, op_right = ops[method]
+    j = xp.where(p < 0.5, op_left(p*n), op_right(n*p - 1))
+    return xp.take_along_axis(y, xp.astype(j, xp.int64), axis=-1)
+
+
+def _quantile_trim(y, p, n, method, xp):
+    ops = dict(trim_less=(xp.floor, xp.ceil),
+               trim_more=(xp.ceil, xp.floor),
+               trim_round=(xp.round, xp.round))
```
mdhaber, line 362: I think we can get away without all these options. Just wanted to create them to explore a bit.

scipy/stats/_quantile.py
```diff
@@ -335,3 +345,25 @@ def _quantile_hd(y, p, n, xp):
...
+    op_left, op_right = ops[method]
+    j = xp.where(p < 0.5, op_left(p*n)-1, op_right(n*p))
+    j_clip = xp.clip(j, 0*n, n-1)
+    res = xp.take_along_axis(y, xp.astype(j_clip, xp.int64), axis=-1)
+    res[j < 0] -= 1
+    res[j >= n] += 1
```
mdhaber, line 368: Whether this is needed depends on what options we end up with and whether we standardize the use of strict vs non-strict inequalities in masking.... easier to discuss in real time.

mdhaber (commented):
Some self-review.

scipy/stats/_quantile.py
mdhaber: Another thing I considered doing to avoid adding all these methods: add a `sample_order_statistic` function. 

Good:
- The user does the rounding themselves, so we don't have to worry about their preferred rounding direction.
- Also facilitates the use-case of trimming/winsorizing a certain *number* of elements.

Bad: 
- This puts most of the work back on the user, including paying attention to the number of non-masked elements.

So at that point, they might as well just sorting the array themselves and doing indexing rather than calling the function. It's probably not worth having a `sample_order_statistic` function because it would do so little. 

---

Hopefully we can just agree on a reduced set of methods.

mdhaber: I've thought about it more carefully and I'm pretty sure we can't possibly need more than three variants. We still might want different rounding methods, but there is no need to return a different value when trimming vs winsorizing. Suppose we want to trim or winsorize `m` values on the left, where `m` is the result of rounding `p*n`.

In that case, we return the value at index `m` of the sorted array. To winsorize, the user `clip`s at that value. To trim, the user masks elements *strictly less* than that value or, equivalently, retains elements *greater than or equal to* that value.

Note that the minimum value of `m` is 0 when `p = 0` or `p*n` is small enough to be rounded down, according to the preferred rounding scheme. These are the conditions under which we'd want to trim zero data, so this makes sense. It does not make sense that my `trim_more` method could return a value less than the minimum element. That clearly shouldn't be needed.

doc/source/tutorial/stats/outliers.md
mdhaber: Not here.

scipy/stats/_quantile.py
mdhaber: Removed `trim` methods and documented `winsor` methods.

doc/source/tutorial/stats/outliers.md
mdhaber: Shall I leave the questions and remove the example code cells?

mdhaber (commented):
I think this is ready for a real review.

I believe we could get away with `winsor_less` and `winsor_round` if we just wanted to be able to reproduce the results of the old functions, given their `inclusive` behavior. If we're going to support two of these methods, though, I feel like we might as well support all three. 

I considered having the rounding direction as a separate argument, but I don't like having a separate argument that is incompatible with all methods *except* `winsor`, and I don't think there is a good way of making it mean something for the other arguments. One thought is to define a `method="winsor"` (e.g. to mean `winsor_round`) and add the rounding options as a sort of footnote - the idea is to make them available, but to avoid complicating the docs. Perhaps instead of parsing the string, we could accept the rounding direction as the second element of a tuple; e.g. `method='winsor'` and `method=('winsor', 'less')` could be valid. Heck, the user could just pass the rounding functions to use: `method=('winsor', np.round)` and `method=('winsor', np.floor, np.ceil)` could be the way to specify `winsor_round` and `winsor_less`, respectively. This would not be so out of place if someday we want to add a `method` that accepts `alpha` and `beta` (currently accepted by `mquantiles`) as a tuple. With those ideas in mind, we could just add `winsor` as the method now, and save specification of the rounding method for a later day, _if_ reqested.

scipy/stats/tests/test_quantile.py
```diff
@@ -14,22 +14,38 @@ def quantile_reference_last_axis(x, p, nan_policy, method):
     if nan_policy == 'omit':
         x = x[~np.isnan(x)]
     p_mask = np.isnan(p)
-    p = p.copy()
+    p = np.asarray(p, copy=True)
```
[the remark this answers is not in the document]
mdhaber: And I think it will address the old NumPy CI failures.

scipy/stats/tests/test_quantile.py
```diff
@@ -14,22 +14,38 @@ def quantile_reference_last_axis(x, p, nan_policy, method):
...
-        res = np.quantile(x, p)
+        res = np.quantile(x, p, method=method)
+    if nan_policy == 'propagate' and np.any(np.isnan(x)):
+        res[:] = np.nan
+    res = np.asarray(res)
     res[p_mask] = np.nan
     return res
 
 
+@np.vectorize(excluded={0, 2})
```
mdhaber, line 34:
```suggestion
@np.vectorize(excluded={0, 2})  # type: ignore[call-arg]
```

scipy/stats/tests/test_quantile.py
```diff
@@ -84,29 +100,34 @@ def test_input_validation(self, xp):
     @pytest.mark.parametrize('method',
```
mdhaber, line 100:
```suggestion
    @skip_xp_backends(cpu_only=True, reason="PyTorch doesn't have `betainc`.",
                      exceptions=['cupy'])
    @pytest.mark.parametrize('method',
```

steppi:
> Heck, the user could just pass the rounding functions to use: `method=('winsor', np.round)` and `method=('winsor', np.floor, np.ceil)` could be the way to specify `winsor_round` and `winsor_less`, respectively. This would not be so out of place if someday we want to add a `method` that accepts `alpha` and `beta` (currently accepted by `mquantiles`) as a tuple. With those ideas in mind, we could just add `winsor` as the method now, and save specification of the rounding method for a later day, _if_ reqested.

I like this idea of being able to pass the rounding method. I can even imagine cases where users might prefer an ad-hoc custom rounding method. It's a good consideration that we may end up with other methods with method specific options. 

> I considered having the rounding direction as a separate argument, but I don't like having a separate argument that is incompatible with all methods _except_ `winsor`, and I don't think there is a good way of making it mean something for the other arguments.

Are you aware of prior art for packing options specific to a particular value `v` of a keyword argument `x` into some kind of structured value for `x` rather than having additional keyword arguments which are ignored if `x != v`? I think keyword options which are applicable only for particular choices of other keyword options is pretty common in SciKit-Learn for instance, see `degree` for [SVC](https://scikit-learn.org/stable/modules/generated/sklearn.svm.SVC.html), or `l1_ratio` for [SDGClassifier](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.SGDClassifier.html#sklearn.linear_model.SGDClassifier) for instance. I think option specific options are fairly common in SciPy too.

I can see a wide proliferation of option specific options (which I don't think we have in this case) potentially being a sign that a function with different methods might be better off split into separate functions if the API deviates too much between the methods, but I don't really see any issues with a separate `round` (or whatever we want to call it) option that takes callables and only applies when `method=winsor`. Maybe it's not ideal, but it's at least familiar. Has the community been moving away from such things? I admit to not really being up on the latest thoughts on Python API design.

mdhaber:
> I can see a wide proliferation of option specific options 

I think I've seen this enough that I just avoid it, whether or not I think it's likely to happen for `quantile`.

So how about I just make this `method='winsor'`, and also accept a three-tuple with `winsor` and the desired rounding functions?

mdhaber:
Case in point - gh-22794 just came up asking for a `weights` options that might only apply to some methods.

steppi:
What I'm saying is, keyword options which are only applicable for specific values of other options seem to be very common. This happens all throughout Scikit-Learn for instance. I'm not aware of any general principles that say not to do this. Packing options into structured values is more of an unknown to me. Are there conventions around it?  If everyone were to pack option specific options into values, it seems like there's more than one an obvious way to do it in many cases. It might add cognitive burden to have to remember how specifically the options should be packed for specific functions across the ecosystem.

.circleci/config.yml
```diff
@@ -76,8 +76,10 @@ jobs:
           command: |
             pip install numpy cython pybind11 pythran ninja meson
             pip install -r requirements/doc.txt
+
```
mdhaber:
```suggestion
```

.circleci/config.yml
```diff
@@ -76,8 +76,10 @@ jobs:
...
             # asv upper bound due to issue in gh-23611, remove when that is resolved
-            pip install mpmath gmpy2 "asv<0.6.5" pooch threadpoolctl spin
+            pip install mpmath gmpy2 "asv<0.6.5" pooch threadpoolctl spin marray
+
```
mdhaber:
```suggestion
```

mdhaber:
@dschmitz89 @j-bowhay @lucascolley if I could get one more PR into 1.17.0 release, it would be this one[^1]. No special background is needed.

If you're willing, the best way to get started would be to read the motivation in the top post, then take a look at the new ["Trimming and Winsorization"](https://output.circle-artifacts.com/output/job/f57c60a0-ec9c-4f2f-b813-b35dc5e92c01/artifacts/0/html/tutorial/stats/outliers.html#trimming-and-winsorization-transition-guide) guide, which is the bulk of this PR. That will motivate the <10 lines of implementation code and minor extensions to the existing tests.

---

CI failure is just an error in the `find_peaks` docstring example when `SCIPY_ARRAY_API=1`. @ev-br in the following:
https://github.com/scipy/scipy/blob/65deafff93fcd495ff3cf60812be08829089d1d6/scipy/signal/_delegators.py#L317-L321

OK to remove `prominence` from the call to `array_namespace`, since it can be a number or sequence, not necessarily an array?
If so, I'll adjust that with `[docs only]`.

[^1]: There's too much other work there in `quantile` right now (e.g. gh-23941, gh-23949, and incoming PR about quantile confidence intervals) to let this continue to accumulate merge conflicts, especially since there are no technical concerns, just an interface one. To resolve that, I'd suggest that we include the three new `method`s `winsor_less`, `winsor_more`, and `winsor_round` as proposed. We can add an `options` keyword or something else later if deemed necessary and de-emphasize these `method`s in the docs later, if desired.

josef-pkt:
I don't see what use trimming or winsorizing has in a quantile function.
The values are NOT affected by  outliers 
If I want the median or the 20% and 80% quantiles, then it is irrelevant what values the bottom and top 10% have and whether they are outliers or not.

Trimming or winsorizing by values could be different, but "quantile" trimming for a quantile function seems redundant to me.

I did not check what all the `method` options are doing.
Maybe there are cases that estimate a parameter on the full data. Then my comment does not apply.

mdhaber:
> I did not check what all the method options are doing... Then my comment does not apply.

Yes, that. Thanks.

(Well, just the "did not check" and "does not apply" parts. The 'harell-davis' method does estimate based on all the data, but this has nothing to do with that. This is not about taking a "winsorized quantile"; it's about enabling users to winsorize a percentage of their data.)

ev-br:
> `find_peaks` ... OK to remove `prominence` from the call to `array_namespace`, since it can be a number or sequence, not necessarily an array?

Not sure it's OK. It's documented as "number or ndarray or sequence", so it can be an array, apparently.

Oh, I see what's going on in the docstring example: it uses `prominences=(None, 0.6)`, which gets converted to an object array, and everything crashes and burns from that point on.

So the easiest fix is to always return `np_compat` from `find_peaks_signature` + add a code comment saying it needs fixing when `find_peaks` is converted.

mdhaber:
@ev-br would the fix be to just check if it an array with `xpx.is_array_object` and, if so, include it in the call to `array_namespace`? If so, that is easy enough.

ev-br:
I've to admit I'm not sure about these `is_array_api_object` helpers. The one from `_lib/_array_api` does not seem to cut it:

```
In [6]: from scipy._lib._array_api import array_namespace, is_array_api_obj

In [7]: is_array_api_obj((1, 2, 3))
Out[7]: False

In [8]: is_array_api_obj([1, 2, 3])
Out[8]: False

In [9]: import numpy as np

In [10]: is_array_api_obj(np.asarray([1, 2, 3]))
Out[10]: True

In [11]: is_array_api_obj(np.asarray([1, 2, None]))
Out[11]: True
```

Elsewhere in signal we had to resort a try-except:

```
try:
   xp = array_namespace(...)
except Exception:
   # object arrays, ugh
   xp = np_compat
```

This is fairly ugly, and is prone to masking genuine errors. A proper, disciplined way is to convert `find_peaks` and its test suite, to see what the actual contract is (which I don't know off the top of my head).

The current `find_peaks_signature` is basically written based on whatever is written in the documentation, by adding everything which reads like an `array_like`. And as you discovered, it is incorrect. Therefore, I'd suggest to either do the conversion (mainly: convert the tests), or apply the simplest band-aid fix to unblock what you are actually doing, and add a TODO-style comment.

mdhaber:
I have experienced bugs in that function, too, like it also reports True for NumPy dtypes. I will use the band-aid + comment as you suggest, but I would suggest that it should be OK to use `is_array_api_obj` now and report the bug. I don't think we should let development be blocked by these things, and it's fair to expect the bug to be fixed before this is non-experimental. Also, as long as `is_array_api_obj` reports True for bona-fide arrays and False for strings and tuples, it is working enough for this use case.

lucascolley:
This is the commit which added the function and made it return `True` for `np.generic` subclasses: https://github.com/data-apis/array-api-compat/commit/fd798e15f2e39fbc71eb3a0e0bc6ae69b2aa471e

ev-br:
> This is the commit which added the function and made it return True for np.generic subclasses

IIUC the problem is whether an object array is an array-api-object (cf cell [11] in https://github.com/scipy/scipy/pull/22644#issuecomment-3571579317.)

But anyhow `signal.find_peaks` shortcomings shouldn't block `stats.quantile`, I'd think.

lucascolley:
Do object arrays subclass `ndarray`? The original commit has a comment asking whether that support should be removed.

ev-br:
> Do object arrays subclass ndarray?

I mean these: 
```
>>> np.asarray([1, 2, None])
array([1, 2, None], dtype=object)
```

lucascolley:
so yes:

```python
In [2]: isinstance(np.asarray([1, 2, None]), np.ndarray)
Out[2]: True
```

would someone be interested in making that change and seeing what breaks?

lucascolley:
if we don't want to worry about breaking other consumers, we could make the change in https://github.com/scipy/scipy/blob/main/scipy/_lib/_array_api_override.py.

doc/source/tutorial/stats/outliers.md
```diff
@@ -0,0 +1,603 @@
...
+
+An outlier is an observation that differs substantially from other observations in a sample. There are many references available about identifying outliers, whether to make adjustments for outliers, and if so, what adjustments to make. This guide does not make any recommendations on these topics. Instead, the focus is how to perform two of the most common procedures:
+
+- trimming: removing outliers from the dataset, and
+- winsorization: replacing outliers with a more central value,
+
+using tools available in SciPy, with an emphasis on transitioning from use of legacy functions to the preferred approach.
+
+## Motivation
+In the past, SciPy has offered several "convenience functions" that combine trimming with computation of a statistic. Consider, for instance, `scipy.stats.trim_mean`.
```
j-bowhay:
```suggestion
In the past, SciPy has offered several "convenience functions" that combine trimming with computation of a statistic. Consider, for instance, {class}`scipy.stats.trim_mean <scipy.stats.trim_mean>`.
```

doc/source/tutorial/stats/outliers.md
j-bowhay: > Shall I leave the questions and remove the example code cells?

Yes I think that would be my preference

doc/source/tutorial/stats/outliers.md
```diff
@@ -0,0 +1,603 @@
...
+
+One could take the mean after using one of the seven functions for trimming data. But which one(s) will do what we need?
+
+- `stats.mstats.trim`
+- `stats.mstats.trima`
+- `stats.mstats.trimr`
+- `stats.mstats.trimtail`
+- `stats.mstats.trimboth`
+- `stats.trimboth`
+- `stats.trim1`
```
j-bowhay: Same here

doc/source/tutorial/stats/outliers.md
```diff
@@ -0,0 +1,603 @@
...
+### Missing Data
+
++++
+
+When there are missing or ignorable values but omitting them would lead to a ragged array, the two historical options have been to:
+
+- cover the missing/ignorable values with the "mask" of a NumPy masked array and use `stats.mstats` functions for analysis, or
+- place NaNs in place of the missing/ignorable values and use `stats` functions with `nan_poliy='omit'` for analysis.
+
+The downsides of these approaches are documented elsewhere; here, we recommend using an [MArray](https://mdhaber.github.io/marray/tutorial.html) with `stats` functions.
```
j-bowhay: Could you link to where

doc/source/tutorial/stats/outliers.md
```diff
@@ -0,0 +1,603 @@
...
+```{code-cell} ipython3
+x2d_winsorized = np.clip(x2d, low2d, high2d)
+x2d_winsorized
+```
+
+### Trimming
+
++++
+
+For 1-D data, trimming is equivalent to another fundamental routine: indexing.
```
j-bowhay:
```suggestion
For 1-D data, trimming is equivalent to another fundamental concept: indexing.
```
nit, but I don't know if routine is the best descriptions of indexing

doc/source/tutorial/stats/outliers.md
```diff
@@ -0,0 +1,603 @@
...
+   :new_tab: True
+```
+
+(outliers)=
+
++++
+
+# Trimming and Winsorization Transition Guide
+
+An outlier is an observation that differs substantially from other observations in a sample. There are many references available about identifying outliers, whether to make adjustments for outliers, and if so, what adjustments to make. This guide does not make any recommendations on these topics. Instead, the focus is how to perform two of the most common procedures:
```
j-bowhay:
```suggestion
An outlier is an observation that differs substantially from other observations in a sample. There are many references available about identifying outliers, whether to make adjustments for outliers, and if so, what adjustments to make. This guide does not make any recommendations on these topics. Instead, the focus is on how to perform two of the most common procedures:
```

doc/source/tutorial/stats/outliers.md
```diff
@@ -0,0 +1,603 @@
...
+```
+
+### Missing Data
+
++++
+
+When there are missing or ignorable values but omitting them would lead to a ragged array, the two historical options have been to:
+
+- cover the missing/ignorable values with the "mask" of a NumPy masked array and use `stats.mstats` functions for analysis, or
+- place NaNs in place of the missing/ignorable values and use `stats` functions with `nan_poliy='omit'` for analysis.
```
j-bowhay:
```suggestion
- place NaNs in place of the missing/ignorable values and use `stats` functions with `nan_policy='omit'` for analysis.
```

doc/source/tutorial/stats/outliers.md
```diff
@@ -0,0 +1,603 @@
...
+stats.quantile(x, [p, 1-p], method='winsor_more')
+```
+
+Rounding the number of points to trim to the nearest integer:
+
+```{code-cell} ipython3
+stats.quantile(x, [p, 1-p], method='winsor_round')
+```
+
+Note that these methods are all subtlely different from estimating the 10th and 90th percentiles of the data, which might be called for in different circumstances. Using the default method:
```
j-bowhay:
```suggestion
Note that these methods are all subtly different from estimating the 10th and 90th percentiles of the data, which might be called for in different circumstances. Using the default method:
```

doc/source/tutorial/stats/outliers.md
```diff
@@ -0,0 +1,603 @@
...
+low, high = stats.quantile(x, [p, 1-p], method='winsor_less')
+mask = (x < low) | (high < x)
+np.count_nonzero(mask)
+```
+
+```{code-cell} ipython3
+mxp.mean(mxp.asarray(x, mask=mask))
+```
+
+What about `inclusive=(False, False)`? According to the documentation, the number of elements "masked" should be truncated. In our case, `p*len(x) = 1.9`, so we would trunctate to 1 on each side for a total of two. This should give us the same result as the `winsor_less` calculation just above.
```
j-bowhay:
```suggestion
What about `inclusive=(False, False)`? According to the documentation, the number of elements "masked" should be truncated. In our case, `p*len(x) = 1.9`, so we would truncate to 1 on each side for a total of two. This should give us the same result as the `winsor_less` calculation just above.
```

doc/source/tutorial/stats/outliers.md
```diff
@@ -0,0 +1,603 @@
...
+```
+
+But instead, it gives us the same result as the `winsor_round` (or a `winsor_more`) calculation. Apparently, the documentation is backwards.
+
++++
+
+As it turns out, `stats.mstats.trimmed_mean` is a convenience function equivalent to:
+
+```python3
+stats.mstatats.trim(a, limits=limits, inclusive=inclusive, relative=relative).mean(axis=axis)
```
j-bowhay:
```suggestion
stats.mstats.trim(a, limits=limits, inclusive=inclusive, relative=relative).mean(axis=axis)
```

j-bowhay (commented):
Thanks, this looks in good shape and the guide was easy to follow. A few points/questions

- At the risk of bike shedding when there isn't much time left before branching, is it confusing having the methods being called `winsor_*` when they can be used for both winsorising **and** trimming. Is there a way to convey this in the name?
- Can you add links to the functions referenced in the tutorial so it is easier to navigate
- Nit, the headings in the tutorial are a mix of sentence and title case can we stick to one

mdhaber:
Thanks, Jake! I'll make those fixes shortly. The comment about the method name makes sense. These methods are for dealing with "outliers" - maybe something to do with that root word? The word "fence" is also associated with outlier thresholds, (but not necessarily observed values *at* the threshold).

scipy/stats/_quantile.py
```diff
@@ -249,6 +259,22 @@ def quantile(x, p, *, method='linear', axis=0, nan_policy='propagate', keepdims=
     :math:`I` is the regularized, lower incomplete beta function
     (`scipy.special.betainc`).
 
+    ``method='winsor_round'`` is equivalent to indexing ``y[j]``, where::
+
+        j = int(np.round(p*n) if p < 0.5 else np.round(n*p - 1))
+
+    This is useful when winsorizing data: replacing ``p*n`` of the most extreme
```
j-bowhay, line 266: Could also mention trimming here too?

j-bowhay:
> Thanks, Jake! I'll make those fixes shortly. The comment about the method name makes sense. These methods are for dealing with "outliers" - maybe something to do with that root word? The word "fence" is also associated with outlier thresholds, but fences are not in general observed values.

That could work. ChatGPT suggested `round_nearest`, `round_inward`, and `round_outward` which I quite like but I can understand the argument for hinting at the application in the name

doc/source/tutorial/stats/outliers.md
```diff
@@ -141,23 +106,23 @@ The Harrell-Davis method:
...
-### Missing Data
+### Missing data
 
 +++
 
 When there are missing or ignorable values but omitting them would lead to a ragged array, the two historical options have been to:
 
 - cover the missing/ignorable values with the "mask" of a NumPy masked array and use `stats.mstats` functions for analysis, or
-- place NaNs in place of the missing/ignorable values and use `stats` functions with `nan_poliy='omit'` for analysis.
+- replce missing/ignorable values with NaNs and use `stats` functions with `nan_policy='omit'` for analysis.
```
j-bowhay:
```suggestion
- replace missing/ignorable values with NaNs and use `stats` functions with `nan_policy='omit'` for analysis.
```

scipy/stats/_quantile.py
```diff
@@ -140,6 +141,10 @@ def quantile(x, p, *, method='linear', axis=0, nan_policy='propagate', keepdims=
 
         'harrell-davis' is also available to compute the quantile estimate
         according to [2]_.
+
+        `round_outward`, `round_inward`, and `round_nearest` are available for use
```
j-bowhay:
```suggestion
        'round_outward', 'round_inward', and 'round_nearest' are available for use
```
For consistency with the other methods?

mdhaber: Ah good catch. Committed with skip ci.

j-bowhay (approved):
LGTM, changes are easy to follow and the motivation provided in the tutorial is convincing

mdhaber:
Thanks Jake! @lucascolley I'll add the release notes shortly.

j-bowhay:
No problem, is there anything that would be good to get in before the release?

mdhaber:
Anything else, you mean? gh-23941 would be great. I'll resolve the merge conflict.
After that, maybe gh-23751, since it fixes some old bugs, and gh-23766,
gh-24030 and gh-24031 would be great since they involve deprecations, but I think I still need to send an email to the mailing list about those.

ev-br:
Am seeing [this in a different PR](https://app.circleci.com/pipelines/github/scipy/scipy/37529/workflows/0bc7b80b-5ab8-42a9-b04e-cdf30da4c793/jobs/129287), can it be related:

```
/home/circleci/repo/doc/source/tutorial/stats/outliers.md: WARNING: Executing notebook failed: CellExecutionError [mystnb.exec] [mystnb.exec]
/home/circleci/repo/doc/source/tutorial/stats/outliers.md: WARNING: Notebook exception traceback saved in: /home/circleci/repo/doc/build/html/reports/tutorial/stats/outliers.err.log [mystnb.exec] [mystnb.exec]
```

mdhaber:
It is related. Not sure if it indicates a problem - could be related to CircleCI not merging main (although I can't explain how it picks up the `.md` at all). Can you try updating the branch? I will check to see if there is any reason for the tutorial to be flaky (e.g. unseeded random data).

PR #22647, opened 2025-03-08
MAINT: sparse: rewrite `sparse._sputils.validateaxis` to centralize axis arg processing

dschult:
The `sum/mean/min/max` methods take an`axis` parameter that needs validation. The various sparse formats and methods duplicate validation code in non-uniform ways. We'd like to centralize the checking and conversion of `axis` into the current `_sputils.validateaxis` function.  That involves:
- making the function return the axis with everything checked and converted to nonnegative integers.
- add a parameter `ndim` to allow conversion to nonnegative integers.
- unify the error messages across methods and formats.
- capture the return value in the various users of this helper function.
- clean up the axis processing from the various formats and methods.

This function `validateaxis` is private because it is accessible only via `sparse._sputils.validateaxis`. But just in case someone is using it, I gave the second argument a default `ndim=2`, and the new version performs all the previous checks plus more. I don't think any deprecation should be needed.

This also prepares the axis validation for nD processing!

~~This is built on top of #22642 (which adds/corrects testing of invalid `out` parameters), so there might be conflicts if that gets changed. I'll keep up with rebasing as needed.~~

### DIFF
--- a/scipy/sparse/_base.py
+++ b/scipy/sparse/_base.py
@@ -1,10 +1,11 @@
 """Base class for sparse matrices"""
 
+import math
 import numpy as np
 
 from ._sputils import (asmatrix, check_reshape_kwargs, check_shape,
                        get_sum_dtype, isdense, isscalarlike,
-                       matrix, validateaxis, getdtype)
+                       matrix, validateaxis, getdtype, isintlike)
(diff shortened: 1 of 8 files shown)


perimosocordiae (approved):
This is great!

PR #22648, opened 2025-03-08
MAINT: stats.quantile: fixup quantile for p < minimum plotting position

mdhaber:
#### Reference issue
-

For some `method`s, `quantile` produced incorrect results for *nonzero* `p` less than the minimum plotting position (probabilty corresponding with the minimum observation). Tests catch it when the number of points checked is bumped up slightly. The fix is simple: when `p` is less than the minimum plotting position, the output is supposed to be clipped to the minimum observation, so we just need to set `g` - the parameter that controls linear interpolation between two adjacent values - to zero.

### DIFF
--- a/scipy/stats/_quantile.py
+++ b/scipy/stats/_quantile.py
@@ -205,8 +205,9 @@ def quantile(x, p, *, method='linear', axis=0, nan_policy='propagate', keepdims=
 
     Note that indices ``j`` and ``j + 1`` are clipped to the range ``0`` to
     ``n - 1`` when the results of the formula would be outside the allowed
-    range of non-negative indices. The ``-1`` in the formulas for ``j`` and
-    ``g`` accounts for Python's 0-based indexing.
+    range of non-negative indices. When ``j`` is clipped to zero, ``g`` is
+    set to zero as well. The ``-1`` in the formulas for ``j`` and ``g``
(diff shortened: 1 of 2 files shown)


mdhaber:
@steppi sounded like you were considering reviewing this. And we discussed that another fixup for a separate PR would be to preserve integer input dtype for discontinuous methods. That should wait until gh-22695 merges to avoid merge conflicts.

mdhaber:
@tylerjereddy I think this is a patch that `quantile` needs if it is to make it in this release. It might not need another stats-focused review, though; the bug is clear and the fix is quite short.

tylerjereddy (approved):
The one CI failure is an unrelated "slow" test failure on Windows.

I'll merge this on the basis of the following:

1. In person discussion with Matt at the summit.
2. Improved matching of behavior to battle-tested `np.quantile` (I verified fail before/pass after the patch).
3. I'd have a slight preference for leaving the old test case in place *and* adding a new one, but rather than asking Matt to do that for this simple case I simply checked manually that the old case is still passing as well.
4. I checked the full testsuite on this branch locally after rebasing since it is getting a bit older--still seems fine.

We're almost at 50 backport-labeled PRs so I need to start getting some of the simpler ones in or bumped.

PR #22651, opened 2025-03-08
MAINT: ndimage.zoom: eliminate noise when `zoom=1`

### LINKED ISSUE #20999: BUG: ndimage.zoom: wrong output with zoom factor of 1
### Describe your issue.

Hey,
I am using scipy.ndimage.zoom for 3d images. Based on the input parameters, I might need to zoom or not. So, I expect the function to act as an identity function if the zoom factor is one. You can see the code and the output.
mask

### Reproducing Code Example

```python
import numpy as np
from scipy.ndimage import zoom

size = 100
mask = np.zeros((100, 100, 100))
center = np.multiply(size,(0.5,0.5,0.5))
print(center)
radius = size/8
for x in range(size):
    for y in range(size):
        for z in range(size):
            if (x - center[0]) ** 2 + (y - center[1]) ** 2 + (z - center[2]) ** 2 < radius ** 2:
                mask[x, y, z] = 1

zoomed_mask = zoom(mask, (1,1,1))

# Plot the original and zoomed masks in 3d
fig = plt.figure()
ax = fig.add_subplot(121, projection='3d')
ax.voxels(mask, edgecolor='k')
ax.set_title('Original Mask')
ax = fig.add_subplot(122, projection='3d')
ax.voxels(zoomed_mask, edgecolor='k')
ax.set_title('Zoomed Mask')
plt.savefig('mask.png')
```

### Error message

```shell
There is no error message. See the output image.
```

```shell
Build Dependencies:
  blas:
    detection method: pkgconfig
    found: true
    lib directory: /usr/local/lib
    name: openblas
      NO_LAPACK= NO_LAPACKE= NO_AFFINITY=1 USE_OPENMP= ZEN MAX_THREADS=64
    version: 0.3.27
  lapack:
    detection method: pkgconfig
    found: true
    lib directory: /usr/local/lib
    name: openblas
      NO_LAPACK= NO_LAPACKE= NO_AFFINITY=1 USE_OPENMP= ZEN MAX_THREADS=64
    version: 0.3.27
  pybind11:
    detection method: config-tool
    include directory: unknown
    name: pybind11
    version: 2.12.0
Compilers:
  c:
    commands: cc
    linker: ld.bfd
    name: gcc
    version: 10.2.1
  c++:
    commands: c++
    linker: ld.bfd
    name: gcc
    version: 10.2.1
  cython:
    commands: cython
    linker: cython
    name: cython
    version: 3.0.10
  fortran:
    commands: gfortran
    linker: ld.bfd
    name: gcc
    version: 10.2.1
  pythran:
    include directory: ../../tmp/pip-build-env-mnl4e8vy/overlay/lib/python3.10/site-packages/pythran
    version: 0.15.0
Machine Information:
  build:
    cpu: x86_64
    endian: little
    family: x86_64
    system: linux
  cross-compiled: false
  host:
    cpu: x86_64
    endian: little
    family: x86_64
    system: linux
Python Information:
  version: '3.10'
```

mdhaber:
#### Reference issue
Closes gh-20999

Eliminate noise in output of `zoom` when `zoom=1` (or a sequence of ones).

#### Additional information
Draft while checking if other arguments should affect the result.


stefanv:
Thanks for this PR, @mdhaber. We don't support the Array API in skimage because of all the Cython extensions, so would be simpler to use NumPy syntax here.

mdhaber:
I'm not sure I understand @stefanv. If you're referring to the use of `array_api_extra`, we vendor that, and that's used throughout SciPy now. `at.set` is the way to mutate where possible and copy otherwise, so I think that's what we need here.

scipy/ndimage/_interpolation.py
```diff
@@ -818,6 +819,9 @@ def zoom(input, zoom, output=None, order=3, mode='constant', cval=0.0,
     complex_output = np.iscomplexobj(input)
     output = _ni_support._get_output(output, input, shape=output_shape,
                                      complex_output=complex_output)
+    if all(z == 1 for z in zoom):  # early exit for gh-20999
+        output = xpx.at(output)[...].set(input)
+        return output
```
mdhaber, line 824:
```suggestion
    if all(z == 1 for z in zoom) and not prefilter:  # early exit for gh-20999
        # could just set order = 0?
        output = xpx.at(output)[...].set(input)
        return output
```

mdhaber (commented):
CuPy and Dask failures are surprising. Not sure what to do about those.

The `prefilter` argument should affect the logic here. If it's `False`, then I don't think we can do much to special-case this.

Assuming it's True, instead of manually setting the output to the input, another option would be to set the `order` to `0`.

@lucascolley thoughts?

mdhaber:
@lucascolley I was hoping to get your thought on https://github.com/scipy/scipy/pull/22651#pullrequestreview-2669466779, since it sounded like you were interested in https://github.com/scipy/scipy/issues/20999#issuecomment-2708372070. If not, no problem. I don't think this is terribly urgent, so I would be fine closing it.

lucascolley:
> CuPy and Dask failures are surprising. Not sure what to do about those.

I would have to check the full traceback, but I imagine the CuPy failure requires a similar PR to this one to https://docs.cupy.dev/en/latest/reference/generated/cupyx.scipy.ndimage.zoom.html.

The Dask failure is *(I think,  currently on mobile so can't check)* because `np.asarray(some_dask_array)` returns a dask array, i.e. our naive conversion to NumPy arrays before this function doesn't work. I would probably suggest just skipping Dask for now and waiting for `xpx.lazy_apply` to land—does that sound about right @crusaderky ?

lucascolley:
@stefanv did you still have an array API query here? If I'm right in my last comment about Dask stuff, I think we actually can just use NumPy syntax here.

crusaderky:
> The Dask failure is _(I think, currently on mobile so can't check)_ because `np.asarray(some_dask_array)` returns a dask array, i.e. our naive conversion to NumPy arrays before this function doesn't work. I would probably suggest just skipping Dask for now and waiting for `xpx.lazy_apply` to land—does that sound about right @crusaderky ?

Correct; I expect *all* other tests for the same function to be skipped on Dask when there is an explicit output array.

stefanv:
> Thanks for this PR, @mdhaber. We don't support the Array API in skimage because of all the Cython extensions, so would be simpler to use NumPy syntax here.

My apologies, folks; I got my wires crossed, thought this was an skimage PR.

mdhaber:
Thanks, all. Looks like the remaining failures are the ones in main.

scipy/ndimage/_interpolation.py
```diff
@@ -841,6 +842,9 @@ def zoom(input, zoom, output=None, order=3, mode='constant', cval=0.0,
     complex_output = np.iscomplexobj(input)
     output = _ni_support._get_output(output, input, shape=output_shape,
                                      complex_output=complex_output)
+    if all(z == 1 for z in zoom) and prefilter:  # early exit for gh-20999
```
j-bowhay, line 845: Sorry if this is mentioned elsewhere but it's not entirely obvious to me why the early exit only applies when `prefilter=True`?

mdhaber: I was thinking that If `prefilter=False`, the argument has been prefiltered to get spline coefficients, so `input` is no longer really an image. it needs to undergo additional processing to return an image.

```python3
from scipy import datasets, ndimage
img = datasets.face()
filtered = ndimage.spline_filter(img)
plt.imshow(filtered)  # this is what it would look like with early exit

# out = np.zeros_like(img)
# out[:] = ndimage.zoom(filtered, 1, prefilter=False)
# plt.imshow(out)  # this is what we expect
```

lucascolley: The docs say

> If setting this to False, the output will be slightly blurred if order > 1, unless the input is prefiltered

so the early exit would avoid the doumented blurring for non pre-filtered input.

That said, maybe the `zoom=1` argument trumps that and should still be a no-op, I don't know.

mdhaber: `zoom=1` does not mean no-op on the `input`; it means return an image identical to the original image. If `prefilter` is False, `input` is not the original image, so processing is needed to return the original image.

lucascolley: yes, that sounds right to me

j-bowhay: Ah I see, would you be able to a quick comment to avoid any future confusion

mdhaber:
```suggestion
    if all(z == 1 for z in zoom) and prefilter:  # early exit for gh-20999
        # zoom 1 means "return original image". If `prefilter=False`,
        # `input` is *not* the original image; processing is still needed
        # to undo the filter. So we only early exit if `prefilter`.
```
If this looks safe, please feel free to commit with skip CI and merge.

j-bowhay (approved):
Thanks!

PR #22652, opened 2025-03-08
ENH/MAINT: add support for `args` as list in `optimize.minimize`

### LINKED ISSUE #3816: wrap_function seems not to be working when wrapper_args is a one element list

dschmitz89:
#### Reference issue
Closes #3816

With this change, `args` can be supplied as list to `minimize`. I am not sure if that was meant to work before, as there was somehow a cast to tuple via `args = (args, )` but that did not work.

#### Additional information
There might be some history involved here. Simply rejecting list input and erroring out could also be a valid solution. Open for opinions here :).

### DIFF
--- a/scipy/optimize/_minimize.py
+++ b/scipy/optimize/_minimize.py
@@ -560,8 +560,11 @@ def minimize(fun, x0, args=(), method=None, jac=None, hess=None,
     if x0.dtype.kind in np.typecodes["AllInteger"]:
         x0 = np.asarray(x0, dtype=float)
 
+    if isinstance(args, list):
+        args = tuple(args)
+
     if not isinstance(args, tuple):
-        args = (args,)
+        args = (args, )
 
     if method is None:
         # Select automatically
--- a/scipy/optimize/tests/test_optimize.py
+++ b/scipy/optimize/tests/test_optimize.py
@@ -3265,6 +3265,61 @@ def test_sparse_hessian(method, sparse_type):
     assert res_dense.nhev == res_sparse.nhev
 
 
+class TestFunctionArguments:
+
+    def fun_one_arg(self, x, a):
+        return rosen(x) + a
+    
+    def fun_grad_one_arg(self, x, a):
+        return rosen_der(x)
+    
+    def fun_hess_one_arg(self, x, a):
+        return rosen_hess(x)
+    
+    def fun_two_args(self, x, a, b):
+        return rosen(x) + a + b
+
+    def fun_grad_two_args(self, x, a, b):
+        return rosen_der(x)
+    
+    def fun_hess_two_args(self, x, a, b):
+        return rosen_hess(x)
+
+    @pytest.mark.parametrize("method", ["powell", "nelder-mead", "cobyla", "cobyqa"])
+    def test_one_arg_derivative_free(self, method):
+        test_args = [-1.0]
+        x0 = [-1.0, 1.0]
+
+        res_tuple_args = optimize.minimize(self.fun_one_arg, x0=x0,
+                                           args=tuple(test_args), method=method)
+        res_list_args = optimize.minimize(self.fun_one_arg, x0=x0, args=test_args, method=method)
+        assert_allclose(res_tuple_args.x, res_list_args.x)
+        assert res_tuple_args.nfev == res_list_args.nfev
+    
+    @pytest.mark.parametrize("method", ["slsqp", "bfgs", "l-bfgs-b", "tnc", "newton-cg"])
+    def test_one_arg_with_gradient(self, method):
+        test_args = [-1.0]
+        x0 = [-1.0, 1.0]
+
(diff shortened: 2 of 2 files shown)


dschmitz89:
CC @andyfaff would you have time for a review?

j-bowhay:
I think this might break backwards compatibility with valid usage such as
`minimize(lambda x, a: a[0]*x**2 + a[1], [0], args=[2,3])`

andyfaff:
> I think this might break backwards compatibility with valid usage such as
`minimize(lambda x, a: a[0]*x**2 + a[1], [0], args=[2,3])`

I don't think that is valid usage. I'm not sure the extent to which this is present in the wild. The definition of the objective function is `fun(x, *args)`. The valid function in that example would be:

`minimize(lambda x, *a: a[0]*x**2 + a[1], [0], args=[2,3])`

andyfaff:
I note that 
```
minimize(lambda x, *a: a[0]*x**2 + a[1], [0], args=np.array([2,3]))
```
doesn't work. Because `args` is turned into `(np.array([2, 3]),)` by the code in this PR.

If we're going to go down the line of allowing sequency/array-like things by turning them into tuples, then I think this example should also work.

In the array-api spec `xp.ndarray`s have an abstract base class?

lucascolley:
> In the array-api spec xp.ndarrays have an abstract base class?

Not yet at least, but that is a question for @jorenham :)

Note that there is not a standardised name for the array class.

jorenham:
> > In the array-api spec xp.ndarrays have an abstract base class?
> 
> Not yet at least, but that is a question for @jorenham :)
> 
> Note that there is not a standardised name for the array class.

They never will have an abstract base class AFAIK. They might get a couple of protocols (the `typing` ones) that they'll be assignable to though.

jorenham:
About this PR, I can't say I'm in favor. Because from the static typing side of things (i.e. [scipy-stubs](https://github.com/scipy/scipy-stubs/)), this would mean that it's not possible to validate the correctness of the argument types anymore.

The technical reason behind that is that tuple types are *variadic*, while list types are not. So type checkers will infer the type of e.g. `(1, "")` as `tuple[int, Literal[""]]`, but `[1, ""]` will be inferred as `list[int | str]`. So with a list, we'll no longer be able to check whether the type of the `*args` of the objective function (which is an "unpacked" tuple type) matches that of `args` that's passed to `minimize`. 

Additionally, this PR would require adding additional overloads to handle this specific `list` case. And as you can see [here](https://github.com/scipy/scipy-stubs/blob/f66b1dbcc600d1c040b8a154d5a9aaab471300b3/scipy-stubs/optimize/_minimize.pyi#L185-L245), the current annotations are already rather complex. 

To summarize, this change would make the `minimize` function less type-safe, and its type definition more complicated. I'm probably biased, but I don't think that's worth it.

andyfaff:
For me that's a good a reason as any to close both the PR and the original issue.

We could do any of the following:
- leave as-is
- check that args is a tuple, and if it's not raise
- strengthen the docstring for tuple

andyfaff:
Ok. Going to close the original issue as won't fix. Thank you for the work you did here @dschmitz89, it helped resolve things.

PR #22653, opened 2025-03-08
DOC: add COBYQA to local optimizer comparison table

dschmitz89:
Adds the recently added COBYQA solver to the local optimizer comparison table in the tutorial.


j-bowhay (approved):
Thanks!

PR #22656, opened 2025-03-09
ENH: signal: New functions `sawtooth_rfft` and `square_rfft` for creating band-limited sawtooth and square signals.

DietBru:
This is a shot at resolving the oldest open issue in the signal module, namely #1507. It provides the new functions [sawtooth_rfft] and [square_rfft], which return the one-sided discrete Fourier transform of a sawtooth and a square waveform. This allows to efficiently create signals by utilizing the [irfft] function.

I chose not to integrate band-limited version into the original [sawtooth] and [square] functions, but to provide the one-sided FFTs because:
*  [sawtooth] and [square] allow arbitrary time stamps—band-limited signals are closely related to equidistant sample intervals. Hence, the existing parameters are not suited to enforce equidistant time stamps.
* Being able to use the efficient [irfft] function comes with some limitations: Only signals with complete cycles can be created. By letting the users apply the [irfft] function themselves, reasoning about those limitations becomes much easier.
* When one cares about band-limited signals, one is probably also interested in the corresponding spectrum.
* Having the explicit Fourier coefficients is beneficial when creating signals with varying frequency (as illustrated in the second example of [square_rfft]).

Additional Changes:
* Some unit tests were added to achieve 100% coverage of file `signal/_waveforms.py`.
* New unit tests for `signal.square` and `signal.sawtooth` were added to check for correct values.
* Beautified plots of in docs of  [square] and [sawtooth].

Notes:
* Links to rendered docs:  [square_rfft], [sawtooth_rfft], [square], [sawtooth]
* The error in CI (``test_hermitian``) seems to be unrelated
* I tried to incorporate the array API standard, but could come up with a good way to implement this without mutating arrays (as demanded by JAX).
* ~~Closes~~ ~~#1507~~ (due to this [comment] below).

[square_rfft]: https://output.circle-artifacts.com/output/job/78c70132-73f0-4811-89dd-4bd5bbafb99a/artifacts/0/html/reference/generated/scipy.signal.square_rfft.html
[sawtooth_rfft]: https://output.circle-artifacts.com/output/job/78c70132-73f0-4811-89dd-4bd5bbafb99a/artifacts/0/html/reference/generated/scipy.signal.sawtooth_rfft.html#scipy.signal.sawtooth_rfft
[sawtooth]: https://output.circle-artifacts.com/output/job/78c70132-73f0-4811-89dd-4bd5bbafb99a/artifacts/0/html/reference/generated/scipy.signal.sawtooth.html#scipy.signal.sawtooth
[square]: https://output.circle-artifacts.com/output/job/78c70132-73f0-4811-89dd-4bd5bbafb99a/artifacts/0/html/reference/generated/scipy.signal.square.html#scipy.signal.square
[irfft]: https://scipy.github.io/devdocs/reference/generated/scipy.fft.irfft.html
 [comment]: https://github.com/scipy/scipy/pull/22656#issuecomment-2886890141

### DIFF
--- a/scipy/signal/__init__.py
+++ b/scipy/signal/__init__.py
@@ -218,13 +218,15 @@ Waveforms
 .. autosummary::
    :toctree: generated/
 
-   chirp        -- Frequency swept cosine signal, with several freq functions.
-   gausspulse   -- Gaussian modulated sinusoid.
-   max_len_seq  -- Maximum length sequence.
-   sawtooth     -- Periodic sawtooth.
-   square       -- Square wave.
-   sweep_poly   -- Frequency swept cosine signal; freq is arbitrary polynomial.
-   unit_impulse -- Discrete unit impulse.
+   chirp         -- Frequency swept cosine signal, with several freq functions.
+   gausspulse    -- Gaussian modulated sinusoid.
+   max_len_seq   -- Maximum length sequence.
+   sawtooth      -- Periodic sawtooth.
+   sawtooth_rfft -- Onesided FFT of a sawtooth or triangle wave signal.
+   square        -- Square wave.
+   square_rfft   -- Onesided FFT of a square wave signal.
+   sweep_poly    -- Frequency swept cosine signal; freq is arbitrary polynomial.
+   unit_impulse  -- Discrete unit impulse.
 
 Window functions
 ================
--- a/scipy/signal/_waveforms.py
+++ b/scipy/signal/_waveforms.py
@@ -147,8 +147,8 @@ def sawtooth_rfft(n: int, m_cyc: int, duty: float = 0.5, *,
                0 &\text{ for } l = 0\ ,\\
                -\frac{\mathbb{j}\tau}{\pi l} &\text{ for } l \neq 0 \wedge d = 0 \ ,\\
                +\frac{\mathbb{j}\tau}{\pi l} &\text{ for } l \neq 0 \wedge d = 1 \ ,\\
-               \frac{\tau}{2\pi^2 l^2 d (d-1)}
-               \left(1 - \operatorname{e}^{\mathbb{j}2\pi d l}\right)
+               \frac{\tau \left(1 - \operatorname{e}^{\mathbb{j}2\pi d l}\right)}{
+                     2\pi^2 l^2 d (d-1)}
                &\text{ for } l \neq 0 \wedge 0 < d < 1 \ .
             \end{cases}
 
@@ -166,8 +166,8 @@ def sawtooth_rfft(n: int, m_cyc: int, duty: float = 0.5, *,
 
     Examples
     --------
-    The following cade generates a plot of 1 Hz square wave and its Magnitude
-    spectrum:
+    The following code generates a plot of 1 Hz square wave and its Magnitude spectrum
+    (Consult the `square_rfft` examples on creating a signal with varying frequency):
 
     >>> import numpy as np
     >>> from matplotlib import pyplot as plt
@@ -368,7 +368,7 @@ def square_rfft(n: int, m_cyc: int, duty: float = 0.5,  *,
             \end{cases}
 
     The close relationship between Fourier series and the discrete Fourier transform
-    allows to utilize :math:`c[l]` to obtain a closed-form llexpression for `X`. I.e.,
+    allows to utilize :math:`c[l]` to obtain a closed-form expression for `X`. I.e.,
     ``X[m_cyc*l] = s * c[l] / tau``, with ``tau = n * T / m_cyc`` being the duration of
     one cycle. The scaling factor ``s`` has a value of ``n`` for parameter
     `norm` = ``'backward'``, ``sqrt(n)`` for `norm` = ``'ortho'``, or ``1`` for
@@ -380,7 +380,7 @@ def square_rfft(n: int, m_cyc: int, duty: float = 0.5,  *,
 
     Examples
     --------
-    The following cade generates a plot of 1 Hz square wave and its Magnitude
+    The following code generates a plot of 1 Hz square wave and its Magnitude
     spectrum:
 
     >>> import numpy as np
--- a/scipy/signal/_waveforms.py
+++ b/scipy/signal/_waveforms.py
@@ -9,6 +9,8 @@ from typing import Literal
 import numpy as np
 from numpy import asarray, zeros, place, nan, mod, pi, extract, log, sqrt, \
     exp, cos, sin, polyval, polyint
+
+from scipy._lib._array_api import array_namespace
 from scipy.fft import rfft
 
 __all__ = ['sawtooth', 'sawtooth_rfft', 'square', 'square_rfft', 'gausspulse', 'chirp',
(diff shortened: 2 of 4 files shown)


DietBru:
[Mention] this link to this PR in our Discussion Forum.

[Mention]: https://discuss.scientific-python.org/t/new-functions-sawtooth-rfft-and-square-rfft-for-creating-band-limited-sawtooth-and-square-signals/1764

endolith:
`irrft` method is probably more efficient at low frequencies than [direct additive synthesis](https://gist.github.com/endolith/407991)?

> band-limited signals are closely related to equidistant sample intervals. 

Not necessarily.  The main reason I didn't try to merge anything in the past was because I wanted to support chirps/sweeps/modulation, but wasn't sure how to implement it without discontinuities at Nyquist.  I think providing a digital filter argument that attenuates the high frequency components to zero as the signal sweeps would be best?  But for fixed-frequency you might not want that?

DietBru:
I did reflect about supporting chirps/sweeps/modulation:

Concerning additive synthesis: Did you have time to look at the second example of the [square_rfft] docs, @endolith? In my understanding that is the same direct synthesis you proposed, with the difference of using complex-valued `exp` functions instead of trigonometric functions. 

The problem with the harmonics hitting the Nyquist frequency, could be circumvented by oversampling the signal by factor of 2 while using the appropriate number of Fourier coefficients. That generated signal could be then down-sampled by utilizing [resample_poly], which a incorporates FIR low-pass filter to prevent aliasing. Note that choosing the anti-aliasing filter can be bit of a rabbit hole: Depending on how the filter flank should look like, it might be better to choose a higher oversampling factor. If it makes sense to provide a generic implementation (for SciPy) depends on the (subjective) assessment how frequent that specific use case comes up. 

An alternative approach might be to generate the derivative of a square wave on a highly over-sampled signal, i.e., positive and negative impulses. Then utilize again [resample_poly], but this time incorporate into the anti-aliasing filter an integrator to obtain a square wave. Note that the big advantage of [resample_poly] is its very high efficiency in applying the anti-aliasing filter—computing-time wise it is the same as applying that filter on the down-sampled signal. An in-depth explanation can be found in the  [book] from Vetterli and Kovačević (section "Multirate Discrete-Time Signal Processing"). Again, there quite a few degrees of freedom in the design (like the tolerance of jitter, using sinc-functions or other approximations for the impulses, …), which make realizing a generic implementation no easy feat. 

The conclusion of my thoughts of on this topic:
 I could not come up with a concrete specification for generating generic square/sawtooth signals with varying frequency, which would be a worthwhile implementation for SciPy—I am open to suggestions of course. With my remarks above, I tried to show that there are different kind of significant optimizations possible, which heavily depend on the design constraints. 
Nonetheless, I think the [square_rfft]/[sawtooth_rfft] functions are useful enough to be included in SciPy, since they provide a closed-from solution on constructing band-limted signals with constant frequency. Furthermore,  the [square_rfft] docs, show to calculate a signal with varying frequency.


PS: Concerning the efficiency, I am quite certain that rfft-based method will easily win over direct synthesis when calculating  a large number of samples / Fourier coefficients. 



[square_rfft]: https://output.circle-artifacts.com/output/job/16a4e4a4-44f0-4d9a-bccf-36924d989112/artifacts/0/html/reference/generated/scipy.signal.square_rfft.html
[sawtooth_rfft]: https://output.circle-artifacts.com/output/job/16a4e4a4-44f0-4d9a-bccf-36924d989112/artifacts/0/html/reference/generated/scipy.signal.sawtooth_rfft.html#scipy.signal.sawtooth_rfft
[resample_poly]: https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.resample_poly.html
[book]: http://waveletsandsubbandcoding.org

DietBru:
The CI failure
```
FAILED scipy\stats\tests\test_continuous.py::TestMakeDistribution::test_custom_variable_support[1]
```
seems unrealted.

DietBru:
Rebased and did some Array API improvements.

Links to new rendered docs:  [square_rfft], [sawtooth_rfft], [square], [sawtooth]

[square_rfft]: https://output.circle-artifacts.com/output/job/8b2a8f0c-c7e2-4204-a28b-bdf8a8c62663/artifacts/0/html/reference/generated/scipy.signal.square_rfft.html
[sawtooth_rfft]: https://output.circle-artifacts.com/output/job/8b2a8f0c-c7e2-4204-a28b-bdf8a8c62663/artifacts/0/html/reference/generated/scipy.signal.sawtooth_rfft.html#scipy.signal.sawtooth_rfft
[sawtooth]: https://output.circle-artifacts.com/output/job/8b2a8f0c-c7e2-4204-a28b-bdf8a8c62663/artifacts/0/html/reference/generated/scipy.signal.sawtooth.html#scipy.signal.sawtooth
[square]: https://output.circle-artifacts.com/output/job/8b2a8f0c-c7e2-4204-a28b-bdf8a8c62663/artifacts/0/html/reference/generated/scipy.signal.square.html#scipy.signal.square

larsoner:
@endolith do you have time to look again? I had a quick look and it seems reasonable but I am out of my depth a bit with the arguments above :point_up: , so it would be best if you could look!

endolith:
Sorry, I've been meaning to look at this.

So the IFFT method can only generate a subset of frequencies that fit exactly into a signal length, right?  It doesn't support arbitrary frequencies, chirps, PWM, starting and ending at arbitrary points in the cycle, etc.?  I was imagining something that supported all of those, like the `scipy.signal.square` does (though not necessarily with the same interface (though having a drop-in replacement convenience function that tries to mimic the same interface, or adding a `bl_approx=True` parameter to the existing functions would be good)).

There are a bunch of different bandlimited methods that have various trade-offs.  Additive Synth, IFFT, [BLIT/DSF](https://ccrma.stanford.edu/~stilti/papers/blit.pdf), [LP-BLIT](https://www.dafx17.eca.ed.ac.uk/papers/DAFx17_paper_59.pdf), BLEP/[MinBLEP](https://www.cs.cmu.edu/~eli/papers/icmc01-hardsync.pdf)/[PolyBLEP](https://mac.kaist.ac.kr/pubs/ValimakiPeknenNam-jasa2012.pdf), [BLAMP](https://www.dafx.de/paper-archive/2016/dafxpapers/18-DAFx-16_paper_33-PN.pdf), Wavetable, [DPW](https://ieeexplore.ieee.org/abstract/document/5153306), etc.

Additive synth is the most flexible and produces the most "pure" waveforms, with zero aliasing, but is heavy to compute at low frequencies with lots of harmonics, so it's not used as much for real-time generation, but for SciPy, where accuracy is more important than computation, it might be best?  (And it would need a parameter to adjust how it rolls off near Nyquist.)
BLEP and the like are also flexible and support all those different features, and easier to compute at low frequencies, but not perfectly clean.  They would need a parameter to adjust how suppressed the aliased frequencies are required to be?

I can imagine a bl_square function that supports multiple implementations and lets you choose them and the parameters for each?  (and choose minimum vs zero phase, I guess.)  I'm not sure the best API though.

> Did you have time to look at the second example of the [square_rfft](https://output.circle-artifacts.com/output/job/16a4e4a4-44f0-4d9a-bccf-36924d989112/artifacts/0/html/reference/generated/scipy.signal.square_rfft.html) docs, @endolith? In my understanding that is the same direct synthesis you proposed, with the difference of using complex-valued exp functions instead of trigonometric functions.

Yes, that looks like it's just additive synthesis?  So it doesn't have any benefits of IFFT?  I had additive synth versions of the functions [here](https://gist.github.com/endolith/407991), but they don't support chirps yet because that would require estimating the frequency from the time steps which felt dirty.

So I guess if you have a specific application in mind for the IFFT functions, they might be useful to include, but if you're just trying to close https://github.com/scipy/scipy/issues/1507, then I would think it is too limited, and we should make something more flexible?

DietBru:
Thanks @endolith for taking the time to write up your thoughts for the literature links.

You're right to point out that this spectral approach is much more limited for applications, so I removed the "closing #1507" tag. The application I had in my mind was simulating a clock signal going through a filter—modeling that in the spectral domain is the most insightful approach, in my opinion. Also for generating signals with constant frequencies, I think the `irfft` approach is probably the most efficient and it present insights about the error one is making. 

> So the IFFT method can only generate a subset of frequencies that fit exactly into a signal length, right? 

Yes, but one can freely choose the signal length, which makes it easier to approximate a given frequency.

If we are interested in the most accurate approach for square waves with varying frequency,  I think decomposing a square into a summation of step functions would be a useful approach: It is not too hard to show that the band-limited version of step function is the [sine integral function]. A minimal implementation looks like this:
```python
from itertools import accumulate

import numpy as np
from scipy.special import sici


def square_bandlimited(t, widths):
    """Generate a square wave at times `t` from a sequential list of half-wave
    `widths`, which is band-limited to a frequency of 0.5 Hz (given `t` is in s). """
    sgn = 1  # alternating sign
    y = sici(np.pi * t)[0]
    for t_c in accumulate(widths):
        sgn = -sgn
        y += sgn * sici(np.pi * (t - t_c))[0]
    return y / np.pi


ww = np.arange(1, 7) * 10  # array with sequential half-wave widths
t_pwm = np.linspace(-ww[0]/2, np.sum(ww) + ww[-1]/2, 1000)  # time stamps

x = square_bandlimited(t_pwm, ww)  # PWM square wave
t_p  = np.cumsum(np.hstack(([0,], ww)))  # flank positions in time

_, ax0 = plt.subplots()
ax0.plot(t_p, np.zeros_like(t_p), 'C0o-', label="Flank positions")
ax0.plot(t_pwm, x, 'C1-', label="Square wave")
ax0.grid(True)
ax0.legend()
plt.show()
```
It produces the following plot:
image

Advantages are:
* It produces exact results (up to numerical errors)
* It allows arbitrary instead of only equidistant timestamps

Disadvantages are:
* It's relatively slow and finding good approximation for speeding things up is not straightforward.
* The interface probably could use some more thought.
* I have not though about sawtooth waves.

I don't know when I have the bandwidth to work on a PR for this. So I am very supportive, if anybody else want to take a shot of implementing something like this. I would be happy to mentor as  far I can.


[sine integral function]: https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.sici.html

PR #22668, opened 2025-03-11
DOC: document Array API support for the constants module and optimize.rosen and associated functions

j-bowhay:
Adds Array API support tables to docs

### DIFF
--- a/scipy/optimize/_optimize.py
+++ b/scipy/optimize/_optimize.py
@@ -41,7 +41,7 @@ from scipy._lib._util import getfullargspec_no_self as _getfullargspec
 from scipy._lib._util import (MapWrapper, check_random_state, _RichResult,
                               _call_callback_maybe_halt, _transition_to_rng)
 from scipy.optimize._differentiable_functions import ScalarFunction, FD_METHODS
-from scipy._lib._array_api import array_namespace
+from scipy._lib._array_api import array_namespace, xp_capabilities
 from scipy._lib import array_api_extra as xpx
 
 
@@ -338,6 +338,7 @@ def _check_clip_x(x, bounds):
     return x
 
 
+@xp_capabilities()
 def rosen(x):
     """
     The Rosenbrock function.
@@ -389,6 +390,7 @@ def rosen(x):
     return r
 
 
+@xp_capabilities(skip_backends=[('jax', "JAX doesn't allow item assignment.")])
 def rosen_der(x):
     """
     The derivative (i.e. gradient) of the Rosenbrock function.
@@ -431,6 +433,7 @@ def rosen_der(x):
     return der
 
 
+@xp_capabilities(skip_backends=[('jax', "JAX doesn't allow item assignment.")])
 def rosen_hess(x):
     """
     The Hessian matrix of the Rosenbrock function.
@@ -474,6 +477,7 @@ def rosen_hess(x):
     return H + xpx.create_diagonal(diagonal, xp=xp)
 
 
+@xp_capabilities(skip_backends=[('jax', "JAX doesn't allow item assignment.")])
 def rosen_hess_prod(x, p):
     """
     Product of the Hessian matrix of the Rosenbrock function with a vector.
--- a/scipy/optimize/tests/test_optimize.py
+++ b/scipy/optimize/tests/test_optimize.py
@@ -35,6 +35,7 @@ from scipy.optimize import rosen, rosen_der, rosen_hess
 from scipy.sparse import (coo_matrix, csc_matrix, csr_matrix, coo_array,
                           csr_array, csc_array)
 from scipy._lib._array_api_no_0d import xp_assert_equal
+from scipy._lib._array_api import make_skip_xp_backends
 from scipy._lib._util import MapWrapper
 
 skip_xp_backends = pytest.mark.skip_xp_backends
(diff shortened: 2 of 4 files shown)


mdhaber:
Should we decorate tests even if they run with all backends? I did in the original PR, but I can see mild arguments either way.

j-bowhay:
> Should we decorate tests even if they run with all backends? I did in the original PR, but I can see mild arguments either way.

I think I agree under the principle that it ensures the documented compatibility is what is tested, even after future changes.

mdhaber:
Yeah, and now that I think about it, it will also help us tell at a glance what has already been treated with this system and what remains to be converted.

scipy/optimize/_optimize.py
```diff
@@ -474,6 +477,7 @@ def rosen_hess(x):
     return H + xpx.create_diagonal(diagonal, xp=xp)
 
 
+@xp_capabilities(skip_backends=[('jax', "JAX doesn't allow item assignment.")])
```
mdhaber, line 480: Since `rosen_hess` and `rosen_hess_prod` share tests, and if we expect that they will be subject to the same backend limitations, should the two functions share an entry in the capabilities table? For example,
```suggestion
@xp_capabilities('rosen_hess',
                 skip_backends=[('jax', "JAX doesn't allow item assignment.")])
```

scipy/optimize/tests/test_optimize.py
```diff
@@ -2455,21 +2456,20 @@ def test_powell_output():
...
+    @make_skip_xp_backends("rosen_der")
     def test_rosen_der(self, xp):
         x = xp.asarray([1, 1, 1, 1])
         xp_assert_equal(optimize.rosen_der(x),
                         xp.zeros_like(x, dtype=xp.asarray(1.).dtype))
 
-    @skip_xp_backends('jax.numpy',
-                      reason="JAX arrays do not support item assignment")
+    @make_skip_xp_backends("rosen_hess_prod")
     def test_hess_prod(self, xp):
```
mdhaber, line 2473: The change here associated with the other suggestion might be:
```suggestion
    @make_skip_xp_backends("rosen_hess")
    def test_hess(self, xp):
```

mdhaber (commented):
Could merge as-is, but I think it's worth discussing best practices. Although `rosen_hess` and `rosen_hess_prod` probably should have separate tests to check their accuracy - and not *just* that they agree - since they currently *do* share a test and backend limitations, should they share an entry in the table?

j-bowhay:
> Could merge as-is, but I think it's worth discussing best practices. Although `rosen_hess` and `rosen_hess_prod` probably should have separate tests to check their accuracy - and not _just_ that they agree - since they currently _do_ share a test and backend limitations, should they share an entry in the table?

My personal opinion is that each function should have its own entry (ideally key being either `"module.function"` or the function handle itself), for the following reason:

- Once Array API support is not experimental we may want to expose the compatibility data to users
- It would allow for per module compatibility tables or lists
- Sharing keys may lead to unintended changes; in comparison having individual entries doesn't seem like that much more effort

Of course I am willing to be convinced of the merits of the current approach and concede that these arguments are mostly based on hypothetical future use cases.

j-bowhay:
> Perhaps all the functions need to be identified to the test decorator factory, and the test is run on backends common to all functions?

I think this is a reasonable approach to take.

> The potential downside is that if, for example, `rosen_hess` and `rosen_hess_product` are _only_ tested together, and we only test on the backends available for both functions, and if one of those functions needs to skip more backends than the other, then what we advertise as supported and what we test will get out of sync. (The solution there, of course, may be to ensure that all functions/classes are adequately tested independently of others - and perhaps supplemented with property-based tests that check the relationships between the functions - but sometimes that's going to require adding tests.)

I would be happy to add independent testing of `rosen_hess` and `rosen_hess_product`.

> All those reasons sound fine, but you have to answer the question of what to do when many functions share the same test class

```
@make_skip_xp_backends("foo")
class TestFoo:
     def test_foo(self): ...

    @make_skip_xp_backends('bar')
    def test_foo_against_bar(self): ...
```

> or function.

```
@make_skip_xp_backends((''foo", "bar"))
def test_foo_bar_property(): ...
```
WDYT?

mdhaber:
Yup, I think that's good. Alhough instead of passing strings like `("foo", "bar")`, wouldn't you need to pass the function references `(foo, bar)` or fully-qualified strings `('scipy.stats.foo', 'scipy.stats.bar')`? And since there are usually not going to be any other arguments, I think it's fine to use `*args` rather than passing a tuple.

LMK what (if any) of that you want to do here.

j-bowhay:
> Yup, I think that's good. Alhough instead of passing strings like `("foo", "bar")`, wouldn't you need to pass the function references `(foo, bar)` or fully-qualified strings `('scipy.stats.foo', 'scipy.stats.bar')`? And since there are usually not going to be any other arguments, I think it's fine to use `*args` rather than passing a tuple.
> 
> LMK what (if any) of that you want to do here.

All good points. I am still on fence regarding function references vs fully-qualified strings, do you have any thoughts here? If we do choose fully-qualified strings, I guess we could drop the `scipy` since everything is in SciPy!

mdhaber:
Unless there is a technical reason not to, I think references for concision:
- you don't need quotes
- unless the test imports the module as `scipy.module` without renaming it (rare, hopefully), there's automatically no need to provide the fully-qualified name
- if the test imports the module with a more concise name or the function itself, it gets even shorter

j-bowhay:
> And you suggested references in the first place. Why are you on the fence?

Just that fully-qualified strings would be less likely to have any hidden technical gotchas. How about: in the this PR add separate tests for `rosen_hess` and `rosen_hess_product` then as a follow up switch to using function handles as the key?

mdhaber:
Actually, I'd prefer to just merge this. The fact that `rosen_hess` and `rosen_hess_product` are not adequately tested is worthy of a separate issue/PR. The fact that they share a test is something we want to be able to deal with, so how about we leave it as is here and I just merge?

lucascolley (approved):
agreed—thanks both!

PR #22671, opened 2025-03-12
MAINT: enforce modularity with `tach`

lucascolley:
Related to discussion in gh-22382, the bulk of the work here in setting up the initial `tach` config was done by @rgommers there.

This PR follows from a conversation on the Scientific Python Discord @mdhaber. Ideally we can enforce this in CI to stop new dependencies between submodules from being created accidentally.

Here we remove the dependency of `_lib` on `linalg`.

### DIFF
--- a/scipy/stats/_distn_infrastructure.py
+++ b/scipy/stats/_distn_infrastructure.py
@@ -26,7 +26,7 @@ from scipy import optimize
 from scipy import integrate
 
 # to approximate the pdf of a continuous distribution given its cdf
-from scipy._lib._finite_differences import _derivative
+from scipy.stats._finite_differences import _derivative
 
 # for scipy.stats.entropy. Attempts to import just that function or file
 # have cause import problems
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -39,7 +39,7 @@ jobs:
 
     - name: Install Python packages
       run: |
-        python -m pip install ruff cython-lint packaging
+        python -m pip install ruff cython-lint packaging tach
 
     - name: Lint
       run: |
@@ -52,3 +52,7 @@ jobs:
       shell: bash
       run: |
         python tools/check_python_h_first.py 
+
+    - name: Check module interdependencies
+      run: |
+        tach check --exact
--- a/scipy/_lib/meson.build
+++ b/scipy/_lib/meson.build
@@ -120,7 +120,6 @@ python_sources = [
   '_disjoint_set.py',
   '_docscrape.py',
   '_elementwise_iterative_method.py',
-  '_finite_differences.py',
   '_gcutils.py',
   '_pep440.py',
   '_sparse.py',
--- a/scipy/stats/_finite_differences.py
+++ b/scipy/stats/_finite_differences.py
@@ -1,4 +1,5 @@
 from numpy import arange, newaxis, hstack, prod, array
+from scipy import linalg
 
 
 def _central_diff_weights(Np, ndiv=1):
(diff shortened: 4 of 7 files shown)


lucascolley:
Example of nice discovery of dependencies:

```
scipy on 🎋 tach [🔨] is 📦 v1.16.0.dev0 via 🐍 
❯ git diff
diff --git a/tach.toml b/tach.toml
index 0fdbb5fc0..a322bbbf2 100644
--- a/tach.toml
+++ b/tach.toml
@@ -163,7 +163,6 @@ path = "scipy.stats"
 depends_on = [
     "scipy.linalg.blas",
     "scipy.stats.distributions",
-    "scipy.fft",
     "scipy.spatial.distance",
     "scipy.stats.contingency",
     "scipy.special",

scipy on 🎋 tach [🔨] is 📦 v1.16.0.dev0 via 🐍 
❯ pixi run tach
✨ Pixi task (tach in default): tach check --exact
Internal Dependencies
❌ scipy/stats/_hypotests.py[L13]: Cannot use 'scipy.fft.ifft'. Module 'scipy.stats' cannot depend on 'scipy.fft'.
❌ scipy/stats/_qmvnt.py[L37]: Cannot use 'scipy.fft.fft'. Module 'scipy.stats' cannot depend on 'scipy.fft'.
❌ scipy/stats/_qmvnt.py[L37]: Cannot use 'scipy.fft.ifft'. Module 'scipy.stats' cannot depend on 'scipy.fft'.

If you intended to add a new dependency, run 'tach sync' to update your module configuration.
Otherwise, remove any disallowed imports and consider refactoring.
```

rgommers:
Thanks for moving this forward Lucas! I think it needs a CI check. Maybe easiest is to fold it into the `lint` task? Not sure, but if it's fast enough then I think that is a reasonable thing to do and no one notices anything until the check fails.

scipy/stats/_distn_infrastructure.py
```diff
@@ -26,7 +26,7 @@
 from scipy import integrate
 
 # to approximate the pdf of a continuous distribution given its cdf
-from scipy._lib._finite_differences import _derivative
+from scipy.linalg._finite_differences import _derivative
```
ev-br, line 29: Why does it move to linalg? AFAICS it's scalar finite differences formulas, have nothing to do with linalg.

lucascolley: It uses `linalg` and is used in `stats`. Can put it in `stats` instead?

lucascolley: Alternatively, from a modularity perspective we can put it in any of the modules which `stats` depend on and which already depend on `linalg` at the minute. So

```
"scipy.stats",
"scipy.special",
"scipy.spatial",
"scipy.linalg",
"scipy.integrate",
"scipy.optimize",
"scipy.interpolate",
```

Let me know if any of those seem better!

mdhaber:
> I think it needs a CI check.

The conversation Lucas mentioned was about something like scikit-image/scikit-image#7749, which uses dependency data (with sub-package granularity) to speed up tests. As part of that, we would also check that new dependencies are not added, and I mentioned that we could easily eliminate some very weak dependencies. @steppi was going to work on a similar PR for SciPy, but if he hasn't started, perhaps that's no longer needed with Stefan and Lucas looking into similar things?

Perhaps `tach` can be used instead of my simple (but very sloppy!) script!

lucascolley:
It sounds like the test speed-ups would be a welcome follow-up to this PR still @steppi !

lucascolley:
Good to go I think! We could also expose this via `dev.py lint`, but I wouldn't personally find that useful.

j-bowhay (approved):
Seems reasonable to give this a go, can always revert if it becomes a pain!

Issue #20544, opened 2024-04-20
ENH: stats: add array API-support


mdhaber:
This issue tracks progress toward the addition of array-API support to `scipy.stats` functions. The functions listed below look ready for conversion, and I'd be happy to review PRs for them. Priority, balancing the ease and importance of the task, is roughly in the order listed. 

- [x] `moment` (gh-20292)
- [x] `skew` (gh-20541 - please see this PR as an example roughly similar `kurtosis` through `circstd`)
- [ ] `kurtosis`
- [ ] `describe`
- [ ] `mode`
- [ ] `variation`
- [ ] `sem`
- [ ] `circmean`
- [ ] `circvar`
- [ ] `circstd`
- [x] `pearsonr` (gh-20284)
- [ ] `ttest_1samp` (I might do this one to help pave a path toward the rest)
- [ ] `ttest_rel`
- [ ] `ttest_ind`
- [ ] `skewtest`
- [ ] `kurtosistest`
- [ ] `normaltest`

Most other functions are either more challenging (e.g. because they do not already natively support an `axis` argument) or lower priority. Let's pick the low-hanging fruit (listed above) first. If you're particularly interested in another function, I'd appreciate it if you'd comment below before starting work.

lucascolley:
> add N-D support to _array_api.cov; consider making it public if array API won't offer it

Public _in SciPy_ seems wrong. Upstreamed into either the standard or a new library for these type of standard-extensions seems right.

lucascolley:
Thinking a little more about `mean`, `var`, `std` etc., (when you get round to it) I think it would be useful to open an issue on the array API repo discussing the fully-featured versions as you would potentially implement them within SciPy. The question is whether or not those versions would belong on the array library level, and hence (eventually) in the standard. If the answer is:
- "yes, we just need time to get the API right / see adoption", then keeping things private in SciPy is definitely the way to go to avoid duplication in the future.
- "_maybe_ (grey area / bordering on array-consumer level)", then I would also suggest keeping things private in SciPy for now, as it may be that they would find a more appropriate home in a hypothetical (say) `array-api-extra` package (potentially making their way into the standard as the implementation matures / the API is adopted).
- "no, these features belong on the array-consumer level", then I would go ahead with adding them to `scipy.stats`, as that seems like the ideal place!

Issue #9873, opened 2019-02-25
[feature request] Majority voting


skjerns:
`ndimage.filters` has quite useful filters, but in my opinion one useful filter is missing:
An occurrence based filter, such as a majority filter.

The filter would go over an array, count occurrences within the given window size and take the n-ths most occuring value as an output, e.g.:

```Python
arr = [1,1,2,3,4,4,5,6,6,6,7,8,8,8]
res = occurence_filter(arr, size=5, rank=-1,  mode='valid')
# pseudocode
for i in range(len(arr)-size):
    w = arr[i:i+size]
    occurences = count_unique_values(w) # e.g [2:1,3:1,4:1,1:2]
    ranked = occurences[rank] # =1
    res[i] = ranked
    
# res = [1,4,4,4,4,4,6,6,6,6,8]
```

would it make sense to implement such a filter?
Or is there any other way I can express such an operation that I haven't seen yet?

rgommers:
Sounds reasonable, and there's at least some demand for it: https://stackoverflow.com/questions/39829716/majority-filter-python

May be a bit specialized though and fit better in scikit-image. @stefanv @jni any opinion on this functionality?

jni:
Seems like a job for `ndi.generic_filter` and LowLevelCallables! =) We would certainly welcome this in scikit-image filters, but given that it is straightforward (cough) to implement with `ndimage` primitives, possibly it could sit there as a model for how to use these features.

@skjerns for references on how to use these:

https://ilovesymposia.com/2017/03/12/scipys-new-lowlevelcallable-is-a-game-changer/
https://ilovesymposia.com/2017/03/15/prettier-lowlevelcallables-with-numba-jit-and-decorators/

Those both use Numba, which neither SciPy nor scikit-image have as a dependency yet. This notebook from @kne42 has a reference about how to do this with Cython:

https://gist.github.com/kne42/b487af595332db468654a5c5b1ce416c

And of course see also the SciPy documentation:
https://docs.scipy.org/doc/scipy/reference/tutorial/ndimage.html#ndimage-ccallbacks

stefanv:
This sounds like it may be implemented effectively with `skimage.filters.rank`.  The generic-filter version is going to be pretty slow without some kind of acceleration (such as the ones mentioned above).

jni:
@stefanv what do you mean exactly? I'm probably missing a step in my brain, but I can't map a rank filter to "bincount" which is what needs to happen here?

[...]

PR #22676, opened 2025-03-12
ENH: stats.mode: vectorize implementation

mdhaber:
#### Reference issue
Toward gh-14651, gh-20544 
Together with the new `vectorized_filter` (gh-22575), this makes gh-9873 much easier.

`scipy.stats.mode` natively works only with 1-D slices; the `_axis_nan_policy` decorator loops over slices to add `axis` support. This PR replaces the implementation with one that is natively vectorized.

#### Additional information
The upside is ~90x speed improvement in cases where we have many short slices; the downside is 2x slowdown when we have a few long slices.

```python3
import numpy as np
from scipy import stats

rng = np.random.default_rng(23458924958234563)

# times are:
# main
# PR

x = rng.integers(0, 10, size=1_000_000)
# %timeit stats.mode(x)
# 5.53 ms ± 208 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
# 11.6 ms ± 137 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

y = x.reshape(100, 10000)
# %timeit stats.mode(y, axis=-1)
# 8.61 ms ± 24 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
# 16 ms ± 1.54 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)

y = x.reshape(10000, 100)
# %timeit stats.mode(y, axis=-1)
# 333 ms ± 3.34 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# 18.6 ms ± 164 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

y = x.reshape(100000, 10)
# %timeit stats.mode(y, axis=-1)
# 3.33 s ± 38.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# 36.8 ms ± 320 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

It would be simple to use the old implementation when the input is 1-D.
As an enhancement, we could also add a heuristic to loop when looping is expected to be faster.

### DIFF
--- a/scipy/stats/_stats_py.py
+++ b/scipy/stats/_stats_py.py
@@ -503,8 +503,7 @@ def _mode_result(mode, count):
     return ModeResult(mode, count)
 
 
-@_axis_nan_policy_factory(_mode_result, override={'vectorization': True,
-                                                  'nan_propagation': False})
+@_axis_nan_policy_factory(_mode_result, override={'nan_propagation': False})
 def mode(a, axis=0, nan_policy='propagate', keepdims=False):
     r"""Return an array of the modal (most common) value in the passed array.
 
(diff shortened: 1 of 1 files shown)


j-bowhay:
Given it is only two lines, it does not seem too bad to keep a separate 1D path

mdhaber:
Done! Confirmed locally that performance for 1-D arrays is preserved.

j-bowhay (approved):

PR #22677, opened 2025-03-13
MAINT: use function handles rather than custom strings in `xp_capabilities_table`

j-bowhay:
Follow up to #22609

Originally in #22609 the keys in `xp_capabilities_table` are a custom string of the developers choosing. Subsequently we have discussed a few reasons why we might want to standardise this:

> My personal opinion is that each function should have its own entry (ideally key being either `"module.function"` or the function handle itself), for the following reason:
>- Once Array API support is not experimental we may want to expose the compatibility data to users
>- It would allow for per module compatibility tables or lists
>- Sharing keys may lead to unintended changes; in comparison having individual entries doesn't seem like that much more effort

https://github.com/scipy/scipy/pull/22668#issuecomment-2715427331

This PR changes the key from being a custom string to a reference to the function. This should also help prevent the keys from becoming a jumble of different naming conventions and prevent collisions.

### DIFF
--- a/scipy/_lib/_array_api.py
+++ b/scipy/_lib/_array_api.py
@@ -734,8 +734,7 @@ def xp_capabilities(capabilities_table=None, **kwargs):
         def wrapper(*args, **kwargs):
             return f(*args, **kwargs)
 
-        if wrapper not in capabilities_table:
-            capabilities_table[wrapper] = kwargs
+        capabilities_table[wrapper] = kwargs
         capabilities = _make_capabilities(**capabilities_table[wrapper])
 
         note = _make_capabilities_note(f.__name__, capabilities)
@@ -755,21 +754,21 @@ def make_skip_xp_backends(*funs, capabilities_table=None):
 
     skip_backends = []
     cpu_only = False
-    cpu_only_reason = ""
+    cpu_only_reason = set()
     np_only = False
     exceptions = []
 
     for fun in funs:
         skip_backends += capabilities_table[fun].get('skip_backends', [])
         cpu_only |= capabilities_table[fun].get('cpu_only', False)
-        cpu_only_reason += capabilities_table[fun].get('reason', "")
+        cpu_only_reason.add(capabilities_table[fun].get('reason', ""))
         np_only |= capabilities_table[fun].get('np_only', False)
         exceptions += capabilities_table[fun].get('exceptions', [])
 
     decorators = []
     if cpu_only:
         kwargs = dict(cpu_only=True, exceptions=exceptions)
-        kwargs |= {'reason': cpu_only_reason}
+        kwargs |= {'reason': "\n".join(cpu_only_reason)}
         decorators.append(pytest.mark.skip_xp_backends(**kwargs))
 
     if np_only:
--- a/scipy/_lib/_array_api.py
+++ b/scipy/_lib/_array_api.py
@@ -717,7 +717,8 @@ def make_skip_xp_backends(*funs, capabilities_table=None):
     for fun in funs:
         skip_backends += capabilities_table[fun].get('skip_backends', [])
         cpu_only |= capabilities_table[fun].get('cpu_only', False)
-        cpu_only_reason.add(capabilities_table[fun].get('reason', ""))
+        # Empty reason causes the decorator to have no effect
+        cpu_only_reason.add(capabilities_table[fun].get('reason', "No reason given."))
         np_only |= capabilities_table[fun].get('np_only', False)
         exceptions += capabilities_table[fun].get('exceptions', [])
 
--- a/scipy/_lib/_array_api.py
+++ b/scipy/_lib/_array_api.py
@@ -693,7 +693,7 @@ def _make_capabilities_table(capabilities):
             val += " " * (len('{numpy.}') + len(attr) - 1)
             setattr(backend, attr, val)
 
-    table = f"""                
+    table = f"""
     +---------+-------------+-------------+
     | Library | CPU         | GPU         |
     +=========+=============+=============+
@@ -725,22 +725,20 @@ def _make_capabilities_note(fun_name, capabilities):
     return textwrap.dedent(note)
 
 
-def xp_capabilities(name=None, capabilities_table=None, **kwargs):
+def xp_capabilities(capabilities_table=None, **kwargs):
     capabilities_table = (xp_capabilities_table if capabilities_table is None
                           else capabilities_table)
 
     def decorator(f):
-        fun_name = f.__name__
-        table_entry = name or fun_name
-        if table_entry not in capabilities_table:
-            capabilities_table[table_entry] = kwargs
-        capabilities = _make_capabilities(**capabilities_table[table_entry])
-
         @functools.wraps(f)
         def wrapper(*args, **kwargs):
             return f(*args, **kwargs)
 
-        note = _make_capabilities_note(fun_name, capabilities)
+        if wrapper not in capabilities_table:
+            capabilities_table[wrapper] = kwargs
+        capabilities = _make_capabilities(**capabilities_table[wrapper])
+
+        note = _make_capabilities_note(f.__name__, capabilities)
         doc = FunctionDoc(wrapper)
         doc['Notes'].append(note)
         wrapper.__doc__ = str(doc).split("\n", 1)[1]  # remove signature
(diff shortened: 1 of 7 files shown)


scipy/stats/tests/test_stats.py
```diff
@@ -383,10 +388,9 @@ def test_pROUNDROUND(self):
         assert_approx_equal(r,1.0)
 
 
-@make_skip_xp_backends('pearsonr')
+@make_skip_xp_backends(stats.pearsonr)
 class TestPearsonr:
-    @skip_xp_backends(np_only=True)
```
j-bowhay, line 388: I got carried away here cleaning up, I can revert if desired!

scipy/stats/tests/test_stats.py
```diff
@@ -3470,15 +3473,6 @@ class TestMoments:
     np.random.seed(1234)
     testcase_moment_accuracy = np.random.rand(42)
 
-    def _assert_equal(self, actual, expect, *, shape=None, dtype=None):
```
j-bowhay, line 3473: unused

scipy/_lib/_array_api.py
```diff
@@ -725,22 +725,20 @@ def _make_capabilities_note(fun_name, capabilities):
...
-        if table_entry not in capabilities_table:
-            capabilities_table[table_entry] = kwargs
-        capabilities = _make_capabilities(**capabilities_table[table_entry])
-
         @functools.wraps(f)
         def wrapper(*args, **kwargs):
             return f(*args, **kwargs)
 
-        note = _make_capabilities_note(fun_name, capabilities)
+        if wrapper not in capabilities_table:
```
j-bowhay, line 737: I guess this check _should_ no longer be needed

j-bowhay:
Oops failures look real, penance for only having array api strict installed

scipy/_lib/_array_api.py
```diff
@@ -749,23 +747,29 @@ def wrapper(*args, **kwargs):
...
-    exceptions = capabilities_table[fun_name].get('exceptions', None)
-    reason = capabilities_table[fun_name].get('reason', None)
+    skip_backends = []
+    cpu_only = False
+    cpu_only_reason = ""
+    np_only = False
+    exceptions = []
+
+    for fun in funs:
+        skip_backends += capabilities_table[fun].get('skip_backends', [])
```
j-bowhay, line 763: I should probably make sure to avoid duplicates here

scipy/stats/tests/test_stats.py
```diff
@@ -129,11 +129,11 @@ def test_tmean(self, xp):
         y_true = [4.5, 10, 17, 21, xp.nan, xp.nan, xp.nan, xp.nan, xp.nan]
         xp_assert_close(y, xp.asarray(y_true))
 
+    @make_skip_xp_backends(stats.tvar)
     @pytest.mark.filterwarnings(
         "ignore:invalid value encountered in divide:RuntimeWarning:dask"
     )
-    @skip_xp_backends('cupy',
-                      reason="cupy/cupy#8391")
+    @skip_xp_backends('cupy', reason="cupy/cupy#8391")
```
mdhaber, line 136: Issue was closed and locally all tests pass. Let's remove all of these.
```suggestion
```

scipy/stats/tests/test_stats.py
```diff
@@ -3639,8 +3633,9 @@ class SkewKurtosisTest:
     testmathworks = [1.165, 0.6268, 0.0751, 0.3516, -0.6965]
 
 
-@make_skip_xp_backends("skew")
+@make_skip_xp_backends(stats.skew)
 class TestSkew(SkewKurtosisTest):
+    @make_skip_xp_backends(stats.kurtosis)
     @pytest.mark.parametrize('stat_fun', [stats.skew, stats.kurtosis])
     def test_empty_1d(self, stat_fun, xp):
```
mdhaber, line 3640: Could you try moving this to `SkewKurtosisTest` and making `stat_fun` an attribute in each class? Would be better than applying `@make_skip_xp_backends(stats.kurtosis)` here, I think.

scipy/stats/tests/test_stats.py
```diff
@@ -5965,226 +5960,223 @@ def test__broadcast_concatenate():
...
-    # regression test for gh-6134: nans in the second arg were not handled
-    x = [np.nan, 2.0, 3.0, 4.0]
-    y = [1.0, 2.0, 1.0, 2.0]
-
-    r1 = stats.ttest_ind(x, y, nan_policy='omit')
-    r2 = stats.ttest_ind(y, x, nan_policy='omit')
-    assert_allclose(r2.statistic, -r1.statistic, atol=1e-15)
-    assert_allclose(r2.pvalue, r1.pvalue, atol=1e-15)
+@make_skip_xp_backends(stats.ttest_ind)
+class TestTTestInd:
```
mdhaber, line 5964: Thank you for doing this. Should the stragglers above be moved over? And maybe it makes sense to have a separate class for combined tests rather than putting this first test in `TestTTestInd`. It's definitely a test of both, not a test of one against the other.

scipy/stats/tests/test_stats.py
mdhaber: I was a little concerned by the size of the diff but it's not too bad.

mdhaber:
The failures could be due to passing an empty exceptions list to the `xp` test skip decorator factory. Or something like that. I seem to remember from one of the original PRs that passing an argument that wasn't really set led to the decorator being ignored.

ev-br:
One possibly relevant comment. I've a MWE level hack where I 
- instrument the import machinery to record the backends of function arguments. 
- hook into a pytest machinery to record the data during a test run.

The goal is to have a record of which scipy functions are tested for which backends. 

The hack is at https://github.com/scipy/scipy/compare/main...ev-br:scipy:messing_with_import_hooks?expand=1 (the import machinery hooks are courtesy of ChatGPT).

Why I'm bringing it up here. Using functions are dict keys is one thing that throws off this kind of hacks, because functions get decorated on import. So if we ever consider using this sort of automated reporting, we may prefer going for using `func.__qualname__` analogs as keys instead of functions themselves).

mdhaber:
Can you describe the problem in more detail? Is the problem that this will screw up your machinery, or is it that our `xp_capabilities_table` (using function handle keys) is difficult to compare against your harvested information?

crusaderky:
I didn't notice this and wrote a parallel PR: https://github.com/scipy/scipy/pull/22686

ev-br:
> Can you describe the problem in more detail? or is it that our xp_capabilities_table (using function handle keys) is difficult to compare against your harvested information? 

I'm not sure if there is a problem. I only noticed that decorating on import messes up `iirfilter` which relies on fetching functions stored in dicts: https://github.com/scipy/scipy/blob/f2445c5758284ea0739972f476622154ae117b12/scipy/signal/_filter_design.py#L2632
and 
https://github.com/scipy/scipy/blob/f2445c5758284ea0739972f476622154ae117b12/scipy/signal/_filter_design.py#L2666

j-bowhay:
I addressed most of the comment but did not have time to debug the CI failures

scipy/_lib/_array_api.py
```diff
@@ -711,21 +710,22 @@ def make_skip_xp_backends(*funs, capabilities_table=None):
...
+    cpu_only_reason = set()
     np_only = False
     exceptions = []
 
     for fun in funs:
         skip_backends += capabilities_table[fun].get('skip_backends', [])
         cpu_only |= capabilities_table[fun].get('cpu_only', False)
-        cpu_only_reason += capabilities_table[fun].get('reason', "")
+        # Empty reason causes the decorator to have no effect
+        cpu_only_reason.add(capabilities_table[fun].get('reason', "No reason given."))
```
mdhaber, line 721: For whatever reason, if an empty `reason` is passed as a keyword argument, the `cpu_only` decorator doesn't do its job. Simplest fix to get CI passing (hopefully) is to provide a reason. If this is all going away, that should be find to get the rest of this PR merged.

scipy/stats/tests/test_stats.py
```diff
@@ -3632,24 +3627,25 @@ class SkewKurtosisTest:
...
                 sup.filter(RuntimeWarning, "Mean of empty slice")
                 sup.filter(RuntimeWarning, "invalid value encountered")
-                res = stat_fun(x)
+                res = self.stat_fun(x)
         xp_assert_equal(res, xp.asarray(xp.nan))
 
+
+@make_skip_xp_backends(stats.skew)
+class TestSkew(SkewKurtosisTest):
+    def stat_fun(self, x):
```
mdhaber, line 3646: With `stat_fun = stats.skew`, `stats.skew` was being passed `self` as the first argument.

mdhaber (approved):

Issue #22635, opened 2025-03-05
DOC: optimize.root: default values of optional parameters not documented


Mark799:
### Issue with current documentation:

`scipy.optimize.root` lacks documentation for default values of optional parameters, for example [here](https://docs.scipy.org/doc/scipy/reference/optimize.root-hybr.html).

Values, like `xtol` and `eps` are unknown to the user, unless specified. To be honest, this can be frustrating. Could you add the default values in the docstring?

lucascolley:
PR welcome!

MsMansiDhruv:
I came across this issue and wanted to contribute by adding the default values for optional parameters in scipy.optimize.root. I checked the documentation and source code but couldn't find explicit default values for parameters like xtol and eps. Are these set dynamically based on the solver method, or is there a reference where these defaults are defined? Any pointers would be appreciated—I'd love to help improve the documentation!

lucascolley:
Hi @MsMansiDhruv, to take the example of `hybr`:
-	The signature with defaults can be found at https://github.com/scipy/scipy/blob/50a007e66f28021887854a769ad639c2903e2dad/scipy/optimize/_minpack_py.py#L193-L195
- You can see that for some arguments like `xtol` here, the default is in the signature. For others, like `eps` here, the default is `None` in the signature and the value is set within the function: https://github.com/scipy/scipy/blob/50a007e66f28021887854a769ad639c2903e2dad/scipy/optimize/_minpack_py.py#L239-L240

I think it would make sense to add these defaults to the descriptions of the parameters, e.g. at https://github.com/scipy/scipy/blob/50a007e66f28021887854a769ad639c2903e2dad/scipy/optimize/_minpack_py.py#L205-L207

Following https://numpydoc.readthedocs.io/en/latest/format.html#parameters, this can follow the type of the parameter, e.g. `xtol : float, default: 1.49012e-08`. Sometimes we instead include the default in the description of parameter, but I don't know of a hard rule about when either style is preferable, so either would be fine for a first PR!

The docstrings of other methods to `hybr` are located in other places in `scipy/optimize`, and I can point you to them if you can't find them. For a first PR, though, it would be fine to just look at `hybr`!

lucascolley:
As a side note, it's a shame we don't have a `[show source]` button on pages like https://docs.scipy.org/doc/scipy/reference/optimize.root-hybr.html#optimize-root-hybr. Perhaps you have an idea why we don't have that @melissawm ?

[...]

PR #22679, opened 2025-03-13
DOC: optimize.root: include default values for method args

MsMansiDhruv:
#### Reference issue
Closes #22635

Updated docstrings to include default values for optional parameters for methods within `scipy.optimize.root`

#### Additional information
The PR only updates docstrings of method: `hybr`, `lm` and `df-sane` within `scipy.optimize.root`

### DIFF
--- a/scipy/optimize/_minpack_py.py
+++ b/scipy/optimize/_minpack_py.py
@@ -199,31 +199,31 @@ def _root_hybr(func, x0, args=(), jac=None,
 
     Options
     -------
-    col_deriv : bool
+    col_deriv : bool, default: 0
         Specify whether the Jacobian function computes derivatives down
         the columns (faster, because there is no transpose operation).
-    xtol : float
+    xtol : float, default: 1.49012e-08
         The calculation will terminate if the relative error between two
         consecutive iterates is at most `xtol`.
-    maxfev : int
+    maxfev : int, default: 0
         The maximum number of calls to the function. If zero, then
         ``100*(N+1)`` is the maximum where N is the number of elements
         in `x0`.
-    band : tuple
+    band : tuple, default: None
         If set to a two-sequence containing the number of sub- and
         super-diagonals within the band of the Jacobi matrix, the
         Jacobi matrix is considered banded (only for ``jac=None``).
-    eps : float
+    eps : float, default: None
         A suitable step length for the forward-difference
         approximation of the Jacobian (for ``jac=None``). If
         `eps` is less than the machine precision, it is assumed
         that the relative errors in the functions are of the order of
         the machine precision.
-    factor : float
+    factor : float, default: 100
         A parameter determining the initial step bound
         (``factor * || diag * x||``). Should be in the interval
         ``(0.1, 100)``.
-    diag : sequence
+    diag : sequence, default: None
         N positive entries that serve as a scale factors for the
         variables.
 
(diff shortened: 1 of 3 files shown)


scipy/optimize/_minpack_py.py
```diff
@@ -199,31 +199,31 @@ def _root_hybr(func, x0, args=(), jac=None,
 
     Options
     -------
-    col_deriv : bool
+    col_deriv : bool, default: 0
```
lucascolley, line 202: @jorenham I see this is annotated as `onp.ToBool` in scipy-stubs. Do you have an opinion on this line of code?

scipy/optimize/_minpack_py.py
```diff
@@ -199,31 +199,31 @@ def _root_hybr(func, x0, args=(), jac=None,
...
         If set to a two-sequence containing the number of sub- and
         super-diagonals within the band of the Jacobi matrix, the
         Jacobi matrix is considered banded (only for ``jac=None``).
-    eps : float
+    eps : float, default: None
         A suitable step length for the forward-difference
         approximation of the Jacobian (for ``jac=None``). If
         `eps` is less than the machine precision, it is assumed
         that the relative errors in the functions are of the order of
         the machine precision.
```
lucascolley, line 221: In order to close gh-22635, I think we ought to give the user an indication of what the default step length is here (i.e. which `float` corresponds to `None`). 

In the code we have

```python
shape, dtype = _check_func('fsolve', 'func', func, x0, args, n, (n,))
if epsfcn is None:
    epsfcn = finfo(dtype).eps
```

so depending on how the `dtype` returned from `_check_func` varies based on the other inputs, we may be able to give a reasonably concise description of how the default is chosen. I haven't looked into this yet, but no worries if you aren't sure what to write here!

lucascolley: Since we also have

> If `eps` is less than the machine precision, it is assumed that the relative errors in the functions are of the order of the machine precision.

I wonder whether it would be backwards-compatible to change the default argument to `0.0` (as is the case for `_root_leastsq`).

lucascolley:
thanks for the PR @MsMansiDhruv ! I left a few comments (not a full review just yet) but this looks pretty good!

scipy/optimize/_minpack_py.py
jorenham: Thanks for the ping.

It's more of an observation that an opinion, but `0` isn't a `bool`:

```pycon
>>> isinstance(0, bool)
False
```

But throughout the scipy Python codebase, "C-style booleans" like these are pretty common. In fact, it's so common, that I defined `onp.ToBool` (in `optype.numpy`) as

```py
ToBool: TypeAlias = bool | np.bool | Literal[0, 1]
```

Perhaps it feels "wrong" to do define it like this, but it's very practical. In situations where you don't want to accept `Literal[0, 1]`, you could instead use `onp.ToJustBool`, which aliases `bool | np.bool`.

Relevant docs: https://github.com/jorenham/optype/blob/master/README.md#array-likes

lucascolley: thanks, makes sense. I would probably like to defer to a more seasoned maintainer on this one on what would be both 'scipy-like' and roughly accurate.

jorenham: Calling the type a `bool-like` or something might be an easy fix to make this type-valid.

Issue #22680, opened 2025-03-13
BUG: error for arrays on non-default device 


lesteve:
### Describe your issue.

`scipy.special.logsumexp` raise `ValueError` when called on an array with non-default device. This was seen in https://github.com/scikit-learn/scikit-learn/pull/30777.

### Reproducing Code Example

```python
from scipy.special import logsumexp
import array_api_strict
arr = array_api_strict.asarray([1, 2, 3], device=array_api_strict.Device('device1'))
logsumexp(arr)
```

### Error message

```shell
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[4], line 1
----> 1 logsumexp(arr)

File ~/micromamba/envs/scikit-learn-dev/lib/python3.13/site-packages/scipy/special/_logsumexp.py:118, in logsumexp(a, axis, b, keepdims, return_sign)
    116 if xp_size(a) != 0:
    117     with np.errstate(divide='ignore', invalid='ignore'):  # log of zero is OK
--> 118         out, sgn = _logsumexp(a, b, axis=axis, return_sign=return_sign, xp=xp)
    119 else:
    120     shape = np.asarray(a.shape)  # NumPy is convenient for shape manipulation

File ~/micromamba/envs/scikit-learn-dev/lib/python3.13/site-packages/scipy/special/_logsumexp.py:215, in _logsumexp(a, b, axis, return_sign, xp)
    209 m = (xp.sum(i_max_dt, axis=axis, keepdims=True, dtype=a.dtype) if b is None
    210      else xp.sum(b * i_max_dt, axis=axis, keepdims=True, dtype=a.dtype))
    212 # Arithmetic between infinities will introduce NaNs.
    213 # `+ a_max` at the end naturally corrects for removing them here.
    214 # TODO hack device=a_max.device
--> 215 shift = xp.where(xp.isfinite(a_max), a_max, xp.asarray(0, dtype=a_max.dtype)) # , device=a_max.device))
    217 # Shift, exponentiate, scale, and sum
    218 exp = b * xp.exp(a - shift) if b is not None else xp.exp(a - shift)

File ~/micromamba/envs/scikit-learn-dev/lib/python3.13/site-packages/array_api_strict/_searching_functions.py:124, in where(condition, x1, x2)
    121     raise TypeError("`condition` must be have a boolean data type")
    123 if len({a.device for a in (condition, x1, x2)}) > 1:
[...]

PR #22683, opened 2025-03-14
MAINT: remove outdated `xp_` functions, `xp.asarray` on elementwise function args

mdhaber:
#### Reference issue
gh-22680

Now that the 2024.12 standard is out and array-api-compat supports it, I thought it was time for a cleanup of some artifacts from the old days of array API translations. 

One of the things we used to have to do was convert arguments to `xp.where`, `xp.minimum`, etc., into arrays rather than leaving them as scalars. Even if we were careful to use the correct `dtype`, I don't think anyone ensured that the array ended up on the correct device. Use of a Python scalar avoids both of these gotchas.

This cleans up most of the now-unnecessary array conversions in the context of `xp.where`, `xp.minimum`, `xp.clip`, and some other elementwise functions mentioned in https://github.com/data-apis/array-api/issues/807#issuecomment-2639477476. I also noticed some old custom `xp_` functions that we no longer need since they are provided by array-api-compat.

Lots of lines and files are changed, but it should be pretty straightforward.

#### Additional information
Based on gh-22680, I think that anywhere we create an array within a function, we need to pay attention to the device. This only avoids some of those places. To resolve that issue, we really should test with non-default devices (and make the required changes).

There may be some test skips we can remove now. I think we can do that as part of follow-up to gh-22609/gh-22677.

Use of `result_type` needs very special attention; see https://github.com/data-apis/array-api-compat/issues/223#issuecomment-2628684979. Now that it's possible, we need to make sure we pass the arguments to `result_type` before they become arrays.

### DIFF
--- a/scipy/_lib/_array_api.py
+++ b/scipy/_lib/_array_api.py
@@ -43,8 +43,7 @@ __all__ = [
     'is_array_api_strict', 'is_complex', 'is_cupy', 'is_jax', 'is_numpy', 'is_torch', 
     'SCIPY_ARRAY_API', 'SCIPY_DEVICE', 'scipy_namespace_for',
     'xp_assert_close', 'xp_assert_equal', 'xp_assert_less',
-    'xp_copy', 'xp_copysign', 'xp_device',
-    'xp_moveaxis_to_end', 'xp_ravel', 'xp_real', 'xp_sign', 'xp_size',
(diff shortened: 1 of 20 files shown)


mdhaber:
Failures are due to data-apis/array-api-compat#271 and data-apis/array-api-strict#131.

j-bowhay (approved):
Thanks for the tidy up!

PR #22686, opened 2025-03-14
TST/DOC: `lazy_xp_backends` in `xp_capabilities`

crusaderky:
- Incorporate `lazy_xp_backends` into `xp_capabilities`. This allows informing final users if a function:
  - Fully works with JAX or only without `jax.jit`;
  - Fully works with Dask, or if there are spurious calls to `compute()` that harm performance

- Test jax.jit and Dask compute() for all functions decorated with `@xp_capabilities`
- Enable IDE autocompletion when typing parameters of `@xp_capabilities`
- In
 `@xp_capabilities(cpu_only=True, reason=...)`
`@xp_capabilities(np_only=True, reason=...)`,
fix issue where the `reason` was silently suppressed and replaced with the default one.
- Add support for explicit `reason=None` in skip_xp_backends` / `xfail_xp_backends`

### Demo render
image

### Follow-ups
These have been left as follow-ups to keep the size of the PR short:
- Decorate `scipy.cluster` with `xp_capabilities` 
- Mass rename `make_skip_xp_backends` to `make_xp_test_case`


scipy/_lib/_array_api.py
```diff
@@ -600,140 +601,303 @@ def is_marray(xp):
...
+            def test_func(xp):
+                ...
+        """
+        self._lazy_xp_function()
+        for mark in self._make_pytest_marks():
+            test_func = mark(test_func)
+        return test_func
 
-    import pytest
+    def pytest_param(self):
```
crusaderky, line 856: This is unused for now but it will feature prominently in `scipy.special.tests.test_alternative_backends`, and in general whenever functions are passed to `@pytest.mark.parametrize`

scipy/constants/tests/test_constants.py
```diff
@@ -2,12 +2,10 @@
 
 import scipy.constants as sc
 from scipy._lib._array_api_no_0d import xp_assert_equal, xp_assert_close
-from scipy._lib._array_api import make_skip_xp_backends
 
-skip_xp_backends = pytest.mark.skip_xp_backends
+lazy_xp_modules = [sc]
```
crusaderky, line 6: requires https://github.com/data-apis/array-api-extra/pull/158

scipy/constants/_constants.py
```diff
@@ -225,7 +225,7 @@
 # functions for conversions that are not linear
 
 
-@xp_capabilities()
+@xp_capabilities(static_argnames=("old_scale", "new_scale"))
```
crusaderky, line 228: I had an offline conversation with @pearu regarding the opportunity to automatically wrap these functions with `jax.jit` _in production_, and not just in unit tests, whenever we detect an jnp.Array argument.

The outcome was that, while you can observe very substantial benefits compared to running eagerly, the jit has a cost every time the shape of the input array changes, so it may be detrimental whenever someone iterates on ever-differently-shaped arrays.

scipy/stats/_stats_py.py
```diff
@@ -696,7 +698,7 @@ def tmean(a, limits=None, inclusive=(True, True), axis=None):
     return mean[()] if mean.ndim == 0 else mean
 
 
-@xp_capabilities("trimmed statistics")
+@xp_capabilities(like=tmean)
```
crusaderky, line 701: I foud that, once you add `static_argnames` to the decorator, the likelyhood of getting exactly identical capabilities between two functions are slim. So I opted for a simpler design that doesn't involve a global state.

scipy/stats/_stats_py.py
```diff
@@ -756,7 +758,9 @@ def tvar(a, limits=None, inclusive=(True, True), axis=0, ddof=1):
         return _xp_var(a, correction=ddof, axis=axis, nan_policy='omit', xp=xp)
 
 
-@xp_capabilities("trimmed statistics")
+@xp_capabilities(skip_backends=[("jax", "JAX doesn't allow item assignment"),
+                                ("dask", "lazywhere doesn't work with dask")],
+    xfail_backends=[("array_api_strict", "broadcast int dtype vs. xp.nan")])
```
crusaderky, line 763: moved this line from unit tests

scipy/_lib/_array_api.py
```diff
@@ -600,140 +601,303 @@ def is_marray(xp):
...
+    # Enforced by xpx.lazy_xp_function.
+    allow_compute: int = 0
+
+    # If False, chunks will be squashed into a single one, which
+    # is likely to harm performance. This typically happens when
+    # the developer implementing Array API support is unsure on
+    # whether there are any reductions along axes (a function that
+    # performs reductions may return outputs that change depending
+    # on chunk size!)
+    chunked: bool = True
```
crusaderky, line 640: Use case for this will appear in https://github.com/scipy/scipy/pull/22342
I'm annoyed that it's a purely informative flag for now that needs to be kept in sync by hand. It would be far from trivial to test its correctness automatically.

scipy/_lib/_array_api.py
```diff
@@ -600,140 +601,303 @@ def is_marray(xp):
...
-        wrapper.__doc__ = str(doc).split("\n", 1)[1]  # remove signature
+    def test_case(self, test_func):
+        """Return a pytest decorator and apply lazy_xp_function::
 
-        return wrapper
-    return decorator
+            from scipy.mymod import func
+
+            @func.capabilities.skip_xp_backends
+            def test_func(..., xp):
```
crusaderky, line 835:
```suggestion
            def test_func(xp):
```

mdhaber:
There were some other valuable things in gh-22677, which I already reviewed and was almost ready to go, so I'd like to get that in, first. In the meantime, although the top post suggests that this adds support lazy backends, it looks like a complete rewrite. It would be helpful to know why lazy backends wouldn't easily fit into the existing structure.

crusaderky:
> There were some other valuable things in [gh-22677](https://github.com/scipy/scipy/pull/22677), which was posted first. It seemed ready to go, so I'd like to get that in first.

Sure, go on and merge it, I'll rework this PR on top of it.

> In the meantime, although the top post suggests that this just adds support for lazy backends, it looks like a complete rewrite of this system. It would be helpful to know why lazy backends wouldn't easily fit into the existing structure.

- the extra jax and dask specific params weren't great with the previously dynamically-defined dataclasses. This looks a lot cleaner.
- I had to rename `make_skip_backends` because it no longer just generates `@skip_xp_backends`; it also applies `lazy_xp_function`. I went for `<func>.capabilities.test_case` but if you prefer I can do `xp_test_case(func)`.
- Adding the `static_argnames` to each function made the benefits given by a table with multiple functions using the same capabilities much less appealing; so I simplified.
- The rst table generation now has to deal with cells of drastically variable width, so the previous semi-static template was no longer a great choice.
- Adding lazy_xp_function increased complexity, which IMHO made the previous function-based design start looking like a poorer fit than a (IMHO cleaner) class object.

j-bowhay:
What's the status here? I would like to start using `xp_capabilities` to advertising support for more functions but I don't want to invest the time to do so if its going to be drastically changed.

crusaderky:
> What's the status here? I would like to start using `xp_capabilities` to advertising support for more functions but I don't want to invest the time to do so if its going to be drastically changed.

I need to rework this PR.
Ultimately you should expect `xp_capabilities` to gain optional parameters and `lazy_xp_backends` to become seldom used explicitly. 
You don't need to wait for this PR to be merged before you start. Please do go on and decorate functions with the current version of `xp_capabilities`.

mdhaber:
> Please do go on and decorate functions with the current version

https://github.com/scipy/scipy/pull/22677#issuecomment-2731583443

scipy/_lib/_array_api.py
```diff
@@ -736,41 +776,83 @@ def wrapper(*args, **kwargs):
...
+
+        lazy_kwargs = {k: capabilities[k]
+                       for k in ('allow_dask_compute', 'jax_jit',
+                                 'static_argnums', 'static_argnames')}
+        lazy_xp_function(func, **lazy_kwargs)
+
+    return marks
+
+
+def make_skip_xp_backends(*funcs, capabilities_table=None):
```
crusaderky, line 811: This is no longer just about skipping tests, as it also applies `lazy_xp_function`.
I will mass rename it to `make_xp_test_case` in a follow-up PR. I didn't to it here in the interest of keeping the PR readable.

scipy/_lib/_array_api.py
```diff
@@ -736,41 +776,83 @@ def wrapper(*args, **kwargs):
...
+
+    - Generate the @pytest.mark.skip_xp_backends and
+      @pytest.mark.xfail_xp_backends decorators for the decorated test function
+    - Tag the function with `xpx.lazy_xp_function`
+    """
+    marks = _make_xp_pytest_marks(*funcs, capabilities_table=capabilities_table)
+    return lambda func: functools.reduce(lambda f, g: g(f), marks, func)
+
+
+def make_xp_pytest_param(func, capabilities_table=None):
```
crusaderky, line 828: Unused for now; will be used in a follow-up in `scipy.special.tests.test_support_alternative_backends`

scipy/conftest.py
```diff
@@ -304,21 +304,20 @@ def _backends_kwargs_from_request(request, skip_or_xfail):
                              f"must be a subset of {list(xp_known_backends)}")
 
         if marker.kwargs.get('np_only', False):
-            reason = marker.kwargs.get(
-                "reason", "do not run with non-NumPy backends")
+            reason = marker.kwargs.get("reason") or "do not run with non-NumPy backends"
```
crusaderky, line 307: Disregard explicit reason=None from `@make_skip_xp_backends`

scipy/conftest.py
```diff
@@ -403,9 +402,10 @@ def skip_or_xfail_xp_backends(request: pytest.FixtureRequest,
         request, skip_or_xfail=skip_or_xfail
     )
     xp = request.param
-    skip_or_xfail = getattr(pytest, skip_or_xfail)
-    reason = skip_xfail_reasons.get(xp.__name__)
-    if reason:
+    if xp.__name__ in skip_xfail_reasons:
```
crusaderky, line 405: This was causing the test not to be skipped as we ended up with something like `{torch: None}` due to the explicit reason=None from make_skip_xp_backends.

scipy/stats/_stats_py.py
```diff
@@ -6463,7 +6476,13 @@ def ttest_ind_from_stats(mean1, std1, nobs1, mean2, std2, nobs2,
 _ttest_ind_dep_msg = "Use ``method`` to perform a permutation test."
 
 
-@xp_capabilities(cpu_only=True, exceptions=["cupy", "jax.numpy"])
```
crusaderky, line 6466: out of scope: we have no means of testing that jax isn't actually quietly dragging everything from GPU memory to CPU memory and converting to numpy because of https://github.com/jax-ml/jax/issues/25998

scipy/stats/_stats_py.py
```diff
@@ -7052,6 +7071,8 @@ def _get_len(a, axis, msg):
     return n
 
 
+@xp_capabilities(cpu_only=True, exceptions=["cupy", "jax.numpy"],
+                 jax_jit=False, allow_dask_compute=999)
```
crusaderky, line 7075: this was missing

crusaderky:
@mdhaber @lucascolley @rgommers this is now ready for review.
I've reworked it to keep the design as similar as it was before.
See updated opening post and various commentary.

lucascolley:
I skimmed the diff and this looks like a very nice improvement!

Now that we have a pretty solid level of granularity, we may want to rethink how we are tracking progress in gh-18867

mdhaber (commented):
A few comments from what I saw, but mostly I'm reviewing by test. 

The tables seem to render nicely and reflect what is specified by the decorator. [`gmean`](https://output.circle-artifacts.com/output/job/327cdea5-66db-4553-903b-3ee781514f81/artifacts/0/html/reference/generated/scipy.stats.mstats.gmean.html), [`hmean`](https://output.circle-artifacts.com/output/job/327cdea5-66db-4553-903b-3ee781514f81/artifacts/0/html/reference/generated/scipy.stats.mstats.hmean.html),  and [`pearsonr`](https://output.circle-artifacts.com/output/job/327cdea5-66db-4553-903b-3ee781514f81/artifacts/0/html/reference/generated/scipy.stats.pearsonr.html) are pretty representative of the cases we can see with the functions that are currently decorated. 

`skip_backends` and `xfail_backends` seem to work. Do we need to have a way to make `xfail_backends` strict or not?

I see that `cpu_only=True` skips CuPy on my machine, as expected. If it does not work to skip other GPU tests, I think we'll notice pretty quickly in CI.

`np_only` isn't working as I would expect. I added `np_only=True` to  the decorator factory on `skew`, and it only skipped CuPy.

I'm not certain how to test the rest.

Thoughts on how `marray` might fit into this? Fine for that to be a follow-up. 
I was thinking about an `marray_backends` argument that accepts the backends with which `marray` should be tested. Would this be represented with a column in the table if specified? Do we need to specify combinations of backend and device separately? Since `marray` support will initially only be in `stats`, I think it's fine for *no* tests to be performed with `marray` by default and for such a column to render only if `marray_backends` is specified.
There would still be some other work to do with this approach. For instance, tests like `xp_assert_equal(stats.skew(xp.asarray([]), xp.asarray(xp.nan))` will fail because MArray would produce a masked value rather than a NaN. We could add logic to `xp_assert_`s to make them equivalent, but it is possible that there are (rare) cases in which the distinction should be made. It might be easiest to only run tests with `marray` if they were written with `marray` in mind (e.g. those in `stats/tests/test_marray.py`).

scipy/_lib/_array_api.py
```diff
@@ -633,101 +633,132 @@ def is_marray(xp):
...
-def xp_capabilities(capabilities_table=None, **kwargs):
+def xp_capabilities(
+    capabilities_table=None,
+    # lists of tuples [(module name, reason), ...]
+    skip_backends=(), xfail_backends=(),
+    # @pytest.mark.skip/xfail_xp_backends kwargs
+    cpu_only=False, np_only=False, reason=None, exceptions=(),
+    # xpx.lazy_xp_backends kwargs
+    allow_dask_compute=0, jax_jit=True,
+    static_argnums=None, static_argnames=None,
```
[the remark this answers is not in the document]
crusaderky: Done

crusaderky:
> Would it help for our assertions to explicitly compare the actual and desired device? e.g., check_device in addition to check_dtype, etc...?

Yes, that's a good idea. It would catch the case of GPU arrays that are sent back to CPU and then left there, but not the case of a full round trip through `xpx.lazy_apply`. Let's do it downstream of https://github.com/data-apis/array-api-extra/pull/267 though.

crusaderky:
FYI the follow-up to this is ready: https://github.com/data-apis/array-api-extra/pull/284
I'd rather not invert the merge order, if anything for the sake of everyone's time.

doc/source/dev/api-dev/array_api.rst
```diff
@@ -92,7 +92,7 @@ More strict array input validation will reject ``np.matrix`` and
 Currently supported functionality
 `````````````````````````````````
 
-The following modules provide array API standard support when the environment
+The following modules provide Array API Standard support when the environment
```
lucascolley, line 95:
```suggestion
The following modules provide array API standard support when the environment
```

https://github.com/data-apis/array-api/discussions/778

lucascolley (approved):
LGTM, thanks Guido!

crusaderky:
@lucascolley @mdhaber is there anything outstanding here?

lucascolley:
Not from my side, will merge soon if no further comments

lucascolley:
thanks again all!

PR #22687, opened 2025-03-14
MAINT: Bump Array API to 2024.12

crusaderky:


scipy/integrate/_tanhsinh.py
```diff
@@ -470,10 +470,7 @@ def customize_result(res, shape):
         # If the integration limits were such that b < a, we reversed them
         # to perform the calculation, and the final result needs to be negated.
         if log and xp.any(negative):
-            dtype = res['integral'].dtype
-            pi = xp.asarray(xp.pi, dtype=dtype)[()]
-            j = xp.asarray(1j, dtype=xp.complex64)[()]  # minimum complex type
-            res['integral'] = res['integral'] + negative*pi*j
+            res['integral'] = res['integral'] + negative * xp.pi* 1.0j
```
lucascolley, line 473:
```suggestion
            res['integral'] = res['integral'] + negative * xp.pi * 1.0j
```

lucascolley (approved):
thanks @crusaderky ! We can deal with any others in a follow-up if we find them.

mdhaber:
No, I haven't kept track. How did you find these? 

gh-22683 took care of most calls to functions like `where` and `minimum`, but I know there are a sprinkling of those left that we will find eventually. I planned to submit a PR for `result_type` shortly.

I think we also need to look out for *all* array creation inside functions because I don't think we've ever thought to match the `device` (gh-22680). Would you know how to set up CI to test that?

crusaderky:
> No, I haven't kept track. How did you find these?

just a text search for `xp.asarray(0`, `xp.asarray(1`, `xp.zeros(()`, `xp.ones(()`, `xp.full(()`

crusaderky:
> I think we also need to look out for _all_ array creation inside functions because I don't think we've ever thought to match the `device` ([gh-22680](https://github.com/scipy/scipy/issues/22680)). Would you know how to set up CI to test that?

We could do this as follows:
1. add a "default device" config flag to array_api_strict
2. in `lazy_xp_function`, temporarily tamper with the default device when xp=array_api_strict and set it to an inane one
3. in `xp_assert_equal` et al, verify that the devices of actual and expect match.

PR #22690, opened 2025-03-15
MAINT/DOC: integrate.tanhsinh: lightly refactor error estimate and correct documentation

### LINKED ISSUE #22681: DOC: integrate.tanhsinh: documentation refers to non-existent attribute

https://github.com/scipy/scipy/blob/2532ffdfa85179b0921e1ef99012fe80b622ba78/scipy/integrate/_tanhsinh.py#L100

In the above `df` does not exist in the actual result object.

I'm guessing this should be `res.integral` but thought I would check first @mdhaber 

_No response_

mdhaber:
#### Reference issue
Closes gh-22681

gh-22681 noted that `integrate.tanhsinh` documentation had some leftover copypasta from `differentiate` in the description of `rtol` and `atol`. This fixes that and does some housekeeping to simplify the error estimate and associated termination condition.

#### Additional information
To do:
- https://github.com/scipy/scipy/issues/22681#issuecomment-2725648042. I think the `d1` estimate should be zero in this case based on the fact that (working in log-space) $\lim_{d_1 \rightarrow -\infty} \frac{d_1^2}{d_2} = -\infty$. Do you agree? If so, I'll add the logic. I've already added the unit test so we can confirm that it fails in main (since it didn't on one of my machines).
- https://github.com/scipy/scipy/issues/22681#issuecomment-2725656715. While I'm at it, I'd like to add the factor of $h$ if it is appropriate. In that case, I'd make the change and add your example problem as a test.


j-bowhay:
Thanks, just need to make time to look at the references and then I'll take a look at this.

j-bowhay:
> * [DOC: integrate.tanhsinh: documentation refers to non-existent attribute #22681 (comment)](https://github.com/scipy/scipy/issues/22681#issuecomment-2725648042). I think the `d1` estimate should be zero in this case based on the fact that (working in log-space) 
Github has slightly butchered the latex here, do you mind restating?

mdhaber:
Thank you for taking a look!

j-bowhay:
> * [DOC: integrate.tanhsinh: documentation refers to non-existent attribute #22681 (comment)](https://github.com/scipy/scipy/issues/22681#issuecomment-2725648042). I think the `d1` estimate should be zero in this case based on the fact that (working in log-space) . Do you agree? If so, I'll add the logic. I've already added the unit test so we can confirm that it fails in main (since it didn't on one of my machines).

To confirm, the issue is that approximation to the integral doesn't change between levels? If so the reference seems to special cases this and just sets the error to zero:
image


> * [DOC: integrate.tanhsinh: documentation refers to non-existent attribute #22681 (comment)](https://github.com/scipy/scipy/issues/22681#issuecomment-2725656715). While I'm at it, I'd like to add the factor of 
>     h
>    if it is appropriate. In that case, I'd make the change and add your example problem as a test.

I don't have an opinion on this part yet...

mdhaber:
> If so the reference seems to special cases this and just sets the error to zero:

Huh, yeah, I guess it does. I recall finding it strange that if $|S_n - S_{n-1}| > 0$, we would use a totally different estimate of the error, which could be much larger, and suddenly, when $|S_n - S_{n-1}| $ is indestinguishable from `0` in the chosen precision, the error would jump to _zero_. And of course, I didn't think it was appropriate to return an error estimate of exactly zero - I would at least return `eps * integral`, where `eps` is precision and `integral` is the integral estimate. I suppose that's why I didn't implement that.

But we can do it if you and @steppi agree it's a good idea.

I would still appreciate your thoughts on the factor of $h$ in $d_3$ and $d_4$.

steppi:
> > If so the reference seems to special cases this and just sets the error to zero:
> 
> Huh, yeah, I guess it does. I recall finding it strange that if | S n − S n − 1 | > 0 , we would use a totally different estimate of the error, which could be much larger, and suddenly, when | S n − S n − 1 | is indestinguishable from `0` in the chosen precision, the error would jump to _zero_. And of course, I didn't think it was appropriate to return an error estimate of exactly zero - I would at least return `eps * integral`, where `eps` is precision and `integral` is the integral estimate. I suppose that's why I didn't implement that.
> 
> But we can do it if you and @steppi agree it's a good idea.
> 
> I would still appreciate your thoughts on the factor of h in d 3 and d 4 .

I commented here too, https://github.com/scipy/scipy/issues/22681#issuecomment-2762494417. I think what's going on in the reference is that its heuristic error estimate assumes $\epsilon << h$, since it's working in an arbitrary precision context. The article says that $h=2^{-12}$ is typically sufficient to get 1000 decimal places of accuracy. The heuristic error estimates work by assuming quadratic convergence and can allow you to declare convergence one level earlier compared to just stopping when $|S_{n} - S_{n-1}| = 0$.

I don't think the error estimate in the reference is actually relevant in a double precision context. Clearly we can't just say the error is 0 when $S_n = S_{n-1}$.I think didn't really understand all that was going on yet when I first commented on this error estimate last year. Sorry about that.

steppi:
I'll try to unpack what we know and make a recommendation.  Let $\epsilon = 10^{-p}$, where $p$ is the desired level of precision.

The heuristic error estimate is based on

$$d_1 = \log_{10}\left|S_n - S_{n-1}\right|$$
$$d_2 = \log_{10}\left|S_n - S_{n-2}\right|$$
$$d_3 = \log_{10}\left(\epsilon \max\left|w_j f(x_j)\right|\right)$$
$$d_4= \log_{10}\max\left(w_l f(x_l), w_r f(x_r)\right)$$

where $x_l$ and $x_r$ are the left and right most abscissas respectively. If

$$d = \max(2d_1, \frac{d_1^2}{d_2}, d_3, d_4)$$

then the error estimate is $10^d$.

In plain language, what the first two entries in `max` above are saying is:

$2d_1$: assuming quadratic convergence, we can double the estimated number of digits of accuracy, because we can assume the jump from $S_{n-1}$ to $S_{n}$ doubled the number of accurate digits.

$\frac{d_1^2}{d_2}$: but make sure the the previous two iterations actually showed at least quadratic convergence.

At least this part is straightforward. There seems to be no harm in keeping this as part of the estimate in a double precision context since it can potentially save an iteration.

As we know, the result is computed with an Euler-Maclaurin sum

$$h \sum_{j=-N}^{N}w_j f(x_j)$$

In the reference for the error, the value of $N$ at a level is chosen so that $\left|w_j f(x_j) \right | < \epsilon = 10^{-p}$ for $|j| > N$. This doesn't always make sense in a fixed precision context, but I see you found a good reference https://arxiv.org/pdf/2007.15057, which works out reasonable `tanh-sinh` modifications for float and double precision arithmetic. In this reference, the window of summation is determined by finding a cutoff $t_max$ where the stored abscissa complements underflow (for 1 dimensional integration this is also where the weights underflow).

`tanh-sinh` makes lots of function evaluations very close to the endpoints $-1$ and $1$ of the integral (finite integrals that aren't over $[-1, 1]$ have to be rescaled to this impact), and these are critical places that have the largest impact on the result. For arbitrary precision arithmetic, $$d_4= \log_{10}\max\left(w_l f(x_l), w_r f(x_r)\right)$$, will approach $-p$ as the number of levels increase, because it's possible to take the window all the way up to the $\left|w_j f(x_j) \right | < \epsilon = 10^{-p}$ cutoff. For fixed precision that's no longer the case. I think it's good to keep this, it shows a little piece of error that can't be avoided due to the limitations of our arithmetic.

 $\log_{10}\left(\epsilon \max\left|w_j f(x_j)\right|\right)$ is good to keep too.

### Recommendation

Having looked at this more, I think I was wrong yesterday about the error estimate given not being relevant in a fixed precision context. It's still useful to be able to converge one level earlier and the rough lower bounds that are used in the reference to check for obstructions to quadratic convergence in the previous iteration are still useful for estimating pieces of error that can remain even when $S_n == S_{n-1}$.

What you've done in this PR is what I would have recommended. When $$S_{n} - S_{n-1}$$ is zero, it makes sense to treat the error components dealing using $d_1$ and $d_2$ as zero and only pay attention to the other components.

### $h$ in error estimate

I looked up David Bailey's Fortran code from `mpfun2020` here https://www.davidhbailey.com/dhbsoftware/. It's in the file `tquad.f90`  (I had to unpack the tarball on my machine to look at it).

```f90
!   Compute s1 = current integral approximation and err = error estimate.
!   Tsum is the sum of all tw1 and tw2 from the loop above.
!   Twmx is the largest absolute value of tw1 and tw2 from the loop above.
!   Twi1 and twi2 are the final nonzero values of abs(tw1) and abs(tw2).

  s1 =  mpreal (ax, nwds1) * h * tsum
  eps1 = twmx * epsilon1
  eps2 = abs (max (twi1, twi2) / s1)
  d1 = dplog10 (abs ((s1 - s2) / s1))
  d2 = dplog10 (abs ((s1 - s3) / s1))
  d3 = dplog10 (eps1) - 1.d0
  d4 = dplog10 (eps2) - 1.d0

  if (k <= 2) then
    err = mpreal (1.d0, nwds1)
  elseif (d1 == -999999.d0) then
    err = mpreal (0.d0, nwds1)
  else
    err = c10 ** nint (min (0.d0, max (d1 ** 2 / d2, 2.d0 * d1, d3, d4)))
  endif
```

Everything is being done in terms of relative error not absolute error like in the reference. If we convert these to absolute error, then there still won't be an $h$ factor in `d3` and `d4`.

My gut says there's it's not likely Bailey is wrong here for the arbitrary precision context and it looks like his heuristic is backed up by empirical evidence. I don't have an explaination though. Are we consistently over-estimating the error by a factor of $1/h$ or is this more about trying to get the heuristics to look more sensible?

steppi (commented):
That was a long comment. But the overall gist is that I thought about this a lot, I agree with what you are doing here in this PR. When the test failures are dealt with I think this can be merged.

mdhaber:
Albert and I just talked about this in real-time. We agreed that $d_3$ and $d_4$ seem to be extremely conservative error estimates in our double precision context. Moreover, the whole error estimation scheme, which is designed to save one level of iteration by relying on quadratic convergence, is probably not helpful for us.

The papers by Bailey (e.g. [here](https://www.davidhbailey.com/dhbpapers/dhb-tanh-sinh.pdf) and [here](https://projecteuclid.org/journals/experimental-mathematics/volume-14/issue-3/A-comparison-of-three-high-precision-quadrature-schemes/em/1128371757.pdf)) say:
> One does not need to rely on this estimation scheme if one is willing to continue computation until the quadrature results from two successive levels are in agreement (to within the last few digits).

and

> One does not need to rely on this estimation scheme if one is willing to continue computation until the quadrature results from two successive levels agree to within the full primary precision....

And we think that ultimately, this would be *faster* for us in many problems. This is because the recommended scheme is designed to save *one level* of computation, but the overly-conservative $d_3$ and/or $d_4$ can delay convergence for many levels - or even preclude the possibility. In Jake's example, for example, it is impossible to reduce $d_4$ below ~`200`, causing the algorithm to continue until `maxlevel` is reached, whereas in reality the accuracy doesn't improve much after the fourth level.

The longer-term plan would be to remove this complicated error estimate scheme and just report the difference in the approximation between two successive levels. For this PR, though, I'm just going to special-case $d_1 ^2 / d_2$ to equal $-\infty$ whenever $d_1 = -\infty$ so we don't get NaN for the error. That will fix the failures in the test I added. (The current CI failure is intentional.) We can overhaul the error estimation later.

mdhaber:
Failures are unrelated.

j-bowhay:
Thanks for digging into this, I have been quite busy recently so don't really have anything to add!

steppi (approved):
Thanks @mdhaber. Looks good to me.

mdhaber:
@steppi @j-bowhay what do you think about backporting this as a bug-fix?

steppi:
That sounds reasonable to me.

PR #22691, opened 2025-03-15
DOC: signal: fix `freqz_sos` and `sosfreqz` docstrings

### LINKED ISSUE #22461: DOC: freqz_sos: claims that it was introduced in 0.19; no mention to sosfreqz; no documentation for sosfreqz

[The documentation for `freqz_sos`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.freqz_sos.html) states that the function was introduced in version 0.19; however, this function doesn't exist in versions older than 1.15.  This is because the function was formerly named `sosfreqz`, and when it was renamed to `freqz_sos`, no mention of the old name was left in the documentation, and the claim that the function was introduced in 0.19 has been preserved from the previous docstring (which is incorrect; the *functionality* was introduced in 0.19, under a different name, but the *function* with this name is new).
This is misleading to users of older SciPy versions, who will be surprised when they find that the online documentation mentions a function that supposedly existed since 0.19 but that they don't have in their 1.11.

Furthermore, the `sosfreqz` function still exists for backwards compatibility, but [doesn't appear in the documentation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.sosfreqz.html), which is a shame because it already includes a docstring indicating that it's deprecated and to use `freqz_sos` instead, and users reading [an older version of the API reference](https://docs.scipy.org/doc/scipy-1.14.0/reference/generated/scipy.signal.sosfreqz.html) will get a link to the newest version that currently leads nowhere.

**Suggested resolution:**
* Amend the `freqz_sos` docstring so that it clarifies that it used to be called `sosfreqz` in older versions.
* Consider amending the *"Added in version 0.19.0"* string so that it states that it was added in 1.15 (but that a functionally identical version called `sosfreqz` has existed since 0.19.0), or leave it as it is if you consider that the previous point clarifies this enough.
* Include the obsolete `sosfreqz` in the online documentation so that users of the old API know that they have to rename it.
* Amend the `sosfreqz` docstring to clarify since which version the function has become obsolete.

_No response_

gsreeja11:
closes https://github.com/scipy/scipy/issues/22461

Changes:
1) Amended the freqz_sos Docstring. 
2) Amended the sosfreqz Docstring.
3) Removed sosfreqz from SKIPLIST so that it shows up in the online documentation.

### DIFF
--- a/scipy/signal/__init__.py
+++ b/scipy/signal/__init__.py
@@ -92,7 +92,7 @@ Filter design
    freqs         -- Analog filter frequency response from TF coefficients.
    freqs_zpk     -- Analog filter frequency response from ZPK coefficients.
    freqz         -- Digital filter frequency response from TF coefficients.
-   sosfreqz      -- Digital filter frequency response for SOS format filter.
+   sosfreqz      -- Digital filter frequency response for SOS format filter (legacy).
    freqz_sos     -- Digital filter frequency response for SOS format filter.
    freqz_zpk     -- Digital filter frequency response from ZPK coefficients.
    gammatone     -- FIR and IIR gammatone filter design.
--- a/scipy/signal/__init__.py
+++ b/scipy/signal/__init__.py
@@ -92,6 +92,7 @@ Filter design
    freqs         -- Analog filter frequency response from TF coefficients.
    freqs_zpk     -- Analog filter frequency response from ZPK coefficients.
    freqz         -- Digital filter frequency response from TF coefficients.
+   sosfreqz      -- Digital filter frequency response for SOS format filter.
    freqz_sos     -- Digital filter frequency response for SOS format filter.
    freqz_zpk     -- Digital filter frequency response from ZPK coefficients.
    gammatone     -- FIR and IIR gammatone filter design.
--- a/scipy/signal/_filter_design.py
+++ b/scipy/signal/_filter_design.py
@@ -792,8 +792,10 @@ def freqz_sos(sos, worN=512, whole=False, fs=2*pi):
 
     Notes
     -----
+    This function used to be called ``sosfreqz`` in older versions (≥ 0.19.0)
+    
     .. versionadded:: 1.15.0
-    This function used to be called sosfreqz in older versions (>=0.19.0)
+    
 
     Examples
     --------
--- a/scipy/signal/_filter_design.py
+++ b/scipy/signal/_filter_design.py
@@ -870,10 +870,12 @@ def sosfreqz(*args, **kwargs):
     """
     Compute the frequency response of a digital filter in SOS format.
 
-    .. warning:: This function is an alias, provided for backward
-                 compatibility. New code should use the function
-                 :func:`scipy.signal.freqz_sos`.
-                 This function became obsolete from version 1.15.0.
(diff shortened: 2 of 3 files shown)


scipy/signal/__init__.py
```diff
@@ -92,6 +92,7 @@
    freqs         -- Analog filter frequency response from TF coefficients.
    freqs_zpk     -- Analog filter frequency response from ZPK coefficients.
    freqz         -- Digital filter frequency response from TF coefficients.
+   sosfreqz      -- Digital filter frequency response for SOS format filter.
```
j-bowhay, line 95: Maybe we should put this at the bottom of the page under an aliases section to avoid too much clutter?

scipy/signal/_filter_design.py
```diff
@@ -788,11 +788,12 @@ def freqz_sos(sos, worN=512, whole=False, fs=2*pi):
...
     See Also
     --------
-    freqz, sosfilt
+    freqz, sosfilt, sosfreqz
 
     Notes
     -----
-    .. versionadded:: 0.19.0
+    .. versionadded:: 1.15.0
+    This function used to be called sosfreqz in older versions (>=0.19.0)
```
DietBru, line 796: The missing empty line here cause a failure on the documentation build in CI. 

```suggestion
    This function used to be called ``sosfreqz`` in older versions (≥ 0.19.0)
    
    .. versionadded:: 1.15.0
    
```
As a matter of (my) personal taste, it might be nicer to switch the lines and highlight `sosfreqz`.

PS: As noted in this [checklist], you can build the documentation [locally] to verify that everything is correct.

[checklist]: https://docs.scipy.org/doc/scipy/dev/contributor/development_workflow.html#checklist-before-submitting-a-pr
[locally]: https://docs.scipy.org/doc/scipy/dev/contributor/rendering_documentation.html

scipy/signal/__init__.py
DietBru: Perhaps it might also help to mark this function as "legacy" like:
```
   sosfreqz      -- Digital filter frequency response for SOS format filter (legacy).
```

DietBru (commented):
All in all, it looks good to me. Only a few minor changes—especially to make CI happy.

scipy/signal/__init__.py
gsreeja11: @DietBru Added the "(legacy)" text so the readers of the documentation know that it is an old function.Thanks for the suggestions.
@j-bowhay good point! I would like to take up the task of cleaning up the code by adding all the functions that have aliases in another PR. Could we please get this merged?

scipy/signal/_filter_design.py
gsreeja11: Thanks for the suggestions!

DietBru:
Thanks for your work, @gsreeja11!  
I will merge this PR, once the CI is done and passes.

DietBru (approved):

DietBru:
CI Failure is unrelated.

j-bowhay:
Thanks for your first contribution to SciPy @gsreeja11 keep them coming!

PR #20035, opened 2024-02-05
BUG: spatial.HalfspaceIntersection: raise on non-feasible half space

tylerjereddy:
* Fixes #19865

* Rather than returning a problematic half space intersection polytope, we raise an error for incremental processing when a new half space is added for which the interior point is no longer clearly interior

[skip cirrus]

Issue #19865, opened 2024-01-11
BUG: HalfspaceIntersection.add_halfspaces() does not seem to work as expected


RichardBowenGM:
### Describe your issue.

`scipy` version `1.11.4`.

If I start with 3 halfspaces and a feasible point, then the resulting intersections form a triangle as expected:

<img width="424" alt="image" src="https://github.com/scipy/scipy/assets/131701699/5e578e33-bae1-458f-b03d-d0378ae38886">

(actual numbers in the code example below; initial feasible point shown as 'x', returned vertices shown as 'o').

The intersections there are: 

```
[[ 0.355004   -0.45821925]
 [-0.24133901  0.46546184]
 [ 0.94809281  0.46546184]]
```

if I start with the square [-1,1] × [-1,1] and incrementally add the halfspaces, I expect to get the same result (because that box strictly contains my triangle). Instead, I get out a pentagon:

<img width="457" alt="image" src="https://github.com/scipy/scipy/assets/131701699/ea9396ec-fb4b-4978-877c-77ada8fc83f8">

it seems like the +/- sense of one of the halfspaces (blue line) is being reversed?

### Reproducing Code Example

```python
halfspaces = np.array([
       [-0.70613882, -0.45589431,  0.04178256],
       [ 0.70807342, -0.45464871, -0.45969769],
       [ 0.        ,  0.76515026, -0.35614825]])

feasible_point = np.array([0.35425774, 0.14104826])
vertices = HalfspaceIntersection(halfspaces, feasible_point).intersections
print(vertices) # prints the 3 vertices of the triangle

initial_square =  np.array(
                    [[1, 0, -1], [0, 1, -1], [-1, 0, -1], [0, -1, -1]]
                )

incremental_intersector = HalfspaceIntersection(initial_square, np.zeros(2), incremental=True)
for h in halfspaces:
    incremental_intersector.add_halfspaces(h[np.newaxis])
incremental_intersector.close()
results = incremental_intersector.intersections
print(results) # prints the five vertices of the pentagon
```

### Error message

```shell
(none)
```

### SciPy/NumPy/Python version and system information

```shell
1.11.4 1.23.5 sys.version_info(major=3, minor=9, micro=18, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    detection method: pkgconfig
    found: true
    include directory: /opt/arm64-builds/include
    lib directory: /opt/arm64-builds/lib
    name: openblas
[...]

PR #22693, opened 2025-03-15
BUG: spatial.HalfspaceIntersection: fix `add_halfspaces` batch size support

rsbowen:
#### Reference issue
This [stemmed](https://github.com/scipy/scipy/pull/20035#discussion_r1951830860) from the investigation of, but does not close, https://github.com/scipy/scipy/issues/19865. Please let me know if I should also file a separate bug for this issue.

`add_halfspaces` fails with an array-size issue unless the number of halfspaces being added is `1`, or (I think) `ndim`.  The only places this function was tested is [here](https://github.com/scipy/scipy/blob/main/scipy/spatial/tests/test_qhull.py#L1135), where only a single halfspace is added at a time. I added a simple test and fixed the broadcasting.

#### Additional information
This is my first PR to scipy or indeed to any open source project with this much complexity. Please LMK if I have missed anything.

### DIFF
--- a/scipy/spatial/tests/test_qhull.py
+++ b/scipy/spatial/tests/test_qhull.py
@@ -1203,7 +1203,21 @@ class Test_HalfspaceIntersection:
         hs.add_halfspaces(small_square[0:k,:])
         hs.add_halfspaces(small_square[k:4,:])
         hs.close()
-        # TODO: add correctness test here.
+
+        # Check the intersections are correct (they are the corners of the small square)
+        expected_intersections = np.array([[1., 1.],
+                                           [1., -1.],
+                                           [-1., 1.],
+                                           [-1., -1.]])
+        actual_intersections = hs.intersections
+        # They may be in any order, so just check that under some permutation 
+        # expected=actual.
+        close = np.isclose(actual_intersections[np.newaxis],
+                           expected_intersections[:, np.newaxis])
+        close = np.all(close, axis=-1)
+
+        assert (np.count_nonzero(close, axis=0) == 1).all()
+        assert (np.count_nonzero(close, axis=1) == 1).all()
 
 @pytest.mark.parametrize("diagram_type", [Voronoi, qhull.Delaunay])
 def test_gh_20623(diagram_type):
--- a/scipy/spatial/tests/test_qhull.py
+++ b/scipy/spatial/tests/test_qhull.py
@@ -1199,7 +1199,10 @@ class Test_HalfspaceIntersection:
                                  [ 0.,  1., -1.],
                                  [ 0., -1., -1.]])
 
-        hs = qhull.HalfspaceIntersection(big_square, np.zeros((2,)), incremental=True)
+        hs = qhull.HalfspaceIntersection(big_square,
+                                         np.array([0.3141, 0.2718]),
+                                         incremental=True)
+
         hs.add_halfspaces(small_square[0:k,:])
         hs.add_halfspaces(small_square[k:4,:])
         hs.close()
--- a/scipy/spatial/tests/test_qhull.py
+++ b/scipy/spatial/tests/test_qhull.py
@@ -1215,12 +1215,10 @@ class Test_HalfspaceIntersection:
         actual_intersections = hs.intersections
         # They may be in any order, so just check that under some permutation 
         # expected=actual.
-        close = np.isclose(actual_intersections[np.newaxis],
-                           expected_intersections[:, np.newaxis])
-        close = np.all(close, axis=-1)
 
-        assert (np.count_nonzero(close, axis=0) == 1).all()
-        assert (np.count_nonzero(close, axis=1) == 1).all()
+        ind1 = np.lexsort((actual_intersections[:, 1], actual_intersections[:, 0]))
+        ind2 = np.lexsort((expected_intersections[:, 1], expected_intersections[:, 0]))
+        assert_allclose(actual_intersections[ind1], expected_intersections[ind2])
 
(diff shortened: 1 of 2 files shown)


scipy/spatial/tests/test_qhull.py
```diff
@@ -1186,6 +1186,38 @@ def test_cube(self):
...
+                                           [-1., -1.]])
+        actual_intersections = hs.intersections
+        # They may be in any order, so just check that under some permutation 
+        # expected=actual.
+        close = np.isclose(actual_intersections[np.newaxis],
+                           expected_intersections[:, np.newaxis])
+        close = np.all(close, axis=-1)
+
+        assert (np.count_nonzero(close, axis=0) == 1).all()
+        assert (np.count_nonzero(close, axis=1) == 1).all()
```
tylerjereddy, line 1223: I wonder if it might be slightly clearer to sort the actual and expected arrays lexically and then compare them directly  with `assert_array_equal`:

```diff
--- a/scipy/spatial/tests/test_qhull.py
+++ b/scipy/spatial/tests/test_qhull.py
@@ -1212,12 +1212,9 @@ class Test_HalfspaceIntersection:
         actual_intersections = hs.intersections
         # They may be in any order, so just check that under some permutation 
         # expected=actual.
-        close = np.isclose(actual_intersections[np.newaxis],
-                           expected_intersections[:, np.newaxis])
-        close = np.all(close, axis=-1)
-
-        assert (np.count_nonzero(close, axis=0) == 1).all()
-        assert (np.count_nonzero(close, axis=1) == 1).all()
+        ind1 = np.lexsort((actual_intersections[:, 1], actual_intersections[:, 0]))
+        ind2 = np.lexsort((expected_intersections[:, 1], expected_intersections[:, 0]))
+        assert_array_equal(actual_intersections[ind1], expected_intersections[ind2])
```

tylerjereddy (commented):
This seems sensible and will likely unblock the other bug fix PR indeed.

It makes sense that the intersection of the two convex sets (squares) is the internal one:

<details>

<summary> Sample Plot of the two convex sets (squares) </summary>

image

</details>

One other thing that might be useful to do is update the docstring for `add_halfspaces`  so that the `halfspaces` parameter has a shape that is a bit more specific, like in the parent class:

image

vs. 

image

I double checked that the test fails before/passes after the patch so that's good.

The only other minor comment I'd make is that we try to format our commit messages per the following, if you continue to contribute: https://docs.scipy.org/doc/scipy/dev/contributor/development_workflow.html#writing-the-commit-message

rsbowen:
There is a little more to it. In existing code [here](https://github.com/scipy/scipy/blob/main/scipy/spatial/_qhull.pyx#L459) the size of `dists` is `(num_new_halfspaces,)` and of `points[:,:-1]` is `(num_new_halfspaces, ndim)`. That's why it works when `ndim` is 1. The broadcasting is also legal when `ndim = num_new_halfspaces`, but this does the division along the wrong axis. So you can get incorrect results

```python
big_square = np.array([[ 1.,  0., -2.],
                        [-1.,  0., -2.],
                        [ 0.,  1., -2.],
                        [ 0., -1., -2.]])

small_square = np.array([[ 1.,  0., -1.],
                          [-1.,  0., -1.],
                          [ 0.,  1., -1.],
                          [ 0., -1., -1.]])

hs = HalfspaceIntersection(big_square, np.array([0.3141, 0.2718]), incremental=True)
hs.add_halfspaces(small_square[:2])
hs.add_halfspaces(small_square[2:])
hs.intersections

array([[ 1.    ,  1.5436],
       [-0.3718,  1.5436],
       [ 1.    , -1.    ],
       [-0.3718, -1.    ]])
```

This happens in the (probably common) case that `dists` has more than one value. However the test I wrote doesn't check this, because it chooses the origin as the initial feasible point, which has the same signed distance to every point. So, I updated the test to use a different feasible point, `[0.3141, 0.2718]` as a center; that way, this test also acts a regression test for that bug ([51a9ae9](https://github.com/scipy/scipy/pull/22693/commits/51a9ae941e79025055967e5e943f4390ca9452d9))

rsbowen:
> One other thing that might be useful to do is update the docstring for `add_halfspaces` so that the `halfspaces` parameter has a shape that is a bit more specific, like in the parent class:

[51e3e7f](https://github.com/scipy/scipy/pull/22693/commits/51e3e7f8a5c61f26c2ced20dd2bf6b64b5981a0a)

scipy/spatial/tests/test_qhull.py
rsbowen: Good idea, [8e7360a](https://github.com/scipy/scipy/pull/22693/commits/8e7360a6bc450978d35ef64b703d6a0a48c6beb1)

tylerjereddy:
sounds good, let's let the CI run then probably merge soon

rsbowen:
Not sure what to do about the CI failure. Could it be CI flakyness? ` python dev.py test -t scipy/sparse` passes on my machine with `13274 passed, 1996 skipped, 4808 deselected, 57 xfailed, 3 xpassed in 87.74s (0:01:27)`

tylerjereddy:
> Not sure what to do about the CI failure. 

It is not related. For example, it also shows up in a recent PR that only changes a docstring: https://github.com/scipy/scipy/pull/22698

tylerjereddy:
I pushed in a small commit to fix up a docstring blemish where a few words were showing up as an "extra parameter:"

image

to 

image

tylerjereddy:
@rsbowen thanks for your first contribution to SciPy. I'll ping you for a review on the PR that started all this after I rebase/update it.

PR #22694, opened 2025-03-15
DOC: interpolate.make_smoothing_spline: improve example visibility

### LINKED ISSUE #22692: DOC: interpolate.make_smoothing_spline: example plot uses the same color for points and the fit spline

In the example plot for ``interpolate.make_smoothing_spline``, it is difficult to distinguish between the points and the fit spline since they are the same color (I guess matplotlib advances color cycles separately for ``plot`` and ``scatter``?).

Link to current plot: https://scipy.github.io/devdocs/reference/generated/scipy.interpolate.make_smoothing_spline.html#scipy.interpolate.make_smoothing_spline

Just need to use ``plot`` for the points too, and plot them first so they don't obscure the spline. The modification to the last part of the docstring would look like:

```python
>>> import matplotlib.pyplot as plt
>>> grid = np.linspace(x[0], x[-1], 400)
>>> plt.plot(x, y, '.')
>>> plt.plot(grid, spl(grid), label='Spline')
>>> plt.plot(grid, func(grid), label='Original function')
>>> plt.legend(loc='best')
>>> plt.show()
```

_No response_

derb12:
Improves the visibility of the curves for the example for ``interpolate.make_smoothing_spline``. [docs only]
#### Reference issue

Closes gh-22692

Makes the color for the points and the fit spline different so it is easier to distinguish the two.

### DIFF
--- a/scipy/interpolate/_bsplines.py
+++ b/scipy/interpolate/_bsplines.py
@@ -2271,9 +2271,9 @@ def make_smoothing_spline(x, y, w=None, lam=None, *, axis=0):
 
     >>> import matplotlib.pyplot as plt
     >>> grid = np.linspace(x[0], x[-1], 400)
+    >>> plt.plot(x, y, '.')
     >>> plt.plot(grid, spl(grid), label='Spline')
     >>> plt.plot(grid, func(grid), label='Original function')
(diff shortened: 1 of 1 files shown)


j-bowhay (approved):
Thanks @derb12, keep them coming!

PR #13193, opened 2020-12-02
MAINT: stats: chisquare check sum of observed/expected frequencies

mdhaber:
#### Reference issue
Closes gh-12282

#### What does this implement/fix?
There was no check in `scipy.stats.chisquare` that the sum of observed frequencies matched the sum of expected frequencies. This PR raises an error if the two do not agree within a certain tolerance. 

#### Additional information
The side effect is that it is no longer possible to express the expected frequencies in terms of probabilities; we can add automatic normalization of the observed frequencies for this case if desired.
A question is whether this should be specific to `chisquare` or is more generally applicable in `power_divergence`. Currently the check is performed in `power_divergence` whenever `lambda_ == 1` (Pearson).

@sam-data-guy would this address the concern?

PR #22695, opened 2025-03-15
MAINT: improve dtype handling now that `xp.result_type` accepts Python scalars

mdhaber:
#### Reference issue
gh-22683

Before `xp.result_type` accepted Python scalars, it was common to use `xp.asarray` before passing in the arguments. This does not always respect the standard array-scalar promotion rules. For example, if `a` is `float32` and `b` is a Python float, the result dtype *should* be `float32`, but the result *would* be `float64` for backends with `float64` default. Also, a standard technique for dealing with integer inputs in functions that expect floats was to convert these manually to the default floating point type. This happens to follow promotion rules in the most common cases, but ideally we should let the backends be responsible for unsafe dtype conversions like this. 

This PR improves the `xp_broadcast_promote` helper, factors out an improved `xp_result_type` function, and replaces instances of the old patterns with uses of these functions as soon after `array_namespace` as possible.

For now, I'm not adding additional tests; just passing the old ones. Let's consider this PR as a refactoring for simplicity and consistency. If someone is inclined to add tests with Python scalars later, great, but I think it's clear that this PR does some good as-is.


scipy/special/tests/test_logsumexp.py
```diff
@@ -415,3 +415,19 @@ def test_log_softmax_3d(self, xp):
...
+@pytest.mark.parametrize('dtype_a', dtypes)
+@pytest.mark.parametrize('dtype_b', dtypes)
+def test_torch_bug(xp, dtype_a, dtype_b):
+    # This is very confusing. I can't produce the failures outside of tests.
+    # I want to see if it fails in CI and whether others can reproduce.
+    a = xp.asarray([2, 1], dtype=getattr(xp, dtype_a))
+    b = xp.asarray([1, -1], dtype=getattr(xp, dtype_b))
+    dtype_1 = xp.result_type(a, b, 1.0)
+    dtype_2 = xp.result_type(b, a, 1.0)
+    assert dtype_1 == dtype_2
```
mdhaber, line 433: Can anyone reproduce these failures outside of tests?

<details>

```python3
FAILED scipy/special/tests/test_logsumexp.py::test_torch_bug[torch-float32-int32] - assert torch.float32 == torch.float64
FAILED scipy/special/tests/test_logsumexp.py::test_torch_bug[torch-float32-int64] - assert torch.float32 == torch.float64
FAILED scipy/special/tests/test_logsumexp.py::test_torch_bug[torch-int32-float32] - assert torch.float64 == torch.float32
FAILED scipy/special/tests/test_logsumexp.py::test_torch_bug[torch-int32-complex64] - assert torch.complex128 == torch.complex64
FAILED scipy/special/tests/test_logsumexp.py::test_torch_bug[torch-int64-float32] - assert torch.float64 == torch.float32
FAILED scipy/special/tests/test_logsumexp.py::test_torch_bug[torch-int64-complex64] - assert torch.complex128 == torch.complex64
FAILED scipy/special/tests/test_logsumexp.py::test_torch_bug[torch-complex64-int32] - assert torch.complex64 == torch.complex128
FAILED scipy/special/tests/test_logsumexp.py::test_torch_bug[torch-complex64-int64] - assert torch.complex64 == torch.complex128
```

</details>

Here is some code to try:
```python3
from array_api_compat import torch as xp

dtypes = ['float32', 'float64', 'int32', 'int64', 'complex64', 'complex128']
for dtype_a in dtypes:
    for dtype_b in dtypes:
        a = xp.asarray([2, 1], dtype=getattr(xp, dtype_a))
        b = xp.asarray([1, -1], dtype=getattr(xp, dtype_b))
        dtype_1 = xp.result_type(a, b, 1.0)
        dtype_2 = xp.result_type(b, a, 1.0)
        assert dtype_1 == dtype_2
```

Any ideas?
@crusaderky ping because I think you know how `xp` is being provided to tests.

ev-br: This seems dependent on the default fp dtype:

```
In [1]: import torch

In [2]: torch.__version__
Out[2]: '2.6.0+cpu'

In [3]: from array_api_compat import torch as xp
   ...: 
   ...: dtypes = ['float32', 'float64', 'int32', 'int64', 'complex64', 'complex128']
   ...: for dtype_a in dtypes:
   ...:     for dtype_b in dtypes:
   ...:         a = xp.asarray([2, 1], dtype=getattr(xp, dtype_a))
   ...:         b = xp.asarray([1, -1], dtype=getattr(xp, dtype_b))
   ...:         dtype_1 = xp.result_type(a, b, 1.0)
   ...:         dtype_2 = xp.result_type(b, a, 1.0)
   ...:         assert dtype_1 == dtype_2
   ...: 

In [7]: torch.set_default_dtype(torch.float64)

In [8]: from array_api_compat import torch as xp
   ...: 
   ...: dtypes = ['float32', 'float64', 'int32', 'int64', 'complex64', 'complex128']
   ...: for dtype_a in dtypes:
   ...:     for dtype_b in dtypes:
   ...:         a = xp.asarray([2, 1], dtype=getattr(xp, dtype_a))
   ...:         b = xp.asarray([1, -1], dtype=getattr(xp, dtype_b))
   ...:         dtype_1 = xp.result_type(a, b, 1.0)
   ...:         dtype_2 = xp.result_type(b, a, 1.0)
   ...:         assert dtype_1 == dtype_2
   ...: 
AssertionError                            Traceback (most recent call last)
Cell In[8], line 10
      8 dtype_1 = xp.result_type(a, b, 1.0)
      9 dtype_2 = xp.result_type(b, a, 1.0)
---> 10 assert dtype_1 == dtype_2

AssertionError:

```

mdhaber: Thanks! Do you agree that it shouldn't be? Can we patch that in array-api-compat?

ev-br: Yes indeed, `result_type` scalars on `torch` needs fixing:

```
In [23]: xp.result_type(xp.float32, 1)
Cell In[23], line 1
----> 1 xp.result_type(xp.float32, 1)

File ~/repos/array-api-compat/array_api_compat/torch/_aliases.py:139, in result_type(*arrays_and_dtypes)
    137 x, y = arrays_and_dtypes
    138 if isinstance(x, _py_scalars) or isinstance(y, _py_scalars):
--> 139     return torch.result_type(x, y)
    141 xdt = x.dtype if not isinstance(x, torch.dtype) else x
    142 ydt = y.dtype if not isinstance(y, torch.dtype) else y

TypeError: result_type() received an invalid combination of arguments - got (torch.dtype, int), but expected one of:
 * (Tensor tensor, Tensor other)
      didn't match because some of the arguments have invalid types: (torch.dtype, int)
 * (Number scalar, Tensor tensor)
      didn't match because some of the arguments have invalid types: (torch.dtype, int)
 * (Tensor tensor, Number other)
      didn't match because some of the arguments have invalid types: (torch.dtype, int)
 * (Number scalar1, Number scalar2)
      didn't match because some of the arguments have invalid types: (torch.dtype, int)
```

mdhaber: Right, I knew about that one, too; it wasn't as surprising to me as the other. I'll report at array-api-compat.

scipy/_lib/_array_api.py
```diff
@@ -513,31 +514,50 @@ def xp_ravel(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
-    try:  # follow library's prefered mixed promotion rules
-        dtype = xp.result_type(*dtypes)
-        if force_floating and xp.isdtype(dtype, 'integral'):
-            # If we were to add `default_float` before checking whether the result
-            # type is otherwise integral, we risk promotion from lower float.
-            dtype = xp.result_type(dtype, default_float)
+    try:  # follow library's preferred promotion rules
+        dtype = (xp.result_type(*args_not_none, 1.0) if force_floating
+                 else xp.result_type(*args_not_none))
     except TypeError:  # mixed type promotion isn't defined
```
mdhaber, line 529: Another benefit of this PR is that it gets most integer->float conversion handled by the same code path. For backends that don't support this, we currently define our own rule: use the result type of the default float and any other floats passed. 

Why define *a* rule? Without it, many existing tests would fail with `array_api_strict`. With some effort, one could ensure that all backends get the required coverage with floating intput and separately check that NumPy (and possibly other backends) continues to accept accepts integers as it always has, but this can substantially increase the diff in array API conversion PRs, and moreover, it is a fair amount of work for a backend that is not even meant to be used. (IIUC, the only real backend that doesn't support *automatic* integer to float promotion is TensorFlow, but even it supports `result_type` on the two.) So it is easiest to just define a rule. 

If someone feels strongly that we need to protect users from *these* unsafe conversions (despite the fact that we don't protect them from all sorts of other gotchas associated with floating point arithmetic, such as promotion from Python `int`s to array floating point types, any sort of overflow, etc.), that's fine, and they are welcome to do that work. This PR makes it easier for them to do so because all the logic is in one place.

Why use *this* rule? This is essentially the logic already implemented in most places. It is probably not ideal because it can result in the same sort of over-promotion we're trying to avoid. However, it is of little practical concern right now, since it only affects Array API strict, so it can be debated and refined later.

scipy/optimize/tests/test_bracket.py
```diff
@@ -249,13 +250,13 @@ def test_input_validation(self, xp):
         with pytest.raises(ValueError, match=message):
             _bracket_root(lambda x: x, -4+1j, 4)
         with pytest.raises(ValueError, match=message):
-            _bracket_root(lambda x: x, -4, 'hello')
```
mdhaber, line 252: It was a bit overkill to expect the same message for extremely odd input like this. Now we call `array_namespace` on all arguments like we should, so when `SCIPY_ARRAY_API=1`, we get the standard message that only numeric input is accepted. When `SCIPY_ARRAY_API` is not set, we get the same message as we always have, but it's annoying to change the test depending on the value of `SCIPY_ARRAY_API`, so I just test for complex input now.

scipy/special/tests/test_support_alternative_backends.py
```diff
@@ -4,13 +4,14 @@
...
 from scipy._lib._array_api_no_0d import xp_assert_close
-from scipy._lib._array_api import (is_cupy, is_dask, is_jax, is_torch,
+from scipy._lib._array_api import (is_cupy, is_dask, is_jax, is_torch, array_namespace,
                                    is_array_api_strict, SCIPY_DEVICE)
 from scipy._lib.array_api_compat import numpy as np
 
 
 def test_dispatch_to_unrecognized_library():
     xp = pytest.importorskip("array_api_strict")
+    xp = array_namespace(xp.asarray(1.))  # xp must match what array_namespace returns
```
mdhaber, line 14: With `SCIPY_ARRAY_API` not set, `array_namespace` always returns the NumPy namespace. This was causing problems and causing the test to fail with `SCIPY_ARRAY_API` not set. Maybe a better solution is to skip this test when `SCIPY_ARRAY_API` is not set, but this was hard enough to debug, so I'd rather not dig into it again. When `SCIPY_ARRAY_API=1` becomes the default and only option, the problem will just disappear.

scipy/stats/_stats_py.py
```diff
@@ -1121,10 +1112,7 @@ def moment(a, order=1, axis=0, nan_policy='propagate', *, center=None):
     xp = array_namespace(a)
     a, axis = _chk_asarray(a, axis, xp=xp)
 
-    if xp.isdtype(a.dtype, 'integral'):
-        a = xp.asarray(a, dtype=xp.float64)
-    else:
-        a = xp.asarray(a)
+    a = xp_broadcast_promote(a, force_floating=True, xp=xp)
```
crusaderky, line 1115: This PR downgrades the dtype from float64 to float32 when default_dtype is float32.

scipy/stats/_stats_py.py
```diff
@@ -6758,11 +6743,9 @@ def ttest_ind(a, b, *, axis=0, equal_var=True, nan_policy='propagate',
...
     xp = array_namespace(a, b)
 
-    default_float = xp.asarray(1.).dtype
-    if xp.isdtype(a.dtype, 'integral'):
-        a = xp.astype(a, default_float)
-    if xp.isdtype(b.dtype, 'integral'):
-        b = xp.astype(b, default_float)
+    dtype = xp_result_type(a, b, force_floating=True, xp=xp)
+    a = xp.asarray(a, dtype=dtype)
+    b = xp.asarray(b, dtype=dtype)
```
crusaderky, line 6748: FWIW this looks cleaner and more readable to me than
```python
a, b = xp_broadcast_promote(a, b, force_floating=True, broadcast=False, xp=xp)
```
which makes a case for deleting the broadcast=True option from the signature of `xp_broadcast_promote`.

scipy/stats/tests/test_continued_fraction.py
```diff
@@ -92,6 +92,7 @@ def test_input_validation(self, xp):
         with pytest.raises(ValueError, match=message):
             _continued_fraction(a1, b1, log=2)
 
+    @pytest.mark.skip_xp_backends('torch', reason='data-apis/array-api-compat#273')
```
crusaderky, line 95:
```suggestion
    @pytest.mark.xfail_xp_backends('torch', reason='data-apis/array-api-compat#273')
```

scipy/stats/tests/test_morestats.py
```diff
@@ -1995,6 +1995,7 @@ def test_brack(self):
 
 class TestBoxcox_llf:
 
+    @pytest.mark.skip_xp_backends('torch', reason='data-apis/array-api-compat#274')
```
crusaderky, line 1998:
```suggestion
    @pytest.mark.xfail_xp_backends('torch', reason='data-apis/array-api-compat#274')
```

scipy/stats/tests/test_morestats.py
```diff
@@ -2026,6 +2027,7 @@ def test_2d_input(self, xp):
     def test_empty(self, xp):
         assert xp.isnan(xp.asarray(stats.boxcox_llf(1, xp.asarray([]))))
 
+    @pytest.mark.skip_xp_backends('torch', reason='data-apis/array-api-compat#274')
```
crusaderky, line 2030:
```suggestion
    @pytest.mark.xfail_xp_backends('torch', reason='data-apis/array-api-compat#274')
```

scipy/stats/tests/test_morestats.py
```diff
@@ -2034,6 +2036,7 @@ def test_gh_6873(self, xp):
         # The expected value was computed with mpmath.
         xp_assert_close(llf, xp.asarray(-17.93934208579061))
 
+    @pytest.mark.skip_xp_backends('torch', reason='data-apis/array-api-compat#274')
```
crusaderky, line 2039:
```suggestion
    @pytest.mark.xfail_xp_backends('torch', reason='data-apis/array-api-compat#274')
```

scipy/stats/tests/test_stats.py
```diff
@@ -430,6 +430,7 @@ def test_constant_input(self, xp):
             xp_assert_close(r, xp.asarray(xp.nan))
             xp_assert_close(p, xp.asarray(xp.nan))
 
+    @pytest.mark.skip_xp_backends('torch', reason='data-apis/array-api-compat#273')
```
crusaderky, line 433:
```suggestion
    @pytest.mark.xfail_xp_backends('torch', reason='data-apis/array-api-compat#273')
```

crusaderky (commented):
Overall this looks like a very nice cleanup.
However, I have concerns for when the default_dtype is float32.

scipy/_lib/_array_api.py
```diff
@@ -513,31 +514,50 @@ def xp_ravel(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
-            dtype = xp.result_type(*float_dtypes, default_float)
+        default_float = xp_default_dtype(xp)
+        float_args = []
+        for arg in args_not_none:
+            arg = xp.asarray(arg) if np.isscalar(arg) else arg
+            dtype = getattr(arg, 'dtype', arg)
+            if not xp.isdtype(dtype, 'integral'):
+                float_args.append(arg)
+        if float_args:
+            dtype = xp.result_type(*float_args, default_float)
```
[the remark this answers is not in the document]
mdhaber: This doesn't run for JAX, right? Only `array-api-strict`. Did you see https://github.com/scipy/scipy/pull/22695#discussion_r1997905891?

scipy/_lib/_array_api.py
```diff
@@ -513,31 +514,50 @@ def xp_ravel(x: Array, /, *, xp: ModuleType | None = None) -> Array:
     return xp.reshape(x, (-1,))
 
 
-# utility to broadcast arrays and promote to common dtype
-def xp_broadcast_promote(*args, ensure_writeable=False, force_floating=False, xp=None):
-    xp = array_namespace(*args) if xp is None else xp
-
-    args = [(_asarray(arg, subok=True) if arg is not None else arg) for arg in args]
+# utility to find common dtype with option to force floating
+def xp_result_type(*args, xp, force_floating=False):
```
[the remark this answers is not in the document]
mdhaber: Ok, probably deserves more than the code comment.

scipy/optimize/tests/test_bracket.py
```diff
@@ -219,6 +219,7 @@ def f(xs, js):
 
         xp_assert_equal(res.status, ref_flags)
 
+    @pytest.mark.skip_xp_backends('torch', reason='data-apis/array-api-compat#273')
```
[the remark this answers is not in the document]
mdhaber: Didn't even know we had that now.

scipy/special/tests/test_support_alternative_backends.py
```diff
@@ -21,6 +22,7 @@ def test_dispatch_to_unrecognized_library():
 @pytest.mark.parametrize('dtype', ['float32', 'float64', 'int64'])
 def test_rel_entr_generic(dtype):
     xp = pytest.importorskip("array_api_strict")
+    xp = array_namespace(xp.asarray(1.))  # xp must match what array_namespace returns
```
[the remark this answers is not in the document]
mdhaber: Somehow I just didn't think about that until I was self-reviewing.

scipy/stats/_stats_py.py
mdhaber: Huh I think that's even why I added the `broadcast` argument, but I forgot to use it.

I drill think it's convenient to have a function that can *just* promote a bunch of arga.

How about calling it `xp_promote` leaving the `broadcast` argument with a default of `False` so `broadcast=True` is always explicit (but not used here).

scipy/_lib/_array_api.py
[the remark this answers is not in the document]
crusaderky: It doesn't. However I'd like not to leave a timebomb for the next backend that defaults to float32 :wink: 
In fact, it makes a strong case for adding a default_dtype config setting to array-api-strict.

scipy/stats/_stats_py.py
crusaderky: sounds good to me

crusaderky: Alternatively, why not just skip the `broadcast=False` parameter and write `xp.broadcast_arrays(xp_promote(*arrays))` when needed?

scipy/_lib/_array_api.py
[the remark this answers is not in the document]
mdhaber: Is there any backend you know of that doesn't define int to float promotion in `result_type`?

scipy/stats/_stats_py.py
mdhaber: Yup, but only for array API strict. And it's temporary until we discuss what we _want_ to have happen when `SCIPY_ARRAY_API=1` becomes the default and only option (https://github.com/scipy/scipy/pull/22695#discussion_r1997905891).

scipy/_lib/_array_api.py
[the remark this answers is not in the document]
mdhaber: I did the experiment:
```python3
from array_api_compat import numpy, cupy, torch
from array_api_compat.dask import array as dask
import tensorflow.experimental.numpy as tensorflow
import jax.numpy as jax

for xp in [numpy, cupy, torch, dask, jax, tensorflow]:
    dtypes = [xp.int8, xp.int16, xp.int32, xp.int64,
              xp.uint8, xp.uint16, xp.uint32, xp.uint64,
              xp.float32, xp.float64, xp.complex64, xp.complex128]
    for dtype_a in dtypes:
        for dtype_b in dtypes:
            try:
                xp.result_type(dtype_a, dtype_b)
            except:
                print(f"{xp}: {dtype_a}, {dtype_b}")
```
<details>

```
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.int8, torch.uint16
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.int8, torch.uint32
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.int8, torch.uint64
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.int16, torch.uint16
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.int16, torch.uint32
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.int16, torch.uint64
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.int32, torch.uint16
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.int32, torch.uint32
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.int32, torch.uint64
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.int64, torch.uint16
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.int64, torch.uint32
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.int64, torch.uint64
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint8, torch.uint16
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint8, torch.uint32
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint8, torch.uint64
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint16, torch.int8
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint16, torch.int16
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint16, torch.int32
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint16, torch.int64
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint16, torch.uint8
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint16, torch.uint32
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint16, torch.uint64
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint16, torch.complex64
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint16, torch.complex128
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint32, torch.int8
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint32, torch.int16
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint32, torch.int32
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint32, torch.int64
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint32, torch.uint8
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint32, torch.uint16
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint32, torch.uint64
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint32, torch.complex64
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint32, torch.complex128
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint64, torch.int8
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint64, torch.int16
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint64, torch.int32
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint64, torch.int64
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint64, torch.uint8
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint64, torch.uint16
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint64, torch.uint32
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint64, torch.complex64
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.uint64, torch.complex128
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.complex64, torch.uint16
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.complex64, torch.uint32
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.complex64, torch.uint64
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.complex128, torch.uint16
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.complex128, torch.uint32
<module 'array_api_compat.torch' from '/usr/local/lib/python3.11/dist-packages/array_api_compat/torch/__init__.py'>: torch.complex128, torch.uint64
```

</details>

The only output is from `torch` which doesn't define some `int`/`uint` result types (some of which should be defined) or integral/complex result types. But we know `torch.result_type` is not production-ready (gh-273, gh-274), so I think we need it to settle before we make decisions based on it.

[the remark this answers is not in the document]
lucascolley: > In fact, it makes a strong case for adding a default_dtype config setting to array-api-strict.

previously suggested in https://github.com/data-apis/array-api-strict/issues/38, and see also https://github.com/data-apis/array-api-strict/issues/64

ev-br:
PSA: https://github.com/data-apis/array-api-tests/pull/348 and https://github.com/data-apis/array-api-compat/pull/277 seem to fix the issue with `result_type`. Please test. 
Once these two land (maybe also https://github.com/data-apis/array-api-compat/pull/272 and https://github.com/data-apis/array-api-tests/pull/348), an `array-api-compat==1.11.2` will be in order.

mdhaber (commented):
Thanks for the review @crusaderky. I think I addressed everything except for the `test_support_alternative_backends` comments - is there a decorator for skipping when `SCIPY_ARRAY_API` is not defined, or should I manually check inside the test?

scipy/special/tests/test_support_alternative_backends.py
lucascolley: `skip_xp_invalid_arg` is functionally equivalent I think (we might want a new name/alias)

mdhaber: I think this is the opposite of that. We don't want to run the test *unless* `SCIPY_ARRAY_API=1`

lucascolley: Right you are!

mdhaber:
Looks like I have a few more `xfail`s to add in `test_bracket`. I'll do that when https://github.com/scipy/scipy/pull/22695#discussion_r1999494400 is resolved. If there is no decorator, I'll use:
```python3
if not int(os.environ.get('SCIPY_ARRAY_API', 0)):
    pytest.skip("Only appropriate to run for non-NumPy backend.")
```

crusaderky:
> Thanks for the review @crusaderky. I think I addressed everything except for the `test_support_alternative_backends` comments - is there a decorator for skipping when `SCIPY_ARRAY_API` is not defined, or should I manually check inside the test?

`from scipy.conftest import skip_xp_invalid_arg`

crusaderky:
> Looks like I have a few more `xfail`s to add in `test_bracket`. I'll do that when [#22695 (comment)](https://github.com/scipy/scipy/pull/22695#discussion_r1999494400) is resolved. If there is no decorator, I'll use:
> 
> ```python
> if not int(os.environ.get('SCIPY_ARRAY_API', 0)):
>     pytest.skip("Only appropriate to run for non-NumPy backend.")
> ```

```python
from scipy._lib._array_api import SCIPY_ARRAY_API

@pytest.mark.skipif(SCIPY_ARRAY_API, reason="...")
def test(...):
    ...
```

crusaderky (changes requested):

scipy/stats/_quantile.py
```diff
@@ -1,24 +1,24 @@
...
-        x = xp.astype(x, xp_default_dtype(xp))
 
-    if not xp.isdtype(p.dtype, 'real floating'):
+    if not xp.isdtype(xp.asarray(p).dtype, 'real floating'):
         raise ValueError("`p` must have real floating dtype.")
 
+    dtype = xp_result_type(x, p, xp=xp)
+    x = xp.asarray(x, dtype=dtype)
+    p = xp.asarray(p, dtype=dtype)
+
```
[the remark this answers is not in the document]
mdhaber: Looks like I may have written it this way before taking broadcasting out of `xp_broadcast_promote`.

scipy/stats/_stats_py.py
```diff
@@ -7368,16 +7347,14 @@ def _power_divergence(f_obs, f_exp, ddof, axis, lambda_, sum_check=True):
...
     f_obs = xp.asarray(f_obs, dtype=dtype)
-    f_obs_float = xp.asarray(f_obs, dtype=xp.float64)
 
     if f_exp is not None:
-        f_exp = xp.asarray(f_exp)
-        dtype = default_float if xp.isdtype(f_exp.dtype, 'integral') else f_exp.dtype
         f_exp = xp.asarray(f_exp, dtype=dtype)
 
+        # not sure why we force to float64, but not going to touch it
+        f_obs_float = xp.asarray(f_obs, dtype=xp.float64)
```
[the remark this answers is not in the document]
mdhaber: I don't really understand the question. This is used only for the `sum_check`, so it doesn't affect the value returned to the user. This line was added in gh-13193 when stats paid no attention at all to dtypes (https://github.com/scipy/scipy/pull/13193#discussion_r534503188), and I chose not to stop to revisit the choice at the moment. I'm making some targeted changes here, and I'd prefer not to fix every oddity I run across along the way.

Which library-wide pattern? If you mean dtype preservation, that is something I've had to fight for recently, when others have argued either that it's not important or we should actually force stats calculations to be done in float64. And pre NEP 50, we had to fight with NumPy to keep dtypes consistent. So I'm all on board for the consistency; I would just rather not change this line right now.

scipy/stats/tests/test_continued_fraction.py
```diff
@@ -92,6 +92,7 @@ def test_input_validation(self, xp):
         with pytest.raises(ValueError, match=message):
             _continued_fraction(a1, b1, log=2)
 
+    @pytest.mark.xfail_xp_backends('torch', reason='TBD')
```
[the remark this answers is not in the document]
mdhaber: Investigated. Dramatically simplified the dtype/shape handling using the new `xp_promote`.

scipy/stats/_stats_py.py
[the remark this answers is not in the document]
crusaderky: I'm referring to the library-wide pattern, that you're making uniform in this very PR, to promote int to default float instead of float64.
Happy to leave this nitpicking to a later PR.

scipy/stats/_continued_fraction.py
```diff
@@ -265,7 +272,7 @@ def _continued_fraction(a, b, *, args=(), tolerances=None, maxiter=100, log=Fals
     """
 
     res = _continued_fraction_iv(a, b, args, tolerances, maxiter, log)
-    a, b, args, eps, tiny, maxiter, log = res
+    a, b, args, eps, tiny, maxiter, log, a0, b0, shape, dtype, xp = res
```
crusaderky, line 275: Code flow: does it even make sense to keep it as a separate function?

scipy/stats/_continued_fraction.py
```diff
@@ -287,36 +293,7 @@ def b(n, *args, b=b):
...
-    xp = array_namespace(fs_a[0], fs_b[0], *args)
-
-    shape = np.broadcast_shapes(shape_a, shape_b)  # OK to use NumPy on tuples
-    dtype = xp.result_type(dtype_a, dtype_b)
-    an = xp.astype(xp_ravel(xp.broadcast_to(xp.reshape(fs_a[0], shape_a), shape)), dtype)  # noqa: E501
-    bn = xp.astype(xp_ravel(xp.broadcast_to(xp.reshape(fs_b[0], shape_b), shape)), dtype)  # noqa: E501
-    args = [xp.astype(xp_ravel(xp.broadcast_to(arg, shape)), dtype) for arg in args]
-
-    status = xp.full_like(an, xp.asarray(eim._EINPROGRESS),
-                          dtype=xp.int32)  # in progress
```
crusaderky, line 319: I confess I feel out of my depth here and I have no idea what this PR functionally changes in this module, if anything.

scipy/stats/_quantile.py
```diff
@@ -1,24 +1,23 @@
...
         raise ValueError("`x` must have real dtype.")
-    if xp.isdtype(x.dtype, 'integral'):
-        x = xp.astype(x, xp_default_dtype(xp))
 
-    if not xp.isdtype(p.dtype, 'real floating'):
+    if not xp.isdtype(xp.asarray(p).dtype, 'real floating'):
         raise ValueError("`p` must have real floating dtype.")
 
+    x, p = xp_promote(x, p, force_floating=True, xp=xp)
+    dtype = x.dtype
```
crusaderky, line 19:
```suggestion
```
unused

scipy/stats/_quantile.py
```diff
@@ -1,24 +1,23 @@
...
+    if not xp.isdtype(xp.asarray(x).dtype, ('integral', 'real floating')):
         raise ValueError("`x` must have real dtype.")
-    if xp.isdtype(x.dtype, 'integral'):
-        x = xp.astype(x, xp_default_dtype(xp))
 
-    if not xp.isdtype(p.dtype, 'real floating'):
+    if not xp.isdtype(xp.asarray(p).dtype, 'real floating'):
         raise ValueError("`p` must have real floating dtype.")
 
+    x, p = xp_promote(x, p, force_floating=True, xp=xp)
```
crusaderky, line 18: The comment about this check being very expensive for lists still stands.
I think it should be removed outright.
With the check in, `force_floating=True` is redundant.

scipy/optimize/_bracket.py
```diff
@@ -450,10 +441,12 @@ def _bracket_minimum_iv(func, xm0, xl0, xr0, xmin, xmax, factor, args, maxiter):
...
-    xl0, xm0, xr0, xmin, xmax, factor = xp.broadcast_arrays(
-        xp.asarray(xl0), xm0, xp.asarray(xr0), xp.asarray(xmin),
-        xp.asarray(xmax), xp.asarray(factor)
-    )
+
+    xm0, xl0, xr0, xmin, xmax, factor = xp_promote(
+        xm0, xl0, xr0, xmin, xmax, factor, broadcast=True, force_floating=True, xp=xp)
+
+    if not xp.isdtype(xm0.dtype, ('integral', 'real floating')):
+        raise ValueError('`xm0` must be numeric and real.')
```
crusaderky, line 449: Could you update the checks on lines 451:468? Probably a good idea to make it into a for loop.

scipy/_lib/_array_api.py
```diff
@@ -513,31 +514,68 @@ def xp_ravel(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+    (see Array API Standard Type Promotion Rules) to the arguments. Augments
+    standard `result_type` in two ways:
+
+    - There is a `force_floating` argument that ensures that the result type
+      is floating point, even when all args are integer.     
+    - When a TypeError is raised (e.g. due to an unsupported promotion),
+      we define a custom rule: use the result type of the default float
+      and any other floats passed. See
+      https://github.com/scipy/scipy/pull/22695/files#r1997905891
+      for rationale.
```
crusaderky, line 530: 1. both xp_result_type and xp_promote accept plain lists, unlike
   xp.result_type. This needs to be documented.

2. This behaviour can be a bit surprising. It makes sense to me,
   but it needs to be documented in both xp_result_type and xp_promote.
```python
>>> import numpy as np
>>> from scipy._lib._array_api import xp_result_type
>>> xp_result_type(1.0, np.asarray(1.0, dtype=np.float32))
dtype('float32')
>>> xp_result_type([1.0], np.asarray(1.0, dtype=np.float32))
dtype('float64')
```

3. torch is broken unless you explicitly pass xp=xp:
```python
>>> import scipy._lib.array_api_compat.torch as xp
>>> xp.result_type(1.0, xp.asarray(1.0, dtype=xp.float32))
torch.float32
>>> xp_result_type(1.0, xp.asarray(1.0, dtype=xp.float32), xp=xp)
torch.float32
>>> xp_result_type(1.0, xp.asarray(1.0, dtype=xp.float32))
TypeError: dtype argument must be a NumPy dtype, but it is a <class 'torch.dtype'>.
```

4. Here, torch is broken in a different way (`xp=xp` makes the error message change):
```python
>>> import scipy._lib.array_api_compat.torch as xp
>>> xp_result_type([1.0], xp.asarray(1.0, dtype=xp.float32))
TypeError: dtype argument must be a NumPy dtype, but it is a <class 'torch.dtype'>.
>>> xp_result_type([1.0], xp.asarray(1.0, dtype=xp.float32), xp=xp)
AttributeError: 'numpy.dtypes.Float64DType' object has no attribute 'is_floating_point'
```

Could you please write a dedicated unit test so that we have confidence that all these cases are covered?

crusaderky (changes requested):

scipy/stats/_continued_fraction.py
```diff
@@ -29,6 +29,13 @@ def _continued_fraction_iv(a, b, args, tolerances, maxiter, log):
     if not np.iterable(args):
         args = (args,)
 
+    # Call each callable once to determine namespace and dtypes
+    a0, b0 = a(0, *args), b(0, *args)
```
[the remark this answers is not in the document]
mdhaber: This is fine.

scipy/stats/_continued_fraction.py
```diff
@@ -29,6 +29,13 @@ def _continued_fraction_iv(a, b, args, tolerances, maxiter, log):
...
+    xp = array_namespace(a0, b0, *args)
+    a0, b0, *args = xp_promote(a0, b0, *args, force_floating=True, broadcast=True)
```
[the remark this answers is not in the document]
mdhaber: This is one of a class of functions that expects that once a common dtype is found, it must be stable such that when passed into the function, the function returns the same type.

This particular function may not actually need that,  but I am treating it the same as all the rest that do. It doesn't need to be simplified further here.

scipy/stats/_continued_fraction.py
mdhaber: No problem.

scipy/stats/_quantile.py
mdhaber: I just committed your last comment about this line. https://github.com/scipy/scipy/pull/22695#discussion_r2014120797

scipy/optimize/_bracket.py
mdhaber: Not here.

scipy/_lib/array_api_compat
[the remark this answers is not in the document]
mdhaber: See commit a251df9 and associated CI run. I don't think that was the tag you were asking for, but it is a later commit, and all array API tests failed.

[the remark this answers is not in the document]
lucascolley: the relevant part of the CI log is

```
scipy/_lib/array_api_compat/torch/_aliases.py:12: in <module>
    from ._typing import Array, Device, DType
E   ModuleNotFoundError: No module named 'scipy._lib.array_api_compat.torch._typing'
```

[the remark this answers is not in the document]
lucascolley: probably just need to add the module to https://github.com/scipy/scipy/blob/5c5b44ceb45fe51ba0813d2b4bd295bcb45007e6/scipy/_lib/meson.build#L209-L218

scipy/_lib/_array_api.py
```diff
@@ -513,31 +514,68 @@ def xp_ravel(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+    Promotes elements of *args to result dtype, ignoring `None`s.
+    Includes options for forcing promotion to floating point and
+    broadcasting the arrays, again ignoring `None`s.
+    """
+    xp = array_namespace(*args) if xp is None else xp
+
+    dtype = xp_result_type(*args, force_floating=force_floating, xp=xp)
+
+    args = [(_asarray(arg, dtype=dtype, subok=True, xp=xp) if arg is not None else arg)
+            for arg in args]
```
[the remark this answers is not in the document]
mdhaber: I considered this when I was writing it. In the future, it's looking like lists aren't even going to be allowed as array-like arguments into xp-compatible functions, so I didn't bother. I'll adjust is as an easy optimization, but I don't think it needs to be carefully profiled and optimized. If someone reports it as a bottleneck, we can take another look. 

I'm not concerned about the overflow issue. It signals to the user that they should have been passing an array with suitable dtype in the first place. If the value is `2**31-1`, addition is going to cause overflow anyway, so I don't think we need to do the pre-NEP 50 things and make type decisions based on numerical values.

scipy/_lib/_array_api.py
mdhaber: Oops, I guess I forgot to make `xp` required. I noticed that and started always passing `xp`.

I did not expect this private function to be robust to out-of-context use. It doesn't need to work with combinations of lists and non-NumPy arrays because `array_namespace` will be called on the same arguments first. It only needs to do what we ask it to do, and we are not asking it to do that.

I can document some of this, sure, and I can write a basic unit test.

scipy/_lib/array_api_compat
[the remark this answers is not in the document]
crusaderky: main now points to array-api-compat 1.11.2 and array-api-extra  0.7.1; you shouldn't need any further changes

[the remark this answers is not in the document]
mdhaber: Looks like this has already been done.

mdhaber:
Documenting this here so I don't lose it:
```
pytest scipy/differentiate/tests scipy/integrate/tests/test_tanhsinh.py scipy/integrate/tests/test_cubature.py scipy/integrate/tests/test_quadrature.py scipy/optimize/tests/test_bracket.py scipy/optimize/tests/test_chandrupatla.py scipy/special/tests/test_logsumexp.py scipy/special/tests/test_support_alternative_backends.py scipy/stats/tests/test_continued_fraction.py scipy/stats/tests/test_entropy.py scipy/stats/tests/test_morestats.py scipy/stats/tests/test_quantile.py scipy/stats/tests/test_stats.py scipy/stats/tests/test_resampling.py
```

scipy/_lib/_array_api.py
mdhaber: Added tests that the behavior of `xp_result_type` is the same as `xp.result_type`, including errors raised, for all combinations of:
- numeric Python scalars
- lists including Python scalars
- `xp` dtypes
- `xp` arrays of all dtypes

with the following exceptions:
- If an argument is iterable, it gets converted to an array with the provided `xp` first.
- When `force_floating=True`, 1.0 is appended to the argument list.
- When `force_floating=True` and a `TypeError` is raised, the result type is the same as `xp.result_type` called with only the  floating point arguments and the default float.

If this is not what you had in mind, feel free to add tests in a follow-up. I'd like to get this merged to avoid further merge conflicts.

scipy/_lib/_array_api.py
```diff
@@ -513,31 +514,85 @@ def xp_ravel(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+                float_args.append(arg)
+        return xp.result_type(*float_args, xp_default_dtype(xp))
+
+
+def xp_promote(*args, broadcast=False, ensure_writeable=False,
+               force_floating=False, xp):
+    """
+    Promotes elements of *args to result dtype, ignoring `None`s.
+    Includes options for forcing promotion to floating point and
+    broadcasting the arrays, again ignoring `None`s.
```
crusaderky, line 574:
```suggestion
    broadcasting the arrays, again ignoring `None`s.
    Type promotion rules follow `xp_result_type` instead of `xp.result_type`.
```

crusaderky (changes requested):
I think we are almost there!

scipy/_lib/_array_api.py
```diff
@@ -513,31 +514,85 @@ def xp_ravel(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+        # Follow NEP 50 promotion rules anyway
+        args_not_none = [arg.dtype if getattr(arg, 'size', 0) == 1 else arg
+                         for arg in args_not_none]
+        return xp.result_type(*args_not_none)
+
+    try:  # follow library's preferred promotion rules
+        return xp.result_type(*args_not_none)
+    except TypeError as e:  # mixed type promotion isn't defined
+        if not force_floating:
+            raise e
```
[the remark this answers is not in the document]
mdhaber: Recall from the top post that I didn't intend to get embroiled in a big discussion over what this policy should be. It just doesn't matter what happens with `array_api_strict` right now if it's not breaking tests. If a developer runs into an error here that they didn't expect, that's a good thing, because hopefully we'll get more opinions about this fine point. The plan is to an open an issue after this merges, and then others can tweak what happens for libraries that don't exist as much as they like. In the meantime, I'll update the documentation.

scipy/_lib/_array_api.py
```diff
@@ -513,31 +514,85 @@ def xp_ravel(x: Array, /, *, xp: ModuleType | None = None) -> Array:
...
+      is floating point, even when all args are integer.     
+    - When a TypeError is raised (e.g. due to an unsupported promotion),
+      we define a custom rule: use the result type of the default float
+      and any other floats passed. See
+      https://github.com/scipy/scipy/pull/22695/files#r1997905891
+      for rationale.
+    - To support use when `SCIPY_ARRAY_API=1` is not set, this function accepts
+      array-like iterables, which are immediately converted to NumPy arrays before
+      result type calculation. Consequently, the result dtype may be different
+      when an argument is `1.` vs `[1.]`.
```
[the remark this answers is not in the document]
mdhaber: I intentionally added `xp` to the call to `_asarray`. Of course, it doesn't affect anything because all incoming arrays will have the same namespace in any real use of this function, which is *private*,but I will update the docstring.

scipy/optimize/_bracket.py
```diff
@@ -33,8 +27,11 @@ def _bracket_root_iv(func, xl0, xr0, xmin, xmax, factor, args, maxiter):
     xmin = -xp.inf if xmin is None else xmin
     xmax = xp.inf if xmax is None else xmax
     factor = 2. if factor is None else factor
-    xl0, xr0, xmin, xmax, factor = xp.broadcast_arrays(
-        xl0, xp.asarray(xr0), xp.asarray(xmin), xp.asarray(xmax), xp.asarray(factor))
+    xl0, xr0, xmin, xmax, factor = xp_promote(
+        xl0, xr0, xmin, xmax, factor, broadcast=True, force_floating=True, xp=xp)
```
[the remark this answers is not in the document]
mdhaber: The change is deliberate. All these should be factored into the result dtype.

mdhaber:
Free-threaded failure seen in https://github.com/scipy/scipy/pull/22524#issuecomment-2761636908, too; unrelated.

crusaderky (approved):
LGTM, let's merge it!

lucascolley:
thanks both, this seems like a nice improvement!

I think it would be nice to update https://scipy.github.io/devdocs/dev/api-dev/array_api.html with some guidance on this, but I understand that there are other open PRs for which the same can probably be said, so that ought not hold us up here. It would be great to get those docs nice and up to date at the end of this round of overhauls, ideally to the point where adding support to a function is a 'sprint-able' task.

Issue #22659, opened 2025-03-09
BUG: spatial: `RigidTransform` does not support zero-length collections


scottshambaugh:
### Describe your issue.

Follow-on issue from https://github.com/scipy/scipy/issues/22589. Should follow a similar implementation and test scheme as https://github.com/scipy/scipy/pull/22643

Should fix before first release of RigidTransform in v1.16.0.

@nmayorov @hubernikus FYI

nmayorov:
I plan to resolve this issues soon.

nmayorov:
Resolved in #22696

PR #22696, opened 2025-03-15
MAINT: spatial: support empty case in `RigidTransform`

nmayorov:
#### Reference issue

#22659

It makes empty (zero length) transformations naturally supported by `RigidTransfrom` class.

#### Additional information

It follows the same approach to implementation (only necessary to remove checks against this case) and testing as #22643

-----

@hubernikus wellcome to review as well.

### DIFF
--- a/scipy/spatial/transform/tests/test_rigid_transform.py
+++ b/scipy/spatial/transform/tests/test_rigid_transform.py
@@ -948,3 +948,13 @@ def test_empty_transform_representation():
     assert tf.as_matrix().shape == (0, 4, 4)
     assert tf.as_exp_coords().shape == (0, 6)
     assert tf.as_dual_quat().shape == (0, 8)
+
+
+def test_empty_transform_application():
+    tf = RigidTransform.identity(0)
+
+    assert tf.apply(np.zeros((3,))).shape == (0, 3)
+    assert tf.apply(np.empty((0, 3))).shape == (0, 3)
+
+    with pytest.raises(ValueError, match="operands could not be broadcast together"):
+        tf.apply(np.zeros((2, 3)))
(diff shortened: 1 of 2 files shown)


nmayorov:
Not sure if it is necessary to classify it as a bug, I think it's rather some maintenance of a minor improvement.

hubernikus:
Looks good and complete, and consistent with the `Rotation` implementation and tests.

The only thing maybe somewhere single-check: 
`assert tf.single == False`

scipy/spatial/transform/tests/test_rigid_transform.py
```diff
@@ -910,3 +909,104 @@ def test_normalize_dual_quaternion():
...
+    assert len(RigidTransform.concatenate([tf_single, tf_empty])) == 1
+    assert len(RigidTransform.concatenate([tf_empty, tf_many])) == 2
+    assert len(RigidTransform.concatenate([tf_many, tf_empty])) == 2
+    assert len(RigidTransform.concatenate([tf_many, tf_empty, tf_single])) == 3
+
+
+def test_empty_transform_inv_and_pow():
+    tf = RigidTransform.identity(0)
+    assert len(tf.inv()) == 0
+    assert len(tf ** 0.5) == 0
```
scottshambaugh, line 994: Could you add the exponents of -1, 0, and +1 which follow shortcut code branches?

scipy/spatial/transform/tests/test_rigid_transform.py
```diff
@@ -750,6 +750,12 @@ def test_indexing():
     assert_allclose(tf_slice.as_matrix()[:, :3, :3], r[0:2].as_matrix(), atol=atol)
     assert_allclose(tf_slice.as_matrix()[:, :3, 3], t[0:2], atol=atol)
 
+    # Test boolean indexing
+    tf_masked = tf[[False, True]]
```
scottshambaugh, line 754: Can you add an all-False case as well?

nmayorov:
All done.

scottshambaugh (approved):
Looks good, thanks for putting this together!

nmayorov:
Failures look unrelated, I'll go ahead and merge.

PR #22711, opened 2025-03-18
DOC: stats: Extend documentation of random_correlation matrix

vrossum:
DOC: _multivariate.py
Extended documentation of random_correlation matrix.
- state conditions on eig input.
- describe properties of output

### DIFF
--- a/scipy/stats/_multivariate.py
+++ b/scipy/stats/_multivariate.py
@@ -3872,7 +3872,7 @@ class random_correlation_gen(multi_rv_generic):
     ----------
     eigs : 1d ndarray
         Eigenvalues of correlation matrix. All eigenvalues need to be non-negative and
-        need to sum up to the dimension.
+        need to sum to the number of eigenvalues.
     seed : {None, int, `numpy.random.Generator`, `numpy.random.RandomState`}, optional
         If `seed` is None (or `np.random`), the `numpy.random.RandomState`
         singleton is used.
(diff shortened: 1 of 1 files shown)


melissawm:
Hi @vrossum !

- If you are addressing an existing issue, please add the expression "Closes #<issue number>"
- If you can, please make sure you use our [commit message conventions] and, for this case, you can also use `[docs only]` at the end of your commit message to make sure we don't spend CI minutes unnecessarily. (see https://docs.scipy.org/doc/scipy/dev/contributor/continuous_integration.html#skipping for more information)

Thank you!

scipy/stats/_multivariate.py
```diff
@@ -3870,7 +3871,8 @@ class random_correlation_gen(multi_rv_generic):
     Parameters
     ----------
     eigs : 1d ndarray
-        Eigenvalues of correlation matrix
+        Eigenvalues of correlation matrix. All eigenvalues need to be non-negative and
+        need to sum up to the dimension.
```
j-bowhay, line 3875: Perhaps slightly clearer?
```suggestion
        need to sum to the number of eigenvalues.
```

mdhaber: Yes, I'd say that "dimension" is not correct (here or most places it's used in `_multivariate.py`). "dimensionality" is often what is meant, but "number of eigenvalues" is even more clear here.

j-bowhay (approved):
Thanks for your first contribution to SciPy, keep them coming!

PR #22713, opened 2025-03-19
DOC: linalg.solve: clarify symmetry requirement

aseyboldt:
The `assume_a` argument in `scipy.linalg.solve` allows to specify that a matrix is positive definite.
It then assumes that the matrix is symmetric as well. Some authors (for instance the English wikipedia) define "positive definite" such that it is always symmetric, but others (for instance the German wikipedia, but I think also many other English sources) do not.

I added "symmetric" to the description in table to avoid ambiguity.

I also changed the code so that it accepts `assume_a="symmetric positive definite"`.

### DIFF
--- a/scipy/linalg/_basic.py
+++ b/scipy/linalg/_basic.py
@@ -85,17 +85,17 @@ def solve(a, b, lower=False, overwrite_a=False,
     corresponding string to ``assume_a`` key chooses the dedicated solver.
     The available options are
 
-    ===================  ================================
-     diagonal             'diagonal'
-     tridiagonal          'tridiagonal'
-     banded               'banded'
-     upper triangular     'upper triangular'
-     lower triangular     'lower triangular'
-     symmetric            'symmetric' (or 'sym')
-     hermitian            'hermitian' (or 'her')
-     positive definite    'positive definite' (or 'pos')
-     general              'general' (or 'gen')
-    ===================  ================================
+    =============================  ================================
+     diagonal                       'diagonal'
+     tridiagonal                    'tridiagonal'
+     banded                         'banded'
+     upper triangular               'upper triangular'
+     lower triangular               'lower triangular'
+     symmetric                      'symmetric' (or 'sym')
+     hermitian                      'hermitian' (or 'her')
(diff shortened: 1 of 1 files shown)


ilayn:
I think the clarification is good but the option name is too long. It would be a hassle to type it. Unsymmetric pos def.ness is really a niche and does not come too often regardless what some tend to define so it should be safe.

j-bowhay:
Thanks for bringing this up. I agree with @ilayn, I would not add a new option but I think it is fine to document.

scipy/linalg/_basic.py
```diff
@@ -208,7 +208,8 @@ def solve(a, b, lower=False, overwrite_a=False,
 
     if assume_a not in {None, 'diagonal', 'tridiagonal', 'banded', 'lower triangular',
                         'upper triangular', 'symmetric', 'hermitian',
-                        'positive definite', 'general', 'sym', 'her', 'pos', 'gen'}:
+                        'positive definite', 'symmetric positive definite', 'general',
+                        'sym', 'her', 'pos', 'gen'}:
```
j-bowhay, line 212:
```suggestion
                        'positive definite', 'general',  'sym', 'her', 'pos', 'gen'}:
```

scipy/linalg/_basic.py
```diff
@@ -208,7 +208,7 @@ def solve(a, b, lower=False, overwrite_a=False,
 
     if assume_a not in {None, 'diagonal', 'tridiagonal', 'banded', 'lower triangular',
                         'upper triangular', 'symmetric', 'hermitian',
-                        'positive definite', 'general', 'sym', 'her', 'pos', 'gen'}:
+                        'positive definite', 'general',  'sym', 'her', 'pos', 'gen'}:
```
j-bowhay, line 211:
```suggestion
                        'positive definite', 'general', 'sym', 'her', 'pos', 'gen'}:
```

j-bowhay (approved):
I changed this to be a docs only change so we can go ahead and merge this. Thanks for your contribution @aseyboldt

aseyboldt:
Thanks!

PR #22716, opened 2025-03-19
DOC: optimize: Clarify use of `xtol` in 1D rootfinder docstrings

### LINKED ISSUE #18740: BUG: scipy.optimize.bisect gives incorrect results for very small roots.
### Describe your issue.

If the root that you want to approximate is smaller than 0.1, bisect will return results that will have anywhere from fewer digits that expected to no valid digits. 

### Reproducing Code Example

```python
# Consider the following code which uses the default values for xtol & rtol (b has been chosen to avoid having f(c) = 0).

_xtol = 2e-12
_rtol = 4 * np.finfo(float).eps
for k in [1e-1, 1e-2, 1e-3, 1e-4, 1e-10, 1e-30, 1e-50]:
    tmp = optimize.bisect(lambda x: x - 1.234567890123456*k, -10, 10+1/3,  xtol=_xtol,
                          rtol=_rtol, maxiter=2000, full_output=True)
    r_root = 1.234567890123456 * k
    print('\nTrue root   = %.15e xtol = %.4e  rtol = %6e' % (r_root ,_xtol, _rtol))
    print('SciPy: root = %.15e calls = %d iterations = %d flag = %s'
          % (tmp[0], tmp[1].function_calls, tmp[1].iterations,  tmp[1].flag))

(I inserted spaces to show where the results differ.)

True root   = 1.2345678901 23456e-01 xtol = 2.0000e-12  rtol = 8.881784e-16
SciPy: root = 1.2345678901 19453e-01 calls = 46 iterations = 44 flag = converged

True root   = 1.234567890 123456e-02 xtol = 2.0000e-12  rtol = 8.881784e-16
SciPy: root = 1.234567890 077304e-02 calls = 46 iterations = 44 flag = converged

True root   = 1.23456789 0123456e-03 xtol = 2.0000e-12  rtol = 8.881784e-16
SciPy: root = 1.23456789 1042795e-03 calls = 46 iterations = 44 flag = converged

True root   = 1.2345678 90123456e-04 xtol = 2.0000e-12  rtol = 8.881784e-16
SciPy: root = 1.2345678 89139551e-04 calls = 46 iterations = 44 flag = converged

True root   = 1.23 4567890123456e-10 xtol = 2.0000e-12  rtol = 8.881784e-16
SciPy: root = 1.23 3324359458235e-10 calls = 46 iterations = 44 flag = converged

True root   = 1.234567890123456e-30 xtol = 2.0000e-12  rtol = 8.881784e-16
SciPy: root = 8.159204944952321e-13 calls = 46 iterations = 44 flag = converged

True root   = 1.234567890123456e-50 xtol = 2.0000e-12  rtol = 8.881784e-16
SciPy: root = 8.159204944952321e-13 calls = 46 iterations = 44 flag = converged
```

### Error message

```shell
The problem is that there is no error message. The following is a description of the problem as well as the solution.

Notice that, as the magnitude of the root decreases, the accuracy of the approximation decreases until the root becomes 8.159204944952321e-13 - the width that |b - a| becomes after 44 iterations.

This does not fit the "Slow but sure." claim made in the documentation.

The C code is
	dm = xb - xa;
	...
	xm = xa + dm;
	...
	if (fm == 0 || fabs(dm) < xtol + rtol*fabs(xm))
	
and it is the absolute error part 'xtol' which causes the problem. It compares decimal places for accuracy - even if all digits before any meaningful digits are reached are zeros.

The relative error condition (5e-s) will give 's' significant digits. The only concern has been that problems can arise if the root is close to zero. As can be seen in the above example, 'bisect' returns a completely inaccurate appoximation to small roots - with no warning of error.  

At a minimum, bisect needs to flag the fact that, if the root is less that 0.1, the approximation will not be accurate.

In my following up on this problem, I came across this -

http://web.mit.edu/10.001/Web/Tips/Converge.htm

Convergence Tests, RTOL and ATOL 

Tolerances are usually specified as either a relative tolerance RTOL or an absolute tolerance ATOL, or both. The user typically desires that
	| True value – Computed value | < RTOL*|True Value| + ATOL            (Eq.1)
where the RTOL controls the number of significant figures in the computed value (a float or a double), and a small ATOL is a just a “safety net” for the case where True Value is close to zero. (What would happen if ATOL = 0 and True Value = 0? Would the convergence test ever be satisfied?) You should write your programs to take both RTOL and ATOL as inputs.

I do not know if this is the origin of the idea of using this method of stopping, but it needs to be tested.

As can be seen above, the 'a small ATOL is a just a “safety net” for the case where True Value is close to zero' results in a termination to a grossly inaccurate approximation to the root.

The question is then asked '(What would happen if ATOL = 0 and True Value = 0? Would the convergence test ever be satisfied?)' - but no attempt is made to answer it.

Suppose that we consider any interval [a, b] and bisect it until we can go no further. 

One of 2 things must happen:
	1) we reach a point at which f(c) = 0 so we have a root, or
	2) the interval can no longer be bisected i.e. the interval has reached the minimum size allowed by the computer. For those using IEEE 754, that minimum is the minimum subnormal value 4.940656458412465e-324 which occurs when a & b differ only in the last bit.
	
This means that setting xtol = 4.940656458412465e-324 will assure that the iterations will stop.

Consider the following code 

_rtol = 4 * np.finfo(float).eps
min_sub = np.finfo(float).smallest_subnormal
for k in [1e-50, 1e-100, 1e-150, 1e-200, 1e-300, 1e-310, 1e-311, 1e-312, 1e-313, 1e-315, 1e-320]:
    tmp = optimize.bisect(lambda x: x - 1.234567890123456*k, -10, 10+1/3, xtol=min_sub, rtol=_rtol,
                          maxiter=2000, full_output=True)
    r_root = 1.234567890123456 * k
    print('\nTrue root   = %.15e xtol = %.15e  rtol = %6e' % (r_root, min_sub, _rtol))
    print('SciPy: root = %.15e calls = %d iterations = %d flag = %s'
          % (tmp[0], tmp[1].function_calls, tmp[1].iterations,  tmp[1].flag))


True root   = 1.234567890123456e-50 xtol = 4.940656458412465e-324  rtol = 8.881784e-16
SciPy: root = 1.234567890123457e-50 calls = 223 iterations = 221 flag = converged

True root   = 1.234567890123456e-100 xtol = 4.940656458412465e-324  rtol = 8.881784e-16
SciPy: root = 1.234567890123456e-100 calls = 389 iterations = 387 flag = converged

True root   = 1.234567890123456e-150 xtol = 4.940656458412465e-324  rtol = 8.881784e-16
SciPy: root = 1.234567890123456e-150 calls = 555 iterations = 553 flag = converged

True root   = 1.234567890123456e-200 xtol = 4.940656458412465e-324  rtol = 8.881784e-16
SciPy: root = 1.234567890123456e-200 calls = 721 iterations = 719 flag = converged

True root   = 1.234567890123456e-300 xtol = 4.940656458412465e-324  rtol = 8.881784e-16
SciPy: root = 1.234567890123456e-300 calls = 1052 iterations = 1050 flag = converged

True root   = 1.234567890123457e-310 xtol = 4.940656458412465e-324  rtol = 8.881784e-16
SciPy: root = 1.234567890123457e-310 calls = 1077 iterations = 1075 flag = converged

True root   = 1.23456789012 3260e-311 xtol = 4.940656458412465e-324  rtol = 8.881784e-16  ***
SciPy: root = 1.23456789012 2765e-311 calls = 1081 iterations = 1079 flag = converged

True root   = 1.234567890119307e-312 xtol = 4.940656458412465e-324  rtol = 8.881784e-16
SciPy: root = 1.234567890119307e-312 calls = 1079 iterations = 1077 flag = converged

True root   = 1.234567890163773e-313 xtol = 4.940656458412465e-324  rtol = 8.881784e-16
SciPy: root = 1.234567890163773e-313 calls = 1080 iterations = 1078 flag = converged

True root   = 1.234567886063028e-315 xtol = 4.940656458412465e-324  rtol = 8.881784e-16
SciPy: root = 1.234567881122372e-315 calls = 1081 iterations = 1079 flag = converged

True root   = 1.234670048957275e-320 xtol = 4.940656458412465e-324  rtol = 8.881784e-16
SciPy: root = 1.234175983311434e-320 calls = 1081 iterations = 1079 flag = converged

After the 'e-311' case, the 'True root' is no longer shown as 1.234567890123456 e-3xx because we are into the 'subnormal' range (min normal: np.finfo(float).tiny
2.2250738585072014e-308). We might also note that we are trying to get 15 digits when we are within 15 digits to the min subnormal.

This shows that, if the c-code is changed to
	if (fm == 0 || fabs(dm) < 4.940656458412465e-324  + rtol*fabs(xm)
bisect will give an accurate approximation.

While this requires a large number of iterations, it satisfies the "Slow but sure" statement. It should also be noted that the need for this number of iterations would be rare but 'hard wiring' the min subnormal as xtol and not allowing the user to change it would guarantee correct results.

The answer to the old question "Do you want it fast or do you want it right?" must always be "Right".
```

```shell
1.10.1 1.25.0 sys.version_info(major=3, minor=11, micro=0, releaselevel='final', serial=0)
{
  "Compilers": {
    "c": {
      "name": "gcc",
      "linker": "ld.bfd",
      "version": "10.3.0",
      "commands": "cc"
    },
    "cython": {
      "name": "cython",
      "linker": "cython",
      "version": "0.29.33",
      "commands": "cython"
    },
    "c++": {
      "name": "gcc",
      "linker": "ld.bfd",
      "version": "10.3.0",
      "commands": "c++"
    },
    "fortran": {
      "name": "gcc",
      "linker": "ld.bfd",
      "version": "10.3.0",
      "commands": "gfortran"
    },
    "pythran": {
      "version": "0.12.1",
      "include directory": "C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pip-build-env-tkj0pjht\\overlay\\Lib\\site-packages\\pythran"
    }
  },
  "Machine Information": {
    "host": {
      "cpu": "x86_64",
      "family": "x86_64",
      "endian": "little",
      "system": "windows"
    },
    "build": {
      "cpu": "x86_64",
      "family": "x86_64",
      "endian": "little",
      "system": "windows"
    },
    "cross-compiled": false
  },
  "Build Dependencies": {
    "blas": {
      "name": "openblas",
      "found": true,
      "version": "0.3.18",
      "include directory": "c:/opt/openblas/if_32/64/include",
      "lib directory": "c:/opt/openblas/if_32/64/lib",
      "openblas configuration": "USE_64BITINT= DYNAMIC_ARCH=1 DYNAMIC_OLDER= NO_CBLAS= NO_LAPACK= NO_LAPACKE= NO_AFFINITY=1 USE_OPENMP= PRESCOTT MAX_THREADS=4",
      "pc file directory": "c:/opt/openblas/if_32/64/lib/pkgconfig"
    },
    "lapack": {
      "name": "openblas",
      "found": true,
      "version": "0.3.18",
      "include directory": "c:/opt/openblas/if_32/64/include",
      "lib directory": "c:/opt/openblas/if_32/64/lib",
      "openblas configuration": "USE_64BITINT= DYNAMIC_ARCH=1 DYNAMIC_OLDER= NO_CBLAS= NO_LAPACK= NO_LAPACKE= NO_AFFINITY=1 USE_OPENMP= PRESCOTT MAX_THREADS=4",
      "pc file directory": "c:/opt/openblas/if_32/64/lib/pkgconfig"
    }
  },
  "Python Information": {
    "path": "C:\\Users\\runneradmin\\AppData\\Local\\Temp\\cibw-run-7zsnxupn\\cp311-win_amd64\\build\\venv\\Scripts\\python.exe",
    "version": "3.11"
  }
}
```

steppi:
#### Reference issue

Closes gh-18740

This PR makes a docs fix to the problem reported in gh-18740 for `bisect`. The problem was that a choice of the default absolute tolerance `xtol` for deciding convergence can lead to very large relative errors for roots near zero because the absolute error is deemed small enough. In that issue, @mdhaber and I both agreed that a smaller default would make more sense if starting from scratch but that it's better to be cautious about changing defaults in widely used, over 20-year-old code. Instead, this just clarifies documentation of the termination condition, makes users aware of potential surprises due to the default value of `xtol`, and suggests taking care to select `xtol` for the use case at hand.

I've also changed appearances of `np.allclose` in the documentation of the termination condition to `np.isclose`, which I think makes more sense in a scalar context, and added `scipy.optimize.elementwise.find_root` to the `See Also` sections of the other bracketing root finders.

#### Additional information

I didn't make any mention of setting `xtol` to the smallest subnormal to o

### DIFF
--- a/scipy/optimize/_zeros_py.py
+++ b/scipy/optimize/_zeros_py.py
@@ -513,11 +513,11 @@ def bisect(f, a, b, args=(),
     b : scalar
         The other end of the bracketing interval [a,b].
     xtol : number, optional
-        The computed root ``x0`` will satisfy ``np.allclose(x, x0,
+        The computed root ``x0`` will satisfy ``np.isclose(x, x0,
         atol=xtol, rtol=rtol)``, where ``x`` is the exact root. The
         parameter must be positive.
     rtol : number, optional
-        The computed root ``x0`` will satisfy ``np.allclose(x, x0,
+        The computed root ``x0`` will satisfy ``np.isclose(x, x0,
         atol=xtol, rtol=rtol)``, where ``x`` is the exact root. The
         parameter cannot be smaller than its default value of
         ``4*np.finfo(float).eps``.
@@ -595,11 +595,11 @@ def ridder(f, a, b, args=(),
     b : scalar
         The other end of the bracketing interval [a,b].
     xtol : number, optional
-        The computed root ``x0`` will satisfy ``np.allclose(x, x0,
+        The computed root ``x0`` will satisfy ``np.isclose(x, x0,
         atol=xtol, rtol=rtol)``, where ``x`` is the exact root. The
         parameter must be positive.
     rtol : number, optional
-        The computed root ``x0`` will satisfy ``np.allclose(x, x0,
+        The computed root ``x0`` will satisfy ``np.isclose(x, x0,
         atol=xtol, rtol=rtol)``, where ``x`` is the exact root. The
         parameter cannot be smaller than its default value of
         ``4*np.finfo(float).eps``.
@@ -710,13 +710,13 @@ def brentq(f, a, b, args=(),
     b : scalar
         The other end of the bracketing interval :math:`[a, b]`.
     xtol : number, optional
-        The computed root ``x0`` will satisfy ``np.allclose(x, x0,
+        The computed root ``x0`` will satisfy ``np.isclose(x, x0,
         atol=xtol, rtol=rtol)``, where ``x`` is the exact root. The
         parameter must be positive. For nice functions, Brent's
         method will often satisfy the above condition with ``xtol/2``
         and ``rtol/2``. [Brent1973]_
     rtol : number, optional
-        The computed root ``x0`` will satisfy ``np.allclose(x, x0,
+        The computed root ``x0`` will satisfy ``np.isclose(x, x0,
         atol=xtol, rtol=rtol)``, where ``x`` is the exact root. The
         parameter cannot be smaller than its default value of
         ``4*np.finfo(float).eps``. For nice functions, Brent's
@@ -828,13 +828,13 @@ def brenth(f, a, b, args=(),
     b : scalar
         The other end of the bracketing interval [a,b].
     xtol : number, optional
-        The computed root ``x0`` will satisfy ``np.allclose(x, x0,
+        The computed root ``x0`` will satisfy ``np.isclose(x, x0,
         atol=xtol, rtol=rtol)``, where ``x`` is the exact root. The
         parameter must be positive. As with `brentq`, for nice
         functions the method will often satisfy the above condition
         with ``xtol/2`` and ``rtol/2``.
     rtol : number, optional
-        The computed root ``x0`` will satisfy ``np.allclose(x, x0,
+        The computed root ``x0`` will satisfy ``np.isclose(x, x0,
         atol=xtol, rtol=rtol)``, where ``x`` is the exact root. The
         parameter cannot be smaller than its default value of
         ``4*np.finfo(float).eps``. As with `brentq`, for nice functions
@@ -1292,11 +1292,11 @@ def toms748(f, a, b, args=(), k=1,
         The number of Newton quadratic steps to perform each
         iteration. ``k>=1``.
     xtol : scalar, optional
-        The computed root ``x0`` will satisfy ``np.allclose(x, x0,
+        The computed root ``x0`` will satisfy ``np.isclose(x, x0,
         atol=xtol, rtol=rtol)``, where ``x`` is the exact root. The
         parameter must be positive.
     rtol : scalar, optional
-        The computed root ``x0`` will satisfy ``np.allclose(x, x0,
+        The computed root ``x0`` will satisfy ``np.isclose(x, x0,
         atol=xtol, rtol=rtol)``, where ``x`` is the exact root.
     maxiter : int, optional
         If convergence is not achieved in `maxiter` iterations, an error is
--- a/scipy/optimize/_zeros_py.py
+++ b/scipy/optimize/_zeros_py.py
@@ -544,6 +544,21 @@ def bisect(f, a, b, args=(),
         Object containing information about the convergence. In particular,
         ``r.converged`` is True if the routine converged.
 
+    Notes
+    -----
+    As mentioned in the parameter documentation, the computed root ``x0`` will
+    satisfy ``np.isclose(x, x0, atol=xtol, rtol=rtol)``, where ``x`` is the
+    exact root. In equation form, this terminating condition is ``abs(x - x0)
+    <= xtol + rtol * abs(x)``.
+
+    A nonzero value of `xtol` may be useful for saving function evaluations
+    when a root is near zero in applications where the tiny absolute
+    differences available between floating point numbers near zero are not
+    meaningful. The default value ``xtol=2e-12`` may lead to surprising
+    behavior if one expects `bisect` to always compute roots with relative
+    error near machine precision. Care should be taken to select `xtol` for the
+    use case at hand.
+
     Examples
     --------
 
@@ -641,6 +656,19 @@ def ridder(f, a, b, args=(),
 
     The routine used here diverges slightly from standard presentations in
     order to be a bit more careful of tolerance.
+
+    As mentioned in the parameter documentation, the computed root ``x0`` will
+    satisfy ``np.isclose(x, x0, atol=xtol, rtol=rtol)``, where ``x`` is the
+    exact root. In equation form, this terminating condition is ``abs(x - x0)
+    <= xtol + rtol * abs(x)``.
+
+    A nonzero value of `xtol` may be useful for saving function evaluations
+    when a root is near zero in applications where the tiny absolute
+    differences available between floating point numbers near zero are not
+    meaningful. The default value ``xtol=2e-12`` may lead to surprising
+    behavior if one expects `ridder` to always compute roots with relative
+    error near machine precision. Care should be taken to select `xtol` for the
+    use case at hand.
 
     References
     ----------
@@ -760,6 +788,19 @@ def brentq(f, a, b, args=(),
     -----
     `f` must be continuous.  f(a) and f(b) must have opposite signs.
 
+    As mentioned in the parameter documentation, the computed root ``x0`` will
+    satisfy ``np.isclose(x, x0, atol=xtol, rtol=rtol)``, where ``x`` is the
+    exact root. In equation form, this terminating condition is ``abs(x - x0)
+    <= xtol + rtol * abs(x)``.
+
+    A nonzero value of `xtol` may be useful for saving function evaluations
+    when a root is near zero in applications where the tiny absolute
+    differences available between floating point numbers near zero are not
+    meaningful. The default value ``xtol=2e-12`` may lead to surprising
+    behavior if one expects `brentq` to always compute roots with relative
+    error near machine precision. Care should be taken to select `xtol` for the
+    use case at hand.
+
     References
     ----------
     .. [Brent1973]
@@ -873,6 +914,20 @@ def brenth(f, a, b, args=(),
     fsolve : N-D root-finding
     brentq, ridder, bisect, newton : 1-D root-finding
     fixed_point : scalar fixed-point finder
+    Notes
+    -----
+    As mentioned in the parameter documentation, the computed root ``x0`` will
+    satisfy ``np.isclose(x, x0, atol=xtol, rtol=rtol)``, where ``x`` is the
+    exact root. In equation form, this terminating condition is ``abs(x - x0)
+    <= xtol + rtol * abs(x)``.
+
+    A nonzero value of `xtol` may be useful for saving function evaluations
+    when a root is near zero in applications where the tiny absolute
+    differences available between floating point numbers near zero are not
+    meaningful. The default value ``xtol=2e-12`` may lead to surprising
+    behavior if one expects `brenth` to always compute roots with relative
+    error near machine precision. Care should be taken to select `brenth` for the
+    use case at hand.
 
     References
     ----------
@@ -1343,6 +1398,19 @@ def toms748(f, a, b, args=(), k=1,
     For higher values of `k`, the efficiency index approaches
     the kth root of ``(3k-2)``, hence ``k=1`` or ``k=2`` are
     usually appropriate.
+
+    As mentioned in the parameter documentation, the computed root ``x0`` will
+    satisfy ``np.isclose(x, x0, atol=xtol, rtol=rtol)``, where ``x`` is the
+    exact root. In equation form, this terminating condition is ``abs(x - x0)
+    <= xtol + rtol * abs(x)``.
+
+    A nonzero value of `xtol` may be useful for saving function evaluations
+    when a root is near zero in applications where the tiny absolute
+    differences available between floating point numbers near zero are not
+    meaningful. The default value ``xtol=2e-12`` may lead to surprising
+    behavior if one expects `toms748` to always compute roots with relative
+    error near machine precision. Care should be taken to select `xtol` for the
+    use case at hand.
 
     References
     ----------
--- a/scipy/optimize/_zeros_py.py
+++ b/scipy/optimize/_zeros_py.py
@@ -549,7 +549,7 @@ def bisect(f, a, b, args=(),
     As mentioned in the parameter documentation, the computed root ``x0`` will
     satisfy ``np.isclose(x, x0, atol=xtol, rtol=rtol)``, where ``x`` is the
     exact root. In equation form, this terminating condition is ``abs(x - x0)
-    <= xtol + rtol * abs(x)``.
+    <= xtol + rtol * abs(x0)``.
 
     The default value ``xtol=2e-12`` may lead to surprising behavior if one
     expects `bisect` to always compute roots with relative error near machine
--- a/scipy/optimize/_zeros_py.py
+++ b/scipy/optimize/_zeros_py.py
@@ -663,7 +663,7 @@ def ridder(f, a, b, args=(),
     As mentioned in the parameter documentation, the computed root ``x0`` will
     satisfy ``np.isclose(x, x0, atol=xtol, rtol=rtol)``, where ``x`` is the
     exact root. In equation form, this terminating condition is ``abs(x - x0)
-    <= xtol + rtol * abs(x)``.
+    <= xtol + rtol * abs(x0)``.
 
     The default value ``xtol=2e-12`` may lead to surprising behavior if one
     expects `ridder` to always compute roots with relative error near machine
@@ -796,7 +796,7 @@ def brentq(f, a, b, args=(),
     As mentioned in the parameter documentation, the computed root ``x0`` will
     satisfy ``np.isclose(x, x0, atol=xtol, rtol=rtol)``, where ``x`` is the
     exact root. In equation form, this terminating condition is ``abs(x - x0)
-    <= xtol + rtol * abs(x)``.
+    <= xtol + rtol * abs(x0)``.
 
     The default value ``xtol=2e-12`` may lead to surprising behavior if one
     expects `brentq` to always compute roots with relative error near machine
@@ -927,7 +927,7 @@ def brenth(f, a, b, args=(),
     As mentioned in the parameter documentation, the computed root ``x0`` will
     satisfy ``np.isclose(x, x0, atol=xtol, rtol=rtol)``, where ``x`` is the
     exact root. In equation form, this terminating condition is ``abs(x - x0)
-    <= xtol + rtol * abs(x)``.
+    <= xtol + rtol * abs(x0)``.
 
     The default value ``xtol=2e-12`` may lead to surprising behavior if one
     expects `brenth` to always compute roots with relative error near machine
@@ -1412,7 +1412,7 @@ def toms748(f, a, b, args=(), k=1,
     As mentioned in the parameter documentation, the computed root ``x0`` will
     satisfy ``np.isclose(x, x0, atol=xtol, rtol=rtol)``, where ``x`` is the
     exact root. In equation form, this terminating condition is ``abs(x - x0)
-    <= xtol + rtol * abs(x)``.
+    <= xtol + rtol * abs(x0)``.
 
     The default value ``xtol=2e-12`` may lead to surprising behavior if one
     expects `toms748` to always compute roots with relative error near machine
--- a/scipy/optimize/_zeros_py.py
+++ b/scipy/optimize/_zeros_py.py
@@ -580,6 +580,7 @@ def bisect(f, a, b, args=(),
     brentq, brenth, bisect, newton
     fixed_point : scalar fixed-point finder
     fsolve : n-dimensional root-finding
+    elementwise.find_root : Efficient elementwise 1-D root-finder.
 
     """
     if not isinstance(args, tuple):
@@ -645,6 +646,7 @@ def ridder(f, a, b, args=(),
     --------
     brentq, brenth, bisect, newton : 1-D root-finding
     fixed_point : scalar fixed-point finder
+    elementwise.find_root : Efficient elementwise 1-D root-finder.
 
     Notes
     -----
@@ -783,6 +785,7 @@ def brentq(f, a, b, args=(),
     fsolve : N-D root-finding
     brenth, ridder, bisect, newton : 1-D root-finding
     fixed_point : scalar fixed-point finder
+    elementwise.find_root : Efficient elementwise 1-D root-finder.
 
     Notes
     -----
@@ -914,6 +917,8 @@ def brenth(f, a, b, args=(),
     fsolve : N-D root-finding
     brentq, ridder, bisect, newton : 1-D root-finding
     fixed_point : scalar fixed-point finder
+    elementwise.find_root : Efficient elementwise 1-D root-finder.
+
     Notes
     -----
     As mentioned in the parameter documentation, the computed root ``x0`` will
@@ -1377,6 +1382,7 @@ def toms748(f, a, b, args=(), k=1,
     --------
     brentq, brenth, ridder, bisect, newton
     fsolve : find roots in N dimensions.
+    elementwise.find_root : Efficient elementwise 1-D root-finder.
 
     Notes
     -----
--- a/scipy/optimize/_zeros_py.py
+++ b/scipy/optimize/_zeros_py.py
@@ -581,7 +581,7 @@ def bisect(f, a, b, args=(),
     brentq, brenth, bisect, newton
     fixed_point : scalar fixed-point finder
     fsolve : n-dimensional root-finding
-    elementwise.find_root : Efficient elementwise 1-D root-finder.
+    elementwise.find_root : efficient elementwise 1-D root-finder
 
     """
     if not isinstance(args, tuple):
@@ -647,7 +647,7 @@ def ridder(f, a, b, args=(),
     --------
     brentq, brenth, bisect, newton : 1-D root-finding
     fixed_point : scalar fixed-point finder
-    elementwise.find_root : Efficient elementwise 1-D root-finder.
+    elementwise.find_root : efficient elementwise 1-D root-finder
 
     Notes
     -----
@@ -787,7 +787,7 @@ def brentq(f, a, b, args=(),
     fsolve : N-D root-finding
     brenth, ridder, bisect, newton : 1-D root-finding
     fixed_point : scalar fixed-point finder
-    elementwise.find_root : Efficient elementwise 1-D root-finder.
+    elementwise.find_root : efficient elementwise 1-D root-finder
 
     Notes
     -----
@@ -920,7 +920,7 @@ def brenth(f, a, b, args=(),
     fsolve : N-D root-finding
     brentq, ridder, bisect, newton : 1-D root-finding
     fixed_point : scalar fixed-point finder
-    elementwise.find_root : Efficient elementwise 1-D root-finder.
+    elementwise.find_root : efficient elementwise 1-D root-finder
 
     Notes
     -----
@@ -1386,7 +1386,7 @@ def toms748(f, a, b, args=(), k=1,
     --------
     brentq, brenth, ridder, bisect, newton
     fsolve : find roots in N dimensions.
-    elementwise.find_root : Efficient elementwise 1-D root-finder.
+    elementwise.find_root : efficient elementwise 1-D root-finder
 
     Notes
     -----
--- a/scipy/optimize/_zeros_py.py
+++ b/scipy/optimize/_zeros_py.py
@@ -551,13 +551,14 @@ def bisect(f, a, b, args=(),
     exact root. In equation form, this terminating condition is ``abs(x - x0)
     <= xtol + rtol * abs(x)``.
 
-    A nonzero value of `xtol` may be useful for saving function evaluations
-    when a root is near zero in applications where the tiny absolute
-    differences available between floating point numbers near zero are not
-    meaningful. The default value ``xtol=2e-12`` may lead to surprising
-    behavior if one expects `bisect` to always compute roots with relative
-    error near machine precision. Care should be taken to select `xtol` for the
-    use case at hand.
+    The default value ``xtol=2e-12`` may lead to surprising behavior if one
+    expects `bisect` to always compute roots with relative error near machine
+    precision. Care should be taken to select `xtol` for the use case at hand.
+    Setting ``xtol=5e-324``, the smallest subnormal number, will ensure the
+    highest level of accuracy. Larger values of `xtol` may be useful for saving
+    function evaluations when a root is at or near zero in applications where
+    the tiny absolute differences available between floating point numbers near
+    zero are not meaningful.
 
     Examples
     --------
@@ -664,13 +665,14 @@ def ridder(f, a, b, args=(),
     exact root. In equation form, this terminating condition is ``abs(x - x0)
     <= xtol + rtol * abs(x)``.
 
-    A nonzero value of `xtol` may be useful for saving function evaluations
-    when a root is near zero in applications where the tiny absolute
-    differences available between floating point numbers near zero are not
-    meaningful. The default value ``xtol=2e-12`` may lead to surprising
-    behavior if one expects `ridder` to always compute roots with relative
-    error near machine precision. Care should be taken to select `xtol` for the
-    use case at hand.
+    The default value ``xtol=2e-12`` may lead to surprising behavior if one
+    expects `ridder` to always compute roots with relative error near machine
+    precision. Care should be taken to select `xtol` for the use case at hand.
+    Setting ``xtol=5e-324``, the smallest subnormal number, will ensure the
+    highest level of accuracy. Larger values of `xtol` may be useful for saving
+    function evaluations when a root is at or near zero in applications where
+    the tiny absolute differences available between floating point numbers near
+    zero are not meaningful.
 
     References
     ----------
@@ -796,13 +798,14 @@ def brentq(f, a, b, args=(),
     exact root. In equation form, this terminating condition is ``abs(x - x0)
     <= xtol + rtol * abs(x)``.
 
-    A nonzero value of `xtol` may be useful for saving function evaluations
-    when a root is near zero in applications where the tiny absolute
-    differences available between floating point numbers near zero are not
-    meaningful. The default value ``xtol=2e-12`` may lead to surprising
-    behavior if one expects `brentq` to always compute roots with relative
-    error near machine precision. Care should be taken to select `xtol` for the
-    use case at hand.
+    The default value ``xtol=2e-12`` may lead to surprising behavior if one
+    expects `brentq` to always compute roots with relative error near machine
+    precision. Care should be taken to select `xtol` for the use case at hand.
+    Setting ``xtol=5e-324``, the smallest subnormal number, will ensure the
+    highest level of accuracy. Larger values of `xtol` may be useful for saving
+    function evaluations when a root is at or near zero in applications where
+    the tiny absolute differences available between floating point numbers near
+    zero are not meaningful.
 
     References
     ----------
@@ -926,13 +929,14 @@ def brenth(f, a, b, args=(),
     exact root. In equation form, this terminating condition is ``abs(x - x0)
     <= xtol + rtol * abs(x)``.
 
-    A nonzero value of `xtol` may be useful for saving function evaluations
-    when a root is near zero in applications where the tiny absolute
-    differences available between floating point numbers near zero are not
-    meaningful. The default value ``xtol=2e-12`` may lead to surprising
-    behavior if one expects `brenth` to always compute roots with relative
-    error near machine precision. Care should be taken to select `brenth` for the
-    use case at hand.
+    The default value ``xtol=2e-12`` may lead to surprising behavior if one
+    expects `brenth` to always compute roots with relative error near machine
+    precision. Care should be taken to select `xtol` for the use case at hand.
+    Setting ``xtol=5e-324``, the smallest subnormal number, will ensure the
+    highest level of accuracy. Larger values of `xtol` may be useful for saving
+    function evaluations when a root is at or near zero in applications where
+    the tiny absolute differences available between floating point numbers near
+    zero are not meaningful.
 
     References
     ----------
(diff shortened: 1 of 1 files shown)


scipy/optimize/_zeros_py.py
```diff
@@ -544,6 +544,22 @@ def bisect(f, a, b, args=(),
         Object containing information about the convergence. In particular,
         ``r.converged`` is True if the routine converged.
 
+    Notes
+    -----
+    As mentioned in the parameter documentation, the computed root ``x0`` will
+    satisfy ``np.isclose(x, x0, atol=xtol, rtol=rtol)``, where ``x`` is the
+    exact root. In equation form, this terminating condition is ``abs(x - x0)
+    <= xtol + rtol * abs(x)``.
```
mdhaber, line 552: Technically,
```suggestion
    <= xtol + rtol * abs(x0)``.
```
right? That's how `isclose` is documented, and I am just guessing we could cook up a situation where the tolerances are loose and the difference could matter. (But it's possible that guess is wrong.)

mdhaber (approved):
Thanks! Just one thought.

scipy/optimize/_zeros_py.py
steppi: Yes, you're absolutely correct. I don't think what I wrote is even possible because the actual root isn't known so it's magnitude can't be used in the stopping criterion like that. I think there are situations where the difference could matter, I'll make a note to revisit that after I start getting ahead of my backlog.

scipy/optimize/_zeros_py.py
```diff
@@ -642,6 +660,20 @@ def ridder(f, a, b, args=(),
     The routine used here diverges slightly from standard presentations in
     order to be a bit more careful of tolerance.
 
+    As mentioned in the parameter documentation, the computed root ``x0`` will
+    satisfy ``np.isclose(x, x0, atol=xtol, rtol=rtol)``, where ``x`` is the
+    exact root. In equation form, this terminating condition is ``abs(x - x0)
+    <= xtol + rtol * abs(x)``.
```
mdhaber, line 666:
```suggestion
    <= xtol + rtol * abs(x0)``.
```

scipy/optimize/_zeros_py.py
```diff
@@ -755,11 +787,26 @@ def brentq(f, a, b, args=(),
...
+    elementwise.find_root : efficient elementwise 1-D root-finder
 
     Notes
     -----
     `f` must be continuous.  f(a) and f(b) must have opposite signs.
 
+    As mentioned in the parameter documentation, the computed root ``x0`` will
+    satisfy ``np.isclose(x, x0, atol=xtol, rtol=rtol)``, where ``x`` is the
+    exact root. In equation form, this terminating condition is ``abs(x - x0)
+    <= xtol + rtol * abs(x)``.
```
mdhaber, line 799:
```suggestion
    <= xtol + rtol * abs(x0)``.
```

scipy/optimize/_zeros_py.py
```diff
@@ -873,6 +920,23 @@ def brenth(f, a, b, args=(),
...
     brentq, ridder, bisect, newton : 1-D root-finding
     fixed_point : scalar fixed-point finder
+    elementwise.find_root : efficient elementwise 1-D root-finder
+
+    Notes
+    -----
+    As mentioned in the parameter documentation, the computed root ``x0`` will
+    satisfy ``np.isclose(x, x0, atol=xtol, rtol=rtol)``, where ``x`` is the
+    exact root. In equation form, this terminating condition is ``abs(x - x0)
+    <= xtol + rtol * abs(x)``.
```
mdhaber, line 930:
```suggestion
    <= xtol + rtol * abs(x0)``.
```

scipy/optimize/_zeros_py.py
```diff
@@ -1344,6 +1409,20 @@ def toms748(f, a, b, args=(), k=1,
     the kth root of ``(3k-2)``, hence ``k=1`` or ``k=2`` are
     usually appropriate.
 
+    As mentioned in the parameter documentation, the computed root ``x0`` will
+    satisfy ``np.isclose(x, x0, atol=xtol, rtol=rtol)``, where ``x`` is the
+    exact root. In equation form, this terminating condition is ``abs(x - x0)
+    <= xtol + rtol * abs(x)``.
```
mdhaber, line 1415:
```suggestion
    <= xtol + rtol * abs(x0)``.
```

mdhaber (approved):
Need to fix it throughout. For that reason, I thought about suggesting that we write this once and use code to put it in the docstrings, but the copy-paste is done, so I think let's roll with it.

steppi:
> Need to fix it throughout. For that reason, I thought about suggesting that we write this once and use code to put it in the docstrings, but the copy-paste is done, so I think let's roll with it.

Oops. Thanks @mdhaber Is there a clean way to add bits of shared boilerplate like this to multiple docstrings? I've wanted to do that before in other situations, but the only thing I could think of was setting `__doc__` equal to an `f-string`.

mdhaber:
Yes, but f-strings is not one. They need to be used with caution to avoid increasing import times. But see, for example, uses of [`scipy._lib.doccer`](https://github.com/scipy/scipy/blob/main/scipy/_lib/doccer.py) [in `ndimage`](https://github.com/scipy/scipy/blob/ae25ba2385e62d5372a47ed59f9cfddc5ab3dc6a/scipy/ndimage/_filters.py#L193) and [`scipy._lib.docscrape`](https://github.com/scipy/scipy/blob/main/scipy/_lib/_docscrape.py) in `_lib._util`, e.g. [`_transition_to_rng`](https://github.com/scipy/scipy/blob/ae25ba2385e62d5372a47ed59f9cfddc5ab3dc6a/scipy/_lib/_util.py#L436) decorator.

steppi:
> > Need to fix it throughout. For that reason, I thought about suggesting that we write this once and use code to put it in the docstrings, but the copy-paste is done, so I think let's roll with it.
> 
> Oops. Thanks @mdhaber Is there a clean way to add bits of shared boilerplate like this to multiple docstrings? I've wanted to do that before in other situations, but the only thing I could think of was setting `__doc__` equal to an `f-string`.


> They need to be used with caution to avoid increasing import times, but see, for example, uses of [`scipy._lib.doccer`](https://github.com/scipy/scipy/blob/main/scipy/_lib/doccer.py) [in `ndimage`](https://github.com/scipy/scipy/blob/ae25ba2385e62d5372a47ed59f9cfddc5ab3dc6a/scipy/ndimage/_filters.py#L193) and [`scipy._lib.docscrape`](https://github.com/scipy/scipy/blob/main/scipy/_lib/_docscrape.py) in `_lib._util`, e.g. [`_transition_to_rng`](https://github.com/scipy/scipy/blob/ae25ba2385e62d5372a47ed59f9cfddc5ab3dc6a/scipy/_lib/_util.py#L436) decorator.

Cool. Good to know.

> Yes, but f-strings is not one - it just doesn't work. They need to be used with caution to avoid increasing import times, but see, for example, uses of [`scipy._lib.doccer`](https://github.com/scipy/scipy/blob/main/scipy/_lib/doccer.py) [in `ndimage`](https://github.com/scipy/scipy/blob/ae25ba2385e62d5372a47ed59f9cfddc5ab3dc6a/scipy/ndimage/_filters.py#L193) and [`scipy._lib.docscrape`](https://github.com/scipy/scipy/blob/main/scipy/_lib/_docscrape.py) in `_lib._util`, e.g. [`_transition_to_rng`](https://github.com/scipy/scipy/blob/ae25ba2385e62d5372a47ed59f9cfddc5ab3dc6a/scipy/_lib/_util.py#L436) decorator.

I meant doing what you did here in the distribution infra

https://github.com/scipy/scipy/blob/ae25ba2385e62d5372a47ed59f9cfddc5ab3dc6a/scipy/stats/_distribution_infrastructure.py#L3780-L3786

not just trying to make the docstring an `f-string`. I know the latter that doesn't work.

EJN38:
BTW - isclose says:
Warning

The default atol is not appropriate for comparing numbers with magnitudes much smaller than one (see Notes).

steppi:
@EJN38. Though I'm sympathetic, please remember that SciPy is a community project and everyone involved in working on your issue did so as a volunteer. There is a process that must be followed to make breaking changes, https://docs.scipy.org/doc/scipy/dev/core-dev/index.html#code, that requires establishing a consensus among maintainers. Updating the documentation to more accurately describe a function and its potential pitfalls is something that individual maintainers are trusted to sign off on. I personally do not have time to spend on trying to get consensus for updating the default for `xtol` for these solvers.

The solvers behave as documented and are not unusable. Asking for better defaults for `xtol` seems more line with a feature request than a bug report, so I think it's reasonable to close your original issue.

EJN38:
"The solvers behave as documented and are not unusable" - but they are only usable For roots that are greater than 1.
I am also sympathetic with the efforts that are put into maintaining such a project. You refer to better defaults for xtoll as being a feature request But I look at it as a change that makes the methods more robust. Without that a reasonable user Might expect that these routines could be used for any value And to me it is a bug when the program could give but, because of some inappropriate choice of values, It fails to do so.

Maybe it would have been better If I had suggested production of some new routines - ridder_rel, newton_rel, bisect_rel ... Which will give double precision accuracy for all numbers in the normal range. Then, at least, Users of the solver routines would not have to write their own programs in order to get accurate results.

of course, this could bring up another problem, - are there other routines that use the absolute error when the relative error is the appropriate method?

steppi:
> Maybe it would have been better If I had suggested production of some new routines - ridder_rel, newton_rel, bisect_rel ... Which will give double precision accuracy for all numbers in the normal range. Then, at least, Users of the solver routines would not have to write their own programs in order to get accurate results.

OK, this actually has me very confused. Why would users need to write their own programs to get accurate results? Why wouldn't

```Python
bisect(f, a, b, xtol=5e-324)
```

suffice?

EJN38:
In order for any of the routines to give DP accuracy, xtol = 0 and rtol = 5e-15. xtol is, in fact, useless as it only gives decimal places. rtol = 5e-t gives t significant digits. so the magnitude of the root is not a problem.
If you run any of the rootfinders with rtol = 5e-15 and xtol = 5e-308 (0 is not accepted) you get DP accuracy for any normal number for functions of the form (x - r)G(x).

steppi:
> In order for any of the routines to give DP accuracy, xtol = 0 and rtol = 5e-15. xtol is, in fact, useless as it only gives decimal places. rtol = 5e-t gives t significant digits. so the magnitude of the root is not a problem.

`xtol` is not useless for the reason now stated in the documentation

*"Larger values of xtol may be useful for saving function evaluations when a root is at or near zero in applications where the tiny absolute differences available between floating point numbers near zero are not meaningful."*

The unfortunate thing in these root-finders is that the default for `xtol` is set surprisingly high, not that an option for `xtol` exists at all.

EJN38:
The problem is that if you get an approximation using xtol, you have no knowledge as to the accuracy of the approx. 
If I set rtol = 5e-2, I know that the result that I get has 2 meaningful digits and, no matter how many iterations are needed - that is the best approx. for those iterations. Set xtol = 0.0000000000005 and approx = 0.00000000000005 will be a good approx to 0.0000000000000000005 i.e. meaningless.

EJN38:
> The following shows why the absolute error is a bad measure to use.
> My code for the bisection and its simplified Newton's method are at the end. 

> Linear 
> f(x)  =  lambda x: (x - 1.23456789012345e1) ** 1
> f'(x) =  lambda x: 1*(x - 1.23456789012345e1) ** 0
> 1.23456789012345e1
> 1.2345678901234500e+01 N=1 converged optimize.ridder(f, -0.1, 13, full_output=True, maxiter=150)
> 1.2345678901234500e+01 N=2 converged optimize.brentq(f,  0, 13, full_output=True, maxiter=150)
> 1.2345678901234500e+01 N=2 converged optimize.brenth(f,  0, 13, full_output=True, maxiter=150)
> 1.2345678901234500e+01 N=0 converged optimize.toms748(f, -1, 13, full_output=True, maxiter=100)
> 1.2345678901234500e+01 N=1 converged optimize.newton(f, x0=0.2, fprime=fp, full_output=True, maxiter=150)
> 1.2345678901234500e+01 N=1 f(c) = 0 newton(f, fp, 1, 5e-15, 222, 'A')
> 1.2345678901234500e+01 N=1 f(c) = 0 newton(f, fp, 1, 5e-15, 800, 'R')
> 1.234567890123**5146**e+01 N=43 converged optimize.bisect(f, 0, 13, full_output=True)
> 1.2345678901234500e+01 N=47 f(c) = 0 bisect(f, 0, 13, 5e-15) 

--1------------------------------------------------------------------------------------------ 


> f(x)  =  lambda x: (x - 1.23456789012345e-1) ** 1
> f'(x) =  lambda x: 1*(x - 1.23456789012345e-1) ** 0
> 1.23456789012345e-1
> 1.2345678901**134496**e-01 N=2 converged optimize.ridder(f, -0.1, 1, full_output=True, maxiter=150)
> 1.2345678901234500e-01 N=2 converged optimize.brentq(f,  0, 1, full_output=True, maxiter=150)
> 1.2345678901234500e-01 N=2 converged optimize.brenth(f,  0, 1, full_output=True, maxiter=150)
> 1.2345678901234500e-01 N=2 converged optimize.toms748(f, -1, 1, full_output=True, maxiter=100)
> 1.2345678901234500e-01 N=1 converged optimize.newton(f, x0=0.2, fprime=fp, full_output=True, maxiter=150)
> 1.2345678901234500e-01 N=2 f(c) = 0 newton(f, fp, 1, 5e-15, 222, 'A')
> 1.2345678901234500e-01 N=2 f(c) = 0 newton(f, fp, 1, 5e-15, 800, 'R')
> 1.2345678901**147039**e-01 N=39 converged optimize.bisect(f, 0, 1, full_output=True)
> 1.2345678901234503e-01 N=52 Tolerance bisect(f, 0, 1, 5e-15)

> =2===================================================================

> f(x)  =  lambda x: (x - 1.23456789012345e-10) ** 1
> f'(x) =  lambda x: 1*(x - 1.23456789012345e-10) ** 0
> 1.23456789012345e-10
> 1.2245680164335582e-10 N=2 converged optimize.ridder(f, -0.1, 1, full_output=True, maxiter=150)
> 1.2345678901234501e-10 N=2 converged optimize.brentq(f,  0, 1, full_output=True, maxiter=150)
> 1.2345678901234501e-10 N=2 converged optimize.brenth(f,  0, 1, full_output=True, maxiter=150)
> 1.2345678901234501e-10 N=2 converged optimize.toms748(f, -1, 1, full_output=True, maxiter=100)
> 1.2345678901234501e-10 N=2 converged optimize.newton(f, x0=0.2, fprime=fp, full_output=True, maxiter=150)
> 1.2345678901234501e-10 N=2 f(c) = 0 newton(f, fp, 1, 5e-15, 222, 'A')
> 1.2345678901234501e-10 N=2 f(c) = 0 newton(f, fp, 1, 5e-15, 800, 'R')
> 1.2**187229003757238**e-10 N=39 converged optimize.bisect(f, 0, 1, full_output=True)
> 1.2345678901234519e-10 N=82 Tolerance bisect(f, 0, 1, 5e-15)

> =3=====================================================================

> optimize.bisect gives bad result - claims 'converged'

> ====================================================================== 
> f(x)  =  lambda x: (x - 1.23456789012345e-100) ** 1
> f'(x) =  lambda x: 1*(x - 1.23456789012345e-100) ** 0 
> 
> 1.23456789012345e-100
> **1.0000056338554941e-12** N=2 converged optimize.ridder(f, -0.1, 1, full_output=True, maxiter=150)
> **0.0000000000000000e+00** N=2 converged optimize.brentq(f,  0, 1, full_output=True, maxiter=150)
> **0.0000000000000000e+00** N=2 converged optimize.brenth(f,  0, 1, full_output=True, maxiter=150)
> 1.2345678901234501e-100 N=2 converged optimize.toms748(f, -1, 1, full_output=True, maxiter=100)
> 1.2345678901234501e-100 N=2 converged optimize.newton(f, x0=0.2, fprime=fp, full_output=True, maxiter=150)
> 1.2345678901234501e-100 N=2 f(c) = 0 newton(f, fp, 1, 5e-15, 222, 'A')
> 1.2345678901234501e-100 N=2 f(c) = 0 newton(f, fp, 1, 5e-15, 800, 'R')
> **1.8189894035458565e-12** N=39 converged optimize.bisect(f, 0, 1, full_output=True)
> 1.2345678901234501e-100 N=381 f(c) = 0 bisect(f, 0, 1, 5e-15)

> =4=====================================================================

> ridder, brent, optimize.bisect give bad result - claims 'converged'

======================================================================

> Cubic

> ======================================================================

> f(x)  =  lambda x: (x - 1.23456789012345e1) ** 3
> f'(x) =  lambda x: 3*(x - 1.23456789012345e1) ** 2
> 1.23456789012345e1
> 1.2345678901234459e+01 N=35 converged optimize.ridder(f, -0.1, 13, full_output=True, maxiter=150)
> 1.2345678901234690e+01 N=127 converged optimize.brentq(f,  0, 13, full_output=True, maxiter=150)
> 1.2345678901234990e+01 N=125 converged optimize.brenth(f,  0, 13, full_output=True, maxiter=150)
> 1.2345678901234857e+01 N=33 converged optimize.toms748(f, -1, 13, full_output=True, maxiter=100)
> 1.2345678872661876e+01 N=49 converged optimize.newton(f, x0=0.2, fprime=fp, full_output=True, maxiter=150)
> 1.2345678901234491e+01 N=86 Tolerance newton(f, fp, 1, 5e-15, 222, 'A')
> 1.2345678901234406e+01 N=80 Tolerance newton(f, fp, 1, 5e-15, 800, 'R')
> 1.2345678901235146e+01 N=43 converged optimize.bisect(f, 0, 13, full_output=True)
> 1.2345678901234500e+01 N=47 f(c) = 0 bisect(f, 0, 13, 5e-15)

 =5=====================================================================

> All OK

> ======================================================================
> f(x)  =  lambda x: (x - 1.23456789012345e-1) ** 3
> f'(x) =  lambda x: 3*(x - 1.23456789012345e-1) ** 2
> 1.23456789012345e-1
> 1.23456789012**00113**e-01 N=32 converged optimize.ridder(f, -0.1, 1, full_output=True, maxiter=150)
> 1.23456789012**49352**e-01 N=115 converged optimize.brentq(f,  0, 1, full_output=True, maxiter=150)
> 1.2345678901**192131**e-01 N=115 converged optimize.brenth(f,  0, 1, full_output=True, maxiter=150)
> 1.23456789012**68434**e-01 N=31 converged optimize.toms748(f, -1, 1, full_output=True, maxiter=100)
> 1.234568123**7542477**e-01 N=37 converged optimize.newton(f, x0=0.2, fprime=fp, full_output=True, maxiter=150)
> 1.234567890123**5217**e-01 N=80 Tolerance newton(f, fp, 1, 5e-15, 222, 'A')
> 1.2345678901234594e-01 N=85 Tolerance newton(f, fp, 1, 5e-15, 800, 'R')
> 1.2345678901**147039**e-01 N=39 converged optimize.bisect(f, 0, 1, full_output=True)
> 1.2345678901234503e-01 N=52 Tolerance bisect(f, 0, 1, 5e-15)

==6====================================================================

> My newton & bisect OK

> ======================================================================
> f(x)  =  lambda x: (x - 1.23456789012345e-10) ** 3
> f'(x) =  lambda x: 3*(x - 1.23456789012345e-10) ** 2
> 1.23456789012345e-10
> 1.23**76889422007515**e-10 N=32 converged optimize.ridder(f, -0.1, 1, full_output=True, maxiter=150)
> 1.23**66664211711684**e-10 N=30 converged optimize.brentq(f,  0, 1, full_output=True, maxiter=150)
> 1.23**66664211497804**e-10 N=30 converged optimize.brenth(f,  0, 1, full_output=True, maxiter=150)
> 1.23**28521210081680**e-10 N=24 converged optimize.toms748(f, -1, 1, full_output=True, maxiter=100)
> **2.7254774823714496e-08** N=39 converged optimize.newton(f, x0=0.2, fprime=fp, full_output=True, maxiter=150)
> 1.234**6496799477964e**-10 N=80 Tolerance newton(f, fp, 1, 5e-15, 222, 'A')
> 1.2345678901234**612**e-10 N=136 Tolerance newton(f, fp, 1, 5e-15, 800, 'R')
> 1.2**187229003757238**e-10 N=39 converged optimize.bisect(f, 0, 1, full_output=True)
> 1.2345678901234519e-10 N=82 Tolerance bisect(f, 0, 1, 5e-15)

>=7=====================================================================

> My newton 'R' & bisect OK - rest 2-3 digits

> ======================================================================

> f(x)  =  lambda x: (x - 1.23456789012345e-100) ** 3
> f'(x) =  lambda x: 3*(x - 1.23456789012345e-100) ** 2
> 1.23456789012345e-100
> **3.1208106315091907e-13** N=32 converged optimize.ridder(f, -0.1, 1, full_output=True, maxiter=150)
> **0.0000000000000000e+00** N=2 converged optimize.brentq(f,  0, 1, full_output=True, maxiter=150)
> **0.0000000000000000e+00** N=2 converged optimize.brenth(f,  0, 1, full_output=True, maxiter=150)
> **4.0251719524511793e-13** N=27 converged optimize.toms748(f, -1, 1, full_output=True, maxiter=100)
> **2.7131318051449888e-08** N=39 converged optimize.newton(f, x0=0.2, fprime=fp, full_output=True, maxiter=150)
> **8.1789824356547741e-15** N=80 Tolerance newton(f, fp, 1, 5e-15, 222, 'A')
> 1.234567**9002467461**e-100 N=613 f(c) = 0 newton(f, fp, 1, 5e-15, 800, 'R')
> **1.0948497219942510e-08** N=39 converged optimize.bisect(f, 0, 1, full_output=True)
> 1.23456789**80942788**e-100 N=357 f(c) = 0 bisect(f, 0, 1, 5e-15)

> ======================================================================

> Only my newton 'R' & bisect even have the correct exponent - rest ???
> bisect does not have as DP - why?
> (1.2345678980942788e-100 - 1.23456789012345e-100)**3 = (7.970828653611526e-109)**3 = 0.0
> so the limitations of the computer mathematics make it impossible to do better.
> 
> Suppose that we look at what happens if we reduce the number of digits that we require: 
> 
> lambda x: (x - 1.23456789012345e-100) ** 3
> (1.2345678980942788e-100, 357, 'Converged', 'f(c) = 0') bisect(f, 0, 1, 5e-15)
> (1.2345678980942788e-100, 357, 'Converged', 'f(c) = 0') bisect(f, 0, 1, 5e-8)
> (1.2347947281308757e-100, 344, 'Converged', 'Tolerance') bisect(f, 0, 1, 5e-4)
> (1.2345658202098768e-100, 351, 'Converged', 'Tolerance') bisect(f, 0, 1, 5e-6)
> (1.2345677277758852e-100, 354, 'Converged', 'Tolerance') bisect(f, 0, 1, 5e-7)

> we see that with a rtol of 5e-8 the termination is from f(c) = 0 while at 5e-7 the termination is from 'Tolerance',
> the solution has 7 significant digits.

> =8=====================================================================

> (x - r)*7

> ======================================================================

> f(x)  =  lambda x: (x - 1.23456789012345e1) ** 7
> f'(x) =  lambda x: 7*(x - 1.23456789012345e1) ** 6
> 1.23456789012345e1
> 1.234567890123**5297**e+01 N=33 converged optimize.ridder(f, -0.1, 13, full_output=True, maxiter=150)
> 1.2345678901234**221**e+01 N=118 converged optimize.brentq(f,  0, 13, full_output=True, maxiter=150)
> 1.2345678901234**020**e+01 N=118 converged optimize.brenth(f,  0, 13, full_output=True, maxiter=150)
> 1.2345678901234**409**e+01 N=41 converged optimize.toms748(f, -1, 13, full_output=True, maxiter=100)
> 1.2345678**818647164**e+01 N=122 converged optimize.newton(f, x0=0.2, fprime=fp, full_output=True, maxiter=150)
> 1.2345678901234**475**e+01 N=219 Tolerance newton(f, fp, 1, 5e-15, 222, 'A')
> 1.2345678901234**159**e+01 N=202 Tolerance newton(f, fp, 1, 5e-15, 800, 'R')
> 1.234567890123**5146**e+01 N=43 converged optimize.bisect(f, 0, 13, full_output=True)
> 1.2345678901234500e+01 N=47 f(c) = 0 bisect(f, 0, 13, 5e-15)

> =9=====================================================================

> optimize.newton not good

> ======================================================================

> f(x)  =  lambda x: (x - 1.23456789012345e-1) ** 7
> f'(x) =  lambda x: 7*(x - 1.23456789012345e-1) ** 6
> 1.23456789012345e-1
> 1.23456789012**46490**e-01 N=32 converged optimize.ridder(f, -0.1, 1, full_output=True, maxiter=150)
> 1.23456789012**43828**e-01 N=108 converged optimize.brentq(f,  0, 1, full_output=True, maxiter=150)
> 1.23456789012**14732**e-01 N=110 converged optimize.brenth(f,  0, 1, full_output=True, maxiter=150)
> 1.2345678901**199564**e-01 N=39 converged optimize.toms748(f, -1, 1, full_output=True, maxiter=100)
> 1.234568**7327640877**e-01 N=89 converged optimize.newton(f, x0=0.2, fprime=fp, full_output=True, maxiter=150)
> 1.234567890123**7127**e-01 N=202 Tolerance newton(f, fp, 1, 5e-15, 222, 'A')
> 1.2345678901234**854**e-01 N=215 Tolerance newton(f, fp, 1, 5e-15, 800, 'R')
> 1.2345678901**147039**e-01 N=39 converged optimize.bisect(f, 0, 1, full_output=True)
> 1.23456789012345**03**e-01 N=52 Tolerance bisect(f, 0, 1, 5e-15)

> =10=====================================================================

> optimize.newton not good

> ======================================================================

> f(x)  =  lambda x: (x - 1.23456789012345e-10) ** 7
> f'(x) =  lambda x: 7*(x - 1.23456789012345e-10) ** 6
> 1.23456789012345e-10
> 1.2**427694973418434**e-10 N=33 converged optimize.ridder(f, -0.1, 1, full_output=True, maxiter=150)
> 1.23**11098132712072**e-10 N=60 converged optimize.brentq(f,  0, 1, full_output=True, maxiter=150)
> 1.23**11098122081924**e-10 N=60 converged optimize.brenth(f,  0, 1, full_output=True, maxiter=150)
> 1.23**26028128536769**e-10 N=20 converged optimize.toms748(f, -1, 1, full_output=True, maxiter=100)
> **8.7437677149320988e-08** N=95 converged optimize.newton(f, x0=0.2, fprime=fp, full_output=True, maxiter=150)
> 1.234**8676327403289**e-10 N=202 Tolerance newton(f, fp, 1, 5e-15, 222, 'A')
> 1.2345678901234**818**e-10 N=351 Tolerance newton(f, fp, 1, 5e-15, 800, 'R')
> 1.2**187229003757238**e-10 N=39 converged optimize.bisect(f, 0, 1, full_output=True)
> 1.2345678901234519e-10 N=82 Tolerance bisect(f, 0, 1, 5e-15)

> ======================================================================

> optimize.newton terrible - others bad

> ======================================================================

> f(x)  =  lambda x: (x - 1.23456789012345e-100) ** 7
> f'(x) =  lambda x: 7*(x - 1.23456789012345e-100) ** 6
> 1.23456789012345e-100
> **-7.0727225522258833e-13** N=32 converged optimize.ridder(f, -0.1, 1, full_output=True, maxiter=150)
> **0.0000000000000000e+00** N=1 converged optimize.brentq(f,  0, 1, full_output=True, maxiter=150)
> **0.0000000000000000e+00** N=1 converged optimize.brenth(f,  0, 1, full_output=True, maxiter=150)
> **0.0000000000000000e+00** N=0 converged optimize.toms748(f, -1, 1, full_output=True, maxiter=100)
> **8.7314220414206299e-08** N=95 converged optimize.newton(f, x0=0.2, fprime=fp, full_output=True, maxiter=150)
> **2.9974261691598338e-14** N=202 Tolerance newton(f, fp, 1, 5e-15, 222, 'A')
> **4.9378523656902344e-47** N=691 f(c) = 0 newton(f, fp, 1, 5e-15, 800, 'R')
> **0.0000000000000000e+00** N=1 converged optimize.bisect(f, 0, 1, full_output=True)
> **4.3790577010150533e-47** N=154 f(c) = 0 bisect(f, 0, 1, 5e-15)

> ======================================================================
> None of them are good but it can be seen that both my newton and bisection methods return a value f(c) = 0 because 4.3790577010150533e-47**7 = 0.
> 
> 

>======================================================================

from scipy import optimize
import numpy as np
import math
import re


```
def bisect(f, a, b, tol, bound=9.8813129168249309e-324):
    # ################################################################################E
    # input: Function f,
    #        endpoint values a, b,
    #        tolerance tol, (if tol = 5e-t and bound = 9.0e-324 the function returns
    #                        t significant digits for a root between the minimum normal
    #                        and the maximum normal),
    #         bound (if bound=9.8813129168249309e-324, the algorithm continues until
    #                the interval cannot be further divided, a larger value may result
    #                in termination before t digits are found).
    # conditions: f is a continuous function in the interval [a, b],
    #             a < b,
    #             and f(a)*f(b) < 0.
    # output:    [root, iterations, convergence, termination condition]    #
    # #################################################################################N
    if b <= a:
        return float("NAN"), 0, "No convergence", "b < a"
    fa = f(a)
    fb = f(b)
    if np.sign(fa) == np.sign(fb):
        return float("NAN"), 0, "No convergence", "f(a)*f(b) > 0"
    en = 0
    while en < 2200:
        en += 1
        if np.sign(a) == np.sign(b):  # avoid overflow
            c = a + (b - a)/2
        else:
            c = (a + b)/2
        fc = f(c)
        if b - a <= bound:
            return bound, en, "No convergence", "Bound reached"
        if fc == 0:
            return c, en, "Converged", "f(c) = 0"
        if b - a <= abs(c) * tol:
            return c, en, "Converged", "Tolerance"
        if np.sign(fa) == np.sign(fc):
            a = c
            fa = fc
        else:
            b = c
    return float("NAN"), en, "No convergence", "Bad function"


def newton(f, fprime, pin, eps, N, type):
    n = 1
    p = 0.  # to ensure the value of p carries out of the while loop
    seq = []
    while n <= N:
        seq.append(pin)
        p = pin - f(pin)/fprime(pin)
        if f(p) == 0:
            return p, n, "Converged", "f(c) = 0"
        if type == 'A':
            if np.abs(p-pin) < eps:
                return p, n, "Converged", "Tolerance"
        else:
            if np.abs(p-pin) <= eps*np.abs(p):
                return p, n, "Converged", "Tolerance"
        pin = p
        n += 1

    return p, -n

ilayn:
@EJN38 digit matching is not always going to work in floating point for very small numbers. There is no such thing as arbitrary digit matching, they are limited by the representable values of the magnitude of exact result. 

There is a time and a place for each tolerance method. So one is right the other one is wrong is obviously false. Absolute tolerance is a must in many numerical computation situations. You need to get to concrete examples; so forget about rootfinding let's say there is a correct value of 

 1. 12.345678910111213
 2. 123456789101.11213
 3. 1.2345678910111213e-9

What value do you want to have here in terms of tolerance values applies to set of all values with the same precision and applies to all three cases?

EJN38:
Of course there is a to use the absolute error. If I am trying to match a piston with a cylinder, I need the absolute tolerance.

But that is not what this is about. The objective of root finding is to get a good approximation to the actual value of the root. As can be seen, use of the absolute error produces results that do not even have the appropriate magnitude.

As I have illustrated for functions of the form (x - r)g(x) It is in fact possible to get up to double precision for any in the normal range. If the degree of the root is larger than 1., then some other problems will arise as the function value at c will be 0 before c becomes a good approximation.

I am not quite sure what the intent of your question was but if I consider the absolute error to be 5e-5, then 

|x - 12.345678910111213| < 5e-5 => -5e-5 < x - 12.345678910111213 < 5e-5 or 
x in [12.345**6**28910111213, 12.345**7**28910111212] 

|x - 1.2345678910111213e9| < 5e-5 => -5e-5 < x - 1.2345678910111213e9 < 5e-5 or
x in [1234567891.0110712, 1234567891.0111713]  and this would require a lot of additional iterations,

|x - 1.2345678910111213e-9| < 5e-5 => -5e-5 < x - 1.2345678910111213e-9 < 5e-5 or
x in [-4.999876543210899e-05, 5.0001234567891015e-05] or [-0.00004999876543210899, 0.000050001234567891015]
This means that  -0.00004999876543210900 would be an acceptable 
approximation to 0.0000000012345678910111213

If I consider the relative error to be 5e-5, then 

|x - 12.345678910111213| < 12.345678910111213*5e-5 =>
 -0.0006172839455055606 < x - 12.345678910111213 < 0.0006172839455055606 or 
x in [12.345061626165707, 12.346296194056718] 

|x - 1.2345678910111213e9| < 1.2345678910111213e9*5e-5 => 
-61728.39455055607 < x - 1.2345678910111213e9 < 61728.39455055607 or
x in [1234506162.6165707, 1234629619.4056718]

|x - 1.2345678910111213e-9| < 1.2345678910111213e-9*5e-5 => 
-6.172839455055606e-14 < x - 1.2345678910111213e-9 < 6.172839455055606e-14 or
x in [-1.2345061626165708e-09, 1.2346296194056718e-09]

For getting an accurate approximation to a value relative error is the correct vay.

ilayn:
Yes great so we are on the same page. And in your last example the algorithms typically don't terminate if there is no maxiter set because the algorithm noise is a few order of magnitude higher than eps so you start dancing around the actual value. 

This is the precise reason why, when you get into that zone you use an abstol to stop trying for no reason. So it becomes `tol = max(abstol, rtol*value)`. If the values are large `rtol` does its job, otherwise `atol` take over. 

This is decades old practice in 99% algorithms out there so with no offense I am not going to get into an argument about this practice. Even running the same algorithm on different machines give you different tail of the floating point number. So there is no point in trying to get to that level of precision.

Modern algorithms typically do not work in those regions and try to normalize the problem so computations are always close to 1.0. Hence `rtol` and `atol` practice is rather going out of fashion. Neither of them are correct depending on which region you are working on. But it is very labor intensive to code that way. Hence typically commercial solvers do that kind of work.

Why relative tolerance does not always give you  "that many correct digits" is already well-established and documented; here is the first google result https://blogs.mathworks.com/matlab/2024/03/20/understanding-tolerances-in-ordinary-differential-equation-solvers/

Having said all that, here we already know the default values are problematic so we are trying to explain the current situation not saying we chose the best values historically.

EJN38:
Is there a function  for which my method will fail but the SciPy method will produce a better result?

Issue #19953, opened 2024-01-23
TST: We should add missing tests for cdflib


steppi:
Since merging https://github.com/scipy/scipy/pull/19560,  @ilayn's translation of cdflib into cython, several bugs have been brought to our attention by statsmodels maintainer @bashtage, https://github.com/scipy/scipy/issues/19896, https://github.com/scipy/scipy/pull/19933, https://github.com/scipy/scipy/pull/19944, who kindly made PRs fixing several of them.

Currently `scipy/special/tests/test_cdflib.py` is missing tests for a number of functions

https://github.com/scipy/scipy/blob/11ce9d798116092189a25d3799a6bcecfdedbc53/scipy/special/tests/test_cdflib.py#L4-L19

I think there are two primary types of tests we can add. Tests against mpmath, (which will require writing arbitrary precision implementations of these functions using mpmath), and tests against the Fortran version (which would be faster, but it won't catch bugs in the Fortran version).

@ilayn offered to add tests if I came up the parameter domains which in which the functions should be tested. I started making such a list, and realized that such special care around hot spots probability isn't necessary just yet. I think we should just use the classes `Arg`, `IntArg`, and `ProbArg` used in `test_cdflib.py` over the domains of definition of the functions. When mpmath is used, we should restrict the domains for cases where mpmath cannot compute a correct answer for sufficiently large values.

While going through all of these, I found that the functions for the non-Central F distribution, `ncfdtr` and friends are documented to require the non-centrality parameter `nc` to in $(0, 1e^4)$. The comment in `_cdflib.pyx` refers to the bound too. I've found the bound is missing from the code though. @ilayn, do you remember if there was a reason to take the bound out? I can't remember if that came up in the review. We'll either need to put the bound back or change the docs.

Bounds below.

### Non-central F distribution
* `ncfdtr`, `ncfdtri`, `ncfdtridfn`, `ncfdtridfd`, and `ncfdtrinc`.
    * `dfn`: `Arg(0, np.inf, inclusive_a=False)`
    * `dfd`: `Arg(0, np.inf, inclusive_a=False)`
    * `nc`: `Arg(0, 1e4)` This is the documented range at least, but it looks like this check has been removed.
    * `f`: `Arg(0, np.inf)`
    * `p`: `ProbArg()`.

### Negative binomial distribution
* `nbdtrik` and `nbdtrin`
    * `k`: `IntArg(1, 1000)`
    * `y`: `ProbArg()`
[...]

PR #22719, opened 2025-03-20
TST: add tests for `ncfdtri`

dschmitz89:
#### Reference issue
Towards #19953

Adds accuracy tests for `ncfdtri` using the mpmath references that were already available from the tests for `ncfdtr`. Originally, I had planned to add tests for the other inverses of `ncfdtr` but they do not pass half of these test cases as they are still backed by cdflib.

### DIFF
--- a/scipy/special/tests/test_cdflib.py
+++ b/scipy/special/tests/test_cdflib.py
@@ -3,7 +3,6 @@ Test cdflib functions versus mpmath, if available.
 
 The following functions still need tests:
 
-- ncfdtri
 - ncfdtridfn
 - ncfdtridfd
 - ncfdtrinc
@@ -491,7 +490,7 @@ def test_bdtrik_nbdtrik_inf():
 
 
 @pytest.mark.parametrize(
-    "dfn,dfd,nc,f,expected",
+    "dfn,dfd,nc,f,expected_cdf",
     [[100.0, 0.1, 0.1, 100.0, 0.29787396410092676],
      [100.0, 100.0, 0.01, 0.1, 4.4344737598690424e-26],
      [100.0, 0.01, 0.1, 0.01, 0.002848616633080384],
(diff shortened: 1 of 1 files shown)


scipy/special/tests/test_cdflib.py
```diff
@@ -548,8 +547,11 @@ def test_ncfdtr(dfn, dfd, nc, f, expected):
     # rng = np.random.default_rng(1234)
     # sample_idx = rng.choice(len(re), replace=False, size=12)
     # cases = np.array(cases)[sample_idx].tolist()
-    assert_allclose(sp.ncfdtr(dfn, dfd, nc, f), expected, rtol=1e-13, atol=0)
-
+    assert_allclose(sp.ncfdtr(dfn, dfd, nc, f), expected_cdf, rtol=1e-13, atol=0)
+    # testing tails where the CDF reaches 0 or 1 does not make sense for inverses
+    # of a CDF as they are not bijective in these regions
+    if 0 < expected_cdf < 1:
+        assert_allclose(sp.ncfdtri(dfn, dfd, nc, expected_cdf), f, rtol=5e-11)
```
tylerjereddy, line 554: I suppose we could wait for Albert to comment, but given that no algorithmic code is changing and the numerical stringency of the original testing remains the same (just a new test case is added)--this is probably pretty safe in any case.

I looked at the few CI failures--they're all unrelated and were recently fixed in `main`.

tylerjereddy:
I took another look at this this evening, and my assessment from > 1 month ago at https://github.com/scipy/scipy/pull/22719#pullrequestreview-2748577487 remains the same, with no additional input from the team during that time. I double checked that the rebased version of this feature branch passes the full testsuite locally on ARM Mac.

We have many PRs waiting to go in for the next release, and this is a fairly safe one to put in with extra testing only.

PR #22721, opened 2025-03-21
MAINT: bump musl image

andyfaff:
Bumps the musl image used in CI. Corresponds to the default cibuildwheel image, and the image used for building numpy wheels.


andyfaff:
https://cibuildwheel.pypa.io/en/stable/options/#linux-image

andyfaff:
Note, the musl test with the older image meant that a numpy wheel would have to be built from source. With the newer image we can simply use the wheel.

PR #22724, opened 2025-03-21
MAINT: integrate: break dependency on optimize

j-bowhay:
Follow up https://github.com/scipy/scipy/pull/22671

Currently, `scipy.integrate` depends on `scipy.optimize` as some functions return instances of `OptimizeResult`. This seems a little unnecessary to import an entire module just for a result object! Instead, we can move `OptimizeResult` right next to `_RichResult` (which it is a subclass of) in `_lib` which both modules already import.

### DIFF
--- a/scipy/_lib/_util.py
+++ b/scipy/_lib/_util.py
@@ -1189,6 +1189,53 @@ class _RichResult(dict):
         return list(self.keys())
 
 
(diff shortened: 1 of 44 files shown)


j-bowhay:
Grrr my quick search failed me, there are uses that cannot be removed

lucascolley:
oh well, at least the CI check is working!

PR #22728, opened 2025-03-22
ENH/BUG: optimize: use `_differentiable_functions.VectorFunction` in `least_squares`

andyfaff:
Updates the internals of `optimize.least_squares` to use `optimize._differentiable_functions.VectorFunction`.

The `lm` method now uses `VectorFunction` to calculate numeric derivatives instead of relying on `_minpack._lmder`. Consequently the `lm` method is now able to take advantage of various features of `approx_derivative` (used by `VectorFunction`). These include the `workers` keyword (parallelise the numerical differentiation), any of the `2-point`/`3-point`/`cs` steps. Down the line this may offer the potential to use LinearOperator, jac_sparsity. `approx_derivative` is also more accurate than the FD done in `_minpack._lmdif`.

The other two methods (`trf`/`dogbox`) also use `VectorFunction`. The `VectorFunction` class offers memoisation of function and jacobian, which can reduce the total number of their evaluations, although `trf` and `dogbox` were careful about their design in that regard.

Finally some bug fixes (introduced in #22549) are made to `VectorFunction`.

### DIFF
--- a/scipy/optimize/_differentiable_functions.py
+++ b/scipy/optimize/_differentiable_functions.py
@@ -684,11 +684,17 @@ class VectorFunction:
     def fun(self, x):
         self._update_x(x)
         self._update_fun()
-        return self.f
+        # returns a copy so that downstream can't overwrite the
+        # internal attribute
+        return self.f.copy()
 
     def jac(self, x):
         self._update_x(x)
         self._update_jac()
+        if hasattr(self.J, "copy"):
+            # returns a copy so that downstream can't overwrite the
+            # internal attribute. But one can't copy a LinearOperator
+            return self.J.copy()
         return self.J
 
     def hess(self, x, v):
@@ -696,6 +702,10 @@ class VectorFunction:
         self._update_v(v)
         self._update_x(x)
         self._update_hess()
+        if hasattr(self.H, "copy"):
+            # returns a copy so that downstream can't overwrite the
+            # internal attribute. But one can't copy non-arrays
+            return self.H.copy()
         return self.H
 
 
--- a/scipy/optimize/_differentiable_functions.py
+++ b/scipy/optimize/_differentiable_functions.py
@@ -429,7 +429,8 @@ class VectorFunction:
     """
     def __init__(self, fun, x0, jac, hess,
                  finite_diff_rel_step=None, finite_diff_jac_sparsity=None,
-                 finite_diff_bounds=(-np.inf, np.inf), sparse_jacobian=None, workers=None):
+                 finite_diff_bounds=(-np.inf, np.inf), sparse_jacobian=None,
+                 workers=None):
         if not callable(jac) and jac not in FD_METHODS:
             raise ValueError(f"`jac` must be either callable or one of {FD_METHODS}.")
 
--- a/scipy/optimize/_differentiable_functions.py
+++ b/scipy/optimize/_differentiable_functions.py
@@ -545,9 +545,9 @@ class VectorFunction:
                     sparse_jacobian is None and sps.issparse(self.J)):
                 def update_jac():
                     self._update_fun()
-                    self.J, dct = sps.csr_array(
-                        approx_derivative(fun_wrapped, self.x, f0=self.f,
-                                          **finite_diff_options))
+                    _J, dct = approx_derivative(fun_wrapped, self.x, f0=self.f,
+                                                **finite_diff_options)
+                    self.J = sps.csr_array(_J)
                     self.nfev += dct['nfev']
                 self.J = sps.csr_array(self.J)
                 self.sparse_jacobian = True
@@ -555,8 +555,9 @@ class VectorFunction:
             elif sps.issparse(self.J):
                 def update_jac():
                     self._update_fun()
-                    self.J, dct = approx_derivative(fun_wrapped, self.x, f0=self.f,
-                                                    **finite_diff_options).toarray()
+                    _J, dct = approx_derivative(fun_wrapped, self.x, f0=self.f,
+                                                    **finite_diff_options)
+                    self.J = _J.toarray()
                     self.nfev += dct['nfev']
                 self.J = self.J.toarray()
                 self.sparse_jacobian = False
--- a/scipy/optimize/_differentiable_functions.py
+++ b/scipy/optimize/_differentiable_functions.py
@@ -429,7 +429,7 @@ class VectorFunction:
     """
     def __init__(self, fun, x0, jac, hess,
                  finite_diff_rel_step=None, finite_diff_jac_sparsity=None,
-                 finite_diff_bounds=None, sparse_jacobian=None, workers=None):
+                 finite_diff_bounds=(-np.inf, np.inf), sparse_jacobian=None, workers=None):
         if not callable(jac) and jac not in FD_METHODS:
             raise ValueError(f"`jac` must be either callable or one of {FD_METHODS}.")
 
--- a/scipy/optimize/_differentiable_functions.py
+++ b/scipy/optimize/_differentiable_functions.py
@@ -451,7 +451,7 @@ class VectorFunction:
         if xp.isdtype(_x.dtype, "real floating"):
             _dtype = _x.dtype
 
-        # promotes to floating
+        # promotes to floating, ensures that it's a copy
         self.x = xp.astype(_x, _dtype)
         self.x_dtype = _dtype
 
@@ -497,7 +497,7 @@ class VectorFunction:
 
         def update_fun():
             self.nfev += 1
-            self.f = fun_wrapped(self.x)
+            self.f = fun_wrapped(xp.astype(self.x, _dtype))
 
         self._update_fun_impl = update_fun
         update_fun()
@@ -507,7 +507,7 @@ class VectorFunction:
 
         # Jacobian Evaluation
         if callable(jac):
-            self.J = jac(self.x)
+            self.J = jac(xp.astype(self.x, _dtype))
             self.J_updated = True
             self.njev += 1
 
@@ -539,7 +539,7 @@ class VectorFunction:
                 self.sparse_jacobian = False
 
             def update_jac():
-                self.J = jac_wrapped(self.x)
+                self.J = jac_wrapped(xp.astype(self.x, _dtype))
 
         elif jac in FD_METHODS:
             self.J, dct = approx_derivative(fun_wrapped, self.x, f0=self.f,
@@ -562,7 +562,7 @@ class VectorFunction:
                 def update_jac():
                     self._update_fun()
                     _J, dct = approx_derivative(fun_wrapped, self.x, f0=self.f,
-                                                    **finite_diff_options)
+                                                **finite_diff_options)
                     self.J = _J.toarray()
                     self.nfev += dct['nfev']
                 self.J = self.J.toarray()
@@ -582,7 +582,7 @@ class VectorFunction:
 
         # Define Hessian
         if callable(hess):
-            self.H = hess(self.x, self.v)
+            self.H = hess(xp.astype(self.x, _dtype), self.v)
             self.H_updated = True
             self.nhev += 1
 
@@ -604,10 +604,10 @@ class VectorFunction:
                 self.H = np.atleast_2d(np.asarray(self.H))
 
             def update_hess():
-                self.H = hess_wrapped(self.x, self.v)
+                self.H = hess_wrapped(xp.astype(self.x, _dtype), self.v)
         elif hess in FD_METHODS:
             def jac_dot_v(x, v):
-                return jac_wrapped(x).T.dot(v)
+                return jac_wrapped(xp.astype(x, _dtype)).T.dot(v)
 
             def update_hess():
                 self._update_jac()
@@ -686,15 +686,17 @@ class VectorFunction:
         self._update_fun()
         # returns a copy so that downstream can't overwrite the
         # internal attribute
-        return self.f.copy()
+        xp = array_namespace(self.f)
+        return xp.astype(self.f, self.f.dtype)
 
     def jac(self, x):
         self._update_x(x)
         self._update_jac()
-        if hasattr(self.J, "copy"):
+        if hasattr(self.J, "astype"):
             # returns a copy so that downstream can't overwrite the
             # internal attribute. But one can't copy a LinearOperator
-            return self.J.copy()
+            xp = array_namespace(self.J)
+            return xp.astype(self.J, self.J.dtype)
         return self.J
 
     def hess(self, x, v):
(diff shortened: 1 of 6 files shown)


andyfaff:
As an example of the memoisation

```python
import numpy as np
from numpy.random import default_rng

def gen_data(t, a, b, c, noise=0., n_outliers=0, seed=None):
    rng = default_rng(seed)
    y = a + b * np.exp(t * c)
    error = noise * rng.standard_normal(t.size)
    outliers = rng.integers(0, t.size, n_outliers)
    error[outliers] *= 10
    return y + error

a = 0.5
b = 2.0
c = -1
t_min = 0
t_max = 10
n_points = 15

t_train = np.linspace(t_min, t_max, n_points)
y_train = gen_data(t_train, a, b, c, noise=0.1, n_outliers=3, seed=18)

class Fun:
    def __init__(self, t, y):
        self.nfev = 0
        self.t = t
        self.y = y
        
    def __call__(self, x):
        self.nfev += 1
        return x[0] + x[1] * np.exp(x[2] * self.t) - self.y
        
x0 = np.array([1.0, 1.0, 0.0])
fun = Fun(t_train, y_train)

res = least_squares(fun, x0, method='lm')
print(fun.nfev)
```

gives 109 for main and 104 for this PR.

nickodell:
Here's a suggested patch which improves test coverage around `sparse_jacobian=False`:

```diff
diff --git a/scipy/optimize/tests/test_differentiable_functions.py b/scipy/optimize/tests/test_differentiable_functions.py
index 5c16672f7..619abc8ec 100644
--- a/scipy/optimize/tests/test_differentiable_functions.py
+++ b/scipy/optimize/tests/test_differentiable_functions.py
@@ -818,6 +818,44 @@ class TestVectorialFunction(TestCase):
         res = vf.jac(x0)
         assert res.dtype == np.float32
 
+    def test_sparse_analytic_jac(self):
+        ex = ExVectorialFunction()
+        x0 = np.array([1.0, 0.0])
+        def sparse_adapter(func):
+            def inner(x):
+                f_x = func(x)
+                return csr_array(f_x)
+            return inner
+
+        # jac(x) returns dense jacobian
+        vf1 = VectorFunction(ex.fun, x0, ex.jac, ex.hess, None, None,
+                            (-np.inf, np.inf), sparse_jacobian=None)
+        # jac(x) returns sparse jacobian, but sparse_jacobian=False requests dense
+        vf2 = VectorFunction(ex.fun, x0, sparse_adapter(ex.jac), ex.hess, None, None,
+                            (-np.inf, np.inf), sparse_jacobian=False)
+
+        res1 = vf1.jac(x0 + 1)
+        res2 = vf2.jac(x0 + 1)
+        assert_equal(res1, res2)
+
+    def test_sparse_numerical_jac(self):
+        ex = ExVectorialFunction()
+        x0 = np.array([1.0, 0.0])
+        N = len(x0)
+
+        # normal dense numerical difference
+        vf1 = VectorFunction(ex.fun, x0, '2-point', ex.hess, None, None,
+                            (-np.inf, np.inf), sparse_jacobian=None)
+        # use sparse numerical difference, but ask it to be converted to dense
+        finite_diff_jac_sparsity = csr_array(np.ones((N, N)))
+        vf2 = VectorFunction(ex.fun, x0, '2-point', ex.hess, None, finite_diff_jac_sparsity,
+                            (-np.inf, np.inf), sparse_jacobian=False)
+
+        res1 = vf1.jac(x0 + 1)
+        res2 = vf2.jac(x0 + 1)
+        assert_equal(res1, res2)
+
+
 
 def test_LinearVectorFunction():
     A_dense = np.array([
```

scipy/optimize/_differentiable_functions.py
```diff
@@ -545,18 +550,19 @@ def update_jac():
...
                 def update_jac():
                     self._update_fun()
-                    self.J, dct = approx_derivative(fun_wrapped, self.x, f0=self.f,
-                                                    **finite_diff_options).toarray()
+                    _J, dct = approx_derivative(fun_wrapped, self.x, f0=self.f,
+                                                    **finite_diff_options)
+                    self.J = _J.toarray()
                     self.nfev += dct['nfev']
                 self.J = self.J.toarray()
                 self.sparse_jacobian = False
```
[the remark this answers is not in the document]
andyfaff: The bug got picked up by the `least_squares` tests.

andyfaff:
Would you like to submit that patch against my PR, so you get credit?

scipy/optimize/_lsq/least_squares.py
```diff
@@ -286,8 +299,7 @@ def least_squares(
         twice as many operations as '2-point' (default). The scheme 'cs'
         uses complex steps, and while potentially the most accurate, it is
         applicable only when `fun` correctly handles complex inputs and
-        can be analytically continued to the complex plane. Method 'lm'
-        always uses the '2-point' scheme. If callable, it is used as
+        can be analytically continued to the complex plane If callable, it is used as
```
andyfaff, line 302:
```suggestion
        can be analytically continued to the complex plane. If callable, it is used as
```

andyfaff:
Added release notes.

nickodell:
I spent a few hours yesterday and today testing this, and I did find one thing, though I'm not sure if you'd consider it a bug.

Specifically, lmder can corrupt the state of VectorFunction.f, because one of the temporary workspaces given to lmder is the return value of VectorFun.fun(). This return value can be given out to multiple callers, so if you call `f0 = vector_fun.fun(x0)`, the array with that value will later be changed by lmder.

You can see this by adding the following print statements:

```diff
diff --git a/scipy/optimize/_lsq/least_squares.py b/scipy/optimize/_lsq/least_squares.py
index 939763f2c..663ab6e37 100644
--- a/scipy/optimize/_lsq/least_squares.py
+++ b/scipy/optimize/_lsq/least_squares.py
@@ -978,8 +978,10 @@ def least_squares(
         if callback is not None:
             warn("Callback function specified, but not supported with `lm` method.",
                  stacklevel=2)
+        print(f0)
         result = call_minpack(vector_fun.fun, x0, vector_fun.jac, ftol, xtol, gtol,
                               max_nfev, x_scale, jac_method=jac)
+        print(f0)
         # VectorFunction tracks number of times fun was actually called (it tries to
         # memoise where possible), including numerical differentiation.
         # Tesult object needs to be patched with the correct number.
```

The result of this is that f0 changes, even though it wasn't a parameter. You can also see this by looking at the pointers of the arrays within VectorFunction.fun and minpack_lmder.

I haven't been able to come up with a user-visible scenario where this could cause a bug, but it seems very "spooky action at a distance."

andyfaff:
Hmm. That's something that might need to be changed in the C code @ilayn. Another workaround is for VectorFunction to return a copy of the array.

ilayn:
I haven't touched the header files and they are coming all the way back from Travis, but here is probably why it happens

https://github.com/scipy/scipy/blob/06e82a4737fc4feac0edd710498e28924bf626ab/scipy/optimize/__minpack.h#L680-L682

So if x0 is overwritten then this https://numpy.org/doc/stable/reference/c-api/array.html#c.PyArray_ContiguousFromObject is not correct as it does not return a new copy. 

However if `wrapped_fun` is passing `x0` through then `f0` shares the data with `x0`. So it gets modified via via. Probably needs a `.copy()` somewhere if this is an issue.

andyfaff:
I think it has something to do with `VectorFunction.fun(x)` returning an array that is actually an attribute of `VectorFunction`. The `jac` method does the same thing. Returning a copy in VectorFunction is simple to do.

ilayn:
@nickodell Do you get a changed result in `x0` if you modify the output array `x`? Because it seems like it should from your description.

ilayn:
> Returning a copy in VectorFunction is simple to do.

Also possible in that function with the flag [NPY_ARRAY_ENSURECOPY](https://numpy.org/doc/stable/reference/c-api/array.html#c.NPY_ARRAY_ENSURECOPY).  Might be a tiny bit faster.

andyfaff:
It's probably a good idea for `VectorFunction` to return a copy.

nickodell:
>@nickodell Do you get a changed result in x0 if you modify the output array x? Because it seems like it should from your description.

I do. If I make this change:

```
index 0c1b7f099..bbe002454 100644
--- a/scipy/optimize/_lsq/least_squares.py
+++ b/scipy/optimize/_lsq/least_squares.py
@@ -67,6 +67,11 @@ def call_minpack(fun, m, x0, jac, ftol, xtol, gtol, max_nfev, x_scale, jac_metho
     x, info, status = _minpack._lmder(
         fun, jac, m, n, x0, (), full_output, col_deriv,
         ftol, xtol, gtol, max_nfev, factor, diag)
+    print(x0)
+    x[0] = 0
+    print(x0)
+    #print("x0", x0.__array_interface__['data'][0])
+    #print("x", x.__array_interface__['data'][0])
 
     f = info['fvec']
     J = jac(x)
```

The result is that changing x[0] also changes x0[0].

>I haven't touched the header files and they are coming all the way back from Travis, but here is probably why it happens

I think it's a couple lines down from that:

```
  /* Setup array to hold the function evaluations */
  ap_fvec = (PyArrayObject *)call_python_function(fcn, n, x, extra_args, 1, minpack_error, -1);
  if (ap_fvec == NULL) goto fail;
  fvec = (double *) PyArray_DATA(ap_fvec);
```

In other words, it calls fun(), then uses the returned array as the fvec parameter. Then lmder modifies fvec to store f(x) of whatever x it is looking at.

This seems strange for three reasons:

1) It's calling the user function to discover how large its return value is, but the caller to this function has already done this as part of validating the arguments to this function. The caller already knows how big fvec ought to be.
2) fvec is an output parameter, so providing the value at f(x) doesn't do anything. You could use zeros just as well.
3) It might be memory that somebody else is using for something.

I had an idea that lmder could be fixed by asking the caller to provide the dimensions of the function/jacobian, and allocating a new fvec array at the start. Every caller to lmder already knows this, so it is not too difficult to provide this information

I made a PR here: https://github.com/andyfaff/scipy/pull/79

ilayn:
> I think it's a couple lines down from that:

Yes but if `x` is not copied inside func but spits out another array that shares data with x, then they are all connected back to x0 which is funny at the very least. 

> This seems strange for three reasons: [...] 

Keep in mind that when these were written (24 years ago to be exact) python/numpy C apis were a different monster and much has changed. So if you think there are better ways to handle all this, please go ahead. Let me know if I can be of any help.

nickodell:
> Yes but if `x` is not copied inside func but spits out another array that shares data with x, then they are all connected back to x0 which is funny at the very least.

Yeah, that is wacky.

Reading the NumPy docs, they suggest that PyArray_ContiguousFromObject is like PyArray_FromAny with NPY_ARRAY_DEFAULT and NPY_ARRAY_ENSUREARRAY flags set.

With that in mind, how about something like this:

```diff
diff --git a/scipy/optimize/__minpack.h b/scipy/optimize/__minpack.h
index bdb1ab93a..b57a5258d 100644
--- a/scipy/optimize/__minpack.h
+++ b/scipy/optimize/__minpack.h
@@ -680,7 +680,7 @@ static PyObject *minpack_lmder(PyObject *dummy, PyObject *args) {
   INIT_JAC_FUNC(fcn,Dfun,extra_args,col_deriv,minpack_error);
 
   /* Initial input vector */
-  ap_x = (PyArrayObject *)PyArray_ContiguousFromObject(x0, NPY_DOUBLE, 1, 1);
+  ap_x = (PyArrayObject *)PyArray_FromAny(x0, PyArray_DescrFromType(NPY_DOUBLE), 1, 1, NPY_ARRAY_DEFAULT | NPY_ARRAY_ENSUREARRAY | NPY_ARRAY_ENSURECOPY, NULL);
   if (ap_x == NULL) goto fail;
   x = (double *) PyArray_DATA(ap_x);
   if (n != PyArray_DIMS(ap_x)[0]) goto fail;

```

nickodell (approved):

ilayn:
> With that in mind, what do you think of a fix like this?

Sounds good to me but I'm a bit confused why it did not work as is.

andyfaff:
I don't think the overwriting will occur now as the code sends array copies to minpack.

I think this PR is mergeable now.

andyfaff:
If there are no further objections I will merge this tomorrow.

nickodell:
I'm happy with merging this PR as-is.

ilayn (approved):

andyfaff:
On a final walkthrough I decided that the behaviour around `max_nfev` and `result.nfev` was confusing. Partially due to history, and partially due to this refactor.

### Historically

- The `max_nfev` and `result.nfev` for both 'trf' and 'dogbox' do not include function evaluations consumed during any numerical approximation of the jacobian. default max_nfev is 100*n. i.e. `result.nfev` is the same for `'2-point'` and `callable(jac)`.
- For 'lm' with `callable(jac)`, the situation is the same as 'trf' and 'dogbox'.
- For 'lm' with numerical approximation of the jacobian the `max_nfev` and `result.nfev` included function evaluations consumed during numerical approximation of the jacobian. default max_nfev is 100 * n * (n + 1).

### This PR
Everything is brought into line.
- The `max_nfev` and `result.nfev` for 'trf', 'dogbox', 'lm' do not include function evaluations consumed during numerical approximation of the jacobian. default max_nfev is 100*n.

This may catch out end users who were setting max_nfev to be larger for 'lm' with numerical approximation of jacobian, and who were expecting `result.nfev` to be correspondingly larger. I could've kept the historical behaviour of `result.nfev`, but there was no way to replicate the before/after behaviour of `max_nfev` (`max_nfev` means different things for lmdif compared to lmder, and now we're only using lmder).
The sensible option to me was to unify the behaviour and provide explanation to the end user.

scipy/optimize/_lsq/least_squares.py
```diff
@@ -390,13 +405,12 @@ def least_squares(
...
+        For all methods this parameter controls the maximum number of function
+        evaluations separate to those used in numerical approximation of the jacobian.
+        If None (default), the value is chosen automatically as 100 * n.
 
-        * For 'trf' and 'dogbox' : 100 * n.
-        * For 'lm' :  100 * n if `jac` is callable and 100 * n * (n + 1)
-          otherwise (because 'lm' counts function calls in Jacobian
-          estimation).
+        .. versionchanged:: 1.16.0
+            The default for the 'lm' method is changed to 100 * n.
```
andyfaff, line 413:
```suggestion
            The default for the 'lm' method is changed to 100 * n, for both a callable and
            a numerically estimated jacobian. Previously the default when using an estimated
            jacobian was 100 * n * (n + 1), because the method included evaluations used in
            the estimation. 
```

rgommers:
This PR seems to have broken something and uncovered a gap in our CI - see gh-22774.

Issue #21972, opened 2024-11-29
STY: fix and enable lint rule UP031


lucascolley:
`ruff check . --target-version=py310 --select UP031` currently yields 346 errors with `ruff 0.8.1`. I disabled the rule in gh-21973.

To close this issue, someone should:

- In a first commit, remove `, "UP031"` from https://github.com/scipy/scipy/blob/4c83de1632ef2deeb44cf90738c427358e405ee9/tools/lint.toml#L23
- run `ruff check . --target-version=py310 --select UP031`
- fix all of the errors in a second commit

You may want to only fix some of the errors and ask for feedback via a draft PR first!

lucascolley:
Closed by gh-21987

PR #21987, opened 2024-12-01
STY: fix lint rule UP031

Transurgeon:
#### Reference issue

#21972
#### What does this implement/fix?

This PR aims to tackle the issue above by replacing printf strings with either f-strings or using the ``.format()`` method.
#### Additional information

There are over 200 more errors left. I will try to fix them over the upcoming week.

PR #22730, opened 2025-03-22
STY: Fixed UP031 errors

jazman7:
- Ran the unit tests and only one failed outside of what I touched

#### Reference issue
Closes  Issue #21972

Changed to f-strings instead of old % formatting

#### Additional information
There was 1 unit test that failed, but in code I did not touch

### DIFF
--- a/scipy/_build_utils/tempita/_looper.py
+++ b/scipy/_build_utils/tempita/_looper.py
@@ -41,8 +41,7 @@ class looper:
         return looper_iter(self.seq)
 
     def __repr__(self):
-        return '<%s for %r>' % (
-            self.__class__.__name__, self.seq)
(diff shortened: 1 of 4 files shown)


lucascolley:
Ah, sorry @jazman7, that issue should have been closed by gh-21987. That's my mistake that it was still open.

The files that you have touched here are not checked by the linter's default config, see lines 1-5 and 30 of https://github.com/scipy/scipy/blob/main/tools/lint.toml. The files under `_lib` are external code that we just 'vendor' in SciPy, and the mmio file has some tricky sections in it that we decided were best to ignore.

Thank you for your work and the PR regardless! You're more than welcome to work on another open issue. Apologies again for wasting your time here.

PR #22733, opened 2025-03-23
MAINT: migrate `gdtria` from cdflib to cephes and scipy's own rootfinder

dschmitz89:
#### Reference issue
Towards #21966

Refactors `gdtria` in the same way as `gdtrib` was reimplemented in #21454

#### Additional information
Edge cases are not really handled yet, opening for initial feedback and to get familiarized with the new root finding machinery.

### DIFF
--- a/scipy/special/cython_special.pyx
+++ b/scipy/special/cython_special.pyx
@@ -1116,6 +1116,7 @@ cdef extern from r"xsf_wrappers.h":
     double special_ber(double) nogil
     double special_berp(double) nogil
     double xsf_gdtrib(double, double, double) nogil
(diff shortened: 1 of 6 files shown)


steppi:
`gdtria` actually doesn't invert `gdtr` it can be implemented directly in terms of the inverse of the incomplete gamma function. I discovered this while I was working on the `gdtrib` PR. Using the rootfinder will be inferior to the more direct implementation.

https://github.com/scipy/scipy/blob/06e82a4737fc4feac0edd710498e28924bf626ab/scipy/special/cdflib.c#L3244-L3271

PR #22735, opened 2025-03-23
ENH: stats: lazy trimmed stats for Dask and JAX

crusaderky:
Run `scipy.stats.tmean`, `tvar`, `tstd`, `tsem`, `tmin`, `tmax` in jax.jit and prevent spurious computation in Dask.

tmin and tmax have a data-dependent output dtype: when the input is integer, the output is integer if the output doesn't need to contain NaNs; otherwise it's changed to float.
This PR makes all lazy backends always return float from these two functions. This includes JAX in eager mode, deliberately, as it is understood by the jax maintainers that eager mode should mostly be used to debug jitted mode so it's best to keep it as similar as possible here.


mdhaber:
Would it be simpler if these functions were to just raise rather than switching to float and returning NaN?

crusaderky:
> Would it be simpler if these functions were to just raise rather than switching to float and returning NaN? You were concerned about the loss of precision in [gh-22626](https://github.com/scipy/scipy/issues/22626).

Yes, that is a concern here too. However lazy arrays _can't_ raise based on array values without materializing the graph - which is destructive for performance in Dask and leads to undefined behaviour in JAX.

One option would be to add a `fill_value=math.nan` optional parameter and let `where` do its type broadcasting, which would happen only in case of error for eager backends and at all times for lazy ones. So eager backends would benefit from it too. It may also be a good idea to issue a warning whenever this causes the dtype to change from int to float, again on all backends?

mdhaber:
> However lazy arrays can't raise based on array

Right. Hopefully at some point that will become second nature. 

Actually, then, I'd rather not give extra attention to this right now. I will take it into consideration when working on the function to replace `mstats.trim`, which will become a better way of working with trimmed data.

crusaderky:
FWIW, this is ready for merge.

mdhaber:
I'll take another look. I'm having a hard time seeing what changed with the force pushes since it's obscured by merges. Anything notable? 
Removing draft status in case someone gets to it first.

scipy/stats/tests/test_stats.py
```diff
@@ -3698,8 +3725,9 @@ def test_skew_propagate_nan(self, xp):
 
     def test_skew_constant_value(self, xp):
         # Skewness of a constant input should be NaN (gh-16061)
-        with pytest.warns(RuntimeWarning, match="Precision loss occurred"):
-            a = xp.asarray([-0.27829495]*10)  # xp.repeat not currently available
+        a = xp.asarray([-0.27829495]*10)  # xp.repeat not currently available
+
+        with eager_warns(a, RuntimeWarning, match="Precision loss occurred"):
```
mdhaber, line 3730: Hmm, I see that only *one* of these has to warn for the test to pass. Looks like that's not what we want. Just thought I'd mention it, but no need to fix it here, since it's not new with this PR.

scipy/stats/tests/test_stats.py
```diff
@@ -9596,6 +9630,12 @@ def test_nan_policy(self, xp):
         ref = xp.var(x[~mask])
         xp_assert_close(res, ref)
 
+    @skip_xp_backends(eager_only=True)
+    def test_nan_policy_warns(self, xp):
+        x = xp.arange(10.)
+        mask = (x == 3)
+        x = xp.where(mask, xp.nan, x)
```
mdhaber, line 9637: NBD, but I saw you did it above.
```suggestion
        x = xp.where(x == 3, xp.nan, x)
```

scipy/stats/tests/test_stats.py
```diff
@@ -74,15 +75,38 @@
...
+    @staticmethod
+    def lazy_expect(expect, xp):
+        """The output dtype of tmin/tmax is data-dependent: int inputs are cast to
+        float if there are any NaNs. On lazy backends (JAX/Dask), since content tests
+        are impossible, int inputs are always cast to float.
+        """
+        expect = xp.asarray(expect)
+        if is_lazy_array(expect) and xp.isdtype(expect, "integral"):
+            return xp.astype(expect, xp.float64)
+        return expect
```
mdhaber, line 107: What do you think about changing `test_tmin` and `test_tmax` to work with floating point input and adding one short, separate test that confirms that integer input still works? Then, I think this logic in this function could go right in the new test, instead.

mdhaber (approved):
LGTM; just a few thoughts.

scipy/stats/tests/test_stats.py
crusaderky: This is a great idea

scipy/stats/_stats_py.py
```diff
@@ -805,11 +806,15 @@ def tmin(a, lowerlimit=None, axis=0, inclusive=True, nan_policy='propagate'):
...
-    # Output dtype is data-dependent
-    # Possible loss of precision for int types
-    res = min_ if xp.all(valid) else xp.where(valid, min_, xp.nan)
+    res = xp.min(a, axis=axis)
+    invalid = xp.all(mask, axis=axis)  # All elements are below lowerlimit
+
+    if is_lazy_array(invalid) or xp.any(invalid):
+        # For eager backends, output dtype is data-dependent
+        # Possible loss of precision for int types
+        res, = xp_broadcast_promote(res, force_floating=True, xp=xp)
```
mdhaber, line 815: Looks good, but I'd like to get gh-22695 merged first to avoid the conflict (`xp_broadcast_promote` will no longer exist).

lucascolley (approved):
thanks Guido!

rgommers:
This broke the free-threading parallel CI job, with failures that are also visible on this PR:

```
 FAILED scipy/stats/tests/test_stats.py::TestZscore::test_zscore_constant_input_1d[numpy] - Failed: DID NOT WARN. No warnings of type (<class 'RuntimeWarning'>,) were emitted.
 Emitted warnings: [].
FAILED scipy/stats/tests/test_stats.py::TestZscore::test_zscore_constant_input_2d[numpy] - Failed: DID NOT WARN. No warnings of type (<class 'RuntimeWarning'>,) were emitted.
 Emitted warnings: [].
FAILED scipy/stats/tests/test_stats.py::TestZscore::test_zscore_constant_input_2d_nan_policy_omit[numpy] - Failed: DID NOT WARN. No warnings of type (<class 'RuntimeWarning'>,) were emitted.
 Emitted warnings: [].
FAILED scipy/stats/tests/test_stats.py::TestZscore::test_degenerate_input[numpy] - Failed: DID NOT WARN. No warnings of type (<class 'RuntimeWarning'>,) were emitted.
 Emitted warnings: [].
FAILED scipy/stats/tests/test_stats.py::TestSkew::test_skew_constant_value[numpy] - Failed: DID NOT WARN. No warnings of type (<class 'RuntimeWarning'>,) were emitted.
 Emitted warnings: [].
FAILED scipy/stats/tests/test_stats.py::TestKurtosis::test_kurtosis_constant_value[numpy] - Failed: DID NOT WARN. No warnings of type (<class 'RuntimeWarning'>,) were emitted.
 Emitted warnings: [].
FAILED scipy/stats/tests/test_stats.py::TestTTestInd::test_ttest_ind_zero_division[numpy] - Failed: DID NOT WARN. No warnings of type (<class 'RuntimeWarning'>,) were emitted.
 Emitted warnings: [].
= 7 failed, 59090 passed, 3939 skipped, 16086 deselected, 192 xfailed, 13 xpassed, 211 warnings in 1462.67s (0:24:22) =
```

I know that that job is relatively new and has some intermittent failures, but please don't ignore it completely. Known failures are tracked in gh-22758; it's pretty clear here that the failing tests are due to this PR.

The reason is that `eager_warns` prevented `pytest-run-parallel` from detecting `pytest.warns` (see https://github.com/Quansight-Labs/pytest-run-parallel/pull/19). 

I think we should revert this and figure out an ergonomic way to deal with `eager_warns`. Maybe a special private attribute on `eager_warns` that can get detected in `WarningNodeVisitor` in `pytest-run-parallel`? Cc @andfoy

rgommers:
For completeness: fixed in gh-22771, we need to revert that PR once we have a more ergonomic way to mark `eager_warns` as thread-unsafe in a way that pytest-run-parallel recognizes.

PR #22350, opened 2025-01-17
ENH: Replace Fortran COBYLA with Python version from PRIMA

nbelakovski:
PRIMA now has a pure python implementation of COBYLA!

I'm pleased to announce that after a lot of effort the PRIMA COBYLA algorithm is now implemented in Python. A considerable amount of effort went into validating this implementation against the Fortran and I'm happy to report that against a battery of over 500 problems from cutest, the Fortran and Python implementations get the same result bit for bit.

#### Reference issue
Closes gh-18118

#### Additional information
This commit also
- Updates COBYLA to accept the new style constraints
- Adds several options, including ftarget and more granular iprint
- Updates COBYLA to use actual bounds and linear constraints instead of simply adding them to the nonlinear constraint
- Updates COBYLA to use the new style callback interface
- Updates COBYLA to be threadsafe
- Updates all relevant documentation and tests
- Removes the old implementation

PR #22738, opened 2025-03-25
DOC: PRIMA licence and reference fix

andyfaff:
This PR appends the PRIMA licence to the other bundled licences we use. In addition it adds references to PRIMA. Are you happy with this approach @zaikunzhang and @nbelakovski?

@j-bowhay

### DIFF
--- a/LICENSES_bundled.txt
+++ b/LICENSES_bundled.txt
@@ -305,4 +305,4 @@ License 2-Clause BSD
 Name: pyprima
 Files: scipy/_lib/pyprima
 License: BSD-3-Clause
-  For details, see scipy/_lib/pyprima/LICENCE.txt
+  For details, see scipy/_lib/pyprima/prima/LICENCE.txt
--- a/LICENSES_bundled.txt
+++ b/LICENSES_bundled.txt
@@ -301,3 +301,8 @@ Name: getLebedevSphere
 Files: scipy/integrate/_lebedev.py
 License 2-Clause BSD
   For details, see scipy/integrate/_lebedev.py
+
+Name: pyprima
+Files: scipy/_lib/pyprima
+License: BSD-3-Clause
+  For details, see scipy/_lib/pyprima/LICENCE.txt
--- a/LICENSES_bundled.txt
+++ b/LICENSES_bundled.txt
@@ -302,7 +302,7 @@ Files: scipy/integrate/_lebedev.py
 License 2-Clause BSD
   For details, see scipy/integrate/_lebedev.py
 
-Name: pyprima
+Name: prima
 Files: scipy/_lib/pyprima
 License: BSD-3-Clause
   For details, see scipy/_lib/pyprima/prima/LICENCE.txt
--- a/doc/source/dev/core-dev/vendored-code.rst.inc
+++ b/doc/source/dev/core-dev/vendored-code.rst.inc
@@ -20,6 +20,7 @@ Currently, this includes the following parts of the codebase:
 - QHull_, at ``scipy/spatial/qhull_src``
 - trlib_, at ``scipy/optimize/_trlib``
 - UNU.RAN_, at ``scipy/stats/_unuran``
+- PRIMA_, at ``scipy/_lib/pyprima``
 
 .. _ARPACK: https://github.com/opencollab/arpack-ng
 .. _SuperLU: https://github.com/xiaoyeli/superlu
@@ -27,3 +28,4 @@ Currently, this includes the following parts of the codebase:
 .. _trlib: https://github.com/felixlen/trlib
 .. _UNU.RAN: https://statmath.wu.ac.at/unuran/
 .. _DIRECT: https://github.com/stevengj/nlopt/tree/master/src/algs/direct
+.. _PRIMA: https://github.com/nbelakovski/prima
--- a/scipy/_lib/meson.build
+++ b/scipy/_lib/meson.build
@@ -287,7 +287,6 @@ py3.install_sources(
 py3.install_sources(
   [
     'pyprima/prima/pyprima/src/pyprima/__init__.py',
-    'pyprima/prima/LICENCE.txt',
   ],
   subdir: 'scipy/_lib/pyprima'
 )
--- a/scipy/_lib/meson.build
+++ b/scipy/_lib/meson.build
@@ -287,6 +287,7 @@ py3.install_sources(
 py3.install_sources(
   [
     'pyprima/prima/pyprima/src/pyprima/__init__.py',
+    'pyprima/prima/LICENCE.txt',
   ],
   subdir: 'scipy/_lib/pyprima'
 )
--- a/scipy/optimize/_cobyla_py.py
+++ b/scipy/optimize/_cobyla_py.py
@@ -98,9 +98,10 @@ def fmin_cobyla(func, x0, cons, args=(), consargs=None, rhobeg=1.0,
     how these issues are resolved, as well as how the points v_i are
     updated, refer to the source code or the references below.
 
-    .. versionchanged:: 1.16.0
-    The original Powell implementation was replaced by a pure Python version from
-    the PRIMA package, with bug fixes and improvements being made.
+        .. versionchanged:: 1.16.0
+            The original Powell implementation was replaced by a pure Python version
+            from the PRIMA package, with bug fixes and improvements being made.
+
 
     References
     ----------
@@ -138,6 +139,8 @@ def fmin_cobyla(func, x0, cons, args=(), consargs=None, rhobeg=1.0,
         array([-0.70710685,  0.70710671])
 
     The exact solution is (-sqrt(2)/2, sqrt(2)/2).
+
+
 
     """
     err = "cons must be a sequence of callable functions or a single"\
@@ -213,9 +216,10 @@ def _minimize_cobyla(fun, x0, args=(), constraints=(),
     f_target : float
         Stop if the objective function is less than `f_target`.
 
-    .. versionchanged:: 1.16.0
-    The original Powell implementation was replaced by a pure Python version from
-    the PRIMA package, with bug fixes and improvements being made.
+        .. versionchanged:: 1.16.0
+            The original Powell implementation was replaced by a pure Python version from
+            the PRIMA package, with bug fixes and improvements being made.
+
 
     References
     ----------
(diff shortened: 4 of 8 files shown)


scipy/optimize/_cobyla_py.py
```diff
@@ -112,6 +112,9 @@ def fmin_cobyla(func, x0, cons, args=(), consargs=None, rhobeg=1.0,
     Powell M.J.D. (2007), "A view of algorithms for optimization without
     derivatives", Cambridge University Technical Report DAMTP 2007/NA03
 
+    Zhang Z. (2023), "PRIMA: Reference Implementation for Powell's Methods with
+    Modernization and Amelioration", http://www.libprima.net,
+    DOI: 10.5281/zenodo.8052654
```
j-bowhay, line 117:
```suggestion
    :doi:`10.5281/zenodo.8052654`
```

scipy/optimize/_cobyla_py.py
```diff
@@ -209,6 +212,11 @@ def _minimize_cobyla(fun, x0, args=(), constraints=(),
     f_target : float
         Stop if the objective function is less than `f_target`.
 
+    References
+    ----------
+    Zhang Z. (2023), "PRIMA: Reference Implementation for Powell's Methods with
+    Modernization and Amelioration", http://www.libprima.net,
+    DOI: 10.5281/zenodo.8052654
```
j-bowhay, line 219:
```suggestion
    :doi:`10.5281/zenodo.8052654`
```

j-bowhay:
Maybe we should also put something in https://github.com/scipy/scipy/blob/main/LICENSES_bundled.txt ? I have to admit I didn't know about those 3 other files!

andyfaff:
@j-bowhay that's a better place to make the addition.

zaikunzhang:
Hello Andrew @andyfaff and others,

Thank you very much for taking care of this. I appreciate it. 

I would like to suggest some changes. See 

https://github.com/andyfaff/scipy/pull/80

Note that the PRIMA version of COBYLA is significantly different from Powell's. An example is `DELTA_j` versus `RHO_j`, as explained here: https://github.com/andyfaff/scipy/pull/80/commits/ae0dbee339f7b916c469bcd77d94d28cff4ef622

I also suggest removing two references that did not discuss COBYLA (Powell 1998 and Powell 2007) --- COBYLA itself is difficult enough to understand, let's not confusing users by papers that are essentially irrelevant. 

I suggest (and humbly request) that the new version of COBYLA comes from [the PRIMA package](https://www.libprima.net) by [Zaikun Zhang](https://www.zhangzk.net), so that people know who to blame if something goes wrong due to my changes to the algorithm. 

Forgive me for [repeating](https://github.com/scipy/scipy/pull/22350#issuecomment-2749934407) the following:

> Note that PRIMA contains [bug fixes](https://github.com/libprima/prima?rgh-link-date=2025-03-25T03%3A03%3A28Z#bug-fixes) and [improvements](https://github.com/libprima/prima?rgh-link-date=2025-03-25T03%3A03%3A28Z#improvements) that do not exist in Powell's original implementation of the solvers. Results produced by PRIMA are surely different from those produced by the original solvers by Powell. Therefore, it is simply wrong to pretend that PRIMA is just Powell's solvers.
>
> It took me years to develop PRIMA, which almost cost my career and my life.
>
> As a professor, [all I need is a citation / acknowledgment when my work is useful](https://github.com/libprima/prima?rgh-link-date=2025-03-25T03%3A03%3A28Z#citing-prima), which is not well recognized in academia when it comes to software development. In addition, for obvious reasons, people tend to cite Powell without mentioning me when they use PRIMA instead of Powell's original solvers. (I have no doubt that my contribution is insignificant compared with Powell's, and I acknowledge this whenever I introduce PRIMA to others. See [libprima.net](https://www.libprima.net/) for example). This is crucial for my career.

Thank you very much.

Zaikun

zaikunzhang:
A question, probably can be answered by @nbelakovski  : does the interface of COBYLA in SciPy use `rhoend` or `tol` as the name for the final trust-region radius? The1.16.0 release note mentions that it is `tol`, but I guess it is a typo so I changed in in my PR. 

I did notice `tol` in 

https://github.com/zaikunzhang/scipy_prima/blob/prima/scipy/optimize/_cobyla_py.py#L170

It confuses me a bit. It seems that we receive `rhoend`, change its name to `tol`, and then pass it to `rhoend`. Why? 

Thank you.

scipy/_lib/meson.build
```diff
@@ -287,6 +287,7 @@ py3.install_sources(
 py3.install_sources(
   [
     'pyprima/prima/pyprima/src/pyprima/__init__.py',
+    'pyprima/prima/LICENCE.txt',
```
rgommers, line 290: This file should not be installed. No other license files are. They will instead end up in the `dist-info` directory in the future (once PEP 639 support is out). There's no problem today; there is no actual requirement to install license files, only to ship them.

rgommers (commented):
Thanks for taking care of this @andyfaff.

Can I suggest also adding `scipy/_lib/pyprima` to https://scipy.github.io/devdocs/dev/core-dev/index.html#vendored-code?

scipy/_lib/meson.build
andyfaff: Ok, can remove. How are we shipping? Do we already meet that ba, or do I need to make a change?

andyfaff: Should the licenses_bundled file point to where they are in the source, or installed?

rgommers: > How are we shipping? Do we already meet that bar, or do I need to make a change?

sdists automatically include license files. wheels rely on `LICENSES_bundled.txt` being appended to `LICENSE.txt`. No change needed.

> Should the licenses_bundled file point to where they are in the source, or installed?

The source location.

nickodell:
> A question, probably can be answered by @nbelakovski : does the interface of COBYLA in SciPy use `rhoend` or `tol` as the name for the final trust-region radius? The 1.16.0 release note mentions that it is `tol`, but I guess it is a typo so I changed it in my PR.
>
> [...]
> 
> It confuses me a bit. It seems that we receive `rhoend`, change its name to `tol`, and then pass it to `rhoend`. Why?

My understanding of this is that we actually have two interfaces to COBYLA. We have `scipy.optimize.fmin_cobyla()`, and we have `scipy.optimize.minimize(..., method='COBYLA')`. The `fmin_cobyla()` interface existed first, and must be maintained for backward compatibility. The `minimize()` interface is the recommended way for new code to call COBYLA.

`minimize()` calls this argument `tol`, and `fmin_cobyla()` calls it `rhoend`. This is because `minimize()` is trying to provide a consistent interface for many different solvers. Both eventually call the same code, but they use different names for the parameters, which is why there is conversion back and forth.

Doing a spot check using GitHub code search, the two ways of calling COBYLA seem roughly equally popular.

andyfaff:
I note that there is already a separate pyPRIMA package on PyPI. We should keep this in mind.

nbelakovski:
> I note that there is already a separate pyPRIMA package on PyPI. We should keep this in mind.

It lists Zaikun as the maintainer and points to PDFO so no conflict there. @zaikunzhang, @nickodell's answer above to the question about tol vs rhoend is correct.

LICENSES_bundled.txt
```diff
@@ -300,4 +300,9 @@ License 3-Clause BSD
 Name: getLebedevSphere
 Files: scipy/integrate/_lebedev.py
 License 2-Clause BSD
-  For details, see scipy/integrate/_lebedev.py
\ No newline at end of file
+  For details, see scipy/integrate/_lebedev.py
+
+Name: pyprima
```
nbelakovski, line 305:
```suggestion
Name: prima
```

pyprima is merely a subfolder of PRIMA, I think we should use the proper name here?

zaikunzhang:
> > I note that there is already a separate pyPRIMA package on PyPI. We should keep this in mind.
> 
> It lists Zaikun as the maintainer and points to PDFO so no conflict there. 

It is a placeholder by me, which will be used for PRIMA.

andyfaff:
What we have looks appropriate to me now. Authorship of PRIMA is referenced and the license is pointed to.

doc/source/dev/core-dev/vendored-code.rst.inc
```diff
@@ -20,10 +20,12 @@ Currently, this includes the following parts of the codebase:
 - QHull_, at ``scipy/spatial/qhull_src``
 - trlib_, at ``scipy/optimize/_trlib``
 - UNU.RAN_, at ``scipy/stats/_unuran``
+- PRIMA_, at ``scipy/_lib/pyprima``
```
lucascolley, line 23: On second thought, this is unnecessary as we are actually vendoring this as a git submodule. Not a big deal though

rgommers: Indeed. I didn't realize we had changed plans; I thought we wouldn't add a submodule (no bandwidth to follow along unfortunately).

lucascolley: my understanding was that we converged on:
- not adding a submodule to Nikolai's repo, instead 'mirroring' under the SciPy org (for security reasons)
- not adding a 'full' submodule, but rather using a script to trim off the parts we don't need to use (like the Fortran PRIMA implementation) while vendoring a specific commit hash

That script is https://github.com/scipy/pyprima/blob/main/vendor_pyprima.sh. I suppose we could just maintain the script and the vendored source in the main `scipy` repo, but I think the separation is nice.

rgommers: > I suppose we could just maintain the script and the vendored source in the main scipy repo, but I think the separation is nice.

That's what I had thought we settled on indeed. The separation now just gives us the overhead of a git submodule without the ability to pull in or work on upstream commits without much of an upside, right?

Anyway, I'm not suggesting to change it. We have enough submodules, and this probably won't be updated too often.

lucascolley: Agreed. I think the only upside of the obvious separation is making it clear that users shouldn't (/can't) modify the source in this repo.

But that could equally be achieved (at some point in the future) by a refactor of `_lib` and which cleanly separates internal from external code (scikit-learn separates `sklearn.utils` and `sklearn.externals` for example).

PR #22742, opened 2025-03-26
CI/MAINT: make special.errstate thread-safe and run pytest-run-parallel on special/stats

rgommers:
This is a follow-up to gh-22419 for `special.errstate`, and supersedes gh-21956.

This makes tests pass under `pytest-run-parallel`; the logic was already broken for use of the default build with Python-level threads (e.g., through the `threading` module) though. This is per-module global state that wasn't robust against
altering it from separate threads.

The opportunity to remove the `WITH_THREAD` usage was pointed out in gh-21956. It's always defined for Python >=3.7, threadless builds no longer exist.


tylerjereddy (approved):
I checked locally on x86_64 Linux and saw fail before (on `main`)/pass after (on this branch) for the following incantation on 3.13t with GIL disabled: `python dev.py test -j 8 -t scipy/stats -- --parallel-threads=10 --iterations=1`

The CI seems to agree.

The diff looks solid to me. I initially wondered about the `meson` changes--presumably that's to provide access to the config header though.

tylerjereddy:
thanks Ralf

rgommers:
> initially wondered about the `meson` changes--presumably that's to provide access to the config header though.

yep, that's correct

Issue #20682, opened 2024-05-09
RfC:Switch to `cython_<lapack,blas>` based wrappers and deprecate custom `scipy.linalg.<lapack,blas>`


ilayn:
Historically, we have gone through many stages of BLAS/LAPACK wrappers. And consequently we have a lot of home-baked wrappers to the formal BLAS/LAPACK routines. 

This actually allowed the users to conveniently call native routines through `f2py` glue. While this is very valuable, our way of carrying the historical inconsistencies have been quite high. Just to name a few

- Character switches sometimes kept ala `uplo='L'` or converted into `0`, `1` ala `lower=1` (since bool is not really available). 
- The pivot arrays sometimes converted at the wrapper with `C` injection, sometimes kept in Fortran order.
- `lwork` queries sometimes included as separate public function `<routine>_lwork` but sometimes left to user by calling the same routine twice with exposed `lwork` keyword set to `-1`.
- This would also help us handling the OpenBLAS linkage or other libs a bit more ergonomic.

I think we can switch to cython based linkage to have more automated wrapper generation and less strain on our testing framework. And also it would make things more straightforward to compile. 

On the downside, we let f2py to handle the normalization of arguments and this has to be robust in the cythonized wrappers. Moreover, we will have to make sure that the F-contiguous and `<s,d,c,z>` flavorizing happens properly. 

We touched upon this many times over the years but let's consolidate the discussions here (last time we mentioned this was I think https://github.com/scipy/scipy/pull/13663#issuecomment-1179792099) 

One particular difficulty is that if we want to solve the historical issues, unfortunately, we have to get rid of all wrappers and offer new ones where the naming would be different. I'm not necessarily a big fan of this but doing this without breaking anybody's current code this seems like the only option. And probably needs more than 2 versions for deprecation cycles. 

Hence it would be the case that `scipy.linalg.lapack` and `scipy.linalg.blas` gets deprecated goes away and new ones, whatever the name might be will be the new auto-generated submodule names.

ev-br:
cross-ref https://github.com/scipy/scipy/issues/4516

ilayn:
Just occured to me that we can at least do this for our own internal use and leave the deprecation for later.

rgommers:
This proposal isn't very clear to me I'm afraid. We should consider public API we expose separately from what we do internally. For public API:
[...]

PR #19816, opened 2024-01-06
BLD: Add Accelerate support for macOS 13.3+

thalassemia:
I added a script ([scipy/_build_utils/_generate_blas_wrapper.py](https://github.com/scipy/scipy/compare/main...thalassemia:accelerate?expand=1#diff-b9fb73b1113defdbecdd9276cd1a893eac534c6b47a53e3512195ee85a2f6203)) to generate C wrappers for Accelerate BLAS/LAPACK. These wrappers are compiled into a static library that other modules can link against with no code modifications (e.g. a call to `dgemm` in a Fortran module will be routed through a wrapper and turned into a call to `dgemm$NEWLAPACK`). I also added Accelerate support to the script ([scipy/linalg/_generate_pyx.py](https://github.com/scipy/scipy/compare/main...thalassemia:accelerate?expand=1#diff-23487f1e98e4e3ec6489ff8fa3919ba87494ddfb3dea4af817b9c61de6c41c52)) that generates the Cython for `_fblas`/`_flapack` using similar C wrappers. In the process of making these two changes, I included special logic to handle the G77 ABI wrappers, eliminating the need for additional Fortran wrappers.

Other changes:
- Functions in `_fblas`/`_flapack` that have a return value were written as such (previously, f2py generated Fortran wrappers for these)
  
  - UPDATE: Reverting this change because the intermediate Fortran wrappers are needed to guarantee compatibility of complex return types, without which segfaults can occur on x86 machines (#19855)
  - UPDATE 2: Brought this change back to avoid [erroneous F2PY boilerplate](https://github.com/scipy/scipy/pull/19816#issuecomment-1970338432) that referenced incorrect BLAS/LAPACK symbols. Fortran subroutine wrappers were translated into C and appended to G77 ABI wrapper file (#20232)

- Resolved missing prototype build warnings in [scipy/integrate/lsoda.pyf](https://github.com/scipy/scipy/compare/main...thalassemia:accelerate?expand=1#diff-bcfb240e4cf0b9546fdc187135f5e1711944e85a992c31e5b8057d62b3d10819) and [scipy/integrate/vode.pyf](https://github.com/scipy/scipy/compare/main...thalassemia:accelerate?expand=1#diff-3a625b9354fc6062bb9c6c0d28d527e376da21034e7cd296e8754afb63cf9051)

  - UPDATE: Separated into #19821
 
- Added test for `cython_blas.cladiv` and `cython_blas.dladiv`

  - UPDATE: Separated into #19820
 
[...]

PR #12085, opened 2020-05-10
ENH: special: support ILP64 Lapack

pv:
#### Reference issue
Part of gh-11193

#### What does this implement/fix?
Add support for ilp64 Lapack in scipy.special.
It's not used there in many places, only in the C code of ellip_harm*

PR #12030, opened 2020-05-05
 ENH: optimize/_trlib: support ILP64 blas/lapack 

pv:
#### Reference issue
Part of gh-11193

#### What does this implement/fix?
Support 64-bit BLAS/LAPACK in trlib.

Add wrapper routines for trlib BLAS/LAPACK calls, with correct integer types.

Because trlib_int_t is long, trlib can take advantage of ilp64 BLAS on
64-bit systems.

This change also fixes an issue in the previous code: trlib_int_t is not
necessarily compatible with BLAS integers even without ilp64, so the
wrappers are necessary in any case.

The replacement `cblas.h` is the same as used in Numpy, and needed
here mostly for the symbol mangling macros. In principle it also makes
the BLAS ABI wrappers work properly with symbol name mangling,
but this is a bit theoretical corner case.

PR #22743, opened 2025-03-26
ENH: ILP64 BLAS/LAPACK support

### LINKED ISSUE #20002: ENH: linking LAPACK from C/C++

`cython_{blas,lapack}` does a great job to make using LAPACK from Cython easy: just

```
from scipy.linalg.cython_lapack cimport dgeev
```

and you're up and running.

Using it from C or C++ should be as easy, but it currently isn't. 

The ugly part is dealing with Fortran name mangling, which `cython_lapack`  hides. It actually does the job internally:

```
$ ll build/scipy/linalg/_lapack_subroutine*
-rw-rw-r-- 1 br br 246836 янв 25 21:59 build/scipy/linalg/_lapack_subroutines.h
-rw-rw-r-- 1 br br  34384 янв 25 21:59 build/scipy/linalg/_lapack_subroutine_wrappers.f
```

Here `_lapack_subroutines.h` exports all the correctly named pointers:

```
$ less build/scipy/linalg/_lapack_subroutines.h
/* This file was generated by _generate_pyx.py. */
/* Do not edit this file directly. */

#ifndef SCIPY_LINALG_LAPACK_FORTRAN_WRAPPERS_H
#define SCIPY_LINALG_LAPACK_FORTRAN_WRAPPERS_H
#include "fortran_defs.h"
#include "numpy/arrayobject.h"

<snip>

#ifdef __cplusplus
extern "C" {
#endif

void F_FUNC(chla_transtypewrp, CHLA_TRANSTYPEWRP)(char *ret, int *trans);
void F_FUNC(cladivwrp, CLADIVWRP)(npy_complex64 *ret, npy_complex64 *x, npy_complex64 *y);
void F_FUNC(clangbwrp, CLANGBWRP)(float *ret, char *norm, int *n, int *kl, int *ku, npy_complex64 *ab, int *ldab, float *work);

...

```

so that `#including`  `_lapack_signatures` looks like the solution for user code, and it even _almost_ works. 
Using e.g. 

```patch
$ git diff
diff --git a/scipy/interpolate/meson.build b/scipy/interpolate/meson.build
index b9b3ed852b..eaa5e68c33 100644
--- a/scipy/interpolate/meson.build
+++ b/scipy/interpolate/meson.build
@@ -124,7 +124,7 @@ _bspl = py3.extension_module('_bspl',
   cython_gen.process('_bspl.pyx'),
   c_args: cython_c_args,
   include_directories: 'src/',
-  dependencies: np_dep,
+  dependencies: [lapack, np_dep],
   link_args: version_link_args,
   install: true,
   subdir: 'scipy/interpolate'
diff --git a/scipy/interpolate/src/__fitpack.h b/scipy/interpolate/src/__fitpack.h
index 97ac44a950..e70b081140 100644
--- a/scipy/interpolate/src/__fitpack.h
+++ b/scipy/interpolate/src/__fitpack.h
@@ -1,3 +1,22 @@
+
+#include "../linalg/_lapack_subroutines.h"
+
+
+/*
+
+#include "../linalg/fortran_defs.h"
```

fails to build with

```
$ python dev.py build
<snip>
Found ninja-1.11.1 at /home/br/mambaforge/envs/scipy-dev/bin/ninja
Cleaning... 0 files.
[32/43] Compiling C object scipy/interpolate/_fitpack.cpython-310-x86_64-linux-gnu.so.p/src__fitpackmodule.c.o
FAILED: scipy/interpolate/_fitpack.cpython-310-x86_64-linux-gnu.so.p/src__fitpackmodule.c.o 
/home/br/mambaforge/envs/scipy-dev/bin/x86_64-conda-linux-gnu-cc -Iscipy/interpolate/_fitpack.cpython-310-x86_64-linux-gnu.so.p -Iscipy/interpolate -I../scipy/interpolate -I../scipy/interpolate/src -I../../../../mambaforge/envs/scipy-dev/lib/python3.10/site-packages/numpy/core/include -I/home/br/mambaforge/envs/scipy-dev/include/python3.10 -fvisibility=hidden -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -std=c99 -O2 -g -Wno-unused-but-set-variable -Wno-unused-function -Wno-conversion -Wno-misleading-indentation -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem /home/br/mambaforge/envs/scipy-dev/include -DNDEBUG -D_FORTIFY_SOURCE=2 -O2 -isystem /home/br/mambaforge/envs/scipy-dev/include -fPIC -DNPY_NO_DEPRECATED_API=NPY_1_9_API_VERSION -MD -MQ scipy/interpolate/_fitpack.cpython-310-x86_64-linux-gnu.so.p/src__fitpackmodule.c.o -MF scipy/interpolate/_fitpack.cpython-310-x86_64-linux-gnu.so.p/src__fitpackmodule.c.o.d -o scipy/interpolate/_fitpack.cpython-310-x86_64-linux-gnu.so.p/src__fitpackmodule.c.o -c ../scipy/interpolate/src/_fitpackmodule.c
In file included from ../scipy/interpolate/src/__fitpack.h:2,
                 from ../scipy/interpolate/src/_fitpackmodule.c:7:
scipy/interpolate/../linalg/_lapack_subroutines.h:6:10: fatal error: fortran_defs.h: No such file or directory
    6 | #include "fortran_defs.h"
      |          ^~~~~~~~~~~~~~~~
compilation terminated.
[33/43] Compiling C object scipy/interpolate/_bspl.cpython-310-x86_64-linux-gnu.so.p/meson-generated__bspl.c.o
FAILED: scipy/interpolate/_bspl.cpython-310-x86_64-linux-gnu.so.p/meson-generated__bspl.c.o 
/home/br/mambaforge/envs/scipy-dev/bin/x86_64-conda-linux-gnu-cc -Iscipy/interpolate/_bspl.cpython-310-x86_64-linux-gnu.so.p -Iscipy/interpolate -I../scipy/interpolate -I../scipy/interpolate/src -I../../../../mambaforge/envs/scipy-dev/lib/python3.10/site-packages/numpy/core/include -I/home/br/mambaforge/envs/scipy-dev/include -I/home/br/mambaforge/envs/scipy-dev/include/python3.10 -fvisibility=hidden -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -std=c99 -O2 -g -Wno-unused-but-set-variable -Wno-unused-function -Wno-conversion -Wno-misleading-indentation -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem /home/br/mambaforge/envs/scipy-dev/include -DNDEBUG -D_FORTIFY_SOURCE=2 -O2 -isystem /home/br/mambaforge/envs/scipy-dev/include -fPIC -DNPY_NO_DEPRECATED_API=NPY_1_9_API_VERSION -DCYTHON_CCOMPLEX=0 -MD -MQ scipy/interpolate/_bspl.cpython-310-x86_64-linux-gnu.so.p/meson-generated__bspl.c.o -MF scipy/interpolate/_bspl.cpython-310-x86_64-linux-gnu.so.p/meson-generated__bspl.c.o.d -o scipy/interpolate/_bspl.cpython-310-x86_64-linux-gnu.so.p/meson-generated__bspl.c.o -c scipy/interpolate/_bspl.cpython-310-x86_64-linux-gnu.so.p/_bspl.c
In file included from ../scipy/interpolate/src/__fitpack.h:2,
                 from scipy/interpolate/_bspl.cpython-310-x86_64-linux-gnu.so.p/_bspl.c:1198:
scipy/interpolate/../linalg/_lapack_subroutines.h:6:10: fatal error: fortran_defs.h: No such file or directory
    6 | #include "fortran_defs.h"
      |          ^~~~~~~~~~~~~~~~
```

Have a canonical, single header file to include. 

Of course replicating the name mangling manually at include sites works:

```diff
diff --git a/scipy/interpolate/src/__fitpack.h b/scipy/interpolate/src/__fitpack.h
index 97ac44a950..e70b081140 100644
--- a/scipy/interpolate/src/__fitpack.h
+++ b/scipy/interpolate/src/__fitpack.h
@@ -1,3 +1,22 @@
+#include "../linalg/fortran_defs.h"
+
+#define DLARTG F_FUNC(dlartg, DLARTG)
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+void DLARTG(double *f, double *g, double *cs, double *sn, double *r);
+#ifdef __cplusplus
+}
+#endif
```

This is what we had here and there before `cython_lapack.pxd`, and I'd much prefer to not have these beauties sprinkled across the codebase :-).

I don't know if forward slashes in the include path are a problem on Windows. If they are, no biggie, can copy-paste the `F_FUNC` macro. But the point is it's ugly, it's brittle, it's error-prone, and it's reinventing a small wheel at each include site.  

I'm mostly interested in internal usage in scipy. Whether we want to make this a part of the public SciPy contract (as cython_lapack is) is a separate question.

ev-br:
#### Reference issue

towards https://github.com/scipy/scipy/issues/21889

fixes https://github.com/scipy/scipy/issues/23219

Add changes needed to build SciPy with 64-bit (ILP64) BLAS and LAPACK.

The main bulk of work for LAPACK detection is done by Ralf, see the commit history. (I squashed several commits to simplify rebasing, please let me know if some useful history got lost @rgommers).

On top of Ralf's work:
- hammer the MKL ILP64 detection into shape
- add source changes needed for the with-ILP64 MKL test run run to completion.
- EDIT: `scipy-openblas64` now works, too.

Am marking this PR as an enhancement even though we had all of this before moving off distutils to meson, so this PR just restores the status quo ante :-).

#### Additional information

The approach here is the option 2 of gh-21889 : we expect a LAPACK provider to provide two sets of symbols, 32-bit and 64-bit. The former must be always available, the latter are requested via a meson option.

The approach does not work for a general OpenBLAS build at the moment, so this PR only enables it for from-source builds with MKL and `scipy-openblas64`. See the `linux_blas_ilp64.yml` workflow for details.

How to test this: sadly, we cannot easily test linear algebra with sizes $> 2^{31}$ on CI. At the moment the only CI indicator is that the test run does not segfault. Locally, there's a test in linalg, `test_basic/test_large_vector`, which fails without ILP64 and passes with it.

Potential follow-ups:

- fix up "slow" tests. ATM `dev.py test -m "full"` fails with ILP64 for trivial reasons
- currently several submodules (optimize, special, cluster, odr) are using 32-bit integer size linear algebra; it is not very clear which of these really need 64-bit integers (SLSQP with $2^{31}$ variables, cough.
- generalize to other lapacks: Accelerate et al
- resolve https://github.com/scipy/scipy/issues/20002, provide a "canonical" header for LAPACK symbols to use from C/C++

Bigger issues:

- Decide on `cython_blas`  and  `cython_lapack`. Do we provide additional `cython_blas64` and `cython_lapack64` cython extensions?
- Do something in/with openblas which is not `scipy-openblas{32,64}`, so that we can have ILP64 by default, maybe even in the official wheels. That needs coordination with OpenBLAS and NumPy, it seems.


rgommers:
Nice to see progress on this, thanks @ev-br!

Once CI is happy, I think it makes sense to split off the changes to `.c|h|pyf` files and merge those, since that can immediately be done. 

> Am marking this PR as an enhancement even though we had all of this before moving off distutils to meson, so this PR just restores the status quo ante :-).

Not really, I am fairly sure that things only ever worked with a pretty janky OpenBLAS setup, not with MKL or Accelerate (different symbol name conventions).

ev-br:
Okay, some progress: CI is all green for both existing jobs and the added ILP64 job;
Locally, the smoking gun test is passing, too:

   $ python dev.py build -C-Dblas=mkl -C-Duse-ilp64=true
   $ python dev.py test -s linalg -- -k'large_vector'

This test is skipped if ILP64 BLAS is not available, and fails if it is available but
LP64 routines get used instead.

To get there,

- removed the blas-order selection;
- reverted the ILP64 usage in most submodules apart from linalg
- simplified the ILP64 linking and made it less reliant on MKL details


> Once CI is happy, I think it makes sense to split off the changes to `.c|h|pyf` files and merge those, since that can immediately be done.

CI is now, and some the majority of `c|h|pyf` changes are in; 
but let's maybe do it the other way around and start from the build system changes.
This PR then is a minimal reasonable product (MRP): it has the needed tweaks to meson.build files + minimal source changes to have
scipy.linalg working.
Once we're happy with this (am ready to trim more fat from this one if you want me to),
I'll work on making ILP64 work in some of the newer C submodules. WDYT?

ev-br:
> Is it going to be compile-time or runtime switch you have in mind?

Build time only, sorry. Need to have MKL when building scipy. Can you build with MKL locally?

 > Not sure how to help but great that this is happening. The new C modules are pretty much in the "find/replace int -> ILP_INT" state. But the calls to BLAS/LAPACK, I don't know if possible to do if the symbol names are going to be varying.

The basic idea is in interpolate in this PR. Let's iron it out on one of the additions, shall we?
It'd be best to have an example which clearly fails if LAPACK is 32-bit and passes if it is  64-bit.

rgommers:
> This PR then is a minimal reasonable product (MRP): it has the needed tweaks to meson.build files + minimal source changes to have
> scipy.linalg working.

Ah okay, I am also fine with trying to merge this whole PR as long as we're a bit careful with how it's advertised. The CI job now only works because of:
```
pip install git+https://github.com/numpy/meson.git@main-numpymeson
```
which means the new build options need a disclaimer to not use/rely on them for distro packaging etc., they are subject to change.

ev-br:
> But the calls to BLAS/LAPACK, I don't know if possible to do if the symbol names are going to be varying.

Here's a branch: https://github.com/ev-br/scipy/compare/ilp64...ev-br:scipy:nnls_ilp64?expand=1
Currently it passes all tests locally with the "usual" lapack; `test_nnls.py` passes with ILP64, and SLSQP segfaults with it :-). I rather strongly suspect the culprit is in rather obscure `SLSQP_vars` handling, which involves `PyLong_FromLong` etc and those are quite unforgiving to C int type mismatches. Anyhow, these things is one reason why I think it makes sense to defer porting these C extensions to a separate PR.


> a bit careful with how it's advertised. The CI job now only works because of:
   >> pip install git+https://github.com/numpy/meson.git@main-numpymeson

Sure, it's ways off from being ready for distro packaging. For now it's just "there exists at least one not completely insane setup where it works". That said, the need for a numpy meson fork is orthogonal to scipy-internal code/build changes, I'd think. Once we have this single CI job, we can incrementally iron out smaller issues, generalize to other lapacks etc. And once numpy's meson fork lands upstream, we just stop using the fork and all is well.

ilayn:
That's a very common Python dict <-> C struct mapping. So if there is a problem it should error out but I think it might be that LongLong is not necessarily 64 bit. Separate PR also seems good to me so that we don't conflate with side effects.

Also if we are doing this, we should map to exact integer types from `stdint.h` https://en.cppreference.com/w/c/types/integer so that we never fall into these architectural issues.

scipy/_build_utils/src/wrap_g77_abi_ilp64.c
```diff
@@ -0,0 +1,114 @@
+/*
+ * This file is an analog of wrap_g77_abi.c, only for the ILP64 BLAS build.
+ * See `wrap_g77_abi.c` for the purpose and scope.
+ *
+ * The only difference between this and wrap_g77_abi.c is that here we
+ * export the wrappers mangled with BLAS_FUNC (i.e. with BLAS_PREFIX) instead
+ * of mangling with F_FUNC from fortran_defs.h.
+ * This mirrors the usage in scipy/linalg/fblas_64.pyf.src
```
rgommers, line 8: Did you consider avoiding the duplication by, for example, defining a new `WRAP_BLAS_FUNC` which is an alias to either `BLAS_FUNC` or `F_FUNC` depending on whether `HAVE_BLAS_ILP64` is defined?

scipy/interpolate/meson.build
```diff
@@ -123,25 +128,33 @@ py3.extension_module('_rgi_cython',
 
 __fitpack_lib = static_library('__fitpack',
     ['src/__fitpack.h', 'src/__fitpack.cc'],
-    dependencies:[lapack, np_dep, py3_dep],
+    dependencies:[lapack_ilp64, np_dep, py3_dep],
+    c_args: c_flags_ilp64
```
rgommers, line 132: This looks wrong. It's possible that the code is actually correct, but only if `lapack_ilp64` can also contain an LP64 lapack. In which case it's still confusing; best to have a cleaner solution in `scipy/meson.build` then - both these variables need renaming at a minimum.

scipy/interpolate/meson.build
```diff
@@ -123,25 +128,33 @@ py3.extension_module('_rgi_cython',
...
 )
 
-# TODO: Add flags for 64 bit ints
-py3.extension_module('_fitpack',
+# Build _fitpack and dfitpack extensions: both are FITPACK wrappers.
+# XXX: some functions from dfitpack use the F_INT macro defined in dfitpack.pyf, while
+#      others rely on the build flags only. Consider cleaning this up at some point.
+
+
+_fitpack = py3.extension_module('_fitpack',
```
rgommers, line 155: Defining `_fitpack` here doesn't seem to do anything. Please consider reverting to what it was before:
```meson
py3.extension_module('_fitpack',
```

scipy/interpolate/meson.build
```diff
@@ -151,12 +164,20 @@ py3.extension_module('_fitpack',
   subdir: 'scipy/interpolate'
 )
 
-# TODO: Add flags for 64 bit ints
+if use_ilp64
+   # generator only accepts strings, not files
+   f2c_map_file = f2py_ilp64_opts[1]
+   extra_arg = f2py_ilp64_opts[0] + '=' + fs.parent(f2c_map_file) / fs.name(f2c_map_file)
+  _dfitpackmodule_obj = f2py_gen.process('src/dfitpack.pyf', extra_args: extra_arg)
```
rgommers, line 171: This deserves an explanation in a code comment for why only `dfitpack` needs this `f2c_map_file`.

scipy/interpolate/meson.build
rgommers: `c_flags_ilp64` can probably also be wrapped into the `lapack` object in `scipy/meson.build` so that there's no way to get it wrong anywhere. It's not good practice to carry around required flags that go with a dependency in a separate variable. Something like (untested):
```meson
lapack = declare_dependency(
    dependencies: lapack,
    compile_args: c_flags_ilp64,
)
```

scipy/linalg/flapack_64.pyf.src
```diff
@@ -16,9 +16,23 @@
 
 python module _flapack_64
     usercode '''
-#if defined(BLAS_SYMBOL_PREFIX) || defined(BLAS_SYMBOL_SUFFIX)
-#include "blas64-prefix-defines.h"
```
rgommers, line 20: xref https://github.com/scipy/scipy/pull/19816#issuecomment-1879792156 for what this used to do

(no change to the code here needed)

rgommers: That said, the lines below are now repeated twice, for `fblas_64` and `flapack_64` - would be good to put that into a new header.

scipy/interpolate/meson.build
rgommers: Same comment applies to `_fflag_lp64`, `_fflag_ilp64`

scipy/linalg/meson.build
```diff
@@ -66,9 +66,12 @@ fblas_module = custom_target('fblas_module',
 # LAPACK - we have historically put these in `_fblas`.
 py3.extension_module('_fblas',
   fblas_module,
+  fortran_args: _fflag_lp64,
   link_args: version_link_args,
   dependencies: [lapack_dep, blas_dep, fortranobject_dep],
+  link_with: [g77_abi_wrappers],
   install: true,
+  link_language: 'fortran',
```
rgommers, line 74: Question: why did this change? It seems wrong, this should be a pure C module.

rgommers (commented):
Thanks Evgeni. Overall looks pretty reasonable. Here is a first set of review comments.

scipy/interpolate/meson.build
ev-br: > This looks wrong. It's possible that the code is actually correct, but only if lapack_ilp64 can also contain an LP64 lapack.

Almost. Does not contain it, but may be aliased to it
- `lapack_dep` is always LP64
- `lapack_ilp64` is ILP64 if available, equals `lapack_dep` otherwise (https://github.com/scipy/scipy/pull/22743/files#diff-b39b7bda5a39c101be8d9e0b3b6d4da335b6daf7385c2a43463125b0a1f33778R445)

Then a C extension links to `lapack_dep` if it's always LP64, or to `lapack_ilp64` if it is prepared to switch between LP64 and ILP64, if available.

Am open to a better naming scheme if this one's confusing.
----------------

An alternative is, of course, to have `lapack_ilp64` to be always ILP64 or not defined; then extensions will need to if-else in their `py3.extension_module` stanzas.

scipy/linalg/meson.build
ev-br: No reason apparently. Is likely a leftover from past iterations. Removed now, thanks!

scipy/interpolate/meson.build
ev-br: > c_flags_ilp64 can probably also be wrapped into the lapack object in scipy/meson.build so that there's no way to get it wrong anywhere

Done for this specific instance: things which depend on `lapack_ilp64` indeed get the right flags already (via https://github.com/scipy/scipy/pull/22743/files#diff-b39b7bda5a39c101be8d9e0b3b6d4da335b6daf7385c2a43463125b0a1f33778R442)

Other instances of `c_flags_ilp64` and `_fflag_ilp64`  I'm not sure how to remove: they are used to build fitpack_lib, which needs both C and Fortran build flags, and does not depend on `blas_ilp64`. Note that this is specific to a Fortran library which does not use BLAS/LAPACK.
Not sure what is the best practice is then. Make it artificially depend on `blas_ilp64` or make a special dependency just for the flags (how is this better than passing the flags around, as is currently done), or just leave as is. Happy to follow the best practice, just not sure what it is.

ev-br:
Rebased on main to resolve conflicts this PR picked up, and to replace `dev.py` with `spin` in the docs.

meson.options
```diff
@@ -2,6 +2,12 @@ option('blas', type: 'string', value: 'openblas',
         description: 'option for BLAS library switching')
 option('lapack', type: 'string', value: 'openblas',
         description: 'option for LAPACK library switching')
+option('use-ilp64', type: 'boolean', value: false,
+       description: 'Use ILP64 (64-bit integer) BLAS and LAPACK interfaces')
+option('blas-symbol-suffix', type: 'string', value: 'auto',
+        description: 'BLAS and LAPACK symbol suffix to use, if any')
+option('mkl-threading', type: 'string', value: 'auto',
+        description: 'MKL threading method, one of: `seq`, `iomp`, `gomp`, `tbb`')
```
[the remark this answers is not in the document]
ev-br: Done in a code comment just above. Or do want to have this wording in the `description` field?

doc/source/building/blas_lapack.rst
```diff
@@ -96,6 +96,44 @@ user wants to override this autodetection mechanism for building against plain
...
+Building with ILP64 support requires several NumPy additions to ``meson``, which have
+not been merged to upstream yet::
+
+    $ pip install git+https://github.com/numpy/meson.git@main-numpymeson
+
+For a development build with MKL, install the library and its development headers, and
+give use the ``ilp64=true`` option::
+
+    $ pip install mkl mkl-devel
+    $ spin build -C-Dblas=mkl -C-Duse-ilp64=true
```
ev-br:
```suggestion
    $ spin build -S-Dblas=mkl -S-Duse-ilp64=true
```

doc/source/building/blas_lapack.rst
```diff
@@ -96,6 +96,44 @@ user wants to override this autodetection mechanism for building against plain
...
+
+For a development build with ``scipy-openblas64``, make sure you have installed both
+``scipy-openblas32`` and ``scipy-openblas64``, and generate the pkg-config file
+for the ILP64 variant::
+
+    >>> python -c'import scipy_openblas64 as so64; print(so64.get_pkg_config())' > scipy-openblas64.pc
+    >>> export PKG_CONFIG_PATH=`pwd`
+    >>> spin build --with-scipy-openblas -C-Duse-ilp64=true
```
ev-br:
```suggestion
    >>> spin build --with-scipy-openblas -S-Duse-ilp64=true
```

rgommers:
Doing another review pass now. @ev-br may be good to rebase once more.

.github/workflows/linux_blas_ilp64.yml
```diff
@@ -0,0 +1,129 @@
+name: BLAS tests (Linux)
+
+# This file is meant for testing different BLAS/LAPACK flavors and build
+# options on Linux. All other yml files for Linux will only test without BLAS
```
rgommers: That's copied from `numpy` I think - it's not possible to build `scipy` without any BLAS. Can you update/remove?

.github/workflows/linux_blas_ilp64.yml
```diff
@@ -0,0 +1,129 @@
...
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
+jobs:
+
+  mkl-lp64:
+    runs-on: ubuntu-latest
+    name: "MKL LP64"
+    steps:
+    - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
```
rgommers: can you bump to v4.2.2 like in the other yml files? same applies further down

.github/workflows/linux_blas_ilp64.yml
```diff
@@ -0,0 +1,129 @@
...
+      with:
+        submodules: recursive
+        fetch-depth: 0
+    - uses: actions/setup-python@0a5c61591373683505ea898e09a3ea4f39ef2b9c # v5.0.0
```
rgommers: can you bump to v5.5.0 like in the other yml files? same applies further down

.github/workflows/linux_blas_ilp64.yml
```diff
@@ -0,0 +1,129 @@
...
+        pip install git+https://github.com/numpy/meson.git@main-numpymeson
+        pip install mkl mkl-devel
+
+    - name: Build with defaults (LP64)
+      run: |
+        pkg-config --libs mkl-dynamic-lp64-seq  # check link flags
+        spin build -S-Dblas=mkl
+
+    - name: Test
+      run: spin test -j 2
```
rgommers: nit: can you spell as `-j2` like in other files? and same further down

.github/workflows/linux_blas_ilp64.yml
```diff
@@ -0,0 +1,129 @@
...
+
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y gfortran
+        pip install cython numpy pybind11 pythran pytest hypothesis pytest-xdist pooch
+        pip install -r requirements/dev.txt
+        pip install git+https://github.com/numpy/meson.git@main-numpymeson
+        pip install scipy-openblas32 scipy-openblas64
+        # spin does this for scipy-openblas32
```
rgommers: Can you add a separate job step for this?
```
name: Write out scipy-openblas64.pc
```

.github/workflows/linux_blas_ilp64.yml
```diff
@@ -0,0 +1,129 @@
...
+        python -c'import scipy_openblas64 as so64; print(so64.get_pkg_config())' > scipy-openblas64.pc
+        export PKG_CONFIG_PATH=`pwd`
```
rgommers: This `export` should not be doing anything, since it doesn't get persisted between job steps. You normally need `echo xxx >> "$GITHUB_ENV"`. If things work for this job, you should be able to delete this line.

doc/source/building/blas_lapack.rst
```diff
@@ -96,6 +96,44 @@ user wants to override this autodetection mechanism for building against plain
     $ python -m build -C-Duse-g77-abi=true -Csetup-args=-Dblas=blas -Csetup-args=-Dlapack=lapack
 
 
+64-bit integer (ILP64) BLAS/LAPACK
+----------------------------------
+
+Support for ILP64 BLAS and LAPACK is still experimental; at the time of writing
+(Apr 2025) it is only available for two BLAS/LAPACK configurations: MKL and
```
rgommers: can be updated to (Jul 2025)

rgommers (commented):
If MKL works, I'd expect Accelerate to work as well pretty much unchanged. Did you test? If not, should I?

rgommers:
(argh partial review, not done yet)

doc/source/building/blas_lapack.rst
```diff
@@ -96,6 +96,44 @@ user wants to override this autodetection mechanism for building against plain
...
+``scipy-openblas``.
+
+SciPy always requires LP64 (32-bit integer size) BLAS/LAPACK. You can build SciPy
+with *additional* ILP64 support. This will result in SciPy requiring both BLAS and
+LAPACK variants, where some extensions link to the ILP64 variant, while other
+extensions link to the LP64 variant. From python, choosing the variant is done
```
rgommers: nit Python with capital P

rgommers: More importantly, this should be in a separate paragraph and with a bit more context, since it now looks like `get_*_funcs` is somehow important to regular usage while in reality it's super niche and doesn't affect how any of the normal public APIs behave.

doc/source/building/blas_lapack.rst
```diff
@@ -96,6 +96,44 @@ user wants to override this autodetection mechanism for building against plain
...
+    >>> daxpy.int_dtype
+    dtype('int64')
+
+Building with ILP64 support requires several NumPy additions to ``meson``, which have
+not been merged to upstream yet::
+
+    $ pip install git+https://github.com/numpy/meson.git@main-numpymeson
+
+For a development build with MKL, install the library and its development headers, and
+give use the ``ilp64=true`` option::
```
rgommers:
```suggestion
use the ``ilp64=true`` build option::
```

scipy/_build_utils/src/_blas64_defines.h
```diff
@@ -0,0 +1,32 @@
...
+#endif
+
+#ifdef F_FUNC
+#undef F_FUNC
+#endif
+
+#include "npy_cblas.h"
+#define F_FUNC(f, F) BLAS_FUNC(f)
+
+#ifdef FIX_MKL_2025_ILP64_MISSING_SYMBOL
```
rgommers, line 51: Can you open an upstream bug report and link to it, if there isn't one already? On the Intel forum I think. It will probably be resolved only slowly, but it may help.

scipy/_build_utils/src/wrap_g77_abi.c
```diff
@@ -25,6 +25,12 @@ return values, struct complex arguments work without segfaulting.
 #include "npy_cblas.h"
 #include "fortran_defs.h"
 
+#ifdef HAVE_BLAS_ILP64
+/* NB: this redefines F_FUNC */
+#include "_blas64_defines.h"
```
rgommers, line 30: This looks slightly fishy since `npy_blas.h` is included higher up as well as in `_blas64_defines.h`. Haven't tried yet, but I'd expect this to give compiler warnings as well.

scipy/integrate/vode.pyf
```diff
@@ -50,11 +50,18 @@ python module zvode__user__routines
 end python module zvode__user__routines
 
 python module _vode
+  usercode '''
+#ifdef HAVE_BLAS_ILP64
+typedef npy_int64 F_INT;
```
rgommers: In existing code (`odrpack.h`, `_odepackmodule.c`) this is `#define F_INT npy_int64` instead. Not a big issue if both work, but might be nice to keep it uniform?

scipy/interpolate/meson.build
```diff
@@ -123,13 +128,14 @@ py3.extension_module('_rgi_cython',
 
 __fitpack_lib = static_library('__fitpack',
     ['src/__fitpack.h', 'src/__fitpack.cc'],
-    dependencies:[lapack, np_dep, py3_dep],
+    dependencies:[lapack_ilp64, np_dep, py3_dep],
```
rgommers: TODO: continue review here.

scipy/interpolate/meson.build
```diff
@@ -151,12 +162,20 @@ py3.extension_module('_fitpack',
...
 )
 
-# TODO: Add flags for 64 bit ints
+if use_ilp64
+   # generator only accepts strings, not files
+   f2c_map_file = f2py_ilp64_opts[1]
+   extra_arg = f2py_ilp64_opts[0] + '=' + fs.parent(f2c_map_file) / fs.name(f2c_map_file)
+  _dfitpackmodule_obj = f2py_gen.process('src/dfitpack.pyf', extra_args: extra_arg)
+else
+  _dfitpackmodule_obj = f2py_gen.process('src/dfitpack.pyf')
```
rgommers: It'd be cleaner to keep this line in its original location and only add `extra_args` to it (empty or containing ilp64 f2cmap).

scipy/linalg/meson.build
```diff
@@ -66,8 +66,10 @@ fblas_module = custom_target('fblas_module',
 # LAPACK - we have historically put these in `_fblas`.
 py3.extension_module('_fblas',
   fblas_module,
+  fortran_args: _fflag_lp64,
   link_args: version_link_args,
   dependencies: [lapack_dep, blas_dep, fortranobject_dep],
+  link_with: [g77_abi_wrappers],
```
rgommers: Please remove this line, it's already included in `lapack_dep` and `blas_dep`.

scipy/linalg/meson.build
```diff
@@ -66,8 +66,10 @@ fblas_module = custom_target('fblas_module',
 # LAPACK - we have historically put these in `_fblas`.
 py3.extension_module('_fblas',
   fblas_module,
+  fortran_args: _fflag_lp64,
```
rgommers: This should also be included in `blas_dep`, `lapack_dep`.

scipy/linalg/meson.build
```diff
@@ -95,12 +97,74 @@ flapack_module = custom_target('flapack_module',
 py3.extension_module('_flapack',
   flapack_module,
   c_args: [Wno_empty_body],
+  fortran_args: _fflag_lp64,
```
rgommers: Same here

scipy/meson.build
```diff
@@ -292,6 +298,30 @@ else
...
+  )
+  # Insert a second try with MKL, because we may be rejecting older versions
+  # or missing it because no pkg-config installed. If so, we need to retry
+  # with MKL SDL, and drop the version constraint (this always worked).
+  if not blas.found() and mkl_may_use_sdl
+    blas = dependency('mkl', modules: ['sdl: true'], required: false)
+  endif
+endif
+
+# fallback BLAS detection
```
rgommers: This isn't a detection at all, this is a declaration that wraps together the dependency object and the compile args that go with it. I think the code is fine, just this comment should be deleted.

scipy/meson.build
```diff
@@ -335,6 +369,104 @@ else
   )
 endif
 
+# Reuse the names, so we ensure we don't lose the arguments wrapped in with
+# declare_dependency. Also, avoids changing `dependencies: blas` to blas_dep in other files.
+# XXX: unused, remove?
```
rgommers: This is a bit messy indeed. I'd be inclined to say that we should make this more consistent at the end and use the `_dep` flavors everywhere.

scipy/meson.build
```diff
@@ -335,6 +369,104 @@ else
...
+     c_flags_ilp64 += ['-DBLAS_SYMBOL_SUFFIX=_64']
+
+  elif blas_name == 'scipy-openblas'
+      # scipy_openblas64, a separate library
+      blas_ilp64 = dependency('scipy-openblas64')
+      lapack_ilp64 = blas_ilp64
+
+      if not blas_ilp64.found()
+        error('scipy-openblas64 not found.')
+      endif
```
rgommers: These 3 lines are not needed, since the `dependency` calls above will fail if the library isn't found. A separate check is only useful if the detection is done with `required: false`

scipy/meson.build
```diff
@@ -335,6 +369,104 @@ else
...
+    endif
+    message(f'BLAS symbol suffix (ILP64): @blas_symbol_suffix@')
+  endif
+  _blas_incdir = []
+  if blas_symbol_suffix != ''
+    _args_blas_ilp64 += ['-DBLAS_SYMBOL_SUFFIX=' + blas_symbol_suffix]
+    _blas_incdir = ['.']
+  endif
+  # Declare the ILP64 dependencies
+  message('BLAS / LAPACK ILP64 detected: ', blas_ilp64.name(), ', ', lapack_ilp64.name())
```
rgommers: Remove these two lines I'd say. The `message` is duplicate; the regular `dependency()` calls already give useful log output.

scipy/linalg/meson.build
```diff
@@ -95,12 +97,74 @@ flapack_module = custom_target('flapack_module',
...
+  )
+
+  py3.extension_module('_fblas_64',
+    fblas64_module,
+    #['_fblas_64module.c'],
+    fortran_args: _fflag_ilp64,
+    link_args: version_link_args,
+    include_directories: ['../_build_utils/src'],   # for npy_cblas.h
+    dependencies: [lapack_ilp64, blas_ilp64, fortranobject_dep],
+    link_with: [g77_abi_wrappers_ilp64],
```
rgommers: Not needed, remove this line

scipy/linalg/meson.build
```diff
@@ -95,12 +97,74 @@ flapack_module = custom_target('flapack_module',
...
+
+  py3.extension_module('_flapack_64',
+    flapack64_module,
+    #['_flapack_64module.c'],
+    c_args: [Wno_empty_body],
+    fortran_args: _fflag_ilp64,
+    link_args: version_link_args,
+    include_directories: ['../_build_utils/src'],   # for npy_cblas.h
+    dependencies: [lapack_ilp64, blas_ilp64, fortranobject_dep],
+    link_with: [g77_abi_wrappers_ilp64],
```
rgommers: Same here, remove this line

scipy/meson.build
```diff
@@ -504,13 +636,51 @@ fortran_ignore_warnings = ff.get_supported_arguments(
...
+  f2py_ilp64_opts = ['--f2cmap', int64_f2cmap]
+endif
+
+if blas_symbol_suffix != ''
+  # We need to patch source files that use BLAS/LAPACK symbols.
+  # In addition, we now need to enable the Fortran preprocessor on all targets
+  # that depend on BLAS/LAPACK
+  # Note: this came from `scipy/build_utils/_fortran.py` in the distutils build.
+  _fflag_lp64 += _fflag_preprocess  # TODO: propagate _fflag_lp64
+  _fflag_ilp64 += _fflag_preprocess
```
rgommers: Can you include this in the `declare_dependency` calls right below, and then remove the separate usages elsewhere? And the `TODO` comment?

scipy/meson.build
```diff
@@ -335,6 +369,104 @@ else
...
+     c_args: _args_blas_ilp64,
+     gnu_symbol_visibility: 'hidden',
+   )
+else
+  # we're not using ILP64; user code will link to the always-available LP64 blas/lapack
+  # (all users must use preprocessor macros BLAS_NAME to handle the two options)
+  blas_ilp64 = blas
+  lapack_ilp64 = lapack
+  c_flags_ilp64 = []
+  message('LAPACK ILP64 not requested / not detected.')
```
rgommers: I think this whole `else` block needs deleting, it's more confusing than helpful.

scipy/meson.build
```diff
@@ -504,13 +636,51 @@ fortran_ignore_warnings = ff.get_supported_arguments(
...
+
+# Gfortran does run the preprocessor for .F files, and PROPACK is the only
+# component which needs the preprocessor (unless we need symbol renaming for
+# blas_symbol_suffix).
+_fflag_preprocess = []
+_gfortran_preprocess = ['-cpp', '-ffree-line-length-none', '-ffixed-line-length-none']
+if ff.has_multi_arguments(_gfortran_preprocess)
+  _fflag_preprocess = _gfortran_preprocess
+else
+  _fflag_preprocess = ff.first_supported_argument(['-fpp', '/fpp', 'cpp'])
```
rgommers: The `cpp` at the end here is missing a dash, that can't be correct.

scipy/special/meson.build
```diff
@@ -176,7 +176,7 @@ py3.extension_module('_ellip_harm_2',
   cpp_args: ['-DSP_SPECFUN_ERROR'],
   include_directories: ['..', '../_lib', '../_build_utils/src'],
   link_args: version_link_args,
-  dependencies: [xsf_dep, lapack_dep, np_dep],
+  dependencies: [xsf_dep, np_dep, lapack_dep],
```
rgommers: nit: this is an empty change now, drop it?

rgommers (changes requested):
Looks pretty close - I have lots of comments, but they should be easy to address.

scipy/_build_utils/src/wrap_g77_abi.c
ev-br: I don't see any compiler warnings locally on linux at least. Not sure what the problem is that you're after. We need an `#include "npy_cblas.h"` for both LP64 and ILP64 builds, so in principle we can remove one include from `_blas64_defines.h` but why.

scipy/meson.build
ev-br: Confusing yes, helpful also yes though.

The key issue is we need to somehow express "link to 64-bit LAPACK if available else to 32-bit LAPACK". Therefore there's a need for an `if-else` somewhere. Currently it's here, so that all extensions simply depend on `lapack_ilp64` which may be ILP64 or LP64 depending on the meson option.

Alternatives:
- rename the dependency to `lapack_maybe_ilp64` or some such, or
- push the if-else to individual extensions  (`dependcies: [lapack_ilp64 if available lapack_dep otherwise]`).

Are you suggesting the second variant?

ev-br (commented):
Thanks for the review @rgommers. Let me clarify two larger items first, (and I'll address smaller items in a separate update)

scipy/meson.build
ev-br: I tried this in https://github.com/scipy/scipy/pull/22743/commits/814dc4634dae82375eb216eaa008bb57120d472d but the compile flag gets added to both Fortran and C sources, which then triggers

```
[15/19] Compiling C object scipy/linalg/_fblas_64.cpython-312-x86_64-linux-gnu.so.p/meson-generated_..__fblas_64module.c.o
cc1: warning: command-line option '-fdefault-integer-8' is valid for Fortran but not for C
```

What would be a way only apply `compile_args` to Fortran sources and not C sources?

ev-br:
Okay, I think I addressed all review comments modulo two higher-level questions, https://github.com/scipy/scipy/pull/22743#pullrequestreview-3020065024 and a techical glitch, https://github.com/scipy/scipy/pull/22743#discussion_r2207857664


> If MKL works, I'd expect Accelerate to work as well pretty much unchanged. Did you test? If not, should I?

IIRC name mangling needs a tweak. Is a bit annoying to debug on CI, so I did not go all that far.

scipy/meson.build
rgommers: Oh grumble, yeah I've run into that before, there's no language-specific `compile_args`. We'll need two separate `_dep` objects, one for C/C++ and one for targets that also contain Fortran code.

Let me actually test the latest changes and then circle back to this.

ev-br: > We'll need two separate _dep objects, one for C/C++ and one for targets that also contain Fortran code.

To make it more "fun", some targets need `-fdefault-integer-8` but do not link to LAPACK (dfitpack is a prime example). 
All in all, if we expect that all fortran targets are gone in a not-so-distant future, I can't help wondering if the cure is not worse than the decease

.github/workflows/linux_blas_ilp64.yml
```diff
@@ -0,0 +1,129 @@
+name: BLAS tests (Linux)
```
rgommers: These jobs aren't running: https://github.com/scipy/scipy/actions/workflows/linux_mkl_ilp64.yml

ev-br: Oops, a bad change yesterday. Should be fixed now: https://github.com/scipy/scipy/actions/runs/16315705534

scipy/meson.build
rgommers: No, I'm not suggesting the second variant. We need clear names for the first variant. We have `2**3` options:
- BLAS and LAPACK
- LP64-only and maybe-ILP64
- C/C++ and C/C++/Fortran

So how about we have eight `_dep` objects, and each build target will depend on exactly one of those objects with no other dependencies or compile arg objects needing to be supplied?

Names can be: 
- `blas_lp64_dep`
- `blas_dep`
- `blas_lp64_fortran_dep`
- `blas_fortran_dep`

and the same for `lapack_xxx`? That way `blas_dep` is the generic one, may be LP64 or ILP64 which everything should support in the future; `blas_lp64` is the special case. And the `_fortran` flavors will be disappearing completely over the next year.

Also `lapack_xxx_dep` should include `dependencies: [blas_xxx]` so that we no longer have to depend on both blas and lapack anywhere.

rgommers: `message` call can go since the info is already present both in the build invocation and in this output:
```
Build targets in project: 199

scipy 1.17.0.dev0+git20250716.9aeb39e

  Subprojects
    boost_math: YES
    highs     : YES
    qhull_r   : YES
    xsf       : YES

  User defined options
    blas      : mkl
    prefix    : /usr
    use-ilp64 : true
```

scipy/_build_utils/src/wrap_g77_abi.c
rgommers: I was thinking `F_FUNC` was in `npy_cblas.h`, but it isn't. That said, this still looks very odd - I'm unclear on the logic. `F_FUNC` goes from _only_ dealing with Fortran uppercase/append to dealing with BLAS prefix/suffix, which isn't its original function at all.

Its original purpose is, from `_build_utils/src/fortran_defs.h`:
```c
/*
 * Handle different Fortran conventions.
 */

#if defined(NO_APPEND_FORTRAN)
#if defined(UPPERCASE_FORTRAN)
#define F_FUNC(f,F) F
#else
#define F_FUNC(f,F) f
#endif
#else
#if defined(UPPERCASE_FORTRAN)
#define F_FUNC(f,F) F##_
#else
#define F_FUNC(f,F) f##_
#endif
#endif
```

After you redefine it in `_blas64_defines.h` it becomes:

```c
#define BLAS_FUNC(name) BLAS_FUNC_EXPAND(name,BLAS_SYMBOL_PREFIX,BLAS_SYMBOL_SUFFIX,BLAS_FORTRAN_SUFFIX)
#define F_FUNC(f, F) BLAS_FUNC(f)
```

I'm not following that logic.

scipy/meson.build
rgommers: > To make it more "fun", some targets need `-fdefault-integer-8` but do not link to LAPACK (dfitpack is a prime example).

`dfitpack` does link to lapack?

ev-br: No, it doesn't

scipy/_build_utils/src/wrap_g77_abi.c
ev-br: There's a brief discussion in the commit message of https://github.com/scipy/scipy/pull/22743/commits/76b30881a0d2d1030bb2b369ccaf599fa581025b, but I see it's too brief, so let me try rephrasing.

In the f2py-generated code, `F_FUNC` macro comes from f2py itself: https://github.com/numpy/numpy/blob/main/numpy/f2py/cfuncs.py#L112
The name of the macro and its content are fixed but f2py itself and I don't see a way to change it. Now, if you look into the f2py-generated `build/scipy/linalg/_fblas64_module.c`:

```
/********************** See f2py2e/cfuncs.py: cppmacros **********************/

#if defined(PREPEND_FORTRAN)
#if defined(NO_APPEND_FORTRAN)
#if defined(UPPERCASE_FORTRAN)
#define F_FUNC(f,F) _##F
#else
#define F_FUNC(f,F) _##f
#endif
#else
#if defined(UPPERCASE_FORTRAN)
#define F_FUNC(f,F) _##F##_
#else
#define F_FUNC(f,F) _##f##_
#endif
#endif
#else
#if defined(NO_APPEND_FORTRAN)
#if defined(UPPERCASE_FORTRAN)
#define F_FUNC(f,F) F
#else
#define F_FUNC(f,F) f
#endif
#else
#if defined(UPPERCASE_FORTRAN)
#define F_FUNC(f,F) F##_
#else
#define F_FUNC(f,F) f##_
#endif
#endif
#endif

// ... snip ...

/********************* See f2py2e/capi_rules.py: usercode *********************/
  /* start usercode multiline (0) */
 
#include "_blas64_defines.h"

  /* end multiline (0)*/

/* See f2py2e/rules.py */
extern void F_FUNC(srotg,SROTG)(float*,float*,float*,float* );
extern void F_FUNC(drotg,DROTG)(double*,double*,double*,double* );
extern void F_FUNC(crotg,CROTG)(complex_float*,complex_float*,complex_float*,complex_float* );
extern void F_FUNC(zrotg,ZROTG)(complex_double*,complex_double*,complex_double*,complex_double* );

//....
```

So what happens is that f2py defines the `F_FUNC` macro (we cannot change it), then we include `_blas64_defines.h` from the usercode section, and then `F_FUNC` macro is used to mangle the BLAS names.

So in the usercode section, we undefine the not-useful `F_FUNC` macro and redefine it with `BLAS_FUNC` from `npy_cblas.h` --- which does the LP64/ILP64 name mangling we want.

rgommers: Ah thanks - that would be really helpful to put in a code comment.

scipy/meson.build
rgommers: It's right there in the extension module definition, so can you please clarify if you mean something else?

https://github.com/scipy/scipy/blob/ef5faacd6356bfd139ff9582887e9ddadfd608b3/scipy/interpolate/meson.build#L155-L159

ev-br: It was on main, but it's not needed and is removed in this PR: 
https://github.com/scipy/scipy/pull/22743/files#diff-85ab739d8d3eb989d275834cd1da1d3302bedb33b8ecbef8656831d6ad026edbR176
(no idea why it was there at all, might have been my previous error)

scipy/_build_utils/src/wrap_g77_abi.c
ev-br: Any suggestions as to where to add this comment? (I stared for too long too recently into this, my eyes are biased)

rgommers: Maybe extend the code comment in `_blas_defines.h` with what's in the commit message?

ev-br: Thanks, done.

rgommers: Great, resolving this

scipy/meson.build
ev-br: 8 dependency objects, ugh.
How about dropping the `_fortran` variants now? ATM, we only use it for `dfitpack` and we don't test it even there (and with my `scipy.interpolate` hat on, 64-bit integer in FITPACK is not crucial). Other remaining Fortran modules (vode, dop, odepack) we're not touching in this PR anyway.

ATM FITPACK served as a poster child for ILP64 support; we can use the new batched linalg instead (in a follow-up PR, which I've on my TODO list anyway). 

If we do this, we're down to four dependency objects, and with a `lapack -> blas` dependency, submodules only need two objects. So it is then basically a rename, `lapack_dep -> lapack_lp64_dep` and `lapack_ilp64 -> lapack_dep`. WDYT?

rgommers: Okay great, let's drop the Fortran variants - let's just add a code comment above the four `_dep` objects that we didn't create those due to limited need, but it could be done if Fortran usage unexpectedly proliferates.

> So it is then basically a rename, `lapack_dep -> lapack_lp64_dep` and `lapack_ilp64 -> lapack_dep`. WDYT?

SGTM

ev-br: Okay, I believe all comments are addressed: lapack_* dependencies are renamed, there are no spurious build warnings, and the smoke tests,

```
$ spin test -- -k'large_vector'
```

passes locally with both MKL and scipy-openblas.

rgommers: Thanks, this looks good now - resolving.

scipy/_build_utils/meson.build
```diff
@@ -12,7 +12,7 @@ if generate_blas_wrappers
   blas_lapack_wrapper_lib = static_library('blas_lapack_wrapper_lib',
     blas_lapack_wrappers,
     c_args: numpy_nodepr_api,
-    dependencies: [blas_dep, lapack_dep, np_dep, py3_dep],
+    dependencies: [lapack_lp64_dep, np_dep, py3_dep],
```
rgommers, line 15: This can't be right, can it? The purpose of this wrapper should be at least as relevant for ILP64. xref gh-19816 for why this was introduced, it already mentions ILP64.

scipy/interpolate/meson.build
rgommers: Resolving - this indeed cannot be done perfectly right now, because `declare_dependency` isn't language-specific.

scipy/meson.build
```diff
@@ -504,13 +622,42 @@ fortran_ignore_warnings = ff.get_supported_arguments(
...
+# blas_symbol_suffix).
+_fflag_preprocess = []
+_gfortran_preprocess = ['-cpp', '-ffree-line-length-none', '-ffixed-line-length-none']
+if ff.has_multi_arguments(_gfortran_preprocess)
+  _fflag_preprocess = _gfortran_preprocess
+else
+  _fflag_preprocess = ff.first_supported_argument(['-fpp', '/fpp', '-cpp'])
+endif
+
+_fflag_lp64 = []
```
rgommers: `_fflag_lp64` is never defined to anything, so can be removed.

scipy/_build_utils/meson.build
ev-br: It is certainly not needed for MKL and OpenBLAS, and I cannot easily test Accelerate. Maybe it's the missing bit for Accelerate; however `generate_blas_wrappers = true` for openblas, too, and things work as they are now. Not sure.

rgommers:
> currently several submodules (optimize, special, cluster, odr) are using 32-bit integer size linear algebra; it is not very clear which of these really need 64-bit integers (SLSQP with variables `2**31`, cough).

This stood out more to me after the change to `*_lp64_dep`. A lot of these should already be ready for ILP64, e.g. `special` (gh-12085), `lbfbgs` (gh-12009), `trlib` (gh-12030). See all the PRs linking to gh-11193. Did that regress in the meantime? If not, it seems easier to leave them using `blas_dep`/`lapack_dep`.

scipy/_build_utils/meson.build
ev-br: Actually, 

```
$ cat build/scipy/_build_utils/blas_lapack_wrappers.c  |more
/*
This file was generated by _generate_blas_wrapper.py.
Do not edit this file directly.
*/


#include "npy_cblas.h"
#include "fortran_defs.h"

// ... snip ....

void BLAS_FUNC(caxpy)(int *n, npy_complex64 *ca, npy_complex64 *cx, int *incx, npy_complex64 *cy, int *incy);
void F_FUNC(caxpy,CAXPY)(int *n, npy_complex64 *ca, npy_complex64 *cx, int *incx, npy_complex64 *cy, int *incy){
    return BLAS_FUNC(caxpy)(n, ca, cx, incx, cy, incy);
}

// ...

```

So this basically makes a bunch of trampolines from `F_FUNC` to `BLAS_FUNC`? If so, the ILP64 version is using the `#undef F_FUNC` construction instead which we discussed here yesterday. 

EDIT: this one: https://github.com/scipy/scipy/pull/22743#discussion_r2210848534

ev-br:
> This stood out more to me after the change to *_lp64_dep. A lot of these should already be ready for ILP64, e.g. special (gh-12085), lbfbgs (gh-12009), trlib (gh-12030). See all the PRs linking to gh-11193. Did that regress in the meantime?

Yes, a lot of them did regress while we did not have a CI job.
Pretty much all Fortran->C rewrites are not ILP64-ready (e.g. https://github.com/scipy/scipy/blob/main/scipy/optimize/__lbfgsb.h#L24 and a bunch of `int` variables elsewhere in that file).

scipy/_build_utils/meson.build
rgommers: > So this basically makes a bunch of trampolines from `F_FUNC` to `BLAS_FUNC`? If so, the ILP64 version is using the `#undef F_FUNC` construction instead which we discussed here yesterday.

I think that's part of why that confused me. All symbol prefixes/suffixes were handled by this generated wrapper. ILP64 adds symbol suffixes, so I expected it to be handled by these wrappers as well (of course that's made more complex by us needing both LP64 and ILP64). Don't we now have two similar but different ways of doing the same thing?

In addition, `_generate_blas_wrappers.py` it generates wrappers for all LAPACK functions as well. So I think what could be done in principle is to call `_generate_blas_wrappers.py` twice, once for LP64 wrappers and once for ILP64 wrappers - after updating `scipy/_build_utils/_wrappers_common.py` and whatever that pulls in for ILP64 (so we'd have `cython_lapack64_signatures.txt` in addition to `cython_lapack_signatures.txt`).

- `_blas64_defines.h` is reintroducing what we had with `distutils` way back when, but is specific to `fblas_64.pyf.src` and `flapack_64.pyf.src`, and our `f2py` signature file method.
- The `blas_lapack_wrappers` method is using generated wrappers for the Cython API, **and** was used for all symbol suffix/prefix handling in current LP64 usage (`blas_dep`, `lapack_dep`). 
- xref gh-20682 which is related to the difference between these methods.
 
* * *

Okay, I think I finally have it sharp again how this all worked. I think we should merge this PR as is, but also discuss this topic further since we now both are looking at it in detail. This brings a bunch of extra tech debt as it stands (disclaimer: in my current understanding, which may still be missing something) because it splits suffix/prefix handling into two and leaves ILP64 LAPACK incomplete while LP64 LAPACK is complete.

rgommers:
Okay, CI is happy and review comments were addressed, so I'll do one last pass with maybe a few last cleanups, and look into Accelerate locally, then get this in.

@ev-br the commit history is a bit messy, I'm considering squash-merging it - unless you have a different preference?

ev-br:
> @ev-br the commit history is a bit messy, I'm considering squash-merging it - unless you have a different preference?

+1 to squash merge, yes.

scipy/_build_utils/meson.build
ev-br: Agreed that we now have two ways of doing the same thing. And that it's best to unify it. 

As to which way to use. I'd suggest to get rid of the wrappers, actually. Both ways work around f2py ways (cython_lapack piggy-backs onto the same f2py-d module IIUC); one is a dozen lines, while the other is autogenerating a 1000+ wrapper functions. 
Either way, it'd be best to do in a follow-up PR.

EDIT: IOW, currently `#undef F_FUNC` is behind the `#ifdef HAVE_BLAS_ILP64`; I'm pretty sure we can _almost_ have it if we do the same thing twice, with and without `HAVE_BLAS_ILP64`.

EDIT2 :  
> leaves ILP64 LAPACK incomplete while LP64 LAPACK is complete.

ILP64 LAPACK is complete, what you're saying is missing?

ev-br: IOW,

```
$ nm _flapack_64.cpython-312-x86_64-linux-gnu.so |grep dtrsyl
000000000019aee0 d doc_f2py_rout__flapack_64_dtrsyl
                 U dtrsyl_64_
000000000005e9c0 t f2py_rout__flapack_64_dtrsyl
$ nm _flapack.cpython-312-x86_64-linux-gnu.so |grep dtrsyl
000000000017dee0 d doc_f2py_rout__flapack_dtrsyl
                 U dtrsyl_
000000000009a9f0 t f2py_rout__flapack_dtrsyl
```

So my suggestion is to generate the latter using the same way as the former instead of generating wrappers.
Then `cython_lapack` can link `_fblas.so` the same way is before, since the symbols names are correct.

.github/workflows/linux_blas_ilp64.yml
```diff
@@ -0,0 +1,130 @@
...
+  contents: read # to fetch code (actions/checkout)
+
+jobs:
+
+  mkl-lp64:
+    runs-on: ubuntu-latest
+    name: "MKL LP64"
+    steps:
+    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
```
rgommers: Odd blank line here.

.github/workflows/linux_blas_ilp64.yml
```diff
@@ -0,0 +1,130 @@
...
+    - name: Build with defaults (LP64)
+      run: |
+        pkg-config --libs mkl-dynamic-lp64-seq  # check link flags
+        spin build -S-Dblas=mkl
+
+    - name: Test
+      run: spin test -j2
+
+
+  mkl-ilp64:
```
rgommers: Unify with job above? EDIT: leave that for some other time

doc/source/building/blas_lapack.rst
```diff
@@ -96,6 +96,59 @@ user wants to override this autodetection mechanism for building against plain
...
+LAPACK variants, where some extensions link to the ILP64 variant, while other
+extensions link to the LP64 variant.
+
+Building with ILP64 support requires several NumPy additions to ``meson``, which have
+not been merged to upstream yet::
+
+    $ pip install git+https://github.com/numpy/meson.git@main-numpymeson
+
+For a development build with MKL, install the library and its development headers, and
+give use the ``ilp64=true`` build option::
```
rgommers: typo "give use"

doc/source/building/blas_lapack.rst
```diff
@@ -96,6 +96,59 @@ user wants to override this autodetection mechanism for building against plain
...
+Support for ILP64 BLAS and LAPACK is still experimental; at the time of writing
+(July 2025) it is only available for two BLAS/LAPACK configurations: MKL and
+``scipy-openblas``.
+
+SciPy always requires LP64 (32-bit integer size) BLAS/LAPACK. You can build SciPy
+with *additional* ILP64 support. This will result in SciPy requiring both BLAS and
+LAPACK variants, where some extensions link to the ILP64 variant, while other
+extensions link to the LP64 variant.
+
+Building with ILP64 support requires several NumPy additions to ``meson``, which have
```
rgommers: rephrase

scipy/interpolate/meson.build
```diff
@@ -87,10 +87,15 @@ fitpack_src = [
   'fitpack/pardtc.f'
 ]
 
-# TODO: Add flags for 64 bit ints
+if use_ilp64
+  _fflag_intsize = _fflag_ilp64
+else
+  _fflag_intsize = []
```
rgommers: move up

scipy/meson.build
```diff
@@ -301,40 +330,129 @@ elif lapack_name == 'openblas'
...
+lapack_lp64_dep = declare_dependency(
+  dependencies: [lapack, blas_lp64_dep],
   compile_args: _args_blas_lapack
 )
 
 dependency_map = {
-  'BLAS': blas,
-  'LAPACK': lapack,
+  'BLAS': blas_lp64_dep,
+  'LAPACK': lapack_lp64_dep,
```
rgommers: This change yields broken output for `scipy.show_config`:
```
Build Dependencies:
  blas:
    detection method: internal
    found: true
    include directory: unknown
    lib directory: unknown
    name: dep123773294823904
    pc file directory: unknown
    version: 1.17.0.dev0+git20250717.ee3cf12
```

ilayn:
> Pretty much all Fortran->C rewrites are not ILP64-ready

I did not want to barge in to the discussion so I'm waiting for this to stabilize and I'll start the F_INTing process with them so no worries. I don't want to release ARPACK only 32bit capable state.

ilayn:
Regarding the Cython based BLAS/LAPACK we learned to mimic f2py by stride tricks and silent F-major conversions etc. in our linalg work. Hence we can follow the

argument sanitizing -> Cython wrapper -> BLAS/LAPACK linkage

instead of 

f2py -> argument sanitizing -> Fortran wrapper -> BLAS LAPACK linkage

That's basically the nice thing to remove Fortran compiler dependence in its entirety contingent upon the remaining code disappearing.

rgommers:
I did a full set of wheel builds [CI log](https://github.com/rgommers/scipy/actions/runs/16352000484) and local testing for Accelerate - all looks fine. 

A few more little things to fix up, I'll push them to this branch.

Could you open a new tracking issue for per-module ILP64 support wherever that needs to be added back?

scipy/_build_utils/src/npy_cblas.h
```diff
@@ -26,6 +26,21 @@ enum CBLAS_SIDE {CblasLeft=141, CblasRight=142};
...
+    #if __MAC_OS_X_VERSION_MAX_ALLOWED < 130300
+        #ifdef HAVE_BLAS_ILP64
+            #error "Accelerate ILP64 support is only available with macOS 13.3 SDK or later"
+        #endif
+    #else
+       /* #define NO_APPEND_FORTRAN */
+        #ifdef HAVE_BLAS_ILP64
+            #define BLAS_SYMBOL_SUFFIX $NEWLAPACK$ILP64
+        #else
+            #define BLAS_SYMBOL_SUFFIX $NEWLAPACK
```
rgommers, line 39: This is undone/redone a few lines below, but that can be cleaned up when adding Accelerate ILP64 support in a follow-up PR.

rgommers (approved):
Submitting last few review comments for which I just pushed two commits.

scipy/_build_utils/meson.build
rgommers: What's incomplete in both actually is a header with LAPACK signatures. I think we can get them from `scipy/linalg/cython_lapack_signatures.txt` fairly easily (`blas_lapack_wrappers.h` in addition to `blas_lapack_wrappers.c`), but not from the f2py method.

> I'd suggest to get rid of the wrappers, actually. Both ways work around f2py ways (cython_lapack piggy-backs onto the same f2py-d module IIUC

No it doesn't use that (unless I'm really missing something). That would mean reusing `py3.extension_module('_flapack',` or `flapack_module = custom_target('flapack_module',` in `linalg/meson.build`.

The wrapper method is more complete, easier to update when we move to a new minimum LAPACK version, allows us to generate the missing header, and doesn't depend on `f2py` at all. 

I think we park this for a while, no need to do anything now. I'll have a go at updating my docs PR for this and get that merged.

rgommers:
All right, it's in. Thanks a lot Evgeni!

ev-br:
> Could you open a new tracking issue for per-module ILP64 support wherever that needs to be added back?

https://github.com/scipy/scipy/issues/23351

tools/generate_f2pymod.py
```diff
@@ -290,10 +293,13 @@ def main():
 
     # Now invoke f2py to generate the C API module file
     if args.infile.endswith(('.pyf.src', '.pyf')):
-        p = subprocess.Popen(
-            ['f2py', fname_pyf, '--build-dir', outdir_abs] + nogil_arg,
-            stdout=subprocess.PIPE, stderr=subprocess.PIPE, cwd=os.getcwd()
-        )
+        cmd = [sys.executable, '-m', 'numpy.f2py', fname_pyf,
```
rgommers, line 296: Missed this in review, so a note for the record: this caused a regression (see gh-24406). `sys.executable` should not be used.

ev-br: Oh wow so one exception to the knee-jerk reaction of "use sys.executable -m entry-point".

PR #22749, opened 2025-03-27
ENH: stats: add support for multiple parameterizations for custom distributions in new distribution infra

steppi:
#### Reference issue

There is no issue, but this was discussed in https://github.com/scipy/scipy/pull/22560#issuecomment-2676153771.

This updates `make_distribution` to allow for specifying custom distributions with multiple parameterizations following @mdhaber's suggestion here https://github.com/scipy/scipy/pull/22560#issuecomment-2676153771. Previously the `parameterization` attribute of user defined custom distribution classes could only be a dictionary defining a single parameterization. Now it can also be a tuple of such dictionaries. If a user defined custom distribution class has multiple parameterizations defined through such a tuple, the class also must define the method `_process_parameters` that allows mapping between parameterizations.

I added an example and test that defines a custom Beta distribution with two parameterizations, the regular one, and an alternative parameterization in terms of the mean $\mu = \frac{a}{a + b}$ and a dispersion parameter $\nu = a + b$. The alternative parameterization is used in [beta-binomial regression](https://www.researchgate.net/publication/7184584_A_better_lemon_squeezer_Maximum-likelihood_regression_with_beta-distributed_dependent_variables).

#### Additional information

@mdhaber, is this what you had in mind? Is it fine that methods like `pdf` should take all parameters from all parameterizations even if they aren't used? Is it fine that users have to define `_process_parameters` like this?

Also, this isn't really related to the content of this PR, but I noticed while writing tests that `CustomDistribution`'s raise when calling methods for which they don't define an implementation instead of using a default implementation. So for example, for the distribution from the test

```python
    class MyBeta:
            @property
            def __make_distribution_version__(self):
                return "1.16.0"

            @property
            def parameters(self):
                return (
                    {"a": (0, np.inf), "b": (0, np.inf)},
                    {"mu": (0, np.inf), "nu": (0, np.inf)},
                )

            def _process_parameters(self, a=None, b=None, mu=None, nu=None):
                if a is not None and b is not None and mu is None and nu is None:
                    nu = a + b
                    mu = a / nu
                elif mu is not None and nu is not None and a is None and b is None:
                    a = mu * nu
                    b = nu - a
                else:
                    raise ValueError("Invalid parameterization of MyBeta.")
                return {"a": a, "b": b, "mu": mu, "nu": nu}

            @property
            def support(self):
                return {'endpoints': (0, 1)}

            def pdf(self, x, a, b, mu, nu):
                return special._ufuncs._beta_pdf(x, a, b)

            def cdf(self, x, a, b, mu, nu):
                return special.betainc(a, b, x)

    MyBeta = stats.make_distribution(MyBeta())
    X = MyBeta(a=2.0, b=2.0)
```

`X.entropy()` will raise with

```
NotImplementedError: `CustomDistribution` does not provide an accurate implementation of the required method. Consider leaving `method` and `tol` unspecified to use another implementation.
```

Is that expected? Should we change it so that the default implementation is used instead? I haven't untangled exactly why this is happening yet.

### DIFF
--- a/scipy/stats/_distribution_infrastructure.py
+++ b/scipy/stats/_distribution_infrastructure.py
@@ -3689,8 +3689,8 @@ def make_distribution(dist):
     ...             nu = a + b
     ...             mu = a / nu
     ...         elif mu is not None and nu is not None and a is None and b is None:
-    ...             b = mu * nu
-    ...             a = nu - b
+    ...             a = mu * nu
+    ...             b = nu - a
     ...         else:
     ...             raise ValueError("Invalid parameterization of MyBeta.")
     ...         return {"a": a, "b": b, "mu": mu, "nu": nu}
@@ -3700,19 +3700,10 @@ def make_distribution(dist):
     ...         return {'endpoints': (0, 1)}
     ...
     ...     def pdf(self, x, a, b, mu, nu):
-    ...         return x**(a - 1) * (1 - x)**(b - 1) / special.beta(a, b)
-    ...
-    ...     def logpdf(self, x, a, b, mu, nu):
-    ...         return (a - 1) * np.log(x) + (b - 1)*np.log1p(-x) - special.betaln(a, b)
+    ...         return special._ufuncs._beta_pdf(x, a, b)
     ...
     ...     def cdf(self, x, a, b, mu, nu):
     ...         return special.betainc(a, b, x)
-    ...
-    ...     def moment(self, order, kind='raw', *, a, b, mu, nu):
-    ...         if order == 1 and kind == "raw":
-    ...             return mu
-    ...         if order == 2 and kind == "central":
-    ...             return mu * (1 - mu) / (1 + nu)
     >>>
     >>> MyBeta = stats.make_distribution(MyBeta())
     >>> X = MyBeta(a=2.0, b=2.0)
--- a/scipy/stats/_distribution_infrastructure.py
+++ b/scipy/stats/_distribution_infrastructure.py
@@ -3510,10 +3510,17 @@ def make_distribution(dist):
             is defined. The preferred interface may change in future SciPy versions,
             in which case support for an old interface version may be deprecated
             and eventually removed.
-        parameters : dict
-            Each key is the name of a parameter,
+        parameters : dict or tuple
+            If a ``tuple`` it should be a ``tuple`` of ``dict``s with the structure
+            described below. Each ``dict`` in the ``tuple`` will correspond to a
+            separate parametrization of the distribution. If multiple parameterizations
+            are specified, ``dist``'s class must also define a method
+           ``_process_parameters`` to allow mapping between the different parameterizations.
+           The requirements for this method are described further below.
+
+            If a ``dict``, each key is the name of a parameter,
             and the corresponding value is either a dictionary or tuple.
-            If a dictionary, it may have the following items, with default
+            If the value is a dictionary, it may have the following items, with default
             values used for entries which aren't present.
 
             endpoints : tuple, default: (-inf, inf)
@@ -3555,7 +3562,17 @@ def make_distribution(dist):
         ``sample`` method must accept ``shape`` by position or keyword, but contrary
         to the public method of the same name, the argument it receives will be the
         *full* shape of the output array - that is, the shape passed to the public
-        method prepended to the broadcasted shape of random variable parameters.
+        method prepended to the broadcasted shape of random variable parameters. If
+        multiple parameterizations are defined, these methods must accept all
+        parameters from all parameterizations.
+
+        If the class's ``parameters`` attribute contains a ``tuple`` defining multiple
+        parameterizations, then the class must also define a ``__process_parameters``
+        method for mapping between the parameterizations. It must take all parameters
+        from all parameterizations as optional keyword arguments and return a dictionary
+        mapping parameters to values, filling in values from other parameterizations
+        using values from the supplied parameterization.
+
 
     Returns
     -------
@@ -3649,6 +3666,58 @@ def make_distribution(dist):
     >>> X.support()
     (-8.0, 8.0)
     >>> np.isclose(X.cdf(2.1), Y.cdf(2.1))
+    np.True_
+
+    Create a custom distribution with multiple parameterizations. Here we create a
+    custom version of the Beta distribution that has an alternative parameterization
+    in terms of the mean ``mu`` and a dispersion parameter ``nu``.
+
+    >>> class MyBeta:
+    ...     @property
+    ...     def __make_distribution_version__(self):
+    ...         return "1.16.0"
+    ...
+    ...     @property
+    ...     def parameters(self):
+    ...         return (
+    ...             {"a": (0, np.inf), "b": (0, np.inf)},
+    ...             {"mu": (0, np.inf), "nu": (0, np.inf)},
+    ...         )
+    ...
+    ...     def _process_parameters(self, a=None, b=None, mu=None, nu=None):
+    ...         if a is not None and b is not None and mu is None and nu is None:
+    ...             nu = a + b
+    ...             mu = a / nu
+    ...         elif mu is not None and nu is not None and a is None and b is None:
+    ...             b = mu * nu
+    ...             a = nu - b
+    ...         else:
+    ...             raise ValueError("Invalid parameterization of MyBeta.")
+    ...         return {"a": a, "b": b, "mu": mu, "nu": nu}
+    ...
+    ...     @property
+    ...     def support(self):
+    ...         return {'endpoints': (0, 1)}
+    ...
+    ...     def pdf(self, x, a, b, mu, nu):
+    ...         return x**(a - 1) * (1 - x)**(b - 1) / special.beta(a, b)
+    ...
+    ...     def logpdf(self, x, a, b, mu, nu):
+    ...         return (a - 1) * np.log(x) + (b - 1)*np.log1p(-x) - special.betaln(a, b)
+    ...
+    ...     def cdf(self, x, a, b, mu, nu):
+    ...         return special.betainc(a, b, x)
+    ...
+    ...     def moment(self, order, kind='raw', *, a, b, mu, nu):
+    ...         if order == 1 and kind == "raw":
+    ...             return mu
+    ...         if order == 2 and kind == "central":
+    ...             return mu * (1 - mu) / (1 + nu)
+    >>>
+    >>> MyBeta = stats.make_distribution(MyBeta())
+    >>> X = MyBeta(a=2.0, b=2.0)
+    >>> Y = MyBeta(mu=0.5, nu=4.0)
+    >>> np.isclose(X.pdf(0.3), Y.pdf(0.3))
     np.True_
 
     """
@@ -3795,13 +3864,19 @@ def _get_domain_info(info):
 
 
 def _make_distribution_custom(dist):
-    parameters = []
+    dist_parameters = (
+        dist.parameters if isinstance(dist.parameters, tuple) else (dist.parameters, )
+    )
+    parameterizations = []
+    for parameterization in dist_parameters:
+        parameters = []
 
-    for name, info in dist.parameters.items():
-        domain_info, typical = _get_domain_info(info)
-        domain = _RealDomain(**domain_info)
-        param = _RealParameter(name, domain=domain, typical=typical)
-        parameters.append(param)
+        for name, info in parameterization.items():
+            domain_info, typical = _get_domain_info(info)
+            domain = _RealDomain(**domain_info)
+            param = _RealParameter(name, domain=domain, typical=typical)
+            parameters.append(param)
+        parameterizations.append(_Parameterization(*parameters) if parameters else [])
 
     domain_info, _ = _get_domain_info(dist.support)
     _x_support = _RealDomain(**domain_info)
@@ -3809,8 +3884,7 @@ def _make_distribution_custom(dist):
     repr_str = dist.__class__.__name__
 
     class CustomDistribution(ContinuousDistribution):
-        _parameterizations = ([_Parameterization(*parameters)] if parameters
-                              else [])
+        _parameterizations = parameterizations
         _variable = _x_param
 
         def __repr__(self):
@@ -3846,6 +3920,9 @@ def _make_distribution_custom(dist):
         CustomDistribution._moment_raw_formula = _moment_raw_formula
         CustomDistribution._moment_central_formula = _moment_central_formula
         CustomDistribution._moment_standardized_formula = _moment_standardized_formula
+
+    if hasattr(dist, '_process_parameters'):
+        setattr(CustomDistribution, "_process_parameters", getattr(dist, "_process_parameters"))
 
     support_etc = _combine_docs(CustomDistribution, include_examples=False).lstrip()
     docs = [
--- a/scipy/stats/_distribution_infrastructure.py
+++ b/scipy/stats/_distribution_infrastructure.py
@@ -3511,14 +3511,6 @@ def make_distribution(dist):
             in which case support for an old interface version may be deprecated
             and eventually removed.
         parameters : dict or tuple
-            If a ``tuple`` it should be a ``tuple`` of ``dict``s with the structure
-            described below. Each ``dict`` in the ``tuple`` will correspond to a
-            separate parametrization of the distribution. If multiple parameterizations
-            are specified, ``dist``'s class must also define a method
-            ``_process_parameters`` to allow mapping between the different
-            parameterizations. The requirements for this method are described further
-            below.
-
             If a ``dict``, each key is the name of a parameter,
             and the corresponding value is either a dictionary or tuple.
             If the value is a dictionary, it may have the following items, with default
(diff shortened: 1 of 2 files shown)


mdhaber:
> is this what you had in mind? Is it fine that methods like pdf should take all parameters from all parameterizations even if they aren't used? Is it fine that users have to define _process_parameters like this?

From the description, precisely! Yes, I intended for methods to receive all the parameters from all parameterizations. `_process_parameters` can even return named constants like `1/sqrt(2*np.pi)`, I think, so all methods get them. They should all be consisent with one another, so different methods can pick whichever is more convenient for them. We might re-think this when we think about mutability and getting over the limitations of parameter name conflicts, but it was intended to be a feature. Yes, users *should* define `_process_parameters`, otherwise, they would have to deal with the different parameterizations in every method.

> Is that expected? Should we change it so that the default implementation is used instead? 

That is not expected. Fortunately, it is not true in `main`. Adapting your `Beta` to have just one parameterization:

```python3
class MyBeta:
    @property
    def __make_distribution_version__(self):
        return "1.16.0"

    @property
    def parameters(self):
        return {"a": (0, np.inf), "b": (0, np.inf)}

    @property
    def support(self):
        return {'endpoints': (0, 1)}

    def pdf(self, x, a, b):
        return special._ufuncs._beta_pdf(x, a, b)

    def cdf(self, x, a, b):
        return special.betainc(a, b, x)


MyBeta = stats.make_distribution(MyBeta())
X = MyBeta(a=2.0, b=2.0)
X.entropy()  # np.float64(-0.1250928025613885) on main
```

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3651,6 +3669,49 @@ class or its methods for more information.
     >>> np.isclose(X.cdf(2.1), Y.cdf(2.1))
     np.True_
 
+    Create a custom distribution with multiple parameterizations. Here we create a
+    custom version of the Beta distribution that has an alternative parameterization
```
mdhaber, line 3673:
```suggestion
    custom version of the beta distribution that has an alternative parameterization
```

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3651,6 +3669,49 @@ class or its methods for more information.
...
+    ...
+    ...     def _process_parameters(self, a=None, b=None, mu=None, nu=None):
+    ...         if a is not None and b is not None and mu is None and nu is None:
+    ...             nu = a + b
+    ...             mu = a / nu
+    ...         elif mu is not None and nu is not None and a is None and b is None:
+    ...             a = mu * nu
+    ...             b = nu - a
+    ...         else:
+    ...             raise ValueError("Invalid parameterization of MyBeta.")
```
mdhaber, line 3696: Are you sure this is needed? The infrastructure should raise before getting here unless either (`a` and `b` are provided) OR (`mu` and `nu` are provided). See e.g.:
```python3
from scipy.stats._new_distributions import _LogUniform
_LogUniform(a=1, b=10).entropy()  # np.float64(1.9853249917449787)
_LogUniform(log_a=0, log_b=np.log(10)).entropy()  # np.float64(1.9853249917449787)
_LogUniform(log_a=0, b=np.log(10)).entropy()
# ValueError: The provided parameters `{b, log_a}` do not match a supported parameterization of the `_LogUniform` distribution family.
```
So you really should only have to do `if a is not None` and `else`, really.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3555,7 +3562,17 @@ def make_distribution(dist):
         ``sample`` method must accept ``shape`` by position or keyword, but contrary
         to the public method of the same name, the argument it receives will be the
         *full* shape of the output array - that is, the shape passed to the public
-        method prepended to the broadcasted shape of random variable parameters.
+        method prepended to the broadcasted shape of random variable parameters. If
+        multiple parameterizations are defined, these methods must accept all
+        parameters from all parameterizations.
+
+        If the class's ``parameters`` attribute contains a ``tuple`` defining multiple
+        parameterizations, then the class must also define a ``__process_parameters``
```
mdhaber, line 3570:
```suggestion
        parameterizations, then the class must also define a ``_process_parameters``
```
We *could* allow the users to make this a public function (`process_parameters`), instead.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3510,10 +3510,17 @@ def make_distribution(dist):
...
             in which case support for an old interface version may be deprecated
             and eventually removed.
-        parameters : dict
-            Each key is the name of a parameter,
+        parameters : dict or tuple
+            If a ``tuple`` it should be a ``tuple`` of ``dict``s with the structure
+            described below. Each ``dict`` in the ``tuple`` will correspond to a
+            separate parametrization of the distribution. If multiple parameterizations
+            are specified, ``dist``'s class must also define a method
+           ``_process_parameters`` to allow mapping between the different parameterizations.
```
mdhaber, line 3518: Right now, this looks redundant. If you intended it to be, maybe mention "see below".
```suggestion
            separate parametrization of the distribution. If multiple parameterizations
            are specified, ``dist``'s class must also define a method
           ``_process_parameters`` to map between the different parameterizations.
```

But actually, I think this will be the exception rather than the rule. I would start by describing only the dict. So basically, undo this change at the top, and modify only the first line: "If a dict, each key is the name of a parameter,...."

Then afterward you can describe what a *tuple* should be. I think this would both eliminate redundancy and be easier to read, tbh.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3651,6 +3669,49 @@ class or its methods for more information.
...
+    >>> class MyBeta:
+    ...     @property
+    ...     def __make_distribution_version__(self):
+    ...         return "1.16.0"
+    ...
+    ...     @property
+    ...     def parameters(self):
+    ...         return (
+    ...             {"a": (0, np.inf), "b": (0, np.inf)},
+    ...             {"mu": (0, np.inf), "nu": (0, np.inf)},
```
mdhaber, line 3685:
```suggestion
    ...             {"mu": (0, 1), "nu": (0, np.inf)},
```
Right?

Same in test.

mdhaber (commented):
Great start. This is really exciting! I'll review the code once we get to the bottom of the issue you reported.

scipy/stats/_distribution_infrastructure.py
steppi: Yes. It popped in my mind that I did this wrong as I was falling asleep last night. It's because I just copy-pasted to correct the second entry, and only changed the parameter names.

scipy/stats/_distribution_infrastructure.py
steppi: Sounds good.

steppi:
It's curious. I'm not seeing that strange error about an efficient `entropy` not being defined any longer, but the changes I've made are all very small and I'm not sure what exactly I did that fixed it. I've still left `entropy` out of the tests though because it seems there is a genuine `scipy.special` bug that still causes the test to fail. This bug impacts `scipy.stats.beta` too

```python
import scipy.stats

stats.beta(4.45014772e-308, 4.0, 8.0)
```

results in

```
~/scipy/build-install/lib/python3.12/site-packages/scipy/stats/_continuous_distns.py in ?(self, x, a, b)
    746         #                     gamma(a+b) * x**(a-1) * (1-x)**(b-1)
    747         # beta.pdf(x, a, b) = ------------------------------------
    748         #                              gamma(a)*gamma(b)
    749         with np.errstate(over='ignore'):
--> 750             return scu._beta_pdf(x, a, b)

OverflowError: Error in function ibeta_derivative<d>(%1%,%1%,%1%): Overflow Error
```.

The Boost ufuncs used in `stats` were originally in a separate extension module and were only added to `special` last year, and they have some quirks, such as not catching errors and dealing with them. I can make an issue about this.

I think I've addressed all of the other comments, and have tried to remove the redundancy in the documentation. Let me know what you think @mdhaber.

File: scipy/stats/_distribution_infrastructure.py [hunk 6]
```diff
@@ -3533,7 +3533,17 @@ def make_distribution(dist):
                 ``endpoints`` tuple above, and should define a subinterval of the
                 domain given by ``endpoints``.
 
-            A ``tuple`` value ``(a, b)`` is equivalent to ``{endpoints: (a, b)}``.
+            A ``tuple`` value ``(a, b)`` associated to a key in the ``parameters``
+            ``dict`` is equivalent to ``{endpoints: (a, b)}``.
```
mdhaber, line 3537: I think it might be easier to read with this as English; maybe elsewhere. 
```suggestion
            dictionary is equivalent to ``{endpoints: (a, b)}``.
```

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3533,7 +3533,17 @@ def make_distribution(dist):
...
+
+            Custom distributions with multiple parameterizations can be defined by
+            having the ``parameters`` attribute be a ``tuple`` of ``dict``s with
+            the structure described above. In this case, ``dist``'s class must also
+            define a method ``process_parameters` to map between the different
```
mdhaber, line 3542:
```suggestion
            define a method ``process_parameters`` to map between the different
```

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3533,7 +3533,17 @@ def make_distribution(dist):
...
+            parameterizations. It must take all parameters from all parameterizations
+            as optional keyword arguments and return a ``dict`` mapping parameters to
+            values, filling in values from other parameterizations using values from
+            the supplied parameterization.
```
mdhaber, line 3546:
```suggestion
            the supplied parameterization. See example.
```
While this is about as clear as it can be, it's worth mentioning to the confused reader that there is an example coming.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3548,14 +3558,16 @@ def make_distribution(dist):
...
+        the corresponding method of `ContinuousDistribution`. When multiple
+        parameterizations are defined, these methods must accept all parameters from
+        all parameterizations. The ``moment`` method must accept the ``order`` and
+        ``kind`` arguments by position or keyword, but may return ``None`` if a
+        formula is not available for the arguments; in this case, the infrastructure
+        will fall back to a default implementation. The ``sample`` method must accept
+        ``shape`` by position or keyword, but contrary to the public method of the
+        same name, the argument it receives will be the *full* shape of the output
+        array - that is, the shape passed to the public method prepended to the
+        broadcasted shape of random variable parameters.
```
mdhaber, line 3570: Just want to see if anything else changed. GH is not showing the diff very nicely.
```suggestion
        the corresponding method of `ContinuousDistribution`. 
        When multiple parameterizations are defined, these methods must accept
        all parameters from all parameterizations. The ``moment`` method
        must accept the ``order`` and ``kind`` arguments by position or keyword, but
        may return ``None`` if a formula is not available for the arguments; in this
        case, the infrastructure will fall back to a default implementation. The
        ``sample`` method must accept ``shape`` by position or keyword, but contrary
        to the public method of the same name, the argument it receives will be the
        *full* shape of the output array - that is, the shape passed to the public
        method prepended to the broadcasted shape of random variable parameters.
```

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3533,7 +3533,17 @@ def make_distribution(dist):
                 ``endpoints`` tuple above, and should define a subinterval of the
                 domain given by ``endpoints``.
 
-            A ``tuple`` value ``(a, b)`` is equivalent to ``{endpoints: (a, b)}``.
+            A ``tuple`` value ``(a, b)`` associated to a key in the ``parameters``
+            ``dict`` is equivalent to ``{endpoints: (a, b)}``.
+
+            Custom distributions with multiple parameterizations can be defined by
+            having the ``parameters`` attribute be a ``tuple`` of ``dict``s with
+            the structure described above. In this case, ``dist``'s class must also
```
mdhaber, line 3541:
```suggestion
            having the ``parameters`` attribute be a tuple of dictionaries with
            the structure described above. In this case, ``dist``\'s class must also
```

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3533,7 +3533,17 @@ def make_distribution(dist):
...
+            define a method ``process_parameters` to map between the different
+            parameterizations. It must take all parameters from all parameterizations
+            as optional keyword arguments and return a ``dict`` mapping parameters to
```
mdhaber, line 3544:
```suggestion
            as optional keyword arguments and return a dictionary mapping parameters to
```

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3510,10 +3510,10 @@ def make_distribution(dist):
             is defined. The preferred interface may change in future SciPy versions,
             in which case support for an old interface version may be deprecated
             and eventually removed.
-        parameters : dict
-            Each key is the name of a parameter,
+        parameters : dict or tuple
+            If a ``dict``, each key is the name of a parameter,
```
mdhaber, line 3514:
```suggestion
            If a dictionary, each key is the name of a parameter,
```

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3533,7 +3533,17 @@ def make_distribution(dist):
                 ``endpoints`` tuple above, and should define a subinterval of the
                 domain given by ``endpoints``.
 
-            A ``tuple`` value ``(a, b)`` is equivalent to ``{endpoints: (a, b)}``.
+            A ``tuple`` value ``(a, b)`` associated to a key in the ``parameters``
```
mdhaber, line 3536:
```suggestion
            A tuple value ``(a, b)`` associated to a key in the ``parameters``
```

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3651,6 +3664,47 @@ class or its methods for more information.
...
+    ...     @property
+    ...     def __make_distribution_version__(self):
+    ...         return "1.16.0"
+    ...
+    ...     @property
+    ...     def parameters(self):
+    ...         return (
+    ...             {"a": (0, np.inf), "b": (0, np.inf)},
+    ...             {"mu": (0, 1), "nu": (0, np.inf)},
+    ...         )
```
mdhaber, line 3681: Sorry, I can't help myself. To the user, I think two lines will look simpler than four, even if it's mostly whitespace.
```suggestion
    ...         return ({"a": (0, np.inf), "b": (0, np.inf)},
    ...                 {"mu": (0, 1), "nu": (0, np.inf)})
```

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3651,6 +3664,47 @@ class or its methods for more information.
...
+    ...
+    ...     def process_parameters(self, a=None, b=None, mu=None, nu=None):
+    ...         if a is not None and b is not None and mu is None and nu is None:
```
mdhaber, line 3684: Technically only need to check *one*, but perhaps the symmetry matters more:
```suggestion
    ...         if a is not None and b is not None:
```

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3651,6 +3664,47 @@ class or its methods for more information.
...
+    ...             nu = a + b
+    ...             mu = a / nu
+    ...         else:
+    ...             a = mu * nu
+    ...             b = nu - a
+    ...         return {"a": a, "b": b, "mu": mu, "nu": nu}
```
mdhaber, line 3690: While I'm changing things completely subjectively...
```suggestion
    ...         return dict(a=a, b=b, mu=mu, nu=nu)
```

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3795,22 +3849,27 @@ def _get_domain_info(info):
...
-    for name, info in dist.parameters.items():
-        domain_info, typical = _get_domain_info(info)
-        domain = _RealDomain(**domain_info)
-        param = _RealParameter(name, domain=domain, typical=typical)
-        parameters.append(param)
+    dist_parameters = (
+        dist.parameters if isinstance(dist.parameters, tuple) else (dist.parameters, )
+    )
+    parameterizations = []
+    for parameterization in dist_parameters:
```
mdhaber, line 3856: Interesting... I was going to comment about a confusing variable name, but I see why it's named this way. The attribute `parameters` of the custom distribution class really takes the place of `parameterization` in the infrastructure itself. 

Do you think the attribute should still be named `parameters` or should it be `parameterization`? Probably fine to keep it `parameters`, but address it as a comment in the code?

scipy/stats/tests/test_continuous.py
```diff
@@ -1272,6 +1273,70 @@ def cdf(self, x, *, c, mu, sigma):
         assert_allclose(X1.icdf(p), X2.icdf(p))
         assert_allclose(X1.iccdf(p), X2.iccdf(p))
 
+    @pytest.mark.parametrize("a", [0.5, 1.0, 2.0, 4.0, 8.0])
+    @pytest.mark.parametrize("b", [0.5, 1.0, 2.0, 4.0, 8.0])
```
mdhaber, line 1277: To save some test time and strengthen the test, maybe:
```suggestion
    @pytest.mark.parametrize("a", [0.5, np.asarray([0.5, 1.0, 2.0, 4.0, 8.0])])
    @pytest.mark.parametrize("b", [0.5, np.asarray([0.5, 1.0, 2.0, 4.0, 8.0])])
```
That way we test with arrays and not just scalars, but we only run the test four times instead of 25.

mdhaber (approved):
Looks great. I'm going to commit some of these doc changes and run a `[docs only]` build to see if that fixes the rendering issues, e.g. 

image

Then there are two minor things to consider, but either way, I can merge this tomorrow.

scipy/stats/_distribution_infrastructure.py
steppi: 100% agreed. I felt a little confused writing this even.

scipy/stats/_distribution_infrastructure.py
steppi: Yeah, I struggled a little bit naming things here. I agree with keeping it as `parameters` and addressing it with a code comment. `parameters`  seems reasonable from a user-perspective.

mdhaber:
Great. Please check how the docs look, then I'll let you push any last changes. Thanks for this!

mdhaber:
I think this pretty conclusively closes gh-4538 and gh-12133, if they hadn't been closed before. Thanks @steppi!

PR #22755, opened 2025-03-28
MAINT: special: refine `logsumexp` writeback behaviour

crusaderky:
- Revisit how `logsumexp` avoids writing back to its arguments.
- Remove unnecessary deep copy of `b`
- On JAX and (in the future) PyData Sparse, remove an unnecessary copy of `a`.
- Add test that the arguments are effectively preserved.
- Minor style cleanup.

### DIFF
--- a/scipy/_lib/_array_api.py
+++ b/scipy/_lib/_array_api.py
@@ -514,7 +514,7 @@ def xp_ravel(x: Array, /, *, xp: ModuleType | None = None) -> Array:
 
 
 # utility to broadcast arrays and promote to common dtype
-def xp_broadcast_promote(*args, ensure_writeable=False, force_floating=False, xp=None):
+def xp_broadcast_promote(*args, force_floating=False, xp=None):
     xp = array_namespace(*args) if xp is None else xp
 
     args = [(_asarray(arg, subok=True) if arg is not None else arg) for arg in args]
@@ -561,9 +561,10 @@ def xp_broadcast_promote(*args, ensure_writeable=False, force_floating=False, xp
             kwargs = {'subok': True} if is_numpy(xp) else {}
             arg = xp.broadcast_to(arg, shape, **kwargs)
 
-        # convert dtype/copy only if needed
-        if (arg.dtype != dtype) or ensure_writeable:
-            arg = xp.astype(arg, dtype, copy=True)
+        # This is much faster than xp.astype(arg, dtype, copy=False)
+        if arg.dtype != dtype:
+            arg = xp.astype(arg, dtype)
+
         out.append(arg)
 
     return out
--- a/scipy/special/_logsumexp.py
+++ b/scipy/special/_logsumexp.py
@@ -104,7 +104,7 @@ def logsumexp(a, axis=None, b=None, keepdims=False, return_sign=False):
 
     """
     xp = array_namespace(a, b)
-    a, b = xp_broadcast_promote(a, b, ensure_writeable=True, force_floating=True, xp=xp)
+    a, b = xp_broadcast_promote(a, b, force_floating=True, xp=xp)
     a = xpx.atleast_nd(a, ndim=1, xp=xp)
     b = xpx.atleast_nd(b, ndim=1, xp=xp) if b is not None else b
     axis = tuple(range(a.ndim)) if axis is None else axis
@@ -115,10 +115,10 @@ def logsumexp(a, axis=None, b=None, keepdims=False, return_sign=False):
             # delegate edge case handling to the behavior of `xp.log` and `xp.exp`,
             # which should follow the C99 standard for complex values.
             b_exp_a = xp.exp(a) if b is None else b * xp.exp(a)
-            sum = xp.sum(b_exp_a, axis=axis, keepdims=True)
-            sgn_inf = _sign(sum, xp) if return_sign else None
-            sum = xp.abs(sum) if return_sign else sum
-            out_inf = xp.log(sum)
+            sum_ = xp.sum(b_exp_a, axis=axis, keepdims=True)
+            sgn_inf = _sign(sum_, xp=xp) if return_sign else None
+            sum_ = xp.abs(sum_) if return_sign else sum_
+            out_inf = xp.log(sum_)
 
         with np.errstate(divide='ignore', invalid='ignore'):  # log of zero is OK
             out, sgn = _logsumexp(a, b, axis=axis, return_sign=return_sign, xp=xp)
@@ -138,11 +138,11 @@ def logsumexp(a, axis=None, b=None, keepdims=False, return_sign=False):
     if xp.isdtype(out.dtype, 'complex floating'):
         if return_sign:
             real = xp.real(sgn)
-            imag = xp_float_to_complex(_wrap_radians(xp.imag(sgn), xp))
+            imag = xp_float_to_complex(_wrap_radians(xp.imag(sgn), xp=xp), xp=xp)
             sgn = real + imag*1j
         else:
             real = xp.real(out)
-            imag = xp_float_to_complex(_wrap_radians(xp.imag(out), xp))
+            imag = xp_float_to_complex(_wrap_radians(xp.imag(out), xp=xp), xp=xp)
             out = real + imag*1j
 
     # Deal with shape details - reducing dimensions and convert 0-D to scalar for NumPy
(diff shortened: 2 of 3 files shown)


scipy/special/_logsumexp.py
```diff
@@ -104,7 +104,7 @@ def logsumexp(a, axis=None, b=None, keepdims=False, return_sign=False):
 
     """
     xp = array_namespace(a, b)
-    a, b = xp_broadcast_promote(a, b, ensure_writeable=True, force_floating=True, xp=xp)
+    a, b = xp_broadcast_promote(a, b, copy=True, force_floating=True, xp=xp)
```
crusaderky, line 107: This parameter was a misnomber from the time when there was only numpy and a copy of a read-only array was always writeable.
The key intent of this flag is to ensure we don't write back to the input, not to ensure that the argument is writeable.
So I've renamed it to match the copy parameter of `astype`.

I'll write it in the docstring once the conflict with https://github.com/scipy/scipy/pull/22695 is resolved.

scipy/_lib/_array_api.py
```diff
@@ -514,7 +515,7 @@ def xp_ravel(x: Array, /, *, xp: ModuleType | None = None) -> Array:
 
 
 # utility to broadcast arrays and promote to common dtype
-def xp_broadcast_promote(*args, ensure_writeable=False, force_floating=False, xp=None):
+def xp_broadcast_promote(*args, copy=False, force_floating=False, xp=None):
```
crusaderky, line 518: This was misnamed from the time when args could only be numpy. a copy of a numpy read-only array is only writeable - which is not something you can say about JAX or PyData Sparse arrays.

The intent of this parameter is not to ensure that you can write to the arguments, but that you do not accidentally write back to the inputs.
I've renamed the parameter to match the same in `astype`.

I'll add documentation once conflict with https://github.com/scipy/scipy/pull/22735 is resolved.

scipy/_lib/_array_api.py
```diff
@@ -561,9 +561,10 @@ def xp_broadcast_promote(*args, ensure_writeable=False, force_floating=False, xp
             kwargs = {'subok': True} if is_numpy(xp) else {}
             arg = xp.broadcast_to(arg, shape, **kwargs)
 
-        # convert dtype/copy only if needed
-        if (arg.dtype != dtype) or ensure_writeable:
-            arg = xp.astype(arg, dtype, copy=True)
+        # This is much faster than xp.astype(arg, dtype, copy=False)
+        if arg.dtype != dtype:
+            arg = xp.astype(arg, dtype)
```
crusaderky, line 566: This was from the time when args could only be numpy. a copy of a numpy read-only array is only writeable - which is not something you can say about JAX or PyData Sparse arrays.

The parameter name was a bit misleading - the intent was not to ensure that the arguments are writeable, but that you do not accidentally write back to the inputs.
Anyway, this becomes obsolete thanks to the `copy=True` optional parameter of `xpx.at`.

Minor merge conflict with https://github.com/scipy/scipy/pull/22735.

scipy/special/_logsumexp.py
```diff
@@ -188,37 +186,36 @@ def _elements_and_indices_with_max_real(a, axis=-1, xp=None):
...
 
-def _logsumexp(a, b, axis, return_sign, xp):
-
+def _logsumexp(a, b, *, axis, return_sign, xp):
     # This has been around for about a decade, so let's consider it a feature:
     # Even if element of `a` is infinite or NaN, it adds nothing to the sum if
     # the corresponding weight is zero.
     if b is not None:
-        a = xpx.at(a, b == 0).set(-xp.inf)
+        a = xpx.at(a, b == 0).set(-xp.inf, copy=True)
```
crusaderky, line 206: I like this pattern a lot more than before.
Instead of writing back to the arguments of the function (which, generically speaking, is an antipattern unless thoroughly documented) and then making sure somewhere else that it is OK to do so because at some point upstream in the stack of private function calls the argument had been deep-copied, each private function is individually responsible for not polluting its inputs. Which is a lot easier to spot.

scipy/_lib/_array_api.py
mdhaber: It has only existed for about 6-9 months, depending on how you count.

I don't recally exactly why that name was given, but fine to remove it if it's still only used by `logsumexp`.

scipy/special/_logsumexp.py
```diff
@@ -155,16 +154,15 @@ def logsumexp(a, axis=None, b=None, keepdims=False, return_sign=False):
     return (out, sgn) if return_sign else out
 
 
-def _wrap_radians(x, xp=None):
-    xp = array_namespace(x) if xp is None else xp
```
mdhaber, line 159: At some point it was encouraged to do this. I don't know why, but I followed along. I don't see why private functions shouldn't insist on being passed `xp`, so this LGTM.

crusaderky:
CI failure is unrelated

scipy/special/_logsumexp.py
```diff
@@ -189,37 +186,36 @@ def _elements_and_indices_with_max_real(a, axis=-1, xp=None):
...
         a = xp.where(mask, a, 0.)
-        max = xp.sum(a, axis=axis, dtype=a.dtype, keepdims=True)
+        max_ = xp.sum(a, axis=axis, dtype=a.dtype, keepdims=True)
     else:
-        max = xp.max(a, axis=axis, keepdims=True)
-        mask = a == max
+        max_ = xp.max(a, axis=axis, keepdims=True)
+        mask = a == max_
 
-    return xp.asarray(max), xp.asarray(mask)
```
mdhaber, line 197: This was probably here for a reason at some point due to NumPy returning scalars, which can't be mutated via `__setitem__`. I imagine that the change to `at.set` or something else makes this unnecessary.

scipy/special/_logsumexp.py
mdhaber: For private functions that will see a lot of re-use, I would agree. These private functions would be thoroughly documented, tested independently, etc., and be treated more like public functions except that they are only for use within the library. Perhaps this would lead to duplication of effort (e.g. many copies where one would be sufficient), but it is probably safer overall.

Just note that isn't the state of many private functions in SciPy. In this case, `_logsumexp` was only used by `logsumexp`, and it was only intended to by used by `logsumexp`. The purpose of the distinction between the two functions was just to keep the private `_logsumexp` conceptually focused on the algorithmic things rather than details like this. If another function *were* to call `_logsumexp` (or any other undocumented private function), they would do so at their own risk. There are a lot of things organized this way (including many that I have nothing to do with), and they do sometimes lead to bugs, so I don't mean to get into a best practices debate - it's just a statement of the way many things are.

So I don't have a problem with changing this instance. Mostly, I'd just caution against assuming that other private functions in SciPy have these sort of safeguards.

scipy/_lib/_array_api.py
lucascolley: my recollection is that @crusaderky's assessment is accurate

lucascolley (approved):
thanks Guido, LGTM!

mdhaber (commented):
This looks like a reasonably safe refactor. 

Previously, the arguments were copied up front in the public function, The `b` copy was technically unnecessary, since only `a` is currently modified in the implementation. Also, for immutable backends, this separate copy may be unnecessary, since they are going to perform a copy at the first `at.set` anyway.

Now, `a` is copied in (only) one of two calls to `at.set` in the private function. This is argued to be better practice because if another function wants to re-use the private function, the private function protects against writeback. (If the calling function has already copied the arrays, this might duplicate their effort, but TBH, both possibilities are probably irrelevant because the private `_logsumexp` function is unlikely to be used besides by the public `logsumexp` function.)

If something changes such that `b` is modified in the future, it is no longer copied, so this might start writing back to the user's `b` argument. However, there is now a test that guards against this.

Before I merge, @crusaderky, can you just confirm why the `xp.asarray`s mentioned [here](https://github.com/scipy/scipy/pull/22755/files#r2021772733) are no longer necessary?

I might also suggest that one of the elements of `b` equal `0` in the test, since that is one of the conditions under which `a` gets modified.

mdhaber:
OK, well, that works, too.

scipy/_lib/_array_api.py
mdhaber: Well, the assertion that "This was from the time when args could only be numpy" made me pause, because this function was certainly added as part of the array api translation of `logsumexp`. 

At that time, we were only really thinking about mutable backends, though. I agree that the name of the parameter is no longer accurate for immutable backends, which cannot be writable, and one of the purposes it served in this function was to avoid writeback

But it's not even worth discussing the original motivations now - the end of the story is what is relevant - this PR removes the argument, which is fine because it's no longer needed by the only function that ever used it. If it were to be added again later, another name might be more appropriate.

scipy/special/_logsumexp.py
crusaderky: That's my understanding too.

Issue #22758, opened 2025-03-29
CI: Failures in pytest-run-parallel CI job


rgommers:
After we enabled `pytest-run-parallel` to stress test the free-threading support in SciPy in gh-22658, we've seen a number of failures of the CI job on PRs, maybe 20% of the time the job is failing - mostly with a few different test failures that repeat. These are:

- [ ] segfault inside CPython's exception handling
- [x] a `stats.weightedtau` test failure (addressed in gh-22759)
- [ ] two `optimize.differential_evolution` failures
- [ ] a `sparse.linalg.factorize` failure
- [ ] a job timeout after all tests pass
- [ ] a `_page_trend_test` failure

Details:

1. A segfault inside CPython's exception handling ([example log](https://github.com/scipy/scipy/actions/runs/14120895339/job/39560826909)):
```
scipy/interpolate/tests/test_fitpack.py ................................ [ 13%]
Fatal Python error: Segmentation fault

Current thread 0x00007f445d615b80 (most recent call first):
  File "/opt/hostedtoolcache/Python/3.13.2/x64-freethreaded/lib/python3.13t/traceback.py", line 200 in _safe_string
  File "/opt/hostedtoolcache/Python/3.13.2/x64-freethreaded/lib/python3.13t/traceback.py", line 1053 in __init__
  File "/opt/hostedtoolcache/Python/3.13.2/x64-freethreaded/lib/python3.13t/traceback.py", line 1134 in __init__
  File "/opt/hostedtoolcache/Python/3.13.2/x64-freethreaded/lib/python3.13t/traceback.py", line 176 in format_exception_only
```
I believe this is fixed in CPython main, but not in 3.13.2 (it may not be backported to 3.13.x at all, not sure yet).

2. A test for `stats.weightedtau` ([example log](https://github.com/scipy/scipy/actions/runs/14132063045/job/39594674823)):
```
________________________ test_weightedtau_vs_quadratic _________________________
scipy/stats/tests/test_stats.py:2129: in test_weightedtau_vs_quadratic
    assert_approx_equal(expected, actual)
E   AssertionError: 
E   Items are not equal to 7 significant digits:
E    ACTUAL: 5.836994531322873e-17
E    DESIRED: 5.836994531322869e-16
        _          = 1
        a          = [1, 2, 3, 3, 2, 3]
        actual     = np.float64(5.836994531322869e-16)
        add        = False
        b          = [2, 3, 3, 3, 2, 1]
[...]

PR #22759, opened 2025-03-30
TST: weightedtau rng thread safety

tylerjereddy:
* This deals with one of the items in gh-22758.

* `test_weightedtau_vs_quadratic()` has been adjusted to use a local "scoped"/modern NumPy random `Generator` rather than a global random seed, which allows the following parallel stress test incantation to pass 3 times in a row locally on x86_64 Linux after failing 3 times in a row on latest `main`:

`python dev.py test -t scipy/stats/tests/test_stats.py::test_weightedtau_vs_quadratic -- -- --parallel-threads=40 --iterations=40`

* For a bit more detail, the shuffled arrays (`a` and `b`) were not shuffled deterministically in the repeat incantations on SciPy `main`, and possibly other data structures were also not deterministic as a result. I believe I've seen swaps away from global NumPy random state in a few thread safety patches now so I believe this is not unexpected. I could also perhaps use i.e., 17 digits instead of 2 for `default_rng` but ...

### DIFF
--- a/scipy/stats/tests/test_stats.py
+++ b/scipy/stats/tests/test_stats.py
@@ -2111,15 +2111,15 @@ def test_weightedtau_vs_quadratic():
     def weigher(x):
         return 1. / (x + 1)
 
-    np.random.seed(42)
+    rng = np.random.default_rng(42)
     for s in range(3,10):
         a = []
         # Generate rankings with ties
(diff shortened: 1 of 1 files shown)


mdhaber:
Thanks, Tyler!

PR #22760, opened 2025-03-30
BUG: optimize: `VectorFunction.f_updated` wasn't being set on initialisation

### LINKED ISSUE #18487: max_nfev is ignored by least_squares method
I am setting max_nfev to 1 in the least_squares method, so that I can debug something downstream of that method call. I expect my objective function to be called only once, but the optimization process continues for more function evaluations than those required for a finite diff jacobian evaluation, certainly more than 1 evaluation.

The function call is as follows:

            result = opt.least_squares(fun, 
                        x0,
                        method='lm', 
                        ftol=1e-08, 
                        xtol=1e-08,
                        gtol=1e-08,
                        x_scale=1.0,
                        loss='linear', 
                        f_scale=1.0, 
                        diff_step=None, 
                        tr_solver=None, 
                        tr_options={}, 
                        jac_sparsity=None, 
                        max_nfev=1, 
                        verbose=0,
                        args=(), 
                        kwargs={})

Where fun calls an external binary, and reads some output text files, and writes some screen output. I call tell how many evaluations are made by monitoring this screen output.

I have tried with other values than one, but it seems to be ignored anyway. Since from other messages it seems that this option works for other people it may be something wrong with my set-up.

My scipy version is 1.9.0, loaded in an anaconda environment.

Thanks

andyfaff:
Found a bug that `optimize._differentiable_functions.VectorFunction` didn't have its `f_updated` attribute set correctly upon initialisation. This PR fixes that, and adds a check that the attribute is set correctly. 

Other tests are added to check that `least_squares` have the correct number of `nfev`, which is how I discovered the bug.

### DIFF
--- a/scipy/optimize/_differentiable_functions.py
+++ b/scipy/optimize/_differentiable_functions.py
@@ -500,7 +500,7 @@ class VectorFunction:
             self.f = fun_wrapped(xp_copy(self.x))
 
         self._update_fun_impl = update_fun
-        update_fun()
+        self._update_fun()
 
         self.v = np.zeros_like(self.f)
         self.m = self.v.size
--- a/scipy/optimize/_lsq/least_squares.py
+++ b/scipy/optimize/_lsq/least_squares.py
@@ -406,7 +406,8 @@ def least_squares(
         of crucial importance.
     max_nfev : None or int, optional
         For all methods this parameter controls the maximum number of function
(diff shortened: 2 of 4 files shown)


tylerjereddy:
This was considered for backporting but seemed too messy per note in gh-22840, so I'll remove the label.

Ping me directly if that is a concern, but I'll otherwise start finalizing for `1.15.3`.

Issue #22736, opened 2025-03-24
ENH: Updating scipy.io.loadmat to extract data from MATLAB objects/classes


foreverallama:
### Is your feature request related to a problem? Please describe.

Hi All!
Recently I was working with some .mat files on Python, using `scipy.io.loadmat` to load the variables from the files. However, scipy lacks enough functionality to extract contents from classes/objects defined in MATLAB. These may be both user-defined or datatypes like `string` and `datetime` which are implemented as class objects.

Similar issues have been opened in the past #11818 #12408 #13636 #15984 

### Describe the solution you'd like.

I believe loadmat's functionality can be easily extended to parse and extract object data from MAT files as well. I have spent some time reverse engineering the MAT file structure for contents, taking some help from [mahalex’s guide on github](https://github.com/mahalex/MatFileHandler/blob/master/MatFileHandler/Objects.md) which gives a detailed breakdown on how MAT files are structured for objects.

I am quite familiar with the structure, and I’m also still working on further decoding certain elements of this file structure, as it is not publicly documented.

I would be happy to contribute to this feature addition myself. I am confident in my understanding of the MAT-file structure and I’m also sufficiently familiar with the `scipy.matlab.io` codebase. However, I’m not too familiar with code optimization techniques, particularly in Cython, which most of the codebase is written in.

### Additional context (e.g. screenshots, GIFs)

Going through the existing code in `scipy.io.matlab`, this functionality can be added on top without changing any of the existing codebase. As `loadmat` already returns the subsystem data to `__function_workspace__`, only a couple of changes are required:

1. Decode the MatlabOpaque object and extract object metadata from it. This contains information like `object_id`, `class_id`, `array_name`, `n_dims` and `dims`. This information can then be saved to the header in `VarHeader5`.

2. Parse the subsystem data stored in `__function_workspace__` as another MAT-file. The subsystem data contains information about every object in the MAT-file and their relevant properties. Each unique object and class in a MAT-file is assigned a unique ID, and the subsystem data essentially contains instructions on how to link the object properties to its ID.

Once this is done, every object extracted from the subsystem data can be assigned to the variable extracted in Step 1. I'll draw up an image explaining the MAT-file structure for objects, and a plan visualizing how the current `loadmat` algorithm can be updated.

Tagging @matthew-brett since he's been involved in the relevant issues

foreverallama:
## More context surrounding the MAT-file format

[...]

Issue #15984, opened 2022-04-13
ENH: Reverse engineer MATLAB datetime format


max3-2:
### Is your feature request related to a problem? Please describe.

Loading `*mat` files which contain matlab datetime objects produce unreadable results for that type of data.

### Describe the solution you'd like.

Enabling a method to parse the datetime class from Matlab to a readable format

### Describe alternatives you've considered.

Converting in matlab to string or timestamp does work but requires Matlab

### Additional context (e.g. screenshots, GIFs)

Having a datetime object in matlab an saving this to a file
```matlab
date = datetime(2022, 4, 13, 12,0,5)

  datetime

   13-Apr-2022 12:00:05

save('test_date.mat')
```

and loading into python
```python
from scipy.io import loadmat

test = loadmat('test_date.mat')['None']
print(test)

[(b'date', b'MCOS', b'datetime', array([[3707764736],
        [         2],
        [         1],
        [         1],
        [         1],
        [         1]], dtype=uint32))                ]
```
yields the above. has anyone been able to decode MATLAB datetime binary representation?

Issue #13636, opened 2021-03-02
`scipy.io.loadmat` does not load tform correctly


aleksejs-fomins:
A matlab function [fitgeotrans](https://ch.mathworks.com/help/images/ref/fitgeotrans.html) produces a tform struct which is then saved to a file. I have uploaded a tiny example file [here](https://filebin.net/aan18idmc1p7gi38). The scipy function `scipy.io.loadmat` can load this file, but it does not resemble the original file at all. If you open the file with matlab, it looks like [this](https://filebin.net/aan18idmc1p7gi38/Screenshot_from_2021-03-02_09-32-55.png?t=vnqutk2m). When you open it with python, it looks like this

```
{'__header__': b'MATLAB 5.0 MAT-file, Platform: PCWIN64, Created on: Tue Jul 31 17:37:21 2018',
 '__version__': '1.0',
 '__globals__': [],
 'None': MatlabOpaque([(b'tform', b'MCOS', b'images.geotrans.PolynomialTransformation2D', array([[3707764736],
        [         2],
        [         1],
        [         1],
        [         1],
        [         1]], dtype=uint32))],
              dtype=[('s0', 'O'), ('s1', 'O'), ('s2', 'O'), ('arr', 'O')]),
 '__function_workspace__': array([[ 0,  1, 73, ...,  0,  0,  0]], dtype=uint8)}
```

Please help. Any suggestions welcome. My goal is to automatically read many such files in python. I'm only interested in this particular structure, so I'm happy to do some hacks if necessary

tupui:
Did you have a look at the tutorial? https://docs.scipy.org/doc/scipy/reference/tutorial/io.html

tupui:
Also see this: https://stackoverflow.com/questions/15512560/access-mat-file-containing-matlab-classes-in-python

Looks like this is not supported. Maybe @matthew-brett could tell.

aleksejs-fomins:
@tupui Thanks, I figured this much. Since I have many such files, changing them to structs manually is not an option. I do not know matlab enough to be able to write a loop that would convert multiple files from object to struct. I guess my only option is hex analysis...

seb5433:
> @tupui Thanks, I figured this much. Since I have many such files, changing them to structs manually is not an option. I do not know matlab enough to be able to write a loop that would convert multiple files from object to struct. I guess my only option is hex analysis...

Did you do it?

Issue #12408, opened 2020-06-22
Support the new Matlab string arrays in scipy.io


AKuederle:
**Is your feature request related to a problem? Please describe.**
Collaborating with other researches sometimes requires reading and writing matlab files. Scipy already has some awesome tools for this. However, the `scipy.io` module can not handle Matlab strings and string arrays. When reading files that contain string arrays, the output is "nonsense" and when writing arrays of Python strings to matlab, they will be converted to char arrays instead of a matlab string array.

**Describe the solution you'd like**

It would be awesome, if we could get native support for this datatype, but I honestly have no idea, where to start with this issue.

I hadn’t had time to prepare a proper example yet. Until then here is a Stack Overflow post that describes the issue as well: https://stackoverflow.com/questions/60045637/matlab-string-array-to-python-numpy

I would be open to work on this issue myself, but I do have limited experience with Matlab and how one would go about reverse engineering the file format. So I might require some hints to get started here.

mdhaber:
It looks like the format is documented [here](https://www.mathworks.com/help/pdf_doc/matlab/matfile_format.pdf). It would take closer inspection to tell whether it is complete enough to guide better support for string arrays.

matthew-brett:
Matt - can you give a page reference in that doc?  I see the documentation for character arrays,  which we do support, but I'm sure that @AKuederle is referring the newish Matlab "string" type, and I can't find a reference to that type.

PR #22762, opened 2025-03-30
ENH: io.loadmat: extract data from MATLAB objects/classes

foreverallama:
This pull request is part of an enhancement to `io.loadmat` to incorporate reading and writing of MATLAB objects/classes in MAT-files.

This has been detailed in #22736, and references other issues opened in the past #11818 #12408 #13636 #15984 

The purpose of this PR is to contribute to the development process iteratively, allowing for feedback along the way. The initial commit includes:
- Updated `MatFile5Reader.read_file_header()` to return the subsystem offset. This will be used as a flag later on to detect and process subsystem data

- Updated `VarReader5.read_header()` to return the variable name for `mxOPAQUE_CLASS`
    - Additional headers i.e., Type System Name and Class Name, are read but not included as part of the header. As per my knowledge, they are not required for processing `mxOPAQUE_CLASS` arrays
    
- Updated `VarReader5.read_opaque()` to return the object ID of the variable.
    - Object ID will be used to link field contents from the subsystem to the variable name
    - Other metadata like class ID and object reference is not required
    - TODO: Extract `ndims` and `dims` from this data element and update the var header
    
- Updated `OPAQUE_DTYPE` to represent `object_id`. This is simply to check if the variable is an object when finally linking it with the subsystem

### DIFF
--- a/scipy/io/matlab/_mio5.py
+++ b/scipy/io/matlab/_mio5.py
@@ -90,7 +90,7 @@ from ._miobase import (MatFileReader, docfiller, matdims, read_dtype,
                       MatReadError, MatReadWarning, MatWriteWarning)
 
 # Reader object for matlab 5 format variables
-from ._mio5_utils import VarReader5
+from ._mio5_utils import VarReader5, SubsystemReader5
 
 # Constants and helper objects
 from ._mio5_params import (MatlabObject, MatlabFunction, MDTYPES, NP_TO_MTYPES,
@@ -237,6 +237,28 @@ class MatFile5Reader(MatFileReader):
         # reader for matrix streams
         self._matrix_reader = VarReader5(self)
 
+    def initialize_subsystem(self, sso):
+        ''' Initialize subsystem reader '''
+        cur_pos = self.mat_stream.tell()
+        self.mat_stream.seek(sso)
+        mdtype, byte_count = self._file_reader.read_full_tag()
+        if not byte_count > 0:
+            raise ValueError("Did not read any bytes in subsystem")
+        if mdtype == miCOMPRESSED:
+            # Make new stream from compressed data
+            stream = ZlibInputStream(self.mat_stream, byte_count)
+            mdtype = miMATRIX
+            is_compressed = True
+        else:
+            stream = self.mat_stream
+            is_compressed = False
+        if not mdtype == miMATRIX:
+            raise TypeError(f'Expecting miMATRIX type here, got {mdtype}')
+
+        subsystem = SubsystemReader5(stream, is_compressed)
+        self.mat_stream.seek(cur_pos)
+        return subsystem
+
     def read_var_header(self):
         ''' Read header, return header, next position
 
@@ -272,7 +294,7 @@ class MatFile5Reader(MatFileReader):
         header = self._matrix_reader.read_header(check_stream_limit)
         return header, next_pos
 
-    def read_var_array(self, header, process=True):
+    def read_var_array(self, header, subsystem, process=True):
         ''' Read array, given `header`
 
         Parameters
@@ -289,7 +311,7 @@ class MatFile5Reader(MatFileReader):
            array with post-processing applied or not according to
            `process`.
         '''
-        return self._matrix_reader.array_from_header(header, process)
+        return self._matrix_reader.array_from_header(header, process, subsystem)
 
     def get_variables(self, variable_names=None):
         ''' get variables from stream as dictionary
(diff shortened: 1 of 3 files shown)


foreverallama:
`loadmat` now returns headers for `mxOPAQUE_CLASS` which are required to extract its data from the subsystem. 

**Next steps**
If `subsystem_offset` is set, then initialize the subsystem. This includes parsing metadata within the subsystem and create some links between object IDs, class IDs, and their field contents. This would be set only once per MAT-file, maybe as an instance of a new class `SubsystemReader`.

scipy/io/matlab/_mio5.py
```diff
@@ -237,6 +237,28 @@ def initialize_read(self):
...
+        else:
+            stream = self.mat_stream
+            is_compressed = False
+        if not mdtype == miMATRIX:
+            raise TypeError(f'Expecting miMATRIX type here, got {mdtype}')
+
+        subsystem = SubsystemReader5(stream, is_compressed)
+        self.mat_stream.seek(cur_pos)
+        return subsystem
+
```
foreverallama, line 261: This works because the subsystem is stored within a normal `mxINT8` array. Need to check how `miCOMPRESSED` tags affect the byte markers

scipy/io/matlab/_mio5_utils.pyx
```diff
@@ -997,3 +997,189 @@ cdef class VarReader5:
         # Cython (0.23.4).
         return None
         #! Placeholder, replace with parsing methods from metadata
+
+cdef class SubsystemReader5:
+    cdef int is_swapped # endian indicator derived from subsystem
+    cdef list names # list of all field names and class names
+    cdef list class_names # list of class names ordered by class_id from names
+    cdef list object_list # list of all objects with their attributes like type_id, dependency_id, and class_id
+    cdef list field_content_pos # list of byte markers for field contents
```
foreverallama, line 1006: Using lists to store these data. Possible optimizations could come from handling these datatypes

foreverallama:
The initial parsing of the subsystem is working well. Algo flowchart below for reference. 
- Time with original `loadmat` implementation: `78.3 μs ± 1.36 μs`
- Time including initializing subsystem: `87 μs ± 976 ns`

### Next steps
1. Implement `VarReader5.read_opaque()`

algo_flowchart

foreverallama:
Current implementation now reads both user-defined objects and MATLAB datatypes like `datetime`, `string`, `table`, etc. 
Few points
- There are still quite a few unknowns in the subsystem. However, these are required for some of the more advanced OOP concepts in MATLAB. This does not affect reading of MATLAB datatypes like `string` and `datetime` and even basic user-defined objects.
- Implementation also takes care of the nested case, i.e., the property of an object array being another object array itself. This is done by calling the function recursively.
- Currently field contents are returned as raw data. As per my knowledge, only Type 1 objects like `string` need a separate parsing and cannot be read as a normal array.
- Timing came to about `127 μs ± 3.24` for a MAT-file with a single `datetime` object 1x1 array. This is about ~50μs more than the stable `loadmat` implementation which does not parse the subsystem.

### Example Output:
MAT-file contains a 1x1 datetime array `dt1`.
```
mdict['dt1'] = 
array([
    (
        'datetime',
        array([
            [
                array([
                    (
                        array([
                            [1.74291009e+12 + 9.57027078e-05j]
                        ]),
                    )
                ], dtype=[('data', 'O')])
            ]
        ], dtype=object)
    )
], dtype=[('class', 'O'), ('properties', 'O')])
```

The top level is a `(1, )` structured array with two fields - `class` and `properties`. `properties` is a `[dims]` object array which contains a structured array whose field names are the property names of the class. Each field is in turn an array itself as returned by `read_mi_matrix()`. This could use some tuning.

Todo:
1. Bug: Fix algorithm for `miCOMPRESSED` case
2. Bug: Update all calls of `read_mi_matrix()` within `_mio5_utils.pyx` to check for object references. This is useful for example such as `table` where the string metadata is stored within a `struct` and needs to be treated as object reference.
3. Include reading of default property values 
4. Add tests

Note: Will probably clean up and open a new PR later, went a bit messy with my git hist

foreverallama:
I didn't realize that you couldn't move the file pointer backwards with Zlib compressed streams. Anyways, I also got some new information on how the subsystem data is stored, which makes a lot of the workflow more clear.

I also did some timing tests and memory profiling on the MATLAB load function. It appears that MATLAB loads the complete subsystem data into memory first before linking the subsystem data to the object variable. You can observe this by creating very large arrays and assigning them to properties of objects of different classes. Load the objects in various different combinations (individually, two at a time, etc.), and you'll see they take almost the exact same time, and peak memory usage is the same as well. Compare this with normal arrays. It's clear that the data element loading is skipped if the variable is not required.

So I'll be closing this PR and opening a new one later according to the new workflow.

Issue #20768, opened 2024-05-22
BUG: optimize.minimize: garbage collection in `lbfgs`


lorentzenchr:
### Describe your issue.

Iterated calls to LBFGS increase memory consumption if garbage collection is not explicitly called.

This pipped up in https://github.com/scikit-learn/scikit-learn/issues/28993.

### Reproducing Code Example

```python
import tracemalloc

from scipy.optimize import minimize
import numpy as np


def do_something(*args, **kwargs):
    return np.random.rand()


def do_train():
    minimize(
        do_something,
        np.random.rand(300),
        args=(np.random.rand(2000, 3000)),
        method="L-BFGS-B"
    )


def run(reps):
    tracemalloc.start()
    for _ in range(reps):
        do_train()

    base, peak = tracemalloc.get_traced_memory()
    tracemalloc.stop()
    print(f"| {reps:^6} | {base/ 1024**2:^9.2f} | {peak/ 1024**2:^9.2f} | {(peak-base)/ 1024**2:^9.2f} |")


if __name__ == "__main__":
    print("| # runs | base (MB) | peak (MB) | diff (MB) |")
    print("| ------ | --------- | --------- | --------- |")

    run(1)
    run(2)
    run(3)
    run(4)
    run(5)
    run(6)
    run(7)
    run(8)
    run(9)
    run(14)
    run(16)
    run(8)
```

### Error message

```shell
| # runs | base (MB) | peak (MB) | diff (MB) |
| ------ | --------- | --------- | --------- |
|   1    |   45.79   |   46.59   |   0.80    |
|   2    |   91.57   |   92.37   |   0.80    |
|   3    |  137.36   |  138.16   |   0.80    |
|   4    |  183.15   |  183.94   |   0.80    |
|   5    |  228.93   |  229.73   |   0.80    |
|   6    |  274.72   |  275.52   |   0.80    |
|   7    |   45.79   |  321.21   |  275.42   |
|   8    |  366.29   |  367.08   |   0.80    |
|   9    |  412.07   |  412.87   |   0.80    |
|   14   |  641.00   |  641.80   |   0.80    |
|   16   |  366.29   |  412.07   |   45.78   |
|   8    |  366.29   |  367.08   |   0.80    |
```

### SciPy/NumPy/Python version and system information

```shell
python: 3.10.11
scipy: 1.11.1
```

Tialo:
Also tested with scipy==1.13.0

nickodell:
[...]

PR #22763, opened 2025-03-30
BUG: sparse: Remove reference cycle to improve memory use

### LINKED ISSUE #10634: BUG: optimize: `least_squares` with `'trf'` and `'trf_sover=lsmr'` options leaks reference cycles
I use scipy's least_squares in my paleochrono model:
http://github.com/parrenin/paleochrono
I use python as distributed by anaconda.

Specifically, I use the 'trf' method with the tr_solver='lsmr' option.
For large systems (~9000x7000), it solves the problem quite rapidly.
The problem I have is that it takes more and more memory at each iteration.
For a 9000x7000 system, it can take typically 8 Go of memory at the end, which is quite big.

If I use tr_sover='exact', the memory usage does not grow up at each iteration but it is slower.

So I would like to know, is this growing memory usage a necessary feature, or is it a bug?

>>> import sys, scipy, numpy; print(scipy.__version__, numpy.__version__, sys.version_info)
1.2.1 1.16.2 sys.version_info(major=3, minor=7, micro=3, releaselevel='final', serial=0)

nickodell:
#### Reference issue
Fixes #10634

When users run `least_squares()` with `method='trf'`, or `method='dogbox`, with `tr_solver='lsmr'`, this creates a MatrixLinearOperator. It also takes the adjoint of this operator. The MatrixLinearOperator caches its adjoint to avoid re-creating it, and the adjoint caches the original matrix, so that `operator.adjoint().adjoint()` returns the original MatrixLinearOperator. Since this creates a reference cycle, this has the potential to increase memory use.

This can be fixed by changing _AdjointLinearOperator to cache the array backing the MatrixLinearOperator rather than cache the MatrixLinearOperator itself. In other words, the MatrixLinearOperator can keep the _AdjointLinearOperator object alive, but not vice-versa. The MatrixLinearOperator is re-created when required.

There are two test programs that demonstrate the practical impact of this reference cycle.

The first is a example based on the issue report's original example, which can be found here: https://github.com/nickodell/paleochrono_hacks_donotuse/ You can use `python paleochrono.py AICC2012-Hulu-VLR` to run the test script that I ran. The change I am making here reduced memory usage from a peak of 4GB to a peak of 1GB.

I made another example, which is as simple as possible to demonstrate the bug.

```
import psutil
import numpy as np


def fun(x):
    return x + 1

def jac(x):
    return np.diag(np.ones_like(x))

def run_problem():
    x0 = np.random.rand(1000)
    res = least_squares(fun, x0, jac=jac, method='trf', tr_solver='lsmr')
    #print(res)


for i in range(1000):
    run_problem()
    if i % 100 == 0:
        print(psutil.Process().memory_info().rss / 1e6, "MB")
```

In this example, memory usage is reduced from a peak of 2243 MB to 862 MB.

~~This example also shows that there is unfortunately still a reference cycle inside `scipy.optimize`, specifically VectorFunction. This PR does not fix that. See #20768 for discussion of the reference cycle in VectorFunction, and [this test](https://github.com/scipy/scipy/blob/2c16b8bfab4d98a824d4723aebd133a402b53c83/scipy/optimize/tests/test_differentiable_functions.py#L995).~~

#### Additional information

While I tried to avoid any performance regressions, converting a matrix into adjoint and back again is unfortunately 5000 ns slower with this patch.

See this benchmark: https://gist.github.com/nickodell/07244788e2433bf4b9b74f0164124712

On main:

```
Took 4792 -/+ 82 ns for <function convert_to_mlo at 0x7f96489b4360>
Took 6373 -/+ 319 ns for <function take_adjoint_twice at 0x7f9610a85580>
Took 6233 -/+ 50 ns for <function get_original_from_adjoint at 0x7f9610a85620>
```

With patch:

```
Took 4885 -/+ 108 ns for <function convert_to_mlo at 0x7f9973904360>
Took 6734 -/+ 86 ns for <function take_adjoint_twice at 0x7f993b9d9580>
Took 11670 -/+ 65 ns for <function get_original_from_adjoint at 0x7f993b9d9620>
```

### DIFF
--- a/scipy/sparse/linalg/tests/test_interface.py
+++ b/scipy/sparse/linalg/tests/test_interface.py
@@ -13,6 +13,7 @@ import scipy.sparse as sparse
 
 import scipy.sparse.linalg._interface as interface
 from scipy.sparse._sputils import matrix
+from scipy._lib._gcutils import assert_deallocated, IS_PYPY
 
 
 class TestLinearOperator:
@@ -524,3 +525,13 @@ def test_sparse_matmat_exception():
         A @ np.identity(4)
     with assert_raises(ValueError):
         np.identity(4) @ A
+
+
+@pytest.mark.skipif(IS_PYPY, reason="Test not meaningful on PyPy")
+def test_MatrixLinearOperator_refcycle():
+    # gh-10634
+    # Test that MatrixLinearOperator can be automatically garbage collected
+    A = np.eye(2)
+    with assert_deallocated(interface.MatrixLinearOperator, A) as op:
+        op.adjoint()
+        del op
--- a/scipy/sparse/linalg/_interface.py
+++ b/scipy/sparse/linalg/_interface.py
@@ -824,19 +824,19 @@ class MatrixLinearOperator(LinearOperator):
             self.__adj = _AdjointMatrixOperator(self.A)
         return self.__adj
 
+
 class _AdjointMatrixOperator(MatrixLinearOperator):
     def __init__(self, adjoint_array):
-        self.__adjoint = adjoint_array
         self.A = adjoint_array.T.conj()
         self.args = (adjoint_array,)
         self.shape = adjoint_array.shape[1], adjoint_array.shape[0]
 
     @property
     def dtype(self):
-        return self.__adjoint.dtype
+        return self.args[0].dtype
 
     def _adjoint(self):
-        return MatrixLinearOperator(self.__adjoint)
+        return MatrixLinearOperator(self.args[0])
 
 
 class IdentityOperator(LinearOperator):
--- a/scipy/sparse/linalg/_interface.py
+++ b/scipy/sparse/linalg/_interface.py
@@ -821,22 +821,22 @@ class MatrixLinearOperator(LinearOperator):
 
     def _adjoint(self):
         if self.__adj is None:
-            self.__adj = _AdjointMatrixOperator(self)
+            self.__adj = _AdjointMatrixOperator(self.A)
         return self.__adj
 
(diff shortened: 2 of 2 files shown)


andyfaff:
I'm working on the `VectorFunction` reference cycle.

scipy/sparse/linalg/tests/test_interface.py
```diff
@@ -524,3 +525,13 @@ def test_sparse_matmat_exception():
         A @ np.identity(4)
     with assert_raises(ValueError):
         np.identity(4) @ A
+
+
+@pytest.mark.skipif(IS_PYPY, reason="Test not meaningful on PyPy")
```
tylerjereddy, line 530: I wonder if it might be slightly preferable to use `HAS_REFCOUNT` from `numpy.testing`:

`@pytest.mark.skipif(not HAS_REFCOUNT, reason="Python lacks refcounts")`

perhaps slightly more general--I've seen it used a bit more recently

andyfaff:
This looks fine to me. Regarding the extra overhead, how many times would people want to do the round trip on the MLO (o.adjoint().adjoint())?
I'll leave it up to some linalg people to merge though.

scipy/sparse/linalg/tests/test_interface.py
nickodell: Possibly - I was doing it this way for consistency with the existing `assert_deallocated` tests.

j-bowhay (approved):
Thanks for getting to bottom of this @nickodell! Changes look good to me and the slight slow down for adjoint adjoint seems acceptable as you would have the original object after all!

PR #20770, opened 2024-05-23
MAINT: optimize: remove circular reference in `ScalarFunction`

andyfaff:
This PR would close #20768.

In that issue it was determined that circular references in `optimize._differentiable_functions.ScalarFunction` prevent it from being immediately garbage collected when functions that use the class go out of scope. 

> functions defined inside `ScalarFunction.__init__()` have a closure that allows them to access local variables, such as `self`. Then those functions get assigned as instance variables of `ScalarFunction`, which creates the reference cycle.

This PR removes those circular references by lifting references to `self` out of closures used for wrapping. I have checked that the issue is fixed using the example given by @nickodell:

```
| # runs | base (MB) | peak (MB) | diff (MB) | SF objs |
| ------ | --------- | --------- | --------- | ------- |
|   10   |   0.01    |   45.92   |   45.90   |    0    |
```

Performance should be unchanged, although there are a few more `if` checks.

PR #22764, opened 2025-03-31
MAINT: optimize: `VectorFunction`: remove reference cycle

andyfaff:
Removes a reference cycle akin from `VectorFunction`, akin to #20768 (`ScalarFunction`) that was fixed by #20770

### DIFF
--- a/scipy/optimize/_differentiable_functions.py
+++ b/scipy/optimize/_differentiable_functions.py
@@ -406,8 +406,112 @@ class ScalarFunction:
         return self.f, self.g
 
 
-def _VectorFunWrapper(fun, x):
-    return np.atleast_1d(fun(x))
+class _VectorFunWrapper:
+    def __init__(self, fun):
+        self.fun = fun
+        self.nfev = 0
+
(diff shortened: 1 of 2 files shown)


nickodell (approved):
Glad to see this. The new commit fixes the tests for me. Checked that the new code has coverage under the test suite.

I also made a merged version of this and my PR for the MatrixLinearOperator reference cycle, and tried the reprex from that PR:

```
import psutil
import numpy as np


def fun(x):
    return x + 1

def jac(x):
    return np.diag(np.ones_like(x))

def run_problem():
    x0 = np.random.rand(1000)
    res = least_squares(fun, x0, jac=jac, method='trf', tr_solver='lsmr')
    #print(res)


for i in range(1000):
    run_problem()
    if i % 100 == 0:
        print(psutil.Process().memory_info().rss / 1e6, "MB")
```

This PR reduces memory usage from my PR from 862 MB to 93 MB.

lucascolley:
thanks Andrew, and for the review Nick!

PR #22766, opened 2025-03-31
DOC: improve docstrings of boxcox and yeojohnson

lorentzenchr:
#### Reference issue
None

This PR improves the docstrings of the Box-Cox and Yeo-Johnson transformations.

#### Additional information
None

### DIFF
--- a/scipy/stats/_morestats.py
+++ b/scipy/stats/_morestats.py
@@ -900,10 +900,12 @@ def boxcox_llf(lmb, data):
 
     .. math::
 
-        llf = (\lambda - 1) \sum_i(\log(x_i)) -
-              N/2 \log(\sum_i (y_i - \bar{y})^2 / N),
+        llf = (\lambda - 1) \sum_i^N(\log(x_i)) -
+              \frac{N}{2} \log(\sum_i^N (y_i - \bar{y})^2 / N),
 
     where ``y`` is the Box-Cox transformed input data ``x``.
+    This corresponds to the profile log- likelihood of the original data ``x`` with
+    some constant terms dropped.
 
     Examples
     --------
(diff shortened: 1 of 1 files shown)


scipy/stats/_morestats.py
```diff
@@ -900,10 +900,12 @@ def boxcox_llf(lmb, data):
 
     .. math::
 
-        llf = (\lambda - 1) \sum_i(\log(x_i)) -
-              N/2 \log(\sum_i (y_i - \bar{y})^2 / N),
+        llf = (\lambda - 1) \sum_i^N(\log(x_i)) -
+              \frac{N}{2} \log(\sum_i^N (y_i - \bar{y})^2 / N),
 
     where ``y`` is the Box-Cox transformed input data ``x``.
+    This corresponds to the profile log- likelihood of the original data ``x`` with
```
mdhaber, line 907: I'm thinking that the intersection of readers who are interested, don't know this, and understand "profile" without looking at another reference may be small. I'd either leave it out or be a bit more explicit.

scipy/stats/_morestats.py
```diff
@@ -900,10 +900,12 @@ def boxcox_llf(lmb, data):
 
     .. math::
 
-        llf = (\lambda - 1) \sum_i(\log(x_i)) -
-              N/2 \log(\sum_i (y_i - \bar{y})^2 / N),
+        llf = (\lambda - 1) \sum_i^N(\log(x_i)) -
```
mdhaber, line 903: In the literature, is there a common symbol instead of $llf$? I'd avoid making it look like $fl^2$. $\text{LLF}$ or something?

mdhaber (commented):
I'd rather not make a project out of something probably intended to be a one-commit cleanup, but some things stood out.

mdhaber:
Before and after:
image
image

image
image

So better, but $llf$ still standard out. If the purpose is solely to improve the appearance, I think we might as well improve that.

scipy/stats/_morestats.py
steppi: I think it's useful to keep this information. I had to look up profile likelihood because it had been a while since I'd seen the term. I think it's helpful to know what term to look up.

mdhaber: Based on https://github.com/scipy/scipy/pull/22766#discussion_r2023059863, I see now that the intent was to disambiguate this from the full log-likelihood (including scale and location). That makes it seem more useful to me to those who are interested.

mdhaber (approved):
With [that](https://github.com/scipy/scipy/pull/22766#issuecomment-2772937027) change.or any of the suggestions in https://github.com/scipy/scipy/pull/22766#discussion_r2022205256, I think we can merge this. Please include `[docs only]` in the commit message.

scipy/stats/_morestats.py
mdhaber: Anything with precedent other than $l^2f$ is fine with me.

lorentzenchr: I looked up several books. :math:`l` seems quite common for the log likelihood (e.g. Yeo & Johnson).

lucascolley:
thanks @lorentzenchr that looks good

image

As a separate issue, wide math formulas are truncated on mobile (on safari at least) 🫠

PR #22770, opened 2025-04-01
TST: stats: add marray tests for _length_nonmasked directly

zitongzhoueric:
This PR adds two tests that add coverage for `_length_nonmasked` in `scipy/stats/_stats_py.py` directly, handling the case when the input is a masked array `marray`. This was introduced in https://github.com/scipy/scipy/pull/22393

### DIFF
--- a/scipy/stats/tests/test_marray.py
+++ b/scipy/stats/tests/test_marray.py
@@ -275,22 +275,6 @@ def test_ttest_ind_from_stats(xp):
     assert res.statistic.shape == shape
     assert res.pvalue.shape == shape
 
-@pytest.mark.parametrize("keepdims", [False, True])
-def test_length_nonmasked_marray_axis0(keepdims):
-    xp = marray._get_namespace(np)
-
-    data = [[1.0, 2.0, 3.0],
-            [4.0, 5.0, 6.0]]
-    mask = [[False, True, False],
-            [False, False, True]]
-
-    marr = xp.asarray(data, mask=mask)
-
-    result = _length_nonmasked(marr, axis=0, keepdims=keepdims, xp=xp)
-    expected = np.array([[2, 1, 1]]) if keepdims else np.array([2, 1, 1])
-    np.testing.assert_array_equal(result.data, expected)
-
-
 def test_length_nonmasked_marray_iterable_axis_raises():
     xp = marray._get_namespace(np)
 
--- a/scipy/stats/tests/test_marray.py
+++ b/scipy/stats/tests/test_marray.py
@@ -282,7 +282,8 @@ def test_length_nonmasked_marray_iterable_axis_raises():
     mask = [[False, False], [True, False]]
     marr = xp.asarray(data, mask=mask)
 
-    # This should raise because axis is a tuple, and that's not supported for MArray
+    # Axis tuples are not currently supported for MArray input.
+    # This test can be removed after support is added.
     with pytest.raises(NotImplementedError,
         match="`axis` must be an integer or None for use with `MArray`"):
         _length_nonmasked(marr, axis=(0, 1), xp=xp)
--- a/scipy/stats/tests/test_marray.py
+++ b/scipy/stats/tests/test_marray.py
@@ -3,7 +3,7 @@ import numpy as np
 from scipy import stats
 
 from scipy._lib._array_api import xp_assert_close, xp_assert_equal
-from scipy.stats._stats_py import _xp_mean, _xp_var
+from scipy.stats._stats_py import _xp_mean, _xp_var, _length_nonmasked
 from scipy.stats._axis_nan_policy import _axis_nan_policy_factory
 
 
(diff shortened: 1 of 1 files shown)


scipy/stats/tests/test_marray.py
```diff
@@ -274,3 +274,31 @@ def test_ttest_ind_from_stats(xp):
...
+    xp = marray._get_namespace(np)
+
+    data = [[1.0, 2.0], [3.0, 4.0]]
+    mask = [[False, False], [True, False]]
+    marr = xp.asarray(data, mask=mask)
+
+    # This should raise because axis is a tuple, and that's not supported for MArray
+    with pytest.raises(NotImplementedError,
+        match="`axis` must be an integer or None for use with `MArray`"):
+        _length_nonmasked(marr, axis=(0, 1), xp=xp)
```
mdhaber, line 304: This is a temporary restriction. Let's make that clear with a comment that this is better than undefined behavior, but that the plan is to add support. (Not right now, though.)

mdhaber (commented):
Thanks for your interest! How did you come across this topic (use of `MArray`, adding tests for a private function), by the way? 

I'm not sure that we need these as written, since the tests for correct results of the functions probably put `_length_nonmasked` through a much wider variety of situations. I would not be opposed to adding tests if they were property-based like the other tests in the file, though. How about generating the arrays with the `get_arrays` helper and computing the result programmatically, without looking at the implementation of `_length_nonmasked`, with some independent, easy-to-verify code?

zitongzhoueric:
> I'm not sure that we need these as written, since the tests for correct results of the functions probably put `_length_nonmasked` through a much wider variety of situations. I would not be opposed to adding tests if they were property-based like the other tests in the file, though. How about generating the arrays with the `get_arrays` helper and computing the result independently, without looking at the implementation of `_length_nonmasked`, with some other simple, easy to verify method?

I debated adding tests for `_length_nonmasked` directly and testing it through some properties in the file. Sure if you prefer it the other way let me rewrite the tests.

> How did you come across this topic (use of MArray, adding tests for a private function), by the way?

We have a script running that looks for Scipy PRs with missed patch coverage.

mdhaber:
I imagine it found that there was no test for this error message? 
https://github.com/scipy/scipy/blob/efdffabfca26d1924577f8f4aa0c1bd018a41f1a/scipy/stats/_stats_py.py#L1247

I would be surprised if there were not coverage for the other lines. It's quite short..

mdhaber:
Don't worry about the CI failure, BTW. That seems to be in main.

zitongzhoueric:
> I imagine it found that there was no test for the error message?

Yes, my script reported that these three lines are not covered by the regression test suite. 
https://github.com/scipy/scipy/blob/efdffabfca26d1924577f8f4aa0c1bd018a41f1a/scipy/stats/_stats_py.py#L1247-L1250

I should have Array API enabled ```SCIPY_ARRAY_API=1```. I'm not entirely sure, could it be that the other tests in `test_marray.py` only uses `_xp_mean, _xp_var` and therefore do not test `_length_nonmasked` enough and fail to pass this guard?

https://github.com/scipy/scipy/blob/efdffabfca26d1924577f8f4aa0c1bd018a41f1a/scipy/stats/_stats_py.py#L1244-L1246

> it would make an assertion about a property that should be true of the output regardless of (randomly generated) input.

Understood. Maybe the best way is to keep the test for the exception handling (for now) and rewrite the other into a property-based test?

zitongzhoueric:
I removed the bad test. 
BTW it looks like my automated script forgot to install `marray` (because it's wasn't added as a dependency until https://github.com/scipy/scipy/pull/22393)? Consequently it skipped every test in `stats/tests/test_marray.py` . I verified that `_length_nonmasked` is indeed covered by the other tests, so only the test for exception handling is needed.

scipy/stats/tests/test_marray.py
```diff
@@ -274,3 +274,15 @@ def test_ttest_ind_from_stats(xp):
...
     assert res.pvalue.shape == shape
+
+def test_length_nonmasked_marray_iterable_axis_raises():
+    xp = marray._get_namespace(np)
+
+    data = [[1.0, 2.0], [3.0, 4.0]]
+    mask = [[False, False], [True, False]]
+    marr = xp.asarray(data, mask=mask)
+
+    # This should raise because axis is a tuple, and that's not supported for MArray
```
mdhaber, line 285:
```suggestion
    # Axis tuples are not currently supported with MArray input.
    # This test can be removed after support is added.
```

mdhaber:
> because it's wasn't added as a dependency 

Right. This is an feature that is currently under development, so although it is tested in CI, it is not listed as a dependency in `pyproject.toml`. 

This test will probably be removed by the time `marray` is documented as an experimental feature - still behind the `SCIPY_ARRAY_API` environment variable. But the test it is written and quite fast to run, so I'll go ahead and merge. Thanks!

PR #22771, opened 2025-04-01
TST: stats: don't encapsulate `pytest.warns`

crusaderky:
Follow-up to #22735 
- Fix regression in free-threaded tests caused by encapsulation of `pytest.warns` (https://github.com/scipy/scipy/pull/22735#issuecomment-2768833079)
- Address issue where only one call has to warn for the pytest.warns context to be satisfied (https://github.com/scipy/scipy/pull/22735#discussion_r2014601340)

CC @rgommers @mdhaber @lucascolley

### DIFF
--- a/scipy/stats/tests/test_stats.py
+++ b/scipy/stats/tests/test_stats.py
@@ -84,13 +84,6 @@ lazy_xp_function(stats.tmin, static_argnames=("inclusive", "axis"))
 lazy_xp_function(stats.tmax, static_argnames=("inclusive", "axis"))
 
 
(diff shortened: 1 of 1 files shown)


lucascolley (approved):
LGTM if CI is happy, thanks!

rgommers (approved):
Thanks @crusaderky. This is also a pragmatic fix. Let's remember to revert this change when we have a better way to handle this kind of utility function - I've opened https://github.com/Quansight-Labs/pytest-run-parallel/issues/35 for that.

PR #22772, opened 2025-04-01
BUG: sparse.linalg: Transposed `LinearOperator` multiplication fallback

### LINKED ISSUE #22637: BUG: Transposed LinearOperator fails on vector multiplication
### Describe your issue.

`LinearOperator`s are supposed to automatically implement their transpose operator. This usually works, but there is a bug in the logic that can lead the transpose operator to fail with `NotImplementedError` when the following two conditions hold:

1. The original matrix was defined using `_matmat` and `_rmatmat`, instead of `_matvec` and `_rmatvec`
2. You multiply the transpose with a vector, instead of a matrix

The reason is [this line](https://github.com/scipy/scipy/blob/ef4c939f7d6a1801d61c33c56af20b818e8c0bee/scipy/sparse/linalg/_interface.py#L667) in the implementation of `_TransposedLinearOperator` (same for `_AdjointLinearOperator`):
```
    def _matvec(self, x):
        # NB. np.conj works also on sparse matrices
        return np.conj(self.A._rmatvec(np.conj(x)))
```
In our case, `A._rmatvec` was never overridden, so it calls [the default](https://github.com/scipy/scipy/blob/ef4c939f7d6a1801d61c33c56af20b818e8c0bee/scipy/sparse/linalg/_interface.py#L321) `_rmatvec`:
```
    def _rmatvec(self, x):
        """Default implementation of _rmatvec; defers to adjoint."""
        if type(self)._adjoint == LinearOperator._adjoint:
            # _adjoint not overridden, prevent infinite recursion
            raise NotImplementedError
        else:
            return self.H.matvec(x)
```
But `A._adjoint` was not overridden, so it fails. I believe a fix would be to add another case in the conditional of the `LinearOperator._rmatvec` that falls back to `rmatmat` when `self._adjoint` is not defined, same as in `LinearOperator._matvec` [here](https://github.com/scipy/scipy/blob/ef4c939f7d6a1801d61c33c56af20b818e8c0bee/scipy/sparse/linalg/_interface.py#L225).

Note: in my use case, I can't really implement `adjoint` for my operator directly because it is wrapping `scipy.sparse.linalg.factorized`, which is some opaque type.

### Reproducing Code Example

```python
import numpy as np
from scipy.sparse.linalg import aslinearoperator as alo, LinearOperator

class MyMatrix(LinearOperator):
    def __init__(self, A):
        super().__init__(A.dtype, A.shape)
        self.A = A
    def _matmat(self, other): return self.A @ other
    def _rmatmat(self, other): return self.A.T @ other

    # If the following line is included, everything works fine,
    # but _rmatvec should be implemented automatically using _rmatmat.
    # def _rmatvec(self, other): return self.M.T @ other


op = MyMatrix(np.array([[1, 2], [3, 4]]))
x = np.array([1, 2])
B = np.array([[10, 20], [30, 40]])
print(op @ B)  # Succeeds
print(B.T @ op)  # Succeeds
print(op.T @ B)  # Succeeds
print(op @ x)  # Succeeds
print(x.T @ op)  # Raises NotImplementedError
print(op.T @ x)  # Raises NotImplementedError

x2 = x.reshape(-1, 1)
print(op @ x2)  # Succeeds
print(x2.T @ op)  # Raises NotImplementedError
print(op.T @ x2)  # Raises NotImplementedError
```

### Error message

```shell
  File "/Users/noah/Documents/PhD3/Musco/HSS-approximation/test.py", line 23, in <module>
    print(x.T @ op)  # Raises NotImplementedError
          ~~~~^~~~
  File "/Users/noah/miniconda3/envs/py3_12/lib/python3.12/site-packages/scipy/sparse/linalg/_interface.py", line 461, in __rmatmul__
    return self.__rmul__(other)
           ^^^^^^^^^^^^^^^^^^^^
  File "/Users/noah/miniconda3/envs/py3_12/lib/python3.12/site-packages/scipy/sparse/linalg/_interface.py", line 467, in __rmul__
    return self._rdot(x)
           ^^^^^^^^^^^^^
  File "/Users/noah/miniconda3/envs/py3_12/lib/python3.12/site-packages/scipy/sparse/linalg/_interface.py", line 499, in _rdot
    return self.T.matvec(x.T).T
           ^^^^^^^^^^^^^^^^^^
  File "/Users/noah/miniconda3/envs/py3_12/lib/python3.12/site-packages/scipy/sparse/linalg/_interface.py", line 236, in matvec
    y = self._matvec(x)
        ^^^^^^^^^^^^^^^
  File "/Users/noah/miniconda3/envs/py3_12/lib/python3.12/site-packages/scipy/sparse/linalg/_interface.py", line 648, in _matvec
    return np.conj(self.A._rmatvec(np.conj(x)))
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/noah/miniconda3/envs/py3_12/lib/python3.12/site-packages/scipy/sparse/linalg/_interface.py", line 303, in _rmatvec
    raise NotImplementedError
NotImplementedError
```

```shell
1.14.1 2.1.2 sys.version_info(major=3, minor=12, micro=0, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    found: true
    include directory: unknown
    lib directory: unknown
    name: accelerate
    pc file directory: unknown
    version: unknown
  lapack:
    found: true
    include directory: unknown
    lib directory: unknown
    name: accelerate
    pc file directory: unknown
    version: unknown
  pybind11:
    detection method: config-tool
    include directory: unknown
    name: pybind11
    version: 2.12.0
Compilers:
  c:
    commands: cc
    linker: ld64
    name: clang
    version: 15.0.0
  c++:
    commands: c++
    linker: ld64
    name: clang
    version: 15.0.0
  cython:
    commands: cython
    linker: cython
    name: cython
    version: 3.0.11
  fortran:
    commands: gfortran
    linker: ld64
    name: gcc
    version: 13.3.0
  pythran:
    include directory: ../../../../../../private/var/folders/zq/1mkv5ktx0_5cz1rlynrmdmlh0000gn/T/pip-build-env-pw6uu9w1/overlay/lib/python3.12/site-packages/pythran
    version: 0.16.1
Machine Information:
  build:
    cpu: x86_64
    endian: little
    family: x86_64
    system: darwin
  cross-compiled: false
  host:
    cpu: x86_64
    endian: little
    family: x86_64
    system: darwin
Python Information:
  path: /private/var/folders/zq/1mkv5ktx0_5cz1rlynrmdmlh0000gn/T/cibw-run-mpfy6fi_/cp312-macosx_x86_64/build/venv/bin/python
  version: '3.12'
```

DVarela56:
Added a case in `LinearOperator._rmatvec` to use
`_rmatmat` as a fallback when self._adjoint is not defined.
Included a test that ensures that LinearOperator handles transposed multiplications when only _rmatmat is defined.

#### Reference issue
Closes gh-22637

Transposed LinearOperator fails with NotImplementedError when the original matrix is defined using _matmat and _rmatmat and you multiply it with a vector, instead of a matrix.

### DIFF
--- a/scipy/sparse/linalg/_interface.py
+++ b/scipy/sparse/linalg/_interface.py
@@ -322,6 +322,10 @@ class LinearOperator:
         """Default implementation of _rmatvec; defers to adjoint."""
         if type(self)._adjoint == LinearOperator._adjoint:
             # _adjoint not overridden, prevent infinite recursion
+            if (hasattr(self, "_rmatmat")
+                    and type(self)._rmatmat != LinearOperator._rmatmat):
+                # Try to use _rmatmat as a fallback
+                return self._rmatmat(x.reshape(-1, 1)).reshape(-1)
             raise NotImplementedError
         else:
             return self.H.matvec(x)
--- a/scipy/sparse/linalg/tests/test_interface.py
+++ b/scipy/sparse/linalg/tests/test_interface.py
@@ -513,6 +513,30 @@ def test_transpose_noconjugate():
     assert_equal(B.dot(v), Y.dot(v))
     assert_equal(B.T.dot(v), Y.T.dot(v))
 
+def test_transpose_multiplication():
+    class MyMatrix(interface.LinearOperator):
+        def __init__(self, A):
+            super().__init__(A.dtype, A.shape)
+            self.A = A
+        def _matmat(self, other): return self.A @ other
+        def _rmatmat(self, other): return self.A.T @ other
+
+    A = MyMatrix(np.array([[1, 2], [3, 4]]))
+    X = np.array([1, 2])
+    B = np.array([[10, 20], [30, 40]])
+    X2 = X.reshape(-1, 1)
+    Y = np.array([[1, 2], [3, 4]])
+
+    assert_equal(A @ B, Y @ B)
+    assert_equal(B.T @ A, B.T @ Y)
+    assert_equal(A.T @ B, Y.T @ B)
+    assert_equal(A @ X, Y @ X)
+    assert_equal(X.T @ A, X.T @ Y)
+    assert_equal(A.T @ X, Y.T @ X)
+    assert_equal(A @ X2, Y @ X2)
+    assert_equal(X2.T @ A, X2.T @ Y)
+    assert_equal(A.T @ X2, Y.T @ X2)
+
 def test_sparse_matmat_exception():
     A = interface.LinearOperator((2, 2), matvec=lambda x: x)
     B = sparse.eye_array(2)


lucascolley (approved):
thanks @DVarela56, looks good to me!

@noahamsel would you have time to take a look?

NoahAmsel:
Looks good to me. Thanks @DVarela56

PR #22777, opened 2025-04-01
ENH: spatial.transform: add array API standard support

### LINKED ISSUE #22500: ENH: spatial.transform: Add array API standard support

Working towards https://github.com/scipy/scipy/issues/18867. 

There are a few decisions that are better discussed before moving forward with the implementation. It would be great if you could add your thoughts on that @lucascolley @rgommers .

There is existing cython code for the rotation implementation that offers significant performance benefits over a generic array API when using numpy arrays. Sacrificing that speed for compatibility with other frameworks than numpy is no real option. At the same time, I think we want to avoid having two `Rotation` classes / types. To work around this, I see two viable design options:

1. Inheriting from Cython
- Implement a new `Rotation` class that inherits from the existing Cython implementation. This is the one that is exposed to users.
- Operations are delegated back to the Cython superclass when called with a numpy array.
- The existing Cython code remains unchanged.

2. Rewriting with a native Python class (flexible backend support):
- Rewrite `Rotation` (and the other classes) as a native Python class with a generic transform backend.
- Implement array API functions as a fallback for all operations, and refactor the Cython code into a library of functions matching the signature.
- Register the Cython backend for NumPy arrays to enjoy (hopefully) the same speed as the current version.
- Benefits: eliminates duplicate class implementations, simplifies the code, uses a pattern that may work for the whole spatial submodule, allows other frameworks to register their own, specialized backends.
- Drawback: requires rewriting the Cython code and potentially greater refactoring effort.

I am heavily inclined towards the second solution. It's a clean way of separating the Cython code from the array API, it should work for the whole module, and it feels less hacky than the first.

Nevertheless, I have experimented with mixing the Cython `Rotation` class and a Rotation class using the array API to make sure the approach is viable. Currently, applying rotations works for torch, numpy, jax (jit and no jit) and the strict array API. So if desired, we could go down that route.

See here for implementation details
<details>

```python
# ...
from functools import wraps
from ._rotation import Rotation as CythonRotation


def _maybe_cython(f):
    cython_fn = getattr(CythonRotation, f.__name__)

    @wraps(f)
    def wrapper(self, *args, **kwargs):
        if self._use_cython:
            return cython_fn(self, *args, **kwargs)
        return f(self, *args, **kwargs)

    return wrapper


class Rotation(CythonRotation):
    def __init__(self, quat, normalize=True, copy=True, scalar_first=False):
        xp = array_namespace(quat)
        # Fast path for numpy arrays: If quat is a numpy object, we call the Cython version of the
        # class and forward all calls to the function's methods to super()
        self._use_cython = is_numpy(xp)
        self._xp = xp
        if self._use_cython:
            super().__init__(
                quat, normalize=normalize, copy=copy, scalar_first=scalar_first
            )
            return
        # TODO: Do we want this dtype check?
        dtype = (
            xp.float64 if not is_jax(xp) or jax.config.jax_enable_x64 else xp.float32
        )
        quat = xp.asarray(quat, dtype=dtype)
        if quat.shape[-1] != 4 or quat.shape[0] == 0:
            raise ValueError(
                f"Expected `quat` to have shape (..., 4), got {quat.shape}."
            )
        scalar_first = xp.asarray(scalar_first)
        normalize = xp.asarray(normalize)
        copy = xp.asarray(copy)
        # All operations are done non-branching to enable JIT-compilation
        quat = xp.where(scalar_first, xp.roll(quat, -1, axis=-1), quat)
        quat = xp.where(normalize | copy, xp.asarray(quat, copy=True), quat)
        quat_norm = xp.linalg.vector_norm(quat, axis=-1, keepdims=True)

        # TODO: This is a deviation from the behavior of the cython version because JIT code needs
        # to be strictly non-branching. If we have zero quaternions, we return nan values
        # if xp.any(quat_norm == 0):
        #     raise ValueError("Found zero norm quaternions in `quat`.")
        quat_norm = xp.where(quat_norm == 0, xp.asarray([xp.nan]), quat_norm)
        self._quat = xp.where(normalize, quat / quat_norm, quat)

    @classmethod
    def from_quat(cls, quat: Array, *, scalar_first=False) -> Rotation:
        return cls(quat, normalize=True, scalar_first=scalar_first)

    @_maybe_cython
    def as_quat(self, canonical=False, *, scalar_first=False):
        return _as_quat(self._quat, canonical=canonical, scalar_first=scalar_first)

    @_maybe_cython
    def as_matrix(self) -> Array:
        return _as_matrix(self._quat)

    @_maybe_cython
    def apply(self, points: Array) -> Array:
        return _apply(self._quat, points)
```
</details>

I'm not sure how other submodules of scipy have dealt with the problem. If there is already an established practice that differs from the proposed ones, I'm happy to take a look at them.

## Additional modifications
There are a two other things that warrant discussion. 

NaN / Producing non-branching code is important for frameworks that offer JIT compilation. I suggest we take that into account for the array API implementation. The drawback is that this prevents value-based checks/errors in the code. One example are zero-norm quaternions, which currently result in a `ValueError` when passed to a `Rotation`. The implementation above returns NaN values instead. I think both behaviors are reasonable, and producing NaNs enables jit-compiling `Rotation`, but it also is a change in behavior.

The second issue concerns batching. At the moment, `Rotation` is only implemented for a single rotation or a list of rotations. I think it would make a lot of sense to extend this such that rotations become broadcastable just like any other array. E.g., one should be able to apply a tensor of `Rotation`s of dimensions (N, 1, M) to a tensor of 3D points (1, K, M, 3). This eliminates special `_single` cases and extends the functionality in a principled way that is clearly outlined by the broadcasting rules of arrays. In fact, the generic array API implementations I wrote so far support this out of the box.

amacati:
## General overview
This PR adds array API support for the `scipy.spatial.transform` module as a step towards #18867. It allows users to create objects from the module with arrays from frameworks like `torch`, `jax` etc., and leverages these frameworks for performing the underlying operations. 

Since work started when `Rotation` was the primary part of the module, `RigidTransform` has not yet been ported. However, it would be great to discuss some of the general design decisions before moving forward with the implementation of `RigidTransform`.

### Motivation
Previously, `Rotation` was a cython class centered around the rotation data represented as a quaternion. With the addition of support for the array API we can no longer use this cython code path as the universal implementation, as some arrays from other frameworks cannot be passed into cython (e.g. tensors on a GPU, jax tracers etc.). 

Since the cython implementation can significantly speed up calculations with numpy arrays, and numpy remains the primary use case for the foreseeable future, we do not want to add too much overhead for the array API support.

### Leveraging backends
This PR proposes an architecture that separates the implementation into:

1. A pure Python `Rotation` class that handles the interface and data storage
2. Backend modules that implement the actual computations.

These backends are functional implementations of `Rotation`'s methods. Each time a `Rotation` is created from arrays, we check if a backend specifically for this array type has been added. If so, we use the specialized backend for all operations. Otherwise, we fall back to the new array API backend that is compatible with _any_ array API compatible array type.

This design allows us to nicely integrate the previous, optimized cython implementation by rewriting it into a backend. We register this cython backend for numpy arrays by default, which means that numpy arrays still enjoy the speedup from compiled cython code. In addition, this also gives us a straightforward path to include optimized implementations for individual frameworks if desired, and could even be exposed to other modules that may want to register their own, specialized implementations. Here is an example of how this works in the new implementation:

```python
# We use cython as the backend for numpy
backend_registry = {array_namespace(np.empty(0)): cython_backend}

# ... and use specialized backends if they are available in Rotation
class Rotation:
    ...
    @classmethod
    def from_matrix(cls, matrix: ArrayLike) -> Rotation:
        backend = backend_registry.get(array_namespace(matrix), array_api_backend)
        quat = backend.from_matrix(matrix)
        return cls(quat, normalize=False, copy=False)

```

### Compiling functions
Some frameworks like `jax` heavily rely on jit compiled code with statically known shapes and value-independent control flow. The generic array API backend is written (and tested) to comply with these restrictions. However, this has two implications:

1. Error handling: Since value-based checks are no longer possible, we cannot emit certain errors. Instead, we return NaN values. However, some frameworks will raise Errors (e.g. on ill-conditioned matrices for SVDs), while others don't. This leads to an inconsistent error behaviour that we should address.
2. Performance: Value-based control flow is no longer possible under these conditions, meaning that both code paths often have to be run in parallel. For eager execution frameworks, this leads to unnecessary performance degradation. We could add code paths for lazy and eager execution frameworks, but that would increase the complexity of the array API backend.

### `spatial.transform.RigidTransform` support
Just like the previous `Rotation` implementation, `RigidTransform` is currently a cython class that relies on `Rotation` being one as well. The solution is to also convert RigidTransform into a Python class as interface, and use the same backend mechanic as with `Rotation`. Since the current PR already makes significant code changes with many design decisions, it makes sense to first discuss and review these before moving on to converting `RigidTransform` as well. The final draft may then also include full `RigidTransform` support.

For now, this version of `spatial.transform` **INTENTIONALLY DISABLES** `RigidTransform` to let the build succeed. Hence, all tests for `RigidTransform` are **expected to fail**.

### array API version
The current implementation uses the latest array API standard 2024.12 with `array-api-strict>=2.3` and `array-api-compat>=1.11`. This is necessary for advanced indexing.

### Randomness
The array API does not define a standard for generating random numbers. This is probably out of scope anyways, as different frameworks have different models for randomness (compare e.g. `numpy` generators vs `jax` PRNG keys). Therefore, random constructors _always_ return the numpy version of rotations.

### Rotation splines
Rotation splines currently use `scipy.interpolate`, which is not array API compatible. Therefore, the current implementation errors on creating splines from a `Rotation` that is not backed by a numpy array.

### Open issues
`xp.roll` needs to be clarified by the array API and `array-api-compat` needs to be updated for torch, or `array-api-strict` requires additional checks for `xp.roll`. See https://github.com/data-apis/array-api/issues/914

### Open design decisions
Rewriting `Rotation` with pure array API compatible code that may also be compiled by lazy frameworks requires some deviations from the original cython implementation. Sections that deviate and require some form of discussion or decision are marked with 
```python
# DECISION: ... (or sometimes with TODO:)
```
to make it easier for reviewers to find critical points for discussion. They mostly also add a reason why the current decision was made, and may include alternatives.

## Tests
Currently, all tests for rotations have been rewritten in terms of the array API and pass locally with `numpy`, `array-api-strict`, `torch` and `jax`.

## Performance
I plan on benchmarking the current implementation for various workloads similar to the benchmarks in #22500. Ideally this includes comparing it against the current version to measure the overhead of going through the backend. This may have to wait until I have some more time at my disposal though. In case this should be integrated into the airspeed suite, I'd be grateful for guidance and/or help with that.

## TODOs

- [x] Add support for RigidTransform
- [x] Benchmark performance for different frameworks
- [x] Unify error handling
- [x] Discuss design decisions marked as # DECISION in the code
- [x] Move all docs to the Python Rotation class

##  Reference issue
Closes #22500

### DIFF
--- a/.spin/cmds.py
+++ b/.spin/cmds.py
@@ -798,7 +798,7 @@ def bench(ctx, tests, submodule, compare, verbose, quick,
             )
 
         cmd_compare = [
-            'asv', 'continuous', '--factor', '1.05', '--quick'
+            'asv', 'continuous', '--factor', '1.05'
         ] + bench_args + [commit_a, commit_b]
         _run_asv(cmd_compare)
 


tylerjereddy:
> Error handling: Since value-based checks are no longer possible, we cannot emit certain errors. Instead, we return NaN values.

At first glance, this is a compromise I'm not particularly keen on. I'm definitely in the camp of error handling being a measure of correctness for which performance shouldn't take priority. That said, the main point in saying that is just to make sure reviewers consider the tradeoffs carefully. It may be sensible to spell out our policy on this somewhere in a tracking issue or something like that if there is already a consensus I'm not aware of. I made a similar comment about "batching" using NaNs instead of hard errors recently, but this is a separate scenario I think.

I suppose the very existence of very popular lazy execution libraries is a sort of counter point that there is substantial demand for performance over error handling. Some of the lazy libraries have eager modes for better error handling/debugging though, right? Anyway, this is something I'm worried about.

lucascolley:
> > Error handling: Since value-based checks are no longer possible, we cannot emit certain errors. Instead, we return NaN values.
> 
> At first glance, this is a compromise I'm not particularly keen on. I'm definitely in the camp of error handling being a measure of correctness for which performance shouldn't take priority. That said, the main point in saying that is just to make sure reviewers consider the tradeoffs carefully. It may be sensible to spell out our policy on this somewhere in a tracking issue or something like that if there is already a consensus I'm not aware of.

@crusaderky may have words to add here

crusaderky:
Please follow the pattern from https://github.com/scipy/scipy/pull/22342:
- use `xpx.lazy_apply` to wrap around Cython kernels
- skip error checks and warnings on lazy backends (`is_lazy_array(x)`)
- if skipping an error check can cause a segfault or infinite loop in Cython code, it needs to be pushed into the function wrapped by `xpx.lazy_apply`
- if worthwile, offer explicit, eager validation functions
- be careful about injecting NaNs, particularly if your code is meant to work with integers
- use `lazy_xp_function` to test that your code is actually lazy


I'm going to write a proper guide in the documentation soon.

ev-br:
> At first glance, this is a compromise I'm not particularly keen on. I'm definitely in the camp of error handling being a measure of correctness for which performance shouldn't take priority. That said, the main point in saying that is just to make sure reviewers consider the tradeoffs carefully. It may be sensible to spell out our policy on this somewhere in a tracking issue or something like that if there is already a consensus I'm not aware of.

A big +1 for spelling out the policy and documenting it in a centralized location. https://github.com/scipy/scipy/issues/22781

amacati:
> Please follow the pattern from https://github.com/scipy/scipy/pull/22342:
> - use xpx.lazy_apply to wrap around Cython kernels
> ...

Just to make sure I understand correctly: You mean you want to exclusively use the cython implementation for performing actual computations and add a conversion wrapper for frameworks other than numpy?

amacati:
> It's not a matter of performance. Lazy backends cannot materialize the array contents with __bool__ so that you can raise/warn.

I was about to remark the same. The NaN vs eager raise discussion is not about performance. It's just not possible to raise with some lazy frameworks. The question is how consistent error handling needs to be between frameworks. 

- If it should be consistent, we either have to use NaNs also for eager frameworks, or we cannot support lazy frameworks at all. 
- If we allow inconsistencies between frameworks, we probably should make that explicit in the docs.

crusaderky:
> > Please follow the pattern from #22342:
> > 
> > * use xpx.lazy_apply to wrap around Cython kernels
> >   ...
> 
> Just to make sure I understand correctly: You mean you want to exclusively use the cython implementation for performing actual computations and add a conversion wrapper for frameworks other than numpy?

- _If_ a native dispatch is not available,
- _and_ a array API agnostic variant is not feasible,
- _then_ as a last resort backends other than numpy - including dask and jax - should use `xpx.lazy_apply` to use a cython function, as long as the data is not on GPU memory.

lucascolley:
right, and this PR addresses "an array API agnostic variant is not feasible (or delayed to a future date)" for `spatial.transform`

tylerjereddy:
> I was about to remark the same. The NaN vs eager raise discussion is not about performance.

I'm not sure I follow--why would we want to add lazy backend support if not for performance reasons?

The fact that we're contemplating adding the lazy backend support implies that there's a performance benefit to doing so--if performance is not involved here, what is the true motivation? What I'm saying is that if we want a single common code path for lazy and eager backends that would mean that we sacrifice error handling for the eager case, which I'm not fond of.

If we specialize code paths for lazy and eager and provide them with different error handling, that would seem to have different drawbacks that somewhat move away from the original array API implementation ideals (though I think specialized code paths have proliferated somewhat regardless).

I also don't really understand the design of the `is_lazy` or `lazy_` shims of various sorts--can these legitimately distinguish between a potentially-lazy backend that is currently in eager mode vs. its default lazy mode for example?

In short, if you're not compromising, shimming or adjusting the error handling of the eager libraries at all, then I'm less worried. If you are making *any* adjustments to the error handling behavior of the eager libraries, then I maintain that this is ultimately a performance (support for lazy backends) vs. error handling matter.

scipy/spatial/transform/_array_api_backend.py
```diff
@@ -0,0 +1,1076 @@
...
+
+    gramians = matrix @ xp.matrix_transpose(matrix)
+    # TODO: We need to orthogonalize only the non-orthogonal matrices, but jax does not support
+    # non-concrete boolean indexing or any form of computation without statically known shapes. We
+    # either have to branch depending on lazy/non-lazy frameworks or pay the performance penalty for
+    # the SVD.
+    is_orthogonal = xp.all(xpx.isclose(gramians, xp.eye(3), atol=1e-12, xp=xp))
+    U, _, Vt = xp.linalg.svd(matrix)
+    orthogonal_matrix = U @ Vt
+    matrix = xp.where(is_orthogonal, matrix, orthogonal_matrix)
```
lucascolley, line 52: We can use https://data-apis.org/array-api-extra/generated/array_api_extra.apply_where.html here, to avoid paying the extra cost for libraries which do have the boolean indexing capability—see https://github.com/data-apis/array-api-extra/blob/66e9282df165df219797f65f1cc061665fe7e95f/src/array_api_extra/_lib/_funcs.py#L156-L158.

scipy/spatial/transform/_array_api_backend.py
```diff
@@ -0,0 +1,1076 @@
...
+        axis=-1,
+    )
+    quat = xp.where((choice == 3), quat_3, quat)
+
+    return _normalize_quaternion(quat)
+
+
+def from_rotvec(rotvec: Array, degrees: bool = False) -> Array:
+    xp = array_namespace(rotvec)
+    rotvec = xp.asarray(rotvec, copy=True, dtype=atleast_f32(rotvec))
```
lucascolley, line 118: We may want to use https://github.com/scipy/scipy/blob/d72842a3cfb653e2fb764e6670590492206827f2/scipy/_lib/_array_api.py#L568 in cases like this.

scipy/spatial/transform/_array_api_backend.py
```diff
@@ -0,0 +1,1076 @@
...
+            "sequences or 'i'/'intrinsic' for intrinsic "
+            "sequences, got {}".format(order)
+        )
+    if axes.shape[0] != 3:
+        raise ValueError(f"Expected 3 axes, got {axes.shape}.")
+    if axes.shape[1] != 3:
+        raise ValueError("Axes must be vectors of length 3.")
+
+    # normalize axes
+    axes = axes / xp.linalg.vector_norm(axes, axis=-1, keepdims=True)
```
lucascolley, line 376: https://github.com/scipy/scipy/blob/d72842a3cfb653e2fb764e6670590492206827f2/scipy/_lib/_array_api.py#L483 is slightly more robust in handling the case that `xp` does not implement the `linalg` extension.

scipy/spatial/transform/_array_api_backend.py
```diff
@@ -0,0 +1,1076 @@
...
+            "{} values and {} rotations.".format(weights.shape[0], n_rot)
+        )
+    # DECISION: We cannot check for negative weights because jit code needs to be non-branching. We
+    # return NaN instead
+    weights = xp.where(weights < 0, xp.asarray(xp.nan, device=device(quat)), weights)
+
+    # Make sure we can transpose quat
+    quat = xpx.atleast_nd(quat, ndim=2, xp=xp)
+    K = (weights * quat.T) @ quat
+    _, v = xp.linalg.eigh(K)
```
lucascolley, line 463: strictly, all uses of `xp.linalg` should be wrapped in `if hasattr(xp, "linalg")`. Alternatively, if no fallback implementation is easy, we can just declare that the module requires `xp` to implement the `linalg` extension. In that case, it is probably better to add a higher-level check for an early exit.

scipy/spatial/transform/_array_api_backend.py
```diff
@@ -0,0 +1,1076 @@
...
+    n_rot = quat.shape[0] if quat.ndim > 1 else 1
+    if weights.shape[0] != n_rot:
+        raise ValueError(
+            "Expected `weights` to have number of values "
+            "equal to number of rotations, got "
+            "{} values and {} rotations.".format(weights.shape[0], n_rot)
+        )
+    # DECISION: We cannot check for negative weights because jit code needs to be non-branching. We
+    # return NaN instead
+    weights = xp.where(weights < 0, xp.asarray(xp.nan, device=device(quat)), weights)
```
lucascolley, line 458: perhaps a use-case for `is_lazy`?

scipy/spatial/transform/_array_api_backend.py
```diff
@@ -0,0 +1,1076 @@
...
+    # 1. Computing both for eager execution models is expensive.
+    # 2. Some operations will fail when running the unused branch because of numerical and
+    # algorithmical issues. Numpy e.g. will raise an exception when trying to compute the svd of a
+    # matrix with infinite weights. To prevent this, we only compute the branch that is needed. Jax
+    # jit however requires us to take the full compute graph. Therefore, we use xp.where for jax and
+    # a branching version for non-jax frameworks. Might apply for other lazy frameworks as well.
+    #
+    # Note that we could also solve this by exploiting the externals of xpx.apply_where. However,
+    # we'd have to rely on the implementation details of apply_where, which is something we should
+    # avoid in the long run.
```
lucascolley, line 639: could you explain further the problem with 'relying on implementation details of apply_where' here? Is there anything we can make public API / documented in apply_where which would take things out of just 'implementation details'?

cc @crusaderky

scipy/spatial/transform/_array_api_backend.py
```diff
@@ -0,0 +1,1076 @@
...
+    if is_jax(xp):
```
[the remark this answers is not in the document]
crusaderky: ...or just use `apply_where`, which should fix your problem unless we missed something while writing it.

crusaderky:
> > I was about to remark the same. The NaN vs eager raise discussion is not about performance.
> 
> I'm not sure I follow--why would we want to add lazy backend support if not for performance reasons?

Namely, because _this_ function may not get a substantial performance benefit from running on Dask or JAX, but everything else in the user's workflow before and after it could.

> What I'm saying is that if we want a single common code path for lazy and eager backends that would mean that we sacrifice error handling for the eager case

Specifically for error handling, we _don't_ want a single common code path because error handling is valuable for eager backends and impossible for lazy ones.

> If we specialize code paths for lazy and eager and provide them with different error handling, that would seem to have different drawbacks that somewhat move away from the original array API implementation ideals

It is important to point out that we are talking about _two_ code paths, total: one for eager and one for lazy. Which are two very broad categories of backends.
All hacks and special cases specific to a _single_ backend are designed to be self-contained in `array-api-compat` and `array-api-extra`.

> (though I think specialized code paths have proliferated somewhat regardless).

We're working to remove them, at least from the production code. Unit tests are harder to simplify as there are frequently quirks in behaviour when this or that backend diverges from the Standard.

> I also don't really understand the design of the `is_lazy` or `lazy_` shims of various sorts--can these legitimately distinguish between a potentially-lazy backend that is currently in eager mode vs. its default lazy mode for example?

They don't, by design.
JAX in eager mode is generally regarded by its developers only as a debugging/development clutch, with very little value in production use cases as it is enormously slower than numpy/cupy. So there was a conscious decision not to expose a "jitted/eager" flag, in order to reduce the testing surface in downstream projects.

Another "potentially lazy" backend is ndonnx (not yet supported by scipy), which I understand is intended mainly for ML training where it switches to lazy mode.

tylerjereddy:
> It is important to point out that we are talking about two code paths, total: one for eager and one for lazy. 

Understood, I think my main concern evaporates then, at least for this specific case. I was worried both were being unified to the detriment of eager error handling to keep maintenance burden under control.

scipy/spatial/transform/_array_api_backend.py
amacati: Yes, I was thinking about that. But I was under the impression that `apply_where `applies element-wise, and here we would need to apply this to whole matrices. I have experimented a bit with `apply_where`, and it seems to flatten the input and then applies the function. We _could_ reassemble the matrices from the resulting [M * 3 * 3] array, but that would mean relying on the internals of `apply_where`

I have not dived deeper though and might have missed something.

scipy/spatial/transform/_array_api_backend.py
amacati: Oh nice, yes, that would work. Do you know if that handles float64 / float32 for `jax`? The code uses a bunch of `atleast_f32` promotions that could be replaced in that case.

scipy/spatial/transform/_array_api_backend.py
amacati: Agreed, though it depends on how we want to handle errors. Are we okay with lazy and eager frameworks producing different outcomes?

scipy/spatial/transform/_array_api_backend.py
amacati: The issue is that (as far as I tested it), `apply_where` flattens the input and the applies the function to the flattened array. We could write a function that unflattens the array again, performs the operation, and then flattens it, only for `apply_where` to unflatten it again. 

However, as far as I understood, elementwise does not imply we are guaranteed the function is called with the whole array. It would also be possible for implement something where each array element is being processed individually (although it would be nonsensical). 

That's what was meant here. I have not studied `apply_where` in greater detail though, so there might be something I'm missing.

scipy/spatial/transform/_array_api_backend.py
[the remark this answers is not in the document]
amacati: Agreed, there are a bunch of places where the code checks for jax instead of using is_lazy. I started using the array API seriously for the first time and wasn't aware that function existed. I have commented about `apply_where` further up, I believe lucas cc'ed you there as well. Seems like I misunderstood something about the function. I was under the impression it only applies a function element-wise to an array and flattens it in the process.

(https://github.com/data-apis/array-api-extra/blob/66e9282df165df219797f65f1cc061665fe7e95f/src/array_api_extra/_lib/_funcs.py#L160, indexing with broadcasted cond should effectively flatten the arrays, right?)

scipy/spatial/transform/_array_api_backend.py
lucascolley: discussion in gh-22781 but it seems like yes

scipy/spatial/transform/_array_api_backend.py
lucascolley: yes I think that will use `jax.numpy.result_type(rotvec, 1.0)` under the hood

scipy/spatial/transform/_array_api_backend.py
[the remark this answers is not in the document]
crusaderky: it doesn't flatten it. As long as your backend supports `x[cond] = y[cond]`, plus special-casing for JAX and Dask, apply_where is expected to work.

scipy/spatial/transform/_array_api_backend.py
crusaderky, line 61:
```suggestion
    is_orthogonal = xp.all(xpx.isclose(gramians, xp.eye(3), atol=1e-12, xp=xp))
    is_lazy = is_lazy_array(is_orthogonal)
    if is_lazy or is_orthogonal:
        U, _, Vt = xp.linalg.svd(matrix)
        orthogonal_matrix = U @ Vt
        matrix = (xp.where(is_orthogonal, matrix, orthogonal_matrix)
                  if is_lazy else orthogonal_matrix)
```

crusaderky: Note that `is_lazy or is_orthogonal` is very different from `is_orthogonal or is_lazy`, as the latter will materialize the array.
This shouldn't warrant a comment here, but it needs to be paired by `lazy_xp_function` in the unit tests to catch it.

amacati: Thanks for the suggestion, will have a look later! One issue I was trying to get at with my comment though: `is_orthogonal` really should be a boolean mask here. Ideally we'd have something like 
```python
is_orthogonal = xpx.isclose(gramians, xp.eye(3), atol=1e-12, xp=xp)
U, _, Vt = xp.linalg.svd(matrix[~is_orthogonal[..., None, None])
matrix[~is_orthogonal[..., None, None]] = U @ Vt
```
for non-lazy frameworks. Is there a convenient way to express this with `apply_where`? This is one of the cases where the function would not have to be applied element-wise, but only over the first axis. Then we could have an identity for the first function, and the svd as the second function.

crusaderky: The functions you pass to `apply_where` are executed on a flattened array, because `apply_where` has no way to know that your mask has been broadcasted along axes 1 and 2.
What you are basically asking for is a variant of `apply_where` that works along an axis, e.g. 
```python
def make_orthogonal(matrix):
    U, _, Vt = xp.linalg.svd(matrix)
    return U @ Vt

matrix = apply_along_axis(~is_orthogonal, matrix, make_orthogonal, fill_value=matrix, axis=0)
```
but it does not exist today.

Untested code below:

```python
is_orthogonal = xpx.isclose(gramians, xp.eye(3), atol=1e-12, xp=xp)
if is_lazy_array(is_orthogonal) or not xp.all(is_orthogonal):
    U, _, Vt = xp.linalg.svd(matrix)
    orthogonal_matrix = U @ Vt
    matrix = xp.where(is_orthogonal[..., None, None], matrix, orthogonal_matrix)
```

Fancier, and I would expect faster (but please run a performance test!)
Again very, VERY untested.

```python
def make_orthogonal(m):
    U, _, Vt = xp.linalg.svd(m)
    return U @ Vt

is_orthogonal = xpx.isclose(gramians, xp.eye(3), atol=1e-12, xp=xp)
lazy = is_lazy_array(is_orthogonal)
if lazy or not xp.all(is_orthogonal):
    if not lazy and not xp.any(is_orthogonal):
        matrix = make_orthogonal(matrix)
    # Temporarily _may_ need to special-case Dask and/or JAX to use this code path
    elif not xp.__array_api_info__().capabilities()["boolean indexing"]:
        matrix = xp.where(is_orthogonal[..., None, None], matrix, make_orthogonal(matrix))
    else:    
        # Calculate dot product ( ~O(n^3) ) only on the needed planes

        # As of 2024.12, the Array API standard does not allow for
        # non_orthogonal_matrix = matrix[non_orthogonal, :, :]
        int_idx = xp.arange(is_orthogonal.shape[0], device=xp_device(matrix))[~is_orthogonal]
        non_orthogonal_matrix = xp.take(matrix, int_idx, axis=0)

        orthogonal_matrix = make_orthogonal(non_orthogonal_matrix)

        # As of 2024.12, the Array API standard does not allow for
        # matrix[non_orthogonal, :, :] = orthogonal_matrix
        # reshape orthogonal_matrix to match matrix.
        # indices along axis 0 of `matrix` of the planes of `orthogonal_matrix`, or
        # arbitrary dummy indices when non_orthogonal is false
        reverse_int_indx = xp.cumulative_sum(xp.where(is_orthogonal, 0, 1)) - 1
        orthogonal_matrix = xp.take(orthogonal_matrix, reverse_int_indx, axis=0)
        matrix = xp.where(is_orthogonal[..., None, None], matrix, orthogonal_matrix)
```

lucascolley: Am I right in thinking that this would be in a sense an extension of `put_along_axis` discussed in https://github.com/data-apis/array-api/issues/177 to also apply an arbitrary function where we have `cond`?

crusaderky: not `put_along_axis`, just `put`. Which however would not be compatible with read-only backends.

crusaderky: I think that an ideal place to reach with the API, which would require both `__getitem__` and `__setitem__` to be allowed with tuple indices of one flat bool array and a bunch of `slice(None)`, would be:

```python
def make_orthogonal(m):
    U, _, Vt = xp.linalg.svd(m)
    return U @ Vt

non_orthogonal = ~xpx.isclose(gramians, xp.eye(3), atol=1e-12, xp=xp)
if xp.__array_api_info__().capabilities()["boolean indexing"]:
    non_orthogonal_matrix = matrix[non_orthogonal, :, :]
    orthogonal_matrix = make_orthogonal(non_orthogonal_matrix)
    matrix = xpx.at(matrix)[non_orthogonal, :, :].set(orthogonal_matrix)
else:
    matrix = xp.where(non_orthogonal[..., None, None], make_orthogonal(matrix), matrix)
```

amacati: Hmm, the latter seems like it's going in the direction of "fast, but high maintenance burden" territory. 

Having the capability to use `apply_where` along axes would be fantastic and is exactly what I was referring to in earlier comments. We can emulate that here somewhat by reshaping the arrays back within the two functions (because we know that we are applying over axes with known shapes), but that was what I meant with "relying on the implementation specifics of `apply_where`". It means relying on the way the function is written, not on its guarantees. 

_(All just gists, not actual code)_
```python
def fn_orthogonal(x):
    return x

def fn_non_orthogonal(x):
    x = xp.reshape(x, (-1, 3, 3))
    U, _, Vt = xp.linalg.svd(x)
    return (U @ Vt).flatten()
```

which would break if for some reason the `apply_where` function changes to

```python
def _apply_where(...):
    ...
    # temp1 = f1(*(arr[cond] for arr in args))  <- Current implementation
    temp1 = xp.asarray([f1(*(arr[c] for arr in args)) for c in cond])
    ...
```

crusaderky: Right. This is something you can try today with `apply_where` (because you know that cond is broadcasted along the last two dimensions):

```python
if is_dask(xp):
    # Do not apply make_orthogonal to parts of the planes of the matrix.
    matrix = matrix.rechunk("auto", -1, -1)
    gramians = gramians.rechunk(matrix.chunks[0], )

def make_orthogonal(x):
    # x may or may not have been flattened by apply_where
    xm = xp.reshape(x, (-1, 3, 3))
    U, _, Vt = xp.linalg.svd(xm)
    return xp.reshape(U @ Vt, x.shape)

non_orthogonal = ~xpx.isclose(gramians, xp.eye(3), atol=1e-12, xp=xp)
matrix = xpx.apply_where(non_orthogonal[..., None, None], matrix,
                         make_orthogonal, fill_value=matrix, xp=xp)
```

For clarity, `matrix[i, :, :]` needs to be always small enough to fit in memory.
Which is fine if it's hardcoded (3, 3).
I need to go and write a huge warning in the documentation of `apply_where` about that rechunk bit...

amacati:
I have now added full support for `RigidTransform` as well, and all tests are passing locally. Also, error handling is now unified. Eager frameworks raise just like NumPy, lazy frameworks return NaN values.

I'll try to create some benchmarks that compare the new implementation for different frameworks and devices to the old one next. Once we have some numbers and the performance looks okay, we can maybe discuss what's left to do before this PR can be merged.

amacati:
One frequent issue that is coming up while I'm going through the benchmarks: Some functions may work just fine on the CPU, but will raise device errors when the original `Rotation` is created on the GPU, but intermediate arrays are created on the CPU. The tests do not capture these cases at the moment. In theory, it would be necessary to run the tests for all devices. Is there precedence for this in scipy?

lucascolley:
we're still thinking about how best to test devices in gh-22680

In array-api-extra, we use an enum, https://github.com/data-apis/array-api-extra/blob/main/src/array_api_extra/_lib/_backends.py. But in SciPy we also have the environment variable `SCIPY_DEVICE`.

lucascolley:
If the important part for now is just that the same tests are ran on the GPU, that should be covered by https://github.com/scipy/scipy/blob/main/.github/workflows/gpu-ci.yml

amacati:
Here are the benchmarks for `Rotation`. They include `numpy` with `scipy 1.15.2`, `numpy` with this PR, `torch` on CPU and GPU, `jax` with just-in-time compilation on CPU and GPU, and `cupy`. The full report is added as PDF. 

There is a performance penalty for determining the array framework etc. that is mainly noticeable at sample sizes from 1 to 10. Nevertheless, this should be a very common use case, so we should probably take that serious and try to close the gap. I'm also not a cython expert, so if someone has more expertise in that area feel free to jump in with any ideas on how to improve typing etc. for more performance / less overhead.

Some samples:

from_quat
as_quat
from_euler
as_euler
mul
apply

>Note: Some benchmarks are missing `cupy`. That is because the xp backend currently makes use of xp.array.mT, which will be supported in the upcoming release of `cupy 14`. See https://github.com/data-apis/array-api-compat/issues/316

I will try to produce the same benchmark for `RigidTransform` asap, and then I think we are finally at a point where a serious review makes sense.

[rotation_benchmark.pdf](https://github.com/user-attachments/files/19864065/rotation_benchmark.pdf)

amacati:
> You mean like this 😃 https://github.com/data-apis/array-api-compat/pull/308

Yes, very much like this! Exciting to see that the overhead before jumping into cython will get smaller. 

Another thing that may be an easy optimization is to use canonical input shapes and types for cython, because that would allow for better typing. If xp operations get faster, it might make sense to perform these conversions in Python before handing over the arrays to cython. In some places this is what's already happening, but in others I didn't see performance gains. Would be great to see how this evolves with faster xp and optimizations from someone with better cython knowledge.

crusaderky:
> Some samples:

Are the JAX benchmarks using eager JAX, or are they running in `jax.jit`?

amacati:
All running with `jax.jit` and compiled during the benchmark setup

amacati:
Here's the benchmark for `RigidTransform`. Overall the picture is similar to `Rotation`.

The PR should now be ready for review. Let me know what you think @crusaderky @lucascolley 

[tf_benchmark.pdf](https://github.com/user-attachments/files/19885472/tf_benchmark.pdf)

scipy/spatial/transform/tests/test_rigid_transform.py
```diff
@@ -29,824 +36,1151 @@ def test_repr():
...
-        invalid[3, 3] = 2  # Invalid last row
-        RigidTransform.from_matrix(invalid)
+    invalid = xp.eye(4)
+    invalid = xpx.at(invalid)[..., 3, 3].set(2)  # Invalid last row
+    if is_lazy_array(invalid):
+        tf = RigidTransform.from_matrix(invalid)
+        assert xp.all(xp.isnan(tf.as_matrix()))
+    else:
+        with pytest.raises(ValueError):
+            RigidTransform.from_matrix(invalid)
```
crusaderky, line 155: Do users of lazy backends have the means to validate their input? e.g. is there an eager user-facing validation function like `scipy.cluster.is_valid_linkage`?

scipy/spatial/transform/tests/test_rigid_transform.py
```diff
@@ -29,824 +36,1151 @@ def test_repr():
...
+        tf.as_matrix()[:, 3, :], xp.asarray([[0.0, 0, 0, 1], [0, 0, 0, 1]]), atol=atol
+    )
     assert not tf.single
 
 
-def test_from_translation():
+def test_from_rotation_jax_compile():
+    pytest.importorskip("jax")
+    import jax
+    import jax.numpy as jp
```
crusaderky, line 72: canonical short name is `jnp`

scipy/spatial/transform/tests/test_rigid_transform.py
```diff
@@ -1002,28 +1373,35 @@ def test_empty_transform_concatenation():
...
         tf_zero[0]
 
-    with pytest.raises(IndexError):
-        tf_zero[[0, 2]]
+    if is_jax(xp):  # DECISION: Should we manually check and raise for jax here?
+        with pytest.raises(TypeError, match="Slice size at index 0 in gather op"):
+            tf_zero[xp.asarray([0, 2])]
+    else:
+        with pytest.raises(IndexError):
+            tf_zero[xp.asarray([0, 2])]
```
crusaderky, line 1404: I assume this is because of https://docs.jax.dev/en/latest/notebooks/Common_Gotchas_in_JAX.html#out-of-bounds-indexing?
Does the out of bounds index derive from a Python scalar or from the values of the input array (e.g. via `xp.take`)? In the former case, couldn't your own Python code check it?

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -839,445 +940,474 @@ def test_as_euler_degenerate_symmetric_axes(seq_tuple, intrinsic):
...
+            angle_estimates = rotation.as_euler(seq, degrees=True)
     mat_estimated = Rotation.from_euler(seq, angle_estimates, degrees=True).as_matrix()
 
-    assert_array_almost_equal(mat_expected, mat_estimated)
-
-
-@pytest.mark.thread_unsafe
-@pytest.mark.parametrize("seq_tuple", permutations("xyz"))
-@pytest.mark.parametrize("intrinsic", (False, True))
-def test_as_euler_degenerate_compare_algorithms(seq_tuple, intrinsic):
```
crusaderky, line 852: What's the reason for deleting? Did the behaviour on numpy change?

scipy/spatial/transform/tests/test_rotation_spline.py
```diff
@@ -160,3 +161,19 @@ def test_error_handling():
     raises(ValueError, s, 10, -1)
 
     raises(ValueError, s, np.arange(10).reshape(5, 2))
+
+
+def test_xp_errors(xp):
+    if is_numpy(xp):
+        return
```
crusaderky, line 168:
```suggestion
@pytest.mark.skip_xp_backends("numpy")
def test_xp_errors(xp):
```

scipy/spatial/transform/_rigid_transform_xp.py
```diff
@@ -0,0 +1,350 @@
...
+    xp = array_namespace(matrix)
+    # Shape check should be done before calling this function
+    if normalize or copy:
+        matrix = xp.asarray(matrix, copy=True)
+
+    last_row_ok = xp.all(
+        matrix[..., 3, :] == xp.asarray([0, 0, 0, 1.0], device=device(matrix)), axis=-1
+    )
+    if is_lazy_array(matrix):
+        matrix = xp.where(last_row_ok[..., None, None], matrix, xp.nan)
```
crusaderky, line 28: To my limited understanding, the input can be integer.
This causes lazy backends to automatically promote everything to float64, if they support int to float promotion (which is undefined behaviour in the Standard).
```suggestion
        matrix = xp_promote(matrix, force_floating=True)
        matrix = xp.where(last_row_ok[..., None, None], matrix, xp.nan)
```
Is this behaviour documented? Is it tested?

crusaderky, line 28: To my limited understanding, the input can be integer.
This causes lazy backends to automatically promote everything to float64, if they support int to float promotion (which is undefined behaviour in the Standard).
```suggestion
        matrix = xp_promote(matrix, force_floating=True, xp=xp)
        matrix = xp.where(last_row_ok[..., None, None], matrix, xp.nan)
```
Is this behaviour documented? Is it tested?

crusaderky (commented):
I gave a very quick, very _very_ incomplete glance at it.
The short summary:

- 14500 lines worth of change in a single PR makes it extremely hard for any human reviewer to check if there is anything spurious going on.
- There are a wealth of changes that could hopefully be cherry picked out of this and moved into one or more prerequisite refactoring PRs.
- JAX-specific design issues (see comments)
- Is there an eager user API for validating lazy inputs, like `scipy.cluster.is_valid_linkage`? (It may already be there, again I didn't read through the whole thing).
- Is the auto-promotion from int to float of lazy backends tested and documented?
 
Finally, ahead of further review, could you merge with main and reach a green-ish CI?

amacati:
> 14500 lines worth of change in a single PR makes it extremely hard for any human reviewer to check if there is anything spurious going on.

Yes, this is a huge headache. There are four main reasons:

- The Array API compatible backend is just straight up a lot of new code
- The rewrite of `_rotation.pyx` into `rotation.py` and `_rotation_cy.pyx` lets git believe that the whole docs have been removed and added again. This alone are probably several thousand lines worth of change. Similarly, it cannot recognize the split of `_rigid_transform.pyx` into `_rigid_transform.py` and `_rigid_transform.pyx`
- The tests changed to be Array API compatible
- I ran the scipy linting as recommended with `python dev.py lint`, but this does not enforce unified formatting, so I guess my ruff formatter changes were preserved

There are probably two low-hanging fruits here. It would be easy to factor out the test changes into a PR that introduces Array API compatible tests. 

Also, all spurious format changes can also probably be factored out. On that note, I may have missed this, but is there no consistent format style for `scipy`? Having a ruff section in the `pyproject.toml` or a `ruff.toml` would be very helpful for this.

The first and the second issue are a bit tougher. I could introduce two PRs that split the changes into two: Splitting the cython classes into pure Python "frontend" classes and functional Cython backends, and then introducing the Array API backend in a separate PR. 

What's your opinion on that @crusaderky ?

ev-br:
> It would be easy to factor out the test changes into a PR that introduces Array API compatible tests.

+1. A strategy of converting tests in a separate PR works well.

> On that note, I may have missed this, but is there no consistent format style for scipy? Having a ruff section in the pyproject.toml or a ruff.toml would be very helpful for this.

Please drop this. This conversation we had more than once, so please limit the your autoformatter (or factor autoformatting of the parts you are going to change into a separate PR or at least a separate commit).

> The first and the second issue are a bit tougher. I could introduce two PRs that split the changes into two: Splitting the cython classes into pure Python "frontend" classes and functional Cython backends, and then introducing the Array API backend in a separate PR.

Splitting these sounds good.

-------------------------

On a higher level, I am still very concerned about the maintenance load from duplicating the code into the "array api backend".
I'm not going to block it of course, if `scipy.spatial` maintainers are OK with this.
Still I suggest to seriously consider why this is in scipy and somewhere backend-specific, which hooks into scipy via some form of delegation.

lucascolley:
> is there no consistent format style for scipy? Having a ruff section in the pyproject.toml or a ruff.toml would be very helpful for this.

@amacati check out the thread at https://github.com/astral-sh/ruff/discussions/7310#discussioncomment-7266904 if you are interested in what the community has felt we need from ruff before this seems like a good idea. Also, if you are interested in helping work towards this end, you could drop a review at https://github.com/scientific-python/specs/pull/326 !

---

> I suggest to seriously consider why this is in scipy and somewhere backend-specific, which hooks into scipy via some form of delegation.

On the contrary (with a bit of forward thinking), if we are wanting to split backend-specific code out into a separate repo, it is the NumPy-specific Cython which ought to be maintained separately.

scipy/spatial/transform/_rotation_xp.py
lucascolley, line 1: high-level comment: I think this file should just be called `_rotation.py`. I think it is fine to not pick out the generic implementation with a special name.

lucascolley: EDIT 2: actually, for `_rigid_transform_xp`, it is probably better to have the split since we don't have a public `rigid_transform.py` file. And it is probably good to match that pattern here. So feel free to disregard/resolve.

amacati: Actually, I would like to follow the pattern of `_rotation_<backend>.py` and keep `_rotation_xp.py` for another reason. The `rotation.py` file is scheduled for deprecation in scipy 2.0, and then you could simply rename `rotation.py` -> `_rotation.py` to mark the submodule as private. Keeping `_rotation_xp.py` prevents the name collision that would occur in that case.

amacati:
> Still I suggest to seriously consider why this is in scipy and somewhere backend-specific, which hooks into scipy via some form of delegation.

> On the contrary (with a bit of forward thinking), if we are wanting to split backend-specific code out into a separate repo, it is the NumPy-specific Cython which ought to be maintained separately.

This is why I think the backend design makes a lot of sense. It should be very easy to separate either backend out at a later time into its own package if you wish to. Adding backends back in is then just be a matter of installing `scipy-accel` or whatever that package would be called, and then let it register

scipy/spatial/transform/rotation.py
```diff
@@ -2,20 +2,2649 @@
...
+        r._backend = backend_registry.get(array_namespace(c[0]), array_api_backend)
+        r._quat = c[0]
+        # We set _single to False for jax because the Array API backend supports broadcasting by
+        # default and hence returns the correct shape without the _single workaround
+        r._single = False
+        return r
+
+    register_pytree_node(Rotation, lambda v: ((v._quat,), None), rot_unflatten)
+except ImportError:
+    pass
```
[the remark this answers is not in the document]
crusaderky: FYI https://github.com/data-apis/array-api-extra/pull/284

crusaderky:
> > There is a performance penalty for determining the array framework etc. that is mainly noticeable at sample sizes from 1 to 10. Nevertheless, this should be a very common use case, so we should probably take that serious and try to close the gap.
> 
> You mean like this 😃 [data-apis/array-api-compat#308](https://github.com/data-apis/array-api-compat/pull/308)

The PR above is now in scipy main; if you merge from main and rerun the benchmarks you should see its impact.

scottshambaugh (commented):
I find this a rather difficult PR to review, since big chunks of code are split out to new files and I'm unfamiliar with the array API & its conventions.

However I am familiar with the existing tests here, and the testing of all the behavior of this module is very thoroughly covered. I reviewed the test files and barring the comments I left, I can confirm that the test suite remains functionally unchanged. This serves to prove that the code changes here have not modified the expected behavior.

scipy/spatial/transform/tests/test_rigid_transform.py
```diff
@@ -29,824 +36,1151 @@ def test_repr():
...
+            ],
+            [
+                [0.50440981, 0.2957028, 0.81125249, 1.20934468],
+                [0.08979911, 0.91647262, -0.3898898, -0.70540077],
+                [-0.8587822, 0.26951399, 0.43572393, -0.47776265],
+                [0.0, 0.0, 0.0, 1.0],
+            ],
+        ]
+    )
+    # The tolerance is set to 1e-8, because xp_assert_close compares the absolute difference instead
```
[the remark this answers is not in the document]
lucascolley: > The previous 1e-12 should still be a bounding absolute error to compare against here, and not including rtol would only lower the expected error

I'm not sure I follow the logic here. Yes, adjusting the implicit `rtol` lowers the expected error. Hence, the tests fail, as the implementation no longer meets the stricter expectation.

https://github.com/scipy/scipy/blob/9c021071c05bca708df316d38b2d4316ea413c48/scipy/_lib/_array_api.py#L366-L373

`1e-12` was not acting as a bounding absolute error, due to the implicit `rtol`.

So I don't think the change implies that the results are becoming less precise (of course, we should still check).

The important question is what mixture of `atol` and `rtol` would be most appropriate here.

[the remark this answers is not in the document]
lucascolley: I think the code comment here seems inaccurate, though: the linked lines show that `rtol` is set to a nonzero value when the argument is omitted, so it isn't just comparing the absolute difference.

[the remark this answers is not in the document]
scottshambaugh: Ah, yeah I was getting things backwards, should be fine as-is. I think the correct comparison for hardcoded digits as is done here is an atol equal to the resolution of the hardcoded number. I.e. 8 decimal places would compare to 1e-8 as is done in the new code.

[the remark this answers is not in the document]
crusaderky: If swapping `assert_allclose` for `xp_assert_close` is nontrivial, it should be moved to a separate PR which can become a dependency of this one. `xp_assert_close` runs on non-array-api numpy too.
Basically everything that makes sense on its own as a propaedeutic step to this should be moved out in order to reduce the size of this PR as much as possible.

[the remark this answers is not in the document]
lucascolley: yes, agreed

lucascolley:
BTW @amacati, it looks like you aren't the only one interested in using this! https://github.com/magpylib/magpylib/issues/792

amacati:
Sorry for taking so long, had a lot going on recently. I have now split up the testing into a separate PR #22939. The PR is optimized to reduce the amount of line changes, but still comes in at ~ +2000/-1200 lines. I don't think I can improve significantly on that, it's just a lot of code that has changed.

The plan for moving forward (in this order) is:

- A PR to disentangle `RigidTransform` into a functional-style cython backend and a pure python frontend class
- A PR to disentangle `Rotation` into a functional-style cython backend and a pure python frontend class
- A PR to add the Array API backend for `Rotation`
- A PR to add the Array API backend for `RigidTransform`

All reviews can focus on these isolated PRs to make things less overwhelming. Once these PRs are merged, we can close this PR (probably without merging, as its parts have been merged individually).

@maintainers please let me know if that works for you. Splitting this PR up into stages without breaking the `transform` module in between is a lot more work than one might think, so it would be great to know that this effort is not in vain.

Also, I know that `scipy` has no consistent formatter and does not intent to have one anytime soon (see https://github.com/scipy/scipy/pull/22777#issuecomment-2827585033). However, it would be nice to at least have a formatter for `scipy.spatial.transform`, even if it's just a recommendation. Trying to match the format style of the old implementations to reduce the amount of random format changes is tedious work. Having at least a comment on how to replicate that style somewhere in the files would be immensely helpful.

scipy/spatial/transform/tests/test_rotation_groups.py
```diff
@@ -9,7 +9,7 @@
 from scipy.spatial import cKDTree
 
 
-TOL = 1E-12
+TOL = 1e-12
```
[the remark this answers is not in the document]
amacati: Done (by avoiding for now). See #22939

scipy/spatial/transform/tests/test_rigid_transform.py
```diff
@@ -29,824 +36,1151 @@ def test_repr():
...
-         [0., 0., 0., 1.]],
-        [[-0.78446989, 0.61157488, 0.10287448, 1.33330055],
-         [-0.58017785, -0.78232107, 0.22664378, 0.52660831],
-         [0.21909052, 0.11810973, 0.96852952, -0.02968529],
-         [0., 0., 0., 1.]]
-    ]
-    assert_allclose(
+    exp_coords = xp.asarray(
+        [
+            [-2.01041204, -0.52983629, 0.65773501, 0.10386614, 0.05855009, 0.54959179],
```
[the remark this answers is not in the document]
amacati: See the isolated testing PR #22939. I strongly agree on this one. It would be great to have at least a consistent formatter for `scipy.spatial.transform`.

scipy/spatial/transform/tests/test_rigid_transform.py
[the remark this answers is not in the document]
amacati: See #22939

lucascolley:
sounds like a plan! We can add to the list: converting the tests to use the `xp_assert` infrastructure, separately and before making the changes needed in the implementations.

amacati:
> sounds like a plan! We can add to the list: converting the tests to use the xp_assert infrastructure, separately and before making the changes needed in the implementations.

Ah yes, but this is already completed in #22939 and ready for review, so I didn't list it

amacati:
Array API tests from #22939 have now been merged. The PR to refactor RigidTransform into a functional-style cython backend and a pure python frontend class is now open and ready for review in #22972

purepani:
Did you mean to clear all the commits here?

amacati:
I thought it might make sense to keep this thread open for the general discussion until everything is merged from smaller PRs and then close this PR without merging.

ev-br:
Now that I looked at https://docs.jax.dev/en/latest/jax.scipy.html, apparently jax has a native implementation of `jax.scipy.Rotation`. Did somebody try to benchmark it against the implementation from this effort? Ideally, both in eager mode and under jit.

amacati:
No, not yet. It would definitely be interesting to compare against those functions that are implemented in the jax version (it only supports a subset)

amacati:
Just for future reference: Most of the functionality that was discussed here has been implemented in the PRs that reference this initial PR. https://github.com/scipy/scipy/issues/23391 provides a good overview over what has been done and what's still ongoing / planned. 

Note that this includes some functionalities enabled by the generic xp backend like broadcasting support that extend beyond mirroring the original capabilities.

PR #22776, opened 2025-04-01
DOC: Refine the documentation regarding the switching of COBYLA to PRIMA

zaikunzhang:
#### What does this implement/fix?

This PR refines the documentation regarding the switching of COBYLA to the PRIMA implementation. 

- d7a4a86: 

  --  Thanks to the efforts of @nbelakovski, [PyPRIMA is now merged into the main repo of PRIMA](https://fortran-lang.discourse.group/t/prima-gets-integrated-into-scipy-but-not-the-fortran-version/9480/19?u=zaikunzhang), and hence the change of the vendored code URL for PRIMA. 

  -- Clarify in `scipy/optimize/_minimize.py` that COBYLA uses the PRIMA implementation of the algorithm. 
  
  -- Update HTTP to HTTPS in the URL of PRIMA.

- 59e4a0b: Revise 1.16.0 release notes regarding the switching of COBYLA to the PRIMA implementation. 

  -- Mention that PRIMA fixes many bugs to justify the switching (in addition to other justifications).

  -- Explicitly refer to the PRIMA documentation at the end of the paragraph. 

  -- Removed the names (I particularly changed "old Powell implementation" to "old Fortran 77 implementation") to make it sound less personal.

- 04eb466: Highlight the switching in 1.16.0 release notes --- feel free to revert this if it does not qualify. 

- 37fde67: Add a link to the bug fixes made by PRIMA in case users are interested --- I suppose many of them will be. 

#### Additional information

This PR is made (thankfully) according to the [comments and suggestion on Fortran Discourse](https://fortran-lang.discourse.group/t/prima-gets-integrated-into-scipy-but-not-the-fortran-version/9480/11?u=zaikunzhang) made by @lucascolley , in particular, 

>> [ricriv](https://fortran-lang.discourse.group/u/ricriv): The release notes are also very much underselling PRIMA. A user that never heard of PRIMA before will wonder why on Earth they switched to something different. Even a simple copy of the About section in the Github repo would have been better.

> [lucascolley](https://fortran-lang.discourse.group/u/lucascolley): Please open an issue (or PR!) about this on the SciPy repo to suggest any changes!

Disclaimer:  [ricriv](https://fortran-lang.discourse.group/u/ricriv) is not me. 

Thank you for taking a look.

PR #22778, opened 2025-04-01
MAINT: switch to vendoring libprima/prima

lucascolley:
on top of and closes gh-22776


lucascolley:
@czgdp1807 do you have any idea why the `check_installation` is failing with this?

> Test for extension modules should be installed but are not found in the installation directory.

rgommers:
That `check_installation` failure is probably because the new directory isn't added as a git submodule. Git submodules are excluded automatically via `tools/get_submodule_paths.py`.


The vendoring script in this PR looks like it's going in the right direction, but can you exclude more files please? E.g., the many files in `pyprima/.github`, and the `profiles_vs_matlib`.

lucascolley:
> Git submodules are excluded automatically via tools/get_submodule_paths.py

Thanks, but I have already adjusted for that in 598f4d0b09ac2f5ec38f5a1511bce55756beea20. That fixed an error from `check_installation` about test modules not being installed, but not this error which seems specific to extension modules.

> The vendoring script in this PR looks like it's going in the right direction, but can you exclude more files please? E.g., the many files in pyprima/.github, and the profiles_vs_matlib.

Sure, can do. I had originally cherry-picked only the directories we needed, but @nbelakovski was concerned about reaching too deep into the repo, so it was suggested to switch to this setup in https://github.com/nbelakovski/scipy/pull/1#issuecomment-2733394211. We can either extend the list of removals a bit, or revert to cherry-picking what is necessary.

tools/get_submodule_paths.py
```diff
@@ -14,6 +14,8 @@ def get_submodule_paths():
                         datum.startswith('\tpath = ')]
         submodule_paths = [os.path.join(root_directory, path) for path in
                            submodule_paths]
+    # vendored with a script rather than via gitmodules
+    submodule_paths += os.path.join(root_directory, 'scipy/_lib/pyprima')
```
tylerjereddy, line 18: `check_installation.py` will pass with a small shim here:

```diff
--- a/tools/get_submodule_paths.py
+++ b/tools/get_submodule_paths.py
@@ -15,7 +15,7 @@ def get_submodule_paths():
         submodule_paths = [os.path.join(root_directory, path) for path in
                            submodule_paths]
     # vendored with a script rather than via gitmodules
-    submodule_paths += os.path.join(root_directory, 'scipy/_lib/pyprima')
+    submodule_paths += [os.path.join(root_directory, 'scipy/_lib/pyprima')]
```

The problem was that every single character in the `pyprima` path string got appended to the list separately, and then various shenanigans ensued.

lucascolley: 🤦‍♂️ thanks Tyler, rookie error!

lucascolley:
> can you exclude more files please? E.g., the many files in pyprima/.github, and the profiles_vs_matlib.

done, please let me know if any more changes are desired

tools/vendor_pyprima.sh
```diff
@@ -6,7 +6,7 @@ set -o nounset
 set -o errexit
 
 REPO_URL="https://github.com/libprima/prima"
-COMMIT_HASH="5a128beb5f29155adfc4dd302a6dcb3f9e885aea"
+COMMIT_HASH="e7083b9eccbc74d26d13a38ca395e5df6062ab8b"
```
zaikunzhang, line 9: Hello Lucas @lucascolley , could you update this to commit 6fbf15d9b4aac20d2e3ee5b3fcb41f0ca17c4330 ? I made it just now, fixing some grammar mistakes and sentences with ambiguity in the documentation of PyPRIMA. See

https://github.com/libprima/prima/commit/6fbf15d9b4aac20d2e3ee5b3fcb41f0ca17c4330

Thank you very much. 

Zaikun

tools/vendor_pyprima.sh
```diff
@@ -0,0 +1,29 @@
...
+git checkout $COMMIT_HASH
+rm -rf .development
+rm -rf .git
+rm .gitmodules
+rm -rf benchmark/
+rm -rf fortran/
+rm -rf matlab/
+rm -rf c/
+rm -rf python/
+rm -rf .github/
```
j-bowhay, line 27: I think we should do something like this, it removes all but the pyprima directory and we don't have to worry so much about upstream reorganization breaking our script 
```suggestion
rm -rf $(ls -A | grep -v pyprima)
```

lucascolley: I committed something slightly different in https://github.com/scipy/scipy/pull/22778/commits/22b64cff104241464b76ed8c9957e3fe557bb25c, let me know what you think

j-bowhay (approved):
LGTM thanks! Failure looks unrelated

lucascolley:
thanks Jake!

rgommers:
> @rgommers would it be appropriate to mark https://github.com/scipy/pyprima as 'public archive' now?

Yep, done now.

PR #22782, opened 2025-04-02
MAINT: fix failures in free-threading(parallel=1) job

rgommers:
xref gh-22758. Fixes two of the failures tracked in that issue. See individual commit messages for more details.

### DIFF
--- a/scipy/sparse/linalg/_dsolve/tests/test_linsolve.py
+++ b/scipy/sparse/linalg/_dsolve/tests/test_linsolve.py
@@ -4,7 +4,6 @@ import threading
 import numpy as np
 from numpy import array, finfo, arange, eye, all, unique, ones, dot
 from numpy.exceptions import ComplexWarning
-import numpy.random as random
 from numpy.testing import (
         assert_array_almost_equal, assert_almost_equal,
         assert_equal, assert_array_equal, assert_, assert_allclose,
@@ -61,7 +60,6 @@ class TestFactorized:
         d = arange(n) + 1
         self.n = n
         self.A = dia_array(((d, 2*d, d[::-1]), (-3, 0, 5)), shape=(n,n)).tocsc()
-        random.seed(1234)
 
     def _check_singular(self):
         A = csc_array((5,5), dtype='d')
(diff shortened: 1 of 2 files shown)


mdhaber (commented):
I would have pushed a commit adding
```python3
os.path.join('sparse', 'linalg', '_dsolve', 'tests', 'test_linsolve.py'),  # gh-22782
```
to `allowed_filters` in `_lib/tests/test_warnings.py` to fix CI, but I wonder why this changed away from the approach recommended by that test. Is:

> warning ignore filter should not be used, instead, use numpy.testing.suppress_warnings (in tests only);

wrong now?

The rest looks OK.

rgommers:
Ah good point Matt. That test is still correct, however the code comment in it isn't - we should indeed no longer be using `suppress_warnings`. It's not urgent to move away from it, but there's no benefit from it anymore. Plus (a) in CPython 3.14 the `warnings` module will actually be better, and (b) currently in the free-threaded job we get a wall of warnings from `suppress_warnings`, which is the reason I'm moving away from it here when touching this code anyway ([example log](https://github.com/scipy/scipy/actions/runs/14206158268/job/39804289305)).

Not sure why I didn't see those failures locally, let me check how to best handle that. I see a lot of tests use `warnings.simplefilter`, maybe I should have used that instead of `warnings.filterwarnings`.

rgommers:
This rabbit hole with warning filters was a little deeper than I had anticipated, so I took the changes out of this PR - will try making a separate update in one go, including documenting the current state of how to do this better in `test_warnings.py`.

> It would be helpful to know why this test has been against warning suppression in code and what the options are when maintainers decide that suppressing the warning is desirable (including in source code, outside of tests).

The historical reason is given in the `suppress_warnings` docstring: adding and then removing a filter again would silence any further such warnings, due to a bug in the stdlib. That was fixed in Python 3.4 though. What remains is thread-safety related. This will be fixed in Python 3.14 at long last (barring unforeseen last hiccups).

tylerjereddy (commented):
The sparse linalg-related diff looks solid. I wasn't able to reproduce the failure on x86_64 Linux with `3.13t` and GIL disabled on `main`, and the following incantation passed 3 times in a row on both `main` and this branch:

`python dev.py test -t scipy/sparse/linalg/_dsolve/tests/test_linsolve.py -- --parallel-threads=40 --iterations=40`

That should be "ok," since the diff looks like an improvement.

I didn't check the `stats` changes, but Matt is already reviewing here.

tylerjereddy:
Looks like Matt said:

> The rest looks OK.

above, and Ralf mentioned backing out the warnings stuff that was the only real source of discussion, so I'll put it in.

PR #22783, opened 2025-04-02
TST/MAINT: signal.symiirorder2: r, omega, precision are floats; use default dtypes

crusaderky:
- The parameters r, omega, and precision of `scipy.signal.symiirorder2` are floats, as per the documentation.
This PR undoes an improper wrapping in `xp.asarray` in the test.
This fixes a regression vs. array-api-strict git tip, which has started crashing on `__mul__` and other ops vs. np.float64 types, on the line `a2 = 2 * r * np.cos(omega)` because r was an `array_api_strict.Array`.
- When the torch default dtype is float32, `symiirorder1` and `symiirorder2` now return float32 when fed with int inputs, accordingly to the general `xp_promote` policy.
- Re-enabled several tests on CuPy
- Cosmetic enhancement of the code.

### DIFF
--- a/scipy/signal/_spline_filters.py
+++ b/scipy/signal/_spline_filters.py
@@ -5,7 +5,7 @@ from numpy import (zeros_like, array, tan, arange, floor,
                    moveaxis, abs, complex64, float32)
 import numpy as np
 
-from scipy._lib._array_api import array_namespace
+from scipy._lib._array_api import array_namespace, xp_promote
 
 from scipy._lib._util import normalize_axis_index
 
@@ -797,10 +797,9 @@ def symiirorder2(input, r, omega, precision=-1.0):
         The filtered signal.
     """
     xp = array_namespace(input)
-
-    # internals are numpy-only
-    input = np.asarray(input)
-    omega = np.asarray(omega)
+    input = xp_promote(input, force_floating=True, xp=xp)
+    # This function uses C internals
+    input = np.ascontiguousarray(input)
 
     if r >= 1.0:
         raise ValueError('r must be less than 1.0')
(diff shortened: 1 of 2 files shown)


scipy/signal/_spline_filters.py
```diff
@@ -797,41 +797,34 @@ def symiirorder2(input, r, omega, precision=-1.0):
...
     if input.ndim == 1:
         input = input[None, :]
         squeeze_dim = True
 
-    if np.issubdtype(input.dtype, np.integer):
-        input = input.astype(np.promote_types(input.dtype, np.float32))
-
     rsq = r * r
-    a2 = 2 * r * np.cos(omega)
+    a2 = 2 * r * math.cos(omega)
```
ev-br, line 816: Despite what the docs say, I wouldn't be surprised if historic usage allows 0D arrays or `np.float64`s. This part of the codebase is straight from the multipack era. Let's see if tests complain.

crusaderky:
I've slightly extended the scope of this PR; see updated initial post

scipy/signal/tests/test_splines.py
```diff
@@ -367,13 +366,15 @@ def test_symiir2(self, dtype, precision, xp):
...
-    @skip_xp_backends(cpu_only=True)
+    @skip_xp_backends(cpu_only=True, exceptions=["cupy"], reason="C internals")
     @pytest.mark.parametrize('dtyp', ['float32', 'float64'])
     def test_symiir2_values(self, dtyp, xp):
         rng = np.random.RandomState(1234)
         s = rng.uniform(size=16).astype(dtyp)
         s = xp.asarray(s)
-        dtyp = getattr(xp, dtyp)
+
+        # cupy returns f64 for f32 inputs
```
ev-br, line 376: This one is a bit funny: I'm pretty sure CuPy's version was made to follow SciPy's, and now we're adjusting SciPy's to match CuPy's :-).

ev-br (approved):
LGTM. Changes look reasonable, the resulting code is cleaner, and the test coverage has expanded. Looks like wins all around. Thanks Guido!

PR #22786, opened 2025-04-02
BUG: illustrate milp crash

andyfaff:
This change illustrates an interpreter crash that occurs in CI, but seemingly only on macOS.

All the PR does is change how random number generation occurs in `optimize.tests.test_linprog::magic_square`. The change causes `scipy.optimize.tests.test_milp::test_result` a hard crash, specifically [on this line](https://github.com/scipy/scipy/blob/main/scipy/optimize/tests/test_milp.py#L100).

Weirdly, if I copy and paste the `magic_square` function and those two test lines into an interpreter everything works fine, crashes only occur when the test suite is run.

### DIFF
--- a/scipy/optimize/tests/test_linprog.py
+++ b/scipy/optimize/tests/test_linprog.py
@@ -78,14 +78,14 @@ def _assert_success(res, desired_fun=None, desired_x=None,
                         rtol=rtol, atol=atol)
 
 
-def magic_square(n):
+def magic_square(n, rng=11322891):
     """
     Generates a linear program for which integer solutions represent an
     n x n magic square; binary decision variables represent the presence
     (or absence) of an integer 1 to n^2 in each position of the square.
     """
 
-    rng = np.random.RandomState(0)
+    rng = np.random.default_rng(rng)
     M = n * (n**2 + 1) / 2
 
     numbers = np.arange(n**4) // n**2 + 1
@@ -139,7 +139,7 @@ def magic_square(n):
 
     A = np.array(np.vstack(A_list), dtype=float)
     b = np.array(b_list, dtype=float)
-    c = rng.rand(A.shape[1])
+    c = rng.random(A.shape[1])
 
     return A, b, c, numbers, M
 


andyfaff:
@mdhaber

mdhaber:
Sounds like you know, but just thought I'd confirm: this test should not be sensitive to the seed. It is really a feasibility problem (generate _any_ [magic square](https://en.wikipedia.org/wiki/Magic_square) of specified size); the objective function is arbitrary. It might affect the particular result, but the problem will still be feasible for any objective function.

andyfaff:
For reference, https://github.com/scipy/scipy/actions/runs/14230868239/job/39881097083?pr=22786#step:4:2284

ilayn:
I don't claim that this is the case but typically the crash can be coming from an earlier test but nuking when a new interaction is done. The easiest to see whether you have the same issue or not is, after the build, recreate the test and run `%timeit` on the solver step so that the function is invoked numerous times to flesh out the leaks and buffer overflows. For compiled extensions it is quite normal to die in the third invocation etc. I don't know what plays a role in ipython kernel death. 

Also assuming you have the same issue, then you need to run all highs tests with `-v -v` flag and without any workers to catch exactly where it goes bad. This happens to me very often when I am testing the new C translations. I can't remember in which module it was but I spent a full week on the wrong test and lost my shit because of this because `pytest` was printing a few more test names before dying.

andyfaff:
Using `python dev.py test -t scipy.optimize.tests.test_milp::test_result`, i.e. the single test, is enough to trigger the crash. So there's no interaction with any other tests.
I made sure that I was rebased on main, had done a `git clean -xdf`, `git submodule update --init`, etc. I'm building with Accelerate.

However, the issue only comes up if I run that command. If I extract the relevant bits and run the same code in an interpreter no crash occurs.

> Equivalently, maybe check out an old commit and cherry-pick this one

So go back to a version before Highs was updated and apply this patch and see if things work?

mdhaber:
> So go back to a version before Highs was updated

I think that would be a shortcut to help debug. If it happens before and after, it seems more likely to me that the problem is due to HiGHS, since I doubt the two entirely different approaches to wrapping would produce similar failures. If it only happens after, it *could* be a problem with the new wrapper, or it could be a regression in HiGHS.

andyfaff:
I looked at 33335695aef244321260b149ec3c1286b7fde4d6 and 1.14.1. The former is bad, the latter is good. For 1.14.1 I had to install the wheel because I couldn't build for some reason. With both I patched the `test_milp.py` file with these changes.

andyfaff:
I can't really bisect against older versions because I run up against a 
```
E   ImportError: dlopen(/Users/andrew/Documents/Andy/programming/scipy/build-install/lib/python3.11/site-packages/scipy/special/_ufuncs.cpython-311-darwin.so, 0x0002): Library not loaded: /Users/andrew/Documents/Andy/programming/scipy/build-install/lib/python3.11/site-packages/scipy/special/libsf_error_state.dylib
E     Referenced from: <2A5B9E16-3F55-358F-B46B-15998F3FE836> /Users/andrew/Documents/Andy/programming/scipy/build-install/lib/python3.11/site-packages/scipy/special/_ufuncs.cpython-311-darwin.so
E     Reason: tried: '/Users/andrew/Documents/Andy/programming/scipy/build-install/lib/python3.11/site-packages/scipy/special/libsf_error_state.dylib' (missing LC_LOAD_DYLIB (must link with at least libSystem.dylib)), '/System/Volumes/Preboot/Cryptexes/OS/Users/andrew/Documents/Andy/programming/scipy/build-install/lib/python3.11/site-packages/scipy/special/libsf_error_state.dylib' (no such file), '/Users/andrew/Documents/Andy/programming/scipy/build-install/lib/python3.11/site-packages/scipy/special/libsf_error_state.dylib' (missing LC_LOAD_DYLIB (must link with at least libSystem.dylib))
``` 
error

mdhaber:
Do you need to? It looks like the 14.1 version is before the change in wrapper/HiGHS version change?

mdhaber:
I wanted to try to help bisect, but I can't reproduce the crash in the development environment on my Mac.

<details>

```
>>> import sys, scipy, numpy; print(scipy.__version__, numpy.__version__, sys.version_info); \
scipy.show_config()
1.16.0.dev0+git20250403.82b6ddb 2.1.2 sys.version_info(major=3, minor=13, micro=0, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    detection method: pkgconfig
    found: true
    include directory: /Users/matthaberland/miniforge3/envs/scipy-dev/include
    lib directory: /Users/matthaberland/miniforge3/envs/scipy-dev/lib
    name: openblas
    openblas configuration: USE_64BITINT=0 DYNAMIC_ARCH=0 DYNAMIC_OLDER= NO_CBLAS=
      NO_LAPACK=0 NO_LAPACKE= NO_AFFINITY=1 USE_OPENMP=1 VORTEX MAX_THREADS=128
    pc file directory: /Users/matthaberland/miniforge3/envs/scipy-dev/lib/pkgconfig
    version: 0.3.28
  lapack:
    detection method: pkgconfig
    found: true
    include directory: /Users/matthaberland/miniforge3/envs/scipy-dev/include
    lib directory: /Users/matthaberland/miniforge3/envs/scipy-dev/lib
    name: openblas
    openblas configuration: USE_64BITINT=0 DYNAMIC_ARCH=0 DYNAMIC_OLDER= NO_CBLAS=
      NO_LAPACK=0 NO_LAPACKE= NO_AFFINITY=1 USE_OPENMP=1 VORTEX MAX_THREADS=128
    pc file directory: /Users/matthaberland/miniforge3/envs/scipy-dev/lib/pkgconfig
    version: 0.3.28
  pybind11:
    detection method: pkgconfig
    include directory: /Users/matthaberland/miniforge3/envs/scipy-dev/include
    name: pybind11
    version: 2.13.6
Compilers:
  c:
    args: -ftree-vectorize, -fPIC, -fstack-protector-strong, -O2, -pipe, -isystem,
      /Users/matthaberland/miniforge3/envs/scipy-dev/include, -D_FORTIFY_SOURCE=2,
      -isystem, /Users/matthaberland/miniforge3/envs/scipy-dev/include
    commands: arm64-apple-darwin20.0.0-clang
    linker: ld64
    linker args: -Wl,-headerpad_max_install_names, -Wl,-dead_strip_dylibs, -Wl,-rpath,/Users/matthaberland/miniforge3/envs/scipy-dev/lib,
      -L/Users/matthaberland/miniforge3/envs/scipy-dev/lib, -ftree-vectorize, -fPIC,
      -fstack-protector-strong, -O2, -pipe, -isystem, /Users/matthaberland/miniforge3/envs/scipy-dev/include,
      -D_FORTIFY_SOURCE=2, -isystem, /Users/matthaberland/miniforge3/envs/scipy-dev/include
    name: clang
    version: 17.0.6
  c++:
    args: -ftree-vectorize, -fPIC, -fstack-protector-strong, -O2, -pipe, -stdlib=libc++,
      -fvisibility-inlines-hidden, -fmessage-length=0, -isystem, /Users/matthaberland/miniforge3/envs/scipy-dev/include,
      -D_FORTIFY_SOURCE=2, -isystem, /Users/matthaberland/miniforge3/envs/scipy-dev/include
    commands: arm64-apple-darwin20.0.0-clang++
    linker: ld64
    linker args: -Wl,-headerpad_max_install_names, -Wl,-dead_strip_dylibs, -Wl,-rpath,/Users/matthaberland/miniforge3/envs/scipy-dev/lib,
      -L/Users/matthaberland/miniforge3/envs/scipy-dev/lib, -ftree-vectorize, -fPIC,
      -fstack-protector-strong, -O2, -pipe, -stdlib=libc++, -fvisibility-inlines-hidden,
      -fmessage-length=0, -isystem, /Users/matthaberland/miniforge3/envs/scipy-dev/include,
      -D_FORTIFY_SOURCE=2, -isystem, /Users/matthaberland/miniforge3/envs/scipy-dev/include
    name: clang
    version: 17.0.6
  cython:
    commands: cython
    linker: cython
    name: cython
    version: 3.0.11
  fortran:
    args: -march=armv8.3-a, -ftree-vectorize, -fPIC, -fno-stack-protector, -O2, -pipe,
      -isystem, /Users/matthaberland/miniforge3/envs/scipy-dev/include
    commands: /Users/matthaberland/miniforge3/envs/scipy-dev/bin/arm64-apple-darwin20.0.0-gfortran
    linker: ld64
    linker args: -Wl,-headerpad_max_install_names, -Wl,-dead_strip_dylibs, -Wl,-rpath,/Users/matthaberland/miniforge3/envs/scipy-dev/lib,
      -L/Users/matthaberland/miniforge3/envs/scipy-dev/lib, -march=armv8.3-a, -ftree-vectorize,
      -fPIC, -fno-stack-protector, -O2, -pipe, -isystem, /Users/matthaberland/miniforge3/envs/scipy-dev/include
    name: gcc
    version: 13.2.0
  pythran:
    include directory: ../../../miniforge3/envs/scipy-dev/lib/python3.13/site-packages/pythran
    version: 0.16.1
Machine Information:
  build:
    cpu: aarch64
    endian: little
    family: aarch64
    system: darwin
  cross-compiled: false
  host:
    cpu: aarch64
    endian: little
    family: aarch64
    system: darwin
Python Information:
  path: /Users/matthaberland/miniforge3/envs/scipy-dev/bin/python
  version: '3.13'
```

</details>

ilayn:
It is simpler to setup adress sanitizer and gdb to get the reason of the crash

andyfaff:
@mdhaber what happens if you use Accelerate? I tried building with the `asan` option, but there's a link error.
```
ld: library 'asan' not found
collect2: error: ld returned 1 exit status
```

mdhaber:
I'm looking for information about how to do that, and search engines are turning up results about using `site.cfg`. That's out of date, right? I use an in-place build, in case it matters.

lucascolley:
you use `pip install -e .` right? can you try `pip -Csetup-args=-Dblas=accelerate install -e .`?

mdhaber:
Thanks, I follow meson's edtiable install command, but I was able to add that option. I still couldn't replicate the failure, though.

PR #15641, opened 2022-02-23
ENH: scipy.optimize.minimize with method=SLSQP returns KKT multipliers TAKE 2

andyfaff:
@msparapa this is an updated version of your PR #9839 to return the KKT multipliers from SLSQP. (To save git hassles).

For reviewers:

1. This needs tests. This PR naively returns the values that SLSQP said should be returned (https://github.com/scipy/scipy/blob/main/scipy/optimize/slsqp/slsqp_optmz.f#L154). EDIT: I've added some simple tests.
2. A dict containing the KKT multipliers is returned. The `'eq'` key contains a list of the multipliers for the equality constraints, in the order in which they appear in the `constraints` list. A similar approach is used for the inequality constraints. 
3. **HOWEVER**, if new-style `NonlinearConstraint` or `LinearConstraint` are used, then it's possible for a single new-style constraint to simultaneously contain both inequality and equality constraints. The first step in SLSQP is to separate these mixed new-style constraints into separated old-style constraints. This means that if there is mixing within a single constraint, then the returned list of multipliers is going to appear to be of a different length to the original new-style constraints. Hopefully this doesn't lead to confusion.
4. The original PR said that it seems to be necessary to call `slsqp` again to force the multipliers to be updated. This won't cause any more function evaluations, but it will add extra overhead. I don't particularly want to dive into the fortran code to see if this can be 'fixed'.

PR #22787, opened 2025-04-02
DOC: optimize: Add the multiplier details to SLSQP funcs

### LINKED ISSUE #14394: ENH: optmize.slsqp: return Lagrange multipliers

### LINKED ISSUE #22769: ENH: optimize: Return bound multiplier for SLSQP
Now that we have #22524 in, we have the multipliers are being returned by SLSQP. But these multipliers are only corresponding to the problem constraints. The result is now reported in the `OptimizeResult` object under the `multipliers` key (the name we can change before the release). 

However, the multipliers corresponding to the bounds on the variables were being NaN'd and never survived the algorithm. Now that we have a grip on the code, we can surface those back to python as requested in #14394 

There are a few difficulties about this operation I'd like to describe below; 

When the user bounds are processed by SLSQP, for some strange reason they are handled together with the problem's internal dummy variables for inconsistent linearization cases. Say we have three decision variables, and we lower bound x1 x2, with values L1, L2, and upper bound x2, x3. with U2, U3. After some black magic these constraints converted into inequality constraints with the following 

```
[ prob ]               [ ]
[ ineqs ]  [x1]        [ ]
[1  0  0]  [x2]  >=    [ L1]
[0  1  0]  [x3]        [ L2]
[0 -1  0]              [-U2]
[0  0 -1]              [-U3]
```

But SLSQP code sometimes augments the problem and uses its own dummy variable [0, 1] to regularize the problem so the problem becomes 

```
[ prob     ]               [   ]
[ ineqs    ]  [x1 ]        [   ]
[1  0  0  0]  [x2 ]  >=    [ L1]
[0  1  0  0]  [x3 ]        [ L2]
[0  0  0  1]  [dum]        [0  ]
[0 -1  0  0]               [-U2]
[0  0 -1  0]               [-U3]
[0  0  0 -1]               [-1 ] 

```

because pain is addictive. Hence it squeezes two more constraints. Of course, this does not happen every time. Only when problem encounters these linearization issues. 

So while the bounds vectors are [L1, L2, NaN] and [NaN, U2, U3], we have a multiplier vector coming in either 

```
[ineq mults ... m1, m2, m3, m4, NaN, NaN]
                ^   ^   ^   ^   
                L1  L2  U2  U3 
```

```
[ineq mults ... m1, m2, m3, m4, m5, m6]
                ^   ^   ^   ^   ^   ^
                L1  L2  dum U2  U3  dum
```

So then, in C code when we detect inconsistent linearizations, we NaN the entry m3 to mark that it is not U2 but there was a dummy variable. If there are no dummy vars then all finite values correspond to multipliers and keep mapping finite bounds to finite multipliers. Otherwise this is kind of a pigeonholing bunch of numbers and hoping for the best. I don't know if there is any other mechanism that can create NaNs in the multiplier vector through bad math.

It's ugly but not impossible, not sure if this will work out with large problems though. What do you think @mdhaber  @andyfaff ?

ilayn:
#### Reference issue
Closes #22769 
Closes #14394 
Closes #15641 

Adds more information about what is returned under the `OptimizeResult` regarding the new `multipliers` key. Also brought the text from #15641 here to the `minimize` tutorial.

### DIFF
--- a/scipy/optimize/_minimize.py
+++ b/scipy/optimize/_minimize.py
@@ -554,7 +554,7 @@ def minimize(fun, x0, args=(), method=None, jac=None, hess=None,
     Karush-Kuhn-Tucker (KKT) multipliers, which are a generalization
     of Lagrange multipliers to inequality-constrained optimization problems.
 
-    Notice that at solution, the first constraint is active, let's evaluate the
+    Notice that at the solution, the first constraint is active. Let's evaluate the
     function at solution:
 
     >>> cons[0]['fun'](res.x)
--- a/scipy/optimize/_minimize.py
+++ b/scipy/optimize/_minimize.py
@@ -529,7 +529,6 @@ def minimize(fun, x0, args=(), method=None, jac=None, hess=None,
         [ 0.09495377,  0.18996269,  0.38165151,  0.7664427,   1.53713523]
     ])
 
-
     Next, consider a minimization problem with several constraints (namely
     Example 16.4 from [5]_). The objective function is:
 
@@ -547,10 +546,42 @@ def minimize(fun, x0, args=(), method=None, jac=None, hess=None,
 
     The optimization problem is solved using the SLSQP method as:
 
-    >>> res = minimize(fun, (2, 0), method='SLSQP', bounds=bnds,
-    ...                constraints=cons)
+    >>> res = minimize(fun, (2, 0), method='SLSQP', bounds=bnds, constraints=cons)
 
-    It should converge to the theoretical solution (1.4 ,1.7).
+    It should converge to the theoretical solution [1.4 ,1.7]. *SLSQP* also
+    returns the multipliers that are used in the solution of the problem. These
+    multipliers, when the problem constraints are linear can be thought as the
+    Karush-Kuhn-Tucker (KKT) multipliers, which are a generalization
+    of Lagrange multipliers to inequality-constrained optimization problems.
+
+    Notice that at solution, the first constraint is active, let's evaluate the
+    function at solution:
+
+    >>> cons[0]['fun'](res.x)
+    np.float64(1.4901224698604665e-09)
+
+    Also notice that at optimality there is a non-zero multiplier:
+
+    >>> res.multipliers
+    array([0.8, 0. , 0. ])
+
+    This can be understood as the local sensitivity of the optimal value of the
+    objective function with respect to changes in the first constraint. If we
+    tighten the constraint by a small amount ``eps``:
+
+    >>> eps = 0.01
+    ... cons[0]['fun'] = lambda x: x[0] - 2 * x[1] + 2 - eps
+
+    we expect the optimal value of the objective function to increase by
+    approximately ``eps * res.multipliers[0]``:
+
+    >>> eps * res.multipliers[0]  # Expected change in f0
+    np.float64(0.008000000027153205)
+    >>> f0 = res.fun  # Keep track of the previous optimal value
+    ... res = minimize(fun, (2, 0), method='SLSQP', bounds=bnds, constraints=cons)
+    ... f1 = res.fun  # New optimal value
+    ... f1 - f0
+    np.float64(0.008019998807885509)
 
     """
     x0 = np.atleast_1d(np.asarray(x0))
--- a/scipy/optimize/_minimize.py
+++ b/scipy/optimize/_minimize.py
@@ -570,7 +570,7 @@ def minimize(fun, x0, args=(), method=None, jac=None, hess=None,
     tighten the constraint by a small amount ``eps``:
 
     >>> eps = 0.01
-    ... cons[0]['fun'] = lambda x: x[0] - 2 * x[1] + 2 - eps
+    >>> cons[0]['fun'] = lambda x: x[0] - 2 * x[1] + 2 - eps
 
     we expect the optimal value of the objective function to increase by
     approximately ``eps * res.multipliers[0]``:
@@ -578,9 +578,9 @@ def minimize(fun, x0, args=(), method=None, jac=None, hess=None,
     >>> eps * res.multipliers[0]  # Expected change in f0
     np.float64(0.008000000027153205)
     >>> f0 = res.fun  # Keep track of the previous optimal value
-    ... res = minimize(fun, (2, 0), method='SLSQP', bounds=bnds, constraints=cons)
-    ... f1 = res.fun  # New optimal value
-    ... f1 - f0
+    >>> res = minimize(fun, (2, 0), method='SLSQP', bounds=bnds, constraints=cons)
+    >>> f1 = res.fun  # New optimal value
+    >>> f1 - f0
     np.float64(0.008019998807885509)
 
     """
--- a/scipy/optimize/_minimize.py
+++ b/scipy/optimize/_minimize.py
@@ -548,9 +548,9 @@ def minimize(fun, x0, args=(), method=None, jac=None, hess=None,
 
     >>> res = minimize(fun, (2, 0), method='SLSQP', bounds=bnds, constraints=cons)
 
-    It should converge to the theoretical solution [1.4 ,1.7]. *SLSQP* also
+    It should converge to the theoretical solution ``[1.4 ,1.7]``. *SLSQP* also
     returns the multipliers that are used in the solution of the problem. These
-    multipliers, when the problem constraints are linear can be thought as the
+    multipliers, when the problem constraints are linear, can be thought of as the
     Karush-Kuhn-Tucker (KKT) multipliers, which are a generalization
     of Lagrange multipliers to inequality-constrained optimization problems.
 
@@ -560,7 +560,7 @@ def minimize(fun, x0, args=(), method=None, jac=None, hess=None,
     >>> cons[0]['fun'](res.x)
     np.float64(1.4901224698604665e-09)
 
-    Also notice that at optimality there is a non-zero multiplier:
+    Also, notice that at optimality there is a non-zero multiplier:
 
     >>> res.multipliers
     array([0.8, 0. , 0. ])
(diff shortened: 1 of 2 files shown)


ilayn:
Any comments for this one?

scipy/optimize/_minimize.py
```diff
@@ -543,10 +542,42 @@ def minimize(fun, x0, args=(), method=None, jac=None, hess=None,
...
-    ...                constraints=cons)
+    >>> res = minimize(fun, (2, 0), method='SLSQP', bounds=bnds, constraints=cons)
+
+    It should converge to the theoretical solution [1.4 ,1.7]. *SLSQP* also
+    returns the multipliers that are used in the solution of the problem. These
+    multipliers, when the problem constraints are linear can be thought as the
+    Karush-Kuhn-Tucker (KKT) multipliers, which are a generalization
+    of Lagrange multipliers to inequality-constrained optimization problems.
+
+    Notice that at solution, the first constraint is active, let's evaluate the
```
j-bowhay, line 553:
```suggestion
    Notice that at the solution, the first constraint is active. Let's evaluate the
```

scipy/optimize/_minimize.py
```diff
@@ -543,10 +542,42 @@ def minimize(fun, x0, args=(), method=None, jac=None, hess=None,
 
     The optimization problem is solved using the SLSQP method as:
 
-    >>> res = minimize(fun, (2, 0), method='SLSQP', bounds=bnds,
-    ...                constraints=cons)
+    >>> res = minimize(fun, (2, 0), method='SLSQP', bounds=bnds, constraints=cons)
+
+    It should converge to the theoretical solution [1.4 ,1.7]. *SLSQP* also
```
j-bowhay, line 547:
```suggestion
    It should converge to the theoretical solution ``[1.4 ,1.7]``. *SLSQP* also
```

scipy/optimize/_minimize.py
```diff
@@ -543,10 +542,42 @@ def minimize(fun, x0, args=(), method=None, jac=None, hess=None,
...
+    returns the multipliers that are used in the solution of the problem. These
+    multipliers, when the problem constraints are linear can be thought as the
```
j-bowhay, line 549:
```suggestion
    multipliers, when the problem constraints are linear, can be thought of as the
```

scipy/optimize/_minimize.py
```diff
@@ -543,10 +542,42 @@ def minimize(fun, x0, args=(), method=None, jac=None, hess=None,
...
+    Karush-Kuhn-Tucker (KKT) multipliers, which are a generalization
+    of Lagrange multipliers to inequality-constrained optimization problems.
```
j-bowhay, line 551: Optional, but it might be nice to add a reference here. It's the sort of thing numerical people know about but maybe scientists/engineers don't!

scipy/optimize/_minimize.py
```diff
@@ -543,10 +542,42 @@ def minimize(fun, x0, args=(), method=None, jac=None, hess=None,
...
+
+    Notice that at solution, the first constraint is active, let's evaluate the
+    function at solution:
+
+    >>> cons[0]['fun'](res.x)
+    np.float64(1.4901224698604665e-09)
+
+    Also notice that at optimality there is a non-zero multiplier:
```
j-bowhay, line 559:
```suggestion
    Also, notice that at optimality there is a non-zero multiplier:
```

scipy/optimize/_slsqp_py.py
```diff
@@ -250,6 +250,45 @@ def _minimize_slsqp(func, x0, args=(), jac=None, bounds=None,
 
         .. versionadded:: 1.16.0
 
+    Returns
+    -------
+    res : OptimizeResult
+        The optimization result represented as an ``OptimizeResult`` object.
```
j-bowhay, line 256:
```suggestion
        The optimization result represented as an `OptimizeResult` object.
```

scipy/optimize/_slsqp_py.py
```diff
@@ -250,6 +250,45 @@ def _minimize_slsqp(func, x0, args=(), jac=None, bounds=None,
...
+        termination, and ``multipliers`` which contains the Karush-Kuhn-Tucker
+        (KKT) multipliers for the QP approximation used in solving the original
+        nonlinear problem. See ``Notes`` below. See also `OptimizeResult` for a
+        description of other attributes.
+
+    Notes
+    -----
+    The KKT multipliers are returned in the ``OptimizeResult.multipliers``
+    attribute as a NumPy array. Denoting the dimension of the equality constraints
+    with ``meq``, and of inequality constraints with ``mineq``, then the returned
```
j-bowhay, line 269: nit ``mineq`` is never referenced, do we need to introduce it?

j-bowhay (changes requested):
A few typos but otherwise all good I think

scipy/optimize/_slsqp_py.py
ilayn: Insufficient coffee 🤦‍♂️

scipy/optimize/_slsqp_py.py
```diff
@@ -250,6 +250,46 @@ def _minimize_slsqp(func, x0, args=(), jac=None, bounds=None,
...
+    -------
+    res : OptimizeResult
+        The optimization result represented as an `OptimizeResult` object.
+        In this dict-like object the following fields are of particular importance:
+        ``x`` the solution array, ``success`` a Boolean flag indicating if the
+        optimizer exited successfully, ``message`` which describes the reason for
+        termination, and ``multipliers`` which contains the Karush-Kuhn-Tucker
+        (KKT) multipliers for the QP approximation used in solving the original
+        nonlinear problem. See ``Notes`` below. See also `OptimizeResult` for a
+        description of other attributes.
```
j-bowhay, line 263: I'm not sure we should add a returns section here because of the way of injecting the options into the docs. Adding a return section makes it render a bit funny and is duplicated with the `minimize` docs
image
https://output.circle-artifacts.com/output/job/015648c2-e43f-4cae-be5d-a05451719575/artifacts/0/html/reference/optimize.minimize-slsqp.html#optimize-minimize-slsqp

```suggestion
```

ilayn: That's probably something for @mdhaber to decide actually, I'm following his pattern on this one. Also if we don't include that then there is no obvious place to document these output particulars.

mdhaber: > I'm following his pattern on this one

Not really. I only contributed the tests to gh-15641. I still didn't see those here, so I decided to sit this one out.

If I approved this text in gh-15641, it was probably because the content looked good, not that it rendered well. I have no special knowledge of `minimize` method-specific documentation. I just know that none of the method-specific documentation looks great, and I see that only [`"powell"`](https://docs.scipy.org/doc/scipy/reference/optimize.minimize-powell.html) and [`"trust-constr"`](https://docs.scipy.org/doc/scipy/reference/optimize.minimize-trustconstr.html) seem to have a `Returns` section.

ilayn: I probably misread https://github.com/scipy/scipy/issues/22769#issuecomment-2773624591. Then let me reshuffle this so that `Returns` go under the argument section. Then it is a regular docstring.

ilayn:
Any more comments/concerns over this one?

j-bowhay:
It would be great if we could bring the tests over from #15641 so we can verify correctness before advertising this more loudly

ilayn:
We did not bring over all the functionality but just the docstring which was the result of #22769, that's why I just took the docstring from that.

j-bowhay:
I guess my point is, as far as I could tell, there are no tests for the new `multipliers` so maybe it is a little hasty to advertise while still untested?

ilayn:
I get what you mean so no resistance from my side. These multipliers are actually used in the computations anyways and the algorithm terminates based on its values. So if there is an issue we have bigger problems. So I'm only exposing the array not doing anything that wasn't there before. 

For the testing I don't have any good examples that are not toy examples like the one in the docstring example. If we write something that is not trivial, then we are testing the algorithm with itself, hence if you have anything nontrivial it's always good to add more.

j-bowhay:
No blockers but I unfortunately won't have time to look before branching

ilayn:
The feature is already in and the docstring would be nice to have. The discussion above is for extra tests which we can add anytime. So I'll click the button not to miss the version the feature is introduced.

PR #22792, opened 2025-04-03
BUG: signal.ShortTimeFFT.upper_border_begin: Document parameter `n`.

DietBru:
In `ShortTimeFFT.upper_border_begin` the parameter `n` is not documented, which is fixed here. Furthermore:
* Fix the same problem in `ShortTimeFFT._post_padding` and `ShortTimeFFT.k_max`
* Document return parameters in `ShortTimeFFT.upper_border_begin`.
* Test that `n` is large enough in `ShortTimeFFT.upper_border_begin` and `ShortTimeFFT._post_padding` (includes unit test).
* Fix small variable name inconsistency in `ShortTimeFFT` docstr.

See also PR #22784.

### DIFF
--- a/scipy/signal/_short_time_fft.py
+++ b/scipy/signal/_short_time_fft.py
@@ -381,14 +381,14 @@ class ShortTimeFFT:
 
     It is possible to calculate the STFT of signal parts:
 
-    >>> p_q = SFT.nearest_k_p(N // 2)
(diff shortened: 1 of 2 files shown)


j-bowhay (approved):
Thanks!

DietBru:
Thanks for the quick merge, @j-bowhay! :rocket:

PR #22718, opened 2025-03-20
TST: special: overhaul test_support_alternative_backends

crusaderky:
This is part of a set of three PRs, which can be merged in any order with minor conflicts:
- #22641
- #22639
- #22718

# In this PR
- Test all functions in support_alternative_backends vs. int64
- Revamp the hypothesis test (disabled by default)
- Add test for mismatched int and float parameters (which `scipy.stats` relies upon)
- Fix `xlogy` dtype on dask and array-api-strict

The hypothesis test makes a wealth of bugs and inconsistencies show up in various backends.
Most of these are around inf/nan behaviour but not only.
The mismatched args test also highlights several issues.

PR #21401, opened 2024-08-15
MAINT: special: Accommodate changed integer handling in NumPy 2.0.

WarrenWeckesser:
* Add the 'p' character to the typecodes recognized by the script `_generate_pyx.py`.  This code corresponds to the integer types `npy_intp` and `Py_ssize_t`.
* In Cython code:
  * Change the fused type `dl_number_t` to `dlp_number_t`, and include `Py_ssize_t` in the types it fuses.
  * Change other uses of `long` to `Py_ssize_t`.
* In the type signatures of ufuncs configured in `functions.json`, change occurrences of `i` or `l` in an input to `p`.
* Change use of `int` in the wrappers of `xsf` functions to `Py_ssize_t`. The thin wrappers cast this to `int` before calling the `xsf` function.

PR #22795, opened 2025-04-03
ENH: special: add more functions to `_support_alternative_backends`

NeilGirdhar:


NeilGirdhar:
Based on https://github.com/data-apis/array-api/issues/725#issuecomment-2775874787

mdhaber:
@crusaderky should this wait for gh-22718 or gh-22639, or would it be helpful to have more functions to test those with?

scipy/special/_support_alternative_backends.py
```diff
@@ -19,6 +20,12 @@ def get_array_special_func(f_name, xp, n_array_args):
     if is_numpy(xp):
         f = getattr(_ufuncs, f_name, None)
     elif is_torch(xp):
+        # These functions are not implemented.
+        if f_name in {'gamma', 'betaln', 'beta', 'betainc', 'expi', 'expn'}:
+            raise NotImplementedError
+        # These functions have incompatible interaces.
+        if f_name in {'logsumex'}
```
mdhaber, line 27:
```suggestion
        if f_name in {'logsumexp'}
```
But this can just be removed. `scipy.special.logsumexp` has a generic implementation, and it does not go through this machinery.

NeilGirdhar: Oh, so it will call through the other Array API functions?

mdhaber: It is implemented in terms of array API standard calls.
https://github.com/scipy/scipy/blob/25ca572f6dcaf2d7ad38d96fdc9def33b2f8a385/scipy/special/_logsumexp.py#L13

NeilGirdhar: Awesome!

scipy/special/_support_alternative_backends.py
```diff
@@ -51,22 +55,38 @@ def wrapped(*args, **kwargs):
...
+    'gammaincc': 2,
+    'betaln': 2,
+    'beta': 2,
+    'betainc': 3,
     'erf': 1,
     'erfc': 1,
+    'erfinv': 1,
+    'zeta': 2,
+    # No one implements binom yet.
+    # 'binom': 2,
```
crusaderky, line 78:
```suggestion
    'binom': 2,
```
Dask and MArray implement _all elementwise ufuncs_ when they wrap around numpy!

scipy/special/_support_alternative_backends.py
```diff
@@ -19,6 +20,9 @@ def get_array_special_func(f_name, xp, n_array_args):
     if is_numpy(xp):
         f = getattr(_ufuncs, f_name, None)
     elif is_torch(xp):
+        # These functions are not implemented.
+        if f_name in {'gamma', 'betaln', 'beta', 'betainc', 'expi', 'expn'}:
+            raise NotImplementedError
```
crusaderky, line 25:
```suggestion
```
I expect them to work on CPU through the numpy round-trip a few lines below.
On GPU, they should be skipped in the test (there is already a section for it).

crusaderky (changes requested):

crusaderky:
> @crusaderky should this wait for [gh-22718](https://github.com/scipy/scipy/pull/22718) or [gh-22639](https://github.com/scipy/scipy/pull/22639), or would it be helpful to have more functions to test those with?

merge order doesn't matter

crusaderky:
None of these changes will be picked up (and the unit tests should definitely tell you) unless you also change `scipy/special/__init__.py:787:791`.
I suggest to just change it to 
```python
from ._support_alternative_backends import *
```

scipy/special/_support_alternative_backends.py
```diff
@@ -51,22 +55,38 @@ def wrapped(*args, **kwargs):
...
+    'betainc': 3,
     'erf': 1,
     'erfc': 1,
+    'erfinv': 1,
+    'zeta': 2,
+    # No one implements binom yet.
+    # 'binom': 2,
+    'expi': 1,
+    'expn': 2,
+    'softmax': 1,
```
crusaderky, line 81: This function is not in `_ufuncs.py`. Even if you were to fix the import, it is not elementwise and cannot use this delegation system without structural changes to it. 

This delegation system can't realise that the function is already API agnostic, so it will try to convert everything to numpy before pushing data through it. On Dask, this will cause reductions to be applied locally to chunks instead of globally. On MArray, it will incorrectly calculate the mask.

I expect (hope) that the Dask test will fail  really loudly saying that the numbers coming out of it are wrong. Note that there is no test for MArray.

NeilGirdhar:
@crusaderky Thanks for your thorough and quick review!  My apologies that I'm a bit new to this code, so thanks for being patient.  I'll answer some of your comments now to try and gather some information before addressing your concerns.

NeilGirdhar:
I looked and couldn't find the justification for vendoring array-api-compat.  My issue is that the code here uses a very old version of compat and very awkward tests.  In the current version of compat, we would use:
```
    is_numpy_namespace
    is_cupy_namespace
    is_torch_namespace
    is_ndonnx_namespace
    is_dask_namespace
    is_jax_namespace
    is_pydata_sparse_namespace
    is_array_api_strict_namespace
```
It's unfortunate that we can't use these functions as they would lead to more readable code.

scipy/special/_support_alternative_backends.py
NeilGirdhar: Great!

If we expose `is_dask_namespace`, I can test for that and let `binom` through for that case.

How do I test for MArray?

scipy/special/_support_alternative_backends.py
NeilGirdhar: Do we have a way to test for PyTorch-cpu?  If not, does it makes more sense to block them on PyTorch since it would lead to the odd behavior whereby a program works just fine on PyTorch-cpu, but fails on PyTorch-gpu?

crusaderky:
```bash
python dev.py test -b torch  # on CPU float64
SCIPY_DEVICE=cuda python dev.py test -b torch  # on GPU float64
SCIPY_DEFAULT_DTYPE=float32 python dev.py test -b torch  # on CPU float32
```
CI runs all three of them.

>  a program works just fine on PyTorch-cpu, but fails on PyTorch-gpu?

We already have tons of stuff that works fine on torch/jax CPU but fails on GPU. Basically anything with a Cython or C internal call.

NeilGirdhar: Sure, but it seems that you're asking me to let these functions pass through here?  Do you want me to unconditionally let them pass through or just when PyTorch is running on the CPU?

scipy/special/_support_alternative_backends.py
crusaderky: > Great!
> 
> If we expose `is_dask_namespace`, I can test for that and let `binom` through for that case.

Please merge from main, there's news for you :wink: #22639

> How do I test for MArray?

At the moment, specific special functions that are needed by scipy.stats are indirected tested in there.
For a more systematic test effort, I would advise to talk it through thoroughly with @mdhaber and write a dedicated PR for it.

scipy/special/_support_alternative_backends.py
crusaderky: Please don't overload this module, which was never meant for anything that is not an elementwise ufunc and is not fit for purpose for it.
I'm very happy to go through a more structural redesign and unification of the delegation system, but it should be its own separate design effort and PR. This is out of scope here.

crusaderky:
> I looked and couldn't find the justification for vendoring array-api-compat. My issue is that the code here uses a very old version of compat and very awkward tests. In the current version of compat, we would use:
> 
> ```
>     is_numpy_namespace
>     is_cupy_namespace
>     is_torch_namespace
>     is_ndonnx_namespace
>     is_dask_namespace
>     is_jax_namespace
>     is_pydata_sparse_namespace
>     is_array_api_strict_namespace
> ```
> 
> It's unfortunate that we can't use these functions as they would lead to more readable code.

We're currently vendoring the version released 2 weeks ago and we can bump to git tip if needed.
Those functions are imported through `scipy._lib._array_api`, where they're renamed to `is_numpy` etc.

scipy/special/_support_alternative_backends.py
crusaderky: `binom` is not a function; it's a namespace for distributions (`binoms.ppf` etc.). This delegation system is not fit for purpose for it.

scipy/special/_support_alternative_backends.py
crusaderky: unconditionally. They will then crash and burn when `np.asarray` tries to implicitly move the GPU buffer to host memory a few lines later.

scipy/special/_support_alternative_backends.py
NeilGirdhar: Sorry, I'm confused here.  `scipy.stats.binom` is a function:
```python
[utM] In [10]: ss.binom(5, 2)
Out[10]: np.float64(10.0)
```

NeilGirdhar:
Got it.  (My mistake about it being old).  Why weren't the original function names kept?

lucascolley:
> Why weren't the original function names kept?

just a historical artifact from when we were using much less of this stuff. Feel free to send a refactor PR if that would help avoid confusion.

scipy/special/_support_alternative_backends.py
NeilGirdhar: Removed `softmax`, and `log_softmax`

scipy/special/_support_alternative_backends.py
crusaderky: Apologies, I got confused with `scipy.stats.binom`

scipy/special/tests/test_support_alternative_backends.py
```diff
@@ -56,15 +56,22 @@ def test_rel_entr_generic(dtype):
...
+                  for name, info in array_special_func_map.items()]))
 @pytest.mark.filterwarnings("ignore:invalid value encountered:RuntimeWarning:dask")
 @pytest.mark.parametrize('dtype', ['float32', 'float64'])
 @pytest.mark.parametrize('shapes', [[(0,)]*4, [tuple()]*4, [(10,)]*4,
                                     [(10,), (11, 1), (12, 1, 1), (13, 1, 1, 1)]])
 def test_support_alternative_backends(xp, f_name, n_args, dtype, shapes):
+    if f_name in {'expn', 'polygamma'}:
+        pytest.skip(f"`{f_name}` has limited domain.")
+    if f_name in {'multigammaln'}:
+        pytest.skip(f"`{f_name}` requires some integer arguments.")
```
crusaderky, line 71: It would be _very_ good to get https://github.com/scipy/scipy/pull/22718 in so that we can think how to build some tooling for these on top of it.

NeilGirdhar: Agreed.  Should I wait for that PR?

crusaderky: I think it's ok to leave these untested for the time being. Just know that the more changes you do to the test module the more either of us will have to rework to fix a conflict.

NeilGirdhar:
> Could you clarify your question here?

Sorry, I'm just frustrated with my inability to build scipy (which is making finishing this feature take longer than I anticipated).  I just thought that Jax has Jaxlib (which I assume has all the compiled code), and this partition makes Jax easier to develop since it can use ordinary setuptools (or whatever) rather than Meson.

lucascolley:
> I just thought that Jax has Jaxlib (which I assume has all the compiled code), and this partition makes Jax easier to develop since it can use ordinary setuptools (or whatever) rather than Meson.

Maybe one day :) but the reality is that compiled code is scattered all over SciPy in various languages and this is a fast moving area, with various things getting translated and rewritten etc.

`xsf` is an example taking a step in this direction, where part of the compiled code backend will eventually be broken out into a separate library (of course we might still vendor that in the repo rather than distributing a separate dependency for the time-being). But it will be an enormous amount of work to do this for every corner of SciPy, so I wouldn't keep your hopes up for it happening soon :sweat_smile:.

> I'm just frustrated with my inability to build scipy

Don't worry, this a common experience. Seriously though, this is why https://github.com/rgommers/pixi-dev-scipystack will become very valuable.

NeilGirdhar:
Awesome thanks for explaining, and thanks for your help!

NeilGirdhar:
Sorry that I haven't had time to work on this.  Has this been implemented now by someone else?  Should we close?

crusaderky:
It hasn't; however the module has changed quite a lot in the meantime.

steppi:
I can clean up the merge conflicts and will take a close look at everything.

steppi:
I've resolved the merge conflicts and carried on from there to add dispatch to all available special functions across backends. I'll post a comment explaining everything tomorrow. I needed to add some new fields to `_FuncInfo` to handle various edge cases.

steppi:
I'll investigate the failures and try to figure out why they aren't happening locally.

scipy/special/tests/test_support_alternative_backends.py
```diff
@@ -21,41 +22,92 @@
...
+        return "dask.array"
+    if is_jax(xp):
+        return "jax.numpy"
+    if is_numpy(xp):
+        return "numpy"
+    if is_torch(xp):
+        return "torch"
+    if is_array_api_strict(xp):
+        return "array_api_strict"
+    raise ValueError(f"Received unsupported array namspace {xp}")
```
lucascolley: Not blocking for this PR: it would be good to refactor tests to make use of an enum of backends, like https://github.com/data-apis/array-api-extra/blob/main/src/array_api_extra/_lib/_backends.py.

Blocking for this PR: general utilities like this belong in `_lib`.

scipy/special/tests/test_support_alternative_backends.py
```diff
@@ -21,41 +22,92 @@
...
+        x = x.take(0)
+    dtype = x.dtype
+    if xp.isdtype(dtype, "integral"):
+        return int(x)
+    if xp.isdtype(dtype, "real floating"):
+        return float(x)
+    if xp.isdtype(dtype, "complex floating"):
+        return complex(x)
+    if xp.isdtype(dtype, "bool"):
+        return bool(x)
```
lucascolley: likewise, this belongs in `_lib`

scipy/special/_support_alternative_backends.py
```diff
@@ -254,38 +288,518 @@ def fun(t, df, p):  return stdtr(df, t) - p
 # Inventory of automatically dispatched functions
 # IMPORTANT: these must all be **elementwise** functions!
 
-# PyTorch doesn't implement `betainc`.
-# On torch CPU we can fall back to NumPy, but on GPU it won't work.
-_needs_betainc = xp_capabilities(cpu_only=True, exceptions=['jax.numpy', 'cupy'])
+# PyTorch doesn"t implement `betainc`.
+# On torch CPU we can fall back to NumPy, but on GPU it won"t work.
```
lucascolley: erroneous formatting changes

lucascolley (changes requested):

scipy/special/tests/test_support_alternative_backends.py
```diff
@@ -21,41 +22,92 @@
...
+        return "torch"
+    if is_array_api_strict(xp):
+        return "array_api_strict"
+    raise ValueError(f"Received unsupported array namspace {xp}")
+
+
+def _to_python_scalar(x):
+    xp = array_namespace(x)
+    if x.ndim != 0:
+        x = x.take(0)
```
ev-br: `.take` method is not array API compliant, I think?

Also it's a bit tricky to follow the logic --- or I am just being dense --- if `x` can be a scalar, could you write it roughly

```
if isinstance(x, bool |int | float | complex):
     return x 

xp = array_namespace(x)

if xp.isdtype(x.dtype, ....):
  ....
```

steppi: https://data-apis.org/array-api/latest/API_specification/generated/array_api.take.html#take

I didn't spend any time thinking about writing this in a nice way, so can definitely be simplified though. Will revise, (and add to `_lib`).

scipy/special/tests/test_support_alternative_backends.py
steppi: Yeah, `enum` of backends sounds like a good idea.

scipy/special/tests/test_support_alternative_backends.py
lucascolley: I think Evgeni means that we should be using `xp.take` rather than `x.take`.

steppi: Oh, ha. Sorry, I didn't notice I didn't do that

steppi: The very existence of this function is kind of just a hack around the current test infra, so after I get all tests passing, I'll try to refactor so it doesn't need to exist

steppi:
Hopefully all of the tests will pass now. The issues

1. I didn't realize we need to set `SCIPY_DEVICE="cuda"` to get the torch tests to run on GPU.
2. `betaln` isn't actually available in PyTorch but it was marked as if it was.
3. `alt_names_map` wasn't actually working because I didn't normalize the namespace names to the native ones (didn't strip out the `array_api_compat.` prefix).

I got rid of that silly `_to_python_scalar` function. I'll have more detailed comments tomorrow and will remove the now unused imports the linter is complaining in a `skip-ci` commit if CI settles.

steppi:
The JAX GPU tests are failing because the JAX in CI is `v0.6.0`, but `hyp2f1` landed in `v0.6.1`. For whatever reason, `v0.6.1` and `v0.6.2` are up on PyPI but haven't landed in the conda feedstock yet, https://github.com/conda-forge/jax-feedstock/pulls. I can just remove it for now. Having a `requires_version` option in `_FuncInfo` seems like it would be useful though, but leaving out for now.

There's also some unexpected failures to pickle for random functions. I'll keep looking into that.

steppi:
```
=================================== FAILURES ===================================
______________________________ test_pickle[cbrt] _______________________________

func = <ufunc 'cbrt'>

    @pytest.mark.parametrize("func", [nfo.wrapper for nfo in _special_funcs])
    def test_pickle(func):
>       roundtrip = pickle.loads(pickle.dumps(func))
                                 ^^^^^^^^^^^^^^^^^^

func       = <ufunc 'cbrt'>

scipy/special/tests/test_support_alternative_backends.py:294: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/numpy/core/__init__.py:32: in __getattr__
    _raise_warning(attr_name)
        attr       = <ufunc 'cbrt'>
        attr_name  = 'cbrt'
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

attr = 'cbrt', submodule = None

    def _raise_warning(attr: str, submodule: str | None = None) -> None:
        new_module = "numpy._core"
        old_module = "numpy.core"
        if submodule is not None:
            new_module = f"{new_module}.{submodule}"
            old_module = f"{old_module}.{submodule}"
>       warnings.warn(
            f"{old_module} is deprecated and has been renamed to {new_module}. "
            "The numpy._core namespace contains private NumPy internals and its "
            "use is discouraged, as NumPy internals can change without warning in "
            "any release. In practice, most real-world usage of numpy.core is to "
            "access functionality in the public NumPy API. If that is the case, "
            "use the public NumPy API. If not, you are using NumPy internals. "
            "If you would still like to access an internal attribute, "
            f"use {new_module}.{attr}.",
            DeprecationWarning,
            stacklevel=3
        )
E       DeprecationWarning: numpy.core is deprecated and has been renamed to numpy._core. The numpy._core namespace contains private NumPy internals and its use is discouraged, as NumPy internals can change without warning in any release. In practice, most real-world usage of numpy.core is to access functionality in the public NumPy API. If that is the case, use the public NumPy API. If not, you are using NumPy internals. If you would still like to access an internal attribute, use numpy._core.cbrt.

attr       = 'cbrt'
new_module = 'numpy._core'
old_module = 'numpy.core'
submodule  = None

/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/numpy/core/_utils.py:10: DeprecationWarning
______________________________ test_pickle[exp2] _______________________________

func = <ufunc 'exp2'>

    @pytest.mark.parametrize("func", [nfo.wrapper for nfo in _special_funcs])
    def test_pickle(func):
>       roundtrip = pickle.loads(pickle.dumps(func))
                                 ^^^^^^^^^^^^^^^^^^

func       = <ufunc 'exp2'>

scipy/special/tests/test_support_alternative_backends.py:294: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/numpy/core/__init__.py:32: in __getattr__
    _raise_warning(attr_name)
        attr       = <ufunc 'exp2'>
        attr_name  = 'exp2'
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

attr = 'exp2', submodule = None

    def _raise_warning(attr: str, submodule: str | None = None) -> None:
        new_module = "numpy._core"
        old_module = "numpy.core"
        if submodule is not None:
            new_module = f"{new_module}.{submodule}"
            old_module = f"{old_module}.{submodule}"
>       warnings.warn(
            f"{old_module} is deprecated and has been renamed to {new_module}. "
            "The numpy._core namespace contains private NumPy internals and its "
            "use is discouraged, as NumPy internals can change without warning in "
            "any release. In practice, most real-world usage of numpy.core is to "
            "access functionality in the public NumPy API. If that is the case, "
            "use the public NumPy API. If not, you are using NumPy internals. "
            "If you would still like to access an internal attribute, "
            f"use {new_module}.{attr}.",
            DeprecationWarning,
            stacklevel=3
        )
E       DeprecationWarning: numpy.core is deprecated and has been renamed to numpy._core. The numpy._core namespace contains private NumPy internals and its use is discouraged, as NumPy internals can change without warning in any release. In practice, most real-world usage of numpy.core is to access functionality in the public NumPy API. If that is the case, use the public NumPy API. If not, you are using NumPy internals. If you would still like to access an internal attribute, use numpy._core.exp2.

attr       = 'exp2'
new_module = 'numpy._core'
old_module = 'numpy.core'
submodule  = None

/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/numpy/core/_utils.py:10: DeprecationWarning
FAILED scipy/special/tests/test_support_alternative_backends.py::test_pickle[cbrt] - DeprecationWarning: numpy.core is deprecated and has been renamed to numpy._core. The numpy._core namespace contains private NumPy internals and its use is discouraged, as NumPy internals can change without warning in any release. In practice, most real-world usage of numpy.core is to access functionality in the public NumPy API. If that is the case, use the public NumPy API. If not, you are using NumPy internals. If you would still like to access an internal attribute, use numpy._core.cbrt.
FAILED scipy/special/tests/test_support_alternative_backends.py::test_pickle[exp2] - DeprecationWarning: numpy.core is deprecated and has been renamed to numpy._core. The numpy._core namespace contains private NumPy internals and its use is discouraged, as NumPy internals can change without warning in any release. In practice, most real-world usage of numpy.core is to access functionality in the public NumPy API. If that is the case, use the public NumPy API. If not, you are using NumPy internals. If you would still like to access an internal attribute, use numpy._core.exp2.
= 2 failed, 66205 passed, 7604 skipped, 11691 deselected, 213 xfailed, 11 xpassed in 274.70s (0:04:34) =
```

so strange. I can't reproduce this on my Linux machine or my Mac mini. The only commonality is that `exp2` and `cbrt` are ufuncs which are also in NumPy, (though SciPy has its own implementations from Cephes). For now, I guess I'll just suppress the warning and make an issue about it.

lucascolley (approved):
I think I'm happy for this to go in when CI is green!

steppi:
Thanks @lucascolley. It looks like the failures are unrelated so far. A few comments I wanted to add.

1. I didn't add `special.sinc` because it isn't in `jax.scipy.special` but instead is in `jax.numpy` (since `sinc` is in NumPy and SciPy's `sinc` is just an alias for NumPy's). But `sinc` is in CuPy and PyTorch, I just had missed that only the JAX tests were failing. I made a determination to not add any more special case logic, but thought to follow up with allowing functions to come from an alternative namespace. Since `sinc` is in PyTorch and CuPy though, I made as well at it for those in this PR.

2. I wanted to make sure that adding all of these extra attributes to the `_FuncInfo` wouldn't lead to bloat, so measured with [pympler](https://pympler.readthedocs.io/en/latest/). Things don't look too bad.

Memory usage on main after importing `special`.

```
                         types |   # objects |   total size
============================== | =========== | ============
                           str |       28758 |      5.27 MB
                          code |        9387 |      3.79 MB
                          dict |        7840 |      2.65 MB
                          type |        1311 |      1.38 MB
                         tuple |       10553 |    639.02 KB
  Cython.Shadow.CythonMetaType |         198 |    329.48 KB
                          list |        2439 |    285.39 KB
                           set |         266 |    245.11 KB
                   abc.ABCMeta |         131 |    184.99 KB
            wrapper_descriptor |        2421 |    170.23 KB
         weakref.ReferenceType |        1997 |    156.02 KB
    builtin_function_or_method |        1933 |    135.91 KB
                   numpy.ufunc |         505 |    126.25 KB
             method_descriptor |        1354 |     95.20 KB
                     frozenset |         257 |     94.21 KB
```

Memory usage on this PR branch after importing `special`.

```
                         types |   # objects |   total size
============================== | =========== | ============
                           str |       28839 |      5.29 MB
                          code |        9393 |      3.80 MB
                          dict |        7990 |      2.69 MB
                          type |        1311 |      1.38 MB
                         tuple |       10635 |    644.56 KB
  Cython.Shadow.CythonMetaType |         198 |    329.48 KB
                          list |        2934 |    314.41 KB
                           set |         261 |    244.05 KB
                   abc.ABCMeta |         131 |    184.99 KB
            wrapper_descriptor |        2421 |    170.23 KB
         weakref.ReferenceType |        1987 |    155.23 KB
    builtin_function_or_method |        1927 |    135.49 KB
                   numpy.ufunc |         506 |    126.50 KB
             method_descriptor |        1354 |     95.20 KB
                     frozenset |         257 |     94.21 KB
```

3. Some of the weird edge cases, like `multigammaln` having a parameter which can only be a Python int should probably be changed on the function side, rather than maintaining support for handling all of these edge cases.

steppi:
Crossing fingers, but that should take care of the last of the failures. On `torch`, when the default dtype is set to `float32`, the function `_skip_or_tweak_alternative_backends` was converting integer arguments to type `float32` to work around inconsistency in `int64` to float conversions, but this shouldn't be allowed to happen for arguments which can only be integers.

I also added `sinc` support for CuPy and PyTorch. If CI goes green, this is ready from my end.

steppi:
The failure is related

```
..................                                                       [ 80%]
scipy/special/tests/test_support_alternative_backends.py .FFF.FFF.FFF.FF [ 80%]
F.FFF.FFF.FFF.FFF.FFF................................................... [ 81%]
........................................................................ [ 81%]
........................................................................ [ 81%]
........................................................................ [ 81%]
........................................................................ [ 81%]
........................................................................ [ 81%]
........................................................................ [ 81%]
........................................................................ [ 81%]
........................................................................ [ 81%]
........................................................................ [ 81%]
.........ssssssssssss.FFF.FFF.FFF.FFF.FFF.FFF.FFF.FFF.FF
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/scipy/test/lib/python3.11/site-packages/pluggy/_callers.py", line 43, in run_old_style_hookwrapper
INTERNALERROR>     teardown.send(result)
INTERNALERROR>   File "/scipy/test/lib/python3.11/site-packages/_hypothesis_pytestplugin.py", line 329, in pytest_runtest_makereport
INTERNALERROR>     report = (yield).get_result()
INTERNALERROR>              ^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/scipy/test/lib/python3.11/site-packages/pluggy/_result.py", line 103, in get_result
INTERNALERROR>     raise exc.with_traceback(tb)
INTERNALERROR>   File "/scipy/test/lib/python3.11/si
```

I don't understand it yet.

scipy/special/tests/test_support_alternative_backends.py
```diff
@@ -67,37 +87,101 @@ def _skip_or_tweak_alternative_backends(xp, f_name, dtypes):
         # to float32, and sometimes to float64.
         # When they are promoted to float32, explicitly convert the reference
         # numpy arrays to float32 to prevent them from being automatically promoted
-        # to float64 instead.
-        dtypes_np_ref = ['float32' if 'int' in dtype else dtype for dtype in dtypes]
+        # to float64 instead. Do not convert integer only args to float32 though.
+        dtypes_np_ref = [
+            'float32' if 'int' in dtype and not needs_int else dtype
+            for dtype, needs_int in zip(dtypes, int_only)
```
lucascolley: This line relies on `int_only` being iterable, but elsewhere it is passed as a single bool

steppi: I saw that's the problem in the array api / torch etc tests. I wonder if the strange failures in the Linux / 32 bit tests are for the same reason though. Those are the ones that confuse me.

steppi:
OK, I think that should take care of the remaining test failures.

scipy/special/tests/test_support_alternative_backends.py
```diff
@@ -21,83 +21,167 @@
...
-        dtypes_np_ref = ['float32' if 'int' in dtype else dtype for dtype in dtypes]
+    if (is_torch(xp) and xp_default_dtype(xp) == xp.float32):
+        # On PyTorch with float32 default dtype, all ints are promoted
+        # to float32, but when falling back to NumPy/SciPy int64 is promoted
+        # instead to float64. Integer only parameters essentially do not
+        # participate in determination of the result type in PyTorch with
+        # float32 default dtype, but will impact the output dtype as if
+        # they were float64 when falling back to NumPy/SciPy.
+        if not nfo.torch_native:
+            pytest.xfail("dtypes no not match")
```
steppi: The previous comment made it seem like PyTorch randomly mapped to `float32` or `float64` in this case, but I discovered `float64` output only occurs when PyTorch falls back to NumPy/SciPy, which have no concept of `default_dtype`. I added a flag to the `_FuncInfo` objects to encode whether PyTorch falls back, and just insert the `xfails` here.

scipy/special/tests/test_support_alternative_backends.py
```diff
@@ -21,83 +21,167 @@
...
-    dtypes_np_ref = dtypes
-    if (is_torch(xp) and xp_default_dtype(xp) == xp.float32
-        and f_name not in {'betainc', 'betaincc', 'stdtr', 'stdtrit'}
-    ):
-        # On PyTorch with float32 default dtype, sometimes ints are promoted
-        # to float32, and sometimes to float64.
-        # When they are promoted to float32, explicitly convert the reference
-        # numpy arrays to float32 to prevent them from being automatically promoted
-        # to float64 instead.
-        dtypes_np_ref = ['float32' if 'int' in dtype else dtype for dtype in dtypes]
```
steppi, line 71: This handled the different promotion rules by modifying the inputs, but doesn't work for instance when there are integer only arguments. I changed it to modify the output to the correct type instead, directly at the point where the reference output is computed.

scipy/special/tests/test_support_alternative_backends.py
```diff
@@ -108,10 +192,25 @@ def test_support_alternative_backends(xp, func, nfo, dtype, shapes):
...
     ref = nfo.func(*args_np)  # Unwrapped ufunc
-
+    if (
+            is_torch(xp)
+            and xp_default_dtype(xp) == xp.float32
+            and "float64" not in dtypes
+    ):
+        # int64 promotes like float32 on torch with default dtype = float32
+        # cast reference if needed
+        ref = np.float32(ref)
```
steppi: Eventually we should remove code duplication rather than just pasting this line in everywhere a reference output is computed, but I'd prefer to punt on what the best way to do this is. Using duplication to postpone the decision basically.

steppi (commented):
Added a few comments explaining what I did. I also realized I wasn't annotating functions which rely on NumPy/SciPy fallback for JAX correctly. I updated those to use `jax_jit=False`.

steppi:
Oh, ~looks like `test_public_api.py` needs to be updated for how I handled giving alternative backend support for the alias `digamma` for `psi`~, the `digamma` alias still needs to be importable from `special.basic` because that deprecation hasn't expired yet.

steppi:
It looks like the weird memory error on 32 bit Linux still happens even after everything else is fixed. I'll see if I can reproduce in a 32 bit Linux Docker container. 

Update: I can reproduce it locally now. Will investigate further tomorrow.

steppi:
> It looks like the weird memory error on 32 bit Linux still happens even after everything else is fixed. I'll see if I can reproduce in a 32 bit Linux Docker container.
> 
> Update: I can reproduce it locally now. Will investigate further tomorrow.

Ah, it's a confluence of things. On 32 bit systems, the NumPy default `int` is `int32`, the tests use `int64` type. The overloads for some functions look like this

```python
    "bdtr": {
        "_legacy.pxd": {
            "bdtr_unsafe": "ddd->d"
        },
        "xsf_wrappers.h": {
            "cephes_bdtr_wrap": "dpd->d"
        }
    },
```

the `p` in `"dpd->d"` is for `ptrdiff_t` which will always be the NumPy default `int`. This `int32` on 32 bit systems, so when you pass an `int64`, the deprecated "`ddd->d`" signature is chosen, which leads to deprecation warnings. A lot of deprecation warnings. So many deprecation warnings that the limited 4GB of RAM you get with a 32 bit address space gets flooded when pytest is doing its teardown and we see an out of memory error.

For a principled fix, maybe we should provide the `int32` and `int64` overload for all integer parameters, not just an overload for the NumPy default `int`. Sebastian Berg had mentioned this here, https://github.com/scipy/scipy/pull/21401#issuecomment-2296164860, using a `long long` `q` typecode, as a way to allow passing `int64` on 32 bit platforms. For now, I can just conditionally make the integer types `int32` on 32 bit systems in the tests.

steppi:
@lucascolley, it looks like I finally got all tests passing.

scipy/special/tests/test_support_alternative_backends.py
lucascolley: please can you investigate getting rid of `xp_default_dtype` and replacing it with `xpx.default_dtype`? https://data-apis.org/array-api-extra/generated/array_api_extra.default_dtype.html

scipy/special/tests/test_support_alternative_backends.py
```diff
@@ -21,83 +21,172 @@
...
-        dtypes_np_ref = ['float32' if 'int' in dtype else dtype for dtype in dtypes]
+    if (is_torch(xp) and xp_default_dtype(xp) == xp.float32):
+        # On PyTorch with float32 default dtype, all ints are promoted
+        # to float32, but when falling back to NumPy/SciPy int64 is promoted
+        # instead to float64. Integer only parameters essentially do not
+        # participate in determination of the result type in PyTorch with
+        # float32 default dtype, but will impact the output dtype as if
+        # they were float64 when falling back to NumPy/SciPy.
+        if not nfo.torch_native:
+            pytest.xfail("dtypes no not match")
```
lucascolley:
```suggestion
            pytest.xfail("dtypes do not match")
```

scipy/special/_support_alternative_backends.py
```diff
@@ -256,40 +299,577 @@ def fun(t, df, p):  return stdtr(df, t) - p
...
+        xp_capabilities(
+            cpu_only=True, exceptions=["cupy"],
+            jax_jit=False,
+        ),
+        torch_native=False,
+    ),
     _FuncInfo(_ufuncs.chdtr, 2, generic_impl=_chdtr),
-    _FuncInfo(_ufuncs.chdtrc, 2, generic_impl=_chdtrc),
+    _FuncInfo(_ufuncs.chdtrc, 2, generic_impl=_chdtrc,
+              # gh-20972
```
lucascolley: totally optional nit but just for future reference, this makes a nice hyperlink in the GitHub web UI

```suggestion
              # scipy/scipy#20972
```

lucascolley (changes requested):
thanks @steppi , one request

NeilGirdhar:
Just returning from vacation and want to congratulate everyone on this Herculean effort.  This looks amazing!  Great job @steppi

steppi:
@lucascolley, I changed it from using `xp_default_dtype` to `xpx.default_dtype`. For the future, I'll check if utilities I use from `scipy._lib._array_api` are being phased out in favor of an `xpx` version.

lucascolley:
thanks! You said:

> The previous comment made it seem like PyTorch randomly mapped to float32 or float64 in this case, but I discovered float64 output only occurs when PyTorch falls back to NumPy/SciPy, which have no concept of `default_dtype`.

Can you check whether that last phrase still holds true now that we are using `xpx.default_dtype`? If it doesn't, are there any changes to be made?

EDIT: thinking a bit more, I guess there won't be any more work to do for this specific case, as there is still no variable default dtype for NumPy. But maybe this affects something on the JAX side of things?

lucascolley (approved):
unless my previous comment prompts any further thoughts, happy to click the green button. Thanks all!

steppi:
> thanks! You said:
> 
> > The previous comment made it seem like PyTorch randomly mapped to float32 or float64 in this case, but I discovered float64 output only occurs when PyTorch falls back to NumPy/SciPy, which have no concept of `default_dtype`.
> 
> Can you check whether that last phrase still holds true now that we are using `xpx.default_dtype`? If it doesn't, are there any changes to be made?
> 
> EDIT: thinking a bit more, I guess there won't be any more work to do for this specific case, as there is still no variable default dtype for NumPy. But maybe this affects something on the JAX side of things?

Yeah, nothing should change here due to using `xpx.default_dtype`. Like you wrote in the edit, NumPy has no concept of default dtype, so computations going through NumPy can't match PyTorch's  `default_dtype=float32` promotion behavior. I'm not sure if we have JAX tests with the default dtype set to `float32`, but in any case, when `SCIPY_DEFAULT_DTYPE="float32"` then in JAX, `int64` isn't even an available type, since what happens when `SCIPY_DEFAULT_DTYPE="float32"` is that `JAX_ENABLE_X64` gets set to `False`. I don't think JAX has the same notion of default dtype that PyTorch has, instead you can turn `X64` support off and on. This means that the problems that arise when PyTorch falls back to SciPy/NumPy cannot occur, since it's not possible to mix `float32` and `int64` in the first place.

lucascolley:
> I'm not sure if we have JAX tests with the default dtype set to float32

Indeed we don't, AFAIK since JAX basically has two modes: either default `float32` and throw errors when trying to use double precision, or default to double precision. Unless anything has moved recently, I don't think there is a mode which enables double precision while still allowing you to default to `float32`. So I don't think these tests are possible atm.

---

gh-23081 next :) ?

steppi:
> [gh-23081](https://github.com/scipy/scipy/pull/23081) next :) ?

Yes, working on finishing it now. Main thing left is a comment explaining what is and isn't included in the tables and why some submodules are missing.

steppi:
> > match PyTorch's `default_dtype=float32`
> 
> I haven't been following the whole discussion, but just in case you don't know that Jax also has a default dtype of `float32`.

Not exactly, see above. You can only turn on and off `JAX_ENABLE_X64` so it's not possible to mix `int64` and `float32` types, so these differences in promotion depending on fallback to numpy/scipy or native computation cannot occur.

PR #22796, opened 2025-04-03
ENH/BLD: cython: share memoryview utility between extension modules

rgommers:
This reduces the size of our wheels by ~1.8 MB, by compiling a bunch of Cython's utility code once in total rather than once per extension module. See https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#shared-utility-module for context.

Closes gh-22775

### DIFF
--- a/scipy/meson.build
+++ b/scipy/meson.build
@@ -380,18 +380,41 @@ _cython_tree = [fs.copyfile('__init__.py')]
 cython_args = ['-3', '--fast-fail', '--output-file', '@OUTPUT@', '--include-dir', '@BUILD_ROOT@', '@INPUT@']
 if cy.version().version_compare('>=3.1.0')
   cython_args += ['-Xfreethreading_compatible=True']
+
+  cython_shared_src = custom_target(
+    install: false,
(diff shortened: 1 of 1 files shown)


rgommers:
Draft only because I want to see regular CI run before looking at `[wheel build]`. Good chance this is fine as is.

rgommers:
Looks like this is good to go. Cython 3.1.0b1 is available on PyPI and on conda-forge, so there's no blocker to merging it.

lucascolley (approved):

rgommers:
@lesteve this may interest you - the benefit for scikit-learn is probably similar in Mb's, so even larger in a relative sense - I'd expect this to shrink your wheel sizes by 15% or so.

ilayn:
@rgommers Do you have a recent breakdown of our wheel binaries at your convenience?

ilayn:
These are uncompressed values



linalg_pythran is unnecessarily big that I'll have a look and `sparse` is mostly propack and arpack which triggered me. At some point we'll circle back to `cython_special` apparently, there must be a better way to do it.

PR #22798, opened 2025-04-03
TST: stats: mark some tests as slow

mdhaber:
I've seen these tests (`TestMakeDistribution::test_custom_variable_support[1]` and `TestMakeDistribution::test_rv_generic` with `norminvgauss`) and failing as slightly too slow recently. 

I think it's safe to mark them as slow to run them on fewer platforms and give them a little more time to finish.


mdhaber:
Are you comfortable with these being marked slow @steppi?

steppi (approved):
Looks good to me.

PR #22803, opened 2025-04-04
MAINT: use `xp.asarray` instead of `xp.array`

Saransh-cpp:

A quick PR to get rid of a couple of final/leftover instances on `xp.array`. (No more `xp.array` in the codebase now!)


lucascolley (approved):
LGTM, thanks!

PR #22805, opened 2025-04-06
CI: start using the `CIBW_ENABLE` env var

rgommers:
The `CIBW_FREE_THREADED_SUPPORT` and `CIBW_PRERELEASE_PYTHONS` are deprecated (see https://cibuildwheel.pypa.io/en/stable/options/#enable).

There is no functional change, this is just a cleanup.

I skipped CI since there are no changes affecting regular CI, and I ran wheel builds on my fork ([in this job](https://github.com/rgommers/scipy/actions/runs/14294661858)) for completeness.


charris:
Thanks Ralf.

PR #22810, opened 2025-04-07
BUG: special.hyp2f1: fix for extreme inputs

### LINKED ISSUE #20988: BUG: `scipy.special.hyp2f1` gives wrong result for extreme inputs
### Describe your issue.

With https://github.com/scipy/scipy/pull/20089, the behavior of `scipy.special.hyp2f1` has changed for extreme inputs.

For example, calling `hyp2f1(1.0, -np.inf, 3.0, 4.0+1.0j)` yields `(1+0j)`. Based on my understanding of `hyp2f1`, this isn't the correct result. As the `-np.inf` argument is outside of the domain that is handled correctly, the call should return `(nan+nanj)`. Indeed, this was the behavior in the latest released version (v1.13.1), i.e., before https://github.com/scipy/scipy/pull/20089.

For comparison, calling `hyp2f1(1.0, -1000, 3.0, 4.0+1.0j)`, i.e., with a small but not `-np.inf` argument, yields `(nan+nanj)`. By extension, I'd expect `hyp2f1(1.0, -np.inf, 3.0, 4.0+1.0j)` to also yield `(nan+nanj)`.

### Reproducing Code Example

```python
import numpy as np
from scipy.special import hyp2f1

r = hyp2f1(1.0, -np.inf, 3.0, 4.0+1.0j)
print(r)   # prints (1+0j) in main, but (nan+nanj) in v1.13.1
```

### Error message

```shell
(no error)
```

```shell
1.15.0.dev0+950.c22b657 1.26.4 sys.version_info(major=3, minor=12, micro=3, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    detection method: pkgconfig
    found: true
    include directory: /root/conda/envs/scipy-dev/include
    lib directory: /root/conda/envs/scipy-dev/lib
    name: openblas
      NO_LAPACK=0 NO_LAPACKE= NO_AFFINITY=1 USE_OPENMP=0 PRESCOTT MAX_THREADS=128
    pc file directory: /root/conda/envs/scipy-dev/lib/pkgconfig
    version: 0.3.27
  lapack:
    detection method: pkgconfig
    found: true
    include directory: /root/conda/envs/scipy-dev/include
    lib directory: /root/conda/envs/scipy-dev/lib
    name: openblas
      NO_LAPACK=0 NO_LAPACKE= NO_AFFINITY=1 USE_OPENMP=0 PRESCOTT MAX_THREADS=128
    pc file directory: /root/conda/envs/scipy-dev/lib/pkgconfig
    version: 0.3.27
  pybind11:
    detection method: pkgconfig
    include directory: /root/conda/envs/scipy-dev/include
    name: pybind11
    version: 2.12.0
Compilers:
  c:
    args: -march=nocona, -mtune=haswell, -ftree-vectorize, -fPIC, -fstack-protector-strong,
      -fno-plt, -O2, -ffunction-sections, -pipe, -isystem, /root/conda/envs/scipy-dev/include,
      -DNDEBUG, -D_FORTIFY_SOURCE=2, -O2, -isystem, /root/conda/envs/scipy-dev/include
    commands: /root/conda/envs/scipy-dev/bin/x86_64-conda-linux-gnu-cc
    linker: ld.bfd
    linker args: -Wl,-O2, -Wl,--sort-common, -Wl,--as-needed, -Wl,-z,relro, -Wl,-z,now,
      -Wl,--disable-new-dtags, -Wl,--gc-sections, -Wl,--allow-shlib-undefined, -Wl,-rpath,/root/conda/envs/scipy-dev/lib,
      -Wl,-rpath-link,/root/conda/envs/scipy-dev/lib, -L/root/conda/envs/scipy-dev/lib,
      -march=nocona, -mtune=haswell, -ftree-vectorize, -fPIC, -fstack-protector-strong,
      -fno-plt, -O2, -ffunction-sections, -pipe, -isystem, /root/conda/envs/scipy-dev/include,
      -DNDEBUG, -D_FORTIFY_SOURCE=2, -O2, -isystem, /root/conda/envs/scipy-dev/include
    name: gcc
    version: 12.3.0
  c++:
    args: -fvisibility-inlines-hidden, -fmessage-length=0, -march=nocona, -mtune=haswell,
      -ftree-vectorize, -fPIC, -fstack-protector-strong, -fno-plt, -O2, -ffunction-sections,
      -pipe, -isystem, /root/conda/envs/scipy-dev/include, -DNDEBUG, -D_FORTIFY_SOURCE=2,
      -O2, -isystem, /root/conda/envs/scipy-dev/include
    commands: /root/conda/envs/scipy-dev/bin/x86_64-conda-linux-gnu-c++
    linker: ld.bfd
    linker args: -Wl,-O2, -Wl,--sort-common, -Wl,--as-needed, -Wl,-z,relro, -Wl,-z,now,
      -Wl,--disable-new-dtags, -Wl,--gc-sections, -Wl,--allow-shlib-undefined, -Wl,-rpath,/root/conda/envs/scipy-dev/lib,
      -Wl,-rpath-link,/root/conda/envs/scipy-dev/lib, -L/root/conda/envs/scipy-dev/lib,
      -fvisibility-inlines-hidden, -fmessage-length=0, -march=nocona, -mtune=haswell,
      -ftree-vectorize, -fPIC, -fstack-protector-strong, -fno-plt, -O2, -ffunction-sections,
      -pipe, -isystem, /root/conda/envs/scipy-dev/include, -DNDEBUG, -D_FORTIFY_SOURCE=2,
      -O2, -isystem, /root/conda/envs/scipy-dev/include
    name: gcc
    version: 12.3.0
  cython:
    commands: cython
    linker: cython
    name: cython
    version: 3.0.10
  fortran:
    args: -march=nocona, -mtune=haswell, -ftree-vectorize, -fPIC, -fstack-protector-strong,
      -fno-plt, -O2, -ffunction-sections, -pipe, -isystem, /root/conda/envs/scipy-dev/include
    commands: /root/conda/envs/scipy-dev/bin/x86_64-conda-linux-gnu-gfortran
    linker: ld.bfd
    linker args: -Wl,-O2, -Wl,--sort-common, -Wl,--as-needed, -Wl,-z,relro, -Wl,-z,now,
      -Wl,--disable-new-dtags, -Wl,--gc-sections, -Wl,--allow-shlib-undefined, -Wl,-rpath,/root/conda/envs/scipy-dev/lib,
      -Wl,-rpath-link,/root/conda/envs/scipy-dev/lib, -L/root/conda/envs/scipy-dev/lib,
      -march=nocona, -mtune=haswell, -ftree-vectorize, -fPIC, -fstack-protector-strong,
      -fno-plt, -O2, -ffunction-sections, -pipe, -isystem, /root/conda/envs/scipy-dev/include
    name: gcc
    version: 12.3.0
  pythran:
    include directory: ../../../root/conda/envs/scipy-dev/lib/python3.12/site-packages/pythran
    version: 0.15.0
Machine Information:
  build:
    cpu: x86_64
    endian: little
    family: x86_64
    system: linux
  cross-compiled: false
  host:
    cpu: x86_64
    endian: little
    family: x86_64
    system: linux
Python Information:
  path: /root/conda/envs/scipy-dev/bin/python3.12
  version: '3.12'
```

czgdp1807:
#### Reference issue

Closes https://github.com/scipy/scipy/issues/20988

The fix is simple. There are two instances where we check, `max_degree <= UINT64_MAX`. However, just after these checks we do, `max_degree + 1` and `max_degree + 2`. So consider the edge case when `max_degree = UINT64_MAX` in case when `a` or `b` is `np.inf`. Then `max_degree + 1` and `max_degree + 2` will circle to `0`. And therefore the number of terms in the series expansion is `0`. So we get `0` as the output which the issue highlights.

To fix the problem I changed `max_degree <= UINT64_MAX - 1` and `max_degree <= UINT64 - 2`. So we get a `nan` now.

Tests added for the same.

#### Additional information

cc: @michaelpradel @steppi

### DIFF
--- a/scipy/special/xsf/hyp2f1.h
+++ b/scipy/special/xsf/hyp2f1.h
@@ -553,7 +553,7 @@ XSF_HOST_DEVICE inline std::complex<double> hyp2f1(double a, double b, double c,
      * the series at a or b of smaller magnitude. This is to ensure proper
      * handling of situations like a < c < b <= 0, a, b, c all non-positive
      * integers, where terminating at a would lead to a term of the form 0 / 0. */
-    std::uint64_t max_degree;
+    double max_degree;
     if (a_neg_int || b_neg_int) {
         if (a_neg_int && b_neg_int) {
             max_degree = a > b ? std::abs(a) : std::abs(b);
@@ -562,7 +562,7 @@ XSF_HOST_DEVICE inline std::complex<double> hyp2f1(double a, double b, double c,
         } else {
             max_degree = std::abs(b);
         }
-        if (max_degree <= UINT64_MAX - 1) {
+        if (max_degree <= (double) UINT64_MAX) {
             auto series_generator = detail::HypergeometricSeriesGenerator(a, b, c, z);
             return detail::series_eval_fixed_length(series_generator, std::complex<double>{0.0, 0.0}, max_degree + 1);
         } else {
@@ -583,7 +583,7 @@ XSF_HOST_DEVICE inline std::complex<double> hyp2f1(double a, double b, double c,
      * (DLMF 15.8.1) */
     if (c_minus_a_neg_int || c_minus_b_neg_int) {
         max_degree = c_minus_b_neg_int ? std::abs(c - b) : std::abs(c - a);
-        if (max_degree <= UINT64_MAX - 2) {
+        if (max_degree <= (double) UINT64_MAX) {
             result = std::pow(1.0 - z, c - a - b);
             auto series_generator = detail::HypergeometricSeriesGenerator(c - a, c - b, c, z);
             result *=
--- a/scipy/special/xsf/hyp2f1.h
+++ b/scipy/special/xsf/hyp2f1.h
@@ -562,7 +562,7 @@ XSF_HOST_DEVICE inline std::complex<double> hyp2f1(double a, double b, double c,
         } else {
             max_degree = std::abs(b);
         }
-        if (max_degree <= UINT64_MAX) {
+        if (max_degree <= UINT64_MAX - 1) {
             auto series_generator = detail::HypergeometricSeriesGenerator(a, b, c, z);
             return detail::series_eval_fixed_length(series_generator, std::complex<double>{0.0, 0.0}, max_degree + 1);
         } else {
@@ -583,7 +583,7 @@ XSF_HOST_DEVICE inline std::complex<double> hyp2f1(double a, double b, double c,
      * (DLMF 15.8.1) */
     if (c_minus_a_neg_int || c_minus_b_neg_int) {
         max_degree = c_minus_b_neg_int ? std::abs(c - b) : std::abs(c - a);
-        if (max_degree <= UINT64_MAX) {
+        if (max_degree <= UINT64_MAX - 2) {
             result = std::pow(1.0 - z, c - a - b);
             auto series_generator = detail::HypergeometricSeriesGenerator(c - a, c - b, c, z);
             result *=
--- a/scipy/special/tests/test_hyp2f1.py
+++ b/scipy/special/tests/test_hyp2f1.py
@@ -2544,6 +2544,8 @@ class TestHyp2f1:
             for case in mark.args[1]
         ]
 
+class TestHyp2f1ExtremeInputs:
+
     @pytest.mark.parametrize("a", [1.0, 2.0, 3.0, -np.inf, np.inf])
     @pytest.mark.parametrize("b", [3.0, 4.0, 5.0, -np.inf, np.inf])
     @pytest.mark.parametrize("c", [3.0, 5.0, 6.0, 7.0])
--- a/scipy/special/tests/test_hyp2f1.py
+++ b/scipy/special/tests/test_hyp2f1.py
@@ -2558,3 +2558,7 @@ class TestHyp2f1:
 
         assert(np.isnan(hyp2f1(1.0, 10**7, 3.0, 4.0 + 1.0j)))
         assert(np.isnan(hyp2f1(1.0, -10**7, 3.0, 4.0 + 1.0j)))
+
+        # Already correct in main but testing for surety
+        assert(np.isnan(hyp2f1(np.inf, 1.0, 3.0, 4.0)))
+        assert(np.isnan(hyp2f1(1.0, np.inf, 3.0, 4.0)))
--- a/scipy/special/tests/test_hyp2f1.py
+++ b/scipy/special/tests/test_hyp2f1.py
@@ -2543,3 +2543,18 @@ class TestHyp2f1:
             if mark.name == 'parametrize'
             for case in mark.args[1]
         ]
+
+    @pytest.mark.parametrize("a", [1.0, 2.0, 3.0, -np.inf, np.inf])
+    @pytest.mark.parametrize("b", [3.0, 4.0, 5.0, -np.inf, np.inf])
+    @pytest.mark.parametrize("c", [3.0, 5.0, 6.0, 7.0])
+    @pytest.mark.parametrize("z", [4.0 + 1.0j])
+    def test_inf_a_b(self, a, b, c, z):
+        if np.any(np.isinf(np.asarray([a, b]))):
+            assert(np.isnan(hyp2f1(a, b, c, z)))
+
+    def test_large_a_b(self):
+        assert(np.isnan(hyp2f1(10**7, 1.0, 3.0, 4.0 + 1.0j)))
+        assert(np.isnan(hyp2f1(-10**7, 1.0, 3.0, 4.0 + 1.0j)))
+
+        assert(np.isnan(hyp2f1(1.0, 10**7, 3.0, 4.0 + 1.0j)))
+        assert(np.isnan(hyp2f1(1.0, -10**7, 3.0, 4.0 + 1.0j)))


scipy/special/xsf/hyp2f1.h
```diff
@@ -562,7 +562,7 @@ XSF_HOST_DEVICE inline std::complex<double> hyp2f1(double a, double b, double c,
         } else {
             max_degree = std::abs(b);
         }
-        if (max_degree <= UINT64_MAX) {
+        if (max_degree <= UINT64_MAX - 1) {
```
tylerjereddy, line 565: Something isn't right here yet. On Linux `max_degree` keeps showing up as `0` in many of the new tests added here while it is a huge integer on MacOS, which probably explains their deviation in CI.

After reading the source and debug printing a bit locally, I found that the patch below allows `test_hyp2f1.py` to pass on Linux on this branch:

```diff
--- a/scipy/special/xsf/hyp2f1.h
+++ b/scipy/special/xsf/hyp2f1.h
@@ -553,7 +553,7 @@ XSF_HOST_DEVICE inline std::complex<double> hyp2f1(double a, double b, double c,
      * the series at a or b of smaller magnitude. This is to ensure proper
      * handling of situations like a < c < b <= 0, a, b, c all non-positive
      * integers, where terminating at a would lead to a term of the form 0 / 0. */
-    std::uint64_t max_degree;
+    double max_degree;
     if (a_neg_int || b_neg_int) {
```

MacOS seems to remain happy with the same patch. 

Maybe we should check with Albert, but there are a lot of `double` values and casts floating around here so the integer type strikes me as a bit weird, especially if we're approaching the max?

steppi: I see. It's platform dependent when casting `inf` to a `uint64_t` because it's undefined behavior to cast a `double` to a `uint64_t` if it doesn't fit.

@tylerjereddy is right.  This was a mistake on my part due to sticking too close to the original Fortran I was translating.  Casting to `uint64_t` shouldn't take place until after verification that `max_degree` will fit into a `uint64_t`.

I think that series evaluator will take at best hundreds of years to finish on current hardware if `max_degree` is anywhere close to `UINT64_MAX` so the bound for `max_degree` could be much lower without causing problems. I think it should be lower, to avoid having the function hang indefinitely if passed the wrong values for `a` and `b`, but that could be left for future work.

czgdp1807: I updated to use the patch in [@tylerjereddy's comment above](https://github.com/scipy/scipy/pull/22810#discussion_r2038793092). It seems to work on my Mac. Let's see how CI responds.

> I think that series evaluator will take at best hundreds of years to finish on current hardware if `max_degree` is anywhere close to `UINT64_MAX` so the bound for `max_degree` could be much lower without causing problems.

I think that lower bound can be just based on heuristics I guess. Let me know if you some other way. Definitely for future work though.

czgdp1807:
Seems like the test pass. The failures seem to be irrelevant to my fix here.

tylerjereddy:
> The failures seem to be irrelevant to my fix here.

The CI failure is in the same testing module you've added code to:

> special/tests/test_hyp2f1.py::TestHyp2f1::test_test_hyp2f1 - AttributeError: 'complex' object has no attribute 'values'

That's a test that is marked as `slow` because it verifies the `mpmath` outcome with the same arguments against the SciPy one for each new test function added to the class. It appears that, in general, the expectation here is that new test additions in this class will use the `Hyp2f1TestCase` framework to support this.

I'm not sure how closely the `special` folks want that to be followed--in principle I suppose it may be possible to do that here as well (does `mpmath` have the correct extreme value handling behavior already?).

An alternative that works locally is to exclude the newly-added tests from the `mpmath` checking test harness:

<details>

```diff
--- a/scipy/special/tests/test_hyp2f1.py
+++ b/scipy/special/tests/test_hyp2f1.py
@@ -2528,8 +2528,8 @@ class TestHyp2f1:
             if test_method.startswith('test') and
             # Filter properties and attributes (futureproofing).
             callable(getattr(self, test_method)) and
-            # Filter out this test
-            test_method != 'test_test_hyp2f1'
+            # Filter out these tests
+            (test_method not in {'test_test_hyp2f1', 'test_inf_a_b', 'test_large_a_b'})
```

</details>

An argument might be made that rather than starting to balloon the testing class with exceptions to its original design intent, it might be preferable to just make your new tests free-floating `pytest` tests that are not subject to the `mpmath` "meta-testing" of the parent class if it isn't suitable/necessary here.

To iterate locally you can use i.e., `python dev.py test -m full -t scipy/special/tests/test_hyp2f1.py`.

czgdp1807:
@tylerjereddy Instead of your workaround, I created `TestHyp2f1ExtremeInputs` and added my tests there. Before this I tried testing large inputs on `mp_hyp2f1` and seems like `mpmath` cannot handle such inputs (gives, `OverflowError` with `np.inf` inputs and slow convergence error with large values of `a` or `b`). So I decided that we can't test `mpmath` with extreme inputs. Hence I separated my newly added tests in a separate class.



> An argument might be made that rather than starting to balloon the testing class with exceptions to its original design intent

My separate class doesn't touch the design of `TestHyp2f1`. So changes should be safe. 

Let me know if you think otherwise.

tylerjereddy:
Alright, I'm happy with this now. I did a final check that the new tests still fail before and pass after the source patch, and that the two specific examples called out in the matching issue indeed both yield the expected `(nan+nanj)` on this branch.

The bug appears to have been new in the `1.14.x` series--I'm not likely to backport that far back, but I'll add a tentative backport label for the likely `1.15.3` release. I'll squash merge to support backporting.

From the discussion, it sounds like this is a band-aid that Albert is "ok" with, but that more attention is probably warranted to this algorithm/section of code longer term.

Thanks

PR #22814, opened 2025-04-08
BUG: signal.resample: Fix bug for parameter num=2 (including refactor and improved docstr)

### LINKED ISSUE #13788: Documentation for scipy.signal.resample should say what to use with nonperiodic signals
The documentation at https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.resample.html  says that this resampling function assumes the input is periodic. It would be helpful if it also linked to the appropriate function to use for resampling non-periodic signals.

Thanks!

### LINKED ISSUE #13789: Documentation for scipy.signal.decimate doesn't say what to use for non-integer downsampling factor
The documentation at https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.decimate.html   says that the downsampling factor q has to be an integer, but it doesn't really draw any attention to the fact that it only works for integer factors of q. It would be better if it mentioned in the intro paragraph that this method is specifically for downsampling by an integer factor, and if it linked to the appropriate function to use for non-integer factors. (And why you would use this one, since it's limited.)

### LINKED ISSUE #14569: BUG: signal.resample: inconsistency across dtypes
When calling `scipy.signal.resample` with the same real-valued input, but provided as arrays of real and complex type, the returned result is different when the length of the resampled array is 2.

```
import numpy as np


def test(L,L2):
    import scipy.signal
    a = np.random.random(L)
    k = np.random.random(L)
    x=scipy.signal.resample(a,L2,axis=0,domain='time')
    x2=scipy.signal.resample(a.astype(np.complex128),L2,axis=0,domain='time')
    if np.max(np.abs(x-x2)) > 1e-4:
        print("resampling inconsistency for", L, "->", L2)


for l1 in [2,3,4,5,6,10,11]:
    for l2 in [2,3,4,5,6,10,11]:
        test(l1, l2)
```

#### Output:

```
resampling inconsistency for 3 -> 2
resampling inconsistency for 4 -> 2
resampling inconsistency for 5 -> 2
resampling inconsistency for 6 -> 2
resampling inconsistency for 10 -> 2
resampling inconsistency for 11 -> 2
```

1.6.0 1.19.5 sys.version_info(major=3, minor=9, micro=2, releaselevel='final', serial=0)

### LINKED ISSUE #15153: BUG: signal.resample: incorrect with `datetime[ns]` for `t` param
### Describe your issue.

When using `scipy.signal.resample` with a `datetime[ns]` array given for the `t` parameter, the function doesn't resample the times correctly.

For example, the code I've provided for reproduction of this bug will take uniformly spaced datetimes, resample these datetimes (through the `t` parameter of the resample function), then prints the unique time differences in the resampled datetimes.

The output shows that the time stamps for the resampled data are **not** uniformly spaced, and there are periodically large negative values that appear.

### Reproducing Code Example

```python
import scipy.signal
import numpy as np

times = np.arange('2021-02', '2021-03', np.timedelta64(1, 's'), dtype='datetime64[ns]')

_, resampled_times = scipy.signal.resample(np.arange(len(times)), len(times), t=times)

print("Unique time differences:", np.unique(np.diff(times)))
print("Unique time differences after resampling:", np.unique(np.diff(resampled_times)))
# OUTPUT:
# Unique time differences: [1000000000]
# Unique time differences after resampling: [-7624142226235 -7624142226234      999999999     1000000000]
```

### Error message

```shell
N.A.
```

1.7.3 1.19.5 sys.version_info(major=3, minor=7, micro=12, releaselevel='final', serial=0)

DietBru:
In function `resample`:
* Refactor to make code a little bit more accessible.
* Rework of docstr (also Closes #13788).
* Fix indexing bug for parameter `num=2` (Closes #14569 )
* Fix a corner case when parameter `t` is of integer type (Closes #15153)

In `signal.test.tests_signaltools.py`:
* Add unit test for verifying fix of #14569
* Increase coverage of `resample` and `resample_polyphase` to 100%.

Additionally:
* Handle unpaired bins correctly in `signal.envelope()`.
* Improve docstr of `signal.decimate` clarifying what to do with non-integer down-sampling factors (Closes #13789).

### DIFF
--- a/scipy/signal/_signaltools.py
+++ b/scipy/signal/_signaltools.py
@@ -2871,6 +2871,7 @@ def envelope(z: np.ndarray, bp_in: tuple[int | None, int | None] = (1, None), *,
     >>> fg0.subplots_adjust(left=0.08, right=0.97, wspace=0.15)
     >>> plt.show()
     """
+    xp = array_namespace(z)
     if not (-z.ndim <= axis < z.ndim):
         raise ValueError(f"Invalid parameter {axis=} for {z.shape=}!")
     if not (z.shape[axis] > 0):
@@ -2893,12 +2894,12 @@ def envelope(z: np.ndarray, bp_in: tuple[int | None, int | None] = (1, None), *,
                          f"for n={z.shape[axis]=} and {bp_in=}!")
 
     # moving active axis to end allows to use `...` for indexing:
-    z = np.moveaxis(z, axis, -1)
+    z = xp.moveaxis(z, axis, -1)
 
-    if np.iscomplexobj(z):
+    if xp.isdtype(z.dtype, 'complex floating'):
         Z = sp_fft.fft(z)
     else:  # avoid calculating negative frequency bins for real signals:
-        Z = np.zeros_like(z, dtype=sp_fft.rfft(z.flat[:1]).dtype)
+        Z = xp.zeros_like(z, dtype=sp_fft.rfft(z.flat[:1]).dtype)
         Z[..., :n//2 + 1] = sp_fft.rfft(z)
         if bp.start > 0:  # make signal analytic within bp_in band:
             Z[..., bp] *= 2
@@ -2910,8 +2911,8 @@ def envelope(z: np.ndarray, bp_in: tuple[int | None, int | None] = (1, None), *,
         bp_shift = slice(bp.start + n//2, bp.stop + n//2)
         z_bb = sp_fft.ifft(sp_fft.fftshift(Z, axes=-1)[..., bp_shift], n=n_out) * fak
 
-    z_env = np.abs(z_bb) if not squared else z_bb.real ** 2 + z_bb.imag ** 2
-    z_env = np.moveaxis(z_env, -1, axis)
+    z_env = xp.abs(z_bb) if not squared else z_bb.real ** 2 + z_bb.imag ** 2
+    z_env = xp.moveaxis(z_env, -1, axis)
 
     # Calculate the residual from the input bandpass filter:
     if residual is None:
@@ -2926,13 +2927,14 @@ def envelope(z: np.ndarray, bp_in: tuple[int | None, int | None] = (1, None), *,
         else:
             Z[..., bp.start:], Z[..., 0:(n + 1) // 2] = 0, 0
 
-    if np.iscomplexobj(z):  # resample() accounts for unpaired bins:
+    if xp.isdtype(z.dtype, 'complex floating'):  # resample accounts for unpaired bins:
         z_res = resample(Z, n_out, axis=-1, domain='freq')  # ifft() with corrections
     else:  # account for unpaired bin at m//2 before doing irfft():
         if n_out != n and (m := min(n, n_out)) % 2 == 0:
             Z[..., m//2] *= 2 if n_out < n else 0.5
         z_res = fak * sp_fft.irfft(Z, n=n_out)
-    return np.stack((z_env, np.moveaxis(z_res, -1, axis)), axis=0)
+    return xp.stack((z_env, xp.moveaxis(z_res, -1, axis)), axis=0)
+
 
 def _cmplx_sort(p):
     """Sort roots based on magnitude.
@@ -3657,9 +3659,10 @@ def resample(x, num, t=None, axis=0, window=None, domain='time'):
     if domain not in ('time', 'freq'):
         raise ValueError(f"Parameter {domain=} not in ('time', 'freq')!")
 
-    x = np.asarray(x)
+    xp = array_namespace(x, t)
+    x = xp.asarray(x)
     if x.ndim > 1:  # moving active axis to end allows to use `...` in indexing:
-        x = np.swapaxes(x, axis, -1)
+        x = xp.moveaxis(x, axis, -1)
     n_x = x.shape[-1]  # number of samples along the time/frequency axis
     s_fac = n_x / num  # scaling factor represents sample interval dilatation
     m = min(num, n_x)  # number of relevant frequency bins
@@ -3669,15 +3672,15 @@ def resample(x, num, t=None, axis=0, window=None, domain='time'):
         W = None
     elif callable(window):
         W = window(sp_fft.fftfreq(n_x))
-    elif isinstance(window, np.ndarray):
+    elif hasattr(window, 'shape'): # must be an array object
         if window.shape != (n_x,):
             raise ValueError(f"{window.shape=} != ({n_x},), i.e., window length " +
                              "is not equal to number of frequency bins!")
-        W = window.copy()  # we do not want not modify the function's parameters
+        W = xp.asarray(window, copy=True)  # prevent modifying the function parameters
     else:
         W = sp_fft.fftshift(get_window(window, n_x))
 
-    if domain == 'time' and np.isrealobj(x):  # utilize more efficient one-sided FFT:
+    if domain == 'time' and xp.isdtype(x.dtype, 'real floating'):  # utilize rfft():
         X = sp_fft.rfft(x)
         if W is not None:  # fold window, i.e., W1[l] = (W[l] + W[-l]) / 2 for l > 0
             n_X = X.shape[-1]
@@ -3692,7 +3695,7 @@ def resample(x, num, t=None, axis=0, window=None, domain='time'):
         X = sp_fft.fft(x) if domain == 'time' else x
         if W is not None:
             X = X * W  # writing X *= W could modify parameter x
-        Y = np.zeros(X.shape[:-1] + (num,), dtype=X.dtype)
+        Y = xp.zeros(X.shape[:-1] + (num,), dtype=X.dtype)
         Y[..., :m2] = X[..., :m2]  # copy part up to Nyquist frequency
         if m2 < m:  # == m > 2
             Y[..., m2-m:] = X[..., m2-m:]  # copy negative frequency part
@@ -3705,9 +3708,9 @@ def resample(x, num, t=None, axis=0, window=None, domain='time'):
         x_r = sp_fft.ifft(Y / s_fac, n=num, overwrite_x=True)
 
     if x_r.ndim > 1:  # moving active axis back to original position:
-        x_r = np.swapaxes(x_r, -1, axis)
+        x_r = xp.moveaxis(x_r, -1, axis)
     if t is not None:
-        return x_r, t[0] + (t[1] - t[0]) * s_fac * np.arange(num)
+        return x_r, t[0] + (t[1] - t[0]) * s_fac * xp.arange(num)
     return x_r
 
 
--- a/scipy/signal/_signaltools.py
+++ b/scipy/signal/_signaltools.py
@@ -2926,8 +2926,12 @@ def envelope(z: np.ndarray, bp_in: tuple[int | None, int | None] = (1, None), *,
         else:
             Z[..., bp.start:], Z[..., 0:(n + 1) // 2] = 0, 0
 
-    z_res = fak * (sp_fft.ifft(Z, n=n_out) if np.iscomplexobj(z) else
-                   sp_fft.irfft(Z, n=n_out))
+    if np.iscomplexobj(z):  # resample() accounts for unpaired bins:
+        z_res = resample(Z, n_out, axis=-1, domain='freq')  # ifft() with corrections
+    else:  # account for unpaired bin at m//2 before doing irfft():
+        if n_out != n and (m := min(n, n_out)) % 2 == 0:
+            Z[..., m//2] *= 2 if n_out < n else 0.5
+        z_res = fak * sp_fft.irfft(Z, n=n_out)
     return np.stack((z_env, np.moveaxis(z_res, -1, axis)), axis=0)
 
 def _cmplx_sort(p):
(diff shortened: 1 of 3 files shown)


DietBru:
Rendered docs: [resample], [decimate]

[resample]: https://output.circle-artifacts.com/output/job/fe5bb357-984b-4745-99a8-decbcc21ed8a/artifacts/0/html/reference/generated/scipy.signal.resample.html
[decimate]: https://output.circle-artifacts.com/output/job/fe5bb357-984b-4745-99a8-decbcc21ed8a/artifacts/0/html/reference/generated/scipy.signal.decimate.html

scipy/signal/_signaltools.py
```diff
@@ -3500,38 +3506,58 @@ def invresz(r, p, k, tol=1e-3, rtype='avg'):
...
+        here that ``n_x = x.shape[axis]`` specifies the number of samples and ``T`` the
+        sampling interval.
     num : int
-        The number of samples in the resampled signal.
+        The number of samples of the resampled output signal. It may be larger or
+        smaller than ``n_x``.
     t : array_like, optional
-        If `t` is given, it is assumed to be the equally spaced sample
-        positions associated with the signal data in `x`.
+        If `t` is not ``None`` (default), then the timestamps of the resampled signal
```
ev-br, line 3527: Minor, trivial: this got me pause for a moment, to parse whether the default is None or not None. Consider adding the "Default is `None`" at the end instead.

scipy/signal/_signaltools.py
```diff
@@ -3500,38 +3506,58 @@ def invresz(r, p, k, tol=1e-3, rtype='avg'):
...
     window : array_like, callable, string, float, or tuple, optional
-        Specifies the window applied to the signal in the Fourier
-        domain.  See below for details.
-    domain : string, optional
-        A string indicating the domain of the input `x`:
-        ``time`` Consider the input `x` as time-domain (Default),
-        ``freq`` Consider the input `x` as frequency-domain.
+        If not ``None`` (default), specifies a filter in the Fourier domain, which is
+        applied before resampling. I.e., the FFT ``X`` of `x` is calculated by
+        ``X = W * fft(x, axis=axis)``. ``W`` may be interpreted as a sepctral
```
ev-br, line 3538:
```suggestion
        ``X = W * fft(x, axis=axis)``. ``W`` may be interpreted as a spectral
```

scipy/signal/_signaltools.py
```diff
@@ -3500,38 +3506,58 @@ def invresz(r, p, k, tol=1e-3, rtype='avg'):
...
+        windowing function ``W(f_X)`` which consumes the frequencies
+        ``f_X = fftfreq(n_x, T)``.
+
+        If `window` is a 1d array of length `n_x` then ``W=window``.
+        If `window` is a function  then ``W = window(f_X)``.
```
ev-br, line 3543:
```suggestion
        If `window` is a callable  then ``W = window(f_X)``.
```

scipy/signal/_signaltools.py
```diff
@@ -3500,38 +3506,58 @@ def invresz(r, p, k, tol=1e-3, rtype='avg'):
...
+        Otherwise, `window` is passed to `~scipy.signal.get_window`, i.e.,
+        ``W = fftshift(signal.get_window(window, n_x))``.
```
ev-br, line 3545: Just to double-check: is `fftshift` not extraneous here?

scipy/signal/_signaltools.py
```diff
@@ -3540,170 +3566,152 @@ def resample(x, num, t=None, axis=0, window=None, domain='time'):
...
+    ``x.astype(np.complex128)`` produce the same numeric result.
+
+    If the number of input  or output samples are prime or have few prime factors, this
+    function may be slow due to utilizing FFTs. Consult `~scipy.fft.prev_fast_len` and
+    `~scipy.fft.next_fast_len` for determining efficient signals lengths.
+    Alternatively, utilizing `resample_poly` to calculate an intermediate signal (as
+    illustrated in the example below) can result in significant speed increases.
+
+    `resample` is intended to be used for periodic signals with equidistant sampling
+    intervals. For non-periodic signals, `resample_poly` may be a better choice.
```
ev-br, line 3587: Consider stressing this in the See Also link to `resample_poly`.

scipy/signal/_signaltools.py
```diff
@@ -3540,170 +3566,152 @@ def resample(x, num, t=None, axis=0, window=None, domain='time'):
...
-    Consider the following signal  ``y`` where ``len(y)`` is a large  prime number:
-
-    >>> N = 55949
-    >>> freq = 100
-    >>> x = np.linspace(0, 1, N)
-    >>> y = np.cos(2 * np.pi * freq * x)
 
-    Due to ``N`` being prime,
+    The following example illustrates that `~scipy.fft.rfft` / `~scipy.fft.irfft`
+    combination does not always produce the correct resampling result:
```
ev-br, line 3621: Consider 

```suggestion
    combination does not always produce the correct resampling result, while `resample` does:
```

or something to this effect.

scipy/signal/_signaltools.py
```diff
@@ -3540,170 +3566,152 @@ def resample(x, num, t=None, axis=0, window=None, domain='time'):
...
+    >>> x0 = irfft([0, 0, 8, 0, 0], n=n0)  # input signal
+    >>> x1 = resample(x0, num=n1)
+    >>> y1 = irfft(rfft(x0), n=n1) * n1/n0
+    >>> for x_, n_ in zip((x0, x1, y1), ('x0', 'x1', 'y1')):
+    ...    print(f"{n_} = {x_}")
+    x0 = [ 2.  0. -2.  0.  2.  0. -2.  0.]
+    x1 = [ 2. -2.  2. -2.]
+    y1 = [ 1. -1.  1. -1.]
+
+    It can be easily verified that ``x1`` is correct and ``y1`` is not. To produce
```
ev-br, line 3636: Consider showing how to verify it.

If the docstring gets too long, consider moving some material into the tutorial, and linking from here.

scipy/signal/_signaltools.py
```diff
@@ -3540,170 +3566,152 @@ def resample(x, num, t=None, axis=0, window=None, domain='time'):
...
-    >>> f = signal.resample(signal.resample_poly(y, 1, N // num), num)
+    >>> from scipy.fft import irfft, rfft
+    >>> from scipy.signal import resample
+    ...
+    >>> n0, n1 = 8, 4  # number of samples
+    >>> x0 = irfft([0, 0, 8, 0, 0], n=n0)  # input signal
+    >>> x1 = resample(x0, num=n1)
+    >>> y1 = irfft(rfft(x0), n=n1) * n1/n0
+    >>> for x_, n_ in zip((x0, x1, y1), ('x0', 'x1', 'y1')):
+    ...    print(f"{n_} = {x_}")
```
ev-br, line 3631: Optionally, consider simplifying to

```
>>> x0
<copy-paste the output>
>>> x1
<copy-paste the output>
>>> y1
<copy-paste the output>
```

If you prefer the current version for readability, fine.
The suggestion makes it a little friendlier to the docstring checker (it has some logic to parse printed arrays from text, but here it won't manage and will fall back to literal symbol-by-symbol comparison.

scipy/signal/_signaltools.py
```diff
@@ -3540,170 +3566,152 @@ def resample(x, num, t=None, axis=0, window=None, domain='time'):
...
+    >>> n0 = 19937 # number of input samples - prime
+    >>> n1 = 128  # number of output samples - fast FFT length
+    >>> x0 = np.random.rand(n0)  # input signal
+    ...
+    >>> y1 = resample(x0, n1)  # slow due to n0 being prime
+    >>> y1_p = resample(resample_poly(x0, 1, n0 // n1), n1)  # This is faster
 
-    >>> f = signal.resample(y, num)
+    Note that the `y1` and `y1_p` are not identical due to the differences in the
+    utilized antialiasing filter.
```
ev-br, line 3657: This gets me confused a bit. (Maybe it's very very obvious for DSP regulars, but).

Results are not identical, OK. As a user, do I need to worry about it:
Is the difference just fp noise? Is the difference significant? Is there a way to make them identical? Is there actually a need to make them identical?

scipy/signal/_signaltools.py
```diff
@@ -3540,170 +3566,152 @@ def resample(x, num, t=None, axis=0, window=None, domain='time'):
...
+        if window.shape != (n_x,):
+            raise ValueError(f"{window.shape=} != ({n_x},), i.e., window length " +
+                             "is not equal to number of frequency bins!")
+        W = xp.asarray(window, copy=True)  # prevent modifying the function parameters
     else:
-        new_t = np.arange(0, num) * (t[1] - t[0]) * Nx / float(num) + t[0]
-        return y, new_t
+        W = sp_fft.fftshift(get_window(window, n_x))
+
+    if domain == 'time' and xp.isdtype(x.dtype, 'real floating'):  # utilize rfft():
```
ev-br, line 3683: So integer signals go through the two-sided FFT code path, even in time domain, is this intentional?

ev-br (commented):
Looks very reasonable overall from a non-specialist POV.
Cannot comment on fine DSP details, am ready to trust your judgement.

I've left several comments, all of them minor and mostly optional.

scipy/signal/_signaltools.py
```diff
@@ -2893,12 +2894,12 @@ def envelope(z: np.ndarray, bp_in: tuple[int | None, int | None] = (1, None), *,
...
     # moving active axis to end allows to use `...` for indexing:
-    z = np.moveaxis(z, axis, -1)
+    z = xp.moveaxis(z, axis, -1)
 
-    if np.iscomplexobj(z):
+    if xp.isdtype(z.dtype, 'complex floating'):
         Z = sp_fft.fft(z)
     else:  # avoid calculating negative frequency bins for real signals:
-        Z = np.zeros_like(z, dtype=sp_fft.rfft(z.flat[:1]).dtype)
+        Z = xp.zeros_like(z, dtype=sp_fft.rfft(z.flat[:1]).dtype)
```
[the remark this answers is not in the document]
DietBru: IMO, the `fft` function does have some surprising behavior in that regard:
```python
In [1]: import numpy as np

In [2]: from scipy.fft import fft

In [3]: fft(np.array([1,], dtype=np.float32)).dtype
Out[3]: dtype('complex64')

In [4]: fft(np.array([1,], dtype=np.int16)).dtype
Out[4]: dtype('complex128')
```
Furthermore, you can [swap in] your own back-end, if you want.

[swap in]: https://docs.scipy.org/doc/scipy/reference/generated/scipy.fft.set_backend.html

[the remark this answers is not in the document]
ev-br: Ok got it. Best kept as is then, agreed.

scipy/signal/_signaltools.py
DietBru: Well, you should always use an anti-aliasing filter, when resampling. The default for `resample`, i.e. `window=None`, corresponds to passing `np.ones(len(x))`. To replicate the default behavior of `resample_poly` you would have to pass something along the lines of `window=fft(firwin(.., window=('kaiser', 5.0)), n=len(x))`.
To reproduce identical behavior with `resample_poly`, the filter length would have to be same as the signal length, which would defeat the point of using it.

I noticed while rereading that in the snippet above that `resample_poly` is missing the parameter `padtype='wrap' to account for the periodicity of the signal. I'll fix that in the next commit.

scipy/signal/_signaltools.py
DietBru: It is not, since it is used in the code as well. The reason is  that you typically want your spectral window to be centered over the zero frequency bin. For the standard `fft()`, the first entry corresponds to the zero frequency bin, hence `fftshift` is needed.

scipy/signal/_signaltools.py
ev-br: Consider adding this kind of discussion to the tutorial and linking from this docstring.

scipy/signal/_signaltools.py
DietBru: Fixed in commit 76636f9e7

scipy/signal/_signaltools.py
DietBru: Reworked example in commit 76636f9e7.

scipy/signal/_signaltools.py
DietBru: Good point. Fixed in commit 76636f9e7.

scipy/signal/_signaltools.py
DietBru: Added in commit 76636f9e7.

scipy/signal/_signaltools.py
DietBru: Good catch! Addressed in commit 76636f9e7.

scipy/signal/_signaltools.py
DietBru: In commit 76636f9e7, I added a pointer to the relationship between the `window` parameter and the anti-aliasing filter.

I have never thought about how to explain within two or three paragraphs what is going on spectrally when resampling. This topic deems me to demand at least two to three pages in a well written text book. Unfortunately, I am not aware of any good English references (learned this stuff too long ago...). In summary, I am unsure how I would approach this in a tutorial. I am very open to suggestions for references or to approaches for concise explanations.

DietBru:
Thanks a lot for the detailed review, @ev-br! I hope I addressed all your points.

Rendered docs: [resample], [decimate]

[resample]: https://output.circle-artifacts.com/output/job/2db5a33b-bd5e-4b92-8f31-fd4de048a07e/artifacts/0/html/reference/generated/scipy.signal.resample.html
[decimate]: https://output.circle-artifacts.com/output/job/2db5a33b-bd5e-4b92-8f31-fd4de048a07e/artifacts/0/html/reference/generated/scipy.signal.decimate.html

scipy/signal/_signaltools.py
ev-br: I've to admit  https://github.com/scipy/scipy/pull/22814#discussion_r2039204850 is clearer than either of the docstring additions, to me at least. Consider just this explanation somewhere, here or in a follow-up.
I don't have more specific suggestions though, so up to you.

ev-br (approved):
Looks good from a non-specialist perspective.

Let's keep this open for a short while, in a hope that one of DSP experts looks through the PR. 
Otherwise, I think it's good to go in.

scipy/signal/_signaltools.py
DietBru: Hmm, it's interesting how subjective perspective differ. What my explanation are lacking, is i) what the purpose of the anti-aliasing filter serves and ii) what criteria it needs to fulfill. They key (of course, from my biased point of view) lies in understanding how sampled spectra are periodic with an interval width of 1 / sampling interval...  Would a link to this [Wikipedia section] on sampling & DFT be in some way enlightening (imhoi, it's slightly besides the key points)? A more complete treatment can be found in Sec. 2.4 and 2.5 of  [Vetterli's book] (manuscript is online—imho, it's a bit too compact for newcomers)

Would it be a possibility, if you would formulate in a new issue what you think is worthwhile explaining in a tutorial section. The process of resampling, aliasing, ...?

[Wikipedia section]: https://en.wikipedia.org/wiki/Discrete-time_Fourier_transform#Sampling_the_DTFT
[Vetterli's book]: http://waveletsandsubbandcoding.org/

DietBru:
:partying_face: Merge whenever you think it's appropriate. Thanks again for your helpful input, @ev-br!

lucascolley:
> Looks good from a non-specialist perspective.
> 
> Let's keep this open for a short while, in a hope that one of DSP experts looks through the PR.

If you have some time @endolith @larsoner !

scipy/signal/_signaltools.py
```diff
@@ -3500,210 +3506,230 @@ def invresz(r, p, k, tol=1e-3, rtype='avg'):
...
+    ...
+    >>> y1 = resample(x0, n1)  # slow due to n0 being prime
+    >>> # This is faster:
+    >>> y1_p = resample(resample_poly(x0, 1, n0 // n1, padtype='wrap'), n1)
+
+    Note that the `y1` and `y1_p` are not identical due to the differences in the
+    utilized antialiasing filter in each function. In both functions the antialiasing
+    filter can be customized by modifying the `window` parameter, though `resample`
+    interprets it as a spectral window function, whereas `resample_poly` assumes it to
+    be some form of impulse response.
```
larsoner, line 3675: I would say something like this... the "some form of impulse response" is what caught my eye as being unclear. The purpose in `resample_poly` is to perform the antialiasing. If you pass a bad antialiasing filter, you risk aliasing there. There is no risk of aliasing in `resample` since the ideal brick-wall filter is always implicitly applied by the method itself.
```suggestion
    filter can be customized by modifying the `window` parameter. `resample`
    interprets it as a *spectral* window function that is always applied in addition to
    the implicit ideal (brick wall) antialiasing filter. `resample_poly` applies
    `window` as a FIR filter that gets applied in the *time* domain and should be
    designed appropriately to avoid aliasing.
```

larsoner (commented):
I like the docstring updates! The code changes are extensive but I had a look at them and they looked okay, and I think we can trust the existing tests here. Just a couple of comments/ideas for the docs then I think we're good to go

ev-br:
gentle ping @DietBru -- would you be open to addressing Eric's comments in this PR or a follow-up, or you two are going to speak some DSP and decide to not implement his suggestions?

DietBru:
Sorry for not being able to finish this sooner. In commit [b3c1993](https://github.com/scipy/scipy/pull/22814/commits/b3c199349fd5cc424dbbaa77f3b615881156a666), I added two additional plots to the resample examples. I hope this addresses all your review remarks, @larsoner and perhaps address some aspects of this [comment], @ev-br.


Rendered docs: [resample], [decimate]

[comment]: https://github.com/scipy/scipy/pull/22814#discussion_r2037119941
[resample]: https://output.circle-artifacts.com/output/job/609d1aa8-ab9c-4da7-9076-ef7843219850/artifacts/0/html/reference/generated/scipy.signal.resample.html
[decimate]: https://output.circle-artifacts.com/output/job/609d1aa8-ab9c-4da7-9076-ef7843219850/artifacts/0/html/reference/generated/scipy.signal.decimate.html

scipy/signal/_signaltools.py
DietBru: Addressed in commit [b3c1993](https://github.com/scipy/scipy/pull/22814/commits/b3c199349fd5cc424dbbaa77f3b615881156a666)

ev-br:
Let's land this. Thanks @DietBru , @larsoner !

Issue #22412, opened 2025-01-27
RFC/API: move `clarkson_woodruff_transform` from `linalg` to `sparse.linalg`?


lucascolley:
While investigating gh-22382, I found that `linalg` has a dependency on `sparse` via `clarkson_woodruff_transform`, which uses `csc_matrix` for computation (introduced in https://github.com/scipy/scipy/pull/9884).

Since `scipy.fft` has a dependency on `scipy.linalg` via `scipy.special`, this introduces a transitive dependency on `scipy.sparse`, which accounts for ~20% of `scipy.fft`'s import time on my machine.

Would it be acceptable to move the function under `sparse.linalg`?

cc @perimosocordiae @rgommers @ilayn

lucascolley:
Actually, I think there is a much easier fix here. I think we can just use lazy loading at https://github.com/scipy/scipy/blob/abcbaec60d654f6362731216eb025db2a0d37e40/scipy/special/_orthogonal.py#L179
which breaks the `special`->`linalg` import dependency

ilayn:
We also had a small discussion here and following comments https://github.com/scipy/scipy/issues/21466#issuecomment-2583587737

lucascolley:
> moving the import into `_gen_roots_and_weights` seems like the right thing to do here.

Would that be equivalent to `import scipy` and calling `scipy.linalg.eigvals_banded`?

rgommers:
Equivalent but worse I'd say. If you need `linalg`, better to import it (explicit is better than implicit). Also, tools like `tach` won't understand dynamic imports hidden with `__getattr__`.

PR #22815, opened 2025-04-08
MAINT: sparse: add lazy loading for csgraph and linalg

### LINKED ISSUE #22382: ENH: sparse.spmatrix: allow fast import

j-bowhay:
#### Reference issue

closes #22382

Adds lazy loading for the two submodules in order to speed up import times.

### DIFF
--- a/scipy/sparse/_matrix.py
+++ b/scipy/sparse/_matrix.py
@@ -49,8 +49,9 @@ class spmatrix:
 
     # Restore matrix power
     def __pow__(self, power):
-        import scipy
-        return scipy.sparse.linalg.matrix_power(self, power)
+        from .linalg import matrix_power
+
+        return matrix_power(self, power)
 
     ## Backward compatibility
 
--- a/scipy/sparse/_matrix.py
+++ b/scipy/sparse/_matrix.py
@@ -49,9 +49,8 @@ class spmatrix:
 
     # Restore matrix power
     def __pow__(self, power):
-        from .linalg import matrix_power
-
-        return matrix_power(self, power)
+        import scipy
+        return scipy.sparse.linalg.matrix_power(self, power)
 
     ## Backward compatibility
 
--- a/scipy/sparse/__init__.py
+++ b/scipy/sparse/__init__.py
@@ -299,6 +299,7 @@ sorted indices are required (e.g., when passing data to other libraries).
 # Nathan Bell, and Jake Vanderplas.
 
 import warnings as _warnings
+import importlib as _importlib
 
 from ._base import *
 from ._csr import *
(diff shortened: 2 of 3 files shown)


dschult (commented):
This change looks like what we want. 
- `csgraph will no longer be imported at the time sparse is imported; only when `sparse.csgraph` is accessed.
- `linalg` will be auto-imported when accessed (it currently is not imported with sparse)

I'm not sure about the details of `tach.toml`.  Why does `linalg` stay in the "depends_on" for `scipy.sparse`? Does something in `sparse` depends on `linalg`? If so does `linalg` get imported when importing `sparse` when using "tach"?

dschult (approved):
That's a nice way to import matrix_power and allow lazy loading. Thanks!

This looks good to me!

lucascolley:
> For whatever reason tach didn't like the style of one of the imports. I tweaked it slightly which now allows us to remove both csgraph and linalg from the  tach.toml.

I would suggest reverting that change. It is true that `sparse` depends on `sparse.linalg` for `spmatrix.matrix_power`, even though the dependency is not at import-time.

One could argue that keeping track of import-time dependencies only is more useful than dependencies in general, but I think at this stage it is helpful to be explicit about potential runtime gotchas.

lucascolley:
> The original complaint was about import-time so I think we should focus on that.

Sure, but I could imagine a similar issue about runtime instead, so I don't think either way is favoured. Hence I think explicit rather than implicit is better.

> If we also use also track runtime dependencies then it looses its "sharpness" as I could add sparse.linalg as an import-time dependency without complaint

There should be other ways to achieve this (or better) sharpness. [Reading the docs](https://docs.gauge.sh/usage/interfaces#example%3A-exclusive-interface), it looks like we could add something like:

```toml
[[interfaces]]
expose = ["matrix_power"]
from = ["scipy.sparse.linalg"]
visibility = ["scipy.sparse"]
exclusive = true
```

I'm running into some errors when trying to get recent versions of `tach` to run locally, though

scipy/sparse/_matrix.py
```diff
@@ -49,9 +49,8 @@ def __rmul__(self, other):
 
     # Restore matrix power
     def __pow__(self, power):
-        from .linalg import matrix_power
-
-        return matrix_power(self, power)
+        import scipy
+        return scipy.sparse.linalg.matrix_power(self, power)
```
lucascolley, line 53: I don't mean to block the PR as tach-bikeshedding is insignificant compared to the performance improvement, but as another data-point, Ralf was against this pattern at https://github.com/scipy/scipy/issues/22412#issuecomment-2615500392

lucascolley (approved):
thanks Jake, and for the review Dan!

PR #22817, opened 2025-04-08
fix: "bca" method to "BCa"

lga-zurich:
#### Reference issue
None

I believe the 'BCa' method name is being used inconsistently and breaks the default invocation.

#### Additional information
This change seems to fix the computation of confidence intervals in Unitxt [GlobalMetric](https://github.com/IBM/unitxt/blob/main/src/unitxt/metrics.py#L1004)


lga-zurich:
Hi @mdhaber 
Ultimately, I would just like the default "method" parameter to match the "if" statement on line 634, so the default invocation works. If that is "bca" so be it, I don't mind, but it should match, unless I'm missing something.

lga-zurich:
I just saw that the "method" should be set to lowercase in bootstrap_iv ... sorry for the noise!

PR #22820, opened 2025-04-09
MAINT: signal: consolidate `order_filter` tests

ev-br:
For some historic reasons, `order_filter` tests were split between two test modules, with some tests hiding in `test_filter_design.py`.

Thus, parametrize these previously hidden tests across the Array API backends, and move them to `test_signaltools.py`.

Two commits are on purpose: one commit adds backend testing, the other one moves tests.


ilayn (approved):

lucascolley (approved):
thanks Evgeni and for the review Ilhan!

PR #18799, opened 2023-06-30
ENH: simplify `gausshyper.pdf`

dschmitz89:
#### Reference issue
Part of #10106 and #17832
 
#### What does this implement/fix?
There is no analytical expression available for the gausshyper distribution PPF. To make things worse, implementing the CDF is not possible without high precision codes for Appel's hypergeometric functions (see #18056). That means that the default random variate generation goes through numerical integration to calculate the CDF and then inverts it via a rootfinder. At least the quadrature integration can be speeded up by providing a lowlevel callable of the PDF in cython.

I pretty much blindly copied the approach from the genhyperbolic distribution implemented in #18134.

#### Additional information

I did not further check the accuracy of the PDF in the tails, here I focussed on performance only. Locally, this passed the generic distribution tests for me at least. Going via the logpdf could be done in a follow up.

Benchmark:

```python
import numpy as np
from scipy import stats
from time import time

a, b, c, z = 1.5, 2.5, 2, 0

gh = stats.gausshyper(a=a, b=b, c=c, z=z)

x = np.linspace(0, 1, 1000)
t0 = time()
cdf_vals = gh.cdf(x)
print(f"Time for 1000 CDF values: {time() - t0}")

t0 = time()
samples = gh.rvs(size=1000)
print(f"Time for 1000 samples: {time() - t0}")
``` 

*This PR*:
Time for 1000 CDF values: 0.03378629684448242
Time for 1000 samples: 1.2594482898712158

*Main*:
Time for 1000 CDF values: 0.8763303756713867
Time for 1000 samples: 8.12756633758545

Sampling is still not "fast" but at least faster. Maybe with a vectorized rootfinder (#17719), the PPF calculation will become much faster in future.

PR #22823, opened 2025-04-10
MAINT: integrate.tanhsinh: simplify error estimate

mdhaber:
#### Reference issue
gh-22681
https://github.com/scipy/scipy/pull/22690#issuecomment-2767402826

This simplifies the `tanhsinh` error estimate as discussed in https://github.com/scipy/scipy/pull/22690#issuecomment-2767402826.

#### Additional information
Marking as draft pending additional benchmarking.

### DIFF
--- a/scipy/integrate/_tanhsinh.py
+++ b/scipy/integrate/_tanhsinh.py
@@ -100,7 +100,7 @@ def tanhsinh(f, a, b, *, args=(), log=False, maxlevel=None, minlevel=2,
         the result dtype), respectively. Must be non-negative and finite if
         `log` is False, and must be expressed as the log of a non-negative and
         finite number if `log` is True. Iteration will stop when
-        ``res.error < atol`` or  ``res.error < res.integral * rtol`.
+        ``res.error < atol`` or  ``res.error < res.integral * rtol``.
     preserve_shape : bool, default: False
         In the following, "arguments of `f`" refers to the array ``xi`` and
         any arrays within ``argsi``. Let ``shape`` be the broadcasted shape
--- a/scipy/integrate/_tanhsinh.py
+++ b/scipy/integrate/_tanhsinh.py
@@ -97,13 +97,10 @@ def tanhsinh(f, a, b, *, args=(), log=False, maxlevel=None, minlevel=2,
     atol, rtol : float, optional
         Absolute termination tolerance (default: 0) and relative termination
         tolerance (default: ``eps**0.75``, where ``eps`` is the precision of
-        the result dtype), respectively.  Iteration will stop when
-        ``res.error < atol`` or  ``res.error < res.integral * rtol``. The error
-        estimate is as described in [1]_ Section 5 but with a lower bound of
-        ``eps * res.integral``. While not theoretically rigorous or
-        conservative, it is said to work well in practice. Must be non-negative
-        and finite if `log` is False, and must be expressed as the log of a
-        non-negative and finite number if `log` is True.
+        the result dtype), respectively. Must be non-negative and finite if
+        `log` is False, and must be expressed as the log of a non-negative and
+        finite number if `log` is True. Iteration will stop when
+        ``res.error < atol`` or  ``res.error < res.integral * rtol`.
     preserve_shape : bool, default: False
         In the following, "arguments of `f`" refers to the array ``xi`` and
         any arrays within ``argsi``. Let ``shape`` be the broadcasted shape
@@ -174,6 +171,12 @@ def tanhsinh(f, a, b, *, args=(), log=False, maxlevel=None, minlevel=2,
     Implements the algorithm as described in [1]_ with minor adaptations for
     finite-precision arithmetic, including some described by [2]_ and [3]_. The
     tanh-sinh scheme was originally introduced in [4]_.
+
+    Two error estimation schemes are described in [1]_ Section 5: one attempts to
+    detect and exploit quadratic convergence; the other simply compares the integral
+    estimates at successive levels. While neither is theoretically rigorous or
+    conservative, both work well in practice. Our error estimate uses the minimum of
+    these two schemes with a lower bound of ``eps * res.integral``.
 
     Due to floating-point error in the abscissae, the function may be evaluated
     at the endpoints of the interval during iterations, but the values returned by
@@ -771,8 +774,8 @@ def _estimate_error(work, xp):
         d4 = work.d4
         d5 = log_e1 + xp.real(work.Sn)
         temp = xp.where(d1 > -xp.inf, d1 ** 2 / d2, -xp.inf)
-        ds = xp.stack([temp, 2 * d1, d3, d4, d5])
-        aerr = xp.minimum(d1, xp.max(ds, axis=0))
+        ds = xp.stack([temp, 2 * d1, d3, d4])
+        aerr = xp.clip(xp.max(ds, axis=0), d5, d1)
         rerr = aerr - xp.real(work.Sn)
     else:
         # Note: explicit computation of log10 of each of these is unnecessary.
@@ -782,8 +785,8 @@ def _estimate_error(work, xp):
         d4 = work.d4
         d5 = e1 * xp.abs(work.Sn)
         temp = xp.where(d1 > 0, d1**(xp.log(d1)/xp.log(d2)), 0)
-        ds = xp.stack([temp, d1**2, d3, d4, d5])
-        aerr = xp.minimum(d1, xp.max(ds, axis=0))
+        ds = xp.stack([temp, d1**2, d3, d4])
+        aerr = xp.clip(xp.max(ds, axis=0), d5, d1)
         rerr = aerr/xp.abs(work.Sn)
 
     return rerr, aerr
--- a/scipy/integrate/_tanhsinh.py
+++ b/scipy/integrate/_tanhsinh.py
@@ -765,18 +765,24 @@ def _estimate_error(work, xp):
         # complex values have imaginary part in increments of pi*j, which just
         # carries sign information of the original integral, so use of
         # `xp.real` here is equivalent to absolute value in real scale.
-        ones = xp.ones_like(work.Sn)
-        d1 = xp.real(special.logsumexp(xp.stack([work.Sn, Snm1]),
-                                       b=xp.stack([ones, -ones]), axis=0))
+        d1 = xp.real(special.logsumexp(xp.stack([work.Sn, Snm1 + work.pi*1j]), axis=0))
+        d2 = xp.real(special.logsumexp(xp.stack([work.Sn, Snm2 + work.pi*1j]), axis=0))
+        d3 = log_e1 + xp.max(xp.real(work.fjwj), axis=-1)
+        d4 = work.d4
         d5 = log_e1 + xp.real(work.Sn)
-        ds = xp.stack([d1, d5])
+        temp = xp.where(d1 > -xp.inf, d1 ** 2 / d2, -xp.inf)
+        ds = xp.stack([temp, 2 * d1, d3, d4, d5])
         aerr = xp.max(ds, axis=0)
         rerr = aerr - xp.real(work.Sn)
     else:
         # Note: explicit computation of log10 of each of these is unnecessary.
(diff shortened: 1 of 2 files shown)


mdhaber:
I benchmarked on the [818 problem set](https://github.com/boostorg/math/issues/706#issuecomment-943722047). The following are with default settings.

For integration of a single, scalar function, wall clock time of this PR tends to be a little faster (execution time gmean is 86.5%):
image

However, this PR usually requires twice as many function evaluations. That's 510 problems that require about twice as many function evaluations, 276 problems that require the same number of function evaluations, and a handful of other cases.
image

This suggested that when function evaluations are somewhat costly and execution time is limited by actual operations rather than overhead, main might be a little faster. So I tried the [`gausshyper.cdf`](https://github.com/scipy/scipy/pull/18799#issuecomment-1615084118) problem:

<details>

```python3
import numpy as np
from scipy import stats
from scipy.integrate import tanhsinh
from time import perf_counter
from mpmath import mp
mp.dps = 50

a, b, c, z = 1.5, 2.5, 2, 0

def _pdf(x):
    Cinv = sc.gamma(a) * sc.gamma(b) / sc.gamma(a + b) * sc.hyp2f1(c, a, a + b, -z)
    return 1.0 / Cinv * x ** (a - 1.0) * (1.0 - x) ** (b - 1.0) / (1.0 + z * x) ** c

def _pdf_mpmath(x):
    a, b, c, z = 1.5, 2.5, 2, 0
    a, b, c, z = mp.mpf(a), mp.mpf(b), mp.mpf(c), mp.mpf(z)
    Cinv = mp.gamma(a) * mp.gamma(b) / mp.gamma(a + b) * mp.hyp2f1(c, a, a + b, -z)
    return mp.one / Cinv * x ** (a - mp.one) * (mp.one - x) ** (b - mp.one) / (mp.one + z * x) ** c

x = np.linspace(0, 1, 10000)[1:]
tic = perf_counter()
res = tanhsinh(_pdf, 0, x[1:])
toc = perf_counter()
t1 = toc-tic
print(f"Time for 1000 CDF value (main): {t1}")
# Time for 1000 CDF value (pr): 0.0061378999962471426
# Time for 1000 CDF value (main): 0.004455100002815016

ref = np.asarray([mp.quad(_pdf_mpmath, (mp.mpf(0), mp.mpf(xi))) for xi in x[1:]], dtype=np.float64)

rerr = abs((res.integral - ref)/(ref))
print(rerr)
```

</details>

Indeed, for that problem, the PR terminates after 131 function evaluations whereas main needs only 67, and main is faster[^1]: 
```
# Time for 1000 CDF value (pr): 0.0061378999962471426
# Time for 1000 CDF value (main): 0.004455100002815016
```

[^1]: Of course, the PR is a little more accurate, essentially always agreeing with `mpmath` within 2 ULPs (whereas the worst-case error in main is on the order of 1e-13), but that's a bit overkill given the default `rtol`.

The actual error tends the same or a little better (as we might expect from more function evaluations):
image

But maybe more relevant is the number of problems solved to within a specified (log10) relative error (with default tolerances). The default relative tolerance is supposed to be ~`-11.74`.
```
           PR Solved  Main Solved
Tolerance                        
-15              521          410
-14              678          593
-13              707          669
-12              717          708
-11              719          717
-10              722          722
-9               728          729
-8               767          764
-7               787          787
-6               794          794
-5               798          798
-4               802          802
```

Something else that is important is the accuracy of the error estimate. This histogram shows the log10 of the "error ratio", that is, the actual error over the reported error. A conservative integator will have actual error smaller than reported error (negative).

image

So this PR tends to report the error a little more accurately, but it looks like it ends up being a bit less conservative in the estimate (which is a little surprising, but not impossible). 

---

Conclusions? Well, it turns out that the error estimate that we're using _does_ seem to work as it is intended to - often detecting the convergence early and terminating with half as many function evaluations. This doesn't always translate to faster wall clock execution, but probably _does_ for the slowest problems - large arrays, costly function evaluations. I hate to say it, but we might want to keep it. Something to try, however, would be to add a factor of `h` to `d3` and `d4` (https://github.com/scipy/scipy/issues/22681#issuecomment-2767554114). What do you think of that @steppi?

steppi:
Ah, that makes sense. Since $h$ gets cut in half in each step, then saving one iteration will halve the number of function evaluations.

> Conclusions? Well, it turns out that the error estimate that we're using _does_ seem to work as it is intended to - often detecting the convergence early and terminating with half as many function evaluations.

Right, the error estimate assumes quadratic convergence and will converge one step earlier than the simplified estimate unless the other estimates show an obstruction to quadratic convergence.

> I hate to say it, but we might want to keep using that error estimate. Some things to try, however, would be to add a factor of `h` to `d3` and `d4` ([#22681 (comment)](https://github.com/scipy/scipy/issues/22681#issuecomment-2767554114)). Or maybe just _remove_ `d3` and `d4`, just to see what happens? What do you think

I have concerns about multiplying by a factor of `h` and making the other estimates tighter, because then they may no longer work to test for obstructions to quadratic convergence. I haven't thought about how these estimates might work to check for such obstructions, so I'd rather leave them alone for now and assume Bailey knew what he was doing.

We don't want to remove them either for the same general reason. To best follow the spirit of the paper, which converges based on an error estimate $10^d$ for

$$d = \max\left(\frac{d_1^2}{d_2}, 2d_1, d_3, d_4\right)$$

or otherwise when $S_n$ and $S_{n-1}$ agree up to all but the last few digits

I think what could do is instead use the error estimate $d^{'} = \min(10^{d_1}, 10^{d})$

In code, (where we don't think in terms of decimal digits and leave out the $\log_{10}$'s and $10^x$'s), I think this would look like

```python
 d1 = xp.abs(work.Sn - Snm1)
        d2 = xp.abs(work.Sn - Snm2)
        d3 = e1 * xp.max(xp.abs(work.fjwj), axis=-1)
        d4 = work.d4
        d5 = e1 * xp.abs(work.Sn)
        temp = xp.where(d1 > 0, d1**(xp.log(d1)/xp.log(d2)), 0)
        ds = xp.stack([temp, d1**2, d3, d4, d5])
        aerr = xp.min(d1, xp.max(ds, axis=0))
        rerr = aerr/xp.abs(work.Sn)
```

mdhaber:
Ok. I thought we would test those modifications, hoping these functions would be representative enough, but I understand if you're more comfortable keeping the authors' recommendation(s) intact.

Taking the minimum of the two is a really good idea. (With `xp.minimum`, though.) Performance can only be better, and we're still as conservative as the authors would suggest. I'll implement that.

scipy/integrate/_tanhsinh.py
```diff
@@ -772,7 +772,7 @@ def _estimate_error(work, xp):
         d5 = log_e1 + xp.real(work.Sn)
         temp = xp.where(d1 > -xp.inf, d1 ** 2 / d2, -xp.inf)
         ds = xp.stack([temp, 2 * d1, d3, d4, d5])
```
mdhaber, line 774: Oops, maybe I need to do:
```suggestion
        ds = xp.stack([temp, 2 * d1, d3, d4])
```
and below:
```python3
aerr = xp.maximum(xp.minimum(d1, xp.max(ds, axis=0)), d5)
```
?

I also need to update the description of the error estimate.

scipy/integrate/tests/test_tanhsinh.py
```diff
@@ -752,8 +752,13 @@ def test_gh_22681_finite_error(self, xp):
         b = complex(12, 39)
         def f(t):
             return xp.sin(a * (1 - t) + b * t)
-        res = _tanhsinh(f, xp.asarray(0.), xp.asarray(1.), atol=0, rtol=0, maxlevel=10)
-        assert xp.isfinite(res.error)
+        ref = _tanhsinh(f, xp.asarray(0.), xp.asarray(1.), atol=0, rtol=0, maxlevel=10)
+        assert xp.isfinite(ref.error)
+        # Previously, the tanhsinh would not detect convergence
+        res = _tanhsinh(f, xp.asarray(0.), xp.asarray(1.))
```
mdhaber, line 758: *This* wouldn't pass in main, but does pass in this PR.
```suggestion
        res = _tanhsinh(f, xp.asarray(0.), xp.asarray(1.), rtol=1e-14)
```

mdhaber (commented):
@steppi Supposing I adjust as shown here, I would remove this description of `res.error` from the description of the `atol` and `rtol` parameters:

> The error estimate is as described in [1]_ Section 5 but with a lower bound of ``eps * res.integral``. While not theoretically rigorous or conservative, it is said to work well in practice.

and add to the description of `res.error`:

> Two error estimation schemes are described in [1]_ Section 5: one attempts to detect and exploit quadratic convergence; the other simply compares the integral estimates at successive levels. While neither is theoretically rigorous or conservative, both work well in practice. Our error estimate uses the minimum of these two schemes with a lower bound of ``eps * res.integral``.

scipy/integrate/_tanhsinh.py
steppi: Oh, yeah. I didn't think of that. That sounds right.

mdhaber: I think I can use `xp.clip` to simplify, though.

steppi:
Nice, this looks good to me now. Is it ready for review?

mdhaber:
Yeah, I was just waiting for CI to finish.

I'm not sure if the CircleCI failure is real. I can't see what went wrong:
image

Can you?

mdhaber:
What does line 74 looks like in the CircleCI output?

steppi:
> What does line 74 looks like in the CircleCI output?

```
    73          **atol, rtol** : float, optional
    74              Absolute termination tolerance (default: 0) and relative termination
    75              tolerance (default: ``eps**0.75``, where ``eps`` is the precision of
    76              the result dtype), respectively. Must be non-negative and finite if
    77              `log` is False, and must be expressed as the log of a non-negative and
    78              finite number if `log` is True. Iteration will stop when
    79              ``res.error < atol`` or  ``res.error < res.integral * rtol`.
```

mdhaber:
Missing backtick at the end.

steppi:
> Missing backtick at the end.

Oh yeah, on line 79. It's weird that it complained about line 74.

scipy/integrate/_tanhsinh.py
```diff
@@ -97,13 +97,10 @@ def tanhsinh(f, a, b, *, args=(), log=False, maxlevel=None, minlevel=2,
...
-        ``res.error < atol`` or  ``res.error < res.integral * rtol``. The error
-        estimate is as described in [1]_ Section 5 but with a lower bound of
-        ``eps * res.integral``. While not theoretically rigorous or
-        conservative, it is said to work well in practice. Must be non-negative
-        and finite if `log` is False, and must be expressed as the log of a
-        non-negative and finite number if `log` is True.
+        the result dtype), respectively. Must be non-negative and finite if
+        `log` is False, and must be expressed as the log of a non-negative and
+        finite number if `log` is True. Iteration will stop when
+        ``res.error < atol`` or  ``res.error < res.integral * rtol`.
```
mdhaber, line 103:
```suggestion
        ``res.error < atol`` or  ``res.error < res.integral * rtol``.
```

scipy/integrate/tests/test_tanhsinh.py
```diff
@@ -236,7 +237,8 @@ def test_integral_transforms(self, limits, ref, xp):
         assert (xp.isdtype(logres.integral.dtype, "real floating") if ref > 0
                 else xp.isdtype(logres.integral.dtype, "complex floating"))
 
-        xp_assert_close(xp.exp(logres.error), res.error, atol=1e-16, check_dtype=False)
+        atol = xp.finfo(res.error.dtype).eps
```
mdhaber, line 240:
```suggestion
        atol = 2 * xp.finfo(res.error.dtype).eps
```

steppi (approved):
Looks good. Thanks @mdhaber. I'll merge after CI settles.

steppi:
Failure is unrelated, due to a slow test (that I wrote and should have marked `slow` it seems).

mdhaber:
Thanks. Would you mark that? I've seen it fail as slow elsewhere, too.

PR #22827, opened 2025-04-11
ENH: stats: add `marray` support to a few remaining functions

mdhaber:
#### Reference issue
gh-22490

Tests `marray` support for `stats.gstd` (no modifiications required).
Adds support to `stats.directional_stats` (see inline comment).
Adds support to `stats.bartlett` (see inline comments).
Adds support to `stats.variation` (see inline comments).
Adds support to `stats.pearsonr`/`stats.pointbiserialr` (see inline comments).
Add support to `stats.entropy`.

#### Additional information
Easier to review with "Hide whitespace".
`stats.differential_entropy` looks more challenging; saving for a separate PR.


scipy/stats/_morestats.py
```diff
@@ -4386,6 +4387,12 @@ def directional_stats(samples, *, axis=0, normalize=True):
         raise ValueError("samples must at least be two-dimensional. "
                          f"Instead samples has shape: {tuple(samples.shape)}")
     samples = xp.moveaxis(samples, axis, 0)
+
+    if is_marray(xp):
+        _xp = array_namespace(samples.mask)
+        mask = _xp.any(samples.mask, axis=-1, keepdims=True)
+        samples = xp.asarray(samples.data, mask=mask)
```
mdhaber, line 4394: @dschmitz89 Do you have a sense for what the preferred behavior would be when, say, only one element of a row is masked? 

The behavior here propagates the mask to all elements of the row (observation). This matches the behavior of ignoring the entire observation.

It's unclear whether another option is available. Suppose there is only one masked element in the entire array. If we ignore *just* the originally masked element, I think this would have have exactly the same affect on the other components as if the masked element were an unmasked zero. (The effect is a little different on the component that has a masked entry.) That doesn't seem quite right.

scipy/stats/_morestats.py
```diff
@@ -2901,15 +2904,20 @@ def bartlett(*samples, axis=0):
     if k < 2:
         raise ValueError("Must enter at least two input sample vectors.")
 
-    samples = _broadcast_arrays(samples, axis=axis, xp=xp)
-    samples = [xp.moveaxis(sample, axis, -1) for sample in samples]
+    if axis is None:
+        samples = [xp_ravel(sample) for sample in samples]
```
mdhaber, line 2908: Really, this is needed for any array-API backend. It is taken care of by the `_axis_nan_policy` decorator for NumPy, and eventually will probably be done for the other backends, too. But for now, it's simple to add it here.

scipy/stats/_morestats.py
```diff
@@ -2901,15 +2904,20 @@ def bartlett(*samples, axis=0):
...
+    else:
+        samples = _broadcast_arrays(samples, axis=axis, xp=xp)
+        samples = [xp.moveaxis(sample, axis, -1) for sample in samples]
 
-    Ni = [xp.asarray(sample.shape[-1], dtype=sample.dtype) for sample in samples]
+    Ni = [xp.asarray(_length_nonmasked(sample, axis=-1, xp=xp), dtype=sample.dtype)
```
mdhaber, line 2913: This is a standard generalization for adding marray support.

scipy/stats/_morestats.py
```diff
@@ -2901,15 +2904,20 @@ def bartlett(*samples, axis=0):
...
+          for sample in samples]
     Ni = [xp.broadcast_to(N, samples[0].shape[:-1]) for N in Ni]
     ssq = [xp.var(sample, correction=1, axis=-1) for sample in samples]
     Ni = [arr[xp.newaxis, ...] for arr in Ni]
     ssq = [arr[xp.newaxis, ...] for arr in ssq]
     Ni = xp.concat(Ni, axis=0)
+    Ni = xpx.at(Ni)[Ni == 0].set(xp.nan)
```
mdhaber, line 2920: This is the only uncommon correction for masked arrays. Previously, empty arrays would produce a *masked* variance, which gets *omitted* by `xp.sum`. We introduce a NaN that can't be ignored when there are zero non-masked elements.

mdhaber (commented):
Thoughts about adjustments to `bartlett`.

scipy/stats/_variation.py
```diff
@@ -96,33 +99,25 @@ def variation(a, axis=0, nan_policy='propagate', ddof=0, *, keepdims=False):
...
-    if ddof == n:
-        # Another special case.  Result is either inf or nan.
-        std_a = xp.std(a, axis=axis, correction=0)
-        result = xp.where(std_a > 0, xp.copysign(xp.inf, mean_a), NaN)
-        return result[()] if result.ndim == 0 else result
-
-    with np.errstate(divide='ignore', invalid='ignore'):
-        std_a = xp.std(a, axis=axis, correction=ddof)
-        result = std_a / mean_a
+        axis = -1
```
mdhaber, line 106: No need to change this.
```suggestion
        axis = 0
```

scipy/stats/_variation.py
```diff
@@ -96,33 +99,25 @@ def variation(a, axis=0, nan_policy='propagate', ddof=0, *, keepdims=False):
...
+
+    n = xp.asarray(_length_nonmasked(a, axis=axis), dtype=a.dtype)
```
mdhaber, line 108: When `n` is an array (masked arrays), we can't have `if` statements that depend on `n`, so this needed to be rewritten. Fortunately, it's a very simple function. If we ignore the warnings, there is only one special case.

scipy/stats/_variation.py
```diff
@@ -96,33 +99,25 @@ def variation(a, axis=0, nan_policy='propagate', ddof=0, *, keepdims=False):
...
+
+    with (np.errstate(divide='ignore', invalid='ignore'), warnings.catch_warnings()):
+        warnings.simplefilter("ignore")
+        mean_a = xp.mean(a, axis=axis)
+        std_a = xp.std(a, axis=axis)
+        correction = (n / (n - ddof))**0.5  # we may need uncorrected std below
+        result = std_a * correction / mean_a
+
+    def special_case(std_a, mean_a):
+        return xp.where(std_a > 0, xp.copysign(xp.inf, mean_a), xp.nan)
```
mdhaber, line 118: See `if ddof == n` special case above.

scipy/stats/_stats_py.py
```diff
@@ -4236,33 +4236,35 @@ def _pearsonr_fisher_ci(r, n, confidence_level, alternative):
     """
     xp = array_namespace(r)
 
-    with np.errstate(divide='ignore'):
-        zr = xp.atanh(r)
-
     ones = xp.ones_like(r)
     n = xp.asarray(n, dtype=r.dtype)
     confidence_level = xp.asarray(confidence_level, dtype=r.dtype)
-    if n > 3:
```
mdhaber, line 4245: This is just more of the usual generalization to array `n`.

mdhaber:
Marking as draft to avoid merge, but I think this is ready for review. The failures are due to https://github.com/data-apis/array-api-extra/discussions/196#discussioncomment-12816449; maybe we'll be able to fix, or maybe we'll just xfail the Dask tests for now.

scipy/stats/_morestats.py
dschmitz89: I don't have a strong opinion but the implemented policy would work for me as a user.

scipy/_lib/_array_api.py
```diff
@@ -616,11 +617,31 @@ def xp_default_dtype(xp):
         return xp.float64
 
 
+### MArray Helpers ###
 def is_marray(xp):
     """Returns True if `xp` is an MArray namespace; False otherwise."""
     return "marray" in xp.__name__
 
 
+def _length_nonmasked(x, axis, keepdims=False, xp=None):
```
mdhaber, line 626: Copied from `_stats_py` for easier use in other modules.

dschmitz89:
@mdhaber are you happy with this PR besides the merge conflict? I am a bit concerned with the number of necessary test skips for the different array backends, but I guess these are the necessary growing pains during the whole array API adoption.

mdhaber:
Yup, I think this is alright. I wouldn't be too concerned about the skips - I think some will go away once I merge some PRs on the `marray` side, and some are probably due to issues with the backends. NumPy + MArray is the initial goal, and the rest will follow in time.

dschmitz89 (approved):

PR #22830, opened 2025-04-13
ENH: sparse.csgraph.laplacian: add `signed_graph_variant` parameter and support 'opposing' and 'unsigned'

### LINKED ISSUE #22685: ENH: generalize laplacian matrix to signed networks

The function `scipy.sparse.csgraph.laplacian` currently computes node degrees by summing (possibly signed) edge weights:

`w = np.asarray(m.sum(axis=axis)).ravel() - m.diagonal()`

This works well for unsigned graphs, but in signed graphs, negative edges can cancel out positive ones, causing nodes with many connections to appear as if they have near-zero degree. This leads to issues in spectral analysis and clustering on signed networks, as noted in Kunegis et al. (2010).

Introduce an option (e.g., `signed=False` or `use_abs=False`) that, when set to True, uses the absolute values of the edge weights when computing the degree:

```
if use_abs:
    w = np.asarray(np.abs(m).sum(axis=axis)).ravel() - np.abs(m.diagonal())
else:
    w = np.asarray(m.sum(axis=axis)).ravel() - m.diagonal()
```
This change generalizes the Laplacian to signed graphs without altering the current default behavior for unsigned graphs. Otherwise, one could simply use the absolute value without introducing an extra option, as this change does not affect the results in unsigned graphs. 

Overwriting the Laplacian locally: Monkey-patching the function, but this is not ideal or maintainable.
Writing a separate Laplacian function: A custom solution external to SciPy could address signed graphs, but having it in SciPy benefits the community directly and ensures consistency with other graph utilities.

Reference: Kunegis, Jérôme, et al. "Spectral analysis of signed graphs for clustering, prediction and visualization." Proceedings of the 2010 SIAM International Conference on Data Mining, SIAM, 2010.
Why This Matters: Properly accounting for negative edges is necessary when analyzing social networks, biological data with inhibitory interactions, or correlation matrices.

math-hiyoko:
#### Reference issue
Closes #22685 

This PR adds a new `signed_graph_variant` option to `scipy.sparse.csgraph.laplacian` to improve support for signed networks (signed_graph_variant='standard').

- `signed_graph_variant='standard'` is standard degree calculation using edge weights.
- `signed_graph_variant='opposing'` computes the degree using absolute value only for the degree computation, so that positive and negative weights cannot cancel out.
- `signed_graph_variant='unsigned'` computes both degree and adjacency using absolute value, discarding all sign information.

This change follows the proposal in #22685 and is based on the approach discussed in [Kunegis et al. (2010)](https://www.researchgate.net/publication/220906703_Spectral_Analysis_of_Signed_Graphs_for_Clustering_Prediction_and_Visualization)
Detailed discussions on the reasoning behind each type of variant can be found in [Shi et al. (2019)](https://doi.org/10.1137/17M1134172)

The default behavior (`signed_graph_variant='standard'`) remains unchanged to preserve backward compatibility.

#### Additional information
When both `signed_graph_variant='opposing'` and `symmetrized=True`, the symmetrization is applied *after* taking the absolute value of the input matrix. This ensures that the computed degrees reflect the magnitude of all connections symmetrically, which is particularly relevant for signed adjacency matrices where negative values represent inhibitory relationships.


scipy/sparse/csgraph/_laplacian.py
```diff
@@ -20,6 +20,7 @@ def laplacian(
     form="array",
     dtype=None,
     symmetrized=False,
+    use_abs=False,
```
j-bowhay, line 23: Just a drive by comment: if we're adding a new argument, let's make it keyword only
```suggestion
    *,
    use_abs=False,
```

math-hiyoko:
Everything is ready now

lucascolley:
@fernandodiazdiaz would you like to take a look?

fernandodiazdiaz:
> @fernandodiazdiaz would you like to take a look?

Looks perfect to me!

j-bowhay:
Thanks @math-hiyoko, would you be able to change the `use_abs` to what is suggested in https://github.com/scipy/scipy/issues/22685#issuecomment-2813538308

math-hiyoko:
Just to confirm—I’ll add a new variant parameter to sparse.csgraph.laplacian, defaulting to "default", and "use_abs" will activate the functionality implemented in this PR.

math-hiyoko:
I’ve updated the PR to replace the boolean use_abs parameter with the new variant parameter

scipy/sparse/csgraph/tests/test_graph_laplacian.py
```diff
@@ -176,42 +184,71 @@ def _check_laplacian_dtype(
...
 @pytest.mark.parametrize("use_out_degree", [True, False])
+@pytest.mark.parametrize("variant", ["default", "use_abs"])
 def test_asymmetric_laplacian(use_out_degree, normed,
-                              copy, dtype, arr_type):
+                              copy, dtype, arr_type, variant):
     # adjacency matrix
-    A = [[0, 1, 0],
-         [4, 2, 0],
-         [0, 0, 0]]
+    A = [[0, 2, -2],
```
dschult, line 191: This is changing the tests that were already in place. Is there a reason to do that?

dschult (commented):
Thanks for this @math-hiyoko 

I'd like other folks take on having the string `"default"` be the default variant. We could use `None`, or `"default"` or `"degree"` (or some other descriptive name). It seems like the string "default" doesn't convey a meaning -- beyond the fact that it is being used as the default variant -- which we already know because it is in the function signature. Using `None` provides easy/quick look-up with `variant is None`. And a descriptive word provides meaning and allows potential sqitching of the default without switching the name of the default variant.  I think `"degree"` is best of these 3 options, but maybe there is a better one-word description of what the standard laplacian uses to fill the diagonal.

My second comment is that the changes to the tests add new tests, but also change the existing tests (see comment below). Is there a reason you chose to change the current arrays that are tested? (Are these better for the original laplacian as well as for the new variant we are introducing?)

scipy/sparse/csgraph/tests/test_graph_laplacian.py
math-hiyoko: The `use_abs` variant specifically addresses cases where edge weights can be negative. The original test cases did not include negative edge weights, making them unsuitable for testing the behavior of the `use_abs` variant.

math-hiyoko: This PR also adds the method="use_abs" case to this test. Now both cases are being tested. 
https://github.com/math-hiyoko/scipy/blob/bf8bcb89b7020028f753398cae2f36a2b2d451ad/scipy/sparse/csgraph/tests/test_graph_laplacian.py#L225-L251

dschult: Maybe I'm not understanding something. The test is filled with if statements for each case and sets matrices `A` and vector `d`.  The cases that have `method="default"` dont have anything to do with `use_abs`.

In fact, maybe all the different cases could be separated nicely using pytest fixtures. But that's not a blocker for this PR.

I would like that the existing tests are still run (unless there is a reason they shouldn't be). Adding new ones is fine. But I'd prefer not to take remove existing tests.

math-hiyoko: Understood—I'll restore the original test cases and have them coexist alongside the new tests introduced in this PR.

dschult (commented):
- We need the docs to have a reference for Kunegis et al (2010).
- We should get a better name than `use_abs`. What should this variant be called?
- see below for comments about duplicate/unneeded computation.

Can you check that the formula is right for this `use_abs` variant? 
I thought it should be L = |A|.sum() - |A|   but you have it as L = |A|.sum() - A.
The latter is strange for a laplacian because the rows don't sum to zero (that is the usual eigenvector of all ones is not an eigenvector). That is a fairly universal property of matrices that are called laplacians.

fernandodiazdiaz:
Perhaps I can be of some assistance with the mathematical details. I can confirm that the correct expression is L=|A|.sum() - A. Indeed, this Laplacian has the unusual property that its columns do not add up to one. This is the price to pay to have a positive semidefinite operator. There are some authors that use instead L=A.sum() - A. This alternative choice has rows that add up to zero, but it ceases to be positive semidefinite, and therefore the associated dynamical systems may exhibit unstable behaviors. When both Laplacians are compared, the first one is typically called the "opposing Laplacian", while the second one is the "repelling Laplacian". Let me emphasize that the object L=|A|.sum() - |A| is another Laplacian that is not suited for the analysis of signed networks, as it removes all the information contained in the sign. I typically refer to it as the "unsigned Laplacian", although there is not a single name for it in the literature. 

A more detailed discussion on the reasoning behind each type of Laplacian can be found here : Shi, Guodong, Claudio Altafini, and John S. Baras. "Dynamics over signed networks." Siam review 61.2 (2019): 229-257.

dschult:
Thanks for this comment @fernandodiazdiaz 
Do you have insight into when one might use the laplacian variant `A.sum() + A`? Or the normalized laplacians like: `I - (1 / np.sqrt(A.sum()) * A * (1 / np.sqrt(A.sum()) )`?  There seem to be so many different variants!

This says to me that we definitely need a better name for the variant than `use_abs` -- other laplacian definitions use abs is other ways.  What is a good name for this variant?

fernandodiazdiaz:
The Laplacian variant A.sum() + A is typically referred to as the "signless Laplacian". Paradoxically, its main application is when A is unsigned, where it is used to get some insight into the system's bipartivity. For example, a network is bipartite if and only if the singless Laplacian is singular, and the eigenvector corresponding to the null eigenvalue provides the partition of the node set that makes all edges fall between nodes in different subsets. 

With respect to the normalized Laplacians, their use is completely analogous to that of the unsigned case. For instance, one can define a random walk opposing Laplacian L_rw = I-|A|.sum()^{-1}A, which governs the behavior of a random walk on a signed network. When using these normalizations, the degree matrix must be computed using the absolute value; otherwise one risks divisions by zero. 

I agree that 'use_abs' may be too vague for some users, as different Laplacians use the absolute value. A more concrete option could be 'type': {'opposing', 'repelling', 'unsigned'}, where 'unsigned' corresponds to the default.

dschult:
The normalized cases are analogous to the unnormalized cases. But do we want this function to be able to provide the normalized cases too by choosing a variant? Or is it better to have this function only  provide unnormalized cases and have the user create normalized versions from the unnormalized results. Your example of `I-|A|.sum()^{-1}A` makes it seem like the user could create this normalized version from the unnormalized by multiplying by `|A|.sum()^{-1}`. And I guess the example `I - A.sum()^(-1/2) * A * A.sum()^(-1/2)` would require pre and post multiplication by `D=A.sum()^(-1/2)`. 

Are there normalized versions that won't be easily created from the unnormalized versions?  I'm just trying to get a feel for how many variants we'll need in the long term.

math-hiyoko:
> - We need the docs to have a reference for Kunegis et al (2010).
> - We should get a better name than use_abs. What should this variant be called?
> - see below for comments about duplicate/unneeded computation.
>
> Can you check that the formula is right for this use_abs variant?
> I thought it should be L = |A|.sum() - |A| but you have it as L = |A|.sum() - A.
> The latter is strange for a laplacian because the rows don't sum to zero (that is the usual eigenvector of all ones is not an eigenvector). That is a fairly universal property of matrices that are called laplacians.

@dschult 

Thanks for the comments!
I’ve addressed the first and third points:
- Added the reference to Kunegis et al. (2010) in the docs
- Removed the duplicate/unnecessary computation based on the variant setting.

math-hiyoko:
@fernandodiazdiaz @dschult 
As part of this PR, would it make sense to rename the current default variant to "unsigned" and use_abs to "opposing", and additionally introduce a new "repelling" variant?
Please let me know if this direction sounds good.

math-hiyoko:
@fernandodiazdiaz @dschult
I’ve gone through Shi, Altafini & Baras (2019) “Dynamics over Signed Networks” to make sure I fully understood the mathematical distinctions.
From that reading it’s clear that:
- the Laplacian already in main is the repelling variant (L = D – A, degrees from signed weights), so changing its behaviour would break existing code;

I’ve therefore left the current implementation as variant="repelling" for backward-compatibility and added fully-tested implementations of "opposing" and "unsigned", plus docstrings, references and unit tests for each.

Let me know if anything should be tweaked further!

math-hiyoko:
fixed some tests and docs

math-hiyoko:
Addresses an environment-dependent bug.

scipy/sparse/csgraph/_laplacian.py
```diff
@@ -69,6 +70,19 @@ def laplacian(
         sparse matrices unless the sparsity pattern is symmetric or
         `form` is 'function' or 'lo'.
         Default: False, for backward compatibility.
+    variant : 'repelling', 'opposing' or 'unsigned'
+        Specifies the method used to compute vertex degrees.
```
dschult:
```suggestion
        Specifies the method used to compute vertex degrees.
        All variants are the same in the case of non-negative edge weights.
```

dschult (commented):
For the "unsigned" variant we replace the edge weights with their absolute value. And in some places, the PR overwrites the values in the graph object to avoid memory allocation `np.abs(m.data, out=m.data)`.  Does this change the input graph? Should we warn users about this? Or do we make a copy before the code ever gets to this spot?

I think this is a reasonable approach for capturing all the variants we have discussed.
:)

math-hiyoko:
> For the "unsigned" variant we replace the edge weights with their absolute value. And in some places, the PR overwrites the values in the graph object to avoid memory allocation np.abs(m.data, out=m.data). Does this change the input graph? Should we warn users about this? Or do we make a copy before the code ever gets to this spot?

With the current implementation, we do not overwrite the input graph’s data in-place—the original adjacency matrix remains unchanged. This behavior is explicitly verified by the _check_symmetric_graph_laplacian helper (exercised via test_symmetric_graph_laplacian and test_sparse_formats), which asserts that when `copy=True` the input stays intact.
https://github.com/math-hiyoko/scipy/blob/5bcfdcd47a23c510ae00641ab15a7065f945a9fa/scipy/sparse/csgraph/tests/test_graph_laplacian.py#L68-L75

dschult:
> With the current implementation, we do not overwrite the input graph’s data in-place

I see that those tests check are supposed to check for in-place changes.
But this example shows it doesn't always work. 
```python
import scipy as sp;import numpy as np
mat=np.array([[0,-1,1],[-1,0,-1],[1,-1,0]])
spmat=sp.sparse.coo_array(mat)
spmat_copy=spmat.copy()

# works fine
sp.sparse.csgraph.laplacian(spmat)
assert np.array_equal(spmat.data, spmat_copy.data)

# "unsigned" variant sets data to abs(data) using the `out=` input to `np.abs`
sp.sparse.csgraph.laplacian(spmat, variant="unsigned")
assert np.array_equal(spmat.data, spmat_copy.data)  # Fails
```
Maybe the unsigned variant isn't being tested with those tests?

math-hiyoko:
Thank you for catching this.
I’ve updated the tests to validate the behavior of the copy parameter when edge weights are negative, and I’ve also corrected the laplacian implementation accordingly.

math-hiyoko:
Thank you for catching this.
I’ve updated the tests to validate the behavior of the copy parameter when edge weights are negative, and I’ve also corrected the laplacian implementation accordingly.

scipy/sparse/csgraph/_laplacian.py
```diff
@@ -69,6 +70,20 @@ def laplacian(
...
+        All variants are the same in the case of non-negative edge weights.
+        Detailed discussions on the reasoning behind each type of variant
+        can be found in [2].:
+
+        * 'repelling' is standard degree calculation using edge weights.
+        * 'opposing' computes the degree using absolute values of edge weights.
+          This prevents cancellation of positive and negative edge weights in
+          signed graphs, as described in [3].
+        * 'unsigned' discards the sign of edge weights.
+          This removes all the information contained in the sign.
```
dschult: This description talks about calculation of degree values. But not the sign used in the definition of L itself. I think this will be confusing for someone coming to this function who has never heard of or cares about signed graphs. 

Do I understand correctly that in addition to the treatment of signs when computing degree, that some of these variants compute `L=D-A` and others compute `L=D+A`?  And maybe it would help if this description started with some indication that this keyword handles variants for signed graphs. That will help folks who are using standard edge weights figure out that they don't need to worry about this keyword arg.

scipy/sparse/csgraph/_laplacian.py
```diff
@@ -271,6 +298,57 @@ def laplacian(
...
+    ... )
+    >>> L_opposing
+    array([[ 2, -1, 1],
+           [-1,  2, -1],
+           [1, -1,  2]])
+    >>> d_opposing
+    array([2, 2, 2])
+
+    Setting variant="unsigned" discards the sign of edge weights
+    and computes the Laplacian using absolute edge weights
```
dschult: It's not clear in this sentence what "computes the Laplacian using absolute edge weights" refers to. Knowing the answer I must guess that "computes the Laplacian" refers to handling the adjacency matrix part of the laplacian formula. But computing the laplacian involves both computing degrees and combining the degrees with the adjacency info. Let's either find better words to use, use formulas, or define terms for the degree handling and for the adjacency handling and use them.

dschult (commented):
I think the words "repelling" "opposing" and "unsigned" are all from the signed graph literature. Most people coming to this function will not be using signed graphs. So we want a clear path for them to use/ignore the variant keyword.  One way to do this would be to rename the keyword to something like: `signed_graph_variant`. That's long and ugly but contains the right information.  We could instead leave the name `variant` and make the default name be something like `standard` and use the text to explain that the signed graph literature calls this `repelling`. 

We also should make sure each description (in keyword description and examples) explains both the treatment of signs in degree and treatment of the sign of adjacency matrix in the formula `L = D - A`. We either need to define the words we are using or use formulas. 

In the end I'd like the signed graph part of the laplacian function to clearly be a special case, and have the words used to describe the options be clearly defined and complete in their description.

math-hiyoko:
@dschult 
Thank you for the feedback!
Based on your suggestions, I’ve renamed the parameter to signed_graph_variant (defaulting to standard), updated the docstring to include the explicit formulas for each variant, and verified the changes by running spin docs.

dschult (approved):
Thanks @math-hiyoko For the changes here.
This looks ready to be merged to me.

It keeps the current features of `csgraph.laplacian` and adds features for signed graphs. The docs are clear that for people not using signed graphs, they can just ignore that keyword and for people using signed graphs they can see the options described clearly with citations for more information. The additional tests look good.

I approve this PR.

scipy/sparse/csgraph/_laplacian.py
```diff
@@ -69,6 +70,28 @@ def laplacian(
         sparse matrices unless the sparsity pattern is symmetric or
         `form` is 'function' or 'lo'.
         Default: False, for backward compatibility.
+    signed_graph_variant : 'standard', 'opposing' or 'unsigned'
+        This option only has an effect when the graph contains negative edge weights;
+        it does nothing for graphs with all non-negative weights.
+        The Laplacian is always assembled as :math:`L = D - A`,
+        but D and A are replaced as described below.
+        See [2] for a detailed discussion of each variant:
```
j-bowhay:
```suggestion
        See [2]_ for a detailed discussion of each variant:
```

scipy/sparse/csgraph/_laplacian.py
```diff
@@ -69,6 +70,28 @@ def laplacian(
...
+
+        * 'standard': No modification of edge weights when computing D or A.
+          This is called "repelling" in [2].
```
j-bowhay:
```suggestion
          This is called "repelling" in [2]_.
```

scipy/sparse/csgraph/_laplacian.py
```diff
@@ -69,6 +70,28 @@ def laplacian(
...
+            - :math:`D_i = \\sum_j G_{ij}`
+            - :math:`A_{ij} = G_{ij}`
+        * 'opposing': Absolute value only for the degree computation,
+          so that positive and negative weights cannot cancel out.
+          see [3] for a discussion of this variant.
```
j-bowhay:
```suggestion
          see [3]_ for a discussion of this variant.
```

scipy/sparse/csgraph/_laplacian.py
```diff
@@ -115,6 +138,18 @@ def laplacian(
     References
     ----------
     .. [1] Laplacian matrix. https://en.wikipedia.org/wiki/Laplacian_matrix
+    .. [2] Shi, Guodong and Altafini, Claudio and Baras, John S. (2019).
+           Dynamics over Signed Networks,
+           SIAM Review 61(2), 229-257
+           https://doi.org/10.1137/17M1134172
```
j-bowhay:
```suggestion
           :doi:`10.1137/17M1134172`
```

scipy/sparse/csgraph/_laplacian.py
```diff
@@ -115,6 +138,18 @@ def laplacian(
...
+    .. [3] Jérôme Kunegis and Stephan Schmidt and Andreas Lommatzsch
+           and Jürgen Lerner and Ernesto W. De Luca
+           and Sahin Albayrak. (2010).
+           Spectral Analysis of Signed Graphs for Clustering,
+           Prediction and Visualization.
+           In Proceedings of the 2010 SIAM International Conference
+           on Data Mining (SDM), 559-570.
+           https://doi.org/10.1137/1.9781611972801.49
```
j-bowhay:
```suggestion
           :doi:`10.1137/1.9781611972801.49`
```

scipy/sparse/csgraph/_laplacian.py
```diff
@@ -271,6 +306,59 @@ def laplacian(
...
+
+    >>> G = np.array([[0, 1, -1],
+    ...               [1, 0,  1],
+    ...               [-1, 1, 0]])
+    >>> G
+    array([[ 0,  1, -1],
+           [ 1,  0,  1],
+           [-1,  1,  0]])
+
+    Setting `signed_graph_variant="standard"` (default),
```
j-bowhay:
```suggestion
    Setting ``signed_graph_variant="standard"`` (default),
```

scipy/sparse/csgraph/_laplacian.py
```diff
@@ -271,6 +306,59 @@ def laplacian(
...
+
+    >>> L_standard, d_standard = csgraph.laplacian(G, return_diag=True)
+    >>> L_standard
+    array([[ 0, -1,  1],
+           [-1,  2, -1],
+           [ 1, -1,  0]])
+    >>> d_standard
+    array([0, 2, 0])
+
+    Setting `signed_graph_variant="opposing"` prevents cancellation
```
j-bowhay:
```suggestion
    Setting ``signed_graph_variant="opposing"`` prevents cancellation
```

scipy/sparse/csgraph/_laplacian.py
```diff
@@ -271,6 +306,59 @@ def laplacian(
...
+    ...     signed_graph_variant="opposing",
+    ... )
+    >>> L_opposing
+    array([[ 2, -1, 1],
+           [-1,  2, -1],
+           [1, -1,  2]])
+    >>> d_opposing
+    array([2, 2, 2])
+
+    Setting `signed_graph_variant="unsigned"` uses absolute value
```
j-bowhay:
```suggestion
    Setting ``signed_graph_variant="unsigned"`` uses absolute value
```

scipy/sparse/csgraph/_laplacian.py
```diff
@@ -271,6 +306,59 @@ def laplacian(
...
+    for both degree and adjacency, discarding all sign information.
+    This is same as
+    `csgraph.laplacian(np.abs(G), return_diag=True, signed_graph_variant="standard")`:
```
j-bowhay:
```suggestion
    ``csgraph.laplacian(np.abs(G), return_diag=True, signed_graph_variant="standard")``:
```

scipy/sparse/csgraph/_laplacian.py
```diff
@@ -404,19 +493,40 @@ def _linearoperator(mv, shape, dtype):
...
+        # Normalization affects only the diagonal computation, not the rest.
+        # abs before symmetrize (for opposing), seems implied by Kunegis et al (2010).
+        graph_abs = np.abs(graph)
+        graph_diagonal_abs = np.abs(graph_diagonal)
+        diag = np.asarray(graph_abs.sum(axis=axis)).ravel() - graph_diagonal_abs
+        if symmetrized:
+            diag += np.asarray(graph_abs.sum(axis=1 - axis)).ravel() \
+                    - graph_diagonal_abs
+    else:
+        raise ValueError(f"Invalid signed_graph_variant: {signed_graph_variant!r}")
```
j-bowhay, line 535: Can you add test coverage for invalid input

j-bowhay (commented):
Thanks, some minor comments

j-bowhay:
Also CI failure is related

scipy/sparse/csgraph/_laplacian.py
math-hiyoko: Thank you for your comments
edited existing test `test_format_error_message`
https://github.com/math-hiyoko/scipy/blob/06bbab850289af4b8b1630a833ae861a9d40237d/scipy/sparse/csgraph/tests/test_graph_laplacian.py#L526-L531

j-bowhay:
CI failures look related, do you mind taking a look

math-hiyoko:
~The remaining failing test is the issue of #23412 so not related to this PR.~
This seems to have been fixed in #23381

j-bowhay (changes requested):
Sorry for drawing this out even longer but there seems to be a slight issue, the function is documented to work with sparse arrays however the new argument can break this:
```
In [20]: G = np.array([[0, 1, -1], [1, 0,  1], [-1, 1, 0]])

In [21]: G = sparse.dok_array(G)

In [22]: laplacian(G, form="lo")
Out[22]: <3x3 _CustomLinearOperator with dtype=int64>

In [23]: laplacian(G, signed_graph_variant="unsigned", form="lo")
AttributeError                            Traceback (most recent call last)
Cell In[23], line 1
----> 1 laplacian(G, signed_graph_variant="unsigned", form="lo")

File ~/development/scipy/scipy/sparse/csgraph/_laplacian.py:453, in laplacian(csgraph, normed, return_diag, use_out_degree, copy, form, dtype, symmetrized, signed_graph_variant)
    445     create_lap = (
    446         _laplacian_sparse_flo
    447         if issparse(csgraph)
    448         else _laplacian_dense_flo
    449     )
    451 degree_axis = 1 if use_out_degree else 0
--> 453 lap, d = create_lap(
    454     csgraph,
    455     normed=normed,
    456     axis=degree_axis,
    457     copy=copy,
    458     form=form,
    459     dtype=dtype,
    460     symmetrized=symmetrized,
    461     signed_graph_variant=signed_graph_variant,
    462 )
    463 if is_pydata_sparse:
    464     lap = pydata_sparse_cls.from_scipy_sparse(lap)

File ~/development/scipy/scipy/sparse/csgraph/_laplacian.py:513, in _laplacian_sparse_flo(***failed resolving arguments***)
    509     dtype = graph.dtype
    511 if signed_graph_variant == "unsigned":
    512     # Discard the sign of edge weights.
--> 513     np.abs(graph.data, out=graph.data)
    515 graph_sum = np.asarray(graph.sum(axis=axis)).ravel()
    516 if symmetrized:

AttributeError: 'dok_array' object has no attribute 'data'
```

math-hiyoko:
@j-bowhay 
Thank you for pointing that out!
I’ve added handling for sparse matrices in both DOK and LIL formats.

math-hiyoko:
Hi @j-bowhay,  I’ve made the requested changes.
Could you take another look when you get a chance? Thanks!

j-bowhay:
Apologies @math-hiyoko for the delay, I still need to find the time to look at this carefully.

dschult:
I merged with main again to pull in some recent fixes in main.
Still some failures apparently unrelated to this PR. @math-hiyoko I just wanted to warn you that you should pull from upstream main to bring this merge to your local repo. :)

tylerjereddy:
Looks like a core dev still wants to do a detailed review here per https://github.com/scipy/scipy/pull/22830#issuecomment-3288875235 and it seems unlikely this will cross the line before we branch in a day or two, so I'll bump the milestone.

tylerjereddy:
The concern from https://github.com/scipy/scipy/pull/22830#issuecomment-3619181816 is still valid and this hasn't seen activity apart from merging in `main` in the last six months, so I'll bump the milestone with branching in ~10 days (the PR is still "only" a year old, so maybe there's still some hope to get a maintainer to take a closer look).

PR #20868, opened 2024-06-03
DEP: interpolate: deprecate complex dtypes in `{Akima1D, Pchip}Interpolator`

j-bowhay:
follow up to https://github.com/scipy/scipy/pull/19752, scheduled for this release

PR #22832, opened 2025-04-13
DOC: interpolate: remove outdated deprecation notices

j-bowhay:
The actual deprecation happened as scheduled but this notices snuck through


mdhaber (approved):
Agreed, relevant PR gh-20868.

PR #21233, opened 2024-07-22
ENH: `stats.boxcox_llf`: vectorize for n-D arrays

lucascolley:
#### Reference issue
Towards gh-20544

#### What does this implement/fix?
Adds native `axis` support.

PR #22835, opened 2025-04-14
MAINT: stats.boxcox_llf: refactor for simplicity

mdhaber:
#### Reference issue
gh-21233

During review of gh-21233, I noticed some simplifications now possible after updates to `logsumexp` and `array-api-compat`. This implements those simplifications.

### DIFF
--- a/scipy/stats/_morestats.py
+++ b/scipy/stats/_morestats.py
@@ -1006,6 +1006,7 @@ def _boxcox_llf(data, axis=0, *, lmb):
         logvar = _log_var(logx, xp, axis) - 2 * math.log(abs(lmb))
 
     res = (lmb - 1) * xp.sum(logdata, axis=axis) - N/2 * logvar
+    res = xp.astype(res, data.dtype, copy=False)  # compensate for NumPy <2.0
     res = res[()] if res.ndim == 0 else res
     return res
 
--- a/scipy/stats/_morestats.py
+++ b/scipy/stats/_morestats.py
@@ -864,15 +864,11 @@ def _log_mean(logx, axis):
 
 def _log_var(logx, xp, axis):
     # compute log of variance of x from log(x)
-    logmean = _log_mean(logx, axis=axis)
-    # get complex dtype with component dtypes same as `logx` dtype;
-    dtype = xp.result_type(logx.dtype, 1j)
-    pij = xp.full(logx.shape, pi * 1j, dtype=dtype)
-    logxmu = special.logsumexp(xp.stack((logx, logmean + pij)), axis=0)
-    return (
-        xp.real(xp.asarray(special.logsumexp(2 * logxmu, axis=axis)))
-        - math.log(logx.shape[axis])
-    )
+    logmean = xp.broadcast_to(_log_mean(logx, axis=axis), logx.shape)
+    ones = xp.ones_like(logx)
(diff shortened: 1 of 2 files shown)


scipy/stats/_morestats.py
```diff
@@ -864,15 +864,11 @@ def _log_mean(logx, axis):
...
-    logxmu = special.logsumexp(xp.stack((logx, logmean + pij)), axis=0)
-    return (
-        xp.real(xp.asarray(special.logsumexp(2 * logxmu, axis=axis)))
-        - math.log(logx.shape[axis])
-    )
+    logmean = xp.broadcast_to(_log_mean(logx, axis=axis), logx.shape)
+    ones = xp.ones_like(logx)
+    logxmu, _ = special.logsumexp(xp.stack((logx, logmean), axis=0), axis=0,
+                                  b=xp.stack((ones, -ones), axis=0), return_sign=True)
+    return special.logsumexp(2 * logxmu, axis=axis) - math.log(logx.shape[axis])
```
mdhaber, line 871: Previously, we implemented subtraction by adding $\pi j$ to the `logmean`, and we took the absolute value by looking at the real of the result. Here, we use a weight of -1 on the `logmean`, and we take the absolute value by ignoring the sign. (Passing `return_sign=True` is required, though; otherwise, elements with negative sign would be NaN.)

scipy/stats/_morestats.py
```diff
@@ -1010,7 +1006,6 @@ def _boxcox_llf(data, axis=0, *, lmb):
         logvar = _log_var(logx, xp, axis) - 2 * math.log(abs(lmb))
 
     res = (lmb - 1) * xp.sum(logdata, axis=axis) - N/2 * logvar
-    res = xp.astype(res, data.dtype)
```
mdhaber, line 1013: I think this was only needed due to old `xp.sum` dtype behavior.

scipy/stats/_morestats.py
```diff
@@ -991,7 +987,7 @@ def boxcox_llf(lmb, data, *, axis=0, keepdims=False, nan_policy='propagate'):
                           result_to_tuple=lambda x: (x,))
 def _boxcox_llf(data, axis=0, *, lmb):
     xp = array_namespace(data)
-    data = xp_promote(data, force_floating=True, xp=xp)
+    lmb, data = xp_promote(lmb, data, force_floating=True, xp=xp)
```
mdhaber, line 990: Typically, `lmb` will be a Python scalar, but this will help with NumPy scalars and 0-d arrays.

mdhaber:
Dtype failures are the usual NumPy <2.0 gotchas. I'll try to debug later on another machine.

scipy/stats/_morestats.py
mdhaber: Not so.
`copy=False` was removed in gh-21233. Was that intentional? Added it back, because it seems fine not to copy for NumPy 2.0+.

mdhaber:
Failures are unrelated (`interpolate`/`spatial`) and probably temporary.

scipy/stats/_morestats.py
lucascolley: > Was that intentional?

may well have been a merge-conflict mistake

lucascolley (approved):
thanks Matt!

PR #22836, opened 2025-04-14
DOC, MAINT: Make stats distributions methods links to parent

melissawm:
#### Reference issue
--

For distributions inheriting from `scipy.stats._distribution_infrastructure.ContinuousDistribution`,
makes autosummary list the methods of the child class as links to the methods in the parent class.

Also exposes the `ContinuousDistribution` class so that these links are created.

#### Additional information
Also addresses two small docs build issues.

This is a work in progress, as it could be expanded to other distributions as well, and there may be more work to be done to improve the code.

Stub pages for the inherited classes' methods are still generated; I can investigate not generating them at all.

cc @mdhaber


doc/source/conf.py
```diff
@@ -581,5 +588,84 @@ def run(self):
...
+
+
+def custom_generate_autosummary_docs(*args, **kwargs):
+    """Filter out inherited methods before generating links to stub pages."""
+    names = args[0]
+    filtered_names = []
+
+    for name in names:
+        try:
+            if "scipy.stats.Normal" in name or "scipy.stats.Uniform" in name:
```
mdhaber, line 644: Can we get these names from `__all__` of `_new_distributions.py`?

mdhaber (commented):
Thanks for putting this together! Please do investigate not generating the other pages.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -30,7 +30,7 @@ def _isnull(x):
     return type(x) is object or x is None
 
 __all__ = ['make_distribution', 'Mixture', 'order_statistic',
-           'truncate', 'abs', 'exp', 'log']
+           'truncate', 'abs', 'exp', 'log', 'ContinuousDistribution']
```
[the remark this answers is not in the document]
melissawm: Yes it works - I'll remove :)

melissawm:
OK - unfortunately this took me way longer than I expected due to having to use internal Autosummary classes to do it, but it seems to work. I also added a filter since there are info messages we can ignore about 'document is referenced in multiple toctrees' for the methods we are altering.

@mdhaber hopefully this closes out the issue but if not, please let me know!

mdhaber:
This is great - it dramatically cuts down on the number of files generated. I've confirmed that separate distribution-specific method documentation files no longer exist, and the links go to the documentation of the parent class.

> hopefully this closes out the issue but if not, please let me know!

I know that there might not be time left to work on this, but I thought I'd check in about these things I noticed:
- There are still distribution-specific documentation files for parameters/attributes `tol`, `validation_policy`, and `cache_policy` - but actually, I don't see any links to them. Is it possible to:
  - eliminate these distributon-specific files
  - document distribution-specific parameters (e.g. `mu` and `sigma` for `Normal`)?
- In https://github.com/scipy/scipy/pull/22371#discussion_r1924702763, I mentioned the trouble with the rendering of:

image

I guess it appears that way because it is expecting the name of an attribute - can I skip that line and include the note in a regular font? Or can we just remove the `Attribute` section and provide the text in the Notes?

If there is not time for changes, it would be good to merge main (since we want this strategy to be used with the new `DiscreteDistribution` class, too) and merge the PR, but maybe I should open an issue about the remaining items above.

mdhaber:
Oops @melissawm, looks like something broke, but I can't tell whether it's the changes from gh-22312 (now there's a `UnivariateDistribution` class in-between `ContinuousDistribution`/`DiscreteDistribution` and `_ProbabilityDistribution`) or just what I did with the Attributes section. Do you know?

melissawm:
Hmmm I will take a look, maybe later this week. I'm a bit more confident now that I've done it once 😁 I'll let you know if I get stuck and we can think of a different plan.

melissawm:
Ok! In order to reference the parent methods here I had to expose UnivariateDistribution as well, hopefully this is ok. 

I see the Parameters and Attributes sections showing up nicely in my local build for Normal, Binomial and Uniform, let me know if you are still seeing anything wrong there.

mdhaber:
> I had to expose UnivariateDistribution as well, hopefully this is ok.

That's fine. 

But I suppose this raises the fact that since everything is inherited from `UnivariateDistribution`, we only need *those* versions of the documentation, whereas now we have separate copies for `ContinuousDistribution`, `DiscreteDistribution`, and `UnivariateDistribtion`. It's not a big deal, since that's only two more copies than necessary - just thought we should make an explicit decision to let it be. What do you think?

Can you also comment on:

image

---

> I see the Parameters and Attributes sections showing up nicely

They are now looking good.

Before:
image

After:
image

I still need to figure out how to document distribution-specific parameters/attributes (like `mu` and `sigma` of `Normal`), but that doesn't have to be in this PR.

melissawm:
Sorry I merged main hoping the preview would be fixed by not yet.

If you prefer removing ContinuousDistribution and DiscreteDistribution completely, we can do that since we are detecting these methods as coming from UnivariateDistribution correctly. I will address that in my next commit.

For distribution-specific attributes, I think maybe an extra "docstring" parameter to `_RealParameter` may work. In fact, if I navigate to `doc/build/html/reference/generated/scipy.stats.Normal.html`, the docstring of Normal mentions `mu` and `Sigma` and links to the correct page:

Captura de imagem_20250522_173000

So I think it's possible but we can think of something in a follow-up.

About the files we don't want to generate, I'll also address in the next commit 🙏🏻

mdhaber:
> If you prefer removing ContinuousDistribution and DiscreteDistribution completely,

Hmm.. Would that mean that appearances of `ContinuousDistribution` in the docs would go to the documentation of `UnivariateDistribution` (in which case let's remove them), or would they not work (in which case let's leave it alone)?

melissawm:
> Hmm.. Would that mean that appearances of ContinuousDistribution in the docs would go to the documentation of UnivariateDistribution (in which case let's remove them), or would they not work (in which case let's leave it alone)?

I think that would mean they would disappear completely, although I could also redirect those docs to UnivariateDistribution. Let me know what you prefer.

My last commit addresses the properties issue - now only distribution-specific properties have stub files, while the inherited ones don't.

Next, I'll take a look at the distribution-specific docstrings.

mdhaber:
> Let me know what you prefer.

Let's make them both link to `UnivariateDistribution` docs. All docs are shared between the two right now.

melissawm:
I think I got it! Let me know if this is acceptable. I chose to take a simpler route for remove ContinuousDistribution and DiscreteDistribution from being cited but we can explore a more general version if you prefer. Cheers!

scipy/stats/_distribution_infrastructure.py
```diff
@@ -721,6 +724,30 @@ class _RealParameter(_Parameter):
...
+        typical : 2-tuple of floats or strings, optional
+            Defines the endpoints of a typical range of values of the parameter.
+            Used for sampling. Defaults to `domain`.
+        docstring : str, optional
+            A string that describes the parameter. Defaults to None.
+
+        """
+        super().__init__(name, domain=domain, symbol=symbol, typical=typical,
+                         docstring=docstring)
+        self.__doc__ = self.docstring = docstring
```
mdhaber, line 695: Would it work if this were brought up to `_Parameter` in case we have other subclasses in the future?

scipy/stats/_new_distributions.py
```diff
@@ -29,10 +35,23 @@ class Normal(ContinuousDistribution):
...
+        docstring="The location parameter, which is also the mean, median, and mode. "\
+                  "Default: 0.0.",
+    )
+    
+    _sigma_param = _RealParameter(
+        'sigma',
+        symbol=r'\sigma',
+        domain=_sigma_domain,
+        typical=(0.5, 1.5),
+        docstring="The scalar parameter, which is also the standard deviation. "\
```
mdhaber, line 54: Oops, my mistake.
```suggestion
        docstring="The scale parameter, which is also the standard deviation. "\
```

scipy/stats/_distribution_infrastructure.py
```diff
@@ -806,13 +833,18 @@ class _Parameterization:
     """
     def __init__(self, *parameters):
         self.parameters = {param.name: param for param in parameters}
+        for param in parameters:
+            self.parameters[param.name].__doc__ = param.docstring
```
mdhaber, line 783: Doesn't this happen when the parameter is initialized?

scipy/stats/_new_distributions.py
```diff
@@ -29,10 +35,23 @@ class Normal(ContinuousDistribution):
     _sigma_domain = _RealInterval(endpoints=(0, inf))
     _x_support = _RealInterval(endpoints=(-inf, inf))
 
-    _mu_param = _RealParameter('mu',  symbol=r'\mu', domain=_mu_domain,
-                               typical=(-1, 1))
-    _sigma_param = _RealParameter('sigma', symbol=r'\sigma', domain=_sigma_domain,
-                                  typical=(0.5, 1.5))
+    _mu_param = _RealParameter(
```
mdhaber, line 40: It's great that this information appears at the links now! Since these are specified as parameters in the class initializer and available as attributes, can they appear at the front of the lists of Parameters and Attributes? It looks like I could implement it in `_combine_docs` if need be.

mdhaber (commented):
This is great! I'm tempted to just merge, but I wanted to check on just a few things.

scipy/stats/_new_distributions.py
melissawm: Hmm good question. I can try and take a look this weekend. If you get to it first, feel free to commit here!

melissawm:
Thanks @mdhaber - like I mentioned above I can try and finish this up on the weekend. If you have time before then and want to move this forward feel free to commit here 😄

mdhaber:
Hi @melissawm were you still willing to finish this? If you don't want to make changes anymore, it would still be helpful if you could reply to the last comments. Thanks!

lucascolley:
gentle ping @melissawm as we are close to the deadline for the next SciPy release

melissawm:
So sorry. Will get back to this in the next couple of days 🙏

tylerjereddy:
@melissawm planning to get this across the line in ~10 days or should I bump the milestone?

Issue #21466, opened 2024-08-27
ENH: linalg: N-D array batch support 


mdhaber:
### Is your feature request related to a problem? Please describe.

The array API standard and NumPy `linalg` functions support batched operation over stacks of matrices, vectors, etc. We occasionally get requests for SciPy to do the same (e.g. gh-16090, gh-17344, gh-20072), and it would help internally sometimes, too. (For instance, I've wanted batch support for many linear algebra operations in `multivariate_normal`.) Some of our functions do have this support, so for consistency, we should add support for the rest.

### Describe the solution you'd like.

`linalg` functions would support N-D input as they do in NumPy.

Conceptually, this means following normal NumPy broadcasting rules with the following adjustment: vectors and matrices, which are slices along the last one or two axes of an array, are treated as though they occupy only one element of the array. We ignore their "core dimensions" for the sake of broadcasting, and these core dimensions are always the last one or two of the array (rather than being specified by an `axis` argument).

For implementation, it is more precise to think of the shape of each N-D input as the sum (concatenation) of a batch shape and a core shape (e.g. the shape of a single vector or matrix). The net batch shape is the result of broadcasting the batch shapes of the inputs. The shape of each result array is the sum of the net batch shape and the core shape of each result. We perform the computation by looping over the batch dimensions.

Since `solve` and similar functions can accept a 1-D vector or 2-D stack of vectors already, there are some special cases, but the descriptions above work in most cases.

Initially, this can be handled throughout `linalg` by a decorator (e.g. gh-21462). More performant, lower-level implementations can be added at our leisure and tested against the decorator.

Here's a tracking list. Items marked with `+` are included in the array API standard linear algebra extension. It's organized in categories that made some sense to me, but it's not perfect. After writing gh-21462, I realized that categorization isn't really needed, anyway; the decorator can be used as-is for almost all cases. I might have left out some functions or neglected to note that they have batch support already. In any case, feel free to modify as you see fit.

Matrix in | scalar out
- [x] det+
- [ ] norm+
- [ ] issymmetric
- [ ] ishermitian
- [ ] expm_cond

Matrices in | vector out
- [ ] subspace_angles
- [ ] svdvals+
- [ ] eigvalsh+

Matrices in | matrices out
- [ ] inv+
- [ ] pinv+
- [ ] pinvh
- [ ] null_space
- [x] expm
- [ ] logm
- [ ] cosm
- [ ] sinm
- [ ] tanm
- [ ] coshm
[...]

PR #22838, opened 2025-04-14
POC/RFC: linalg: low-level nD support

ev-br:
#### Reference issue

towards https://github.com/scipy/scipy/issues/21466
supersedes and closes https://github.com/scipy/scipy/pull/21935

Add infrastructure for low-level batching in scipy.linalg. 

This is an alternative to gh-21935, which copy-pasted the gufunc infrastructure from numpy. This PR, instead, does manual looping over the batch dimensions, with the iterator from `sqrtm`, cc https://github.com/scipy/scipy/pull/22406#issuecomment-2796989303

Similar to gh-21935, here I convert `inv`. Which by itself is not a very interesting function; it's just simple enough to be useful as a guinea pig for the infrastructure. The PR looks large, but PRs for additional functions will be much smaller.

Remaining TBDs and action items:

- <s>`overwrite_a` not done yet; probably worth tackling in a follow-up;</s>
- "weird" dtypes, float16 and longdouble; also backwards compatible handling of integer array-likes
- what strategy we we use for when operation fails for some subarrays and succeeds for others: do we silently fill the failing parts with nans, or raise or fill with nans and emit a warning, cf https://github.com/scipy/scipy/issues/22476.  Currently this PR does quiet fill-with-nans.

#### Additional information

Quick-and-dirty performance measurements: basically, we are on par with numpy.linalg, which is to say 5-10x faster than the current scipy main for deep stacks of matrices of small core dimension.

```
In [1]: from scipy.linalg import inv

In [2]: from scipy.linalg._basic import inv0

In [3]: import numpy as np

In [4]: n = 10

In [5]: a = np.ones((n//2, n//5, n//10, n, n), dtype=float) + 8*np.eye(n)

In [6]: %timeit np.linalg.inv(a)
26.4 μs ± 409 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [7]: %timeit inv(a)
23.4 μs ± 187 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [8]: %timeit inv0(a)
211 μs ± 834 ns per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

### DIFF
--- a/scipy/linalg/_basic.py
+++ b/scipy/linalg/_basic.py
@@ -10,7 +10,7 @@ from itertools import product
 import numpy as np
 from numpy import atleast_1d, atleast_2d
 from scipy._lib._util import _apply_over_batch
-from .lapack import get_lapack_funcs, _compute_lwork
+from .lapack import get_lapack_funcs, _compute_lwork, _normalize_lapack_dtype
 from ._misc import LinAlgError, _datacopied, LinAlgWarning
 from ._decomp import _asarray_validated
 from . import _decomp, _decomp_svd
@@ -22,17 +22,6 @@ from . import _batched_linalg
 __all__ = ['solve', 'solve_triangular', 'solveh_banded', 'solve_banded',
            'solve_toeplitz', 'solve_circulant', 'inv', 'det', 'lstsq',
            'pinv', 'pinvh', 'matrix_balance', 'matmul_toeplitz']
-
-
-# The numpy facilities for type-casting checks are too slow for small sized
-# arrays and eat away the time budget for the checkups. Here we set a
-# precomputed dict container of the numpy.can_cast() table.
-
-# It can be used to determine quickly what a dtype can be cast to LAPACK
-# compatible types, i.e., 'float32, float64, complex64, complex128'.
-# Then it can be checked via "casting_dict[arr.dtype.char]"
-lapack_cast_dict = {x: ''.join([y for y in 'fdFD' if np.can_cast(x, y)])
-                    for x in np.typecodes['All']}
 
 
 # Linear equations
@@ -1245,17 +1234,12 @@ def inv(a, overwrite_a=False, check_finite=True):
         dt = inv(np.eye(2, dtype=a1.dtype)).dtype
         return np.empty_like(a1, dtype=dt)
 
-    # dtypes
-    if issubclass(a1.dtype.type, np.integer):
-        # XXX float16, longdouble, check backwards compat
+    # Also check if dtype is LAPACK compatible
+    a1, overwrite_a = _normalize_lapack_dtype(a1, overwrite_a)
+
+    if not (a1.flags['ALIGNED'] or a1.dtype.byteorder == '='):
         overwrite_a = False
-        a1 = a1.astype(np.float64)            
-    elif not issubclass(a1.dtype.type, np.number):
-        raise ValueError(f"Expected a numeric array, got f{a1.dtype=}")
-    else:
-        if not (a1.flags['ALIGNED'] or a1.dtype.byteorder == '='):
-            overwrite_a = False
-            a1 = a1.copy()
+        a1 = a1.copy()
 
     if a1.ndim == 2:
         a1 = a1[None, ...]
@@ -1344,14 +1328,7 @@ def det(a, overwrite_a=False, check_finite=True):
                          f' but received shape {a1.shape}.')
 
     # Also check if dtype is LAPACK compatible
-    if a1.dtype.char not in 'fdFD':
-        dtype_char = lapack_cast_dict[a1.dtype.char]
-        if not dtype_char:  # No casting possible
-            raise TypeError(f'The dtype "{a1.dtype.name}" cannot be cast '
-                            'to float(32, 64) or complex(64, 128).')
-
-        a1 = a1.astype(dtype_char[0])  # makes a copy, free to scratch
-        overwrite_a = True
+    a1, overwrite_a = _normalize_lapack_dtype(a1, overwrite_a)
 
     # Empty array has determinant 1 because math.
     if min(*a1.shape) == 0:
--- a/scipy/linalg/_basic.py
+++ b/scipy/linalg/_basic.py
@@ -1237,23 +1237,13 @@ def inv(a, overwrite_a=False, check_finite=True):
     # Also check if dtype is LAPACK compatible
     a1, overwrite_a = _normalize_lapack_dtype(a1, overwrite_a)
 
-    # XXX move flags/strides check to C
     if not (a1.flags['ALIGNED'] or a1.dtype.byteorder == '='):
         overwrite_a = True
         a1 = a1.copy()
 
-    # batched arrays always copy
-    overwrite_a = overwrite_a and a1.flags["F_CONTIGUOUS"] and a1.ndim == 2
-
-    extradim = False
-    if a1.ndim == 2:
-        a1 = a1[None, ...]
-        extradim = True
-
-    # by now, we have at least a 3D well behaved array of a BLAS-compatible dtype
+    # a1 is well behaved, invert it.
     inv_a = _batched_linalg.inv(a1, overwrite_a)
-
-    return inv_a[0, ...] if extradim else inv_a
+    return inv_a
 
 
 # Determinant
--- a/scipy/linalg/_basic.py
+++ b/scipy/linalg/_basic.py
@@ -1117,6 +1117,82 @@ def solve_circulant(c, b, singular='raise', tol=None,
 
 
 # matrix inversion
+@_apply_over_batch(('a', 2))
+def inv0(a, overwrite_a=False, check_finite=True):
+    """
+    Compute the inverse of a matrix.
+
+    Parameters
+    ----------
+    a : array_like
+        Square matrix to be inverted.
+    overwrite_a : bool, optional
+        Discard data in `a` (may improve performance). Default is False.
+    check_finite : bool, optional
+        Whether to check that the input matrix contains only finite numbers.
+        Disabling may give a performance gain, but may result in problems
+        (crashes, non-termination) if the inputs do contain infinities or NaNs.
+
+    Returns
+    -------
+    ainv : ndarray
+        Inverse of the matrix `a`.
+
+    Raises
+    ------
+    LinAlgError
+        If `a` is singular.
+    ValueError
+        If `a` is not square, or not 2D.
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> from scipy import linalg
+    >>> a = np.array([[1., 2.], [3., 4.]])
+    >>> linalg.inv(a)
+    array([[-2. ,  1. ],
+           [ 1.5, -0.5]])
+    >>> np.dot(a, linalg.inv(a))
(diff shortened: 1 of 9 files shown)


scipy/linalg/_basic.py
```diff
@@ -1117,7 +1118,7 @@ def solve_circulant(c, b, singular='raise', tol=None,
 
 # matrix inversion
 @_apply_over_batch(('a', 2))
```
mdhaber, line 1120: IIRC this wouldn't be needed anymore?
```suggestion
```

ev-br: Yes, totally. ATM though, this is `inv0`, temporarily kept to simplify perf comparisons.

ilayn:
Thank you for this. I get stomach aches by looking at Arpack these days. I'll get back to this once I got over my nausea. I can't say I can follow the OOP parts though. Seems like we have 10x boilerplate instead of 4x boilerplate now 😝 Let me finish that arpack thing up then we can go back to our skirmish about C vs. ++

ev-br:
Thanks Ilhan. Would be great to detail your specific concerns about this, once you're safely back from your travels in the wondrous ARPACK-land of course :-).

Meanwhile, the last commit
- adds `overwrite_a=True` in a backwards compatible way;
- refactors the code a bit to make it more C-like and hopefully easier to follow.

I don't think I'm using any OOP, and the ++ features are at minimum. In fact, it's pretty much all C (manual memory management, `goto done` etc etc) with a few basic templates to cut down in 4x duplication of algorithmic parts.

And I believe the boilerplate here is enough to very easily implement all LU-related functionality (det, lu, lu_{factor,solve}). Can do it either here or in  a follow-up.

One thing that bugs me still is error handling, gh-22476. `numpy.linalg` does something that their gufuncs talk to `np.errstate`. This is hidden from the user in numpy (https://github.com/numpy/numpy/blob/v2.2.0/numpy/linalg/_linalg.py#L607), so maybe the right thing for scipy.linalg is to use a sensible default (raise in 2D, fill with nans for ndim > 2) and make it talk to `np.errstate`, so that users can use the context manager. I'll take a look at how hard it is to make it a thing.

scipy/linalg/_basic.py
```diff
@@ -1191,6 +1181,81 @@ def inv(a, overwrite_a=False, check_finite=True):
...
+    # XXX move flags/strides check to C
+    if not (a1.flags['ALIGNED'] or a1.dtype.byteorder == '='):
+        overwrite_a = True
+        a1 = a1.copy()
+
+    # batched arrays always copy
+    overwrite_a = overwrite_a and a1.flags["F_CONTIGUOUS"] and a1.ndim == 2
+
+    extradim = False
+    if a1.ndim == 2:
```
ilayn, line 1249: See below for this

scipy/linalg/meson.build
```diff
@@ -113,6 +113,22 @@ py3.extension_module('_decomp_interpolative',
   subdir: 'scipy/linalg'
 )
 
+
+py3.extension_module('_batched_linalg',
```
ilayn, line 117: If you don't do this as a static (and shared for BLAS/LAPACK dependency) library, other C code can't use it. That's why it is just a header file in common utils right now for sqrtm and eventually for others.

scipy/linalg/src/_batched_linalg.h
```diff
@@ -0,0 +1,305 @@
...
+#include "Python.h"
+#include <limits>
+#include <cassert>
+#include "numpy/arrayobject.h"
+#include "npy_cblas.h"
+#include "_lapack_trampolines.h"
+#include "_npymath.h"
+
+
+using namespace _numpymath;
```
ilayn, line 12: why is this namespace `_numpymath` required? Can't we not template using the regular dependencies to NumPy without our own extra boilerplates?

scipy/linalg/src/_batched_linalg.h
```diff
@@ -0,0 +1,305 @@
...
+
+
+// parroted from sqrtm
+// XXX can probably be replaced by ?copy from BLAS
```
ilayn, line 16: No unfortunately, this is basically to transpose a matrix in smaller chunks that fits into L1 cache to reduce the cache misses. 

At least in theory. I think we should unroll these to 8 for reals and 4 for complexes.I need a proper godbolt session to get a feeling for what compilers are choosing to do. I'll get to it eventually. 

 `dcopy` is BLAS' `memcpy`. They do different things.

scipy/linalg/src/_batched_linalg.h
```diff
@@ -0,0 +1,305 @@
...
+    }
+}
+
+
+/*
+ * Visit (m, n)-shaped slices of a ndim>=3 array in C order.
+ * Copy each slice into a buffer, make the buffer F-ordered
+ * for LAPACK.
+ */
+struct iter_data_t
```
ilayn, line 56: I don't get your comment about not using OOP. This is pretty much a class with a single method. Instead of loop optimizations now you are converting it into a Python generator and `yield`ing a slice here if I can read C++ correctly.

scipy/linalg/src/_batched_linalg.h
```diff
@@ -0,0 +1,305 @@
...
+
+
+/***********************************
+ *   Inversion via GETRF / GETRI
+ ***********************************/
+
+/*
+ * Invert a 2D slice:
+ *   - Input slice is in `getrf_data.a`.
+ *   - The result is in `getri_data.a`.
```
ilayn, line 120: This is a class instance here. I really don't see the complication helping us here a class with a single method is pretty much a function. Still missing the call to `gecon` though.

scipy/linalg/src/_batched_linalg.h
```diff
@@ -0,0 +1,305 @@
...
+ * The main batched-inv routine, to be called from the python API layer.
+ *
+ * This is a computational routine, and it assumes that the caller has done all input
+ * validation before calling it.
+ * In particular:
+ *
+ * For the input array `arr`:
+ *
+ * - The caller must check that the dtype is LAPACK-compatible before calling this routine.
+ * - The array is at least 3D (for a single 2D matrix, the caller must
```
ilayn, line 159: This is not necessary, you can detect this from the ndim property of the array. Then you loop only once in the outer slice loop if ndim = 2. That's the setting to `1` in sqrtm outer dimension multiplication.

scipy/linalg/src/_lapack_trampolines.h
```diff
@@ -0,0 +1,221 @@
...
+);
+
+
+} // extern "C"
+
+
+/*
+ * Hold the GESV related variables, handle allocation/deallocation.
+ */
+template<typename T>
```
ilayn, line 87: I understand the temptation but this is clearly a C++ class. We need to justify the need for a class within us first. 

If it is always a single instance then this is just unnecessary code making things complicated and holding bunch of variables together. I can't reuse anything inside this class it is just wasted memory. Say, I already have a larger integer array, I want to use it with a `gesv` call. Now you are taking away that possibility with this or making it 10x harder to achieve because then I have to overwrite this with *dependency injection* or whatever the terminology is. 

If we had 100 instances of this entity, like a computer game with 100 characters, yes by all means let's class'ify it but for single instances Class usage is an abstraction that carries the burden of proof. Independent from C++, we also have such code in Python, equally difficult to keep track of. The most recent I'm working on is `_arpack.py` with three big classes and single instances. It is really not a great experience I can tell you to troubleshoot. 

We just need three consecutive lapack calls for getrf/gecon/getri. Look at all the preparation code that goes into it. And I can't import it to a C code if we have to.

scipy/linalg/src/_npymath.h
```diff
@@ -0,0 +1,89 @@
...
+    static constexpr npy_cdouble nan = {std::numeric_limits<double>::quiet_NaN(),
+                                        std::numeric_limits<double>::quiet_NaN()};
+};
+
+
+
+/* 
+ * Grab a real part of a possibly complex array.
+ * This is for the work queries.
+ * There must be a better way, I'm sure.
```
ilayn, line 47: There isn't, this is the templating bloat I was mentioning before. Even to get the real part this is the boilerplate code we have to inject.

scipy/linalg/src/_npymath.h
```diff
@@ -0,0 +1,89 @@
...
+
+
+/* identity square matrix generation */
+template<typename typ>
+static inline void
+identity_matrix(typ *matrix, ptrdiff_t n)
+{
+    ptrdiff_t i;
+    /* in IEEE floating point, zeroes are represented as bitwise 0 */
+    memset((void *)matrix, 0, n*n*sizeof(typ));
```
ilayn, line 68: If there is no padding injected or some other thing happened. 

This is dancing around the UB and compiler dependent behavior. I already broke scikit-learn with this in the first `expm` write that took us a few days to figure out. I don't know if now it is safer but I wouldn't touch it. Safer to do the loop and set the number to 0.0 like you did with NaN below, compiler would optimize it anyways (speaking like a true fortran person).

ilayn (commented):
I think I have to write a similar implementation to explain myself. 

I feel that this is not fair to criticize over your work without putting any effort but just typing terse sentences that I know we will have a back and forth. So please take them lightly until I write something *very arcane* so you would have the fair chance to dunk on the shortcomings of what I do 😃 

Regardless of what we do in the end, this is really more performant and cleaner in my opinion than doing bunch of `get_lapack_funcs` and losing lots of performance hidden away. And even in your version we only have 200ish lines and the rest is going to be recycled for other funcs. So overall, I think this is worth the effort.

ev-br (commented):
Agreed, let's try to converge on a version with minimum viable amount of boilerplate :-), so that we can proceed chipping away on `get_lapack_funcs`.

There are a couple of questions below, to try clarifying what kind of reworks would work for you. We could hop on a higher-bandwidth conversation, too, if that's helpful.

For me, the main thing is that currently the main loop, `inv_loop` is under 100 LOCs, with memory allocations and comments.

To keep it in that ballpark, we do need some sort of templating, be it `template<typename T>` or Tempita or even distutils templates [1] if push comes to shove.


[1] not my first choice, mind you :-).

scipy/linalg/src/_npymath.h
ev-br: Minor comment that it's not templating, really. It's just overloading, so would be exactly the same if it all were pure C.

ev-br:
Okay, following up on https://github.com/scipy/scipy/pull/22838#discussion_r2050587438, https://github.com/scipy/scipy/pull/22838/commits/96c1831fe066efcb9f643b7a8b7676bdac6035ab ditches the `getrf_data` structs and inlines the LAPACK calls.

ilayn:
Just sent you a link to the board currently showing this so we can organize things a bit better

shapes at 25-04-19 22 26 19

ilayn:
Talking to np.errstate is an excellent idea but if it is possible by the way. I'll dig a bit about it.

j-bowhay:
It would be great to coordinate with #28782 so there aren't multiple ways of handling errors in stacks of matrices

ev-br:
superseded by https://github.com/scipy/scipy/pull/22924, closing

PR #22841, opened 2025-04-16
ENH: stats.qmc.geometric_discrepancy: improve memory complexity for `method='mindist'`

### LINKED ISSUE #22837: ENH: stats.qmc.geometric_discrepancy: `method='mindist'` can improve memory complexity

`geometric_discrepancy(method='mindist')` calculates a full pairwise distance matrix just to return the minimum value of the matrix. This requires quadratic memory, which has very poor scalability, as the example below shows.  With 32768 points, `geometric_discrepancy(method='mindist')` requires more than 4GB of memory!

Chunking the computation per row as in the code below has the same quadratic computation time but only uses linear memory. 

```python
import numpy as np
import pandas as pd
import timeit
from scipy.stats import qmc
from scipy.spatial import distance
import seaborn as sns
sns.set_theme(style="darkgrid")

def pmindist(x):
    n = x.shape[0]
    d = np.inf
    for i in range(1, n):
        d = min(d, distance.cdist(x[i:, :], x[(i-1):i, :]).min())
    return d

res = []
for k in range(10,29):
    n = int(np.sqrt(2**k))
    x = np.random.default_rng().random((n, 2))
    start = timeit.default_timer()
    v1 = pmindist(x)
    t1 = timeit.default_timer()
    v2 = qmc.geometric_discrepancy(x)
    t2 = timeit.default_timer()
    assert v1 == v2
    res += [ dict(n=n, time=t1 - start, method="pmindist")]
    res += [ dict(n=n, time=t2 - t1, method="qmc")]
    
df = pd.DataFrame(res)
plt.figure() 
ax = sns.lineplot(data=df, x="n", y="time", hue="method", marker='o')
ax.set(yscale='log', xscale='log')
#plt.xticks(df['n'].unique(), [ fr'$2^{{{int(p)}}}$' for p in df['n'].unique()])
plt.show()
```

A C/C++ implementation would be even faster.

Scikit-learn has [`pairwise_distances_chunked`](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise_distances_chunked.html), which may be faster but more complex to implement.

Image

MLopez-Ibanez:
…plexity (#22837)

#### Reference issue

Closes gh-22837

Improves memory complexity of `qmc.geometric_discrepancy`.


scipy/stats/_qmc.py
```diff
@@ -336,10 +336,57 @@ def discrepancy(
...
+    elif metric == 'cityblock':
+        p = 1
+        distance_fun = distance.cityblock
+    elif metric == 'chebyshev':
+        p = np.inf
+        distance_fun = distance.chebyshev
+    else:
+        # Slow path for metrics unsupported by KDTree.
+        distances = distance.pdist(sample, metric=metric)  # type: ignore[call-overload]
+        return distances.min()
```
dschmitz89, line 375: Could this be prettified using a match/case block?

dschmitz89 (commented):
This is a bit outside my wheelhouse but still doing a cursory review. We definitely need tests for the new `workers` keyword.

scipy/stats/_qmc.py
```diff
@@ -336,10 +336,57 @@ def discrepancy(
...
+
+    distance_upper_bound = distance_fun(sample[0,...], sample[1,...])
```
tylerjereddy, line 378: In local testing I see a test failure because `distance_upper_bound` can of course be `0`, and KDTree then returns non-finite distances instead of `0` distances. That's a bit weird, but anyway, I've suggested a potential shim below the fold that makes the test happy locally (properly warns when there are duplicated points).

<details>

```diff
--- a/scipy/stats/_qmc.py
+++ b/scipy/stats/_qmc.py
@@ -375,6 +375,7 @@ def pmindist(
         return distances.min()
 
     distance_upper_bound = distance_fun(sample[0,...], sample[1,...])
+    distance_upper_bound = max(distance_upper_bound, 1e-15)
     tree = KDTree(sample)
     d, _ = tree.query(sample,
                       k=[2], p = p,
@@ -493,7 +494,10 @@ def geometric_discrepancy(
         raise ValueError("Sample must contain at least two points")
     
     if method == "mindist":
-        return pmindist(sample, metric=metric, workers=workers)
+        min_d = pmindist(sample, metric=metric, workers=workers)
+        if np.allclose(min_d, 0.0):
+            warnings.warn("Sample contains duplicate points.", stacklevel=2)
+        return min_d
```

</details>

scipy/stats/_qmc.py
```diff
@@ -336,10 +336,57 @@ def discrepancy(
...
+    tree = KDTree(sample)
+    d, _ = tree.query(sample,
+                      k=[2], p = p,
```
tylerjereddy, line 381: minor/nit: remove spacing between `p = p`, which anyway would be consistent with the rest of your formatting and what we usually do

scipy/stats/_qmc.py
```diff
@@ -353,6 +400,9 @@ def geometric_discrepancy(
         The distance metric to use. See the documentation
         for `scipy.spatial.distance.pdist` for the available metrics and
         the default.
+    workers : int, optional
+        Number of workers to use for parallel processing. If -1 is given all
+        CPU threads are used. Default is 1.
```
tylerjereddy, line 406: Minor: could be worth mentioning that this only applies with `mindist` `method` option (I believe we have seen some credible proposals to add concurrency/acceleration to distance metrics, but that's another matter).

IIRC there was some discussion between Matt and Albert (?) about avoiding kwargs that only apply to a subset of other kwargs, but I don't really see a better option.

tylerjereddy (commented):
I saw two test failures in `test_qmc.py` when checking this branch locally--added a suggestion related to one of them.

If we don't already have `asv` benchmarks in `benchmarks/stats.py` that would capture this, that may be useful to add, but we don't usually block PRs on it.

Let me see if I can understand the remaining test fail I see locally. The rest of the changes look broadly sensible re: computational geometry. Would be good if Pamphile could scan it over.

scipy/stats/_qmc.py
```diff
@@ -336,10 +336,58 @@ def discrepancy(
...
+    match metric:
+        case 'euclidean':
+            p = 2
+            distance_fun = distance.euclidean
+        case 'cityblock':
+            p = 1
+            distance_fun = distance.cityblock
+        case 'chebyshev':
+            p = np.inf
+            distance_fun = distance.chebyshev
```
tylerjereddy, line 372: Looks like the remaining test failure I saw locally was because KDTree was giving a non-finite result with `chebyshev`.

If you don't find a workaround, one crude fix that worked locally was just to send `cheybshev` to `pdist` as well for now:

<details>

```diff
--- a/scipy/stats/_qmc.py
+++ b/scipy/stats/_qmc.py
@@ -366,15 +366,13 @@ def pmindist(
     elif metric == 'cityblock':
         p = 1
         distance_fun = distance.cityblock
-    elif metric == 'chebyshev':
-        p = np.inf
-        distance_fun = distance.chebyshev
```

</details>

scipy/stats/_qmc.py
MLopez-Ibanez: Why not simply return 0 before calling KDTree? I don't see the point of warning. A value of zero is a valid value of the the minimum distance that happens if (and only if) there are duplicates (or if the points are so close that they are for all purposes duplicates).

scipy/stats/_qmc.py
MLopez-Ibanez: OK. I was hoping that the automatic pre-commit linter (like ruff) would fix those automatically.

scipy/stats/_qmc.py
tylerjereddy: This is what the regression tests currently enforce, so it would be a behavior change not to warn. I don't think it is really worth trying to push for a behavior change and a performance enhancement at the same time, but I'll leave that for Pamphile to decide.

scipy/stats/_qmc.py
MLopez-Ibanez: The documentation of KDTree gives the impression that `np.inf` should work but happy to use your work-around.

scipy/stats/tests/test_qmc.py
```diff
@@ -224,11 +224,23 @@ def test_geometric_discrepancy(self):
         assert_allclose(
             qmc.geometric_discrepancy(sample, method='mst'), 0.19704396643366182
         )
-
-    @pytest.mark.xfail(
-            reason="minimum_spanning_tree ignores zero distances (#18892)",
-            strict=True,
-    )
```
tylerjereddy, line 231: Why was this mark removed?

scipy/stats/tests/test_qmc.py
```diff
@@ -224,11 +224,23 @@ def test_geometric_discrepancy(self):
...
+        sample = np.array([[0, 0], [0, 1], [0.5, 1]])
+        assert_allclose(qmc.geometric_discrepancy(sample, workers=8), 0.5)
+        
+        sample = np.array([[0, 0], [0.25, 0.25], [1, 1]])
+        assert_allclose(qmc.geometric_discrepancy(sample, workers=8), np.sqrt(2) / 4)
+        assert_allclose(qmc.geometric_discrepancy(sample, metric="chebyshev", workers=8), 0.25)
+        
+        rng = np.random.default_rng(191468432622931918890291693003068437394)
+        sample = qmc.LatinHypercube(d=3, rng=rng).random(50)
+        assert_allclose(qmc.geometric_discrepancy(sample, workers=8), 0.05106012076093356)
```
tylerjereddy, line 242: here and above, `8` workers may be a bit aggressive for conventional CI testing

scipy/stats/_qmc.py
MLopez-Ibanez: OK, you convinced me.

scipy/stats/tests/test_qmc.py
MLopez-Ibanez: Fixed now.

MLopez-Ibanez:
> This is a bit outside my wheelhouse but still doing a cursory review. We definitely need tests for the new `workers` keyword.

tests added now.

scipy/stats/_qmc.py
tupui: The alternative would be to rely on Array API support here. Not sure about the plan on the dist part now.

scipy/stats/_qmc.py
```diff
@@ -336,10 +336,57 @@ def discrepancy(
                          f" {set(methods)!r}")
 
 
+def pmindist(
```
tupui, line 339: Almost look like this should go into spatial. What do you think Tyler?

tupui (commented):
Thanks for making a PR 😃 I have some notes, nothing blocking from my side otherwise.

scipy/stats/_qmc.py
MLopez-Ibanez: OK, I have documented it now.

scipy/stats/_qmc.py
tylerjereddy: Sure, that's fair, and maybe we could prefix the name with `_` so it is "informally private?" In the more distant future we could write detailed tests on the `spatial` side and then try to support more metrics for efficient "closest distance in a sample" if we need to. I don't know if it would ever become public on `spatial` side, but seems like a useful internal utility at least.

One other comment I have is that it is helpful to leave the review comment "resolution" to the maintainers so we can double check that items are resolved on our end rather than collapsing them right away, but minor point.

scipy/stats/_qmc.py
tupui: The warning is arguable, this was written for QMC use in mind first and in that context having duplicate points is seen as a bad thing (because if you use that point to evaluate a function it has a cost for no extra information you get on the output, provided you have sort of deterministic model.) So yes in the QMC function I would keep the warning.

scipy/stats/_qmc.py
MLopez-Ibanez: OK, moved to `spatial` with a `_` prefix.

MLopez-Ibanez:
A question about process. Should I squash all commits in my fork once you approve all the changes or should I squash every commit so there is only one commit in the pull request?

tupui:
> A question about process. Should I squash all commits in my fork once you approve all the changes or should I squash every commit so there is only one commit in the pull request?

Thanks for the question. Don't worry about that. The review can take some time (i.e. more commits) and in the end it's up to the maintainer merging the changes to either squash merge or merge. If it's a merge we either ask that you rewrite a bit the history to make it nicer or sometimes the maintainer does it. TLDR, there is no enforced convention. What matters is the quality of the code getting in 😃

scipy/stats/_qmc.py
MLopez-Ibanez: The warning is kept in `geometric_discrepancy`.

MLopez-Ibanez:
> Thanks for the question. Don't worry about that. The review can take some time (i.e. more commits) and in the end it's up to the maintainer merging the changes to either squash merge or merge. If it's a merge we either ask that you rewrite a bit the history to make it nicer or sometimes the maintainer does it. TLDR, there is no enforced convention. What matters is the quality of the code getting in 😃

OK. Happy to squash all the commits to keep the history clean. And happy if someone else does it.

scipy/stats/_qmc.py
MLopez-Ibanez: @tylerjereddy Could you check again? I think I found the reason and it was a more general bug. If the distance between the two points (`distance_upper_bound`) is actually the minimum distance, then `query` will return infinity for all points. In that case, we should return `distance_upper_bound`.

tylerjereddy:
I activated the full CI run here to see if anything pops up.

tylerjereddy:
I drafted some benchmarks for the proposed improvements here over at gh-22907. One question--there are some slowdowns at higher dimensions, probably because of the curse of dimensionality for `KDTree`--are high dims a common use case in `qmc`? If so, we may want to be a little cautious, although figuring out an appropriate crossover point (how many dims?) for KDTree vs. "brute force" in terms of dims for each metric may be annoying.

MLopez-Ibanez:
> I drafted some benchmarks for the proposed improvements here over at [gh-22907](https://github.com/scipy/scipy/pull/22907). One question--there are some slowdowns at higher dimensions, probably because of the curse of dimensionality for `KDTree`--are high dims a common use case in `qmc`? If so, we may want to be a little cautious, although figuring out an appropriate crossover point (how many dims?) for KDTree vs. "brute force" in terms of dims for each metric may be annoying.

It depends what you mean by "common". Usually low-discrepancy sequences become worse with higher number of dimensions. There is also less research about many-dimensional sequences, AFAIK. 

I wonder how my originally proposed "brute-force" approach behaves for large dimensions:

```python
def pmindist(x):
    n = x.shape[0]
    d = np.inf
    for i in range(1, n):
        d = min(d, distance.cdist(x[i:, :], x[(i-1):i, :]).min())
    return d
``` 

If you look at the plots in issue #2287, the KD-tree approach is also slightly slower for less than 300 points. Could you benchmark that approach for higher dimensions?

tupui:
> It depends what you mean by "common". Usually low-discrepancy sequences become worse with higher number of dimensions. There is also less research about many-dimensional sequences, AFAIK.

QMC is essential for high dimensional problem as it really shows its strength compared with crude MC. Just plot the canonical 2D subprojections on a 10D hypercube, you will see large empty regions if you use MC or random LHS. Halton and Sobol' here don't suffer much.

The nuance here is where to find what is high dimensionality. There is not much guidelines there besides saying that above 20 I would say is considered high already by some people. It's very much field dependent as well.

tylerjereddy:
Right, there is probably a larger constant factor for the KDTree approach so small datasets suffer a little for the tradeoff of better asymptotic scaling.

> If you look at the plots in issue https://github.com/scipy/scipy/issues/2287, the KD-tree approach is also slightly slower for less than 300 points. Could you benchmark that approach for higher dimensions?

I can help with benchmarking, sure. I'm just worried about the explosion of performance-related codepaths that might come up like:

- if `<=N` dimensions, use KDTree for those metrics that allow it
- if `>N` dimensions or for some other metrics, use the brute force/batching approach

And there is also the question of whether adding concurrency to KDTree at higher dims allows us to partially rescue the performance, such that we wouldn't want a hard switch on `N` dimensions. Does this concern about getting messy with many specialized code paths make sense? I'm not saying we shouldn't do something, just a little worried about where it might lead us.

MLopez-Ibanez:
> > It depends what you mean by "common". Usually low-discrepancy sequences become worse with higher number of dimensions. There is also less research about many-dimensional sequences, AFAIK.
> 
> QMC is essential for high dimensional problem as it really shows its strength compared with crude MC. Just plot the canonical 2D subprojections on a 10D hypercube, you will see large empty regions if you use MC or random LHS. Halton and Sobol' (scrambled) here don't suffer much.

Apologies, I was basing my opinion on conclusions such as: "The answer to the question of whether QMC is superior to MC and how well QMC could work for high-dimensional integrals depends strongly on the structure of the function. If a function is dominated by its first-order part, or if the earlier variables are much more important than others, then good results can be expected from using QMC. However, if higher-order ANOVA terms play an important role, then the superiority of QMC is not guaranteed for relatively small ."  https://doi.org/10.1016/j.cam.2007.01.005

But I'm certainly not an expert on QMC.

> The nuance here is where to find what is high dimensionality. There is not much guidelines there besides saying that above 20 I would say is considered high already by some people. It's very much field dependent as well.

Complete agreement here.

MLopez-Ibanez:
> Ah, Pamphile commented at same time as me. Ok yeah, so can't ignore higher dim

Adding a threshold for higher dimensions to not use KDTree will also disable parallel processing. I am unsure the slowdown of KDTree is really worth losing the parallel processing capability.

MLopez-Ibanez:
Hi @tylerjereddy , is there anything that remains to do here?

dschmitz89:
@MLopez-Ibanez : could you fix the lint failures?

MLopez-Ibanez:
> @MLopez-Ibanez : could you fix the lint failures?

Some of those errors are unrelated to my patch but OK, I fixed them. 

it would be great to let the pre-commit.ci action apply these trivial fixes and also fix trailing whitespaces automatically.

tylerjereddy:
> is there anything that remains to do here?

My main concern is the performance tradeoffs noted at gh-22907. It just makes it a bit harder to declare a clear victory over the previous approach. If the previous space complexity was intractable maybe that can allow us to excuse some high dimensional slowdowns now that we leverage KDTree.

MLopez-Ibanez:
> My main concern is the performance tradeoffs noted at [gh-22907](https://github.com/scipy/scipy/pull/22907). It just makes it a bit harder to declare a clear victory over the previous approach. If the previous space complexity was intractable maybe that can allow us to excuse some high dimensional slowdowns now that we leverage KDTree.

It may be useful to recall that there is no such performance trade-off with the solution that I originally proposed and it only arises with the KDTree that I was asked to implement. 

I have contributed to many open-source projects and this pattern is quite common. A first-time/sporadic contributor proposes a solution that, although imperfect, it is better than the status-quo. The contributor is asked to perfect the solution through several rounds of review just to find out that the "more perfect" solution is non-viable. At this point, most people would have already given up. Since I know very well that this pattern is actually pretty common and often unconscious, I don't want to give up just yet.

I have opened #22993 that gets rid of the KDTree. This should be an overall win in all cases with respect to the current code. There are many possible improvements but I would suggest that all of them should be implemented as a follow up.

tylerjereddy:
> This should be an overall win in all cases with respect to the current code.

I can't imagine it would be reasonable to expect the original approach of batching to "win" in all cases--batching `cdist` in a Python for loop is going to have some kind of performance cost that you don't pay if you don't batch it, otherwise it wouldn't have been ported to C/C++ in the first case and we could have just used Python loops.

I suspect the reviewer (probably me) will be left with some tricky decisions to consider. Thanks for your patience--I'll try to run the `asv` benchmarks I drafted for this scenario to get an initial sense of the situation for your alternate proposal.

MLopez-Ibanez:
I will close this as I see zero-chance that this approach will be accepted. A pity because the parallel processing capability seemed very useful. 

I will keep gh-22993 open as I think the new version should be an overall (but minor) improvement, so hopefully it has a chance to get in.

Having better benchmarks (gh-22907) for high dimensions and high number of points that test the turning point of the memory explosion would be welcome. In my computer, 32768 2D points is already noticeably slower.

PR #21215, opened 2024-07-19
MAINT: `_lib`: update `_docscrape` from upstream

lucascolley:
[skip cirrus]

#### Reference issue
n/a

#### What does this implement/fix?

Brings `_lib._docscrape` up to date with upstream.

#### Additional information
This is used in `stats._axis_nan_policy`.

Accidentally pushed a branch to the main repo again, will fix my codespaces setup.

PR #22843, opened 2025-04-16
DOC: `special`: add `xp_capabilities` to logsumexp

crusaderky:
Automatically generates documentation

image

### DIFF
--- a/scipy/special/_logsumexp.py
+++ b/scipy/special/_logsumexp.py
@@ -1,6 +1,7 @@
 import numpy as np
 from scipy._lib._array_api import (
     array_namespace,
+    xp_capabilities,
     xp_device,
     xp_size,
     xp_promote,
@@ -11,6 +12,7 @@ from scipy._lib import array_api_extra as xpx
 __all__ = ["logsumexp", "softmax", "log_softmax"]
 
 
+@xp_capabilities()
 def logsumexp(a, axis=None, b=None, keepdims=False, return_sign=False):
     """Compute the log of the sum of exponentials of input elements.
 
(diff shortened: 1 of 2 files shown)


scipy/special/_logsumexp.py
```diff
@@ -58,7 +60,8 @@ def logsumexp(a, axis=None, b=None, keepdims=False, return_sign=False):
 
     See Also
     --------
-    numpy.logaddexp, numpy.logaddexp2
+    :data:`numpy.logaddexp`
+    :data:`numpy.logaddexp2`
```
crusaderky, line 64: Work around to 
- https://github.com/numpy/numpy/issues/28734
- https://github.com/numpy/numpydoc/issues/618

Note that docscrape has been bumped up recently-ish: https://github.com/scipy/scipy/pull/21215

scipy/special/tests/test_logsumexp.py
```diff
@@ -38,6 +39,7 @@ def test_wrap_radians(xp):
 @pytest.mark.filterwarnings("ignore:invalid value encountered:RuntimeWarning")
 @pytest.mark.filterwarnings("ignore:divide by zero encountered:RuntimeWarning")
 @pytest.mark.filterwarnings("ignore:overflow encountered:RuntimeWarning")
+@make_skip_xp_backends(logsumexp)
```
crusaderky, line 42: These don't do anything _yet_ because no backends are skipped, but they'll become very important as soon as https://github.com/scipy/scipy/pull/22686 is ready

mdhaber (approved):
Looks good; I'll just take a look at the rendered docs when CI finishes.

scipy/special/tests/test_logsumexp.py
mdhaber: Sure, we've been adding those whether or not something is skipped. It is a sign that the `logsumexp` is participating in the `xp_capabilities` system, and _if_ we want to add a skip, all that needs to be changed is the capabilities table.

PR #22846, opened 2025-04-16
MAINT: _lib/differentiate: update EIM with `at.set`

mdhaber:
#### Reference issue
-

Updates `_elementwise_iterative_method.py` and `_differentiate.py` to use `xpx.at.set` instead of `__setitem__`. This allows `scipy.differentiate` to be used with JAX arrays. Functions in other modules can be updated separately.

#### Additional information
Execution is quite slow with JAX, though.

### DIFF
--- a/scipy/_lib/_elementwise_iterative_method.py
+++ b/scipy/_lib/_elementwise_iterative_method.py
@@ -314,7 +314,7 @@ def _update_active(work, res, res_work_pairs, active, mask, preserve_shape, xp):
     if mask is not None:
         if preserve_shape:
             active_mask = xp.zeros_like(mask)
-            active_mask = xpx.at(active_mask)[active].set(1)
+            active_mask = xpx.at(active_mask)[active].set(True)
             active_mask = active_mask & mask
             for key, val in update_dict.items():
                 try:
--- a/scipy/_lib/_elementwise_iterative_method.py
+++ b/scipy/_lib/_elementwise_iterative_method.py
@@ -16,6 +16,7 @@ import math
 import numpy as np
 from ._util import _RichResult, _call_callback_maybe_halt
 from ._array_api import array_namespace, xp_size, xp_result_type
+import scipy._lib.array_api_extra as xpx
 
 _ESIGNERR = -1
 _ECONVERR = -2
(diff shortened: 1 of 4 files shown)


scipy/differentiate/_differentiate.py
```diff
@@ -477,9 +478,9 @@ def pre_func_eval(work):
         n_new = 2*n if work.nit == 0 else 2  # number of new abscissae
         x_eval = xp.zeros((work.hdir.shape[0], n_new), dtype=work.dtype)
         il, ic, ir = work.il, work.ic, work.ir
-        x_eval[ir] = work.x[ir][:, xp.newaxis] + hr[ir]
-        x_eval[ic] = work.x[ic][:, xp.newaxis] + hc[ic]
-        x_eval[il] = work.x[il][:, xp.newaxis] - hr[il]
+        x_eval = xpx.at(x_eval)[ir].set(work.x[ir][:, xp.newaxis] + hr[ir])
+        x_eval = xpx.at(x_eval)[ic].set(work.x[ic][:, xp.newaxis] + hc[ic])
+        x_eval = xpx.at(x_eval)[il].set( work.x[il][:, xp.newaxis] - hr[il])
```
lucascolley, line 483:
```suggestion
        x_eval = xpx.at(x_eval)[il].set(work.x[il][:, xp.newaxis] - hr[il])
```

lucascolley (approved):
LGTM, thanks Matt!

scipy/_lib/_elementwise_iterative_method.py
```diff
@@ -313,28 +314,28 @@ def _update_active(work, res, res_work_pairs, active, mask, preserve_shape, xp):
...
-            active_mask[active] = 1
+            active_mask = xpx.at(active_mask)[active].set(True)
             active_mask = active_mask & mask
             for key, val in update_dict.items():
                 try:
-                    res[key][active_mask] = val[active_mask]
+                    res[key] = xpx.at(res[key])[active_mask].set(val[active_mask])
                 except (IndexError, TypeError, KeyError):
-                    res[key][active_mask] = val
+                    res[key] = xpx.at(res[key])[active_mask].set(val)
```
crusaderky, line 323: This is going to fail in jax.jit, and the exception filter is so wide that it _might_ quietly suppress the jax tracer error depending on what it subclasses. Unsure what the original intent was?
Is `val` sometimes a scalar? In which case this should be

```python
if getattr(val, ndim, 0) == 0:
    res[key] = xpx.at(res[key])[active_mask].set(val)
else:
    assert val.shape == res[key].shape
    res[key] = xp.where(active_mask, val, res[key])

mdhaber: I thought this might come up. My plan was to fix this `try`-`except` business right after this PR. I happened to have already done this a few days ago, and I didn't want to cause merge conflicts.

Can we judge this on the basis of converting in-place operations to `xpx.at`, and tackle that separately?

crusaderky: works for me

PR #22856, opened 2025-04-17
MAINT: stats: ensure functions work on non-default device

mdhaber:
#### Reference issue
Toward gh-22680

Makes adjustments needed to ensure that `stats` functions work when arrays are not on the default device. Adds tests for device and dtype preservation.

#### Additional information
To do:
- [x] add `device` to `marray` `asarray`
- [ ] ~~preserve device in `special/_support_alternative_backends.py`?~~ *A lot of work has been done on `_support_alternative_backends` since opening this; I don't think I need to do anything here.*
- [ ] ~~ensure common device in `xp_promote`?~~ *Sidestepped. Our private `_asarray` needs to support `device`.*
- [x] add tests for remaining functions


scipy/stats/tests/test_device_dtype.py
```diff
@@ -0,0 +1,138 @@
...
+        assert xp_device(res) == xp_device(x)
+        assert x.dtype == dtype
+
+
+@pytest.mark.parametrize('fun, kwargs',
+    [(stats.moment, {'order': 2}),
+     (stats.skew, {}),
+     (stats.skew, {'bias': False}),
+     (stats.kurtosis, {}),
+     (stats.kurtosis, {'bias': False}),git
```
mdhaber, line 47:
```suggestion
     (stats.kurtosis, {'bias': False}),
```

rgommers:
Thanks Matt. This looks pretty good to me already.

scipy/stats/_morestats.py
```diff
@@ -987,7 +989,8 @@ def boxcox_llf(lmb, data, *, axis=0, keepdims=False, nan_policy='propagate'):
                           result_to_tuple=lambda x: (x,))
 def _boxcox_llf(data, axis=0, *, lmb):
     xp = array_namespace(data)
-    lmb, data = xp_promote(lmb, data, force_floating=True, xp=xp)
+    dtype = xp_result_type(lmb, data, force_floating=True, xp=xp)
+    data = xp.asarray(data, dtype=dtype)
```
mdhaber, line 993: It would be nice if `xp_promote` could convert Python scalar `lmb` to an array on the right device, but `xp_promote` uses our `_asarray`, which doesn't support `device`. I'd like to leave it as-is in this PR, but it's something to keep in mind in gh-22049.

scipy/stats/tests/test_marray.py
```diff
@@ -7,6 +7,7 @@
 from scipy.stats._axis_nan_policy import _axis_nan_policy_factory
 
 
+pytest.skip(allow_module_level=True)  # temporary until `marray` supports `device`
```
mdhaber, line 10: `marray` 0.0.9 added `device`, but broke mdhaber/marray#123, so I pinned us at 0.0.8. I'll fix that soon (thoughts on the best way to do that?) and remove these skips.

scipy/conftest.py
```diff
@@ -440,6 +440,10 @@ def test_device(xp, devices):
         # verbose to skip the test for each jit-capable function and run it for
         # those that only support eager mode.
         pytest.xfail(reason="jax-ml/jax#26000")
+    if is_torch(xp):
+        devices = xp.__array_namespace_info__().devices()
+        # open an issue about this - cannot branch based on `any`/`all`?
+        return (device for device in devices if device.type != 'meta')
```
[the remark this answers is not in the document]
rgommers: Related to https://github.com/data-apis/array-api/issues/945 it looks like. `'meta'` backend is brand new and WIP I think - do we need to worry about it here if CI is green?

scipy/spatial/distance.py
```diff
@@ -2294,7 +2294,8 @@ def pdist(X, metric='euclidean', *, out=None, **kwargs):
 
     X = _asarray(X)
     if X.ndim != 2:
-        raise ValueError(f'A 2-dimensional array must be passed. (Shape was {X.shape}).')
+        raise ValueError('A 2-dimensional array must be passed. '
+                         f'(Shape was {X.shape}).')
```
rgommers, line 2298: It'd be nice to drop the changes to this file, since it's style-only in a different module. Can be done in a separate PR if this bothers your local setup somehow.

scipy/stats/_morestats.py
```diff
@@ -2950,7 +2954,7 @@ def bartlett(*samples, axis=0):
              * ((xp.sum(1/(Ni - 1), axis=0)) - 1/(Ntot - k)))
     T = numer / denom
 
-    chi2 = _SimpleChi2(xp.asarray(k-1))
+    chi2 = _SimpleChi2(xp.asarray(k-1, dtype=dtype))
```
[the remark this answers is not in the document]
rgommers: Looks right. This PR looks about ready to merge.

scipy/conftest.py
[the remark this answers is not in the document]
mdhaber: I'm guessing not many existing tests exercise it because only tests that use the `devices` fixture see it, right? In any case, it causes several of these new tests to fail. 

I don't think I can make all the functions work with it in this PR, so currently I used this like to skip it everywhere. But I'm open to thoughts on a more targeted way to skip or xfail the offending tests.

scipy/spatial/distance.py
mdhaber: Yeah it came up with `spin lint`. I'll take it out.

mdhaber:
Removed changes to `distance.py` and added the one missing use of `device` to `bartlett`. Since `meta` is new and essentially untested, I can open an issue for it. If we want to support it, I think we need a help function to look for it (like `is_meta_array`, since it will require special cases like lazy arrays do), and I think the test skip/`xp_capabilities` system needs to be updated to deal with it.

scipy/conftest.py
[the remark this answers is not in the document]
rgommers: > I'm guessing not many existing tests exercise it because only tests that use the `devices` fixture see it, right?

Yes, that seems right. And I think the choice you made here for skipping at the scipy-global level is the right approach for now.

rgommers:
> Since `meta` is new and essentially untested, I can open an issue for it. If we want to support it, I think we need a help function to look for it (like `is_meta_array`, since it will require special cases like lazy arrays do), and I think the test skip/`xp_capabilities` system needs to be updated to deal with it.

An issue would be great. My current impression is that this is low-priority, and we should leave it completely disabled until we have dealt with JAX/Dask more. Because the PyTorch `'meta'` device has similar requirements to them.

rgommers (approved):
Thanks @mdhaber. I think this looks good in its current form, let's squash-merge when CI comes back green I'd say.

scipy/stats/tests/test_device_dtype.py
```diff
@@ -0,0 +1,240 @@
...
+
+@pytest.mark.skip_xp_backends('dask.array', reason='no take_along_axis')
+@pytest.mark.parametrize('method', ['inverted_cdf', 'averaged_inverted_cdf',
+                                    'closest_observation', 'interpolated_inverted_cdf',
+                                    'hazen', 'weibull', 'linear', 'median_unbiased',
+                                    'normal_unbiased', 'harrell-davis'])
+@pytest.mark.parametrize('dtype', dtypes)
+def test_quantile(method, dtype, xp, devices):
+    if (is_array_api_strict(xp) or is_torch(xp)) and method == 'harrell-davis':
+        pytest.skip("'harrell-davis' not currently not supported on GPU.")
```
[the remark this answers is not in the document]
rgommers: Good point. It's easier to write it with `if ...` rather than `pytest.param` (no need to remember unusual `pytest` syntax), but the `pytest.param` method is a bit more concise and gives better messages.

I'm happy for this to go in either way though, before it accumulates conflicts again.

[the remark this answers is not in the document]
lucascolley: actually, it looks like `cpu_only=True` does not skip array-api-strict's `device1` as that comes via the `devices` fixture, not the `xp` fixture.

@crusaderky is there a better solution already available here?

lucascolley:
merged to avoid conflicts, hopefully https://github.com/scipy/scipy/pull/22856#discussion_r2117688376 can be addressed in a follow-up

scipy/conftest.py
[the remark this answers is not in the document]
crusaderky: Supporting the torch `meta` device requires 
1. adding it to `array_api_compat.is_lazy_array`
2. switching scipy to use array-api-extra `xp_assert_equal` et al. to benefit from https://github.com/data-apis/array-api-extra/pull/300.

scipy/stats/tests/test_device_dtype.py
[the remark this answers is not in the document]
crusaderky: What you're asking for is a flag for all backends that have devices other than CPU. Which doesn't exist, and it's a tad too niche IMHO.
Worth pointing out that, on a non-pixi deployment with CUDA installed,
```python
@pytest.mark.skip_xp_backends(cpu_only=True)
def test1(xp, devices):
    ...
```
will in fact run on torch GPU and JAX GPU.

scipy/stats/tests/test_device_dtype.py
```diff
@@ -0,0 +1,240 @@
...
+     (stats.tvar, {'limits': (0.1, 0.9)}),
+     (stats.tmin, {'lowerlimit': 0.1}),
+     (stats.tmax, {'upperlimit': 0.9}),
+     (stats.tstd, {'limits': (0.1, 0.9)}),
+     (stats.tsem, {'limits': (0.1, 0.9)}),
+     ])
+@pytest.mark.parametrize('dtype', dtypes)
+def test_one_in_one_out(fun, kwargs, dtype, xp, devices):
+    if is_dask(xp) and fun == stats.variation:
+        pytest.skip("dtype inference failing in xpx.apply_where")
```
crusaderky, line 144: @mdhaber this is quite worrysome:
> E           TypeError("Multiple namespaces for array inputs: {<module 'scipy._lib.array_api_compat.numpy' [...], <module 'scipy._lib.array_api_compat.dask.array' [...]

scipy/stats/tests/test_device_dtype.py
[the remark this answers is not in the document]
crusaderky: I drafted a PR that uses `pytest.param(..., marks=skip_xp_backends(...))`, but while it's more idiomatic I found it to be a lot less readable.

scipy/stats/tests/test_device_dtype.py
mdhaber: I'm going to look into it, but it wasn't worth holding up the whole PR. Dask can be very creative about how to fail with code that works for other backends. There are other things I find more worrisome.

rgommers: > Dask can be very creative about how to fail with code that works for other backends.

Yeah I looked at this, and it's very much not obvious what is going on there. All the input variables seemed to be Dask arrays of the same dtype and device, and internally somewhere, one is converted to a numpy array.

mdhaber: @crusaderky I looked and I can't see what's wrong with the inputs to `apply_where`, but the error is happening inside that.

In the call to `map_blocks`:
https://github.com/data-apis/array-api-extra/blob/dd7650aa737f4ded37e5d406067f117e05035ec3/src/array_api_extra/_lib/_funcs.py#L142

- `cond` is a Dask array
- `f1` is a function that operates on Dask arrays and returns Dask arrays (defined in `variation`)
- `f2` is `None`
- `args_` is a tuple of two Dask arrays
- `f1(*args_)` is a Dask array
- `meta_xp` is SciPy's array API compat NumPy

Then if we go inside the call to `_apply_where`, all the array arguments are NumPy arrays, but `f1` is still a function that is supposed to operate on Dask arrays. So [`temp1`](https://github.com/data-apis/array-api-extra/blob/dd7650aa737f4ded37e5d406067f117e05035ec3/src/array_api_extra/_lib/_funcs.py#L160) is a Dask array, and in [`return at(out, cond).set(temp1)`](https://github.com/data-apis/array-api-extra/blob/dd7650aa737f4ded37e5d406067f117e05035ec3/src/array_api_extra/_lib/_funcs.py#L175), it's trying to set elements of a NumPy `out` array with elements of a Dask `temp1` array.

mdhaber: Does the callable `f1` need to accept any type of array and return any type of array? I guess so:
> On Dask, `f1` and `f2` are applied to the individual chunks and should use functions from the namespace of the chunks.

I can probably fix that.

crusaderky: > Does the callable `f1` need to accept any type of array and return any type of array? 

Yes. See https://github.com/data-apis/array-api-extra/discussions/196

crusaderky: - https://github.com/scipy/scipy/pull/23104
- https://github.com/scipy/scipy/pull/23105

PR #22857, opened 2025-04-18
ENH: stats: add array API support to much of `_axis_nan_policy` decorator

mdhaber:
#### Reference issue
Toward gh-20544

Adds array API support to much of the `_axis_nan_policy` decorator. This makes support for `nan_policy`, `keepdims`, tuple `axis`, `SmallSampleWarning`, etc. in stats more consistent across backends.


scipy/stats/_axis_nan_policy.py
```diff
@@ -268,12 +271,12 @@ def _check_empty_inputs(samples, axis):
     return output
 
 
-def _add_reduced_axes(res, reduced_axes, keepdims):
+def _add_reduced_axes(res, reduced_axes, keepdims, xp=np):
```
mdhaber, line 274: The changes in this file are pretty straightforward xp-translations that allow non-numpy arrays to go through the first part of the decorator, which makes a few things much easier for wrapped functions.
  - The dimensions specified by `axis=None` and `axis` tuples are raveled (so functions only have to consider scalar, in `axis`)
  - The remaining `axis` is moved to `-1`, which tends to make indexing along it easier.
  - `keepdims` is supported automatically by adding back any axes that get reduced.

Eventually, I'll let non-numpy arrays pass a little further through the decorator so when the input array is too small (e.g. empty), the right thing happens.

scipy/stats/_axis_nan_policy.py
```diff
@@ -520,34 +523,43 @@ def hypotest_fun_out(*samples, **kwds):
...
-                samples = [sample.reshape(new_shape)
+                samples = [xp.reshape(sample, new_shape)
                            for sample, new_shape in zip(samples, new_shapes)]
             axis = -1  # work over the last axis
+
+            if not is_numpy(xp):
+                res = hypotest_fun_out(*samples, axis=axis, **kwds)
+                res = result_to_tuple(res, n_out)
+                res = _add_reduced_axes(res, reduced_axes, keepdims, xp=xp)
+                return tuple_to_result(*res)
```
mdhaber, line 561: These are all pretty standard invocations; e.g., see 599-602.

scipy/stats/_stats_py.py
```diff
@@ -483,14 +484,18 @@ def _mode_result(mode, count):
...
     if i.shape == ():
-        count = np.asarray(0, dtype=count.dtype)[()] if i else count
+        count = xp.asarray(0, dtype=count.dtype)[()] if i else count
     else:
-        count[i] = 0
+        count = xpx.at(count)[i].set(0)
     return ModeResult(mode, count)
 
 
+@xp_capabilities(skip_backends=[('dask', "can't compute chunk size"),
```
mdhaber, line 496: See top post. Feedback appreciated.

scipy/stats/_stats_py.py
```diff
@@ -555,40 +560,48 @@ def mode(a, axis=0, nan_policy='propagate', keepdims=False):
...
-        vals, cnts = np.unique(a, return_counts=True)
-        modes, counts = vals[cnts.argmax()], cnts.max()
-        return ModeResult(modes[()], counts[()])
+    # if a.ndim == 1:
+    #     # vals, cnts = xp.unique(a, return_counts=True)  # desired NaN behavior
+    #     vals, cnts = xp.unique_counts(a)  # array API
+    #     modes, counts = vals[xp.argmax(cnts)], xp.max(cnts)
+    #     modes = modes[()] if modes.ndim == 0 else modes
+    #     counts = counts[()] if counts.ndim == 0 else counts
+    #     return ModeResult(modes, counts)
```
mdhaber, line 583: See top post. Feedback appreciated.

scipy/stats/_stats_py.py
```diff
@@ -1005,15 +1018,16 @@ def _moment_outputs(kwds, default_order=1):
...
+    return xp.stack(args)
 
 
 # When `order` is array-like with size > 1, moment produces an *array*
 # rather than a tuple, but the zeroth dimension is to be treated like
 # separate outputs. It is important to make the distinction between
 # separate outputs when adding the reduced axes back (`keepdims=True`).
 def _moment_tuple(x, n_out):
-    return tuple(x) if n_out > 1 else (x,)
+    return tuple(x[i, ...] for i in range(x.shape[0])) if n_out > 1 else (x,)
```
mdhaber, line 1030: See code comment. For better or for worse, this is supposed to iterate over the zeroth axis.

rgommers:
> I got a bit carried away, and now this also overhauls `_axis_nan_policy` to work with non-numpy arrays., since `mode` needed it for many of the tests. LMK if this needs to get broken up.

That might be helpful. Now that I'm in the "review lots of array API PRs mode" I'm inclined to keep going, but it's easier to find time to review and merge in one go if they're smaller diffs and more self-contained.

This now has merge conflicts. Splitting it in two when updating may be nice.

mdhaber:
Will do.

scipy/stats/_stats_py.py
```diff
@@ -486,11 +486,12 @@ def _mode_result(mode, count):
     # When a slice is empty, `_axis_nan_policy` automatically produces
     # NaN for `mode` and `count`. This is a reasonable convention for `mode`,
     # but `count` should not be NaN; it should be zero.
-    i = np.isnan(count)
+    xp = array_namespace(mode, count)
```
mdhaber, line 489: These could be reverted

scipy/stats/_stats_py.py
```diff
@@ -1655,10 +1657,6 @@ def skewtest(a, axis=0, nan_policy='propagate', alternative='two-sided'):
 
     n = xp.asarray(_length_nonmasked(a, axis), dtype=b2.dtype, device=xp_device(a))
     n = xpx.at(n, n < 8).set(xp.nan)
-    if xp.any(xp.isnan(n)):
-        message = ("`skewtest` requires at least 8 valid observations;"
-                   "slices with fewer observations will produce NaNs.")
-        warnings.warn(message, SmallSampleWarning, stacklevel=2)
```
mdhaber, line 1661: Small sample warnings are emitted by the decorator for eager backends; no warnings are emitted for lazy backends (so we can remove restrictions on Dask arrays and JAX JIT).

mdhaber:
CI failure is just an unused import; I'll take care of that when there is other feedback to address.

scipy/stats/tests/test_axis_nan_policy.py
```diff
@@ -1414,6 +1415,13 @@ def __init__(self, x, dtype):
         def __array__(self, dtype=None, copy=None):
             return np.asarray(x, dtype=self._dtype)
 
+        def __iter__(self):
+            # I don't know of a canonical way to determine whether an object should
+            # be coerced to a NumPy array or not. Currently, `xp_promote` checks
+            # whether they are iterable, and if so uses `_asarray` with whatever
+            # `xp` is. So for this to get coerced, it needs to be iterable.
+            return iter(self._x)
```
lucascolley, line 1423: seems okay to me. As far as I remember we would just throw anything we didn't know how to handle at `np.asarray` and see what sticks

scipy/stats/_stats_py.py
```diff
@@ -6022,9 +6015,10 @@ def pack_TtestResult(statistic, pvalue, df, alternative, standard_error,
                      estimate):
     # this could be any number of dimensions (including 0d), but there is
     # at most one unique non-NaN value
-    alternative = np.atleast_1d(alternative)  # can't index 0D object
-    alternative = alternative[np.isfinite(alternative)]
-    alternative = alternative[0] if alternative.size else np.nan
+    xp = array_namespace(statistic, pvalue)
+    alternative = xpx.atleast_nd(xp.asarray(alternative), ndim=1, xp=xp)
+    alternative = alternative[xp.isfinite(alternative)]
+    alternative = alternative[0] if xp_size(alternative) else xp.nan
```
lucascolley: completely optional nit but I would write `xp_size(alternative) != 0` for readability

scipy/stats/_axis_nan_policy.py
```diff
@@ -499,8 +505,13 @@ def hypotest_fun_out(*samples, **kwds):
...
             # Extract the things we need here
             try:  # if something is missing
-                samples = [np.atleast_1d(kwds.pop(param))
-                           for param in (params[:n_samp] + kwd_samp)]
+                samples = [kwds.pop(param) for param in (params[:n_samp] + kwd_samp)]
+                xp = array_namespace(*samples)
+                samples = xp_promote(*samples, xp=xp)
+                samples = (samples,) if not isinstance(samples, tuple) else samples
+                atleast_1d = (np.atleast_1d if is_numpy(xp)
+                              else lambda x: xpx.atleast_nd(x, ndim=1))
```
lucascolley: why would we want to do this instead of just `xpx.atleast_nd`? Is it significantly faster (until we check off `atleast_nd` on https://github.com/data-apis/array-api-extra/issues/100)?

scipy/stats/_axis_nan_policy.py
```diff
@@ -623,7 +648,7 @@ def hypotest_fun(x):
                     if is_too_small(samples, kwds):
                         warnings.warn(too_small_nd_omit, SmallSampleWarning,
                                       stacklevel=4)
-                        return np.full(n_out, NaN)
+                        return np.full(n_out, xp.nan, dtype=NaN.dtype)
```
lucascolley: can we just use `np.nan` here for simplicity?

lucascolley (approved):
thanks Matt, LGTM!

scipy/stats/_axis_nan_policy.py
mdhaber: ~~I think it's because we can still have masked arrays at this point? I'll check and see what fails.~~ Not sure why I did that at some point. Not seeing any failures locally, so let's change it and see if CI still passes.

scipy/stats/_axis_nan_policy.py
mdhaber: Not just `xp.nan`; I think it has to have the right dtype. Do you mean leave it as is (`NaN`)? I think at the time some array libraries weren't accepting 0d arrays in `full`. They seem to now, though.

scipy/stats/_stats_py.py
mdhaber: OK, done.

scipy/stats/tests/test_axis_nan_policy.py
mdhaber: Ok. If we want to change something about this, we can in the context of gh-23670, where we probably need to adjust `xp_promote`'s use of `iter` anyway.

scipy/stats/_axis_nan_policy.py
lucascolley: Sorry, I meant use `np` instead of `xp`, since we are NumPy only here and using `np` on the same line?

mdhaber: Got it. Yeah I must have gotten carried away before. Fixed this (just passing `NaN` because I can with NumPy), and reverted the rest.

mdhaber:
W00t! CI is green. Thanks for the review @lucascolley!

PR #22859, opened 2025-04-19
DOC: rewording of "ties" into "tied pairs" for clearer meaning

tomasel:

The use of "ties" in this context is misleading, possibly hinting to only counting tied elements and not actually all the possible tied pairs. This looks more appropriate and also keeps the phrase consistent, together with "concordant pairs" and "discordant pairs" above.

### DIFF
--- a/scipy/stats/_stats_py.py
+++ b/scipy/stats/_stats_py.py
@@ -5617,7 +5617,7 @@ def kendalltau(x, y, *, nan_policy='propagate',
       tau_c = 2 (P - Q) / (n**2 * (m - 1) / m)
 
     where P is the number of concordant pairs, Q the number of discordant
-    pairs, T the number of ties only in `x`, and U the number of ties only in
+    pairs, T the number of tied pairs only in `x`, and U the number of tied pairs only in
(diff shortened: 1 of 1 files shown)


mdhaber (approved):

mdhaber:
Content is fine, but needs to be adjusted for line length.

scipy/stats/_stats_py.py
```diff
@@ -5617,7 +5617,7 @@ def kendalltau(x, y, *, nan_policy='propagate',
       tau_c = 2 (P - Q) / (n**2 * (m - 1) / m)
 
     where P is the number of concordant pairs, Q the number of discordant
-    pairs, T the number of ties only in `x`, and U the number of ties only in
+    pairs, T the number of tied pairs only in `x`, and U the number of tied pairs only in
     `y`.  If a tie occurs for the same pair in both `x` and `y`, it is not
```
lucascolley, line 5621:
```suggestion
    pairs, T the number of tied pairs only in `x`, and U the number of tied pairs only
    in `y`.  If a tie occurs for the same pair in both `x` and `y`, it is not
```

lucascolley:
thanks for your first contribution to SciPy @tomasel !

PR #22863, opened 2025-04-20
TST: reintroduce `eager_warns` and fix free-threading test failures

rgommers:
The first commit reverts gh-22771, which is possible now that `pytest-run-parallel` 0.4.0 is available. The second moves to to `_lib._array_api` where I think it belongs, and uses it to address a few more test failures that occurred a few times in the last few days in the free-threading CI job (see https://github.com/scipy/scipy/issues/22758#issuecomment-2816855461).

### DIFF
--- a/scipy/_lib/_array_api.py
+++ b/scipy/_lib/_array_api.py
@@ -6,9 +6,10 @@ https://data-apis.org/array-api/latest/purpose_and_scope.html
 The SciPy use case of the Array API is described on the following page:
 https://data-apis.org/array-api/latest/use_cases.html#use-case-scipy
 """
-import os
+import contextlib
 import dataclasses
 import functools
+import os
 import textwrap
 
 from collections.abc import Generator, Iterable, Iterator
@@ -39,8 +40,8 @@ from scipy._lib._docscrape import FunctionDoc
 
 __all__ = [
     '_asarray', 'array_namespace', 'assert_almost_equal', 'assert_array_almost_equal',
-    'default_xp', 'is_lazy_array', 'is_marray',
-    'is_array_api_strict', 'is_complex', 'is_cupy', 'is_jax', 'is_numpy', 'is_torch', 
+    'default_xp', 'eager_warns', 'is_lazy_array', 'is_marray',
+    'is_array_api_strict', 'is_complex', 'is_cupy', 'is_jax', 'is_numpy', 'is_torch',
     'SCIPY_ARRAY_API', 'SCIPY_DEVICE', 'scipy_namespace_for',
     'xp_assert_close', 'xp_assert_equal', 'xp_assert_less',
     'xp_copy', 'xp_device', 'xp_ravel', 'xp_size',
@@ -246,7 +247,7 @@ _default_xp_ctxvar: ContextVar[ModuleType] = ContextVar("_default_xp")
 @contextmanager
 def default_xp(xp: ModuleType) -> Generator[None, None, None]:
     """In all ``xp_assert_*`` and ``assert_*`` function calls executed within this
-    context manager, test by default that the array namespace is 
+    context manager, test by default that the array namespace is
     the provided across all arrays, unless one explicitly passes the ``xp=``
     parameter or ``check_namespace=False``.
 
@@ -260,6 +261,17 @@ def default_xp(xp: ModuleType) -> Generator[None, None, None]:
         _default_xp_ctxvar.reset(token)
 
 
+def eager_warns(x, warning_type, match=None):
+    """pytest.warns context manager, but only if x is not a lazy array."""
+    import pytest
+    # This attribute is interpreted by pytest-run-parallel, ensuring that tests that use
+    # `eager_warns` aren't run in parallel (since pytest.warns isn't thread-safe).
+    __thread_safe__ = False  # noqa: F841
+    if is_lazy_array(x):
+        return contextlib.nullcontext()
+    return pytest.warns(warning_type, match=match)
+
+
 def _strict_check(actual, desired, xp, *,
                   check_namespace=True, check_dtype=True, check_shape=True,
                   check_0d=True):
@@ -270,7 +282,7 @@ def _strict_check(actual, desired, xp, *,
             xp = _default_xp_ctxvar.get()
         except LookupError:
             xp = array_namespace(desired)
- 
+
     if check_namespace:
         _assert_matching_namespace(actual, desired, xp)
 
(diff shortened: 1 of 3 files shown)


mdhaber:
Haven't looked into the `TestBoxcox_llf` failure, but if you want to ignore it for now, I can fix it in gh-22857 or whatever that gets split into.

rgommers:
Not sure why the free-threading (parallel=1) job failed with a , since all tests passed. This should be good to go though. Re-running that job to see if it happens again.

scipy/stats/tests/test_stats.py
```diff
@@ -3734,25 +3719,17 @@ def test_skew_propagate_nan(self, xp):
...
         # Skewness of a constant input should be NaN (gh-16061)
         a = xp.asarray([-0.27829495]*10)  # xp.repeat not currently available
-        warn_ctx = (
-            contextlib.nullcontext() if is_lazy_array(a)
-            else pytest.warns(RuntimeWarning, match="Precision loss occurred..."))
 
-        with warn_ctx:
+        with eager_warns(a, RuntimeWarning, match="Precision loss occurred"):
             xp_assert_equal(stats.skew(a), xp.asarray(xp.nan))
-        with warn_ctx:
```
mdhaber, line 3743: This changes the test, but I suppose we're not really testing the warning here. 👍 (We're not testing much, from the looks of it, but that can be considered separately.)

scipy/_lib/_array_api.py
```diff
@@ -260,6 +261,17 @@ def default_xp(xp: ModuleType) -> Generator[None, None, None]:
         _default_xp_ctxvar.reset(token)
 
 
+def eager_warns(x, warning_type, match=None):
```
mdhaber, line 264: If we're going to see this a lot, I'd prefer a signature closer to that of `pytest.warns` (and all of our `xp_` functions), like
```suggestion
def eager_warns(warning_type, match=None, xp=xp):
```

I suppose the reason that [`is_lazy_array`](https://github.com/data-apis/array-api-compat/blob/5e14b53a3558765a8f9b921c72f0249cc0c1c5b9/array_api_compat/common/_helpers.py#L959) accepts an array (rather than having a function `is_lazy(xp)` that accepts a namespace) is that an unknown backend could in principle have both lazy and eager arrays? But I'm not sure that is important to us right now. 

So this is fine to get tests passing, and it's much better than than having ternaries everywhere, but I think it's worth considering a signature adjustment later.

mdhaber (approved):

scipy/_lib/_array_api.py
rgommers: That seems reasonable to me - I think `xp=xp` fits better indeed. I didn't give it much thought here, since I focused on fixing the CI failures and the `eager_warns` signature was a reintroduction of what was previously reviewed.

scipy/_lib/_array_api.py
```diff
@@ -260,6 +261,17 @@ def default_xp(xp: ModuleType) -> Generator[None, None, None]:
...
+    """pytest.warns context manager, but only if x is not a lazy array."""
+    import pytest
+    # This attribute is interpreted by pytest-run-parallel, ensuring that tests that use
+    # `eager_warns` aren't run in parallel (since pytest.warns isn't thread-safe).
+    __thread_safe__ = False  # noqa: F841
+    if is_lazy_array(x):
+        return contextlib.nullcontext()
```
mdhaber, line 271: @rgommers So if we were to change the signature of `eager_warns` to accept `xp` instead of an array, we'd either change this to:

```suggestion:
    if is_lazy_array(xp.asarray(0)):
        return contextlib.nullcontext()
```

or we'd want an `is_lazy_namespace` function that would accept `xp`.

Would this function go in `array-api-compat`, or would it be private in SciPy? I thought that maybe `array-api-compat` would refuse to define an `is_lazy_namespace` function in case arrays from a namespace could be either lazy or eager. But according to the [documentation](https://data-apis.org/array-api-compat/helper-functions.html#array_api_compat.is_lazy_array):

> This function errs on the side of caution for array types that may or may not be lazy, e.g. JAX arrays, by always returning True for them.

This suggests that there might as well be an `is_lazy_namespace` function that also "errs on the side of caution". If using on an unrecognized namespace, it could either raise, or it could call `is_lazy_array(xp.asarray(0))` or similar.

rgommers: Ah, great question. That does seem a little problematic for libraries that have both lazy and eager flavors. Looking at the current implementation, so far we have had no need to deal with `torch.compile` or other PyTorch backends as lazy, but that could easily change - there are fully lazy backends there after all.

I think the best choice is to leave `eager_warns` as is, and not change it to using `xp=xp`. The only upside of the change is consistency in signature with other functions, but this question above shows that the design is better off accepting an array rather than a namespace.

@crusaderky may have things to add here, since he originally wrote this function - but I suspect he considered both options.

crusaderky: At the moment, torch is always eager according to `is_lazy_array`, but that may change pretty soon in the future.
JAX is always treated as lazy by design (eager JAX is considered a development and debug tool of jitted JAX).

Another backend which can be eager or lazy is blosc2. Under the hood, it lazifies chained operations with `numexpr`. It's been off our radar so far due to trivial issues (`blosc2.asarray` doesn't accept an ArrayLike) but I expect it to join the fray pretty soon.

```python
>>>x = blosc2.asarray(np.asarray([1,2,3]))
>>> x
<blosc2.ndarray.NDArray at 0x7b3d2f461900>
>>> x + 1
<blosc2.lazyexpr.LazyExpr at 0x7b3e5bfbb6a0>
```

mdhaber: If `is_lazy_array` will return different results for a given backend in the near future, it sounds fine to leave the `array-api-compat` function alone.

As for the tests, I would prefer the more familiar interface if we can almost always make the distinction based on `xp`. I'd even volunteer to add a third possibility for the context that *ignores* the specified warning (rather than failing if it is emitted or failing if it is not emitted) if we're comfortable with the slight loss in test coverage. Given that the functions being tested will be using the arrays to make the distinction with `is_lazy_array`, it seems unlikely that the warning would not be emitted correctly based on the laziness of the array - and unimportant, since warning behavior for such backends will probably seem inconsistent to users anyway.

mdhaber: Cool. @rgommers does that work for you - assert that the warning is emitted when backends are inherently eager, and ignore the warnings otherwise? If that's OK, I can modify `eager_warns` to accept `xp` by keyword instead of an array as the first argument without changing `is_lazy_array` - it would just manually specify the context for each backend we test.

rgommers: > ignoring the warning instead of asserting that there is none sounds good to me.

This SGTM too.

> (Do we need to test that warnings are not emited for lazy backends, or would there be another failure if the function attempted to warn for a lazy backend?)

None of that I think, that'll just give us a bunch of extra fragility without any upside I'd expect.

I'd be fine with testing eager-only warnings and leaving out the rest completely, or even only testing numpy-only warnings.

PR #22864, opened 2025-04-20
MAINT: linalg.svd: raise correct error message for GESDD when info=-4

j-bowhay:
image
Makes the error message closer to the infomation shown above when `gesdd` returns `info=-4`


ilayn:
This came with LAPACK 3.12 in https://github.com/Reference-LAPACK/lapack/pull/567 but I think it won't hurt to switch early anyways.

j-bowhay:
> This came with LAPACK 3.12 in [Reference-LAPACK/lapack#567](https://github.com/Reference-LAPACK/lapack/pull/567) but I think it won't hurt to switch early anyways.

Indeed, since A is the forth argument even if using an old LAPACK I don't think this would be too misleading

ilayn:
Thanks Jake.

PR #22865, opened 2025-04-20
BUG: optimize: DCSRCH: fix promoting float32 input to float64

melnikovsky:
In some cases the *linesearch* optimization routines in `DCSRCH` cast `dtype` of the argument to (numpy default) `np.float64`.
Suggested fix allows for `np.float32` optimization without `dtype` promotion.


tylerjereddy (commented):
I'll let the `optimize` regulars comment on the idea, but for now I'll just note that the testsuite doesn't seem sensitive to this change locally--any chance a suitable test could be added to enforce this behavior? (probably want to wait for feedback from `optimize` regulars before putting in the time to do that though)

melnikovsky:
In my case the problem was triggered inside a huge optimization problem (hence float32 usage) but I'll try to produce a compact synthetic example.

scipy/optimize/_dcsrch.py
```diff
@@ -480,7 +480,7 @@ def _iterate(self, stp, f, g, task):
             self.stmax = stp + xtrapu * (stp - self.stx)
 
         # Force the step to be within the bounds stpmax and stpmin.
-        stp = np.clip(stp, self.stpmin, self.stpmax)
+        stp = max(min(stp, self.stpmax), self.stpmin)
```
lucascolley, line 483: I'm not familiar with this section of code, but this kind of fix looks unideal to me. Is the problem that `self.stpmax` and/or `self.stpmin` can be `float64` while `stp` is `float32`? If so, can that be resolved higher up the call-chain than here?

melnikovsky: No, the problem is that `stp` is just a Python number. And `np.clip` builds a one-element numpy array with default dtype. An alternative solution would be to pick this single value and use
`stp = np.clip(stp, self.stpmin, self.stpmax).item()`
Honestly, I don't like it. Do you think this is more readable?

lucascolley: ah, thanks, I see now that `stp`, `stpmax`, and `stpmin` are supposed to be Python scalars. In which case, +1 from me to use the builtin `max` and `min`!

lucascolley (commented):
thanks @melnikovsky, these changes seem good to me. We should add a test which fails before this patch and passes afterwards.

You could probably add a simple check following a relevant docstring example, using https://github.com/scipy/scipy/blob/5bc3d8814d566ef328f41cfa69ccd797c68b0d02/scipy/_lib/_array_api.py#L343-L345 which automatically checks for matching dtypes.

scipy/optimize/_dcsrch.py
```diff
@@ -650,7 +650,9 @@ def dcstep(stx, fx, dx, sty, fy, dy, stp, fp, dp, brackt, stpmin, stpmax):
 
         # The case gamma = 0 only arises if the cubic does not tend
         # to infinity in the direction of the step.
-        gamma = s * np.sqrt(max(0, (theta / s) ** 2 - (dx / s) * (dp / s)))
+        gamma = max(0, (theta / s) ** 2 - (dx / s) * (dp / s))
```
ilayn, line 653: gamma is guaranteed to be positive here so the check is not needed.

ilayn (commented):
Just as a reminder; this code was never meant to work with `np.float32` nor the F77 code that this is translated from. While the `dtype` is now preserved, it won't have any guarantee on the accuracy of the result.

scipy/optimize/_dcsrch.py
lucascolley:
```suggestion
        gamma = (theta / s) ** 2 - (dx / s) * (dp / s)
```

melnikovsky (commented):
Errr, then we need to take into account another case I think:
```
if gamma>0:
         gamma = s * np.sqrt(gamma)
else:
         gamma=0
```

lucascolley:
I interpreted https://github.com/scipy/scipy/pull/22865#discussion_r2052312941 as saying that `gamma > 0` will only fail if `gamma` is already `0`, even before the clipping which was here previously

scipy/optimize/_dcsrch.py
ilayn: No max is still needed here for catastrophic cancellations

ilayn:
`s*sqrt(0.0)` will still be 0.0 so maybe I'm not at my best at understanding today but I don't see the issue.

scipy/optimize/_dcsrch.py
lucascolley: ah sorry, I misinterpreted what you meant

scipy/optimize/_dcsrch.py
```diff
@@ -650,7 +650,9 @@ def dcstep(stx, fx, dx, sty, fy, dy, stp, fp, dp, brackt, stpmin, stpmax):
 
         # The case gamma = 0 only arises if the cubic does not tend
         # to infinity in the direction of the step.
-        gamma = s * np.sqrt(max(0, (theta / s) ** 2 - (dx / s) * (dp / s)))
+        gamma = (theta / s) ** 2 - (dx / s) * (dp / s)
+        if gamma > 0:
+            gamma = s * np.sqrt(gamma)
```
lucascolley, line 655:
```suggestion
        gamma = s * np.sqrt(max(0, (theta / s) ** 2 - (dx / s) * (dp / s)))
```

lucascolley (commented):
ah, I think I see the problem; `np.sqrt` is also converting python scalars to `float64`.

@melnikovsky we would need more than just avoiding `np.sqrt` for the `gamma == 0` case, though, right? Can we use `math.sqrt` instead?

melnikovsky:
...while designing a synthetic example I discovered another unnoticed issue. Sorry, I am slow.

melnikovsky:
> ah, I think I see the problem; `np.sqrt` is also converting python scalars to `float64`.

Right, `np.sqrt(0).dtype==np.float64`

> @melnikovsky we would need more than just avoiding `np.sqrt` for the `gamma == 0` case, though, right? Can we use `math.sqrt` instead?

I do not think so: other variables, like `dx`, are numpy arrays with (hopefully) correct `dtype`

lucascolley:
> other variables, like `dx`, are numpy arrays with (hopefully) correct dtype

Okay, no worries. Let's see how many other changes are needed then for a suitable test to pass, if that seems feasible.

melnikovsky:
I hope no other issues remaining with `np.float32` support in **CG** minimizer. A possible test is below:

```
import numpy as np
import scipy

def f(x):
    return(np.sin(x)-x+x**2/1000)

def g(x):
    return(np.cos(x)-1+x/500)

x0=np.array([0.01], dtype=np.float32)
result=scipy.optimize.minimize(f, x0, method='CG', jac=g)
print(result.x,result.x.dtype)

x0=np.array([0.12], dtype=np.float32)
result=scipy.optimize.minimize(f, x0, method='CG', jac=g)
print(result.x,result.x.dtype)
```

I will be away a couple of days, hope to look carefully again into it after I am back.

scipy/optimize/_dcsrch.py
ilayn: No problem, but is the sqrt issue resolved?

scipy/optimize/_optimize.py
```diff
@@ -1442,8 +1442,8 @@ def _minimize_bfgs(fun, x0, args=(), jac=None, callback=None,
         try:
             alpha_k, fc, gc, old_fval, old_old_fval, gfkp1 = \
                      _line_search_wolfe12(f, myfprime, xk, pk, gfk,
-                                          old_fval, old_old_fval, amin=1e-100,
-                                          amax=1e100, c1=c1, c2=c2)
+                                          old_fval, old_old_fval, amin=1e-30,
```
ilayn, line 1445: These numbers need to be defined by conditionals otherwise `np.float64` branch will fail. That is to say, `np.float64` branch should not be affected.

melnikovsky: What is the conventional way to deal with those fundamental constants? `np.finfo` is not helpful here. Just an `if/else` block for both locations? Please advise.

melnikovsky: I used a simple `if/else` block, please have a look

melnikovsky:
> Let's see how many other changes are needed then for a suitable test to pass, if that seems feasible.
I added a test, please review

lucascolley (commented):
looks like the doctests have picked up on a problem:

```
1667     >>> res1 = optimize.fmin_cg(f, x0, fprime=gradf, args=args)
UNEXPECTED EXCEPTION: ValueError("data type <class 'numpy.int64'> not inexact")
  File "/home/circleci/.pyenv/versions/3.11.12/lib/python3.11/doctest.py", line 1355, in __run
    exec(compile(example.source, filename, "single",
  File "<doctest scipy.optimize.fmin_cg[6]>", line 1, in <module>
  File "/home/circleci/repo/build-install/lib/python3.11/site-packages/scipy/optimize/_optimize.py", line 1705, in fmin_cg
    res = _minimize_cg(f, x0, args, fprime, callback=callback, c1=c1, c2=c2,
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/circleci/repo/build-install/lib/python3.11/site-packages/scipy/optimize/_optimize.py", line 1787, in _minimize_cg
    if float(np.finfo(x0.dtype).tiny) > amin:
             ^^^^^^^^^^^^^^^^^^
  File "/home/circleci/.pyenv/versions/3.11.12/lib/python3.11/site-packages/numpy/_core/getlimits.py", line 525, in __new__
    raise ValueError("data type %r not inexact" % (dtype))
ValueError: data type <class 'numpy.int64'> not inexact
```

lucascolley:
You can probably use `xp_result_type(x0, force_floating=True, xp=np).dtype` from https://github.com/scipy/scipy/blob/6d9b6c90c0cd32ddb58ef884be0f9c85a7715cd1/scipy/_lib/_array_api.py#L494C5-L494C19 instead of `x0.dtype`

melnikovsky:
> You can probably use `xp_result_type(x0, force_floating=True, xp=np).dtype` from

Done per your suggestion, seems to be OK.

lucascolley:
the failure at https://github.com/scipy/scipy/actions/runs/14693746974/job/41232707866?pr=22865 looks relevant

lucascolley:
@ilayn would you like to take another look now that this is passing CI?

scipy/optimize/_linesearch.py
```diff
@@ -157,7 +157,7 @@ def scalar_search_wolfe1(phi, derphi, phi0=None, old_phi0=None, derphi0=None,
     if derphi0 is None:
         derphi0 = derphi(0.)
 
-    if old_phi0 is not None and derphi0 != 0:
+    if old_phi0 is not None and abs(derphi0) > abs(phi0 - old_phi0):
```
ilayn, line 160: Not sure why we introduced the right hand side instead of the previous 0 here.

scipy/optimize/_optimize.py
```diff
@@ -1783,6 +1783,13 @@ def _minimize_cg(fun, x0, args=(), jac=None, callback=None,
     old_fval = f(x0)
     gfk = myfprime(x0)
 
+    dtype=xp_result_type(x0, force_floating=True, xp=array_namespace(x0))
+    for amin,amax in ((1e-100,1e100),(1e-30,1e30),(1e-4,1e4)):
+        if float(np.finfo(dtype).tiny) < amin and amax < float(np.finfo(dtype).max):
```
ilayn, line 1788: I don't quite understand what is happening here. I would expect something like 

```python
amin = 1e-100 if dtype == np.float64 else 1e-30
amax = 1e100 if dtype == np.float64 else 1e30
```

scipy/optimize/_linesearch.py
melnikovsky: An overflow can happen in the division next line. If the denominator is bigger than the numerator by the absolute value, then `alpha1=1` anyway.
 I assume (?) I can induce this overflow for any precision, but it was triggered by the test for `float16`.

scipy/optimize/_optimize.py
melnikovsky: This is basically the same - we are looking for the widest bracket interval from the list which lies within `dtype` precision. I am trying to avoid hard-coded `dtype` but support `64/32/16-bit` floats in a uniform way.

scipy/optimize/_dcsrch.py
melnikovsky: I hope it is:

- if `gamma` is negative (or zero), then it is set to a Python scalar zero
- if it is positive (with a correct `dtype`), then a `np.sqrt` is applied to it.

melnikovsky:
> the failure at https://github.com/scipy/scipy/actions/runs/14693746974/job/41232707866?pr=22865 looks relevant

Thank you, it does. I missed the older numpy peculiarities.

scipy/optimize/_optimize.py
ilayn: See above for float16 bit.

scipy/optimize/_linesearch.py
melnikovsky: > you are introducing a tolerance value

No, I don't. The old behavior (assuming no overflow) for `abs(derphi0) <= abs(phi0 - old_phi0)` was:
- to compute `1.01*2*(phi0 - old_phi0)/derphi0` (and get a result `<=-2.02` or `>=2.02`)
- if positive, let `alpha1=1.0`
- if negative, let `alpha1=1.0`, but one line below

The only thing I suggest is to avoid division when an overflow is possible.

melnikovsky: > You can ignore float16 since they don't have enough range to do any meaningful number crunching. It's way too easy to overflow and there are no support in SciPy for half precision. We don't need to create a false impression in the eyes of the user that float16 is somehow supported.

I don't use `float16` myself. Amazingly, `_numdiff.py` explicitly mentions `float16` support (I have not tested whether it does). So, as you say, I was under an "impression that float16 is somehow supported" :-)

Anyway, `fmin_cg` seems to kind of work with `float16` now. We can remove `float16` from the tests so as not to give false hope, however.

andyfaff: `_numdiff.approx_derivative` is not public. Just because it can handle float16 doesn't mean it's a good idea for it to be used for that purpose. It was modified to make it usable for float64/float32/float128.

tylerjereddy:
Looks like this PR hasn't seen activity in 8 months and may still have some touching up needed? Probably not going to come back to life in the day or two before we branch? I'll bump the milestone I think.

melnikovsky:
> may still have some touching up needed?

I’m not sure what, if anything, needs touching up. Any feedback is welcome.

scipy/optimize/_optimize.py
andyfaff: I don't think we'll do this in any of the `_minimize` functions. Instead we'll have a separate function along the lines of:

```
def _linesearch_defaults_for_dtype(dtype):
     kwargs = {}
     # fill out for float16, float32, float64, float128
     # do amax/amin/xtol
     return kwargs
```
    
Because then we can call it from BFGS/CG/etc.

scipy/optimize/_optimize.py
```diff
@@ -1782,6 +1782,13 @@ def _minimize_cg(fun, x0, args=(), jac=None, callback=None,
     old_fval = f(x0)
     gfk = myfprime(x0)
 
+    dtype=xp_result_type(x0, force_floating=True, xp=array_namespace(x0))
```
andyfaff: One thing that's not clear to me - what signifies the floating precision of the problem?

Is it `fun(x0).dtype`, `x0.dtype`, `grad(x0).dtype`, ... ?

If `x0.dtype is float64` there's no point if `fun(x0).dtype is np.float32`.

andyfaff: My thoughts are that it should be the smallest dtype of either `x0` or `f(x0)`

melnikovsky: I don't see how to make sense of an optimization problem where `x0.dtype`, `f(x0).dtype`, and `g(x0).dtype` are different. We must allow expressions like `x0*g(x0)+f(x0)`, right?

Honestly, "smallest dtype" is also ambiguous. I wonder if you meant smallest `max` or smallest `min`?

scipy/optimize/_optimize.py
melnikovsky: Thanks, I have factored out `_line_search_wolfe_bracket_for_dtype()` following your suggestion. A good bracket selection for `float128` is an open question.

`xtol` is not handled, though, because it is not used in CG / not propagated to line search in BFGS

andyfaff:
@melnikovsky I'm thinking I prefer #24228 to this PR. Could you have a look + experiment with that and let me know how things go for you?

melnikovsky:
> I'm thinking I prefer #24228 to this PR. Could you have a look + experiment with that and let me know how things go for you?

They fix separate problems. There is hardly any overlap:

- #24228 returns `np.float64` even for `dtype`-consistent `x0`/`f`/`jac`
- #22865 fails if `jac` is missing and numerical `np.float32` differentiation is needed

Issue #22868, opened 2025-04-22
BUG: re-importing `scipy` fails


smurfix:
### Describe your issue.

`scipy.optimize._highspy._core` is a C module. While it contains internal submodules, most notably `simplex_constants`, it's not a package by the strict definition of Python 3.12's `importlib`.

Importing directly from `scipy.optimize._highspy._core.simplex_constants` thus may fail.

![image](https://github.com/user-attachments/assets/ac25f159-e77c-45ee-ae66-2297ebadbfa3)

### Reproducing Code Example

```python
* install cq-editor
* create a model
* render it
* render it again

Unfortunately this is somewhat non-deterministic, for whatever reason.
```

### Error message

```shell
ModuleNotFoundError: no module named "scipy.optimize._highspy._core.simplex_constants"
```

### SciPy/NumPy/Python version and system information

```shell
Python 3.12.10 | packaged by conda-forge | (main, Apr 10 2025, 22:21:13) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
<__, numpy.__version__, sys.version_info); scipy.show_config()
1.15.2 2.2.5 sys.version_info(major=3, minor=12, micro=10, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    detection method: pkgconfig
    found: true
    include directory: /opt/conda/include
    lib directory: /opt/conda/lib
    name: openblas
    openblas configuration: USE_64BITINT= DYNAMIC_ARCH=1 DYNAMIC_OLDER= NO_CBLAS=
      NO_LAPACK=0 NO_LAPACKE= NO_AFFINITY=1 USE_OPENMP=0 PRESCOTT MAX_THREADS=128
    pc file directory: /opt/conda/lib/pkgconfig
    version: 0.3.21
  lapack:
    detection method: pkgconfig
    found: true
    include directory: /opt/conda/include
    lib directory: /opt/conda/lib
    name: openblas
    openblas configuration: USE_64BITINT= DYNAMIC_ARCH=1 DYNAMIC_OLDER= NO_CBLAS=
      NO_LAPACK=0 NO_LAPACKE= NO_AFFINITY=1 USE_OPENMP=0 PRESCOTT MAX_THREADS=128
    pc file directory: /opt/conda/lib/pkgconfig
    version: 0.3.21
  pybind11:
    detection method: pkgconfig
    include directory: /opt/conda/include
    name: pybind11
    version: 2.13.6
Compilers:
  c:
[...]

PR #22869, opened 2025-04-22
BUG: optimize._highspy: don't import from inside a C module

smurfix:
`scipy.optimize._highspy._core` is a C module. While it contains C modules, most notably `simplex_constants`, it may or may not be a package by the strict definition of Python 3.12's `importlib`. Thus directly importing from it may fail.

#### Reference issue
#22868 

Reloading scipy didn't work reliably.

### DIFF
--- a/scipy/optimize/_linprog_highs.py
+++ b/scipy/optimize/_linprog_highs.py
@@ -23,10 +23,13 @@ from ._highspy._core import(
     HighsDebugLevel,
     ObjSense,
     HighsModelStatus,
-    simplex_constants as s_c,
+    simplex_constants as s_c,  # [1]
 )
 from scipy.sparse import csc_array, vstack, issparse
 
+# [1]: Directly importing from "._highspy._core.simplex_constants"
+# causes problems when reloading.
+# See https://github.com/scipy/scipy/pull/22869 for details.
 
 def _highs_to_scipy_status_message(highs_status, highs_message):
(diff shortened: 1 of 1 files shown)


lucascolley (commented):
Just one question from me over at https://github.com/scipy/scipy/issues/22868#issuecomment-2820862913.

It would be nice to have a test for this to prevent regressions, but I don't know if we have identified a deterministic reproducer.

fyi @HaoZeke

lucascolley:
@tylerjereddy FYI I added the backport-candidate label, if this PR does fix the linked issue then we may want to try to sneak it into 1.15.3

smurfix:
Yes I have confirmed that this patch fixes the problem.
No I didn't get a clean reproducer. It's reasonably consistent once I get the cq-editor into a state where it triggers the problem, but that obviously doesn't work for the CI queue.

lucascolley (approved):

lucascolley:
alright, thanks! Maybe another maintainer will want us to find a test to be added before merge.

If we don't add a test, we should probably add a code comment warning against reverting this change.

smurfix:
> If we don't add a test, we should probably add a code comment warning against reverting this change.

I'lll do that.

rgommers (approved):
LGTM, thanks @smurfix! Explanation looks correct, and I agree we can't reasonably test this in a reliable way.

PR #22873, opened 2025-04-22
ENH: sparse: Support nD sum/mean/min/max/argmin for sparse arrays

dschult:
This PR provides sparse nD array functionality focusing on unary operations.  It splits off part of the large nD PRs #22540 which itself was a split off of #21613.  This PR does:
- Add nD support for sum, mean, min, max, argmmin, argmax
- Adjust validateaxis() to return tuple or None. Not int.
- Replace `_ravel_non_reduced_axes()` with `_convert_to_2d()` cuz we are always converting to a 2d COO anyway.
- Test power()

All this is building on the base submitted by Anushka Suyal -- though much has been rewritten too. 

This PR also rewrites the ravel helper function to construct the 2d arrays (now called `_convert_to_2d()`) that hold the data for the nD arrays. It also simplifies the `nD` part of `min/max` and `sum` with one method each (like `_min_or_max_axis_nd()`). Instead of copying whole methods to the COO class, the approach here is to provide small nD utility methods that the main methods call. Those small methods essentially do this: ravel nd to 2d, apply operator, unravel.

The `dot` series of methods also use this `_convert_to_2d()` interface. So I converted those existing methods to use the new helper function.

Changes by file:
- In `_sputils.py` and `test_sputils.py` we update `validateaxis` to return tuples for 1D axis values. 
- In `_data.py` we enable min/max/argmin/etc to work for nD with nD method calls to the COO class.
- The bulk of the changes are in `_coo.py` and `test_coo.py`.  The tests of nD behavior are in `test_coo.py`, and we have a much smaller footprint in `_coo.py` than currently on main.  That code is either pushed to `_base.py` where much already existed or was duplicatedand was removed via the new helper `_convert_to_2d` function.
- _base.py has methods sum/mean/etc adjusted to work for nD while using calls to small nD helper functions in COO.  
- changes to `test_base.py` include better test coverage of min/max over `explicit=True` kwarg, 
- Changes in wording of error messages lead to a few other test files having changes.

### DIFF
--- a/scipy/sparse/_sputils.py
+++ b/scipy/sparse/_sputils.py
@@ -403,7 +403,7 @@ def isdense(x) -> bool:
     return isinstance(x, np.ndarray)
 
 
-def validateaxis(axis, *, ndim=2) -> int | tuple[int, ...]:
+def validateaxis(axis, *, ndim=2) -> tuple[int, ...] | None:
     if axis is None:
         return None
 
--- a/scipy/sparse/_sputils.py
+++ b/scipy/sparse/_sputils.py
@@ -437,8 +437,6 @@ def validateaxis(axis, *, ndim=2) -> int | tuple[int, ...]:
         raise ValueError("axis tuple has too many elements")
     elif len_axis == ndim:
         return None
-    elif len_axis == 1:
-        return canon_axis[0]
     else:
         return tuple(canon_axis)
 
--- a/scipy/sparse/_base.py
+++ b/scipy/sparse/_base.py
@@ -5,7 +5,7 @@ import numpy as np
 
 from ._sputils import (asmatrix, check_reshape_kwargs, check_shape,
                        get_sum_dtype, isdense, isscalarlike, _todata,
-                       matrix, validateaxis, getdtype, isintlike)
+                       matrix, validateaxis, getdtype)
 from scipy._lib._sparse import SparseABC, issparse
 
 from ._matrix import spmatrix
--- a/scipy/sparse/_base.py
+++ b/scipy/sparse/_base.py
@@ -4,7 +4,7 @@ import math
 import numpy as np
 
 from ._sputils import (asmatrix, check_reshape_kwargs, check_shape,
-                       get_sum_dtype, isdense, isscalarlike,
+                       get_sum_dtype, isdense, isscalarlike, _todata,
                        matrix, validateaxis, getdtype, isintlike)
 from scipy._lib._sparse import SparseABC, issparse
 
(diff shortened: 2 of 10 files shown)


dschult:
Confusing test failures on free-threaded (1) seem to report that the arrays differ on a relative scale of inf. But the printed values for both arrays agree. Is there a good way to track down these errors (perhaps it is not thread safe in some way)? Perhaps it is checking for array agreement before all threads are completed, and the report of the failure is done after the threads are all complete?

```
_______________ ERROR at call of test_sum[shape22-axis22-out22] ________________
scipy/sparse/tests/test_coo.py:962: in test_sum
    assert_equal(res, exp)
E   AssertionError: 
E   Arrays are not equal
E   
E   Mismatched elements: 5 / 5 (100%)
E   Max absolute difference among violations: 6.98210571e+19
E   Max relative difference among violations: inf
E    ACTUAL: array([ 5.369791e+19,  1.334737e+19,  1.093169e+19,  6.982106e+19,
E          -6.520094e+18])
E    DESIRED: array([ 5.369791e+19,  1.334737e+19,  1.093169e+19,  6.982106e+19,
E          -6.520094e+18])
        a          = <COOrdinate sparse array of dtype 'int64'
	with 504 stored elements and shape (3, 7, 8, 5)>
        axis       = (0, 1, 2)
        exp        = array([ 5.36979055e+19,  1.33473720e+19,  1.09316873e+19,  6.98210571e+19,
       -6.52009378e+18])
        out        = array([ 5.36979055e+19,  1.33473720e+19,  1.09316873e+19,  6.98210571e+19,
       -6.52009378e+18])
        res        = array([ 5.36979055e+19,  1.33473720e+19,  1.09316873e+19,  6.98210571e+19,
       -6.52009378e+18])
        rng        = Generator(PCG64) at 0x339D06EE380
        shape      = (3, 7, 8, 5)
```

scipy/sparse/_sputils.py
```diff
@@ -437,8 +437,6 @@ def validateaxis(axis, *, ndim=2) -> int | tuple[int, ...]:
         raise ValueError("axis tuple has too many elements")
     elif len_axis == ndim:
         return None
```
perimosocordiae, line 439: Let's update the function type signature, which should now be `tuple[int, ...] | None`

perimosocordiae (approved):
Let's double-check that we've fixed all the callsites for validateaxis, to make sure we don't assume an integer return anywhere.

dschult:
Checked all use of validateaxis:  All assume output is a tuple or `None`.
(actually only 5)
**_base**: `sum`, `mean`, 
**_dia**: `sum`, 
**_data**: `_min_or_max`, `_argminmax`

Updated the type annotation on `validateaxis`.

Rebased on main.

lucascolley:
> Updated the type annotation on validateaxis.

FYI @jorenham

jorenham (approved):
the typing annotation change looks good 👌🏻

lesteve:
It seems like this PR modified the result of sparse matrix `.mean` to be a numpy array rather than a `np.matrix`, maybe this was not intended?

This was originally noticed in scikit-learn CI that uses scipy dev from scientific-python-nightly-wheels https://github.com/scikit-learn/scikit-learn/issues/31370.

With the following snippet:
```py
import scipy
from scipy import sparse

print(f"{scipy.__version__=}")
m = sparse.csr_matrix([[1, 2, 3], [4, 5, 6]])
print(f"{m.mean(axis=0)=}")
``` 

In the commit associated to this PR c1e2c2c898 this returns a numpy array
```
scipy.__version__='1.16.0.dev0+git20250514.c1e2c2c'
m.mean(axis=0)=array([[2.5, 3.5, 4.5]])
```

In the previous merge commit on `main` 8e7578828d (and 1.15.2 as well) this returns a `np.matrix`
```
scipy.__version__='1.16.0.dev0+git20250514.8e75788'
m.mean(axis=0)=matrix([[2.5, 3.5, 4.5]])
```

jorenham:
> It seems like this PR modified the result of sparse matrix `.mean` to be a numpy array rather than a `np.matrix`, maybe this was not intended?

From the `np.matrix` docs (https://numpy.org/doc/stable/reference/generated/numpy.matrix.html):

> It is no longer recommended to use this class, even for linear algebra. Instead use regular arrays. The class may be removed in the future.

But that being said, since this change isn't reflected in the docstring (which still states that it should return a `np.matrix`), this smells like an accidental breaking change to me, as well.

dschult:
I did not intend for this to change the return type of spmatrix.mean.
I am investigating and hopefully get a have a fix PR soon.

dschult:
PR #23005  has a fix for the `mean` return type error and tests.
Thanks @lesteve

PR #22875, opened 2025-04-23
CI: limit pytest-fail-slow usage to a single CI job

### LINKED ISSUE #20806: Failures for new `pytest-fail-slow` check in Windows CI jobs
Issue to keep track of spurious failures after gh-20672 was merged yesterday. 

This one showed up on gh-17924:
```
 ______________ TestDifferentialEvolutionSolver.test_strategy_fn _______________
[gw0] win32 -- Python 3.12.3 C:\hostedtoolcache\windows\Python\3.12.3\x64\python.exe
Test passed but took too long to run: Duration 1.9977539000001343s > 1.0s
```

rgommers:
Addresses the flakiness of the Windows CI jobs. Implements what I proposed in https://github.com/scipy/scipy/issues/20806#issuecomment-2585776891. I picked the Linux aarch64 job as the only one where it was easy to cleanly add these flags.

Also marks a couple of tests that take a second or more as slow. I ran the job with a limit of 0.7 seconds, and that flagged two of these tests, the rest were fine. That gives 30% room for jitter, with the unchanged 1.0 second limit. In case this turns out to still result in spurious failures, we should remove support completely.

Closes gh-20806

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -233,7 +233,7 @@ jobs:
   #################################################################################
   gcc9:
     # Purpose is to examine builds with oldest-supported gcc and test with pydata/sparse.
-    name: Oldest GCC & pydata/sparse, fast, py3.11/npMin, pip+pytest
+    name: Oldest GCC & pydata/sparse, full, py3.11/npMin, pip+pytest
     needs: get_commit_message
     if: >
       needs.get_commit_message.outputs.message == 1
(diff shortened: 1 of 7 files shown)


.github/workflows/linux.yml
```diff
@@ -233,7 +233,7 @@ jobs:
   #################################################################################
   gcc9:
     # Purpose is to examine builds with oldest-supported gcc and test with pydata/sparse.
-    name: Oldest GCC & pydata/sparse, fast, py3.11/npMin, pip+pytest
+    name: Oldest GCC & pydata/sparse, full, py3.11/npMin, pip+pytest
```
rgommers, line 236: This was just a mistake in the job name I caught. Running `pytest --pyargs scipy` will run the full rather than the fast test suite.

lucascolley (approved):
LGTM! Maybe @mdhaber wants to take a look

mdhaber (approved):
Yup, the timings in the other jobs has been all over the place. Also, we get most of the benefit running only on the fast test suite.

PR #22886, opened 2025-04-25
ENH: signal: filter design functions array API standard support

ev-br:
#### Reference issue

towards https://github.com/scipy/scipy/issues/20678

Continue going through the "filter design" rubric of gh-20678. In this PR,

- [x] `findfreqs`
- [x] `freqs`
- [x] `freqs_zpk`
- [x] `freqz`
- [x] `freqz_zpk`
- [x] `freqs_sos`
- [x] `freqz_sos`

These frequency response functions block pretty much the rest of scipy signal which is still unconverted, so convert them to unblock the rest.

#### Additional information

Two things worth noting:

- Filter design functions heavily rely on numpy polynomials. Since those polynomials are in numpy, there's little chance they will ever get array API compatible---unless somebody extracts them and maintains them, and we add a dependency. With this in mind, it's easiest to just copy-paste the few routines we need, drop them into `scipy/signal/_polyutils.py` and make them array API compatible. If this exercise grows beyong a handful of functions, well we'll be able to reconsider the approach.

- Going from coefficients to zeros and poles relies on `np.roots`, which, in turn, relies on `eigvals`. Which itself is not in the spec, and is not universally available. This might change in medium term, but for now I'm dropping a sort of best-effort routine which will work on torch, jax and numpy; CuPy has its own routine via delegation, and in fact it round-trips to numpy under the hood. So that we don't have to do that here.

### DIFF
--- a/scipy/signal/_filter_design.py
+++ b/scipy/signal/_filter_design.py
@@ -797,16 +797,16 @@ def group_delay(system, w=512, whole=False, fs=2*pi):
     return w, gd
 
 
-def _validate_sos(sos):
+def _validate_sos(sos, xp):
     """Helper to validate a SOS input"""
-    sos = np.asarray(sos)
-    sos = np.atleast_2d(sos)
+    sos = xp.asarray(sos)
+    sos = xpx.atleast_nd(sos, ndim=2, xp=xp)
     if sos.ndim != 2:
         raise ValueError('sos array must be 2D')
     n_sections, m = sos.shape
     if m != 6:
         raise ValueError('sos array must be shape (n_sections, 6)')
-    if not (sos[:, 3] == 1).all():
+    if not xp.all(sos[:, 3] == 1):
         raise ValueError('sos[:, 3] should be all ones')
     return sos, n_sections
 
@@ -925,13 +925,16 @@ def freqz_sos(sos, worN=512, whole=False, fs=2*pi):
     >>> plt.show()
 
     """
+    xp = array_namespace(sos)
+
     fs = _validate_fs(fs, allow_none=False)
 
-    sos, n_sections = _validate_sos(sos)
+    sos, n_sections = _validate_sos(sos, xp)
     if n_sections == 0:
         raise ValueError('Cannot compute frequencies with no sections')
     h = 1.
-    for row in sos:
+    for j in range(sos.shape[0]):
+        row = sos[j, :]
         w, rowh = freqz(row[:3], row[3:], worN=worN, whole=whole, fs=fs)
         h *= rowh
     return w, h
--- a/scipy/signal/_filter_design.py
+++ b/scipy/signal/_filter_design.py
@@ -72,7 +72,12 @@ def _logspace(start, stop, num=50, endpoint=True, base=10.0, dtype=None, *, xp):
             for a in (start, stop, base)
         )
         base = xp.expand_dims(base)
-    y = xp.linspace(start, stop, num=num, endpoint=endpoint)
+    try:
+        result_dt = xp.result_type(start, stop, base)
+    except ValueError:
+        # all of start, stop and base are python scalars
+        result_dt = xp_default_dtype(xp)
+    y = xp.linspace(start, stop, num=num, endpoint=endpoint, dtype=result_dt)
 
     yp = xp.pow(base, y)
     if dtype is None:
@@ -486,6 +491,7 @@ def freqz(b, a=1, worN=512, whole=False, plot=None, fs=2*pi,
     b, a = map(xp.asarray, (b, a))
     if xp.isdtype(a.dtype, 'integral'):
         a = xp.astype(a, xp_default_dtype(xp))
+    res_dtype = xp.result_type(b, a)
 
     b = xpx.atleast_nd(b, ndim=1, xp=xp)
     a = xpx.atleast_nd(a, ndim=1, xp=xp)
@@ -507,7 +513,7 @@ def freqz(b, a=1, worN=512, whole=False, plot=None, fs=2*pi,
         # if include_nyquist is true and whole is false, w should
         # include end point
         w = xp.linspace(0, lastpoint, N,
-                        endpoint=include_nyquist and not whole)
+                        endpoint=include_nyquist and not whole, dtype=res_dtype)
         n_fft = N if whole else 2 * (N - 1) if include_nyquist else 2 * N
         if (xp_size(a) == 1 and (b.ndim == 1 or (b.shape[-1] == 1))
                 and n_fft >= b.shape[0]
@@ -536,7 +542,10 @@ def freqz(b, a=1, worN=512, whole=False, plot=None, fs=2*pi,
                 # Move the first axis of h to the end.
                 h = xp.moveaxis(h, 0, -1)
     else:
-        w = xpx.atleast_nd(xp.asarray(worN), ndim=1, xp=xp)
+        if isinstance(worN, complex):
+            # backwards compat
+            worN = worN.real
+        w = xpx.atleast_nd(xp.asarray(worN, dtype=res_dtype), ndim=1, xp=xp)
         if xp.isdtype(w.dtype, 'integral'):
             w = xp.astype(w, xp_default_dtype(xp))
         del worN
@@ -643,6 +652,8 @@ def freqz_zpk(z, p, k, worN=512, whole=False, fs=2*pi):
 
     z = xpx.atleast_nd(z, ndim=1, xp=xp)
     p = xpx.atleast_nd(p, ndim=1, xp=xp)
+    res_dtype = xp.result_type(z, p)
+    res_dtype = xp.float64 if res_dtype in (xp.float64, xp.complex128) else xp.float32
 
     fs = _validate_fs(fs, allow_none=False)
 
@@ -653,9 +664,9 @@ def freqz_zpk(z, p, k, worN=512, whole=False, fs=2*pi):
 
     if worN is None:
         # For backwards compatibility
-        w = xp.linspace(0, lastpoint, 512, endpoint=False)
+        w = xp.linspace(0, lastpoint, 512, endpoint=False, dtype=res_dtype)
     elif _is_int_type(worN):
-        w = xp.linspace(0, lastpoint, worN, endpoint=False)
+        w = xp.linspace(0, lastpoint, worN, endpoint=False, dtype=res_dtype)
     else:
         w = xp.asarray(worN)
         if xp.isdtype(w.dtype, 'integral'):
@@ -665,7 +676,7 @@ def freqz_zpk(z, p, k, worN=512, whole=False, fs=2*pi):
 
     zm1 = xp.exp(1j * w)
     func = _pu.npp_polyvalfromroots
-    h = xp.asarray(k) * func(zm1, z, xp=xp) / func(zm1, p, xp=xp)
+    h = xp.asarray(k, dtype=res_dtype) * func(zm1, z, xp=xp) / func(zm1, p, xp=xp)
 
     w = w*(fs/(2*pi))
 
--- a/scipy/signal/_filter_design.py
+++ b/scipy/signal/_filter_design.py
@@ -18,7 +18,7 @@ from scipy.signal import _polyutils as _pu
 
 import scipy._lib.array_api_extra as xpx
 from scipy._lib._array_api import (
-    array_namespace, xp_promote, xp_size, xp_default_dtype, is_jax
+    array_namespace, xp_promote, xp_size, xp_default_dtype, is_jax, xp_float_to_complex,
 )
 from scipy._lib.array_api_compat import numpy as np_compat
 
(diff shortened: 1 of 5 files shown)


scipy/signal/_polyutils.py
```diff
@@ -0,0 +1,83 @@
...
+This module contains both "old-style", np.poly1d, routines from the main numpy
+namespace, and "new-style", np.polynomial.polynomial, routines.
+
+To distinguish the two sets, the "new-style" routine names start with `npp_`
+"""
+import scipy._lib.array_api_extra as xpx
+
+
+def polyroots(coef, *, xp):
+    """numpy.roots, best-effor replacement
```
ev-br, line 12: This routine can potentially be considered for `array-api-extra`. That said, it might not be worth it:
- it's basically the `# companion matrix` part below
- the real blocker is `eigvals`: it is not standardized, and it was previously not available on CUDA at all. Now the low-level blocker is removed (https://github.com/cupy/cupy/issues/6359#issuecomment-2425994358), so CuPy is hopefully getting `eig / eigvals` soon (https://github.com/cupy/cupy/pull/8980). If JAX CUDA grows this routine, too, and we can have it in the array API spec, then SciPy can just retain the companion matrix computation.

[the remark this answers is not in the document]
j-bowhay: Great news CuPy has came to their senses and realised the world is not hermitian! If only generalized eigenvalue problems too...

scipy/signal/tests/test_filter_design.py
```diff
@@ -548,49 +549,52 @@ def test_bad_args(self):
...
-        w2, h2 = freqs([1.0], [1.0], None)
+        w1, h1 = freqs(xp.asarray([1.0]), xp.asarray([1.0]))
+        w2, h2 = freqs(xp.asarray([1.0]), xp.asarray([1.0]), None)
         assert_array_almost_equal(w1, w2)
         assert_array_almost_equal(h1, h2)
 
-    def test_w_or_N_types(self):
+    @skip_xp_backends(np_only=True, reason="numpy scalars")
+    def test_w_or_N_types(self, xp):
         # Measure at 8 equally-spaced points
```
j-bowhay, line 598: I wasn't quite sure what the preferred pattern here is? I thought things has converged on not adding the `xp`
```suggestion
    def test_w_or_N_types(self):
        # Measure at 8 equally-spaced points
```

scipy/signal/tests/test_filter_design.py
```diff
@@ -668,237 +681,276 @@ def test_w_or_N_types(self):
...
+        ):
+
             w, h = freqz(b, a, worN=4, whole=True)
-            expected_w = np.linspace(0, 2 * np.pi, 4, endpoint=False)
+            expected_w = xp.linspace(0, 2 * xp.pi, 4, endpoint=False)
             assert_array_almost_equal(w, expected_w)
             assert_array_almost_equal(h, h_whole)
+
             # simultaneously check int-like support
             w, h = freqz(b, a, worN=np.int32(4), whole=True)
```
j-bowhay, line 722: Does it matter this line is still np?

scipy/signal/tests/test_filter_design.py
```diff
@@ -668,237 +681,276 @@ def test_w_or_N_types(self):
...
             w, h = freqz(b, a, worN=w, whole=True)
             assert_array_almost_equal(w, expected_w)
             assert_array_almost_equal(h, h_whole)
 
-    def test_basic_whole(self):
-        w, h = freqz([1.0], worN=8, whole=True)
-        assert_array_almost_equal(w, 2 * np.pi * np.arange(8.0) / 8)
-        assert_array_almost_equal(h, np.ones(8))
+    @skip_xp_backends(np_only=True, reason="numpy scalars")
+    def test_basic3(self, xp):
```
j-bowhay, line 731: The last test also uses a numpy scalar but is not skipped?

scipy/signal/tests/test_filter_design.py
```diff
@@ -668,237 +681,276 @@ def test_w_or_N_types(self):
...
-    def test_broadcasting2(self):
+    def test_broadcasting2(self, xp):
         # Test broadcasting with worN an integer or a 1-D array,
         # b is an n-dimensional array, and a is left at the default value.
         np.random.seed(123)
         b = np.random.rand(3, 5, 1)
+        b = xp.asarray(b)
         for whole in [False, True]:
-            for worN in [16, 17, np.linspace(0, 1, 10)]:
+            for worN in [16, 17, xp.linspace(0, 1, 10)]:   # FIXME
```
j-bowhay, line 860: FIXME?

scipy/signal/tests/test_filter_design.py
```diff
@@ -668,237 +681,276 @@ def test_w_or_N_types(self):
...
+                                   worN=xp.reshape(worN, (-1,)),
                                    whole=whole)
-                    xp_assert_close(ww, worN.ravel(), rtol=1e-14)
-                    xp_assert_close(hh, h[k, :, :].ravel())
+                    xp_assert_close(ww, xp.reshape(worN, (-1,)), rtol=1e-14)
+                    xp_assert_close(hh, xp.reshape(h[k, :, :], (-1,)))
 
-    def test_backward_compat(self):
+    @skip_xp_backends(np_only=True)
+    def test_backward_compat(self, xp):
```
j-bowhay, line 906: Same query about skipping vs not adding `xp`

scipy/signal/tests/test_filter_design.py
```diff
@@ -668,237 +681,276 @@ def test_w_or_N_types(self):
...
             w1, h1 = freqz(b, a, w, fs=fs)
-            w2, h2 = freqz(b, a, 2*pi*np.array(w)/fs)
+            w2, h2 = freqz(b, a, 2*pi*xp.asarray(w, dtype=xp.float64)/ fs)
             xp_assert_close(h1, h2)
-            xp_assert_close(w, w1, check_dtype=False)
+            xp_assert_close(w1, xp.asarray(w), check_dtype=False)
 
-    def test_w_or_N_types(self):
+    @skip_xp_backends(np_only=True, reason="numpy scalars")
+    def test_w_or_N_types(self, xp):
```
j-bowhay, line 953: Same query about skipping vs not adding xp

scipy/signal/tests/test_filter_design.py
```diff
@@ -1117,52 +1199,67 @@ def test_freqz_sos_against_mp(self):
...
 
         # N = 5, whole=True
         w1, h1 = freqz_sos(sos, 5, whole=True, fs=fs)
         w2, h2 = freqz_sos(sos, 5, whole=True)
         xp_assert_close(h1, h2)
-        xp_assert_close(w1, np.linspace(0, fs, 5, endpoint=False))
+        xp_assert_close(w1, xp.linspace(0, fs, 5, endpoint=False))
+
+    @skip_xp_backends(np_only=True, reason="array-likes")
+    def test_fs_param2(self, xp):
```
j-bowhay, line 1244: Same query about skipping vs not adding xp

j-bowhay: Also wasn't sure if we should add array-like coverage for other funcs?

scipy/signal/tests/test_filter_design.py
```diff
@@ -1117,52 +1199,67 @@ def test_freqz_sos_against_mp(self):
...
+                  xp.asarray([50, 123, 230])):
             w1, h1 = freqz_sos(sos, w, fs=fs)
-            w2, h2 = freqz_sos(sos, 2*pi*np.array(w)/fs)
+            w1, h1 = map(xp.asarray, (w1, h1))
+
+            w2, h2 = freqz_sos(sos, 2*pi*xp.asarray(w, dtype=sos.dtype)/fs)
             xp_assert_close(h1, h2)
             xp_assert_close(w, w1, check_dtype=False)
 
+    @skip_xp_backends(np_only=True, reason="numpy scalars")
```
j-bowhay, line 1262: No `xp` arg (guessing it was accidentally missed given the pattern used elsewhere) so no skip required

scipy/signal/tests/test_filter_design.py
```diff
@@ -1185,6 +1282,7 @@ def test_w_or_N_types(self):
             assert_array_almost_equal(w_out, [8])
             assert_array_almost_equal(h, [1])
 
+    @skip_xp_backends(np_only=True)
```
j-bowhay, line 1285: same here?

scipy/signal/tests/test_filter_design.py
```diff
@@ -1287,9 +1403,10 @@ def test_w_or_N_types(self):
...
```
j-bowhay, line 1406: Skip not needed as no `xp` arg?

j-bowhay (commented):
A first pass, looks in pretty good shape

scipy/signal/_polyutils.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+    if coef.shape[0] < 2:
+        return xp.asarray([], dtype=coef.dtype)
+
+    root_func = getattr(xp, 'roots', None)
+    if root_func:
+        # OK on numpy and jax, is broken on cupy
```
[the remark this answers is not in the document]
j-bowhay: Or CuPy delegates so we never reach here?

scipy/signal/_filter_design.py
```diff
@@ -93,27 +120,46 @@ def findfreqs(num, den, N, kind='ba'):
...
+        tz[(xp.abs(tz) < 1e5) & (xp.imag(tz) >= 0)]
+    ))
 
-    w = np.logspace(lfreq, hfreq, N)
+    integ = xp.astype(xp.abs(ez) < 1e-10, ez.dtype) # XXX True->1, False->0
+    hfreq = xp.round(
+        xp.log10(xp.max(3*xp.abs(xp.real(ez) + integ) + 1.5*xp.imag(ez))) + 0.5
+    )
+
+    fudge = 1e-14 if is_jax(xp) else 0
```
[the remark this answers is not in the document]
ev-br: It's just that `np.round(-1.5)` can evaluate to either -1 or -2 depending on the floating-point jitter in the value which computes to -1.5. That jax differs from all other known backends is an experimental fact from running the test suite. 

The root of the problem is of course that the computation here is brittle. Changing it has backwards compat concern, and it's been criticized by the author of `harold` as "should be done very differently were it not for backwards compat". 

So here I opted for a minimal fix to keep the existing tests passing. 
Have added a code comment to this effect, there's also a longer comment in the commit message which introduced this line.

scipy/signal/_polyutils.py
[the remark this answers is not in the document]
ev-br: Exactly. Was rephrasing the comment, but if you've a preferred wording, I'll pick yours.

scipy/signal/_polyutils.py
```diff
@@ -0,0 +1,83 @@
...
+
+# ### Old-style routines ###
+
+
+# https://github.com/numpy/numpy/blob/v2.2.0/numpy/lib/_polynomial_impl.py#L702-L779
+def polyval(p, x, *, xp):
+    """ Old-style polynomial, `np.polyval`
+    """
+    p = xp.asarray(p)
+    x = xp.asarray(x)
```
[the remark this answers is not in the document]
ev-br: Tried locally and tests seem to pass without this `asarray`, so removed here and elsewhere.

scipy/signal/tests/test_filter_design.py
ev-br: To me, missing `xp` looks like a test which we forgot to convert, but no problem, will change here and elsewhere if this is a preferred way now.

scipy/signal/tests/test_filter_design.py
ev-br: Apparently, it doesn't, at least according to the CI?

scipy/signal/tests/test_filter_design.py
ev-br: I'd rather add these new tests in one go in a follow-up TBH.

ev-br:
Thanks Jake! I believe I addressed your review comments.

j-bowhay (approved):
LGMT! There were two numpy only skips left that could be tidied up by removing the `xp` argument but not sure it's worth running another round of CI just for that. Could you get them up while rebasing the next pr?

On to the next!

ev-br:
Thanks for the review Jake!

> There were two numpy only skips left that could be tidied up by removing the xp argument but not sure it's worth running another round of CI just for that. Could you get them up while rebasing the next pr?

Agreed, and done in https://github.com/scipy/scipy/pull/22896/

jorenham:
`freqz_sos` is missing in the PR description list :)

ev-br:
> freqz_sos is missing in the PR description list :)

Thanks, edited it in. (don't you have enough maintainer chops Joren to edit PR descriptions?)

In general, if you need to check what is converted and what isn't in scipy.signal, the ground truth currently is to rerun https://github.com/scipy/scipy/issues/20678#issuecomment-2900866545

jorenham:
> don't you have enough maintainer chops Joren to edit PR descriptions?

my super cow powers are limited to scipy/scipy-stubs :)

PR #22888, opened 2025-04-25
BUG:interpolate.Akima1DInterpolator Improving Numerical Stability

### LINKED ISSUE #22880: BUG: interpolate: Akima1DInterpolator raises ValueError for valid inputs
### Describe your issue.

It is possible to get a `ValueError: `dydx` must contain only finite values.` when calling Akima1DInterpolator with finite-valued x and ys, for example if the input contains a sharp feature at a region where the value of y is also large. This can happen even if (y[i+1]-y[i])/(x[i+1]-x[i]) is finite everywhere and well within the limits of 64 bit floating point precision.  The error message itself is also confusing, because Akima1DInterpolator does not take dydx as an argument at all. 

This appears to occur because of a floating point overflow in line 539/540 of `_cubic.py`, where the multiplication of f1 and m in the numerator can overflow a 64 bit float even if the denominator f12 is also large and would therefore cancel the overflow. It is straightforward to rearrange this equation to avoid the potential for overflow (e.g. (f1/f12)*m1+(f2/f12)*m2 could be one possibility for an alternate arrangement). 

I attached a simple example I created to replicate the behavior. GSL's akima spline implementation performs as expected for this test case, so it should be a case that a similar spline implementation could be reasonably expected to handle without a floating point overflow.

### Reproducing Code Example

```python
import numpy as np
from scipy.interpolate import Akima1DInterpolator

xs = np.arange(1,10)
ys = 1.e160*np.heaviside(xs-4,0.5)

m = (ys[1:]-ys[:-1])/(xs[1:]-xs[:-1])

assert np.all(np.isfinite(m))
assert np.all(m < 1.e-148 * np.finfo(np.float64).max)

akima_spline = Akima1DInterpolator(xs, ys, extrapolate=False,method='makima')
```

### Error message

```shell
conda/envs/Akima/lib/python3.12/site-packages/scipy/interpolate/_cubic.py:539: RuntimeWarning: overflow encountered in multiply
  t[ind] = (f1[ind] * m[(x_ind + 1,) + y_ind] +
conda/envs/Akima/lib/python3.12/site-packages/scipy/interpolate/_cubic.py:540: RuntimeWarning: overflow encountered in multiply
  f2[ind] * m[(x_ind + 2,) + y_ind]) / f12[ind]
  File "scipy_spline_bug_demo.py", line 12, in <module>
    akima_spline = Akima1DInterpolator(xs, ys, extrapolate=False,method='akima')
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "conda/envs/Akima/lib/python3.12/site-packages/scipy/interpolate/_cubic.py", line 542, in __init__
    super().__init__(x, y, t, axis=0, extrapolate=extrapolate)
  File "conda/envs/Akima/lib/python3.12/site-packages/scipy/interpolate/_cubic.py", line 143, in __init__
    x, dx, y, axis, dydx = prepare_input(x, y, axis, dydx)
  File "conda/envs/Akima/lib/python3.12/site-packages/scipy/interpolate/_cubic.py", line 59, in prepare_input
    raise ValueError("`dydx` must contain only finite values.")
ValueError: `dydx` must contain only finite values.
```

```shell
{
  "Compilers": {
    "c": {
      "name": "clang",
      "linker": "ld64",
      "version": "18.1.8",
      "commands": "arm64-apple-darwin20.0.0-clang",
      "args": "-ftree-vectorize, -fPIC, -fstack-protector-strong, -O2, -pipe, -isystem, /Users/mcdigman/conda/envs/Akima/include, -fdebug-prefix-map=/Users/runner/miniforge3/conda-bld/scipy-split_1739790747129/work=/usr/local/src/conda/scipy-split-1.15.2, -fdebug-prefix-map=/Users/mcdigman/conda/envs/Akima=/usr/local/src/conda-prefix, -D_FORTIFY_SOURCE=2, -isystem, /Users/mcdigman/conda/envs/Akima/include, -mmacosx-version-min=11.0, -mmacosx-version-min=11.0",
      "linker args": "-Wl,-headerpad_max_install_names, -Wl,-dead_strip_dylibs, -Wl,-rpath,/Users/mcdigman/conda/envs/Akima/lib, -L/Users/mcdigman/conda/envs/Akima/lib, -ftree-vectorize, -fPIC, -fstack-protector-strong, -O2, -pipe, -isystem, /Users/mcdigman/conda/envs/Akima/include, -fdebug-prefix-map=/Users/runner/miniforge3/conda-bld/scipy-split_1739790747129/work=/usr/local/src/conda/scipy-split-1.15.2, -fdebug-prefix-map=/Users/mcdigman/conda/envs/Akima=/usr/local/src/conda-prefix, -D_FORTIFY_SOURCE=2, -isystem, /Users/mcdigman/conda/envs/Akima/include, -mmacosx-version-min=11.0, -mmacosx-version-min=11.0"
    },
    "cython": {
      "name": "cython",
      "linker": "cython",
      "version": "3.0.12",
      "commands": "cython"
    },
    "c++": {
      "name": "clang",
      "linker": "ld64",
      "version": "18.1.8",
      "commands": "arm64-apple-darwin20.0.0-clang++",
      "args": "-ftree-vectorize, -fPIC, -fstack-protector-strong, -O2, -pipe, -stdlib=libc++, -fvisibility-inlines-hidden, -fmessage-length=0, -isystem, /Users/mcdigman/conda/envs/Akima/include, -fdebug-prefix-map=/Users/runner/miniforge3/conda-bld/scipy-split_1739790747129/work=/usr/local/src/conda/scipy-split-1.15.2, -fdebug-prefix-map=/Users/mcdigman/conda/envs/Akima=/usr/local/src/conda-prefix, -D_FORTIFY_SOURCE=2, -isystem, /Users/mcdigman/conda/envs/Akima/include, -mmacosx-version-min=11.0, -mmacosx-version-min=11.0",
      "linker args": "-Wl,-headerpad_max_install_names, -Wl,-dead_strip_dylibs, -Wl,-rpath,/Users/mcdigman/conda/envs/Akima/lib, -L/Users/mcdigman/conda/envs/Akima/lib, -ftree-vectorize, -fPIC, -fstack-protector-strong, -O2, -pipe, -stdlib=libc++, -fvisibility-inlines-hidden, -fmessage-length=0, -isystem, /Users/mcdigman/conda/envs/Akima/include, -fdebug-prefix-map=/Users/runner/miniforge3/conda-bld/scipy-split_1739790747129/work=/usr/local/src/conda/scipy-split-1.15.2, -fdebug-prefix-map=/Users/mcdigman/conda/envs/Akima=/usr/local/src/conda-prefix, -D_FORTIFY_SOURCE=2, -isystem, /Users/mcdigman/conda/envs/Akima/include, -mmacosx-version-min=11.0, -mmacosx-version-min=11.0"
    },
    "fortran": {
      "name": "gcc",
      "linker": "ld64",
      "version": "13.2.0",
      "commands": "/Users/runner/miniforge3/conda-bld/scipy-split_1739790747129/_build_env/bin/arm64-apple-darwin20.0.0-gfortran",
      "args": "-march=armv8.3-a, -ftree-vectorize, -fPIC, -fno-stack-protector, -O2, -pipe, -isystem, /Users/mcdigman/conda/envs/Akima/include, -fdebug-prefix-map=/Users/runner/miniforge3/conda-bld/scipy-split_1739790747129/work=/usr/local/src/conda/scipy-split-1.15.2, -fdebug-prefix-map=/Users/mcdigman/conda/envs/Akima=/usr/local/src/conda-prefix",
      "linker args": "-Wl,-headerpad_max_install_names, -Wl,-dead_strip_dylibs, -Wl,-rpath,/Users/mcdigman/conda/envs/Akima/lib, -L/Users/mcdigman/conda/envs/Akima/lib, -march=armv8.3-a, -ftree-vectorize, -fPIC, -fno-stack-protector, -O2, -pipe, -isystem, /Users/mcdigman/conda/envs/Akima/include, -fdebug-prefix-map=/Users/runner/miniforge3/conda-bld/scipy-split_1739790747129/work=/usr/local/src/conda/scipy-split-1.15.2, -fdebug-prefix-map=/Users/mcdigman/conda/envs/Akima=/usr/local/src/conda-prefix"
    },
    "pythran": {
      "version": "0.17.0",
      "include directory": "../../_build_env/venv/lib/python3.12/site-packages/pythran"
    }
  },
  "Machine Information": {
    "host": {
      "cpu": "arm64",
      "family": "aarch64",
      "endian": "little",
      "system": "darwin"
    },
    "build": {
      "cpu": "x86_64",
      "family": "x86_64",
      "endian": "little",
      "system": "darwin"
    },
    "cross-compiled": true
  },
  "Build Dependencies": {
    "blas": {
      "name": "blas",
      "found": true,
      "version": "3.9.0",
      "include directory": "/Users/mcdigman/conda/envs/Akima/include",
      "lib directory": "/Users/mcdigman/conda/envs/Akima/lib",
      "pc file directory": "/Users/mcdigman/conda/envs/Akima/lib/pkgconfig"
    },
    "lapack": {
      "name": "lapack",
      "found": true,
      "version": "3.9.0",
      "include directory": "/Users/mcdigman/conda/envs/Akima/include",
      "lib directory": "/Users/mcdigman/conda/envs/Akima/lib",
      "pc file directory": "/Users/mcdigman/conda/envs/Akima/lib/pkgconfig"
    },
    "pybind11": {
      "name": "pybind11",
      "version": "2.13.6",
      "include directory": "/Users/mcdigman/conda/envs/Akima/include"
    }
  },
  "Python Information": {
    "path": "/Users/runner/miniforge3/conda-bld/scipy-split_1739790747129/_build_env/bin/python3.12",
    "version": "3.12"
  }
}
```

### LINKED ISSUE #22882: BUG: interpolate: Corner finding in Akima1DInterpolator silently fails
### Describe your issue.

Related but separate issue from #22880, but this one is far more likely to occur in realistic use.  

If the dynamic range of the input ys to Akima1DInterpolator is more than a factor of ~10^9, the corner finding algorithm is _completely_ inoperable for smaller input ys, even if the corner at the smaller input y occurs in a completely non-overlapping sub-spline. While it is possibly reasonable to expect that the corner-finding could fail for a dynamic range of 10^9 in a single sub-spline (e.g, +/-2 points to either sides), there should not be any non-local effect on other sub-splines. 

This occurs because the line

`ind = np.nonzero(f12 > 1e-9 * np.max(f12, initial=-np.inf))`

considers the maximum value of f12 _anywhere_, not just the maximum value in the same vicinity, so if the maximum value anywhere is more than 10^9 larger it cannot detect any corners. 

I attached a test case that demonstrates the misbehavior using two widely separated step functions, and a plot of the misbehaving region of the interpolation in the test case. 

I am not sure if the best mitigation would be to only consider the maximum of nearby f12 values (I think just within an index +/- 1 would be sufficient?) , or a different change to how this check works. I think 

`ind = np.nonzero(f12 > 1.e-9 * dm[1:-1])`

might better reflect the intended behavior of the check. This alternate proposed check would at least work as expected for the attached test case, though I have not exhaustively considered how it might behave in the case of other possible corner behavior. 

Image

### Reproducing Code Example

```python
import numpy as np
from scipy.interpolate import Akima1DInterpolator

xs = 1.*np.arange(1,25)                  # grid of spline points
xs_sample = np.linspace(1.,14.,100)      # grid of points to sample splines at
ys1 = np.heaviside(xs-6,0.5)             # small step function
ys2 = ys1 + 1.e9*np.heaviside(xs-16,0.5) # increase of at least ~1e9, widely separated

# create spline objects with both ys
# spline that behaves as expected
akima_spline1 = Akima1DInterpolator(xs, ys1, extrapolate=False,method='akima')
ys_expected = akima_spline1(xs_sample)

# spline that does not behave as expected
akima_spline2 = Akima1DInterpolator(xs, ys2, extrapolate=False,method='akima')
ys_got = akima_spline2(xs_sample)


print('Should be True: ',np.allclose(ys_expected,ys_got,atol=1.e-2,rtol=1.e-2))
print('Should be True: ',np.all(ys_expected==ys_got))

plt.scatter(xs[:10],ys2[:10], label='spline points')
plt.plot(xs_sample,ys_expected, label='expected interpolation')
plt.plot(xs_sample,ys_got, label='actual interpolation')
plt.legend()
plt.xlabel('x')
plt.ylabel('y')
plt.show()
```

### Error message

```shell
Should be True:  False
Should be True:  False
```

```shell
{
  "Compilers": {
    "c": {
      "name": "clang",
      "linker": "ld64",
      "version": "18.1.8",
      "commands": "arm64-apple-darwin20.0.0-clang",
      "args": "-ftree-vectorize, -fPIC, -fstack-protector-strong, -O2, -pipe, -isystem, /Users/mcdigman/conda/envs/Akima/include, -fdebug-prefix-map=/Users/runner/miniforge3/conda-bld/scipy-split_1739790747129/work=/usr/local/src/conda/scipy-split-1.15.2, -fdebug-prefix-map=/Users/mcdigman/conda/envs/Akima=/usr/local/src/conda-prefix, -D_FORTIFY_SOURCE=2, -isystem, /Users/mcdigman/conda/envs/Akima/include, -mmacosx-version-min=11.0, -mmacosx-version-min=11.0",
      "linker args": "-Wl,-headerpad_max_install_names, -Wl,-dead_strip_dylibs, -Wl,-rpath,/Users/mcdigman/conda/envs/Akima/lib, -L/Users/mcdigman/conda/envs/Akima/lib, -ftree-vectorize, -fPIC, -fstack-protector-strong, -O2, -pipe, -isystem, /Users/mcdigman/conda/envs/Akima/include, -fdebug-prefix-map=/Users/runner/miniforge3/conda-bld/scipy-split_1739790747129/work=/usr/local/src/conda/scipy-split-1.15.2, -fdebug-prefix-map=/Users/mcdigman/conda/envs/Akima=/usr/local/src/conda-prefix, -D_FORTIFY_SOURCE=2, -isystem, /Users/mcdigman/conda/envs/Akima/include, -mmacosx-version-min=11.0, -mmacosx-version-min=11.0"
    },
    "cython": {
      "name": "cython",
      "linker": "cython",
      "version": "3.0.12",
      "commands": "cython"
    },
    "c++": {
      "name": "clang",
      "linker": "ld64",
      "version": "18.1.8",
      "commands": "arm64-apple-darwin20.0.0-clang++",
      "args": "-ftree-vectorize, -fPIC, -fstack-protector-strong, -O2, -pipe, -stdlib=libc++, -fvisibility-inlines-hidden, -fmessage-length=0, -isystem, /Users/mcdigman/conda/envs/Akima/include, -fdebug-prefix-map=/Users/runner/miniforge3/conda-bld/scipy-split_1739790747129/work=/usr/local/src/conda/scipy-split-1.15.2, -fdebug-prefix-map=/Users/mcdigman/conda/envs/Akima=/usr/local/src/conda-prefix, -D_FORTIFY_SOURCE=2, -isystem, /Users/mcdigman/conda/envs/Akima/include, -mmacosx-version-min=11.0, -mmacosx-version-min=11.0",
      "linker args": "-Wl,-headerpad_max_install_names, -Wl,-dead_strip_dylibs, -Wl,-rpath,/Users/mcdigman/conda/envs/Akima/lib, -L/Users/mcdigman/conda/envs/Akima/lib, -ftree-vectorize, -fPIC, -fstack-protector-strong, -O2, -pipe, -stdlib=libc++, -fvisibility-inlines-hidden, -fmessage-length=0, -isystem, /Users/mcdigman/conda/envs/Akima/include, -fdebug-prefix-map=/Users/runner/miniforge3/conda-bld/scipy-split_1739790747129/work=/usr/local/src/conda/scipy-split-1.15.2, -fdebug-prefix-map=/Users/mcdigman/conda/envs/Akima=/usr/local/src/conda-prefix, -D_FORTIFY_SOURCE=2, -isystem, /Users/mcdigman/conda/envs/Akima/include, -mmacosx-version-min=11.0, -mmacosx-version-min=11.0"
    },
    "fortran": {
      "name": "gcc",
      "linker": "ld64",
      "version": "13.2.0",
      "commands": "/Users/runner/miniforge3/conda-bld/scipy-split_1739790747129/_build_env/bin/arm64-apple-darwin20.0.0-gfortran",
      "args": "-march=armv8.3-a, -ftree-vectorize, -fPIC, -fno-stack-protector, -O2, -pipe, -isystem, /Users/mcdigman/conda/envs/Akima/include, -fdebug-prefix-map=/Users/runner/miniforge3/conda-bld/scipy-split_1739790747129/work=/usr/local/src/conda/scipy-split-1.15.2, -fdebug-prefix-map=/Users/mcdigman/conda/envs/Akima=/usr/local/src/conda-prefix",
      "linker args": "-Wl,-headerpad_max_install_names, -Wl,-dead_strip_dylibs, -Wl,-rpath,/Users/mcdigman/conda/envs/Akima/lib, -L/Users/mcdigman/conda/envs/Akima/lib, -march=armv8.3-a, -ftree-vectorize, -fPIC, -fno-stack-protector, -O2, -pipe, -isystem, /Users/mcdigman/conda/envs/Akima/include, -fdebug-prefix-map=/Users/runner/miniforge3/conda-bld/scipy-split_1739790747129/work=/usr/local/src/conda/scipy-split-1.15.2, -fdebug-prefix-map=/Users/mcdigman/conda/envs/Akima=/usr/local/src/conda-prefix"
    },
    "pythran": {
      "version": "0.17.0",
      "include directory": "../../_build_env/venv/lib/python3.12/site-packages/pythran"
    }
  },
  "Machine Information": {
    "host": {
      "cpu": "arm64",
      "family": "aarch64",
      "endian": "little",
      "system": "darwin"
    },
    "build": {
      "cpu": "x86_64",
      "family": "x86_64",
      "endian": "little",
      "system": "darwin"
    },
    "cross-compiled": true
  },
  "Build Dependencies": {
    "blas": {
      "name": "blas",
      "found": true,
      "version": "3.9.0",
      "include directory": "/Users/mcdigman/conda/envs/Akima/include",
      "lib directory": "/Users/mcdigman/conda/envs/Akima/lib",
      "pc file directory": "/Users/mcdigman/conda/envs/Akima/lib/pkgconfig"
    },
    "lapack": {
      "name": "lapack",
      "found": true,
      "version": "3.9.0",
      "include directory": "/Users/mcdigman/conda/envs/Akima/include",
      "lib directory": "/Users/mcdigman/conda/envs/Akima/lib",
      "pc file directory": "/Users/mcdigman/conda/envs/Akima/lib/pkgconfig"
    },
    "pybind11": {
      "name": "pybind11",
      "version": "2.13.6",
      "include directory": "/Users/mcdigman/conda/envs/Akima/include"
    }
  },
  "Python Information": {
    "path": "/Users/runner/miniforge3/conda-bld/scipy-split_1739790747129/_build_env/bin/python3.12",
    "version": "3.12"
  }
}
```

mcdigman:
Addresses two numerical stability concerns in the initialization of Akima Splines 

Adds two unit tests that demonstrate the fixes of the code examples in #22880 and #22882

#### Reference issue
closes #22880
closes #22882

The changes are:

1. Rearranging the equation for t in such a way that the denominator is not as susceptible to becoming enormous 
2. Handling setting the slope in the dm[1:-1]==0. case separately/explicitly 
3. Setting the finite threshold for switching formulas as the denominator of the slope approaches zero to identically 0 for method='makima', because per the citation in the documentation the method is specifically constructed to not require such a switch (see comment in #22882). 
4. Changing the scaling multiplier on the threshold to a different formula that only consider points in the same subspline. I think dm[1:-1] is a better choice than max(f12) anyway, because it implicitly appears twice in the formula for the slope, and so actually sets the scale for when the formula could potentially become numerically inaccurate.

#### Additional information

Note that in my opinion the zero-denominator threshold could also just be set to zero _always_ (i.e., also for method='akima'); the formula will be equally qualitatively discontinuous somewhere no matter where the threshold is set, and making the threshold nonzero expands the region where the formula is different to a surface with a far larger volume of points, instead of just a single point discontinuity at denom=0. Having a finite switch does not meaningfully improve the numerical stability as far as I can tell. The implementation in e.g. GSL does not have a finite cutoff like this. However, for now I left the finite cutoff present in this pull request because this is my first pull request to scipy and I do not really know whether the review standards for approving a change that (albeit slightly) changes the behavior differ from those for just a bug fix.

### DIFF
--- a/scipy/interpolate/_cubic.py
+++ b/scipy/interpolate/_cubic.py
@@ -547,7 +547,7 @@ class Akima1DInterpolator(CubicHermiteSpline):
             f12 = f1 + f2
 
             # mask to handle m2=m3 case exactly
-            ind = np.nonzero((dm[1:-1] == 0.))
+            ind = np.nonzero(dm[1:-1] == 0.)
             x_ind, y_ind = ind[0], ind[1:]
             t[ind] = m[(x_ind + 1,) + y_ind]
 
--- a/scipy/interpolate/_cubic.py
+++ b/scipy/interpolate/_cubic.py
@@ -535,16 +535,29 @@ class Akima1DInterpolator(CubicHermiteSpline):
                 pm = np.abs(m[1:] + m[:-1])
                 f1 = dm[2:] + 0.5 * pm[2:]
                 f2 = dm[:-2] + 0.5 * pm[:-2]
+                # makima is more numerically stable for small f12,
+                # so a finite cutoff should not improve any behavior
+                break_mult = 0.0
             else:
                 f1 = dm[2:]
                 f2 = dm[:-2]
+                # akima has a qualitative discontinuity near f12=0
+                # a finite cutoff moves it, but cannot remove it.
+                break_mult = 1.e-9
             f12 = f1 + f2
+
+            # mask to handle m2=m3 case exactly
+            ind = np.nonzero((dm[1:-1] == 0.))
+            x_ind, y_ind = ind[0], ind[1:]
+            t[ind] = m[(x_ind + 1,) + y_ind]
+
             # These are the mask of where the slope at breakpoint is defined:
-            ind = np.nonzero(f12 > 1e-9 * np.max(f12, initial=-np.inf))
+            ind = np.nonzero((dm[1:-1] != 0.) & (f12 > break_mult * dm[1:-1]))
             x_ind, y_ind = ind[0], ind[1:]
             # Set the slope at breakpoint
-            t[ind] = (f1[ind] * m[(x_ind + 1,) + y_ind] +
-                    f2[ind] * m[(x_ind + 2,) + y_ind]) / f12[ind]
+            t[ind] = m[(x_ind + 1,) + y_ind] + (
+                    (f2[ind] / f12[ind]) * (
+                        m[(x_ind + 2,) + y_ind] - m[(x_ind + 1,) + y_ind]))
 
         super().__init__(x, y, t, axis=0, extrapolate=extrapolate)
         self.axis = axis
--- a/scipy/interpolate/tests/test_interpolate.py
+++ b/scipy/interpolate/tests/test_interpolate.py
@@ -1052,15 +1052,19 @@ class TestAkima1DInterpolator:
 
         y_eval1_1 = ak1(x_sample1)
         y_eval2_1 = ak2(x_sample1)
+        y_eval3_1 = x_sample1 - 3  # the known result for this segment
 
         y_eval1_2 = ak1(x_sample2)
         y_eval2_2 = ak2(x_sample2)
+        y_eval3_2 = 6 - x_sample2  # the known result for this segment
 
         xp_assert_equal(np.isnan(y_eval1_1), np.isnan(y_eval2_1))
         xp_assert_close(y_eval1_1[~np.isnan(y_eval1_1)], y_eval2_1[~np.isnan(y_eval1_1)])
+        xp_assert_close(y_eval3_1, y_eval1_1)
 
         xp_assert_equal(np.isnan(y_eval1_2), np.isnan(y_eval2_2))
         xp_assert_close(y_eval1_2[~np.isnan(y_eval1_2)], y_eval2_2[~np.isnan(y_eval1_2)])
+        xp_assert_close(y_eval3_2, y_eval1_2)
 
 
     def test_no_overflow(self):
--- a/scipy/interpolate/tests/test_interpolate.py
+++ b/scipy/interpolate/tests/test_interpolate.py
@@ -993,7 +993,7 @@ class TestAkima1DInterpolator:
         # Testing extrapoation to actual function.
         xp_assert_close(y_ext, ak_true(x_ext), atol=1e-15)
 
-    def test_large_dynamic_rannge(self):
+    def test_large_dynamic_range(self):
         # check a large step does not change non-overlapping subsplines
         x = 1.*np.arange(1, 12)                    # grid of spline points
         x_sample = np.linspace(1., 7., 21)         # grid of points to sample splines at
@@ -1014,6 +1014,24 @@ class TestAkima1DInterpolator:
 
         xp_assert_equal(y_eval1, y_eval2)
         xp_assert_equal(y_eval3, y_eval4)
+
+    def test_large_dynamic_range2(self):
+        # variant of test_large_dynamic_range specifically engineered to fail if m2=m3 handling is not on
+        x = np.linspace(-1., 9., num=40)
+        x_sample = np.linspace(0.,3.,41)
+
+        y1 = np.heaviside(x - 2, 0.5)                # small step function
+        y2 = y1 + 1.e9 * np.heaviside(x - 7, 0.5)    # very large increase
+
+        ak1 = Akima1DInterpolator(x, y1, method='akima')
+
+        ak2 = Akima1DInterpolator(x, y2, method='akima')
+
+        y_eval1 = ak1(x_sample)
+        y_eval2 = ak2(x_sample)
+
+        xp_assert_equal(np.isnan(y_eval1), np.isnan(y_eval2))
+        xp_assert_close(y_eval1[~np.isnan(y_eval2)], y_eval2[~np.isnan(y_eval2)])
 
     def test_no_overflow(self):
         # check a large jump does not cause a float overflow
--- a/scipy/interpolate/tests/test_interpolate.py
+++ b/scipy/interpolate/tests/test_interpolate.py
@@ -1033,6 +1033,36 @@ class TestAkima1DInterpolator:
         xp_assert_equal(np.isnan(y_eval1), np.isnan(y_eval2))
         xp_assert_close(y_eval1[~np.isnan(y_eval2)], y_eval2[~np.isnan(y_eval2)])
 
+    def test_large_dynamic_range3(self):
+        # variant of test_large_dynamic_range that more specifically isolates the failure if m2=m3 handling is not on
+        x = np.arange(0, 14)
+
+        # evaluate at x points chosen specifically to isolate only the misbehavior in m2=m3 handling
+        x_sample1 = np.linspace(1.,3.,11)
+        x_sample2 = np.linspace(6.,8.,11)
+
+        # points to create spline
+        y1 = np.array([-3, -2, -1, 0, 1, 0, 0, -1, -2, -3, -4, -5, -6, -7])
+        y2 = y1.copy()
+        y2[-1] -= 1.e10  # make a copy identical except for one very large value at one end
+
+        ak1 = Akima1DInterpolator(x, y1, method='akima')
+
+        ak2 = Akima1DInterpolator(x, y2, method='akima')
+
+        y_eval1_1 = ak1(x_sample1)
+        y_eval2_1 = ak2(x_sample1)
+
+        y_eval1_2 = ak1(x_sample2)
+        y_eval2_2 = ak2(x_sample2)
+
+        xp_assert_equal(np.isnan(y_eval1_1), np.isnan(y_eval2_1))
+        xp_assert_close(y_eval1_1[~np.isnan(y_eval1_1)], y_eval2_1[~np.isnan(y_eval1_1)])
+
+        xp_assert_equal(np.isnan(y_eval1_2), np.isnan(y_eval2_2))
+        xp_assert_close(y_eval1_2[~np.isnan(y_eval1_2)], y_eval2_2[~np.isnan(y_eval1_2)])
+
+
     def test_no_overflow(self):
         # check a large jump does not cause a float overflow
         x = np.arange(1, 10)
--- a/scipy/interpolate/tests/test_interpolate.py
+++ b/scipy/interpolate/tests/test_interpolate.py
@@ -993,6 +993,42 @@ class TestAkima1DInterpolator:
         # Testing extrapoation to actual function.
         xp_assert_close(y_ext, ak_true(x_ext), atol=1e-15)
 
+    def test_large_dynamic_rannge(self):
+        # check a large step does not change non-overlapping subsplines
+        x = 1.*np.arange(1, 12)                    # grid of spline points
+        x_sample = np.linspace(1., 7., 21)         # grid of points to sample splines at
+        y1 = np.heaviside(x - 3, 0.5)              # small step function
+        y2 = y1 + 1.e9 * np.heaviside(x - 9, 0.5)  # increase of at least ~1e9
+
+        ak1 = Akima1DInterpolator(x, y1, method='akima')
+        y_eval1 = ak1(x_sample)
+
+        ak2 = Akima1DInterpolator(x, y2, method='akima')
+        y_eval2 = ak2(x_sample)
+
+        ak3 = Akima1DInterpolator(x, y1, method='makima')
+        y_eval3 = ak3(x_sample)
+
+        ak4 = Akima1DInterpolator(x, y2, method='makima')
+        y_eval4 = ak4(x_sample)
+
+        xp_assert_equal(y_eval1, y_eval2)
+        xp_assert_equal(y_eval3, y_eval4)
+
+    def test_no_overflow(self):
+        # check a large jump does not cause a float overflow
+        x = np.arange(1, 10)
+        y = 1.e160*np.heaviside(x-4, 0.5)
+
+        ak1 = Akima1DInterpolator(x, y, method='makima')
+        ak2 = Akima1DInterpolator(x, y, method='akima')
+
+        y_eval1 = ak1(x)
+        y_eval2 = ak2(x)
+
+        assert np.isfinite(y_eval1).all()
+        assert np.isfinite(y_eval2).all()
+
 
 @pytest.mark.parametrize("method", [Akima1DInterpolator, PchipInterpolator])
 def test_complex(method):


mcdigman:
Just to document the logic of the simplification/rearrangement of t I used for clarity:

t = (f1*m1+f2*m2)/(f1+f2)
  = m1 - (f1+f2)/(f1+f2)*m1 + (f1*m1+f2*m2)/(f1+f2)
  = m1 + ((-f1 - f2 + f1)*m1 + f2*m2)/(f1+f2)
  = m1 + (- f2*m1 + f2*m2)/(f1+f2)
  = m1 + (f1/(f1+f2))*(m2 - m1)

This way the division happens before the multiplication so that the multiplication involves smaller numbers, significantly mitigating the potential for overflow unless the variables were very close to overflow anyway. 

This way of writing it (with alpha = f1/(f1+f2)) is used in e.g. algorithm 13.1 in ["Akima and Renner Subsplines"](https://link.springer.com/chapter/10.1007/978-3-642-61074-5_13) from "Numerical Algorithms with C", describing the non-rounded corner akima splines method used in e.g. GSL.

j-bowhay:
Thanks for the contribution @mcdigman! Please be careful not to reference Numerical Recipes or GSL as they are not compatible with SciPy's license

ev-br:
Could you please confirm that you did not look at the GSL implementation @mcdigman ?

Like Jake said, https://github.com/scipy/scipy/pull/22888#issuecomment-2831942458, we'll need to do clean room reimplementation if you did.
Reading the docs or experimenting with input and output is OK, while reading the code itself is not, sadly.
To this end, I did not look at your implementation, and will not look at it until you confirm.

From reading your comments in the two issues:
- rearranging the formula to avoid a potential overflow is certainly fine;
- messing with the heuristics for the `max(f12)` threshold I'm not sure about TBH. In the end of the day, it is a heuristics, so there will always a corner case which breaks it. Making the threshold  user-controllable:  a new kwarg with the default of `1e9` instead of a hardcoded `1e9` is fine. 
- Changing the logic (threshold in the local neighborhood or globally) has backwards compatibility concerns. Given that we have a strictly local alternative in `PchipInterpolator`, I would only consider changing the logic in `Akima1DInterpolator` if there's an overwhelming evidence from the literature and alternative implementations (GSL, matlab, octave, Wolfram etc). Would be great to see a survey

mcdigman:
@ev-br @j-bowhay Oh no, I didn’t look at GSL’s actual code implementation or anything; I don’t know if they actually have it factored that way or not. The equation for t is just a different rearrangement from what I originally suggested in the GitHub issue that I decided was better at keeping the quantities small after playing around with the equation for a while. I didn’t mean to cite the book as a source for the equation; it is describing a totally different algorithmic modification to the original Akima splines from the ‘makima’ method, that instead uses left and right handed slopes instead of different weights like ‘makima’ does. The ‘makima’ modification is simpler and probably better in many/most applications (it lacks discontinuities). The change in factoring has nothing to do with that book; it never mentions the concept of weights at all. I can remove the mention of  it from my previous comment if it is unintentionally misleading about the origin of the equation. 

I can try to type it up more mathematically later if needed, but I think for the makima case I setting the cutoff to zero isn’t really a break in backwards compatibility so much as an improvement in the numerical behavior; I think the only case where they could disagree would be if the cutoff was misbehaving numerically. It’s a much smaller change for the makima case anyway, because it is better behaved to start with.

For the Akima case I totally agree it would be a bigger change/backwards compatibility break to set the cutoff to zero. I do not think it is quite as significant to change it from being calculated globally to locally. My concern with leaving the cutoff as calculated globally is that the failure case is totally silent, so users may be unaware that the potential for this particular misbehavior exists at all. I think it is totally plausible that someone could be interpolating say a modulated exponential function or broken power law or something and be unaware that extending the x range beyond what they strictly need could change its behavior everywhere (a factor of 10^9 is a plausible dynamic range for such a case; for example in gravitational waves we have a frequency as a function of time that can evolve all the way from the nanohertz to the kilohertz, so if I used this interpolator on that frequency function without some other mitigation, it could actually realistically trigger this issue). This could result in e.g. inconsistent behavior between different runs of the same scientific code with different boundary cutoffs (say hypothetically I evolved one run from just the nano hertz to the millihertz and was comparing to another run that was identical except that it continued all the way to the kilohertz).

scipy/interpolate/_cubic.py
```diff
@@ -535,16 +535,29 @@ def __init__(self, x, y, axis=0, *, method: Literal["akima", "makima"]="akima",
...
             f12 = f1 + f2
+
+            # mask to handle m2=m3 case exactly
+            ind = np.nonzero((dm[1:-1] == 0.))
+            x_ind, y_ind = ind[0], ind[1:]
+            t[ind] = m[(x_ind + 1,) + y_ind]
+
             # These are the mask of where the slope at breakpoint is defined:
-            ind = np.nonzero(f12 > 1e-9 * np.max(f12, initial=-np.inf))
+            ind = np.nonzero((dm[1:-1] != 0.) & (f12 > break_mult * dm[1:-1]))
```
ev-br, line 555: This is potentially a backwards compat break IIUC. Some datasets will silently produce different results.

ev-br (commented):
I think I stand by my initial assessment: rearranging the formular to help numerical stability is totally fine; changing the algorithm in a backwards incompatible way is risky. Your argument in https://github.com/scipy/scipy/pull/22888#issuecomment-2832530123 make total sense --- if we were designing the function from scratch.
At this stage of the lifetime of this function, it feels that what you discuss is potential deficiencies of the algorithm. They are totally real of course. What I'm saying is I'm not convinced that fixing these potential deficiencies justifies a potential disruption of existing uses.
Now that you mention your use cases --- would Pchip work for you instead?

To summarize, I'd rather keep the `1e-9 * max(f12)` heauristics, but if, say, @j-bowhay thinks we should change it, I won't block it.

ev-br:
> I guess if you think this behavior should be kept in one or both 'method' cases for maximum backwards compatibility, then at least some kind of warning about this behavior could be added to the docs? Perhaps in the context of the description of a new keyword argument to set a custom threshold?

Absolutely, that'd be great!

Depending on how ambitious you feel, you may also consider adding a DIY recipe using `CubicHermiteSpline` to the tutorial (https://docs.scipy.org/doc/scipy/tutorial/interpolate.html).

The `jacobi` CI failure under free-threading is indeed unrelated and can be ignored here.

j-bowhay:
I think I would prefer to split this pr up. The rearrangement for numerical stability is great and could be merged straight away. The other changes might need some further discussion but shouldn't hold up the numerical stability improvements.

mcdigman:
@j-bowhay That makes sense; should I make this request the one for the identical improvement and then also make another, or vice versa? I haven't split a pull request in two before.

@ev-br There is one specific other case to decide how to handle, which is whether to keep my special/separate handling of the m2=m3 exactly case, where I have replaced the fill value with the analytic answer. 

As an existing comment in the code mentions, 

> if m1 == m2 != m3 == m4, the slope at the breakpoint is not 
> defined. This is the fill value:

the fill value is not really meant to be used in the m2=m3 case, because the slope _is_ defined.

This _shouldn't_ really change anything because it is an analytic simplification. However, it is possible; after some playing around, the easiest way I found to get skipping this check to trigger misbehavior is to pick a case that would _also_ trigger the high dynamic range behavior. However, I suspect it would be possible to find a case that trigger this misbehavior independently though (it might just be a less visibly obvious difference?). 

The commit I just added changes none of the functionality I previously proposed in `_cubic`, but adds a variant of the high dynamic range test case that would specifically pass or fail based on whether the m2=m3 handling was included in the old code. I attached a plot comparing the behavior it triggers; I think it is visually rougher than the other misbehavior, and really just a bug that should be fixed. 

This issue is specific to method='akima', method='makima' is almost totally unaffected. 

akima_m2=m2_case

mcdigman:
I haven't yet found a case where the m2=m3 handling changes anything by more than ~floating point epsilon unless you also trigger the case where the max(f12) handling misbehaves. However I attached a plot of another case that more clearly illustrates how the behavior is worse without the m2=m3 handling in a high dynamic range case (i.e it eventually climbs to 10^10 somewhere off screen). In this case, without the m2=m3 handling it struggles to reproduce _straight lines_ even away from the discontinuity (the behavior also differs in the discontinuity, but it may be less clear which curve should be the expected behavior there).
dynamic_range_hill_fail

mcdigman:
Oops I can fix the tests to work in 32-bit and also the linting failure. I can't make any sense of the other failures. 

I can wait to commit that until it is exactly in the form you want it to be merged for the first pull request, so let me know if you think it is ok to include the m2=m3 handling or not.

ev-br:
I'd suggest the following process:
- A pure maintenance PR: make a separate PR with just the identical rearrangement for numeric stability. Include a test case to show the improved stability if you have one.
- merge the maintenance PR -- this can be done today, so that it is included into the 1.16 release
- rebase on main and work on your further improvements in the 1.17 cycle.

mcdigman:
Sorry I missed this before the deadline. I made a new pull request rolled back to just the strictly identical changes; I left `break_mult=1e.-9` factored out as a variable (with the same hard-coded value) with the view of possibly making it a keyword argument in the future. If it is made a keyword argument, we would still need to decide whether/how to include the explicit m2=m3 case handling.

ev-br:
This now needs a rebase after your other PR being merged. Also some of CI failures are related (it's OK to ignore connection errors in `datasets`, those are likely a fluke).

lucascolley:
would you like to return to this @mcdigman ?

mcdigman:
Ah I forgot there was more to do after the pull request merged. So, to remind myself and the thread the steps were  

- [ ]  Make the multiplier a keyword argument
- [ ] Handle m2=m3 edge case somehow. Ideas for this:
1.) Just use the analytic answer always, which more accurately reflects the intended behavior but is technically a break from backwards compatibility in a few edge cases 
2.) Use the analytic answer only if the user specifies a value for the new keyword argument, which completely prevents any break from backwards compatibility though it doesn't fix that sub-issue unless the user knows about it
3.) Use the analytic answer only if the threshold is specified and is set to exactly zero, indicating that they want all parts of this solution implemented
4.) Have a second new keyword argument to select the m2=m3 handling (IMO this would add unnecessary complexity and confusion because AFAIK there is no reason you would want the current handling other than to preserve near-identical backward compatibility)

ev-br:
I'd prefer to keep it backwards compatible, and as simple as possible otherwise.

Mohsalam:
Hi! I’m interested in working on this issue and can start investigating right away.

PR #22891, opened 2025-04-25
DOC: Document allowed NumPy / Python versions

### LINKED ISSUE #22881: DOC: Update minimum NumPy and Python in toolchain roadmap

The current [toolchain roadmap](https://docs.scipy.org/doc/scipy/dev/toolchain.html) documentation does not document:

* Minimum NumPy for SciPy versions 1.14, 1.15, and 1.16.
* Minimum Python supported in 2025

Proposed values for supported NumPy versions:

* 1.14: >=1.23.5, <2.3.0
* 1.15: >=1.23.5, <2.3.0
* 1.16: >=1.25.2, <2.3.0

I got these values from this file https://github.com/scipy/scipy/blob/main/.github/workflows/linux.yml at various tags

Proposed values for supported Python versions:

* 2025: 3.11+

The logic of this is that Python 3.10 was released in Oct 2022. 42 months after this is Apr 2025. This is assuming we don't release 1.16 in the next week : ).

_No response_

nickodell:
Document which versions of NumPy / Python are supported on Toolchain Roadmap page.

#### Reference issue

Closes #22881 

Documents the compatible NumPy / Python versions.


doc/source/dev/toolchain.rst
```diff
@@ -87,6 +88,9 @@ needs to be written using what is common in all of those 4 `NumPy releases`_.
     =================  ========================    =======================
      SciPy version      Python versions             NumPy versions
     =================  ========================    =======================
+     1.16               >=3.11, <3.14               >=1.25.2, <2.6.0
+     1.15               >=3.10, <3.13               >=1.23.5, <2.5.0
```
tylerjereddy, line 92: `1.15.0` included `3.13.0` support from the start right? https://github.com/scipy/scipy/releases/tag/v1.15.0

doc/source/dev/toolchain.rst
```diff
@@ -87,6 +88,9 @@ needs to be written using what is common in all of those 4 `NumPy releases`_.
...
+     1.14               >=3.10, <3.13               >=1.23.5, <2.3.0
```
tylerjereddy, line 93: I can't quite tell from the comment above if this is supposed to mean the `1.14.x` series or `1.14.0` proper:

> This table does not distinguish SciPy patch versions (e.g. when a new Python
    version is released, SciPy will generally issue a compatible patch version).

If it means the series, then as we just discussed, `3.13` support was added in `1.14.1`: https://github.com/scipy/scipy/releases/tag/v1.14.1

tylerjereddy (commented):
Thanks for taking the lead on documenting this--I added some small comments

doc/source/dev/toolchain.rst
rgommers: I think we wave our hands just slightly here and mean the latest bug fix release for each feature release. So I think it's should be `<3.14` here.

And I think the scipy 1.16 release will include Python 3.14 support. The only thing missing is that 1.16.0 will not have `cp314` wheels yet - but we'll add them to 1.16.x as soon as we can after CPython 3.14.0rc1.

nickodell: >I think we wave our hands just slightly here and mean the latest bug fix release for each feature release. So I think it's should be <3.14 here.

This seems to be the interpretation that the rest of the table uses. For example, in the 1.11 series, it documents support for `>=3.9, <3.13`, but support for 3.12 was not added until 1.11.2.

I wrote a program to look at all released wheel files on PyPI.org and see what Python versions are available for each scipy version.

```

1.15.2 >=3.10, <3.14
1.15.1 >=3.10, <3.14
1.15.0 >=3.10, <3.14

1.14.1 >=3.10, <3.14
1.14.0 >=3.10, <3.13

1.13.1 >=3.9, <3.13
1.13.0 >=3.9, <3.13

1.12.0 >=3.9, <3.13

1.11.4 >=3.9, <3.13
1.11.3 >=3.9, <3.13
1.11.2 >=3.9, <3.13
1.11.1 >=3.9, <3.12
1.11.0 >=3.9, <3.12

1.10.1 >=3.8, <3.12
1.10.0 >=3.8, <3.12

1.9.3 >=3.8, <3.12
1.9.2 >=3.8, <3.12
1.9.1 >=3.8, <3.11
1.9.0 >=3.8, <3.11

1.8.1 >=3.8, <3.11
1.8.0 >=3.8, <3.11

1.7.3 >=3.7, <3.11
1.7.2 >=3.7, <3.11
1.7.1 >=3.7, <3.10
1.7.0 >=3.7, <3.10

1.6.3 >=3.7, <3.10
1.6.2 >=3.7, <3.10
1.6.1 >=3.7, <3.10
1.6.0 >=3.7, <3.10

1.5.4 >=3.6, <3.10
```

nickodell: > If it means the series, then as we just discussed, `3.13` support was added in `1.14.1`: https://github.com/scipy/scipy/releases/tag/v1.14.1

Fixed in 38060c4520bff73d6720c5103cba47b1a0793ff9.

doc/source/dev/toolchain.rst
nickodell: Yes, you're right. Fixed in commit 38060c4520bff73d6720c5103cba47b1a0793ff9.

doc/source/dev/toolchain.rst
nickodell: > I can't quite tell from the comment above if this is supposed to mean the `1.14.x` series or `1.14.0` proper:
> 
> > This table does not distinguish SciPy patch versions (e.g. when a new Python
> > version is released, SciPy will generally issue a compatible patch version).

I've attempted to clarify this in commit ef4e44037 by changing the documentation to match @rgommers suggested interpretation. I've tried to rewrite this to make the bottom line clearer for users.

nickodell:
In commit f54f60f, I removed the documentation for 1.16, as it seems like there are several factors like NumPy version and Python version that are not yet 100% settled. I want to avoid adding incorrect documentation, as I think that's harder to notice and fix than missing documentation.

rgommers (approved):
LGTM now. The added textual clarification makes more clear exactly how to interpret the table, and it keeps the table more concise then trying to start tracking bug fix versions.

Thanks @nickodell, and thanks @tylerjereddy for the review.

Issue #17751, opened 2023-01-08
Use Meson subprojects to allow a system version of Boost, Qhull, & co


rgommers:
This can be done to make distro packagers happy, without introducing more issues around having extra build time dependencies or different versions of dependencies. https://mesonbuild.com/Subprojects.html#toggling-between-system-libraries-and-embedded-sources explains most of how that works. The one thing it does not show is how to restrict `dependency(...)` to a single version, but that is possible. That way, if the system has the right version of a library available already, it can be used. And if it's not found, we use our vendored copy (in the SciPy source tree or in a git submodule).

This came up several times:
- recently for Qhull, which conda-forge builds separately (see [this mailing list thread](https://mail.python.org/archives/list/scipy-dev@python.org/thread/I535LGRTWUXJIP2VM7VQGF3HELHQJ4B2/)),
- for Boost in gh-15260 (Boost is less interesting because it's header-only, but still some people seem to care)
- the Fortran-lang folks have asked about relying on their Minpack in the future (could be vendored still of course)
- there's other libraries that I can't remember questions about, but may be present in distros (ARPACK-ng, HiGHS, etc.)

This is not high-prio, but it would be nice to implement.

rgommers:
xref https://github.com/mesonbuild/meson-python/discussions/410

rgommers:
This blog post is a good example of using WrapDB for this kind of thing: https://nibblestew.blogspot.com/2023/05/advanced-dependency-management-and.html. It's even simpler than keeping a git submodule in `subprojects/`, but of course at the cost of needing network access if the dependency is not installed (seems fine in principle).

This could also be useful to start enforcing a result enough OpenBLAS version. Once there's a Meson OpenBLAS build and it's present in WrapDB (which is in the plan for next year), that can be the fallback when we specify `dependency('openblas', version='>=0.3.21')`. Right now we have no such check, and we do sometimes get bug reports from users on distros with several-years-old versions of OpenBLAS.

doronbehar:
Are you planning to also support system wide installations of these dependencies?

rgommers:
Yes, because distro packagers like that. The way it works now is that we vendor lots of stuff, and a small part of it gets unvendored in an ad-hoc way in some distros. The way we want it to work is to declare dependencies like so:
```
dependency('openblas', version: '>=0.3.21')
dependency('boost', modules : ['math'], version: '1.82')
```
[...]

Issue #22887, opened 2025-04-25
DEV: follow-up tasks after merge of `spin` support


rgommers:
gh-21674 added support for [spin](https://github.com/scientific-python/spin), which can now be used instead of `python dev.py` for all development tasks (e.g., `python dev.py test` becomes `spin test`). See `spin --help` and `spin <task> --help` for more details.

This issue is to track follow-up items and allow people to report any hiccups they may encounter:

- [ ] Update the documentation (contributor guide etc.) - search for `dev.py` and replace with `spin`
- [ ] One final PR once we're happy with `spin`, at least a few weeks out):
    - [ ] Remove `dev.py`
    - [ ] Update the few CI jobs that are still using `dev.py`
    - [ ] Update `dev` dependencies in `pyproject.toml` and sync to requirements files with `tools/generate_requirements.py`
    - [ ] Update `environment.yml`
    - [ ] Check throughout the repo that `click`, `rich-click`, `doit` and `pydevtool` are no longer used

stefanv:
`spin` should function correctly with editable installs; Ralf or Gagandeep would know if there are any SciPy-specific constraints.

rgommers:
> `spin` should function correctly with editable installs;

I'm not sure that that helps us in its current state. My understanding is that there is `spin install` which can be used to perform an editable install, but then running `spin test` will still trigger `build` under the hood right? So it's still not possible to use `spin` if we default to `build`.

Hence, as of today, the status in SciPy is the same as with `dev.py`: you can either use `spin` or you can use editable installs (`pip install -e . --no-build-isolation` and `pytest ...`), but not mix them.

If we want to allow `spin` itself to do an editable install, I think there's some logic in `spin` needed, plus a UX knob like `SPIN_PREFER_EDITABLE=true` env var), which tells `spin` that `test` and other commands shouldn't call `build`.

Does that make sense @stefanv or am I missing something?

lucascolley:
> My understanding is that there is `spin install` which can be used to perform an editable install

I think we would need to add something to https://github.com/scipy/scipy/blob/a82a5782fbf1b348a07ca3e7999caf5ed52aa102/pyproject.toml#L173 for that:

```console
scipy on 🎋 main [?] is 📦 v1.16.0.dev0 via 🐍 
❯ pixi run spin install
Usage: spin [OPTIONS] COMMAND [ARGS]...
Try 'spin --help' for help.

Error: No such command 'install'.
```

rgommers:
[...]

PR #22893, opened 2025-04-26
MAINT: vendor qhull as subproject and add `-Duse-system-libraries`

lucascolley:
#### Reference issue
Towards gh-17751

TODO:
- [x] test with qhull from conda-forge in CI
- [x] use a script for vendoring?


lucascolley (commented):
this gets us most of the way. However, there are two files which explicitly include directories from qhull:

https://github.com/scipy/scipy/blob/a82a5782fbf1b348a07ca3e7999caf5ed52aa102/scipy/spatial/qhull_misc.h#L12 and https://github.com/scipy/scipy/blob/a82a5782fbf1b348a07ca3e7999caf5ed52aa102/scipy/spatial/_qhull.pyx#L57

Advice on how to proceed appreciated.

rgommers:
I don't think this will be useful unless we first upgrade to QHull 2020.2, which is the release shipped by the vast majority of distros: https://repology.org/project/qhull/versions. Maybe try that in a separate PR? With a bit of luck, it'll "just work".

For the headers, `dependency('qhull_r')` should return a dependency object that contains the right include dir. E.g., for `qhull` from conda-forge: 
```
$ pkg-config --libs qhull_r
-L/home/rgommers/mambaforge/envs/qhull/lib -lqhull_r
```
That will certainly work for `qhull_misc.c|h`. Not sure if it works for the Cython code, I can't remember if Meson dependencies are handled correctly with Cython code (I hope so though, give it a try).

subprojects/qhull_r/meson.build
```diff
@@ -0,0 +1,49 @@
...
+    # private symbols don't get re-exported.
+    gnu_symbol_visibility: 'inlineshidden',
+    pic: true,
+    install: false,
+)
+
+qhull_r_dep = declare_dependency(
+    link_with: libqhull_r,
+    include_directories: 'qhull/src/libqhull_r',
+)
```
lucascolley, line 49: this may not be exactly what we want, I took inspiration from what we use for highs

lucascolley:
cc @gillins, in case you are still interested in allowing external libqhull with SciPy 2 years on from https://mail.python.org/archives/list/scipy-dev@python.org/thread/I535LGRTWUXJIP2VM7VQGF3HELHQJ4B2/. This gets us most of the way

subprojects/qhull_r/meson.build
```diff
@@ -0,0 +1,49 @@
...
```
lucascolley, line 49: this may not be exactly what we want, I took inspiration from what we use for highs

lucascolley (commented):
@rgommers should we try using a script to vendor relevant parts from qhull rather than using a git submodule?

lucascolley:
> Not sure if it works for the Cython code, I can't remember if Meson dependencies are handled correctly with Cython code (I hope so though, give it a try).

It looks like this is working for the meson dependency I have created from the sources, but not for the dependency from  conda-forge `qhull`.

gillins:
Thanks @lucascolley !

musvaage:
@lucascolley 

The last commit to qhull appears to be from Jul 23, 2024.

https://github.com/qhull/qhull/pull/148 {Closed)

A ping to @cbbarber who has write authority over there was not responded to.

I wonder how actively qhull is factually being maintained though I must admit my pull was non-essential fixes.

rgommers:
> @rgommers should we try using a script to vendor relevant parts from qhull rather than using a git submodule?

Yes, that sounds like a good idea. Unlike some of the more active subprojects, we are unlikely to want to develop on QHull itself to send fixes or pull in updates regularly, so avoiding a git submodule is nice.

> I wonder how actively qhull is factually being maintained

Well, the last release is from 2020, so not actively at all I'd say. At least it has an upstream.

rgommers:
> It looks like this is working for the meson dependency I have created from the sources, but not for the dependency from conda-forge `qhull`.

I think the commit I just pushed fixes it. If not completely, it'll certainly be in the right direction.

Note that I didn't review completely yet, just pushed a quick fix. Also, I think we may want to change the pattern to opt-in, because QHull packaging seems fairly broken and even versioning isn't consistent. So defaulting to the subproject and adding a `--use-system-libraries` flag is probably healthy to avoid problems.

scipy/spatial/qhull_src/README.txt
lucascolley, line 1: the README seemed superfluous with both COPYING and ANNOUNCE kept, but obviously I'm fine either way

scipy/spatial/qhull_misc_README.txt
```diff
@@ -1,10 +1,8 @@
...
-version source code.
-
-The file scipy/spatial/qhull_misc.c additionally contains a function
+The file scipy/spatial/qhull_misc.c contains a function
 "qh_new_qhull_scipy" derived from Qhull sources, via the following
 patch:
 
+TODO: regenerate patch
 --- a/scipy/spatial/qhull_src/src/user_r.c
 +++ b/scipy/spatial/qhull_src/src/user_r.c
```
lucascolley, line 7: TODO

subprojects/qhull_r/meson.build
```diff
@@ -0,0 +1,44 @@
...
+
+libqhull_r = static_library(
+    'libqhull_r',
+    [libqhull_r_sources],
+    c_args: cc.get_supported_arguments('-Wno-unused-but-set-variable'),
+    include_directories: 'libqhull_r',
+    # Ensure that if we link a static library into a shared library,
+    # private symbols don't get re-exported.
+    gnu_symbol_visibility: 'inlineshidden',
+    pic: true,
```
lucascolley, line 37: are these lines relevant?

subprojects/qhull_r/meson.build
```diff
@@ -0,0 +1,44 @@
...
+  'libqhull_r/user_r.c',
+  'libqhull_r/usermem_r.c',
+  'libqhull_r/userprintf_r.c',
+  'libqhull_r/userprintf_rbox_r.c'
+]
+
+libqhull_r = static_library(
+    'libqhull_r',
+    [libqhull_r_sources],
+    c_args: cc.get_supported_arguments('-Wno-unused-but-set-variable'),
```
lucascolley, line 32: should we be passing any other `c_args` defined elsewhere? Or are any global ones we need automatically injected?

scipy/spatial/qhull_src/README.txt
rgommers: Sure, fine either way with me

subprojects/qhull_r/meson.build
rgommers: `pic: true` probably not (remove it I'd say), symbol visibility yes.

subprojects/qhull_r/meson.build
rgommers: I think the ones from `add_project_arguments` in the top-level `meson.build` are used (please check), which means that this line isn't needed either.

lucascolley: ah, no, that doesn't seem to be the case. This line was necessary to get the build to pass.

subprojects/qhull_r/meson.build
lucascolley:
```suggestion
    # Ensure that if we link a static library into a shared library,
    # private symbols don't get re-exported.
    gnu_symbol_visibility: 'inlineshidden',
```

doc/source/dev/core-dev/vendored-code.rst.inc
```diff
@@ -3,29 +3,44 @@
...
 - DIRECT_, at ``scipy/optimize/_direct``
 - ARPACK_, at ``scipy/sparse/linalg/_eigen/arpack/ARPACK``
 - SuperLU_, at ``scipy/sparse/linalg/_dsolve/SuperLU``
-- QHull_, at ``scipy/spatial/qhull_src``
 - trlib_, at ``scipy/optimize/_trlib``
 - UNU.RAN_, at ``scipy/stats/_unuran``
-- PRIMA_, at ``scipy/_lib/pyprima``
+- Cython/Tempita_, at ``scipy/_build_utils/tempita``
+- fast-matrix-market_, at ``scipy/io/_fast_matrix_market``
+- numpydoc/docscrape_, at ``scipy/_lib/_docscrape.py``
```
lucascolley, line 33: this syntax needs to be fixed

```
source/dev/core-dev/vendored-code.rst.inc:31: ERROR: Unknown target name: "tempita".
source/dev/core-dev/vendored-code.rst.inc:33: ERROR: Unknown target name: "docscrape".
```

subprojects/qhull_r/meson.build
rgommers: Confirmed. That is a little annoying, the Meson docs could have been more clear here. It looks like we should switch back to using `add_global_arguments` then in `meson.build` for some of the calls (e.g., the Intel compiler flag handling).

I had expected `add_project_arguments` to avoid propagating upwards and sideways, but to still propagate downwards into subprojects of the main project.

rgommers: I reviewed them all, the rest looks fine as `add_project_*`, just the Intel ones should be global.

doc/source/dev/core-dev/vendored-code.rst.inc
```diff
@@ -3,29 +3,46 @@
...
-This is mainly for historical reasons, and it is possible that some of these parts
-will see patches contributed upstream and become git submodules in the future.
+Other parts are not vendored as git submodules or under the ``subprojects`` directory,
+despite having a maintained upstream. This is usually either because:
+
+#. a subset of the upstream repo is vendored with a script.
+(It is possible that these parts will be moved under ``subprojects`` in the future.)
+#. Code has been copied into SciPy and modified since.
+(It is possible that some of these parts will see patches contributed upstream and
+become git submodules or be moved under ``subprojects`` in the future.)
```
lucascolley, line 16:
```suggestion
1. a subset of the upstream repo is vendored with a script.
    (It is possible that these parts will be moved under ``subprojects`` in the future.)

2. Code has been copied into SciPy and modified since.
    (It is possible that some of these parts will see patches contributed upstream and
become git submodules or be moved under ``subprojects`` in the future.)
```

lucascolley:
> I think we may want to change the pattern to opt-in ... defaulting to the subproject and adding a --use-system-libraries flag is probably healthy to avoid problems.

done, let's see if CI is happy

lucascolley:
I would suggest that we save documenting the `--use-system-libraries` flag until we switch the docs over to `spin` (https://github.com/scipy/scipy/issues/22887) and/or we get more subprojects converted over to this pattern

rgommers (commented):
Thanks for adding in `use-system-libraries`! The main comment I have is that it should be an `array` rather than `boolean` option, with values `none` (default), `auto`, `boost.math`, `highs`, `qhull` (etc). That way the packager can have much more control over it. 

> I would suggest that we save documenting the --use-system-libraries flag until we switch the docs over to spin (https://github.com/scipy/scipy/issues/22887) and/or we get more subprojects converted over to this pattern

Agreed, not too worried about `spin`, that can be done at any time in the future.

meson.build
```diff
@@ -163,11 +163,41 @@ endif
...
-  fallback : ['boost_math', 'boost_math_dep'],
-)
+use_system_libraries = get_option('use-system-libraries')
+auto_system_libraries = false
+
+if use_system_libraries.contains('none')
+  use_system_libraries = ['none']
+elif use_system_libraries.contains('auto')
+  auto_system_libraries = true 
+endif
```
lucascolley, line 173: The logic is as follows:
1. If we contain `'none'`, use no system libraries, regardless of the rest of the array
2. Otherwise, use all libraries set to 'auto' (currently all of them), and potentially others specified in the array
3. Otherwise, use libraries specified in the array

lucascolley:
CI failure is unrelated (calling ARPACK)

meson.build
rgommers: Very close I think. https://packaging.python.org/en/latest/discussions/downstream-packaging/#id5 suggests one change that I also had in mind, erroring out if `-Duse-system-libraries=libname` is used and `libname` isn't given. So it's a tri-state thing:
- default to vendored
- "auto" to try system with fallback to vendored
- named lib(s) to force use of the system libraries

We should probably also add `'all'` as the opposite of `'none'`.

WDYT?

rgommers: Last suggestion: if the value is `auto,qhull`, I think that should mean "force `qhull`, use `auto` for the rest

lucascolley: > erroring out if -Duse-system-libraries=libname is used and libname isn't given.

sounds good

> if the value is auto,qhull, I think that should mean "force qhull, use auto for the rest

:+1: yes the current implementation does that, I worded (2) above ambiguously

meson.build
```diff
@@ -163,11 +163,41 @@ endif
...
+    'boost_math',
+    version : '1.88.0',
+  )
+  boost_math_dep = boost_math.get_variable('boost_math_dep')
+endif
 
+if auto_system_libraries or use_system_libraries.contains('qhull')
+  qhull_r_dep = dependency(
+    'qhull_r',
+    version : '8.0.2',
```
rgommers, line 192: This should be reliable I think, due to 
https://github.com/qhull/qhull/blob/c7bee59d068a69f427b1273e71cdc5bc455a5bdd/CMakeLists.txt#L77-L78

The _package_ version can be either 8.0.2 or 2020.2 depending on the distro. 

Just a note for future reference; nothing to change here.

meson.build
```diff
@@ -163,11 +163,41 @@ endif
...
+    'boost',
+    version : '1.88.0',
+    fallback : ['boost_math', 'boost_math_dep'],
+  )
+else
+  boost_math = subproject(
+    'boost_math',
+    version : '1.88.0',
+  )
+  boost_math_dep = boost_math.get_variable('boost_math_dep')
```
rgommers, line 186: These `dependency` and `subproject` calls may be easier to read if they're on a single line, especially if it becomes `if-elif-else`. 

If you prefer to leave it like this, also fine with me.

meson.build
rgommers: > 👍 yes the current implementation does that, I worded (2) above ambiguously

It doesn't yet, but it will after the "error out" cases are added.

rgommers (commented):
I re-reviewed everything, looks great. I think we can (and should) get this in before the 1.16.x branch is created.

dschmitz89:
I just noticed that after the recent qhull update the scipy specific readme was not updated: https://github.com/scipy/scipy/blob/main/scipy/spatial/qhull_src/README.scipy

Feel free to ignore if this file anyway becomes obsolete with the changes proposed here.

lucascolley:
https://github.com/lucascolley/scipy/blob/qhull/scipy/spatial/qhull_misc_README.txt

meson.build
rgommers: Yep, LGTM now! resolving.

meson.build
```diff
@@ -163,11 +163,33 @@ endif
...
+  auto_system_libraries = true 
+endif
+
+if auto_system_libraries or use_system_libraries.contains('boost.math')
+  boost_math_dep = dependency(
+    'boost', version : '1.88.0',
+    fallback : ['boost_math', 'boost_math_dep']
+  )
+else
+  boost_math = subproject('boost_math',version : '1.88.0')
```
lucascolley, line 181: minor typo

```suggestion
  boost_math = subproject('boost_math', version : '1.88.0')
```

meson.build
rgommers: Not quite what I meant. Let me try and implement it in the next couple of hours.

lucascolley: ohh, I finally see what you mean

meson.build
```diff
@@ -163,11 +163,40 @@ endif
...
+  boost_math_dep = dependency('boost', version : '1.88.0')
+elif auto_system_libraries
+  boost_math_dep = dependency(
+    'boost', version : '1.88.0',
+    fallback : ['boost_math', 'boost_math_dep']
+  )
+else
+  boost_math = subproject('boost_math', version : '1.88.0')
+  boost_math_dep = boost_math.get_variable('boost_math_dep')
+endif
```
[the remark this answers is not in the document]
rgommers: Yes, looks like it!

rgommers (approved):
Everything looks happy, in it goes. Thanks a lot Lucas, nice work!

meson.build
```diff
@@ -163,11 +163,40 @@ endif
...
+auto_system_libraries = false
+
+if use_system_libraries.contains('none')
+  use_system_libraries = ['none']
+elif use_system_libraries.contains('all')
+  all_system_libraries = true
+elif use_system_libraries.contains('auto')
+  auto_system_libraries = true 
+endif
+
```
eli-schwartz, line 177: Just saw this PR. It feels very complex and fragile. :(

Meson has a builtin:
- `-Dwrap_mode=` 
  - forcefallback ("use system libraries" == none)
  - default ("use system libraries" == auto)
  - nofallback ("use system libraries" == all)
- `-Dforce_fallback_for=` ("use system libraries" == 'boost.math', 'qhull')

meson.build
```diff
@@ -163,11 +163,40 @@ endif
...
+if all_system_libraries or use_system_libraries.contains('boost.math')
+  boost_math_dep = dependency('boost', version : '1.88.0')
+elif auto_system_libraries
+  boost_math_dep = dependency(
+    'boost', version : '1.88.0',
+    fallback : ['boost_math', 'boost_math_dep']
+  )
+else
+  boost_math = subproject('boost_math', version : '1.88.0')
+  boost_math_dep = boost_math.get_variable('boost_math_dep')
```
eli-schwartz, line 187: All of this would be handled using
```meson
   boost_math_dep = dependency(
    'boost', version : '1.88.0',
    fallback : ['boost_math', 'boost_math_dep']
  )
```

meson.build
lucascolley: feel free to send a PR to simplify!

rgommers: @eli-schwartz I looked at that, but I don't think there's a way to implement the exact same logic with the built-in wrap modes? I really want the default to be `none`, not `auto`, because it's going to be more robust at both build and install time. Distro packagers can opt into system libs, but end users building from source (a concern specific to PyPI) should not since there's nothing to gain there except new failure modes.

In addition, we're going to have cases where the default mode is dependency-specific. E.g., if we add a wrap file for OpenBLAS, that should have either the `auto` or `nofallback` behavior by default.

eli-schwartz: > @eli-schwartz I looked at that, but I don't think there's a way to implement the exact same logic with the built-in wrap modes? I really want the default to be `none`, not `auto`, because it's going to be more robust at both build and install time. Distro packagers can opt into system libs, but end users building from source (a concern specific to PyPI/pip/wheels) should not since there's nothing to gain there except new failure modes.

You would simply set:
```meson
project(...,
    default_options: ['wrap_mode=forcefallback'],
)
```

?

> In addition, we're going to have cases where the default mode is dependency-specific. E.g., if we add a wrap file for OpenBLAS, that should have either the `auto` or `nofallback` behavior by default.

Or,

```meson
project(...,
    default_options: ['force_fallback_for=boost,qhull'],
)
```

I'm also quite happy to discuss tweaking the allowed values in meson itself, if you have ideas for enhancing the granularity...

PR #22895, opened 2025-04-27
MAINT: signal: correct the `get_window` delegator

ev-br:
`scipy.signal.get_window` has an `xp=` argument, but its delegator `get_window_signature` was always returning `np` for the namespace, so correct that.


lucascolley (approved):
thanks Evgeni!

PR #22896, opened 2025-04-27
ENH: signal: `tf2zpk` et al  Array API

ev-br:
#### Reference issue

on top of https://github.com/scipy/scipy/pull/22886

Convert to array API:

- [x] `zpk2tf`
- [x] `tf2zpk`
- [x] `tf2sos`
- [x] `sos2tf`
- [x] `sos2zpk`
- [x] `zpk2sos`

#### Additional information

Am taking a shortcut with `zpk2sos` : it currently goes via numpy. Making the algorithm properly backend agnostic is a bit annoying, and can be left for a follow-up.

### DIFF
--- a/scipy/signal/_filter_design.py
+++ b/scipy/signal/_filter_design.py
@@ -1438,8 +1438,8 @@ def sos2zpk(sos):
     k = 1.
     for section in range(n_sections):
         zpk = tf2zpk(sos[section, :3], sos[section, 3:])
-        z[2*section:2*section + zpk[0].shape[0]] = zpk[0]
-        p[2*section:2*section + zpk[1].shape[0]] = zpk[1]
+        z = xpx.at(z, slice(2*section, 2*section + zpk[0].shape[0])).set(zpk[0])
+        p = xpx.at(p, slice(2*section, 2*section + zpk[1].shape[0])).set(zpk[1])
         k *= zpk[2]
     return z, p, k
 
--- a/scipy/signal/_filter_design.py
+++ b/scipy/signal/_filter_design.py
@@ -1652,14 +1652,7 @@ def zpk2sos(z, p, k, pairing=None, *, analog=False):
 
     # convert to numpy, convert back on exit   XXX
     z, p = map(np.asarray, (z, p))
-
-    k = xp.asarray(k)
-    if xp.isdtype(k.dtype, 'complex floating'):
-        if xp.imag(k) != 0:
-            raise ValueError('k must be real')
-        k = float(xp.real(k))
-    else:
-        k = float(k)
+    k = np.asarray(k)
 
     if pairing is None:
         pairing = 'minimal' if analog else 'nearest'
@@ -1674,9 +1667,9 @@ def zpk2sos(z, p, k, pairing=None, *, analog=False):
 
     if len(z) == len(p) == 0:
         if not analog:
-            return xp.asarray([[k, 0., 0., 1., 0., 0.]])
+            return xp.asarray(np.asarray([[k, 0., 0., 1., 0., 0.]]))
         else:
-            return xp.asarray([[0., 0., k, 0., 0., 1.]])
+            return xp.asarray(np.asarray([[0., 0., k, 0., 0., 1.]]))
 
     if pairing != 'minimal':
         # ensure we have the same number of poles and zeros, and make copies
--- a/scipy/signal/_filter_design.py
+++ b/scipy/signal/_filter_design.py
@@ -4,7 +4,7 @@ import operator
 import warnings
 
 import numpy as np
-from numpy import (atleast_1d, poly, roots, asarray,
+from numpy import (atleast_1d, asarray,
                    pi, absolute, sqrt, tan, log10,
                    arcsinh, sin, exp, cosh, arccosh, ceil, conjugate,
                    sinh, concatenate, prod, array)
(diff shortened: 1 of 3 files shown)


scipy/signal/_polyutils.py
```diff
@@ -0,0 +1,171 @@
...
+# https://github.com/numpy/numpy/blob/v2.2.0/numpy/lib/_polynomial_impl.py#L34-L157
+def poly(seq_of_zeros, *, xp):
+    # Only reproduce the 1D variant of np.poly
+    seq_of_zeros = xp.asarray(seq_of_zeros)
+    seq_of_zeros = xpx.atleast_nd(seq_of_zeros, ndim=1, xp=xp)
+
+    if seq_of_zeros.shape[0] == 0:
+        return 1.0
+
+    # prefer np.convolve etc, if available
```
ev-br, line 106: `np.convolve` might be an option to consider for `array-api-extra`. 

We cannot always default to `scipy.signal.convolve` or `scipy.ndimage.convolve` since those are CPU only.
CuPy and JAX have it available, and a there's a pytorch equivalent:
https://github.com/pytorch/pytorch/blob/main/torch/_numpy/_funcs_impl.py#L612

There is no usable pure-python, array-agnostic implementation though.

scipy/signal/_filter_design.py
```diff
@@ -1214,13 +1214,21 @@ def tf2zpk(b, a):
...
     """
+    xp = array_namespace(b, a)
     b, a = normalize(b, a)
-    b = (b + 0.0) / a[0]
-    a = (a + 0.0) / a[0]
+
+    if xp.isdtype(b.dtype, 'integral'):
+        b = xp.astype(b, xp.float64)
+    if xp.isdtype(a.dtype, 'integral'):
+        a = xp.astype(a, xp.float64)
```
j-bowhay, line 1223:
```suggestion
    a, b = xp_promote(a, b, force_floating=True)
```

scipy/signal/_filter_design.py
```diff
@@ -1214,13 +1214,21 @@ def tf2zpk(b, a):
...
+
+    b = b / a[0]
```
j-bowhay, line 1225: I've probably just forgotten but how come we can no longer do this in-place?

scipy/signal/_filter_design.py
```diff
@@ -1633,6 +1654,19 @@ def zpk2sos(z, p, k, pairing=None, *, analog=False):
...
 
+    xp = array_namespace(z, p)
+
+    # convert to numpy, convert back on exit   XXX
+    z, p = map(np.asarray, (z, p))
+
+    k = xp.asarray(k)
+    if xp.isdtype(k.dtype, 'complex floating'):
+        if xp.imag(k) != 0:
+            raise ValueError('k must be real')
```
j-bowhay, line 1665: I'm in two minds, would it be clearer to say zero imaginary part as the dtype doesn't have to be real?

scipy/signal/tests/test_filter_design.py
```diff
@@ -462,17 +519,19 @@ def test_basic(self):
...
-            p = np.array([mag * np.exp(theta * deg2rad * 1j)
-                          for theta, mag in zip(thetas, mags)])
-            p = np.concatenate((p, np.conj(p)))
+            thetas = xp.asarray([22.5, 45, 77.5])
+            mags = xp.asarray([0.8, 0.6, 0.9])
+     #       z = xp.asarray([xp.exp(theta * deg2rad * 1j) for theta in thetas])
+            z = xp.exp(1j * deg2rad * thetas)
+            z = xp.concat((z, xp.conj(z)))
+#            p = xp.asarray([mag * np.exp(theta * deg2rad * 1j)
+#                          for theta, mag in zip(thetas, mags)])
```
j-bowhay, line 532: wasn't sure if these comments were meant to be left here

scipy/signal/tests/test_filter_design.py
```diff
@@ -481,12 +540,15 @@ def test_basic(self):
...
                     [1, -0.43288, 1, 1, -0.38959, 0.81]]
+            sos2 = xp.asarray(sos2)
             assert_array_almost_equal(sos, sos2, decimal=4)
 
             # second example
-            z = np.array([np.exp(theta * deg2rad * 1j)
-                          for theta in (85., 10.)])
-            z = np.concatenate((z, np.conj(z), [1, -1]))
+#            z = xp.asarray([xp.exp(theta * deg2rad * 1j)
+#                          for theta in (85., 10.)])
```
j-bowhay, line 548: same here

scipy/signal/_filter_design.py
j-bowhay: also can you add test coverage for this new check

j-bowhay (commented):
A few minor comments

scipy/signal/_filter_design.py
ev-br: Actually, I'm no longer sure why this is here at all. Will double-check.

scipy/signal/_filter_design.py
ev-br: a three-letter word: jax

scipy/signal/_filter_design.py
```diff
@@ -1214,13 +1214,15 @@ def tf2zpk(b, a):
         array([ -2.5+2.59807621j ,  -2.5-2.59807621j]),
         3.0)
     """
+    xp = array_namespace(b, a)
     b, a = normalize(b, a)
-    b = (b + 0.0) / a[0]
-    a = (a + 0.0) / a[0]
+
+    a, b = xp_promote(a, b, xp=xp, force_floating=True)
```
j-bowhay, line 1220: Oops I just noticed I didn't read things as carefully as I thought sorry. For some reason I thought there was an operation involving both `a` and `b` but this isn't the case so originally they were never promoted against each other, is it going to cause an issue that we now promote them?

ev-br: TL;DR Hard to be sure, but on balance I think it's strictly not worse than what we had.

Long version: If I'm to divine the intent of the multipack-era idiom, `b = (b + 0.0) / a[0]; a = (a + 0.0) / a[0]`, it means to ensure that both `b` and `a` and float64. Single precision floats were not a concern back then.  What we definitely do not want is mixed dtypes, where `b` is float32 and `a` is float64 or vice versa. The `xp_promote(a, b, force_floating=True)` idiom makes sure that we have them as either both float32 or float64. 
What is the quality of float32 results, I don't know and I won't be too surprised if it's not stellar. Shall we do something about it? Maybe, at some point (cf https://github.com/scipy/scipy/pull/22896#issuecomment-2852442506). Should we do something about it right now? As long as our existing test suite does not complain about f32, I believe we're in the clear.

j-bowhay (approved):
Looks ready to go in!

ev-br:
Thanks Jake!

PR #22897, opened 2025-04-27
ENH: sparse: ND binary operations support

dschult:
This PR builds on #22873 and between the two of them largely replace the bigger #22540 which split out from #21613

Idea: nD requires COO and means our base class can no longer always and directly convert to CSR to do arithmetic. Rather than copying the CSR methods to COO and updating (as with the previous variants of this PR), here we move the general parts of code to `base.py` while leaving the specialized parts in either COO or CSR. The biggest advantage of moving to `_base.py` is that we delay converting to CSR until we need to. And we centralize a lot of code reducing duplication that was slightly different in each place. Also, we now don't reshape or `tocsr()` until we are about to call the CSR methods `_scalar_binopt` and `_binopt`. This saves a conversion for many cases of the if-logic.

In the process of moving to _base, many simplifications were possible:
- refactor helper `ravel_non_reduced_coords()` to `_convert_to_2d()`. This helper now includes the code to build the 2d array and that was the only reason the old helper was used. This code is used in previously merged functionality too -- like    `nD dot/tensordot/matmul`. We refactored them wherever they convert to 2d.
- splitting `_inequality, __eq__, __ne__` in CSR avoided. Now single `_comparison` method in _base.py. `_comparison(self, other, op)` is the signature. No more `op_name` or `bad_scalar_msg`. The messages are unified via f-string magic. And the `op_name` is obtained from `op.__name__`. The logic surrounding warnings is simplified by checking `op(0, other)` for scalars and `op(0, 0)` for non-scalars. This is more direct than many separate cases, e.g. `le` and `ge` are inefficient for scalar 0 and `gt` for negative scalars, etc.
- `_minimum_maximum` similarly only needs args `(self, other, op)`. Messages now use `op.__name__` and `op(0, other)` replaces the need for a dense_check/lambda function.
- `multiply` and `_divide` follow a similar flow/logic to the others. No shift of code needed because these already allowed formats beyond CSR to work

The processing for all these follows this kind of pattern:
    - Process list-like `other`.
    - handle scalarlike other
    - handle dense other
    - handle sparse other
    - report NotImplemented if no cases above used.

Commits:
- nD bool compares `eq/ne/lt/gt/le/ge` and `dot/tensordot/matmul`, etc
  Refactor `_convert_to_2d` and unify approach to eq and the others.
- nD `multiply` and `_divide` with tests
- nD `maximum` and `minimum` with tests

Left to do from the big nD PR (#21613):
- decide whether to remove `todense_nd` or speed it up
- support broadcasting without copying data

This is built on #22873 so the diff is smaller than it appears. Once that is merged this will look smaller and I'll rebase.  Looking at the most recent 3 commits should show what changes. And they separate into "bool comparisons", "multi & divide" and "maximum/minimum".


scipy/sparse/_base.py
```diff
@@ -475,18 +489,85 @@ def asformat(self, format, copy=False):
...
+                other = other_a
+
+        if self.ndim < 3 and other.ndim < 3:
+            return self.tocsr().multiply(other)
+
+        if self.shape != other.shape:
+            raise ValueError("inconsistent shapes: >2D multiply() does not yet "
+                             "support broadcasting")
+
+        # self is >2D so much be COO
```
perimosocordiae, line 513: much -> must?

scipy/sparse/_base.py
```diff
@@ -475,18 +489,85 @@ def asformat(self, format, copy=False):
...
+            other_a = np.asanyarray(other)
+            if other_a.ndim == 0 and other_a.dtype == np.object_:
+                return NotImplemented
+            try:
+                other.shape
+            except AttributeError:
+                other = other_a
+
+        if self.ndim < 3 and other.ndim < 3:
+            return self.tocsr().multiply(other)
```
perimosocordiae, line 507: non-blocking: Ideally we should have this base class provide all the public API methods, and these owuld dispatch exclusively to private methods (similar to how we already handle `_mul_scalar`). That way, each subclass can have stronger guarantees about the types and shapes of its inputs.

scipy/sparse/_base.py
```diff
@@ -516,23 +597,88 @@ def _broadcast_to(self, shape, copy=False):
...
+            if not op(0, other):
+                if np.isnan(other):  # op is not `ne`, so results are all False.
+                    return self.__class__(self.shape, dtype=np.bool_)
+                else:
+                    csr_ready = (self if self.ndim < 3 else self.reshape(1, -1)).tocsr()
+                    res = csr_ready._scalar_binopt(other, op)
+                    return res if self.ndim < 3 else res.tocoo().reshape(self.shape)
+            else:
+                warn(f"Comparing a sparse matrix with {other} using {op_sym[op]} "
+                     "is inefficient. Try using {op_sym[op_neg[op]]} instead.",
```
perimosocordiae, line 625: missing the f-prefix on this string

scipy/sparse/_base.py
```diff
@@ -516,23 +597,88 @@ def _broadcast_to(self, shape, copy=False):
...
+            return op(self.todense(), other)
+
+        elif issparse(other):
+            # TODO sparse broadcasting
+            if self.shape != other.shape:
+                if op in (operator.eq, operator.ne):
+                    return op == operator.eq
+                raise ValueError("inconsistent shape")
+
+            csr_ready = (self if self.ndim < 3 else self.reshape(1, -1)).tocsr()
```
perimosocordiae, line 648: `csr_self` makes more sense to me

scipy/sparse/_base.py
```diff
@@ -516,23 +597,88 @@ def _broadcast_to(self, shape, copy=False):
...
+            csr_other = (other if other.ndim < 3 else other.reshape(1, -1)).tocsr()
+            if not op(0, 0):
+                result = csr_ready._binopt(csr_other, f'_{op.__name__}_')
+                return result if self.ndim < 3 else result.tocoo().reshape(self.shape)
+            else:
+                # result will not be sparse. Use negated op and then negate.
+                warn(f"Comparing two sparse matrices using {op_sym[op]} "
+                     "is inefficient. Try using {op_sym[op_neg[op]]} instead.",
```
perimosocordiae, line 656: f-string prefix here

perimosocordiae (changes requested):

scipy/sparse/_base.py
dschult: I have moved the public `multiply` from `_compressed.py` to `_base.py` as described.

The `isscalarlike` case is now handled in `_base.py` before dispatching to csr's `_multiply_2d_with_broadcasting`. 

As we improve broadcasting support, we can/should move more out of `csr` and into `base`. For example, the case `isdense(other)` doesn't need to convert to CSR -- once it is unwrapped from all the broadcasting code we can call it from `_base`.

In summary, we now have all the public API methods we've been messing with in _base.py and they dispatch to private methods in the subclasses.  A followup PR can attempt to do this with other methods not considered here (like `sum` and `diagonal` and maybe others).

scipy/sparse/_base.py
```diff
@@ -475,18 +489,86 @@ def asformat(self, format, copy=False):
...
             return self._mul_scalar(other)
-        return self.tocsr().multiply(other)
+
+        if self.ndim < 3:
+            return self.tocsr()._multiply_2d_with_broadcasting(other)
+
+        if not (issparse(other) or isdense(other)):
+            # If it's a list or whatever, treat it like an array
+            other_a = np.asanyarray(other)
+            if other_a.ndim == 0 and other_a.dtype == np.object_:
```
stefanv, line 502: May be helpful (for future us) to add comment to explain this.

scipy/sparse/_base.py
```diff
@@ -516,23 +598,89 @@ def _broadcast_to(self, shape, copy=False):
...
+                result = all_true - inv
+                return result if self.ndim < 3 else result.tocoo().reshape(self.shape)
+
+        elif isdense(other):
+            return op(self.todense(), other)
+
+        elif issparse(other):
+            # TODO sparse broadcasting
+            if self.shape != other.shape:
+                if op in (operator.eq, operator.ne):
```
stefanv, line 646: Comment this one as well.

scipy/sparse/_base.py
```diff
@@ -752,6 +899,17 @@ def __rmatmul__(self, other):
     ####################
 
     def _divide(self, other, true_divide=False, rdivide=False):
+        if not (issparse(other) or isdense(other) or isscalarlike(other)):
```
stefanv, line 902: Document arguments?

stefanv (approved):

stefanv (approved):

PR #22900, opened 2025-04-28
MAINT: fix `np.copyto` warnings on Dask

crusaderky:
As per the Array API Standard, `fill_value` can't be an Array.
Discussion [here](https://github.com/data-apis/array-api/issues/909) to allow it in the next draft.
This triggers a warning in Dask.

This PR also fixes a bug in `stats.ttest_ind` where the returned dtype was the default float, regardless of inputs.
Finally, this PR returns a flaw in device propagation from input to output when returning NaN, with the device reverting to the default one instead of the device of the inputs.

### DIFF
--- a/scipy/stats/_stats_py.py
+++ b/scipy/stats/_stats_py.py
@@ -6762,9 +6762,7 @@ def ttest_ind(a, b, *, axis=0, equal_var=True, nan_policy='propagate',
         raise NotImplementedError(message)
 
     result_shape = _broadcast_array_shapes_remove_axis((a, b), axis=axis)
-    NaN = _get_nan(a, b, xp=xp)
-    if result_shape != ():
-        NaN = xp.asarray(xp.broadcast_to(NaN, result_shape), copy=True)
+    NaN = _get_nan(a, b, shape=result_shape, xp=xp)
     if xp_size(a) == 0 or xp_size(b) == 0:
         return TtestResult(NaN, NaN, df=NaN, alternative=NaN,
                            standard_error=NaN, estimate=NaN)
--- a/scipy/stats/_stats_py.py
+++ b/scipy/stats/_stats_py.py
@@ -6762,8 +6762,9 @@ def ttest_ind(a, b, *, axis=0, equal_var=True, nan_policy='propagate',
         raise NotImplementedError(message)
 
     result_shape = _broadcast_array_shapes_remove_axis((a, b), axis=axis)
-    NaN = xp.full(result_shape, _get_nan(a, b, xp=xp))
-    NaN = NaN[()] if NaN.ndim == 0 else NaN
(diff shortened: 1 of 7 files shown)


scipy/stats/_variation.py
```diff
@@ -110,8 +110,9 @@ def variation(a, axis=0, nan_policy='propagate', ddof=0, *, keepdims=False):
         # The return values, if any, are all nan.
         shp = list(a.shape)
         shp.pop(axis)
-        result = xp.full(shp, fill_value=NaN)
-        return result[()] if result.ndim == 0 else result
+        if shp:
+            return xp.asarray(xp.broadcast_to(NaN, tuple(shp)), copy=True)
```
crusaderky, line 114: Do we care about the return value to be writeable?

scipy/stats/_stats_py.py
```diff
@@ -6762,8 +6762,9 @@ def ttest_ind(a, b, *, axis=0, equal_var=True, nan_policy='propagate',
         raise NotImplementedError(message)
 
     result_shape = _broadcast_array_shapes_remove_axis((a, b), axis=axis)
-    NaN = xp.full(result_shape, _get_nan(a, b, xp=xp))
-    NaN = NaN[()] if NaN.ndim == 0 else NaN
+    NaN = _get_nan(a, b, xp=xp)
+    if result_shape != ():
+        NaN = xp.asarray(xp.broadcast_to(NaN, result_shape), copy=True)
```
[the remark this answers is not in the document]
crusaderky: This was also unwittingly reverting the result type back to the default dtype on torch

scipy/stats/tests/test_stats.py
```diff
@@ -6149,7 +6138,7 @@ def test_ttest_ind_axis_size_zero(self, b, expected_shape, xp):
         # The results should be arrays containing nan with shape
         # given by the broadcast nonaxis dimensions.
         a = xp.empty((3, 1, 0))
-        b = xp.asarray(b)
+        b = xp.asarray(b, dtype=a.dtype)
```
crusaderky, line 6141: When SCIPY_DEFAULT_DTYPE=float32, this was float64. The return value was, incorrectly float32 instead of float64 so the test passed anyway.

scipy/stats/_variation.py
mdhaber: Please just pass `full` a `dtype` here and above.

mdhaber (commented):
I have another PR open that adds dtype tests, so fine that those are left out.

mdhaber:
The device propagation is also fixed in gh-22856, but I can resolve the conflicts there.

mdhaber (approved):
`xp_result_device` looks good to have around.

lucascolley (approved):
thanks both!

PR #22908, opened 2025-04-29
MAINT: bump qhull to 2020.2

lucascolley:
gh-22893


lucascolley:
@rgommers ready for merge, I just checked out the 2020.2 tag and copied the directory across, removing files which didn't already exist.

Moving over to a submodule/script like we use for prima is more involved and can be done in a follow-up.

Jose1654 (dismissed):

rgommers (commented):
Nice, glad to see it was fairly painless:)

You missed the three `.txt` files in `scipy/spatial/qhull_src/`, could you please include those? Then this looks good to go.

lucascolley:
thanks for catching that!

PR #22915, opened 2025-04-30
BLD: optimize sdist contents through a dist script

rgommers:
This reduces the size of the sdist by about 50%, from 60 MB to 30 MB.

Also remove some images from `doc/source/_static/gitpod`. The doc content that used them was removed a long time ago.


lucascolley (approved):
LGTM!

tylerjereddy (approved):
The command I usually use for `sdist` generation for releases indeed has a resulting tarball that drops in size from 57 MB->29 MB locally.

The changes look sensible to me as well, and the CI seems to have at least two `sdist`-related jobs that passed.

`meson` docs suggest this command runs after the sources are moved in, but before the archive is populated. I guess that makes sense.

tylerjereddy:
thanks Ralf, Lucas

PR #22917, opened 2025-05-01
DEV: fix invocation of linter on Windows

stefanv:
On that platform, depending on the shell, the `#!/usr/bin/env python` may not be used. To work around that problem, launch `sys.executable`, instead of trying to run the Python script directly.

Since `spin` is always run from the root of the repo, also no need to provide the full path to the linting tool.


rgommers (approved):
Thanks Stefan!

PR #22918, opened 2025-05-01
TST: `linalg` add test coverage to exception handling for invalid shapes in `linalg._decomp_cossin._cossin`

zitongzhoueric:
This PR adds a test case to ensure exceptions are properly raised when `linalg._decomp_cossin._cossin` has arguments with non-decomposable shapes.


j-bowhay (approved):
Thanks!

PR #22919, opened 2025-05-01
TST: `optimize` add test coverage to callback warning in `optimize.least_squares`

zitongzhoueric:
Adds a test to ensure warning is emitted when `optimize.least_squares` is called with `method='lm'` and `callback` function, and that the `callback` function is not used.


j-bowhay (approved):
Thanks!

Issue #22904, opened 2025-04-28
BUG: Wrong use of `__builtin_prefetch()`


MLopez-Ibanez:
### Describe your issue.

This code is wrong:
https://github.com/scipy/scipy/blob/6a918e90245f0c456cb2ddd73e9e8d34e44c4259/scipy/spatial/ckdtree/src/query.cxx#L295

The 3rd argument of `__builtin_prefetch()` should be either 0, 1, 2 or 3 not `m`. https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html#index-_005f_005fbuiltin_005fprefetch

When this code was rewritten in https://github.com/scipy/scipy/commit/6975656966bd526570af5616096487e10921f163 by @rainwoodman to use `NPY_PREFETCH()`, the previous code used `m` to fill a cache line. However, `__builtin_prefetch()` always reads a full cache line and the third argument is the expected level of locality. Values larger than 3 are invalid but they are possibly interpreted as 3. But the real bug is that the locality of the prefetch function depends on `m` which is the dimension of the points!

This line:
https://github.com/scipy/scipy/blob/6a918e90245f0c456cb2ddd73e9e8d34e44c4259/scipy/spatial/ckdtree/src/ckdtree_decl.h#L12C1-L12C63

should simply be:

```C
#define CKDTREE_PREFETCH(x)  NPY_PREFETCH(x,0, 3)
```

### Reproducing Code Example

```python
NA
```

### Error message

```shell
NA
```

### SciPy/NumPy/Python version and system information

```shell
NA
```

lucascolley:
Would you like to submit a PR? And is there a test that could be added for this?

MLopez-Ibanez:
> Would you like to submit a PR? And is there a test that could be added for this?

I don't know how it could be tested. It is unclear to me if the prefetch is actually doing anything useful, so even benchmarking it may not show any differences.

tylerjereddy:
Cool! Out of curiosity, how did you find this? Static analyzer or just reading the source?

cc @sturlamolden, but I agree this would be hard to test IIUC

MLopez-Ibanez:
> Cool! Out of curiosity, how did you find this? Static analyzer or just reading the source?

Just reading the source. I wanted to create a stand-alone version of the code that I could use in a project of mine.

MLopez-Ibanez:
> cc [@sturlamolden](https://github.com/sturlamolden), but I agree this would be hard to test IIUC

[...]

PR #22922, opened 2025-05-01
BUG: Fix wrong use of prefetching in cKDTree

sturlamolden:
#### Reference issue
#22904

Wrong use of prefetching

#### Additional information
The solution is suboptimal for Apple silicon with 128B cache line and 16KB page size.
The hardware dependent lines are tagged.

### DIFF
--- a/scipy/spatial/ckdtree/src/ckdtree_decl.h
+++ b/scipy/spatial/ckdtree/src/ckdtree_decl.h
@@ -7,9 +7,66 @@
  * */
 #include <numpy/npy_common.h>
 #include <cmath>
-#define CKDTREE_LIKELY(x) NPY_LIKELY(x)
-#define CKDTREE_UNLIKELY(x)  NPY_UNLIKELY(x)
-#define CKDTREE_PREFETCH(x, rw, loc)  NPY_PREFETCH(x, rw, loc)
+#include <new>
+
+#if defined(__cpp_lib_hardware_interference_size)
+    #define CACHE_LINE std::hardware_destructive_interference_size
+#else
+    // 64 is OK for x86 and amd64, not for Apple silicon with 128
+    // On Apple silicon: Use a C++17 library that has  std::size_t hardware_destructive_interference_size
+    // defined or suffer the pain of extra prefetching.
+    #define CACHE_LINE 64
+#endif // __cpp_lib_hardware_interference_size
+
+#if defined(__GNUC__) 
+
+#define CKDTREE_LIKELY(expr)    (__builtin_expect(!!(expr), 1))
+#define CKDTREE_UNLIKELY(expr)  (__builtin_expect(!!(expr), 0))
+
+inline void 
+prefetch_datapoint(const npy_float64 *x, const npy_intp m) 
+{
+    // The data point can live on multiple cache lines
+    // so we must hit all of them
+    char *cur = (char*)x;
+    char *end = (char*)(x+m);    
+    do { 
+        __builtin_prefetch((void*)cur, 0, 3);
+        cur += CACHE_LINE;
+    } while (CKDTREE_UNLIKELY(cur < end));
+}
+  
+#else 
+
+#define CKDTREE_LIKELY(x) NPY_LIKELY(x)     // FIXME: Does not do anything as of May 2025, better replace with C++20 [[likely]]
+#define CKDTREE_UNLIKELY(x) NPY_UNLIKELY(x) // FIXME: Does not do anything as of May 2025, better replace with C++20 [[unlikely]]
+  
+#if defined(_WIN32) 
+  
+#include <xmmintrin.h> 
+  
+inline void 
+prefetch_datapoint(const npy_float64 *x, const npy_intp m) 
+{
+    // The data point can live on multiple cache lines
+    // so we must hit all of them
+    char *cur = (char*)x;
+    char *end = (char*)(x+m);    
+    do { 
+        _mm_prefetch((const char*)cur,_MM_HINT_T0);
+        cur += CACHE_LINE;
+    } while (CKDTREE_UNLIKELY(cur < end));
+}
+
+#else 
+
+#define prefetch_datapoint(x,m)
+
+#endif // _WIN32
+#endif // __GNUC__
+
+// rw is always 0
+#define CKDTREE_PREFETCH(x,rw,m) prefetch_datapoint(x,m)
 
 #define ckdtree_intp_t npy_intp
 #define ckdtree_fmin(x, y)   fmin(x, y)
--- a/scipy/spatial/ckdtree/src/query.cxx
+++ b/scipy/spatial/ckdtree/src/query.cxx
@@ -157,11 +157,22 @@ struct nodeinfo_pool {
     char *arena;
     char *arena_ptr;
 
+    // A cache line is 64 on x86 and amd64, but 128 on some Apple silicon
+    // This mainly for alignment, 64 will do fine on Apple silicon as well.
+    // CACHE_LINE is defined in ckdtree_decl.h
+
+    // A page is 4096 on x86 and amd64, but 16384 on Apple silicon.
+    // Arena size in full pages prevents new from searching the heap.
+    // The purpose of nodeinfo_pool is to have an O(1) allocator instead of
+    // asking new or malloc to scan the heap for the smallest fitting piece 
+    // of free memory.
+    #define ARENA 16384
+
     nodeinfo_pool(ckdtree_intp_t m) {
         alloc_size = sizeof(nodeinfo) + (3 * m -1)*sizeof(double);
-        alloc_size = 64*(alloc_size/64)+64;
-        arena_size = 4096*((64*alloc_size)/4096)+4096;
-        arena = new char[arena_size];
+        alloc_size = (ckdtree_intp_t) CACHE_LINE*(alloc_size/CACHE_LINE)+CACHE_LINE; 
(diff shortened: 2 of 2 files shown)


tylerjereddy (commented):
To clarify--you've labelled this a bug, but there is no way the issue can cause numerically incorrect results, right? It is just sub-optimal performance?

I checked the `asv` benchmarks on x86_64 Linux and they were neutral on this change via `asv continuous -E virtualenv -e -b ".*time_query.*" main fix-prefetching`. I believe that is not surprising based on your description of the need for very large input for the cache efficiency to become relevant.

I wonder if we could convince @seiko2plus to take a quick look--I believe Sayed may have sufficient knowledge of low-level caching to sanity check the change.

sturlamolden:
> To clarify--you've labelled this a bug, but there is no way the issue can cause numerically incorrect results, right? It is just sub-optimal performance?
> 
> I checked the `asv` benchmarks on x86_64 Linux and they were neutral on this change via `asv continuous -E virtualenv -e -b ".*time_query.*" main fix-prefetching`. I believe that is not surprising based on your description of the need for very large input for the cache efficiency to become relevant.
> 
> I wonder if we could convince @seiko2plus to take a quick look--I believe Sayed may have sufficient knowledge of low-level caching to sanity check the change.

It is a bug because the locale argument should only be 0, 1, 2 or 3, and it must be a compile-time integer. We are passing the dimensionality of the data set as a run-time integer to this argument. I am surprised the compiler eats it.

But say that it does use the locale argument, then ```m=2``` would mean to leave data in L2, ```m=3``` would mean to prefetch to L1, ```m>3``` would be bogus. 

I do not know what the compiler will do when we feed garbage to ```__builtin_prefetch```. I cannot find a description of what might happen. So for now I would say we get undefined behaviour. 

It is also a bug due to the address. A data point can reside on more than one cache line.

Note that this is not new code in SciPy. We had this before. The very reason the query functions pass the ```m``` argument to ```CKDTREE_PREFETCH``` is because we had the code in this PR before.

sturlamolden:
Here is the equivalent code from SciPy 1.0.X. 

https://github.com/scipy/scipy/blob/83eb32d32f9a4d8fcb03dc75478f39c8f1cb7686/scipy/spatial/ckdtree/src/ckdtree_methods.h#L28-L60

sturlamolden:
This is the commit that introduced the bug

https://github.com/scipy/scipy/commit/6975656966bd526570af5616096487e10921f163

sturlamolden:
> I checked the `asv` benchmarks on x86_64 Linux and they were neutral on this change via `asv continuous -E virtualenv -e -b ".*time_query.*" main fix-prefetching`. I believe that is not surprising based on your description of the need for very large input for the cache efficiency to become relevant.

But it is good that this change was neutral! 

It means that the extra overhead in the ```CKDTREE_PREFETCH``` function were of no consequence compared to just a single call to ```NPY_PREFETCH```. Since it is called on each data point it could have been harmful for the performance of low-dimensional kd-trees, where a single prefetch statement is optimal. 

I changed the loop to do-while and added ```NPY_UNLIKELY``` for the conditional to make the common case of a single prefetch be fast. do-while makes it run at least once, not needing a conditional check for the first pass, and then informs the compiler that more than one pass is unlikely. Compared with the while-loop we had up to SciPy 1.2, it should now allow for better optimization of the common case. Potentially it should allow the compiler to remove all overhead. Since you found the PR to be neutral it means this goal was achieved. Simply calling ```NPY_PREFETCH``` once did not make queries on a low-dimensional kd-tree go any faster.

seiko2plus:
@tylerjereddy, thank you for pinging me.

> It is a bug because the locale argument should only be 0, 1, 2 or 3, and it must be a compile-time integer. We are passing the dimensionality of the data set as a run-time integer to this argument. I am surprised the compiler eats it.

Hi @sturlamolden, IDK :) its a bug and not a bug. The `NPY_PREFETCH` macro is controlled by configure-time checks that are only defined in NumPy's private config headers. When used in SciPy, it becomes `#define NPY_PREFETCH(x, rw, loc)` - an empty macro that does nothing.

That's why the compiler accepts any thing without complaint about immediate value. See the NumPy's header:
https://github.com/numpy/numpy/blob/49056192330487ff49a142e5280e24461b638723/numpy/_core/include/numpy/npy_common.h#L71-L84
We shouldn't be exposing these internal macros for external use.

> I do not know what the compiler will do when we feed garbage to __builtin_prefetch. I cannot find a description of what might happen. So for now I would say we get undefined behaviour.

I think it won't cause your program to crash, but its a waste of bandwidth, leads to flush other data out of cache so conflict misses will definitely increased. Some CPUs may not handle invalid pointers probably which can cause performance regression.

sturlamolden:
Are NPY_LIKELY and NPY_UNLIKEKLY empty macros as well?

sturlamolden:
Can we use C++20 in SciPy?

seiko2plus:
> Are NPY_LIKELY and NPY_UNLIKEKLY empty macros as well?

Yes.

Regarding `NPY_PREFETCH`, I didn't dive into the functions that use it, but maybe be before adding software prefetching, would it be better improve memory access patterns (if not) instead? IMHO, hardware prefetching nowdays is more than enough.

> Can we use C++20 in SciPy?

No, C++17 is the limit:
https://github.com/scipy/scipy/blob/4674ab33af519eddc34709c9f75106f95eaf0482/meson.build#L10-L11

sturlamolden:
Squashing to one commit...

sturlamolden:
> To clarify--you've labelled this a bug, but there is no way the issue can cause numerically incorrect results, right? It is just sub-optimal performance?

As we have now established that ```NPY_PREFETCH```, ```NPY_LIKELY``` and ```NPY_UNLIKELY``` are empty macros, it means the issue cannot cause any logic errors. The errorneous code compiles to nothing.

What prevents the code from crashing when compiling is that the  ```NPY_PREFETCH``` is empty. It is only for internal use in NumPy.

The issue is a regression though, that happened between SciPy 1.2 and 1.3. I think @rainwoodman might not have realised the macros were empty. Also he did not use the locale argument correctly, which i what @MLopez-Ibanez observed.

This PR restores it and also does something to cater for Apple Silicon (if the C++17 compiler is up to date).

nickodell:
> But it is good that this change was neutral!

Do we have a benchmark that shows that this change *helps*? Prefetching, and marking a branch as likely or unlikely, does have a cost. CPUs already contain hardware to attempt to predict memory and branch access.

I think we have a plausible theoretical argument that this change could help. But can we show that it does?

sturlamolden:
@nickodell 

> > But it is good that this change was neutral!
> 
> Do we have a benchmark that shows that this change _helps_? Prefetching, and marking a branch as likely or unlikely, does have a cost. CPUs already contain hardware to attempt to predict memory access and whether a branch will be taken.
> 
> I think we have a plausible theoretical argument that this change could help. But can we show that it does?

In 2015 when I wrote this code, we did a lot of testing. Both prefetching and marking branches as likely/unlikely helped a lot. But that was _then_. Both our C++ code and computers have evolved since then.

E.g. we started to use templates and C++17 closures to replace some of the branches with specialized kd-trees. Distance functions are now templated. We used to have a switch that was called whenever a distance were to be computed. This was likely where likely/unlikely helped the most. It allowed us to make sure Euclidian distances were prioritized. Now we are doing it with templates instead. So today I expect likely/unlikely to have less effect than it did in 2015. Maybe software branch-prediction is not needed anymore.

For software prefetching the condition that caused it to be useful is still present. The memory access pattern while running a query is _completely random_ and cannot be predicted without some extra information. We do have that extra information, but is requires special knowledge about the data structure. 

It does not have to be this way. The culprit is the way we set up cKDTree internally. 

The way we found out what was useful in 2015 was to sit down and test various versions of the code (separated from SciPy) and see what worked best. If I remember correctly we used an ipython (now Jupyter) notebook. We could do the same thing now. We would have to use large data sets for testing, though. The ones we use for benchmarking are way too small. 

**Thus, all I can say now is that the prefetching and branch-prediction code was broken by accident in SciPy 1.3 and that it was very useful 10 years ago.** I will not make too strong claims about it now, particularly as most of the branch prediction code was replaced by using templates.

sturlamolden:
I have now tested with 3 GB data sets. I could not get any improvement from software prefetching or software branch prediction. This is on Apple silicon M4.

Very different from what I observed in 2015.

I think we should just throw it out. 

Quadrupling the arena size did help though, which is expected with larger page size. 

```python
import numpy as np
import ckdtree0
import ckdtree1
import scipy
from scipy.spatial import cKDTree
```
```python
data = np.random.rand(30000000,12)
qdata = np.random.rand(1000,12)
```
```python
kdt0 = ckdtree0.cKDTree(data)
kdt1 = ckdtree1.cKDTree(data)
kdt2 = cKDTree(data)
```
```python
%%timeit 
kdt0.query(qdata) # no software branch-prediction or software prefetching, 16KB arena
```
    521 ms ± 1.12 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

```python
%%timeit 
kdt1.query(qdata) # software branch-prediction and software prefetching, 16KB arena
```
    548 ms ± 9.12 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

```python
%%timeit 
kdt2.query(qdata) # SciPy 1.13.1, no software branch-prediction or software prefetching, 4KB arena
```
    735 ms ± 853 μs per loop (mean ± std. dev. of 7 runs, 1 loop each)

sturlamolden:
I am closing this PR. I will post a new one where software prefetching and branch prediction are removed.

PR #22867, opened 2025-04-21
WIP: linalg: Comparison PR for #22838

ilayn:
#### Reference issue
This is a PR for comparing C/C++ implementation flows discussed in #22838, not to be merged

cc: @ev-br

PR #22924, opened 2025-05-02
ENH: linalg: low-level nD support, take 3

ev-br:
#### Reference issue

This is a synthesis of https://github.com/scipy/scipy/pull/22838 and https://github.com/scipy/scipy/pull/22867

Add nD batch support and structure detection to `linalg.inv`, with the following features:

- low-level looping over slices of an ndim>2 array
- structure detection to transparently choose between a triangular, Cholesky or general inverse, per slice
- `assume_a` argument to shortcut the detection when the structure is known
- condition number checks to detect ill-conditioned problems

What is not in this PR and is left for follow-ups:

- further memory and performance improvements (there are quite a few, potentially)
- avoiding a potential UB from using `malloc`-ed buffers for LAPACK. This is IMO OK for a follow-up since essentially the same problem exists in numpy.linalg, for more than a decade, and caused no problems so far;
- reenable `overwrite_a`
- make error handling more flexible, along the lines of the discussion in https://github.com/numpy/numpy/pull/28782. ATM, a single singular matrix in a slice halts processing the rest of the batch.
- deduplicate the internals, make it easier to use things from both C++ (as here) and C (elsewhere)
- add a "sym/herm" mode for symmetric/hermitian but not positive definite matrices (https://github.com/scipy/scipy/pull/22924#issuecomment-2862357697)
- investigate if using "packed" LAPACK routines give a measurable speed-up (https://github.com/scipy/scipy/pull/22924#issuecomment-2855256821)
- extend/generalize the approach to `solve`

#### Additional information

Crude timings. First, stacks of small matrices: low-level looping is quite a bit faster than 1.5.2, is within a factor of $\pi$ to `np.linalg.inv`. Some of this perf difference can surely be gotten back.

```
In [43]: n = 10

In [44]: a = np.ones((n//2, n//2, n//2, n, n), dtype=float) + 8*np.eye(n)

In [45]: %timeit inv0(a, check_finite=False)     # SciPy 1.5.2
1.27 ms ± 10.5 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [46]: %timeit np.linalg.inv(a)
269 μs ± 1.78 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [47]: %timeit inv(a, check_finite=False)
524 μs ± 14.5 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [48]: %timeit inv(a, check_finite=False, assume_a="pos def upper")
507 μs ± 1.28 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

Next, a single moderately large matrix: algorithm selection makes a difference

```
In [32]: n = 500; b = np.ones((n, n)) + np.eye(n)

In [33]: %timeit np.linalg.inv(b)
7.34 ms ± 162 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [34]: %timeit inv(b, assume_a="general", check_finite=False)
8.13 ms ± 395 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [35]: %timeit inv(b, assume_a="pos def upper", check_finite=False)
6.08 ms ± 176 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [36]: %timeit inv(b,  check_finite=False)
6.49 ms ± 143 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```


scipy/linalg/_common_array_utils.h
```diff
@@ -316,4 +324,109 @@ isschur(const double* restrict data, const Py_ssize_t n)
...
+        }
+        if (c <= ub) { break; }
+    }
+    *lower_band = lb;
+    *upper_band = ub;
+}
+
+
+inline void
+bandwidth_c(const SCIPY_C* restrict data, const Py_ssize_t n, const Py_ssize_t m, Py_ssize_t* lower_band, Py_ssize_t* upper_band)
```
Kai-Striega, line 385: As someone who hasn't been keeping up with the changes here; what's ``SCIPY_C``? Is this type documented anywhere?

Kai-Striega:
@ev-br I haven't really been keeping up with the changes, but if you want me to spend some time over the next week to review this code I'd be happy to.

scipy/linalg/_common_array_utils.h
ev-br: IIUC, this is a C `complex float`. This source file, `linalg/_common_array_utils.h`, is used in C code in `sqrtm` etc and is not used here in this PR.

A technical challenge AFAICS, is to make this set of basic routines reusable from both C++ code in `_inv` here and from C code in `sqrtm` and its relatives. Specific steps:
- deduplicate the LAPACK declarations. The C code uses C complex types; to make it C/C++ ready we need to make them use `npy_cfloat/npy_cdouble`.
- Cook up a way to cast to either `double complex` in C and `std::complex` in C++ (In this PR the latter is done via `reinterpret_cast`, as a shortcut.) Alternatively, ditch `complex` and always work with `npy_cdouble`. Either of this should be done in both `sqrtm` and here.
- `#define` around the `restrict` keyword not available in C++.
- Move templated functions this PR adds to C and add type overloads in C++. 

This is not too difficult I think, but it will inflate the diff size here and will touch unrelated code, so ISTM is best deferred to a follow-up.

ev-br:
> I haven't really been keeping up with the changes, but if you want me to spend some time over the next week to review this code I'd be happy to.

Would be great indeed @Kai-Striega !

scipy/linalg/_common_array_utils.h
ev-br: > deduplicate the LAPACK declarations. The C code uses C complex types; to make it C/C++ ready we need to make them use npy_cfloat/npy_cdouble.

This part IMO is as simple as committing the `build/_lapack_subroutines.h` file which is autogenerated during the `cython_lapack` build,  cf https://github.com/scipy/scipy/issues/20002 + cleaning up ad hoc declarations across the codebase.
Which is one more reason to make it a separate follow-up PR.

ev-br:
> use the packed versions of the LAPACK routines

Let's chalk it up to the list of potential improvements.

ilayn:
I couldn't find where the fallback is but if cholesky fails `pos` should fall back to `her` and not `gen` since it's implied and also you already have the other half.

ev-br:
The fallback is here (no `break` in a `switch`) : https://github.com/scipy/scipy/pull/22924/files#diff-68a304e3b40ad565a25f847144c46b1a46b2a81e978593d9cb465f26180a3bd4R281

sym/herm is not implemented ATM. Can be added without too much difficulty. Let's gather the set of hard requirements for a minimal viable version though, before expanding the scope here.

ev-br:
This is ready from my side. 
One potential API decision is what is the way for express "I know the input is symmetric positive definite, and please only use an upper triangle".
Elsewhere (most notably, `solve`), there is an extra keyword, `assume_a="pos", lower=False`. This additional keyword argument only used for symmetric matrices and ignored otherwise. IMO such arguments which are used or ignored depending on other arguments are an antipattern, and this PR uses `assume_a="pos upper"`. This way, 
- `assume_a="pos"` means "feel free to pick either triangle"
- `assume_a="pos upper"` means "use the upper triangle only" and `assume_a="pos lower"` means "use the lower triangle only".

This API avoids an antipattern and is consistent with `assume_a="upper triangular"` for triangular matrices. Thoughts?

Also, since this PR adds warnings for ill-conditioned inputs, which was not checked before, it is expected that some usage will see new warnings. While we do want these new checks, it'd be best to act on this PR early in the 1.17 cycle, so that downstream has time to adapt.

ev-br:
It's been two weeks. In case somebody wants to take a look and needs more time, great, please drop a note here. Otherwise I'm planning to land this in a week (around June 9th) and iterate.

ev-br:
Given that there are no further comments, let's land this and iteratively improve in the 1.17 release cycle. Cross-ref https://github.com/scipy/scipy/issues/23141 for a laundry list of potential improvements.
Thanks @ilayn , @Kai-Striega!

scipy/linalg/src/_common_array_utils.hh
```diff
@@ -0,0 +1,460 @@
...
+#include "npy_cblas.h"
+
+/*
+ * declare LAPACK prototypes
+ */
+
+
+/* ?GESV */
+extern "C" {
+CBLAS_INT
```
ilayn, line 17: @ev-br Can you remind me why or what we are returning integers from these LAPACK functions? Is there something specific for these three sets?

ev-br: No reason. This is me parroting numpy, https://github.com/numpy/numpy/blob/main/numpy/linalg/umath_linalg.cpp#L98, in early iterations. Can be cleaned up or left as is no problem.

ilayn: These don't return any integers so I was wondering where the integers are coming from.

ev-br: At a pure guess, this comes from whoever first wrote umath_linalg predecessor all the way back, https://github.com/ContinuumIO/gulinalg/blob/master/gulinalg/src/gulinalg.c.src#L81

And C is relatively lax (as in `int main` vs `void main` etc) so nobody bothered or even noticed it for over a decade.

PR #22925, opened 2025-05-02
ENH: stats: Add matrix variate t distribution

### LINKED ISSUE #22870: ENH: stats: add matrix t-distribution

Adding a matrix t-distribution would make `stats` more complete and improve support for Bayesian methodology. 
    
  - The matrix t-distribution is well-understood in the literature. It appears in the standard monograph by Gupta and Nagar https://www.taylorfrancis.com/books/mono/10.1201/9780203749289/matrix-variate-distributions-nagar-gupta. 

  - It also comes up naturally in Bayesian inference, as the posterior predictive distribution of a multivariate linear regression model with a normal-inverse-Wishart prior and a matrix normal likelihood.

  - There are matrix t-distribution implementations in R, Mathematica, and Julia. 

I wrote a scipy-compatible implementation of the matrix t-distribution for a recent project. This implementation has methods analogous to the ones for `stats.matrix_normal`. The code would fit into `stats._multivariate`, without the need to change any of the existing code.

I am not aware of any alternatives. 

There's a potential hazard when rolling one's own: Unlike the matrix normal distribution, the matrix t-distribution cannot be sampled by vectorizing a matrix and sampling the corresponding multivariate t-distribution. 
See the note at https://github.com/scipy/scipy/blob/1e9a58feebf8a18bfa6c13d649fd24286f13866c/scipy/stats/_multivariate.py#L1074
This happens because of the way that the among-row covariance matrix and among-column covariance matrix interact in the PDF.

_No response_

johnmdusel:
#### Reference issue
Closes  [issue #22870](https://github.com/scipy/scipy/issues/22870).

Background explanation in [this post on the developer forum](https://discuss.scientific-python.org/t/feature-addition-adding-matrix-t-distribution-to-scipy-stats/1845)

#### Additional information

From the PR checklist:

- This code can be distributed under a BSD license. I followed the existing random matrix implementation for data validation and wrote the methods from scratch.

- I provided unit tests in `stats.tests.test_multivariate.TestMatrixT`. These are mostly simiar to the ones for `matrix_normal`, except that the moment checking and sample checking also perform a spot-check of two basic theorems from [Gupta and Nagar (2000)](https://www.taylorfrancis.com/books/mono/10.1201/9780203749289/matrix-variate-distributions-nagar-gupta). All unit tests pass locally.

- Re: style: I used `black` to format my additions to `stats._multivariate`. For unit tests and benchmarking I followed the style of existing code.

- I provided benchmarks in `stats.MatrixSampling`. This also includes benchmarks for `rvs` methods of `matrix_normal` and `invwishart`.

- All public functions have docstrings with examples, and I've verified that the documentation builds locally and renders correctly.


scipy/stats/_multivariate.py
```diff
@@ -1448,6 +1449,631 @@ def entropy(self):
...
+        """
+
+        # Process mean
+        if mean is not None:
+            mean = np.asarray(mean, dtype=float)
+            meanshape = mean.shape
+            if len(meanshape) != 2:
+                raise ValueError("Array `mean` must be 2D.")
+            if np.any(meanshape == 0):
+                raise ValueError("Array `mean` has invalid shape.")
```
tylerjereddy, line 1621: Lacks test coverage if it matters.

scipy/stats/_multivariate.py
```diff
@@ -1448,6 +1449,631 @@ def entropy(self):
...
+                rowcov = rowcov * np.identity(1)
+        elif rowcov.ndim == 1:
+            rowcov = np.diag(rowcov)
+        rowshape = rowcov.shape
+        if len(rowshape) != 2:
+            raise ValueError("Array `rowcov` must be a scalar or a 2D array.")
+        if rowshape[0] != rowshape[1]:
+            raise ValueError("Array `rowcov` must be square.")
+        if rowshape[0] == 0:
+            raise ValueError("Array `rowcov` has invalid shape.")
```
tylerjereddy, line 1640: The 3 error code paths above are not covered by tests, in case it matters.

scipy/stats/_multivariate.py
```diff
@@ -1448,6 +1449,631 @@ def entropy(self):
...
+        numrows = rowshape[0]
+        if np.isnan(rowcov).any():
+            import pdb; pdb.set_trace()
```
tylerjereddy, line 1643: I suspect we don't want to drop the user into the debugger here--residual debugging code to clean up?

scipy/stats/_multivariate.py
```diff
@@ -1448,6 +1449,631 @@ def entropy(self):
...
+                colcov = colcov * np.identity(1)
+        elif colcov.ndim == 1:
+            colcov = np.diag(colcov)
+        colshape = colcov.shape
+        if len(colshape) != 2:
+            raise ValueError("Array `colcov` must be a scalar or a 2D array.")
+        if colshape[0] != colshape[1]:
+            raise ValueError("Array `colcov` must be square.")
+        if colshape[0] == 0:
+            raise ValueError("Array `colcov` has invalid shape.")
```
tylerjereddy, line 1662: Above 3 error code paths not covered by tests, if it matters.

scipy/stats/_multivariate.py
```diff
@@ -1448,6 +1449,631 @@ def entropy(self):
...
+    if size == 1:
+        iw_samples = iw_samples[np.newaxis, ...]
+    chol_samples = np.empty_like(iw_samples)
+    for idx in range(size):
+        # try:
+        chol_samples[idx] = scipy.linalg.cholesky(
+            iw_samples[idx], lower=True, check_finite=False
+        ).reshape(iw_samples.shape[1:])
+        # except ValueError:
+        #     import pdb; pdb.set_trace()
```
tylerjereddy, line 2073: minor: leftover debug comments

tylerjereddy (commented):
Looks like you made a solid effort to follow our usual practices. I added a few surface-level comments ahead of the review by `stats` regulars/domain experts.

scipy/stats/_multivariate.py
johnmdusel: Resolved in e45cb56. Thanks for catching this oversight.

scipy/stats/_multivariate.py
johnmdusel: Resolved in e45cb56.

scipy/stats/_multivariate.py
johnmdusel: Added coverage in b1916d6.

scipy/stats/_multivariate.py
johnmdusel: Added coverage in b1916d6.

scipy/stats/_multivariate.py
johnmdusel: Added coverage in b1916d6.

johnmdusel:
Is this PR still active? Some tests show as failed, but I’m unsure how to resolve them on my end. I’m happy to make any necessary changes to fix them.

scipy/stats/_multivariate.py
```diff
@@ -1448,6 +1449,626 @@ def entropy(self):
...
+    .. math::
+        \mathcal{T}_{m,n} = \\frac{\Gamma_n \left( \\frac{\mathrm{df} + m + n - 1}{2} \\right)}{
+        \Gamma_n \left( \\frac{\mathrm{df} + n - 1}{2} \\right)\pi^{mn / 2}\left|\Sigma\\right|^{n/2}
+        \left|\Omega\\right|^{m/2}}\left|I_n + ( X - \mathrm{M} )^T \Sigma^{-1} ( X - \mathrm{M} ) \Omega^{-1}\\right|^{
+        -\\frac{\mathrm{df} + m + n - 1}{2}}
+
+    or, alternatively,
+
+    .. math::
+        \mathcal{T}_{m,n} = \\frac{\Gamma_m \left( \\frac{\mathrm{df} + m + n - 1}{2} \\right)}{
```
lucascolley, line 1538: at a guess, this may fix most of the CI failures.

```suggestion
    .. math::

        \mathcal{T}_{m,n} = \\frac{\Gamma_n \left( \\frac{\mathrm{df} + m + n - 1}{2} \\right)}{
        \Gamma_n \left( \\frac{\mathrm{df} + n - 1}{2} \\right)\pi^{mn / 2}\left|\Sigma\\right|^{n/2}
        \left|\Omega\\right|^{m/2}}\left|I_n + ( X - \mathrm{M} )^T \Sigma^{-1} ( X - \mathrm{M} ) \Omega^{-1}\\right|^{
        -\\frac{\mathrm{df} + m + n - 1}{2}}

    or, alternatively,

    .. math::

        \mathcal{T}_{m,n} = \\frac{\Gamma_m \left( \\frac{\mathrm{df} + m + n - 1}{2} \\right)}{
```

lucascolley (commented):
the lint failures shown at https://github.com/scipy/scipy/actions/runs/14981775469/job/42087379243?pr=22925 are an easy fix—they are just due to some lines being too long

scipy/stats/_multivariate.py
johnmdusel: Thanks for the tip. Docstrings are updated and and docs build/render correctly locally.

scipy/stats/_multivariate.py
```diff
@@ -1448,6 +1449,657 @@ def entropy(self):
...
+
+    The covariance matrices specified by `rowcov` and `colcov` must be
+    (symmetric) positive definite. If the samples in `X` have shape `(m,n)`
+    then `rowcov` must have shape `(m,m)` and `colcov` must have shape `(n,n)`.
+    Covariance matrices must be full rank.
+
+    The probability density function for `matrix_t` is
+
+    .. math::
+        \mathcal{T}_{m,n} = \\frac{
```
lucascolley, line 1530:
```suggestion
    .. math::

        \mathcal{T}_{m,n} = \\frac{
```

scipy/stats/_multivariate.py
```diff
@@ -1448,6 +1449,657 @@ def entropy(self):
...
+            \\right)
+            \pi^{mn / 2}
+            \left( \det \Sigma \\right)^{n/2}
+            \left( \det \Omega \\right)^{m/2}
+        }
+
+    or, alternatively,
+
+    .. math::
+        \mathcal{T}_{m,n} =
```
lucascolley, line 1551:
```suggestion
    .. math::

        \mathcal{T}_{m,n} = 
```

dschmitz89 (changes requested):
This looks very promising, thank you @johnmdusel . Could you add references to the paper in the docstring?

About tests: how can we evaluate that this is indeed the matrix t distribution? Can we test the PDF method against some external reference? Or is there a special case where this reduces to a distribution we already have implemented?

johnmdusel:
> This looks very promising, thank you @johnmdusel . Could you add references to the paper in the docstring?

Yes, I will add the reference to Gupta and Nagar.

> About tests: how can we evaluate that this is indeed the matrix t distribution? Can we test the PDF method against some external reference? Or is there a special case where this reduces to a distribution we already have implemented?

When I was developing this class, I tested the `rvs` and `pdf` and `logpdf` methods against Mathematica's `MatrixTDistribution`. The `matrix_t_gen` implemented here does agree with the Mathematica version. I can add some or all of that validation to the `TestMatrixT` class, if there's interest.

lucascolley:
it looks like some commits were accidentally included with a rebase

johnmdusel:
> it looks like some commits were accidentally included with a rebase

Apologies for that, and thanks for fixing it.

lucascolley:
> thanks for fixing it.

I don't think that it is fixed: the PR is still showing as 26 commits

dschmitz89:
> > This looks very promising, thank you @johnmdusel . Could you add references to the paper in the docstring?
> 
> Yes, I will add the reference to Gupta and Nagar.
> 

Thanks!

> > About tests: how can we evaluate that this is indeed the matrix t distribution? Can we test the PDF method against some external reference? Or is there a special case where this reduces to a distribution we already have implemented?
> 
> When I was developing this class, I tested the `rvs` and `pdf` and `logpdf` methods against Mathematica's `MatrixTDistribution`. The `matrix_t_gen` implemented here does agree with the Mathematica version. I can add some or all of that validation to the `TestMatrixT` class, if there's interest.

That would be great. I do onot have Mathematica but I think at least @mdhaber had access in case we want to counter check the reference values.

johnmdusel:
> I don't think that it is fixed: the PR is still showing as 26 commits

Having difficulty but still working on removing the unwanted commits from 13 May. I will hold off on updating tests per @dschmitz89's request until they're gone.

lucascolley:
> until they're gone

gone now :) you just want to make sure you have pulled the latest changes before committing anything else

johnmdusel:
I noticed many `DeprecationWarning: invalid escape sequence '\m'`(also `\G`) while the docs were building.

mdhaber:
> I do not have Mathematica...

I can execute code if you can provide it

johnmdusel:
> I can execute code if you can provide it

For the record, here's Mathematica info for my system:

- Version 13.0.0 for Linux x86 (64-bit) (December 3, 2021)

- Release ID 13.0.0.0 (7522564, 2021120311723)

- Patch Level 0

I generated matrix t-distribution samples and their PDF values using these commands

```
mu={{1,2,3},{4,5,6}};
sigma={{1,0.5},{0.5,1}};
omega={{1,0.3,0.2},{0.3,1,0.4},{0.2,0.4,1}};
df=5;
sampleSize=1000000;
SeedRandom[42];
dist=MatrixTDistribution[mu,sigma,omega,alpha];
samples=RandomVariate[dist,sampleSize];
Export["matrixt_samples.csv",Flatten[#]&/@samples,"CSV"];
pdfs=PDF[dist,#]&/@samples;
Export["matrixt_pdfs.csv",pdfs,"CSV"];
```

I saved the first 10,000 values from each CSV into a `.npz` file with fields `"samples"` and `"pdfs`. My unit tests use that `.npz` file. Let me know if I should use the one you generated instead.

rgommers:
> I saved the first 10,000 values from each CSV into a `.npz` file with fields `"samples"` and `"pdfs`. My unit tests use that `.npz` file. Let me know if I should use the one you generated instead.

We can't commit such a large `.npz` file to the repo. Perhaps use it for testing and then remove it, leaving only the code comment about how to reproduce the testing for someone with access to Mathematica.

dschmitz89:
@johnmdusel Sorry if you interpreted my request in a way that we would need that many test cases. I thought of just a handful of reference values computed by Mathematica. What we typically do is put the code for reference value creation  into a comment and then add a few accuracy tests using pytest parametrization.

johnmdusel:
Running `dev.py doc` produces errors that seem unrelated to this PR. See line 55 in attached file.
[doc-build.log](https://github.com/user-attachments/files/20411805/doc-build.log)

melissawm:
@johnmdusel we recently updated the build infrastructure of scipy to use `spin build` instead of dev.py - it looks like main was merged into your branch so you may need to use this command instead of `dev.py build`.

johnmdusel:
@melissawm thank you for the reminder. Confirming that `spin build` `spin test` `spin docs` `spin bench` run successfully locally.

@dschmitz89 please advise if the current state satisfies the requested changes

johnmdusel:
The [failing check above](https://github.com/scipy/scipy/actions/runs/15397216680/job/43325858387?pr=22925) looks to be about the `match=""` argument on [line 242 of test_lsq_linear.py](https://github.com/scipy/scipy/blame/f2d22b2308bc0269e18fe2f27fe6ab17540a4ac9/scipy/optimize/tests/test_lsq_linear.py#L242). 

I can fix this but want to confirm that the intent is to verify that an `AssertionError` is raised, regardless of the message. Can anyone confirm? Thank you!

scipy/stats/_multivariate.py
```diff
@@ -1448,6 +1449,685 @@ def entropy(self):
...
+
+_matt_doc_default_callparams = """\
+mean : array_like, optional
+    Mean of the distribution (default: `None`)
+rowcov : array_like, optional
+    Among-row covariance matrix of the distribution (default: ``1``)
+colcov : array_like, optional
+    Among-column covariance matrix of the distribution (default: ``1```)
+df : scalar, optional
+    Degrees of freedom (default: ```1```)
```
dschmitz89:
```suggestion
colcov : array_like, optional
    Among-column covariance matrix of the distribution (default: ``1``)
df : scalar, optional
    Degrees of freedom (default: ``1``)
```

scipy/stats/_multivariate.py
```diff
@@ -1448,6 +1449,685 @@ def entropy(self):
...
+            raise ValueError("Array `colcov` must be square.")
+        if np.any(colshape == 0):
+            raise ValueError("Array `colcov` has invalid shape.")
+        numcols = colshape[0]
+
+        # Ensure mean and 2nd moments are conformal
+        if mean is not None:
+            if meanshape[0] != numrows:
+                raise ValueError(
+                    "Arrays `mean` and `rowcov` must have the " "same number of rows."
```
dschmitz89: Copy paste error?

scipy/stats/_multivariate.py
```diff
@@ -1448,6 +1449,685 @@ def entropy(self):
...
+                raise ValueError(
+                    "Arrays `mean` and `colcov` must have the "
+                    "same number of columns."
+                )
+        else:
+            mean = np.zeros((numrows, numcols))
+
+        dims = (numrows, numcols)
+
+        if df is None:
```
dschmitz89, line 1743: Note to self: this is the same default as for the `multivariate_t`.

scipy/stats/tests/test_multivariate.py
```diff
@@ -1429,6 +1429,315 @@ def test_samples(self):
...
+        num_rows = 4
+        num_cols = 3
+        df = 5
+        M = np.full((num_rows,num_cols), 0.3)
+        U = 0.5 * np.identity(num_rows) + np.full((num_rows, num_rows), 0.5)
+        V = 0.7 * np.identity(num_cols) + np.full((num_cols, num_cols), 0.3)
+
+        # Nonpositive degrees of freedom
+        assert_raises(ValueError, matrix_t, 0, 1, 1, 0)
+        assert_raises(ValueError, matrix_t, 0, 1, 1, -1)
```
dschmitz89: These and all following `assert_raises` need to check the error message using the `match=` clause from pytest.

scipy/stats/tests/test_multivariate.py
```diff
@@ -1429,6 +1429,315 @@ def test_samples(self):
...
+        assert_equal(matrix_t(mean=M, rowcov=Uv, colcov=Vv, df=df).colcov,
+                     0.1*Ic)
+        assert_equal(matrix_t(mean=M, rowcov=Us, colcov=Vs, df=df).rowcov,
+                     0.2*Ir)
+        assert_equal(matrix_t(mean=M, rowcov=Us, colcov=Vs, df=df).colcov,
+                     0.1*Ic)
+
+    def test_frozen_matrix_t(self):
+        for i in range(1,5):
+            for j in range(1,5):
```
dschmitz89: Can this be done using
`@pytest.mark.parametrize("i", (range(1, 5))`
`@pytest.mark.parametrize("j", (range(1, 5))`

instead of the two loops?

dschmitz89: Also just three test cases should suffice instead of 4 to save some testing time.

scipy/stats/tests/test_multivariate.py
```diff
@@ -1429,6 +1429,315 @@ def test_samples(self):
...
+                separate_logpdf = matrix_t.logpdf(X[i,j], mean=M,
+                                                       rowcov=U, colcov=V,
+                                                       df=df)
+                assert_allclose(separate_logpdf, array_logpdf[i,j], 1E-10)
+
+    @staticmethod
+    def relative_error(vec1: np.ndarray, vec2: np.ndarray):
+        numerator = np.linalg.norm(vec1 - vec2) ** 2
+        denominator = np.linalg.norm(vec1) ** 2 + np.linalg.norm(vec2) ** 2
+        return 2 * numerator / denominator
```
dschmitz89: Why the factor of 2?

scipy/stats/tests/test_multivariate.py
```diff
@@ -1429,6 +1429,315 @@ def test_samples(self):
...
+    def matrix_divergence(mat_true: np.ndarray,
+                            mat_est: np.ndarray) -> float:
+        det_true = np.linalg.det(mat_true)
+        det_est = np.linalg.det(mat_est)
+        if (det_true <= 0) or (det_est <= 0):
+            return np.inf
+        trace_term = np.trace(np.linalg.inv(mat_est) @ mat_true)
+        log_detratio = np.log(det_est / det_true)
+        return (trace_term + log_detratio - len(mat_true)) / 2
+
```
dschmitz89: Is this necessary to compute a norm in matrix space? Seems very involved if it is only needed for testing. Also explicity inverses should be avoided if possible.

dschmitz89 (changes requested):
Sorry for taking a bit longer @johnmdusel. Overall, the implementation looks good. A few comments inline and one more request regarding testing: can we add a test that for a 1D matrix it produces the same output (pdf, entropy) as the `multivariate_t` ? We typically do that to check if two implementations are the same when they should be mathematically.

scipy/stats/_multivariate.py
johnmdusel: Done

scipy/stats/_multivariate.py
johnmdusel: Thanks for catching

scipy/stats/_multivariate.py
johnmdusel: Acknowledged

scipy/stats/tests/test_multivariate.py
johnmdusel: Done

scipy/stats/tests/test_multivariate.py
johnmdusel: Done

scipy/stats/tests/test_multivariate.py
johnmdusel: This was a normalization that I needed in another context but is not important here. I removed it from this PR; thanks for catching it.

scipy/stats/tests/test_multivariate.py
johnmdusel: I streamlined the calculation to use `_PSD`, which also avoids explicitly taking an inverse. This calculation is useful for comparing random variates because it is invariant under scale and the row-spread and column-spread matrices (see below) are [unique to to a positive real constant](https://reference.wolfram.com/language/ref/MatrixTDistribution.html).

johnmdusel:
@dschmitz89 Thank you for requesting the comparison with `multivariate_t`. 
While writing this test I realized that there's a mathematical subtlety in the way the matrix t versus the multivariate t are parameterized. Because of this I'm proposing a change to the parameter names for `matrix_t`. 

I'll explain the issue first:

The matrix t ($p \times m$) is parameterized by two matrices usually denoted $\Sigma (p \times p)$ and $\Omega (m \times m)$. These are raw 2nd order central moments. On the other hand, the multivariate t is parameterized by a covariance matrix--which is the normalized raw 2nd order central moment--usually denoted $U$ and called `shape` in the `multivariate_t` implementation. The matrix normal is also parameterized by covariance matrices $U (p \times p)$ and $V(m \times m)$, called `rowcov` and `colcov` in the `matrix_normal` implementation.

This difference could result in confusion, so I'm going to change the terminology used in `matrix_t` from `rowcov` and `colcov` to `row_spread` and `col_spread`. Gupta and Nagar (2000) on p.134 attribute these terms to a paper "Subjective-probability assessment methods for multivariate-t and matrix-t models" by JM Dickey, AP Dawid, JB Kadane (1986). I haven't found a digital copy, yet. 

It's now a feature that the interface for `matrix_t` doesn't match the ones for `multivariate_t` and `matrix_normal`.

johnmdusel:
@lucascolley -- Sorry but I've made a mess of my history again. It all started when I tried to put another change into [c5df254](https://github.com/scipy/scipy/pull/22925/commits/c5df25484b3bc6207c5b9a15cbc35d1cda29e0af). Can you work your magic trick again?

dschmitz89:
> @dschmitz89 Thank you for requesting the comparison with `multivariate_t`. While writing this test I realized that there's a mathematical subtlety in the way the matrix t versus the multivariate t are parameterized. Because of this I'm proposing a change to the parameter names for `matrix_t`.
> 
> I'll explain the issue first:
> 
> The matrix t ( p × m ) is parameterized by two matrices usually denoted Σ ( p × p ) and Ω ( m × m ) . These are raw 2nd order central moments. On the other hand, the multivariate t is parameterized by a covariance matrix--which is the normalized raw 2nd order central moment--usually denoted U and called `shape` in the `multivariate_t` implementation. The matrix normal is also parameterized by covariance matrices U ( p × p ) and V ( m × m ) , called `rowcov` and `colcov` in the `matrix_normal` implementation.
> 
> This difference could result in confusion, so I'm going to change the terminology used in `matrix_t` from `rowcov` and `colcov` to `row_spread` and `col_spread`. Gupta and Nagar (2000) on p.134 attribute these terms to a paper "Subjective-probability assessment methods for multivariate-t and matrix-t models" by JM Dickey, AP Dawid, JB Kadane (1986). I haven't found a digital copy, yet.
> 
> It's now a feature that the interface for `matrix_t` doesn't match the ones for `multivariate_t` and `matrix_normal`.

Puh, I have never used these distributions myself, so I do not have a strong opinion. Are you saying that the old naming `rowcov` would be wrong for `matrix_t` while it is correct for `matrix_normal`?

johnmdusel:
@dschmitz89  Like you said, `rowcov` and `colcov` are correct for `matrix_normal`. It's not wrong to use them for `matrix_t`, either.  [Mathematica's documentation](https://reference.wolfram.com/language/ref/MatrixTDistribution.html) calls them "row/column covariance matrix". 

IMO calling them "spread matrices" is more descriptive, especially in the context of `matrix_normal` and `multivariate_t`.

lucascolley:
> @lucascolley -- Sorry but I've made a mess of my history again. It all started when I tried to put another change into [c5df254](https://github.com/scipy/scipy/pull/22925/commits/c5df25484b3bc6207c5b9a15cbc35d1cda29e0af). Can you work your magic trick again?

🎩🪄

(feel free to just make regular commits and pushes from now on, I don't think there is a need to rebase and force-push)

scipy/stats/tests/test_multivariate.py
```diff
@@ -1429,6 +1429,427 @@ def test_samples(self):
...
+        atol = 1e-10
+
+        samples_m = np.array(
+            [
+                [
+                    [0.6399716994253741, 2.171718671534955, 2.5758260933527706],
+                    [4.031082477912233, 5.0216809585266375, 6.268126154787008],
+                ],
+                [
+                    [1.1648428842062322, 2.5262970999930445, 3.7813752298650685],
```
dschmitz89: Does Mathematica have a function to convert its arbitrary precision numbers to common double precision? These numbers have up to 17 digits but double precision typically only yields up to 15 significant digits.

scipy/stats/tests/test_multivariate.py
```diff
@@ -1429,6 +1429,427 @@ def test_samples(self):
...
+        Gupta and Nagar (2000) p.133f
+        When the number of rows or the number of columns equals 1 the
+        matrix t reduces to the multivariate t. But, the matrix t
+        is parameterized by raw 2nd moments whereas the multivariate t is
+        parameterized by a covariance (raw 2nd central moment normalized by df).
+        We can see the difference by comparing the author's notation
+            $t_p(n, \omega, \mathbf{\mu}, \Sigma)$
+        for a matrix t with a single column
+        to the formula (4.1.2) for the PDF of the multivariate t.
+        """
```
dschmitz89: Thanks for the explanation :). Could you parametrize the test so that it also covers both cases (1 column and 1 row)?

scipy/stats/tests/test_multivariate.py
johnmdusel: Done (d8b5b53ea)

scipy/stats/tests/test_multivariate.py
johnmdusel: Always glad to write about math :). Done (af59a7ae5).

dschmitz89: > Always glad to write about math :).

That's the spirit :).

scipy/stats/_multivariate.py
```diff
@@ -1448,6 +1449,609 @@ def entropy(self):
...
+    .. [1] Gupta, A.K., & Nagar, D.K. (2000). Matrix Variate Distributions (1st ed.).
+           Chapman and Hall/CRC.
+
+    Examples
+    --------
+
+    >>> import numpy as np
+    >>> from scipy.stats import matrix_t
+    >>> M = np.arange(6).reshape(3,2); M
+    array([[0, 1],
```
dschmitz89: Typically, we would put `M` in its own line. See for example here for `uniform_direction`:

image

To have a consistent style throughout our docs, could you update this and the following lines?

scipy/stats/_multivariate.py
```diff
@@ -1448,6 +1449,609 @@ def entropy(self):
...
+            if 0 in meanshape:
+                raise ValueError("Array `mean` has invalid shape.")
+            if len(meanshape) != 2:
+                raise ValueError("Array `mean` must be 2D.")
+
+        # Process row-wise spread
+        row_spread = np.asarray(row_spread, dtype=float)
+        if row_spread.ndim == 0:
+            if np.isnan(row_spread).any():
+                row_spread = np.identity(1)
```
dschmitz89: Didn't notice this before, so sorry about the new review comment.

Not sure we have a precedent for what to do with nans but silently changing it to an identity matrix does not "feel" right and will be a cause for silent fails. I would just remove it.

scipy/stats/_multivariate.py
```diff
@@ -1448,6 +1449,609 @@ def entropy(self):
...
+            \right)
+            \pi^{mn / 2}
+            \left( \det \Sigma \right)^{n/2}
+            \left( \det \Omega \right)^{m/2}
+        }
+
+    where :math:`\mathrm{M}` is the mean,
+    :math:`\Sigma` is the row-wise spread matrix,
+    :math:`\Omega` is the column-wise matrix,
+    and :math:`\mathrm{df}` is the degrees of freedom.
```
dschmitz89: Could you also add an explanation of $\Gamma_n$ as the multivariate gamma function?

scipy/stats/_multivariate.py
```diff
@@ -1448,6 +1449,609 @@ def entropy(self):
...
+
+    When :math:`\mathrm{df} = 1` this distribution is known as the matrix
+    variate Cauchy.
+
+    .. versionadded:: 1.16.0
```
dschmitz89:
```suggestion
    .. versionadded:: 1.17.0
```

scipy/stats/_multivariate.py
```diff
@@ -1448,6 +1449,609 @@ def entropy(self):
...
+        std_norm = random_state.standard_normal(
+            size=(dims[1], size, dims[0])
+        ).transpose(1, 2, 0)
+        if dims[0] <= dims[1]:
+            rowchol = _cholesky_invwishart_rvs(df, row_spread, size, random_state)
+            colchol = scipy.linalg.cholesky(col_spread, lower=True)[np.newaxis, ...]
+        else:
+            rowchol = scipy.linalg.cholesky(row_spread, lower=True)[np.newaxis, ...]
+            colchol = _cholesky_invwishart_rvs(df, col_spread, size, random_state)
+        t_raw = np.einsum("ijp,ipq,ikq->ijk", rowchol, std_norm, colchol, optimize=True)
```
dschmitz89, line 1984: Could you add some explanatory comments for this code or a reference?

dschmitz89 (changes requested):
A few more comments mostly about the interface @johnmdusel. Would you have time to also have a look @mdhaber (especially the mathematica reference)?

scipy/stats/_multivariate.py
johnmdusel: Done (7dc12e61ff8ffab7654daa83f1c18d2aeb729170) 

Originally, I was following the style o fthe [`matrix_normal` example](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.matrix_normal.html#scipy.stats.matrix_normal)

```
>>> M = np.arange(6).reshape(3,2); M
array([[0, 1],
       [2, 3],
       [4, 5]])
>>> U = np.diag([1,2,3]); U
array([[1, 0, 0],
       [0, 2, 0],
       [0, 0, 3]])
>>> V = 0.3*np.identity(2); V
array([[ 0.3,  0. ],
       [ 0. ,  0.3]])
>>> X = M + 0.1; X
array([[ 0.1,  1.1],
       [ 2.1,  3.1],
       [ 4.1,  5.1]])
>>> matrix_normal.pdf(X, mean=M, rowcov=U, colcov=V)
0.023410202050005054
```

scipy/stats/_multivariate.py
johnmdusel: Thanks for catching this. 

Done (0dceecb0620fdbd95141213ac54440071de317b6)

scipy/stats/_multivariate.py
johnmdusel: Done (7dc12e61ff8ffab7654daa83f1c18d2aeb729170)

scipy/stats/_multivariate.py
johnmdusel: Done (7dc12e61ff8ffab7654daa83f1c18d2aeb729170)

scipy/stats/_multivariate.py
johnmdusel: Done (550fe3d78638059fb8a3437b07b47228ad37dd10)

mdhaber:
> Would you have time to also have a look @mdhaber (especially the mathematica reference)?

I'm limited to using Mathematica version "14.2.1 for Linux x86 (64-bit) (March 16, 2025)". When I execute that code, I don't get the sample sample or pdf values.

dschmitz89:
> > Would you have time to also have a look @mdhaber (especially the mathematica reference)?
> 
> I'm limited to using Mathematica version "14.2.1 for Linux x86 (64-bit) (March 16, 2025)". When I execute that code, I don't get the sample sample or pdf values.

Hm, unfortunate. Can we do a minimal test with a fixed set of matrices to test the pdf values if we cannot rely on the random variates? I do not expect differences in the first few digits between different mathematica versions. @johnmdusel  @mdhaber

johnmdusel:
> Can we do a minimal test with a fixed set of matrices to test the pdf values if we cannot rely on the random variates?

@dschmitz89 I could set up a minimal Julia script to generate random variates and compute PDF values. Would that be alright, as an alternative to Mathematica? I would provide a Dockerfile and a `.jl` file and copy-paste the outputs as is currently done in the unit test.

mdhaber:
Oops that was a misclick. I'll see if I can just verify the PDfF values @johnmdusel  generated.

johnmdusel:
@dschmitz89 @mdhaber I added a unit test that checks PDF values against Julia. The source in the test's docstring and this [small repo](https://github.com/johnmdusel/matrix-t-julia). All tests pass locally.

scipy/stats/tests/test_multivariate.py
```diff
@@ -1429,6 +1429,536 @@ def test_samples(self):
...
+                ],
+                [
+                    [1.045187318995198, 1.645467189679729, 3.284396214544507],
+                    [3.648493466445393, 5.004212508553601, 6.301624351328048],
+                ],
+            ]
+        )
+
+        pdfs_m = np.array(
+            [
```
mdhaber, line 1869: I was able to confirm agreement with Mathematica 14.2.1 for Linux x86 (64-bit) (March 16, 2025) for these PDF values (given the sample of matrices above).

dschmitz89 (approved):
I am happy with this PR and the pdf values were also confirmed against an external source (Mathematica). There is one last stylistic nit regarding tests: we typically use relative tolerances, not absolute ones. Will wait for other reviewers before merging.

johnmdusel:
@rgommers The tests against [Julia](https://github.com/johnmdusel/scipy/blob/d861da77ec67b3bd2d46fc31f1955a5a3b9bc393/scipy/stats/tests/test_multivariate.py#L1703) and [Mathematica](https://github.com/johnmdusel/scipy/blob/d861da77ec67b3bd2d46fc31f1955a5a3b9bc393/scipy/stats/tests/test_multivariate.py#L1800) are now based on a handful of reference values, instead of a file.

scipy/stats/tests/test_multivariate.py
```diff
@@ -1892,7 +1892,7 @@ def test_samples(self):
         U = 0.5 * np.identity(num_rows) + np.full((num_rows, num_rows), 0.5)
         V = 0.7 * np.identity(num_cols) + np.full((num_cols, num_cols), 0.3)
         N = 10**4
-        atol = 1e-1
+        rtol = 1e-1
```
dschmitz89: A relative tolerance of 10% is quite high, can we lower it?

johnmdusel: I set the relative tolerance to 1%, trading off on the sample size.

johnmdusel:
@dschmitz89 -- questions about the failing CI checks above

- [Linux tests](https://github.com/scipy/scipy/actions/runs/16263981032/job/46069887969?pr=22925): The `test_samples` test took too long; it takes more samples for the test to pass with a lower relative tolerance. How would you like me to manage this tradeoff? With 10k samples it will pass with a 5% relative tolerance; with 100k it will pass with 2.5% but will take > 1 second.

- [Windows tests](https://github.com/scipy/scipy/actions/runs/16263981027/job/46069887230?pr=22925): This seems to be related to HiGHS; can I assume it'll be fixed in some other PR?

dschmitz89:
> @dschmitz89 -- questions about the failing CI checks above
> 
> - [Linux tests](https://github.com/scipy/scipy/actions/runs/16263981032/job/46069887969?pr=22925): The `test_samples` test took too long; it takes more samples for the test to pass with a lower relative tolerance. How would you like me to manage this tradeoff? With 10k samples it will pass with a 5% relative tolerance; with 100k it will pass with 2.5% but will take > 1 second.
> 

Let's go with the 5% tolerance for simplicity. Sry about the churn.

> - [Windows tests](https://github.com/scipy/scipy/actions/runs/16263981027/job/46069887230?pr=22925): This seems to be related to HiGHS; can I assume it'll be fixed in some other PR?

That one is already fixed in main.

johnmdusel:
> > @dschmitz89 -- questions about the failing CI checks above
> > 
> > * [Linux tests](https://github.com/scipy/scipy/actions/runs/16263981032/job/46069887969?pr=22925): The `test_samples` test took too long; it takes more samples for the test to pass with a lower relative tolerance. How would you like me to manage this tradeoff? With 10k samples it will pass with a 5% relative tolerance; with 100k it will pass with 2.5% but will take > 1 second.
> 
> Let's go with the 5% tolerance for simplicity. Sry about the churn.

No problem! I made the change.

johnmdusel:
> @rgommers The tests against [Julia](https://github.com/johnmdusel/scipy/blob/d861da77ec67b3bd2d46fc31f1955a5a3b9bc393/scipy/stats/tests/test_multivariate.py#L1703) and [Mathematica](https://github.com/johnmdusel/scipy/blob/d861da77ec67b3bd2d46fc31f1955a5a3b9bc393/scipy/stats/tests/test_multivariate.py#L1800) are now based on a handful of reference values, instead of a file.

@rgommers would you be able to approve if no further changes are required? On the other hand, I'm happy to make further  changes as needed. Thank you very much!

dschmitz89:
This PR fell off my radar. I will merge it next week unless there are more comments or unexpected CI failures after merging main.

dschmitz89:
Thanks for your first (and large) contribution @johnmdusel !

johnmdusel:
Thank you, @dschmitz89 and @lucascolley! It's been a pleasure.

PR #22926, opened 2025-05-03
MAINT: spatial.cKDTree: remove software prefetching and software branch-prediction

### LINKED ISSUE #22904: BUG: Wrong use of `__builtin_prefetch()`

sturlamolden:
#### Reference issue
Closes gh-22904

See also gh-22922

Removes prefetching and likely/unlikely statements in the code. It was useful in 2015 but not in 2025. Also it has been broken and defunct since a coding error in SciPy 1.3 (due to https://github.com/scipy/scipy/commit/6975656966bd526570af5616096487e10921f163). It complicates the code so it is time to remove it.

### DIFF
--- a/scipy/spatial/_ckdtree.pyx
+++ b/scipy/spatial/_ckdtree.pyx
@@ -31,10 +31,6 @@ cdef extern from "<limits.h>":
 
 __all__ = ['cKDTree']
 
-cdef extern from *:
-    int NPY_LIKELY(int)
-    int NPY_UNLIKELY(int)
-
 
 # C++ implementations
 # ===================
@@ -188,7 +184,7 @@ cdef class coo_entries:
         _dtype = [('i',np.intp),('j',np.intp),('v',np.float64)]
         res_dtype = np.dtype(_dtype, align = True)
         n = <np.intp_t> self.buf.size()
-        if NPY_LIKELY(n > 0):
+        if (n > 0):
             pr = self.buf.data()
             uintptr = <np.uintp_t> (<void*> pr)
             dtype = np.dtype(np.uint8)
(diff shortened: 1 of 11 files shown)


nickodell (approved):
This looks reasonable.

Benchmarking shows no meaningful changes in performance, which is not surprising, as many of these branches are probably fairly predictable, and the prefetch macro has compiled to nothing since 1.3.

As this simplifies the code, and has no performance cost, I approve.

tylerjereddy (approved):
I just read the full diff and agree with comments/approval above, thanks both.

PR #22928, opened 2025-05-03
BUG: spatial.cKDTree.query: fix allocator alignment bugs

sturlamolden:
#### Reference issue
See gh-22922

```cKDTree.query``` has a custom O(1) memory allocator because ```new[]``` or ```std::malloc``` are too slow (yes, really). The allocator assumes a page is 4K and a cache line is 64 bytes. 

- The memory returned by the allocator should be aligned according to page and cache line boundaries, for faster access and to avoid false sharing in parallel queries. The logic has been broken for many years. 

- Also there is a bug that results in unnecessary memory allocation if the data already fits within the alignment boundaries.

- There is also a bug (line 186) that can cause the allocator to compute a memory address that might not exist, which teoretically can cause undefined behaviour in very rare occations. In this PR the operation is moved to line 248 where it is safe.

### additional information
Apple Silicon uses 16K pages and 128 byte cache lines. This PR does not cater for this ATM.

### CODE REVIEW

**Comment by sturlamolden** [comment]
> are you looking to remove the prefetching in this PR or a separate one?

Prefetching is removed in gh-22926.

**Comment by lucascolley** [comment]
thanks, I missed that

**Review by tylerjereddy — COMMENTED** [review]
Benchmarking via `asv continuous -E virtualenv -e -b ".*time_query.*" main better-tuning-for-Apple-silicon` I saw a slight drop in performance on M3 Max Mac and no change on `x86_64 Linux (i9-13900K)`.

M3 Max Mac:

```
| Change   | Before [0242738d] <main>   | After [e755ce31] <better-tuning-for-Apple-silicon>   |   Ratio | Benchmark (Parameter)                                               |
|----------|----------------------------|------------------------------------------------------|---------|---------------------------------------------------------------------|
| +        | 26.1±0.9ms                 | 31.2±0.7ms                                           |    1.19 | spatial.Radius.time_query_pairs((3, 1000, 30), inf, 0.5, None, 128) |

PERFORMANCE DECREASED.
```

If the argument is that this change is only effective for much larger numbers of coordinates I/we could try repeating the benchmarks with temporarily increased sizes I suppose.

**Comment by sturlamolden** [comment]
> | +        | 26.1±0.9ms                 | 31.2±0.7ms                                           |    1.19 | spatial.Radius.time_query_pairs((3, 1000, 30), inf, 0.5, None, 128) |

The code  for```cKDTree.query_pairs``` is not changed and this PR should not have any effect on it. 


> If the argument is that this change is only effective for much larger numbers of coordinates

No, that was prefetching, not this PR.

**Comment by tylerjereddy** [comment]
maybe clarify if you think we need to add a new `asv` benchmark that can capture the improvement -- I can try to draft it if you want

**Comment by tylerjereddy** [comment]
In the absence of another person verifying the performance improvement in some way, preferably via `asv`, I'm hesitant to merge this just yet. I think I'll bump the milestone so that this isn't counted as release/branching "blocking" for `1.16.0`.

Just ping me if this gets updated and the demonstration of performance improvement is clearer.

**Comment by nickodell** [comment]
@sturlamolden I've been meaning to get back to you about this PR; apologies for the delay. I disagree with some aspects of this PR.

- For the cache alignment changes, I don’t think it is necessary to avoid sharing cache lines between multiple nodeinfo structs. It isn’t needed to avoid false sharing, as each thread has its own nodeinfo pool. For alignment, it is only needed to align the pointer to the minimum alignment of the struct, which is alignof(nodeinfo), or 8.
- The code which aligns the returned memory to a cache line boundary only works if the arena was originally allocated at a cache line boundary in the first place. The memory allocator is only guaranteed to align the allocation to a 16 byte boundary when allocating a `new char[N]`.
- The code which rounds the allocation up to the nearest multiple of the cache line has a bug in it that makes it use extra memory: if the allocation is already an exact multiple of the cache line size, then the code still rounds the allocation up. This wastes 64 / 128 bytes depending on how large the cache line is.
- For the page size change, I don’t see it as necessary to make sure that the area is an exact multiple of the page size. My understanding is that if the page’s size is below the mmap threshold (which is 128 KiB on linux glibc/OS X libmalloc) then the memory is allocated as if it were a large, non-page sized allocation, which still involves searching through free lists. I see the arena size as an arbitrary parameter, where tuning the parameter higher results in larger allocations but fewer of them.

I also did some benchmarking. I found that it regressed on Apple Silicon for some sizes. I couldn’t find a case that had a 30% performance improvement, though some benchmarks do improve.

Here's a benchmark, spatial.QueryUnbalanced.time_query, that regresses on these changes. Note that I have modified the benchmark a little, to make it larger. You can see the benchmark changes in njo-ckdtree-base.

Legend:

- njo-ckdtree-change1 is the prefetch change
- njo-ckdtree-change2 is the OS X tuning
- njo-ckdtree-change3 is the prefetch change + OS X tuning
- njo-ckdtree-change4 is the prefetch change + OS X tuning + reducing alignment to 8
- njo-ckdtree-base is the commit that these changes originally branched off of, for comparison purposes

(Sorry about the order these are graphed in. I couldn't figure out how to change it.)

In this graph, lower is better.

asv benchmark comparison

If you'd like to look at any of these branches, they are available at my repo: https://github.com/nickodell/scipy/

In summary, in order to merge this, I would like to see

1. More detail about the benchmark that improves. A regression on certain benchmarks could be justifiable, depending on what the gain is elsewhere.
2. Avoiding wasting memory if the allocation is already cache-sized aligned.

**Comment by tylerjereddy** [comment]
It seems like neither Nick nor I are ready to merge this yet, and it has been inactive for a while, so I'll bump the milestone to `1.18.0`. We have around 50 milestoned PRs with 2 weeks or so until branching, and I don't think this one is critical.

I believe Nick and I both saw performance regressions in some cases, though Nick obviously did a deeper dive.

I should be clear that, as Sturla obviously knows as one of the main devs, KDTree is very important and improvements are most welcome--I've often used it in my own research for lightning fast distance assessments in simulations.

**Comment by nickodell** [comment]
Should this PR be closed? We haven't observed an improvement in the benchmark, and there hasn't been progress on creating a benchmark where this shows a positive improvement.

I propose that we close this in a week if there's no movement.

**Comment by sturlamolden** [comment]
Just close this.

**Comment by sturlamolden** [comment]
Actually, there is still a bug we need to squash, as pointed out by @nickodell

**Comment by sturlamolden** [comment]
@nickodell 

Oh, I see... It should always round up. Because in the chunk returned by ```new char [n]``` there should be room for an aligned ```nodeinfo``` struct, even ```new``` only gives us 16 byte alignment. It seems a very long time ago the logic of the ```allocate``` method I wrote has been messed up, so now this is just bogus.

**Comment by sturlamolden** [comment]
It is all messed up since the introduction of support for toroid universes. The ```nodeinfo``` struct is messed up too. 

This is in need of a serious bugfix.

**Comment by sturlamolden** [comment]
Now it looks better. The four fails are unrelated I think.

**Comment by tylerjereddy** [comment]
> This is in need of a serious bugfix.

Anything we can hit/reproduce with a test of a public function?

**Comment by nickodell** [comment]
1. Can you update the first comment explaining the PR? It's quite out of date.
2. I would rather leave the Apple specific tuning out of the PR. It's not clear to me that it's necessary, even on Apple hardware.

**Comment by sturlamolden** [comment]
> > This is in need of a serious bugfix.
> 
> Anything we can hit/reproduce with a test of a public function?

That is a good question. A regression test would be nice. I will have to think about this.

**Comment by sturlamolden** [comment]
> 1. Can you update the first comment explaining the PR? It's quite out of date.

Will do.

> 2. I would rather leave the Apple specific tuning out of the PR. It's not clear to me that it's necessary, even on Apple hardware.

Sure.

**Comment by sturlamolden** [comment]
I am trying to checkout this branch locally to squash, but I just get...

```
error: Entry 'scipy/_lib/array_api_compat/.github/workflows/array-api-tests-dask.yml' not uptodate. Cannot merge.
error: Submodule 'scipy/_lib/array_api_compat' could not be updated.
error: Entry 'scipy/_lib/array_api_extra/.all-contributorsrc' not uptodate. Cannot merge.
error: Submodule 'scipy/_lib/array_api_extra' could not be updated.
error: Entry 'scipy/_lib/cobyqa/.github/actions/spelling/expect.txt' not uptodate. Cannot merge.
error: Submodule 'scipy/_lib/cobyqa' could not be updated.
error: Entry 'scipy/sparse/linalg/_propack/PROPACK/Changelog' not uptodate. Cannot merge.
error: Submodule 'scipy/sparse/linalg/_propack/PROPACK' could not be updated.
error: Entry 'subprojects/boost_math/math/.drone.star' not uptodate. Cannot merge.
error: Submodule 'subprojects/boost_math/math' could not be updated.
error: Entry 'subprojects/highs/.bazelrc' not uptodate. Cannot merge.
error: Submodule 'subprojects/highs' could not be updated.
error: Entry 'subprojects/xsf/README.md' not uptodate. Cannot merge.
error: Submodule 'subprojects/xsf' could not be updated.
error: Cannot update submodule:
	scipy/_lib/array_api_compat
	scipy/_lib/array_api_extra
	scipy/_lib/cobyqa
	scipy/sparse/linalg/_propack/PROPACK
	subprojects/boost_math/math
	subprojects/highs
	subprojects/xsf

Aborting
```

I might have to ask for help squashing or close this PR and start all over again with a fresh fork of scipy.

sturlamolden:
I am closing this PR because my local repo is fubar. It is not synced with scipy/scipy anymore and I just get errors. I cannot get anything done. Because I am an imbecille with git I will just have to delete it and start with a fresh fork of scipy. Sorry for this inconvinience.

I will open a new PR to supercede this.

sturlamolden:
Superceded by gh-24269.

PR #22934, opened 2025-05-05
MAINT: optimize.zeros: fix error message

ChVeen:
#### Reference issue
No issue involved.

This tiny PR fixes the error message in `scipy/optimize/zeros.c` at
```
if (iter < 0) {
	PyErr_SetString(PyExc_ValueError, "maxiter should be > 0");
	return NULL;
}
```
Only negative `iter` are disallowed which means that `iter == 0` is allowed which leads to something like `iter must be >= 0`.

### DIFF
--- a/scipy/optimize/zeros.c
+++ b/scipy/optimize/zeros.c
@@ -96,7 +96,7 @@ call_solver(solver_type solver, PyObject *self, PyObject *args)
         return NULL;
     }
     if (iter < 0) {
-        PyErr_SetString(PyExc_ValueError, "iter must be >= 0");
+        PyErr_SetString(PyExc_ValueError, "maxiter must be >= 0");
         return NULL;
     }
 
--- a/scipy/optimize/zeros.c
+++ b/scipy/optimize/zeros.c
@@ -96,7 +96,7 @@ call_solver(solver_type solver, PyObject *self, PyObject *args)
         return NULL;
     }
     if (iter < 0) {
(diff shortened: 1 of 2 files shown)


scipy/optimize/zeros.c
```diff
@@ -96,7 +96,7 @@ call_solver(solver_type solver, PyObject *self, PyObject *args)
         return NULL;
     }
     if (iter < 0) {
-        PyErr_SetString(PyExc_ValueError, "maxiter should be > 0");
+        PyErr_SetString(PyExc_ValueError, "iter must be >= 0");
```
j-bowhay, line 99: I think this should stay as maxiter since this is the name on the python side
```suggestion
        PyErr_SetString(PyExc_ValueError, "maxiter must be >= 0");
```

ChVeen: Okay, changed back :-)

tylerjereddy (commented):
Sample test that fails before/passes after:

```diff
--- a/scipy/optimize/tests/test_zeros.py
+++ b/scipy/optimize/tests/test_zeros.py
@@ -190,6 +190,10 @@ class TestBracketMethods(TestScalarRootFinders):
         assert r.converged
         assert_allclose(root, 1.0, atol=self.xtol, rtol=self.rtol)
 
+    def test_gh_22934(self):
+        with pytest.raises(ValueError, match="maxiter must be >= 0"):
+            zeros.brentq(lambda x: x**2 - 1, -2, 0, maxiter=-1)
+
```

PR #22935, opened 2025-05-05
ENH: Handle missing attrs `.dtype` or `.dtype.kind` in sparse indexer

jggatter:
#### Reference issue

None found

Here, we enable the passing of certain array-like objects such as a `torch.Tensor` for indexing sparse matrices. Currently, an object that has `ndim` such as `torch.Tensor` does not get converted by an `np.asanyarray` call. As a result, an exception occurs when the code attempts to access its non-existent `.dtype.kind` (it has `.dtype` but this lacks a `.kind` attribute unlike for a numpy array).

Notably, I did not include handling to detect a boolean `torch.Tensor`, but I can take a stab at that if desired (without importing `torch` of course). Pending feedback I'm also open to making other changes or improvements.

#### Additional information

Reproduction of error:
1. Use a CUDA environment with NVIDIA drivers installed
2. `pip install scanpy torch` 
3. Run the below code

```python
import anndata
import numpy as np
import scanpy as sc
import torch


# Initialize CUDA
print("Emptying CUDA cache...")
torch.cuda.empty_cache()
print("Initializing CUDA...")
torch.cuda.init()
# Set float32 matmul precision for better performance on NVIDIA GPUs
print("Setting float32 matmul precision to medium...")
torch.set_float32_matmul_precision("medium")

# Test CUDA initialization
print("Testing CUDA via matrix multiplication of random tensors...")
a = torch.rand((100, 100), device="cuda")
b = torch.rand((100, 100), device="cuda")
_ = torch.matmul(a, b)
print("Finished testing CUDA initialization.")

print("Loading small AnnData dataset...")
adata = sc.datasets.pbmc3k()
x = adata.X

# Subsample the data
indices = torch.arange(x.shape[0])
subsampled_indices = indices.split(512)
for batch in subsampled_indices:
    x_batch = x[batch, :]
    # ...

```
Yields exception:
```
  File "/home/ubuntu/Projects/cyrapids/scarches_tensor_index.py", line 31, in <module>
    x_batch = x[batch, :]
              ~^^^^^^^^^^
  File "/opt/pytorch/lib/python3.12/site-packages/scipy/sparse/_index.py", line 30, in __getitem__
    index, new_shape = self._validate_indices(key)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/pytorch/lib/python3.12/site-packages/scipy/sparse/_index.py", line 234, in _validate_indices
    elif (ix := _compatible_boolean_index(idx, self.ndim)) is not None:
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/pytorch/lib/python3.12/site-packages/scipy/sparse/_index.py", line 442, in _compatible_boolean_index
    if idx.dtype.kind == 'b':
       ^^^^^^^^^^^^^^
AttributeError: 'torch.dtype' object has no attribute 'kind'
```

### DIFF
--- a/scipy/sparse/_index.py
+++ b/scipy/sparse/_index.py
@@ -231,7 +231,7 @@ class IndexMixin:
             elif isinstance(idx, slice) or isintlike(idx):
                 index_1st.append(idx)
                 prelim_ndim += 1
-            elif (ix := _compatible_boolean_index(idx, self.ndim)) is not None:
+            elif (ix := _get_compatible_boolean_index(idx, self.ndim)) is not None:
                 index_1st.append(ix)
                 prelim_ndim += ix.ndim
             elif issparse(idx):
@@ -420,8 +420,8 @@ class IndexMixin:
         self._set_arrayXarray(row, col, x)
 
 
-def _compatible_boolean_index(idx, desired_ndim):
-    """Check for boolean array or array-like. peek before asarray for array-like"""
+def _get_compatible_boolean_index(idx, desired_ndim):
+    """Check for compatible boolean array-like value, converting to one if necessary."""
     # use attribute ndim to indicate a compatible array and check dtype
     # if not, look at 1st element as quick rejection of bool, else slower asanyarray
     if not hasattr(idx, 'ndim'):
@@ -439,6 +439,7 @@ def _compatible_boolean_index(idx, desired_ndim):
         # since first is boolean, construct array and check all elements
         idx = np.asanyarray(idx)
 
-    if idx.dtype.kind == 'b':
-        return idx
+    if hasattr(idx, 'dtype'):
+        if hasattr(idx.dtype, 'kind') and idx.dtype.kind == 'b':
+            return idx
     return None


jggatter:
Apologies that I did not have time to check unit tests locally, will do so tomorrow if possible!

jggatter:
Updating to check against scipy 1.15.3

tylerjereddy (commented):
Can you clarify if you're trying to take a SciPy sparse array/matrix, and index it with a `torch` tensor? The reproducing example seems to involve a number of libraries, so is perhaps a bit indirect.

I think my intrepretation may be correct, because I can reproduce your error with a simpler reproducer:

```python
import numpy as np
import scipy
import torch

indices = torch.arange(2)
rng = np.random.default_rng()
S = scipy.sparse.random(3, 4, density=0.25, rng=rng).tocsr()
S[indices]
```

I wonder where support for this kind of thing would stand in the context of the array API standard. For `sparse` data structures, I believe array API support is still in flux per https://github.com/data-apis/array-api/issues/840 for example.

On the other hand, if the data underneath a SciPy `sparse` matrix/array is NumPy-based, would we want to guarantee indexing compatibility with various other array types? I might have guessed we'd rather aim to support indexing with a matching array type.

Maybe @crusaderky or @steppi will have thoughts on where this kind of thing would stand on the array API community side, and perhaps @dschult knows of the "sparse team" plans here.

crusaderky:
> I wonder where support for this kind of thing would stand in the context of the array API standard.

The standard does not support inter-operability between libraries, with a handful of explicit exceptions (dlpack and buffer interfaces). Whether `scipy.sparse` (which is not Array API compliant) wants to implement bespoke support for this or that library should be an internal discussion to the scipy team.

> On the other hand, if the data underneath a SciPy `sparse` matrix/array is NumPy-based, would we want to guarantee indexing compatibility with various other array types?

It boils down to how much Cython/C/C++ `scipy.sparse` has under the hood, and whether that can be avoided in favour of Array API agnostic(ish) algorithms.

ev-br:
Note that `dtype.kind` attribute is not in the array API spec, and is unlikely to be ever added to the spec. So ISTM scanpy is not prepared to handle non-numpy arrays. 

For array API compatible sparse arrays, you might want to take a look at pydata/sparse. 

For scipy.sparse compatible sparse arrays on cuda, you might want to take a look at cupy/cupyx.scipy.sparse. if you've cycles to spare, I guess bringing them to feature parity with recent scipy.sparse enhancements would be welcome. 

None of these is scipy related though, and none of these immediately helps the OP use case which is at least two layers up the stack.

PR #22936, opened 2025-05-06
TST: sparse: more tests for sparse `has_canonical_format` flag

dschult:
This PR adds some tests to `scipy/sparse/tests/test_base.py` that illuminate the power and show a few minor limits of our attribute flag `has_canonical_format` for sparse formats CSR, CSC, BSR and COO.  Note that this meaning of canonical refers to 1) sorted indices and 2) no duplicate entries.

Summary: Every use of CSR and BSR formats for element-wise binary operations in the test suite of SciPy, scikit-learn, scikit-image, networkx and cvxpy is correctly indicated by the attribute `has_canonical_format`. That suggests that we could reduce or remove the check of canonical format from the binary operations like multiply, add, minimum, eq, gt, etc. 

There are still some ways that users could mess up the integrity of the attribute -- though they are hard to arrange without at least wondering if it would work. One is by manually setting the `has_canonical_format` to `True` when the array is not in canonical format. The other is by manually changing the index arrays without resetting `has_canonical_format`. Other, more typical interactions via the API are safe: changing the array via assignment statements or sorting, reshaping, resizing. They all maintain the canonical nature of CSR, CSC, BSR and COO. Verification was explored in the above mentioned packages using [code in a branch of my forked repo](https://github.com/scipy/scipy/compare/main...dschult:scipy:canonical_track). 

The tests here codify cases where `has_canonical_format` correctly tracks the canonical nature. They also demonstrate the two found cases where it doesn't correctly track changes: when a user manually changes the attribute, when a user manually changes the index arrays and fails to reset the attribute. In the latter case, if a user is unsure whether the result is in canonical form, they should set the attribute to `False`. It will get reset to `True` when needed -- or they could trigger a reset by setting it to `False` followed by calling `A.sum_duplicates()`.

Adding tests is easy -- I suspect we should also **update docs** and perhaps change code to **reduce or remove checking** the validity of the `has_canonical_format`. By trusting the flag, we could reduce each binary operation by two function calls to `csr_has_canonical_format`.  The downside of the code change could be a user changing the index arrays directly while leaving the attribute set as `True`. The binary operation in those cases can be surprising: sometimes it gives the correct result in non-canonical form, and sometimes it returns wrong results. I have not been able to create a crash/segfault but I've only tried simple things.  I think the question is whether we can improve the docs for `has_canonical_format` and the element-wise binary operation functions to the point where people will not manually update the index arrays without resetting the flag.  :)

### DIFF
--- a/scipy/sparse/tests/test_base.py
+++ b/scipy/sparse/tests/test_base.py
@@ -4286,31 +4286,70 @@ class TestCSR(sparse_test_class()):
     def test_has_canonical_format(self):
         "Ensure has_canonical_format memoizes state for sum_duplicates"
 
-        M = self.csr_container((np.array([2]), np.array([0]), np.array([0, 1])))
+        info_no_dups = (np.array([2]), np.array([0]), np.array([0, 1]))
+        info_with_dups = (np.array([1, 1]), np.array([0, 0]), np.array([0, 2]))
+
+        M = self.csr_container(info_no_dups)
         assert_equal(True, M.has_canonical_format)
 
-        indices = np.array([0, 0])  # contains duplicate
-        data = np.array([1, 1])
-        indptr = np.array([0, 2])
-
-        M = self.csr_container((data, indices, indptr)).copy()
+        M = self.csr_container(info_with_dups).copy()
         assert_equal(False, M.has_canonical_format)
         assert isinstance(M.has_canonical_format, bool)
 
-        # set by deduplicating
+        # set flag by deduplicating
         M.sum_duplicates()
         assert_equal(True, M.has_canonical_format)
         assert_equal(1, len(M.indices))
 
-        M = self.csr_container((data, indices, indptr)).copy()
-        # set manually (although underlyingly duplicated)
+        # manually set flag True (although underlyingly duplicated)
+        M = self.csr_container(info_with_dups).copy()
         M.has_canonical_format = True
         assert_equal(True, M.has_canonical_format)
         assert_equal(2, len(M.indices))  # unaffected content
-
         # ensure deduplication bypassed when has_canonical_format == True
         M.sum_duplicates()
-        assert_equal(2, len(M.indices))  # unaffected content
+        assert_equal(2, len(M.indices))  # still has duplicates!!!!
+        # ensure deduplication reenabled when has_canonical_format == False
+        M.has_canonical_format = False
+        M.sum_duplicates()
+        assert_equal(1, len(M.indices))
+        assert_equal(True, M.has_canonical_format)
+
+        # manually set flag False (although underlyingly canonical)
+        M.has_canonical_format = False
+        assert_equal(False, M.has_canonical_format)
+        Mcheck = self.csr_container((M.data, M.indices, M.indptr))
+        assert_equal(True, Mcheck.has_canonical_format)
+        # sum_duplicates does not complain when no work to do
+        M.sum_duplicates()
+        assert_equal(True, M.has_canonical_format)
+
+        # check assignments maintain canonical format
+        M = self.csr_container((np.array([2]), np.array([2]), np.array([0, 1, 1, 1])))
+        assert_equal(M.shape, (3, 3))
+        with suppress_warnings() as sup:
+            sup.filter(SparseEfficiencyWarning, "Changing the sparsity structure")
+            M[0, 1] = 2
+            M[1, :] *= 5
+            M[0, 2] = 3
+        assert_equal(True, M.has_canonical_format)
+        Mcheck = self.csr_container((M.data, M.indices, M.indptr))
+        assert_equal(True, Mcheck.has_canonical_format)
+
+        # resetting index arrays before accessing M.has_canonical_format is OK
+        M = self.csr_container(info_no_dups)
+        M.data, M.indices, M.indptr = info_with_dups
+        assert_equal(False, M.has_canonical_format)
+        assert_equal(2, len(M.indices))  # dups and has_canonical_format is False
+
+        # but reset after accessing M.has_canonical_format can break flag
+        M = self.csr_container(info_no_dups)
+        M.has_canonical_format  # underlying attr is set here
+        M.data, M.indices, M.indptr = info_with_dups
+        assert_equal(True, M.has_canonical_format)
+        assert_equal(2, len(M.indices))  # dups but has_canonical_format is True
+        M.sum_duplicates()
+        assert_equal(2, len(M.indices))  # still has duplicates!!!!
 
     def test_scalar_idx_dtype(self):
         # Check that index dtype takes into account all parameters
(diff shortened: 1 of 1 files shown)


perimosocordiae (approved):
Test-only change that will help us detect behavior changes in potential future updates to has_canonical_format handling.

stefanv:
We could make the API more explicit for setting `has_canonical_format`, e.g., make users have to go via `set_canonical_format(True, void_index_check=True)` or similar.

Have we considered putting guards around accessing `.indices` too? E.g., we could expose `.indices` read-only, and only provide the settable version via `.get_indices(void_canonical=True)`.

scipy/sparse/tests/test_base.py
```diff
@@ -4286,31 +4286,70 @@ def test_has_sorted_indices(self):
...
+        info_with_dups = (np.array([1, 1]), np.array([0, 0]), np.array([0, 2]))
 
-        indices = np.array([0, 0])  # contains duplicate
-        data = np.array([1, 1])
-        indptr = np.array([0, 2])
+        M = self.csr_container(info_no_dups)
+        assert_equal(True, M.has_canonical_format)
 
-        M = self.csr_container((data, indices, indptr)).copy()
+        M = self.csr_container(info_with_dups).copy()
```
stefanv, line 4295: Is the copy here an additional test?

scipy/sparse/tests/test_base.py
```diff
@@ -4286,31 +4286,70 @@ def test_has_sorted_indices(self):
...
         assert_equal(True, M.has_canonical_format)
         assert_equal(2, len(M.indices))  # unaffected content
-
         # ensure deduplication bypassed when has_canonical_format == True
         M.sum_duplicates()
-        assert_equal(2, len(M.indices))  # unaffected content
+        assert_equal(2, len(M.indices))  # still has duplicates!!!!
+        # ensure deduplication reenabled when has_canonical_format == False
+        M.has_canonical_format = False
+        M.sum_duplicates()
```
stefanv, line 4314: This function *always* returns True; do we have any functions that return True/False depending on circumstances, that we should check here? Same below.

scipy/sparse/tests/test_base.py
```diff
@@ -4286,31 +4286,70 @@ def test_has_sorted_indices(self):
...
+            M[1, :] *= 5
+            M[0, 2] = 3
+        assert_equal(True, M.has_canonical_format)
+        Mcheck = self.csr_container((M.data, M.indices, M.indptr))
+        assert_equal(True, Mcheck.has_canonical_format)
+
+        # resetting index arrays before accessing M.has_canonical_format is OK
+        M = self.csr_container(info_no_dups)
+        M.data, M.indices, M.indptr = info_with_dups
+        assert_equal(False, M.has_canonical_format)
```
stefanv, line 4342: Only works on *first* calculation, right? Maybe another reason to void `has_canonical_format` on any assignment of those attributes, which can be detected via property.

scipy/sparse/tests/test_base.py
```diff
@@ -4868,6 +4907,51 @@ def test_tocompressed_duplicates(self):
...
+        assert A.has_canonical_format is True
+        A = self.coo_container(A_coords_form)
+        assert A.has_canonical_format is False
+        A.sum_duplicates()
+        assert A.has_canonical_format is True
+
+        A = self.coo_container(A, copy=True)
+        assert A.has_canonical_format is True
+        A = self.coo_container(A, copy=False)
+        assert A.has_canonical_format is False
```
stefanv, line 4930: Why does the `copy` flag affect whether it the flag is True or False?

scipy/sparse/tests/test_base.py
```diff
@@ -4286,31 +4286,70 @@ def test_has_sorted_indices(self):
...
+        M = self.csr_container(info_with_dups).copy()
         M.has_canonical_format = True
         assert_equal(True, M.has_canonical_format)
         assert_equal(2, len(M.indices))  # unaffected content
-
         # ensure deduplication bypassed when has_canonical_format == True
         M.sum_duplicates()
-        assert_equal(2, len(M.indices))  # unaffected content
+        assert_equal(2, len(M.indices))  # still has duplicates!!!!
+        # ensure deduplication reenabled when has_canonical_format == False
```
dschult, line 4312: This is the start of the added tests. Everything above is refactoring to obtain data/indices arrays that are reusable and named in a hopefully helpful way.

scipy/sparse/tests/test_base.py
dschult: The test without the copy (new line 4292-4293) replaces old line 4289-4290 test.
The test with the copy (new line 4295-4296) replaces old line 4296-4297 test. 
So it is a second test within this function, but it is not an additional test compared to before this PR.

scipy/sparse/tests/test_base.py
dschult: We currently don't have a public python function to check whether the array is actually canonical format. We do have the sparsetools function `_csr_has_canonical_format()` which returns bool indicating whether it has canonical format -- and doesn't do anything else.

So, on the next line (4315) the test is checking that *this particular array* has its duplicates removed.

It might be helpful for users to have a public python function that returns a bool to indicate whether canonical form. Let's do that as part of the code changes surrounding `has_canonical_format`.

scipy/sparse/tests/test_base.py
dschult: Yes!  We should add the code to detect assignment to `.data`, `.indices` and `indptr` and set the `has_canonical_format` to `False` upon assignment. 

That isn't fool proof in the sense that someone can change the values in those 3 attribute arrays to make it noncanonical instead of assigning to the attributes.  But monitoring the attribute assignments would make it that much more difficult for a user to mess up the flag -- they'd have to change the attribute arrays in-place and forget to set the flag.

scipy/sparse/tests/test_base.py
dschult: When `copy` is True, and the input is sparse, the `has_canonical_format` flag is set based on the incoming sparse flag value. When `copy` is False the flag is set to False.  This [happens in `_coo.py`](https://github.com/scipy/scipy/blob/e27588f9f232a525eb130f7dc6243c0ef12887ac/scipy/sparse/_coo.py#L66) 
I'm not sure of the history for why not copy from the incoming flag to the new flag no matter what the `copy` argument is. I think that would be a good thing to make more consistent when we change the code.  Let's set the has_canonical_format flag for COO format based on the flag's value on the incoming sparse object.

dschult (commented):
Good questions -- Most indicate places where we can/should improve the code to make this flag more useful. I have tried to indicate these spots with answering comments. 

We'll make those code changes (and more I'm sure) in the next PR and have to come change these tests to indicate the new behavior (with deprecations, etc).

To respond to your questions before your comments:
- yes -- moving the api for changing `has_canonical_format` in `_compress.py` to an explicit function instead of the property setting hook could make it more obvious that they need to be careful. +1     Note that the flag for COO is not a property atm. So we should make that a property and add the explicit changing api to do this for COO.
- your suggestion for how to make `indices` harder to change manually is nice (and sneaky). Could be done with `indptr` and the COO `coords` too. :)

stefanv (approved):

stefanv:
Thanks, Dan! Also for discussing with me the various improvements that can be made to `has_canonical_format`.

PR #22938, opened 2025-05-06
TST: `signal` add test coverage to the handling of new set of points being filtered down to empty in `cspline1d_eval` and `qspline1d_eval`

zitongzhoueric:
This PR adds two tests to ensure that `signal.cspline1d_eval` and `signal.qspline1d_eval` behaves correctly when the new set of points `newx` is filtered down to empty.

### DIFF
--- a/scipy/signal/tests/test_bsplines.py
+++ b/scipy/signal/tests/test_bsplines.py
@@ -162,6 +162,12 @@ class TestBSplines:
         r = signal.cspline1d_eval(xp.asarray([1., 0, 1], dtype=xp.float64),
                                xp.asarray([], dtype=xp.float64))
         xp_assert_equal(r, xp.asarray([], dtype=xp.float64))
+        
+        # Test case for newx that gets filtered down to empty
+        r = signal.cspline1d_eval(xp.asarray([1.0, 0, 1], dtype=xp.float64), 
+                                  xp.asarray([-1.0], dtype=xp.float64))
+        xp_assert_close(r, xp.asarray([0.33333333], dtype=xp.float64))
+
         x = [-3, -2, -1, 0, 1, 2, 3, 4, 5, 6]
         dx = x[1] - x[0]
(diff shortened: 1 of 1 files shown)


scipy/signal/tests/test_bsplines.py
```diff
@@ -162,6 +162,12 @@ def test_cspline1d_eval(self, xp):
         r = signal.cspline1d_eval(xp.asarray([1., 0, 1], dtype=xp.float64),
                                xp.asarray([], dtype=xp.float64))
         xp_assert_equal(r, xp.asarray([], dtype=xp.float64))
+        
+        # Test case for newx that gets filtered down to empty
+        r = signal.cspline1d_eval(xp.asarray([1.0, 0, 1], dtype=xp.float64), 
+                                  xp.asarray([-1.0], dtype=xp.float64))
+        xp_assert_close(r, xp.asarray([0.33333333], dtype=xp.float64))
```
j-bowhay, line 169: To save me some time, could you spell out how the reference value here is computed

zitongzhoueric: Not an expert on polynomial interpolation. I used the cubic B-spline [formula](https://sepwww.stanford.edu/sep/sergey/sepsergey/bspl/paper_html/node5.html ) and worked out the math.

By mirroring $$s(-1) = s(1)$$
The offsets $j$ that contribute to non-zero spline value are $${0,1,2}$$
Evaluating the basis spline at $x - j$ we get 

$$β_3(1-0) = β_3(1) = \frac{1}{6}, β_3(1-1) = β_3(0)  = \frac{2}{3}, β_3(1-2) = β_3(1) = \frac{1}{6}$$

Finally the weighted sum of the spline values with coefficients cj:

$$s(1) = c_0 β_3(1) + c_1 β_3(0) + c_2 β_3(-1) = 1 \times \frac{1}{6} + 0  \times \frac{2}{3} + 1 \times \frac{1}{6} = \frac{1}{3}$$

j-bowhay (approved):

Issue #21044, opened 2024-06-25
RFC: quo vadis, `xp_assert_*` infrastructure?


h-vetinari:
This topic sits at the interface of at least three-four different issues, at least two of which I seem to have a knack of running into. I don't consider myself the right (much less the best) person to write about this, but here we are...

Those four issues are:
1. [Array API](https://data-apis.org/array-api/latest/) standard adoption
1. Treatment of 0d-arrays in NumPy & SciPy
1. General testing facilities
1. Desirable fidelity in testing

I won't write about the first topic, which is covered elsewhere (e.g. https://github.com/scipy/scipy/issues/18286), I just take it as an obviously beneficial goal and acknowledge that its implementation is happening already.

## Treatment of 0d-arrays

NumPy arrays are nicely generic over their dimension, which can range from 0 to `NPY_MAXDIMS` (==32). NumPy also has its own scalars to complement the Python-native ones (e.g. `np.int64(...)` is obviously limited in range but has constant footprint, whereas `int(...)` is unbounded in both value and footprint). One long-standing historical inaccuracy has been that one only needs to breathe at a 0d-array the wrong way, and it decays into a scalar:
```python
>>> x= np.asarray(1.)
>>> -x
np.float64(-1.0)
```

This is BAD from a number of perspectives, principally those two objects have different APIs (although many methods/fields actually are compatible). Another issue is typing, e.g. it's ~impossible~ very hard to write a function
```
func(x: np.array) -> np.array
```
because there are so many ways that 0d inputs might actually return scalars, forcing the return type to be `np.array | scalar`.

For these reasons (and others), all array standard API implementations except NumPy actually keep 0d arrays as 0d arrays. Great! However, SciPy is still _very_ closely related to NumPy, and we cannot really get away from the reality of 20+ years of legacy that has grown around this. [NEP 56](https://numpy.org/neps/nep-0056-array-api-main-namespace.html) even call this out specifically (quoted):

> [!NOTE]
[...]

PR #22939, opened 2025-05-06
TST: spatial.transform: Add array API standard support for testing

amacati:
Add Array API support to `scipy.spatial.transform` tests in preparation of #22777 

#### Reference issue
#22500, #22777 

All tests are converted to be compatible with the Array API standard.

#### `jax.jit`
This PR is derived from #22777 where we already have a running Array API backend. To isolate changes to files, this PR already includes tests that check if the functions inside `scipy.spatial.transform` are compatible with `jax.jit`. In addition, three tests use `jax.jit` to accelerate testing, as the runtime without jitting increases significantly. Importantly, these changes do not influence the current behavior with numpy in any way.

#### Additional information
This PR does not enforce consistent formatting in the tests. Instead, it tries to match the existing code style to keep the amount of changes as small as possible and make reviewing easier. It would be great to have another PR after this one that makes clear which formatter should be used within `scipy.spatial.transform`. A general format style for `scipy` is not planned. See e.g. https://github.com/scipy/scipy/pull/22777#issuecomment-2827585033.


lucascolley:
> It would be great to have another PR after this one that makes clear which formatter should be used within scipy.spatial.transform. A general format style for scipy is not planned.

I'm happy to review a style-only PR 👍 just keep it separate from functional changes as you have planned

amacati:
> I'm happy to review a style-only PR 👍 just keep it separate from functional changes as you have planned

Will do once this one is merged. Thanks!

scipy/spatial/transform/tests/test_rigid_transform.py
```diff
@@ -1,21 +1,37 @@
...
+    is_numpy,
+    xp_assert_close,
+)
+import scipy._lib.array_api_extra as xpx
 
 
-def test_repr():
-    actual = repr(RigidTransform.identity())
+def rotation_to_xp(r: Rotation, xp):
+    return Rotation.from_quat(xp.asarray(r.as_quat()))
```
scottshambaugh, line 16: I'm unfamiliar with the array API, what are these two functions doing?

scipy/spatial/transform/tests/test_rigid_transform.py
```diff
@@ -228,103 +302,126 @@ def test_from_exp_coords():
...
         scalar_first=True)
-    expected_matrix = np.eye(4)
-    expected_matrix[:3, :3] = actual_rot.as_matrix()
-    assert_allclose(actual.as_matrix(), expected_matrix, atol=1e-12)
+    expected_matrix = xp.eye(4)
+    expected_matrix = xpx.at(expected_matrix)[..., :3, :3].set(actual_rot.as_matrix())
+    xp_assert_close(actual.as_matrix(), expected_matrix, atol=1e-12)
 
     # rotation and translation
+    # The tolerance is set to 1e-8, because xp_assert_close compares the absolute
```
scottshambaugh, line 414: Following up from the comments on the other PR, I think the tolerance change here is fine and doesn't need to be broken out to another PR. However I do believe that there is still a default rtol applied for floating point numbers, so this comment should change

scottshambaugh (approved):
I can confirm that all the tests are functionally unchanged and exercise the same behavior before and after this PR. Again, I'm not too familiar with the array API or the "why" behind these changes, but the changes themselves are no issue. The CI errors right now look like a setup issue rather than due to bad test values.

There's clearly an immense amount of effort here - well done! I'm good for a merge once the CI issues are worked through.

scipy/spatial/transform/tests/test_rigid_transform.py
amacati: `rotation_to_xp` is a helper that converts `Rotation` objects from `numpy` to `Rotation` objects with the correct Array API type. There are some functions in `Rotation` and `RigidTransform` that will always use numpy arrays, even in the new implementation. Examples are `Rotation.random` or `Rotation.identity` where it is impossible to decide which Array API framework should be used based on the input types, or where the Array API cannot provide a generic function (e.g. it does not specify how to create random numbers).

The conversion works like this:
`Rotation` in `numpy` -> `NDArray` with `as_quat()` -> `xp.Array` with `xp.asarray()` -> `Rotation` in `xp` with `R.from_quat()`

Long-term, we can change the input types to make it clear which framework to use, and then get rid of this helper.

lucascolley: > it is impossible to decide which Array API framework should be used based on the input types

There is also the option of adding an `xp` kwarg rather than changing allowed input types, where it really does make sense to accept Python scalars rather than 0D arrays. See e.g. https://docs.scipy.org/doc/scipy/reference/generated/scipy.fft.fftfreq.html.

Either way, it can be done in a follow-up long-term, using some NumPy for now is fine.

amacati: Yes, I was thinking along those lines. Some things like `random` will either need a specialized backend implementation or will have to use numpy though.

lucascolley:
A sample CI failures:

```
scipy/spatial/transform/tests/test_rotation.py:898: in test_as_euler_asymmetric_axes
rotation = Rotation.from_euler(seq, angles)
...
 _rotation.pyx:518: in scipy.spatial.transform._rotation._format_angles
     ???
...
E   TypeError: can't convert cuda:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first.
```

amacati:
> I guess we want to skip the whole modules for PyTorch and other backends for now?

Ah right, I thought that failed Array API tests are expected until we have merged the xp backend. They should be disabled now, let's see if the CI passes

lucascolley (commented):
thanks for the quick review @scottshambaugh !

I'll give @crusaderky some time to potentially check the JAX / lazy-specific sections, but otherwise we can probably merge this soon

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -897,139 +1068,199 @@ def test_as_euler_degenerate_compare_algorithms(seq_tuple, intrinsic):
...
-def test_identity_magnitude():
+@pytest.mark.array_api_backends
+def test_identity_magnitude(xp):
     n = 10
-    assert_allclose(Rotation.identity(n).magnitude(), 0)
-    assert_allclose(Rotation.identity(n).inv().magnitude(), 0)
+    r = Rotation.identity(n)
+    r = Rotation.from_quat(xp.asarray(r.as_quat()))
+    xp_assert_close(r.magnitude(), xp.asarray(0.0), check_shape=False)
+    xp_assert_close(r.inv().magnitude(), xp.asarray(0.0), check_shape=False)
```
crusaderky, line 1170: Why can't we test the output shape?

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -897,139 +1068,199 @@ def test_as_euler_degenerate_compare_algorithms(seq_tuple, intrinsic):
...
 
-def test_magnitude():
-    r = Rotation.from_quat(np.eye(4))
+
+@pytest.mark.array_api_backends
+def test_magnitude(xp):
+    r = Rotation.from_quat(xp.eye(4))
     result = r.magnitude()
-    assert_array_almost_equal(result, [np.pi, np.pi, np.pi, 0])
+    xp_assert_close(result, xp.asarray([np.pi, np.pi, np.pi, 0]))
```
crusaderky, line 1220:
```suggestion
    xp_assert_close(result, xp.asarray([xp.pi, xp.pi, xp.pi, 0]))
```

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -897,139 +1068,199 @@ def test_as_euler_degenerate_compare_algorithms(seq_tuple, intrinsic):
...
 
 
-def test_magnitude_single_rotation():
-    r = Rotation.from_quat(np.eye(4))
+@pytest.mark.array_api_backends
+def test_magnitude_single_rotation(xp):
+    r = Rotation.from_quat(xp.eye(4))
     result1 = r[0].magnitude()
-    assert_allclose(result1, np.pi)
+    xp_assert_close(result1, np.pi)
```
crusaderky, line 1231:
```suggestion
    xp_assert_close(result1, xp.pi)
```

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1040,40 +1271,68 @@ def test_approx_equal_single_rotation():
...
+    r = Rotation.from_matrix(jax.numpy.eye(3))
+    jax.block_until_ready(approx_equal(r, r))
+
+
+@pytest.mark.array_api_backends
+def test_mean(xp):
+    axes = xp.concat((-xp.eye(3), xp.eye(3)))
     axes = np.concatenate((-np.eye(3), np.eye(3)))
-    thetas = np.linspace(0, np.pi / 2, 100)
+    thetas = xp.linspace(0, np.pi / 2, 100)
```
crusaderky, line 1286:
```suggestion
    thetas = xp.linspace(0, xp.pi / 2, 100)
```

crusaderky, line 1286:
```suggestion
    thetas = xp.linspace(0, xp.pi / 2, 100)
```

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1303,80 +1622,86 @@ def test_random_rotation_shape():
...
+                                [0. , 1. , 1.]])
     atol = 1e-6
-    a = [[0, 1, 0], [0, 1, 1], [0, 1, 1]]
-    b = [[1, 0, 0], [1, 1.1, 0], [1, 0.9, 0]]
+    a = xp.asarray([[0, 1, 0], [0, 1, 1], [0, 1, 1]])
+    b = xp.asarray([[1, 0, 0], [1, 1.1, 0], [1, 0.9, 0]])
     rot, rssd, sens = Rotation.align_vectors(a, b, return_sensitivity=True)
-    assert np.isclose(rssd, rssd_expected, atol=atol)
-    assert np.allclose(sens, sens_expected, atol=atol)
+    assert xp.all(xpx.isclose(rssd, rssd_expected, atol=atol, xp=xp))
```
crusaderky, line 1670:
```suggestion
    xp_assert_close(rssd, rssd_expected, atol=atol)
```

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1303,80 +1622,86 @@ def test_random_rotation_shape():
...
-def test_align_vectors_scaled_weights():
+@pytest.mark.array_api_backends
+def test_align_vectors_scaled_weights(xp):
     n = 10
-    a = Rotation.random(n, rng=0).apply([1, 0, 0])
-    b = Rotation.random(n, rng=1).apply([1, 0, 0])
-    scale = 2
+    a = xp.asarray(Rotation.random(n, rng=0).apply([1, 0, 0]))
+    b = xp.asarray(Rotation.random(n, rng=1).apply([1, 0, 0]))
+    scale = xp.asarray(2.0)
```
crusaderky, line 1679:
```suggestion
    scale = 2
```

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1303,80 +1622,86 @@ def test_random_rotation_shape():
...
-    est1, rssd1, cov1 = Rotation.align_vectors(a, b, np.ones(n), True)
-    est2, rssd2, cov2 = Rotation.align_vectors(a, b, scale * np.ones(n), True)
+    est1, rssd1, cov1 = Rotation.align_vectors(a, b, xp.ones(n), True)
+    est2, rssd2, cov2 = Rotation.align_vectors(a, b, scale * xp.ones(n), True)
 
-    assert_allclose(est1.as_matrix(), est2.as_matrix())
-    assert_allclose(np.sqrt(scale) * rssd1, rssd2, atol=1e-6)
-    assert_allclose(cov1, cov2)
+    xp_assert_close(est1.as_matrix(), est2.as_matrix())
+    xp_assert_close(xp.sqrt(scale) * rssd1, rssd2, atol=1e-6)
```
crusaderky, line 1685:
```suggestion
    xp_assert_close(math.sqrt(scale) * rssd1, rssd2, atol=1e-6)
```

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1388,99 +1713,137 @@ def test_align_vectors_noise():
...
-    R, rssd = Rotation.align_vectors(a, b, weights=[np.inf, 1])
-    assert_allclose(R.as_matrix(), m_expected, atol=atol)
-    assert_allclose(R.apply(b), a, atol=atol)  # Pri and sec align exactly
-    assert np.isclose(rssd, 0, atol=atol)
+    b = xp.asarray([[1, 0, 0], [1, 1, 0]])
+    a = xp.asarray([[0.0, 1, 0], [0, 1, 1]])
+    m_expected = xp.asarray([[0.0, 0, 1],
+                             [1, 0, 0],
+                             [0, 1, 0]])
+    R, rssd = Rotation.align_vectors(a, b, weights=xp.asarray([np.inf, 1]))
```
crusaderky, line 1821:
```suggestion
    R, rssd = Rotation.align_vectors(a, b, weights=xp.asarray([xp.inf, 1]))
```

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1388,99 +1713,137 @@ def test_align_vectors_noise():
...
+    xp_assert_close(R.as_matrix(), m_expected, atol=atol)
+    xp_assert_close(R.apply(b), a, atol=atol)  # Pri and sec align exactly
+    assert xpx.isclose(rssd, xp.asarray(0.0), atol=atol, xp=xp)
```
crusaderky, line 1824:
```suggestion
    assert xpx.isclose(rssd, 0, atol=atol, xp=xp)
```

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1388,99 +1713,137 @@ def test_align_vectors_noise():
...
-    b = [[1, 2, 3], [-2, 3, -1]]
-    a = [[-1, 3, 2], [1, -1, 2]]
-    rssd_expected = 1.3101595297515016
-    R, rssd = Rotation.align_vectors(a, b, weights=[np.inf, 1])
-    assert_allclose(R.apply(b)[0], a[0], atol=atol)  # Only pri aligns exactly
-    assert np.isclose(rssd, rssd_expected, atol=atol)
+    b = xp.asarray([[1, 2, 3], [-2, 3, -1]])
+    a = xp.asarray([[-1.0, 3, 2], [1, -1, 2]])
+    rssd_expected = xp.asarray(1.3101595297515016)
+    R, rssd = Rotation.align_vectors(a, b, weights=xp.asarray([np.inf, 1]))
```
crusaderky, line 1840:
```suggestion
    rssd_expected = 1.3101595297515016
    R, rssd = Rotation.align_vectors(a, b, weights=xp.asarray([xp.inf, 1]))
```

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1489,60 +1852,80 @@ def test_align_vectors_near_inf():
...
                            [[0, 0, 1], [0, 1, 0]]])
+
+    # We jit these tests for JAX because the non-jitted version significantly slows down
+    # the test
+    align_vectors = Rotation.align_vectors
+    if is_jax(xp):
+        import jax  # Must be installed because is_jax succeeds
+
+        align_vectors = jax.jit(Rotation.align_vectors)
+
```
crusaderky, line 1922:
```suggestion
```
ditto

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1551,247 +1934,340 @@ def test_align_vectors_antiparallel():
...
     mats_b = Rotation.random(100, rng=1).as_matrix()
+
+    # We jit these tests for JAX because the non-jitted version significantly slows down
+    # the test
+    align_vectors = Rotation.align_vectors
+    if is_jax(xp):
+        import jax  # Must be installed because is_jax succeeds
+
+        align_vectors = jax.jit(Rotation.align_vectors)
+
```
crusaderky, line 1958:
```suggestion
```

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1551,247 +1934,340 @@ def test_align_vectors_antiparallel():
...
                             [0., 0., 1.]]))"""
-    assert actual == expected
+        assert actual == expected
 
 
-def test_repr_rotation_sequence():
-    q = np.array([[0, 1, 0, 1], [0, 0, 1, 1]]) / np.sqrt(2)
+@pytest.mark.array_api_backends
+def test_repr_rotation_sequence(xp):
+    q = xp.asarray([[0.0, 1, 0, 1], [0, 0, 1, 1]]) / np.sqrt(2)
```
crusaderky, line 1996:
```suggestion
    q = xp.asarray([[0.0, 1, 0, 1], [0, 0, 1, 1]]) / math.sqrt(2)
```

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1821,20 +2297,23 @@ def test_rotation_within_numpy_array():
...
+    # TODO: Can we support this for all Array API frameworks?
+    r = Rotation.from_quat(xp.asarray([0, 0, np.sin(np.pi/4), np.cos(np.pi/4)]))
     pkl = pickle.dumps(r)
     unpickled = pickle.loads(pkl)
-    assert_allclose(r.as_matrix(), unpickled.as_matrix(), atol=1e-15)
+    xp_assert_close(r.as_matrix(), unpickled.as_matrix(), atol=1e-15)
 
 
 def test_deepcopy():
+    # TODO: Can we support this for all Array API frameworks?
```
crusaderky, line 2309:
```suggestion
    # Note: Array API makes no provision for arrays to support the `__copy__`
    # protocol, so it's OK to skip this test for the backends that don't
```

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1976,23 +2502,39 @@ def test_as_davenport_degenerate():
...
             ax = ax_lamb if order == 'intrinsic' else ax_lamb[::-1]
-            rot = Rotation.from_davenport(ax, order, angles)
-            with pytest.warns(UserWarning, match="Gimbal lock"):
-                angles_dav = rot.as_davenport(ax, order)
+            rot = Rotation.from_davenport(xp.asarray(ax), order, angles)
+            if is_lazy_array(rot.as_quat()):
+                angles_dav = rot.as_davenport(xp.asarray(ax), order)
+            else:  # We can only warn in non-lazy frameworks.
+                with pytest.warns(UserWarning, match="Gimbal lock"):
+                    angles_dav = rot.as_davenport(xp.asarray(ax), order)
```
crusaderky, line 2519:
```suggestion
            with eager_warns(rot_as_quat(), UserWarning, match="Gimbal lock"):
                angles_dav = rot.as_davenport(xp.asarray(ax), order)
```

crusaderky (changes requested):
- `test_*_jax_compile` are problematic for the reasons listed here: https://github.com/scipy/scipy/pull/22777#discussion_r2057765298
  What I would like to see:
  
  - For all these tests to be removed
  - Have all methods decorated with `@xp_capabilities`
  - At the top of each test module, add `lazy_xp_modules = [Rotation, RigidTransform]`
  - Have all tests decorated with `@make_skip_xp_backends(Rotation.<method>, ...)`

  _However_, this pattern

  - requires https://github.com/data-apis/array-api-extra/pull/284 to work (it should be merged any day now);
  - is untested with instance methods and class methods (need to add unit tests to array-api-extra);
  - I expect won't work with the current Cython classes.
  
  In the interest of pushing this PR through, I ask for now to just delete all `test_*_jax_compile` and leave JAX-specific treatment to a follow-up.

- Lazy backends don't have any way of eagerly validating inputs, as per the pattern established by `scipy.cluster.is_valid_linkage`.
  In the interest of pushing this PR through, I'd be content to have a follow-up issue to track the problem. 

- These methods, and probably a few more, have lost all test cases for when the input is a scalar or array-like:

  - from_euler
  - from_matrix
  - from_quat
  - from_translation
  - align_vectors
  - `__getitem__`

  Please add specific test cases for them (unless they're already there and I missed them).

- A bunch of minor observations as noted inline.

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -18,732 +30,869 @@ def basis_vec(axis):
...
 
 
-def test_from_square_quat_matrix():
+@pytest.mark.array_api_backends
+def test_from_square_quat_matrix(xp):
     # Ensure proper norm array broadcasting
-    x = np.array([
-        [3, 0, 0, 4],
+    x = xp.asarray([
+        [3.0, 0, 0, 4],
```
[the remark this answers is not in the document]
lucascolley: Typically these sorts of changes are fine if an extra test is added per function for dtype preservation

scipy/spatial/transform/tests/test_rigid_transform.py
```diff
@@ -29,176 +48,234 @@ def test_repr():
...
-        invalid[3, 3] = 2  # Invalid last row
-        RigidTransform.from_matrix(invalid)
+    invalid = xp.eye(4)
+    invalid = xpx.at(invalid)[..., 3, 3].set(2)  # Invalid last row
+    if is_lazy_array(invalid):
+        tf = RigidTransform.from_matrix(invalid)
+        assert xp.all(xp.isnan(tf.as_matrix()))
+    else:
+        with pytest.raises(ValueError):
+            RigidTransform.from_matrix(invalid)
```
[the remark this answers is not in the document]
amacati: Are you suggesting to add something like a `RigidTransform.is_valid_matrix`? That should be fairly easy to add for all options. However, where would you put this addition? The lazy distinctions (and `..jax_compile`) in the tests were only made to have all test changes in one commit. This would be a code change.

I could also remove the special lazy handling for now, since the tests are just supposed to work for numpy (as of now). If we add new methods to `RigidTransform` and `Rotation`, the PRs start to get mixed up again.

scipy/spatial/transform/tests/test_rigid_transform.py
```diff
@@ -1,21 +1,40 @@
...
+
+def rotation_to_xp(r: Rotation, xp):
+    return Rotation.from_quat(xp.asarray(r.as_quat()))
+
+
+def rigid_transform_to_xp(r: RigidTransform, xp):
+    return RigidTransform.from_matrix(xp.asarray(r.as_matrix()))
+
+
+@pytest.mark.array_api_backends
```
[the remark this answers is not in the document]
amacati: Oh, good to know, thanks! Will remove them everywhere. Sometimes the pytest magic of inserting fixtures is a bit too convoluted for my taste..

scipy/spatial/transform/tests/test_rigid_transform.py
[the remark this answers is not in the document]
crusaderky: As said elsewhere, I'm happy to have such a change discussed and implemented separately.

> I could also remove the special lazy handling for now

The point of this PR is to keep the size of the main PR down.
The moment you add lazy backends you'll need these code branches anyway.

To clarify: these changes to the already existing tests are OK, but they highlight a missing feature (eager validation) that impacts lazy backends only.

scipy/spatial/transform/tests/test_rigid_transform.py
```diff
@@ -228,103 +305,126 @@ def test_from_exp_coords():
...
-    assert_allclose(actual.as_matrix(), expected_matrix, atol=1e-12)
+    expected_matrix = xp.eye(4)
+    expected_matrix = xpx.at(expected_matrix)[..., :3, :3].set(actual_rot.as_matrix())
+    xp_assert_close(actual.as_matrix(), expected_matrix, atol=1e-12)
 
     # rotation and translation
+    # The tolerance is set to 1e-8 because xp_assert_close deviates from 
+    # np.testing.assert_allclose in that it does not automatically default to 1e-7 for
+    # floating point inputs.
+    # See https://numpy.org/doc/2.2/reference/generated/numpy.testing.assert_allclose.html
```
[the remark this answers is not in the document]
amacati: At the moment, the line reads https://github.com/scipy/scipy/blob/c41054dfd5f3779bc350d9896191df10efe6bd91/scipy/spatial/transform/tests/test_rigid_transform.py#L336
which explicitly sets `atol`. If we do set `rtol` here, I assume we'd have to check for fp32 / fp64 manually, which also feels off. The comment was mainly meant to remind me where the difference in behaviour between xp and np stems from, because I wasn't aware of it before. I wasn't sure which combinations of `rtol` and `atol` would make fp32 computations pass, so I went for the easiest solution by reducing `atol`. Maybe this should be left at `1e-12` for now and revisited once the Array API backend and fp32 become relevant?

scipy/spatial/transform/tests/test_rigid_transform.py
```diff
@@ -333,520 +433,599 @@ def test_from_dual_quat():
...
         [-0.2547655, 1.23506123, 0.20230088, 0.24247194,  # norm 1.3
          0.38559628, 0.08184063, 0.1755943, -0.1582222]  # orthogonal
     )
-    assert pytest.approx(np.linalg.norm(unnormalized_dual_quat[:4])) == 1.3
-    assert pytest.approx(np.dot(unnormalized_dual_quat[:4],
-                                unnormalized_dual_quat[4:]), abs=8) == 0.0
+    xp_assert_close(xp_vector_norm(unnormalized_dual_quat[:4]), xp.asarray(1.3),
+                    atol=1e-12, check_0d=False)
+    xp_assert_close(xp.vecdot(unnormalized_dual_quat[:4], unnormalized_dual_quat[4:]),
+                    xp.asarray(0.0), atol=1e-8, check_0d=False)
```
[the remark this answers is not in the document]
amacati: The result is not exactly zero, so the tolerances are needed, but I'll get rid of `check_0d`. Just out of curiosity: Why does `check_0d` exist if we don't use it in the tests?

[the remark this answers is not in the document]
lucascolley: FWIW I wouldn't mind `check_0d` here. Probably cleaner without. But there is also the argument that `[()]` is undefined behaviour on general arrays (https://github.com/data-apis/array-api/issues/833).

> Why does check_0d exist if we don't use it in the tests?

Maybe it doesn't need to exist anymore, if it is really unused.

The history of 0d checking with these assertions is long and complicated—https://github.com/scipy/scipy/issues/21044 for the interested reader.

scipy/spatial/transform/tests/test_rigid_transform.py
```diff
@@ -333,520 +433,599 @@ def test_from_dual_quat():
...
+    t = xp.asarray(rng.normal(size=(3,)))
+    r = rotation_to_xp(Rotation.random(10, rng=rng), xp=xp)
     random_dual_quats = RigidTransform.from_components(t, r).as_dual_quat()
 
     # ensure that random quaternions are not normalized
-    random_dual_quats[:, :4] = random_dual_quats[:, :4].round(2)
-    assert not np.any(np.isclose(
-        np.linalg.norm(random_dual_quats[:, :4], axis=1), 1.0, atol=0.0001))
+    non_normal_dual_quats = random_dual_quats[:, :4] + 0.01
+    random_dual_quats = xpx.at(random_dual_quats)[:, :4].set(non_normal_dual_quats)
```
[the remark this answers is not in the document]
amacati: Yes. The rounding before was to make sure there is a deviation from the normalized (i.e. norm 1) quaternions. Since `round()` was not available, the new version adds a constant to the quaternion to achieve the same effect. The assert afterwards double-checks that we really have non-normalized quaternions.

scipy/spatial/transform/tests/test_rigid_transform.py
```diff
@@ -860,86 +1039,105 @@ def test_indexing_validation():
...
 
     # Test invalid value type
     with pytest.raises(TypeError, match="value must be a RigidTransform"):
-        tf[0] = np.eye(4)
+        tf[0] = xp.eye(4)
 
 
-def test_copy_flag():
+@pytest.mark.skip_xp_backends(eager_only=True,
+                              reason="lazy backends do not support memory sharing")
```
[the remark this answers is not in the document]
amacati: Thanks for the correction. I've only worked with jax as lazy framework before and was assuming that other lazy frameworks have the same restriction on mutability.

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -18,732 +30,869 @@ def basis_vec(axis):
...
-            expected_quat)
+    xp_assert_close(Rotation.from_matrix(mat).as_quat(), expected_quat)
 
 
-def test_from_matrix_calculation():
-    expected_quat = np.array([1, 1, 6, 1]) / np.sqrt(39)
-    mat = np.array([
+@pytest.mark.array_api_backends
+def test_from_matrix_calculation(xp):
+    atol = 1e-8
```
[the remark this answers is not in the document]
amacati: No, but same as the other one I'm not sure which settings are preferred here. `atol` was just used before to control the precision, so I stuck with it. This should only be required for fp32

scipy/spatial/transform/tests/test_rotation.py
amacati: The previous assert just checked against 0, so it also didn't cover the shape. Using `check_shape=False` replicates that as closely as possible. I have changed it now to include the shape using `xp.zeros(n)`.

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1489,60 +1852,80 @@ def test_align_vectors_near_inf():
...
         mats.append(Rotation.random(n, rng=10 + i).as_matrix())
 
+    # We jit these tests for JAX because the non-jitted version significantly slows down
+    # the test
+    align_vectors = Rotation.align_vectors
+    if is_jax(xp):
+        import jax  # Must be installed because is_jax succeeds
+
+        align_vectors = jax.jit(Rotation.align_vectors)
+
```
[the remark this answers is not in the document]
amacati: I believe these are not yet ready, right? Will remove the jit anyways

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1551,247 +1934,340 @@ def test_align_vectors_antiparallel():
...
+    jax.block_until_ready(jax.jit(Rotation.align_vectors)(a, b))
+
+
+@pytest.mark.array_api_backends
+def test_repr_single_rotation(xp):
+    q = xp.asarray([0, 0, 0, 1])
     actual = repr(Rotation.from_quat(q))
-    expected = """\
+    # TODO: Decide how we want to print that the inner array may be a different type
+    # (e.g. a torch.Tensor)
```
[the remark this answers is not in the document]
amacati: What is your opinion on this? Ideally the xp framework would be visible in the print

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1886,61 +2381,90 @@ def test_len_and_bool():
...
+    rot = Rotation.from_rotvec(axes * np.pi/4)
+    rot_dav = Rotation.from_davenport(axes, 'e', [np.pi/4])
+    xp_assert_close(rot.as_quat(canonical=True),
                     rot_dav.as_quat(canonical=True))
 
     # Single rotation, two axes, axes.shape == (2, 3)
-    rot = Rotation.from_rotvec([np.array(ez) * np.pi/4,
-                                np.array(ey) * np.pi/6])
+    axes = xp.stack([ez, ey], axis=0)
+    rot = Rotation.from_rotvec(axes * xp.asarray([[np.pi/4], [np.pi/6]]))
```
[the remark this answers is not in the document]
amacati: Done

amacati:
I have addressed most of the review comments by removing the jax.jit tests, adding tests for array_like inputs and scalars where applicable, and fixing the inline comments from @crusaderky 

The main remaining item is the change of tolerances for some tests. I'm not sure what your preference or policy is there

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1236,47 +1417,62 @@ def test_getitem():
...
+                     [1, 0, 0],
+                     [0, 0, 1]],
+
+                    [[1, 0, 0],
+                     [0, 0, -1],
+                     [0, 1, 0]]])
+    mat = xp.asarray(mat)
+    r = Rotation.from_matrix(mat)
+    xp_assert_close(r[[0]].as_matrix(), mat[[0]], atol=1e-15)
+    xp_assert_close(r[[0, 1]].as_matrix(), mat[[0, 1]], atol=1e-15)
```
crusaderky, line 1444: I'd rather leave this behaviour untested for xp != np. Intuitively, this should be aligned to the Array API `__getitem__` where array-likes are not supported (read: their behaviour is undefined).
```suggestion
def test_getitem_array_like():
    mat = np.array([[[0.0, -1, 0],
                     [1, 0, 0],
                     [0, 0, 1]],
                    [[1, 0, 0],
                     [0, 0, -1],
                     [0, 1, 0]]])
    r = Rotation.from_matrix(mat)
    xp_assert_close(r[[0]].as_matrix(), mat[[0]], atol=1e-15)
    xp_assert_close(r[[0, 1]].as_matrix(), mat[[0, 1]], atol=1e-15)
```

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1821,20 +2112,25 @@ def test_rotation_within_numpy_array():
     assert_equal(array.shape, (3, 2))
 
 
-def test_pickling():
-    r = Rotation.from_quat([0, 0, np.sin(np.pi/4), np.cos(np.pi/4)])
+def test_pickling(xp):
+    # Note: Array API makes no provision for arrays to be pickleable, so
+    # it's OK to skip this test for the backends that don't support it
+    r = Rotation.from_quat(xp.asarray([0, 0, math.sin(xp.pi/4), math.cos(xp.pi/4)]))
```
crusaderky, line 2118:
```suggestion
    r = Rotation.from_quat(xp.asarray([0, 0, math.sin(np.pi/4), math.cos(np.pi/4)]))
```

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1821,20 +2112,25 @@ def test_rotation_within_numpy_array():
...
     pkl = pickle.dumps(r)
     unpickled = pickle.loads(pkl)
-    assert_allclose(r.as_matrix(), unpickled.as_matrix(), atol=1e-15)
+    xp_assert_close(r.as_matrix(), unpickled.as_matrix(), atol=1e-15)
 
 
 def test_deepcopy():
```
crusaderky, line 2124:
```suggestion
def test_deepcopy(xp):
```

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1821,20 +2112,25 @@ def test_rotation_within_numpy_array():
...
-    r = Rotation.from_quat([0, 0, np.sin(np.pi/4), np.cos(np.pi/4)])
+    # Note: Array API makes no provision for arrays to support the `__copy__`
+    # protocol, so it's OK to skip this test for the backends that don't
+    r = Rotation.from_quat([0, 0, math.sin(np.pi/4), math.cos(np.pi/4)])
```
crusaderky, line 2127:
```suggestion
    r = Rotation.from_quat(xp.asarray([0, 0, math.sin(np.pi/4), math.cos(np.pi/4)]))
```

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1886,180 +2185,215 @@ def test_len_and_bool():
...
 
     # symmetric sequences
-    angles[:, 0] = rnd.uniform(low=-np.pi, high=np.pi, size=(n,))
-    angles_middle = [rnd.choice([0, np.pi]) for i in range(n)]
-    angles[:, 2] = rnd.uniform(low=-np.pi, high=np.pi, size=(n,))
-    lambdas = rnd.uniform(low=0, high=np.pi, size=(5,))
+    angles[:, 0] = rnd.uniform(low=-xp.pi, high=xp.pi, size=(n,))
+    angles_middle = [rnd.choice([0, xp.pi]) for i in range(n)]
+    angles[:, 2] = rnd.uniform(low=-xp.pi, high=xp.pi, size=(n,))
+    lambdas = rnd.uniform(low=0, high=xp.pi, size=(5,))
```
crusaderky, line 2309:
```suggestion
    angles[:, 0] = rnd.uniform(low=-np.pi, high=np.pi, size=(n,))
    angles_middle = [rnd.choice([0, np.pi]) for i in range(n)]
    angles[:, 2] = rnd.uniform(low=-np.pi, high=np.pi, size=(n,))
    lambdas = rnd.uniform(low=0, high=np.pi, size=(5,))
```

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1886,180 +2185,215 @@ def test_len_and_bool():
...
     n = 100
     angles = np.empty((n, 3))
 
     # symmetric sequences
-    angles[:, 0] = rnd.uniform(low=-np.pi, high=np.pi, size=(n,))
-    angles[:, 1] = rnd.uniform(low=0, high=np.pi, size=(n,))
-    angles[:, 2] = rnd.uniform(low=-np.pi, high=np.pi, size=(n,))
+    angles[:, 0] = rnd.uniform(low=-xp.pi, high=xp.pi, size=(n,))
+    angles[:, 1] = rnd.uniform(low=0, high=xp.pi, size=(n,))
+    angles[:, 2] = rnd.uniform(low=-xp.pi, high=xp.pi, size=(n,))
```
crusaderky, line 2337:
```suggestion
    angles[:, 0] = rnd.uniform(low=-np.pi, high=np.pi, size=(n,))
    angles[:, 1] = rnd.uniform(low=0, high=np.pi, size=(n,))
    angles[:, 2] = rnd.uniform(low=-np.pi, high=np.pi, size=(n,))
```

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1886,180 +2185,215 @@ def test_len_and_bool():
...
-            dav = Rotation.from_davenport(ax, order, angles)
-            assert_allclose(eul.as_quat(canonical=True), dav.as_quat(canonical=True),
+            eul = Rotation.from_euler(seq, xp.asarray(angles))
+            dav = Rotation.from_davenport(xp.asarray(ax), order, xp.asarray(angles))
+            xp_assert_close(eul.as_quat(canonical=True), dav.as_quat(canonical=True),
                             rtol=1e-12)
 
     # asymmetric sequences
-    angles[:, 1] -= np.pi / 2
+    angles[:, 1] -= xp.pi / 2
```
crusaderky, line 2350:
```suggestion
    angles[:, 1] -= np.pi / 2
```

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1886,180 +2185,215 @@ def test_len_and_bool():
...
     n = 100
     angles = np.empty((n, 3))
 
     # symmetric sequences
-    angles[:, 0] = rnd.uniform(low=-np.pi, high=np.pi, size=(n,))
-    angles[:, 1] = rnd.uniform(low=0, high=np.pi, size=(n,))
-    angles[:, 2] = rnd.uniform(low=-np.pi, high=np.pi, size=(n,))
+    angles[:, 0] = rnd.uniform(low=-xp.pi, high=xp.pi, size=(n,))
+    angles[:, 1] = rnd.uniform(low=0, high=xp.pi, size=(n,))
+    angles[:, 2] = rnd.uniform(low=-xp.pi, high=xp.pi, size=(n,))
```
crusaderky, line 2370:
```suggestion
    angles[:, 0] = rnd.uniform(low=-np.pi, high=np.pi, size=(n,))
    angles[:, 1] = rnd.uniform(low=0, high=np.pi, size=(n,))
    angles[:, 2] = rnd.uniform(low=-np.pi, high=np.pi, size=(n,))
```

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1886,180 +2185,215 @@ def test_len_and_bool():
...
+            rot = Rotation.from_euler(seq, xp.asarray(angles))
             eul = rot.as_euler(seq)
-            dav = rot.as_davenport(ax, order)
-            assert_allclose(eul, dav, rtol=1e-12)
+            dav = rot.as_davenport(xp.asarray(ax), order)
+            xp_assert_close(eul, dav, rtol=1e-12)
 
     # asymmetric sequences
-    angles[:, 1] -= np.pi / 2
+    angles[:, 1] -= xp.pi / 2
```
crusaderky, line 2383:
```suggestion
    angles[:, 1] -= np.pi / 2
```

crusaderky (changes requested):
I don't have a very strong opinion re. tolerances.
On one hand, yes it would be nice if they were in a separate, specific PR; on the other hand I don't want to add unnecessary attrition.

So, I defer the decision to you.

A bunch of nits below. I'll be away until next Thursday; please don't wait for my formal thumbs up to merge.

scipy/spatial/transform/tests/test_rigid_transform.py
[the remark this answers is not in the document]
crusaderky: As a final user, I would treat it as a regression if a function that started to return a np.ndarray all of a sudden while until the previous version it returned a np.float64. Namely, everything that relies on the returned value to be a subclass of float would break; everything that expects it to be hashable would follow the same fate.

I agree that we should reconsider the usefulness of check_0d. To me it makes sense only in a case where you can't predict if you're going to receive an array or a generic, but... you can always do that even if you don't know the shape of your `expect`'ed array by just applying `[()]` to it.

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -18,732 +32,847 @@ def basis_vec(axis):
...
 
     rnd = np.random.RandomState(0)
     n = 1000
     angles = np.empty((n, 3))
-    angles[:, 0] = rnd.uniform(low=-np.pi, high=np.pi, size=(n,))
-    angles[:, 1] = rnd.uniform(low=-np.pi / 2, high=np.pi / 2, size=(n,))
-    angles[:, 2] = rnd.uniform(low=-np.pi, high=np.pi, size=(n,))
+    angles[:, 0] = rnd.uniform(low=-xp.pi, high=xp.pi, size=(n,))
+    angles[:, 1] = rnd.uniform(low=-xp.pi / 2, high=xp.pi / 2, size=(n,))
+    angles[:, 2] = rnd.uniform(low=-xp.pi, high=xp.pi, size=(n,))
```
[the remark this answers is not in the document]
amacati: No worries, it makes sense to be strict. This was changed by a bulk search and rename. Good catch

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1224,7 +1388,24 @@ def test_apply_shapes():
...
+    t = rng.uniform(-100, 100, size=(3,)).tolist()
+    v = r.apply(t)
+    v_expected = r.apply(xp.asarray(t))
+    xp_assert_close(v, v_expected, atol=1e-12)
+
+    # Multiple vectors
+    t = rng.uniform(-100, 100, size=(2, 3)).tolist()
+    v = r.apply(t)
+    v_expected = r.apply(xp.asarray(t))
+    xp_assert_close(v, v_expected, atol=1e-12)
```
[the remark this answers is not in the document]
amacati: This change was intentional. The thought here was that if we already have an xp rotation, we may want to be able to pass in lists to those as well. But it also makes sense to only test this for numpy. After all, passing lists when dealing with xp other than numpy is most likely unintentional

scipy/spatial/transform/tests/test_rotation.py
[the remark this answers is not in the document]
purepani: > This is just defensive in case one day some backend defines pi as something else.

Would be great for an april fools backend :smile: : https://en.wikipedia.org/wiki/Indiana_pi_bill

amacati:
Okay, I think this is ready to be merged if the CI passes. Thanks a lot @lucascolley, @scottshambaugh and @crusaderky for the detailed reviews. That was a lot of code to work through. Now that the tests are in place we can hopefully start to merge the Array API backend soon.

scipy/spatial/transform/tests/test_rotation.py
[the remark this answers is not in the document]
crusaderky: I was actually implying the possibility of a `float128`, or a ML backend defining it as `float32` or even `float16` or `bfloat16`, but `3.14` will do too I suppose 😆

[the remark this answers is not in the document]
purepani: Not even that! It was implied to be 3.2 in that bill!

amacati:
By the way, I see that `xp_default_dtype` returns `float64` for all backends but torch. How is this consistent with `jax`'s default precision of `float32`?

lucascolley:
https://docs.jax.dev/en/latest/default_dtypes.html#the-x64-flag-enabling-64-bit-values, https://github.com/scipy/scipy/blob/main/scipy/conftest.py#L193. It seems that JAX effectively offers 2 modes: either you cannot use 64-bit dtypes at all, or they are the default.

lucascolley (approved):
This has been thoroughly reviewed at this point and is passing CI—thanks a lot @amacati ! Excited for the follow-ups!

amacati:
> https://github.com/scipy/scipy/blob/main/scipy/conftest.py#L193. It seems that JAX effectively offers 2 modes: either you cannot use 64-bit dtypes at all, or they are the default.

I wasn't aware `scipy` sets the f64 variable for testing. This might be of interest for the coming PRs. Since `jax` is usually running on f32, it would make sense to also test that. In addition, the Array API backend draft is currently designed to preserve f32 for **all** backends but `numpy` (because that would be a change in behaviour from the current implementation). I.e., a `Rotation` created with a `torch.float32` tensor will preserve the `dtype` and calculate in single precision.

This design makes most sense to me, but we'll probably have to discuss this decision once the backend PR is ready.

lucascolley:
1. I think there is definitely room to make 'no f64 support assumed' work in some parts of SciPy
2. Doing that for the whole of SciPy seems like a monumental task. I think some parts just won't work 
3. Differing dtype behaviour for NumPy and alternative backends may be justified, but we will have to consider this again when we come to it.

On (2), I think we probably want to wait for some movement on https://github.com/data-apis/array-api-compat/issues/162. It would be interesting to see the quantity if failures if we implemented a 32-bit dtype which 'pretends' to be `float64` (with some sort of shim to expect 32-bit precision results in tests).

crusaderky:
> 1. I think there is definitely room to make 'no f64 support assumed' work in some parts of SciPy
> 2. Doing that for the whole of SciPy seems like a monumental task. I think some parts just won't work

Worth pointing out that without that flag JAX is falling foul of the Array API specs. That's very different from PyTorch's default type, which impacts exactly just that and which is covered by the Standard. 

> 3. Differing dtype behaviour for NumPy and alternative backends may be justified, but we will have to consider this again when we come to it.

The default dtype being float32 instead of float64 is covered. The rest shouldn't IMHO.

lucascolley:
> I assume that's because type promotions etc do not work the way one expects them to in fp32 mode?

It's more simple than that—the standard requires that e.g. `float64` is supported (https://data-apis.org/array-api/draft/API_specification/data_types.html#api-specification-data-types--page-root). I think it was briefly discussed at some point to figure out how to relax that requirement, especially with MLX being 32-bit only. But I don't think much work has happened in that direction.

PR #22941, opened 2025-05-07
MAINT: stats.qmc.Sobol: fix stacklevel of warning

j-bowhay:
A pretty trivial change but it cropped up while using this in my own research so thought I'd quickly fix. The stacklevel is currently set to 2 however the call stack is `.random() ->._random() -> warnings.warn()` so it should be set to 3.


lucascolley (approved):
thanks Jake!

tupui:
Nice 👍 glad to hear you are using this 😃 feel free to DM me if you have any questions!

PR #22944, opened 2025-05-08
MAINT: fix regressions in array-api-strict after disabling np.float64

crusaderky:
Fix regressions after
- data-apis/array-api-strict#145
- data-apis/array-api-strict#148

### DIFF
--- a/scipy/fft/_fftlog_backend.py
+++ b/scipy/fft/_fftlog_backend.py
@@ -1,3 +1,5 @@
+import math
+
 import numpy as np
 from warnings import warn
 from ._basic import rfft, irfft
@@ -8,7 +10,7 @@ from scipy._lib._array_api import array_namespace
 __all__ = ['fht', 'ifht', 'fhtoffset']
 
 # constants
-LN_2 = np.log(2)
+LN_2 = math.log(2)
 
 
 def fht(a, dln, mu, offset=0.0, bias=0.0):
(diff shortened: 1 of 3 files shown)


scipy/fft/_fftlog_backend.py
```diff
@@ -168,11 +170,11 @@ def fhtoffset(dln, mu, initial=0.0, bias=0.0):
...
-    y = np.pi/(2*dln)
-    zp = loggamma(xp + 1j*y)
-    zm = loggamma(xm + 1j*y)
-    arg = (LN_2 - lnkr)/dln + (zp.imag + zm.imag)/np.pi
-    return lnkr + (arg - np.round(arg))*dln
+    y = math.pi/(2*dln)
+    zp = complex(loggamma(xp + 1j*y))
+    zm = complex(loggamma(xm + 1j*y))
+    arg = (LN_2 - lnkr)/dln + (zp.imag + zm.imag)/math.pi
+    return lnkr + (arg - round(arg))*dln
```
crusaderky, line 177: I'm a bit conflicted about this one.
On one hand, this used to _accidentally_ allow numpy array inputs, and now it won't anymore.
On the other hand, the documentation clearly states that all inputs must be plain floats, and there are no tests whatsoever for vectorized inputs.
Is it OK to just disallow numpy arrays?

lucascolley: as much as I would like to, we probably shouldn't break the undocumented behaviour if there is a suitable alternative—is there?

crusaderky: We can force the output to float in the unit tests

lucascolley: That sounds fine?

crusaderky: Changed it

lucascolley (approved):
thanks Guido!

PR #22953, opened 2025-05-08
REL, MAINT: prep for 1.15.4

tylerjereddy:
* Prepare for SciPy `1.15.4`, just in case.

This is very unlikely to happen unless there is an emergency, since branching for `1.16.x` is approaching, some of us are at the summit next week, etc.


rgommers (approved):
Thanks Tyler!

Issue #21424, opened 2024-08-21
REL: missing MacOS 10.13 Py3.12 wheel


axiomofjoy:
### Describe your issue.

`scipy==1.14.1` appears to be causing installation issues with our package on MacOS with Python 3.12.

### Reproducing Code Example

I have not reproduced locally, but it consistently happens on our MacOS runners on GitHub Actions. Resolved the issue with [this PR](https://github.com/Arize-ai/phoenix/pull/4298). See an example run [here](https://github.com/Arize-ai/phoenix/actions/runs/10481800150/job/29031869117).

### Error message

```shell
Installing project in development mode
  error: subprocess-exited-with-error
  
  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [43 lines of output]
      + meson setup /private/var/folders/k_/2n6sfcjj6b7dm7znt8sx5lgw0000gn/T/pip-install-89_auglf/scipy_5a4541f396854fd580955c262fa91e93 /private/var/folders/k_/2n6sfcjj6b7dm7znt8sx5lgw0000gn/T/pip-install-89_auglf/scipy_5a4541f396854fd580955c262fa91e93/.mesonpy-45vze25y -Dbuildtype=release -Db_ndebug=if-release -Db_vscrt=md --native-file=/private/var/folders/k_/2n6sfcjj6b7dm7znt8sx5lgw0000gn/T/pip-install-89_auglf/scipy_5a4541f396854fd580955c262fa91e93/.mesonpy-45vze25y/meson-python-native-file.ini
      The Meson build system
      Version: 1.5.1
      Source dir: /private/var/folders/k_/2n6sfcjj6b7dm7znt8sx5lgw0000gn/T/pip-install-89_auglf/scipy_5a4541f396854fd580955c262fa91e93
      Build dir: /private/var/folders/k_/2n6sfcjj6b7dm7znt8sx5lgw0000gn/T/pip-install-89_auglf/scipy_5a4541f396854fd580955c262fa91e93/.mesonpy-45vze25y
      Build type: native build
      Project name: scipy
      Project version: 1.14.1
      C compiler for the host machine: cc (clang 14.0.0 "Apple clang version 14.0.0 (clang-1400.0.29.202)")
      C linker for the host machine: cc ld64 820.1
      C++ compiler for the host machine: c++ (clang 14.0.0 "Apple clang version 14.0.0 (clang-1400.0.29.202)")
[...]

PR #22955, opened 2025-05-08
MAINT: wheel downloader

andyfaff:
This fixes the downloading of all the wheels as part of the release process. This is to fix the issue mentioned in the comment at https://github.com/scipy/scipy/issues/21424#issuecomment-2863526109. The problem in that comment arises because https://anaconda.org/multibuild-wheels-staging/scipy/files is a "fancy" webpage, which only has so much space for listing files before one has to manually press next to see more files. The page isn't able to list all files for a given scipy version at once, which means that not all the wheel names for that version are scrapable.

This PR changes the staging URL to https://pypi.anaconda.org/multibuild-wheels-staging/simple/scipy/, which I think gives all the files. When I use this different URL all 45 of the wheels are downloaded.

@tylerjereddy

### DIFF
--- a/tools/download-wheels.py
+++ b/tools/download-wheels.py
@@ -16,7 +16,8 @@ from bs4 import BeautifulSoup
 __version__ = '0.1'
 
 # Edit these for other projects.
-STAGING_URL = "https://pypi.anaconda.org/multibuild-wheels-staging/simple/scipy/"
+STAGING_FILE_URL = "https://pypi.anaconda.org/multibuild-wheels-staging/simple/scipy/"
+STAGING_URL = 'https://anaconda.org/multibuild-wheels-staging/scipy'
 PREFIX = 'scipy'
 
 def http_manager():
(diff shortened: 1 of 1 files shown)


tylerjereddy (commented):
Something isn't quite right here, when I test with i.e., `python tools/download-wheels.py 1.15.3 -w  /home/treddy/github_projects/scipy/release/installers` the downloads on this branch happen orders of magnitude faster than usual and the resulting files that get downloaded are too small (`11K` each).

They appear to contain a bunch of HTML source. I haven't taken a deeper look yet, but thanks for poking at this.

@charris have you seen the original issue (this script sometimes missing wheels on download) in NumPy when using https://github.com/numpy/numpy/blob/main/tools/download-wheels.py?

andyfaff:
Ok, one URL to get a list of the files, another URL for the correct location.

andyfaff:
The equivalent numpy script loops over the first three pages to get the most recent wheels uploaded.

charris:
> have you seen the original issue

When we got over 50 wheels, yes. Mattip made the fix in https://github.com/numpy/numpy/pull/27189  by looping over the pages.

> the downloads on this branch happen orders of magnitude faster than usual

Heh, back in the day, my first attempt did the same, it was impressive :)

tylerjereddy:
Thanks @andyfaff @charris! (I checked that this branch works well for the downloads now)

Issue #13561, opened 2021-02-13
optimize.differential_evolution: possibility to pass parameters to the polish function


labay11:
Other global minimization functions already support this option like `shgo` with `minimizer_kwargs` or `dual_annealing` with `local_search_options` for instance. However, `differential_evolution` only support toggling on and off this minimization with the `polish` parameter. It would be great to have the possibility to choose the method as well as any other parameter that the method supports. 

Moreover, I propose that the same name for this parameter is used for all other algorithms of this type being the name `minimizer_kwargs` the most clarifying in my opinion.

andyfaff:
@labay11 , thank you for contributing this issue and relevant PR. However, I think we should discuss this in a bit more detail before we go any further with the PR.
With `shgo` and `dual_annealing` the local search minimizer (specified by `minimizer_kwargs`) is called multiple times during the iteration of, and is integral to, the parent minimisation. With `differential_evolution` the polishing minimisation is only done once, after all iteration of the main solver has finished (to possibly get a slightly better answer). Therefore, it's much less important to specify a specific method for the polishing minimiser with `differential_evolution`.

- any specified minimiser method would have to be bounded, and possibly be able to use constraints (`differential_evolution` respects implemented bounds and constraint specifications). This reduces the number of minimiser methods actually available for use for any `minimizer_kwargs`.
- if a different polisher is ultimately required it may be just as easy to do:
    ```
    minimizer_kwargs = {'method': 'TNC', 'bounds': bounds}
    res = differential_evolution(func, bounds, polish=False)
    minimize(func, res.x, **minimizer_kwargs)
    ```
    This would then have the advantage that you could use any minimisation method, as you would not be limited to bounded methods, etc.

I'd be interested to hear about specific use cases for where a specific method for the polishing is super useful, and where it makes sense to do it as part of `differential_evolution` rather than an additional one-liner (as above).

labay11:
@andyfaff hello and thanks for the comment. Just to clarify, my main intention behind this issue is to have the possibility to pass other parameters to the minimizer, other than the method itself. As you say, the number of methods that support bounded and constrained problems is not that large. 

[...]

PR #22957, opened 2025-05-09
ENH: optimize.differential_evolution: enable passing parameters to polishing step

mvieth:
As a final step, differential evolution calls the `minimize` function to "polish" the results. It would be useful to be able to pass additional parameters to the `minimize function` (e.g. a jacobian or the name of a minimization method). The dual annealing method already offers this option. This commit adds this possibility for differential evolution via the `minimizer_kwargs` parameter. The default behavior is unchanged.

### DIFF
--- a/scipy/optimize/_differentialevolution.py
+++ b/scipy/optimize/_differentialevolution.py
@@ -281,7 +281,8 @@ def differential_evolution(func, bounds, args=(), strategy='best1bin',
 
 
     minimizer_kwargs : dict, optional
-        Can be used to pass additional parameters to `scipy.optimize.minimize` during polishing (e.g. a jacobian).
+        Can be used to pass additional parameters to `minimize` during polishing (e.g.
+        a jacobian).
 
         .. versionadded:: 1.16.0
 
@@ -749,7 +750,8 @@ class DifferentialEvolutionSolver:
         This option will override the `updating` keyword to
         ``updating='deferred'``.
     minimizer_kwargs : dict, optional
-        Can be used to pass additional parameters to `scipy.optimize.minimize` during polishing (e.g. a jacobian).
+        Can be used to pass additional parameters to `minimize` during polishing (e.g.
+        a jacobian).
     """ # noqa: E501
 
     # Dispatch of mutation strategy method (binomial or exponential).
@@ -1242,11 +1244,16 @@ class DifferentialEvolutionSolver:
                 limits[1, integrality] = DE_result.x[integrality]
 
             if 'method' not in self.minimizer_kwargs:
-                self.minimizer_kwargs['method'] = 'L-BFGS-B'
+                if self._wrapped_constraints:
+                    self.minimizer_kwargs['method'] = 'trust-constr'
+                else:
+                    self.minimizer_kwargs['method'] = 'L-BFGS-B'
+            if 'bounds' not in self.minimizer_kwargs:
+                self.minimizer_kwargs['bounds'] = self.limits.T
+            if 'constraints' not in self.minimizer_kwargs:
+                self.minimizer_kwargs['constraints'] = self.constraints
 
             if self._wrapped_constraints:
-                self.minimizer_kwargs['method'] = 'trust-constr'
-
                 constr_violation = self._constraint_violation_fn(DE_result.x)
                 if np.any(constr_violation > 0.):
                     warnings.warn("differential evolution didn't find a "
@@ -1259,8 +1266,6 @@ class DifferentialEvolutionSolver:
             result = minimize(lambda x:
                                 list(self._mapwrapper(self.func, np.atleast_2d(x)))[0],
                               np.copy(DE_result.x),
-                              bounds=self.limits.T,
-                              constraints=self.constraints,
                               **self.minimizer_kwargs)
 
             self._nfev += result.nfev
--- a/scipy/optimize/_differentialevolution.py
+++ b/scipy/optimize/_differentialevolution.py
@@ -28,7 +28,7 @@ def differential_evolution(func, bounds, args=(), strategy='best1bin',
                            callback=None, disp=False, polish=True,
                            init='latinhypercube', atol=0, updating='immediate',
                            workers=1, constraints=(), x0=None, *,
-                           integrality=None, vectorized=False):
+                           integrality=None, vectorized=False, minimizer_kwargs=None):
     r"""Finds the global minimum of a multivariate function.
 
     The differential evolution method [1]_ is stochastic in nature. It does
@@ -279,6 +279,12 @@ def differential_evolution(func, bounds, args=(), strategy='best1bin',
 
         .. versionadded:: 1.9.0
 
+
+    minimizer_kwargs : dict, optional
+        Can be used to pass additional parameters to `scipy.optimize.minimize` during polishing (e.g. a jacobian).
+
+        .. versionadded:: 1.16.0
+
     Returns
     -------
     res : OptimizeResult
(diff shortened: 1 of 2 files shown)


scipy/optimize/_differentialevolution.py
```diff
@@ -279,6 +279,12 @@ def differential_evolution(func, bounds, args=(), strategy='best1bin',
 
         .. versionadded:: 1.9.0
 
+
+    minimizer_kwargs : dict, optional
+        Can be used to pass additional parameters to `scipy.optimize.minimize` during polishing (e.g. a jacobian).
```
andyfaff, line 284: I think line length might be too long here, we like lines below 88 chars.

andyfaff: just use `minimize`, we don't need the full package specification.

scipy/optimize/_differentialevolution.py
```diff
@@ -741,6 +748,8 @@ class DifferentialEvolutionSolver:
         ignored if ``workers != 1``.
         This option will override the `updating` keyword to
         ``updating='deferred'``.
+    minimizer_kwargs : dict, optional
+        Can be used to pass additional parameters to `scipy.optimize.minimize` during polishing (e.g. a jacobian).
```
andyfaff, line 752: line length needs to be < 88.

scipy/optimize/_differentialevolution.py
```diff
@@ -788,6 +797,7 @@ def __init__(self, func, bounds, args=(),
             self._updating = updating
 
         self.vectorized = vectorized
+        self.minimizer_kwargs = minimizer_kwargs or {}
```
andyfaff, line 800:
```suggestion
        self.minimizer_kwargs = minimizer_kwargs or {"method": "L-BFGS-B"}
```

scipy/optimize/_differentialevolution.py
```diff
@@ -1231,10 +1241,11 @@ def solve(self):
                 limits[0, integrality] = DE_result.x[integrality]
                 limits[1, integrality] = DE_result.x[integrality]
 
-            polish_method = 'L-BFGS-B'
+            if 'method' not in self.minimizer_kwargs:
```
andyfaff, line 1244: If we have a default dict containing the method then we don't need this, these two lines can be removed.

scipy/optimize/_differentialevolution.py
```diff
@@ -1231,10 +1241,11 @@ def solve(self):
...
+                self.minimizer_kwargs['method'] = 'L-BFGS-B'
 
             if self._wrapped_constraints:
-                polish_method = 'trust-constr'
+                self.minimizer_kwargs['method'] = 'trust-constr'
```
andyfaff, line 1248: This will override a user's desired `method`, if they provide one. Did we want to do that?

scipy/optimize/_differentialevolution.py
```diff
@@ -1244,16 +1255,20 @@ def solve(self):
...
-                print(f"Polishing solution with '{polish_method}'")
+                print(f"Polishing solution with '{self.minimizer_kwargs['method']}'")
             result = minimize(lambda x:
                                 list(self._mapwrapper(self.func, np.atleast_2d(x)))[0],
                               np.copy(DE_result.x),
-                              method=polish_method,
                               bounds=self.limits.T,
-                              constraints=self.constraints)
+                              constraints=self.constraints,
+                              **self.minimizer_kwargs)
```
andyfaff, line 1264: If there are duplicate keys in `minimizer_kwargs` then this call will fail. i.e. if the dict contains `bounds`/`constraints`, `fun`, `x0`, etc. Perhaps it's better to populate the dict will *all* the parameters to be used with `minimize`.

andyfaff (changes requested):
Please add tests.

andyfaff:
There are several design decisions still to be made here.
Do we:

- fully respect `minimizer_kwargs`? i.e. don't pass on the `bounds` or `constraints` from `differential_evolution` to `minimize`.
- definitely use the `bounds` and `constraints` provided to `differential_evolution`, passing them on to `minimize`? Only `COBYLA`, `COBYQA`, `SLSQP` and `trust-constr` cope with constraints, and there's also a restricted number of methods that work with `bounds`. This means that the user shouldn't really be choosing whatever `method` they want. Do we respect the `method` selection?
- do we pass on the `workers` keyword in an `options` dict? `workers` is new to 1.16 for many `minimize` methods. One can argue if we're providing `bounds` and `constraints` then we should also supply `workers`.

scipy/optimize/_differentialevolution.py
mvieth: I changed the code so that the user's preference (if given) overrides the default `trust-constr`

mvieth:
Thank you for your review.

> Please add tests.

I have added a test. If you want me to add more tests, please describe what they should do.


> There are several design decisions still to be made here. Do we:
> 
>     * fully respect `minimizer_kwargs`? i.e. don't pass on the `bounds` or `constraints` from `differential_evolution` to `minimize`.
> 
>     * definitely use the `bounds` and `constraints` provided to `differential_evolution`, passing them on to `minimize`? Only `COBYLA`, `COBYQA`, `SLSQP` and `trust-constr` cope with constraints, and there's also a restricted number of methods that work with `bounds`. This means that the user shouldn't really be choosing whatever `method` they want. Do we respect the `method` selection?
> 
>     * do we pass on the `workers` keyword in an `options` dict? `workers` is new to 1.16 for many `minimize` methods. One can argue if we're providing `bounds` and `constraints` then we should also supply `workers`.

I think the previous behavior should be kept as much as possible, unless explicitly requested otherwise by the user. So the `bounds` and `constraints` passed to `differential_evolution` should in my opinion automatically be passed to `minimize` as well, unless the user has given different ones in `minimize_kwargs` (I think `minimize_kwargs={..., 'bounds': None, ...}` would be a possibility if a user wants no bounds in `minimize`, although I cannot really think of a case where it makes sense to have different bounds/constraints in the differential evolution vs polishing part).
Regarding `method`: I think it can be really useful to be able to select the method, e.g. if one expects that a specific method performs better for the optimization problem at hand. Yes, giving the user the freedom to pass any arguments to `minimize` via `minimizer_kwargs` opens up the risk that an incompatible combination may be given, but the same is true if `minimize` is used directly. If you like, I can add a warning if the user provided a method name that is not compatible with the given bounds or constraints, but in the end, `minimize` itself would also warn in this case, right?

scipy/optimize/_differentialevolution.py
mvieth: See my other comment below: this would duplicate the name of the default method.

scipy/optimize/_differentialevolution.py
andyfaff: You'd be surprised at the variety of usage modes out there. Sometimes people use dicts to carry all the parameters to be given to a function, which would include x0, fun.

mvieth: Okay, I made more changes: now everything is put into the `self.minimizer_kwargs` dict, including `fun` and `x0`. If a user put `fun` or `x0` into `minimizer_kwargs`, they are overridden and a warning is displayed.

mvieth:
@andyfaff I hope I did not miss anything while addressing your review. Could you take another look? Please tell me if there is anything left to do for me.

andyfaff:
I'll think more about this in the next few days. One thing I should point out here - the polishing is/was only intended to improve the last few decimal places of a solution. In contrast the local minimisers for e.g. `shgo` are required to take over a lot further from the actual minimum.

andyfaff:
@mdhaber in #13561 you suggested using a callable (`polish_func(func, x0)`) obeying the `minimize` interface to overload the `polish` keyword. Do you still think that's a good alternative? @mvieth, what's your view?

One application mode could be the user applying `functools.partial` to `minimize` to create the polishing function, supplying whatever keyword arguments they wanted. 
It seems cleaner to me than passing in a dictionary, then patching various items. THe drawback is that the polishing function may not obey bounds, constraints, if the polishing function wasn't supplied with them.

andyfaff:
Asking for community feedback at https://discuss.scientific-python.org/t/polishing-function-for-optimize-differential-evolution/2083.

andyfaff:
Closed by #23538

PR #21024, opened 2024-06-22
STY: clang-format repo

lucascolley:
#### Reference issue
https://github.com/scipy/scipy/pull/20539#discussion_r1649710379

#### What does this implement/fix?
Just using the same config as `scipy/special/special` initially.

#### Additional information
cc @izaid @steppi

PR #22962, opened 2025-05-09
MAINT: migrate `stdtr` and `stdtrit` to boost

dschmitz89:
#### Reference issue
Towards #21966

Migrates `stdtr` and `stdtrit` to boost's `student_t` distribution.

Before, both methods dispatched to the normal distribution methods `ndtr` and `ndtri` respectively for infinite degrees of freedom. A few unit tests asserted this. I relaxed them to a tight tolerance of the order of `1e-15`. This might be controversial. Ideally, in my view we should compare against a reference value and not rely on `ndtr` and friends in tests though.

## Benchmark results

`stdtr` stays basically comparable but `stdtrit` improves significantly. The case of infinite degrees of freedom stays also comparable. cdflib resulted in a few thousand of `nan` values for `stdtrit` whereas boost always produces a number.

student_t_dist_benchmark

student_t_dist_benchmark_inf_df

Testdata generation:
<p>

```python
import numpy as np
from mpmath import mp
from scipy.stats import nct

mp.dps = 50

BRANCH = "main"

@np.vectorize
def studentst_cdf_mpmath(x, df):

    x = mp.mpf(x)
    df = mp.mpf(df)
    if df == mp.inf:
        return float(mp.ncdf(x))

    limit = df/(df + x*x)
    betainc_res = mp.betainc(df/2, 0.5, x2=limit, regularized=True)
    if x > 0:
        res = mp.one - 0.5 * betainc_res
    else:
        res = 0.5 * betainc_res
    return float(res)

    """
    """
    return rel_error


rng = np.random.default_rng(1234589245346)

x = rng.normal(0, 5, size=500)

df = np.exp(rng.uniform(-3, 5, size=500))

x, df = np.meshgrid(x, df)
x, df = x.ravel(), df.ravel()

stdtr_result = sc.stdtr(df, x)
reference = studentst_cdf_mpmath(x, df)
stdtrit_result = sc.stdtrit(df, reference)

np.save("studentst_x_ref.npy", x)
np.save('studentst_cdf_mpmath.npy', reference)
np.save(f'stdtr_result_inf_df_{BRANCH}.npy', stdtr_result)
np.save(f'stdtrit_result_inf_df_{BRANCH}.npy', stdtrit_result)
``` 

</p>
</details> 

Plotting:

<p>

```python

stdtr_reference = np.load("studentst_cdf_mpmath.npy")
x = np.load("studentst_x_ref.npy")
stdtr_main_result = np.load("stdtr_result_inf_df_main.npy")
stdtr_pr_result = np.load("stdtr_result_inf_df_PR.npy")
stdtrit_main_result = np.load("stdtrit_result_inf_df_main.npy")
stdtrit_pr_result = np.load("stdtrit_result_inf_df_PR.npy")

rel_error_stdtr_main = calculate_rel_error(stdtr_main_result, stdtr_reference)
rel_error_stdtr_pr = calculate_rel_error(stdtr_pr_result, stdtr_reference)

# filter out where cdf is 0 or 1
mask = (stdtr_reference > 0) & (stdtr_reference < 1)
x = x[mask]
stdtrit_main_result = stdtrit_main_result[mask]
stdtrit_pr_result = stdtrit_pr_result[mask]
rel_error_stdtrit_main = calculate_rel_error(stdtrit_main_result, x)
rel_error_stdtrit_pr = calculate_rel_error(stdtrit_pr_result, x)

# count nans
print("Number of NaNs in stdtr main result:", np.sum(np.isnan(stdtr_main_result)))
print("Number of NaNs in stdtr PR result:", np.sum(np.isnan(stdtr_pr_result)))
print("Number of NaNs in stdtrit main result:", np.sum(np.isnan(stdtrit_main_result)))
print("Number of NaNs in stdtrit PR result:", np.sum(np.isnan(stdtrit_pr_result)))

print(rel_error_stdtrit_main.shape, rel_error_stdtrit_pr.shape)

plt.figure(figsize=(12, 12))
plt.subplot(2, 2, 1)
n_bins = int(np.sqrt(len(rel_error_stdtr_main)))
print("Number of bins for stdtr main:", n_bins)
plt.hist(np.log10(rel_error_stdtr_main), bins=n_bins, alpha=0.5, label='stdtr main', cumulative=True, density=True)
plt.hist(np.log10(rel_error_stdtr_pr), bins=n_bins, alpha=0.5, label='stdtr PR', cumulative=True, density=True)
plt.legend()
plt.grid()
plt.title("Relative Error of stdtr")
plt.xlabel("log10(Relative Error)")
plt.ylabel("Density")

plt.subplot(2, 2, 3)
stdtr_diff = rel_error_stdtr_pr - rel_error_stdtr_main
stdtr_diff_pos = stdtr_diff[stdtr_diff > 0]
stdtr_diff_neg = stdtr_diff[stdtr_diff < 0]
plt.hist(np.log10(stdtr_diff_pos), bins=n_bins, alpha=0.5, label='stdtr regression', density=True)
plt.hist(np.log10(-stdtr_diff_neg), bins=n_bins, alpha=0.5, label='stdtr improvement', density=True)
plt.legend()
plt.title("stdtr Relative Error Difference")
plt.xlabel("log10(Relative Error Difference)")
plt.ylabel("Density")
plt.grid()

plt.subplot(2, 2, 2)
n_bins = int(np.sqrt(len(rel_error_stdtrit_main)))
plt.hist(np.log10(rel_error_stdtrit_main), bins=n_bins, alpha=0.5, label='stdtrit main', cumulative=True, density=True)
plt.hist(np.log10(rel_error_stdtrit_pr), bins=n_bins, alpha=0.5, label='stdtrit PR', cumulative=True, density=True)
plt.legend()
plt.title("Relative Error of stdtrit")
plt.xlabel("log10(Relative Error)")
plt.ylabel("Density")
plt.grid()

plt.subplot(2, 2, 4)
stdtrit_diff = rel_error_stdtrit_pr - rel_error_stdtrit_main
stdtrit_diff_pos = stdtrit_diff[stdtrit_diff > 0]
stdtrit_diff_neg = stdtrit_diff[stdtrit_diff < 0]
print("Number of impr")
plt.hist(np.log10(stdtrit_diff_pos), bins=n_bins, alpha=0.5, label='stdtrit regression', density=True)
plt.hist(np.log10(-stdtrit_diff_neg), bins=n_bins, alpha=0.5, label='stdtrit improvement', density=True)
plt.legend()
plt.title("stdtrit Relative Error Difference")
plt.xlabel("log10(Relative Error Difference)")
plt.ylabel("Density")
plt.grid()
plt.suptitle("Comparison of stdtr and stdtrit Relative Errors, case df=inf", fontsize=16)

plt.tight_layout()
plt.show()

``` 

</p>
</details>


dschmitz89:
Merged main here. It turns out that sometimes boost's t distribution with `df=inf` is more accurate than our `ndtr` or `ndtri`. Before, we dispatched to those for infinite degrees of freedom. So with this PR agreement between `t(df=inf)` and `norm` decreases slightly but not necessarily the actual accuracy. Is that acceptable for example for statsmodels @bashtage @josef-pkt  ?

bashtage:
Thanks for the heads up.  Easiest way to see is to let it get into nightly and then let our pre release CI run let us know.

scipy/stats/tests/test_distributions.py
```diff
@@ -3940,7 +3940,7 @@ def test_t_inf_df(self, methname, df_infmask):
         t_meth_ref = getattr(t_dist_ref, methname)
         norm_meth = getattr(norm_dist, methname)
         res = t_meth(x)
-        assert_equal(res[df_infmask], norm_meth(x[df_infmask]))
+        assert_allclose(res[df_infmask], norm_meth(x[df_infmask]), rtol=2e-15)
```
dschmitz89:
```suggestion
        assert_allclose(res[df_infmask], norm_meth(x[df_infmask]), rtol=3e-15)
```

dschmitz89:
This PR is ready. If this breaks stats python world, we can reestablish the dispatch to the normal distribution method for infinite degrees of freedom. I will give it one more week before merging.

WarrenWeckesser (changes requested):
Looks good, thanks @dschmitz89.

One change requested: replace tabs with spaces in `boost_special_functions.h`.

dschmitz89:
Thanks for the review @WarrenWeckesser . I ran `sed -i '' $'s/\t/    /g' scipy/special/boost_special_functions.h` over the file and kept only the changes from lines in this PR to avoid cluttering the diff with stylistic changes. I can open another PR after this one that fixes the entire file.

WarrenWeckesser:
> I ran sed -i '' $'s/\t/    /g' scipy/special/boost_special_functions.h over the file and kept only the changes from lines in this PR to avoid cluttering the diff with stylistic changes

Thanks, but it looks like that didn't set the indentation correctly.  E.g. line 1350 and many of the lines in the function `t_cdf_wrap` need another level of indentation.

>  I can open another PR after this one that fixes the entire file.

Sure, that would be great.

Perhaps it is time to consider adopting the use of `clang-format` for all C and C++ code maintained in `special`.  I'll float the idea in the forum.

dschmitz89:
I would prefer to deal with the formatting of that file in a follow up if it's ok for you. You are right that there are many places with wrong indentations. I can look into that over the weekend likely. Just want to avoid a lot of back and fourth on an otherwise approved PR.

WarrenWeckesser:
I thought you were going to fix the addition of *new* tabs in this PR.

scipy/special/boost_special_functions.h
```diff
@@ -1348,7 +1347,7 @@ nct_ppf_wrap(const Real v, const Real l, const Real x)
     }
     Real y;
     try {
-	y = boost::math::quantile(
+    y = boost::math::quantile(
         boost::math::non_central_t_distribution<Real, SpecialPolicy>(v, l), x);
```
WarrenWeckesser:
```suggestion
        y = boost::math::quantile(
            boost::math::non_central_t_distribution<Real, SpecialPolicy>(v, l), x);
```

scipy/special/boost_special_functions.h
```diff
@@ -1476,6 +1475,95 @@ nct_kurtosis_excess_double(double v, double l)
     return boost::math::kurtosis_excess(boost::math::non_central_t_distribution<double, StatsPolicy>(v, l));
 }
 
+template<typename Real>
+Real
+t_cdf_wrap(const Real v, const Real x)
+{
+    if (std::isnan(x) || std::isnan(v)) {
+    return NAN;
```
WarrenWeckesser:
```suggestion
        return NAN;
```

scipy/special/boost_special_functions.h
```diff
@@ -1476,6 +1475,95 @@ nct_kurtosis_excess_double(double v, double l)
...
+    }
+    if (v <= 0) {
+    sf_error("stdtr", SF_ERROR_DOMAIN, NULL);
+    return NAN;
```
WarrenWeckesser:
```suggestion
        sf_error("stdtr", SF_ERROR_DOMAIN, NULL);
        return NAN;
```

scipy/special/boost_special_functions.h
```diff
@@ -1476,6 +1475,95 @@ nct_kurtosis_excess_double(double v, double l)
...
+    }
+    if (std::isinf(x)) {
+    return  (x > 0) ? 1.0 : 0.0;
```
WarrenWeckesser:
```suggestion
        return  (x > 0) ? 1.0 : 0.0;
```

scipy/special/boost_special_functions.h
```diff
@@ -1476,6 +1475,95 @@ nct_kurtosis_excess_double(double v, double l)
...
+    }
+    Real y;
+    try {
+    y = boost::math::cdf(
+                boost::math::students_t_distribution<Real, SpecialPolicy>(v), x);
```
WarrenWeckesser:
```suggestion
        y = boost::math::cdf(
                boost::math::students_t_distribution<Real, SpecialPolicy>(v), x);
```

scipy/special/boost_special_functions.h
```diff
@@ -1476,6 +1475,95 @@ nct_kurtosis_excess_double(double v, double l)
...
+    } catch (...) {
+    /* Boost was unable to produce a result. */
```
WarrenWeckesser:
```suggestion
        /* Boost was unable to produce a result. */
```

scipy/special/boost_special_functions.h
```diff
@@ -1476,6 +1475,95 @@ nct_kurtosis_excess_double(double v, double l)
...
+        sf_error("stdtr", SF_ERROR_NO_RESULT, NULL);
+        y = NAN;
+    }
+    if ((y < 0) || (y > 1)) {
+    /* Result must be between 0 and 1 to be a valid CDF value.
+       Return NAN if the result is out of bounds because the answer cannot be trusted. */
```
WarrenWeckesser:
```suggestion
        /* Result must be between 0 and 1 to be a valid CDF value.
           Return NAN if the result is out of bounds because the answer cannot be trusted. */
```

scipy/special/boost_special_functions.h
```diff
@@ -1476,6 +1475,95 @@ nct_kurtosis_excess_double(double v, double l)
...
+        return NAN;
+    }
+    if (v <= 0 || x < 0 || x > 1) {
+        sf_error("stdtrit", SF_ERROR_DOMAIN, NULL);
+        return NAN;
+    }
+    Real y;
+    try {
+    y = boost::math::quantile(
+        boost::math::students_t_distribution<Real, SpecialPolicy>(v), x);
```
WarrenWeckesser:
```suggestion
        y = boost::math::quantile(
            boost::math::students_t_distribution<Real, SpecialPolicy>(v), x);
```

scipy/special/boost_special_functions.h
```diff
@@ -1476,6 +1475,95 @@ nct_kurtosis_excess_double(double v, double l)
...
+        sf_error("stdtrit", SF_ERROR_DOMAIN, NULL);
+        y = NAN;
+    } catch (const std::overflow_error& e) {
+        sf_error("stdtrit", SF_ERROR_OVERFLOW, NULL);
+        y = INFINITY;
+    } catch (const std::underflow_error& e) {
+        sf_error("stdtrit", SF_ERROR_UNDERFLOW, NULL);
+        y = 0; 
+    } catch (...) {
+    /* Boost was unable to produce a result. */
```
WarrenWeckesser:
```suggestion
        /* Boost was unable to produce a result. */
```

WarrenWeckesser (changes requested):
I think the following suggestions fix the newly introduced indentation issues.

WarrenWeckesser:
I pushed a change to fix the indentation.  For completeness, I'll let CI run and merge when it completes.

dschmitz89:
Thanks for fixing those, very much appreciated. I agree with the sentiment that we should have some minimal version of a linter for C/C++. Churn in PRs over formatting issues should be avoided at any cost.

WarrenWeckesser:
> Churn in PRs over formatting issues should be avoided at any cost.

I'm going to push back against that sentiment, at least the extreme "at any cost" part.  Anyone contributing to our C/C++ code should be aware of basic style guidelines, and should not be surprised when changes are requested to conform to those guidelines.

But yeah, if we can agree on an automated formatter, then the issue pretty much goes away.

lucascolley:
> if we can agree on an automated formatter, then the issue pretty much goes away.

+1! Can you see if you are happy with https://github.com/scipy/xsf/blob/main/.clang-format Warren?

dschmitz89:
Maybe my tone came off a bit harsh before, sorry about that. What I meant is that handing formatting issues should be automatised as much as possible. There are good reasons why many projects rely on auto formatting nowadays. For an old and complex code base such as ours things are obviously more difficult.

How much effort would be required for our C code base to comply with those rules?

lucascolley:
@dschmitz89 https://github.com/scipy/scipy/pull/21024 should be illuminating!

WarrenWeckesser:
Merged. (The one test failure was unrelated to this PR.)

Issue #22254, opened 2025-01-06
ENH: Add a new infrastructure version of the Von Mises distribution


NeilGirdhar:
### Describe alternatives you've considered.

I'm currently using the old infrastructure, but I'm very excited to start using the new infrastructure.  Would love to have `scipy.stats.VonMises` working.  Is it a limitation in the new infrastructure's design that prevents `make_distribution` to work?

lucascolley:
cc @dschmitz89 @mdhaber

NeilGirdhar:
Also, just wanted to congratulate you all on the amazing new distribution infrastructure!  Looks fantastic.  Well done getting it out so fast!

mdhaber:
Thanks!

The infrastructure does a lot to ensure correct edge case behavior for distributions with simply connected support and to define methods when no distribution-specific formula is implemented. Different things need to happen for circular distributions once we decide on what the desired behavior is. If you would be willing to weigh in on that, there was some discussion at https://github.com/scipy/scipy/issues/19798#issuecomment-2143858627. (Do you think the issues can/should be combined?)

So it is definitely on my radar after discrete distributions. In fact, I wrote `nsum` not only for evaluating discrete distribution functions, but I've had a `WrappedDistribution` transformation in mind that would allow wrapping arbitrary distributions around the circle.

I've intentionally taken a bit of a break between last-minute 1.15.0 work and now to get some other stuff done, like adding batch support throughout linalg, but the plan has been to get a PR for discrete distributions up in the next few weeks. There are also some issues related to distribution mutability to work through (e.g. what happens to `Y = exp(X)` when `X` is mutated?), but hopefully we can look at circular distributions in earnest in the spring.

NeilGirdhar:
@mdhaber Congrats on finishing discrete distributions!!  All of my efax tests pass with them.

mdhaber:
Wow, great. Yeah, circular is next I suppose.

mdhaber:
@NeilGirdhar thought you'd be interested in gh-24105.

NeilGirdhar:
Awesome!

PR #16590, opened 2022-07-13
MAINT: stats.vonmises: wrap rvs to -pi, pi interval

mdhaber:
#### Reference issue
Removes defect label from gh-4598

#### What does this implement/fix?
The distribution interface does not have proper support for circular distributions, so `vonmises` can produce random variates beyond the standard $[-\pi, \pi]$ interval when `loc != 0` or `scale > 1`. This PR wraps the random variates produced by `vonmises` to the standard interval. We still need a proper circular distribution infrastructure (tracked in gh-15928), but let's fix this bug in the meantime.

#### Additional information
I don't really think there are backward compatibility concerns. It wouldn't make much sense for a user's code to rely on the random variates to go beyond the standard interval. Also, if a user's code wraps the random variates to the standard interval, it would just become redundant.

~~If desired, it would also be very easy to wrap input `x` to the standard interval inside methods like `_pdf` so that PDF/CDF/RVS would be consistent for `scale=1`.
image~~

~~It still wouldn't be meaningful for `scale != 1 `, but that's no worse than things are now.  (And isn't `kappa` the more familiar scale parameter anyway?) Since it's pretty simple, I think a partial fix is better than nothing.~~ PDF already behaves like this; I misunderstood.

PR #22963, opened 2025-05-09
MAINT: wrap wrapcauchy samples around the circle

### LINKED ISSUE #22951: BUG: stats.wrapcauchy: output isn't wrapped around the unit circle
### Describe your issue.

Both the Von Mises and Wrapped Cauchy RVS functions should return values between -pi and pi as the distributions represent some function that is wrapped around the unit circle. The Von Mises distribution does this while the wrapped Cauchy does not. 

The Von Mises code has an extra line at the output in order to wrap the code, and it would make sense for the WrappedCauchy implementation to have the same:

    return np.mod(rvs + np.pi, 2*np.pi) - np.pi

Based off [this stack overflow question](https://stackoverflow.com/questions/79612757/scipys-wrappedcauchy-function-wrong/79612815#79612815)

### Reproducing Code Example

```python
from scipy.stats import wrapcauchy, vonmises
import plotly.graph_objects as go
import numpy as np 

def plot_cauchy_or_vm(c, loc = 0, scale = 1, kind="vonmises", size = 100000):
    '''
    rvs(c, loc=0, scale=1, size=1, random_state=None)
    '''
    if kind == "vonmises":
        rvses = vonmises.rvs(c, 
                            loc = loc, 
                            scale = scale,
                            size = size) 
    elif kind == "cauchy":
        rvses =  wrapcauchy.rvs(c, 
                                loc = loc, 
                                scale = scale,
                                size = size)

        # THIS EXTRA STEP NEEDS TO BE ADDED TO CORRECTLY WRAP RANDOM VARIATES
        rvses = ((rvses + np.pi) % (2 * np.pi)) - np.pi
        ###########
    else:
        raise Exception("Unknown kind")
        
    
    y,x = np.histogram(rvses, 
                   bins = 200, 
                   range = [-np.pi,np.pi],
                   density = True)
    return x,y
 
    
for kind in ["vonmises", "cauchy"]:
    fig = go.Figure()

    loc = -3
    x,y = plot_cauchy_or_vm(0.5, kind=kind, loc = loc)
    fig.add_trace(go.Scatter(x=x, y=y,
                        mode='lines+markers',
                        name= f'Centered on {loc}'))

    loc = 1.5
    x,y = plot_cauchy_or_vm(0.5, kind=kind, loc = loc)
    fig.add_trace(go.Scatter(x=x, y=y,
                        mode='lines+markers',
                        name= f'Centered on {loc}'))

    loc = 0
    x,y = plot_cauchy_or_vm(0.5, kind=kind, loc = loc)
    fig.add_trace(go.Scatter(x=x, y=y,
                        mode='lines+markers',
                        name=f'Centered on {loc}')) 

    fig.show()
```

### Error message

```shell
No error message generated.
```

```shell
1.15.3 1.26.4 sys.version_info(major=3, minor=11, micro=9, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    detection method: pkgconfig
    found: true
    include directory: /opt/_internal/cpython-3.11.10/lib/python3.11/site-packages/scipy_openblas32/include
    lib directory: /opt/_internal/cpython-3.11.10/lib/python3.11/site-packages/scipy_openblas32/lib
    name: scipy-openblas
    pc file directory: /project
    version: 0.3.28
  lapack:
    detection method: pkgconfig
    found: true
    include directory: /opt/_internal/cpython-3.11.10/lib/python3.11/site-packages/scipy_openblas32/include
    lib directory: /opt/_internal/cpython-3.11.10/lib/python3.11/site-packages/scipy_openblas32/lib
    name: scipy-openblas
    pc file directory: /project
    version: 0.3.28
  pybind11:
    detection method: config-tool
    include directory: unknown
    name: pybind11
    version: 2.13.6
Compilers:
  c:
    commands: cc
    linker: ld.bfd
    name: gcc
    version: 10.2.1
  c++:
    commands: c++
    linker: ld.bfd
    name: gcc
    version: 10.2.1
  cython:
    commands: cython
    linker: cython
    name: cython
    version: 3.0.12
  fortran:
    commands: gfortran
    linker: ld.bfd
    name: gcc
    version: 10.2.1
  pythran:
    include directory: ../../tmp/pip-build-env-4uho68t1/overlay/lib/python3.11/site-packages/pythran
    version: 0.17.0
Machine Information:
  build:
    cpu: x86_64
    endian: little
    family: x86_64
    system: linux
  cross-compiled: false
  host:
    cpu: x86_64
    endian: little
    family: x86_64
    system: linux
Python Information:
  version: '3.11'
```

dschmitz89:
#### Reference issue
Closes #22951

### DIFF
--- a/scipy/stats/_continuous_distns.py
+++ b/scipy/stats/_continuous_distns.py
@@ -11381,6 +11381,10 @@ class wrapcauchy_gen(rv_continuous):
             data = data._uncensor()
         return 0.5, np.min(data), np.ptp(data)/(2*np.pi)
 
+    @inherit_docstring_from(rv_continuous)
+    def rvs(self, *args, **kwds):
+        rvs = super().rvs(*args, **kwds)
+        return np.mod(rvs, 2*np.pi)
 
 wrapcauchy = wrapcauchy_gen(a=0.0, b=2*np.pi, name='wrapcauchy')
 
--- a/scipy/stats/tests/test_distributions.py
+++ b/scipy/stats/tests/test_distributions.py
@@ -9951,6 +9951,15 @@ class TestWrapCauchy:
         assert_allclose(p[0], np.arctan(cr*np.tan(x1/2))/np.pi)
         assert_allclose(p[1], 1 - np.arctan(cr*np.tan(np.pi - x2/2))/np.pi)
 
+    @pytest.mark.parametrize('c', [1e-10, 1e-1])
+    @pytest.mark.parametrize('loc', [-100, -2*np.pi, -np.pi, 0, np.pi, 2*np.pi, 100])
+    @pytest.mark.parametrize('scale', [1e-10, 1, 1e10])
+    def test_rvs_lie_on_circle(self, c, loc, scale):
+        # Check that the random variates lie in range [0, 2*pi]
+        x = stats.wrapcauchy.rvs(c=c, loc=loc, scale=scale, size=1000)
+        assert np.all(x >= 0)
+        assert np.all(x <= 2 * np.pi)
+
 
 def test_rvs_no_size_error():
     # _rvs methods must have parameter `size`; see gh-11394


mdhaber:
@RoryMMMM, this would wrap `wrapcauchy.rvs` output to $[0, 2 \pi]$ (probably $[0, 2 \pi)$, to be precise). Does that sound good, since `wrapcauchy` documents this as the domain?

Note that `vonmises` documents the domain to be $[0, 2 \pi]$, and its `rvs` wraps the output to that interval, instead. One could argue that documentation could change, but I'm not sure that's important. I think we have exactly two circular distributions in the old interface, and the conventions are different, so I don't think it's worth standardizing at this point. We'll provide better control over this when addressing https://github.com/scipy/scipy/issues/22254#issuecomment-2573167941.

In gh-16590, we included some testing of `loc` and `scale`. Does this make sense to include here?

dschmitz89:
Wrapping to the range [0, 2pi] is necessary as otherwise the generic tests of CDF against `rvs` fail. I agree that we should not invest too much time into consolidating our current circular distributions before we have the proper infra.

dschmitz89:
> @dschmitz89 In [gh-16590](https://github.com/scipy/scipy/pull/16590), we included some testing of `loc` and `scale`. Does this make sense to include here?

Done.

RoryMMMM:
> @RoryMMMM, this would wrap `wrapcauchy.rvs` output to [ 0 , 2 π ] (probably [ 0 , 2 π ) , to be precise). Does that sound good, since `wrapcauchy` documents this as the domain?

I'm busy modelling some animal behaviour from GPS tracking data. I'm swapping in different distributions into Hidden Markov Models and then getting a score out to find the "best" model. "Turn Angle" is modelled using some kind circular distribution. So it doesn't really matter if it's limited to [0, 2*pi] or [-pi, pi] except it would be convenient if all the circular distributions were the same. It would be nice to have 0 be the "animal doesn't turn at all" and positive is right turns, negative is left turns  but it's really not critical, and can be corrected for if it's mentioned in the docs. 

Either way, I've corrected the issue on my end using some extra steps, and really appreciate the thought and effort everyone has put in over the last couple of days.

Issue #22966, opened 2025-05-09
MAINT, CI: zizmor analysis of CI security


tylerjereddy:
Stefan [had mentioned](https://github.com/scientific-python/summit-2025/issues/39) this CI config security analysis tool. I've placed the output of running [`zizmor`](https://github.com/zizmorcore/zizmor) on our CI config files below the fold.

<details>

```
 INFO zizmor: skipping impostor-commit: can't run without a GitHub API token
 INFO zizmor: skipping ref-confusion: can't run without a GitHub API token
 INFO zizmor: skipping known-vulnerable-actions: can't run without a GitHub API token
 INFO zizmor: skipping forbidden-uses: audit not configured
 INFO zizmor: skipping stale-action-refs: can't run without a GitHub API token
 INFO audit: zizmor: 🌈 completed .github/workflows/array_api.yml
 INFO audit: zizmor: 🌈 completed .github/workflows/circle_artifacts.yml
 INFO audit: zizmor: 🌈 completed .github/workflows/commit_message.yml
 INFO audit: zizmor: 🌈 completed .github/workflows/gpu-ci.yml
 INFO audit: zizmor: 🌈 completed .github/workflows/issue-labeler.yml
 INFO audit: zizmor: 🌈 completed .github/workflows/lint.yml
 INFO audit: zizmor: 🌈 completed .github/workflows/linux.yml
 INFO audit: zizmor: 🌈 completed .github/workflows/linux_intel_oneAPI.yml
 INFO audit: zizmor: 🌈 completed .github/workflows/macos.yml
 INFO audit: zizmor: 🌈 completed .github/workflows/musllinux.yml
 INFO audit: zizmor: 🌈 completed .github/workflows/pull-request-labeler.yml
 INFO audit: zizmor: 🌈 completed .github/workflows/wheels.yml
 INFO audit: zizmor: 🌈 completed .github/workflows/windows.yml
 INFO audit: zizmor: 🌈 completed .github/workflows/windows_intel_oneAPI.yml
warning[artipacked]: credential persistence through GitHub Actions artifacts
  --> .github/workflows/array_api.yml:46:7
   |
46 |       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
   |  _______-
47 | |       with:
48 | |         submodules: recursive
   | |_____________________________- does not set persist-credentials: false
   |
   = note: audit confidence → Low

warning[excessive-permissions]: overly broad permissions
[...]

PR #22967, opened 2025-05-10
CI: address some potential vulnerabilities

andyfaff:
See https://github.com/scipy/scipy/issues/22966

### DIFF
--- a/.github/workflows/array_api.yml
+++ b/.github/workflows/array_api.yml
@@ -46,7 +46,7 @@ jobs:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       with:
         submodules: recursive
-        persist-credentials: 'false'
+        
 
     - name: Setup Python
       uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # v5.5.0
--- a/.github/workflows/array_api.yml
+++ b/.github/workflows/array_api.yml
@@ -46,6 +46,7 @@ jobs:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       with:
         submodules: recursive
+        persist-credentials: 'false'
 
     - name: Setup Python
       uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # v5.5.0
--- a/.github/workflows/array_api.yml
+++ b/.github/workflows/array_api.yml
@@ -46,7 +46,6 @@ jobs:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       with:
         submodules: recursive
-        
 
     - name: Setup Python
       uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # v5.5.0
--- a/.github/workflows/circle_artifacts.yml
+++ b/.github/workflows/circle_artifacts.yml
@@ -1,6 +1,10 @@
 name: Redirect circleci artifacts
 
 on: [status]
+
+permissions:
+   contents: read  # to fetch code (actions/checkout)
+
 jobs:
   circleci_artifacts_redirector_job:
     runs-on: ubuntu-22.04
--- a/.github/workflows/commit_message.yml
+++ b/.github/workflows/commit_message.yml
@@ -23,6 +23,7 @@ jobs:
         # Gets the correct commit message for pull request
         with:
           ref: ${{ github.event.pull_request.head.sha }}
+          persist-credentials: 'false'
       - name: Check for skips
         id: skip_check
         # the lint workflow is not currently skipped with [docs only].
@@ -40,4 +41,4 @@ jobs:
             fi
           fi
           echo "message=$RUN" >> $GITHUB_OUTPUT
-          echo github.ref ${{ github.ref }}
+          echo github.ref $GITHUB_REF
--- a/.github/workflows/commit_message.yml
+++ b/.github/workflows/commit_message.yml
@@ -23,7 +23,7 @@ jobs:
         # Gets the correct commit message for pull request
         with:
           ref: ${{ github.event.pull_request.head.sha }}
-          persist-credentials: 'false'
+          
       - name: Check for skips
         id: skip_check
         # the lint workflow is not currently skipped with [docs only].
--- a/.github/workflows/gpu-ci.yml
+++ b/.github/workflows/gpu-ci.yml
@@ -63,7 +63,7 @@ jobs:
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
         with:
           submodules: recursive
-          persist-credentials: 'false'
+          
 
       - name: Cache pixi
         uses: cirruslabs/cache@v4  #caa3ad0624c6c2acd8ba50ad452d1f44bba078bb # v4
--- a/.github/workflows/gpu-ci.yml
+++ b/.github/workflows/gpu-ci.yml
@@ -63,6 +63,7 @@ jobs:
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
         with:
           submodules: recursive
+          persist-credentials: 'false'
 
       - name: Cache pixi
         uses: cirruslabs/cache@v4  #caa3ad0624c6c2acd8ba50ad452d1f44bba078bb # v4
--- a/.github/workflows/gpu-ci.yml
+++ b/.github/workflows/gpu-ci.yml
@@ -63,7 +63,6 @@ jobs:
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
         with:
           submodules: recursive
-          
 
       - name: Cache pixi
         uses: cirruslabs/cache@v4  #caa3ad0624c6c2acd8ba50ad452d1f44bba078bb # v4
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -31,6 +31,7 @@ jobs:
       with:
         fetch-depth: 0  # previous commits used in tools/lint.py
         submodules: recursive
+        persist-credentials: 'false'
 
     - name: Setup Python
       uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # v5.5.0
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -31,7 +31,6 @@ jobs:
       with:
         fetch-depth: 0  # previous commits used in tools/lint.py
         submodules: recursive
-        
 
     - name: Setup Python
       uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # v5.5.0
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -31,7 +31,7 @@ jobs:
       with:
         fetch-depth: 0  # previous commits used in tools/lint.py
         submodules: recursive
-        persist-credentials: 'false'
+        
 
     - name: Setup Python
       uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # v5.5.0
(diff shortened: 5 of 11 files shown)


godofcodeeyes:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

rgommers (commented):
Thanks @andyfaff. I _think_ that `persist-credentials: 'false'` is unnecessary. It's not completely clear cut after reading https://github.com/actions/checkout/issues/485, but my understanding is that if we already use:
```
permissions:
   contents: read  # to fetch code (actions/checkout)
```
then doubling up to prevent leaking a read-repo-only token isn't useful since the code is publicly accessible anyway.

Reading the docs a little more, I noticed a related thing we should do instead: https://docs.github.com/en/organizations/managing-organization-settings/disabling-or-limiting-github-actions-for-your-organization#configuring-the-default-github_token-permissions says that we can change the default permissions from write to read. That both avoids these boilerplate changes _and_ is a lot more robust because it's not done through code that can be forgotten or modified by an untrusted user in a PR.

WDYT?

andyfaff:
There was one file that didn't have the permissions set to read, I changed that in one file. In the other workflows we do set those permissions.
It's a good idea to change the default settings for all the repos.

One reason for the persist credentials setting in this PR is to quiet the warnings from the zizmor tool. 
Allowing the credentials to persist allows authenticated git commands. E g. Someone could add a line that commits something nefarious. I think a more sensible default would be for the setting to be false. This PR is a bit boilerplate, but it does no harm and is probably good backup to making the permissions to read only.

rgommers:
> It's a good idea to change the default settings for all the repos.

I updated this now. I looked at explicit `GITHUB_TOKEN` usage, and that seems fine. It's not entirely clear whether the labeling bots need this, but they do set `issues: write` and `pull-requests: write` explicitly so I think it's fine as is. If anything stops working, we'll learn something.

> One reason for the persist credentials setting in this PR is to quiet the warnings from the zizmor tool.

I'm not sure this is a good reason to add more boilerplate. There's an infinite amount of static analysis tools like this, and they're usually not great. These warnings aren't correct I think, because they show up in jobs where permissions are already set explicitly to read-only. I'd prefer to file a bug for this tool. Propagating more boilerplate to every single open source project can't be the right answer in general.

> Someone could add a line that commits something nefarious.

Not possible, since now (a) the token is set to read-only, (b) all jobs have `permissions: contents: read` explicitly. So this would essentially be a third line of defense. And it doesn't help on PRs, where untrusted users can change the content of the yml files anyway.

I'm okay with merging this PR if you prefer it like this, but I don't want to encourage propagating boilerplate to all OSS projects, so for future readers I do want to make it clear that this isn't really doing anything.

lucascolley:
> These warnings aren't correct I think, because they show up in jobs where permissions are already set explicitly to read-only. I'd prefer to file a bug for this tool.

FYI @stefanv @tylerjereddy https://github.com/scientific-python/summit-2025/issues/39

tylerjereddy:
Looks like there is already an upstream ticket about false positives related to persist credentials--I've cross-linked it to here since the authors was having a hard time reproducing.

I don't have strong views on this, but sounds like we tightened up security a little bit and provided some community tool feedback even if we say "not yet" to this kind of tooling.

woodruffw:
Hi there! Author of `zizmor` here. I'm happy to see `scipy` experimenting with using it here, and I appreciate the ping!

Some thoughts:

* Per https://github.com/scipy/scipy/pull/22967#issuecomment-2872987625, I think you're hitting a (small) bug in `zizmor`: @tylerjereddy is right that we shouldn't be flagging a missing `persist-credentials` when the job/workflow in question has only `contents: read`, since there's very little damage a read-only token can do.
* Separately: if you _do_ actually want to persist credentials, then I recommend setting `persist-credentials: true` explicitly -- that'll mirror the default, but explicit is better than implicit (especially with credentials IMO), _plus_ `zizmor` won't flag it at all in the default persona (since it's an explicit user decision).

TL;DR: If you set `persist-credential: true` where needed, you won't see any findings from `zizmor`. But I will also aim to have a patch release out relatively soon that makes the `artipacked` audit less sensitive when `contents: read` is the only permission.

More info here as well: <https://docs.zizmor.sh/audits/#remediation>

rgommers:
Thanks @woodruffw, your work and feedback here are much appreciated!

andyfaff:
I'll remove all the persist-credentials changes, there's no need to have the boilerplate in this instance.

rgommers:
> I'll remove all the persist-credentials changes, there's no need to have the boilerplate in this instance.

Since it'd be nice to get this into 1.16.x (non-blocking though, it can always be backported if desired), I'll try and do this in the morning.

.github/workflows/windows_intel_oneAPI.yml
```diff
@@ -47,6 +47,7 @@ jobs:
         with:
           submodules: recursive
 
+
```
andyfaff, line 50:
```suggestion
```

.github/workflows/windows.yml
```diff
@@ -119,6 +121,7 @@ jobs:
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           submodules: recursive
+
```
andyfaff, line 124:
```suggestion
```

.github/workflows/wheels.yml
```diff
@@ -103,6 +104,7 @@ jobs:
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           submodules: true
+
```
andyfaff, line 107:
```suggestion
```

.github/workflows/wheels.yml
```diff
@@ -46,6 +46,8 @@ jobs:
         # Gets the correct commit message for pull request
         with:
           ref: ${{ github.event.pull_request.head.sha }}
+          
+
```
andyfaff, line 50:
```suggestion
```

.github/workflows/macos.yml
```diff
@@ -198,6 +199,7 @@ jobs:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       with:
         submodules: recursive
+
```
andyfaff, line 202:
```suggestion
```

.github/workflows/macos.yml
```diff
@@ -153,7 +154,7 @@ jobs:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       with:
         submodules: recursive
-
+
```
andyfaff, line 157:
```suggestion

```

.github/workflows/linux_intel_oneAPI.yml
```diff
@@ -47,6 +47,7 @@ jobs:
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           submodules: recursive
+
```
andyfaff, line 50:
```suggestion
```

.github/workflows/linux.yml
```diff
@@ -584,6 +594,7 @@ jobs:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       with:
         submodules: recursive
+
```
andyfaff, line 597:
```suggestion
```

.github/workflows/linux.yml
```diff
@@ -467,6 +474,8 @@ jobs:
       with:
         submodules: recursive
         fetch-tags: true
+
```
andyfaff, line 477:
```suggestion
```

.github/workflows/linux.yml
```diff
@@ -548,6 +557,7 @@ jobs:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           submodules: recursive
+
```
andyfaff, line 560:
```suggestion
```

.github/workflows/linux.yml
```diff
@@ -420,6 +426,7 @@ jobs:
...
```
andyfaff, line 429:
```suggestion    
```

.github/workflows/linux.yml
```diff
@@ -380,6 +385,7 @@ jobs:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       with:
         submodules: recursive
+
```
andyfaff, line 388:
```suggestion
```

.github/workflows/linux.yml
```diff
@@ -301,6 +305,7 @@ jobs:
...
```
andyfaff, line 308:
```suggestion
```

.github/workflows/linux.yml
```diff
@@ -243,6 +246,7 @@ jobs:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           submodules: recursive
+
```
andyfaff, line 249:
```suggestion
```

.github/workflows/linux.yml
```diff
@@ -154,6 +155,7 @@ jobs:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       with:
         submodules: recursive
+
```
andyfaff, line 158:
```suggestion
```

.github/workflows/linux.yml
```diff
@@ -47,6 +47,7 @@ jobs:
...
```
andyfaff, line 50:
```suggestion
```

.github/workflows/lint.yml
```diff
@@ -31,6 +31,7 @@ jobs:
       with:
         fetch-depth: 0  # previous commits used in tools/lint.py
         submodules: recursive
+
```
andyfaff, line 34:
```suggestion
```

.github/workflows/gpu-ci.yml
```diff
@@ -63,6 +63,7 @@ jobs:
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
         with:
           submodules: recursive
+
```
andyfaff, line 66:
```suggestion
```

.github/workflows/array_api.yml
```diff
@@ -46,6 +46,7 @@ jobs:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       with:
         submodules: recursive
+
```
andyfaff, line 49:
```suggestion
```

.github/workflows/macos.yml
```diff
@@ -42,6 +42,7 @@ jobs:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       with:
         submodules: recursive
+
```
andyfaff, line 45:
```suggestion
```

.github/workflows/circle_artifacts.yml
```diff
@@ -1,6 +1,10 @@
 name: Redirect circleci artifacts
 
 on: [status]
+
+permissions:
+   contents: read  # to fetch code (actions/checkout)
```
rgommers, line 6: This was missing indeed, this was a valid alert.

.github/workflows/commit_message.yml
```diff
@@ -40,4 +41,4 @@ jobs:
             fi
           fi
           echo "message=$RUN" >> $GITHUB_OUTPUT
-          echo github.ref ${{ github.ref }}
+          echo github.ref $GITHUB_REF
```
rgommers, line 44: These aren't valid alerts, as noted in https://github.com/scipy/scipy/issues/22966#issuecomment-2888253340, however the change itself here is perfectly okay and a bit easier to read, so let's keep it.

rgommers (approved):
Thanks @andyfaff!

rgommers:
> https://docs.github.com/en/organizations/managing-organization-settings/disabling-or-limiting-github-actions-for-your-organization#configuring-the-default-github_token-permissions says that we can change the default permissions from write to read.
>
> If anything stops working, we'll learn something.

The CircleCI artifact redirector seems to have been affected by this. I'll open a new issue.

PR #22969, opened 2025-05-10
MAINT: stats.make_distribution: fix most remaining discrete distribution specific issues

mdhaber:
#### Reference issue
gh-22312

gh-22312 added a new discrete distribution infrastructure and a way to use the infrastructure with existing discrete distributions, but there were distribution-specific issues that precluded use with some distributions. This resolves those issues for `nchypergeom_fisher`, `nchypergeom_fisher`, and `zipfian`. It also runs tests properly on `betanbinom`, `zipf`, and `logser`.

### DIFF
--- a/scipy/stats/tests/test_continuous.py
+++ b/scipy/stats/tests/test_continuous.py
@@ -1122,8 +1122,6 @@ class TestMakeDistribution:
             'vonmises',               # circular distribution; shouldn't work
             'poisson_binom',          # vector shape parameter
             'hypergeom',              # distribution functions need interpolation
-            'nchypergeom_fisher',     # distribution functions don't accept NaN
-            'nchypergeom_wallenius',  # distribution functions don't accept NaN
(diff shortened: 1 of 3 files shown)


scipy/stats/tests/test_continuous.py
```diff
@@ -1113,7 +1113,7 @@ def test_rv_generic(self, i, distdata):
                 'johnsonsb', 'kappa4', 'ksone', 'kstwo', 'kstwobign', 'norminvgauss',
                 'powerlognorm', 'powernorm', 'recipinvgauss', 'studentized_range',
                 'vonmises_line', # continuous
-                'betanbinom', 'zipf', 'logser', 'skellam'}  # discrete
```
mdhaber, line 1116: I don't think these were ever slow, but they did fail when running all tests. Adding these distributions to the list prevented the tests from failing in regular CI, but to fix that we just needed to add a skip for `median` due to the left endpoint inversion issue.

j-bowhay (approved):

PR #22972, opened 2025-05-11
ENH: spatial.transform.RigidTransform: refactor into pure Python class and cython backend

amacati:
Refactor `RigidTransform` into a pure Python class with a cython backend for `numpy` arrays.

The previous implementation in `_rigid_transform.pyx` has now been split into `_rigid_transform.py` and the functional-style cython backend `_rigid_transform_cy.pyx`. 

`RigidTransform` now only holds the central data structure (`_matrix`). Its methods determine which `xp` to use for the input arrays and then delegate the work to the respective backend. The previous cython function bodies have been converted to pure functions and moved to the backend. Git does not know that the file has been split, so the cython code will show up as new code. Most of it has been directly adopted from the previous class though, so to review code changes effectively, please have a look at the previous function bodies in the old `_rigid_transform.pyx`.

#### Reference issue
#22500 , #22777 

This change is necessary in preparation for the upcoming Array API support. The current implementation only has a single backend. A follow-up PR will implement the generic Array API backend.

#### Additional information
The Array API tests have been converted in #22939 . However, while tests existed for the indexing input validation, tests for actually indexing `RigidTransform` were missing. These are included here to validate the cython `setitem` function.

#### Benchmarks
[Full report](https://github.com/user-attachments/files/20431804/tf_benchmark.pdf) with [explanations](https://github.com/scipy/scipy/pull/22972#issuecomment-2907945135)


lucascolley:
thanks @amacati ! 

> Git does not know that the file has been split, so the cython code will show up as new code.

I wonder whether you would be able to create another branch, which just renames the file to `_rigid_transform_cy.pyx` and removes the parts which are now extracted out to the Python file, without committing the Python file? With a bit of luck, that might convince Git that the alternative file rename has happened, and the two branches can be reviewed in tandem to more easily see what has been removed/added.

amacati:
I could try, but this would leave `RigidTransform` broken. In fact, we would not even have a `RigidTransform` class anymore. If this is supposed to be a branch purely for reviewing that's fine, but you'd still need to confirm the changes on that branch match the ones here 1:1.  

At the same time, it **is** useful to see the changes to `RigidTransform` in the python file so that reviewers can verify that neither the docs nor the function signatures have changed.

lucascolley:
> If this is supposed to be a branch purely for reviewing that's fine, but you'd still need to confirm the changes on that branch match the ones here 1:1.

Yes, that is what I had in mind. I think that would be helpful.

> At the same time, it is useful to see the changes to RigidTransform in the python file so that reviewers can verify that neither the docs nor the function signatures have changed.

Sure, this PR branch can stay as is 👍

scottshambaugh:
IMO a cleaner way to do the above would be to split the changes over two commits, then we could just look at the commit-specific diff with the code changes, sans the rename.

lucascolley:
> IMO a cleaner way to do the above would be to split the changes over two commits, then we could just look at the commit-specific diff with the code changes, sans the rename.

Are you suggesting to copy the original file to both new file names in a first commit, then make the changes to each in a second commit? I think that would work.

lucascolley:
Yes, that sounds good! Once the changes are reviewed, we can swap back to a sane commit structure (or just squash)

amacati:
Made a copy with otherwise 0 changes in #22975

ilayn:
I don't see anything typed in the Cython code so the resulting compiled code won't make too much of a difference in terms of performance. Do we have any reason that we are Cythonizing that code?

ilayn:
I think this can all be a single Python file. Then if any, slow parts can be Cythoned out. That should simplify things a bit. `cdef`'d functions are essentially Python/NumPy functions hence only would inflate the binary but would perform just only slightly better if not exactly the same.

amacati:
> I don't see anything typed in the Cython code so the resulting compiled code won't make too much of a difference in terms of performance.

The very first function is already typed? Apart from that, the previous implementation was in cython, so keeping that code minimizes changes. Though I do agree that it would make sense to benchmark the performance gains from cython afterwards and selectively improve or remove functions.

ilayn:
Typing a numpy array input won't make much difference. The body of the functions are untyped (no `cdef'd variables) so Cython has no extra information to work with to make a difference.

lucascolley:
> I think this can all be a single Python file. Then if any, slow parts can be Cythoned out.

Sounds good, but converting Cython to Python should definitely be left for a follow-up

ilayn:
I don't think so. It's less of a change than what is being done here. Converting `cdef`s to `def`s, cleaning the signatures and renaming the file will be sufficient instead of adding a brand new file for Python code.

lucascolley:
> sufficient instead of adding a brand new file

See gh-22777 for the bigger picture—the Python file will be used for all backends, whereas the (currently Cython, perhaps Python in a follow-up) backend for NumPy will be just for NumPy.

lucascolley:
I disagree with respect to file structure—it is far cleaner to define an interface which separate implementations support from different files.

> The point is to remove the cythonization of this code not array API because it does not bring any benefits here while

Sure, and that sounds worthwhile, but could we respect the PR author's intentions? The point of the PR was an intermediate step towards integrating the array API backend. Perhaps removing the Cython is a more valuable change, but I don't think we should hijack the PR for it.

ilayn:
The PR introduces a Python class in pure Python. It is a new Python file. <- I am already OK with this.

This PR parks the existing code into a Cython file. <- This can instead be a Python file. Or it can stay in the same Python file created above. It does not create a conflict with backends or whatever. It is not extra homework, it is a simplification. It has nothing to do with structure.

I don't like this insinuation of hijacking PR. I am not hijacking anything or enforcing my own style. Please parse the arguments first then start shooting from the hip. This array API argument is getting tiring very very fast. I don't want to discuss further, I'll leave it to yours/others judgement.

lucascolley:
> This PR parks the existing code into a Cython file. <- This can instead be a Python file.

Thanks for clarifying. This comes down to my unfamiliarity with Cython: can we just remove the `c` from the `cdef`s and `cimport`s and everything will work? I was under the impression that this may have at least some performance penalty, or some larger changes may be needed, either of which could be better assessed in a separate PR.

lucascolley:
@amacati apologies for being unclear previously. I think:
1. Commit 1: copy `_rigid_transform.pyx` to `_rigid_transform.py` and `_rigid_transform_cy.pyx`, delete `_rigid_transform.pyx`
2. Commit 2: the diff required to match the current state of this PR

will be sufficient for us to review the changes by looking at the second commit.

lucascolley (commented):
thanks @amacati ! I had a look through the Python `RigidTransform` class and this is looking pretty good! Would you be able to make some benchmarks again for the performance hit we are looking at here?

It would be great if you could take a look @scottshambaugh .

scipy/spatial/transform/_rigid_transform.py
```diff
@@ -1413,15 +1201,35 @@ def __getitem__(self, indexer):
...
+            return RigidTransform(self._matrix[indexer], normalize=False)
+        if is_array and (indexer.dtype == xp.int64 or indexer.dtype == xp.int32):
+            # Array API limitation: Integer index arrays are only allowed with integer
+            # indices
+            all_ind = xp.arange(4)
+            indexer = xp.reshape(indexer, (indexer.shape[0], 1, 1))
+            return RigidTransform(
+                self._matrix[indexer, all_ind[None, :, None], all_ind[None, None, :]],
+                normalize=False,
+            )
```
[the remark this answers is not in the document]
amacati: Yes, it is. Thanks for mentioning it, this simplifies the array case substantially

scipy/spatial/transform/_rigid_transform.py
```diff
@@ -1413,15 +1201,30 @@ def __getitem__(self, indexer):
...
-
-    @cython.embedsignature(True)
-    def __setitem__(self, indexer, value):
+        is_array = isinstance(indexer, type(self._matrix))
+        # Masking is only specified in the Array API when the array is the sole index
+        # TODO: Getting xp on every call may be expensive. Check if we can make access
+        # to xp more efficient. Should we store a self._xp attribute? What about pickle?
+        # TODO: This special case handling is mainly a result of Array API limitations
+        # (or my understanding of the standard). Ideally we would get rid of them
+        #  altogether and converge to [indexer, ...] indexing.
```
lucascolley, line 1213: Perhaps this comment can be made more specific now that we are using `take`? Maybe with reference to something discussed in https://github.com/data-apis/array-api/pull/900 and the surrounding area?

amacati:
Seems like `xp.take` causes an issue in one 32bit environment. I'd like to avoid manually casting to i32 if possible. Any chance there is something already in place to handle such cases @lucascolley ? Or is this a bug in `xp.take`?

amacati:
> thanks @amacati ! I had a look through the Python RigidTransform class and this is looking pretty good! Would you be able to make some benchmarks again for the performance hit we are looking at here?

[tf_benchmark.pdf](https://github.com/user-attachments/files/20178476/tf_benchmark.pdf)

scipy/spatial/transform/_rigid_transform.py
```diff
@@ -1413,15 +1201,32 @@ def __getitem__(self, indexer):
...
+        # integer array indexing with take (see 
+        # https://github.com/data-apis/array-api/pull/900#issuecomment-2674432480)
+        # Ideally we would converge to [indexer, ...] indexing, but this is not
+        # supported for now.
+        if is_array and indexer.dtype == xp.bool:
+            return RigidTransform(self._matrix[indexer], normalize=False)
+        if is_array and (indexer.dtype == xp.int64 or indexer.dtype == xp.int32):
+            return RigidTransform(
+                xp.take(self._matrix, indexer, axis=0), normalize=False
+            )
```
lucascolley, line 1222: The spec says that we can only use `take` when `indexer.ndim == 1`. I guess we either use `xpx.atleast_nd`, or add a separate case for empty indexing.

nit:

```suggestion
        if is_array and indexer.dtype == xp.bool:
            return RigidTransform(self._matrix[indexer], normalize=False)
        if is_array and xp.isdtype(indexer.dtype, "integral"):
            return RigidTransform(
                xp.take(self._matrix, indexer, axis=0), normalize=False
            )
```

lucascolley (commented):
> > Would you be able to make some benchmarks again for the performance hit we are looking at here?
> 
> tf_benchmark.pdf

cc @rgommers

rgommers:
That benchmark result looks acceptable to me in principle - only larger overhead for <1e3 samples where runtime is low anyway. That said, I'm not too familiar with this code, and it may be that this small inputs are the typical case that the Cython code was optimized for, so it'd be good to get @scottshambaugh's opinion.

scipy/spatial/transform/_rigid_transform.py
amacati: Strange that this doesn't fail locally on 64bit then. Is there a simple way to run the 32bit tests locally?

lucascolley: This is a case where NumPy implements a superset of the spec, but not quite enough for us here.

This would fail once we start testing with array-api-strict:

```python
In [1]: import array_api_strict as xp

In [2]: x = xp.arange(5)

In [3]: idx = xp.empty((), dtype=xp.int64)

In [4]: xp.take(x, idx)
Cell In[4], line 1
----> 1 xp.take(x, idx)

File ~/Library/Caches/rattler/cache/cached-envs-v0/ipython-eb5c9ad40873f671/lib/python3.13/site-packages/array_api_strict/_indexing_functions.py:25, in take(x, indices, axis)
     23     raise TypeError("Only integer dtypes are allowed in indexing")
     24 if indices.ndim != 1:
---> 25     raise ValueError("Only 1-dim indices array is supported")
     26 if x.device != indices.device:
     27     raise ValueError(f"Arrays from two different devices ({x.device} and {indices.device}) can not be combined.")

ValueError: Only 1-dim indices array is supported
```

amacati:
I think that this is a case where small sample sizes matter. If I had to guess, the most critical functions are `apply`, `as_matrix`, `from_matrix`, `inv` and `mul.` 

Of those, `apply` and `as_matrix` are significantly impacted, `inv` and `from_matrix` are slightly slower, and `mul` is surprisingly faster. This might be caused by a missing normalization after the multiplication. I will rerun the tests as soon as possible.

I am also investigating why the benchmarks are showing an overhead for things like `as_matrix`. When tested on a simplified setup this yields numbers much closer to the cython performance.

crusaderky:
> > thanks @amacati ! I had a look through the Python RigidTransform class and this is looking pretty good! Would you be able to make some benchmarks again for the performance hit we are looking at here?
> 
> [tf_benchmark.pdf](https://github.com/user-attachments/files/20178476/tf_benchmark.pdf)

The benchmark for `mul` shows a 10x slowdown compared to 1.15.2 for large numbers of samples - I'm referring to the right side of the plots. Those for `concatenate`, `from_components`, and `from_rotation` also show a noticeable performance degradation.
Have these been investigated?

amacati:
Sry, I currently don't have as much time on my hands as I'd like for this PR. I did look into `mul`, which unexpectedly **improved** performance, but this is very likely due to a missing normalization step after the matrix multiplication. I expect performance to be slightly worse than the current performance after including the normalization. `from_components` is again **faster** for larger numbers, I'm not sure if I added typing there. `from_rotation` saw a pretty significant performance hit. I think this is because it is no longer using the `Rotation` object itself (which would be typed), but instead uses the underlying quaternion. This is currently untyped, hence no memory view and slow. One could definitely type this, I am just no cython expert and did not know how to express "either 1D or 2D" in terms of memory views.

rgommers:
This is a large PR with potential performance implications and active discussion, so I removed the 1.16.0 milestone - we're not getting this one merged by tomorrow.

crusaderky:
> This is currently untyped, hence no memory view and slow. One could definitely type this, I am just no cython expert and did not know how to express "either 1D or 2D" in terms of memory views.

From personal experience, looking at the C code that cython generates, AFAIK you can't. Unless you can work on a flattened array? (did not look at the algorithm).

ev-br:
> This is currently untyped, hence no memory view and slow. One could definitely type this, I am just no cython expert and did not know how to express "either 1D or 2D" in terms of memory views.

You can't, not without going away from typed memoryviews. A typed memoryview has a fixed number of dimensions at compile time. A pattern we use elsewhere once in a while is to add an length-one axis, grad a 2D memoryview, remove the axis from the result:

```
is_1D = x.ndim == 1
if is_1D:
  x = x[:, None]
  result = cython_function_accepting_a_2D_memoryview(x)
if is_1D:
   result = result[:, 0]
return result
```

whether it's applicable or useful here, I don't know.

amacati:
> whether it's applicable or useful here, I don't know.

It should be possible to use that pattern with minor changes here. Will give an update with renewed benchmarks once these issues are resolved

lucascolley:
Not as far as I know.

But I think we already identified the problem above: `indexer` cannot be empty.

scipy/spatial/transform/_rigid_transform.py
amacati: On my end, this works just fine:

```python
import array_api_strict as xp

x = xp.arange(5)

idx = xp.asarray([], dtype=xp.int64)
y = xp.take(x, idx)  # empty((0,), dtype=array_api_strict.int64)
```
which is exactly what we are using in the tests. How do the tests work for 32bit systems? Do they just set the default of non-explicit types to the 32bit version? If so, explicitly requesting an int64 indices array might be the issue if they cannot be used as index array on 32bit systems. This would also explain the attempted cast from i64 to i32, which then throws.

lucascolley: ah sorry, I guess `xp.asarray([], dtype=xp.int64)` counts as 1-d

> How do the tests work for 32bit systems?

As far as I know, we don't do anything special. It is just about the default NumPy behaviour on such a system.

> explicitly requesting an i64 indices array might be the issue if they cannot be used as index array on 32bit systems. This would also explain the attempted cast from i64 to i32, which then throws.

Yes, that sounds right.

amacati: Resolved by using i32 in tests

lucascolley:
> The drawback would be that we cannot pickle anymore, since modules are non-pickleable as far as I know.

Worth checking whether the consideration from https://github.com/scipy/scipy/pull/23020#issuecomment-2897535533 applies here

ev-br:
Please explain the benchmarks. What are the "numpy no XP" labels? Ideally also list the numbers from asv themselves.
Also https://github.com/scipy/scipy/pull/22777#issuecomment-2905421193 : how does this compare to the jax-native Rotation.

amacati:
> Please explain the benchmarks. What are the "numpy no XP" labels? Ideally also list the numbers from asv themselves.
Also https://github.com/scipy/scipy/pull/22777#issuecomment-2905421193 : how does this compare to the jax-native Rotation.

As explained [here](https://github.com/scipy/scipy/pull/22972#issuecomment-2907904146): 
- scipy main: scipy as of commit [98eaf7b](https://github.com/scipy/scipy/commit/98eaf7b0359848d742c4ae3ad1825395469f4664)
- XP disabled: `SCIPY_ARRAY_API` not set to `1`. Array API features are disabled, which leads to `array_namespace` having near-zero overhead
- XP enabled: `SCIPY_ARRAY_API` set to `1`. Array API features are enabled, which incurs additional overhead

These benchmarks are not done with airspeed velocity, but are my own. I am unfamiliar with airspeed velocity and asked for help integrating these in the official benchmarks in the main thread, but have not received an answer on that front yet.

There is no comparison against `jax` because this PR only concerns `RigidTransform`, which does not have a native `jax` implementation. I do have that on my list and will include it when adding the PR for adding support for the Array API backend for `Rotation`.

rgommers:
> These benchmarks are not done with airspeed velocity, but are my own. I am unfamiliar with airspeed velocity and asked for help integrating these in the official benchmarks in the main thread, but have not received an answer on that front yet.

I'd say that your graphs are very nice, and a lot better than what one can achieve with `asv` (which is primarily set up for detecting performance regressions), so no need to change them. I suggest adding the up-to-date version of the graph to the PR description so it's easy to find.

ev-br:
Thanks for clarifying @amacati !

Indeed, `asv` is brittle and capricious, and it's not uncommon to resort to `%timeit` or other bespoke benchmarking once you need something even just outside of `asv` typical use case. Feel free to ping me for help with `asv`.

For your benchmarks, to make sure I got it right:
- you're comparing what used to be pure cython code and its refactor to a python wrapper + some cython for numpy inputs
- for the time being, you're only running the "numpy backend", i.e. this python-wrapped cython code
- the overhead comes from `array_namespace` (2-5 us per call IME) and extra function calls from 
`array_api_compat.numpy`
- the overhead is visible for `apply` and `as_matrix`, and basically invisible otherwise

Does this sound about right?

For ~100 samples, `apply` gets 2-5x slower. This looks unforturnate, but the absolute times are under 10us, which is nothing --- unless called in a tight loop. How likely is it to be called in a tight loop?
`as_matrix` I'd expect to be mostly one-time calls, so the extra overhead is probably quite tolerable.

amacati:
I have one more proposal to address overhead: One could construct `RigidTransform` by passing `xp` and `_backend` directly, and skip any conversions when we already know that the matrix buffer is valid. 

We could either add more flags to `__init__`, or use `__new__` and set the attributes directly. I don't like that the first option would overcrowd the `__init__` of `RigidTransform.` On the other hand, the second option might be less readable and more error-prone. Using

```python
@staticmethod
def _from_raw_data(matrix: Array, xp: ModuleType, backend: ModuleType | None = None) -> RigidTransform:
    """Create a RigidTransform from raw data to avoid any overhead."""
    tf = RigidTransform.__new__(RigidTransform)
    tf._single = matrix.ndim == 2 and is_numpy(xp)
    if tf._single:
        matrix = xpx.atleast_nd(matrix, ndim=3, xp=xp)
    tf._matrix = matrix
    if backend is None:
        backend = backend_registry[xp]
    tf._backend = backend
    return tf
```
yields noticeable performance improvements for functions that need to create a new `RigidTransform` instance and passes all tests locally. Benchmarks using this direct initialization from data are shown in red (numpy optimized XP disabled).
[tf_benchmark.pdf](https://github.com/user-attachments/files/20433120/tf_benchmark.pdf)

amacati:
> Could you add a `__reduce__` hook that yank out the cached `xp` and recalculate it upon unpickle?

Do you already want to include the feature in this PR, or should this be a follow-up? It is not required for the conversion, bundles nicely with the proposed improvements in https://github.com/scipy/scipy/pull/22972#issuecomment-2908144402 and does not interfere with any of the planned changes, so could be worked on and reviewed in parallel.

amacati:
> I don't mind a minor temporary slowdown.

Then I'd leave this one as it is to make reviews easier.

amacati:
@lucascolley Are there remaining blockers for this?

lucascolley:
sorry @amacati, my finals exams just started for the year. I'll have some time to come back to this in a while.

lucascolley:
would you be able to squash the commits back to a state where we can separately see the changes for the Python and Cython parts @amacati ? No worries if that is too involved, but I think it would aid reviewers.

lucascolley (approved):
I haven't reviewed at line-by-line detail, but this looks good to me now from a high level. I think we may want to wait for gh-23012 and gh-23057 before deciding to merge here, but I think we should put this in and continue iterating once we have a 'final' set of benchmarks for this batch of changes, and everything is passing CI.

@scottshambaugh would you have some time to take a look at these changes?

scipy/spatial/transform/_rigid_transform.py
```diff
@@ -1908,14 +1668,13 @@ def translation(self):
         >>> np.allclose(tf.translation, t)
         True
         """
+        matrix = self._matrix
         if self._single:
-            return np.array(self._matrix[0, :3, 3])
-        else:
-            return np.array(self._matrix[:, :3, 3])
+            matrix = matrix[0, ...]
```
scottshambaugh: This seems to have dropped the correct indexing. Need to make sure we have test coverage of this.

scipy/spatial/transform/_rigid_transform.py
```diff
@@ -1875,14 +1636,13 @@ def rotation(self):
         >>> np.allclose(tf.rotation.as_matrix(), r.as_matrix())
         True
         """
+        matrix = self._matrix
         if self._single:
-            return Rotation.from_matrix(self._matrix[0, :3, :3])
-        else:
-            return Rotation.from_matrix(self._matrix[:, :3, :3])
+            matrix = matrix[0, ...]
```
scottshambaugh: This seems to have dropped the correct indexing. Need to make sure we have test coverage of this.

scipy/spatial/transform/_rigid_transform.py
```diff
@@ -1356,13 +1140,13 @@ def __len__(self):
             ...
         TypeError: Single transform has no len().
         """
-        if self._single:
-            raise TypeError("Single transform has no len().")
-
+        if self.single:
```
scottshambaugh, line 1151: dropped underscore. I think using the property is fine, but should be consistent across all methods

scipy/spatial/transform/_rigid_transform.py
```diff
@@ -654,23 +515,19 @@ def from_rotation(cls, rotation):
         >>> len(tf)
         2
         """
+        # TODO: Should this not raise a TypeError?
```
scottshambaugh, line 524: I think a TypeError is more applicable, yeah

scipy/spatial/transform/_rigid_transform.py
```diff
@@ -1926,4 +1685,10 @@ def single(self):
             True if this instance represents a single transform, False
             otherwise.
         """
-        return self._single
+        return self._single or self._matrix.ndim == 2
```
scottshambaugh, line 1695: self._matrix should always have ndim == 3, has this changed somewhere?

scipy/spatial/transform/_rigid_transform.py
```diff
@@ -1926,4 +1685,10 @@ def single(self):
...
+
+    def __repr__(self):
```
scottshambaugh: could you move this back up just for the cleaner diff?

scottshambaugh (commented):
Looks good apart from a few comments

scipy/spatial/transform/_rigid_transform.py
amacati: Should I change it here directly or is this something for a separate PR?

lucascolley: Follow-up is fine IMO

scipy/spatial/transform/_rigid_transform.py
amacati: This is in fact something in preparation of the Array API backend. The transform matrix of Array API frameworks is not stored in 3D and uses `self._single` to return the correct shape, but keeps its original shape. All Array API functions are written to correctly broadcast (which is possible because they are not restricted by cython's dimension typing).

This means that `self._single` is only used for numpy arrays and the Cython backend. In this particular case we want to raise an error for any single `RigidTransform` though, which is accounted for by the `self.single` property since it also checks the matrix dimensions for Array API matrices.

This only makes sense once the Array API backend has been introduced, but if you are fine with it I would keep it here since it does not change the behaviour of numpy and is necessary for the next steps.

scipy/spatial/transform/_rigid_transform.py
amacati: This should still be correct I think. Previously we were selecting either the first matrix in case of `self._single` and sliced or selected all matrices and sliced `[..., :3, :3]` to cut out the rotation matrix. Now we select the first matrix in case of `self._single`, and then for all cases slice along `[..., :3, :3]`. The result should be the same, since the ellipsis also covers the case that we don't have a third dimension.

scipy/spatial/transform/_rigid_transform.py
amacati: Same thing as above. We can revert to the old formulation though, should be equivalent. One thing I noticed though: Previously we created new Arrays in this method, whereas now we return a view. This means that this is now possible

```python
from scipy.spatial.transform import RigidTransform as TF

tf = TF.identity()
y = tf.translation # [0, 0, 0]
y[0] = 100
tf.translation  # [100, 0, 0]
```

Is that something we want? If not, we should probably test for it.

scipy/spatial/transform/_rigid_transform.py
amacati: See the comment on the Array API for `self.single` vs `self._single`. For the Cython backend this remains unchanged, but not for the Array API version which will be a follow-up. We could remove it here, but will need to add it later anyways.

amacati:
Addressed most of the comments. The last important point that has come up is the copy on `TF.translation`. As soon as we have made a decision on that we should be good to merge, or I can add a test that checks if we are giving out views and add the copy back.

Edit: Failing tests are unrelated to this PR

scipy/spatial/transform/_rigid_transform.py
scottshambaugh: Ah, yeah that makes sense to me.

scipy/spatial/transform/_rigid_transform.py
scottshambaugh: Good catch - I would match existing behavior where we return a copy. Agree we should test for that. Could you also add explicit shape checks for these two properties while you're at it? I'm not seeing those explicitly tested anywhere - I know the np.assert_allclose calls will implicitly broadcast and not verify shape, though I'm not as clear on xp_assert_close.

scipy/spatial/transform/_rigid_transform.py
scottshambaugh: Works for me!

scipy/spatial/transform/_rigid_transform.py
scottshambaugh: Makes sense, I'm fine with phasing this in.

scipy/spatial/transform/_rigid_transform.py
amacati: `xp_assert_close` checks shapes by default, so I don't think we need another check there. I have added the copy test to `test_properties`.

amacati:
The latest commit fixes the last concerns. Should be good to merge now. Test failures are unrelated

scottshambaugh (approved):

lucascolley:
thanks for the review @scottshambaugh !

Alright, let's put this in, thanks all. Looking forward to the follow-ups @amacati !

PR #22976, opened 2025-05-12
ENH: stats: Implement _munp for gennorm

jdtatz:
Adds an analytic implementation for generalized normal moments

#### Additional information
The computation is done in log-space to be consistent with the `_stats` implementation and the formatting is copied from `stats.norm_gen._munp`

### DIFF
--- a/scipy/stats/_continuous_distns.py
+++ b/scipy/stats/_continuous_distns.py
@@ -11453,6 +11453,15 @@ class gennorm_gen(rv_continuous):
     def _isf(self, x, beta):
         return -self._ppf(x, beta)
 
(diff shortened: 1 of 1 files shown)


mdhaber (approved):
LGTM. No additional tests are needed: generic tests demonstrate consistency with the distribution, and I don't see an obvious way to improve the accuracy of this implementation[^1].

[^1]: Aside from maybe performing the calculation without log-space for moderate values of the arguments. But we've lived for long enough without those last few digits, so this should be fine, and the simplicity is preferred.

dschmitz89:
Thanks @jdtatz , keep them coming :).

PR #22982, opened 2025-05-13
BUG: signal: fix incorrect vendoring of `npp_polyval`

### LINKED ISSUE #22981: Bug with freqz when specifying worN after #22886
On 1.15.0:
```
>>> from scipy.signal import freqz
>>> freqz([1.], worN=[0., 0.1])
(array([0. , 0.1]), array([1.+0.j, 1.+0.j]))
```
i.e., the length of `w` matches the length of `h`, both of which match the length of `worN`. On scientific-python-nightly-wheels:
```
>>> freqz([1.], worN=[0., 0.1])
(array([0. , 0.1]), np.float64(1.0))
```
i.e., the length of `h` does not match, and the dtype is wrong (should be complex but it's a float64). I am very surprised that our unit tests didn't catch this somehow!

cc @ev-br @j-bowhay as I think this comes from #22886 but I could be wrong

ev-br:
#### Reference issue

fixes gh-22981

Add a missing broadcast-promote operation to our array-api compatible clone of `numpy.polynomial`: the clone was missing `c0 = c[-1] + x*0` (https://github.com/numpy/numpy/blob/v2.2.0/numpy/polynomial/polynomial.py#L752), and this PR adds a slightly more verbose variant --- which is a harder to dismiss as a no-op (which I initially did).

### DIFF
--- a/scipy/signal/_polyutils.py
+++ b/scipy/signal/_polyutils.py
@@ -6,6 +6,7 @@ namespace, and "new-style", np.polynomial.polynomial, routines.
 To distinguish the two sets, the "new-style" routine names start with `npp_`
 """
 import scipy._lib.array_api_extra as xpx
+from scipy._lib._array_api import xp_promote, xp_default_dtype
 
 
 def _sort_cmplx(arr, xp):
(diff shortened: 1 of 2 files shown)


scipy/signal/_polyutils.py
```diff
@@ -138,15 +139,18 @@ def polymul(a1, a2, *, xp):
 # ### New-style routines ###
 
 
-# https://github.com/numpy/numpy/blob/v2.2.0/numpy/polynomial/polynomial.py#L845-L894
+# https://github.com/numpy/numpy/blob/v2.2.0/numpy/polynomial/polynomial.py#L663
 def npp_polyval(x, c, *, xp, tensor=True):
+    if xp.isdtype(c.dtype, 'integral'):
+        c = xp.astype(c, xp_default_dtype(xp))
+
```
j-bowhay, line 146:
```suggestion
```

scipy/signal/_polyutils.py
```diff
@@ -138,15 +139,18 @@ def polymul(a1, a2, *, xp):
...
     c = xpx.atleast_nd(c, ndim=1, xp=xp)
     if isinstance(x, tuple | list):
         x = xp.asarray(x)
     if tensor:
         c = xp.reshape(c, (c.shape + (1,)*x.ndim))
 
-    c0 = c[-1, ...]
+    c0, _ = xp_promote(c[-1, ...], x, broadcast=True, xp=xp)
```
j-bowhay, line 153:
```suggestion
    c0, _ = xp_promote(c[-1, ...], x, broadcast=True, force_floating=True,
                       xp=xp)
```

j-bowhay (approved):
Thanks for the quick fix! Happy that this serves the same purpose as the missing line.

I think we can streamline the fix slightly however this isn't blocking

tylerjereddy (approved):
I confirmed fail before/pass after patch locally and discussed in person with Jake.

This is "ok" as-is so I'm going to merge, thx.

PR #22983, opened 2025-05-13
ENH: sparse.linalg: a restarted Krylov method for evaluating f(A)b.

### LINKED ISSUE #22974: ENH: sparse.linalg: Krylov methods for matrix functions

I am currently investigating the numerical evaluation of matrix functions and I noticed that SciPy only has `expm_multiply` for evaluating $$f(A) b$$ (with $$f(A) = e^A$$ in this case). For other functions, it is required to first form the full matrix $$f(A)$$ and then multiply with $$b$$, which is very inefficient (and sometimes impossible if $$A$$ is large). 

I suggest to add another routine in `scipy.sparse.linalg` based on the  (restarted) Krylov subspace method from [1, 2]. The idea of the method is to use the Arnoldi/Lanczos iteration to construct an orthonormal Krylov basis $$V_m$$ and the projection $$H_m = V_m^* A V_m$$. Then, $$f(A) b$$ can be approximated as $$f(A) b \approx \|b\|_2 V_m f(H_m) e_1$$, where $$e_1$$ is the first canonical basis vector. Since $$H_m$$ is often small, we can just use the dense numerical methods from `scipy.linalg`, like `expm`, `cosm`, `sinm`, etc. Thus, this enables the computation of the action of any matrix function. The Krylov method is also usually more efficient than the scheme from `expm_multiply` for the matrix exponential. 

The original implementation is in [MATLAB](http://guettel.com/funm_kryl/), but I wrote a C++ version for my own linear algebra library [HighFM](https://gitlab.com/highfm/highfm). 

References:

[1] M. Afanasjew, M. Eiermann, O. G. Ernst, and S. Güttel, “Implementation of a restarted Krylov subspace method for the evaluation of matrix functions,” Linear Algebra and its Applications, vol. 429, no. 10, pp. 2293–2314, Nov. 2008, doi: [10.1016/j.laa.2008.06.029](https://doi.org/10.1016/j.laa.2008.06.029).

[2] M. Eiermann and O. G. Ernst, “A Restarted Krylov Subspace Method for the Evaluation of Matrix Functions,” SIAM J. Numer. Anal., vol. 44, no. 6, pp. 2481–2504, Jan. 2006, doi: [10.1137/050633846](https://doi.org/10.1137/050633846).

None

None

nlg550:
#### Reference issue

Closes #22974 for vectors.

Implement a restarted Krylov method for computing `f(A) b` for an arbitrary function `f` and vector `b`. The routine is called `krylov_funmv`  in the ``sparse.linalg`` module. The tests cover the `exp`, `sin`, `cos`, `sinh` and `cosh` functions as well as a user-provided function for both hermitian and non-hermitian matrices.  Note that the method may not be convergent for some functions (e.g., `sqrt`), and thus, they are not included in the tests.

### DIFF
--- a/scipy/sparse/linalg/__init__.py
+++ b/scipy/sparse/linalg/__init__.py
@@ -22,6 +22,7 @@ Matrix Operations
    inv -- compute the sparse matrix inverse
    expm -- compute the sparse matrix exponential
    expm_multiply -- compute the product of a matrix exponential and a matrix
+   krylov_funmv -- use a Krylov method to compute f(A)b for a general f 
    matrix_power -- compute the matrix power by raising a matrix to an exponent
 
 Matrix norms
--- a/scipy/sparse/linalg/__init__.py
+++ b/scipy/sparse/linalg/__init__.py
@@ -136,6 +136,7 @@ from ._matfuncs import *
 from ._onenormest import *
 from ._norm import *
 from ._expm_multiply import *
+from ._krylov_funmv import *
 from ._special_sparse_arrays import *
 
 # Deprecated namespaces, to be removed in v2.0.0
--- a/scipy/sparse/linalg/__init__.py
+++ b/scipy/sparse/linalg/__init__.py
@@ -22,7 +22,7 @@ Matrix Operations
    inv -- compute the sparse matrix inverse
    expm -- compute the sparse matrix exponential
    expm_multiply -- compute the product of a matrix exponential and a matrix
-   krylov_funmv -- use a Krylov method to compute f(A)b for a general f 
+   funm_multiply_krylov -- use a Krylov method to compute f(A)b for a general f
    matrix_power -- compute the matrix power by raising a matrix to an exponent
 
 Matrix norms
@@ -137,7 +137,7 @@ from ._matfuncs import *
 from ._onenormest import *
 from ._norm import *
 from ._expm_multiply import *
-from ._krylov_funmv import *
+from ._funm_multiply_krylov import *
 from ._special_sparse_arrays import *
 
 # Deprecated namespaces, to be removed in v2.0.0
--- a/scipy/sparse/linalg/_funm_multiply_krylov.py
+++ b/scipy/sparse/linalg/_funm_multiply_krylov.py
@@ -1,4 +1,39 @@
-"""Restared Krylov method for evaluating f(A)b"""
+"""
+  Restared Krylov method for evaluating f(A)b
+
+  The original code was written in MATLAB by Stefan Guttel.
+  It was later adapted to C++ and Python by Nicolas Guidotti.
+  Both authors agrees to relicense the code under the BSD license.
+
+  Copyright (C) 2025 Nicolas Guidotti and Stefan Guttel.
+  All rights reserved.
+
+  Redistribution and use in source and binary forms, with or without
+  modification, are permitted provided that the following conditions are met:
+
+  1. Redistributions of source code must retain the above copyright notice, this
+  list of conditions and the following disclaimer.
+
+  2. Redistributions in binary form must reproduce the above copyright notice,
+  this list of conditions and the following disclaimer in the documentation
+  and/or other materials provided with the distribution.
+
+  3. Neither the name of the copyright holder nor the names of its contributors
+  may be used to endorse or promote products derived from this software without
+  specific prior written permission.
+
+  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+  IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+  DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+  FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+  DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+  SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+  OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+"""
 
 import numpy as np
 from scipy.linalg import norm
@@ -116,16 +151,6 @@ def funm_multiply_krylov(f, A, b, *, assume_a = "general", t = 1.0, atol = 0.0,
     """
     A restarted Krylov method for evaluating ``y = f(tA) b`` from [1]_ [2]_.
 
-    If the structure of ``A`` is known, then supplying the
-    corresponding string to ``assume_a`` key chooses the dedicated algorithm.
-    The available options are
-
-    =============================  ================================
-     symmetric                      'symmetric' (or 'sym')
-     hermitian                      'hermitian' (or 'her')
-     general                        'general' (or 'gen')
-    =============================  ================================
-
     Parameters
     ----------
     f : callable
@@ -140,8 +165,10 @@ def funm_multiply_krylov(f, A, b, *, assume_a = "general", t = 1.0, atol = 0.0,
         A vector to multiply the ``f(tA)`` with.
 
     assume_a : string, optional
-        Indicate the structure of ``A`` as described above. If ommited, then
-        it is assumed that ``A`` has a general structure.
+        Indicate the structure of ``A``. The algorithm will use this information
+        to select the appropriated code path. The available options are
+        'hermitian'/'her' and 'general'/'gen'. If ommited, then it is assumed
+        that ``A`` has a 'general' structure.
 
     t : float, optional
         The value to scale the matrix ``A`` with. The default is ``t = 1.0``
@@ -186,11 +213,26 @@ def funm_multiply_krylov(f, A, b, *, assume_a = "general", t = 1.0, atol = 0.0,
     Compute ``y = exp(tA) b``.
 
     >>> y = funm_multiply_krylov(expm, A, b, t = t)
+    >>> y
     [3.6164913 3.88421511 0.96073457]
 
     >>> ref = expm(t * A.todense()) @ b
-    >>> y - ref
+    >>> err = y - ref
+    >>> err
     [4.44089210e-16 0.00000000e+00 2.22044605e-16]
+
+    Compute :math:`y = (A^3 - A) b`.
+
+    >>> def polynomial(X):
+    >>>     return X @ X @ X - X
+    >>> y = funm_multiply_krylov(polynomial, A, b)
+    >>> y
+    [132.  24.  70.]
+
+    >>> ref = polynomial(A.todense()) @ b
+    >>> err = y - ref
+    >>> err
+    [ 0.00000000e+00  7.10542736e-15 -2.84217094e-14]
 
     Compute :math:`y = f(tA) b`, where  :math:`f(X) = X^{-1}(e^{X} - I)`. This is
     known as the "phi function" from the exponential integrator literature.
@@ -198,10 +240,12 @@ def funm_multiply_krylov(f, A, b, *, assume_a = "general", t = 1.0, atol = 0.0,
     >>> def phim_1(X):
     >>>     return solve(X, expm(X) - np.eye(X.shape[0]))
     >>> y = funm_multiply_krylov(phim_1, A, b, t = t)
+    >>> y
     [ 2.76984306  3.92769192 -0.03111392]
 
     >>> ref = phim_1(t * A.todense()) @ b
-    >>> y - ref
+    >>> err = y - ref
+    >>> err
     [ 0.00000000e+00  8.88178420e-16 -4.60742555e-15]
 
     References
@@ -222,10 +266,10 @@ def funm_multiply_krylov(f, A, b, *, assume_a = "general", t = 1.0, atol = 0.0,
 
     """
 
-    if assume_a not in {'symmetric', 'hermitian', 'general', 'sym', 'her', 'gen'}:
+    if assume_a not in {'hermitian', 'general', 'her', 'gen'}:
         raise ValueError(f'scipy.sparse.linalg.funm_multiply_krylov: {assume_a} '
                          'is not a recognized matrix structure')
-    is_hermitian = (assume_a != 'gen') and (assume_a != 'general')
+    is_hermitian = (assume_a == 'her') or (assume_a == 'hermitian')
 
     if len(b.shape) != 1:
         raise ValueError("scipy.sparse.linalg.funm_multiply_krylov: "
--- a/scipy/sparse/linalg/_funm_multiply_krylov.py
+++ b/scipy/sparse/linalg/_funm_multiply_krylov.py
@@ -1,11 +1,13 @@
 """Restared Krylov method for evaluating f(A)b"""
+from warnings import warn
 
 import numpy as np
+from scipy.linalg import norm
 from ._isolve.iterative import _get_atol_rtol
 
-__all__ = ['krylov_funmv']
+__all__ = ['funm_multiply_krylov']
 
-def _krylov_funmv_arnoldi(A, b, bnorm, V, H, m):
+def _funm_multiply_krylov_arnoldi(A, b, bnorm, V, H, m):
     """
     The Arnoldi iteration for constructing the basis V and the projection H = V * A V
     for the Krylov subspace Km(A, b) of order m.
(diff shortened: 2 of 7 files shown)


nlg550:
`krylov_funmv` is just a suggestion. If there is a better name, I will gladly changed it.

scipy/sparse/linalg/__init__.py
```diff
@@ -22,6 +22,7 @@
    inv -- compute the sparse matrix inverse
    expm -- compute the sparse matrix exponential
    expm_multiply -- compute the product of a matrix exponential and a matrix
+   krylov_funmv -- use a Krylov method to compute f(A)b for a general f
```
ilayn, line 25: We better call this `funmv_multiply_krylov` both to match expm_multiply and in case we implement the exact version via another method. It's long and a bit ugly but it's always easier to rename than replace.

scipy/sparse/linalg/_krylov_funmv.py
```diff
@@ -0,0 +1,266 @@
...
+        H[k, k] = dotprod(V[:, k + 1], V[:, k])
+        V[:, k + 1] = V[:, k + 1] - H[k, k] * V[:, k]
+
+        H[k + 1, k] = np.linalg.norm(V[:, k + 1])
+        if H[k + 1, k] != 0:
+            V[:, k + 1] = V[:, k + 1] / H[k + 1, k]
+            if k < m - 1:
+                H[k, k + 1] = H[k + 1, k]
+        else:
+            raise RuntimeError("krylov_funmv: The Lanczos iteration broke down "
```
ilayn, line 88: Same as above

scipy/sparse/linalg/_krylov_funmv.py
```diff
@@ -0,0 +1,266 @@
...
+
+
+def krylov_funmv(f, t, A, b, *, atol = 0.0, btol = 1e-6, restart_length = None,
+                 max_restarts = 20, ortho_method = "arnoldi", verbose = False):
+    """
+    A restarted Krylov method for evaluating ``y = f(tA) b`` from [1]_ [2]_.
+
+    Parameters
+    ----------
+    f : function
```
ilayn, line 99: The type for this is `callable`

scipy/sparse/linalg/_krylov_funmv.py
```diff
@@ -0,0 +1,266 @@
...
+        Callable object that computes the matrix function ``F = f(X)``.
+
+    t : float
```
ilayn, line 102: This should be optional and should assume the value `1.0` since it won't be always needed so it shouldn't get in the way.

scipy/sparse/linalg/_krylov_funmv.py
```diff
@@ -0,0 +1,266 @@
...
+
+    A : {sparse array, ndarray, LinearOperator}
+        A real or complex N-by-N matrix.
+        Alternatively, `A` can be a linear operator which can
+        produce ``Ax`` using, e.g., ``scipy.sparse.linalg.LinearOperator``.
+
+    b : ndarray, sparse array
+        A vector to multiply the ``f(tA)`` with.
+
+    atol, btol : float, optional
```
ilayn, line 113: We use the `atol` and `rtol` terminology for having absolute and relative tolerance. atol, btol is confusing here since the arrays are also called `A` and `b`. The explanation is very clear.

scipy/sparse/linalg/_krylov_funmv.py
```diff
@@ -0,0 +1,266 @@
...
+        Parameters for the convergence test. For convergence,
+        ``norm(||y_k - y_k-1||) <= max(btol*norm(b), atol)`` should be satisfied.
+        The default is ``atol=0.`` and ``btol=1e-6``.
+
+    restart_length : integer
```
ilayn, line 118: `restart_iter_limit`, or `restart_every_n` seems a bit more informative here but just a suggestion, not a blocker. The text can be a bit more clearer if we mention that "_if iteration number reaches this value a restart is triggered_".

scipy/sparse/linalg/_krylov_funmv.py
```diff
@@ -0,0 +1,266 @@
...
+        Number of iterations between restarts. Larger values increase
+        iteration cost, but may be necessary for convergence.
+        If omitted, ``min(20, n)`` is used.
+
+    max_restarts : int, optional
+        Maximum number of restart cycles.  The algorithm will stop
```
ilayn, line 124:
```suggestion
        Maximum number of restart cycles. The algorithm will stop
```

scipy/sparse/linalg/_krylov_funmv.py
```diff
@@ -0,0 +1,266 @@
...
+        after max_restarts cycles even if the specified tolerance has not been
+        achieved. The default is ``max_restarts=20``
+
+    ortho_method : string, optional
```
ilayn, line 128: In the linalg module we use the keyword `assume_a` with values `gen` and `her` for this (see `linalg.solve` for example). If it is `her`, it can dispatch to Lanczos otherwise does the Arnoldi. Probably should go first in the order of keywords.

scipy/sparse/linalg/_krylov_funmv.py
```diff
@@ -0,0 +1,266 @@
...
+        Orthogonalization method to use. ``lanczos`` is faster, but it requires
+        that ``A`` is Hermitian. ``arnoldi`` is the default option.
+
+    verbose : bool
```
ilayn, line 132: You can remove this. `print`ing things turned to out to be not useful in any function we had and we started removing them all together.

scipy/sparse/linalg/_krylov_funmv.py
```diff
@@ -0,0 +1,266 @@
...
+
+    .. [3] A. Frommer, S. Güttel, and M. Schweitzer, "Convergence of Restarted
+           Krylov Subspace Methods for Stieltjes Functions of Matrices," SIAM J.
+           Matrix Anal. Appl., vol. 35, no. 4, pp. 1602-1624,
+           Jan. 2014, :doi:`10.1137/140973463`.
+
+    """
+
+    if len(b.shape) != 1:
+        raise RuntimeError("krylov_funmv: b must be a vector.")
```
ilayn, line 196: This is a `ValueError` and we spell out the entire function name in case this is hit in somebody else's code hence `scipy.sparse.linalg.funm_multiply_krylov: argument "b" must be a 1D array.`

scipy/sparse/linalg/_krylov_funmv.py
```diff
@@ -0,0 +1,266 @@
...
+
+    n = b.shape[0]
+
+    if restart_length is None:
+        restart_length = min(20, n)
+    m = restart_length
+    max_restarts = min(max_restarts, int(n / restart_length) + 1)
+    mmax = restart_length * max_restarts
+
+    bnorm = np.linalg.norm(b)
```
ilayn, line 206: np.linalg.norm does not accept sparse input if you want the RHS to be sparse

scipy/sparse/linalg/_krylov_funmv.py
```diff
@@ -0,0 +1,266 @@
...
+    atol, _ = _get_atol_rtol("krylov_funmv", bnorm, atol, btol)
+
+    if bnorm == 0:
+        y = b
```
ilayn, line 210: You need to copy here to detach the output from the input argument.

ilayn (changes requested):
Thank you @nlg550 It looks good and I've placed some comments for the first pass.

scipy/sparse/linalg/__init__.py
nlg550: Sounds good. I dropped the `v` part to allow it to be extended to matrix blocks in the future.

scipy/sparse/linalg/_krylov_funmv.py
```diff
@@ -0,0 +1,266 @@
...
+        # against the previous basis vectors
+        for i in range(0, k + 1):
+            H[i, k] = dotprod(V[:, i], V[:, k + 1])
+            V[:, k + 1] = V[:, k + 1] - H[i, k] * V[:, i]
+
+        H[k + 1, k] = np.linalg.norm(V[:, k + 1])
+        if H[k + 1, k] != 0:
+            V[:, k + 1] = V[:, k + 1] / H[k + 1, k]
+        else:
+            raise RuntimeError("krylov_funmv: The Arnoldi iteration broke down "
```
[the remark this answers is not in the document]
nlg550: You are right. When the Arnoldi/Lanczos iteration break down, we get the "exact" value of `f(A)`. I modified the code to handle this case. There is also a test case that checks the breakdown.

Explanation: When the Arnoldi/Lanczos iteration break down, the spectrum of the projection `H` coincides with the spectrum of `A`, hence `f(H)` is exactly equal to `f(A)` (at least in theory). This is why unrestarted Krylov method converges to  `f(A) b` eventually. With restarting, the convergence is no longer guaranteed, but if a breakdown occurs, then it will be on the first cycle anyway.

scipy/sparse/linalg/_krylov_funmv.py
nlg550: Done.

scipy/sparse/linalg/_krylov_funmv.py
nlg550: I think it is better to drop the `sparse array` support. `b` will be converted to a column of a dense matrix in the Arnoldi/Lanczos iteration anyway. Looking in the `gmres` and other iterative linear solvers, they also only supports `b` as a `ndarray`.

scipy/sparse/linalg/_krylov_funmv.py
nlg550: Done

scipy/sparse/linalg/_krylov_funmv.py
nlg550: Ok

scipy/sparse/linalg/_krylov_funmv.py
nlg550: Done

scipy/sparse/linalg/_krylov_funmv.py
nlg550: I changed to `restart_every_m`.

scipy/sparse/linalg/_krylov_funmv.py
nlg550: Fixed in [b334447](https://github.com/scipy/scipy/pull/22983/commits/b33444708b80fc96cedb33e25f8db954af0c918a)

scipy/sparse/linalg/_krylov_funmv.py
nlg550: Fixed at [b334447](https://github.com/scipy/scipy/pull/22983/commits/b33444708b80fc96cedb33e25f8db954af0c918a)

scipy/sparse/linalg/_krylov_funmv.py
nlg550: Verbose option removed in [b334447](https://github.com/scipy/scipy/pull/22983/commits/b33444708b80fc96cedb33e25f8db954af0c918a)

scipy/sparse/linalg/_krylov_funmv.py
nlg550: Fixed in [b334447](https://github.com/scipy/scipy/pull/22983/commits/b33444708b80fc96cedb33e25f8db954af0c918a)

scipy/sparse/linalg/_krylov_funmv.py
```diff
@@ -0,0 +1,266 @@
...
+           Matrix Anal. Appl., vol. 35, no. 4, pp. 1602-1624,
+           Jan. 2014, :doi:`10.1137/140973463`.
+
+    """
+
+    if len(b.shape) != 1:
+        raise RuntimeError("krylov_funmv: b must be a vector.")
+
+    n = b.shape[0]
+
```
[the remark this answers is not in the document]
nlg550: Fixed in [b334447](https://github.com/scipy/scipy/pull/22983/commits/b33444708b80fc96cedb33e25f8db954af0c918a)

scipy/sparse/linalg/_krylov_funmv.py
```diff
@@ -0,0 +1,266 @@
...
+        print(f"ortho_method = {ortho_method}")
+        print(f"||b|| = {bnorm}")
+
+    restart = 1
+    if ortho_method == "lanczos":
+        _krylov_funmv_lanczos(A, b, bnorm, V, H[:m + 1, :m], m)
+    elif ortho_method == "arnoldi":
+        _krylov_funmv_arnoldi(A, b, bnorm, V, H[:m + 1, :m], m)
+    else:
+        raise RuntimeError("krylov_funmv: Invalid orthogonalization method. "
```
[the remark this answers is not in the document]
nlg550: Fixed in [b334447](https://github.com/scipy/scipy/pull/22983/commits/b33444708b80fc96cedb33e25f8db954af0c918a)

scipy/sparse/linalg/_krylov_funmv.py
nlg550: Fixed in [b334447](https://github.com/scipy/scipy/pull/22983/commits/b33444708b80fc96cedb33e25f8db954af0c918a)

scipy/sparse/linalg/_krylov_funmv.py
```diff
@@ -0,0 +1,266 @@
...
+    atol, _ = _get_atol_rtol("krylov_funmv", bnorm, atol, btol)
+
+    if bnorm == 0:
+        y = b
+        return y
+
+    # Pre-allocate the maximum memory space.
+    # Using the column major order here since we work with
+    # each individual column separately.
+    V = np.zeros((n, m + 1), dtype = b.dtype, order = 'F')
```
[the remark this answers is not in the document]
nlg550: Fixed in [b334447](https://github.com/scipy/scipy/pull/22983/commits/b33444708b80fc96cedb33e25f8db954af0c918a)

scipy/sparse/linalg/_krylov_funmv.py
ilayn: OK we can try to add if there is a need later.

scipy/sparse/linalg/_krylov_funmv.py
ilayn: Ah yes, it does not need to detect automatically but a user can hint that the operator is Hermitian. So we select the method based on that.

scipy/sparse/linalg/_krylov_funmv.py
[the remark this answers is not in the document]
ilayn: That's a very nice addition, thanks.

scipy/sparse/linalg/_funm_multiply_krylov.py
```diff
@@ -0,0 +1,306 @@
...
+                         rtol = 1e-6, restart_every_m = None, max_restarts = 20):
+    """
+    A restarted Krylov method for evaluating ``y = f(tA) b`` from [1]_ [2]_.
+
+    If the structure of ``A`` is known, then supplying the
+    corresponding string to ``assume_a`` key chooses the dedicated algorithm.
+    The available options are
+
+    =============================  ================================
+     symmetric                      'symmetric' (or 'sym')
```
ilayn, line 124: One particular annoyance about symmetric is that a complex array can be symmetric but not hermitian. Hence we use hermitian to cover both dtypes complex and real. Symmetric we reserve for just symmetric arrays including complexes. So probably we want to avoid this one as a valid entry.

I know this is very annoying but officially welcome on board of dtype olympics in SciPy :)

scipy/sparse/linalg/_funm_multiply_krylov.py
```diff
@@ -0,0 +1,306 @@
...
+    Compute ``y = exp(tA) b``.
+
+    >>> y = funm_multiply_krylov(expm, A, b, t = t)
+    [3.6164913 3.88421511 0.96073457]
+
+    >>> ref = expm(t * A.todense()) @ b
+    >>> y - ref
+    [4.44089210e-16 0.00000000e+00 2.22044605e-16]
+
+    Compute :math:`y = f(tA) b`, where  :math:`f(X) = X^{-1}(e^{X} - I)`. This is
```
ilayn, line 195: Should we also add a simple polynomial function? A low order polynomial just to warm-up the user to what `funm` is before we jump to phi function? Not a blocker, just a suggestion.

ilayn (commented):
I've added a few minor things but other than that this is in good shape.

scipy/sparse/linalg/_funm_multiply_krylov.py
nlg550: Ok. Dropped the `symmetric` option in [7f3998e](https://github.com/scipy/scipy/pull/22983/commits/7f3998e9f2a824e4a8d197d199a73f857b564759)

scipy/sparse/linalg/_funm_multiply_krylov.py
nlg550: Added a polynomial example in [7f3998e](https://github.com/scipy/scipy/pull/22983/commits/7f3998e9f2a824e4a8d197d199a73f857b564759). Although for short polynomials, I think it is more efficient to just compute them directly instead of using a Krylov method.

scipy/sparse/linalg/_funm_multiply_krylov.py
```diff
@@ -0,0 +1,306 @@
+"""Restared Krylov method for evaluating f(A)b"""
```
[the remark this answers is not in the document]
nlg550: Added the licence agreement in the beginning of the file. This is enough?

scipy/sparse/linalg/_funm_multiply_krylov.py
```diff
@@ -0,0 +1,306 @@
...
+    >>> from scipy.sparse import csr_array
+    >>> from scipy.sparse.linalg import funm_multiply_krylov
+    >>> from scipy.linalg import expm, solve
+    >>> A = csr_array([[3, 2, 0], [1, -1, 0], [0, 5, 1]], dtype=float)
+    >>> b = np.array([2, 4, -1], dtype=float)
+    >>> t = 0.1
+
+    Compute ``y = exp(tA) b``.
+
+    >>> y = funm_multiply_krylov(expm, A, b, t = t)
```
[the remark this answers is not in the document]
nlg550: Fixed in [4a67f53](https://github.com/scipy/scipy/pull/22983/commits/4a67f53b1b2e34f78f59cd3dc24a8aebdb891af8)

ilayn:
All looks good. The failure is not related. Thank you for your contribution @nlg550, much appreciated.

j-bowhay:
Thanks for a great first contribution to SciPy @nlg550

PR #22987, opened 2025-05-14
DOC: sparse: sparray migration guide updates

dschult:
This PR updates the sparray migration guide to display the latest helper script.
It also updates terms surrounding e.g. random_state vs rng. And it tweaks some language.

The updates are intended to appear in v1.16 so it has the latest info.

@perimosocordiae, @j-bowhay


j-bowhay (approved):
One minor error but otherwise looks good to me!

PR #22988, opened 2025-05-14
ENH: stats: allow shifting of discrete distributions

j-bowhay:
Allows shifting of discrete distributions by integral values.

Tests to follow!

### DIFF
--- a/scipy/stats/_distribution_infrastructure.py
+++ b/scipy/stats/_distribution_infrastructure.py
@@ -4453,6 +4453,10 @@ class TruncatedDistribution(TransformedDistribution):
                           _Parameterization(_ub_param)]
 
     def __init__(self, X, /, *args, lb=-np.inf, ub=np.inf, **kwargs):
+        if not isinstance(X, ContinuousDistribution):
+            message = ("Truncated distributions are currently only supported for "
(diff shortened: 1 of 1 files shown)


scipy/stats/_distribution_infrastructure.py
```diff
@@ -4365,9 +4385,6 @@ def wrapped(self, p, *args, loc, scale, sign, **kwargs):
 
 class TransformedDistribution(ContinuousDistribution):
```
j-bowhay, line 4386: Hmmm this makes a discrete distribution continuous which clearly isn't right!

mdhaber:
```suggestion
class TransformedDistribution(UnivariateDistribution):
```
?

j-bowhay: I'll double check but I think I tried this but it broke the tests for continuous transformed distributions as they no longer have a _logpxf_dispatch or the other methods defined by ContinuousDistribution

mdhaber:
Thanks, Jake. I'll pick this up again when the circular distributions are in, since we'll want to be able to shift and scale them, too.

PR #22989, opened 2025-05-15
ENH: `cluster`: vectorize `is_isomorphic`

crusaderky:
Rewrite `cluster.is_isomorphic` using the Array API for a 10~100x speedup.

# Benchmark

5 isomorphic clusters (arrays contain `n_obs * 5` points). All times in microseconds.

image
### Notes
- On main, JAX GPU only works if the transfer guard is disabled
- On main, Dask timings are likely going to be much higher in real life due to workers→client transfer
- On main, `jax.pure_callback`  blocks all unrelated work until done


scipy/cluster/hierarchy.py
```diff
@@ -3855,26 +3850,19 @@ def is_isomorphic(T1, T2):
...
+
+    # O(n*log(n)) algorithm.
+    # It is also possible to write a O(n) algorithm on top of unique_all(),
+    # but in practice it's much slower even for large clusters.
+    idx = xp.argsort(T1)
+    T1 = xp.take(T1, idx)
+    T2 = xp.take(T2, idx)
+    changes1 = T1[:-1] == T1[1:]
+    changes2 = T2[:-1] == T2[1:]
+    return xp.all(changes1 == changes2)
```
crusaderky, line 3864: Changes from aligning to `is_valid_linkage` and `is_valid_im` to aligning to `is_monotonic`, which feels much more natural.
Note that this changes the return type on numpy from `bool` to `np.bool_`. It shouldn't be a problem worth worrying about IMHO?

scipy/cluster/hierarchy.py
```diff
@@ -3855,26 +3850,19 @@ def is_isomorphic(T1, T2):
...
-            else:
-                d1[t1] = t2
-                d2[t2] = t1
-        return True
+    # For each pair of (i, j) indices, test that
+    # T1[i] == T1[j] <--> T2[i] == T2[j]
+
+    # O(n*log(n)) algorithm.
+    # It is also possible to write a O(n) algorithm on top of unique_all(),
+    # but in practice it's much slower even for large clusters.
```
crusaderky, line 3858: tested up to 1m observations, 5m nodes. It just never seems to catch up.

crusaderky:
windows failures are unrelated

crusaderky:
test failures are unrelated

scipy/cluster/hierarchy.py
tylerjereddy: > This PR changes the return value

Can you clarify--this is a change in the public API return value, even when using NumPy? It is getting a bit hard for me to track/review all this is going on, but that kind of change without i.e., deprecation/some kind of transition would give me pause.

Perhaps a concrete example of the change when using NumPy might suit for reviewers with limited time.

tylerjereddy (commented):
I added one comment.

Also, why do the benchmarks sometimes show a "speedup"  row and other times a "delta" row? It is taking some effort to try to parse the meaning of those benchmarks, which is part of the original appeal of the standardization of `asv` benchmarks. I guess `asv` may not be ready/suitable for array API benchmarking at this time, but seeing i.e., JAX on CPU and GPU with speedups reported using "delta" vs. "speedup" (fold) values in a bit distracting.

scipy/cluster/hierarchy.py
crusaderky: Yes, this changes the return type on numpy from `bool` to `np.bool_`.
I could add a switch that reverts _just numpy_ to `bool`, but it feels like an antipattern?

crusaderky:
> I added one comment.
> 
> Also, why do the benchmarks sometimes show a "speedup" row and other times a "delta" row? It is taking some effort to try to parse the meaning of those benchmarks, which is part of the original appeal of the standardization of `asv` benchmarks. I guess `asv` may not be ready/suitable for array API benchmarking at this time, but seeing i.e., JAX on CPU and GPU with speedups reported using "delta" vs. "speedup" (fold) values in a bit distracting.

Yes, apologies, I should have spent some extra time and figured out how to rig the benchmark in asv.
"delta" is an artefact caused by me composing the final output by hand.

scipy/cluster/hierarchy.py
tylerjereddy: I'm just saying we don't normally make backwards-incompatible changes because of intuition (you probably know this)--normally they need to be justified as bug fixes or go through some kind of process. That specific example sounds fairly minor, but hard to judge as a non-consumer of the functionality.

Anyway, just making sure this gets discussed before merge.

crusaderky:
Turns out, there was no framework whatsosever for running asv on Array API. So I had to write it (https://github.com/scipy/scipy/pull/23012). Let's just say it's been a _long_ weekend.
Here's the ASV output (stitched by hand, as I couldn't convince `dev.py bench --compare` to cooperate with the SCIPY_ARRAY_API env variable).
```
================== =========== ========== ==========
    backend           nobs       main      #22989    
------------------ ----------- ---------- ----------
      numpy            100      156μs      143μs  
      numpy            1000     403μs      173μs  
      numpy           10000     2.93ms     372μs 
      numpy           100000    28.2ms     4.01ms
      numpy          1000000    290ms      46.6ms
      cupy            100        n/a       58.5ms 
      cupy            1000       n/a       58.2ms 
      cupy           10000       n/a       58.5ms 
      cupy           100000      n/a       58.8ms 
      cupy          1000000      n/a       58.8ms 
      cupy          10000000     n/a       63.4ms 
      cupy         100000000     n/a       121ms 
    torch:cpu          100      371μs      1.36ms 
    torch:cpu          1000     646μs      1.39ms 
    torch:cpu         10000     3.29ms     1.58ms 
    torch:cpu         100000    29.1ms     5.94ms 
    torch:cpu        1000000    271ms      72.7ms 
    torch:cuda         100        n/a      43.3ms 
    torch:cuda         1000       n/a      43.2ms 
    torch:cuda        10000       n/a      45.2ms 
    torch:cuda        100000      n/a      45.5ms 
    torch:cuda       1000000      n/a      45.9ms 
    torch:cuda       10000000     n/a      50.8ms 
    torch:cuda      100000000     n/a      98.7ms 
    dask.array         100      4.84ms     11.9ms 
    dask.array         1000     5.16ms     12.6ms 
    dask.array        10000     8.00ms     12.5ms 
    dask.array        100000    33.6ms     19.3ms 
    dask.array       1000000    291ms      83.2ms 
  jax.numpy:cpu        100      730μs      51.8μs 
  jax.numpy:cpu        1000     1.01ms     145μs  
  jax.numpy:cpu       10000     3.69ms     2.34ms 
  jax.numpy:cpu       100000    30.9ms     30.6ms 
  jax.numpy:cpu      1000000    288ms      213ms  
  jax.numpy:cuda       100        n/a      101μs  
  jax.numpy:cuda       1000       n/a      95.4μs 
  jax.numpy:cuda      10000       n/a      184μs  
  jax.numpy:cuda      100000      n/a      263μs  
  jax.numpy:cuda     1000000      n/a      2.84ms 
  jax.numpy:cuda     10000000     n/a      43.9ms 
  jax.numpy:cuda    100000000     n/a      617ms  
================== =========== ========== ==========
```

scipy/cluster/hierarchy.py
crusaderky: I've changed it to return `bool` when SCIPY_ARRAY_API is not defined.

You still have a breaking change on `SCIPY_ARRAY_API=1`; most notably `is_isomorphic(a, b) is True` and `is False` (which are commonly regarded as antipatterns) will always return False.

Would this be sufficient? If not, how would you suggest implementing a deprecation cycle for a return type?

rgommers: I wouldn't consider a change from `bool` to `np.bool_` a breaking change - `np.bool_` is designed as a drop-in replacement, and any reasonable usage of the return value will not change in behavior.

> most notably `is_isomorphic(a, b) is True` and `is False` (which are commonly regarded as antipatterns)

Yeah that's a don't-care: user error.

I'm pretty sure we don't want or need a deprecation cycle for something like this.

scipy/cluster/hierarchy.py
rgommers: > As if `unique_all` was not implemented on top of a hashmap...

It is now:) https://numpy.org/devdocs/release/2.3.0-notes.html#performance-improvements-to-np-unique

scipy/cluster/hierarchy.py
crusaderky: Reverted compatibility layer. Now returning `np.bool_` also when SCIPY_ARRAY_API is not set.

crusaderky:
@lucascolley @rgommers can we push this through the finishing line?

lucascolley (commented):
thanks Guido, LGTM!

lucascolley (approved):

lucascolley:
For those interested in `pixi run bench-array-api -t cluster.IsIsomorphic`, NumPy diff vs `main`:

```console
| Change   | Before [dbd8a0c3] <main>   | After [a159ec15] <isomorphic>   |   Ratio | Benchmark (Parameter)                                    |
|----------|----------------------------|---------------------------------|---------|----------------------------------------------------------|
| -        | 32.4±2μs                   | 24.5±2μs                        |    0.76 | cluster.IsIsomorphic.time_is_isomorphic('numpy', 100)    |
| -        | 19.8±2ms                   | 5.35±0.03ms                     |    0.27 | cluster.IsIsomorphic.time_is_isomorphic('numpy', 100000) |
```

timings for array types here:

```console
[100.00%] ··· ================== ============= =============
              --                             nobs           
              ------------------ ---------------------------
                   backend            100          100000   
              ================== ============= =============
                    numpy         19.6±0.07μs   5.26±0.01ms 
               array_api_strict    94.4±0.2μs   5.39±0.01ms 
                     cupy             n/a           n/a     
                  torch:cpu        37.6±0.2μs   1.29±0.01ms 
                  torch:cuda          n/a           n/a     
                  dask.array      5.86±0.02ms    12.5±0.1ms 
                jax.numpy:cpu      38.5±0.6μs    12.7±0.1ms 
                jax.numpy:cuda        n/a           n/a     
              ================== ============= =============
```

timings on `main`:

```console
[100.00%] ··· ================== ============= ============
              --                            nobs           
              ------------------ --------------------------
                   backend            100         100000   
              ================== ============= ============
                    numpy         25.3±0.08μs   16.5±0.2ms 
               array_api_strict    89.7±0.4μs   16.5±0.3ms 
                     cupy             n/a          n/a     
                  torch:cpu        45.0±0.1μs   16.5±0.2ms 
                  torch:cuda          n/a          n/a     
                  dask.array      2.35±0.01ms   19.0±0.2ms 
                jax.numpy:cpu       362±1μs     17.4±0.7ms 
                jax.numpy:cuda        n/a          n/a     
              ================== ============= ============
```

lucascolley:
thank you for your patience Guido :)

PR #22990, opened 2025-05-15
ENH: fix typo in differential_evolution docstring for atol parameter

PriyadarshiAkshay:
#### Reference issue
ENH: fix typo in differential_evolution docstring for atol parameter

In scipy.optimize._differentialevolution.py there is a typo in the description of `atol` parameter in differential_evolution function. In line 207, it should be np.std(population_energies) instead of np.std(pop). This can be checked from the description of `tol` parameter, line 110.


melissawm (approved):
Looks correct to me, thanks @PriyadarshiAkshay !

PR #22992, opened 2025-05-15
ENH: `signal.cspline1d_eval,qspline1d_eval` throw exception for empty cj array

zitongzhoueric:
Empty cj array should be invalid argument to spline evaluation. Without the guard the mirror symmetry implementation would lead to infinite recursion. 

 image


ev-br (approved):
Confirmed it fails with the recursion error on scipy 1.5.2, too, so the problem is not a recent regression. Thus no backwards compat considerations, and a descriptive error is nice. 

Let's land this.
Thanks @zitongzhoueric , keep them coming!

PR #22907, opened 2025-04-29
BENCH: stats.qmc.geometric_discrepancy: add benchmarks

tylerjereddy:
* Add `asv` benchmarks for `geometric_discrepancy`, to accompany the improvements in gh-22841.

[skip actions]

The results do indicate one point of caution, which is the well-known curse of dimensionality. I'll check with Pamphile in the matching PR to see if should add a switch for dimensions (depends on if high dims is a common use case in `qmc`).

`asv continuous -E virtualenv -e -b ".*_geo_discrepancy" main pr_22841`

```
| Change   | Before [6a918e90] <main>   | After [6779c807] <MLopez-Ibanez/main>   |   Ratio | Benchmark (Parameter)                                                                  |
|----------|----------------------------|-----------------------------------------|---------|----------------------------------------------------------------------------------------|
| +        | 3.33±0.08ms                | 12.5±0.05ms                             |    3.74 | stats.BenchQMCGeometricDiscrepancy.time_geo_discrepancy('mindist', 'cityblock', 10)    |
| +        | 3.48±0.02ms                | 4.11±0.03ms                             |    1.18 | stats.BenchQMCGeometricDiscrepancy.time_geo_discrepancy('mindist', 'euclidean', 10)    |
| -        | 100M                       | 88.1M                                   |    0.88 | stats.BenchQMCGeometricDiscrepancy.peakmem_geo_discrepancy('mindist', 'chebyshev', 10) |
| -        | 101M                       | 88.4M                                   |    0.88 | stats.BenchQMCGeometricDiscrepancy.peakmem_geo_discrepancy('mindist', 'chebyshev', 3)  |
| -        | 100M                       | 88.6M                                   |    0.88 | stats.BenchQMCGeometricDiscrepancy.peakmem_geo_discrepancy('mindist', 'cityblock', 10) |
| -        | 100M                       | 88.6M                                   |    0.88 | stats.BenchQMCGeometricDiscrepancy.peakmem_geo_discrepancy('mindist', 'cityblock', 3)  |
| -        | 100M                       | 88.6M                                   |    0.88 | stats.BenchQMCGeometricDiscrepancy.peakmem_geo_discrepancy('mindist', 'euclidean', 10) |
| -        | 100M                       | 88.6M                                   |    0.88 | stats.BenchQMCGeometricDiscrepancy.peakmem_geo_discrepancy('mindist', 'euclidean', 3)  |
[...]

PR #22993, opened 2025-05-16
ENH: `qmc.geometric_discrepancy`: method='mindist' improve memory com…

### LINKED ISSUE #22837: ENH: stats.qmc.geometric_discrepancy: `method='mindist'` can improve memory complexity

MLopez-Ibanez:
…plexity (#22837)

#### Reference issue

Closes gh-22837

Improves memory complexity of `qmc.geometric_discrepancy`.

#### Additional information

This is an alternative version of #22841 that does not suffer from the slowdown of KDTree in high dimensions.

### DIFF
--- a/scipy/stats/_qmc.py
+++ b/scipy/stats/_qmc.py
@@ -442,14 +442,15 @@ def geometric_discrepancy(
     if sample.shape[0] < 2:
         raise ValueError("Sample must contain at least two points")
 
-    distances = distance.pdist(sample, metric=metric)  # type: ignore[call-overload]
-
-    if np.any(distances == 0.0):
-        warnings.warn("Sample contains duplicate points.", stacklevel=2)
-
     if method == "mindist":
-        return np.min(distances[distances.nonzero()])
+        min_d = distance._pmindist(sample, metric=metric)
+        if np.isclose(min_d, 0.0):
+            warnings.warn("Sample contains duplicate points.", stacklevel=2)
+        return min_d
     elif method == "mst":
+        distances = distance.pdist(sample, metric=metric)  # type: ignore[call-overload]
+        if np.any(distances == 0.0):
+            warnings.warn("Sample contains duplicate points.", stacklevel=2)
         fully_connected_graph = distance.squareform(distances)
         mst = minimum_spanning_tree(fully_connected_graph)
         distances = mst[mst.nonzero()]
--- a/scipy/stats/_qmc.py
+++ b/scipy/stats/_qmc.py
@@ -443,7 +443,7 @@ def geometric_discrepancy(
         raise ValueError("Sample must contain at least two points")
 
     if method == "mindist":
-        min_d = distance._pmindist(sample, metric=metric)
+        min_d = distance.pdist(sample, metric=metric).min()
         if np.isclose(min_d, 0.0):
             warnings.warn("Sample contains duplicate points.", stacklevel=2)
         return min_d
--- a/scipy/spatial/distance.py
+++ b/scipy/spatial/distance.py
@@ -106,10 +106,6 @@ import warnings
 import dataclasses
 from collections.abc import Callable
 from functools import partial
-from typing import TYPE_CHECKING
-
-if TYPE_CHECKING:
-    import numpy.typing as npt
 
 import numpy as np
 
@@ -2109,34 +2105,6 @@ def pdist(X, metric='euclidean', *, out=None, **kwargs):
                           shape=((n * (n - 1)) // 2, ), dtype=X.dtype,
                           as_numpy=True, metric=metric, **kwargs)
 
-def _pmindist(sample: "npt.ArrayLike", metric: str) -> float:
-    """Mininum distance between points in the given sample.
-
-    Parameters
-    ----------
-    sample : array_like (n, d)
-        The sample to compute the minimum distance from.
-    metric : str or callable, optional
-        The distance metric to use. See the documentation
-        for `scipy.spatial.distance.pdist` for the available metrics and
-        the default.
-
-    Returns
-    -------
-    distance : float
-        Minimum distance.
-
-    """
-    sample = _asarray(sample)
-    n = sample.shape[0]
-    d = cdist(sample[0:1,...], sample[1:,...], metric=metric).min()
-    if np.isclose(d, 0.0):
-        return 0.0
-
-    for i in range(2, n):
-        d = min(d, cdist(sample[(i-1):i,...], sample[i:,...], metric=metric).min())
-    return d
-
 
 def _np_pdist(X, out, w, V, VI, metric='euclidean', **kwargs):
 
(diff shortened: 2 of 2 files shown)


tylerjereddy (commented):
Thanks for prototyping the alternate approach as well (this was indeed your original suggestion).

I ran the benchmarks I drafted for evaluating these proposals in gh-22907, and as I speculated at https://github.com/scipy/scipy/pull/22841#issuecomment-2887197041, there is some time performance cost to pay for the space complexity improvement (see below) since we are batching in pure Python.

Now that you've presented both approaches we can carefully consider the path forward--it may not happen today, but I'll try to push in whatever changes might be needed if you need to move on.

On x86_64 Linux with `i9-13900K`:

`asv continuous -E virtualenv -e -b ".*_geo_discrepancy" main no-kde-tree`

```
| Change   | Before [a011b598] <main>   | After [3a3205ea] <no-kde-tree>   |   Ratio | Benchmark (Parameter)                                                                  |
|----------|----------------------------|----------------------------------|---------|----------------------------------------------------------------------------------------|
| +        | 3.47±0.05ms                | 4.68±0.02ms                      |    1.35 | stats.BenchQMCGeometricDiscrepancy.time_geo_discrepancy('mindist', 'cosine', 10)       |
| +        | 3.14±0ms                   | 4.26±0.03ms                      |    1.35 | stats.BenchQMCGeometricDiscrepancy.time_geo_discrepancy('mindist', 'cosine', 3)        |
| -        | 100M                       | 88.6M                            |    0.88 | stats.BenchQMCGeometricDiscrepancy.peakmem_geo_discrepancy('mindist', 'chebyshev', 10) |
| -        | 100M                       | 88.6M                            |    0.88 | stats.BenchQMCGeometricDiscrepancy.peakmem_geo_discrepancy('mindist', 'chebyshev', 3)  |
| -        | 100M                       | 88.6M                            |    0.88 | stats.BenchQMCGeometricDiscrepancy.peakmem_geo_discrepancy('mindist', 'cityblock', 10) |
| -        | 100M                       | 88.6M                            |    0.88 | stats.BenchQMCGeometricDiscrepancy.peakmem_geo_discrepancy('mindist', 'cityblock', 3)  |
| -        | 99.9M                      | 88.1M                            |    0.88 | stats.BenchQMCGeometricDiscrepancy.peakmem_geo_discrepancy('mindist', 'cosine', 10)    |
| -        | 100M                       | 88.1M                            |    0.88 | stats.BenchQMCGeometricDiscrepancy.peakmem_geo_discrepancy('mindist', 'cosine', 3)     |
| -        | 101M                       | 88.4M                            |    0.88 | stats.BenchQMCGeometricDiscrepancy.peakmem_geo_discrepancy('mindist', 'euclidean', 10) |
| -        | 101M                       | 88.4M                            |    0.88 | stats.BenchQMCGeometricDiscrepancy.peakmem_geo_discrepancy('mindist', 'euclidean', 3)  |

PERFORMANCE DECREASED.
```

MLopez-Ibanez:
Since this was going nowhere, I decided to simplify it. This version simply uses pmin(), which is hopefully faster than cdist() and has a smaller memory cost. It is still quadratic memory, so it doesn't solve the main problem I pointed out, but it is an improvement.

scipy/stats/_qmc.py
```diff
@@ -442,14 +442,15 @@ def geometric_discrepancy(
...
-    distances = distance.pdist(sample, metric=metric)  # type: ignore[call-overload]
-
-    if np.any(distances == 0.0):
-        warnings.warn("Sample contains duplicate points.", stacklevel=2)
-
     if method == "mindist":
-        return np.min(distances[distances.nonzero()])
+        min_d = distance.pdist(sample, metric=metric).min()
+        if np.isclose(min_d, 0.0):
+            warnings.warn("Sample contains duplicate points.", stacklevel=2)
```
mdhaber, line 448: This now has a slightly different criterion for warning. Is it correct to warn about duplicate points if the points are just close?

scipy/stats/_qmc.py
```diff
@@ -442,14 +442,15 @@ def geometric_discrepancy(
...
+        return min_d
```
mdhaber, line 449: Now, this might return `0`. It looks like the original author was avoiding that.

mdhaber (commented):
ISTM that both branches use `pdist`, so `distances` could be computed outside the branching as it was before. The other changes alter the logic - but could you explain why those changes are OK? It looks like the original author was careful to avoid returning `0` in the `mindist` branch.

Is the main idea that this avoids `np.min(distances[distances.nonzero()])`, which could effectively duplicate `distances`?

If so, could you tolerate a temporary boolean array with the same shape as distances (but which uses a fraction of the memory)? If so, you could preserve the logic but save some memory by reverting all the changes except doing this in the `mindist` branch:
```python3
distances[distances == 0] = np.inf
return np.min(distances)
```

mdhaber:
@MLopez-Ibanez @tylerjereddy were you still interested in pursuing this/gh-22907?

MLopez-Ibanez:
This has now diverted so much from the original proposal that I'm not sure where it is going. I have created my own implementation outside scipy that doesn't suffer from the problems reported here and can also calculate other types of discrepancy. Somebody else will have to figure out how to fix gh-22837.

tylerjereddy:
> Somebody else

Probably me, yeah. The tradeoffs were a bit annoying to deal with though. Sorry for the frustrating contributor experience. If a future solution uses part of one of your approaches we'll try to credit you.

PR #22996, opened 2025-05-16
MAINT: _lib: bump vendored array_api_compat to 1.12

ev-br:


rgommers:
Does it fix anything, or just good to have?

ev-br:
The latter, most likely. 
There is an incompatibility of array-api-compat==1.11.2 and pytorch==2.7 , https://github.com/data-apis/array-api-compat/issues/320, but our vendored copy pinned some in-development commit which likely contained the fix already. So this PR is not critical for scipy 1.16, I suppose.

ev-br:
superseded by gh-22999 (do we not get PR races now :-))

PR #22997, opened 2025-05-16
CI: temporarily disable free-threaded job with parallel-threads

rgommers:
Reasons:
1. A segfault in `stats/_distribution_infrastructure.py` since ~1 week, see https://github.com/scipy/scipy/issues/22758#issuecomment-2883968653
2. Preferred not to have this not-so-stable job running on the 1.16.x branch

Re-enable on `main` after 1.16.x has branched and (1) is investigated.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -456,7 +456,7 @@ jobs:
     needs: get_commit_message
     strategy:
       matrix:
-        parallel: ['0', '1']
+        parallel: ['0']  # '1' temporarily disabled, `stats` segfault (see gh-22758)
 
     runs-on: ubuntu-latest
     if: >


tylerjereddy (approved):
Makes sense, I tried to reproduce locally as described at https://github.com/scipy/scipy/issues/22758#issuecomment-2887351400, but ran into other issues rather than the segfault... anyway, let's put this in so the release branch has a sensible CI for now and we can consider backports/reactivating later.

tylerjereddy:
thanks Ralf

mdhaber:
For future reference, would putting `@pytest.mark.thread_unsafe` on the revealing test do the trick?

tylerjereddy:
Ya, my experience described at https://github.com/scipy/scipy/issues/22758#issuecomment-2887351400 was almost identical to yours I think. As I note there:

> In the full testsuite via spin test -- --parallel-threads=4 I was seeing crazy slow timing just to collect the tests for some reason... slow enough that I gave up for the moment.

Similar to you, I also pondered running `stats` stuff in isolation vs. full suite, but the latter was intractably slow to collect tests for me as well.

I wonder if using slightly-outdated versions of `3.13.x` or the supporting tools/libs can lead to pretty bad experiences still (possibly/probably?).

> For future reference, would putting @pytest.mark.thread_unsafe on the revealing test do the trick if we don't want to disable the whole element of the job?

I would have thought so--that's why I was trying to reproduce it (probably you had a similar motivation). I suspect we will want to backport a fix at some point and perhaps reactivate that job eventually.

I suspect the team is just noticing that things are a bit behind schedule and trying to reduce the barriers as much as possible, which makes sense for now I suppose.

rgommers:
> For future reference, would putting `@pytest.mark.thread_unsafe` on the revealing test do the trick if we don't want to disable the whole element of the job?

Yes, it avoids the problem. _However_, please only put a `thread_unsafe` marker on something if you actually understand why the test is failing. If it's a known issue then we skip it, but if the code under test is actually unsafe, it should be fixed rather than the problem hidden. At a minimum we then need a bug report to track the problem so it can be investigated/fixed later.

> Similar to you, I also pondered running `stats` stuff in isolation vs. full suite, but the latter was intractably slow to collect tests for me as well.

Running only the submodule (or test file/class/function) that has a problem is always more efficient. Full test suite collection takes about 45 seconds at the moment for the default config, and I think with `--parallel-threads=4` it may get 4x slower. Is that what you saw, or is it way more than that?

mdhaber:
> Yes, it avoids the problem

I think I see. So same as standard practice for at least the past eight years, right?

> Running only the submodule (or test file/class/function) that has a problem is always more efficient.

That's what I did in gh-23002, but it doesn't seem to be reproducing the problem.

Collection takes less than a second when I run `spin test -t scipy.stats.tests.test_continuous` and at least 45 seconds when I run `spin test -t scipy.stats.tests.test_continuous -- --parallel-threads=4`. I'm using an editable install on Windows, if it makes a difference.

rgommers:
> Collection takes less than a second when I run `spin test -t scipy.stats.tests.test_continuous` and at least 45 seconds when

That's bad indeed. The problem is a recent regression in `pytest-run-parallel` 0.4.0: https://github.com/Quansight-Labs/pytest-run-parallel/issues/56. For now, downgrading to 0.3.1 will avoid the problem - that may then run into some failures related to not auto-detection `pytest.warns`, `eager_warns` & co, but that can be ignored so it may still be a helpful workaround.

> So same as standard practice for at least the past eight years, right?

Yep.

> I wonder if using slightly-outdated versions of `3.13.x` or the supporting tools/libs can lead to pretty bad experiences still (possibly/probably?).

Perhaps, yes. The difference between CPython 3.13.2 and 3.13.3 is not large anymore, the flood of backports in the RCs and .0/.1 releases slowed down a lot. And NumPy 2.2.6 is also fine. Some other libraries/tools with more recent support may show larger differences.

PR #22946, opened 2025-05-08
ENH: `special`: add `xp_capabilities`

crusaderky:
- Add `@xp_capabilities` to the ufuncs in `scipy.special._support_alternative_backends`
- Speed up function dispatch by ~4us by applying the same caching pattern already used in array-api-compat helpers
- Fix bug where the wrapped functions could not be pickled
- Fix bug where the wrapped functions would have poor `repr()`, `__qualname__`, and `__module__`

### Demo render
image

PR #23001, opened 2025-05-16
DOC: `special`: update top-level docs to reflect `xp_capabilities`

crusaderky:
Follow-up to #22946
CC @steppi


steppi (approved):
Looks good. If you make another docs only PR like this in the future, you can put `[docs only]` at the end of your commit messages to skip all checks except Circle CI.

PR #22991, opened 2025-05-15
DOC: update SciPy 1.16.0 release notes

tylerjereddy:
* Update the SciPy `1.16.0` release notes prior to branching. I'm a bit behind schedule--we had 50 milestone-labeled PRs at the start of the week are we now hovering around 25. I'll probably have to get tougher about bumping milestones for things that aren't 1-2 days away from crossing the line.

[ci skip] [skip ci]

TODO:

- [ ] `.mailmap` updates
- [x] warnings in the doc build
- [x] manually scanning through release note PRs with `enhancement` labels (Matt will be doing this for `stats`)
- [x] check the release notes Wiki again for updates--initial transcription was performed on May 15/2025; checked again on May 17th
- [x] release notes highlights section
- [x] gradually reactivate the CI here, when ready
- [x] Remove unused/empty (sub)sections in the release notes
- [x] re-run author/list scripts again just before branching

PR #23004, opened 2025-05-16
DOC: duplicate contributor

crusaderky:


tylerjereddy (commented):
We usually handle this by adjusting `.mailmap`, otherwise we'd need to manually edit each time after running the author script.

This also shows up in gh-22991 for the upcoming `1.16.0` release, so is indeed useful to point out (thanks).

I don't know if we'd want to go back and start removing duplicates from previous releases like this, but certainly for the new draft release notes a patch like the following does the trick and is worth adding over there:

```diff
--- a/.mailmap
+++ b/.mailmap
@@ -207,7 +207,8 @@ Franziska Horn <cod3licious@users.noreply.github.com> cod3licious <cod3licious@u
 Fukumu Tsutsumi <levelfourslv@gmail.com> levelfour <levelfourslv@gmail.com>
 G Young <gfyoung17@gmail.com> gfyoung <gfyoung17@gmail.com>
 G Young <gfyoung17@gmail.com> gfyoung <gfyoung@mit.edu>
-Gagandeep Singh <gdp.1807@gmail.com> czgdp1807 <gdp.1807@gmail.com>
+ਗਗਨਦੀਪ ਸਿੰਘ (Gagandeep Singh) <gdp.1807@gmail.com> czgdp1807 <gdp.1807@gmail.com>
+ਗਗਨਦੀਪ ਸਿੰਘ (Gagandeep Singh) <gdp.1807@gmail.com> Gagandeep Singh <gdp.1807@gmail.com>
 Ganesh Kathiresan <ganesh3597@gmail.com> ganesh-k13 <ganesh3597@gmail.com>
```

czgdp1807 (changes requested):
`ਗਗਨਦੀਪ ਸਿੰਘ (Gagandeep Singh)` is preferable to me. Thanks for unifying my contributions  under this name.

crusaderky:
@tylerjereddy gentle ping

tylerjereddy:
I'm a bit confused by the `.mailmap` change--there is more than one alias we need to contend with (hence the two remappings in `.mailmap`) and when I pull the `.mailmap` change from here into the branch at gh-23098 for example, I see this change to the author list, which I don't think we want, when I run `python tools/authors.py 36733273bac61e5e..9babbe47 >& authors_draft.txt` over there:

```diff
--- a/authors_draft.txt
+++ b/authors_draft.txt
@@ -95,6 +95,7 @@ Authors
 * Scott Shambaugh (4)
 * ShannonS00 (1) +
 * sildater (3) +
+* Gagandeep Singh (32)
 * Param Singh (1) +
 * G Sreeja (7) +
 * Albert Steppi (133)
@@ -125,9 +126,9 @@ Authors
 * Zhenyu Zhu ajz34 (1) +
 * Eric Zitong Zhou (11) +
 * Case Zumbrum (2) +
-* ਗਗਨਦੀਪ ਸਿੰਘ (Gagandeep Singh) (45)
+* ਗਗਨਦੀਪ ਸਿੰਘ (Gagandeep Singh) (13)
 
-    A total of 124 people contributed to this release.
+    A total of 125 people contributed to this release.
```

You're basically reverting the fix for the problem I just merged that removed the duplication in gh-22991. Most likely you manually deleted the duplicated names and then assumed that you can just purge the alias in the `.mailmap` or something, but in any case, I'd probably vote to just close this. If we really want to salvage this, we can just purge the duplicated names from the old release notes (though I'm not sure that's worth doing in general, once the release is out, maybe...), but let's not move backwards with the `.mailmap` stuff.

crusaderky:
just closing makes sense.

PR #23005, opened 2025-05-17
BUG: sparse: fix mean/sum change in return of np.matrix for sparse matrices

dschult:
This PR corrects a mistake from #22873 where `spmatrix.mean` and `spmatrix.sum` no longer return `np.matrix`. The trouble was a dropped `self._ascontainer(...)` during creation of the return object. 

It also adds tests to check the return object type of (array, matrix, sparray, spmatrix) for methods (sum, mean,  min, max, nanmin, nanmax, argmin, argmax).

This bug was spotted by @lesteve from scikit-learn based on [tests against the nightly wheels](https://github.com/scikit-learn/scikit-learn/issues/31370)!   Thanks @lesteve 

cc: @perimosocordiae @stefanv and/or @j-bowhay

### DIFF
--- a/scipy/sparse/_base.py
+++ b/scipy/sparse/_base.py
@@ -1343,7 +1343,7 @@ class _spbase(SparseABC):
 
         if out is None:
             # create out array with desired dtype
-            out = np.zeros(new_shape, dtype=dtype or res_dtype)
+            out = self._ascontainer(np.zeros(new_shape, dtype=dtype or res_dtype))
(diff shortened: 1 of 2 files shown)


stefanv (approved):

stefanv:
Go ahead and merge on green, given that this is an important fix.

lesteve:
Great, thanks a lot for the fix @dschult!

Issue #23007, opened 2025-05-17
CircleCI docs artifact redirector action affect by change in security setting


rgommers:
In https://github.com/scipy/scipy/pull/22967#pullrequestreview-2831335239 we changed a security-related setting for GitHub Actions, changing the default permissions for `GITHUB_TOKEN` from _write_ to _read_ (see [docs](https://docs.github.com/en/organizations/managing-organization-settings/disabling-or-limiting-github-actions-for-your-organization#configuring-the-default-github_token-permissions)).

Since then, the redirector has been failing ([example log](https://github.com/scipy/scipy/actions/runs/14987831958/job/42105000405)):

```
RequestError [HttpError]: Resource not accessible by integration
    at /home/runner/work/_actions/scientific-python/circleci-artifacts-redirector-action/4e13a10d89177f4bfc8007a7064bdbeda848d8d1/dist/index.js:8573:21
```

<details>
<summary>Full log contents:</summary>

```
> Run scientific-python/circleci-artifacts-redirector-action@4e13a10d89177f4bfc8007a7064bdbeda848d8d1
  with:
    repo-token: ***
    api-token: ***
    artifact-path: 0/html/index.html
    circleci-jobs: build_docs
    job-title: Check the rendered docs here!
/home/runner/work/_actions/scientific-python/circleci-artifacts-redirector-action/4e13a10d89177f4bfc8007a7064bdbeda848d8d1/dist/index.js:8573
      const error = new requestError.RequestError(toErrorMessage(data), status, {
                    ^

RequestError [HttpError]: Resource not accessible by integration
    at /home/runner/work/_actions/scientific-python/circleci-artifacts-redirector-action/4e13a10d89177f4bfc8007a7064bdbeda848d8d1/dist/index.js:8573:21
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  status: 403,
  response: {
    url: 'https://api.github.com/repos/scipy/scipy/statuses/933ea9d3f818c663a3e1da5b30b6763f3dd4a44b',
    status: 403,
    headers: {
      'access-control-allow-origin': '*',
[...]

PR #23008, opened 2025-05-17
CI: give redirect action status write permission

nickodell:
Give the redirect action permission to write statuses

#### Reference issue

#23007

The `scientific-python/circleci-artifacts-redirector-action` action uses the `client.rest.repos.createCommitStatus()` function, which I believe requires `statuses: write` permission.

#### Additional information

Note: I think this ought to fix it, but I am opening this PR to run CI.

### DIFF
--- a/.github/workflows/circle_artifacts.yml
+++ b/.github/workflows/circle_artifacts.yml
@@ -3,7 +3,8 @@ name: Redirect circleci artifacts
 on: [status]
 
 permissions:
-   contents: read  # to fetch code (actions/checkout)
+  contents: read  # to fetch code (actions/checkout)
+  statuses: write  # to report circleci status (scientific-python/circleci-artifacts-redirector-action)
 
 jobs:
   circleci_artifacts_redirector_job:


larsoner:
Annoyingly you won't know until it's merged in `main` because the commit status event action is executed from there even for PRs IIRC.

One way to deal with this annoyance without creating too much noise in the main repo is to replicate the problem in your fork, then push commits to `main` in your fork (with actions enabled) and a PR open in your fork to see if it works there.

nickodell:
> One way to deal with this annoyance without creating too much noise in the main repo is to replicate the problem in your fork, then push commits to `main` in your fork (with actions enabled) and a PR open in your fork to see if it works there.

Thanks, will do.

rgommers:
Thanks @nickodell for looking into this. I think that this shouldn't work, because of (from the [docs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#changing-the-permissions-in-a-forked-repository):

*You can use the permissions key to add and remove read permissions for forked repositories, but typically you can't grant write access. The exception to this behavior is where an admin user has selected the Send write tokens to workflows from pull requests option in the GitHub Actions settings.*

Hence it'd be good to test on the `main` branch of your fork indeed.

If this doesn't work indeed, can you try changing `GITHUB_TOKEN` to `SCIPY_CIRCLECI_REDIRECTOR_TOKEN`? I think that should do it.

nickodell:
Currently, I'm having some trouble testing this because of https://github.com/scientific-python/circleci-artifacts-redirector-action/issues/65.

I'm also having trouble building this action to modify it. I've opened https://github.com/scientific-python/circleci-artifacts-redirector-action/issues/70 for that reason. @larsoner Would you have any suggestions about how to fix either issue? Happy to help resolve the target_url parsing issue.

.github/workflows/circle_artifacts.yml
```diff
@@ -3,7 +3,8 @@ name: Redirect circleci artifacts
 on: [status]
 
 permissions:
-   contents: read  # to fetch code (actions/checkout)
+  contents: read  # to fetch code (actions/checkout)
```
larsoner, line 6: I don't think you need this one actually. And https://github.com/scientific-python/circleci-artifacts-redirector-action/issues/70 is merged and `v1.1.0` is out so I'd suggest to update to that and merge this and see if it works. If it doesn't, then we could even try using https://github.com/scientific-python/circleci-artifacts-redirector-action/pull/66 in `main` for a short while and merge that PR if it does fix things...

lucascolley:
any update here @larsoner @nickodell ? It would be really useful to have the docs previews back up-and-running!

nickodell:
> any update here @larsoner @nickodell ? It would be really useful to have the docs previews back up-and-running!

Apologies, no. I have not been able to create a working 'dev environment' to check my work.

I started writing a test suite for the circleci-artifacts-redirector-action action, so I could debug the [target_url parsing PR](https://github.com/scientific-python/circleci-artifacts-redirector-action/pull/66) locally, but Javascript is not my strong suit. I've figured out how to mock the initial event, but I got stuck trying to mock either octokit or the API call to circle - I forget which.

I know that that PR has a bug with respect to pipelines where the docs job is not the first job in the workflow (as in SciPy) but I have not gotten further than that.

For that reason, I am not sure if the permissions change in this PR would fix things.

nickodell:
@lucascolley I don't have a way to test this PR before merging, but it should be relatively low-risk to try merging it.

What would you think about merging this PR, and checking if it helps? If it doesn't, we can revert it.

That way I can find out if I'm on the right track. If not, we can try one of the other options mentioned in #23007.

lucascolley:
yeah, let's give it a go!

lucascolley (approved):

nickodell:
That looks promising so far.

Screenshot 2025-08-21 at 10 15 17 AM

When testing this, I found pressing "rerun failed jobs" on failed "Redirect circleci artifacts" actions does not work. I think this might be because manually re-running jobs uses the permissions from the branch, not from main. Not sure though.

PR #21180, opened 2024-07-14
ENH: sparse: efficient arithmetic operations for DIA format

MikhailRyazanov:
#### Reference issue
Partially addresses gh-21039.

#### What does this implement/fix?
Implements most important arithmetic operations with sparse arrays in DIA format. Previously most operations used fall-back conversion to CSR format with very poor overall efficiency.

For comparison, the standard benchmarks (``python dev.py bench -s sparse.Arithmetic``):

##### Before
```
sparse.Arithmetic.time_arithmetic                        
======== ==== ========== ========== ========== ==========
--                                 op                    
------------- -------------------------------------------
 format   XY   __add__    __sub__    multiply   __mul__  
======== ==== ========== ========== ========== ==========
  csr     AA   1.95±0ms   1.15±0ms   1.97±0ms   10.4±0ms 
  csr     AB   7.12±0ms   13.8±0ms   7.64±0ms   21.1±0ms 
  csr     BA   6.89±0ms   7.05±0ms   6.52±0ms   24.4±0ms 
  csr     BB   7.06±0ms   5.21±0ms   6.98±0ms   46.0±0ms 
  csc     AA   3.63±0ms   6.53±0ms   3.49±0ms   10.2±0ms 
  csc     AB   6.55±0ms   6.45±0ms   7.40±0ms   24.4±0ms 
  csc     BA   6.40±0ms   6.43±0ms   7.31±0ms   20.8±0ms 
  csc     BB   7.04±0ms   5.65±0ms   7.07±0ms   44.9±0ms 
  coo     AA   18.5±0ms   15.4±0ms   17.9±0ms   27.2±0ms 
  coo     AB   25.9±0ms   26.0±0ms   26.2±0ms   42.0±0ms 
  coo     BA   25.8±0ms   29.4±0ms   26.5±0ms   49.8±0ms 
  coo     BB   35.0±0ms   30.8±0ms   33.0±0ms   72.9±0ms 
  dia     AA   1.44±0ms   28.2±0ms   29.3±0ms   38.2±0ms 
  dia     AB   2.37±0ms   46.7±0ms   46.8±0ms   61.5±0ms 
  dia     BA   2.16±0ms   46.0±0ms   42.3±0ms   66.9±0ms 
  dia     BB   1.51±0ms   64.6±0ms   66.8±0ms   108±0ms  
======== ==== ========== ========== ========== ==========
```
(Only addition was implemented.)

##### After
```
sparse.Arithmetic.time_arithmetic                        
[...]

PR #23009, opened 2025-05-17
ENH: sparse: efficient conversion from DIA to CSR format

### LINKED ISSUE #21039: ENH: Implement all relevant methods for `sparse.dia_array`

Currently almost all multiplicative operations on `dia_array` convert it to `csr_array`, which is inefficient and unnecessary.
#20445 has enabled scalar multiplication, but element-wise multiplication by a row or column (with broadcasting) and matrix (“dot”) multiplication by another `dia_array` still involve conversion to `csr_array`, even though all these operations should produce produce sparse diagonal arrays and can be efficiently performed directly in the diagonal representation.

Element-wise multiplication and division by an `ndarray` row (with broadcasting) can be done simply by `self.data * other` and `self.data / other` (or using a correspondingly truncated `other`, if `self.data` doesn't span the full `dia_array` width).

Element-wise multiplication/division by an `ndarray` column (with broadcasting) is slightly more complicated and involves a loop over `self.offsets` but also shouldn't change the array structure and is more efficient than conversion to `csr_array` (and significantly more efficient that conversion back to `dia_array`, if that format is needed for the result).

Matrix (“dot”) multiplication of two sparse diagonal arrays can also be done by operating directly on their diagonals, and even with a Python loop seems to be more efficient than going through conversion to `csr_array` (moreover, theoretically, with a lower-level implementation, it must be more much efficient than `csr_array` multiplication due to operating on sequential data instead of random indirect memory access).

The results can be explicitly cast to `dia_array`, but that is quite inefficient.

I have some sketches that work for square `dia_array`s with full-length diagonals (my interest was limited to efficient storage and operations on covariance and correlation matrices) and can share them here if anybody is interested in implementing and optimizing this for the general case.

MikhailRyazanov:
#### Reference issue

Addendum to gh-21180, should close gh-21039.

Direct and efficient conversion from DIA format to CSR format, to speed-up all operations not directly implemented for DIA.

#### Additional information
After merging gh-21180, almost all arithmetic operations with DIA arrays became faster than conversion from DIA to other formats, needed for other operations, such as arithmetic with arrays in other formats and passing DIA arrays to methods not implemented for them directly.

For comparison — arithmetic operations:
```
sparse.Arithmetic.time_arithmetic
======== ==== ============= ============= ============= =============
--                                       op                          
------------- -------------------------------------------------------
 format   XY     __add__       __sub__       multiply      __mul__   
======== ==== ============= ============= ============= =============
  dia     AA     600±20μs      617±20μs    1.34±0.05ms    6.18±0.2ms 
  dia     AB   2.18±0.01ms   2.92±0.02ms   1.33±0.04ms     12.8±2ms  
  dia     BA   2.22±0.04ms   2.21±0.05ms   1.28±0.04ms     12.7±2ms  
  dia     BB   1.26±0.01ms   1.32±0.02ms   3.17±0.04ms    26.5±0.3ms 
======== ==== ============= ============= ============= =============
```
and conversions:
```
sparse.Conversion.time_conversion
============= ============= ============= ============= ============= ============ ============ =============
--                                                       to_format                                           
------------- -----------------------------------------------------------------------------------------------
 from_format       csr           csc           coo           dia          lil          dok           bsr     
============= ============= ============= ============= ============= ============ ============ =============
     dia       1.74±0.05ms     866±40μs      691±10μs      212±5ns     19.8±0.5ms   22.4±0.3ms   2.28±0.06ms 
============= ============= ============= ============= ============= ============ ============ =============
```
(keep in mind that conversion benchmarks use $100^2 \times 100^2$ arrays, while arithmetic benchmarks are for $250^2 \times 250^2$, and generally with more diagonals, so the relative performance of conversions is even worse than these benchmarks show). This is because currently direct conversion is implemented only to CSC and COO, which is quite inefficient.

The direct and efficient conversion to CSR format implemented in this PR gives it ~5-fold speed-up (which also significantly improves conversion to BSR and slightly to LIL and DOK), making it no longer much slower than arithmetic operations:
```
sparse.Conversion.time_conversion
============= ============= ============= ============= ============= ============ ============ =============
--                                                       to_format                                           
------------- -----------------------------------------------------------------------------------------------
 from_format       csr           csc           coo           dia          lil          dok           bsr     
============= ============= ============= ============= ============= ============ ============ =============
     dia         273±2μs       814±5μs       597±8μs       217±9ns     17.4±0.1ms   20.5±0.07ms     781±2μs   
============= ============= ============= ============= ============= ============ ============ =============
```

<details>
<summary>P.S.</summary>

* All the benchmarks above were done without the `--quick` option for `asv`, for better reproducibility and more meaningful numbers.
* When I tried to use the `--compare main` approach, it apparently started rebuilding the whole SciPy instead of just recompiling the changed files. Am I missing something?
</details>

UPD: Conversion to CSC and COO formats can also be made faster by simply removing current `tocsc()` and `tocoo()` methods, see below.


MikhailRyazanov:
The failed tests seem to be failing due to some completely unrelated network problems; there is nothing I can do, but perhaps maintainers can check what is wrong with them and rerun.

MikhailRyazanov:
By the way, simply removing the current Python + NumPy implementations of `.tocsc()` and `.tocoo()` methods, thus making these conversions to be done indirectly as DIA → CSR → CSC/COO in fact speeds them up by a factor of ~2:
```
============= ============= ============= ============= ============= ============= ============= =============
--                                                        to_format                                            
------------- -------------------------------------------------------------------------------------------------
 from_format       csr           csc           coo           dia           lil           dok           bsr     
============= ============= ============= ============= ============= ============= ============= =============
     dia         270±2μs       484±7μs       386±3μs       207±2ns      17.2±0.2ms    21.2±0.1ms     777±5μs   
============= ============= ============= ============= ============= ============= ============= =============
```

However, [`TestDIA::test_tocoo_gh10050`](https://github.com/scipy/scipy/blob/7ef6aeb632f4d99078d1332108455c401a11da10/scipy/sparse/tests/test_base.py#L5032-L5037) starts failing because the converted COO matrix has `.has_canonical_format = False` even though it is in fact canonical. This seems to be caused by the COO constructor indiscriminately setting
https://github.com/scipy/scipy/blob/7ef6aeb632f4d99078d1332108455c401a11da10/scipy/sparse/_coo.py#L63
even for canonical data. So there are two possible solutions:
1. Relax the test such that it fails only for unsafe results (meaning that `m.has_canonical_format == False` is safe when `inds_are_sorted == True`, like it is implied by the COO constructor, but the opposite is an error).
2. Explicitly add `tocoo()` to `_csr.py` which will transfer `has_canonical_format` to the conversion result (as far as I understand, canonical CSR always produces canonical COO).

MikhailRyazanov:
While implementing the second “solution” from the previous comment, I've discovered missing bound checks for ill-formed DIA arrays (when diagonal are longer than the array width or lie completely outside), so this is corrected in [677f10b](https://github.com/scipy/scipy/pull/23009/commits/677f10b963052f94aa3ab82af618b3d11a6e189a).

The remaining new commits are related to eliminating the inefficient `tocsc()` and `tocoo()` methods and then dealing with the revealed deficiencies in COO construction. I'm not sure that the last change ([1d03664](https://github.com/scipy/scipy/pull/23009/commits/1d03664c9f24d1da82887e6ddeed32ae46ff3b3b)) is completely correct, but it seems to be more appropriate than setting the default `copy=True` in all `tocoo()` methods (or assuming that internal/intermediate operations will copy the relevant data for methods with default `copy=False`).

MikhailRyazanov:
Here are more extensive benchmarks for arrays of various sizes and comparing conversions with arithmetic operations on equal footing. `main:` means how it is now, `this:` means this PR; arithmetic benchmarks are not affected.

Using `poisson2d(20, ...)` = 400 × 400 array:
```
      sparse.Conversion.time_conversion
      --                                      to_format
      ------------- ---------------------------------------------------------
       from_format      csr        csc       coo      lil     dok      bsr   
      ============= =========== ========= ========= ======= ======= =========
main:      dia      225±2μs     122±10μs  117±0.3μs 882±3μs 721±5μs 296±0.3μs
this:      dia       57.0±0.3μs 104±0.7μs 111±0.3μs 648±5μs 522±2μs 114±0.3μs

sparse.Arithmetic.time_arithmetic
--                              op
------------- ---------------------------------------
 format   XY     __add__      __sub__      multiply  
======== ==== ============= ============ ============
  dia     AA   43.0±0.5μs    43.3±0.4μs   64.9±0.5μs  
```

Using `poisson2d(100, ...)` = 10000 × 10000 array:
```
      sparse.Conversion.time_conversion
      --                                    to_format
      ------------- ---------------------------------------------------------------
       from_format      csr       csc      coo       lil        dok         bsr    
      ============= =========== ======== ======== ========== ========== ===========
main:     dia       1.74±0.05ms 866±40μs 691±10μs 19.8±0.5ms 22.4±0.3ms 2.28±0.06ms
this:     dia         270±2μs   484±7μs  386±3μs  17.2±0.2ms 21.2±0.1ms   777±5μs  

sparse.Arithmetic.time_arithmetic
--                               op
------------- -----------------------------------------
 format   XY     __add__       __sub__       multiply  
======== ==== ============= ============= =============
  dia     AA    99.8±0.2μs    101±0.3μs      168±1μs   
```

Using `poisson2d(250, ...)` = 62500 × 62500 array:
```
      sparse.Conversion.time_conversion
      --                                        to_format
      ------------- ----------------------------------------------------------------
       from_format     csr         csc        coo        lil       dok        bsr   
      ============= =========== ========== ========== ========= ========= ==========
main:      dia      15.5±0.6ms  7.31±0.6ms 8.89±1ms   126±1ms   142±1ms   19.3±0.6ms
this:      dia       4.55±0.3ms 6.65±0.1ms 5.24±0.3ms 114±0.7ms 130±0.4ms 7.99±0.2ms

sparse.Arithmetic.time_arithmetic
--                               op
------------- -----------------------------------------
 format   XY     __add__       __sub__       multiply  
======== ==== ============= ============= =============
  dia     AA     600±20μs      617±20μs    1.34±0.05ms 
```

So conversion from DIA becomes 3–6 times faster to CSR, >2 times faster to BSR, and somewhat faster to all other formats. Still slower than simple arithmetic operations, but not as much as it currently is (like an order of magnitude for the 10000 × 10000 size).

dschult (commented):
Thanks for this! 
The timing results look good -- and I agree with your choice to implement the second approach to `has_canonical_format` in `tocoo`. Adding the `tocoo` method and setting `has_canonical_format` inside `_csrbase.tocoo` should work fine. The result of conversion from canonical CSR to COO is canonical. Using the `has_canonical_format` property on CSR can trigger a check of canonical if that property has not been accessed before. But this is a case when that is desirable IIUC.

This C++ code allocates the space for the output directly, whereas most other functions in this suite of tools have the allocation occur before calling and then "prune" the results after returning. I don't know enough about `std::vector` to know the impacts of that approach. And I see that the vectors get `resize`d twice. But we don't have to prune after the return. Can you say a few words about this choice to allocate inside the C++ function and to resize twice?  Are there advantages over preallocating and pruning via numpy?

The addition of `copy=copy` in COO `__init__` as well as the very minor tweaks to the 
dia_matmat code look good.

And I have a request below for you to add a comment block at the top of the new C++  `dia_tocsr` function.

Thanks!

MikhailRyazanov:
@dschult, thank you for reviewing so quickly! (This time I was busy, so replying only now.)

> This C++ code allocates the space for the output directly, whereas most other functions in this suite of tools have the allocation occur before calling and then "prune" the results after returning. I don't know enough about `std::vector` to know the impacts of that approach. And I see that the vectors get `resize`d twice. But we don't have to prune after the return. Can you say a few words about this choice to allocate inside the C++ function and to resize twice? Are there advantages over preallocating and pruning via numpy?

I'm not familiar with SciPy internals and strategies to be sure. Were can I read about that? And how the connection between NumPy arrays C++ vectors is done internally (the '*V' and '*W' codes in `_generate_sparsetools.py`)?

The general idea was to reduce memory allocations and copying.

`nnz` is not used in the beginning because, as I understand, DIA doesn't store its value but calculates it using `_getnnz()` with a Python loop, so doing the same thing in C++ is faster (and trivial to implement). Also, pruning with `scipy._lib._prune_array()` makes a copy, which C++ `vector::resize()` avoids (only change the internal size variable, since memory reallocation is most likely unnecessary in this case).

On the other hand, now I see that `_prune_array()` actually tolerates up to a factor-of-two memory overhead. From this perspective, it might be actually fine to just directly allocate DIA's `data.size` for the output arrays, since `nnz` never exceeds `data.size` and on average (for uniformly distributed diagonals) `nnz` ≈ `data.size`/2. Then all estimations and pruning can be completely removed. What do you think?

> And I have a request below for you to add a comment block at the top of the new C++ `dia_tocsr` function.

Definitely will do. And also will look at the rest slightly later.

P.S. By the way, there is a bug in `_getnnz()` for ill-constructed arrays:
```
>>> from scipy.sparse import dia_array
>>> d = dia_array(([[1, 2, 3]], [5]), shape=(3, 3))
>>> d.toarray()
array([[0, 0, 0],
       [0, 0, 0],
       [0, 0, 0]])
>>> d.nnz
-2
```
Since `sparse/tests` currently do have some tests with such weird inputs, I suspect that `_getnnz()` must return 0.
Should I just add a commit to correct that? Or open a separate issue/PR?

dschult:
Can you add a separate PR for the fix of DIA.nnz in ill-formed cases?
I think it will likely speed both sets of changes to have them separated -- it helps focus the discussions.

I am intrigued by the idea of reducing allocation and copying -- with `std::vector` or using numpy allocations. But I suspect the current sparse approach looked at that question back in the day. 

Perhaps @perimosocordiae has some perspective on tradeoffs of "pre-allocate via numpy && `_prune_array`" vs "allocate and resize `std:vector` in C++".

MikhailRyazanov:
@dschult,
> Can you add a separate PR for the fix of DIA.nnz in ill-formed cases?

Sure, see gh-23055.

Your review was in fact very helpful, because I've tried preallocating NumPy output arrays using the new `_getnzz()` and discovered that the overall conversion performance improves significantly. Apparently because `sparsetools.cxx` constructs `ndarray` from `std:vector` by always making a copy (not sure why this is necessary), while writing from C++ to existing `ndarray` is direct.

So I'll put this PR on hold until gh-23055 is merged, and then will refactor to accommodate those changes.

perimosocordiae:
@MikhailRyazanov is this PR still in draft mode, or are you ready for review?

scipy/sparse/_csr.py
```diff
@@ -60,6 +60,13 @@ def tocsr(self, copy=False):
 
     tocsr.__doc__ = _spbase.tocsr.__doc__
 
+    def tocoo(self, copy=False):
```
perimosocordiae, line 63: A comment here would be nice, explaining that CSR->COO preserves the sorted ordering property, while the CSC->COO conversion does not.

scipy/sparse/_dia.py
```diff
@@ -407,57 +407,29 @@ def diagonal(self, k=0):
...
+        # expansion for everything else even if not needed (see gh19245);
+        # at the same time, CSR wants common dtype for indices and indptr;
+        # so care should be taken to use appropriate indexing dtype throughout.
+        # Estimate smallest safe dtype for indptr.
+        idx_dtype = self._get_index_dtype(maxval=self.data.size)
+        order = np.argsort(self.offsets).astype(idx_dtype, copy=False)
+        indices, indptr, data = dia_tocsr(*self.shape, *self.data.shape,
+                                          self.offsets, self.data, order)
+        # If possible, shrink indexing dtype using actual nnz.
+        idx_dtype = self._get_index_dtype(maxval=indptr[-1])
```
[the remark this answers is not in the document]
perimosocordiae: +1, let's use self.shape here as well (in case we have indices on the minor axis that are really huge).

scipy/sparse/tests/test_base.py
```diff
@@ -5095,6 +5095,28 @@ def test_tocoo_tocsr_tocsc_gh19245(self):
...
         assert csc.indices.dtype == np.int32
 
+    def test_tocsr(self):
+        # test bound checks (other pathological cases are tested by
+        # TestConstructUtils::test_spdiags, and normal operation is ensured by
+        # many other tests here using .toarray())
+        d1 = [[1]]        # diagonal shorter than width
+        d3 = [[1, 2, 3]]  # diagonal longer than width
+
+        cases = [(d1, [-1], 1, [[0, 0], [1, 0]]),  # within
```
perimosocordiae: Let's convert each of these comments into a string, and include that in the assert message (i.e., as the third argument to `assert_array_equal`). This will make it a lot easier to figure out which case failed in the event of a test failure.

perimosocordiae (changes requested):

MikhailRyazanov:
> @MikhailRyazanov is this PR still in draft mode, or are you ready for review?

Still a draft. I need to refactor it using the gh-23055 changes (so the recent commit was only to get them in this branch) and the discussions above, then will reopen as "ready for review".

MikhailRyazanov:
@dschult, @perimosocordiae:
OK, here is the improved variant, with the output arrays allocated by NumPy, based on `nnz`, and then pruned. And hopefully also addressing all the review comments above.

Benchmark comparisons for various sizes (see [above](https://github.com/scipy/scipy/pull/23009#issuecomment-2889491978)); “`main`” is the current implementation from the main branch, “`this`” is this PR (other lines are explained below):
```
          sparse.Conversion.time_conversion
          --                                                 to_format
          ------------- ------------------------------------------------------------------------------------
           from_format        csr           csc          coo          lil           dok            bsr
          ============= ============== ============= ============ ============= ============= ==============
400 × 400:
  main:        dia       225±2μs        122±10μs      117±0.3μs    882±3μs       721±5μs       296±0.3μs
  this:        dia        71.5±0.09μs   122±0.5μs     123±0.3μs    661±1μs       544±6μs       131±0.1μs

  no nnz:      dia        51.9±2μs       97.5±0.3μs   106±0.8μs    635±3μs       511±1μs       109±0.5μs
  +copy:       dia        54.1±0.2μs    101±0.7μs     107±0.6μs    644±2μs       518±2μs       111±0.3μs

  cntnz:       dia        50.4±0.2μs     99.5±0.3μs   108±1μs      644±5μs       565±60μs      110±1μs

10000 × 10000:
  main:        dia       1.74±0.05ms    866±40μs      691±10μs     19.8±0.5ms    22.4±0.3ms    2.28±0.06ms
  this:        dia         175±1μs      392±4μs       290±7μs      17.2±0.08ms   18.8±0.1ms      692±8μs

  no nnz:      dia         157±0.5μs    366±3μs       266±2μs      16.9±0.08ms   18.6±0.06ms     662±3μs
  +copy:       dia         198±1μs      417±3μs       318±3μs      17.3±0.07ms   18.8±0.09ms     714±7μs

  cntnz:       dia         195±1μs      421±10μs      305±0.7μs    17.1±0.1ms    18.6±0.2ms      695±7μs

62500 × 62500:
  main:        dia       15.5±0.6ms     7.31±0.6ms    8.89±1ms     126±1ms       142±1ms       19.3±0.6ms
  this:        dia        1.06±0.02ms   5.94±0.07ms   3.05±0.6ms   113±0.5ms     126±0.3ms      6.80±0.06ms

  no nnz:      dia        1.07±0.04ms   5.76±0.08ms   3.66±0.9ms   113±1ms       127±0.5ms      6.81±0.2ms
  +copy:       dia        4.06±0.2ms    6.93±0.5ms    4.66±0.2ms   115±0.9ms     130±0.6ms      7.76±0.3ms

  cntnz:       dia        1.38±0.04ms   6.24±0.2ms    3.50±0.9ms   113±0.7ms     131±0.8ms      7.13±0.2ms
```

For small sizes, the performance is still better than the original but slightly worse than from the previous attempt with `nnz` calculated during the conversion (likely because `._getnnz()` adds a constant, size-independent, overhead), however, for larger sizes, the performance gain is now even better than before.

I've also tried two alternative approaches:
1. Using `max_nnz = self.data.size` instead of `max_nnz = self.nnz`, labeled “`no nnz`” in the benchmarks above. This improves the timings by ~20 μs, which is noticeable for small sizes, but at the same time is more likely to trigger copying when pruning the output, with a huge impact for large sizes (see the “`+copy`” lines above, which include artificially triggering it), plus higher memory usage (might be very undesirable when sizes are large).
2. Using `self.count_nonzero()` to get the exact output size before the conversion and thus avoid any memory overhead and reallocations. This would be useless with the current `count_nonzero()` implementation, which is slower than the conversion itself, but reimplementing it in C++ (see the [dia-cntnz](https://github.com/MikhailRyazanov/scipy/tree/dia-cntnz) branch in my fork) gives ~7-fold speed-up, and the overall results are shown as “`cntnz`” above. They are quite good for small sizes, but for large sizes are ~25% slower than “`this`”, although still ~10 faster than “`main`” and cannot degrade to the worst-case “`+copy`”. However, another C++ function (`dia_count_nonzero()`) additionally increases the library size by about as much as this PR does (with only added `dia_tocsr()`), so using this approach would be reasonable only if fast `count_nonzero()` for DIA arrays is useful on its own.

I don't think that the first alternative is really better, despite its slightly better best-case performance. The second alternative has some advantages and some disadvantages, so I'm not sure. What do you think?

scipy/sparse/tests/test_base.py
MikhailRyazanov: Done, please take a look.

scipy/sparse/_csr.py
MikhailRyazanov: Makes sense, please take a look.

scipy/sparse/_dia.py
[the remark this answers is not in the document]
MikhailRyazanov: Done.

scipy/sparse/sparsetools/dia.h
```diff
@@ -227,4 +227,54 @@ void dia_matvecs(const I A_rows,
 }
 
 
+template <class I, class T>
```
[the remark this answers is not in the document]
MikhailRyazanov: Sure! Forgot about it initially...

scipy/sparse/sparsetools/dia.h
```diff
@@ -227,4 +227,71 @@ void dia_matvecs(const I A_rows,
...
+    I nnz = 0;
+    // loop over rows
+    for (I i = 0; i < n_rows; ++i) {
+        // loop over offsets in ascending order
+        for (I k = 0; k < n_diags; ++k) {
+            const I n = order[k], // index of diagonal
+                    j = i + offsets[n]; // column
+            if (j < 0 or j >= j_end)
+                continue;
+            const T x = (data + npy_intp(L) * n)[j];
```
perimosocordiae: Let's use a different indexing expression here, to be more consistent with the rest of the usages in sparsetools:

```
data[L*n + j]
```

perimosocordiae (approved):

perimosocordiae:
Thanks @MikhailRyazanov! Based on your benchmark results, I think it's reasonable to keep the current approach (`max_nnz = self.nnz`) for now. We can then test using the more precise `count_nonzero()` version as a followup (if you're interested).

scipy/sparse/sparsetools/dia.h
MikhailRyazanov: This was basically to be consistent with the idiom
```C++
const T diag = diags + npy_intp(L) * n;
... diag[j] ...;
```
used throughout `dia.h`.
Do you also want to change other occurrences, like line 219 here:
https://github.com/scipy/scipy/blob/98e24c549749fee06bf2da11bb17b0bdf82da852/scipy/sparse/sparsetools/dia.h#L218-L224
to `const T a = A_data[npy_intp(A_L) * n + k];`?

perimosocordiae: Yeah, let's avoid mixing pointer arithmetic and indexing in the same expression.

MikhailRyazanov: Done!

dschult (approved):
This looks good to me!
Thanks @MikhailRyazanov !

PR #22024, opened 2024-12-08
MAINT: version pins/prep for 1.15.0rc1

tylerjereddy:
Important: CI is skipped completely for now because this will not work--it mostly demonstrates what I'd *like* to be able to do, but that apparently cannot be done because of the limited hooks available in https://packaging.python.org/en/latest/specifications/dependency-specifiers/#dependency-specifiers. I'm pretty sure our only option at the moment for free-threaded CPython `3.13` wheels will be to treat `pyproject.toml` more or less like free-threaded CPython doesn't exist, and disable build isolation for `cibuildwheel` (which is basically what we do on `main`). 

However, if you wanted to do a more formally-correct build-isolation-enabled CPython `3.13` free-threaded wheel build it looks like you'd be in trouble at the moment?

* Update dependency version ranges for build/runtime as we prepare for the SciPy `1.15.0rc1` pre-release, paying special attention to the complications related to free-threaded CPython `3.13` support.

* In addition to the differences caused by free-threading support, there are also some version handling changes that mean this update differs from the previous one performed in the last release cycle at gh-20831.

[skip ci] [ci skip] [skip cirrus] [skip circle]

TODO:

- [ ] regular CI passing
- [ ] wheels CI passing

PR #23013, opened 2025-05-19
MAINT: version pins/prep for 1.16.0rc1

tylerjereddy:
* Adjust the build and runtime dependency version upper bounds (on maintenance branch) as we prepare for the release of SciPy `1.16.0rc1`. Many of the details are based on the code changes and discussion at gh-22024. In particular, see that PR for some discussion on free-threading support. This assumes that we will offer a release that is free of NumPy `2.3.0` deprecation warnings, which seems reasonable based on current release cycles.

Getting this just right can be tricky, thanks for any review suggestions. I'll prepare the `.mailmap` updates under separate cover today, and currently intend to try to keep to schedule for RC1 on May 21.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,10 +17,19 @@
 [build-system]
 build-backend = 'mesonpy'
 requires = [
-    "meson-python>=0.15.0",
-    "Cython>=3.0.8",        # when updating version, also update check in meson.build
-    "pybind11>=2.13.2",     # when updating version, also update check in scipy/meson.build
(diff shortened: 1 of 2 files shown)


rgommers (approved):
Thanks Tyler, LGTM!

> This assumes that we will offer a release that is free of NumPy 2.3.0 deprecation warnings, which seems reasonable based on current release cycles.

Agreed. 2.3.x will be branched well before the 1.16.0 release.

PR #23014, opened 2025-05-19
DOC: .mailmap updates for 1.16.0rc1

tylerjereddy:
* Update the `.mailmap` for SciPy `1.16.0rc1` release. This is generally done by checking GitHub user profiles for names/academic web pages (when the currently-listed name looks like a handle rather than proper name), and occasionally by searching online for names associated with a commit-provided email address.

[docs only]

In general, I've assumed that folks would prefer their proper given and surname(s) for acknowledgement. This is just a convention we've been following, and there are sometimes folks who prefer to remain a bit more anonymous (let me know if you want your name reverted). I took note of a few GitHub usernames where I was less confident about my ability to assign a proper name given the available information--these folks are free to let me know their preferred acknowledgement name and I'll update accordingly (otherwise, I'll leave them as is):

- [ ] @aiudirog
- [ ] @Backfisch263
- [ ] @lciti
- [ ] @pmav99
- [ ] @ciaokitty
- [ ] @ShannonS00
- [ ] @sildater
- [ ] @gsreeja11
- [ ] @Tearyt
- [ ] @vfdev
- [ ] @vpecanins
- [ ] @windows-server-2003

Doc build was still passing locally; update release note author list still looks "ok," and makes sense that it has a few less entries after catching duplicates with updates here.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -182,8 +182,10 @@ Eric Larson <larson.eric.d@gmail.com> Eric89GXL <larson.eric.d@gmail.com>
 Eric Quintero <eric.antonio.quintero@gmail.com> e-q <eric.antonio.quintero@gmail.com>
 Eric Quintero <eric.antonio.quintero@gmail.com> Eric Quintero <e-q@users.noreply.github.com>
 Eric Soroos <eric-github@soroos.net> wiredfool <eric-github@soroos.net>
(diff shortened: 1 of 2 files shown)


tylerjereddy:
Alright, self-merging so I can proceed with the (pre-)release process. If folks pinged here follow up later we can consider updating the release notes with their desired names.

tylerjereddy:
`1.16.0rc1` wheel builds initiated: https://github.com/scipy/scipy/actions/runs/15153009479

fingers crossed..

PR #23015, opened 2025-05-19
DOC: scipy.signal.ShortTimeFFT.k_max: Fix slice index

danra:
As depicted in https://docs.scipy.org/doc/scipy/tutorial/signal.html#sliding-windows, `k_max - 1` is the largest sample index of the slice `p_max - 1`, not `p_max`.

[docs only]


DietBru:
Good catch, thank you! I took the liberty to move the closing accent to have internal link for `p_max` int the HTML docs.
Before:
image

After:
image

DietBru (approved):

DietBru:
Thanks @danra for your PR. We are looking forward to receiving more input from you.

PR #23020, opened 2025-05-20
ENH: interpolate.BSpline: add array API standard interface

ev-br:
This PR adds a minimal Array API interface to `BSpline` and `make_interp_spline`.

There are several minor details worth spelling out. First, `BSpline` uses a C extension, therefore non-numpy array inputs need to go through the now-standard `np.asarray(input); result = c_function(input); return xp.asarray(result)` dance.

Second, `BSpline` instances used to store numpy arrays. Namely, for array-like `t` and `c`, and  `spl = BSpline(t, c, k)`, then `spl.t` and `spl.c` are numpy array attributes derived from the `__init__` arguments.
This PR changes these attributes into properties, so that if `t` and `c` are torch tensors, and `spl = BSpline(t, c, k)`, 
a new private attribute `spl._t` is a numpy array, and `spl.t` is a propery which constructs a torch tensor from `spl._t`.

The advantage of this is that at the call time, we can use `spl._t` without paying the price of `np.asarray` (which is not negligible here, actually). An unpleasant consequence is that the property abstraction leaks for mutable properties: `spl.t.fill(0)` goes through the getter not the setter. There's nothing to be done about it though, and this does not lead to new failure modes AFAICS.

Third, if `t`, `c` and `x` are torch tensors, then `spl = BSpline(t, c, k); spl(x)` should produce a torch tensor, too. To achieve this, we store the namespace on an instance, via `self.xp = array_namespace(t, c)` and the `__call__` is

```
def __call__(self, x):
     x_np = np.asarray(x)
     out = heavy_lifting_in_c(x_np)
     return self.xp.asarray(out)
```

At the moment, the new attribute `xp` is public, since I don't see a particularly strong need to hide it. If anything, it can be useful to a user: otherwise there's no easy way for a user to tell what is the type of `spl(1)` without calling it.
If this turns out to be problematic, we can always hide the attribute behind a property. Or store `xp.asarray` as a private method. 

One last small thing: since module objects are not picklable, we need to add a custom pickler to correctly handle the `xp` attribute.

### DIFF
--- a/scipy/_lib/_array_api.py
+++ b/scipy/_lib/_array_api.py
@@ -38,6 +38,8 @@ from scipy._lib._array_api_override import (
     array_namespace, SCIPY_ARRAY_API, SCIPY_DEVICE
 )
 from scipy._lib._docscrape import FunctionDoc
+from scipy._lib import array_api_extra as xpx
+
 
 __all__ = [
     '_asarray', 'array_namespace', 'assert_almost_equal', 'assert_array_almost_equal',
@@ -549,6 +551,15 @@ def xp_result_device(*args):
         if is_array_api_obj(arg):
             return xp_device(arg)
     return None
+
+
+# np.r_ replacement
+def concat_1d(xp, *arys):
+    """A replacement for np.r_: `xp.concat` does not accept python scalars or 0D arrays.
+    """
+    arys = [xp.asarray(a) for a in arys]
+    arys = [xpx.atleast_nd(a, ndim=1, xp=xp) for a in arys]
+    return xp.concat(arys)
 
 
 def is_marray(xp):
--- a/scipy/_lib/_array_api.py
+++ b/scipy/_lib/_array_api.py
@@ -16,6 +16,7 @@ from contextlib import contextmanager
 from contextvars import ContextVar
 from types import ModuleType
 from typing import Any, Literal, TypeAlias
+from collections.abc import Iterable
 
 import numpy as np
 import numpy.typing as npt
@@ -554,11 +555,11 @@ def xp_result_device(*args):
 
 
 # np.r_ replacement
-def concat_1d(xp, *arys):
-    """A replacement for np.r_: `xp.concat` does not accept python scalars or 0D arrays.
+def concat_1d(xp: ModuleType | None, *arrays: Iterable[ArrayLike]) -> Array:
+    """A replacement for `np.r_` as `xp.concat` does not accept python scalars
+       or 0-D arrays.
     """
-    arys = [xp.asarray(a) for a in arys]
-    arys = [xpx.atleast_nd(a, ndim=1, xp=xp) for a in arys]
+    arys = [xpx.atleast_nd(xp.asarray(a), ndim=1, xp=xp) for a in arrays]
     return xp.concat(arys)
 
 
--- a/scipy/interpolate/_bsplines.py
+++ b/scipy/interpolate/_bsplines.py
@@ -280,6 +280,8 @@ class BSpline:
         """
         return self.t, self.c, self.k
 
+    # Under the hood, self._c and self._t are always saved as numpy array
+    # because they are used in a C extension expecting numpy arrays.
     @property
     def t(self):
         return self._asarray(self._t)
--- a/scipy/interpolate/_bsplines.py
+++ b/scipy/interpolate/_bsplines.py
@@ -214,7 +214,7 @@ class BSpline:
     def __init__(self, t, c, k, extrapolate=True, axis=0):
         super().__init__()
 
-        self.xp = array_namespace(c, t)
+        self._xp = array_namespace(c, t)
 
         self.k = operator.index(k)
         self._c = np.asarray(c)
@@ -271,7 +271,7 @@ class BSpline:
         self._t, self._c, self.k = np.asarray(t), np.asarray(c), k
         self.extrapolate = extrapolate
         self.axis = axis
-        self.xp = array_namespace(t, c)
+        self._xp = array_namespace(t, c)
         return self
 
     @property
@@ -282,7 +282,7 @@ class BSpline:
 
     @property
     def t(self):
-        return self.xp.asarray(self._t)
+        return self._xp.asarray(self._t)
 
     @t.setter
     def t(self, t):
@@ -290,19 +290,19 @@ class BSpline:
 
     @property
     def c(self):
-        return self.xp.asarray(self._c)
+        return self._xp.asarray(self._c)
 
     @c.setter
     def c(self, c):
         self._c = np.asarray(c)
 
     def __getstate__(self):
-        return self._t, self._c, self.k, self.axis, self.extrapolate, self.xp.empty(1)
+        return self._t, self._c, self.k, self.axis, self.extrapolate, self._xp.empty(1)
 
     def __setstate__(self, state):
-        # {get,set}state needed because storing a module as self.xp breaks pickling
+        # {get,set}state needed because storing a module as self._xp breaks pickling
         self._t, self._c, self.k, self.axis, self.extrapolate, token = state
-        self.xp = array_namespace(token)
+        self._xp = array_namespace(token)
 
     @classmethod
     def basis_element(cls, t, extrapolate=True):
@@ -571,7 +571,7 @@ class BSpline:
             l = list(range(out.ndim))
             l = l[x_ndim:x_ndim+self.axis] + l[:x_ndim] + l[x_ndim+self.axis:]
             out = out.transpose(l)
-        return self.xp.asarray(out)
+        return self._xp.asarray(out)
 
     def _ensure_c_contiguous(self):
         """
@@ -603,7 +603,7 @@ class BSpline:
         splder, splantider
 
         """
-        xp = self.xp
+        xp = self._xp
         c = xp.asarray(self.c, copy=True)
         t = self.t
         npr = _fitpack_impl.npr   # np.r_ replacement
@@ -641,7 +641,7 @@ class BSpline:
         splder, splantider
 
         """
-        xp = self.xp
+        xp = self._xp
         c = xp.asarray(self.c, copy=True)
         t = self.t
         npr = _fitpack_impl.npr   # np.r_ replacement
@@ -730,7 +730,7 @@ class BSpline:
                 # Fast path: use FITPACK's routine
                 # (cf _fitpack_impl.splint).
                 integral = _fitpack_impl.splint(a, b, (self._t, self._c, self.k))
-                return self.xp.asarray(integral * sign)
+                return self._xp.asarray(integral * sign)
 
         # Compute the antiderivative.
         c = self._c
@@ -788,7 +788,7 @@ class BSpline:
             integral = out[1] - out[0]
 
         integral *= sign
-        return self.xp.asarray(integral.reshape(ca.shape[1:]))
+        return self._xp.asarray(integral.reshape(ca.shape[1:]))
 
     @classmethod
     def from_power_basis(cls, pp, bc_type='not-a-knot'):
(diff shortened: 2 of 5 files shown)


scipy/interpolate/_fitpack_impl.py
```diff
@@ -745,13 +748,23 @@ def insert(x, tck, m=1, per=0):
         return (tt, cc, k)
 
 
-def splder(tck, n=1):
+# np.r_ replacement
+def npr(xp, *arys):
```
rgommers, line 752: A readable name here please! I suggest `concat_1d`.

rgommers: With a code comment or docstring explaining that this exists because `xp.concat` doesn't take scalars or 0-D arrays.

rgommers (commented):
Overall LGTM, thanks Evgeni.

> > At the moment, the new attribute `xp` is public, since I don't see a particularly strong need to hide it.

Private please! There is no need to leak this, even after array API support becomes fully public.

> If anything, it can be useful to a user: otherwise there's no easy way for a user to tell what is the type of `spl(1)` without calling it. 

If this info is needed - which should be rare - `isinstance` on one of the array attributes is still quite cheap, and more idiomatic.

> If this turns out to be problematic, we can always hide the attribute behind a property. Or store `xp.asarray` as a private method.

Renaming it from `.xp` to `._xp` seems most straightforward here.

scipy/interpolate/_fitpack_impl.py
lucascolley: I think we should also consider putting helpers like this in `_lib._array_api`. I haven't blocked on this when merging previous PRs, but it does seem a bit messy to have such general functions living in specific submodules. It also would make it easier to see at a glance the helpers which we have written.

ev-br:
Comments addressed. 

Moving `concat_1d` into `_lib` runs into some sort of  import cycle between array-api-compat and array-api-extra. Using a local import fixes the immediate problem, but I suppose it is brittle either way. The multivendoring contraption there is frankly rather crazy IMO. Bringing it back to sanity is out of scope for this PR, so if we'd rather avoid local imports, I can move concat_1d back to scipy.interpolate.

lucascolley:
Good point. I think the simplest fix may be to move `_array_api.array_namespace` to a different file which doesn't import from `_array_api`.

ev-br:
>One last small thing: since module objects are not picklable, we need to add a custom pickler to correctly handle the xp attribute.

Actually, we don't if we store a private function instead of a private namespace: instead of `self._xp.asarray` we store `self._asarray = array_namespace(t, c).asarray` and all is well. The need to call other namespaced functions should be rare, and here it only happens in `derivative` and `antiderivative`, which are essentially constructors. So no performance considerations here, and the code is simpler, I'd say it's worth it.
The last commit makes this change.

scipy/_lib/_array_api.py
```diff
@@ -651,6 +653,15 @@ def xp_result_device(*args):
     return None
 
 
+# np.r_ replacement
+def concat_1d(xp, *arys):
+    """A replacement for np.r_: `xp.concat` does not accept python scalars or 0D arrays.
+    """
+    arys = [xp.asarray(a) for a in arys]
+    arys = [xpx.atleast_nd(a, ndim=1, xp=xp) for a in arys]
```
j-bowhay: Do we need to run the loop twice?
```suggestion
    arys = (xpx.atleast_nd(xp.asarray(a), ndim=1, xp=xp) for a in arys)
```

scipy/interpolate/_bsplines.py
```diff
@@ -328,11 +349,15 @@ def basis_element(cls, t, extrapolate=True):
...
-        k = len(t) - 2
+        xp = array_namespace(t)
+        t = np.asarray(t)
+        k = t.shape[0] - 2
         t = _as_float_array(t)
         t = np.r_[(t[0]-1,) * k, t, (t[-1]+1,) * k]
         c = np.zeros_like(t)
         c[k] = 1.
+
+        t, c = xp.asarray(t), xp.asarray(c)
```
j-bowhay, line 365: It slightly bugs me that we have to call `xp.asarray` on these for the sole purpose of being able to get the array namespace later but I guess the cost of doing so is pretty low

scipy/interpolate/_bsplines.py
```diff
@@ -568,12 +593,15 @@ def derivative(self, nu=1):
         splder, splantider
 
         """
-        c = self.c.copy()
+        c = self._asarray(self.c, copy=True)
+        t = self.t
+        xp = array_namespace(t, c)
```
j-bowhay, line 603: Can we not store `xp` when the object is initialized?

j-bowhay (commented):
Thanks, not looked at the tests yet

scipy/interpolate/_bsplines.py
lucascolley: cf. https://github.com/scipy/scipy/blob/8f51c63d384f0e9c966355362f982ea7ec2de1cc/scipy/spatial/transform/_rigid_transform.py#L369

ev-br: That's deliberate, actually:

https://github.com/scipy/scipy/pull/23020#issuecomment-2897535533

scipy/interpolate/_bsplines.py
ev-br: Bugs me too, a bit. This method is not meant to be in a tight loop though, and the parts that are, have numpy arrays to work on.

scipy/_lib/_array_api.py
```diff
@@ -651,6 +653,15 @@ def xp_result_device(*args):
     return None
 
 
+# np.r_ replacement
+def concat_1d(xp, *arys):
+    """A replacement for np.r_: `xp.concat` does not accept python scalars or 0D arrays.
```
lorentzenchr:
```suggestion
    """A replacement for `np.r_` as `xp.concat` does not accept python scalars or 0-D arrays.
```

scipy/_lib/_array_api.py
```diff
@@ -651,6 +653,15 @@ def xp_result_device(*args):
     return None
 
 
+# np.r_ replacement
+def concat_1d(xp, *arys):
```
lorentzenchr:
```suggestion
def concat_1d(xp: ModuleType | None, *arrays: Iterable[ArrayLike]) -> Array:
```
In line with other functions in this file.

scipy/interpolate/_bsplines.py
```diff
@@ -271,6 +276,22 @@ def tck(self):
         """
         return self.t, self.c, self.k
 
+    @property
```
lorentzenchr, line 285: A comment here or above in the init would help, something like
```suggestion
    # Under the hood, self._c and self._t are always saved as numpy array
    # because they are used in a C extension expecting numpy arrays.
    @property
```

lorentzenchr (approved):
What happens if c and t are different arrays like `BSpline(t=torch_array, c=numpy_array, 2)`? (Not that it makes sense, but in case a user does it.)

lucascolley:
> What happens if c and t are different arrays like BSpline(t=torch_array, c=numpy_array, 2)? (Not that it makes sense, but in case a user does it.)

`raise TypeError(f"Multiple namespaces for array inputs: {namespaces}")`

https://github.com/data-apis/array-api-compat/blob/6c708d13e826fb850161babf34f306fe80cae875/array_api_compat/common/_helpers.py#L665

ev-br:
> What happens if c and t are different arrays like BSpline(t=torch_array, c=numpy_array, 2)? (Not that it makes sense, but in case a user does it.)

What Lucas said: mixing array types in the constructor is an automatic TypeError from `array_namespace`. What is somewhat less clear-cut is mixing array types between `__init__` and `__call__` (in the scikit-learn language: `.fit` and `.predict`):  what is `BSpline(t=np_array, c=np_array, k)(torch_tensor)`?

In this PR --- and I think this is correct in general --- the logic is that the array namespace is fixed by the arguments of the constructor; at call time, the arguments are cast to the array type _of that fixed namespace_:

```
In [1]: import torch

In [2]: import numpy as np

In [3]: from scipy.interpolate import BSpline, make_interp_spline

In [4]: spl = make_interp_spline(np.arange(10), np.arange(10)**2)

In [5]: spl_ = BSpline(torch.asarray(spl.t), torch.asarray(spl.c), spl.k)

In [7]: spl_(1)
Out[7]: tensor(1., dtype=torch.float64)

In [8]: spl_(np.asarray(1))
Out[8]: tensor(1., dtype=torch.float64)
```

scipy/_lib/_array_api.py
ev-br: Included, thanks.

ev-br:
Addressed all review comments.

lorentzenchr:
> In this PR --- and I think this is correct in general --- the logic is that the array namespace is fixed by the arguments of the constructor; at call time, the arguments are cast to the array type of that fixed namespace

Maybe it does not matter much and consistency within scipy is more import, but as a user I would expect to get the same array back as I pass into `__call__`.

scipy/interpolate/_bsplines.py
```diff
@@ -331,11 +354,15 @@ def basis_element(cls, t, extrapolate=True):
         >>> plt.show()
 
         """
-        k = len(t) - 2
+        xp = array_namespace(t)
+        t = np.asarray(t)
+        k = t.shape[0] - 2
         t = _as_float_array(t)
```
lucascolley, line 360: maybe consider renaming `_as_float_array`? It was not clear to me that it was something NumPy-specific. perhaps include 'c_contiguous'?

---

What is the precise reason we need to convert `t` to and from `np` in this method? Can the `np.r_` call not be replaced with `concat_1d`?

If we only need to worry about C-contiguity for NumPy arrays, could we not make `_as_float_array` operate on `xp` arrays by using `_array_api._asarray(..., order='C')`?

scipy/interpolate/tests/test_bsplines.py
```diff
@@ -838,7 +892,6 @@ def check_splev(b, j, der=0, atol=1e-14, rtol=1e-14):
                 check_splev(b1, j, der, 1e-12, 1e-12)
 
 
-### stolen from @pv, verbatim
```
lucascolley, line 841: 😄

lucascolley (commented):
thanks Evgeni, very close!

scipy/interpolate/_bsplines.py
ev-br: No opinion on renaming `_as_float_array`.  Send a patch if you have feel strongly about it?

----- 

Conversions to/from numpy: there are of course multiple ways to skin this one. It's used ten times, 
```
$ git grep -n "_as_float_array" scipy/interpolate/ | wc -l
11
 ```
and IIRC other uses are after the conversion to numpy. So it felt easier to just do it like it's done here and be done with it. Care to send a patch if you see a way to make it nicer across the module?

lucascolley: Just to clarify, the only reason we are moving to `np` here is to use `_as_float_array`? If so, that seems bad.

ev-br: ATM all alternative constructors, including `basis_element`, use this pattern. If a clean-up is desired, we can do it in a sequel, after https://github.com/scipy/scipy/pull/23052 which converts other users of `_as_float_array`.

lucascolley: alright, if you add a TODO comment here to avoid the namespace conversion I am happy to merge!

Alternatively, I could try to come up with a patch in the coming days.

lucascolley (approved):
thanks Evgeni, and for the review @lorentzenchr !

scipy/interpolate/_bsplines.py
ev-br: A patch would be great indeed @lucascolley! Meanwhile I'll then add a TODO comment to gh-23052 to avoid flushing the CI here.

ev-br:
Merging as approved. Thanks to all reviewers!

Issue #22980, opened 2025-05-13
DEV: `spin test -b all` adds `--pyargs scipy` twice


ev-br:
Running 

```
$ spin test  -v -b all
```

creates a process which lists `--pyargs scipy` twice:

```
$ ps  2854732
    PID TTY      STAT   TIME COMMAND
2854732 pts/0    Rl+   55:20 /home/ev-br/.conda/envs/scipy-dev2/bin/python3.13 -P -m pytest -v -m not slow --pyargs scipy --pyargs scipy
```

Looking at the `spin` output shows a couple of other oddities:

```
$ spin test  -v -b array-api-strict
/home/ev-br/.conda/envs/scipy-dev2/lib/python3.13/site-packages/click/core.py:1193: UserWarning: The parameter -j is used more than once. Remove its duplicate as parameters should be unique.
  parser = self.make_parser(ctx)
/home/ev-br/.conda/envs/scipy-dev2/lib/python3.13/site-packages/click/core.py:1186: UserWarning: The parameter -j is used more than once. Remove its duplicate as parameters should be unique.
  self.parse_args(ctx, args)
Invoking `build` prior to running tests:
$ meson compile -j 32 -C build
INFO: autodetecting backend as ninja
INFO: calculating backend command to run: /home/ev-br/.conda/envs/scipy-dev2/bin/ninja -C /home/ev-br/repos/scipy/build -j 32
ninja: Entering directory `/home/ev-br/repos/scipy/build'
[2/338] Generating subprojects/highs/src/HConfig.h with a custom command
$ meson install --only-changed -C build --destdir ../build-install --tags=runtime,python-runtime,tests,devel
$ export PYTHONPATH="/home/ev-br/repos/scipy/build-install/usr/lib/python3.13/site-packages"
$ /home/ev-br/.conda/envs/scipy-dev2/bin/python3.13 -P -c 'import scipy'
$ cd /home/ev-br/repos/scipy/build-install/usr/lib/python3.13/site-packages
$ /home/ev-br/.conda/envs/scipy-dev2/bin/python3.13 -P -m pytest -v -m 'not slow' --pyargs scipy --pyargs scipy
^C
```

1. What is the "-j is used more than once" warning?
2. It's great that `spin` lists the CLI command, and I expected it to be the exact analog of what it runs. However, it does not show anything about the backends selection, `-b` arguments. It's be nice to have it output how exactly it sets the backend(s).

rgommers:
[...]

PR #23025, opened 2025-05-21
DEV: spin: remove duplicate argument `-j` from parent commands

czgdp1807:
#### Reference issue

https://github.com/scipy/scipy/issues/22980

This PR only deals with duplication of `-j` in parent command and in the extended command in `.cmds/spin.py`. This went unnoticed in the initial addition of `spin` because only after recent update of `click`, it has started raising `UserWarning` on duplication of arguments.

This `UserWarning` was being raised in `build`, `test`, and `docs` on my macOS and Linux machines. The problem is OS independent.

For fixing this problem, I have used `remove_args` functionality of `@spin.util.extend_command`. I have removed those arguments from parent command to keep SciPy's build interface backwards compatible (SciPy uses `--parallel` everywhere, `spin` use `--jobs`). However, `spin` still handles the job argument, we in SciPy just take the value from `parallel` and pass it on to `spin` via `parent_callback`. This fix helps in keeping APIs backwards compatible with minimal change.

#### Additional information

I still see warnings related to duplication of `-C` when I run `spin build`. The reason is we use `-C` for short hand of `--setup-args` and `spin` uses `-C` for short hand of `--build-dir`.  Check the links below. Removing this duplication would be tricky, a separate topic of discussion though.

```zsh
(scipy-dev) 14:34:19:~/Quansight/scipy % spin build
/Users/czgdp1807/mambaforge/envs/scipy-dev/lib/python3.13/site-packages/click/core.py:1193: UserWarning: The parameter -C is used more than once. Remove its duplicate as parameters should be unique.
  parser = self.make_parser(ctx)
/Users/czgdp1807/mambaforge/envs/scipy-dev/lib/python3.13/site-packages/click/core.py:1186: UserWarning: The parameter -C is used more than once. Remove its duplicate as parameters should be unique.
  self.parse_args(ctx, args)
$ meson compile -j 10 -C build
.
.
.
```

https://github.com/scipy/scipy/blob/2ebb23877c780f69db32ae856fa50172e5caff15/.spin/cmds.py#L40-L43

https://github.com/scientific-python/spin/blob/fee6abe42cea118c2e4577cce5e0ca034c55748e/spin/cmds/meson.py#L275-L283

https://github.com/scientific-python/spin/blob/fee6abe42cea118c2e4577cce5e0ca034c55748e/spin/cmds/meson.py#L311-L312

cc: @rgommers @ev-br @tylerjereddy

### DIFF
--- a/.spin/cmds.py
+++ b/.spin/cmds.py
@@ -179,7 +179,7 @@ def build(*, parent_callback, meson_args, verbose, werror, asan, debug,
         "'jax.numpy', 'dask.array')."
     )
 )
-@spin.util.extend_command(spin.cmds.meson.test)
+@spin.util.extend_command(spin.cmds.meson.test, remove_args=("n_jobs",))
 def test(*, parent_callback, pytest_args, tests, coverage,
          durations, submodule, mode, parallel,
          array_api_backend, **kwargs):
@@ -263,10 +263,6 @@ def test(*, parent_callback, pytest_args, tests, coverage,
         if markexpr != "full":
             pytest_args = ('-m', markexpr) + pytest_args
 
-    n_jobs = parallel
-    if (n_jobs != 1) and ('-n' not in pytest_args):
-        pytest_args = ('-n', str(n_jobs)) + pytest_args
-
     if durations:
         pytest_args += ('--durations', durations)
 
@@ -274,7 +270,7 @@ def test(*, parent_callback, pytest_args, tests, coverage,
         os.environ['SCIPY_ARRAY_API'] = json.dumps(list(array_api_backend))
 
     parent_callback(**{"pytest_args": pytest_args, "tests": tests,
-                    "coverage": coverage, **kwargs})
+                    "coverage": coverage, "n_jobs": parallel, **kwargs})
 
 @click.option(
         '--list-targets', '-t', default=False, is_flag=True,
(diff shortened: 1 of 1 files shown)


rgommers:
> For fixing this problem, I have used `remove_args` functionality of `@spin.util.extend_command`. I have removed those arguments from parent command to keep SciPy's build interface backwards compatible (SciPy uses `--parallel` everywhere, `spin` use `--jobs`).

Would the resulting code be simpler if we went with `--jobs`? Is it just the naming that's different, or anything else?

rgommers:
CI is unhappy:
```
__main__.py: error: unrecognized arguments: -n scipy
```

czgdp1807:
> Is it just the naming that's different, or anything else?

Yes only the naming is different. Another way out is to just remove `--parallel` from SciPy and rely on `spin` for multiple jobs during build/test.

rgommers:
> Yes only the naming is different. Another way out is to just remove `--parallel` from SciPy and rely on `spin` for multiple jobs during build/test.

I think that's the better approach. We don't have to be 100% compatible here. Most people either use the defaults or the short form `-j 10`) anyway. Can you try that in a separate PR or branch so we can compare?

czgdp1807:
Sure. Let me create a separate PR. All CI tests should pass because there isn't any usage of `--parallel` in `.yml` files within SciPy codebase.

PR #23029, opened 2025-05-21
DOC: add documentation for the new `use-system-libraries` build option

rgommers:
Adds documentation for the build option added recently in gh-22893. Also xref gh-17751.


lucascolley (approved):
sounds accurate!

PR #23039, opened 2025-05-22
TYP: fix and improve the ``linalg._decomp_lu_cython`` stub

jorenham:
Stubtest was complaining in scipy/scipy-stubs that the stubbed `lu_decompose` function did not exist (it's a `cdef`), so I removed it. I also fixed two incorrect parameter names, which matters if called with keyword arguments.

I also took the liberty of tightening up the input annotations. This way, static type-checkers will reject (combinations of) array dtypes that would otherwise cause runtime exceptions.


jorenham:
For what it's worth; in scipy/scipy-stubs#535 I've changed this stub in the same way. And stubtest seems to be happy with it: scipy/scipy-stubs#535

lucascolley (approved):
thanks Joren!

PR #23041, opened 2025-05-22
DOC/DEV: Updated developer docs for `spin`

czgdp1807:

I have updated the commands everywhere in the docs to now use `spin`. Removed one additional line related to `dev.py`.

I have done this change in multiple commits. I would suggest to review this PR commit by commit because I think it would be better to do all the documentation changes in a single PR. So to make review process easier, I would suggest to check each commit's changes.

#### Additional information

Note that we don't have `spin notes` yet. `python dev.py` has it.

cc: @rgommers @ev-br


doc/source/dev/core-dev/releasing.rst.inc
```diff
@@ -148,7 +148,7 @@ Here is a complete list of artifacts created for a release:
 
 An ``sdist`` is generated by running ``python -m build --sdist`` (note: we
 still need to move this into a CI job!), and the Changelog and README are built
-by running ``python dev.py notes`` (with tags, see ``python dev.py notes
+by running ``spin notes`` (with tags, see ``spin notes
```
rgommers, line 151: `spin notes` doesn't work

rgommers (commented):
Thanks @czgdp1807. This looks pretty good. I noticed that `spin notes` and `spin `authors` commands seem to be missing compared to `dev.py`. I can't remember if we left it out or delayed implementing it on purpose - but I don't think so. Unless I'm wrong about that, can you implement those in a separate PR?

lucascolley (commented):
seems good to me once Ralf's comment is addressed!

I think "`.github/workflows/pixi.toml` can be changed to use spin" is the only remaining follow-up action that I had noted

rgommers:
> I think "`.github/workflows/pixi.toml` can be changed to use spin" is the only remaining follow-up action that I had noted

That's perhaps best done as part of the PR that deletes `dev.py` probably - already captured in gh-22887.

doc/source/dev/contributor/rendering_documentation.rst
```diff
@@ -62,7 +62,7 @@ with Sphinx`_ \ *.*
 To render the documentation on your own machine:
 
 0. Ensure that you have a working SciPy build (see :ref:`building-from-source`).
-#. Then run ``python dev.py doc`` to build the documentation.
+#. Then run ``spin doc`` to build the documentation.
```
melissawm, line 65:
```suggestion
#. Then run ``spin docs`` to build the documentation.
```

doc/source/dev/contributor/rendering_documentation.rst
```diff
@@ -74,7 +74,7 @@ To render the documentation on your own machine:
    - Changes to certain documents do not take effect when Sphinx documentation
      is rebuilt. In this case, you can build from scratch by deleting the
      directories ``scipy/doc/build`` and ``source/reference/generated``, or by
-     running ``python dev.py doc clean`` then building the docs again.
+     running ``spin doc clean`` then building the docs again.
```
melissawm, line 77:
```suggestion
     running ``spin docs clean`` then building the docs again.
```

This is actually currently not working, it just rebuilds the docs again and does not clean anything. It is working on NumPy though, so it may be worth looking into how it's done there.

melissawm (changes requested):
Left some comments on the only command I've actually ran so far which is docs. Cheers!

doc/source/dev/core-dev/releasing.rst.inc
rgommers: this works now, resolving.

doc/source/dev/contributor/rendering_documentation.rst
rgommers: There are multiple issues with the `docs` command, in SciPy and also in NumPy:

- For both: docstring looks bad, needs improving:
    - `SPHINXOPTS` should be removed since it's both too niche and incorrect,
    -  It should explain what targets are supported and how to build exactly those - `html`, `dist` only for SciPy, and in addition `latex` for NumPy (better renamed to `pdf` probably)
    - `spin docs targets` should not be mentioned, that's just confusing for no good reason.
- For both: `sphinx-gallery` is not used and hence should not be an option, that is confusing
- For SciPy: docstring is displayed twice (see below)
- For SciPy: `--clean` doesn't work indeed, deferring to `spin` should work
- For SciPy: `--no-cache` doesn't seem useful, it says it forces a full rebuild, so seems duplicate with `--clean` at that point.
- For SciPy: `--list-targets` can then also be removed

image

I'll link to this from gh-22887

czgdp1807: Sure. So seems like this PR doesn't anything more from me. :-)

rgommers (approved):
LGTM, in it goes. Thanks @czgdp1807

PR #23043, opened 2025-05-23
DEP: linalg: remove kron

j-bowhay:
`kron` is due to be removed in this release.

Also in the second commit tidies up some outdated warning filters etc


lucascolley (approved):
LGTM, thanks Jake!

PR #21754, opened 2024-10-25
MAINT: interpolate: rename `interpnd` -> `_interpnd`

ev-br:
#### Reference issue

closes https://github.com/scipy/scipy/issues/19907

#### What does this implement/fix?

Add  a leading undescore to the `interpolate/_interpnd.so` extension module.

#### Additional information

This is the last non-underscored private module in scipy/interpolate (dfitpack was done in https://github.com/scipy/scipy/pull/20612)

PR #23045, opened 2025-05-23
DEP: interpolate.interpnd: remove incidental imports from private but present module

j-bowhay:
follow up to https://github.com/scipy/scipy/pull/21754, removing these accidentally exposed functions is due for this release


lucascolley (approved):
thanks Jake!

Issue #23046, opened 2025-05-23
ENH: `vectorized_filter`: special case of scalar `size` with `axes`


ego-thales:
### Is your feature request related to a problem? Please describe.

Hello,

I observed the following.
```python
input = np.ones((1,2))
function = np.median

vectorized_filter(input, function, size=2)                 # OK
vectorized_filter(input, function, size=2, axes=0)         # FAILS
vectorized_filter(input, function, size=(2,), axes=0)      # OK
vectorized_filter(input, function, size=(2,), axes=(0,1))  # FAILS
vectorized_filter(input, function, size=2, axes=(0,1))     # OK
```

The [scipy.ndimage.vectorized_filter](https://scipy.github.io/devdocs/reference/generated/scipy.ndimage.vectorized_filter.html#scipy.ndimage.vectorized_filter) doc:
> axes (tuple of int, optional)
>
>    If None, input is filtered along all axes. Otherwise, input is filtered along the specified axes. When axes is specified, the dimensionality of footprint and the length of any tuples used for size or origin must match the length of axes. The ith axis of footprint and the ith element in these tuples corresponds to the ith element of axes.

According the the docs, 5th should fail (and doesn't, which I think that is a good thing).

### Describe the solution you'd like.

My question is:
**Could we make it so that 2nd also naturally work for scalar `size`?**

### Additional context (e.g. screenshots, GIFs)

v1.16.0rc1

mdhaber:
I did run into this and thought about changing it - I'm surprised someone else is already asking! 

IIRC, scalar `size` gets converted to a tuple with the dimensionality of `input`, not that of `axes`. That's why 5 ends up working - `input` and `axes` have the same number of dimensions. It should be simple to change it to become a tuple with the same number of dimensions as `axes`, instead. I think that will always do the same thing in cases that currently work, but enable cases that don't work right now.

ego-thales:
> I'm surprised someone else is already asking!

I was eagerly waiting for float16 support with `median_filter`, so I ran into this while updating :)

>  It should be simple to change it to become a tuple with the same number of dimensions as `axes`, instead. I think that will always do the same thing in cases that currently work, but enable cases that don't work right now.

I think it is indeed a sound approach, I personally don't see a good argument in favor of matching dimensionality of `input`.

mdhaber:
[...]

Issue #22576, opened 2025-02-23
DOC/BUG?: ndimage: correspondence between `axes` and `size`?


mdhaber:
The documentation of the `axes` argument of filters tends to look like:

> When `axes` is specified, any tuples used for `size` or `origin` must match the length of `axes`. The ith entry in any of these tuples corresponds to the ith entry in `axes`.

Based on this, I would expect that reversing the order of both `size` and `axes` should leave the output unchanged (as it does when `size` is square). However, this is not the case when `size` is not square:

```python3
import numpy as np
from scipy import ndimage
rng = np.random.default_rng(38952489254)

input = rng.integers(5, 10, size=(3, 3))
footprint_size, axes = (3, 1), (0, 1)

# the filter should have length of 3 along axis 0 and length 1 along axis 1
# The minimum filter should produce zero rows at the top and bottom 
out1 = ndimage.generic_filter(input, np.min, size=footprint_size, axes=axes, mode='constant')

# the filter should *still* have length of 3 along axis 0 and length 1 along axis 1
# The output should look the same, but instead I get zero columns at the left and right
out2 = ndimage.generic_filter(input, np.min, size=footprint_size[::-1], axes=axes[::-1], mode='constant')

np.testing.assert_equal(out1, out2)  # fails

out1
# array([[0, 0, 0],
#        [6, 6, 5],
#        [0, 0, 0]])

out2
# array([[0, 5, 0],
#        [0, 5, 0],
#        [0, 6, 0]])
```

I know that there are similar issues in higher dimensions and with `median_filter`.
More testing is needed, but this might affect all filters, and it might affect the correspondence between `axis` and `origin`. 

Does the documentation reflect the intended behavior, and am I reading it correctly?

Dronten:
can i work on this one

PR #23048, opened 2025-05-23
MAINT: ndimage.vectorized_filter: fix behavior of axes when length equals dimensionality of image

mdhaber:
#### Reference issue
Toward gh-23046

While investigating the feasibility of gh-23046, I noticed that `vectorized_filter` would not treat `axes` correctly when its length is equal to the dimensionality of the image. This fixes PR by treating it the same was is if the length of `axes` were less than the dimensionality of the image.

#### Additional information
This was harder to catch/test because other `ndimage` filters don't treat `axes` as documented (gh-22576).
I'd like to squeeze the gh-23046 into here if there's time, but it's more important that the bug fix makes the release.

### DIFF
--- a/scipy/ndimage/_filters.py
+++ b/scipy/ndimage/_filters.py
@@ -101,9 +101,10 @@ def _vectorized_filter_iv(input, function, size, footprint, output, mode, cval, 
                        "(`len(size)` or `footprint.ndim`) does not equal the number "
                        "of axes of `input` (`input.ndim`).")
             raise ValueError(message)
-        axes = (axes,) if np.isscalar(axes) else axes
     else:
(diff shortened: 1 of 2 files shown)


tylerjereddy (approved):
I checked locally that the test fails before and passes after the patch is applied.

This is brand new functionality in `1.16.0rc1`, and on top of that it doesn't even change the old tests for this functionality, just adds a new regression test.

Folks have had two weeks to comment and CI is passing.

I'm inclined to merge this before RC2 given the low risk and change in intended behavior of a brand new function and I believe @ev-br has a similar view on tweaks to this function at https://github.com/scipy/scipy/pull/23051#issuecomment-2909402842.

Issue #15994, opened 2022-04-16
ENH: how to treat single value `args` when a sequence is expected?


mdhaber:
gh-14138 reported [this](https://stackoverflow.com/questions/53262552/cant-integrate-with-dblquad) stack overflow issue in which a user attempted to pass a single value into the `args` argument of `dblquad` 
```python3
integrate.dblquad(g, 0, 1, lambda x: -1, lambda x: 2, args=(a))  # should be (a,)
```
instead of a tuple with one element 
```python3
integrate.dblquad(g, 0, 1, lambda x: -1, lambda x: 2, args=(a,))
```
and got an error message they didn't understand.

For better or worse, `args` is all over SciPy, so what should we do in these cases? A few suggestions from that issue:
- Add additional reminders in the documentation
- Make the error message more helpful
- Use `args=np.atleast_1d(args)` so that single values get packed into a sequence and can be treated like a valid input

dschmitz89:
I tested adding `np.atleast_1d()` for some simple optimization problems with DIRECT. For inexpensive functions, it causes severe overhead, increased optimization time by up to 30%. 

What about adding a simple test that `args` is indeed an iterable and if not a better error message.
```
from collections.abc import Iterable

if not isinstance(args, Iterable):
      raise ValueError("args must be an iterable. If only one argument is required, it should be written as (arg, ).")
```

Kai-Striega:
I'm +1 on adding a more useful error message. I consider `(a)` erroneous input and it should be treated as such.

mdhaber:
@dschmitz89 why did it cause the overhead? `np.atleast_1d` does not need to be inside the `fun` being optimized/integrated/etc; it can be called once as part of input validation. Do you mean that a single call was extremely slow?

dschmitz89:
> @dschmitz89 why did it cause the overhead? `np.atleast_1d` does not need to be inside the `fun` being optimized/integrated/etc; it can be called once as part of input validation. Do you mean that a single call was extremely slow?

Sorry for the late answer, busy days here unfortunately. I am not sure what was going on on my system but `%timeit np.atleast_1d((1., ))` actually only takes about 2 micro seconds, so it is probably going to be negligible as you wrote.

WarrenWeckesser:
A problem with using `atleast_1d` is that it always converts its argument to a numpy array, which could break existing code.  The values passed in as `args` can be arbitrary Python objects; these should be passed through unchanged.

[...]

PR #23051, opened 2025-05-24
MAINT/ENH: ndimage.vectorized_filter: adjust scalar size to match number of axes

### LINKED ISSUE #23046: ENH: `vectorized_filter`: special case of scalar `size` with `axes`

mdhaber:
#### Reference issue
Closes gh-23046

Implements the feature requested by gh-23046: allow `size` scalar to be converted to a tuple with the same length as `axes` rather than `input.ndim`.

#### Additional information
See discussion in gh-23046.


scipy/ndimage/_filters.py
```diff
@@ -150,10 +146,9 @@ def footprinted_function(input, *args, axis=-1, **kwargs):
 
     # For simplicity, work with `axes` at the end.
     working_axes = tuple(range(-n_axes, 0))
-    if axes is not None:
-        input = xp.moveaxis(input, axes, working_axes)
-        output = (xp.moveaxis(output, axes, working_axes)
-                  if output is not None else output)
+    input = xp.moveaxis(input, axes, working_axes)
+    output = (xp.moveaxis(output, axes, working_axes)
+              if output is not None else output)
```
mdhaber, line 151: No need to branch anymore because `axes` is always a tuple.

scipy/ndimage/_filters.py
```diff
@@ -455,7 +450,7 @@ def vectorized_filter(input, function, *, size=None, footprint=None, output=None
     res = function(view)
 
     # move working_axes back to original positions
-    return xp.moveaxis(res, working_axes, axes) if axes is not None else res
+    return xp.moveaxis(res, working_axes, axes)
```
mdhaber, line 453: No need to branch anymore because `axes` is always a tuple.

scipy/ndimage/_filters.py
```diff
@@ -68,12 +68,23 @@ def _vectorized_filter_iv(input, function, size, footprint, output, mode, cval,
...
         raise ValueError("Either `size` or `footprint` may be provided, not both.")
 
-    # Either footprint or size must be provided, and these determine the core
-    # dimensionality...
+    if axes is None:
+        axes = tuple(range(-input.ndim, 0))
+    elif np.isscalar(axes):
+        axes = (axes,)
+    n_axes = len(axes)
+    n_batch = input.ndim - n_axes
```
mdhaber, line 76: We now ensure that `axes` is always a tuple from the outset. If `axes` is `None`, the default is that all dimensions are core dimensions; otherwise `axes` determines the number of core/batch dimensions.

scipy/ndimage/_filters.py
```diff
@@ -68,12 +68,23 @@ def _vectorized_filter_iv(input, function, size, footprint, output, mode, cval,
...
+
+    if n_axes > input.ndim:
```
mdhaber, line 78: This check is similar in purpose to the one removed below - the core dimensionality can't exceed the `input` dimensionality.

scipy/ndimage/_filters.py
```diff
@@ -68,12 +68,23 @@ def _vectorized_filter_iv(input, function, size, footprint, output, mode, cval,
...
+        message = ("The length of `axes` may not exceed the dimensionality of `input`"
+                   "(`input.ndim`).")
+        raise ValueError(message)
+
+    # Either footprint or size must be provided
     footprinted_function = function
     if size is not None:
         # If provided, size must be an integer or tuple of integers.
-        size = (size,)*input.ndim if np.isscalar(size) else tuple(size)
+        size = (size,)*n_axes if np.isscalar(size) else tuple(size)
```
mdhaber, line 87: Here's the key change - `size` becomes a tuple with length `n_axes` rather than `input.ndim`. Consider the possibilities when `size` is a scalar:
- if `axes` was `None`, `n_axes == input.ndim`, so no change
- otherwise, if `axes` is a scalar (which gets converted to a length-1 tuple) or tuple with length not equal to `input.ndim`, this would have failed before, but works now.

So I think this only *adds* reasonable behavior.

scipy/ndimage/_filters.py
```diff
@@ -84,27 +95,12 @@ def _vectorized_filter_iv(input, function, size, footprint, output, mode, cval,
...
-    n_batch = input.ndim - n_axes
-
-    # ...which can't exceed the dimensionality of `input`.
-    if n_axes > input.ndim:
-        message = ("The dimensionality of the window (`len(size)` or `footprint.ndim`) "
-                   "may not exceed the number of axes of `input` (`input.ndim`).")
+    # And by now, the dimensionality of the footprint must equal the number of axes
+    if n_axes != len(size):
+        message = ("`axes` must be compatible with the dimensionality "
+                   "of the window specified by `size` or `footprint`.")
```
mdhaber, line 101: The check about whether `axes` is provided is no longer relevant. 

This check actually should have been here before, because instead of this nice error message, one would get e.g.:

```
>>> ndimage.vectorized_filter(img, np.mean, size=(2,), axes=(0, 1))
  File "<python-input-10>", line 1, in <module>
    ndimage.vectorized_filter(img, np.mean, size=(2,), axes=(0, 1))
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\matth\Desktop\scipy\scipy\ndimage\_support_alternative_backends.py", line 52, in wrapper
    result = func(*args, **kwds)
  File "C:\Users\matth\Desktop\scipy\scipy\ndimage\_filters.py", line 413, in vectorized_filter
    ) = _vectorized_filter_iv(input, function, size, footprint, output, mode, cval,
        ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
       origin, axes, batch_memory)
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\matth\Desktop\scipy\scipy\ndimage\_filters.py", line 154, in _vectorized_filter_iv
    input = xp.moveaxis(input, axes, working_axes)
  File "C:\Users\matth\miniforge3\envs\scipy-dev\Lib\site-packages\numpy\_core\numeric.py", line 1506, in moveaxis
    raise ValueError('`source` and `destination` arguments must have '
                     'the same number of elements')
ValueError: `source` and `destination` arguments must have the same number of elements
```

I suppose it just became more obvious that it's needed now.

mdhaber:
@ev-br I wouldn't necessarily ask for a review before 1.16 releases, but would you be able to assess whether you agree that this change is backward-compatible?

--

Currently, scalar `size` arguments are expanded into a tuple with length equal to the `input` dimensionality, like `(size,) * input.ndim`. So it's assumed that scalar `size` means "apply the filter with the same size across *all* dimensions of the input". From the documentation:

> We adjust `size` to the number of dimensions of the input array, so that, if the input array is shape (10,10,10), and `size` is 2, then the actual `size` used is (2,2,2).

This is fine when `axis=None` or `axis` is a tuple with the same number of elements as the `input` shape. But when `axes` is a smaller `tuple` - meaning that some dimensions are just batch dimensions - it *fails*. 

The request is to change the behavior so that instead of expanding `size` to to a tuple with length equal to the `input` dimensionality, it should be expanded to a tuple with the same number of elements as `axes`. The documentation might sound quite different in that case:

> We adjust `size` to the number of dimensions indicated by `axes`, so that, if `axes` is (0, 2, 1), and `size` is 2, then the actual `size` used is (2,2,2).

(The `axes` description already mentions that "If `None`, `input` is filtered along all axes.", so `axes=None` indicates `input.ndim` dimensions.)

So I think this preserves the existing behavior when `size` is a scalar and `axes=None` or is a tuple with length `input.ndim`, but adds behavior when it's a tuple with length `< input.ndim`. So do you think we can add this later even if it doesn't make this release?

scipy/ndimage/_filters.py
```diff
@@ -68,12 +68,23 @@ def _vectorized_filter_iv(input, function, size, footprint, output, mode, cval,
     if size is not None and footprint is not None:
         raise ValueError("Either `size` or `footprint` may be provided, not both.")
 
-    # Either footprint or size must be provided, and these determine the core
-    # dimensionality...
+    if axes is None:
+        axes = tuple(range(-input.ndim, 0))
```
eliegoudout, line 72: Just looking at this line, I'm wondering why not use `tuple(range(input.ndim))`?

mdhaber: Because that is the opposite of what I want (leftmost vs rightmost axes).

scipy/ndimage/_filters.py
eliegoudout: I think this works as perfectly as intended. However, note that this works perfectly with `size={1: "foo", 2: "bar"}` since you override `size` as long as it's not `None`or a scalar. It might silence wrong uses which may no be desirable. I would add a (albeit longer) triple condition
```python
if np.isscalar(size):
  ...
elif isinstance(size, tuple):
  ...
else:
  ...
```
Another simplification may be using `np.atleast_1d`?

scipy/ndimage/_filters.py
eliegoudout: I'm not sure I understand. I though this had the purpose of selecting all axes, in which case the two seem exactly equivalent to me. But I may be missing something since I don't have the whole context.

eliegoudout: Ok ok. I get where you come from then. I would personally change it though for readability/maintainability, but I don't think it matters much 👍

scipy/ndimage/_filters.py
mdhaber: >  It might silence wrong uses which may no be desirable.

Yeah - there's a balance to be struck. For better or for worse, SciPy has historically been very inclusive when it comes to this sort of input. I'd love to just insist on a `tuple` and be done with it, and maybe in SciPy 2.0 that's what we'll do. But until then, this maintains the expected flexibility. It's tough to say exactly where to draw the line, so I'd prefer not to try to filter out pathological inputs. If that were really important to SciPy, we'd have developed a private function for it by now rather than re-inventing it in each PR.

eliegoudout: For future reference:

The option `size = np.atleast_1d(size)` would remove the need for the `if` branching *and* make it so `arraylike` sizes are valid. Then one would only need to check that it is broadcatable with `axes`.

mdhaber: And I've used it in the past for similar purposes, so it's not new. See e.g. gh-15994. You can see there that there was a preference for something more like what is here. Now, the context was a little different there because `args` can have a little more variety than `axis`. But then again, the fact that the discussion didn't really converge suggested that this sort of thing wasn't really important to others, so I didn't attempt to optimize this line here. It sounds important to you, so if the maintainer who reviews this PR agrees, I can change it.

eliegoudout: It is kind of you to be open minded about this.

To be clear, I would personally not object to curent version being merged, **I think it is fine** :) I simply saw a good occasion to discuss these possibilities, and I thank you for you POVs.

All the best!

ev-br:
@mdhaber I'm not sure we need to worry about backwards compatibility already. The function is fresh from the foundry, so fixing edge cases is OK, I'd think. If anything, this can go in in the rc phase, too.
So it boil down to what is the best behavior when `size` is a scalar: does it get expanded to `input.shape` or `axes.shape`. The latter makes sense to me, but I don't have a very good intuition here. 
On balance, I think we can make the change in the 1.16.x cycle.

mdhaber:
Ok. I already considered making this the behavior in the original PR, so I think among the three of us there is enough confidence that this is the desired behavior. I'll update the docstring accordingly. If you're willing to review/merge, awesome; if not, let me know and I'll reach out to others.

tylerjereddy:
@ev-br @mdhaber do you want this in RC2 or not?

mdhaber:
I'd like to get it in since I'm quite confident that this is useful behavior, but it's up to the reviewer (maybe @ev-br?). I'll push doc updates shortly so the ball is in their court.

tylerjereddy:
Ok, note that I merged the "sister" PR as cross-linked above, since that was an easier decision to make and it had an actual backport label.

I'll check this PR again in the morning, maybe put a backport label on it if you want it in, there are no other unmerged backport candidates left.

ev-br:
This looks reasonable, let's include it in 1.16.0 indeed. Thanks @mdhaber for the fix, @eliegoudout for the review.

scipy/ndimage/tests/test_filters.py
```diff
@@ -2986,7 +2990,19 @@ def test_edge_cases(self, xp):
...
+        # The intent of gh-23046 was to always allow `size` to be a scalar.
+        rng = np.random.default_rng(45982734597824)
+        img = xp.asarray(rng.random((5, 5)))
+
+        ref = ndimage.vectorized_filter(img, xp.mean, size=2)
+        res = ndimage.vectorized_filter(img, xp.mean, size=2, axes=(0, 1))
+        xp_assert_close(res, ref)
+
+        ref = ndimage.vectorized_filter(img, xp.mean, size=(2,), axes=(0,))
+        res = ndimage.vectorized_filter(img, xp.mean, size=2, axes=0)
```
tylerjereddy, line 3003: @mdhaber was the `xp_assert_close` left out on purpose here? At first I thought it might be just make sure things don't error in these statements, but then assigning to `ref` and `res` may not be of much use?

mdhaber: Oops, no : / They should be compared. I can submit a PR tonight.

tylerjereddy: I already have it on the release branch locally, so no rush, maybe just do it on `main` when you have time so I don't have to forward port it

PR #23012, opened 2025-05-18
BENCH: asv benchmarks for alternative array backends

crusaderky:
- New infrastructure for running asv on the Array API backends, which shares machinery with `conftest` and `@xp_capabilities`.
- Bew bench test for `cluster.hierarchy.is_isomorphic` (part of #22989; kept in a separate PR for obvious reasons)
- New bench test for `cluster.vq.whiten`
- Bench test for `cluster.hierarchy.linkage` now runs on alternative backends

# Known issues
`spin bench -b all --compare`, as well as bare `asv`, create an environment without the Array API optional dependencies. I did not figure out how to add them.

Without the `--compare` flag it works as expected.

# Demo
```
$ spin bench -b all -t cluster.IsIsomorphic 
[50.00%] ··· ================== ========== ==========
             --                          nobs        
             ------------------ ---------------------
                  backend          100       100000  
             ================== ========== ==========
                   numpy         168±0μs    27.8±0ms 
              array_api_strict   343±0μs    29.0±0ms 
                    cupy           n/a        n/a    
                 torch:cpu       391±0μs    28.5±0ms 
                 torch:cuda        n/a        n/a    
                 dask.array      5.06±0ms   34.5±0ms 
               jax.numpy:cpu     736±0μs    30.4±0ms 
               jax.numpy:cuda      n/a        n/a    
             ================== ========== ==========

[50.00%] ···· For parameters: 'cupy', 100
              asv: skipped: SkipNotImplemented('cupy skipped by @xp_capabilities')
              
              For parameters: 'cupy', 100000
              asv: skipped: SkipNotImplemented('cupy skipped by @xp_capabilities')
              
              For parameters: 'torch:cuda', 100
              asv: skipped: SkipNotImplemented('torch skipped by @xp_capabilities')
              
              For parameters: 'torch:cuda', 100000
              asv: skipped: SkipNotImplemented('torch skipped by @xp_capabilities')
              
              For parameters: 'jax.numpy:cuda', 100
              asv: skipped: SkipNotImplemented('jax.numpy skipped by @xp_capabilities')
              
              For parameters: 'jax.numpy:cuda', 100000
              asv: skipped: SkipNotImplemented('jax.numpy skipped by @xp_capabilities')
```
# Lessons learned
- On CUDA, do not trust the documentation about the default stream being synchronous or not. Turns out that the default stream on CuPy is asynchronous.
[...]

PR #23057, opened 2025-05-26
ENH: Speed up `array_namespace` through caching

crusaderky:
Related:
- https://github.com/data-apis/array-api-compat/pull/308 (included in all benchmarks below)
- https://github.com/data-apis/array-api-compat/pull/329 (all benchmarks below both with and without)
- https://github.com/scipy/scipy/pull/23175

In this PR:
- 4x to 8x speed-up when calling `array_namespace` on numpy arrays, array-likes, and scalars
- 1.4x speed-up on other Array API objects.
- add test for passing a mix of `np.float64` and non-numpy arrays to functions
- fix JAX zero-gradient arrays, which would previously raise
- run more Array API tests in CI

```python
>>> import numpy as np
>>> from scipy._lib._array_api import array_namespace
>>> a = np.empty(0)

# Scalar
>>> %timeit array_namespace(1)
# main           1.87 μs ± 8.14 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
# #329           1.91 μs ± 6 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
# #23507         212 ns ± 7.02 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
# #329 + #23057  213 ns ± 5.56 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

# NumPy Array
>>> %timeit array_namespace(a)
# main           2.69 μs ± 19.6 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
# #329           2.07 μs ± 16.1 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
# #23507         299 ns ± 2.14 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
# #329 + #23057  299 ns ± 3.16 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

# Array-like
>>> %timeit array_namespace([1, 2])
# main           3.56 μs ± 30.5 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
# #329           2.87 μs ± 5.01 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
# #23507         833 ns ± 13.6 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
# #329 + #23057  841 ns ± 19.9 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

# Array API object other than NumPy
>>> import jax.numpy as jnp
>>> ja = jnp.empty(0)
>>> %timeit array_namespace(ja)
# main           2.37 μs ± 24.3 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
# #329           1.2 μs ± 9.84 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
# #23507         1.68 μs ± 23.9 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
# #329 + #23057  757 ns ± 22.3 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

# Unwrapped array_api_compat.array_namespace
>>> from scipy._lib.array_api_compat import array_namespace as compat_namespace
>>> %timeit compat_namespace(a)
# main           1.2 μs ± 14.6 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
# #329           506 ns ± 2.39 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

>>> %timeit compat_namespace(ja)
# main           1.25 μs ± 16.1 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
# #329           355 ns ± 1.36 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

# SCIPY_ARRAY_API not set
>>> import scipy
>>> scipy._lib._array_api_override.SCIPY_ARRAY_API = False
>>> %timeit array_namespace(a)
# main           61.5 ns ± 0.438 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
# #23507         65.6 ns ± 0.248 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

Downstream of this PR, we can still squeeze quite a bit out of `array_api_compat.array_namespace` to further speed up the case of non-numpy array api objects.

ASV tests blocked by https://github.com/scipy/scipy/pull/23012.


scipy/_lib/_array_api_override.py
```diff
@@ -99,24 +97,66 @@ def array_namespace(*arrays: Array) -> ModuleType:
...
+                else:
+                    # void dtype, but not a JAX zero gradient array
+                    raise TypeError(f"An argument has dtype `{array.dtype!r}`; "
+                        f"only boolean and numerical dtypes are supported.")
+            else:
+                numpy_arrays.append(array)
+        elif arr_info is _ArrayClsInfo.no_info and is_array_api_obj(array):
+            api_arrays.append(array)
+        else:
+            # list, tuple, or arbitrary object
```
crusaderky, line 146: we could drastically speed up the use case of ArrayLikes by pushing this validation to `_asarray`. However, that would require sanitizing all calls to `xp.asarray` throughout the codebase, so I'm leaving it to a follow-up.

scipy/_lib/_array_api_override.py
```diff
@@ -99,24 +81,70 @@ def array_namespace(*arrays: Array) -> ModuleType:
 
     Notes
     -----
-    Thin wrapper around `array_api_compat.array_namespace`.
```
crusaderky, line 102: This isn't thin :stuck_out_tongue_closed_eyes:

tylerjereddy (commented):
The benchmarks seem consistent locally on an M3 mac as well:

`main`:
```
In [2]: %timeit array_namespace(1)
902 ns ± 6.21 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [3]: %timeit array_namespace(1)
900 ns ± 4.4 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```
this branch:
```
In [2]: %timeit array_namespace(1)
108 ns ± 0.246 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [3]: %timeit array_namespace(1)
108 ns ± 0.318 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

Switching to the sometimes-preferred `@cache` didn't make much difference:

```
In [2]: %timeit array_namespace(1)
107 ns ± 0.263 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [3]: %timeit array_namespace(1)
107 ns ± 0.197 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

crusaderky:
I've updated the opening post to include measures with and without https://github.com/data-apis/array-api-compat/pull/329.
PR to pull it in: https://github.com/scipy/scipy/pull/23175

crusaderky:
@lucascolley @rgommers can we push this through the finishing line?

andyfaff:
Has this been checked for compatibility with multiprocessing?  E.g. functions that use array_namespace still need to be pickleable.

crusaderky:
> Has this been checked for compatibility with multiprocessing? E.g. functions that use array_namespace still need to be pickleable.

This is a `@lru_cache` (which itself is never included in pickles), by a function called internally by the actual functions which users will end up pickling. This is safe for multiprocessing.

lucascolley (approved):
thanks Guido, this LGTM! Will merge soon if no other feedback.

andyfaff:
Have there been any memory leak checks done here? We've had several instances of caches causing leaks because objects don't get garbage collected.  @nickodell is good at checking for things like that.

nickodell:
> Have there been any memory leak checks done here? We've had several instances of caches causing leaks because objects don't get garbage collected. @nickodell is good at checking for things like that.

I did not test this, but this looks like a safe design.

Since `lru_cache()` is being applied to classes of arrays, not arrays, the cache shouldn't keep the arrays themselves around for longer. Those array classes will probably live for the duration of the Python process anyway. (The user could call `np.ndarray` at any time, so the Python interpreter must keep that class alive forever.)

I also agree with the decision put a bound on the number of items in cache here. It is unlikely, but the user might repeatedly dynamically create a class and pass it to some SciPy function. In that case we don't want to hold on to an unlimited number of cache entries.

scipy/_lib/_array_api_override.py
```diff
@@ -28,60 +29,41 @@
...
+                'would work instead.')
+        raise ValueError(msg)
+
+    if issubclass(cls, np.ma.MaskedArray):
+        raise TypeError("Inputs of type `numpy.ma.MaskedArray` are not supported.")
+
+    if issubclass(cls, np.matrix):
+        raise TypeError("Inputs of type `numpy.matrix` are not supported.")
+
+    if issubclass(cls, np.ndarray | np.generic):
```
nickodell: I'm not sure if this is a meaningful amount of performance in the big picture, but

```
issubclass(cls, (x, y))
```

is slightly faster than

```
issubclass(cls, x | y)
```

Micro-benchmark:

```python
import scipy
import numpy as np

my_array = np.zeros((10, 10))
my_matrix = scipy.sparse.coo_matrix(my_array)

M_type = type(my_array)
%timeit issubclass(M_type, np.ndarray | np.generic)
# 417 ns ± 2.94 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
%timeit issubclass(M_type, (np.ndarray, np.generic))
# 224 ns ± 1.37 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

# Difference increases with more types checked
M_type = type(1234)
%timeit issubclass(M_type, int | float | complex | bool | type(None))
# 1.11 µs ± 4.59 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
%timeit issubclass(M_type, (int, float, complex, bool, type(None)))
# 299 ns ± 2.57 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

You should consider whether to use `issubclass(cls, (x, y))` here in the three places where you check whether a class is a subclass of a union of types. Alternatively, you could decide that the performance inside the cached function doesn't matter. I have no strong feelings one way or another.

lucascolley: I like how the `|` syntax looks and have argued on that ground in non-performance-critical places before, however I see now that the ruff rule recommending it has been deprecated: https://docs.astral.sh/ruff/rules/non-pep604-isinstance/.

Given that, I think the 'more readable' syntax is likely to become more rare over time, so the benefit I saw is no longer there. Let's replace it with the faster syntax here.

scipy/_lib/_array_api_override.py
```diff
@@ -28,60 +29,41 @@
...
-                raise TypeError(message)
-            # Ignore 0-dimensional arrays, coherently with array-api-compat.
-            # Raise if there are 1+-dimensional array-likes mixed with non-numpy
-            # Array API objects.
-            if array.ndim:
-                yield array
+
+@lru_cache(100)
+def _validate_array_cls(cls: type) -> _ArrayClsInfo:
+    if issubclass(cls, list | tuple):
```
lucascolley:
```suggestion
    if issubclass(cls, (list,  tuple)):
```

scipy/_lib/_array_api_override.py
lucascolley:
```suggestion
    if issubclass(cls, (np.ndarray, np.generic)):
```

scipy/_lib/_array_api_override.py
```diff
@@ -28,60 +29,41 @@
...
+    if issubclass(cls, np.matrix):
+        raise TypeError("Inputs of type `numpy.matrix` are not supported.")
+
+    if issubclass(cls, np.ndarray | np.generic):
+        return _ArrayClsInfo.numpy
+
+    # Note: this must happen after the test for np.generic, because
+    # np.float64 and np.complex128 are subclasses of float and complex respectively.
+    # This matches the behavior of array_api_compat.
+    if issubclass(cls, int | float | complex | bool | type(None)):
```
lucascolley:
```suggestion
    if issubclass(cls, (int, float, complex, bool, type(None))):
```

lucascolley:
Looks like we'll need to add UP038 to https://github.com/scipy/scipy/blob/main/tools/lint.toml#L24 until the deprecation is executed.

lucascolley:
thanks all!

PR #23058, opened 2025-05-26
CI: clean up free-threading job, use released Cython/Pythran

rgommers:


lucascolley (approved):
thanks Ralf!

crusaderky:
Could it be useful to add a [freethreading] optional dependency that pins the higher version of pythran?

rgommers:
It's a build dependency, and there are no optional build dependencies. So not really. There is no way to get this wrong really when creating a new environment, so I'd prefer to avoid a hack until we can properly express it with an environment marker in `build-system.requires` in `pyproject.toml`.

PR #23059, opened 2025-05-26
MAINT: remove nan-handling from _factorialx_array_exact

h-vetinari:
As a consequence of tightening the types allowed in `factorial{,2,k}` -- in particular, that `exact=True` is only allowed for actual integers -- we can now remove some unused nan-handling in the exact calculation.


lucascolley (approved):
thanks @h-vetinari !

PR #23060, opened 2025-05-27
DOC: Revise a footnote in meson-distutils

star1327p:

Revise a footnote in meson-distutils.
https://docs.scipy.org/doc/scipy/building/distutils_equivalents.html

#### Additional information

The commit has the label `[docs only]` to save CI resources.


rgommers (approved):
Thanks for the fix @star1327p!

PR #23062, opened 2025-05-27
DEV: help debug free-threading build of Fortran modules

crusaderky:
This just cost me way too many hours to debug.
XREF https://github.com/mesonbuild/meson/issues/14651

Scipy warning (which is an error when running tests) below for the benefit of github searches:

> RuntimeWarning: The global interpreter lock (GIL) has been enabled to load module 'scipy.linalg._fblas', which has not declared that it can run safely without the GIL. To override this behavior and keep the GIL disabled (at your own risk), run with PYTHON_GIL=0 or -Xgil=0.

besides `linalg`, I can see fortran modules using this in `integrate`, `interpolate`, and `io`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -153,6 +153,7 @@ benchmarks/html
 benchmarks/scipy-benchmarks
 .github/workflows/.pixi
 .openblas
+scipy-openblas.pc
 scipy/_distributor_init_local.py
 scipy/__config__.py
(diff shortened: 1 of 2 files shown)


lucascolley (approved):
thanks Guido! Not sure if we want the meson issue link in long-term, but seems like a good idea to include it for now.

PR #23065, opened 2025-05-27
ENH: stats.tukey_hsd: improve array index iteration in ``TukeyHSDResult.__str__``

jorenham:
This fixes an issue where the wrong axis length was used during iteration in ``scipy.stats._hypotests.TukeyHSDResult.__str__``, by using the less error-prone ``np.ndindex`` as index iterator. Note that this wasn't causing any problems at runtime at the moment, because as far as I can tell, the relevant arrays are always square. Either way, this way the code is less confusing :shrug:.

### DIFF
--- a/scipy/stats/_hypotests.py
+++ b/scipy/stats/_hypotests.py
@@ -1749,13 +1749,12 @@ class TukeyHSDResult:
         s = ("Pairwise Group Comparisons"
              f" ({self._ci_cl*100:.1f}% Confidence Interval)\n")
         s += "Comparison  Statistic  p-value  Lower CI  Upper CI\n"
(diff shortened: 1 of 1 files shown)


mdhaber (approved):
Yup, the arrays are always square, but I agree that this is a bit simpler. I'll merge when CI finishes.

tylerjereddy:
There's no clearly stated performance improvement or behavior change, so I'll strip the release note label. Just make a suggestion over at gh-24100 is there is a concrete user-facing gain we should mention.

jorenham:
Yea, this probably worth mentioning in the release notes. Thanks for the reviews :)

Issue #23064, opened 2025-05-27
ENH: Add a normalize parameter to scipy.stats.entropy to optionally skip input normalization


NeriCarcasci:
### Is your feature request related to a problem? Please describe.

`scipy.stats.entropy` automatically normalizes the input distributions `pk` and `qk` to sum to 1. This is helpful in many cases, but limits flexibility when the inputs are already normalized or when users want to work with unnormalized weights (e.g. for cross-entropy or variational methods).

It also adds unnecessary computation when users are working with large batches of normalized data, such as model outputs or preprocessed datasets.

### Describe the solution you'd like.

Add a keyword-only parameter normalize (defaulting to True) to scipy.stats.entropy that allows users to disable internal normalization.

_Example:_
```
from scipy.stats import entropy
import numpy as np

pk = np.array([2.0, 8.0])   # Not normalized
qk = np.array([1.0, 9.0])   # Not normalized

D1 = entropy(pk, qk)                    # Normalized by default
D2 = entropy(pk, qk, normalize=False)   # Uses raw inputs
```

### Describe alternatives you've considered.

- Using scipy.special.entr and rel_entr to manually compute entropy without normalization is possible, but it’s verbose, error-prone, and inconsistent with the convenience scipy.stats.entropy offers.
- Reimplementing entropy logic from scratch, which duplicates code already available in SciPy.

steppi:
At a glance I'm -1 on this one and would prefer to keep the API simpler.  `scipy.special.entr` and `scipy.special.rel_entr` seem like a good fit for cases where one has large batches of normalized data or is working with something more exotic than a regular probability distribution.  I'd be interested in seeing the problems you've run into with `scipy.special.entr` and `scipy.special.rel_entr` though.

mdhaber:
@NeriCarcasci could we resolve this with an example? For instance, "`stats.entropy` normalizes the argument(s) to have unit sum. If the argument(s) are already normalized, consider using `special.entr` for Shannon entropy or `special.rel_entr` for relative entropy and summing the output."

```python3
from scipy import special
pk = [0.1, 0.2, 0.3, 0.4]  # already normalized
qk = [0.2, 0.2, 0.3, 0.3]
np.allclose(special.entr(pk).sum(), entropy(pk))  # Shannon entropy
np.allclose(special.rel_entr(pk).sum(), entropy(pk))  # relative entropy
```

PR #23069, opened 2025-05-28
ENH: stats.entropy: add normalize parameter to skip input normalization

NeriCarcasci:
#### Reference issue
#23064 

Added a keyword-only argument `normalize` to scipy.stats.entropy, defaulting to True.
	•	When `normalize=True` (default), input arrays pk and qk are normalized to sum to 1 along the specified axis.
	•	When `normalize=False`, inputs are treated as already normalized, and no internal normalization is performed.

This addition offers more flexibility and avoids redundant normalization when it’s unnecessary.

### DIFF
--- a/scipy/stats/_entropy.py
+++ b/scipy/stats/_entropy.py
@@ -27,8 +27,9 @@ __all__ = ['entropy', 'differential_entropy']
 def entropy(pk: np.typing.ArrayLike,
             qk: np.typing.ArrayLike | None = None,
             base: float | None = None,
-            axis: int = 0
-            ) -> np.number | np.ndarray:
+            axis: int = 0,
+            *, 
+            normalize: bool = True) -> np.number | np.ndarray:
     """
     Calculate the Shannon entropy/relative entropy of given distribution(s).
 
@@ -39,7 +40,7 @@ def entropy(pk: np.typing.ArrayLike,
     ``D = sum(pk * log(pk / qk))``. This quantity is also known
     as the Kullback-Leibler divergence.
 
-    This routine will normalize `pk` and `qk` if they don't sum to 1.
+    By default, this routine normalizes pk and qk so they sum to 1 along the given axis. Set normalize=False to skip this step.
 
     Parameters
     ----------
@@ -54,6 +55,9 @@ def entropy(pk: np.typing.ArrayLike,
         The logarithmic base to use, defaults to ``e`` (natural logarithm).
     axis : int, optional
         The axis along which the entropy is calculated. Default is 0.
+    normalize : bool, optional
+        If True (default), `pk` and `qk` will be normalized to sum to 1 along the specified axis.
+        If False, the inputs are assumed to already represent valid distributions, and no normalization is applied. 
 
     Returns
     -------
@@ -137,6 +141,13 @@ def entropy(pk: np.typing.ArrayLike,
     >>> CE == -np.sum(pk * np.log(qk)) / np.log(base)
     True
 
+    Using unnormalized inputs with `normalize=False`:
+
+    >>> pk = np.array([2.0, 8.0])
+    >>> qk = np.array([1.0, 9.0])
+    >>> entropy(pk, qk, base=2, normalize=False)
+    0.08164398904051034
+
     """
     if base is not None and base <= 0:
         raise ValueError("`base` must be a positive number or `None`.")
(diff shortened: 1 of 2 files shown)


scipy/stats/_entropy.py
```diff
@@ -137,6 +141,14 @@ def entropy(pk: np.typing.ArrayLike,
...
     True
 
+    Using unnormalized inputs with `normalize=False`:
+
+    >>> pk = np.array([2.0, 8.0])
+    >>> qk = np.array([1.0, 9.0])
+    >>> entropy(pk, qk, base=2, normalize=False)
+    0.6405999884615012
+
+    .. versionadded:: 1.15.4
```
tylerjereddy, line 151: Two comments here:

1) If the team decides to move forward with this, we'll need to increase this version number, probably to `1.17.0`. We also pretty much never add new features in patch releases, but that's just unnecessary commentary from the release manager :)
2) I'm pretty sure we usually place `versionadded` directives right under the new parameters in the docstring rather than in the `Examples` block. For examples, `git grep` should do the trick.

scipy/stats/tests/test_entropy.py
```diff
@@ -115,6 +115,26 @@ def test_input_validation(self, xp):
...
+    def test_entropy_normalize_true_equals_false_when_inputs_normalized(self, xp):
+        pk = xp.asarray([0.05, 0.2, 0.1, 0.05, 0.15, 0.05, 0.2, 0.1, 0.05, 0.05])
+        qk = xp.asarray([0.05, 0.15, 0.1, 0.1, 0.15, 0.1, 0.15, 0.1, 0.05, 0.05])
+
+        res1 = stats.entropy(pk, qk, base=2)
+        res2 = stats.entropy(pk, qk, base=2, normalize=False)
+
+        expected = xp.asarray(0.0660149997115376) 
+        xp_assert_close(res1, expected)
+        xp_assert_close(res2, expected)
```
tylerjereddy, line 136: The full testsuite passes for me locally on ARM Mac, including with array API testing active. The CI failures that I see here are unrelated and were fixed by Guido a while back now, so rebasing can solve those.

scipy/stats/tests/test_entropy.py
```diff
@@ -115,6 +115,26 @@ def test_input_validation(self, xp):
...
+
+        expected = xp.asarray(2.190590872098243)
+        xp_assert_close(result, expected)
+
+    def test_entropy_normalize_true_equals_false_when_inputs_normalized(self, xp):
+        pk = xp.asarray([0.05, 0.2, 0.1, 0.05, 0.15, 0.05, 0.2, 0.1, 0.05, 0.05])
+        qk = xp.asarray([0.05, 0.15, 0.1, 0.1, 0.15, 0.1, 0.15, 0.1, 0.05, 0.05])
+
+        res1 = stats.entropy(pk, qk, base=2)
+        res2 = stats.entropy(pk, qk, base=2, normalize=False)
```
tylerjereddy, line 132: A minor comment is that we could also parametrize across the value of `normalize`, but it hardly seems worth the churn.

tylerjereddy (commented):
I see that this one has been sitting for a bit without review so I added a few comments.

Before putting more effort into it, we should probably ping @mdhaber for a "temperature check" on whether `stats` folks are interested in it. The original motivations make sense to a non-expert, but I believe we do also have to be conscious of i.e., API bloat where we keep adding more and more kwargs to bolt-on new functionality.

mdhaber:
> Before putting more effort into it, we should probably ping @mdhaber for a "temperature check" on whether stats folks are interested in it

I was not, but I didn't want to get in the way. But if I'm asked:
- I think there are a lot of functions that could skip input standardization steps. I'd prefer not to treat this function in isolation. (I'm not particularly interested in tackling the bigger picture at this time, either, though.)
- gh-23064 acknowledges that "Using `scipy.special.entr` and `rel_entr` to manually compute entropy without normalization is possible" but concludes that "it’s verbose, error-prone, and inconsistent with the convenience `scipy.stats.entropy` offers". I don't really agree with the sentiment - `stats.entropy` does so little that it's reasonable to either use it as-is or call the relevant special function and sum if speed is that important.

mdhaber:
I'll go ahead and close this since we have three maintainers concerned about the API bloat (together with @steppi in gh-23064). We can continue the discussion about resolving the need in gh-23064, though.

PR #21683, opened 2024-10-09
ENH:MAINT:integrate: Rewrite DOP F77 code in C

ilayn:
Towards #18566
Towards #23067 

Mechanical translation of dopri5 and dopri853 integration routines. Only a bit of restructuring and glue code is introduced instead of the f2py bridges.

Closes #4487

PR #23070, opened 2025-05-28
MAINT: Convert `_fpumode` to multi-phase init (PEP 489)

AA-Turner:
Towards GH-23067.

This is an initial conversion to multi-phase init for one of the simplest extension modules, `_fpumode`. cc @ev-br @ilayn.

A


rgommers (commented):
Thanks @AA-Turner. This looks good, the one thing to discuss is the value of `Py_mod_multiple_interpreters`.  I think it'd be better to use `Py_MOD_MULTIPLE_INTERPRETERS_NOT_SUPPORTED` here. There are multiple reasons why subinterpreters are not supported yet, (NumPy, Pybind11, static types within SciPy, etc.) so having to reason about this per extension module makes review harder - the rest is pretty much mechanical. Cython also generates code with `_NOT_SUPPORTED` at the moment; there are no instances within SciPy of `Py_MOD_PER_INTERPRETER_GIL_SUPPORTED`.

ilayn:
From the look of it, this does not have too much work since it is just a switch to designated initializers (if we skip the subinterpreter details). We can do this in one go.

rgommers:
> We can do this in one go.

Agreed, or in a handful of PRs, so they're of reasonable size - we don't need one PR per extension module.

AA-Turner:
> Lucas Colley changed the title
> Convert _fpumode to multi-phase init (PEP 489)
> ENH: Convert _fpumode to multi-phase init (PEP 489)

nota bene: NumPy classified the equivalent series of PRs as "MAINT", though I'm not familar with the finer points of the distinction.

AA-Turner:
> this looks good, the one thing to discuss is the value of `Py_mod_multiple_interpreters`. I think it'd be better to use `Py_MOD_MULTIPLE_INTERPRETERS_NOT_SUPPORTED` here.

Happy to do so. Note that `Py_MOD_MULTIPLE_INTERPRETERS_NOT_SUPPORTED` both only exists for Python 3.12+ and in fact only restricts execution to the main interpreter, so I would suggest also adding the [opt-out `ImportError`](https://docs.python.org/dev/howto/isolating-extensions.html#opt-out-limiting-to-one-module-object-per-process) for explicitness. Are you happy to proceed thus?

> From the look of it, this does not have too much work since it is just a switch to designated initializers

Designated initializers are strictly optional, I'm using them for better readability. I can remove that part of the change if you'd prefer. The key details are moving 'module execution' to the `{mod}_exec` function (and changing the returns), adding the slots array, and making `m_size` non-negative.

A

rgommers:
> nota bene: NumPy classified the equivalent series of PRs as "MAINT", though I'm not familar with the finer points of the distinction.

Yeah I think this is MAINT; nothing really in it for end users, just moving us ahead on tech debt

rgommers:
> Note that `Py_MOD_MULTIPLE_INTERPRETERS_NOT_SUPPORTED` both only exists for Python 3.12+

Argh, I missed that. That is not helpful - let me have a closer look. We have 6 more months to go until dropping 3.11.

AA-Turner:
For reference, the `ImportError` approach is preferred in the C API documentation: https://docs.python.org/dev/c-api/module.html#multi-phase-initialization: "*A simpler way to avoid these issues is raising an error on repeated initialization.*".

rgommers:
> Note that `Py_MOD_MULTIPLE_INTERPRETERS_NOT_SUPPORTED` both only exists for Python 3.12+ and in fact only restricts execution to the main interpreter, so I would suggest also adding the [opt-out `ImportError`](https://docs.python.org/dev/howto/isolating-extensions.html#opt-out-limiting-to-one-module-object-per-process) for explicitness. Are you happy to proceed thus?

Yes, assuming you mean using the exact same pattern as in your merged NumPy PRs (?). E.g. https://github.com/numpy/numpy/pull/29028, that seems fine. SciPy has the same Python version support range and other requirements as NumPy. It's similar as the example you linked to, except for the `PyMutex` usage. That `PyMutex` usage also isn't portable to earlier Python versions, so the linked example isn't really usable by any widely used library yet.

AA-Turner:
Thank you Lucas!

A

ev-br:
> Designated initializers are strictly optional, I'm using them for better readability. 

Agreed, they are a readability improvement where available. in C++ they are a c++20 feature, correct? If so, would be best to include a comment to this effect in converted c++ modules (scipy is on c++17, similar to numpy).

ilayn:
That's part of CPython API so it won't matter what C++ is, it will be extern'd. C99 onward they are in the standard for C.

scipy/_lib/_fpumode.c
```diff
@@ -47,32 +47,28 @@ static struct PyMethodDef methods[] = {
     {NULL, NULL, 0, NULL}
 };
 
+static struct PyModuleDef_Slot _fpumode_slots[] = {
+#if PY_VERSION_HEX >= 0x030c00f0  // Python 3.12+
+    // signal that this module can be imported in isolated subinterpreters
+    {Py_mod_multiple_interpreters, Py_MOD_PER_INTERPRETER_GIL_SUPPORTED},
```
tylerjereddy, line 51: @rgommers I took a stab at addressing your comments locally (diff below the fold), and the full testsuite passed after rebase for both my modified version and the current version.

<details>

```diff
--- a/scipy/_lib/_fpumode.c
+++ b/scipy/_lib/_fpumode.c
@@ -9,6 +9,7 @@
 #pragma fenv_access (on)
 #endif
 
+static int module_loaded = 0;
 
 static char get_fpu_mode_doc[] = (
     "get_fpu_mode()\n"
@@ -49,8 +50,7 @@ static struct PyMethodDef methods[] = {
 
 static struct PyModuleDef_Slot _fpumode_slots[] = {
 #if PY_VERSION_HEX >= 0x030c00f0  // Python 3.12+
-    // signal that this module can be imported in isolated subinterpreters
-    {Py_mod_multiple_interpreters, Py_MOD_PER_INTERPRETER_GIL_SUPPORTED},
+    {Py_mod_multiple_interpreters, Py_MOD_MULTIPLE_INTERPRETERS_NOT_SUPPORTED},
 #endif
 #if PY_VERSION_HEX >= 0x030d00f0  // Python 3.13+
     // signal that this module supports running without an active GIL
@@ -70,5 +70,10 @@ static struct PyModuleDef moduledef = {
 PyMODINIT_FUNC
 PyInit__fpumode(void)
 {
+    if (module_loaded) {
+        PyErr_SetString(PyExc_ImportError,
+                        "cannot load module more than once per process");
+        return -1;
+    }
     return PyModuleDef_Init(&moduledef);
```

</details>

Does that look right? Is that what you want? NumPy seems to have one more level of abstraction for the sub interpreter guard with `Py_mod_exec`, but not sure it matters (@ilayn did use that in the cross-ref'd SciPy example as well).

I also note that some things appear to have changed in the past few months, since `git grep -E -i "Py_MOD_PER_INTERPRETER_GIL_SUPPORTED"` now has hits in our codebase.

An example of where this was added was @ilayn's work at gh-21683. Do we want `module_loaded` guards like above added in each case of this in SciPy for now or?

rgommers: Thanks for the update Tyler. I'll have to re-read this more carefully, am at a conference now.

I think we can drop 3.11 support right after the release branch creation, so we can then use `Py_MOD_MULTIPLE_INTERPRETERS_NOT_SUPPORTED` indeed.

ilayn: The ones I have been adding these things support multiple interpreters hence my additions but if we don't want to deal with it then we can remove them. Either way is fine with me. We can also remove them completely if they are causing trouble for you @tylerjereddy and live with the default settings. 

The reason for the guards is that, if we don't put them 3.11 jobs were failing as they don't have these macros available.

tylerjereddy:
@AA-Turner do you want to take a stab at simplifying this now that we no longer have to support Python 3.11? We're planning to branch for the next release in a few days.

scipy/_lib/_fpumode.c
rgommers: I pushed a commit that addresses this. I kept `Py_MOD_PER_INTERPRETER_GIL_SUPPORTED`, because that's what we've been doing in other PRs, and there's no problem with that - so let's take the path of least resistance here.

rgommers (approved):
Resolved the last comment, this is good to go now. Thanks @AA-Turner & all reviewers.

PR #23074, opened 2025-05-28
DOC: special.softmax: document how to use a temperature parameter

jpmikhail:
#### Reference issue
N/A

This implements an additional parameter `beta` to the `softmax` function. 

#### Additional information

`beta` is an inverse temperature parameter that changes the weights assigned by `softmax`. The default value, 1, does not change the return value of `softmax`. Values greater than 1 correspond with lower temperature and thus place comparatively greater weight on greater values of the input array, `x`. Likewise, values less than 1 correspond with higher temperatures and thus place comparatively greater weight on lesser values of `x`.

### DIFF
--- a/scipy/special/_logsumexp.py
+++ b/scipy/special/_logsumexp.py
@@ -255,7 +255,7 @@ def _logsumexp(a, b, *, axis, return_sign, xp):
 
 
 @xp_capabilities()
-def softmax(x, axis=None, beta=None):
+def softmax(x, axis=None, beta=1.0):
     r"""Compute the softmax function.
 
     The softmax function transforms each element of a collection by
@@ -274,7 +274,7 @@ def softmax(x, axis=None, beta=None):
         computed over the entire array `x`.
     beta : float, optional
         Reciprocal temperature parameter for adjusting softmax weights.
-        Default is None and softmax will essentially use beta=1.
+        Default is 1.0 softmax weights will not be adjusted.
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


j-bowhay:
Looks like you have accidentally picked up some other commits alongside your changes

jpmikhail:
> Looks like you have accidentally picked up some other commits alongside your changes

Yep my bad, should be cleaned up now.

rgommers:
@jpmikhail do you have a justification for this addition? I quickly checked PyTorch and JAX, and they both don't have any such parameter:
- https://docs.pytorch.org/docs/main/generated/torch.nn.Softmax.html
- https://docs.jax.dev/en/latest/_autosummary/jax.nn.softmax.html

peterbell10:
Why do you think `softmax(x, beta=beta)` is better than `softmax(x * beta)`?

jpmikhail:
> Why do you think `softmax(x, beta=beta)` is better than `softmax(x * beta)`?

Functionally they're identical, of course. I think having the capability documented along with the function itself is cleaner and makes the use case more obvious. Users unfamiliar with the other contexts of softmax may not assume multiplying the input array by a scalar would be a well-established method of adjusting weights, or they may not know how to effectively communicate the meaning if they did.

Potentially, some additional documentation could be added instead of the parameter if people think that approach is better or more consistent with scipy practices.

lucascolley:
> I think having the capability documented along with the function itself is cleaner and makes the use case more obvious. Users unfamiliar with the other contexts of softmax may not assume multiplying the input array by a scalar would be a well-established method of adjusting weights, or they may not know how to effectively communicate the meaning if they did.
> 
> Potentially, some additional documentation could be added instead of the parameter if people think that approach is better or more consistent with scipy practices.

I think that sounds like a better idea. Would you like to rework the PR to add to the Notes section of the docstring, explaining the functionality?

PR #23081, opened 2025-05-29
DOC: Add tables summarizing array API backend capabilities across modules and functions

### LINKED ISSUE #23021: DEV: can we 'scrape' package-wide `make_xp_test_case` info?
I was wondering whether we could somewhat automate our tracking of support for alternative array types in this way.

I don't think we really need a docs page that covers every single function in detail, but it might be nice to have that information as e.g. a csv file, perhaps generate-able 'on demand', to inform tracking.

Is this feasible / worth working on? Any bright ideas? cc @crusaderky @j-bowhay @mdhaber

steppi:
#### Reference issue

Could potentially close #23021 

This PR is a proof of concept for adding a table to the documentation quantifying alternative array API backend support across modules. Each module  entry in the table links to another table which shows support across functions in that module. I did not try to format the tables nicely and hope for feedback on how best to present the tables. I wrote Sphinx directives to generate and insert the tables because I think this is a good solution for ensuring the tables are kept up to date.

Here's a screen shot of the top of the main table:

image

It's been inserted at the bottom of this page, https://docs.scipy.org/doc/scipy/dev/api-dev/array_api.html, on "Support for the array API standard". The entries in the table show the proportion of supported functions out of the total number of functions in the given modules public API.

Knowledge of what is and isn't supported comes from `scipy._lib._array_api.xp_capabilities_table` which is populated based on use of the `xp_capabilities` decorator. The proportions in the above table are currently lower than they should be due to functions with alternative backend support that still need the decorator. Before this current PR is merged, these missing decorators should be inserted; I can help work on that.

Clicking on a module in the above table will take you to an additional table which looks like this:

image

Again, this currently just a proof of concept, and this is not my proposal for the final form for such tables. 

Clicking on the links for individual functions will take you to the individual function's documentation pages. This lead me find three functions which seem to be in the public API because they are included in `__all__` from within a public module, but for which there are no generated docs.

```
<unknown>:69: WARNING: unknown document: '../../../reference/generated/scipy.ndimage.sum' [ref.doc]
<unknown>:38: WARNING: unknown document: '../../../reference/generated/scipy.optimize.line_search_wolfe2' [ref.doc]
<unknown>:39: WARNING: unknown document: '../../../reference/generated/scipy.signal.czt' [ref.doc]
```

I haven't looked into why this is occurring yet. 

Another thing that needs to be addressed: 

I haven't accounted for out-of-scope functions. I'd asked @rgommers about this, and my understanding is that there should be a formal policy to determine what should and shouldn't be considered out of scope, and that discussions and decision making around this are coming soon. In the table of functions per module, I think one could simply put NA for out of scope entries, and counts of out of scope entries out of scope functions can be excluded from the denominators in the proportions in the support per module table.  There can be a note before the table explaining this, perhaps with a link to the policy on what should be considered out-of-scope.

Edit:
There is an unrelated change here adding the `__name__` attribute to `MultiUFunc` callables. The reason I did this is because my logic for what parts of SciPy to include was named callables, and this made me realize that these were missing `__name__` within the `__all__` of a public module.


lucascolley:
> three functions which seem to be in the public API because they are included in __all__ from within a public module

my guess is that these are undocumented aliases for backwards-compatibility. At least I think that is the case for `ndimage.sum` and https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.sum_labels.html.

lucascolley (commented):
would you be able to also generate a table which is granular at the function-level which we could throw in the Wiki for reference?

steppi:
> my guess is that these are undocumented aliases for backwards-compatibility. At least I think that is the case for `ndimage.sum` and https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.sum_labels.html.

Ah, that makes sense. For `scipy.signal.czt` it's because of this

https://github.com/scipy/scipy/blob/7d936fa25e2e8fa74ec6399620dd6490e4dfb33e/doc/source/conf.py#L317-L326

It seems easy enough just to hardcode handling these special cases.

steppi:
For `scipy.optimize.line_search_wolfe2`, I discovered that `scipy.optimize.line_search` is an alias for `scipy.optimize.line_search_wolfe2` created with an `import as`

https://github.com/scipy/scipy/blob/7d936fa25e2e8fa74ec6399620dd6490e4dfb33e/scipy/optimize/_optimize.py#L36-L38

so `scipy.optimize.line_search.__name__ == 'line_search_wolfe2', and I was using `.__name__` for the names in the tables. I changed it to just use the relevant entry in `<module>.__all__` instead.

steppi:
OK, it looks like the failures are due to more aliases I missed. I guess I'll just add them all to the disallow list.

scipy/_lib/_array_api_docs_tables.py
```diff
@@ -0,0 +1,226 @@
...
+    multiple_modules = True
+    if isinstance(modules, str):
+        modules = [modules]
+        multiple_modules = False
+    if capabilities_table is None:
+        capabilities_table = xp_capabilities_table
+
+    output = []
+
+    from scipy.stats._distn_infrastructure import rv_generic
```
mdhaber, line 138: You mentioned this in your last message, but I think there needs to be another way. (Maybe even looking at the `__name__` attribute of `__thing__`'s class, if there's nothing else?)

steppi: Yeah, importing from `stats` to do this here is atrocious. I've gone back to skipping unnamed callables.

rgommers:
This looks pretty good, thanks @steppi. I think it's maintainable like this, so +1 on expanding the usage of `xp_capabilities`. Maybe focus on completeness of a couple of modules first. I think `cluster` is complete, `special` may be next?

For the per-module table, it looks like that won't fit very soon due to too many columns. I know you already said that this isn't the final form, so I'll just throw out a few high-level suggestions:
- Split the per-module table in three tables for: CPU, GPU, and JIT. That makes it easier to read, allows adding more libraries without running out of horizontal space on the rendered page, and also allows adding missing parts for existing libraries. E.g.:
    - PyTorch has a JIT as well - we are likely to want to add info on, and testing with, `torch.compile` in the future
    - Numba can fit in the JIT column
    - Adding `ndonnx`, MLX, etc.
- Try to reduce the number of characters per cell in all the tables. E.g.,:
    - `100/400 | 100/400` -> `25% | 25%` (add total number of functions once, as a last column or right behind the module name
    - `No | No | Yes` -> ` - |  - |  ✔️ `

steppi:
That all sounds good @rgommers. I started expanding `special` yesterday, building off of https://github.com/scipy/scipy/pull/22795.

steppi:
I've updated this to include three tables for CPU, GPU, and JIT as suggested by @rgommers. I've also incorporated the suggestions to reduce the number of characters per cell.

It's a little redundant how the directive needs to be included separately for each table, with the same lists of modules included in each, but I think this is good for now because of the flexibility it enables.

rgommers:
That looks pretty good to me!

To save others some clicking: [latest rendered version](https://output.circle-artifacts.com/output/job/7c88acf1-7cb2-4f0a-b53b-2b209c6601c3/artifacts/0/html/dev/api-dev/array_api.html#support)

lucascolley:
agreed, looks great!

j-bowhay:
It doesn't have to be in this PR but we need a way of marking functions that won't be converted. For example the datasets sub-module displays 0% coverage however these functions will never be converged since users can just call `xp.asarray` on their output. Likewise `scipy.constants.value` won't be converted as it doesn't operate on arrays.

steppi:
> It doesn't have to be in this PR but we need a way of marking functions that won't be converted. For example the datasets sub-module displays 0% coverage however these functions will never be converged since users can just call `xp.asarray` on their output. Likewise `scipy.constants.value` won't be converted as it doesn't operate on arrays.

Agreed, except I think it should be done in this PR. This is currently just a proof of concept,  and is also blocked until use of the `xp_capabilities` decorator becomes universal.

It's understandable you missed it, because my initial comment was a pretty long wall of text, but this is what I had written originally:

> I haven't accounted for out-of-scope functions. I'd asked @rgommers about this, and my understanding is that there should be a formal policy to determine what should and shouldn't be considered out of scope, and that discussions and decision making around this are coming soon. In the table of functions per module, I think one could simply put NA for out of scope entries, and counts of out of scope entries out of scope functions can be excluded from the denominators in the proportions in the support per module table. There can be a note before the table explaining this, perhaps with a link to the policy on what should be considered out-of-scope.

rgommers:
Agreed, it'd be useful to add it to this PR. No need to wait for decisions I think to add the infra for it, we're sure that we will have functions/objects and submodules that will be out of scope.

steppi:
I'm thinking that it may be best to use the `xp_capabilities` decorator to give the source of truth for which functions are out of scope. I'm curious about the desired level of granularity. If we decide that functions are either in-scope for alternative array API backend support or out of scope, then it would be straightforward to extend the existing decorator, but if there are functions which will be in-scope for some backends but not others (perhaps Jax or Dask) then I think we may want to reorganize the interface to `xp_capabilities`  to prevent it from exploding in complexity.

rgommers:
> if there are functions which will be in-scope for some backends but not others (perhaps Jax or Dask)

I don't think that there's such a distinction. It's either in scope or not for multiple array types. If the function cannot support JIT, GPU, or a specific array library for some reason specific to how the function is implemented, I think that it's then good to show as "not covered" rather than "out of scope".

steppi:
Declaring things out of scope with `xp_capabilities` is working now. For proof of concept, I've added the decorator to the relevant functions in `constants`. I used the below symbol for out-of-scope:

image

This is tentative. If we use the symbol instead of N/A then it seems like there should be a legend. Perhaps N/A is better.
I'll post the link to the deployed docs after they finish building if no one else gets to it first.

steppi:
OK, using "N/A" looks much better.

image

steppi:
docs link https://output.circle-artifacts.com/output/job/91a0013e-5321-4ccd-adee-e1ec1e2a2d39/artifacts/0/html/index.html

j-bowhay:
Thanks! Agree N/A looks good. I think it would be great to add a sentence or two explaining to users what the `N/A` means and why functions might be marked as out of scope

mdhaber:
Is there a better way to get to the table than from developer documentation -> SciPy API Development Guide -> Support for the array API standard? Maybe not for this PR, but it would be nice if the user part of that page were accessible from the tutorials (if it's not already).

lucascolley:
Hey @steppi, can I help move this along?

steppi:
> Hey @steppi, can I help move this along?

Yes, there are things you could do. We need to make `xp_capabilities` use universal in SciPy before this goes in. I've been working on adding `xp_capabilities` to special, carrying on from this PR, https://github.com/scipy/scipy/pull/22795, but there are still some hiccups to sort out. If you want to pick a submodule and work on adding the decorators, that would be helpful. I could review your PRs. You could also take a look at #22795 and review for me when I wrap it up, or maybe even give another look to the CI failures. I think some are just because I need to add PyTorch to the skips.

j-bowhay:
It might be useful to merge this and work on adding `xp_capabilities` incrementally` as thing like the out of scope marker added in this pr would be quite useful

steppi:
> It might be useful to merge this and work on adding `xp_capabilities` incrementally` as thing like the out of scope marker added in this pr would be quite useful

I'm open to that. And then when we add the decorators we can see right away how the tables were impacted. I marked it as ready for review.

requirements/doc.txt
```diff
@@ -13,3 +13,4 @@ pooch
 jupyterlite-sphinx>=0.19.1
 jupyterlite-pyodide-kernel
 linkify-it-py
+tabulate
```
j-bowhay, line 16: I think this also needs to be added to `pyproject.toml` and the `enviroment.yml

j-bowhay (commented):
Could you merge main as the CI artifacts have been cleared and to see if the additions #23287 etc show up correctly

requirements/doc.txt
lucascolley: Yes, please note https://github.com/scipy/scipy/blob/da62128ca614776e694098590b7c590b48f0933c/requirements/doc.txt#L1-L2

steppi: Thinking back, I think I did add to pyproject.toml and run `python tools/generate_requirements.py`, but may have lost those commits trying to clean up history.

steppi:
I'm OK with merging this now if anyone wants to push the button on it.

doc/source/dev/api-dev/array_api_modules_tables/differentiate.rst
```diff
@@ -0,0 +1,29 @@
+array API Support: differentiate
```
lucascolley: I don't like how this title looks:

Image

```suggestion
Array API Standard Support: ``differentiate``
```

please could we use full caps where appropriate for titles, lowercase "array" and "standard" elsewhere, and typically include "standard"? ref https://github.com/data-apis/array-api/discussions/778.

I think making module names monospace would also be a good change.

lucascolley (changes requested):
One point on style, otherwise this looks great, thanks Albert!

> Could you merge main as the CI artifacts have been cleared and to see if the additions https://github.com/scipy/scipy/pull/23287 etc show up correctly

looks like they have done 👍

> > would you be able to also generate a table which is granular at the function-level which we could throw in the Wiki for reference?
> 
> Yeah, that sounds doable. The tabulate package could be used to produce the tables in wiki format.

Are you able to follow-up on this now?

lucascolley:
>>>would you be able to also generate a table which is granular at the function-level which we could throw in the Wiki for reference?
>>
>>Yeah, that sounds doable. The tabulate package could be used to produce the tables in wiki format.
>
> Are you able to follow-up on this now?

Actually, since we have function-level granularity in the tables in the docs now, I don't think anything else is needed. My original thought was that we may not want to display this level of granularity in the docs, but I think it is fine now.

j-bowhay:
Also the coverage of sub-modules doesn't seem complete e.g. there is `spatial.distance` but no `spatial.transform` and a few other cases

steppi:
> Also the coverage of sub-modules doesn't seem complete e.g. there is `spatial.distance` but no `spatial.transform` and a few other cases

Currently the tables only include functions and other callables but there are only non-callable classes in the public API of `spatial.transforms`. I had intended to write a short paragraph before the tables in `array_api.rst` discussing what is and isn't out-of-scope. I was thinking that we could make it clear that only callables are considered, that `N/A` is for callables which are are out of scope, and sub-modules without functions in them are not included in the tables. We could also put everything in the public API in the tables, and have more things that get `N/A`. I don't really have a strong opinion on the matter. We may also want to add alternative backend support to non-callables; I was just starting out with callables only since 1) there's nothing like `xp_capabilities` for non-callables yet. 2) It's easier to get started this way.

lucascolley:
> there are only non-callable classes in the public API of spatial.transforms

https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Slerp.html#scipy.spatial.transform.Slerp has a `__call__` method, no?

steppi:
> > there are only non-callable classes in the public API of spatial.transforms
> 
> https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Slerp.html#scipy.spatial.transform.Slerp has a `__call__` method, no?

Oh, strange. I'll look into it and see if anything else is missing. Perhaps there's a bug in the script for putting together the tables.

steppi:
> > > there are only non-callable classes in the public API of spatial.transforms
> > 
> > 
> > https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Slerp.html#scipy.spatial.transform.Slerp has a `__call__` method, no?
> 
> Oh, strange. I'll look into it and see if anything else is missing. Perhaps there's a bug in the script for putting together the tables.

steppi:
> > > there are only non-callable classes in the public API of spatial.transforms
> > 
> > 
> > https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Slerp.html#scipy.spatial.transform.Slerp has a `__call__` method, no?
> 
> Oh, strange. I'll look into it and see if anything else is missing. Perhaps there's a bug in the script for putting together the tables.

OK, so the distinction here is between something that is `callable`, and a class which is a factory for generating `callable`s. The script I wrote to look for entries for the table only checks for `callables`, not classes with a `__call__` method which can be used to generate a `callable`.

https://github.com/scipy/scipy/blob/4d3c62958cd8e2410ada44cc295129db05ad787b/scipy/_lib/_array_api_docs_tables.py#L212

steppi:
Sorry, I didn't mean to close the PR and reopen. I was distracted by my son while typing my response. I'm trying to push a skip ci commit to stop ci, but the push is timing out for some reason.

steppi:
@lucascolley I fixed the headings. They were programmatically generated, and I hadn't really paid attention to what I put there. I agree they looked awful I had them.

The note describing what's in the tables may be a little verbose. I can cut it down if you think that would be better.

Update: Here's a link to the built docs: https://output.circle-artifacts.com/output/job/e82faba8-ecc1-4d66-9e4d-b8b6340e2dd6/artifacts/0/html/dev/api-dev/array_api.html#support. I forgot the preview links don't show up on this page like they used to.

steppi:
@lucascolley, I've made some further edits to the passage explaining the array API backend support tables. I wasn't very happy with what I wrote before.

doc/source/dev/api-dev/array_api_modules_tables/spatial
lucascolley, line 1: delete this file

doc/source/dev/api-dev/array_api_modules_tables/stats_contingency.rst
```diff
@@ -0,0 +1,30 @@
+Array API Standard Support: ``stats.contingency``
+=================================================
+
+
```
lucascolley: extra blank line here?

doc/source/dev/api-dev/array_api.rst
```diff
@@ -384,3 +384,158 @@ helped during the development phase:
 
 .. _RFC: https://github.com/scipy/scipy/issues/18286
 .. _the tracker issue: https://github.com/scipy/scipy/issues/18867
+
+
+Support
+-------
```
lucascolley: nit: I find this subheading uninformative, given the title of the top heading is "Support for the array API standard". How about something like "API coverage"?

doc/source/dev/api-dev/array_api.rst
```diff
@@ -384,3 +384,158 @@ helped during the development phase:
...
+The below tables show the current state of alternative backend support across
+SciPy's modules. Currently only public functions and function-like callable
+objects are included in the tables, but it is planned to eventually also include
+relevant public classes. Functions which are deemed out-of-scope are excluded
+from consideration. If a module or submodule contains no in-scope functions, it
+is excluded from the tables. For example, `scipy.spatial.transform` is currently
+excluded because it's API contains no functions, but may be included in the future
```
lucascolley, line 397:
```suggestion
excluded because its API contains no functions, but may be included in the future
```

doc/source/dev/api-dev/array_api.rst
```diff
@@ -384,3 +384,158 @@ helped during the development phase:
...
+
+As an example. The contents of `scipy.odr` are considered out-of-scope for a
+combination of reasons 2 and 3 above. `scipy.odr` essentially provides a direct
+wrapper of the monolithic ODRPACK Fortran library, and it's API is tied to the
+structure of this monolithic library. Creation of an efficient GPU accelerated
+implementation of nonlinear weighted orthogonal distance regression is also a
+challenging problem in its own right. Nevertheless, considerations of what to
+consider in-scope are evolving, and something which is now considered out-of-scope
+may be decided to be in-scope in the future if sufficient user interest and
+feasability are demonstrated.
```
lucascolley, line 417: thanks, this section reads very well!

doc/source/dev/api-dev/array_api.rst
```diff
@@ -384,3 +384,158 @@ helped during the development phase:
...
+
+
```
lucascolley, line 426: what do you think about adding a note that indication of a lack of support in the tables does not necessarily imply lack of support in the implementations, due to us being behind/fallible in adding `xp_capabilities()`? I'm not sure what level of detail is a good idea to include here w.r.t. that, but I think e.g. `fft` showing all 0% is a bit confusing.

lucascolley (changes requested):
thanks Albert, very close!

doc/source/dev/api-dev/array_api.rst
steppi: That's a good consideration. What do you think about using a third symbol (some kind of question mark maybe?) for things without `xp_capabilities`? We'd need to add a legend to explain the symbols too.

lucascolley: Yes, sounds great!

steppi:
@lucascolley, I've incorporated your suggestions. There is now a note explaining that the percentages may be below the true values, and I've added asterisks to percentages which may be impacted by missing `xp_capabilities` decorators. I've also added table legends to the per function tables. The legends look like this:

image

I've chosen to use blank cells for functions missing `xp_capabilities` decorators. I think it looks pretty good, but on reflection, I'm not happy with the the text `"not currently documented"`. It may not be clear enough that this means that the backend may or may not be supported, but whether it is supported is currently not documented. I thought `"unknown"` seemed a little odd too though. I'll try to think of something better. Also, maybe it's better not to have the bell icon in the legend. That's there because I used an `admonition` for the legend to put it in a box. There may be a better node type to use that won't add the bell. I think that doesn't really matter that much though.

lucascolley (approved):
> on reflection, I'm not happy with the the text "not currently documented". It may not be clear enough that this means that the backend may or may not be supported, but whether it is supported is currently not documented. I thought "unknown" seemed a little odd too though. I'll try to think of something better. Also, maybe it's better not to have the bell icon in the legend. That's there because I used an admonition for the legend to put it in a box. There may be a better node type to use that won't add the bell. I think that doesn't really matter that much though.

I'm happy for both of these things to be left for a follow-up!

---

thanks again!

steppi:
Thanks @lucascolley! In any case, it shouldn't take more than a week or two to add all the missing decorators for everything.

j-bowhay:
Thanks for pushing this through, unless I've missed something I think something has gone slightly wrong in the closing stages as lot of functions are showing up as unsupported when they are correctly marked https://scipy.github.io/devdocs/dev/api-dev/array_api_modules_tables/cluster_vq.html#array-api-support-cluster-vq-cpu

lucascolley:
Good spot, that needs to be fixed

PR #23082, opened 2025-05-30
TST: linalg: thread-safe rng

crusaderky:
- Make tests deterministic when run with `pytest-run-parallel`
- Fix infrequent random failure in `test_lapack` when running on a very high number of threads
- Review all instances of `pytest.mark.thread_unsafe`. Remove redundant ones (e.g. in case of `with pytest.warns`, which is now detected automatically)

Stress tested on a 32-core Linux machine with `--parallel-threads=32`.


scipy/linalg/tests/test_decomp.py
```diff
@@ -43,6 +43,7 @@
 
 def _random_hermitian_matrix(n, posdef=False, dtype=float):
     "Generate random sym/hermitian array of the given size n"
+    # FIXME non-deterministic rng
```
j-bowhay, line 46: I just wanted to double check that this is a future fix me and not one that was intended for this pr

j-bowhay (approved):
Looks like a nice modernization too! Just one minor query

scipy/linalg/tests/test_sketches.py
```diff
@@ -45,6 +45,7 @@ class TestClarksonWoodruffTransform:
 
     # Test vector with norm ~1
     x = rng.random((n_rows, 1)) / np.sqrt(n_rows)
+    del rng  # Not deterministic in pytest-run-parallel
```
rgommers, line 48: I think it's because this test is written in a weird way, normally this code would go into `setup_method`. Fine with me to leave as is though.

rgommers (approved):
LGTM too!

scipy/linalg/tests/test_decomp.py
crusaderky: yes, it's left for the future as fixing it would involve major refactoring. It's non-deterministic with the GIL too, so not terribly concerned.

scipy/linalg/tests/test_sketches.py
crusaderky: Yeah, the way it's written it's just working around not being able to have a `__init__` method. My addition just makes it clear that the rng should not be reused in the test methods.

Note that `setup_method` to my understanding would be problematic with `pytest-run-parallel`.

rgommers: > Note that `setup_method` to my understanding would be problematic with `pytest-run-parallel`.

It shouldn't be AFAIK, it's used extensively in the SciPy test suite. It should run just once, and since all it does is set up arrays/objects as attributes for read-only usage, it should be fine.

crusaderky: > It should run just once, and since all it does is set up arrays/objects as attributes for read-only usage, it should be fine.

I've been seeing it used a lot to set up a random state (like in many points that are cleaned up by this PR), a temporary directory, a temporary file handle, etc.

PR #23083, opened 2025-05-30
TST: linalg: pytest-run-parallell out of memory

crusaderky:
These two tests killed by 32-core, 64GiB RAM machine when run with `--parallel-threads=32`.


rgommers (approved):
Thanks @crusaderky

PR #23084, opened 2025-05-30
TST: linalg: race condition in BLAS `ger*`

### LINKED ISSUE #21936: BUG: `*ger*` routines in `linalg.blas` are not thread-safe
The `TestFBLAS2Simple.test_gemv` test is consistently segfaulting when run with [pytest-run-parallel](https://github.com/Quansight-Labs/pytest-run-parallel) under free-threaded CPython. The segfaults happen for:
- all of the `*ger`, `*geru`, and `*gerc` routines
- with all BLAS libraries - I tested OpenBLAS, MKL and Netlib on Linux as well as OpenBLAS and Accelerate on macOS

That makes it likely that the problem is in our `f2py` wrappers. They do contain something unusual, namely the line `a = <0.0,\0,(0.0\,0.0),\2,\2,\2>`, which is syntax I don't recognize:

https://github.com/scipy/scipy/blob/d249082f25b3555b9f8d283c97148253526d9cb7/scipy/linalg/fblas_l2.pyf.src#L210-L229

That parameter `a` is mutated in-place say [the docs](https://netlib.org/lapack/explore-html/d8/d75/group__ger_ga95baec6bb0a84393d7bc67212b566ab0.html), so that could well be the issue.

For the default CPython build, we didn't actually exercise multi-threaded calls since the GIL was serializing them. For free-threaded CPython we're actually calling the bindings in parallel. All other low-level functions in `linalg.blas` and `linalg.lapack` appear to be robust, except for this one.

crusaderky:
- Closes #21936
- Blocked by https://github.com/numpy/numpy/pull/29091

Tested many times on a 32-CPUs host with `--parallel-threads=32`.


rgommers:
Thanks, glad to see that fixed now. We should only merge this after NumPy 2.3.0 is out with the fix.

crusaderky:
NumPy 2.3.0 final is out and this should be now mergeable (if CI agrees)

rgommers (commented):
CI is happy, but that's expected since parallel testing is currently disabled. Some local testing shows that there's still a problem, this crashes regularly:
```
pixi r test-nogil -t scipy.linalg.tests.test_blas -- --parallel-threads=5 -k test_ger
```

crusaderky:
> CI is happy, but that's expected since parallel testing is currently disabled. Some local testing shows that there's still a problem, this crashes regularly:
> 
> ```
> pixi r test-nogil -t scipy.linalg.tests.test_blas -- --parallel-threads=5 -k test_ger
> ```

Can't reproduce on my box on 32 threads. Could you `rm -rf build-nogil build-nogil-install ../.pixi/envs/free-threading` and try again?

rgommers (approved):
I can no longer reproduce the issue after thoroughly cleaning build env and caches - not sure what happened there. In it goes, thanks @crusaderky!

PR #23086, opened 2025-05-30
CI: update windows-2019 runner image to windows-2022

abhishek-iitmadras:
As per info in : https://github.com/actions/runner-images/issues/12045
We need to change window runner.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -77,7 +77,7 @@ jobs:
         # https://github.com/github/feedback/discussions/7835#discussioncomment-1769026
         buildplat:
         # should also be able to do multi-archs on a single entry, e.g.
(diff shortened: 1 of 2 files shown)


rgommers (approved):
Thanks @abhishek-iitmadras.

This didn't run the wheel build jobs, however I just tested the wheel build jobs on my fork ([logs](https://github.com/rgommers/scipy/actions/runs/15353928501)) and there were no hiccups, so this should be good to go. 

Let me also link your NumPy PR: https://github.com/numpy/numpy/pull/28955. @tylerjereddy we should backport this to the 1.16.x branch straight away, to avoid potential issues with NumPy wheels having been built with a newer toolchain than the SciPy ones (xref https://github.com/numpy/numpy/pull/28955#issuecomment-2922880437).

PR #23089, opened 2025-05-31
ENH/DEP: linalg: support batching for toeplitz and related functions

j-bowhay:
Removes the deprecated raveling of inputs and adds batch support

Checks boxes in gh-21466.


scipy/linalg/_basic.py
```diff
@@ -2134,9 +2112,16 @@ def matmul_toeplitz(c_or_cr, x, check_finite=False, workers=None):
...
 
     from ..fft import fft, ifft, rfft, irfft
+    c, r = c_or_cr if isinstance(c_or_cr, tuple) else (c_or_cr, np.conjugate(c_or_cr))
+
+    return _matmul_toepltiz(r, c, x, workers, check_finite, fft, ifft, rfft, irfft)
+
 
-    r, c, x, dtype, x_shape = _validate_args_for_toeplitz_ops(
-        c_or_cr, x, check_finite, keep_b_shape=False, enforce_square=False)
+@_apply_over_batch(('c', 1), ('r', 1), ('x', '1|2'))
```
mdhaber: I don't remember if it matters, but
```suggestion
@_apply_over_batch(('r', 1), ('c', 1), ('x', '1|2'))
```
right?

mdhaber (approved):
I think it looks good. Just the one question/comment.

scipy/linalg/_basic.py
j-bowhay: Practically I don't think it makes a difference, but yes let's make it correct

ilayn (approved):

PR #23091, opened 2025-05-31
BUG: sparse: correct eq and ne with different shapes. (and add test)

### LINKED ISSUE #23068: BUG: sparse: `!=` operator with csr matrices
### Describe your issue.

This behavior could be elsewhere but I have definitely found it here where `!=` does not work as expected in the release candidate, but did previously work

### Reproducing Code Example

```python
from scipy import sparse
assert sparse.random(10,10,format="csr") != sparse.random(10, 20, format="csr")
```

### Error message

```shell
  File "<python-input-0>", line 2, in <module>
    assert sparse.random(10,10,format="csr") != sparse.random(10, 20, format="csr")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

```shell
1.16.0rc1 2.3.0rc1 sys.version_info(major=3, minor=13, micro=0, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    found: true
    include directory: unknown
    lib directory: unknown
    name: Accelerate
    pc file directory: unknown
    version: unknown
  lapack:
    found: true
    include directory: unknown
    lib directory: unknown
    name: Accelerate
    pc file directory: unknown
    version: unknown
  pybind11:
    detection method: config-tool
    include directory: unknown
    name: pybind11
    version: 2.13.6
Compilers:
  c:
    commands: cc
    linker: ld64
    name: clang
    version: 15.0.0
  c++:
    commands: c++
    linker: ld64
    name: clang
    version: 15.0.0
  cython:
    commands: cython
    linker: cython
    name: cython
    version: 3.1.1
  fortran:
    commands: gfortran
    linker: ld64
    name: gcc
    version: 13.3.0
  pythran:
    include directory: ../../../../../../private/var/folders/w5/_8wgjw3j5cg6mgrth3s2kg9m0000gn/T/pip-build-env-og4zq41h/overlay/lib/python3.13/site-packages/pythran
    version: 0.17.0
Machine Information:
  build:
    cpu: aarch64
    endian: little
    family: aarch64
    system: darwin
  cross-compiled: false
  host:
    cpu: aarch64
    endian: little
    family: aarch64
    system: darwin
Python Information:
  path: /private/var/folders/w5/_8wgjw3j5cg6mgrth3s2kg9m0000gn/T/cibw-run-fkc4ao5g/cp313-macosx_arm64/build/venv/bin/python
  version: '3.13'
```

dschult:
Fixes #23068 

Operators != and == return incorrect results for differently shaped sparse arrays/matrices. This was introduced in #22897 and has been released in 1.16rc.  The result of differently shaped boolean eq/ne was not tested and fell through the cracks.

This PR adds a test and fixes the bug.

There is also a bigger question here: 
Why doesn't sparse raise when the shapes don't match. 
Numpy raises a ValueError because the sizes do not broadcast to the same shape. 

We should at least change sparray treatment of this up-to-now untested case.  If/when we support broadcasting the results will differ.    But that's for a separate PR.  This is a bug-fix.

### DIFF
--- a/scipy/sparse/_base.py
+++ b/scipy/sparse/_base.py
@@ -658,7 +658,7 @@ class _spbase(SparseABC):
                 # eq and ne return True or False instead of an array when the shapes
                 # don't match. Numpy doesn't do this. Is this what we want?
                 if op in (operator.eq, operator.ne):
-                    return op == operator.eq
+                    return op is operator.ne
                 raise ValueError("inconsistent shape")
 
             csr_self = (self if self.ndim < 3 else self.reshape(1, -1)).tocsr()
--- a/scipy/sparse/tests/test_base.py
+++ b/scipy/sparse/tests/test_base.py
@@ -433,6 +433,13 @@ class _TestCommon:
         for dtype in self.checked_dtypes:
             check(dtype)
 
+    def test_eq_ne_different_shapes(self):
+        if self.datsp.format not in ['bsr', 'csc', 'csr']:
+            pytest.skip("Bool comparisons only implemented for BSR, CSC, and CSR.")
+        # Is this what we want? numpy raises when shape differs. we return False.
+        assert (self.datsp == self.datsp.T) is False
+        assert (self.datsp != self.datsp.T) is True
+
     def test_lt(self):
         sup = suppress_warnings()
         sup.filter(SparseEfficiencyWarning)


dschult:
@perimosocordiae @stefanv @j-bowhay  
Any chance we can review this before 1.16rc2 on Wednesday?

It's a simple bug. I compared to the wrong operator. This also adds a test.

stefanv (approved):

stefanv:
I'm going to merge this bugfix. If other reviews disagree, we'll go from there.

perimosocordiae:
Sorry for the late review! I agree that this is weird behavior. Thanks for the bug fix and regression test.

I'm in favor of aligning sparse array behavior with numpy.

stefanv:
Agreed.

PR #23092, opened 2025-06-01
DEV: Address issues in `spin docs` command

czgdp1807:
#### Reference issue

Addresses https://github.com/scipy/scipy/pull/23041#discussion_r2105773841

There are multiple issues with the `docs` command, in SciPy and also in NumPy:

- For both: docstring looks bad, needs improving:
    - [x] `SPHINXOPTS` should be removed since it's both too niche and incorrect,
    - [x] It should explain what targets are supported and how to build exactly those - `html`, `dist` only for SciPy, and in addition `latex` for NumPy (better renamed to `pdf` probably)
    - [x] `spin docs targets` should not be mentioned, that's just confusing for no good reason.
- [x] For both: `sphinx-gallery` is not used and hence should not be an option, that is confusing
- [x] For SciPy: docstring is displayed twice (see below)
- [x] For SciPy: `--clean` doesn't work indeed, deferring to `spin` should work
- [x] For SciPy: `--no-cache` doesn't seem useful, it says it forces a full rebuild, so seems duplicate with `--clean` at that point.
- [x] For SciPy: `--list-targets` can then also be removed

_Originally posted by @rgommers in https://github.com/scipy/scipy/pull/23041#discussion_r2105773841_

### DIFF
--- a/.spin/cmds.py
+++ b/.spin/cmds.py
@@ -262,59 +262,26 @@ def test(*, parent_callback, pytest_args, tests, coverage,
     parent_callback(**{"pytest_args": pytest_args, "tests": tests,
                     "coverage": coverage, **kwargs})
 
-@click.option(
-        '--list-targets', '-t', default=False, is_flag=True,
-        help='List doc targets',
-    )
-@click.option(
-    '--no-cache', default=False, is_flag=True,
-    help="Forces a full rebuild of the docs. Note that this may be " + \
-            "needed in order to make docstring changes in C/Cython files " + \
-            "show up."
-)
-@spin.util.extend_command(spin.cmds.meson.docs)
-def docs(*, parent_callback, sphinx_target, clean, jobs,
-         list_targets, no_cache, **kwargs):
+@spin.util.extend_command(spin.cmds.meson.docs, remove_args=("sphinx_gallery_plot", ))
+def docs(*, parent_callback, sphinx_target, clean, jobs, **kwargs):
     """📖 Build Sphinx documentation
 
-    By default, SPHINXOPTS="-W", raising errors on warnings.
-    To build without raising on warnings:
+    Following Sphinx targets are supported:
 
-      SPHINXOPTS="" spin docs
+    html:
 
-    To list all Sphinx targets:
+      spin docs html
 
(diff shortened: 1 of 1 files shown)


rgommers:
Thanks @czgdp1807. xref my comment on the `doc` keyword to `extend_command`: https://github.com/scientific-python/spin/pull/282#issuecomment-2927104742

Once the `docs` command is fixed, it would be good to also address the docstring of the `test` command and ensure the `--help` output of the other commands also looks good.

czgdp1807:
> Once the `docs` command is fixed, it would be good to also address the docstring of the `test` command and ensure the `--help` output of the other commands also looks good.

Addressed all. Works well now. Check below,

```zsh
Usage: spin build [OPTIONS] [MESON_ARGS]...

  🔧 Build package with Meson/ninja and install

  MESON_ARGS are passed through e.g.:

  spin build -- -Dpkg_config_path=/lib64/pkgconfig

  The package is installed to build-install

  By default builds for release, to be able to use a debugger set CFLAGS appropriately. For example, for linux use

  CFLAGS="-O0 -g" spin build

Options:
```

```zsh
Usage: spin test [OPTIONS] [PYTEST_ARGS]...

  🔧 Run tests

  PYTEST_ARGS are passed through directly to pytest, e.g.:

    spin test -- --pdb

  To run tests on a directory or file:

    spin test scipy/linalg

  To report the durations of the N slowest tests:

    spin test -- --durations=N

  To run tests that match a given pattern:

    spin test -- -k "geometric"
    spin test -- -k "geometric and not rgeometric"

  By default, spin will run `-m 'not slow'`. To run the full test suite, use `spin test -m full`

  For more, see `pytest --help`.

Options:
```

```zsh
(scipy-dev) 17:35:28:~/Quansight/scipy % spin docs --help
Usage: spin docs [OPTIONS] [SPHINX_TARGET]

  📖 Build Sphinx documentation

  Following Sphinx targets are supported:

  html:

    spin docs html

  dist: to build a zipfile of the html docs for distribution

    spin docs dist

Options:
```

rgommers (approved):
LGTM now, thanks @czgdp1807!

PR #23094, opened 2025-06-01
DOC: linalg: fix overview for `cdf2rdf` and `khatri_rao`

roryyorke:
In both cases took overview from `linalg.__init__.py`.

[docs only]

#### Reference issue
None; searching open issues for `khatri_rao` and `cdf2rdf` yields no hits.

Fixes minor errors in linalg overview for the two functions.


ev-br (approved):
Both updates are certainly improvements, thanks @roryyorke !

Reading the `khatri_rao` docstring still leaves me puzzled and reaching for wikipedia. Not sure how to improve it without an example, and wikipedia has it. So if anything to be done for it, it can be done in a follow-up, and this PR LGTM.

PR #23096, opened 2025-06-01
CI: move to `cibuildwheel` 3.0b4 and to `manylinux_2_28`

rgommers:
These changes were made in NumPy already, the `manylinux_2_28` one a few months ago (will be in 2.3.0) and the `cibuildwheel` one more recently (reason: a release is imminent, and this is needed for Python 3.14 support).

We need to backport the `cibuildwheel` change to the 1.16.x branch within 2-3 months when we do a release to add `cp314` wheels, but I wouldn't do that now since it's not a final release yet (even though it should be stable).

The changes to `pyproject.toml` also match those in NumPy; the other changes are just for explicitness - `musllinux_1_2` was already the default, and we were already not building `universal2` wheels. The `cp36-* cp37-* pp*` are no longer built by default, so the skips can be removed (leaving them in would cause a warning actually).

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -212,7 +212,7 @@ jobs:
           echo "CIBW_BUILD_FRONTEND=$CIBW" >> "$GITHUB_ENV"
 
       - name: Build wheels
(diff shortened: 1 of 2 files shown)


tylerjereddy (approved):
Thanks Ralf, sounds plausible to me. Diff looks clean and CI is passing with wheel builds activated.

PR #23103, opened 2025-06-03
ENH: stats: Dask support for `differential_entropy(method='ebrahimi')`

crusaderky:
Also a prerequisite for `jax.jit` (untested)

### DIFF
--- a/scipy/stats/_entropy.py
+++ b/scipy/stats/_entropy.py
@@ -10,7 +10,6 @@ from scipy import special
 from ._axis_nan_policy import _axis_nan_policy_factory
 from scipy._lib._array_api import (array_namespace, xp_promote, xp_device,
                                    is_marray, _share_masks)
-from scipy._lib import array_api_extra as xpx
 
 __all__ = ['entropy', 'differential_entropy']
 
@@ -409,8 +408,7 @@ def _ebrahimi_entropy(X, m, *, xp):
 
     i = xp.arange(1, n+1, dtype=X.dtype, device=xp_device(X))
     ci = xp.where(i <= m, 1 + (i - 1)/m, 2.)
-    cond = i >= n - m + 1
-    ci = xpx.at(ci, cond).set(1 + (n - i[cond])/m)
+    ci = xp.where(i >= n - m + 1, 1 + (n - i)/m, ci)
 
     logs = xp.log(n * differences / (ci * m))
     return xp.mean(logs, axis=-1)
--- a/scipy/stats/tests/test_device_dtype.py
+++ b/scipy/stats/tests/test_device_dtype.py
@@ -5,6 +5,7 @@ from scipy import stats
 from scipy._lib._array_api import xp_device, is_array_api_strict, is_dask, is_torch
 from scipy.stats._stats_py import _xp_mean, _xp_var
 
+skip_xp_backends = pytest.mark.skip_xp_backends
 
 dtypes = ['float32', 'float64']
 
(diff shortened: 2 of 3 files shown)


scipy/stats/tests/test_device_dtype.py
```diff
@@ -68,14 +69,21 @@ def test_directional_stats(dtype, xp, devices):
...
+    pytest.param(
+        "correa",
+        marks=[
+            skip_xp_backends("array_api_strict", reason="Invalid fancy indexing"),
+            skip_xp_backends("dask.array", reason="Invalid fancy indexing"),
+        ],
+    ),
+    "ebrahimi",
+    "auto",
+])
```
mdhaber, line 84: @rgommers In https://github.com/scipy/scipy/pull/22856#discussion_r2117586032 it sounded like you were OK with either syntax for skipping tests. Here, `parametrize` is less concise. Is this change necessary?

rgommers: Probably not _necessary_. That said, there may be two advantages:
- It may reduce the total number of tests being configured/run (since the skip happens at the test discovery phase)
- The balance on conciseness may shift as more skips are added, like for `jax.jit` (as the PR description says)

So I can see arguments either way.

mdhaber: I'm fine with the change here. I would just like to clarify that both syntaxes may be acceptable. I will probably continue to use `pytest.skip` when it is not obvious that configuring marks with `pytest.param` would have significant advantages.

scipy/stats/_entropy.py
```diff
@@ -409,8 +408,7 @@ def _ebrahimi_entropy(X, m, *, xp):
 
     i = xp.arange(1, n+1, dtype=X.dtype, device=xp_device(X))
     ci = xp.where(i <= m, 1 + (i - 1)/m, 2.)
-    cond = i >= n - m + 1
-    ci = xpx.at(ci, cond).set(1 + (n - i[cond])/m)
+    ci = xp.where(i >= n - m + 1, 1 + (n - i)/m, ci)
```
mdhaber, line 411: @crusaderky Why does this change fix the issue?

crusaderky: Because `xpx.at(x)[mask].set(y[mask])` never works on Dask or in `jax.jit` due to boolean masking.

mdhaber: My mental model of `at.set` is that it effectively does `where` when necessary and `__setitem__` otherwise. That was the point of using `at.set`, I thought - we use it so we don't have to choose between mutation for mutable arrays and `where` for the rest when either is acceptable.

What is wrong with that model? 

Why can't/shouldn't we make `at.set` behave that way so that this change would not be necessary?

mdhaber (approved):
This seems to work as advertised, and I like the consolidation of `methods` in `test_entropy.py`.

Most of the changes in `test_device_dtype` seems like personal refactoring preferences orthogonal to the main content of this, so I would have prepared for them to be a separate PR. But they are here and they don't look problematic, so in they go.

I'd still appreciate a reply to https://github.com/scipy/scipy/pull/23103/files#r2124465085, though. Thanks!

scipy/stats/_entropy.py
crusaderky: array-api-extra can't prevent the Python interpreter from executing `y[mask]`. You never get to call `at`; JAX has already raised an exception. On Dask, one would need to crawl back up the graph (no can do; not outside of the dask package) to undo `y[mask]` and retrieve the original `y`.

mdhaber: IIUC, the problem here is `i[cond]`, we can't (ever) use boolean mask indexing, and `at.set` can only really help us for integer indexing replacements. Is that right?

crusaderky: Correct, `i[cond]` fails inside jax.jit.

PR #23105, opened 2025-06-03
ENH: stats: Dask and JAX support for `pearsonr`

crusaderky:


scipy/stats/_stats_py.py
```diff
@@ -4254,6 +4255,7 @@ def _pearsonr_fisher_ci(r, n, confidence_level, alternative):
         rhi = ones
 
     mask = (n <= 3)
+    mask, rlo, rhi = xp.broadcast_arrays(mask, rlo, rhi)
```
mdhaber: It looks like some of the failures in [full, py3.11/npMin, spin](https://github.com/scipy/scipy/actions/runs/15413780597/job/43371686407?pr=23105#logs) can be addressed here.

scipy/stats/_stats_py.py
```diff
@@ -4780,8 +4794,11 @@ def statistic(x, y, axis):
     pvalue = _get_pvalue(r, dist, alternative, xp=xp)
 
     mask = (n == 2)   #  return exactly 1.0 or -1.0 values for n == 2 case as promised
-    def special_case(r): return xp.where(xp.isnan(r), xp.nan, xp.ones_like(r))
-    r = xpx.apply_where(mask, (r,), xp.round, fill_value=r)
+    # data-apis/array-api-extra#196
+    mxp = array_namespace(r._meta) if is_dask(xp) else xp    
+    def special_case(r):
+        return mxp.where(mxp.isnan(r), mxp.nan, mxp.ones_like(r))
```
mdhaber, line 4795: To check my understanding, an alternative fix would be for `special_case` to detect the `xp` of `r`?
```suggestion
    def special_case(r):
        xp = array_namespace(r)
        return mxp.where(xp.isnan(r), xp.nan, xp.ones_like(r))
```

scipy/stats/_stats_py.py
```diff
@@ -4780,8 +4794,11 @@ def statistic(x, y, axis):
...
+    r = xpx.apply_where(mask, (r,), mxp.round, fill_value=r)
```
mdhaber: IIRC, this should work now, too, right?
```suggestion
    r = xpx.apply_where(mask, r, mxp.round, fill_value=r)
```

scipy/stats/_stats_py.py
```diff
@@ -4254,6 +4255,15 @@ def _pearsonr_fisher_ci(r, n, confidence_level, alternative):
...
 
     mask = (n <= 3)
+
+    # Slightly convoluted logic to
+    # 1. suppress NumPy warning for reading flags.writeable after np.broadcast_arrays
+    # 2. let xpx.at overwrite rlo and rhi if no broadcasting is needed
+    shape = xpx.broadcast_shapes(mask.shape, rlo.shape, rhi.shape)
+    mask = xp.broadcast_to(mask, shape)
+    rlo = xp.broadcast_to(rlo, shape) if shape != rlo.shape else rlo
+    rhi = xp.broadcast_to(rhi, shape) if shape != rhi.shape else rhi
```
mdhaber: I haven't thought too carefully yet, but if you have, can you explain why broadcasting is only necessary to make Dask/JAX work - or was it always needed? 

I'm also a litle unclear on how `broadcast_to` instead of `broadcast_shapes` fixes the problem.

scipy/stats/_stats_py.py
crusaderky: Yes, it would fix `special_case`. However you also need `mxp.round` one line later.

scipy/stats/_stats_py.py
crusaderky: Broadcasting is necessary for Dask because in `da.Array.__setitem__` boolean indices are exclusively accepted if they have exactly the same shape as the array being indexed, instead of just being broadcastable (as the Array API allows for).
Admittedly, this could have been fixed in `xpx.at` instead.

`broadcast_arrays` works to fix Dask; however it breaks NumPy:
- it generates arrays where the first access to `ndarray.flags.writeable` will issue a warning, and
- it incorrectly marks arrays that do need to be broadcasted as writeable (with warning), and
- it will in the future unnecessarily marks arrays that don't need to be broadcasted as read-only

mdhaber: OK, I think I understand. Please check me:

Previously, this function was taking advantage of indexing rules that the array API standard does not guarantee. You fixed this with explicit broadcasting.

The problem with `broadcast_arrays` is that it currently doesn't set `writeable` properly - but `broadcast_to` does. `broadcast_arrays` will work properly in the future. However, another issue with the current code is that when the arrays are already of the same size, `broadcast_arrays` and `broadcast_to` both unncessarily mark arrays as read-only, hence the ternary. In the future (when the warning goes away and NumPy `broadcast_arrays` is fixed), would something like this be OK?
```suggestion
    if not (rlo.shape == rhi.shape == mask.shape == shape):
        mask, rlo, rhi == xp.broadcast_arrays(mask, rlo, rhi)
```

`at.set` looks at the `writeable` flag for NumPy arrrays, and if they are not writeable, it does something `where`-like rather than `__setitem__`-like, right?

scipy/stats/_stats_py.py
mdhaber: I see.

I think I'm a little confused by the terminology, though. I think `m` in `mxp` refers to `meta`, since it comes from `r._meta`. But take the case of Dask with NumPy as the underlying library - if we are to use the term "meta" to refer to one of the libraries, I would expect it to refer to the higher-level library, Dask, not NumPy. But the callable passed to `apply_where` needs to work on arrays from underlying library (NumPy). Is the use of "meta" reversed compared to what I would expect?

mdhaber (approved):
Looks good, but I'd prefer to wait for an answers to https://github.com/scipy/scipy/pull/23105#discussion_r2124514847 before merging; it might be worth commenting in the code about just replacing with `broadcast_arrays` in the future, if that's an option.

I'd also appreciate a reply to https://github.com/scipy/scipy/pull/23105#discussion_r2124532107, but that is not blocking.

scipy/stats/_stats_py.py
crusaderky: > In the future (when the warning goes away and NumPy `broadcast_arrays` is fixed), would something like this be OK?
> ```python
> if not (rlo.shape == rhi.shape == mask.shape == shape):
>         mask, rlo, rhi == xp.broadcast_arrays(mask, rlo, rhi)
> ```
Correct.

> at.set looks at the writeable flag for NumPy arrrays, and if they are not writeable, it does something where-like rather than __setitem__-like, right?

No, it copies the array to a writeable one and then calls `__setitem__` on it.
This is because there is no way to express all the indexing rules in a where.

crusaderky: I've looked into this more - the Array API definition does not allow for the index to be broadcastable to the indexed array. Instead, it supports some special cases where the index has less dimensions than the indexed array: https://data-apis.org/array-api/latest/API_specification/indexing.html#boolean-array-indexing which as a general rule behave differently from broadcasting.

pearsonr triggers an edge case where the index has _exactly_ 0 dimensions, which is not supported by dask, and thus it _looks like_ broadcasting it works but it's just coincidental that broadcasting and bool index rules are the same in this case.

I've simplified the code to reflect this.

scipy/stats/_stats_py.py
crusaderky: Dask has used the word "meta" for almost a decade to indicate the wrapped namespace.
I would advise against overloading it now to mean the opposite in some cases.
"wrapper" and "wrapped" are more explicit, but unfortunately don't abbreviate easily.

mdhaber: I did a little more research, and now understand that `_meta` is used in the sense of "metadata". I don't think Dask is referring to NumPy as the "meta" namespace. The `_meta` attribute stores metadata about what a chunk will look like, and it happens to be an instance of the underlying library's array type, so we are able to use `array_namespace` on it to get the underlying namespace.

scipy/stats/_stats_py.py
mdhaber: > No, it copies the array to a writeable one and then calls `__setitem__` on it.

OK, so not `where`, but all I meant by `where`-like is that it creates a copy with some elements changed. It happens to implement this by copying and using `__setitem__`.

mdhaber (approved):

PR #23108, opened 2025-06-03
MAINT: Remove optional arguments from `*_impl` functions in  `_fitpack_repro.py`

czgdp1807:
#### Reference issue

Reported in https://github.com/czgdp1807/scipy/pull/19#discussion_r2104438220

The above review comment reports that `xb=None` and `xe=None` are filled by default values in `_validate_inputs` before calling `_generate_knots_impl`. I think all `*_impl` functions don't need any optional arguments with default values because the user facing APIs and `_validate_inputs` fills them up with correct values. So removed optional arguments from `*_impl` functions. This reduces confusion and addresses https://github.com/czgdp1807/scipy/pull/19#discussion_r2104438220 in a broader context.

#### Additional information

@ev-br


ev-br (approved):
LGTM, thanks @czgdp1807 !

Issue #21569, opened 2024-09-17
RFC: spatial.distance: a roadmap for Boolean distance metrics


fancidev:
### Is your feature request related to a problem? Please describe.

Upon reviewing the Boolean distance metrics in scipy.spatial.distance, I find a number of inconsistencies that I’d like to collect below and address them in steps.

### Describe the solution you'd like.

The following distance functions are referred to as *Boolean metrics* in the [documentation](https://docs.scipy.org/doc/scipy/reference/spatial.distance.html):

dice, hamming, jaccard, kulczynski1, rogerstanimoto, russelrao, sokalmichener, sokalsneath, yule 

The functions vary in their handling of non-Boolean input, fuzzy Boolean input, and zero input. The Python implementation may behave differently from the C++ (pdist/cdist) implementation. Some functions are defined in a questionable way (in my opinion). These are summarized below.

**Problem with  kulczynski1**. All metrics, are _documented_ to be dissimilarity metrics, except that kulczynski1 is _implemented_ as a similarity metric. This is clearly inconsistent. I propose to deprecate and remove kulczynski1.

- [ ] gh-21187

**Problem with hamming**. Hamming is documented to operate on Boolean vectors, but it is often treated as an “edit distance”. This seems to be an unfortunate consequence of “merging” matching with hamming, which are fundamentally different metrics that happen to share the same implementation when the inputs are pure Boolean vectors.

- [ ] (issue to be opened)

**Handling of zero input vector**. When both input vectors are zero, the formula of certain metrics involve …

**Handling of non-Boolean input**.

**Handling of fuzzy Boolean input**.

**General documentation**. The documentation of some metrics lack a reference or source of definition. This has created trouble in the past of identifying the “definitive” definition of the metric. Separately, weighted metric or fuzzy input are not documented. This should be enriched as we go.

tylerjereddy:
I feel like I'm going to be reading more LaTeX math in the near future :)

Issue #21135, opened 2024-07-08
DOC: spatial: document 0/0 handling of Boolean metrics


fancidev:
### Issue with current documentation:

In `spatial.distance`, some of the distance metrics defined for Boolean vectors are computed as a ratio `num/denom`, where `denom` could become zero. The handling of such situation is diverse between metrics and sometimes between the Python and C++ implemention of the same metric, and is only sparsely documented. It could be helpful to systematically document the behavior if only for completeness.

### Idea or request for content:

The following table summarizes the handling of 0/0 for the Boolean metrics by the Python implementation and the C++ implementation, as well as whether documentation is available.

(Table to be completed)

### Additional context (e.g. screenshots, GIFs)

Whether to reconcile the differences between metrics or between implementations of the same metric is out of the scope of this issue and is left for future discussion.

fancidev:
The Boolean metrics may be classified into a few groups where metrics in the same group are monotone transforms of each other (and in this sense “equivalent”).

For ease of notation, let $a=n_\textrm{TT}$, $b=n_\textrm{TF}$, $c=n_\textrm{FT}$, $d=n_\textrm{FF}$.

Group 1. Simple matching, and friends.

*Remark*. The simple matching distance is (unfortunately) known as hamming distance in spatial.distance.

$$
\begin{align*}
d_\textrm{matching} &:= \frac{b+c}{(a+d)+(b+c)} \\
\end{align*}
$$

Group 2. Jaccard, 

Group 3. Yule.

$$
d_\textrm{yule} := \frac{2bc}{ad+bc}
$$

Group 4. Russels-Rao.

$$
d_\textrm{russelsrao} := \frac{b+c+d}{a+b+c+d}
$$

fancidev:
While out of the scope of this PR, it would be interesting to see how the other (real) metrics defined in `scipy.spatial.distance` handle empty vector inputs. This is reported below.

<details>

```python
import scipy.spatial.distance as sd

metrics = ["braycurtis", "canberra", "chebyshev", "cityblock",
           "correlation", "cosine", "euclidean", "jensenshannon",
           "mahalanobis", "minkowski", "seuclidean", "sokalmichener",
           "sqeuclidean"]

for metric in metrics:
    u = v = []
    try:
        v1 = getattr(sd, metric)(u, v)
    except Exception as e:
        v1 = type(e).__name__
    v2 = sd.pdist([u, v], metric)
    v3 = sd.cdist([u], [v], metric)
[...]

Issue #21090, opened 2024-07-01
RFC: spatial.distance: consolidate Cython and Python metrics 


fancidev:
### Is your feature request related to a problem? Please describe.

The spatial.distance module contains two sets of implementations for the 20+ metrics: the “one-per-call” Python version, such as `euclidean`, and the “many-per-call” Cython version, such as `cdist(‘euclidean’)`. Recently I come across three PRs where such “duality” brings some inconvenience:

- gh-20786 improves the numerical accuracy of jensenshannon, but it has to change both Python and Cython because the two implementations are separate.
- gh-21002 improves checks on dtype promotion for the Python implementation, but the impact/consistency with the Cython version is unclear.
- gh-21085 deprecates complex input to the Python version of cosine and correlation, but the remaining Python metrics remain unspecified in terms of complex handling, even though their Cython counterparts only support real input.

These examples show that having two sets of implementations leads to inconsistency and adds complexity to maintenance.

### Describe the solution you'd like.

Consolidate the implementation of the Python and Cython versions of the 20+ metrics. Ideally, for each metric, there is only one place that provides the underlying formula.

There need be no (backward-incompatible) change to the interface; only the implementation has to be updated.

### Additional context (e.g. screenshots, GIFs)

I’d like to hear your opinion on whether this objective is desirable and feasible.

I’d also very much like to learn if you know about any historical reason that leads to the current situation (of having two separate implementations of metrics).

fancidev:
A quick search of “spatial.distance” under Issues brings up more interesting and relevant discussions, such as gh-13629. It appears continuous improvements have been made to this module but perhaps more is necessary to make it into a fully good shape.

It seems to me the most promising way to get this done once and for all is to vectorize the Python metric functions one by one, and finally update cdist/pdist to dispatch to the Python versions and drop the Cython versions altogether.

A bonus of this approach is that we get Array API compatibility almost automatically. The main risk of this approach is degradation of performance due to Python overhead, especially for tiny inputs. But I could pick three metrics to try this out if there’s some support for doing so.

tylerjereddy:
> A bonus of this approach is that we get Array API compatibility almost automatically. 

I don't follow--most of the performance-critical distance code for i.e., `cdist` was rewritten in C++/pybind11 by Peter et al. How does unifying the "pure Python" implementations to that help with array API compat? 

[...]

PR #17538, opened 2022-12-04
ENH: Extending _distance_pybind with additional distance metrics (continued)

jjerphan:
#### Reference issue

Follow-up of #14640, authored by @czgdp1807.

Alternative to #17522 whose CI is failing continuously.

> #### Reference issue
> 
> #14077
>
> #### What does this implement/fix?
> 
> In this PR following metrics will be ported to `_distance_pybind11`,
>  * [x]  dice
>  * [x]  hamming
>  * [x]  jaccard
>  * [x]  kulsinski
>  * [x]  matching
>  * [x]  rogerstanimoto
>  * [x]  russellrao
>  * [x]  sokalmichener
>  * [x]  sokalsneath
>  * [x]  yule
> 
> 
> The remaining distance metrics (tricky and potential blocker ones) are being dealt with in #14611
> 
> Please feel free to leave your reviews.
> #### Additional information
> 
> Benchmark results.
> 
> ```python
>        before           after         ratio
>      [453aad79]       [24897be1]
>      <master>         <pybind11_dist_framework>
> +     1.59±0.02ms      1.98±0.04ms     1.24  spatial.Xdist.time_pdist(1000, 'dice')
> +     3.20±0.01ms      3.89±0.02ms     1.21  spatial.Xdist.time_cdist(1000, 'dice')
> +     31.9±0.07μs       38.5±0.5μs     1.21  spatial.Xdist.time_cdist(100, 'dice')
> +     17.0±0.07μs      20.2±0.06μs     1.19  spatial.Xdist.time_pdist(100, 'dice')
> +     1.90±0.01ms      2.19±0.03ms     1.15  spatial.Xdist.time_pdist(1000, 'yule')
> +     3.85±0.02ms      4.40±0.07ms     1.14  spatial.Xdist.time_cdist(1000, 'yule')
> +     20.2±0.07μs       23.0±0.3μs     1.14  spatial.Xdist.time_pdist(100, 'yule')
> +     38.3±0.02μs       43.2±0.2μs     1.13  spatial.Xdist.time_cdist(100, 'yule')
> +     19.9±0.07μs       22.0±0.3μs     1.11  spatial.Xdist.time_pdist(100, 'jaccard')
> +      38.0±0.2μs      41.6±0.06μs     1.09  spatial.Xdist.time_cdist(100, 'jaccard')
> +     1.35±0.01ms      1.47±0.02ms     1.09  spatial.Xdist.time_pdist(1000, 'russellrao')
[...]

PR #23110, opened 2025-06-04
BUG: spatial.distance: yule implementation in `distance_metrics.h`

ajz34:
#### Reference issue

Closes gh-23109 

Fixes C++ header implementation of yule metric.

Previously, internal variable type is `intptr_t`. This changes to generic type `T`, in accordance to other boolean distance metric implementations, and also fixes bug that result of `cdist` function does not equal to python implementation of `yule`.

### DIFF
--- a/scipy/spatial/src/distance_metrics.h
+++ b/scipy/spatial/src/distance_metrics.h
@@ -722,7 +722,7 @@ struct YuleDistance {
     template <typename T>
     struct Acc {
         Acc(): ntt(0), nft(0), nff(0), ntf(0) {}
-        intptr_t ntt, nft, nff, ntf;
+        T ntt, nft, nff, ntf;
     };
 
     template <typename T>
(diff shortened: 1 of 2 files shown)


tylerjereddy (approved):
I'm approving this because, after checking locally, I noticed that the regression test passes in SciPy `1.10.1` and fails in SciPy `1.11.1` (the failure at `1.11.1` is identical to the failure on `main`). The reason is almost certainly an unintended change in behavior during the move to C++ implementation at https://github.com/scipy/scipy/pull/17538.

The CI is happy and I think I might as well allow this in for `1.16.0rc2` release as well.

One thing I will say is that we've had enormous confusion with weighted distance metrics, boolean distance metrics, and the consistencies between Python and compiled implementations.

Just a small sampling of complex issues/discussion:

- https://github.com/scipy/scipy/issues/21090
- https://github.com/scipy/scipy/issues/21569
- https://github.com/scipy/scipy/issues/21135

@fancidev did a good job of some detailed writeups, and in some cases things got stuck because I probably wasn't smart enough to understand the gory details. But for this PR I don't think there was ever a discussion on changing the behavior of `yule` so simultaneously matching the Python implementation and the old flat C implementation seems hard to argue with in the absence of concrete evidence for some other behavior.

tylerjereddy:
squash-merged for backporting; thanks for your first SciPy contribution

PR #15729, opened 2022-03-10
ENH: stats: add weighted power mean

qbarthelemy:
#### Reference issue
Closes issue #12456. @candalfigomoro

#### What does this implement/fix?
It implements `pmean`, the weighted power mean.

PR #23112, opened 2025-06-04
MAINT: remove duplicated test in `stats.tests.test_stats.py`

ChVeen:
#### Reference issue
No issue involved.

This tiny PR removes an unnecessary duplicated test in `stats.tests.test_stats.py`.


mdhaber:
I see that the original PR, gh-15729, intended to run the same test for both list and array input. During the conversion to array API, the test with list input as converted to array (because with `SCIPY_ARRAY_API=1`, lists aren't accepted), resulting in two identical tests. I think it's fine that we lost the list input test. `pmean` is covered by a decorator that will ensure that list inputs are converted to NumPy arrays when `SCIPY_ARRAY_API` is not set.

mdhaber (approved):

lucascolley (approved):
thanks @ChVeen !

PR #23114, opened 2025-06-05
DOC: Add missing BLAS Level 2 functions

jorenham:
I noticed that not all of the available BLAS (level 2)  functions were listed in the docs in https://github.com/scipy/scipy-stubs/pull/594. This should take care of that


rgommers (approved):
Thanks @jorenham.

jorenham:
Perhaps this could be worth backporting

rgommers:
Sure, added the label - easy enough and makes the docs more correct

PR #23118, opened 2025-06-05
TYP: generic ``signal.ShortTimeFFT`` type

### LINKED ISSUE #545: `signal.ShortTimeFFT`: generic scalar type-parameter
The scalar type-parameter should be covariant and bound to `np.inexact`, matching the scalar type of `win`.

It should also be made generic upstream, so that `ShortTimeFFT` is also subscriptable at runtime.

- [x] `scipy-stubs` (#547)
- [ ] `scipy` (scipy/scipy#23118)

jorenham:
In scipy/scipy-stubs#547, `signal.ShortTimeFFT`  was made into an optional generic type. But it can currently only be used that way from a `.pyi`, a `if TYPE_CHECKING` block, or in certain situations with `from __future__ import annotations`. In practice, this can be rather confusing and annoying for users.

This change will allow `scipy-stubs` users (or anyone else for that matter) to subscript the `ShortTimeFFT` type in all situations, e.g.:

```py
# from __future__ import annotations 
#    ^-- not need anymore

import numpy as np
from typing import Never
from scipy.signal import ShortTimeFFT

def create_black_hole(stft: ShortTimeFFT[np.complex128]) -> Never:
    # --implementation left as an exercise to the reader--
```

Closes scipy/scipy-stubs#545

### DIFF
--- a/scipy/signal/_short_time_fft.py
+++ b/scipy/signal/_short_time_fft.py
@@ -21,6 +21,7 @@
 # Linter does not allow to import ``Generator`` from ``typing`` module:
 from collections.abc import Generator, Callable
 from functools import cache, lru_cache, partial
+from types import GenericAlias
 from typing import get_args, Literal
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


lucascolley (approved):
thanks Joren!

Out of interest, is this a temporary measure until we eventually get to a recent-enough minimum Python version? Or is it permanent?

rgommers (approved):
Thanks @jorenham. Looks fine to me. I'll leave it open for a bit in case @DietBru or any other `signal` expert has a comment.

jorenham:
> thanks Joren!
> 
> Out of interest, is this a temporary measure until we eventually get to a recent-enough minimum Python version? Or is it permanent?

Permanent probably. 3.14 will only defer annotations, which won't solve much for runtime typecheckers

Issue #21232, opened 2024-07-22
META: vendored code tracker


lucascolley:
I think that this will be a useful thing to have (to complement https://docs.scipy.org/doc/scipy/dev/core-dev/index.html#vendored-code) for a few reasons below. I thought that I'd open this before writing out the tracker in case we would prefer a different format (wiki page, docs only etc.).

1. There is less maintenance overhead for an issue than a docs page.
2. We don't have an exhaustive list yet - I only accidentally realised that we had a vendored copy of `numpydoc.docscrape` the other day.
3. It would be useful to categorise which code we do/don't maintain a fork of and maintain a list of upstream repos.
4. This will be a useful place to track potential progress from upstreams towards support for meson subprojects.

nickodell:
Hello, I performed a search of SciPy for README files, using the command `find . -name README\*` and filtered out the ones which did not seem to correspond to external software.

List of additional vendored software:

- scipy/optimize/tnc
- scipy/optimize/lbfgsb_src
- scipy/optimize/_direct
- scipy/io/_fast_matrix_market
- scipy/interpolate/fitpack
- scipy/stats/libnpyrandom
- scipy/_lib/highs

Of note:

* It seems like `scipy/stats/libnpyrandom` is part of NumPy.
* Someone seems to be maintaining a modern Fortran version of `scipy/interpolate/fitpack` at https://github.com/perazz/fitpack.
* Highs is being maintained at https://github.com/ERGO-Code/HiGHS and seems pretty active.

dschmitz89:
- DIRECT is taken from [nlopt](https://github.com/stevengj/nlopt) and maintained in the sense that bugs will likely be fixed but no new features implemented as NLopt also has its own DIRECT implementation.
- L-BFGS-B is taken from [here](https://users.iems.northwestern.edu/~nocedal/lbfgsb.html) and de facto unmaintained

lucascolley:
Thanks @nickodell ! Added above.

And thanks @dschmitz89 ! Please feel free to update the tracker and sort any of the entries from the "to be audited" section.

A few findings:

- The URL in the TNC README (http://www.jeannot.org/~js/code/index.en.html) is dead
- `scipy.interpolate` README has
```
Author: Enthought, Inc
	Austin, TX
	enthough-dev@mail.enthought.com
```
perhaps time to change to
```
Authors:
    SciPy developers
        https://discuss.scientific-python.org/c/contributor/scipy/
    Enthought, Inc
        Austin, TX
[...]

PR #23119, opened 2025-06-06
MAINT: Remove dead code in ``_lib.decorator``

jorenham:
This also takes care of a bunch of accidental exports that were reported by `stubtest`: https://github.com/scipy/scipy-stubs/blob/ebf72360308600dd84f023781230a3ba7943e51e/.mypyignore#L13


rgommers (commented):
Thanks Joren. Note that that file is a vendored one: https://github.com/micheles/decorator, https://scipy.github.io/devdocs/dev/core-dev/index.html#dependencies

It'd be better to leave it alone, excluding it from static checks or fixing issues upstream and then bumping the version that we vendor. Local changes like this are going to be overwritten otherwise when we do pull in a new version.

lucascolley:
looks like this is missing from https://scipy.github.io/devdocs/dev/core-dev/index.html#vendored-code and #21232 (and perhaps other places in `tools` that control checks)

PR #23120, opened 2025-06-07
Added code base visualization

ivanmilevtues:
I have generated and added a diagram driven documentation for the scipy project.
You can see how these document render here: https://github.com/CodeBoarding/GeneratedOnBoardings/blob/main/scipy/on_boarding.md

#### Additional information

Me and my friend are building a product for automatic document generation as we've noticed how hard it can be for new people to understand new codebases. I would love to hear what is your take on on-boarding and how do you handle it especially considering that the maintainers team is one of volunteers. I strive to create a doc which can be autogenerated whenever you think it is needed (maybe part of your CI).

Any feedback is more than welcome! I would usually open a discussion first, but they don't seem to be enabled for the scipy project so decided to go ahead and open this PR.

Full disclosure: we're trying to turn this into a startup, but we're still in a very early stage and figuring out what will actually be useful for people.

### DIFF
--- a/.codeboarding/Core Mathematical Operations.md
+++ b/.codeboarding/Core Mathematical Operations.md
@@ -1,0 +1,319 @@
+```mermaid
+graph LR
+    Dense_Linear_Algebra["Dense Linear Algebra"]
+    Sparse_Linear_Algebra["Sparse Linear Algebra"]
+    BLAS_LAPACK_Interface["BLAS/LAPACK Interface"]
+    Special_Functions["Special Functions"]
+    FFT_Module["FFT Module"]
+    Numerical_Integration["Numerical Integration"]
+    Optimization_Linear_Programming["Optimization - Linear Programming"]
+    SciPy_Utilities["SciPy Utilities"]
+    Dense_Linear_Algebra -- "uses" --> BLAS_LAPACK_Interface
+    Dense_Linear_Algebra -- "uses" --> SciPy_Utilities
+    Dense_Linear_Algebra -- "uses" --> FFT_Module
+    Dense_Linear_Algebra -- "uses" --> Special_Functions
+    Special_Functions -- "uses" --> SciPy_Utilities
+    Special_Functions -- "uses" --> Dense_Linear_Algebra
+    Numerical_Integration -- "uses" --> Special_Functions
+    Numerical_Integration -- "uses" --> SciPy_Utilities
+    BLAS_LAPACK_Interface -- "uses" --> SciPy_Utilities
+    FFT_Module -- "uses" --> SciPy_Utilities
+    Sparse_Linear_Algebra -- "uses" --> SciPy_Utilities
+    Sparse_Linear_Algebra -- "uses" --> Dense_Linear_Algebra
+    Sparse_Linear_Algebra -- "uses" --> BLAS_LAPACK_Interface
+```
+[![CodeBoarding](https://img.shields.io/badge/Generated%20by-CodeBoarding-9cf?style=flat-square)](https://github.com/CodeBoarding/GeneratedOnBoardings)[![Demo](https://img.shields.io/badge/Try%20our-Demo-blue?style=flat-square)](https://www.codeboarding.org/demo)[![Contact](https://img.shields.io/badge/Contact%20us%20-%20contact@codeboarding.org-lightgrey?style=flat-square)](mailto:contact@codeboarding.org)
+
+## Component Details
+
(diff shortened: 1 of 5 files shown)


lucascolley (changes requested):
first reaction: I don't think these pages will be helpful for SciPy contributors. To get familiar with the various submodules of SciPy, there is much clearer and more organised information at https://docs.scipy.org/doc/scipy/tutorial/index.html. And we describe various parts of the internal codebase in https://docs.scipy.org/doc/scipy/dev/contributor/contributor_toc.html.

The mermaid graphs also look a bit random in terms of which items are picked out. I think they would just confuse me as a new contributor, to be honest.

I could see something along the lines of these pages having value for a project, but I think they would take significant vetting and effort from those maintainers with deep knowledge of the project, to weed out all of the LLM generated oddities and potential inaccuracies.

ivanmilevtues:
Thank you for the response @lucascolley! 

Thanks for linking your existing documentation. I wonder what is your opinion on diagram first docs as the one we have with the mermaid, do you prefer the classic one which is more of a list of documents rather?

Further on the feedback you gave about the LLMs being inaccurate, we are trying to leverage Static Anlaysis to help with this problem, however I can see that it is still inaccurate. I expect that even if a person wrote it you would still like to proof read it, so I would like to hear what is you take in more general terms about documentation maintaince?

Lastly, again thanks for your time but it would be awesome if you can share how do you imagine the process of getting to know the SciPy codebase to be i.e. you read through the contribution guides and then play with the codebase around by doing your first task and intreact with people in the community or something else?

Thanks for the feedback we will try to improve and address the points you made!

ilayn:
I don't think this is a fair way of gathering feedback for a prospective commercial offering using open source project of this size with an incomplete product. Please refrain from requesting feedback into big projects like this. I see you already opened many issues in many projects. 

However, we (OSS maintainers in general) already have very limited time to maintain the projects and we shouldn't be forced to shape your to-be-product at this stage. I think Lucas already gave quite valuable feedback and let's keep it at this point since we don't have any use case for this. If needed, folks can trigger your product on the codebases themselves should the adoption happens. 

I wish you great success with your initiatives.

lucascolley:
> how do you imagine the process of getting to know the SciPy codebase to be i.e. you read through the contribution guides and then play with the codebase around by doing your first task and intreact with people in the community

Yes, that sounds about right! Given that SciPy is a very large codebase, I think it is expected that new contributors will not start out with a good understanding of the whole library, rather they will contribute to some specific small areas. If they stick around and start contributing more heavily, then yes, I think most of the learning comes from interacting with the community, watching how maintainers react to issues and pull requests, and reading both user documentation and things like our roadmap: https://docs.scipy.org/doc/scipy/dev/roadmap.html.

> do you prefer the classic one which is more of a list of documents rather?

I feel like 3 things are important for me, in this order:
1. Accuracy
2. Clarity
3. Being quick/easy to process/understand

I think the list style, tutorials + API reference in the docs will always be better for accuracy and depth of information. I think it probably also wins on clarity, as it doesn't have to categorise and distill information into nodes on a graph.

The visualisation will definitely help some people with ease/speed of understanding, *if* the information it contains is sufficiently accurate and clear. But I think it would be quite difficult to decide how to organise that information in a useful way.

> I expect that even if a person wrote it you would still like to proof read it, so I would like to hear what is you take in more general terms about documentation maintaince?

Large docs contributions are rare, as they can take a long time to review and more importantly require significant understanding on the side of contributors. Hence our documentation maintenance is typically about iterating, making small improvements piece by piece. This is greatly helped by users raising issues when documentation is unclear.

I would say there is nothing especially problematic about it being LLM generated when we look at it from a pure quality perspective. The problem is that it's basically the same as if a developer with lots of experience but who has never developed SciPy before submitted a PR of a similar size—we would expect that to be of similar quality.

---

You may find this interesting, a relatively recent map of our module structure. It is a difficult task to convey all of the useful information contained here to interested contributors.

image

I'll close this PR out now as I don't think this will be merged, but I hope the feedback was somewhat helpful.

EDIT: Ilhan beat me to it

ivanmilevtues:
Thank you both for the responses @lucascolley and @ilayn!

I do agree that maintainers are busy and that is a part of why we want to build something to help them!
Firstly, I am thankful for the time spent giving us feedback, you and everyone who does so in the OSS community! And I really don't want to force people to spend time on giving us feedback, that is why whenever possible I open discussions, which I think are more "casual" from the maintainers side and more about the community of the project. This said SciPy doesn't have the discussion panel open so I decided to do a PR.

The insights you've given us are invaluable and we will definetly take them into account.

Thanks for the effort and time, I wish you all the best!

ivanmilevtues:
Hey @lucascolley and @ilayn, just wanted to do a quick update that we are now open-source so if you are interested you can take a look: https://github.com/CodeBoarding/CodeBoarding

Issue #22825, opened 2025-04-11
ENH: new GH runners available for macOS 15 and Windows 2025


dschmitz89:
### Is your feature request related to a problem? Please describe.

Github action runners for macOS15 and Windows 2025 are now [available](https://github.blog/changelog/2025-04-10-github-actions-macos-15-and-windows-2025-images-are-now-generally-available/#:~:text=GitHub%20Actions%3A%20macOS%2015%20and%20Windows%202025%20images%20are%20now%20generally%20available,-actions&text=April%2010%2C%202025-,macOS%2015%20and%20Windows%202025%20images%20are%20now%20generally%20available,hosted%20standard%20or%20larger%20runners.)

We should include at least maxOS 15 in the wheels of the next release.

rgommers:
Thanks for sharing that these images are available @dschmitz89.

> We should include at least macOS 15 in the wheels of the next release.
 
I think this misunderstands how wheels work. We already have wheels that work on macOS 15 (and will work on 16 as well). So this is not related to our release wheels. I'll remove the label for that from this issue.

This is really business as usual. The announcement says _"The latest tag will migrate to these images later in the year."_ It may be useful for someone to switch a CI job from `macos-latest` and `windows-latest` to these images to see if there are any problems, and preemptively fix them if they occur. So this is more CI maintenance than anything else.

tylerjereddy:
Loosely related--I think I saw Matti point out the general availability of Windows + ARM runners on GHA: https://github.blog/changelog/2025-04-14-windows-arm64-hosted-runners-now-available-in-public-preview/

If the wait times for that platform/hardware are not too long, IIRC we do get some performance patches for that combo these days.

andyfaff:
Ah, the mythical winarm runners have arrived. Do we know when we'll have OpenBLAS wheels? 
@matthew-brett do you have any material for a PR?

tylerjereddy:
https://github.com/MacPython/openblas-libs/pull/193 may be related

rgommers:
I think there's nothing left to do here now that we're using the Windows on Arm runners, so closing.

PR #23122, opened 2025-06-07
BLD: Windows Arm64 wheel support

### LINKED ISSUE #21562: BLD: Windows Arm builds
### Describe your issue.

Can't install scipy on win arm machine for native Python. No pre-built wheels, from source compilation fails.

### Reproducing Code Example

```python
pip install scipy
```

### Error message

```shell
Downloading scipy-1.14.1.tar.gz (58.6 MB)
     --------------------------------------- 58.6/58.6 MB 24.2 MB/s eta 0:00:00
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'done'
  Installing backend dependencies: started
  Installing backend dependencies: finished with status 'done'
  Preparing metadata (pyproject.toml): started
  Preparing metadata (pyproject.toml): finished with status 'error'
  error: subprocess-exited-with-error

  Preparing metadata (pyproject.toml) did not run successfully.
  exit code: 1

  [39 lines of output]
  + meson setup C:\Users\gkhmyznikov\AppData\Local\Temp\pip-install-1pan2rrf\scipy_d66596b99a184dd3a6302c554d0a9d7c C:\Users\gkhmyznikov\AppData\Local\Temp\pip-install-1pan2rrf\scipy_d66596b99a184dd3a6302c554d0a9d7c\.mesonpy-m6g0z9mh -Dbuildtype=release -Db_ndebug=if-release -Db_vscrt=md --native-file=C:\Users\gkhmyznikov\AppData\Local\Temp\pip-install-1pan2rrf\scipy_d66596b99a184dd3a6302c554d0a9d7c\.mesonpy-m6g0z9mh\meson-python-native-file.ini
  The Meson build system
  Version: 1.5.1
  Source dir: C:\Users\gkhmyznikov\AppData\Local\Temp\pip-install-1pan2rrf\scipy_d66596b99a184dd3a6302c554d0a9d7c
  Build dir: C:\Users\gkhmyznikov\AppData\Local\Temp\pip-install-1pan2rrf\scipy_d66596b99a184dd3a6302c554d0a9d7c\.mesonpy-m6g0z9mh
  Build type: native build
  Project name: scipy
  Project version: 1.14.1
  Activating VS 17.8.3
  C compiler for the host machine: cl (msvc 19.38.33133 "Microsoft (R) C/C++ Optimizing Compiler Version 19.38.33133 for ARM64")
  C linker for the host machine: link link 14.38.33133.0
  C++ compiler for the host machine: cl (msvc 19.38.33133 "Microsoft (R) C/C++ Optimizing Compiler Version 19.38.33133 for ARM64")
  C++ linker for the host machine: link link 14.38.33133.0
  Cython compiler for the host machine: cython (cython 3.0.11)
  Host machine cpu family: aarch64
  Host machine cpu: aarch64
  Program python found: YES (X:\Github\PWABuilder\Windows-DevRel\Tools\ML-tests\.temp\env_scipy\Scripts\python.exe)
  Run-time dependency python found: YES 3.11
  Program cython found: YES (C:\Users\gkhmyznikov\AppData\Local\Temp\pip-build-env-cw5uc0or\overlay\Scripts\cython.EXE)
  Compiler for C supports arguments -Wno-unused-but-set-variable: NO
  Compiler for C supports arguments -Wno-unused-function: NO
  Compiler for C supports arguments -Wno-conversion: NO
  Compiler for C supports arguments -Wno-misleading-indentation: NO
  Library m found: NO

  ..\meson.build:84:0: ERROR: Unknown compiler(s): [['ifort'], ['gfortran'], ['flang'], ['pgfortran'], ['g95']]
  The following exception(s) were encountered:
  Running `ifort --version` gave "[WinError 2] The system cannot find the file specified"
  Running `ifort -V` gave "[WinError 2] The system cannot find the file specified"
  Running `gfortran --version` gave "[WinError 2] The system cannot find the file specified"
  Running `gfortran -V` gave "[WinError 2] The system cannot find the file specified"
  Running `flang --version` gave "[WinError 2] The system cannot find the file specified"
  Running `flang -V` gave "[WinError 2] The system cannot find the file specified"
  Running `pgfortran --version` gave "[WinError 2] The system cannot find the file specified"
  Running `pgfortran -V` gave "[WinError 2] The system cannot find the file specified"
  Running `g95 --version` gave "[WinError 2] The system cannot find the file specified"
  Running `g95 -V` gave "[WinError 2] The system cannot find the file specified"

  A full log can be found at C:\Users\gkhmyznikov\AppData\Local\Temp\pip-install-1pan2rrf\scipy_d66596b99a184dd3a6302c554d0a9d7c\.mesonpy-m6g0z9mh\meson-logs\meson-log.txt
  [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed
System.Management.Automation.RemoteException
Encountered error while generating package metadata.
System.Management.Automation.RemoteException
See above for output.
System.Management.Automation.RemoteException
```

```shell
scipy 1.14.1
python 3.11.9 win-arm64
```

khmyznikov:
#### Reference issue
Closes: https://github.com/scipy/scipy/issues/21562 https://github.com/scipy/scipy/issues/22825

Added support for building wheels for Windows Arm64

#### Additional information
Bumped scipy-openblas32==0.3.29.265.1
Added llvm-strip into repair_windows.sh for WoA


khmyznikov:
@Mugundanmcw can you look at this too? I saw you worked on that too.

MugundanMCW:
@khmyznikov 

- The workaround that made for "fortranRuntime.lib" issue from meson LGTM, since the PR for this fix merged recently we can have this workaround until we have the latest meson wheel being released via PIP.
- I have one doubt with the paths being added in this patch wrt to LLVM lib paths, do we really need them?
image
- Also after building the wheel, we need to link OpenBLAS via repair_windows.sh with scipy, are we doing this? else we are importing scipy_openblas32 while importing scipy? @rgommers let me know if I am missing something here

requirements/openblas.txt
```diff
@@ -1 +1 @@
-scipy-openblas32==0.3.28.0.2
+scipy-openblas32==0.3.29.265.0
```
rgommers: This should be `scipy-openblas32==0.3.29.265.1` I think, since that contained a fix: https://github.com/MacPython/openblas-libs/pull/200

.github/workflows/wheels.yml
```diff
@@ -212,12 +213,53 @@ jobs:
...
+      - name: Install LLVM for win_arm64
+        if: matrix.buildplat[1] == 'win' && matrix.buildplat[2] == 'ARM64'
+        uses: KyleMayes/install-llvm-action@v2
+        id: install_llvm
+        with:
+          version: 20.1.6
+          force-url: https://github.com/llvm/llvm-project/releases/download/llvmorg-20.1.6/LLVM-20.1.6-woa64.exe
+          env: false # Do not set CC/CXX by the action          
+
+      # pkgconf and FortranRuntime hacks may be removed after fixes in according libraries
```
rgommers: @Mugundanmcw's fix for this landed in Meson's `main` branch, so I think we'd rather install `meson` from a fixed commit on its `main` branch and do a non-isolated build to ensure we pick up that version of `meson`. 

Note: I'm not completely sure it's easier, and @Mugundanmcw just commented that they think this approach is fine, so please don't change it yet - but I do find this harder to understand at least.

rgommers:
> * Also after building the wheel, we need to link OpenBLAS via repair_windows.sh with scipy, are doing this?

Yes, this is already taken care of. It's hooked up via `repair-wheel-command` in `pyproject.toml`, which `cibuildwheel` will invoke.

.github/workflows/wheels.yml
```diff
@@ -212,12 +213,53 @@ jobs:
...
+        shell: cmd
+        run: |
+          CALL "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" arm64
+
+          pip install build
+          pip install -r requirements/build.txt -r requirements/openblas.txt
+          python -c "import scipy_openblas32; print(scipy_openblas32.get_pkg_config())" > %GITHUB_WORKSPACE%\scipy-openblas.pc
+          
+          set VCPKG_ROOT=C:\vcpkg
+          vcpkg install pkgconf
```
rgommers: I assume that this is because we're still getting the x86 version from Chocolatey, and that doesn't have win-arm64? 

Once we add `win_arm64` support to https://pypi.org/project/pkgconf/, we can use that instead I think - easier to understand and should work equally well.

.github/workflows/wheels.yml
```diff
@@ -212,12 +213,53 @@ jobs:
...
+
+          set "LLVM_LIB_DIRS=C:\Program Files\LLVM\lib\clang\20\lib\windows;C:\Program Files\LLVM\lib"
+          set "LIB=%LLVM_LIB_DIRS%;%LIB%"
+          set "LIBPATH=%LLVM_LIB_DIRS%;%LIBPATH%"
+
+          set CC=clang-cl
+          set CXX=clang-cl
+          set FC=flang
+
+          python -m build --wheel --no-isolation -o wheelhouse .
```
rgommers: This is incorrect. `cibuildwheel` should do the wheel building, you cannot invoke it manually, that will not result in a portable wheel.

For disabling isolation, you can do the same as in the step `Disable build isolation for python free-threaded version`.

.github/workflows/wheels.yml
```diff
@@ -212,12 +213,53 @@ jobs:
...
 
       - name: Build wheels
+        if: matrix.buildplat[1] != 'win' && matrix.buildplat[2] != 'ARM64' # Skip for win_arm64 for manual build
         uses: pypa/cibuildwheel@cf078b0954f3fd08b8445a7bf2c3fb83ab3bb971 # v3.0.0b4
         env:
           CIBW_BUILD: ${{ matrix.python[0] }}-${{ matrix.buildplat[1] }}*
           CIBW_ARCHS: ${{ matrix.buildplat[2] }}
           CIBW_ENABLE: cpython-freethreading cpython-prerelease
 
+      # can be remove after https://github.com/actions/partner-runner-images/issues/90
```
rgommers: That's unfortunate. 

Note to self: inspect the added action here.

rgommers (changes requested):
Thanks @khmyznikov! Here is an initial set of comments and questions.

.github/workflows/wheels.yml
MugundanMCW: @rgommers installing meson from that commit should be easier and proper way to build SciPy without this workaround. This involves cloning meson and `cd meson\` then followed by `pip install .` which can install meson from source with actual fix

@khmyznikov Please let me know if you need any assistance

.github/workflows/wheels.yml
rgommers: Actually, the `choco` version does seem to work fine for NumPy: https://github.com/numpy/numpy/pull/29039

rgommers: That said, `choco` has always been slow and a little annoying to deal with. So maybe `vcpkg` is an improvement either way. If you put it in a separate job step, which is useful anyway I think, you can see how much faster it is compared to `choco` (that takes about a minute).

.github/workflows/wheels.yml
andyfaff: You mean the `KyleMayes/install-llvm-action`? That should be given a hash, not a version.

.github/workflows/wheels.yml
```diff
@@ -212,12 +213,53 @@ jobs:
...
+      - name: Install LLVM for win_arm64
+        if: matrix.buildplat[1] == 'win' && matrix.buildplat[2] == 'ARM64'
+        uses: KyleMayes/install-llvm-action@v2
```
andyfaff: Use a hash, not a version.

.github/workflows/wheels.yml
rgommers: No need for cloning, you can use:
```
pip install git+https://github.com/mesonbuild/meson.git@<commit-hash>
```

.github/workflows/wheels.yml
andyfaff: Yes, needs to go through cibuildwheel. I think that all the environment variable may need to be put in `GITHUB_PATH`/`GITHUB_ENV`.

.github/workflows/wheels.yml
rgommers: Yes that's the one. Reading through all source code of new actions that are added and assessing whether it's safe to rely on security-wise is always an unfortunate chore.

andyfaff: I'm surprised that it's not installed on the runner. I wonder if we could ask GH to include the software.

MugundanMCW: LLVM is already preinstalled in GH runners for Win-ARM64 as I could see from [here](https://github.com/actions/partner-runner-images/blob/main/images/arm-windows-11-image.md)

.github/workflows/wheels.yml
rgommers: Answering my own question on timing from @Mugundanmcw's [CI log](https://github.com/Mugundanmcw/scipy_for_winarm64/actions/runs/15515986619/job/43682938920): it also takes just over a minute with `vcpkg`.

.github/workflows/wheels.yml
rgommers: The last two comments are already addressed by the exact link you're commenting on here: the runner image has a non-native `libclang` installed.

rgommers: Seems like it can be done in a few lines of code, from https://github.com/Mugundanmcw/scipy_for_winarm64/actions/runs/15515986619/workflow#L158-L162:

```
      - name: Download and install LLVM installer
        run: |
          Invoke-WebRequest https://github.com/llvm/llvm-project/releases/download/llvmorg-19.1.5/LLVM-19.1.5-woa64.exe -UseBasicParsing -OutFile LLVM-woa64.exe
          Start-Process -FilePath ".\LLVM-woa64.exe" -ArgumentList "/S" -Wait
          echo "C:\Program Files\LLVM\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
```

Assuming that that's equivalent, that seems nicer than using this action - just need to trust that one links, rather than lots of TypeScript code that's hard to review.

.github/workflows/wheels.yml
khmyznikov: Got success with cibuildwheel, will update the PR soon

.github/workflows/wheels.yml
khmyznikov: @Mugundanmcw I've replaced the step with more simple Invoke-WebRequest approach, but if it can be done differently I'll be glad if you can assist on this.

khmyznikov:
@andyfaff @rgommers @larsoner @Mugundanmcw how it looks now? Is anything left?

MugundanMCW (approved):

.github/workflows/wheels.yml
MugundanMCW: @rgommers & @khmyznikov  I see that Meson had a [release](https://github.com/mesonbuild/meson/releases/tag/1.8.2) recently which includes the fix for flang. We can remove the copy operation of "fortranruntime.lib" and install meson directly via pip.

tools/wheels/repair_windows.sh
```diff
@@ -4,6 +4,17 @@ WHEEL="$1"
 DEST_DIR="$2"
 OPENBLAS_DIR=$(python -c"import scipy_openblas32 as sop; print(sop.get_lib_dir())")
 
+# Determine the strip command based on TARGET_ARCH
+# TARGET_ARCH should be set by the CI environment (e.g., ARM64, AMD64)
+TARGET_ARCH="${TARGET_ARCH:-}" # Default to empty string if not set
+STRIP_COMMAND="strip"
+if [ "$TARGET_ARCH" = "ARM64" ]; then
+  STRIP_COMMAND="llvm-strip"
+  echo "Using llvm-strip for ARM64 target."
```
rgommers: Are you able to check locally whether this stripping actually changes anything? IIRC the only reason we have it is because `mingw-w64` is leaving (or used to leave) some bad content at the end of each `.pyd` file. For LLVM-based compilers, this may well do nothing, since we're not building with debug info or anything like that.

.github/workflows/wheels.yml
rgommers: Ah awesome, always nice if fixes propagate quickly

.github/workflows/wheels.yml
```diff
@@ -210,13 +211,46 @@ jobs:
         run: |
           CIBW="pip; args: --no-build-isolation"
           echo "CIBW_BUILD_FRONTEND=$CIBW" >> "$GITHUB_ENV"
-
+      
       - name: Build wheels
+        if: matrix.buildplat[1] != 'win' || matrix.buildplat[2] != 'ARM64' # win_arm64 build configured separately
```
rgommers: This doesn't actually need to be done separately, right? It's more maintainable to keep common things common. You can set the env vars you need in a step before the regular "Build wheels" step, and pass them on with:
```
echo "CC=clang-cl" >> $GITHUB_ENV
```

.github/workflows/wheels.yml
```diff
@@ -210,13 +211,46 @@ jobs:
...
+          CIBW_ENVIRONMENT_WINDOWS: >
+            PATH="C:\\vcpkg\\installed\\arm64-windows\\tools\\pkgconf;$PATH"
+            CC=clang-cl
+            CXX=clang-cl
+            FC=flang
+            PKG_CONFIG_PATH=$GITHUB_WORKSPACE
+            TARGET_ARCH=${{ matrix.buildplat[2] }}
+          CIBW_BEFORE_BUILD_WINDOWS: >
+            pip install wheel delvewheel &&
+            pip install -r requirements/build.txt -r requirements/openblas.txt &&
```
rgommers: These installs aren't necessary, they're already done elsewhere, e.g. in `tools/wheels/cibw_before_build_win.sh`

rgommers (changes requested):
Thanks @khmyznikov, this is getting there. It can still be made significantly more concise it looks like to me, we should be able to get this down to adding the `windows-11-arm` entry to the list of platforms, and two win-arm64 specific steps of installing the needed LLVM/pkgconf and setting the env vars. Plus maybe the stripping change if that's needed.

.github/workflows/wheels.yml
rgommers: Actually the syntax in gh-23133 looks even more concise, `$env:cc = "clang-cl"`

rgommers:
I plan to merge gh-23133 first, so the LLVM and pkgconf changes are already packaged up in a reusable action. That will make the diff here really small.

tools/wheels/repair_windows.sh
khmyznikov: No I didn't checked that. I guess you are right. But how it can be checked? Can I compare the ending of x64 without repair and arm64?

.github/workflows/wheels.yml
khmyznikov: @rgommers It was looked a bit messy if I try to wrap to IFs all extra code for arm64 to keep everything inside one build wheel step. I'm not sure the syntax allows to keep the commands in a mupliple rows as it is for now. But I'll do it if you suggest that.

.github/workflows/wheels.yml
khmyznikov: Let me try again, I think it was needed for some reason.

tools/wheels/repair_windows.sh
rgommers: How about inserting debug output that just checks the file size right before and after stripping. `ls -l` on the .whl file?

.github/workflows/wheels.yml
rgommers: It should allow you to keep it in multiple rows. You should only need one extra statement per env var to export it to the next job steps.

rgommers:
@khmyznikov also note that CI runs on this repo aren't useful to you, since they don't run `wheels.yml` unless you add `[wheel build]` to the first line of the most recent commit message. I suggest to use `[ci skip]` by default on pushes here, and run the windows part of `wheels.yml` on your fork until it passes. And right at the end we do a full set of wheel builds on this PR before merging.

khmyznikov:
> And right at the end we do a full set of wheel builds on this PR before merging.

I've runned all and individual wheels inside my fork to make sure everything else is building.

requirements/openblas.txt
```diff
@@ -1 +1 @@
-scipy-openblas32==0.3.28.0.2
+scipy-openblas32==0.3.29.265.1
```
rgommers: This change has been made in `main` now, so you can drop it from this PR.

tools/wheels/repair_windows.sh
khmyznikov: @rgommers yeah the size difference for x64 after strip is 14mb less, for arm it's 0.158kb bigger, almost identical. I guess we can count that as something we can just skip for arm right?

khmyznikov: but the `delvewheel repair` still needed right?

rgommers: Indeed, let's skip it then.

> but the `delvewheel repair` still needed right?

Yep. What that does is vendor in `libopenblas.dll`.

.github/workflows/wheels.yml
khmyznikov: Added the cibw_before_build_win.sh call

.github/workflows/wheels.yml
khmyznikov: I've refactored that. Back to common cibuildwheel. Envs for Windows was combined, separate step for arm64 CIBW_BEFORE_BUILD_WINDOWS

tools/wheels/repair_windows.sh
khmyznikov: Done

khmyznikov:
@rgommers what's next? 😃

lucascolley:
It looks like gh-23133 is still hitting some failures after https://github.com/scipy/scipy/pull/23134

rgommers:
Now that gh-23133 is merged, we're reasonably confident that there are no remaining issues on `win_arm64` anymore. So hopefully we can shrink this to the minimum required to build wheels - I'll add some review comments.

.github/workflows/wheels.yml
```diff
@@ -214,6 +199,50 @@ jobs:
...
+
+      - name: Install rtools for win_amd64
+        if: ${{ matrix.buildplat[1] == 'win' && matrix.buildplat[2] == 'AMD64' }}
+        run: |
+          # mingw-w64
+          choco install rtools -y --no-progress --force --version=4.0.0.20220206
+          echo "c:\rtools40\ucrt64\bin;" >> $env:GITHUB_PATH
+
+      # can be remove after https://github.com/actions/partner-runner-images/issues/90
+      - name: Install LLVM for win_arm64
```
rgommers: Installing LLVM and `pkg-config` can now be done as a one-liner, invoking https://github.com/scipy/scipy/blob/main/.github/windows_arm64_steps/action.yml. As done here:

https://github.com/scipy/scipy/blob/fbb6ff839cfdf02304a3bd21d46a1189c01832e9/.github/workflows/windows.yml#L193-L194

.github/workflows/wheels.yml
```diff
@@ -214,6 +199,50 @@ jobs:
...
+          if ('${{ matrix.buildplat[1] }}' -eq 'win' -and '${{ matrix.buildplat[2] }}' -eq 'ARM64') {
+            echo "CC=clang-cl" >> $env:GITHUB_ENV
+            echo "CXX=clang-cl" >> $env:GITHUB_ENV
+            echo "FC=flang" >> $env:GITHUB_ENV
+            echo "TARGET_ARCH=${{ matrix.buildplat[2] }}" >> $env:GITHUB_ENV
+            echo "PATH=C:\vcpkg\installed\arm64-windows\tools\pkgconf;$($env:PATH)" >> $env:GITHUB_ENV
+          }
+
+      - name: Install rtools for win_amd64
+        if: ${{ matrix.buildplat[1] == 'win' && matrix.buildplat[2] == 'AMD64' }}
```
rgommers: Can you put this back in its original place to make the diff smaller please? I think all it needs is adding `&& matrix.buildplat[2] == 'AMD64'`

.github/workflows/wheels.yml
```diff
@@ -214,6 +199,50 @@ jobs:
           CIBW="build; args: --no-isolation"
           echo "CIBW_BUILD_FRONTEND=$CIBW" >> "$GITHUB_ENV"
 
+      - name: Windows - set environment variables
```
rgommers: I have a slight preference for keeping the common part unchanged, and adding the "ARM64 specific settings in a separate step. That shrinks the diff further, and also it's easier to see in the GitHub UI that the step didn't run for AMD64.

.github/workflows/wheels.yml
rgommers: Oh even better!

khmyznikov: Ended up having this step, something is still off without it, need deeper look

.github/workflows/wheels.yml
khmyznikov: I've tried to organize it to have all windows stuff together but, if smaller patch preferred - I don't mind.

.github/workflows/wheels.yml
khmyznikov: Moved Arm specific to own step

khmyznikov:
WoA wheels generated just fine: https://github.com/khmyznikov/scipy/actions/runs/16532181797

rgommers:
Thanks @khmyznikov. I built a set of wheels here as well to verify: https://github.com/rgommers/scipy/actions/runs/16537298741. Looking at the logs now.

I also posted https://discuss.scientific-python.org/t/scipy-native-wheel-for-windows-on-arm/1448/6 to make visible that we're about to merge this PR.

rgommers (approved):
Okay, CI logs look good, let's declare victory here. Thanks @khmyznikov!

rgommers:
I kicked off nightly wheel builds in https://github.com/scipy/scipy/actions/runs/16543418442, they should appear in half an hour or so at https://anaconda.org/scientific-python-nightly-wheels/scipy/files.

khmyznikov:
@rgommers thanks for your guidance and @Mugundanmcw for test pipelines and deeper investigation🤝

khmyznikov:
@rgommers we are already started the contribution to [scikit-learn](https://github.com/scikit-learn/scikit-learn/issues/30567), any ETAs for scipy wheel broadly available on pypi?

rgommers:
If we backport it, then it'll land in 1.16.2 probably in 6-10 weeks or so. If not, then 1.17.0 will only come out around the end of the year. That's more a question for @tylerjereddy as the release manager though. Tyler, are you comfortable backporting  gh-23133 plus this PR and release `win_arm64` wheels for 1.16.2?

khmyznikov:
Sooner is always better in this case, I hope it will be possible to land it with 1.16.2

tylerjereddy:
> Tyler, are you comfortable backporting https://github.com/scipy/scipy/pull/23133 plus this PR and release win_arm64 wheels for 1.16.2?

@rgommers sure, I went ahead and put a tentative backport label on both PRs after reading through their diffs. I put a calendar reminder to start preparing the release ~6 weeks from now (September 10th).

PR #23130, opened 2025-06-08
DEV: spin: change shorthand of `--setup-args` to `-S` to avoid `UserWarning`

### LINKED ISSUE #23116: BUG: `UserWarning` due to `spin build`
`spin build` invocation gives the following,

```zsh
(scipy-dev) 2:32:50:~/Quansight/scipy % spin build
/Users/czgdp1807/mambaforge/envs/scipy-dev/lib/python3.13/site-packages/click/core.py:1193: UserWarning: The parameter -C is used more than once. Remove its duplicate as parameters should be unique.
  parser = self.make_parser(ctx)
/Users/czgdp1807/mambaforge/envs/scipy-dev/lib/python3.13/site-packages/click/core.py:1186: UserWarning: The parameter -C is used more than once. Remove its duplicate as parameters should be unique.
  self.parse_args(ctx, args)
```

This is coming from duplicate use of `-C` short hand in the following places,

https://github.com/scipy/scipy/blob/9326f58e02a54218dcb4ebee9839a8df3f51ae41/.spin/cmds.py#L37-L40

https://github.com/scientific-python/spin/blob/fee6abe42cea118c2e4577cce5e0ca034c55748e/spin/cmds/meson.py#L275-L283

i.e., we are using `-C` short hand for `--setup-args` and `spin` uses `-C` short hand for `--build-dir`. Two ways I can think of,

i. Change the short hand for `--setup-args` in SciPy. Easy way out. Would definitely work (AFAICT).
ii. Change the short hand for `--build-dir` by overriding on our end.

Let me know which one to pick or if there is a third way to fix this.

cc: @rgommers

czgdp1807:
#### Reference issue

Closes https://github.com/scipy/scipy/issues/23116

Implements the solution in,

>  If there isn't, we can change to:
> 
> ```
> -C, --build-dir
> -S, --setup-args
> ```

#### Additional information

cc: @rgommers

### DIFF
--- a/.github/workflows/pixi.toml
+++ b/.github/workflows/pixi.toml
@@ -33,7 +33,7 @@ gmpy2 = "*"
 ccache = ">=4.10.1,<5"
 
 [feature.build.tasks]
-build = { cmd = "spin build -C-Dblas=blas -C-Dlapack=lapack -C-Duse-g77-abi=true", cwd = "../..", env = { CC = "ccache $CC", CXX = "ccache $CXX", FC = "ccache $FC" } }
+build = { cmd = "spin build -S-Dblas=blas -S-Dlapack=lapack -S-Duse-g77-abi=true", cwd = "../..", env = { CC = "ccache $CC", CXX = "ccache $CXX", FC = "ccache $FC" } }
 wheel = { cmd = "python -m build -wnx -Cbuild-dir=build-whl && cp dist/*.whl ../../wheelhouse/", cwd = "../.." }
 
 [feature.test.tasks]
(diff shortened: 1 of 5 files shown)


lucascolley (commented):
CI failure looks real, should just need to update

https://github.com/scipy/scipy/blob/c0c93b8d2b1f0e3f96d45d6b9308c30ef5e278d8/.github/workflows/pixi.toml#L36

czgdp1807:
Ah! okay. I see. I just filtered out `.yml` files. Let me search the whole code base for `spin build`. :-)

czgdp1807:
The tests seem to pass. :-).

rgommers (approved):
Looks like you got them all, so approving. @lucascolley @crusaderky for visibility, this will break the `build` and `build-nogil` commands in https://github.com/rgommers/pixi-dev-scipystack/, making it a bit annoying temporarily to switch back and forth in branches that do or don't have the merge commit for this PR. We can adjust the timing for this merge, or perhaps better - change `pixidev` to the full `--setup-args` first. 

I'm out of time for today, so I won't hit the merge button now.

lucascolley:
No concerns from me either way 👍

rgommers:
Okay, replaced now with `--setup-args` in the pixi dev setup. So in it goes. Thanks again!

PR #22181, opened 2024-12-24
REL: set 1.15.0rc3 unreleased

tylerjereddy:
* Set the version to SciPy `1.15.0rc3.dev0` (unreleased).

[ci skip] [skip ci] [skip cirrus] [skip circle]

Note that while there are no plans to actually do an RC3, only time will tell what is really needed. In any case, if we don't need RC3 we can simply switch the version string after any additional backports, when ready.

PR #23131, opened 2025-06-08
REL: set 1.16.0rc3 unreleased

tylerjereddy:
* Set the version of SciPy to `1.16.0rc3.dev0` (unreleased).

[ci skip] [skip ci]

We're unlikely to do an RC3, and this is just a convention (did it last cycle too: https://github.com/scipy/scipy/pull/22181) with the CI turned off.

In all likelihood, if there are no major problems, we'll just do `1.16.0` "final" by adjusting the `version` string again, etc.

We're still on track for the originally-scheduled final release on June 18th--that's 1.5 weeks away still.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,7 +43,7 @@ requires = [
 
 [project]
 name = "scipy"
-version = "1.16.0rc2"
+version = "1.16.0rc3.dev0"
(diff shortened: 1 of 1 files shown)


rgommers (approved):
Thanks Tyler!

PR #23132, opened 2025-06-08
BUG: signal.sawtooth: handle small negative values

### LINKED ISSUE #22940: BUG: nan value from signal.sawtooth ?
### Describe your issue.

Hi, 
I noticed a strange nan output value when using waveforms.py (scipy signal.sawtooth with a width of 1 and a phase shift). See attached example code. Let me know if I'm doing something dumb or if this is a bug, I am not an expert! 
Thanks.

### Reproducing Code Example

```python
import numpy as np
from numpy import pi
import scipy
from scipy import signal as s
import sys
print('python version', sys.version)
print('nympy version', np.__version__)
print('scipy verions', scipy.__version__)

Nmax=50
t = np.linspace(0, 1, Nmax, endpoint=False) 

phase=0.22
w=1 # this provides an unexpected nan value in this example code!
#w=0.9999999999999999 #this hack seems to fix the problem, but not great...

def signal(phase):  
    saw = s.sawtooth(2*pi*(t-0.5)-2*pi*phase, width=w) # phase shifted saw     
    return saw

nan_mask = np.isnan(signal(phase))
nan_indices = np.where(nan_mask)[0]
print("NaN indices (NumPy):", nan_indices)

print(signal(phase))
```

### Error message

```shell
python version 3.9.13 (main, Aug 25 2022, 23:51:50) [MSC v.1916 64 bit (AMD64)]
nympy version 1.21.5
scipy verions 1.9.1
NaN indices (NumPy): [36]
[-0.44 -0.4  -0.36 -0.32 -0.28 -0.24 -0.2  -0.16 -0.12 -0.08 -0.04  0.
  0.04  0.08  0.12  0.16  0.2   0.24  0.28  0.32  0.36  0.4   0.44  0.48
  0.52  0.56  0.6   0.64  0.68  0.72  0.76  0.8   0.84  0.88  0.92  0.96
   nan -0.96 -0.92 -0.88 -0.84 -0.8  -0.76 -0.72 -0.68 -0.64 -0.6  -0.56
 -0.52 -0.48]
```

```shell
python version 3.9.13 (main, Aug 25 2022, 23:51:50) [MSC v.1916 64 bit (AMD64)]
nympy version 1.21.5
scipy verions 1.9.1
```

roryyorke:
#### Reference issue
Closes gh-22940.

The remainder operation performed in `sawtooth` mapped -1e-16 (and smaller magnitude values) to `2*pi`, resulting in the check `tmod < w * 2 * pi` omitting this value, and when `width=1` (the default), giving divide-by-zero in evaluating the `mask3` result for (actually empty) interval [`width*2*pi`, `2*pi`]. [edit: `*`s caused havoc in Markdown, back-ticked them.]

To fix this, I've used `np.fmod` instead instead of `np.mod`, so that the remainder has the sign of the dividend, and there are now four branches in the evaluation of the function to handle the negative values.

I've added a regression test for this case.

#### Additional information
I've also somewhat modernized the code, and documented two pre-existing behaviours:
  - t and width are broadcast
  - where width is out-of-range, output is NaN

I've added tests for both of these.  This could be further improved, e.g., raise a warning on out-of-range width.

Do I need a `.. versionadded` for this?  Seems like a fix, so I assume not.

### DIFF
--- a/scipy/signal/_waveforms.py
+++ b/scipy/signal/_waveforms.py
@@ -5,7 +5,7 @@
 #   Rewrote much of chirp()
 #   Added sweep_poly()
 import numpy as np
-from numpy import asarray, zeros, place, nan, mod, pi, extract, log, sqrt, \
+from numpy import asarray, zeros, place, nan, mod, pi, log, sqrt, \
     exp, cos, sin, polyval, polyint
 
 
@@ -33,8 +33,9 @@ def sawtooth(t, width=1):
         Width of the rising ramp as a proportion of the total cycle.
         Default is 1, producing a rising ramp, while 0 produces a falling
         ramp.  `width` = 0.5 produces a triangle wave.
-        If an array, causes wave shape to change over time, and must be the
-        same length as t.
+        If an array, causes wave shape to change over time.  Must be
+        broadcastable with t.
+        Where width not in interval [0, 1], output will be NaN.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


roryyorke:
reasonably sure I forgot to handle the case of `t`, or elements thereof, being non-finite.  Fix straightforward (final branch is inverse of union of masks used thus far).  Will add a test and fix.

roryyorke:
I've added a test that non-finite t results in NaN, and change `sawtooth` so that this happens.

scipy/signal/_waveforms.py
```diff
@@ -33,8 +33,9 @@ def sawtooth(t, width=1):
         Width of the rising ramp as a proportion of the total cycle.
         Default is 1, producing a rising ramp, while 0 produces a falling
         ramp.  `width` = 0.5 produces a triangle wave.
-        If an array, causes wave shape to change over time, and must be the
-        same length as t.
+        If an array, causes wave shape to change over time.  Must be
+        broadcastable with t.
+        Where width not in interval [0, 1], output will be NaN.
```
DietBru, line 38: A little nitpick:

```suggestion
        If the wave shape should vary over time, an array with different widths can be 
        used. Its shape has to be either the same shape as `t` or be broadcastable 
        with `t` (as in `~numpy.broadcast_arrays`).
        Where `width` is not in the interval of [0, 1], the output will be NaN.
```

Perhaps a simple example illustrating broadcasting would not hurt.

scipy/signal/_waveforms.py
```diff
@@ -52,32 +53,40 @@ def sawtooth(t, width=1):
...
-    t = asarray(t + (w - w))
+    t, w = np.broadcast_arrays(np.asarray(t), np.asarray(width))
+
     y = zeros(t.shape, dtype="d")
 
     # width must be between 0 and 1 inclusive
-    mask1 = (w > 1) | (w < 0)
-    place(y, mask1, nan)
+    maskbadw = ~((0 <= w)  & (w <= 1))
+    y[maskbadw] = np.nan
```
DietBru, line 62: I think, the code is slightly easier to understand if the logic is written like this:
```python
def sawtooth(t, width=1):
    
    t, width = np.broadcast_arrays(np.asarray(t), np.asarray(width))
    with np.errstate(invalid='ignore'):  # t may have nan values
        u = (t % (2*np.pi)) / (2*np.pi)  # remainder of t normalized to range [0, 1]

    y = np.full(t.shape, np.nan)  # invalid values are set to nan
    ii = (u <= width) & (0 < width) & (width <= 1)  # indexes with rising slope
    y[ii] = u[ii] / width[ii]

    jj = (width <= u) & (0 <= width) & (width < 1)  # indexes with falling slope
    y[jj] = (1 - u[jj]) / (1 - width[jj])

    return 2*y - 1
```

But this is a matter of personal taste—I'm fine with either version.

DietBru (commented):
Looks good to me—my comments are mostly cosmetic.

scipy/signal/_waveforms.py
roryyorke: `numpy.arctan2` has the wording below, which is succinct and clear.  I also didn't mention the shape of the output in the docstring update.  What do you think of something on these lines?

```
    If ``x1.shape != x2.shape``, they must be broadcastable to a common
    shape (which becomes the shape of the output).
```

Do we need to elaborate on broadcasting?  It's basic numpy knowledge.  (I'm in favour of terser docs...)

DietBru: That's good too. Maybe adding something like "The output will have the shape ``np.broadcast_shapes(t, w)``" would make it more beginner friendly.

PS: My experience with broadcasting is more or less limited to doing linear algebra wit matrices and tensors. I added the comment, since I find using broadcasting like this not too intuitive.

ev-br:
What is the status here, folks, is this ready to land or is there something to address still?

roryyorke:
There's disagreement about wording in the docstring.

DietBru:
> There's disagreement about wording in the docstring.

So would you like to keep the wording as is, @roryyorke?  Did you have a chance to look at the code simplification, I proposed?

roryyorke:
> So would you like to keep the wording as is, @roryyorke?

I suggested changing args to match numpy.arctan2, you suggested adding more to that.

> Did you have a chance to look at the code simplification, I proposed?

We need to use `fmod`: https://numpy.org/doc/2.3/reference/generated/numpy.fmod.html#numpy-fmod to fix the bug, not `%`, which is equivalent to `mod` https://numpy.org/doc/2.3/reference/generated/numpy.mod.html#numpy.mod  .  As for the scaling-by-2pi change, I'm neutral, but not inclined to do it myself.  Initially I attempted to keep the "feel" of the original, but with needing to now handle 4 branches that didn't really work.

`square`, just below `sawtooth`, should get similar `fmod`ification.

ev-br:
A gentle ping. We kicked the can down the road once for 1.17.0 release for what looked like disagreement about finer details; now that 1.18.0 is happening, is there maybe a way to resolve the disagreements?

PR #22228, opened 2025-01-02
ENH: linalg.lapack: wrap ?sytri and ?hetri

j-bowhay:
#### Reference issue

Closes #21764
#### What does this implement/fix?

As requested in the above issue wraps the functions from https://www.netlib.org/lapack/explore-html/da/dfa/group__hetri.html

PR #23134, opened 2025-06-09
BUG: linalg.lapack: fix incorrectly sized work array for {c,z}syrti

j-bowhay:
https://github.com/scipy/scipy/issues/21562#issuecomment-2953714487 https://github.com/scipy/scipy/pull/22228#discussion_r2134547140

As @ilayn pointed out the work array should be `2*n` for complex variants

https://www.netlib.org/lapack/explore-html/da/dfa/group__hetri_gafbb5b5ed1e4ad17dbc809478165c3d37.html#gafbb5b5ed1e4ad17dbc809478165c3d37

cc @ilayn @rgommers

### DIFF
--- a/scipy/linalg/flapack_sym_herm.pyf.src
+++ b/scipy/linalg/flapack_sym_herm.pyf.src
@@ -1411,24 +1411,39 @@ subroutine <prefix2c>heequb(lower,n,a,lda,s,scond,amax,work,info)
 
 end subroutine <prefix2c>heequb
 
-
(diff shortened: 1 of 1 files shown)


ilayn:
Thanks @j-bowhay This looks good to me. I couldn't find any issues with the `he` one but at least this should fix the `sy` crashes. Strange that they did not crash on our CI but on WoA runs.

rgommers (approved):
LGTM too, let's get this in so we can check it on a rebased `win_arm64` CI run. Thanks both!

PR #23138, opened 2025-06-09
MAINT: sparse: remove python 2 remnants from `_spbase`

jorenham:
This removes the Python 2 `__nonzero__` and `__[{i,r}]div__` dunder methods from `sparse._spbase`, and cleans up `_spbase._divide`.

Note that `np.true_divide` is an alias of `np.divide`, so the `true_divide` kwarg wasn't really doing anything.

---

The remaining `rdivide` kwarg of `_divide` does not seem to be used. So while I'm at it, I could remove it, or use it in `__rtruediv__`?

### DIFF
--- a/scipy/sparse/_base.py
+++ b/scipy/sparse/_base.py
@@ -440,7 +440,6 @@ class _spbase(SparseABC):
         else:
             raise ValueError("The truth value of an array with more than one "
                              "element is ambiguous. Use a.any() or a.all().")
-    __nonzero__ = __bool__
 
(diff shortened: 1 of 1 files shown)


perimosocordiae:
re: rdivide, I'm ok with merging this as-is and then deciding whether to clean up that code path later.

perimosocordiae (approved):

dschult (commented):
This LGTM as is... Thanks!

dschult (approved):

PR #23140, opened 2025-06-09
TST: Fix missing exceptions

emmanuel-ferdman:
This small PR fixes missing exceptions in `scipy/linalg/tests/test_decomp_update.py`.


tylerjereddy (approved):
Full testsuite and linting seems "ok" locally on this branch--I don't think we need to flush the full CI matrix here.

Changes seem obviously correct with the original `ValueError`s "dangling."  How did you run into this in practice? A linter of some sort, reading the code, or?

PR #23146, opened 2025-06-10
DOC: stats.Mixture: add example

longyuxi:
#### Reference issue

None.

Added two usage examples to the documentation of [stats.Mixture](https://scipy.github.io/devdocs/reference/generated/scipy.stats.Mixture.html).

#### Additional information

None.

### DIFF
--- a/scipy/stats/_distribution_infrastructure.py
+++ b/scipy/stats/_distribution_infrastructure.py
@@ -5100,6 +5100,30 @@ class Mixture(_ProbabilityDistribution):
     .. [1] Mixture distribution, *Wikipedia*,
            https://en.wikipedia.org/wiki/Mixture_distribution
 
+
+    Examples
+    --------
+    A mixture of normal distributions:
+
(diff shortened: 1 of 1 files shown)


mdhaber:
The second commit would not have fixed the problem, so I went ahead and force-pushed the commit I had been preparing.

One issue was PEP8 line length, the other was the order of the `print` statement and results - the output had to come immediately after the `print` line.

I removed the second example because it added length and some complexity but not much about the use of the `Mixture` class itself. Also, it was a very unusual distribution - without reading carefully, it looked like the output was wrong somehow.

I also went ahead and simplified the first example and made it a bit more visually pleasing (to me, anyway - I think the contributions of the original distributions should be visually obvious in  the mixture). I'll merge when CI finishes.

longyuxi:
Sounds good. Thank you! @mdhaber

mdhaber:
Thanks for the PR. Not sure how this ended up without and example.

PR #23151, opened 2025-06-11
CI: use 4 cores on GitHub Actions

rgommers:
Windows and Linux runners were upgraded from 2 to 4 cores last year, see https://github.blog/news-insights/product-news/github-hosted-runners-double-the-power-for-open-source/

Note: this uses -j3 instead of -j4` to avoid a failure with 60 sec timeout for `TestBasinHopping::test_all_nograd_minimizers`. Scaling beyond 2 or 3 parallel processes with `pytest-xdist` can result in little to no improvements or even a degradation sometimes, it really depends.

The timing gains aren't large it looks like, but it should help and also clean up the config a bit.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -140,7 +140,7 @@ jobs:
     - name: Test SciPy
       run: |
         export OMP_NUM_THREADS=2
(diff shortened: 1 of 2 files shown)


lucascolley (approved):
thanks Ralf!

PR #23153, opened 2025-06-11
TST: optimize: cap pytest-run-parallel memory usage

crusaderky:
When running `pytest --parallel-threads=32` on my box, this single test causes a 13~14 GiB spike in memory usage. It's best to cap it.

### DIFF
--- a/scipy/optimize/tests/test_least_squares.py
+++ b/scipy/optimize/tests/test_least_squares.py
@@ -400,6 +400,7 @@ class BaseMixin:
                                 method=self.method)
             assert_allclose(res.x, x_opt)
 
+    @pytest.mark.parallel_threads(4)  # 0.4 GiB per thread RAM usage
     @pytest.mark.fail_slow(5.0)
     def test_workers(self):
         serial = least_squares(fun_trivial, 2.0, method=self.method)


scipy/optimize/tests/test_least_squares.py
```diff
@@ -400,6 +400,7 @@ def test_convergence_with_only_one_tolerance_enabled(self):
                                 method=self.method)
             assert_allclose(res.x, x_opt)
 
+    @pytest.mark.parallel_threads(4)  # 0.4 GiB per thread RAM usage
```
tylerjereddy, line 403: This is probably sensible enough. Note that when I run this test locally on `main` on x86_64 Linux with Python `3.13.0` free-threaded via `spin test -t scipy/optimize/tests/test_least_squares.py -- -k "test_workers" --parallel-threads=32`, I actually see a `multiprocessing` error about too many open files, and sometimes a segfault too.

So, that seems like two reasons for this patch I suppose (the errors go away with the thread limiting as well).

tylerjereddy (approved):

scipy/optimize/tests/test_least_squares.py
tylerjereddy: Presumably mixing `Pool()` with free-threading may cause some resource issues with sufficiently high free-threading concurrency.

PR #23155, opened 2025-06-11
ENH: Add support for 64 bit indices in `sparse.csgraph.minimum_spanning_tree`

### LINKED ISSUE #20817: BUG: 64bit indices not supported by sparse.csgraph.minimum_spanning_tree
### Describe your issue.

As noted in #18716, due to #18509 users are (since 1.11) quite likely to hit the problem that many sparse csgraph routines don't support 64-bit indices.  For the specific case of connected_components this was fixed in #18913, but there are still other problematic routines, e.g. minimum_spanning_tree.

### Reproducing Code Example

```python
import numpy as np, scipy.sparse
# Using astype("i4") below works, and i8 worked before scipy 1.11.
adj = scipy.sparse.coo_array((np.random.rand(5), np.random.randint(0, 9, (2, 5)).astype("i8")), (9, 9))
sparse.csgraph.minimum_spanning_tree(adj)
```

### Error message

```shell
Cell In[23], line 1
----> 1 sparse.csgraph.minimum_spanning_tree(adj)

File _min_spanning_tree.pyx:109, in scipy.sparse.csgraph._min_spanning_tree.minimum_spanning_tree()

ValueError: Buffer dtype mismatch, expected 'ITYPE_t' but got 'long'
```

```shell
1.11.3 1.26.4 sys.version_info(major=3, minor=12, micro=2, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    detection method: pkgconfig
    found: true
    include directory: /opt/arm64-builds/include
    lib directory: /opt/arm64-builds/lib
    name: openblas
      NO_LAPACK= NO_LAPACKE= NO_AFFINITY=1 USE_OPENMP= SANDYBRIDGE MAX_THREADS=3
    pc file directory: /opt/arm64-builds/lib/pkgconfig
    version: 0.3.21.dev
  lapack:
    detection method: pkgconfig
    found: true
    include directory: /opt/arm64-builds/include
    lib directory: /opt/arm64-builds/lib
    name: openblas
      NO_LAPACK= NO_LAPACKE= NO_AFFINITY=1 USE_OPENMP= SANDYBRIDGE MAX_THREADS=3
    pc file directory: /opt/arm64-builds/lib/pkgconfig
    version: 0.3.21.dev
  pybind11:
    detection method: config-tool
    include directory: unknown
    name: pybind11
    version: 2.11.0
Compilers:
  c:
    commands: cc
    linker: ld64
    name: clang
    version: 14.0.0
  c++:
    commands: c++
    linker: ld64
    name: clang
    version: 14.0.0
  cython:
    commands: cython
    linker: cython
    name: cython
    version: 0.29.36
  fortran:
    commands: gfortran
    linker: ld64
    name: gcc
    version: 12.1.0
  pythran:
    include directory: ../../pip-build-env-9n_61ofj/overlay/lib/python3.12/site-packages/pythran
    version: 0.14.0
Machine Information:
  build:
    cpu: aarch64
    endian: little
    family: aarch64
    system: darwin
  cross-compiled: false
  host:
    cpu: aarch64
    endian: little
    family: aarch64
    system: darwin
Python Information:
  path: /private/var/folders/76/zy5ktkns50v6gt5g8r0sf6sc0000gn/T/cibw-run-dp6he_rc/cp312-macosx_arm64/build/venv/bin/python
  version: '3.12'
```

czgdp1807:
#### Reference issue

Closes https://github.com/scipy/scipy/issues/20817

I have implemented overloading of `minimum_spanning_tree` for `np.int64`. It uses a similar approach as used in `_traversal.pyx`. 

Right now I had to use `_min_spanning_tree2` called from `_min_spanning_tree` because otherwise cython was giving me `ambigious overload error` (followed the pattern in `_traversal.pyx`) and it worked out.

Binary size comparison,

On `main`,

```zsh
(scipy-dev) 1:29:25:~/Quansight/scipy/build-install/usr/lib/python3.13/site-packages/scipy/sparse/csgraph % ls -l _min_spanning_tree.cpython-313-darwin.so
-rwxr-xr-x  1 czgdp1807  staff  115048 Jun 12 01:27 _min_spanning_tree.cpython-313-darwin.so
```

On this PR,

```zsh
(scipy-dev) 2:37:33:~/Quansight/scipy/build-install/usr/lib/python3.13/site-packages/scipy/sparse/csgraph % ls -l _min_spanning_tree.cpython-313-darwin.so
-rwxr-xr-x  1 czgdp1807  staff  153080 Jun 12 02:36 _min_spanning_tree.cpython-313-darwin.so
```

An increase of 33% for `_min_spanning_tree.cpython-313-darwin.so `.

Full wheels,

```zsh
(scipy-dev) 3:00:12:~/Quansight/scipy % ls -l *.whl
-rw-r--r--  1 czgdp1807  staff  20236016 Jun 12 02:59 scipy-1.17.0.dev0-cp313-cp313-macosx_15_0_arm64-mst_i32_i64.whl # this PR
-rw-r--r--  1 czgdp1807  staff  20219016 Jun 12 03:00 scipy-1.17.0.dev0-cp313-cp313-macosx_15_0_arm64.whl # main
```

**0.084 %** for overall SciPy wheel.

I think the increase is very small when we look from the perspective of full SciPy wheel size.

#### Additional information

I tried benchmarking but seems like there no benchmarks for `minimum_spanning_tree` in SciPy. Should we add one?

cc: @rgommers @perimosocordiae

### DIFF
--- a/scipy/sparse/csgraph/tests/test_spanning_tree.py
+++ b/scipy/sparse/csgraph/tests/test_spanning_tree.py
@@ -1,5 +1,6 @@
 """Test the minimum spanning tree function"""
 import numpy as np
+import pytest
 from numpy.testing import assert_
 import numpy.testing as npt
 from scipy.sparse import csr_array
@@ -64,3 +65,32 @@ def test_minimum_spanning_tree():
 
         npt.assert_array_equal(mintree.toarray(), expected,
             'Incorrect spanning tree found.')
+
+@pytest.mark.parametrize("dtype", [np.int32, np.int64])
+def test_mst_with_various_index_dtypes(dtype):
+    # Row indices
+    indptr = np.array([0, 2, 4, 5], dtype=dtype)
+    indices = np.array([1, 2, 0, 2, 1], dtype=dtype)
+    data = np.array([2, 0, 2, 3, 3], dtype=float)
+
+    graph = csr_array((data, indices, indptr), shape=(3, 3))
+
+    # Check whether the dtype of indices is as expected
+    assert graph.indices.dtype == dtype
+    assert graph.indptr.dtype == dtype
+
+    # Compute MST
+    mst = minimum_spanning_tree(graph)
+
+    # Check whether the dtype of indices is as expected
+    assert mst.indices.dtype == dtype
+    assert mst.indptr.dtype == dtype
+
+    # Expected MST has 2 edges: (0->1, weight 2)
+    expected = np.array([
+        [0, 2, 0],
+        [0, 0, 0],
+        [0, 0, 0]
+    ], dtype=float)
+
+    npt.assert_array_almost_equal(mst.toarray(), expected)
--- a/scipy/sparse/csgraph/tests/test_spanning_tree.py
+++ b/scipy/sparse/csgraph/tests/test_spanning_tree.py
@@ -3,7 +3,7 @@ import numpy as np
 import pytest
 from numpy.testing import assert_
 import numpy.testing as npt
-from scipy.sparse import csr_array
+from scipy.sparse import csr_array, coo_array
 from scipy.sparse.csgraph import minimum_spanning_tree
 
 
@@ -68,6 +68,7 @@ def test_minimum_spanning_tree():
 
 @pytest.mark.parametrize("dtype", [np.int32, np.int64])
 def test_mst_with_various_index_dtypes(dtype):
+    #CSR array
     # Row indices
     indptr = np.array([0, 2, 4, 5], dtype=dtype)
     indices = np.array([1, 2, 0, 2, 1], dtype=dtype)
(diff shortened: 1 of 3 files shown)


scipy/sparse/csgraph/tests/test_spanning_tree.py
```diff
@@ -64,3 +65,62 @@ def test_minimum_spanning_tree():
...
+
+    # Expected MST will include all 5 edges since graph is disconnected elsewhere
+    expected = np.zeros((9, 9), dtype=float)
+    expected[0, 1] = 1.0
+    expected[1, 2] = 4.0
+    expected[2, 3] = 3.0
+    expected[3, 4] = 2.0
+    expected[4, 5] = 5.0
+
+    npt.assert_array_almost_equal(mst.toarray(), expected)
```
tylerjereddy: Minor point -- here and above, any objection to using `assert_allclose()` as the docs for this NumPy function indicate is preferred?

I realize we still use old way in older code, but maybe in new test code we can start using the new way.

tylerjereddy (commented):
I just did a quick pass late this evening. Quick review notes:

- the new test fails on `main` for the `np.int64` case and passes here, so that's good
- adding some basic `asv` benchmarking might be nice since it is completely missing for this functionality as you say; that said, I don't think we've typically blocked PRs on lack of `asv` benchmarks 
- in principle, the fact that you've not mutated any current regression tests and only added a new one should make this easier for a reviewer to be comfortable with I'd think
- I have at least one question--is there a practical example of needing the extra dtype support that clearly motivates the extra complexity and (admittedly very small) binary size increase? Or is this a regression and it used to be supported? The ticket isn't super clear to me on these matters from a quick look.

scipy/sparse/csgraph/tests/test_spanning_tree.py
czgdp1807: I addressed this review. Now I am using `xp_assert_close` (I guess its most generic and reliable).

PR #23156, opened 2025-06-12
MAINT: _lib: remove dead code

j-bowhay:
As the title suggests, it looks like this utility function is completely unused


dschmitz89:
Thanks!

PR #23159, opened 2025-06-13
ENH: integrate: finite horizon RDE solver

### LINKED ISSUE #21237: ENH: Riccati Differential Equation

Hey Guys! 

My feature request regards the most common type of optimal control problem: the continuous LQR. When solving a continuous LQR problem with finite horizon and boundary conditions, it is necessary to solve a Riccati Differential Equation. I've noticed that scipy only provides solutions for the case with infinite horizon where the equation become static ([scipy.linalg.solve_continuous_are](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.solve_continuous_are.html#scipy.linalg.solve_continuous_are)). 

However, this doesn't really represent most of the practical uses of the continuous problem. Moreover, since there is an official implementation of the solution of the RDE for matlab (https://www.mathworks.com/matlabcentral/fileexchange/104030-solve-riccati-differential-equation-solve_riccati_ode), I just thought it could be a good idea to also have it in scipy.  

I have most of my background in control theory and sometimes I feel like this area needs some love from scipy maintaners. I'd love to contribute, even though I don't know how.

P.S.: two years ago, someone in stackoverflow was also looking for an implementation of this in python (https://stackoverflow.com/questions/72055651/how-to-solve-matrix-riccati-ode-in-python). 

Regards!

_No response_

_No response_

_No response_

DVarela56:
Implements a numerical solver for finite-horizon Riccati Differential Equations using `integrate.solve_ivp` to integrate the RDE backward in time. Includes tests for output validation and batch input support.

#### Reference issue
Closes gh-21237

This PR adds a numerical solver for finite-horizon Riccati Differential Equations (RDEs), using scipy.integrate.solve_ivp to integrate the equations backward in time. The solver supports both single and batched inputs, with time-varying coefficient matrices and final-time boundary conditions. It includes input validation and test coverage to verify correctness and stability of the results.

#### Additional information

Although we couldn’t fully verify the results, we include checks to ensure that the solution matrices remain symmetric, and, when the state-control cross-weight S = 0 is used, that the solution is also semidefinite. This solver can be useful in control theory applications, such as solving continuous-time finite-horizon LQR problems with terminal constraints.

### DIFF
--- a/scipy/linalg/_solvers.py
+++ b/scipy/linalg/_solvers.py
@@ -25,7 +25,7 @@ from ._special_matrices import block_diag
 __all__ = ['solve_sylvester',
            'solve_continuous_lyapunov', 'solve_discrete_lyapunov',
            'solve_lyapunov',
-           'solve_continuous_are', 'solve_discrete_are']
+           'solve_continuous_are', 'solve_discrete_are', 'solve_continuous_rde']
 
 
 @_apply_over_batch(('a', 2), ('b', 2), ('q', 2))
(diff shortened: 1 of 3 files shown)


ilayn:
Thanks @DVarela56 for this one. We have to do a few things before we do more work on this, one is that this probably should live in integrate and not linalg as this is actually an IVP after all. Second this method typically is not suitable for even moderate systems. Hence we need to change the solution method underneath if we want this to go into SciPy.

The classical way to deal with this is the Davison-Maki method. But Riccati-Lyapunov folks have worked on this for some time. An example you can consult is the `linalg.expm` based algorithms in https://arxiv.org/abs/1910.13362. Unfortunately, their code is GPL hence we can't read it but we can contact and ask them if they are willing to relicense it.

DVarela56:
Thanks for the feedback!

I’ll move the implementation to scipy.integrate.

Regarding the numerical method, I understand the limitations of the current IVP-based approach for larger systems. I’ll start looking into the Davison–Maki method and work on a new implementation based on that. I also appreciate the pointer to the matrix exponential approach, I’ll keep it in mind as a potential alternative.

PR #23173, opened 2025-06-18
ENH/DOC: signal.hilbert:  Refactor to avoid extra scaling array and improve docstr 

JesseLivezey:
Added a `workers` argument to `signal.hilbert()` that gets passed into the underlying `ifft/fft` calls.  I see reasonable speedups when using this in cases where I am doing many transforms in one call. I modified a test to run with the default and 2 workers. Is there a reason why `workers` couldn't be added as an argument to any function in `signal` that uses scipy's fft functions?

I also refactored the scaling so that it is done in-place on the fft coefficients rather than allocating an extra scaling array. I think this change should be reasonably well covered by existing tests, but added a 3d example. I see a small speedup on arrays that are fairly large on the transform axis. Would it be helpful to add an `asv` benchmark for `hilbert`?

I also tried to clarify a few parts of the docstring that I found unclear, but couldn't build them locally to check for formatting issues. I see
```
ERROR: [numpydoc] While processing docstring for 'scipy.interpolate.BSpline'
```

### DIFF
--- a/scipy/signal/_signaltools.py
+++ b/scipy/signal/_signaltools.py
@@ -2473,7 +2473,7 @@ def deconvolve(signal, divisor):
     return quot, rem
 
 
-def hilbert(x, N=None, axis=-1):
+def hilbert(x, N=None, axis=-1, workers=None):
     r"""FFT-based computation of the analytic signal.
 
     The analytic signal is calculated by filtering out the negative frequencies and
@@ -2491,6 +2491,10 @@ def hilbert(x, N=None, axis=-1):
         Number of Fourier components.  Default: ``x.shape[axis]``
     axis : int, optional
         Axis along which to do the transformation.  Default: -1.
+    workers : int, optional
+        Maximum number of workers to use for parallel computation. If negative,
+        the value wraps around from ``os.cpu_count()``. See ``scipy.fft.fft()``
+        for more details.
 
     Returns
     -------
@@ -2577,7 +2581,7 @@ def hilbert(x, N=None, axis=-1):
     if N <= 0:
         raise ValueError("N must be positive.")
 
-    Xf = sp_fft.fft(x, N, axis=axis)
+    Xf = sp_fft.fft(x, N, axis=axis, workers=workers)
     h = xp.zeros(N, dtype=Xf.dtype)
     if N % 2 == 0:
         h[0] = h[N // 2] = 1
@@ -2590,7 +2594,7 @@ def hilbert(x, N=None, axis=-1):
         ind = [xp.newaxis] * x.ndim
         ind[axis] = slice(None)
         h = h[tuple(ind)]
-    x = sp_fft.ifft(Xf * h, axis=axis)
+    x = sp_fft.ifft(Xf * h, axis=axis, workers=workers)
     return x
 
 
--- a/scipy/signal/_signaltools.py
+++ b/scipy/signal/_signaltools.py
@@ -2491,8 +2491,8 @@ def hilbert(x, N=None, axis=-1):
     x : array_like
         Signal data.  Must be real.
     N : int, optional
-        Number of Fourier components. ``x`` is cropped or zero-padded to length
-        ``N`` along ``axis``.  Default: ``x.shape[axis]``
+        Number of output samples. `x` is initially cropped or zero-padded to length
+        `N` along `axis`.  Default: ``x.shape[axis]``
     axis : int, optional
         Axis along which to do the transformation.  Default: -1.
 
--- a/scipy/signal/_signaltools.py
+++ b/scipy/signal/_signaltools.py
@@ -2476,7 +2476,7 @@ def deconvolve(signal, divisor):
 def hilbert(x, N=None, axis=-1, workers=None):
     r"""FFT-based computation of the analytic signal.
 
-    The analytic signal is calculated by filtering out the negative frequencies and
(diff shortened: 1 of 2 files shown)


JesseLivezey:
>Is there a reason why workers couldn't be added as an argument to any function in signal that uses scipy's fft functions?

I guess the torch array failures are probably an answer to this. Is this a reasonable way to special-case numpy arrays?

```python
if not is_numpy(xp):
    workers_dict = dict()
else:
    workers_dict = dict(workers=workers)
````

**Comment by lucascolley** [comment]
Yes, looks good. Please also add to the docstring that the parameter is NumPy only

**Comment by DietBru** [comment]
Wouldn't using [set_workers] achieve the same goal? I.e.:
```python
from scipy.signal import hilbert
from scipy.fft import set_workers

x = [1, 0, -1, 0]
with set_workers(8):
    z = hilbert(x)
```
I would prefer this solution over adding an extra parameter to `hilbert()`. The reason: Especially in the signal module, FFTs are used in a quite a few functions. For consistency reasons it would make sense to add a `workers` parameter to each of them, which is quite a bit of work. Perhaps adding a section on how to use workers in the [FFT user guide] and link to from the `hilbert` docstr might make this approach more discoverable.

Avoiding the extra array `h` is beneficial, in my opinion.

[set_workers]: https://scipy.github.io/devdocs/reference/generated/scipy.fft.set_workers.html
[FFT user guide]: https://scipy.github.io/devdocs/tutorial/fft.html

scipy/signal/_signaltools.py
```diff
@@ -2577,20 +2582,20 @@ def hilbert(x, N=None, axis=-1):
...
     if N % 2 == 0:
-        h[0] = h[N // 2] = 1
-        h[1:N // 2] = 2
+        Xf[tuple(ind + [slice(1, N // 2)])] *= 2.0
+        Xf[tuple(ind + [slice(N // 2 + 1, N)])] = 0.0
     else:
-        h[0] = 1
-        h[1:(N + 1) // 2] = 2
+        Xf[tuple(ind + [slice(1, (N + 1) // 2)])] *= 2.0
+        Xf[tuple(ind + [slice((N + 1) // 2, N)])] = 0.0
```
DietBru: A cosmetic suggestion: Utilizing `np.moveaxis` makes the code a little more succinct, i.e.: 

```python
Xf = sp_fft.fft(x, N, axis=axis)
Xf1 = np.moveaxis(Xf, axis, -1)  # view on Xf
i = N // 2 if N % 2 == 0 else (N + 1) // 2 
Xf1[..., 1:i] *= 2.0
Xf1[..., i:N] = 0.0
```

JesseLivezey: In the even case, the end and start values are different by 1, so I don't think it can clean up quite this nicely. `moveaxis` does still make it cleaner.

JesseLivezey:
@DietBru, I wasn't aware of the `set_workers` function! Agree that documenting it in the signal functions that use ffts is better than adding an argument everywhere.

scipy/signal/_signaltools.py
```diff
@@ -2476,19 +2476,23 @@ def deconvolve(signal, divisor):
...
+    workers used for the FFTs.
+
     Parameters
     ----------
     x : array_like
         Signal data.  Must be real.
     N : int, optional
-        Number of Fourier components.  Default: ``x.shape[axis]``
+        Number of Fourier components. ``x`` is cropped or zero-padded to length
+        ``N`` along ``axis``.  Default: ``x.shape[axis]``
```
DietBru: A cosmetic remark: I think denoting `N` as number of Fourier components is not precise, since `x_a` has at most ``N//2 + 1`` non-zero Fourier components. Furthermore, the FFT is shortened or zero-padded.
What about something like:
```suggestion
        Number of output samples. The output length is adjusted by cropping or or zero-padding
        the FFT of `x` accordingly. Default: ``x.shape[axis]``
```

PS: In docstrs function parameters are enclosed by single ticks, e.g., `` `x` `` and non-parameters and code by two single ticks, e.g., ``` ``x.shape[axis]`` ```.

DietBru (changes requested):
It looks good now. I only have one little  remark in the docstr..

The CI failure is unrelated and is addressed in PR #23199.

scipy/signal/_signaltools.py
JesseLivezey: Agreed on the original version being imprecise. I updated to something similar to your suggestion, although I believe `fft()` pads the original signal, not the fourier coefficients.
[
Docs](https://output.circle-artifacts.com/output/job/870fee4e-e67a-416e-b886-527dbab091f0/artifacts/0/html/reference/generated/scipy.signal.hilbert.html) seem to render/link correctly.

DietBru (approved):
Looks good now.

scipy/signal/_signaltools.py
DietBru: Regarding the padding: you're of course correct. I didn't read carefully and thought it was done like in the [envelope] function, were the cropping/zero-padding is done in the Fourier space.

[envelope]: https://scipy.github.io/devdocs/reference/generated/scipy.signal.envelope.html

DietBru: I like you're corrected version.

DietBru:
CI Test failure still unrelated, hence merging...
Thanks for your work, @JesseLivezey!

Issue #23177, opened 2025-06-18
BUG: `stats`: regression in JAX 0.6.2


crusaderky:
### Describe your issue.

A test started failing after upgrading from jax 0.6.1 to 0.6.2.

Note that scipy CPU CI uses pypi, which automatically pulls in 0.6.2, whereas both scipy GPU CI and https://github.com/rgommers/pixi-dev-scipystack use conda, which is still at 0.6.0 even after you run `pixi update jax jaxlib`.

### Reproducing Code Example

```python
To reproduce locally with https://github.com/rgommers/pixi-dev-scipystack:


diff --git a/scipy/pixi.toml b/scipy/pixi.toml
index 0e6e5a8..0e4252d 100644
--- a/scipy/pixi.toml
+++ b/scipy/pixi.toml
@@ -173,9 +173,9 @@ test-cupy = { cmd = "spin test -b cupy -m 'array_api_backends and not slow'", cw
 # Windows support pending: https://github.com/conda-forge/jaxlib-feedstock/issues/161
 platforms = ["linux-64", "osx-arm64"]
 
-[feature.jax-cpu.dependencies]
-jax = ">=0.6.0"
-jaxlib = { version = "*", build = "*cpu*" }
+[feature.jax-cpu.pypi-dependencies]
+jax = "==0.6.2"
+jaxlib = "==0.6.2"
 
 [feature.jax-cuda]
 platforms = ["linux-64"]


pixi r test-jax -t scipy.stats -- -k test_ddof_equal_to_number_of_observations

The above passes with 0.6.0 (conda) and 0.6.1 (pixi), but fails with 0.6.2.
```

### Error message

```shell
e.g. https://github.com/scipy/scipy/actions/runs/15727179767/job/44319546946?pr=23175


________ TestGSTD.test_ddof_equal_to_number_of_observations[jax.numpy] _________

self = <scipy.stats.tests.test_stats.TestGSTD object at 0x7f93248e73e0>
xp = <module 'jax.numpy' from '/opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/site-packages/jax/numpy/__init__.py'>

    def test_ddof_equal_to_number_of_observations(self, xp):
        x = xp.asarray(self.array_1d)
        res = stats.gstd(x, ddof=x.shape[0])
>       xp_assert_equal(res, xp.asarray(xp.nan))
E       AssertionError: 
E       Arrays are not equal
E       
E       nan location mismatch:
[...]

PR #23178, opened 2025-06-18
CI: skip JAX 0.6.2

crusaderky:
https://github.com/scipy/scipy/issues/23177
Unblock CI


crusaderky:
CI failure is unrelated

rgommers (approved):
Thanks @crusaderky! 

@tylerjereddy this can be backported to avoid the same failure on the 1.16.x branch.

PR #23179, opened 2025-06-18
TST: ndimage: avoid implicit dtype conversions in vectorized_filter tests

ev-br:
Here's a test which does   `vectorized_filter(input, xp.sum, output)`, with output being `xp.empty_like(input)`. This is problematic for short integers because
- the output of xp.sum(int8_array) is int64 (the default int dtype), and
- internally, vectorized_filter does roughly, `output[some_indices] = xp.sum(input[some_indices])`

So what happens here is that the output of the `sum` in the r.h.s. gets silently downcast to the dtype of the l.h.s. by `__setitem__`.

This kind of sort of works in many array libraries, but is technically unspecified by the Array API standard. And indeed this starts failing with array-api-strict in https://github.com/data-apis/array-api-strict/pull/157

Thus, a test only change to make downcasting explicit.

### DIFF
--- a/scipy/ndimage/tests/test_filters.py
+++ b/scipy/ndimage/tests/test_filters.py
@@ -2864,7 +2864,14 @@ class TestVectorizedFilter:
             assert res.dtype == sum_dtype
 
             output = xp.empty_like(input)
-            res = ndimage.vectorized_filter(input, xp.sum, output=output, **kwargs)
+            res = ndimage.vectorized_filter(
+                input,
+                lambda x, *args, **kw: xp.astype(
+                    xp.sum(x, *args, **kw), x.dtype, copy=False
+                ),
+                output=output,
+                **kwargs
+            )
             xp_assert_close(res, xp.astype(xp.stack(ref), dtype))
             assert res.dtype == dtype
 


ev-br:
`scipy/stats/tests/test_stats.py::TestGSTD::test_ddof_equal_to_number_of_observations[jax.numpy]` CI failure is unrelated.

mdhaber:
Failure is gh-23177, and that seems to be an upstream bug.

By default, the output dtype is chosen as that of `function(input[some_indices])`, so I think this would only be a problem when either:
- `output` is specified and is of a different dtype than `function(input[some_indices])` OR
- `function(input)` is inconsistent about its output dtype (e.g. dtype is data-dependent)

But it looks like we're passing `out` in that test, so I see how that could trigger complaints about implicit dtype conversions. It looks like the diff will fix that.

---

This brings up that the documentation is not exactly right, specifically the parameter description:
image

I think that should read:
> By default, an array of the dtype returned by `function` will be created.

(Technically, it the dtype of `function(input[some_indices])`, but I think it's OK to gloss over that fine point.)

mdhaber (approved):

PR #23180, opened 2025-06-18
CI: remove free-threading workarounds in wheel builds

rgommers:
These are no longer needed since Cython, Pythran and NumPy all have public releases that support free-threading now.

Also include two small dependency bumps: 
- numpy: 2.0.0rc1 -> 2.0.0
- cibuildwheel 3.0.0rc3 -> 3.0.0 

I did a full set of wheel builds on my fork, all green: [CI log](https://github.com/rgommers/scipy/actions/runs/15734610009).

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -212,7 +212,7 @@ jobs:
           echo "CIBW_BUILD_FRONTEND=$CIBW" >> "$GITHUB_ENV"
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@3c5ff0988806752c5a6502c845f9edc2d98095d6 # v3.0.0rc3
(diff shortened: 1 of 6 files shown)


rgommers:
The one failure is unrelated: `TestGSTD.test_ddof_equal_to_number_of_observations[jax.numpy]` is also failing on other PRs right now.

This is good to merge.

andyfaff:
LGTM. Thanks @rgommers

rgommers:
@tylerjereddy I've marked this for backporting. You can leave it out for 1.16.0 final if you prefer, it's not critical. It is an improvement though, prevents picking up dev versions of dependencies in wheels builds - and also it will be needed on the 1.16.x branch before we add Python 3.14 support there (PR for that is coming as a follow-up to this one).

tylerjereddy:
Looks like a clean win here, I'll backport it. We have slight release delay at the moment anyway and you're already helping me with that.

PR #23181, opened 2025-06-18
DOC: spatial: Rigid Transformation docs cleanup

scottshambaugh:
Some of the rigid transformation docs weren't rendering as intended, this fixes that and IMO gives a more intuitive matrix representation.

**Before:** 
image

**After:**
image

### DIFF
--- a/scipy/spatial/transform/_rigid_transform.py
+++ b/scipy/spatial/transform/_rigid_transform.py
@@ -404,17 +404,17 @@ class RigidTransform:
 
         Notes
         -----
-        4x4 rigid transformation matrices are of the form:
+        4x4 rigid transformation matrices are of the form::
 
-        ..
+            [       tx]
+            [   R   ty]
+            [       tz]
+            [ 0 0 0  1]
 
-            [R | t]
-            [0 | 1]
-
-        where ``R`` is a 3x3 rotation matrix and ``t`` is a 3x1 translation
-        vector ``[tx, ty, tz]``. As rotation matrices must be proper
-        orthogonal, the rotation component is orthonormalized using singular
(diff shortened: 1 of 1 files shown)


scottshambaugh:
CI test failure unrelated, these are only docs changes.

scipy/spatial/transform/_rigid_transform.pyx
```diff
@@ -536,17 +536,17 @@ cdef class RigidTransform:
...
-            [0 | 1]
-
-        where ``R`` is a 3x3 rotation matrix and ``t`` is a 3x1 translation
-        vector ``[tx, ty, tz]``. As rotation matrices must be proper
-        orthogonal, the rotation component is orthonormalized using singular
-        value decomposition before initialization.
+        where ``R`` is a 3x3 rotation matrix and ``t = [tx, ty, tz]`` is a 3x1
+        translation vector. As rotation matrices must be proper orthogonal, the
+        rotation component is orthonormalized using singular value decomposition
+        before initialization.
```
j-bowhay: May as well fix a typo too while we're here
```suggestion
        rotation component is orthonormalized using the singular value decomposition
        before initialization.
```

scottshambaugh: I think it's more correct as-is, since it's referring to the SVD method instead of its singular result.

j-bowhay (approved):
Thanks, still not entirely convinced by the grammar around the SVD but I won't hold things up over it:)

Can you resolve the merge conflict then I will merge

scottshambaugh:
Rebased!

lucascolley (approved):
thanks both!

PR #23182, opened 2025-06-18
DOC: fft: tutorial improvements

### LINKED ISSUE #23147: DOC: example in the tutorial for FFT
## Problem

The first example with a plot in the tutorial for FFT (https://docs.scipy.org/doc/scipy/tutorial/fft.html) uses `fftfreq` to calculate the frequency points, and then cuts the segment containing non-negative frequencies
> xf = fftfreq(N, T)[:N//2]

This segment does not include `fs/2` (half the sampling frequency), which is not important for the particular signal in this example, but in general it is.

## Fix

Use `rfftfreq`, which is suitable for one-sided spectrum, instead of `fftfreq`, which is more suitable for two-sided spectrum. Also fix the limits in `yf`.

```
xf = rfftfreq(N, T)
plt.plot(xf, 2.0/N * np.abs(yf[:len(xf)]))
```

## Reference

See also the discussion in https://github.com/numpy/numpy/issues/29162

_No response_

_No response_

georgeef:
#### Reference issue
Implements the improvements discussed in #23147.

The section "Fast Fourier transforms" -> "1-D DFT" is completely re-written. The main changes are:
* Corrected the use of `fftfreq`: changed to `rfftfreq` which is more suitable for one-sided spectrum.
* Extended the text with more clarifications.
* Extended the example plots with more details.
* Moved the description of spectral leakage and windows to a new section "Appendix" at the end.

#### Additional information

Appendix: More advanced topics which do not introduce the main fft functions, have been moved at the end, in order to have more space without disturbing the introduction of the main functions. The appendix could be extended with more topics in the future (I could contribute a new section about the power spectral density of random signals).

Decorations in plots: Several examples contain code which improves the appearance of the plot, or adds more details, but in general it is not useful for a real application. This code is labeled as "decorations" with a comment, it appears at the end of the example code, and it is separated from the main code with a blank line, in order to make easier the copy of the useful part in the example code.

Tight layout in plots: There seems to be a problem with the margins in plots, at least with my local renderer. I use tight layout because it makes the plot a little better, but the renderer cuts the labels at the edges (maybe a bug). Anyway the labels are not important, and the same code in jupyterlab produces correct plots, therefore users should not have problems when they copy code from the examples. I tried without the tight layout option, but it looks worse.

Punctuation after `:math:`: Punctuation (comma or full-stop) immediately after a closing backtick has been moved inside the backticks. Without this change, the rendered (or the browser) allows a line break between the math expression and the punctuation, i.e., the next line may start with a comma of full-stop, which is bad style. 

@lucascolley, @DietBru: FYI

### DIFF
--- a/doc/source/tutorial/fft.rst
+++ b/doc/source/tutorial/fft.rst
@@ -795,9 +795,8 @@ as shown in the following example.
     >>> Ax = np.array([0.2, 0.8, 0.6])               # signal amplitudes
     >>> t = np.linspace(0.0, Tp, N, endpoint=False)  # time in s (one segment)
     >>> f = rfftfreq(M*N, Ts)                        # frequency in Hz
-    >>> x = np.zeros(M*N)
-    >>> x[:N] = np.sum(Ax[:,np.newaxis]*np.cos(2.0*np.pi*np.outer(Fx,t)), axis=0)
-    >>> yr = rfft(x)
+    >>> x = np.sum(Ax[:,np.newaxis]*np.cos(2.0*np.pi*np.outer(Fx,t)), axis=0)
+    >>> yr = rfft(x, M*N)
     >>> ya = 20.0*np.log10(2.0/N * np.abs(yr))
     >>>
     >>> # plot ya vs. f in the range [0, Fs/2]
@@ -858,7 +857,7 @@ two bin frequencies, and the amplitude in these two bins is scaled
 by :math:`2/\pi`, i.e., it appears 4 dB below the actual peak.
 The side peaks are close to integer-plus-half multiples of :math:`F_p`
 relative to :math:`F_x.`
-When the spectrum is plotted in dB, the envelop of the spectral leakage
+When the spectrum is plotted in dB, the envelope of the spectral leakage
 near the main peak takes the form of a negative logarithmic function,
 which falls by 6 dB (or 1 bit in log2) per doubling of the distance
 from the main peak.
@@ -899,8 +898,8 @@ the Blackman window and the Dolph–Chebyshev window with 100 dB attenuation.
     >>> import matplotlib.pyplot as plt
     >>> fig, ax = plt.subplots(figsize=(7,4))
     >>> fig.tight_layout()
-    >>> ax.plot(n, windows.blackman(N), 'g', label='blackman')
-    >>> ax.plot(n, windows.chebwin(N, at=100), 'r', label='cheb(100)')
+    >>> ax.plot(n, windows.blackman(N, sym=False), 'g', label='blackman')
+    >>> ax.plot(n, windows.chebwin(N, at=100, sym=False), 'r', label='cheb(100)')
     >>> ax.legend()
     >>>
     >>> # decorations
@@ -914,7 +913,7 @@ Their spectrum, which shows also their spectral leakage pattern,
 is calculated and plotted with the following code:
 
 .. plot::
-    :alt: "This code generates calculates and plots the spectrum of three windows: (1) the rectangular window, (2) the  Blackman window, and (3) the Dolph–Chebyshev window with 100 dB attenuation. The main axes shows the spectrum at integer-plus-half bin width, which is approximately equal to the envelop of the spectral leakage. An inset axes shows a detail of the spectrum for the first 6 bins."
+    :alt: "This code generates calculates and plots the spectrum of three windows: (1) the rectangular window, (2) the  Blackman window, and (3) the Dolph–Chebyshev window with 100 dB attenuation. The main axes shows the spectrum at integer-plus-half bin width, which is approximately equal to the envelope of the spectral leakage. An inset axes shows a detail of the spectrum for the first 6 bins."
 
     >>> from scipy.fft import rfft, rfftfreq
     >>> from scipy.signal import windows
@@ -922,14 +921,14 @@ is calculated and plotted with the following code:
     >>> M = 10   # number of segments
     >>> N = 400  # window length
     >>> f = rfftfreq(M*N, 1.0/N)  # normalized frequency
-    >>> x = np.zeros((3, M*N))    # space for 3 windows
-    >>> x[0,:N] = np.ones(N)
-    >>> x[1,:N] = windows.blackman(N)
-    >>> x[2,:N] = windows.chebwin(N, at=100)
-    >>> yr = rfft(x, axis=1)
+    >>> x = np.zeros((3, N))      # space for 3 windows
+    >>> x[0,:] = np.ones(N)
+    >>> x[1,:] = windows.blackman(N, sym=False)
+    >>> x[2,:] = windows.chebwin(N, at=100, sym=False)
+    >>> yr = rfft(x, M*N, axis=1)
     >>> eps = np.spacing(1.0)
     >>> ya = 20.0*np.log10(eps + np.abs(yr) / np.sum(x, axis=1, keepdims=True))
-    >>> # envelop of ya
+    >>> # envelope of ya
     >>> fm = f[:-1].reshape(-1,M)
     >>> yam = ya[:,:-1].reshape(3,-1,M)
     >>> ke = np.argmax(yam, axis=-1)
@@ -969,7 +968,7 @@ by the sum of the window sequence).
 The inset shows the amplitude for the first few bins.
 As it is seen, the amplitude has large variation over the continuous frequency,
 due to the zero values at or near every bin frequency.
-The main plot shows the envelop of the amplitude vs. frequency,
+The main plot shows the envelope of the amplitude vs. frequency,
 which is also the worst-case pattern for the spectral leakage.
 
 These three windows are applied to a signal as shown below.
@@ -990,9 +989,9 @@ These three windows are applied to a signal as shown below.
     >>> t = np.linspace(0.0, Tp, N, endpoint=False)  # time in s
     >>> f = rfftfreq(N, Ts)                          # frequency in Hz
     >>> w = np.empty((3, N))                         # space for 3 windows
-    >>> w[0,:N] = np.ones(N)
-    >>> w[1,:N] = windows.blackman(N)
-    >>> w[2,:N] = windows.chebwin(N, at=100)
+    >>> w[0,:] = np.ones(N)
+    >>> w[1,:] = windows.blackman(N, sym=False)
+    >>> w[2,:] = windows.chebwin(N, at=100, sym=False)
     >>> x = w * np.sum(Ax[:,np.newaxis]*np.cos(2.0*np.pi*np.outer(Fx,t)), axis=0)
     >>> yr = rfft(x, axis=1)
     >>> eps = np.spacing(1.0)
--- a/doc/source/tutorial/fft.rst
+++ b/doc/source/tutorial/fft.rst
@@ -8,15 +8,16 @@ Fourier Transforms (:mod:`scipy.fft`)
 .. contents::
 
 
-Fourier analysis is a method for expressing a function as a sum of periodic
-components, and for recovering the signal from those components. When both
-the function and its Fourier transform are replaced with discretized
-counterparts, it is called the discrete Fourier transform (DFT). The DFT has
-become a mainstay of numerical computing in part because of a very fast
-algorithm for computing it, called the Fast Fourier Transform (FFT), which was
-known to Gauss (1805) and was brought to light in its current form by Cooley
-and Tukey [CT65]_. Press et al. [NR07]_ provide an accessible introduction to
-Fourier analysis and its applications.
+Fourier analysis is a method which decomposes a function into periodic
+components, and reconstructs it from those components.
+When both the function and its Fourier transform are discretized,
+it is called the "Discrete Fourier Transform" (DFT).
+The DFT has become a mainstay of numerical computing in part because
+of a fast algorithm for computing it, with complexity
+:math:`O(n\,{\rm log}\,n),` called the "Fast Fourier Transform" (FFT).
+This algorithm was known to Gauss (1805) and was brought to light in its
+current form by Cooley and Tukey [CT65]_. Press et al. [NR07]_ provide an
+accessible introduction to Fourier analysis and its applications.
 
 
 .. _tutorial_FFT:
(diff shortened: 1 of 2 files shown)


georgeef:
Minor corrections:
* Use the argument `n` in `rfft` instead of padding the signal with zeros.
* Add `sym=False` in windows.
* Spelling: envelop -> envelope.

georgeef:
@DietBru 
Something is wrong with `scipy.signal.windows.chebwin(N, at=100, sym=False)`. Its FFT should have almost flat envelope, as it happens with `sym=False`, but the deviation is quite noticeable.

Looking into the implementation, I see that `M` (the wanted length) is first extended by 1, then the spectrum is constructed, then the window function is calculated with an inverse FFT, and finally the result is truncated. This algorithm seems to give a distorted window function. I think the correction for `sym=False` is to keep `M` as it is (no extension, no truncation), since the inverse FFT is already truncated at the end. For `sym=True`, the sub-sequence `w[0], ..., w[M-2]` can be calculated with `M` reduced by 1, and the final result can be extended by appending `w[0]` to make it symmetric. Appending `w[0]` is essentially padding with one more 0, so the spectrum shape does not change, it only slightly scales in the frequency axis. Therefore in either case the FFT of the result has the wanted properties of the intermediate FFT during the construction. (I didn't try to modify the code, and I didn't test if the previous description gives the expected result).

Anyway, I revert the `sym` argument for `chebwin` in the examples. This shows a spectrum closer to the expected.

georgeef:
Another issue: `windows.triang(4, sym=False)` gives
> [1/3, 2/3, 3/3, 2/3]

The standard definition, at least according to Harris (1978), eq. (23b), gives
> [0, 1/2, 2/2, 1/2]

`windows.triang(5, sym=False)` also gives wrong result.

For `sym=True`, the symmetric window should be the non-symmetric window of size `M-1` followed by `w[0]`, which is true, but for the wrong non-symmetric window.

EDIT: I now see that `triang` in Matlab also has a non-standard definition, so it is ok to be compatible with Matlab. On the other hand, Matlab gives only symmetric result.

doc/source/tutorial/fft.rst
```diff
@@ -8,187 +8,228 @@ Fourier Transforms (:mod:`scipy.fft`)
...
+...
+>>> x = np.array([2. , 1.5, 0.5, 0. , 0.5, 1.5])
+>>> y = fft(x)
+>>> y
+array([6.-0.j, 3.+0.j, 0.+0.j, 0.-0.j, 0.-0.j, 3.-0.j])
+>>> ifft(y)  # equals x, but is complex-valued
+array([2. +0.j, 1.5+0.j, 0.5+0.j, 0. +0.j, 0.5-0.j, 1.5+0.j])
+>>> # the following are equal for any complex x:
+>>> np.isclose(ifft(x), fft(x.conj()).conj()/len(x))
+array([ True,  True,  True,  True,  True,  True])
```
DietBru, line 74: This is marginally easier to read:
```suggestion
>>> np.allclose(ifft(x), fft(x.conj()).conj()/len(x))
True
```

doc/source/tutorial/fft.rst
```diff
@@ -8,187 +8,228 @@ Fourier Transforms (:mod:`scipy.fft`)
...
-[WPW]_). Windowing the signal with a dedicated window function helps mitigate
-spectral leakage. The example below uses a Blackman window from scipy.signal
-and shows the effect of windowing (the zero component of the FFT has been
-truncated for illustrative purposes).
+    x(t) = \sum_{k\in\mathbb{Z}} c_k\, e^{j 2 \pi k t}
+
+and the DFT. :math:`x(t)` may be interpreted as a two-dimensional parametric closed
+curve :math:`x: t\in[0, 1] \mapsto \mathbb{C}`, where the real and the imaginary part
+of :math:`x(t)` represent represent the two dimensions. This curve is sampled at
+equistant intervals, i.e., :math:`x[n] := x(n/N)` and the Fourier coefficents
```
DietBru, line 91: Typos:
```suggestion
equidistant intervals, i.e., :math:`x[n] := x(n/N)` and the Fourier coefficients
```

doc/source/tutorial/fft.rst
```diff
@@ -8,187 +8,228 @@ Fourier Transforms (:mod:`scipy.fft`)
...
+
+The samples are depicted as dots and the value of the parameter :math:`t` is encoded in
+the color of the curve. Note that the `~scipy.signal.resample` function, which uses the
+:func:`fft`/:func:`ifft` functions internally, can be used as a more efficient
+alternative for calculating the Fourier expansion for each interpolation value.
+
+
+In the usual case in which :math:`x[n]` is real, the subsequence
+:math:`y[1],\ldots,y[N-1]` is conjugate-symmetric, i.e.,
+:math:`y[N-k] = y^*[k],` for :math:`k=1,\ldots,N-1.`
```
DietBru, line 110: For consistency with above:
```suggestion
:math:`y[N-k] = \overline{y^*[k]},` for :math:`k=1,\ldots,N-1.`
```

doc/source/tutorial/fft.rst
```diff
@@ -8,187 +8,228 @@ Fourier Transforms (:mod:`scipy.fft`)
...
+Therefore :math:`y[N/2]` is real for even :math:`N,`
+while :math:`y[0]` is real for either even or odd :math:`N.`
+The subsequence :math:`y[0],\ldots,y[N//2]` contains :math:`N`
+independent real values, and it is sufficient to reconstruct :math:`x[n].`
+Typically only this subsequence is plotted for real :math:`x[n].`
```
DietBru, line 115: Something along the lines of 
```suggestion
The `rfft` / `irfft` functions are designed for this use case.
```
would make more sense to me.

doc/source/tutorial/fft.rst
```diff
@@ -8,187 +8,228 @@ Fourier Transforms (:mod:`scipy.fft`)
...
+In analogy to :math:`x[n],` the sequence :math:`y[k]` is considered to be
+the samples of a complex function of frequency :math:`y(f),` at a fixed
+frequency interval :math:`F_p,` i.e., :math:`y[k] = y(k\,F_p).`
+The frequencies :math:`k\,F_p` are called "bin frequencies".
+
+The following example models the samples taken from a real signal
+containing two sinusoidal components and a bias (such that its
+spectrum can be easily predicted). It calculates the DFT of the signal
+using :func:`fft`, and then it plots its amplitude vs. frequency.
+The frequency points are caclulated using the convenience function
```
DietBru, line 134: Typo:
```suggestion
The frequency points are calculated using the convenience function
```

doc/source/tutorial/fft.rst
```diff
@@ -8,187 +8,228 @@ Fourier Transforms (:mod:`scipy.fft`)
...
+:func:`rfftfreq`, which takes as arguments the parameters :math:`N`
+and :math:`T_s` and returns the multiples of the bin width :math:`F_p`
+in the range :math:`[0,F_s/2].` These are the frequency points of
+interest for the FFT of a real signal.
```
DietBru, line 139: I am aware that the original tutorial has also comparable plots—but I am not really convinced that this example is useful in this form. In my opinion, it glosses over too many relevant aspects. To make this clearer, I modified the example a little bit: 
image

The signal is $x(t) = \sum a_i \cos(2\pi f_i t)$ with $f_i\in[0, 20, 55]$ Hz and $a_i\in[2, 4, 4]$. 

This plot has the following oddities: 
1. Due to the spectral leakage the heights of the DC-component and the first peak do not have the expected values of $2$ and $4$. 
2. The 55 Hz lobe is not symmetric, even though is lies exactly in the middle of two frequency bins. This is due to the spectral leakage of the negative frequency component.
3. Doubling the magnitude is achieved by adding the negative frequency bins values to the positive ones. The 0-th component and the Nyquist frequency at 100 Hz do not have a negative part (since $N$ is even). Hence they should not be doubled. This can lead to counterintuitive results when oversampling the spectrum by zero-padding. This also explains why here the value at $f=100$ Hz is half the expected height.

In summary: It is not clear from this example how to infer signal properties, like the cosine amplitude, from a spectral plot, IMHO.

The code for the plot:
```python
import numpy as np

from scipy.fft import rfft, rfftfreq

N, fs = 100, 200  # number of samples and sampling frequency in Hz
T = 1 / fs  # sampling interval in s
tau = N * T  # signal duration in s
df = fs / N  # bin width (frequency resolution) in Hz

# Create signal made up of three cosine components:
ff = np.array([0, 20, 55])  # signal frequencies in Hz
aa = np.array([2,  4,  4])  # signal amplitudes
t = np.arange(N) * T  # timestamps in s
x = sum(a_ * np.cos(2*np.pi * f_ * t) for a_, f_ in zip(aa, ff))

# Determine one-sided magnitude spectrum:
f = rfftfreq(N, T)  # frequency in Hz
X = abs(rfft(x) / N)
X[1:(N+1)//2] *= 2  # add values of negative frequencies to positive ones

fg, ax = plt.subplots(1, 1, tight_layout=True)
ax.set_title(f"One-sided Magnitude Spectrum of a {N} sample Signal")
ax.set(xlabel=rf"Frequency in hertz ($\Delta f={df}\,$Hz, $f_S={fs}\,$Hz)",
       ylabel="Doubled Magnitude")
ax.stem(f, X, basefmt=' ', markerfmt=f'C0.', linefmt=f'C0-')
ax.grid(True)
plt.show()
```

doc/source/tutorial/fft.rst
```diff
@@ -8,187 +8,228 @@ Fourier Transforms (:mod:`scipy.fft`)
...
->>> fftshift(x)
-array([4, 5, 6, 7, 0, 1, 2, 3])
+Since :math:`x` is real and the phase (angle) of :math:`y` is of no interest,
+we can obtain the same result by calculating the inverse DFT of :math:`x`
+using :func:`ifft`, and then scaling its amplitude by :math:`2.`
 
-The example below plots the FFT of two complex exponentials; note the
-asymmetric spectrum.
+>>> yi = ifft(x)
+>>> ya = 2.0 * np.abs(yi[:len(f)])
```
DietBru, line 214: Wouldn't it me more informative to prove the more general relation
 `ifft(x.conj()).conj() == fft(x) / len(x)` as well as 
 `ifft(x.conj(), norm='ortho').conj() == fft(xnorm='ortho')`.

Furthermore, this context gives some insights for what the `norm` keyword is for.

doc/source/tutorial/fft.rst
```diff
@@ -210,24 +252,112 @@ array([ 4.5       +0.j        ,  2.08155948-1.65109876j,
...
+The corresponding amplitudes may be symmetric (by chance or
+by construction), but usually they are not.
+In this case it is common to plot the spectrum of a signal with the
+zero frequency at the center of the horizontal axis, with the
+understanding that the bin frequency is wrapped with period :math:`F_s.`
+For an even :math:`N,` the bin frequency :math:`F_s/2` can be shown
+either as positive or negative (but usually not both).
+This kind of plot is called "two-sided", while the plot of the spectrum
+of a real signal in the range :math:`[0,F_s/2]` is called "one-sided".
+
```
DietBru, line 283: It might be easier to concisely present those ideas with the help of a plot—what about something like this:
image

It shows:
* The discrete Fourier transform of an $n$ sample sequences of even and odd length (blue stems). The gray/white background illustrate the periodicity in $n$. The sampling interval was chosen to be one.
* The `fft()` function returns the DFT values in the range from `[-n//2:(n+1)//2]`. The frequency can be generated with `fftfreq()`, which returns first the positive then the negative frequencies. 
* A caveat for signals with an even number of samples: `fftfreq()` returns the negative (here: $l=-4$) and not the positive Nyquist frequency (here: $l=4$). It does not matter, since those values are always same. This problem does not appear in signals with an odd number of samples, since the Nyquist frequency does not fall onto a bin.
* The `rfft()` / `rfftfreq()` functions are convenience function for `fft()` / `fftfreq()`, which only return the bins of non-negative frequencies (which is sufficient for real-valued inputs). The number of `rfft()` bins are the same in both plots, in spite of having different number of input samples. For that reason the number of output samples should always be specified when using the `irfft()` function.

The code for the plot:
```python
import numpy as np
from scipy.fft import fft, rfft, fftfreq, rfftfreq
from scipy.signal import windows

l0, l1 = -7, 8  # span of frequency bins for Fourier series
lX0 = np.arange(l0, l1)  # frequency bins for X0
fg0, axx = plt.subplots(2, 1, sharex='all', tight_layout=True)
for ax_, n in zip(axx, (7, 6)):
    x = windows.gaussian(n, .8,sym=False)   # input signal
    # Calculate discrete Fourier transform:
    X0 = np.array([sum(x_ * np.exp(2j * np.pi * k_ * l_ / n)
                       for k_, x_ in enumerate(x)) for l_ in lX0])
    X1, lX1 = fft(x), fftfreq(n, 1 / n)
    X2, lX2 = rfft(x), rfftfreq(n, 1 / n)

    ax_.set_title(f"Comparison of fft() / rfft() for ${n}$ sample Signal")
    ax_.set(ylabel="Magnitude $|X[l]|$", facecolor='lightgray')
    ax_.axvspan(-n / 2, n / 2, color='white')  # white background for primary period
    ax_.stem(lX0, abs(X0), basefmt=' ', markerfmt=f'', linefmt=f'C0-', label="DFT")
    ax_.plot(lX1, abs(X1), 'C0x', alpha=0.5, label="fft()")
    ax_.plot(lX2, abs(X2), 'C1^', alpha=0.5, label="rfft()")
    ax_.legend()
    ax_.grid()
axx[-1].set(xlabel="Frequency Bins $l$", xlim=(l0 - .5, l1 - .5), xticks=lX0, )
plt.show()
```

DietBru (changes requested):
Sorry for not finding the time to finish this review sooner. It is great that this section is seeing some love. I am aware that this page evolved over a long time, but I am not sure it is evolving into the best direction: IMHO, the big topics of the discrete Fourier transform (DFT) and spectral analysis are intermingled a little bit too much, making it hard for non-experts to understand what is going on.

E.g.: Spectral analysis assumes there is an underlying continuous-time band-limited signal, which is sampled an at equidistant intervals. The example plot of the letter ω illustrates that for some applications the notion of equidistant sampling is not always insightful. Generally, there a lot of confusing nooks and crannies (in topics like aliasing, resampling, windowing, etc), which can only be picked apart by carefully looking at the underlying assumptions. For this reason, some people also distinguish between the [Discrete Fourier transform] (DFT) and the [Discrete-time Fourier transform] (DTFT).

There's a relatively new [Spectral Analysis Section] section in the signal processing module's user guide (disclaimer: I wrote most of it), which discusses relevant spectral analysis caveats by analyzing a single finite duration sine wave.
I am wondering, if it would make more sense to move the spectral analysis parts, namely the spectral plots, spectral leakage and windowing parts, there as needed (there is quite a bit of overlap). 
The contribution of the PR in this regard would be explaining the difference between the `fft` / `fftfreq` and `rfft` / `rfftfreq` functions (e.g., by a plot like in the comment below).
What do you think about that, @georgeef?

PS: The other parts look good to me.

[Discrete Fourier transform]: https://en.wikipedia.org/wiki/Discrete_Fourier_transform
[Discrete-time Fourier transform]: https://en.wikipedia.org/wiki/Discrete-time_Fourier_transform
[Spectral Analysis Section]: https://scipy.github.io/devdocs/tutorial/signal.html#spectral-analysis

georgeef:
@DietBru 
I need some time to look into the details, but I fully agree that this introduction to FFT functions should better be restricted on how to use the functions and what is their mathematical definition. Signal processing aspects, like their relationship to continuous-time signals, windowing, aliasing, etc., can be better described in the signal processing section.

In my proposal I tried to keep the content as it was, and I focused only on a few clarifications. I think it makes sense to cut some parts and elaborate more on the usage of FFT functions. On the other hand, an introduction to FFT just as a vector transformation, completely isolated from underlying signals, it wouldn't be useful to most users. Maybe we can keep one or two examples showing the spectrum of a sampled signal, but we don't need to go deeper into the details of aliasing and windowing. Also we don't need to explain the relationships between Fourier transforms.

DietBru:
> Maybe we can keep one or two examples showing the spectrum of a sampled signal, but we don't need to go deeper into the details of aliasing and windowing. Also we don't need to explain the relationships between Fourier transforms.

I am fully onboard with this. It makes sense  to show a spectrum, but care needs to be taken when talking about the values between the calculated values (for periodic signals they are zero, whereas for finite duration signals they are the continuous Fourier transform of the respective Fourier series expansion). This [comment] was meant to serve as a starting point for thinking into that direction.


[comment]: https://github.com/scipy/scipy/pull/23182#discussion_r2235496028

DietBru:
> Another issue: `windows.triang(4, sym=False)` gives
> 
> > [1/3, 2/3, 3/3, 2/3]
> 
> The standard definition, at least according to Harris (1978), eq. (23b), gives
> 
> > [0, 1/2, 2/2, 1/2]

You probably wanted a [bartlett] window:
```python
In [1]: from scipy.signal import windows

In [2]: windows.bartlett(4, sym=False)
Out[2]: array([0. , 0.5, 1. , 0.5])
```
I know, the naming is not always helpful. :shrug: 

[bartlett]: https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.windows.bartlett.html

lucascolley:
hey @georgeef, just wondering if you have found some time to take another look at this? No rush :)

georgeef:
@lucascolley 
I have limited time now, but it is in my list.

lucascolley:
just FYI @georgeef we are approaching the deadline for the next release of SciPy, if you wanted to come back to this.

tylerjereddy:
While improving the tutorial documentation is most welcome, this hasn't seen activity in a while and can safely be bumped to the `1.18.0` release schedule, so I'll bump the milestone now.

DietBru:
I took inspiration from this PR and started over in PR #24376. The focus is a bit different—comments are always welcome.
For this reason I am closing this PR—feel free to reopen if you disagree and want to pursue further here.

Issue #23184, opened 2025-06-19
BUG: min Python version enforcement in 1.16.x series? (and main)


tylerjereddy:
The supported Python version range in the `1.16.0` release should be `3.11-3.13` (and eventually `3.14`).

However, I just noticed that Python `3.10.12` is being used for the array API test CI run that was failing for other reasons at https://github.com/scipy/scipy/actions/runs/15748282268/job/44388579936 and https://github.com/scipy/scipy/pull/23164.

That's not great--has our runtime Python version enforcement disappeared from `scipy/__init__.py`? `pyproject.toml` has the lower bound set.

`spin test -j 32` passes on x86_64 Linux in a Python `3.10` venv on the `maintenance/1.16.x` branch.

On latest `main` (4a734661), this is also true. Maybe it is just getting late, but I would have expected a preference for failing early when importing at an unsupported Python version, if only to prevent unwanted tickets in the future.

Of course, the correct error is produced via `python -m pip install -v .` because of the `pyproject.toml` guard.

This gives me some pause though--@rgommers can/should we reinstate the runtime CPython version guard in `scipy/__init__.py` or did I just miss a discussion somewhere?

rgommers:
I have no memory of removing it on purpose. That doesn't mean that it didn't happen on purpose of course. I agree it's nicer to catch this at runtime. I'll do a bit of digging.

rgommers:
@tylerjereddy after some head scratching I concluded that we never had a guard in `scipy/__init__.py`. We did have one in `setup.py`, and we only issue warnings in `__init__.py` for numpy version mismatches:

- https://github.com/scipy/scipy/blob/maintenance/1.11.x/_setup.py#L465-L474
- https://github.com/scipy/scipy/blob/maintenance/1.11.x/scipy/__init__.py#L127-L135

I think that actually makes sense - if you cannot build for py310, then you also don't need a runtime check for it. So the only thing that can go wrong are with `spin` or non-isolated installs that avoid the `pyproject.toml` check. 

What we need to introduce is a check in `meson.build` I'd say, to match what we used to check in `setup.py`. If that makes sense to you, I can open a PR.

tylerjereddy:
> If that makes sense to you, I can open a PR.

That sounds good--failing at build time saves the user/developer even more time too.

You can also link to the relevent `meson` docs if you want me to take a stab at it, but I'm definitely less proficient with `meson`.

PR #23144, opened 2025-06-10
MAINT: array types: array-api-strict got stricter

ev-br:
Small fixes to make `$ spin test -b array_api_strict` pass with the array-api-strict main branch.

PR #23188, opened 2025-06-19
MAINT: a few more backports for 1.16.0

tylerjereddy:
A few more things to tie up before the final `1.16.0` release. Skipping CI for now, until we're ready to go.

Backports included (so far):

1. gh-23178
2. gh-23180
3. gh-23189
4. gh-23144
5. gh-23197
6. gh-23199

TODO:
- [x] handle the Python version check at build time per gh-23184 (Ralf or I will do this in `meson`)
- [x] bump the version of Python used for the array API testing job (should error out with current `3.10` once first point above is handled) on this branch (unless this happens to not be bumped on `main` either, in which case we could backport, but anyway fix that here either way) -- EDIT: need to backport gh-23144
- [x] check the status of CI here again
- [ ] maybe update release notes one more time to get final changes (i.e., scrape in gh-23199); if I do that, probably `[docs only]`

### DIFF
--- a/scipy/signal/tests/test_signaltools.py
+++ b/scipy/signal/tests/test_signaltools.py
@@ -1295,7 +1295,7 @@ class TestMedFilt:
         # us into wrong memory if used (but it does not need to be used)
         dummy = xp.arange(10, dtype=xp.float64)
         a = dummy[5:6]
-        a = xp.lib.stride_tricks.as_strided(a, strides=(16,))
+        a = np.lib.stride_tricks.as_strided(a, strides=(16,))
         xp_assert_close(signal.medfilt(a, 1),  xp.asarray([5.]))
 
     @skip_xp_backends(
--- a/scipy/signal/tests/test_signaltools.py
+++ b/scipy/signal/tests/test_signaltools.py
@@ -1295,7 +1295,7 @@ class TestMedFilt:
         # us into wrong memory if used (but it does not need to be used)
         dummy = xp.arange(10, dtype=xp.float64)
         a = dummy[5:6]
-        a.strides = 16
+        a = xp.lib.stride_tricks.as_strided(a, strides=(16,))
         xp_assert_close(signal.medfilt(a, 1),  xp.asarray([5.]))
 
     @skip_xp_backends(
--- a/.github/workflows/array_api.yml
+++ b/.github/workflows/array_api.yml
@@ -60,7 +60,8 @@ jobs:
 
     - name: Install Python packages
       run: |
-        python -m pip install numpy cython pytest pytest-xdist pytest-timeout pybind11 mpmath gmpy2 pythran ninja meson click pooch hypothesis array-api-strict spin
+        # array-api-strict pinned because gh-23183
+        python -m pip install numpy cython pytest pytest-xdist pytest-timeout pybind11 mpmath gmpy2 pythran ninja meson click pooch hypothesis "array-api-strict<2.4" spin
 
     - name: Install PyTorch CPU
       run: |
--- a/.github/workflows/array_api.yml
+++ b/.github/workflows/array_api.yml
@@ -69,7 +69,8 @@ jobs:
 
     - name: Install JAX
       run: |
-        python -m pip install "jax[cpu]"
+        # Skip 0.6.2: https://github.com/jax-ml/jax/issues/29537
+        python -m pip install "jax[cpu]!=0.6.2"
 
     - name: Install Dask
       run: |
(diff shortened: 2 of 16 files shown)


ev-br:
> bump the version of Python used for the array API testing job (should error out with current 3.10 once first point above is handled) on this branch (unless this happens to not be bumped on main either, in which case we could backport, but anyway fix that here either way)

For this you'd want to backport https://github.com/scipy/scipy/pull/23144 (two commits: one bumps the python version, the other fixes small issues that the last released array-api-strict barfs on). I can send a backport PR to the maintenance branch if that helps.

tylerjereddy:
Thanks Evgeni, I don't think you need to open a PR, I've added it to the TODO list above. I'm a bit tied up with grant reviews today and tomorrow, but sounds like the path forward is there.

tylerjereddy:
Well, I've backported what I thought I needed.

Now the array API job is failing with Python `3.10` usage despite `python-version: ['3.12']`, and I see `Warning: Neither 'python-version' nor 'python-version-file' inputs were supplied. Attempting to find '.python-version' file.` Furthermore, `git diff main -- .github/workflows/array_api.yml` doesn't show anything that seems obviously related to that issue.

tylerjereddy:
Backporting gh-23199 also caused an issue that didn't show up on `main`. I need to move on for now, I'll try to revisit this fresh tomorrow.

lucascolley:
something is wrong with the templating: https://github.com/scipy/scipy/blob/3d56b36b28523c4e27c67596ea94c2ecd67b9d31/.github/workflows/array_api.yml#L96 and https://github.com/scipy/scipy/actions/runs/15788333109/job/44509424319?pr=23188 show that `${{ matrix.python-version }}` is evaluating to an empty string.

tylerjereddy:
Do you see anything in `git diff main -- .github/workflows/array_api.yml` that could explain that?

Coming from another "included" file perhaps? Must be something obvious I'm missing on that one.

lucascolley:
I am completely puzzled after staring at the files for a few minutes.

tylerjereddy:
Ok, maybe a thing for tomorrow morning, thanks for checking.

scipy/signal/tests/test_signaltools.py
```diff
@@ -1295,7 +1295,7 @@ def test_odd_strides(self, xp):
         # us into wrong memory if used (but it does not need to be used)
         dummy = xp.arange(10, dtype=xp.float64)
         a = dummy[5:6]
-        a.strides = 16
+        a = xp.lib.stride_tricks.as_strided(a, strides=(16,))
```
ev-br:
```suggestion
        a = np.lib.stride_tricks.as_strided(a, strides=(16,))
```

because otherwise (https://github.com/scipy/scipy/actions/runs/15788333110/job/44509424300?pr=23188), grumble grumble

```
>       a = xp.lib.stride_tricks.as_strided(a, strides=(16,))
            ^^^^^^
E       AttributeError: module 'scipy._lib.array_api_compat.numpy' has no attribute 'lib'
```

lucascolley: gh-23199 now merged as so

tylerjereddy:
I applied the suggested change from https://github.com/scipy/scipy/pull/23188#discussion_r2159983946, so that the testsuite may pass with older and newer versions of NumPy.

For now though, I've skipped the CI again because the shenanigans from https://github.com/scipy/scipy/pull/23188#issuecomment-2993011648 are reproducing with manual CI job restarts and need a closer look.

tylerjereddy:
oh yuck (see below), that must be the problem... I'll try to patch it out and probably forward port to `main` too if that removal fixes it--I can't think of a reason we'd want this anymore

```yaml
    strategy:
      matrix:
        python-version: ['3.12']
        maintenance-branch:
          - ${{ contains(github.ref, 'maintenance/') || contains(github.base_ref, 'maintenance/') }}
        exclude:
          - maintenance-branch: true
```

tylerjereddy:
The full diff looks exactly as I'd expect it to now. I took a bit of a risk and started a full CI run, including wheels, with the latest fixes.

If that turns out well, the only other thing I might want to do here is update the release notes one more time, although the changes from doing so are almost not worth it (basically just pulling in gh-23199 to the list since it was recently merged after its backport happened here).

tylerjereddy:
I think the single prerelease job failure may be "ok," as long as all the wheel builds (and everything else) are green.

NumPy has been picking up on the merging of bigger changes after their recent release I think, so the new warning could be related to that for example. Handling `main` branch NumPy changes on this release branch is lower prio.

tylerjereddy:
Merging based on analysis above, we need to prioritize getting the release out over minor non-blocking matters at this point I think. The 1 pre-release test failure seems safe to ignore for the moment.

tylerjereddy:
The wheel builds (https://github.com/scipy/scipy/actions/runs/15801847747) all succeeded, so it should be smooth sailing to finish up the release process tomorrow (Sunday, June 22nd).

PR #23189, opened 2025-06-19
BLD: implement build-time version check for minimum Python version

### LINKED ISSUE #23184: BUG: min Python version enforcement in 1.16.x series? (and main)

rgommers:
Closes gh-23184

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -20,6 +20,12 @@ py3 = import('python').find_installation(pure: false)
 py3_dep = py3.dependency()
 
 min_numpy_version = '1.26.4'  # keep in sync with pyproject.toml
+min_python_version = '3.11'   # keep in sync with pyproject.toml
+
+python_version = py3.language_version()
+if python_version.version_compare(f'<@min_python_version@')
+  error(f'Minimum supported Python version is @min_python_version@, found @python_version@')
+endif
 
 # Emit a warning for 32-bit Python installs on Windows; users are getting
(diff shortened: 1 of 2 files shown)


tylerjereddy (approved):
Works as advertised locally in a Python `3.10` venv on x86_64 Linux.

Two CI failures cannot possibly be related, they are specific to a single test each.

The array API CI job is still running, but it has already reached the testing stage, so any failure there would be disconnected from the diff here anyway.

thanks Ralf

PR #23176, opened 2025-06-18
BUG: `spin bench` ignores `--quick` parameter

crusaderky:
Towards #23166.

Fix bug where `spin bench` and `spin bench --quick` would both call `asv --quick`.

Issue #23166, opened 2025-06-16
BENCH/DEV: `spin bench` option `--quick` is useless (always active)


MikhailRyazanov:
### Describe your issue.

`spin bench --help` shows
```
...
Options:
...
  -q, --quick                Run each benchmark only once (timings won't be accurate)
```
but if fact all benchmark are *always* running only once (and thus obtaining accurate timings is impossible) because the `--quick` option is *always unconditionally added* to the `asv` command line here:
https://github.com/scipy/scipy/blob/a465e2ce014c1b20b0e4b949e46361e5c2fb727e/.spin/cmds.py#L787-L791
and here:
https://github.com/scipy/scipy/blob/a465e2ce014c1b20b0e4b949e46361e5c2fb727e/.spin/cmds.py#L803-L806
And `asv` has no options to “undo” it.

I guess, this unconditional addition should be removed, and the declaration of the `--quick` option here:
https://github.com/scipy/scipy/blob/a465e2ce014c1b20b0e4b949e46361e5c2fb727e/.spin/cmds.py#L715-L718
should be modified to make `default=True` (to preserve the current behavior), but also the `--no-quick` variant should be added to enable accurate benchmarks on request.

### Reproducing Code Example

Try any benchmarks without and with the `-q` option.

### Error message

N/A

### SciPy/NumPy/Python version and system information

<details>

```shell
1.17.0.dev0+git20250518.ff473db 2.3.0 sys.version_info(major=3, minor=12, micro=9, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    detection method: pkgconfig
    found: true
    include directory: /usr/include/x86_64-linux-gnu/openblas-openmp/
    lib directory: /usr/lib/x86_64-linux-gnu/openblas-openmp/
    name: openblas
    openblas configuration: USE_64BITINT=0 DYNAMIC_ARCH=1 DYNAMIC_OLDER=1 NO_CBLAS=
      NO_LAPACK= NO_LAPACKE=1 NO_AFFINITY=1 USE_OPENMP=1 GENERIC MAX_THREADS=128
    pc file directory: /usr/lib/x86_64-linux-gnu/pkgconfig
    version: 0.3.28
  lapack:
    detection method: pkgconfig
    found: true
    include directory: /usr/include/x86_64-linux-gnu/openblas-openmp/
    lib directory: /usr/lib/x86_64-linux-gnu/openblas-openmp/
    name: openblas
[...]

PR #23190, opened 2025-06-19
BUG/BENCH: remove unconditional `--quick` from `spin bench --compare`

MikhailRyazanov:
#### Reference issue
Follow-up to gh-23176, see the discussion at gh-23166.

Allows `spin bench` to produce precise timings for comparison benchmarks. The `--quick` option now works as documented (previously it was always active and could not be disabled).

#### Additional information
Should not affect CI benchmark tests, as they call `asv` directly.


rgommers (approved):
Thanks @MikhailRyazanov

PR #23193, opened 2025-06-20
 ENH: spatial: optimize performance of `RigidTransform`

amacati:
This PR improves the performance of `RigidTransform` after the refactor into a pure `Python` class and a `Cython` backend. 

The bulk of performance losses stem from using `MemoryViews` in `Cython.` While the initial transfer into `Cython` is zero-copy and cheap, some `Cython` functions subsequently require these inputs to be `numpy` arrays instead of `MemoryViews,` which creates new `numpy` arrays. There are two solutions to this: 

- We can change the `Cython` types to `numpy` arrays. This is not recommended by `Cython` as the previous `numpy` typing interface is considered outdated.
- We can avoid calling `Cython` functions where possible

This PR opts for the second solution. In addition, it implements some smaller improvements like caching `xp`.

### Motivation
The refactor in #22972 has introduced some overhead that can be further reduced. See the benchmark below as a motivating example.

image

### Reference issue
This is a continuation of #22777 and a direct follow-up of #22972.

### Benchmarks
Below is a full benchmark. It compares the previous pure Cython implementation pre-#22972 ("scipy pure cython") with the performance of #22972 ("XP disabled / enabled") and this PR ("optimized XP enabled / disabled"). If Array API features are disabled, some functions have short paths and thus better performance. Hence, we benchmark #22972 and this PR twice, once with XP enabled and once with XP disabled.

[tf_benchmark.pdf](https://github.com/user-attachments/files/20833580/tf_benchmark.pdf)

### DIFF
--- a/scipy/spatial/transform/_rigid_transform.py
+++ b/scipy/spatial/transform/_rigid_transform.py
@@ -1,7 +1,8 @@
 from __future__ import annotations
 
 from collections.abc import Iterable
-from types import EllipsisType
+from types import EllipsisType, ModuleType
+from collections.abc import Callable
 
 import numpy as np
 
@@ -12,7 +13,6 @@ from scipy._lib._array_api import (
     Array,
 )
 from scipy.spatial.transform import Rotation
-
 import scipy.spatial.transform._rigid_transform_cy as cython_backend
 import scipy._lib.array_api_extra as xpx
 from scipy._lib.array_api_compat import device
@@ -366,6 +366,7 @@ class RigidTransform:
         transform : `RigidTransform` instance
         """
         xp = array_namespace(matrix)
+        self._xp = xp
         matrix = xp_promote(matrix, force_floating=True, xp=xp)
         if matrix.ndim not in (2, 3) or matrix.shape[-2:] != (4, 4):
             raise ValueError(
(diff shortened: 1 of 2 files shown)


lucascolley (approved):
thanks @amacati, this looks great to me! Nice to see the performance improvements over the original implementation already :)

I'll merge soon, but maybe give @scottshambaugh @crusaderky a few days in case they want to check the diff

scottshambaugh:
The asymptotic performance definitely looks a lot better. Any thoughts around if there's a way to keep `apply` faster at low N? I expect that will be one of the most common operations.

amacati:
> The asymptotic performance definitely looks a lot better. Any thoughts around if there's a way to keep `apply` faster at low N? I expect that will be one of the most common operations.

There might be if the `vectors` input can be statically typed without creating new arrays. I will play around some more and check if there is room for further performance improvements.

amacati:
Double-checking `apply` with typed vectors unfortunately did not result in significant speed-ups. Maybe someone with more knowledge in Cython can chime in. This is very likely a matter of making sure we are avoiding copies wherever possible.

scottshambaugh (approved):

scottshambaugh:
Got it, just making sure there wasn't any low hanging fruit. Good to merge from my end!

lucascolley:
thanks both :)

PR #23194, opened 2025-06-20
MAINT: Improve error type for RigidTransform.from_rotation

amacati:
A minor change that has come up during #22972. 

`RigidTransform` raises a `ValueError` when the type passed to `RigidTransform.from_rotation` is anything other than an instance of `Rotation,` which really should be a `TypeError.`

### DIFF
--- a/scipy/spatial/transform/_rigid_transform.py
+++ b/scipy/spatial/transform/_rigid_transform.py
@@ -521,9 +521,8 @@ class RigidTransform:
         >>> len(tf)
         2
         """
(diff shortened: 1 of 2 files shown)


lucascolley (approved):
thanks @amacati, I think this is probably fine as a very small breaking change

scottshambaugh (approved):

rgommers (approved):
I'm fine with treating this as a bug fix as well, seems pretty clear cut.

Issue #18768, opened 2023-06-27
RFC: Ufunc return type for zero dimensional array input.


steppi:
## Introduction

(skip to TLDR if you don't care to read this)

The longstanding behavior for NumPy ufuncs has been to return a scalar in cases where the output would otherwise be expected to be a zero dimensional array. This is accomplished in the NumPy C API with the function [PyArray_Return](https://github.com/numpy/numpy/blob/85385d6eb9da2b45d1e5fa9994e82608e562ef32/numpy/core/src/multiarray/scalarapi.c#L770C1-L779C34).

```c

/* Return Array Scalar if 0-d array object is encountered */


/*NUMPY_API
 *
 * Return either an array or the appropriate Python object if the array
 * is 0d and matches a Python type.
 * steals reference to mp
 */
NPY_NO_EXPORT PyObject *
PyArray_Return(PyArrayObject *mp)
```

ufuncs in `special` generated with the standard infrastructure [_generate_pxd.py](https://github.com/scipy/scipy/blob/main/scipy/special/_generate_pyx.py) will automatically have this behavior.

There are however functions in special which have ufunc like behavior but are not generated with the standard infrastructure. These include the factorial functions [factorial](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.factorial.html#scipy.special.factorial), [factorial2](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.factorial2.html#scipy.special.factorial2), and [factorialk](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.factorialk.html#scipy.special.factorialk). These functions have an `exact=True` option where exact calculations using Python's arbitrary-precision integers can be made for array-like inputs, returning an `object` array of Python `int`s.

Until recently, factorial functions in SciPy were inconsistent with each other in terms of API and behavior. @h-vetinari made a heroic effort to resolve these inconsistencies, discussion took place in issue gh-15600, and the work was accomplished over a number of PRs. I did not participate in or follow the discussions due to dealing with difficult circumstances over that time period. One thing which was decided was that each factorial function should return a 0d array when given 0d input. This decision appears to have been made from first principles, and if I interpret correctly, I think @h-vetinari believes that the existing standard ufunc behavior to return a scalar when the output would otherwise be expected to be a 0d array is incorrect.

[...]

PR #23198, opened 2025-06-20
ENH: spatial.transform.Rotation: refactor into pure Python class and cython backend

amacati:
Refactor `Rotation` into a pure Python class with a Cython backend for `numpy` arrays.

The previous implementation in `_rotation.pyx` has now been split into `_rotation.py` and the functional-style cython backend `_rotation_cy.pyx`. In addition, the functional Cython backend for `RigidTransform` has been adapted to now use the functional `Rotation` backend.

This PR follows exactly the same structure outlined in #22972. `Rotation` now only holds the central data structure (`_quat`). Its methods determine which `xp` to use for the input arrays and then delegate the work to the respective backend. The previous cython function bodies have been converted to pure functions and moved to the backend. Git does not know that the file has been split, so the cython code will show up as new code. Most of it has been directly adopted from the previous class.

## Reference issue
#22500 , #22777

This change is necessary in preparation for the upcoming Array API support. The current implementation only has a single backend. A follow-up PR will implement the generic Array API backend.

## Additional information
The Array API tests have been converted in #22939 . These tests have been left unchanged with two exceptions:
- `Rotation.magnitude` now returns a 0d `ndarray` for single rotations instead of a `float.` While this is more consistent with treating everything as Arrays and makes subsequent PRs for the Array API adoption easier, it is a breaking change that needs to be discussed.
- Previously, `Rotation` had an `_as_euler_from_matrix` and an `_as_euler_from_quat` method, which are both tested for correctness and compared with each other. However, the public interface of `Rotation` does not expose these methods and exclusively uses `_as_euler_from_matrix`. To reduce the amount of code that later has to be ported to the Array API and to clean up the `Rotation` class, I think it would make sense to drop the unused `_as_euler_from_matrix`. For now, tests are only being skipped, so depending on the overall sentiment here we can also easily add it back in.

## Benchmarks
The benchmark compares three cases:
- scipy pure cython: Current implementation as pure Cython class
- XP disabled: SCIPY_ARRAY_API not set to 1. Array API features are disabled, which leads to array_namespace having near-zero overhead
- XP enabled: SCIPY_ARRAY_API set to 1. Array API features are enabled, which incurs additional overhead

[rotation_benchmark.pdf](https://github.com/user-attachments/files/20838168/rotation_benchmark.pdf)


scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -991,6 +991,7 @@ def test_as_euler_degenerate_symmetric_axes(xp, seq_tuple, intrinsic):
 def test_as_euler_degenerate_compare_algorithms(xp, seq_tuple, intrinsic):
     # this test makes sure that both algorithms are doing the same choices
     # in degenerate cases
+    pytest.skip("Skipping this test for now")
```
lucascolley: it would be worth mentioning `_as_euler_from_matrix` just to be clear

lucascolley:
> `Rotation.magnitude` now returns a 0d ndarray for single rotations instead of a float. While this is more consistent with treating everything as Arrays and makes subsequent PRs for the Array API adoption easier, it is a breaking change that needs to be discussed.

Could you expand on how this makes subsequent PRs easier? In other places where we haven't made the call to make a breaking change, we are using https://github.com/scipy/scipy/blob/main/scipy/_lib/_array_api_no_0d.py to keep the tests concise.

scipy/spatial/transform/_rotation.py
```diff
@@ -3112,10 +1963,12 @@ def create_group(cls, group, axis='Z'):
            <https://en.wikipedia.org/wiki/Point_groups_in_three_dimensions>`_
            on Wikipedia.
         """
+        # DECISION: We create groups as numpy arrays only. We have no nice way of specifying the
+        # array framework here. Another option would be to pass in an array_module argument, but
+        # that would change the function signature, and having modules as argument feels off.
         return create_group(cls, group, axis=axis)
```
lucascolley: there is precedent for array module args: https://docs.scipy.org/doc/scipy/reference/generated/scipy.fft.fftfreq.html

scipy/spatial/transform/_rotation.py
```diff
@@ -3179,12 +2032,25 @@ def __getitem__(self, indexer):
...
-        return self.__class__(np.asarray(self._quat)[indexer], normalize=False)
-
-    def __setitem__(self, indexer, value):
+        is_array = isinstance(indexer, type(self._quat))
+        # Masking is only specified in the Array API when the array is the sole index
+        # TODO: Getting xp on every call may be expensive. Check if we can make access to xp more
+        # efficient. Should we store a self._xp attribute?
+        # TODO: This special case handling is mainly a result of Array API limitations. Ideally we
+        # would get rid of them altogether and converge to [indexer, ...] indexing.
+        xp = array_namespace(self._quat)
```
lucascolley: this can be updated after gh-23193 is merged

scipy/spatial/transform/_rotation.py
```diff
@@ -3291,19 +2150,22 @@ def random(cls, num=None, rng=None):
...
-            sample = rng.normal(size=(num, 4))
-
-        return cls(sample)
+        """
+        # DECISION: How do we handle random numbers in other frameworks?
+        # TODO: The array API does not have a unified random interface. This method only creates
+        # numpy arrays. If we do want to support other frameworks, we need a way to handle other rng
+        # implementations.
+        sample = cython_backend.random(num, rng)
+        return cls(sample, normalize=True, copy=False)
```
lucascolley: x-ref https://github.com/data-apis/array-api-extra/issues/308. For SciPy's purposes, wrapping NumPy's rng is fine for now. See above for potential array module parameter.

lucascolley (commented):
cursory review

amacati:
> > `Rotation.magnitude` now returns a 0d ndarray for single rotations instead of a float. While this is more consistent with treating everything as Arrays and makes subsequent PRs for the Array API adoption easier, it is a breaking change that needs to be discussed.
> 
> Could you expand on how this makes subsequent PRs easier? In other places where we haven't made the call to make a breaking change, we are using https://github.com/scipy/scipy/blob/main/scipy/_lib/_array_api_no_0d.py to keep the tests concise.

Using a real Python float would break being able to jit compile things for jax or doing stuff with autograd as far as I know. It's not that we guarantee users can do these things, but I tried to keep them in mind whereever possible.

lucascolley:
> Using a real Python float would break being able to jit compile things for jax or doing stuff with autograd as far as I know. It's not that we guarantee users can do these things, but I tried to keep them in mind whereever possible.

Ah, does main currently return a NumPy scalar, or a Python float?

If the former, it would be acceptable to keep returning a NumPy scalar for NumPy, and return an array for other backends, with wishful thinking in the direction of https://github.com/scipy/scipy/issues/18768. I can't remember if we've hit cases of the latter yet for code that has been converted to have JAX jit support

scipy/spatial/transform/_rotation_cy.pyx
```diff
@@ -0,0 +1,1425 @@
...
+
+    seq = seq.lower()
+    if not extrinsic:
+        seq = seq[::-1]
+
+    # The algorithm assumes extrinsic frame transformations. The algorithm
+    # in the paper is formulated for rotation quaternions, which are stored
+    # directly by Rotation.
+    # Adapt the algorithm for our case by reversing both axis sequence and
+    # angles for intrinsic rotations when needed
```
scottshambaugh: Marking to myself that this algorithm has changed, need to come back and review more thoroughly.

scipy/spatial/transform/_rotation_cy.pyx
```diff
@@ -0,0 +1,1425 @@
...
+
+@cython.embedsignature(True)
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def magnitude(double[:, :] quat) -> double[:, :]:
+    cdef Py_ssize_t num_rotations = quat.shape[0]
+    cdef double[:] angles = _empty1(num_rotations)
+
+    for ind in range(num_rotations):
+        angles[ind] = 2 * atan2(_norm3(quat[ind, :3]), abs(quat[ind, 3]))
```
scottshambaugh, line 906: This might be vectorizable now

scipy/spatial/transform/_rotation.py
```diff
@@ -2575,43 +1559,14 @@ def apply(self, vectors, inverse=False):
...
-            return np.matmul(matrix, vectors)
-
-        if self.single:
-            matrix = matrix[None, :, :]
-
-        if n_rotations == 1:
-            # Single rotation/many vectors, use matmul for speed: The axes argument
-            # is such that the input arguments don't need to be transposed and the
-            # output argument is contineous in memory.
-            return np.matmul(matrix, vectors, axes=[(-2, -1), (-1, -2), (-1, -2)])[0]
```
scottshambaugh, line 2608: This fast path was removed, does einsum show equal in speed now?

scottshambaugh, line 2608: This fast path was removed, does einsum show equal in speed now?

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -884,11 +884,11 @@ def test_stats(error, mean_max, rms_max):
     angles_quat = rotation.as_euler(seq)
     # TODO: Why are we using _as_euler_from_matrix here? As a sanity check? It is not
     # part of the public API and should not be used anywhere else
```
scottshambaugh: Should remove the TODO and lines here and below rather than commenting out, if we decide to remove them.

amacati: Agreed. This was meant as a reminder to ask the reviewers about the original reasoning behind having two implementations and if removing one is fine or not.

amacati:
> > Using a real Python float would break being able to jit compile things for jax or doing stuff with autograd as far as I know. It's not that we guarantee users can do these things, but I tried to keep them in mind whereever possible.
> 
> Ah, does main currently return a NumPy scalar, or a Python float?
> 
> If the former, it would be acceptable to keep returning a NumPy scalar for NumPy, and return an array for other backends, with wishful thinking in the direction of #18768. I can't remember if we've hit cases of the latter yet for code that has been converted to have JAX jit support

The latter, main currently returns Python floats.

scipy/spatial/transform/_rotation.py
amacati: Not to my knowledge, this might have been by accident. Will double-check

scipy/spatial/transform/_rotation_cy.pyx
scottshambaugh: Ok, I see this is just the switch to _compute_euler_from_quat. Note that "the paper" is no longer in any comments, so should remove that bit of this comment.

scottshambaugh (commented):
Not an easy diff to review through the github UI, so I copied each function into an external diff tool to review it. Looks good to me, a few comments but nothing blocking.

The biggest change is that the `_compute_euler_from_quat` algorithm has been inlined into `as_euler`, which means that `_compute_euler` has been removed and neither `_as_euler_from_matrix` or `_compute_euler_from_matrix` are being called anywhere any more. 

I dug up the original implementation of the `_compute_euler_from_quat` algorithm, which was merged in this PR: https://github.com/scipy/scipy/pull/17392

That PR made it such that `_as_euler_from_matrix` was no longer able to be called through any path in the public API, so I'm not sure why it was kept in the codebase. The quaternion function has been working in the released library for ~2.5 years, and these are all private functions, so I would support removing both `_as_euler_from_matrix` and `_compute_euler_from_matrix` entirely (keeping any testing that they alone were implicitly covering).

scipy/spatial/transform/_rotation_cy.pyx
scottshambaugh: Or even better, update to cite the paper: https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0276302

scottshambaugh:
Ok, it looks like the `*_euler_from_matrix` functions were left in a bit arbitrarily, to help give confidence in the new method through those comparison tests even if they were never publicly accessed. See this comment thread: https://github.com/scipy/scipy/pull/17392#discussion_r1024259309. I would vote for removing them now, @evbernardes or @nmayorov do you have thoughts on that?

evbernardes:
> Ok, it looks like the `*_euler_from_matrix` functions were left in a bit arbitrarily, to help give confidence in the new method through those comparison tests even if they were never publicly accessed. See this comment thread: https://github.com/scipy/scipy/pull/17392#discussion_r1024259309. I would vote for removing them now, @evbernardes or @nmayorov do you have thoughts on that?

@scottshambaugh indeed, it was left there only to be sure that my new method was returning the same set of Euler angles as the old matrix method (as there are always, even in non degenerate cases, two alternative sets of Euler angles that produce the same final rotation).

My reasoning is that even if the final rotations would be equivalent, the angles should also be the same as not to break any code that uses the angle values directly.

If it's not that important anymore to write this explicitly on the code/testing, then indeed it's not that important to leave this part there.

scipy/spatial/transform/_rotation.py
amacati: This would work. Should we introduce that in this PR already? Might make sense to leave this as it is for now and add that with the Array API implementation

lucascolley: Yes, I would suggest leaving that for a follow-up 👍 please just adjust the comments in this PR

scipy/spatial/transform/_rotation.py
amacati: At the moment the optimizations in https://github.com/scipy/scipy/pull/23193 are focused on `RigidTransform`. The plan was to maybe follow up this PR with a similar performance optimization PR for `Rotation`, which then could get rid of this TODO.

scipy/spatial/transform/_rotation_cy.pyx
amacati: Will add the reference in the next commit.

scipy/spatial/transform/_rotation_cy.pyx
amacati: I was planning on doing another optimization PR similar to the one for `RigidTransform`. This might be a good fit for it. Some Cython loops are faster than using the `numpy` broadcasting though, so this will have to get benchmarked.

amacati:
> My reasoning is that even if the final rotations would be equivalent, the angles should also be the same as not to break any code that uses the angle values specifically.
> If it's not that important anymore to write this explicitly on the code/testing, then indeed it's not that important to leave this part there.

Given that it has been used for several years now apparently and has been tested to produce the same angles I think it would be fine to get rid of the `*_euler_from_matrix` versions and streamline the code. What do you think @scottshambaugh @lucascolley.

lucascolley:
+1

scipy/spatial/transform/_rotation.py
amacati: After checking this function again it seems like a lot has changed between the old implementation and the one that was added here, not sure why. The last commit has revised the function, so it would be great if you could have another look at it. This one should have the same optimizations.

amacati:
Regarding the performance of `Rotation.apply`: I have added some optimizations specifically for the `numpy` / Cython code path. One source of latency at the moment is determining the `device` which we could cache on initialization. I would leave this for the performance optimization follow-up, and it's not certain that `apply` reaches the same speeds as before, but it would certainly improve performance.

scipy/spatial/transform/_rotation_cy.pyx
```diff
@@ -1155,32 +1156,32 @@ def reduce(double[:, :] quat, left=None, right=None):
...
+        matrix = np.swapaxes(matrix, -1, -2)
 
-    return result
+    if n_vectors == 1 and n_rotations == 1:
+        return np.matmul(matrix, vectors[0])
+
+    if n_rotations == 1:
+        # Single rotation/many vectors, use matmul for speed: The axes argument
+        # is such that the input arguments don't need to be transposed and the
+        # output argument is contineous in memory.
```
scottshambaugh:
```suggestion
        # output argument is continuous in memory.
```

scipy/spatial/transform/_rotation_cy.pyx
```diff
@@ -916,6 +916,7 @@ def as_euler(double[:, :] quat, seq, bint degrees=False) -> double[:, :]:
     # The algorithm assumes extrinsic frame transformations. The algorithm
     # in the paper is formulated for rotation quaternions, which are stored
     # directly by Rotation.
+    # See https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0276302
```
scottshambaugh:
```suggestion
    # directly by Rotation. See:
    # Bernardes E, Viollet S (2022) Quaternion to Euler angles conversion: A
    # direct, general and computationally efficient method.
    # PLoS ONE 17(11): e0276302. https://doi.org/10.1371/journal.pone.0276302
```

scottshambaugh (commented):
Yeah I agree we can remove the `*_euler_from_matrix` functions, best to take out any unused code to make things cleaner.

scipy/spatial/transform/_rotation.py
scottshambaugh: Looks good to me!

amacati:
The last point that requires attention seems to be the type of `magnitude` for the 0d case. The current implementation returns plain Python floats, which makes typing more difficult, forces us to synchronize when using GPU-backed arrays, prevents jit compilation with `jax` in subsequent PRs etc.

Given these drawbacks and reading the discussion in #18768 (in particular @rgommers's comment https://github.com/scipy/scipy/issues/18768#issuecomment-1612881324), I would strongly favor returning 0d arrays here. We could avoid this case by introducing special handling for `numpy,` but I wonder if it's better to make a small breaking change here instead of introducing inconsistencies. Not sure what's your stance on this @lucascolley

lucascolley:
@crusaderky have we encountered the case of functions returning python scalars before? My memory fails me.

@mdhaber I think we had something like this come up in `stats`?

crusaderky:
> @crusaderky have we encountered the case of functions returning python scalars before? My memory fails me.

Yes.
- `cluster.is_valid_linkage` and `cluster.is_valid_im` return a python bool for all eager backends and a 0d array for lazy ones. Not terribly happy with the design, in hindsight.
- `cluster.is_monotonic` returns a `np.bool_` on numpy and a 0d array on other backends. This IMHO is the correct approach.
- `cluster.is_isomorphic` (until now numpy-only) returns a Python bool. https://github.com/scipy/scipy/pull/22989 proposes to change it to match `is_monotonic` when SCIPY_ARRAY_API=1.

lucascolley:
thanks! In that case, I would suggest including special handling for NumPy (if needed) to return a NumPy scalar instead of a Python scalar, which IIUC shouldn't break anything @amacati

amacati:
> thanks! In that case, I would suggest including special handling for NumPy (if needed) to return a NumPy scalar instead of a Python scalar, which IIUC shouldn't break anything @amacati

Done with the latest commit. The section is also commented with a link to our discussion. Any changes to the tests regarding `magnitude` have been reverted. Should be ready for reviews.

amacati:
Added `NotImplemented` for `__mul__` with tests.

scipy/spatial/transform/_rotation_cy.pyx
```diff
@@ -0,0 +1,1425 @@
...
+            angles[idx] -= 2 * pi
+
+    if case != 0:
+        warnings.warn("Gimbal lock detected. Setting third angle to zero "
+                      "since it is not possible to uniquely determine "
+                      "all angles.", stacklevel=3)
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+cdef double[:, :] _compute_euler_from_matrix(
```
scottshambaugh: This can get removed in this PR or a follow-on, your call.

scottshambaugh (approved):
This looks ready to ship to me, thank you for the tremendous effort that went into these cython backends!

scipy/spatial/transform/_rotation_cy.pyx
amacati: Good catch, removed it.

scipy/spatial/transform/_rotation.py
```diff
@@ -2844,23 +1799,20 @@ def magnitude(self):
...
-    @cython.boundscheck(False)
-    @cython.wraparound(False)
-    def approx_equal(Rotation self, Rotation other, atol=None, degrees=False):
+            # Special handling for numpy and single rotations. self._single is only set
+            # if xp is numpy. We therefore know that magnitude is a numpy array and
+            # that it contains a single element. Previously this code returned a Python
+            # float in that case. Here we return a numpy float64 scalar. All other
+            # Array API libraries return 0d arrays instead.
+            # See https://github.com/scipy/scipy/pull/23198#issuecomment-3003757848
+            return magnitude[0].item()
```
lucascolley:
```suggestion
            return magnitude[0]
```

I think we just want this?

```console
(Pdb) magnitude
array([0.])
(Pdb) type(magnitude[0])
<class 'numpy.float64'>
(Pdb) type(magnitude[0].item())
<class 'float'>
```

scipy/spatial/transform/_rotation_cy.pyx
```diff
@@ -0,0 +1,1283 @@
...
+
+    q = identity(len(angles))
+    for i in range(num_axes):
+        qi = from_rotvec(angles[:, i, np.newaxis] * axes[i])
+        if extrinsic:
+            q = compose_quat(qi, q)
+        else:
+            q = compose_quat(q, qi)
+
+    return q[0] if is_single else q
```
lucascolley:
```suggestion
    return q[0] if is_single else q

```

lucascolley (approved):
thanks @amacati, great work! Just one more comment about `magnitude`.

scipy/spatial/transform/_rotation.py
amacati: True. I thought `item()` would also yield a `np.float64`. Applied both your suggestions.

lucascolley:
thanks all :)

PR #23199, opened 2025-06-20
MAINT, TST: NumPy stride deprecation shim

tylerjereddy:
* This patch allows the SciPy testsuite to pass against NumPy `main` branch, after https://github.com/numpy/numpy/pull/28925 was merged in the last day or so.

We shouldn't really have to backport this, since it is slated for NumPy `2.4.x` series, but if it becomes annoying on the maintenance branch we could backport it since it is a test-only shim (though not running dev version tests on maintenance branches is also something we discussed in the past).

### DIFF
--- a/scipy/signal/tests/test_signaltools.py
+++ b/scipy/signal/tests/test_signaltools.py
@@ -1295,7 +1295,7 @@ class TestMedFilt:
         # us into wrong memory if used (but it does not need to be used)
         dummy = xp.arange(10, dtype=xp.float64)
         a = dummy[5:6]
-        a = xp.lib.stride_tricks.as_strided(a, strides=(16,))
+        a = np.lib.stride_tricks.as_strided(a, strides=(16,))
         xp_assert_close(signal.medfilt(a, 1),  xp.asarray([5.]))
 
     @skip_xp_backends(
--- a/scipy/signal/tests/test_signaltools.py
+++ b/scipy/signal/tests/test_signaltools.py
@@ -1295,7 +1295,7 @@ class TestMedFilt:
         # us into wrong memory if used (but it does not need to be used)
         dummy = xp.arange(10, dtype=xp.float64)
         a = dummy[5:6]
-        a.strides = 16
+        a = xp.lib.stride_tricks.as_strided(a, strides=(16,))
         xp_assert_close(signal.medfilt(a, 1),  xp.asarray([5.]))
 
     @skip_xp_backends(


scipy/signal/tests/test_signaltools.py
```diff
@@ -1295,7 +1295,7 @@ def test_odd_strides(self, xp):
         # us into wrong memory if used (but it does not need to be used)
         dummy = xp.arange(10, dtype=xp.float64)
         a = dummy[5:6]
-        a.strides = 16
+        a = xp.lib.stride_tricks.as_strided(a)
```
tylerjereddy: I don't know if using `xp` here contributes that much--pretty sure the stride trick stuff can't be part of the array API standard anyway.

tylerjereddy: I should probably set the value to `16` though...

ev-br: `xp.asarray(np.stride_tricks.as_strided(np.asarray(a),...))) `

lucascolley (approved):
thanks Tyler, LGTM now that `strides=(16,)` is here!

scipy/signal/tests/test_signaltools.py
ev-br: Or make the test numpy-only via `@skip_xp_baxkends(np_only=True) `

tylerjereddy: it is also NumPy-only in the decorator I think, unless you're thinking of farther into the future (although some array API backends still don't propagate the strides do they?)

tylerjereddy:
note that this patch causes a failure on the maintenance branch when backported: https://github.com/scipy/scipy/actions/runs/15788333092/job/44509424585?pr=23188

The array API stuff and its deps are pretty tricky to handle portably on both branches..

scipy/signal/tests/test_signaltools.py
```diff
@@ -1295,7 +1295,7 @@ def test_odd_strides(self, xp):
         # us into wrong memory if used (but it does not need to be used)
         dummy = xp.arange(10, dtype=xp.float64)
         a = dummy[5:6]
-        a.strides = 16
+        a = xp.lib.stride_tricks.as_strided(a, strides=(16,))
```
lucascolley:
```suggestion
        a = np.lib.stride_tricks.as_strided(a, strides=(16,))
```

lucascolley:
thanks Tyler

PR #23200, opened 2025-06-20
Fix miscalculation of norminvgauss CDF when x is very large

### LINKED ISSUE #23196: BUG: norminvgauss cdf becomes inaccurate in far right tail of the distribution
### Describe your issue.

`norminvgauss(...).cdf(x)` returns wildly inaccurate values when it is evaluated for values of `x` in the far right tail of the distribution, i.e. where the true CDF is near 1. 

Perhaps related [this issue](https://github.com/scipy/scipy/issues/18117)?

### Reproducing Code Example

```python
from scipy.stats import norminvgauss
nig = norminvgauss(b=0, loc=0, a=1, scale=1)
print(nig.cdf(10)) #Correctly returns a value near 1.0
print(nig.cdf(100)) #Should return a value near 1.0, but returns a value near 0.0
```

### Error message

```shell
None
```

```shell
{
  "Compilers": {
    "c": {
      "name": "clang-cl",
      "linker": "lld-link",
      "version": "19.1.7",
      "commands": "clang-cl",
      "linker args": "-Wl,-defaultlib:C:/bld/scipy-split_1739790745069/_build_env/lib/clang/19/lib/windows/clang_rt.builtins-x86_64.lib"
    },
    "cython": {
      "name": "cython",
      "linker": "cython",
      "version": "3.0.12",
      "commands": "cython"
    },
    "c++": {
      "name": "clang-cl",
      "linker": "lld-link",
      "version": "19.1.7",
      "commands": "clang-cl",
      "linker args": "-Wl,-defaultlib:C:/bld/scipy-split_1739790745069/_build_env/lib/clang/19/lib/windows/clang_rt.builtins-x86_64.lib"
    },
    "fortran": {
      "name": "llvm-flang",
      "linker": "lld-link",
      "version": "19.1.7",
      "commands": "flang-new",
      "args": "-D_CRT_SECURE_NO_WARNINGS, -fms-runtime-lib=dll, -fuse-ld=lld, -IC:\\bld\\scipy-split_1739790745069\\_h_env\\Library\\include",
      "linker args": "-Wl,-defaultlib:C:/bld/scipy-split_1739790745069/_build_env/lib/clang/19/lib/windows/clang_rt.builtins-x86_64.lib, -D_CRT_SECURE_NO_WARNINGS, -fms-runtime-lib=dll, -fuse-ld=lld, -IC:\\bld\\scipy-split_1739790745069\\_h_env\\Library\\include"
    },
    "pythran": {
      "version": "0.17.0",
      "include directory": "..\\..\\_h_env\\Lib\\site-packages\\pythran"
    }
  },
  "Machine Information": {
    "host": {
      "cpu": "x86_64",
      "family": "x86_64",
      "endian": "little",
      "system": "windows"
    },
    "build": {
      "cpu": "x86_64",
      "family": "x86_64",
      "endian": "little",
      "system": "windows"
    },
    "cross-compiled": false
  },
  "Build Dependencies": {
    "blas": {
      "name": "blas",
      "found": true,
      "version": "3.9.0",
      "include directory": "C:/Users/vberge01/.conda/envs/pymc/Library/include",
      "lib directory": "C:/Users/vberge01/.conda/envs/pymc/Library/lib",
      "pc file directory": "C:\\bld\\scipy-split_1739790745069\\_h_env\\Library\\lib\\pkgconfig"
    },
    "lapack": {
      "name": "lapack",
      "found": true,
      "version": "3.9.0",
      "include directory": "C:/Users/vberge01/.conda/envs/pymc/Library/include",
      "lib directory": "C:/Users/vberge01/.conda/envs/pymc/Library/lib",
      "pc file directory": "C:\\bld\\scipy-split_1739790745069\\_h_env\\Library\\lib\\pkgconfig"
    },
    "pybind11": {
      "name": "pybind11",
      "version": "2.13.6",
      "include directory": "C:/Users/vberge01/.conda/envs/pymc/Library/include"
    }
  },
  "Python Information": {
    "path": "C:\\bld\\scipy-split_1739790745069\\_h_env\\python.exe",
    "version": "3.12"
  }
}
```

rishi-kulkarni:
#### Reference issue
Closes #23196 

`norminvgauss` inappropriately used its inherited `_cdf_single` method to calculate the CDF, which worked fine for small values but not for large values. On the other hand, doing `1 - self_sf` in this case should always work. 

#### Additional information
I added a test that fails on the current implementation, but passes on this one.

### DIFF
--- a/scipy/stats/_continuous_distns.py
+++ b/scipy/stats/_continuous_distns.py
@@ -5526,6 +5526,10 @@ class norminvgauss_gen(rv_continuous):
         sq = np.hypot(1, x)  # reduce overflows
         return fac1 * sc.k1e(a * sq) * np.exp(b*x - a*sq + gamma) / sq
 
+    def _cdf(self, x, a, b):
+        # Use 1 - sf(x) for numerical stability instead of integrating from -inf to x
+        return 1.0 - self._sf(x, a, b)
+
     def _sf(self, x, a, b):
         if np.isscalar(x):
             # If x is a scalar, then so are a and b.
--- a/scipy/stats/tests/test_distributions.py
+++ b/scipy/stats/tests/test_distributions.py
@@ -1258,6 +1258,37 @@ class TestNormInvGauss:
         vals = stats.norminvgauss.ppf(x_test, a, b)
         assert_allclose(x_test, stats.norminvgauss.cdf(vals, a, b))
 
+    def test_cdf_precision_large_x(self):
+        """Test for norminvgauss CDF precision bug with large x values.
+        
+        Regression test for gh-XXXXX: norminvgauss.cdf(x) returns
+        incorrect tiny values for large x instead of values close to 1.0.
+        
+        The bug occurs because the default _cdf implementation integrates
+        from -∞ to x, which fails for large x due to numerical precision
+        issues, while the custom _sf method works correctly.
+        """
+        # Test symmetric case (b=0) where mathematical properties are clearest
+        a, b = 1.0, 0.0
+        dist = stats.norminvgauss(a, b)
+        
+        # Test points where the bug is most severe
+        x_values = [20, 30, 50]
+        
+        for x in x_values:
+            # The fundamental relationship cdf(x) + sf(x) = 1.0 must hold
+            cdf_x = dist.cdf(x)
+            sf_x = dist.sf(x)
+            sum_val = cdf_x + sf_x
+            
+            assert_allclose(sum_val, 1.0, rtol=1e-10, atol=1e-15,
+                           err_msg=f"cdf({x}) + sf({x}) = {sum_val}, expected 1.0")
+            
+            # For large positive x, cdf(x) should be very close to 1.0
+            assert_allclose(cdf_x, 1.0, rtol=1e-10, atol=1e-14,
+                           err_msg=f"cdf({x}) = {cdf_x}, expected ≈ 1.0")
+        
+
 
 class TestGeom:
     def setup_method(self):


mdhaber:
I don't think this really fixes the problem since `sf` fails for large negative input:
```python3
from scipy.stats import norminvgauss
nig = norminvgauss(a=1, b=0)
nig.sf(-100)
# np.float64(1.8452928491904642e-10)
```
Also, the subtraction will limit the precision of the CDF for values in the left tail, so we would only want to use that implementation when taking the integral from the left is going to fail.

Please see https://github.com/scipy/scipy/issues/23196#issuecomment-2992978287.

rishi-kulkarni:
Ah, yes, good point. As you note, the precise value of X this starts happening at depends on the values of a and b. At the expense of performance, a simple test may be checking that sf + cdf = 1, and if not, trusting the larger one. What do you think?

mdhaber:
> if not, trusting the larger one

Well, you can see that wouldn't work as described.

```python3
from scipy.stats import norminvgauss
X = norminvgauss(a=1, b=0)
X.sf(100)  # np.float64(3.969811595787509e-47)
# X.cdf(100)  # np.float64(1.8452928491904642e-10)
```
Here, the SF should be trusted, but it is the smaller of the two.

If by "larger" you meant "smaller", that might work for this example, but I'm not certain it would *always* work.

rishi-kulkarni:
Yes, sorry, the smaller of the two. As you noted in the issue, it is indeed difficult to fix this in general, but I think having a heuristic that's correct more often than the current implementation would still be an improvement. Given that there's no closed form for the median, that's probably as good as it gets.

mdhaber:
I'd be a bit uncomfortable doubling the computation time in all calls to the function to deal with this.
I think it would be better for me to sit this one out, so this will need another reviewer. I'll put the time into https://github.com/statistical-python/scikit-stats instead so people who use distributions can do what they want with them.

mdhaber:
@steppi and I discussed closing this, but let's continue to discuss a more robust fix in gh-23196.

PR #21085, opened 2024-06-30
DEP: spatial: deprecate complex input to `cosine` and `correlation`

fancidev:
#### Reference issue
Closes gh-20994.

#### What does this implement/fix?
If `u` or `v` supplied to `cosine()` or `correlation()` is complex, issue a `DeprecationWarning`.

#### Additional information
Before this PR, these functions do not reject or warn about complex inputs explicitly. However, `ValueError`, `ComplexWarning`, or `TypeError` may occur down the code path depending on the value of the inputs. And even if no error occurs, the returned answer is usually not meaningful. In addition, the `cdist`/`pdist` counterparts of these distance metrics only support real input.

To ensure consistency and well-defined behavior, this PR emits a `DeprecationWarning` when either `u` or `v` is complex. (`w` is not checked because a `TypeError` is always raised if `w` is complex.) The plan is to reject complex input in a future release.

See the referenced issue as well as gh-21001 for more details.

PR #23202, opened 2025-06-21
DEP: spatial: raise error for complex input to cosine and correlation

j-bowhay:
follow up to #21085

Complex inputs are deprecated and scheduled to be removed in this release.


rgommers (commented):
Thanks @j-bowhay. Did you consider raising a `TypeError`? That seems a little more appropriate for unsupported dtypes, given that numpy arrays can be parametrized by dtype in type annotations. I'm not sure we do this fully consistently in other places, but I think `TypeError` is expected.

rgommers (approved):
LGTM now, in it goes.

PR #23207, opened 2025-06-21
CI: always set Python in array API testing

tylerjereddy:
* Forward port for the fix for: https://github.com/scipy/scipy/pull/23188#issuecomment-2993011648

* In short, I can't think of a good reason to disregard the setting of the matrix Python version in the array API CI job on maintenance branches at this point. This behavior also caused nontrivial loss of developer time for Lucas and I last night, since the `main` branch deviated from the maintenance branch behavior and I'd rather not have this repeated in the next release cycle since it would be easy to forget about and cause more head scratching.

[skip circle]

### DIFF
--- a/.github/workflows/array_api.yml
+++ b/.github/workflows/array_api.yml
@@ -37,10 +37,6 @@ jobs:
     strategy:
       matrix:
         python-version: ['3.12']
-        maintenance-branch:
-          - ${{ contains(github.ref, 'maintenance/') || contains(github.base_ref, 'maintenance/') }}
-        exclude:
-          - maintenance-branch: true
 
     steps:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2


ev-br (approved):
Wow, great catch. No reason to continue doing it indeed. (no idea how I missed it when bumping the version just one line above.)

rgommers:
I think the intent here was to avoid running this CI job on maintenance branches at all, to avoid failures with changes in dependencies. The `exclude` entry was missing an indent to achieve this it looks like.

I'm fine with this solution as well.

lucascolley:
> I think the intent here was to avoid running this CI job on maintenance branches at all, to avoid failures with changes in dependencies. The exclude entry was missing an indent to achieve this it looks like.

Yes, this did cross my mind when taking a look at it the other day. I figured I must have been missing something, as I thought we would have observed this on the 1.15 maintenance branch — maybe something with a setup-python release turning a warning into an error?

PR #23214, opened 2025-06-22
REL, MAINT: prep for 1.16.1

tylerjereddy:
* Prepare for SciPy `1.16.1`.

We know we'll need at least a `1.16.1` for Python `3.14` support (if not sooner than that).


rgommers (approved):
Thanks Tyler

PR #23218, opened 2025-06-23
ENH: sparse: Add COO indexing for 1d, 2d, and nD

dschult:
Add indexing (`__getitem__` and `__setitem__`) for COO format including for nD sparse.

This includes `Ellipsis`, `newaxis`/`None`, int/slice/array, basic slicing and fancy indexing.

It is likely that optimizations will improve the performance here and some code could be rewritten in C++ as part of sparsetools. The goal of this PR is to get a working numpy-assisted version that provides features and can be used to identify areas for further development including performance enhancements.

Traits:
- The approach here in `__setitem__` is to remove any entries that match the index and then construct and add coords and data for the assigned values, concatenating the new coords with old coords that didn't get assigned to.
- new coords created during `__setitem__` are not actively sorted, so `has_canonical_format` is set to False.
- focused solely on COO format. I haven't tried to pull apart code and/or features that
would be common with e.g. nD compressed or DOK formats.
- When broadcasting the incoming data (i.e. `x` in `A[idx] = x`), this differs from NumPy's indexing in that NumPy raises an IndexError for `x` shaped as (2,1) assigning into a (2,) slot (while pre-pended 1s are removed, post-pended 1s raise). The sparse arrays allow assigning `x` with shape (2, 1) into an index that indicates shape (2,). Probably the difference comes from np.matrix api encouraging 2d column vectors.  Continuing this current support of 2d vector input seems harmless -- and avoids people having to rewrite spmatrix code for sparray. This PR extends the 2D support to any number of
post-pended 1s. Thus x can have shape (2, 1, 1, 1, 1) and be assigned to a slot with shape (2,), or (2, 1), etc. I believe this provides a superset of NumPy functionality in a case where the intent is clear. But I may be missing corner cases.
- I believe I've got the correct handling of `newaxis/None` for `__setitem__`. It works as
tested and seems to match NumPy behavior. But it is possible I am missing some corner
cases. As far as I can tell, using `None` in an index for assignment does not affect
the part of the array being assigned to. It only affects the allowed shape of the incoming data.
- It's not clear to me that NumPy specifies what should be done in some cases of fancy
array indexing with multiple incoming data getting mapped to the same location. For example, `A[[1,1,1] = [2, 4, 5]`. Should the result have `A[1]` be the 2, 4, or 5? I guess numpy assigns the incoming values to the same location in an implementation dependent way. The docs (when [explaining in-place operation surprises](https://numpy.org/devdocs/user/basics.indexing.html#assigning-values-to-indexed-arrays)) talk about creating a temporary matrix to build the values that then get moved to the array. The order of operations controls which of the three assigned values actualy appear in the array after assignment. This PRs sparse implementation builds coords and data for each of the 3 values. It then uses `np.unique` to identify/remove duplicate coords in this case when multiple inputs get mapped to the same location. We could instead neglect the removal of duplicates and support creating duplicate entries in this way as a feature. But that wouldn't match the NumPy behavior, so this PR removes the duplicates leaving the first entry found by `np.unique`.

Possible improvements:
- add benchmarks specifically for nD
- remove duplicates from idx arrays before constructing new_coords might avoid some work/time/space.
- The `__getitem__` and `__setitem`` code is in `_coo.py` instead of `_index.py` to avoid circular import dependencies involving `_ravel_coords`. We could instead refactor `_ravel_coords` to `_index.py` or refactor `__setitem__` to use a private helper method living in `coo_array`. Or we could wait until another format supports nD indexing and refactor at that time.
- new coords created during `__setitem__` are not actively sorted, so has_canonical_format is set False.

Testing:
I added a number of nD tests in `test_coo.py` and "turned on" the extensive 1d and 2d
tests in `test_base.py` for COO format. Of course there could be corner cases this is missing.
For example, boolean array indexing with multiple boolean arrays covering
only some axes, e.g. `A[A[:, 4] > 5, A[3, :] < 6]`. I have tried to make this code replace
the boolean arrays with integer arrays that come from `idx.nonzero()` where `idx` is the 
boolean array index. That's what the numpy indexing description says should happen. But
there may be come corner cases near this that I haven't tested.

Timing:
I turned on the COO format for `Getset` benchmarks with the following results:
```
--                                            format                   
-------------------------- --------------------------------------------
   N     sparsity pattern    csr      csc      lil      dok      coo   
======= ================== ======== ======== ======== ======== ========
   1        different       57.5μs   57.9μs   661ns    4.45μs   45.6μs 
   1           same         15.1μs   15.0μs   667ns    4.39μs   41.7μs 
   10       different       112μs    112μs    16.8μs   21.7μs   57.0μs 
   10          same         26.3μs   26.5μs   16.6μs   21.9μs   53.2μs 
   100       different       454μs    449μs    20.1μs   34.6μs   82.0μs 
   100          same         27.0μs   26.8μs   20.9μs   38.3μs   112μs  
  1000      different       2.73ms   2.69ms   62.9μs    n/a     409μs  
  1000         same         33.6μs   33.9μs   70.5μs    n/a     3.93ms 
 10000      different       5.22ms   5.25ms   860μs     n/a     4.81ms 
 10000         same         155μs    160μs    798μs     n/a     321ms  
======= ================== ======== ======== ======== ======== ========
```


benchmarks/benchmarks/sparse.py
```diff
@@ -292,6 +293,8 @@ def setup(self, N, sparsity_pattern, format):
             v = float(v)
 
         base = A.asformat(format)
+        if format == 'coo':
+            base = coo_array(A)
```
perimosocordiae: Let's make a new sparse_array.py next to this sparse.py benchmark, and use sparray versions of all the formats instead of spmatrix. That way, we can add COO to that one without special-casing.

scipy/sparse/_coo.py
```diff
@@ -524,6 +526,271 @@ def _with_data(self, data, copy=True):
             coords = self.coords
         return self.__class__((data, coords), shape=self.shape, dtype=data.dtype)
 
+    def __getitem__(self, key):
+        index, new_shape, arr_int_pos, none_pos = self._validate_indices(key)
```
perimosocordiae: Let's move `_validate_indices` and `_asindices` out of IndexMixin, as they don't really need access to anything but the shape.

scipy/sparse/_coo.py
```diff
@@ -524,6 +526,271 @@ def _with_data(self, data, copy=True):
...
+                # Not contiguous. Put all array coords at front
+                new_coords = new_arr_coords + new_coords
+
+        if none_pos:
+            if new_coords:
+                coord_like = new_coords[0]
+            else:
+                coord_like = np.zeros(len(new_data), dtype=self.coords[0].dtype)
+            for i in none_pos:
+                new_coords.insert(i, np.zeros_like(coord_like))
```
perimosocordiae: This makes one extra array. We could move the zeros_like to line 593, and then copy on insert `if i != 0`

perimosocordiae (changes requested):

jorenham:
I'll probably be needing some help when I have to annotate this in scipy-stubs for the next scipy release 😅

scipy/sparse/_coo.py
dschult: Done in #23267

scipy/sparse/tests/test_base.py
```diff
@@ -3512,37 +3517,42 @@ def test_fancy_assign_empty(self):
...
-            assert_raises(IndexError, S.__getitem__, (I,J_bad))
+            assert_raises(IndexError, S.__getitem__, (I_bad, J))
+            assert_raises(IndexError, S.__getitem__, (I, J_bad))
 
-            # This would generate 3-D arrays -- not supported
-            assert_raises(IndexError, S.__getitem__, ([I, I], slice(None)))
-            assert_raises(IndexError, S.__getitem__, (slice(None), [J, J]))
+            if S.format != 'coo':
+                    assert_raises(IndexError, S.__getitem__, ([I, I], slice(None)))
+                    assert_raises(IndexError, S.__getitem__, (slice(None), [J, J]))
```
perimosocordiae:
```suggestion
            if S.format != 'coo':
                assert_raises(IndexError, S.__getitem__, ([I, I], slice(None)))
                assert_raises(IndexError, S.__getitem__, (slice(None), [J, J]))
```

scipy/sparse/_coo.py
```diff
@@ -524,6 +526,291 @@ def _with_data(self, data, copy=True):
...
+                # Not contiguous. Put all array coords at front
+                new_coords = new_arr_coords + new_coords
+
+        if none_pos:
+            if new_coords:
+                coord_like = np.zeros_like(new_coords[0])
+            else:
+                coord_like = np.zeros(len(new_data), dtype=self.coords[0].dtype)
+            new_coords.insert(none_pos[0], coord_like)
+            if len(none_pos) > 1:
```
perimosocordiae: Technically this if condition isn't required. If you have a list with one element, then `none_pos[1:]` will give an empty list.

scipy/sparse/_coo.py
```diff
@@ -524,6 +526,291 @@ def _with_data(self, data, copy=True):
...
+                for i in none_pos[1:]:
+                    new_coords.insert(i, coord_like.copy())
+        return coo_array((new_data, new_coords), shape=new_shape, dtype=self.dtype)
+
+    def __setitem__(self, key, x):
+        # enact self[key] = x
+        index, new_shape, arr_int_pos, none_pos = _validate_indices(
+            key, self.shape, self.format
+        )
+        print(f"{key=} {index=} {new_shape=}")
```
perimosocordiae: Don't forget to remove this :-)

scipy/sparse/_coo.py
```diff
@@ -524,6 +526,291 @@ def _with_data(self, data, copy=True):
...
+
+        # remove None's at beginning of index. Should not impact indexing coords
+        # and will mistakenly align with x_coord columns if not removed.
+        if none_pos:
+            new_shape = list(new_shape)
+            for j in none_pos[::-1]:
+                new_shape.pop(j)
+            new_shape = tuple(new_shape)
+
+        # get coords and data from x
```
perimosocordiae, line 626: This block would be nice to have as one or two helper functions, though I'm having trouble coming up with concise names for them:

```python
if issparse(x):
    x_coords, x_data = _get_sparse_coords_and_data(x, self.dtype)
else:
    x_coords, x_data = _get_dense_coords_and_data(x, self.dtype)
if len(x_data) == 0:
    return  # Nothing to set.
  ```

perimosocordiae (approved):
This is a big one! I have some minor comments, but overall this looks great.

scipy/sparse/_coo.py
dschult: Thanks for the review and the suggestions!  This is a big one, for sure!

Checking `len(x_data)` doesn't work because x might be all zeros -> `len(x_data) == 0`.
So I moved that check into the if-structure and implemented the two helper functions.
Names seem fine to me: long but not used much and won't cause line-wrapping. 

Should be ready to go. :)

j-bowhay:
Looks impressive! 

> * I believe I've got the correct handling of `newaxis/None` for `__setitem__`. It works as
>   tested and seems to match NumPy behavior. But it is possible I am missing some corner
>   cases. As far as I can tell, using `None` in an index for assignment does not affect
>   the part of the array being assigned to. It only affects the allowed shape of the incoming data.

I wonder if this would be a good use case for [hypothesis](https://hypothesis.works/) to compare the behavior against numpy rather than writing explicit test cases?

dschult:
Are there other oplaces in scipy that use Hypothesis? I'm having a hard time searching because of all the stats hypothesis tests. :)

perimosocordiae:
Merged 🎉

PR #22911, opened 2025-04-29
ENH: signal: Array API for filter design functions

ev-br:
#### Reference issue

towards https://github.com/scipy/scipy/issues/20678

#### What does this implement/fix?

Make "filter design" functions Array API compatible:

- [x] butter, buttap, buttord
- [x] cheby1, cheb1ap, cheb1ord
- [x] cheby2, cheb2ap, cheb2ord
- [x] bessel, besselap
- [x] ellip, ellipap, ellipord 

#### Additional information

This PR is on top of https://github.com/scipy/scipy/pull/22896, so either manually close that one when this one lands, or land gh-22896 first and rebase this one.

PR #23222, opened 2025-06-23
BUG: signal.tf2sos: fix a new `ComplexWarning`

ev-br:
#### Reference issue

reported in https://github.com/scipy/scipy/issues/23221#issuecomment-2997264496

fixes gh-23221, fixes gh-23277

Cherry-pick a part of gh-22911 to fix an issue reported for scipy==1.16.0 in gh-23221

https://github.com/scipy/scipy/issues/23221#issuecomment-2997264496

#### Additional information

The PR is against maintenance/1.16.x since the problem is already fixed in main.

### DIFF
--- a/scipy/signal/_filter_design.py
+++ b/scipy/signal/_filter_design.py
@@ -60,6 +60,17 @@ def _is_int_type(x):
         return True
 
 
+def _real_dtype_for_complex(dtyp, *, xp):
+    if xp.isdtype(dtyp, 'real floating'):
+        return dtyp
+    if dtyp == xp.complex64:
+        return xp.float32
+    elif dtyp == xp.complex128:
+        return xp.float64
+    else:
+        raise ValueError(f"Unknown dtype {dtyp}.")
+
+
 # https://github.com/numpy/numpy/blob/v2.2.0/numpy/_core/function_base.py#L195-L302
 def _logspace(start, stop, num=50, endpoint=True, base=10.0, dtype=None, *, xp):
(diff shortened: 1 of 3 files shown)


j-bowhay (approved):

tylerjereddy (commented):
To confirm--is there no "regression" test? We can't check for absence of warning if that is the issue? Or the plan is to fix more correctly later (I thought I saw something about that in the ticket).

ev-br:
> To confirm--is there no "regression" test? We can't check for absence of warning if that is the issue?

OK, I added a test which will fail if the OP case from gh-23221 emits eny warnings.

>  Or the plan is to fix more correctly later (I thought I saw something about that in the ticket).

The issue is specific to scipy==1.6.0 and is already fixed in main.

For context: a series of array API PRs got split across the branching point; it turns out that what landed in 1.16.0 emits a warning in some cases; the fix is in main only; this PR backports the fix to maintenance/1.16.x.

ev-br:
Have added one more partial cherry-pick to fix https://github.com/scipy/scipy/issues/23277  (please let me know if you prefer a separate PR Tyler).

tylerjereddy:
I confirmed locally that both of the newly-added tests fail when the source changes in the PR are reverted.

There was a small typo in one of the tests talking about SciPy `1.6.0` instead of `1.16.0`. I pushed in a quick fix for that with CI skipped, since CI was already passing, and I'll now proceed with the backport merge as we're interested in getting moving with SciPy `1.16.1`.

Thanks Evgeni

PR #23223, opened 2025-06-23
MAINT: port `_test_fortran.f` to Python functions in `test_fortran.py`

czgdp1807:

In this PR, I am porting the code written in `_test_fortran.f` to `_test_fortran_py.py`. Basically, moving to Python from Fortran. AFAICT, this is an effort towards moving away from Fortran at places where Python can be used.

If needed I can use C++ instead of pure Python. Let me know.

@rgommers Let me know if this is what's expected.


rgommers (commented):
Thanks @czgdp1807. A few quick comments:

- Why add a new `.py` file for this? These function can simply live in `tests/test_fortran.py` right?
- It'd be good to remove the `.f` and `.pyf` files.

@ev-br would you be able to review this PR? I won't have time for the next 2 weeks at least.

czgdp1807:
> * Why add a new `.py` file for this? These function can simply live in `tests/test_fortran.py` right?

I don't have any strong reason to do this. Just did it for following the structure. I am +1 with having these Python functions in the same file.

> * It'd be good to remove the `.f` and `.pyf` files.

Sure thing. Removing in a few minutes.

ilayn:
Indeed same file would be the right place for this. Marked this done in #18566 as well.

scipy/io/tests/test_fortran.py
```diff
@@ -131,6 +130,75 @@ def test_fortranfile_write_mixed_record(tmpdir):
         for aa, bb in zip(a, b):
             assert_equal(bb, aa)
 
+def read_unformatted_double(m, n, k, filename):
+    """
+    Read a Fortran-style unformatted binary file written with a single write() call,
```
ev-br, line 135: If we add this, it'd be helpful to add a link to the explicit fortran code which reads the file, https://github.com/scipy/scipy/blob/maintenance/1.15.x/scipy/io/_test_fortran.f

ev-br (approved):
Previously, tests were checking that a file written by `FortranFile.write` roundtrips with both `FortranFile.read` _and_ a native Fortran file reading code. This PR removes the Fortran reader and replaces it with pure python code.
So this essentially double-checks `FortranFile.read` with an independent python implementation. 

The code added here looks sufficiently different from https://github.com/scipy/scipy/blob/v1.16.0/scipy/io/_fortran.py#L33-L354, so it can be regareded as an independent implementation.

The format for Fortran I/O is very unlikely to change, so anything which worked in scipy 1.5 is going to keep working in the future, so removing a native Fortran reader is probably not essential. I suppose we could just remove the reader and rely on roundtripping between `FortranFile` reads and writes. 
Now that there is an extra implementation, it does not hurt either. If anything, it could serve as simple demo of using `np.fromfile` on Fortran-writte files.

scipy/io/tests/test_fortran.py
czgdp1807: I think you mean to refer the old Fortran code in these functions so that developers know what was the original Fortran logic for this Python implementation. Am I right?

ev-br: Yes, exactly. If there's ever a failure in this test, the link would help debugging it.

czgdp1807:
@ev-br I have addressed your review. It is ready to go now.

ev-br:
LGTM, thank you @czgdp1807

PR #23233, opened 2025-06-25
ENH: Implement support for `bc_type="periodic"` in `make_splrep` and `make_splprep`

czgdp1807:

This PR moves the periodic B-spline fitting logic from FITPACK’s old Fortran code to a modern C++ + Python implementation. It helps reduce the dependency on legacy Fortran and brings the periodic case in line with the rest of the SciPy spline infrastructure.

The new code is exposed via `bc_type="periodic"` in both `make_splrep` and `make_splprep`. It follows the same logic as FITPACK, using extra matrix space to correctly handle basis functions that wrap around at the edges. Proper tests have been added, and the core algorithm is explained in the code with clear comments and ASCII diagrams to improve readability and maintainability.

#### Additional information

cc: @ev-br @rgommers

### DIFF
--- a/scipy/interpolate/_bsplines.py
+++ b/scipy/interpolate/_bsplines.py
@@ -1880,7 +1880,7 @@ def make_lsq_spline(x, y, t, k=3, w=None, axis=0, check_finite=True, *, method="
         c = cho_solve_banded((cho_decomp, lower), rhs.reshape(m, -1), overwrite_b=True,
                              check_finite=check_finite).reshape(rhs.shape)
     elif method == "qr":
-        _, _, c = _lsq_solve_qr(x, yy, t, k, w)
+        _, _, c, _, _ = _lsq_solve_qr(x, yy, t, k, w)
 
         if was_complex:
             c = c.view(complex)
@@ -1900,21 +1900,51 @@ def make_lsq_spline(x, y, t, k=3, w=None, axis=0, check_finite=True, *, method="
 # LSQ spline helpers #
 ######################
 
-def _lsq_solve_qr(x, y, t, k, w):
+def _lsq_solve_qr_for_root_rati_periodic(x, y, t, k, w):
     """Solve for the LSQ spline coeffs given x, y and knots.
 
     `y` is always 2D: for 1D data, the shape is ``(m, 1)``.
     `w` is always 1D: one weight value per `x` value.
 
     """
-    assert y.ndim == 2
-
     y_w = y * w[:, None]
-    A, offset, nc = _dierckx.data_matrix(x, t, k, w)
-    _dierckx.qr_reduce(A, offset, nc, y_w)         # modifies arguments in-place
-    c = _dierckx.fpback(A, nc, y_w)
+    # Ref: https://github.com/scipy/scipy/blob/maintenance/1.16.x/scipy/interpolate/fitpack/fpperi.f#L221-L238
+    R, H1, H2, offset, nc = _dierckx.data_matrix_periodic(x, t, k, w, False)
+    # Ref: https://github.com/scipy/scipy/blob/maintenance/1.16.x/scipy/interpolate/fitpack/fpperi.f#L239-L314
+    A1, A2, Z, p, _ = _dierckx.qr_reduce_periodic(
+        R, H1, H2, offset, nc, y_w, k,
+        len(t), True
+    )         # modifies arguments in-place
+    # Ref: https://github.com/scipy/scipy/blob/main/scipy/interpolate/fitpack/fpbacp.f
+    c, residuals, _ = _dierckx.fpbacp(A1, A2, Z, k, k, x, y, t, w)
+    return R, A1, A2, Z, y_w, c, p, residuals
 
-    return A, y_w, c
+
+def _lsq_solve_qr(x, y, t, k, w, periodic=False):
+    """Solve for the LSQ spline coeffs given x, y and knots.
+
+    `y` is always 2D: for 1D data, the shape is ``(m, 1)``.
+    `w` is always 1D: one weight value per `x` value.
+
+    """
+    y_w = y * w[:, None]
+    if not periodic:
+        A, offset, nc = _dierckx.data_matrix(x, t, k, w)
+        _dierckx.qr_reduce(A, offset, nc, y_w)         # modifies arguments in-place
+        c, residuals, fp = _dierckx.fpback(A, nc, x, y, t, k, w, y_w)
+        return A, y_w, c, fp, residuals
+    else:
+        # Ref: https://github.com/scipy/scipy/blob/maintenance/1.16.x/scipy/interpolate/fitpack/fpperi.f#L221-L238
+        R, H1, H2, offset, nc = _dierckx.data_matrix_periodic(x, t, k, w, False)
+        # Ref: https://github.com/scipy/scipy/blob/maintenance/1.16.x/scipy/interpolate/fitpack/fpperi.f#L239-L314
+        A1, A2, Z, fp = _dierckx.qr_reduce_periodic(
+            R, H1, H2, offset, nc, y_w, k,
+            len(t), False)         # modifies arguments in-place
+        # Ref: https://github.com/scipy/scipy/blob/main/scipy/interpolate/fitpack/fpbacp.f
+        c, residuals, _ = _dierckx.fpbacp(A1, A2, Z, k, k, x, y, t, w)
+        return R, y_w, c, fp, residuals
+
+
 
 
 #############################
@@ -2430,29 +2460,52 @@ def make_smoothing_spline(x, y, w=None, lam=None, *, axis=0):
 #  FITPACK look-alikes #
 ########################
 
-def fpcheck(x, t, k):
-    """ Check consistency of the data vector `x` and the knot vector `t`.
+def fpcheck(x, t, k, periodic=False):
+    """Check consistency of data vector `x` and knot vector `t`.
 
-    Return None if inputs are consistent, raises a ValueError otherwise.
+    Parameters
+    ----------
+    x : array_like, shape (m,)
+        1D sorted array of data points.
+    t : array_like, shape (n,)
+        1D non-decreasing knot vector.
+    k : int
+        Degree of the spline.
+    periodic : bool, optional
+        Whether the spline is periodic. Default is False.
+
+    Raises
+    ------
+    ValueError
+        If the configuration of `x`, `t`, and `k` violates any required condition.
     """
-    # This routine is a clone of the `fpchec` Fortran routine,
-    # https://github.com/scipy/scipy/blob/main/scipy/interpolate/fitpack/fpchec.f
-    # which carries the following comment:
+    # This routine is a unified clone of the FITPACK Fortran routines `fpchec.f`
+    # and `fpchep.f`:
+    # - https://github.com/scipy/scipy/blob/main/scipy/interpolate/fitpack/fpchec.f
+    # - https://github.com/scipy/scipy/blob/main/scipy/interpolate/fitpack/fpchep.f
     #
-    # subroutine fpchec verifies the number and the position of the knots
-    #  t(j),j=1,2,...,n of a spline of degree k, in relation to the number
-    #  and the position of the data points x(i),i=1,2,...,m. if all of the
-    #  following conditions are fulfilled, the error parameter ier is set
-    #  to zero. if one of the conditions is violated ier is set to ten.
-    #      1) k+1 <= n-k-1 <= m
-    #      2) t(1) <= t(2) <= ... <= t(k+1)
-    #         t(n-k) <= t(n-k+1) <= ... <= t(n)
-    #      3) t(k+1) < t(k+2) < ... < t(n-k)
-    #      4) t(k+1) <= x(i) <= t(n-k)
-    #      5) the conditions specified by schoenberg and whitney must hold
-    #         for at least one subset of data points, i.e. there must be a
-    #         subset of data points y(j) such that
-    #             t(j) < y(j) < t(j+k+1), j=1,2,...,n-k-1
+    # These routines verify the number and position of the knots t(j), j=1,...,n,
+    # of a spline of degree k, relative to the number and distribution of data points
+    # x(i), i=1,...,m. If all of the following conditions are fulfilled,
+    # validation passes.
+    #
+    # For non-periodic splines:
+    #   1) k+1 <= n-k-1 <= m
+    #   2) t(1) <= t(2) <= ... <= t(k+1)
+    #      t(n-k) <= t(n-k+1) <= ... <= t(n)
+    #   3) t(k+1) < t(k+2) < ... < t(n-k)
+    #   4) t(k+1) <= x(i) <= t(n-k)
+    #   5) Schoenberg-Whitney conditions hold: there exists a subset y(j) such that
+    #        t(j) < y(j) < t(j+k+1), for j = 1, ..., n-k-1
+    #
+    # For periodic splines:
+    #   1) k+1 <= n-k-1 <= m + k - 1
+    #   2) Same boundary knot monotonicity as above
+    #   3) Same strict interior knot increase as above
+    #   4) t(k+1) <= x(i) <= t(n-k)
+    #   5) Schoenberg-Whitney conditions must hold for *some periodic shift*
+    #        of the data sequence; i.e. wrapped data points x(i) must satisfy
+    #        t(j) < y(j) < t(j+k+1), j = k+1, ..., n-k-1
     x = np.asarray(x)
     t = np.asarray(t)
 
@@ -2464,9 +2517,14 @@ def fpcheck(x, t, k):
     nk1 = n - k - 1
 
     # check condition no 1
-    # c      1) k+1 <= n-k-1 <= m
-    if not (k + 1 <= nk1 <= m):
-        raise ValueError(f"Need k+1 <= n-k-1 <= m. Got {m = }, {n = } and {k = }.")
+    if periodic:
+        # c      1) k+1 <= nk1 <= m+k-1
+        if not (k + 1 <= nk1 <= m + k - 1):
+            raise ValueError(f"Need k+1 <= n-k-1 <= m+k-1. Got {m = }, {n = }, {k = }")
+    else:
+        # c      1) k+1 <= n-k-1 <= m
+        if not (k + 1 <= nk1 <= m):
+            raise ValueError(f"Need k+1 <= n-k-1 <= m. Got {m = }, {n = } and {k = }.")
 
     # check condition no 2
     # c      2) t(1) <= t(2) <= ... <= t(k+1)
@@ -2489,27 +2547,59 @@ def fpcheck(x, t, k):
         raise ValueError(f"Out of bounds: {x = } and {t = }.")
 
     # c  check condition no 5
-    # c      5) the conditions specified by schoenberg and whitney must hold
-    # c         for at least one subset of data points, i.e. there must be a
-    # c         subset of data points y(j) such that
-    # c             t(j) < y(j) < t(j+k+1), j=1,2,...,n-k-1
+    # c      5) the conditions specified by Schoenberg and Whitney must hold
+    # c         for at least one subset of data points y(j) such that
+    # c             t(j) < y(j) < t(j+k+1)
+    # c
+    # c         For non-periodic splines:
+    # c             j = 1, 2, ..., n-k-1 (i.e., j in [1, n-k-1])
+    # c             The data points must lie strictly inside some B-spline supports.
+    # c
+    # c         For periodic splines:
+    # c             j = k+1, ..., n-k-1
+    # c             The condition must hold for a wrapped subset of the data points,
+    # c             i.e., there exists a cyclic shift of the data such that
+    # c                 t(j) < x(i) < t(j+k+1)
+    # c             holds for all j in that range. The test must account for the
+    # c             periodic domain length: per = t(n-k) - t(k+1), and wrap around x(i)
+    # c             as x(i) + per if needed.
     mesg = f"Schoenberg-Whitney condition is violated with {t = } and {x =}."
 
-    if (x[0] >= t[k+1]) or (x[-1] <= t[n-k-2]):
+    if periodic:
+        per = t[n - k - 1] - t[k]
+        m1 = m - 1
+        for shift in range(1, m):
+            for j in range(k, nk1):
+                tj = t[j]
+                tl = t[j + k + 1]
+                found = False
+                for i in range(shift, shift + m1 + 1):
+                    idx = i if i < m else i - m
+                    xi = x[idx] + (0 if i < m else per)
+                    if tj < xi < tl:
+                        found = True
+                        break
+                if not found:
+                    break
+            else:
+                return
         raise ValueError(mesg)
+    else:
+        if (x[0] >= t[k+1]) or (x[-1] <= t[n-k-2]):
+            raise ValueError(mesg)
 
-    m = x.shape[0]
-    l = k+1
-    nk3 = n - k - 3
-    if nk3 < 2:
+        m = x.shape[0]
+        l = k+1
+        nk3 = n - k - 3
+        if nk3 < 2:
+            return
+        for j in range(1, nk3+1):
+            tj = t[j]
+            l += 1
+            tl = t[l]
+            i = np.argmax(x > tj)
+            if i >= m-1:
+                raise ValueError(mesg)
+            if x[i] >= tl:
+                raise ValueError(mesg)
         return
-    for j in range(1, nk3+1):
-        tj = t[j]
-        l += 1
-        tl = t[l]
-        i = np.argmax(x > tj)
-        if i >= m-1:
-            raise ValueError(mesg)
-        if x[i] >= tl:
-            raise ValueError(mesg)
-    return
--- a/scipy/interpolate/_fitpack_repro.py
+++ b/scipy/interpolate/_fitpack_repro.py
@@ -24,7 +24,8 @@ import numpy as np
 from scipy._lib._array_api import array_namespace, concat_1d
 
 from ._bsplines import (
-    _not_a_knot, make_interp_spline, BSpline, fpcheck, _lsq_solve_qr
+    _not_a_knot, make_interp_spline, BSpline, fpcheck, _lsq_solve_qr,
+    _lsq_solve_qr_for_root_rati_periodic, _periodic_knots
 )
 from . import _dierckx      # type: ignore[attr-defined]
 
@@ -41,10 +42,7 @@ TOL = 0.001
 MAXIT = 20
 
 
-def _get_residuals(x, y, t, k, w):
-    # FITPACK has (w*(spl(x)-y))**2; make_lsq_spline has w*(spl(x)-y)**2
-    w2 = w**2
-
+def _get_residuals(x, y, t, k, w, periodic=False):
     # inline the relevant part of
     # >>> spl = make_lsq_spline(x, y, w=w2, t=t, k=k)
     # NB:
@@ -56,19 +54,21 @@ def _get_residuals(x, y, t, k, w):
     #         * For 2D (parametric=True), the summation is actually how the
     #           'residuals' are defined, see Eq. (42) in Dierckx1982
     #           (the reference is in the docstring of `class F`) below.
-    _, _, c = _lsq_solve_qr(x, y, t, k, w)
-    c = np.ascontiguousarray(c)
-    spl = BSpline(t, c, k)
-    residuals = _compute_residuals(w2, spl(x), y)
-    fp = residuals.sum()
-    if np.isnan(fp):
+    _, _, _, fp, residuals = _lsq_solve_qr(x, y, t, k, w, periodic=periodic)
+    if np.isnan(residuals.sum()):
         raise ValueError(_iermesg[1])
     return residuals, fp
 
+def _validate_bc_type(bc_type):
+    if bc_type is None:
+        return "not-a-knot"
 
-def _compute_residuals(w2, splx, y):
-    delta = ((splx - y)**2).sum(axis=1)
-    return w2 * delta
+    if bc_type not in ("not-a-knot", "periodic"):
+        raise ValueError("Only 'not-a-knot' and 'periodic' "
+                         "boundary conditions are recognised, "
+                         f"found {bc_type}")
+
+    return bc_type
 
 
 def add_knot(x, t, k, residuals):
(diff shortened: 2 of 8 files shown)


czgdp1807:
All tests pass. Marking it ready for review. :-).

rgommers:
Can you remove the unrelated whitespace changes to `.github/workflows` please?

rgommers:
Also a minor comment on commit message prefixes: the right ones are `ENH`, `MAINT`, `TST`. `DEV` is specifically for development workflow related changes.

scipy/interpolate/tests/test_fitpack.py
```diff
@@ -117,7 +128,11 @@ def err_est(k, d):
     def test_smoke_splrep_splev(self):
         self.check_1(s=1e-6)
         self.check_1(b=1.5*np.pi)
+
+    @pytest.mark.skipif(sys.maxsize <= 2**32, reason="Segfaults on 32-bit system")
```
ev-br: Now would be a good time to look at this segfault, I'd think.

czgdp1807:
> Can you remove the unrelated whitespace changes to `.github/workflows` please?

Sure. It's in a separate commit. I added that commit, just in case if its needed. I will drop it from the commit history here.

> Also a minor comment on commit message prefixes: the right ones are `ENH`, `MAINT`, `TST`. `DEV` is specifically for development workflow related changes.

I see. I will use `ENH`, `TST` in this PR.

scipy/interpolate/tests/test_fitpack.py
czgdp1807: Sure. I would be needing a 32-bit system for it. I am not sure if it's easily accessible. I tried on docker but it's hard to debug there. Azure doesn't have 32-bit system AFAIK.

czgdp1807:
Please let me know if there are any other reviews.

scipy/interpolate/tests/test_fitpack.py
ev-br: Ok to keep it an xfail/skip if it's hard to debug.

rgommers: @ev-br I just had a chat with @czgdp1807 and made the point that it isn't, since it'll be blocking for removing the equivalent Fortran code I think? 

If it's a very localized segfault, it'd need to be changed in the code to raise an exception I'd think, rather than only hiding the segfault by skipping the test.

czgdp1807: Sure. I will debug it on Linux (with a docker, it's hard to debug on Docker but still I will try). Meanwhile if you have any other reviews, feel free to let me know. I will address those as well.

czgdp1807: I removed all the `@pytest.mark.skipif` for 32-bit systems. They are working already (verified on docker) and in https://github.com/czgdp1807/scipy/pull/20.

There is `scipy/interpolate/tests/test_bsplines.py::TestMakeSplrepBase::test_s_too_small` which was failing on 32-bit system. This was not due to `make_splrep(..., bc_type='periodic')`. In fact, `splrep(..., per=1)` was giving `NaN` coefficients when `s < 1e-31`. `make_splrep(..., bc_type='periodic')` was working totally fine with correct results.

So, I decided to use `s = 1e-31` if the system is 32-bit and `bc_type="periodic"` so that `splrep(..., per=1)` gives non-NaN coefficients. IMO, fixing `splrep` wouldn't be worth it because eventually we will be shifting to `make_splrep` which is working fine in all cases. 

@rgommers @ev-br What do you folks say?

Please feel free to resolve this review if you agree with my decision.

scipy/interpolate/tests/test_bsplines.py
```diff
@@ -3365,94 +3367,226 @@ def __call__(self, p):
...
+        self._test_default_s(x, y, k)
+
+    @pytest.mark.thread_unsafe
+    def test_s_too_small(self):
+        # both splrep and make_splrep warn that "s too small": ier=2
+        s = 1e-50
+        if self.bc_type == 'periodic':
+            x = np.linspace(0, 2*np.pi, 14)
+            y = np.sin(x)
+            s = 1e-31 if sys.maxsize <= 2**32 else 1e-50
```
ev-br: Nit: This should not be needed. s=1e-31 should be "too small" for all arches and all boundary conditions.

scipy/interpolate/tests/test_fitpack.py
ev-br: To double-check the status: 
- the "old" `splrep` was segfaulting (?) on 32-bit system with `s < 1e-31`.
- the "new" `make_splrep` was not segfaulting on either system or boundary conditions.
- this PR now does not have any `skipif`s.

If this is indeed correct, then it's good as is, modulo removing an unneeded branching, commented below.

scipy/interpolate/tests/test_bsplines.py
czgdp1807: I see. So I will set `s=1e-31` for `test_s_too_small`.

scipy/interpolate/tests/test_fitpack.py
czgdp1807: > * the "old" `splrep` was segfaulting (?) on 32-bit system with `s < 1e-31`.

Not segfaulting but giving `NaN` coefficients, which is not expected.



> * the "new" `make_splrep` was not segfaulting on either system or boundary conditions.

Correct. The new `make_splrep` API doesn't segfault in any case. It also produces non-NaN results in all cases and both 32-bit and 64-bit systems.



> * this PR now does not have any `skipif`s.

AFAIK, yes no `skipif`s.

czgdp1807:
Seems like, `splrep` is giving NaN with `s=1e-31` if `per=1` (for periodic splines). Seems like `splrep` is flaky. We might need to increase `s` to found a value for `splrep` which always works.

czgdp1807:
@ev-br @rgommers With `s=1e-30`, `splrep` works fine. `make_splrep` (the one I have updated in this PR) is perfect with any value of `s`. This is ready to go from my end.

czgdp1807:
@ev-br @rgommers This PR is complete from my end. No `skipif`s for tests. The problem was anyways in older Fortran backed APIs i.e., `splrep`. `make_splrep` is 100% working on all tests (even where `splrep` fails).

Let me know if anything else is to be done. Let's get this in otherwise.

scipy/interpolate/tests/test_bsplines.py
```diff
@@ -3365,94 +3366,225 @@ def __call__(self, p):
...
+        # both splrep and make_splrep warn that "s too small": ier=2
+        s = 1e-30
+        if self.bc_type == 'periodic':
+            x = np.linspace(0, 2*np.pi, 14)
+            y = np.sin(x)
+        else:
+            x = np.arange(14)
+            y = x**3
+
+        with suppress_warnings() as sup:
```
ev-br: Could you replace `suppress_warnings` similar to https://github.com/scipy/scipy/pull/23275 please?

scipy/interpolate/tests/test_bsplines.py
```diff
@@ -3578,6 +3700,73 @@ def test_s0_vs_not(self):
...
+            make_splrep(x, y, s=1e-8, bc_type=self.bc_type)
+
+    def test_make_splrep_periodic_m_eq_2_k_eq_1(self):
+        # Two data points (m = 2)
+        x = np.array([0.0, 1.0])
+        y = np.array([5.0, 5.0])  # constant function
+        w = np.array([1.0, 1.0])
+
+        # Degree 1 periodic spline
+        spl = make_splrep(x, y, w=w, k=1, bc_type="periodic", s=1e-50)
```
ev-br: Minor: Can parametrize this test for `s=[0, None, 1e-50]` to make sure it's consistent.

ev-br (approved):
This is a great feature. Converting 40-year old F77 code is a no small feat. Well done @czgdp1807! 

This code went through several review rounds already at https://github.com/czgdp1807/scipy/pull/19.

While there are always things to improve, I think this is now at a state where we can land this and improve iteratively at leisure.

The PR is LGTM modulo a small fix-up for a recent change to remove `np.suppress_warnings`.
Unless someone wants more time to review, I'd merge this PR as soon as `suppress_warnings` is gone. Please speak up if you would like more time to review.

scipy/interpolate/_bsplines.py
```diff
@@ -1871,11 +1871,45 @@ def _lsq_solve_qr(x, y, t, k, w):
...
+
+
+def _lsq_solve_qr(x, y, t, k, w, periodic=False):
+    """Solve for the LSQ spline coeffs given x, y and knots.
+
+    `y` is always 2D: for 1D data, the shape is ``(m, 1)``.
+    `w` is always 1D: one weight value per `x` value.
+
+    """
+    assert y.ndim == 2
```
j-bowhay: nit but I don't think we should have `assert`s in user facing code

scipy/interpolate/_fitpack_repro.py
```diff
@@ -139,10 +139,15 @@ def _validate_inputs(x, y, w, k, s, xb, xe, parametric):
     if xe is None:
         xe = max(x)
 
+    if periodic and not np.allclose(y[0], y[-1], atol=1e-15):
+        raise ValueError("First and last points does not match while "
+                         "periodic case expected")
```
j-bowhay: nit, I am not sure the grammar here is correct
```suggestion
        raise ValueError("First and last points does not match which is required "
                         "for `bc_type='periodic'`.")
```

scipy/interpolate/_fitpack_repro.py
```diff
@@ -163,6 +168,16 @@ def generate_knots(x, y, *, w=None, xb=None, xe=None, k=3, s=0, nest=None):
         The smoothing factor. Default is ``s = 0``.
     nest : int, optional
         Stop when at least this many knots are placed.
+          ends are equivalent.
+    bc_type : str, optional
+        Boundary conditions.
+        Default is None.
```
j-bowhay: I think it would be clearer if it just said `not-a-knot` as this is what `None` corresponds with

scipy/interpolate/_fitpack_repro.py
```diff
@@ -765,6 +1037,20 @@ def make_splrep(x, y, *, w=None, xb=None, xe=None, k=3, s=0, t=None, nest=None):
         The actual number of knots returned by this routine may be slightly
         larger than `nest`.
         Default is None (no limit, add up to ``m + k + 1`` knots).
+    periodic : bool, optional
+        If non-zero, data points are considered periodic with period ``x[m-1]`` -
+        ``x[0]`` and a smooth periodic spline approximation is returned. Values of
+        ``y[m-1]`` and ``w[m-1]`` are not used.
+        The default is zero, corresponding to boundary condition 'not-a-knot'.
```
j-bowhay: Can we refer to True/False rather than non-zero/zero

scipy/interpolate/_fitpack_repro.py
```diff
@@ -765,6 +1037,20 @@ def make_splrep(x, y, *, w=None, xb=None, xe=None, k=3, s=0, t=None, nest=None):
...
+    bc_type : str, optional
+        Boundary conditions.
+        Default is None.
```
j-bowhay: same comment as before

scipy/interpolate/_fitpack_repro.py
```diff
@@ -906,6 +1198,15 @@ def make_splprep(x, *, w=None, u=None, ub=None, ue=None, k=3, s=0, t=None, nest=
         The actual number of knots returned by this routine may be slightly
         larger than `nest`.
         Default is None (no limit, add up to ``m + k + 1`` knots).
+    bc_type : str, optional
+        Boundary conditions.
+        Default is None.
```
j-bowhay: here too

scipy/interpolate/_fitpack_repro.py
```diff
@@ -219,50 +234,111 @@ def generate_knots(x, y, *, w=None, xb=None, xe=None, k=3, s=0, nest=None):
...
     """
+    bc_type = _validate_bc_type(bc_type)
+    periodic = bc_type == 'periodic'
+
     if s == 0:
         if nest is not None or w is not None:
             raise ValueError("s == 0 is interpolation only")
-        t = _not_a_knot(x, k)
+        if periodic:
+            t = _periodic_knots(x, k)
```
j-bowhay, line 249: Can you add test coverage for this branch

scipy/interpolate/_fitpack_repro.py
```diff
@@ -504,16 +583,181 @@ def __call__(self, p):
...
+        # If QR factorization or auxiliary matrices (A1, A2, Z) are not provided,
+        # compute them via least squares on the data (x,y) with weights w.
+        # These matrices come from fitting B-spline basis to data.
+        if ((A1 is None or A2 is None or Z is None) or
+            (R is None and Y is None)):
+            # _lsq_solve_qr_for_root_rati_periodic computes QR factorization of
+            # data matrix and returns related matrices.
+            # Refer: https://github.com/scipy/scipy/blob/maintenance/1.16.x/scipy/interpolate/fitpack/fpperi.f#L171-L215
+            (R, A1, A2, Z), _, _, _, _ = _lsq_solve_qr_for_root_rati_periodic(
+                x, y, t, k, w)
```
j-bowhay: Again flagging this branch as not hit by any test coverage

scipy/interpolate/tests/test_bsplines.py
```diff
@@ -3686,6 +3875,103 @@ def test_1D(self):
...
+        assert np.shape(y)[0] == 2
+
+        # assert the behavior of FITPACK's splrep
+        tck, u = splprep(y, s=s, per=1)
+        tck_a, u_a = splprep(np.asarray(y), s=s, per=1)
+        xp_assert_close(u, u_a, atol=s)
+        xp_assert_close(tck[0], tck_a[0], atol=1e-15)
+        assert len(tck[1]) == len(tck_a[1])
+        for c1, c2 in zip(tck[1], tck_a[1]):
+            xp_assert_close(c1, c2, atol=1e-15)
```
j-bowhay, line 4105:
```suggestion
        xp_assert_close(tck[1], tck_a[1], atol=1e-15)
```

j-bowhay (commented):
Thanks, some general comments

scipy/interpolate/tests/test_bsplines.py
czgdp1807: Here, `s=None` is an invalid input for `make_splrep`.

For `s = 0`, `w` should be `None` (a condition which pre-exists in `main` branch as well).

I parametrised this test accordingly and made some adjustments.

scipy/interpolate/_fitpack_repro.py
czgdp1807: I have addressed your other reviews. Will look into this one in a while.

czgdp1807: Addressed this as well.

czgdp1807:
@ev-br @j-bowhay I have addressed your reviews and the CI seems to pass. Please let me know if there is anything else to do in this PR for final merging.

P.S. If there are more minor changes, let's do it in future PRs and let's get it in? Anyways I am fine both ways. 

Thank you for the reviews.

czgdp1807:
@ev-br @j-bowhay Let's merge this PR if everything looks good?

dschmitz89:
Not a blocker but these two functions are the last ones in `interpolate` without examples acc. to https://github.com/scipy/scipy/issues/7168. Can we use the opportunity to add them (with and without periodic boundary conditions for example)?

ev-br:
These can probably could copy examples from `splrep` and `splprep`. Here or in a separate PR.

czgdp1807:
> Here or in a separate PR.

Can we do this in a separate PR? The reason being this one is waiting for a very long time so if there is an option to address the examples related stuff in a separate PR then it would be great.

Thanks.

dschmitz89:
> > Here or in a separate PR.
> 
> Can we do this in a separate PR? The reason being this one is waiting for a very long time so if there is an option to address the examples related stuff in a separate PR then it would be great.
> 
> Thanks.

Follow up PR is perfectly fine. Note that four weeks is not a long waiting time for a PR of this size in open source though. Reviewing 2000 lines of numerical linear algebra in C++ is not a simple task.

czgdp1807:
Sure. Actually, I opened this PR in my own fork as well, where it was thoroughly reviewed by @ev-br. Let’s wait for a couple of weeks more to see if there are any critical reviews.

dschmitz89:
> Sure. Actually, I opened this PR in my own fork as well, where it was thoroughly reviewed by @ev-br. Let’s wait for a couple of weeks more to see if there are any critical reviews.

Ah, did not know that of course, sorry if my reply came off the wrong way.

ev-br:
Indeed, we iterated with Gagan on his fork for a while, where the PR went through several review rounds.
This means the PR was open for a "long time" for Gagan and me, but is only open for 4 weeks for everybody else. We should give it a while as a scipy PR, so that all interested parties have time to take a look. This is what we get for brewing a patch in a fork @czgdp1807.

Therefore, while I personally am happy with the current state of the PR, I'm not going to rush merging this until @dschmitz89 , @j-bowhay and/or other interested folks indicate they are happy with landing this or sufficient time has passed for folks to take a look.

Re: docstring updates, that separate PR would be great to have and can be merged independently of this one IMO.

j-bowhay:
I don't have any further comments but I haven't check the algorithm side of things so I probably won't formally "approve" (although I trust @ev-br here!)

czgdp1807:
Sure let's wait for two weeks more to give this PR sufficient time. Thank you.

dschmitz89:
> Indeed, we iterated with Gagan on his fork for a while, where the PR went through several review rounds. This means the PR was open for a "long time" for Gagan and me, but is only open for 4 weeks for everybody else. We should give it a while as a scipy PR, so that all interested parties have time to take a look. This is what we get for brewing a patch in a fork @czgdp1807.
> 
> Therefore, while I personally am happy with the current state of the PR, I'm not going to rush merging this until @dschmitz89 , @j-bowhay and/or other interested folks indicate they are happy with landing this or sufficient time has passed for folks to take a look.
> 
> Re: docstring updates, that separate PR would be great to have and can be merged independently of this one IMO.

I trust your judgement here as well @ev-br , I am not at all familiar with the math involved here.

ev-br:
Conceptually the math here is not rocket science: it's basically solving an LSQ problem using an unpivoted QR factorization of the design matrix. One complication is that the matrix is banded, thus the QR procedure is implemented manually for a specific matrix storage scheme; The really tricky thing that Gagan pulled in this PR is that periodicity makes the matrix banded + corrections, and these corrections need to be tracked during the QR reduction. Again conceptually it is not too dissimilar from dealing with cyclic tridiagonal systems, but details are hairy.

So it'd be much appreciated if you feel like giving it a read-through @dschmitz89; no pressure though.

dschmitz89:
> Conceptually the math here is not rocket science: it's basically solving an LSQ problem using an unpivoted QR factorization of the design matrix. One complication is that the matrix is banded, thus the QR procedure is implemented manually for a specific matrix storage scheme; The really tricky thing that Gagan pulled in this PR is that periodicity makes the matrix banded + corrections, and these corrections need to be tracked during the QR reduction. Again conceptually it is not too dissimilar from dealing with cyclic tridiagonal systems, but details are hairy.
> 
> So it'd be much appreciated if you feel like giving it a read-through @dschmitz89; no pressure though.

You should not trust me about numerical linear algebra, never done that myself, so I cannot do a real review.

From 10000 feet, what I would like to see are some crude benchmarks/comparisons between the legacy FITPACK and the reimplementation regarding accuracy and performance. What are good benchmark metrics here? Difference between the estimated coefficients and knot positions? Simply comparing the data and the interpolation is probably not good as it would reward overfitting (I believe FITPACK uses something to control smoothness and splines are anyway good for this).

czgdp1807:
I would like to highlight that the implementation which I have done is same as the FITPACK one. It's just re-organised between Python and C++ so that it's maintainable, readable and easy to debug. It's lesser convoluted than the FITPACK one. However the underlying is algorithm is same as FITPACK.

dschmitz89:
> I would like to highlight that the implementation which I have done is same as the FITPACK one. It's just re-organised between Python and C++ so that it's maintainable, readable and easy to debug. It's lesser convoluted than the FITPACK one. However the underlying is algorithm is same as FITPACK.

Rewriting this monster is highly appreciated. My comment about benchmarking simply comes from the experience that such rewrites often cause subtle bugs that are not caught by our test suite (see for example the number of [issues](https://github.com/scipy/scipy/issues?q=is%3Aissue%20state%3Aclosed%20nnls) for a much simpler code,`nnls`).

ev-br:
Can a test of this sort, I think: 

```
In [65]: rng = np.random.default_rng(123)

In [66]: x = np.r_[0, np.sort(rng.uniform(0, 2*np.pi, size=n-2)), 2*np.pi]

In [67]: y = np.sin(x) + np.cos(x)

In [68]: for s in [10, 8, 5, 1, 1e-2]:
    ...:     t, c, k = splrep(x, y, s=s, per=False)
    ...:     spl = make_splrep(x, y, s=s, bc_type="not-a-knot")
    ...:     assert len(t) == len(spl.t)
    ...:     xp_assert_close(spl.t, t, atol=1e-15)
    ...:     xp_assert_close(spl.c, c[:-k-1], atol=1e-15)      # atol might be optimistic
    ...: 

In [69]: for s in [10, 8, 5, 1, 1e-2]:
    ...:     print("s = ", s)
    ...:     t, c, k = splrep(x, y, s=s, per=True)
    ...:     spl = make_splrep(x, y, s=s, bc_type="periodic")
    ...:     assert len(t) == len(spl.t)
    ...:     xp_assert_close(spl.t, t, atol=1e-15)
    ...:     xp_assert_close(spl.c, c[:-k-1], atol=1e-15)
```

We've seen differences in some cases for a non-periodic case due to fp noise, would be nice to understand what causes differences in the code added in this PR.

ev-br:
> What are good benchmark metrics here? Difference between the estimated coefficients and knot positions? Simply comparing the data and the interpolation is probably not good as it would reward overfitting (I believe FITPACK uses something to control smoothness and splines are anyway good for this).

Yes, knots and coefficients should match to a reasonable accuracy, I'd say. The line between "reasonable"  and "not reasonable" is a bit blurry though, and it's possible to invent ill-conditioned cases where heuristics fail (e.g. https://github.com/scipy/scipy/issues/22704).

czgdp1807:
With, https://github.com/scipy/scipy/pull/23233/commits/815a816c07cb2b21b90bdc07e26a3be16c78b2a3, the test in https://github.com/scipy/scipy/pull/23233#issuecomment-3117913129 passes successfully ( without any modifications to tolerances ). Seems like the subtle difference in coefficients between `splrep` and `make_splrep` was due to the differences in computation of `fp0` and `fpinf` between my code and FITPACK. I updated my code to use the same logic as FITPACK for computing `fp0` and `fpinf`. Once these two are correct then solver for `f(p=0) = s` gives exactly the same coefficients as `splrep`.

ev-br (changes requested):
> With, 815a816, the test in #23233 (comment) passes successfully

Great! Please add it as a test then. 

In fact, trying it locally, I see the test passing for `n=51` and failing for `n=11`. The key is not the length of data per se though; it's that the values of `s` should probe several different numbers of added knots. The differences start at a later stage: first few knots seem to agree, but then the knot vectors start to differ.
Could you please check this?

If it so happens that the differences are due to fp jitter in FITPACK-specific heuristics, a good test is to check that `make_splrep(x, y, t=<knots_from_splrep>).c` agrees with `c` from splrep. Please add tests of this kind for the combinations of data where the knot vectors differ and the difference cannot be fixed. 

 Speaking of which, I get an exception for the `x` and `y` from the comment above, `n=15` and

```
In [29]: t, c, k = splrep(x, y, s=1e-2, per=True)

In [30]: make_splrep(x, y, s=1e-2, bc_type="periodic", t=t).c

... snip

File ~/repos/scipy/scipy/build-install/usr/lib/python3.13/site-packages/scipy/interpolate/_fitpack_repro.py:940, in _make_splrep_impl(x, y, w, xb, xe, k, s, t, nest, periodic)
    938     t = list(gen)[-1]
    939 else:
--> 940     fpcheck(x, t, k)
    942 if t.shape[0] == 2 * (k + 1):
    943     # nothing to optimize
    944     _, _, c, _, _ = _lsq_solve_qr(x, y, t, k, w, periodic=periodic)

File ~/repos/scipy/scipy/build-install/usr/lib/python3.13/site-packages/scipy/interpolate/_bsplines.py:2494, in fpcheck(x, t, k)
   2491 # check condition no 1
   2492 # c      1) k+1 <= n-k-1 <= m
   2493 if not (k + 1 <= nk1 <= m):
-> 2494     raise ValueError(f"Need k+1 <= n-k-1 <= m. Got {m = }, {n = } and {k = }.")
   2496 # check condition no 2
   2497 # c      2) t(1) <= t(2) <= ... <= t(k+1)
   2498 # c         t(n-k) <= t(n-k+1) <= ... <= t(n)
   2499 if (t[:k+1] > t[1:k+2]).any():

ValueError: Need k+1 <= n-k-1 <= m. Got m = 15, n = 21 and k = 3.
```

czgdp1807:
@ev-br Added tests and resolved conflicts. Let me know if there is anything else to do in this PR. See https://github.com/scipy/scipy/pull/23233/commits/e24362835dfd8b0b3f0db2bab7c04138bae902a7

scipy/interpolate/tests/test_bsplines.py
```diff
@@ -3469,95 +3472,256 @@ def __call__(self, p):
...
+
+        assert isinstance(spl, BSpline)
+        assert spl.t.shape[0] == 2 * (k + 1)
+        assert spl.k == k
+        xp_assert_close(spl.t[:k+1], np.asarray([xb] * (k + 1)))
+        xp_assert_close(spl.t[-(k+1):], np.asarray([xe] * (k + 1)))
+
+    @pytest.mark.parametrize("n", [100, 51, 15, 11])
+    @pytest.mark.parametrize("s", [10, 8, 5, 1, 1e-2])
+    def test_make_splrep_matches_scipy_splrep_periodic(self, n, s):
```
ev-br: Thanks.
The logic of this test is very hard to read though. Could you  rewrite the test to something like

```
 t, c, k = splrep(x, y, s=s, per=(self.bc_type == "periodic"))
 spl = make_splrep(x, y, s=s, bc_type=self.bc_type)

if (... known conditions where knots differ):
    # the knot vector differs; reuse the knots from FITPACK and only check coefficients
      spl_1 = make_splrep(x, y, s=s, bc_type=self.bc_type, t=t)
      xp_assert_close(spl_1.c, c[:-k - 1], atol=1e-15)
else:
       xp_assert_close(spl.t, t, atol=1e-15)
       xp_assert_close(spl.c, c[:-k - 1], atol=1e-15)
```

ev-br (changes requested):

scipy/interpolate/tests/test_bsplines.py
czgdp1807: Sure. :-). I will address it by tonight (let's see if more reviews come in by that time).

scipy/interpolate/tests/test_bsplines.py
```diff
@@ -3469,95 +3472,256 @@ def __call__(self, p):
...
         fp = np.sum(self.w**2 * (spl(self.x) - self.y)**2)
 
         self.spl = spl   # store it
 
         return fp - self.s
 
-
 @skip_xp_backends(cpu_only=True)
-class TestMakeSplrep:
+class TestMakeSplrepBase:
```
ev-br:
```suggestion
class _TestMakeSplrepBase:
```

So that this class itself does not get collected by pytest. To check: run a single test from this subclass with `-v`, check that it is run twice not three times.

ev-br:
Oh, and --- sorry for a piecemeal review, this is the last comment of this batch --- I think I found the reason for the 
`ValueError: Need k+1 <= n-k-1 <= m. Got m = 15, n = 21 and k = 3.` exception I saw above. 

The problem is that https://github.com/scipy/scipy/blob/main/scipy/interpolate/_fitpack_repro.py#L680 calls `fpcheck(x, y, t)`. This function is a translation of https://github.com/scipy/scipy/blob/main/scipy/interpolate/fitpack/fpchec.f, while for periodic boundary conditions we need https://github.com/scipy/scipy/blob/main/scipy/interpolate/fitpack/fpchep.f

This code path is only triggered if you explicitly supply knots, `make_splrep(..., t=t, bc_type="periodic")`. So what is needed here is to trace the difference between `fpchec.f` and `fpchep.f` and port them to the python-level `fpcheck` routine. Can be done either here or in an immediate follow-up PR.

czgdp1807:
@ev-br I have addressed your reviews. Please let me know if there are more changes to make.

ev-br (approved):
All my comments have been addressed. AFAIU Jake and Daniel have no blocking concerns either. Since the last commit added new code, I'd give it a week before merging. Please speak up if you want more time to take a look.

Meanwhile @czgdp1807 not all of 19 commits are useful to keep. Could you squash it some please?

czgdp1807:
I will do it by tonight. :-).

czgdp1807:
I am leaving this comment as a reminder that this PR will be merged soon. If anyone has reviews to offer, please go ahead and share.

If reviews come after merging then I will promptly address those in follow up PRs.

Thanks for the review cycle. Amazing experience to work on this PR. Grateful to all. :D.

ev-br:
Time to merge this. Thank you @czgdp1807 , this was a very heavy lift. Keep them coming! And thanks to all reviewers.

czgdp1807:
Wohooooo!!! My largest PR in SciPy. I am legit crying out of happiness. 😭🥺. 

Thank you so much @ev-br @rgommers @dschmitz89 and all other folks for reviewing.

scipy/interpolate/_fitpack_repro.py
```diff
@@ -773,6 +1047,20 @@ def make_splrep(x, y, *, w=None, xb=None, xe=None, k=3, s=0, t=None, nest=None):
         The actual number of knots returned by this routine may be slightly
         larger than `nest`.
         Default is None (no limit, add up to ``m + k + 1`` knots).
+    periodic : bool, optional
+        If True, data points are considered periodic with period ``x[m-1]`` -
+        ``x[0]`` and a smooth periodic spline approximation is returned. Values of
+        ``y[m-1]`` and ``w[m-1]`` are not used.
+        The default is False, corresponding to boundary condition 'not-a-knot'.
```
j-bowhay, line 1054: @czgdp1807 `make_splrep` has no argument `perodic` could you open a PR to fix the docs

czgdp1807: Doing it right now.

PR #23235, opened 2025-06-25
BUG: stats.differential_entropy: check case when window_length is a param and is None

LucaCerina:
#### Reference issue
When the parameter `window_length` is passed to differential_entropy, but is None, `kwargs.get('window_length')` would not trigger the default value, keeps `window_length=None`, and raises a TypeError `2*window_length` --> int*NoneType

Now `kwargs.get('window_length')` would be None both in the case of empty/default parameter, and a `window_length=None` parameter. The check is performed later on the actual value of `window_length`.

#### Additional information
None to add

### DIFF
--- a/scipy/stats/tests/test_entropy.py
+++ b/scipy/stats/tests/test_entropy.py
@@ -234,6 +234,14 @@ class TestDifferentialEntropy:
         with pytest.raises(ValueError, match=message):
             stats.differential_entropy(x, method='ekki-ekki')
 
+    def test_window_length_is_none(self, xp):
+        x = np.random.rand(10)
+        window_length = None
+
+        expected = stats.differential_entropy(x)
+        res = stats.differential_entropy(x, window_length=window_length)
+        xp_assert_close(res, expected, rtol=0.005)
+
     @methods
     def test_consistency(self, method, xp):
         # test that method is a consistent estimator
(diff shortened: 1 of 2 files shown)


scipy/stats/tests/test_entropy.py
```diff
@@ -234,6 +234,14 @@ def test_input_validation(self, xp):
         with pytest.raises(ValueError, match=message):
             stats.differential_entropy(x, method='ekki-ekki')
 
+    def test_window_length_is_none(self, xp):
+        x = np.random.rand(10)
+        window_length = None
+
+        expected = stats.differential_entropy(x)
+        res = stats.differential_entropy(x, window_length=window_length)
+        xp_assert_close(res, expected, rtol=0.005)
```
mdhaber: Looks like we need to update the rest of this file, but just to keep this patch in line with current standards:
```suggestion
        rng = np.random.default_rng(358923459826738562)
        x = xp.asarray(rng.random(size=10))
        ref = stats.differential_entropy(x)
        res = stats.differential_entropy(x, window_length=None)
        xp_assert_close(res, ref, rtol=0.005)
```

The main changes are:
- Using a `Generator` rather than global `RandomState` for reproducibility
- Using `xp.asarray` to ensure test runs with each backend (Torch, JAX, etc.)

mdhaber:
Test fails on main, fix is simple, and test passes with PR. 

`linalg` test failures (`test_expm_multiply`) can't be related. It looks like this is the first place we're seeing them; will file an issue if we see it elsewhere.

Thanks @LucaCerina!

PR #23243, opened 2025-06-26
DEP: special: remove deprecated functions

j-bowhay:
`sph_harm`, `clpmn`, `lpn`, `lpmn` are all due to be removed in this release


scipy/special/__init__.py
```diff
@@ -484,11 +484,7 @@
    :toctree: generated/
 
    lpmv                       -- Associated Legendre function of integer order and real degree.
```
j-bowhay, line 486: I didn't remove these other functions as they don't raise a deprecation warning

lucascolley (approved):
thanks Jake!

lucascolley:
I updated the tracker

scipy/special/__init__.py
j-bowhay: @steppi what is the plan with these other functions? Should they be raising a warning?

steppi: Sorry for missing the ping. The replacements haven't been written yet, so no warnings yet. It's something I've wanted to do but haven't had time for.

PR #23234, opened 2025-06-25
TYP: generic ``interpolate`` types

jorenham:
This is the same story as with ##23118, but this time for these ``scipy.interpolate`` types:

- ``AAA``
- ``FloaterHormannInterpolator``
- ``BSpline``
- ``CubicHermiteSpline``
- ``CubicSpline``
- ``LinearNDInterpolator``
- ``PPoly``
- ``BPoly``
- ``NdPPoly``
- ``NdBSpline``
- ``NearestNDInterpolator``
- ``BarycentricInterpolator``
- ``KroghInterpolator``
- ``RBFInterpolator``
- ``RegularGridInterpolator``

---

Closes scipy/scipy-stubs#653

PR #23244, opened 2025-06-26
TYP: generic ``optimize`` types

jorenham:
This is the same story as with #23234 and #23118, but this time for these ``scipy.optimize`` types, as  documented in the `scipy-stubs` readme: https://github.com/scipy/scipy-stubs#scipyoptimize


lucascolley (approved):
thanks Joren

PR #22748, opened 2025-03-27
ENH:MAINT: sparse.linalg: rewrite ARPACK in C

ilayn:
#### Reference issue
Towards #18566

#### What does this implement/fix?
ARPACK in C as a static (or shared lib with BLAS/LAPACK dependency, it's complicated...) library.

#### Additional information

 - I included a hand-written version of the notorious `(cz)dotc` BLAS functions so hopefully we will never hear from them.
 - Now ARPACK is thread-safe and reentrant since all SAVE'd variables are taken out to the reverse communication layer with a struct. Hence closed but still needs testing against #5846 EDIT: **Checked and verified**
 - Some of the functions are very confusing and perform Dijkstra-raging goto tricks such as `?NAPPS.f` so I partially rewrote them instead of translating, because each time another week was being spent in debugging if something was not right.
 - Now the pseudo-random generation happens outside ARPACK hence `rng` keyword is added. To be able to use it, new job codes `ido = 4` and `ido = 5` are introduced and `ido = -1` is removed.
 - Since now ARPACK is thread-safe/reentrant, those tests that are testing for "no crash" behavior are removed/let passed.
 - LAPACK `dlamch('E')` by default resolves to `0.5*eps`, in this PR `eps` is used for tolerance tests.

#### All open issues regarding ARPACK

 - [ ] #3716 
 - [ ] #6827 
 - [ ] #9284 
 - [ ] #11232 
 - [ ] #11261  ILP64
 - [ ] #13033 Add the picture
 - [ ] #17632 LinearOperator support
 - [ ] #19235 Performance
 - [ ] #20808 More hanging tests

Closes #9023 
Closes #9185
Closes #9460
Closes #10744
Closes #11805 
Closes #13222 
Closes #13692 
Closes #13780 
Closes #14823 
Closes #18191 
Closes #18736
Closes #19756 
Closes #20339
Closes #21137 
Closes #22310

PR #23133, opened 2025-06-09
CI: validate build on Win-ARM64 

MugundanMCW:
#### Reference issue

Closes: https://github.com/scipy/scipy/issues/21562 https://github.com/scipy/scipy/issues/22825

#### What does this implement/fix?

Enables new CI environment for building & validating SciPy for Win-ARM64.

#### Additional information

- The PR introduces setting up a new CI environment for building and testing SciPy locally for Win-ARM64.
- This CI setup depends on the latest stable LLVM toolchain & vcpkg(for native pkg-config).

Issue #22757, opened 2025-03-28
MAINT: Collect low-level code under SRC folder per module respectively


ilayn:
Over #22524, @rgommers and I exchanged some comments about where the low level code should reside. And it seems like collecting them under a common SRC folder is better. This issue is to have a reminder after we are done with `optimize` translations and the module is F77-free.

rgommers:
Lower-case `src/` please - moving us out of the all-caps age:)

ilayn:
You don't get to complain about all-caps if using 19th century languages like C. Until we use `zig build`, we are graybeards forever. 

Joking aside yes it will be `src/` anyways.

Kai-Striega:
Is this realistic for 1.16.0? Otherwise I'll have a bit more free time over the next few weeks. I'd be happy to take on a couple of modules if you want to allocate them to me.

lucascolley:
Yes, I don't see any blockers here. We should just check whether any other PRs close to merge would get conflicts

ilayn:
It's about moving the low level code files into a subfolder and modifying the meson.build file. But there is no rush. I'll remove the milestone.

rgommers:
Agreed, no hurry - it's actually better to do that when working on code. Standalone PRs moving code around are low-value and may yield merge conflicts on other PRs.

PR #23246, opened 2025-06-27
RFC: avoid X-macros in arpack wrappers

h-vetinari:
Follow-up to #22748, since the macro games with `X(...)` being constantly redefined provoked antibodies on my end.

The main idea is to move from
https://github.com/scipy/scipy/blob/56e3c8d6153cde6e40cfcc6ca95cce318d800ebb/scipy/sparse/linalg/_eigen/arpack/_arpackmodule.c#L22-L29
to something that encodes the same information (we cannot do reflection on the arpack struct itself), i.e. field names and types, and then use that to do explicit loops further down, rather than said macro games.

This is complicated by the fact that `ARPACK_arnoldi_update_vars_s` and `ARPACK_arnoldi_update_vars_d` come in two different variants for single & double precision, c.f.

https://github.com/scipy/scipy/blob/56e3c8d6153cde6e40cfcc6ca95cce318d800ebb/scipy/sparse/linalg/_eigen/arpack/ARPACK/_arpack.h#L160-L167
and
https://github.com/scipy/scipy/blob/56e3c8d6153cde6e40cfcc6ca95cce318d800ebb/scipy/sparse/linalg/_eigen/arpack/ARPACK/_arpack.h#L205-L212

While that difference is very likely benign on 64bit platforms, it problematic in the sense that the compiler might produce different offsets of the struct fields.

On top of that, these structs appear in the function signatures and thus the API/ABI, and \*of course\* the only fields of potentially differing size are at the _beginning_ of the struct, which completely messes up any shared offset-based computations between the two structs. 😑 

So despite wanting to avoid macros where possible, I've used one to avoid duplication of the list of fields, because we do need to duplicate the array which contains the name/offset information between single & double precision. The rest is straight-forward translation from the macros to C code. The only changes I've made was:
* remove explicit initialization - I cannot see where that's used, since we require all fields to be populated by the dict anyway
* use CPython's `PyErr_Format` instead of handspun `PYERR`

There'd be further opportunity to condense much of the wrapper code (which repeats for each initialization/readout), but for now I've tried to keep the changes small.

### DIFF
--- a/scipy/sparse/linalg/_eigen/arpack/_arpackmodule.c
+++ b/scipy/sparse/linalg/_eigen/arpack/_arpackmodule.c
@@ -17,19 +17,7 @@
     #define ARPACK_cplxf(real, imag) ((real) + (imag)*I)
 #endif
 
-#define PYERR(errobj,message) {PyErr_SetString(errobj,message); return NULL;}
 static PyObject* arpack_error;
-
-// The following macros are used to define the field names in the ARPACK struct.
-#define STRUCT_INEXACT_FIELD_NAMES X(tol) X(getv0_rnorm0) X(aitr_betaj) X(aitr_rnorm1) X(aitr_wnorm) X(aup2_rnorm)
-#define STRUCT_INT_FIELD_NAMES X(ido) X(which) X(bmat) X(info) X(iter) X(maxiter) X(mode) \
-                               X(n) X(nconv) X(ncv) X(nev) X(np) \
-                               X(shift) X(getv0_first) X(getv0_iter) X(getv0_itry) X(getv0_orth) \
-                               X(aitr_iter) X(aitr_j) X(aitr_orth1) X(aitr_orth2) X(aitr_restart) \
-                               X(aitr_step3) X(aitr_step4) X(aitr_ierr) X(aup2_initv) X(aup2_iter) \
-                               X(aup2_getv0) X(aup2_cnorm) X(aup2_kplusp) X(aup2_nev0) X(aup2_np0) \
-                               X(aup2_numcnv) X(aup2_update) X(aup2_ushift)
-#define STRUCT_FIELD_NAMES STRUCT_INT_FIELD_NAMES STRUCT_INEXACT_FIELD_NAMES
 
 typedef enum {
     FIELD_INT,
@@ -328,49 +316,52 @@ cnaupd_wrap(PyObject* Py_UNUSED(dummy), PyObject* args)
     // Declare and Initialize the ARPACK struct that will be populated from dict with zeros
     struct ARPACK_arnoldi_update_vars_s Vars = {0};
 
-    #define X(name) Vars.name = 0;
-    STRUCT_FIELD_NAMES
-    #undef X
+    for (size_t i = 0; i < ARPACK_FIELD_COUNT; ++i) {
+        const char* name = arpack_fields_s[i].name;
+        PyObject* obj = PyDict_GetItemString(input_dict, name);
+        if (!obj) {
+            PyErr_Format(arpack_error, "%s not found in the dictionary.", name);
+        }
+        void* ptr = (char*)&Vars + arpack_fields_s[i].offset;
 
-    #define X(name) \
-        PyObject* name##_obj = PyDict_GetItemString(input_dict, #name); \
-        if (!name##_obj) { PYERR(arpack_error, #name " not found in the dictionary."); } \
-        Vars.name = (float)PyFloat_AsDouble(name##_obj);
-        STRUCT_INEXACT_FIELD_NAMES
-    #undef X
-
-    #define X(name) \
-        PyObject* name##_obj = PyDict_GetItemString(input_dict, #name); \
-        if (!name##_obj) { PYERR(arpack_error, #name " not found in the dictionary."); } \
-        Vars.name = (int)PyLong_AsLong(name##_obj);
-        STRUCT_INT_FIELD_NAMES
-    #undef X
+        switch (arpack_fields_s[i].type) {
+            case FIELD_INT:
+                *(int*)ptr = (int)PyLong_AsLong(obj);
+                break;
+            case FIELD_FLOATING:
+                *(float*)ptr = (float)PyFloat_AsDouble(obj);
+                break;
+            default:
+                PyErr_Format(arpack_error, "Unknown field type for %s", name);
+        }
+    }
 
     // Call ARPACK function
     ARPACK_cnaupd(&Vars, resid, v, Vars.n, ipntr, workd, workl, rwork);
 
     // Unpack the struct back to the dictionary
-    #define X(name) do { \
-            PyObject* tmp_##name = PyFloat_FromDouble((double)Vars.name); \
-            if ((!tmp_##name) || (PyDict_SetItemString(input_dict, #name, tmp_##name) < 0)) { \
-            Py_XDECREF(tmp_##name); \
-            PYERR(arpack_error, "Setting '" #name "' failed."); \
-            } \
-            Py_DECREF(tmp_##name); \
-        } while (0);
-        STRUCT_INEXACT_FIELD_NAMES
-    #undef X
+    for (size_t i = 0; i < ARPACK_FIELD_COUNT; ++i) {
+        const char* name = arpack_fields_s[i].name;
+        void* ptr = (char*)&Vars + arpack_fields_s[i].offset;
 
-    #define X(name) do { \
-            PyObject* tmp_##name = PyLong_FromLong((long)Vars.name); \
-            if ((!tmp_##name) || (PyDict_SetItemString(input_dict, #name, tmp_##name) < 0)) { \
-                Py_XDECREF(tmp_##name); \
-                PYERR(arpack_error, "Setting '" #name "' failed."); \
-            } \
-            Py_DECREF(tmp_##name); \
-        } while (0);
-        STRUCT_INT_FIELD_NAMES
-    #undef X
+        PyObject* obj = NULL;
+        switch (arpack_fields_s[i].type) {
+            case FIELD_INT:
+                obj = PyLong_FromLong((long)(*(int*)ptr));
+                break;
+            case FIELD_FLOATING:
+                obj = PyFloat_FromDouble((double)(*(float*)ptr));
+                break;
+            default:
+                PyErr_Format(arpack_error, "Unknown field type for %s", name);
+        }
+
+        if (!obj || PyDict_SetItemString(input_dict, name, obj) < 0) {
+            Py_XDECREF(obj);
+            PyErr_Format(arpack_error, "Setting '%s' failed.", name);
+        }
+        Py_DECREF(obj);
+    }
 
     Py_RETURN_NONE;
 }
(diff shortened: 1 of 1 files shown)


h-vetinari:
> On top of that, these structs appear in the function signatures and thus the API/ABI, [...]

Perhaps I misunderstood something there. I just saw your [comment](https://discuss.scientific-python.org/t/alternative-name-for-the-new-c-translation-of-arpack/1995/4) on discourse that the API/ABI break is intentional. In that case, I have to ask why we should even bother to have the single/double distinction.

I'm OK with keeping the code runnable on 32-bit systems, but I'm against extra maintenance burden because of that. If we decided to back everything in "our" API by the double-flavoured struct, we could greatly simplify the implementation (e.g. what I started out with when I started translating, see https://github.com/scipy/scipy/pull/23246/commits/c7f0cbcd2f5679bc977da62a1da4a12281dd486d).

ilayn:
Not sure I understand your objection here so let me spell out my confusion. 

The macro "game" is pretty well-defined and kind of standard practice in C which is called X-macros. https://en.wikipedia.org/wiki/X_macro. The use case is to read the Python dict values to a struct or write the struct values to the Python dict with the updated values. And instead of writing the same thing 20 something times we use X macros. 

The struct with double and single precision is not about 32bit/64bit systems. It is the np.float32 and np.float64 distinction that is coming from the user side. The reason why they are at the top of the struct is to pack the struct a bit more efficiently such that otherwise if the number of ints are odd numbered we don't get padding and alignment issues because float and double are typically larger than or equal to ints in C. Hence bigger ones are placed early to avoid strange packing/padding causing alignment issues.

h-vetinari:
> The macro "game" is pretty well-defined and kind of standard practice in C which is called X-macros. https://en.wikipedia.org/wiki/X_macro.

Thanks, I hate it. (this is tongue-in-cheek and not directed at you personally, obviously)

> The use case is to read the Python dict values to a struct or write the struct values to the Python dict with the updated values. And instead of writing the same thing 20 something times we use X macros.

I got that, but I don't see why we can't do that with something else than macros. We actually have capable compilers now, we can do better than 20th century code style IMO.

> The struct with double and single precision is not about 32bit/64bit systems. It is the np.float32 and np.float64 distinction that is coming from the user side.

I still fail to see why we couldn't use a double field in the struct even for the `np.float32` values. We have to cast anyway in the wrappers (as we do between int/long). I care very little about the packing density on 32bit machines, I consider that really not a worthwhile tradeoff for the maintenance cost of duplicating the struct (with all that entails).

ilayn:
For the reason why there are two versions of the struct is because ARPACK is written very much like LAPACK for 4 different flavors, that is s, d, c, z. And s, c uses single precision `float`s and d, z uses double precision `double`s. 

Because C does not have struct composition in a convenient way, that is to say, if you have a struct already for ints you can't write 

```C
typedef struct
{
inexact_struct
int_struct
} new_struct;
```

ilayn:
> I got that, but I don't see why we can't do that with something else than macros. We actually have capable compilers now, we can do better than 20th century code style IMO.

I don't follow this part. It's basically a text-based for loop at the pre-processing stage with explicit access to the struct elements without bothering with bit offsets manually which is more fragile due to compiler/architecture differences.

By the way, though my comments suggest otherwise, I truly hate writing in Fortran/C/C++. They are all equally unbearable in my book. So please don't take it the wrong way that somehow I am defending any of this in terms of ergoonomics, so at a meta level I completely agree with you :smile:

h-vetinari:
> For the reason why there are two versions of the struct is because ARPACK is written very much like LAPACK for 4 different flavors, that is s, d, c, z. And s, c uses single precision `float`s and d, z uses double precision `double`s.

I understand that. The types for individual values as handled by ARPACK wouldn't change (float for s/c, double for d/z). But if we're free to break the API/ABI, then the functions here
https://github.com/scipy/scipy/blob/56e3c8d6153cde6e40cfcc6ca95cce318d800ebb/scipy/sparse/linalg/_eigen/arpack/ARPACK/_arpack.h#L250-L253

could still share the same struct type (the one currently called `ARPACK_arnoldi_update_vars_d`) for the first argument.

What I mean is, if we're not constrained by API/ABI, then we get to design the signature; I'd like to avoid that struct distinction in the signature.

> Because C does not have struct composition in a convenient way, [...]

Can you explain in a bit more detail what you mean here? I understand that the operation you describe is not possible, but I don't see how that pertains to the situation right now.

> I don't follow this part. It's basically a text-based for loop at the pre-processing stage with explicit access to the struct elements without bothering with bit offsets manually which is more fragile due to compiler/architecture differences.

Using the preprocessor should be a last resort. In this case it produces code that's basically impossible to analyse correctly (everything from static analysis to syntax highlighting to compiler/debugger diagnostics), because you need to run the preprocessor to see what it ends up as. I could imagine keeping them for the field listing, but using them for the translation layer logic tips the balance from helpful to detrimental[^1]

[^1]: also, with static field info, we could actually turn the prolog/epilogue before the actual ARPACK calls into an actual function call, rather than repeating the same macro soup every time.

W.r.t. the offsets, they're entirely controlled by the compiler/architecture we're compiling for, which means they're just part of the ABI we have to commit to anyway when compiling. Plus, it's an part of the C standard that's been there since the beginning, so I don't see what should be fragile about it.

More to the point, even the macro-variant ultimately gets compiled to memory offsets (i.e. `Vars.field_name` gets replaced by something involving `offsetof(..., field_name)`), so it's not like we're changing anything at the ABI level either. In both cases we cannot change the order of fields in the struct without recompiling everything.

ilayn:
> What I mean is, if we're not constrained by API/ABI, then we get to design the signature; I'd like to avoid that struct distinction in the signature.

Yes I wrote three different versions until I settle down with this. Because then you have to downcast the doubles to floats everywhere these variables are used in the implementation which involve norms of vectors or tolerances etc. So you will be doing a lot of casting that you can't control in a debugging situation and create a lot of head-scratching in case something does not go through. 

It happened already because I was trying to use a `double` tol value in a `float` accepting flavor. After three days I traced it back to the wrapper only to discover that I was reading a `double` as a `float` which was giving zeros and then going awry. 

By the way, we will also have the ILP64 switch for 64bit LAPACK ints. So it will get even worse pretty soon. 

Also if this is used purely in C context which `igraph` project is hoping to use and potentially a few more, then you need a separated API for single precision and double precision use. That is to say, one might opt out for only using single precision version. 

The immediate reaction to this is "use C++ boomer!" but that comes with a different set of problems so I don't see a  satisfactory answer to this while using C/C++.



> Using the preprocessor should be a last resort. 

Unfortunately you have to make peace with this if you are in C/C++. Even `#include` is a text replacement op. They sound sophisticated but `pragma`s or macros, these primitive languages only offer text replacement as comp-time tooling. I hope you never look at MINPACK header file code that is coming from Travis (which should not be in a header file yes, I'll circle back to all of them for #22757 but be my guest if you feel like it).

h-vetinari:
> > Using the preprocessor should be a last resort.
> 
> Unfortunately you have to make peace with this if you are in C/C++.

There's a vast gulf between using `#include` and recursive macros. This PR is currently using a macro as well, in order to populate the offset arrays without duplicating the list of fields. As always, there are trade-offs. Handling the conversion logic in the X-macros is to me a distinctly bad trade-off.

> The immediate reaction to this is "use C++ boomer!" but that comes with a different set of problems so I don't see a satisfactory answer to this while using C/C++.

I wasn't proposing to switch to C++, just to limit the use of the preprocessor to where it's unavoidable or at least not harmful.

> Yes I wrote three different versions until I settle down with this. Because then you have to downcast the doubles to floats everywhere these variables are used in the implementation which involve norms of vectors or tolerances etc.

Fine by me to keep both then. I was responding to the "we can break the API/ABI" part. If the implementation needs different structs for single/double to be maintainable, I'm happy to drop that line of thought. Then the state of this PR is actually reviewable. :)

> By the way, we will also have the ILP64 switch for 64bit LAPACK ints. So it will get even worse pretty soon.

That's a one-time choice at compilation-time, I don't see how it would have a lot of impact here... 🤷

ilayn:
> There's a vast gulf between using #include and recursive macros. 

Yes but only to you that is (no recursion here by the way). A complicated regex and a simple find/replace are the same thing. It does not matter what the operation is. They are modifying text the same way with same shortcomings/dangers. So essentially there is no difference. You might have a fundamental objection to X-macros or macros in general but they **are** valid tools and used everywhere you don't want them to be :smile:

Because your version looks objectively cleaner, though raise the hair on the back of my neck in return let's switch to a flow like that. Note that I can make the same "why are we dealing with memory offsets in 2025" hence we did not achieve any modernization here. Is there a version that we actually set attributes and not compute the offsets as a compromise? Because unlike macros that is compiler dependent operation and sets up an unnecessary fragile ABI boundary (including the paddings and so on).

And you **must** zero initialize a struct by the way, but that's a C nonsense, I agree I wish we did not have to do that.

h-vetinari:
> > There's a vast gulf between using #include and recursive macros.
> 
> Yes but only to you that is (no recursion here by the way).

Macros have well-documented downsides; it's hard to take the "only to you" seriously from anyone who's been around programming as long as you have - fire up your favourite search engine or LLM if it's really in doubt.

With "recursion" I meant that one macro (e.g. `STRUCT_INT_FIELD_NAMES`) calls another macro (`X`) that's by necessity variable in its definition. You cannot do it with a single layer of (function) macros, and that macro-recursion (if you permit me the abuse of the term) is really bad™️.

> Note that I can make the same "why are we dealing with memory offsets in 2025" hence we did not achieve any modernization here.

We've gone from the preprocessor to native C, which I consider an improvement already. It would be nice to do this with reflection (just accepted into C++26!), but that technology doesn't exist yet for C, and perhaps never will. So for now we're down to hard-coding the field names.

The reason why I went with also encoding the type of the field (which in turn requires the `switch` in the loop body) is that it seemed cleaner and more extensible than having `int_fields` and `float_fields`; plus, it allows unifying much of the loops that would otherwise have to be duplicated.

> Is there a version that we actually set attributes and not compute the offsets as a compromise? Because unlike macros that is compiler dependent operation and sets up an unnecessary fragile ABI boundary (including the paddings and so on).

There might be a misunderstanding (I think on your side, but perhaps I don't understand what you mean). As I mentioned further up ("more to the point, [...]"), even the currently-checked-in-on-`main` version ends up encoding offsets into the end result. Those struct offsets are part of the ABI already!

That said, if someone has a way to do this without using either the preprocessor or `offsetof`, I'm all ears 🙃

h-vetinari:
> @h-vetinari: There'd be further opportunity to condense much of the wrapper code (which repeats for each initialization/readout), but for now I've tried to keep the changes small.

I've decided to give this a shot, and it yields a nice reduction in LoC. :)

ilayn:
> Macros have well-documented downsides; it's hard to take the "only to you" seriously from anyone who's been around programming as long as you have - fire up your favourite search engine or LLM if it's really in doubt.

> With "recursion" I meant that one macro (e.g. STRUCT_INT_FIELD_NAMES) calls another macro (X) that's by necessity variable in its definition. You cannot do it with a single layer of (function) macros, and that macro-recursion (if you permit me the abuse of the term) is really bad™️.

Like I said X-macro is defined like that but it is not a recursion. And macros are valid tools with downsides not valid in this use case. The downsides are starting to creep in if you have a look at `__minpack.h`. So I guess we will have to disagree at this, given your strong but fresh reaction to X-macros.

> We've gone from the preprocessor to native C, which I consider an improvement already. It would be nice to do this with reflection (just accepted into C++26!), but that technology doesn't exist yet for C, and perhaps never will. So for now we're down to hard-coding the field names.

Again native C / preprocessor distinction is only valid in your line of argument. They are pieces of the same machinery. You might not like it but you can't reject their usage because it does not look native to you. 

> There might be a misunderstanding (I think on your side, but perhaps I don't understand what you mean). A [...] That said, if someone has a way to do this without using either the preprocessor or offsetof, I'm all ears 🙃

Yes and that's exactly my point is that I feel the same way about your usage of offsets. So currently it's my way vs your way of argumentation. I am not claiming that I did the cleanest solution but at least I am using the struct fields instead of dealing with their offsets which is a no-no unless it is absolutely necessary (you can LLM for using the offsets yourself). So let's not take it to a place where we start assuming that each side thinks they are right because I can do the same and we achieve nothing. 

I'm also all ears and I think there might be a way. But please don't start an argument for the sake of these obscure details that wraps a library because I'm already very tired of working on this for months and I might not be able to approach to it as calmly as I can in normal times dealing with it for so long.

h-vetinari:
Your work is highly appreciated (as I noted right away in your PR). At the same time, some of the arguments that you're making here ("essentially there is no difference" between `#include` and arbitrary macro complexity) are just... demonstrably wrong.

> Again native C / preprocessor distinction is only valid in your line of argument.

For example this one. You may consider the preprocessor as indivisible from a C-compiler[^1], but they really are distinct. They even have separate binaries in a standard GCC setup. The translation phases they correspond to during compilation are completely different.

[^1]: and far be it for me to argue that C can be compiled _without_ a preprocessor

This has a lot of consequences that I mentioned already as well (static analysis, diagnostic quality, syntax highlighting etc.), all of which gets substantially _worse_ with macros, and are thus generally inferior for tasks that can equally be achieved with regular C[^2].

Compilers and IDEs and analyzers and debuggers can introspect headers. They can't resolve arbitrarily nested macros.

[^2]:  again, no argument that some things can't be achieved in plain C, or sometimes avoiding repetition for maintainability is justification enough

You may consider "it's all just C anyway", but that's because the differences are apparently not important to you. They are to other people though. 

>  I am not claiming that I did the cleanest solution but at least I am using the struct fields instead of dealing with their offsets which is a no-no unless it is absolutely necessary (you can LLM for using the offsets yourself).

I've already addressed the point about offsets twice. The compiler decides how to lay out a struct in memory - it is literally the best-placed actor to determine the offset of a field. Any fragility there comes from the fact that it's part of the ABI (so you can't mix artefacts if the struct changes), but that's exactly the same as the status quo - even your field-based code ends up using memory offsets in the final binary. Can you explain what we would gain by denying ourselves the use of a mechanism that is in fact at play in any case?

h-vetinari:
The existence of crusty code from the bad old days is not a reason to keep using those patterns indefinitely. I would have guessed that you'd be the first one to agree to this, given your valiant efforts to rid us of F77 code (imagine your reaction if someone tried to add some new F77 spaghetti code to scipy... 😆).

Macro soup is an almost equally outdated coding style, and we should avoid it where we can. Just as we do for Fortran, we should move to high-quality, modern C code over time.

If the fields of that struct were used in a lot of different places, I could see the point of X-macros. Here though, we just need a static list of the fields and a loop. IMO that's reduced complexity (and LoC count) as well as improved clarity & code quality.

ilayn:
I don't have enough patience to write a wall of text but here are some comments and I won't add much more to this.

You started with "macro games" because that's how I code things, then I explained and you said thanks I hate them. Then the title changed to X macros. Now I'm being lectured on how C works. From this, I must conclude that this is a you problem. With all respect, I won't get a X-macro lecture from someone who just figured out their existence. This is not much different than what tempita does and they all get parsed first **THEN** compiled in the bowels. You can google it yourself like you are advising me quite liberally. 

The macros that I included here are 100% LSP compatible, because that's exactly how I was able to code them in VS Code. Otherwise I would not have the patience. Please spare me some knowledge capacity that I do care as much as you do. Because I have been debugging these code for months now. So I know a bit better than you what works and what does not. "It's not all C code anyways", it is the much verbose but safer option. Open in VS Code or VS or whichever IDE you want. They will inspect fully the macros. If you want to combine them into a single place that's great! Then do that without this gratis C course I'm receiving and without offsets. I never claimed expertise but this much I know quite well. 

I'll expand it one last time about this. Offsets are the most fragile parts of struct handling. Go find some other experts to explain to you if you wish to know more about it since clearly I'm writing outdated code and randomly introducing dusty crust. You might be surprised by the responses you get. But offsets are definitely no go. So with all respect I won't budge from that opinion and I can't accept their introduction just because you don't find macros aesthetically pleasing. Otherwise you are free to form your opinions about macro usage.

With all due respect, please keep your C knowledge to yourself and demonstrate a technical argument. If there are no apparent ways to break the code then this is a pointless argument.

h-vetinari:
> With all respect, I won't get a X-macro lecture from someone who just figured out their existence.

Are we going to use Duff's device too, just because it has a canonical name? 

Whatever man, if you cannot see the objective improvements here, then I'm not going to bother.

ilayn:
Others can take this in I don't wish to participate what happens to the code next. Next time I'll try to google harder to write more native C code.

andyfaff:
Hey, the temperature of the conversation seems to have risen too far here. We're all passionate about improving the scipy code, but let's remember that we're on the same side at the end of the day.

lorentzenchr:
How about this? We agree:
- @ilayn did a great job translating Fortran to valid C.
- @h-vetinari did a great job with this PR increasing maintainability.
- We stop arguing about different general programming practices and instead focus on the code at hand.

I think this PR is a net improvement.

lucascolley:
I would be happy to merge if you are willing to review @lorentzenchr  and @h-vetinari is willing to re-open the PR. Ilhan has expressed a preference for not participating further in https://github.com/scipy/scipy/pull/23246#issuecomment-3012690312.

ilayn:
I took some of the principles from this PR and applied it already see the new version here. https://github.com/scipy/scipy/blob/main/scipy/sparse/linalg/_eigen/arpack/_arpackmodule.c

Macros are not frowned upon in C. That's essentially a C++ point of view. I don't agree about the maintainability by using offsets at all.

h-vetinari:
Please take whatever changes you want from this PR. I consider it a blindingly obvious improvement, but I don't have the energy to fight Ilhan for it.

> Macros are not frowned upon in C. 

We'll have to disagree on that.

ilayn:
For everyone who wants to chime in; please stop pinging contested issues like this. Not only you are re-annoying the participants by barging in but you are not mediating as you might think you are. Moreover I'm receiving statements, I already wrote 5 comments ago back to me. "Blindly obvious" or other insinuating silly comments are not going to change my view on things. I can give 15000 lines of reasons because I wrote them. This wrapper is the last tiny part that has nothing to do with the actual maintenance but just the glue. The code runs either way and I already fixed many things said here. So not sure what the point of this ping after 2 weeks. I did not have the patience or the energy already. I can link it above where I explicitly mentioned this if it is not loading up on the screens. 

If anybody wants to really maintain; here is one #23133 go maintain. Otherwise stop wasting time on bikeshedding. With this attitude and absolute confidence I don't give a fuck about if it is macros or whatever. Funny how I started at the top and now what I am typing.

h-vetinari:
> "Blindly obvious" or other insinuating silly comments

You conveniently ignored the "I consider it" - I do have a right to my own opinion. In any case, your reaction is a good illustration why I'm disengaging completely. Peace ✌️

lorentzenchr:
Sorry for the noise then.
fyi (better not reply, really just fyi): I would have opened a new PR based on this one, but the main branch changed too much.

PR #23247, opened 2025-06-27
ENH/DOC: signal.hilbert2: Refactor to avoid extra scaling array and improve docstr

JesseLivezey:
Similar to #23173 for `hilbert2`. This PR
- avoids allocating an extra scaling array (I get modest speedups on a laptop),
- improves the docstring ~and adds a reference to the particular definition of 2d analytic signal used here
- add an argument to allow performing batched `2d` transform on `nd` arrays for `n >= 2`

TODOs:
- [x] add a test for using non-default axes
- [x] add a direct test that `ifft2` of analytic signal has expected orthant power.
- [x] what's the right version to add for the added arg? 1.16.2 was added.
- [ ] check that docs look okay

FYI, @DietBru , similar to the `hilbert()` PR.


scipy/signal/tests/test_signaltools.py
```diff
@@ -3497,21 +3499,85 @@ def test_bad_args(self, xp):
...
+            while k > 2:
+                h = h[:, xp.newaxis]
+                k -= 1
+            x = sp_fft.ifft2(Xf * h, axes=(0, 1))
+            return x
+
+        x = xp.arange(xp.prod(shape)).reshape(shape)
+        xh_old = _hilbert2(x)
+        xh = hilbert2(x)
+        assert_allclose(xh_old, xh)
```
j-bowhay:
```suggestion
        xp_assert_close(xh_old, xh)
```

scipy/signal/tests/test_signaltools.py
```diff
@@ -3497,21 +3499,85 @@ def test_bad_args(self, xp):
...
+
+            h = h1[:, xp.newaxis] * h2[xp.newaxis, :]
+            k = x.ndim
+            while k > 2:
+                h = h[:, xp.newaxis]
+                k -= 1
+            x = sp_fft.ifft2(Xf * h, axes=(0, 1))
+            return x
+
+        x = xp.arange(xp.prod(shape)).reshape(shape)
```
JesseLivezey:
```suggestion
        x = xp.arange(shape[0] * shape[1]).reshape(shape)
```

ev-br: reshape is a function not a method in array API, thus you want `xp.reshape(...)`

DietBru:
I found a little time to look at it today:
The referenced [Wikipedia article] does not give a definition but only states that there are different approaches for defining two-dimensional analytic signals.. It would be helpful, if the docstr would state how the 2d Hilbert transform / analytic signal is defined for this function. An authoritative reference would even be better.

The following example illustrates what is happening:
```python
import numpy as np
from scipy.fft import fft2, fftshift, ifftshift
from scipy.signal import hilbert2

# Input signal is unit impulse with DC offset:
x = np.ones((5, 5)) / 25
x[0, 0] += 1
X = fftshift(fft2(x))  # Zero frequency bin is at center

y = hilbert2(x)
Y = fftshift(fft2(y))

print("FFT2 of Input:")
print(X)
print("FFT2 of Output:")
print(np.round(Y, 3))
```
It produces:
```text
FFT2 of Input:
[[1.-0.j 1.-0.j 1.-0.j 1.+0.j 1.+0.j]
 [1.-0.j 1.-0.j 1.-0.j 1.+0.j 1.+0.j]
 [1.-0.j 1.-0.j 2.-0.j 1.+0.j 1.+0.j]
 [1.-0.j 1.-0.j 1.+0.j 1.+0.j 1.+0.j]
 [1.-0.j 1.-0.j 1.+0.j 1.+0.j 1.+0.j]]
FFT2 of Output:
[[ 0.+0.j  0.+0.j -0.+0.j  0.+0.j  0.+0.j]
 [ 0.+0.j  0.+0.j -0.+0.j  0.+0.j  0.+0.j]
 [ 0.+0.j  0.+0.j  2.+0.j  2.+0.j  2.+0.j]
 [ 0.+0.j  0.+0.j  2.+0.j  4.+0.j  4.+0.j]
 [ 0.+0.j  0.+0.j  2.+0.j  4.+0.j  4.+0.j]]
```
It illustrates that the nonnegative frequency bins are multiplied by either two or four. 
I am not completely sure that this result concurs to the [Wikipedia article] and I was not able to find a reference (though I did not search too long).

It would be great if in this PR or in a follow-up, the docstr could be improved.

 [Wikipedia article]: https://en.wikipedia.org/wiki/Analytic_signal#Extensions_of_the_analytic_signal_to_signals_of_multiple_variables

DietBru:
To illustrate  a problem of this definition, the following example produces a zero-value analytic signal from a non-zero input:
```python
import numpy as np
from scipy.fft import fft2, fftshift, ifft2, ifftshift
from scipy.signal import hilbert2

# Create a real signal by ensuring `Z[-k,-l] == np.conj(Z[k,l])` holds:
Z = np.array([[0, 0, 0, 0, 0],
              [0, 0, 0, 1, 0],
              [0, 0, 0, 0, 0],
              [0, 1, 0, 0, 0],
              [0, 0, 0, 0, 0]])
z = ifft2(ifftshift(Z))
assert np.allclose(z.imag, 0)  # z is a real signal
x = z.real

y = hilbert2(x)
print(np.allclose(y, 0))  # => True
```

JesseLivezey:
Didn't mean to close and re-open.

@DietBru, agreed although I haven't found a reference for the particular definition used in scipy. I'm aware of

1. [Multidimensional complex signals with single-orthant spectra](https://ieeexplore.ieee.org/abstract/document/158601/)
2. [A Novel Approach to the 2D Analytic Signal?](https://www.informatik.uni-kiel.de/inf/Sommer/doc/Publications/tbl/caip99.pdf)

The 1d-AS is computed like
```
AS(f(x)) = InvFourier( 2 * H(w_X) * Fourier(f(x)) )
```
and 1. generalizes this like
```
AS(f(x, y)) = InvFourier( 2 * H(w_x) * H(w_y) * Fourier(f(x, y)) )
```
which scales the Fourier orthants like
```
0|4
---
0|0
```

The current scipy implementation (which this PR preserves) does this instead
```
AS(f(x, y)) = InvFourier( 2 * H(w_x) **+** H(w_y) * Fourier(f(x, y)) )
```
which scales the Fourier orthants like
```
2|4
---
0|2
```
I haven't been able to find a reference that uses this definition.

As 2. points out, the definition of 1. means that f(x, y) is not reconstructable from AS(f(x, y)), which is a desirable property for a 2d extension of the AS to have. The definition in 2. relies on a qaternion fourier transform and retains more of the desirable properties of the 1d AS.

JesseLivezey:
I can add better docs for the current version and a method kwarg with an option to use the single orthant definition

DietBru:
I am not sure if I am understanding  your definitions correctly, @JesseLivezey: Currently the values are multiplied by four, if $\omega_x > 0 \wedge \omega_y > 0$ and multiplied by two if $\omega_{x,y}=0 \wedge \omega_{y,x}>0$.

If $H(x) := 1 + \text{sign}(x)$, then `hilbert2` would follow the convention
```
AS(f(x, y)) = InvFourier(  H(w_x) * H(w_y) * Fourier(f(x, y)) )
```


> I can add better docs for the current version and a method kwarg with an option to use the single orthant definition

:+1: Improved docs would be very welcome.

JesseLivezey:
@DietBru correcting my previous comment. The current scipy implementation _is_ based on the "Single orthant" paper to the best of my understanding (added a reference). Roughly, it zeros out any amplitude where either of the frequencies is negative and multiplies the amplitudes of the positive-positive frequency amplitudes by 4. I've tried to clarify these details in the docstring.

As you point out, this definition has some drawbacks which are pointed out in the new third reference.

scipy/signal/_signaltools.py
```diff
@@ -2587,37 +2587,64 @@ def hilbert(x, N=None, axis=-1):
     return x
 
 
-def hilbert2(x, N=None):
-    """
-    Compute the '2-D' analytic signal of `x`
+def hilbert2(x, N=None, axes=(-2, -1)):
+    """Compute the '2-D' analytic signal of `x`.
```
DietBru: For the docstr, it is helpful to precisely describe what is implemented and have an illustrative example. What do you think about  something along the line of:
```python
r"""Compute the '2-D' analytic signal of `x`.

    The 2-D analytic signal is calculated as a so-called "single-orthant" transform.
    This is achieved by applying one-dimensional Hilbert function (as in
    `~signal.hilbert`) to the first and to the second array axis in Fourier space.

    For NumPy arrays, `scipy.fft.set_workers` can be used to change the number of
    workers used for the FFTs.

    Parameters
    ----------
    x : array_like
        Input signal. Must be at least two-dimensional.
    N : int or tuple of two ints, optional
        Number of output samples. `x` is initially cropped or zero-padded to length
        `N` along `axes`.  Default: ``x.shape[i] for i in axes``
    axes : tuple of two ints, optional
        Axes along which to do the transformation.  Default: (-2, -1).

    Returns
    -------
    xa : ndarray
        Analytic signal of `x` taken along given axes.

    .. versionadded:: 1.16.2

    Notes
    -----
    The "single-orthant" transform, as defined in [2]_, is calculated by performing the
    following steps:

    1. Calculate the two-dimensional FFT of the input, i.e.,

       .. math::
       
            X[p,q] = \sum_{k,l=0}^{N_0,N_1} x[k,l]\,
                                                 e^{-2j\pi k p/N_0}\, e^{-2j\pi l q/N_1}

    2. Zero negative frequency bins and double their positive counterparts, i.e.,

       .. math::

           X_a[p,q] = \big(1 + s_{N_0}(p)\big) \big(1 + s_{N_1}(q)\big) X[p,q]

       with :math:`s_N(.)` being a modified sign function defined as

       .. math::

           s_N(p) := \begin{cases}
                                 -1 & \text{ for } p < 0\ ,\\
                       \phantom{-}0 & \text{ for } p = 0\ ,\\
                                 +1 & \text{ for } 1 \leq p < (N+1) // 2\ ,\\
                       \phantom{-}0 & \text{ elsewhere.}
                     \end{cases}

       The limitation of the ":math:`+1`" case to the range of ``[1:(N+1)//2]``
       accounts for the unpaired Nyquist frequency bin at :math:`N/2` for even
       :math:`N`. Note that :math:`X_a[p] = \big(1 + s_N(p)\big) X[p]` is the
       one-dimensional Hilbert function (as in `~signal.hilbert`) in Fourier
       space.

    3. Produce the analytic signal by performing the inverse FFT, i.e.,

       .. math::

           x_a[k, l] = \frac{1}{N_0 N_1}
                 \sum_{p,q=0}^{N_0,N_1} X_a[p,q]\, e^{2j\pi k p/N_0}\, e^{2j\pi l q/N_1}

    The "single-orthant" transform is not the only possible definition of an analytic
    signal in multiple dimensions (as noted in [1]_). Consult [3]_ for a description of
    properties that this 2-D transform does and does not share with the 1-D transform.
    The second example below shows one of the downsides of this approach.

    References
    ----------
    .. [1] Wikipedia, "Analytic signal",
        https://en.wikipedia.org/wiki/Analytic_signal
    .. [2] Hahn, Stefan L. "Multidimensional complex signals with
        single-orthant spectra." Proceedings of the IEEE 80.8
        (1992): 1287-1300.
        `PDF <https://ieeexplore.ieee.org/iel1/5/4083/00158601.pdf>`__
    .. [3] Bülow, Thomas, and Gerald Sommer. "A novel approach to the 2D analytic
        signal." In International Conference on Computer Analysis of Images and
        Patterns, pp. 25-32. Berlin, Heidelberg: Springer Berlin Heidelberg, 1999.
        `PDF <https://www.informatik.uni-kiel.de/inf/Sommer/doc/Publications/tbl/caip99.pdf>`__

    Examples
    --------
    The following example calculates the two-dimensional analytic signal from a single
    impulse with an added constant offset. The impulse produces an FFT where each bin
    has a value of one and the constant offset component produces only a non-zero
    component at the ``(0,0)`` bin.

    >>> import numpy as np
    >>> from scipy.fft import fft2, fftshift, ifftshift
    >>> from scipy.signal import hilbert2
    ...
    >>> # Input signal is unit impulse with a constant offset:
    >>> x = np.ones((5, 5)) / 5
    >>> x[0, 0] += 1
    ...
    >>> X = fftshift(fft2(x))  # Zero frequency bin is at center
    >>> print(X)
    [[1.-0.j 1.-0.j 1.-0.j 1.+0.j 1.+0.j]
     [1.-0.j 1.-0.j 1.-0.j 1.+0.j 1.+0.j]
     [1.-0.j 1.-0.j 6.-0.j 1.+0.j 1.+0.j]
     [1.-0.j 1.-0.j 1.+0.j 1.+0.j 1.+0.j]
     [1.-0.j 1.-0.j 1.+0.j 1.+0.j 1.+0.j]]
    >>> x_a = hilbert2(x)
    >>> X_a = fftshift(fft2(x_a))
    >>> print(np.round(X_a, 3))
    [[ 0.+0.j  0.+0.j -0.+0.j  0.+0.j  0.+0.j]
     [ 0.+0.j  0.+0.j -0.+0.j  0.+0.j  0.+0.j]
     [ 0.+0.j  0.+0.j  6.+0.j  2.+0.j  2.+0.j]
     [ 0.+0.j  0.+0.j  2.+0.j  4.+0.j  4.+0.j]
     [ 0.+0.j  0.+0.j  2.+0.j  4.+0.j  4.+0.j]]

    The FFT of the result illustrates that the values of the lower right quadrant (or
    orthant) with purely positive frequency bins have been quadrupled. The values at its
    borders, where only one frequency component is zero, are doubled. The zero frequency
    bin ``(0, 0)`` has not been altered. All other quadrants have been set to zero.

    This second example illustrates a problem with the "single-orthant" convention. A
    purely real signal can produce an analytic signal which is completely zero:

    >>> from scipy.fft import fft2, fftshift, ifft2, ifftshift
    >>> from scipy.signal import hilbert2
    ...
    >>> # Create a real signal by ensuring `Z[-p,-q] == np.conj(Z[p,q])` holds:
    >>> Z = np.array([[0, 0, 0, 0, 0],
    ...               [0, 0, 0, 1, 0],
    ...               [0, 0, 0, 0, 0],
    ...               [0, 1, 0, 0, 0],
    ...               [0, 0, 0, 0, 0]]) * 25
    >>> z = ifft2(ifftshift(Z))
    >>> np.allclose(z.imag, 0)  # z is a real signal
    True
    >>> np.sum(z.real**2)  # z.real is non-zero
    np.float64(50.0)
    >>> x_a = hilbert2(z.real)
    >>> np.allclose(x_a, 0)  # analytic signal is zero
    True

    """
```

scipy/signal/_signaltools.py
```diff
@@ -2626,24 +2653,18 @@ def hilbert2(x, N=None):
...
-    h = h1[:, xp.newaxis] * h2[xp.newaxis, :]
-    k = x.ndim
-    while k > 2:
-        h = h[:, xp.newaxis]
-        k -= 1
-    x = sp_fft.ifft2(Xf * h, axes=(0, 1))
+    Xf = sp_fft.fft2(x, N, axes=axes)
+    Xf = xp.moveaxis(Xf, axes, (-2, -1))
+    end = [N[i] // 2 if N[i] % 2 == 0 else (N[i] + 1) // 2 for i in [0, 1]]
+    start = [N[i] // 2 + 1 if N[i] % 2 == 0 else (N[i] + 1) // 2 for i in [0, 1]]
```
DietBru: Couldn't this be simplified to:
```python
nn = np.array(N)
end, start = (nn + 1) // 2, nn // 2 + 1
```

A cosmetic note of personal taste: Having the variable `N` not denote a single value but a tuple feels off to me.

scipy/signal/tests/test_signaltools.py
```diff
@@ -3497,21 +3498,108 @@ def test_bad_args(self, xp):
...
 
     @pytest.mark.parametrize('dtype', ['float32', 'float64'])
     def test_hilbert2_types(self, dtype, xp):
         dtype = getattr(xp, dtype)
         in_typed = xp.zeros((2, 32), dtype=dtype)
         assert xp.real(signal.hilbert2(in_typed)).dtype == dtype
+    
+    @pytest.mark.parametrize('shape', [(4, 5), (5, 4), (4, 4), (5, 5)])
+    @pytest.mark.parametrize('axes', [(-2, -1), (0, 1)])
+    def test_hilbert2_old_implementation(self, xp, shape, axes):
```
DietBru: I do not think that testing against the old implementation brings a lot of additional value. It is sufficient to test that manually once. It's probably preferable to:

* Write tests for different shapes like in the example in previous [comment], where the result can be easily verified by hand.
* Additionally a test of a 3d signal, where the axes are permuted (with `np.moveaxis`), would be helpful.


 [comment]: https://github.com/scipy/scipy/pull/23247/files#r2278882880

scipy/signal/tests/test_signaltools.py
```diff
@@ -3497,21 +3498,108 @@ def test_bad_args(self, xp):
...
+        x = xp.reshape(xp.arange(sh0 * sh1, dtype=xp.float64), shape)
+        x_as = hilbert2(x)
+        x_as_f = sp_fft.fft2(x_as)
+        n_quads = (freq0 < 0) | (freq1 < 0)
+        if sh0 % 2 == 0:
+            n_quads = n_quads & (freq0 != xp.min(freq0))
+        if sh1 % 2 == 0:
+            n_quads = n_quads & (freq1 != xp.min(freq1))
+        zero_quad = x_as_f[n_quads]
+        xp_assert_close(zero_quad, xp.zeros_like(zero_quad), atol=1e-12)
```
DietBru: It would be great if the values in the non-zero quadrant could be verified too.

DietBru (changes requested):
The refactorization of the code is a helpful simplification. Also allowing higher dimensional arrays by adding an ´axis` parameter` is a win and follows the convention of similar functions (should be added to the release notes). 

I added some suggestions to improve the docstr and the unit tests.

scipy/signal/_signaltools.py
JesseLivezey: That's great. I've included this.

scipy/signal/_signaltools.py
JesseLivezey: Yes, that's nicer.

scipy/signal/tests/test_signaltools.py
JesseLivezey: That's fair, I removed it. I wrote 2 tests that use the types of signals you included in the docstring and 2 tests that check that 2d transforms on 3d arrays work as expected.

scipy/signal/tests/test_signaltools.py
JesseLivezey: Done!

DietBru:
I found a little indexing bug and took the liberty to fix it. While I was at it, I also:

* Refactored test `test_quadrant_power` (now `test_quadrant_values`) to verify that all bins have the correct value.
* Removed `test_impulse` since it does effectively the same thing as `test_quadrant_values` does now.
* Make `xp` last parameter (for sake of convention) in tests.
* Increased test overage of `hilbert2` to 100%
* Cosmetic improvements to the docstr.
* Remove some unneeded trailing spaces.

Feel free to revert anything you do not agree with, @JesseLivezey

DietBru:
I rebased this PR onto the current branch hoping CI would pass. There's seems to be a [problem] in CuPy & friends when doing in this [test]:
```python
       ...
        x_f = xp.ones(shape, dtype=xp.complex128)  # FFT of input signal
        x_f[0 , 0] += 7
        x = xp.real(sp_fft.ifft2(x_f))  # != 0

        x_as = hilbert2(x)  # <--  == 0. (!?)
```
I do not understand why `x` became zero here (while it passes in the other tests). Do you have any insight, @j-bowhay,perhaps?

Rendered docs can be found [here].

[problem]: https://github.com/scipy/scipy/actions/runs/17317409940/job/49162890897?pr=23247#step:12:18059
[test]: https://github.com/JesseLivezey/scipy/blob/2323b0653ce037b0f8edd76ff7b3eaf78b6f5395/scipy/signal/tests/test_signaltools.py#L3412
[here]: https://output.circle-artifacts.com/output/job/62efe5e1-d68c-4026-b6c2-e0da80fc020d/artifacts/0/html/reference/generated/scipy.signal.hilbert2.html#scipy.signal.hilbert2

scipy/signal/tests/test_signaltools.py
```diff
@@ -3372,31 +3372,109 @@ def test_hilbert_types(self, dtype, xp):
...
+
+        x_as = hilbert2(x)
+        x_as_f = sp_fft.fft2(x_as)
+
+        # Create slices for bins with purely positive and purely negative frequencies
+        # (can be verified with `sp_fft.fftfreq()`):
+        f0_pos, f0_neg = slice(1, (shape[0] + 1) // 2), slice((shape[0] + 1) // 2, None)
+        f1_pos, f1_neg = slice(1, (shape[1] + 1) // 2), slice((shape[1] + 1) // 2, None)
+        # Verify all values:
+        atol = np.finfo(x_as_f.dtype).resolution
```
j-bowhay: Unfortunately the `resolution` isn't in the array api standard, maybe something like sqrt(eps) would work?
https://data-apis.org/array-api/latest/API_specification/generated/array_api.finfo.html#finfo

DietBru: Thanks for the hint. This solves only part of the problem. There seems also a CI problem–consult issue #23539.

JesseLivezey:
@DietBru , thanks for the changes. I'll be able to take a look in the next day or two.

JesseLivezey:
@DietBru, your changes look good. I assume we should skip the cupy backend for any test that uses the new `axes` kwarg. I added a commit to try that.

Do you know if cupy tries to stay current on scipy API changes. I can make an issue there.

DietBru:
> 
> Do you know if cupy tries to stay current on scipy API changes.

All I know about CuPy is @j-bowhay's [comment]—I do not understand what is going on there. If @j-bowhay does not object, I would suggest to skip CuPy for the failing tests..




[comment]: https://github.com/scipy/scipy/issues/23539#issuecomment-3244013094

JesseLivezey:
@DietBru , this might explain some of the discrepancies for CuPy
https://github.com/cupy/cupy/blob/f78e78e09cdb5c2d326b069e54f6e8445d7b270d/cupyx/scipy/signal/_signaltools.py#L1751
I think there is a bug in their implementation (they have `h1` twice).

DietBru (approved):
Looks good now. Thanks @JesseLivezey!

DietBru:
Since this PR has now 40 commits, would it be okay, if you squashed those into one or a few commits, @JesseLivezey? There's also merge conflict with main, which requires a rebase onto it. I could take care of that for you, if you do not feeling like doing it.

JesseLivezey:
@DietBru, I'm getting a lot of conflicts internal to this PR when I rebase and squash, which I keep screwing up when I try and resolve them correctly. If you want to take a pass, feel free. Otherwise, I'll try again in a day or two.

FWIW, I was going to squash into 2 commits, roughly 1 that captured the main docs/function commits and 1 that captured the dask/cupy commits.

DietBru:
The squashing required a little bit of manual work. I hope did not overlook anything. If it's fine with you, @JesseLivezey, and CI passed, I think we might be ready for merging...

ev-br:
> I think there is a bug in their implementation (they have h1 twice).

Would be great to report it to CuPy (or even send the PR with a fix straight away, since this looks like a clear typo bug). In SciPy though, we can only rely on a released CuPy, so we'll need to wait until a release with a fix (which is a couple of months at least). So what we could do I guess, is to add a code comment with a link to that issue / PR now, and revisit later.

scipy/signal/tests/test_signaltools.py
```diff
@@ -3388,31 +3388,119 @@ def test_hilbert_types(self, dtype, xp):
...
+
+    def test_parameter_N(self, xp):
+        """Compare passing tuple to single int. """
+        x = xp.zeros((5, 5))
+        x0_a = hilbert2(x, N=4)
+        x1_a = hilbert2(x, N=(4, 4))
+        xp_assert_equal(x1_a, x0_a)
+
+    @pytest.mark.parametrize('shape', [(4, 5), (5, 4), (4, 4), (5, 5)])
+    @skip_xp_backends("cupy", reason="Bug in cupy implementation")
```
JesseLivezey:
```suggestion
    @skip_xp_backends("cupy", reason="Bug in cupy implementation, see cupy#9396")
```

JesseLivezey:
@ev-br, I made a CuPy PR and added a note in the skip reason. I offered to port the scipy tests that caught the problem to CuPy, but I'm not a CuPy user so that might take some time.

I can also make a scipy issue to track unskipping later.

ev-br:
Great. If you point me to the test, I can quickly hammer it into a cupy-compatible shape

scipy/signal/tests/test_signaltools.py
```diff
@@ -3388,31 +3388,119 @@ def test_hilbert_types(self, dtype, xp):
...
+    def test_parameter_N(self, xp):
+        """Compare passing tuple to single int. """
+        x = xp.zeros((5, 5))
+        x0_a = hilbert2(x, N=4)
+        x1_a = hilbert2(x, N=(4, 4))
+        xp_assert_equal(x1_a, x0_a)
+
+    @pytest.mark.parametrize('shape', [(4, 5), (5, 4), (4, 4), (5, 5)])
+    @skip_xp_backends("cupy", reason="Bug in cupy implementation, see cupy#9396")
+    def test_quadrant_values(self, shape, xp):
```
JesseLivezey, line 3436: @ev-br, this test should be ported to CuPy. I think the bug in CuPy means that `hilbert2(x)` will always return `zeros_like(x)` for any `x`.

DietBru:
Unsure what's going on with CI—all failures seem to be caused by the same problem:
```text
Successfully installed Jinja2-3.1.6 MarkupSafe-3.0.2 alabaster-1.0.0 babel-2.17.0 docutils-0.21.2 imagesize-1.4.1 roman-numerals-py-3.1.0 snowballstemmer-3.0.1 sphinx-8.2.3 sphinxcontrib-applehelp-2.0.0 sphinxcontrib-devhelp-2.0.0 sphinxcontrib-htmlhelp-2.1.0 sphinxcontrib-jsmath-1.0.1 sphinxcontrib-qthelp-2.0.0 sphinxcontrib-serializinghtml-2.0.0
Invoking `build` prior to running mypy tests:
$ meson compile -j Sentinel.UNSET -C build
usage: meson compile [-h] [--clean] [-C WD] [-j JOBS] [-l LOAD_AVERAGE] [-v]
                     [--ninja-args NINJA_ARGS] [--vs-args VS_ARGS]
                     [--xcode-args XCODE_ARGS]
                     [TARGET ...]
meson compile: error: argument -j/--jobs: invalid int value: 'Sentinel.UNSET'
Error: Process completed with exit code 2.
```
This branch is based off @lucascolley's  commit d206e414 from today (2025-09-18).

JesseLivezey:
@DietBru, looks like the issue is tracked here https://github.com/scipy/scipy/issues/23642

scipy/signal/tests/test_signaltools.py
ev-br: Here it is @JesseLivezey, as a comment to your CuPy PR:  https://github.com/cupy/cupy/pull/9396#issuecomment-3311419533

Note the process (bit convoluted, yes): 
- SciPy tests are skipped (can actually xfail them instead, but taht's minor) until the CuPy releases a version with your fix
- CuPy tests are skipped until SciPy releases 1.17 and the CuPy CI runs with scipy 1.17

Therefore, this PR can be merged at any time; CuPy PR can be merged any time; once we have the right releases, both libs can undo their skips/xfails. The last steps are for 2026, most likely :-).

JesseLivezey:
@DietBru, looks like things run again after another rebase. The one failure looks unrelated (I hope).

DietBru:
Thank you for your persistence, @JesseLivezey! I agree that this PR is now in good shape for merging

JesseLivezey:
Thanks for your help @DietBru!

I figured this would be a simple repeat of the `hilbert()` PR 🤣

PR #23250, opened 2025-06-27
Remove NumFOCUS as CoC contact

matthew-brett:
As discussed on scipy-core-dev mailing list.

The problem with the current state is that the alternative contact (Ralf) is in fact on the CoC committee, so is not an alternative.

I'm perfectly happy to be that other person, but perhaps we should select someone else, outside Scipy, to be that person - with the understanding it will likely involve somewhere between no work and a very small amount of work.

[skip ci]

### DIFF
--- a/doc/source/dev/conduct/code_of_conduct.rst
+++ b/doc/source/dev/conduct/code_of_conduct.rst
@@ -118,14 +118,11 @@ scipy-conduct@googlegroups.com. Currently, the committee consists of:
 - Nathaniel J. Smith
 - Ralf Gommers
 
-If your report involves any members of the committee, or if they feel they have
-a conflict of interest in handling it, then they will recuse themselves from
-considering your report. Alternatively, if, for any reason, you feel
-uncomfortable making a report to the committee, then you can also contact:
-
-- Chair of the SciPy Steering Committee: Ralf Gommers, or
-- Senior `NumFOCUS staff <https://numfocus.org/code-of-conduct#persons-responsible>`__: conduct@numfocus.org
-
+If your report involves any members of the committee, or if they feel they
+have a conflict of interest in handling it, then they will recuse themselves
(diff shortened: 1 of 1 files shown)


rkern:
I think you got a stray submodule change in there.

matthew-brett:
Oops, thanks, I removed the stray submodule change.

doc/source/dev/conduct/code_of_conduct.rst
```diff
@@ -118,14 +118,11 @@ scipy-conduct@googlegroups.com. Currently, the committee consists of:
...
-uncomfortable making a report to the committee, then you can also contact:
-
-- Chair of the SciPy Steering Committee: Ralf Gommers, or
-- Senior `NumFOCUS staff <https://numfocus.org/code-of-conduct#persons-responsible>`__: conduct@numfocus.org
-
+If your report involves any members of the committee, or if they feel they
+have a conflict of interest in handling it, then they will recuse themselves
+from considering your report. Alternatively, if, for any reason, you feel
+uncomfortable making a report to the committee, then you can also contact the
+chair of the SciPy Steering Committee: Ralf Gommers.
```
tylerjereddy, line 125: Do you want to move forward with the NumFOCUS removal and delay the potential policy modification for the fallback contact for now?

Or do you want to have that fallback discussion inline here now? For folks in other projects, I wonder if someone on the NumPy side who is less/not active in SciPy could make sense. Some "orthogonality" in the alternative contact employer might also make sense perhaps. Not sure how much COI there is for "same employer" situations, but seems potentially messy (i.e., getting "punished" for the same action twice in separate venues, all that stuff...). With a fairly global team there could be all kinds of weird perceived conflicts, so maybe something simple-but-not-perfect for the fallback is the best we could do.

lucascolley: that sounds like a good suggestion to me

Issue #22291, opened 2025-01-09
ENH: interpolate: regular grid linear interpolation for a vector-valued function


Nicholaswogan:
Suppose I have a vector-valued function `f(x) = y`, where `x` and `y` are both vectors of some length (same or different length). `f` could be evaluated over a regular grid of `x` values, producing a set of `y` vectors on that grid. 

I think Scipy should have a function/class for efficiently linearly interpolating such a regular grid, because this kind of interpolation happens very often in many scientific applications. For example, researchers might compute a grid of simulations with a model that takes a lot of computation time. Then, they might want to interpolate that grid to provide an approximate at intermediate points.

It is possible to quickly create this sort of regular grid interpolator with `scipy.interpolate.RegularGridInterpolator`, but it is ridiculously inefficient. For example:

```python
import numpy as np
from scipy.interpolate import RegularGridInterpolator

class VectorLinearInterpolator:

    def __init__(self, points, values_list):    
        self.interps = []
        for values in values_list:
            interp = RegularGridInterpolator(points, values)
            self.interps.append(interp)

    def __call__(self, x):
        res = np.empty(len(self.interps))
        for i in range(len(self.interps)):
            res[i] = self.interps[i](x)[0]
        return res
```

The above class is very similar to `RegularGridInterpolator`, except it takes in the list `values_list` which is the data to be interpolated at each index of `y`. Here is the class in action:

```python
def f(x1, x2, x3, n):
    y = []
    for i in range(n):
        vals = np.sin(np.pi * x1 / 2) * np.exp(x2 / 2) + x3 + i
        y.append(vals)
    return y

np.random.seed(0)

# Array length
n = 5

# Make a grid
x1 = np.linspace(-np.pi, np.pi, 20)
x2 = np.linspace(0.0, np.pi, 20)
x3 = np.linspace(0.0, np.pi/2, 20)
points = (x1,x2,x3)

# Make some fake data on the grid
X1, X2, X3 = np.meshgrid(x1, x2, x3, indexing="ij")
values_list = f(X1, X2, X3, n)

# Coordinates to compute the interpolation at
x = np.array([0.2, np.pi/2.1, np.pi/4.1])

# Initialize interpolator
interp = VectorLinearInterpolator(points, values_list)

# Compute the interpolation
y = interp(x)
print(y)
print(np.array(f(x[0], x[1], x[2], n)))
```

[...]

PR #23253, opened 2025-06-27
BUG: interpolate.RegularGridInterpolator: fix linear interp. on 2D grid with a single point

miccoli:
#### Reference issue
Closes GH-23171

`RegularGridInterpolator` (method=linear) works also in the degenerate case of single-points grids.
However in the specific case of a 2D, float64 grid a Nan is returned.

Tests for generic single-point grids are added and the spurious case is fixed.

#### Additional information
Changes are kept to a bare minimum, however unnecessary branches (where `out[point] = NAN` is set) were simply discarded.


scipy/interpolate/_rgi_cython.pyx
```diff
@@ -27,29 +27,29 @@ def evaluate_linear_2d(const double_or_complex[:, :] values, # cannot declare as
         double_or_complex y0, y1, result
     assert out.shape[0] == num_points
 
-    if grid[1].shape[0] == 1:
-        # linear interpolation along axis=0
+    if grid[0].shape[0] == grid[1].shape[0] == 1:
+        # single grid point along both axes: no interpolation
+        for point in range(num_points):
+            assert indices[0, point] == indices[1, point] == -1
```
tylerjereddy: We usually try to avoid plain `assert` in our source code proper (it is "ok" in `pytest` tests) for a few reasons, including its omission in optimized CPython runs.

It probably isn't too much work to i.e., raise a custom `ValueError` if this check is really needed (and possibly add a regression test for the raising of that error as well, if there is a straightforward way to trigger it).

scipy/interpolate/tests/test_rgi.py
```diff
@@ -335,6 +335,41 @@ def test_fillvalue_type(self):
...
+    def test_length_one_axes(self, dtype, ndim, method):
+        # gh-23171: length-1 axes in all dimensions are legal
+        # for all methods parametrized above
+        xi = np.array([0], dtype=dtype)
+        points = (xi, ) * ndim
+        val = dtype(1/7)
+        fill = dtype(1/42)
+        nshape = (1, ) * ndim
+        values = np.array(val, dtype=dtype)
+        values.shape = nshape
```
tylerjereddy: perhaps a bit more compact/idiomatic to avoid manually setting the shape of the array?

<details>

```diff
--- a/scipy/interpolate/tests/test_rgi.py
+++ b/scipy/interpolate/tests/test_rgi.py
@@ -343,11 +343,9 @@ class TestRegularGridInterpolator:
         # for all methods parametrized above
         xi = np.array([0], dtype=dtype)
         points = (xi, ) * ndim
-        val = dtype(1/7)
-        fill = dtype(1/42)
-        nshape = (1, ) * ndim
-        values = np.array(val, dtype=dtype)
-        values.shape = nshape
+        val = np.asarray(1 / 7, dtype=dtype)
+        fill = np.asarray(1 / 42, dtype=dtype)
+        values = np.full((1,) * ndim, val, dtype=dtype)
```

</details>

scipy/interpolate/tests/test_rgi.py
```diff
@@ -335,6 +335,41 @@ def test_fillvalue_type(self):
...
+        x_nan_0 = np.copy(x0)
+        x_nan_0[0] = np.nan
+        # out-of-bounds point
+        x_oob = np.copy(x0)
+        x_oob[0] += 1
+
+        sample = np.array([x0, x_nan, x_nan_0, x_oob], dtype=dtype)
+        wanted = np.array([val, np.nan, np.nan, fill], dtype=dtype)
+        result = interp(sample)
+        assert_array_equal(result, wanted)
```
tylerjereddy: It may not matter too much, but this test is a fair bit more complex looking than the one in your original ticket--is that just to make it more realistic than the original toy case?

tylerjereddy (commented):
It seems like Evgeni was happy to receive a fix for this so I'll let him review as domain expert.

I added a few small comments/suggestions. I checked locally to confirm fail before/pass after the patch here for the new test, and the fact that you haven't had to modify any old/current tests should help for reviewer confidence.

scipy/interpolate/_rgi_cython.pyx
ev-br: Agreed, we should not be using `assert` for checking things. 

I'm also confused by the logic, actually. This piece of code should not raise at all. Previously, the logic was roughly "if the grid size in this direction is 1, return exactly the corresponding value or nan". I'd expect that the doubly degenerate case might need an extra check, but your patch below removes `out[point] = NAN` for extrapolated values.

scipy/interpolate/tests/test_rgi.py
```diff
@@ -335,6 +335,41 @@ def test_fillvalue_type(self):
         RegularGridInterpolator(points, values)
         RegularGridInterpolator(points, values, fill_value=0.)
 
+    @pytest.mark.parametrize("dtype", [np.float32, np.float64])
+    @pytest.mark.parametrize("ndim", [1, 2, 3])
+    @pytest.mark.parametrize("method", ["linear", "nearest"])
+    def test_length_one_axes(self, dtype, ndim, method):
```
ev-br: Nit: could you rename the test please (took me a while to register the difference between `length_one_axes` and `length_one_axis` below).

scipy/interpolate/tests/test_rgi.py
```diff
@@ -335,6 +335,41 @@ def test_fillvalue_type(self):
...
+        fill = dtype(1/42)
+        nshape = (1, ) * ndim
+        values = np.array(val, dtype=dtype)
+        values.shape = nshape
+        interp = RegularGridInterpolator(
+            points, values, method=method, bounds_error=False, fill_value=fill
+        )
+
+        # grid point
+        x0 = np.concatenate(points)
```
ev-br: Nit: why is it not `points = np.asarray([0]*ndim, dtype=dtype)`

scipy/interpolate/tests/test_rgi.py
ev-br: Nit: please check for nans via `assert np.isnan`

scipy/interpolate/tests/test_rgi.py
```diff
@@ -335,6 +335,41 @@ def test_fillvalue_type(self):
...
+        x_nan_0[0] = np.nan
+        # out-of-bounds point
+        x_oob = np.copy(x0)
+        x_oob[0] += 1
+
+        sample = np.array([x0, x_nan, x_nan_0, x_oob], dtype=dtype)
+        wanted = np.array([val, np.nan, np.nan, fill], dtype=dtype)
+        result = interp(sample)
+        assert_array_equal(result, wanted)
+
```
ev-br, line 359:
```suggestion
```

scipy/interpolate/tests/test_rgi.py
```diff
@@ -335,6 +335,41 @@ def test_fillvalue_type(self):
...
+
```
ev-br, line 370:
```suggestion
```

ev-br (changes requested):
This looks like a good start. Still, 
- the cython implementation should neither `assert` nor raise. IIUC you want to check that if the grid is indeed a single point, then return either the `value` or a nan.
- the test is a little hard to read. IIUC, you are checking that the return values are correct for `xi` being a nan, `xi` being excatly the `point` and `xi` being away from it. If so, great! Please just rewrite the test to separate the cases.

scipy/interpolate/_rgi_cython.pyx
miccoli: The above assert should never raise: it is added only to explicitly state an invariant that is assumed to be true by the current implementation.

I understand that such use of asserts is not common/idiomatic in python... but I felt authorized to add my assert looking at line 28, where there is indeed an assert of this very same nature:

https://github.com/scipy/scipy/blob/4c1a4da8110cc709dc2ff5050d882318fc649e20/scipy/interpolate/_rgi_cython.pyx#L19-L29

When learning C I was taught that asserts should be used only for debugging, and that it is good practice to leave them also in production code (which is typically compiled with `-DNDEBUG`).
The reasons:
- explicitly document the invariants which are assumed to be valid in some point of the code
- help debugging regressions if for some reason the invariants fail in some future release.

So actually, the above assert should never raise and could be simply deleted or commened out.
Please let me know if I should 
- delete this assert (but then also the one at line 28), 
- add a message that clarifies its debugging-only use
- comment it out.

scipy/interpolate/tests/test_rgi.py
miccoli: Here setting the shape directly stresses the fact that, regardless of the number of dimensions `ndim`, the `values` array should have size=1: so it is first constructed as a 0D array which is promoted to `ndim`D by assigning the shape.

Similarly `val` and `fill` are meant to be numpy scalars and not 0D arrays... 

Currently numpy is a little bit confusing in the treatment of 0D arrays vs. scalars, so I prefer to be 100% explicit on what is meant.

scipy/interpolate/tests/test_rgi.py
miccoli: These test are meant to expliciltly show the desired behaviour... I added them to be sure that my patch (expressly deleting `out[point] = NAN` clause) would not introduce other regressions.

scipy/interpolate/tests/test_rgi.py
miccoli: OK, I will rename the new test...

miccoli: Just a question of style: some projects implement a strict requirement of a single assert per unit test (this to have a single point of failure... instead of a sequnce of `assert A`, some code, `assert B`, some other code `assert C`).

Personally I like the idea of having a single assert, and indeed this was the reason for dumping all the tests in a single vector.

Should I split the different cases as multiple asserts in the same method, or should I write multiple methods with a single assert?

scipy/interpolate/_rgi_cython.pyx
ev-br: > delete this assert (but then also the one at line 28),

This please. 

This whole thing is quite firmly in the "tastes differ" territory and indeed there might be very good reasons to keep asserts for self-documenting reasons, either in python or C. Here though, having both these asserts is more confusing then helpful --- and I'm to blame for the one in line 28, so thank you for cleaning up the mess I left in this code!

My reasoning for leaving the assert on line 28 was probably along the same lines of reasoning you discuss, but then a couple of years later it confuses multiple people, which means it was a bad idea.

scipy/interpolate/tests/test_rgi.py
ev-br: This is even further into the "tastes differ" territory indeed. We certainly do not have a strict requirement of a single assertion for a single test. In the opposite limit, having a single test with 25 assertions is not nice either. Where to draw a line is a matter of taste in the end. Here though, you've what three or four assertions, and it seems at least to me that having separate assertions makes the test more readable (and saves a step in debugging if it ever fails).

scipy/interpolate/tests/test_rgi.py
ev-br: numpy scalars vs 0D arrays are confusing indeed! That said, the way it's written now makes it more confusing then just creating the right-shaped array (have I mentioned that we're in the "matters of taste" territory already?)

> val and fill are meant to be numpy scalars and not 0D arrays

why though? You want to reject fill_value and values being numpy scalars? Then please add an explicit assertion that it raises. I wouldn't though, exactly because it's confusing in numpy, and it's not always easy to predict if something is a scalar or a 0D array. (and in the spirit of array API, consider actually just calling `np.asarray` on something that might be a scalar)

scipy/interpolate/tests/test_rgi.py
ev-br: Thank you for making sure of not introducing regressions, this is great! The ask to only split a single "assert four things in one line" into several assertions.

TBH I find it a bit surprising that these `out[point]=NAN` branches can be just removed. Will need to audit test coverage I suppose.

As an aside, this whole 2D fast path should be rewritten at some point, cf gh-22291. This is of course absolutely separate, am only mentioning it in case you'd be interested in working on it after this one :-)

scipy/interpolate/tests/test_rgi.py
miccoli: I thought `val` (scalar) as a building block for `values` (array). Similarly with `fill` (scalar), I build the array of desired values in `wanted`.

Why? Actually there is no compelling reason, apart from the fact that I seems more natural to me to build N-dimensional arrays from scalars, and not 0-dimensional arrays.

Again I see here that this is not idiomatic, so I will rewrite the tests along your lines.

scipy/interpolate/tests/test_rgi.py
miccoli: OK, clear enough, hope to fix this soon.

scipy/interpolate/tests/test_rgi.py
miccoli: > TBH I find it a bit surprising that these `out[point]=NAN` branches can be just removed. Will need to audit test coverage I suppose.

Nan and out-of-bounds are handled by the caller here:
https://github.com/scipy/scipy/blob/4c1a4da8110cc709dc2ff5050d882318fc649e20/scipy/interpolate/_rgi.py#L443-L448
so no need to add a sentinel in `out[point]`.

miccoli:
A few comments on the last modifications:
- 2f313bcb006887b468e3cb104a29ae6bfde44b85: delete `assert` from code
- 18d1ac8dd5569c640ce4152490292d21f34a3db1, 47ff469edcc08257d4dea614f65c7c27d8cfa152: refactor tests, using only array API functions. Tests are split in independent asserts. Code is a little less DRY, but I hope it is more legible, and not too verbose.
- 936e7c1c55566b27ae88b3e4e27328c0537cbb62: merge main branch, to resolve conflicts. I noticed that the top module import of `np.testing.array_assert_equal` was deleted. Is there a preferred method for array equality testing?

ev-br:
> Is there a preferred method for array equality testing

Please use `xp_assert_close` or `xp_assert_equal` for exact equality.

ev-br (approved):
LGTM, thank you @miccoli. And congrats with the first scipy contribution, keep them coming!

miccoli:
> LGTM, thank you @miccoli. And congrats with the first scipy contribution, keep them coming!

Thanks for merging. 

This was a rather straightforward contribution (but perhaps ideal to learn the scipy coding conventions): hope to do some more substantial contributions in the future.

PR #23255, opened 2025-06-28
MAINT: ``spatial._plotutils``: remove matplotlib 1 code

jorenham:
I believe that the lower bound for matplotlib is currentlly set to 3.5.

I also removed the ``_lib.decorator`` stuff for the sake of simplicity. Mypy should also be able to understand the code a bit better this way.

### DIFF
--- a/scipy/spatial/_plotutils.py
+++ b/scipy/spatial/_plotutils.py
@@ -1,28 +1,12 @@
 import numpy as np
-from scipy._lib.decorator import decoratorx as _decorator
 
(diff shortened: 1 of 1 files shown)


lucascolley (approved):
thanks Joren!

tylerjereddy (approved):
Agreed, it has been almost 9 years since the last `1.x` series `matplotlib` release and this is easier to read as well.

The CI does have two jobs that run with `matplotlib`, and they are both passing.

Incidentally, both of those jobs have `py3.11` hardcoded in their name even though one of them run with Python `3.12` because of a `matrix` config split, but that is unrelated and super low prio.

tylerjereddy:
thanks both

PR #9623, opened 2018-12-21
DOC: Build docs with warnings as errors

larsoner:
Inspired by @rgommers in #9618 I'm trying to build with `-nWT --keep-going` (nitpicky, turn warnings into errors, full traceback on error, and keep going until the end to show an error). I now have it down to a few dozen lines of errors, then I'll enable "don't allow warnings in examples" (with exceptions of course) because it seems like good practice.

PR #23257, opened 2025-06-28
DOC: add orphaned modules to toctree

scottshambaugh:
[docs only]

#### Reference issue
Closes https://github.com/scipy/scipy/issues/23237

There are four modules which are currently marked as orphaned and not part of the toctree, which hurts navigation in the docs: `io.matlab`, `signal.window`, `spatial.transform`, and `special.cython_special`. Screenshots and examples in the linked issue.

This PR includes these modules in the toctree. I don't know of a good reason why they would have been excluded in the first place, so please weigh in if there is rationale besides this being an oversight.

### DIFF
--- a/doc/source/reference/io.matlab.rst
+++ b/doc/source/reference/io.matlab.rst
@@ -1,5 +1,3 @@
-:orphan:
-
 .. automodule:: scipy.io.matlab
    :no-members:
    :no-inherited-members:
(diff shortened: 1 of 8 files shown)


scottshambaugh:
Built docs look good to me

scottshambaugh:
It looks like the orphan tags were added in https://github.com/scipy/scipy/pull/9623 to suppress the build warnings. So IMO bringing them into the toctree is the proper fix

melissawm (approved):
Looks great, and I agree this is appropriate. Thanks @scottshambaugh !

j-bowhay (approved):
Also LGTM

PR #23258, opened 2025-06-28
MAINT: ``sparse`` testing without ``decorator``

jorenham:
After #23256 and #23255, this was the last place where ``scipy._lib.decorator`` was being used. So my plan is to remove it in a follow-up.

The refactored decorator is equivalent to what it previously was, and is only in a couple of ``sparse`` tests.


jorenham:
oh woops, I see I messed up the submodules

lucascolley (approved):
thanks Joren

PR #23259, opened 2025-06-28
ENH: stats: ``Logistic`` distribution

jorenham:
I thought I'd try out the new distribution infrastructure for a bit, and then somehow this happened :shrug:.

---

Fun fact about the *standard* logistic distribution:
Its *standard* deviation is $\sigma = \pi/\sqrt{3}$, but its second L-moment, the L-scale, is $\lambda_2 = 1$. And for the general L-moment of order $r > 1$ it's $\displaystyle \lambda_r = \frac{(-1)^r+1}{(r-1) \cdot r}$, which is a lot simpler than for the conventional (raw/central/standardized) moments.


jorenham:
Something tells me that you might be interested in this one, @mdhaber  :)

scipy/stats/_new_distributions.py
```diff
@@ -200,6 +200,76 @@ def _sample_formula(self, full_shape, rng, **kwargs):
...
+
+    The probability density function of the standard logistic distribution is:
+
+    .. math::
+
+        f(x) = \frac{1}{\left( e^{x / 2} + e^{-x / 2} \right)^2}
+
+    """
+    _x_support = _RealInterval(endpoints=(-inf, inf))
+    _variable =_x_param = _RealParameter('x', domain=_x_support, typical=(-9, 9))
```
lucascolley:
```suggestion
    _variable = _x_param = _RealParameter('x', domain=_x_support, typical=(-9, 9))
```

jorenham:
Should I add the release notes directly to `doc/source/release/1.17.0-notes.rst`?

lucascolley:
> Should I add the release notes directly to doc/source/release/1.17.0-notes.rst?

I think we typically compile things in the Wiki first, but we can see what @tylerjereddy would prefer.

tylerjereddy:
Yeah, if you could use the wiki (https://github.com/scipy/scipy/wiki/*-Release-Note-Entries-for-SciPy-1.17.0-(asterisk-makes-this-appear-at-the-top)) for now that would be great. Ilan already started editing it.

Maybe we'll change that workflow at some point, but its simplicity has suited reasonably well so far.

jorenham:
> Yeah, if you could use the wiki ([Wiki: * Release Note Entries for SciPy 1.17.0 (asterisk makes this appear at the top)](https://github.com/scipy/scipy/wiki/*-Release-Note-Entries-for-SciPy-1.17.0-(asterisk-makes-this-appear-at-the-top))) for now that would be great. Ilan already started editing it.

Eh.. how 😅?

lucascolley:
> Eh.. how 😅?

If you don't have permissions to edit the wiki, just post a comment here with the text, and a maintainer will copy it across on merge and remove the label

jorenham:
> > Eh.. how 😅?
> 
> If you don't have permissions to edit the wiki, just post a comment here with the text, and a maintainer will copy it across on merge and remove the label

I suppose that'd look something like this:

> # ``scipy.stats`` improvements
> 
> - ...
> - A new distribution, ``scipy.stats.Logstic``, was added.
> - ...

mdhaber:
@jorenham What do you think about (separately) adding l-moments to the infrastructure? We already have `order_statistic`, which can get the expectations needed. Would it be terrible to add it as a a `kind` of [`moment`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.Normal.moment.html#scipy.stats.Normal.moment) alongside `raw`, `central`, and `standardized`?  I suppose l-moment ratios would be another `kind`. We could have a separate `lmoment` method. I suppose the benefit is bit simpler documentation and the con is a little duplication, since the interfaces of the two will be essentially identical.

BTW, will you be at the SciPy conference next week?

jorenham:
> What do you think about (separately) adding l-moments to the infrastructure?

Best idea since the invention of food./

> We could have a separate `lmoment` method.?

Yea I think that's the best approach. 
The difference between raw/central/standardized is on completely different scale than the difference between conventional- and L-moments. We also have separate `stats.moment` and `stats.lmoment` functions. 

---

I've also considered this before, and that's in fact why I was looking into the new distribution infrastructure, which side-tracked into this PR 😛.

I was picturing something like

```pyi
@overload
def lmoment(
    self,
    order: int, 
    *,
    ratio: bool = False,
    trim: int | tuple[int, int] = 0,
) -> np.float64: ...
@overload
def lmoment(
    self,
    order: Sequence[int], 
    *,
    ratio: bool = False,
    trim: int | tuple[int, int] = 0,
) -> Array1D[np.float64]: ...
```

When working with L-moments, you often need more than one, so that's what the second overload is for. 

---

FWIW, [Lmo](https://github.com/jorenham/lmo/) currently patches the `rv_continuous` and `rv_continuous_frozen` classes so that you can do something similar:

```pycon
>>> import lmo
>>> from scipy.stats import logistic
>>> logistic().l_moment([1, 2, 3, 4])
array([0.        , 1.        , 0.        , 0.16666667])
```

And not to pat myself on the back or anything, but I found that in practice, it's very convenient this way.

jorenham:
> BTW, will you be at the SciPy conference next week?

I'm afraid I won't. I'm planning on going to PyData Amsterdam in september though.

mdhaber:
Technically, this all looks good.
- Docs render well
- CDF definition agrees with literature, and I trust the test suite to check for consistency
- It looks like the location/scale parameterization is standard, so I think it's fine to leave this without parameters

Reading this brings to mind that I've wanted to add logic for symmetric distributions that would avoid the need to separately define cdf/ccdf, etc. But that can wait. Also, for distributions without parameters, there is currently no way to control the dtype of some methods (e.g. `mean`, `sample`). I bring this up because I noticed the Python math and use of `float` - that's OK here because it turns into `float64` for now. We've considered adding a standard `dtype` parameter for that reason. But that, too, can wait.

One thought is that we should have added one distribution-specific test for each of the existing distributions to show that they are indeed the distribution they purport to be. One could say the same here, but I'd prefer to let that be in this PR and add all those needed tests in a separate PR.

The main question is whether we want to add `Logistic`. That's a question for the mailing list. 
(Personally, I'd like to not go too far down the list of popular distributions because at some point I want to get `scikit-stats` up and running to accept community-contributed and maintained distributions much more flexibly. But `Logistic` might be OK because everything can be expressed pretty neatly, so we're unlikely to need much maintenance on this one.)

mdhaber:
@jorenham By mailing list, I meant [Discuss](https://discuss.scientific-python.org/c/contributor/scipy/32). Did you want to propose this? I can't merge without it. 
If it's approved, I'll add a test, then merge.

scipy/stats/_new_distributions.py
```diff
@@ -200,6 +200,76 @@ def _sample_formula(self, full_shape, rng, **kwargs):
...
+        return special.logit(x)
+
+    def _iccdf_formula(self, x, **kwargs):
+        return -special.logit(x)
+
+    def _entropy_formula(self, **kwargs):
+        return 2.0
+
+    def _logentropy_formula(self, **kwargs):
+        return np.log(2)
```
ev-br, line 249: Nit, optional: math.log(2) ? 
And similar for the _scale classvar

jorenham: I was first thinking of going with `scipy.integrate.cubature(lambda t: 1/t, [1], [2]).estimate.item()` but I like your idea better :P

jorenham: Hmm, it'll be a inconsistent with `Normal` this way 🤔 

https://github.com/scipy/scipy/blob/d206e414575046a315781b1a8cfc9a167fd8d86f/scipy/stats/_new_distributions.py#L41-L42

It also requires an additional `math` import.

So do we still want this?

mdhaber: Go ahead and change `Normal` to use `math`, too. Eventually this will be converted to array API, and in that case there are some additional advantages to using math instead of `xp`.

(Please check that the return value is still a NumPy float, though? It should currently be a `float64`.)

jorenham: I think that it would be cleaner to address that in a separate PR.

jorenham:
> @jorenham By mailing list, I meant [Discuss](https://discuss.scientific-python.org/c/contributor/scipy/32). Did you want to propose this? I can't merge without it. If it's approved, I'll add a test, then merge.

I opened https://discuss.scientific-python.org/t/new-scipy-stats-logistic-distribution/2118

mdhaber:
There's a heart here, an implicit approval of the idea from a maintainer, another maintainer mentioned approval to me, and there's lazy consensus from the Dicuss forum, so in it goes. Thanks @jorenham.

moizescbf:
@mdhaber out of curiosity, what is **scikit-stats** that you said you want to get "up and running to accept community-contributed and maintained distributions much more flexibly"? Because I really like the new statistical distribution infrastructure and wonder what could this scikit-stats be to facilitate so much the creation of new distributions on that new infrastructure

mdhaber:
The `scikit-stats` repo is [here](https://github.com/statistical-python/scikit-stats). The plan has been to set up a standardized system for testing and review so that progress on distributions is made by those who are interested and not held up by SciPy stats maintainers (e.g. me). Progress on that is slow because there are many other projects and there has been limited interest. In the meantime, you can use most old distributions via the new interface with [`make_distribution`](https://docs.scipy.org/doc/scipy-1.16.2/reference/generated/scipy.stats.make_distribution.html), and you can also use `make_distribution` to create new distributions. When the infrastructure of `scikit-stats` is set up, you will be able to contribute it there to share it with others.

PR #23266, opened 2025-07-01
BUG: signal.remez: fix handling of `weight` array

gmgunter:

The `signal.remez` function can take an optional array of relative weights. In f45ba7bd311ee60c5c2f74c272cd3cd6c1dece86, logic was added that checks to see if weights were supplied and, if so, convert them to an array object.

https://github.com/scipy/scipy/blob/f45ba7bd311ee60c5c2f74c272cd3cd6c1dece86/scipy/signal/_fir_filter_design.py#L836-L837

However, the `if weight` check raises a `ValueError` if `weight` is a NumPy array with more than one element.

```
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

It seems best to explicitly check `if weight is not None`.

### DIFF
--- a/scipy/signal/tests/test_fir_filter_design.py
+++ b/scipy/signal/tests/test_fir_filter_design.py
@@ -546,9 +546,11 @@ class TestRemez:
         with pytest.raises(ValueError, match="Sampling.*single scalar"):
             remez(11, .1, 1, fs=np.array([10, 20]))
 
-    def test_gh_23266(self):
-        weight = np.asarray([1.0, 2.0])
-        remez(21, [0.0, 0.2, 0.3, 0.5], [1.0, 0.0], weight=weight)
+    def test_gh_23266(self, xp):
+        bands = xp.asarray([0.0, 0.2, 0.3, 0.5])
+        desired = xp.asarray([1.0, 0.0])
+        weight = xp.asarray([1.0, 2.0])
+        remez(21, bands, desired, weight=weight)
 
 
 @skip_xp_backends(cpu_only=True, reason="lstsq")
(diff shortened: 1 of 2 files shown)


tylerjereddy (commented):
I'm not a domain expert, but if we do this we may want a regression test like the one below that fails before/passes after the patch is applied (could probably be improved).

<details>

```diff
--- a/scipy/signal/tests/test_fir_filter_design.py
+++ b/scipy/signal/tests/test_fir_filter_design.py
@@ -546,6 +546,12 @@ class TestRemez:
         with pytest.raises(ValueError, match="Sampling.*single scalar"):
             remez(11, .1, 1, fs=np.array([10, 20]))
 
+    def test_gh_23266(self):
+        numtaps = 21
+        bands = [0, 0.1, 0.15, 0.35]
+        desired = [1, 0]
+        weights = np.asarray([1, 2])
+        remez(numtaps, bands, desired, weight=weights, fs=1.0)
```


</details>

scipy/signal/tests/test_fir_filter_design.py
```diff
@@ -546,6 +546,9 @@ def test_fs_validation(self):
         with pytest.raises(ValueError, match="Sampling.*single scalar"):
             remez(11, .1, 1, fs=np.array([10, 20]))
 
+    def test_gh_23266(self):
+        weight = np.asarray([1.0, 2.0])
+        remez(21, [0.0, 0.2, 0.3, 0.5], [1.0, 0.0], weight=weight)
```
j-bowhay: This function supports the array api so we should use the `xp` fixture to test with other backends too
```suggestion
    def test_gh_23266(self, xp):
        weight = xp.asarray([1.0, 2.0])
        remez(21, xp.asarray([0.0, 0.2, 0.3, 0.5]), xp.asarray([1.0, 0.0]), weight=weight)
```

j-bowhay (approved):
Thanks for your first contribution to Scipy @gmgunter! Keep them coming

tylerjereddy:
Looks like this is related to a PR merged in`1.16.0`, so tentatively adding backport label.

PR #23267, opened 2025-07-01
MAINT: sparse: move `_validate_indices` and `_asindices` from methods to functions

dschult:
This PR prepares `_indexing.py` for the nD indexing provided in #23218 
by refactoring two methods into functions.  Based on [#23218(comment)](https://github.com/scipy/scipy/pull/23218#discussion_r2162344895).

The idea is to avoid subclassing just to get those two methods. They only use `self.shape` and `self.format` so we pass those in as arguments.  `self.format` is used in two ways within `_asindices`:
- if format is LIL we don't need to check index bounds (they are checked in the updating methods).
- if the format is COO we don't need to force index arrays to be 1D or 2D.

This is not part of #23218 to make review easier. 

Note: To ease review, look at the changes in the second commit. The first commit only changes the indent and position within the file. Splitting these two makes the code changes easier to see.

### DIFF
--- a/scipy/sparse/_index.py
+++ b/scipy/sparse/_index.py
@@ -27,7 +27,7 @@ class IndexMixin:
     This class provides common dispatching and validation logic for indexing.
     """
     def __getitem__(self, key):
-        index, new_shape = self._validate_indices(key)
+        index, new_shape = _validate_indices(key, self.shape, self.format)
 
         # 1D array
         if len(index) == 1:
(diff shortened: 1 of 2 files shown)


lucascolley (approved):
thanks Dan, LGTM!

Maybe CJ would like to take a look.

dschult:
@perimosocordiae when you get a chance can you look at this PR? 
It is fairly straight foward and merging it will allow me to refactor the nD indexing to use this approach.

The first commit just changes the indent level -- the second updates the code to work as functions instead of methods.

perimosocordiae (approved):

perimosocordiae:
Thanks @dschult , merged!

PR #23273, opened 2025-07-01
DOC: interpolate.make_splprep: fix dimensions ordering

anntzer:
It is maybe a bit unfortunate that make_splprep takes input with (ndim, npoints) shape rather than the other way round, but changing that would be quite tedious wrt. deprecation and API changes; instead, just document the actual behavior.

To verify this, try e.g. running the make_splprep examples in smoothing_splintes.rst but transposing the arrays (replace `[xn, yn]` by `np.column_stack([xn, rn])`); this results in a ValueError.

### DIFF
--- a/doc/source/tutorial/interpolate/smoothing_splines.rst
+++ b/doc/source/tutorial/interpolate/smoothing_splines.rst
@@ -343,8 +343,8 @@ problem it solves.
 The main user-visible difference of the parametric case is the user interface:
 
 - instead of two data arrays, ``x`` and ``y``, `make_splprep` receives a single
-  two-dimensional array, where the second dimension has size :math:`d` and each
-  data array is stored along the first dimension (alternatively, you can supply
+  two-dimensional array, where the first dimension has size :math:`d` and each
+  data array is stored along the second dimension (alternatively, you can supply
   a list of 1D arrays).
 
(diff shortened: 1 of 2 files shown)


ev-br (approved):
Yeah, thanks. The docs were clearly wrong before!

Re the ordering of dimensions: `make_splprep` tries to be API-compatible with `splprep` which accepts the list of 1D arrays. That behavior is fixed from day one, so we're going to have to stick to it.

Now that you recently worked through this @anntzer , you've got fresh eyes on this mess: would it be helpful to phrase it starting from a list of 1D arrays, alternatively a 2D array of shape `(ndim, m)`?

anntzer:
> Now that you recently worked through this @anntzer , you've got fresh eyes on this mess: would it be helpful to phrase it starting from a list of 1D arrays, alternatively a 2D array of shape `(ndim, m)`?

The docstring already states that: `The typical use is a list of 1D arrays, each of length m.`

ev-br:
Thanks @anntzer for noticing and fixing this inconsistency. 

> The docstring already states that

My question was if it'd be helpful to switch the order and mention the list of arrays first and move a 2D array input to the "alternatively" clause, but I see that the answer is a no. So let's roll with the now-fixed wording. Thanks again!

PR #23275, opened 2025-07-01
TST: migrate away from suppress_warnings and assert_warns

ngoldbaum:
#### Reference issue
xref https://github.com/numpy/numpy/issues/29293

Both `np.testing.suppress_warnings` and `np.testing.assert_warns` are not thread-safe. Now that the `warnings` module is thread-safe by default on free-threaded 3.14, I finally noticed that this is a problem. See the NumPy issue for more details.

This migration enables running all the modified tests in parallel on free-threaded 3.14. I have a follow-up to update the tests marked as thread-unsafe and re-enable parallel testing.

#### Additional information

Question for reviewers: should I try to make the diff more minimal? 

I tried to avoid leaving behind out-of-date idioms from the era of `nosetests` so there's a bit more test rewriting than is strictly necessary to make the diff as minimal as possible. 

I think it's probably better for the SciPy tests to use more modern idioms, even if it means the diff here is pretty big.


steppi:
I think the diff doesn't look too bad. +1 for using the modern idioms.

scipy/_lib/tests/test_warnings.py
```diff
@@ -54,7 +54,8 @@ def visit_Call(self, node):
                 case _:
                     raise ValueError("unknown ast node type")
             # check if filter is set to ignore
-            if argtext == "ignore":
+            any_tests = any(p.name.endswith("tests") for p in self.__filename.parents)
+            if argtext == "ignore" and not any_tests:
```
steppi: I skimmed the diff and this seems to be the only change that isn't a mechanical change in warning suppression from the old way to the new way. This logic for allowing the use of `warnings.catch_warnings` to ignore warnings in tests looks correct.

A minor nit: I think

python
in_test = "tests" in self.__filename.parts

might more clearly show intent and should work just as well since there aren't tests folders with names matching ".+tests" as far as I'm aware. I get that you must have wrote this quickly just to get CI to pass again though.

I also suggest updating the comment to say something like.

```python
# check if filter is set to ignore outside of test code
```

ngoldbaum:
CI failures look to me like unrelated flakiness.

steppi (approved):
Looks good to me. Thanks @ngoldbaum. I'll merge when CI settles

steppi:
Sorry, hesitant to push the button because I can't tell if some the failures could possibly be related. It at least seems unlikely that they are related though.

ngoldbaum:
@steppi here's a CI run failing in exactly the same way a couple weeks ago: https://github.com/scipy/scipy/actions/runs/15731877458/job/44334684270

I think the circleci one is because of transient issues with debian's or canonical's package infrastructure.

I think if you re-trigger CI for both of the failing runs, it will almost certainly pass.

ngoldbaum:
another one: https://github.com/scipy/scipy/actions/runs/15819266405/job/44584492596

ngoldbaum:
In NumPy's CI I dealt with hanging tests like this by adding a dependency on `pytest-timeout` and setting a per-test timeout. That also gives you a C traceback if the timeout is triggered because it calls faulthandler.

steppi:
Thanks @ngoldbaum for checking into that.

PR #23276, opened 2025-07-01
BUG: fix signature of ``@_transition_to_rng`` functions (SPEC 7)

### LINKED ISSUE #23216: BUG: `@_transition_to_rng` leads to inconsistent function signatures
For example in the `stats.qmc` engines, the `seed` parameter can still be used in place of `rng` (and is not deprecated yet). For this SPEC 7 transition a convenient `_lib._util._transition_to_rng` decorator is used, effectively mapping calls with `seed={}` to `rng={}`. And even though `rng` and `seed` are not allowed to be used together, this "added" `seed` parameter should also be present in the function signature:

```py
>>> import inspect
>>> import scipy
>>> list(inspect.signature(scipy.stats.qmc.Sobol).parameters)
['d', 'scramble', 'bits', 'rng', 'optimization']
>>> scipy.__version__
'1.16.0'
```

Mypy's stubtest relies on these signatures, and uses that to validate if `scipy-stubs` matches the runtime implementation. The result is that I have removed all `seed` kwargs from methods and functions that are decorated with `@_transition_to_rng`, because I assumed that stubtest was right (as it almost always is). I just discovered that this was not the case, leading to false rejections (i.e. CI failures) in libraries such as `optuna` that use `scipy-stubs` (scipy/scipy-stubs/issues/644). 

If I now restore the `seed` parameter, stubtest gets angry. This forces me to mark all of these methods as ignored by stubtest as a whole, which results in blind spots.

The `@_transition_to_rng` decorator looks pretty complicated, and I couldn't immediately find an obvious solution to this issue. I'm also not very familiar with the low-level machinery behind`inspect.signature`, so I'm afraid that there's no solution I can propose here.

jorenham:
This adds the missing parameter to the signature of functions decorated with ``scipy._lib._util._transition_to_rng``, and closes #23216.

**Before:**

```pycon
>>> from scipy.linalg import clarkson_woodruff_transform
>>> import inspect
>>> inspect.signature(clarkson_woodruff_transform)
<Signature (input_matrix, sketch_size, rng=None)>
```

**After:**

```pycon
>>> from scipy.linalg import clarkson_woodruff_transform
>>> import inspect
>>> inspect.signature(clarkson_woodruff_transform)
<Signature (input_matrix, sketch_size, rng=None, *, seed=None)>
```

---

The reason I call this is bug, is because it's (still) perfectly valid to use the `seed` parameter for SPEC 7 functions:

```pycon
>>> import numpy as np
>>> clarkson_woodruff_transform(np.eye(2), 1, seed=1)
array([[-1., -1.]])
```

i.e. no (deprecation- or future-) warnings are shown.

See #23216 for the consequences of this missing signature parameter.

---

For reference: <https://scientific-python.org/specs/spec-0007/>

### DIFF
--- a/scipy/_lib/_util.py
+++ b/scipy/_lib/_util.py
@@ -342,6 +342,13 @@ def _transition_to_rng(old_name, *, position_num=None, end_version=None,
 
             return fun(*args, **kwargs)
 
+        # Add the old parameter name to the function signature
+        wrapped_signature = inspect.signature(fun)
+        wrapper.__signature__ = wrapped_signature.replace(parameters=[
+            *wrapped_signature.parameters.values(),
+            inspect.Parameter(old_name, inspect.Parameter.KEYWORD_ONLY, default=None),
+        ])
+
         if replace_doc:
             doc = FunctionDoc(wrapper)
             parameter_names = [param.name for param in doc['Parameters']]


mdhaber (approved):
Ah, I thought you wanted more to change than *just* getting `inspect` to report this as a keyword. If that's all you need, LGTM if CI looks good.

tylerjereddy:
@jorenham I'll tentatively add a backport label, although if this is typing specific not sure if it is critical. Also, we're starting to think about the `1.16.1` release as mentioned this morning, in case you need a heads up on the stubs side of things.

jorenham:
> @jorenham I'll tentatively add a backport label, although if this is typing specific not sure if it is critical. 

That would help a lot for scipy-stubs!

> Also, we're starting to think about the `1.16.1` release as mentioned this morning, in case you need a heads up on the stubs side of things.

Good to know :)

PR #23279, opened 2025-07-01
BUG: linalg: Fix pointer casting order for sqrtm

### LINKED ISSUE #23278: BUG: <Please write a comprehensive title after the 'BUG: ' prefix>
### Describe your issue.

`sqrtm` return wrong results with 1.16.0 for simple matrices:

The issue seems to happen when the matrix has at least one isolated diagonal element and is complex.

### Reproducing Code Example

```python
mat = np.array([
    [1, 0, 0],
    [0, 1, -1j],
    [0, 1j, 2],
])
sq = scipy.linalg.sqrtm(mat)
sq @ sq
```

### Error message

```shell
Should return the original matrix, but return:

array([
    [ 2.10557281+0.j,  0. + 0.82917961j,  -0.40162283+0.j],
    [ 0. -0.82917961j,  1.2763932 + 0.j  , 0. -0.64983939j],
    [-0.40162283+0.j,  0. + 0.64983939j, 0.76393202+0.j,
])
```

```shell
1.16.0 2.3.1 sys.version_info(major=3, minor=13, micro=2, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    detection method: pkgconfig
    found: true
    include directory: /opt/_internal/cpython-3.13.4/lib/python3.13/site-packages/scipy_openblas32/include
    lib directory: /opt/_internal/cpython-3.13.4/lib/python3.13/site-packages/scipy_openblas32/lib
    name: scipy-openblas
    pc file directory: /project
    version: 0.3.28
  lapack:
    detection method: pkgconfig
    found: true
    include directory: /opt/_internal/cpython-3.13.4/lib/python3.13/site-packages/scipy_openblas32/include
    lib directory: /opt/_internal/cpython-3.13.4/lib/python3.13/site-packages/scipy_openblas32/lib
    name: scipy-openblas
    pc file directory: /project
    version: 0.3.28
  pybind11:
    detection method: config-tool
    include directory: unknown
    name: pybind11
    version: 2.13.6
Compilers:
  c:
    commands: cc
    linker: ld.bfd
    name: gcc
    version: 10.2.1
  c++:
    commands: c++
    linker: ld.bfd
    name: gcc
    version: 10.2.1
  cython:
    commands: cython
    linker: cython
    name: cython
    version: 3.1.2
  fortran:
    commands: gfortran
    linker: ld.bfd
    name: gcc
    version: 10.2.1
  pythran:
    include directory: ../../tmp/build-env-gy_g1nhu/lib/python3.13/site-packages/pythran
    version: 0.18.0
Machine Information:
  build:
    cpu: x86_64
    endian: little
    family: x86_64
    system: linux
  cross-compiled: false
  host:
    cpu: x86_64
    endian: little
    family: x86_64
    system: linux
Python Information:
  path: /tmp/build-env-gy_g1nhu/bin/python
  version: '3.13'
```

ilayn:
#### Reference issue
Closes #23278 

#### Additional information

The pointer casting was done in the reverse order, causing to overwrite the original array.

### DIFF
--- a/scipy/linalg/_matfuncs_sqrtm.c
+++ b/scipy/linalg/_matfuncs_sqrtm.c
@@ -533,7 +533,7 @@ matrix_squareroot_c(const PyArrayObject* ap_Am, SCIPY_C* restrict ret_data, int*
     SCIPY_C* restrict data = &buffer[0];
     SCIPY_C* restrict vs = &buffer[n*n];
     SCIPY_C* restrict w = &buffer[2*n*n];
-    float* restrict rwork = &((float*)buffer)[2*n*n + n];
+    float* restrict rwork = (float*)(&buffer[2*n*n + n]);
     SCIPY_C* restrict work = &buffer[2*n*n + 2*n];
 
     for (npy_intp idx = 0; idx < outer_size; idx++) {
@@ -635,7 +635,7 @@ matrix_squareroot_z(const PyArrayObject* ap_Am, SCIPY_Z* restrict ret_data, int*
     SCIPY_Z* restrict data = &buffer[0];
     SCIPY_Z* restrict vs = &buffer[n*n];
     SCIPY_Z* restrict w = &buffer[2*n*n];
-    double* restrict rwork = &((double*)buffer)[2*n*n + n];
+    double* restrict rwork = (double*)(&buffer[2*n*n + n]);
     SCIPY_Z* restrict work = &buffer[2*n*n + 2*n];
 
     for (npy_intp idx = 0; idx < outer_size; idx++) {
--- a/scipy/linalg/tests/test_matfuncs.py
+++ b/scipy/linalg/tests/test_matfuncs.py
@@ -454,6 +454,13 @@ class TestSqrtM:
         np.fill_diagonal(M, 1)
         assert np.isrealobj(sqrtm(M))
 
+    def test_gh23278(self):
+        M = np.array([[1., 0., 0.], [0, 1, -1j], [0, 1j, 2]])
+        sq = sqrtm(M)
+        assert_allclose(sq @ sq, M, atol=1e-14)
+        sq = sqrtm(M.astype(np.complex64))
+        assert_allclose(sq @ sq, M, atol=1e-6)
+
     def test_data_size_preservation_uint_in_float_out(self):
         M = np.eye(10, dtype=np.uint8)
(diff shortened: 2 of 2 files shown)


lucascolley:
would you be able to point to the original lines here pre-translation? Or is this new code? (so that I can review)

ilayn:
I think I didn't parse the question correctly. Which lines are you asking about? This is modifying two lines without adding removing any.

lucascolley:
Sorry, let me try to form a proper sentence.

Was this bug a translation error? If so, could you point to the pre-translation lines for comparison?

Or is this in new code that was introduced alongside the translation?

ilayn:
I wrote this from scratch for linalg without any translation. Maybe you have some other code in mind? The issue here is my C mistake but not an algorithm issue.

lucascolley:
Ah apologies, I didn't realise it wasn't a translation. Do I understand correctly:

- buggy version: cast buffer to `float*` -> index -> get address
- fixed version: index -> get address -> cast pointer to `float*`

> overwrite the original array

Is this because we are casting a complex buffer to float in the buggy version?

tylerjereddy (approved):
I did a bisection (wasn't sure Ilan was working on it already) and found this as first bad commit locally on ARM Mac:

<details>

```
2bbd1939756296018715a11eef72847f154243ff is the first bad commit
commit 2bbd1939756296018715a11eef72847f154243ff
Author: Ilhan Polat <ilhanpolat@gmail.com>
Date:   Tue Jan 28 20:25:01 2025 +0100

    MAINT: linalg: move sqrtm function next to other matfuncs
```

</details> 

Does that make sense? Anyway,  when I check locally the test does indeed fail before/pass after the source patch and the test also seems to match the original reproducer quite well.

The CI is also happy. It looks like Lucas is more ambitious than I and actually trying to understand the change :) In any case, seems sensible to me.

ilayn:
> Do I understand correctly:

Yes. Instead of

 - jumping `(2n*n+n)*complex128` bytes and casting the array to double from that point on, 
 - it first casts to double and jumps `(2n*n+n)*float64` many bytes which falls short and walks into somebody else's lawn.

lucascolley:
Got it. Thanks for the learning opportunity!

ilayn:
It's refreshing to have small PRs once in a while.

PR #23282, opened 2025-07-03
TST: differentiate: remove float16 test

mdhaber:
#### Reference issue
Closes gh-23248

We don't really need a `float16` test, so if it's causing problems, let's remove it.


dschmitz89 (approved):
Finite differencing with float16 is definitely yolo computing.

PR #23283, opened 2025-07-03
DOC: small typo in the doc of class Model

pierre-veron:
correction of an inconsistency in the order of the arguments to be taken by the fjacd and fjacb functions

correction of an inconsistency in the order of the arguments to be taken by the fjacd and fjacb functions


rkern (approved):
LGTM! Thank you!

Issue #21910, opened 2024-11-18
BUG: interpolate.griddata: NaNs inside convex hull (despite no NaNs or duplicates in data)


paulbrodersen:
### Describe your issue.

Hi, thanks for taking the time to look at this bug report. 

# Short summary

`scipy.interpolate.griddata` is returning NaNs with `method = 'linear'` and `method = 'cubic'`. By construction, the data points used in the interpolation correspond to the convex hull of the data. Also, there are no duplicate data points.

Potentially related to issue #12945. 

# Detailed description

I am trying to correct for uneven illumination in a series of microscopy images. The region of interest in these images is always in the center, such that the border of the images only contains background illumination. This allows me to extract the pixel intensities on the border of the image:

![Figure_1](https://github.com/user-attachments/assets/5fdebde5-9371-404e-9731-ae636475b85a)

I then compute smooth estimates of the pixel intensities for the left/right/top/bottom borders. 

![Figure_4](https://github.com/user-attachments/assets/fd1765ad-dd35-4c80-b636-bfe00305fb24)

I can then interpolate between the fitted border pixel intensities using `scipy.interpolate.griddata`. However, the result contains a number of NaNs (run code below with `fill_nans=False`). 

![Figure_5](https://github.com/user-attachments/assets/e021eeee-d597-4468-b50b-55275c1178c7)

I can fill in these NaNs by simply running the interpolation again (run code below with `fill_nans=True`) , however as the images are large, this is a slow process that I would like to avoid. 

![Figure_3](https://github.com/user-attachments/assets/ba5dbea0-3929-4c26-99a0-c735279c6234)

**Any insight why the output of the first interpolation contains NaN would be appreciated.** 

[Test image](https://drive.google.com/file/d/1ykZP1voNEMKdC79A3_JMjs1DbByJK8wK/view?usp=drive_link)

### Reproducing Code Example

```python
import numpy as np
import matplotlib.pyplot as plt

from skimage.io import imread
from scipy.optimize import curve_fit
from scipy.interpolate import griddata


def correct_uneven_illumination(image, border_width=50, show=False, fill_nans=True):

    # Extract border pixel intensities
    border_pixel_intensities = dict(
        Left   = image[:, :border_width].mean(axis=1),
        Right  = image[:, -border_width:].mean(axis=1),
        Top    = image[:border_width, :].mean(axis=0),
        Bottom = image[-border_width:, :].mean(axis=0)
    )

    if show:
        image_with_border = image.copy()
[...]

PR #23285, opened 2025-07-04
DOC: interpolate.LinearNDInterpolator: mention limitations

DWesl:
#### Reference issue

This documents the issues uncovered in #21910.

This documents that linear interpolation is not intended to fill the interior of a very large array from the boundary.

#### Additional information

There's probably a dozen points with the given setup, starting around 150

### DIFF
--- a/scipy/interpolate/_interpnd.pyx
+++ b/scipy/interpolate/_interpnd.pyx
@@ -258,6 +258,15 @@ class LinearNDInterpolator(NDInterpolatorBase):
     with Qhull [1]_, and on each triangle performing linear
     barycentric interpolation.
 
+    If you provide points only on the boundary of a shape rather
+    than the interior, the default tolerance will cause issues
+    assigning a simplex to certain points in the interior with
+    enough points on each side, resulting in missing values in the
+    output.  This is easiest to reproduce by providing 300
+    evenly-spaced points on the positive x and y axes and
+    requesting points to fill the grid: all points with `x+y<300`
+    should have values.
+
     .. note:: For data on a regular grid use `interpn` instead.
 
     Examples


ev-br (changes requested):
Thanks.

Two questions/comments:

1. ISTM it could be nicer to move this discussion to the tutorial, 
https://docs.scipy.org/doc/scipy/tutorial/interpolate/ND_unstructured.html#scattered-data-interpolation-griddata, and expand to add code corresponding to the suggested recipe; this way it's easier for a user to experiment. Can link to the tutorial  from the `Notes` section of the docstring (see the RegularGridInterpolator docstring for an example)

2. I'm a bit confused by the "default tolerance" comment. Is this something we could fix by e.g. exposing the tolerance to a user as an optional argument with a backwards-compatible default?

DWesl:
>     1. ISTM it could be nicer to move this discussion to the tutorial,
>        https://docs.scipy.org/doc/scipy/tutorial/interpolate/ND_unstructured.html#scattered-data-interpolation-griddata, and expand to add code corresponding to the suggested recipe; this way it's easier for a user to experiment. Can link to the tutorial  from the `Notes` section of the docstring (see the RegularGridInterpolator docstring for an example)

That's probably a good idea

>     2. I'm a bit confused by the "default tolerance" comment. Is this something we could fix by e.g. exposing the tolerance to a user as an optional argument with a backwards-compatible default?

I think I got the example from the other issue to avoid NaNs by increasing the tolerance and recompiling SciPy, so this would likely be possible

ev-br:
> I think I got the example from the other issue to avoid NaNs by increasing the tolerance and recompiling SciPy, so this would likely be possible

Then it's definitely worth trying to plumb though an optional argument. If that gets very ugly or if the meaning of the parameter is very difficult to explain, then let's fall back to a DOC patch.

DWesl:
> > I think I got the example from the other issue to avoid NaNs by increasing the tolerance and recompiling SciPy, so this would likely be possible
> 
> Then it's definitely worth trying to plumb though an optional argument. If that gets very ugly or if the meaning of the parameter is very difficult to explain, then let's fall back to a DOC patch.

Should the argument go in the constructor or the interpolation argument?  Should the other interpolators in [`_ndgriddata.pyx`](https://github.com/scipy/scipy/blob/main/scipy/interpolate/_ndgriddata.py) have the same modificaton?

ev-br:
This depends on where the check with the user-supplied tolerance occurs , I'd say. If it's in constructing the triangulation in `__init__`, then the python argument goes to the constructor; if it's in computing the interpolation in `__call__`, then it's where the extra argument goes. 
Same for other interpolators--- ideally those which use the check should get an argument (and a regression test).

ev-br:
Gentle ping @DWesl , is this something you plan/have bandwidth to work on?

DWesl:
Added #23737 with the optional argument

I got frustrated back in July, but the work from then was almost complete.  Is the test there what you had in mind?

lucascolley:
superseded by gh-23951

PR #23287, opened 2025-07-05
MAINT: differentiate: add `xp_capabilities`

j-bowhay:
https://github.com/scipy/scipy/pull/23081#issuecomment-3037388701

Records the array api capabilites of the `differentiate` module

### DIFF
--- a/scipy/differentiate/_differentiate.py
+++ b/scipy/differentiate/_differentiate.py
@@ -3,7 +3,7 @@ import warnings
 import numpy as np
 import scipy._lib._elementwise_iterative_method as eim
 from scipy._lib._util import _RichResult
-from scipy._lib._array_api import array_namespace, xp_copy, xp_promote
+from scipy._lib._array_api import array_namespace, xp_copy, xp_promote, xp_capabilities
 import scipy._lib.array_api_extra as xpx
 
 _EERRORINCREASE = -1  # used in derivative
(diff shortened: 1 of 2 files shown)


lucascolley (approved):
thanks Jake!

lucascolley:
looks like some bits are not JAX JIT-compatible yet

j-bowhay:
seems odd I didn't remove any jax skips

lucascolley:
If I understand correctly, activating JIT in tests requires calling both:
1. `xpx.testing.lazy_xp_function`
2. `xpx.testing.patch_lazy_xp_functions`

The `xp` fixture calls (2) automatically, but on `main` there is nothing calling (1) for the `differentiate` tests.

This PR starts using `make_xp_test_case`, which calls (1) at https://github.com/scipy/scipy/blob/main/scipy/_lib/_array_api.py#L771.

Please let us know if I am wrong @crusaderky !

---

So, I think this PR should either include `jax_jit=False` in the appropriate `xp_capabilities` calls, or fix the implementation under JIT.

crusaderky:
> If I understand correctly, activating JIT in tests requires calling both:
> 
> 1. `xpx.testing.lazy_xp_function`
> 2. `xpx.testing.patch_lazy_xp_functions`
> 
> The `xp` fixture calls (2) automatically, but on `main` there is nothing calling (1) for the `differentiate` tests.
> 
> This PR starts using `make_xp_test_case`, which calls (1) at https://github.com/scipy/scipy/blob/main/scipy/_lib/_array_api.py#L771.
> 
> Please let us know if I am wrong @crusaderky !

`@xp_capabilities` + `@make_xp_test_case` internally call `xpx.testing.lazy_xp_function` and should be preferred to it.
~Your test module must declare `lazy_xp_modules = [<namespace where @xp_capabilities was decclared>]`.~
[EDIT] In this case you're importing the tested function directly into the namespace of the test module, so there is no need for `lazy_xp_modules`.

PR #23291, opened 2025-07-07
BENCH: sparse: parameterize benchmark with sparse arrays and matrices

dschult:
This PR starts the shift from spmatrix to sparray in our benchmark suite.

This first step copies the `sparse.py` benchmarks to `sparse_array.py` and changes all spmatrix usage to sparray. My naive checks suggest that any difference in classes is within variation from run to run. But I'd like someone with more benchmark experience to take a look.

Is this the way we should compare speeds between the classes? Or should we change the benchmarks to sparray without duplicating (so `asv` can plot the change in benchmark performance at the shift in classes)?


rgommers:
> Or should we change the benchmarks to sparray without duplicating

It'd be much better to parametrize the existing benchmarks over array/matrix I think. That will yield both easier to maintain code and direct side-by-side comparison in the output for array and matrix classes.

j-bowhay (approved):
All looks reasonable to me

PR #21620, opened 2024-09-24
ENH: estimate condition number for sparse matrices

maxaehle:
#### Reference issue
Closes gh-18969

#### What does this implement/fix?
Add
- `scipy.sparse.linalg.SuperLU.rinvnormest` to compute reciprocal norm of inverse matrix given an LU decomposition,
- `scipy.sparse.linalg.rinvnormest` to compute LU decomposition and call the function mentioned previously,
- `scipy.sparse.linalg.cond1est` to compute the 1-norm condition number, calling the function mentioned previously  and `scipy.sparse.linalg.onenormest`

PR #23292, opened 2025-07-07
ENH: sparse.linalg: condition number for sparse matrices

### LINKED ISSUE #18969: ENH: Estimate condition number in SuperLU

As far as I can tell, there is no good way to estimate the condition number of sparse matrices in scipy.

SuperLU provides a function `dgscon()` (https://github.com/scipy/scipy/blob/main/scipy/sparse/linalg/_dsolve/SuperLU/SRC/dgscon.c) that estimates the reciprocal of the condition number given a SuperLU factorization. I would like to request that a wrapper for `dgscon` be added as an instance method on `scipy.sparse.linalg.SuperLU`, for example 
```python
scipy.sparse.linalg.SuperLU.estimate_rcond(A_norm: float, norm=np.inf [or 1]) -> float
```
where `A_norm` is the infinity-norm [or 1-norm] of the matrix `A` that was factorized.

MATLAB has a function for this, called [condest](https://www.mathworks.com/help/matlab/ref/condest.html). The [pyAMG](https://github.com/pyamg/pyamg/tree/main) package has a function that estimates the condition number for sparse matrices (https://github.com/pyamg/pyamg/blob/main/pyamg/util/linalg.py#L388-L437), but given that the functionality is built in to SuperLU, which is vendored in scipy, I think this would be a straightforward and valuable addition to scipy.

_No response_

broesler:
#### Reference issue
Closes https://github.com/scipy/scipy/issues/18969
Builds on PR #21620, which is currently open.

This PR is a continuation of the work done in PR #21620, which introduced the
`scipy.sparse.linalg.cond1est` function, and the
`scipy.sparse.linalg.rinvnormest` function.

Changes herein:
- Retain `cond1est`
- Remove `rinvnormest` from `scipy.sparse.linalg`
- Refactor `rinvnormest` to `normest_inv` in `scipy.sparse.linalg.SuperLU`

#### Additional information
The `cond1est` function estimates the condition number of a sparse matrix in the
1-norm, which is defined as the product of the 1-norm of the matrix, and the
1-norm of its inverse. This PR slightly changes the internal functionality of
`cond1est` by computing the exact 1-norm of the matrix with
`sparse.linalg.norm`, as opposed to the estimate `sparse.linalg.onenormest`.
This PR also adds input checking to `cond1est`, and updates the docstring.

This PR removes the `scipy.sparse.linalg.rinvnormest` function from
`scipy.sparse.linalg`, but retains the functionality in the
`scipy.sparse.linalg.SuperLU` class. There is no analogous function in either
numpy or scipy for dense matrices, so it is unnecessary to expose it at the top
level of `scipy.sparse.linalg`.

This PR renames the `SuperLU.rinvnormest` method to `normest_inv`, which better
reflects the mathematical syntax ``norm(inv(A))``. The method now returns the
norm itself rather than its reciprocal. The API has also been updated to accept
either the integer `1` or float `np.inf` as the norm order, as opposed to
a string argument.

Finally, this PR adds extensive tests for the `normest_inv` and `cond1est` for
sparse matrices of various sizes, densities, and data types. The tests also
include checks for the correct return types and values, as well as error
handling for invalid inputs.

There is an additional script in
`scipy/sparse/linalg/_dev_scripts/check_normcond_behavior.py` that demonstrates
the existing behavior of `np.linalg.norm`, `np.linalg.cond`,
`scipy.linalg.norm`, and `scipy.sparse.linalg.onenormest`.


ilayn:
This is really impressive work but before we go further, note the discussion on #21620 until the end. 

We don't really need the norm of the inverse but we need a condition number which is theoretically defined as $\kappa$ but typically it is not derived using separate two norm estimate products. That's why we were discussing `lacon2` calls in the end and that's still need to be done.

broesler:
Ok I was going to ask if that is still under consideration.

I looked into the SuperLU function calls, and I think the best way to achieve that goal is actually to use the [langs](https://portal.nersc.gov/project/sparse/superlu/superlu_code_html/dlangs_8c.html) function to compute the exact 1-norm of the original `A` (in C), then calling `gscon` on L and U to estimate the condition number. The referenced Hager and Higham papers both discuss computing the 1-norm of A exactly, and estimating the 1-norm of A inverse. If we use `lacon2` to estimate the 1-norm of A from L and U, we are essentially just multiplying L and U back together to get the original matrix, which is inefficient.

The caveat here is that the 1-norm computation in `langs` would have to be done during the *construction* of the SuperLU object, while we still have the original matrix (as opposed to just the L and U factors). That should be a cheap operation relative to the factorization, but I can add testing to make sure that is the case.

broesler:
Hi, just checking in on the status of this PR. Let me know if you need me to make any more changes before merging.

scipy/sparse/linalg/_dev_scripts/check_normcond_behavior.py
```diff
@@ -0,0 +1,295 @@
+#!/usr/bin/env python3
```
ilayn: This file needs to be converted to actual tests such that they always run in case they are not exact copies of the tests below. We don't keep additional development versions of the designs.

scipy/sparse/linalg/_dsolve/_superluobject.c
```diff
@@ -105,10 +106,159 @@ static PyObject *SuperLU_solve(SuperLUObject * self, PyObject * args,
...
+        } else {
+            PyErr_SetString(PyExc_ValueError, "ord must be 1 or np.inf");
+            return NULL;
+        }
+    } else if (PyFloat_Check(ord_obj)) {  /* check if it's a float (np.inf) */
+        double ord = PyFloat_AsDouble(ord_obj);
+        if (ord == -1.0 && PyErr_Occurred()) {
+            return NULL;  /* conversion error */
+        }
+        if (ord == NPY_INFINITY) {
```
ilayn: I think norm1est is OK with getting only the 1-norm so you don't need to perform this norm detection dance. Infinity norm can be obtained by the transpose of the sparse array anyways.

scipy/sparse/linalg/_dsolve/_superluobject.c
```diff
@@ -105,10 +106,159 @@ static PyObject *SuperLU_solve(SuperLUObject * self, PyObject * args,
...
+        } else {
+            PyErr_SetString(PyExc_ValueError, "ord must be 1 or np.inf");
+            return NULL;
+        }
+    } else {
+        PyErr_SetString(PyExc_TypeError, "ord must be 1 or np.inf");
+        return NULL;
+    }
+
+    const bool return_float = (self->type == NPY_FLOAT || self->type == NPY_CFLOAT);
```
ilayn: Use dtype NPY_FLOAT32 and NPY_COMPLEX64 for explicit bit width specs

scipy/sparse/linalg/_dsolve/_superluobject.c
```diff
@@ -105,10 +106,159 @@ static PyObject *SuperLU_solve(SuperLUObject * self, PyObject * args,
...
+        goto fail;
+    }
+
+    /* Outputs from gscon */
+    volatile float rcondf;
+    volatile double rcondd;
+    volatile int info;
+
+    switch (self->type) {
+        case NPY_FLOAT:
```
ilayn: Explicit NumPy dtypes here too.

scipy/sparse/linalg/_dsolve/_superluobject.c
```diff
@@ -105,10 +106,159 @@ static PyObject *SuperLU_solve(SuperLUObject * self, PyObject * args,
...
+                (double *)&rcondd, (SuperLUStat_t *)&stat, (int *)&info);
+            break;
+        default: 
+            PyErr_SetString(PyExc_ValueError, "unsupported data type");
+            return NULL;
+    }
+
+    SLU_END_THREADS;
+
+    if (info) {
```
ilayn: needs an explicit check for `info < 0` in case something goes weird and somehow info comes back positive overflow or whatever the case might be because ... C.

scipy/sparse/linalg/_dsolve/_superluobject.c
```diff
@@ -721,6 +871,26 @@ PyObject *newSuperLUObject(SuperMatrix * A, PyObject * option_dict,
     self->py_csc_construct_func = NULL;
     self->type = intype;
 
+    /* Call the appropriate function to compute the 1-norm of the matrix */
```
ilayn: This has to be a lazy evaluation otherwise every instance will have to have the norm computation even if they won't need it. Hence you have to call a class method inside the `norm1est` to get the norm.

scipy/sparse/linalg/_dsolve/_superluobject.c
```diff
@@ -721,6 +871,26 @@ PyObject *newSuperLUObject(SuperMatrix * A, PyObject * option_dict,
...
+    char norm_c = '1';  /* default to 1-norm */
+    switch (self->type) {
+        case NPY_FLOAT:
```
ilayn: It is better to spell out the bit width for more explicit dtypes in case some architecture changes these, that is NPY_FLOAT64, NPY_COMPLEX128 etc. even though they are typically aliases.

scipy/sparse/linalg/_dsolve/_superluobject.h
```diff
@@ -25,6 +26,20 @@
 
 #define _CHECK_INTEGER(x) (PyArray_ISINTEGER((PyArrayObject*)x) && PyArray_ITEMSIZE((PyArrayObject*)x) == sizeof(int))
 
+
+/* 
+ * The norm of the matrix is determined by the type of the SuperLUObject.
+ * It is either a float or a double, depending on the type of the matrix.
+ * This union allows us to use the same variable for both types.
+ */
+typedef union _SuperLUAnorm_t {
```
ilayn: This is better to have a struct otherwise you are bit fiddling for 4 bytes. There is no need for such frugality here and for type punning.

ilayn (changes requested):
This looks quite good. Thanks @broesler . I've added some comments.

scipy/sparse/linalg/_dev_scripts/check_normcond_behavior.py
broesler: Removed the file in a54a74212. Relevant sections were already covered in the unit tests.

scipy/sparse/linalg/_dsolve/_superluobject.c
broesler: Removed the "ord" kwarg in 698a10101.

scipy/sparse/linalg/_dsolve/_superluobject.c
broesler: Explicit dtypes used in 47caff508.

scipy/sparse/linalg/_dsolve/_superluobject.c
broesler: Explicit dtypes used in 47caff508.

scipy/sparse/linalg/_dsolve/_superluobject.c
broesler: Explicit dtypes used in 47caff508.

scipy/sparse/linalg/_dsolve/_superluobject.h
broesler: struct used in 4497567c0.

scipy/sparse/linalg/_dsolve/_superluobject.c
broesler: Addressed in e8f997001.

scipy/sparse/linalg/_dsolve/_superluobject.c
broesler: I understand the need for lazy evaluation here. The challenge is: how do we compute the norm of the original matrix *A* after creation of the SuperLU object, since *A* is not stored in the object? If we use the SuperLU `lacon2` function (as discussed in #21620) to compute $\lVert A \rVert_1$ within the `SuperLU_cond1est` C function, we would only have access to the factors *L* and *U*, so we would be multiplying them back together to compute the norm. This operation is potentially quite expensive if *L* and *U* have a lot of fill-in relative to the sparsity of *A*.

The other option is to revert back to an earlier version of this PR (b3c80ed) where we compute only the estimate of $\lVert A^{-1} \rVert_1$ in the SuperLU object, and rely on the existing `scipy.sparse.linalg.norm(A, ord=1)` to compute $\lVert A \rVert_1$ in the Python `scipy.sparse.linalg.cond1est` function. The only Python computation that would happen in that case is the final multiplication of two floats:

$$ \kappa \equiv \lVert A \rVert_1 \lVert A^{-1} \rVert_1 $$

We can make the `normest_inv` function "private" to SuperLU, since there is no equivalent public function in numpy or scipy for dense matrices.

@dschmitz89 do you have any input following from the discussion in #21620?

broesler: For reference, I benchmarked the code on my local laptop. Here is a comparison of the addition of the norm computation into the SuperLU object creation:

```
% asv compare main feature_cond1est

All benchmarks:

| Change   | Before [601dc92c] <main>   | After [7572f59e] <feature_cond1est>   |   Ratio | Benchmark (Parameter)                                          |
|----------|----------------------------|--------------------------------------------|---------|----------------------------------------------------------------|
|          | 285M                       | 286M                                       |    1    | sparse_linalg_cond1est.Norm1EstBench.peakmem_splu(5000, 0.001) |
|          | 501M                       | 499M                                       |    1    | sparse_linalg_cond1est.Norm1EstBench.peakmem_splu(5000, 0.1)   |
|          | 62.5M                      | 62.8M                                      |    1    | sparse_linalg_cond1est.Norm1EstBench.peakmem_splu(5000, 1e-05) |
|          | 2.69±0.02s                 | 2.70±0.02s                                 |    1    | sparse_linalg_cond1est.Norm1EstBench.time_splu(5000, 0.001)    |
|          | 7.95±0.01s                 | 8.00±0.04s                                 |    1.01 | sparse_linalg_cond1est.Norm1EstBench.time_splu(5000, 0.1)      |
|          | 313±10μs                   | 318±2μs                                    |    1.02 | sparse_linalg_cond1est.Norm1EstBench.time_splu(5000, 1e-05)    |
```

The norm computation adds 1-2% additional time in constructing the SuperLU object for a 5000 x 5000 matrix. 

Here is a comparison of the `cond1est` computation time using the C computation of the norm vs. a `sparse.linalg.norm` call:

```
% asv compare feature_cond1est feature_cond1est_pynorm

All benchmarks:

| Change   | Before [eec2b0c0] <feature_cond1est>   | After [18f3d2bf] <feature_cond1est_pynorm>   |   Ratio | Benchmark (Parameter)                                              |
|----------|----------------------------------------|----------------------------------------------|---------|--------------------------------------------------------------------|
|          | 296M                                   | 286M                                         |    0.97 | sparse_linalg_cond1est.Cond1EstBench.peakmem_cond1est(5000, 0.001) |
|          | 503M                                   | 520M                                         |    1.03 | sparse_linalg_cond1est.Cond1EstBench.peakmem_cond1est(5000, 0.1)   |
|          | 62.7M                                  | 62.5M                                        |    1    | sparse_linalg_cond1est.Cond1EstBench.peakmem_cond1est(5000, 1e-05) |
|          | 2.78±0.03s                             | 2.70±0s                                      |    0.97 | sparse_linalg_cond1est.Cond1EstBench.time_cond1est(5000, 0.001)    |
|          | 8.31±0.03s                             | 8.19±0.03s                                   |    0.99 | sparse_linalg_cond1est.Cond1EstBench.time_cond1est(5000, 0.1)      |
| +        | 459±4μs                                | 516±10μs                                     |    1.12 | sparse_linalg_cond1est.Cond1EstBench.time_cond1est(5000, 1e-05)    |
```

There is a bit of noise in the least dense matrix since there are only ~250 entries, but the denser matrices show that the methods are essentially equivalent in both time and memory usage. We expect this result since the norm computation is almost negligible vs. the LU factorization itself.

broesler: Comparison of the branches is here:

[commit comparison](https://github.com/broesler/scipy/compare/9e86866e61...4c28774fd7)

I will merge the `feature_cond1est_pynorm` version with the `feature_cond1est` branch in this PR if we would like to use that one, but I wanted to make sure the reviewers could see both without opening an additional PR.

(edit: updated the link to compare the commits instead of the branches)

broesler: I realize 5,000 x 5,000 isn't particularly large, but 10,000 x 10,000 timed out the `asv` run on my local machine (60s max). In running a simple test with a smaller random matrix, I noticed that `splu` is quite slow relative to the dense `lu`:

```python
>>> from scipy import sparse, linalg as la
>>> from scipy.sparse import linalg as spla
>>> N = 5000
>>> d = 0.001
>>> A = sparse.random_array((N, N), density=d, format='csc')
>>> A.setdiag(N)  # set diagonal large to avoid pivoting
>>> A
===
<Compressed Sparse Column sparse array of dtype 'float64'
        with 29997 stored elements and shape (5000, 5000)>
>>> %timeit la.lu(A.toarray())
609 ms ± 10.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
>>> %timeit spla.splu(A)
2.74 s ± 3.18 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

The sparse version is ~4.4x slower than the dense version.

This difference reverses using a structured matrix:

```python
>>> from scipy.sparse.linalg import LaplacianNd
>>> sqrtN = 71  # 71 * 71 = 5041 ~ 5000
>>> A = LaplacianNd((sqrtN, sqrtN)).tosparse().tocsc().astype(float)
>>> A
===
<Compressed Sparse Column sparse array of dtype 'float64'
        with 24921 stored elements and shape (5041, 5041)>
>>> %timeit la.lu(A.toarray())
3.71 s ± 1.36 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
>>> %timeit spla.splu(A)
34.3 ms ± 497 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Each of these matrices has essentially the same density, so I am not quite sure why there are such major runtime discrepancies.

If I update the benchmark to use a `LaplacianNd` matrix instead of a `random_array`, which allows me to use larger matrices, we get these results:

```
(scipy-dev) (feature_cond1est_pynorm *) benchmarks % asv compare feature_cond1est feature_cond1est_pynorm

All benchmarks:

| Change   | Before [9e86866e] <feature_cond1est>   | After [4c28774f] <feature_cond1est_pynorm>   |   Ratio | Benchmark (Parameter)                                                |
|----------|----------------------------------------|----------------------------------------------|---------|----------------------------------------------------------------------|
|          | 71.9M                                  | 71.4M                                        |    0.99 | sparse_linalg_cond1est.Cond1EstBench.peakmem_cond1est(np.int64(100)) |
|          | 58.8M                                  | 58.9M                                        |    1    | sparse_linalg_cond1est.Cond1EstBench.peakmem_cond1est(np.int64(22))  |
|          | 59.3M                                  | 60.1M                                        |    1.01 | sparse_linalg_cond1est.Cond1EstBench.peakmem_cond1est(np.int64(31))  |
|          | 60.6M                                  | 61.8M                                        |    1.02 | sparse_linalg_cond1est.Cond1EstBench.peakmem_cond1est(np.int64(44))  |
|          | 64.6M                                  | 64.7M                                        |    1    | sparse_linalg_cond1est.Cond1EstBench.peakmem_cond1est(np.int64(70))  |
|          | 15.2±0.05ms                            | 15.3±0.04ms                                  |    1.01 | sparse_linalg_cond1est.Cond1EstBench.time_cond1est(np.int64(100))    |
| +        | 342±10μs                               | 402±0.9μs                                    |    1.18 | sparse_linalg_cond1est.Cond1EstBench.time_cond1est(np.int64(22))     |
|          | 839±8μs                                | 888±8μs                                      |    1.06 | sparse_linalg_cond1est.Cond1EstBench.time_cond1est(np.int64(31))     |
|          | 2.12±0.01ms                            | 2.17±0.01ms                                  |    1.02 | sparse_linalg_cond1est.Cond1EstBench.time_cond1est(np.int64(44))     |
|          | 6.62±0.03ms                            | 6.69±0.01ms                                  |    1.01 | sparse_linalg_cond1est.Cond1EstBench.time_cond1est(np.int64(70))     |
```

The difference in computing the norm outside of the `SuperLU` object becomes negligible for larger matrices.

dschmitz89:
The CI failures are related to the new test.

broesler:
Ah, the `strict=True` kwarg in `np.testing.assert_allclose` was introduced in numpy v2.0.0, so all of the failures are due to the test being run with an earlier version of numpy. I see that the `scipy/pyproject.toml` requires '"numpy>=2.0.0"', but do these CI runs just override that requirement?

dschmitz89:
@broesler: this is indeed confusing. Our toolchain roadmap [states](https://docs.scipy.org/doc/scipy/dev/toolchain.html#numpy) that "SciPy attempts to be compatible with at least the 4 previous releases of NumPy". I think what you find in `pyproject.toml` is meant for local development on scipy itself.

dschmitz89:
@broesler CI shows issues/crashes for Windows on ARM. Do you have the possibility to investigate? macOS/accelerate should be fixable by merging main.

broesler:
Agreed, the MacOS fix seems unrelated to this PR.

The Windows version seems to be failing in `np.linalg.cond`, which is the first line of the failing test. See lines 1078-1079 of the "Test" output in the Windows CI job:

```
 File "C:\hostedtoolcache\windows\Python\3.12.10\arm64\Lib\site-packages\numpy\linalg\_linalg.py", line 2035 in cond
  File "C:\a\scipy\scipy\build-install\Lib\site-packages\scipy\sparse\linalg\tests\test_cond1est.py", line 257 in test_random_nonsingular_matrix
```

so it is not even getting to the SuperLU calls. This issue might also be fixed by rebasing on main?

dschmitz89:
If there is a problem with numpy's `linalg.cond` on WoA, could you open an issue in the numpy repo?

broesler:
I opened a [numpy Issue](https://github.com/numpy/numpy/issues/30089).

dschmitz89:
Does the error in `np.linalg.cond` on WoA occur for all matrices or just pathological, near singular ones? In case of the latter, I think we can work around it in the test here by skipping it just for this platform. I do not think we have a pytest skip marker for WoA yet but should not be too difficult.

broesler:
The error actually occurred only on *non*-singular matrices in `test_random_nonsingular_matrix`. The smallest eigenvalues of the failing matrices are all of order ~ 10.

PR #23293, opened 2025-07-07
BUG: ndimage.median_filter: single element array handling

### LINKED ISSUE #23075: BUG: ndimage.median_filter: always returns `0` on single element arrays.
### Describe your issue.

After an upgrade to scipy `v1.15.3`, my `median_filter` is always returning `[0]` regardless of what single-element array I provide.

### Reproducing Code Example

```python
import numpy as np
import scipy
import sys
import platform

for samples in ([1, 2], [2]):
    sample_array = np.fromiter(samples, dtype=np.single)
    print(f"sample_array: {sample_array}")
    
    filtered_samples = scipy.ndimage.median_filter(sample_array, size=5, mode='reflect')
    print(f"filtered_samples: {filtered_samples}")
```

Output:

```python
sample_array: [1. 2.]
filtered_samples: [2. 1.]
sample_array: [2.]
filtered_samples: [0.]
```

Expected output:

```python
sample_array: [1. 2.]
filtered_samples: [2. 1.]
sample_array: [2.]
filtered_samples: [2.]
```

### Error message

```shell
N/A
```

```shell
SciPy version: 1.15.3
NumPy version: 2.0.2
Python version: 3.11.12
Machine: x86_64
```

tylerjereddy:
* Fixes gh-23075

* This patch adds special handling for the single-element array scenario in the above ticket, along with several regression test cases that get their expected values from running the same tests against SciPy `1.14.1`, before we overhauled the `median_filter` algorithm.

[skip circle]

### DIFF
--- a/scipy/ndimage/_filters.py
+++ b/scipy/ndimage/_filters.py
@@ -1989,7 +1989,8 @@ def _rank_filter(input, rank, size=None, footprint=None, output=None,
                 "A sequence of modes is not supported by non-separable rank "
                 "filters")
         mode = _ni_support._extend_mode_to_code(mode, is_filter=True)
-        if input.ndim == 1:
+        lim2 = input.size - ((footprint.size - 1) / 2 - origin)
+        if input.ndim == 1 and ((lim2 >= 0) or (input.size == 1)):
             if input.dtype in (np.int64, np.float64, np.float32):
                 x = input
                 x_out = output
--- a/scipy/ndimage/_filters.py
+++ b/scipy/ndimage/_filters.py
@@ -1989,7 +1989,7 @@ def _rank_filter(input, rank, size=None, footprint=None, output=None,
                 "A sequence of modes is not supported by non-separable rank "
                 "filters")
         mode = _ni_support._extend_mode_to_code(mode, is_filter=True)
-        lim2 = input.size - ((footprint.size - 1) / 2 - origin)
+        lim2 = input.size - ((footprint.size - 1) // 2 - origin)
         if input.ndim == 1 and ((lim2 >= 0) or (input.size == 1)):
             if input.dtype in (np.int64, np.float64, np.float32):
                 x = input
--- a/scipy/ndimage/_filters.py
+++ b/scipy/ndimage/_filters.py
@@ -1989,6 +1989,12 @@ def _rank_filter(input, rank, size=None, footprint=None, output=None,
                 "A sequence of modes is not supported by non-separable rank "
                 "filters")
         mode = _ni_support._extend_mode_to_code(mode, is_filter=True)
+        # Some corner cases are currently not allowed to use the
+        # "new"/fast 1D rank filter code, including when the
+        # footprint is large compared to the array size.
+        # See discussion in gh-23293; longer-term it may be possible
+        # to allow the fast path for these corner cases as well,
+        # if algorithmic fixes are found.
         lim2 = input.size - ((footprint.size - 1) // 2 - origin)
         if input.ndim == 1 and ((lim2 >= 0) or (input.size == 1)):
             if input.dtype in (np.int64, np.float64, np.float32):
--- a/scipy/ndimage/src/_rank_filter_1d.cpp
+++ b/scipy/ndimage/src/_rank_filter_1d.cpp
@@ -155,7 +155,24 @@ void _rank_filter(T *in_arr, int rank, int arr_len, int win_len, T *out_arr,
                   int mode, T cval, int origin) {
   int i, arr_len_thresh, lim = (win_len - 1) / 2 - origin;
   int lim2 = arr_len - lim;
-  if (lim2 < 0) return;
+  if (arr_len == 1) {
+      switch (mode) {
+          case REFLECT:
+          case NEAREST:
+          case WRAP:
+          case MIRROR:
+              out_arr[0] = in_arr[0];
+              return;
+          case CONSTANT:
+              if (win_len == 1) {
+                  out_arr[0] = in_arr[0];
+              }
+              else {
+                  out_arr[0] = cval;
+              }
+              return;
+      }
+  }
   int offset;
   Mediator *m = MediatorNew(win_len, rank);
   T *data = new T[win_len]();
--- a/scipy/ndimage/src/_rank_filter_1d.cpp
+++ b/scipy/ndimage/src/_rank_filter_1d.cpp
@@ -174,9 +174,6 @@ void _rank_filter(T *in_arr, int rank, int arr_len, int win_len, T *out_arr,
                   return;
           }
       }
-      else {
-          return;
-      }
   }
   int offset;
   Mediator *m = MediatorNew(win_len, rank);
--- a/scipy/ndimage/src/_rank_filter_1d.cpp
+++ b/scipy/ndimage/src/_rank_filter_1d.cpp
@@ -155,22 +155,27 @@ void _rank_filter(T *in_arr, int rank, int arr_len, int win_len, T *out_arr,
                   int mode, T cval, int origin) {
   int i, arr_len_thresh, lim = (win_len - 1) / 2 - origin;
   int lim2 = arr_len - lim;
-  if (arr_len == 1) {
-      switch (mode) {
-          case REFLECT:
-          case NEAREST:
-          case WRAP:
-          case MIRROR:
-              out_arr[0] = in_arr[0];
-              return;
-          case CONSTANT:
-              if (win_len == 1) {
+  if (lim2 < 0) {
+      if (arr_len == 1) {
+          switch (mode) {
+              case REFLECT:
+              case NEAREST:
+              case WRAP:
+              case MIRROR:
                   out_arr[0] = in_arr[0];
-              }
-              else {
-                  out_arr[0] = cval;
-              }
-              return;
+                  return;
+              case CONSTANT:
+                  if (win_len == 1) {
+                      out_arr[0] = in_arr[0];
+                  }
+                  else {
+                      out_arr[0] = cval;
+                  }
+                  return;
+          }
+      }
+      else {
+          return;
       }
   }
   int offset;
--- a/scipy/ndimage/src/_rank_filter_1d.cpp
+++ b/scipy/ndimage/src/_rank_filter_1d.cpp
@@ -155,7 +155,9 @@ void _rank_filter(T *in_arr, int rank, int arr_len, int win_len, T *out_arr,
                   int mode, T cval, int origin) {
   int i, arr_len_thresh, lim = (win_len - 1) / 2 - origin;
   int lim2 = arr_len - lim;
-  if (lim2 < 0) {
+  /* Note: `arr_len == 1` is the only case implemented here for `lim2 < 0`; the calling code */
+  /* in _filters.py ensures that this function isn't called otherwise. xref gh-23293 for details. */
+  if (lim2 < 0 && arr_len == 1) {
       switch (mode) {
           case REFLECT:
           case NEAREST:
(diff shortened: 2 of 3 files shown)


scipy/ndimage/src/_rank_filter_1d.cpp
```diff
@@ -155,7 +155,24 @@ void _rank_filter(T *in_arr, int rank, int arr_len, int win_len, T *out_arr,
...
+          case CONSTANT:
+              if (win_len == 1) {
+                  out_arr[0] = in_arr[0];
+              }
+              else {
+                  out_arr[0] = cval;
+              }
+              return;
+      }
+  }
```
tylerjereddy, line 177: @leventis @tqbl @aiudirog

This seems to do the trick--can you think of any problem scenarios/test cases where this approach doesn't hold?

tylerjereddy:
Interesting, I don't see the testing issues locally. We've seen lots of segfaulting problems with this code of course--that's why the original `return` was there.

scipy/ndimage/src/_rank_filter_1d.cpp
tqbl: Thanks @tylerjereddy. I just want to be sure we're not introducing another regression. The predicate `lim2 < 0` can be rewritten as `win_len > 2 * (arr_len + origin + 1)`, e.g. `origin=0`, `arr_len=8`, `win_len=19` would trigger it. Since the new check is only for single-element arrays, the other cases would be handled by the later parts of the code. Couldn't that potentially lead to the issues that were encountered before, like segfaults?

tylerjereddy: Good point, there are two problems showing up in the CI testing at the moment.

1. `assert_allclose()` only has the `strict` kwarg in NumPy `2.x` series--that is easily fixed by using the `xp..` version of that test, which preserves the ability to check shape and dtype correctness.
2. I'm able to (stochastically) reproduce the segfault showing up in CI if use NumPy `1.26.4`--seems to be caused in `scipy/ndimage/tests/test_filters.py::test_gh_22586_crash_property`, which is the exact test we wrote to prevent this regression (so that's good, I suppose, it is doing its job). I'll see if I can do a "partial fix" for now perhaps, for example by preserving the original conditional, but special casing the length-1 arrays that are the specific source of complaint here. I suspect this will still let some corner cases cause problems, but move us a bit closer to correctness.

scipy/ndimage/src/_rank_filter_1d.cpp
```diff
@@ -155,7 +155,29 @@ void _rank_filter(T *in_arr, int rank, int arr_len, int win_len, T *out_arr,
...
+                      out_arr[0] = in_arr[0];
+                  }
+                  else {
+                      out_arr[0] = cval;
+                  }
+                  return;
+          }
+      }
+      else {
+          return;
```
tylerjereddy: @tqbl This latest patch seems to prevent the crashing caught by the regression guard, `test_gh_22586_crash_property`, by restoring the original `lim2` conditional and only special casing the scenario that is the subject of the original ticket.

That said, I suspect this is a move in the right direction, but still not the full story since the case you describe at https://github.com/scipy/scipy/pull/23293/files#r2190998466 may very well still not return the correct answer (though it should no longer segfault).

Can I convince you that this might be "ok" to delay to another issue/ticket? Probably a few permutations of cases similar to the one you describe should be tested against SciPy `1.14.1`, and we should try to see if we can find a sensible algorithm that allows us to match that previous behavior without the "lazy" `return` here.

Or, we could try to make this PR a more complete fix, though that may take a fair bit more time.

tqbl: The problem is that prematurely returning is definitely not the right answer. For example, this will fail:
```python
sample_array = np.ones(8)
expected = np.ones(8)
filtered_samples = scipy.ndimage.median_filter(sample_array, size=19, mode="reflect")
assert np.isclose(filtered_samples, expected).all()
```

How about falling back to the nd filter implementation?
https://github.com/scipy/scipy/blob/v1.16.0/scipy/ndimage/_filters.py#L1990
```python
if input.ndim == 1:
```
becomes
```python
if input.ndim == 1 and (x.shape[0] == 1 or footprint.size <= 2 * (input.shape[0] + origin + 1)):
```
The `x.shape[0] == 1` check can even be dropped if there's no performance benefit.

tylerjereddy: Ok, I'll see what I can do. I believe my work here today is still a step in the right direction since we've reduced the number of scenarios where the premature `return` applies (i.e., the original ticket).

tylerjereddy: Ok, my fallback checks use a slightly different form, but the test case you provide and the others are passing in my hands now (locally, at least).

tylerjereddy:
And now the extra shim code isn't array API compatible... pretty annoying that it is that easy to break compliance with basic arithmetic. I'll check it tomorrow.

scipy/ndimage/src/_rank_filter_1d.cpp
tqbl: Thanks! Just one minor thing. `(footprint.size - 1) / 2` can be `(footprint.size - 1) // 2` to match the C++ code.

tylerjereddy:
Applying the suggested patch from @tqbl to perform truncating division on the Python side to match the C++ guards seems to conveniently also fix the array API compatibility issue in local testing.

We'll see if the CI is happy here now.

tylerjereddy:
@ev-br do we have someone for reviewing `ndimage` stuff? Maybe someone on the `skimage` team if we have to? I believe they call our `ndimage` in a few of their functions for example.

ev-br:
@tylerjereddy it's a bit of a problem that we are very thin on ndimage reviewers ATM. Maybe @grlee77 would have the bandwidth to take a look? (We're aware that you're very busy Greg!)

tylerjereddy:
@lagru could you take a look maybe? I think the Wikipedia entry for Median Filter is probably sufficient background to look this over (in fact, that's most of what I used to fix it, on top of checking the previous behavior).

tylerjereddy:
@rgommers as discussed in the call a few minutes ago, if you don't mind looking this over that would be great--for me, reading the first few paragraphs of the Wikipedia Median Filter page was sufficient to understand the improvement in behavior.

lagru:
Thanks for pinging me @tylerjereddy. I see what I can do time wise but no promises. :wink:

lagru:
PS: You can request a review from me. That would increase the chance that I'll get back to this.

scipy/ndimage/tests/test_filters.py
```diff
@@ -3036,3 +3036,55 @@ def test_gh23046_fix(self, xp):
...
+    ([2], "nearest", 100, [2]),
+    ([2], "wrap", 100, [2]),
+    ([2], "mirror", 100, [2]),
+    ([2], "constant", 100, [0]),
+])
+def test_gh_23075(samples, mode, size, expected):
+    # results verified against SciPy 1.14.1, before the median_filter
+    # overhaul
+    sample_array = np.asarray(samples, dtype=np.single)
+    expected = np.asarray(expected, dtype=np.single)
```
rgommers: nit: `np.float32` for these two lines please

scipy/ndimage/src/_rank_filter_1d.cpp
```diff
@@ -155,7 +155,26 @@ void _rank_filter(T *in_arr, int rank, int arr_len, int win_len, T *out_arr,
                   int mode, T cval, int origin) {
   int i, arr_len_thresh, lim = (win_len - 1) / 2 - origin;
   int lim2 = arr_len - lim;
-  if (lim2 < 0) return;
+  if (lim2 < 0) {
+      if (arr_len == 1) {
```
rgommers: There is no `else` clause here, which looks off. Is it impossible, missing, or did you intend for that case to fall through now to beyond line 177 in this diff, rather than hit `return;`?

scipy/ndimage/_filters.py
```diff
@@ -1989,7 +1989,8 @@ def _rank_filter(input, rank, size=None, footprint=None, output=None,
                 "A sequence of modes is not supported by non-separable rank "
                 "filters")
         mode = _ni_support._extend_mode_to_code(mode, is_filter=True)
-        if input.ndim == 1:
+        lim2 = input.size - ((footprint.size - 1) // 2 - origin)
+        if input.ndim == 1 and ((lim2 >= 0) or (input.size == 1)):
```
rgommers, line 1999: So the fix is that for the corner case of `footprint` being large compared to array size, we're skipping this branch and therefore hitting the old `_nd_image.rank_filter` C implementation rather than the new `_rank_filter_1d.rank_filter` C++ code?

I think that deserves a comment here, because future readers will wonder why the 1-D specific implementation isn't being used for `input.ndim == 1`.

rgommers (commented):
Thanks for tackling this @tylerjereddy & reviewers. 

The added test cases look good to me. A couple of questions about the implementation.

scipy/ndimage/_filters.py
tylerjereddy: I've added a comment similar to that, which also refers the reader back to this PR, and suggests that it may be possible to restore the fast path more broadly with algorithmic fixes.

That said, the original fast path author has clearly indicated they don't have the bandwidth to help us maintain it. Anyway, that's clearly why we're going through this. I don't know if it has been worth it to try to keep that fast path, and what percentage of real world cases have already been sent to the slower path as a result of working around problems, so things are getting a bit murky here, but the regression tests are the one thing that makes me think we're moving forward either way, because they were obviously missing before.

scipy/ndimage/src/_rank_filter_1d.cpp
tylerjereddy: I've purged the `arr_len` check and the full suite still passes locally.

That said, I had intended for the last thing you mention, I really only wanted to intercept the size-1 array case and the code inside the conditional was intended for the simple scenario that was failing in the original ticket. `_filters.py` also has the size-1 exception/allowance for the fast path.  But I can't prove with any testing that we need that extra guard anymore, so while that bothers me, I've removed it.

I did double check that we still need the broader `if (lim2 < 0)` guard--there are gruesome segfaults locally without that.

scipy/ndimage/tests/test_filters.py
tylerjereddy: done

scipy/ndimage/_filters.py
rgommers: The added comment LGTM. 



> I don't know if it has been worth it to try to keep that fast path,

Agreed it's not clear cut, but now that we have this fix let's roll with it.

scipy/ndimage/src/_rank_filter_1d.cpp
rgommers: The added code only works for `arr_len == 1`, so I don't think the change you just pushed is correct. How about adding that check back, and adding a code comment. Something like:
```c
    /* Note: `arr_len == 1` is the only case implemented here for `lim2 < 0`, the calling code */
    / * in _filters.py ensures that this function isn't called otherwise. xref gh-23293 for details. */
    if (lim2 < 0 && arr_len == 1) {
```

tylerjereddy: That's why I had it originally, but we can't prove a need for the guard with a public function test right? So this would just be a preference to have a private function behave more robustly on its own?

rgommers: Yes indeed. The code explicitly sets a single element, so we should just make it clear that that's what it's doing. It's confusing the way it is now.

tylerjereddy: alright, I pushed that up

rgommers (approved):
LGTM, seems good to merge and backport once CI comes back green.

PR #23299, opened 2025-07-08
MAINT/TST: switch to array_api_extra's xp_assert_* functions at several places

Saransh-cpp:
Switched in:

cluster.tests.test_hierarchy
signal.tests.test_savitzky_golay
signal.tests.test_spectral
signal.tests.test_spline
signal.tests.test_upfirdn
stats.tests.test_marray
stats.tests.test_rank

#### Additional information
cc: @lucascolley 😄

### DIFF
--- a/.github/workflows/pixi.toml
+++ b/.github/workflows/pixi.toml
@@ -31,6 +31,7 @@ pooch = "*"
 mpmath = "*"
 gmpy2 = "*"
 ccache = ">=4.10.1,<5"
+array_api_extra = "*"
 
 [feature.build.tasks]
 build = { cmd = "spin build -S-Dblas=blas -S-Dlapack=lapack -S-Duse-g77-abi=true", cwd = "../..", env = { CC = "ccache $CC", CXX = "ccache $CXX", FC = "ccache $FC" } }
--- a/.github/workflows/pixi.toml
+++ b/.github/workflows/pixi.toml
@@ -31,7 +31,6 @@ pooch = "*"
 mpmath = "*"
 gmpy2 = "*"
 ccache = ">=4.10.1,<5"
-array_api_extra = "*"
 
 [feature.build.tasks]
 build = { cmd = "spin build -S-Dblas=blas -S-Dlapack=lapack -S-Duse-g77-abi=true", cwd = "../..", env = { CC = "ccache $CC", CXX = "ccache $CXX", FC = "ccache $FC" } }
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,7 +75,6 @@ test = [
     "pytest-cov",
     "pytest-timeout",
     "pytest-xdist",
-    "array_api_extra",
     "asv",
     "mpmath",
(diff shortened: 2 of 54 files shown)


scipy/cluster/tests/test_hierarchy.py
```diff
@@ -443,21 +443,21 @@ def test_is_valid_linkage_various_size(self, nrow, ncol, valid, xp):
         Z = xp.asarray([[0, 1, 3.0, 2, 5],
                         [3, 2, 4.0, 3, 3]], dtype=xp.float64)
         Z = Z[:nrow, :ncol]
-        xp_assert_equal(is_valid_linkage(Z), valid, check_namespace=False)
+        assert is_valid_linkage(Z) == valid
```
Saransh-cpp, line 445: I am not sure why the tests were using `xp_assert_equal` here. I will find it out through the reviews or the failing CI 🙂

scipy/signal/tests/test_savitzky_golay.py
Saransh-cpp, line 1: All the functions tested in this file do return a numpy array, so equating the output explicitly with a numpy array should be fine IMO.

Saransh-cpp:
Now that I think of it, should the entire migration be done in one PR or multiple PRs? I'll mark this as draft for now.

requirements/test.txt
```diff
@@ -4,6 +4,7 @@ pytest
 pytest-cov
 pytest-timeout
 pytest-xdist
+array_api_extra
```
lucascolley: Please use the vendored version of `xpx` which exists in `scipy/_lib` rather than adding a new dependency. (We may need to update the submodule for new commits, not sure).

lucascolley (commented):
> should the entire migration be done in one PR or multiple PRs?

I think we should just wait and see how this turns out.

Note to other maintainers: this is primarily an exercise in battle-testing these functions from `xpx` before we make them public API. It can and should be merged eventually, but there is no rush.

.github/workflows/pixi.toml
```diff
@@ -31,6 +31,7 @@ pooch = "*"
 mpmath = "*"
 gmpy2 = "*"
 ccache = ">=4.10.1,<5"
+array_api_extra = "*"
```
Saransh-cpp:
```suggestion
```

pyproject.toml
```diff
@@ -75,6 +75,7 @@ test = [
     "pytest-cov",
     "pytest-timeout",
     "pytest-xdist",
+    "array_api_extra",
```
Saransh-cpp:
```suggestion
```

requirements/test.txt
Saransh-cpp:
```suggestion
```

Saransh-cpp:
Been a while since I worked on it, and unfortunately I will not be able to pick it up again any time in the near future :(

My work has diverged from array API as well :(

Closing this for now.

PR #23306, opened 2025-07-09
TST: free-threading performance tweaks

crusaderky:
Speed up a bunch of tests that take tens or hundreds of seconds to complete with `--parallel-threads=32`.

CC @ngoldbaum


tylerjereddy (approved):
This looks pretty sensible to me, thanks.

PR #23307, opened 2025-07-09
TST: free-threading: More local RNGs

crusaderky:
Replace a bunch of global RNGs with local ones. This fixes random failures when running 3.13t with `pytest-run-parallel`.

CC @ngoldbaum


rgommers (approved):
Thanks @crusaderky! All LGTM and CI is happy, so in it goes.

Issue #7695, opened 2017-08-01
BUG: special: ncfdtridfn and ncfdtridfd do not work reliably


WarrenWeckesser:
In `scipy.special`, `ncfdtridfn` and `ncfdtridfd` do not work reliably:

```
In [12]: from scipy.special import ncfdtr, ncfdtridfd, ncfdtridfn

In [13]: p = ncfdtr(5, 3, 1.5, 1.25)

In [14]: ncfdtridfn(p, 3, 1.5, 1.25)  # Should be 5
Out[14]: 1e-100

In [15]: p = ncfdtr(2, 2, 1.5, 1.25)

In [16]: ncfdtridfd(2, p, 1.5, 1.25)  # Should be 2
Out[16]: 1e+100
```

Version info:

```
In [42]: import sys

In [43]: sys.version_info
Out[43]: sys.version_info(major=3, minor=5, micro=2, releaselevel='final', serial=0)

In [44]: import numpy as np

In [45]: np.__version__
Out[45]: '1.12.1'

In [46]: import scipy

In [47]: scipy.__version__
Out[47]: '1.0.0.dev0+2b3ce2a'
```

WarrenWeckesser:
Note that, as pointed out in https://github.com/scipy/scipy/issues/7693, the docstrings for these functions show the arguments in the wrong order.  The problem in this issue was discovered while checking out that issue.

cdouglass:
Apparently `special` errors are ignored by default (from the `special` module docstring, since at least as far back as 2008: "Some of the special function routines can emit warnings or raise exceptions when an error occurs. By default this is disabled..."). I don't know the reasoning behind that and it definitely seems odd to me.

Would it make sense to change the default throughout `special` to warn instead of ignoring?

[...]

Issue #15633, opened 2022-02-21
missing fdtridfn


grlee77:
### Describe your issue.

According to `cdf_wrappers.c` and `_add_docs.py` there should be a function `fdtridfn` which is the inverse of `fdtri` with respect to `dfn`, but this function is not currently available.

This came up in #15622 where the docstrings to these function were edited, but the actual function is not present in the docs or importable in released versions of SciPy.

The cause seems to be because it is missing from `functions.json`, but I think that is [perhaps intentional based on this old comment](https://github.com/scipy/scipy/commits?author=teoliphant)](https://github.com/scipy/scipy/blob/33f24234af42fa484854e97489b222d4cb764ce4/scipy/special/cdf_wrappers.c#L222) by @teoliphant

I compiled it, but encountered unexpected behavior as mentioned here: https://github.com/scipy/scipy/pull/15622#discussion_r811440983 so perhaps there is a bug in the underlying Fortran library.

### SciPy/NumPy/Python version information

SciPy 1.9.0.dev0+1570.124e8a2

dschmitz89:
Since this function was never exposed by scipy and there has never been a user request for it, I am -1 on adding it. In principle, it is not super difficult but I do not find it important enough to be added. Once `ncfdtridfd` is ported to boost, it can be expected to work reasonably well for nc=0  which is exactly this case.

PR #23308, opened 2025-07-09
ENH: reimplement `special.ncfdtrinc` with boost's root finder

dschmitz89:
#### Reference issue
Towards #21966 and #19953

This implements `ncfdtrinc`, the inverse of the noncentral F distribution CDF with respect to the noncentrality parameter,  using boost's root finding machinery.

Boost contains a very convenient function called [bracket_and_solve_root](https://www.boost.org/doc/libs/latest/libs/math/doc/html/math_toolkit/roots_noderiv/bracket_solve.html) that automagically finds the root for us without having to set many parameters. Algorithmically, it is pretty much similar to xsf's root finder: it first brackets the root and then employes TOMS748 instead of Chandrupatla's algorithm. The big difference though is that it is templated (like pretty much everything in boost), so we get single and double precision out of the box. My implementation is mostly a copy of the example from boost's [documentation](https://www.boost.org/doc/libs/latest/libs/math/doc/html/math_toolkit/root_finding_examples/cbrt_eg.html) adapted to only template over real types.

**Benchmark results**
I carried out a benchmark based on the roundtrip `ncfdtrinc(ncfdtr(...))`, one with our own `ncfdtr` (upper image) using 10000 samples and once with an mpmath reference for `ncfdtr` (lower image) using 200 samples. The script was executed at an earlier commit when the new function was implemented as a private function (see details).
The new implementation outperforms cdflib by a large margin. This comes with a slight performance penalty: for the 10000 points used in the benchmark, the new implementation takes about 3 times as long as the old one. Performance tips would be welcome but do not need to block this PR.

<p>

```python
from time import perf_counter
import numpy as np
from mpmath import mp


# runs at commit fe5e38a42b044306507e8c33c383242765782a39

mp.dps = 100

NCFDTR_REFERENCE = "scipy"

np.vectorize
def ncfdtr_reference(dfn, dfd, nc, x):
    
    # Uses formula 26.2.20 from Abramowitz and Stegun.
    dfn, dfd, nc, x = map(mp.mpf, (dfn, dfd, nc, x))
    def term(j):
        result = mp.exp(-nc/2)*(nc/2)**j / mp.factorial(j)
        result *= mp.betainc(
            dfn/2 + j, dfd/2, 0, x*dfn/(x*dfn + dfd), regularized=True
        )
        return result
    try:
        result = mp.nsum(term, [0, mp.inf])
    except Exception as e:
        print(f"Error in nsum: {e}")
        return np.nan
    return float(result)

    """
    """
    mask = np.isnan(computed)
    reference = reference[~mask]
    computed = computed[~mask]
    return rel_error


rng = np.random.default_rng(12345673423235)

if NCFDTR_REFERENCE == "mpmath":
    number_of_samples = 200
else:
    number_of_samples = 10000

dfn = sc.exp10(rng.uniform(-4, 6, number_of_samples))
dfd = sc.exp10(rng.uniform(-5, 6, number_of_samples))
x = sc.exp10(rng.uniform(-10, 5, number_of_samples))
nc = sc.exp10(rng.uniform(-5, 4, number_of_samples))

if NCFDTR_REFERENCE == "mpmath":
    # Use mpmath for high precision reference values
    p = np.array([ncfdtr_reference(a, b, c, d) for a, b, c, d in zip(dfn, dfd, nc, x)])
else:
    p = sc.ncfdtr(dfn, dfd, nc, x)
mask = (p > 0) & (p < 1)

dfn_masked = dfn[mask]
dfd_masked = dfd[mask]
x_masked = x[mask]
nc_masked = nc[mask]
p_masked = p[mask]

t0 = perf_counter()
ncfdtrinc_old = sc.ncfdtrinc(dfn_masked, dfd_masked, p_masked, x_masked)
t1 = perf_counter()
print(f"cdflib ncfdtrinc took {t1 - t0:.4f} seconds")

t0 = perf_counter()
ncfdtrinc_new = sc._ufuncs._ncfdtrinc_new(dfn_masked, dfd_masked, p_masked, x_masked)
t1 = perf_counter()
print(f"boost ncfdtrinc took {t1 - t0:.4f} seconds")

rel_error_old = calculate_rel_error(ncfdtrinc_old, nc_masked)
rel_error_new = calculate_rel_error(ncfdtrinc_new, nc_masked)

n_bins = int(np.sqrt(len(nc_masked)))

if NCFDTR_REFERENCE == "mpmath":
    title = "Relative Error Distribution of roundtrip ncfdtrinc(ncfdtr_mpmath(...))"
else:
    title = "Relative Error Distribution of roundtrip ncfdtrinc(ncfdtr(...))"
plt.hist(np.log10(rel_error_old), bins=n_bins, alpha=0.5, label='cdflib', density=True, cumulative=True)
plt.hist(np.log10(rel_error_new), bins=n_bins, alpha=0.5, label='boost', density=True, cumulative=True)
plt.ylabel('Density')
plt.title(title)
plt.legend()
plt.grid()
plt.show()
``` 

</p>
</details> 

ncfdtrinc_benchmark

ncfdtrinc_benchmark_mpmath


scipy/special/boost_special_functions.h
```diff
@@ -1,11 +1,11 @@
 #ifndef BOOST_SPECIAL_FUNCTIONS_H
 #define BOOST_SPECIAL_FUNCTIONS_H
 
+#include <iostream>
```
dschmitz89:
```suggestion
```

Leftover from debugging.

scipy/special/boost_special_functions.h
```diff
@@ -1869,4 +1871,102 @@ landau_isf_double(double p, double loc, double scale)
...
+_ncfdtrinc(const Real dfn, const Real dfd, const Real p, const Real x)
+{
+    if (std::isnan(dfn) || std::isnan(dfd) || std::isnan(p) || std::isnan(x)) {
+        return NAN;
+    }
+    if (dfn <= 0 || dfd <= 0 || p < 0 || p > 1 || x < 0) {
+        sf_error("ncfdtrinc", SF_ERROR_DOMAIN, NULL);
+        return NAN;
+    }
+    Real guess = std::max(Real(1.0), dfn * dfd / (dfn + dfd)); // Crude initial guess.
```
dschmitz89: The guess was proposed by a LLM. Before, I had simply used 1. This line improved performance by a factor of 2 in my benchmark.

In principle, the expression can overflow but in that regime the noncentral CDF is anyway pretty much uncomputable. Open to other ideas for a starting guess if anyone has any.

dschmitz89:
CC @ev-br and @ilayn : this PR actually uses heavily templated C++, would be interested in your opinion since there has been some debate about this topic recently. As a rare user of C/C++, I found this quite pleasant to work with. This is likely due to the quality of the underlying Boost code,

ilayn:
No problem of C++ from my part since it is integrating with Boost anyways. It is a bit better on maintainability if it does not go into the "advanced" feature set of C++ otherwise all good. 

It's an eternal argument between proponents of C users with the C++ users that each thinks the other is using the wrong instrument. I would not pay too much attention to that. Both are terrible enough so nobody can ever win. Our debate was over something else and not the language selection.

scipy/special/boost_special_functions.h
mborland: This issue with overflowing in your guess would be that in Python you'll get an error for setting the overflow [`fenv`](https://en.cppreference.com/w/cpp/numeric/fenv/FE_exceptions.html) flag here, instead of a much more readable `throw` later down the line. Claude 4.1 Opus (Reasoning) recommended this change to avoid such a case:

```
Real guess {};
if (p < 0.5) {
    // For lower tail, start closer to 0
    guess = std::max(Real(0.1), dfn / (dfn + dfd));
} else if (p > 0.9) {
    // For upper tail, use larger initial value
    guess = std::max(Real(2.0), Real(2.0) * dfn / std::max(Real(1.0), dfd));
} else {
    // For middle probabilities, use ratio avoiding overflow
    Real ratio = dfn / (dfn + dfd);  // Always between 0 and 1
    guess = std::max(Real(1.0), ratio * dfd / std::max(Real(1.0), dfd - Real(2.0)));
}
```

scipy/special/boost_special_functions.h
```diff
@@ -1870,135 +1872,101 @@ landau_isf_double(double p, double loc, double scale)
...
 {
-    if (std::isnan(x) || std::isnan(dfn) || std::isnan(dfd)) {
-	return NAN;
+    if (std::isnan(dfn) || std::isnan(dfd) || std::isnan(p) || std::isnan(x)) {
+        return NAN;
     }
-    if ((dfn <= 0) || (dfd <= 0) || (x < 0) || (x > 1)) {
-	sf_error("fdtri", SF_ERROR_DOMAIN, NULL);
-	return NAN;
+    if (dfn <= 0 || dfd <= 0 || p < 0 || p > 1 || x < 0) {
```
mborland, line 2056: You should also probably check `std::isfinite` for `dfn` and `dfd` since +INF can be filtered out. You could also combine that with the above checks since that will catch NANs too.

scipy/special/boost_special_functions.h
```diff
@@ -1870,135 +1872,101 @@ landau_isf_double(double p, double loc, double scale)
...
+        sf_error("ncfdtrinc", SF_ERROR_DOMAIN, NULL);
+        return NAN;
     }
+    Real guess = std::max(Real(1.0), dfn * dfd / (dfn + dfd)); // Crude initial guess.
+    Real factor = 8;                                 // How big steps to take when searching.
+    const std::uintmax_t maxit = 500;            // Limit to maximum iterations.
```
mborland: We have a function that allows the policy to determine the maximum number of iterations:

`const std::uintmax_t maxit = boost::math::policies::max_root_iterations<SpecialPolicy>();`

scipy/special/boost_special_functions.h
```diff
@@ -1870,135 +1872,101 @@ landau_isf_double(double p, double loc, double scale)
...
+    std::pair<Real, Real> result_bracket;
     try {
-	y = boost::math::quantile<Real, SpecialPolicy>(
-                boost::math::fisher_f_distribution<Real, SpecialPolicy>(dfn, dfd), x);
-    } catch (const std::domain_error& e) {
-        sf_error("fdtri", SF_ERROR_DOMAIN, NULL);
+        result_bracket = boost::math::tools::bracket_and_solve_root(
+                            ncfdtrinc_target<Real>(x, dfn, dfd, p), guess, factor, is_rising, tol, it);
+    }
+    catch (const std::domain_error& e) {
```
mborland: You can do `catch (const std::domain_error&)` (no e) to avoid warnings in your builds.

mborland:
Does this need to run on GPU, since I see that those jobs are failing here? There's a few other changes that can be made using `boost::math` types and functions if that's needed.

scipy/special/boost_special_functions.h
dschmitz89: Claude's suggestion seems arbitrary to me but it avoids the overflow (as long as `dfn + dfd` does not).

scipy/special/boost_special_functions.h
dschmitz89: Yup, this was already done before in [this commit](https://github.com/scipy/scipy/pull/23308/commits/288ff7219c4b49d6d9f6f67487dcace04c30d1ef). In scipy, we typically commit without force pushing or rewriting history until reviewers are happy.

dschmitz89:
> Does this need to run on GPU, since I see that those jobs are failing here? There's a few other changes that can be made using `boost::math` types and functions if that's needed.

Boost functions are not running on GPUs yet in scipy (part of the long term roadmap though!). The failure is a CI/build problem that was fixed in main already from what I know.

scipy/special/boost_special_functions.h
dschmitz89: In main, `inf` is an acceptable input for this function. Although it will not give useful results likely, I kept it that way to avoid breaking backwards compatability.

scipy/special/boost_special_functions.h
dschmitz89: I decided against the complex guess and simply went with 10. My rationale is that for the vast majority of statistical applications, the noncentrality will be rather low.

dschmitz89:
@mborland I responded to your review comments. Let me know if anything else is required. On top of your review, the last commit removed the legacy cdflib implementation of this function that we are replacing here.

For visbility: In https://github.com/boostorg/math/issues/1305 the boost maintainers expressed that contributions of functions such as this one would also be welcome in boost. Boost also has GPU support and CI testing, so I think this would be the best way forward in the long run to avoid adding SciPy specific GPU cruft on top of boost. I need to familiarize myself with boost though before I can tackle that.

Once this PR is accepted, I can use the same approach to resolve #7695 and #15633 .

mborland (approved):
This looks good to me. If you do end up opening a PR to move this into boost I can always add commits to your PR for GPU support, so you don't have to worry about that piece.

dschmitz89:
This PR was approved by a boost maintainer and benchmarks show that it greatly outperforms cdflib in accuracy. Since we have not used boost's root finders before, I will wait until October or so before merging to give others time to have a look.

dschmitz89:
Closing as this will be implemented in boost itself.

PR #23309, opened 2025-07-10
Switch FFT library from pocketfft to ducc0

### LINKED ISSUE #15104: BUG: Possible memory leaks in signal.welch?
### Describe your issue.

When invoking welch on different sized data, it seems memory leaks are introduced even when the result of welch call is not stored in any variables and gc is collected afterwards. The data I feed into welch in the example below are just the original data with 1 byte removed (no copies), to ensure the input size is different. It also seems that after several cycles the memory use goes down a bit again (but again it should not be garbage collector since it is explicitly collected after each call).

I suspect there might beinternal caches created for distinct input sizes different, not sure if there is a way to clean these after each welch (https://github.com/scipy/scipy/issues/5986 suggests some methods to clean the cache using methods as destroy_zfft_cache etc, but I don't see these in the latest version of scipy anymore...).

OS: Windows 10

### Reproducing Code Example

```python
from scipy import signal
import numpy as np
import gc
import os, psutil

n = 60000000
data = np.random.randn(n)
process = psutil.Process(os.getpid())

for i in range(10):
    fft_size = n
    segment_size = int(fft_size / 2)
    overlap_fac = 0.2
    overlap_size = int(overlap_fac * segment_size)

    signal.welch(data, 8000, window='hann', nfft=fft_size,
                                return_onesided=True, detrend='constant', average='mean', nperseg=segment_size,
                                noverlap=overlap_size)
    gc.collect()

    print(f"Memory use loop {i}: {process.memory_info().rss} bytes")

    n -= 1
    data = np.delete(data, -1)
```

### Error message

```shell
Memory use loop 0: 1.557362 GB
Memory use loop 1: 2.302464 GB
Memory use loop 2: 8.255886 GB
Memory use loop 3: 12.674786 GB
Memory use loop 4: 12.141304 GB
Memory use loop 5: 7.263657 GB
Memory use loop 6: 13.386681 GB
Memory use loop 7: 5.997337 GB
Memory use loop 8: 12.672852 GB
Memory use loop 9: 7.488586 GB
```

1.7.3 1.21.4 sys.version_info(major=3, minor=9, micro=9, releaselevel='final', serial=0)

mreineck:
#### Reference issue
Closes https://github.com/scipy/scipy/issues/15104.

This switches the FFT library from `pocketfft` to its successor `ducc0`. The changes are not dramatic, but they include
- more compact plan storage (consumes `O(sqrt(N))` memory instead of `O(N)`)
- many incremental performance improvements

#### Additional information
There are a few things to be done to make this work:
- [ ] create a clone of https://gitlab.mpcdf.mpg.de/mtr/ducc in the Github `scipy` organization. Don't worry that the overall package is GPL licensed - everything related to FFT is also licensed under BSD3.
- [ ] compile a subset of the C++ files in there to a shared C++ library (I'll provide a list)
- [ ] adjust the calls within `scipy` itself (probably not a big amount of work, because the interface has not changed much).
- [ ] deal with situations that `ducc0` can't handle (mainly data arrays with strides that are not a multiple of the data type size).

@lucascolley, @peterbell10, I'm definitely going to need your help with this!


rkern:
This isn't [the way we work](https://scipy.github.io/devdocs/dev/contributor/development_workflow.html) and not how we use draft PRs. Please make a PR when you have functioning code that's ready to review. You can have conversations in an issue rather than a PR. If you need to collaborate before your branch is ready, be sure to do so on your own fork.

lucascolley:
Apologies @rkern, I said that this would be okay. Feel free to take over and direct the process if you have time to spare.

lucascolley:
@mreineck I am also happy to do this on your fork as Robert suggested

rkern:
Okay, yeah, I see that invitation now. @mreineck I apologize for the whipsaw of being told opposite things in quick succession. But I do think it would be better to use an issue for tracking work and discussion (regardless of where the work is done, this needs a separate issue made) and then a branch on one's own fork for collaboration until there's something reviewable. We don't want to be running all of our CI on every tentative commit in the meantime.

mreineck:
I'm fine with any approach! And I completely understand you want to avoid triggering CI whenever something happens on my branch.

Let's close this PR for the time being, and I'll open an issue, OK?

PR #23310, opened 2025-07-10
TST: stats: add test for CustomDistribution `__string__` & `__repr__`

hippowm:
This PR adds a test for the `__string__` & `__repr__` methods of `CustomDistribution` classes in the `stats.make_distribution` function. The test ensures that the string representations of the superclass `ContinuousDistribution` are correctly replaced with a custom string. PR https://github.com/scipy/scipy/pull/22371 added the methods without tests. This new test adds coverage to the following lines (that were modified in the PR^^):

```py
def _make_distribution_custom(dist):
    parameters = []
    support = getattr(dist, 'support')

    ...
    repr_str = dist.__class__.__name__

    class CustomDistribution(ContinuousDistribution):
        _parameterizations = ([_Parameterization(*parameters)] if parameters
                              else [])
        _variable = _x_param

        def __repr__(self):
            s = super().__repr__() #✅ NOW COVERED
            return s.replace('CustomDistribution', repr_str) #✅ NOW COVERED

        def __str__(self):
            s = super().__str__() #✅ NOW COVERED
            return s.replace('CustomDistribution', repr_str) #✅ NOW COVERED

    methods = {'sample', 'logentropy', 'entropy',
    ...
```

Note: Parts of this test have been automatically generated by a novel technique that we're currently developing as part of an academic research project aiming at improving test coverage. *To not waste developer time, two researchers manually checked the test before submitting it.* We appreciate the developers' time and any feedback is welcomed.

### DIFF
--- a/scipy/stats/tests/test_continuous.py
+++ b/scipy/stats/tests/test_continuous.py
@@ -1446,6 +1446,36 @@ class TestMakeDistribution:
             assert repr(dist(beta=2)) == "HalfGeneralizedNormal(beta=np.float64(2.0))"
         assert 'HalfGeneralizedNormal' in dist.__doc__
 
+    def test_custom_distribution_repr_str(self):
+            # Create a custom distribution class
+            class MyLogUniform:
+                @property
+                def __make_distribution_version__(self):
+                    return "1.16.0"
+
+                @property
+                def parameters(self):
+                    return {'a': {'endpoints': (0, np.inf), 
+                                  'inclusive': (False, False)},
+                            'b': {'endpoints': ('a', np.inf), 
+                                  'inclusive': (False, False)}}
+
+                @property
+                def support(self):
+                    return 'a', 'b'
+
+                def pdf(self, x, a, b):
+                    return 1 / (x * (np.log(b) - np.log(a)))
+
+            LogUniform = stats.make_distribution(MyLogUniform())
+            dist_instance = LogUniform(a=2.0, b=3.0)
+
+            # Test __str__ method
+            assert str(dist_instance) == 'MyLogUniform(a=2.0, b=3.0)'
+            # Test __repr__ method
+            assert repr(dist_instance) == \
+                'MyLogUniform(a=np.float64(2.0), b=np.float64(3.0))'
+
 
 class TestTransforms:
 
--- a/scipy/stats/tests/test_continuous.py
+++ b/scipy/stats/tests/test_continuous.py
@@ -1464,19 +1464,17 @@ class TestMakeDistribution:
                 def support(self):
                     return 'a', 'b'
 
-                def pdf(self, x, a, b):
-                    return 1 / (x * (np.log(b) - np.log(a)))
-
             LogUniform = stats.make_distribution(MyLogUniform())
             dist_instance = LogUniform(a=2.0, b=3.0)
 
             # Test __str__ method
             assert str(dist_instance) == 'MyLogUniform(a=2.0, b=3.0)'
             # Test __repr__ method
-            assert repr(dist_instance) in {
-                'MyLogUniform(a=np.float64(2.0), b=np.float64(3.0))',
-                'MyLogUniform(a=2.0, b=3.0)',
-            }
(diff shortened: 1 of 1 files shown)


scipy/stats/tests/test_continuous.py
```diff
@@ -1446,6 +1446,38 @@ def test_repr_str_docs(self):
...
+                                  'inclusive': (False, False)},
+                            'b': {'endpoints': ('a', np.inf), 
+                                  'inclusive': (False, False)}}
+
+                @property
+                def support(self):
+                    return 'a', 'b'
+
+                def pdf(self, x, a, b):
+                    return 1 / (x * (np.log(b) - np.log(a)))
```
tylerjereddy: For the purposes of this test, it looks like you can just purge out `pdf()` and it will still pass (see below). I suppose we may want to keep the test as simple as possible, even if defining `pdf()` is more realistic in many cases.

```diff
--- a/scipy/stats/tests/test_continuous.py
+++ b/scipy/stats/tests/test_continuous.py
@@ -1464,9 +1464,6 @@ class TestMakeDistribution:
                 def support(self):
                     return 'a', 'b'
 
-                def pdf(self, x, a, b):
-                    return 1 / (x * (np.log(b) - np.log(a)))
-

```

scipy/stats/tests/test_continuous.py
```diff
@@ -1446,6 +1446,38 @@ def test_repr_str_docs(self):
...
+            LogUniform = stats.make_distribution(MyLogUniform())
+            dist_instance = LogUniform(a=2.0, b=3.0)
+
+            # Test __str__ method
+            assert str(dist_instance) == 'MyLogUniform(a=2.0, b=3.0)'
+            # Test __repr__ method
+            assert repr(dist_instance) in {
+                'MyLogUniform(a=np.float64(2.0), b=np.float64(3.0))',
+                'MyLogUniform(a=2.0, b=3.0)',
+            }
```
tylerjereddy: pretty minor point, but above we have similar checks that are guarded on NumPy version, which is what determines which of these strings get matched, i.e., above there is:

```python
1445         if np.__version__ >= "2":
1446             assert repr(dist(beta=2)) == "HalfGeneralizedNormal(beta=np.float64(2.0))"
```

One advantage to that is that we can more easily identify which shims can be removed when the minimum supported NumPy is `2.0.0` (makes i.e., `git grep` chores easier), but it likely doesn't matter very much.

tylerjereddy (commented):
Since the CI is passing and you're just adding a new test, with no old tests changed, this is probably pretty safe.

I added some minor comments.

tylerjereddy:
We may want Albert or Matt to check it real quick, but the risk is very small with just adding a test.

mdhaber:
Is there a discussion somewhere about the tool used to generate this code? (Was it already decided that it's OK for SciPy?)

tylerjereddy:
good point, may be good to check on provenance, especially with the proliferation of license-dubious LLM-inspired solutions (though I've mostly tried to stay out of that discussion/debate)

lucascolley:
I can't imagine it would be significantly more objectionable than what is widely available to the public and has likely already been used in many PRs. I've been getting involved in the discussion on LLM policy and would like to see it move forward, but I don't think that should hold up PRs like this.

mdhaber:
I'll accept whatever consensus there is about this sort of thing. But personally, I'm not really keen on mixing AI generated code into this part of the project. This particular test is not very strong (or important, IMO), so I don't see much motivation to use the new distribution infrastructure as a testing ground for this sort of contribution. If it is really important to others that this test be added, I'd prefer to do it manually in one or two lines elsewhere.

hippowm:
Thank you for your opinions. We aim to contribute useful tests. When we manually review these tests, we occasionally run into tests that we are not sure of their worthiness (like Matt pointed out). In this case, we submit them and rely on the developers' judgement. So far our tests to scipy have seen mostly positive feedback.

j-bowhay:
Superseded by #23395

Issue #19848, opened 2024-01-10
`_lib._util.MapWrapper` uses multiprocessing with `fork`, deprecated in Python 3.12


rgommers:
Test failure from the new CI job with Python 3.12 I was just working on:

```
 ___________ TestDifferentialEvolutionSolver.test_immediate_updating ____________
[gw1] linux -- Python 3.12.1 /opt/hostedtoolcache/Python/3.12.1/x64/bin/python
scipy/optimize/tests/test__differential_evolution.py:703: in test_immediate_updating
    with warns(UserWarning):
E   DeprecationWarning: This process (pid=12452) is multi-threaded, use of fork() may lead to deadlocks in the child.
        bounds     = [(0.0, 2.0), (0.0, 2.0)]
        self       = <scipy.optimize.tests.test__differential_evolution.TestDifferentialEvolutionSolver object at 0x7f60117f4e30>
        solver     = <scipy.optimize._differentialevolution.DifferentialEvolutionSolver object at 0x7f5ffc3b0e30>

...

 FAILED scipy/optimize/tests/test__differential_evolution.py::TestDifferentialEvolutionSolver::test_immediate_updating - DeprecationWarning: This process (pid=12452) is multi-threaded, use of fork() may lead to deadlocks in the child.
```

This was inside a `pytest.warns` that was filtering another warning, so this is probably being surfaced now because of the combination of:
- the deprecation which is new in Python 3.12
- use of `pytest` 8.0, which bubbles up more warnings
- running the tests in parallel

Note that the warning isn't deterministic, because Python cannot reliably determine whether a program is using threading. I cannot reproduce it locally, but the cause is clear. See the notes on using the `fork` method with `multiprocessing` in https://docs.python.org/3/whatsnew/3.12.html.

For now I'll just avoid the warning, but I think the thing to do here is to switch the start method to `'spawn'` or `'forkserver'` (which one is TBD - the latter is faster, but not always available).

Cc @andyfaff

andyfaff:
The `DeprecationWarning` message isn't very clear, it should be pointing to info like the release notes for cp312 give. i.e. `fork` is deprecated in 3.12 (for linux), and will be moving to something else in 3.14. If you don't want to filter the deprecation warning in the test one you'll have to provide your own `workers`:

```python
ctx = multiprocessing.get_context('spawn')
with ctx.Pool() as p:
    with warns(UserWarning):
[...]

PR #23311, opened 2025-07-10
MAINT/TST: always use forkserver multiprocessing on POSIX

### LINKED ISSUE #23301: BUG: fft free-threading: test_multiprocessing segfaults
### Describe your issue.

`scipy.fft.tests.test_basic.py::test_multiprocess` segfaults consistently on my 32-core machine when I torture test it with `--parallel-threads=32`.

The segfault is visible in the pytest output.
However, even when it segfaults, on most occasions pytest returns that the test was successful and exits 0.
Occasionally, however, the test deadlocks.

### Reproducing Code Example

```python
$ pixi r test-nogil -t scipy.fft -- --parallel-threads=32 -v -k test_multiprocess

Collected 4 items to run in parallel

scipy/fft/tests/test_basic.py::test_multiprocess[numpy-fft] Fatal Python error: Segmentation fault

Current thread 0x00007cb556ffd6c0 (most recent call first):
  File "/home/crusaderky/github/pixi-dev-scipystack/scipy/.pixi/envs/free-threading/lib/python3.13t/multiprocessing/popen_fork.py", line 67 in _launch
  File "/home/crusaderky/github/pixi-dev-scipystack/scipy/.pixi/envs/free-threading/lib/python3.13t/multiprocessing/popen_fork.py", line 20 in __init__
  File "/home/crusaderky/github/pixi-dev-scipystack/scipy/.pixi/envs/free-threading/lib/python3.13t/multiprocessing/context.py", line 282 in _Popen
  File "/home/crusaderky/github/pixi-dev-scipystack/scipy/.pixi/envs/free-threading/lib/python3.13t/multiprocessing/process.py", line 121 in start
  File "/home/crusaderky/github/pixi-dev-scipystack/scipy/.pixi/envs/free-threading/lib/python3.13t/multiprocessing/pool.py", line 329 in _repopulate_pool_static
  File "/home/crusaderky/github/pixi-dev-scipystack/scipy/.pixi/envs/free-threading/lib/python3.13t/multiprocessing/pool.py", line 306 in _repopulate_pool
  File "/home/crusaderky/github/pixi-dev-scipystack/scipy/.pixi/envs/free-threading/lib/python3.13t/multiprocessing/pool.py", line 215 in __init__
  File "/home/crusaderky/github/pixi-dev-scipystack/scipy/.pixi/envs/free-threading/lib/python3.13t/multiprocessing/context.py", line 119 in Pool
  File "/home/crusaderky/github/pixi-dev-scipystack/scipy/scipy/build-nogil-install/usr/lib/python3.13t/site-packages/scipy/fft/tests/test_basic.py", line 455 in test_multiprocess
  File "/home/crusaderky/github/pixi-dev-scipystack/scipy/.pixi/envs/free-threading/lib/python3.13t/site-packages/pytest_run_parallel/plugin.py", line 99 in closure
  File "/home/crusaderky/github/pixi-dev-scipystack/scipy/.pixi/envs/free-threading/lib/python3.13t/threading.py", line 994 in run
  File "/home/crusaderky/github/pixi-dev-scipystack/scipy/.pixi/envs/free-threading/lib/python3.13t/threading.py", line 1043 in _bootstrap_inner
  File "/home/crusaderky/github/pixi-dev-scipystack/scipy/.pixi/envs/free-threading/lib/python3.13t/threading.py", line 1014 in _bootstrap

Extension modules: numpy._core._multiarray_umath, numpy.linalg._umath_linalg, cython.cimports.libc.math, scipy._lib._ccallback_c, scipy._lib._fpumode, numpy.random._common, numpy.random.bit_generator, numpy.random._bounded_integers, numpy.random._pcg64, numpy.random._mt19937, numpy.random._generator, numpy.random._philox, numpy.random._sfc64, numpy.random.mtrand, scipy._lib._uarray._uarray, scipy.special._ufuncs_cxx, scipy.special._ellip_harm_2, scipy.special._special_ufuncs, scipy.special._gufuncs, scipy.special._ufuncs, scipy.special._specfun, scipy.special._comb, scipy.linalg._fblas, scipy.linalg._flapack, _cyutility, scipy._cyutility, scipy.linalg.cython_lapack, scipy.linalg._cythonized_array_utils, scipy.linalg._solve_toeplitz, scipy.linalg._batched_linalg, scipy.linalg._decomp_lu_cython, scipy.linalg._matfuncs_schur_sqrtm, scipy.linalg._matfuncs_expm, scipy.linalg._linalg_pythran, scipy.linalg.cython_blas, scipy.linalg._decomp_update, scipy.sparse._sparsetools, _csparsetools, scipy.sparse._csparsetools, scipy.fftpack.convolve (total: 40)
PARALLEL PASSED                                                               [ 25%]
scipy/fft/tests/test_basic.py::test_multiprocess[numpy-ifft] PARALLEL PASSED                                                              [ 50%]
scipy/fft/tests/test_basic.py::test_multiprocess[numpy-rfft] PARALLEL PASSED                                                              [ 75%]
scipy/fft/tests/test_basic.py::test_multiprocess[numpy-irfft] PARALLEL PASSED                                                             [100%]

============================================================= slowest 50 durations ==============================================================
1.57s call     build-nogil-install/usr/lib/python3.13t/site-packages/scipy/fft/tests/test_basic.py::test_multiprocess[numpy-fft]
0.53s call     build-nogil-install/usr/lib/python3.13t/site-packages/scipy/fft/tests/test_basic.py::test_multiprocess[numpy-rfft]
0.52s call     build-nogil-install/usr/lib/python3.13t/site-packages/scipy/fft/tests/test_basic.py::test_multiprocess[numpy-irfft]
0.49s call     build-nogil-install/usr/lib/python3.13t/site-packages/scipy/fft/tests/test_basic.py::test_multiprocess[numpy-ifft]

(8 durations < 0.005s hidden.  Use -vv to show these durations.)
********************************************************** pytest-run-parallel report ***********************************************************
All tests were run in parallel! 🎉
====================================================== 4 passed, 5807 deselected in 5.03s =======================================================
```

### Error message

```shell
.
```

```shell
python 3.13.5t
pytest-run-parallel 0.4.4
numpy 2.3.0
scipy git tip
Linux x64 32-core host
```

ngoldbaum:
#### Reference issue
Closes https://github.com/scipy/scipy/issues/23301.

See https://github.com/scipy/scipy/issues/23301#issuecomment-3057645030 for details. I set `force=True` so if anyone adds a test that intentionally spawns in the future, we'll get an error.

### DIFF
--- a/scipy/conftest.py
+++ b/scipy/conftest.py
@@ -86,13 +86,12 @@ def pytest_configure(config):
             "markers",
             "iterations(n): run the given test function `n` times in each thread",
         )
-    else:
-        if int(config.getoption('--parallel-threads', '0')) > 0:
-            if os.name == 'posix' and sys.version_info < (3, 14):
-                # On POSIX, Python 3.13 and older uses the 'fork' context by
-                # default. Calling fork() from multiple threads leads to
-                # deadlocks. This has been changed in 3.14 to 'forkserver'.
-                multiprocessing.set_start_method('forkserver', force=True)
+
+    if os.name == 'posix' and sys.version_info < (3, 14):
+        # On POSIX, Python 3.13 and older uses the 'fork' context by
+        # default. Calling fork() from multiple threads leads to
+        # deadlocks. This has been changed in 3.14 to 'forkserver'.
+        multiprocessing.set_start_method('forkserver', force=True)
 
 
 def pytest_runtest_setup(item):
--- a/scipy/conftest.py
+++ b/scipy/conftest.py
@@ -2,6 +2,7 @@
 import json
 import multiprocessing
 import os
+import sys
 import warnings
 import tempfile
 from contextlib import contextmanager
@@ -87,8 +88,11 @@ def pytest_configure(config):
         )
     else:
         if int(config.getoption('--parallel-threads', '0')) > 0:
-            # forking from multiple threads leads to deadlocks
-            multiprocessing.set_start_method('spawn', force=True)
+            if os.name == 'posix' and sys.version_info < (3, 14):
+                # On POSIX, Python 3.13 and older uses the 'fork' context by
+                # default. Calling fork() from multiple threads leads to
+                # deadlocks. This has been changed in 3.14 to 'forkserver'.
+                multiprocessing.set_start_method('forkserver', force=True)
 
 
 def pytest_runtest_setup(item):
--- a/scipy/conftest.py
+++ b/scipy/conftest.py
@@ -1,5 +1,6 @@
 # Pytest customization
 import json
+import multiprocessing
 import os
 import warnings
 import tempfile
@@ -84,6 +85,10 @@ def pytest_configure(config):
             "markers",
             "iterations(n): run the given test function `n` times in each thread",
         )
+    else:
+        if int(config.getoption('--parallel-threads', '0')) > 0:
+            # forking from multiple threads leads to deadlocks
+            multiprocessing.set_start_method('spawn', force=True)
 
 
 def pytest_runtest_setup(item):
--- a/scipy/_lib/_util.py
+++ b/scipy/_lib/_util.py
@@ -1,3 +1,4 @@
+import os
 import re
 from contextlib import contextmanager
 import functools
@@ -617,18 +618,20 @@ class MapWrapper:
             self.pool = pool
             self._mapfunc = self.pool
         else:
-            from multiprocessing import Pool
+            from multiprocessing import get_context
+
             # user supplies a number
+            method = "forkserver" if os.name == "posix" else None
             if int(pool) == -1:
                 # use as many processors as possible
-                self.pool = Pool()
+                self.pool = get_context(method=method).Pool()
                 self._mapfunc = self.pool.map
                 self._own_pool = True
             elif int(pool) == 1:
                 pass
             elif int(pool) > 1:
                 # use the number of processors requested
-                self.pool = Pool(processes=int(pool))
+                self.pool = get_context(method=method).Pool(processes=int(pool))
                 self._mapfunc = self.pool.map
                 self._own_pool = True
             else:
--- a/scipy/_lib/_util.py
+++ b/scipy/_lib/_util.py
@@ -623,7 +623,7 @@ class MapWrapper:
 
             method = get_start_method(allow_none=True)
 
-            if method is None and os.name=='posix' and sys.version_info() < (3, 14):
+            if method is None and os.name=='posix' and sys.version_info < (3, 14):
                 # Python 3.13 and older used "fork" on posix, which can lead to
                 # deadlocks. This backports that fix to older Python versions.
                 method = 'forkserver'
--- a/scipy/_lib/_util.py
+++ b/scipy/_lib/_util.py
@@ -1,4 +1,3 @@
-import os
 import re
 from contextlib import contextmanager
 import functools
@@ -8,6 +7,8 @@ import numbers
 from collections import namedtuple
 import inspect
 import math
+import os
+import sys
 from types import ModuleType
 from typing import Literal, TypeAlias, TypeVar
 
@@ -618,10 +619,16 @@ class MapWrapper:
             self.pool = pool
             self._mapfunc = self.pool
         else:
-            from multiprocessing import get_context
+            from multiprocessing import get_context, get_start_method
+
+            method = get_start_method(allow_none=True)
+
+            if method is None and os.name=='posix' and sys.version_info() < (3, 14):
+                # Python 3.13 and older used "fork" on posix, which can lead to
+                # deadlocks. This backports that fix to older Python versions.
+                method = 'forkserver'
 
             # user supplies a number
-            method = "forkserver" if os.name == "posix" else None
             if int(pool) == -1:
                 # use as many processors as possible
                 self.pool = get_context(method=method).Pool()
--- a/scipy/optimize/tests/test__differential_evolution.py
+++ b/scipy/optimize/tests/test__differential_evolution.py
@@ -1,7 +1,6 @@
 """
 Unit tests for the differential global minimization algorithm.
 """
-import multiprocessing
 from multiprocessing.dummy import Pool as ThreadPool
 import platform
 import warnings
@@ -705,25 +704,18 @@ class TestDifferentialEvolutionSolver:
         )
         assert res.success
 
-    @pytest.mark.thread_unsafe
     def test_immediate_updating(self):
         # check setting of immediate updating, with default workers
         bounds = [(0., 2.), (0., 2.)]
         solver = DifferentialEvolutionSolver(rosen, bounds)
         assert_(solver._updating == 'immediate')
 
-        # Safely forking from a multithreaded process is
-        # problematic, and deprecated in Python 3.12, so
-        # we use a slower but portable alternative
-        # see gh-19848
-        ctx = multiprocessing.get_context("spawn")
-        with ctx.Pool(2) as p:
-            # should raise a UserWarning because the updating='immediate'
-            # is being overridden by the workers keyword
-            with warns(UserWarning):
-                with DifferentialEvolutionSolver(rosen, bounds, workers=p.map) as s:
-                    solver.solve()
-            assert s._updating == 'deferred'
+        # should raise a UserWarning because the updating='immediate'
+        # is being overridden by the workers keyword
+        with warns(UserWarning):
+            with DifferentialEvolutionSolver(rosen, bounds, workers=2) as s:
+                solver.solve()
+        assert s._updating == 'deferred'
 
     @pytest.mark.fail_slow(10)
     def test_parallel(self):


scipy/conftest.py
```diff
@@ -84,6 +85,10 @@ def pytest_configure(config):
             "markers",
             "iterations(n): run the given test function `n` times in each thread",
         )
+    else:
+        if int(config.getoption('--parallel-threads', '0')) > 0:
+            # forking from multiple threads leads to deadlocks
```
crusaderky:
```suggestion
        if (sys.version_info[:2] == (3, 13) 
            and int(config.getoption('--parallel-threads', '0')) > 0):
            # On POSIX, Python 3.13 uses the 'fork' context by default.
            # Calling fork() from multiple threads leads to deadlocks.
            # This has been changed in 3.14 to 'forkserver'.
```

scipy/conftest.py
```diff
@@ -1,5 +1,6 @@
 # Pytest customization
 import json
+import multiprocessing
 import os
```
crusaderky, line 4:
```suggestion
import multiprocessing
import os
import sys
```

crusaderky (changes requested):
Could you also clean up the ad-hoc fix in `scipy/optimize/tests/test__differential_evolution.py::test_immediate_updating`?

scipy/conftest.py
ngoldbaum: I don't think this is right, the default start method was "fork" in 3.12 and older as well.

IMO it's better to just unconditionally set this as I have it in this PR right now. I'll tweak the comment though.

ngoldbaum: Went ahead and made it only do something for Python < 3.13 and POSIX. Also made it use forkserver instead of spawn, since that's supposed to be faster.

crusaderky:
I've been running all tests that use multiprocessing.Pool directly or indirectly with forkserver on 3.13t with --parallel-threads=32 and it looks stable.

However, CI shows a genuine regression where it seems like the global change is not being picked up in all environments:

> FAILED ../optimize/tests/test__differential_evolution.py::TestDifferentialEvolutionSolver::test_immediate_updating - DeprecationWarning: This process (pid=13442) is multi-threaded, use of fork() may lead to deadlocks in the child.

scipy/conftest.py
```diff
@@ -84,6 +86,13 @@ def pytest_configure(config):
             "markers",
             "iterations(n): run the given test function `n` times in each thread",
         )
+    else:
+        if int(config.getoption('--parallel-threads', '0')) > 0:
+            if os.name == 'posix' and sys.version_info < (3, 14):
+                # On POSIX, Python 3.13 and older uses the 'fork' context by
+                # default. Calling fork() from multiple threads leads to
+                # deadlocks. This has been changed in 3.14 to 'forkserver'.
+                multiprocessing.set_start_method('forkserver', force=True)
```
crusaderky: In order to fix the regression this should run unconditionally to `PARALLEL_RUN_AVAILABLE`.

rgommers:
The change to this test came from the fix for gh-19848, where I commented: *For now I'll just avoid the warning, but I think the thing to do here is to switch the start method to 'spawn' or 'forkserver' (which one is TBD - the latter is faster, but not always available).*

I think that is still true. Moving only the tests away from `'fork'` but leaving the `MapWrapper` implementation defaulting to it feels inherently wrong, we're just no longer testing what the default behavior of the functionality is. We have to either change the functionality or add prominent warnings that this is up to the user. @andyfaff WDYT?

crusaderky:
> Moving only the tests away from `'fork'` but leaving the `MapWrapper` implementation defaulting to it feels inherently wrong, we're just no longer testing what the default behavior of the functionality is. We have to either change the functionality or add prominent warnings that this is up to the user. @andyfaff WDYT?

IMHO this is a Python issue (fixed in 3.14), not a scipy one. scipy is just a wrapper around `multiprocessing.Pool()`; I don't think it should tamper itself with the default context.
Maybe a note in the documentation would suffice?

I feel that this falls neatly in the much larger bucket of "3.13 is kinda, sorta, free-threading compatible if you feel adventurous; otherwise please wait for 3.14".

andyfaff:
`MapWrapper` itself uses the default start method. The actual test uses a `'spawn'` multiprocessing context. Is that incompatible with threading?

crusaderky:
> `MapWrapper` itself uses the default start method. The actual test uses a `'spawn'` multiprocessing context. Is that incompatible with threading?

`fork`, which is the default on Linux python<3.14, is incompatible with threading.
The test was previously changing the context to a safe one; this is now done centrally by conftest.

scipy/conftest.py
```diff
@@ -85,6 +87,12 @@ def pytest_configure(config):
             "iterations(n): run the given test function `n` times in each thread",
         )
 
+    if os.name == 'posix' and sys.version_info < (3, 14):
+        # On POSIX, Python 3.13 and older uses the 'fork' context by
+        # default. Calling fork() from multiple threads leads to
+        # deadlocks. This has been changed in 3.14 to 'forkserver'.
+        multiprocessing.set_start_method('forkserver', force=True)
```
andyfaff, line 94: why forkserver and not spawn?

andyfaff:
If spawn should be compatible with threading, then why was the test marked as threadunsafe?

ngoldbaum:
Because it catches warnings. The marker is now unnecessary because pytest-run-parallel detects that.

scipy/conftest.py
ngoldbaum: It's slightly faster according to the Python docs. This also makes things consistent on 3.13 and older with the behavior on 3.14 and going forward.

rgommers:
> scipy is just a wrapper around `multiprocessing.Pool()`; I don't think it should tamper itself with the default context.

It doesn't have to mess with global state, but the easiest way to use these APIs is for the user to add a keyword like `workers=8`. It's logically inconsistent to make that use `'fork'` but remove test coverage that that works as advertised. 

I'd think making the implementation use `multiprocessing.get_context(method='spawn').Pool` rather than `multiprocessing.Pool` when the user provides a plain integer is a much more consistent fix.

scipy/conftest.py
crusaderky: Because forkserver is faster, and also it is the default on 3.14. This PR is effectively backporting 3.14 policy to the unit tests running on 3.11~3.13.

crusaderky:
> I'd think making the implementation use `multiprocessing.get_context(method='spawn').Pool` rather than `multiprocessing.Pool` when the user provides a plain integer is a much more consistent fix.

Can we settle on forkserver, for the sake of consistency with multiprocessing 3.14?

rgommers:
> Can we settle on forkserver, for the sake of consistency with multiprocessing 3.14?

Yes, I didn't think about it too hard, I'd be fine with either. Picking whatever the platform defaults are on 3.14 seems like a good strategy.

ngoldbaum:
> I'd think making the implementation use multiprocessing.get_context(method='spawn').Pool rather than multiprocessing.Pool when the user provides a plain integer is a much more consistent fix.

Done in the latest commit, but with forksersever on POSIX, otherwise deferring to the CPython default.

Does SciPy use towncrier or otherwise is there a spot I should add a release note for the change? I don't see anything in the developer guide.

rgommers:
> Does SciPy use towncrier or otherwise is there a spot I should add a release note for the change? I don't see anything in the developer guide.

We use wiki pages. Only editable by maintainers, if you propose an edit here, I'll copy it over.

andyfaff:
We can't use the approach made in the last commit because it forces `'posix'` to using `'forkserver'` whenever a `MapWrapper` is used. i.e. not only in tests, but in production as well. It disrespects [`multiprocessing.set_start_method`](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.set_start_method), where the user may have already selected what they'd like to use for starting child processes.

rgommers:
@andyfaff is that a problem? I think there's two ways to use the public `workers` API in various places:
1. Pass in a `Pool` or a `Pool`-compatible object. Full control over start method here.
2. Pass in a plain integer. Here SciPy only promises to parallelize to the requested parallelism. The underlying implementation is free to do what it wants - and we even have a mix of `multiprocessing` and native threadpools to back this. 

I don't think there should be an expectation of (2) having specific semantics beyond running as efficiently as possible in parallel. Global state isn't a good way to pick a start method compared to using (1).

andyfaff:
> Pass in a Pool or a Pool-compatible object. Full control over start method here.

That's not quite True. The idea behind `'workers'` is that for full control one passes a `map-like` function/method, e.g. `map`, `Pool.map` (several processes), `ThreadPoolExecutor.map` (several threads), `ProcessPoolExecutor.map`, etc. In that vein, test code can provide a `ThreadPoolExecutor.map` if they wish the underlying parallelisation to use threads instead of processes.

For number 2, for me as an end-user I would be expecting library code that was using `multiprocessing.Pool` to respect `set_start_method`. I agree that using global state is not generally a good idea, but as I understand python generally that's the current landscape. This expectation might be biased however.

I think I'd be happy for:

```python
method = multiprocessing.get_start_method(allow_none=True)

if method is None and os.name=='posix' and sys.version_info() < (3, 14):
    method = 'forkserver'
```

rgommers:
+1 that's a nice improvement, `allow_none` is a very helpful keyword here.

scipy/conftest.py
ngoldbaum: IMO it's still worth keeping this despite the change to `MapWrapper`, since some tests use `multiprocessing.Pool` or `ProcessPoolExecutor` directly:

```
goldbaum at Nathans-MBP in ~/Documents/scipy/scipy on force-spawn-pytest-run-parallel!
± rg "from multiprocessing import Pool" **/tests/*
stats/tests/test_generation/studentized_range_mpmath_ref.py
26:from multiprocessing import Pool, cpu_count

optimize/tests/test__shgo.py
6:from multiprocessing import Pool

optimize/tests/test_least_squares.py
4:from multiprocessing import Pool

_lib/tests/test__util.py
1:from multiprocessing import Pool

goldbaum at Nathans-MBP in ~/Documents/scipy/scipy on force-spawn-pytest-run-parallel!
± rg "multiprocessing.Pool" **/tests/*
fft/tests/test_basic.py
454:    with multiprocessing.Pool(2) as p:

fft/tests/test_multithreading.py
38:    with multiprocessing.Pool(2) as p:
```

ngoldbaum:
Release note suggestion:

On POSIX operating systems, SciPy will use the forkserver multiprocessing context on Python 3.13 and older. This follows the default behavior on Python 3.14.

rgommers (approved):
LGTM, everyone's comments were addressed, and CI seems happy. So in it goes. Thanks @ngoldbaum, @crusaderky and @andyfaff!

rgommers:
Release note added to https://github.com/scipy/scipy/wiki/*-Release-Note-Entries-for-SciPy-1.17.0-(asterisk-makes-this-appear-at-the-top)

PR #23313, opened 2025-07-10
BLD: support build warnings for flang/clang-cl on Windows

rgommers:
This gets rid of a few thousand lines of build warnings on this Windows build config. There are some left, most prominently these:
```
flang-new: warning: argument unused during compilation: '-Wall' [-Wunused-command-line-argument]
```

The problem with the linker script flag was that `clang-cl` does accept it so the check passes, but then complains about not knowing it. This is caused by the more general design choice it made about accepting rather than erroring out on unknown flags.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -98,6 +98,19 @@ if ff.has_argument('-Wno-conversion')
   add_project_arguments('-Wno-conversion', language: 'fortran')
 endif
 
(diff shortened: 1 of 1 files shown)


lucascolley (approved):
Thanks Ralf! I noticed a lot of warnings too...

PR #23314, opened 2025-07-10
TST: odr: add more tests for odrpack

alteholz:
Add more tests to increase coverage.
The new tests provide strange parameters to some methods and check whether the expected exceptions or results appear.

This PR increases the test coverage of scipy/odr/_odrpack.py from 74% (94 lines missing) to 93% (27 lines missing).

#### Additional information
I ran "spin lint" and autopep8 and got no warnings for my additions. There are still other warnings from commits before, that I didn't want to change.

### DIFF
--- a/scipy/odr/tests/test_odr.py
+++ b/scipy/odr/tests/test_odr.py
@@ -1,7 +1,9 @@
+
 import pickle
 import tempfile
 import shutil
 import os
+import re
 
 import numpy as np
 from numpy import pi
@@ -13,7 +15,8 @@ from pytest import raises as assert_raises
 
 from scipy.odr import (Data, Model, ODR, RealData, OdrStop, OdrWarning,
                        multilinear, exponential, unilinear, quadratic,
-                       polynomial)
+                       polynomial, Output)
+from scipy.odr._odrpack import (_report_error, OdrError)
 
 
 class TestODR:
@@ -75,6 +78,12 @@ class TestODR:
         explicit_odr.set_job(deriv=2)
         explicit_odr.set_iprint(init=0, iter=0, final=0)
 
+        with pytest.raises(
+            OdrError,
+            match=re.escape("cannot restart: run() has not been called before")
+        ):
+            _ = explicit_odr.restart()
+
         out = explicit_odr.run()
         assert_array_almost_equal(
             out.beta,
@@ -94,6 +103,13 @@ class TestODR:
                [-8.0978217468468912e-04, -1.9453521827942002e-03,
                   1.6827336938454476e-05]]),
         )
+        assert_equal(explicit_mod.meta['name'], 'Sample Explicit Model')
+
+        out_restart = explicit_odr.restart()
+        assert_array_almost_equal(out.beta, out_restart.beta, decimal=4)
+        assert_array_almost_equal(out.sd_beta, out_restart.sd_beta, decimal=5)
+        assert_array_almost_equal(
+            out.cov_beta, out_restart.cov_beta, decimal=5)
 
     # Implicit Example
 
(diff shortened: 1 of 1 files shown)


scipy/odr/tests/test_odr.py
```diff
@@ -94,6 +103,13 @@ def test_explicit(self):
                [-8.0978217468468912e-04, -1.9453521827942002e-03,
                   1.6827336938454476e-05]]),
         )
+        assert_equal(explicit_mod.meta['name'], 'Sample Explicit Model')
+
+        out_restart = explicit_odr.restart()
+        assert_array_almost_equal(out.beta, out_restart.beta, decimal=4)
```
tylerjereddy, line 109: My usual comment that for new tests, we should probably prefer `assert_allclose()` per the NumPy docs for the function.

scipy/odr/tests/test_odr.py
```diff
@@ -605,3 +621,167 @@ def test_pickling_output(self):
...
+        problems = _report_error(41000)
+        assert_equal(problems, ['Error in derivatives'])
+
+        # I0=5; I1=1 (I1 does not matter here)
+        problems = _report_error(51000)
+        assert_equal(problems, ['Error occurred in callback'])
+
+        # I0=6; I1=1 (I1 does not matter here)
+        problems = _report_error(61000)
+        assert_equal(problems, ['Numerical error detected'])
```
tylerjereddy, line 710: Just a purely mechanical comment, but this is a lot of assertions in one test; `pytest` parametrization might make sense

scipy/odr/tests/test_odr.py
```diff
@@ -605,3 +621,167 @@ def test_pickling_output(self):
...
+
+    def test_data_meta(self):
+        p = Data([0., 0., 5., 7., 7.5, 10.])
+        p.set_meta(name='Sample Data Meta', ref='ODRPACK')
+        assert_equal(p.meta['name'], 'Sample Data Meta')
+        with pytest.raises(AssertionError, match="hurzname"):
+            assert_equal(p.meta,
+                         {'hurzname': 'Sample Data Meta', 'ref': 'ODRPACK'})
```
tylerjereddy, line 718: This looks a bit weird to me -- are you trying to check that `hurzname` is not in `p.meta`? Any reason not to do this directly?

scipy/odr/tests/test_odr.py
```diff
@@ -605,3 +621,167 @@ def test_pickling_output(self):
...
+
+        p_dat = RealData([0., 0., 5., 7., 7.5, 10.],
+                         meta={'name': 'Sample RealData Meta'})
+        assert_equal(p_dat.meta['name'], 'Sample RealData Meta')
+
+    def test_output(self):
+        beta3 = ["parameter1", "parameter2", "parameter3"]
+        output3 = Output(beta3)
+
+        output3.pprint()
```
tylerjereddy, line 735: Just checking for absence of an error on `pprint()` here? If it is worth checking, could probe the actual string contents with i.e., `capsys` https://docs.pytest.org/en/stable/reference/reference.html#std-fixture-capsys

tylerjereddy (commented):
I just added a few mechanical comments

scipy/odr/tests/test_odr.py
```diff
@@ -605,3 +621,167 @@ def test_pickling_output(self):
...
+        data = Data(x, y, wd=1.0, we=1.0)
+
+        model_only_linear = Model(linear)
+        # should not give any exception
+        _ = ODR(data, model_only_linear,
+                beta0=[0.02, 0.0], overwrite=True)
+        _ = ODR(data, model_only_linear,
+                beta0=[0.02, 0.0], overwrite=True, rptfile="gagahurzi")
+        _ = ODR(data, model_only_linear,
+                beta0=[0.02, 0.0], overwrite=True, errfile="gagahurzi")
```
rkern, line 756: If these files are created, I think we need to ensure they are deleted by the test before completion.

rkern (approved):
I think all of these make sense, if a little more implementation-specific than is my usual taste (I would not try to match the details of the exception message text, for instance), but it's within reasonable bounds.

FWIW, I don't think `scipy.odr` needs much more attention to it. I'd prefer folks spend their efforts improving [`odrpack`](https://pypi.org/project/odrpack/) instead.

alteholz:
Thanks a lot for reviewing and approving my tests @rkern, @lucascolley and @tylerjereddy . I would like to become more familiar with scipy and try to achieve this by adding tests. So yes, some tests might look strange, but at least I learned something. I will also have a look at odrpack-python.

scipy/odr/tests/test_odr.py
alteholz: There should have been no file created.

scipy/odr/tests/test_odr.py
alteholz: I just used the stuff that others used before. Would it make sense to globally replace assert_array_almost_equal() by assert_allclose() all at once? Or would it be too error-prone to automatically replace them in about 2500 lines?

tylerjereddy:
> I just used the stuff that others used before. Would it make sense to globally replace assert_array_almost_equal() by assert_allclose() all at once? 

Just for the new cases, as my comment indicates

lucascolley:
hey @alteholz, looks like this is very close. Would you have time to come back and finish it up?

lucascolley:
gentle ping @alteholz

j-bowhay:
We are about to deprecate `scipy.odr` so it probably isn't worth the effort to finish this. Thanks anyway @alteholz

PR #22208, opened 2024-12-28
ENH: io.wavfile: read unseekable files

smammy:
#### Reference issue
Closes gh-11328

#### What does this implement/fix?
Allow scipy.io.wavfile.read() to read non-seekable files.

#### Additional information
Allow scipy.io.wavfile.read() to read non-seekable files.

  * If passed stream is not seekable, wrap it in a SeekEmulatingReader, which tracks current position and implements tell() to report it.

Most seek()s in wavfile.read() simply seek forward from the current position by a fixed amount. To address these:

  * Support limited SeekEmulatingReader.seek() that emulates seek()s that can be supported by read()ing and discarding the result.

Befor returning, read() attempts to seek back to the start of the file. This is nice when possible, but if it isn't we shouldn't blow up. So:

  * Avoid final fid.seek(0) if underlying stream isn't seekable.

The remaining seek()s are in RF64 support, where chunk size is global and stored in the RIFF header rather than in the chunk header. _read_data_chunk() currently seeks back to a fixed offset from the file origin to grab the chunk size. To address this:

  * Modify RF64 support to parse and store the chunk size in _read_riff_chunk() so that _read_data_chunk() doesn't have to seek() back for it. This applies whether or not reading from a seekable stream.

To test this, I added test_streams(), which ensures that wavfile reads the same data no matter whether a path, seekable stream, or unseekable stream is passed to read().

PR #23319, opened 2025-07-11
TST: io.wavfile: add test for `SeekEmulatingReader.seek`

hippowm:
This PR adds a test for `io.wavfile::SeekEmulatingReader.seek`. It ensures seek works correctly when seeking exactly to the end (os.SEEK_END) or unsupoorted seek is used. PR https://github.com/scipy/scipy/pull/22208 introduced this emulator class but some test coverage were missing. 

```py
class SeekEmulatingReader:
    """
    Tracks stream position, provides tell(), and emulates only those
    seeks that can be supported by reading forward. Other seeks raise
    io.UnsupportedOperation. Note that this class implements only the
    minimum necessary to keep wavfile.read() happy.
    """
    def __init__(self, reader):
        self.reader = reader
        self.pos = 0

    def read(self, size=-1, /):
        data = self.reader.read(size)
        self.pos += len(data)
        return data
    
    def seek(self, offset, whence=os.SEEK_SET, /):
        match whence:
            case os.SEEK_SET if offset >= self.pos:
                self.read(offset - self.pos) # convert to relative
            case os.SEEK_CUR if offset >= 0:
                self.read(offset) # advance by offset
            case os.SEEK_END if offset == 0: #✅ NOW COVERED
                self.read() # advance to end of stream #✅ NOW COVERED
            case _: #✅ NOW COVERED
                raise io.UnsupportedOperation("SeekEmulatingReader was asked to emulate" #✅ NOW COVERED
                                              " a seek operation it does not support.") #✅ NOW COVERED
        return self.pos
```

Note: Parts of this test have been automatically generated by a novel technique that we're currently developing as part of an academic research project aiming at improving test coverage. *To not waste developer time, two researchers manually checked the test before submitting it.* We appreciate the developers' time and any feedback is welcomed.

### DIFF
--- a/scipy/io/tests/test_wavfile.py
+++ b/scipy/io/tests/test_wavfile.py
@@ -513,8 +513,5 @@ def test_seek_emulating_reader_invalid_seek():
         reader.seek(-1, 0)  # Negative position with SEEK_SET
     
     # Test SEEK_END with valid parameters (should not raise)
-    try:
-        pos = reader.seek(0, os.SEEK_END)  # Valid usage
-        assert pos == 2  # Check the position after seeking (corrected)
-    except Exception as e:
-        pytest.fail(f'Seek to SEEK_END raised an exception: {e}')
+    pos = reader.seek(0, os.SEEK_END)  # Valid usage
+    assert pos == 2, f"Failed to seek to end, got position {pos}"
--- a/scipy/io/tests/test_wavfile.py
+++ b/scipy/io/tests/test_wavfile.py
@@ -1,6 +1,6 @@
 import os
 import sys
-from io import BytesIO
+from io import (BytesIO, UnsupportedOperation)
 import threading
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


scipy/io/tests/test_wavfile.py
```diff
@@ -499,3 +499,22 @@ def test_wavfile_dtype_unsupported(tmpdir, dtype):
...
+    # Test with negative seek value
+    with pytest.raises(UnsupportedOperation):
+        reader.seek(-1, 0)  # Negative position with SEEK_SET
+    
+    # Test SEEK_END with valid parameters (should not raise)
+    try:
+        pos = reader.seek(0, os.SEEK_END)  # Valid usage
+        assert pos == 2  # Check the position after seeking (corrected)
+    except Exception as e:
+        pytest.fail(f'Seek to SEEK_END raised an exception: {e}')
```
j-bowhay: the try except block is not needed as the test will fail anyway if an exception is raised

nickodell (approved):

PR #23320, opened 2025-07-13
BUG:DOC: sparse.linalg: Fix more arnaud bugs and add C code documentation

ilayn:
This PR adds more C documentation and also fixes a few more bugs; 

- Eigenvector purification off-by-one error in complex valued solvers
- Not handling negative increments properly in custom `zdotc` and `cdotc` (not that arnaud was using any but just in case)
- Shellsort algorithm did not initialize properly in none of the flavors. Now cleaned up and restructured.

Also minor project restructuring is added regarding the type declarations and includes.

### DIFF
--- a/scipy/sparse/linalg/_eigen/arpack/arnaud/include/arnaud/arnaud.h
+++ b/scipy/sparse/linalg/_eigen/arpack/arnaud/include/arnaud/arnaud.h
@@ -1,52 +1,49 @@
 #ifndef ARNAUD_H
 #define ARNAUD_H
 
-#include <math.h>
-#include <complex.h>
-
-#if defined(_MSC_VER)
-    // MSVC definitions
-    typedef _Dcomplex ARNAUD_CPLX_TYPE;
-    typedef _Fcomplex ARNAUD_CPLXF_TYPE;
-
-#else
-    // C99 compliant compilers
-    typedef double complex ARNAUD_CPLX_TYPE;
-    typedef float complex ARNAUD_CPLXF_TYPE;
-
-#endif
-
+#include "types.h"
 
 enum ARNAUD_which {
-    which_LM = 0,    // want the NEV eigenvalues of largest magnitude.
-    which_SM = 1,    // want the NEV eigenvalues of smallest magnitude.
-    which_LR = 2,    // want the NEV eigenvalues of largest real part.
-    which_SR = 3,    // want the NEV eigenvalues of smallest real part.
-    which_LI = 4,    // want the NEV eigenvalues of largest imaginary part.
-    which_SI = 5,    // want the NEV eigenvalues of smallest imaginary part.
-    which_LA = 6,    // compute the NEV largest (algebraic) eigenvalues. (sym)
-    which_SA = 7,    // compute the NEV smallest (algebraic) eigenvalues. (sym)
-    which_BE = 8     // compute NEV eigenvalues, half from each end of the spectrum. (sym)
-};
-
-enum ARNAUD_ido {
-    ido_FIRST      = 0,  // First call
-    ido_OPX        = 1,  // OP*x needed
-    ido_BX         = 2,  // B*x needed
-    ido_USER_SHIFT = 3,  // User shifts are needed
-    ido_RANDOM     = 4,  // A random vector is needed to be written in resid
-    ido_RANDOM_OPX = 5,  // Force random vector to be in the range of OP
-    ido_DONE       = 99  // Done
+    which_LM = 0,    /**> want the NEV eigenvalues of largest magnitude. */
+    which_SM = 1,    /**> want the NEV eigenvalues of smallest magnitude. */
+    which_LR = 2,    /**> want the NEV eigenvalues of largest real part. */
+    which_SR = 3,    /**> want the NEV eigenvalues of smallest real part. */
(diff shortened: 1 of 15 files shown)


tylerjereddy:
> fixes a few more bugs

I'll ask the simple question I guess -- are any of these possible to regression test via public API of sparse linalg?

ilayn:
You are right. I found these by source reading but I don't know how to hit these algorithmically. Tried a bunch of examples but didn't manage to fail in a meaningful way.

 Shellsort parts seemed weird if not wrong but somehow I cannot demonstrate they were wrong. Instead went for the canonical style.

scipy/sparse/linalg/_eigen/arpack/arnaud/include/arnaud/arnaud.h
```diff
@@ -1,156 +1,371 @@
...
-    which_LM = 0,    // want the NEV eigenvalues of largest magnitude.
-    which_SM = 1,    // want the NEV eigenvalues of smallest magnitude.
-    which_LR = 2,    // want the NEV eigenvalues of largest real part.
-    which_SR = 3,    // want the NEV eigenvalues of smallest real part.
-    which_LI = 4,    // want the NEV eigenvalues of largest imaginary part.
-    which_SI = 5,    // want the NEV eigenvalues of smallest imaginary part.
-    which_LA = 6,    // compute the NEV largest (algebraic) eigenvalues. (sym)
-    which_SA = 7,    // compute the NEV smallest (algebraic) eigenvalues. (sym)
-    which_BE = 8     // compute NEV eigenvalues, half from each end of the spectrum. (sym)
+    which_LM = 0,    /** want the NEV eigenvalues of largest magnitude.                      */
```
rgommers, line 7: If you're changing style, why the double `**` at the start? Normal is `/* text-here */`

scipy/sparse/linalg/_eigen/arpack/arnaud/include/arnaud/arnaud.h
```diff
@@ -1,156 +1,371 @@
...
+    which_LR = 2,    /** want the NEV eigenvalues of largest real part.                      */
+    which_SR = 3,    /** want the NEV eigenvalues of smallest real part.                     */
+    which_LI = 4,    /** want the NEV eigenvalues of largest imaginary part.                 */
+    which_SI = 5,    /** want the NEV eigenvalues of smallest imaginary part.                */
+    which_LA = 6,    /** compute the NEV largest (algebraic) eigenvalues. (sym)              */
+    which_SA = 7,    /** compute the NEV smallest (algebraic) eigenvalues. (sym)             */
+    which_BE = 8     /** compute NEV eigenvalues, half from each end of the spectrum. (sym)  */
 };
 
+/**
```
rgommers, line 18: Same for multiline comment style, normal is starting line `/*` rather than `/**`

scipy/sparse/linalg/_eigen/arpack/arnaud/include/arnaud/types.h
```diff
@@ -0,0 +1,23 @@
+#ifndef ARNAUD_TYPES_H
+#define ARNAUD_TYPES_H
+
+#include <stdint.h>
+#include <complex.h>
+
+
+#if defined(_MSC_VER)
+    // MSVC definition
```
rgommers, line 9: nit: if you're changing code style everywhere to `/* */`, may as well be consistent in this file as well.

scipy/sparse/linalg/_eigen/arpack/arnaud/src/arnaud_n_double_complex.c
```diff
@@ -1844,15 +1844,20 @@ zdotc_(const int* n, const ARNAUD_CPLX_TYPE* restrict x, const int* incx, const
         }
 
     } else {
+        // Handle negative increments correctly
```
rgommers, line 1847: and here too

rgommers (approved):
Overall LGTM, a few comments on the comment style since you're deliberately changing it here. I didn't verify the `*dotc` fix in detail, going to assume that you checked that carefully enough and CI is happy.

ilayn:
Thanks Ralf, the `/** ... */` is the doxygen documentation convention which, I presume, probably where this is going to be looked at or when hovered over in the IDE. I don't use these anywhere else but I tried to apply some "modern" C documentation practices (which are very weird at some topics). 

I didn't go for any specific standard as it seemed like `"""`/`"""` vs. `#` in Python.  Since I kept on forgetting closing `*/` I only used them in the header file doc parts. Otherwise it should be `//` elsewhere (emphasis on should).

rgommers:
> the `/** ... */` is the doxygen documentation convention

Didn't realize that. Doxygen calls it the "Javadoc style" which is a little 🙃 🙈  (https://www.doxygen.nl/manual/docblocks.html). 

Feel free to hit the merge button if you prefer it like this I'd say.

ilayn:
> Doxygen calls it the "Javadoc style" which is a little 🙃 🙈

Indeed sometimes we forget how much Python improved the quality of life with its tooling and with its conventions. 

> Feel free to hit the merge button if you prefer it like this I'd say.

I just noticed 4 compiler warnings in the MSVC runs. I'll fix them and we can send it in.

ilayn:
Thank you Tyler, Ralf. Clicking the button.

PR #23326, opened 2025-07-14
DEV: remove non-generated file from `.gitignore`

lucascolley:
https://scipy-community.slack.com/archives/C03JL2YGCNR/p1752488458012589


ev-br:
The linked slack comment is

> No it's an accident. It used to be a Tempita-templated file (.pyx.in), which is why the .pyx one was ignored. Once templating was removed, the .gitignore entry should have been removed.

PR #23249, opened 2025-06-27
ENH: spatial.transform.Rotation: add array API backend

amacati:
Add an array API backend to `Rotation`.

This PR implements a generic version `_rotation_xp.py` of the functional-style cython backend `_rotation_cy.pyx` using only array API features. Since we want to support lazy backends and `jax.jit` compilation, the backend avoids value-dependent branching wherever possible.

### Additional considerations
Due to shape typing restrictions, the cython backend can only use broadcasting in a few select places. The array API backend instead relies heavily on broadcasting to directly produce correct shapes without the `_single` distinction. This reduces special-case handling and would allow us to add support for multiple batch dimensions for `Rotation` in future PRs.

## Reference issue
#22500 

## Related PRs
#22777, #23198

## What does this implement/fix?
Implement the generic Array API backend for `Rotation`.

## Additional information
The Array API tests have been converted in #22939 . These tests have been left unchanged with a few exceptions:
- Tests using `xp.pi` are failing with a namespace mismatch if the constant is used without wrapping. We now use `xp.asarray(xp.pi)[()]` to force the correct namespace for `xp.pi`.
- `cupy` does not support array.mT as of the latest release `13.4.1` . This will be fixed in the upcoming release `14.0`. See https://github.com/data-apis/array-api-compat/issues/316. Tests using this property are skipped for now
- `jax` emits a different error for zero-slice indexing. The tests have been adapted to this special case. Instead we could also catch this error ourselves and raise an `IndexError` ourselves.
- All tests have been adapted to also work with `float32` precision

## Benchmarks
See https://github.com/scipy/scipy/pull/23249#issuecomment-3071232726.

## TODOs
- [x] Benchmarks
- [x] Investigate slow iteration over `jax`-backed `Rotation`s
- [x] Decide jax index error handling
- [x] **Optional:** Accelerate tests with jax.jit where possible to keep test times low

PR #23331, opened 2025-07-15
TST: Read `@xp_capabilities` from `pytestmark`

crusaderky:
In support of https://github.com/scipy/scipy/pull/23249.
Allow adding module-wide `skip_xp_backends` marks from `@xp_capabilities`.

### DIFF
--- a/scipy/_lib/_array_api.py
+++ b/scipy/_lib/_array_api.py
@@ -51,7 +51,8 @@ __all__ = [
     'xp_assert_close', 'xp_assert_equal', 'xp_assert_less',
     'xp_copy', 'xp_device', 'xp_ravel', 'xp_size',
     'xp_unsupported_param_msg', 'xp_vector_norm', 'xp_capabilities',
-    'xp_result_type', 'xp_promote'
+    'xp_result_type', 'xp_promote',
+    'make_xp_test_case', 'make_xp_pytest_marks', 'make_xp_pytest_param',
 ]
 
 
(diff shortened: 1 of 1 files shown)


scipy/_lib/_array_api.py
```diff
@@ -851,13 +848,72 @@ def test(func, verb, xp):
...
     array_api_extra.testing.lazy_xp_function
     """
     import pytest
 
-    marks = _make_xp_pytest_marks(func, capabilities_table=capabilities_table)
+    marks = make_xp_pytest_marks(func, capabilities_table=capabilities_table)
     return pytest.param(func, *args, marks=marks, id=func.__name__)
 
 
+def make_xp_pytest_marks(*funcs, capabilities_table=None):
```
crusaderky, line 860: Moved the function for ease of reading, made public, and added documentation.
Implementation has not been changed.

lucascolley (approved):
thanks Guido, happy if CI is happy!

scipy/_lib/_array_api.py
```diff
@@ -799,13 +770,39 @@ def make_xp_test_case(*funcs, capabilities_table=None):
...
+        @make_xp_test_case(f1, f2)
+        def test_f1_and_f2(xp):
+            ...
+
+    The above is equivalent to::
+        @pytest.mark.skip_xp_backends(...)
+        @pytest.mark.skip_xp_backends(...)        
+        @pytest.mark.xfail_xp_backends(...)
+        @pytest.mark.xfail_xp_backends(...)
+        def test_f1(xp):
```
tylerjereddy, line 792: For the four decorator case, populated from two originally-decorated funcs, should this be `test_f1_and_f2()` instead since `test_f1(xp)` above has a mono-function `make_xp_test_case`?

Or can a single-function `make_xp_test_case(f1)` expand to multiple `skip_xp_backends` decorators?

crusaderky: It expands to multiple ones.
`@skip_xp_backends` accepts exactly one backend as its parameter. In tests that need to skip more than one and that have not been yet migrated to `@xp_capabilities` you will read e.g.

```python
@pytest.mark.skip_xp_backends("cupy")        
@pytest.mark.skip_xp_backends("torch")
def test_f1(xp):
    ...
```

PR #23332, opened 2025-07-15
BLD: Add bigobj flag for MSVC to fix object file section limit

### LINKED ISSUE #23329: BUG: Compilation failure in HiGHs with MSVC
### Describe your issue.

Recently, the following error shows up quite often in CI: https://github.com/scipy/scipy/actions/runs/16271430193/job/45939816354?pr=22580

The error message points to the likely fix already.

### Reproducing Code Example

```python
n/a
```

### Error message

```shell
D:\a\scipy\scipy\subprojects\highs\src\highs_bindings.cpp : fatal error C1128: number of sections exceeded object fileformat limit: compile with /bigobj
```

```shell
main branch
```

Greenie0701:
Closes #23329

PR Description:

- MSVC fails to compile highs_bindings.cpp due to exceeding the default object file section limit.
- This patch adds the /bigobj flag when using MSVC to resolve the C1128 compiler error.

Reference:
image

CI Job failure : https://github.com/scipy/scipy/actions/runs/16298418130/job/46026301158

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -111,6 +111,10 @@ if cc.get_id() == 'clang-cl'
   add_project_arguments('-Qunused-arguments', language: ['c', 'cpp'])
 endif
 
+if cpp.get_id() == 'msvc'
+  add_project_arguments('/bigobj', language: 'cpp')
+endif
+
 if host_machine.system() == 'darwin'
   if cc.has_link_argument('-Wl,-dead_strip')
     # Allow linker to strip unused symbols


lucascolley (commented):
@galabovaa I wonder whether this belongs in https://github.com/ERGO-Code/HiGHS/blob/364c83a51e44ba6c27def9c8fc1a49b1daf5ad5c/meson.build#L71-L95 instead? Do you build with MSVC in CI upstream / are you familiar with an error like https://github.com/scipy/scipy/issues/23329 ?

MGN-GIT:
I was getting a similar kind of error on Windows on Arm builds too using MSVC, so I added this flag in build file to fix the issue. Not sure that adding this flag in high proj can resolve this issue, but I can give a try

rgommers:
Due to `pybind11` 3.0.0, but doesn't seem like a regression - it looks like we should be using `/bigobj` for MSVC with `pybind11` indeed:

https://github.com/pybind/pybind11/blob/cc69a3789cace002d6bb77d50ad8abf558bb5176/tools/pybind11Config.cmake.in#L59

https://github.com/pybind/pybind11/blob/cc69a3789cace002d6bb77d50ad8abf558bb5176/pybind11/setup_helpers.py#L149-L150

https://pybind11.readthedocs.io/en/stable/compiling.html#advanced-interface-library-targets

> @galabovaa I wonder whether this belongs in https://github.com/ERGO-Code/HiGHS/blob/364c83a51e44ba6c27def9c8fc1a49b1daf5ad5c/meson.build#L71-L95 instead?

It may be useful to someone to put the same fix into HiGHS, but we shouldn't be relying on that - it's connected to `pybind11` usage and we use that outside of HIGHS targets as well.

rgommers (approved):
Thanks @lowkeyrossi. I think it's reasonable to do this for all C++ targets rather than the immediately problematic target, since we have other heavy C++ components and this may prevent the problem with MSVC from reoccurring.

MGN-GIT:
That sounds reasonable @rgommers , thanks!

PR #23346, opened 2025-07-17
DOC: correct dimensions of PPoly.c in docstring

### LINKED ISSUE #23345: DOC: Incorrect dimension of PPoly.c in docstring

For a PPoly of degree `k`, the shape of the coefficients array is  (k+1, m, ...) and not `(k, m, ...)` as currently written.
( https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.PPoly.html ).

In accordance with standard definition of polynomial degree, a polynomial of degree k has k+1 coefficients.
This can also be verified by comparison with the formula that appears just above the parameter definitions, in which the  loop index `m` goes over `k+1` (not k) different values.

_No response_

_No response_

AmitAronovitch:
#### Reference issue

Closes gh-23345

Fix incorrect array dimension in parameters block of the docstring

#### Additional information

See in the issue

### DIFF
--- a/scipy/interpolate/_interpolate.py
+++ b/scipy/interpolate/_interpolate.py
@@ -830,7 +830,7 @@ class PPoly(_PPolyBase):
     Parameters
     ----------
     c : ndarray, shape (k+1, m, ...)
-        Polynomial coefficients, order `k` and `m` intervals.
+        Polynomial coefficients, degree `k` and `m` intervals.
     x : ndarray, shape (m+1,)
         Polynomial breakpoints. Must be sorted in either increasing or
         decreasing order.
--- a/scipy/interpolate/_interpolate.py
+++ b/scipy/interpolate/_interpolate.py
@@ -829,7 +829,7 @@ class PPoly(_PPolyBase):
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


scipy/interpolate/_interpolate.py
```diff
@@ -829,7 +829,7 @@ class PPoly(_PPolyBase):
 
     Parameters
     ----------
-    c : ndarray, shape (k, m, ...)
+    c : ndarray, shape (k+1, m, ...)
         Polynomial coefficients, order `k` and `m` intervals.
```
ev-br:
```suggestion
        Polynomial coefficients, degree `k` and `m` intervals.
```

ev-br (changes requested):
Nice catch. I think the intention here was to use "order" being degree +1 (so that a linear function has degree 1 and order 2, a cubic has degree 3 and order 4 etc). I see this gets confusing, so we should make the change here and change "order" into "degree" throughout.

ev-br:
LGTM, thank you @AmitAronovitch, and congrats with what I believe is your first scipy contribution. Keep them coming!

PR #23349, opened 2025-07-18
BENCH, TST: benchmarks without hypothesis

tylerjereddy:
* Fixes gh-23344.

* The new array API `asv` benchmarks depend on `scipy.conftest`, which has a mandatory dependency on `hypothesis`. This patch makes it possible to extract the necessary array API information from `conftest` without needing the `hypothesis` dependency, so that benchmarks can run in that situation.

* I'm not sure if it is desirable long-term to have this kind of cross-talk between the `asv` benchmarks and our testing infrastructure, but this seemed simplest for now. Note that `hypothesis` is still mandatory for running our full testsuite.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -78,7 +78,7 @@ jobs:
             pip install -r requirements/doc.txt
             pip install mpmath gmpy2 "asv>=0.6.0" pooch threadpoolctl spin
             # extra benchmark deps
-            pip install pyfftw cffi pytest hypothesis
+            pip install pyfftw cffi pytest
 
       - run:
           name: build SciPy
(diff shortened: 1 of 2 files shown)


tylerjereddy:
It may be worth noting that Guido did add `hypothesis` to the CircleCI config in gh-23012, and the benchmarks do run with `--python=same` there, so there may be something else unintended happening. That said, I think I still agree with Evgeni that we shouldn't need `hypothesis` in this situation regardless.

crusaderky (changes requested):
With this change, you can remove hypothesis from here:
https://github.com/scipy/scipy/blob/38cb4d90e871d3c8265e040c172ec818b54eb700/.circleci/config.yml#L81

tylerjereddy:
thanks, I pushed in a revision to do that

lucascolley (approved):
thanks both!

Issue #23072, opened 2025-05-28
Deprecate `lower` and `transposed` arguments of `linalg.solve`


ev-br:
The signature of `linalg.solve` is a bit crowded, and the two parameters, `lower` and `transposed`, seem to be either redundant or confusing. Their use has been mostly superseded by `assume_a` keyword. 

TL;DR: I think it's a good idea to get rid of them.

Consider `lower={True,False}`. First of all, whether it has any effect depends on other arguments (which is in itself an antipattern): The [docstring](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.solve.html) says _Ignored unless assume_a is one of 'sym', 'her', or 'pos'_. 

So let's consider some usage:

- `>>> solve(a, b, lower=True)   # redundant`  :  The routine runs structure detection anyway, so the keyword does not do anything;
- `>>> solve(a, b, assume_a="general", lower=True)   # redundant`  :   The keyword is ignored, as documented : 
- `>>> solve(a, b, assume_a="lower triangular", lower=True)    # redundant`  :   The triangle is already selected via `assume_a`.
- `>>> solve(a, b, assume_a="upper triangular", lower=True)    # ???`  :   No idea which triangle will it use without reading the code.

A legit usage is `>>> solve(a, b, assume_a="sym", lower=True)`.

An alternative API is suggested in https://github.com/scipy/scipy/pull/22924 : use a single string `assume_a="sym lower"`. Here the allowed values are `"sym"`  (the matrix is symmetrix, pick a triangle automatically), ``"sym lower"` (the matrix is symmetric, the upper triangle is not referenced) and `"sym upper"` (the lower triangle is not referenced).

------------------

Now consider `transposed={True, False}` keyword. The [docstring](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.solve.html) says _If True, solve `a.T @ x == b`_.  If so, why do we need a special keyword for this, a user can transpose the input just as well.

What the keyword actually does, it forwards the value to the LAPACK `trans={'N', 'T', 'C'}` parameter --- but the argument is boolean, and we cannot express `'C'`, so it raises a `NotImplementedError`. 

The interaction of the `transposed` argument with other arguments is... something:

```
>>> solve(a, b, assume_a='lower triangular', transposed=True)    # is it `np.triu(a.T)` or  `np.triu(a).T` ?
>>> solve(a, b, assume_a='lower triangular', transposed=True, lower=False)`    # ???
```

[...]

PR #23350, opened 2025-07-18
BUG: linalg: fix `solve(..., assume_a='pos', lower=False)`

ev-br:
#### Reference issue

Reported in https://github.com/scipy/scipy/pull/23071#issuecomment-3085826112

Actually plumb the 'lower' argument to the Cholesky routines in `scipy.linalg.solve`. This is a regression from gh-23071, is thus only relevant to the -dev version.

#### Additional information

While at it, improve validation for conflicting `assume_a` and `lower` arguments. At some point we should consider https://github.com/scipy/scipy/issues/23072 but for now we should error out for definitely wrong combinations.

### DIFF
--- a/scipy/linalg/_basic.py
+++ b/scipy/linalg/_basic.py
@@ -63,7 +63,7 @@ def _find_matrix_structure(a):
     return kind, n_below, n_above
 
 
-def solve(a, b,lower=None, overwrite_a=False,
+def solve(a, b, lower=None, overwrite_a=False,
           overwrite_b=False, check_finite=True, assume_a=None,
           transposed=False):
     """
--- a/scipy/linalg/_basic.py
+++ b/scipy/linalg/_basic.py
@@ -63,7 +63,7 @@ def _find_matrix_structure(a):
     return kind, n_below, n_above
 
 
-def solve(a, b,lower=False, overwrite_a=False,
+def solve(a, b,lower=None, overwrite_a=False,
           overwrite_b=False, check_finite=True, assume_a=None,
           transposed=False):
     """
@@ -178,6 +178,8 @@ def solve(a, b,lower=False, overwrite_a=False,
         'sym', 'her', 'symmetric', 'hermitian', 'diagonal', 'tridiagonal', 'banded'
     ]:
         # TODO: handle these structures in this function
+        if lower is None:
+            lower = False
         return solve0(
             a, b, lower=lower, overwrite_a=overwrite_a, overwrite_b=overwrite_b,
             check_finite=check_finite, assume_a=assume_a, transposed=transposed
@@ -196,6 +198,15 @@ def solve(a, b,lower=False, overwrite_a=False,
     }.get(assume_a, 'unknown')
     if structure == 'unknown':
         raise ValueError(f'{assume_a} is not a recognized matrix structure')
+
+    if (
+        (assume_a == 'pos upper' and lower is True) or
+        (assume_a == 'pos lower' and lower is False)
+    ):
+        raise ValueError(f"Conflicting {assume_a = } and {lower = }.")
+
+    if lower is None:
+        lower = False
 
     a1 = np.atleast_2d(_asarray_validated(a, check_finite=check_finite))
     b1 = np.atleast_1d(_asarray_validated(b, check_finite=check_finite))
@@ -248,7 +259,7 @@ def solve(a, b,lower=False, overwrite_a=False,
         return out[..., 0] if b_is_1D else out
 
     # heavy lifting
-    result = _batched_linalg._solve(a1, b1, structure, transposed)
+    result = _batched_linalg._solve(a1, b1, structure, lower, transposed)
     x, is_ill_cond, is_singular, info = result
 
     if info < 0:
--- a/scipy/linalg/src/_batched_linalg_module.cc
+++ b/scipy/linalg/src/_batched_linalg_module.cc
@@ -101,9 +101,10 @@ _linalg_solve(PyObject* Py_UNUSED(dummy), PyObject* args) {
     St structure = St::NONE;
     int overwrite_a = 0;
     int transposed = 0;
+    int lower=0;
 
     // Get the input array
-    if (!PyArg_ParseTuple(args, "O!O!|npp", &PyArray_Type, (PyObject **)&ap_Am, &PyArray_Type, (PyObject **)&ap_b, &structure, &transposed, &overwrite_a)) {
+    if (!PyArg_ParseTuple(args, "O!O!|nppp", &PyArray_Type, (PyObject **)&ap_Am, &PyArray_Type, (PyObject **)&ap_b, &structure, &lower, &transposed, &overwrite_a)) {
         return NULL;
     }
 
@@ -154,16 +155,16 @@ _linalg_solve(PyObject* Py_UNUSED(dummy), PyObject* args) {
     void *buf = PyArray_DATA(ap_x);
     switch(typenum) {
         case(NPY_FLOAT32):
-            _solve<float>(ap_Am, ap_b, (float *)buf, structure, transposed, overwrite_a, &isIllconditioned, &isSingular, &info);
+            _solve<float>(ap_Am, ap_b, (float *)buf, structure, lower, transposed, overwrite_a, &isIllconditioned, &isSingular, &info);
             break;
         case(NPY_FLOAT64):
-            _solve<double>(ap_Am, ap_b, (double *)buf, structure, transposed, overwrite_a, &isIllconditioned, &isSingular, &info);
+            _solve<double>(ap_Am, ap_b, (double *)buf, structure, lower, transposed, overwrite_a, &isIllconditioned, &isSingular, &info);
             break;
         case(NPY_COMPLEX64):
-            _solve<npy_complex64>(ap_Am, ap_b, (npy_complex64 *)buf, structure, transposed, overwrite_a, &isIllconditioned, &isSingular, &info);
+            _solve<npy_complex64>(ap_Am, ap_b, (npy_complex64 *)buf, structure, lower, transposed, overwrite_a, &isIllconditioned, &isSingular, &info);
             break;
         case(NPY_COMPLEX128):
-            _solve<npy_complex128>(ap_Am, ap_b, (npy_complex128 *)buf, structure, transposed, overwrite_a, &isIllconditioned, &isSingular, &info);
+            _solve<npy_complex128>(ap_Am, ap_b, (npy_complex128 *)buf, structure, lower, transposed, overwrite_a, &isIllconditioned, &isSingular, &info);
             break;
         default:
             PYERR(PyExc_RuntimeError, "Unknown array type.")
--- a/scipy/linalg/src/_linalg_solve.hh
+++ b/scipy/linalg/src/_linalg_solve.hh
@@ -110,7 +110,7 @@ inline CBLAS_INT solve_slice_cholesky(
 
 
 template<typename T>
-void _solve(PyArrayObject* ap_Am, PyArrayObject *ap_b, T* ret_data, St structure, int transposed, int overwrite_a, int* isIllconditioned, int* isSingular, int* info)
+void _solve(PyArrayObject* ap_Am, PyArrayObject *ap_b, T* ret_data, St structure, int lower, int transposed, int overwrite_a, int* isIllconditioned, int* isSingular, int* info)
 {
     using real_type = typename type_traits<T>::real_type; // float if T==npy_cfloat etc
 
@@ -119,7 +119,7 @@ void _solve(PyArrayObject* ap_Am, PyArrayObject *ap_b, T* ret_data, St structure
     char trans = transposed ? 'T' : 'N'; 
     npy_intp lower_band = 0, upper_band = 0;
     bool is_symm = false;
-    char uplo;
+    char uplo = 'X';    // sentinel
     St slice_structure = St::NONE;
     bool posdef_fallback = true;
 
@@ -173,10 +173,9 @@ void _solve(PyArrayObject* ap_Am, PyArrayObject *ap_b, T* ret_data, St structure
     if (irwork == NULL) { free(irwork); *info = -102; return; }
 
     // normalize the structure detection inputs
-    uplo = 'U';
     if (structure == St::POS_DEF) {
-        uplo = 'U';
         posdef_fallback = false;
+        uplo = lower ? 'L' : 'U';
     }
     else {
         if (structure == St::POS_DEF_UPPER) {
--- a/scipy/linalg/tests/test_basic.py
+++ b/scipy/linalg/tests/test_basic.py
@@ -1008,6 +1008,33 @@ class TestSolve:
         b = np.ones((2, 2)) * [1, 2]
         assert_allclose(solve(a, b, assume_a=assume_a), np.linalg.solve(a, b))
 
+    def test_pos_lower(self):
+        # regression test for
+        # https://github.com/scipy/scipy/pull/23071#issuecomment-3085826112
+        rng = np.random.default_rng(0)
+        a = rng.normal(size=(4, 4))
+        a = np.tril(np.matmul(a, np.conj(a.T)))  # lower triangle of hermitian array
+        b = rng.normal(size=(4, 2))
+        out = solve(a, b, assume_a='pos', lower=True)
+
+        aa = a + a.T - np.diag(np.diag(a))   # the full hermitian array
+        result_np = np.linalg.solve(aa, b)
+        assert_allclose(out, result_np, atol=1e-15)
+
+        out = solve(a, b, assume_a='pos lower')
+        assert_allclose(out, result_np, atol=1e-15)
+
+        # repeat with uplo='U'
+        out = solve(a.T, b, assume_a='pos', lower=False)
+        assert_allclose(out, result_np, atol=1e-15)
+
+        out = solve(a.T, b, assume_a='pos upper')
+        assert_allclose(out, result_np, atol=1e-15)
+
+        # conflicting assume_a & lower
+        with assert_raises(ValueError):
+            solve(a, b, assume_a='pos upper', lower=True)
+
     def test_readonly(self):
         a = np.eye(3)
         a.flags.writeable = False


scipy/linalg/_basic.py
```diff
@@ -63,7 +63,7 @@ def _find_matrix_structure(a):
     return kind, n_below, n_above
 
 
-def solve(a, b,lower=False, overwrite_a=False,
+def solve(a, b,lower=None, overwrite_a=False,
```
lucascolley:
```suggestion
def solve(a, b, lower=None, overwrite_a=False,
```

ilayn:
I think pos should not have upper or lower. That's a data spec not an array spec

ev-br:
> ... upper or lower. That's a data spec not an array spec

No idea what this means TBH :-). My only points are that `solve(a, b, assume_a="gen", lower=True)` ignores `lower` and that `solve(a, b, assume_a="upper triangular", lower=True)` makes no sense.

ilayn:
Pos is a property of the matrix. It is full so not like triangular. lower is about telling where the data lives. There is no matrix that is positive definite and upper triangular. So it does not belong into assume_a 

Upper triangular with lower=True makes sense but we don't use it. You can tell the algorithm to use the upper side of the array which represents a lower triangular array. Historical luggage combined with modern api's cause this type of issues.

ev-br:
> It is full so not like triangular.

If you look at the OP, it zeros out the "other" triangle.

ilayn:
That's what I mean. Same example should also work whether you zero it or not. Because pos definite is independent from the data. It means you can assume that Cholesky will pass. 

Then when switching to Cholesky solver where should we look at the data? If lower=True we take the lower half, otherwise we take the upper side. If not given we default to one of them. This is not related to pos upper or pos lower. This is `assume_a = pos` then discover where the data is via `lower`. Hence pos is an array type spec. `lower` is about where the data is. 

Unfortunately, and historically lower is also used to do which Cholesky solver we should use LL^T or U^TU, depending on the lower keyword. This complicated things further so we overloaded the lower keyword meaning and leaked it to internal implementation.

ev-br:
There would've been no argument from me if we were discussing `solve_cholesky`. But since this all is in `solve` for all matrix structures, mixing structure and storage is IMO a lesser evil than conflicting arguments, https://github.com/scipy/scipy/pull/23350#issuecomment-3089344244.

ilayn:
I don't think there is any mixing structure. `pos` is very clear switch to Cholesky, `sym` is very clear for symmetric and triangulars are very clear. Note that we are pretty much following matlab and others here so this is not something we are cooking up ourselves. So I don't agree that `pos upper` and others are suitable things because our scheme is getting complicated. In fact pos upper is the mix that I am objecting to. If I write `assume_a=pos`  I want all the checks to be skipped and jump to cholesky. Then `lower` tells me where do I copy the input data. That part has no mix in it. 

If you remember this is exactly why I started with those drawings as this is really, really not so trivial to bundle up in low-level code.

ilayn:
Same with triangular too but we are a bit too late for making triangular without Upper and Lower. I didn't catch it at the time.

ev-br:
Do you consider either of these a problem:
-  `assume_a="lower triangular", lower=False`
- `assume_a="general", lower=False`

Because either we state that these are not a problem (and my point is that they are), or we need to get rid of `lower`, with a deprecation etc.

ilayn:
What I initially though is something like a `solver_options` kwarg. Back then I was experimenting with using 

```
solve(a, assume_a=`bidiag`, solver_options={lower=1, format='packed'})
```

 to accomodate for, say, `solve_banded`, `solve_tridiagonal` and other cases. But that also looked a bit strange and wordy so I did not pursue further.

In your examples;  the first one is what I mean by not catching it on time because those two are the odd one out currently. `assume_a` should only be `tri` or `triangular`. That kwarg is only meant for the type of the array. Not the data location but I don't know if we can back pedal it. If we can, that would be really nice since then we can type in the documentation that `lower` value is ignored for all but `sym`, `her`, `pos`, `tri`, and `bidiag`. This also fixes your second issue as a no-op. In fact that is already mentioned in the docs as such

>lower bool, default: False
>    Ignored unless assume_a is one of 'sym', 'her', or 'pos'. If True, the calculation uses only the data in the lower triangle of a; entries above the diagonal are ignored. If False (default), the calculation uses only the data in the upper triangle of a; entries below the diagonal are ignored.

ilayn:
For deprecating `lower` entirely, I think that would require a full-blown redesign of all types. It can be done but folks are very sticky to their old habits so we need to come up with upper pos, upper tri, upper sym, upper her upper bidiag and the lower counterparts so that there is no feature loss. I don't know if this is a better way due to very lengthy option list.

ev-br:
> For deprecating lower entirely, I think that would require a full-blown redesign of all types. It can be done but folks are very sticky to their old habits so we need to come up with upper pos, upper tri, upper sym, upper her upper bidiag and the lower counterparts so that there is no feature loss. I don't know if this is a better way due to very lengthy option list.

This is exactly what I'm advocating for. `assume_a = "sym upper"` or `"upper sym"` or with a comma if needs be: `"sym, upper"`. And then deprecate a separate `lower={True, False}`.

ilayn:
I get that but now you are baking the entries in with data and specs are mixed up. Like I mentioned `pos upper` is quite unorthodox and hard to guess about because it has no correspondence to the actual array structure but some weird memory property tagged to the name. We still have Hessenberg, tridiag and others in the queue and the `assume_a` list will grow into a very long list if we keep adding these rather non-intuitive names. Say we wanted to add packed formatted input. Then it would also become `pos upper packed`. It will be a list with no end at some point. 

It might be making "our" lives easier but I don't think it is making a better feature. We have to provide a generic and customizable APIs instead of just enumerating different possibilities. 

I don't know how else to mention this but `solve` API is as complicated as the other parts of SciPy but I have not managed to convey this message so far since #12824. While we are taking time making these designs at other parts, it seems rushed about `solve` for details as such are being fixed on the go.

jakevdp (approved):

jakevdp:
Approving these changes – I think the discussion here is useful to hash out, but probably separately from this fix, which we should merge in order to fix the regression at head.

ev-br:
> Say we wanted to add packed formatted input. Then it would also become pos upper packed. It will be a list with no end at some point.

The number of options is exactly the same whether we have a separate "lower" argument or not.

> It might be making "our" lives easier but I don't think it is making a better feature. We have to provide a generic and customizable APIs instead of just enumerating different possibilities.

I again don't see why separate `lower` is anyhow more customizable or generic. Other than allowing nonsense combinations of parameters.

> I don't know how else to mention this but solve API is as complicated as the other parts of SciPy but I have not managed to convey this message so far since #12824. While we are taking time making these designs at other parts, it seems rushed about solve for details as such are being fixed on the go.

Sure, let's continue the discussion in https://github.com/scipy/scipy/issues/23072, which is open and is hopefully more discoverable than a closed PR from a few years back.

ilayn:
> The number of options is exactly the same whether we have a separate "lower" argument or not.

Of course not. If you have 4 options to kwarg1 and lower=1, 0 instead of 8 separate lower upper combos you are dealing with four options and choosing where the data lives. I mean can we conclude that part at least? What is your strong reaction to `lower` ? I don't see the point of removing it at all for no additional benefit but renaming existing options. 

> I again don't see why separate lower is anyhow more customizable or generic. Other than allowing nonsense combinations of parameters.

It does  not change the `assume_a` keyword options if we decide to add other solver options. Like I said array type is independent from what we do with the array. This is also used everywhere else in linalg with zero issues previously. So why do you want to go against the grain in these ones? 

> Sure, let's continue the discussion in https://github.com/scipy/scipy/issues/23072, which is open and is hopefully more discoverable than a closed PR from a few years back.

The point of the closed PR is that there is a clear reasoning for adding `assume_a` to start with and implementing a backslash operator. That plan is not the plan we have apparently and we are freestyling as we encounter issues. And I don't like how we are handling it. I am -1 on diverting from that without any benefit but keyword style which we introduced keywords for a very particular reason.

jakevdp:
Can we merge this fix and continue the design discussion in https://github.com/scipy/scipy/issues/23072?

(We're trying to get out a new JAX release and our tests are broken because scipy nightly has an unannounced behavior change in its `solve` API).

ilayn:
We can but can we get some freedom on breaking nightly behavior? Otherwise why do you release over SciPy's nightlies?  That should not have any stability guarantee. We are already stressing over the regular releases, we don't need extra promises made to downstream.

lucascolley (approved):

lucascolley:
thanks all!

ilayn:
@lucascolley Please don't do this. I know exactly where the button is.

jakevdp:
Thanks all!

> We can but can we get some freedom on breaking nightly behavior?

Sure, but this is a clear regression (the change in question is a refactor with no changelog entry) so it seemed to me to pretty clearly be an unintentional breakage. Sorry if I'm misinterpreting things.

ev-br:
@jakevdp Thank you for running scipy-nightlies in CI. It's been and continues to be a great source of early-warnings for us.  Hope it'll continue! I suppose a released JAX only depends on released scipy though? IOW, scipy-dev regressions should not hold a jax release, though.

This PR is indeed just a bugfix, I'd think.

>> The number of options is exactly the same whether we have a separate "lower" argument or not.

>Of course not. If you have 4 options to kwarg1 and lower=1, 0 instead of 8 separate lower upper combos you are dealing with four options and choosing where the data lives. I mean can we conclude that part at least?

The actual number of options is exactly the same, `number_of_structures \times 2`, whether they are spelled via two arguments or a single argument. I personally fail to see how having a longer list of allowed string values of `assume_a` is a problem. If anything, the total number of variant is larger for the current setup because of nonsense combinations.

> What is your strong reaction to lower ? 

1. https://github.com/scipy/scipy/pull/23350#issuecomment-3094469477

2. For consistency across `scipy.linalg`, we'd need to introduce it to `inv` and potentially other routines instead of `pos lower` etc.

And yes, I personally consider consistency across `scipy.linalg` to be more valuable than following matlab legacy.

ilayn:
I already provided the answers to those under the linked comment. `lower` is not a matlab thing (matlab does not allow you fiddle with arrays like NumPy does). It's our API and currently in all relevant functions solve_triangular, eigh and so on.

> For consistency across scipy.linalg, we'd need to introduce it to inv and potentially other routines instead of pos lower etc.

Yes because I introduced `assume_a` later on. Previously, inv only handled generic inverses. So the plan is to put it into all, just like batched decorator went into all to complete suite wherever applicable. If you want to keep inv generic that's also an option.

ilayn:
The difference spelled out

```
assume_a = gen, sym, her, pos, tri, hess, diag, bidiag, tridiag
lower = True, False
```

vs. 

```
assume_a = gen, diagonal, tridiagonal, banded, upper triangular,
     lower triangular, sym upper, sym lower, her upper, her lower, pos upper, pos lower
```

In the first one, I don't need to choose lower if I don't care. In the second one, I must choose even if I am not interested in the data location. This is what I mean by inflexible API choice.

ev-br:
The second one is not what I'm discussing though. 
In my mind (and in main for `inv`) it is --- am picking a single structure for clarity: `assume_a  = sym, sym upper, sym lower`, where the "sym" option is exactly as you discuss. In detail:


| a single argument   |    with `lower`      |
|----------|:-------------:|
| `assume_a="sym"` |  `assume_a="sym", lower=default` |
| `assume_a="sym upper"` |  `assume_a="sym", lower=False` |
| `assume_a="sym lower"` |  `assume_a="sym", lower=True` |

So there's no need to specify a triangle if you're happy with the default choice --- this functionality is exactly equivalent.

One potential decision point is what is the default triangle for `assume_a="sym"`. The options are:
1. unspecified; the routine is free to choose either triangle
2. the default is documented.

The second variant is exactly backwards compatible. Starting from scratch, I'd go with the first variant though. But that's a mild preference.

ilayn:
I get that and internally that's quite OK whatever we use. Since we very recently finally got rid of `sym_pos` and `turbo` keywords, I am indeed splitting the hair to not have the same issues in the future. I don't see the string chaining as a good replacement for a lower though. 

Let's say we also implemented the `input_format` option for compressed representation for triangular solve or the packed form for sym/her. Now we want to add this feature in. Take an example; my array is symmetric, and my data is the lower triangle and it's in packed form to save half the space for my 2000x2000 array. 

So currently, it would look 

```python
any_linalg_func(a, ..., assume_a='sym', lower=True, input_format='packed')
```
or 
```python
any_linalg_func(a, ..., assume_a='sym lower', input_format='packed')
```
or, why stop at lower, let's tag on more structure.  
```python
any_linalg_func(a, ..., assume_a='sym lower packed')
```
and if we don't do this one, then why `lower` is special and `input_format` is not becomes the sore point. But it can be symmetric and banded too. So I want to use [dsbev](https://www.netlib.org/lapack/explore-html/df/d26/group__hbev_gac58eb3bf8b79359d6125b0e95dce0efd.html#gac58eb3bf8b79359d6125b0e95dce0efd) then where should I squeeze these options now. But my `input_format` is banded now and not packed. Now we have to parse all options for existence and validity. So it will get harder to place new features and they would start to look wonky when we can't. Or alternatively we have to wait a couple of versions and deprecation cycles etc. Instead we can test the keywords and we would be done. 

For my taste, the last two are overloading the option for no apparent new feature or convenience and creates more edge cases to be parsed such as _'is it just pos or something tagged'_. 

I think clean knobs for each option is the better way. Also it avoids the "you can do it like this but also like that" situation, and provides a single way to do it, Python's zen and whatnot.

ev-br:
There's no need to invoke Zen on Python, not on me at least. As far as sayings go, I'd say "split hairs any finer, we get fission" is apt :-).

Seriously though, I think `any_linalg_func(a, ..., assume_a='sym lower packed')` is the right thing to do. This way we  implement the relevant modes where they are actually needed, instead of having to ignore `a_func(..., assume_a="diagonal", lower=True, format="packed")`. The set of options is going to be sparse, while separate keywords enforce a Cartesian product.

All that said, ISTM our disagreement is pretty clearly articulated at this point. Maybe we should ask somebody to serve as a tie-breaker?

Whichever way we go, we should ideally act in the 1.17 time frame. At a minimum, we should either
- add `lower={True, False}` to `inv`, remove `lower pos` from `inv` and `solve`, consider removing `upper triangular`; or
- start the deprecation of `lower`

ilayn:
If it was a decision such as going left or right, say, write it in C++ or not etc. I would not object because I don't have strong opinions about it so I don't need to stand ground but just argue for a while but you were determined. So, OK have it. But this is unfortunately not that kind of an argument. 

What you are doing here is picking the most sophisticated function that is `solve` that require the most complicated flow chart and combining with `inv` which does a single thing. And beceause `inv` does not have `lower` keyword you are arguing we should remove it from solve and others. For the `transposed` keyword, I lost the argument back when it was implemented and I was arguing against it. Now we want to remove it which is fine I agree with that part. But for the rest, if I have any resident hammer power, that's not what I would prefer to do.

What you are discussing here is quite a disruptive change; for example, 

- It touches almost all `lower` having functions to have deprecation cycles for no apparent (as far as I can understand) benefit that is not otherwise possible. 
- There exists already a well-defined API for all the features that is listed on the initial roadmap for `assume_a` functionality (that closed issue). Even we asked for feedback of the name `assume_a` so it's not like it was just a keyword but a step in that direction. 
- It is bringing string manipulation instead of keywords which is against all other APIs around from say numpy, pandas, to scikits. It is also inherently a very nontrivial thing to parse strings for options compared to simple booleans. I have no examples that I can find this is preferred instead of keywords anywhere.
- It is breaking Array API linalg parts substantially which was brought up every time I tried to touch something on the linalg module for a single thing. This is touching much more stuff but somehow this time there is no resistance or even mention of it whatsoever which I find it a bit strange to be honest. Maybe it is about me and not about Array API.

Given all these items, I think we are not discussing technical parts but personal preferences. However I find this unfair that while I respect other folks/residents plans on other parts of SciPy and concede the arguments when it gets too long be it stats or special or whatever, I have to defend every decision as if my life depends on it lately, in every linalg parts. This is telling me a different story.  

So if that is the case then we don't need to play this resident game and majority wins over everything instead of technical arguments. It can be the preferred way of doing things in Scipy, and such things happen so no problem either way. But at some point, I think I deserve the right to ask "why" and then receive a well-thought out technical argument instead of an essentially "I like it better" answer. Because I like a lot of things that we don't prefer but I am trying to make a case and if not accepted oh well. So at this time, I tried my best in the three other PRs and here and other places to give all the tech arguments but if it is not working then you folks let me know. Because I am starting to drag my feet to even open the notifications because I know I have a truckload of argument to make once again the same things I tried to lay out numerous times.

lucascolley:
> It is breaking Array API linalg parts substantially which was brought up every time I tried to touch something on the linalg module for a single thing. This is touching much more stuff

Would you mind explaining this point for me? Maybe I misunderstand, but in [`xp.linalg`](https://data-apis.org/array-api/draft/extensions/linear_algebra_functions.html), neither `solve` nor `inv` have any kwargs, so I don't see the direct relevance to this discussion. Sorry if this was already explained further up the thread.

ilayn:
It's in one of those issues somewhere that are possibly converted to discussions. So far only cholesky has it, which will be affected by this it seems. But there was a discussion to add some of the keywords to solve, maybe they had a change of heart. I haven't followed the discussions.

But the existing keywords in that standard also don't seem to follow any logic but added by different folks it appears to be. Because that keyword should have been called `lower` to follow suite. But again, mistakes were made.

ev-br:
Okay, after stepping away for several days to cool off--- you don't need to vent an alleged disregard of the "my way of highway" maintainer priviledge on me, Ilhan.
I believe my recent scipy.linalg work paid quite an attention to your preferences, and that we had productive discussions to reach some form of acceptable solutions on points we had disagreements on.
And sorry, calling an argument not a "well though-out technical argument" is not a little insulting. 

Anyhow, I think I'm going to slow down working on scipy.linalg. There is one other enhancement on my TODO list I might get to at some point; my small follow-up PRs are open for scipy.linalg regulars to consider, and potential follow-up issues have been filed---again, for regular or irregular contributors to consider at their own pace. Any further bugs coming from my recent work on scipy.linalg I'm happy to fix of course.

With this, and with all due respect, I bow out of this thread.

ilayn:
I think I am really bad at communicating because I am typing an entire wall of text arguing just to avoid that one thing to get in the discussion to only get exactly the opposite. It is precisely asking for a technical argument only to be taken as insisting on my way. Fair enough.

Moreover, lately, I feel like all I am doing is arguing with SciPy maintainers with quite strong resistance to anything I propose. It leaves me with the conclusion that it must be about me somehow. So please by all means do whatever you see fit. Allow me to finish the fortran translation stuff then we can all get back our lives because there is nothing good coming out of this for me and it feels like I'm doing constantly doing something against the current.

PR #23354, opened 2025-07-18
DEV: Use `sys.executable` everywhere in spin lint

### LINKED ISSUE #23334: DEV: `spin lint`: use `sys.executable` always
I use Python installed from source, with the old `./configure, make, make install` incantation.  This does not create an executable called `python`.  Instead, there is `python3`, and the longer (for example) `python3.13`.  As this is the default behavior for CPython downloaded from python.org, I would expect this to be supported in our `spin` commands.

Most of the `spin` commands have worked fine, but `spin lint` does not.  The `lint()` function in `.spin/cmds.py` starts like this:

```
def lint(ctx, fix, diff_against, files, all, no_cython):
    """🔦 Run linter on modified files and check for
    disallowed Unicode characters and possibly-invalid test names."""
    cmd_prefix = [sys.executable] if sys.platform == "win32" else []
```

So it uses `sys.executable` on only the `win32` platform.  Shouldn't it use that everywhere?  On my system (a Linux workstation), `spin lint` generates an error, because there is no executable called `python`:

```
$ spin lint
$ tools/lint.py --diff-against=main --no-cython
/usr/bin/env: ‘python’: No such file or directory
```

If I change the first line to `cmd_prefix = [sys.executable]`, the command works.

Of course, I can create a symlink somewhere in my command search path for the `python` executable, but that should not be necessary.

kianelbo:
#### Reference issue
Closes #23334 

Use `sys.executable` regardless of platform in spin lint

#### Additional information
Nope!

### DIFF
--- a/.spin/cmds.py
+++ b/.spin/cmds.py
@@ -600,7 +600,7 @@ def authors(ctx_obj, revision_args):
 def lint(ctx, fix, diff_against, files, all, no_cython):
     """🔦 Run linter on modified files and check for
     disallowed Unicode characters and possibly-invalid test names."""
-    cmd_prefix = [sys.executable] if sys.platform == "win32" else []
+    cmd_prefix = [sys.executable]
 
     cmd_lint = cmd_prefix + [
         os.path.join('tools', 'lint.py'),


lucascolley (approved):
thanks @kianelbo !

PR #23357, opened 2025-07-20
ENH, MAINT: linalg: assume_a="diagonal" for solve and inv

ev-br:
Move `assume_a="diagonal"` mode of `scipy.linalg.solve` to C, add this mode to `inv`.

The patch is an enhancement for `inv` and MAINT for `solve`.

### DIFF
--- a/scipy/linalg/src/_linalg_inv.hh
+++ b/scipy/linalg/src/_linalg_inv.hh
@@ -159,6 +159,45 @@ void invert_slice_triangular(
 }
 
 
+// Diagonal array inversion
+template<typename T>
+inline void invert_slice_diagonal(
+    CBLAS_INT N, T *data, SliceStatus& status
+) {
+    using real_type = typename type_traits<T>::real_type;
+    using value_type = typename type_traits<T>::value_type;
+    value_type *pdata = reinterpret_cast<value_type *>(data);
+
+    value_type zero(0.), one(1.);
+    real_type maxa(0.), maxinva(0.);
+
+    for (CBLAS_INT j=0; j<N; j++) {
+        value_type ajj = pdata[j*N + j];
+
+        status.is_singular = (ajj == zero);
+        if (status.is_singular) {
+            status.lapack_info  = j;
+            return;
+        }
+
+        value_type inv_ajj = one / ajj;
+        pdata[j*N + j] = inv_ajj;
+
+        // condition number
+        real_type absa = std::abs(ajj), absinva = std::abs(inv_ajj);
+
+        if(absa != absa) {
+            status.is_ill_conditioned = true;
+        } else {
+            if(absa > maxa) {maxa = absa;}
+            if(absinva > maxinva) {maxinva = absinva;}
+        }
+    }
+    status.is_ill_conditioned = status.is_ill_conditioned || (maxa * maxinva > 1./ numeric_limits<real_type>::eps);
+    status.rcond = maxa * maxinva;
+}
+
+
 template<typename T>
 int
 _inverse(PyArrayObject* ap_Am, T* ret_data, St structure, int lower, int overwrite_a, SliceStatusVec& vec_status)
@@ -288,7 +327,10 @@ _inverse(PyArrayObject* ap_Am, T* ret_data, St structure, int lower, int overwri
             // Get the bandwidth of the slice
             bandwidth(data, n, n, &lower_band, &upper_band);
 
-            if(lower_band == 0) {
+            if ((upper_band == 0) && (lower_band == 0)) {
+                slice_structure = St::DIAGONAL;
+            }
+            else if(lower_band == 0) {
                 slice_structure = St::UPPER_TRIANGULAR;
                 uplo = 'U';
             } else if (upper_band == 0) {
@@ -310,6 +352,19 @@ _inverse(PyArrayObject* ap_Am, T* ret_data, St structure, int lower, int overwri
         init_status(slice_status, idx, slice_structure);
 
         switch(slice_structure) {
+            case St::DIAGONAL:
+            {
+                invert_slice_diagonal(intn, data, slice_status);
+
+                if ((slice_status.lapack_info < 0) || (slice_status.is_singular )) {
+                    vec_status.push_back(slice_status);
+                    goto free_exit;
+                }
+                else if (slice_status.is_ill_conditioned) {
+                    vec_status.push_back(slice_status);
+                }
+                break;
+            }
             case St::UPPER_TRIANGULAR:
             case St::LOWER_TRIANGULAR:
             {
--- a/scipy/linalg/src/_linalg_inv.hh
+++ b/scipy/linalg/src/_linalg_inv.hh
@@ -186,14 +186,10 @@ inline void invert_slice_diagonal(
         // condition number
         real_type absa = std::abs(ajj), absinva = std::abs(inv_ajj);
 
-        if(absa != absa) {
-            status.is_ill_conditioned = true;
-        } else {
-            if(absa > maxa) {maxa = absa;}
-            if(absinva > maxinva) {maxinva = absinva;}
-        }
+        if(absa > maxa) {maxa = absa;}
+        if(absinva > maxinva) {maxinva = absinva;}
     }
-    status.is_ill_conditioned = status.is_ill_conditioned || (maxa * maxinva > 1./ numeric_limits<real_type>::eps);
+    status.is_ill_conditioned = maxa * maxinva > 1./ numeric_limits<real_type>::eps;
     status.rcond = maxa * maxinva;
 }
 
--- a/scipy/linalg/src/_linalg_solve.hh
+++ b/scipy/linalg/src/_linalg_solve.hh
@@ -157,6 +157,47 @@ void solve_slice_sym_herm(
 
 
 
+// Diagonal array solve
+template<typename T>
+inline void solve_slice_diagonal(
+    CBLAS_INT N, CBLAS_INT NRHS, T *data, T *b_data, SliceStatus& status
+) {
+    using real_type = typename type_traits<T>::real_type;
+    using value_type = typename type_traits<T>::value_type;
+    value_type *pdata = reinterpret_cast<value_type *>(data);
+    value_type *p_bdata = reinterpret_cast<value_type *>(b_data);
+
+    value_type zero(0.), one(1.);
+    real_type maxa(0.), maxinva(0.);
+
+    for (CBLAS_INT j=0; j<N; j++) {
+        value_type ajj = pdata[j*N + j];
+
+        status.is_singular = (ajj == zero);
+        if (status.is_singular) {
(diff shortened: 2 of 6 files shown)


ev-br:
Benchmarks: batches of small diagonal matrices get a ~20-50x speedup. A single large matrix gets a bit slower though:


```
| Change   | Before [fc8d4e7f] <main>   | After [9aada1e0] <inv_solve_diagonal>   |   Ratio | Benchmark (Parameter)                                                   |
|----------|----------------------------|-----------------------------------------|---------|-------------------------------------------------------------------------|
| +        | 39.7±1μs                   | 52.7±0.3μs                              |    1.33 | linalg.BatchedSolveBench.time_solve((100, 100), 'diagonal', 'scipy')    |
| -        | 3.26±0.07ms                | 118±2μs                                 |    0.04 | linalg.BatchedSolveBench.time_solve((100, 20, 20), 'diagonal', 'scipy') |
| -        | 3.30±0.03ms                | 58.7±1μs                                |    0.02 | linalg.BatchedSolveBench.time_solve((100, 10, 10), 'diagonal', 'scipy') |
```

Either way, it is way faster than `numpy.linalg.solve`:

```
[50.00%] · For scipy commit 9aada1e0 <inv_solve_diagonal> (round 2/2):
[75.00%] ··· linalg.BatchedSolveBench.time_solve                                                                                                                          ok
[75.00%] ··· =============== =========== ============= ============
             --                                    module          
             --------------------------- --------------------------
                  shape       structure      scipy        numpy    
             =============== =========== ============= ============
              (100, 10, 10)    diagonal     58.7±1μs    125±0.2μs  
              (100, 20, 20)    diagonal     118±2μs     401±0.4μs  
                (100, 100)     diagonal    52.7±0.3μs    89.6±1μs  
             =============== =========== ============= ============

[75.00%] · For scipy commit fc8d4e7f <main> (round 2/2):
[100.00%] ··· linalg.BatchedSolveBench.time_solve                                                                                                                          ok
[100.00%] ··· =============== =========== ============= ============
              --                                    module          
              --------------------------- --------------------------
                   shape       structure      scipy        numpy    
              =============== =========== ============= ============
               (100, 10, 10)    diagonal   3.30±0.03ms   125±0.2μs  
               (100, 20, 20)    diagonal   3.26±0.07ms    397±6μs   
                 (100, 100)     diagonal     39.7±1μs    88.4±0.4μs 
              =============== =========== ============= ============
```

All in all, I think we could merge this as is, and investigate performance of a single-array batch separately: somewhat similar slowdowns has been seen by Tyler in https://github.com/scipy/scipy/pull/23547#pullrequestreview-3503268178 for other modes.

ev-br:
Using `$ py-spy record --native` shows that for `(100,100)`-shaped diagonal matrix the bulk of time is spent in `copy_slice` and `swap_cf`. Applying 

```diff
diff --git a/scipy/linalg/src/_linalg_solve.hh b/scipy/linalg/src/_linalg_solve.hh
index 3971b47f5d..216e8f7eb2 100644
--- a/scipy/linalg/src/_linalg_solve.hh
+++ b/scipy/linalg/src/_linalg_solve.hh
@@ -313,8 +313,9 @@ _solve(PyArrayObject* ap_Am, PyArrayObject *ap_b, T* ret_data, St structure, int
             temp_idx /= shape[i];
         }
         T* slice_ptr = (T *)(Am_data + (offset/sizeof(T)));
-        copy_slice(scratch, slice_ptr, n, n, strides[ndim-2], strides[ndim-1]); // XXX: make it in one go
-        swap_cf(scratch, data, n, n, n);
+        copy_slice_F(data, slice_ptr, n, n, strides[ndim-2], strides[ndim-1]);
 
         // copy the r.h.s, too; XXX: dedupe
```

brings the 100x100 diagonal benchmark to within the noise w.r.t. main:

```
| Change   | Before [fc8d4e7f] <main>   | After [a62c68e6] <inv_solve_diagonal>   |   Ratio | Benchmark (Parameter)                                                   |
|----------|----------------------------|-----------------------------------------|---------|-------------------------------------------------------------------------|
| +        | 31.5±0.6μs                 | 33.3±0.4μs                              |    1.06 | linalg.BatchedSolveBench.time_solve((100, 100), 'diagonal', 'scipy')    |
| -        | 167±4μs                    | 158±3μs                                 |    0.95 | linalg.BatchedSolveBench.time_solve((100, 100), 'gen', 'scipy')         |
| -        | 132±2μs                    | 126±2μs                                 |    0.95 | linalg.BatchedSolveBench.time_solve((100, 100), 'pos', 'scipy')         |
| -        | 2.60±0.04ms                | 74.7±2μs                                |    0.03 | linalg.BatchedSolveBench.time_solve((100, 20, 20), 'diagonal', 'scipy') |
| -        | 2.60±0.06ms                | 40.8±0.2μs                              |    0.02 | linalg.BatchedSolveBench.time_solve((100, 10, 10), 'diagonal', 'scipy') |
```

Rationalizing the way we copy-and-transpose slices gets us deep into the weeds of fast paths (for symmetric modes we don't need to swap to the F order at all; if the input matrix is already C ordered, we can `memcpy` instead of looping etc etc etc). 

Therefore, I'd prefer to put this into 1.17 as is, and make a separate sweep through the copying strategies without the pressures of the impending release branch.

scipy/linalg/src/_linalg_inv.hh
```diff
@@ -159,6 +159,45 @@ void invert_slice_triangular(
...
+            return;
+        }
+
+        value_type inv_ajj = one / ajj;
+        pdata[j*N + j] = inv_ajj;
+
+        // condition number
+        real_type absa = std::abs(ajj), absinva = std::abs(inv_ajj);
+
+        if(absa != absa) {
```
tylerjereddy: Any reason to prefer this over `std::isnan(absa)`?

And in my usual "is this really tested" stance, if I change this to a codepath that can't be hit `spin test -j 8` still passes in my hands on x86-64 Linux:

```diff
-        if(absa != absa) {
+        if(0) {
             status.is_ill_conditioned = true;
```

scipy/linalg/src/_linalg_inv.hh
```diff
@@ -310,6 +352,19 @@ _inverse(PyArrayObject* ap_Am, T* ret_data, St structure, int lower, int overwri
...
 
         switch(slice_structure) {
+            case St::DIAGONAL:
+            {
+                invert_slice_diagonal(intn, data, slice_status);
+
+                if ((slice_status.lapack_info < 0) || (slice_status.is_singular )) {
+                    vec_status.push_back(slice_status);
+                    goto free_exit;
+                }
```
tylerjereddy, line 358: random deletion analysis -- can purge as below the fold and tests still pass with `spin test -j 8`

<details>

```diff
--- a/scipy/linalg/src/_linalg_inv.hh
+++ b/scipy/linalg/src/_linalg_inv.hh
@@ -356,11 +356,7 @@ _inverse(PyArrayObject* ap_Am, T* ret_data, St structure, int lower, int overwri
             {
                 invert_slice_diagonal(intn, data, slice_status);
 
-                if ((slice_status.lapack_info < 0) || (slice_status.is_singular )) {
-                    vec_status.push_back(slice_status);
-                    goto free_exit;
-                }
-                else if (slice_status.is_ill_conditioned) {
+                if (slice_status.is_ill_conditioned) {
```

</details>

scipy/linalg/src/_linalg_solve.hh
```diff
@@ -157,6 +157,47 @@ void solve_slice_sym_herm(
...
+
+        value_type inv_ajj = one / ajj;
+        for (CBLAS_INT i=0; i<NRHS; i++) {
+            p_bdata[j + i*N] *= inv_ajj;
+        }
+
+        // condition number
+        real_type absa = std::abs(ajj), absinva = std::abs(inv_ajj);
+
+        if(absa != absa) {
```
tylerjereddy: I guess this code block has a lot of similarities to the other one changed in `_linalg_inv.hh`, `if(0)` still passes tests in my hands, etc.

I won't repeat the rest of the random deletion analysis a second time. I'm not saying it is wrong, just adding my observations.

scipy/linalg/tests/test_basic.py
```diff
@@ -1473,6 +1473,18 @@ def test_triangular_2(self):
...
+        inv_a = inv(a)
+
+        # basic diagonal invert
+        assert_allclose(inv_a[1], np.diag(1 / np.arange(1, 4)), atol=1e-14)
+
+        # ill-conditioned inputs warn
+        a = np.asarray([[1e30, 0], [0, 1]])
+        with pytest.warns(LinAlgWarning):
+            inv(a, assume_a="diagonal")
+
```
tylerjereddy, line 1506: From the testsuite perspective, this PR seems quite safe for the usual reasons, you've added more testing without mutating old tests.

tylerjereddy (commented):
I added a few small comments and did my usual thing where I delete code and ask why things don't break.

Overall though, it seems fine/pretty safe, and the plan for future performance refinements after branching sounds good to me. Could be nice to get `linalg` regular opinion, but I'd probably be inclined to merge before branching regardless.

scipy/linalg/src/_linalg_inv.hh
ev-br: TIL about `std::isnan`, thank you! 

Writing a test for this branch made me realize it's quite awkward: we check for nans with `check_finite`, so I needed to explicitly switch off the check, and then manually add a nan: `inv([[np.nan, 0], [0, 1], check_finite=False)`. This looks weird, and in the LAPACK-aware solvers we explicitly say that the result is undefined: the exact phrase is _may result in problems (crashes, non-termination) if the inputs do contain infinities or NaNs_. Therefore, the last commit simplifies this branch away. We'll be able to reinstate it if users complain.

ev-br: Aside: your random deletion analysis is great actually. 

AI fuzzers seem to be fashionable these days; making an AI system imitate Tyler's random deletion analysis--- _that_ would be a worthy project. Potentially much more impactful than typical "AI fuzzing" over inputs.

scipy/linalg/src/_linalg_inv.hh
ev-br: Fixed in the updated PR, thanks

scipy/linalg/src/_linalg_solve.hh
ev-br: This branch is gone in the update, together with a matching branch in `inv`

mdhaber:
Can we see what happens with the failure when we update this branch with `main`?

ev-br:
Rebased on main (5bd7eaf6b53a43d0e610f5bcd4b4c8fa5f3785d6) and force-pushed.

tylerjereddy:
Thanks Evgeni, I think I'm happy with this now. It has been open since July without comment from another `linalg` expert, but testing-wise and for benchmarks it seems sensible, and my few comments were addressed.

PR #23362, opened 2025-07-21
MAINT: linalg.{inv, solve}: better error reporting for batched arrays

ev-br:
- Return a rich(-er) structure from the low-level inversion routine, account for e.g. multiple slices being ill-conditioned.

For each "problematic" slice of batched array, we fill in a struct of diagostic values (whether it is singular or ill-conditioned, whether a LAPACK routine detects an internal error etc), collect all such slices into vector, and convert it into a list of dicts, which can be easily manipulated on the python side. python-side handling is minimal ATM--- the main point of this PR is to lay the groundwork, and we'll be able to polish the reporting as we gain some usage.

cross-ref the "make error handling more flexible" and "improve error reporting" bullets of https://github.com/scipy/scipy/issues/23141

- remove `assume_a = "pos lower" from inv and solve. This mode was not released yet, therefore we should remove it before they are frozen by the 1.17 release.

- tighten the memory allocation strategy, plug a potential memory leak.

### DIFF
--- a/scipy/linalg/_basic.py
+++ b/scipy/linalg/_basic.py
@@ -63,6 +63,36 @@ def _find_matrix_structure(a):
     return kind, n_below, n_above
 
 
+def _format_errors(err_lst):
+    """Format/emit errors/warnings from a lowlevel batched routine.
+
+    See inv, solve.
+    """
+    singular, lapack_err, ill_cond = [], [], []
+    for i, dct in enumerate(err_lst):
+        if dct["is_singular"]:
+            singular.append(i)
+        if dct["lapack_info"] < 0:
+            lapack_err.append(f"slice {i} emits lapack info={dct['lapack_info']}")
+        if dct["is_ill_conditioned"]:
+            ill_cond.append(f"slice {i} has rcond = {dct['rcond']}")
+
+    if singular:
+        raise LinAlgError(
+            f"An ill-conditioned matrix detected: slice(s) {singular} are singular."
+        )
+
+    if lapack_err:
+        raise ValueError(f"Internal LAPACK errors: {','.join(lapack_err)}.")
+
+    if ill_cond:
+       warnings.warn(
+            f"An ill-conditioned matrix detected: {','.join(ill_cond)}.",
+            LinAlgWarning,
+            stacklevel=3
+        )
+
+
 def solve(a, b, lower=False, overwrite_a=False,
           overwrite_b=False, check_finite=True, assume_a=None,
           transposed=False):
@@ -250,15 +280,11 @@ def solve(a, b, lower=False, overwrite_a=False,
         return out[..., 0] if b_is_1D else out
 
     # heavy lifting
-    result = _batched_linalg._solve(a1, b1, structure, lower, transposed)
-    x, is_ill_cond, is_singular, info = result
+    x, err_lst = _batched_linalg._solve(a1, b1, structure, lower, transposed)
 
-    if info < 0:
-        raise ValueError("Internal LAPACK error.")
-    if is_singular:
-        raise LinAlgError("A singular matrix detected")
-    if is_ill_cond:
-        warnings.warn("An ill-conditioned matrix detected", LinAlgWarning, stacklevel=2)
+    # emit helpful errors/warnings
+    if err_lst:
+        _format_errors(err_lst)
 
     if b_is_1D:
         x = x[..., 0]
@@ -1422,29 +1448,7 @@ def inv(a, overwrite_a=False, check_finite=True, assume_a=None, lower=False):
 
     # emit helpful errors/warnings
     if err_lst:
-        singular, lapack_err, ill_cond = [], [], []
-        for i, dct in enumerate(err_lst):
-            if dct["is_singular"]:
-                singular.append(i)
-            if dct["lapack_info"] < 0:
-                lapack_err.append(f"slice {i} emits lapack info={dct['lapack_info']}")
-            if dct["is_ill_conditioned"]:
-                ill_cond.append(f"slice {i} has rcond = {dct['rcond']}")
-
-        if singular:
-            raise LinAlgError(
-                f"An ill-conditioned matrix detected: slice(s) {singular} are singular."
-            )
-
-        if lapack_err:
-            raise ValueError(f"Internal LAPACK errors: {','.join(lapack_err)}.")
-
-        if ill_cond:
-           warnings.warn(
-                f"An ill-conditioned matrix detected: {','.join(ill_cond)}.",
-                LinAlgWarning,
-                stacklevel=2
-            )
+        _format_errors(err_lst)
 
     return inv_a
 
--- a/scipy/linalg/_basic.py
+++ b/scipy/linalg/_basic.py
@@ -1324,7 +1324,7 @@ def solve_circulant(c, b, singular='raise', tol=None,
 
 
 # matrix inversion
-def inv(a, overwrite_a=False, check_finite=True, assume_a=None):
+def inv(a, overwrite_a=False, check_finite=True, assume_a=None, lower=False):
     r"""
     Compute the inverse of a matrix.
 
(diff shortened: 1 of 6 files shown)


ev-br:
"One other enhancement" I mentioned over in https://github.com/scipy/scipy/pull/23350 is adding symmetric/hermitian modes to `inv/solve`. IIRC we were all in agreement that these are worth it, and I'm willing and have time to work on these. It'd be helpful to sort this PR out before that. WDYT @ilayn?

ev-br:
Updated to use the same reporting for `inv` and `solve`.

While at it, removed last vestiges of `assume_a="pos lower"`. This should close the book on the discussion at https://github.com/scipy/scipy/pull/23350

ev-br:
A gentle ping. While the main part of this PR is a small clean-up in the interaction between the new-in-1.16 C extension and its python frontend, it also removes `assume_a="pos upper"`, as requested, and adds a new keyword argument to `linalg.inv`. I'm thus a little nervous to self-merge it and would appreciate a second pair of eyes on the patch.

The most important thing IMO is that if we remove `"pos upper"`, we do it before it's released in `scipy==1.17`. To this end, I've attached a milestone.

ev-br:
Updated to fix a memory leak in the new-in-scipy-1-17 code path.

ev-br:
To recap, this PR fixes several small issues with new features added for the 1.17 release:

- fix a memory leak in error handling 
- improve error reporting
- fix user API for symmetric/pos.def. inputs

If someone wants more time to review this, please speak up. Otherwise I plan to land this in a week (Nov 26).

lucascolley:
I'll take a look today 👍

scipy/linalg/_basic.py
```diff
@@ -263,15 +280,11 @@ def solve(a, b, lower=None, overwrite_a=False,
...
 
-    if info < 0:
-        raise ValueError("Internal LAPACK error.")
-    if is_singular:
-        raise LinAlgError("A singular matrix detected")
-    if is_ill_cond:
-        warnings.warn("An ill-conditioned matrix detected", LinAlgWarning, stacklevel=2)
+    # emit helpful errors/warnings
+    if err_lst:
+        _format_errors(err_lst)
```
lucascolley: completely optional nit: we could remove the need for this comment by just making the function name more descriptive? e.g. `_format_emit_errors_warnings` would make the comment obsolete.

scipy/linalg/_basic.py
```diff
@@ -1336,11 +1349,11 @@ def inv(a, overwrite_a=False, check_finite=True, assume_a=None):
...
      upper triangular               'upper triangular'
      lower triangular               'lower triangular'
-     symmetric positive definite    'pos', 'pos upper', 'pos lower'
+     symmetric positive definite    'pos'
     =============================  ================================
 
-    For the 'pos upper' and 'pos lower' options, only the specified
-    triangle of the input matrix is used, and the other triangle is not referenced.
+    For the 'pos' option, only the specified triangle of the input matrix is used, and
+    the other triangle is not referenced.
```
lucascolley:
```suggestion
    For the 'pos' option, only the triangle of the input matrix specified in
    the `lower` argument is used, and the other triangle is not referenced.
```

scipy/linalg/_basic.py
```diff
@@ -1324,7 +1337,7 @@ def solve_circulant(c, b, singular='raise', tol=None,
 
 
 # matrix inversion
-def inv(a, overwrite_a=False, check_finite=True, assume_a=None):
+def inv(a, overwrite_a=False, check_finite=True, assume_a=None, lower=False):
```
lucascolley, line 1339: this seems completely fine to add to me, for consistency. Even _if_ we ended up getting rid of `lower` across the board some day, I don't think this would add a significant amount to the bundle of churn there.

scipy/linalg/_basic.py
```diff
@@ -1360,6 +1373,11 @@ def inv(a, overwrite_a=False, check_finite=True, assume_a=None):
         Valid entries are described above.
         If omitted or ``None``, checks are performed to identify structure so the
         appropriate solver can be called.
+    lower : bool, optional
+        Ignored unless assume_a is one of 'sym', 'her', or 'pos'. If True, the
```
lucascolley:
```suggestion
        Ignored unless `assume_a` is one of 'sym', 'her', or 'pos'. If True, the
```

scipy/linalg/src/_linalg_inv.hh
```diff
@@ -3,18 +3,28 @@
...
+}
+
 // Dense array inversion with getrf, gecon and getri
 template<typename T>
-inline CBLAS_INT invert_slice_general(
+void invert_slice_general(
     CBLAS_INT N, T *data, CBLAS_INT *ipiv, void *irwork, T *work, CBLAS_INT lwork,
-    int* isIllconditioned, int* isSingular
+    SliceStatus& status
 ) {
```
lucascolley, line 19: so rather than returning an `info` status int, we now return nothing here, but modify `status` appropriately?

scipy/linalg/src/_linalg_inv.hh
```diff
@@ -96,31 +107,32 @@ inline CBLAS_INT invert_slice_triangular(
...
         }
     }
-    return info;
 }
 
 
 template<typename T>
-void _inverse(PyArrayObject* ap_Am, T* ret_data, St structure, int overwrite_a, int* isIllconditioned, int* isSingular, CBLAS_INT* info)
+int
+_inverse(PyArrayObject* ap_Am, T* ret_data, St structure, int overwrite_a, SliceStatusVec& vec_status)
```
lucascolley: and this is effectively the reverse, avoiding the need to pass in an `info` object from the call site?

scipy/linalg/src/_linalg_solve.hh
```diff
@@ -147,11 +146,12 @@ void _solve(PyArrayObject* ap_Am, PyArrayObject *ap_b, T* ret_data, St structure
     // --------------------------------------------------------------------
     // Workspace computation and allocation
     // --------------------------------------------------------------------
-    CBLAS_INT intn = (CBLAS_INT)n, int_nrhs = (CBLAS_INT)nrhs, lwork = -1;
+    CBLAS_INT intn = (CBLAS_INT)n, int_nrhs = (CBLAS_INT)nrhs, lwork = -1, info;
 
+    // XXX: workspace query
```
lucascolley, line 151: what does this mean?

lucascolley (approved):
I re-read the discussion in gh-23350.

> fix user API for symmetric/pos.def. inputs

👍 agreed that this is important to get in before 1.17.0. Test changes LGTM.

> improve error reporting

nice improvement and implementation LGTM.

> fix a memory leak in error handling

https://github.com/scipy/scipy/pull/23362/commits/edc08c1ae58f1033ae874ab70f483700c720a56a seems good to me 👍

Thanks Evgeni!

scipy/linalg/src/_linalg_solve.hh
ev-br: LAPACK lingo :-). 
- LAPACK accept work arrays as a pair `double *work, int lwork` --- the data pointer + size. (Fortran details slightly differ, but the idea is the same).
- to find out the optimal work array size, you call with `lwork = -1` and read off the value pointed to by  `*work`.
(see, e.g., https://www.netlib.org/lapack/explore-html/da/d28/group__getri_ga8b6904853957cfb37eba1e72a3f372d2.html)

So the dance is always: query the workspace (lwork=-1, array pointers can be null, shapes need to be correct), then allocate according to what it returned, and call again, now with real arrays.

lucascolley: alright, thanks!

scipy/linalg/src/_linalg_inv.hh
ev-br: Essentially, yes.  Just a small separation of concerns:
Previously, `info` return was overloaded to signal either a malloc error, or something that LAPACK returns. Thus info = -101 could have meant either a failure to allocate one of scratch arrays or a problem in 101-th minor of a matrix.

With this PR, non-zero return means memory error, and linalg errors are all in the vec_status vector.

ev-br:
CI failure are not related, let's merge this as approved. Thanks Lucas for the review!

PR #23374, opened 2025-07-23
DOC: corrected the docstring of NearestNDInterpolator

### LINKED ISSUE #23369: DOC: NearestNDInterpolator works with multi-dimensional arrays but docs suggest otherwise

Currently, the [docs](https://github.com/scipy/scipy/blob/de2e92529ca53ae7f67f493ce88ac086348d285b/scipy/interpolate/_ndgriddata.py#L31) for `NearestNDInterpolator` indicate that the `y` array of values should have shape `(npoints, )` and be a 1-D ndarray, but in fact any multi-dimensional array with shape `(npoints, ...)` will work.

For instance, here's a snippet that works just fine on `scipy==1.15.2`:

```python
from scipy.interpolate import NearestNDInterpolator
import numpy as np
x = np.array([[0., 0.], [1., 0.]])
y = np.array([[1., 0.], [0., 1.]])
interp = NearestNDInterpolator(x, y)
res = interp(x)
```

Printing `res` will correctly show:

```shell
array([[1., 0.],
       [0., 1.]])
```

Unless I'm missing something, the docs should probably be updated to reflect the current behavior:

```python
    """Nearest-neighbor interpolator in N > 1 dimensions.

    ...

    Parameters
    ----------
    x : (npoints, ndims) 2-D ndarray of floats
        Data point coordinates.
    y : (npoints, ...) N-D array of floats or complex
        Data values.
        The length of `y` along the first axis must be equal to
        the length of `x`.

    ...

    """
```

_No response_

SuriyaaMM:
#### Reference issue
closes https://github.com/scipy/scipy/issues/23369

Fixes the docstring accordingly

### DIFF
--- a/scipy/interpolate/_ndgriddata.py
+++ b/scipy/interpolate/_ndgriddata.py
@@ -28,8 +28,9 @@ class NearestNDInterpolator(NDInterpolatorBase):
     ----------
     x : (npoints, ndims) 2-D ndarray of floats
         Data point coordinates.
-    y : (npoints, ) 1-D ndarray of float or complex
-        Data values.
+    y : (npoints, ...) N-D ndarray of float or complex
+        Data values. The length of `y` along the first axis must be equal to
+        the length of `x`.
     rescale : boolean, optional
         Rescale points to unit cube before performing interpolation.
         This is useful if some of the input dimensions have


melissawm (approved):
Looks reasonable to me.

For information @SuriyaaMM : next time, if a PR only touches documentation, you can use `[docs only]` in your commit message to only run docs-related CI checks and not spend extra resources if they are not needed. See https://docs.scipy.org/doc/scipy/dev/contributor/continuous_integration.html#skipping

PR #23375, opened 2025-07-23
ENH: Add ``NdBSpline.derivative`` method

czgdp1807:
#### Reference issue

Coming from second point of, https://github.com/czgdp1807/scipy/pull/21#issuecomment-3067047601,

> `RectBivariateSpline.partial_derivative` : we'd want to add a corresponding method to NdBSpline. Implementation wise AFAICS the low-level routine, `fitpack/pardtc.f`, is equivalent to applying `splder` in a loop over dimensions. This should be done in a separate PR though. The name of the method should probably follow that of NdPPoly, https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.NdPPoly.derivative.html#scipy.interpolate.NdPPoly.derivative

This PR adds a `.derivative()` method to the `NdBSpline` class, enabling evaluation of partial derivatives of B-splines along each axis.

## 🔧 What’s Added

- `NdBSpline.derivative(nu)` method, where `nu` is a list/array of derivative orders per axis.
- Internally uses 1D `BSpline.derivative()` applied along each axis.

## ✅ Tests

Added thorough unit tests to validate correctness of `NdBSpline.derivative()` by comparing results against `RectBivariateSpline.partial_derivative`:

- `test_partial_derivative_method_grid`: full grid-based derivative comparison
- `test_partial_derivative_method`: pointwise (grid=False) diagonal derivative check
- DRY principle applied across tests for clarity and maintainability
- Currently tests only the 2D case using `RectBivariateSpline` for validation
- ~More generic test coverage for **higher-dimensional splines** will follow in subsequent commits.~ I have added the tests for higher dimensions as well.

#### Additional information

cc: @rgommers @ev-br

### DIFF
--- a/scipy/interpolate/_ndbspline.py
+++ b/scipy/interpolate/_ndbspline.py
@@ -246,21 +246,24 @@ class NdBSpline:
 
         return csr_array((data, indices, indptr))
 
-    def _bspline_derivative_along_axis(self, c, t, k, axis):
+    def _bspline_derivative_along_axis(self, c, t, k, axis, nu=1):
         # Move the selected axis to front
         c = np.moveaxis(c, axis, 0)
-        shape = c.shape
-        n = shape[0]
-        trailing_shape = shape[1:]
+        n = c.shape[0]
+        trailing_shape = c.shape[1:]
         c_flat = c.reshape(n, -1)
 
         new_c_list = []
         new_t = None
 
         for i in range(c_flat.shape[1]):
-            b = BSpline(t, c_flat[:, i], k)
-            db = b.derivative(1)
-            db.c = db.c[:len(db.t) - db.k - 1]
+            if k >= nu:
+                b = BSpline.construct_fast(t, c_flat[:, i], k)
+                db = b.derivative(nu)
+                # truncate coefficients to match new knot/degree size
+                db.c = db.c[:len(db.t) - db.k - 1]
+            else:
+                db = BSpline.construct_fast(t, np.zeros(len(t) - 1), 0)
 
             if new_t is None:
                 new_t = db.t
@@ -308,18 +311,10 @@ class NdBSpline:
             if n == 0:
                 continue
 
-            for _ in range(n):
-                k_current = k_new[axis]
-                if k_current < 1:
-                    # No degree left -> derivative is identically zero
-                    c_new[...] = 0.0
-                    k_new[axis] = 0
-                    break  # No need to continue differentiating this axis
-
-                c_new, t_new[axis] = self._bspline_derivative_along_axis(
-                    c_new, t_new[axis], k_current, axis
-                )
-                k_new[axis] = k_current - 1
+            c_new, t_new[axis] = self._bspline_derivative_along_axis(
+                c_new, t_new[axis], k_new[axis], axis, nu=n
+            )
+            k_new[axis] = max(k_new[axis] - n, 0)
 
         return NdBSpline(tuple(t_new), c_new,
                          tuple(k_new), extrapolate=self.extrapolate)
--- a/scipy/interpolate/_ndbspline.py
+++ b/scipy/interpolate/_ndbspline.py
@@ -11,7 +11,7 @@ from . import _dierckx  # type: ignore[attr-defined]
 import scipy.sparse.linalg as ssl
 from scipy.sparse import csr_array
 
-from ._bsplines import _not_a_knot
+from ._bsplines import _not_a_knot, BSpline
 
 __all__ = ["NdBSpline"]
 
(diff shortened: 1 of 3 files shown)


scipy/interpolate/_ndbspline.py
```diff
@@ -246,6 +246,83 @@ def design_matrix(cls, xvals, t, k, extrapolate=True):
 
         return csr_array((data, indices, indptr))
 
+    def _bspline_derivative_along_axis(self, c, t, k, axis):
+        # Move the selected axis to front
+        c = np.moveaxis(c, axis, 0)
+        shape = c.shape
+        n = shape[0]
```
ev-br: nit: `shape` variable can be removed?

scipy/interpolate/_ndbspline.py
```diff
@@ -246,6 +246,83 @@ def design_matrix(cls, xvals, t, k, extrapolate=True):
...
+        trailing_shape = shape[1:]
+        c_flat = c.reshape(n, -1)
+
+        new_c_list = []
+        new_t = None
+
+        for i in range(c_flat.shape[1]):
+            b = BSpline(t, c_flat[:, i], k)
```
ev-br: nit: BSpline.construct_fast?

scipy/interpolate/_ndbspline.py
```diff
@@ -246,6 +246,83 @@ def design_matrix(cls, xvals, t, k, extrapolate=True):
...
+            db = b.derivative(1)
```
ev-br: What is the reason you do not do the derivative of order `nu` at once?

scipy/interpolate/_ndbspline.py
```diff
@@ -246,6 +246,83 @@ def design_matrix(cls, xvals, t, k, extrapolate=True):
...
+            db.c = db.c[:len(db.t) - db.k - 1]
```
ev-br: This is to undo an extension of the `c` array which `BSpline.derivative` does before calling `splder`? Looks awkward, I wonder if we can avoid it, maybe call `splder` directly

ev-br (commented):
I've to admit I do not understand how this works. 

In math-y notation, consider a 2D spline, a linear combination of products of  `1D` basis elements,  $B_i(x)$,

$S(x, y) = \sum_{i, j} c[i, j] B_i(x) B_j(y)$

An application of the 1D differentiation formula to differentiate w.r.t $x$ seems to rely on an identical rewrite,

$S(x, y) = \sum_i \left(  \sum_j c[i, j] B_j(y) \right) B_i(x)$ 

I see now that my original thought to just apply `splder` to the coefficients along an axis was not correct, we need to take $B_j(y)$ into account .
Could you explain a bit if this is indeed what you are doing and how your code relates to it --- of if it's something else, then please explain it?

czgdp1807:
**🔍 Explanation: Derivative Computation Matches B-Spline Derivatives**

This implementation directly follows the mathematical definition of the partial derivative of a tensor-product B-spline.

Given a multivariate spline:

    S(x₁, x₂, ..., x_N) = ∑_{i₁,...,i_N} c[i₁,...,i_N] * B_{i₁}^{(k₁)}(x₁) * ... * B_{i_N}^{(k_N)}(x_N)

The mixed partial derivative of order ν = (ν₁, ν₂, ..., ν_N) is:

    ∂^{|ν|}S / ∂x₁^{ν₁} ∂x₂^{ν₂} ... ∂x_N^{ν_N}
    = ∑_{i₁,...,i_N} c[i₁,...,i_N] * B_{i₁}^{(k₁)(ν₁)}(x₁) * ... * B_{i_N}^{(k_N)(ν_N)}(x_N)

That is, each 1D B-spline basis function is replaced by its derivative of appropriate order, and the coefficients are retained.

---

### ✅ How the code matches this:

1. **Per-axis derivation in `derivative()`**:

    ```python
    for axis, n in enumerate(nu_arr):
        ...
        c_new, t_new[axis] = self._bspline_derivative_along_axis(...)
        k_new[axis] = max(k_new[axis] - n, 0)
    ```

    - Loops over each axis and applies the derivative of order `nu[axis]`
    - Reduces the spline degree per axis accordingly
    - Constructs updated knot vectors and coefficients

2. **Actual derivative calculation in `_bspline_derivative_along_axis(...)`**:

    ```python
    b = BSpline.construct_fast(t, c_flat[:, i], k)
    db = b.derivative(nu)
    ```

    - For each fixed slice orthogonal to the target axis, we:
      - Construct a 1D B-spline `b(x)`
      - Compute its derivative of order `nu` directly using `BSpline.derivative(nu)`
      - This returns a new spline object `db` with:
        - updated degree: `k - nu`
        - updated knot vector
        - transformed coefficients
    - We truncate `db.c` manually to `len(db.t) - db.k - 1` to ensure consistency with `construct_fast`.

3. **Edge case handling**:

    ```python
    if k < nu:
        db = BSpline.construct_fast(t, np.zeros(len(t) - 1), 0)
    ```

    - If the degree is less than the requested derivative order (e.g. `k=1`, `nu=2`), the result is identically zero.
    - We replace the basis with a degree-0 spline filled with zeros to reflect this.

---

### 🎯 Final result:

The returned `NdBSpline` object represents:

    S_derived(x₁, ..., x_N) = ∑ c'[i₁,...,i_N] * B_{i₁}^{(k₁ - ν₁)}(x₁) * ... * B_{i_N}^{(k_N - ν_N)}(x_N)

which is the correct tensor-product spline representing the mixed partial derivative of order `nu`.

scipy/interpolate/_ndbspline.py
czgdp1807: I tried it. We have to anyways first pad the coefficients and send it to `splder` (which `BSpline.derivative` already does). Then `splder` returns padded coefficients, we have to truncate them which we are doing. I think calling `splder` doesn't result in any difference from what we are doing here.

scipy/interpolate/_ndbspline.py
```diff
@@ -246,6 +246,78 @@ def design_matrix(cls, xvals, t, k, extrapolate=True):
...
+        nu : int or array_like of shape (ndim,)
+            Order(s) of the derivative to compute along each dimension.
+            A single integer means the same order for all dimensions.
+
+        Returns
+        -------
+        NdBSpline
+            A new NdBSpline representing the derivative.
+        """
+        nu_arr = np.atleast_1d(nu)
```
ev-br: These two should be consistent:

```
(Pdb) bspl2(xi, nu=42)
*** ValueError: invalid number of derivative orders nu = array(42) for ndim = 2.
(Pdb) bspl2.derivative(42).k
(np.int64(0), np.int64(0))
```

Here `bspl2` is from https://github.com/scipy/scipy/issues/23510 (which is separate).

scipy/interpolate/_ndbspline.py
```diff
@@ -246,6 +246,78 @@ def design_matrix(cls, xvals, t, k, extrapolate=True):
...
+
+        return new_c, new_t
+
+    def derivative(self, nu=1):
+        """
+        Construct a new NdBSpline representing the derivative.
+
+        Parameters
+        ----------
+        nu : int or array_like of shape (ndim,)
```
ev-br: So here you follow the convention for `k`, where `k=3` means "cubics in all dimensions".

For a user in me,  it's not immediately obvious that `spl.derivative(1)` means a derivative of order `d` for a `d`-dimensional spline, so let's make it stricter and always require that `nu` is an array-like of the correct length, similar to the in-place derivatives.

If this proves limiting, we can generalize and allow a single integer; the reverse is harder, so let's start with a stricter API.

ev-br (changes requested):
Thanks @czgdp1807 . I've a nagging feeling this can be simplified a bit by using `splder` directly instead of constructing `BSpline` objects per dimension, but if it's not immediately obvious, as you say, then it's not worth spending time on. 

There's a small API suggestion below. Once it's addressed, I think this is good to go in.

czgdp1807:
@ev-br I have addressed your review. Please take a look and let me know if I have done the right thing. I think what you suggested is `nu` should be an array of length equal to number of dimensions in `NdBSpline`. Both reviews were related to this, right?

ev-br:
OK, there's still a small hiccup for non-integer values of `nu`:

```
(Pdb) bspl2.derivative([1, 2.5])(xi)
array([ 11.76,  37.5 , 121.5 ])
(Pdb) bspl2(xi, nu=[1, 2.5])
array([ 11.76,  37.5 , 121.5 ])
(Pdb) bspl2(xi, nu=[1, 2])
array([ 11.76,  37.5 , 121.5 ])
```

Since this is pre-existing and consistent between in-place and out-of-place `derivative`, we can ignore this for now. 

One other thing: the new `.derivative` method is public. It thus should be listed in the `Methods` section of the docstring, https://github.com/scipy/scipy/pull/23375/files#diff-9f498f78b1600784ab26e11e416a730d110c761749fcdb09d8cf43f8357fb0b4R70, and also should get a release note snippet. 

Let's not hold this PR for these minor things, I'll see to these two now. Merging, thanks @czgdp1807 and keep them coming!

PR #23187, opened 2025-06-19
CI: add cp314/cp314t nighly wheel builds

rgommers:
We're about 5-6 weeks away from needing to do a release with Python 3.14 support (see the release schedule in [PEP 745](https://peps.python.org/pep-0745/)), so it'll be good to enable nightly wheels now and fix/report issues. 

There are a couple of tests that needed skipping - issues filed for those: gh-23172, gh-23185. Also, there is one wheel (`cp314t-musllinux_1_2_aarch64`) for which the test suite is hanging. The hang is the most serious problem and needs investigating next. That wheel is probably the least critical of all wheels though, so it's best to get the other ones deployed now.

We shouldn't backport this PR just yet; once 3.14.0rc1 is out and numpy has done a release that supports it, this will be significantly simpler to support on the `maintenance/1.16.x` branch.

PR #23380, opened 2025-07-24
CI: bump to cibuildwheel 3.1.0 for 3.14.0rc1, add cp314t musllinux wheels

rgommers:
This follows up on gh-23187, adding the last two missing wheel builds for Python 3.14 and building against 3.14.0rc1.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -215,7 +215,7 @@ jobs:
           echo "CIBW_BUILD_FRONTEND=$CIBW" >> "$GITHUB_ENV"
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@5f22145df44122af0f5a201f93cf0207171beca7 # v3.0.0
+        uses: pypa/cibuildwheel@ffd835cef18fa11522f608fc0fa973b89f5ddc87 # v3.1.0
         env:
           CIBW_BUILD: ${{ matrix.python[0] }}-${{ matrix.buildplat[1] }}*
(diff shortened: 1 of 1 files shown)


rgommers:
Note that one more thing is needed for 3.14: after NumPy 2.3.2 is out with `cp314` wheels on PyPI, remove the workarounds for pulling in nightly wheels. Can be done in this PR or in a follow-up, depending on timing of the `numpy` release.

tylerjereddy (approved):
CI looks happy, and I spot checked that `3.14.0rc1` is indeed being used in a few logs that I randomly sampled from the matrix.

I'll go ahead and put this in. One question I had was the original gh-23187 has the comment below, if you have advice on that it may be helpful, otherwise I'll probably find out when noticing merge conflicts or whatever:

> We shouldn't backport this PR just yet; once 3.14.0rc1 is out and numpy has done a release that supports it, this will be significantly simpler to support on the maintenance/1.16.x branch.

Presumably it just means I can be a bit selective in backporting only parts of that without certain shims, or something like that.

tylerjereddy:
thanks Ralf

rgommers:
> Presumably it just means I can be a bit selective in backporting only parts of that without certain shims, or something like that.

That first gh-23187 PR had a lot of code under `- name: Inject environment variable for python dev versions`, those changes can all be reverted once `numpy` does a release. Also the `- exclude` part of this PR is a revert of part of gh-23187. So the easiest will be to wait until the last PR goes into `main` and then cherry-pick them all at once (perhaps even squash them together) - the diff will be quite small.

PR #23381, opened 2025-07-24
MAINT: stats: Apply xp_capabilities decorator to all public functions

### LINKED ISSUE #23412: TST: stats: `marray` tests failing with latest array-api-strict
e.g.

```
/opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/site-packages/array_api_strict/_array_object.py:959: in __setitem__
    other = self._promote_scalar(value)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
        key        = Array(True, dtype=array_api_strict.bool)
        np_key     = array(True)
        other      = False
        self       = Array(0.22652482, dtype=array_api_strict.float64)
        value      = False
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = Array(0.22652482, dtype=array_api_strict.float64), scalar = False

    def _promote_scalar(self, scalar: complex) -> Array:
        """
        Returns a promoted version of a Python scalar appropriate for use with
        operations on self.
    
        This may raise an OverflowError in cases where the scalar is an
        integer that is too large to fit in a NumPy integer dtype, or
        TypeError when the scalar type is incompatible with the dtype of self.
        """
        from ._data_type_functions import iinfo
    
        target_dtype = self.dtype
        # Note: Only Python scalar types that match the array dtype are
        # allowed.
        if isinstance(scalar, bool):
            if self.dtype not in _boolean_dtypes:
>               raise TypeError(
                    "Python bool scalars can only be promoted with bool arrays"
E                   TypeError: Python bool scalars can only be promoted with bool arrays

iinfo      = <function iinfo at 0x7f13b953f6a0>
scalar     = False
self       = Array(0.22652482, dtype=array_api_strict.float64)
target_dtype = array_api_strict.float64
```

cc @ev-br @mdhaber

steppi:

This PR applies the `xp_capabilities` decorator to all public functions and function-likes (named non-type callables) in `scipy.stats`. I added a test to `scipy._lib.tests.test_array_api` to check for presence of the decorator which now passes for `scipy.stats`. To create this PR a followed a recipe given by @crusaderky:

1. Temporarily tamper with xp_capabilities to always skip array_api_strict; then run `python dev.py tests -b array_api_strict -m array_api_backends` to see what is still running. Proceed to apply `@xp_capabilities` and `@make_xp_test_case` and remove the explicit `@skip_xp_backends` and `@xfail_xp_backends` to everything that crops up.
2. Review the surviving explicit `skip_xp_backends` and `xfail_xp_backends` . They should only apply to edge cases and not to the general functionality of a public function.
3. Investigate `__all__` in the scipy submodules to see if you can extract an inventory of everything that isn't decorated without too much noise.

#### Additional information

I'm marked this at draft because it branches off from #23081, the PR adding tables summarizing alternative backend support, not main, so #23081 should go in first. GPU coverage for `stats` is now up to 33% for CuPy and Jax and 28% for PyTorch. The slightly lower percentage for PyTorch is due to PyTorch missing the `betainc` function.

I marked functions as out of scope if they take distributions from the old infrastructure as input. I didn't mark transform functions from the new infrastructure like `stats.exp`, `stats.log`, `stats.abs` as out-of-scope, because the intention is to make distributions from the new infrastructure work across backends, so even though these functions don't take arrays as inputs, I think it still makes sense to consider them in-scope. This current work is only for functions and function-like callables, and there still needs to be remaining work to handle classes (like distributions from the new infra) that we'd to consider in-scope.

### DIFF
--- a/scipy/stats/tests/test_marray.py
+++ b/scipy/stats/tests/test_marray.py
@@ -1,15 +1,31 @@
-import pytest
 import numpy as np
+import pytest
 from scipy import stats
+from packaging import version
 
 from scipy._lib._array_api import xp_assert_close, xp_assert_equal, _length_nonmasked
+from scipy._lib._array_api import make_xp_pytest_param, make_xp_test_case
+from scipy._lib._array_api import SCIPY_ARRAY_API
 from scipy.stats._stats_py import _xp_mean, _xp_var
 from scipy.stats._axis_nan_policy import _axis_nan_policy_factory
 
 
 marray = pytest.importorskip('marray')
-skip_backend = pytest.mark.skip_xp_backends
+pytestmark = [
+    pytest.mark.skipif(
+        not SCIPY_ARRAY_API,
+        reason=(
+            "special function dispatch to marray required for these tests"
+            " is hidden behind SCIPY_ARRAY_API flag."
+        ),
+    ),
+    pytest.mark.skip_xp_backends(
+        "array_api_strict",
+        reason="scipy/#23412",
+    ),
+]
 
+skip_backend = pytest.mark.skip_xp_backends
 
 def get_arrays(n_arrays, *, dtype='float64', xp=np, shape=(7, 8), seed=84912165484321):
     mxp = marray._get_namespace(xp)
@@ -39,9 +55,9 @@ def get_arrays(n_arrays, *, dtype='float64', xp=np, shape=(7, 8), seed=849121654
 @skip_backend('dask.array', reason='Arrays need `device` attribute: dask/dask#11711')
 @skip_backend('jax.numpy', reason="JAX doesn't allow item assignment.")
 @skip_backend('torch', reason="marray#99")
-@pytest.mark.parametrize('fun, kwargs', [(stats.gmean, {}),
-                                         (stats.hmean, {}),
-                                         (stats.pmean, {'p': 2})])
+@pytest.mark.parametrize('fun, kwargs', [make_xp_pytest_param(stats.gmean, {}),
+                                         make_xp_pytest_param(stats.hmean, {}),
+                                         make_xp_pytest_param(stats.pmean, {'p': 2})])
 @pytest.mark.parametrize('axis', [0, 1])
 def test_xmean(fun, kwargs, axis, xp):
     mxp, marrays, narrays = get_arrays(2, xp=xp)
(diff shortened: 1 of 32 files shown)


mdhaber:
gh-20544 has some thoughts about what else might be out of scope - see bottom of top post.

steppi:
> [gh-20544](https://github.com/scipy/scipy/issues/20544) has some thoughts about what else might be out of scope - see bottom of top post.

Thanks @mdhaber. I'll have to check all of those to make sure we're being consistent. The below is what I wrote in the documentation regarding what to consider out-of-scope

---

There is not yet a formal policy for which functions should be considered out-of-scope for alternative backend support. Some general rules of thumb that are being followed are to exclude:

* functions which do not operate on arrays such as [scipy.constants.value](https://output.circle-artifacts.com/output/job/112c7244-f3c5-4700-87e9-c268c1617d7c/artifacts/0/html/reference/generated/scipy.constants.value.html)

* functions which are too implementation specific such as those in [scipy.linalg.blas](https://output.circle-artifacts.com/output/job/112c7244-f3c5-4700-87e9-c268c1617d7c/artifacts/0/html/reference/linalg.blas.html#module-scipy.linalg.blas) which give direct wrappers to low-level BLAS routines.

* functions which would inherently be very difficult or even impossible to compute efficiently on accelerated computing devices.

As an example. The contents of [scipy.odr](https://output.circle-artifacts.com/output/job/112c7244-f3c5-4700-87e9-c268c1617d7c/artifacts/0/html/reference/odr.html#module-scipy.odr) are considered out-of-scope for a combination of reasons 2 and 3 above. [scipy.odr](https://output.circle-artifacts.com/output/job/112c7244-f3c5-4700-87e9-c268c1617d7c/artifacts/0/html/reference/odr.html#module-scipy.odr) essentially provides a direct wrapper of the monolithic ODRPACK Fortran library, and it’s API is tied to the structure of this monolithic library. Creation of an efficient GPU accelerated implementation of nonlinear weighted orthogonal distance regression is also a challenging problem in its own right. Nevertheless, considerations of what to consider in-scope are evolving, and something which is now considered out-of-scope may be decided to be in-scope in the future if sufficient user interest and feasability are demonstrated.

---

I think the NASA ROSES grant deliverables offer some leeway on not having to cover everything which could be potentially considered in-scope. My thought for instance is that `scipy.odr` is mainly out-of-scope because if we did have a capable GPU accelerated `odr` implementation, it would be a waste of time to try to preserve the `ODRPACK` API. `odr` that fits into the `scipy.optimize` API  would probably be reasonable to consider in-scope though, but wouldn't be prioritized because it's difficult.

scipy/stats/tests/test_marray.py
```diff
@@ -1,8 +1,8 @@
 import pytest
-import numpy as np
 from scipy import stats
 
 from scipy._lib._array_api import xp_assert_close, xp_assert_equal, _length_nonmasked
+from scipy._lib.array_api_compat import numpy as np
```
lucascolley: @mdhaber should this be necessary for `marray`? What went wrong without this @steppi ?

lucascolley (approved):
thanks @steppi, the diff looks good to me!

@mdhaber did you want to weigh in any further on what is marked out-of-scope before we merge this? Or shall we save that for later?

scipy/stats/tests/test_marray.py
steppi: I've been running a very thorough grid of tests locally. These tests were failing in the oldest supported NumPy because `np.astype` isn't available there. I don't think we have anything in CI which tests `marray` on NumPy < 2. This was just something I did without thinking to get the local tests passing. Maybe it should be rethought. It at least deserves a comment.

steppi: Comment added

lucascolley: I would rather use an alternative to `astype` than stop testing bare `np` here, although this is certainly unimportant enough to be left for an optional follow-up

steppi: I'll change it here. I was just cargo culting something something I saw in the special alternative backends test.

steppi:
@lucascolley, I've gone back to using bare `np`. I just used `ndarray.astype` instead of `np.astype` for the case mentioned earlier. There was also another test which failed for older NumPy versions when using bare `np`, but I just added `skipif` for it. 

I noticed something though. Here, and on `main`, the `marray` tests run and fail if `marray` is installed and `SCIPY_ARRAY_API` is unset. They should instead be skipped. I can take care of this here or in a separate PR. @mdhaber, had you seen this before?

steppi:
Also, I found I was missing a bunch of `make_xp_test_case` and/or `make_xp_pytest_param` decorators in `test_marray.py`. Added now.

steppi:
It looks like I accidentally messed up the diff trying to clean up history. I will fix it tomorrow.

steppi:
Oh, I was wrong about the diff being corrupted. Late last night I saw that this PR touched `scipy/_lib/_array_api_docs_tables.py` and assumed I must have broken something, because I had been switching between many different branches, and stashing and popping repeatedly. I hadn't checked, but my assumption was that I had some uncommitted changes locally, and that's why tests pass locally but not on CI.

I forgot that the PR actually does change something in `_array_api_docs_tables.py`. The actual problem comes out of the update from `array_api_strict=2.4.0` to `2.4.1`. I can reproduce locally using `2.4.1`. Investigating now.

steppi:
Oh, what a bizarre coincidence. These failures are appearing elsewhere. They are not even related to this PR.

lucascolley:
yep, unfortunate coincidence! I opened an issue for that already

steppi:
> yep, unfortunate coincidence! I opened an issue for that already

Thanks @lucascolley! I was about to submit one myself. I think this is fine to merge then.

steppi:
I pushed a commit that makes the `marray` tests get skipped if `SCIPY_ARRAY_API=1` isn't set.

steppi:
I guess until the underlying issue is fixed, I can `skip` or `xfail` those `marray` tests here to get rid of the noise in CI.

lucascolley:
1 CI failure left to skip or fix, then we can merge this after a full run is green

steppi:
> 1 CI failure left to skip or fix, then we can merge this after a full run is green

I fixed the unrelated failure in `spatial.transform`. Before merging, I'll interactive rebase this into two commits. One for stats xp_capabilities stuff, and one for the spatial.transform test fix.

steppi:
Cool. Everything passed. I'll squash all the stats stuff together into one commit, and then this can be regular merged as two commits, with the `spatial.transform` commit separate.

PR #23384, opened 2025-07-25
CI: disable special-casing of 3.14-dev now that Python 3.14rc1 is out

rgommers:
This reverts changes made in scipy#23187, keeping the commented out code for once 3.15-dev arrives.

Tested on my fork [here](https://github.com/rgommers/scipy/actions/runs/16522174701).


rgommers:
@tylerjereddy this should be the last one, I think after merging this and then backporting gh-23187, gh-23380 and this PR together will get `1.16.x` into good shape.

tylerjereddy (approved):
thanks Ralf!

Issue #8410, opened 2018-02-12
odeint sometimes skips critical points


jcalbert:
While working on #7379 I found a bigger problem with `odeint`.  It only uses `tcrit` to prevent `LSODA` from overshooting points in `t` by enough to also go past `tcrit`, but doesn't make the `LSODA` integrator stop at internal critical points which aren't near points in `t`.

The example below uses integrating `y'(t) = max( 0, 1-t )` from `0<t,20`, which has a critical point at `t=1`

### Reproducing code example:
```
yprime = lambda y,t: max(0,1-t)
y0 = np.zeros(1)
times_coarse = np.linspace(0,20,5)
times_fine = np.linspace(0,20,21)
tcrit = np.ones(1)
res_fine = odeint(funk, y0, times_fine, tcrit=tcrit, full_output=1)
res_coarse = odeint(funk, y0, times_coarse, tcrit=tcrit, full_output=1)
assert np.allclose(res_fine[0][-1], res_coarse[0][-1])
```

### Scipy/Numpy/Python version information:
```
('1.0.0', '1.14.0', sys.version_info(major=2, minor=7, micro=13, releaselevel='final', serial=0))
```

As before, I'm fairly sure the problem is in the loop at lines 696-736 of `integrate/_odepackmodule.c`

it repeatedly calls `LSODA` with `itask=4`, which essentially means "advance the integrator past t, but without passing tcrit, and return a back-interpolated value for y(t)" but it *should* sometimes call `LSODA` just to get to the next `tcrit`, which would look something like do something like:

```
i = 0
j = 0
y_out = np.empty(len(t))
while i < len(t):
    if tcrit[j] < t[i]:
        ADVANCE INTEGRATOR TO tcrit[j]
        j+= 1
        continue
    elif tcrit[j] == t[i]:
        j+= 1

    ADVANCE INTEGRATOR TO t[i] AND SAVE TO y_out[i]
    i+=1
```

v-ber:
I have a similar issue. Is there any update? Thanks in advance! BR

tylerjereddy:
I'll try adding a milestone to this to see if it helps from slipping under the radar..

h-vetinari:
> I'll try adding a milestone to this to see if it helps from slipping under the radar..

Apparently it didn't 🙈

I've remilestoned it to the next release.

ev-br:
Bumping the milestone once again.

Issue #7379, opened 2017-05-06
odeint (LSODA) fails if two t_crits occurs between two t_outs


jcalbert:
I've found that in running integrate's `odeint`, LSODA will fail if, given an array of critical points `tcrit` and output times `t`,  there exists m, n such that `t[m] <= tcrit[n] < tcrit[n+1] < t[m+1]`.  That is, if there are two critical points within a single span of t.

### Reproducing code example:
```
import numpy as np
from scipy.integrate import odeint

def deriv(y,t):
    return np.zeros_like(y) + np.floor(t)

tc = np.arange(0,3,1.)
    

t1 = np.arange(0., 3.0, .5)
t2 = np.arange(0., 3.0, 1.5)

y0 = np.zeros(1)

for t in [t1, t2]:
    print 't = ' + str(t)
    r, e = odeint(deriv, y0, t, tcrit = tc, full_output=1)
    print e['message'] + '\n'

```

### Error message:
```
t = [ 0.   0.5  1.   1.5  2.   2.5]
Integration successful.

t = [ 0.   1.5]
 lsoda--  itask = 4 or 5 and tcrit (=r1) behind tout (=r2)   
      in above,  r1 =  0.1000000000000D+01   r2 =  0.1500000000000D+01
<path>/scipy-dev/local/lib/python2.7/site-packages/scipy-1.0.0.dev0+fe4ee43-py2..
  warnings.warn(warning_msg, ODEintWarning)
Illegal input detected (internal error).

```

### Scipy/Numpy/Python version information:
```
<<('1.0.0.dev0+fe4ee43', '1.12.1', sys.version_info(major=2, minor=7, micro=12, releaselevel='final', serial=0))>>
```

I'm fairly certain the problem is in lines 697-710 of `integrate/_odepackmodule.c`

```
        tout_ptr = tout + k;
        /* Use tcrit if relevant */
        if (itask == 4 && *tout_ptr > *(tcrit + crit_ind)) {
            crit_ind++;
            rwork[0] = *(tcrit+crit_ind);
        }
        if (crit_ind >= numcrit) {
            itask = 1;  /* No more critical values */
        }

        LSODA(ode_function, &neq, y, &t, tout_ptr, &itol, rtol, atol, &itask,
              &istate, &iopt, rwork, &lrw, iwork, &liw,
              ode_jacobian_function, &jt);
```
[...]

PR #23385, opened 2025-07-25
ENH: integrate.solve_ivp: critical time support 

### LINKED ISSUE #14246: DiscreteEvents: solve_ivp should evaluate fun() at specific, user-specified timepoints
I'd like to force solve_ivp to evaluate the ODE RHS `fun()` at specific timepoints, in order to implement the adjoint method (as in [this paper](https://arxiv.org/pdf/1806.07366.pdf)). 

- Registering an event at specific timepoints with `solve_ivp` returns the ODE solution at those timepoints using interpolation, and skips evaluating `fun()` at those points. 
- Similarly, supplying `t_eval` to `solve_ivp` at those points _does not actually evaluate fun() at those t_, but instead, gives you back a solution with the interpolated values at those points. 

I'd like to be able to specify a set of timepoints at which the solver _must_ evaluate `fun()`.

MatthewFlamm:
#### Reference issue

Likely closes #14246

This PR implements critical time support for `solve_ivp`. The detailed scipy roadmap currently says

> Complete the feature set of solve_ivp, including all functionality of the odeint interface.

One of the features of `odeint` is to provide a `tcrit` input parameter as

>tcrit: ndarray, optional
>> Vector of critical points (e.g., singularities) where integration care should be taken.

In `odeint` this means that the solution should integrate to exactly those points, then proceed further.

#### Additional information
Currently the end point of simulation is being used as a critical time point in the LSODA solver.  This PR modifies the logic to progressively step through a series of critical points.  A test is added that ensures this point is hit exactly.

~This feature is likely possible to implement for all the solve_ivp solvers, e.g. RK could limit the next time step similar to the `t_bound` check below.  While I'm willing to implement for all solvers, I wanted to get feedback on this approach before spending that time.~

This section has now been modified for all solvers.

https://github.com/scipy/scipy/blob/fbb6ff839cfdf02304a3bd21d46a1189c01832e9/scipy/integrate/_ivp/rk.py#L143-L144

#### Areas of feedback
- Should the monotonicity of tcrit be checked explicitly or assumed to be correct (current behavior)?
- ~The current code silently excludes the critical times outside the interval of integration, should it raise an error or warning instead?~
- ~Implement for all other solvers?~
  - ~If so, move common tcrit logic to ODESolver~

### TODO

These items depend on the feedback above.

- [ ] Add tests for tcrit sequences that are non-monotonically increasing or decreasing ~or fall outside t0 or tbound.~
- [x] Add tests for the tcrit attribute on LSODA class.
- [x] use for other solvers


scipy/integrate/_ivp/lsoda.py
```diff
@@ -207,6 +227,38 @@ def _dense_output_impl(self):
...
+            tcrit = tcrit[tcrit>t0]
+            # remove critical times after end time
+            tcrit = tcrit[tcrit<t_bound]
+        else:
+            # should be unreachable keep?
+            raise ValueError("Unexpected direction")
+        self.tcrit = np.append(tcrit, t_bound)
+
+    def _find_next_tcrit(self):
+        if self.tcrit.size == 0:
```
MatthewFlamm: From a performance perspective, these extra if statements will give extra overhead.  It would be better if the minimum length of `tcrit` is `1`, i.e. `t_bound` is always there.  Then we only need to handle this extra logic when a critical point is hit, which is infrequent.

MatthewFlamm:
Since the implementation of `LSODA` and `solve_ivp` do not specify intermediate time points of integration, the issues https://github.com/scipy/scipy/issues/8410 and https://github.com/scipy/scipy/issues/7379 should not be relevant for `LSODA` and `solve_ivp` in this implementation.

scipy/integrate/_ivp/lsoda.py
MatthewFlamm: changed in https://github.com/scipy/scipy/pull/23385/commits/a96c26db331c01aba1a15f433865aac951053983

ilayn:
Ah dangit, I hope I did not create too much hassle for your work here with the rewrite. I should have checked before, apologies.

MatthewFlamm:
No problem, the failing tests look unrelated, I assume that the new c version is working correctly since the tests in this PR are passing.  I still need to go check the new c version closely and whether there is some unexpected edge case that could arise.

MatthewFlamm:
In a look, it seems it should be like for like for this functionality, so I don't think the rewrite impacts this PR.

scipy/integrate/_ivp/tests/test_ivp.py
```diff
@@ -1293,3 +1293,35 @@ def test_inital_maxstep():
                                             method_order,
                                             rtol, atol)
             assert_equal(max_step, step_with_max)
+
+
+@pytest.mark.thread_unsafe(reason="lsoda solver is not thread-safe")
```
MatthewFlamm: After #23914, lsoda is now thread safe

```suggestion
```

MatthewFlamm:
This PR was extended to add the functionality to all solvers in `solve_ivp`.  I updated the PR description and title accordingly.

There is a lot of common functionality in the Python solvers that could be considered for refactoring, but I kept it simple for the PR and implemented the same logic in each.

scipy/integrate/_ivp/base.py
```diff
@@ -101,6 +101,15 @@ class OdeSolver:
...
         Default is False.
+    tcrit : float and array_like, optional
+        Critical points to take care during integration.  Forces
+        solver to integrate to this time point exactly before proceeding.
+        If an array of values is passed in, the solver will treat each
+        value as critical. The array of values must be sorted either
+        ascending or descending in the same manner as the direction
+        between ``t0`` and ``t_bound``.
+
+        .. versionadded:: 1.17.0
```
MatthewFlamm:
```suggestion
        .. versionadded:: 1.18.0
```

scipy/integrate/_ivp/bdf.py
```diff
@@ -155,6 +155,15 @@ class BDF(OdeSolver):
...
         execution overall in some circumstances (e.g. small ``len(y0)``).
+    tcrit : float and array_like, optional
+        Critical points to take care during integration.  Forces
+        solver to integrate to this time point exactly before proceeding.
+        If an array of values is passed in, the solver will treat each
+        value as critical. The array of values must be sorted either
+        ascending or descending in the same manner as the direction
+        between ``t0`` and ``t_bound``.
+
+        .. versionadded:: 1.17.0
```
MatthewFlamm:
```suggestion
        .. versionadded:: 1.18.0
```

scipy/integrate/_ivp/ivp.py
```diff
@@ -360,6 +360,15 @@ def solve_ivp(fun, t_span, y0, method='RK45', t_eval=None, dense_output=False,
...
         By default `min_step` is zero.
+    tcrit : float and array_like, optional
+        Critical points to take care during integration.  Forces
+        solver to integrate to this time point exactly before proceeding.
+        If an array of values is passed in, the solver will treat each
+        value as critical. The array of values must be sorted either
+        ascending or descending in the same manner as the direction
+        defined by ``t_span``.
+
+        .. versionadded:: 1.17.0
```
MatthewFlamm:
```suggestion
        .. versionadded:: 1.18.0
```

scipy/integrate/_ivp/radau.py
```diff
@@ -258,6 +258,15 @@ class Radau(OdeSolver):
...
         execution overall in some circumstances (e.g. small ``len(y0)``).
+    tcrit : float and array_like, optional
+        Critical points to take care during integration.  Forces
+        solver to integrate to this time point exactly before proceeding.
+        If an array of values is passed in, the solver will treat each
+        value as critical. The array of values must be sorted either
+        ascending or descending in the same manner as the direction
+        between ``t0`` and ``t_bound``.
+
+        .. versionadded:: 1.17.0
```
MatthewFlamm:
```suggestion
        .. versionadded:: 1.18.0
```

scipy/integrate/_ivp/rk.py
```diff
@@ -339,6 +341,15 @@ class RK45(RungeKutta):
...
         Whether `fun` is implemented in a vectorized fashion. Default is False.
+    tcrit : float and array_like, optional
+        Critical points to take care during integration.  Forces
+        solver to integrate to this time point exactly before proceeding.
+        If an array of values is passed in, the solver will treat each
+        value as critical. The array of values must be sorted either
+        ascending or descending in the same manner as the direction
+        between ``t0`` and ``t_bound``.
+
+        .. versionadded:: 1.17.0
```
MatthewFlamm:
```suggestion
        .. versionadded:: 1.18.0
```

scipy/integrate/_ivp/lsoda.py
```diff
@@ -106,6 +115,10 @@ class LSODA(OdeSolver):
         Number of evaluations of the right-hand side.
     njev : int
         Number of evaluations of the Jacobian.
+    tcrit : ndarray
+        Critical time points to take care of integration.
```
MatthewFlamm:
```suggestion
```

PR #23387, opened 2025-07-26
MAINT: Python 3.14 to classifiers

tylerjereddy:
* With NumPy `2.3.2` now out with Python `3.14.0rc1` support, and Ralf's PRs that test SciPy wheels against that same version of Python now passing/merged, it is probably time to add this to the classifiers and backport it.

* See: https://github.com/numpy/numpy/pull/29444

[ci skip] [skip ci]


rgommers (approved):
Good catch, thanks Tyler. Agreed we're all set here for 3.14.

PR #23393, opened 2025-07-26
DOC: remove manual array api support list

j-bowhay:
Now we have an automatically generated list we should ditch this manual one


lucascolley (approved):
thanks Jake!

Issue #23396, opened 2025-07-27
BUG: interpolate: `test_bispev_integer_overflow` crashes for ILP64 build on macOS


rgommers:
### Describe your issue.

After adding ILP64 support to Accelerate, I found that `test_bispev_integer_overflow` is the one test that fails. The `_dfitpack` code it calls doesn't even use BLAS, so this crash will also show up for other ILP64 builds on macOS, and possibly on other platforms as well.

This test is a regression test for gh-12471, and that issue was closed after multiple attempts in gh-16649. The PRs that superseded, gh-16084 and gh-12473, contain more relevant discussion. The questions about "will this `F_INT` work when ILP64 is brought back" in there were justified it looks like.

For now I'll just skip the test on Accelerate ILP64, but it warrants fixing properly or working around structurally by just rejecting trying to allocate arrays larger than `2**31`.

### Reproducing Code Example

```python
% python -m pytest scipy/interpolate/tests/test_fitpack.py -k integer_overflow
```

### Error message

```shell
LLDB traceback for `test_bispev_integer_overflow` crash:


% lldb python -- -m pytest scipy/interpolate/tests/test_fitpack.py -k integer_overflow
(lldb) target create "python"
Current executable set to '/Users/rgommers/code/pixi-dev/scipy/.pixi/envs/accelerate/bin/python' (arm64).
(lldb) settings set -- target.run-args  "-m" "pytest" "scipy/interpolate/tests/test_fitpack.py" "-k" "integer_overflow"
(lldb) run
Process 19633 launched: '/Users/rgommers/code/pixi-dev/scipy/.pixi/envs/accelerate/bin/python' (arm64)
============================= test session starts ==============================
platform darwin -- Python 3.12.11, pytest-8.4.1, pluggy-1.6.0
rootdir: /Users/rgommers/code/pixi-dev/scipy/scipy
configfile: pytest.ini
plugins: xdist-3.8.0, timeout-2.4.0, cov-6.2.1, hypothesis-6.135.32
collected 45 items / 44 deselected / 1 selected                                

Process 19633 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGKILL
    frame #0: 0x000000019bc87f60 libsystem_platform.dylib`__bzero + 64
libsystem_platform.dylib`:
->  0x19bc87f60 <+64>: dc     zva, x3
    0x19bc87f64 <+68>: add    x3, x3, #0x40
[...]

PR #23399, opened 2025-07-27
ENH: add Accelerate ILP64 support and CI job

rgommers:
This is a follow-up to gh-22743, fixing a few small issues with that PR. That makes Accelerate build and test fine on macOS >=13.3 with the latest Meson release, the only manual bit being that one has to  add the BLAS symbol suffix manually with `-Dblas-symbol-suffix=$NEWLAPACK$ILP64`.
    
The one hiccup is an interpolate test for integer overflow that starts crashing, see gh-23396.

For the CI job, cleans up the existing Accelerate job a bit and makes the CI logs easier to read by breaking  up the steps.  Adds a new ILP64 job, and switches the jobs to `-m full` because it only takes 2 minutes or so longer and any ILP64-specific tests are likely to be marked as slow. Also there are no other `-m full` jobs for macOS.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -28,8 +28,6 @@ jobs:
   test_meson:
     name: Conda & umfpack/scikit-sparse, fast, py3.11/npAny, spin
     needs: get_commit_message
-    # If using act to run CI locally the github object does not exist and
-    # the usual skipping should not be enforced
     if: >
       needs.get_commit_message.outputs.message == 1
       && (github.repository == 'scipy/scipy' || github.repository == '')
@@ -143,8 +141,6 @@ jobs:
   test_scipy_openblas:
     name: M1 & OpenBLAS, fast, py3.11/npAny, spin
     needs: get_commit_message
-    # If using act to run CI locally the github object does not exist and
-    # the usual skipping should not be enforced
     if: >
       needs.get_commit_message.outputs.message == 1
       && (github.repository == 'scipy/scipy' || github.repository == '')
@@ -184,19 +180,17 @@ jobs:
         spin test
 
 
-  test_meson_accelerate:
-    name: Accelerate, fast, py3.11/npAny, spin
+  test_accelerate:
+    name: Accelerate, full, py3.13/npAny, spin
     needs: get_commit_message
-    # If using act to run CI locally the github object does not exist and
-    # the usual skipping should not be enforced
     if: >
       needs.get_commit_message.outputs.message == 1
       && (github.repository == 'scipy/scipy' || github.repository == '')
     runs-on: macos-15
     strategy:
+      fail-fast: false
       matrix:
-        python-version: ["3.11"]
-
+        blas-interface: ["LP64", "ILP64"]
     steps:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       with:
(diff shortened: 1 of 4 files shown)


.github/workflows/macos.yml
```diff
@@ -205,24 +199,34 @@ jobs:
...
+    - name: Make gfortran-13 on runner image usable
       run: |
-        sudo xcode-select -s /Applications/Xcode_16.app
+        # Ensure we use gfortran-13 and that its runtime is on the library search path
+        FC=$(which gfortran-13)
+        GFORTRAN_LIB=$(dirname `$FC --print-file-name libgfortran.dylib`)
+        echo DYLD_LIBRARY_PATH=$GFORTRAN_LIB >> "$GITHUB_ENV"
+        echo FC=$FC >> "$GITHUB_ENV"
 
-        git submodule update --init
```
andyfaff, line 215: Does the GH checkout update all submodules? Presumably this was originally added because it wasn't?

.github/workflows/macos.yml
```diff
@@ -205,24 +199,34 @@ jobs:
...
       with:
-        python-version: ${{ matrix.python-version }}
-        cache: 'pip'
+        python-version: "3.13"
 
-    - name: Build and Install SciPy
+    - name: Make gfortran-13 on runner image usable
       run: |
-        sudo xcode-select -s /Applications/Xcode_16.app
+        # Ensure we use gfortran-13 and that its runtime is on the library search path
```
andyfaff, line 206: on the macOS-15 runners there's gfortran-12/13/14, I can't remember the implications of using too old/too modern.

.github/workflows/macos.yml
```diff
@@ -205,24 +199,34 @@ jobs:
...
+        echo DYLD_LIBRARY_PATH=$GFORTRAN_LIB >> "$GITHUB_ENV"
+        echo FC=$FC >> "$GITHUB_ENV"
 
-        git submodule update --init
-        GFORTRAN_LOC=$(which gfortran-13)
-        ln -s $GFORTRAN_LOC gfortran
-        export PATH=$PWD:$PATH
+    - name: Install dependencies from PyPI
+      run: |
+        # These are the required build and test dependencies only
```
andyfaff:
```suggestion
```

.github/workflows/macos.yml
```diff
@@ -205,24 +199,34 @@ jobs:
...
 
-        # Ensure we have gfortran dylib
-        GFORTRAN_LIB=$(dirname `gfortran --print-file-name libgfortran.dylib`)
-        export DYLD_LIBRARY_PATH=$GFORTRAN_LIB
+    - name: Build SciPy with Accelerate (LP64)
+      if: matrix.blas-interface == 'LP64'
+      run: |
+        # The b_ndebug flag is a workaround for `_LIBCPP_ENABLE_ASSERTIONS` warnings, can
+        # be removed once Meson 1.8.3 is out with the fix (see meson#14440).
+        spin build --with-accelerate -S-Db_ndebug=true
```
andyfaff, line 221: unrelated - is it worth making the `--with-accelerate` default for spin building? Instead having a `--with-openblas`  flag or similar?

andyfaff (approved):
Few minor comments from me. Otherwise looks good to merge.

.github/workflows/macos.yml
rgommers: It does, that's what this bit in the `actions/checkout` invocation does:
```
    with:
        submodules: recursive
```

I suspect this separate `git` invocation was added by mistake instead of that `with:`, it's never had an issue AFAIK.

.github/workflows/macos.yml
rgommers: Me neither, I decided to just not touch it since 13 is new enough.

.github/workflows/macos.yml
rgommers: I added this as a note saying "this is deliberate, don't replace it with `pip install -r requirements/xxx.txt`. Perhaps I should make that more clear?

.github/workflows/macos.yml
rgommers: Yes, I think so. It's a lot easier to set up, and I don't think that there will be many contributors on macOS <13.3

.github/workflows/macos.yml
```diff
@@ -184,19 +180,17 @@ jobs:
         spin test
 
 
-  test_meson_accelerate:
-    name: Accelerate, fast, py3.11/npAny, spin
+  test_accelerate:
+    name: Accelerate, full, py3.13/npAny, spin
```
ev-br, line 184: Minor, trivial, optional: not sure `npAny` label is that useful; what is it even supposed to mean

.github/workflows/macos.yml
```diff
@@ -205,24 +199,34 @@ jobs:
...
+        # The b_ndebug flag is a workaround for `_LIBCPP_ENABLE_ASSERTIONS` warnings, can
+        # be removed once Meson 1.8.3 is out with the fix (see meson#14440).
+        spin build --with-accelerate -S-Db_ndebug=true
 
-        pip install meson cython pythran pybind11 ninja numpy spin
-        spin build --with-accelerate
+    - name: Build SciPy with Accelerate (ILP64)
+      if: matrix.blas-interface == 'ILP64'
+      run: |
+        # Once Meson has full BLAS support, the `blas-symbol-suffix` flag can be dropped
```
ev-br: Is there an upstream issue/PR to track here?

.github/workflows/macos.yml
```diff
@@ -205,24 +199,34 @@ jobs:
...
+        spin build -S-Dblas=accelerate -S-Duse-ilp64=true -S-Dblas-symbol-suffix='$NEWLAPACK$ILP64' -S-Db_ndebug=true
```
ev-br, line 228: So an upstream meson works modulo this `blas-symbol-suffix` on MacOS but on linux we need the numpy fork. Why? Some things are bound to remain a mystery, I guess.

scipy/_build_utils/src/npy_cblas.h
```diff
@@ -47,11 +47,10 @@ enum CBLAS_SIDE {CblasLeft=141, CblasRight=142};
 #define BLAS_FORTRAN_SUFFIX _
 #endif
 
-// New Accelerate suffix is always $NEWLAPACK (no underscore)
+/* Accelerate doesn't use an underscore as suffix, so fix that up here */
```
ev-br, line 50: Wow.

ev-br (approved):
LGTM!

A couple of small nits are all optional.

.github/workflows/macos.yml
rgommers: Yeah agreed. It was added at a time when numpy nightlies were failing a lot and hence it was useful to see at a glance whether a job was using that. It's not so useful anymore and right now making it harder to read more relevant parts of the label in some cases. 

I'd prefer to change that separately though, since nothing about it is specific to this job.

.github/workflows/macos.yml
rgommers: Yes, will link it in an update.

.github/workflows/macos.yml
rgommers: Not a mystery. It works fine on Linux too, if and only if you set the exact library names and symbol suffixes you need manually. What the Meson fork contains is the "do this automatically part".

Because there' much more variation across Linux distros than on macOS, you do in practice want that automatic behavior, it's too hard to get right without it.

.github/workflows/macos.yml
rgommers: Okay I'll just remove this - tried to extend the comment but it's too hard to explain clearly in a comment. Let's keep this under control through PR review and discouraging overzealous cleanups.

rgommers: applied this change, resolving the comment.

PR #23403, opened 2025-07-28
BUG: Skip installing gmpy2(test deps) for Win-ARM64

Greenie0701:
Issue Description:
- SciPy started providing wheels for Windows ARM64 in nightly builds. 
- Upon installing python dependencies from requirements folder, I found that installation of test dependencies are failing due to unavailability of gmpy2 for Win-ARM64
- Due to this, the installation of complete test dependencies are failing.

Fix:
- Added conditional statement to avoid installing gmpy2 for Windows ARM64 which leads to successful installation of test dependencies for Win-ARM64.

### DIFF
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -6,7 +6,7 @@ pytest-timeout
 pytest-xdist
 asv
 mpmath
-gmpy2
+gmpy2; sys_platform != "win32" or platform_machine != "ARM64"
 threadpoolctl
 # scikit-umfpack  # circular dependency issues
 pooch
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -6,7 +6,7 @@ pytest-timeout
 pytest-xdist
 asv
 mpmath
-gmpy2; sys_platform != "win32" and platform_machine != "ARM64"
+gmpy2;
 threadpoolctl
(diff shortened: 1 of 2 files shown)


lucascolley (changes requested):
As per the note at the top of the file, this file is generated by a script. Please could you add a special-case for this in the `tools/generate_requirements.py` script?

MGN-GIT:
thanks for your suggestion! made the relevant changes, let me know if there is any additional changes are required

lucascolley:
nice! Now you just need to run the script :)

MGN-GIT:
> nice! Now you just need to run the script :)

Thanks @lucascolley , yes after running the script, the test requirements are installing fine as expected!

lucascolley:
Can you commit the generated change to `requirements/test.txt`?

lucascolley (approved):
thanks @lowkeyrossi !

tylerjereddy (approved):
CI failures are clearly unrelated.

This seems "ok" to me as well.

5 commits is perhaps a bit much for this patch, and actually I suppose I might consider backporting it as well since `1.16.2` is aiming at Win ARM64 support, so that's two reasons to squash merge.

tylerjereddy:
thanks both

Issue #21782, opened 2024-10-30
RFC: special: Return type of special functions that return derivatives.


steppi:
In relatively recent work (https://github.com/scipy/scipy/pull/20539), @izaid developed a more consistent API for the Legendre functions, and developed tools for evaluating the recurrences used in their computation that simplified the implementation and reduced code duplication. These functions are able to return derivatives of Legendre functions up to order 2, and this is done through a keyword argument, `diff_n`, which takes values `0`, `1`, or `2`. The output of the function is then a tuple of length 1, 2, or 3 with the following values:

|diff_n| result |
|--------|----------|
| 0       |      (function value, )      |
| 1       |      (function value, gradient)      |
| 2       |      (function value, gradient, hessian)    |

This was discussed in the discourse post, https://discuss.scientific-python.org/t/naming-convention-for-generalized-ufuncs-in-special/1144, and in the issue, https://github.com/scipy/scipy/issues/20448. None of this has made it into a release yet. The functions with this behavior are all new functions, and we plan to deprecate the older Legendre function API.

Having a keyword argument control the number of output values is not ideal. In a comment on the discourse post mentioned above, @rgommers said:

*Some things are worse than others, obviously. Here is how I’d order them from most to least impactful (in a negative way) for behavior based on the value of a keyword:*

* Returning a different number of output values
* Returning an array with a different dtype
* Returning an array with a different number of dimensions

At the time #20539 went in, it would have been difficult to take another path, but since then, @izaid has done additional work improving how derivatives are computed for special functions, adding forward autodifferentiation in https://github.com/scipy/scipy/pull/21483. This makes computation of derivatives much more flexible (previously it was done in Zhang and Jin's code by manually differentiating code line by line). With this change, @izaid has informed me that it would be much easier to have legendre functions, and other special functions that return derivatives, each return a single array.

For functions of two variables $f(x, y)$, the idea is to return, for scalar input a 2d array `out` such that

$$\mathrm{out}\left[i, j\right] = \frac{\partial^{i+j}f}{\partial x^{i} \partial y^{j}}$$

and for array input to add two new dimensions for the derivatives. For functions of a single variable, it's straightforward what to do, and for functions with more than two variables, I hope it's clear how to generalize this.

There's a question of whether dimensions for the derivatives should be appended to the beginning or end of the output. I think having them at the front would be more consistent with what has already been done with the legendre functions, but will let @izaid weigh in on this.

[...]

PR #23404, opened 2025-07-28
DOC: special: fix output shape of `*_all` functions

JaRoSchm:
Hi, while working on https://github.com/scipy/xsf/issues/50 I noticed that the output shapes documented for `legendre_p_all`, `assoc_legendre_p_all`, `sph_legendre_p_all`, and `sph_harm_y_all` do not take the `diff_n` parameter into account. This should be fixed now.

### DIFF
--- a/scipy/special/_multiufuncs.py
+++ b/scipy/special/_multiufuncs.py
@@ -201,11 +201,13 @@ sph_legendre_p_all = MultiUFunc(
     """sph_legendre_p_all(n, m, theta, *, diff_n=0)
 
     All spherical Legendre polynomials of the first kind up to the
-    specified degree ``n`` and order ``m``.
+    specified degree ``n``, order ``m``, and all derivatives up
+    to order ``diff_n``.
 
-    Output shape is ``(n + 1, 2 * m + 1, ...)``. The entry at ``(j, i)``
-    corresponds to degree ``j`` and order ``i`` for all  ``0 <= j <= n``
-    and ``-m <= i <= m``.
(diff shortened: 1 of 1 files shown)


steppi:
Thanks, much appreciated @JaRoSchm! A lot of discussion and some controversy went into the API around returning derivatives for special functions, and the API changed several times before these functions were made public. My guess is the documentation for `diff_n` got lost in one of the changes. I had actually forgotten that the output was made an `ndarray` in the case of functions of one variable, when we don't have to worry about increasing `ndim` as the order of differentiation increases. Before that it was a `tuple` in all cases.

JaRoSchm:
Yeah, actually I was surprised by the functions which return tuples as I never used them before. You probably refer to the discussion in https://github.com/scipy/scipy/issues/21782 which I also found before suggesting this change here. In principle, I like the current API.

steppi:
> Yeah, actually I was surprised by the functions which return tuples as I never used them before. You probably refer to the discussion in #21782 which I also found before suggesting this change here. In principle, I like the current API.

Yeah, there was that GitHub issue, and also a lot of internal discussion off GitHub. Some other ideas that had popped up were:

1. Having `ndim` in the output depend on the highest order derivative, and have values and lower order derivatives be constant along the extra dimensions.
2. Use patterns like:
$$a[i, j] = \frac{\partial^i{f}}{\partial x^i}\frac{\partial^j{f}}{\partial y^i}$$
to add dimensions for derivatives to the output array.

What we ended up with seemed most convenient for users though, but there was a lot of back and forth.

steppi (approved):
Looks good to me. Thanks @JaRoSchm!

PR #23405, opened 2025-07-28
MAINT:BLD:_lib: Remove deprecated Python TLS fallback code from `ccallback.h`

ilayn:
Since we are way past supporting Python 3.7, this file can be cleaned-up with a more portable TLS Api instead of the old band-aids.

### DIFF
--- a/scipy/_lib/src/ccallback.h
+++ b/scipy/_lib/src/ccallback.h
@@ -67,9 +67,6 @@ struct ccallback {
 /*
  * Thread-local storage
  */
-
-#if !defined(SCIPY_TLS_EMPTY)
-
 static SCIPY_TLS ccallback_t *_active_ccallback = NULL;
 
 static void *ccallback__get_thread_local(void)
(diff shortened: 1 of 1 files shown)


lucascolley:
CI segfault sounds related

scipy/_lib/src/ccallback.h
```diff
@@ -68,104 +67,31 @@ struct ccallback {
...
- * Obtain a pointer to the current ccallback_t structure.
- */
-static ccallback_t *ccallback_obtain(void)
-{
-    return (ccallback_t *)ccallback__get_thread_local();
-}
-
-#else
-
-/* Fallback implementation with Python thread API */
```
rgommers, line 96: It looks like this fallback can indeed be removed, since `SCIPY_TLS_EMPTY` is never defined, and I cannot find any external usages nor issues mentioning it.

rgommers (changes requested):
Thanks @ilayn. The fallback path can be removed indeed. The rest of this PR does not look right to me. We aren't using any of the [old deprecated Python TLS API](https://docs.python.org/3/c-api/init.html#thread-local-storage-tls-api), so there is no reason to try and use the [new TSS API](https://docs.python.org/3/c-api/init.html#thread-specific-storage-tss-api) that replaced it. Using that API is going to be slower than using normal C17/C++17 TLS support. So I suggest limiting this PR to removing the fallback path.

andyfaff:
Tangential question, what would be the work required to get the ccallback interface working with the native minimize methods? I looked at it several years ago but couldnt wrap my head around how it worked.

scipy/_lib/src/ccallback.h
ilayn: That's indeed right, Python API is slow (but still nanoseconds) however we use it to wrap Python functions that are much slower operations in comparison hence negligable to anything we do with TLS (unless we decide to do a loop over a python function call which would be very strange). 

My concern was about the upcoming, if ever,  subinterpreter business, that native TLS might not respect. But either way I don't have any opinions about it.

ilayn:
> Tangential question, what would be the work required to get the ccallback interface working with the native minimize methods? I looked at it several years ago but couldnt wrap my head around how it worked.

Do you mean wrapping Python level f, grad, jac computing functions to be sent to C code? MINPACK does that as an example. But not sure I got your question right.

scipy/_lib/src/ccallback.h
ilayn: I think Pauli had some plans but then never materialized. Or he left it as an escape hatch if integer keys did not work.

rgommers: There shouldn't be a problem there, this code should work as expected AFAIK - and in case there is a problem, let's cross that bridge when we get to it.

> however we use it to wrap Python functions that are much slower operations

The callback mechanism is designed to avoid going through Python when used with `scipy.LowLevelCallable`, it should be fast.

rgommers: > I think Pauli had some plans but then never materialized. Or he left it as an escape hatch if integer keys did not work.

I'm pretty sure that the fallback was necessary in 2016 when he wrote this code. However it's 2025 now and there are no platforms without TLS support left, aside from embedded systems and the like where Python won't run. CPython >=3.12 requires TLS support I believe, it can't be built without.

ilayn: Ah yes I forgot the LowLevelCallable thing. Let me revert it back.

andyfaff:
> Do you mean wrapping Python level f, grad, jac computing functions to be sent to C code? MINPACK does that as an example. But not sure I got your question right.

Sorry, I should've been clearer. How easy would it be for `minimize` to be supplied with a `LowLevelCallable` for the objective function. If we go in via one of those presumably there's a reduction in call overhead.

ilayn:
I think there are two cases, one is of the form as in `L-BFGS-B` that uses reverse communication (basically goes back to Python and asks for new evaluations). The other one is passing the relevant Python callables to the low level code like minpack does. 

In the first one, the difference will be the evaluation speed of the callables, but this is done on the user side. In other words, you can do it already. However, the python overhead will be the same. So maybe I am missing a crucial point but that won't make too much difference (unless your evals are very demanding).

In the second case, ccallback already supports LowLevelCallables through ctypes or cffi so you can pass it to minpack based ones for testing. It will make a difference but I haven't tried it but "should" work. You can use the same mechanism elaborated here https://docs.scipy.org/doc/scipy/tutorial/integrate.html#quad-callbacks

ilayn:
Array api failure unrelated and happening on other PRs too. Thanks all.

rgommers:
Thanks Ilhan. Post-hoc review: code changes LGTM, the commit message and new PR title are slightly misleading, since this wasn't actually using deprecated CPython C API. This PR as merged simply removed unreachable dead code.

ilayn:
What I meant with that is the code that was removed was using deprecated TLS API even though it wasn't used by us. But was possible to reach if you set that env variable. Now you can't fallback no matter what.

rgommers:
Still not sure what you mean, because APIs like [PyThreadState_GetDict](https://docs.python.org/3/c-api/init.html#c.PyThreadState_GetDict) that are removed in this PR are not deprecated. There are six deprecated `PyThread_xxx` APIs (see https://docs.python.org/3/c-api/init.html#thread-local-storage-tls-api), we were not using any of those.

ilayn:
It's not the commands but the way we keep the thread state. If you don't use TLS, the extension must have a unique key defined via an int and we are creating a scipy specific entry into the thread dict. 

That behavior is replaced with opaque pointers in the newer api. So it's the behavior that is deprecated/outdated way of doing things. Pauli basically created a mini manual TLS implementation similar to the one that is deprecated, in the fallback code and that's what we removed.

PR #23406, opened 2025-07-28
DOC: stats.false_discovery_control: added note that input does not need to be sorted

noamteyssier:
#### Reference issue
No reference issue

This adds some extra detail on `false_discovery_control` to note that the inputs do not need to be sorted.

#### Additional information
Small quality of life improvement on the documentation which can save some devs from checking the source to see if they need to pre-sort the inputs or not.


mdhaber:
There is no indication in the existing documentation that the values need to be sorted, so without this addition, would the user assume that sorting might be required? I don't think we have any functions that implicitly assume sorted input, even for rank-based statistics.

noamteyssier:
It may be assumed you don’t need to sort but I wasn’t sure if I needed to for the procedure (because its implementation must work on sorted values). I looked through the code and saw it sorts then recreates the original sorting after correction. So I figured this note could save somebody some time in the future who has the same question as I did.

mdhaber:
OK; I'll let someone else take a look. For consistency, I would suggest also considering other statistics that sort internally - e.g. `wilcoxon`, `spearmanr`. Unless there is a reason that *this* function needs such a statement but the others don't, I'd make a decision (one way or the other) that applies to all of them.

steppi:
@noamteyssier, as an alternative to adding the note, do you think it would be sufficiently informative to change the example to use non-sorted input?

mdhaber:
I think we can also point out in the example the correspondence between input and output p-values. LMK if you'd like to make the change @noamteyssier or if I should go ahead.

noamteyssier:
@mdhaber feel free to make that change

steppi (approved):
This looks good to me.

PR #23410, opened 2025-07-29
MAINT: integrate: add xp_capabilities to public functions

j-bowhay:
Along the same lines of #23381, applies `xp_capabilities` and `make_xp_test_case`

### DIFF
--- a/scipy/integrate/_cubature.py
+++ b/scipy/integrate/_cubature.py
@@ -57,7 +57,7 @@ class CubatureResult:
     rtol: float
 
 
-@xp_capabilities()
+@xp_capabilities(allow_dask_compute=True, jax_jit=False)
 def cubature(f, a, b, *, rule="gk21", rtol=1e-8, atol=0, max_subdivisions=10000,
              args=(), workers=1, points=None):
     r"""
--- a/scipy/integrate/_cubature.py
+++ b/scipy/integrate/_cubature.py
@@ -10,7 +10,8 @@ from scipy._lib._array_api import (
     array_namespace,
     xp_size,
     xp_copy,
-    xp_promote
+    xp_promote,
+    xp_capabilities
 )
 from scipy._lib._util import MapWrapper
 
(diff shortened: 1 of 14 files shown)


steppi:
Thanks @j-bowhay. I'll take a close look tomorrow. To avoid duplication, I'm currently working on `ndimage` and plan to submit a PR for it tomorrow.

scipy/integrate/tests/test_tanhsinh.py
```diff
@@ -43,15 +44,6 @@ def wrapped(*arg_arrays):
     return decorator
 
 
-@pytest.mark.skip_xp_backends(
```
steppi, line 46: These got stripped out, but no `make_xp_test_case` was inserted in its place.

scipy/integrate/_cubature.py
```diff
@@ -56,6 +57,7 @@ class CubatureResult:
     rtol: float
 
 
+@xp_capabilities()
```
steppi:
```suggestion
@xp_capabilities(allow_dask_compute=True, jax_jit=False)
```

This is a little confusing, but applying  `@xp_capabilities()` to `f` and then using `@make_xp_test_case(f)` doesn't do the same thing as directly using the `xp` fixture in a test with no pytest marks to control skips etc. Default `@xp_capabilities()` also adds tests that the JAX jit works, and that dask remains lazy. Need to add these kwargs to get tests passing.

steppi (commented):
Thanks @j-bowhay. I think my two comments should take care of all the related test failures.

scipy/integrate/tests/test_tanhsinh.py
```diff
@@ -43,15 +44,6 @@ def wrapped(*arg_arrays):
...
-    'array_api_strict', reason='Currently uses fancy indexing assignment.'
-)
-@pytest.mark.skip_xp_backends(
-    'dask.array', reason='boolean indexing assignment'
-)
-@pytest.mark.skip_xp_backends(
-    'jax.numpy', reason='JAX arrays do not support item assignment.'
-)
 class TestTanhSinh:
```
j-bowhay, line 48:
```suggestion
@make_xp_test_case(_tanhsinh)
class TestTanhSinh:
```

steppi (approved):
Thanks  @j-bowhay! This looks good to me.

PR #23167, opened 2025-06-16
BUG: optimize.shgo: Complex cache `split_edge` differently

andyfaff:
@nickodell, I have my fingers crossed that this simple PR might fix #23158. Would it be possible to run your refcycle analysis on this patch?

PR #23414, opened 2025-07-31
BUG: signal: Avoid unbounded cache for ShortTimeFFT

### LINKED ISSUE #23411: BUG: Easy to reproduce memory leak in scipy.signal.periodogram
### Describe your issue.

Running below code returns:

scipy 1.16.0 (with memory leak)
```
[ Top 3 memory allocations ]
venv/lib/python3.13/site-packages/scipy/signal/_short_time_fft.py:1018: size=386 MiB, count=100000, average=4048 B
venv/lib/python3.13/site-packages/scipy/signal/_spectral_py.py:914: size=9377 KiB, count=100000, average=96 B
venv/lib/python3.13/site-packages/scipy/signal/_short_time_fft.py:1674: size=4904 KiB, count=50000, average=100 B
```

scipy 1.15.3 (without memory leak)
```
[ Top 3 memory allocations ]
venv/lib/python3.13/site-packages/numpy/lib/_stride_tricks_impl.py:326: size=23.2 KiB, count=496, average=48 B
venv/lib/python3.13/site-packages/numpy/lib/_stride_tricks_impl.py:101: size=23.2 KiB, count=495, average=48 B
venv/lib/python3.13/site-packages/scipy/fft/_pocketfft/helper.py:122: size=11.7 KiB, count=249, average=48 B
```

### Reproducing Code Example

```python
import tracemalloc

import numpy as np
import scipy.signal

tracemalloc.start()

array = np.random.randn(1000).astype(float)

for _ in range(50_000):
    frequencies, Pxx = scipy.signal.periodogram(
        array,
        1e12,
        window="hamming",
        detrend="linear",
    )

snapshot = tracemalloc.take_snapshot()
top_stats = snapshot.statistics("lineno")
print("[ Top 3 memory allocations ]")
for stat in top_stats[:3]:
    print(stat)
```

### Error message

```shell
No error message
```

```shell
* Windows 10 and Ubuntu 24.10
* python 3.13.3
* numpy 2.3.1
* scipy 1.16.0 and 1.16.1
```

nickodell:
If `@functools.cache()` is used on an instance method, this will hold references to all instances that were provided as the `self` argument. This leaks memory. Avoid this by using `@functools.cached_property` instead.

#### Reference issue

Closes #23411 

The `@cached_property` decorator works by adding an caching the property value in  the `__dict__` of the instance when the method is called. This avoids keeping the ShortTimeFFT object alive just because the cache entry has not yet expired. Instead, the cached property will expire when the ShortTimeFFT object is garbage collected.

This reduces memory usage in the following reproducer.

```
import numpy as np
import scipy.signal

array = np.random.randn(1000).astype(float)

for _ in range(50_000):
    frequencies, Pxx = scipy.signal.periodogram(
        array,
        1e12,
        window="hamming",
        detrend="linear",
    )
```

Before:

newplot(1)

After:

newplot(2)

#### Additional information

This change is covered by existing unit tests.


nickodell:
@DietBru I see you wrote most of this file. Would you be interested in reviewing this?

DietBru:
The changes look good to me. This [modifaction](https://github.com/nickodell/scipy/pull/10) removes all `lru_cache` decorators because they leak memory as well (though probably not as much).

nickodell:
# Discussion about removing lru_cache via explicit memoization

@DietBru

> The changes look good to me. This [modifaction](https://github.com/nickodell/scipy/pull/10) removes all `lru_cache` decorators because they leak memory as well (though probably not as much).

I do have two concerns about this approach.

1. Complexity. This takes a decent amount of code, and it is interleaved with the code that calculates the answer. There are non-obvious requirements about how to modify this. For example, if one adds an early return to improve performance, one must remember to add the returned value to the cache.
2. Thread safety. My understanding is that it is not thread-safe to concurrently mutate an OrderedDict, in two circumstances. The first is if the pure-Python fallback implementation of OrderedDict is used. The second is if the free-threaded build is used. I do not know if anyone is using ShortTimeFFT in a concurrent manner - you probably have a better idea of that. In contrast, `lru_cache()` is implemented in a thread-safe manner.

As a simpler alternative, what would you think about [Andrew's approach that he used to fix shgo](https://github.com/scipy/scipy/pull/23167/files)? This creates an `lru_cache()` instance inside the `__init__()` method, which ensures that the cache is scoped to the object. The code involved is fairly simple - it adds 2 lines per method that must be cached.

# Benchmark of memory use of the various approaches

I wrote a quick benchmark of how much memory each of the three approaches used, using a slightly modified version of the example script I used initially. This uses a much bigger array to stress the caches that hold an array-sized amount of info.

Here is a comparison of the memory used by the three approaches:

 * This PR, which is used as a baseline.
 * Adding OrderedDict to cache entries
 * Andrew's lru_cache approach

## This PR (baseline)

This was measured using the following test script.

```
# Run this with memray
# python -m memray run --aggregate scipy.py
# memray target with big array - intended to be as simple as possible
import numpy as np
import scipy.signal


array = np.random.randn(100000).astype(float)

for _ in range(2000):
    frequencies, Pxx = scipy.signal.periodogram(
        array,
        1e12,
        window="hamming",
        detrend="linear",
    )
```

Branch: https://github.com/scipy/scipy/compare/main...nickodell:scipy:njo-remove-sfft-ub-cache

Memray Usage Plot

newplot(3)

Timings

```
real	1m2.173s
user	0m53.863s
sys	0m9.268s
```

Peak Memory: 377.0 MiB

## OrderedDict on each method

This was measured using the previous test script.

Branch: https://github.com/scipy/scipy/compare/main...DietBru:ShortTimeFFT_remove_LRUcache_decorator

Memray Usage Plot

newplot(4)

Timings

```
real	1m3.026s
user	0m54.753s
sys	0m9.227s
```

Peak Memory: 182.4 MiB

## lru_cache inside `__init__`

This was measured using the previous test script.

Branch: https://github.com/scipy/scipy/compare/main...nickodell:scipy:njo-remove-sfft-ub-cache-2

Memray Usage Plot

newplot(5)

Timings

```
real	0m53.430s
user	0m52.311s
sys	0m2.036s
```

Peak Memory: 216.0 MiB

# Summary

* In terms of complexity, OrderedDict is most complex.
* In terms of memory use, OrderedDict used the least, with 40 MiB less than using lru_cache in `__init__`. Both used much less memory than leaving lru_cache unchanged.
* In terms of speed, using lru_cache in `__init__` was the fastest, being about 10s faster. (I have not investigated this benchmark in detail, so I may be doing something silly that is throwing off my result.)

DietBru:
After thinking a bit more about usage scenarios, I came to the conclusion that removing all LRU caches and just manually cache the last calculated value is probably the cleanest solution. I could not come up with a scenario where LRU caches would bring measurable speed gains (I originally added the decorators, because it was only one line ...). 
The changes can be found [here].

> 1.  Complexity. This takes a decent amount of code, and it is interleaved with the code that calculates the answer. There are non-obvious requirements about how to modify this. For example, if one adds an early return to improve performance, one must remember to add the returned value to the cache.

In general, I think you do have a valid point. In this context, I do think that the methods are short enough for such modifications to be feasible.

> 2. Thread safety. My understanding is that it is not thread-safe to concurrently mutate an OrderedDict, in two circumstances. The first is if the pure-Python fallback implementation of OrderedDict is used. The second is if the free-threaded build is used. I do not know if anyone is using ShortTimeFFT in a concurrent manner - you probably have a better idea of that. In contrast, lru_cache() is implemented in a thread-safe manner.

An interesting point, I have not thought about yet. The `ShortTimeFFT` class is not designed to be thread-safe. If the signal related caching attributes `_post_padding_last_val`,  `_upper_border_begin_last_val` and `_t_last_val` (in the version [here]) would be made thread-safe, a `ShortTimeFFT` instance could be used to calculate different length STFTs in different threads. This would still not make the `ShortTimeFFT` not thread-safe when modifying transform related parameters (e.g., `_fft_mode `). I do think such scenarios are a bit contrived, since swapping in a multi-threadable [FFT backend] would be expected to bring the biggest speed increases.

> In terms of speed, using lru_cache in __init__ was the fastest, being about 10s faster. (I have not investigated this benchmark in detail, so I may be doing something silly that is throwing off my result.)

I would expect the speed differences to be neglible. I strongly suspect this is difference is solely caused by the utilizing the `memray` module.

[here]: https://github.com/nickodell/scipy/pull/11
[FFT backend]: https://scipy.github.io/devdocs/reference/generated/scipy.fft.set_backend.html

nickodell:
@DietBru 

>In general, I think you do have a valid point. In this context, I do think that the methods are short enough for such modifications to be feasible.

That makes sense. If you're happy with it, I have no objection. I did want to mention the alternatives in case they were more appealing.

> An interesting point, I have not thought about yet. The `ShortTimeFFT` class is not designed to be thread-safe.

That's a valid point. I might not be focused on the most important factors here. It's hard for me to gauge how thread-safe code is intended to be, or what the possible gain, if any, is.

>After thinking a bit more about usage scenarios, I came to the conclusion that removing all LRU caches and just manually cache the last calculated value is probably the cleanest solution. I could not come up with a scenario where LRU caches would bring measurable speed gains (I originally added the decorators, because it was only one line ...).  
>The changes can be found https://github.com/nickodell/scipy/pull/11.

Thanks. I re-benchmarked the memory usage, and I found that this single-element cache had similar memory usage to the version where `lru_cache()` was completely removed. Very nice work.

Would you mind opening a PR within the scipy/scipy repo, targeting `main`, rather than targeting my branch? I can review it for you. I would prefer to merge this without the `lru_cache()` changes, because it makes it easier to bisect bugs if each PR is kept small.

rgommers (approved):
This LGTM, it's a clear win and easy to backport. So I agree with merging this as is now, and dealing with the `lru_cache` usage on methods separately.

> The ShortTimeFFT class is not designed to be thread-safe. If the signal related caching attributes _post_padding_last_val, _upper_border_begin_last_val and _t_last_val (in the version https://github.com/nickodell/scipy/pull/11) would be made thread-safe, a ShortTimeFFT instance could be used to calculate different length STFTs in different threads.

I'd say that everything in SciPy is supposed to be (or be made) thread-safe to the extent possible. In this case for a class with quite a bit of state, I think you got the goal right of making it possible to use a single instance in multiple threads for repeated calculations given the same state, but not trying to support mutation while using it in different threads (that can be explicitly documented as not supported).

DietBru:
You're right @nickodell, it's better to merge this PR as is, making it easier to back-port.
The chance that  the  `@lru_cache` effects are noticed in practice, are much slimmer (should be dealt with nonetheless). I will work on a new PR.

Thanks @nickodell, @rgommers for your helpful comments. Merging now.

PR #23415, opened 2025-07-31
MAINT: ndimage: add `xp_capabilities` to functions in public API

steppi:

Following up on #23410 and #23381, this adds the `xp_capabilities` decorator to all public functions in `scipy.ndimage`. `ndimage` uses a `support_alternative_backends` dispatching infrastructure similar to what `special` does, so the decorators are added in `scipy/ndimage/_support_alternative_backends.py` after creating dispatching wrappers.

### DIFF
--- a/scipy/ndimage/_support_alternative_backends.py
+++ b/scipy/ndimage/_support_alternative_backends.py
@@ -103,7 +103,7 @@ capabilities_dict = {
         cpu_only=True, allow_dask_compute=True, jax_jit=False
     ),
     "generate_binary_structure": xp_capabilities(out_of_scope=True),
-    "map_coodinates": xp_capabilities(
+    "map_coordinates": xp_capabilities(
         cpu_only=True, exceptions=["cupy", "jax.numpy"],
         allow_dask_compute=True, jax_jit=True
     )
--- a/scipy/ndimage/_support_alternative_backends.py
+++ b/scipy/ndimage/_support_alternative_backends.py
@@ -76,11 +76,23 @@ def delegate_xp(delegator, module_name, capabilities=None):
         return capabilities(wrapper)
     return inner
 
+default_capabilities = xp_capabilities(
+    cpu_only=True, exceptions=['cupy', 'jax.numpy']
+)
+
+capabilities_dict = {
+    "geometric_transform": xp_capabilities(
+        cpu_only=True, exceptions=['jax.numpy']
+    ),
+}
+
 # ### decorate ###
 for func_name in _ndimage_api.__all__:
     bare_func = getattr(_ndimage_api, func_name)
     delegator = getattr(_delegators, func_name + "_signature")
 
-    f = delegate_xp(delegator, MODULE_NAME)(bare_func)
+    capabilities = capabilities_dict.get(func_name, default_capabilities)
+
+    f = delegate_xp(delegator, MODULE_NAME, capabilities)(bare_func)
     # add the decorated function to the namespace, to be imported in __init__.py
     vars()[func_name] = f
--- a/scipy/ndimage/_support_alternative_backends.py
+++ b/scipy/ndimage/_support_alternative_backends.py
@@ -84,6 +84,9 @@ capabilities_dict = {
     "geometric_transform": xp_capabilities(
         cpu_only=True, exceptions=['jax.numpy']
     ),
+    "find_objects": xp_capabilities(
+        cpu_only=True, exceptions=['jax.numpy']
+    ),
 }
 
 # ### decorate ###
--- a/scipy/ndimage/_support_alternative_backends.py
+++ b/scipy/ndimage/_support_alternative_backends.py
@@ -87,6 +87,13 @@ capabilities_dict = {
     "find_objects": xp_capabilities(
         cpu_only=True, exceptions=['jax.numpy']
     ),
+    "distance_transform_bf": xp_capabilities(
+        cpu_only=True, exceptions=['jax.numpy']
+    ),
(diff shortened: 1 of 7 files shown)


scipy/ndimage/_support_alternative_backends.py
```diff
@@ -25,60 +25,98 @@ def _maybe_convert_arg(arg, xp):
...
-                elif isinstance(result, dict):
-                    # value_indices: result is {np.int64(1): (array(0), array(1))} etc
-                    return {
-                        k.item(): tuple(xp.asarray(vv) for vv in v)
-                        for k,v in result.items()
-                    }
-                elif result is None:
-                    # inplace operations
-                    return result
+        if SCIPY_ARRAY_API:
```
lucascolley: Was support always turned on prior to this change? I guess that's a good thing if no users noticed :)

scipy/ndimage/tests/test_measurements.py
```diff
@@ -147,48 +147,55 @@ def test_basic(self, xp):
             assert result[1].dtype.kind == 'i'
 
 
+@make_xp_test_case(ndimage.label)
```
lucascolley, line 150: no need to do it here (this was easier to review :) ) but I guess an alternative would have been to put these tests in a class

scipy/ndimage/tests/test_morphology.py
```diff
@@ -12,12 +14,11 @@
 
 skip_xp_backends = pytest.mark.skip_xp_backends
 xfail_xp_backends = pytest.mark.xfail_xp_backends
-pytestmark = [skip_xp_backends(cpu_only=True, exceptions=['cupy', 'jax.numpy'])]
 
 
 class TestNdimageMorphology:
 
-    @xfail_xp_backends('cupy', reason='CuPy does not have distance_transform_bf.')
+    @make_xp_test_case(ndimage.distance_transform_bf)
```
lucascolley, line 21: this class could also do with being broken up into focused classes at some point :)

lucascolley (approved):
thanks Albert, looks solid! Very pleasing diff (apart from the lack of organised classes which pre-exists this PR), nice to see it coming together!

scipy/ndimage/_support_alternative_backends.py
steppi: No. The check was just in a different place. Previously things looked like this. 

```python
  f = (delegate_xp(delegator, MODULE_NAME)(bare_func)
         if SCIPY_ARRAY_API
         else bare_func)
```

Somehow I got it in my mind that it makes sense to pass `capabilities` to `delegate_xp`, but I could have well enough done this:

```python
  f = capabilities(delegate_xp(delegator, MODULE_NAME)(bare_func)
         if SCIPY_ARRAY_API
         else bare_func)
```

I think I like the latter way of doing it better now, to avoid the extra nesting in the body of `delegate_xp`. Do you have a preference?

scipy/ndimage/tests/test_measurements.py
steppi: I used emacs macros to help add these. I agree that it would be nice to refactor these tests, but thought it would be better for a separate PR.

steppi:
Thanks for the review @lucascolley. All failures look unrelated.

scipy/ndimage/_support_alternative_backends.py
lucascolley: No preference from me, whatever you think is best!

lucascolley:
> All failures look unrelated.

Yes but they are blocking other tests from running in the array API job. Please could we get tests green before merging?

scipy/ndimage/_support_alternative_backends.py
steppi: Cool. I changed it to the second way of doing this. This makes the diff much simpler too.

steppi:
> > All failures look unrelated.
> 
> Yes but they are blocking other tests from running in the array API job. Please could we get tests green before merging?

Oh, ok. I got most of them passing in #23381. There's one more outside of stats that also needs an `array_api_strict` skip. I'll just add it in there, and then they should all pass.

steppi:
@lucascolley, don't merge yet. The behavior around `CUPY_BLOCK_LIST` is wrong, and I just accepted it uncritically. That needs to go. If a function isn't available in CuPy, trying to use it on a CuPy array should raise, not convert to NumPy and back.

Update: It's worse than I thought. It will convert to NumPy for `torch` and anything else. `delegate_xp` needs some surgery.

steppi:
> @lucascolley, don't merge yet. The behavior around `CUPY_BLOCK_LIST` is wrong, and I just accepted it uncritically. That needs to go. If a function isn't available in CuPy, trying to use it on a CuPy array should raise, not convert to NumPy and back.
> 
> Update: It's worse than I thought. It will convert to NumPy for `torch` and anything else. `delegate_xp` needs some surgery.

Oh. False alarm. It's not as bad as I thought. It's a tests problem. There are a bunch of GPU tests that pass for `torch` when presumably they shouldn't, but it looks like most either 1. test if an error is raised for invalid input, 2. don't actually use the `xp` fixture, are hard coded to just test with NumPy, but are  just part of test classes that have `@make_xp_test_case`.

steppi:
@lucascolley, I think this is actually fine to merge. We aren't actually running GPU tests for functions on backends where this isn't supported. Having some of the tests pass if they are run because they check an error is raised, but not specifically enough, could be addressed in a follow-up where the test suite is refactored.

lucascolley:
thanks again!

Issue #23408, opened 2025-07-29
ENH: stats.kendall_tau: MergeSort to achieve O(n log n) complexity


HDembinski:
### Is your feature request related to a problem? Please describe.

The naive computation of concordant pairs has $O(n^2)$ complexity, which makes this computation prohibitive for large data sets (millions). This is what scipy currently implements, see https://github.com/scipy/scipy/blob/main/scipy/stats/_stats.pyx#L84

### Describe the solution you'd like.

As described in the Wikipedia, one can compute the number of concordant pairs using MergeSort in O(n log n)
https://en.wikipedia.org/wiki/Kendall_rank_correlation_coefficient#Algorithms, original source https://www.tandfonline.com/doi/abs/10.1080/01621459.1966.10480879
which is a huge advantage, obviously, for large datasets.

### Describe alternatives you've considered.

I don't know of alternatives.

lucascolley:
sounds like PR welcome!

SuriyaaMM:
@lucascolley forgive me if I am wrong, but I think the implementation there already uses an fenwick tree, each update and query operations takes at worst O(logn) the outer loop runs for at most O(n) (both k and i advance together).

HDembinski:
@SuriyaaMM I noticed that on second glance, too. At first glance, I just noticed the two nested while loops, so I thought it is the $O(n^2)$ algorithm. In my defense, there is no documentation given how this algorithm works, and the scipy docs also don't mention the complexity. The implementation does not look like the Wikipedia solution with the MergeSort, so I initially thought it is the $O(n^2)$ implementation.

I now checked the corresponding PR https://github.com/scipy/scipy/pull/10108 which is mentioned in the code, which gives some benchmarks, and the runtime indeed looks more like $O(n log n)$, in that table every row doubles the elements in the list.

I confirmed that scaling myself on the console.

So, in conclusion, this is just a documentation issue. I shall open another issue focused on that.

PR #23416, opened 2025-08-01
DOC: Add complexity remark to scipy.stats.kendalltau

HDembinski:
[docs only]

#### Reference issue

Follow-up of #23408 

Naive computation of kendalltau has O(n^2) complexity, it is important to note that this implementation is way faster with O(n log(n)) complexity, to avoid misunderstandings.

#### Additional information

None.


dschmitz89:
Thanks @HDembinski

PR #23420, opened 2025-08-02
MAINT: break import dependencies between special-linalg and linalg-sparse

### LINKED ISSUE #22412: RFC/API: move `clarkson_woodruff_transform` from `linalg` to `sparse.linalg`?

j-bowhay:
closes #22412

Removes some of the spaghetti by making the imports lazy

### DIFF
--- a/scipy/linalg/_sketches.py
+++ b/scipy/linalg/_sketches.py
@@ -44,6 +44,7 @@ def cwt_matrix(n_rows, n_columns, rng=None):
     .. math:: \|SA\| = (1 \pm \epsilon)\|A\|
     Where the error epsilon is related to the size of S.
     """
+    # lazy import to prevent to prevent sparse dependency for whole module (gh-23420)
     from scipy.sparse import csc_matrix
     rng = check_random_state(rng)
     rows = rng_integers(rng, 0, n_rows, n_columns)
@@ -174,6 +175,7 @@ def clarkson_woodruff_transform(input_matrix, sketch_size, rng=None):
     166.58473879945151
 
     """
+    # lazy import to prevent to prevent sparse dependency for whole module (gh-23420)
     from scipy.sparse import issparse
     if issparse(input_matrix) and input_matrix.ndim > 2:
         message = "Batch support for sparse arrays is not available."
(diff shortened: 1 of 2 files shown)


scipy/linalg/_sketches.py
```diff
@@ -45,6 +44,7 @@ def cwt_matrix(n_rows, n_columns, rng=None):
     .. math:: \|SA\| = (1 \pm \epsilon)\|A\|
     Where the error epsilon is related to the size of S.
     """
+    from scipy.sparse import csc_matrix
```
ev-br, line 48: Would be nice to have an explanatory comment, potentially with a link to the gh issue.

So that it's clearer to future contributors that it's a deliberate exception to the rule that by default all imports are eager and are at the top level.

j-bowhay, line 48:
```suggestion
    # lazy import to prevent to prevent sparse dependency for whole module (gh-23420)
    from scipy.sparse import csc_matrix
```

scipy/linalg/_sketches.py
```diff
@@ -174,6 +174,7 @@ def clarkson_woodruff_transform(input_matrix, sketch_size, rng=None):
     166.58473879945151
 
     """
+    from scipy.sparse import issparse
```
j-bowhay, line 179:
```suggestion
    # lazy import to prevent to prevent sparse dependency for whole module (gh-23420)
    from scipy.sparse import issparse
```

scipy/special/_orthogonal.py
```diff
@@ -172,6 +171,7 @@ def _gen_roots_and_weights(n, mu0, an_func, bn_func, f, df, symmetrize, mu):
     mu ( = h_0 )        is the integral of the weight over the orthogonal
                         interval
     """
+    from scipy import linalg
```
j-bowhay, line 175:
```suggestion
    # lazy import to prevent to prevent linalg dependency for whole module (gh-23420)
    from scipy import linalg
```

rgommers (approved):
LGTM, thanks @j-bowhay

PR #23424, opened 2025-08-03
ENH: spatial: optimize performance of Rotation

amacati:
This PR improves the performance of Rotation after the refactor into a pure Python class and a Cython backend.

Optimizations are pretty much the same as in #23193. It also removes `broadcastable` from `_rotation.py` which has moved to `scipy._lib._util`.

### Motivation
Calling into the backends always costs some performance. We now skip validations for backend functions that are guaranteed and tested to return valid quaternions. An example of performance gains can be seen below

image

### Reference issue(s)
This is a continuation of #22777 working towards #22500 and a direct follow-up of #23198 and #23249. For more information on the progress towards #22500, see #23391.

### Changes to the Cython backend
The changes to the Cython backends make sure we always return numpy arrays instead of memory views. I think this is justified and should have already been done in the original PR as it is natural to assume that all backends return "their" `Array` type.

### Benchmarks
Full benchmarks are given below. We compare the current implementation with the optimized (this PR) version. Current performance is plotted in dashed lines, the optimized performance in solid lines. jax has a native implementation for some of the functions. These are plotted in dotted lines. Since the plots are getting a bit crowded, each framework is plotted in a dedicated subplot. Axes are consistent across individual functions (e.g. the axes for `Rotation.from_matrix` have the same ticks across all frameworks). The array API feature flag is always enabled.

[rotation_benchmark.pdf](https://github.com/user-attachments/files/21566714/rotation_benchmark.pdf)

### DIFF
--- a/scipy/spatial/transform/_rotation.py
+++ b/scipy/spatial/transform/_rotation.py
@@ -2183,8 +2183,8 @@ class Rotation:
 
         self._quat = self._backend.setitem(self._quat, value.as_quat(), indexer)
 
-    @classmethod
-    def identity(cls, num: int | None = None) -> Rotation:
+    @staticmethod
+    def identity(num: int | None = None) -> Rotation:
         """Get identity rotation(s).
 
         Composition with the identity rotation has no effect.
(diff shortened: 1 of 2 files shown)


amacati:
Any news / comments @lucascolley ?

lucascolley (approved):
LGTM, thanks as always @amacati!

PR #23425, opened 2025-08-03
ENH: spatial.transform.RigidTransform: add array API backend

amacati:
Add an array API backend to RigidTransform.

This PR implements a generic version `_rigid_transform_xp.py` of the functional-style cython backend `_rigid_transform_cy.pyx` using only array API features. Since we want to support lazy backends and jax.jit compilation, the backend avoids value-dependent branching wherever possible.

### Additional considerations
Due to shape typing restrictions, the cython backend can only use broadcasting in a few select places. The array API backend instead relies heavily on broadcasting to directly produce correct shapes without the `_single` distinction. This reduces special-case handling and would allow us to add support for multiple batch dimensions for `RigidTransform` in future PRs.

## Reference issue
#22500. Progress tracker at #23391

## Related PRs

- #22777
- #22972 
- #23198
- #23249 (backend required by this PR)

Implement the generic Array API backend for `RigidTransform`.

## Additional information
The Array API tests have been converted in #22939 . These tests have been left unchanged with the following exceptions:
- Test cases are now using `make_xp_test_case`
- Test cases unintentionally using `np.array()` have been fixed
- `dtype` errors from array-api-strict 2.4.1 have been fixed
- One test case from `test_empty_transform_indexing` has been disabled for now. There needs to be a discussion if the indexing that we previously tested for should be supported.
- Added one test case for iterating over transforms

## Current blockers
One test (namely test_setitem) still fails locally with array_api_strict because setitem is not supported with fancy indexing in array_api_extra. Perhaps @lucascolley and @crusaderky have ideas on how to resolve this issue without too many manual case handlings.

## Benchmarks
Currently missing, will run benchmarks once the PR is ready for merging. Local tests look as expected.

## TODOs

- [x] Fix `__setitem__`
- [ ] Benchmarks


amacati:
Any ideas on supporting fancy indexing in setindex @crusaderky ?

crusaderky:
> Any ideas on supporting fancy indexing in setindex @crusaderky ?

integer array index on `__setitem__` is unsupported by the Array API:
https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__setitem__.html#array_api.array.__setitem__

if you think it is important to have it, you should make a case for it on https://github.com/data-apis/array-api/issues/864.

amacati:
This should also be an issue in the current `Rotation` implementation, but it seems like it's lacking testing and thus did not come up during the discussion.

scipy/spatial/transform/tests/test_rigid_transform.py
```diff
@@ -1254,7 +1311,8 @@ def test_empty_transform_indexing(xp):
     tf_zero = tf_many[xp.asarray([], dtype=xp.int32)]
     assert len(tf_zero) == 0
 
-    assert len(tf_zero[xp.asarray([], dtype=xp.int32)]) == 0
+    # TODO: We cannot take from an already empty transform. Remove this test?
+    # assert len(tf_zero[xp.asarray([], dtype=xp.int32)]) == 0
```
lucascolley: seems fine to me given the check above

lucascolley (commented):
changes to tests LGTM as of 29e011f11d3b47730b0b559722f33ce0d57bab79

amacati:
> Any ideas on supporting fancy indexing in setindex @crusaderky ?

It seems like fancy `__setitem__` semantics are fundamentally flawed at the moment (see https://github.com/data-apis/array-api/issues/864#issuecomment-3197669514). @lucascolley What's your sentiment on this? Should we only allow `__setitem__`  + integer arrays for the numpy backend for now?

crusaderky:
> > Any ideas on supporting fancy indexing in setindex @crusaderky ?
> 
> It seems like fancy `__setitem__` semantics are fundamentally flawed at the moment (see [data-apis/array-api#864 (comment)](https://github.com/data-apis/array-api/issues/864#issuecomment-3197669514)). @lucascolley What's your sentiment on this? Should we only allow `__setitem__` + integer arrays for the numpy backend for now?

I would not call them "fundamentally flawed". 
The specific use case of duplicate indices is what is problematic.

amacati:
Sry for the imprecision, that was essentially what was meant. But it seems like there is indeed a common behavior specifically for `__setitem__` which could be implemented in array-api-extra (https://github.com/data-apis/array-api/issues/864#issuecomment-3199917384). Implementing that for array-api-strict should unlock progress with this PR, so I'll give that a stab once I find some time.

lucascolley:
I'll mark this as draft pending resolution to https://github.com/data-apis/array-api-extra/pull/395

amacati:
After some discussion over in https://github.com/data-apis/array-api-extra/pull/395 it seems unlikely that fancy indexing and `__setitem__` will make it into array-api-extra.

Since the only failing test was the one using integer array indices for array-api-strict, I have now split up the tests for `__setitem__` into the ones supported by the standard and the ones leveraging fancy indexing. array-api-strict is disabled for fancy indexing, all other frameworks are tested. This should unblock the PR and we can move forward with the reviews. @lucascolley @crusaderky

scipy/spatial/transform/_rigid_transform.py
```diff
@@ -1153,6 +1191,10 @@ def __len__(self) -> int:
             raise TypeError("Single transform has no len")
         return self._matrix.shape[0]
 
+    @xp_capabilities(
+        jax_jit=False,
+        skip_backends=[("dask.array", "cannot handle zero-length rotations")],
```
lucascolley: is there a comment somewhere explaining the underlying reason for this limitation? What is missing/bugged?

scipy/spatial/transform/_rigid_transform_xp.py
```diff
@@ -0,0 +1,385 @@
...
+    result = xpx.at(result)[..., 1, 2].set(-vec[..., 0])
+    result = xpx.at(result)[..., 2, 0].set(-vec[..., 1])
+    result = xpx.at(result)[..., 2, 1].set(vec[..., 0])
+    return result
+
+
+def _normalize_dual_quaternion(
+    real_part: Array, dual_part: Array
+) -> tuple[Array, Array]:
+    """Ensure that unit norm of the dual quaternion.
```
lucascolley: sentence feels incomplete?

lucascolley (approved):
thanks @amacati, this looks great!

scipy/spatial/transform/_rigid_transform.py
amacati: This first surfaced in the `Rotation` PR as a consequence of how dask handles zero dimensional arrays. See https://github.com/scipy/scipy/pull/23249#issuecomment-3092170244. The capabilities were taken from rotations by the way, hence the "rotations" mislabel. Changed the description to "rigid transforms".

lucascolley: makes sense, thanks

scipy/spatial/transform/_rigid_transform_xp.py
amacati: This was copied over from the doc string of the cython backend. Both should be correct now.

amacati:
MacOS tests fails are unrelated. Any chance you have some bandwidth to review @crusaderky ?

scipy/spatial/transform/_rigid_transform.py
```diff
@@ -24,10 +26,11 @@
 backend_registry = {array_namespace(np.empty(0)): cython_backend}
```
crusaderky, line 26: Cosmetic suggestion; feel free to ignore:

```python
from scipy._lib._array_api import is_numpy

def backend_for(xp):
    return cython_backend if is_numpy(xp) else xp_backend

[...]
return backend_for(xp).normalize_dual_quaternion(dual_quat)
```
note how `is_numpy_namespace` is wrapped by `@lru_cache` in array_api_compat so there should not be a performance penalty.

scipy/spatial/transform/_rigid_transform_xp.py
```diff
@@ -0,0 +1,385 @@
...
+        )
+    return tf_matrix @ other_tf_matrix
+
+
+def inv(matrix: Array) -> Array:
+    xp = array_namespace(matrix)
+    r_inv = xp.matrix_transpose(matrix[..., :3, :3])
+    # Matrix multiplication of r_inv and translation vector
+    t_inv = -(r_inv @ matrix[..., :3, 3][..., None])[..., 0]
+    matrix = xp.zeros((*matrix.shape[:-2], 4, 4), dtype=matrix.dtype)
```
crusaderky:
```suggestion
    matrix = xp.zeros((*matrix.shape[:-2], 4, 4), dtype=matrix.dtype, device=xp_device(matrix))
```

scipy/spatial/transform/_rigid_transform_xp.py
```diff
@@ -0,0 +1,385 @@
...
+        return xp.tile(identity, (*matrix.shape[:-2], 1, 1))
+    elif n == -1:
+        return inv(matrix)
+    elif n == 1:
+        return matrix
+    return from_exp_coords(as_exp_coords(matrix) * n)
+
+
+def setitem(
+    matrix: Array, indexer: Array | int | slice | EllipsisType | None, value: Array
```
crusaderky: is it deliberate that indexer can't be a tuple?

scipy/spatial/transform/tests/test_rigid_transform.py
```diff
@@ -10,19 +10,26 @@
...
     is_numpy,
     xp_assert_close,
+    make_xp_pytest_marks,
+    make_xp_test_case,
+    xp_assert_equal,
+    xp_promote
 )
 import scipy._lib.array_api_extra as xpx
 
+lazy_xp_modules = [RigidTransform, normalize_dual_quaternion]
```
crusaderky:
```suggestion
lazy_xp_modules = [RigidTransform]
```
functions imported in the local namespace are automatically caught.

scipy/spatial/transform/tests/test_rigid_transform.py
```diff
@@ -10,19 +10,26 @@
...
 
-pytestmark = pytest.mark.skip_xp_backends(np_only=True)
+pytestmark = make_xp_pytest_marks(RigidTransform.as_matrix)
```
crusaderky: I'm not a fan of this. This makes it quite hard to spot apparently unmarked tests, and has the unexpected side effect of marking non-array-API tests. I'd rather have a marker on each test although I would avoid unnecessary verbosity if there is already a much stricter marker. In other words, mark a test explicitly if it tests `as_matrix`; don't add a mark for `as_matrix` if the test tests something else (and there is already a `make_xp_pytest_marks` for it) and just uses as_matrix to fetch the output.

scipy/spatial/transform/_rigid_transform.py
```diff
@@ -1707,6 +1772,19 @@ def __reduce__(self) -> tuple[Callable, tuple]:
             matrix = matrix[0, ...]
         return (self.__class__.from_matrix, (matrix,))
 
+    @xp_capabilities(jax_jit=False)
```
crusaderky: I assume it is an artefact of an iterator object not being able to get out of the JIT? This function looks JIT-compatible to me; it's just wrapping it individually in the jit that is not possible.
This feels like a use case where `xp_capabilities` may not be a good fit.

scipy/spatial/transform/_rigid_transform.py
amacati: Actually this will become necessary when adding broadcasting support (next PR), so might as well add it now. See https://github.com/scipy/scipy/blob/377ee11cf92a9e3bb1dc303e140ed5daebd2e169/scipy/spatial/transform/_rotation.py#L27

I would like to keep the registry though. If at any point someone wants to switch out the backend with a specialized version for e.g. torch, this would allow them to extend the functionality with zero changes by just registering the backend.

scipy/spatial/transform/_rigid_transform_xp.py
amacati: No, missed that one.

scipy/spatial/transform/_rigid_transform.py
lucascolley: Eventually it would be good to use https://github.com/scipy/scipy/pull/23493 to allow backend registration

scipy/spatial/transform/tests/test_rigid_transform.py
amacati: A bit too much magical stuff is happening implicitly for my taste. But I guess this is also a pytest thing. The fact that `xp` is never declared in this file bugs me as well.

scipy/spatial/transform/_rigid_transform.py
crusaderky: You could try removing `xp_capabilities` from here and to the test module add

```python
def jitted_list(x):
    """Test x.__iter__ inside jax.jit"""
    return list(x)

lazy_xp_function(jitted_list)


def test_iter(xp):
    data = ...
    actual = jitted_list(data)
```

amacati: Added the test.

scipy/spatial/transform/tests/test_rigid_transform.py
amacati: Somewhat agree, though the `as_matrix` marker had the benefit of skipping all dask testing because we do not support it at the moment. Some tests that only use `as_matrix` coincidentally and are really checking something else have an additional marker now.

scipy/spatial/transform/_rigid_transform.py
crusaderky: Upstream fix: https://github.com/data-apis/array-api-extra/pull/418

amacati: So with the upstream fix we can write @xp_capabilities again?

amacati:
Do we wait for the iterator support before merging this?

lucascolley:
> Do we wait for the iterator support before merging this?

Up to you, I'm happy either way. Happy to make a quick release upstream if that would be preferable to making a follow-up PR for you.

amacati:
Then let's merge the iterator fix and add that here instead of opening another PR.

lucascolley:
I'll release xpx tomorrow if I don't get round to it today

scipy/spatial/transform/_rigid_transform.py
crusaderky: yes.

amacati: Nice. Then we can also include that in `Rotation`.

crusaderky:
FYI, in the past we vendored the git tip of array-api-compat and/or array-api-extra

lucascolley:
tagged: https://github.com/data-apis/array-api-extra/tree/v0.9.0

amacati:
@lucascolley Can you let me know when the xpx 0.9 release is vendored under `scipy._lib` so that I can push the final updates to `__iter__`?

amacati:
CI is happy. Anything else or are we good for merging?

lucascolley (approved):

amacati:
Minor change: `select_backend` was not used consistently, and concatenate with a single `RigidTransform` was not using the `_from_raw_matrix` optimization.

amacati:
@crusaderky Any additional review comments on this one?

crusaderky:
> @crusaderky Any additional review comments on this one?

I'm on holiday and won't be able to give this another review pass until late next week. I'm happy to defer to the other reviewers judgement.

lucascolley:
Thanks all!

PR #10380, opened 2019-07-01
ENH: add geometric_slerp

tylerjereddy:
**Purpose**: Provide an N-dimensional implementation of [geometric spherical linear interpolation](https://en.wikipedia.org/wiki/Slerp#Geometric_Slerp), which is not confined to 3-dimensions like [`scipy.spatial.transform.Slerp`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Slerp.html#scipy.spatial.transform.Slerp) & also not chained to infrastructure related to or understanding of quaternions.

**Context**: This is broken off from a much larger effort to support the calculation of the surface areas of spherical polygons in #8069. There, Slerp served the purpose of discretizing spherical surface polygons to achieve higher or lower accuracy in the surface area (integral) values, as desired. 

Arguably, having two different kinds of Slerp in SciPy may be contentious. I did try to [bring this up](https://github.com/scipy/scipy/pull/8994#discussion_r200010310), but didn't get a response when reviewing the original PR that uses quaternions. Regardless, I can always just move this to another repo if we don't want it in SciPy.

A more common / popular application of Slerp is for spherical trajectory simulation / animation on curved surfaces -- see my [example in matplotlib issue tracker from a while ago](https://github.com/matplotlib/matplotlib/issues/5294#issuecomment-367877945).

If we do like this, probably needs at least:

- [x] `asv` benchmarks
- [x] reduce verbosity in some places maybe

PR #23426, opened 2025-08-03
ENH: spatial.geometric_slerp: allow for extrapolation

tylerjereddy:
* Fixes gh-23400.

* Add support and regression tests for extrapolation behavior in `geometric_slerp()`, and update the documentation for the interpolation parameter accordingly.

* The `Raises` section of the `geometric_slerp()` docstring was also updated to remove a spurious claim about antipodes--we now actually allow i.e., North and South pole points despite the infinite number of paths between them on the sphere, and enforce this behavior in our regression tests (a warning, but not error, is issued). That is not directly related to the changes here, but the documented innacurracy of behavior would otherwise make this harder to reason about.

[ci skip] [skip ci]

TODO:

- [x] reactivate the CI and make sure it is passing (it passes locally, for now)
- [x] think of some more useful extrapolation test cases to add?
- [x] expand the `geometric_slerp()` docstring to include an example (+ plot) of extrapolation on the circle or sphere

@Ockenfuss since you were enthusiastic about adding this capability, would you like to suggest a few more interesting test cases and possibly make a suggestion for a docstring example of extrapolation you'd like to see?

### DIFF
--- a/scipy/spatial/_geometric_slerp.py
+++ b/scipy/spatial/_geometric_slerp.py
@@ -175,6 +175,26 @@ def geometric_slerp(
     ...         result[...,2],
     ...         c='k')
     >>> plt.show()
+
+    It is also possible to perform extrapolations outside
+    the interpolation interval by using interpolation parameter
+    values below 0 or above 1. For example, the above example
+    may be adjusted to extrapolate to an antipodal position:
+
+    >>> fig = plt.figure()
+    >>> ax = fig.add_subplot(111, projection='3d')
+    >>> ax.plot_surface(x, y, z, color='y', alpha=0.1)
+    >>> start = np.array([1, 0, 0])
+    >>> end = np.array([0, 0, 1])
+    >>> t_vals = np.linspace(0, 2, 400)
+    >>> result = geometric_slerp(start,
+    ...                          end,
+    ...                          t_vals)
+    >>> ax.plot(result[...,0],
+    ...         result[...,1],
+    ...         result[...,2],
+    ...         c='k')
+    >>> plt.show()
     """
 
     start = np.asarray(start, dtype=np.float64)
--- a/scipy/spatial/_geometric_slerp.py
+++ b/scipy/spatial/_geometric_slerp.py
@@ -190,10 +190,7 @@ def geometric_slerp(
     >>> result = geometric_slerp(start,
     ...                          end,
     ...                          t_vals)
-    >>> ax.plot(result[...,0],
-    ...         result[...,1],
-    ...         result[...,2],
-    ...         c='k')
+    >>> ax.plot(result[...,0], result[...,1], result[...,2], c='k')
     >>> plt.show()
     """
 
--- a/scipy/spatial/_geometric_slerp.py
+++ b/scipy/spatial/_geometric_slerp.py
@@ -52,10 +52,13 @@ def geometric_slerp(
         object. `n` must be greater than 1.
     t : float or (n_points,) 1D array-like
         A float or 1D array-like of doubles representing interpolation
-        parameters, with values required in the inclusive interval
-        between 0 and 1. A common approach is to generate the array
+        parameters. A common approach is to generate the array
         with ``np.linspace(0, 1, n_pts)`` for linearly spaced points.
         Ascending, descending, and scrambled orders are permitted.
+
+        .. versionchanged:: 1.17.0
+            Extrapolation is permitted, allowing values below `0`
+            and above `1`.
     tol : float
         The absolute tolerance for determining if the start and end
         coordinates are antipodes.
@@ -73,7 +76,7 @@ def geometric_slerp(
     Raises
     ------
     ValueError
-        If ``start`` and ``end`` are antipodes, not on the
+        If ``start`` or ``end`` are not on the
         unit n-sphere, or for a variety of degenerate conditions.
 
     See Also
@@ -225,9 +228,6 @@ def geometric_slerp(
 
     if t.size == 0:
         return np.empty((0, start.size))
-
-    if t.min() < 0 or t.max() > 1:
-        raise ValueError("interpolation parameter must be in [0, 1]")
 
     if t.ndim == 0:
         return _geometric_slerp(start,
--- a/scipy/spatial/tests/test_slerp.py
+++ b/scipy/spatial/tests/test_slerp.py
@@ -1,3 +1,4 @@
+import math
 import warnings
 
 import numpy as np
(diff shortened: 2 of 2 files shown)


Ockenfuss:
wow, thanks @tylerjereddy! Yes, I will have a look at the docs and tests in the next days.

Ockenfuss:
I had a look at the test cases and they basically cover the same cases I would have proposed. Regarding the doc example, you can find an updated version of the example plots [in my branch here](https://github.com/Ockenfuss/scipy/tree/treddy_issue_23400_extrapolation). (When trying to setup and build the scipy documentation, I got stuck into a rabbit hole of errors, but I checked the plotting codes separately).

scipy/spatial/_geometric_slerp.py
```diff
@@ -172,6 +175,26 @@ def geometric_slerp(
...
+    >>> end = np.array([0, 0, 1])
+    >>> t_vals = np.linspace(0, 2, 400)
+    >>> result = geometric_slerp(start,
+    ...                          end,
+    ...                          t_vals)
+    >>> ax.plot(result[...,0],
+    ...         result[...,1],
+    ...         result[...,2],
+    ...         c='k')
+    >>> plt.show()
```
tylerjereddy, line 194: So far the feedback seems to be positive so I've pushed in a docstring adjustment that includes an example of the newly-proposed extrapolation capability.

I'll place the pertinent new parts of the docstring render below the fold.

<details>

image
image


</details>

tylerjereddy: I'll also let the CI flush through now, to see where we stand according to that.

tylerjereddy:
I don't think the free-threaded CI issue is related. `spatial.transform` has a Slerp function as well (in quaternion context). If worst comes to worst we could ask folks maintaining that for a review--there may be enough conceptual similarity that they could have an educated opinion. But this not urgent anyway, I'll move on to other stuff for now.

nickodell:
> I don't think the free-threaded CI issue is related.

I agree. I investigated the issue in more detail, and found that it was a bug in pytest-run-parallel which causes it to no longer be able to auto-detect that `test_pyyaml_not_found` is not thread safe. I've opened a new issue about it.

scipy/spatial/_geometric_slerp.py
```diff
@@ -172,6 +175,26 @@ def geometric_slerp(
...
+    >>> start = np.array([1, 0, 0])
+    >>> end = np.array([0, 0, 1])
+    >>> t_vals = np.linspace(0, 2, 400)
+    >>> result = geometric_slerp(start,
+    ...                          end,
+    ...                          t_vals)
+    >>> ax.plot(result[...,0],
+    ...         result[...,1],
+    ...         result[...,2],
+    ...         c='k')
```
rgommers: Nit: the line breaks look unusual, the above 7 lines can be 2 lines instead.

scipy/spatial/tests/test_slerp.py
```diff
@@ -416,3 +403,64 @@ def test_interpolation_param_ndim(self, t):
             geometric_slerp(start=arr1,
                             end=arr1,
                             t=t)
+
+
+    @pytest.mark.parametrize("start, end, t, expected", [
+    # verify intuitive cases for t > 1 or < 0 (extrapolation)
```
rgommers: This needs a 4-space indent for the whole block for readability

scipy/spatial/tests/test_slerp.py
```diff
@@ -416,3 +403,64 @@ def test_interpolation_param_ndim(self, t):
...
+    ([0, 0, 1], [0, 0, -1], -2, [0, 0, 1]),
+    # move "backwards" to South pole via an infinite number
+    # of possible routes
+    ([0, 0, 1], [0, 0, -1], -1, [0, 0, -1]),
+    ])
+    def test_extrapolation_antipodes(self, start, end, t, expected):
+        with pytest.warns(UserWarning, match='antipodes'):
+            actual_path = geometric_slerp(start=start,
+                                          end=end,
+                                          t=t)
```
rgommers: fits on a single line

scipy/spatial/tests/test_slerp.py
```diff
@@ -416,3 +403,64 @@ def test_interpolation_param_ndim(self, t):
...
+    # North to South on unit sphere, then back to North again;
+    # there are an infinite number of possible routes, but they all
+    # return back to North
+    ([0, 0, 1], [0, 0, -1], -2, [0, 0, 1]),
+    # move "backwards" to South pole via an infinite number
+    # of possible routes
+    ([0, 0, 1], [0, 0, -1], -1, [0, 0, -1]),
+    ])
+    def test_extrapolation_antipodes(self, start, end, t, expected):
+        with pytest.warns(UserWarning, match='antipodes'):
```
rgommers, line 463: This doesn't look quite right. If extrapolation is supported, it should be working without warnings being emitted. If this warning filter is for a subset of the test cases, can you please split the list of tests into two?

rgommers (commented):
Thanks @tylerjereddy, this overall looks good to me and the discussion about why to add this seems to have converged. A few comments.

@Ockenfuss did you want to look at this more?

Ockenfuss:
@rgommers I posted my version of the docs and examples [in this fork here](https://github.com/Ockenfuss/scipy/commit/23625936341ee6da0404514179445a5432890da9), but tylerjereddy's version also looks good to me. So there's nothing against a merge from my side!

tylerjereddy:
@rgommers ok, I've revised to address all comments except for https://github.com/scipy/scipy/pull/23426#discussion_r2312305016. The antipodal warnings are intended and documented--we could remove them for both interpolated and extrapolated scenarios, but I'm not sure we should do that here.

The original discussion about this is at https://github.com/scipy/scipy/pull/10380#discussion_r336760595. It looks like I was pushing for an error for antipodes six years ago, but ended up with a warning. I still can't decide what is best--we return one of the infinite possible paths between poles in i.e. 3D, and the lack of a unique solution makes me think that a warning might still suit. I'm not super attached to that idea though, and I suppose removing a warning isn't a huge backcompat break, but this would also affect the original interpolation behavior added in 2020, so perhaps I could convince you to delay on that adjustment (both test cases in this new test are intentionally checking for this warning when extrapolating, for consistency with interpolating).

rgommers (approved):
Sure, I'm fine with your assessment that a warning for the antipodal usage is best - thanks for the context. 

All LGTM now, thanks @tylerjereddy and @Ockenfuss!

PR #23433, opened 2025-08-04
MAINT:optimize.minimize: Restructure solver C code and some fixes

### LINKED ISSUE #23430: BUG: scipy.optimize is missing error checking for slsqp inputs
### Describe your issue.

When trying to upgrade to SciPy 1.16 in our code base, I saw this error:

```
...
File ".../scipy/optimize/_minimize.py", line 796, in minimize
    res = _minimize_slsqp(fun, x0, args, jac, bounds,
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../scipy/optimize/_slsqp_py.py", line 517, in _minimize_slsqp
    slsqp(state_dict, fx, g, C, d, x, mult, xl, xu, buffer, indices)
SystemError: <built-in function slsqp> returned a result with an exception set
```

When building with Py_DEBUG mode, this became:
```
...
Fatal Python error: _Py_CheckFunctionResult: a function returned a result with an exception set
Python runtime state: initialized
TypeError: 'float' object cannot be interpreted as an integer
```

I traced the problem in a debugger to this line of code:

https://github.com/scipy/scipy/blob/93af8eeed5b33ea6b1d0bbce62eb877b5c156d3d/scipy/optimize/__slsqp.h#L262

If `PyLong_AsLong` fails, e.g., because the input was a python float, we need to return NULL or we'll get the above error from the python interpreter.

I presume the same problem exists here also: https://github.com/scipy/scipy/blob/93af8eeed5b33ea6b1d0bbce62eb877b5c156d3d/scipy/optimize/__slsqp.h#L255

@ilayn (https://github.com/scipy/scipy/pull/22524)

### Reproducing Code Example

```python
Don't have one but can make one if needed.
```

### Error message

```shell
See above.
```

```shell
SciPy 1.16.1, Python 3.12.
```

ilayn:
#### Reference issue
Closes #23429 
Closes #23430

Towards #23067 
Towards #22757

 - Adds multi-phase initializers for SLSQP, NNLS, MINPACK, L-BFGS-B
 - Fixes the reported memory leak in NNLS
 - Guards against non-integer `maxiter` values in SLSQP
 - Extracts and compiles Licenses of old F77 code with added SciPy licenses.
 - Places the source code as standalone libraries under `/src` folder for SLSQP, NNLS, MINPACK, L-BFGS-B
 - Creates new `???module.c` files for the extension modules. 

cc @agriyakhetarpal ; Apologies for shifting sands again but I think this will be much easier for the void->int conversion
cc @hawkinsp ; If you can test this one out for the memory leak, it would be great.

Not sure if this is worth of backporting though.


rgommers (changes requested):
@ilayn thanks for the quick fixes. Can you please split this up though? The bug fixes should be backported to 1.16.x, as well as the symbol hiding fix. The major code restructure isn't backportable though, and very hard to review like this. 

Moving code around, if you want to do that, should be done in a standalone PR with no functional changes, and then the relevant commit hash can be added to `.git-blame-ignore-revs` so `git blame` remains useful for the code in question.

rgommers:
> Adds multi-phase initializers for SLSQP, NNLS, MINPACK, L-BFGS-B

This should also be its own PR.

ilayn:
It seems much but there is only about 20 lines added. I don't know why git is making a mess showing it as such. But I honestly don't have the patience to make multiple PRs for this since it is just text shuffling. I wouldn't mind if someone tackles it though. 

For review purposes, the commits are cleanly separated and the first one is the relevant one for the code changes.

ilayn:
Not sure if it is worth it, but for the backporting purposes, let me try to separate that part so it can be cherry-picked at least.

ilayn:
I separated the changes into cleaner pieces for review. The first two commits are for bug-fixes that can be backported, if desired. The rest is just shuffling things around. I still don't understand why GitHub is showing file moves as file deletion and additions even when I use explicit `git mv` ops.

rgommers:
> I still don't understand why GitHub is showing file moves as file deletion and additions even when I use explicit `git mv` ops.

It shows it like a move just fine if you only move the code (see, e.g., the diff for your third commit). Once you mix moving code and modifying it, that is no longer possible so the diff view on the PR becomes a mess.

rgommers:
I'll try to split this PR up when reviewing. First bit in gh-23442.

agriyakhetarpal:
Thanks @ilayn! I'm stuck in https://github.com/pyodide/pyodide-recipes/pull/141 unfortunately due to a bug in how `PyInit` symbols are being loaded from extension modules, and it is most certainly Pyodide-specific. However, this PR would certainly be helpful. I will leave this to Ralf and the rest of the maintainers for review in detail or split into multiple PRs, though the `void`s-to-`int`s problem is usually overestimated in terms of difficulty in fixing – it is usually only a bunch of `sed` commands that we apply quite liberally, so having a few less files to patch is welcome. I can backport this if it makes it to 1.16.2 or over 1.16.1 if possible cleanly.

ilayn:
@rgommers Do you still have any capacity for the remaining parts? I now have some time to spare and since works such as @andrewellis55 is slowly coming in, I'd like to put the source code in their resting place.

rgommers:
I do, sorry for the delay. I tried adding an AddressSanitizer CI build to see that this bug fix for the memory leak was actually correct and didn't introduce another problem. That turned up multiple other issues in `optimize` to work through first though, so I'm not actually going to get that done any time soon. So back to regular review and merging this in a few parts. I'll see if I can get the next piece in by tomorrow.

rgommers:
Second fix split off in gh-23563.

tylerjereddy:
From a release management standpoint, and a very quick look, it seems that the parts that can be easily backported have been split off, so I'm not intending to try to pull anything else from here for `1.16.2`--let me know if I've misunderstood but I'm going to see if I can get the release process going.

ilayn:
Apologies for making a mess with this one @tylerjereddy. Ralf already sorted it out. I was not able to spare time lately hence it looks disorganized but you already have the necessary changes extracted out for 1.16.2; once with #23442 and then with the memory leak one for nnls and sqrtm. 

The rest of this PR is reorg of files and licenses that you can ignore for now.

rgommers:
@ilayn the next bit I'd like to split off is moving the code and extracting licenses into separate files (which actually overlaps for `lbfgsb` with gh-23451). Looking at the diff in this PR, I'd like to make one change: have separate submodules per solver (e.g., `src/lbfgsb/`). They are independent solvers, and if they're in separate subdirs then it's clearer that the license in each subdir applies to the files right next to it. WDYT?

ilayn:
I think the best is to close this one if you agree and do it from scratch. Since the remaining changes are mostly file renaming and move operations, there is not much left to salvage from this one, especially the license work has progressed already.

Should we add our license to the rewrites?

rgommers:
> Should we add our license to the rewrites?

I'm more thinking a README that just mentions "the top-level `LICENSE.txt` license applies to these files as well, because SciPy contributors have made major changes to these sources". Including many copies of the same SciPy license file doesn't seem ideal.

> I think the best is to close this one if you agree and do it from scratch. 

I was thinking to leave it open as a reminder to actually do the work of this again, and close it when at least most of it has landed in other PRs - otherwise it's pretty likely to fall through the cracks for me.

ilayn:
>  Including many copies of the same SciPy license file doesn't seem ideal.

Indeed that's not ideal. There is some interest to reuse the C sources outside SciPy project. Not sure if we should add BSD with clause at the top of source files. The original F77 source licenses can be handled with PEP639 but then not sure if we need yet another copy of our BSD license for external use. I think the implication of being attached to SciPy project is sufficient with a one-liner.

>  They are independent solvers, and if they're in separate subdirs then it's clearer that the license in each subdir applies to the files right next to it. WDYT?

Makes sense. The idea for the rest of this PR was to empty out the header files and move the wrappers to `xxxmodule.c` files. Feels like we just made this practice up back in the day but I guess it is now familiar to everyone.

nickodell:
@ilayn Is this PR still targeting 1.17.0? If not perhaps we should bump the milestone.

tylerjereddy:
Yeah, there's a decent volume of compiled code changes here, and merge conflicts--maybe a bit risky to try to merge 1-2 days before the branch point--I'll bump the milestone.

PR #23436, opened 2025-08-05
ENH: fft: switch from `pocketfft` to `ducc0.fft`

### LINKED ISSUE #23318: ENH: Switching from `pocketfft` to `ducc0.fft`

This is an issue to track the changes I'm currently implementing at https://github.com/mreineck/scipy/tree/ducc0_fft.
Only when this is somewhat converged, I'll open a PR to have this merged to `scipy` itself.

_No response_

_No response_

_No response_

mreineck:
#### Reference issue
Previous discussion in #23318

This switches the built-in FFT library from `pocketfft` to its successor `ducc.fft`

#### Additional information
Additional work required:
- [x] decide on how to integrate the external sources into `scipy` (currently done via external submodule, but @rgommers suggested using a vendoring script)
- [x] update `scipy/fft/_duccfft/version.md`
- [x] update `ChangeLog / release notes`

#### Release Note

> SciPy's internal FFT backend has switched from `pocketfft` to its successor package `ducc0.fft`, which features several incremental improvements. The most significant of those from SciPy's perspective is probably that storage requirements for internally cached plans have been significantly reduced for most long 1D transforms. Plans that require more storage than 1MB will no longer be cached; this mainly affects huge 1D transforms of prime and near-prime sizes.


mreineck:
Most test regressions seem to appear in STFT-related tests ... not sure what to make of them. Possibly less accurate trigonometric functions in the math libraries on the targets in question?

Others are Windows-related and seem to cause a `Windows fatal exception: code 0xc00000ff` when trying to load the `ducc0` module(?)

rgommers:
> Most test regressions seem to appear in STFT-related tests ... not sure what to make of them. Possibly less accurate trigonometric functions in the math libraries on the targets in question?

Capturing for posterity:
```
FAILED scipy/signal/tests/test_spectral.py::TestSTFT::test_roundtrip_nola_not_cola - AssertionError: 
Not equal to tolerance rtol=1e-07, atol=2e-15
Signal values  of `istft_wrapper()` differ from `istft()`
Mismatched elements: 25 / 998 (2.51%)
Max absolute difference among violations: 8.25067668e-14
Max relative difference among violations: 4.
```

@DietBru any chance you could look at these?

rgommers:
> Others are Windows-related and seem to cause a `Windows fatal exception: code 0xc00000ff` when trying to load the `ducc0` module(?)

Did you change the implementation of `DUCC0_NO_LOWLEVEL_THREADING` compared to how the corresponding implementation in PocketFFT was done?

mreineck:
I made several changes to the multithreading code over the years, but the low-level interaction with `pthreads` that was implemented by @peterbell10 should still be the same. If I'm interpreting my search results correctly, the error code is `STATUS_BAD_FUNCTION_TABLE`, which doesn't really help me understand the issue.
As a standalone package, `ducc` builds and runs fine with Windows, so it's probably a bad interaction between the code and `scipy`'s build system.

DietBru:
> > Most test regressions seem to appear in STFT-related tests ... not sure what to make of them. Possibly less accurate trigonometric functions in the math libraries on the targets in question?
> 
> Capturing for posterity:
> 
> ```
> FAILED scipy/signal/tests/test_spectral.py::TestSTFT::test_roundtrip_nola_not_cola - AssertionError: 
> Not equal to tolerance rtol=1e-07, atol=2e-15
> Signal values  of `istft_wrapper()` differ from `istft()`
> Mismatched elements: 25 / 998 (2.51%)
> Max absolute difference among violations: 8.25067668e-14
> Max relative difference among violations: 4.
> ```
> 
> @DietBru any chance you could look at these?

Since the methods of calculating an inverse STFT differs in `ShortTimeFFT` from the legacy `istft()` differ substantially, it is difficult to assess, if there is a accuracy problem in the new FFT implementation. The original motivation for writing the `istft_wrapper()` was to ensure that the new  `ShortTimeFFT` implementation replicates all legacy `istft()` functionality. Given that `istft()` is now legacy and that I am quite confident that there are no significant numeric problems in `ShortTimeFFT.istft()`, the  `istft_wrapper()` does not really serve a function anymore.

It's interesting that the test fails only on certain architectures. That problem is not new as the following [code snippet] from `istft_compare()` shows:
```python
# Adapted tolerances to account for resolution loss:
    atol = np.finfo(x.dtype).resolution*2  # instead of default atol = 0
    rtol = 1e-7  # default for np.allclose()

    # Relax atol on 32-Bit platforms a bit to pass CI tests.
    #  - Not clear why there are discrepancies (in the FFT maybe?)
    #  - Not sure what changed on 'i686' since earlier on those test passed
    if x.dtype == np.float32 and platform.machine() == 'i686':
        # float32 gets only used by TestSTFT.test_roundtrip_float32() so
        # we are using the tolerances from there to circumvent CI problems
        atol, rtol = 1e-4, 1e-5
    elif platform.machine() in ('aarch64', 'i386', 'i686'):
        atol = max(atol, 1e-12)  # 2e-15 seems too tight for 32-Bit platforms

    assert_allclose(x_wrapper[k_lo:k_hi], x[k_lo:k_hi], atol=atol, rtol=rtol,
                    err_msg=f"Signal values {e_msg_part}")
```
I do not understand why the results are architecture dependent, since there is only inverse FFTs and scaling happening. Concerning the signal module, I do not see any problem in adding looser platform specific tolerances to make the tests pass.



[code snippet]: https://github.com/scipy/scipy/blob/0f2342f34cc850e7137a74afd33c6877c1c790e6/scipy/signal/tests/_scipy_spectral_test_shim.py#L295

kwsp:
@mreineck Is there anything I can help with to finish this PR?

mreineck:
My problem is that I don't really know what is causing the problem, and that I don't have access to Windows machine, or the expertise to debug what's happening there. This needs to be looked at by someone with very good knowledge of `scipy` internals.

LICENSES_bundled.txt
```diff
@@ -87,10 +87,10 @@ License: BSD-3-Clause AND MIT AND BSD-3-Clause-LBNL AND Apache-2.0 WITH LLVM-exc
   For details, see subprojects/xsf/LICENSE and
   subprojects/xsf/LICENSES_bundled.txt
 
-Name: pypocketfft
-Files: scipy/fft/_pocketfft/pypocketfft.cxx
-License: BSD-3-Clause
-  For details, see scipy/fft/_pocketfft/LICENSE.md
```
lucascolley, line 93: I pushed some commits to switch over to vendoring `ducc` under the `subprojects` directory — should we include the main `ducc` license here?

subprojects/ducc/meson.build
lucascolley, line 1: @mreineck ideally this could be upstreamed into the main `ducc` repo. Of course I have only included what we need for the `duccfft_dep` here.

scipy/fft/_duccfft/version.md
lucascolley, line 1: I wonder whether we should just delete this.

scipy/fft/_duccfft/meson.build
```diff
@@ -1,37 +1,38 @@
...
 fft_deps = []
 if is_windows
   if is_mingw
-    # mingw-w64 does not implement pthread_pfork, needed by pocket_fft
+    # mingw-w64 does not implement pthread_pfork, needed by ducc_fft
     # Disable threading completely, because of freezes using threading for
     # mingw-w64 gcc: https://github.com/mreineck/pocketfft/issues/1
-    pocketfft_threads += ['-DPOCKETFFT_NO_MULTITHREADING']
+    duccfft_threads += ['-DDUCC0_NO_LOWLEVEL_THREADING']
   else
```
lucascolley, line 9: Not sure how relevant things like this are for other users, but in theory things like this could be upstreamed into the `duccfft_dep` definition

lucascolley (commented):
I merged main, updated to idiomatic vendoring practice, and triggered CI, let's see how we are looking.

lucascolley:
Looks like two of the Windows jobs get the fatal Python crash, while another hits `E   ImportError: DLL load failed while importing pyduccfft: A dynamic link library (DLL) initialization routine failed.`

LICENSES_bundled.txt
mreineck: I don't have a strong opinion on this. If this is intended for end users, it might be nice to have the full licensing info there (FFT and its dependencies is BSD3, rest GPL2), so that it's clear that not all the code in there is BSD.
If it's intended to document the licenses of things `scipy` actually depends on, it's probably best to just leave it at BSD-3-clause.

lucascolley: maybe we need to fork `ducc` and remove the GPL sections before we can merge this

lucascolley:
@rgommers I could potentially help debug this on a Windows machine, but I don't know where to start

subprojects/ducc/meson.build
mreineck: I'll think about this. "Officially" I don't support partial compilation out of the box, since there are just too many build systems around to make everyone happy.

LICENSES_bundled.txt
mreineck: You are of course welcome to do that, but it sounds like a lot of work. If I recall correctly, the JAX people work with the unchanged repo, and Google tends to be rather picky about licensing questions. Just in case this helps.

lucascolley: > FFT and its dependencies is BSD3, rest GPL2

I guess my concern is that we have `include_directories: 'ducc/src'` which IIUC pulls in GPL2 headers. Probably rather than forking we can just pull in only what we need via a script like https://github.com/scipy/scipy/blob/main/tools/vendoring/vendor_pyprima.sh, but I think that would be easier if the BSD3 headers were in a separate (sub)directory?

mreineck: All the BSD files are marked with 
`/* SPDX-License-Identifier: BSD-3-Clause OR GPL-2.0-or-later */`
Does this help?
I'd prefer not to rearrange the repo in a hurry :-)

lucascolley: > I'd prefer not to rearrange the repo in a hurry :-)

well no hurry, we need to fix the crashes first!

kwsp:
@lucascolley  on windows with DLL issues, my rule of thumb is to start with the Dependencies GUI (https://github.com/lucasg/Dependencies). It shows exactly what symbols the executable or DLL is trying to pull from where. I can debug this tomorrow

kwsp:
I found the Windows runtime error.

In `subprojects/ducc/ducc/src/ducc0/bindings/pybind_utils.h` line 94, 
```cpp
static const auto None = py::none();
```

The static storage duration on a None object means the constructor is called at module import time, which calls `inc_ref()` while the GIL is not held, causing the fatal error when loading `pyduccfft` on Windows. Luckily, this static object is not used anywhere, as the `pyduccfft` module declares its own None instance with automatic storage duration. Why this only errors on Windows and not unices I don't know.

This patch for ducc fixes the SciPy windows build. @mreineck @lucascolley 

```patch
diff --git a/src/ducc0/bindings/pybind_utils.h b/src/ducc0/bindings/pybind_utils.h
index 46cab725..8805f632 100644
--- a/src/ducc0/bindings/pybind_utils.h
+++ b/src/ducc0/bindings/pybind_utils.h
@@ -91,8 +91,6 @@ using namespace std;
 using shape_t=fmav_info::shape_t;
 using stride_t=fmav_info::stride_t;
 
-static const auto None = py::none();
-
 #ifdef DUCC0_USE_NANOBIND
 using NpArr = py::ndarray<py::numpy, py::device::cpu>;
 using CNpArr = py::ndarray<py::numpy, py::ro, py::device::cpu>;
@@ -434,7 +432,6 @@ using detail_pybind::OptNpArr;
 using detail_pybind::CNpArr;
 using detail_pybind::CNpArrT;
 using detail_pybind::OptCNpArr;
-using detail_pybind::None;
 using detail_pybind::isPyarr;
 using detail_pybind::make_Pyarr;
 using detail_pybind::make_Pyarr_and_vmav;

```

scipy/fft/_duccfft/meson.build
kwsp: I like having the Python wrapper's meson build in SciPy. It'll make it easier for me to implement C++ extensions for oaconvolve using the C++ FFT backend.

mreineck:
Thank you very much! That is a really interesting find, since I'm using this construct heavily since a few years and it has never triggered a problem, not even on Windows CI test runs with Github actions. Just curious: any idea why it triggers in this particular situation?

Hmm, why is the GIL not held at load time? Is that because the DLL is not loaded via `dlopen` from Python, but at some other point in time? If the latter, would it be possible to fetch the GIL in the calling code?

kwsp:
In C++ the constructor of a static duration object is called when the program starts (or when the lib is loaded), but before any other code (like `main`) executes. I didn't step through any of the code in a debugger and just read it, but I believe its this code in PyBind11 that throws that error:

https://github.com/pybind/pybind11/blob/cd538ed1184d41cf685dd4da81c1eca80d24f353/include/pybind11/pytypes.h#L280-L284

This was added to pybind11 4 years ago, so maybe you were testing with an older version of pybind11?

mreineck:
I'm keeping `pybind11` up to date, so I should have noticed this already.

> In C++ the constructor of a static duration object is called when the program starts (or when the lib is loaded),

Yes, and under "normal" circumstances, the library is loaded when Python processes the "import" statement. At that time I think the GIL is actually held. But maybe `scipy` is doing things differently and actively loads the library at some other time when the GIL is not active.

I'm OK with changing the problematic object somehow, I would just like to avoid a function call whenever I refer to `None`, because I'm doing that quite a lot in my interface definitions.

kwsp:
> I'm OK with changing the problematic object somehow, I would just like to avoid a function call whenever I refer to None, because I'm doing that quite a lot in my interface definitions.

Isn't the interface definition only ran once (when the module is imported), so it doesn't affect runtime performance?

mreineck:
I believe that things like default argument substitution happen at every single call, but I may be wrong.

mreineck:
I checked the `pybind11` sources and apparently `py::none()` should be really cheap, I'll try to switch from my `None` to `pybind11::none()` and see whether it makes a difference.

kwsp:
In any case, `None` is a special object in Python and is a singleton (cannot call the constructor or destructor from Python's C API), so I'm not sure why PyBind11 is using reference counting for it. (Looks like pybind11 doesn't do any special handling of none unless I read the source wrong, which is possible)

kwsp:
> I'll try to switch from my `None` to `pybind11::none()` and see whether it makes a difference.

https://github.com/mreineck/scipy/blob/375413e808402dd6d69dd968b54bab4fdccb5fd8/scipy/fft/_duccfft/pyduccfft.cxx#L642

I believe this is fine as it's not causing a crash

mreineck:
You're right of course - much nicer this way!

I have pushed this to the Gitlab and Github repos now.

scipy/fft/_duccfft/meson.build
lucascolley: > I like having the Python wrapper's meson build in SciPy. It'll make it easier for me to implement C++ extensions for oaconvolve using the C++ FFT backend.

The `pyduccfft` extension module part will stay in SciPy for sure, yeah. I was just thinking that some of the `cpp_args` may belong upstream if they are not SciPy-specific or specific to `pyduccfft`, but rather inherent to `duccfft_dep`.

mreineck:
I don't understand the failing tests: an absolute tolerance of `2e-15` for arrays with a range of at least `[-10;10]` feels way too strict. You can't do many useful calculations on that error budget ...

lucascolley:
> I don't understand the failing tests: an absolute tolerance of `2e-15` for arrays with a range of at least `[-10;10]` feels way too strict. You can't do many useful calculations on that error budget ...

feel free to loosen tolerances to get CI passing and we can check if the adjustments look okay before merge

LICENSES_bundled.txt
lucascolley: @rgommers do you have an opinion on how to orchestrate the vendoring in light of the mixed licensing? Is a script where we pull in only what we need / can use the way to go?

mreineck:
OK, let's see if that helps.

There seems to be some confusion in the testcase concerning the tolerances to use.
It has been relaxed before on 32bit systems because things started to fail ... and I have no idea why 32bit systems should be less accurate than 64bit ones.

mreineck:
> FAILED scipy/signal/tests/test_spectral.py::TestSTFT::test_roundtrip_float32 - AssertionError: 
Not equal to tolerance rtol=1e-07, atol=1e-12

I am totally confused. This is a float32 test with an atol of (originally) 2e-15? I think without input from the test author we won't be able to figure out what is going on here.

lucascolley:
> > FAILED scipy/signal/tests/test_spectral.py::TestSTFT::test_roundtrip_float32 - AssertionError:
> > Not equal to tolerance rtol=1e-07, atol=1e-12
> 
> I am totally confused. This is a float32 test with an atol of (originally) 2e-15? I think without input from the test author we won't be able to figure out what is going on here.

Unless @e-q is about to comment, I'd be happy to trust the judgement of @DietBru .

LICENSES_bundled.txt
rgommers: Yep, a script sounds right to me. And I see I looked at that before: https://github.com/scipy/scipy/issues/23318#issuecomment-3067183287

e-q:
I'd be surprised if we had any particular reasoning for the `atol` of the float32 tests. If there is lingering doubt, I think it could be worthwhile to look at the distribution of absolute and relative mismatch of the roundtrip values on different dtypes and architectures to choose "safe" thresholds and see if there is surprising behavior anywhere

mreineck:
Sorry ... my last crude fix for the tolerances was stupidly wrong. Let's hope I got it right this time.

tools/vendoring/vendor_duccfft.sh
```diff
@@ -0,0 +1,25 @@
...
+mkdir $ROOT_DIR
+mkdir $ROOT_DIR/.tmp
+git clone $REPO_URL $ROOT_DIR/.tmp
+pushd $ROOT_DIR/.tmp
+git checkout $COMMIT_HASH
+popd
+mv -v $ROOT_DIR/.tmp/src/ducc0/bindings/ $ROOT_DIR/bindings/
+mv -v $ROOT_DIR/.tmp/src/ducc0/fft/ $ROOT_DIR/fft/
+mv -v $ROOT_DIR/.tmp/src/ducc0/infra/ $ROOT_DIR/infra/
+mv -v $ROOT_DIR/.tmp/src/ducc0/math/ $ROOT_DIR/math/
```
lucascolley: @mreineck do these directories fall under the compatible license?

tools/vendoring/vendor_duccfft.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+# Vendors duccfft from https://gitlab.mpcdf.mpg.de/mtr/ducc.git
+
+set -o nounset
+set -o errexit
+
+REPO_URL="https://gitlab.mpcdf.mpg.de/mtr/ducc.git"
+COMMIT_HASH="837c60ac4b28801cb6b17113495d4bf4cbd82968"
```
lucascolley: (may want to adjust to a different commit)

subprojects/duccfft/meson.build
```diff
@@ -0,0 +1,12 @@
+project('duccfft', version: '0.38.0')
```
lucascolley: should check that this version matches the commit we vendor

subprojects/duccfft/LICENSE.md
lucascolley, line 1: I copied the license over from pyduccfft here too. If that's what we go with then we should probably add this one to `LICENSES_bundled.txt` too.

lucascolley (commented):
thanks for the feedback @e-q and the fix @mreineck.

I pushed changes to vendor the requirements of `duccfft_dep` with a script instead of a git submodule, see questions below.

tools/vendoring/vendor_duccfft.sh
lucascolley: I guess files like `bindings/array_descriptor.h` do not. If it's not so easy to restructure the repo, would you mind suggesting a patch to this script which only grabs what we are allowed to, @mreineck ?

.gitmodules
```diff
@@ -22,14 +22,14 @@
...
 	path = subprojects/unuran
 	url = https://github.com/scipy/unuran
 
 [submodule "subprojects/xsf"]
 	path = subprojects/xsf
 	url = https://github.com/scipy/xsf.git
+
+[submodule "subprojects/ducc/ducc"]
+	path = subprojects/ducc/ducc
+	url = https://gitlab.mpcdf.mpg.de/mtr/ducc.git
```
lucascolley: need to `git rm subprojects/ducc/ducc`

subprojects/duccfft/meson.build
mreineck: I suggest to update to the latest release once we are converged (currently 0.40).

tools/vendoring/vendor_duccfft.sh
mreineck: I pushed an attempt, please have a look!

.gitmodules
mreineck: If it's OK, I'll leave that to you. I don't know exactly what you mean by this, and the potential for screwing up things is too high for my taste :)

tools/vendoring/vendor_duccfft.sh
lucascolley: thanks, please check https://github.com/scipy/scipy/pull/23436/changes/c33f766c3f26742a7c68a8ffdf7841a111bffe06

mreineck:
Thanks for the commits, @lucascolley !
Unfortunately version 0.40 doesn't contain @kwsp's fix for `py::None` yet, so I'll probably have to release 0.41 soon.

lucascolley (approved):
Provisional approval, modulo bumping to a released version and my open comments regarding licenses and `version.md`

mreineck:
If we need an announcement of the change, where should it go, `doc/source/release/1.18.0-notes.rst`? 
I can take care of it; the single most important point would be that FFT plans are now smaller.

lucascolley:
> If we need an announcement of the change, where should it go, `doc/source/release/1.18.0-notes.rst`? I can take care of it; the single most important point would be that FFT plans are now smaller.

Feel free to just post it as a comment here and a maintainer can copy it across to https://github.com/scipy/scipy/wiki/*-Release-Note-Entries-for-SciPy-1.18.0-(asterisk-makes-this-appear-at-the-top)

mreineck:
Great! So, how about:

`Scipy`'s internal FFT backend switched from `pocketfft` to its successor package `ducc0.fft`, which features several incremental improvements. The most significant of those from `scipy`'s perspective is probably that storage requirements for internally cached plans have been significantly reduced for most long 1D transforms. Plans that require more storage than 1MB will no longer be cached; this mainly affects huge 1D transforms of prime and near-prime sizes..

mreineck:
I made a new release and updated the vendoring script and the sources (no change in the relevant source files).

LICENSES_bundled.txt
```diff
@@ -87,10 +87,10 @@ License: BSD-3-Clause AND MIT AND BSD-3-Clause-LBNL AND Apache-2.0 WITH LLVM-exc
...
+Name: pyduccfft
+Files: scipy/fft/_duccfft/[pyduccfft.cxx]
+License: 3-Clause BSD
+  For details, see scipy/fft/_duccfft/LICENSE.md
```
lucascolley, line 93:
```suggestion
Name: pyduccfft
Files: scipy/fft/_duccfft/pyduccfft.cxx
License: 3-Clause BSD
  For details, see scipy/fft/_duccfft/LICENSE.md

Name: ducc0
Files: subprojects/duccfft/ducc0/*
License: 3-Clause BSD
  For details, see subprojects/duccfft/LICENSE.md
```

scipy/fft/_duccfft/version.md
lucascolley: {delete this file}

lucascolley (approved):
I would like to merge this after applying these two suggestions, @mreineck @rgommers. Do we want a forum post first?

LICENSES_bundled.txt
mreineck: Fine with me! No idea whether a forum post is needed.

rgommers: There is no user-visible change, right? If not, and there's no other impact that requires team-wide discussion, then probably no need for a forum post either.

scipy/fft/_duccfft/pyduccfft.cxx
```diff
@@ -1,32 +1,35 @@
...
  *  Python interface.
  *
- *  Copyright (C) 2019 Max-Planck-Society
+ *  Copyright (C) 2019-2025 Max-Planck-Society
  *  Copyright (C) 2019 Peter Bell
  *  \author Martin Reinecke
  *  \author Peter Bell
  */
 
-// Note: do not reorder, pybind11.h must come first (it includes Python.h)
```
rgommers, line 15: The changes here look wrong, stdlib modules are now being included before the `pybind11` include.

rgommers:
Great to see that we have a choice between `pybind11` and `nanobind` - that will make it easier to switch over. Is there a measurable performance benefit from `nanobind` @mreineck?

scipy/fft/_duccfft/pyduccfft.cxx
lucascolley: (reference for those not familiar, https://docs.python.org/3/c-api/intro.html#include-files)

subprojects/duccfft/ducc0/bindings/pybind_utils.h
```diff
@@ -0,0 +1,461 @@
...
+#include <nanobind/nanobind.h>
+#include <nanobind/ndarray.h>
+#include <nanobind/stl/vector.h>
+#include <nanobind/stl/string.h>
+#include <nanobind/stl/complex.h>
+#include <nanobind/stl/optional.h>
+#include <nanobind/stl/variant.h>
+#include <nanobind/stl/function.h>
+#else
+#include <pybind11/pybind11.h>
```
lucascolley: looks like this also contradicts https://docs.python.org/3.15/c-api/intro.html#include-files

LICENSES_bundled.txt
rgommers: this change looks right

rgommers:
Did we review the SIMD support? I don't see `DUCC0_NO_SIMD` usage so quickly.

LICENSES_bundled.txt
mreineck: The main user-visible change should be that it's now almost impossible to make the system run out of memory by doing many huge 1D FFTs of different lengths.

mreineck:
> Great to see that we have a choice between `pybind11` and `nanobind` - that will make it easier to switch over. Is there a measurable performance benefit from `nanobind` @mreineck?

I don't see siginificant performance advantages from switching to `nanobind`; my motivation was to have support for arrays living on GPUs in the future.

BTW, switching to `nanobind` implies losing support for `long double` ... not sure if this acceptable for `scipy`.

scipy/fft/_duccfft/pyduccfft.cxx
mreineck: Most likely my fault; I wasn't aware that this kind of horrible low-level interaction is still a thing.
Let me try to fix this.

rgommers:
> BTW, switching to `nanobind` implies losing support for `long double` ... not sure if this acceptable for `scipy`.

That'd be fine, might need a deprecation but support was always inconsistent and it's a bad type that is completely useless on Windows and macOS arm64, so we have no problem deprecating it.

For now we won't switch, but we may want to to gain features like support for the CPython Stable ABI and DLPack.

scipy/fft/_duccfft/pyduccfft.cxx
rgommers: Unfortunately I don't think CPython is ever going to fix that.

subprojects/duccfft/ducc0/bindings/pybind_utils.h
mreineck: Changing that will require a new `ducc0` tag - or we patch it locally for the time being.

mreineck: I *think* this should be fixed now ... please double-check!

scipy/fft/_duccfft/pyduccfft.cxx
mreineck: I'm glad it doesn't seem to have affected my code so far ... but now it should be compliant!

mreineck:
> Did we review the SIMD support? I don't see `DUCC0_NO_SIMD` usage so quickly.

Let me know if I can help with that. Generally, SIMD support hasn't changed much since `pocketfft`; but I switch to `<experimental/simd>` (which has been voted into C++26, as far as I understand) , if I detect the header file and trust the compiler.

rgommers:
Well I mostly meant that we can't be using that SIMD support, because it doesn't generate portable binaries. So we need to define `DUCC0_NO_SIMD` in a `meson.build` file, and I don't see that being done.

mreineck:
Ah, sorry, I didn't get that one!

I don't think you have to disable SIMD explicitly, as long as  ~there is no optimization~ you do not use any compiler flags for specific targets (or `-march=native`). That will at least allow usage of SSE2 on `x86_64`.
If the code doesn't find any set of SIMD instructions it can use, it will fall back to scalar mode anyway. `DUCC0_NO_SIMD` is mainly for testing purposes.

rgommers:
Ah great, I missed that. Yes, SSE2 is fine - SSE3 would be too. AVX2 isn't.

subprojects/duccfft/ducc0/bindings/pybind_utils.h
lucascolley: looks fixed, thanks!

lucascolley:
alright, seems ready to merge from my side then!

lucascolley:
alright, let me click this in. I'm confident that, in the case that complications are discovered post-merge, @mreineck will be able to help resolve them.

Thanks all involved here!

mreineck:
I'll certainly be here to look at any problems that might turn up!
There's a chance that some users might complain about slight performance degradations with repeated, large 1D FFTs, but that's the price to pay for the size-limited plan management.
The code itself has not changed significantly over the last few years and should have been tested extensively as part of JAX and a few other packages, so I'm optimistic that there won't be too many actual bugs left.

PR #23438, opened 2025-08-05
DOC: fix documentation bug in `Rotation.as_davenport`

Nevsor:
[docs only]

Fix the documentation giving the wrong range of possible values for the second axis when lambda is negative.

If lambda is negative, the code will flip the second axis, calculate it with the now positive value of lambda and flip the sign of the second calculated angle. Therefore the range of possible values for negative lambda is the range of possible values for positive lambda, but mirrored at zero.

The following code demonstrates this behaviour:

```py
from scipy.spatial.transform import Rotation
import numpy as np

# positive lambda:
axes = [
     [1,0,0],
     [0,1,0],
     [1,0,1],
]

# negative lambda:
axes_flipped = [
      [1,0,0],
      [0,-1,0],
      [1,0,1],
]

n1, n2, n3 = (axis.direction for axis in axes)
print(f"λ = {np.rad2deg(np.arctan2(np.cross(n1, n2).dot(n3), n1.dot(n3)))}")
# λ = 45.0

rot = Rotation.from_rotvec(np.deg2rad([0,90,0]))

np.rad2deg(Rotation.identity().as_davenport(axes, "extrinsic"))
# array([0.00000000e+00, 6.36110936e-15, 0.00000000e+00])
np.rad2deg(rot.as_davenport(axes, "extrinsic"))
# array([ 0., 90.,  0.])
np.rad2deg(rot.inv().as_davenport(axes, "extrinsic"))
# array([-1.80000000e+02, -6.36110936e-15,  1.80000000e+02])

np.rad2deg(Rotation.identity().as_davenport(axes_flipped, "extrinsic"))
# array([ 0.00000000e+00, -6.36110936e-15,  0.00000000e+00])
np.rad2deg(rot.as_davenport(axes_flipped, "extrinsic")
# array([  0., -90.,   0.])
np.rad2deg(rot.inv().as_davenport(axes_flipped, "extrinsic"))
# array([-1.80000000e+02,  6.36110936e-15,  1.80000000e+02])
```

### DIFF
--- a/scipy/spatial/transform/_rotation.py
+++ b/scipy/spatial/transform/_rotation.py
@@ -1299,7 +1299,7 @@ class Rotation:
             - Second angle belongs to a set of size 180 degrees,
               given by: ``[-lambda, 180 - lambda]``, if ``lambda ≥ 0`` or
               ``[-180 - lambda, -lambda]``, if ``lambda < 0``. Here ``lambda``
-              is the angle in [-180, 180] that rotates axis 3 unto axis 1 
+              is the angle in [-180, 180] that rotates axis 3 onto axis 1 
               using axis 2 as the rotation vector.
 
         References
--- a/scipy/spatial/transform/_rotation.py
+++ b/scipy/spatial/transform/_rotation.py
@@ -1297,7 +1297,8 @@ class Rotation:
             - First angle belongs to [-180, 180] degrees (both inclusive)
             - Third angle belongs to [-180, 180] degrees (both inclusive)
             - Second angle belongs to a set of size 180 degrees,
-              given by: ``[-abs(lambda), 180 - abs(lambda)]``, where ``lambda``
+              given by: ``[-lambda, 180 - lambda]``, if ``lambda ≥ 0`` or
+              ``[180 - lambda, -lambda]``, if ``lambda < 0``. Here ``lambda``
               is the angle between the first and third axes.
 
         References
--- a/scipy/spatial/transform/_rotation.py
+++ b/scipy/spatial/transform/_rotation.py
@@ -1299,7 +1299,8 @@ class Rotation:
             - Second angle belongs to a set of size 180 degrees,
               given by: ``[-lambda, 180 - lambda]``, if ``lambda ≥ 0`` or
               ``[-180 - lambda, -lambda]``, if ``lambda < 0``. Here ``lambda``
-              is the angle between the first and third axes.
+              is the angle in [-180, 180] that rotates axis 3 unto axis 1 
+              using axis 2 as the rotation vector.
 
         References
         ----------
--- a/scipy/spatial/transform/_rotation.py
+++ b/scipy/spatial/transform/_rotation.py
@@ -1252,9 +1252,9 @@ class Rotation:
         rotations.
 
         For both Euler angles and Davenport angles, consecutive axes must
-        be are orthogonal (``axis2`` is orthogonal to both ``axis1`` and
-        ``axis3``). For Euler angles, there is an additional relationship
-        between ``axis1`` or ``axis3``, with two possibilities:
+        be orthogonal (``axis2`` is orthogonal to both ``axis1`` and ``axis3``). 
+        For Euler angles, there is an additional relationship between ``axis1`` 
+        or ``axis3``, with two possibilities:
 
             - ``axis1`` and ``axis3`` are also orthogonal (asymmetric sequence)
             - ``axis1 == axis3`` (symmetric sequence)
--- a/scipy/spatial/transform/_rotation.py
+++ b/scipy/spatial/transform/_rotation.py
@@ -1298,7 +1298,7 @@ class Rotation:
             - Third angle belongs to [-180, 180] degrees (both inclusive)
             - Second angle belongs to a set of size 180 degrees,
               given by: ``[-lambda, 180 - lambda]``, if ``lambda ≥ 0`` or
-              ``[180 - lambda, -lambda]``, if ``lambda < 0``. Here ``lambda``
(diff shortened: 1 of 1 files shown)


scottshambaugh (commented):
I haven't used davenport angles extensively, but am confused by the docs here - as written it is unclear why lambda would be a negative angle at all, since angles between vectors should always be strictly positive. Is it a rotation angle about the 2nd axis?

Also, would you mind fixing the typo in "consecutive axes must **be are** orthogonal" a few paragraphs up while you're working on this section?

Nevsor:
> I haven't used davenport angles extensively, but am confused by the docs here - as written it is unclear why lambda would be a negative angle at all, since angles between vectors should always be strictly positive. Is it a rotation angle about the 2nd axis?

Exactly λ is the rotation angle around axis2 that would rotate axis3 to align with axis1.

Nevsor:
> Also, would you mind fixing the typo in "consecutive axes must **be are** orthogonal" a few paragraphs up while you're working on this section?

Done

Nevsor:
I added another commit that clarifies the meaning of lambda.

scipy/spatial/transform/_rotation.py
```diff
@@ -1298,8 +1298,10 @@ def as_davenport(self, axes: ArrayLike, order: str, degrees: bool = False) -> Ar
             - First angle belongs to [-180, 180] degrees (both inclusive)
             - Third angle belongs to [-180, 180] degrees (both inclusive)
             - Second angle belongs to a set of size 180 degrees,
-              given by: ``[-abs(lambda), 180 - abs(lambda)]``, where ``lambda``
-              is the angle between the first and third axes.
+              given by: ``[-lambda, 180 - lambda]``, if ``lambda ≥ 0`` or
+              ``[-180 - lambda, -lambda]``, if ``lambda < 0``. Here ``lambda``
+              is the angle in [-180, 180] that rotates axis 3 unto axis 1
```
scottshambaugh:
```suggestion
              is the angle in [-180, 180] that rotates axis 3 onto axis 1 
```
One small typo, then I think it looks good!

scottshambaugh (approved):

scipy/spatial/transform/_rotation.py
```diff
@@ -1297,8 +1297,10 @@ def as_davenport(self, axes: ArrayLike, order: str, degrees: bool = False) -> Ar
             - First angle belongs to [-180, 180] degrees (both inclusive)
             - Third angle belongs to [-180, 180] degrees (both inclusive)
             - Second angle belongs to a set of size 180 degrees,
-              given by: ``[-abs(lambda), 180 - abs(lambda)]``, where ``lambda``
-              is the angle between the first and third axes.
+              given by: ``[-lambda, 180 - lambda]``, if ``lambda ≥ 0`` or
+              ``[-180 - lambda, -lambda]``, if ``lambda < 0``. Here ``lambda``
+              is the angle in [-180, 180] that rotates axis 3 onto axis 1 
+              using axis 2 as the rotation vector.
```
lucascolley: It looks a little off having code-formatted text here, but the bullet points for the first and third angles having no special formatting. Maybe the other two bullet points should be code-formatted too?

Or perhaps move this to a new `Notes` section and use LaTeX: https://numpydoc.readthedocs.io/en/latest/format.html#notes

lucascolley: with the latter, you could just say "see notes for the ranges of the returned angles" like https://docs.scipy.org/doc/scipy-1.16.1/reference/generated/scipy.interpolate.AAA.html

lucascolley:
@Nevsor are you interested in returning to address the remaining comments? The deadline for the next SciPy release is coming up soon.

tylerjereddy:
Doc fixes are most welcome, but I'm going to bump the milestone as we're about to cut the release branch this weekend. If it does get merged before branching please just move the milestone back.

tylerjereddy:
The ping for the original author at https://github.com/scipy/scipy/pull/23438#issuecomment-3575292215 was six months ago with no response since then.

I'll make the same decision I did in the last release cycle and bump the milestone again. If this PR does suddenly come back to life and get merged before we branch, just restore the milestone.

scipy/spatial/transform/_rotation.py
Nevsor: @lucascolley, I changed the formulas in all three bullet points to use inline LaTeX in my newest commit. How do like that approach?

scipy/spatial/transform/_rotation.py
```diff
@@ -1294,11 +1294,17 @@ def as_davenport(self, axes: ArrayLike, order: str, degrees: bool = False) -> Ar
...
+            - Third angle belongs to :math:`[-180°,\ 180°]` (both inclusive)
+            - Second angle belongs to a set of size 180°,
+              given by: 
+              
+                - :math:`[-\lambda,\ 180° {-\lambda}]`, if :math:`\lambda ≥ 0`
+                - :math:`[-180° - \lambda,\ {-\lambda}]`, if :math:`\lambda < 0`
+                
+              Here :math:`\lambda` is the angle in :math:`[-180°,\ 180°]` that rotates 
+              axis 3 onto axis 1 using axis 2 as the rotation vector. (all intervals 
+              inclusive on both sides)
```
lucascolley:
```suggestion
              axis 3 onto axis 1 using axis 2 as the rotation vector
              (all intervals inclusive on both sides).
```

lucascolley (approved):
thanks both, LGTM!

rgommers:
This contains a syntax error, see e.g. https://github.com/scipy/scipy/actions/runs/33639717812/job/100279479989. Reverting.

lucascolley:
ouch, sorry, the build passed locally with `pixi run docs`, not sure what went wrong — thanks for the quick fix. Let me send a new PR

PR #23442, opened 2025-08-06
BUG: fix regression for non-integer `maxiter` in SLSQP

### LINKED ISSUE #23430: BUG: scipy.optimize is missing error checking for slsqp inputs

rgommers:
Closes gh-23430. Fix by @ilayn, split off from gh-23433. I added a regression test and a doc tweak.

### DIFF
--- a/scipy/optimize/_slsqp_py.py
+++ b/scipy/optimize/_slsqp_py.py
@@ -235,7 +235,8 @@ def _minimize_slsqp(func, x0, args=(), jac=None, bounds=None,
         Set to True to print convergence messages. If False,
         `verbosity` is ignored and set to 0.
     maxiter : int
-        Maximum number of iterations.
+        Maximum number of iterations. All inputs are passed through ``int()``.
+        Default value is 100.
     finite_diff_rel_step : None or array_like, optional
         If ``jac in ['2-point', '3-point', 'cs']`` the relative step size to
         use for numerical approximation of `jac`. The absolute step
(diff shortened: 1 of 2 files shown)


rgommers:
I'll plan to squash-merge, just kept commits separate since I cherry-picked @ilayn's one.

tylerjereddy (approved):
I checked that reversion of the source patch causes the regression test to fail for `slsqp` only, as expected I think. I checked that the lines in the failure/traceback match those reported by the JAX developer in the matching ticket, so seems pretty likely to be a fix for that.

I'll squash merge it now for backporting.

tylerjereddy:
Ilan was appropriately captured as a co-author in the squash.

tylerjereddy:
thanks both

PR #23443, opened 2025-08-07
BUG: stats.qmc: Avoid infinite loop in sobol when requesting more than 2**32 samples

### LINKED ISSUE #23409: BUG: stats.qmc.Sobol: stuck in infinite loop in low_0_bit after generating 2**32 samples
### Describe your issue.

After generating 2**32 samples, it appears that low_0_bit can get stuck in an infinite loop.

### Reproducing Code Example

```python
from scipy.stats import qmc
import math


def run():
    batch_size = 2 ** 26
    samples = 2 ** 33
    n_batches = samples // batch_size
    sampler = qmc.Sobol(d=2, bits=64)
    samples = 0

    for i in range(n_batches):
        print(f"at step {i}")
        qmc_samples = sampler.random(n=batch_size)
        samples += len(qmc_samples)
        print(f"generated 2**{math.log2(samples):.1f} samples")

if __name__ == "__main__":
    run()
```

### Error message

```shell
at step 0
generated 2**26.0 samples
at step 1
generated 2**27.0 samples
at step 2
generated 2**27.6 samples
at step 3
generated 2**28.0 samples
at step 4
generated 2**28.3 samples
at step 5
generated 2**28.6 samples
at step 6
generated 2**28.8 samples
at step 7
generated 2**29.0 samples
at step 8
generated 2**29.2 samples
at step 9
generated 2**29.3 samples
at step 10
generated 2**29.5 samples
at step 11
generated 2**29.6 samples
at step 12
generated 2**29.7 samples
at step 13
generated 2**29.8 samples
at step 14
generated 2**29.9 samples
at step 15
generated 2**30.0 samples
at step 16
generated 2**30.1 samples
at step 17
generated 2**30.2 samples
at step 18
generated 2**30.2 samples
at step 19
generated 2**30.3 samples
at step 20
generated 2**30.4 samples
at step 21
generated 2**30.5 samples
at step 22
generated 2**30.5 samples
at step 23
generated 2**30.6 samples
at step 24
generated 2**30.6 samples
at step 25
generated 2**30.7 samples
at step 26
generated 2**30.8 samples
at step 27
generated 2**30.8 samples
at step 28
generated 2**30.9 samples
at step 29
generated 2**30.9 samples
at step 30
generated 2**31.0 samples
at step 31
generated 2**31.0 samples
at step 32
generated 2**31.0 samples
at step 33
generated 2**31.1 samples
at step 34
generated 2**31.1 samples
at step 35
generated 2**31.2 samples
at step 36
generated 2**31.2 samples
at step 37
generated 2**31.2 samples
at step 38
generated 2**31.3 samples
at step 39
generated 2**31.3 samples
at step 40
generated 2**31.4 samples
at step 41
generated 2**31.4 samples
at step 42
generated 2**31.4 samples
at step 43
generated 2**31.5 samples
at step 44
generated 2**31.5 samples
at step 45
generated 2**31.5 samples
at step 46
generated 2**31.6 samples
at step 47
generated 2**31.6 samples
at step 48
generated 2**31.6 samples
at step 49
generated 2**31.6 samples
at step 50
generated 2**31.7 samples
at step 51
generated 2**31.7 samples
at step 52
generated 2**31.7 samples
at step 53
generated 2**31.8 samples
at step 54
generated 2**31.8 samples
at step 55
generated 2**31.8 samples
at step 56
generated 2**31.8 samples
at step 57
generated 2**31.9 samples
at step 58
generated 2**31.9 samples
at step 59
generated 2**31.9 samples
at step 60
generated 2**31.9 samples
at step 61
generated 2**32.0 samples
at step 62
generated 2**32.0 samples
at step 63
generated 2**32.0 samples
at step 64
^C^C^C^C^C^C^C^\Quit (core dumped)
```

```shell
1.17.0.dev0+git20250728.a0520a1 2.2.5 sys.version_info(major=3, minor=12, micro=10, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    detection method: pkgconfig
    found: true
    include directory: /home/nodell/.miniforge3/envs/scipy-dev/include
    lib directory: /home/nodell/.miniforge3/envs/scipy-dev/lib
    name: openblas
      NO_LAPACK=0 NO_LAPACKE= NO_AFFINITY=1 USE_OPENMP=0 PRESCOTT MAX_THREADS=128
    pc file directory: /home/nodell/.miniforge3/envs/scipy-dev/lib/pkgconfig
    version: 0.3.29
  lapack:
    detection method: pkgconfig
    found: true
    include directory: /home/nodell/.miniforge3/envs/scipy-dev/include
    lib directory: /home/nodell/.miniforge3/envs/scipy-dev/lib
    name: openblas
      NO_LAPACK=0 NO_LAPACKE= NO_AFFINITY=1 USE_OPENMP=0 PRESCOTT MAX_THREADS=128
    pc file directory: /home/nodell/.miniforge3/envs/scipy-dev/lib/pkgconfig
    version: 0.3.29
  pybind11:
    detection method: pkgconfig
    include directory: /home/nodell/.miniforge3/envs/scipy-dev/include
    name: pybind11
    version: 2.13.6
Compilers:
  c:
    args: -march=nocona, -mtune=haswell, -ftree-vectorize, -fPIC, -fstack-protector-strong,
      -fno-plt, -O2, -ffunction-sections, -pipe, -isystem, /home/nodell/.miniforge3/envs/scipy-dev/include,
      -DNDEBUG, -D_FORTIFY_SOURCE=2, -O2, -isystem, /home/nodell/.miniforge3/envs/scipy-dev/include
    commands: ccache, cc
    linker: ld.bfd
    linker args: -Wl,-O2, -Wl,--sort-common, -Wl,--as-needed, -Wl,-z,relro, -Wl,-z,now,
      -Wl,--disable-new-dtags, -Wl,--gc-sections, -Wl,--allow-shlib-undefined, -Wl,-rpath,/home/nodell/.miniforge3/envs/scipy-dev/lib,
      -Wl,-rpath-link,/home/nodell/.miniforge3/envs/scipy-dev/lib, -L/home/nodell/.miniforge3/envs/scipy-dev/lib,
      -march=nocona, -mtune=haswell, -ftree-vectorize, -fPIC, -fstack-protector-strong,
      -fno-plt, -O2, -ffunction-sections, -pipe, -isystem, /home/nodell/.miniforge3/envs/scipy-dev/include,
      -DNDEBUG, -D_FORTIFY_SOURCE=2, -O2, -isystem, /home/nodell/.miniforge3/envs/scipy-dev/include
    name: gcc
    version: 13.3.0
  c++:
    args: -fvisibility-inlines-hidden, -fmessage-length=0, -march=nocona, -mtune=haswell,
      -ftree-vectorize, -fPIC, -fstack-protector-strong, -fno-plt, -O2, -ffunction-sections,
      -pipe, -isystem, /home/nodell/.miniforge3/envs/scipy-dev/include, -DNDEBUG,
      -D_FORTIFY_SOURCE=2, -O2, -isystem, /home/nodell/.miniforge3/envs/scipy-dev/include
    commands: ccache, c++
    linker: ld.bfd
    linker args: -Wl,-O2, -Wl,--sort-common, -Wl,--as-needed, -Wl,-z,relro, -Wl,-z,now,
      -Wl,--disable-new-dtags, -Wl,--gc-sections, -Wl,--allow-shlib-undefined, -Wl,-rpath,/home/nodell/.miniforge3/envs/scipy-dev/lib,
      -Wl,-rpath-link,/home/nodell/.miniforge3/envs/scipy-dev/lib, -L/home/nodell/.miniforge3/envs/scipy-dev/lib,
      -fvisibility-inlines-hidden, -fmessage-length=0, -march=nocona, -mtune=haswell,
      -ftree-vectorize, -fPIC, -fstack-protector-strong, -fno-plt, -O2, -ffunction-sections,
      -pipe, -isystem, /home/nodell/.miniforge3/envs/scipy-dev/include, -DNDEBUG,
      -D_FORTIFY_SOURCE=2, -O2, -isystem, /home/nodell/.miniforge3/envs/scipy-dev/include
    name: gcc
    version: 13.3.0
  cython:
    commands: cython
    linker: cython
    name: cython
    version: 3.0.12
  fortran:
    args: -march=nocona, -mtune=haswell, -ftree-vectorize, -fPIC, -fstack-protector-strong,
      -fno-plt, -O2, -ffunction-sections, -pipe, -isystem, /home/nodell/.miniforge3/envs/scipy-dev/include,
      -I/home/nodell/.miniforge3/envs/scipy-dev/include
    commands: /home/nodell/.miniforge3/envs/scipy-dev/bin/x86_64-conda-linux-gnu-gfortran
    linker: ld.bfd
    linker args: -Wl,-O2, -Wl,--sort-common, -Wl,--as-needed, -Wl,-z,relro, -Wl,-z,now,
      -Wl,--disable-new-dtags, -Wl,--gc-sections, -Wl,--allow-shlib-undefined, -Wl,-rpath,/home/nodell/.miniforge3/envs/scipy-dev/lib,
      -Wl,-rpath-link,/home/nodell/.miniforge3/envs/scipy-dev/lib, -L/home/nodell/.miniforge3/envs/scipy-dev/lib,
      -march=nocona, -mtune=haswell, -ftree-vectorize, -fPIC, -fstack-protector-strong,
      -fno-plt, -O2, -ffunction-sections, -pipe, -isystem, /home/nodell/.miniforge3/envs/scipy-dev/include,
      -I/home/nodell/.miniforge3/envs/scipy-dev/include
    name: gcc
    version: 13.3.0
  pythran:
    include directory: ../../.miniforge3/envs/scipy-dev/lib/python3.12/site-packages/pythran
    version: 0.17.0
Machine Information:
  build:
    cpu: x86_64
    endian: little
    family: x86_64
    system: linux
  cross-compiled: false
  host:
    cpu: x86_64
    endian: little
    family: x86_64
    system: linux
Python Information:
  path: /home/nodell/.miniforge3/envs/scipy-dev/bin/python3.12
  version: '3.12'
```

### GDB info

It's getting stuck on this line:

```
  __pyx_v_i = 0;

  /* "scipy/stats/_sobol.pyx":198
 *     """
 *     cdef int i = 0
 *     while x & (1 << i) != 0:             # <<<<<<<<<<<<<<
 *         i += 1
 *     return i + 1
 */
  while (1) {
   // infinite loop here
    __pyx_t_1 = ((__pyx_v_x & (1 << __pyx_v_i)) != 0);
    if (!__pyx_t_1) break;

    /* "scipy/stats/_sobol.pyx":199
 *     cdef int i = 0
 *     while x & (1 << i) != 0:
 *         i += 1             # <<<<<<<<<<<<<<
 *     return i + 1
 * 
 */
```

## StackOverflow link

https://stackoverflow.com/questions/79718895/why-does-generating-the-sobol-sequence-in-scipy-hang-after-completing-25-of-the

nickodell:
Fix bug where low_0_bit could loop forever. Also, add tests for low_0_bit.

#### Reference issue

Closes #23409 

#### First approach

I changed the following things:

* ~~When evaluating a 64-bit Sobol sample, the bit shift was done in 32 bits, before the 32-bit result was promoted to a 64-bit number. The shift is now always done in the same type as `uint_32_64`.~~
* ~~When evaluating `low_0_bit` where x was all ones, the program would loop forever, because `1 << x` can wrap around when x is larger than the number of bits in the type. `low_0_bit` now shifts x right one place each loop, so it is guaranteed to eventually reach zero.~~
   * ~~This changes the return value for `low_0_bit(2**64 - 1)` to 65. I think this is the most natural way to extend this value in this situation. For all previous values of x, `low_0_bit(2**x - 1) == x + 1`.~~
* ~~I experimented with a different change for fixing the infinite loop, which you can see in 6dbb7e5b43bd2a965b933dec001d5e66d2529da9. However, I felt it was too complicated, so I changed it to the version you see here.~~
* ~~I marked `low_0_bit` as cpdef, so that I could call it from Python test code. I wrote some tests for it. I did this because testing this by running stats.qmc.Sobol with 2**32 samples would take too long.~~
* ~~I verified that `draw` calls the C function directly, rather than the Python wrapper.~~

#### Revised approach

The first approach I tried had too high of a performance regression.

The revised version tries to fix this via the smallest possible change.

* To reduce complexity of `low_0_bit()`, replace type specialized function with single function that is used by both 32 and 64 bit code.
* Do bit shift in 64-bit always.
* Avoid infinite loop for 0xFFFFFFFFFFFFFFFF by special-casing this input.

#### Performance notes (outdated)

Since I was introducing a new loop-carried dependency, which can prevent optimization, I wanted to check that this bug fix didn't harm performance too much. I ran the BenchQMCSobol benchmark. In some circumstances, there is a 5% performance regression.

```
$ (cd benchmarks/; asv continuous --factor 1.05 --bench stats.BenchQMCSobol main 5df4c3653c21e30c0fde4c179132cbd24c49ab87 --interleave-rounds -a rounds=10)
[90.00%] · For scipy commit 5df4c365 <njo-sobol-loop> (round 10/10):
[90.00%] ·· Benchmarking virtualenv-py3.12-Cython-meson-python-numpy-pybind11-pytest-pythran
[95.00%] ··· stats.BenchQMCSobol.time_sobol                                                                          ok
[95.00%] ··· ===== ========= ========= ========= ============= ============ ============ =============
             --                                           base2                                       
             ----- -----------------------------------------------------------------------------------
               d       3         10        11          12           14           16            20     
             ===== ========= ========= ========= ============= ============ ============ =============
               1    224±2μs   237±3μs   244±3μs     256±4μs      339±3μs      662±9μs     12.1±0.07ms 
               50   262±3μs   364±6μs   460±3μs     666±7μs     6.79±0.8ms   17.3±0.5ms     222±1ms   
              100   301±4μs   485±6μs   688±6μs   4.83±0.09ms   9.61±0.7ms   30.5±0.4ms     432±3ms   
             ===== ========= ========= ========= ============= ============ ============ =============

[95.00%] · For scipy commit 3fe8b508 <main> (round 10/10):
[95.00%] ·· Building for virtualenv-py3.12-Cython-meson-python-numpy-pybind11-pytest-pythran...
[95.00%] ·· Benchmarking virtualenv-py3.12-Cython-meson-python-numpy-pybind11-pytest-pythran
[100.00%] ··· stats.BenchQMCSobol.time_sobol                                                                          ok
[100.00%] ··· ===== ======= =============
                d    base2               
              ----- ------- -------------
                1      3       227±6μs   
                1      10      239±5μs   
                1      11      248±3μs   
                1      12      260±5μs   
                1      14      336±3μs   
                1      16      630±6μs   
                1      20     11.7±0.1ms 
                50     3       265±4μs   
                50     10      363±5μs   
                50     11      466±6μs   
                50     12      669±9μs   
                50     14    6.90±0.07ms 
                50     16     17.5±0.1ms 
                50     20      224±3ms   
               100     3       317±5μs   
               100     10      499±5μs   
               100     11      689±10μs  
               100     12    4.82±0.08ms 
               100     14     9.76±0.8ms 
               100     16     30.6±0.3ms 
               100     20      438±5ms   
              ===== ======= =============

| Change   | Before [3fe8b508] <main>   | After [5df4c365] <njo-sobol-loop>   |   Ratio | Benchmark (Parameter)                  |
|----------|----------------------------|-------------------------------------|---------|----------------------------------------|
| +        | 630±6μs                    | 662±9μs                             |    1.05 | stats.BenchQMCSobol.time_sobol(1, 16)  |
| -        | 317±5μs                    | 301±4μs                             |    0.95 | stats.BenchQMCSobol.time_sobol(100, 3) |
```

#### Performance notes (new approach)

https://github.com/scipy/scipy/pull/23443#issuecomment-3309912014


scipy/stats/tests/meson.build
```diff
@@ -35,6 +35,7 @@ py3.install_sources([
     'test_resampling.py',
     'test_sampling.py',
     'test_sensitivity_analysis.py',
+    'test_sobol.py',
```
rgommers: A new file for one test is best avoided; another utility function from `_sobol` is already tested in `test_qmc.py` - add to that instead?

rgommers (commented):
Thanks @nickodell, these fixes look useful (infinite loops in corner cases are a pain)!

For the performance regressions, are you sure the 5% is reproducible? And if so, can you make sure it comes from the bug fix changes and not from moving to `cpdef`? The benchmark in question doesn't need the `cpdef` so should be easy to verify that.

scipy/stats/tests/meson.build
nickodell: Fixed in 6926cfea3

nickodell:
> For the performance regressions, are you sure the 5% is reproducible?

Fairly sure. I'm using asv with `--interleave-rounds` and a high round count. I did need to increase base2 to 16 or 20 for the difference to show up, though. I'm guessing that amounts smaller than that get swallowed up by Python overhead.

>And if so, can you make sure it comes from the bug fix changes and not from moving to `cpdef`? The benchmark in question doesn't need the `cpdef` so should be easy to verify that.

Hmm. That may be it. A lot of the difference disappears after taking that out.

rgommers:
Then it may be good to move back to `cdef` and if you want to keep the test, which seems reasonable, add a separate utility `cpdef low_0_bit_py` which forwards to `low_0_bit`?

nickodell:
> Then it may be good to move back to `cdef` and if you want to keep the test, which seems reasonable, add a separate utility `cpdef low_0_bit_py` which forwards to `low_0_bit`?

Sorry for my slow reply.

I benchmarked it more thoroughly, and splitting the cpdef into a cdef and a def doesn't help.

I do have two other ideas:

* It seems like it is accessing memory in an order which is inefficient for C memory layout. The function has more predictable cache usage in Fortran order, which gives an approx 10% speedup.
* One can avoid this loop by using the compiler builtin `_BitScanForward` (MSVC) or `__builtin_ctzll` (gcc/clang/icc). On most architectures this will compile to one or two instructions.

But that is maybe too radical of a change for a bugfix PR.

rgommers:
> But that is maybe too radical of a change for a bugfix PR.

I don't see why, either of the options seem fine to me. The only concern would be if a bug fix needs a backport, because then yes you'd want to split off the larger changes. But this doesn't really need to be backported, so whatever makes the code better is fine I'd say. I'd judge it independently: if the implementation of one or both of your ideas has a good performance/maintainability tradeoff, then it's fine to include and merge. If not, then taking a small performance hit to fix a bug is also fine.

mdhaber:
Shall we get the bugfix in and consider the enhancements in a follow-up?

rgommers:
> Shall we get the bugfix in and consider the enhancements in a follow-up?

I'm always 👍🏼 on being pragmatic, so sounds good in principle. That said, I don't think we want to merge this in its current form, because the performance regression is more important than the corner case infinite loop. One of @nickodell's ideas still needs to be implemented I believe (please correct me if I'm wrong Nick).

nickodell:
I took another crack at this, and tried fixing the bug while making as few changes as possible. The performance difference is very small. (This benchmark difference would be hidden normally, but I am using `--factor 1.01` to show more differences.)

```
$ (cd benchmarks/; asv continuous --factor 1.01 --bench stats.BenchQMCSobol main 6123b2def2 --interleave-rounds -a rounds=10)
...

[90.00%] ·· Benchmarking virtualenv-py3.13-Cython-meson-python-numpy-pybind11-pytest-pythran
[95.00%] ··· stats.BenchQMCSobol.time_sobol                                                                          ok
[95.00%] ··· ===== ========= ========= ========= =============
             --                       base2                   
             ----- -------------------------------------------
               d       3         10        11          12     
             ===== ========= ========= ========= =============
               1    225±3μs   241±3μs   248±3μs     261±2μs   
               50   264±4μs   364±5μs   455±8μs     659±6μs   
              100   299±3μs   489±5μs   679±9μs   6.15±0.02ms 
             ===== ========= ========= ========= =============

[95.00%] · For scipy commit d206e414 <main> (round 10/10):
[95.00%] ·· Building for virtualenv-py3.13-Cython-meson-python-numpy-pybind11-pytest-pythran...
[95.00%] ·· Benchmarking virtualenv-py3.13-Cython-meson-python-numpy-pybind11-pytest-pythran
[100.00%] ··· stats.BenchQMCSobol.time_sobol                                                                          ok
[100.00%] ··· ===== ========= ========= ========== =============
              --                       base2                    
              ----- --------------------------------------------
                d       3         10        11           12     
              ===== ========= ========= ========== =============
                1    225±3μs   239±3μs   246±3μs      259±2μs   
                50   262±4μs   360±4μs   455±7μs      659±9μs   
               100   298±2μs   480±6μs   673±10μs   6.17±0.03ms 
              ===== ========= ========= ========== =============

| Change   | Before [d206e414] <main>   | After [6123b2de] <njo-sobol-loop2>   |   Ratio | Benchmark (Parameter)                   |
|----------|----------------------------|--------------------------------------|---------|-----------------------------------------|
| +        | 480±6μs                    | 489±5μs                              |    1.02 | stats.BenchQMCSobol.time_sobol(100, 10) |

PERFORMANCE DECREASED.
```

I tried two versions of this.

* 0887d9f - This commit fixes the infinite loop at 2 ** 32 samples by doing the bit shift using 64-bit integers. However, it does not fix the infinite loop that occurs at 2 ** 64 samples. That one is less important, as that's harder to hit.
* 6123b2d - This commit also fixes the loop at 2 ** 64 samples, by checking for that corner case. I also removed the type-specialization of low_0_bit, as I don't think it makes any difference to performance, and it complicates checking for the corner-case.

mdhaber:
Hmm several jobs are failing with `meson compile: error: argument -j/--jobs: invalid int value: 'Sentinel.UNSET'`. Does that ring a bell?

nickodell:
It does,  #23642

mdhaber:
The Linux prerelease failure is in other PRs. Haven't seen the CircleCI failure elsewhere yet, but I can't imagine it's related to this PR - more likely to related to the Linux prerelease failure, since both have to do with plotting.

nickodell:
Thanks - was having trouble figuring out why that was failing. For posterity, in case that CI log gets deleted, the error it shows during build_docs is

https://app.circleci.com/pipelines/github/nickodell/scipy/19/workflows/1a854c93-53db-4082-b6a2-443a30588f0c/jobs/76?invite=true#step-104-366235_100

```
/home/circleci/repo/build-install/usr/lib/python3.11/site-packages/scipy/stats/_morestats.py:docstring of scipy.stats._morestats.bayes_mvs:91: WARNING: Exception occurred in plotting scipy-stats-bayes_mvs-1
 from /home/circleci/repo/doc/source/reference/generated/scipy.stats.bayes_mvs.rst:
  File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/matplotlib/sphinxext/plot_directive.py", line 702, in render_figures
    figman.canvas.figure.savefig(img.filename(fmt), dpi=dpi)
  File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/matplotlib/figure.py", line 3490, in savefig
    self.canvas.print_figure(fname, **kwargs)
  File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/matplotlib/backend_bases.py", line 2186, in print_figure
    result = print_method(
             ^^^^^^^^^^^^^
  File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/matplotlib/backend_bases.py", line 2042, in <lambda>
    print_method = functools.wraps(meth)(lambda *args, **kwargs: meth(
                                                                 ^^^^^
  File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/matplotlib/backends/backend_agg.py", line 481, in print_png
    self._print_pil(filename_or_obj, "png", pil_kwargs, metadata)
  File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/matplotlib/backends/backend_agg.py", line 429, in _print_pil
    FigureCanvasAgg.draw(self)
  File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/matplotlib/backends/backend_agg.py", line 382, in draw
    self.figure.draw(self.renderer)
  File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/matplotlib/artist.py", line 94, in draw_wrapper
    result = draw(artist, renderer, *args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/matplotlib/artist.py", line 71, in draw_wrapper
    return draw(artist, renderer)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/matplotlib/figure.py", line 3257, in draw
    mimage._draw_list_compositing_images(
  File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/matplotlib/image.py", line 134, in _draw_list_compositing_images
    a.draw(renderer)
  File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/matplotlib/artist.py", line 71, in draw_wrapper
    return draw(artist, renderer)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/matplotlib/axes/_base.py", line 3226, in draw
    mimage._draw_list_compositing_images(
  File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/matplotlib/image.py", line 134, in _draw_list_compositing_images
    a.draw(renderer)
  File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/matplotlib/artist.py", line 71, in draw_wrapper
    return draw(artist, renderer)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/matplotlib/legend.py", line 753, in draw
    bbox = self._legend_box.get_window_extent(renderer)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/matplotlib/offsetbox.py", line 369, in get_window_extent
    px, py = self.get_offset(bbox, renderer)
  File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/matplotlib/offsetbox.py", line 60, in get_offset
    return meth(params["self"], bbox, params["renderer"])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/matplotlib/offsetbox.py", line 306, in get_offset
    self._offset(bbox.width, bbox.height, -bbox.x0, -bbox.y0, renderer)
  File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/matplotlib/legend.py", line 722, in _findoffset
    x, y = self._find_best_position(width, height, renderer)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/matplotlib/legend.py", line 1180, in _find_best_position
    _api.warn_external(
  File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/matplotlib/_api/__init__.py", line 391, in warn_external
    warnings.warn(message, category, **kwargs)
UserWarning: Creating legend with loc="best" can be slow with large amounts of data. [docutils]
looking for now-outdated files... none found
```

mdhaber:
Thanks @nickodell. Looks like that was a temporary glitch. The other failures are in main.

nickodell:
> Thanks nickodell. Looks like that was a temporary glitch. The other failures are in main.

Thanks for triaging that.

I realized that I forgot to update the cover letter on this PR, so I've done that now.

This should be ready for review, if anyone wants to comment on the approach I used here?

mdhaber:
@rgommers you had started the review here. Do you have a moment to take another look? It is only a few lines, but it requires some familiarity with the background, and it involves Cython, which I haven't used in a long time.

scipy/stats/_sobol.pyx
```diff
@@ -195,7 +195,10 @@ cdef int low_0_bit(uint_32_64 x) noexcept nogil:
 
     """
     cdef int i = 0
-    while x & (1 << i) != 0:
+    if (~x) == 0:
+        # If number is all 1s, the first 0 is at one after the last 1
+        return 65
+    while x & (1L << i) != 0:
         i += 1
     return i + 1
```
WarrenWeckesser: @nickodell, what do you think of this version?
```
    cdef int i = 1
    while x & 1UL:
        x >>= 1
        i += 1
    return i
```
Running the benchmark against the current main branch (thanks for giving the asv incantation!) shows an improvement:

```
| Change   | Before [da7011c6] <main>   | After [d64f66e6] <nickodell-njo-sobol-loop>   |   Ratio | Benchmark (Parameter)                   |
|----------|----------------------------|-----------------------------------------------|---------|-----------------------------------------|
| -        | 645±30μs                   | 597±10μs                                      |    0.93 | stats.BenchQMCSobol.time_sobol(100, 12) |
```
(da7011c6 is the current main branch; d64f66e6 is my local commit of this suggestion.)

nickodell: @WarrenWeckesser Seems simpler than this version; it removes a magic number. I'd be happy doing that. I suggest that you push that change to this branch. I have the "allow edits by maintainers" option checked on this PR.

nickodell:
Merging main to pick up PR #23777. This appears to be the cause of the CI error [here](https://app.circleci.com/pipelines/github/nickodell/scipy/35/workflows/b4f22063-e438-4389-902f-07a90fdf82b3/jobs/141), as it says it is pulling `cimg/python:3.11`.

scipy/stats/_sobol.pyx
nickodell: @WarrenWeckesser I re benchmarked your changes after the merge at 8ed713ab to fix CI, and I also see the performance increase vs. main. (Or at least no regression vs main, which was my only goal.)

```
$ (cd benchmarks/; asv continuous --factor 1.01 --bench stats.BenchQMCSobol upstream/main HEAD --interleave-rounds -a rounds=10)

| Change   | Before [91458d76] <upstream/main>      | After [8ed713ab] <njo-sobol-loop>   |   Ratio | Benchmark (Parameter)                   |
|----------|----------------------------------------|-------------------------------------|---------|-----------------------------------------|
| -        | 248±3μs                                | 244±3μs                             |    0.98 | stats.BenchQMCSobol.time_sobol(1, 11)   |
| -        | 493±7μs                                | 481±6μs                             |    0.98 | stats.BenchQMCSobol.time_sobol(100, 10) |
| -        | 262±3μs                                | 254±3μs                             |    0.97 | stats.BenchQMCSobol.time_sobol(1, 12)   |
```

scipy/stats/_sobol.pyx
```diff
@@ -442,3 +443,8 @@ def _test_find_index(p_cumulative, size, value):
     # type: (np.ndarray, int, float) -> int
     """Wrapper for testing in python"""
     return _find_index(p_cumulative, size, value)
+
+cpdef _test_low_0_bit(cnp.uint64_t x):
```
WarrenWeckesser: Nit: this will be called from Python only, so plain `def` is sufficient:
```suggestion


def _test_low_0_bit(cnp.uint64_t x):
```
(Plus a PEP-8 blank line addition.)

scipy/stats/tests/test_qmc.py
```diff
@@ -883,6 +883,26 @@ def test_64bits(self):
         assert_array_equal(self.unscramble_nd, sample)
 
 
+
```
WarrenWeckesser: Extraneous blank line.
```suggestion
```

WarrenWeckesser (changes requested):
Just a couple nits to pick, then I think this is good to go.

WarrenWeckesser:
@nickodell, I'll make those little changes, let the tests run again (overabundance of caution and all that), then merge when CI shows all green.

WarrenWeckesser (approved):

WarrenWeckesser:
Merged. Thanks @nickodell!

mdhaber:
Thanks, all!

PR #23446, opened 2025-08-07
DOC: special.entr: added example

wongaokay:
[docs only]

#### Reference issue

towards #7168 

Adds examples to `special.entr` (covering cases where x>0)

#### Additional information

I chose not to include examples where x<0, but I'm happy to add them if reviewers feel they would be helpful.


j-bowhay:
Thanks for your contribution @wongaokay. Other reviewers may disagree but I would prefer the example to show a motivated example rather than just plugging numbers in

wongaokay:
Ahh good point @j-bowhay! I agree, maybe it would be clearer if we used an array that could represent a probability distribution? I’ll update the example.

fbourgey:
@j-bowhay this lgtm

Issue #23437, opened 2025-08-05
ENH: spatial.transform: Add broadcasting support


amacati:
# Feature description

At the moment, `Rotation` and `RigidTransform` only accept either single rotations/transforms, or an array of rotations/transforms. This limitation stems from the original Cython implementation that did not use broadcasting. With the newly introduced generic backends which already use broadcasting internally, it is possible to remove this restriction and apply regular broadcasting logic (see #23249 and #23425). 

This  would enable users to use generically shaped rotations, i.e. a batch of 3x2 quaternions.

# Design
Broadcasting rules are well established and offer a straightforward way of determining the result each operation should have. Ideally, we state that all operations in `Rotation`/`RigidTransform` follow standard broadcasting rules. This would remove the arbitrary limitation to a list of rotations/transforms and would make using `Rotation`/`RigidTransform` feel a bit more like dealing with an array wrapper around quaternions/matrices.

## Implementation considerations
Overall, the scope and intended result are clearly defined by building on broadcasting rules. However, the implementation requires some consideration.

### Fully support broadcasting in the backends
The xp backends already make heavy use of broadcasting to produce the correct output shapes without distinction of cases (what a happy coincidence!). However, some functions don't have full broadcasting support yet. These would need to be addressed.

### numpy handling
We currently use the Cython backend for single rotations and lists of rotations. We cannot open this up to arbitrary shapes because of Cython type restrictions. I do not recommend dropping the Cython backend, as it provides speedups for some of the most common cases (single Rotation or smaller lists of Rotations). However, for tensors we could use the xp backends. Users requiring batches with multiple dimensions will probably work with larger data sizes anyways, where the xp backends sometimes become more efficient than the Cython backend. In addition, this minimizes maintenance load. The xp backends are already tested.

### Changes to functions
Some functions (e.g. `mean` or groups) would greatly benefit from additional functionality or require additional parameters for clarification. For example, `mean` should probably take an axis argument if broadcasting is possible.

## Alternative options

We could also expose the functional array API backends to users which use broadcasting for most implementations. However, these functions rely on the frontend for some sanity checks (e.g. they assume input quaternions are valid and normalized), are not documented at the moment and thus are less safe for users to use.

@lucascolley @scottshambaugh @crusaderky

lucascolley:
[...]

PR #23447, opened 2025-08-07
do not merge: experiment with Array API backends for RBFInterpolator

ev-br:
TL;DR: We add Array API backends to RBFInterpolator. The baseline is Pythran AOT-compiled code working with numpy arrays, and we want to generalize to other Array API backends (with eager and JIT-compiled execution models). We also throw `numba` into the mix because why not.

*Goals*: We want to 
- allow various Array API backends (torch, jax, cupy);
- allow execution on GPUs and other accelerators,
- ... without having to write dedicated CUDA kernels;
- ideally, we want to keep the same Python codebase, and just AOT or JIT compile it across backends and devices.

*Maths*:

Interpolation of given data involves two steps: constructing the interpolant, and evaluating it on the input data.
- Constructing an interpolant involves filling in and solving a certain system of linear equations. Let's call its solution the "interpolation coefficients".
- Evaluating the interpolant involves filling in a certain matrix (call it the "kernel matrix") and matrix-multiplying it with the interpolation coefficients.

Typical usage is to construct the interpolant once, and evaluate it multiple times. The "kernel matrix" depends on the input data, so that we construct the kernel matrix multiple times (in the `__call__` method), while the coefficients are fixed and are evaluated only once (in the `__init__` method).

Given the set of $n$ input vectors, $x$ and $y$, the kernel matrix is $K_{ij} = k(|x_i - y_j|)$, where $k(.)$ is a fixed univariate _kernel_ function, and $|.|$ is a vector norm. In numpy-type pseudocode, and glossing over details

```
def kernel_matrix(x, y, kernel_func):
    return kernel_func(np.linalg.vector_norm(x[:, None, :] - y[None, :, :], axis=-1)
```

Here `x` and `y` are 2D arrays which contain the vectors $x_i$ and $y_i$, respectively, stacked along the last dimension.

*Implementation*:

In addition to the existing numpy/pythran backend, we add a "generic" backend which operates on Array API compatible objects. This is sufficient for eager mode of relevant Array API libraries (`pytorch`, `jax.numpy` and `cupy`). For the JIT mode, we manually override select functions of the "generic" backend by their jit-compiled versions. For example, eager pytorch execution uses the generic `_rbfinterp_xp.py` module, and JIT-compiled variants reside in the `_rbfinterpr_dynamo.py` which overrides parts of the "generic" backend module. Likewise, eager execution with JAX arrays uses the "generic" backend, and the JIT-compiled overrides live in the `_rbfinterp_jaxjit.py` module.

Yes, the current implementation is deliberately simplistic. How to best avoid this from evolving into a sprawling mess is
an open question. It is also full well possible that I did something suboptimal for specific backends, especially around JIT compiled parts.
Any kinds of comments are greatly appreciated!

The first lesson from this exercise is that the dream of having a single source for all of pythran/AOT and array-api/JIT
modes is not realistic. There are several reasons, some technical and some conceptual. 

One technical reason is that `pythran` relies on source code analysis; Given an explicit import, "import numpy as np", it knows how to compile `np.linalg.norm` to efficient low-level C code. Our "generic Array API" style implies that functions receive a namespace as an `xp` argument which is resolved at runtime. I did not find an easy way to express "consider this `xp` argument to be fact `numpy`" to `pythran` (one other backend which has the same issue is `numba`).

One conceptual reason is that some backends prefer vectorized code while others prefer explicit loops. An extreme example is `torch.compile`, which agressively unrolls all `for` loops. This way, iterating over an axis of an array in a loop,
`for i in range(x.shape[0]): do_something(x[i])`, actually generates C++ code with `x.shape[0]` copies of `do_something(x[0])`, `do_something(x[1])` and so on.

One other thing I still need to check is whether `jax` requirement of having immutable arrays interferes with `torch.dynamo` relying on object identity to elide temporaries during codegen (so far this did not seem to cause serious problems; I took only a cursory look though).

*Performance benchmarks*:

To assess the performance of various array API backends, we use a simple benchmark, where we construct an interpolator with the fixed sized data to interpolate, and measure the evaluation time for an input array of a varying size. 

We run this benchmark on two machines: a "laptop" (a 4-core intel i5 laptop) and a "server" (a 32-core Threadpiper machine hosted by QS). On the "server" we run the benchmark on a CPU and a CUDA GPU (an RTX 2060 card). 

On CPU, we run the benchmark with:
- (the baseline) numpy arrays + Pythran AOT compiled modules,
- numpy arrays with `numba` JIT compiled evaluations (because why not check out `numba` when we talk JIT),
- torch tensors in the eager mode and `torch.dynamo` JIT compiled evaluations, and
- `jax` arrays in the eager mode and under `jax.jit`.

On GPU, we run the benchmarks with
- `torch`, eager and jit,
- `jax.numpy`, eager and jit, and
- `cupy` eager.

Benchmarking results are available at this Jupyter notebook: https://github.com/ev-br/rbfinterp_xp_timings/blob/master/timings_cpu.ipynb; the timing data and the --- very crude! --- benchmark script are at the same repository next to the notebook.

To summarize the results:

On CPU,

- Unsurprisingly, the eager mode array API backends generally perform worse than the pythran compiled version: `torch` eager is 3-5 times worse than the baseline; `jax` eager runtime depends on the machine: on the laptop it is 30-50% worse, and on the server it is---surprisingly---comparable with the baseline (50% worse to 50% better).
- In the JIT mode, `jax.jit` and `torch.compile` give ~3x speedup on a laptop, and a 5-15x speedup on the server.

These results---to a large degree---are due to automatic parallelization by backends. While `pythran` generated code always runs single-threaded, both `jax` and `torch`, especially in the JIT mode, infer the number of threads from the number of CPU cores. 
There is definitely ample room for tuning the runtime parallelization: first, there are signs of oversubscription due to hyperthreading (e.g. 64 threads run by default on a 32-core Intel machine); second, where LAPACK is used for the vector norm, LAPACK library itself spawns threads (and I saw some interaction of the reported runtime with environment variables like `MKL_NUM_THREADS`). More work is needed to systematically investigate this further.

Quite surprisingly, out of the box, `numba` performs way worse than either `jax.jit` or `torch.compile`. By default, numba-jitted code runs single-threaded, and the runtime is some 5-6 times worse than (also single-threaded!) pythran-compiled version. Using explicit source-level parallelization, `numba.prange`, for for-loops brings its multithreaded performance to be on par with the (single-threaded) baseline.

Also implementation-wise, `numba` seems to be more limiting than either `torch.compile` or `jax.jit`. For instance, it requires explicit for loops in place of "vectorized" evaluations for basic numpy ufuncs (`np.prod(x)` compiles by `np.prod(x, axis=-1)` does not).

On CUDA,

- `cupy` delivers about a 3x speedup relative to the numpy/pythran baseline. I should note however that CuPy's implementation is rather simple, and is similar to the "generic" implementation here. It is quite likely that not having manually written CUDA kernels leaves quite a bit of performance on the table.

- `jax.jit` and `torch.compile` typically get a 5x to 40x perf improvement relative to numpy on CPU. The difference between jit and eager is not as drastic as in CPU, typically within an factor of two. 

The scaling with the problem size is somewhat bizzare though, both on `torch` and `jax`, which hints at a PEBKAC level problem with the benchmarking script.

-------------------------------------

Overall, this is a very crude WIP experiment. I plan to refine it at some point. Meanwhile, any kinds of comments or suggestions or fixes are most welcome.


rgommers:
Thanks for sharing @ev-br. I think the first order of business is to improve the benchmarks so that they can be run with a lot of manual work - that will make it more feasible for others to try this out and weigh in. There's a couple of ways of doing that:

- Improve the current notebook to loop over all available array libraries, save the benchmark results, and then use them in the same notebook so that a single "Run all cells" gets you the graphs you want
- We have support for `asv` benchmarks that run across all `xp` backends at once (see gh-23012) and there is an existing asv benchmark for `RBFInterpolator`

Also @amacati has been producing some very polished benchmarks with graphs across libraries and devices (e.g., gh-23424). I don't see benchmarking code there though, just the end result - @amacati do you have code somewhere?

For the baseline of the benchmark plots, are you comparing with the implementation in `main`? If not, are you sure the refactoring isn't slower? `_rbfinterp_pythran.py` was kept unchanged, but there's some more overhead in how it's plumbed in it looks like.

* * *

Other comments:
- It would be insightful to distinguish between single-threaded (apples-to-apples) and scaling performance. Probably an easy way to do that is to run the benchmarks after using `os.sched_setaffinity` to ensure only a single CPU core is available. Or launching `python` under `taskset`.
- The "write separate code for each array library" approach is interesting - I assume it needs to be compared with "what if we had a single extra `_xp` implementation", since that's going to be the maintainable thing? 
- Please make sure not to reinvent [Transonic](https://transonic.readthedocs.io/en/latest/index.html)

ev-br:
> Also @amacati has been producing some very polished benchmarks with graphs across libraries and devices (e.g., gh-23424). I don't see benchmarking code there though, just the end result - @amacati do you have code somewhere?

Yup, I remember seeing @amacati's code somewhere, but did not manage to find it again, and slapdashed a quick version, if only to get the ball rolling.

asv benchmarks... Yes we have some. I've been looking for a reason to look into an improvement over `asv` based on `pytest-benchmark`. This might be it, now :-). Two directions to work towards:
- make time measurements work across backends and devices (e.g. hide CUDA events/synchronization python API differences)
- decouple a benchmark viewer from the runner.

> For the baseline of the benchmark plots, are you comparing with the implementation in main? If not, are you sure the refactoring isn't slower? _rbfinterp_pythran.py was kept unchanged, but there's some more overhead in how it's plumbed in it looks like.

No rigorous comparisons to main, not yet. The baselines here are using the code in this PR.
There is definitely one thing this PR does at at runtime which should be moved to `__init__` time https://github.com/scipy/scipy/pull/23447/files#diff-cc6e641abc62d2e6ce38c679bdb188c4173ede68f7ce2c03fc34a919bb9225fbR55. Other than that, there's some overhead of `array_namespace` and related array-api decisions. Not sure if we ever measured this beyond "it should be small" though.

> The "write separate code for each array library" approach is interesting - I assume it needs to be compared with "what if we had a single extra _xp implementation", since that's going to be the maintainable thing?
Please make sure not to reinvent [Transonic](https://transonic.readthedocs.io/en/latest/index.html)

This is the original "single python source for everything" dream :-).

Big picture, there are basically two backends: one for pythran and one for "generic xp". backend-jit-specific modules only invoke JIT on select functions. Two directions to unify all this:
- check if pythran handles "vectorized" evaluations, improve pythran if necessary;
- invert the "what to jit logic": it's a user decision not the library's; This probably needs @crusaderky 's `lazy_xp_function` or a generalization.

amacati:
The benchmarks were created outside the regular scipy benchmarking infrastructure for tracking performance during development and for the final PRs, that's why they only live in my own benchmark branch.

https://github.com/amacati/scipy/tree/benchmark/xp_bench and in particular https://github.com/amacati/scipy/blob/benchmark/xp_bench/rot_benchmark.py. These have not been updated for the PR @rgommers was referencing, but I can push that iteration later today if you want. Also note that returning scipy objects from JITted functions requires them to be registered as `PyTree`s in jax. In the case of `Rotation`, the registration code was kept out of scipy. In case you want to run the benchmarks, add this to the top of your file.

```python
from scipy.spatial.transform import Rotation
from jax.tree_util import register_pytree_node

def rot_unflatten(_, c):
    # Optimization: We do not want to call __init__ here because it would perform normalizations
    # twice. More importantly, it would call the non-jitted Array API backend and therefore
    # incur a significant performance hit
    r = Rotation.__new__(Rotation)
    # Someone could have registered a different backend for jax, so we attempt to fetch the
    # updated backend here. If not, we fall back to the Array API backend.
    xp = array_namespace(c[0])
    r._xp = xp
    r._backend = backend_registry.get(xp, array_api_backend)
    r._quat = c[0]
    # We set _single to False for jax because the Array API backend supports broadcasting by
    # default and hence returns the correct shape without the _single workaround
    r._single = False
    return r

register_pytree_node(Rotation, lambda v: ((v._quat,), None), rot_unflatten)
```

ev-br:
> but I can push that iteration later today if you want. 

Yes please!

Two questions on the Rotation clarifications:

1. The benchmarking script is jax-only, correct? More generally, have you tried / looked into / thought about seeding how generalize the Rotation work to `torch.compile` ? 
2. What is the suggested user story for the registration code? Surely we don't expect users to be able to come up with `rot_unflatten` above. Do we include this into scipy, do we spell this as a recipe somewhere in the docs, do we add some helpers for this in scipy or jax itself?

amacati:
> Two questions on the Rotation clarifications:
> 
> 1. The benchmarking script is jax-only, correct? More generally, have you tried / looked into / thought about seeding how generalize the Rotation work to `torch.compile` ?

No, the code works with jax, torch, numpy and cupy at the moment. Only jax jits it though. See https://github.com/amacati/scipy/blob/1d5e0ec9e82cc246e9961647d1bba4ba9796226f/xp_bench/rot_benchmark.py#L885

> 2. What is the suggested user story for the registration code? Surely we don't expect users to be able to come up with `rot_unflatten` above. Do we include this into scipy, do we spell this as a recipe somewhere in the docs, do we add some helpers for this in scipy or jax itself?

This was shortly debated in the generic xp `Rotation` PR. My view was that it would be great to register `Rotation`s because it would allow them to act as return values from jitted functions, enables auto-broadcasting via `vmap` etc. Initially I thought we would include it here in scipy, but @crusaderky pointed out that there are several issues with that. 
<details>
We don't want to import jax because it leads to unexpected behavior. Some jax settings have to be set as environment variables before the first import. Users importing scipy, setting those variables and finally importing jax would not see the changes applied. In addition, importing jax is expensive and preallocates GPU memory by default on the first operation.
</details>

My next preferred option would be to include the optimized registration in jax. Since the code is using internals of `Rotation`, we might provide a `jax_register_rotations` function for them if they are not comfortable with having code that relies on these. Jax would then call this function on init to register `Rotation` as a `PyTree`, and all would work as expected. Scipy is a dependency in jax anyways, so that shouldn't be a large issue.

The other way of using `Rotation` in jax with jit compilation is to create `Rotation`s inside a function and use it as part of the computations, but never return the object itself. This is what is possible with the current state of things. Jitting such a function does not require `Rotation` to be a `PyTree`. It only requires functions from `Rotation` to be jit compatible (which we already test). For example, one could write this:

```python
import jax


# Possible without registration
@jax.jit
def rotate_points_by_degrees(x: Array, degrees: Array)  -> Array:
    return R.from_euler("z", degrees, degrees=True).apply(x)


# Only possible with registration
@jax.jit
def rotate_points(x: Array, rot: R) -> Array:
    return rot.apply(x)


@jax.jit
def rotate_rotations(r1: R, r2: R) -> R:
    return r1 * r2
```

ev-br:
Agreed, making all scipy users pay the price of importing jax just for those who want to jit compile a class or two is not desirable.

Making users copy-paste an estoteric piece of code, which relies of private implementation details, https://github.com/scipy/scipy/pull/23447#issuecomment-3167316573, is also very much not desirable.

Forcing users of a class to always use the object only once is not a great solution either (why on earth do you have a class then). 

Adding things to jax itself could be a way to go. This would mean somewhat tighter coupling between SciPy (a consumer) with JAX (a producer) than desired though. Are JAX devs aware of the issue, are they open to adding code for Rotation --- and a bunch of other things from SciPy in the future, is there an upstream issue or a PR?

Unless the answers from JAX are yes yes and yes, the only reasonable thing I can think of is cook up a SciPy registration API. Something along the lines of

```
>>> scipy.register_jax_jit_func(scipy.spatial.Rotation)
```

which under the hood does basically  `rot_unflatten` from https://github.com/scipy/scipy/pull/23447#issuecomment-3167316573 + conditionally importing JAX  etc.

The only other alternative I can see is what this PR does: we jit things ourselves internally. That's not great from the maintenance perspective either.

amacati:
The discussion over in jax was what lead to the array API push in `Rotation` in the first place: https://github.com/jax-ml/jax/issues/25491 
The jax team wants to drop support for `Rotation` long-term ([see here](https://docs.jax.dev/en/latest/jep/18137-numpy-scipy-scope.html#scipy-spatial)), but I hope they would be open to include the registration code instead since that's minimal effort on their side. However, the current implementation does not match the features of the jax implementation because it cannot broadcast to arbitrary dimensions. Changing this should be feasible (https://github.com/scipy/scipy/issues/23437) but will take some time. Once that's done and we match (or rather exceed) jax's feature I was planning to approach the people from jax and open up a PR to integrate this.

This would make deprecation of `jax.spatial.transform.Rotation` easy for them as well because they can just point to using `scipy.spatial.transform.Rotation` and everything should work exactly as before.

Maybe @jakevdp could quickly comment on this to see if that's realistic?

purepani:
@amacati maybe it's worth making a separate issue for this, since it seems like a decently generic issue.

amacati:
> @amacati maybe it's worth making a separate issue for this, since it seems like a decently generic issue.

You mean registering scipy classes in jax? Maybe, though `Rotation` would be the first to do so as far as I know (_if_ that happens). And it's probably best to ask the jax people first if they would even support such an effort

Also, this PR should not be hijacked by `Rotation` discussions. We should take it to the progress tracker #23391 .

amacati:
> Yes please!

Sorry @ev-br , the update took a bit longer than expected. The benchmark branch is updated now. All benchmarks live in the xp_bench folder. The benchmarks can be run for individual frameworks and/or functions. All results get saved to disk. _plot_comparison.py_ then generates the plots, and _create_report.py_ puts everything into a single report pdf. 

Runs may need manual renaming from you to make the comparison work. See e.g. the labeling where I renamed previous results to _xxx_no_opt_ 
```python
    label = f"{device} " + ("optimized" if "_no_opt" not in xp else "")
```
These were just for PR comparisons, so I didn't bother polishing everything.

ev-br:
Updated benchmarks: https://github.com/ev-br/bench_playground/blob/master/viz.rbf.ipynb
(the benchmarking code is the same repository, relies on `$ pip install pytest-benchmark`)

Several observations:
- Potential performance gains from JIT are huge, up to 30x or more
- Performance characteristics strongly depend on the hardware, esp on the CPU count.
- One can make numpy/pythran twice faster by just asking OpenBLAS to *not* multithread :-). 
- Controlling the threading state across backends is surprisingly brittle, especially in the JIT context (e.g. https://github.com/pytorch/pytorch/issues/160812#issuecomment-3198919245)
- I'm getting convinced that we should not JIT anything ourselves, and leave it to the user. For class-based APIs this might mean that users need to monkeypatch methods with their jitted versions. So the SciPy message should be roughly "you may want to experiment with jitting specific methods, and you can potentially unlock huge perf gains, but we do not officially support it; if you find dragons, we sympathize but ultimately it's up to you".
- `pytest-benchmarks` is nice, and can be made to improve on `asv` pain points with some work.

lucascolley:
> One can make numpy/pythran twice faster by just asking OpenBLAS to not multithread :-).

https://pythonspeed.com/articles/concurrency-control/, right?

rgommers:
> I'm getting convinced that we should not JIT anything ourselves, and leave it to the user. 

That sounds right, that's how the `torch.compile`/`jax.jit`/`numba.jit` functionality is meant to be used. And you don't want to pre-JIT on import, that's very expensive.

> For class-based APIs this might mean that users need to monkeypatch methods with their jitted versions.

That's far less clear, I don't think that can be the right answer. Seems related to https://github.com/scipy/scipy/issues/23391#issuecomment-3172634371.

ev-br:
@lucascolley 
> https://pythonspeed.com/articles/concurrency-control/, right?

This post IIUC talks about oversubscription due to multiple thread pools. While this is very much a valid concern, here we're seeing something dfferent: a single python thread, no multiprocessing or python treads. The only threadpool is that internal to OpenBLAS. And running with `OPENBLAS_NUM_THREADS=1` is twice faster than running with whatever the default OpenBLAS thread pool size is ( num_cpus or 2*num_cpus due to hyperthreading).

This is likely _very_ problem-specific and openblas version specific and is possibly due to how pythran interacts with openblas (No idea, and I did not check pythran-generated code).
This also might be related to the fact that here we're doing *a lot of* calls to `np.linalg.vector_norm` with length-2 arrays. I wouldn't be surprised if this is not what OpenBLAS defaults are tuned towards.

That post is very valid regardless, and is relevant to what we're seeing here on other backends. For instance, with `jax/jit` running single-threaded is 5x slower than the default on 32 cores. Which means if you can split the work into 32 chunks and limit each chunk to a single core, you get a 32/5 = 6.4 speedup. 

So for a user recommendation, it's always a good idea to check how the application reacts to `OPENBLAS_NUM_THREADS` or `MKL_NUM_THREADS`; you might be pleasantly surprised :-).

@rgommers
>> .. we should not JIT anything ourselves, and leave it to the user.

> That sounds right, that's how the torch.compile/jax.jit/numba.jit functionality is meant to be used. And you don't want to pre-JIT on import, that's very expensive.

Absolutely yes for functional APIs, esp pure functions. When you want to store a state, things start geting muddy.

>> For class-based APIs this might mean that users need to monkeypatch methods with their jitted versions.

> That's far less clear, I don't think that can be the right answer. Seems related to #23391 (comment).

Sorta. It's not very clear-cut though. Let's consider specifics. Suppose we have a class, which has a couple of methods. Jitting all methods does not necessarily make sense (depends on details). Hooking a whole class to JIT machinery may look like this: https://github.com/scipy/scipy/pull/23447#issuecomment-3167316573 --- note that this is also jax-specific, and who knows what it takes to make it talk to torch Dynamo.
So, we can't add this to SciPy, we can only tell the user to copy-paste this.

Alternatively, consider jitting a single method. The options are:

1. what this PR does. Hardcode jit technologies (torch.dynamo, jax.jit), add jit-specific "backends"; where each backend is like three LOC, https://github.com/scipy/scipy/pull/23447/files#diff-38f028899642eff3eaef89422219703e5053a3c6225a5a32ca3c05350d2a731b

2. attach the things to jit as methods and ask the user to do 

```
from scipy.interpolate import RBFInterpolator
rbf = RBFInterpolator(....)

# fiddling with private methods!
rbf._compute_interpolation = jax.jit(rbf._compute_interpolation, static_argnames=["kernel", "xp"])

# or

rbf._compute_interpolation = torch.compile(fullgraph=True, dynamic=True)(rbf._compute_interpolation)
```

EDITED: This is "Strategy 2" of https://docs.jax.dev/en/latest/faq.html#how-to-use-jit-with-methods. Reading that link, I think we should stick to only jitting helper functions ("Strategy 1").

3. invent a scipy-specific way to paper over the monkey-patching above (which is what https://github.com/scipy/scipy/issues/23391#issuecomment-3172634371 suggests)


Choosing betwen 2 or 3, I'm not 100% sure. The knee-jerk reaction is we don't support mucking with private methods.
OTOH if we go with 3, it'll be scipy-specific and not very discoverable.

purepani:
Don't have time at the moment to test this(and I may also be being dumb), but can't we make it so that the user just compile `rbf` in that example?
```python
from scipy.interpolate import RBFInterpolator
rbf = RBFInterpolator(....)

rbf = jax.jit(rbf, ...)
```

Also, you definitely don't want to pre-jit compile these functions, and I would argue option 2 isn't super great.
I believe certain other features can break if you don't do the jit at the top level for jax(like you can't vmap over a precompiled jit function).

ev-br:
>  can't we make it so that the user just compile rbf in that example?

This is "Strategy 3" in https://docs.jax.dev/en/latest/faq.html#how-to-use-jit-with-methods, and is what https://github.com/scipy/scipy/pull/23447#issuecomment-3167316573 suggests.

AFAICS to make it work, we need to find a way to 1) make it so that users who do not jax jit do not pay for importing jax, 2) make it work across multiple jit technologies (at least torch.dynamo and jax.jit).

Re-reading the jax faq entry above, I see that my option 2 is Jax's "Strategy 2" and is fraught with subtle difficulties, so probably best not go there.

Which leaves us with Jax's "Strategy 1", which is what this PR does currently: only jit compile helper functions. Then the choice for us IMO is between leaving to users to jit private helpers, and inventing a scipy-specific public API to hide these private helpers.

rgommers:
> Which leaves us with Jax's "Strategy 1", which is what this PR does currently: only jit compile helper functions. Then the choice for us IMO is between leaving to users to jit private helpers, and inventing a scipy-specific public API to hide these private helpers.

The former is not an acceptable option - really, no monkeypatching or other hacks and no private API that needs to be documented or touched. The latter does seem like an option; at that point it looks like it'd be similar to having that same API register a pytree (strategy 3).

> The knee-jerk reaction is we don't support mucking with private methods.

Yep, that. 

> OTOH if we go with 3, it'll be scipy-specific and not very discoverable.

There is no way around having the user call at least some one-liner, so I would not worry about discoverability here, that is what it is. 

* * *

Looking at the discussion in https://github.com/scipy/scipy/issues/23391#issuecomment-3172671518 and at how different `torch.compile` approaches look (pytrees are still under discussion, see e.g. https://github.com/pytorch/pytorch/issues/65761, but ) as well as Numba's`@jitclass`, this is quite tricky to design as a common API either way.

I think there are a couple of other design principles that are hard constraints here:

1. No pre-compilation (agreeing with what @purepani said above). It's the user who should be calling `jax.jit`/`torch.compile`, not `scipy`. We need jit-compatible code, but not actually do the jitting.
2. No `if USE_JIT: ... else: ...` (or `if detect_we_are_jitting: ... else:`). That's what this PR relies on, but we established as part of the work that adding JAX support that it's a no-go for JAX.

The key choice seems to be between (a) doing some kind of registration or (b) constraining the user code to directly JIT a public method and use that. The former can look something like:

```python
# possibly do per module rather than per class, or even for all modules at once
import jax
scipy.register_jit(jax, scipy.interpolate.RBFInterpolator)

@jax.jit
def some_user_code(...):
    # Can now use SciPy API as normal, best composability
    ...
    rbf = RBFInterpolator(....)
    y_interp = rbf(x_interp)
    ...
```

It's unclear to me though if we can have an API like that that is a good fit for at least `torch.compile` as well.

amacati:
Just as a meta comment: I think what all jit approaches rely on is a form of pure data container that can be passed and verified to compiled functions. This is what makes it so easy to register `Rotation` and `RigidTransform` as `PyTree`s, they revolve around a single quaternion array / a matrix. 

If possible, new OOP APIs should be designed with that in mind: Have one data container and then operate on that with pure functions. I have not looked too much into torch.compile, but my guess is it would work best with pure functional APIs on tensors as well.

> ```python
> # possibly do per module rather than per class, or even for all modules at once
> import jax
> scipy.register_jit(jax, scipy.interpolate.RBFInterpolator)
> 
> @jax.jit
> def some_user_code(...):
>     # Can now use SciPy API as normal, best composability
>     ...
>     rbf = RBFInterpolator(....)
>     y_interp = rbf(x_interp)
> ```

If the constructor of the interpolator is jit compatible, this should be possible without registering anything. JAX will trace the construction of the object and the use of its functions and compile all of that. We just run into problems if we want to pass an interpolator out of a jit context or give the function an interpolator as argument. Assuming of course that the internals of the class are non-branching etc.

ev-br:
> It's unclear to me though if we can have an API like that that is a good fit for at least torch.compile as well.

That's exactly the key problem here. Your design just hides a pytree or jitclass registration and requires that the jit decorator understands the class. Which I do not believe is the case (am happy to be proven wrong though).

So the problem I think is this: we want the _jit callable_ to live in the userland, and apply it to a private detail. So if we make the user spell out the callable itself, we can plumb it through roughly like

```
import jax
scipy.apply_jit(jax.jit, scipy.interpolate.RBFInterpolator)    # jax.jit, a callable

# or

import torch
scipy.apply_jit(torch.compile, scipy.interpolate.RBFInterpolator)     # torch.compile, a callable

# and then

# no @jax.jit decorator
def some_user_code(...):
        ...
    rbf = RBFInterpolator(....)
    y_interp = rbf(x_interp)
```

Internally, `scipiy.apply_jit` maintains a mapping from `RBFInterpolator` to `scipy.interpolate._rbfinterp_xp.compute_interpolation`, and applies to it the JIT callable it receives as the first argument:

```
def apply_jit(callable, Klass):
   if Klass == RBFInterpolator:
        import scipy.interpolate._rbfinterp_xp as _module
        _module.compute_interpolation = callable(_module.compute_interpolation)
   elif Klass == Rotation:
        .... apply the `callable` to `_from_davenport` or something
```

Not entirely sure whether this sort of contraption is better than making the user just do

```
        import scipy.interpolate._rbfinterp_xp as _module
        _module.compute_interpolation = callable(_module.compute_interpolation)
```


> Assuming of course that the internals of the class are non-branching etc.

That's not an assumption we can make across SciPy as it currently is, for better or worse.

rgommers:
@amacati I'm not sure I got your points. This is true for every library I believe:

> my guess is it would work best with pure functional APIs on tensors as well.

however when we do have a class-based API then (almost?) always it's because we are grouping a mutable data container with a set of methods specific to that container. If your point is that those methods themselves shouldn't mix mutating self with doing computation, then yes I agree. And that's usually the case, like here for `RBFInterpolator`.

> [...] We just run into problems if we want to pass an interpolator out of a jit context or give the function an interpolator as argument.

Do you mean "method" instead of "function" here? Or do you mean something like this:
```python
rbf = RBFInterpolator(....)

@jax.jit
def some_user_code(rbf, ...):
    ...
    y_interp = rbf(x_interp)
```

rgommers:
> Your design just hides a pytree or jitclass registration 

Yes. It seems _something_ is necessary, and it's just about how to make that public and ergonomic. All I said was essentially "case 3 in the JAX FAQ is the most generic, and it's feasible to make that public and ergonomic".

> and requires that the jit decorator understands the class. Which I do not believe is the case (am happy to be proven wrong though).

It isn't the case for `RBFInterpolator` specifically, or for arbitrarily-complex classes in general?

amacati:
> however when we do have a class-based API then (almost?) always it's because we are grouping a mutable data container with a set of methods specific to that container. If your point is that those methods themselves shouldn't mix mutating self with doing computation, then yes I agree. And that's usually the case, like here for `RBFInterpolator`.

Yes, that was the point. I haven't looked at `RBFInterpolator`, but the ideal case is that there is a very simple data structure, preferably a single tensor, that contains all data. All methods would then look approximately like this:

```python
class RBFInterpolator:
    data: Array
    ...
    def some_fn(self, t: Array):
        self.data = functional_some_fn(self.data, t)
```
Depending on the class structure that might not be possible, but _if_ things are written in this way it could simplify the integration of jit compilation, whatever route scipy takes. This is not meant to solve existing problems, but we should think about this when introducing new classes / APIs. Or stay functional from the beginning, with only primitive types allowed as inputs wherever possible.

> 
> > [...] We just run into problems if we want to pass an interpolator out of a jit context or give the function an interpolator as argument.
> 
> Do you mean "method" instead of "function" here? Or do you mean something like this:
> 
> ```python
> rbf = RBFInterpolator(....)
> 
> @jax.jit
> def some_user_code(rbf, ...):
>     ...
>     y_interp = rbf(x_interp)
> ```
I was referring to the latter. I really see three use-cases: 

1. A user wants to interpolate inside a jit'ed function. They create the interpolator within jit and only pass out results. In that case, it is sufficient if `__init__` and any other functions are called are non-branching and whatever additional constraints the compiler has.
2. A user wants to create an interpolator and pass it into a jit'ed function, or return an interpolator. In that case, the class must be registered (at least for jax, I think something similar might exist for torch.compile).
3. We want to jit methods of the class (like `__call__`). In that case I agree that an API like `scipy.apply_jit(jax.jit, scipy.interpolate.RBFInterpolator)` would be necessary. However, we should think about how often this is the case. It assumes scipy is used at the highest level of a user's logic. I think it's much more likely that users use the library _inside_ a compiled context. For comparison:

```python
# The API is used directly
rbf = RBFInterpolator(....)
y_interp = rbf(x_interp)
some_other_fn(y_interp)

# This should be much more likely for a lot of the APIs

@jax.jit  # Perfectly fine, this jits scipy with it
def calculate_something(x: Array) -> Array:
    rbf = RBFInterpolator(....)
    y_interp = rbf(x_interp)
    return some_other_fn(y_interp)
```
> 
> > Assuming of course that the internals of the class are non-branching etc.
> 
> That's not an assumption we can make across SciPy as it currently is, for better or worse.

Oh yes, absolutely. This was meant as a comment on what to think about when creating new APIs or modifying existing ones with minimal changes. It does not solve the issues with existing APIs. For what it's worth, I think attempting to convert old APIs with multiple blockers for jit compilation will lead to inconsistent/surprising behaviours and suboptimal performance. It might be better to include a "jit ready" flag in the docs instead, focus on writing new APIs to be compatible, and replace old ones whenever people are interested in adding support.

lucascolley:
> It might be better to include a "jit ready" flag in the docs instead

Indeed we have https://scipy.github.io/devdocs/dev/api-dev/array_api.html#support-with-jit

purepani:
Ah, I see. This is just another pytree problem.

The best solution imo is some form of a pytree protocol, where you, in an ideal world, could define `RBFInterpolator().__tree_unflatten__` and `RBFInterpolator().__tree_flatten__`, and every array library that has some concept of structured pytrees could implement support for those, and know what to do with them, without needing to manually register them(https://github.com/jax-ml/jax/issues/8099).

That being said, doing something like this takes a good amount of coordination and effort. Certainly a reasonable short-term solution is https://github.com/scipy/scipy/pull/23447#issuecomment-3227645217. 

> So the problem I think is this: we want the jit callable to live in the userland, and apply it to a private detail.

I feel like I disagree with this. Having the user not directly use the jit directly, and using it through a scipy helper feels like a very unintuitive solution to me, and can't compose with other features.

rgommers:
> 3\. It assumes scipy is used at the highest level of a user's logic. I think it's much more likely that users use the library _inside_ a compiled context.

Agreed, that's an important point. 

> The best solution imo is some form of a pytree protocol,

That sounds right to me - but it doesn't exist yet. Punting on this or having an API that's explicitly unstable/experimental because we reserve the right to migrate to such a protocol in the future could be a way to go. 

> > So the problem I think is this: we want the jit callable to live in the userland, and apply it to a private detail.
> 
> I feel like I disagree with this. Having the user not directly use the jit directly, and using it through a scipy helper feels like a very unintuitive solution to me, and can't compose with other features.

I agree I think, it's ~unintuitive~ less intuitive. It's better to enable `jax.jit`/`torch.compile` to work than to replace it with a scipy-specific equivalent. I also don't agree with "apply it to a private detail" - the goal is to apply it to SciPy's public API, or to even higher-level code that internally uses such public API.

purepani:
I will note that for JAX, there is already a supported way [to define a tree_flatten and tree_unflatten methods](https://docs.jax.dev/en/latest/pytrees.html).

```python
from jax.tree_util import register_pytree_node_class

@register_pytree_node_class
class RegisteredSpecial2(Special):
  def __repr__(self):
    return "RegisteredSpecial2(x={}, y={})".format(self.x, self.y)

  def tree_flatten(self):
    children = (self.x, self.y)
    aux_data = None
    return (children, aux_data)

  @classmethod
  def tree_unflatten(cls, aux_data, children):
    return cls(*children)

show_example(RegisteredSpecial2(1., 2.))
```

This obviously still requires importing jax, but it might work as a starting point to build something that doesn't require that.

Additionally there's this generic pytree library https://github.com/metaopt/optree

amacati:
> I will note that for JAX, there is already a supported way [to define a tree_flatten and tree_unflatten methods](https://docs.jax.dev/en/latest/pytrees.html).

That was the thought for `Rotation` at least. We are a jax dependency anyways, so if we offer a `register_scipy_pytrees()` function that jax or whoever needs to call on init that would be fine imo. The question is if they would be open to something like that, but the jax team has yet to respond to that proposition (see e.g. https://github.com/jax-ml/jax/pull/30826#issuecomment-3191717054).

rgommers:
> This obviously still requires importing jax, but it might work as a starting point to build something that doesn't require that.

I think this is inherently infeasible in the absence of something like the PyTree Protocol idea that you linked above. Within `scipy`, we cannot try to do anything before a JAX array is passed into the class - which is too late.

In addition, it seems like the `tree_flatten`/`tree_unflatten` currently need to be public methods - that's not okay, this has to be customizeable so it doesn't leak into the public API of a `scipy` class.

ev-br:
Two potentially relevant design ideas (no specific suggestions yet, just recording the links here):

- `spatch` considers adding a notion of "jit backend", https://github.com/scientific-python/spatch/issues/32#issuecomment-3258551658

- A somewhat recent blog post, https://blog.scientific-python.org/pytrees/, discusses PyTrees beyond jax. The example with `minimize` they discuss is not immediately related, but maybe it's something worth exploring.

ev-br:
https://github.com/scipy/scipy/pull/23746 separated the "eager" part, so the plan is to rebase this PR to only contain the JIT changes and continue discussing the JIT strategies here, after gh-23746 lands.

ev-br:
Rebased on main, now that gh-23746 has landed.

lucascolley:
> For the JIT mode, we manually override select functions of the "generic" backend by their jit-compiled versions. For example, eager pytorch execution uses the generic _rbfinterp_xp.py module, and JIT-compiled variants reside in the _rbfinterpr_dynamo.py which overrides parts of the "generic" backend module. Likewise, eager execution with JAX arrays uses the "generic" backend, and the JIT-compiled overrides live in the _rbfinterp_jaxjit.py module.

This is the kind of thing I think should belong in an optional spatch backend.

rgommers:
> This is the kind of thing I think should belong in an optional spatch backend.

That seems very much undesirable, an out-of-tree package to support jitting will have higher overhead and be far less reliable and usable. In case `spatch` provides useful plumbing here, it can be used inside this repo.

PR #23453, opened 2025-08-09
MAINT: Support ruff 0.12

DimitriPapadopoulos:

In ruff [0.12.0](https://github.com/astral-sh/ruff/releases/tag/0.12.0):
> * [non-pep604-annotation-union](https://docs.astral.sh/ruff/rules/non-pep604-annotation-union) (`UP007`) has now been split into two rules. `UP007` now applies only to `typing.Union`, while [non-pep604-annotation-optional](https://docs.astral.sh/ruff/rules/non-pep604-annotation-optional) (`UP045`) checks for use of `typing.Optional`. `UP045` has also been stabilized in this release, but you may need to update existing `include`, `ignore`, or `noqa` settings to accommodate this change.

#### Additional information

While we could ignore both rules, I prefer requiring a decently recent version of ruff.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -58,5 +58,5 @@ dependencies:
   # For CLI
   - spin
   # For linting
-  - ruff>=0.0.292
+  - ruff>=0.12.0
   - cython-lint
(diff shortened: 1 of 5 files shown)


lucascolley (approved):
thanks @DimitriPapadopoulos !

PR #23459, opened 2025-08-13
BLD: allow qhull-8.1

negril:

There is no release (yet), but the alpha versions fix a bunch of cmake issues.

#### Additional information

https://github.com/qhull/qhull/compare/v8.0.2..master

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -214,14 +214,14 @@ else
 endif
 
 if all_system_libraries or use_system_libraries.contains('qhull')
-  qhull_r_dep = dependency('qhull_r', version : '8.0.2')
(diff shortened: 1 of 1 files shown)


lucascolley:
We only support a single version of QHull at a time, since by default we use our vendored copy. So I'll mark this as draft for now until there is a release.

Note that this PR should also update the vendored source at https://github.com/scipy/scipy/tree/main/subprojects/qhull_r.

lucascolley:
closing due to inactivity, a PR is still welcome to perform the bump once v8.1 is released and available via conda-forge

PR #23465, opened 2025-08-15
BUG: sparse: nD sparse save_npz and load_npz

### LINKED ISSUE #23462: BUG: sparse: load_npz fails for coo_arry with dim>2
### Describe your issue.

Saving a `coo_array` with more than 2 dimensions with `scipy.sparse.save_npz` and trying to load it with `load_npz` fails with a `ValueError`:

The loading code seems hardcoded to handle 2D only: https://github.com/scipy/scipy/blob/0f2342f34cc850e7137a74afd33c6877c1c790e6/scipy/sparse/_matrix_io.py#L162

### Reproducing Code Example

```python
import numpy as np
import scipy as sp

arr = np.zeros((10, 10, 10))
coo = sp.sparse.coo_array(arr)

sp.sparse.save_npz('test.npz', coo)
sp.sparse.load_npz('test.npz')
```

### Error message

```shell
203 """ Checks data structure for consistency """
    204 if self.ndim != len(self.coords):
--> 205     raise ValueError('mismatching number of index arrays for shape; '
    206                      f'got {len(self.coords)}, expected {self.ndim}')
    208 # index arrays should have integer data types
    209 for i, idx in enumerate(self.coords):

ValueError: mismatching number of index arrays for shape; got 2, expected 3
```

```shell
1.15.3 2.2.6 sys.version_info(major=3, minor=10, micro=12, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    detection method: pkgconfig
    found: true
    include directory: /opt/_internal/cpython-3.10.15/lib/python3.10/site-packages/scipy_openblas32/include
    lib directory: /opt/_internal/cpython-3.10.15/lib/python3.10/site-packages/scipy_openblas32/lib
    name: scipy-openblas
    pc file directory: /project
    version: 0.3.28
  lapack:
    detection method: pkgconfig
    found: true
    include directory: /opt/_internal/cpython-3.10.15/lib/python3.10/site-packages/scipy_openblas32/include
    lib directory: /opt/_internal/cpython-3.10.15/lib/python3.10/site-packages/scipy_openblas32/lib
    name: scipy-openblas
    pc file directory: /project
    version: 0.3.28
  pybind11:
    detection method: config-tool
    include directory: unknown
    name: pybind11
    version: 2.13.6
Compilers:
  c:
    commands: cc
    linker: ld.bfd
    name: gcc
    version: 10.2.1
  c++:
    commands: c++
    linker: ld.bfd
    name: gcc
    version: 10.2.1
  cython:
    commands: cython
    linker: cython
    name: cython
    version: 3.0.12
  fortran:
    commands: gfortran
    linker: ld.bfd
    name: gcc
    version: 10.2.1
  pythran:
    include directory: ../../tmp/pip-build-env-jujiy7q_/overlay/lib/python3.10/site-packages/pythran
    version: 0.17.0
Machine Information:
  build:
    cpu: x86_64
    endian: little
    family: x86_64
    system: linux
  cross-compiled: false
  host:
    cpu: x86_64
    endian: little
    family: x86_64
    system: linux
Python Information:
  version: '3.10'
```

dschult:
Fixes #23462 

`save_npz` writes a `coo_array` as if it has ndim == 2 (whether 1D or nD for n>2).
`load_npz` for coo_array always creates 2D array.

This PR allows COO sparse format to be saved and loaded as nD.

I'm not sure what the backward compatibility tradition is for `save_npz` and `load_npz`.
For `save_npz` this PR saves `A.coords` as well as `A.row` and `A.col`, though all that is needed is to save `coords`.  This makes all stored files bigger than needed, but they can be read by older SciPy versions of load_npz.
For `load_npz` this PR checks the loaded arrays for the name `coords`, and use that if it is present. Otherwise it loads the older version arrays `row` and `col`. This allows us to load files that were saved using older SciPy versions.

I can change it to have `save_npz` store smaller files (with only `coords`) so new versions of SciPy can read older versions of npz files but older versions of SciPy cannot read newer saved npz files.  Or maybe its better to still use row/col for 2D, and coords for 1D and nD.

### DIFF
--- a/scipy/sparse/tests/test_matrix_io.py
+++ b/scipy/sparse/tests/test_matrix_io.py
@@ -2,6 +2,7 @@ import os
 import numpy as np
 import tempfile
 
+import pytest
 from pytest import raises as assert_raises
 from numpy.testing import assert_equal, assert_
 
@@ -72,8 +73,10 @@ def test_sparray_vs_spmatrix():
     assert_(loaded_matrix.dtype == loaded_array.dtype)
     assert_equal(loaded_matrix.toarray(), loaded_array.toarray())
 
-def test_nd_coo_format(tmpdir):
-    A = coo_array([[[0]]])
+@pytest.mark.parametrize("value", [0, 1.2])
+@pytest.mark.parametrize("ndim", [1, 2, 3])
+def test_nd_coo_format(ndim, value, tmpdir):
+    A = coo_array([value]).reshape((1,) * ndim)
 
     #save/load array
     with tmpdir.as_cwd():
--- a/scipy/sparse/tests/test_matrix_io.py
+++ b/scipy/sparse/tests/test_matrix_io.py
@@ -72,17 +72,15 @@ def test_sparray_vs_spmatrix():
     assert_(loaded_matrix.dtype == loaded_array.dtype)
     assert_equal(loaded_matrix.toarray(), loaded_array.toarray())
 
-def test_nd_coo_format():
+def test_nd_coo_format(tmpdir):
     A = coo_array([[[0]]])
 
     #save/load array
-    fd, tmpfile = tempfile.mkstemp(suffix='.npz')
-    os.close(fd)
-    try:
+    with tmpdir.as_cwd():
+        tmpfile = "f.npz"
+
         save_npz(tmpfile, A)
         loaded_A = load_npz(tmpfile)
-    finally:
-        os.remove(tmpfile)
 
     assert isinstance(loaded_A, coo_array)
     assert_(loaded_A.shape == A.shape)
--- a/scipy/sparse/tests/test_matrix_io.py
+++ b/scipy/sparse/tests/test_matrix_io.py
@@ -5,8 +5,9 @@ import tempfile
 from pytest import raises as assert_raises
 from numpy.testing import assert_equal, assert_
 
-from scipy.sparse import (sparray, csc_matrix, csr_matrix, bsr_matrix, dia_matrix,
-                          coo_matrix, dok_matrix, csr_array, save_npz, load_npz)
+from scipy.sparse import (sparray, csr_array, coo_array, save_npz, load_npz,
+                          csc_matrix, csr_matrix, bsr_matrix, dia_matrix,
+                          coo_matrix, dok_matrix)
 
 
 DATA_DIR = os.path.join(os.path.dirname(__file__), 'data')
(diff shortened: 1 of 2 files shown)


scipy/sparse/tests/test_matrix_io.py
```diff
@@ -71,6 +72,22 @@ def test_sparray_vs_spmatrix():
...
+    A = coo_array([[[0]]])
+
+    #save/load array
+    fd, tmpfile = tempfile.mkstemp(suffix='.npz')
+    os.close(fd)
+    try:
+        save_npz(tmpfile, A)
+        loaded_A = load_npz(tmpfile)
+    finally:
+        os.remove(tmpfile)
```
tylerjereddy: Maybe we could cut down on the boilerplate for file handling here -- the diff below seems to work locally. I did also confirm that the test fails fails before/passes after the patch while I was here.

```diff
-def test_nd_coo_format():
+def test_nd_coo_format(tmpdir):
     A = coo_array([[[0]]])
 
     #save/load array
-    fd, tmpfile = tempfile.mkstemp(suffix='.npz')
-    os.close(fd)
-    try:
+    with tmpdir.as_cwd():
+        tmpfile = "f.npz"
         save_npz(tmpfile, A)
         loaded_A = load_npz(tmpfile)
-    finally:
-        os.remove(tmpfile)
```

dschult: simplifying tmpdir fixture is now in place.  Thanks!
CI failure seems to be unrelated.

perimosocordiae:
How about this plan: for now, we keep writing row + col for the 2d case, but start writing coords for any other dimension.

The reading logic would use coords if available, then fall back to row + col.

After a few releases, we can switch the 2d case to save with coords only.

dschult:
OK. I've update the save_npz to save all 2D sparse matrices/arrays using `row` and `col`, with anything else saved as `coords`. I added a Todo comment about switching to use only coords after a few releases.

scipy/sparse/tests/test_matrix_io.py
```diff
@@ -71,6 +72,20 @@ def test_sparray_vs_spmatrix():
     assert_(loaded_matrix.dtype == loaded_array.dtype)
     assert_equal(loaded_matrix.toarray(), loaded_array.toarray())
 
+def test_nd_coo_format(tmpdir):
```
perimosocordiae: Can we parametrize this for 1,2,3 dimensions?

perimosocordiae (approved):

dschult:
I parametrized for 1,2,3-D and also for the value in the array being 0 or 1.2

perimosocordiae (approved):

perimosocordiae:
Thanks, merged!

PR #23467, opened 2025-08-15
ENH: spatial.transform.Rotation: support broadcasting

amacati:
Add support for arbitrary leading batch dimensions for `Rotation`.

# Motivation
Previously, `Rotation` only accepted either single rotations or an array of rotations. This limitation stems from the original Cython implementation that did not use broadcasting. With the newly introduced generic backends it is possible to remove this restriction and apply regular broadcasting logic (see https://github.com/scipy/scipy/pull/23249).

This enables users to create arbitrarily shaped rotations, i.e. rotate a batch of (5, 3, 2) rotations around 90 degrees, apply a batch of rotations with shape (3, 1, 6, (4, )) to a batch of 3D points (1, 8, 6, (3, )) and so on.

Note that this PR **does not introduce any breaking changes**. Any inconsistencies with these rules have been kept and clearly marked in the code. If we decide to make all results consistent with broadcasting logics we can simply remove these special cases.

# Implementation
There are three main concerns of this PR: 

1. How do we determine how differently shaped tensors of rotations and possibly vectors should be combined?
2. How can we make this work with the existing Cython backend?
3. How do we keep the maintenance burden low?

## Combining shapes of arbitrary dimensions
All operations follow the broadcasting rules outlined in https://numpy.org/doc/stable/user/basics.broadcasting.html. E.g., the result of rotating a batch (1, 8, 6, (3, )) 3D points by a Rotation of (3, 1, 6, (4, )) using `Rotation.apply` yields a batch of (3, 8, 6, (3, )) 3D points.

## Overcoming Cython's shape limitation
We currently use the Cython backend for single rotations and lists of rotations for numpy arrays. We cannot open this up to arbitrary shapes because of Cython type restrictions. The arguments to the Cython backend **must** be 2D to enable statically typed memory views into the arrays.

Converting all arguments to 2D and reshaping them back after calling into the Cython backend also does not work without duplicating data for all axes that broadcast, which would add considerable complexity. Instead, we assume that users interested in operations involving tensors of `Rotation`s are working with significant data sizes and prefer operations that natively broadcast anyways. Hence, we simply call the array API backend in those cases. One added benefit is that this requires 0 changes to the current Cython code.

### Why not remove the Cython backend?
The Cython backend provides speedups for some of the most common cases (single `Rotation` or smaller lists of `Rotation`s). Especially for operations on smaller arrays, removing the backend would result in a performance hit for many users.

## Managing maintenance burden
The array API backend has been designed to use broadcasting rules from the beginning (_huh, what a coincidence :D_).  Most backend functions have seen little or 0 changes. The main change comes from delegating to the generic backend in case the input shapes of numpy arrays are incompatible with Cython and from not throwing an error when creating tensors of `Rotation`s. The two main exceptions are `from_euler` and `from_davenport`, with more details given below. 

A significant part of this PR are changes to expand the test suite to Rotations of arbitrary leading dimensions. All previous test cases have either been kept unchanged with additional test functions for the higher-dimensional cases, or have been generalized to multiple dimensions which contain the previous case as a special case for `ndim=1` or `2`.

All other classes interacting with `Rotation`s currently error out on encountering tensors of `Rotation`s.

## Exceptions from broadcasting rules
The resulting `Rotation`s from `from_euler` and `from_davenport` currently do not follow shape rules consistent with broadcasting logic. These functions have been redesigned for consistent behavior with special handling for 1D/2D cases to remain backwards compatible. Note that this arguably decreases the maintenance burden because we can get rid of the overly complex `_format_angles` function.

### from_euler
<details>

`from_euler` yields a (N, 4) quaternion for `seq="x"` and `angles.shape=(N,)`, but a (4,) quaternion for `seq="xyz"` and `angles.shape=(3,)` and an error for `angles.shape != (3,)`. The case of a single axes is treated special, which should not be the case. To see this, consider these two angle shapes which produce the same Rotation shapes at the moment:

```python
import numpy as np

print(R.from_euler("x", np.asarray([1])).as_quat().shape)  # (1, 4), should be (4,)
print(R.from_euler("xyz", np.asarray([1, 1, 1])).as_quat().shape)  # (4,)
print(R.from_euler("x", np.asarray([[1]])).as_quat().shape)  # (1, 4)
```
The rule proposed to generalize this to arbitrary shapes is the following:
- Promote 0D angle arrays, floats, ints to 1D (valid for single sequences)
- `angles.shape[-1]` must be equal to `len(seq)`
- The resulting quaternion has the shape `angles.shape[:-1] + (4,)` by applying the last dimension of angles to each axes described in `seq`.

For now we keep the legacy behavior to not break backward compatibility. I recommend discussing the behavior in a separate PR to sort out these inconsistencies.
</details>

### from_davenport 

<details>

`from_davenport` suffers from the same inconsistency. The only change is that the axes are passed directly to the function.

```python
import numpy as np

ax = np.array([0, 0, 1])

print(R.from_davenport(ax, "e", np.array([1])).as_quat().shape)  # (1, 4), should be (4,)
print(R.from_davenport(ax, "e", np.array([[1]])).as_quat().shape)  # (1, 4), correct
print(R.from_davenport(ax, "e", np.array([1, 1, 1])).as_quat().shape)  # (3, 4), should error. ax.shape[-2] != angles.shape[-1]
```

The rule proposed to generalize this to arbitrary shapes is the following:
- Promote 0D angle arrays, floats, ints to 1D (valid for single sequences)
- Promote axes to 2D
- `angles.shape[-1]` must be equal to `axes.shape[-2]`
- The resulting quaternion has the shape `angles.shape[:-1] + (4,)` by applying the last dimension of angles to each axes described in `axes`.

Note that this behavior is exactly the same as in `from_euler` if we map each element of `seq` to one basis vector. As in `from_euler`, we keep the legacy behavior for now. I recommend discussing the behavior together with `from_euler` in a separate PR.

</details>

# Documentation
It would be great to document the expanded capabilities of Rotation. Since we are not introducing breaking changes, updates to the docs have been deferred to a follow-up PR. This should make reviewing more manageable.

# References

### Issue
#23437 
Overall tracker for `Rotation` in #23391 
General issue for `Rotation` array API support #22500

### Related PRs
#22777, #23249

# What does this implement/fix?
Implement support for arbitrary leading dimensions for `Rotation`.


amacati:
CI failures are not related to this PR. Cupy throws multiple `cupy.cuda.compiler.CompileException`s, but tests for `spatial.transform` are passing as far as I could gather from the raw logs.

Should be ready for review.

amacati:
Test failures are still unrelated to this PR (`cupy` tests throw `cupy.cuda.compiler.CompileException`) . Any news on the review side @nmayorov  @scottshambaugh ?

scottshambaugh (approved):
Hey @amacati, this looks great! Please excuse the delay in reviewing, I just got back from an extended holiday. I agree with keeping the from_euler and from_davenport changes in a separate PR, and like your solution from a quick look.

I don't see any issues with the code changes (caveating that I'm not super knowledgeable about the different backends), and like that things are simplified in quite a few places. I'm not a huge fan of taking out some of the hardcoded values in the tests and replacing them with rand()'s since it makes things harder to inspect, but since it's parameterizing over several ndims it's definitely less messy this way. Agree with the decision to not cythonize higher order tensors.

With broadcasting, we should add a `shape` property to Rotations. Whether you want to include that here or put in a separate PR I'll leave up to you.

amacati:
No worries @scottshambaugh , and thanks for the review.

Adding `shape` makes a lot of sense. I included it with tests. There are some functions where an axis would also make sense now, e.g. `mean()`, but it's probably best to implement them only if there is an interest in using them.

While not part of this PR, what do you think is the best way forward with `from_euler` and `from_davenport`? The proposed changes would be breaking. Is this something for 2.0, or do we start putting out warnings?

scottshambaugh:
The shape property looks good to me! Along with axis arguments it also opens up possibilities around `reshape`, but we don't need to get into those weeds right now.

> While not part of this PR, what do you think is the best way forward with from_euler and from_davenport? The proposed changes would be breaking. Is this something for 2.0, or do we start putting out warnings?

My personal opinion is that when there is internally inconsistent behavior, it's okay to consider that a bug and make breaking changes to correct it. A brief note in the docs about the promotion to 1-D and how to recover a single rotation with `r[0]` indexing should suffice for anyone who needs to transition their code. But some maintainers should weigh in with their take here - I do expect that the single rotation case is the most common use for these functions, so the impact might be large.

amacati:
@lucascolley do you know by chance if @nmayorov is currently active on scipy? Would be great to have a rough timeline for getting this one merged. 

There is one PR I was planning to submit which adds broadcasting to `RigidTransform`, but it depends on this one and is thus currently stuck. Also, the docs should be updated in a follow-up PR to reflect the new capabilities. Finally there is the consistency issue with `from_euler` and `from_davenport`, which is another PR that depends on this one.

scipy/spatial/transform/_rotation.py
```diff
@@ -314,15 +349,19 @@ def __init__(
...
-        # Legacy behavior for cython backend: Differentiate between single quat and
-        # batched quats. We only use this for the cython backend. The Array API backend
-        # uses broadcasting by default and hence returns the correct shape without
-        # additional logic
+        quat = _promote(quat, xp)
+        if quat.shape[-1] != 4:
+            raise ValueError(
+                f"Expected `quat` to have shape (..., 4), got {quat.shape}."
+            )
+        # Single NumPy quats of list of quats are accelerated by the cython backend.
```
lucascolley: Is this supposed to say "or" rather than "of"?

scipy/spatial/transform/_rotation.py
```diff
@@ -2473,7 +2521,11 @@ def align_vectors(
                [0., 1., 2.]])
         """
         xp = array_namespace(a)
-        backend = backend_registry.get(xp, xp_backend)
+        a, b, weights = xp_promote(a, b, weights, force_floating=True, xp=xp)
```
lucascolley: could you add a comment explaining why `xp_promote` is used here while `_promote` is used elsewhere in the file?

scipy/spatial/transform/_rotation_xp.py
```diff
@@ -195,19 +190,57 @@ def from_euler(seq: str, angles: Array, degrees: bool = False) -> Array:
...
+    # 1d case:
+    # seq "x", angles (1,) -> q (1, 4)              Changed, q (4,)
+    # seq "xyz", angles (3,) -> q (4,)              Unchanged
+    # 2d case:
+    # seq "x", angles (N, 1) -> q (N, 4)            Unchanged
+    # seq "xyz", angles (N, 3) -> q (N, 4)          Unchanged
+    # After the 2d case, we follow standard broadcasting rules.
+    # TODO: Discuss with @lucascolley, @crusaderky and @scottshambaugh if this should
+    # be changed to be consistent with broadcasting rules. This would be a breaking
+    # change.
```
lucascolley, line 219: @j-bowhay @h-vetinari perhaps you have an opinion on whether this shape change ought to warrant a deprecation? See also from line 297

lucascolley (commented):
thanks @amacati !

scipy/spatial/transform/_rotation.py
amacati: Yes.

scipy/spatial/transform/_rotation.py
amacati: `_promote` was not designed for multiple arguments, but `a`, `b` and `weights` must be promoted together to yield consistent dtypes. I have rewritten `_promote` to accept multiple inputs. This is also useful in `from_davenport` where `axes` and `angles` were previously promoted independently from each other.

amacati:
Is the discussion on the Euler and Davenport angles a blocker? These changes would go into a separate PR anyways because they are behavior changes and no mere extensions to arbitrary dimensions.

I think keeping the comment is also fine for now because it needs to be resolved either by explicitly documenting the inconsistency or by resolving it properly.

lucascolley (approved):
not a blocker for this PR, let's give this a go. Thanks as always @amacati and for the review @scottshambaugh !

lucascolley:
there is however a merge conflict now

amacati:
Both CI test failures are unrelated to this PR. Should be good to go

lucascolley:
@amacati I noticed a failure coming from one of these tests with torch float32, does this diff look fine to you? It resolves the failures for me.

```diff
@@ -1298,9 +1298,10 @@ def test_mean(xp, ndim: int):
     desired = xp.zeros((1,) * (ndim - 1))
     if desired.ndim == 0:
         desired = desired[()]
+    atol = 1e-6 if xp_default_dtype(xp) is xp.float32 else 1e-10
     for t in thetas:
         r = Rotation.from_rotvec(t * axes)
-        xp_assert_close(r.mean().magnitude(), desired, atol=1e-10)
+        xp_assert_close(r.mean().magnitude(), desired, atol=atol)
```

amacati:
Yes, that is very much in line with what we are doing for other tests as well.

amacati:
@lucascolley By the way, this is completely unrelated, but do you know the approximate release schedule for 1.17? Some of the code bases at my lab are using the new rotation stuff already, and we're looking forward to being able to install via pip again.

lucascolley:
1.12.0 was released on January 20th, 1.15.0 was released on January 3rd. 1.17.0 should be on a similar schedule, so expect some time in January.

lucascolley:
@amacati out of interest, did you end up benchmarking along the lines of https://github.com/amacati/scipy/blob/fe407bfccf9be938b17ae0d48bb87846ad96a52e/xp_bench/rot_benchmark.py for batch support too? Of course we can be sure it is a lot better than serial, but I would be interested to see how much better, in the nice plots you had.

amacati:
No, those benchmarks were only running on batches with one leading dimension, and I haven't adapted them since.

PR #23468, opened 2025-08-15
DOC: optimize.shgo: correct `minimize_every_iter ` description

DKenefake:
Currently, The description for the argument ```minimize_every_iter``` has the following description. The second if true, should be a false. 

```
        minimize_every_iter : bool
            If True then promising global sampling points will be passed to a
            local minimization routine every iteration. If True then only the
            final minimizer pool will be run. Defaults to True.
```

[docs only]


lucascolley (approved):
thanks @DKenefake, looks good to me! Thanks for your first contribution to SciPy :)

PR #23471, opened 2025-08-16
BUG: optimize: fix the incompatible dtypes in L-BFGS-B

nabenabe0928:

This PR fixes dtype incompatibility between the C L-BFGS-B implementation and the Python module which calls the C implementation.

### DIFF
--- a/scipy/optimize/_lbfgsb_py.py
+++ b/scipy/optimize/_lbfgsb_py.py
@@ -440,8 +440,8 @@ def _minimize_lbfgsb(fun, x0, args=(), jac=None, bounds=None,
         raise ValueError('maxls must be positive.')
 
     x = array(x0, dtype=np.float64)
-    f = array(0.0, dtype=np.int32)
-    g = zeros((n,), dtype=np.int32)
+    f = array(0.0, dtype=np.float64)
+    g = zeros((n,), dtype=np.float64)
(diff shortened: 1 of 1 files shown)


andyfaff:
This looks correct.

Happens to work as it was though. On the first time round the loop `g` is cast to a double array, whilst `f` is zero going in, and then corrected the second time around the loop.

nickodell:
~~My first reaction to this was that the initial code looks like a memory safety problem. It allocates `g` as a buffer of n * 4 bytes, then passes that to a function expecting a buffer of size n * 8 bytes. On a second look, it seems this is not a memory safety problem, because in the loop, it runs [`g = g.astype(np.float64)`](https://github.com/scipy/scipy/blob/2887dcb3c34c4115b16948b8f8ece9800f974d79/scipy/optimize/_lbfgsb_py.py#L459), which re-allocates to a larger buffer. Later loops get their value of `g` from `func_and_grad()`.~~ Never mind, Andy made this point already.

Similarly, for `f`, it appears that when PyArgs_ParseTuple is given the `d` format specifier, this calls `PyFloat_AsDouble`, which takes care of the np.int32 -> double conversion by using the `__float__` method on the `np.int32` object.

Still, that seems needlessly confusing, so I think this is a good change.

ilayn:
That's me copy/pasting too fast, coincidence that it kept working. 

The failing tests are not related, thanks @nabenabe0928 and all.

Issue #23458, opened 2025-08-12
CI: CUDA compilation failures for all cupy tests


nickodell:
In #23457, there are many failures to compile CUDA code, in the "Run CuPy tests" job. However, this PR is not touching CUDA code, so it shouldn't have caused this.

[Log](https://github.com/scipy/scipy/actions/runs/16889996340/job/47847115432)

Here is a typical example:

```
2025-08-11T20:07:00.4517158Z =================================== FAILURES ===================================
2025-08-11T20:07:00.4517606Z __________________ TestNdimageFilters.test_correlate01[cupy] ___________________
2025-08-11T20:07:00.4517895Z 
2025-08-11T20:07:00.4518090Z self = <cupy.cuda.compiler._NVRTCProgram object at 0x7c9263f5dd00>
2025-08-11T20:07:00.4519252Z options = ('-I/home/admin/actions-runner/_work/scipy/scipy/.github/workflows/.pixi/envs/cupy/lib/python3.12/site-packages/cupy/_...y/.github/workflows/.pixi/envs/cupy/lib/python3.12/site-packages/cupy/_core/include', '-I/usr/local/cuda/include', ...)
2025-08-11T20:07:00.4520382Z log_stream = None
2025-08-11T20:07:00.4520507Z 
2025-08-11T20:07:00.4520631Z     def compile(self, options=(), log_stream=None):
2025-08-11T20:07:00.4521253Z         try:
2025-08-11T20:07:00.4521485Z             if self.name_expressions:
2025-08-11T20:07:00.4521757Z                 for ker in self.name_expressions:
2025-08-11T20:07:00.4522071Z                     nvrtc.addNameExpression(self.ptr, ker)
2025-08-11T20:07:00.4522387Z >           nvrtc.compileProgram(self.ptr, options)
2025-08-11T20:07:00.4522596Z 
2025-08-11T20:07:00.4523437Z log        = '/home/admin/actions-runner/_work/scipy/scipy/.github/workflows/.pixi/envs/cupy/lib/python3.12/site-packages/cupy/_cor...  ^\n\n6 errors detected in the compilation of "/tmp/tmp7qbmyxwe/f250abfd4bddbbda76b03bdc342c668e83c9c411.cubin.cu".\n'
2025-08-11T20:07:00.4524430Z log_stream = None
[...]

PR #23475, opened 2025-08-17
BUG: signal.lombscargle: replace in-place input adjustments

adammj:
See gh-23474

#### Reference issue
Closes gh-23474.

Replaces the two in-place operations that are performed on the inputs (in certain situations).

#### Additional information
None

### DIFF
--- a/scipy/signal/_spectral_py.py
+++ b/scipy/signal/_spectral_py.py
@@ -243,11 +243,11 @@ def lombscargle(
         )
 
     # weight vector must sum to 1
-    weights *= 1.0 / weights.sum()
+    weights = weights * (1.0 / weights.sum())
 
     # if requested, perform precenter
     if precenter:
-        y -= y.mean()
+        y = y - y.mean()
 
     # transform arrays
     # row vector
--- a/scipy/signal/tests/test_spectral.py
+++ b/scipy/signal/tests/test_spectral.py
@@ -1507,7 +1507,7 @@ class TestLombscargle:
         freqs = [np.pi/2.0] * 2  # must have 2+ elements
 
         lombscargle(t, y, freqs)
-    
+
     def test_input_mutation(self):
         # this tests for mutation of the input arrays
         # https://github.com/scipy/scipy/issues/23474
@@ -1521,8 +1521,8 @@ class TestLombscargle:
         p = 0.7  # Fraction of points to select
 
         # Randomly select a fraction of an array with timesteps
-        rng = np.random.RandomState(2353425)
-        r = rng.rand(nin)
+        rng = np.random.default_rng()
+        r = rng.random(nin)
         t = np.linspace(0.01*np.pi, 10.*np.pi, nin)[r >= p]
 
         # Plot a sine wave for the selected times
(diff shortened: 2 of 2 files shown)


tylerjereddy (commented):
Do you mind adding a regression test based on David's report?

When I check locally, the test below the fold fails before and passes after your patch.

<details>

```diff
--- a/scipy/signal/tests/test_spectral.py
+++ b/scipy/signal/tests/test_spectral.py
@@ -1021,6 +1021,15 @@ class TestSpectrogram:
         assert_allclose(p1, p3)
 
 class TestLombscargle:
+
+
+    def test_gh_23474(self):
+        x = np.arange(3)
+        y = np.arange(3, dtype=np.float64)
+        lombscargle(x, y, [1], precenter=True)
+        assert_allclose(y, [0, 1, 2])
+
+

```

</details>

scipy/signal/_spectral_py.py
```diff
@@ -243,11 +243,11 @@ def lombscargle(
         )
 
     # weight vector must sum to 1
-    weights *= 1.0 / weights.sum()
+    weights = weights * (1.0 / weights.sum())
```
tylerjereddy, line 246: It may also be sensible to add a regression test for this mutation fix as well--the one I just suggested doesn't fail if I revert this part of the patch of course.

We can handle that if you don't want to, should be straightforward/similar to the other one but for `weights`.

tylerjereddy:
I've tentatively added a backport label--if we unintentionally introduced mutation of user input arrays that is something we'd usually try to repair in a bug fix release (even if it involves discouraged/legacy arguments).

scipy/signal/_spectral_py.py
adammj: `weights` didn't exist before it was added by me (and in its current form). But, yes, I can also add a test for mutation.

scipy/signal/tests/test_spectral.py
```diff
@@ -1508,6 +1507,45 @@ def test_simple_div_zero(self):
...
+        # Input parameters
+        ampl = 2.
+        w = 1.
+        phi = 0.5 * np.pi
+        nin = 100
+        nout = 1000
+        p = 0.7  # Fraction of points to select
+
+        # Randomly select a fraction of an array with timesteps
+        rng = np.random.RandomState(2353425)
```
tylerjereddy: we usually try to use `default_rng` in new tests for a variety of reasons

scipy/signal/tests/test_spectral.py
```diff
@@ -1508,6 +1507,45 @@ def test_simple_div_zero(self):
...
+        f_org = f.copy()
+        weights_org = weights.copy()
+
+        pgram = lombscargle(t, y, f, precenter=True, weights=weights)
+
+        # check all 4 array inputs
+        assert_array_equal(t_org, t)
+        assert_array_equal(y_org, y)
+        assert_array_equal(f_org, f)
+        assert_array_equal(weights_org, weights)
```
tylerjereddy: Minor, but helpful convention to follow is `actual` argument first and `desired` value second per the documented function signature:

> assert_array_equal(actual, desired)

DietBru (approved):
The new `test_input_mutation`  could also be merged with the `test_amplitude` test. But I guess that is a matter  of personal taste.

Otherwise it looks good to me. Thanks for your work @adammj.

adammj:
@tylerjereddy I noticed that you add this to the 1.17.0 milestone, and @j-bowhay added the referenced bug (https://github.com/scipy/scipy/issues/23474) to the 1.16.2 milestone.

tylerjereddy:
Thanks Adam, I did one last confirmation that both `y` and `weights` are caught as mutated when the source code patch is reverted, so that looks good.

There's also core dev approval from someone more active than I in `signal`, so I went ahead and squash merged to facilitate backporting.

> I noticed that you add this to the 1.17.0 milestone

I wouldn't worry too much about our milestone convention--what is important here is that we have the "backport candidate" label--that's what I look for as the release manager when deciding what to include in a bug fix release (we then change the milestone after backporting).

tylerjereddy:
And the CI failure is obviously unrelated, that is gh-23458.

PR #21277, opened 2024-07-28
ENH: signal.lombscargle: update to the generalized Lomb-Scargle periodogram

adammj:
The function is a backwards-compatible replacement, but with additional functionality for sample weights and correctly removing any unknown y offset (including where the peaks are clipped, such that a naive mean removal will not work). Additional tests were created for the new options and outputs. Finally, files (and references to them) that are no longer needed were removed (such as the Cython and pythran version of the old code). The compiled versions are not needed, as the vectorized NumPy code is nearly the same speed as the nested for loops.

#### Reference issue
None, but discussed in more detail at [Scientific Python](https://discuss.scientific-python.org/t/fix-or-replace-lombscargle-function/).

#### What does this implement/fix?
Replaces the current Lomb-Scargle peridogram with the generalized version (published in 2009 [arxiv version of published paper link](https://arxiv.org/abs/0901.2573)) which enables sample weighting as well as more correctly removing any unknown y offset. Besides the one additional new parameter, the function is completely backwards compatible in API and the results (for the previous functionality).

#### Additional information
Since the compiled code is no longer needed, I removed the files and all references to them. New tests were added.

PR #23476, opened 2025-08-17
DEP: signal.lombscargle: deprecate argument `precenter`

adammj:
First stab at deprecating `precenter` argument, which was made useless after https://github.com/scipy/scipy/pull/21277 was merged last year.

#### Reference issue
None.

Flags `precenter` for deprecation, as it is legacy and no longer necessary—and less useful than the `floating_mean` argument (as already detailed in the Notes).

#### Additional information
Would need to be committed after gh-23475.


scipy/signal/_spectral_py.py
```diff
@@ -247,6 +249,12 @@ def lombscargle(
 
     # if requested, perform precenter
     if precenter:
+        msg = ("Use of parameter 'precenter' is deprecated as of SciPy "
+               "X.X.X and will be removed in X.X.X. Please leave "
+               "'precenter' unspecified; the operation is no longer necessary when "
+               "'floating_mean' is True. However, it can be exactly substituted by "
+               "passing 'y = (y - y.mean())' into the input.")
+        warnings.warn(msg, DeprecationWarning, stacklevel=2)
```
tylerjereddy, line 263: Since some tests still verify the correctness of `precenter` behavior, I needed a small patch locally to allow them to pass by ignoring the warning (for now).

<details>

```diff
--- a/scipy/signal/tests/test_spectral.py
+++ b/scipy/signal/tests/test_spectral.py
@@ -1095,6 +1095,7 @@ class TestLombscargle:
         # numerical differences when data is removed)
         assert_allclose(pgram[f==w], ampl, rtol=5e-2)
 
+    @pytest.mark.filterwarnings("ignore::DeprecationWarning")
     def test_precenter(self):
         # Test if precenter gives the same result as manually precentering
         # (for a very simple offset)
@@ -1389,6 +1390,7 @@ class TestLombscargle:
         weights = -np.ones(1)
         assert_raises(ValueError, lombscargle, t, y, f, weights=weights)
 
+    @pytest.mark.filterwarnings("ignore::DeprecationWarning")
     def test_list_input(self):
```

</details>

We'd also typically verify that the deprecation warning is correctly issued with i.e., a new test with `pytest.deprecated_call()` or similar.

If folks are in favor of this, we'd most likely do it in release `1.17.0` in late December/early January. Is it worth double checking with the `astropy` folks that were originally involved in the redesign discussion (if I'm remembering correctly)?

tylerjereddy:
I've added the `1.17.0` milestone so this doesn't fall off our radar, but we do have a bit more time to decide on this one vs. your preceding bug fix PR.

scipy/signal/_spectral_py.py
adammj: Yeah, I was lazy just turning out the commits instead of setting up the whole environment again. I'll go ahead and do it right so that I can fix and check for all these warnings.

I'll refresh myself on who was involved, and @ them in here (after I check if their repo even uses precenter).

adammj:
@neutrinoceros, since the changes to lombscargle last year started some discussion with [`astropy`](https://github.com/astropy/astropy), it made sense to check if this deprecation would cause any issues. However, looking over the `astropy` repo, I don't see any uses of the `precenter` argument. 

**correction:** there is one uncommitted `astropy` PR that mentions it. https://github.com/astropy/astropy/pull/17170

adammj:
I'm a little confused as to the purpose of the functions in `signal/_delegators.py`, and if anything needs to be updated there.

rgommers:
> I'm a little confused as to the purpose of the functions in `signal/_delegators.py`, and if anything needs to be updated there.

They are for supporting other array types. E.g., if you pass in a CuPy array (and have `SCIPY_ARRAY_API=1` set), then the code in `_delegators.py` will ensure that [cupyx.scipy.signal.lombscargle](https://docs.cupy.dev/en/stable/reference/generated/cupyx.scipy.signal.lombscargle.html) is called.

The CI failure is unrelated. I don't think you need to do anything to handle this deprecation in `_delegators.py`.

scipy/signal/_spectral_py.py
```diff
@@ -65,6 +65,7 @@ def lombscargle(
         output periodogram. Frequencies are normally >= 0, as any peak at ``-freq`` will
         also exist at ``+freq``.
     precenter : bool, optional
+        Parameter is deprecated and will be removed in version 1.19.0.
         Pre-center measurement values by subtracting the mean, if True. This is
         a legacy parameter and unnecessary if `floating_mean` is True.
```
DietBru, line 69: Wouldn't it be nicer to make the deprecation a little more prominent,like (e.g., compare [logm]):
```text
   precenter : bool, optional
       Pre-center measurement values by subtracting the mean, if True. This is
       a legacy parameter and unnecessary if `floating_mean` is True.
       
       .. versionchanged:: 1.17.0

            Parameter `precenter` is deprecated and will be removed in version 1.19.0.
            The precenter operation can be substituted by changing parameter `y` to
            ``y - y.mean()``.
```

[logm]: https://docs.scipy.org/doc/scipy-1.16.1/reference/generated/scipy.linalg.logm.html

scipy/signal/_spectral_py.py
```diff
@@ -103,12 +104,13 @@ def lombscargle(
     -----
     The algorithm used will not automatically account for any unknown y offset, unless
     floating_mean is True. Therefore, for most use cases, if there is a possibility of
-    a y offset, it is recommended to set floating_mean to True. If precenter is True,
-    it performs the operation ``y -= y.mean()``. However, precenter is a legacy
-    parameter, and unnecessary when floating_mean is True. Furthermore, the mean
-    removed by precenter does not account for sample weights, nor will it correct for
-    any bias due to consistently missing observations at peaks and/or troughs. When the
-    normalize parameter is "amplitude", for any frequency in freqs that is below
+    a y offset, it is recommended to set floating_mean to True. Parameter precenter is
```
DietBru: That `precenter` is deprecated is made clear above. Hence, I think it makes more sense to explain the general concept of "pre-centering" here (perhaps in a separate paragraph). Something along the lines of 

```text
The legacy concept of "pre-centering" entails removing the mean from parameter `y`
before processing, i.e., passing ``y - y.mean()`` instead of setting the parameter 
`floating_mean` to ``True``.
```
would suffice. It would also have the advantage that this paragraph will not have to be changed in version 1.19.

DietBru (changes requested):
I have just some suggestions for the docstr. Otherwise, it looks good to me.

scipy/signal/_spectral_py.py
adammj: I'm not sure where I took my example from, but I like this suggestion. Will tweak.

scipy/signal/_spectral_py.py
adammj: Good point. Again, I think I took that from some example I found elsewhere. But I like the idea of not having to change the paragraph again later.

Also, because I think it's helpful to drive the points home, I slightly reworded and rearranged the downsides of precenter to the upsides of floating_mean.

scipy/signal/_spectral_py.py
```diff
@@ -103,16 +108,17 @@ def lombscargle(
...
+    floating_mean accounts for sample weights, and will also correct for any bias due
+    to consistently missing observations at peaks and/or troughs. The legacy concept of
+    "pre-centering" entails removing the mean from parameter `y` before processing,
+    i.e., passing ``y - y.mean()`` instead of setting the parameter `floating_mean` to
+    ``True``. When the normalize parameter is "amplitude", for any frequency in freqs
+    that is below ``(2*pi)/(x.max() - x.min())``, the predicted amplitude will tend
+    towards infinity. The concept of a "Nyquist frequency" limit (see Nyquist-Shannon
+    sampling theorem) is not generally applicable to unevenly sampled data. Therefore,
+    with unevenly sampled data, valid frequencies in freqs can often be much higher
+    than expected for those familiar with methods like FFT.
```
DietBru: A purely cosmetic suggestion: Separating this text into paragraphs as well as marking `floating_mean` makes this note easier to read:

```suggestion
    The algorithm used will not automatically account for any unknown y offset, unless
    `floating_mean` is ``True``. Therefore, for most use cases, if there is a
    possibility of a y offset, it is recommended to set `floating_mean` to True.
    Furthermore, `floating_mean` accounts for sample weights, and will also correct for
    any bias due to consistently missing observations at peaks and/or troughs.

    The legacy concept of "pre-centering" entails removing the mean from parameter `y`
    before processing, i.e., passing ``y - y.mean()`` instead of setting the parameter
    `floating_mean` to ``True``.

    When the normalize parameter is "amplitude", for any frequency in freqs that is
    below ``(2*pi)/(x.max() - x.min())``, the predicted amplitude will tend towards
    infinity. The concept of a "Nyquist frequency" limit (see Nyquist-Shannon sampling
    theorem) is not generally applicable to unevenly sampled data. Therefore, with
    unevenly sampled data, valid frequencies in freqs can often be much higher than
    expected for those familiar with methods like FFT.
```

Otherwise, everything looks good.

adammj: It wouldn't let me "commit suggestion" using the UI, so I just committed it the normal way.

DietBru (approved):
Looks good now. Thanks, @adammj@adammj.

lucascolley:
Have we posted on the forum about this deprecation?

lucascolley:
> Have we posted on the forum about this deprecation?

it seems not. Just to put this (back) on your radars as the deadline for 1.17.0 is approaching @adammj @DietBru , would be good to have a simple post like https://discuss.scientific-python.org/t/dep-stats-ttest-ind-deprecate-random-state-permutation-require-keywords-for-most-args/1552, then we can merge this if no complaints.

adammj:
Thanks for the reminder, @lucascolley. It's been posted: https://discuss.scientific-python.org/t/dep-signal-lombscargle-deprecate-argument-precenter/2178

tylerjereddy (approved):
Let's give this a try. The CI is all green, and `astropy` folks didn't really object in the cross-linked PR (they mentioned working around the deprecation). The diff seems sensible to me, and the rendered docs for `lombscargle` seem decent to me with  the changes here.

It is true that the forum post has no responses, but that's not unusual for a deprecation query. We could wait a few more days, but this has the appropriate approvals and has been open for a few months now too, so I think I'd prefer to get it in and reduce our list of  milestoned PRs by one more. If someone really objects to the deprecation we may also be able to catch that in the RC phase and revert, but seems unlikey.

Thanks all, I'll probably squash merge, commit history a bit messy here.

PR #23478, opened 2025-08-17
DOC: Clarify parameter selection process in differential_evolution documentation

PriyadarshiAkshay:
#### Reference issue
The documentation for differential evolution mentioned 
`The final parameter is always loaded from ``b'```
which is different from how it is implemented in the code. 

In case of a single worker, `scipy.optimize._differentialevolution.DifferentialEvolutionSolver._mutate` 
defines the strategy of parameter selection. In the case of multiple workers, this is defined at 
`scipy.optimize._differentialevolution.DifferentialEvolutionSolver._mutate_many
`

Both of the functions show that the parameter to be mandatorily loaded from b' is controlled by `fill_point = rng_integers(rng, self.parameter_count)`.
For binomial crossover, it is a randomly selected parameter; for exponential crossover, fill_point is the starting point which is mandatorily loaded from b'

This fixes the comment and the documentation surrounding the mandatory selection of one parameter from b'.

### DIFF
--- a/scipy/optimize/_differentialevolution.py
+++ b/scipy/optimize/_differentialevolution.py
@@ -1728,16 +1728,19 @@ class DifferentialEvolutionSolver:
         crossovers = rng.uniform(size=(S, self.parameter_count))
         crossovers = crossovers < self.cross_over_probability
         if self.strategy in self._binomial:
-            # A randomly selected parameter is always from the bprime vector for binomial
-            # If you fill in modulo with a loop you have to set the last one to
-            # true. If you don't use a loop then you can have any random entry
-            # be True.
+            # A randomly selected parameter is always from the bprime vector for binomial crossover.
+            # The fill_point ensures at least one parameter comes from bprime, preventing
+            # the possibility of no mutation influence in the trial vector.
             i = np.arange(S)
             crossovers[i, fill_point[i]] = True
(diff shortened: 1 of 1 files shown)


scipy/optimize/_differentialevolution.py
```diff
@@ -318,8 +318,7 @@ def differential_evolution(func, bounds, args=(), strategy='best1bin',
     or the original candidate is made with a binomial distribution (the 'bin'
     in 'best1bin') - a random number in [0, 1) is generated. If this number is
     less than the `recombination` constant then the parameter is loaded from
-    ``b'``, otherwise it is loaded from the original candidate. The final
-    parameter is always loaded from ``b'``. Once the trial candidate is built
+    ``b'``, otherwise it is loaded from the original candidate. A randomly selected parameter is always loaded from ``b'``. For binomial crossover, this is a single random parameter. For exponential crossover, this is the starting point of a consecutive sequence of parameters from ``b'``. Once the trial candidate is built
```
andyfaff: These lines are too long. Please adjust line length to match the text above and below.

PriyadarshiAkshay: Fixed, thanks.

andyfaff:
Thank you @PriyadarshiAkshay

PR #22902, opened 2025-04-28
ENH:io.loadmat: Returning variable name for `mxOPAQUE_CLASS` variables

foreverallama:
Hi! 
This is a small change related to earlier PR #22847 on handling `mxOPAQUE_CLASS` object in `.mat` files. 

Currently, `scipy.io.loadmat` return `"None"` as the variable name for `mxOPAQUE_CLASS` types. This results in overwriting of the final dictionary when multiple opaque objects are present in a MAT-file.

Since the other PR on this topic is understandably very complex, and will have a longer developmental cycle, this PR instead focuses on a simple, functional fix: 
1. Return variable name for `mxOPAQUE_CLASS` objects
2. Update `OPAQUE_DTYPE` in `_mio5_params.py` to more appropriately reflect field headers for these object types

These fixes would enable custom wrappers (like [mine](https://github.com/foreverallama/matio)) to read opaque objects, while the development of the other PR continues.

## Additional Context

`mxOPAQUE_CLASS` data element format is as follows:
1. Array Flags (indicating `mxOPAQUE_CLASS`
2. Array Name (`int8` string)
3. Type System Name (`int8` string)
4. Class Name (`int8` string)
5. Object Metadata - Any of the other array types

PR #23481, opened 2025-08-19
ENH:io.matlab: Processing of mxOPAQUE_CLASS type during loadmat and savemat

foreverallama:
Hi!

I'd previously opened a PR on this topic (#22902 ), but accidentally closed it when I was reorganizing. I'd like to reintroduce it again, this time adding functionality for both load and save. Reiterating several issues that references the requirement of processing these types #11818 #12408 #13636 #15984 

Currently, `scipy.io.loadmat` returns `"None"` as the variable name for mxOPAQUE_CLASS types. This prevents the user from obtaining metadata for all variables in file as they get overwritten in the dictionary. Additionally, these variables (`MatlabOpaque`) are not processed during save.

Summary of the changes below:
* _mio5.py (VarWriter5): New method to write mxOPAQUE_CLASS objects to MAT-files
* _mio5.py (put_variables): Update to detect subsystem data in `mdict`
* _mio5_utils.pyx: Return variable name for mxOPAQUE_CLASS objects
* _mio5_params.py: Update OPAQUE_DTYPE field names

Some context about the `mxOPAQUE_CLASS` data element - there are 4 header fields and a data field as follows:
1. Array Flags (indicating `mxOPAQUE_CLASS`)
2. Array Name - int8 string
3. Type System - int8 string
4. Class Name - int8 string
5. Metadata - any of the other datatypes

This PR doesn't fully solve the issue of decoding `MatlabOpaque` types. The metadata and subsystem (what SciPy calls `__function_workspace__`) needs to be decoded to obtain object property values. However, these changes would allow the development of third-party libraries for this task. I've worked on it [separately](https://github.com/foreverallama/matio), and if there's sufficient interest I will port it over here. If you're taking a look, the main file is `subsystem.py`.

Tagging @matthew-brett again! Let me know if you need any further info!


rgommers:
Thanks @foreverallama! Is there a way you could add some tests, similar to some of the existing ones in `io/matlab/tests/`? Roundtrip save/load tests at a minimum seems feasible.

foreverallama:
Done! Added 3 tests with a new MAT-file:
1. Loading multiple `MatlabOpaque` objects from the same MAT-file
2. Saving a `MatlabOpaque` object
3. Saving `__function_workspace__`

In `_mio5.py: MatFileWriter5.put_variables()`, I also renamed the case from `__subsystem__` to `__function_workspace__` for consistency with what its currently called.

foreverallama:
@rgommers bump!

rgommers (approved):
Thanks @foreverallama, and apologies this fell through the cracks before. I now did a full review. The added test case was solid, it just missing the change to `meson.build` to install the file, hence the failures. I fixed that, and then also added the roundtrip (write-read) that I asked for in my initial review. With that, everything looks good.

***

I also investigated (in a container) whether adding these binary files from a new contributor are fine - in this case the answer is yes, the file was well-structured and there are no `pickle`-like concerns.

So, in it goes.

foreverallama:
Thanks for the review, glad to see this go through! My understanding of the MAT-file format has significantly grown since then, so I can help out with any related issues that pop up. Cheers!

rgommers:
Thanks, much appreciated @foreverallama

PR #23484, opened 2025-08-20
fix(signal): improve numerical stability of lfilter_zi

MengAiDev:
#### Reference issue
Closes #23480 

- Implement an approach similar to GNU Octave for better numerical stability in edge cases
- Remove leading zeros from denominator coefficients
- Normalize coefficients if a[0] != 1.0
- Use cumulative sum method for computing the steady-state response

### DIFF
--- a/scipy/signal/_signaltools.py
+++ b/scipy/signal/_signaltools.py
@@ -4238,6 +4238,9 @@ def lfilter_zi(b, a):
     assuming ``a[0]`` is 1.0; if ``a[0]`` is not 1, `a` and `b` are first
     divided by a[0].
 
+    This implementation uses an approach similar to GNU Octave for better
+    numerical stability in edge cases.
+
     Examples
     --------
     The following code creates a lowpass Butterworth filter. Then it
@@ -4269,14 +4272,7 @@ def lfilter_zi(b, a):
     """
     xp = array_namespace(b, a)
 
-    # FIXME: Can this function be replaced with an appropriate
-    # use of lfiltic?  For example, when b,a = butter(N,Wn),
-    #    lfiltic(b, a, y=numpy.ones_like(a), x=numpy.ones_like(b)).
-    #
-
-    # We could use scipy.signal.normalize, but it uses warnings in
-    # cases where a ValueError is more appropriate, and it allows
-    # b to be 2D.
+    # Convert to arrays
     b = xpx.atleast_nd(xp.asarray(b), ndim=1, xp=xp)
     if b.ndim != 1:
         raise ValueError("Numerator b must be 1-D.")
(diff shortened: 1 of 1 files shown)


scipy/signal/_signaltools.py
```diff
@@ -4269,57 +4272,34 @@ def lfilter_zi(b, a):
...
-    #     zi[k] = asum*zi[0] - csum
+    # Use GNU Octave's approach for better numerical stability
+    # This approach directly computes the steady-state response
+    kdc = xp.sum(b) / xp.sum(a)
+    if xp.abs(kdc) < xp.inf:
+        # Compute si using the cumsum approach from GNU Octave
+        si = xp.flip(xp.cumsum(xp.flip(b - kdc * a)))
+        zi = si[1:]  # Remove first element
+    else:
+        zi = xp.zeros(len(a) - 1, dtype=b.dtype)
```
tylerjereddy, line 4302: I'm sorry to bring this up, but it seems that GNU octave have a GPL-3 license, which is copyleft.

We have to ask if you looked at the GNU Octave source code, and if so, I believe we cannot accept such contributions unfortunately.

It would be totally fine to actually execute their code and design a solution based on that, bit if you looked at the source (even just for inspiration, I believe) it would be risky for us to accept such a patch.

If their implementation is based on a published paper, reading the paper would be totally fine.

One other thing I might point out, is that I believe the matching issue/ticket also has the problematic-provenance GNU Octave source pasted in, so arguably anyone who looks at that ticket also can't contribute a potential patch--as crazy as it sounds, we might want to hide that code in the ticket as well and briefly summarize what is going on over there to increase the likelihood that someone coming across the issue can resolve it in a license-compatible way.

nickodell: >If their implementation is based on a published paper, reading the paper would be totally fine.

I've read the GPL'd source code, and there is a comment on the part relating to this code citing Likhterov & Kopeika, 2003. "Hardware-efficient technique for minimizing startup transients in Direct Form II digital filters." I believe it would be fine for an implementer to read this paper.

nickodell: >One other thing I might point out, is that I believe the matching issue/ticket also has the problematic-provenance GNU Octave source pasted in, so arguably anyone who looks at that ticket also can't contribute a potential patch--as crazy as it sounds, we might want to hide that code in the ticket as well and briefly summarize what is going on over there to increase the likelihood that someone coming across the issue can resolve it in a license-compatible way.

Agreed. I've made that edit.

nickodell:
I'm closing this PR due to concerns that it was based upon GPL licensed code. If this is incorrect, please let us know and we'll re-open.

PR #16004, opened 2022-04-18
MAINT: rename MaximumFlowResult.residual to flow

fuglede:
The attribute `MaximumFlowResult.residual` was poorly named: It in fact
represents the flow function of the maximum flow, whereas the residual
is given by the difference between the input graph and this flow
function. As such, we simply rename it from `residual` to `flow` to
avoid the risk of confusion.

A different possible fix would be to change the attribute to actually be
the residual graph, but that would change its value and therefore lead
to a slightly trickier deprecation. The flow is also useful in itself,
so for now, we simply include that, and let it be up to the user to make
the more or less trivial calculation of the residual if that's ever
needed.

This closes https://github.com/scipy/scipy/issues/14731

I think this is the first time I add a `DeprecationWarning`, so I'd appreciate
if the prospective reviewer could take a look at the form of that.

cc @czgdp1807 @pmla

PR #23491, opened 2025-08-22
DOC: signal.detrend: clarify `overwrite_data` behaviour

### LINKED ISSUE #14498: overwrite_data=True kwarg ignored when type="constant" in scipy.signal.detrend
The issue is that the current version of scipy.signal (1.7.0) takes a keyword argument overwrite_data.  The intended behavior is that the input array, called "data" in the example below should be overwritten by the array of detrended_data within the call to scipy.signal.detrend if overwrite_data is set to True.  The expected behaviour is observed when linear detrending is
applied, but not when constant detrending (mean-subtraction) is applied.

In the code sample provided below I have labelled the arrays printed to screen (1) ... (6).
I expect arrays (2) and (3) to be identical and they are, but I likewise expect arrays (5) and (6) to be identical and they are not.

The issue can be probabaly be handled by changing scipy.signal lines 3455-3457 from:

    if type in ['constant', 'c']:
        ret = data - np.mean(data, axis, keepdims=True)
        return ret

to:

    if type in ['constant', 'c']:
        data = data - np.mean(data, axis, keepdims=True)
        return data

```
import numpy as np
import scipy.signal as ssig

np.random.seed(1)

n_rows = 3
n_cols = 2
data = np.random.random((n_rows, n_cols))

print("Expected behavior:\n")
print("(1) Original Data")
print(data)

detrended_data = ssig.detrend(data, type="linear", overwrite_data=True)

print("\n(2) Linear detrended data")
print(detrended_data)
print("\n(3) Data which was successfully overwritten")
print(data)


np.random.seed(1)
data = np.random.random((n_rows, n_cols))
print("\n\n\nUnexpected behavior:\n")
print("(4) Original Data (again)")
print(data)

detrended_data = ssig.detrend(data, type="constant", overwrite_data=True)

print("\n(5) Constant detrended data")
print(detrended_data)
print("\n(6) Data which was not successfully overwritten")
print(data)

```

#### Expected output from sample code message:

```
Expected behavior:

(1) Original Data
[[4.17022005e-01 7.20324493e-01]
 [1.14374817e-04 3.02332573e-01]
 [1.46755891e-01 9.23385948e-02]]

(2) Linear detrended data
[[ 5.55111512e-17  0.00000000e+00]
 [ 2.22044605e-16  1.66533454e-16]
 [-2.77555756e-17 -1.38777878e-17]]

(3) Data which was successfully overwritten
[[ 5.55111512e-17  0.00000000e+00]
 [ 2.22044605e-16  1.66533454e-16]
 [-2.77555756e-17 -1.38777878e-17]]



Unexpected behavior:

(4) Original Data (again)
[[4.17022005e-01 7.20324493e-01]
 [1.14374817e-04 3.02332573e-01]
 [1.46755891e-01 9.23385948e-02]]

(5) Constant detrended data
[[-0.15165124  0.15165124]
 [-0.1511091   0.1511091 ]
 [ 0.02720865 -0.02720865]]

(6) Data which was not successfully overwritten
[[4.17022005e-01 7.20324493e-01]
 [1.14374817e-04 3.02332573e-01]
 [1.46755891e-01 9.23385948e-02]]

Process finished with exit code 0
```

OleBialas:
The argument `overwrite_data=True` made it appear as if overwriting was guaranteed. However, overwriting only happens when ``type == linear`` and ``dtype is in "dtDF"`` and array reshaping returns a view. I changed the argument name to `allow_overwrite` and updated the docstring to indicate overwriting only happens under specific conditions.

#### Reference issue

Closes gh-14498

This makes the behavior for in place modifications more explicit

### DIFF
--- a/scipy/signal/_signaltools.py
+++ b/scipy/signal/_signaltools.py
@@ -4052,7 +4052,7 @@ def vectorstrength(events, period):
 
 def detrend(data: np.ndarray, axis: int = -1,
             type: Literal['linear', 'constant'] = 'linear',
-            bp: ArrayLike | int = 0, overwrite_data = False) -> np.ndarray:
+            bp: ArrayLike | int = 0, overwrite_data:bool = False) -> np.ndarray:
     r"""Remove linear or constant trend along axis from data.
 
     Parameters
--- a/scipy/signal/_signaltools.py
+++ b/scipy/signal/_signaltools.py
@@ -4052,7 +4052,7 @@ def vectorstrength(events, period):
 
 def detrend(data: np.ndarray, axis: int = -1,
             type: Literal['linear', 'constant'] = 'linear',
-            bp: ArrayLike | int = 0, overwrite_data:bool = False) -> np.ndarray:
+            bp: ArrayLike | int = 0, overwrite_data: bool = False) -> np.ndarray:
     r"""Remove linear or constant trend along axis from data.
 
     Parameters
--- a/scipy/signal/_signaltools.py
+++ b/scipy/signal/_signaltools.py
@@ -4075,7 +4075,7 @@ def detrend(data: np.ndarray, axis: int = -1,
     overwrite_data: bool, optional
         If True, allow in place detrending and avoid a copy. Default is
         False. In place modification applies only if ``type == 'linear'``
-        and also depends in the data type and memory layout of `data`.
+        and also depends on the data type and memory layout of `data`.
 
     Returns
     -------
--- a/scipy/signal/_signaltools.py
+++ b/scipy/signal/_signaltools.py
@@ -32,6 +32,7 @@ from scipy._lib._array_api import (
 )
 from scipy._lib.array_api_compat import is_array_api_obj
 import scipy._lib.array_api_extra as xpx
+from scipy._lib.deprecation import _NoValue
 
 
 __all__ = ['correlate', 'correlation_lags', 'correlate2d',
@@ -4052,7 +4053,7 @@ def vectorstrength(events, period):
 
 def detrend(data: np.ndarray, axis: int = -1,
             type: Literal['linear', 'constant'] = 'linear',
-            bp: ArrayLike | int = 0, allow_overwrite: bool = False) -> np.ndarray:
+            bp: ArrayLike | int = 0, overwrite_data = False) -> np.ndarray:
     r"""Remove linear or constant trend along axis from data.
 
     Parameters
@@ -4072,7 +4073,7 @@ def detrend(data: np.ndarray, axis: int = -1,
         performed for each part of `data` between two break points.
         Break points are specified as indices into `data`. This parameter
         only has an effect when ``type == 'linear'``.
-    allow_overwrite: bool, optional
+    overwrite_data: bool, optional
         If True, allow in place detrending and avoid a copy. Default is
         False. In place modification applies only if ``type == 'linear'``
         and also depends in the data type and memory layout of `data`.
--- a/scipy/signal/_signaltools.py
+++ b/scipy/signal/_signaltools.py
@@ -32,7 +32,6 @@ from scipy._lib._array_api import (
 )
 from scipy._lib.array_api_compat import is_array_api_obj
 import scipy._lib.array_api_extra as xpx
-from scipy._lib.deprecation import _NoValue
 
 
 __all__ = ['correlate', 'correlation_lags', 'correlate2d',
--- a/scipy/signal/_signaltools.py
+++ b/scipy/signal/_signaltools.py
@@ -4052,7 +4052,7 @@ def vectorstrength(events, period):
 
 def detrend(data: np.ndarray, axis: int = -1,
             type: Literal['linear', 'constant'] = 'linear',
-            bp: ArrayLike | int = 0, overwrite_data: bool = False) -> np.ndarray:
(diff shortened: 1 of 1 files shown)


OleBialas:
@Kai-Striega

lucascolley:
Thanks @OleBialas !

Would you mind splitting the formatting changes into a separate commit? Then we can add it to https://github.com/scipy/scipy/blob/main/.git-blame-ignore-revs to hide them from git blame. It also makes it easier to review the real changes here.

OleBialas:
The formatting was done by accident, I removed the changes

lucascolley (commented):
I think we will need to introduce a deprecation cycle, where the old parameter name is kept around as an alias for the new one, but raises a deprecation warning if used. There is an example of a similar change at https://github.com/scipy/scipy/pull/16004/files.

I think we should raise an exception if both `overwrite_data` and `allow_overwrite` are passed with different values.

We probably also want to keep the `overwrite_data` parameter in the docstring, but add a `.. deprecated:: 1.17.0` admonition, like those seen at https://github.com/scipy/scipy/commit/ca34a7735cae547568ef88729432ebaed7403275.

j-bowhay (dismissed):
Thanks for your contribution @OleBialas!

I agree if starting from scratch that `allow_overwrite` would be a better name. Unfortunately, we cannot proceed with your proposed renaming as it would break backwards compatibility. I would propose keeping the old name and improving the documentation (as you have done)

lucascolley:
I'm also open to Jake's suggestion of just improving the documentation.

OleBialas:
I'm fine with either introducing the deprecation warning or updating only the documentation

j-bowhay:
I would prefer just to update the documentation as this stage as a deprecation is a lot more effort. Then if someone still finds it confusing in the future we can go through that whole process

DietBru:
What do you think about keeping the function signature they way it is and refactor the function a little:

* Implement overwriting for `type` = `'constant'`.
* Implement the `bp` parameter for `type` = `'constant'`.
* Raise a `ValueError` if `overwrite_data` is `True` and overwriting is not possible.

It shouldn't be too hard and I think that would make the behavior of `detrend` much more intuitive. Since the behavior of the default parameters does not change, it shouldn't annoy too many people. For now, only the docstr could be changed and the refactoring could be done in a later PR.

OleBialas:
I was actually wondering why overwriting is completely ignored for constant detrending. For now, I updated my PR to only change the documentation, but I would be open to work on the suggestion by @DietBru in a separate issue

scipy/signal/_signaltools.py
```diff
@@ -4072,8 +4072,10 @@ def detrend(data: np.ndarray, axis: int = -1,
         performed for each part of `data` between two break points.
         Break points are specified as indices into `data`. This parameter
         only has an effect when ``type == 'linear'``.
-    overwrite_data : bool, optional
-        If True, perform in place detrending and avoid a copy. Default is False
+    overwrite_data: bool, optional
+        If True, allow in place detrending and avoid a copy. Default is
+        False. In place modification applies only if ``type == 'linear'``
+        and also depends in the data type and memory layout of `data`.
```
lucascolley:
```suggestion
        and also depends on the data type and memory layout of `data`.
```

lucascolley (commented):
image

preview looks good!

> overwriting only happens when ``type == linear`` and ``dtype is in "dtDF"`` and array reshaping returns a view.

maybe we should spell out these last two points in the Notes section? What do you think @DietBru @j-bowhay ?

DietBru:
Yup. Might be useful to note that overwriting only works for floats. In a follow-up PR overwriting could be also made workable for complex-valued inputs.

scipy/signal/_signaltools.py
```diff
@@ -4072,8 +4072,10 @@ def detrend(data: np.ndarray, axis: int = -1,
         performed for each part of `data` between two break points.
         Break points are specified as indices into `data`. This parameter
         only has an effect when ``type == 'linear'``.
-    overwrite_data : bool, optional
-        If True, perform in place detrending and avoid a copy. Default is False
+    overwrite_data: bool, optional
+        If True, allow in place detrending and avoid a copy. Default is
+        False. In place modification applies only if ``type == 'linear'``
+        and also depends on the data type and memory layout of `data`.
```
DietBru: Referring to @lucascolley's [comment]:
 
* The `dfDF` dtypes transform to:
   ```python 
   >>> import numpy as np
   >>> {c_: f"{np.zeros(0, dtype=c_).dtype}" for c_ in 'dfDF'}
   {'d': 'float64', 'f': 'float32', 'D': 'complex128', 'F': 'complex64'}
   ```
* As far as I understand the code, the array reshaping always returns a view.

Hence, I think it would be sufficient to complete this PR by adding:
```suggestion
        and `data` is of the floating point dtype ``float32``, ``float64``,
        ``complex64`` or ``complex128``.
```

What do you think, @OleBialas?

[comment]: https://github.com/scipy/scipy/pull/23491#issuecomment-3213649342

OleBialas: Looks good - the more specific the documentation, the better!

DietBru (approved):
The [rendered docstr] looks good to me.

[rendered docstr]: https://output.circle-artifacts.com/output/job/e0037579-9457-4e1e-a505-c8fc98ab5edd/artifacts/0/html/reference/generated/scipy.signal.detrend.html

lucascolley (approved):
Also LGTM, thanks all!

PR #18147, opened 2023-03-14
add a case dealing with improper integral

Shengyis:
Add an integral bound check at the very beginning of quad, if the upper and lower bounds are the same, then directly return 0 with error 0. This will fix a problem related to an improper integral such as \int_0^0 x**(-0.5). Now it will return 0 correctly instead of ZeroDIvisionError. And other n-d integrals depending on quad will perform similarly. For example, \int_0^1 \int_0^0 x**(-0.5)y**(-0.5) dx dy will return 0 as well.

#### Reference issue

Closes #18142

PR #23494, opened 2025-08-22
ENH: integrate.quad: shortcut if upper/lower bounds are equal

### LINKED ISSUE #18142: ENH: scipy.integrate.quad, adding a case that if upper and lower integral bounds are the same then return zero

I know for normal integration with the same upper and lower bounds, quad will return zero without any problem. However, if it is an improper integral, sometimes it will give ZeroDivisionError. For example, I want to calculate \int_0^0 1/x^0.5 dx, the python code is 

def f(x):
    return x**(-0.5)
print(quad(f, 0, 0))

this will give ZeroDivisionError. However, in the view of math, since it is integrable neasr x = 0, this should return 0. 

maybe just a few lines in quad function to judge whether two bounds are the same. If they are the same directly return 0 with err = 0, otherwise following the original quad. 

_No response_

_No response_

JostMigenda:
#### Reference issue
Closes gh-18142.

Shortcut integration if upper and lower integration bounds are equal. See original issue for details.

#### Additional information
This is based on an earlier PR (gh-18147), which was deleted without merging. Compared to that, this PR eliminates a `RuntimeWarning` when running with `full_output=1` (`np.full(limit, np.nan, dtype=np.int32)` tried to cast `np.nan` to an integer) and adds `full_output` to the test.

### DIFF
--- a/scipy/integrate/_quadpack_py.py
+++ b/scipy/integrate/_quadpack_py.py
@@ -432,6 +432,18 @@ def quad(func, a, b, args=(), full_output=0, epsabs=1.49e-8, epsrel=1.49e-8,
     if not isinstance(args, tuple):
         args = (args,)
 
+    # Shortcut for empty interval, also works for improper integrals.
+    if a == b:
+        if full_output == 0:
+            return (0., 0.)
+        else:
+            return (0., 0., {"neval": 0, "last": 0,
+                             "alist": np.full(limit, np.nan, dtype=np.float64),
+                             "blist": np.full(limit, np.nan, dtype=np.float64),
+                             "rlist": np.zeros(limit, dtype=np.float64),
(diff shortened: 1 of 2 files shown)


lucascolley (approved):
thanks @JostMigenda, LGTM!

j-bowhay:
Sorry I am late to the review here but it would be good to add tests to confirm that this leads to the correct behavior `dblquad`, `tplquad`, and `nquad` as these all call `quad` under the hood.

Also what about `complex_func=True`? we should check the behavior here is also correct

lucascolley:
good point, sorry for missing those nuances — would you be able to work on a follow-up @JostMigenda ?

JostMigenda:
I’ll take a look, yep.

Issue #20930, opened 2024-06-09
ENH: integrate: add array API-support


mdhaber:
Towards gh-18867

This issue tracks progress toward the addition of array-API support to `scipy.integrate` functions. The functions listed below look ready for conversion.

- [ ] `trapezoid`
- [ ] `simpson`
- [ ] `romb`
- [ ] `cumulative_trapezoid`
- [ ] `cumulative_simpson`

I'd like to work on:
- [ ] `qmc_quad`

These would be pretty substantial undertakings; there would need a pretty dedicated team of author(s)/reviewer(s). I'd suggest holding off on them until some of the lower-hanging fruit is picked. @nmayorov would you be interested in being part of this?
- [ ] `solve_ivp`
- [ ] `solve_bvp`

Of course, adaptive quadrature given a callable rather than points is also a very high priority. We'll see where gh-20252 is at the end of the summer.

I think we could convert these functions, but I'm not sure whether doing so is a high priority. LMK your thoughts.
- [ ] `newton_cotes`
- [ ] `fixed_quad` (IIUC, automatic differentiation would be easier with something non-adaptive like this. I'm assuming that gh-20252 will keep this in mind.)

I don't think these will be converted, per se.
- `quad_vec` - either the outcome of gh-20252 or `_tanhsinh` would be a suitable array-API compatible replacement
- `quad` - Fortran, and same as above
- `odeint`(and related) - Fortran
- `dblquad`, `tplquad`, `nquad` - only relevant if `quad` were converted

j-bowhay:
Just to throw an idea out there, `solve_ivp`, whilst very solid and battle tested, is now a little behind other offerings (DifferentialEquations.jl) so if it is going to be a substantial effort to rewrite it would be interesting to investigate possible alternatives. For example, perhaps considering the feasibility of a slimed down array API compatible version of https://docs.kidger.site/diffrax/

lucascolley:
> perhaps considering the feasibility of a slimed down array API compatible version of [docs.kidger.site/diffrax](https://docs.kidger.site/diffrax/)

x-ref https://github.com/scipy/scipy/pull/20085#issuecomment-2029349921

PR #23497, opened 2025-08-22
ENH: integrate.romb: array-API support

VeckoTheGecko:
#### Reference issue

Contributes to #20930

- Fixes `romb` test cases to use `xp` fixture (tests were silently skipping before)
- Adds array API standard support to `romb`
- Adds docstring example where table can't be printed

#### Additional information
Not sure exactly what the prior return type of the romb function was (a scalar array, or just a scalar value). Ran out of time today to investigate - happy to followup on this, just uploading now so maintainers can review.

---

*First time contribution as part of sprinting with EuroScipy2025. Feel free to call out teachable moments.*

### DIFF
--- a/scipy/integrate/tests/test_quadrature.py
+++ b/scipy/integrate/tests/test_quadrature.py
@@ -40,11 +40,11 @@ class TestFixedQuad:
 @make_xp_test_case(romb)
 class TestRomb:
     def test_romb(self, xp):
-        xp_assert_equal(romb(xp.arange(17)), 128.0)
+        xp_assert_equal(romb(xp.arange(17, dtype=xp.float64)), 128.0)
 
     def test_romb_gh_3731(self, xp):
         # Check that romb makes maximal use of data points
-        x = xp.arange(2**4+1)
+        x = xp.arange(2**4+1, dtype=xp.float64)
         y = xp.cos(0.2*x)
         val = romb(y)
         val2, err = quad(lambda x: xp.cos(0.2*x), x.min(), x.max())
--- a/scipy/integrate/tests/test_quadrature.py
+++ b/scipy/integrate/tests/test_quadrature.py
@@ -15,7 +15,7 @@ from scipy.integrate._quadrature import _cumulative_simpson_unequal_intervals
 from scipy import stats, special, integrate
 from scipy.conftest import skip_xp_invalid_arg
 from scipy._lib._array_api import make_xp_test_case
-from scipy._lib._array_api_no_0d import xp_assert_close
+from scipy._lib._array_api_no_0d import xp_assert_close, xp_assert_equal
 
 skip_xp_backends = pytest.mark.skip_xp_backends
 
@@ -39,16 +39,16 @@ class TestFixedQuad:
 
 @make_xp_test_case(romb)
 class TestRomb:
-    def test_romb(self):
-        assert_equal(romb(np.arange(17)), 128)
+    def test_romb(self, xp):
+        xp_assert_equal(romb(xp.arange(17)), 128.0)
 
-    def test_romb_gh_3731(self):
+    def test_romb_gh_3731(self, xp):
         # Check that romb makes maximal use of data points
-        x = np.arange(2**4+1)
-        y = np.cos(0.2*x)
+        x = xp.arange(2**4+1)
+        y = xp.cos(0.2*x)
         val = romb(y)
-        val2, err = quad(lambda x: np.cos(0.2*x), x.min(), x.max())
-        assert_allclose(val, val2, rtol=1e-8, atol=0)
+        val2, err = quad(lambda x: xp.cos(0.2*x), x.min(), x.max())
+        xp_assert_close(val, val2, rtol=1e-8, atol=0)
 
 
 @make_xp_test_case(newton_cotes)
--- a/scipy/integrate/tests/test_quadrature.py
+++ b/scipy/integrate/tests/test_quadrature.py
@@ -44,10 +44,10 @@ class TestRomb:
 
     def test_romb_gh_3731(self, xp):
         # Check that romb makes maximal use of data points
-        x = xp.arange(2**4+1, dtype=xp.float64)
-        y = xp.cos(0.2*x)
-        val = romb(y)
-        expected, err = quad(lambda x: np.cos(np.array(0.2*x)), xp.min(x), xp.max(x))
+        x = np.arange(2**4+1, dtype=np.float64)
+        y = np.cos(0.2*x)
+        val = romb(xp.asarray(y, dtype=xp.float64))
+        expected, err = quad(lambda x: np.cos(np.array(0.2*x)), np.min(x), np.max(x))
         xp_assert_close(val, xp.asarray(expected), rtol=1e-8, atol=0)
 
 
--- a/scipy/integrate/tests/test_quadrature.py
+++ b/scipy/integrate/tests/test_quadrature.py
@@ -40,7 +40,7 @@ class TestFixedQuad:
 @make_xp_test_case(romb)
 class TestRomb:
     def test_romb(self, xp):
-        xp_assert_equal(romb(xp.arange(17, dtype=xp.float64)), xp.asarray(128.0))
+        xp_assert_equal(romb(xp.arange(17.0)), xp.asarray(128.0))
 
     def test_romb_gh_3731(self, xp):
         # Check that romb makes maximal use of data points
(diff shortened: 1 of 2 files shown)


VeckoTheGecko:
cc @lucascolley

scipy/integrate/tests/test_quadrature.py
```diff
@@ -39,16 +39,17 @@ def test_vector(self):
 
 @make_xp_test_case(romb)
 class TestRomb:
-    def test_romb(self):
-        assert_equal(romb(np.arange(17)), 128)
+    def test_romb(self, xp):
+        xp_assert_equal(romb(xp.arange(17, dtype=xp.float64)), xp.asarray(128.0))
 
-    def test_romb_gh_3731(self):
+    @pytest.mark.skip_xp_backends(np_only=True, reason="`quad` not yet implemented using array API")
```
j-bowhay:
```suggestion
```

scipy/integrate/tests/test_quadrature.py
```diff
@@ -39,16 +39,17 @@ def test_vector(self):
...
         # Check that romb makes maximal use of data points
-        x = np.arange(2**4+1)
-        y = np.cos(0.2*x)
+        x = xp.arange(2**4+1, dtype=xp.float64)
+        y = xp.cos(0.2*x)
         val = romb(y)
-        val2, err = quad(lambda x: np.cos(0.2*x), x.min(), x.max())
-        assert_allclose(val, val2, rtol=1e-8, atol=0)
+        val2, err = quad(lambda x: xp.cos(0.2*x), x.min(), x.max())
+        xp_assert_close(val, val2, rtol=1e-8, atol=0)
```
j-bowhay:
```suggestion
        xp_assert_close(val, xp.asarray(val2), rtol=1e-8, atol=0)
```

j-bowhay (commented):
Thanks @VeckoTheGecko! RE the test that uses `quad`, we can compute the reference value using numpy then call `xp.asarray` on the output

VeckoTheGecko:
Thanks for the review @j-bowhay ! Done now

VeckoTheGecko:
> Not sure exactly what the prior return type of the romb function was (a scalar array, or just a scalar value). 

Turns out on `main` it returns a numpy float value instead of an array. @lucascolley you mentioned that this behaviour might not be compatible across array backends - any ideas how to proceed? Perhaps this is something that should resolved outside of this PR since it reached beyond just this function (and to all scalar output that scipy gives)

`main` behaviour:
- numpy: numpy float output for 1D input, array output for ND input
- other backends: not implemented

this PR behaviour:
- numpy: unchanged
- other backends: array output

lucascolley:
Thanks for investigating, Nick!

In that case, since we are already using the assertions from https://github.com/scipy/scipy/blob/main/scipy/_lib/_array_api_no_0d.py, I think there is nothing more to do if the tests already pass! They should be checking that we continue to return NumPy scalars rather than 0D arrays.

If the tests fail because we are returning 0D arrays, we should follow the pattern at https://github.com/scipy/scipy/blob/main/scipy/integrate/_tanhsinh.py#L521 and return the result with `[()]`.

cc @mdhaber, I think I am remembering correctly but it has been a while since I thought about this.

lucascolley:
> this PR behaviour:
> - numpy: unchanged

Sounds like things should pass, then!

scipy/integrate/tests/test_quadrature.py
```diff
@@ -39,16 +39,16 @@ def test_vector(self):
...
+    def test_romb_gh_3731(self, xp):
         # Check that romb makes maximal use of data points
-        x = np.arange(2**4+1)
-        y = np.cos(0.2*x)
+        x = xp.arange(2**4+1, dtype=xp.float64)
+        y = xp.cos(0.2*x)
         val = romb(y)
-        val2, err = quad(lambda x: np.cos(0.2*x), x.min(), x.max())
-        assert_allclose(val, val2, rtol=1e-8, atol=0)
+        expected, err = quad(lambda x: np.cos(np.array(0.2*x)), xp.min(x), xp.max(x))
```
lucascolley:
```suggestion
        expected, err = quad(lambda x: np.cos(np.array(0.2*x)), np.min(x), np.max(x))
```

scipy/integrate/tests/test_quadrature.py
```diff
@@ -39,16 +39,16 @@ def test_vector(self):
...
+        xp_assert_equal(romb(xp.arange(17.0)), xp.asarray(128.0))
 
-    def test_romb_gh_3731(self):
+    def test_romb_gh_3731(self, xp):
         # Check that romb makes maximal use of data points
-        x = np.arange(2**4+1)
-        y = np.cos(0.2*x)
+        x = xp.arange(2**4+1, dtype=xp.float64)
+        y = xp.cos(0.2*x)
         val = romb(y)
```
lucascolley:
```suggestion
        x = np.arange(2**4+1)
        y = np.cos(0.2*x)
        val = romb(xp.asarray(y, dtype=xp.float64))
```

lucascolley (commented):
this might fix the JAX errors

VeckoTheGecko:
Done - just fyi @lucascolley with the env in the Pixi stack I couldn't reproduce the Jax error

```
pixi run test-jax -- scipy/integrate/tests/test_quadrature.py::TestRomb --pdb -vv
```

full output
```
(base) 🦎scipy   HEAD(2178b46) [$] ❯ pixi run test-jax -- scipy/integrate/tests/test_quadrature.py::TestRomb --pdb -vv
✨ Pixi task (build in default): spin build --setup-args=-Dblas=blas --setup-args=-Dlapack=lapack --setup-args=-Duse-g77-abi=true
$ meson compile -j 11 -C build
INFO: autodetecting backend as ninja
INFO: calculating backend command to run: /Users/Hodgs004/coding/euroscipy2025/pixi-dev-scipystack/scipy/.pixi/envs/default/bin/ninja -C /Users/Hodgs004/coding/euroscipy2025/pixi-dev-scipystack/scipy/scipy/build -j 11
ninja: Entering directory `/Users/Hodgs004/coding/euroscipy2025/pixi-dev-scipystack/scipy/scipy/build'
[2/338] Generating subprojects/highs/src/HConfig.h with a custom command

✨ Pixi task (test-jax in jax): spin test --no-build -b jax.numpy -m 'array_api_backends and not slow' -- scipy/integrate/tests/test_quadrature.py::TestRomb --pdb -vv
$ export PYTHONPATH="/Users/Hodgs004/coding/euroscipy2025/pixi-dev-scipystack/scipy/scipy/build-install/usr/lib/python3.12/site-packages"
$ /Users/Hodgs004/coding/euroscipy2025/pixi-dev-scipystack/scipy/.pixi/envs/jax/bin/python3.12 -P -c 'import scipy'
$ cd /Users/Hodgs004/coding/euroscipy2025/pixi-dev-scipystack/scipy/scipy/build-install/usr/lib/python3.12/site-packages
$ /Users/Hodgs004/coding/euroscipy2025/pixi-dev-scipystack/scipy/.pixi/envs/jax/bin/python3.12 -P -m pytest -m 'array_api_backends and not slow' scipy/integrate/tests/test_quadrature.py::TestRomb --pdb -vv
===================================================================== test session starts ======================================================================
platform darwin -- Python 3.12.11, pytest-8.4.1, pluggy-1.6.0 -- /Users/Hodgs004/coding/euroscipy2025/pixi-dev-scipystack/scipy/.pixi/envs/jax/bin/python3.12
cachedir: .pytest_cache
hypothesis profile 'deterministic' -> database=None, deadline=None, print_blob=True, derandomize=True, suppress_health_check=(HealthCheck.data_too_large, HealthCheck.filter_too_much, HealthCheck.too_slow, HealthCheck.large_base_example, HealthCheck.function_scoped_fixture, HealthCheck.differing_executors, HealthCheck.nested_given)
rootdir: /Users/Hodgs004/coding/euroscipy2025/pixi-dev-scipystack/scipy/scipy
configfile: pytest.ini
plugins: xdist-3.8.0, timeout-2.4.0, cov-6.2.1, hypothesis-6.135.32
collected 2 items                                                                                                                                              

scipy/integrate/tests/test_quadrature.py::TestRomb::test_romb[jax.numpy] PASSED                                                                          [ 50%]
scipy/integrate/tests/test_quadrature.py::TestRomb::test_romb_gh_3731[jax.numpy] PASSED   
```

lucascolley:
possible that only `test-jax-cuda` ran into issues

j-bowhay:
I pushed a commit to hopefully fix the last failure which was happening when torch runs with float32 as the default floating dtype

lucascolley (approved):
All green 🎉

The existing tests are pretty minimal, but I don't think we should hold up this PR on that.

Thanks for your first contribution to SciPy @VeckoTheGecko, keep them coming!

lucascolley:
and thanks for the review Jake

PR #23498, opened 2025-08-22
TST: additional tests if `integrate.quad` upper/lower bounds are equal

JostMigenda:
#### Reference issue
Follow-up to gh-23494.

Adds tests for `complex_func=True` as well as for `dblquad` and `tplquad`. Since the latter two already call `nquad` under the hood, I don’t think an additional test for `nquad` is useful. @j-bowhay, let me know if you disagree!


j-bowhay (approved):
Thanks you very much for the speedy follow up! Looks great

Issue #21725, opened 2024-10-19
MAINT: consolidate beta quantile wrappers `special.betaincinv` and `special._ufuncs._beta_ppf`


dschmitz89:
### Is your feature request related to a problem? Please describe.

`special` contains a wrapper of Boosts' `ibeta_inv` that is used for `special.betaincinv`: https://github.com/scipy/scipy/blob/a1ab237a7dfafc3e327005977531f23266dc1bb4/scipy/special/boost_special_functions.h#L160

On top of that, `beta_ppf` wraps this wrapper with a slightly different boost policy so that a Boost error is converted to a Python Warning:
https://github.com/scipy/scipy/blob/a1ab237a7dfafc3e327005977531f23266dc1bb4/scipy/special/boost_special_functions.h#L528

Replacing `_beta_ppf` with `betaincinv` results in one test failure in `stats`:

https://github.com/scipy/scipy/blob/a1ab237a7dfafc3e327005977531f23266dc1bb4/scipy/stats/tests/test_distributions.py#L4884

It boils down to the following different behaviour for extreme inputs between both wrappers:

```python

from scipy.special._ufuncs import _beta_ppf
from scipy.special import betaincinv

q, a, b = 0.995, 1.0e11, 1.0e13

_beta_ppf(q, a, b), betaincinv(q, a, b)

np.float64(0.009901070347411528), np.float64(nan)
```

The `_beta_ppf` result is accurate up to a few digits for a roundtrip comparison: `stats.beta.cdf(0.009901070347411528, a, b)` yields `0.9950000001700496`.

`_beta_ppf` also raises:

`RuntimeWarning: Error in function boost::math::tgamma<d>(%1%,%1%): Series evaluation exceeded %1% iterations, giving up now.`

### Describe the solution you'd like.

Ideally, we would have only one inverse incomplete beta function used in both `stats` and `special`.

Since there is a lot of history involved here, I would like to start a discussion if the `stats` policy could be applied to `special` as well. Or maybe there is another boost policy to increase the number of iterations for `tgamma`.

CC @WarrenWeckesser since a lof of this originates from your PRs I think.
CC @jzmaddock in case you have any points from the boost side here. I think extreme cases for the incomplete beta function are also on your radar from https://github.com/boostorg/math/issues/1198

jzmaddock:
[...]

PR #23499, opened 2025-08-22
ENH: update boost to 1.89

### LINKED ISSUE #22682: BUG: special.betainc: inaccurate/incorrect for small a and b values when using single-precision floats
### Describe your issue.

As in the title.

It is likely `boost::math::ibeta` implementation issue in the region where `a` and `b` parameters are smaller than a certain threshold. See `expected result 2` example below.

### Reproducing Code Example

```python
>>> sp.betainc(np.float32(1e-20), np.float32(1e-21), np.float32(0.5))  # incorrect value
np.float32(0.0)
>>> sp.betainc(1e-20, 1e-21, 0.5)  # expected result 1
np.float64(0.09090909090909091)
>>> sp.betainc(np.float32(1e-15), np.float32(1e-16), np.float32(0.5))  # expected result 2
np.float32(0.090910465)
```

### Error message

```shell
See incorrect results above.
```

<details>

```shell
>>> import sys, scipy, numpy; print(scipy.__version__, numpy.__version__, sys.version_info); scipy.show_config()
1.16.0.dev0+git20250311.52a0913 2.3.0.dev0+git20250311.a651643 sys.version_info(major=3, minor=11, micro=10, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    detection method: pkgconfig
    found: true
    include directory: /opt/_internal/cpython-3.11.10/lib/python3.11/site-packages/scipy_openblas32/include
    lib directory: /opt/_internal/cpython-3.11.10/lib/python3.11/site-packages/scipy_openblas32/lib
    name: scipy-openblas
    pc file directory: /project
    version: 0.3.28
  lapack:
    detection method: pkgconfig
    found: true
    include directory: /opt/_internal/cpython-3.11.10/lib/python3.11/site-packages/scipy_openblas32/include
    lib directory: /opt/_internal/cpython-3.11.10/lib/python3.11/site-packages/scipy_openblas32/lib
    name: scipy-openblas
    pc file directory: /project
    version: 0.3.28
  pybind11:
    detection method: config-tool
    include directory: unknown
    name: pybind11
    version: 2.13.6
Compilers:
  c:
    commands: cc
    linker: ld.bfd
    name: gcc
    version: 10.2.1
  c++:
    commands: c++
    linker: ld.bfd
    name: gcc
    version: 10.2.1
  cython:
    commands: cython
    linker: cython
    name: cython
    version: 3.0.12
  fortran:
    commands: gfortran
    linker: ld.bfd
    name: gcc
    version: 10.2.1
  pythran:
    include directory: ../../tmp/pip-build-env-2wd8u1q0/overlay/lib/python3.11/site-packages/pythran
    version: 0.17.0
Machine Information:
  build:
    cpu: x86_64
    endian: little
    family: x86_64
    system: linux
  cross-compiled: false
  host:
    cpu: x86_64
    endian: little
    family: x86_64
    system: linux
Python Information:
  version: '3.11'
```

</details>

dschmitz89:
#### Reference issue
Closes #22682

Updates boost to 1.89.0. This directly fixes #22682. I also added a new test for the noncentral F CDF which was reported by us to boost (see https://github.com/boostorg/math/issues/1198). As this boost update contains a large update to the incomplete beta function for large arguments, improvements can be expected for many distributions.

#### Additional information
#21725 is fixed in boost but requires some more policy massaging on the scipy side, so I left it for a followup. At least the runtime warning does not appear anymore (see according test).

One thing that bugs me a little is that we still use boost's develop branch although we rely on releases. I will switch to their master branch in a followup.


scipy/special/tests/test_basic.py
```diff
@@ -1491,6 +1491,21 @@ def test_gh21426(self, dtype):
         result = special.betaincinv(a, a, x)
         assert_allclose(result, x, rtol=10 * np.finfo(dtype).eps)
 
+    @pytest.mark.parametrize("dtype, rtol",
+                             [(np.float32, 1e-4),
+                              (np.float64, 1e-15)])
```
dschmitz89, line 1496: For single precision, the test passed only with `1000 * eps`, so I decided to manually set tolerances instead.

doc/source/tutorial/stats/resampling.rst
```diff
@@ -81,7 +81,7 @@ The demon replies:
     >>> from scipy.stats import binom
     >>> prob = binom.cdf(x, n, p)
     >>> print(f"The correct answer is approximately {prob}")
-    The correct answer is approximately 0.18410080866334788
+    The correct answer is approximately 0.18410080866334808
```
dschmitz89, line 84: The correct value acc. to mpmath is `0.18410080866334813`, so the new value is closer.

scipy/special/tests/test_cdflib.py
```diff
@@ -691,7 +693,7 @@ def test_nctdtr_gh8344(self):
 
     @pytest.mark.parametrize(
         "df, nc, x, expected, rtol",
-        [[3., 5., -2., 1.5645373999149622e-09, 5e-9],
+        [[3., 5., -2., 1.5645373999149622e-09, 5e-8],
```
tylerjereddy: Since the other changes in this PR appear to be in the "correct direction," can you explain what's going on here?

I checked out your branch locally on ARM Mac and didn't need the looser tol for it to pass--has something got slightly worse on a particular platform or is this some unrelated flake you're fixing while you're in the neighborhood?

tylerjereddy:
I guess the CI failure that currently shows up is easy to guess the solution for from a quick look at `macos.yml  `:

```
116         # configure system boost
117         mamba install libboost-headers=1.88.0
```

hopefully we could just bump that

scipy/special/tests/test_cdflib.py
dschmitz89: I needed this locally on Linux/x86 for the test to pass. Will roll it back and see what CI says.

dschmitz89:
Okay, the CI test failures [show](https://github.com/scipy/scipy/pull/23499/checks?check_run_id=48738793394) that the noncentral T CDF became slightly less accurate for one test case on x86 CPUs on Linux and Windows (relative accuracy dropped from `5e-9` to `2e-8`. I will report this and our [test case](https://github.com/scipy/scipy/blob/5ab86fb0390ae48a60332c454c12c6a75814afa3/scipy/special/tests/test_cdflib.py#L673) where the noncentral CDF becomes negative to the boost developers . In the meantime, I will bump the test tolerance as this slight regression is not serious and should not block the improvements in other areas.

dschmitz89:
I found out that our boost policy may be responsable for the test case with decreased precision and opened https://github.com/boostorg/math/issues/1308 . This PR can be merged independently though.

dschmitz89:
Allright, https://github.com/boostorg/math/issues/1308 was fixed in boost meaning with the next boost release we will be able to iron out the edge cases for the noncentral T distribution. This PR is ready from my side.

scipy/special/tests/test_cdflib.py
```diff
@@ -691,7 +693,7 @@ def test_nctdtr_gh8344(self):
 
     @pytest.mark.parametrize(
         "df, nc, x, expected, rtol",
-        [[3., 5., -2., 1.5645373999149622e-09, 5e-9],
+        [[3., 5., -2., 1.5645373999149622e-09, 2e-8],
```
rgommers, line 700: It'd be nice to add a comment here saying that this value can be changed back to 5e-9 after we upgrade to boost-math 1.90.0

rgommers (approved):
LGTM, thanks @dschmitz89. One tiny comment about adding a comment for the test precision change. Other than that, go ahead and squash-merge I'd say.

PR #23500, opened 2025-08-23
ENH: optimize.differential_evolution: add Population and Energy History Tracking

PriyadarshiAkshay:
Added class attributes to store the population and energy history across generations. This would be helpful in tracking history and in generating cornerplots of samples.

The ``result`` now includes result.population_history (shape: generations x popsize x parameter size) and result.energy_history (shape: generations x popsize).

### DIFF
--- a/scipy/optimize/_differentialevolution.py
+++ b/scipy/optimize/_differentialevolution.py
@@ -289,6 +289,10 @@ def differential_evolution(func, bounds, args=(), strategy='best1bin',
         ``population`` the solution vectors present in the population, and
         ``population_energies`` the value of the objective function for each
         entry in ``population``.
+        Additional attributes include ``population_history`` which contains
+        the population state at each generation, and ``energy_history`` which
+        contains the corresponding objective function values throughout the
+        optimization process.
         See `OptimizeResult` for a description of other attributes. If `polish`
         was employed, and a lower minimum was obtained by the polishing, then
         OptimizeResult also contains the ``jac`` attribute.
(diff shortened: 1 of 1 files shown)


lucascolley:
thanks @PriyadarshiAkshay . As per https://scipy.github.io/devdocs/dev/hacking.html#contributing-new-code, please could you post on [the forum](https://discuss.scientific-python.org/c/contributor/scipy/32), summarising and motivating the change?

PriyadarshiAkshay:
@lucascolley thank you for the suggestion; I have now created a post.

andyfaff:
Firstly, it's already possible to track the evolution of the `population` and `population_energies` per iteration using the `callback` keyword. The `OptimizeResult: intermediate_result` contains those attributes. It would be simple for the user to construct the entire history in their own callback. e.g.

```
class Callback:
    def __init__(self):
        self.population_history = []

    def __call__(self, intermediate_result):
        self.population_history.append(intermediate_result.population
```

Secondly, there are circumstances where tracking history would lead to large memory usage. Imagine a situation where there are e.g. 1000 parameters. If `popsize=15` that means the total number of population members is 1_000 * 1_000 * 15 = 15_000_000, which is ~120 megabytes. If there are 1000 iterations that would be 120 gigabytes of memory you'd have to find to store the population history. Therefore, it would seem unwise to store the history of the population and its associated energies.

andyfaff:
In future please don't add the extra commits onto a PR that achieves a different purpose. If the later features aren't desired than the commits have to be dropped to allow the wanted first change through.

lucascolley (changes requested):
The memory usage concerns sound particularly problematic.

andyfaff:
Please drop the commits that affect population history, leaving only the docstring changes.

PriyadarshiAkshay:
@andyfaff The solution with the callback parameter is more elegant and serves the purpose. The other changes are already mentioned in a different PR.

PR #23502, opened 2025-08-24
DOC: add header to array api support in docstring

j-bowhay:
towards #23398

Currently array api support is appended to the notes section of the docstring which isn't great for visibility. This PR adds a subheader to separate it from the rest of the notes section

### DIFF
--- a/scipy/_lib/_array_api.py
+++ b/scipy/_lib/_array_api.py
@@ -682,6 +682,8 @@ def _make_capabilities_note(fun_name, capabilities):
         # that explains what is and isn't in-scope, but such a section
         # doesn't exist yet. Using :ref:`dev-arrayapi` as a placeholder.
         note = f"""
+        **Array API Support**
+
         `{fun_name}` is not in-scope for support of Python Array API Standard compatible
         backends other than NumPy.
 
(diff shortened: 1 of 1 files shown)


lucascolley (changes requested):
I'm not convinced this looks great, can we get it to be a proper subheading like `Notes` and `Examples`?

image

j-bowhay:
> I'm not convinced this looks great, can we get it to be a proper subheading like `Notes` and `Examples`?
> 
> image

It doesn't seem possible to have a proper section, xref https://github.com/numpy/numpydoc/issues/610 , but I could try making the text less excessively large

j-bowhay:
Ok I have made it so the text is just bold which I think is a good compromise. I agree long term it would be nice to have a proper section but I think this is a good solution in the meantime to keep the separate from the rest of the notes section

scipy/_lib/_array_api.py
```diff
@@ -682,6 +682,8 @@ def _make_capabilities_note(fun_name, capabilities):
         # that explains what is and isn't in-scope, but such a section
         # doesn't exist yet. Using :ref:`dev-arrayapi` as a placeholder.
         note = f"""
+        **Array API Support**
```
lucascolley:
```suggestion
        **Array API Standard Support**
```

scipy/_lib/_array_api.py
```diff
@@ -691,6 +693,8 @@ def _make_capabilities_note(fun_name, capabilities):
 
     # Note: deliberately not documenting array-api-strict
     note = f"""
+    **Array API Support**
```
lucascolley:
```suggestion
    **Array API Standard Support**
```

lucascolley (approved):
thanks Jake!

PR #23504, opened 2025-08-25
BUG: stats: trapezoid.fit does not converge

### LINKED ISSUE #23503: BUG: stats: `trapezoid.fit()` doesn't converge
### Describe your issue.

`scipy.stats.trapezoid` was implemented in 2016 in issue https://github.com/scipy/scipy/issues/6028. However, `.fit` doesn't converge when no starting value is given for args `c` and `d`.
I've only come across this issue with `scipy.stats.trapezoid`, not with any other `scipy.stats` distributions.
The issue happens both when the default MLE is used and when Method of Moments is used. In both cases it can easily be fixed by specifying some args0.
Following the principle that things should "just work" out of the box in the majority of cases, I propose that sensible default args0 be chosen. Then, when `fit` is called without specifying args0, those default args0 can be used for the MLE minimisation so that it will converge.

Here is the output of the following code which illustrates the issue:

Image

The green line is a fit with no args0, and always results in `c=d=1` i.e. a triangular distribution. This happens for all the cases tried. The red dotted line is using a fit with args0 specified (`c=0.33, d=0.66`), and while not perfect always fits much better, as reflected in the much higher log likelihood.

### Reproducing Code Example

```python
import scipy.stats
import numpy as np

cd = scipy.stats.uniform(0,1).rvs(size=(10,2), random_state=0).round(2)
cd = np.sort(cd, axis=1)

fig, axs = plt.subplots(len(cd), 2, figsize=(10, 2*len(cd)), sharex=True)

starting_cds = [
    (),
    # (.99,),
    # (.99,.99),
    # (.01, .01),
    (.33,.66),
]

methods = [
    "MM", 
    # "MLE",
    ]
x = np.linspace(-.01, 1.01, 103)

for i, (c,d) in enumerate(cd):
    true_dist = scipy.stats.trapezoid(c,d, 0,1)
    rvs = true_dist.rvs(1_000, random_state=0)

    # print("true generating distribution", true_dist.args, -scipy.stats.trapezoid.nnlf(true_dist.args, rvs), sep="   \t")

    axs[i][0].hist(rvs, bins=50, density=True, alpha=.2, label='rvs')
    axs[i][0].plot(x, true_dist.pdf(x), label=f'true pdf. LL={-scipy.stats.trapezoid.nnlf(true_dist.args, rvs):.2f}')

    axs[i][1].hist(rvs, bins=50, density=True, cumulative=True, alpha=.2, label='rvs')
    axs[i][1].plot(x, true_dist.cdf(x), label=f'true cdf. (c,d)={c,d}. LL={-scipy.stats.trapezoid.nnlf(true_dist.args, rvs):.2f}')

    for starting_cd in starting_cds:
        for method in methods:
            fitted_args = scipy.stats.trapezoid.fit(rvs, *starting_cd, floc=0, scale=1, method=method)
            fitted_dist = scipy.stats.trapezoid(*fitted_args)

            # print(starting_cd, method, fitted_dist.args, -scipy.stats.trapezoid.nnlf(fitted_dist.args, rvs), sep="   \t")

            label = f'fitted cdf. Method={method}. Starting (c,d)={starting_cd}. LL={-scipy.stats.trapezoid.nnlf(fitted_dist.args, rvs):.2f}'
            axs[i][0].plot(x, fitted_dist.pdf(x), ls=":", )
            axs[i][1].plot(x, fitted_dist.cdf(x), ls=":", label=label)

    axs[i][1].ecdf(rvs, alpha=.4, label='ecdf')
    # axs[i][0].legend(fontsize='xx-small')
    axs[i][1].legend(fontsize='x-small', loc='center left', bbox_to_anchor=(1, 0.5))

axs[0][0].set_title('pdf')
axs[0][1].set_title('cdf')

plt.tight_layout()
plt.show()
```

### Error message

```shell
na
```

```shell
1.13.0 1.26.4 sys.version_info(major=3, minor=12, micro=3, releaselevel='final', serial=0)
```

SumitkCodes:
Closes #23503

**Problem:** `scipy.stats.trapezoid.fit` fails to converge when no starting values are provided for `c` and `d` parameters, defaulting to triangular distribution (`c=d=1`) with poor log-likelihood.

This PR sets more robust defaults (`c=0.33`, `d=0.66`) to improve convergence while keeping full backward compatibility.  

Follows suggestion from issue report (`kwds.setdefault`)  
Backward compatible
Tested against edge cases and existing validation


nickodell:
Looks like the benchmark job failure is related to this PR:

```
             For parameters: 'trapezoid', 'fit'
               File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/asv_runner/server.py", line 179, in _run_server
                 _run(run_args)
               File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/asv_runner/run.py", line 72, in _run
                 result = benchmark.do_run()
                          ^^^^^^^^^^^^^^^^^^
               File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/asv_runner/benchmarks/_base.py", line 661, in do_run
                 return self.run(*self._current_params)
               File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/asv_runner/benchmarks/time.py", line 165, in run
                 samples, number = self.benchmark_timing(
                                   ^^^^^^^^^^^^^^^^^^^^^^
               File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/asv_runner/benchmarks/time.py", line 289, in benchmark_timing
                 timing = timer.timeit(number)
                          ^^^^^^^^^^^^^^^^^^^^
               File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/timeit.py", line 180, in timeit
                 timing = self.inner(it, self.timer)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^
               File "<timeit-src>", line 6, in inner
               File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/asv_runner/benchmarks/time.py", line 90, in func
                 self.func(*param)
               File "/home/circleci/repo/benchmarks/benchmarks/stats.py", line 279, in time_distribution
                 self.method(*self.args, **self.kwds)
               File "/home/circleci/repo/build-install/usr/lib/python3.11/site-packages/scipy/stats/_continuous_distns.py", line 10007, in fit
                 return super().fit(data, *args, **kwds)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
               File "/home/circleci/repo/build-install/usr/lib/python3.11/site-packages/scipy/stats/_distn_infrastructure.py", line 2764, in fit
                 raise TypeError(f"Unknown arguments: {kwds}.")
             TypeError: Unknown arguments: {'c': 0.33, 'd': 0.66}.
             
```

You should be able to reproduce this locally by running `spin bench -t stats.DistributionsAll` to run just that benchmark.

SumitkCodes:
@nickodell can you check the changes now

nickodell:
I don't feel comfortable reviewing changes against scipy.stats, but I've tagged two people who are listed in CODEOWNERS / have reviewed similar contributions before.

Three comments:

1. In the associated issue, @Gabriel-Kissin suggested using `_fitstart` instead, rather than overriding `fit()` to add this.  I think that's a good suggestion.
2. Can you make sure you look at [the page in the docs](https://docs.scipy.org/doc/scipy/dev/contributor/development_workflow.html#writing-the-commit-message) on writing good commit messages?
3. This change lacks a test. In general it is good for bug fixes to have tests to verify that they fix the bug in question and that future changes don't re-introduce the same bug while refactoring. See [Checklist before submitting a PR](https://docs.scipy.org/doc/scipy/dev/contributor/development_workflow.html#checklist-before-submitting-a-pr).

SumitkCodes:
@nickodell Thanks for the feedback, Should I start implementing these changes now, or wait for further guidance?

nickodell:
>Thanks for the feedback, Should I start implementing these changes now, or wait for further guidance?

Start now. The reviewer may have further requests for you, but they'll certainly want the things I mentioned.

tylerjereddy (commented):
One thing that may help re: Nick's comment about writing a regression test is that I believe we already have one for both MLE and MM. If you apply the patch below the fold on `main`, to remove the `xfail`s, it will fail both MM and MLE tests for `trapezoid` with `fit not very good in trapezoid`.

<details>

```diff
--- a/scipy/stats/tests/test_fit.py
+++ b/scipy/stats/tests/test_fit.py
@@ -36,7 +36,6 @@ mle_failing_fits = [
         'tukeylambda',
         'vonmises',
         'levy_stable',
-        'trapezoid',
         'truncweibull_min',
         'studentized_range',
 ]
@@ -71,7 +70,7 @@ mm_failing_fits = ['alpha', 'betaprime', 'burr', 'burr12', 'cauchy', 'chi',
                    'johnsonsu', 'kappa3', 'ksone', 'kstwo', 'landau', 'levy', 'levy_l',
                    'levy_stable', 'loglaplace', 'lomax', 'mielke', 'nakagami',
                    'ncf', 'nct', 'ncx2', 'pareto', 'powerlognorm', 'powernorm',
-                   'rel_breitwigner', 'skewcauchy', 't', 'trapezoid', 'triang',
+                   'rel_breitwigner', 'skewcauchy', 't', 'triang',
                    'truncpareto', 'truncweibull_min', 'tukeylambda',
                    'studentized_range']
```

</details>

Both `trapezoid` cases pass on this feature branch, so I suppose that may be a good sign.

One thing that gives me a little pause is this statement in that test, which clearly indicates that folks were aware of this for some time:

```python
 115         # The generic `fit` method can't be expected to work perfectly for all              
 116         # distributions, data, and guesses. Some failures are expected.   
```

I'm not commenting on backwards compatibility or anything else at this time--just pointing this out in case it helps for you/the reviewers.

Incantation: `spin test -m full scipy/stats/tests/test_fit.py::test_cont_fit`

SumitkCodes:
@nickodell I've implemented all three feedback points:
1.  Changed from `fit()` override to `_fitstart` override as suggested by @Gabriel-Kissin
2. Updated commit message to follow SciPy guidelines 
3. Added comprehensive tests to prevent regression
The fix now uses `_fitstart` to provide sensible defaults (c=0.33, d=0.66) without polluting the `fit()` method. All tests pass and backward compatibility is maintained.

let me know if it need more work

andyfaff:
@mdhaber, does it make sense to start adding more complexity to the fitting mechanism here? There are several situations where using the following could make sense:

- one shape parameter should be less than the other (`LinearConstraint`/`NonlinearConstraint`)
- where shape parameters should remain within limits (`Bounds`)

Or would effort be better directed towards the new `ContinuousDistribution` infrastructure?

scipy/stats/_continuous_distns.py
```diff
@@ -9994,6 +9994,15 @@ def _entropy(self, c, d):
         # the following result.
         return 0.5 * (1.0-d+c) / (1.0+d-c) + np.log(0.5 * (1.0+d-c))
 
+    def _fitstart(self, data):
+        """Override _fitstart to provide sensible default starting values for c and d.
+        
+        The default starting values c=0.33, d=0.66 provide better convergence
+        than the generic defaults, especially for trapezoidal distributions
+        that are not close to triangular.
+        """
```
nickodell: This docstring isn't correct. The problem is not only with non-triangular data, but with any data, because the points in the initial simplex become c=1.05 and d=1.05., which are not valid parameters to the triangular distribution.

Additionally, I would suggest a shorter comment. The reason I would suggest that is that the comment's length should be proportional to how complex / hairy the associated code is. Proposed wording:

```suggestion
        # Arbitrary, but c=d=1 fails due to being on edge of bounds
```

scipy/stats/tests/test_trapezoid_fix.py
nickodell, line 1: This test file won't run automatically when someone runs `spin tests`. (This is what the error inside "[mypy (py3.11) & dev deps (py3.14), fast, spin (3.11, false)](https://github.com/scipy/scipy/actions/runs/17238393146/job/48975622181?pr=23504#logs)" is about.)

Although that can be fixed by adding the file to the `meson.build`, generally that should be avoided.

Instead, the test should go in an existing file. `scipy/stats/tests/test_continuous.py` would be the file I would suggest, as that contains tests for other continuous distributions.

For that reason, entrypoint `if __name__ == "__main__":` should also be removed.

nickodell: PS: I would suggest that you try Tyler's suggestion about using the existing fit() test. I think that would be less work than writing your own tests, and it is more meaningful as a reviewer, as it is an independent point of confirmation that the behavior is desirable.

scipy/stats/tests/test_trapezoid_fix.py
```diff
@@ -0,0 +1,85 @@
...
+    fitted_c, fitted_d = fitted_args[0], fitted_args[1]
+    
+    # Should not converge to triangular distribution (c=d=1)
+    assert not (abs(fitted_c - 1.0) < 0.01 and abs(fitted_d - 1.0) < 0.01), \
+        "fit() still converging to triangular distribution (c=d=1)"
+    
+    # Should converge to reasonable values
+    assert 0 <= fitted_c <= 1, f"fitted c={fitted_c} not in [0,1]"
+    assert 0 <= fitted_d <= 1, f"fitted d={fitted_d} not in [0,1]"
+    assert fitted_d >= fitted_c, f"fitted d={fitted_d} < c={fitted_c}"
```
nickodell: Can you make this test more rigorous by testing that the values are an expected value, instead of testing the properties that the values ought to have? I would suggest `numpy.testing.assert_allclose()` with an appropriate tolerance.

Given that this uses a seeded RNG and a deterministic optimizer, that should be fairly reliable.

scipy/stats/tests/test_trapezoid_fix.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def test_trapezoid_fit_with_explicit_values():
```
nickodell: I don't think you need `test_trapezoid_fit_with_explicit_values`, `test_trapezoid_fit_methods` and `test_trapezoid_fit_edge_cases`. This is testing argument parsing code that you didn't change, and is likely covered by existing tests already.

scipy/stats/tests/test_trapezoid_fix.py
nickodell: On a related note, the test will likely be more reliable if you use both floc and fscale.

Here loc is forced to be 0, and scale is set initially to 1 but allowed to vary.

```
fitted_args = scipy.stats.trapezoid.fit(rvs, floc=0, scale=1)
```

Gabriel-Kissin:
> @mdhaber, does it make sense to start adding more complexity to the fitting mechanism here? There are several situations where using the following could make sense:
> 
> * one shape parameter should be less than the other (`LinearConstraint`/`NonlinearConstraint`)
> * where shape parameters should remain within limits (`Bounds`)
> 
> Or would effort be better directed towards the new `ContinuousDistribution` infrastructure?

There is discussion in this thread about generally giving bounds and constraints to the solver. I thought I'd move that into a separate issue https://github.com/scipy/scipy/issues/23515, to avoid mixing the discussion about trapezoid with that

nickodell:
@SumitkCodes 

> TST: stats: trapezoid.fit: improve test rigor and remove redundant tests

By the way, this commit doesn't remove the [scipy/stats/tests/test_trapezoid_fix.py](https://github.com/scipy/scipy/pull/23504/files/322fcff4153c6b98d16f6fce49591e6dd787c0a9#diff-b6eb3f0d899b637774141c6392df178147f046c08552d874b1805c7ad3a76279) file, so the test "[mypy (py3.11) & dev deps (py3.14), fast, spin (3.11, false)](https://github.com/scipy/scipy/actions/runs/17238393146/job/48975622181?pr=23504#logs)" is still going to fail.

You may have deleted the file from your local repository without committing that deletion to Git. [Here is a page which explains how to delete a file in Git.](https://stackoverflow.com/questions/2047465/how-do-i-delete-a-file-from-a-git-repository)

Can fix this by deleting this file?

SumitkCodes:
@nickodell Thanks for catching that! I've now properly removed the `test_trapezoid_fix.py` file from Git  and committed the deletion. The file should no longer cause test failures.

scipy/stats/tests/test_distributions.py
```diff
@@ -7955,6 +7955,30 @@ def test_trapezoid_vect(self):
 
         assert_allclose(v, res.T.reshape(v.shape), atol=1e-15)
 
+    def test_trapezoid_fit_convergence_gh23503(self):
```
mdhaber, line 7958: `test_continuous` is not well named, but it is just for the new distribution infrastructure. `test_distribution.py` (here) or maybe `test_fit` are more appropriate places for an (old) distribution-specific fit test.

As mentioned above, we do not necessarily need to add a distribution-specific test here because we have generic `fit` tests that run for all distributions. I've allowed these tests to run on `trapezoid` again (https://github.com/scipy/scipy/pull/23504#discussion_r2347685718) but I went ahead and kept a distribution-specific test because some of those tests will be too slow to always run in CI.

scipy/stats/tests/test_distributions.py
```diff
@@ -7955,6 +7955,30 @@ def test_trapezoid_vect(self):
...
+
+        # Should not converge to triangular distribution (c=d=1)
+        fitted_c, fitted_d = fitted_args[:2]
+        assert not np.allclose(fitted_c, 1, atol=0.1)
+        assert not np.allclose(fitted_d, 1, atol=0.1)
+
+        # objective function is better than with true values of parameters
+        true_llf = stats.trapezoid.nnlf(true_args, x)
+        fitted_llf = stats.trapezoid.nnlf(fitted_args, x)
+        assert fitted_llf < true_llf
```
mdhaber, line 7980: The job of MLE is to optimize the LLF, which is not necessarily the same as matching the true distribution's parameters. Here, we check that the objective function with the fitted parameters is better than the objective function with the true parameters.

scipy/stats/_continuous_distns.py
```diff
@@ -9994,6 +9994,12 @@ def _entropy(self, c, d):
         # the following result.
         return 0.5 * (1.0-d+c) / (1.0+d-c) + np.log(0.5 * (1.0+d-c))
 
+    def _fitstart(self, data, args=None):
+        # Arbitrary, but c=d=1 fails due to being on edge of bounds
+        if args is None:
```
mdhaber, line 9999: Is it possible that the user passed only one of `c` or `d` but not both? (In that case, what happens?)

scipy/stats/tests/test_continuous_basic.py
mdhaber, line 1: Here and in `test_fit`, I'm removing `trapezoid` from all skip/xfail lists so that it runs generic tests in CI. It is quite slow, so I will add it back to some of the `xslow` lists.

mdhaber (approved):
@SumitkCodes I think the approach is fine. (We could probably spend time to come up with better guesses, but this is clearly better than `c=d=1`.) Please see my adjustments to the tests. If this still looks good to you or @nickodell, I'll go ahead and add `trapezoid` back to some slow test lists (so it won't always run in CI) and merge.

*Note that the CI failures are unrelated; they can be seen in other PRs right now.*

scipy/stats/_continuous_distns.py
mdhaber: @SumitkCodes Can you address this? I think this is the thing preventing merge.

nickodell:
@mdhaber Sorry about my slow response. Could you provide more information about what I should look for in a review on this PR?

1. I looked at my previous review comments. All of them have been addressed.
2. I also looked at the changes you made to the tests. Those look reasonable to me and implement the suggestions that Tyler and I made.
3. The tests pass, with the exception of an issue being tracked as #23689.

Is there anything beyond those items that I should look at? If there's nothing else I should look at I'm happy to approve.

**Edit**: Oh, you said what I should look for two weeks ago. Nvm.

nickodell (approved):

mdhaber:
Thanks very much @nickodell! I think I'll just check out the issue with https://github.com/scipy/scipy/pull/23504#discussion_r2347685402, ~~fix if needed~~, then merge. Thanks @SumitkCodes.

scipy/stats/_continuous_distns.py
mdhaber: Never mind. Turns out the infrastructure ignores parameter guesses unless both are passed and never calls this function with `args`.

PR #19260, opened 2023-09-19
ENH: linalg: support array API for standard extension functions

lucascolley:
_For context, this work was started as part of my internship at [Quansight Labs](https://github.com/Quansight-Labs), which ran until the end of September 2023._
#### Reference issue
Towards gh-19068 and gh-18867.
Please see gh-19068 for context.

#### What does this implement/fix?
Support is added for the functions in the [array API standard `linalg` extension](https://data-apis.org/array-api/latest/extensions/linear_algebra_functions.html). This allows users to input arrays from any compatible array library.

Tests are modified to allow testing with `numpy.array_api`, `cupy` and `torch`. Some new tests are added for exceptions for unsupported parameters.

#### Additional information
I was not able to convert every relevant test since lots of NumPy-specifc things are used. We may want to find ways to convert some of these, or write new tests to serve the same purpose.

`TestSVD` is a little strange due to the way the `lapack_driver` parameter is tested. I have tried to apply a minimal refactor here, but a more substantial refactor may result in something more readable. It is a bit of a misnomer that all of our array API compatible tests are under `TestSVD_GESDD`, just because `gesdd` is the default value.

Lots of tests are failing for PyTorch CUDA, but hopefully we just need to wait for https://github.com/pytorch/pytorch/issues/106773 to come through.

Issue #19068, opened 2023-08-15
ENH: linalg: array library interoperability


lucascolley:
### Is your feature request related to a problem? Please describe.

**Plans for addressing https://github.com/scipy/scipy/issues/18286 in `scipy.linalg`.**

Currently, `scipy.linalg` only supports the use of NumPy arrays. Work towards interoperability has the potential to bring large performance improvements now and, with adoption of the [array API standard](https://data-apis.org/array-api/latest/index.html), will bring compatibility with any future library which complies with the standard.

`linalg` is an [array API standard extension](https://data-apis.org/array-api/latest/extensions/linear_algebra_functions.html), which means that supporting it will bring automatic dispatching for all standard functions. [`array-api-compat`](https://github.com/data-apis/array-api-compat) has implemented `linalg` wrappers for `numpy`, `cupy` and `torch`, so that automatic dispatching is available today.

For those functions which use compiled code and are not in the standard, non-CPU device support will be limited to start off with. From the RFC:

> When a non-NumPy array type sees compiled code in SciPy (which tends to use the NumPy C API), we have a couple of options:
> 
> 1. dispatch back to the other library (PyTorch, CuPy, etc.).
> 2. convert to a NumPy array when possible (i.e., on CPU via the buffer protocol, DLPack, or __array__), use the compiled code in question, then convert back.
> 3. don't support this at all
>
> I'll note that (1) is the long-term goal; how to implement this is out of scope for this proposal - for more on that, see the Appendix section. For now we choose to do (2) when possible, and (3) otherwise. Switching from that approach to (1) in the future will be backwards-compatible.

This does not mean that GPU support is off-limits for all of these functions right now. Firstly, if we can write non-standard functions in terms of standard functions, then they will become compatible with all compliant libraries. Secondly, we have the option to call to CuPy and PyTorch for some functions. While this is not the general, library-agnostic solution of (1) - the "long-term goal" - it may be worth doing this to unlock GPU support for some important functions.

#### A. Conform to the standard extension.

This should be (almost) possible in one PR, similar to https://github.com/scipy/scipy/pull/19005 in `fft`. This involves validating array inputs with `array_namespace` and using `xp.linalg`, if it is available, for all standard functions. If it is not available, a conversion to NumPy is attempted. For NumPy arrays, the existing SciPy implementation is used.

[...]

PR #23507, opened 2025-08-25
ENH: add array api support for `linalg.orthogonal_procrustes` and `_apply_over_batch`

j-bowhay:
Adds array api support to `linalg.orthogonal_procrustes` since this is a pretty straightforward function. Also converts the utility function `_apply_over_batch`

### DIFF
--- a/scipy/linalg/_procrustes.py
+++ b/scipy/linalg/_procrustes.py
@@ -11,7 +11,7 @@ from scipy._lib._array_api import array_namespace, xp_capabilities, _asarray, is
 __all__ = ['orthogonal_procrustes']
 
 
-@xp_capabilities(jax_jit=False, allow_dask_compute=1)
+@xp_capabilities(jax_jit=False, allow_dask_compute=2)
 @_apply_over_batch(('A', 2), ('B', 2))
 def orthogonal_procrustes(A, B, check_finite=True):
     """
--- a/scipy/linalg/_procrustes.py
+++ b/scipy/linalg/_procrustes.py
@@ -11,7 +11,7 @@ from scipy._lib._array_api import array_namespace, xp_capabilities, _asarray, is
 __all__ = ['orthogonal_procrustes']
 
 
-@xp_capabilities(jax_jit=False)
+@xp_capabilities(jax_jit=False, allow_dask_compute=1)
 @_apply_over_batch(('A', 2), ('B', 2))
 def orthogonal_procrustes(A, B, check_finite=True):
     """
--- a/scipy/linalg/_procrustes.py
+++ b/scipy/linalg/_procrustes.py
@@ -2,9 +2,10 @@
 Solve the orthogonal Procrustes problem.
 
 """
-import numpy as np
+
 from scipy._lib._util import _apply_over_batch
-from scipy._lib._array_api import array_namespace, xp_capabilities, _asarray
+from ._decomp_svd import svd
+from scipy._lib._array_api import array_namespace, xp_capabilities, _asarray, is_numpy
 
 
 __all__ = ['orthogonal_procrustes']
@@ -51,8 +52,6 @@ def orthogonal_procrustes(A, B, check_finite=True):
     Note that unlike higher level Procrustes analyses of spatial data, this
     function only uses orthogonal transformations like rotations and
     reflections, and it does not use scaling or translation.
-
-    .. versionadded:: 0.15.0
 
     References
     ----------
@@ -104,11 +103,14 @@ def orthogonal_procrustes(A, B, check_finite=True):
         raise ValueError(f'expected ndim to be 2, but observed {A.ndim}')
     if A.shape != B.shape:
         raise ValueError(f'the shapes of A and B differ ({A.shape} vs {B.shape})')
-    breakpoint()
+
     # Be clever with transposes, with the intention to save memory.
     # The conjugate has no effect for real inputs, but gives the correct solution
     # for complex inputs.
-    u, w, vt = xp.linalg.svd((B.T @ xp.conj(A)).T)
+    if is_numpy(xp):
+        u, w, vt = svd((B.T @ xp.conj(A)).T)
+    else:
+        u, w, vt = xp.linalg.svd((B.T @ xp.conj(A)).T)
     R = u @ vt
-    scale = w.sum()
+    scale = xp.sum(w)
     return R, scale
--- a/scipy/linalg/_procrustes.py
+++ b/scipy/linalg/_procrustes.py
@@ -4,12 +4,13 @@ Solve the orthogonal Procrustes problem.
 """
 import numpy as np
 from scipy._lib._util import _apply_over_batch
-from ._decomp_svd import svd
+from scipy._lib._array_api import array_namespace, xp_capabilities, _asarray
 
 
 __all__ = ['orthogonal_procrustes']
 
 
+@xp_capabilities()
 @_apply_over_batch(('A', 2), ('B', 2))
 def orthogonal_procrustes(A, B, check_finite=True):
     """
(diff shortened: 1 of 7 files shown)


scipy/linalg/_procrustes.py
```diff
@@ -2,14 +2,16 @@
...
+
 from scipy._lib._util import _apply_over_batch
 from ._decomp_svd import svd
+from scipy._lib._array_api import array_namespace, xp_capabilities, _asarray, is_numpy
 
 
 __all__ = ['orthogonal_procrustes']
 
 
+@xp_capabilities()
```
j-bowhay:
```suggestion
@xp_capabilities(jax_jit=False)
```

scipy/linalg/_procrustes.py
```diff
@@ -2,14 +2,16 @@
...
+
 from scipy._lib._util import _apply_over_batch
 from ._decomp_svd import svd
+from scipy._lib._array_api import array_namespace, xp_capabilities, _asarray, is_numpy
 
 
 __all__ = ['orthogonal_procrustes']
 
 
+@xp_capabilities(jax_jit=False)
```
j-bowhay:
```suggestion
@xp_capabilities(jax_jit=False, allow_dask_compute=1)
```

scipy/linalg/_procrustes.py
```diff
@@ -2,14 +2,16 @@
...
+
 from scipy._lib._util import _apply_over_batch
 from ._decomp_svd import svd
+from scipy._lib._array_api import array_namespace, xp_capabilities, _asarray, is_numpy
 
 
 __all__ = ['orthogonal_procrustes']
 
 
+@xp_capabilities(jax_jit=False, allow_dask_compute=1)
```
j-bowhay:
```suggestion
@xp_capabilities(jax_jit=False, allow_dask_compute=2)
```

scipy/linalg/_procrustes.py
```diff
@@ -94,20 +97,23 @@ def orthogonal_procrustes(A, B, check_finite=True):
...
         raise ValueError(f'the shapes of A and B differ ({A.shape} vs {B.shape})')
+
     # Be clever with transposes, with the intention to save memory.
     # The conjugate has no effect for real inputs, but gives the correct solution
     # for complex inputs.
-    u, w, vt = svd((B.T @ np.conjugate(A)).T)
+    if is_numpy(xp):
+        u, w, vt = svd((B.T @ xp.conj(A)).T)
+    else:
+        u, w, vt = xp.linalg.svd((B.T @ xp.conj(A)).T)
```
ev-br, line 116: This `if is_numpy` branching looks like an antipattern on a first glance. On a reflection though, this function looks like one of exceptions where this pattern is fine. 
`scipy.linalg.svd` may differ from the Array API standard `svd` function in two ways: additional features (`compute_uv`) and additional performance or stability improvements (`lapack_driver`). Here in `orthogonal_procrustes` we are not using the former but may benefit from the latter, now or in the future.
Therefore, the pattern used here allows to both unlock large speed-ups on e.g. CUDA via `xp.linalg` and keep scipy-specific performance/stability improvements for numpy/cpu. 

Given a long discussion in https://github.com/scipy/scipy/issues/19068, I believe that using an explicit `if is_numpy` branching on a case by case basis is an acceptable pragmatic choice, and `orthogonal_procrustes` is one place where it's fine, at least until we agree on a scipy.linalg wide course of action in gh-19068.

scipy/linalg/tests/test_basic.py
```diff
@@ -2326,7 +2327,7 @@ def test_empty(self, dt_c, dt_b):
 
 
 class TestMatrix_Balance:
-
+    @skip_xp_invalid_arg
```
ev-br, line 2330: Note to self: this is a fancy way to say "when Array API mode is on, the error is different, so skip the test".

scipy/linalg/tests/test_procrustes.py
```diff
@@ -2,213 +2,209 @@
...
 import pytest
-from numpy.testing import assert_array_less, assert_allclose
+from numpy.testing import assert_allclose
 from pytest import raises as assert_raises
 
-from scipy.linalg import inv, eigh, norm, svd
 from scipy.linalg import orthogonal_procrustes
 from scipy.sparse._sputils import matrix
+from scipy._lib._array_api import make_xp_test_case
+from scipy._lib._array_api_no_0d import xp_assert_close
```
ev-br: Question: why is this needed? IIUC the `_no_0d` version is needed to for functions which emulate numpy reductions and return python floats instead of 0D arrays. Does something in linalg do this?

ev-br (approved):
LGTM modulo a small optional question about testing

scipy/linalg/_procrustes.py
j-bowhay: Yes exactly, my reasoning was that scipy uses `gesdd` whereas numpy uses `gesvd` so I didn't want to accidentally slow things down for users with big matrices

scipy/linalg/tests/test_procrustes.py
j-bowhay: I will double check

j-bowhay: Although minor point I think the distinction is for numpy scalars rather than python floats

scipy/linalg/tests/test_basic.py
j-bowhay: Yes needed because a string is not a valid input to `array_namespace`

scipy/linalg/_procrustes.py
lucascolley: yes indeed, this was the pattern that seemed appropriate to me when working on gh-19260

scipy/linalg/tests/test_basic.py
lucascolley: yeah, I think the best paraphrase is "this is testing behaviour that we would not support / deprecate were we to make `SCIPY_ARRAY_API=1` default". It is for things that work without `SCIPY_ARRAY_API` but that we probably plan to leave in the past.

lucascolley: not necessarily that we wouldn't support the functionality, but that we would at least find a different way to support it

scipy/linalg/tests/test_procrustes.py
j-bowhay: Ok the numpy scalar was coming from within the test itself not `orthogonal_procrustes`. Updated to handle this in a better way in the latest commit

scipy/linalg/_procrustes.py
```diff
@@ -2,14 +2,19 @@
...
 from ._decomp_svd import svd
+from scipy._lib._array_api import array_namespace, xp_capabilities, _asarray, is_numpy
 
 
 __all__ = ['orthogonal_procrustes']
 
 
+@xp_capabilities(
+    jax_jit=False,
+    skip_backends=[("dask.array", "full_matrics=True is not supported by dask")],
```
lucascolley: I assume this? But to which function is `full_matric{e}s` a parameter?
```suggestion
    skip_backends=[("dask.array", "full_matrices=True is not supported by dask")],
```

scipy/linalg/_procrustes.py
lucascolley, line 116: what about this?

```python
if is_numpy(xp):
    ...
elif hasattr(xp, "linalg"}:
    ...
else: # try to fallback to NumPy
```

lucascolley: I would be in favour of adding this now, but I won't block on it.

A bridge we can definitely save crossing until we come to it is making `xp_capabilities` and thus the docs aware of which extensions are required (e.g. "GPU supported[^1] [^1]: warning: only if `xp.linalg` exists)

scipy/linalg/tests/test_procrustes.py
```diff
@@ -2,213 +2,208 @@
...
+        B_orig = xp.asarray([[3, 40], [1, 0], [3, -40], [5, 0]], dtype=xp.float64)
+        A, A_mu = _centered(A_orig, xp)
+        B, B_mu = _centered(B_orig, xp)
+        R, s = orthogonal_procrustes(A, B)
+        scale = s / xp.linalg.matrix_norm(A)**2
+        B_approx = scale * A @ R + B_mu
+        expected = xp.asarray([[3, 21], [-18, 0], [3, -21], [24, 0]], dtype=xp.float64)
+        xp_assert_close(B_approx, expected, atol=1e-8)
+        # Check disparity symmetry.
+        expected_disparity = xp.asarray(0.4501246882793018, dtype=xp.float64)[()]
```
lucascolley, line 142: Non-blocking: a comment might be nice, optionally also saving the `[()]` until the matching coercion from NumPy (I assume line 148?) Something like

```python
A_approx = scale * B @ R + A_mu
expected_disparity = expected_disparity[()] # match NumPy coercion to scalar
```

lucascolley (commented):
thanks both!

scipy/linalg/_procrustes.py
j-bowhay: image

https://data-apis.org/array-api/latest/extensions/generated/array_api.linalg.svd.html#array_api.linalg.svd

scipy/linalg/_procrustes.py
j-bowhay: I have to admit I feel slightly reluctant to make the code more complicated... It's already slightly annoying we have to have the `is_numpy` branch.

It doesn't seem that unreasonable to me to say that if you want to do linear algebra your library of choice should have `xp.linalg`.

lucascolley: sure 👍let's leave this comment unresolved but merge

lucascolley (approved):

PR #23511, opened 2025-08-26
DOC: interpolate: Ndbspline derivative methods list

ev-br:
A follow-up to https://github.com/scipy/scipy/pull/23375 :

- add `.derivative` to the `Methods` listing
- remove the default `nu=1` since we now require that `nu` is an array-like of the correct length
- minor tweaks of  wording in the docstring and error messages.

### DIFF
--- a/scipy/interpolate/_ndbspline.py
+++ b/scipy/interpolate/_ndbspline.py
@@ -277,19 +277,20 @@ class NdBSpline:
 
         return new_c, new_t
 
-    def derivative(self, nu=1):
+    def derivative(self, nu):
         """
-        Construct a new NdBSpline representing the derivative.
+        Construct a new NdBSpline representing the partial derivative.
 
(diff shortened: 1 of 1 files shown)


czgdp1807 (approved):

scipy/interpolate/_ndbspline.py
```diff
@@ -299,7 +301,7 @@ def derivative(self, nu=1):
                 f"ndim = {len(self.t)}.")
 
         if any(n < 0 for n in nu_arr):
```
j-bowhay: Sorry not part of your changes but is this not just
```suggestion
        if any(nu_arr < 0):
```

tylerjereddy (approved):
This looks like some straightforward follow-up tweaks to a feature merged in the last day or so, with CI passing, let's put it in.

PR #23512, opened 2025-08-26
MAINT: remove unused cdflib code

dschmitz89:
#### Reference issue
Towards #21966 

This deletes unused code and hopefully shaves off a few KB from our binary size already as a few cython functions are removed.

In fact, several of the C functions were not used by `special` as ufuncs at all even before I started #21966. Might be leftovers from history, who knows...


rgommers (approved):
Great! Thanks @dschmitz89

PR #23513, opened 2025-08-26
ENH: integrate.cumulative_simpson: add array api support

j-bowhay:
towards #20930

### DIFF
--- a/scipy/integrate/tests/test_quadrature.py
+++ b/scipy/integrate/tests/test_quadrature.py
@@ -548,6 +548,7 @@ class TestCumulativeSimpson:
             # only even-interval terms are "exact"
             xp_assert_close(res[i0::2], ref[i0::2], rtol=2e-15)
 
+    @skip_xp_backends(cpu_only=True)  # uses np.apply_along_axis
     @pytest.mark.parametrize('axis', np.arange(-3, 3))
     @pytest.mark.parametrize('x_ndim', (1, 3))
     @pytest.mark.parametrize('x_len', (1, 2, 7))
--- a/scipy/integrate/tests/test_quadrature.py
+++ b/scipy/integrate/tests/test_quadrature.py
@@ -469,7 +469,7 @@ class TestQMCQuad:
         assert_allclose(res.integral, ref, 1e-2)
 
 
-def cumulative_simpson_nd_reference(y, *, x=None, dx=None, initial=None, axis=-1):
+def cumulative_simpson_nd_ref(y, *, x=None, dx=None, initial=None, axis=-1):
     # Use cumulative_trapezoid if length of y < 3
     if y.shape[axis] < 3:
         if initial is None:
@@ -513,7 +513,7 @@ class TestCumulativeSimpson:
 
     @pytest.mark.parametrize('use_dx', (False, True))
     @pytest.mark.parametrize('use_initial', (False, True))
-    def test_1d(self, use_dx, use_initial):
+    def test_1d(self, use_dx, use_initial, xp):
         # Test for exact agreement with polynomial of highest
         # possible order (3 if `dx` is constant, 2 otherwise).
         rng = np.random.default_rng(82456839535679456794)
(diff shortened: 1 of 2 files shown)


scipy/integrate/tests/test_quadrature.py
```diff
@@ -469,7 +469,7 @@ def func(x):
         assert_allclose(res.integral, ref, 1e-2)
 
 
-def cumulative_simpson_nd_reference(y, *, x=None, dx=None, initial=None, axis=-1):
+def cumulative_simpson_nd_ref(y, *, x=None, dx=None, initial=None, axis=-1):
```
tylerjereddy: Did you just change the function name for the sole purpose of reducing the line length in the one case it is used below?

scipy/integrate/tests/test_quadrature.py
```diff
@@ -587,17 +598,20 @@ def test_nd(self, axis, x_ndim, x_len, i_ndim, dx):
...
-    def test_special_cases(self):
+    def test_special_cases(self, xp):
         # Test special cases not checked elsewhere
         rng = np.random.default_rng(82456839535679456794)
-        y = rng.random(size=10)
-        res = cumulative_simpson(y, dx=0)
-        assert_equal(res, 0)
+        y = xp.asarray(rng.random(size=10))
+        res = cumulative_simpson(y, dx=0.)
+        xp_assert_equal(res, 0*res)
```
tylerjereddy: Maybe a bit pedantic, but do we really need to include the `actual` value in the `expected` value as well? This seems undesirable on some levels for avoiding weird issues, and also just feels a bit like cross-contamination into the expected value (yeah it is multiplied by 0, but weird things can happen...).

scipy/integrate/tests/test_quadrature.py
```diff
@@ -563,20 +566,28 @@ def test_nd(self, axis, x_ndim, x_len, i_ndim, dx):
...
-        initial = None if i_ndim is None else rng.random(size=i_shape)
+            x = xp.asarray(x)
+        initial = None if i_ndim is None else xp.asarray(rng.random(size=i_shape))
 
         # compare results
         res = cumulative_simpson(y, x=x, dx=dx, initial=initial, axis=axis)
-        ref = cumulative_simpson_nd_reference(y, x=x, dx=dx, initial=initial, axis=axis)
-        np.testing.assert_allclose(res, ref, rtol=1e-15)
+        ref = cumulative_simpson_nd_ref(
+            np.asarray(y), x=np.asarray(x), dx=None if dx is None else np.asarray(dx),
```
tylerjereddy, line 585: Should there be a comment explaining why we are using `np.` here and `xp.` above? This seems like just partial support being added/partially converted test? `cpu_only` on its own doesn't seem sufficient

tylerjereddy (commented):
The diff mostly looks "ok" to me. I note that reverting the changes to the source file actually doesn't cause any tests to fail via `spin test scipy/integrate/tests/test_quadrature.py -b all` locally with the backends I have available. I suppose that's a result of how our skipping works.

On this branch, that incantation produces `777 passed, 3 skipped, 6 deselected` while after reverting the changes to the source file only I see: `319 passed, 461 skipped, 6 deselected`.

scipy/integrate/tests/test_quadrature.py
j-bowhay: Yes although in hindsight I don't no why because it makes no difference due to the formatting I finally settled on

scipy/integrate/tests/test_quadrature.py
```diff
@@ -563,20 +566,28 @@ def test_nd(self, axis, x_ndim, x_len, i_ndim, dx):
...
+            dx = xp.asarray(dx)
         else:
             x = (np.sort(rng.random(size=shape), axis=axis) if x_ndim > 1
                  else np.sort(rng.random(size=shape[axis])))
-        initial = None if i_ndim is None else rng.random(size=i_shape)
+            x = xp.asarray(x)
+        initial = None if i_ndim is None else xp.asarray(rng.random(size=i_shape))
 
         # compare results
         res = cumulative_simpson(y, x=x, dx=dx, initial=initial, axis=axis)
```
j-bowhay, line 581:
```suggestion
        res = cumulative_simpson(y, x=x, dx=dx, initial=initial, axis=axis)
        # use np to generate `ref` as `cumulative_simpson_nd_ref`
        # uses `apply_along_axis`
```

scipy/integrate/tests/test_quadrature.py
j-bowhay: comment added

scipy/integrate/tests/test_quadrature.py
j-bowhay: done

scipy/integrate/tests/test_quadrature.py
j-bowhay: reverted

scipy/integrate/tests/test_quadrature.py
```diff
@@ -563,20 +566,30 @@ def test_nd(self, axis, x_ndim, x_len, i_ndim, dx):
...
+        # uses `apply_along_axis`
+        ref = cumulative_simpson_nd_reference(
+            np.asarray(y), x=np.asarray(x), dx=None if dx is None else np.asarray(dx),
+            initial=None if initial is None else np.asarray(initial), axis=axis
+        )
+        xp_assert_close(res, xp.asarray(ref), rtol=1e-15)
 
+    @pytest.mark.filterwarnings(
+        "ignore:Creating a tensor from a list of numpy.ndarrays:UserWarning"
+    )
```
lucascolley: can we get rid of this by adding an `np.asarray` on line 606?

lucascolley (approved):
thanks Jake, LGTM!

Issue #23522, opened 2025-08-28
MAINT: Replace deprecated usage of numpy array shape, dtype and strides setting


eendebakpt:
### Describe your issue.

In Numpy the direct setting of the dtype, shape and strides attributes is being deprecated as these operations are not safe (in particular in the free-threading build). See https://github.com/numpy/numpy/issues/28800 for details.

Most direct setting of the attributes can be replaced with alternatives like array.view(dtype=...) or array.reshape(new_shape). These alternatives have a minor performance penalty (since a new object has to be created), but in general these methods create a view so no data is copied.

### Reproducing Code Example

Potential issues in the code: https://github.com/search?q=repo%3Ascipy%2Fscipy+%22.shape+%3D+%22+language%3APython&type=code

### SciPy/NumPy/Python version and system information

```shell
All
```

lucascolley:
Would you be interested in sending a PR @eendebakpt ?

eendebakpt:
> Would you be interested in sending a PR [@eendebakpt](https://github.com/eendebakpt) ?

Yes, although this issue is also good for any newcomers to implement part of the changes.

lucascolley:
> > Would you be interested in sending a PR [@eendebakpt](https://github.com/eendebakpt) ?
> 
> Yes, although this issue is also good for any newcomers to implement part of the changes. 

Would you mind writing a sentence or two describing how to tackle this as a good first issue? Thanks!

eendebakpt:
How to tackle this:
 
1. Identify usage of `shape` or `dtype` setting on numpy arrays. Searching on `.shape =` using github gives potential cases. Another option is to locally install numpy with the deprecations (e.g. https://github.com/numpy/numpy/pull/29536 or https://github.com/numpy/numpy/pull/29575) and run the scipy tests
2. For `shape` setting use `array.reshape(...)` instead (see https://github.com/scipy/scipy/pull/23523 for examples). For `dtype` setting use `array.view(...)` (see https://github.com/numpy/numpy/pull/29575 for examples)
3. Sometimes one really has to set the attribute of the array (for example in `__array_finalize__`, or subclasses initializers) and it is not enough to return a view. Make separate PRs for these cases so that can be reviewed in more detail. In the proposed numpy changes there are private setters available (e.g. `np.ndarray._set_dtype`)

nickodell:
I looked through the .dtype uses, and I only found one that seems questionable.

[...]

PR #23523, opened 2025-08-28
MAINT: Use reshape instead of setting the shape attribute for numpy arrays

eendebakpt:
We replace setting of the `shape` attribute with calls to `reshape(...)`. In this PR we only handle a part of the cases to keep the PR small for reviewers. In followup PRs we can handle the remaining cases.

#### Reference issue

Addresses part of #23522

Addresses (upcoming) deprecations from numpy. See https://github.com/numpy/numpy/issues/28800


lucascolley:
Since this isn't a deprecation itself but rather a response to an upstream deprecation, let's reclassify this as maintenance.

lucascolley (approved):
thanks @eendebakpt, LGTM!

PR #23524, opened 2025-08-28
MAINT: Use reshape instead of setting the shape attribute for numpy

eendebakpt:
#### Reference issue

Addresses part of https://github.com/scipy/scipy/issues/23522

We replace setting of the shape attribute with calls to reshape(...). In this PR we only handle a part of the cases to keep the PR small for reviewers. In followup PRs we can handle the remaining cases.


lucascolley (approved):
thanks @eendebakpt !

tylerjereddy (approved):
This looks sensible to me as well. The doc build CI failure looks unrelated (what is that, hundreds of warnings about NumPy?).

I may squash merge though--don't think we need four commits for this one, especially since they all have an almost identical message.

PR #23533, opened 2025-08-31
ENH:optimize: Implement backstep behavior in SLSQP on raise of Analysis Error

andrewellis55:
#### Reference issue
None

This feature implements the ability for a user to trigger backstopping in the SLSQP algorithm, via raising the newly added AnalysisError. 

In large scale optimization problems, it can be common to have numerical iterative processes as part of the objective or constraint functions. In cases where these processes fail to converge, the optimization may fail to recover. Furthermore, many problem have invalid regions where either constraints or the objective cannot be evaluated. In cases where a user can detect this occurrence during function evaluation, they can now force SLSQP to backstep by raising an AnalysisError.

It should be noted that this is not a substitute for constraints. In the convergence failure cases, regions that cause this can be hard to predict making the user of constraints to avoid them challenging. In the invalid region cases, the SLSQP algorithm will often iterate through infeasible space, meaning that constraints cannot fully prevent the optimization from stepping through the region. 

This functionality is current provided by a number of modern optimization (IPOPT, SNOPT) and is build upon by MDO frameworks such as [OpenMDAO](https://openmdao.org/newdocs/versions/latest/advanced_user_guide/analysis_errors/analysis_error.html).

#### Simple Example
Here is an example using a function similar to the Rosenbrock function, where a number of infeasible regions have been place throughout the design space. The attached screenshots show the behavior from a starting point of (2, 2) and (2, 4).

```python
def fun(x):
    if np.cos(2.0*x[0]) + np.cos(2.0*x[1]) > 0.5:
        raise AnalysisError("Invalid Region")
    return (1-x[0])**2 + (1-x[1])**2 + 0.5*(2.0*x[1]-x[0]**2)**2

res = minimize(fun, starting_point, method="SLSQP")
```
In the case of the starting point of (2, 2), the function hits the invalid region, and is able to step out with a single backstep (note that after iter 2, the optimizer take additional minor iterations unrelated to the user triggered backstep)
opt_progress_start_2 00_2 00
```
     message: Optimization terminated successfully
     success: True
      status: 0
         fun: 0.09194417645660757
           x: [ 1.213e+00  8.241e-01]
         nit: 6
         jac: [-2.326e-03  1.569e-03]
        nfev: 19
        njev: 6
```

In the case of the starting point of (2,4), the function hits numerous invalid regions during the first iteration, and is still able to recover and complete the optimization
opt_progress_start_2 00_4 00
```
     message: Optimization terminated successfully
     success: True
      status: 0
         fun: 0.09194382638399702
           x: [ 1.213e+00  8.242e-01]
         nit: 7
         jac: [-8.721e-05  2.895e-04]
        nfev: 22
        njev: 7
```

### DIFF
--- a/scipy/optimize/__init__.py
+++ b/scipy/optimize/__init__.py
@@ -335,6 +335,14 @@ Benchmark problems
    rosen_hess - The Hessian matrix of the Rosenbrock function.
    rosen_hess_prod - Product of the Rosenbrock Hessian with a vector.
 
+User Errors
+-----------
+
+.. autosummary::
+   :toctree: generated/
+
+   AnalysisError - Error that can be raised in SLSQP to trigger backstepping.
+
 Legacy functions
 ================
 
@@ -429,7 +437,7 @@ from ._lbfgsb_py import fmin_l_bfgs_b, LbfgsInvHessProduct
 from ._tnc import fmin_tnc
 from ._cobyla_py import fmin_cobyla
 from ._nonlin import *
-from ._slsqp_py import fmin_slsqp
+from ._slsqp_py import fmin_slsqp, AnalysisError
 from ._nnls import nnls
 from ._basinhopping import basinhopping
 from ._linprog import linprog, linprog_verbose_callback
@@ -446,6 +454,7 @@ from ._dual_annealing import dual_annealing
 from ._qap import quadratic_assignment
 from ._direct_py import direct
 from ._milp import milp
+
 
 # Deprecated namespaces, to be removed in v2.0.0
 from . import (
--- a/scipy/optimize/__slsqp.c
+++ b/scipy/optimize/__slsqp.c
@@ -1,4 +1,5 @@
 #include "__slsqp.h"
+#include <stdio.h>
 
 void __nnls(const int m, const int n, double* restrict a, double* restrict b, double* restrict x, double* restrict w, double* restrict zz, int* restrict indices, const int maxiter, double* rnorm, int* info);
 static void ldp(int m, int n, double* g, double* h, double* x, double* buffer, int* indices, double* xnorm, int* mode);
@@ -240,10 +241,18 @@ MODE1:
     }
     S->h1 = S->t - S->t0;
 
-    if ((S->h1 > (S->h3 / 10.0)) && (S->line <= 10))
+    if ((S->h1 > (S->h3 / 10.0) || S-> fnc_failure) && (S->line <= 10))
     {
         S->alpha = fmax(S->h3/(2.0*(S->h3 - S->h1)), alfmin);
         goto LINE_SEARCH;
+    } else if (S->line == 11 && S-> fnc_failure) {
+        // Prevents termination at a step when a fnc_failure occured &
+        //   reset to last viable point before linesearch started so
+        //   that the optimization does not exit to a point that can't
+        //   be evaluated
+        for (int i = 0; i < n; i++) { sol[i] = x0[i]; }
+        S->mode = 9;
+        return;
     }
 
     // Check convergence
--- a/scipy/optimize/__slsqp.h
+++ b/scipy/optimize/__slsqp.h
@@ -76,7 +76,7 @@ void dtrsv_(char* uplo, char* trans, char* diag, int* n, double* a, int* lda, do
 // and back such that it is thread-safe.
 struct SLSQP_vars {
     double acc, alpha, f0, gs, h1, h2, h3, h4, t, t0, tol;
-    int exact, inconsistent, reset, iter, itermax, line, m, meq, mode, n;
+    int exact, inconsistent, reset, iter, itermax, line, m, meq, mode, n, fnc_failure;
 };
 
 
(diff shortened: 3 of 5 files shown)


andyfaff:
@andrewellis55 normally we request that changes of this magnitude are first floated on the SciPy [discussion forum](https://discuss.scientific-python.org/c/contributor/scipy/32). This is to give a greater level of visibility to proposed changes. 
Discussing them first guards against unnecessary work on the part of the contributor, if the suggested change is regarded as not required, or a different approach is suggested.

This would be regarded as a big change, as it would be reasonable for an end user to expect this kind of functionality across the entire `optimize` stable. Behaviour of the feature would have to be consistent with all methods.
I can see the attraction to it. Do you have experience as to how it could be implemented for the other `minimize` methods?

andrewellis55:
@andyfaff thanks for the feedback! I've created a topic on the discussion board [here](https://discuss.scientific-python.org/t/user-trigger-able-backstopping-in-slsqp/2084) to get some wider community feedback.

Personally, I don't have a huge amount of time to dedicate to bringing this feature across the board for `minimize`, however I feel that this feature is quite critical to make SciPy a viable tool in many optimization use cases like those I've mentioned (nested iterative processes and optimization around black box tools with strict design spaces). I've been working in the aerospace industry on optimization tooling for a number of years, and in my experience, myself and other users have been forced to switch to [pyOptSparse](https://github.com/mdolab/pyoptsparse) when these cases arise as it integrates [IPOPT](https://coin-or.github.io/Ipopt/) (free, but not an SQP based algorthim, which can be needed in some situations) and [SNOPT](https://ccom.ucsd.edu/~optimizers/solvers/snopt/) (which costs thousands of dollars) which can each perform this sort of backtracking. I think SciPy is definitely the most widely available optimization tool in python and so it would be a shame not to have at least one optimizer with functional parity in this regard.

That said, I can definitely see how my current implementation might imply the availability of the `AnalysisError` across `minimize`  and so I'd definitely be open to seeing how it could be made to appear more SLSQP focused. I chose the implementation of a newly added error class mainly for backwards compatibility. There are other methods, like in IPOPT, I believe it's triggered by passing a `nan` to the optimizer, and there's also the option of just auto-triggering this on a general `ValueError` however these would change the behavior of SLSQP for existing users, and I know the stability of a widely used tool like SciPy is particularly important. I figured with a new error class, there's no way this functionality could start triggering in someone's legacy codebase that depends on SLSQP.

Anyway, I'm definitely open to community feedback here, but bringing up the feature for the rest of the optimizers (which I don't use nearly as much as SLSQP) would probably be outside what I can commit to doing.

andrewellis55:
@robfalck maybe it could be good to get some thoughts from you/the OpenMDAO team as this should enable some new functionality for SciPy user using OpenMDAO that want to trigger AnalysisErrors (there's probably a little bring up on the OpenMDAO side, but I wouldn't mind doing that if this PR is accepted here)

robfalck:
I agree that this is a valuable capability. From the Scipy side I can see the desire to not limit this capability to SLSQP, but at the same time doing so across a lot of optimization algorithms is probably a big task.

From my perspective on the OpenMDAO side, calling this `AnalysisError` in both packages while they are different objects will probably lead to confusion at some point.

Since Numpy will optionally raise a FloatingPointError if invalid values (nan, inf) are detected, maybe scipy.minimize should just detect if that is raised in `fun`, `jac`, `hess`, or `hessp`.  That would make a lot of existing code compatible with this just be setting a few flags.

andrewellis55:
I figured I'd give this a bit of time to see if any other community feedback popped up.

Thanks for the feedback @robfalck, I agree that detecting a nan or inf would be the easiest, but that could cause unexpected behavior for users that currently expect optimizations to fail if these are encountered. I felt that an explicitly raised error was the most transparent implementation for existing users given SciPy's widespread usage. @andyfaff not sure if you have any preferences either way.

Additionally, I've just pushed a new commit that adds a few more test problems that encompass what I believe to be the three most common use cases:

**Case 1: Function cannot be evaluated**
I added a simple test case to show that the optimizer can toss negative values into a square root function in a constraint and backstep successfully

**Case 2: General Invalid Regions** 
This is the case I've already provided and shown in the images I attached to the PR

**Case 3: Diverge Simulations**
I tried looking for a good example of a simple simulation with divergent behavior and came across a cool Nature paper [Simple mathematical models with very complicated dynamics](https://www.nature.com/articles/261459a0#citeas), available free [here](https://www.researchgate.net/publication/237005499_Simple_Mathematical_Models_With_Very_Complicated_Dynamics). I added a test case based on this with some intentionally poor problem scaling to push the optimizer into the divergent region. I made a demos graph [here](https://www.desmos.com/calculator/livifwtf8i) where you can see what's going on. z is the design variable, and as you vary it, you can see the simulation diverges. In the new test case I provided the optimizer steps into the diveregent region on the second iteration, causing all sorts of overflow, then backsteps to a valid region and converges successfully.

Lastly I added a test case to demonstrate that callbacks still work just fine as @rgommers raised this as a concern.

andrewellis55:
[This commit](https://github.com/scipy/scipy/compare/a2f5a18087da8b13f9d613e020b59a6ead32e9f8..40fe458a7655b74fd04095d6d6fe10ef40bd5203) should fix the test that was failing.

PR #20190, opened 2024-03-05
ENH: sparse: first pass at array API standard compat

ilan-gold:
#### Reference issue
Towards #18915

#### What does this implement/fix?
This begins the works towards an `array_api` compliant implementation.  I've take some first steps here just to see what works and what doesn't and document how to do this.  Of note is that most of the current errors arise from either indexing problems or lack of 1d support.   See also #20128

#### Additional information
In terms of current failing tests on account of missing names/functions (i.e., not 1D issues), we have a sampling here, although this is probably the majority:

<details>

```
FAILED array_api_tests/test_data_type_functions.py::test_isdtype - AssertionError: isdtype is not defined in scipy.array_api
FAILED array_api_tests/test_has_names.py::test_has_names[linalg-cross] - AssertionError: scipy.array_api is missing the linalg extension function cross()
FAILED array_api_tests/test_has_names.py::test_has_names[linalg-diagonal] - AssertionError: scipy.array_api is missing the linalg extension function diagonal()
FAILED array_api_tests/test_has_names.py::test_has_names[linalg-matmul] - AssertionError: scipy.array_api is missing the linalg extension function matmul()
FAILED array_api_tests/test_has_names.py::test_has_names[linalg-matrix_norm] - AssertionError: scipy.array_api is missing the linalg extension function matrix_norm()
FAILED array_api_tests/test_has_names.py::test_has_names[linalg-matrix_power] - AssertionError: scipy.array_api is missing the linalg extension function matrix_power()
FAILED array_api_tests/test_has_names.py::test_has_names[linalg-matrix_rank] - AssertionError: scipy.array_api is missing the linalg extension function matrix_rank()
FAILED array_api_tests/test_has_names.py::test_has_names[linalg-matrix_transpose] - AssertionError: scipy.array_api is missing the linalg extension function matrix_transpose()
FAILED array_api_tests/test_has_names.py::test_has_names[linalg-outer] - AssertionError: scipy.array_api is missing the linalg extension function outer()
FAILED array_api_tests/test_has_names.py::test_has_names[linalg-slogdet] - AssertionError: scipy.array_api is missing the linalg extension function slogdet()
FAILED array_api_tests/test_has_names.py::test_has_names[linalg-tensordot] - AssertionError: scipy.array_api is missing the linalg extension function tensordot()
FAILED array_api_tests/test_has_names.py::test_has_names[linalg-trace] - AssertionError: scipy.array_api is missing the linalg extension function trace()
[...]

PR #23534, opened 2025-08-31
ENH: sparse: add funcs `swapaxes`, `permute_dims`, `expand_dims` with tests and docs

dschult:
Adding 3 sparse functions that manipulate axes.
These are helpful now that we support nD sparse arrays, and have been requested as a desirable nD feature.

- `swapaxes()` mimics the NumPy function.
- `expand_dims()` and `permute_dims()` mimic the array-api functions (and in numpy).

I used tests from numpy and added tests for sparse aspects and axis validity. But please suggest other tests if needed.


lucascolley (commented):
nice!

> please suggest other tests if needed

https://github.com/data-apis/array-api-tests :) we should resolve the merge conflicts in gh-20190 and see if the `{permute,expand}_dims` tests pass with this PR!

scipy/sparse/_construct.py
```diff
@@ -28,6 +29,141 @@
...
+    idx = axis if axis >= 0 else axis + A.ndim + 1
+    if idx < 0 or idx > A.ndim:
+        raise ValueError(f"Invalid axis {axis} for N={A.ndim}. Must be in [-N-1, N].")
+
+    A = A.tocoo(copy=True)
+    new_coord = np.zeros_like(A.coords[0])
+
+    A.coords = A.coords[:idx] + (new_coord,) + A.coords[idx:]
+    A._shape = A.shape[:idx] + (1,) + A.shape[idx:]
+    A.has_canonical_format = False  # data not guaranteed sorted
```
perimosocordiae: I don't think we need to set this here:
 - adding a new trivial dimension won't create any duplicate indices
 - if indices were already sorted, adding a zero at any position won't change the sort order

If I'm missing something, please let me know.

scipy/sparse/_construct.py
```diff
@@ -28,6 +29,141 @@
...
+    if not isintlike(axis1):
+        raise ValueError(f"Invalid axis1 {axis1}. Must be an integer.")
+    idx1 = axis1 if axis1 >= 0 else axis1 + A.ndim
+    if idx1 < 0 or idx1 > A.ndim - 1:
+        raise ValueError(f"Invalid axis1 {axis1} for N={A.ndim}. Must be in [-N, N-1].")
+    if not isintlike(axis2):
+        raise ValueError(f"Invalid axis2 {axis2}. Must be an integer.")
+    idx2 = axis2 if axis2 >= 0 else axis2 + A.ndim
+    if idx2 < 0 or idx2 > A.ndim - 1:
+        raise ValueError(f"Invalid axis2 {axis2} for N={A.ndim}. Must be in [-N, N-1].")
```
perimosocordiae: Side note: it would be nice if we had a helper function for this kind of integer axis validation. We have `_sputils.validateaxis` but that's more for the general case, and wouldn't produce very helpful error messages. Might be worth adding a related function like `validate_single_axis()` or something, that does the type checking, negative handling, bounds checking, and returns an int.

scipy/sparse/_construct.py
```diff
@@ -28,6 +29,141 @@
...
+    A = A.tocoo(copy=True)
+    if idx1 == idx2:
+        return A
+    new_coords = list(A.coords)
+    new_coords[idx1], new_coords[idx2] = new_coords[idx2], new_coords[idx1]
+    A.coords = tuple(new_coords)
+    new_shape = list(A.shape)
+    new_shape[idx1], new_shape[idx2] = new_shape[idx2], new_shape[idx1]
+    A._shape = tuple(new_shape)
+    A.has_canonical_format = False  # data not guaranteed sorted
```
perimosocordiae: This operation almost certainly unsorts the indices, in fact!

scipy/sparse/_construct.py
```diff
@@ -28,6 +29,141 @@
...
+    new_coord = np.zeros_like(A.coords[0])
+
+    A.coords = A.coords[:idx] + (new_coord,) + A.coords[idx:]
+    A._shape = A.shape[:idx] + (1,) + A.shape[idx:]
+    A.has_canonical_format = False  # data not guaranteed sorted
+    return A
+
+
+def swapaxes(A, axis1, axis2):
+    """Interchange two axes of an array.
```
perimosocordiae, line 76: Could we write swapaxes in terms of permute_dims? That might simplify things a bit.

perimosocordiae (changes requested):

scipy/sparse/_construct.py
dschult: The `axis` args to `swapaxes` avoid this concern now that I rewrote `swapaxes` to call `permute_dims`. But we still need something like this for `permute_dims`.

`_sputils.validateaxis` could work well for the `axes` arg to `permute_dims`, except `validateaxis` turns any tuple without dups and of length `ndim` into `None` (special treatment for `min`/`max`-like cases). So the one case that we need, isn't provided! :/     Such is the fate of centralized utilities where there are lots of special cases. :}

I could change validateaxis to return a length ndim tuple and change the code elsewhere that uses it to check for a length ndim tuple instead of checking for `None`.  `validateaxis` is not in `_sputils.__all__` so its a private function. (Has to be called by importing directly from `_sputils` which makes it private I think.)  But there might be some folks who are using it.  (Is this a case of "regularly breaking private backward compatibility"?)

Another approach is to use/abuse the function `_index._asindices(idx, length, format)` which might do the trick. It is written for an array of indices along an axis of size `length`, not a tuple of axis numbers with max-value `ndim`. But the treatment is basically the same and the unrelated checks all pass if we input `_asindices(idx=axis1, length=A.ndim, format="coo")`. The `format="coo"` avoids the check for axis1 being either 1D or 2D  as well as avoiding the shortcut for `lil` format.  The downside for use with `permute_dims` is no check for duplicates. But we could do that on our own.

I'd prefer to make the `validateaxis` function work for this case too and update the other uses of that function as needed. But that should be in a separate follow-up PR because it impacts other code and backward compatibility of a private function.  

So, in this PR I copy/pasted the guts of `validateaxis`, removing the lines that check for length `ndim` and abort. Then in a follow-up PR I can replace that code with a call to a changed validateaxis function.

Thoughts?

scipy/sparse/_construct.py
```diff
@@ -28,6 +29,144 @@
...
+    ValueError
+        If provided a non-integer or out of range ``[-N, N-1]`` axis,
+        where ``N`` is ``A.ndim``.
+    """
+    axes = np.arange(A.ndim)
+    try:
+        axes[[axis1, axis2]] = axes[[axis2, axis1]]
+    except IndexError as err:
+        # message is like: Invalid axis -4 is out of bounds for axis 0 with size 2
+        raise ValueError(f"Invalid axis {str(err)[6:]}")
```
perimosocordiae:
```suggestion
        # original msg looks like: "index -4 is out of bounds for axis 0 with size 2"
        msg = str(e)
        msg = msg.removeprefix('index ').split(' axis ', 1)[0]
        # Final error is: "Invalid axis: -4 is out of bounds for ndim=2"
        raise ValueError(f"Invalid axis: {msg} ndim={A.ndim}")
```

scipy/sparse/_construct.py
```diff
@@ -28,6 +29,144 @@
...
+    ValueError
+        If provided a non-integer or out of range ``[-N, N-1]`` axis,
+        where ``N`` is ``A.ndim``.
+    """
+    ndim = A.ndim
+    if axes is None:
+        axes = tuple(range(ndim)[::-1])
+    elif len(axes) != ndim:
+        raise ValueError(f"Incorrect number of axes: {ndim} instead of {A.ndim}")
+
```
perimosocordiae, line 143: Should we check for axes equal to `range(ndim)` (i.e, the trivial permutation?) We could return early if so.

perimosocordiae (approved):

scipy/sparse/_construct.py
perimosocordiae: That plan sounds good to me.

dschult:
I added the suggested error message improvement, and a check in `permute_dims` for the identity permutation. When `axes == list(range(ndim))` we now return `A.copy()` or `A` depending on the `copy` input param.  Numpy returns a view or a copy. We return the actual object or a copy.

New tests check the identity permutation behavior as well as `swapaxes(A, 0, 0)` which leads to an identity permutation in `permute_dims`.

The CI failures seem unrelated.

lucascolley:
shall we merge @perimosocordiae ?

perimosocordiae:
Thanks for the reminder. Merged!

WarrenWeckesser:
This is a great addition!  But please 🙏 , when you add new public functions, be sure to include examples in the docstrings. We want the total in https://github.com/scipy/scipy/issues/7168 to be monotonically decreasing over time.

lucascolley:
Good point, I forgot that that is mentioned in https://scipy.github.io/devdocs/dev/contributor/development_workflow.html#checklist-before-submitting-a-pr

PR #23535, opened 2025-08-31
BUG: special: Fix switched overflow and underflow error messages.

WarrenWeckesser:


scipy/special/tests/test_sf_error.py
```diff
@@ -142,3 +142,14 @@ def test_errstate_all_but_one():
...
+
+def test_check_overflow_message():
+    # Regression test for a bug where the overflow and underflow
+    # messages were switched.
+    with sc.errstate(all='raise'):
+        try:
+            # This should trigger an overflow:
+            sc.yn(3, 1e-105)
+        except sc.SpecialFunctionError as exc:
+            assert 'overflow' in exc.args[0]
```
lucascolley: can we use `with pytest.raises` to ensure that this raises an exception?

lucascolley (commented):
thanks Warren!

scipy/special/tests/test_sf_error.py
WarrenWeckesser: Yes, that would be better.

WarrenWeckesser: Done.  I went ahead and force-pushed the change.  It is a small PR, so I hope you don't mind.  It should save a bit of squash-merge commit message editing.

lucascolley (approved):

scipy/special/tests/test_sf_error.py
lucascolley: Not a problem in the slightest! 

P.S. if you mean removing the list of extra commits from the description of the default squash message, https://github.com/refined-github/refined-github does that automatically. Not sure about the mobile app though, that's where I just pressed the green button now!

PR #23537, opened 2025-09-01
ENH: Add two private ufuncs related to generalized harmonic numbers.

WarrenWeckesser:
Two private ufuncs are added to `scipy.special._ufuncs`:

* `_gen_harmonic(n, a)` computes the generalized harmonic number

$$\sum_{i=1}^n i^{-a}$$

* `_normalized_gen_harmonic(j, k, n, a)` computes
  
$$\frac{\displaystyle\sum_{i=j}^k i^{-a}}{\displaystyle\sum_{i=1}^n i^{-a}}$$

These functions are used in `scipy.stats.zipfian` to improve performance and to improve precision when a is greater than but close to 1.

These aren't made part of the public API, so I don't think they need to become part of `xsf`.  That could be done in the future if there is downstream interest in such functions.

#### Reference issue

This change was inspired by https://github.com/scipy/scipy/issues/23487, but it doesn't close that issue.  The change makes the generation of random variates in `zipfian.rvs()` significantly faster, but there is another change (to be submitted in a follow-up PR) that should definitively resolve the slowness issue.

Before this change, the calculation of the harmonic numbers used in `stats.zipfian` when a <= 1 was implemented with a Python for-loop, so it was very slow.  This was especially evident in the generation of random variates, as noted in gh-23487.

This PR also improves the precision of the calculations in `zipfian` when `a` is greater than but close to 1.  When `a` > 1, the sum of the powers can be written $\zeta(a, m) - \zeta(a, n + 1)$, where $\zeta(a, m)$ is the Hurwitz zeta function.  Because of the subtraction, this formula loses numerical precision when $\zeta(a, m) \approx \zeta(a, n + 1)$.  In that case, the new function falls back to the simple sum of the powers. (See the comments in the code for more precise details.)

### DIFF
--- a/scipy/special/_special_ufuncs.cpp
+++ b/scipy/special/_special_ufuncs.cpp
@@ -35,6 +35,7 @@
 #include <xsf/trig.h>
 #include <xsf/wright_bessel.h>
 #include <xsf/zeta.h>
+#include "gen_harmonic.h"
 
 // This is the extension module for the NumPy ufuncs in SciPy's special module. To create such a ufunc, call
 // "xsf::numpy::ufunc" with a braced list of kernel functions that will become the ufunc overloads. There are
@@ -46,8 +47,10 @@
 
 extern const char *_cospi_doc;
 extern const char *_sinpi_doc;
+extern const char *_gen_harmonic_doc;
 extern const char *_log1mexp_doc;
 extern const char *_log1pmx_doc;
+extern const char *_normalized_gen_harmonic_doc;
 extern const char *airy_doc;
 extern const char *airye_doc;
 extern const char *bei_doc;
@@ -232,6 +235,14 @@ PyMODINIT_FUNC PyInit__special_ufuncs() {
                            static_cast<xsf::numpy::F_F>(xsf::cospi), static_cast<xsf::numpy::D_D>(xsf::cospi)},
                           "_cospi", _cospi_doc);
     PyModule_AddObjectRef(_special_ufuncs, "_cospi", _cospi);
+
+    PyObject *_gen_harmonic =
+        xsf::numpy::ufunc({gen_harmonic}, "_gen_harmonic", _gen_harmonic_doc);
+    PyModule_AddObjectRef(_special_ufuncs, "_gen_harmonic", _gen_harmonic);
+
+    PyObject *_normalized_gen_harmonic =
+        xsf::numpy::ufunc({normalized_gen_harmonic}, "_normalized_gen_harmonic", _normalized_gen_harmonic_doc);
+    PyModule_AddObjectRef(_special_ufuncs, "_normalized_gen_harmonic", _normalized_gen_harmonic);
 
     PyObject *_lambertw = xsf::numpy::ufunc(
         {static_cast<xsf::numpy::Dld_D>(xsf::lambertw), static_cast<xsf::numpy::Flf_F>(xsf::lambertw)}, "_lambertw",
--- a/scipy/special/functions.json
+++ b/scipy/special/functions.json
@@ -9,16 +9,6 @@
             "cosine_invcdf": "d->d"
         }
     },
-    "_gen_harmonic": {
-        "gen_harmonic.h++": {
-            "gen_harmonic": "qd->d"
-        }
-    },
-    "_normalized_gen_harmonic": {
-        "gen_harmonic.h++": {
-            "normalized_gen_harmonic": "qqqd->d"
-        }
-    },
     "_ellip_harm": {
         "_ellip_harm.pxd": {
             "ellip_harmonic": "ddiiddd->d"
--- a/scipy/special/functions.json
+++ b/scipy/special/functions.json
@@ -9,6 +9,16 @@
             "cosine_invcdf": "d->d"
         }
     },
+    "_gen_harmonic": {
+        "gen_harmonic.h++": {
+            "gen_harmonic": "qd->d"
+        }
+    },
+    "_normalized_gen_harmonic": {
+        "gen_harmonic.h++": {
+            "normalized_gen_harmonic": "qqqd->d"
+        }
+    },
     "_ellip_harm": {
         "_ellip_harm.pxd": {
             "ellip_harmonic": "ddiiddd->d"
--- a/scipy/special/gen_harmonic.h
+++ b/scipy/special/gen_harmonic.h
@@ -150,12 +150,12 @@ normalized_gen_harmonic(int64_t j, int64_t k, int64_t n, double a)
     //
     // Now we know 1 <= j <= k <= n
     //
+    if (n == 1) {
+        // IEEE: pow(1.0, _) is 1.0.
+        // n == 1 implies j == k == 1.
+        return 1.0;
+    }
     if (std::isnan(a)) {
-        if (n == 1) {
-            // n == 1 implies j == k == 1.
-            // IEEE: pow(1.0, nan) is 1.0.
-            return 1.0;
-        }
         return NAN;
     }
     if (std::isinf(a)) {
@@ -172,11 +172,7 @@ normalized_gen_harmonic(int64_t j, int64_t k, int64_t n, double a)
         }
         else {
             // a = -inf
-            if (n == 1) {
-                // Numerator and denominator are both 1.
-                return 1.0;
-            }
-            else if (k == 1) {
+            if (k == 1) {
                 // Numerator is 1, denominator is +inf.
                 return 0.0;
             }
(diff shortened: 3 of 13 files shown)


scipy/stats/_discrete_distns.py
```diff
@@ -5,7 +5,8 @@
 from functools import partial
 
 from scipy import special
-from scipy.special import entr, logsumexp, betaln, gammaln as gamln, zeta
+from scipy.special import entr, logsumexp, betaln, gammaln as gamln
+from scipy.special._ufuncs import _gen_harmonic
```
Peiffap: The rest of the file seems to use imports as `scu.<fn>`.

WarrenWeckesser: Good point! I'll fix that to be consistent with the other uses of `scu`.

WarrenWeckesser: Done.

WarrenWeckesser:
The inefficiency of the Zipfian CDF and SF calculations,

    CDF:  scu._gen_harmonic(1, k, a) / scu._gen_harmonic(1, n, a)
    SF:   scu._gen_harmonic(k + 1, n, a) / scu._gen_harmonic(1, n, a)

irks me.  A normalized generalized harmonic function would make each of those a single function call, in which the partial sum is computed in the same loop as the total sum.  As long as I'm adding private ufuncs, I might as well handle that too.

I'm moving the PR to draft while I see how well that works out.

*Edit:* Updated, and now the Zipfian calculations are

    PMF:  scu._normalized_gen_harmonic(k, k, n, a)
    CDF:  scu._normalized_gen_harmonic(1, k, n, a)
    SF:   scu._normalized_gen_harmonic(k + 1, n, n, a)

WarrenWeckesser:
Updated, and ready for review again.

scipy/special/gen_harmonic.h
```diff
@@ -0,0 +1,231 @@
...
+            // a = -inf
+            if (n == 1) {
+                // IEEE: pow(1.0, inf) = 1
+                return 1.0;
+            }
+            else {
+                return INFINITY;
+            }
+        }
+    }
```
Peiffap, line 78: Is there a reason not to do this? Same can be done in the normalized version.
```suggestion
    if (n == 1) {
        // IEEE: pow(1.0, _) is 1.0.
        return 1.0;
    }
    if (std::isnan(a)) {
        return NAN;
    }
    if (std::isinf(a)) {
        if (a > 0) {
            // a = +inf
            return 1.0;
        }
        else {
            // a = -inf
            return INFINITY;
        }
    }
```

scipy/special/gen_harmonic.h
```diff
@@ -0,0 +1,231 @@
...
+    if (a == 0) {
+        return static_cast<double>(n);
+    }
+    if (a <= 1) {
+        return sum_powers(1, n, a);
+    }
+    else {
+        // If here, we know a is finite and a > 1, and n >= 1.
```
Peiffap: This would need to change too then.
```suggestion
        // If here, we know a is finite and a > 1, and n > 1.
```

Peiffap (commented):
A quick question...

scipy/special/gen_harmonic.h
WarrenWeckesser: That has the same result with fewer lines, so yes, that looks good.  Thanks!

scipy/special/functions.json
steppi, line 1: Adding entries to this file isn't necessary since you've registered the ufuncs in `_special_ufuncs.cpp`.

WarrenWeckesser: Ah, right.  I was on autopilot for this.  Fixed.

scipy/special/_special_ufuncs.cpp
```diff
@@ -233,6 +236,14 @@ PyMODINIT_FUNC PyInit__special_ufuncs() {
                           "_cospi", _cospi_doc);
     PyModule_AddObjectRef(_special_ufuncs, "_cospi", _cospi);
 
+    PyObject *_gen_harmonic =
+        xsf::numpy::ufunc({gen_harmonic}, "_gen_harmonic", _gen_harmonic_doc);
+    PyModule_AddObjectRef(_special_ufuncs, "_gen_harmonic", _gen_harmonic);
+
+    PyObject *_normalized_gen_harmonic =
+        xsf::numpy::ufunc({normalized_gen_harmonic}, "_normalized_gen_harmonic", _normalized_gen_harmonic_doc);
+    PyModule_AddObjectRef(_special_ufuncs, "_normalized_gen_harmonic", _normalized_gen_harmonic);
```
steppi, line 245: Just pointing this out for the general audience. This  works because `gen_harmonic` and `normalized_gen_harmonic` each only have one type signature so we don't need to specify an overload with something like `static_cast<xsf::numpy::qd_d>(normalized_gen_harmonic)`.

 My preference would be to include the `static_cast` anyway just for the sake of consistency and readability. Technically though, `q` is for `long long` which in practice for every platform SciPy supports should be the same as `int64`, then again, who knows what the future holds. Do you think it would be better if there more explicit signatures in `xsf/numpy.h` like `xsf::numpy::i8d_d`? I'm fine leaving this as is for now.

There's also a question of whether we may actually additional overloads, but as is, this does reproduce the behavior with respect to types of existing `zipfian` functions, since the old `_gen_harmonic` explicitly made the output have `dtype` `float`. That's a question I'm not really concerned about for now, but just noting it for the future.

I'd also like to point out for the general audience that the current implementation does avoid issues surrounding NumPy default ints on different platforms (https://github.com/scipy/scipy/pull/21401) since the ufunc explicitly uses `int64`. This ties in with the question above of whether we want different overloads, since maybe it would be nice to allow computation with native 32 bit ints on 32 bit platforms, but it's not something I think matters that much for now.

steppi (approved):
Thanks @WarrenWeckesser. I think the implementation looks solid. I left one comment on some things I think are worth thinking about regarding `dtypes`, targeted for the general audience (so may sound like I'm trying to explain things that I'm actually confident you know already). 

These functions will probably find their way into `xsf` for the sake of CuPy support, but it's definitely easier for now to add them here. I'd like to set things up so that contributing to `xsf` is painless, but there's more work needed to get to that point.

scipy/special/_special_ufuncs.cpp
WarrenWeckesser: Thanks @steppi.

> My preference would be to include the static_cast anyway just for the sake of consistency and readability. 

I don't have a strong preference either way (OK, maybe a *slight* preference for minimizing boilerplate), so if you prefer the explicit cast, that's fine with me.

> Do you think it would be better if there more explicit signatures in xsf/numpy.h like xsf::numpy::i8d_d?

I think that would be great. Using `q` was just a pragmatic choice, since, as you noted, it should correspond to `int64` for any platform that we care about these days. A signature that allows `int64` to be specified explicitly would be much better.

WarrenWeckesser:
Thanks @steppi!

At the risk of being "that annoying pedant", and at the risk of looking like I'm singling out @steppi when in fact this comment applies to us all, I have one workflow-related request.  I realize that once the hard work of reviewing is done, we just want to click the merge or squash-merge button and move on.  But when I squash-merge (which was appropriately done here), I know I have one more small task to finish: copy-edit the commit message.  Many of the small commits made during review are undoing changes from a previous commit in the PR, or making trivial changes such as fixing typos or adjusting whitespace.  The commit messages for these changes are almost never needed in the commit message of the squash-merged commit.

For example, in this PR, all the messages other than the original commit message are irrelevant, and in fact slightly contradictory and confusing.  The message of the merged commit includes:

```
* Rearrange handling of special cases in gen_harmonic().
* Rearrange handling of special cases in normalized_gen_harmonic().
* Use numeric_limits instead of NAN and INFINITY macros.
* Be less conservative about switching from the zeta formula to summing the powers.
* Fix typo in comment.
* Remove _gen_harmonic and _normalized_gen_harmonic from functions.json.
```
but these all refer to changes related to a *previous commit* and not to changes to the main development branch, so they provide no useful information about the squashed commit.  For example, that last line says `Remove _gen_harmonic and _normalized_gen_harmonic from functions.json`, but once the commits have been squashed, those names were never in `functions.json`, so the text is meaningless (and therefore confusing to future readers).

For reference, we have a [brief note about commit messages](https://scipy.github.io/devdocs/dev/contributor/development_workflow.html#writing-the-commit-message) in our development docs. I wanted to say "a web search for 'good commit message' will provide links to good suggestions..." but these days, there is so much fluff and AI generated regurgitations of that fluff, that I'm not so sure about that.  A [google styleguide website](https://google.github.io/styleguide/) includes a link to [How to Write a Git Commit Message](https://cbea.ms/git-commit/), which looks reasonable.

Again, thanks @steppi for the review and merge of a nontrivial PR.  I don't mean to single you out--many of us have been accepting the generated merge commit message without thought for a long time.  I hope this (longer than expected) comment will get some of us to spend the extra minute or two required to clean up the commit message so it reflects the actual changes in the commit.

steppi:
Sorry @WarrenWeckesser. I should have done better and don't mind being singled out as an example. This seems like something where it would be helpful to have guidance on editing the commit message when squash merging a large PR in the development docs. A good process to follow might be for the reviewer to show how they plan to edit the commit message in a comment, and the PR author can give a thumbs up if they think the edits are accurate and acceptable. One could also keep only the PR title I suppose, but for large PRs like this, it does seem good to add a concise explanation of what has been changed and where and why these changes have been made to the body of the commit message. I'll try to be more disciplined from now.

PR #23538, opened 2025-09-01
MAINT: differential_evolution, polish takes a callable

andyfaff:
Alternate to #22957.

@mdhaber @mvieth


scipy/optimize/tests/test__differential_evolution.py
```diff
@@ -1607,6 +1608,43 @@ def rosen_vec(x):
...
+        res = differential_evolution(
+            rosen,
+            self.bounds,
+            polish=pf,
+            maxiter=1
+        )
+        # res.success will be False because of the small number of iterations
+        # The solution produced by DE would be bad after only one iteration.
+        # However, we still expect a good answer if the polishing worked
+        assert res.jac is not None
```
mdhaber, line 1626: Oh, I see above - this is populated from the result returned by the `polish` function if available.

mdhaber (commented):
Looks good from a quick review. I'd be willing to finish the review and merge if this route is deemed OK.

andyfaff:
I'm happy with this solution.
On of the drawbacks of other functions that use local minimisers, e.g. shgo, is that using `minimize_kwargs` leads to a lot of internal complexity. Offloading it to the user is a lot simpler.

scipy/optimize/_differentialevolution.py
```diff
@@ -159,20 +159,33 @@ def differential_evolution(func, bounds, args=(), strategy='best1bin',
...
         If True (default), then `scipy.optimize.minimize` with the `L-BFGS-B`
         method is used to polish the best population member at the end, which
         can improve the minimization slightly. If a constrained problem is
         being studied then the `trust-constr` method is used instead. For large
         problems with many constraints, polishing can take a long time due to
         the Jacobian computations.
+        Alternatively supply a callable that has a `minimize`-like signature,
+        ``polish_func(func, x0, **kwds)`` and returns an `OptimizeResult`. This
+        allows the user to have fine control over how the polishing occurs.
+        `bounds` and `constraints` will be present in `kwds`. Extra keywords
```
mdhaber:
```suggestion
        `bounds` and `constraints` will be present in ``kwds``. Extra keywords
```
(not a parameter of `differential_evolution`)

scipy/optimize/_differentialevolution.py
```diff
@@ -159,20 +159,33 @@ def differential_evolution(func, bounds, args=(), strategy='best1bin',
...
+        could be supplied to `polish_func` using `functools.partial`. It is the
+        user's responsibility to ensure that the polishing function obeys
+        bounds, any constraints (including integrality constraints), and that
+        appropriate attributes are set in the `OptimizeResult`, such as `fun`,
+        `x`, `nfev`, `jac`.
```
mdhaber: Fine points: `fun`, `x`, `nfev`, and `jac` are also not parameters or returns of the function, so double backticks would be most appropriate. 
More importantly, though - can you comment on why all this is emphasized? I'm thinking it may have been left over from before the `polish` callable was passed `bounds` and `constraints`? But since it was just stated that the user-provided `minimize` function is provided bounds and constraints, it seems natural that this function is responsible for ensuring that bounds/constraints are satisfied. (Why would it be reasonable for the user to assume that they can be ignored?)
Maybe we should list all the attributes that *must* be present and note any optional ones, if any, that will be copied over?

scipy/optimize/_differentialevolution.py
```diff
@@ -451,6 +464,21 @@ def differential_evolution(func, bounds, args=(), strategy='best1bin',
     >>> result.x, result.fun
     (array([0., 0.]), 4.440892098500626e-16)
 
+    The final polishing step can be customised by providing a callable that
+    mimics the `minimize` interface. The user is responsible for ensuring that
+    the minimizer obeys any bounds and constraints.
```
mdhaber, line 469: Same as above regarding obeying bounds and constraints.

scipy/optimize/_differentialevolution.py
```diff
@@ -451,6 +464,21 @@ def differential_evolution(func, bounds, args=(), strategy='best1bin',
...
+
+    >>> from functools import partial
+    >>> from scipy.optimize import minimize
+    >>> # supply extra parameters to the polishing function using partial
+    >>> polish_func = partial(minimize, method="SLSQP")
+    >>> result = differential_evolution(
+    ...     ackley, bounds, vectorized=True, updating='deferred', rng=1,
+    ...     polish=polish_func
```
mdhaber, line 477: Ideally, the fact that the function is polished would give a better value of the objective function. Is it easy to tweak the example to demonstrate this? Not blocking, though; anyone can make better later.

scipy/optimize/_differentialevolution.py
```diff
@@ -1245,16 +1283,31 @@ def solve(self):
...
-            result = minimize(lambda x:
-                                list(self._mapwrapper(self.func, np.atleast_2d(x)))[0],
-                              np.copy(DE_result.x),
-                              method=polish_method,
-                              bounds=self.limits.T,
-                              constraints=self.constraints)
-
-            self._nfev += result.nfev
+
+            if callable(self.polish):
```
mdhaber: Hmm. Ideally we wouldn't have to branch to *call* the polishing method  - if the user doesn't provide a `polish` callable, we just provide one, like:

```python
if not callable(polish):
    # define `polish` callable

# use `polish` callable
```
Either way, we would use the callable the same way.

I'm not familiar with `_mapwrapper` - when `polish_method` is one of the defaults (`'L-BFGS-B'` or `'trust-constr'`), does that make use of a `Pool`?

scipy/optimize/tests/test__differential_evolution.py
mdhaber: Can you comment one whether/how this test would fail before this PR? (Can't check myself right now.)

mdhaber (commented):
Looking good; just a few thoughts.

scipy/optimize/_differentialevolution.py
andyfaff: I have a blind spot for when one should use single or double backticks.

scipy/optimize/_differentialevolution.py
andyfaff: >  can you comment on why all this is emphasized?

It's new text here. When one creates an `OptimizeResult` there are no attributes that are mandatory during initialisation. So a deficient polishing function could return an object that hasn't filled out `x`, `fun`, etc. Since `differential_evolution` takes the `OptimizeResult` returned by the polisher to flesh out its own result this will cause problems. Using `minimize` as a polishing function is never an issue.

If the user has their own polishing function it could very well enter forbidden zones, i.e. outside bounds and constraints. Since the output from the polisher is then used to create an `OptimizeResult` for `differential_evolution` it'll give the impression that DE didn't manage to find solutions, when really it was the polishing functions's fault. I don't want to be seeing bug issues along the lines of "why did DE give me a result outside the bounds?". It wouldn't be reasonable for the user to assume they can be ignored, but sometimes users have to have things spelled out for them.

scipy/optimize/_differentialevolution.py
andyfaff: TBH the only reason polishing is used is to give a few extra digits of a relative tolerance. I don't think it's worth providing an example for this.

scipy/optimize/_differentialevolution.py
andyfaff: The code is written how my mind interprets the logic flow, which would be to have a branch. But I can make the change.

`_mapwrapper` can make use of a Pool, if the user specified it. Most of the time it's simply `map`.

Supplying the original function is important here because giving the user full control of the polisher might mean they expect to have the original objective function supplied to it. e.g. if they want to do a separate parallelisation or something.

scipy/optimize/tests/test__differential_evolution.py
andyfaff: It wouldn't fail before this PR because the polisher would be `minimize` and the polishing methods used would automatically have `jac` as an attribute.

mdhaber: Ok, but that means it doesn't really test the functionality added by the PR. I think we need to have a test that does that. A simple modification would be to check that the output with `polish=True` is different from that with a `polish` callable. That could be made stronger by also checking that the output with a `polish` callable agrees with the result when the output of DE is provided manually to the callable as the guess. Stronger tests still would check that the plumbing works properly when there are constraints. Since you're also a maintainer and more active in `optimize` than me, whatever you decide on is fine. But I hope you agree with the need for a test sensitive to the new behavior.

If you'd prefer, I can push the changes.

scipy/optimize/_differentialevolution.py
mdhaber: > So a deficient polishing function could return an object that hasn't filled out x, fun, etc.

Ok, so are the minimum requirements `x`, `fun`, `nfev`, and `jac` (if available)?

> sometimes users have to have things spelled out for them.

Ok, I'll let it go.

scipy/optimize/tests/test__differential_evolution.py
andyfaff: This added test does check the functionality added by the PR. The result from the DE call on lin 1619 is done with only 1 DE iteration. Without polishing that will be nowhere close to the optimum, with polishing it will be very close to the optimum, i.e. within an rtol of 1e-6.

Thus test checks that a polish callable can be passed in, and that polishing occurs correctly. There are already other tests that check that `polish=True` works.

Lines below here check that bounds and constraints are passed to the polishing function, and that an `OptimizeResult` is returned.

scipy/optimize/_differentialevolution.py
mdhaber: K. Maybe later I'll apply super small tweaks to demonstrate it. Currently the example shows how the option is used. I figure if it's no more complex to show the *effect* it has, it's just better.

scipy/optimize/_differentialevolution.py
andyfaff: x, fun, nfev. Having jac is good as well.

scipy/optimize/tests/test__differential_evolution.py
mdhaber: I know that the test exercises the code in the PR; i.e. it satisfies code coverage. I just don't think it's a very strong test of the new behavior if it also passes in main. I hope I'm not being picky : / This sort of thing comes up all the time in reviews. See e.g. https://github.com/scipy/scipy/pull/23557#discussion_r2326182476.

I won't try to block merge either way, but I'll suggest a tweak. With something like it committed, I'm happy to approve and press the button myself.

andyfaff:
If you'd like to add further tests you're welcome.

j-bowhay:
Not blocking but it's not entirely clear to me that we should make our API all singing and dancing vs just showing how the user can polish their result themselves. Is it really more convenient to pass a callable as an argument vs calling it themselves?

andyfaff:
I think it's sufficiently useful to have polishing incorporated within `differential_evolution` because it's a frequent pattern for end users. It therefore reduces boilerplate downstream.
THere's been several requests over the years for people to have finer control over how the polishing gets done. This seems to be a cleaner option than adding `minimizer_kwargs` which is overkill.

mdhaber:
Sorry @andyfaff, I meant to do that. Busy week : / I can probably take a look/merge Friday.

@j-bowhay re: https://github.com/scipy/scipy/pull/23538#issuecomment-3265402352, I said about the same in https://discuss.scientific-python.org/t/polishing-function-for-optimize-differential-evolution/2083/3 and https://github.com/scipy/scipy/issues/13561#issuecomment-1272452537. The problem was that the proposal to deprecate `polish` stalled out before. The idea is that this is a compromise between what was requested and what might have been best (user simply calls `minimize`).

scipy/optimize/_differentialevolution.py
```diff
@@ -1284,27 +1285,27 @@ def solve(self):
...
+                _f,
+                np.copy(DE_result.x),
+                bounds=Bounds(lb=self.limits[0], ub=self.limits[1]),
+                constraints=self.constraints
+            )
+
+            if not isinstance(result, OptimizeResult):
+                raise ValueError(
+                    "The result from a user defined polishing function "
+                     "should return an OptimizeResult."
```
mdhaber, line 1307:
```suggestion
                    "The result from a user defined polishing function "
                    "should return an OptimizeResult."
```

mdhaber (approved):
Thanks @andyfaff. I'll fix the errors in CI then merge.

scipy/optimize/_differentialevolution.py
```diff
@@ -3,6 +3,7 @@
 Added by Andrew Nelson 2014
 """
 import warnings
+from functools import partial
```
mdhaber:
```suggestion
```

scipy/optimize/_differentialevolution.py
```diff
@@ -1245,16 +1283,31 @@ def solve(self):
...
-                              method=polish_method,
-                              bounds=self.limits.T,
-                              constraints=self.constraints)
-
-            self._nfev += result.nfev
+
+            pf = self.polish
+            _f = self.original_func
+            if not callable(pf):
+                pf = partial(minimize, method=polish_method)
```
WarrenWeckesser, line 1290: `partial` has not been imported, so this fails.  How did the tests pass?

scipy/optimize/tests/test__differential_evolution.py
```diff
@@ -4,13 +4,14 @@
 from multiprocessing.dummy import Pool as ThreadPool
 import platform
 import warnings
+from functools import partial
```
WarrenWeckesser, line 7: `partial` is not used in this file. Maybe you added this line to the wrong file?

mdhaber:
Yup, I'll fix it right now. I did a `skip ci` after "fixing" a lint error, and it looks like I removed it from the wrong file. gh-23538.

PR #23541, opened 2025-09-02
ENH: linalg: add symmetric/hermitian modes to `linalg.inv`; add `lower={True, False}` argument

ev-br:
#### Reference issue

a part of https://github.com/scipy/scipy/issues/23141

Add specialized inversion routines for symmetric and hermitian matrices to `linalg.inv`.

When an argument of `inv` is auto-detected as symmetric or hermitian, we first try the Cholesky based inversion. If that fails, it means that the matrix is symmetric but not positive definite. On main, we fall back to the generic solver; 
In this PR, we fall back to the new "sym" or "her" routines ,backed by LAPACK's   `?sytrf/?sycon/?sytri` and `?hetrf/?hecon/?hetri`.

Using explicit `assume_a="sym"` or `"her"` bypasses the autodetection, similar to other modes.

#### Additional information

For "pos", "sym" and "her" modes, we need to be able to select the upper or lower triangle. Given the long discussion in https://github.com/scipy/scipy/pull/23350, this PR adds a new `lower={True/False}` argument to `inv`  and _removes the "pos lower" mode_. Removing the mode has no backwards compatible concerns because it was only added in scipy 1.17 which has not been released yet (in https://github.com/scipy/scipy/pull/22924).

Also this PR is based off https://github.com/scipy/scipy/pull/23362, if only to minimize code churn. If that PR lands before this one, I'll rebase.

There are three separate commits: one MAINT for gh-23362, one ENH for the new `lower` argument, and the third one adds the new "sym/her" modes.

### DIFF
--- a/scipy/linalg/_basic.py
+++ b/scipy/linalg/_basic.py
@@ -1410,15 +1410,33 @@ def inv(a, overwrite_a=False, check_finite=True, assume_a=None):
     }[assume_a]
 
     # a1 is well behaved, invert it.
-    result = _batched_linalg._inv(a1, structure, overwrite_a)
-    inv_a, is_ill_cond, is_singular, info = result
+    inv_a, err_lst = _batched_linalg._inv(a1, structure, overwrite_a)
 
-    if info < 0:
-        raise ValueError("Internal LAPACK error.")
-    if is_singular:
-        raise LinAlgError("A singular matrix detected")
-    if is_ill_cond:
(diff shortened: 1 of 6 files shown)


ev-br:
To be on the safe side, ran the full battery of the `[wheel build]` tests, which all passed (https://github.com/scipy/scipy/actions/workflows/wheels.yml).

ev-br:
Superseded by gh-23547

PR #23547, opened 2025-09-03
`linalg.solve`: move "sym"/"her" solvers to C

ev-br:
#### Reference issue

Is based off and includes https://github.com/scipy/scipy/pull/23541 <s>and https://github.com/scipy/scipy/pull/23362</s>

Move `"sym"` and `"her"` modes of `linalg.solve` to C. The method is similar to that of `inv`, gh-23541, --- we first try Cholesky solve, and if it fails, fall back to sym/her instead of a general solve. 

#### Additional information

This PR is the third PR is the "stack", with two previous items being
- refactor of error reporting from C to user, gh-23362
- implementation of "sym"/"her" modes for `linalg.inv`, gh-23541

If either of these lands, I'll rebase this one. Or can land this one in one go instead of going PR-by-PR. Reviewer's preferernce :-).

### DIFF
--- a/scipy/linalg/src/_batched_linalg_module.cc
+++ b/scipy/linalg/src/_batched_linalg_module.cc
@@ -6,7 +6,6 @@
 #include "_common_array_utils.hh"
 
 
-#define PYERR(errobj,message) {PyErr_SetString(errobj,message); return NULL;}
 static PyObject* _linalg_inv_error;
 
 
@@ -45,7 +44,8 @@ _linalg_inv(PyObject* Py_UNUSED(dummy), PyObject* args) {
     npy_intp* shape = PyArray_SHAPE(ap_Am);      // Array shape
     npy_intp n = shape[ndim - 1];                // Slice size
     if (n != shape[ndim - 2]) {
-        PYERR(PyExc_ValueError, "Last two dimensions of the input must be the same.")
+        PyErr_SetString(PyExc_ValueError, "Last two dimensions of the input must be the same.");
+        return NULL;
     }
 
     overwrite_a = 0; // TODO: enable it
@@ -79,13 +79,15 @@ _linalg_inv(PyObject* Py_UNUSED(dummy), PyObject* args) {
             info = _inverse<npy_complex128>(ap_Am, (npy_complex128 *)buf, structure, lower, overwrite_a, vec_status);
             break;
         default:
-            PYERR(PyExc_RuntimeError, "Unknown array type.")
+            PyErr_SetString(PyExc_RuntimeError, "Unknown array type.");
+            return NULL;
     }
 
     if (info < 0) {
         // Either OOM or internal LAPACK error.
         Py_DECREF(ap_Ainv);
-        PYERR(PyExc_RuntimeError, "Memory error in scipy.linalg.inv.")
+        PyErr_SetString(PyExc_RuntimeError, "Memory error in scipy.linalg.inv.");
+        return NULL;
     }
     PyObject *ret_lst = convert_vec_status(vec_status);
 
@@ -127,7 +129,8 @@ _linalg_solve(PyObject* Py_UNUSED(dummy), PyObject* args) {
     int ndim = PyArray_NDIM(ap_Am);
     npy_intp* shape = PyArray_SHAPE(ap_Am);
     if ((ndim < 2) || (shape[ndim - 1] != shape[ndim - 2])) {
-        PYERR(PyExc_ValueError, "Last two dimensions of `a` must be the same.")
+        PyErr_SetString(PyExc_ValueError, "Last two dimensions of `a` must be the same.");
+        return NULL;
     }
 
     // At the python call site, 
@@ -171,13 +174,15 @@ _linalg_solve(PyObject* Py_UNUSED(dummy), PyObject* args) {
             info = _solve<npy_complex128>(ap_Am, ap_b, (npy_complex128 *)buf, structure, lower, transposed, overwrite_a, vec_status);
             break;
         default:
-            PYERR(PyExc_RuntimeError, "Unknown array type.")
+            PyErr_SetString(PyExc_RuntimeError, "Unknown array type.");
+            return NULL;
     }
 
     if (info < 0) {
-        // Either OOM or internal LAPACK error.
+        // Either OOM error or requiested lwork too large.
         Py_DECREF(ap_x);
-        PYERR(PyExc_RuntimeError, "Memory error in scipy.linalg.solve.")
+        PyErr_SetString(PyExc_MemoryError, "Memory error in scipy.linalg.solve.");
+        return NULL;
     }
     PyObject *ret_lst = convert_vec_status(vec_status);
 
--- a/scipy/linalg/src/_common_array_utils.hh
+++ b/scipy/linalg/src/_common_array_utils.hh
@@ -398,6 +398,33 @@ GEN_HECON_CZ(c, npy_complex64, float, float)
 GEN_HECON_CZ(z, npy_complex128, double, double)
 
 
+#define GEN_SYTRS(PREFIX, TYPE) \
+inline void \
+sytrs(char* uplo, CBLAS_INT* n, CBLAS_INT *nrhs, TYPE *a, CBLAS_INT *lda, CBLAS_INT *ipiv, TYPE *b, CBLAS_INT *ldb, CBLAS_INT* info) \
+{ \
+    BLAS_FUNC(PREFIX ## sytrs)(uplo, n, nrhs, a, lda, ipiv, b, ldb, info); \
+};
+
+GEN_SYTRS(s, float)
+GEN_SYTRS(d, double)
+GEN_SYTRS(c, npy_complex64)
+GEN_SYTRS(z, npy_complex128)
+
+
+// dispatch to sSYtrs for "float hermitian"
+#define GEN_HETRS(PREFIX, L_PREFIX, TYPE) \
+inline void \
+hetrs(char *uplo, CBLAS_INT *n, CBLAS_INT *nrhs, TYPE *a, CBLAS_INT *lda, CBLAS_INT *ipiv, TYPE *b, CBLAS_INT *ldb, CBLAS_INT* info) \
+{ \
+    BLAS_FUNC(PREFIX ## L_PREFIX ## trs)(uplo, n, nrhs, a, lda, ipiv, b, ldb, info); \
+};
+
+GEN_HETRS(s, sy, float)
+GEN_HETRS(d, sy, double)
+GEN_HETRS(c, he, npy_complex64)
+GEN_HETRS(z, he, npy_complex128)
+
+
 // Structure tags; python side maps assume_a strings to these values
 enum St : Py_ssize_t
 {
--- a/scipy/linalg/src/_common_array_utils.hh
+++ b/scipy/linalg/src/_common_array_utils.hh
@@ -488,7 +488,9 @@ CBLAS_INT _calc_lwork(T _lwrk, double fudge_factor=1.0) {
     using real_type = typename type_traits<T>::real_type;
 
     real_type value = real_part(_lwrk) * fudge_factor;
-    if(type_traits<T>::is_single_precision) {
+    if((std::is_same<real_type, float>::value) ||
+       (std::is_same<real_type, npy_complex64>::value)
+    ) {
         // Single-precision routine -- take next fp value to work
         // around possible truncation in LAPACK code
         value = std::nextafter(value, std::numeric_limits<real_type>::infinity());
--- a/scipy/linalg/src/_common_array_utils.hh
+++ b/scipy/linalg/src/_common_array_utils.hh
@@ -4,6 +4,7 @@
 #ifndef _SCIPY_COMMON_ARRAY_UTILS_H
 #define _SCIPY_COMMON_ARRAY_UTILS_H
 #include "Python.h"
+#include <tuple>
 #include "numpy/npy_math.h"
 #include "npy_cblas.h"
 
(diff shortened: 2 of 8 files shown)


ev-br:
CI failures in sparse/free-threaded, stats and interpolate do not seem related.

scipy/linalg/tests/test_basic.py
```diff
@@ -951,8 +951,8 @@ def test_structure_detection(self, dtype, assume_a, nrhs, transposed,
         elif assume_a in {'hermitian', 'her'}:
             A = A + A.conj().T
         elif assume_a in {'positive definite', 'pos'}:
-            A = A + A.T
             A += np.diag(A.sum(axis=1))
+            A = A + A.T.conj()
```
tylerjereddy: It looks like I can add these two new test cases on `main`, and make this same change to their expected value, and have the test pass without the C changes here.

So, this is an unrelated change in the diff here then, where the unused expected value was wrong, but the correction of the wrong expected value is independent of the low-level changes in the PR?

scipy/linalg/tests/test_basic.py
```diff
@@ -1397,10 +1400,58 @@ def test_posdef_not_posdef(self):
...
+        assert_allclose(b_inv0, b_inv1, atol=1e-15)
+
+        # check that the "other" triangle is not referenced
+        mask = np.where(1 - np.tri(*a.shape, -1) == 0, np.nan, 1)
+        b_inv2 = inv(b*mask, check_finite=False, assume_a="her", lower=False)
+        assert_allclose(b_inv2, b_inv0, atol=1e-15)
+
+        # repeat with the upper triangle
+        b_inv3 = inv(b*mask.T, check_finite=False, assume_a="her", lower=True)
+        assert_allclose(b_inv3, b_inv0, atol=1e-15)
```
tylerjereddy, line 1438: This test and the one above it are *very* similar. I suppose there are subtle differences that makes parametrization annoying.

Beyond that, the testing changes look pretty "safe" in the sense of adding new test cases but not changing old ones, modulo the one comment I made about fixing an old test.

scipy/linalg/src/_npymath.hh
```diff
@@ -56,25 +56,29 @@ template<> struct type_traits<float> {
...
     static constexpr int typenum = NPY_COMPLEX64;
     static constexpr bool is_complex = true;
+    static constexpr bool is_single_precision = true;
 };
 template<> struct type_traits<npy_complex128> {
     using real_type = double;
     using value_type = std::complex<double>;
     static constexpr int typenum = NPY_COMPLEX128;
     static constexpr bool is_complex = true;
+    static constexpr bool is_single_precision = false;
```
tylerjereddy: Isn't this extra structure member adding information that can already be obtained through the `real_type`?

The diff below the fold passes the full suite locally for example.

<details>

```diff
--- a/scipy/linalg/src/_common_array_utils.hh
+++ b/scipy/linalg/src/_common_array_utils.hh
@@ -488,7 +488,7 @@ CBLAS_INT _calc_lwork(T _lwrk, double fudge_factor=1.0) {
     using real_type = typename type_traits<T>::real_type;
 
     real_type value = real_part(_lwrk) * fudge_factor;
-    if(type_traits<T>::is_single_precision) {
+    if(std::is_same<real_type, float>::value) {
         // Single-precision routine -- take next fp value to work
         // around possible truncation in LAPACK code
         value = std::nextafter(value, std::numeric_limits<real_type>::infinity());
diff --git a/scipy/linalg/src/_npymath.hh b/scipy/linalg/src/_npymath.hh
index c025e0c764..076978e11a 100644
--- a/scipy/linalg/src/_npymath.hh
+++ b/scipy/linalg/src/_npymath.hh
@@ -56,7 +56,6 @@ template<> struct type_traits<float> {
     using value_type = float;
     static constexpr int typenum = NPY_FLOAT;
     static constexpr bool is_complex = false;
-    static constexpr bool is_single_precision = true;
 
 };
 template<> struct type_traits<double> {
@@ -64,21 +63,18 @@ template<> struct type_traits<double> {
     using value_type = double;
     static constexpr int typenum = NPY_DOUBLE;
     static constexpr bool is_complex = false;
-    static constexpr bool is_single_precision = false;
 };
 template<> struct type_traits<npy_complex64> {
     using real_type = float;
     using value_type = std::complex<float>;
     static constexpr int typenum = NPY_COMPLEX64;
     static constexpr bool is_complex = true;
-    static constexpr bool is_single_precision = true;
 };
 template<> struct type_traits<npy_complex128> {
     using real_type = double;
     using value_type = std::complex<double>;
     static constexpr int typenum = NPY_COMPLEX128;
     static constexpr bool is_complex = true;
-    static constexpr bool is_single_precision = false;
 };

```

</details>

The testsuite also didn't seem to care if I used `double` or `float` for the conditional, maybe I need to run the `slow` tests or something for that, or maybe we don't really have a check that is sensitive for it.

tylerjereddy (commented):
Can you expand the explanation of the benefits/gains here a bit? The user gets a bit more custom control to short-circuit to their chosen matrix structure/solver if known, and the fall-back behavior is also more resilient? Is that a good summary?

Are any performance changes expected here?

When I ran `asv continuous -E virtualenv -e -b "time_solve" main solve_sy_he` on x86_64 Linux it reported some changes, but could just be noise.

<details>

```
| Change   | Before [6151470b] <main>   | After [24c9e7b7] <solve_sy_he>   |   Ratio | Benchmark (Parameter)                               |
|----------|----------------------------|----------------------------------|---------|-----------------------------------------------------|
| +        | 3.10±0.05ms                | 3.63±0.3ms                       |    1.17 | sparse_linalg_solve.Bench.time_solve(100, 'minres') |
| -        | 16.7±0.09ms                | 14.3±0.3ms                       |    0.86 | sparse_linalg_solve.Bench.time_solve(100, 'tfqmr')  |
| -        | 249±1μs                    | 199±1μs                          |    0.8  | linalg.Bench.time_solve(100, 'contig', 'scipy')     |

PERFORMANCE DECREASED.

```

</details>

scipy/linalg/tests/test_basic.py
ev-br: Yes, thanks. This change is unrelated and should not be here. Removed it from here, will investigate and submit separately.

scipy/linalg/tests/test_basic.py
ev-br: Thanks, deduplicated, at a cost of a minor `if-else` branching in a parametrized test.

scipy/linalg/src/_npymath.hh
ev-br: Wanted to avoid a mouthful of `if((std::is_same<real_type, float>::value) || (std::is_same<real_type, npy_complex64>::value)`. No argument from me, changed.

ev-br:
Thanks for the review @tylerjereddy !

> expand the explanation of the benefits/gains here a bit? The user gets a bit more custom control to short-circuit to their chosen matrix structure/solver if known, and the fall-back behavior is also more resilient? Is that a good summary?

Yes, pretty much. I'd add a bit about performance, too. Two sources of perf improvements:
- looping over slices in a batch is in C (on main, it loops in python)
- symmetric solve is potentially faster than the general solve (this is currently not necessarily there, meaning there's a budget for further perf optimization)

> Are any performance changes expected here?

The last commit added a small benchmark, which on my laptop shows up to a 3-5x improvement in some scenarios (details under the fold). If I rerun it several times on a non-empty machine, some runs do show slowdowns of ~5-10% for a "general" solver, similar to what you saw for the existing benchmark.  So maybe these 10-20% are indeed noise?

One other thing the added benchmark shows is that scipy `solve` is still quite a bit slower than `numpy`'s, even though both are now mostly in C. Not sure how much of it is that the condition number estimation, which we do and numpy does not, is very much not free. Might warrant some profiling, preferably in a separate PR.

<details>

```
$ spin bench --compare main -t BatchedSolveBench.time_solve
$ cd benchmarks
$ asv continuous --factor 1.05 --bench BatchedSolveBench.time_solve 00791095931961235a4cd4b9cec22680a9156c1b cf214abefb1e881f574507fefb93b4d91ea05409
Couldn't load asv.plugins._mamba_helpers because
No module named 'libmambapy'
· Creating environments
· Discovering benchmarks.
·· Uninstalling from virtualenv-py3.12-Cython-meson-python-numpy-pybind11-pytest-pythran
·· Installing cf214abe <solve_sy_he> into virtualenv-py3.12-Cython-meson-python-numpy-pybind11-pytest-pythran..
· Running 2 total benchmarks (2 commits * 1 environments * 1 benchmarks)
[ 0.00%] · For scipy commit 00791095 <main> (round 1/2):
[ 0.00%] ·· Building for virtualenv-py3.12-Cython-meson-python-numpy-pybind11-pytest-pythran...
[ 0.00%] ·· Benchmarking virtualenv-py3.12-Cython-meson-python-numpy-pybind11-pytest-pythran
[25.00%] ··· Running (linalg.BatchedSolveBench.time_solve--).
[25.00%] · For scipy commit cf214abe <solve_sy_he> (round 1/2):
[25.00%] ·· Building for virtualenv-py3.12-Cython-meson-python-numpy-pybind11-pytest-pythran...
[25.00%] ·· Benchmarking virtualenv-py3.12-Cython-meson-python-numpy-pybind11-pytest-pythran
[50.00%] ··· Running (linalg.BatchedSolveBench.time_solve--).
[50.00%] · For scipy commit cf214abe <solve_sy_he> (round 2/2):
[75.00%] ··· linalg.BatchedSolveBench.time_solve                                                                                                                           ok
[75.00%] ··· =============== ============= ============= ============= ============= ============= =============
             --                                               structure / module                                
             --------------- -----------------------------------------------------------------------------------
                  shape       gen / scipy   gen / numpy   pos / scipy   pos / numpy   sym / scipy   sym / numpy 
             =============== ============= ============= ============= ============= ============= =============
              (100, 10, 10)     531±3μs      104±0.6μs      461±3μs       113±1μs       832±10μs      104±1μs   
              (100, 20, 20)   1.15±0.02ms     327±3μs       887±10μs      387±3μs     1.92±0.02ms     334±10μs  
                (100, 100)      182±1μs      73.2±0.9μs     147±1μs       72.0±2μs      216±3μs      73.6±0.4μs 
             =============== ============= ============= ============= ============= ============= =============

[75.00%] · For scipy commit 00791095 <main> (round 2/2):
[100.00%] ··· linalg.BatchedSolveBench.time_solve                                                                                                                           ok
[100.00%] ··· =============== ============= ============= ============= ============= ============= =============
              --                                               structure / module                                
              --------------- -----------------------------------------------------------------------------------
                   shape       gen / scipy   gen / numpy   pos / scipy   pos / numpy   sym / scipy   sym / numpy 
              =============== ============= ============= ============= ============= ============= =============
               (100, 10, 10)     540±3μs       103±4μs       476±4μs       113±1μs     3.99±0.08ms    101±0.1μs  
               (100, 20, 20)   1.17±0.03ms     334±9μs       900±10μs      393±8μs     5.05±0.07ms     334±8μs   
                 (100, 100)     182±0.9μs      72.2±2μs      145±3μs      71.2±0.6μs     216±3μs       71.7±3μs  
              =============== ============= ============= ============= ============= ============= =============

| Change   | Before [00791095] <main>   | After [cf214abe] <solve_sy_he>   |   Ratio | Benchmark (Parameter)                                              |
|----------|----------------------------|----------------------------------|---------|--------------------------------------------------------------------|
| -        | 5.05±0.07ms                | 1.92±0.02ms                      |    0.38 | linalg.BatchedSolveBench.time_solve((100, 20, 20), 'sym', 'scipy') |
| -        | 3.99±0.08ms                | 832±10μs                         |    0.21 | linalg.BatchedSolveBench.time_solve((100, 10, 10), 'sym', 'scipy') |

PERFORMANCE INCREASED.
```

</details>

ev-br:
Forgot to respond to

> The testsuite also didn't seem to care if I used double or float for the conditional, maybe I need to run the slow tests or something for that, or maybe we don't really have a check that is sensitive for it.

Unfortunately the latter IIUC. The whole business of second-guessing LAPACK's `lwork` feels like a collection of hacks for _some_ versions of _some_ LAPACK libs, and we've no idea if any of the workarounds relate to computer museum items. What I did so far was to transcribe the existing logic from python to C:
- [some LAPACK versions (<= 3.5.0 at least)](https://github.com/scipy/scipy/blob/main/scipy/linalg/lapack.py#L1032), and
- [a curious segfault](https://github.com/scipy/scipy/blob/v1.15.2/scipy/linalg/_basic.py#L1154)

So I guess we either keep these defensive workarounds indefinitely, or rip them out and see if somebody complains. So far this PR hesitates to do the latter. WDYT?

scipy/linalg/src/_batched_linalg_module.cc
```diff
@@ -171,13 +174,15 @@ _linalg_solve(PyObject* Py_UNUSED(dummy), PyObject* args) {
...
             break;
         default:
-            PYERR(PyExc_RuntimeError, "Unknown array type.")
+            PyErr_SetString(PyExc_RuntimeError, "Unknown array type.");
+            return NULL;
     }
 
     if (info < 0) {
-        // Either OOM or internal LAPACK error.
+        // Either OOM error or requiested lwork too large.
```
tylerjereddy: minor typo, not worth flushing CI for

tylerjereddy:
The new benchmark seemed to be net positive for me as well, though still some slowdowns (but they are less dramatic than the speedups) via `asv continuous -E virtualenv -e -b "BatchedSolveBench" main solve_sy_he`.

<details>

```
| Change   | Before [7bbabe57] <main>   | After [a7e27eb9] <solve_sy_he>   |   Ratio | Benchmark (Parameter)                                              |
|----------|----------------------------|----------------------------------|---------|--------------------------------------------------------------------|
| +        | 198±0.8μs                  | 250±3μs                          |    1.26 | linalg.BatchedSolveBench.time_solve((100, 100), 'gen', 'scipy')    |
| +        | 67.7±0.3μs                 | 78.1±0.2μs                       |    1.15 | linalg.BatchedSolveBench.time_solve((100, 100), 'sym', 'scipy')    |
| -        | 1.68±0ms                   | 481±10μs                         |    0.29 | linalg.BatchedSolveBench.time_solve((100, 20, 20), 'sym', 'scipy') |
| -        | 1.40±0ms                   | 205±0.5μs                        |    0.15 | linalg.BatchedSolveBench.time_solve((100, 10, 10), 'sym', 'scipy') |

PERFORMANCE DECREASED.
```

</details>

Some of the stuff in `scipy/linalg/src/_common_array_utils.hh` I didn't fully follow, and it is a bit annoying when I can delete random bits of code and the tests still pass, but I'll trust your judgement/the legacy shims are correct there since we're digging down to linalg backend territory. There was a minor typo `requiested` in the diff, not worth flushing CI for again.

I re-read the full diff and my comments were addressed. I'm not the best maintainer to review low-level linalg but the status of the testing changes (no old test cases with different behavior) and net performance/functionality improvement seem worth a try.

This has been open for almost 3 months now without comments--if a `linalg` regular has an issue with it during i.e., the `1.17.0` RC phase we can deal with it then, but I'm happy to give it a try given the potential upsides described/demonstrated.

tylerjereddy:
thanks Evgeni

ev-br:
Thanks Tyler! 

> and it is a bit annoying when I can delete random bits of code and the tests still pass

Do you happen to remember still which parts of were these, apart from the defensive "bump lwork just in case" Chesterton fences?

15-25% slowdowns on non-batched workloads look suspicious indeed. It's a bit hard to imagine a noise level of 25%. I made a note at https://github.com/scipy/scipy/issues/23141 to investigate performance.

scipy/linalg/tests/test_basic.py
ev-br: > submit separately.

https://github.com/scipy/scipy/pull/24063

scipy/linalg/src/_common_array_utils.hh
```diff
@@ -477,17 +686,26 @@ bandwidth(T* data, npy_intp n, npy_intp m, npy_intp* lower_band, npy_intp* upper
...
+            all_sym = all_sym && (elem1 == elem2);
+            all_herm = all_herm && (elem1 == std::conj(elem2));
+            if(!(all_sym || all_herm)) {
+                // short-circuit : it's neither symmetric not hermitian
+                return std::make_tuple(false, false);
+            }
         }
     }
-    return true;
+    return std::make_tuple(true, all_sym ? true : false);
```
WarrenWeckesser, line 708: I think this should be

    return std::make_tuple(true, all_sym && !all_herm);

PR #23554, opened 2025-09-04
BUG: signal: Control output dtypes in `abcd_normalize` by adding new …

### LINKED ISSUE #13125: signal.step function behaves differently when system parameters are integers/floats
This issue is about the [`signal.step`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.step.html) function when used with a [`signal.StateSpaceContinuous`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.StateSpace.html) system.  The issue is that the step response can be zero when the parameters are integers.  (I tried to see if this issue has been raised already but couldn't find it.)

```
a = -4
b = 2
c = 1
d = 0
sys = signal.StateSpace(a, b, c, d)
print(sys)
t, y = signal.step(sys)
print(y[0:5])

# Output:
# StateSpaceContinuous(
# array([[-4]]),
# array([[2]]),
# array([[1]]),
# array([[0]]),
# dt: None
# )
# [0 0 0 0 0]
```

Expected response:

```
a = -4.0  # This triggers a to be of type float
b = 2
c = 1
d = 0
sys = signal.StateSpace(a, b, c, d)
print(sys)

# Output:
# StateSpaceContinuous(
# array([[-4.]]),
# array([[2]]),
# array([[1]]),
# array([[0]]),
# dt: None
# )
# [0.         0.03413261 0.06593515 0.09556668 0.1231754 ]
```

The problem is the same with MIMO systems:
```
a = np.array([[0, 1], [0, 0]])
b = np.array([[0], [1]])
c = np.array([[1, 0]])
d = np.array([[0]])
sys = signal.StateSpace(a, b, c, d)
t, y = signal.step(sys)
print(y[0:5])

# Output:
# [0 0 0 0 0]
```

In my view, we should be able to define systems with either floats, ints, or array_like structures of floats/ints and get the same (correct) behaviour.

### LINKED ISSUE #18982: BUG: signal: `lsim` returns incorrect results when the state space matrix `A` has integer type
### Describe your issue.

This problem showed up in a [stackoverflow question](https://stackoverflow.com/questions/76787449/scipys-lsim-not-behaving-as-expected-compared-to-python-control-systems-library).  The output of `lsim` did not match the output produced by another library, and in fact the output was mostly zeros.

Here is a simple example.  The components of the solution should exhibit exponential decay, but instead they immediately jump to zero. And note that solution is an integer array:
```
In [24]: from scipy.signal import lsim

In [25]: A = np.array([[-1, 0], [0, -1]])

In [26]: B = np.array([[0], [1]])

In [27]: C = np.array([0, 0])

In [28]: lsim((A, B, C, None), U=None, T=np.linspace(0, 1, 5), X0=[1, 1])
Out[28]: 
(array([0.  , 0.25, 0.5 , 0.75, 1.  ]),
 array([0, 0, 0, 0, 0]),
 array([[1, 1],
        [0, 0],
        [0, 0],
        [0, 0],
        [0, 0]]))
```

If `A` is converted to `float64`, the result are as expected:

```
In [10]: lsim((A.astype(np.float64), B, C, None), U=None, T=np.linspace(0, 1, 5), X0=[1, 1])
Out[10]: 
(array([0.  , 0.25, 0.5 , 0.75, 1.  ]),
 array([0., 0., 0., 0., 0.]),
 array([[1.        , 1.        ],
        [0.77880078, 0.77880078],
        [0.60653066, 0.60653066],
        [0.47236655, 0.47236655],
        [0.36787944, 0.36787944]]))

```

The source of the problem is here: https://github.com/scipy/scipy/blob/9e3f5fde7295cdba2c06f7e8fcaac34626a96000/scipy/signal/_ltisys.py#L2094

The type of the `xout` array should not be copied directly from the type of `A`.  When `A` is an integer type, `out` should be `float64`.

### Reproducing Code Example

```python
See above.
```

### Error message

```shell
No error, just incorrect output.
```

```shell
In [9]: import sys, scipy, numpy; print(scipy.__version__, numpy.__version__, sys.version_info); scipy.show_config()
1.12.0.dev0+1399.9e3f5fd 1.25.1 sys.version_info(major=3, minor=11, micro=4, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    detection method: pkgconfig
    found: true
    include directory: /usr/include/x86_64-linux-gnu/openblas-pthread/
    lib directory: /usr/lib/x86_64-linux-gnu/openblas-pthread/
    name: openblas
    openblas configuration: USE_64BITINT= DYNAMIC_ARCH=1 DYNAMIC_OLDER=1 NO_CBLAS=
      NO_LAPACK= NO_LAPACKE=1 NO_AFFINITY=1 USE_OPENMP=0 generic MAX_THREADS=64
    pc file directory: /usr/lib/x86_64-linux-gnu/pkgconfig
    version: 0.3.20
  lapack:
    detection method: pkgconfig
    found: true
    include directory: /usr/include/x86_64-linux-gnu/openblas-pthread/
    lib directory: /usr/lib/x86_64-linux-gnu/openblas-pthread/
    name: openblas
    openblas configuration: USE_64BITINT= DYNAMIC_ARCH=1 DYNAMIC_OLDER=1 NO_CBLAS=
      NO_LAPACK= NO_LAPACKE=1 NO_AFFINITY=1 USE_OPENMP=0 generic MAX_THREADS=64
    pc file directory: /usr/lib/x86_64-linux-gnu/pkgconfig
    version: 0.3.20
  pybind11:
    detection method: config-tool
    include directory: unknown
    name: pybind11
    version: 2.10.4
Compilers:
  c:
    commands: cc
    linker: ld.bfd
    name: gcc
    version: 11.3.0
  c++:
    commands: c++
    linker: ld.bfd
    name: gcc
    version: 11.3.0
  cython:
    commands: cython
    linker: cython
    name: cython
    version: 0.29.35
  fortran:
    commands: gfortran
    linker: ld.bfd
    name: gcc
    version: 11.3.0
  pythran:
    include directory: ../../../../../py3.11.4/lib/python3.11/site-packages/pythran
    version: 0.13.1
Machine Information:
  build:
    cpu: x86_64
    endian: little
    family: x86_64
    system: linux
  cross-compiled: false
  host:
    cpu: x86_64
    endian: little
    family: x86_64
    system: linux
Python Information:
  path: /home/warren/py3.11.4/bin/python3.11
  version: '3.11'
```
```

### PR DESCRIPTION
**Pull request by DietBru**
With this PR `signal.abcd_normalize` gains the capability of controlling the dtype of the returned arrays through the new keyword parameter ~~`cast_to`~~ `dtype`. The new default behavior is to make the output dtype float64 / complex128 (depending on input dtypes). Alternatively, either a specific dtype can be provided or type conversion can be turned off.

The idea behind adding a new keyword parameter was to make it clearly visible what `abcd_normalize` is actually doing.

Additionally:
* `signal.ss2tf`, `signal.StateSpace` now convert parameters to float64 / complex128 due to relying on `abcd_normalize`.
* ~~The initializer of class `signal.StateSpace` gained the keyword `cast_to`, which is passed to `abcd_normalize` for normalizing matrix parameters. Hence, those are converted to floating by default.~~
* Unit tests were added to ensure 100% coverage.
* Marginal docstr improvements.
* Supersedes PR #13235 from 2020: The `abcd_normalize` code diverged too much for making sensible adaptions in that PR.
* Closes #13125.
* Closes #18982.
* Closes #19629.

### DIFF
--- a/scipy/signal/_delegators.py
+++ b/scipy/signal/_delegators.py
@@ -58,8 +58,7 @@ def _skip_if_poly1d(arg):
 
 ###################
 
-def abcd_normalize_signature(A=None, B=None, C=None, D=None, *, dtype=None,
-                             device=None):
+def abcd_normalize_signature(A=None, B=None, C=None, D=None, *, dtype=None):
     return array_namespace(A, B, C, D)
 
 def argrelextrema_signature(data, *args, **kwds):
--- a/scipy/signal/_delegators.py
+++ b/scipy/signal/_delegators.py
@@ -58,8 +58,8 @@ def _skip_if_poly1d(arg):
 
 ###################
 
-def abcd_normalize_signature(A=None, B=None, C=None, D=None, *, dtype=None, xp=None,
-                   device=None):
+def abcd_normalize_signature(A=None, B=None, C=None, D=None, *, dtype=None,
+                             device=None):
     return array_namespace(A, B, C, D)
 
 def argrelextrema_signature(data, *args, **kwds):
--- a/scipy/signal/_delegators.py
+++ b/scipy/signal/_delegators.py
@@ -58,6 +58,10 @@ def _skip_if_poly1d(arg):
 
 ###################
 
+def abcd_normalize_signature(A=None, B=None, C=None, D=None, *, dtype=None, xp=None,
+                   device=None):
+    return array_namespace(A, B, C, D)
+
 def argrelextrema_signature(data, *args, **kwds):
     return array_namespace(data)
 
--- a/scipy/signal/_delegators.py
+++ b/scipy/signal/_delegators.py
@@ -58,10 +58,6 @@ def _skip_if_poly1d(arg):
 
 ###################
 
-def abcd_normalize_signature(A=None, B=None, C=None, D=None):
-    return array_namespace(A, B, C, D)
-
-
 def argrelextrema_signature(data, *args, **kwds):
     return array_namespace(data)
 
--- a/scipy/signal/_lti_conversion.py
+++ b/scipy/signal/_lti_conversion.py
@@ -8,7 +8,7 @@ from numpy import (r_, eye, atleast_2d, poly, dot,
                    asarray, zeros, array, outer)
 from scipy import linalg
 
-from scipy._lib._array_api import array_namespace, xp_capabilities, xp_size
+from scipy._lib._array_api import array_namespace, xp_size
 import scipy._lib.array_api_extra as xpx
 from ._filter_design import tf2zpk, zpk2tf, normalize
 
@@ -114,7 +114,6 @@ def tf2ss(num, den):
     return A, B, C, D
 
 
-@xp_capabilities()
 def abcd_normalize(A=None, B=None, C=None, D=None, *, dtype=None):
     r"""Check state-space matrices compatibility and ensure they are 2d arrays.
 
--- a/scipy/signal/_lti_conversion.py
+++ b/scipy/signal/_lti_conversion.py
@@ -214,10 +214,10 @@ def abcd_normalize(A=None, B=None, C=None, D=None, *, dtype=None, xp=None,
     >>> print(f" AA: {AA.dtype}, BB: {BB.dtype}\n CC: {CC.dtype}, DD: {DD.dtype}")
      AA: float64, BB: float64
      CC: float64, DD: float64
-    >>> AA, BB, CC, DD = abcd_normalize(A=A, D=D, dtype=np.int32)  # Explicit dtype
+    >>> AA, BB, CC, DD = abcd_normalize(A=A, D=D, dtype=np.float32)  # Explicit dtype
     >>> print(f" AA: {AA.dtype}, BB: {BB.dtype}\n CC: {CC.dtype}, DD: {DD.dtype}")
-     AA: int32, BB: int32
-     CC: int32, DD: int32
+     AA: float32, BB: float32
+     CC: float32, DD: float32
     """
     if A is None and B is None and C is None:
         raise ValueError("Dimension n is undefined for parameters A = B = C = None!")
--- a/scipy/signal/_lti_conversion.py
+++ b/scipy/signal/_lti_conversion.py
@@ -115,8 +115,7 @@ def tf2ss(num, den):
 
 
 @xp_capabilities()
-def abcd_normalize(A=None, B=None, C=None, D=None, *, dtype=None, xp=None,
-                   device=None):
+def abcd_normalize(A=None, B=None, C=None, D=None, *, dtype=None, device=None):
     r"""Check state-space matrices compatibility and ensure they are 2d arrays.
 
     First, the input matrices are converted into two-dimensional arrays with
@@ -146,14 +145,9 @@ def abcd_normalize(A=None, B=None, C=None, D=None, *, dtype=None, xp=None,
             With this new parameter, all return values have identical dtypes.
             In previous versions the dtype of the input was preserved.
 
-    xp : array_namespace, optional
-        The namespace for the return array. When ``None`` (default), then it is derived
-        from the parameters `A`, `B`, `C`, `D`. It should be compatible with the array
-        API standard or supported by array-api-compat. ``numpy`` is used by if
-        ``None`` and the parameters are not arrays.
-    device : any
-        Optional device specification for created matrices. Should match one of the
-        supported device specifications in ``xp``.
+    device : device, optional
+        Optional device specification for returened matrices.
+        Only applicabe when using an alterantive backend.
 
     Returns
     -------
@@ -226,7 +220,7 @@ def abcd_normalize(A=None, B=None, C=None, D=None, *, dtype=None, xp=None,
     if C is None and D is None:
         raise ValueError("Dimension q is undefined for parameters C = D = None!")
 
-    xp = xp or array_namespace(A, B, C, D)
+    xp = array_namespace(A, B, C, D)
 
     # convert inputs into 2d arrays (zero-size 2d array if None):
     A, B, C, D = (xpx.atleast_nd(xp.asarray(M_, device=device), ndim=2, xp=xp)
--- a/scipy/signal/_lti_conversion.py
+++ b/scipy/signal/_lti_conversion.py
@@ -8,7 +8,7 @@ from numpy import (r_, eye, atleast_2d, poly, dot,
                    asarray, zeros, array, outer)
 from scipy import linalg
 
-from scipy._lib._array_api import array_namespace, xp_size
+from scipy._lib._array_api import array_namespace, xp_capabilities, xp_size
 import scipy._lib.array_api_extra as xpx
 from ._filter_design import tf2zpk, zpk2tf, normalize
 
@@ -114,43 +114,65 @@ def tf2ss(num, den):
     return A, B, C, D
 
 
-def abcd_normalize(A=None, B=None, C=None, D=None):
-    """Check state-space matrices compatibility and ensure they are 2d arrays.
+@xp_capabilities()
+def abcd_normalize(A=None, B=None, C=None, D=None, *, dtype=None, xp=None,
+                   device=None):
+    r"""Check state-space matrices compatibility and ensure they are 2d arrays.
 
-    Converts input matrices into two-dimensional arrays as needed. Then the dimensions
-    n, q, p are determined by investigating the non-zero entries of the array shapes.
-    If a parameter is ``None``, or has shape (0, 0), it is set to a
-    zero-array of compatible shape. Finally, it is verified that all parameter shapes
-    are compatible to each other. If that fails, a ``ValueError`` is raised. Note that
-    the dimensions n, q, p are allowed to be zero.
+    First, the input matrices are converted into two-dimensional arrays with
+    appropriate dtype as needed. Then the dimensions n, q, p are determined by
+    investigating the array shapes. If an input is ``None``, or has size zero, it is
+    set to an array of zeros of compatible shape. Finally, it is verified that all
+    parameter shapes are compatible with each other. If that fails, a ``ValueError`` is
+    raised. Note that the dimensions n, q, p are allowed to be zero.
 
     Parameters
     ----------
-    A: array_like, optional
+    A : array_like, optional
         Two-dimensional array of shape (n, n).
-    B: array_like, optional
+    B : array_like, optional
         Two-dimensional array of shape (n, p).
-    C: array_like, optional
+    C : array_like, optional
         Two-dimensional array of shape (q, n).
-    D: array_like, optional
+    D : array_like, optional
         Two-dimensional array of shape (q, p).
+    dtype : dtype | None, optional
+        Cast all matrices to the specified dtype. If set to ``None`` (default), their
+        dtypes will be "complex128" if any of the matrices are complex-valued.
+        Otherwise, they will be of the type "float64".
+
+        .. versionadded:: 1.18.0
+
+            With this new parameter, all return values have identical dtypes.
+            In previous versions the dtype of the input was preserved.
+
+    xp : array_namespace, optional
+        The namespace for the return array. When ``None`` (default), then it is derived
+        from the parameters `A`, `B`, `C`, `D`. It should be compatible with the array
+        API standard or supported by array-api-compat. ``numpy`` is used by if
+        ``None`` and the parameters are not arrays.
+    device : any
+        Optional device specification for created matrices. Should match one of the
+        supported device specifications in ``xp``.
 
     Returns
     -------
     A, B, C, D : array
-        State-space matrices as two-dimensional arrays.
-
-    Notes
-    -----
-    The :ref:`tutorial_signal_state_space_representation` section of the
-    :ref:`user_guide` presents the corresponding definitions of continuous-time and
-    disrcete time state space systems.
+        State-space matrices as two-dimensional arrays with identical dtype.
 
     Raises
     ------
     ValueError
         If the dimensions n, q, or p could not be determined or if the shapes are
         incompatible with each other.
+
+    Notes
+    -----
+    If a matrix is not modified, the original matrix (not a copy) is returned.
+
+    The :ref:`tutorial_signal_state_space_representation` section of the
+    :ref:`user_guide` presents the corresponding definitions of continuous-time and
+    disrcete time state space systems.
 
     See Also
     --------
@@ -181,6 +203,21 @@ def abcd_normalize(A=None, B=None, C=None, D=None):
     ((2, 2), (2, 1), (1, 2), (1, 1))
     >>> CC
     array([[0., 0.]])
+
+    The following snippet shows the effect of the `dtype` parameter:
+
+    >>> import numpy as np
+    >>> from scipy.signal import abcd_normalize
+    >>> A, D = [[1, 2], [3, 4]], 2.5
+    ...
+    >>> AA, BB, CC, DD = abcd_normalize(A=A, D=D)  # default type casting
+    >>> print(f" AA: {AA.dtype}, BB: {BB.dtype}\n CC: {CC.dtype}, DD: {DD.dtype}")
+     AA: float64, BB: float64
+     CC: float64, DD: float64
+    >>> AA, BB, CC, DD = abcd_normalize(A=A, D=D, dtype=np.int32)  # Explicit dtype
+    >>> print(f" AA: {AA.dtype}, BB: {BB.dtype}\n CC: {CC.dtype}, DD: {DD.dtype}")
+     AA: int32, BB: int32
+     CC: int32, DD: int32
     """
     if A is None and B is None and C is None:
         raise ValueError("Dimension n is undefined for parameters A = B = C = None!")
@@ -189,18 +226,37 @@ def abcd_normalize(A=None, B=None, C=None, D=None):
     if C is None and D is None:
         raise ValueError("Dimension q is undefined for parameters C = D = None!")
 
-    xp = array_namespace(A, B, C, D)
-    A, B, C, D = (xpx.atleast_nd(xp.asarray(M_), ndim=2) if M_ is not None else
-                  xp.zeros((0, 0)) for M_ in (A, B, C, D))
+    xp = xp or array_namespace(A, B, C, D)
 
-    n = A.shape[0] or B.shape[0] or C.shape[1] or 0  # try finding non-zero dimensions
-    p = B.shape[1] or D.shape[1] or 0
-    q = C.shape[0] or D.shape[0] or 0
+    # convert inputs into 2d arrays (zero-size 2d array if None):
+    A, B, C, D = (xpx.atleast_nd(xp.asarray(M_, device=device), ndim=2, xp=xp)
+                  if M_ is not None else xp.zeros((0, 0), device=device)
+                  for M_ in (A, B, C, D))
 
-    A = xp.zeros((n, n)) if xp_size(A) == 0 else A  # Create zero-matrices if needed
-    B = xp.zeros((n, p)) if xp_size(B) == 0 else B
-    C = xp.zeros((q, n)) if xp_size(C) == 0 else C
-    D = xp.zeros((q, p)) if xp_size(D) == 0 else D
+    if dtype is None:
+        to_comp = any(xp.isdtype(M_.dtype, 'complex floating') for M_ in (A, B, C, D))
+        dtype = xp.complex128 if to_comp else xp.float64
+    else:
+        try:
+            is_numeric = xp.isdtype(dtype, 'numeric')
+        except Exception as dtype_error:
+            err_msg = f"Parameter {dtype=} must be None or a numeric dtype!"
+            raise ValueError(err_msg) from dtype_error
+        if not is_numeric:
+            raise ValueError(f"Parameter {dtype=} is not a numeric dtype!")
+
+    n = A.shape[0] or B.shape[0] or C.shape[1] # try finding non-zero dimensions
+    p = B.shape[1] or D.shape[1]
+    q = C.shape[0] or D.shape[0]
+
+    # Create zero matrices as needed:
+    A = xp.zeros((n, n), device=device) if xp_size(A) == 0 else A
+    B = xp.zeros((n, p), device=device) if xp_size(B) == 0 else B
+    C = xp.zeros((q, n), device=device) if xp_size(C) == 0 else C
+    D = xp.zeros((q, p), device=device) if xp_size(D) == 0 else D
+
+    A, B, C, D = (xp.astype(M_, dtype, copy=False, device=device)
+                  for M_ in (A, B, C, D))
 
     if A.shape != (n, n):
         raise ValueError(f"Parameter A has shape {A.shape} but should be ({n}, {n})!")
@@ -243,6 +299,17 @@ def ss2tf(A, B, C, D, input=0):
         Denominator of the resulting transfer function(s). `den` is a sequence
         representation of the denominator polynomial.
 
+    Notes
+    -----
+    Before calculating `num` and `den`, the function `abcd_normalize` is called to
+    convert the parameters `A`, `B`, `C`, `D` into two-dimesional arrays. Their dtypes
+    will be "complex128" if any of the matrices are complex-valued. Otherwise, they
+    will be of type "float64".
+
+    The :ref:`tutorial_signal_state_space_representation` section of the
+    :ref:`user_guide` presents the corresponding definitions of continuous-time and
+    disrcete time state space systems.
+
     Examples
     --------
     Convert the state-space representation:
@@ -271,7 +338,7 @@ def ss2tf(A, B, C, D, input=0):
     """
     # transfer function is C (sI - A)**(-1) B + D
 
-    # Check consistency and make them all rank-2 arrays
+    # Check consistency and make them all floating point 2d arrays:
     A, B, C, D = abcd_normalize(A, B, C, D)
 
     nout, nin = D.shape
--- a/scipy/signal/_lti_conversion.py
+++ b/scipy/signal/_lti_conversion.py
@@ -115,7 +115,7 @@ def tf2ss(num, den):
 
 
 @xp_capabilities()
-def abcd_normalize(A=None, B=None, C=None, D=None, *, dtype=None, device=None):
+def abcd_normalize(A=None, B=None, C=None, D=None, *, dtype=None):
     r"""Check state-space matrices compatibility and ensure they are 2d arrays.
 
     First, the input matrices are converted into two-dimensional arrays with
@@ -144,10 +144,6 @@ def abcd_normalize(A=None, B=None, C=None, D=None, *, dtype=None, device=None):
 
             With this new parameter, all return values have identical dtypes.
             In previous versions the dtype of the input was preserved.
-
-    device : device, optional
-        Optional device specification for returened matrices.
-        Only applicabe when using an alterantive backend.
 
     Returns
     -------
@@ -223,9 +219,8 @@ def abcd_normalize(A=None, B=None, C=None, D=None, *, dtype=None, device=None):
     xp = array_namespace(A, B, C, D)
 
     # convert inputs into 2d arrays (zero-size 2d array if None):
-    A, B, C, D = (xpx.atleast_nd(xp.asarray(M_, device=device), ndim=2, xp=xp)
-                  if M_ is not None else xp.zeros((0, 0), device=device)
-                  for M_ in (A, B, C, D))
+    A, B, C, D = (xpx.atleast_nd(xp.asarray(M_), ndim=2, xp=xp)
+                  if M_ is not None else xp.zeros((0, 0)) for M_ in (A, B, C, D))
 
     if dtype is None:
         to_comp = any(xp.isdtype(M_.dtype, 'complex floating') for M_ in (A, B, C, D))
@@ -244,13 +239,12 @@ def abcd_normalize(A=None, B=None, C=None, D=None, *, dtype=None, device=None):
     q = C.shape[0] or D.shape[0]
 
     # Create zero matrices as needed:
-    A = xp.zeros((n, n), device=device) if xp_size(A) == 0 else A
-    B = xp.zeros((n, p), device=device) if xp_size(B) == 0 else B
-    C = xp.zeros((q, n), device=device) if xp_size(C) == 0 else C
-    D = xp.zeros((q, p), device=device) if xp_size(D) == 0 else D
+    A = xp.zeros((n, n)) if xp_size(A) == 0 else A
+    B = xp.zeros((n, p)) if xp_size(B) == 0 else B
+    C = xp.zeros((q, n)) if xp_size(C) == 0 else C
+    D = xp.zeros((q, p)) if xp_size(D) == 0 else D
 
-    A, B, C, D = (xp.astype(M_, dtype, copy=False, device=device)
-                  for M_ in (A, B, C, D))
+    A, B, C, D = (xp.astype(M_, dtype, copy=False) for M_ in (A, B, C, D))
 
     if A.shape != (n, n):
         raise ValueError(f"Parameter A has shape {A.shape} but should be ({n}, {n})!")
--- a/scipy/signal/_ltisys.py
+++ b/scipy/signal/_ltisys.py
@@ -177,10 +177,10 @@ class lti(LinearTimeInvariant):
 
     >>> signal.lti(1, 2, 3, 4)
     StateSpaceContinuous(
-    array([[1]]),
-    array([[2]]),
-    array([[3]]),
-    array([[4]]),
+    array([[1.]]),
+    array([[2.]]),
+    array([[3.]]),
+    array([[4.]]),
     dt: None
     )
 
@@ -350,19 +350,19 @@ class dlti(LinearTimeInvariant):
 
     >>> signal.dlti(1, 2, 3, 4)
     StateSpaceDiscrete(
-    array([[1]]),
-    array([[2]]),
-    array([[3]]),
-    array([[4]]),
+    array([[1.]]),
+    array([[2.]]),
+    array([[3.]]),
+    array([[4.]]),
     dt: True
     )
 
     >>> signal.dlti(1, 2, 3, 4, dt=0.1)
     StateSpaceDiscrete(
-    array([[1]]),
-    array([[2]]),
-    array([[3]]),
-    array([[4]]),
+    array([[1.]]),
+    array([[2.]]),
+    array([[3.]]),
+    array([[4.]]),
     dt: 0.1
     )
 
@@ -1254,11 +1254,20 @@ class StateSpace(LinearTimeInvariant):
 
     Notes
     -----
+    If the parameter `system` is a tuple (A, B, C, D) with four state space matrices,
+    then those matrices are converted into two-dimensional arrays by calling
+    `abcd_normalize`. Their dtypes will be "complex128" if any of the matrices are
+    complex-valued. Otherwise, they will be of type "float64".
+
     Changing the value of properties that are not part of the
     `StateSpace` system representation (such as `zeros` or `poles`) is very
     inefficient and may lead to numerical inaccuracies.  It is better to
     convert to the specific system representation first. For example, call
     ``sys = sys.to_zpk()`` before accessing/changing the zeros, poles or gain.
+
+    The :ref:`tutorial_signal_state_space_representation` section of the
+    :ref:`user_guide` presents the corresponding definitions of continuous-time and
+    disrcete time state space systems.
 
     Examples
     --------
@@ -1272,12 +1281,12 @@ class StateSpace(LinearTimeInvariant):
     >>> sys = signal.StateSpace(a, b, c, d)
     >>> print(sys)
     StateSpaceContinuous(
-    array([[0, 1],
-           [0, 0]]),
-    array([[0],
-           [1]]),
-    array([[1, 0]]),
-    array([[0]]),
+    array([[0., 1.],
+           [0., 0.]]),
+    array([[0.],
+           [1.]]),
+    array([[1., 0.]]),
+    array([[0.]]),
     dt: None
     )
 
@@ -1287,8 +1296,8 @@ class StateSpace(LinearTimeInvariant):
            [0. , 1. ]]),
     array([[0.005],
            [0.1  ]]),
-    array([[1, 0]]),
-    array([[0]]),
+    array([[1., 0.]]),
+    array([[0.]]),
     dt: 0.1
     )
 
@@ -1301,8 +1310,8 @@ class StateSpace(LinearTimeInvariant):
            [0. , 1. ]]),
     array([[0.005],
            [0.1  ]]),
-    array([[1, 0]]),
-    array([[0]]),
+    array([[1., 0.]]),
+    array([[0.]]),
     dt: 0.1
     )
 
@@ -1344,6 +1353,7 @@ class StateSpace(LinearTimeInvariant):
         self._C = None
         self._D = None
 
+        # Convert A, B, C, D into 2d arrays of dtype float64 or complex128:
         self.A, self.B, self.C, self.D = abcd_normalize(*system)
 
     def __repr__(self):
--- a/scipy/signal/_ltisys.py
+++ b/scipy/signal/_ltisys.py
@@ -791,8 +791,8 @@ class TransferFunctionContinuous(TransferFunction, lti):
 
     >>> signal.TransferFunction(num, den)
     TransferFunctionContinuous(
-    array([ 1.,  3.,  3.]),
-    array([ 1.,  2.,  1.]),
+    array([1., 3., 3.]),
+    array([1., 2., 1.]),
     dt: None
     )
 
@@ -871,8 +871,8 @@ class TransferFunctionDiscrete(TransferFunction, dlti):
 
     >>> signal.TransferFunction(num, den, dt=0.5)
     TransferFunctionDiscrete(
-    array([ 1.,  3.,  3.]),
-    array([ 1.,  2.,  1.]),
+    array([1., 3., 3.]),
+    array([1., 2., 1.]),
     dt: 0.5
     )
 
@@ -1682,12 +1682,12 @@ class StateSpaceContinuous(StateSpace, lti):
     >>> sys = signal.StateSpace(a, b, c, d)
     >>> print(sys)
     StateSpaceContinuous(
-    array([[0, 1],
-           [0, 0]]),
-    array([[0],
-           [1]]),
-    array([[1, 0]]),
-    array([[0]]),
+    array([[0., 1.],
+           [0., 0.]]),
+    array([[0.],
+           [1.]]),
+    array([[1., 0.]]),
+    array([[0.]]),
     dt: None
     )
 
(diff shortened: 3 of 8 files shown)

### CODE REVIEW

**Comment by DietBru** [comment]
Rendered docs: [abcd_normalize], [ss2tf], [StateSpace].

The [CI error] seems unrelated:
```text
Run conda activate scipy-dev
Channels:
 - conda-forge
 - defaults
Platform: linux-64
Collecting package metadata (repodata.json): ...working... failed

CondaHTTPError: HTTP 000 CONNECTION FAILED for url <https://conda.anaconda.org/conda-forge/noarch/repodata.json>
Elapsed: -

An HTTP error occurred when trying to retrieve this URL.
HTTP errors are often intermittent, and a simple retry will get you on your way.
'https//conda.anaconda.org/conda-forge/noarch'


Error: Process completed with exit code 1.
```


[abcd_normalize]: https://output.circle-artifacts.com/output/job/1f41068a-449a-4e31-bf9c-12b3a172e24f/artifacts/0/html/reference/generated/scipy.signal.abcd_normalize.html
[ss2tf]: https://output.circle-artifacts.com/output/job/1f41068a-449a-4e31-bf9c-12b3a172e24f/artifacts/0/html/reference/generated/scipy.signal.ss2tf.html#scipy.signal.ss2tf
[StateSpace]: https://output.circle-artifacts.com/output/job/1f41068a-449a-4e31-bf9c-12b3a172e24f/artifacts/0/html/reference/generated/scipy.signal.StateSpace.html
[CI error]: https://github.com/scipy/scipy/actions/runs/17468741082/job/49611333397?pr=23554#step:8:1

DietBru:
Updated rendered docs: [abcd_normalize]


[abcd_normalize]: https://output.circle-artifacts.com/output/job/e9fd051a-3518-42b4-94ea-3cccd715d984/artifacts/0/html/reference/generated/scipy.signal.abcd_normalize.html

scipy/signal/_ltisys.py
```diff
@@ -1242,6 +1242,13 @@ class StateSpace(LinearTimeInvariant):
         Sampling time [s] of the discrete-time systems. Defaults to `None`
         (continuous-time). Must be specified as a keyword argument, for
         example, ``dt=0.1``.
+    cast_to: dtype | 'floating64' | None, optional
```
ilayn: Everytime we do this, after 5-6 years it becomes an issue when something needs to change or a deprecation comes. If there is no use case for this, I'd suggest we don't bubble up this functionality to `StateSpace`. It's much easier to request from the user to use proper dtype for it. Integers to floats are expected to be done automatically elsewhere. StateSpace function API is not the right place to do data casting unless we implement the entire NumPy casting machinery.

scipy/signal/_lti_conversion.py
```diff
@@ -134,6 +134,10 @@ def abcd_normalize(A=None, B=None, C=None, D=None):
         Two-dimensional array of shape (q, n).
     D: array_like, optional
         Two-dimensional array of shape (q, p).
+    cast_to: dtype | 'floating64' | None, optional
```
ilayn: Here if the default is `None` then you don't need the string so that the option here will have a single type which is also easier to handle in the future. If they don't bother to enter an option to the keyword then you don't need to bother string parsing. NumPy uses `dtype` keyword in such use cases we might want to follow that.

scipy/signal/_ltisys.py
DietBru: Removed in commit 3708d92. If needed, it can be easily added again.

scipy/signal/_lti_conversion.py
DietBru: You're right, having unified dtypes are probably easier to handle—done in commit 3708d92.

I still can think of some esoteric use cases, where keeping the input dtype is beneficial—if needed a case of `dtype='keep'` can still be added again.

I also added the keyword parameters `xp` and `device` in commit 3708d92 to work around some dtype oddities in the unit tests (if A, B, C, D are all lists, "guessing" the "xp"-namespace is not always correct).

DietBru:
Rendered docs (update): [abcd_normalize], [ss2tf], [StateSpace].




[abcd_normalize]: https://output.circle-artifacts.com/output/job/f4d6e682-4d0e-448e-b399-2c8afb2e1bae/artifacts/0/html/reference/generated/scipy.signal.abcd_normalize.html
[ss2tf]: https://output.circle-artifacts.com/output/job/f4d6e682-4d0e-448e-b399-2c8afb2e1bae/artifacts/0/html/reference/generated/scipy.signal.ss2tf.html#scipy.signal.ss2tf
[StateSpace]: https://output.circle-artifacts.com/output/job/f4d6e682-4d0e-448e-b399-2c8afb2e1bae/artifacts/0/html/reference/generated/scipy.signal.StateSpace.html

scipy/signal/_lti_conversion.py
```diff
@@ -181,27 +191,67 @@ def abcd_normalize(A=None, B=None, C=None, D=None):
     ((2, 2), (2, 1), (1, 2), (1, 1))
     >>> CC
     array([[0., 0.]])
+
+    The following snippet shows the effect of the `cast_to` parameter:
```
ev-br: `cast_to` remnant

scipy/signal/_lti_conversion.py
```diff
@@ -181,27 +191,67 @@ def abcd_normalize(A=None, B=None, C=None, D=None):
...
+     AA: float64, BB: float64
+     CC: float64, DD: float64
+    >>> AA, BB, CC, DD = abcd_normalize(A=A, D=D, dtype=np.int32)  # Explicit dtype
+    >>> print(f" AA: {AA.dtype}, BB: {BB.dtype}\n CC: {CC.dtype}, DD: {DD.dtype}")
+     AA: int32, BB: int32
+     CC: int32, DD: int32
     """
+    # Implementation note (2025-09): As of SciPy 1.17, the minimum NumPy version is
+    # 1.26.4. In Numpy 2.1, `np.asarray` learned the `device` keyword parameter and
+    # the `np.astype` function was added (hence, the special NumPy case below).
```
ev-br: This is typically a sign that something is off with numpy vs numpy-compat. 

`from scipy._array_api_compat import numpy as np_compat` --- and then `np_compat` always has the `.astype` methods etc.

But the need for this should be rare, IME is only needed if you're dealing with object arrays. Are you having to handle object arrays?

scipy/signal/_lti_conversion.py
```diff
@@ -181,27 +191,67 @@ def abcd_normalize(A=None, B=None, C=None, D=None):
...
     if B is None and D is None:
         raise ValueError("Dimension p is undefined for parameters B = D = None!")
     if C is None and D is None:
         raise ValueError("Dimension q is undefined for parameters C = D = None!")
 
-    xp = array_namespace(A, B, C, D)
-    A, B, C, D = (xpx.atleast_nd(xp.asarray(M_), ndim=2) if M_ is not None else
-                  xp.zeros((0, 0)) for M_ in (A, B, C, D))
+    if xp is None:  # try to determine namespace if not given as parameter:
+        xp = array_namespace(A, B, C, D)
```
ev-br: This pattern differs from other array-api compliant functions in scipy. We typically add `xp=..., device=...` for functions which always accept python objects and never accept any arrays (e.g. https://docs.scipy.org/doc/scipy/reference/generated/scipy.fft.fftfreq.html#scipy.fft.fftfreq) and return arrays. Then there is no other way to tell what is the namespace of the return values should be.

For functions which accept arrays, we just do `xp = array_namespace(relevant_input_arguments)`. For array-like inputs, the convention is that `array_namespace(list) -> numpy`, so that this usage is backwards-compatible.

While this is not written in stone, I would say we'd need a rather fundamental reason to deviate from an established pattern. What would be that reason?

ev-br (changes requested):
I've to admit I'm a bit confused. There seems to be two things going on here: fix of how integer dtypes are handled, and the array api conversion? It would be easier to follow if these two things are separate into array api conversion (+ strong backwards compat for array-like inputs) and a dtype fix---either as two PRs or as clearly separate commits. Would it be possible to split this PR (for me, either separate PRs or separate commits are fine, so whichever is more convenient to you).
I left a couple of comments mostly relating to the former part, the array api conversion.

For the dtype fix, I am also not clear why there is a need for the new `dtype=` kwarg actually. If the problem is that integer arrays do not get properly converted to floats, would it work to just convert them internally?   Naively---and it's full well possible I'm missing something---would https://github.com/scipy/scipy/blob/main/scipy/interpolate/_bsplines.py#L23-L41 work?

lucascolley:
are you interested in getting back to this for 1.17.0 @DietBru ?

DietBru:
I'm sorry for not finding the time to continue working sooner on this PR. I did some refactoring to achieve a more stringent Array API behavior and be more explicit about the new behavior in the docstr.

> For the dtype fix, I am also not clear why there is a need for the new dtype= kwarg actually. 

The reasoning for adding the `dtype` keyword is to be able to explicitly do:
```python
>>> abcd_normalize(A=0, D=[1, 2], dtype=np.float32)
(array([[0.]], dtype=float32),
 array([[0., 0.]], dtype=float32),
 array([[0.]], dtype=float32),
 array([[1., 2.]], dtype=float32))
```
If I have interpreted a previous failing CI tests right, pytorch can be configured to have `float32` instead of `float64` as default. So being able to be explicit here is useful, in my opinion.

Check the rendered docs [here](https://output.circle-artifacts.com/output/job/6f8b1e24-1c7b-464c-8b1f-d4ed2fa56733/artifacts/0/html/reference/generated/scipy.signal.abcd_normalize.html).

PS: Some docstr changes were also added for making `spin smoke-docs -s signal` pass.

scipy/signal/_lti_conversion.py
DietBru: Fixed in commit bd776702da5bbabdb99c6056e939dd9e89f4415d.

scipy/signal/_lti_conversion.py
DietBru: Fixed in bd776702da5bbabdb99c6056e939dd9e89f4415d.

scipy/signal/_lti_conversion.py
DietBru: Fixed in bd776702da5bbabdb99c6056e939dd9e89f4415d. I guess I got confused between the differences of `numpy.asarray` and `np_compat.asarray`.

scipy/signal/_delegators.py
```diff
@@ -58,10 +58,6 @@ def _skip_if_poly1d(arg):
 
 ###################
 
-def abcd_normalize_signature(A=None, B=None, C=None, D=None):
-    return array_namespace(A, B, C, D)
```
ev-br, line 62: This removal makes the corresponding `delegator` to be None, meaning it will skip delegating to the `cupyx.scipy` namesake. Is this intended? I'd agree there's not much point calling into CuPy just for essentially atleast_2d.

scipy/signal/_lti_conversion.py
```diff
@@ -114,44 +114,66 @@ def tf2ss(num, den):
...
+
+        .. versionadded:: 1.18.0
+
+            With this new parameter, all return values have identical dtypes.
+            In previous versions the dtype of the input was preserved.
+
+    xp : array_namespace, optional
+        The namespace for the return array. When ``None`` (default), then it is derived
+        from the parameters `A`, `B`, `C`, `D`. It should be compatible with the array
+        API standard or supported by array-api-compat. ``numpy`` is used by if
```
ev-br: What should be the expected behavior for `abcd_normalize(numpy.eye(3), None, None, None, xp=torch)`?
Same question for the `device` argument below: if `A=torch.eye(3, device="cuda")` and `device="cpu"`, who wins.

scipy/signal/_lti_conversion.py
```diff
@@ -181,6 +203,21 @@ def abcd_normalize(A=None, B=None, C=None, D=None):
...
+
+    >>> import numpy as np
+    >>> from scipy.signal import abcd_normalize
+    >>> A, D = [[1, 2], [3, 4]], 2.5
+    ...
+    >>> AA, BB, CC, DD = abcd_normalize(A=A, D=D)  # default type casting
+    >>> print(f" AA: {AA.dtype}, BB: {BB.dtype}\n CC: {CC.dtype}, DD: {DD.dtype}")
+     AA: float64, BB: float64
+     CC: float64, DD: float64
+    >>> AA, BB, CC, DD = abcd_normalize(A=A, D=D, dtype=np.int32)  # Explicit dtype
```
ev-br: Nit, minor, optional: are integer state space matrices useful anywhere? If not, consider using `float32` or complex

scipy/signal/_signaltools.py
```diff
@@ -3013,9 +3013,9 @@ def _cmplx_sort(p):
 
     Examples
     --------
-    >>> from scipy import signal
+    >>> from scipy.signal._signaltools import _cmplx_sort
     >>> vals = [1, 4, 1+1.j, 3]
-    >>> p_sorted, indx = signal.cmplx_sort(vals)
+    >>> p_sorted, indx = _cmplx_sort(vals)
```
ev-br, line 3018: Note to self: this is a docstring of a private function, so importing private functions in the code samples is fine.

scipy/signal/tests/test_cont2discrete.py
```diff
@@ -359,8 +359,8 @@ class TestC2dLti:
     def test_c2d_ss(self, xp):
         # StateSpace
         A = np.array([[-0.3, 0.1], [0.2, -0.7]])
-        B = np.array([[0], [1]])
-        C = np.array([[1, 0]])
+        B = np.array([[0.], [1.]])
+        C = np.array([[1., 0.]])
```
ev-br: Do I understand it right that the code change is backwards compatible: if `A` was integer, after `abcd_normalize` it will become float64? If so, the test does not need changing.

ev-br (commented):
A couple of minor comments below. The main question IMO is about mixing array arguments and an explicit `xp=` argument: what is the target behavior for `A=numpy_array` and `xp=torch`? I don't know the answer, and I actually don't know if there's a compelling reason to allow this (and once added, it's hard to remove something). My gut feeling is to not add it, but it's not my hill and I won't press the point if domain experts consider it desirable in this particular function.

scipy/signal/_lti_conversion.py
lucascolley: Agreed, the pattern followed elsewhere has been to infer from the input whenever there is an array parameter. I'm not aware of compelling reasons to support e.g. passing in scalars and getting back a torch array.

scipy/signal/_lti_conversion.py
DietBru: Agreed—fixed in cf1ce6a93501e337e273e90c3719e01252f43c72

scipy/signal/tests/test_cont2discrete.py
DietBru: You' right: It becomes a `float64`. Hence, the check of the results have to be modified.
Fixed in cf1ce6a93501e337e273e90c3719e01252f43c72.

scipy/signal/_delegators.py
DietBru: Do I understand it correctly: Eventually each function in signal should have a signature entry to work correctly with CuPy and JAX?

Reverted in commit cf1ce6a93501e337e273e90c3719e01252f43c72.

scipy/signal/_lti_conversion.py
DietBru: Here, the winner is decided by `xp.astype` (perhaps with some help of `xp.asarray` and `xpx.atleast_nd`). 
Isn't combining more or less compatible arrays, e.g., `torch.eye(3, device="cuda") + jax.zeros((3, 3))`, quite a generic problem for the Array API to deal with? 

I don't have a strong opinion (due to lack of experience), if a `xp` and a `device` keyword is appropriate here. 
My reasoning was "explicit is better than implicit". 

>  I'm not aware of compelling reasons to support e.g. passing in scalars and getting back a torch array.

Well, one purpose of this function is the ability to create arrays from scalars and arrays like:
```python
abcd_normalize(A=1, B=[2, 3], C=4, dtype=float32) 
```
Would this work without `xp` and `device` keywords? If yes, I would agree to remove them.

scipy/signal/_delegators.py
ev-br: Not necessarily.

The North Star goal is to have all scipy functions _work correctly and with good performance_ for CuPy, jax, torch, and  generally all array api compatible array libraries. That's a long-term goal, and I'm to harzard a guess, we'll only approach it asymptotically :-).
`*_signature` entries in `_delegators.py` is only a means to this end.  In some more details:

- A default is to make the internals of a function array-api compatible. Like you did here: since you had `xp=array_namespace(....)`, and used array-api compatible functions from the `xp` namespace, the whole function is automatically compatible. 
This works for a pure python function, like `abcd_normalize`. 
- For things which need a compiled kernel (e.g. `signal.convolve` or `lfilter`), this won't work because we'd need a CUDA kernel. That's where delegation via `delegators.py` and `_support_alternative_backends.py` comes into play:  CuPy has a dedicated kernel for `lfilter`, so we call `cupyx.scipy.lfilter` instead of `scipy.lfilter`. 

A caveat is that `cupyx.scipy.signal` module is API-compatible to _some_ scipy version, hence if you're making improvements in scipy, CuPy does not have them automatically, unless somebody sends a matching PR to CuPy :-).

For `abcd_normalize`, do we _need_ to delegate to CuPy? Probably not, the function is pure python and is hardly performance critical.

So I think removing the entry in `_delegators.py` is OK, and my comment was mainly to make a deliberate decision to keep it or to remove it.

scipy/signal/_lti_conversion.py
ev-br: The behavior of `array_namespace` is to treat "array_like" inputs as numpy objects (and ignore non-array things like None and python scalars). To illustrate:

```
$ SCIPY_ARRAY_API=1 spin ipython      # NB: env variable
...

In [1]: import numpy as np

In [2]: import torch

In [3]: from scipy._lib._array_api import array_namespace

In [4]: array_namespace(1, [2, 3], 4)    # Note: numpy namespace
Out[4]: <module 'scipy._lib.array_api_compat.numpy' from '/home/br/repos/scipy/scipy/build-install/usr/lib/python3.12/site-packages/scipy/_lib/array_api_compat/numpy/__init__.py'>

In [5]: array_namespace(1, [2, 3], torch.asarray(4))    # this errors out

....

--> 665 raise TypeError(f"Multiple namespaces for array inputs: {namespaces}")

TypeError: Multiple namespaces for array inputs: {<module 'scipy._lib.array_api_compat.numpy' from '/home/br/repos/scipy/scipy/build-install/usr/lib/python3.12/site-packages/scipy/_lib/array_api_compat/numpy/__init__.py'>, <module 'scipy._lib.array_api_compat.torch' from '/home/br/repos/scipy/scipy/build-install/usr/lib/python3.12/site-packages/scipy/_lib/array_api_compat/torch/__init__.py'>}

In [6]:  array_namespace(1, torch.asarray(3))    # This works again to give back the torch namespace 
Out[6]: <module 'scipy._lib.array_api_compat.torch' from '/home/br/repos/scipy/scipy/build-install/usr/lib/python3.12/site-packages/scipy/_lib/array_api_compat/torch/__init__.py'>

```


EDIT: IOW, what doesn't work is `array_namespace([1, 2, 3], torch.asarray(3))` because the list is an _array_like_, so it leads to a numpy namespace, and the other argument cannot be combined with a torch tensor. This is however consistent with the rest of scipy.

DietBru: Ahh, I was not aware that lists are implicitly interpreted as NumPy arrays in such contexts. 

I removed the `xp` keyword in commit 10793b6149e8e9c3f1b7f6ab16a22d03d6781873, but kept the `device` keyword (in the spirit of [fftfreq]).

[fftfreq]: https://scipy.github.io/devdocs/reference/generated/scipy.fftpack.fftfreq.html

scipy/signal/_delegators.py
DietBru: Thanks for the explanations, @ev-br. 

During testing on my local machines, I noticed that tests with `xp` fixtures are skipped, when the entry in `_delegators.py` is removed, i.e.:
```console
$ SCIPY_ARRAY_API=1 spin test  -t scipy.signal.tests.test_ltisys::Test_abcd_normalize -v
...
scipy/signal/tests/test_ltisys.py::Test_abcd_normalize::test_param_dtype[numpy] PASSED                                                                                                                                       [ 96%]
scipy/signal/tests/test_ltisys.py::Test_abcd_normalize::test_param_dtype[array_api_strict] SKIPPED (do not run with non-NumPy backends)                                                                                      [ 98%]
scipy/signal/tests/test_ltisys.py::Test_abcd_normalize::test_param_dtype[torch] SKIPPED (do not run with non-NumPy backends)   
...
```
Those tests are executed, when the entry exists. For that reason I put it back in again.

scipy/signal/_lti_conversion.py
ev-br: At a danger of sounding like a broken record, what is the intended outcome of `abcd_normalize(A=torch.eye(3, device="cpu"), device="cuda")`, is the result on CPU or on GPU? 
`fftfreq` is a bit different that it never gets any arrays, so it has nothing to grab the device from.
My impression is that we either want both `xp=..., device=...`  or neither of them.

scipy/signal/_delegators.py
ev-br: Ah, it's probably the `xp-capabilities` dance in [`_support_alternative_backends.py`](https://github.com/scipy/scipy/blob/main/scipy/signal/_support_alternative_backends.py) makes an assumption that all delegators exist, and does not apply them if it is not.
This is literally under two weeks old, so if it gets in a way, we totally can fix it. Up to you, I suppose.

scipy/signal/_lti_conversion.py
DietBru: So be it. Keyword `device` is removed in commit a7fc9b038a368a1bccd740677e81523638b4ab66.

scipy/signal/_lti_conversion.py
```diff
@@ -114,44 +114,56 @@ def tf2ss(num, den):
     return A, B, C, D
 
 
-def abcd_normalize(A=None, B=None, C=None, D=None):
-    """Check state-space matrices compatibility and ensure they are 2d arrays.
+@xp_capabilities()
```
j-bowhay: We don't need to add this since it is automatically added here:
https://github.com/scipy/scipy/blob/ae771970481349390a1fbb3dbf9b53a8ae1af93a/scipy/signal/_support_alternative_backends.py#L382-L386

scipy/signal/_delegators.py
j-bowhay: This is because it is automatically added here:
https://github.com/scipy/scipy/blob/ae771970481349390a1fbb3dbf9b53a8ae1af93a/scipy/signal/_support_alternative_backends.py#L382-L386
if you remove the delegator then you need to add `xp_capabilites` yourself

scipy/signal/_lti_conversion.py
lucascolley: > My impression is that we either want both `xp=..., device=...` or neither of them.

Yes, that is my impression too

scipy/signal/_delegators.py
DietBru: > This is literally under two weeks old, so if it gets in a way, we totally can fix it. Up to you, I suppose.

I do not have any real preference concerning that—it would be helpful though, if best practices of dealing with such caveats would be documented somewhere in the developer documentation.

scipy/signal/_lti_conversion.py
DietBru: Removed in commit b6543b0.

DietBru:
The CI errors seem unrelated:

build-docs:
```
/home/circleci/repo/build-install/usr/lib/python3.13/site-packages/scipy/interpolate/_fitpack2.py:docstring of scipy.interpolate._fitpack2.LSQSphereBivariateSpline:93: WARNING: Exception occurred in plotting scipy-interpolate-LSQSphereBivariateSpline-1
/home/circleci/repo/doc/source/tutorial/interpolate/ND_unstructured.rst:102: WARNING: Exception occurred in plotting ND_unstructured-2
```

smoke-docs:
```
_____________ [doctest] scipy.interpolate.LSQSphereBivariateSpline _____________
2398 We need to set up the interpolator object. Here, we must also specify the
2399 coordinates of the knots to use.
...

File: scipy/signal/_delegators.py (hunk 6 above)
Comment (ev-br) [reply scipy/signal/_delegators.py:62]: Agreed. In https://github.com/scipy/scipy/pull/24130 or in a follow-up?

**Review by ev-br — APPROVED** [review]
LGTM now. The circle CI failures are unrelated, and are fixed now in main, so probably safe to ignore.

Thank you @DietBru , it's great to see this part of scipy getting more dev attention!

File: scipy/signal/_delegators.py (hunk 6 above)
Comment (DietBru) [reply scipy/signal/_delegators.py:62]: Thanks for your patient reviewing, @ev-br.

> Agreed. In #24130 or in a follow-up?

Whatever works better—in a follow-up would probably be easier.

**Comment by steppi** [comment]
> > For the dtype fix, I am also not clear why there is a need for the new dtype= kwarg actually.
> 
> The reasoning for adding the `dtype` keyword is to be able to explicitly do:
> 
> ```python
> >>> abcd_normalize(A=0, D=[1, 2], dtype=np.float32)
> (array([[0.]], dtype=float32),
>  array([[0., 0.]], dtype=float32),
>  array([[0.]], dtype=float32),
>  array([[1., 2.]], dtype=float32))
> ```

I still find the use of a `dtype` kwarg odd here. My understanding is that a `dtype` kwarg is only necessary for array creation functions which have no array inputs to use to infer a dtype. My preferred solution would have involved using [`scipy._lib._array_api.xp_promote`](https://github.com/scipy/scipy/blob/d92f54e6f606d43e5ceeb9a4dd3f157288883bb6/scipy/_lib/_array_api.py#L524-L536) with `force_floating=True`.

When dealing with PyTorch's default dtype, my understanding  is that what we need to look out for is internal use of array creation functions like `xp.zeros` without setting an explicit dtype, because then results can change depending on Torch's default dtype. The goal should be that SciPy's behavior on PyTorch arrays should not depend on the default dtype at all, but instead we should be consistent that output dtypes are determined by input dtypes (and document as a limitation when this is not the case). To be fair, I only came to this understanding fairly recently, and prior to that was very confused about how SciPy should interact with default dtype. I wrote something to this effect in the docs PR I've been working on: 

```rst
``torch`` has a concept of default dtype. By default, arrays of floating point
numbers will have dtype ``torch.float32`` when no explicit dtype is specified.
The default ``dtype`` can be changed in ``torch`` with ``torch.set_default_dtype``.
A third environment variable (``SCIPY_DEFAULT_DTYPE``, again only used in the
test suite) can be used to control the default dtype used by ``torch`` in tests.
Valid values are ``"float64"`` and ``"float32"``. If ``SCIPY_DEFAULT_DTYPE`` is unset,
then ``torch``'s default dtype will be ``float64``.

The intention behind testing with different default dtypes is primarily to catch
subtle bugs that can arise with the ``torch`` backend due to internal array creation
that does not explicitly specify a dtype. The intention is not to implicitly test
that functions are numerically accurate with both ``float32`` and ``float64`` inputs,
or that input dtype controls output dtype. These tasks should instead be accomplished
mindfully by explicitly setting dtypes within tests. For the sake of consistency,
tests intended to be run with the ``torch`` backend should not use array creation
functions without explicitly setting the dtype. At the time of writing, there are many
tests in the test suite which do not follow this practice, and this could be a good source
of first issues for new contributors.
```

I'm not trying to be difficult, but would like if this public API change could be reconsidered because I think SciPy should try to move in the direction of consistency in these matters. I don't think there was a corresponding post on the Scientific Python discourse or RFC issue, otherwise I would have noticed this proposed API change and have commented before the PR was merged.

scipy/signal/_delegators.py
steppi: I added a paragraph in #24130 about this.

```rst
Note that in some modules a systematic process for delegation to native
implementations is set up, where functions are replaced with wrappers
that perform delegation. In this case, `xp_capabilities` is not always
applied as a decorator with `@` syntax, but may instead be applied
programatically on the wrappers. When working on array API standard
support within a module, its important to be aware of how such delegation
is set up, if any, and how `xp_capabilities` is being applied. A common
practice currently is to have a file, `_support_alternative_backends.py`
within a module that sets up such delegation. See for instance
`scipy.signal._support_alternative_backends.py <https://github.com/scipy/scipy/blob/main/scipy/signal/_support_alternative_backends.py>`_.
```

steppi:
I've added a subsection in #24130 which contains the recommendation I made above

```rst
For array creation functions without array inputs,
adding array API standard support can be accomplished by adding keyword
only arguments ``xp`` and ``device`` to specify the desired backend and
device respectively. See for instance `~scipy.signal.buttap` for creating
the analog prototype of an Nth-order Butterworth filter. It may also be
desirable to add a ``dtype`` kwarg to control the output ``dtype`` for
such functions. Note that none of these keyword arguments are necessary when
there are array arguments from which the backend, device, and desired output dtype
can be inferred. For the sake of API simplicity and consistency and in the
spirit of "There should be one-- and preferably only one --obvious way to do it.",
it is recommended to avoid the use of these kwargs in functions which take
at least one array argument.
```

DietBru:
Thank you a lot @steppi for taking on the tedious task of ironing out API Array corner cases!

> My understanding is that a dtype kwarg is only necessary for array creation functions which have no array inputs to use to infer a dtype.

Casting the output to `float64` by default and having the ability choose another output `dtype` is deliberate here:

* Many operations in filter/controller design are numerically quite sensitive (e.g., eigenvalue calculations in [place_poles]). Having `float64` by default was one motivation to write this PR.
* Many embedded systems, like signal processors, use `float32` for speed. Comparing simulation outputs of  `float32` / `float64`  (better: `float128`) is a pragmatic way to asses potential numeric problems for implementations on such devices. 
* Choosing the correct tolerance in unit tests is no easy feat: Balancing tight tolerance for spotting numeric problems and robustness for various back-ends, especially when supporting`float64` and `float32` default dtypes, is often a trial-and-error thing. 


In summary: Comparing behavior of different (floating) dtypes is an important use case in signal processing (I put an Array API back-end for [mpmath] on my wish-list for Santa—but I have been probably too  naughty...). I am not really sure what the new guidance on such use case should be.

> The goal should be that SciPy's behavior on PyTorch arrays should not depend on the default dtype at all, but instead we should be consistent that output dtypes are determined by input dtypes (and document as a limitation when this is not the case). 

A side note: Shouldn't guidance like this be (additionally) placed in the [Contributor guide]? If it is mandated for PyTorch, it applies to all of SciPy doesn't it? In a perfect world, a SciPy contributor would not need to learn about the back-end specifics but have a clear list of "do's and don'ts" for writing portable code.  A further example for this topic is the immutability of Dask arrays, which makes to my understanding constructs like `x[1:] += y` unusable.

>  The intention is not to implicitly test that functions are numerically accurate with both ``float32`` and ``float64`` inputs, or that input dtype controls output dtype. These tasks should instead be accomplished
mindfully by explicitly setting dtypes within tests.

It seems to me that (not only here), the annoying task of dealing with the accuracy of different dtypes is hidden behind nice words. Wouldn't it be more helpful to address these points in the [contributor guide] explicitly? Something along the line of:

* Often functions are allowed to be called with parameters of various dtypes. Their documentation should give guidance on choosing the appropriate dtypes, if applicable. An important case is the loss of accuracy when using `float32` instead of a `float64`.
* Unit tests should account for the use of various dtypes. Special consideration should be given to the common case of using `float32` instead `float64`.



[place_poles]: https://scipy.github.io/devdocs/reference/generated/scipy.signal.place_poles.html
[mpmath]: https://mpmath.org/
[Contributor guide]: https://scipy.github.io/devdocs/dev/contributor/contributor_toc.html

steppi:
> Casting the output to `float64` by default and having the ability choose another output `dtype` is deliberate here:
> 
>     * Many operations in filter/controller design are numerically quite sensitive (e.g., eigenvalue calculations in [place_poles](https://scipy.github.io/devdocs/reference/generated/scipy.signal.place_poles.html)). Having `float64` by default was one motivation to write this PR.
> 
>     * Many embedded systems, like signal processors, use `float32` for speed. Comparing simulation outputs of  `float32` / `float64`  (better: `float128`) is a pragmatic way to asses potential numeric problems for implementations on such devices.
> 
>     * Choosing the correct tolerance in unit tests is no easy feat: Balancing tight tolerance for spotting numeric problems and robustness for various back-ends, especially when supporting`float64` and `float32` default dtypes, is often a trial-and-error thing.

Yeah, I understand. We went with `float64` by default for filter prototypes and filter functions here: https://github.com/scipy/scipy/pull/23921 as a pragmatic solution, with the idea that numerical investigations into accurate native `float32` implementations could be left for future work. I added notes to the docstrings documenting this limitation and stating the situation could change in the future though.

My original draft of that PR actually added a `dtype` kwarg for the filter prototypes which take no array arguments, but @ev-br pointed out that it seemed to offer no real benefit over just doing `xp.astype(x, dtype=xp.float32)` on the output, so I dropped it.

My thinking though is that deviations from the expected [promotion rules](https://data-apis.org/array-api/latest/API_specification/type_promotion.html) violate the principle of least surprise. Exceptions are fine in some cases if they are documented, but adding a `dtype` kwarg while failing to respect input dtypes introduces an API principle that differs from the common way of doing things. Having spent a few months looking over SciPy's entire public API surface and working on formulating a plan for adding array API standard support for as many things as possible in the smoothest way, I've become more sensitive to little quirks and inconsistencies in different parts of the code-base. I think its natural that these quirks appeared through the historical process of SciPy's development, but for new code I think its best to aim in the direction of consistency. 

> In summary: Comparing behavior of different (floating) dtypes is an important use case in signal processing (I put an Array API back-end for [mpmath](https://mpmath.org/) on my wish-list for Santa—but I have been probably too naughty...). I am not really sure what the new guidance on such use case should be.

I agree, I just think that adding a `dtype` kwarg and not respecting input `dtype` doesn't offer any possibilities that can't be done by either respecting input `dtype`, or using `xp.astype`.

Also, are you aware of [mparray](https://github.com/mdhaber/mparray)? I think this could be what you're looking for, but it's still a work in progress. You should get in touch with @mdhaber if you'd like to see more progress on that front and would be interested in helping out. Its something I wish I had time to help with, but I just have too much on my plate.

> A side note: Shouldn't guidance like this be (additionally) placed in the [Contributor guide](https://scipy.github.io/devdocs/dev/contributor/contributor_toc.html)? If it is mandated for PyTorch, it applies to all of SciPy doesn't it? In a perfect world, a SciPy contributor would not need to learn about the back-end specifics but have a clear list of "do's and don'ts" for writing portable code. A further example for this topic is the immutability of Dask arrays, which makes to my understanding constructs like `x[1:] += y` unusable.

I'm in the progress of adding it here: https://github.com/scipy/scipy/pull/24130. For immutability, I mention JAX but not Dask because I think Dask support should be de-prioritized (https://github.com/scipy/scipy/issues/24205). Feel free to weigh in on #24130 with suggestions for how to increase visibility of this guidance to general contributors or any other additional suggestions you want to make. I appreciate the review you already left. I added this additional guidance after your first review.

> It seems to me that (not only here), the annoying task of dealing with the accuracy of different dtypes is hidden behind nice words. Wouldn't it be more helpful to address these points in the [contributor guide](https://scipy.github.io/devdocs/dev/contributor/contributor_toc.html) explicitly? Something along the line of:

Yeah, I agree it would be helpful to make these points more visible. I hope that #24130 is a good start. Before dismissing what I have so far as merely nice words, please understand that this was in response to the current status quo where tests ran with both `SCIPY_DEFAULT_DTYPE=float32` and `SCIPY_DEFAULT_DTYPE=float64`, but no guidance was given on the intention behind this, or what to do when `SCIPY_DEFAULT_DTYPE=float32` tests fail. In all existing tests, no one had to worry about setting a dtype in array creation functions like `zeros` or `arange`. This means that when converting existing tests to use the `xp` fixture and test on alternative backends, what ends up happening is that tests end up running in CI on PyTorch with both `float32` and `float64` inputs. Naively, at first myself and others working on this thought that the intention behind this was that tests should pass with both `float32` and `float64` inputs. In some cases, `float32` tests just got skips added to them, in others, work was spent to make the tests pass with `float32` inputs. This was kind of crazy though. We shouldn't have to get all tests passing with `float32` inputs at once to say there's meaningful PyTorch support, and adding the skips hid actual bugs. The inconsistent dtype issues on PyTorch I fixed here, https://github.com/scipy/scipy/pull/23921, did not get uncovered in the test suite precisely because `SCIPY_DEFAULT_DTYPE=float32` tests were being skipped. I unearthed them when I removed the skips to see what would happen. The reason for the inconsistencies was array creation functions that did not specify a dtype that were only hit on some code-paths. 

My understanding is that SciPy's support for different levels of floating point precision is what it is. This is something that is interesting to work on independent of array API concerns, and something where I think everyone would want accurate algorithms in both single and double precision if its feasible and someone had time to do it. When enabling PyTorch support in SciPy, I don't think we need to suddenly have great `float32` support everywhere to declare victory. We just need to support as much of SciPy's current capabilities as we can with PyTorch, and work on improving those capabilities is a separate thing. I think tests with `SCIPY_DEFAULT_DTYPE=float32` are better run with all tested input types because of the bugs they can unmask, testing with different dtypes should be explicitly done by developers.

lucascolley:
> I just think that adding a `dtype` kwarg and not respecting input `dtype` doesn't offer any possibilities that can't be done by either respecting input `dtype`, or using `xp.astype`.

+1

DietBru:
> I agree, I just think that adding a dtype kwarg and not respecting input dtype doesn't offer any possibilities that can't be done by either respecting input dtype, or using xp.astype.

Well, `abcd_normalize` [uses]  `xp.astype`  to ensure identical output dtypes—so why not expose that parameter?  Maybe renaming the parameter `dtype` to `astype` would make that clearer. The difference in our views probably boils down to the perception on how the state space functionality gets typically used. My guess would be that short semi-interactive scripts with NumPy for educational purposes are the main use case. However, it's probably not unfair to assume that the state space stuff does not see too much use. For that reason I think that having or not having  the `dtype` parameter would both be okay.

> Also, are you aware of [mparray](https://github.com/mdhaber/mparray)?

Very interesting! I was not aware of it—I'll definitely have a closer look. Thanks.

> testing with different dtypes should be explicitly done by developers.

:+1: I fully agree with that. 

Thank you for taking the time for explaining your thoughts in detail. 
#24130 is a great improvement over the status quo and in the right direction.


[uses]: https://github.com/scipy/scipy/blob/95af6337fff1f49bb10c99bca3b4f40802335305/scipy/signal/_lti_conversion.py#L246

steppi:
> Well, `abcd_normalize` [uses](https://github.com/scipy/scipy/blob/95af6337fff1f49bb10c99bca3b4f40802335305/scipy/signal/_lti_conversion.py#L246) `xp.astype` to ensure identical output dtypes—so why not expose that parameter?

Instead of `xp.astype` like this, I think pattern here should be to use `xp_promote`:

https://github.com/scipy/scipy/blob/5b2328f18fe7520efa2773f9105f8fc20db02f21/scipy/_lib/_array_api.py#L524-L536

with the option `force_floating=True`.  This takes care of type promotions happening in the expected way. When I have a chance I'll submit a PR that loses the `dtype` kwarg and uses `xp_promote` that you can take a look at. Arg conversion with `xp_promote` and `force_floating=True` is a pre-existing pattern that's used elsewhere in SciPy that should hopefully feel familiar to users.

ilayn:
> The goal should be that SciPy's behavior on PyTorch arrays should not depend on the default dtype at all, but instead we should be consistent that output dtypes are determined by input dtypes (and document as a limitation when this is not the case).

Just a note on this type of arguments that come up very often; 

As @DietBru mentioned this is not how control and signal processing world operates. This is how Array API wants to operate. So we are moving to a typed Python world and that's fine. I don't see the point of it, but fine. We already made the decision and a lot of smart people pouring their time and sweat into making this happen.

However I think it's a bit unfair to make statements that goes against entire workflow of the engineering discipline. We chose to go this way because JAX or whatever library cannot operate otherwise and we are committed to work with that library. It can be that simple and we can agree on it. 

In my opinion, we should start saying it so instead of trying to justify it with other means. It would save a lot of time and back and forth. This dtype business here is not an inconsistency for this particular work, this is unfortunately the nature of the work that is involved. Filter/controller design is notoriously sensitive to numerical noise and there is a reason why folks upcast things even as an intermediate step to 128 bit floats internally. If you are constantly switching things between zpk, ss and tf representations, no type consistency can save you from wrong results. It is unfortunate, very ugly and annoying but it is what it is especially in a production-grade output, well `scipy.signal` has other issues before its type problems but let's park that for another day.

Alternatively you can be type-consistent and wrong half of the time or you can accept the nature of the work and be correct all the time with inconsistent type behavior. I don't know who is going to work with a torch array in filter design which is inherently an exploratory work but OK once we are committed, we are committed. I'm not questioning this choice. 

Though indeed @DietBru could have chosen an alternative phrasing for the "nice words" parts (it's OK we all write quickly and can't figure out which word becomes a grenade later :smile: ) , I agree with the underlying statement that we should make it more transparent about this choice and convert it into guidelines. Because it is essentially a choice.

steppi:
Cheers @ilayn! Thanks for weighing in. I don't think we disagree as much as you think, so I'll try to lay everything out clearly. I was guilty of writing things too quickly. I think if I'd actually been trying to say what @DietBru thought I was trying to say then I'd deserve a few nice words. 

I don't actually think it should be a requirement that SciPy should always be consistent about preserving input `dtype`. I know I wrote that above, and probably sounded blindly dogmatic, but what I was really trying to say is that if the goal is to provide a way to control output `dtype` and allow for native computations in the `dtype` of choice, then respecting input `dtype`s with typical promotion rules in my opinion offers a clean API which is straightforward to implement. https://github.com/scipy/scipy/pull/24232 should make `abcd_normalize` easier to maintain  my opinion. If one doesn't care about the function allowing native computation in the `dtype` of choice, wants it to always upcast because single precision can't be trusted, but still cares about changing output `dtype`, then anything one can do with the `dtype` kwarg could also be accomplished by applying `astype` after the fact. If you thinks its better if `abcd_normalize` always upcasts to `float64`, I'd be fine with that too so long as its documented.

I was told the same thing about a `dtype` kwarg not being needed if you can just use `astype` after the fact in the review for this PR, https://github.com/scipy/scipy/pull/23921, where I was taking the opposite side and wanted to add a `dtype` kwarg to filter prototype functions. @ev-br and I ended up just making all filter prototype functions return double precision outputs regardless of input type, for pretty much the same reasons your saying. I documented that this might change in the future, but I don't care if these functions keep the current behavior.

PyTorch arrays probably never should have become a core part of this discussion. My opinion on those is that supporting PyTorch shouldn't have to change how SciPy developers think about dtypes. Just because there are tests with PyTorch having default dtype float32, doesn't mean SciPy devs need to start adding `float32` support for everything to get those tests to pass. If one blindly converts tests to support the array API, then that's what the inevitable failures you get with the torch float32 tests seem to be saying. @DietBru had mentioned adding the `dtype` kwarg because of the torch float32 failures, and I was just trying to say, "no, you shouldn't worry about those at all", and was trying to speak from my experience trying to add an unnecessary `dtype` kwarg to filter prototype functions precisely to make it easier to get those `float32` tests to pass. But things came out wrong and I emphasized the wrong things.

ilayn:
Yes I think there is no disagreement indeed. What I am trying to emphasize is that we made a decision and many changes we are doing are due to that commitment. We can hold on to that argumentation for the reasoning, and we don't need to invoke some type consistency argument or some type-theoretical argument to justify it. Because then it opens a type discussion which is not going to help because we already made the decision. 

We can save quite a bit of friction and kilobytes of text I believe if we can avoid that. Other than that, all good.

PR #23557, opened 2025-09-04
BUG: fix levene and fligner test incorrect results with trimmed mean

tristanleclercq:
#### Reference issue

Closes gh-23551

This PR changes the implementation of the Levene test in the case where the selected center is the trimmed mean, in order to trim the data within each sample only when computing the center, instead of the whole test as was the case before.

The corresponding unit test is changed accordingly, and the results are tested against values obtained with R's `leveneTest` method from the `car` package. An additional test is added for the case where the selected center is the mean, also tested against the same R function.

A similar change is made in the implementation of the Fligner test, because I believe the same error was made there, but no unit tests are added because I couldn't find any R library that implements the same test to compute values to compare to. I removed the corresponding unit test instead, because otherwise it would have been broken.

### DIFF
--- a/scipy/stats/tests/test_morestats.py
+++ b/scipy/stats/tests/test_morestats.py
@@ -1123,18 +1123,6 @@ class TestFligner:
         assert_almost_equal(Xsq1, Xsq2)
         assert_almost_equal(pval1, pval2)
 
-    def test_trimmed2(self):
-        x = [1.2, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 100.0]
-        y = [0.0, 3.0, 3.5, 4.0, 4.5, 5.0, 5.5, 200.0]
-        # Use center='trimmed'
-        Xsq1, pval1 = stats.fligner(x, y, center='trimmed',
-                                    proportiontocut=0.125)
-        # Trim the data here, and use center='mean'
-        Xsq2, pval2 = stats.fligner(x[1:-1], y[1:-1], center='mean')
-        # Result should be the same.
-        assert_almost_equal(Xsq1, Xsq2)
-        assert_almost_equal(pval1, pval2)
-
     # The following test looks reasonable at first, but fligner() uses the
     # function stats.rankdata(), and in one of the cases in this test,
     # there are ties, while in the other (because of normal rounding
--- a/scipy/stats/tests/test_morestats.py
+++ b/scipy/stats/tests/test_morestats.py
@@ -801,6 +801,13 @@ class TestLevene:
         assert_almost_equal(W, 1.7059176930008939, 7)
         assert_almost_equal(pval, 0.0990829755522, 7)
 
+    def test_mean(self):
+        # numbers from R: leveneTest in package car
+        args = [g1, g2, g3, g4, g5, g6, g7, g8, g9, g10]
+        W, pval = stats.levene(*args, center="mean")
+        assert_almost_equal(W, 2.15945985647285, 7)
+        assert_almost_equal(pval, 0.032236826559783, 7)
+
     def test_trimmed1(self):
         # Test that center='trimmed' gives the same result as center='mean'
         # when proportiontocut=0.
--- a/scipy/stats/tests/test_morestats.py
+++ b/scipy/stats/tests/test_morestats.py
@@ -811,22 +811,11 @@ class TestLevene:
         assert_almost_equal(pval1, pval2)
 
     def test_trimmed2(self):
-        x = [1.2, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 100.0]
-        y = [0.0, 3.0, 3.5, 4.0, 4.5, 5.0, 5.5, 200.0]
-        np.random.seed(1234)
-        x2 = np.random.permutation(x)
-
-        # Use center='trimmed'
-        W0, pval0 = stats.levene(x, y, center='trimmed',
-                                 proportiontocut=0.125)
-        W1, pval1 = stats.levene(x2, y, center='trimmed',
-                                 proportiontocut=0.125)
-        # Trim the data here, and use center='mean'
-        W2, pval2 = stats.levene(x[1:-1], y[1:-1], center='mean')
-        # Result should be the same.
-        assert_almost_equal(W0, W2)
-        assert_almost_equal(W1, W2)
-        assert_almost_equal(pval1, pval2)
+        # numbers from R: leveneTest in package car
+        args = [g1, g2, g3, g4, g5, g6, g7, g8, g9, g10]
+        W, pval = stats.levene(*args, center="trimmed", proportiontocut=0.25)
+        assert_almost_equal(W, 2.07712845686874, 7)
+        assert_almost_equal(pval, 0.0397269688035377, 7)
 
     def test_equal_mean_median(self):
         x = np.linspace(-1, 1, 21)
--- a/scipy/stats/tests/test_morestats.py
+++ b/scipy/stats/tests/test_morestats.py
@@ -1123,6 +1123,16 @@ class TestFligner:
         assert_almost_equal(Xsq1, Xsq2)
         assert_almost_equal(pval1, pval2)
 
+    def test_trimmed_nonregression(self):
+        # This is a non-regression test
+        # Expected results are *not* from an external gold standard,
+        # we're just making sure the results remain consistent
+        # in the future in case of changes
+        args = [g1, g2, g3, g4, g5, g6, g7, g8, g9, g10]
+        W, pval = stats.fligner(*args, center="trimmed", proportiontocut=0.25)
+        assert_almost_equal(W, 15.953569890010614, 7)
+        assert_almost_equal(pval, 0.06785752327432863, 7)
+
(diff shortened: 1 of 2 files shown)


lucascolley:
@WarrenWeckesser the ghost of git blame past has chosen you :) https://github.com/scipy/scipy/commit/726f6cc5743653bb6b0869ccf760c33d99d08eda

scipy/stats/tests/test_morestats.py
```diff
@@ -1127,18 +1123,6 @@ def _perturb(g):
...
-        x = [1.2, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 100.0]
-        y = [0.0, 3.0, 3.5, 4.0, 4.5, 5.0, 5.5, 200.0]
-        # Use center='trimmed'
-        Xsq1, pval1 = stats.fligner(x, y, center='trimmed',
-                                    proportiontocut=0.125)
-        # Trim the data here, and use center='mean'
-        Xsq2, pval2 = stats.fligner(x[1:-1], y[1:-1], center='mean')
-        # Result should be the same.
-        assert_almost_equal(Xsq1, Xsq2)
-        assert_almost_equal(pval1, pval2)
```
tylerjereddy, line 1140: If I reverse your patch to `flinger()` we have no tests that fail as a result. Typically we'd want some kind of regression guard, perhaps a regression test clearly labelled that it is just guarding against regression and not getting its expected result from a "gold standard?"

The test below the fold seems to accomplish that, failing in the absence of your patch to `flinger()`.

<details>

```diff
--- a/scipy/stats/tests/test_morestats.py
+++ b/scipy/stats/tests/test_morestats.py
@@ -1123,6 +1123,17 @@ class TestFligner:
         assert_almost_equal(Xsq1, Xsq2)
         assert_almost_equal(pval1, pval2)
 
+    def test_trimmed2(self):
+        x = [1.2, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 100.0]
+        y = [0.0, 3.0, 3.5, 4.0, 4.5, 5.0, 5.5, 200.0]
+        # Use center='trimmed'
+        Xsq1, pval1 = stats.fligner(x, y, center='trimmed',
+                                    proportiontocut=0.125)
+        expected_X = 0.3903540022781948
+        expected_p = 0.5321133818269028
+        assert_allclose(Xsq1, expected_X)
+        assert_allclose(pval1, expected_p)
```

</details>

Does that make sense?

Of course, there is no substitute for domain expert review and/or gold standards/references, but if you have a good reason for making the change we'd typically want to make sure it doesn't break again.

tristanleclercq: Sure that makes sense. I've added a non-regression test in dbbd692eabd6773a666b84349a1449304abc378b

lucascolley:
@WarrenWeckesser were you interested in taking a look at this? In case my last message was too cryptic, this is effectively reverting a change you made 15 years ago.

WarrenWeckesser:
@lucascolley, yeah, I got it :).  Obviously I missed an important point back then!  To be honest, I don't recall what I was thinking then, or how that change was reviewed, but now it is clear that the change was not correct.  Like @tristanleclercq, I haven't had any luck finding prior art for the use of different measures of central tendency in the Fligner-Killeen test.  I'll spend a bit more time today looking into it, and then review the PR.

scipy/stats/tests/test_morestats.py
```diff
@@ -801,6 +801,13 @@ def test_data(self):
         assert_almost_equal(W, 1.7059176930008939, 7)
         assert_almost_equal(pval, 0.0990829755522, 7)
 
+    def test_mean(self):
+        # numbers from R: leveneTest in package car
+        args = [g1, g2, g3, g4, g5, g6, g7, g8, g9, g10]
+        W, pval = stats.levene(*args, center="mean")
+        assert_almost_equal(W, 2.15945985647285, 7)
+        assert_almost_equal(pval, 0.032236826559783, 7)
```
WarrenWeckesser:
```suggestion
        assert_allclose(W, 2.15945985647285, rtol=5e-14)
        assert_allclose(pval, 0.032236826559783, rtol=5e-14)
```

scipy/stats/tests/test_morestats.py
```diff
@@ -811,22 +818,11 @@ def test_trimmed1(self):
...
-        W2, pval2 = stats.levene(x[1:-1], y[1:-1], center='mean')
-        # Result should be the same.
-        assert_almost_equal(W0, W2)
-        assert_almost_equal(W1, W2)
-        assert_almost_equal(pval1, pval2)
+        # numbers from R: leveneTest in package car
+        args = [g1, g2, g3, g4, g5, g6, g7, g8, g9, g10]
+        W, pval = stats.levene(*args, center="trimmed", proportiontocut=0.25)
+        assert_almost_equal(W, 2.07712845686874, 7)
+        assert_almost_equal(pval, 0.0397269688035377, 7)
```
WarrenWeckesser:
```suggestion
        assert_allclose(W, 2.07712845686874, rtol=5e-14)
        assert_allclose(pval, 0.0397269688035377, rtol=5e-14)
```

scipy/stats/tests/test_morestats.py
```diff
@@ -1127,17 +1123,15 @@ def _perturb(g):
...
-        assert_almost_equal(pval1, pval2)
+    def test_trimmed_nonregression(self):
+        # This is a non-regression test
+        # Expected results are *not* from an external gold standard,
+        # we're just making sure the results remain consistent
+        # in the future in case of changes
+        args = [g1, g2, g3, g4, g5, g6, g7, g8, g9, g10]
+        W, pval = stats.fligner(*args, center="trimmed", proportiontocut=0.25)
+        assert_almost_equal(W, 15.953569890010614, 7)
+        assert_almost_equal(pval, 0.06785752327432863, 7)
```
WarrenWeckesser:
```suggestion
        assert_allclose(W, 15.953569890010614, rtol=5e-14)
        assert_allclose(pval, 0.06785752327432863, rtol=5e-14)
```

WarrenWeckesser (changes requested):
@tristanleclercq, thanks for finding and fixing my old mistake 😳.

The fix looks good. I have just a few suggestions--really just one suggestion, to use a different test function in the unit tests.  While we use `assert_almost_equal()` a lot in the test file, that function is no longer recommended, and in new code `assert_allclose()` should be used instead.  (We mention this in the ["missing bits" developer docs](https://scipy.github.io/devdocs/dev/missing-bits.html#test-functions-from-numpy-testing), but we don't actually expect new contributors to have discovered that page.)  I've checked the reference values against R and against a reimplementation of `fligner` that I put together based on Conover et al, and found that we can safely go down to a relative tolerance of 5e-14 on all the modified tests.

scipy/stats/tests/test_morestats.py
tristanleclercq: Thanks for the review @WarrenWeckesser ! Indeed, I hadn't discovered the page you linked, thanks for sharing. I modified the tests according to your suggestion in c2d5805dfc92b152cb346aa580c7b88b3d21321a

scipy/stats/tests/test_morestats.py
tristanleclercq: c2d5805dfc92b152cb346aa580c7b88b3d21321a

scipy/stats/tests/test_morestats.py
tristanleclercq: c2d5805dfc92b152cb346aa580c7b88b3d21321a

scipy/stats/tests/test_morestats.py
WarrenWeckesser: Thanks!

WarrenWeckesser (approved):

tristanleclercq:
@lucascolley @WarrenWeckesser I'm not sure what the procedure is at this point, it seems the "owner approval" CI check is still pending. Also, should I merge this myself, and if so, should I update with merge commit or with rebase?

WarrenWeckesser:
A SciPy developer (e.g. me, @lucascolley, quite a few others) has to merge it.  I didn't merge yet, because of that Cirrus job not being run, and because I wouldn't mind another approval.

I think there should be a CI approval button visible to SciPy devs for the Cirrus job, but I don't see one.  (That may have required that I set up a CirrusCI account some time ago, which I never did.)

tristanleclercq:
> A SciPy developer (e.g. me, @lucascolley, quite a few others) has to merge it. I didn't merge yet, because of that Cirrus job not being run, and because I wouldn't mind another approval.
> 
> I think there should be a CI approval button visible to SciPy devs for the Cirrus job, but I don't see one. (That may have required that I set up a CirrusCI account some time ago, which I never did.)

Thanks for the clarification @WarrenWeckesser!

lucascolley:
the cirrus thing looks like a red herring, we don't have any cirrus jobs anymore

WarrenWeckesser:
OK, I'll merge in two days unless anyone objects or merges it before then.

WarrenWeckesser:
Merged.  Thanks again, @tristanleclercq!

PR #23563, opened 2025-09-05
BUG: fix memory leaks in `optimize.nnls` and `linalg.sqrtm`

rgommers:
Split off from gh-23433. Closes gh-23429.

See gh-23429 for the rationale.

### DIFF
--- a/scipy/linalg/_matfuncs_sqrtm.h
+++ b/scipy/linalg/_matfuncs_sqrtm.h
@@ -13,6 +13,13 @@ void matrix_squareroot_z(const PyArrayObject* ap_Am, SCIPY_Z* restrict ap_ret, i
 
 static PyObject* sqrtm_error;
 
+// A simple destructor for buffer attached to a NumPy array via a capsule.
+static void
+capsule_destructor(PyObject *capsule) {
+    void *ptr = PyCapsule_GetPointer(capsule, NULL);
+    free(ptr);
+}
+
 /*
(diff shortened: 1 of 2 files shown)


rgommers:
@ilayn I think your fix is correct indeed. I added a second one for the same code pattern in `sqrtm`.

rgommers:
I expect CI to pass; if it doesn't I'll have time on Sunday to look at it.

ilayn:
Looks good. This also needs to be in minpack and other places, which traditionally we did not care (pretty much since Travis wrapped them). But better to get them in if needs to be backported.

But I still don't quite get why umfpack job suddenly started to cough all of a sudden in this and PROPACK rewrite.

rgommers:
> But I still don't quite get why umfpack job suddenly started to cough all of a sudden in this and PROPACK rewrite.

See gh-23564.

> But better to get them in if needs to be backported.

Yeah, I'd go with this because of 1.16.2 being scheduled very soon. These are the only two issues that were actually reported. Certainly there'll be some more to do along these lines.

tylerjereddy (approved):
The one failing CI  job has errors that match those in the closed ticket about the upstream dependency issue that is resolved by pinning umfpack.

I'm not really qualified to review this one, but the code looks "ok" to me at a surface level, and Ilan's approval modulo the CI failure meets our usual standard for merging.

Given that I'd like to get `1.16.2` out the door soon, let's put this in, thanks both.

PR #23566, opened 2025-09-06
MAINT: Use np.view instead of setting dtype

kianelbo:
#### Reference issue
Addresses part of #23522 

Replaces .dtype assignments with `.view()`.


dschmitz89:
Thanks @kianelbo !

Issue #21489, opened 2024-08-31
CI: use pixi to manage environments


lucascolley:
Skrub is a nice example of using pixi to manage environments for CI jobs: https://github.com/skrub-data/skrub/blob/main/.github/workflows/testing.yml. It gives natural separation between the environment setup details and the actual content of the job. For some SciPy jobs right now, these parts are mixed in together and can be tricky to follow.

I propose that we transition to using pixi in CI to manage the majority of environments (and tasks), hopefully improving readability and clarity in the workflow files. Perhaps we keep one job each for plain pip and conda-forge, or something like that.

@rgommers has already done some experimental work with pixi for SciPy at https://github.com/rgommers/pixi-dev-scipystack, which works nicely for me locally.

I think this seems useful and shouldn't be controversial, but perhaps good to ping resident CI wizards @andyfaff @larsoner in case there are any concerns?

---

As a side-note, this would be a good way to test the waters with pixi, and see whether we would like to start recommending it as a setup method for contributors in the future.

As another side-note, did anything come of your investigation into ccache @nickodell , or did you conclude that there aren't really improvements we could / would like to make right now?

nickodell:
>As another side-note, did anything come of your investigation into ccache @nickodell , or did you conclude that there aren't really improvements we could / would like to make right now?

I found two improvements that could be made:

* The cache load action searches for a cache key that must match down to the second. The only way that can work is if two jobs begin in the same second, one finishes building and saves its cache, and then the other loads the cache.

  The upshot of this is that we are not benefitting from ccache at all, even in jobs that use it.
* The cache load action will only load caches from main branch or from previous builds on the same PR. Since there are no ccache enabled builds happening on main branch, a PR which has one commit can't benefit from ccache. If main were regularly rebuild (once per week?) and the cache saved, this cost could be reduced.

lucascolley:
thanks @nickodell ! I've moved your comment to a new issue, gh-21491.

rgommers:
> I think this seems useful and shouldn't be controversial,

"let's use a new tool" is almost always at least mildly controversial:)

> I propose that we transition to using pixi in CI to manage the majority of environments (and tasks), hopefully improving readability and clarity in the workflow files. Perhaps we keep one job each for plain pip and conda-forge, or something like that.

[...]

PR #23567, opened 2025-09-06
WIP: CI: use Pixi to manage environments and tasks

lucascolley:
gh-21489


rgommers:
@lucascolley can you comment on the caching strategy you have in mind? The current one in this PR is unclear to me, since we don't run CI jobs on merges to `main` and hence this (which I guess was your starting point based on https://pixi.sh/latest/integration/ci/github_actions/#caching) doesn't work:
```
          cache-write: ${{ github.event_name == 'push' && github.ref_name == 'main' }}
```

lucascolley:
> @lucascolley can you comment on the caching strategy you have in mind? The current one in this PR is unclear to me, since we don't run CI jobs on merges to `main` and hence this (which I guess was your starting point based on [pixi.sh/latest/integration/ci/github_actions#caching](https://pixi.sh/latest/integration/ci/github_actions/#caching)) doesn't work:
> 
> ```
>           cache-write: ${{ github.event_name == 'push' && github.ref_name == 'main' }}
> ```

I haven't thought about it properly yet. @glemaitre what do you do in skrub?

.github/workflows/macos.yml
```diff
@@ -151,35 +65,34 @@ jobs:
...
-        export DYLD_LIBRARY_PATH=$GFORTRAN_LIB
-
-        pip install meson cython pythran pybind11 ninja numpy spin
-        pip install -r requirements/openblas.txt
-        spin build --with-scipy-openblas
-
-        pip install pooch pytest hypothesis
-        spin test
-
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
rgommers, line 68: It would be good to avoid large-scale cosmetic changes.

lucascolley: my intention was to break this PR into multiple sensible diffs eventually 👍

.github/workflows/array_api.yml
```diff
@@ -34,76 +32,44 @@ jobs:
...
+          echo "timestamp=${NOW}" >> $GITHUB_OUTPUT
+
+      - name: Setup compiler cache
+        uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
+        id: cache-ccache
+        with:
+          path: ${{ steps.prep-ccache.outputs.dir }}
+          key: ${{ github.workflow }}-ccache-linux-${{ steps.prep-ccache.outputs.timestamp }}
+          restore-keys: |
+            ${{ github.workflow }}-ccache-linux-
```
rgommers: It would probably be helpful to move all the `ccache` prep into a reusable action. It would keep all the weird timestamp munging away from the casual reader, as well as make it easier to start using `ccache` in more jobs.

rgommers:
The job we don't actually have that would perhaps be nice to add here is one that exercises the minimum version of every build and test dependency.

.github/workflows/macos.yml
```diff
@@ -32,112 +32,26 @@ jobs:
...
+          manifest-path: .github/workflows/pixi.toml
+          cache-write: ${{ github.event_name == 'push' && github.ref_name == 'main' }}
+          environments: py311-system-libs-osx
+
+      - name: Test SciPy
+        shell: bash -l {0}
+        working-directory: .github/workflows
+        run: |
+          rm -rf subprojects/boost_math  # so will fail if system boost doesn't work
+          rm -rf subprojects/qhull_r  # so will fail if system qhull doesn't work
```
lucascolley:
```suggestion
          rm -rf ../../subprojects/boost_math  # so will fail if system boost doesn't work
          rm -rf ../../subprojects/qhull_r  # so will fail if system qhull doesn't work
```

.github/workflows/macos.yml
```diff
@@ -188,46 +101,22 @@ jobs:
...
+          cache-write: ${{ github.event_name == 'push' && github.ref_name == 'main' }}
+          environments: >-
+            accelerate-lp64
+            accelerate-ilp64
+
+      - name: Test SciPy
+        working-directory: .github/workflows
+        run: |
+          pixi run --environment=accelerate-lp64 test-accelerate-full
+          pixi run --environment=accelerate-ilp64 test-accelerate-full
```
lucascolley: This should go in a matrix

.github/workflows/pixi.toml
```diff
@@ -1,149 +1,365 @@
...
+[feature.accelerate.dependencies]
+libblas = { version = ">=3.9.0,<4", build = "*accelerate" }
+
+[feature.accelerate-lp64.tasks.build-accelerate]
+cmd = "spin build --build-dir=build-accelerate --with-accelerate"
+cwd = "../.."
+env = { CC = "ccache $CC", CXX = "ccache $CXX", FC = "ccache $FC" }
+
+[feature.accelerate-ilp64.tasks.build-accelerate]
+cmd = "spin build --build-dir=build-accelerate S-Dblas=accelerate -S-Duse-ilp64=true -S-Dblas-symbol-suffix='$NEWLAPACK$ILP64'"
```
```suggestion
cmd = "spin build --build-dir=build-accelerate -S-Dblas=accelerate -S-Duse-ilp64=true -S-Dblas-symbol-suffix='$NEWLAPACK$ILP64'"
```

.github/workflows/array_api.yml
```diff
@@ -34,76 +32,28 @@ jobs:
...
+      - uses: prefix-dev/setup-pixi@ba3bb36eb2066252b2363392b7739741bb777659 # v0.8.1
+        with:
+          pixi-version: v0.54.1
+          manifest-path: .github/workflows/pixi.toml
+          cache-write: ${{ github.event_name == 'push' && github.ref_name == 'main' }}
+          environments: >-
+            build
+            array-api-cpu
+
+      - uses: .github/workflows/ccache.yml
```
lucascolley, line 49:
```suggestion
      - uses: .github/workflows/ccache.yml
```

.github/workflows/array_api.yml
```diff
@@ -34,76 +32,28 @@ jobs:
...
+
+      - name: Test SciPy
+        working-directory: .github/workflows
+        run: |
+          export OMP_NUM_THREADS=2
+          pixi run test-cpu "--" --durations 3 --timeout=60
+
+      - name: Test SciPy with torch/float32
+        working-directory: .github/workflows
+        run: pixi run --environment=array-api-cpu test-torch-float32 "--" --durations 3 --timeout=60
```
lucascolley:
```suggestion
        run: pixi run --environment=array-api-cpu test-torch-float32 -- --durations 3 --timeout=60
```

.github/workflows/array_api.yml
```diff
@@ -34,76 +32,28 @@ jobs:
...
+            build
+            array-api-cpu
+
+      - uses: .github/workflows/ccache.yml
+
+      - name: Test SciPy
+        working-directory: .github/workflows
+        run: |
+          export OMP_NUM_THREADS=2
+          pixi run test-cpu "--" --durations 3 --timeout=60
```
lucascolley:
```suggestion
          pixi run test-cpu -- --durations 3 --timeout=60
```

.github/workflows/ccache.yml
```diff
@@ -0,0 +1,21 @@
...
+        shell: bash
+        run: |
+          mkdir -p "${CCACHE_DIR:-$HOME/.ccache}"
+          echo "dir=${CCACHE_DIR:-$HOME/.ccache}" >> $GITHUB_OUTPUT
+          NOW=$(date -u +"%F-%T")
+          echo "timestamp=${NOW}" >> $GITHUB_OUTPUT
+
+      - name: Setup compiler cache
+        id: cache-ccache
+        uses: actions/cache@5a3ec84eff66
```
lucascolley:
```suggestion
        uses: actions/cache@5a3ec84eff66
        with:
          path: ${{ steps.prep-ccache.outputs.dir }}
          key:  ${{ github.workflow }}-${{ matrix.python-version }}-ccache-linux-${{ steps.prep-ccache.outputs.timestamp }}
          restore-keys: ${{ github.workflow }}-${{ matrix.python-version }}-ccache-linux-
```

.github/workflows/array_api.yml
lucascolley: I factored this out into a composite action, and it seems to be working: https://github.com/scipy/scipy/actions/runs/17801280630/job/50601496587?pr=23567

lucascolley:
@rgommers ultimately I think this PR will be split up and merged elsewhere rather than being merged itself, but do you have any opinions on merge strategy? Is there enough here to start work on a lock file update workflow and then start merging what we have so far? Or would it seem better to convert more workflows and get them green first?

.github/workflows/update_lock_file.yml
```diff
@@ -0,0 +1,55 @@
...
+          committer: "Lock file bot <noreply@github.com>"
+          author: "Lock file bot <noreply@github.com>"
+          delete-branch: true
+          branch: auto-update-pixi-lock-file
+          title: "CI: update Pixi lock file :lock:"
+          body: |
+            Update the Pixi lock file.
+
+            ### Note
+            If CI fails, create a new branch based on this PR and add the required fixes to that branch.
```
lucascolley, line 44: this would need some extra setup

.github/workflows/array_api.yml
rgommers: That refactor looks good and is useful. That said, let's not expand our cache usage right now, since we're already way over the limit and GitHub just sent out an email that they will start enforcing the 10 GB limit.

rgommers:
> do you have any opinions on merge strategy? Is there enough here to start work on a lock file update workflow and then start merging what we have so far? Or would it seem better to convert more workflows and get them green first?

Good question. Maybe 1 or max 2 jobs more would be useful, but then switching over to so figuring out the lock file updates sounds right. I agree that CI jobs and lock file updates should probably two separate PRs; I think the merge order doesn't matter much, but we have to be convinced it all looks good together anyway before doing the first merge. Which I guess is the function of this PR.

.github/workflows/update_lock_file.yml
rgommers: > If CI fails, create a new branch based on this PR and add the required fixes to that branch.

If there would be a way to keep branch creation out of this repo, that would be nice. Not worth a ton of complexity probably, but it's better to not have bots running around creating contributor-visible branches.

lucascolley: > > If CI fails, create a new branch based on this PR and add the required fixes to that branch.
> 
> If there would be a way to keep branch creation out of this repo, that would be nice. Not worth a ton of complexity probably, but it's better to not have bots running around creating contributor-visible branches.

could you clarify what you mean here? Is a bot like https://github.com/skrub-bot sending PRs from its fork fine?

@glemaitre do you know why the workflow I copied from suggested to "create a new branch based on this PR", rather than just pushing to the bot's branch?

rgommers: For a fork seems better, yes.

Not sure about frequency, probably once every 3-4 weeks or so I'd think.

lucascolley: cool. I guess the question is then just how to set up this bot account and its token in a secure way?

rgommers: Yes indeed. Another option is to create a new repo under the scipy org which does do this kind of thing and that does allow creating in-repo branches for bots. 

I think you want to use Renovate for this one, right? Or a custom workflow?

There's also `gha-update` to think about. And, separately, creating a release repo similar to https://github.com/numpy/numpy-release. And possibly needing Docker images in the future for full reproducibility. 

Just random thoughts for now, I don't have a clear idea here yet. @andyfaff you may also have suggestions or wishes for this bot or other things we'd want?

lucascolley: > I think you want to use Renovate for this one, right? Or a custom workflow?

I had planned to just use this workflow, as it is small and easy to reason about, unlike renovate.

I would also be happy to try to set up renovate as it is much more powerful, but it would be trickier to get right and perhaps require a bit more maintenance.

andyfaff: It's nice to have updated hashes of github actions. gha-update is kind of ok, can be flaky if it doesn't read the markup files right. On the otherhand dependabot triggers too frequently. I'd prefer if dependabot rolled updates together in an e.g. weekly update.

What is renovate?

What is the purpose for updating a lockfile?

lucascolley: Renovate is akin to dependabot, but it is highly configurable, and supports Pixi. https://docs.renovatebot.com/

lucascolley: > What is the purpose for updating a lockfile?

To get latest versions of dependencies. Depending on our pinning strategy, this could mean just picking up bug fix releases (or new builds of the same releases), or bumping to new minor/major releases.

Locking somewhat frequently also helps prevent the situation where locked environments become un-installable locally (e.g. due to packages being yanked).

rgommers: > I'd prefer if dependabot rolled updates together in an e.g. weekly update.

It can. Because two regular contributors asked for it, I grudgingly accepted its use in PyWavelets, with a monthly update frequency: https://github.com/PyWavelets/pywt/blob/40a06711b4bee93814cab17c0a494699ba6afc13/.github/dependabot.yml#L12-L13. It's kinda okay on a repo that size with not nearly as many moving parts as SciPy has.

Renovate is better though, since it has broader support, like for Conda and Pixi, and lacks the history of being ultra-spammy with a completely non-responsive team behind it.

rgommers: > Locking somewhat frequently also helps prevent the situation where locked environments become un-installable locally (e.g. due to packages being yanked).

I don't think that is true. The whole purpose of "yanking" rather than outright deleting is that direct links and `==1.2.3` type requests still work, while for other requests like `install pkgname` or `install pkgname>1.2.3` the yanked version becomes invisible.

There is no real upside to daily or weekly updates, other than causing much more churn and reinstalls of packages for all users of the lock file.

lucascolley: off-topic but in case you didn't see what I posted on the Scientific Python discord @rgommers, Dependabot just released a very broken attempt at supporting Conda for Python packages: https://github.com/dependabot/dependabot-core/issues/2227#issuecomment-3300088866.

It seems best to steer clear of dependabot if one wants a team that actually listens to feedback, like that from Jannis in that thread.

---

> lacks the history of being ultra-spammy

if we ignore the array-api-extra incident 😆 that one was my fault, though

rgommers: The proof is in the pudding here. I think our GPU CI job has been more stable than the average job, despite the much more complex set of dependencies, exactly because of the lock file keeping things stable. And that lock file doesn't need to be updated regularly, it has had only 4 commits this year (https://github.com/scipy/scipy/commits/main/.github/workflows/pixi.lock) and only 1 of those was for a failure.

rgommers: Wow, I did not see that indeed - that is taking both technical incompetence and lack of listening back to a ridiculously high level.

andyfaff: It all we're talking about is quarterly, then there's no need for a bot.

andyfaff: i.e. someone can update it locally then make a PR.

lucascolley: > that is taking both technical incompetence and lack of listening back to a ridiculously high level.

The bigger concern is the management structure that is leaving relevant technical oversight devoid! Also https://discord.com/channels/1082332781146800168/1082338253925003385/1417616762203279430 ...

---

Back on topic, we effectively need to decide if we are willing to try out renovate, or setting up our own bot account to open PRs from a workflow like this.

For the renovate route, the first step will be someone with permissions granting access to the repo. Then I can be _extremely careful_ to get a config which only does infrequent lock file updates (and perhaps also the equivalent of `gha_update`, since it covers that too).

> Another option is to create a new repo under the scipy org which does do this kind of thing and that does allow creating in-repo branches for bots.

We would still need automation at some point for making PRs to the SciPy repo, right?

lucascolley: > i.e. someone can update it locally then make a PR.

This is also an option. If you are volunteering to be that someone then I guess that is solved!

rgommers: The main point of a semi-frequent update is to detect failures for new versions of dependencies. CI is the worst place for those failures to show up, since they then bother every contributor (hence pinning is good). That said, ideally we would get a quick signal about it, so we can file an upstream bug report or fix something. That doesn't mean that we actually need to update the lock file, that doesn't really do much.

Maybe the nicest thing would be some separate repo that did update everything, ran CI and checked it passed, and only if it didn't then would open an issue about the failure.

lucascolley: > Maybe the nicest thing would be some separate repo that did update everything, ran CI and checked it passed, and only if it didn't then would open an issue about the failure.

In which case I suggest we:
- open an issue for this idea
- start merging conversions to Pixi, and handle lock file updates manually for now
- open a new issue to track the status of which jobs use Pixi, and more broadly what each job is doing. I imagine there is plenty of room for improvements there which should be revealed by making it all declarative.

rgommers: Sounds good for the first two bullets. I wouldn't bother with 3 to be honest, it's easiest to read from the config files and we've never been successful at keeping that kind of detailed developer docs in sync. Better to just add more code comments in yml files.

lucascolley:
this PR has served its purpose!

PR #23569, opened 2025-09-06
ENH: interpolate.FloaterHormannInterpolator: add axis parameter

### LINKED ISSUE #22513: ENH: add `axis` parameter to FloaterHormannInterpolator and AAA

j-bowhay:
closes #22513 (since adding batch support to AAA is infeasible IMO as arrays would be ragged)

1. Adds an `axis` argument to `FloaterHormannInterpolator`
2. Raises an error for `roots` and `residues` for batched inputs since these currently do not work so it's better to raise a clean error than garbage.


mdhaber:
I can take a look at this after gh-23538. If you can say there whether that approach looks good to you from a high level (compared to the alternative PR, doing nothing, or something backward incompatible), it might speed things up : )

scipy/interpolate/_bary_rational.py
```diff
@@ -40,13 +40,16 @@ class _BarycentricRational:
     # generic type compatibility with scipy-stubs
     __class_getitem__ = classmethod(GenericAlias)
 
+    _axis = 0
```
ev-br: Nit: consider not using class variables? Had to check in an interactive session that different instances do indeed not share the value of `_axis`, while mutable class variables do get shared between instances. TBH I'm not sure why and whether this relies on small integers being special. 
So while it works as is, maybe best follow the principle of least surprise.

j-bowhay: Good point for some reason I overlooked the obvious, idiomatic way to do this. Updated

scipy/interpolate/_bary_rational.py
```diff
@@ -40,13 +40,17 @@ class _BarycentricRational:
...
+    def __init__(self, x, y, axis=0, **kwargs):
+        self._axis = axis
+
         # input validation
         z = np.asarray(x)
         f = np.asarray(y)
 
         self._input_validation(z, f, **kwargs)
 
+        f = np.moveaxis(f, self._axis, 0)
```
mdhaber, line 52: Why not before `_input_validation`? (Then `_input_validation` wouldn't need modification, right?)

Should `axis` be validated or rely on the error from `moveaxis`?

scipy/interpolate/_bary_rational.py
```diff
@@ -124,7 +129,8 @@ def __call__(self, z):
                 # Find the corresponding node and set entry to correct value:
                 r[jj] = support_values[zv[jj] == self._support_points].squeeze()
 
-        return np.reshape(r, z.shape + self._shape)
+        res = np.reshape(r, z.shape + self._shape)
+        return np.moveaxis(res, 0, self._axis) if z.ndim > 0 else res
 
     def poles(self):
         """Compute the poles of the rational approximation.
```
mdhaber, line 136: Is this docstring supposed to refer to AAA? Same for `roots`. I know it's outside of this PR, but happened to notice.

scipy/interpolate/tests/test_bary_rational.py
```diff
@@ -351,6 +366,9 @@ def test_trailing_dim(self, y_shape, xx_shape):
         assert rr.shape == xx.shape + y_shape
         assert_allclose(rr, yy, rtol=1e-6)
 
+        pytest.raises(NotImplementedError, r.roots)
+        pytest.raises(NotImplementedError, r.residues)
```
mdhaber: `poles`, too, for completeness?
And why add these here instead of the new batch test?

scipy/interpolate/tests/test_bary_rational.py
```diff
@@ -81,7 +81,7 @@ def test_integer_promotion(method, dtype):
 
 class TestAAA:
     def test_input_validation(self):
-        with pytest.raises(ValueError, match="same size"):
+        with pytest.raises(ValueError, match="size"):
```
mdhaber: A little more specific message check would be good, if only to make it obvious where this is coming from. While we're at it, maybe manually use the whole error message, including numbers? I've seen a few cases where this would find that an f-string wasn't actually evaluated because it wasn't actually marked to be an f-string.

scipy/interpolate/tests/test_bary_rational.py
```diff
@@ -260,6 +260,21 @@ def test_spiral_cleanup(self):
         assert_allclose(r(z), np.tan(np.pi*z/2), atol=6e-12, rtol=3e-12)
 
 
+class BatchFloaterHormann:
+    # BSpline-line class with reference batch behavior
```
mdhaber: Does this mean it was copy-pasted from the reference BSpline class?

scipy/interpolate/tests/test_bary_rational.py
```diff
@@ -260,6 +260,21 @@ def test_spiral_cleanup(self):
...
+    def __init__(self, x, y, axis):
+        y = np.moveaxis(y, axis, -1)
+        self._batch_shape = y.shape[:-1]
+        self._interps = [FloaterHormannInterpolator(x, yi,)
+                         for yi in y.reshape(-1, y.shape[-1])]
+        self._axis = axis
+
+    def __call__(self, x):
+        y = [intep(x) for intep in self._interps]
```
mdhaber:
```suggestion
        y = [interp(x) for interp in self._interps]
```

mdhaber (approved):
LGTM with a few comments. 
Is 0d output intended to be a scalar or 0d array? Either is fine with me as long as it's consistent with nearby features.

j-bowhay:
> LGTM with a few comments. Is 0d output intended to be a scalar or 0d array? Either is fine with me as long as it's consistent with nearby features.

0d seems to preferred choice for interpolate
```
In [6]: t = scipy.interpolate.BarycentricInterpolator(np.linspace(0, 1), np.linspace(0, 1))

In [7]: t(0)
Out[7]: array(0.)

In [8]: t = scipy.interpolate.FloaterHormannInterpolator(np.linspace(0, 1), np.linspace(0, 1))

In [9]: t(0)
Out[9]: array(0.)

In [10]: t = scipy.interpolate.CubicSpline(np.linspace(0, 1), np.linspace(0, 1))

In [11]: t(0)
Out[11]: array(0.)

In [13]: t = scipy.interpolate.interp1d(np.linspace(0, 1), np.linspace(0, 1))

In [14]: t(0)
Out[14]: array(0.)
```

scipy/interpolate/_bary_rational.py
j-bowhay: Because `_input_validation` checks that `f` is at least 1D which is needed for `moveaxis`

mdhaber: OK. You happy with `axis` input validation? It's fine with me to leave it as-is.

scipy/interpolate/tests/test_bary_rational.py
mdhaber: Did you want to change this one, too?

mdhaber (commented):
CI failures are gh-24052. You happy with this?

scipy/interpolate/_bary_rational.py
j-bowhay: I think lets start simple, it would be easy to add this if a user complains

scipy/interpolate/tests/test_bary_rational.py
j-bowhay: Done

j-bowhay:
> CI failures are [gh-24052](https://github.com/scipy/scipy/issues/24052). You happy with this?

All good from my end

mdhaber:
Thanks @j-bowhay!

PR #17616, opened 2022-12-16
MAINT: optimize.minimize: enhance `callback` for remaining `minimize` methods and `fmin`s

mdhaber:
#### Reference issue
Follow-up to gh-17486

#### What does this implement/fix?
Does for other `minimize` methods and `fmin` optimizers what gh-17486 did for `trust-constr` and Nelder-Mead.

#### Additional information
`fmin_` optimizers need to first wrap the callback method (it was already done by `minimize` for the minimize methods), and the `_minimize` functions need to use `_call_callback_maybe_halt` instead of just calling the callback.

PR #23573, opened 2025-09-07
MAINT: slsqp callback accepts intermediate_result

andyfaff:
Toward gh-23570.

There were two holdouts of minimize callbacks not accepting an intermediate_result. SLSQP was one of them. This PR addresses that.


mdhaber:
I'll take a look today. Failures seem unrelated. ~~If you recall whether there was a standard test we used to loop over all methods, it would be good to use that. (I'll look later.)~~ Yeah it looks like `slsqp` can be added to:
https://github.com/scipy/scipy/blob/377ee11cf92a9e3bb1dc303e140ed5daebd2e169/scipy/optimize/_minimize.py#L48

That should use it in:
https://github.com/scipy/scipy/blob/377ee11cf92a9e3bb1dc303e140ed5daebd2e169/scipy/optimize/tests/test_optimize.py#L1639

Hopefully that means we don't have to add the new tests here. But if there is anything that the standard test doesn't check that you think it should, I'd prefer we add it there rather than in an SLSQP-specific test.

Otherwise, there's not too much to review, and comparing against what we did in gh-17616, it LGTM.

@mattlevine22 are you able to test that this allows `slsqp` to support both the old `xk` and new `intermediate_result` callback interfaces? (There will be unit tests here, of course, but always good to get a user to try it.)

mattlevine22:
Yes, can confirm these changes allowed SLSQP to use callback successfully both via the old `xk` and new `intermediate_result` interfaces.

andyfaff:
I added slsqp to `MINIMIZE_METHODS_NEW_CB`. However, `scipy.optimize.tests.test_optimize::TestOptimizeSimple::test_callback_stopiteration[2-slsqp]` fails (and 0+1). I think it's because of the way SLSQP deals with iteration counting and when the callback is called.

~~I modified the slsqp_py code to add print statements `print(f"{x=} in callback")` and `print(f"{x=} on exit")`
From running that test I see that the callback isn't actually being called when maxiter is set.~~

~~i.e. [`state_dict['iter'] > iter_prev`](https://github.com/scipy/scipy/blob/main/scipy/optimize/_slsqp_py.py#L527) is never True in the test, so the callback never stops the iteration, and  the asserts (assert res.fun == ref.fun, assert_equal(res.x, ref.x)) fail.~~

I know from the test that I added though that `StopIteration` does work.

andyfaff:
More investigation. `scipy/optimize/tests/test_optimize.py::TestOptimizeSimple::test_callback_stopiteration[2-slsqp]` fails against SLSQP because the central loop halts at [L538](https://github.com/scipy/scipy/blob/main/scipy/optimize/_slsqp_py.py#L538) because the maximum number of iterations has been reached. At this point `x` has been updated since the last major iteration number had increased, meaning that callback.x is out of sync with the final result.x.
I think the test is bogus for SLSQP because the termination criterion assumes that the halt point is always at the end of a major iteration.

mdhaber:
> I think the test is bogus for SLSQP because the termination criterion assumes that the halt point is always at the end of a major iteration.

I activated the common test for SLSQP, giving it an exception for the part it's not passing. I'm not sure if the test is "bogus" for assuming that SLSQP would behave like all the other methods or if SLSQP's behavior is an unfortunate exception; it is what it is. 

In any case, adding SLSQP to the test checks several valuable things that weren't tested here before (iteration count, status code), and the excercise led to strengthening the common test with a check of the error message and `success` attribute, which caught a bug in COBYQA (gh-23660).

If tests pass and if the updates look OK to someone, I'll go ahead and merge.

scipy/optimize/_trustregion_constr/minimize_trustregion_constr.py
```diff
@@ -306,7 +306,7 @@ def _minimize_trustregion_constr(fun, x0, args, grad,
         * 0 : The maximum number of function evaluations is exceeded.
         * 1 : `gtol` termination condition is satisfied.
         * 2 : `xtol` termination condition is satisfied.
-        * 3 : `callback` function requested termination.
+        * 3 : `callback` raised `StopIteration`.
```
andyfaff, line 309: Won't this be an issue for downstream projects if we change the message text, code might be looking for specific text.

andyfaff (commented):
LGTM, apart from the question mark over the specific text.

scipy/optimize/_trustregion_constr/minimize_trustregion_constr.py
mdhaber: I recall other discussions of changing error messages where the opinion has been that the text is not considered to be stable API and that users relying on them programmatically is bad practice. It seems that Python doesn't hesitate to change error messages, since there is a section "Improved error messages" several release notes (e.g. [3.10](https://docs.python.org/3/whatsnew/3.10.html#better-error-messages),  [3.12](https://docs.python.org/3/whatsnew/3.12.html#improved-error-messages), [3.13](https://docs.python.org/3/whatsnew/3.13.html#improved-error-messages)). NumPy has also documented changes in release note messages (e.g. [2.3](https://numpy.org/devdocs/release/2.3.0-notes.html#changes), [2.4](https://numpy.org/devdocs/release/2.4.0-notes.html#improved-error-message-for-assert-array-compare)). Instead, we provide `status` codes, and I've restrained myself from making those consistent.

andyfaff: I'm fine with messages that aren't relied upon not to be strictly unchanging. Having said that, I think there's various items scattered around the scipy code base that do this kind of check.
The only thing that should definitely be consistent is a status code equalling zero for success

mdhaber: Feel free to change that part back if you consider it blocking. It's just to make things be consistent, but not worth holding up this PR.

> The only thing that should definitely be consistent is a status code equalling zero for success

Yeah, and enhancing the common test found that wasn't true for cobyqa (gh-23660).

andyfaff: I think we should merge as is

mdhaber: I would, but we have a prerelease test failure. Reverting the last commit to see if somehow that is the cause.

andyfaff:
It could've been a flake. Rather than reverting the commit, could just run the test again?

mdhaber:
Maybe, but I ran that twice (restarted the failed job), and it failed both times.
Now, after the revert, it's passing.
I'll revert the revert and see if it is still failing.

mdhaber:
Wow. It actually looks real.

andyfaff:
The failing test is 
```
FAILED scipy/optimize/tests/test_optimize.py::TestWorkers::test_smoke[CG-2] - ExceptionGroup: multiple unraisable exception warnings (6 sub-exceptions)
```
with messages along the lines of:
```
    | Traceback (most recent call last):
    |   File "/opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/multiprocessing/resource_tracker.py", line 207, in _send
    |     self.ensure_running()
    |   File "/opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/multiprocessing/resource_tracker.py", line 126, in ensure_running
    |     return self._reentrant_call_error()
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |   File "/opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/multiprocessing/resource_tracker.py", line 70, in _reentrant_call_error
    |     raise ReentrantCallError(
    | multiprocessing.resource_tracker.ReentrantCallError: Reentrant call into the multiprocessing resource tracker
    | 
    | During handling of the above exception, another exception occurred:
    | 
    | Traceback (most recent call last):
    |   File "/opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/multiprocessing/util.py", line 227, in __call__
    |     res = self._callback(*self._args, **self._kwargs)
    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |   File "/opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/multiprocessing/synchronize.py", line 88, in _cleanup
    |     unregister(name, "semaphore")
    |   File "/opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/multiprocessing/resource_tracker.py", line 203, in unregister
    |     self._send('UNREGISTER', name, rtype)
    |   File "/opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/multiprocessing/resource_tracker.py", line 213, in _send
    |     warnings.warn(
    | UserWarning: ResourceTracker called reentrantly for resource cleanup, which is unsupported. The semaphore object '/mp-w68sugdi' might leak.
    | 
    | The above exception was the direct cause of the following exception:
    | 
    | Traceback (most recent call last):
    |   File "/opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/site-packages/_pytest/unraisableexception.py", line 67, in collect_unraisable
    |     warnings.warn(pytest.PytestUnraisableExceptionWarning(msg))
    | pytest.PytestUnraisableExceptionWarning: Exception ignored in: <Finalize object, dead>
    | Enable tracemalloc to get traceback where the object was allocated.
    | See https://docs.pytest.org/en/stable/how-to/capture-warnings.html#resource-warnings for more info.
```

The test is using `multiprocessing`, so the presence of those ResourceTracker warnings seems related. I can't understand why the test is not failing for this PR and not others. There's nothing in this PR that is related to the "failing" test.

andyfaff:
and it's only failing for 3.12 and not 3.11

mdhaber:
Agreed. Well, let's see what happens if I comment out `CG` in the test. If this passes, I'll merge and open up a PR trying to re-enable it. If the test fails in that PR, I'll open an issue.

andyfaff:
https://github.com/nedbat/coveragepy/issues/1306 might be related.

andyfaff:
So I wonder if the test would pass if we didn't install pytest-cov

mdhaber:
OK, not so surprisingly, it didn't really have to do with `method='CG'`. Comment out one test, another begins to fail.

mdhaber:
K. I just want to see if it has something to do with the changes to `trust-constr` or maybe something about the testing of `slsqp`. If worst comes to worst I'll comment out the whole test tonight and we can open that as an issue. There is nothing we've done here that *should* be causing this.

mdhaber:
Interesting. The test that was failing is passing after commenting out `trust-constr`. If the rest pass, I'll merge this and open a PR trying to bring back `trust-constr`. If that passes, great; if not, I'll close the PR and open an issue.

mdhaber:
Now it's failing in a different job : / Let's see if it starts appearing in other PRs.

andyfaff:
I wonder if it's due to any bumps in upstream packages.  pytest/pytest-cov were updated a couple of weeks ago.

I have some ideas about things to investigate further, but at the moment I can't reproduce it locally.

scipy/optimize/tests/test_optimize.py
```diff
@@ -3430,9 +3433,9 @@ def sparse_rosen_hess(x):
     ['l-bfgs-b',
      'bfgs',
      'slsqp',
-     'trust-constr',
+     # 'trust-constr',
```
mdhaber:
```suggestion
     'trust-constr',
```

scipy/optimize/tests/test_optimize.py
```diff
@@ -3430,9 +3433,9 @@ def sparse_rosen_hess(x):
...
      'Newton-CG',
-     'CG',
+     'CG',  # failing in gh-23573; will try to re-enable immediately
```
mdhaber:
```suggestion
     'CG',
```

mdhaber:
Just saw this failure in the same test in gh-23655.
```
FAILED ../optimize/tests/test_optimize.py::TestWorkers::test_equal_bounds[slsqp-2] - ExceptionGroup: multiple unraisable exception warnings (24 sub-exceptions)
```
Merging!

PR #23574, opened 2025-09-07
ENH: linalg.fiedler: add native batch support and array api support

j-bowhay:
Firstly, changes batch support to be native rather than using `apply_along_axis`. Secondly, adds array api support.


lucascolley (approved):
looks great, thanks Jake!

PR #23575, opened 2025-09-07
MAINT: use boost master branch

dschmitz89:
Since we now use boost releases only, it makes more sense to simply rely on their master branch instead of pointing to specific commits in their develop branch. This is applied in this PR.


lucascolley (approved):
indeed this aligns with the `tag` on the boost repo:

```console
math on 🎋 HEAD (5e088ff) via △ v4.1.1
❯ git log
commit 5e088ffe2ed0e237b9069e3a7352865283d8f196 (HEAD, tag: boost-1.89.0)
```

thanks Daniel!

dschmitz89:
Wow, this was quick, thanks Lucas.

PR #23579, opened 2025-09-07
ENH: interpolate.AAA: add diagonal scaling to improve conditioning

j-bowhay:
For tricky example, e.g. the added test, the Loewner matrix becomes ill-conditioned. This pr adds a check for this which then triggers diagonal scaling.


ev-br (approved):
A tricky case indeed :-). 
While I'm not 100% familiar with the approach, am ready to defer to Jake, and to the fact that the added test indeed fails on main and passes on the PR. 

Merging, thanks @j-bowhay !

PR #23580, opened 2025-09-08
DOC: spatial: Better document align_vectors weighting

scottshambaugh:
Using `Rotation.align_vectors` with non-unit length vectors will return different results than if the vectors are unit length. I've messed this up even as someone very familiar with this code, and the docs updates here help highlight and explain this behavior.

See https://github.com/scipy/scipy/issues/23583 for some more discussion

#### Additional information

[docs only]


nmayorov:
I think this is a good addition, not much questions.

PR #23584, opened 2025-09-08
ENH: migrate `special.chdtriv` to boost

dschmitz89:
#### Reference issue
Towards #21966

Migrates `chdtriv` to boost. Boost contains a function [gamma_p_inva](https://www.boost.org/doc/libs/latest/libs/math/doc/html/math_toolkit/sf_gamma/igamma_inv.html) that inverts the incomplete gamma with respect to a. This function could be a useful standalone private function in special on its own. Here it comes in quite handy to implement `chdtriv`.

#### Benchmark results
chdtriv_benchmark

<p>

```python
from mpmath import mp
import numpy as np
from scipy.special import chdtr, chdtriv, exp10


BRANCH = "main"

@mp.workdps(50)
@np.vectorize
def chdtr_ref(v, x):
    v = mp.mpf(v)
    x = mp.mpf(x)
    res = mp.gammainc(v/2, a=0, b=x/2, regularized=True)
    return float(res)

    """
    """
    return rel_error

rng = np.random.default_rng(12348430813753)

points_per_variable = 100

v = rng.uniform(-10, 10, points_per_variable)
x = rng.uniform(-5, 5, points_per_variable)

v = exp10(v)
x = exp10(x)

v, x = np.meshgrid(v, x)
v = v.ravel()
x = x.ravel()

cdf_ref = chdtr_ref(v, x)
#chdtriv_vals = chdtriv(cdf_ref, x)

#np.save("chdtriv_bench_data_" + BRANCH + ".npy", chdtriv_vals)


######### Plotting #########

pr_chdtriv_vals = np.load("chdtriv_bench_data_PR.npy")
main_chdtriv_vals = np.load("chdtriv_bench_data_main.npy")
mask = (cdf_ref > 1e-30) & (cdf_ref < 0.99999999)

rel_error_pr = calculate_rel_error(pr_chdtriv_vals[mask], v[mask])
rel_error_main = calculate_rel_error(main_chdtriv_vals[mask], v[mask])

plt.hist(np.log10(rel_error_pr), bins=int(np.sqrt(len(rel_error_pr))), alpha=0.5, label='PR', density=True, cumulative=True)
plt.hist(np.log10(rel_error_main), bins=int(np.sqrt(len(rel_error_main))), alpha=0.5, label='main', density=True, cumulative=True)
plt.legend()
plt.ylabel('Cumulative Density')
plt.title('Cumulative Distribution of Relative Errors for chdtriv')
plt.show()

``` 

</p>
</details>


dschmitz89:
In principle, we could also migrate `chdtr`, `chdtrc` and `chdtri` to boost to have consistency across the distributions. I will let the reviewer decide as adding those would make this PR much bigger.

WarrenWeckesser:
Thanks @dschmitz89.

I'm taking a look at the precision gained here; I should be able to do a review in a day or two. In the meantime, I think you know the changes I'm going to request in `boost_special_functions.h`...

WarrenWeckesser:
I experimented a bit with some code to check the improved precision.  The following plots show the relative error of `chdtriv(p, x)` for a few values of `x` and for a grid of 3000 `p` values on the interval (0, 1).  The first plot was generated with the main development branch (at the time of this comment), and the second is with this pull request.  Short (and not surprising) summary: the Boost code gives us much higher precision than the old code.  Reference values for computing the relative error were computed using `mpmath`.

Main development branch:

chdtriv_relerr_1 17 0 dev0+git20250917 22cf964

This pull request (i.e. `chdtriv` uses `gamma_p_inva` from boost/math):

chdtriv_relerr_1 17 0 dev0+git20250912 83f7b86

WarrenWeckesser:
An unexpected behavior of the new version is that `chdtriv(1.0, x)` returns `4.450147717014403e-308` for any `x > 0`. (That's better than the current main branch, where `nan` is returned.)  E.g.

```
In [18]: x = np.array([0.1, 1.0, 10.0, 100.0])

In [19]: chdtriv(1.0, x)
Out[19]: array([4.45014772e-308, 4.45014772e-308, 4.45014772e-308, 4.45014772e-308])
```

This is a quirk of the underlying Boost implementation, which requires `a > 0` in `gamma_p(a, x)`.  So (apparently for consistency with `gamma_p(a, x)`) they don't return 0 for `gamma_p_inva(x, 1.0)` [1], as that value of `a` would not be a valid input to `gamma_p(a, x)`.

In SciPy, we allow the parameter to be zero (e.g. `chdtr(0, x)` returns 1.0), so we can return 0 from `chdtriv(1.0, x)`.  This should be added as another special case to be handled in `chdtriv_wrap`.

-----

[1] Note the order of the parameters in the Boost functions: `gamma_p_inva(x, p)` is the inverse with respect to `a` of `gamma_p(a, x)`.

WarrenWeckesser (dismissed):
Overall looks good, just a few tweaks needed:
* Return 0 when p=1 and x > 0.
* Fix the examples in the docstring--the improved precision changes the output.
* Fix the code indentation in `chdtriv_wrap`. (We're not autoformatting yet, so for now we'll continue to do it manually.)

dschmitz89:
I incorporated the changes. I wonder why the refguide check did not pick up the changed examples.

scipy/special/boost_special_functions.h
```diff
@@ -2243,4 +2243,54 @@ f_sf_double(double dfn, double dfd, double x)
...
+    } catch (const std::underflow_error& e) {
+        sf_error("chdtriv", SF_ERROR_UNDERFLOW, NULL);
+        y = 0;
+    } catch (...) {
+        sf_error("chdtriv", SF_ERROR_NO_RESULT, NULL);
+        y = NAN;
+    }
+    if (y < 0) {
+        sf_error("chdtriv", SF_ERROR_NO_RESULT, NULL);
+        y = NAN;
```
WarrenWeckesser, line 2281: Sorry, one more question: can this case occur? Can gamma_p_inva() return a negative value?

dschmitz89: I do that check now for all boost wrappers, just to be on the safe side as we have seen that boost can return values outside $[0, 1]$ for different CDFs. Compared to the actual computation, this check is very cheap so I think it does not hurt.

WarrenWeckesser: OK.  I think I've done the same thing in the past.

WarrenWeckesser:
I'm going to push a couple cosmetic changes, let the tests run once more (overabundance of caution and all that) and merge.

scipy/special/boost_special_functions.h
```diff
@@ -2243,4 +2243,54 @@ f_sf_double(double dfn, double dfd, double x)
...
+chdtriv_wrap(const Real p, const Real x)
+{
+    if (std::isnan(p) || std::isnan(x)) {
+        return NAN;
+    }
+    if ((x <= 0) || (p < 0) || (p > 1)) {
+        sf_error("chdtriv", SF_ERROR_DOMAIN, NULL);
+        return NAN;
+    }
+    if ((p == 1)) {
```
WarrenWeckesser:
```suggestion
    if (p == 1) {
```

scipy/special/boost_special_functions.h
```diff
@@ -2243,4 +2243,54 @@ f_sf_double(double dfn, double dfd, double x)
...
+        sf_error("chdtriv", SF_ERROR_OVERFLOW, NULL);
+        y = INFINITY;
+    } catch (const std::underflow_error& e) {
+        sf_error("chdtriv", SF_ERROR_UNDERFLOW, NULL);
+        y = 0;
+    } catch (...) {
+        sf_error("chdtriv", SF_ERROR_NO_RESULT, NULL);
+        y = NAN;
+    }
+    if (y < 0) {
```
WarrenWeckesser, line 2279:
```suggestion
    // In theory, y should never be negative, but with an overabundance of caution
    // we check and return NAN if it happens.
    if (y < 0) {
```

WarrenWeckesser:
Merged.  Thanks @dschmitz89!

dschmitz89:
@WarrenWeckesser You are not listed as co author on the final merge in the main branch, was that intended?

WarrenWeckesser:
@dschmitz89, yeah, if I make some tiny cosmetic changes--things that I could just have easily requested and waited for the author to update--I tend to drop my name as co-author. I don't know if this is necessarily a good practice. I'm happy to keep any contributors to my PRs as coauthor, even for tiny changes, so it is definitely not something I expect on my PRs.

The added comment was probably borderline for "tiny cosmetic change"; maybe you're thinking "I would never use the phrase 'overabundance of caution'" and are annoyed to have it attributed to you.  If so, sorry!

dschmitz89:
> @dschmitz89, yeah, if I make some tiny cosmetic changes--things that I could just have easily requested and waited for the author to update--I tend to drop my name as co-author. I don't know if this is necessarily a good practice. I'm happy to keep any contributors to my PRs as coauthor, even for tiny changes, so it is definitely not something I expect on my PRs.
> 
> The added comment was probably borderline for "tiny cosmetic change"; maybe you're thinking "I would never use the phrase 'overabundance of caution'" and are annoyed to have it attributed to you.  If so, sorry!

No reason to apologize, I just thought there was some glitch during the merge process or so. Your changes are definitely useful!

PR #23590, opened 2025-09-09
ENH: linalg.block_diag: add array api support

j-bowhay:
Another pretty straightforward conversion.


lucascolley (changes requested):
this works for me locally!

```diff
diff --git a/scipy/linalg/_special_matrices.py b/scipy/linalg/_special_matrices.py
index bd492e5749..2915e836aa 100644
--- a/scipy/linalg/_special_matrices.py
+++ b/scipy/linalg/_special_matrices.py
@@ -334 +334 @@ def leslie(f, s):
-@xp_capabilities(skip_backends=[('jax.numpy', "JAX doesn't allow item assignment.")])
+@xp_capabilities(jax_jit=False)
@@ -416 +416 @@ def block_diag(*arrs):
-        out[..., r:r + rr, c:c + cc] = arrs[i]
+        out = xpx.at(out)[..., r:r+rr, c:c+cc].set(arrs[i])
```

otherwise LGTM :)

lucascolley:
looks like just one dask failure left

mdhaber:
Here's an MRE for that:
```python3
import array_api_strict as strict
from array_api_compat import numpy as np, torch
from jax import numpy as jnp
from array_api_compat.dask import array as dask
import array_api_extra as xpx

for xp in [np, strict, torch, jnp, dask]:
    A = xp.zeros((1, 2))
    x = xp.zeros((0, 2))
    # A[1:1, 0:2] = x
    A = xpx.at(A)[1:1, 0:2].set(x)
    print(f"{xp} OK")

# array_api_compat.numpy OK
# array_api_strict OK
# array_api_compat.torch OK
# jax.numpy OK
# ValueError: shape mismatch: value array of shape (0, 2) could not be broadcast to indexing result of shape (0, 2)
```

But it looks like it's not an `array_api_extra` issue - by uncommenting the previous line (and removing JAX from the list), we get the same error.

I'll file an issue.

I think Dask can skip the test.

mdhaber:
Turns out this was reported recently: dask/dask#11800.

scipy/linalg/_special_matrices.py
```diff
@@ -331,6 +331,7 @@ def leslie(f, s):
     return a
 
 
+@xp_capabilities(jax_jit=False, skip_backends=[("dask.array", "dask/dask#1180")])
```
mdhaber: >I think Dask can skip the test.

One test of a degenerate case fails due to a bug in Dask. I'd suggest that we skip the test rather than skipping all tests. I know this will mark Dask as supported, and I think that's OK. Dask is supported; Dask just happens to have a bug.

lucascolley (approved):
LGTM if CI is green, thanks both!

lucascolley:
CI failures look unrelated

PR #23600, opened 2025-09-11
MAINT: fix more memory management and error handling issues in C code

rgommers:
This is a follow-up to gh-23563. Main changes:
- `PyArray_SimpleNewFromData` on plain C arrays requires taking care of memory management of the data being passed in. There are two strategies:
    - In general, creating a capsule to hold the data and making setting the `.base` attribute of the created ndarray to that capsule will do the right thing. This pattern is documented at https://numpy.org/devdocs/reference/c-api/data_memory.html#what-happens-when-deallocating-if-there-is-no-policy-set
    - If one is sure that the data won't go out of scope, it can be done much more concisely by telling the ndarray that it doesn't own the data. This is fairly rare probably, but callback functions do fit this pattern.
- Proper error handling for `PyCapsule_New` and `PyArray_SetBaseObject`.
- A somewhat larger refactor in `__minpack.h`, because that code was particularly hard to read.
- Clean up casting back and forth between `PyArrayObject` and `PyObject` to be more minimal.
- Fix some very unergonomic tuple constructing that was leaking a bit of memory as well. `PyTuple_Pack` is usually the better constructor to use.

### DIFF
--- a/scipy/integrate/_odepackmodule.c
+++ b/scipy/integrate/_odepackmodule.c
@@ -80,63 +80,56 @@ PyObject *call_odeint_user_function(PyObject *func, npy_intp n, double *x,
     is returned.
     */
 
-    PyArrayObject *sequence = NULL;
+    PyObject *sequence = NULL;
     PyObject *tfloat = NULL;
     PyObject *firstargs = NULL;
     PyObject *arglist = NULL;
     PyObject *result = NULL;
-    PyArrayObject *result_array = NULL;
+    PyObject *result_array = NULL;
 
     /* Build sequence argument from inputs */
-    sequence = (PyArrayObject *) PyArray_SimpleNewFromData(1, &n, NPY_DOUBLE,
-                                                           (char *) x);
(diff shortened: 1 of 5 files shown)


rgommers:
The one CI failure is unrelated.

ilayn (approved):
Looks good to me. MINPACK would need a bit more surgery about the thunks but that's an orthogonal change. It is very nice that we are addressing this, we might want to enable ASAN/UBSAN in one of the jobs to catch them while PRs are fresh

rgommers:
> we might want to enable ASAN/UBSAN in one of the jobs to catch them while PRs are fresh

That'd be nice, but there are more issues to work through still. CI job in [this branch](https://github.com/rgommers/scipy/tree/ci-asan), [last failure](https://github.com/rgommers/scipy/actions/runs/17670072800) is still in `minpack`. Iterating on that locally is needed, since it aborts on first failure and CI is pretty slow.

rgommers:
There's nothing really scipy-submodule-specific about this code and I'm not expecting more reviews, so I'll go ahead and hit the Merge button based on @ilayn's approval. Thanks @ilayn

PR #23606, opened 2025-09-12
MAINT: signal: simplify `cheb2ap` formulas

matthiasstraka:
This is a purely mathematical formula optimization
(1) -conj(0+ja) == 0+ja
(2) complex sinh is equivalent to previous expression

In addition, formulas of `cheb2ap` and `cheb1ap` are now more similar in structure.

### DIFF
--- a/scipy/signal/_filter_design.py
+++ b/scipy/signal/_filter_design.py
@@ -4760,14 +4760,12 @@ def cheb2ap(N, rs, *, xp=None, device=None):
     else:
         m = xp.arange(-N+1, N, 2, dtype=xp_default_dtype(xp), device=device)
 
-    z = -xp.conj(1j / xp.sin(m * xp.pi / (2.0 * N)))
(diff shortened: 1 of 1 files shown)


Peiffap (approved):
LGTM!

DietBru (approved):
As far as I can tell, the modification always produces the same result (up to numerical accuracy). Thanks @matthiasstraka 

PS: It would be great if [cheb2ap]'s docstr would reference  source (e.g., [Wikipedia] seems comprehensive enough) against which the formulas for the implementation could verified. Also, I did not find any explicit unit tests for [chebap] which verify the correctness of the calculations. But it is better to address those things in a new pull request.


[cheb2ap]: https://scipy.github.io/devdocs/reference/generated/scipy.signal.cheb2ap.html
[Wikipedia]: https://en.wikipedia.org/wiki/Chebyshev_filter

PR #23614, opened 2025-09-14
TST: shim for NumPy warning changes

tylerjereddy:
* Fixes gh-23603 by suppressing the match on the final punctuation. Locally, this restores full testsuite passing against latest NumPy `main`.

[skip circle]

Normally I'd leave this as suitable for a new contributor to get involved, but I'm working to revert a hack on the release branch and having the tests all green against NumPy `main` there is nice too, so just doing it..

PR #23543, opened 2025-09-03
MAINT: backports for SciPy 1.16.2

tylerjereddy:
We're about a week away from targeted `1.16.2` release date--getting this rolling to see what kinds of problems arise early on.

Backports included so far:

1. gh-23122
2. gh-23133
3. gh-23403
4. gh-23414
5. gh-23435
6. gh-23442
7. gh-23452
8. gh-23457
9. gh-23471
10. gh-23475
11. gh-23488
12. gh-23495
13. gh-23540
14. gh-23154 (OpenBLAS bump only)
15. gh-23576
16. gh-23563

TODO:

- [x] update release notes, including mention of WoA binaries
- [x] regular CI passing
- [x] wheels CI passing, including new WoA binary builds
- [x] read diff carefully
- [x] any other backports we want/need?
- [x] perhaps suppress the testsuite errors from NumPy warnings deprecations with pre-release NumPy, `DeprecationWarning: NumPy warning suppression and assertion utilities are deprecated. Use warnings.catch_warnings, warnings.filterwarnings, pytest.warns, or pytest.filterwarnings instead. (Deprecated NumPy 2.4)
`; I don't think we're obligated to backport the actual fixes since these are just testsuite annoyances, and execution of the deprecation->removal should fall outside our support window

PR #23615, opened 2025-09-14
TST, MAINT: backports, restore lfilter

tylerjereddy:
This PR against the release branch backports gh-23614 and removes the hacks introduced in gh-23543 now that NumPy has fixed the issue I reported upstream with the patch from https://github.com/numpy/numpy/pull/29745.

Locally, this allows the full testsuite to pass against latest NumPy `main` without having to suppress any of our tests.

This isn't particularly high priority, and `1.16.3` isn't planned, but let's just clean this all up just in case.

### DIFF
--- a/scipy/stats/tests/test_axis_nan_policy.py
+++ b/scipy/stats/tests/test_axis_nan_policy.py
@@ -783,7 +783,7 @@ def test_check_empty_inputs():
                                                                     axis)
                 if output is not None:
                     with np.testing.suppress_warnings() as sup:
-                        sup.filter(RuntimeWarning, "Mean of empty slice.")
+                        sup.filter(RuntimeWarning, "Mean of empty slice")
                         sup.filter(RuntimeWarning, "invalid value encountered")
                         reference = samples[0].mean(axis=axis)
                     np.testing.assert_equal(output, reference)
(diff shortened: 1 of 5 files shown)


tylerjereddy:
oh right, we may need to wait for new NumPy nightlies in a subset of cases, depending on what we mean by "pre"

ilayn:
I reran the tests but I think pre 2.0 and post 2.0 NumPy will create an issue here in the remaining test if I understand the reason correctly.

tylerjereddy:
Ok, I took one more stab at it, let's see how it goes in CI.

tylerjereddy:
ah right, also need to backport gh-23629 now

tylerjereddy:
I suspect this can be merged now, but it is getting late, I'll double check tomorrow.

scipy/stats/tests/test_axis_nan_policy.py
```diff
@@ -870,7 +870,7 @@ def small_sample_generator(n_dims):
                 concat = stats._stats_py._broadcast_concatenate(samples, axis,
                                                                 paired=paired)
                 with np.testing.suppress_warnings() as sup:
-                    sup.filter(RuntimeWarning, "Mean of empty slice.")
+                    sup.filter(RuntimeWarning, r"Mean of empty slice\.?")
```
tylerjereddy, line 873: I was wondering why this one was needed despite its absence in gh-23614, but Matt took care of some more cases in gh-23619, including this one.

I don't know why I ended up having to use a slightly different approach to the punctuation matching, but this is not really important, I'll just backport it since CI passing now, on the off chance we need `1.16.3` at some point.

PR #23616, opened 2025-09-14
BUG: ndimage: fix handling of memory allocation failures.

### LINKED ISSUE #22365: BUG: ndimage: C++ memory management issues in `_rank_filter_1d.cpp`

WarrenWeckesser:
* Allocate the memory used by Mediator in a constructor, and
  free it in the corresponding destructor.  Since a constructor
  can't return an error status, it throws an exception if the
  memory allocation fails.
* In _rank_filter(), the dynamically allocated Mediator instance
  and the work array 'data' are managed with 'unique_ptr'.  The
  exception that might be thrown when either of these is allocated
  is caught and converted to a return value of -1; _rank_filter()
  now returns 0 when it succeeds.
* In rank_filter(), a return value of -1 from _rank_filter() will
  result in a MemoryError exception being set.
* The potential failures of PyArray_FROM_OTF are now handled so
  that a failure of one does not cause the memory  allocated for
  the other to be leaked.

Closes gh-22365.

### DIFF
--- a/scipy/ndimage/src/_rank_filter_1d.cpp
+++ b/scipy/ndimage/src/_rank_filter_1d.cpp
@@ -5,17 +5,47 @@
 #include "Python.h"
 #include "numpy/arrayobject.h"
 
-#include <stdio.h>
-#include <stdlib.h>
+#include <memory>
+#include <new>
 
-struct Mediator // this is used for rank keeping
+//
+// This is used for rank keeping.
+//
+class Mediator
 {
+public:
+  int *mem;  // Single array of ints to hold the memory for `pos` and `heap`.
   int *pos;  // index into `heap` for each value
   int *heap; // max/rank/min heap holding indexes into `data`.
   int N;     // allocated size.
   int idx;   // position in circular queue
   int minCt; // count of items in min heap
   int maxCt; // count of items in max heap
+
+  //
+  // This constructor will throw an exception if memory allocation
+  // fails.  The caller *must* handle this exception so that it
+  // does not propagate out to Python-land.
+  //
+  Mediator(int nItems, int rank) {
+    mem = new int[2 * nItems];  // Might throw std::bad_alloc
+    pos = mem;                  // `pos` uses the first `nItems` elements of `mem`.
+    heap = mem + nItems + rank; // `heap` uses the second `nItems`; it actually
+                                // points to `rank` elements into the second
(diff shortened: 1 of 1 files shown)


WarrenWeckesser:
The change could probably be simplified. I'm moving the PR to *draft* while I look into it.

*Edit:* The idea didn't pan out, so back to "ready for review".

WarrenWeckesser:
I made a small change to compartmentalize the use of the Python C API better.  Otherwise, no significant change from a couple days ago.

ev-br:
Much nicer, thanks Warren!

PR #23620, opened 2025-09-14
MAINT: io: Improve `spmatrix` arg handling and description in `io`

### LINKED ISSUE #23594: DOC: spmatrix argument to scipy.io.loadmat

The documentation states that a `coo_array` is returned if `spmatrix=False`, but that is not universally true. If the underlying mat format is version 4, it is correct since that uses COO natively:

https://github.com/scipy/scipy/blob/bdd3b0e77a3813c22c038c908d992b6de23ffcda/scipy/io/matlab/_mio4.py#L279

but if the underlying format is version 5, the return value is a `csc_array`:

https://github.com/scipy/scipy/blob/bdd3b0e77a3813c22c038c908d992b6de23ffcda/scipy/io/matlab/_mio5_utils.pyx#L802

which makes sense since that mat file version uses CSC natively.

This could be fixed by updating the documentation to state that if `spmatrix=False`, a `coo_array` is returned for mat v4, and a `csc_array` for mat v5.

There may be a deeper issue here though: it seems inefficient to convert a `csc_array` from mat v5 to a `coo_matrix` when `spmatrix=True` (the default), as done by

https://github.com/scipy/scipy/blob/bdd3b0e77a3813c22c038c908d992b6de23ffcda/scipy/io/matlab/_mio.py#L239-L240

Why not convert to a `csc_matrix` in that case? That would recover the legacy behaviour:

https://github.com/scipy/scipy/commit/bddd6df859e2dc7caae5ba39b25b2181d6a976f5#diff-30bb905163b7a3b14deb31286152c114eb491b7cd91f3a5087534a03612290bbL802-R802

_No response_

_No response_

dschult:
Fixes #23594 
Thanks @jornbr

This PR does the following:
- Allow matfile/mio input to convert based on `spmatrix` arg  w/o changing sparse format. Used to convert to COO even when CSC format in the file (MatFile v5).
- Update docs for `spmatrix` arg in for _mio, _mmio, hd, and fast_market_matrix to focus on `spmatrix` vs `sparray` rather than even bringing up the sparse format.

### DIFF
--- a/scipy/io/_fast_matrix_market/__init__.py
+++ b/scipy/io/_fast_matrix_market/__init__.py
@@ -305,7 +305,7 @@ def mmread(source, *, spmatrix=True):
         Matrix Market filename (extensions .mtx, .mtz.gz)
         or open file-like object.
     spmatrix : bool, optional (default: True)
-        If ``True``, return sparse ``coo_matrix``. Otherwise return ``coo_array``.
+        If ``True``, return sparse matrix. Otherwise return sparse array.
 
     Returns
     -------
--- a/scipy/io/_harwell_boeing/hb.py
+++ b/scipy/io/_harwell_boeing/hb.py
@@ -467,7 +467,7 @@ def hb_read(path_or_open_file, *, spmatrix=True):
         If a file-like object, it is used as-is. Otherwise, it is opened
         before reading.
     spmatrix : bool, optional (default: True)
-        If ``True``, return sparse ``coo_matrix``. Otherwise return ``coo_array``.
+        If ``True``, return sparse matrix. Otherwise return sparse array.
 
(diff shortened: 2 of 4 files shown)


Abhijais4896 (commented):
Matrix Market filename (extensions .mtx, .mtz.gz)
        or open file-like object.
    spmatrix : bool, optional (default: True)
        If ``True``, return sparse ``coo_matrix``. Otherwise return ``coo_array``.
        If ``True``, return sparse matrix. Otherwise return sparse array.

    Returns

perimosocordiae (approved):

perimosocordiae:
No comments in the last two weeks, so merging now. Thanks @dschult !

Issue #17194, opened 2022-10-11
BUG: incorrect levy_stable survival function for large values


ivarzap:
### Describe your issue.

levy_stable drops to zero (or very close) for large values of the argument. For not so large, asymptotic theorem 1.2 in Nolan "Univariate Stable Distributions Models for Heavy Tailed Data" Springer 2020 provides the relevant behavior. Scipy's implementation levy_stable.sf(x) follows closely Nolan's result up to a certain point, and suddenly drops to 0, even though before the drop, the levy_stable.sf(x) values where on the order of 1e-4.

The drop value changes with alpha.

The following graph is generated from the Reproducing Code Example below:

![image](https://user-images.githubusercontent.com/25571178/195039011-c3321c7d-f35d-464d-88a1-90589922daab.png)

### Reproducing Code Example

```python
from scipy.stats import levy_stable
from scipy.special import gamma
import numpy as np
import matplotlib.pyplot as plt

levy_stable.parameterization = 'S0'
alpha = 0.93
beta = 1

x = np.linspace(
    levy_stable.ppf(0.90, alpha, beta),
    levy_stable.ppf(0.9999, alpha, beta)*3,
    1000
)
x = x[x>0]
y_sf = levy_stable.sf(x, alpha, beta)

c_alpha = np.sin(np.pi * alpha / 2) * gamma(alpha) / np.pi
y_sf_asymp = c_alpha * (1 + beta) / x ** alpha

plt.plot(x, y_sf, 'r-', lw=5, alpha=0.6, label='sf')
plt.plot(x, y_sf_asymp, 'b-', lw=5, alpha=0.6, label='sf_asymp')
plt.yscale('log')
plt.xscale('log')
plt.legend()
plt.show()
```

### Error message

```shell
The values near the drop point are:
y_sf[330:340] = [3.38289467e-04, 3.37339766e-04, 3.36395581e-04, 1.08801856e-14,
       1.05471187e-14, 9.99200722e-15, 9.54791801e-15, 9.32587341e-15,
       8.99280650e-15, 8.54871729e-15]
```

### SciPy/NumPy/Python version information

1.9.1 1.23.3 sys.version_info(major=3, minor=9, micro=14, releaselevel='final', serial=0)

fbourgey:
I will work on this once https://github.com/scipy/scipy/pull/23622 is completed.

abcde2342:
I have designed a Taylor-expansion-based patch. Please take a look.

```
from __future__ import annotations

[...]

PR #23622, opened 2025-09-14
BUG: stats.levy_stable.logpdf: fix incorrect values for `beta == +-1`

### LINKED ISSUE #20636: BUG: incorrect logpdf values in scipy.stats.levy_stable for beta == +-1
### Describe your issue.

Hello,
I am implementing Levy's alpha-stable distribution for the JAX framework, using scipy as reference and I found some discrepancies.

The tail values of scipy's `levy_stable.logpdf` are incorrect for alpha in (1,2], beta in {-1, +1} and |x| >= 10. In particular, they do not obey the expected exponential law.

See Nolan (2020), Proposition 3.1 page 100 for the correct tail approximation. I have implemented it (in jax) here:
https://github.com/tjhunter/levy-stable-jax/blob/b3e3e05cc572b68742a8d1f0223fd88230836f30/src/levy_stable_jax/distribution.py#L169

Here is an example. In blue is the theoretical tail behaviour (implemented by levy-stable-jax), in orange scipy

image

Link to the notebook that created the plot:
https://github.com/tjhunter/levy-stable-jax/blob/main/notebooks/scipy_issues.ipynb

My suggestion to resolve this issue would be to bypass the  quadrature-based implementation in C for |beta| == 1 and to directly implement the tail approximation as in `levy-stable-jax1. Happy to discuss a possible PR.

### Reproducing Code Example

```python
import seaborn as sns
from scipy.stats import levy_stable as sp_levy_stable
import numpy as np

sp_levy_stable.parameterization = "S0"
(alpha, beta) = (1.6, 1.0)
xs = np.linspace(-15, 10, 100)
sns.lineplot(x=xs,            y=sp_levy_stable.logpdf(xs,alpha=alpha, beta=beta))
```

### Error message

```shell
N/A
```

```shell
1.13.0 1.26.4 sys.version_info(major=3, minor=11, micro=9, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    detection method: pkgconfig
    found: true
    lib directory: /usr/local/lib
    name: openblas
      NO_LAPACK= NO_LAPACKE= NO_AFFINITY=1 USE_OPENMP= ZEN MAX_THREADS=64
    version: 0.3.26.dev
  lapack:
    detection method: pkgconfig
    found: true
    lib directory: /usr/local/lib
    name: openblas
      NO_LAPACK= NO_LAPACKE= NO_AFFINITY=1 USE_OPENMP= ZEN MAX_THREADS=64
    version: 0.3.26.dev
  pybind11:
    detection method: config-tool
    include directory: unknown
    name: pybind11
    version: 2.12.0
Compilers:
  c:
    commands: cc
    linker: ld.bfd
    name: gcc
    version: 10.2.1
  c++:
    commands: c++
    linker: ld.bfd
    name: gcc
    version: 10.2.1
  cython:
    commands: cython
    linker: cython
    name: cython
    version: 3.0.10
  fortran:
    commands: gfortran
    linker: ld.bfd
    name: gcc
    version: 10.2.1
  pythran:
    include directory: ../../tmp/pip-build-env-ilw0s1wu/overlay/lib/python3.11/site-packages/pythran
    version: 0.15.0
Machine Information:
  build:
    cpu: x86_64
    endian: little
    family: x86_64
    system: linux
  cross-compiled: false
  host:
    cpu: x86_64
    endian: little
    family: x86_64
    system: linux
Python Information:
  version: '3.11'
```

fbourgey:
#### Reference issue
Closes #20636

Fix tail asymptotics of logpdf for beta=-1 and beta=1

#### Comments
Expected values in the test were computed via Mathematica.

#### Reproducible example

```python
from scipy.stats import levy_stable
import numpy as np

loc = 0.0
scale = 1.0
(alpha, beta) = (1.6, 1.0)

xs = np.linspace(-15, 0.0, 30)
if beta==-1.0:
    xs *= -1.0

levy_stable.parameterization = "S1"
log_pdf_S1_scipy = levy_stable(alpha, beta, loc=loc, scale=scale).logpdf(xs)

levy_stable.parameterization = "S0"
log_pdf_S0_scipy = levy_stable(alpha, beta, loc=loc, scale=scale).logpdf(xs)

if beta==1.0:
    log_pdf_S1_scipy_old = np.array([
        -68.43963144, -65.49096863, -62.71821277, -60.11736153, -57.68437972,
        -55.41519974, -53.30572237, -51.35181816, -49.54932953, -47.89407275,
        -46.38124626, -44.83221836, -40.12078201, -34.24971308, -28.93942999,
        -24.19115239, -19.98008686, -16.28066725, -13.06671442, -10.31138392,
        -7.98710404,  -6.06550641,  -4.51734947,  -3.31243690,  -2.41953621,
        -1.80630926,  -1.43927776,  -1.28386664,  -1.30459730,  -1.46553721
    ])
    log_pdf_S0_scipy_old = np.array([
        -72.88614827, -69.68370175, -66.66272843, -63.81927226, -61.14934422,
        -58.64892238, -56.31395218, -54.14034706, -52.12398962, -50.26073333,
        -48.54640519, -46.97675954, -45.52808222, -42.51307119, -36.56274208,
        -31.01919155, -26.04640635, -21.62100479, -17.71766602, -14.31046429,
        -11.37283140,  -8.87749786,  -6.79642653,  -5.10073879,  -3.76063464,
        -2.74531030,  -2.02288185,  -1.56033279,  -1.32351906,  -1.27729076
    ])
elif beta==-1.0:
    log_pdf_S1_scipy_old = np.array([
    -1.46553721,  -1.30459730,  -1.28386664,  -1.43927776,  -1.80630926,
    -2.41953621,  -3.31243690,  -4.51734947,  -6.06550641,  -7.98710404,
    -10.31365396, -13.06849249, -16.28208745, -19.98124042, -24.19210317,
    -28.94022648, -34.25060526, -40.14771929, -46.65557006, -53.79771366,
    -61.59729036, -70.07705122, -79.25938194, -89.16632454, -99.81959686,
    -111.24061039,-123.45048640,-136.47007079,-150.31994763,-165.02045163
    ])

    log_pdf_S0_scipy_old = np.array([
    -1.27729076,  -1.32351906,  -1.56033279,  -2.02288185,  -2.74531030,
    -3.76063464,  -5.10073879,  -6.79642653,  -8.88015138, -11.37488254,
    -14.31208425, -17.71896915, -21.62206978, -26.04728904, -31.01994759,
    -36.56482720, -42.70620992, -49.46791323, -56.87332158, -64.94541470,
    -73.70679307, -83.17970083, -93.38604658,-104.34742215,-116.08511970,
    -128.62014739,-141.97324360,-156.16489004,-171.21532384,-187.14454860
    ])
else:
    raise ValueError("beta must be -1.0 or 1.0 to compare with old results")

fig, ax = plt.subplots()
ax.plot(xs, log_pdf_S1_scipy, label="scipy corrected S1")
ax.plot(xs, log_pdf_S0_scipy, label="scipy corrected S0")
ax.plot(xs, log_pdf_S1_scipy_old, "r--", label="scipy S1")
ax.plot(xs, log_pdf_S0_scipy_old, "g--", label="scipy S0")
ax.legend()
ax.set_title(f"alpha={alpha}, beta={beta}")
plt.show()
```
Figure_1
beta_1_levy_log_pdf

### DIFF
--- a/scipy/stats/_levy_stable/__init__.py
+++ b/scipy/stats/_levy_stable/__init__.py
@@ -215,20 +215,29 @@ def _pdf_single_value_piecewise_Z0(x0, alpha, beta, **kwds):
         # cauchy
         return 1 / (1 + x0 ** 2) / np.pi
     elif 1 < alpha < 2 and ((beta == 1.0 and x0 < 0.0) or (beta == -1.0 and x0 > 0.0)):
-        result = _levy_stable_tail_asymptotic_Z0(x0, alpha, beta)
-        if result is not None:
-            return result
+        if kwds.get("asympt", True):
+            # tail asymptotic path for fully skewed tails when appropriate
+            result = _levy_stable_tail_asymptotic_Z0(x0, alpha, beta, **kwds)
+            if result is not None:
+                return result
 
     return _pdf_single_value_piecewise_post_rounding_Z0(
         x0, alpha, beta, quad_eps, x_tol_near_zeta
     )
 
-def _levy_stable_tail_asymptotic_Z0(x0, alpha, beta, cutoff_exponent=8.0):
+def _levy_stable_tail_asymptotic_Z0(
+    x0, alpha, beta, cutoff_exponent=8.0, rel_error_tol=1e-1, **kwds
+):
     """
     Handle tail asymptotic for beta=-1 (right-tail) or beta=1 (left-tail), 1<alpha<2, as
     detailed in [NO] Proposition 3.1 (pg 100). Returns None if not in asymptotic region.
     """
     x0 += beta * np.tan(np.pi * alpha / 2)  # Convert to Z1
+    # numerical (non-asymptotic) reference computed using piecewise Z1 method
+    pdf_non_asympt_z1 = _pdf_single_value_piecewise_Z1(
+        x0, alpha, beta, asympt=False, **kwds
+    )
+    # For left-tail asymptotic (beta == 1) flip sign so x_tail is positive
     if beta == 1:
         x0 = -x0
 
@@ -238,10 +247,17 @@ def _levy_stable_tail_asymptotic_Z0(x0, alpha, beta, cutoff_exponent=8.0):
     c1 /= (2 * np.pi * np.abs(1 - alpha)) ** 0.5
     c2 = np.abs(1 - alpha) * (alpha**alpha / abs_cos) ** (1 / (1 - alpha))
 
-    exponent = c2 * x0**(alpha / (alpha - 1))
-    if exponent >= cutoff_exponent:  # far enough in the tail (empirically)
+    exp_arg = c2 * x0 ** (alpha / (alpha - 1))
+    if exp_arg >= cutoff_exponent:  # far enough in the tail (empirically)
         a = (2.0 - alpha) / (2.0 * alpha - 2.0)
-        return c1 * x0**a * np.exp(-exponent)
+        pdf_asympt_z1 = c1 * x0**a * np.exp(-exp_arg)
+        # check relative error against non-asymptotic and return asymptotic if
+        # appropriate
+        rel_err = (
+            abs(pdf_non_asympt_z1 / pdf_asympt_z1 - 1) if pdf_asympt_z1 != 0 else np.inf
+        )
+        if rel_err > rel_error_tol:
+            return pdf_asympt_z1
     return None
 
 
--- a/scipy/stats/_levy_stable/__init__.py
+++ b/scipy/stats/_levy_stable/__init__.py
@@ -214,39 +214,34 @@ def _pdf_single_value_piecewise_Z0(x0, alpha, beta, **kwds):
     elif alpha == 1.0 and beta == 0.0:
         # cauchy
         return 1 / (1 + x0 ** 2) / np.pi
-    elif 1 < alpha < 2:
-        in_tail_region = (beta == 1.0 and x0 < 0.0) or (beta == -1.0 and x0 > 0.0)
-        if in_tail_region:
-            result = _levy_stable_tail_asymptotic(x0, alpha, beta)
-            if result is not None:
-                return result
+    elif 1 < alpha < 2 and ((beta == 1.0 and x0 < 0.0) or (beta == -1.0 and x0 > 0.0)):
+        result = _levy_stable_tail_asymptotic_Z0(x0, alpha, beta)
+        if result is not None:
+            return result
 
     return _pdf_single_value_piecewise_post_rounding_Z0(
         x0, alpha, beta, quad_eps, x_tol_near_zeta
     )
 
-def _levy_stable_tail_asymptotic(x0, alpha, beta, cutoff_exponent=8.0):
+def _levy_stable_tail_asymptotic_Z0(x0, alpha, beta, cutoff_exponent=8.0):
     """
-    Handle tail asymptotic for beta=-1 (right-tail) or beta=1 (left-tail), 1<alpha<=2.
-    Returns None if not in asymptotic region.
-    Nolan (Proposition 3.1).
+    Handle tail asymptotic for beta=-1 (right-tail) or beta=1 (left-tail), 1<alpha<2, as
+    detailed in [NO] Proposition 3.1 (pg 100). Returns None if not in asymptotic region.
     """
-    if not (1.0 < alpha < 2.0) or beta not in (-1.0, 1.0):
-        return None
-    x0 += beta * np.tan(np.pi * alpha / 2)
+    x0 += beta * np.tan(np.pi * alpha / 2)  # Convert to Z1
     if beta == 1:
         x0 = -x0
-    # Constants (same for both tails)
-    p = alpha / (alpha - 1)
-    a = (2.0 - alpha) / (2.0 * alpha - 2.0)
-    cosa = np.cos(np.pi * alpha / 2)
-    c2 = np.abs(1 - alpha) * (alpha**alpha / np.abs(cosa)) ** (1 / (1 - alpha))
-    c1 = (alpha / abs(cosa)) ** (1 / (2 - 2 * alpha))
-    c1 /= (2 * np.pi * np.abs(1 - alpha)) ** 0.5
 
-    if c2 * x0**p >= cutoff_exponent:
-        # far enough in the tail
-        return c1 * x0**a * np.exp(-c2 * (x0**p))
+    # Use abs rather than assuming alpha > 1 to facilitate future generalization
+    abs_cos = np.abs(np.cos(np.pi * alpha / 2))
+    c1 = (alpha / abs_cos) ** (1 / (2 - 2 * alpha))
+    c1 /= (2 * np.pi * np.abs(1 - alpha)) ** 0.5
+    c2 = np.abs(1 - alpha) * (alpha**alpha / abs_cos) ** (1 / (1 - alpha))
+
+    exponent = c2 * x0**(alpha / (alpha - 1))
+    if exponent >= cutoff_exponent:  # far enough in the tail (empirically)
+        a = (2.0 - alpha) / (2.0 * alpha - 2.0)
+        return c1 * x0**a * np.exp(-exponent)
     return None
 
 
--- a/scipy/stats/_levy_stable/__init__.py
+++ b/scipy/stats/_levy_stable/__init__.py
@@ -214,10 +214,40 @@ def _pdf_single_value_piecewise_Z0(x0, alpha, beta, **kwds):
     elif alpha == 1.0 and beta == 0.0:
         # cauchy
         return 1 / (1 + x0 ** 2) / np.pi
+    elif 1 < alpha < 2:
+        in_tail_region = (beta == 1.0 and x0 < 0.0) or (beta == -1.0 and x0 > 0.0)
+        if in_tail_region:
+            result = _levy_stable_tail_asymptotic(x0, alpha, beta)
+            if result is not None:
+                return result
 
     return _pdf_single_value_piecewise_post_rounding_Z0(
         x0, alpha, beta, quad_eps, x_tol_near_zeta
     )
+
+def _levy_stable_tail_asymptotic(x0, alpha, beta, cutoff_exponent=8.0):
+    """
+    Handle tail asymptotic for beta=-1 (right-tail) or beta=1 (left-tail), 1<alpha<=2.
+    Returns None if not in asymptotic region.
+    Nolan (Proposition 3.1).
+    """
+    if not (1.0 < alpha < 2.0) or beta not in (-1.0, 1.0):
+        return None
+    x0 += beta * np.tan(np.pi * alpha / 2)
+    if beta == 1:
+        x0 = -x0
+    # Constants (same for both tails)
+    p = alpha / (alpha - 1)
+    a = (2.0 - alpha) / (2.0 * alpha - 2.0)
+    cosa = np.cos(np.pi * alpha / 2)
+    c2 = np.abs(1 - alpha) * (alpha**alpha / np.abs(cosa)) ** (1 / (1 - alpha))
+    c1 = (alpha / abs(cosa)) ** (1 / (2 - 2 * alpha))
+    c1 /= (2 * np.pi * np.abs(1 - alpha)) ** 0.5
+
+    if c2 * x0**p >= cutoff_exponent:
+        # far enough in the tail
+        return c1 * x0**a * np.exp(-c2 * (x0**p))
+    return None
 
 
 def _pdf_single_value_piecewise_post_rounding_Z0(x0, alpha, beta, quad_eps,
--- a/scipy/stats/tests/test_distributions.py
+++ b/scipy/stats/tests/test_distributions.py
@@ -6127,6 +6127,23 @@ class TestLevyStable:
                     verbose=False
                 )
 
+    @pytest.mark.parametrize(
+        "param, alpha, beta, xs, expected",
+        [
+            (0, 1.6, 1.0, [-15.0, -14.0, -13.0], [-187.14442, -157.14116, -130.35267]),
+            (1, 1.6, 1.0, [-15.0, -14.0, -13.0], [-165.02031, -137.36717, -112.82235]),
+            (0, 1.4, -1.0, [13.0, 14.0, 15.0], [-367.19433, -464.70496, -579.43993]),
+            (1, 1.4, -1.0, [13.0, 14.0, 15.0], [-258.12165, -334.61399, -426.07241]),
+        ],
+    )
+    def test_logpdf_tail(self, param, alpha, beta, xs, expected):
+        # gh-20636
+        # Test levy_stable.logpdf accuracy in the tails when beta=1 or beta=-1
+        # for both S0/S1 parameterizations.
+        stats.levy_stable.parameterization = f"S{param}"
+        logpdfs = stats.levy_stable(alpha=alpha, beta=beta).logpdf(xs)
+        assert_allclose(logpdfs, expected, rtol=1e-5)
+
     @pytest.mark.parametrize("param", [0, 1])
     @pytest.mark.parametrize("case", ["pdf", "cdf"])
     def test_location_scale(
--- a/scipy/stats/tests/test_distributions.py
+++ b/scipy/stats/tests/test_distributions.py
@@ -6130,16 +6130,42 @@ class TestLevyStable:
     @pytest.mark.parametrize(
         "param, alpha, beta, xs, expected",
         [
-            (0, 1.6, 1.0, [-15.0, -14.0, -13.0], [-187.14442, -157.14116, -130.35267]),
-            (1, 1.6, 1.0, [-15.0, -14.0, -13.0], [-165.02031, -137.36717, -112.82235]),
-            (0, 1.4, -1.0, [13.0, 14.0, 15.0], [-367.19433, -464.70496, -579.43993]),
-            (1, 1.4, -1.0, [13.0, 14.0, 15.0], [-258.12165, -334.61399, -426.07241]),
+            (
+                0,
+                1.6,
+                1.0,
+                [-15.0, -14.0, -13.0],
+                [-187.1444263, -157.1411663, -130.3526688],
+            ),
+            (
+                1,
+                1.6,
+                1.0,
+                [-15.0, -14.0, -13.0],
+                [-165.0203129, -137.3671717, -112.8223545],
+            ),
+            (
+                0,
+                1.4,
+                -1.0,
+                [13.0, 14.0, 15.0],
+                [-367.1943293, -464.7049565, -579.4399286],
+            ),
+            (
+                1,
+                1.4,
+                -1.0,
+                [13.0, 14.0, 15.0],
+                [-258.1216497, -334.6139857, -426.0724094],
+            ),
         ],
     )
     def test_logpdf_tail(self, param, alpha, beta, xs, expected):
         # gh-20636
+        # Expected values computed from Mathematica.
         # Test levy_stable.logpdf accuracy in the tails when beta=1 or beta=-1
         # for both S0/S1 parameterizations.
+        # Testing logpdf instead of pdf as more compact.
         stats.levy_stable.parameterization = f"S{param}"
         logpdfs = stats.levy_stable(alpha=alpha, beta=beta).logpdf(xs)
         assert_allclose(logpdfs, expected, rtol=1e-5)


scipy/stats/_levy_stable/__init__.py
```diff
@@ -214,11 +214,41 @@ def _pdf_single_value_piecewise_Z0(x0, alpha, beta, **kwds):
...
 
     return _pdf_single_value_piecewise_post_rounding_Z0(
         x0, alpha, beta, quad_eps, x_tol_near_zeta
     )
 
+def _levy_stable_tail_asymptotic(x0, alpha, beta, cutoff_exponent=8.0):
+    """
+    Handle tail asymptotic for beta=-1 (right-tail) or beta=1 (left-tail), 1<alpha<=2.
+    Returns None if not in asymptotic region.
+    Nolan (Proposition 3.1).
```
mdhaber: Which of the references is this? Do you have a link? If not, do you think posting the relevant portion.(e.g. screenshot) would be covered under fair use?

I'm there is not a recommendations for `cutoff_exponent` in the reference, where did this value come from? Is it robust for different values if $\alpha$?

scipy/stats/tests/test_distributions.py
```diff
@@ -6109,6 +6109,23 @@ def test_cdf_nolan_samples(
                     verbose=False
                 )
 
+    @pytest.mark.parametrize(
+        "param, alpha, beta, xs, expected",
+        [
+            (0, 1.6, 1.0, [-15.0, -14.0, -13.0], [-187.14442, -157.14116, -130.35267]),
```
mdhaber: Can you provide example Mathematics code as a comment in the code? Are more decimal places available, or would they not match?

And why test log-PDF instead of PDF? Just more compact?

scipy/stats/_levy_stable/__init__.py
fbourgey: Proposition 3.1 (page 100) of [Nolan, 2020, Univariate Stable Distributions](https://doi.org/10.1007/978-3-030-52915-4) 

I chose this `cutoff_exponent` arbitrarily. When `c2 * x0**p >= cutoff_exponent`, the exponential is extremely small.

scipy/stats/tests/test_distributions.py
fbourgey: Mathematica code I used

```mathematica
x = Log @ PDF[
    StableDistribution[0, 1.6, 1, 0, 1],
    {-15, -14, 13}
];
SetPrecision[x, 8]
```

gives

```mathematica
{-187.14443,-157.14117,-130.35267}
```

Should I increase the precision? I need to update the `expected` array. I made a small typo.

No special reason for testing log-pdf vs. pdf. Probably more compact yes.

mdhaber: > Should I increase the precision?

If that would allow us to test with default (or better) `rtol`, yes. If not, fine to leave them with fewer digits and comment on why the relative error tolerance is so coarse. (Mathematica does something more accurate than this approximation, I'd guess?)

Might also leave a comment in the code about why `logpdf` is tested. Fine if that's just because it's more compact.

scipy/stats/_levy_stable/__init__.py
mdhaber: Ok, can you confirm that it is high enough to prevent the completely incorrect values produced by `main` for the full range of $\alpha$? If not, that's fine, but it would just be good to know.

> the exponential is extremely small.

Does this imply that the asymptotic approximation will have good relative accuracy, or does it just suggest the absolute accuracy will be good (because the absolute value of the result will be small).

If so, I think I can check the math and merge shortly.

scipy/stats/_levy_stable/__init__.py
```diff
@@ -214,11 +214,41 @@ def _pdf_single_value_piecewise_Z0(x0, alpha, beta, **kwds):
...
+        if in_tail_region:
+            result = _levy_stable_tail_asymptotic(x0, alpha, beta)
+            if result is not None:
+                return result
 
     return _pdf_single_value_piecewise_post_rounding_Z0(
         x0, alpha, beta, quad_eps, x_tol_near_zeta
     )
 
+def _levy_stable_tail_asymptotic(x0, alpha, beta, cutoff_exponent=8.0):
```
mdhaber: If anyone else is interested in checking this, here is the relevant math:

image

Note that this implementation is only for the PDF when $1 < \alpha < 2$, but both PDF and CDF approximations are available for the full range of $0 < \alpha < 2$.

fbourgey: I will submit a PR very soon for #17194 using the asymptotic shown here for the CDF.

mdhaber: Ok. Do things already work ok for $\alpha \leq 1$?

scipy/stats/_levy_stable/__init__.py
fbourgey: > Ok, can you confirm that it is high enough to prevent the completely incorrect values produced by main for the full range of $\alpha$? If not, that's fine, but it would just be good to know.

I checked numerically on many different combinations of $\alpha$. It seems good.

> Does this imply that the asymptotic approximation will have good relative accuracy, or does it just suggest the absolute accuracy will be good (because the absolute value of the result will be small).

I'd say absolute accuracy. The `cutoff exponent` is arbitrary.

scipy/stats/_levy_stable/__init__.py
fbourgey: For CDF/SF or for the PDF?

scipy/stats/tests/test_distributions.py
fbourgey: Suggested something.

scipy/stats/_levy_stable/__init__.py
mdhaber: PDF. It's fine to limit the scope of this PR to $\alpha > 1$; just wanted to know whether the $\alpha \leq 1$ approximation should be considered for future work.

fbourgey: At a first glance (e.g., $\alpha=0.3$), this needs fixing as well. 

@mdhaber, should I raise a new issue?

scipy/stats/_levy_stable/__init__.py
```diff
@@ -214,11 +214,41 @@ def _pdf_single_value_piecewise_Z0(x0, alpha, beta, **kwds):
...
+    """
+    Handle tail asymptotic for beta=-1 (right-tail) or beta=1 (left-tail), 1<alpha<=2.
+    Returns None if not in asymptotic region.
+    Nolan (Proposition 3.1).
+    """
+    if not (1.0 < alpha < 2.0) or beta not in (-1.0, 1.0):
+        return None
```
mdhaber:
```suggestion
```
I don't think we need to check this again.

scipy/stats/_levy_stable/__init__.py
```diff
@@ -214,11 +214,41 @@ def _pdf_single_value_piecewise_Z0(x0, alpha, beta, **kwds):
     elif alpha == 1.0 and beta == 0.0:
         # cauchy
         return 1 / (1 + x0 ** 2) / np.pi
+    elif 1 < alpha < 2:
```
mdhaber:
```suggestion
    elif (1 < alpha < 2) and ((beta == 1.0 and x0 < 0.0) or (beta == -1.0 and x0 > 0.0)):
```
I think we can reduce this from three branching statements to one without sacrificing much.

scipy/stats/_levy_stable/__init__.py
```diff
@@ -214,11 +214,41 @@ def _pdf_single_value_piecewise_Z0(x0, alpha, beta, **kwds):
...
+            if result is not None:
+                return result
 
     return _pdf_single_value_piecewise_post_rounding_Z0(
         x0, alpha, beta, quad_eps, x_tol_near_zeta
     )
 
+def _levy_stable_tail_asymptotic(x0, alpha, beta, cutoff_exponent=8.0):
+    """
+    Handle tail asymptotic for beta=-1 (right-tail) or beta=1 (left-tail), 1<alpha<=2.
```
mdhaber:
```suggestion
    Handle tail asymptotic for beta=-1 (right-tail) or beta=1 (left-tail), 1<alpha<2.
```
Reference says $\alpha < 2$ and so does implementation.

scipy/stats/_levy_stable/__init__.py
```diff
@@ -214,11 +214,41 @@ def _pdf_single_value_piecewise_Z0(x0, alpha, beta, **kwds):
...
+    """
+    if not (1.0 < alpha < 2.0) or beta not in (-1.0, 1.0):
+        return None
+    x0 += beta * np.tan(np.pi * alpha / 2)
+    if beta == 1:
+        x0 = -x0
+    # Constants (same for both tails)
+    p = alpha / (alpha - 1)
+    a = (2.0 - alpha) / (2.0 * alpha - 2.0)
+    cosa = np.cos(np.pi * alpha / 2)
```
mdhaber: `cosa` is not `cos(a)`, and we might as well include the absolute value[^1], so I'll rename that.
I was also a bit confused about where `p` and `a` came from or were used until I got to the end, so I'll move those definitions. I'll also write `2.0` as `2` for consistency with all the rest.

[^1]: I will leave the absolute values in rather than using `-` in case we want to generalize this to $\alpha \leq 1$.

scipy/stats/_levy_stable/__init__.py
```diff
@@ -214,11 +214,36 @@ def _pdf_single_value_piecewise_Z0(x0, alpha, beta, **kwds):
...
     return _pdf_single_value_piecewise_post_rounding_Z0(
         x0, alpha, beta, quad_eps, x_tol_near_zeta
     )
 
+def _levy_stable_tail_asymptotic_Z0(x0, alpha, beta, cutoff_exponent=8.0):
+    """
+    Handle tail asymptotic for beta=-1 (right-tail) or beta=1 (left-tail), 1<alpha<2, as
+    detailed in [NO] Proposition 3.1 (pg 100). Returns None if not in asymptotic region.
+    """
+    x0 += beta * np.tan(np.pi * alpha / 2)  # Convert to Z1
```
mdhaber, line 235: This converts back to Z1, right @fbourgey?

scipy/stats/_levy_stable/__init__.py
```diff
@@ -214,39 +214,34 @@ def _pdf_single_value_piecewise_Z0(x0, alpha, beta, **kwds):
     elif alpha == 1.0 and beta == 0.0:
         # cauchy
         return 1 / (1 + x0 ** 2) / np.pi
-    elif 1 < alpha < 2:
-        in_tail_region = (beta == 1.0 and x0 < 0.0) or (beta == -1.0 and x0 > 0.0)
-        if in_tail_region:
-            result = _levy_stable_tail_asymptotic(x0, alpha, beta)
-            if result is not None:
-                return result
+    elif 1 < alpha < 2 and ((beta == 1.0 and x0 < 0.0) or (beta == -1.0 and x0 > 0.0)):
```
mdhaber, line 217: Should we be in the Z0 parameterization when these comparisons involving `x0` are performed? (Maybe not - just checking whether this was intentional.)

mdhaber (commented):
I think this is correct, and I think this typically improves the behavior in the left tail when $|\beta| = 1$ and $\alpha > 1$. However, this runs into some numerical issues and may make results a little worse when $\alpha$ is close to $1$. 

For example, when `alpha = 1.0065`, we are above the threshold for which `alpha` is set to `1.0`, and we avoid some other errors I get at `alpha = 1.006`, but this approximation causes the `logpdf` to jump from approximately `-2` to `-inf` around what should be the *mode* of the distribution. On one hand, it looks we're already not very accurate for $\alpha$ so close to 1 anyway, so maybe we could just let that go for now. But then again, it highlights that the `exponent > cutoff_exponent` logic for when we are in the desired tail is not robust.

I'd be willing to merge if there were a more systematic study of the criteria for using the approximation across the range of $\alpha$; otherwise, this (and follow-up PRs) will need another reviewer.

scipy/stats/_levy_stable/__init__.py
fbourgey: correct

scipy/stats/_levy_stable/__init__.py
fbourgey: I don't think so.

mdhaber: OK, well we *are* in the Z0 parameterization when this is performed. So if we should be in Z1, that needs to be taken into account. I can reason about why this check is what it is in the Z1 parameterization; it's not quite as obvious if we're in Z0.

fbourgey: should we then do?
```python
beta == 1.0 and x0 + beta * np.tan(np.pi * alpha / 2) < 0.0
```
and 
```python
beta == -1.0 and x0 + beta * np.tan(np.pi * alpha / 2) > 0.0
```

The reason I added those conditions was to use Proposition 3.1., c. from [NO]

mdhaber: That is the logic, but I wouldn't write it like that. I can suggest something later.

In the meantime, I'd be interested in that "systematic study of the criteria for using the approximation". For example, we could begin by comparing the accuracy of SciPy main and the existing PR against reference values for several values of $\alpha$, especially values for which we expect there might be numerical trouble (like $\alpha \approx 1$). This might take the form of plots like in your top post (https://github.com/scipy/scipy/pull/23622#issue-3415718560), but maybe these would show the relative error vs the reference (e.g. Mathematica) instead of the logpdf value itself. It would also be useful to be able to see on the plots where the approximation starts being used. I  think we only need to study one of the two parameterizations for now.

This will tell us when the PDF value returned in the PR is less accurate than in `main` and give us a clues as to how we might want to adjust the threshold for using the approximation. We might iterate on the criterion for switching to the asymptotic approximation until we're comfortable with the results.

fbourgey: OK, I'll propose something shortly.

fbourgey:
Following your comment https://github.com/scipy/scipy/pull/23622#pullrequestreview-3235489898, it appears that 

```python
from scipy.stats import levy_stable
levy_stable.parameterization = "S1"
levy_stable(1.006, beta=1).pdf(0.1)
```
raises an error
```python
ValueError: The function value at x=0.009368566561400948 is NaN; solver cannot continue.
```

I guess, an issue should be raised for this?

scipy/stats/_levy_stable/__init__.py
fbourgey: Screenshot 2025-09-18 at 9 02 52 AM
Screenshot 2025-09-18 at 9 03 02 AM
Screenshot 2025-09-18 at 9 03 14 AM


I have tested for $\alpha \in [1.001, 1.007, 1.01, 1.07]$. It appears that both the `main` and `PR` scipy versions give the same results for the first three, and differ for $\alpha=1.07$. 

The grid runs from $-15$ to $0$ and this is for $\beta=1$ and $S1$.

mdhaber: Thanks, but some of these are not in the left tail of the distribution. That's one reason it would be helpful to see where the approximation is being used. 
Also, I think $\alpha=1.001$ will actually result in a separate $\alpha=1$ code path being used. Need to stay above $\alpha=1.005$, apparently.

mdhaber:
> I guess, an issue should be raised for this?

I imagine that if you had to stop to file a new issue for every case you ran into here, you wouldn't make it very far! Maybe take note of things here in the PR, then file a single issue at the end?

scipy/stats/_levy_stable/__init__.py
fbourgey: Screenshot 2025-09-18 at 1 22 57 PM
Screenshot 2025-09-18 at 1 22 43 PM


I tried for `S1` and $\alpha=1.007,\beta=1$.

```python
xs = np.array([-96. , -95.4, -94.8, -94.2, -93.6, -93. , -92.4, -91.8, -91.2,
       -90.6, -90. ])
```

The left-tail regime kicks in for:

```python
x0: -96.0
c2 * (-x0)**p = 557.0587288495165

x0: -95.4
c2 * (-x0)**p = 226.046799937591

x0: -94.8
c2 * (-x0)**p = 91.2062114949978

x0: -94.2
c2 * (-x0)**p = 36.58876587766649

x0: -93.6
c2 * (-x0)**p = 14.592725514798076
```

I find that the cutoff is `-(cutoff_exponent / c2)**(1/p) = -93.20972488161942`

mdhaber: Great. What I'm seeing is that the cutoff is close to where the relative error in `main` gets really bad, and that the asymptotic approximation saves it. The relative error of the approximation is not terrific, so there is big jump in the error of the bottom plot at the transition, but the idea is that it's better to suffer some relative error there than to risk extreme relative error at points that are further negative. How do things look at the other values of $\alpha$?

mdhaber:
I thought I should try to create some plots of my own. For the parameters in your top post, I'm plotting the pdf produced by Mathematica, SciPy (main), and the asymptotic approximation - just the asymptotic approximation, without switching between implementations. The point where `cutoff_exponent=8` would cause the switch between SciPy's existing behavior and the asymptotic approximation is marked with an `*`.

image

We see the same divergence between SciPy and the Mathematica reference at low values of `x` that you do in the top post. The asymptotic approximation curve is invisible, so it's tracking the Mathematica curve well, and it's much smoother, so we think SciPy's behavior looks bad.

---

But look at the plot of the absolute error:

image

and the relative error:
image

What we see now is that unless we're in the *very* deep tail, the asymptotic approximation has much larger absolute and relative error. So unless you're actually interested in the very deep tails where the PDF is tiny (<`1e-16`), the asymptotic approximation is much worse.

This is why I'm concerned about the use of an arbitrary threshold for the transition to the asymptotic approximation.

mdhaber:
Maybe better transition logic is based on the observed behavior of the plots - SciPy's calculation of the PDF is too high in the deep tail. Maybe we also detect when the PDF produced by SciPy (main) is much greater than that produced by the asymptotic approximation, and only then do we return the approximation.

fbourgey:
Thanks for all those tests. 

> Maybe better transition logic is based on the observed behavior of the plots - SciPy's calculation of the PDF is too high in the deep tail. Maybe we also detect when the PDF produced by SciPy (main) is much greater than that produced by the asymptotic approximation, and only then do we return the approximation.

Something like?

- Start far from the left tail and iterate until $| \frac{f_{\text{scipy}}(x)}{f_{\text{asympt}}(x)} - 1| < \texttt{rtol}$.
- The resulting cutoff $x_{\text{cutoff}}$ tells us when to use $f_{\text{asympt}}$.

$f$ is PDF here.

mdhaber:
> Something like?

No, I don't think we would want iteration.

Here are plots for a few different values of $\alpha$ that show the two PDFs. The `*` is where the SciPy PDF becomes 10x larger than the approximation.

$\alpha = 1.01$
image

$\alpha = 1.25$
image

$\alpha = 1.7$
image

So, for instance, you know you're in the left tail (e.g. with your `cutoff_exponent` criterion), you could calculate both the usual SciPy PDF *and* the approximation (which is relatively inexpensive), and if the SciPy PDF is much larger than the approximation, you return the approximation. This does a pretty good job of tracking the point where SciPy's existing implementation becomes worse than the approximation. It might leave a small region where the PDF is not very accurate, but it's better than the very large region we have now. I suspect one could do a little better with some tweaking.

fbourgey:
This would still depend on `cutoff_exponent` and an additional parameter for testing how *much larger* the scipy PDF is compared to the approximation. What would be the much larger? I can propose something.

mdhaber:
> What would be the much larger? I can propose something.

Yeah, you'd need to work out the details and demonstrate that it's robust across the usable range of $\alpha$. This is just a suggestion and a demonstration of how I might convince someone that this is a reasonable direction to go in. You can do something different if you find a better idea that is not computationally expensive or extremely complicated. 

Once you have something that you think works well, please produce similar plots to [those above](https://github.com/scipy/scipy/pull/23622#issuecomment-3312452486) that show that the absolute abd relative error stay pretty good for a wide range of $\alpha$ (except potentially in a narrow region of $x$ close to the transition).

An alternative would be to find out what goes wrong with the existing calculation and fix it rather than adding another calculation path.

fbourgey:
When after the `cutoff_exponent` and the relative error is above 10%, we return the asymptotic PDF. It might be useful to also add more tests when far in the tail?

Case $\alpha=1.01$

PDF:

Screenshot 2025-09-22 at 12 44 39 PM

Relative error:

Screenshot 2025-09-22 at 12 56 39 PM

Screenshot 2025-09-22 at 12 56 28 PM



Case $\alpha=1.25$

PDF:
Screenshot 2025-09-22 at 12 45 23 PM

Relative error:

Screenshot 2025-09-22 at 12 57 18 PM


Case $\alpha =1.7$

PDF:

Screenshot 2025-09-22 at 12 45 05 PM

Relative error:

Screenshot 2025-09-22 at 12 57 25 PM

fbourgey:
@mdhaber : 
Another idea could be to have different `cutoff_exponent` parameters depending on $\alpha$. The closer $\alpha$ is to 1, the larger `cutoff_exponent` would be. We can also consider a grid of `cutoff_exponent` and compare the different relative errors `pdf_scipy / pdf_asympt`.

mdhaber:
Sure. It's just that figuring out exactly how `cutoff_exponent` should vary based on $\alpha$ is challenging, and it's not clear that `cutoff_exponent` is the best predictor of whether the asymptotic approximation is better or not. If the benefit is worth the extra complexity, though, please show that in the updated PR.

fbourgey:
Yes, this seems challenging, and I have no solution just now. Regarding https://github.com/scipy/scipy/pull/23622#issuecomment-3320158424, what do you think?

fbourgey:
@mdhaber, following on the below: do you think the changes are enough? Thanks!

mdhaber:
Thanks. What you have may be fine, but I am working on other things and would prefer not to continue developing this solution. I would be willing to review a final version of the PR that demonstrates that it fixes the accuracy in the problem area (deep tails) without introducing significant inaccuracies elsewhere. Ideally, the demonstration would be convincing at a glance - e.g. 3D or colormap versions of the plots above showing sweeps across both $\alpha$ and $x$. (Sorry that would be even more work - I've just become extra cautious about this one because it looked like we had a solution before, but it may have done more harm than good. I want to avoid that.)

dschmitz89:
@mdhaber I can take over the review here, it's been a while. Will take me some time to catch up, this distribution is a real mess ... :/ I agree that a plot would be very helpful to assess the improvements here.

fbourgey:
@mdhaber, @dschmitz89, Thanks. Just to be clear, what is the precise colormap you want?

mdhaber:
I had in mind:
- before (`main`) and after (PR branch) and
- absolute and relative error of pdf (or logpdf)
colormaps covering $1 < \alpha < 2$ and $x$ over a relevant range.

But plotting $x$ over a "relevant range" might be challenging because the range of $x$ that is of interest depends on $\alpha$. One option, would be to choose a very broad range that will be OK for all $\alpha$. The trouble is that this might take a long time to compute, or it might be too zoomed out to easily see the parts of the plot we're interested in.

So maybe you'll find that just several plots like you have above:

image

Would be more effective. There would just need to be enough to show that there are not places (e.g. ranges of $\alpha$) where the accuracy is getting worse. Or, if the relative accuracy is bad, show that at least the absolute accuracy is very good. (For that reason, maybe it would help to also show absolute accuracy on the same axes?)

The key thing is probably figuring out how to reliably get the limits of the plot to show the interesting portion.

image

big enough that it can be interpreted. If there were a script (or even better - a notebook) that would let us specify a value of $\alpha$ (e.g. with a slider) and then see this sort of plot, that's OK if we have to wait a bit for it to be generated. 

But let's wait for @dschmitz89 to respond, since he mentioned he'd take over.

dschmitz89:
@fbourgey A 3D plot of the relative error of the pdf for a range of alpha and x is what I also had in mind.

fbourgey:
$x$ ranges from $-70$ to $-1$ with a step of $1$ and $\alpha$ ranges from 1.01 to 1.99 with a step of 0.01.

I compare main vs. Mathematica and PR vs. Mathematica. Below are the two colormaps. 
rel_err_main
rel_err_pr



@dschmitz89 @mdhaber, what do you think?

JacobHass8:
I think that there might be a much simpler solution. On lines 275 to 294 of [https://github.com/scipy/scipy/blob/main/scipy/stats/_levy_stable/__init__.py](url) there is the following code:
```python
# lastly, we add additional samples at
#   ~exp(-100), ~exp(-10), ~exp(-5), ~exp(-1)
# to improve QUADPACK's detection of rapidly descending tail behavior
# (this choice is fairly ad hoc)
tail_points = [
    optimize.bisect(lambda t: g(t) - exp_height, -xi, np.pi / 2)
    for exp_height in [100, 10, 5]
    # exp_height = 1 is handled by peak
]
intg_points = [0, peak] + tail_points
intg, *ret = integrate.quad(
    integrand,
    -xi,
    np.pi / 2,
    points=intg_points,
    limit=100,
    epsrel=quad_eps,
    epsabs=0,
    full_output=1,
)
```
I'm not exactly sure what the `integrand` function looks like, but I've been playing around with how the integration is computed. It seems like omitting `points=intg_points` matches the behavior of the asymptotic approximation [here](https://github.com/scipy/scipy/pull/23622#issuecomment-3312452486) for $\alpha=1.61$. Unfortunately, not providing input points makes the integration unstable for $\alpha \approx 1$. Something as simple as `points=np.linspace(-xi, np.pi/2, num=50)` seems to retain precision in the tail for $\alpha \approx 1$. I think what is going wrong is that  `optimize.bisect` is not finding points on the correct side of the peak. 

This has all been pretty preliminary though. I'll write up a solution and compare it to Mathematica.

JacobHass8:
Here's a couple of plots using different integration points compared to the asymptotic approximation. 
Figure_1
Figure_2
As you can see, the "S0" parameterization behaves well for all $\alpha$. However, the "S1" parameterization breaks down for $\alpha \rightarrow 1$.  The "S1" parameterization is really delicate delicate for some reason. 

The closest solution I've been able to get is letting the number of integration subintervals be $\frac{10}{\alpha -1)}$. When $\alpha$ gets larger you decrease the number of subintervals needed and when $alpha \rightarrow 1$, you increase the number of subintervals. This works for a lot of points except when $x$ is close to $0$ for some reason. 

I'm more inclined to go with the asymptotic approximation at this point. The integration just requires a lot of finesse that maybe we don't need.

JacobHass8:
There's something wrong going on with $\alpha \approx 1$, $x\approx 0$ and the S1 parameterization. Using this current PR, here's what I'm getting for the PDF. 

S1IfStatementPR

JacobHass8:
I implemented a more accurate integration and extended the bounds on some approximations to get this: 

S1IfStatementPR

One issue is that it's a lot slower than the current implementation. It could probably be optimized though.

mdhaber:
> @dschmitz89 @mdhaber, what do you think?

We discussed above that @dschmitz89 would be taking over, so I haven't responded. 

But I'm ready to think about this again, so I will say that the plot looks like a great summary. However, it looks like the relative error is quite often around `1e-1`, which is not terrific, so it may be worth seeing how gh-24314 does. Can you provide the code to produce that plot? Regardless of the fix, the visualization is super valuable.

dschmitz89:
@JacobHass8 @mdhaber @fbourgey sry, this PR fell from my radar. I am a little confused: could we run the same 2D relative error check plot on this and @JacobHass8 PR to compare?

The 32bit failures in CI we see here are indeed related btw.

mdhaber:
@fbourgey Thanks for all the work on this PR, but based on your (excellent) visualization in https://github.com/scipy/scipy/pull/23622#issuecomment-3618032936, I don't think this asymptotic is as accurate as we'd want it to be in the regime where we need it to kick in. (Relative accuracy of `1e-1` just isn't great when there are other options available.[^1])  gh-24314 seems quite a bit more accurate, so I'll go ahead and close this one.

[^1]: ~~I believe that 2.25/2.29 from https://arxiv.org/pdf/1607.04247 would work, and that approach avoids the need to experimentally determine thresholds for the special case to kick in.~~ No - those asymptotics work in the other tail.

fbourgey:
Of course, thanks a lot, everyone.

PR #23624, opened 2025-09-15
maint: :sparse: name loadnpz test tmpdir in a way to avoid unsafe thread

dschult:
This is a small followup to #23465 which changes a test to avoid failures in freethreading (1) tests.

The test uses the `tmpdir` fixture with a fixed `tmpfile` name. But the function is parametrized so multiple copies of that tmpfile are being used at the same time. It doesn't always raise, but it occasionally leads to failure in the freethreading CI.

This PR makes the name of the file include the parameter values so `tmpfile` differs between tests hopefully eliminating the threading issue.  I'm open to other solutions.


perimosocordiae (approved):

dschult:
Superceeded by #23683

Issue #8338, opened 2018-01-28
dok_matrix no longer allows update


amillb:
In previous versions, scipy allowed an `update` on `sparse.dok_matrix`. This seems to be no longer the case with v1.0.0.

With version 0.19.0, no error is thrown, and the matrix is updated accordingly.

The [docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.dok_matrix.html#scipy.sparse.dok_matrix) imply that the `update` method is still supported. If this isn't a bug, is there a suggested approach to update a `dok_matrix`?

### Reproducing code example:
```
from scipy import sparse
m = sparse.dok_matrix((10, 10))
m.update({(1,1):'a'})
```
### Error message:
```
NotImplementedErrorTraceback (most recent call last)
<ipython-input-15-b3e12160bd9e> in <module>()
      1 from scipy import sparse
      2 m = sparse.dok_matrix((10, 10))
----> 3 m.update({(1,1):'a'})
      4 

/Users/amb/Library/Enthought/Canopy/edm/envs/User/lib/python2.7/site-packages/scipy/sparse/dok.pyc in update(self, val)
    115     def update(self, val):
    116         # Prevent direct usage of update
--> 117         raise NotImplementedError("Direct modification to dok_matrix element "
    118                                   "is not allowed.")
    119 

NotImplementedError: Direct modification to dok_matrix element is not allowed.
```

### Scipy/Numpy/Python version information:
```
('1.0.0', '1.14.0', sys.version_info(major=2, minor=7, micro=13, releaselevel='final', serial=0))
```

WarrenWeckesser:
The change was made in https://github.com/scipy/scipy/pull/7673

@perimosocordiae, can you comment on this?

> [...] is there a suggested approach to update a `dok_matrix`?

You can assign a value to a `dok_matrix`:

```
In [20]: m = sparse.dok_matrix((5, 10))

In [21]: m[1, 2] = -10.0

In [22]: m.A
Out[22]: 
array([[  0.,   0.,   0.,   0.,   0.,   0.,   0.,   0.,   0.,   0.],
       [  0.,   0., -10.,   0.,   0.,   0.,   0.,   0.,   0.,   0.],
       [  0.,   0.,   0.,   0.,   0.,   0.,   0.,   0.,   0.,   0.],
       [  0.,   0.,   0.,   0.,   0.,   0.,   0.,   0.,   0.,   0.],
[...]

PR #23625, opened 2025-09-15
ENH: sparse: Add DOK update method and tests

dschult:
Addresses #8338

DOK sparse format currently does not implement the `update` method (in the spirit of dict).  It used to. Issue 8338 indicates that it is desirable to include an update method with index checking the same as if regular array indexing was used on each entry in the dict.

This implementation allows the incoming data to be a dict, a sparse format dok or an iterable of (key, value) pairs.


scipy/sparse/_dok.py
```diff
@@ -58,8 +58,36 @@ def __init__(self, arg1, shape=None, dtype=None, copy=False, *, maxprint=None):
...
-        raise NotImplementedError("Direct update to DOK sparse format is not allowed.")
+        """Update values from a dict, sparse dok or iterable of 2-tuples like .items()
+
+        Keys of the input must be sequences of nonnegative integers less than the shape
+        for each axis.
+        """
+        if isinstance(val, dict):
+            inputs = val.items()
+        elif issparse(val) and val.format == "dok":
+            inputs = val._dict.items()
```
tylerjereddy: Both the array and matrix version of `dok` are going to get fully captured by the conditional above, right?

```python
In [3]: isinstance(scipy.sparse.dok_array((3, 3)), dict)
Out[3]: True

In [4]: isinstance(scipy.sparse.dok_matrix((3, 3)), dict)
Out[4]: True
```

Looks like this line isn't covered by tests at the moment either, at least when I check locally. Any idea how we'd reach it?

The rest of the diff seems well covered by the new testing.

dschult: Good catch @tylerjereddy!  Thanks!

We already cover the use of `items()` when we make `dok_base` a subclass of `dict`.
So, Yes! This change and behavior will take affect for both sparse arrays and matrix. 
I will take out the unnecessary if clause.

dschult:
The CI results look good. Its ready for re-review. Thanks @tylerjereddy

perimosocordiae (approved):
Looks good, and the tests are comprehensive. We probably want to mention this in the release notes.

tylerjereddy:
Ok, I merged given CJs approval and my minor comment was addressed.

dschult:
I added a bullet to the [release notes wiki](https://github.com/scipy/scipy/wiki/*-Release-Note-Entries-for-SciPy-1.17.0-(asterisk-makes-this-appear-at-the-top)) in the sparse section announcing support for a method `update` that mimics `dict.update` with additional checks that the keys are valid index tuples.

PR #23626, opened 2025-09-15
BUG:sparse: allow setting a singleton via a sparse container

### LINKED ISSUE #23382: BUG: setting with `sparse.xxx_{matrix,array}` 2D singleton
### Describe your issue.

I would like to be able to set on a sparse matrix/array with a singleton-sparse-array the same way I can with a singleton dense array. 

The origin of this issue lies in setting in `dask` where the setter and the set-ee are identically typed.

I think doing something like 
```python
if issparse(x):
    x = x.toarray()
else:
    x = np.asarray(x, dtype=self.dtype)
```
instead of 
https://github.com/scipy/scipy/blob/abb598d7e5634cddb903a0ad7b42f8b57c895719/scipy/sparse/_index.py#L165
is sufficient and indeed, this check already exists in the 1d case:
https://github.com/scipy/scipy/blob/abb598d7e5634cddb903a0ad7b42f8b57c895719/scipy/sparse/_index.py#L131-L134

### Reproducing Code Example

```python
import numpy as np, scipy.sparse as sp

arr = sp.csr_matrix([[2, 3], [4, 5]])
arr[0, 0] = sp.csr_matrix([[0]])
```

### Error message

```shell
Cell In[4], line 1
----> 1 arr[0, 0] = sp.csr_matrix([[0]])

File ~/Projects/Theis/anndata/venv/lib/python3.12/site-packages/scipy/sparse/_index.py:168, in IndexMixin.__setitem__(self, key, x)
    162 row, col = index
    164 if isinstance(row, INT_TYPES) and isinstance(col, INT_TYPES):
--> 165     x = np.asarray(x, dtype=self.dtype)
    166     if x.size != 1:
    167         raise ValueError('Trying to assign a sequence to an item')

TypeError: int() argument must be a string, a bytes-like object or a real number, not 'csr_matrix'
```

```shell
1.16.0 2.3.2 sys.version_info(major=3, minor=12, micro=3, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    found: true
    include directory: unknown
    lib directory: unknown
    name: Accelerate
    pc file directory: unknown
    version: unknown
  lapack:
    found: true
    include directory: unknown
    lib directory: unknown
    name: Accelerate
    pc file directory: unknown
    version: unknown
  pybind11:
    detection method: config-tool
    include directory: unknown
    name: pybind11
    version: 2.13.6
Compilers:
  c:
    commands: cc
    linker: ld64
    name: clang
    version: 15.0.0
  c++:
    commands: c++
    linker: ld64
    name: clang
    version: 15.0.0
  cython:
    commands: cython
    linker: cython
    name: cython
    version: 3.1.2
  fortran:
    commands: gfortran
    linker: ld64
    name: gcc
    version: 13.4.0
  pythran:
    include directory: ../../../../../../private/var/folders/y6/nj790rtn62lfktb1sh__79hc0000gn/T/build-env-4cfkfu1i/lib/python3.12/site-packages/pythran
    version: 0.18.0
Machine Information:
  build:
    cpu: aarch64
    endian: little
    family: aarch64
    system: darwin
  cross-compiled: false
  host:
    cpu: aarch64
    endian: little
    family: aarch64
    system: darwin
Python Information:
  path: /private/var/folders/y6/nj790rtn62lfktb1sh__79hc0000gn/T/build-env-4cfkfu1i/bin/python
  version: '3.12'
```

ilan-gold:
#### Reference issue

Closes #23382

Any recognized sparse object is converted to a dense numpy array, which was the previously handled case for singleton setting (i.e., row and column of `INT_TYPE`).


dschult (commented):
This looks good to me. 
The tests are all for 1D csr-arrays, while the code is for the 2D case.
Could you update [the `test_setelement` method](https://github.com/scipy/scipy/blob/22cf964b6e676e1ce189c18ed927a0fd1f0b792d/scipy/sparse/tests/test_base.py#L2536) in `test_base.py` in a similar to this one? I guess the 1D tests are passing before this PR.

It might also be good to add a test for the nD case at the end of the module `test_coo.py`. Checking a 3D example should be sufficient IIUC.  I think it already works for 3D, but it'd be worth checking.

Thanks!

scipy/sparse/tests/test_base.py
```diff
@@ -2533,42 +2533,44 @@ def check(dtype):
...
-                assert_raises(ValueError, A.__setitem__, (0,0), v)
-
-            if (not np.issubdtype(dtype, np.complexfloating) and
-                    dtype != np.bool_):
-                for v in [3j]:
-                    assert_raises(TypeError, A.__setitem__, (0,0), v)
-
-        for dtype in supported_dtypes:
-            check(np.dtype(dtype))
+    def test_setelement(self, ):
```
dschult:
```suggestion
    def test_setelement(self):
```

dschult (approved):
Looks good to me!
Thanks

scipy/sparse/tests/test_base.py
```diff
@@ -2534,41 +2534,45 @@ def check(dtype):
...
-                for v in [3j]:
-                    assert_raises(TypeError, A.__setitem__, (0,0), v)
-
-        for dtype in supported_dtypes:
-            check(np.dtype(dtype))
+        scalar_containers = [
+            lambda x: csr_array(np.array([[x]])), np.array, lambda x: x
+        ]
+        for scalar_container in scalar_containers:
+            def check(dtype, scalar_container):
```
perimosocordiae: We should move the definition of the `check()` function outside the loop, since we're passing `scalar_container` to it anyway.

perimosocordiae (approved):

scipy/sparse/tests/test_base.py
ilan-gold: done!

dschult:
This looks ready to merge. CI failures seem unrelated. 2 approvals.
Thanks @ilan-gold!

PR #23627, opened 2025-09-16
DOC: spatial: Rotation.random: mention distribution in docstring

PicoCentauri:
[docs only]

#### Reference issue

Closes #23608

Add information about the distribution of the random rotations in 3D.

#### Additional information

One could show that rotations are uniform in the example if we want.

### DIFF
--- a/scipy/spatial/transform/_rotation.py
+++ b/scipy/spatial/transform/_rotation.py
@@ -2279,7 +2279,10 @@ class Rotation:
     def random(
         cls, num: int | None = None, rng: np.random.Generator | None = None
     ) -> Rotation:
-        """Generate uniformly distributed rotations.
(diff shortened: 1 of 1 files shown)


scottshambaugh (commented):
Could you keep the original description and add the specifics about the distribution in a new paragraph below? The Haar measure is a niche topic, and this should still be kept readable for a less familiar audience. I personally like the phrasing "uniform on the sphere" as a lead-in to the more precise mathematical definition.

PicoCentauri:
Sure! I changed the docstring and made it two lines. Let me know what you think.

scottshambaugh (approved):
That looks great, thanks!

lucascolley:
thanks both!

PR #23628, opened 2025-09-16
DOC: Add intructions to build SciPy on WoA

MugundanMCW:
This pull request introduces documentation to guide users for building SciPy from source on Windows on Arm64(WoA).

- Add intructions to build SciPy on WoA:
- This commit updates the source file index.rst to include comprehensive steps for building SciPy on WoA in [documentation](https://docs.scipy.org/doc/scipy/building/index.html)
- The current build setup successfully compiles SciPy on WoA using the following set of compilers:
  - MSVC + Flang-new
  - clang-cl + Flang-new
- The updated documentation includes a reference for linking and compiling OpenBLAS on WoA for SciPy.

### DIFF
--- a/doc/source/building/index.rst
+++ b/doc/source/building/index.rst
@@ -219,6 +219,71 @@ your system.
         try again. The Fortran compiler should be installed as described in
         this section.
 
+  .. tab-item:: Windows on ARM64
+    :sync: windows on ARM64
+
+    A compatible set of C, C++ and Fortran compilers is needed to build SciPy.
+    This is trickier on Windows than on other platforms, because MSVC does not
+    support Fortran, and gfortran and MSVC can't be used together. You will
+    need the following set of compilers:
+
+    1. MSVC + flang (``cl``, ``flang``)
+    2. LLVM + flang (``clang-cl``, ``flang``)
+
+    Compared to macOS and Linux, building SciPy on Windows is a little more
+    difficult, due to the need to set up these compilers. It is not possible to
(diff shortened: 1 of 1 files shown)


rgommers (commented):
Thanks for working on this @MugundanMCW. My main comments right now is that MSVC doesn't actually work (also true for x86-64), because Pythran doesn't support it. We also found other limitations in your earlier PRs with MSVC, so how about just dropping any mention of it and saying that `clang-cl` has to be used?

That will make this a lot shorter. And since some of the diff in this PR is a copy from the Windows tab, I suggest to put Windows on Arm under rather than next to the Windows tab. It can be a `.. note::` with the things specific to Arm64, or inline as comments (e.g., it's a small note to say that you need >=2022 for Arm64 rather than >=2019).

MugundanMCW:
> MSVC doesn't actually work (also true for x86-64), because Pythran doesn't support it. We also found other limitations in your earlier PRs with MSVC, so how about just dropping any mention of it and saying that clang-cl has to be used?

Thanks for the feedback, @rgommers. I understand your point about MSVC lacking support with Pythran, the reason I had included MSVC + Flang is that the existing x64 documentation already contains instructions for setting up MSVC with intel fortran  compiler, so I tried to keep the structure consistent with that.

image

rgommers:
> that the existing x64 documentation already contains instructions for setting up MSVC

Yeah that's my fault I think. It doesn't mention that it won't work because Pythran, but it should. Let's not replicate that here for WoA.

Issue #23611, opened 2025-09-13
CI: benchmark job on CircleCI broken due to asv 0.6.5 release


rgommers:
From [this CI log](https://app.circleci.com/pipelines/github/scipy/scipy/36388/workflows/e7198cbb-95bb-41b5-bf3a-c7b16298188b/jobs/124437):

```
   STDERR -------->
   Traceback (most recent call last):
     File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/asv/benchmark.py", line 99, in <module>
       main()
     File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/asv/benchmark.py", line 91, in main
       commands[mode](args)
     File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/asv_runner/discovery.py", line 310, in _discover
       list_benchmarks(benchmark_dir, fp)
     File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/asv_runner/discovery.py", line 278, in list_benchmarks
       for benchmark in disc_benchmarks(root):
     File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/asv_runner/discovery.py", line 144, in disc_benchmarks
       for module in disc_modules(root_name, ignore_import_errors=ignore_import_errors):
     File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/asv_runner/discovery.py", line 109, in disc_modules
       yield from disc_modules(name, ignore_import_errors)
     File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/site-packages/asv_runner/discovery.py", line 98, in disc_modules
       module = importlib.import_module(module_name)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     File "/home/circleci/.pyenv/versions/3.11.13/lib/python3.11/importlib/__init__.py", line 126, in import_module
       return _bootstrap._gcd_import(name[level:], package, level)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
     File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
     File "<frozen importlib._bootstrap>", line 1147, in _find_and_load_unlocked
[...]

PR #23629, opened 2025-09-16
CI: use asv<0.6.5 to avoid problem with that release - fixes benchmarks job

rgommers:
xref gh-23611

[skip github]


lucascolley (approved):
thanks Ralf! (I think the skip tag is `[skip actions]` 🙃)

PR #23630, opened 2025-09-16
DEV: drop `shallow` from submodules

lucascolley:
@rgommers I know we might not want to merge this, but I'm persistently hitting errors like

```
fatal: transport 'file' not allowed
fatal: Fetched in submodule path 'math', but it did not contain 5e088ffe2ed0e237b9069e3a7352865283d8f196. Direct fetching of that commit failed.
```

Is there a better way around it than this?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,11 +4,9 @@
 [submodule "scipy/sparse/linalg/_propack/PROPACK"]
 	path = scipy/sparse/linalg/_propack/PROPACK
 	url = https://github.com/scipy/PROPACK.git
-	shallow = true
 [submodule "scipy/_lib/unuran"]
 	path = scipy/_lib/unuran
 	url = https://github.com/scipy/unuran.git
-	shallow = true
 [submodule "scipy/_lib/array_api_compat"]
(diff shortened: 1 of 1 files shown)


rgommers:
I was never super happy with `shallow`, and have run into this problem multiple times myself. It's very annoying to resolve even when you know what is happening.

I'd be fine with merging this, and simply accepting that initial checkouts are going to be slower. I don't think it will affect CI jobs much, if at all (but let's verify that).

lucascolley:
From a quick look I do not see evidence of increased CI times

ilayn:
We can just get `--depth 1` if that's what we are after. It also implies shallow anyways

lucascolley:
> We can just get `--depth 1` if that's what we are after. It also implies shallow anyways

just to check I understand, you're just pointing out that devs can override this with `--depth 1` if they need a faster clone?

If so, sounds like we can merge this. I merged main in case anyone wants to take a look at fresh CI timings.

rgommers (approved):
This can always be reverted in case an unexpected issue shows up, but this seems like the right thing to do - let's give it a go. Thanks @lucascolley

PR #23631, opened 2025-09-16
DOC: ndimage: fix example of 'wrap' mode for interpolation functions

mguaypaq:
[docs only]

The current docstring description of `mode='wrap'` for the functions in `scipy/ndimage/_interpolation.py` is wrong, as illustrated by the following snippet:

```py3
>>> import numpy as np, scipy.ndimage
>>> scipy.ndimage.affine_transform(
...     np.array([0, 1, 2, 3]),  # length 4, same as the original example
...     np.eye(1),  # no rotation or scaling
...     offset=-4,  # show 4 entries to the left
...     output_shape=(12,),  # also show 4 entries to the right
...     mode='wrap')
array([2, 3, 1, 2, 0, 1, 2, 3, 1, 2, 0, 1])
```

The text itself is fine; the problem is just with the little example on the array `a b c d`. It might be clearer to show more repetitions:

```
d b c d b c d b c | a b c d | b c a b c a b c a
```

Or a single repetition of length 3, like with `mode='mirror'`:

```
d b c | a b c d | b c a
```

But I chose to just correct the typo with the smallest possible change.

(It would also be simpler to see what's going on if the _left_ repetitions used `a` instead of `d` and the _right_ repetitions used `d` instead of `a`, but that would require a code change, or leaning hard on the warning that "it is not well defined which sample will be chosen at the point of overlap":
```
a b c a b c a b c | a b c d | b c d b c d b c d
```
)

### DIFF
--- a/scipy/ndimage/_ni_docstrings.py
+++ b/scipy/ndimage/_ni_docstrings.py
@@ -109,7 +109,7 @@ _mode_interp_constant_doc = (
     'grid-wrap' (`a b c d | a b c d | a b c d`)
         The input is extended by wrapping around to the opposite edge.
 
-    'wrap' (`d b c d | a b c d | b c a b`)
+    'wrap' (`c d b c | a b c d | b c a b`)
         The input is extended by wrapping around to the opposite edge, but in a
         way such that the last point and initial point exactly overlap. In this
         case it is not well defined which sample will be chosen at the point of
--- a/scipy/ndimage/_ni_docstrings.py
+++ b/scipy/ndimage/_ni_docstrings.py
@@ -109,11 +109,17 @@ _mode_interp_constant_doc = (
     'grid-wrap' (`a b c d | a b c d | a b c d`)
         The input is extended by wrapping around to the opposite edge.
 
-    'wrap' (`c d b c | a b c d | b c a b`)
+    'wrap' (`b c d b c | a b c d | b c a b c`)
         The input is extended by wrapping around to the opposite edge, but in a
         way such that the last point and initial point exactly overlap. In this
         case it is not well defined which sample will be chosen at the point of
-        overlap.""")
+        overlap.
+
+
+        For the example illustrated above, the edge values ``a`` and ``d`` overlap,
+        while the inner pattern ``b c`` repeats. The example demonstrates one
+        possible output, while ``b c a b c | a b c d | b c d b c`` is another,
+        since either ``a`` or ``d`` may occur at each point of overlap.""")
 _mode_interp_mirror_doc = (
     _mode_interp_constant_doc.replace("Default is 'constant'",
                                       "Default is 'mirror'")


mguaypaq:
Note: this affects the docstrings of the following functions, all in `scipy/ndimage/_interpolation.py`:
- `spline_filter1d`
- `spline_filter`
- `geometric_transform`
- `map_coordinates`
- `affine_transform`
- `shift`
- `zoom`
- `rotate`

lucascolley (commented):
thanks @mguaypaq . I took the time to understand what was going on. To use your examples, I think we effectively have this, where the letters under the slashes are the implicit overlaps:

```
d b/a c/d | a b c d | b/a c/d a
```

```
d b/a c d b/a c d b/a c/d | a b c d | b/a c/d a b c/d a b c/d a
```

```
2 3 1/0 2/3 | 0 1 2 3 | 1/0 2/3 0 1
```

And the code-change example:

```
a/d b c a/d b c a/d b c/d | a b c d | b/a c d/a b c d/a b c d/a
```

Is that correct? To be honest I don't think this is clear from the docstring even after this change, but I'm not sure how to best describe the behaviour. Any further ideas for what we could say?

mguaypaq:
@lucascolley I don't think that's quite correct.

One way to think about this is that `mode='wrap'` assumes that `d == a`, so that what gets repeated in a cycle is just `a b c`, instead of the full `a b c d`. So, using slashes for the overlaps, this would look like:

```
(d/a) b c (d/a) b c (d/a) b c | (d/a) b c (d/a) | b c (d/a) b c (d/a) b c (d/a)
```

But, since it may be that `d != a` for callers who don't feel like following the assumptions, it matters at each cycle whether the original `a` or the original `d` is used in the output array. The current implementation, for whatever reason, chooses to pick `d` for things on the left, then `a` and `d` for the original array, then `a` for everything on the right. So, this looks like:

```
(d/-) b c (d/-) b c (d/-) b c | (-/a) b c (d/-) | b c (-/a) b c (-/a) b c (-/a)
```

I think this pattern is extremely hard to understand intuitively, but it _is_ what the code does, so the documentation should reflect it.

mguaypaq:
Sorry, I don't really have a better suggestion for the text than what I already wrote, but thank you for your time already!

mguaypaq:
Sorry for being impatient, but, on a documentation correctness scale of:

1. Confusing and doesn't match the code's behaviour (clearly a bug).
2. Confusing and matches the code's behaviour (honestly a pretty normal state of affairs).
3. Clear and matches the code's behaviour (who knows when).

Accepting this PR should be a low-risk and mechanical way to go from 1 to 2, and wouldn't make it any harder to eventually move to 3. I don't see why it should be relegated all the way to release 1.18.0.

lucascolley:
Hey @mguaypaq, 1.18.0 is the next release in which we will include documentation improvements and enhancements. There may be some bugfix releases in the 1.17.x series, but these are typically reserved for more critical changes to reduce burden on the release manager.

Sorry for the wait, this is still on my list!

lucascolley (approved):
I pushed an update based on my understanding. @mguaypaq could you check if that agrees with your understanding?

One final question if so: is it only the interpolation functions that had this problem?

mguaypaq:
Thanks, your extended example and added explanation match my understanding. I know that `affine_transform` is affected, but I will check for the other functions and get back to you shortly.

mguaypaq:
From a quick look at the code, it looks like all the functions in https://github.com/scipy/scipy/pull/23631#issuecomment-3300150488 eventually pass the `mode` argument to `spline_filter1d`, which passes it on to the underlying C implementation. So, I think all of these functions legitimately have/had this documentation problem.

lucascolley:
but not the non-interpolation functions such as `correlate1d`, which use a different entry here like 'mode_reflect' or 'mode_multiple'? https://github.com/scipy/scipy/blob/c3e55561d8ce323da67a82ad2702e032fa07945c/scipy/ndimage/_ni_docstrings.py#L196-L212

mguaypaq:
Ah, I see: the changes in this PR correctly fix all docstrings which use `%(mode_interp_constant)` or `%(mode_interp_mirror)`, but the question is about the functions which use `%(mode_reflect)` or `%(mode_multiple)`.

TL;DR: I think only the interpolation functions are affected.

Most of the functions with `%(mode_reflect)` or `%(mode_multiple)` boil down to `correlate1d`, and that function does seem to treat `mode='wrap'` as a synonym for `mode='grid-wrap'`, as evidenced by this code snippet:

```py3
>>> from scipy.ndimage import correlate1d
>>> array = [0, 1, 2, 3]
>>> left_shift_weights = [1, 0, 0, 0, 0, 0, 0, 0, 0]
>>> # check that this is a left shift by 4 positions
>>> correlate1d(array, left_shift_weights, mode='reflect')
array([3, 2, 1, 0])
>>> # now compare the modes 'wrap' and 'grid-wrap'
>>> correlate1d(array, left_shift_weights, mode='wrap')
array([0, 1, 2, 3])
>>> correlate1d(array, left_shift_weights, mode='grid-wrap')
array([0, 1, 2, 3])
```

There is also `generic_filter1d` and `generic_filter`. I haven't checked `generic_filter1d`, but for `generic_filter` it seems to correctly treat `wrap` and `grid-wrap` as synonyms:

```py3
>>> from scipy.ndimage import generic_filter
>>> generic_filter([0, 1, 2, 3], (lambda x: x[0]), size=9, mode='reflect')
array([3, 2, 1, 0])
>>> generic_filter([0, 1, 2, 3], (lambda x: x[0]), size=9, mode='wrap')
array([0, 1, 2, 3])
>>> generic_filter([0, 1, 2, 3], (lambda x: x[0]), size=9, mode='grid-wrap')
array([0, 1, 2, 3])
```

The other functions with `%(mode_reflect)` or `%(mode_multiple)` seem to boil down to either `generic_laplace` or `generic_gradient_magnitude`. These ones pass the `mode` down to a user-defined callable. In theory, the callable could do something different with the `mode`, but then that's the user's fault.

lucascolley:
alright, thanks for the investigation! Merging.

lucascolley:
while the next release may be a while, our changes should be visible in the devdocs very soon https://scipy.github.io/devdocs/

Issue #23568, opened 2025-09-06
MAINT: spatial.transform.Rotation: `from_euler/from_davenport` shape consistency


amacati:
As discussed in #23467, the current implementation of `from_euler` and `from_davenport` are not consistent with broadcasting rules. This issue points out the inconsistencies and proposes an alternative behavior. 

Note that the new rules are a **breaking change**.

### Euler Angle Cases and Proposed Behavior

| Case      | Sequence       | Input Shape      | Output Shape     | Behavior    |
|------------|----------------|------------------|------------------|-------------|
| 0D        | `"x"`          | `()`             | `(4,)`           | Unchanged   |
| 0D        | `"xyz"`        | `()`             | `raise`       | Unchanged   |
| 1D        | `"x"`          | `(1,)`           | `(1, 4)` → `(4,)`| **Changed** |
| 1D        | `"xyz"`        | `(3,)`           | `(4,)`           | Unchanged   |
| 2D        | `"x"`          | `(N, 1)`         | `(N, 4)`         | Unchanged   |
| 2D        | `"xyz"`        | `(N, 3)`         | `(N, 4)`         | Unchanged   |

> **Note**: After the 2D case, standard broadcasting rules are followed already.

The logic here is that we promote 0D arrays, `float`s and `int`s to 1D. `angles.shape[-1]` must then be the same as `num_axes`, i.e. the last dimension of angles describes by how much we turn around each axis. Consequently, 0D arrays remain invalid for sequences with more than one axis. The resulting rotation has the shape `np.atleast_1d(angles).shape[:-1]`.

### Davenport Cases and Proposed Behavior

| Case      | Axes Shape  | Angles Shape   | Output Shape      | Behavior             |
|-------------|-------------|----------------|-------------------|----------------------|
| 0D        | `(3,)`      | `()`           | `(4,)`            | Unchanged            |
| 1D         | `(3,)`      | `(1,)`         | `(1, 4)` → `(4,)` | **Changed**          |
| 1D        | `(3,)`      | `(N,)`         | `(N, 4)` → `raise`     | **Changed**, raise if `N != 1` |
| 1D        | `(2, 3)`    | `(2,)`         | `(4,)`            | Unchanged            |
| 2D        | `(3,)`      | `(N, 1)`       | `(N, 4)`          | Unchanged            |
| 2D        | `(2, 3)`    | `(2, 1)`       | `(2, 4)`          | **Changed**, raise `2 != 1` |
[...]

PR #23632, opened 2025-09-16
MAINT: spatial.transform.Rotation: `from_euler/from_davenport` shape consistency

amacati:
This PR removes shape inconsistencies in `from_euler` and `from_davenport`.

#### Reference issue
- Adresses #23568 
- Working towards #22500 
- Makes #23467 fully consistent
- Tracker in #23391

Changes shape results as discussed in #23568 

#### Additional information
I took the liberty and renamed `select_backend` and reformatted the file after recent merges that did not use the ruff formatting.


amacati:
I'm not sure why the benchmarks fail. The output sais `ModuleNotFoundError: No module named 'scipy'`, which sounds like a problem unrelated to this PR. Could be wrong though. Are you more familiar with the CI @lucascolley ?

lucascolley:
that should fix it

amacati:
Yep, that did the trick. Should be good for review. 

Side note: I think this is the first mention of ND rotations in the docs since they were not updated in #23467 , so it might make sense to take a closer look and check if the shape descriptions etc are consistent with what is being used elsewhere.

scipy/spatial/transform/_rotation_cy.pyx
```diff
@@ -474,8 +432,17 @@ def from_euler(seq, angles, bint degrees=False):
...
+    
+    angles = np.asarray(angles, dtype=float)
+    if degrees:
+        angles = np.deg2rad(angles)
 
-    angles, is_single = _format_angles(angles, degrees, num_axes)
+    is_single = angles.ndim < 2
+    angles = np.atleast_2d(angles)
+    
+    if angles.shape[1] != num_axes:
```
scottshambaugh, line 447: Should this not be `shape[-1]`? `np.atleast_2d` allows for higher dimensions, and it doesn't look like this error check is being exercised for higher ndim in the tests.

scipy/spatial/transform/_rotation.py
```diff
@@ -731,7 +719,7 @@ def from_euler(seq: str, angles: ArrayLike, degrees: bool = False) -> Rotation:
 
         Initialize a stack with a single rotation around a single axis:
 
-        >>> r = R.from_euler('x', [90], degrees=True)
+        >>> r = R.from_euler('x', [[90]], degrees=True)
```
scottshambaugh, line 722: May be worth keeping as a single rotation here with quat shape (4,)

scipy/spatial/transform/_rotation_cy.pyx
amacati: The function only takes one or two-dimensional arrays by design in `Rotation`. I originally changed it to `shape[-1]` (and also replaced `np.atleast_2d` with `reshape((-1, 1))` for 1D arrays), but the function currently has no bounds checks, and thus crashes. We would have to reactivate those if we want to make use of `shape[-1]`.

amacati: Do you think it's better to deactivate the bounds checks? Any input > 2D should be considered an error anyways.

scipy/spatial/transform/_rotation.py
amacati: Then the text would be wrong, would it not? A stack of rotations implies (1, 4), not (4,).

scottshambaugh: Ah, I didn't see the existing example for single rotations cut off in the diff above this, good as-is.

scipy/spatial/transform/_rotation_cy.pyx
scottshambaugh: Ah, I didn't fully grok the implications of the `cython_compatible=quat.ndim < 3` conditionals. Let's keep the bounds check.

scottshambaugh (approved):

scipy/spatial/transform/_rotation_cy.pyx
amacati: Should we enforce this with a dimension sanity check? I don't think that would cost any meaningful performance, but it might make it easier to understand what's going on and make the function safer to use in case of future refactorings.

scottshambaugh: That sounds like a prudent defensive measure, yes.

amacati:
Test failures seem to indicate something major is broken with building scipy, but that should be unrelated to this PR(?)

lucascolley:
merging main should hopefully solve that

amacati:
Yep, that fixed it

lucascolley (approved):
thanks both!

Usually this is optional from the contributor's side, but can I ask that you write a release note for this now @amacati ? I'll then copy that into https://github.com/scipy/scipy/wiki/*-Release-Note-Entries-for-SciPy-1.17.0-(asterisk-makes-this-appear-at-the-top).

amacati:
`scipy.spatial.transform` breaking changes: 
The resulting shapes for `Rotation.from_euler` / `from_davenport` have changed to make them consistent with broadcasting rules.

Angle inputs to Euler angles must now strictly match the number of provided axes in the last dimension. The resulting Rotation has the shape `np.atleast_1d(angles).shape[:-1]`.

Angle inputs to Davenport angles must also match the number of axes in the last dimension. The resulting Rotation has the shape `np.broadcast_shapes(np.atleast_2d(axes).shape[:-2], np.atleast_1d(angles).shape[:-1])`. 

Does that work for you Lucas?

lucascolley:
yep thanks!

PR #23633, opened 2025-09-17
MAINT/TST: stats: loosen a test tolerance.

WarrenWeckesser:
The test 'TestStudentT.test_t_inf_df' has been failing occasionally on the "Linux test / free-threaded(1)" CI job, but only by a small amount.


mdhaber (approved):
LGTM. The failures I see have an error `4e-15`, so `5e-15` should do the trick.

dschmitz89:
Thanks @WarrenWeckesser .

PR #23635, opened 2025-09-18
BUG,DOC: `find_peaks_cwt`: reject troughs, improve SNR of actual peaks

antipassive:
* When finding signal level for SNR calculation, use maximum CWT coefficient on the ridge line, not CWT coefficient at the end of the ridge line
* Allow ridge lines with negative signal to be rejected due to low SNR
* Update docs of `min_snr` parameter in `_filter_ridge_lines` and `find_peaks_cwt` accordingly
* Add unit tests for both fixes

#### Reference issue
Closes #23604
Closes #23684

This implements the definition of "signal" in signal-to-noise ratio as described in [1],
which is the reference listed in `_peak_finding.py` itself.
As a result, difference in SNR between true positives and false positives increases.
Higher values of the `min_snr` parameter can then be used without loss of true
positives, leading to better precision for `find_peaks_cwt`.
Furthermore, a subset of false positives (appearing at local minima between peaks)
is eliminated, because ridge lines with negative signal can no more meet a positive `min_snr` requirement.

#### Additional information
* This change may lead to _worse_ precision for some users before they raise the value of `min_snr` in the call to `find_peaks_cwt`. This is because the SNR of some false positives will also increase a little, possibly exceeding the value of `min_snr` currently in use.
* That the SNR calculation in [1] and SciPy ever differed seems to be an unintended omission: issuing
`git log --follow scipy/signal/_peak_finding.py`
and checking out the earliest listed commits reveals that the odd SNR calculation is already there in commit `3399829`, which dates back to 2011.

EDIT: add mention of unit tests and the fix to #23684

#### References
[1] Bioinformatics (2006) 22 (17): 2059-2065.
doi:`10.1093/bioinformatics/btl355`

### DIFF
--- a/scipy/signal/_peak_finding.py
+++ b/scipy/signal/_peak_finding.py
@@ -1180,7 +1180,7 @@ def _filter_ridge_lines(cwt, ridge_lines, window_size=None, min_length=None,
     hf_window, odd = divmod(window_size, 2)
 
     # Filter based on SNR
-    row_one = np.abs(cwt[0, :])
+    row_one = cwt[0, :]
     noises = np.empty_like(row_one)
     for ind, val in enumerate(row_one):
         window_start = max(ind - hf_window, 0)
@@ -1191,7 +1191,7 @@ def _filter_ridge_lines(cwt, ridge_lines, window_size=None, min_length=None,
     def filt_func(line):
         if len(line[0]) < min_length:
             return False
-        snr = max(cwt[line[0], line[1]]) / noises[line[1][0]]
+        snr = max(cwt[line[0], line[1]]) / abs(noises[line[1][0]])
         if snr < min_snr:
             return False
         return True
--- a/scipy/signal/_peak_finding.py
+++ b/scipy/signal/_peak_finding.py
@@ -1154,10 +1154,11 @@ def _filter_ridge_lines(cwt, ridge_lines, window_size=None, min_length=None,
         Minimum length a ridge line needs to be acceptable.
         Default is ``cwt.shape[0] / 4``, ie 1/4-th the number of widths.
     min_snr : float, optional
-        Minimum SNR ratio. Default 1. The signal is the value of
-        the cwt matrix at the shortest length scale (``cwt[0, loc]``), the
-        noise is the `noise_perc`\\ th percentile of datapoints contained within a
-        window of `window_size` around ``cwt[0, loc]``.
+        Minimum SNR ratio. Default 1. The signal is the maximum value
+        of the ``cwt`` matrix found along the ridge line. The noise is
+        the `noise_perc`\\ th percentile of datapoints contained within
+        a window of `window_size` around ``cwt[0, loc]``, where ``loc``
+        is the last column index of the ridge line.
     noise_perc : float, optional
         When calculating the noise floor, percentile of data points
         examined below which to consider noise. Calculated using
@@ -1190,7 +1191,7 @@ def _filter_ridge_lines(cwt, ridge_lines, window_size=None, min_length=None,
     def filt_func(line):
         if len(line[0]) < min_length:
             return False
-        snr = abs(cwt[line[0][0], line[1][0]] / noises[line[1][0]])
+        snr = abs(max(cwt[line[0], line[1]]) / noises[line[1][0]])
         if snr < min_snr:
             return False
         return True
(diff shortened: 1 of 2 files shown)


antipassive:
While running experiments on the "fixed" version of `find_peaks_cwt`, I found another SNR-related problem that I can't afford to leave as it is. I'll open a new issue shortly.

EDIT: The problem has been handled. I opened issue #23684. A fix is included in this PR.

scipy/signal/_peak_finding.py
```diff
@@ -1190,7 +1191,7 @@ def _filter_ridge_lines(cwt, ridge_lines, window_size=None, min_length=None,
     def filt_func(line):
         if len(line[0]) < min_length:
             return False
-        snr = abs(cwt[line[0][0], line[1][0]] / noises[line[1][0]])
+        snr = max(cwt[line[0], line[1]]) / noises[line[1][0]]
```
tylerjereddy: I drafted a regression test and slightly different diff for gh-23684, which I've pasted below the fold. On my branch the test, based on your example in the ticket, fails before and passes after applying the patch. Maybe you could consider adding a similar regression test here since it looks like you're taking care of that one as well?

<details>

```diff
--- a/scipy/signal/_peak_finding.py
+++ b/scipy/signal/_peak_finding.py
@@ -1190,7 +1190,7 @@ def _filter_ridge_lines(cwt, ridge_lines, window_size=None, min_length=None,
     def filt_func(line):
         if len(line[0]) < min_length:
             return False
-        snr = abs(cwt[line[0][0], line[1][0]] / noises[line[1][0]])
+        snr = cwt[line[0][0], line[1][0]] / abs(noises[line[1][0]])
         if snr < min_snr:
             return False
         return True
diff --git a/scipy/signal/tests/test_peak_finding.py b/scipy/signal/tests/test_peak_finding.py
index 8118c505e8..d2b1fed643 100644
--- a/scipy/signal/tests/test_peak_finding.py
+++ b/scipy/signal/tests/test_peak_finding.py
@@ -1,5 +1,4 @@
 import copy
-
 import numpy as np
 import pytest
 from pytest import raises, warns
@@ -911,3 +910,21 @@ class TestFindPeaksCwt:
         found_locs = find_peaks_cwt(test_data, widths)
 
         np.testing.assert_equal(found_locs, 32)
+
+    def test_gh_23684(self):
+        rng = np.random.default_rng(50)
+        x = np.arange(500)
+
+        def lorentz(x, height, width, loc):
+            return height / (1 + np.pow((x-loc)/width, 2))
+
+        # synthesize noisy signal with peaks
+        noise = rng.standard_normal((500,)) * 1e-3
+        signal = (lorentz(x, 0.3, 10, 100) +
+                  lorentz(x, 0.4, 15, 250) +
+                  lorentz(x, 0.2, 5, 400) + noise)
+
+        # find the 3 peaks
+        cwt_scales = np.geomspace(1, 20, 16)
+        peaks = find_peaks_cwt(signal, cwt_scales, min_snr=4, noise_perc=90)
+        xp_assert_equal(peaks, [100, 250, 400])
```

</details>

tylerjereddy: I'm assuming your patch is more sensible than mine, since I'm not a domain expert, so it is mostly the regression test that is worth considering. I do wonder how well this code is tested though, if my poorly-understood patch also works...

antipassive: Yes, I can add a regression test for the signal level.

antipassive: ...now that I think of it again, I understand that you probably meant adding a regression test for trough exclusion (gh-23684) in this branch that already has a draft PR. Yes, I can add one. I'm sorry for the confusion!

But do we need a regression test for the signal level (gh-23604)?

antipassive:
Thanks to @tylerjereddy for finding the right way to fix #23684, as my own patch might have broken existing code!

scipy/signal/_peak_finding.py
antipassive: A regression test now exists for both trough exclusion and signal level.

antipassive:
I'm sorry to convert this PR into a draft, but I should probably figure out why `run_benchmarks` failed on CircleCI. What exactly is being tested in `run_benchmarks`? Local unit tests worked correctly.

antipassive:
It turns out my `refguide_check` fails due to an issue that was fixed in #23685. I'll have to rebase and clean up a bit.

antipassive:
I cleaned up this work and pushed a new branch. It has the same fixes but no misleading work-in-progress commits. Also, it passes all CI checks, which probably makes it easier to review.

EDIT: Please find the new PR below

PR #23638, opened 2025-09-18
CI: adapt `gpu-ci.yml` to new Pixi workspace

lucascolley:
Towards gh-21489. See gh-23567 for context.

@rgommers probably easiest to review each commit individually.


.github/workflows/pixi.toml
```diff
@@ -1,149 +1,271 @@
...
+mkl = ">=2024.2.2,<2025"
 
+[feature.mkl.tasks.build-mkl]
+cmd = "spin build --build-dir=build-mkl --setup-args=-Dblas=mkl-dynamic-ilp64-seq --setup-args=-Duse-g77-abi=true"
+cwd = "scipy"
+env = { CC = "ccache $CC", CXX = "ccache $CXX", FC = "ccache $FC" }
+description = "Build with MKL BLAS"
+
+
+# [feature.accelerate]
```
lucascolley, line 98: lots of this is to be uncommented in follow-ups

.github/workflows/pixi.toml
```diff
@@ -1,149 +1,271 @@
...
+# cwd = "../.."
+# depends-on = [{ task = "build", environment = "build" }]
 
-[feature.torch-cuda.tasks]
-test-torch-cuda = { cmd = "spin test -b torch", cwd = "../..", env = { SCIPY_DEVICE = "cuda" } }
+[feature.torch-cuda.tasks.test-torch-cuda]
+cmd = "spin test --no-build -b torch -m 'array_api_backends and not slow'"
+cwd = "../.."
+env.SCIPY_DEVICE = "cuda"
+depends-on = [{ task = "build", environment = "build" }]
```
lucascolley, line 174: maybe some of these should be `build-mkl` instead, but I'm not sure it makes a difference right now. I think this is at least correct once we have https://github.com/rgommers/pixi-dev-scipystack/issues/41 fixed.

.github/workflows/pixi.toml
rgommers: That seems fine to me - let's make an exception to the regular "don't do that please" because I'm sure you'll follow up on this very soon.

.github/workflows/pixi.toml
```diff
@@ -1,149 +1,270 @@
...
+tach = ">=0.29.0,<0.30"
+cython-lint = ">=0.16.7,<0.17"
+packaging = ">=25.0,<26"
+
+[feature.lint.tasks.lint]
+cmd = "spin lint"
+cwd = "../.."
+description = "Run main lint checks"
+
+[feature.lint.tasks.tach]
```
rgommers: Any idea why this doesn't work when run locally?

```
$ pixi r tach
✨ Pixi task (tach in lint): tach check --exact: (Check module interdependencies with tach)                                  Configuration
⚠  Warning: No first-party imports were found. You may need to use 'tach mod' to update your Python source roots. Docs: https://docs.gauge.sh/usage/configuration#source-roots
Unused Dependencies
❌ 'scipy' does not depend on: ['scipy._lib']
❌ 'scipy._lib' does not depend on: ['scipy']

...
```

.github/workflows/pixi.toml
rgommers: Seems fine to cross that bridge when we get to it.

rgommers (approved):
Thanks Lucas, this generally LGTM. One small question about `tach`, but if you can't reproduce let's ignore for now since it works in Ci.

The main issue I had in running things locally is CUDA 13 for the CuPy job. That may be necessary in CI because of gh-23458, it's just a little unfortunate because CUDA 13 drops support for a lot of older GPUs. Fine for this file, but if we'd use it for regular development we'd want to stick with CUDA 12.8/12.9 and use an older CuPy version instead.

Feel free to merge unless the `tach` thing needs a tweak.

.github/workflows/pixi.toml
lucascolley: it seems like a regression in `tach` that must have happened some time ago. Let's just remove it since it is now unmaintained.

rgommers: That sounds fine to me

lucascolley:
Thanks Ralf, will merge if green. I'll include the array_api and lint workflows in the next PR, then onto the macos workflow

Issue #23642, opened 2025-09-18
CI: meson setup fails with "invalid int value: 'Sentinel.UNSET'"


nickodell:
It seems that jobs have started crashing with this error across multiple CI workflows:

```
$ meson compile -j Sentinel.UNSET -C build
usage: meson compile [-h] [--clean] [-C WD] [-j JOBS] [-l LOAD_AVERAGE] [-v]
                     [--ninja-args NINJA_ARGS] [--vs-args VS_ARGS]
                     [--xcode-args XCODE_ARGS]
                     [TARGET ...]
meson compile: error: argument -j/--jobs: invalid int value: 'Sentinel.UNSET'
```

* [Ubuntu musl latest, fast, py3.11/npAny, spin](https://github.com/scipy/scipy/actions/runs/17836963684/job/50716445482#logs)
* [Mac OS Accelerate, full, py3.13/npAny, spin (ILP64)](https://github.com/scipy/scipy/actions/runs/17841939074/job/50733300193#logs)
* [Windows fast, py3.12/npAny, spin](https://github.com/scipy/scipy/actions/runs/17841939072/job/50733300658#logs)

Possibly a bug in spin, or the function `cpu_count()` within `.spin/cmds.py`?

WarrenWeckesser:
Version 8.3.0 of [`click`](https://pypi.org/project/click/) was released about 6 hours ago.

I was able to reproduce the problem locally, and fix it by installing version 8.2.1 of `click`, so apparently the problem is in the new version of `click`.

I'll submit a PR to pin `click` at version 8.2.1.

WarrenWeckesser:
My attempt at a quick fix in https://github.com/scipy/scipy/pull/23643 didn't work.  Too many of the CI jobs have hard-coded `pip install` commands with packages listed explicitly in the command instead of using the files in `requirements/`.

WarrenWeckesser:
The `click` issue is https://github.com/pallets/click/issues/3065.

The immediate fix would be to pin `click` to version `<8.3`, but for SciPy, `click` is a transitive dependency coming from `spin`.  So `spin` needs an update to either pin its dependency or adapt to click 8.3.0, and then we need to use the updated `spin`.

*Edit:* Unless the issue with click 8.3.0 turns out to be a bug. In that case, if a bug fix is imminent, maybe we can just wait.

rgommers:
Wow, that's the largest set of failures we've had in a long time. Yay for transitive dependencies:(

[...]

PR #23643, opened 2025-09-19
CI: Pin click to 8.2.1.

WarrenWeckesser:
See gh-23642.


WarrenWeckesser:
Nope, this won't work.  Too many CI jobs have hard-coded `pip install <packages...>` instead of using the files in `requirements/`.

Issue #21491, opened 2024-08-31
CI/DEV: better utilise `ccache`


lucascolley:
_Quoting @nickodell from https://github.com/scipy/scipy/issues/21489#issuecomment-2322923995:_

I found two improvements that could be made:

- The cache load action searches for a cache key that must match down to the second. The only way that can work is if two jobs begin in the same second, one finishes building and saves its cache, and then the other loads the cache.

  The upshot of this is that we are not benefitting from ccache at all, even in jobs that use it.

- The cache load action will only load caches from main branch or from previous builds on the same PR. Since there are no ccache enabled builds happening on main branch, a PR which has one commit can't benefit from ccache. If main were regularly rebuild (once per week?) and the cache saved, this cost could be reduced.

cc @andyfaff

nickodell:
Thanks for making this issue.

I took another look at this part:

>The cache load action searches for a cache key that must match down to the second. The only way that can work is if two jobs begin in the same second, one finishes building and saves its cache, and then the other loads the cache.

I don't recall exactly what I was looking at when I found this, but it seems like some ccache is actually helping. For example here it gets a 78% hit rate. So that aspect is working.

https://github.com/scipy/scipy/actions/runs/10647360287/job/29515232289?pr=21483#step:10:15

andyfaff:
ccache was very much more important before the transition to the meson build system, builds would take forever. It's importance is slightly less than it was, but it still has an effect.

With the way that scipy development occurs most of the CI runs are triggered by commits to PRs. Successive pushes to a given PR will utilise the caching, in the entries where it's implemented. Those are mainly in macos.yml and linux.yml.

For example, if one looks at https://github.com/scipy/scipy/actions/runs/10658208038 (linux.yml), then in the entries it's used (e.g. https://github.com/scipy/scipy/actions/runs/10658208038/job/29538932819), the cache is hit ~70% of the time.  Nick saw a 78%. It'll vary from job to job and PR to PR. Where it's used it has a good effect.

Where things could be improved:

- more jobs using it in linux.yml/macos.yml
- More concise way of setting up, better way of setting up cache key.

I'm not sure if it's worth aggregating caches over multiple jobs because the jobs typically have different build conditions (e.g. one uses ATLAS, the other OpenBLAS).

AFAICT the cache for the same job between different PRs is shared.

nickodell:
>AFAICT the cache for the same job between different PRs is shared.

[...]

PR #23649, opened 2025-09-19
CI: use Pixi for `array_api.yml` and `lint.yml`

lucascolley:
- **CI: create ccache action**
- **STY: format `array_api.yml`**
- **CI: use Pixi for `array_api.yml`**
- **CI: use Pixi for `lint.yml`**

towards gh-21489, see gh-23567 for context. Review commits separately for nicer diffs!

### DIFF
--- a/.github/ccache/action.yml
+++ b/.github/ccache/action.yml
@@ -1,0 +1,24 @@
+name: Setup Compiler Cache
+description: Prepare and restore compiler cache using ccache
+inputs:
+  workflow_name:
+    description: "Unique name of the calling workflow to use as cache key prefix"
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+    - name: Prepare compiler cache
+      id: prep-ccache
+      shell: bash
+      run: |
+        mkdir -p "${CCACHE_DIR:-$HOME/.ccache}"
+        echo "dir=${CCACHE_DIR:-$HOME/.ccache}" >> $GITHUB_OUTPUT
+        NOW=$(date -u +"%F-%T")
+        echo "timestamp=${NOW}" >> $GITHUB_OUTPUT
+    - name: Setup compiler cache
+      uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684
+      with:
+        path: ${{ steps.prep-ccache.outputs.dir }}
+        key: ${{ inputs.workflow_name }}-ccache-linux-${{ steps.prep-ccache.outputs.timestamp }}
+        restore-keys: ${{ inputs.workflow_name }}-ccache-linux-
--- a/.github/workflows/array_api.yml
+++ b/.github/workflows/array_api.yml
@@ -41,10 +41,7 @@ jobs:
         with:
           pixi-version: v0.55.0
           manifest-path: .github/workflows/pixi.toml
-          cache-write: ${{ github.event_name == 'push' && github.ref_name == 'main' }}
-          environments: >-
-            build
-            array-api-cpu
+          cache: false
 
       - name: Set up ccache
         uses: ./.github/ccache
--- a/.github/workflows/array_api.yml
+++ b/.github/workflows/array_api.yml
@@ -15,8 +15,6 @@ permissions:
 env:
   CCACHE_DIR: "${{ github.workspace }}/.ccache"
   INSTALLDIR: "build-install"
-  # Only run tests that use the `xp` fixture
-  XP_TESTS: -m array_api_backends
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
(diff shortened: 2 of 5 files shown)


.github/workflows/array_api.yml
```diff
@@ -34,76 +32,31 @@ jobs:
...
-        spin test -b torch $XP_TESTS -- --durations 3 --timeout=60
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          submodules: recursive
+
+      - uses: prefix-dev/setup-pixi@ba3bb36eb2066252b2363392b7739741bb777659 # v0.8.1
+        with:
+          pixi-version: v0.55.0
+          manifest-path: .github/workflows/pixi.toml
+          cache-write: ${{ github.event_name == 'push' && github.ref_name == 'main' }}
```
rgommers: I thought I commented on this elsewhere, but: these jobs don't run on `main`.

rgommers:
I think cache usage here is fine, but let's keep an eye on it because GitHub tightening the limits soon (xref https://github.com/scipy/scipy/issues/21491#issuecomment-3314757541).

.github/workflows/array_api.yml
```diff
@@ -34,76 +32,31 @@ jobs:
...
+
+      - name: Set up ccache
+        uses: ./.github/ccache
         with:
-          path: ${{ steps.prep-ccache.outputs.dir }}
-          key: ${{ github.workflow }}-${{ matrix.python-version }}-ccache-linux-${{ steps.prep-ccache.outputs.timestamp }}
-          restore-keys: |
-            ${{ github.workflow }}-${{ matrix.python-version }}-ccache-linux-
-
-      - name: Setup build and install scipy
```
rgommers: I'd prefer for building to stay a separate step - it makes it much easier to find important data for the build and test steps (e.g., the discovery output `pytest` prints out). And it's shorter logs per step, making navigation in the GitHub UI faster.

.github/workflows/lint.yml
```diff
@@ -22,31 +22,26 @@ jobs:
...
+      - uses: prefix-dev/setup-pixi@ba3bb36eb2066252b2363392b7739741bb777659 # v0.8.1
         with:
-          python-version: ${{ matrix.python-version }}
-
-      - name: Install Python packages
-        run: |
-          python -m pip install ruff cython-lint packaging tach spin
+          pixi-version: v0.55.0
+          manifest-path: .github/workflows/pixi.toml
+          cache-write: ${{ github.event_name == 'push' && github.ref_name == 'main' }}
```
rgommers, line 35: This env is very fast to install without caching, so I think it's better to remove this.

rgommers (commented):
Thanks Lucas. Overall looks good in principle. The array API job changes seem to cause a lot of failures, that will need fixing.

lucascolley:
@mdhaber do you know which versions of marray and array-api-strict result in these failures (if we've seen them before)?

ilayn:
Is there also a way to make these jobs less dense? 

image

This is way too heavy for our test suite though we are essentially testing for others. Not to start the discussion but this is basically something CuPy or Jax can do against us instead of us testing the entire API

lucascolley:
We could try to do something like https://github.com/scikit-image/scikit-image/pull/7749 but that is not easy to get right.

rgommers:
> This is way too heavy for our test suite though we are essentially testing for others. Not to start the discussion but this is basically something CuPy or Jax can do against us instead of us testing the entire API

This is indeed a bit off-topic for this PR, but just to point out: we are definitely not testing _their_ API, we are testing our own SciPy API with particular input types. We have custom code handling those types. Everything is being tested once on CPU and (if applicable) once on GPU, for a single version of each optional dependency, so runtime can't really be reduced easily.

I think the good point is: the total runtime of a single job starts to be too much as coverage grows. We'll need to split that up. We could also move some of the CPU testing to the GPU server, since it has spare capacity at the moment and is way faster than GHA runners.

Let's do this type of stuff separately I'd say. The order of priorities is:
1. Use lock files to stop the continuous new failures of CI jobs because of dependencies changing.
2. Reduce cache usage (we're on a timeline there)
3. Reduce the runtime of some of the jobs.

ilayn:
Yes but I don't think we need to test every function for every input for an input mechanism is pretty much the same. So this is pretty much testing our input checks for every array api function candidate. 

I am not convinced that we have to keep track of this input validation. That is to say, 45 minutes of "is it really accepting jax array properly?" testing is way too much for us as an API consumer.  If the API is doing what it is supposed to be doing, it should not be tested here. I think this does not need to be split up but cleaned up for a few sample cases. 

Since AWS and Azure and other hyperscalers prefer traffic over infra costs for free tiers I think we can also remove caching entirely since like you noticed pip is fast. If this fortran stuff goes away then we won't need all the compiler downloads anyways.

rgommers:
> If this fortran stuff goes away then we won't need all the compiler downloads anyways.

Yep, we can clean up a bunch of things at that point for sure.

> Yes but I don't think we need to test every function for every input for an input mechanism is pretty much the same

I think you're missing some context here. There are both code paths in our code base and test skips in our test suite that are array library-specific. And not just a few, a fair amount. We really do need a test suite run per library. Please take my word for that (and if not, I'm sure everyone who has worked on this topic will agree if you do want to ask them). 

If you're interested in total test suite run time improvements, there are much better candidates. E.g., we're templating over Python versions too often, doing essentially the same thing multiple times with just different Python versions, which is fairly pointless since we test every Python version many times anyway.

lucascolley:
I suppose there is the option of putting some stuff on cron/label-activated rather than every PR. Not really sure how to assess the tradeoff.

ilayn:
> I think you're missing some context here. There are both code paths in our code base and test skips in our test suite that are array library-specific. And not just a few, a fair amount. 

I do believe that. But also I'd like to ask your understanding that I am aware of this effort. What I am trying to emphasize is that if the happy flow works for NumPy array then it should work for other arrays if the Array API is doing its job and the authors are sticking to the subset the standard is standarding which many people are trying to stick to. 

Testing everything once again for these libraries are pretty much the anti pattern here and I am not sure where I am losing you when this discussion is even briefly touched. It should not be full of special cases if the Array API is keeping its promise. That is my generic point. 

But anyways, I promised there will be no discussion let's not get into it now :)  I'll concede my point.


For the caching point, folks at microsoft have asked me if we have considered switching to a full dockerized(or any container for that matter) solution such that 

- we create a container images initially and once
- pull the same container for the job
- get into the container  build scipy inside it
- test scipy
- then kill the container. 
 
Then the caching would make sense since we would pull the same image with the same contents. If needed then the image can be rebuilt for newer compilers etc. But I don't know if the extra effort is worth it to maintain.

lucascolley:
perhaps https://tech.quantco.com/blog/pixi-production is worth a read

lucascolley:
I bumped marray to fix the failures (I hadn't locked since Matt merged https://github.com/conda-forge/marray-python-feedstock/pull/7), turned off caching for `array_api.yml`, and split the build out into a separate step. I think that is all blocking comments addressed.

rgommers:
> folks at microsoft have asked me if we have considered switching to a full dockerized(or any container for that matter) solution 

Maybe, at some point. It's definitely more clunky to deal with Docker, and storing those images is also not easy - remember we tried with Docker Hub for some workflows and they pulled the plug on their free tier so we had to give up on that.

The reason to do go down this path would be build reproducibility. If we want full bitwise reproducibility, we need to control compilers, runner images, etc. Pixi gets us pretty much there, but that doesn't help for wheel build workflows - which is where we'd care most about full reproducibility.

andyfaff:
There are already standard images for linux wheel builds from PyPA, https://cibuildwheel.pypa.io/en/stable/options/#linux-image. One can use those straight away in GHA.

andyfaff:
It's also possible to make macos images using tart and packer, they're used by cirrus. I've never made an image with those tools though, only used them.

lucascolley:
all green 🌳

.github/ccache/action.yml
```diff
@@ -0,0 +1,24 @@
+name: Setup Compiler Cache
+description: Prepare and restore compiler cache using ccache
+inputs:
+  workflow_name:
+    description: "Unique name of the calling workflow to use as cache key prefix"
```
ilayn, line 5: Is this our "unique name of the calling workflow to use as cache key prefix"?

lucascolley: sorry, I don't understand the question

ilayn: I thought we need to replace this with something that is meaningful to us or is this something we are skipping?

lucascolley: the `description` field is just describing the `workflow_name` input parameter. We then pass in `${{ github.workflow }}` at L49 in `array_api.yml`

ilayn: Ok so thats a filler then. Thanks.

ilayn:
Thanks @lucascolley

lucascolley:
I just noticed that I forgot to remove the Pixi cache from the lint job, will fix that in the `macos.yml` PR. For the larger envs maybe we want to cron on main and cache that

PR #23653, opened 2025-09-20
MAINT: Fix `spatial.transform.RigidTransform` xp backend comments

amacati:
Fix two comments left over from the cython port of the RigidTransform xp backend merged in #23425. 

#### Reference issue
- #22500
- #23391
- #23425


lucascolley (approved):
thanks Martin

lucascolley:
just for the future, would have been nice to include `[skip ci]` in the commit message here (I forget all the time)

amacati:
Ah right, forgot about that. Will do, sry

PR #23654, opened 2025-09-20
CI: use Pixi for `macos.yml`

lucascolley:
towards https://github.com/scipy/scipy/issues/21489, see https://github.com/scipy/scipy/pull/23567 for context. Review commits separately for nicer diffs!

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -32,110 +32,34 @@ jobs:
       needs.get_commit_message.outputs.message == 1
       && (github.repository == 'scipy/scipy' || github.repository == '')
     runs-on: macos-latest
-    strategy:
-      matrix:
-        python-version: ["3.11"]
-
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           submodules: recursive
 
-      - name: Install Ccache
-        run: |
-          brew install ccache
-
-      - name: Prepare compiler cache
-        id: prep-ccache
-        shell: bash -l {0}
-        run: |
-          mkdir -p "${CCACHE_DIR}"
-          echo "dir=$CCACHE_DIR" >> $GITHUB_OUTPUT
-          NOW=$(date -u +"%F-%T")
-          echo "timestamp=${NOW}" >> $GITHUB_OUTPUT
-
-      - name: Setup compiler cache
-        uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
-        id: cache-ccache
-        # Reference: https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key
-        # NOTE: The caching strategy is modeled in a way that it will always have
-        # a unique cache key for each workflow run (even if the same workflow is
-        # run multiple times). The restore keys are not unique and for a partial
-        # match, they will return the most recently created cache entry,
-        # according to the GitHub Action Docs.
+      - uses: prefix-dev/setup-pixi@ba3bb36eb2066252b2363392b7739741bb777659 # v0.8.1
         with:
-          path: ${{ steps.prep-ccache.outputs.dir }}
-          key: ${{ github.workflow }}-${{ matrix.python-version }}-ccache-macos-${{ steps.prep-ccache.outputs.timestamp }}
-          # This evaluates to `macOS Tests-3.11-ccache-macos-` which is not
-          # unique. As the CI matrix is expanded, this will need to be updated to
-          # be unique so that the cache is not restored from a different job altogether.
-          restore-keys: |
-            ${{ github.workflow }}-${{ matrix.python-version }}-ccache-macos-
+          pixi-version: v0.55.0
+          manifest-path: .github/workflows/pixi.toml
+          cache: false
 
-      - name: Setup Conda
-        uses: conda-incubator/setup-miniconda@505e6394dae86d6a5c7fbb6e3fb8938e3e863830 # v3.1.1
+      - name: Set up ccache
+        uses: ./.github/ccache
         with:
-          python-version: ${{ matrix.python-version }}
-          channels: conda-forge
-          channel-priority: true
-          activate-environment: scipy-dev
-          use-only-tar-bz2: false
-          miniforge-variant: Miniforge3
-          miniforge-version: latest
-          use-mamba: true
-
-      - name: Get Date
-        id: get-date
-        run: echo "today=$(/bin/date -u '+%Y%m%d')" >> $GITHUB_OUTPUT
-        shell: bash
-
-      - name: Cache conda
-        uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
-        env:
-          # Increase this value to reset cache if environment.yml has not changed
-          CACHE_NUMBER: 1
-        with:
-          path: ${{ env.CONDA }}/envs/scipy-dev
-          key: ${{ runner.os }}--${{ steps.get-date.outputs.today }}-conda-${{ env.CACHE_NUMBER }}-${{ hashFiles('environment.yml') }}
-        id: envcache
-
-      - name: Update Conda Environment
-        run: mamba env update -n scipy-dev -f environment.yml
-        if: steps.envcache.outputs.cache-hit != 'true'
-
-      - name: Build and Install SciPy
-        shell: bash -l {0}
-        run: |
-          conda activate scipy-dev
-          mamba install python=${{ matrix.python-version}}
-
-          # optional test dependencies
-          # scikit-umfpack is pinned because 0.4.2 packages seem broken, see gh-23564
-          mamba install "scikit-umfpack=0.3.3" scikit-sparse
-
-          # configure system boost
-          mamba install libboost-headers=1.89.0
-          export BOOST_INCLUDEDIR=${{ env.CONDA }}/envs/scipy-dev/include/boost
-          export BOOST_LIBRARYDIR=${{ env.CONDA }}/envs/scipy-dev/lib
-          rm -rf subprojects/boost_math  # so will fail if system boost doesn't work
-
-          # configure system qhull
-          mamba install qhull=2020.2
-          rm -rf subprojects/qhull_r  # so will fail if system qhull doesn't work
-
-          CC="ccache $CC" spin build --use-system-libraries
+          workflow_name: ${{ github.workflow }}
 
       - name: Test SciPy
         shell: bash -l {0}
+        working-directory: .github/workflows
         run: |
-          conda activate scipy-dev
+          rm -rf ../../subprojects/boost_math  # so will fail if system boost doesn't work
+          rm -rf ../../subprojects/qhull_r  # so will fail if system qhull doesn't work
           export OMP_NUM_THREADS=2
-          spin test -j2
+          pixi run test-system-libs -j2
 
       - name: Ccache statistics
         shell: bash -l {0}
-        run: |
-          ccache -s
+        run: pixi exec ccache -s
 
   test_scipy_openblas:
     name: M1 & OpenBLAS, fast, py3.11/npAny, spin
@@ -152,31 +76,16 @@ jobs:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           submodules: recursive
-      - name: Setup Python
-        uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # v5.5.0
+
+      - uses: prefix-dev/setup-pixi@ba3bb36eb2066252b2363392b7739741bb777659 # v0.8.1
         with:
-          python-version: ${{ matrix.python-version }}
-          cache: "pip"
+          pixi-version: v0.55.0
+          manifest-path: .github/workflows/pixi.toml
+          cache: false
 
-      - name: Build and Install SciPy
-        run: |
-          sudo xcode-select -s /Applications/Xcode_15.2.app
-
-          git submodule update --init
-          GFORTRAN_LOC=$(which gfortran-13)
-          ln -s $GFORTRAN_LOC gfortran
-          export PATH=$PWD:$PATH
-
-          # Ensure we have gfortran dylib
-          GFORTRAN_LIB=$(dirname `gfortran --print-file-name libgfortran.dylib`)
-          export DYLD_LIBRARY_PATH=$GFORTRAN_LIB
-
-          pip install meson cython pythran pybind11 ninja numpy spin "click<8.3.0"
-          pip install -r requirements/openblas.txt
-          spin build --with-scipy-openblas
-
-          pip install pooch pytest hypothesis
-          spin test
+      - name: Test with scipy-openblas32
+        working-directory: .github/workflows
+        run: pixi run test-scipy-openblas
 
   test_accelerate:
     name: Accelerate, full, py3.13/npAny, spin
@@ -186,45 +95,19 @@ jobs:
       && (github.repository == 'scipy/scipy' || github.repository == '')
     runs-on: macos-15
     strategy:
-      fail-fast: false
       matrix:
-        blas-interface: ["LP64", "ILP64"]
+        environment: [accelerate-lp64, accelerate-ilp64]
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           submodules: recursive
 
-      - name: Setup Python
-        uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # v5.5.0
+      - uses: prefix-dev/setup-pixi@ba3bb36eb2066252b2363392b7739741bb777659 # v0.8.1
         with:
-          python-version: "3.13"
+          pixi-version: v0.55.0
+          manifest-path: .github/workflows/pixi.toml
+          cache: false
 
-      - name: Make gfortran-13 on runner image usable
-        run: |
-          # Ensure we use gfortran-13 and that its runtime is on the library search path
-          FC=$(which gfortran-13)
-          GFORTRAN_LIB=$(dirname `$FC --print-file-name libgfortran.dylib`)
-          echo DYLD_LIBRARY_PATH=$GFORTRAN_LIB >> "$GITHUB_ENV"
-          echo FC=$FC >> "$GITHUB_ENV"
-
-      - name: Install dependencies from PyPI
-        run: |
-          pip install meson cython pythran pybind11 ninja numpy spin pooch pytest hypothesis "click<8.3.0"
-
-      - name: Build SciPy with Accelerate (LP64)
-        if: matrix.blas-interface == 'LP64'
-        run: |
-          # The b_ndebug flag is a workaround for `_LIBCPP_ENABLE_ASSERTIONS` warnings, can
-          # be removed once Meson 1.8.3 is out with the fix (see meson#14440).
-          spin build --with-accelerate -S-Db_ndebug=true
-
-      - name: Build SciPy with Accelerate (ILP64)
-        if: matrix.blas-interface == 'ILP64'
-        run: |
-          # Once Meson has full BLAS support (xref meson#14773), the
-          # `blas-symbol-suffix` flag can be dropped.
-          spin build -S-Dblas=accelerate -S-Duse-ilp64=true -S-Dblas-symbol-suffix='$NEWLAPACK$ILP64' -S-Db_ndebug=true
-
-      - name: Run the test suite
-        run: |
-          spin test -m full
+      - name: Test SciPy
+        working-directory: .github/workflows
+        run: pixi run test-${{ matrix.environment }} -m full
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -10,7 +10,7 @@ on:
       - maintenance/**
 
 permissions:
-   contents: read  # to fetch code (actions/checkout)
+  contents: read # to fetch code (actions/checkout)
 
 env:
   INSTALLDIR: "build-install"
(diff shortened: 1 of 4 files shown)


.github/workflows/array_api.yml
```diff
@@ -50,7 +50,7 @@ jobs:
 
       - name: Build SciPy
         working-directory: .github/workflows
-        run: pixi run build
+        run: pixi run --environment=build build
```
lucascolley: we could instead remove the need for this change by splitting the `build` feature into `build-deps` and `build-tasks`, but I think we can punt on that until we start thinking about moving the manifest to the root of the repo

.github/workflows/macos.yml
```diff
@@ -32,112 +32,34 @@ jobs:
...
-
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          submodules: recursive
+
+      - uses: prefix-dev/setup-pixi@ba3bb36eb2066252b2363392b7739741bb777659 # v0.8.1
+        with:
+          pixi-version: v0.55.0
+          manifest-path: .github/workflows/pixi.toml
+          cache: false
```
lucascolley, line 44: might want to think about turning this caching on via cron jobs on main, but maybe we can leave that for now

.github/workflows/pixi.toml
```diff
@@ -251,6 +270,84 @@ depends-on = [{ task = "build", environment = "build" }]
...
+description = "Test SciPy with system libs"
+
+
+### Alternative python versions ###
+
+[feature.py311-test]
+platforms = ["linux-64", "osx-arm64", "win-64"]
+
+[feature.py311-test.dependencies]
+# Build and test dependencies with Python 3.11
```
lucascolley: we could instead widen the constraints in the main `build` and `test` features and include them in the `py311` env. Not sure on the trade-offs there, might want to assess that when we come to consider lock file updates.

.github/workflows/array_api.yml
rgommers: This doesn't look good indeed. Can you explain what the problem is with having a `default = ["build", "test"]` environment, like in https://github.com/rgommers/pixi-dev-scipystack/?

.github/workflows/macos.yml
rgommers: I'd say leave it for now. I've done some work today on reducing cache usage, I'd like to get that in first and then see where we are before adding more caching.

.github/workflows/pixi.toml
rgommers: This is more than just python 3.11 though - I see also `numpy<2`. Is the idea for this to be a `deps-minimum-versions` kind of thing?

lucascolley: `numpy<2` was probably a hack I put in to (quickly) make it solvable, that should be changed to follow the pinning used for the other deps

.github/workflows/array_api.yml
lucascolley: the problem is that any environment which needs to use `no-default-feature` to avoid any of the `build` or `test` deps would then need to duplicate all of the other deps. That actually seems like only a small problem for the envs we currently define, as most of them are including `build` and `test` anyway, so I'm not against changing it here if you prefer.

I just wasn't sure it would scale well. Typically, as Pixi workspaces get more complex, it has been less of a headache to keep a slim default feature (in my experience). I'm [not totally alone in that](https://github.com/prefix-dev/pixi/issues/4591#issuecomment-3294078064), at least.

.github/workflows/pixi.toml
lucascolley: oh wait, this was it:

```console
E:\sandbox\nplt2
❯ pixi init --platform=osx-arm64
✔ Created E:\sandbox\nplt2\pixi.toml

E:\sandbox\nplt2 via 🧚 v0.55.0
❯ pixi add scikit-umfpack=0.3.3 numpy>2
Error:   × failed to solve requirements of environment 'default' for platform 'osx-arm64'
  ├─▶   × failed to solve the environment
  │
  ╰─▶ Cannot solve the request because of: scikit-umfpack 0.3.3.* cannot be installed because there are no viable options:
      └─ scikit-umfpack 0.3.3 | 0.3.3 | 0.3.3 | 0.3.3 | 0.3.3 | 0.3.3 | 0.3.3 | 0.3.3 | 0.3.3 would require
         └─ numpy >=1.21.6,<2.0a0, for which no candidates were found.
```

we could loosen the upper pin if we wanted and it would be forced transitively, but this is what `pixi upgrade -f py311-test numpy` gives us.

lucascolley: I haven't put thought into whether it makes sense to use a non-default [pinning strategy](https://pixi.sh/latest/reference/pixi_configuration/#pinning-strategy), and I imagine the ideal choice depends on how we end up doing lock file updates, but maybe one of those options could make this situation more intuitive.

andyfaff:
One concern I have with the lock files is that their size is large, will they lead to bloat in the size of the git repo? If they're updated regularly and the diff is always large, then I could imagine that being an issue. Would putting them in a submodule be the solution there?

lucascolley:
Probably better to discuss in gh-23637, but I like the sound of the idea of re-locking infrequently in the main repo, but automating some tests with re-locking frequently in a separate repo, to smoke out problems ahead of time.

I wonder what this would look like with `no-pin` [pinning-strategy](https://prefix-dev.github.io/pixi/v0.55.0/reference/pixi_configuration/) 🤔 perhaps that would eliminate the need for `pixi upgrade`.

.github/workflows/array_api.yml
rgommers: I think it's useful to refactor the dependencies a bit, so they're more modular and reusable, and that the top-level `[dependencies]` is empty. That allows adding them to other envs easily without having to redefine them.

That's different from tasks though. I think the principle of having a task only defined in a single env is possible to adhere to, and that ends up way better for usability.

Something like:
```ini
[dependencies]

[feature.build.dependencies]
# Only build dependencies - we expect these to be the same across all envs
# One needs to add the `build` feature as well as one `py_xxx` feature for functioning builds
compilers = ">=1.11.0,<2"
pkg-config = ">=0.29.2,<0.30"
ninja = ">=1.13.1,<2"
blas-devel = ">=3.9.0,<4"
cython = ">=3.1.4,<4"
spin = ">=0.14,<0.15"
python-build = "*"
...

[feature.py_default.dependencies]
python = ">=3.13.7,<3.14"
numpy = ">=2.3.3,<3"

[feature.py_default.tasks.build]
cmd = "spin build --setup-args=-Dblas=blas --setup-args=-Dlapack=lapack --setup-args=-Duse-g77-abi=true"
...

[feature.py_311.dependencies]
python = ">=3.11.7,<3.12"
numpy = "==1.26.4"

[feature.py_default.tasks.build-py311]
cmd = "spin build --setup-args=-Dblas=blas --setup-args=-Dlapack=lapack --setup-args=-Duse-g77-abi=true"
...
[feature.py_default.tasks.test-py311]
cmd = "spin test"
...


[feature.test.dependencies]
pytest = ">=8.4.2,<9"
...

[environments]
default = ["build", "test", "py_default"]
py311 = { features = ["build", "test", "py_311"], no-default-feature = true }
```

I am not sure if this will actually work out better for defining CI jobs when things grow large. I am fairly sure that if we'd move it two levels up, it's better for contributors because `pixi r build` and `pixi r test` "just work". Having that not work is very clunky, to the point you'd go back to defining shell aliases.



> we could instead remove the need for this change by splitting the `build` feature into `build-deps` and `build-tasks`, but I think we can punt on that until we start thinking about moving the manifest to the root of the repo

That's another option indeed. I will have very limited time over the next couple of weeks, so I can't really give my idea more thought or try it out. If you think your approach is easier overall, I'm fine going that way.

.github/workflows/pixi.toml
rgommers: Okay, then hack is fine until we resolve the `scikit-umfpack` 0.4.x compat issue I think - add a code comment to remind ourselves of this reason?

rgommers:
> I wonder what this would look like with `no-pin` [pinning-strategy](https://prefix-dev.github.io/pixi/v0.55.0/reference/pixi_configuration/) 🤔 perhaps that would eliminate the need for `pixi upgrade`.

I think that will make the problem worse rather than better. The key point is _not_ to run `pixi update` when you make a targeted change. Diffs can certainly be small and contained to a single package version, which is often all you need as part of a PR when working on a bug or CI.

lucascolley:
> > I wonder what this would look like with `no-pin` [pinning-strategy](https://prefix-dev.github.io/pixi/v0.55.0/reference/pixi_configuration/) 🤔 perhaps that would eliminate the need for `pixi upgrade`.
> 
> 
> 
> I think that will make the problem worse rather than better. The key point is _not_ to run `pixi update` when you make a targeted change. Diffs can certainly be small and contained to a single package version, which is often all you need as part of a PR when working on a bug or CI.

I think we are on the same page. The idea would be to maintain any constraints manually (one could perhaps use targeted `pixi upgrade` calls as a helper, but probably easier to just do it by hand). All unnecessary constraints would be absent, and the only time `pixi update` is called is for the infrequent lock file update.

.github/workflows/array_api.yml
lucascolley: thanks Ralf, that is very helpful.

If you can spare one more thought on this, can I check that this would fit with what you have in mind (combining now with the `no-pin` suggestion):

```toml
[dependencies]
# We always need (some version of) `spin`
spin = "*" # transitively includes `python`

[feature.build-task.tasks.build] # in place of your `py_default.tasks.build`
...

[feature.test-task.tasks.test]
...

[feature.py_311.dependencies]
python = "3.11.*"

[feature.py_311.tasks.build-py311]
...

[feature.py_311.tasks.test-py311]
...

[environments]
build = { features = ["build-deps", "build-task"], solve-group = "default" } # the sole env for `pixi run build`
test = { features = ["build-deps", "test-deps", "test-task"], solve-group = "default" } # the sole env for `pixi run test`
py311 = ["build-deps", "test-deps", "py_311"] # includes build deps, test deps, and forces py311; tasks come from `py_311` feature
```

With this, I don't think we would need either of:
- a `default` environment
- use of `no-default-feature` (since everything needs `spin`, and we don't define any default tasks (or any that we wouldn't be happy with having in every env))

lucascolley: I wrote this up properly in gh-23658

rgommers: I think that should work well indeed. Let me have a look at gh-23658

lucascolley:
commits squashed, ready for review!

lucascolley:
@amacati any idea what's up with the CI failures here?

amacati:
That's a weird one. The hard-coded strings were always a brittle choice. Can you somehow show where `__repr__` differs from the expected? Seems like they are identical. What exactly does this PR change? Are np print settings somehow affected?

lucascolley:
> That's a weird one. The hard-coded strings were always a brittle choice. Can you somehow show where `__repr__` differs from the expected? Seems like they are identical. What exactly does this PR change? Are np print settings somehow affected?

I'll look into it. Annoyingly, upgrading macOS seems to have broken my C compiler 😅 so I'm struggling to build SciPy locally. https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/.60missing.20LC_LOAD_DYLIB.60.20from.20.60clang.60.20on.20macOS.2026/with/541444282

lucascolley:
this was my editor being over-eager touching whitespace in the previous PR.

lucascolley:
The accelerate jobs took excessively long, not sure why right now

lucascolley:
looks like a hang in `test_mpmath`

```
Thu, 25 Sep 2025 13:53:37 GMT
scipy/special/tests/test_mpmath.py ..................................... [ 81%]
Thu, 25 Sep 2025 13:53:58 GMT
.................x.x..............................................x.xs.. [ 81%]
Thu, 25 Sep 2025 15:24:54 GMT
x...xsx...x...s...s.xs..xx.x..x........................x                 [ 82%]
```

lucascolley:
`pixi run test-accelerate-lp64 -t scipy.special.tests.test_mpmath::TestSystematic::test_jacobi_int -m full` reliably reproduces the hang for me.

@h-vetinari is this a usual suspect?

lucascolley:
the following diff fixes the hang:

```diff
diff --git a/scipy/special/tests/test_mpmath.py b/scipy/special/tests/test_mpmath.py
index e0f729aca9..6586306b88 100644
--- a/scipy/special/tests/test_mpmath.py
+++ b/scipy/special/tests/test_mpmath.py
@@ -1573,7 +1573,7 @@ class TestSystematic:
             lambda n, a, b, x: sc.eval_jacobi(int(n), a, b, x),
             lambda n, a, b, x: exception_to_nan(jacobi)(n, a, b, x, **HYPERKW),
             [IntArg(), Arg(), Arg(), Arg()],
-            n=20000,
+            n=4095,
             dps=50,
         )
```

but changing to `n=4096` brings back the hang (which can't be a coincidence)

lucascolley:
@steppi any idea why `eval_jacobi` may be hanging for `n>=2**12`? I guess the next step is to figure out whether the hang is coming from `mpmath`, `xsf_binom`, or `hyp2f1`?

Unfortunately the `mpmath` testing infrastructure is a bit of a maze for me to try to understand quickly.

lucascolley:
I can reproduce the failure in the default test environment, this is not specific to accelerate. It only showed up now because this is the job where we use `-m full`.

lucascolley:
backtrace:

```
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGSTOP
  * frame #0: 0x0000000101ac2630 _ufuncs.cpython-313-darwin.so`xsf::cephes::detail::hyt2f1(a=10, b=<unavailable>, c=<unavailable>, x=<unavailable>, loss=0x000000016fdfcbc0) at hyp2f1.h:289:62 [opt]
    frame #1: 0x0000000101ac2e94 _ufuncs.cpython-313-darwin.so`xsf::cephes::hyp2f1(a=<unavailable>, b=<unavailable>, c=<unavailable>, x=<unavailable>) at hyp2f1.h:572:13 [opt]
    frame #2: 0x0000000101ae81b0 _ufuncs.cpython-313-darwin.so`xsf::hyp2f1(a=<unavailable>, b=<unavailable>, c=<unavailable>, x=<unavailable>) at hyp2f1.h:699:87 [opt] [inlined] [artificial]
    frame #3: 0x0000000101ab3c64 _ufuncs.cpython-313-darwin.so`__pyx_fuse_0__pyx_f_5scipy_7special_15orthogonal_eval_hyp2f1(__pyx_v_a=<unavailable>, __pyx_v_b=<unavailable>, __pyx_v_c=<unavailable>, __pyx_v_z=<unavailable>) at _ufuncs.c:16978:13 [opt] [inlined]
    frame #4: 0x0000000101ab3c60 _ufuncs.cpython-313-darwin.so`__pyx_fuse_0__pyx_f_5scipy_7special_15orthogonal_eval_eval_jacobi(__pyx_v_n=NaN, __pyx_v_alpha=10, __pyx_v_beta=0, __pyx_v_x=709.08956571282397) at _ufuncs.c:17168:26 [opt] [inlined]
    frame #5: 0x0000000101ab3c28 _ufuncs.cpython-313-darwin.so`__pyx_f_5scipy_7special_15orthogonal_eval_eval_jacobi_l(__pyx_v_n=<unavailable>, __pyx_v_alpha=<unavailable>, __pyx_v_beta=0, __pyx_v_x=709.08956571282397) at _ufuncs.c:17297:15 [opt]
    frame #6: 0x0000000101ab77a0 _ufuncs.cpython-313-darwin.so`__pyx_f_5scipy_7special_7_ufuncs_loop_d_pddd__As_pddd_d(__pyx_v_args=<unavailable>, __pyx_v_dims=<unavailable>, __pyx_v_steps=0x000000016fdfd050, __pyx_v_data=<unavailable>) at _ufuncs.c:29045:19 [opt]
```

lucascolley:
> `warning: _ufuncs.cpython-313-darwin.so was compiled with optimization - stepping may behave oddly; variables may not be available.`

`--setup-args=-Dbuildtype=debug` doesn't seem to help here 🤔

lucascolley:
@ngoldbaum hello! Without setting `env = { CFLAGS = "-O0 -g" }`, the `pyx` variables here were `unavailable` (see the previous traceback). Is that expected / is that also necessary in NumPy? I thought `--setup-args=-Dbuildtype=debug` would be sufficient.

P.S. thanks for https://numpy.org/devdocs/dev/development_advanced_debugging.html#c-debuggers, giving myself a crash course now!

---

After setting that, here is the less obfuscated traceback:

```
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGSTOP
  * frame #0: 0x0000000101b49a68 _ufuncs.cpython-313-darwin.so`xsf::cephes::detail::hyt2f1(a=10, b=<unavailable>, c=<unavailable>, x=<unavailable>, loss=0x000000016fdfca90) at hyp2f1.h:288:37 [opt]
    frame #1: 0x0000000101b4a2e8 _ufuncs.cpython-313-darwin.so`xsf::cephes::hyp2f1(a=<unavailable>, b=<unavailable>, c=<unavailable>, x=<unavailable>) at hyp2f1.h:572:13 [opt]
    frame #2: 0x0000000101b44698 _ufuncs.cpython-313-darwin.so`__pyx_fuse_0__pyx_f_5scipy_7special_15orthogonal_eval_hyp2f1(__pyx_v_a=10, __pyx_v_b=-8.9884656743110496E+307, __pyx_v_c=-5.666666666666667, __pyx_v_z=4.4942328371555248E+307) at _ufuncs.c:16978:13
    frame #3: 0x0000000101b2b07c _ufuncs.cpython-313-darwin.so`__pyx_fuse_0__pyx_f_5scipy_7special_15orthogonal_eval_eval_jacobi(__pyx_v_n=-10, __pyx_v_alpha=-6.666666666666667, __pyx_v_beta=-8.9884656743110496E+307, __pyx_v_x=-8.9884656743110496E+307) at _ufuncs.c:17168:26
    frame #4: 0x0000000101b2b0f0 _ufuncs.cpython-313-darwin.so`__pyx_f_5scipy_7special_15orthogonal_eval_eval_jacobi_l(__pyx_v_n=-10, __pyx_v_alpha=-6.666666666666667, __pyx_v_beta=-8.9884656743110496E+307, __pyx_v_x=-8.9884656743110496E+307) at _ufuncs.c:17297:15
    frame #5: 0x0000000101b32ec4 _ufuncs.cpython-313-darwin.so`__pyx_f_5scipy_7special_7_ufuncs_loop_d_pddd__As_pddd_d(__pyx_v_args=0x000000016fdfce50, __pyx_v_dims=0x000000016fdfcd48, __pyx_v_steps=0x000000016fdfd050, __pyx_v_data=0x0000000101bd0810) at _ufuncs.c:29045:19
```

lucascolley:
with `n=4095`, many similar calls like `xsf::cephes::detail::hyt2f1(a=-12.333333333333334, b=-12.333333333333334, c=-2.3333333333333335, x=0.66666666666666663, loss=0x000000016fdfc950)` with small finite arguments happen effectively instantly.

with `n=4096`, the first such call is `xsf::cephes::detail::hyt2f1(a=10, b=-8.9884656743110496E+307, c=-5.666666666666667, x=4.4942328371555248E+307, loss=0x000000016fdfca90)`, and it takes about 10 seconds. Further calls continue to take about 10 seconds and also have a huge negative `b` argument.

---

Python reproducer of the slowness:

```python
scipy.special.hyp2f1(10, -1e20, -5.5, 5)
```

question is whether this is unexpected or whether the test suite shouldn't be trying arguments this large

ngoldbaum:
> Is that expected / is that also necessary in NumPy?

I've never tried to do this with Cython code. Sounds vaguely like meson failing to pass these arguments for Cython build targets with the debug build profile set in the `meson setup` config for the build.

ngoldbaum:
That is, sounds like a meson bug.

lucascolley:
Explanation for the power of 2 cutoff: `n=4096` translates to a jump to `n==6` from `n==5` at https://github.com/scipy/scipy/blob/f466f35b7d0ed33e6fd697cceccef175e256f074/scipy/special/_mptestutils.py#L39

For negative inf `a` and positive inf `b` (the defaults for `Arg`), and `n==5`, the "logspace between 10 and b"

https://github.com/scipy/scipy/blob/f466f35b7d0ed33e6fd697cceccef175e256f074/scipy/special/_mptestutils.py#L75-L77

is just

```
(Pdb) logpts2
array([10.])
```

but for `n==6`, it is

```
(Pdb) logpts2
array([1.00000000e+001, 8.98846567e+307])
```

lucascolley:
and 🤦‍♂️ finally I understand why the slowness only just appeared. The workflow on main does not install `mpmath`, so all of `test_mpmath` is skipped. The linux job that uses `-m full` (the only other such job) also doesn't install `mpmath`, so this hanging test just hasn't been exercised at all by our CI until now, IIUC.

lucascolley:
> > Is that expected / is that also necessary in NumPy?
> 
> I've never tried to do this with Cython code. Sounds vaguely like meson failing to pass these arguments for Cython build targets with the debug build profile set in the `meson setup` config for the build.

thanks! I'll see if I can file a reproducer in an issue

.github/workflows/macos.yml
```diff
@@ -189,45 +95,20 @@ jobs:
...
+
+      - name: Test SciPy
+        working-directory: .github/workflows
+        # mpmath tests are too slow for CI
```
lucascolley, line 113: maybe we don't want the `mpmath` tests to be this slow. My investigation (ending around https://github.com/scipy/scipy/pull/23654#issuecomment-3342109949) shows that it seems pretty accidental that `test_jacobi_int` ends up making lots of very slow calls to `hyp2f1` with extreme arguments, while other tests do not, due to the function https://github.com/scipy/scipy/blob/f466f35b7d0ed33e6fd697cceccef175e256f074/scipy/special/_mptestutils.py#L39

I think this shouldn't get mixed any further with this PR, though :)

---

I suggest we leave this for now and come back to it when we convert https://github.com/scipy/scipy/blob/138c5f72dcfe325ea2ad5024be52e218481e89ea/.github/workflows/linux.yml#L289 which seems to be a job which both installs `mpmath` and sets `--mode=full`

lucascolley:
meson issue: https://github.com/mesonbuild/meson/issues/15055

lucascolley (commented):
I'm pretty confident in the correctness of this PR now. I plan to merge it after a few days if there are no comments.

rgommers:
> meson issue: [mesonbuild/meson#15055](https://github.com/mesonbuild/meson/issues/15055)

It would be good to ping me before opening Meson issues in the future, since there's a good chance I know the answer and we avoid sending Meson devs down some rabbit hole. This is a known footgun and we stumbled over it multiple times before, which is why it's explained in `spin build --help`.

lucascolley:
Noted!

> it's explained in `spin build --help`

I think the approach in this PR is better though, in that `-Dc_args` get merged in (with priority over `CFLAGS`), rather than the example there which seems to completely override `CFLAGS`?

.github/workflows/macos.yml
rgommers: Sounds good to defer it. And changing to a non-terribly-slow number and opening an issue to change it back once someone has gotten to the bottom of the issue seems fine - we shouldn't make the full test suite very slow because of corner cases in mpmath.

rgommers (commented):
Had a scan through the diff, looks good! My one question is about the new `debug` environment - it adds only `lldb`, but `compilers` will give us GCC on Linux so you'd need `gdb` rather than `lldb`.

.github/workflows/macos.yml
lucascolley: Yep. Just to clarify, the slowness is coming from corner cases in `xsf` not `mpmath`.

rgommers:
> I think the approach in this PR is better though, in that `-Dc_args` get merged in

Yes agreed, this is nicer, it may be useful to update the `spin build` help text.

lucascolley:
> My one question is about the new debug environment - it adds only lldb, but compilers will give us GCC on Linux so you'd need gdb rather than lldb.

The env isn't including `compilers`, but yes, we should have a `gdb` task too. I hadn't decided yet whether to include both in one env for both platforms, or whether to have separate envs and
keep them to osx and linux respectively.

rgommers:
`gdb` on macOS isn't useful. Vice versa, `lldb` on Linux is - but only if you built with `clang`. So separate and platform-specific envs sounds better.

lucascolley:
> `gdb` on macOS isn't useful. Vice versa, `lldb` on Linux is - but only if you built with `clang`. So separate and platform-specific envs sounds better.

Done! And I opened gh-23696 regarding the slow `mpmath` test. Can we merge if this is green?

rgommers:
> Can we merge if this is green?

👍🏼  SGTM

PR #23655, opened 2025-09-20
ENH: stats.truncpareto: allow negative exponent

rodrigosilva17:
#### Reference issue
Closes gh-#23648 and gh-#21606.

With the help of @mdhaber , we were able to implement a minimal fix that allows the use of power laws on arbitrary domains without changing the functional form of the PDF. This enables generating samples on any domain without relying on shifting and scaling, which would otherwise alter the behavior of the PDF and result in samples that do not follow the intended distribution.

More details, including computations, can be found in the discussion here: https://github.com/scipy/scipy/issues/23648#issuecomment-3313762764


rodrigosilva17:
@mdhaber, I’m currently running into the following issue:
`    mu = self._munp(1, *goodargs)
         ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/circleci/repo/build-install/usr/lib/python3.11/site-packages/scipy/stats/_continuous_distns.py", line 10521, in _munp
    return b / (b-n) * (c**b - c**n) / (c**b - 1)
                        ~^^~
ValueError: Integers to negative integer powers are not allowed.`

I’m not sure how to resolve it. I believe I implemented everything as we discussed, but something seems off.

scipy/stats/tests/test_distributions.py
```diff
@@ -10222,6 +10222,14 @@ def test_pdf(self):
         ref = stats.pareto(b).pdf(x) / stats.pareto(b).cdf(c)
         assert_allclose(res, ref)
 
+    def test_pdf_negative(self):
+        # PDF is that of the truncated pareto distribution
+        b, c = -1.8, 5.3
+        x = np.linspace(1.8, 5.3)
+        res = stats.truncpareto(b, c).pdf(x)
+        ref = stats.pareto(b).pdf(x) / stats.pareto(b).cdf(c)
```
mdhaber: Interesting - does this work? `pareto` should be restricted  to positive `b`, right? I'll suggest an adjustment.

mdhaber (commented):
Thanks @rodrigosilva17! I will check out the `fit` method because that is the only method I think might give trouble, and I'll suggest some updates to the documentation.

mdhaber:
> ValueError: Integers to negative integer powers are not allowed.

Ah, the old infrastructure doesn't ensure that arguments are floats, so we've run into this many times. You should just need to update that function to cast `b` as a `float`.

scipy/stats/tests/test_distributions.py
rodrigosilva17: I think it does. The standard Pareto requires $b > 0$, but I modified it to allow all values except $b = 0$, as we discussed.

**Updated**

```
def _argcheck(self, b, c):
        b = float(b)  # ensure float
        return (b != 0.) & (c > 1.)
```

mdhaber: You modified `truncpareto`. You didn't modify `pareto`, and it *shouldn't* accept $b < 0$ because the integral doesn't converge on an infinite domain. It produces NaNs, and the test fails.

scipy/stats/tests/test_distributions.py
```diff
@@ -10222,6 +10222,14 @@ def test_pdf(self):
...
+        assert_allclose(res, ref)
```
mdhaber: How about:
```suggestion
        # truncpareto is equivalent to more general powerlaw from gh-23648
        # exponent of truncpareto is negative in this case
        a, xmin, xmax = 4, 3, 5
        x = np.linspace(xmin, xmax)

        # compute reference using PDF from gh-23648
        C = a / (xmax ** a - xmin ** a)
        ref = C * x ** (a - 1)

        # compute using `truncpareto` with negative exponent
        b = -a
        c = xmax / xmin
        scale = xmin
        loc = 0
        res = stats.truncpareto(b, c, loc, scale).pdf(x)

        assert_allclose(res, ref)
```
This is what I meant when in https://github.com/scipy/scipy/issues/23648#issuecomment-3313762764 I wrote "where the reference value is obtained using your formulation of the powerlaw PDF."

scipy/stats/tests/test_distributions.py
rodrigosilva17: Ah, I see, I hadn’t considered that. Thank you very much for your help! I’m going to check again to see if it works.

scipy/stats/_continuous_distns.py
```diff
@@ -10476,12 +10476,13 @@ class truncpareto_gen(rv_continuous):
...
 
     def _shape_info(self):
-        ib = _ShapeInfo("b", False, (0.0, np.inf), (False, False))
+        ib = _ShapeInfo("b", False, (-np.inf, np.inf), (False, False))
         ic = _ShapeInfo("c", False, (1.0, np.inf), (False, False))
         return [ib, ic]
 
     def _argcheck(self, b, c):
-        return (b > 0.) & (c > 1.)
+        b = float(b)  # ensure float
```
mdhaber: This is not where the error is occuring, though. I'll push the required adjustments.

mdhaber:
Looks like the `log` methods don't work:
```python3
b = -2.
c = 3

stats.truncpareto.logpdf(2, b, c)  # nan
stats.truncpareto.logcdf(2, b, c)  # nan
stats.truncpareto.logsf(2, b, c)  # nan
```
It is probably just the way that the expressions are implemented that is causing problems (e.g. if we refactor $\log(b^2)$ as $2\log(b)$, it won't work with negative $b$ anymore). Can you take a look into alternative formulations that will not try to take the log of a negative value?

Then let's add tests. They can be very simple, just asserting that `truncpareto.logpdf(x, b, c)` is close to `log(truncpareto.pdf(2, b, c)`, etc.

scipy/stats/_continuous_distns.py
mdhaber: @rodrigosilva17 Please see https://github.com/rodrigosilva17/scipy/pull/2 for fixes and merge if you think that looks good. I'm still leaving the `logpdf`/`logcdf`/`logsf` fixes to you, but LMK if you run into trouble.

rodrigosilva17: Your fixes look good. I’ll go ahead and merge them. I’ll now look into implementing `logpdf`, `logcdf`, and `logsf`. Thanks!

rodrigosilva17:
> Looks like the `log` methods don't work:
> 
> ```python
> b = -2.
> c = 3
> 
> stats.truncpareto.logpdf(2, b, c)  # nan
> stats.truncpareto.logcdf(2, b, c)  # nan
> stats.truncpareto.logsf(2, b, c)  # nan
> ```
> 
> It is probably just the way that the expressions are implemented that is causing problems (e.g. if we refactor log ⁡ ( b 2 ) as 2 log ⁡ ( b ) , it won't work with negative b anymore). Can you take a look into alternative formulations that will not try to take the log of a negative value?
> 
> Then let's add tests. They can be very simple, just asserting that `truncpareto.logpdf(x, b, c)` is close to `log(truncpareto.pdf(2, b, c)`, etc.



I think the easiest way to do it is like this (I tested it in a local console, for instance, for the PDF).

```
def pdf(x, b, c):
        return b * x**-(b+1) / (1 - 1/c**b)

def logpdf(x, b, c):
        pdf = b * x**-(b+1) / (1 - 1/c**b)
        return np.log(pdf)     
```
        
Since the PDF works for $b \neq 0 $, the log-PDF will also work, because it doesn’t have to deal with the workarounds required by the other formulas that fail when $b<0$. I’m just not sure if it will be slower.

I tested it against `stats.truncpareto.logpdf(x, b, c)`, and it returns the same values for $b > 0$

rodrigosilva17:
@mdhaber, there is an issue with refguide ; I’ve included it below.

```
scipy.stats.truncpareto:87: (WARNING/2) Definition list ends without a blank line; unexpected unindent.:: 

scipy.stats.truncpareto:94: (ERROR/3) Error in "rubric" directive::: no content permitted.
```

Do you have any idea what it could be? Thanks!

mdhaber:
> I think the easiest way to do it is like this

Yes, that's the easiest way to do it, but it misses why we implement `logpdf` rather than relying on the generic implementation (which works as you describe if the class doesn't override `_logpdf`). The purpose is to avoid problems with over/underflow during evaluation and to precisely represent values too close to 0 or 1 to be represented in linear-space (as opposed to log-space).

~~I'll go ahead and implement this, but we'll need to get another reviewer to merge this PR.~~ I didn't end up implementing anything custom; I just used `apply_where` to select between the existing implementation for positive `b` and the default implementation for negative `b`. 

I'll fix the CI issue, too.

Thanks for fixing the typos in my LaTeX. The semicolon to separate the argument $x$ from shape parameters of the distribution was intentional, though. It's not used consistently in the docs, but it is useful, common notation in statistics literature.

scipy/stats/_continuous_distns.py
```diff
@@ -8600,7 +8600,13 @@ class powerlaw_gen(rv_continuous):
     For example, the support of `powerlaw` can be adjusted from the default
     interval ``[0, 1]`` to the interval ``[c, c+d]`` by setting ``loc=c`` and
     ``scale=d``. For a power-law distribution with infinite support, see
-    `pareto`.
+    `pareto`. For a power-law distribution described by PDF:
+
+    .. math::
+
+        f(x; a, l, h) = \frac{a}{h^a - l^2} x^{a-1}}
```
mdhaber:
```suggestion
        f(x; a, l, h) = \frac{a}{h^a - l^2} x^{a-1}
```

scipy/stats/_continuous_distns.py
```diff
@@ -10463,6 +10469,21 @@ class truncpareto_gen(rv_continuous):
...
 
+    The ``fit`` method assumes that :math:`b` is positive; it does not produce
+    good results when the data is more consistent with negative :math:`b`.
+
+   `truncpareto` can also be used to model a general power law distribution
+    with PDF:
+
+    .. math::
+
+        f(x, a, l, h) = \frac{a}{h^a - l^a} x^{a-1}
```
mdhaber:
```suggestion
        f(x; a, l, h) = \frac{a}{h^a - l^a} x^{a-1}
```

mdhaber:
@rodrigosilva17 please see https://github.com/rodrigosilva17/scipy/pull/3

rodrigosilva17:
> @rodrigosilva17 please see [rodrigosilva17#3](https://github.com/rodrigosilva17/scipy/pull/3)

Looks good. I’ll go ahead and merge it and run some tests. Thank you!@mdhaber!

mdhaber:
The remaining CI failures are in main.

mdhaber:
This is a pretty straightforward improvement, so I'll go ahead and merge. The only complications were that:

- negative integer `b` ran into the "integer to a negative integer power" gotcha,  but `xp_promote` is handy for ensuring that those integers are promoted to the appropriate floating point type,  and
- the `log` methods didn't work out of the box for negative `b`, but it was easy to fall back to the generic implementation in that case.

If someone wants to remove the $b \neq 0$ restriction or improve `logpdf`/`fit` with negative $b$, that can be another PR. In the meantime, this resolves two issues, so in it goes.

Thanks @rodrigosilva17!

PR #23658, opened 2025-09-21
CI: Pixi: use `no-pin` strategy and refactor environments

lucascolley:
See gh-23654 for the discussion that led to this PR, and which this PR will unblock. See also gh-21489 for the overarching issue for this work, and gh-23637 for discussion specific to lock file updates.

---

The idea for pinning which motivates these changes:
- pins will largely be enforced by the lock file (`pixi.lock`), not by the manifest (`pixi.toml`)
	- except for those maintained manually, like those to select a Python version, or those like:
```toml
# JAX 0.6.2 and 0.7.0 segfault on CUDA
jaxlib = { version = "!=0.6.2,!=0.7.0", build = "cuda12*" }
```
- those pins will be refreshed infrequently in the main repo
	- something like every 3 weeks was suggested (to reduce churn and repo bloat)
	- this could be automated with something like https://github.com/modular/modular-community/blob/main/.github/workflows/update-lockfiles.yml
- a separate repo could automate more frequently refreshing those pins and running tests to smoke out problems with new releases of dependencies ahead of time
	- this would keep that noise away from the main repo, but make sure we can keep on top of our dependencies breaking things more steadily, rather than being hit with it all at once every 3 weeks
	- we could automate opening issues to the main repo with something like https://github.com/scikit-learn/scikit-learn/blob/main/maint_tools/update_tracking_issue.py

---

Motivation for the other changes here:
- solve groups now accurately reflect which environments we expect to be identical for the intersection of their dependencies
- splitting up the `build` and `test` features into `{build, test}-{deps, task}` allows us to include the build and test deps in many environments, without:
	- having the `build` and `test` tasks defined in multiple envs. Hence, for the average contributor who just wants to `build` and `test` with default settings, they are not prompted to select an environment on `pixi run {build, test}`
	- having to include those deps in the default feature, and thus require `no-default-feature = true` in all environments which don't want them (this would only be a small problem right now, but probably wouldn't scale well)
- `--skip-deps` is included in workflows for a small speedup where we have already ran the `build` task in an earlier step
- this:
```
warning: The following rules have been removed and ignoring them has no effect:
    - UP038
```


.github/workflows/pixi.toml
```diff
@@ -6,53 +6,82 @@ channels = ["https://prefix.dev/conda-forge"]
...
+ninja = "*"
+meson = "*"
+meson-python = "*"
+cython = "*"
+python-build = "*"
+pip = "*"
+blas-devel = "*"
+numpy = "*"
+pybind11 = "*"
+pythran = "*"
```
lucascolley, line 104: This can probably be simplified to remove `meson` and `ninja` (transitive dependencies of `meson-python` that are unlikely to change), and `python-build` and `pip` (I think they would belong in a separate feature for wheel-building?).


I guess we could also remove `numpy` as a transitive dependency of `pythran`, but at that point we almost certainly hit zen of python line 2.

.github/workflows/pixi.toml
```diff
@@ -6,53 +6,82 @@ channels = ["https://prefix.dev/conda-forge"]
...
+hypothesis = "*"
+pytest-cov = "*"
+pytest-timeout = "*"
+pytest-xdist = "*"
+threadpoolctl = "*"
+pooch = "*"
+mpmath = "*"
+gmpy2 = "*"
+ccache = "*"
+marray-python = "*"
```
lucascolley: we could split optional test deps out from core test deps, but I think it is fine to punt on that for now

.github/workflows/pixi.toml
lucascolley: I remembered that including transitive dependencies can sometimes be useful anyway to coerce the solver into choosing more recent versions, so let's keep them here. I will remove `python-build` and `pip`, though.

lucascolley:
@bollwyvl I wonder whether I can bother you to ask whether this looks idiomatic? (Both the manifest here and the strategy in the top-post.) The vision is to eventually support Pixi as a first-class developer tool in-tree, once it is battle-tested in CI.

lucascolley:
The latest commit does some more feature splitting to ensure that tasks are isolated to a single environment.

lucascolley:
We need to ensure that Python versions match across build envs and their dependent test envs.

One way to do that would be to introduce a `py_default` feature, however it may be more robust to just have a separate build task per solve group.

.github/workflows/pixi.toml
```diff
@@ -119,77 +184,102 @@ description = "Build with MKL BLAS"
...
+
+[feature.build-cpu.tasks.build-cpu]
+cmd = "spin build --build-dir=build-cpu --setup-args=-Dblas=blas --setup-args=-Dlapack=lapack --setup-args=-Duse-g77-abi=true"
+cwd = "../.."
+env = { CC = "ccache $CC", CXX = "ccache $CXX", FC = "ccache $FC" }
+description = "Build SciPy (for CPU environments)"
+
+# XXX: workaround for meson mkl blas detection
+[feature.build-cpu.target.linux-64.tasks.build-cpu]
+cmd = "spin build --build-dir=build-cpu --setup-args=-Dblas=mkl-dynamic-ilp64-seq --setup-args=-Duse-g77-abi=true"
```
lucascolley, line 207: not sure if there is a good issue at which to point for this

.github/workflows/pixi.toml
```diff
@@ -6,53 +6,123 @@ channels = ["https://prefix.dev/conda-forge"]
...
+
+[environments.build]
+# tasks: build
+features = ["build-deps", "build-task"]
+solve-group = "default"
+
+[environments.test]
+# tasks: test
+features = ["build-deps", "test-deps", "test-task"]
+solve-group = "default"
```
lucascolley, line 20: It may make more sense to roll the `default` and `array-api-cpu` solve groups into one, to allow for no-rebuild switching between `test` and `test-cpu`.

We could add a `latest` solve group down the line if we wanted to test in an env which is likely to solve to the very latest versions of our core deps.

lucascolley: that can of course wait while this lives under `.github`

.github/workflows/pixi.toml
rgommers: Good to keep transitive dependencies in this case I think. Also note that `ninja` isn't a dependency of `meson-python` for isolated wheel builds, it's installed from PyPI if not already installed. So best to have it explicitly listed for that reason.

We're going to need `python-build` and/or `pip` in some specialty jobs, but fine to remove here - we can always put them back later anyway.

.github/workflows/pixi.toml
rgommers: Fine to skip for now, it'll be useful at some point.

.github/workflows/pixi.toml
rgommers: Not really, this is fine as is.

.github/workflows/pixi.toml
rgommers: agreed, fine to leave for later

.github/workflows/pixi.toml
```diff
@@ -6,53 +6,123 @@ channels = ["https://prefix.dev/conda-forge"]
...
-threadpoolctl = ">=3.6.0,<4"
-pooch = ">=1.8.2,<2"
-mpmath = ">=1.3.0,<2"
-gmpy2 = ">=2.2.1,<3"
-ccache = ">=4.11.3,<5"
-marray-python = ">=0.0.12,<0.0.13"
+[feature.test-deps.dependencies]
+pytest = "*"
+hypothesis = "*"
+pytest-cov = "*"
```
rgommers: Can we remove `pytest-cov` here? It's not used at all, except for in one prerelease job which we probably won't do with Pixi anyway.

.github/workflows/pixi.toml
```diff
@@ -6,53 +6,123 @@ channels = ["https://prefix.dev/conda-forge"]
...
+pytest-timeout = "*"
+pytest-xdist = "*"
+threadpoolctl = "*"
+pooch = "*"
+mpmath = "*"
+gmpy2 = "*"
+ccache = "*"
+marray-python = "*"
```
rgommers: `marray` is an array API dependency only, right? If so, it should be moved and treated similarly to, e.g., `dask`.

.github/workflows/pixi.toml
```diff
@@ -6,53 +6,123 @@ channels = ["https://prefix.dev/conda-forge"]
...
+
+[feature.build-deps.dependencies]
+compilers = "*"
+pkg-config = "*"
+ninja = "*"
+meson = "*"
+meson-python = "*"
+cython = "*"
+blas-devel = "*"
+numpy = "*"
```
rgommers: It doesn't quite make sense to me to have `numpy` only in build deps and not test deps. It's always needed, it should probably be in the default dependencies.

.github/workflows/pixi.toml
```diff
@@ -6,53 +6,123 @@ channels = ["https://prefix.dev/conda-forge"]
...
+pytest = "*"
+hypothesis = "*"
+pytest-cov = "*"
+pytest-timeout = "*"
+pytest-xdist = "*"
+threadpoolctl = "*"
+pooch = "*"
+mpmath = "*"
+gmpy2 = "*"
+ccache = "*"
```
rgommers: `ccache` is a build dep, not a test one

rgommers:
Overall this look good. My main concern is the `= '*'` versions. I don't see how that helps, all it does is remove info from `pixi.toml` and keep it only in `pixi.lock`, making it harder to see what the actual versions are without running pixi locally. A version bump is a very slightly smaller diff yes, but I don't think that's a useful gain. Am I missing something?

.github/workflows/pixi.toml
lucascolley: sounds like we should introduce a `host-deps` feature

EDIT: or rather `run-deps`

lucascolley:
> Am I missing something?

I think so!

> all it does is remove info from pixi.toml and keep it only in pixi.lock ... A version bump is a very slightly smaller diff

The point is that version bumps can now happen automatically on `pixi update`. Without this, automating version bumps becomes quite tricky, needing effectively a `pixi upgrade` for all dependencies *apart from* those which we want to keep pinned manually. That seems feasible but I'm not sure it's particularly easy or maintainable.

> making it harder to see what the actual versions are without running pixi locally

Sure, but that could be mitigated by automating placing `pixi list --explicit` output somewhere for each env? That also seems feasible and would require no maintenance effort when manual pins change.

---

EDIT: deleted a further comment because it didn't make a great point

lucascolley:
> making it harder to see what the actual versions are without running pixi locally

And supposing we implemented something in `pixi upgrade` to make having an 'ignore-list' more maintainable, you then run into the opposite problem that it makes it harder to see which versions are pinned out of necessity, and which are pinned to just some recent version (because everything is pinned in the manifest).

That problem could be somewhat mitigated by code comments in the manifest or more considered documentation, but my gut feeling is that we're still in a worse situation after that than if we go with `no-pin`.

lucascolley:
I know you said you are busy, so rolling with this and noting in the issue to follow up at some point might be a good idea for now.

.github/workflows/pixi.toml
```diff
@@ -6,53 +6,132 @@ channels = ["https://prefix.dev/conda-forge"]
...
+pkg-config = "*"
+ninja = "*"
+meson = "*"
+meson-python = "*"
+cython = "*"
+pythran = "*"
+pybind11 = "*"
+# needed at build-time as well as run-time
+numpy = "*"
+blas-devel = "*"
```
lucascolley, line 108: does this look correct now @rgommers ? Now only `numpy` and `blas-devel` are included in the non-build envs.

.github/workflows/pixi.toml
```diff
@@ -6,53 +6,132 @@ channels = ["https://prefix.dev/conda-forge"]
...
+solve-group = "array-api-cpu"
+
+[environments.array-api-cpu]
+# tasks: test-cpu
+features = ["run-deps", "test-deps", "test-cpu", "mkl", "array_api_strict", "dask", "jax-cpu", "marray", "torch-cpu"]
+solve-group = "array-api-cpu"
+
+[environments.build-cuda]
+# tasks: build-cuda
+# note: py-cuda feature ensures compatibility with both cuda12 and cuda13 solve groups
```
```suggestion
# note: py-cuda feature aims to ensure compatibility with both cuda12 and cuda13 solve groups
```

bollwyvl:
> @bollwyvl I wonder whether I can bother you to ask whether this looks idiomatic? (Both the manifest here and the strategy in the top-post.) The vision is to eventually support Pixi as a first-class developer tool in-tree, once it is battle-tested in CI.

As for deps: I think pins (especially bottom pins) are _great_ and helpful if they are _real_, as looking at intentional changes on a PR are more useful than having to resort to "forensic" analysis at the lockfile level. For big solves, they are probably also slightly faster, and reduce the risk of weird local minima, especially "big rocks" like `python`, compilers, etc.

That being said: making as-small-as-possible features, as this appears to do, is a good direction for a big project. I do wish there were more places to put `description` (e.g. on `feature`, `environment`), provided there were good places to put these in CLI output, especially when things go wrong.

As for tasks: being specific enough to remove the "which env?" prompt is very good. I might be tempted to use some `args` and template tasks, and potentially some whitespace to really tease out where there coudl be reuse... but I've had mixed results with mixing args and env vars things in complex task trees... as noted elsewhere, it's not _entirely_ clear which fields of `depends-on` and `tasks` will use `jinja` features.

.github/workflows/pixi.toml
rgommers: yep looks good now!

lucascolley:
alright, I'll merge on Thursday and update relevant issues if there are no objections

lucascolley:
rebasing gh-23654 on top of this seems to have gone smoothly

EDIT: CI failure here is just https://github.com/scipy/scipy/pull/23467#issuecomment-3328510476

PR #23662, opened 2025-09-22
ENH: signal: Convert Savitzky-Golay filter to Array API

ev-br:
#### Reference issue

towards https://github.com/scipy/scipy/issues/20678

Convert `signal.savgol_filter` and `signal.savgol_coeffs` to be Array API compatible

#### Additional information

- `savgol_coeffs` receives scalars and return arrays, hence gets new arguments, `xp=None, device=None`, similar to `signal.windows` and `fft`
-<s> both functions used to rely on `lstsq`, which is not a part of Array API linalg extension; Replace it with `pinv` and a manual call to SVD (where we emit a warning for rank-deficient problems)</s> EDIT: Use `scipy.linalg.lstsq` for numpy arrays, backend-specific `xp.linalg.lstsq` if available, and fall back to a manual svd computation otherwise.
- `dask` backend is limited in several ways, needs to be skipped all around

### DIFF
--- a/scipy/signal/_polyutils.py
+++ b/scipy/signal/_polyutils.py
@@ -7,7 +7,15 @@ To distinguish the two sets, the "new-style" routine names start with `npp_`
 """
 import warnings
 import scipy._lib.array_api_extra as xpx
-from scipy._lib._array_api import xp_promote, xp_default_dtype, xp_size, xp_device
+from scipy._lib._array_api import (
+    xp_promote, xp_default_dtype, xp_size, xp_device, is_numpy
+)
+
+try:
+    from numpy.exceptions import RankWarning
+except ImportError:
+    # numpy 1.x
+    from numpy import RankWarning
 
 
 def _sort_cmplx(arr, xp):
@@ -68,6 +76,39 @@ def _trim_zeros(filt, trim='fb'):
     return filt[first:last]
 
 
+# For numpy arrays, use scipy.linalg.lstsq;
+# For other backends,
+#   - use xp.linalg.lstsq, if available (cupy, torch, jax.numpy);
+#   - otherwise manually compute pseudoinverse via SVD factorization
+def _lstsq(a, b, xp=None, rcond=None):
+    a, b = xp_promote(a, b, force_floating=True, xp=xp)
+
+    if rcond is None:
+        rcond = xp.finfo(a.dtype).eps * max(a.shape[-1], a.shape[-2])
+
+    if is_numpy(xp):
+        from scipy.linalg import lstsq as s_lstsq
+        return s_lstsq(a, b, cond=rcond)
+    elif lstsq_func := getattr(xp.linalg, "lstsq", None):
+        # cupy, torch, jax.numpy all have xp.linalg.lstsq
+        return lstsq_func(a, b, rcond=rcond)
+    else:
+        # unknown array library: LSQ solve via pseudoinverse
+        u, s, vt = xp.linalg.svd(a, full_matrices=False)
+
+        sing_val_mask = s > rcond
+        s = xpx.apply_where(sing_val_mask, (s,), lambda x: 1. / x, fill_value=0.)
+
+        sigma = xp.eye(s.shape[0]) * s    # == np.diag(s)
+        x = vt.T @ sigma @ u.T @ b
+
+        rank = xp.count_nonzero(sing_val_mask)
+
+        # XXX actually compute residuals, when there's a use case
+        residuals = xp.asarray([])
+        return x, residuals, rank, s
+
+
 # ### Old-style routines ###
 
 
@@ -92,6 +133,8 @@ def polyval(p, x, *, xp):
     x = xp.asarray(x)
     y = xp.zeros_like(x)
 
+    # NB: cannot do `for pv in p` since array API iteration
+    # is only defined for 1D arrays.
     for j in range(p.shape[0]):
         y = y * x + p[j, ...]
     return y
@@ -171,21 +214,11 @@ def polyfit(x, y, deg, *, xp, rcond=None):
     scale = xp.sqrt(xp.sum(lhs * lhs, axis=0))
     lhs /= scale
 
-    # LSQ solve via pseudoinverse + rank check
-    u, s, vt = xp.linalg.svd(lhs, full_matrices=False)
-
-    sing_val_mask = s > rcond
-    s = xpx.apply_where(sing_val_mask, (s,), lambda x: 1. / x, fill_value=0.)
-
-    sigma = xp.eye(s.shape[0]) * s    # == np.diag(s)
-    c = vt.T @ sigma @ u.T @ y
+    c, _, rank, _ = _lstsq(lhs, y, rcond=rcond, xp=xp)
     c = (c.T / scale).T  # broadcast scale coefficients
 
     # warn on rank reduction, which indicates an ill conditioned matrix
-    rank = xp.count_nonzero(sing_val_mask)
     if rank != order:
-        from numpy.exceptions import RankWarning
-
         msg = "Polyfit may be poorly conditioned"
         warnings.warn(msg, RankWarning, stacklevel=2)
 
--- a/scipy/signal/_polyutils.py
+++ b/scipy/signal/_polyutils.py
@@ -178,7 +178,7 @@ def polyfit(x, y, deg, *, xp, rcond=None):
     s = xpx.apply_where(sing_val_mask, (s,), lambda x: 1. / x, fill_value=0.)
 
     sigma = xp.eye(s.shape[0]) * s    # == np.diag(s)
-    c = vt.mT @ sigma @ u.mT @ y
+    c = vt.T @ sigma @ u.T @ y
     c = (c.T / scale).T  # broadcast scale coefficients
 
     # warn on rank reduction, which indicates an ill conditioned matrix
--- a/scipy/signal/_polyutils.py
+++ b/scipy/signal/_polyutils.py
@@ -5,8 +5,9 @@ namespace, and "new-style", np.polynomial.polynomial, routines.
 
 To distinguish the two sets, the "new-style" routine names start with `npp_`
 """
+import warnings
 import scipy._lib.array_api_extra as xpx
-from scipy._lib._array_api import xp_promote, xp_default_dtype
+from scipy._lib._array_api import xp_promote, xp_default_dtype, xp_size, xp_device
 
 
 def _sort_cmplx(arr, xp):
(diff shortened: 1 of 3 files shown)


ev-br:
All green, finally. This PR is ready from my side.

scipy/signal/_polyutils.py
```diff
@@ -87,10 +88,12 @@ def _poly1d(c_or_r, *, xp):
...
     """ Old-style polynomial, `np.polyval`
     """
+    p = xp.asarray(p)
+    x = xp.asarray(x)
     y = xp.zeros_like(x)
 
-    for pv in p:
-        y = y * x + pv
+    for j in range(p.shape[0]):
+        y = y * x + p[j, ...]
```
DietBru, line 139: Nitpick: A comment here would be helpful, why it is not possible to iterate over the array.

scipy/signal/_polyutils.py
```diff
@@ -136,6 +139,59 @@ def polymul(a1, a2, *, xp):
     return val
 
 
+# https://github.com/numpy/numpy/blob/v2.3.3/numpy/lib/_polynomial_impl.py#L459
+def polyfit(x, y, deg, *, xp, rcond=None):
```
DietBru, line 186: I did not find any unit tests for `polyfit`. Is that intentional? 

Doing
```shell-session
$ pytest --cov=scipy/signal scipy/signal/tests/  && coverage html
```
revealed that none of the exceptions are raised during testing.

scipy/signal/_savitzky_golay.py
```diff
@@ -115,36 +118,41 @@ def savgol_coeffs(window_length, polyorder, deriv=0, delta=1.0, pos=None,
...
-    y = np.zeros(polyorder + 1)
+    y = xp.zeros(polyorder + 1, dtype=xp.float64, device=device)
     # The coefficient assigned to y[deriv] scales the result to take into
     # account the order of the derivative and the sample spacing.
-    y[deriv] = float_factorial(deriv) / (delta ** deriv)
+    y = xpx.at(y, deriv).set(float_factorial(deriv) / (delta ** deriv))
 
     # Find the least-squares solution of A*c = y
-    coeffs, _, _, _ = lstsq(A, y)
+    coeffs = xp.linalg.pinv(A) @ y
```
DietBru, line 151: The numerics of [pinv] appear to me to be quite different than [lstsq] (which is based on LAPACK's [dgelsd] and friends). I have no intuition if it makes a palpable difference here—I doubt thou  that the test cases are suited to reveal such problems. 

It might be helpful, to mention the change in the release notes in case somebody's unit tests break. 

[pinv]: https://scipy.github.io/devdocs/reference/generated/scipy.linalg.pinv.html
[lstsq]: https://scipy.github.io/devdocs/reference/generated/scipy.linalg.lstsq.html
[dgelsd]: https://www.netlib.org/lapack/explore-html/d9/d67/group__gelsd_ga0bee7e1b9e7e43f59ecf2419b2759c42.html

DietBru (commented):
Overall, looks good to me. A better coverage of `_polyutils.py` would make it look even better—though that could also be done in future PR.

scipy/signal/_savitzky_golay.py
j-bowhay: I have to admit this made me slightly nervous too, I wonder if we should use an `if_numpy` branch so that the results for existing users are not made worse?

WarrenWeckesser: I suspect this is probably not a good change.  I don't think I'll be able to take a closer look until later next week, but my guess is that computing an explicit pseudo-inverse and multiplying will be significantly slower than using `lstsq`.

I haven't been closely following the xp updates.  Are benchmarks (at least informal ones) being run on the changes?

ev-br: IIUC, the numerics of `lstsq` and `pinv` are equivalent---both are (rather thin) wrappers on top of an SVD factorization. The main (only?) difference is the LAPACK routine:
- scipy.linalg.lstsq uses `gelsd`, https://github.com/scipy/scipy/blob/main/scipy/linalg/_basic.py#L1754
- numpy.linalg.lstsq uses `gelsd`, https://github.com/numpy/numpy/blob/v2.0.0/numpy/linalg/umath_linalg.cpp#L4037C22-L4037C32

- scipy.linalg.pinv uses `gesdd`, via the default of `scipy.linalg.svd`, https://github.com/scipy/scipy/blob/main/scipy/linalg/_decomp_svd.py#L17 
- numpy.linalg.svd uses `gesdd`, also via the `numpy.linalg.svd`'s default https://github.com/numpy/numpy/blob/v2.0.0/numpy/linalg/umath_linalg.cpp#L2962

I've to admit I don't know if there's a non-historical reason behind the choice of the driver?

---------------------------------------

The `is_numpy` suggestion is, I suppose, the same as https://github.com/scipy/scipy/pull/23507#pullrequestreview-3184134519 

In this case my initial though is to minimize the use of `is_numpy` branching unless there's a pressing reason to. Here we already carry some risk for small backwards compatible changes due to `polyfit`, so it did not pass my threshold. Am happy to add branching if there are strong opinions though.

Timings: I think the main reasoning is a promise of potentially large performance improvements on alternative backends. Those strongly depend on details though --- see, for instance, https://github.com/scipy/scipy/pull/23447, where we can get from x3 slowdown to x40 speedup, depending on details (https://github.com/ev-br/bench_playground/blob/master/viz.rbf.ipynb).

That said, I agree we should prefer to avoid severe performance regressions for existing usage. 

Here, quick-and-dirty performance checks does show _some_ slowdown from `pinv(A) @ y` (see just below). A question however is whether `savgol_filter` is likely to end up in a hot path where <= 1ms / call  becomes a bottleneck?

```
n [1]: import numpy as np

In [2]: from scipy.linalg import pinv, lstsq

In [3]: rng = np.random.default_rng(1234567)

In [4]: N = 10

In [6]: a = rng.uniform(size=(N, N))

In [7]: b = rng.uniform(size=N)

In [8]: %timeit lstsq(a, b)
47.4 μs ± 345 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [9]: %timeit pinv(a) @ b
78.2 μs ± 283 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [10]: N = 100

In [11]: a = rng.uniform(size=(N, N))

In [12]: b = rng.uniform(size=N)

In [13]: %timeit lstsq(a, b)
1.95 ms ± 304 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [14]: %timeit pinv(a) @ b
3.07 ms ± 60.6 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [15]: %timeit pinv(a)
2.03 ms ± 27.3 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

```

scipy/signal/_polyutils.py
ev-br: Sigh. The rabbit hole deepens :-).

`_polyutils/polyfit` here is a helper for `savgol` only, and is tested through `savgol_filter` and `savgol_coeffs`, indeed. This is not a complete coverage, of course. I probably should port some tests from `numpy/polynomial`.
I am still hoping that the whole `_polyutils` vendor is a temporary plug for a more general solution, though:
https://github.com/numpy/numpy/issues/29680#issuecomment-3342656197

scipy/signal/_polyutils.py
ev-br: Short reason: iteration over an array is only defined for 1D arrays in the Array API. 
Will add a comment to this effect, together with other updates.

scipy/signal/_savitzky_golay.py
ev-br: One other `lstsq` related thing. Do we expect a need to change the LAPACK driver @WarrenWeckesser ? 
The `lapack_driver` argument is what tells `scipy.linalg` version from the `numpy.linalg` version. So if using `np.linalg.lstsq` is OK here, we can think of doing `getattr(xp.linalg, "lstsq")` with a fall-back to `xp.linalg.pinv`--- either in scipy or potentially in array-api-extra. 

The `getattr` approach works with `numpy`, `torch`, `jax.numpy` and `cupy` --- all of which have `linalg.lstsq`.

DietBru: The function `lstsq` is used six times in the signal module. Hence, this is kind discussion will probably pop up five more times. Would it be a way forward to implement for `lstsq` a `pinv` backend, i.e., `lstsq(..., lapack_driver='pinv')`? It could be chosen by default for Array API backends. If someone complains, the `lapack_driver` parameter could also become a parameter for `savgol_coeffs`.

ilayn: `pinv` + `matmul` is numerically worse than `lstsq` similar to `inv` + `matmul` is than `solve`. Speed is not so much of an issue but numerics are. 

The question is why `pinv` has array api entry which has a much much less usage (since it is a glorified truncated SVD) and `lstsq` does not, which is pretty much the workhorse of nonsquare and rank deficient problems.

ev-br: Okay, updated to use the fully backwards compatible `scipy.linalg.lstsq`  for numpy arrays, and backend-specific `xp.linalg.lstsq` where available (torch, jax, cupy). And only otherwise fall back to a manual SVD computation: 
https://github.com/scipy/scipy/pull/23662/files#diff-530e4a76e112775eb03501a786ce61cd0c0ec330d6fc3317ae619c3e46d353abR83

> pinv + matmul is numerically worse than lstsq similar to inv + matmul is than solve

Wow, interesting. I thought that the main reason `solve(a, b)` is superior to  `inv(a) @ b` was pivoting, not the matmul itself.  Where does the error accumulate in `pinv(a) @ b` vs svd-based `lstsq(a, b)`?

> why pinv has array api entry which has a much much less usage (since it is a glorified truncated SVD) and lstsq does not

GH archeology shows some discussion from 2021, https://github.com/data-apis/array-api/pull/240 and references therein. Apparently, the API around returning residuals and singular values is somewhat clunky. 

Pragmatically, how about we try working around it using the version from this PR, and then see about one more attempt to add it to the array api spec.

steppi: I concur with @ilayn that `dgelsd` is better numerically than `dgessd` plus a matmul. The Wikipedia article, https://en.wikipedia.org/wiki/Numerical_methods_for_linear_least_squares, seems to give a good overview, see the section on orthogonal decomposition methods. It's kind of a shame that `lstsq` got left out of the standard. I'm on board with working around it for now, but still trying to get it added to the spec.

steppi (approved):
This looks good @ev-br. I'm willing to die on the hill that something like `lstsq` belongs in the standard, but it's a good pragmatic solution for now.

PR #23663, opened 2025-09-23
DOC: O(100) elements -> the order of 100 elements

star1327p:

We should write "100 elements," not "O(100) elements."


rkern:
The `O()` is intentional. Plain `100` would be false precision. You could write out "on the order of 100 elements", though, if you wanted to avoid the jargon notation.

star1327p:
@rkern Thank you for clarifying. I revised the phrase to "larger than the order of 100 elements".

j-bowhay:
Thanks @star1327p, on this occasion I think I'll close this as I think it is probably just semantics to which phrasing we use here

PR #23664, opened 2025-09-23
CI: bump pre-release deps job to 3.13. Avoids resource tracker issue

andyfaff:


lucascolley (approved):
thanks Andrew!

andyfaff:
I don't think the issue is in our code, it's probably somewhere upstream.

PR #23665, opened 2025-09-23
ENH:optimize: Improve SLSQP Progress Reporting

andrewellis55:
#### Reference issue
None

This is a small enhancement to the `iprint` feature of SLSQP. Currently, the `iprint` tracks the value of the objective function and the norm of the gradient as an indication of optimization progress. While these are great metrics for unconstrained optimization, the norm metric is not the best for constrained optimization as this won't tend towards 0 when there are active constraints. Instead, SLSQP given us some great metrics already in the form of `h1` and `h2` which are indications of optimality (or stationary) and feasibility respectively. In fact someone [made a whole repo](https://github.com/anugrahjo/PySLSQP) with one of the main goals being to expose these parameters.

Recent a reviewer of a paper I'm writing asked for a plot of the optimality vs time for a problem where I used SLSQP, and I ended up have to make this change to expose that. I figured it could be useful to others as well.

Additionally, I've added what I believe to be a concise interpretation of these parameters in the docstring. This can be removed if it's felt that this is unnecessary and @ilayn, please let me know if I misinterpreted the exact math there as I know you just spent a lot of time re-writing this so you're probably the most familiar with the exact formulation.

#### Example Output
Simple Problem:
```python
import numpy as np


def objective(x):
    """Quadratic well with optimum near (1, 0.5)."""
    return (x[0] - 1.0) ** 2 + (x[1] - 0.5) ** 2

def equality_constraint(x):
    return x[0] + x[1] - 1.0

def inequality_constraint(x):
    return x[1] - 10

x0 = np.array([0.5, 0.5])
constraints = [
    {"type": "eq", "fun": equality_constraint},
    {"type": "ineq", "fun": inequality_constraint},
]

result = minimize(
    objective,
    x0,
    method="SLSQP",
    constraints=constraints,
    options={"maxiter": 100, "ftol": 1e-9, "iprint": 2, "disp":True},
)
```
Optimality and Feasibility tend towards 0 as the problem converges, despite the active constraint
```
  NIT    FC           OBJFUN       OPTIMALITY      FEASIBILITY
    1     4     1.902500E+02     1.900000E+02     9.500000E+00 
    2     8     1.902500E+02     1.852500E+02     4.750000E+00 
    3    10     1.902500E+02     6.567547E-11     1.200817E-12 
```

Output from main (for reference). The GNORM increases over time, not really indicating convergence in any clear manner
```
  NIT    FC           OBJFUN            GNORM
    1     4     1.902500E+02     1.000000E+00
    2     8     1.902500E+02     1.415980E+01
    3    10     1.902500E+02     2.758623E+01
```


andyfaff:
I doubt that we would add functionality to iprint. We're much better off with adding attributes of an `intermediate_result` in a callback.
I like the additional documentation though.

andrewellis55:
I fully agree that if anyone is going to plot the results, that a callback is the best way forwards, but what's the opposition to improving the intermediate prints? I think dropping the printed output into an outfile is fairly common practice (especially among users that use SLSQP for optimization that take hours to run). If the worry is that this could be a breaking change for some users, I could always change this to require `iprint = 3` that way anyone who is currently running `iprint = 2` won't see any change in behavior.

Perhaps I can add a few more points to support this change.

**1. It bring SLSQP more in line with expensive moderns optimizers like SNOPT**
Here is an example of the [SNOPT outfield](https://ccom.ucsd.edu/~optimizers/static/pdfs/sndoc7.pdf)
image
Norm of the gradient is not provided as a field as it's not relevant to constrained optimization. Instead measures of optimality and feasibility are used to show the user how they are to minimizing the objecting and solving the constraints respectively. Both parameters, like those I've selected in SLSQP, are normalized in a manner that the closer they are to 0, the closer the problem is to being solved.

**2. It shows the user the progress of the algorithm towards the tolerance they selected**
In SLSQP, the user can define a tolerance, however without digging through the code, it's very unclear to an end user how this tolerance is bing used. SLSQP will iterate until both h1 and h2 are below that tolerance and with this print scheme, the user can now see how the algorithm progresses towards this threshold they previously defined. While the existing prints did give some indication how the progress towards the objective, no indication of if the constraints were becoming more or less feasible is provided without knowing the details of the algorithm.

**3. It provides an example of what to track in a callback**
I do like your suggestion of using a callback if a user needs to plot this for a publication, and this just gives a few lines of example code telling them what to plot for both optimality and feasibility, as h1 and h2 are not the most clear parameter names on their own. Frankly, when the reviewers of my paper asked me to plot this, it took me a few hours of work to dig through the code and figure out how exactly to do so with SLSQP. I'm just trying to make this a less painful process for users in the future, as convergence plots are not an uncommon ask with optimization papers.

andrewellis55:
I'm happy to make a post the on the SciPy discussion forum as well if you'd like, I just felt this likely wasn't such a major change as before.

andyfaff:
I've no opposition to providing more information on progress, but I would like it done by adding extra attributes to the `intermediate_result` callback. I'm happy with adding extra documentation.

One aspect of past development is that we didn't pay enough attention to uniformity. That is, if we add something to one minimize method we should consider how other methods do it, or how it would eventually be rolled out to those other methods. One example is the variety of parameters for tolerances that are used across the code base.
The most uniform way of doing this is by adding appropriate attributes to the callback, printing to screen is too hodge-podge across methods (keyword names, what value and type of value is provided to keyword). In any case the callback could just print things out if people wanted it.
The `disp` option is deprecated for L-BFGS-B and will be removed in 1.18, so printing things out is likely to be less common.

If we add appropriate attributes to the `intermediate_result` we should think about:

- what nomenclature for these values is used by other methods with constraints, e.g. `trust-constr`, `cobyqa`, etc. Are h1/h2 going to make sense?
- what nomenclature is used in the literature
- whether those values represent exactly the same thing for each of the methods.

ilayn:
There are a few orthogonal choices to be made in Python-based scientific tools. Especially when wrapping up archaic code like SLSQP or L-BFGS-B this is quite problematic because these code originate from an era where there were only one option to interact with the computation that is the terminal. Hence they dump everything and happily move forward.

Today we know better not to `printf` or `write` arbitrarily when it is uncalled for, and there are places things even run headless. Plus, there are decades went by until we now settle for actual logging instead of dummy printing so that we can direct the output to different channels stderr, stream or a file etc with formatting we choose. It is not the case with these old code where they just print and done. 

Having said that, there is a also an undeniable feature request to follow the progress for long running computations such that we don't wait for a futile search for hours only to find out at the end the futility of it. So the algorithms should give way to track progress. I think there is no discussion on this point that we need to enable this wherever possible. 

It is the "how" that is indeed the discussion. Some solvers use reverse-communication such that the control is passed to Python for jac and func evaluations, and things can be done there. Some accept the callback and stay in the low-level code. We can patch those callbacks to do the same, it's a bit more work but doable. C level printing or logging is very limited and I think it is not worth the effort to do more surgery.

I don't have any opinions on what should be printed really as optimize residents are more knowledgeable about the daily usage.

andrewellis55:
@andyfaff, looks like `intermediate_results` for SLSQP [ merged in the past few days](https://github.com/scipy/scipy/pull/23573). I actually wasn't aware of it before you mentioned it here. When you were referring to callback stuff I just assumed it would be on the user to implement, but I see now how `intermediate_results` gives it a bit more of a framework. Given this great new feature, I have no problem adding the progress tracking I've proposed to it. I do think that the printing should also be supported as the use case for outfiles is a very real use case, especially when optimizations are being run on remote or distributed infratructure.

I see what you're saying about uniformity across optimize if we're going to add parameters to the `OptimizeResult` object. `h1` and `h2` are definitely not the appropriate names to use in a shared object as they are just arbitrary parameters names chosen for SLSQP by the original code author. Personally I feel that `optimality` and `feasibility` are probably the best general terms to use for parameters as most algorithms have some concept of this (where `feasibility` of course only applies to constrained optimization). 

For gradient based unconstrained optimization, the  L2 norm of the gradient could probably be used for the `optimality` parameter (in the absence of an optimization specific output like SLSQP has). `feasibility` can be usually computed as the sum or max constraint violation (L1 or infinity norm), once again the the absence of anything optimizer specific. I'm a little less familiar with metrics for the gradient free optimizers, but I know they are quite widely used so I'm sure it would be worth some thought.

In terms of next steps, are you asking that I put together progress tracking for all optimizers through the `OptimizeResult` object? Or just that I include these parameters in an agreed upon and expandable way for SLSQP. If the ask is the former, I think that puts a pretty high barrier for entry for new contributions to optimize, in terms of both time commitment and required technical competency (i.e. for myself, there are quite a few optimizers in here that I have no experience at all with).

andyfaff:
> I do think that the printing should also be supported as the use case for outfiles is a very real use case, especially when optimizations are being run on remote or distributed infratructure.

I think that the export to file functionality could be done fairly simply in a callback.

> L2 norm of the gradient could probably be used for the optimality

The value of `jac` is easily placed into the `intermediate_result` (can't remember if SLSQP has that yet). I think it's better to provide that than its L2 norm - the user can easily calculate whatever statistic they want. Then again for `trust-constr` its output has `optimality`.

> feasibility can be usually computed as the sum or max constraint violation (L1 or infinity norm), once again the the absence of anything optimizer specific.

`trust-constr` has the following in its output:

```
optimality: float
Infinity norm of the Lagrangian gradient at the solution.

constr_violation: float
Maximum constraint violation at the solution.

lagrangian_grad: ndarray, shape (n,)
Gradient of the Lagrangian function at the solution.
```

I'm not sure if they're also present in the `intermediate_result`. `differential_evolution` has `maxcv` in its output, which would be the rough equivalent of `constr_violation`. I'm not sure that I like `constr_violation` for the maximum violation, `maxcv` is more descriptive.

So the way forward is to identify a common name across the methods, e.g. `constr_violation` for your desired attribute, then add it to the `intermediate_result` for SLSQP.

I'm not expecting that you implement something for all method. Rather, if you e.g. want to add a constraint violation, then work out the best name, then propose it, then implement it.

andrewellis55:
@andyfaff I've added these to `intermediate_results` in my latest commit and added a simple test case. I've taken your suggestion of `constr_violation`. I think this is better than `maxcv` as different methods track different convergence criteria, i.e. SLSQP uses sum of constraint violation rather than the max.

As a side note, this does raise the unrelated question of perhaps should SLSQP use max violation? As the sum of the violation scales with number of constraints, and so a problem with a million constraints could have trouble converging to typical tolerances on this basis alone, and I've definitely in the past had to loosen my tolerances as my number of constraints grew. Topic for another day perhaps.

andrewellis55:
Also skimming through the failed tests here, it seems to me like the failures are unrelated to my changes. Are sporadic failures ever expected in the SciPy CI? The only failures (from what I can tell) are as follows:

**run_benchmarks**
```
[ 9.04%] ··· [0;39mintegrate.NquadOscillatory.time_oscillatory[0m             1/6 failed
[ 9.04%] ··· [0;39m====== ========== ==========[0m
             --              rtol        
             ------ ---------------------
              ndim    1e-10      1e-11   
             ====== ========== ==========
               1     1.17±0ms   1.13±0ms 
               3     110±0ms    126±0ms  
               5      failed    48.3±0s  
             ====== ========== ==========
             For parameters: 5, 1e-10
             
             
             asv: benchmark timed out (timeout 60.0s)
```

**free-threaded**
```
PARALLEL FAILED scipy/sparse/tests/test_matrix_io.py::test_nd_coo_format[3-0] - zipfile.BadZipFile: File is not a zip file
PARALLEL FAILED scipy/sparse/tests/test_sparsetools.py::TestInt32Overflow::test_coo_todense - pytest.PytestUnraisableExceptionWarning: Exception ignored while finalizing file <_io.FileIO name='f.npz' mode='rb' closefd=True>: None
```

mdhaber:
Those failures do not look related to your changes. Failures in CI due to code in `main` are not ideal, but they do happen. If you see them occuring in other PRs but don't see an issue about them, please consider opening one.

andrewellis55:
I did some more testing and realized that as `h1` and `h2` are used for a few different things in different sections of the c code, and they can take unexpected values depending on the exit mode. [This change](https://github.com/scipy/scipy/compare/ed9517edd5eb914abc047a8d4fffd288631fb3eb..f1c51d4620f028d01eab7cc9feda6c0a4f9fe421) is quick fix to make sure that we're only ever reporting the values of `h1` and `h2` from the convergance check.

andyfaff:
Please only force push in a PR if it's really necessary. I'd started considering the changes, but then what I was looking at a couple of days ago has changed.

There's non uniformity here w.r.t feasibility and constr_violation referring to the same thing.

andrewellis55:
@andyfaff sorry, different places have different policies on how they expect contributions formatted. My interpretation of SciPy guidances was that a single commit was expected hence the force pushing. Is it typical to squash on merge to handle the commit message requirements?

And good point, I can unify those now to `constr_violation` everywhere if you can provide guidance as to whether you'd like it as a new commit or rebased into the original.

andrewellis55:
Here is the diff from the force pushes: https://github.com/scipy/scipy/compare/7d5fd09690eab23d4fd080842c6657d07853116d..f9c4d72f86d627f7d2fed08444b721308e4e3a11

nickodell:
>I think this is better than `maxcv` as different methods track different convergence criteria, i.e. SLSQP uses sum of constraint violation rather than the max.

I agree that `maxcv` is  a confusing name.

>sorry, different places have different policies on how they expect contributions formatted. My interpretation of SciPy guidances was that a single commit was expected hence the force pushing.

No, there's no rule against having multiple commits on a PR. It's very common. I would suggest making sure each individual commit follows the commit message guidelines. This is the best way to comply with them.

The developer documentation isn't supposed to imply that people should only make one commit. If you'd like to prevent future contributors from encountering the same confusion, then I would suggest making a documentation issue that points out the confusing language.

>Is it typical to squash on merge to handle the commit message requirements?

It depends on how messy the applicable PR is. From the "SciPy Core Developer Guide:"

>Squashing commits or cleaning up commit messages of a PR that you consider too messy is OK. Make sure though to retain the original author name when doing this. Squashing is highly recommended whenever commit messages do not (roughly) follow the guidelines in Writing the commit message.

However, multiple sequential commits is not the kind of "mess" this is referring to.

andrewellis55:
@andyfaff the  uniformity between feasibility and constr_violation is [now addressed](https://github.com/scipy/scipy/pull/23665/commits/8130e7a96426b071398751d67cfb43225db68bcc).

andrewellis55:
@andyfaff have you had the opportunity to consider this or https://github.com/scipy/scipy/pull/23533 any further? I'm happy to take some time to address any additional concerns or suggestions you may have.

scipy/optimize/_slsqp_py.py
```diff
@@ -221,25 +220,29 @@ def _minimize_slsqp(func, x0, args=(), jac=None, bounds=None,
...
+        Maximum number of iterations. Default value is 100.
     ftol : float
         Precision target for the value of f in the stopping criterion. This value
         controls the final accuracy for checking various optimality conditions;
         gradient of the lagrangian and absolute sum of the constraint violations
         should be lower than ``ftol``. Similarly, computed step size and the
         objective function changes are checked against this value. Default is 1e-6.
-    eps : float
-        Step size used for numerical approximation of the Jacobian.
+    iprint : int
```
andyfaff: My preference for altering the docstring would be to avoid describing `iprint`. The rationale for this is that:

- user has fuller control over where the printing goes via a callback. e.g. it can go to a logger, etc.
- the callback approach is more uniform over all the minimize methods, so if you implement a good 'printer' then it can be used widely.

Can you remove this little bit describing `iprint`.

scipy/optimize/tests/test_slsqp.py
```diff
@@ -396,6 +396,32 @@ def test_callback(self):
...
+            return x[1] - 10
+
+        x0 = np.array([0.5, 0.5])
+        cons = [
+            {"type": "eq", "fun": eq_con},
+            {"type": "ineq", "fun": ineq_con},
+        ]
+
+        minimize(obj, x0, method="SLSQP", constraints=cons, callback=callback)
+
```
andyfaff, line 496: Please add tests in the callback that do a separate calculation of `optimality` and `constr_violation` at a given position `intermediate_result.x`. These values should be compared to those reported in `intermediate_result`, to check that the latter is always correct.

scipy/optimize/_slsqp_py.py
andrewellis55: done!

scipy/optimize/tests/test_slsqp.py
andrewellis55: For this one, this could warrant a little discussion. `h1` and `h2` are computed at the start of the iteration (near the `ITER_START` goto in the C code), but we return the `x` to the intermediate callback at the end of the iteration. For example, if I print out the values after every sleep call from the test I originally added we see

```
 ITER     MODE               X0               X1               H1               H2
    1      FUN    -9.000000E+00     1.000000E+01     1.900000E+02     9.500000E+00
    1      FUN    -4.250000E+00     5.250000E+00     8.242296E-13     9.500000E+00
    1     GRAD    -4.250000E+00     5.250000E+00     0.000000E+00     9.500000E+00   <- h1 and h2 computed  at the start of ITER 2 using x=(0, 9.5) from the end if ITER 1
    2      FUN    -9.000000E+00     1.000000E+01     1.852500E+02     4.750000E+00   <- iter incremented, triggering the callback with x=(-9,10)
    2     GRAD    -9.000000E+00     1.000000E+01     0.000000E+00     4.750000E+00
    3      END    -9.000000E+00     1.000000E+01     6.567547E-11     1.200817E-12
```

`h1` and `h2` also rely on some internal values like Lagrange multipliers and `gs` which are not returned as part of the callback function (and probably not values that a user would normally need access to in an intermediate result.

How would you feel about instead asserting that the final values of these fall below `ACC` (`10*tol`) at the end of the optimization?

andrewellis55: I've added the above [here](https://github.com/scipy/scipy/pull/23665/commits/58369f6bba29f63923638e7c494bfca02522ba86), but I can take a look into what else can eve done to test this if you've like something more in depth

nickodell: I feel that if we're going to add these variables to aid debugging, we should be very sure that the formula we give is correct. As I understand it, this test would allow h1 and h2 to be off by an order of magnitude, and the only effect is that either SLSQP would loop an extra iteration, or h1 and h2 would be lower than expected (which the test allows.)

It also seems strange to add these variables for the purpose of reporting on optimization progress, and not test what they do while the optimization is in progress. That seems very important!

>h1 and h2 are computed at the start of the iteration (near the ITER_START goto in the C code), but we return the x to the intermediate callback at the end of the iteration. For example, if I print out the values after every sleep call from the test I originally added we see

Could you perhaps store the previous value of x in the callback, then look it up again when you get the callback for the next iteration?

e.g.

```
prev_x = None
def callback(intermediate_result):
    nonlocal prev_x
    if prev_x is not None:
        # do something here with the last x value
        pass
    # tests involving new variables
    prev_x = intermediate_result.x
```

andrewellis55: @nickodell Sorry for the delay, was quite busy with work over the past month. I've now added tests in [this commit ](https://github.com/scipy/scipy/pull/23665/commits/2af578ea14dd9cae12cb9e2d79c4f051ceeb547e) to show the intermediate outputs conform to the equations provided.

Also the `10*tol` that you noted before wasn't saying that we're within an order of magnitude, it was there because internal sleep uses an `ACC` variable for it's exit criteria. This is now made more clear

andrewellis55: @andyfaff just a note that the force push I did [here](https://github.com/scipy/scipy/compare/13cb89dc83f52612579c634e4ef39de151933064..2af578ea14dd9cae12cb9e2d79c4f051ceeb547e) was just to combine my linting and test addition into a single commit for a slightly cleaner git history. The two commits were made only a few minutes apart, I just forgot to lint before pushing the new changes.

andrewellis55:
Seems like tests weren't running because of some lock file issue. I've rebased off the latest main branch to see if that will fix it

andrewellis55:
Seems some tests are still not running, but the branch is up to date with main. Not too sure what the issue is
```
pixi install -e lint --locked
  /home/runner/.pixi/bin/pixi install -e lint --locked --manifest-path pixi.toml --color always
  Error:   × lock-file not up-to-date with the workspace
  
Error: The process '/home/runner/.pixi/bin/pixi' failed with exit code 1
```

mdhaber:
@andrewellis55 it seems to be a problem in `main` right now. pinging @lucascolley as our resident pixi expert.

lucascolley:
my bad, forgot to lock in https://github.com/scipy/scipy/pull/24071. I merged the fix in https://github.com/scipy/scipy/pull/24077, it should be fine if you merge main now.

andrewellis55:
@nickodell @andyfaff I've just fixed the merge conflicts on this PR. Would it be possible to do a review? Thanks.

andrewellis55:
@nickodell @andyfaff just another ping on this in case you have some time. I can resolve the conflict after the review to ensure there are no other major issues

PR #23669, opened 2025-09-23
DOC: linalg: document batch support

mdhaber:
#### Reference issue
Toward gh-21466

Adds documentation about batch support where it is missing. Sometimes there is already some reference to batching, but I think there should always be a link to the batch tutorial.

### DIFF
--- a/scipy/linalg/_matfuncs.py
+++ b/scipy/linalg/_matfuncs.py
@@ -233,8 +233,7 @@ def logm(A, disp=_NoValue):
 def expm(A):
     """Compute the matrix exponential of an array.
 
-    The documentation is written assuming array arguments are of specified
-    "core" shapes. However, array argument(s) of this function may have additional
+    Array argument(s) of this function may have additional
     "batch" dimensions prepended to the core shape. In this case, the array is treated
     as a batch of lower-dimensional slices; see :ref:`linalg_batch` for details.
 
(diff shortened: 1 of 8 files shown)


scipy/linalg/_matfuncs.py
```diff
@@ -233,6 +233,11 @@ def logm(A, disp=_NoValue):
 def expm(A):
     """Compute the matrix exponential of an array.
 
+    The documentation is written assuming array arguments are of specified
+    "core" shapes. However, array argument(s) of this function may have additional
+    "batch" dimensions prepended to the core shape. In this case, the array is treated
+    as a batch of lower-dimensional slices; see :ref:`linalg_batch` for details.
```
lucascolley: the documented shape already includes `...`, did you mean to put the other message here? Same for `sqrtm`

mdhaber: No, in those cases, I meant to start with the second sentence: 
```
    Array argument(s) of this function may have additional
    "batch" dimensions prepended to the core shape. In this case, the array is treated
    as a batch of lower-dimensional slices; see :ref:`linalg_batch` for details.
```
I'm going to take another pass once I get the docs to build.

mdhaber: Looks like I need to add that message to `clarkson_woodruff_transform`, too.

lucascolley: >  I meant to start with the second sentence:

yeah, that's what I meant

mdhaber:
I'm removing the note from `solve_circulant`. It has two `axis` arguments rather than assuming the core dimensions come at the end, and one of the defaults has the core dimensions at the beginning.

Adjusted `expm` and `sqrtm` as suggested. Added notes to `clarkson_woodruff_transform` and the matrix generation functions that I left out before.

This should be in pretty good shape now[^1], so marking ready for review.

[^1]: It's possible that there are still some functions with inconsistent documentation in the sense that they hint at batch behavior in some places but not others. If it were up to me, we would just include the standard note and not try to show `...` correctly and explain the implications precisely throughout the documentation. But for now, I'm leaving them be.

j-bowhay (approved):

PR #23671, opened 2025-09-24
DOC: Update stable version switcher button styles in CSS

### LINKED ISSUE #23587: DOC: Selected text in version dropdown not compatible with dark mode
https://docs.scipy.org/doc/scipy/dev/

Please see screenshot below:

Image

Desired outcome: The selected version text has lighter color.

OmBiradar:
Closes #23587

**Before**:

The default text is black in color
- If the dev version is opened, then the background color is orange
- If the old version is opened, then the background color is red
- If it's the latest version, it's just the dark background

**After**: 
An intuitive thing to do would be to have a green background for the latest version, which would, with minimal changes, fix the issue

The default text is black in color
- If the dev version is opened, then the background color is orange
- If the old version is opened, then the background color is red
- If it's the latest version, then the background color is green


j-bowhay:
Should this happen upstream in the pydata sphinx theme rather than in scipy?

OmBiradar:
I think it's how we choose to handle the colouring of the version selector. I see that in other projects using Sphinx, they don't use any color codes

references
1. https://matplotlib.org/stable/index.html
2. https://scikit-learn.org/dev/index.html

Maybe there is a way to handle it, but I guess that would require looking into the Sphinx community guidelines on these themes

stefanv:
Here's the PDST SCSS default definition:

```scss
button.btn.version-switcher__button {
  border-color: var(--pst-color-border);
  color: var(--pst-color-text-base);

  // Add a margin on narrow screens to avoid feeling cramped
  margin-bottom: 1em;

  @include media-breakpoint-up($breakpoint-sidebar-primary) {
    margin-bottom: unset;
  }

  &:hover {
    box-shadow: 0 0 0 $focus-ring-width var(--pst-color-secondary);
    border-color: transparent;
  }

  &:active {
    color: var(--pst-color-text-base);
    border-color: var(--pst-color-border);
  }

  &:focus-visible {
    border-color: transparent;
  }
}
```

OmBiradar:
The color-coding (`dev=orange`, `old=red`) is SciPy-specific, not PDST default, so fixing the dark-mode contrast here makes sense I guess. I'm happy to switch to a neutral, text-only approach if that’s preferred..

lucascolley (approved):
I reached out on the Scientific Python discord to check about this[^1], and seems like we are fine to proceed with the PR as is. We can always tweak in a follow-up if this doesn't work quite as intended.

Thanks @OmBiradar !

[^1]: https://discord.com/channels/786703927705862175/1045809495021998101/1433109939060084867

PR #23675, opened 2025-09-25
ENH: spatial.transform.RigidTransform: support broadcasting

amacati:
Add support for arbitrary leading batch dimensions for `RigidTransform`. 

This is the final planned PR for the series of updates to `scipy.spatial.transform` that enables full array API support. Remaining PRs should consist only of documentation updates and bug fixes.

## Motivation
Enables broadcasting for RigidTransforms in the same form as #23467. Implementation details are identical. There are no exceptions from broadcasting rules, everything is already consistent.

## Test times
I'm a bit worried about the test times. At the moment, testing `RigidTransform` for all backends takes approximately two minutes. This is in part driven by a few functions that test if all combinations of broadcastable shapes combine correctly. We could reduce test times by only testing a few select combinations if that becomes an issue.

## Documentation
As with `Rotation`, it would be great to document the expanded capabilities of `RigidTransform`. Since we are not introducing breaking changes, updates to the docs have been deferred to a follow-up PR. This should make reviewing more manageable.

## References
- #22500
- #23391
- #23425


scipy/spatial/transform/_rigid_transform_xp.py
```diff
@@ -37,12 +37,8 @@ def from_matrix(matrix: Array, normalize: bool = True, copy: bool = True) -> Arr
     # We delay lazy branch checks until after normalization to avoid overwriting nans
     # with the rotation matrix
     if not lazy and xp.any(~last_row_ok):
-        last_row_ok = xpx.atleast_nd(last_row_ok, ndim=1, xp=xp)
-        matrix = xpx.atleast_nd(matrix, ndim=3, xp=xp)
-        ind = int(xp.nonzero(~last_row_ok)[0][0])
         raise ValueError(
-            f"Expected last row of transformation matrix {ind} to be "
-            f"exactly [0, 0, 0, 1], got {matrix[ind, 3, xp.arange(4)]}."
+            "Expected last row of transformation matrix to be exactly [0, 0, 0, 1]."
```
scottshambaugh: This is a little less helpful in tracking down the errant entry, any reason for simplifying it?

scipy/spatial/transform/tests/test_rigid_transform.py
```diff
@@ -194,49 +174,27 @@ def test_from_matrix_array_like():
 
 
 @make_xp_test_case(RigidTransform.from_components)
-def test_from_components(xp):
+@pytest.mark.parametrize("r_ndim", range(1, 4))
+@pytest.mark.parametrize("t_ndim", range(1, 4))
```
scottshambaugh: The combinatorics here are probably overkill if testing is taking a long time, I think should be sufficient to drop this down to a pair of `range(1, 2)`. But should keep `range(1, 4)` on most of the tests to make sure we cover each path without having to think hard about where they branch for the different cases.

scipy/spatial/transform/tests/test_rigid_transform.py
```diff
@@ -601,48 +589,54 @@ def test_as_dual_quat(xp):
...
 @make_xp_test_case(RigidTransform.from_components, RigidTransform.as_components,
                    RigidTransform.from_exp_coords, RigidTransform.as_exp_coords,
                    RigidTransform.from_matrix, RigidTransform.as_matrix,
                    RigidTransform.from_dual_quat, RigidTransform.as_dual_quat)
-def test_from_as_internal_consistency(xp):
+@pytest.mark.parametrize("ndim", range(1, 4))
+def test_from_as_internal_consistency(xp, ndim: int):
     dtype = xpx.default_dtype(xp)
     atol = 1e-12
     n = 1000
```
scottshambaugh: Another way to speed up the tests would be to cut the `n` throughout to e.g. 100. Even 10 should be fine. I think that's fine to do without loss of robustness.

scottshambaugh (approved):
A few suggestions but nothing blocking. Great work as always!

scipy/spatial/transform/_rigid_transform_xp.py
amacati: Since the matrix is N dimensions large, we cannot easily point out a single index now. E.g. if the transformation is (3, 5, 10, 4, 4), and one single row is malformed, it's much harder to specify which one. I'm definitely open to suggestions though, the old message was indeed more helpful.

scipy/spatial/transform/tests/test_rigid_transform.py
amacati: I think it should be at least 3 to test that we can go beyond the current implementation and test that the cython dispatch is handled correctly, but yes, reducing might be sensible.

scipy/spatial/transform/_rigid_transform_xp.py
scottshambaugh: I haven't run this to confirm, but what about
```python
if not lazy and xp.any(~last_row_ok):
    idxs = xp.nonzero(~last_row_ok)
    idx = tuple(int(i[0]) for i in idxs)
    if last_row_ok.shape == ():
        vals = matrix[3, :]
        idx = ()
    else:
        vals = matrix[idx + (3, ...)]
    raise ValueError(f"Expected last row of transformation matrix {idx} to be "
                     f"exactly [0, 0, 0, 1], got {vals}")
```

scottshambaugh (approved):

scipy/spatial/transform/_rigid_transform_xp.py
amacati: Added a variation of this

amacati:
@lucascolley Seems like test failures are unrelated

WarrenWeckesser:
The failure of the job `Linux tests / mypy (py3.11) & dev deps (py3.14), fast, spin (3.14-dev, false) (pull_request)`, which uses the development version of NumPy, shows the fallout from the recently merged NumPy change https://github.com/numpy/numpy/pull/29813.  There are a few places in our code where we use the `where` parameter of a ufunc without also giving an explicit `out` parameter.  That has been a footgun for users for a long time, so now the ufuncs require that the `out` parameter be given explicitly. Giving `out=None` is sufficient to silence the warning, but this is probably a good time to verify that our use of `where` in these cases is actually correct. (Not in this PR, of course--I'm just pointing this out here because I think this is the first PR to hit the new NumPy behavior.)

lucascolley (approved):
LGTM, thanks as always Martin, Scott!

PR #19962, opened 2024-01-25
ENH: sparse: speedup csr/csc setdiag by converting to coo

dschult:
#### Reference issue
Fixes gh-19943

#### What does this implement/fix?
Implement setdiag for compressed format that avoids index value setting by converting to coo format, calling setdiag, and then converting back.

The return conversion took two steps -- converting to csr/csc and then updating the data on self. By creating a private convenience function to shortcut the tocsr/tocsc function, we avoid one of those steps. This also helps a few places in `_compressed.__init__` where we convert to coo and back again. We can avoid some of the input checking.

#### Additional information
Some crude benchmarks. New code is 5-125 times faster. All have linear dependence on nnz. The convert-to-coo version has three curves on top of each other with a lower slope.  Below is the picture with just the convert-to-coo curves to show that they do increase, just at a lower slope.

cross_branch

new_code

[Edit -- legend in the second plot should indicate the new branch -- not `main`]

PR #23676, opened 2025-09-25
BUG: Fixes #23644: sparse: csc_array.setdiag() makes indices unsorted

broesler:
#### Reference issue
Fixes #23644,

This PR:

* fixes #23644 by sorting COO indices in `_cs_matrix._setdiag`.
* refactors `mask` -> `is_existing` and `is_new` in `_setdiag`.
* fixes the condition for converting to COO format in `_setdiag`.
* adds tests for the logic branches in `_setdiag`.

This PR checks whether `has_sorted_indices` is set in the COO branch, and sorts the
newly-created diagonal indices accordingly.

It also updates the condition for converting to COO format in `_setdiag`.
The previous condition was:

```python
self.nnz - mask.sum() < self.nnz * 0.001
```

where `mask` was the logical array that was True for existing elements
on the `k`-th diagonal. This condition did not have the intended effect
when there were non-zeros *not* on the diagonal being set. The new
condition is:

```python
N_new < self.nnz * 0.001
```

where `N_new` is the number of non-existing elements on the `k`-th
diagonal.

This PR also refactors the `mask` array to be named explicitly
`is_existing` because it represents the logical indexing of existing
elements. Similarly, `is_new = ~is_existing`.

This PR creates a new mixin class for testing the
`_cs_matrix._setdiag` function branches from gh-19962. 

The existing `test_setdiag_cs[cr]` methods are now encompassed by the
`_CompressedMixin.test_setdiag_noconvert` method. Remove them to avoid
repeating logic and redundant tests.

#### Additional information
Just to note, the `_cs_matrix._setdiag` function, and similarly the `_coo_base._setdiag` function, cannot create duplicate
entries, so we don't actually need to check `has_canonical_format` or run
`sum_duplicates()`. Both `_setdiag` functions allow explicit zeros,
but that is not part of the SciPy definition of "canonical" format. Thus, we can just check `has_sorted_indices` and run
`sort_indices()` if necessary.

### DIFF
--- a/scipy/sparse/_compressed.py
+++ b/scipy/sparse/_compressed.py
@@ -821,6 +821,10 @@ class _cs_matrix(_data_matrix, _minmax_mixin, IndexMixin):
             coo._setdiag(values, k)
             arrays = coo._coo_to_compressed(self._swap)
             self.indptr, self.indices, self.data, _ = arrays
+            # Sort the indices (like in _insert_many)
+            if self.has_sorted_indices:
+                self.has_sorted_indices = False  # force a sort
+                self.sort_indices()
 
     def _prepare_indices(self, i, j):
         M, N = self._swap(self._shape_as_2d)
--- a/scipy/sparse/_compressed.py
+++ b/scipy/sparse/_compressed.py
@@ -817,12 +817,13 @@ class _cs_matrix(_data_matrix, _minmax_mixin, IndexMixin):
             self._insert_many(i[is_new], j[is_new], x[is_new])
         else:
             # convert to coo for _set_diag
+            do_sort = self.has_sorted_indices
             coo = self.tocoo()
             coo._setdiag(values, k)
             arrays = coo._coo_to_compressed(self._swap)
             self.indptr, self.indices, self.data, _ = arrays
             # Sort the indices (like in _insert_many)
-            if self.has_sorted_indices:
+            if do_sort:
                 self.has_sorted_indices = False  # force a sort
                 self.sort_indices()
 
--- a/scipy/sparse/_compressed.py
+++ b/scipy/sparse/_compressed.py
@@ -813,7 +813,8 @@ class _cs_matrix(_data_matrix, _minmax_mixin, IndexMixin):
             # replace existing entries
             self.data[offsets[is_existing]] = x[is_existing]
             # create new entries
-            is_new = ~is_existing
+            is_new = np.logical_not(is_existing, out=is_existing)
+            del is_existing
             self._insert_many(i[is_new], j[is_new], x[is_new])
         else:
             # convert to coo for _set_diag
--- a/scipy/sparse/_compressed.py
+++ b/scipy/sparse/_compressed.py
@@ -804,17 +804,17 @@ class _cs_matrix(_data_matrix, _minmax_mixin, IndexMixin):
             self.data[offsets] = x
             return
 
-        mask = (offsets >= 0)
+        is_existing = (offsets >= 0)
+        is_new = ~is_existing
+        N_new = is_new.sum()
+
         # Boundary between csc and convert to coo
         # The value 0.001 is justified in gh-19962#issuecomment-1920499678
-        if self.nnz - mask.sum() < self.nnz * 0.001:
+        if N_new < self.nnz * 0.001:
             # replace existing entries
-            self.data[offsets[mask]] = x[mask]
+            self.data[offsets[is_existing]] = x[is_existing]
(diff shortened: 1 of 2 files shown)


scipy/sparse/_compressed.py
```diff
@@ -804,23 +804,27 @@ def _setdiag(self, values, k):
...
-            self._insert_many(i, j, x[mask])
+            self._insert_many(i[is_new], j[is_new], x[is_new])
         else:
             # convert to coo for _set_diag
             coo = self.tocoo()
             coo._setdiag(values, k)
             arrays = coo._coo_to_compressed(self._swap)
             self.indptr, self.indices, self.data, _ = arrays
+            # Sort the indices (like in _insert_many)
+            if self.has_sorted_indices:
```
dschult: Be careful here -- the attribute `self.has_sorted_indices` is a property. If it hasn't been accessed before, this will compute whether the indices are sorted (right now!) rather than indicating that the indices were already sorted before the changes. So, an unsorted input that hasn't got the cache set yet will lead to being sorted here. And a sorted input that hasn't got the cache set yet will check for sorted indices AFTER the setdiag has been done and then sort the indices. I think we want to know the sorted_indices state BEFORE the setdiag has been done to determine whether it needs it after the set is done.

The approach used in `insert_many` is to store the flag value before setting anything. That bool value determines whether a sort is done after setting the values.
```python
    do_sort = self.has_sorted_indices

    ... set stuff ...

    if do_sort:
        self.has_sorted_indices = False  # force a sort
        self.sort_indices()
```

dschult (commented):
Thanks @broesler for this!
I have a comment below about when to check for sorted indices. I'm in favor of matching what `insert_many` does with `do_sort`.

Also there is a bigger issue here about whether the sort should be done after converting back to `cs[rc]`, or whether it should be done before converting back (that is, while still in COO format).  I'm not sure which is better. So maybe that answer will wait until more testing has been done. 

We have to reevaluate the timing to check the formula with this new behavior anyway. So I think its OK to put off checking where the sort should occur to another PR where we re-examine the formula.

scipy/sparse/_compressed.py
broesler: Thanks for pointing out this issue. I moved the `do_sort` line before the `self.tocoo()` etc.

scipy/sparse/_compressed.py
```diff
@@ -804,23 +804,28 @@ def _setdiag(self, values, k):
             self.data[offsets] = x
             return
 
-        mask = (offsets >= 0)
+        is_existing = (offsets >= 0)
+        is_new = ~is_existing
+        N_new = is_new.sum()
```
perimosocordiae:
```suggestion
        N_new = len(is_existing) - np.count_nonzero(is_existing)
```

perimosocordiae: This way we can avoid allocating the full `is_new` array until we're inside the conditional where we know we'll need it.

perimosocordiae (changes requested):

scipy/sparse/_compressed.py
```diff
@@ -804,23 +804,28 @@ def _setdiag(self, values, k):
...
+        if N_new < self.nnz * 0.001:
             # replace existing entries
-            self.data[offsets[mask]] = x[mask]
+            self.data[offsets[is_existing]] = x[is_existing]
             # create new entries
-            mask = ~mask
-            i = i[mask]
-            j = j[mask]
-            self._insert_many(i, j, x[mask])
+            is_new = ~is_existing
```
perimosocordiae: This is going to mean we store the boolean array twice (once for `is_existing` and another time for `is_new`), which is 2x the RAM cost. I agree that the old "mask" naming wasn't super clear, but it did allow us to save some allocations by reusing the same storage for the inverted array.

To be most efficient, we can do the negation in-place while assigning it to a new name for clarity: `is_new = np.logical_not(is_existing, out=is_existing)`

That will make `is_new` a simple alias for the modified `is_existing` array, but since we don't use `is_existing` anywhere else that should be fine. To be extra safe, we can `del is_existing` afterward.

perimosocordiae (changes requested):

broesler:
I believe I have resolved all the requested changes. Let me know if there is anything else.

dschult (approved):
This looks good to me. I think all of @perimosocordiae suggestions are included as well.

I approve this PR.

perimosocordiae:
Merged! (Test failures were unrelated)

PR #23677, opened 2025-09-26
ENH: sparse: Add nD kron sparse construction

dschult:
This adds `kron` support for nD sparse arrays. And it adds tests.

Returns a COO array that represents the kronecker product of the 2 input arrays. If either is nD the result will be nD. This also adds support for 1D.  This is part of the expanding support of nD sparse arrays, and was requested by CVXPY as an nD tool they would use.

I tried to maintain the special case of 2D and fairly dense using BSR format. Also the 2D case follows the old code-path. 
The remaining code-path constructs a COO representation of the kron and then returns the constructed COO array.

Tests are building on the basic 2D framework with additional cases for 3D and 1D.  The exceptions testing changes too because we don't have exceptions for nD and because for 1D we need to ensure the output is spmatrix unless an input is sparray.


scipy/sparse/_construct.py
```diff
@@ -591,64 +595,69 @@ def kron(A, B, format=None):
...
-            raise ValueError(f"kron requires 2D input arrays. `A` is {A.ndim}D.")
-        output_shape = (A.shape[0]*B.shape[0], A.shape[1]*B.shape[1])
+    # B is 2D and fairly dense, use BSR
+    if (
+        (format is None or format == "bsr") and
+        B.ndim == 2 and 2*B.nnz >= math.prod(B.shape)
+    ):
+        A = coo_sparse(A, copy=True)
+        if A.ndim == 2:
+            A = A.tocsr()
```
perimosocordiae: I'm a little confused here: why do we first copy A into COO format, then immediately convert to CSR?

And if A.ndim isn't 2, then we'll end up copying A again on line 626.

perimosocordiae (changes requested):

scipy/sparse/_construct.py
dschult: Good catch!  
Looks like I didn't think through the `copy` attribute well.  We need a copy for the BSR path. But we don't need a copy for the COO path because we use `np.repeat` on the `coords` and `data`.

I refactored to avoid copy where possible and to avoid double-constructor where possible. One path now involves two constructors:  If `A` does not have an `ndim` attribute and `B` is dense enough to justify the BSR approach, we use `coo_sparse(A)` to get the `ndim` value. If `ndim==2` then we use `csr_sparse(A)`.  This is acceptable because the CSR constructor passes creation to COO anyway when the input has no `ndim` and then converts from the coo to csr anyway. By doing that in 2 steps, we can check ndim along the way and stop in COO land when ndim is not 2.

perimosocordiae (approved):

scipy/sparse/_construct.py
```diff
@@ -550,17 +550,21 @@ def eye(m, n=None, k=0, dtype=float, format=None):
...
-    B : sparse or dense matrix
-        second matrix of the product
+    A : sparse or dense array
+        first array of the product
+    B : sparse or dense array
+        second array of the product
     format : str, optional (default: 'bsr' or 'coo')
         format of the result (e.g. "csr")
-        If None, choose 'bsr' for relatively dense array and 'coo' for others
+        If None, choose 'bsr' for relatively dense 2D arrays and 'coo' for others
```
lucascolley, line 713: if the user provides `format="bsr"` with a `>2`-D array, should we error-out? Or should be document that case if we proceed with COO?

Should we also add test coverage for this either way?

lucascolley (commented):
thanks @dschult, one comment

scipy/sparse/_construct.py
dschult: The "bsr" format only supports 2-D arrays. So no problem with dimensions for 'bsr' format inputs. 

When `format is None`, ndim and sparsity of the array determine the output format. With fairly dense 2D arrays we return 'bsr' format. With fairly sparse 2D arrays (and any non-2D arrays), we output 'coo' format.  Is there a better way to say that here?

We currently don't test our choice of the 'bsr' approach vs 'coo' approach, so yes -- its a good idea to test that. I'll try to add that here.  There's a lot more that could be done to modernize these tests too, but not in this PR. :)

Thanks!

lucascolley: > The "bsr" format only supports 2-D arrays. So no problem with dimensions for 'bsr' format inputs.

What I meant was, supposing `x` is a 3-D dense array, `sparse.kron(x, format='bsr')`.

dschult: Ah... sorry for misunderstanding...

We cover the errors in our `tobsr` suite of methods. In the example of `'bsr'` we would raise a ValueError with a confusing message: "Cannot convert a 1d sparse array to bsr format" (there are likely other places where we need to update error messages to refer to nD conversion).  That is probably better done is a separate PR though.

lucascolley: where do we hit the `bsr` constructor in that case, though? Since this condition is not met with `B.ndim == 3`

```python
if (
    (format is None or format == "bsr") and
    B.ndim == 2 and 2*B.nnz >= math.prod(B.shape)
):
```

don't we just use the COO path?

Should `format == "bsr"` instead be forcing us to hit `tobsr`, regardless of `B.ndim`?

dschult: The constructor uses the COO path all the way up to the `.asformat(format)` call just before returning.
So, the coo_array is created but then reformatted. The `asformat` method just calls `tobsr()` or similar.

It was confusing to me that the BSR path uses `return bsr_sparse(...)` while all COO paths use `return coo_sparse(...).asformat(format)`.  But I figured out that the BSR path is only used when format is bsr or "None but chosen to be bsr".  So the `asformat` call at the end would have no impact.  The COO path does need the `asformat` call because it does need to convert, e.g. when `format == 'bsr'` and inoput is 3D, or `format=='bsr'` and input is sparse. The COO code path is used and the result is converted to BSR.

lucascolley: Ah thanks, I missed that on line 664!

What do you think about elaborating on line 631? It might be useful to distinguish between "use COO for computation" and "use COO for return".

lucascolley (approved):

PR #23679, opened 2025-09-26
MAINT/STY: `spatial._kdtree`: Fix type of default value of eps parameter in query methods

r-menezes:
Change type of default eps value from 0 (int) to 0. (float)

This small change would prevent static type checkers like Pyright from incorrectly inferring that eps is an integer.
[skip ci]

### DIFF
--- a/scipy/spatial/_kdtree.py
+++ b/scipy/spatial/_kdtree.py
@@ -361,7 +361,7 @@ class KDTree(cKDTree):
                          balanced_tree, boxsize)
 
     def query(
-            self, x, k=1, eps=0, p=2, distance_upper_bound=np.inf, workers=1):
+            self, x, k=1, eps=0., p=2, distance_upper_bound=np.inf, workers=1):
         r"""Query the kd-tree for nearest neighbors.
 
         Parameters
@@ -477,7 +477,7 @@ class KDTree(cKDTree):
             i = np.intp(i)
         return d, i
 
-    def query_ball_point(self, x, r, p=2., eps=0, workers=1,
+    def query_ball_point(self, x, r, p=2., eps=0., workers=1,
                          return_sorted=None, return_length=False):
         """Find all points within distance r of point(s) x.
 
@@ -554,7 +554,7 @@ class KDTree(cKDTree):
         return super().query_ball_point(
             x, r, p, eps, workers, return_sorted, return_length)
 
-    def query_ball_tree(self, other, r, p=2., eps=0):
+    def query_ball_tree(self, other, r, p=2., eps=0.):
         """
         Find all pairs of points between `self` and `other` whose distance is
         at most r.
@@ -605,7 +605,7 @@ class KDTree(cKDTree):
         """
         return super().query_ball_tree(other, r, p, eps)
 
-    def query_pairs(self, r, p=2., eps=0, output_type='set'):
+    def query_pairs(self, r, p=2., eps=0., output_type='set'):
         """Find all pairs of points in `self` whose distance is at most r.
 
         Parameters


scipy/spatial/_kdtree.py
```diff
@@ -361,7 +361,7 @@ def __init__(self, data, leafsize=10, compact_nodes=True, copy_data=False,
                          balanced_tree, boxsize)
 
     def query(
-            self, x, k=1, eps=0, p=2, distance_upper_bound=np.inf, workers=1):
+            self, x, k=1, eps=0., p=2, distance_upper_bound=np.inf, workers=1):
```
tylerjereddy, line 364: https://github.com/scipy/scipy-stubs has `query` defined with `eps: onp.ToFloat = ...` already.

scipy/spatial/_kdtree.py
```diff
@@ -477,7 +477,7 @@ def query(
             i = np.intp(i)
         return d, i
 
-    def query_ball_point(self, x, r, p=2., eps=0, workers=1,
+    def query_ball_point(self, x, r, p=2., eps=0., workers=1,
```
tylerjereddy, line 480: `scipy-stubs` has `query_ball_point` defined with `eps: onp.ToFloat = ...` in several overloads

scipy/spatial/_kdtree.py
```diff
@@ -554,7 +554,7 @@ def query_ball_point(self, x, r, p=2., eps=0, workers=1,
         return super().query_ball_point(
             x, r, p, eps, workers, return_sorted, return_length)
 
-    def query_ball_tree(self, other, r, p=2., eps=0):
+    def query_ball_tree(self, other, r, p=2., eps=0.):
```
tylerjereddy, line 557: For `query_ball_tree`, `scipy-stubs` has an informative comment on top of specifying the float type already:

> eps: onp.ToFloat = ...,  # defaults to `0.0`, but is overridden in `KDTree` with `0` as default

scipy/spatial/_kdtree.py
```diff
@@ -605,7 +605,7 @@ def query_ball_tree(self, other, r, p=2., eps=0):
         """
         return super().query_ball_tree(other, r, p, eps)
 
-    def query_pairs(self, r, p=2., eps=0, output_type='set'):
+    def query_pairs(self, r, p=2., eps=0., output_type='set'):
```
tylerjereddy, line 608: in `scipy-stubs`, `query_pairs` has two overloads with `eps: onp.ToFloat = 0`

tylerjereddy (commented):
@jorenham can you weigh in here--I went ahead and grepped the stubs repo for the pertinent function signatures, and it seems to me that you may already have this covered over there? Is the use case described here compelling, or can we generally just indicate that the stubs repo should resolve the scenario described in the OC?

scipy/spatial/_kdtree.py
jorenham: for reference: https://github.com/scipy/scipy-stubs/blob/d897b4a4d15e7c0e9505cd75d1a6bbaab31e9885/scipy-stubs/spatial/_ckdtree.pyi#L251

r-menezes:
TBH, I think this is so minor that should not warrant a PR, but I thought that it would be better to open the PR than an issue. As I understood, the solution is already implemented in the stubs, right?

jorenham:
> This small change would prevent static type checkers like Pyright from incorrectly inferring that eps is an integer.

Like @tylerjereddy also noted, this problematic type inference will only occur if [scipy-stubs](https://github.com/scipy/scipy-stubs) isn't installed.

> Is the use case described here compelling, or can we generally just indicate that the stubs repo should resolve the scenario described in the OC?

The problem this aims to address can indeed also be resolved by installing scipy-stubs. But even so, this makes things a bit more consistent. And most importantly, this would allow me to get rid of that comment you mentioned ([src](https://github.com/scipy/scipy-stubs/blob/d897b4a4d15e7c0e9505cd75d1a6bbaab31e9885/scipy-stubs/spatial/_ckdtree.pyi#L251)) 😛.
So I think it's pretty compelling.

tylerjereddy:
Alright, sounds like Joren is +1, and this is quite a minor patch. CI failure is obviously not related to the diff. In it goes.

tylerjereddy:
thanks both

PR #22193, opened 2024-12-26
DEP: linalg.solve_toeplitz/matmul_toeplitz: warn on n-D `c`, `r`

mdhaber:
#### Reference issue
Toward gh-21466
gh-21446

#### What does this implement/fix?
`linalg.toeplitz`, `linalg.solve_toeplitz`, `linalg.matmul_toeplitz` document that `c` and `r` with dimensionality greater than 1 will be `ravel`ed.

gh-21446 added a warning that`toeplitz` would no longer `ravel` in 1.17; rather, multidimensional arrays would be treated as batches of 1-D arrays.

This adds similar warnings to `linalg.solve_toeplitz` and `linalg.matmul_toeplitz`. It also removes the documentation that n-D arrays will be raveled; it is still implicit in the admonition, and we don't really want to advertise it for new uses.

#### Additional information
@tylerjereddy any chance we can get this into 1.15 from the perspective that `toeplitz`, `solve_toeplitz`, `matmul_toeplitz` are a tuple and _should_ have been treated together? This also fixes typos in the existing warnings.

If not, I propose that we still document that this behavior will be changed in 1.17. Our [public deprecation policy](https://docs.scipy.org/doc/scipy/reference/#api-definition) is
> a deprecation warning will be raised for one SciPy release before the change is made.

and in the [Core Developer Guide](https://scipy.github.io/devdocs/dev/core-dev/index.html#deprecations), it says the change can be implemented 6 months after the release that includes the warning (and historically, the winter release tends to be >6 months after the summer release). I know we usually give two releases, but it would be nice to finish these up at the same time, and I expect these will be very low impact.

PR #21446, opened 2024-08-24
ENH: linalg: enable N-D batch support in special matrix functions

mdhaber:
#### Reference issue
https://github.com/scipy/scipy/pull/16090#issuecomment-2308420650

#### What does this implement/fix?
Begins to add N-dimensional batch  support to the special matrix functions. This PR adds documentation, tests, and a simple implementation for API consistency. More performant implementations can come in follow-up PRs.

#### Additional information
~~gh-16090 will add support for `companion`.~~ Done!
gh-21419 will add support for `circulant`
~~A follow-up will add support for `leslie` and `toeplitz`. These accept two arguments of different shapes. This is easy to handle with existing code; I'll just need move the machinery from `scipy.stats.axis_nan_policy` to `_lib`.~~ I decided to leave that code where it is and just import it. The import happens when the function is called for the first time to avoid import cycles. This is temporary and an immediate follow-up PR can either move the machinery to `_lib` to avoid circular imports or eliminate the need for the import by adding batch support directly.

~~I left out `toeplitz` because it documents that it ravels arrays of any number of dimensions, so we can't add batch support until after a cycle of warnings for >1 dimensional input. **Which of the following is preferred?**~~
~~1. Emit a `FutureWarning` that the behavior will change in SciPy 1.17; to ensure consistent behavior, the user must `ravel` input before passing it into `toeplitz`.~~
~~2. Emit a `DeprecationWarning` about support for >1 dimensional input. Raise an error in SciPy 1.17, and add N-D batch support after that.~~
`toeplitz` gets a `FutureWarning` because it documents that it ravels N-D input. In SciPy 1.17 we can add N-D batch support as we do for `leslie`.

I went ahead adding the code inside the function rather than using a decorator because I was going to have to edit the documentation of each function anyway. If we could leave the documentation alone and maybe just add a standard note about batch support, I might do all that with a decorator, instead.

Here is a wrapper we can use for most other `linalg` functions:
<details>

```python3
from functools import wraps
import numpy as np
from scipy import linalg

def _apply_over_batch(*argdefs, result_packer=None):
    names, ndims = list(zip(*argdefs))
    n_arrays = len(names)
    result_packer = result_packer if result_packer is not None else lambda res: res[0]

    def wrapper(f):
        @wraps(f)
        def wrapped(*args, **kwargs):
            args = list(args)

            # Ensure all arrays in `arrays`, other arguments in `other_args`/`kwargs`
            arrays, other_args = args[:n_arrays], args[n_arrays:]
[...]

PR #23680, opened 2025-09-26
MAINT: linalg.hankel: warn against changing behavior for multidimensional arrays

mdhaber:
#### Reference issue
Toward gh-21466
Addendum to gh-22193

Like `linalg.toeplitz`, `linalg.solve_toeplitz`, and `linalg.matmul_toeplitz` did, `linalg.hankel` currently documents that `c` and `r` with dimensionality greater than 1 will be `ravel`ed.

gh-21446 and gh-22193 added warnings that the `toeplitz` functions would no longer `ravel` in 1.17; rather, multidimensional arrays would be treated as batches of 1-D arrays.

This adds a similar warning to `linalg.hankel`. It also removes the documentation that n-D arrays will be raveled; it is still implicit in the admonition, and we don't really want to advertise it for new uses.

#### Additional information
Discussion post in https://discuss.scientific-python.org/t/maint-linalg-hankel-warn-against-changing-behavior-for-multidimensional-arrays/2129

### DIFF
--- a/scipy/linalg/_special_matrices.py
+++ b/scipy/linalg/_special_matrices.py
@@ -1,4 +1,5 @@
 import math
+import warnings
 
 import numpy as np
 from numpy.lib.stride_tricks import as_strided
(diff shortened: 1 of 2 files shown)


j-bowhay (approved):
Thanks for catching this!

PR #23681, opened 2025-09-26
BUG: linalg:solve...are: fix bug when optional array is skipped

mdhaber:
#### Reference issue
Closes gh-23336

gh-23336 reported that `linalg.solve_continuous_are` failed when the optional array argument `e` was skipped. This fixes the bug in `solve_continuous_are` and `solve_discrete_are` removing the decorator from the public function. Then, when all array arguments are available (possibly with default value), a private function with the decorator is used to produce the result.

#### Additional information
I think these are the only batched `linalg` functions that allow an array argument to be skipped. Please check me on that.

### DIFF
--- a/scipy/linalg/tests/test_solvers.py
+++ b/scipy/linalg/tests/test_solvers.py
@@ -577,6 +577,22 @@ class TestSolveDiscreteAre:
         assert_raises(LinAlgError, solve_continuous_are, A, B, Q, R)
 
 
+class TestSolveCommonAre:
+    @pytest.mark.parametrize("solver", [solve_continuous_are, solve_discrete_are])
+    def test_with_skipped_array_argument_gh23336(self, solver):
+        # gh-23336 reported a failure when optional argument `e` was skipped
+        A = np.array([[-0.9, 0.25], [0, -1.1]])
+        B = np.array([[0.23], [0.45]])
+        Q = np.eye(2)
+        R = np.atleast_2d(0.45)
+        E = np.eye(2)
+        S = np.array([[0.1], [0.2]])
+
+        res = solver(A, B, Q, R, s=S)
+        ref = solver(A, B, Q, R, E, S)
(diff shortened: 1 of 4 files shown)


scipy/linalg/tests/test_solvers.py
```diff
@@ -577,6 +577,22 @@ def test_infeasible(self):
...
+        A = np.array([[-0.9, 0.25], [0, -1.1]])
+        B = np.array([[0.23], [0.45]])
+        Q = np.eye(2)
+        R = np.atleast_2d(0.45)
+        E = np.eye(2)
+        S = np.array([[0.1], [0.2]])
+
+        res = solver(A, B, Q, R, s=S)
+        ref = solver(A, B, Q, R, E, S)
+        np.testing.assert_allclose(res, ref)
```
tylerjereddy, line 593: This test does indeed fail when the source patch is reversed, and in a manner consistent with the original ticket, with `ValueError: Matrix e should be square.` So, that's a good sign. It is also good that no old test cases were modified, only new cases were added.

So, mechanically it looks pretty good to me. The CI failure is just https://github.com/scipy/scipy/issues/23689. That said, I've not been super active in the batching work apart from complaining about error handling things, so may make sense for someone else to give it a quick look.

Matt--how severe is this? Does it justify, on its own, another bug fix release for example? (assuming `1.17.0` is not coming out until early January 2026)

mdhaber: Some factors to consider:
- gh-23336 has been waiting patiently for the past 2.5 months without further complaints
- user's get an error (rather than incorrect result)
- the workaround is easy (i.e. pass `e=None`)
- the affected functions are rarely mentioned on SO and in SciPy issues/PRs

To me, these all point toward it not justifying a bug-fix release on its own.

j-bowhay (approved):
Thanks!

PR #23683, opened 2025-09-26
MAINT: sparse: avoid recent free-threading `test_matrix_io.py` failures

dschult:
I've been seeing quite a few `freethread (1)` CI failures with origins in `sparse/tests/test_matrix_io.py` since my PR #23465 was merged. I think the `tmpdir` idiom in the test is not closing the tmpfile used so an exception is raised.

For example: https://github.com/scipy/scipy/actions/runs/18046087738/job/51357077699
shows errors occurring in two tests -- the test changed here, as well as an unrelated test that I think is getting caught in trouble due to parallel garbage collection. The test I've changed in this PR is always involved and the other one changes from failure to failure.  Unfortunately the failures do not even always happen.

I thought about skipping this test for the threadsafe tests, but realized we are no longer closing the file. So I'm just adding back code to close the file.

### DIFF
--- a/scipy/sparse/tests/test_matrix_io.py
+++ b/scipy/sparse/tests/test_matrix_io.py
@@ -75,18 +75,17 @@ def test_sparray_vs_spmatrix():
 
 @pytest.mark.parametrize("value", [0, 1.2])
 @pytest.mark.parametrize("ndim", [1, 2, 3])
-def test_nd_coo_format(ndim, value, tmpdir):
+def test_nd_coo_format(ndim, value):
     A = coo_array([value]).reshape((1,) * ndim)
 
     #save/load array
-    with tmpdir.as_cwd():
-        tmpfile = "f.npz"
-
-        try:
-            save_npz(tmpfile, A)
-            loaded_A = load_npz(tmpfile)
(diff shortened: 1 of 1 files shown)


rgommers (changes requested):
Thanks @dschult. This isn't quite the right fix, since this test will still attempt to write the exact same file multiple times when the body of this test function is run in parallel with `pytest-run-parallel`. The better fix, also from a maintainability point of view, is to remove the `tmpdir` fixture usage and use the exact same pattern with `mkstemp` as all the other tests in this file.

dschult:
I've removed the `tmpdir` fixture in favor of `mkstemp`. That was my first inclination too.

rgommers (approved):
Thanks @dschult, LGTM now and local parallel testing shows that (as expected) all these tests pass reliably with a high number of parallel threads.

nickodell:
BTW, something I learned recently is that you can use a fixture provided by `pytest-run-parallel`, tmp_path. pytest-run-parallel patches this fixture so that it gives a different temp path to each test. https://labs.quansight.org/blog/numpy-thread-safe-test-suite

PR #23685, opened 2025-09-27
MAINT: optimize: fix usage of `nextafter(..., where=)` without `out=`.

rgommers:
This raises a warning now (after https://github.com/numpy/numpy/pull/29813) because it returns uninitialized memory. The usage seems correct, so I left the code as is, although the performance difference of using `where=` over `np.where` is probably negligible.

Fixes CI failures in jobs using numpy nightlies that look like:
```
 FAILED scipy/optimize/tests/test_minimize_constrained.py::test_gh11649 - UserWarning: 'where' used without 'out', expect unitialized memory in output. If this is intentional, use out=None.
```


andyfaff:
This looks good to me.

lucascolley (approved):

PR #23690, opened 2025-09-28
DOC: spatial.transform.Rotation: update docs

amacati:
The recent updates to `Rotation` with broadcasting support are not yet reflected in the docs.

#### Reference issue
- #22500
- #23391
- #23467

Updates the docs to include the broadcasting capabilities. This update is quite minimal, and I feel like we could advertise the new capabilities more, but this PR is only meant to bring the docs up-to-date.

### DIFF
--- a/scipy/spatial/transform/_rotation.py
+++ b/scipy/spatial/transform/_rotation.py
@@ -81,9 +81,9 @@ class Rotation:
     - Rotation Indexing
 
     A `Rotation` instance can contain a single rotation transform or rotations of
-    multiple leading dimensions. E.g., it is possible to have a tensor of (N, M, K)
-    rotations. When applied to other rotations or vectors, standard broadcasting rules
-    apply.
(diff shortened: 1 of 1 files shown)


scipy/spatial/transform/_rotation.py
```diff
@@ -80,8 +80,13 @@ class Rotation:
     - Rotation Inversion
     - Rotation Indexing
 
-    Indexing within a rotation is supported since multiple rotation transforms
-    can be stored within a single `Rotation` instance.
+    A `Rotation` instance can contain a single rotation transform or rotations of
+    multiple leading dimensions. E.g., it is possible to have a tensor of (N, M, K)
```
scottshambaugh: IMO we should call them N-dimensional arrays rather than tensors throughout, to parallel numpy arrays.

scottshambaugh (approved):

amacati:
@lucascolley Any opinions?

lucascolley (approved):
thanks both!

PR #23691, opened 2025-09-28
DOC: stats.wasserstein_distance: add example from reference

mdhaber:
#### Reference issue
Closes gh-23252

gh-23252 reported an apparent discrepancy between the output of `stats.wasserstein_distance` and the example in a reference. This PR adds the example from the reference to the `stats.wasserstein_distance` documentation to clarify the relationship between the function's output and typical earth mover's distance.

#### Additional information
@rkern LMK what you think we should do with the existing examples, if anything.

@twstephens26 does this help clarify the behavior of `wasserstein_distance`?

### DIFF
--- a/scipy/stats/_stats_py.py
+++ b/scipy/stats/_stats_py.py
@@ -9662,6 +9662,27 @@ def wasserstein_distance(u_values, v_values, u_weights=None, v_weights=None):
     ...                      [1.4, 0.9, 3.1, 7.2], [3.2, 3.5])
     4.0781331438047861
 
+    From the example in [2]_, we take the `u_values` and `v_values` to be the values
+    :math:`i` associated with the piles :math:`P_i` and :math:`Q_i`. The `u_weights`
+    and `v_weights` are the "shovelsful of dirt" in each pile:
+
+    >>> u_values = v_values = np.asarray([1, 2, 3, 4])
+    >>> u_weights = np.asarray([3, 2, 1, 4])
+    >>> v_weights = np.asarray([1, 2, 4, 3])
+    >>> wasserstein_distance(u_values, v_values,
+    ...                      u_weights=u_weights, v_weights=v_weights)
+    np.float64(0.5)
+
+    In our context of probability distributions, however, the weights have been
+    normalized and treated as probability masses associated with each value. This
+    accounts for the factor of ten between the output of `wasserstein_distance`,
+    ``0.5``, and the result in the reference, :math:`5`. The Wasserstein distance does
+    depend on the scales of the *values* of the distributions, though:
+
+    >>> wasserstein_distance(100*u_values, 100*v_values,
+    ...                      u_weights=u_weights, v_weights=v_weights)
+    np.float64(50.0)
+
     """
     return _cdf_distance(1, u_values, v_values, u_weights, v_weights)
 


rkern:
I wouldn't recapitulate the confusing example from the link; using the same set of numbers for both weights and locations is just begging for trouble. I'd find a better link (or even just removing the link) than try to explain the link's example. The explanatory text is good, but they're probably better applied to the numbers from the existing examples rather than referring to that link.

mdhaber:
I tried to find an example before starting; but I didn't find one that I liked. If you find an example, I'll make the adjustment.

rkern:
I thought the existing examples were perfectly fine, as far as the code went. They just didn't have the explanatory text that you've added. Your explanatory text doesn't need any particular numbers; it could have been wrapped around the existing examples just fine.

mdhaber:
> Your explanatory text doesn't need any particular numbers; it could have been wrapped around the existing examples just fine.

Hmm... I don't that makes as much sense:

```
    From the example in [2]_, we take the `u_values` and `v_values` to be the values
    :math:`i` associated with the piles :math:`P_i` and :math:`Q_i`. The `u_weights`
    and `v_weights` are the "shovelsful of dirt" in each pile:

    >>> from scipy.stats import wasserstein_distance
    >>> wasserstein_distance([0, 1, 3], [5, 6, 8])
    5.0
    >>> wasserstein_distance([0, 1], [0, 1], [3, 1], [2, 2])
    0.25
    >>> wasserstein_distance([3.4, 3.9, 7.5, 7.8], [4.5, 1.4],
    ...                      [1.4, 0.9, 3.1, 7.2], [3.2, 3.5])
    4.0781331438047861

    In our context of probability distributions, however, the weights have been
    normalized and treated as probability masses associated with each value. This
    accounts for the factor of ten between the output of `wasserstein_distance`,
    ``0.5``, and the result in the reference, :math:`5`. The Wasserstein distance does
    depend on the scales of the *values* of the distributions, though.

```

It might work if I took out the words and added new ones. Maybe later.

rkern:
I mean, I'm definitely saying that you should remove the reference to `[2]`, even if you do use their numbers. They failed to make a good example, and it's not our job to fix that up for them. If `[2]` is confusing people, fixing it by deleting the link.

I said "the existing examples were perfectly fine, as far as the code went" but I actually didn't mean the "code" verbatim, just the numbers, as I mentioned earlier. The value that your example brought was labeling the inputs so it was really clear about the semantics of each input.

mdhaber:
> If [2] is confusing people, fixing it by deleting the link.

I see. I must have misunderstood "and I think the link is fine" (https://github.com/scipy/scipy/issues/23252#issuecomment-3013728449). Anyway, that's why tried to explain it rather than removing it.

I could certainly improve this example section; I just don't really have a vision for massaging _this PR_ to work without [2]_.[^1]

[^1]: As unfortunate as the numbers in [2] were, it was doing some useful things. It made `wasserstein_distance` intuitive by relating it to EM distance and explaining the essential ideas. Without it, I wouldn't be able to talk about "piles" or "shovelsful of dirt" anymore, so I'd have to think about what those would become. I'd need to explain three examples instead of just one. I'd have no expected values to compare the output of `wasserstein_distance` against. So I might as well start fresh later.

rkern:
My full position is that links that don't follow our conventions are fine to include as references if they're otherwise helpful even if sometimes someone gets confused when trying to apply their understanding of that text to our API. Adding in additional explanations in our docstrings to correlate our terminology/conventions to those used in our references are also fine if those references are using common conventions that people will likely see elsewhere (talking about "piles", etc., here or the varying `theta/phi` conventions in polar coordinates elsewhere).

Referring to specific details all the way down to the numbers in the examples in the references in order to clarify their confusing choices is unnecessary, IMO. Not least, links can get edited. We want our docstrings to stand alone as much as possible. So *if* the confusion stemming from our link's example is too onerous to bear, the fix is to remove the link (but again, I don't think it's too onerous to bear).

It did highlight that we were lacking good explanatory text around our API usage examples to make clear which numbers were serving which purpose. That would have helped the reader connect their understanding from the link's text to our API, even when our examples don't use the same numbers as the reference link. Use the "piles" terminology of `[2]` in that text, by all means, just don't try to explain their specific example; it's not our job, and the numbers in the example are not valuable to retain to help the reader make the connections. At minimum, change the `u_values/v_values` to something distinct from the weights, like `[11, 12, 13, 14]`, so that we do not recapitulate our reference's mistakes.

PR #23692, opened 2025-09-28
MAINT: stats.pmean: enforce finite `p` requirement

mdhaber:
#### Reference issue
Closes gh-23111

I started to correct `scipy.stats.pmean` when `p` is infinite by adding special cases, but there are a lot of degenerate cases to consider/test that don't just fall out of the obvious implementation. Beyond zero weights (mentioned in the issue), one would need to consider _all_ zero weights, infinite weights, weight and value arrays with incompatible shapes, etc. So I decided to just document the limitation that `p` must be finite and raise an error for infinite `p`.

### DIFF
--- a/scipy/stats/_stats_py.py
+++ b/scipy/stats/_stats_py.py
@@ -457,6 +457,10 @@ def pmean(a, p, *, axis=0, dtype=None, weights=None):
         return gmean(a, axis=axis, dtype=dtype, weights=weights)
     elif math.isinf(p):
         fun = xp.max if p > 0 else xp.min
+        if weights is not None:
+            zero_weight_fun = xp.min if p > 0 else xp.max
+            a_zero_weight = zero_weight_fun(a, axis=None)
+            a = xpx.at(a, weights==0).set(a_zero_weight)
         return fun(a, axis=axis)
 
     with np.errstate(divide='ignore', invalid='ignore'):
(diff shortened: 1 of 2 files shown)


nickodell:
@mdhaber 

I have a meta / process question. This PR is making a behavioral change, right? It seems like it should be tagged as either BUG (in the sense that it fixes incorrect output) or DEP (in the sense that this is questionable input that used to not raise an exception, and now it does.) If a PR is both a doc change and a bugfix, which takes priority?

mdhaber:
From [Development workflow](https://docs.scipy.org/doc/scipy/dev/contributor/development_workflow.html):
> BUG: bug fix
DEP: deprecate something, or remove a deprecated object
...
DOC: documentation
...
MAINT: maintenance commit (refactoring, typos, etc.)

Definitely not DEP, since there's no deprecation happening, and none is needed: I don't think we're changing something that worked, was supposed to work (presumably), or really needs to work (my opinion).

MAINT is also probably wrong according to the definition above, but I think you'll find in the commit records that - for better or for worse - it has taken on a much broader meaning.

I tend not to label many things as BUG. Maybe I should. I won't express opinions about that here. 

If it were a BUG, that would probably take precedence over DOC. I suppose my thinking when I wrote DOC was that I was just documenting the developer's (apparent) intent for the function's scope to be limited to finite `p`[^1]. I suppose I took a liberal interpretation of the error message as a form of documentation.

But these are just my opinions. You're welcome to collect other opinions or change the labels and commit message.

[^1]: I think finite `p` is a reasonable limitation. We could put in a fair amount of effort to make this work for infinite `p`, and it might be fun to do exactly the right thing with all possible inputs, but I think that time would serve users better if it were invested elsewhere. I don't think inifinite `p` adds much utility, and explicit use of `max` and `min` probably should be preferred in those cases anyway - maybe even if the user needs to branch to use them.

nickodell (approved):

j-bowhay (approved):

PR #23693, opened 2025-09-28
ENH: stats.mode: add array API support

mdhaber:
#### Reference issue
Toward gh-20544
Follow-up of gh-22857

Adds array API support to `scipy.stats.mode`.

#### Additional information
~~Not sure what's going on with `rankdata` with torch on GPU. It is failing most of `test_cases` that have `method='ordinal'`. I'm seeing one of those failures in CI with torch on CPU, but not locally.~~ *Torch doesn't default to stable argsort; see array-api-compat#354.*
Had to skip tests with JAX because the arrays are lazy so `_axis_nan_policy` decorator can't preprocess the inputs.

### DIFF
--- a/scipy/stats/_stats_py.py
+++ b/scipy/stats/_stats_py.py
@@ -465,15 +465,18 @@ def _mode_result(mode, count):
     # When a slice is empty, `_axis_nan_policy` automatically produces
     # NaN for `mode` and `count`. This is a reasonable convention for `mode`,
     # but `count` should not be NaN; it should be zero.
-    i = np.isnan(count)
+    xp = array_namespace(mode, count)
+    i = xp.isnan(count)
     if i.shape == ():
-        count = np.asarray(0, dtype=count.dtype)[()] if i else count
+        count = xp.asarray(0, dtype=count.dtype)[()] if i else count
     else:
-        count[i] = 0
+        count = xpx.at(count)[i].set(0)
     return ModeResult(mode, count)
 
 
-@xp_capabilities(np_only=True)
+@xp_capabilities(skip_backends=[('dask.array', "can't compute chunk size"),
+                                ('cupy', "data-apis/array-api-compat#312")],
+                 jax_jit=False)
 @_axis_nan_policy_factory(_mode_result, override={'nan_propagation': False})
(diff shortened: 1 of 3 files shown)


scipy/stats/_stats_py.py
```diff
@@ -465,15 +465,18 @@ def _mode_result(mode, count):
...
+        count = xp.asarray(0, dtype=count.dtype)[()] if i else count
     else:
-        count[i] = 0
+        count = xpx.at(count)[i].set(0)
     return ModeResult(mode, count)
 
 
-@xp_capabilities(np_only=True)
+@xp_capabilities(skip_backends=[('dask.array', "can't compute chunk size"),
+                                ('jax.numpy', "relies on _axis_nan_policy"),
```
mdhaber, line 478: Previously, I had `jax_jit=False`, so I expected the arrays to be eager. However, it seems they were not, because it was triggering this code:

https://github.com/scipy/scipy/blob/434fc553a403eea11357bfeb0febb66b5175271e/scipy/stats/_axis_nan_policy.py#L439-L444

(The way that `temp` is determined in general is not ideal, since it basically selects only the first (zeroth) argument of the function to check for laziness. But that is exactly right for `mode`, so I don't think that's the problem.)

Why is the array lazy? JAX arrays should be fine when they're eager - they were working back in the original incarnation of gh-22857 before we made them lazy by default.

lucascolley: See https://data-apis.org/array-api-compat/helper-functions.html#array_api_compat.is_lazy_array. IIRC the advice from JAX is not to branch based on JIT. Regardless of JIT, JAX arrays are always treated as lazy.

mdhaber: I see. Well, then JAX arrays will always miss out on some functionality, JIT or not.

Abhijais4896 (commented):
def test_large_int(self, xp):
        data = xp.asarray([2**60, 2**60+1], dtype=xp.int64)
        r = rankdata(data)
        xp_assert_equal(r, xp.asarray([1.0, 2.0], dtype=self.desired_dtype(xp=xp)))

j-bowhay (approved):
LGTM!

PR #23694, opened 2025-09-29
DOC: Correct minor formatting issues in `scipy.sparse.linalg`

star1327p:

Correct minor formatting issues in `scipy.sparse.linalg`


tylerjereddy (approved):
The `linsolve` changes are clearly correct. The `onenormest` links I confirmed are fixed in the new docs render here and broken on the devocs.

So, in it goes, thanks.

Issue #6499, opened 2016-08-19
scipy-0.18.0 test failures on NetBSD


0-wiz-0:
I've just finished running the self-tests for scipy-0.18.0 on NetBSD-7.99.35/amd64 with pyhon-3.5.2, numpy-1.11.1, cython-0.24, fftw-3.3.4.
There are a couple of errors:

```
======================================================================
ERROR: Failure: ImportError (.../scipy-0.18.0/build/testenv/lib/python3.5/site-packages/scipy/integrate/_test_odeint_banded.so: Undefined symbol "banded5x5_jac_" (symnum = 34))
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/pkg/lib/python3.5/site-packages/nose/failure.py", line 39, in runTest
    raise self.exc_val.with_traceback(self.tb)
  File "/usr/pkg/lib/python3.5/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/usr/pkg/lib/python3.5/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/usr/pkg/lib/python3.5/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/usr/pkg/lib/python3.5/imp.py", line 234, in load_module
    return load_source(name, filename, file)
  File "/usr/pkg/lib/python3.5/imp.py", line 172, in load_source
    module = _load(spec)
  File "<frozen importlib._bootstrap>", line 693, in _load
  File "<frozen importlib._bootstrap>", line 673, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 665, in exec_module
  File "<frozen importlib._bootstrap>", line 222, in _call_with_frames_removed
  File ".../scipy-0.18.0/build/testenv/lib/python3.5/site-packages/scipy/integrate/tests/test_odeint_jac.py", line 5, in <module>
    import scipy.integrate._test_odeint_banded as banded5x5
ImportError: .../scipy-0.18.0/build/testenv/lib/python3.5/site-packages/scipy/integrate/_test_odeint_banded.so: Undefined symbol "banded5x5_jac_" (symnum = 34)

======================================================================
ERROR: test_check_finite (test_basic.TestDet)
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #23700, opened 2025-09-29
CI: use Pixi for some linux jobs 

lucascolley:
towards gh-21489

Please review the first two commits with "hide whitespace" and the fourth commit separately.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -450,6 +450,7 @@ jobs:
     # (i.e., we're using it as a binary). This isn't a normal setup, but it's
     # sometimes being used that way - see gh-20535 and gh-24406
     name: build with global meson
+    needs: get_commit_message
     if: >
       needs.get_commit_message.outputs.message == 1
       && (github.repository == 'scipy/scipy' || github.repository == '')
@@ -501,10 +502,12 @@ jobs:
 
   #################################################################################
   free-threaded:
+    name: free-threaded${{ matrix.parallel == '1' && ' (pytest-run-parallel)' || '' }}
     needs: get_commit_message
     strategy:
       matrix:
         parallel: ["0", "1"]
+      fail-fast: false
 
     runs-on: ubuntu-latest
     if: >
@@ -515,36 +518,15 @@ jobs:
         with:
           submodules: recursive
           fetch-tags: true
-
-      - uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # v5.5.0
+      
+      - uses: prefix-dev/setup-pixi@ba3bb36eb2066252b2363392b7739741bb777659 # v0.8.1
         with:
-          python-version: "3.14t-dev"
-
-      - name: Install Ubuntu dependencies
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y gfortran
-
-      - name: Install Python dependencies
-        run: |
-          pip install -r requirements/build.txt
-          pip install -r requirements/openblas.txt
-          pip install spin pytest pytest-xdist threadpoolctl pooch hypothesis "click<8.3.0"
-
-      - name: Install pytest-run-parallel
-        if: ${{ matrix.parallel == '1'}}
-        run: |
-          pip install pytest-run-parallel
-          pip uninstall --yes pytest-xdist
-
-      - name: Build SciPy
-        run: |
-          spin build --with-scipy-openblas
+          pixi-version: v0.63.2
+          cache: false
 
       - name: Run tests (full)
         if: ${{ matrix.parallel == '0'}}
-        run: |
-          spin test -j4 -m full --durations=10
+        run: pixi run test-freethreading -j4 -m full --durations=10
 
       - name: Run tests (fast, with pytest-run-parallel)
         if: ${{ matrix.parallel == '1'}}
@@ -576,7 +558,7 @@ jobs:
           #       and it'll be quite slow with pytest-run-parallel. Also skip
           #       tests that won't run in parallel, those are covered by the
           #       parallel == 0 run.
-          spin test -t $TEST_SUBMODULES -- --parallel-threads=4 --skip-thread-unsafe=true
+          pixi run test-parallel-freethreading -t $TEST_SUBMODULES -- --parallel-threads=4 --skip-thread-unsafe=true
 
   #################################################################################
   clang-17-build-only:
@@ -629,30 +611,13 @@ jobs:
         with:
           submodules: recursive
 
-      - name: Setup Python
-        uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # v5.5.0
+      - uses: prefix-dev/setup-pixi@ba3bb36eb2066252b2363392b7739741bb777659 # v0.8.1
         with:
-          python-version: "3.12"
-
-      - name: Install Ubuntu dependencies
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y libopenblas-dev libatlas-base-dev liblapack-dev gfortran libgmp-dev libmpfr-dev libsuitesparse-dev ccache libmpc-dev
-
-      - name: Install Python packages
-        run: |
-          python -m pip install -r requirements/build.txt -r requirements/test.txt
-          # We want to check for test timing only in a single job, on Linux, running the
-          # fast test suite. This is that job. See gh-20806 for previous issues
-          # after running this job on Windows and in multiple jobs.
-          python -m pip install pytest-fail-slow
-
-      - name: Install SciPy
-        run: |
-          python -m pip install . --no-build-isolation
+          pixi-version: v0.63.2
+          cache: false
 
       - name: Test SciPy
         run: |
-          export OMP_NUM_THREADS=2
-          cd ..
-          pytest --pyargs scipy -m 'not slow' --durations=0 --durations-min=0.5 --fail-slow=1.0
+          export OMP_NUM_THREAD=2
+          # various factorial tests give a fatal python error in CI
+          pixi run test-fail-slow -k 'not (test_factorial or test_nan_inputs)'
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -10,7 +10,7 @@ on:
       - maintenance/**
 
 permissions:
-   contents: read  # to fetch code (actions/checkout)
+  contents: read # to fetch code (actions/checkout)
 
 env:
   CCACHE_DIR: "${{ github.workspace }}/.ccache"
@@ -43,106 +43,106 @@ jobs:
             python-version: '3.14-dev'
 
     steps:
-    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-      with:
-        submodules: recursive
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          submodules: recursive
 
-    - name: Setup Python
-      uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # v5.5.0
-      with:
-        python-version: ${{ matrix.python-version }}
-        allow-prereleases: true
+      - name: Setup Python
+        uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # v5.5.0
+        with:
+          python-version: ${{ matrix.python-version }}
+          allow-prereleases: true
 
-    - name: Install Ubuntu dependencies
-      run: |
-        # NOTE: not the same OpenBLAS version as in upstream CI (I'm being lazy here)
-        sudo apt-get update
-        sudo apt-get install -y libopenblas-dev libatlas-base-dev liblapack-dev gfortran libgmp-dev libmpfr-dev libsuitesparse-dev ccache libmpc-dev
+      - name: Install Ubuntu dependencies
+        run: |
+          # NOTE: not the same OpenBLAS version as in upstream CI (I'm being lazy here)
+          sudo apt-get update
+          sudo apt-get install -y libopenblas-dev libatlas-base-dev liblapack-dev gfortran libgmp-dev libmpfr-dev libsuitesparse-dev ccache libmpc-dev
 
-    - name: Install Python packages
-      if: matrix.python-version == '3.12'
-      run: |
-        python -m pip install numpy cython pytest pytest-xdist pytest-timeout pybind11 mpmath gmpy2 pythran ninja meson pooch hypothesis spin "click<8.3.0"
+      - name: Install Python packages
+        if: matrix.python-version == '3.12'
+        run: |
+          python -m pip install numpy cython pytest pytest-xdist pytest-timeout pybind11 mpmath gmpy2 pythran ninja meson pooch hypothesis spin "click<8.3.0"
 
-    - name: Install Python packages from repositories
-      if: matrix.python-version == '3.14-dev' # this run will use python dev versions when available
-      run: |
-        python -m pip install git+https://github.com/numpy/numpy.git
-        python -m pip install ninja cython pytest pybind11 pytest-xdist pytest-timeout spin pooch hypothesis "setuptools<67.3" meson "click<8.3.0"
-        python -m pip install git+https://github.com/serge-sans-paille/pythran.git
-        # Disable Meson master testing until upstream option handling is fixed, see scipy#22534
-        # python -m pip install git+https://github.com/mesonbuild/meson.git
+      - name: Install Python packages from repositories
+        if: matrix.python-version == '3.14-dev' # this run will use python dev versions when available
+        run: |
+          python -m pip install git+https://github.com/numpy/numpy.git
+          python -m pip install ninja cython pytest pybind11 pytest-xdist pytest-timeout spin pooch hypothesis "setuptools<67.3" meson "click<8.3.0"
+          python -m pip install git+https://github.com/serge-sans-paille/pythran.git
+          # Disable Meson master testing until upstream option handling is fixed, see scipy#22534
+          # python -m pip install git+https://github.com/mesonbuild/meson.git
 
-    - name:  Prepare compiler cache
-      id:    prep-ccache
-      shell: bash
-      run: |
-        mkdir -p "${CCACHE_DIR}"
-        echo "dir=$CCACHE_DIR" >> $GITHUB_OUTPUT
-        NOW=$(date -u +"%F-%T")
-        echo "timestamp=${NOW}" >> $GITHUB_OUTPUT
+      - name: Prepare compiler cache
+        id: prep-ccache
+        shell: bash
+        run: |
+          mkdir -p "${CCACHE_DIR}"
+          echo "dir=$CCACHE_DIR" >> $GITHUB_OUTPUT
+          NOW=$(date -u +"%F-%T")
+          echo "timestamp=${NOW}" >> $GITHUB_OUTPUT
 
-    - name: Setup compiler cache
-      uses:  actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
-      id:    cache-ccache
-      # Reference: https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key
-      # NOTE: The caching strategy is modeled in a way that it will always have a unique cache key for each workflow run
-      # (even if the same workflow is run multiple times). The restore keys are not unique and for a partial match, they will
-      # return the most recently created cache entry, according to the GitHub Action Docs.
-      with:
-        path: ${{ steps.prep-ccache.outputs.dir }}
-        # Restores ccache from either a previous build on this branch or on main
-        key:  ${{ github.workflow }}-${{ matrix.python-version }}-ccache-linux-${{ steps.prep-ccache.outputs.timestamp }}
-        # This evaluates to `Linux Tests-3.12-ccache-linux-` which is not unique. As the CI matrix is expanded, this will
-        # need to be updated to be unique so that the cache is not restored from a different job altogether.
-        restore-keys: |
-          ${{ github.workflow }}-${{ matrix.python-version }}-ccache-linux-
+      - name: Setup compiler cache
+        uses:  actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
+        id:    cache-ccache
+        # Reference: https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key
+        # NOTE: The caching strategy is modeled in a way that it will always have a unique cache key for each workflow run
+        # (even if the same workflow is run multiple times). The restore keys are not unique and for a partial match, they will
+        # return the most recently created cache entry, according to the GitHub Action Docs.
+        with:
+          path: ${{ steps.prep-ccache.outputs.dir }}
+          # Restores ccache from either a previous build on this branch or on main
+          key:  ${{ github.workflow }}-${{ matrix.python-version }}-ccache-linux-${{ steps.prep-ccache.outputs.timestamp }}
+          # This evaluates to `Linux Tests-3.12-ccache-linux-` which is not unique. As the CI matrix is expanded, this will
+          # need to be updated to be unique so that the cache is not restored from a different job altogether.
+          restore-keys: |
+            ${{ github.workflow }}-${{ matrix.python-version }}-ccache-linux-
 
-    - name: Setup build and install scipy
-      run: |
-        spin build --release
+      - name: Setup build and install scipy
+        run: |
+          spin build --release
+  
+      - name: Ccache performance
+        shell: bash -l {0}
+        run: ccache -s
 
-    - name: Ccache performance
-      shell: bash -l {0}
-      run: ccache -s
+      - name: Check installed files
+        run: |
+          spin check --installed-files --no-build
+  
+      - name: Check symbol hiding
+        run: |
+          spin check --symbol-hiding --no-build
+  
+      - name: Check usage of install tags
+        run: |
+          rm -r build-install
+          spin build --tags=runtime,python-runtime,devel
+          python tools/check_installation.py build-install --no-tests
+          rm -r build-install
+          spin build --tags=runtime,python-runtime,devel,tests
+          spin check --installed-files --no-build
+  
+      - name: Check xp markers
+        run: |
+          spin check --xp-markers --no-build
+  
+      - name: Check build-internal dependencies
+        run: ninja -C build -t missingdeps
 
-    - name: Check installed files
-      run: |
-        spin check --installed-files --no-build
+      - name: Mypy
+        if: matrix.python-version == '3.12'
+        run: |
+          # Packages that are only needed for their annotations
+          python -m pip install mypy==1.10.0 types-psutil typing_extensions
+          python -m pip install pybind11 sphinx
+  
+          spin mypy
 
-    - name: Check symbol hiding
-      run: |
-        spin check --symbol-hiding --no-build
-
-    - name: Check usage of install tags
-      run: |
-        rm -r build-install
-        spin build --tags=runtime,python-runtime,devel
-        python tools/check_installation.py build-install --no-tests
-        rm -r build-install
-        spin build --tags=runtime,python-runtime,devel,tests
-        spin check --installed-files --no-build
-
-    - name: Check xp markers
-      run: |
-        spin check --xp-markers --no-build
-
-    - name: Check build-internal dependencies
-      run: ninja -C build -t missingdeps
-
-    - name: Mypy
-      if: matrix.python-version == '3.12'
-      run: |
-        # Packages that are only needed for their annotations
-        python -m pip install mypy==1.10.0 types-psutil typing_extensions
-        python -m pip install pybind11 sphinx
-
-        spin mypy
-
-    - name: Test SciPy
-      run: |
-        export OMP_NUM_THREADS=2
-        spin test -j3 -- --durations 10 --timeout=60
+      - name: Test SciPy
+        run: |
+          export OMP_NUM_THREADS=2
+          spin test -j3 -- --durations 10 --timeout=60
 
   #################################################################################
   test_venv_install:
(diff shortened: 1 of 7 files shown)


pixi.toml
```diff
@@ -3,7 +3,7 @@ name = "scipy-ci"
 description = "Environment and task management for SciPy CI"
 authors = ["SciPy Developers <scipy-dev@python.org>"]
 channels = ["https://prefix.dev/conda-forge"]
-platforms = ["linux-64", "osx-arm64", "win-64"]
+platforms = ["linux-64", "osx-arm64", "win-64", "linux-aarch64"]
```
lucascolley, line 6: We could remove `linux-aarch64` from some other features if we want. Everything locks at least with the current setup.

lucascolley:
not sure what's up with the `aarch` job:

```
scipy/special/tests/test_basic.py ...................................... [ 73%]
Fx............................................................F......... [ 73%]
............xxx......................................................... [ 73%]
........................................................................ [ 73%]
........................................................................ [ 73%]
Fatal Python error: Aborted

Current thread 0x0000ff35ef51da60 (most recent call first):
  File "/home/runner/work/scipy/scipy/build-install/usr/lib/python3.13/site-packages/scipy/special/_basic.py", line 2732 in _gamma1p
  File "/home/runner/work/scipy/scipy/build-install/usr/lib/python3.13/site-packages/scipy/special/_basic.py", line 2750 in _factorialx_approx_core
  File "/home/runner/work/scipy/scipy/build-install/usr/lib/python3.13/site-packages/scipy/special/_basic.py", line 2715 in _factorialx_array_approx
  File "/home/runner/work/scipy/scipy/build-install/usr/lib/python3.13/site-packages/scipy/special/_basic.py", line 2921 in _factorialx_wrapper
  File "/home/runner/work/scipy/scipy/build-install/usr/lib/python3.13/site-packages/scipy/special/_basic.py", line 2978 in factorial
  File "/home/runner/work/scipy/scipy/build-install/usr/lib/python3.13/site-packages/scipy/special/tests/test_basic.py", line 2341 in test_factorialx_inf_nan
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/_pytest/python.py", line 157 in pytest_pyfunc_call
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/pluggy/_callers.py", line 121 in _multicall
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/pluggy/_manager.py", line 120 in _hookexec
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/pluggy/_hooks.py", line 512 in __call__
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/_pytest/python.py", line 1671 in runtest
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/_pytest/runner.py", line 178 in pytest_runtest_call
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/pluggy/_callers.py", line 121 in _multicall
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/pluggy/_manager.py", line 120 in _hookexec
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/pluggy/_hooks.py", line 512 in __call__
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/_pytest/runner.py", line 246 in <lambda>
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/_pytest/runner.py", line 344 in from_call
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/_pytest/runner.py", line 245 in call_and_report
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/_pytest/runner.py", line 136 in runtestprotocol
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/_pytest/runner.py", line 117 in pytest_runtest_protocol
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/pluggy/_callers.py", line 121 in _multicall
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/pluggy/_manager.py", line 120 in _hookexec
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/pluggy/_hooks.py", line 512 in __call__
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/_pytest/main.py", line 367 in pytest_runtestloop
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/pluggy/_callers.py", line 121 in _multicall
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/pluggy/_manager.py", line 120 in _hookexec
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/pluggy/_hooks.py", line 512 in __call__
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/_pytest/main.py", line 343 in _main
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/_pytest/main.py", line 289 in wrap_session
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/_pytest/main.py", line 336 in pytest_cmdline_main
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/pluggy/_callers.py", line 121 in _multicall
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/pluggy/_manager.py", line 120 in _hookexec
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/pluggy/_hooks.py", line 512 in __call__
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/_pytest/config/__init__.py", line 175 in main
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/_pytest/config/__init__.py", line 201 in console_main
  File "/home/runner/work/scipy/scipy/.github/workflows/.pixi/envs/test-fail-slow/lib/python3.13/site-packages/pytest/__main__.py", line 9 in <module>
  File "<frozen runpy>", line 88 in _run_code
  File "<frozen runpy>", line 198 in _run_module_as_main

Extension modules: numpy._core._multiarray_umath, numpy.linalg._umath_linalg, scipy._lib._ccallback_c, scipy._lib._fpumode, numpy.random._common, numpy.random.bit_generator, numpy.random._bounded_integers, numpy.random._pcg64, numpy.random._mt19937, numpy.random._generator, numpy.random._philox, numpy.random._sfc64, numpy.random.mtrand, _cyutility, scipy._cyutility, scipy.sparse._sparsetools, _csparsetools, scipy.sparse._csparsetools, scipy.spatial._ckdtree, scipy._lib.messagestream, scipy.linalg._fblas, scipy.linalg._flapack, scipy.linalg.cython_lapack, scipy.linalg._cythonized_array_utils, scipy.linalg._solve_toeplitz, scipy.linalg._batched_linalg, scipy.linalg._decomp_lu_cython, scipy.linalg._matfuncs_schur_sqrtm, scipy.linalg._matfuncs_expm, scipy.linalg._linalg_pythran, scipy.linalg.cython_blas, scipy.linalg._decomp_update, scipy.spatial._qhull, scipy.spatial._voronoi, scipy.special._ufuncs_cxx, scipy.special._ellip_harm_2, scipy.special._special_ufuncs, scipy.special._gufuncs, scipy.special._ufuncs, scipy.special._specfun, scipy.special._comb, scipy.spatial._hausdorff, scipy.spatial._distance_wrap, scipy.spatial.transform._rotation_cy, scipy.spatial.transform._rigid_transform_cy, scipy.cluster._vq, scipy.cluster._hierarchy, scipy.cluster._optimal_leaf_ordering, scipy.interpolate._fitpack, scipy.interpolate._dfitpack, scipy.sparse.linalg._dsolve._superlu, scipy.sparse.linalg._eigen.arpack._arpacklib, scipy.sparse.linalg._propack, scipy.optimize._group_columns, scipy.optimize._trlib._trlib, scipy.optimize._lbfgsb, _moduleTNC, scipy.optimize._moduleTNC, scipy.optimize._slsqplib, scipy.optimize._minpack, scipy.optimize._lsq.givens_elimination, scipy.optimize._zeros, scipy._lib._uarray._uarray, scipy.linalg._decomp_interpolative, scipy.optimize._bglu_dense, scipy.optimize._lsap, scipy.optimize._direct, scipy.interpolate._dierckx, scipy.interpolate._ppoly, scipy.interpolate._interpnd, scipy.interpolate._rbfinterp_pythran, scipy.interpolate._rgi_cython, scipy.integrate._odepack, scipy.integrate._quadpack, scipy.integrate._vode, scipy.integrate._dop, scipy.integrate._lsoda, scipy.special.cython_special, scipy.stats._stats, scipy.stats._biasedurn, scipy.stats._stats_pythran, scipy.stats._levy_stable.levyst, scipy.stats._ansari_swilk_statistics, scipy.sparse.csgraph._tools, scipy.sparse.csgraph._shortest_path, scipy.sparse.csgraph._traversal, scipy.sparse.csgraph._min_spanning_tree, scipy.sparse.csgraph._flow, scipy.sparse.csgraph._matching, scipy.sparse.csgraph._reordering, scipy.stats._sobol, scipy.stats._qmc_cy, scipy.stats._rcont.rcont, scipy.stats._qmvnt_cy, scipy.ndimage._nd_image, scipy.ndimage._rank_filter_1d, _ni_label, scipy.ndimage._ni_label, scipy._lib._test_ccallback, scipy.fftpack.convolve, scipy.integrate._test_odeint_banded, scipy.io.matlab._mio_utils, scipy.io.matlab._streams, scipy.io.matlab._mio5_utils, scipy.linalg._matfuncs_sqrtm_triu, scipy.ndimage._ctest, _cytest, scipy.ndimage._cytest, scipy.odr.__odrpack, scipy.optimize.cython_optimize._zeros, scipy.signal._sigtools, scipy.signal._max_len_seq_inner, scipy.signal._upfirdn_apply, scipy.signal._spline, scipy.signal._sosfilt, scipy.signal._peak_finding_utils, gmpy2.gmpy2, scipy.special._test_internal, scipy.stats._unuran.unuran_wrapper, _cffi_backend, scipy._lib._test_deprecation_def, scipy._lib._test_deprecation_call, _brotli, zstandard.backend_c (total: 124)
```

lucascolley:
@rgommers any other linux jobs we might want to convert?

.github/workflows/linux_blas.yml
```diff
@@ -7,6 +7,7 @@ name: BLAS tests (Linux)
 #
 #   - mkl:
 #         Tests MKL installed from PyPI (because easiest/fastest) in
+#           XXX: do we still want to isntall MKL from PyPI?
```
rgommers: No, that's fine to drop - that was only for convenience

pixi.toml
rgommers: I'd keep this minimal I think, and only add it to `build`/`test` until we need more than that.

rgommers:
> @rgommers any other linux jobs we might want to convert?

I don't see any others, they're all special cases.

pixi.toml
lucascolley: currently it has:
- default, build, test, test-fail-slow, ipython, test-mkl, scipy-openblas

That might be fine? Or should we remove ipython, test-mkl, scipy-openblas?

rgommers: I'd remove it I think, since we probably don't have any contributors on this platform and we're not using it in CI.

lucascolley:
the aarch64 fatal python error is coming from `scipy/special/tests/test_basic.py::TestFactorialFunctions::test_factorialx_inf_nan`

lucascolley:
after adding a few skips in `special` tests, the test suite now runs to completion on `aarch64`, with 8 failures, again all in `special`.

Are we happy to put these down to conda-forge special cases @rgommers ?

rgommers:
The six `test_negative_real_gh14582` failures look like an error in the test, because the signs of the results are off. The differences are large, so it's not just a tolerance bump that is needed. Might be an invalid assumption on the branch cut or some such thing (no linalg functions used in `spherical_*n` it looks like, that was my first guess). Maybe it's obvious to you what is going on there @steppi?

The failures for `test_expm1_complex` and `test_log1p_complex` look like compiler issues; these were reported once before at gh-6499 for NetBSD, and that just went away. Also unexpected. They don't seem skipped in conda-forge builds nor are marked as slow, so unclear what is going on there.

.github/workflows/linux.yml
```diff
@@ -501,82 +501,57 @@ jobs:
...
+            scipy.signal
+            scipy.sparse
+            scipy.special
+            scipy.stats
+        run: |
+          # Note: Only fast tests; full test suite is unlikely to uncover anything more,
+          #       and it'll be quite slow with pytest-run-parallel. Also skip
+          #       tests that won't run in parallel, those are covered by the
+          #       parallel == 0 run.
+          pixi run test-parallel -t $TEST_SUBMODULES -- --parallel-threads=4 --skip-thread-unsafe=true
```
lucascolley: is it problematic to have `pytest-xdist` as an active plugin here?

lucascolley: could/should we move `--parallel-threads=4 --skip-thread-unsafe=true` into the task?

pixi.toml
```diff
@@ -644,3 +721,30 @@ platforms = ["linux-64", "osx-arm64", "win-64"]
...
+# freethreaded python
+
+[feature.py-freethreading]
+platforms = ["linux-64", "osx-arm64", "win-64"]
+
+[feature.py-freethreading.dependencies]
+python-freethreading = "*"
+
+[feature.build-freethreading.tasks.build-freethreading]
+# `-Dc_args='-O0'` needed to override `-O2` from clang activation `CFLAGS`
```
lucascolley:
```suggestion
```

pixi.toml
```diff
@@ -3,7 +3,7 @@ name = "scipy"
 description = "Fundamental algorithms for scientific computing in Python."
 authors = ["SciPy Developers <scipy-dev@python.org>"]
 channels = ["https://prefix.dev/conda-forge"]
-platforms = ["linux-64", "osx-arm64", "win-64"]
+platforms = ["linux-64", "osx-arm64", "win-64", "linux-aarch64"]
```
lucascolley, line 6: only environments in the default solve group support this new platform, so we aren't taking on loads of unnecessary environments.

lucascolley:
> The six test_negative_real_gh14582 failures look like an error in the test, because the signs of the results are off. The differences are large, so it's not just a tolerance bump that is needed. Might be an invalid assumption on the branch cut or some such thing (no linalg functions used in spherical_*n it looks like, that was my first guess). Maybe it's obvious to you what is going on there @steppi?

Gentle ping @steppi

steppi:
> > 
> 
> Gentle ping @steppi

I'll take a look tomorrow

steppi:
The problem seems to be that this line in `sph_bessel_j`'s kernel and similar lines in the kernels for other spherical bessel functions

```C++
   std::complex<T> out = std::sqrt(static_cast<T>(M_PI_2) / z) * cyl_bessel_j(n + 1 / static_cast<T>(2), z);
```

relies on complex division giving a consistent result for the sign of zero in the imaginary part in the expression

`static_cast<T>(M_PI_2) / z` when IEEE-754 makes no such guarantees about complex arithmetic. This means that the wrong branch of `sqrt` is getting chosen.

A fix in xsf should be relatively straightforward and we can add `linux-arch64` tests to the xsf CI too.

tylerjereddy:
@lucascolley reviewer activity slowed down a bit in recent months, and there's a potential dependency on a change in `xsf`, and this makes some substantial changes to CI that might best be done after branching in ~12 days?

I bumped the milestone as a result--switch it back if I've misunderstood.

.github/workflows/linux.yml
```diff
@@ -565,48 +566,15 @@ jobs:
...
-      - name: Install Ubuntu dependencies
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y ccache
-
-      - name: Set up ccache
-        uses: ./.github/ccache
-        with:
-          workflow_name: ${{ github.workflow }}
-          job_name: ${{ github.job }}
```
lucascolley: let's check ccache before merge

.github/workflows/linux_blas.yml
```diff
@@ -154,40 +151,14 @@ jobs:
...
-
-      - name: Install dependencies
-        run: |
-          pip install -r requirements/build.txt
-          pip install build pytest hypothesis pytest-xdist pooch
-          pip install -r requirements/openblas64.txt
-
-      - name: Write out scipy-openblas64.pc
-        run: |
-          python -c'import scipy_openblas64 as so64; print(so64.get_pkg_config())' > scipy-openblas.pc
```
lucascolley: may need to include this at the Pixi or spin level

lucascolley:
hey @ngoldbaum, do you know what the right response is to errors like this?
```
>   import _brotli
E   RuntimeWarning: The global interpreter lock (GIL) has been enabled to load module '_brotli', which has not declared that it can run safely without the GIL. To override this behavior and keep the GIL disabled (at your own risk), run with PYTHON_GIL=0 or -Xgil=0.
```

ngoldbaum:
Set PYTHON_GIL=0 to ignore it with a TODO to remove it and wait for movement on https://github.com/google/brotli/issues/1241#issuecomment-4359720657. Google is unfortunately inscrutable on stuff like this.

lucascolley:
do you know why it would only appear now with the changes in this PR?

lucascolley:
the main change I can see was switching from installing `cp314t-dev` with setup-python, to `python-freethreading 3.14.5` from conda-forge

ngoldbaum:
> do you know why it would only appear now with the changes in this PR?

Sorry, not sure offhand. Maybe convert warnings into errors to get a traceback?

rgommers:
It looks like that shows up now because you install `pooch` from conda-forge rather than PyPI. And it depends on `urllib`, which has only an optional dependency on `brotli`. And since conda-forge is unable to do optional dependencies, it adds a regular unconditional dependency on it instead.

lucascolley:
> It looks like that shows up now because you install `pooch` from conda-forge rather than PyPI. And it depends on `urllib`, which has only an optional dependency on `brotli`. And since conda-forge is unable to do optional dependencies, it adds a regular unconditional dependency on it instead.

ah that indeed explains it, thanks Ralf :)

lucascolley:
> Set PYTHON_GIL=0 to ignore it with a TODO to remove it and wait for movement on [google/brotli#1241 (comment)](https://github.com/google/brotli/issues/1241#issuecomment-4359720657). Google is unfortunately inscrutable on stuff like this.

The fix I have implemented for now is to force `python-gil` for all environments using the default feature, and set `PYTHON_GIL=0` explicitly for the `python-freethreading` feature.

.github/workflows/linux.yml
```diff
@@ -681,43 +648,28 @@ jobs:
...
+      - uses: prefix-dev/setup-pixi@ba3bb36eb2066252b2363392b7739741bb777659 # v0.8.1
+        with:
+          pixi-version: v0.70.1
+          cache: false
+
+      - name: Test SciPy
+        run: |
+          export OMP_NUM_THREAD=2
+          # various factorial tests were giving a fatal python error in CI
+          # pixi run test-fail-slow -k 'not (test_factorial or test_nan_inputs)'
```
lucascolley:
```suggestion
```

.github/workflows/linux.yml
lucascolley: @nickodell any thoughts on the desired cache setup?

.github/workflows/linux.yml
```diff
@@ -691,32 +662,14 @@ jobs:
...
-          sudo apt-get update
-          sudo apt-get install -y libopenblas-dev libatlas-base-dev liblapack-dev \
-            libgmp-dev libmpfr-dev libsuitesparse-dev ccache libmpc-dev
-
-      - name: Install Python packages
-        run: |
-          python -m pip install -r requirements/build.txt -r requirements/test.txt -r requirements/dev.txt
-          # We want to check for test timing only in a single job, on Linux, running the
-          # fast test suite. This is that job. See gh-20806 for previous issues
-          # after running this job on Windows and in multiple jobs.
```
lucascolley: this comment has moved above

pixi.toml
```diff
@@ -195,20 +224,21 @@ features = ["build-deps", "run-deps", "test-deps", "py312", "scikit-sparse", "sy
 features = ["build-deps", "run-deps", "test-deps", "scipy-openblas"]
 
 [environments.scipy-openblas64]
-# tasks: build-scipy-openblas64, test-scipy-openblas64
-features = ["build-deps", "run-deps", "test-deps", "scipy-openblas64"]
+# tasks: build-scipy-openblas64, test-scipy-openblas64, build-scipy-openblas, test-scipy-openblas
+features = ["build-deps", "run-deps", "test-deps", "scipy-openblas", "scipy-openblas64"]
```
lucascolley: as in `requirements/openblas64`, we install both the 32 and 64 versions of `scipy-openblas` for this environment

.github/workflows/linux_blas.yml
lucascolley: I believe this part of the comment is outdated
```suggestion
#    - scipy-openblas64:
#         Tests ILP64-enabled build with scipy-openblas64.
```

.github/workflows/linux.yml
```diff
@@ -550,95 +549,63 @@ jobs:
...
-      run: |
-        pip install -r requirements/build.txt
-        pip install -r requirements/openblas.txt
-        pip install spin pytest pytest-xdist threadpoolctl pooch hypothesis "click<8.3.0"
-
-    - name: Install pytest-run-parallel
-      if: ${{ matrix.parallel == '1'}}
-      run: |
-        pip install pytest-run-parallel
-        pip uninstall --yes pytest-xdist
```
lucascolley, line 595: `pytest-xdist` is included in the environment now — is that expected to cause problems with `pytest-run-parallel` @ngoldbaum ?

pixi.toml
```diff
@@ -791,6 +844,36 @@ platforms = ["linux-64", "osx-arm64", "win-64"]
...
+cmd = "spin test --no-build --build-dir=build-freethreading"
+depends-on = "build-freethreading"
+description = "Test SciPy (freethreaded Python)"
+
+[feature.test-parallel-freethreading.dependencies]
+pytest-run-parallel = "*"
+psutil = "*"
+
+[feature.test-parallel-freethreading.tasks.test-parallel-freethreading]
+cmd = "spin test --no-build --build-dir=build-freethreading --parallel-threads=4 --skip-thread-unsafe=true"
```
lucascolley: does it make sense to hardcode `--parallel-threads=4 --skip-thread-unsafe=true` here, or may local invocations often want something different to CI?

.github/workflows/linux.yml
ngoldbaum: As long as it's not invoked it shouldn't cause CPU oversubscription.

pixi.toml
ngoldbaum: This should be fine. I would include `skip_thread_unsafe` because otherwise you're running tests that are redundant with other CI jobs.

.github/workflows/linux_blas.yml
```diff
@@ -151,45 +136,27 @@ jobs:
...
+
+      - name: Build with ILP64
+        run: pixi run build-scipy-openblas64
+
+      - name: Test
+        run: pixi run --skip-deps test-scipy-openblas64
+
+      - name: Install and test linalg ILP64 downstream test package
+        run: |
+          uv pip install --no-build-isolation -v scipy/linalg/tests/_cython_examples/ilp64_test_package
```
lucascolley: probably this line can go back into a separate step

.github/workflows/linux.yml
```diff
@@ -550,95 +549,36 @@ jobs:
...
-    - name: Run tests (full)
-      if: ${{ matrix.parallel == '0'}}
-      run: |
-        spin test -j4 -m full --durations=10
-
-    - name: Run tests (fast, with pytest-run-parallel)
-      if: ${{ matrix.parallel == '1'}}
-      env:
-        # Excluded modules:
-        # - scipy.spatial has multiple issues  in kdtree/qhull, and gh-20655 is pending.
```
lucascolley, line 616: looks like this is resolved, a PR superseding the linked PR was merged

.github/workflows/linux.yml
```diff
@@ -550,95 +549,36 @@ jobs:
...
-          scipy.optimize
-          scipy.signal
-          scipy.sparse
-          scipy.special
-          scipy.stats
-      run: |
-        # Note: Only fast tests; full test suite is unlikely to uncover anything more,
-        #       and it'll be quite slow with pytest-run-parallel. Also skip
-        #       tests that won't run in parallel, those are covered by the
-        #       parallel == 0 run.
```
lucascolley, line 640: comment moved into the manifest

pixi.toml
lucascolley: @rgommers maybe this is undesirable?

.github/workflows/linux_blas.yml
```diff
@@ -106,42 +100,37 @@ jobs:
...
+        shell: pixi run -e build-mkl-cyilp64 spin run --build-dir=build-build-mkl-cyilp64 bash -e {0}
+      
+      - name: Test linalg ILP64 downstream test package
+        run: |
+          cp -r scipy/linalg/tests/_cython_examples/ilp64_test_package/tests /tmp/ilp64_tests
+          cd /tmp
+          python -m pytest ilp64_tests
+        shell: pixi run -e build-mkl-cyilp64 spin run --build-dir=build-mkl-cyilp64 bash -e {0}
```
lucascolley: maybe we can install `uv` and `pytest` with `pixi global` in the workflow instead of adding them to the Pixi environments as is currently done

.github/workflows/linux_blas.yml
```diff
@@ -106,42 +100,37 @@ jobs:
...
+          pixi run build-mkl-cyilp64
+          pixi run check-mkl-cyilp64-sharedlibs
+
+      - name: Test
+        run: pixi run --skip-deps test-mkl-cyilp64
+
+      - name: Install linalg ILP64 downstream test package
+        run: |
+          uv pip install --no-build-isolation -v scipy/linalg/tests/_cython_examples/ilp64_test_package
+        shell: pixi run -e build-mkl-cyilp64 spin run --build-dir=build-build-mkl-cyilp64 bash -e {0}
```
lucascolley:
```suggestion
        shell: pixi run -e build-mkl-cyilp64 spin run --build-dir=build-mkl-cyilp64 bash -e {0}
```

.github/workflows/linux.yml
nickodell: @lucascolley Major thing is CCACHE_COMPILERCHECK.

If you're using Pixi to provide compilers, then be aware that Pixi sets mtime for those compilers to the last modified time. In contrast, apt and yum set mtime to consistent values.

By default, ccache uses compiler mtime as a fast way of checking if it needs to invalidate the cache entry due to the compiler changing. The result is that ccache will never get any cache hits in this configuration: it will always detect the compiler as having changed since the last CI run.

To avoid this, set the environment variable `CCACHE_COMPILERCHECK` to `content`.

Here's an example of this from our codebase.

https://github.com/scipy/scipy/blob/d4a7a2463e1eb2825848a61f6913013fd1f5c21b/.github/workflows/array_api.yml#L26-L27

See [ccache docs](https://ccache.dev/manual/4.13.6.html) for more details.

nickodell: Oh, and the other thing is that Pixi often sets CC and CXX, which disables meson autodetection of ccache. But I think you addressed that already in pixi.toml.

.github/workflows/linux_blas.yml
```diff
@@ -151,45 +140,32 @@ jobs:
...
+        run: pixi run build-scipy-openblas64
+
+      - name: Test
+        run: pixi run --skip-deps test-scipy-openblas64
+
+      - name: Install linalg ILP64 downstream test package
+        run: |
+          pixi global install uv
+          uv pip install --no-build-isolation -v scipy/linalg/tests/_cython_examples/ilp64_test_package
+        shell: pixi run -e scipy-openblas64 spin run --build-dir=build-scipy-openblas64 bash -e {0}
```
lucascolley:
```suggestion
          pixi global install uv
          # `spin run` sets the system Python, so pass `--system`
          uv pip install --system --no-build-isolation -v scipy/linalg/tests/_cython_examples/ilp64_test_package
        # `spin run` needs build deps available
        shell: pixi run -e scipy-openblas64 spin run --build-dir=build-scipy-openblas64 bash -e {0}
```

.github/workflows/linux_blas.yml
```diff
@@ -154,40 +150,29 @@ jobs:
...
+          uv pip install --system --no-build-isolation -v scipy/linalg/tests/_cython_examples/ilp64_test_package
 
-      - name: Test ILP64 downstream test package
+      - name: Test linalg ILP64 downstream test package
+        shell: pixi run -e scipy-openblas64 spin run --build-dir=build-scipy-openblas64 bash -e {0}
         run: |
           cp -r scipy/linalg/tests/_cython_examples/ilp64_test_package/tests /tmp/ilp64_tests
           cd /tmp
-          python -m pytest ilp64_tests
+          pytest ilp64_tests
```
lucascolley, line 183: @ev-br does this make sense?

Previously we were building and installing a SciPy wheel, building and installing `ilp64_test_package`, and running the tests, all in the global venv from `setup-python`.

Now, we are
- building SciPy into `build-scipy-openblas64` in the `scipy-openblas64` Pixi env
- installing `uv` globally (there is no longer a `pip` hanging around)
- spinning up an env with the built SciPy via `spin run` and installing `ilp64_test_package` into it
- spinning up that env again to run `pytest`

pixi.toml
rgommers: Indeed, that seems undesirable - there are then two environments connected to the same build task, which might be confusing. Also, we will probably stop installing the 32-bit variant in `requirements/openblas64`, given that we can now (after Fortran is really gone) have a pure 64-bit build, and the mixed 32/64-bit build is very odd so we definitely don't want to encourage that setup.

.github/workflows/linux_blas.yml
```diff
@@ -151,45 +145,34 @@ jobs:
...
+      - name: Test
+        run: pixi run --skip-deps test-scipy-openblas64
+
+      - name: Install linalg ILP64 downstream test package
+        # `spin run` needs build deps available
+        shell: pixi run -e scipy-openblas64 spin run --build-dir=build-scipy-openblas64 bash -e {0}
+        run: |
+          pixi global install uv
+          # `spin run` sets the system Python, so pass `--system`
+          uv pip install --system --no-build-isolation -v scipy/linalg/tests/_cython_examples/ilp64_test_package
```
rgommers: Can you avoid dropping `-Csetup-args=--werror` here?

.github/workflows/linux_blas.yml
rgommers: Yes this looks fine.

.github/workflows/linux_blas.yml
```diff
@@ -43,28 +43,25 @@ jobs:
...
+      - uses: prefix-dev/setup-pixi@ba3bb36eb2066252b2363392b7739741bb777659 # v0.8.1
+        with:
+          pixi-version: v0.70.1
+          cache: false
+
+      - name: Build with defaults (LP64)
+        run: |
+          pixi run --environment=build-mkl-lp64 pkg-config --libs mkl-dynamic-lp64-seq  # check link flags
+          pixi run build-mkl-lp64
+          pixi run check-mkl-lp64-sharedlibs
```
rgommers, line 62: This is a change, not quite sure why it's added here. Won't this be hiding a nonzero exit status for the command on the line above if a build failure occurs?

.github/workflows/linux_blas.yml
```diff
@@ -74,29 +71,26 @@ jobs:
...
+      - uses: prefix-dev/setup-pixi@ba3bb36eb2066252b2363392b7739741bb777659 # v0.8.1
+        with:
+          pixi-version: v0.70.1
+          cache: false
+
+      - name: Build with ILP64, cython-blas LP64
+        run: |
+          # Note: the `lp64` in the .pc name isn't a mistake, due to cython-blas-abi=lp64
+          pixi run -e build-mkl-ilp64 pkg-config --libs mkl-dynamic-lp64-seq  # check link flags
+          pixi run build-mkl-ilp64
```
rgommers, line 90: This changes the linked shared library from `mkl-dynamic-lp64-seq` to `mkl-dynamic-ilp64-seq`, that seems wrong. I think a change is needed in the `build-mkl-ilp64` task definition; it should be linking `-lp64`.

.github/workflows/linux_blas.yml
```diff
@@ -106,42 +100,42 @@ jobs:
...
+
+      - uses: prefix-dev/setup-pixi@ba3bb36eb2066252b2363392b7739741bb777659 # v0.8.1
+        with:
+          pixi-version: v0.70.1
+          cache: false
+
+      - name: Build with ILP64, cython-blas ILP64
+        run: |
+          pixi run -e build-mkl-ilp64 pkg-config --libs mkl-dynamic-ilp64-seq  # check link flags
+          pixi run build-mkl-cyilp64
```
rgommers, line 120: This one does do the right thing

.github/workflows/linux_blas.yml
lucascolley: > Won't this be hiding a nonzero exit status for the command on the line above if a build failure occurs?

Oh, I hadn't considered that! Do we need a `set -euo pipefail` ?

> This is a change, not quite sure why it's added here.

I figured exercising the task _somewhere_ in CI was an improvement over nowhere.

rgommers (commented):
I went through the whole diff and some of the CI logs. This LGTM, modulo the few things I commented on.

.github/workflows/linux_blas.yml
rgommers: I'd just leave it out, or otherwise put it in a separate job step. It's always better to have two command that both can fail (due to things you're actually testing, not a random connection error or some such thing) in separate steps.

lucascolley:
Let me look into a PR to drop `scipy-openblas32` from the `openblas64` requirements and the CI job. If that is quick then we can merge that and come back here.

.github/workflows/linux_blas.yml
ev-br: If you ask me, `pixi run -e ... spin run ... bash - e {0}` is inscrutable. OTOH the original was messy as well, and for a good reason. 
So if that's what it takes, well, good that the stanza is written out explicitly.

.github/workflows/linux.yml
lucascolley: thanks Nick, pushing a change to reintroduce ccache into the one job that had it, and include the `CCACHE_COMPILERCHECK` workaround

rgommers (approved):
LGTM now, thanks @lucascolley. One last request: could you please add the commit hashes for the two STY commits to `.git-blame-ignore-revs` in an extra commit (which can skip CI), and then merge?

lucascolley:
awesome, thanks a lot Ralf, Nick, Nathan, Evgeni!

PR #23702, opened 2025-09-30
ENH: stats.lmoment: add array API support

mdhaber:
#### Reference issue
Toward gh-20544

Adds array API support to `stats.lmoment`.

### DIFF
--- a/scipy/stats/_stats_py.py
+++ b/scipy/stats/_stats_py.py
@@ -71,6 +71,7 @@ from scipy._lib._array_api import (
     is_dask,
     is_numpy,
     is_cupy,
+    is_array_api_strict,
     xp_size,
     xp_vector_norm,
     xp_promote,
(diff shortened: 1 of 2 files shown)


mdhaber:
Failures are mostly just `np.isdtype` not existing pre-2.0 and `special.binom` not working for torch on GPU. Will fix when there are other commends to address. *Update: went ahead and fixed.*

scipy/stats/_stats_py.py
```diff
@@ -10315,7 +10320,18 @@ def _prk(r, k):
     return (-1)**(r-k)*special.binom(r, k)*special.binom(r+k, k)
 
 
-@xp_capabilities(np_only=True)
+def _triu(x, xp):
+    # not array API compatible, but supported by cupy, jax.numpy, torch, dask.array
```
j-bowhay: Maybe my brain is still asleep but `triu` is in the standard?

mdhaber: Still asleep, I think : )

image

So, no, it is not, but just as we sometimes use special functions that are not in the standard, I am using `triu` here.

j-bowhay: Sorry what I meant was what is the difference between this and https://data-apis.org/array-api/2024.12/API_specification/generated/array_api.triu.html#triu ?

mdhaber: Ah : ) So I was asleep when writing this.[^1] I thought it was, too... not sure how I didn't find it when I was looking. I'll fix that!

[^1]: Ah, I did submit the PR at midnight.

scipy/stats/_stats_py.py
```diff
@@ -71,6 +71,7 @@
     is_dask,
     is_numpy,
     is_cupy,
+    is_array_api_strict,
```
j-bowhay:
```suggestion
```

mdhaber:
CI run for 21034f2ac344b48693ccac7fe2dab0dc291d6895 had three failures.
- Lint failure (to be fixed and re-checked)
- Linux tests/ Prerelease deps - failing [everywhere](https://github.com/scipy/scipy/actions/workflows/linux.yml), see matplotlib/matplotlib#30617.
- macOS tests - failing [everywhere](https://github.com/scipy/scipy/actions/workflows/macos.yml). Many functions are failing intermittently, and one of those happens to be `lmoment`. Exactly the same failure we have here shows up in some of those, e.g. [here](https://github.com/scipy/scipy/actions/runs/18247136645/job/51956626119).

j-bowhay (approved):

PR #23705, opened 2025-09-30
MAINT: stats.rv_discrete.ppf/isf: fix behavior with array args, `q=0`/`q=1`

### LINKED ISSUE #18919: BUG: Bad result (broadcasting ?) for binom.ppf and q=1, when all arguments have size > 1
### Describe your issue.

Normally, binom.ppf should return the upper value (the n parameter) when q = 1, for example:
```python3
>>>stats.binom.ppf(q=1, n=10, p=0.5)
10.0
```

but when the arguments are no longer scalars (at least for 'n' and for 'q'), we don't get the expected result : 
```python3
>>>q = [0.5, 1]
>>>n = [10, 11]
>>>p = [0, 0.5]
>>>stats._boost.binom.ppf(q, n, p) # expected result
array([ 0., 11.])
>>>stats.binom.ppf(q, n, p) # but...
array([ 0., 10.])
```

By doing several tests with higher dimensions, it seems that the bug occurs for each q = 1, and instead of taking its correct value, it takes the next value of n.flatten()
```python3
>>> q = [[0.5, 1.0, 0.5],
...      [1.0, 0.5, 1.0],
...      [0.5, 1.0, 0.5]]
>>> n = [[45, 46, 47],
...      [48, 49, 50],
...      [51, 52, 53]]
>>> p = 0.5
>>> stats._boost._binom_ppf(q, n, p) 
array([[22., 46., 23.],
       [48., 24., 50.],
       [25., 52., 26.]])
>>> stats.binom.ppf(q, n, p) 
array([[22., 45., 23.],
       [46., 24., 47.],
       [25., 48., 26.]])
>>> n = [45, 46, 47]
>>> stats.binom.ppf(q, n, p)
array([[22., 45., 23.],
       [46., 23., 47.],
       [22., 45., 23.]])
```

### Reproducing Code Example

```python
import numpy as np
from scipy.stats import binom, _boost

q = [[0.5, 1.0, 0.5],
     [1.0, 0.5, 1.0],
     [0.5, 1.0, 0.5]]

n = [[45, 46, 47],
     [48, 49, 50],
     [51, 52, 53]]

p = 0.5

result1 = _boost._binom_ppf(q, n, p)
result2 = np.vectorize(binom.ppf)(q, n, p)

bad_result = binom.ppf(q, n, p)

print(f"{(result1 == result2).all()=}") # True
print(f"{(result1 == bad_result).all()=}") # False
```

### Error message

```shell
No error message
```

```shell
1.11.1 1.25.1 sys.version_info(major=3, minor=11, micro=4, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    detection method: pkgconfig
    found: true
    include directory: /c/opt/64/include
    lib directory: /c/opt/64/lib
    name: openblas
      NO_LAPACK= NO_LAPACKE= NO_AFFINITY=1 USE_OPENMP= SKYLAKEX MAX_THREADS=2
    pc file directory: c:/opt/64/lib/pkgconfig
    version: 0.3.21.dev
  lapack:
    detection method: pkgconfig
    found: true
    include directory: /c/opt/64/include
    lib directory: /c/opt/64/lib
    name: openblas
      NO_LAPACK= NO_LAPACKE= NO_AFFINITY=1 USE_OPENMP= SKYLAKEX MAX_THREADS=2
    pc file directory: c:/opt/64/lib/pkgconfig
    version: 0.3.21.dev
  pybind11:
    detection method: config-tool
    include directory: unknown
    name: pybind11
    version: 2.10.4
Compilers:
  c:
    commands: cc
    linker: ld.bfd
    name: gcc
    version: 10.3.0
  c++:
    commands: c++
    linker: ld.bfd
    name: gcc
    version: 10.3.0
  cython:
    commands: cython
    linker: cython
    name: cython
    version: 0.29.35
  fortran:
    commands: gfortran
    linker: ld.bfd
    name: gcc
    version: 10.3.0
  pythran:
    include directory: C:\Users\runneradmin\AppData\Local\Temp\pip-build-env-dqdt9tl_\overlay\Lib\site-packages/pythran 
    version: 0.13.1
Machine Information:
  build:
    cpu: x86_64
    endian: little
    family: x86_64
    system: windows
  cross-compiled: false
  host:
    cpu: x86_64
    endian: little
    family: x86_64
    system: windows
Python Information:
  path: C:\Users\runneradmin\AppData\Local\Temp\cibw-run-qsxtw44z\cp311-win_amd64\build\venv\Scripts\python.exe
  version: '3.11'
```

mdhaber:
#### Reference issue
Closes gh-18919

### What does this implement/fix?
gh-18919 reported strange behavior in discrete distribution ppf/isf when arguments are arrays and `q=0`/`q=1`.
In essence it's this gotcha:
```python3
import numpy as np
x = np.asarray([1, 2, 3, 4])
np.place(x, x > 2, 2*x)
x  
# expected:  array([1, 2, 6, 8])
# got:       array([1, 2, 2, 4])
```
From [`np.place` ](https://numpy.org/doc/2.1/reference/generated/numpy.place.html) documentation of the third argument, `vals`:
> Only the first N elements are used, where N is the number of True values in `mask`.

So when `vals` has too many elements, no error is raised; it just starts taking elements from the left.

Looks like this bug has been around since the beginning, or at least since uses of `insert` were replaced with `place`. The same bug was fixed for continuous distributions in https://github.com/scipy/scipy/commit/126d5bf9b96e137f1199c1767962adb3497e1041 but the discrete distributions were not done at the same time.

### DIFF
--- a/scipy/stats/tests/test_discrete_basic.py
+++ b/scipy/stats/tests/test_discrete_basic.py
@@ -600,7 +600,8 @@ def test_gh18919_ppf_array_args():
     res = stats.binom.ppf(q, n, p)
     np.testing.assert_allclose(res, ref)
 
-def test_gh18919_ppf_isf_array_args2():
+@pytest.mark.parametrize("dist", [stats.binom, stats.boltzmann])
+def test_gh18919_ppf_isf_array_args2(dist):
     # a more general version of the test above. Requires that arguments are broadcasted
     # by the infrastructure.
     rng = np.random.default_rng(34873457824358729823)
@@ -612,12 +613,12 @@ def test_gh18919_ppf_isf_array_args2():
     q[rng.random(size=30) > 0.7] = 0
     q[rng.random(size=30) > 0.7] = 1
 
-    res = stats.binom.ppf(q, n, p, loc=loc)
-    ref = _ufuncs._binom_ppf(q, n, p) + loc
-    ref[np.broadcast_to(q, ref.shape) == 0] -= 1
+    args = (q, n, p) if dist == stats.binom else (q, p, n)
+
+    res = dist.ppf(*args, loc=loc)
+    ref = np.vectorize(dist.ppf)(*args) + loc
     np.testing.assert_allclose(res, ref)
 
-    res = stats.binom.isf(q, n, p, loc=loc)
-    ref = _ufuncs._binom_isf(q, n, p) + loc
-    ref[np.broadcast_to(q, ref.shape) == 1] -= 1
+    res = dist.isf(*args, loc=loc)
+    ref = np.vectorize(dist.isf)(*args) + loc
     np.testing.assert_allclose(res, ref)
--- a/scipy/stats/tests/test_discrete_basic.py
+++ b/scipy/stats/tests/test_discrete_basic.py
@@ -7,6 +7,7 @@ import numpy as np
 import pytest
 
 from scipy import stats
+from scipy.special import _ufuncs
 from .common_tests import (check_normalization, check_moment,
                            check_mean_expect,
                            check_var_expect, check_skew_expect,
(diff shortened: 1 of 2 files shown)


scipy/stats/tests/test_discrete_basic.py
```diff
@@ -579,3 +580,44 @@ def _pmf(self, i):
...
+    n = rng.integers(10, 30, size=(10, 1, 1))
+    p = rng.random(size=(4, 1))
+    loc = rng.integers(5, size=(3,))
+
+    q[rng.random(size=30) > 0.7] = 0
+    q[rng.random(size=30) > 0.7] = 1
+
+    res = stats.binom.ppf(q, n, p, loc=loc)
+    ref = _ufuncs._binom_ppf(q, n, p) + loc
+    ref[np.broadcast_to(q, ref.shape) == 0] -= 1
```
mdhaber: This is just due to a difference in conventions between the special function and old distribution infrastructure.
```python3
stats.binom.ppf(0, 10, 0.5)  # np.float64(-1.0), one less then left of support
```

scipy/stats/tests/test_discrete_basic.py
```diff
@@ -579,3 +580,44 @@ def _pmf(self, i):
...
+    # a more general version of the test above. Requires that arguments are broadcasted
+    # by the infrastructure.
+    rng = np.random.default_rng(34873457824358729823)
+    q = rng.random(size=(30, 1, 1, 1))
+    n = rng.integers(10, 30, size=(10, 1, 1))
+    p = rng.random(size=(4, 1))
+    loc = rng.integers(5, size=(3,))
+
+    q[rng.random(size=30) > 0.7] = 0
+    q[rng.random(size=30) > 0.7] = 1
```
mdhaber, line 614: Sprinkle in some 0s and 1s to trigger the bug, if it still exists

scipy/stats/tests/test_discrete_basic.py
```diff
@@ -579,3 +580,44 @@ def _pmf(self, i):
...
+    np.testing.assert_allclose(res, ref)
+
+def test_gh18919_ppf_isf_array_args2():
+    # a more general version of the test above. Requires that arguments are broadcasted
+    # by the infrastructure.
+    rng = np.random.default_rng(34873457824358729823)
+    q = rng.random(size=(30, 1, 1, 1))
+    n = rng.integers(10, 30, size=(10, 1, 1))
+    p = rng.random(size=(4, 1))
+    loc = rng.integers(5, size=(3,))
```
mdhaber, line 611: Rather than broadcasting all the arrays from the outset using `np.broadcast`, the old infrastructure broadcasts every once in a while with idioms like `*(cond == cond)` - multiplication with an all-true array of the right shape. This is another source of potential bugs, so check that the right thing is happening when all the arrays are of different (but compatible) shapes.

mdhaber:
Failures seem to be in `main` - the asv failure is new; it's a timeout in `cubature` (gh-23715):
```
For parameters: 'gk21', 5, 8, 1e-11
             asv: benchmark timed out (timeout 60.0s)
```

dschmitz89 (approved):
Quality work as always.

scipy/stats/tests/test_discrete_basic.py
```diff
@@ -579,3 +580,44 @@ def _pmf(self, i):
...
+         [51, 52, 53]]
+
+    p = 0.5
+
+    ref = _ufuncs._binom_ppf(q, n, p)
+    res = stats.binom.ppf(q, n, p)
+    np.testing.assert_allclose(res, ref)
+
+def test_gh18919_ppf_isf_array_args2():
+    # a more general version of the test above. Requires that arguments are broadcasted
```
dschmitz89: Nitpick: could we perform this test for at least one other distribution to check that it is fixe throughout? I do not mean to add another generic test for the old infra, just spot checks here for more than just `binom`.

dschmitz89:
Thanks @mdhaber

PR #23713, opened 2025-10-01
TST: macos-15-intel

andyfaff:
Highlights several test fails using x86_64+Accelerate+macOS15

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -135,6 +135,7 @@ jobs:
 
       - name: Test SciPy
         run: |
+          alias gfortran=gfortran-13
           pip install -r requirements/build.txt
           pip install -r requirements/dev.txt
           spin build --with-accelerate
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -136,6 +136,7 @@ jobs:
       - name: Test SciPy
         run: |
           pip install -r requirements/build.txt
+          pip install -r requirements/dev.txt
           spin build --with-accelerate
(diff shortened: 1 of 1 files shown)


andyfaff:
e.g. see #23710. @DietBru

andyfaff:
@lucascolley, the macos.yml configuration was fully pixified. I think it's important to retain at least one or two pip using jobs in there. For example, most of the issues with the wheel build infrastructure occur with macos, and that process relies on wheels+pip.

lucascolley:
> most of the issues with the wheel build infrastructure occur with macos, and that process relies on wheels+pip.

Is it just PyPI wheels that you want to exercise? Or the `pip` frontend also?

lucascolley (approved):
changes LGTM!

I spoke to @rgommers in person and it is TBC how much of the wheel infrastructure can move to Pixi, in which case this can be moved too. But we can cross that bridge when we come to it 👍

rgommers (approved):
LGTM, thanks @andyfaff! 

Minor comment, optional to address: I'd usually split build and test into separate job steps, for easier log navigation.

lucascolley:
Did we want to fix the failures before merging? Or will you open an issue for that?

andyfaff:
It's not really possible to see the issues locally, not many people have macos-x86_64. I thought it would aid those fixing the bugs, there are already open issues for them

andyfaff:
In reflection it may have been better to fix some of the issues here.

DietBru:
> In reflection it may have been better to fix some of the issues here.

Consult PR #23728 for further investigations.

PR #23716, opened 2025-10-01
TST: stats.Binomial.logcdf: improves precision in tails

mdhaber:
#### Reference issue
Closes gh-23708
Closes gh-23711

gh-23708 reported a hypothesis-generated test failure of `scipy.stats.Binomial.logccdf` when using the "complement" method in a tail of the distribution. This fixes that by increasing the precision of `logcdf`/`logccdf` in the "other" tail.
 
Separately, since it's a one-line fix reported at the same time (...and I accidentally committed to the wrong branch), it also fixes a test failure observed in gh-23711. It's a floating point equality test that was being performed with `assert_equal`, but it needs a small tolerance, so use `assert_allclose` with `rtol=1e-15`.

### DIFF
--- a/scipy/stats/_new_distributions.py
+++ b/scipy/stats/_new_distributions.py
@@ -479,22 +479,22 @@ class Binomial(DiscreteDistribution):
         return scu._binom_cdf(x, n, p)
 
     def _logcdf_formula(self, x, *, n, p, **kwargs):
-        # todo: add this strategy to infrastructure more generally while avoiding
-        # expensive calculations to figure out which tail we're in
-        mode = self._mode_formula(n=n, p=p)
-        return xpx.apply_where(x < mode, (x, n, p),
+        # todo: add this strategy to infrastructure more generally, but allow dist
+        #   author to specify threshold other than median in case median is expensive
+        median = self._icdf_formula(0.5, n=n, p=p)
+        return xpx.apply_where(x < median, (x, n, p),
             lambda *args: np.log(scu._binom_cdf(*args)),
-            lambda *args: scu._log1mexp(np.log(scu._binom_sf(*args)))
+            lambda *args: np.log1p(-scu._binom_sf(*args))
         )
 
     def _ccdf_formula(self, x, *, n, p, **kwargs):
         return scu._binom_sf(x, n, p)
 
     def _logccdf_formula(self, x, *, n, p, **kwargs):
-        mode = self._mode_formula(n=n, p=p)
-        return xpx.apply_where(x > mode, (x, n, p),
-            lambda *args: np.log(scu._binom_sf(*args)),
-            lambda *args: scu._log1mexp(np.log(scu._binom_cdf(*args)))
+        median = self._icdf_formula(0.5, n=n, p=p)
+        return xpx.apply_where(x < median, (x, n, p),
+            lambda *args: np.log1p(-scu._binom_cdf(*args)),
+            lambda *args: np.log(scu._binom_sf(*args))
         )
 
     def _icdf_formula(self, x, *, n, p, **kwargs):
--- a/scipy/stats/_new_distributions.py
+++ b/scipy/stats/_new_distributions.py
@@ -3,7 +3,9 @@ import sys
 import numpy as np
 from numpy import inf
 
+from scipy._lib import array_api_extra as xpx
 from scipy import special
+from scipy.special import _ufuncs as scu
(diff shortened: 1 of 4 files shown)


scipy/stats/_new_distributions.py
```diff
@@ -474,16 +476,32 @@ def _logpmf_formula(self, x, *, n, p, **kwargs):
...
-        return special._ufuncs._binom_cdf(x, n, p)
+        return scu._binom_cdf(x, n, p)
+
+    def _logcdf_formula(self, x, *, n, p, **kwargs):
+        # todo: add this strategy to infrastructure more generally while avoiding
+        # expensive calculations to figure out which tail we're in
+        mode = self._mode_formula(n=n, p=p)
+        return xpx.apply_where(x < mode, (x, n, p),
+            lambda *args: np.log(scu._binom_cdf(*args)),
+            lambda *args: scu._log1mexp(np.log(scu._binom_sf(*args)))
```
mdhaber:
```suggestion
            lambda *args: np.log1p(-scu._binom_sf(*args))
```

scipy/stats/_new_distributions.py
```diff
@@ -474,16 +476,32 @@ def _logpmf_formula(self, x, *, n, p, **kwargs):
...
 
     def _ccdf_formula(self, x, *, n, p, **kwargs):
-        return special._ufuncs._binom_sf(x, n, p)
+        return scu._binom_sf(x, n, p)
+
+    def _logccdf_formula(self, x, *, n, p, **kwargs):
+        mode = self._mode_formula(n=n, p=p)
+        return xpx.apply_where(x > mode, (x, n, p),
+            lambda *args: np.log(scu._binom_sf(*args)),
+            lambda *args: scu._log1mexp(np.log(scu._binom_cdf(*args)))
```
mdhaber:
```suggestion
            lambda *args: np.log1p(-scu._binom_cdf(*args))
```

scipy/stats/_new_distributions.py
```diff
@@ -474,16 +476,32 @@ def _logpmf_formula(self, x, *, n, p, **kwargs):
         return combiln + special.xlogy(x, p) + special.xlog1py(n-x, -p)
 
     def _cdf_formula(self, x, *, n, p, **kwargs):
-        return special._ufuncs._binom_cdf(x, n, p)
+        return scu._binom_cdf(x, n, p)
+
+    def _logcdf_formula(self, x, *, n, p, **kwargs):
+        # todo: add this strategy to infrastructure more generally while avoiding
+        # expensive calculations to figure out which tail we're in
+        mode = self._mode_formula(n=n, p=p)
```
mdhaber: Here, I've used the mode to determine which tail to use ~~because it is faster than other methods~~(not true, see https://github.com/scipy/scipy/pull/23716#discussion_r2395250383 [^1]. 

Ideally, we'd have this sort of logic in the infrastructure itself rather than in individual distributions. We could *always* make the decision based on the median, but sometimes the median is very expensive to calculate. So I think I'll allow distribution authors to specify a method that identifies the tail in which the CDF/CCDF should be evaluated. Something similar could be useful to determine where the probability mass is concentrated. That will help the integrator out when there are large, finite integration intervals where the integrand is essentially zero except for within a small sub-interval.

[^1]: I should think a bit about whether this can be off by one compared to the optimal choice for extreme parameters, but this is still an improvement over status quo.

mdhaber: For the binomial distribution, the mode is good enough to determine which tail to work with. I chose it because it's a computationally inexpensive way to determine which side of the median we're on. ~~But for the binomial distribution, the median is just as inexpensive to calculate, so might as well use that! I'll change it.~~*not exactly true!*

andyfaff:
If it helps you'll be able to see the test fails in https://github.com/scipy/scipy/actions/runs/18178013400/job/51748162618?pr=23713

andyfaff:
Seems ok to me, but you should probably get someone more familiar with the stats calcs to have a look at it as well.

andyfaff:
@mdhaber, this PR looks ok to me. Is there any barrier stopping this being merged?

mdhaber:
I think this is all set. There are some improvements I had in mind for later, but I wanted to get this in as-is first because doing it this way (`apply_where` based on the median) is the cleanest/simplest way to do the split, and this will close some of the issues you reported.

PR #23717, opened 2025-10-01
ENH: Migrate special.pdtrik to boost

dschmitz89:
#### Reference issue
Towards #19953 and #21966

#### Additional information
This PR migrates `pdtrik` to boost and adds unit tests. I could not get the old `asserts_inverts` machinery to work, so I added spot tests for accuracy and additional tests for special cases.

##### Implementation

The Poisson CDF is given by the upper incomplete gamma function `p=gammaincc(k+1, m)`. The boost equivalent is `gamma_q`. We can then calculate `k` using `gamma_q_inva(p, m) -1`.

##### Benchmark results

pdtrik_benchmark

<p>

```python
from mpmath import mp
import numpy as np

mp.dps = 50  # set decimal places for mpmath

@np.vectorize
def poisson_cdf(k, m):
    """Cumulative distribution function for the Poisson distribution."""
    k, m = mp.mpf(k), mp.mpf(m)
    return float(mp.gammainc(k + 1, a=m, regularized=True))


    """
    """
    #rel_error[np.isnan(rel_error)] = 1
    return rel_error

######### Test data generation and benchmarking #########

# rng = np.random.default_rng(12345)

# size = 5000
# m_min = 5
# m_max = 1000
# m = rng.uniform(m_min, m_max, size=size)  # mean values

# # k must be smaller than m to avoid p=1
# k = rng.uniform(m_min - 1, m_max - 1, size=size)
# mask = k < m

# m = m[mask]
# k = k[mask]

# print(f"left with {len(m)} values after masking")

# # k = np.zeros(size)
# # for i, x in enumerate(m):
# #     k[i] = rng.uniform(19, x - 1, 1)  # k values

# cdf_ref = poisson_cdf(k, m)
# cutoff = 1000 * np.finfo(float).tiny
# mask_cdf = (cdf_ref > cutoff) & (cdf_ref < 1)
# cdf_ref_masked = cdf_ref[mask_cdf]
# m_masked = m[mask_cdf]
# k_masked = k[mask_cdf]

# n_samples = len(m_masked)
# print(f"left with {n_samples} values after masking cdf")

# #plt.hist(cdf_ref_masked, bins=int(np.sqrt(n_samples)))
# #plt.show()
# #print(f"min: {cdf_ref_masked.min()}, max: {cdf_ref_masked.max()}")
# pdtrik = sc.pdtrik(cdf_ref_masked, m_masked)
# rel_error = calculate_rel_error(pdtrik, k_masked)
# np.save("pdtrik_rel_error_main.npy", rel_error)

######### Plotting #########

rel_error_main = np.load("pdtrik_rel_error_main.npy")
rel_error_pr = np.load("pdtrik_rel_error_pr.npy")

plt.figure(figsize=(10, 6))
plt.hist(np.log10(rel_error_main), bins=50, alpha=0.5, density=True, cumulative=True, label="main")
plt.hist(np.log10(rel_error_pr), bins=50, alpha=0.5, density=True, cumulative=True, label="pr")
plt.legend()
plt.title("Cumulative Distribution of Relative Errors in pdtrik")
plt.grid()
plt.show()
``` 

</p>
</details>


scipy/special/functions.json
```diff
@@ -88,7 +88,8 @@
     },
     "bdtrin": {
         "_cdflib_wrappers.pxd": {
-            "bdtrin": "ddd->d"        }
+            "bdtrin": "ddd->d"
```
dschmitz89: Looks like my json auto formatter was active here without me noticing. The changes are good in my opinion though. Will leave it up to the reviewer, if I should remove them.

scipy/special/boost_special_functions.h
```diff
@@ -2295,4 +2295,70 @@ chdtriv_double(double p, double x)
...
+    // than -1, we return NAN. This should never happen.
+    if (y < -1) {
+        sf_error("pdtrik", SF_ERROR_NO_RESULT, NULL);
+        return NAN;
+    }
+    // For small p, the direct formula yields values
+    // between -1 and 0. 
+    // For the Poisson distribution though, the bottom limit is 0,
+    // so we return 0 in that case.
+    if ((-1 <= y) && (y < 0)) {
```
dschmitz89:
```suggestion
    if (y < 0) {
```

scipy/special/functions.json
```diff
@@ -1019,4 +1021,4 @@
             "hypergeom_skewness_double": "ddd->d"
         }
     }
-}
+}
```
dschmitz89:
```suggestion
}

```

WarrenWeckesser:
@dschmitz89, I'm taking a look at this.  First impressions: looks good.  But I haven't turned on nitpick mode yet. :smiley:

Quick question: other than preserving the existing questionable behavior, can you see any reason for `pdtrik(1.0, m)` to return `nan`, when `inf` seems much more natural?

dschmitz89:
> @dschmitz89, I'm taking a look at this. First impressions: looks good. But I haven't turned on nitpick mode yet. 😃
> 

Looking forward to it ;).

> Quick question: other than preserving the existing questionable behavior, can you see any reason for `pdtrik(1.0, m)` to return `nan`, when `inf` seems much more natural?

Nope, I agree that `inf` makes a lot more sense. At the same time, I think that no user should rely on calculating the inverses of CDFs in the tails at 0 or 1. It just does not make sense. Boost for example does not even allow that.

ilayn:
Didn't we move everything to xsf already? I'm surprised to see that we still have a C file and its header in the repo. I thought these were all C++ified later on by Albert and Irwin.

dschmitz89:
They translated cephes and some other (substantial) parts of special but there is still some C and Cython left over. Especially cdflib was not touched. Once all of it is migrated to boost or cephes, we will not need it anymore at all.

steppi:
I actually did translate cdflib on a branch, but my impression was the quality was so low that it would be better to just do it from scratch. This was a proof of concept, https://github.com/scipy/scipy/pull/21454, and I could have carried on doing the whole cdflib in this fashion, but ended up not having bandwidth for it. If Boost devs want to maintain gpu capable C++ implementations of these cdf functions then I think that's a positive.

ilayn:
Ah OK. My bad, sorry for the noise. I don't have too much of an overview anymore hence my ignorant questions.

scipy/special/tests/test_cdflib.py
```diff
@@ -844,3 +843,38 @@ def test_chndtr_edge_cases(self, x, df, nc, expected):
...
+    def test_small_probabilities(self):
+        # Edge case: m = 0 or very small.
+        k = sp.pdtrik([[0], [0.25], [0.95]], [0, 1e-20, 1e-6])
+        assert_array_equal(k, np.zeros((3, 3)))
+
+    def test_roundtrip_against_pdtr(self):
+        m = [10, 50, 500]
+        k = 5
+        p = sp.pdtr(k, m)
+        assert_allclose(sp.pdtrik(p, m), k, atol=1.5e-7, rtol=0)
```
WarrenWeckesser: The computed results are much more precise than suggested by `atol=1.5e-7`, and indeed we should expect (and test for) better precision than that.  `rtol=1e-15` works on my computer.
```suggestion
        assert_allclose(sp.pdtrik(p, m), k, rtol=1e-15)
```

scipy/special/tests/test_cdflib.py
```diff
@@ -844,3 +843,38 @@ def test_chndtr_edge_cases(self, x, df, nc, expected):
...
+
+    @pytest.mark.parametrize("p, m, k",
+                             [(1.8976107553682285e-40, 100, 2),
+                              (0.48670120172085135, 100, 99),
+                              (8.30383406699052e-69, 1000, 500),
+                              (2.252837804125894e-227, 100_000, 90_000)]
+                            )
```
WarrenWeckesser: Style nit: this exact formatting case is covered in PEP 8:
```suggestion
                              (2.252837804125894e-227, 100_000, 90_000)])
```

scipy/special/tests/test_cdflib.py
```diff
@@ -844,3 +843,38 @@ def test_chndtr_edge_cases(self, x, df, nc, expected):
 @pytest.mark.parametrize("x", [0.1, 100])
 def test_chdtriv_p_equals_1_returns_0(x):
     assert sp.chdtriv(1, x) == 0
+
+
+class TestPdtrik:
+    @pytest.mark.parametrize("p, m, expected",
+                             [(0, 0.5, 0),
+                              (0.5, 0, 0),
+                              (0, 0, 0)])
```
WarrenWeckesser: The case p=0, m=0 is covered in `test_small_probabilities()` below.
```suggestion
                              (0.5, 0, 0)])
```

WarrenWeckesser (changes requested):
Thanks @dschmitz89, just a few nits to pick...

dschmitz89:
A little offtopic: @steppi The`gdtrib` function you implemented in `xsf` is also basically just a reparametrization of boost's `gamma_p_inva`. If you prefer we could migrate that also. If you wish to keep it for GPU testing of your root finder etc, also fine for me.

WarrenWeckesser (approved):
:+1:

Issue #22370, opened 2025-01-20
RFC: Feature roadmap for 3D spatial transformations


scottshambaugh:
This is a top-level issue to propose and track a roadmap for 3D spatial transformations in scipy. ✅ indicates complete, and 🏗️ indicates in-work.

Working with spatial transformations is fundamental functionality that is critical to the fields of robotics, aerospace, mechanical engineering, computer graphics, and anything that has to do with observing, modeling, analyzing, or controlling the physical world in 3-D space. I think that this is a natural extension to the `scipy.spatial.transform` module that would be broadly useful to the scientific community over many domains.

This is my take at a roadmap, and feedback is encouraged. In particular, there is consensus around implementing Rigid Transformations, but not yet around Affine Transformations.

**Overall**
* The Rotation and RigidTransformation docs both include plotting functions to help visualize their transformations. We should make plotting helper functions that attach to the objects and make this easy, much like the ones that already exist in the top level [`scipy.spatial`](https://docs.scipy.org/doc/scipy/reference/spatial.html) namespace.
* There is opportunity for speedup via cython functions and faster algorithms throughout.

**Rotations**
* ✅ Already implemented, and IMO feature complete. https://scipy.github.io/devdocs/reference/generated/scipy.spatial.transform.Rotation.html

**Rigid Transformations**
Combination of rotations and translations, equivalent to the special Euclidean SE(3) group.

* 🏗️ Baseline implementation of `RigidTransformation` is in work, issue: https://github.com/scipy/scipy/issues/19254 and pull request: https://github.com/scipy/scipy/pull/22267
* Additional representations, see https://dfki-ric.github.io/pytransform3d/user_guide/transformations.html:
  - Screw matrix
  - Screw axis
  - Matrix logarithm
* `RigidTransformation.mean()` method. Similar to `Rotation.mean()`, to obtain a weighted average of all the 6D poses contained in one RigidTransformation object. This decomposes to a weighted average of the rotation component (already implemented) combined with a weighted arithmetic mean of the translations components.
* `RigidTranformation.rotate_about()` method. Applies a rotation about a point in 3D space.
* Syntactic sugar / operator overloading to allow for easier composition with Rotations and translations. Some discussion about options in https://github.com/scipy/scipy/issues/19254. My personal preference would be to use `__mul__` with Rotations and `__add__` with shape (N, 3) translation arrays. For example,
  - Before: `tf1 = Tf.from_rotation(r) * tf0 * Tf.from_translation(t)`
[...]

PR #23718, opened 2025-10-01
ENH: spatial: Implement `RigidTransform.mean()`

scottshambaugh:
#### Reference issue
N/A

Implements `RigidTransform.mean()`, to mirror the existing `Rotation.mean()`. Part of the spatial transformation roadmap: https://github.com/scipy/scipy/issues/22370

### DIFF
--- a/scipy/spatial/transform/_rigid_transform.py
+++ b/scipy/spatial/transform/_rigid_transform.py
@@ -1049,10 +1049,9 @@ class RigidTransform:
                [ 0.51801458,  0.13833531, -0.84411151,  0.52429339],
                [0., 0., 0., 1.]])
         """
-        backend = select_backend(self._xp, cython_compatible=False)
-        mean = backend.mean(self._matrix, weights=weights)
+        mean = self._backend.mean(self._matrix, weights=weights)
         return RigidTransform._from_raw_matrix(mean, xp=self._xp,
-                                               backend=backend)
+                                               backend=self._backend)
 
     @xp_capabilities(
         skip_backends=[("dask.array", "missing linalg.cross/det functions")]
--- a/scipy/spatial/transform/_rigid_transform.py
+++ b/scipy/spatial/transform/_rigid_transform.py
@@ -1016,8 +1016,8 @@ class RigidTransform:
         Returns
         -------
         mean : `RigidTransform` instance
-            Object containing the mean of the transforms in the current
-            instance.
+            Single transform containing the mean of the transforms in the
+            current instance.
 
         References
         ----------
@@ -1049,18 +1049,10 @@ class RigidTransform:
                [ 0.51801458,  0.13833531, -0.84411151,  0.52429339],
                [0., 0., 0., 1.]])
         """
-        t, r = self.as_components()
-        r_mean = r.mean(weights=weights)
-        xp = array_namespace(t)
-        if weights is None:
-            t_mean = xp.mean(t, axis=tuple(range(t.ndim - 1)))
-        else:
-            axis = tuple(range(t.ndim - 1))
-            norm = xp.sum(weights[..., None], axis=axis)
-            wsum = xp.sum(t * weights[..., None], axis=axis)
-            t_mean = wsum / norm
-        mean = RigidTransform.from_components(t_mean, r_mean)
-        return mean
+        backend = select_backend(self._xp, cython_compatible=False)
+        mean = backend.mean(self._matrix, weights=weights)
+        return RigidTransform._from_raw_matrix(mean, xp=self._xp,
+                                               backend=backend)
 
     @xp_capabilities(
         skip_backends=[("dask.array", "missing linalg.cross/det functions")]
--- a/scipy/spatial/transform/_rigid_transform.py
+++ b/scipy/spatial/transform/_rigid_transform.py
@@ -114,6 +114,7 @@ class RigidTransform:
     as_exp_coords
     as_dual_quat
     concatenate
+    mean
     apply
     inv
     identity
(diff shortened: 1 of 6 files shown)


scottshambaugh:
Test failures are unrelated.

lucascolley:
cc @amacati

scipy/spatial/transform/tests/test_rigid_transform.py
```diff
@@ -1011,6 +1011,34 @@ def test_input_validation(xp):
...
+        weights = None
+        t_mean = t
+    else:
+        weights = xp.asarray(rng.random(size=(ndim,) * (ndim - 1)))
+        axis = tuple(range(t.ndim - 1))
+        norm = xp.sum(weights[..., None], axis=axis)
+        wsum = xp.sum(t * weights[..., None], axis=axis)
+        t_mean = wsum/norm
+    r_mean = r.mean(weights=weights)
+    tf_mean = tf.mean(weights=weights)
```
amacati, line 1046: I think we should also test the non-weighted case at least once for ndim > 1.

scipy/spatial/transform/_rigid_transform.py
```diff
@@ -983,6 +984,84 @@ def concatenate(
...
+        t, r = self.as_components()
+        r_mean = r.mean(weights=weights)
+        xp = array_namespace(t)
+        if weights is None:
+            t_mean = xp.mean(t, axis=tuple(range(t.ndim - 1)))
+        else:
+            axis = tuple(range(t.ndim - 1))
+            norm = xp.sum(weights[..., None], axis=axis)
+            wsum = xp.sum(t * weights[..., None], axis=axis)
+            t_mean = wsum / norm
```
amacati: Can we put this into the xp backend and follow the style we have used for all other functions so far? If you do not want to write the cython code for it, you can set `select_backend(xp, cython_compatible=False)` by default. Although we could see speedups for a smaller number of transforms on numpy.

scipy/spatial/transform/_rigid_transform.py
```diff
@@ -983,6 +984,84 @@ def concatenate(
...
+        mean = RigidTransform.from_components(t_mean, r_mean)
+        return mean
```
amacati: If we return the full matrix from the backend, we could use `_from_raw_matrix` to skip additional normalization steps on lazy backends.

amacati (commented):
I left some comments that should be easy to fix. At least if we decide to skip the cython backend entirely for this one.

scottshambaugh:
@amacati when I was making updates, I noticed that Rotation with ndim>1 is not flattening its input when calculating `mean`. I think that's incorrect behavior for the default, though this will be customizable when we add in the `axis` arg.

```python
rng = np.random.default_rng(123)
q = rng.normal(size=(1, 2, 3, 4))
r = Rotation.from_quat(q)
print(r.shape)
print(r.mean().shape)

# (1, 2, 3)
# (1, 2)
```

scottshambaugh:
Waiting on https://github.com/scipy/scipy/pull/23729 to be merged for further progress, moving back to draft

scottshambaugh:
Hey @amacati, I addressed your comments and added a cython backend for this. I'm feeling much better oriented towards the array API & backend choices now, lmk if it looks good to you.

scipy/spatial/transform/tests/test_rigid_transform.py
```diff
@@ -1011,6 +1011,64 @@ def test_input_validation(xp):
...
+                    atol=atol)
+
+
+@make_xp_test_case(RigidTransform.mean)
+def test_mean_invalid_weights(xp):
+    tf = RigidTransform.from_matrix(xp.tile(xp.eye(4), (4, 1, 1)))
+    if is_lazy_array(tf.as_matrix()):
+        m = tf.mean(weights=-xp.ones(4))
+        # Only the rotation submatrix will be nan
+        assert xp.all(xp.isnan(m.as_matrix()[:3, :3]))
```
scottshambaugh: I'm not totally clear on what the behavior here *should* be when we're working with a lazy array that bypasses the ValueError. Is this nan subset fine?

scottshambaugh:
CI failure unrelated

scipy/spatial/transform/_rigid_transform_cy.pyx
```diff
@@ -243,6 +244,36 @@ def pow(double[:, :, :] matrix, float n):
...
 
+@cython.embedsignature(True)
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def mean(double[:, :, :] matrix, weights=None):
+    if matrix.shape[0] == 0:
+        raise ValueError("Mean of an empty transform set is undefined.")
+    
+    quat = as_quat(from_rot_matrix(matrix[:, :3, :3]))
+    t = np.asarray(matrix[:, :3, 3])
```
amacati, line 255: Unrelated to this PR, more a general comment and food for thought: We use MemoryViews as recommended by cython, but I suspect that we do a lot of unnecessary allocations as a result when converting them to numpy arrays again. For small data sizes, these allocations dominate performance. Is there a way to guarantee that numpy does not allocate? We know that we own the buffers that are passed to the backend, hence reusing it should not have any side effects.

scipy/spatial/transform/_rigid_transform_xp.py
```diff
@@ -245,6 +247,30 @@ def pow(matrix: Array, n: float | Array) -> Array:
     return from_exp_coords(as_exp_coords(matrix) * n)
 
 
+def mean(matrix: Array, weights: ArrayLike | None = None) -> Array:
+    xp = array_namespace(matrix)
+    if matrix.shape[0] == 0:
+        raise ValueError("Mean of an empty rotation set is undefined.")
+
+    quats = quat_from_matrix(matrix[..., :3, :3])
```
amacati, line 256: More food for thoughts on optimizations we could do: At the moment, every call to `from_matrix` is checking if the matrix is a valid rotation matrix, and normalizes the matrix for lazy frameworks. In cases like this where we call the function from within our own backend, the matrix is guaranteed to be valid and there is no need to normalize again. We could have a `from_matrix_unsafe` function that implements only the conversion from a valid matrix to a quaternion, and then only normalize inputs from outside the backend.

Not relevant for this PR, but it could significantly improve performance.

scipy/spatial/transform/_rotation_cy.pyx
```diff
@@ -953,16 +953,15 @@ def mean(double[:, :] quat, weights=None):
...
-            raise ValueError("Expected `weights` to be 1 dimensional, got "
-                                "shape {}.".format(weights.shape))
-        if weights.shape[0] != quat.shape[0]:
-            raise ValueError("Expected `weights` to have number of values "
-                                "equal to number of rotations, got "
-                                "{} values and {} rotations.".format(
-                                weights.shape[0], quat.shape[0]))
         if np.any(weights < 0):
             raise ValueError("`weights` must be non-negative.")
+        if weights.ndim != 1:
```
amacati, line 958: Any reason for the shift? If so, it should maybe get a comment, wasn't clear to me at least.

scipy/spatial/transform/tests/test_rigid_transform.py
```diff
@@ -1011,6 +1011,65 @@ def test_input_validation(xp):
...
+@make_xp_test_case(RigidTransform.mean)
+@pytest.mark.parametrize("ndim", range(1, 4))
+def test_mean(xp, ndim: int):
+    atol = 1e-12
+    rng = np.random.default_rng(123)
+
+    dtype = xpx.default_dtype(xp)
+    t = xp.asarray(rng.normal(size=(ndim,) * (ndim - 1) + (3,)), dtype=dtype)
+    q = xp.asarray(rng.normal(size=(ndim,) * (ndim - 1) + (4,)), dtype=dtype)
+    r = rotation_to_xp(Rotation.from_quat(q), xp=xp)
```
amacati: Should not be necessary since Rotation was already created from an xp array of correct dtype.

scipy/spatial/transform/tests/test_rigid_transform.py
amacati: Not sure either. I think I set the whole matrix to NaNs in a similar case in the rotation backend. But both choices seem ok. We should decide this once and then try to be consistent throughout.

scipy/spatial/transform/_rigid_transform_xp.py
```diff
@@ -245,6 +247,30 @@ def pow(matrix: Array, n: float | Array) -> Array:
...
+    if weights is None:
+        quats_mean = quat_mean(quats)
+    else:
+        quats_mean = quat_mean(quats, weights=weights)
+    r_mean = quat_as_matrix(quats_mean)
```
amacati, line 270: We're implicitly relying on the check https://github.com/scipy/scipy/blob/e407bc4d6ee71ec1a23fce9d95b58e28451e4d94/scipy/spatial/transform/_rotation_xp.py#L496-L500 here. Are we fine with this? We could also relax that check further such that `weights` only needs to be broadcastable to `quat` / `matrix`.

amacati (approved):
Only one unnecessary `rotation_to_xp`, looks good otherwise. The other comments are related to optimizations we could incorporate in the future for speedups.

scipy/spatial/transform/_rotation_cy.pyx
scottshambaugh: It's to match the order that the checks are performed in `_rotation_cy.pyx:mean`, figure it doesn't matter too much

scipy/spatial/transform/_rigid_transform_xp.py
scottshambaugh: Agreed, I like this idea quite a bit

scipy/spatial/transform/_rigid_transform_xp.py
scottshambaugh: Good call, added another check to make things more explicit and call out transforms in the error message

scipy/spatial/transform/tests/test_rigid_transform.py
scottshambaugh: Updated to make the whole matrix NaNs, it seems clearer to me

scottshambaugh:
CI looks good

scottshambaugh:
@lucascolley @amacati any more feedback on this one? I'd like to get this in and then we can add the `axis` kwarg in a follow-on PR.

amacati:
My comments were all addressed. We can think about the optimizations in a separate PR.

lucascolley (approved):
thanks both, LGTM!

amacati:
@scottshambaugh Do you plan to also add the axis argument in a follow-up? That would complete the list of planned features in #23391

scottshambaugh:
@amacati I won't have time to do that for another week, if you feel motivated in the meantime then go for it!

PR #23727, opened 2025-10-02
MAINT: Consistent ``spatial.[c]KDTree`` method parameter defaults for ``eps`` and ``p``

jorenham:
This was partially addressed in #23679 for certain `eps` defaults. But I noticed that there were other inconsistent `eps` defaults, as well as inconsistent `p`. These float parameters now have floats as defaults, and match their docstring signatures.

### DIFF
--- a/scipy/spatial/_ckdtree.pyx
+++ b/scipy/spatial/_ckdtree.pyx
@@ -661,11 +661,11 @@ cdef class cKDTree:
     # -----
 
     @cython.boundscheck(False)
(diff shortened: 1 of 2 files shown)


tylerjereddy (approved):
I read the full diff, this seems straightforward/fine, and I trust you for typing-related conventions. Fractional Minkowski p-norm values are indeed valid as long as $` p \geq 1`$.

The one failing CI job is unrelated, and we have a ticket open for that.

Thanks!

PR #23729, opened 2025-10-03
BUG: spatial: Fix shape of Rotation mean when ndim > 1

scottshambaugh:
#### Reference issue
N/A

When https://github.com/scipy/scipy/pull/23467 was merged, Rotations gained the ability to have shapes with ndim > 1. However the current implementation of `Rotation.mean()` only reduces along the last axis. This does not match the expected behavior of arrays without an `axis` parameter specified (not yet implemented for Rotations).

```python
import numpy as np
from scipy.spatial.transform import Rotation, RigidTransform

rng = np.random.default_rng(123)
q = rng.normal(size=(1, 2, 4))
r = Rotation.from_quat(q)
print(r.shape)
print(r.mean().shape)
# (1, 2)
# (1,)

a = q[..., 0]  # numpy array
print(a.shape)
print(a.mean().shape)
# (1, 2)
# ()
```

This PR updates it so that a single rotation is always returned.
```python
print(r.mean().shape)
# ()
```

#### Additional information
The bug is not yet present in a public release, and can be fixed prior to 1.17.0 going out.

@amacati could you take a look?

### DIFF
--- a/scipy/spatial/transform/_rotation.py
+++ b/scipy/spatial/transform/_rotation.py
@@ -82,8 +82,8 @@ class Rotation:
 
     A `Rotation` instance can contain a single rotation transform or rotations of
     multiple leading dimensions. E.g., it is possible to have an N-dimensional array of
-    (N, M, K) rotations. When applied to other rotations or vectors, standard broadcasting
-    rules apply.
+    (N, M, K) rotations. When applied to other rotations or vectors, standard
+    broadcasting rules apply.
 
     Indexing within a rotation is supported to access a subset of the rotations stored
     in a `Rotation` instance.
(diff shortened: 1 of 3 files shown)


scipy/spatial/transform/_rotation.py
```diff
@@ -2035,8 +2035,10 @@ def mean(self, weights: ArrayLike | None = None) -> Rotation:
         >>> r.mean().as_euler('zyx', degrees=True)
         array([0.24945696, 0.25054542, 0.24945696])
         """
-        mean = self._backend.mean(self._quat, weights=weights)
-        return Rotation._from_raw_quat(mean, xp=self._xp, backend=self._backend)
+        cython_compatible = self._quat.ndim < 3
+        backend = select_backend(self._xp, cython_compatible=cython_compatible)
+        mean = backend.mean(self._quat, weights=weights)
+        return Rotation._from_raw_quat(mean, xp=self._xp, backend=backend)
```
amacati: We don't need to re-determine the backend again, `self._backend` already accounts for `self._quat`'s dimension. We only need to double-check if we interact with new inputs of unknown dimensions (i.e. when multiplying with another quat).

amacati (changes requested):
Nice spot. This looks good except for the backend re-select.

scottshambaugh:
Thanks for the quick look! Updated.

scottshambaugh:
CI failures unrelated

amacati (approved):
Looks good to me with the recent changes

scottshambaugh:
Rebased to main

PR #23734, opened 2025-10-05
MAINT: stats.percentileofscore: raise with multidimensional input

### LINKED ISSUE #21563: ENH: stats.percentileofscore: 2D array input should be fixed or forbidden

[percentileofscore](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.percentileofscore.html) is documented to operate on 1D arrays. If one uses a 2D array for the `a` parameter, bizarre behavior happens.

```
import scipy
import numpy as np

a = np.array([
    [1, 2, 3, 4, 5],
])
print(scipy.stats.percentileofscore(a, 5))
print(scipy.stats.percentileofscore(a.T, 5))
```

Output:

```
[500.]
[20. 20. 20. 20. 20.]
```

The first behavior does not make sense, because 500% is not a percentile. The second behavior does not make sense, because 5 ought to be at the 100% percentile, and I only provided one input, not five.

This function is documented to only accept 1D arrays. Ideally it would do something sensible, but if that's not possible, the next best thing would be to forbid 2D input to the `a` parameter to this function. (Would this need to be phased in with a deprecation?)

It would be best if percentileofscore returned a sensible result in this situation. I am not sure what a sensible result is if the `a` parameter is 2D.

It may also be good to forbid 2D input to `score`, while we're here.

This originally occurred to a user on [StackOverflow](https://stackoverflow.com/questions/78967500/why-would-scipy-stats-percentileofscore-return-an-array-instead-of-a-scaler-begi).

mdhaber:
#### Reference issue
Closes gh-21563

gh-21563 reported that `percentileofscore` produced nonsense results with multidimensional input.
This PR makes `percentileofscore` raise with multidimensional input.

#### Additional information
gh-21563 suggested that `percentileofscore` could be extended to accept multidimensional arrays. Indeed, it could, but there haven't been any moves toward that since the issue was reported, so let's fix the bug.

Also, perhaps the names made it difficult to see, but `stats.percentileofscore` does essentially the same thing as  `mstats.plotting_positions`, `mstats.meppf`, and `stats.ecdf` - which is essentially the inverse of `stats.scoreatpercentile`, `mstats.scoreatpercentile`, `mstats.mquantile`, and `stats.quantile`. I mention it because it seems these were developed in isolation, leading to a lot of redundancy.

It would be better to consider the whole picture at once. gh-22194 tracks the need for a function like this that supports multidimensional input (superseding `mstats.plotting_positions` as `stats.quantile` supersedes `mstats.mquantile`) and native support for `marray` and other array API arrays.

### DIFF
--- a/scipy/stats/_stats_py.py
+++ b/scipy/stats/_stats_py.py
@@ -2102,33 +2102,34 @@ def percentileofscore(a, score, kind='rank', nan_policy='propagate'):
     ----------
     a : array_like
         A 1-D array to which `score` is compared.
-    score : array_like
-        Scores to compute percentiles for.
+    score : float or array_like
+        A float score or array of scores for which to compute the percentile(s).
     kind : {'rank', 'weak', 'strict', 'mean'}, optional
         Specifies the interpretation of the resulting score.
         The following options are available (default is 'rank'):
 
-          * 'rank': Average percentage ranking of score.  In case of multiple
-            matches, average the percentage rankings of all matching scores.
-          * 'weak': This kind corresponds to the definition of a cumulative
-            distribution function.  A percentileofscore of 80% means that 80%
-            of values are less than or equal to the provided score.
-          * 'strict': Similar to "weak", except that only values that are
-            strictly less than the given score are counted.
-          * 'mean': The average of the "weak" and "strict" scores, often used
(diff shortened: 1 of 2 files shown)


j-bowhay (approved):
Looks like there is consensus in the original issue and the changes seems reasonable to me

PR #21872, opened 2024-11-12
MAINT: remove `np.random.seed` from `scipy.optimize`

andyfaff:
Removes instances of `np.random.seed` from `optimize`, and removes as many usages of the global `RandomState` as I can find. The usages of `RandomState` are modified to use Generators.

PR #23736, opened 2025-10-05
ENH: optimize.linprog: remove use of `np.random.seed` from tests

mdhaber:
#### Reference issue
Supplement to gh-21872

gh-21872 removed most use of `np.random.seed` and `np.random.RandomState` from `scipy.optimize` (aside from those needed during the SPEC 7 transition). One exception was `linprog` tests; this removes uses in `test_linprog.py`.

#### Additional information
I probably could have used variables to store reference values for some tests and used them throughout, but I don't think it's too important. Let's just make the minimal changes here.


scipy/optimize/tests/test_linprog.py
```diff
@@ -383,7 +383,7 @@ def test_disp(self):
         A, b, c = lpgen_2d(20, 20)
         res = linprog(c, A_ub=A, b_ub=b, method=self.method,
                       options={"disp": True})
-        _assert_success(res, desired_fun=-64.049494229)
+        _assert_success(res, desired_fun=-63.47967608020187)  # method='highs' solution
```
mdhaber, line 386: I don't remember whether there was a source for these values. In any case, when the tests were written, 'highs'` didn't exist, so I think it's reasonable to use it as a source of ground truth for all the other methods. And if all the other methods agree with it, it's probably right, too.

scipy/optimize/tests/test_linprog.py
```diff
@@ -1814,7 +1814,7 @@ def f(options):
             f(options=options)
 
     def test_crossover(self):
```
mdhaber, line 1816: @jajhall can you suggest a problem for which you'd expect HiGHS IPM to always do crossover iterations? I thought they were typically needed for this magic square problem generator, but maybe not.

jajhall: This small QAP linearisation [qap04.mps](https://github.com/user-attachments/files/22707188/qap04.mps.txt) solves in very little time and requires 14 crossover iterations

mdhaber: Thanks. These magic square problems (relaxation of BILP for creating a magic square) were pretty degenerate, I thought, but I guess not enough. Thanks!

mdhaber:
Marking this as draft until we figure out a problem for `test_crossover`. Is adding a 33kb file a dealbreaker?

jajhall:
> Marking this as draft until we figure out a problem for `test_crossover`. Is adding a 33kb file a dealbreaker?

The problem can be formulated in code if 33kb is a lot!

mdhaber:
Thanks @jajhall. Looks like I'm all set; I found a seed for the magic cube problem generator that works. @andyfaff I think this is ready. Failures are all in main.

andyfaff:
Thank you @mdhaber for finding a way.

I know It feels kind of icky to get things working by fishing around for a seed that gets things to work. Either the test needs redesigning, or the code does. But for now this is good enough.

mdhaber:
> I know It feels kind of icky to get things working

Not necessarily, actually. In this case - and in many cases - the behavior being tested is only supposed to happen sometimes. The random problem generator is just a convenient way of triggering the behavior so we can see that it _can_ happen sometimes. If the implementation were wrong, it would probably _never_ happen. With much more effort, it might be possible to generate a lot of test cases in which you know it definitely should happen, but I don't think it's worth it.

Issue #22831, opened 2025-04-13
BUG: Sparse Interpolator (LinearNDInterpolator and CloughTocher2DInterpolator) return NaN on the points, which is linear combination of triangle point


serhiy-yevtushenko:
### Describe your issue.

I was trying to build extrapolator, which will be returning the value of the nearest point on the sparse interpolators Convex Hull, and while doing it, I have found, that sparse points interpolators (LinearNDInterpolator, CloughTocher2DInterpolator) return in some cases NaN for the point, that is the linear combination of triangle points (i.e., which lays directly on the border)

The bug is reproducible on versions 1.15.1 and 1.15.2

### Reproducing Code Example

```python
import pytest
import scipy
import numpy as np
from scipy.interpolate import (
    CloughTocher2DInterpolator,
    LinearNDInterpolator,
)


@pytest.mark.parametrize("interpolator_factory", [CloughTocher2DInterpolator, LinearNDInterpolator])
def test_reproduction_NaN_on_points_linear_combination(interpolator_factory):
    print(f'{scipy.__version__}')
    inputs = np.array([[ 0.       ,  0.       ],
                      [ 0.       ,  0.008016],
                      [ 3.       , -2.       ]])
    values = np.ones(len(inputs))
    # Create the interpolator
    interpolator = interpolator_factory(inputs, values)
    t1 = 0.6
    p1 = inputs[0]
    p2 = inputs[2]
    point_on_edge = p1 + t1 * (p2 - p1)
    print(f'{p1=} {p2=} {t1=} {point_on_edge=} {interpolator(*point_on_edge)=}')
    assert not np.isnan(interpolator(*point_on_edge))
```

### Error message

```shell
The interpolator return NaN
```

### SciPy/NumPy/Python version and system information

```shell
1.15.2 2.2.2 sys.version_info(major=3, minor=11, micro=4, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    detection method: pkgconfig
    found: true
    include directory: C:/Users/runneradmin/AppData/Local/Temp/cibw-run-chm4s13i/cp311-win_amd64/build/venv/Lib/site-packages/scipy_openblas32/include
    lib directory: C:/Users/runneradmin/AppData/Local/Temp/cibw-run-chm4s13i/cp311-win_amd64/build/venv/Lib/site-packages/scipy_openblas32/lib
    name: scipy-openblas
[...]

Issue #21279, opened 2024-07-28
BUG: qhull incorrect tests if the point is inside convex hull


gf0507033:
### Describe your issue.

`scipy.interpolate.LinearNDInterpolator` should be defined on each input point. Apparently for some input points that's not true. The reason is wrong calculation inside qhull if point is inside convex hull of input points or not.

Sample data for reproducing error are [here](https://gist.github.com/gf0507033/2cfba1709e68c6422d8bcb206d957975).

### Reproducing Code Example

```python
import json

from scipy.interpolate import LinearNDInterpolator

import numpy as np

with open('bug.json', 'r') as f:
    points2d = np.array(json.load(f))    

r = np.ones_like(points2d[:, 0])
r_predictor = LinearNDInterpolator(points2d, r)
assert (~np.isnan(r_predictor(points2d))).all()
```

### Error message

```shell
AssertionError                            Traceback (most recent call last)
File /home/grigory/bug.py:12
     10 r = np.ones_like(points2d[:, 0])
     11 r_predictor = LinearNDInterpolator(points2d, r)
---> 12 assert (~np.isnan(r_predictor(points2d))).all()
```

### SciPy/NumPy/Python version and system information

```shell
1.10.1 1.26.4 sys.version_info(major=3, minor=10, micro=14, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    detection method: cmake
    found: true
    include directory: unknown
    lib directory: unknown
    name: OpenBLAS
    openblas configuration: unknown
    pc file directory: unknown
    version: 0.3.18
  lapack:
    detection method: cmake
    found: true
    include directory: unknown
    lib directory: unknown
    name: OpenBLAS
    openblas configuration: unknown
    pc file directory: unknown
    version: 0.3.18
Compilers:
  c:
    commands: cc
    linker: ld.bfd
    name: gcc
    version: 10.2.1
  c++:
    commands: c++
    linker: ld.bfd
    name: gcc
    version: 10.2.1
  cython:
    commands: cython
    linker: cython
    name: cython
    version: 0.29.33
  fortran:
    commands: gfortran
    linker: ld.bfd
    name: gcc
    version: 10.2.1
  pythran:
    include directory: /tmp/pip-build-env-xob7rtip/overlay/lib/python3.10/site-packages/pythran
    version: 0.12.1
Machine Information:
  build:
    cpu: x86_64
    endian: little
    family: x86_64
    system: linux
  cross-compiled: false
  host:
[...]

PR #21288, opened 2024-07-31
WIP, BUG: `_find_simplex_directed` brute force

tylerjereddy:
* Fixes gh-21279

* In the event that a point is judged to be outside of a simplex in `_find_simplex_directed`, for now it appears to be more robust to fall back to a brute force check on the matter.

[skip circle] [skip cirrus]

Things that could be better here:
- Instead of adding an `.npy` file, search for a smaller reproducing set of generators
- Do some "whiteboard analysis" to find a faster solution than falling back to brute force (though correct > fast for now, I suppose)
- Could test directly in `spatial` via the public `find_simplex()`, though adding a test that exactly mimics the ticket reproducer is perhaps not so terrible.

PR #23737, opened 2025-10-05
ENH: Allow users to increase the QHull simplex assignment tolerance

### LINKED ISSUE #21910: BUG: interpolate.griddata: NaNs inside convex hull (despite no NaNs or duplicates in data)

DWesl:
#### Reference issue

Closes #21910
Replaces #23285

Filling an area from a dense boundary produces holes.  Increasing the tolerance QHull uses to determine whether each test point is in a given simplex allows QHull to assign a simplex to each test point, avoiding those holes.

This PR adds an argument to the `__call__` method of LinearNDInterpolater and CloughTocher2dInterpolator to allow users to increase the tolerance in cases like this.  It also exposes that argument through the `griddata` interface.


scipy/interpolate/tests/test_interpnd.py
```diff
@@ -216,6 +216,27 @@ def do_interp(interpolator, slice_rows, slice_cols):
         worker_thread_3.join()
         worker_thread_4.join()
 
+    @pytest.mark.parametrize("simplex_tolerance", [1, 10])
+    def test_interp_from_boundary(self, simplex_tolerance):
```
ev-br:
```suggestion
    def test_interp_from_boundary(self, simplex_tolerance):
         # regression test for https://github.com/scipy/scipy/issues/21910
```

scipy/interpolate/_interpnd.pyx
```diff
@@ -337,8 +340,8 @@ class LinearNDInterpolator(NDInterpolatorBase):
         nvalues = out.shape[1]
 
         start = 0
-        eps = 100 * DBL_EPSILON
-        eps_broad = sqrt(DBL_EPSILON)
+        eps = 100 * DBL_EPSILON * simplex_tolerance
+        eps_broad = sqrt(DBL_EPSILON) * simplex_tolerance
```
ev-br: `simplex_tolerance` or `sqrt(simplex_tolerance`? AFAIU, either way it's a heuristics; still, is there a reason to prefer one or another? Here and in `CloughTocher2DInterpolator` below.

scipy/interpolate/_ndgriddata.py
```diff
@@ -214,6 +214,9 @@ def griddata(points, values, xi, method='linear', fill_value=np.nan,
         incommensurable units and differ by many orders of magnitude.
 
         .. versionadded:: 0.14.0
+    simplex_tolerance : float
```
ev-br:
```suggestion
    simplex_tolerance : float, optional
```

scipy/interpolate/_ndgriddata.py
```diff
@@ -214,6 +214,9 @@ def griddata(points, values, xi, method='linear', fill_value=np.nan,
...
+        Multiplier for the default tolerance QHull uses to assign
+        a simplex to the xi.
```
ev-br: Please add `Default is 1.0` or some such. Also, this is a multiplier for the `qhull.find_simplex` tolerance, correct? Then we probably need to say this, add a link to `find_simplex`, and maybe state the `find_simplex` tolerance is `100 * epsilon * simplex_tolerance`.

Also a new argument needs a `.. versionadded:: 1.17.0` admonition

scipy/interpolate/src/__fitpack.cc
```diff
@@ -1,3 +1,5 @@
+#include <Python.h>
+/* __fitpack.h includes Python.h */
```
ev-br, line 2: This is a good change, but unrelated --- could you please submit it and its matching addition in `__fitpack.h` as a separate PR?

scipy/interpolate/tests/test_bsplines.py
```diff
@@ -695,6 +696,11 @@ def worker_fn(_, b):
         _run_concurrent_barrier(10, worker_fn, b)
 
 
+    @pytest.mark.xfail(
+        sys.platform == "cygwin",
+        reason="threading.get_native_id not implemented",
+        raises=AttributeError
+    )
```
ev-br, line 699: If it's needed, could you also separate it into its own PR (can be the same PR as `Python.h` additions, just a separate commit).

ev-br (changes requested):
Looks very reasonable!
Several small requests below: the main one is whether `eps_broad` should be use the new multiplier or its square root.

scipy/interpolate/src/__fitpack.cc
DWesl: Those commits are now #23743

Would rebase or revert work better for removing those commits from this PR?

DWesl:
I meant to ask, do you know why the `_do_evaluate` methods are `def` rather than `cdef`?  `cdef` methods would be hard to override from python, but would be slightly faster.  The dispatching goes through `_evaluate_double` and `_evaluate_complex` anyway, so there's a workaround there, but the speedup would only be noticeable if someone is calling the interpolator in an inner loop somewhere, and the vectorization takes care of the obvious cases already.

Changing the `_do_evaluate` methods from def to cdef would change the API, so should probably not go into this PR.

scipy/interpolate/_interpnd.pyx
DWesl: I have no idea which of the above and below would be better:
```suggestion
        eps_broad = sqrt(DBL_EPSILON * simplex_tolerance)
```
Unfortunately, the [`find_simplex` documentation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.Delaunay.find_simplex.html#scipy.spatial.Delaunay.find_simplex) doesn't mention what `eps_broad` is supposed to do.

It's not quite `sqrt(eps)`, which argues for the current form, but things might well work without increasing `eps_broad` at all, or we could change it to `sqrt(eps)` let `simplex_tolerance` propagate through that way.

EDIT: `eps_broad` gets used [here](https://github.com/scipy/scipy/blob/2ee4675fb1d1cfe70828cfc287e6078b33d8be72/scipy/spatial/_qhull.pyx#L1361).  It looks like the outside-visible function that calls `_find_simplex` does so [here](https://github.com/scipy/scipy/blob/2ee4675fb1d1cfe70828cfc287e6078b33d8be72/scipy/spatial/_qhull.pyx#L2148), and sets `eps_broad = sqrt(eps)` [a few lines earlier](https://github.com/scipy/scipy/blob/2ee4675fb1d1cfe70828cfc287e6078b33d8be72/scipy/spatial/_qhull.pyx#L2132)

ev-br: Exactly. So IOW the question is whether here we're scaling the factor of 100 in `tol` or `finfo(double).eps`. I have to admit I don't have an intuition for which one it is,  and I'm ready to roll with a conclusion from a domain expert. You probably looked at these things more than anyone else over last several years at least; in your experiments in https://github.com/scipy/scipy/issues/21910, which of the conditions was active, `eps` or `eps_broad`?
for viz, cc [paulbrodersen](https://github.com/paulbrodersen) as the OP and @tylerjereddy as the spatial expert

scipy/interpolate/src/__fitpack.cc
ev-br: Either way is likely fine, I'd say.
My personal workflow would be to rebase, but feel free to use what's more convenient to you.

Longer version: the main requirement is that commits that land in main are self-contained. This PR has already 7 commits including a merge from main, so it'll need to be squash-merged. This whatever allows the squash-merge is fine.

scipy/interpolate/tests/test_interpnd.py
```diff
@@ -216,6 +216,28 @@ def do_interp(interpolator, slice_rows, slice_cols):
...
+         # regression test for https://github.com/scipy/scipy/issues/21910
+        coords = np.block(
+            [[0, np.arange(400, 500), np.zeros(100), 800],
+             [0, np.zeros(100), np.arange(400, 500), 800]]
+        )
+        values = np.ones(202)
+        interpolator = interpnd.LinearNDInterpolator(coords.T, values)
```
ev-br: A very minor ask if you're updating the PR (might do as a follow-up as well): would be nice to `@pytest.parametrize("interpolator_cls", [LinearNDInterpolator, CloughTocher2DInterpolator])` the test, too.

ev-br (commented):
> I meant to ask, do you know why the _do_evaluate methods are def rather than cdef? cdef methods would be hard to override from python, but would be slightly faster. The dispatching goes through _evaluate_double and _evaluate_complex anyway, so there's a workaround there, but the speedup would only be noticeable if someone is calling the interpolator in an inner loop somewhere, and the vectorization takes care of the obvious cases already.

No idea TBH. Might be worth checking if tests rely on them being `def`, and measuring the perf change from making them `cdef` or `cpdef`, yes. In a follow-up PR though.

> Changing the _do_evaluate methods from def to cdef would change the API, so should probably not go into this PR.

Since `_do_evaluate` is private (has a leading underscore), it won't change the official API. There is some chance that somebody is reaching out to them, yes. Then we'll need to weight the performance improvement "for everybody" with how bad the breakage is for this somebody. Thus yes, a separate PR.

------------------------

One other follow-up is a DOC update (https://github.com/scipy/scipy/pull/23285#pullrequestreview-2990743293) to discuss this new `simplex_tolerance` argument. It'd be best to have an illustration, via a worked example where non-default simplex_tolerance helps. One of MREs from gh-23285 or what you discussed in https://github.com/scipy/scipy/pull/23285 can work, I suppose. But that also can be a follow-up PR.

scipy/interpolate/tests/test_interpnd.py
DWesl: Should I try combining the other tests for LinearNDInterpolator and CloughTocher2DInterpolator as well?  There's a few specific to one or the other, but several would seem to apply to both.

scipy/interpolate/src/__fitpack.cc
DWesl: Commits removed in squash-rebase

scipy/interpolate/tests/test_interpnd.py
ev-br: In a separate PR, yes please

ev-br:
cross-ref https://github.com/scipy/scipy/issues/21279 -- I wonder if this fix also fixes that other issue.

DWesl:
I can avoid the assertion error in the example code for #21279 by passing `simplex_tolerance=10`, so it would seem so.  This PR would then be complementary to #21288

scipy/interpolate/_interpnd.pyx
DWesl: In the previous experiments I think I multiplied both by 1e6 or something, though I suspect 100 would work fine

DWesl:
#22831 also seems related

scipy/interpolate/_ndgriddata.py
```diff
@@ -214,6 +214,11 @@ def griddata(points, values, xi, method='linear', fill_value=np.nan,
         incommensurable units and differ by many orders of magnitude.
 
         .. versionadded:: 0.14.0
+    simplex_tolerance : float, optional
+        Multiplier for the default tolerance QHull uses to assign
+        a simplex to the xi.  Default is 1.0
+
+        .. versionadded:: 0.17.0
```
ev-br:
```suggestion
        .. versionadded:: 1.17.0
```

ev-br (commented):
My comments are all addressed. The only remaining minor point is that if we have two simplex-related tolerances, `eps` and `eps_broad`, where the second one is the square root of the first one; now, if we add a user-adjustable multiplier for the first one, should the second one have the same multiplier or its root, too? Superficially it feels like yes, but I don't have intuition here, and am ready to roll with any suggestion from people more versed in spatial geometry.
Either way, I believe this PR is an improvement over the status quo, as demonstrated by the new test. Thus I think we should merge this PR --- with or without the square root --- for SciPy 1.17.

scipy/interpolate/_interpnd.pyx
```diff
@@ -151,8 +151,8 @@ class NDInterpolatorBase:
         xi = np.ascontiguousarray(xi, dtype=np.float64)
         return self._scale_x(xi), interpolation_points_shape
 
-    def __call__(self, *args):
-        """
+    def __call__(self, *args, simplex_tolerance=1.0):
+        """\
```
tylerjereddy: is the `\` really needed here?

scipy/interpolate/_interpnd.pyx
```diff
@@ -163,13 +163,19 @@ class NDInterpolatorBase:
             Points where to interpolate data at.
             x1, x2, ... xn can be array-like of float with broadcastable shape.
             or x1 can be array-like of float with shape ``(..., ndim)``
+        simplex_tolerance: float, optional
```
tylerjereddy, line 166: Do you mind adding a check that `simplex_tolerance` is a positive value and raising if not? I just checked locally and when I feed `test_interp_from_boundary()` a `simplex_tolerance` of `-1` it seems we get stuck in an infinite loop, or at least some kind of pathological control flow that takes an unreasonable amount of time.

scipy/interpolate/_ndgriddata.py
```diff
@@ -214,6 +214,11 @@ def griddata(points, values, xi, method='linear', fill_value=np.nan,
         incommensurable units and differ by many orders of magnitude.
 
         .. versionadded:: 0.14.0
+    simplex_tolerance : float, optional
+        Multiplier for the default tolerance QHull uses to assign
+        a simplex to the xi.  Default is 1.0
+
+        .. versionadded:: 1.17.0
```
tylerjereddy: same here maybe re: > 0 guard

scipy/interpolate/_interpnd.pyx
tylerjereddy: I'm a bit concerned that the testing isn't sensitive to the code change. For example, I can revert the `eps_broad` change and the regression test still passes:

```diff
--- a/scipy/interpolate/_interpnd.pyx
+++ b/scipy/interpolate/_interpnd.pyx
@@ -344,7 +344,7 @@ class LinearNDInterpolator(NDInterpolatorBase):
 
         start = 0
         eps = 100 * DBL_EPSILON * simplex_tolerance
-        eps_broad = sqrt(DBL_EPSILON) * simplex_tolerance
+        eps_broad = sqrt(DBL_EPSILON)
```

It seems like the comment at https://github.com/scipy/scipy/pull/23737/files#r2412306868 is suggesting that setting `eps_broad` is getting overriden anyway? The test also still passes if I set `eps_broad = 0.0`.

If we can't justify the code change with a test, I might naively prefer to minimize the code change to what we need to make the regression test pass without a clear reason to do otherwise.

tylerjereddy:
Looking at this more closely in between conference things, it looks like `_find_simplex_bruteforce()` does have a check `(-eps_broad <= c[m] <= 1 + eps)` where I think `eps_broad` should be genuinely used in the algorithm. It may be quite painful to write a test that is sensitive to that though.

tylerjereddy:
In fact, in my hands on this branch, I can literally purge the entire degenerate simplex block that uses `eps_broad` (diff below the fold) and `spin test -j 12` passes. Yikes.

<details>

```diff
--- a/scipy/spatial/_qhull.pyx
+++ b/scipy/spatial/_qhull.pyx
@@ -1334,40 +1334,6 @@ cdef int _find_simplex_bruteforce(DelaunayInfo_t *d, double *c,
             inside = _barycentric_inside(d.ndim, transform, x, c, eps)
             if inside:
                 return isimplex
-        else:
-            # transform is invalid (nan, implying degenerate simplex)
-
-            # we replace this inside-check by a check of the neighbors
-            # with a larger epsilon
-
-            for k in range(d.ndim+1):
-                ineighbor = d.neighbors[(d.ndim+1)*isimplex + k]
-                if ineighbor == -1:
-                    continue
-
-                transform = d.transform + ineighbor*d.ndim*(d.ndim+1)
-                if transform[0] != transform[0]:
-                    # another bad simplex
-                    continue
-
-                _barycentric_coordinates(d.ndim, transform, x, c)
-
-                # Check that the point lies (almost) inside the
-                # neighbor simplex
-                inside = 1
-                for m in range(d.ndim+1):
-                    if d.neighbors[(d.ndim+1)*ineighbor + m] == isimplex:
-                        # allow extra leeway towards isimplex
-                        if not (-eps_broad <= c[m] <= 1 + eps):
-                            inside = 0
-                            break
-                    else:
-                        # normal check
-                        if not (-eps <= c[m] <= 1 + eps):
-                            inside = 0
-                            break
-                if inside:
-                    return ineighbor
     return -1
 
 cdef int _find_simplex_directed(DelaunayInfo_t *d, double *c,
```

</details>

scipy/interpolate/_interpnd.pyx
DWesl: It worked fine before, so likely not.

scipy/interpolate/_ndgriddata.py
DWesl: Should be caught by the test in `NDInterpolatorBase.__call__` where relevant, since neither `LinearNDInterpolator` nor `CloughTocher2DInterpolator` override that and both are called a few lines down, and `simplex_tolerance` isn't relevant otherwise, but I added it anyway, just in case.

scipy/interpolate/_interpnd.pyx
DWesl: I wrote the code assuming `simplex_tolerance` would be >= 1, but there might be uses for setting it to 0.1, so I wrote the check to raise if `simplex_tolerance <= 0`.

scipy/interpolate/_interpnd.pyx
```diff
@@ -986,8 +1003,8 @@ class CloughTocher2DInterpolator(NDInterpolatorBase):
         nvalues = out.shape[1]
 
         start = 0
-        eps = 100 * DBL_EPSILON
-        eps_broad = sqrt(eps)
+        eps = 100 * DBL_EPSILON * simplex_tolerance
+        eps_broad = sqrt(eps) * simplex_tolerance
```
DWesl: And also here?
```suggestion
        eps_broad = sqrt(eps)
```

scipy/interpolate/_interpnd.pyx
DWesl: So back to
```suggestion
        eps_broad = sqrt(DBL_EPSILON)
```
?

ev-br:
Close/reopen + merge main to flush the CI.

ev-br:
Looked at this again. I hope it's not too much to ask @DWesl to check if this fixes https://github.com/scipy/scipy/issues/22831 and https://github.com/scipy/scipy/issues/21279, too? 
Both seem to have MWEs. If these can be fixed with a `simplex_tolerance`, it'd be great to have their reproducers added as regression tests.
This will also hopefully help settling the `eps_broad` business. If all known issues can be fixed without touching `eps_broad` at all, then indeed we don't need it.

DWesl:
I somehow managed to get my dev machine into a state where NumPy refuses to import, so it'll be a bit before I can do anything in-depth.  I should be able to add the #22831 reproducer from the web interface, but the #21279 reproducer will likely take an actual editor.

DWesl:
I think I have both reproducers in the tests now: #21279 could reproduce with only four points, which simplified things to a point I know how to handle.

(Oddly, I had to use the JSON version of those points: if I just used the NumPy repr of those points, the test passed fine.)

The last two feel like they'd work as hypothesis tests, though I don't know how to constrain the inputs to actually work with a fixed tolerance.

scipy/interpolate/_interpnd.pyx
```diff
@@ -163,13 +163,30 @@ class NDInterpolatorBase:
...
             or x1 can be array-like of float with shape ``(..., ndim)``
+        simplex_tolerance: float, optional
+            Multiplier for the default tolerance QHull uses to assign
+            a simplex to the xi in :meth:`scipy.spatial.Delaunay.find_simplex`.
+            Default is 1.0.  Increase if there are difficulties assigning points
+            to simplexes; this is most reproducible with points exatly on the
+            border of a very oblique triangle.  Only relevant for linear and 2-D
+            cubic interpolation.
+
+            .. versionadded:: 1.17.0
```
ev-br:
```suggestion
            .. versionadded:: 1.18.0
```

scipy/interpolate/_ndgriddata.py
```diff
@@ -214,12 +214,26 @@ def griddata(points, values, xi, method='linear', fill_value=np.nan,
...
         .. versionadded:: 0.14.0
+    simplex_tolerance : float, optional
+        Multiplier for the default tolerance QHull uses to assign
+        a simplex to the xi.  Default is 1.0.  Increase if there are
+        difficulties assigning points to simplexes; this is most
+        reproducible with points exatly on the border of a very
+        oblique triangle.  Only relevant for linear and 2-D cubic
+        interpolation.
+
+        .. versionadded:: 1.17.0
```
ev-br:
```suggestion
        .. versionadded:: 1.18.0
```

ev-br (approved):
Tyler's request to remove non-essential changes, https://github.com/scipy/scipy/pull/23737#issuecomment-3509112500, has been implemented: the PR now adds a `simplex_tolerance` multiplier to `eps` and does not touch `eps_broad`; this makes my question about the `eps_broad` multiplier moot. Especially given that indeed that adding `simplex_tolerance` of the order of 3-5 fixes gh-21279 and gh-22831.

The code looks nice sand minimal; the API change is backwards compatible in that it only adds a new keyword argument with the backcompatible default. 

Let's merge this now. Closing three tickets with a single patch feels good.
Thank you @DWesl, thanks @tylerjereddy for the review.

Issue #14322, opened 2021-06-29
ENH: refactor RNG usage to use np.random.Generator


tupui:
I propose to work on refactoring `np.random.RandomState` usages in the code base to use the preferred `np.random.Generator` available starting from NumPy 1.17.

It was discussed in some issue (I cannot find it now) that we could start dropping NumPy 1.16 for the next release of SciPy.

This work would be a follow up of #13863 and #13778

rkern:
I'm changing the label to `enhancement` as this is a surprisingly large change to the end-user. Right now, people can and do get reproducible results by calling `np.random.seed(SOME_SEED)` at the top of their program and just calling `scipy` functions afterwards. As we have no global `Generator` (and generally do not wish to add one), simply replacing the behavior of `check_random_state()` to use `Generator` will break programs. Not just because the streams change between different versions of numpy/scipy but because the programs will no longer be reproducible from _run to run_.

tupui:
Indeed, it's something we will need to make sure to communicate. Thanks Robert. I have a feeling that I am collecting controversial issues 😅 Glad to take the shots.

andyfaff:
We had a good discussion on this exact topic, I can't remember if it was on the ML or an issue though. @rgommers might be able to remember. I had a quick search but couldn't find it.
What precise changes were you suggesting @tupui? 
From memory the discussion then was:
For new work use Generator. For existing code retain back compatible, reproducible from release to release if possible, unless there's a glaringbug. In practice this means we can't remove RandomState and replace with Generator, but we can support supplying Generators to random_state keywords (which are then processed by check_random_state).

rgommers:
> Not just because the streams change between different versions of numpy/scipy but because the programs will no longer be reproducible from _run to run_.

That's an important point, and I think one we haven't discussed before. Having this variability in the long run is a good thing - but it's a large change indeed. Now instead of a single `np.random.seed()` call, users have to correctly thread generators all throughout their code.

Something to look at is how this affect downstream libraries. Most of those will be in sync with SciPy regarding NumPy version requirements, but then we have the likes of scikit-learn which supports the lowest numpy version possible for any given Python version (so for py37, `numpy >= 1.14.6`). I _think_ this change shouldn't affect them, but it's worth double checking.

andyfaff:
> having this variability in the long run is a good thing

Do you mean stability/reproducibility/invariance in the long run?

rgommers:
[...]

PR #23739, opened 2025-10-06
MAINT: stats: remove use of `np.random.seed`

mdhaber:
#### Reference issue
Toward gh-14322

Removes uses of `np.random.seed` from `scipy.stats`. 

#### Additional information
There are two exceptions involving tests designed to check behavior with the `RandomState` singleton. These are marked with a comment "valid use of np.random.seed".

### DIFF
--- a/scipy/stats/tests/common_tests.py
+++ b/scipy/stats/tests/common_tests.py
@@ -197,12 +197,8 @@ def check_named_args(distfn, x, shape_args, defaults, meths):
 def check_random_state_property(distfn, args):
     # check the random_state attribute of a distribution *instance*
 
-    # This test fiddles with distfn.random_state. This breaks other tests,
-    # hence need to save it and then restore.
-    rndm = distfn.random_state
-
     # baseline: this relies on the global state
-    np.random.seed(1234)
+    np.random.seed(1234)  # valid use of np.random.seed
     distfn.random_state = None
     r0 = distfn.rvs(*args, size=8)
 
@@ -230,9 +226,6 @@ def check_random_state_property(distfn, args):
 
     # ... and that does not alter the instance-level random_state!
     npt.assert_equal(distfn.random_state.get_state(), orig_state)
-
-    # finally, restore the random_state
-    distfn.random_state = rndm
 
 
 def check_meth_dtype(distfn, arg, meths):
--- a/scipy/stats/tests/test_continuous_basic.py
+++ b/scipy/stats/tests/test_continuous_basic.py
@@ -587,7 +587,6 @@ def test_gh1320_regression():
 
 def test_method_of_moments():
     # example from https://en.wikipedia.org/wiki/Method_of_moments_(statistics)
-    np.random.seed(1234)
     x = [0, 0, 0, 0, 1]
     a = 1/5 - 2*np.sqrt(3)/5
     b = 1/5 + 2*np.sqrt(3)/5
--- a/scipy/stats/tests/test_discrete_distns.py
+++ b/scipy/stats/tests/test_discrete_distns.py
@@ -177,8 +177,8 @@ def test_issue_11134():
 
 
 def test_issue_7406():
-    np.random.seed(0)
-    assert_equal(binom.ppf(np.random.rand(10), 0, 0.5), 0)
+    rng = np.random.default_rng(4763112764)
+    assert_equal(binom.ppf(rng.random(10), 0, 0.5), 0)
 
     # Also check that endpoints (q=0, q=1) are correct
     assert_equal(binom.ppf(0, 0, 0.5), -1)
@@ -186,8 +186,9 @@ def test_issue_7406():
 
 
 def test_issue_5122():
+    rng = np.random.default_rng(8312492117)
     p = 0
-    n = np.random.randint(100, size=10)
+    n = rng.integers(100, size=10)
 
     x = 0
     ppf = binom.ppf(x, n, p)
@@ -397,17 +398,19 @@ class TestZipfian:
 
 
 class TestNCH:
-    np.random.seed(2)  # seeds 0 and 1 had some xl = xu; randint failed
-    shape = (2, 4, 3)
-    max_m = 100
-    m1 = np.random.randint(1, max_m, size=shape)    # red balls
-    m2 = np.random.randint(1, max_m, size=shape)    # white balls
-    N = m1 + m2                                     # total balls
-    n = randint.rvs(0, N, size=N.shape)             # number of draws
-    xl = np.maximum(0, n-m2)                        # lower bound of support
-    xu = np.minimum(n, m1)                          # upper bound of support
-    x = randint.rvs(xl, xu, size=xl.shape)
-    odds = np.random.rand(*x.shape)*2
+    def setup_method(self):
+        rng = np.random.default_rng(7162434334)
+        shape = (2, 4, 3)
+        max_m = 100
+        m1 = rng.integers(1, max_m, size=shape)  # red balls
+        m2 = rng.integers(1, max_m, size=shape)  # white balls
+        N = m1 + m2  # total balls
+        n = randint.rvs(0, N, size=N.shape, random_state=rng)  # number of draws
+        xl = np.maximum(0, n-m2)  # lower bound of support
+        xu = np.minimum(n, m1)  # upper bound of support
+        x = randint.rvs(xl, xu, size=xl.shape, random_state=rng)
+        odds = rng.random(x.shape)*2
+        self.x, self.N, self.m1, self.n, self.odds = x, N, m1, n, odds
 
     # test output is more readable when function names (strings) are passed
     @pytest.mark.parametrize('dist_name',
--- a/scipy/stats/tests/test_distributions.py
+++ b/scipy/stats/tests/test_distributions.py
@@ -77,6 +77,9 @@ def test_distributions_submodule():
 
 
 class TestVonMises:
+    def setup_method(self):
+        self.rng = np.random.default_rng(6320571663)
+
     @pytest.mark.parametrize('k', [0.1, 1, 101])
     @pytest.mark.parametrize('x', [0, 1, np.pi, 10, 100])
     def test_vonmises_periodic(self, k, x):
@@ -257,7 +260,7 @@ class TestVonMises:
         assert_allclose(dist.ppf(0.9), np.pi*0.8, rtol=1e-15)
         assert_allclose(dist.mean(), 0, atol=1e-15)
         assert_allclose(dist.expect(), 0, atol=1e-15)
-        assert np.all(np.abs(dist.rvs(size=10, random_state=1234)) <= np.pi)
+        assert np.all(np.abs(dist.rvs(size=10, random_state=self.rng)) <= np.pi)
 
     def test_vonmises_fit_equal_data(self):
         # When all data are equal, expect kappa = 1e16.
@@ -355,19 +358,19 @@ def test_support(dist):
 
 class TestRandInt:
     def setup_method(self):
-        np.random.seed(1234)
+        self.rng = np.random.default_rng(8826485737)
 
     def test_rvs(self):
-        vals = stats.randint.rvs(5, 30, size=100)
+        vals = stats.randint.rvs(5, 30, size=100, random_state=self.rng)
         assert_(np.all(vals < 30) & np.all(vals >= 5))
         assert_(len(vals) == 100)
-        vals = stats.randint.rvs(5, 30, size=(2, 50))
+        vals = stats.randint.rvs(5, 30, size=(2, 50), random_state=self.rng)
         assert_(np.shape(vals) == (2, 50))
         assert_(vals.dtype.char in typecodes['AllInteger'])
-        val = stats.randint.rvs(15, 46)
+        val = stats.randint.rvs(15, 46, random_state=self.rng)
         assert_((val >= 15) & (val < 46))
         assert_(isinstance(val, np.ScalarType), msg=repr(type(val)))
-        val = stats.randint(15, 46).rvs(3)
+        val = stats.randint(15, 46).rvs(3, random_state=self.rng)
         assert_(val.dtype.char in typecodes['AllInteger'])
 
     def test_pdf(self):
@@ -386,16 +389,16 @@ class TestRandInt:
 
 class TestBinom:
     def setup_method(self):
-        np.random.seed(1234)
+        self.rng = np.random.default_rng(1778595878)
 
     def test_rvs(self):
-        vals = stats.binom.rvs(10, 0.75, size=(2, 50))
+        vals = stats.binom.rvs(10, 0.75, size=(2, 50), random_state=self.rng)
         assert_(np.all(vals >= 0) & np.all(vals <= 10))
         assert_(np.shape(vals) == (2, 50))
         assert_(vals.dtype.char in typecodes['AllInteger'])
-        val = stats.binom.rvs(10, 0.75)
+        val = stats.binom.rvs(10, 0.75, random_state=self.rng)
         assert_(isinstance(val, int))
-        val = stats.binom(10, 0.75).rvs(3)
+        val = stats.binom(10, 0.75).rvs(3, random_state=self.rng)
         assert_(isinstance(val, np.ndarray))
         assert_(val.dtype.char in typecodes['AllInteger'])
 
@@ -469,16 +472,16 @@ class TestArcsine:
 
 class TestBernoulli:
     def setup_method(self):
-        np.random.seed(1234)
+        self.rng = np.random.default_rng(7836792223)
 
     def test_rvs(self):
-        vals = stats.bernoulli.rvs(0.75, size=(2, 50))
+        vals = stats.bernoulli.rvs(0.75, size=(2, 50), random_state=self.rng)
         assert_(np.all(vals >= 0) & np.all(vals <= 1))
         assert_(np.shape(vals) == (2, 50))
         assert_(vals.dtype.char in typecodes['AllInteger'])
-        val = stats.bernoulli.rvs(0.75)
+        val = stats.bernoulli.rvs(0.75, random_state=self.rng)
         assert_(isinstance(val, int))
-        val = stats.bernoulli(0.75).rvs(3)
+        val = stats.bernoulli(0.75).rvs(3, random_state=self.rng)
         assert_(isinstance(val, np.ndarray))
         assert_(val.dtype.char in typecodes['AllInteger'])
 
@@ -854,16 +857,16 @@ class TestCrystalBall:
 
 class TestNBinom:
     def setup_method(self):
-        np.random.seed(1234)
+        self.rng = np.random.default_rng(5861367021)
 
     def test_rvs(self):
-        vals = stats.nbinom.rvs(10, 0.75, size=(2, 50))
+        vals = stats.nbinom.rvs(10, 0.75, size=(2, 50), random_state=self.rng)
         assert_(np.all(vals >= 0))
         assert_(np.shape(vals) == (2, 50))
         assert_(vals.dtype.char in typecodes['AllInteger'])
-        val = stats.nbinom.rvs(10, 0.75)
+        val = stats.nbinom.rvs(10, 0.75, random_state=self.rng)
         assert_(isinstance(val, int))
-        val = stats.nbinom(10, 0.75).rvs(3)
+        val = stats.nbinom(10, 0.75).rvs(3, random_state=self.rng)
         assert_(isinstance(val, np.ndarray))
         assert_(val.dtype.char in typecodes['AllInteger'])
 
@@ -884,34 +887,34 @@ class TestNBinom:
 
 class TestGenInvGauss:
     def setup_method(self):
-        np.random.seed(1234)
+        self.rng = np.random.default_rng(6473281180)
 
     @pytest.mark.slow
     def test_rvs_with_mode_shift(self):
         # ratio_unif w/ mode shift
         gig = stats.geninvgauss(2.3, 1.5)
-        _, p = stats.kstest(gig.rvs(size=1500, random_state=1234), gig.cdf)
+        _, p = stats.kstest(gig.rvs(size=1500, random_state=self.rng), gig.cdf)
         assert_equal(p > 0.05, True)
 
     @pytest.mark.slow
     def test_rvs_without_mode_shift(self):
         # ratio_unif w/o mode shift
         gig = stats.geninvgauss(0.9, 0.75)
-        _, p = stats.kstest(gig.rvs(size=1500, random_state=1234), gig.cdf)
+        _, p = stats.kstest(gig.rvs(size=1500, random_state=self.rng), gig.cdf)
         assert_equal(p > 0.05, True)
 
     @pytest.mark.slow
     def test_rvs_new_method(self):
         # new algorithm of Hoermann / Leydold
         gig = stats.geninvgauss(0.1, 0.2)
-        _, p = stats.kstest(gig.rvs(size=1500, random_state=1234), gig.cdf)
+        _, p = stats.kstest(gig.rvs(size=1500, random_state=self.rng), gig.cdf)
         assert_equal(p > 0.05, True)
 
     @pytest.mark.slow
     def test_rvs_p_zero(self):
         def my_ks_check(p, b):
             gig = stats.geninvgauss(p, b)
-            rvs = gig.rvs(size=1500, random_state=1234)
+            rvs = gig.rvs(size=1500, random_state=self.rng)
             return stats.kstest(rvs, gig.cdf)[1] > 0.05
         # boundary cases when p = 0
         assert_equal(my_ks_check(0, 0.2), True)  # new algo
@@ -926,7 +929,7 @@ class TestGenInvGauss:
 
     def test_invgauss(self):
         # test that invgauss is special case
-        ig = stats.geninvgauss.rvs(size=1500, p=-0.5, b=1, random_state=1234)
+        ig = stats.geninvgauss.rvs(size=1500, p=-0.5, b=1, random_state=1464878613)
         assert_equal(stats.kstest(ig, 'invgauss', args=[1])[1] > 0.15, True)
         # test pdf and cdf
         mu, x = 100, np.linspace(0.01, 1, 10)
@@ -955,8 +958,6 @@ class TestGenInvGauss:
 
 
 class TestGenHyperbolic:
-    def setup_method(self):
-        np.random.seed(1234)
 
     def test_pdf_r(self):
         # test against R package GeneralizedHyperbolic
@@ -1189,9 +1190,6 @@ class TestHypSecant:
 
 
 class TestNormInvGauss:
-    def setup_method(self):
-        np.random.seed(1234)
-
     def test_cdf_R(self):
         # test pdf and cdf vals against R
         # require("GeneralizedHyperbolic")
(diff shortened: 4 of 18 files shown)


scipy/stats/tests/test_distributions.py
```diff
@@ -1350,6 +1346,9 @@ def test_logsf(self):
 
 
 class TestGennorm:
+    def setup_class(self):
```
mdhaber:
```suggestion
    def setup_method(self):
```

mdhaber (commented):
Is setting `self.rng` in `setup_method` OK here? I guess not, considering the test failure?

scipy/stats/tests/test_distributions.py
```diff
@@ -77,6 +77,9 @@ def test_distributions_submodule():
 
 
 class TestVonMises:
+    def setup_class(self):
```
mdhaber, line 80:
```suggestion
    def setup_method(self):
```
Here and throughout.

scipy/stats/tests/test_distributions.py
```diff
@@ -6424,11 +6429,14 @@ def test_args_reduce():
 
 
 class TestFitMethod:
+    def setup_method(self):
+        self.rng = np.random.default_rng(7310515167)
+
```
mdhaber:
```suggestion
```

scipy/stats/tests/common_tests.py
```diff
@@ -197,12 +197,8 @@ def check_named_args(distfn, x, shape_args, defaults, meths):
 def check_random_state_property(distfn, args):
     # check the random_state attribute of a distribution *instance*
 
-    # This test fiddles with distfn.random_state. This breaks other tests,
-    # hence need to save it and then restore.
-    rndm = distfn.random_state
-
     # baseline: this relies on the global state
-    np.random.seed(1234)
+    np.random.seed(1234)  # valid use of np.random.seed
```
andyfaff, line 201: L215 doesn't need the `if` clause any more.

scipy/stats/tests/test_continuous_basic.py
```diff
@@ -587,7 +587,6 @@ def test_gh1320_regression():
 
 def test_method_of_moments():
     # example from https://en.wikipedia.org/wiki/Method_of_moments_(statistics)
-    np.random.seed(1234)
```
andyfaff, line 590: Why was this seed needed in the first place, presumably in the fit method?

scipy/stats/tests/test_discrete_distns.py
```diff
@@ -397,17 +398,19 @@ def test_mean_against_mpmath(self, a, n, ref):
...
-    x = randint.rvs(xl, xu, size=xl.shape)
-    odds = np.random.rand(*x.shape)*2
+    def setup_method(self):
+        rng = np.random.default_rng(7162434334)
+        shape = (2, 4, 3)
+        max_m = 100
+        m1 = rng.integers(1, max_m, size=shape)  # red balls
+        m2 = rng.integers(1, max_m, size=shape)  # white balls
+        N = m1 + m2  # total balls
+        n = randint.rvs(0, N, size=N.shape, random_state=rng)  # number of draws
```
andyfaff, line 408: Is the `random_state` keyword of the distribution going to be changed to `rng` at some point?

scipy/stats/tests/test_entropy.py
```diff
@@ -226,7 +226,8 @@ def test_base_differential_entropy_transposed(self, xp):
         )
 
     def test_input_validation(self, xp):
-        x = np.random.rand(10)
+        rng = np.random.default_rng(4315387321)
```
andyfaff: not clear why random numbers are needed at all here.

scipy/stats/tests/test_hypotests.py
```diff
@@ -1014,16 +1011,16 @@ def test_somersd_large_inputs_gh18132(self):
         # generate lists of random classes 1-2 (binary)
         classes = [1, 2]
         n_samples = 10 ** 6
-        random.seed(6272161)
-        x = random.choices(classes, k=n_samples)
-        y = random.choices(classes, k=n_samples)
+        rng = np.random.default_rng(6889320191)  #
```
andyfaff:
```suggestion
        rng = np.random.default_rng(6889320191)
```

scipy/stats/tests/test_hypotests.py
```diff
@@ -1014,16 +1011,16 @@ def test_somersd_large_inputs_gh18132(self):
...
+        x = rng.choice(classes, n_samples)
+        y = rng.choice(classes, n_samples)
 
         # get value to compare with: sklearn output
         # from sklearn import metrics
         # val_auc_sklearn = metrics.roc_auc_score(x, y)
         # # convert to the Gini coefficient (Gini = (AUC*2)-1)
         # val_sklearn = 2 * val_auc_sklearn - 1
-        val_sklearn = -0.001528138777036947
+        val_sklearn = 0.000624401938730923
```
andyfaff, line 1023: Presumably this is stable?

scipy/stats/tests/test_resampling.py
```diff
@@ -563,8 +563,8 @@ def test_bootstrap_alternative(method):
 
 def test_jackknife_resample():
     shape = 3, 4, 5, 6
-    np.random.seed(0)
-    x = np.random.rand(*shape)
+    rng = np.random.default_rng(5274950392)
+    x = rng.standard_normal(shape)
```
andyfaff: The existing code is a uniform number generator, not a normal distribution.

```suggestion
    x = rng.random(size=shape)
```

scipy/stats/tests/test_resampling.py
```diff
@@ -582,17 +582,15 @@ def test_jackknife_resample():
 
 @pytest.mark.parametrize("rng_name", ["RandomState", "default_rng"])
 def test_bootstrap_resample(rng_name):
-    rng = getattr(np.random, rng_name, None)
-    if rng is None:
-        pytest.skip(f"{rng_name} not available.")
+    rng = getattr(np.random, rng_name)
     rng1 = rng(0)
```
andyfaff: I think we should use a non-zero seed here. Is it sensitive to seed?

scipy/stats/tests/test_stats.py
```diff
@@ -1068,6 +1068,9 @@ class TestCorrSpearmanr:
         your program has them.
     """
 
+    def setup_method(self):
```
andyfaff, line 1071: Is this threadsafe? Should the rng be created in each method?

andyfaff: The tests in this file seem to arbitrarily use this rng, or create it in the test.

scipy/stats/tests/test_stats.py
```diff
@@ -5363,10 +5365,9 @@ class Test_ttest_ind_permutations:
     b3 = [3, 4]
 
     # data for bigger test
-    np.random.seed(0)
```
andyfaff, line 5366: There is a rng seed on L5353 (old)/L5355 (new) ("data for most tests"). Should an rng be created on this line for use in the bigger test with the same seed as used above, i.e. `rng = np.random.default_rng(169708062)`

scipy/stats/tests/test_stats.py
```diff
@@ -6426,8 +6429,8 @@ def test_jarque_bera_against_R(self, xp):
     @skip_xp_backends(np_only=True)
     def test_jarque_bera_array_like(self, xp):
         # array-like only relevant for NumPy
-        np.random.seed(987654321)
-        x = np.random.normal(0, 1, 100000)
+        rng = np.random.default_rng(9294968266)
+        x = rng.normal(0, 1, size=100000)
```
andyfaff:
```suggestion
        x = rng.standard_normal(size=100000)
```

scipy/stats/tests/test_stats.py
```diff
@@ -8766,13 +8775,14 @@ class TestPageTrendTest:
     # results could be checked (to limited precision) against tables in
     # scipy.stats.page_trend_test reference [1]
 
-    np.random.seed(0)
-    data_3_25 = np.random.rand(3, 25)
-    data_10_26 = np.random.rand(10, 26)
+    rng = np.random.default_rng(3113562111)
+    data_3_25 = rng.random((3, 25))
+    rng = np.random.default_rng(3113562111)
```
andyfaff, line 8787: why is rng recreated again? `data_10_26` will have overlapping random numbers with `data_3_25`. The previous code would have those two objects with totally different random numbers.

i.e. shouldn't it just be:
```
    rng = np.random.default_rng(3113562111)
    data_3_25 = rng.random((3, 25))
    data_10_26 = rng.random((10, 26))
```

scipy/stats/tests/test_stats.py
```diff
@@ -8813,10 +8823,8 @@ class TestPageTrendTest:
            [[3, 2, 1], [3, 2, 1], [3, 2, 1], [3, 2, 1], [3, 2, 1], [3, 2, 1],
             [3, 2, 1], [3, 2, 1], [3, 2, 1], [2, 1, 3], [1, 2, 3]]),
          ]
-
```
andyfaff, line 8823: STY: A blank line needs to be between methods of a class.

scipy/stats/tests/test_mgc.py
```diff
@@ -70,25 +70,29 @@ def test_error_infty(self):
 class TestMGCStat:
     """ Test validity of MGC test statistic
     """
-    def _simulations(self, samps=100, dims=1, sim_type=""):
+    def setup_method(self):
+        self.rng = np.random.default_rng(1266219746)
+
+    def _simulations(self, samps=100, dims=1, sim_type="", rng=None):
+        rng = rng if rng is not None else self.rng
```
andyfaff:
```suggestion
        rng = rng or self.rng
```

scipy/stats/tests/test_mgc.py
```diff
@@ -70,25 +70,29 @@ def test_error_infty(self):
...
-                 0.4*np.random.random_sample(size=(x.size, 1)))
+                 0.4*self.rng.random(size=(x.size, 1)))
 
         # independence (tests type I simulation)
         elif sim_type == "independence":
-            u = np.random.normal(0, 1, size=(samps, 1))
-            v = np.random.normal(0, 1, size=(samps, 1))
-            u_2 = np.random.binomial(1, p=0.5, size=(samps, 1))
-            v_2 = np.random.binomial(1, p=0.5, size=(samps, 1))
+            u = self.rng.normal(0, 1, size=(samps, 1))
```
andyfaff:
```suggestion
            u = self.rng.standard_normal(size=(samps, 1))
```

scipy/stats/tests/test_mgc.py
```diff
@@ -70,25 +70,29 @@ def test_error_infty(self):
...
+            v = self.rng.normal(0, 1, size=(samps, 1))
```
andyfaff:
```suggestion
            v = self.rng.standard_normal(size=(samps, 1))
```

scipy/stats/tests/test_mgc.py
```diff
@@ -99,74 +103,74 @@ def _simulations(self, samps=100, dims=1, sim_type=""):
 
         # add dimensions of noise for higher dimensions
         if dims > 1:
-            dims_noise = np.random.normal(0, 1, size=(samps, dims-1))
+            dims_noise = self.rng.normal(0, 1, size=(samps, dims-1))
```
andyfaff:
```suggestion
            dims_noise = self.rng.standard_normal(size=(samps, dims-1))
```

scipy/stats/tests/test_mgc.py
```diff
@@ -99,74 +103,74 @@ def _simulations(self, samps=100, dims=1, sim_type=""):
...
 
     @pytest.mark.xslow
     def test_twosamp(self):
-        np.random.seed(12345678)
-
         # generate x and y
-        x = np.random.binomial(100, 0.5, size=(100, 5))
-        y = np.random.normal(0, 1, size=(80, 5))
+        x = self.rng.binomial(100, 0.5, size=(100, 5))
+        y = self.rng.normal(0, 1, size=(80, 5))
```
andyfaff:
```suggestion
        y = self.rng.standard_normal(size=(80, 5))
```

scipy/stats/tests/test_mgc.py
```diff
@@ -99,74 +103,74 @@ def _simulations(self, samps=100, dims=1, sim_type=""):
...
 
         # test stat and pvalue
-        stat, pvalue, _ = stats.multiscale_graphcorr(x, y)
+        stat, pvalue, _ = stats.multiscale_graphcorr(x, y, random_state=self.rng)
         assert_approx_equal(stat, 1.0, significant=1)
         assert_approx_equal(pvalue, 0.001, significant=1)
 
         # generate x and y
-        y = np.random.normal(0, 1, size=(100, 5))
+        y = self.rng.normal(0, 1, size=(100, 5))
```
andyfaff:
```suggestion
        y = self.rng.standard_normal(size=(100, 5))
```

scipy/stats/tests/test_mgc.py
```diff
@@ -208,8 +209,6 @@ def test_pvalue_literature(self):
 
     @pytest.mark.xslow
     def test_alias(self):
-        np.random.seed(12345678)
```
andyfaff, line 211: why was this seed required?

scipy/stats/tests/test_distributions.py
```diff
@@ -926,7 +929,7 @@ def test_rvs_negative_p(self):
 
     def test_invgauss(self):
         # test that invgauss is special case
-        ig = stats.geninvgauss.rvs(size=1500, p=-0.5, b=1, random_state=1234)
+        ig = stats.geninvgauss.rvs(size=1500, p=-0.5, b=1, random_state=1464878613)
```
andyfaff, line 932: Why does this use an integer and not a `Generator` instance?

scipy/stats/tests/test_distributions.py
```diff
@@ -955,8 +958,6 @@ def test_pdf_zero(self):
 
 
 class TestGenHyperbolic:
-    def setup_method(self):
-        np.random.seed(1234)
```
andyfaff, line 959: I didn't expand the entire class - I'm assuming that no global random state is used within it?

andyfaff:
I looked through all the files and it looks like it's in a good state.

scipy/stats/tests/common_tests.py
mdhaber: Didn't look to me like it was needed. I've found a few of those.

scipy/stats/tests/test_continuous_basic.py
mdhaber: No. Nowhere, AFAICT.

scipy/stats/tests/test_discrete_distns.py
mdhaber: Not planned.

scipy/stats/tests/test_entropy.py
mdhaber: I didn't try to optimize all the tests, just convert. But I can change this to `ones` if preferred. *Done.*

scipy/stats/tests/test_hypotests.py
mdhaber: Yes. I ran the sklearn code to get it.

scipy/stats/tests/test_mgc.py
mdhaber: Was only translating, but I can commit these.

scipy/stats/tests/test_mgc.py
mdhaber: Wasn't AFAICT. If I saw I need, I would have replaced it.

scipy/stats/tests/test_resampling.py
mdhaber: I don't think it is. I think use nonzero seed throughout SciPy could be a whole other PR, so I didn't update all seeds, but I can change this one. *Done.*

scipy/stats/tests/test_resampling.py
mdhaber: Ok. Maybe I still had `standard_normal` on the clipboard. It doesn't affect anything.

scipy/stats/tests/test_stats.py
mdhaber: > The tests in this file seem to arbitrarily use this rng, or create it in the test.

Only tests in the `TestCorrSpearmanr` class can use this rng. I replaced all uses of seeds in `TestCorrSpearmanr` with this rng. In other classes, if it was more convenient to create a new rng in the method rather than use a `setup_method`, I did that. This PR does not attempt to make all uses of RNGs throughout `scipy.stats` consistent in style.

> Is this threadsafe? Should the rng be created in each method?

I don't know. I asked that in my comments on the PR (https://github.com/scipy/scipy/pull/23739#pullrequestreview-3302917087). Please ping someone who would know. 

There is supposed to be a new instance of the class created for each `test_` method, and this code runs to initialize each instance. So `self.rng` should alreay be unique for each method.

With that understanding, I don't see how there would be sharing between threads unless multiple threads are using the same instance of the class to run the same method simultaneously. I don't see why that would be done.

scipy/stats/tests/test_stats.py
mdhaber: Would be nice if the linter caught obvious stuff like this. You assumed I didn't remove it intentionally, right? *Added back the line.*

scipy/stats/tests/test_stats.py
mdhaber: Because I accidentally generated the reference values in R using these values and there was no need to do it again. It requires a lot of manual copy-pasting. There is no effect on the legitimacy of the tests.

scipy/stats/tests/test_distributions.py
mdhaber: Because it did before. I'm not abolishing use of `RandomState` in this PR; just removing `seed`.

scipy/stats/tests/test_distributions.py
mdhaber: I didn't see any.

scipy/stats/tests/test_entropy.py
```diff
@@ -226,7 +226,8 @@ def test_base_differential_entropy_transposed(self, xp):
...
+        x = rng.random(10)
```
mdhaber:
```suggestion
        x = np.ones(10)
```

scipy/stats/tests/test_resampling.py
```diff
@@ -582,17 +582,15 @@ def test_jackknife_resample():
...
     rng2 = rng(0)
```
mdhaber:
```suggestion
    rng1 = rng(3949441460)
    rng2 = rng(3949441460)
```

scipy/stats/tests/test_stats.py
mdhaber: It doesn't appear to matter.

scipy/stats/tests/test_stats.py
```diff
@@ -8813,10 +8823,8 @@ class TestPageTrendTest:
            [[3, 2, 1], [3, 2, 1], [3, 2, 1], [3, 2, 1], [3, 2, 1], [3, 2, 1],
             [3, 2, 1], [3, 2, 1], [3, 2, 1], [2, 1, 3], [1, 2, 3]]),
          ]
```
mdhaber, line 8832:
```suggestion
         ]

```

mdhaber:
@andyfaff committed all suggestions and responded to all remaining comments. If you're not sure about the use of `setup_method`, please ping someone who might know. There was a failure with ARGUS tests in the free-threaded job of the last commit, so I've reverted the changes there. They were not using the `RandomState` singleton before, so there wasn't really a need to change them.

scipy/stats/tests/test_distributions.py
```diff
@@ -9680,9 +9686,12 @@ def test_overflow(self):
 
 
 class TestArgus:
+    def setup_method(self):
+        self.rng = np.random.default_rng(3395603429)
+
```
mdhaber:
```suggestion
```

scipy/stats/tests/test_distributions.py
```diff
@@ -9691,7 +9700,7 @@ def test_argus_rvs_large_chi(self):
             [3.5, 135]    # chi > 1.8: transform conditional Gamma distribution
         ])
     def test_rvs(self, chi, random_state):
-        x = stats.argus.rvs(chi, size=500, random_state=random_state)
+        x = stats.argus.rvs(chi, size=500, random_state=self.rng)
```
mdhaber:
```suggestion
        x = stats.argus.rvs(chi, size=500, random_state=random_state)
```
In principle, most seeds will work, but I didn't intend to change the logic.

scipy/stats/tests/test_distributions.py
```diff
@@ -9700,7 +9709,7 @@ def test_rvs_small_chi(self, chi):
         # test for gh-11699 => rejection method case 1 can even handle chi=0
         # the CDF of the distribution for chi=0 is 1 - (1 - x**2)**(3/2)
         # test rvs against distribution of limit chi=0
-        r = stats.argus.rvs(chi, size=500, random_state=890981)
+        r = stats.argus.rvs(chi, size=500, random_state=self.rng)
```
mdhaber:
```suggestion
        r = stats.argus.rvs(chi, size=500, random_state=890981)
```

scipy/stats/tests/test_distributions.py
```diff
@@ -9680,9 +9686,12 @@ def test_overflow(self):
 
 
 class TestArgus:
+    def setup_method(self):
+        self.rng = np.random.default_rng(3395603429)
+
     def test_argus_rvs_large_chi(self):
         # test that the algorithm can handle large values of chi
-        x = stats.argus.rvs(50, size=500, random_state=325)
+        x = stats.argus.rvs(50, size=500, random_state=self.rng)
```
mdhaber:
```suggestion
        x = stats.argus.rvs(50, size=500, random_state=325)
```

mdhaber (commented):
There was nothing wrong with the ARGUS tests, so might as well leave them alone to avoid the failure.

j-bowhay:
> @andyfaff committed all suggestions and responded to all remaining comments. If you're not sure about the use of `setup_method`, please ping someone who might know. There was a failure with ARGUS tests in the free-threaded job of the last commit, so I've reverted the changes there. They were not using the `RandomState` singleton before, so there wasn't really a need to change them.

Would not claim to be an authority on this but looking at https://github.com/numpy/numpy/issues/29552 and, for example, https://github.com/numpy/numpy/pull/29596 it looks like the answer might be no it is not.

mdhaber:
I see in https://github.com/numpy/numpy/pull/29596#issue-3335594512:

> This makes the tests more thread safe, since xunit setup methods work strangely with threads.

I think I agree - it seems like they *should* be threadsafe, but something unusual is going on.

That said, `setup_method` is used throughout SciPy. If these tests pass consistently, shall we move from `setup_method` to something more verbose after a project-wide decision as part of a separate effort?

mdhaber:
@rgommers @crusaderky  `setup_method` is used frequently in tests of distributions (as in other places in SciPy), so I've kept that pattern here. If tests are passing consistently in this PR, I'd like to leave them as-they are for now, but do we need to think about transitioning these to something else (e.g. creating `rng` manually in each methods) due to thread-safety concerns?

rgommers:
> If tests are passing consistently in this PR, I'd like to leave them as-they are for now,

If they indeed pass consistently under repeated pytest-run-parallel testing, that seems okay to me. I think a lot of the added seeding is for tests that weren't using a seed before, so probably the tests in question are fairly insensitive to seeding to begin with.

That said, this is not a good pattern and should be avoided, whether here or in a follow-up:

```python
class TestLogistic:
    def setup_method(self):
        self.rng = np.random.default_rng(2807014525)

    def test_something(self):
        # uses self.rng

    def test_something_else(self):
        # also uses self.rng
```

This test isn't reproducible under pytest-run-parallel, and even without parallel testing it relies on a `pytest` execution order that may not hold (~EDIT: say you do test selection with `-k test_something_else`, you will now get a different set of random numbers for that test, because the test before it didn't run and hence didn't draw random numbers from `self.rng`~). It stores global state on a test class for very little gain, and makes a test harder to copy-paste for debugging. I'd always write it like this:

```python
class TestLogistic:
    def test_something(self):
        rng = np.random.default_rng(2807014525)
        # uses `rng`

    def test_something_else(self):
        rng = np.random.default_rng(2807014525)
        # uses `rng`
```

That is safe under pytest-run-parallel, copy-pasteable, and the increase in lines of code when one has say 10 test methods is still quite minor.

mdhaber:
> That said, this is not a good pattern and should be avoided...It stores global state on a test class for very little gain, and makes a test harder to copy-paste for debugging.

Right, I asked because I knew there were issues. I was also hoping to understand *why* there was a problem with using `setup_method`. You wrote:

> say you do test selection with -k test_something_else, you will now get a different set of random numbers for that test, because the test before it didn't run and hence didn't draw random numbers from self.rng

That suggests that all the test methods share an instance of the class, so the random state would be preserved between method calls, but according to [Pytest documentation](https://docs.pytest.org/en/stable/getting-started.html#group-multiple-tests-in-a-class):

> Something to be aware of when grouping tests inside classes is that each test has a unique instance of the class. Having each test share the same class instance would be very detrimental to test isolation and would promote poor test practices. 

Clearly using `setup_class` would share global state between tests because `self.rng` would really be an attribute of the class (`cls.rng`); that would lead to the behavior you described. But [`setup_method`](https://docs.pytest.org/en/stable/how-to/xunit_setup.html) is supposed to initialize a separate instance of the class for each method, so the seed should be the same (and therefore consistent) for each test method. ISTM this would be safe as long as pytest-run-parallel doesn't change this aspect of `pytest` (or use multiple threads to run the *same* test using the same object at the same time). What is wrong with this model?

rkern:
`pytest-run-parallel` breaks a lot of the guarantees that plain `pytest` otherwise makes.

rgommers:
Sorry, brain not working on that later EDIT; that wasn't right indeed. 

> ISTM this would be safe as long as pytest-run-parallel doesn't change this aspect of `pytest` (or use multiple threads to run the _same_ test using the same object at the same time). What is wrong with this model?

`pytest-run-parallel` does change that. It runs multiple invocations of the body of the test function. The `setup_method` runs just once, and the implicit `teardown` method that `pytest` pairs with `setup_method` runs as soon as the first invocation completes. Which can make `pytest-run-parallel` fail on this kind of `setup_method` usage.

mdhaber:
OK, I see. I assumed from the name that it was like `pytest-xdist` and that the purpose was to speed up test execution. But I understand now that it designed to expose thread safety concerns, and I suppose it intentionally does unsafe things to find them (at the risk of false positives).

rgommers:
Yes, that is exactly it. And it repeats tests, so it's slower rather than faster. If you run it with N threads, it does roughly N times more work.

mdhaber:
OK, and if I remember correctly, fixtures for `rng`s are also out for the same sort of reason?

I'll probably open a separate issue and do some work along these lines, including adding something to the documentation about it. Looks like the [Testing guidelines](https://numpy.org/doc/stable/reference/testing.html) could be updated if NumPy is also using `pytest-run-parallel`, or else maybe it's worth adding something to the [tips](https://scipy.github.io/devdocs/dev/contributor/writing_test_tips.html#writing-test-tips).

mdhaber:
Resolved merge conflicts and `pytest-run-parallel` passed again. Doc failures are in gh-23771, and macOS failures look like the usual.

andyfaff:
I thought the nan policy fails were fixed I'm #23716? Do the tolerances need to be relaxed further?

mdhaber:
I based the tolerances on the failures that showed up on your experimental build, but they were not consistent in CI, so maybe.

rgommers:
Fixtures have the same issue indeed. 

> I'll probably open a separate issue and do some work along these lines, including adding something to the documentation about it.

SGTM, thanks!

andyfaff:
@mdhaber if you're happy I think this can be merged. I'll do so this afternoon if you don't get around to it.

mdhaber:
Yes, please!

PR #23743, opened 2025-10-08
Cygwin interpolate test xfail

DWesl:
#### Reference issue

Separated from #23737

This gets SciPy compiling again on Cygwin, and all tests in interpolate passing.

In particular, it includes `Python.h` at the top of a few files, and marks a bspline mmap test xfail because Cygwin's threading module doesn't have `get_native_id`

### DIFF
--- a/scipy/interpolate/src/__fitpack.cc
+++ b/scipy/interpolate/src/__fitpack.cc
@@ -1,3 +1,5 @@
+#include <Python.h>
+/* __fitpack.h includes Python.h */
 #include <string>
 #include <cstdint>
 #include <vector>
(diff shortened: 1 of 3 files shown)


tylerjereddy (commented):
Looks like these are all changes Evgeni was in favor of in the cross-linked PR. The `xfail` does seem consistent with supported platform list not including Cygwin at https://docs.python.org/3/library/threading.html#threading.get_native_id.

For some reason, I thought we had a linter-style script for automatically checking the `<Python.h>` include order shenanigans, but maybe I'm mistaken (`tools/check_python_h_first.py`, which you wrote?).

The two CI failures seem clearly unrelated.

DWesl:
That script is supposed to find these include-order problems, yes.  I thought I had made changes that would catch this case, but apparently they didn't work.

ev-br:
AFAICS Tyler's comment is not blocking: it'd be nice to have the Python.h finder script catch everything but it'd be an enhancement and I don't know if updating it is worth the effort.

The patch itself LGTM, merging. Thank you @DWesl , @tylerjereddy

PR #23744, opened 2025-10-08
ENH: stats.bootstrap: array API support

mdhaber:
#### Reference issue
Toward gh-20544

Adds array API support to `stats.bootstrap`.

#### Additional information
`rng` is still a NumPy random number generator (`integers`), and jacknife generates indices with NumPy. The rest uses array API.

Line count is misleading. First commit moves all tests into a class, so it might be easier to review commits after that separately.

### DIFF
--- a/scipy/stats/_resampling.py
+++ b/scipy/stats/_resampling.py
@@ -12,6 +12,7 @@ from scipy._lib._array_api import (
     is_numpy,
     xp_capabilities,
     xp_result_type,
+    xp_size,
 )
 from scipy.special import ndtr, ndtri, comb, factorial
 
@@ -165,6 +166,7 @@ def _bootstrap_iv(data, statistic, vectorized, paired, axis, confidence_level,
                   alternative, n_resamples, batch, method, bootstrap_result,
                   rng):
     """Input validation and standardization for `bootstrap`."""
+    xp = array_namespace(*data)
 
     if vectorized not in {True, False, None}:
         raise ValueError("`vectorized` must be `True`, `False`, or `None`.")
@@ -173,6 +175,14 @@ def _bootstrap_iv(data, statistic, vectorized, paired, axis, confidence_level,
         vectorized = 'axis' in inspect.signature(statistic).parameters
 
     if not vectorized:
+        if not is_numpy(xp):
+            message = (f"When using array library {xp.__name__}, `func` must be "
+                       "be vectorized and accept argument `axis`.")
+            raise TypeError(message)
+
+        message = ("For better performance, `func` should be vectorized and "
+                   "accept argument `axis`.")
+        warnings.warn(message, stacklevel=3)
         statistic = _vectorize_statistic(statistic)
 
     axis_int = int(axis)
@@ -188,14 +198,14 @@ def _bootstrap_iv(data, statistic, vectorized, paired, axis, confidence_level,
     if n_samples == 0:
         raise ValueError("`data` must contain at least one sample.")
 
-    data = _broadcast_arrays(data, axis_int)
+    data = _broadcast_arrays(data, axis_int, xp=xp)
 
     data_iv = []
     for sample in data:
         if sample.shape[axis_int] <= 1:
             raise ValueError("each sample in `data` must contain two or more "
                              "observations along `axis`.")
-        sample = np.moveaxis(sample, axis_int, -1)
+        sample = xp.moveaxis(sample, axis_int, -1)
         data_iv.append(sample)
 
     if paired not in {True, False}:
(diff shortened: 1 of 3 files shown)


scipy/stats/_resampling.py
```diff
@@ -60,63 +62,48 @@ def _jackknife_resample(sample, batch=None):
...
+    return _get_from_last_axis(sample, xp.asarray(i, device=sample.device), xp=xp)
 
 
-def _percentile_of_score(a, score, axis):
+def _get_from_last_axis(sample, i, xp):
+    # Equivalent to `sample[..., i]` as used in `bootstrap`. Assumes i.ndim <=2.
+    if i.ndim == 2:
+        sample = xp.expand_dims(sample, axis=-2)
+    sample, i = _broadcast_arrays((sample, i), axis=-1, xp=xp)
+    return xp.take_along_axis(sample, i, axis=-1)
```
mdhaber, line 90: This function is *much* slower than `sample[..., i]` for NumPy arrays (and quite possibly other backends) because of `take_along_axis`. We should use this implementation for `array_api_strict` and otherwise just return `sample[..., i]`.

j-bowhay: Would it be too heinous just to skip array_api_strict? It seems slightly unfortunate to maintain a separate code path for a backend that no user uses

mdhaber: Hah, I thought *I* was the one who would advocate for that.
Yeah, when all real backends support something, I'm not a big fan of the contortions we go through for compatibility with array_api_strict.

That said, it's not a big deal here. At the beginning of the function:
```python3
    if not is_array_api_strict(xp):
        return sample[..., i]
```
And I suppose that's nice because if anyone wonders what the function is for, that line explains it.

And I don't think I had to jump through a ton of hoops to get the rest of the function working for `array_api_strict` after working on `cupy` and `torch`. (See https://github.com/scipy/scipy/pull/23744/commits/5e2c41c77ad77a9c66d2fbf6a3607d89a5e1c8cc.) So i think I'll keep it.

steppi (approved):
Looks pretty good, With my A4000, I'm seeing a nice speedup with CuPy when `n_resamples` gets large.

Borrowing this setup from your tests

```python
import numpy as np
import cupy
import scipy.stats as stats

rng = np.random.RandomState(0)
# float32 because my consumer GPU isn't as fast with double precision
x = rng.rand(20, 22, 24).astype(np.float32)
```

I'm seeing the following with some quick and dirty benchmarks with `%timeit` in IPython.

```
In [6]: %timeit res1 = stats.bootstrap((x,), np.mean, batch=None, method="basic", axis=0, n_re
      ⋮ samples=100)
1.55 ms ± 413 ns per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [7]: %timeit res2 = stats.bootstrap((cupy.asarray(x),), cupy.mean, batch=None, method="basi
      ⋮ c", axis=0, n_resamples=100)
2.17 ms ± 71 μs per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [8]: %timeit res1 = stats.bootstrap((x,), np.mean, batch=None, method="basic", axis=0, n_re
      ⋮ samples=1000)
23.3 ms ± 24.9 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [9]: %timeit res2 = stats.bootstrap((cupy.asarray(x),), cupy.mean, batch=None, method="basi
      ⋮ c", axis=0, n_resamples=1000)
4.3 ms ± 3.7 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [10]: %timeit res1 = stats.bootstrap((x,), np.mean, batch=None, method="basic", axis=0, n_r
       ⋮ esamples=10000)
293 ms ± 295 μs per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [11]: %timeit res2 = stats.bootstrap((cupy.asarray(x),), cupy.mean, batch=None, method="bas
       ⋮ ic", axis=0, n_resamples=10000)
27.7 ms ± 60.9 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [12]: %timeit res1 = stats.bootstrap((x,), np.mean, batch=None, method="basic", axis=0, n_r
       ⋮ esamples=100000)
3.57 s ± 2.36 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [13]: %timeit res2 = stats.bootstrap((cupy.asarray(x),), cupy.mean, batch=None, method="bas
       ⋮ ic", axis=0, n_resamples=100000)
280 ms ± 416 μs per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

This was all with `method="basic"`, but I see similar speedups with `"bca"` and `"percentile"` (`"percentile"` does less work than `"basic"` so no surprise there of course).

I'm happy merging this.

PR #23746, opened 2025-10-09
ENH: add Array API support to RBFInterpolator

ev-br:
#### Reference issue

Towards https://github.com/scipy/scipy/issues/23754

This PR is a potentially mergeable part of gh-23447 which is blocked on the JIT strategies. So this PR separates the "eager" part, and adds a "generic" array API backend.

Previously, `RBFInterpolator` was using a pythran-compiled extension for performance critical loops. Here we add a "generic, Array API compatible backend for alternative array types". The call chain is:
- for numpy arrays: `_rbfinterp.py` <-- `_rbfinterp_np` (the "numpy backend") <-- `rbfinterp_pythran` (AOT compiled)
- for other array types: `_rbfinterp.py` <-- `_rbfinterp_xp` (the "generic backend") 

This way, handling of numpy arrays is backwards compatible, at a cost of some moving code around (from `_rbfinterp.py` to `_rbfinterp_np.py`, mostly).

Cons:
- we now have two almost-but-but-not-completely-similar backends;

Pros:
- per [benchmarks](https://github.com/ev-br/bench_playground/blob/master/viz.rbf.ipynb) from gh-23447,  we get from 3x to 25x perf increase on CUDA. 

#### Additional information

- only `neighbors=None` version is converted. Non-default values of `neighbors` require KDTree, which is numpy-only;
- mixing array types: an instance infers a namespace from the arguments of `__init__`,  stores it, and coerces the arguments in `__call__`; roughly,
```
def __call__(self, xi):
    xi = self._xp.asarray(xi)
   ...
```
this is consistent with what `BSpline` does after https://github.com/scipy/scipy/pull/23020


scipy/interpolate/_rbfinterp.py
```diff
@@ -48,90 +48,24 @@
     }
 
 
-def _monomial_powers(ndim, degree):
```
ev-br, line 51: This function has moved to `rbfinterp_common.py`

scipy/interpolate/_rbfinterp.py
```diff
@@ -48,90 +48,24 @@
...
-            # multiplicity indicating power (e.g., (0, 1, 1) represents x*y**2)
-            for var in mono:
-                out[count, var] += 1
-
-            count += 1
-
-    return out
-
-
-def _build_and_solve_system(y, d, smoothing, kernel, epsilon, powers):
```
ev-br, line 83: This function moved to `_rbfinterp_np.py`

ev-br:
Potential follow-ups:
- reduce code duplication between backends (the dream is to have a single source for both pythran AOT and array api "generic" JIT and eager; probably only asymptotically achievable). The first blocker is that pythran does source code analysis and special-cases numpy. This is not immediately compatible with our pattern of using `xp=` arguments, where `xp` becomes `numpy` at runtime. We probably need some way to express to `pythran` that "this xp is actually numpy, please compile things under this assumption"
- JIT strategies. This is gh-23447
- See if we can enable `numba` without code-duplicating a dedicated "backend". An immediate issue is that we need `np.prod(..., axis=-1)` and this is https://github.com/numba/numba/issues/1269

scipy/_lib/_array_api.py
```diff
@@ -679,7 +679,7 @@ def _make_sphinx_capabilities(
     return capabilities
 
 
-def _make_capabilities_note(fun_name, capabilities):
+def _make_capabilities_note(fun_name, capabilities, extra_note=None):
```
steppi, line 682: Nice. I've wanted something like this before but didn't think to just add it here. I'll use this in my next `signal` PR.

scipy/interpolate/_rbfinterp_xp.py
```diff
@@ -0,0 +1,266 @@
...
+the following differences:
+
+  -  We used vectorized code not explicit loops in `_build_system` and
+     `_build_evaluation_coefficients`; this is more torch/jax friendly;
+  - RBF kernels are also "vectorized" and not scalar: they receive an
+    array of norms not a single norm;
+  - RBF kernels accept an extra xp= argument;
+
+In general, we would prefer less code duplication. The main blocker ATM is
+that pythran cannot compile functions with an xp= argument where xp is numpy.
```
steppi, line 26: I think it's likely that Serge Sans Paille will work with us on this if we post an enhancement request on the Pythran repo.

steppi (approved):
This looks good to me @ev-br. I'm seeing around 2x speedup with CuPy for scaled up instances of the cases in the tests. It would probably be more if this could work with `float32` arithmetic (my consumer grade GPU isn't great at accelerating `float64`), but that could be saved for a follow-up if someone is interested in doing it. We should follow up with a Pythran issue to see what could be done about reducing code duplication.

scipy/interpolate/_rbfinterp_xp.py
ev-br: Indeed. Here's a pythran feature request: https://github.com/serge-sans-paille/pythran/issues/2367

ev-br:
Thanks @steppi!

> I'm seeing around 2x speedup with CuPy

A couple of quick notes:

First, the CuPy's RBFInterpolator is rather similar to the "generic backend" here, and quite probably can be sped up further: 
https://github.com/cupy/cupy/blob/v13.6.0/cupyx/scipy/interpolate/_rbfinterp.py

In one benchmark, I'm seeing 10x -- 15x speedups with torch and jax  CUDA eager compared to CuPy's 2x: 
https://github.com/ev-br/rbfinterp_xp_timings/blob/master/timings_cpu.ipynb

And the numpy/pythran baseline has a 2x variability depending on an environment variable, https://github.com/scipy/scipy/issues/23732

PR #23749, opened 2025-10-09
MAINT: spatial.transform.Rotation: improve autodiff

amacati:
`Rotation.from_matrix` uses an SVD for normalization. Some autodiff frameworks do not implement the gradient for `full_matrices=True`, e.g. jax: https://github.com/jax-ml/jax/issues/17518. The `full_matrices=False` case is covered (https://github.com/jax-ml/jax/pull/666). Since our matrices are guaranteed to be 3x3, setting the flag to `False` has no drawbacks and improves autodiff capabilities.

Autodiff for `Rotation.from_matrix` on some frameworks.


scottshambaugh (approved):
Agreed no difference for square matrices

j-bowhay (approved):

PR #23135, opened 2025-06-09
ENH: interpolate: Array API for PPoly and related objects

ev-br:
#### Reference issue

Continues gh-23052 ; cross-ref https://github.com/scipy/scipy/issues/23754 

#### What does this implement/fix?

Convert

- [x] PPoly
- [x] BPoly
- [x] CubicHermiteSpline
- [x] PchipInterpolator
- [x] Akima1DInterpolator
- [x] CubicSpline    
.
- [ ] PPoly.roots
- [ ] BPoly.from_derivatives

#### Additional information

Two not-entirely-trivial items, both related to (the lack of) ragged arrays:

- `PPoly.roots` returns either a numeric array or an object array, depending on the data. We discussed it a while ago, https://github.com/scipy/scipy/pull/3260#issuecomment-1166510296, and the conclusion was, basically, "yes it's ugly but not worth breaking backwards compat for". That was before array API though. The only Array API compatible way I can imagine is to make it return either an array or a list of arrays. While not great, it is similar in spirit to what it does now.
- `BPoly.from_derivatives` accepts either an array or a list of arrays. The latter we cannot avoid because otherwise we'd need ragged arrays.

PR #23753, opened 2025-10-10
MAINT: interpolate: use xp-capabilities for make_* functions

ev-br:
While at it, add array API tests for the recently added periodic spline fitting (`make_sprep(...., bc_type="periodic"`).

Am not adding `xp_capabilities` to `BSpline` because it runs into the same problem as https://github.com/scipy/scipy/pull/23135#issuecomment-3265704061 : `make_xp_test_case` turns a class into a function under `dask`, and this messes up classmethods.

### DIFF
--- a/scipy/interpolate/tests/test_bsplines.py
+++ b/scipy/interpolate/tests/test_bsplines.py
@@ -9,7 +9,7 @@ import warnings
 
 import numpy as np
 from scipy._lib._array_api import (
-    xp_assert_equal, xp_assert_close, xp_default_dtype, concat_1d
+    xp_assert_equal, xp_assert_close, xp_default_dtype, concat_1d, make_xp_test_case
 )
 import scipy._lib.array_api_extra as xpx
 from pytest import raises as assert_raises
@@ -1224,7 +1224,7 @@ class TestInterop:
         assert isinstance(tck_n2, tuple)   # back-compat: tck in, tck out
 
 
-@skip_xp_backends(cpu_only=True)
+@make_xp_test_case(make_interp_spline)
 class TestInterp:
     #
     # Test basic ways of constructing interpolating splines.
@@ -1767,7 +1767,7 @@ def make_lsq_full_matrix(x, y, t, k=3):
 parametrize_lsq_methods = pytest.mark.parametrize("method", ["norm-eq", "qr"])
 
 
-@skip_xp_backends(cpu_only=True)
+@make_xp_test_case(make_lsq_spline)
 class TestLSQ:
     #
     # Test make_lsq_spline
@@ -2176,7 +2176,7 @@ def data_file(basename):
                         'data', basename)
 
 
-@skip_xp_backends(cpu_only=True)
+@make_xp_test_case(make_smoothing_spline)
 class TestSmoothingSpline:
     #
     # test make_smoothing_spline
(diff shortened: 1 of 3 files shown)


scipy/interpolate/tests/test_bsplines.py
```diff
@@ -4126,22 +4134,24 @@ def test_array_not_list(self, s):
...
+        x, y, k = self._get_xyk(n=10, xp=xp)
 
         spl_i, u_i = make_splprep(y, s=0, bc_type="periodic")
         spl_n, u_n = make_splprep(y, s=1e-12, bc_type="periodic")
 
         xp_assert_close(u_i, u_n, atol=1e-15)
-        xp_assert_close(spl_i(u_i), y, atol=1e-15)
-        xp_assert_close(spl_n(u_n), y, atol=1e-7, rtol=1e-6)
+
+        y_arr = xp.stack(y)
```
tylerjereddy: Here and elsewhere, the new/extra `stack` caught my eye, seems like `torch` and `array-api-strict` complain otherwise.

It seems we can't adjust `_get_xyk()` itself to perform the stacking because i.e., `make_splprep()` has its own internal `stack()` that expects unstacked input generated by `_get_xyk()`.

Not saying anything is wrong, but it did stand out as a point of confusion for me when reviewing.

scipy/interpolate/tests/test_bsplines.py
```diff
@@ -3890,22 +3895,22 @@ def test_s0_vs_not(self, k):
...
+        y = xp.sin(x) + xp.cos(x)
         spl = make_splrep(x, y, s=1e-12, bc_type=self.bc_type)
-        xp_assert_close(splev(x, spl), y, atol=1e-5, rtol=1e-6)
+        xp_assert_close(spl(x), y, atol=1e-5, rtol=1e-6)
 
-        y = 5*np.sin(x) + np.cos(x)*3
+        y = 5*xp.sin(x) + xp.cos(x)*3
         spl = make_splrep(x, y, s=1e-8, bc_type=self.bc_type)
-        xp_assert_close(splev(x, spl), y, atol=1e-5, rtol=1e-4)
+        xp_assert_close(spl(x), y, atol=1e-5, rtol=1e-4)
```
tylerjereddy: this change caught my eye as well--since `splev` is marked legacy in the docs, it makes sense that we switch to the recommended direct `__call__` approach, but the reason for the change was not obvious without checking that.

tylerjereddy (commented):
Looks pretty clean overall. I spot checked the rendered docs for one of the newly-decorated functions, and the array API support table that gets injected looks nice.

I added minor comments about the testing for a small number of things that confused me at first, but I don't think I have any requested changes as long as those things look fine to other reviewers.

The one failing CI job is not related, and happening elsewhere as well.

scipy/interpolate/tests/test_bsplines.py
ev-br: Internally, `splev` does basically  `if isinstance(tck, BSpline): return tck(x)` so these two forms are equivalent. That said, this is indeed a gratuitous change, which I rolled back in a force-pushed commit.

scipy/interpolate/tests/test_bsplines.py
ev-br: Yeah. Here the `y` argument of `make_splprep` is a list of arrays (why? because `splprep` received a list of array-likes since forever, and we're matching its API).

At the same time, `xp_assert_*`  interprets a list as a numpy object (because array-like) and fails on the namespace mismatch.

I don't see a good way out, other than adding a comment here on the array conversion, which the updated PR does now.

tylerjereddy:
I'm happy with the state of this now, thanks Evgeni. CI fails still not related.

PR #23755, opened 2025-10-10
BUG: interpolate: fail gracefully if all weights are zero

ev-br:
#### Reference issue

closes https://github.com/scipy/scipy/issues/23542

While the meaning of `np.count_nonzero(weights) < k+1` is not immediately obvious, having all weights to be zero is clearly an error.

Thus, emit a clear error for definitely erroneous inputs, and leave the rest to some other day.


scipy/interpolate/tests/test_bsplines.py
```diff
@@ -3414,6 +3414,12 @@ def test_s_too_small(self):
         assert len(r) == 1
         xp_assert_equal(knots[-1], tck[0])
 
+    def test_zero_weights(self):
+        # regression test for https://github.com/scipy/scipy/issues/23542
+        gen = generate_knots([0.,1.,2.,3.], [4.,5.,6.,7.], w=[0.,0.,0.,0.], s=1)
+        with pytest.raises(ValueError, match="weights are zero"):
+            knots = list(gen)
```
ev-br:
```suggestion
            list(gen)
```

j-bowhay (approved):

PR #23758, opened 2025-10-11
ENH: implement `gdtrix` and `gdtria` using xsf's incomplete gamma inverse

dschmitz89:
#### Reference issue
Towards #21966 

This implements the gamma distribution CDF inverses using `gammaincinv`. This seemed quite simple, so I tried it out.

I do not fully understand the whole `xsf` wrapper machinery and the build system, so I added the functions at the same place where `gdtr` was wrapped. There is a compilation error left when the functions are used from cython which I cannot make sense of at the moment (edit: only appears locally for me?).

#### Additional information
`gdtr(a, b, x)=gammainc(b, a*x)` means that the inverses are simply

- `gdtria(p, b, x)=gammaincinv(b, p)/x`
- `gdtrix(a, b, p)=gammaincinv(b, p)/a`

A similar approach could be taken for `nrdtrimn` and `nrdtrisd` but they are also available in boost.

#### Benchmark
I benchmarked `gdtria` using a mpmath implementation of `gdtr`. As expected, xsf (is `gammaincinv` from cephes actually?) achieves a much lower relative error than cdflib. As this is in essence a benchmark of the incomplete gamma function inverses of both libraries, `gdtrix` improves at the same rate.

gdtria_Histograms

<p>

```python
import numpy as np
from mpmath import mp

# mp.dps = 50  # set decimal places for mpmath

# @np.vectorize
# def gdtr_ref(a, b, x):
#     """Reference implementation of the generalized distribution function using mpmath."""
#     a_mp = mp.mpf(a)
#     b_mp = mp.mpf(b)
#     x_mp = mp.mpf(x)
#     ref = mp.gammainc(b_mp, b=a_mp*x_mp, regularized=True)
#     return float(ref)

# def calculate_rel_error(computed, reference):
#     """
#     Calculate the relative error between computed and reference values.
#     """
#     rel_error = np.abs(computed - reference) / np.abs(reference)
#     eps = np.finfo(float).eps
#     rel_error = np.clip(rel_error, eps/10, 1)
#     return rel_error


# #x = 1e-10
# samples_per_variable = 50
# rng = np.random.default_rng(12345)
# b = rng.uniform(1e-5, 10, samples_per_variable)
# a = rng.uniform(1e-5, b, samples_per_variable)
# x = rng.uniform(1e-10, 1e1, samples_per_variable)
# a_grid, x_grid, b_grid = np.meshgrid(a, x, b)
# a_grid = a_grid.flatten()
# x_grid = x_grid.flatten()
# b_grid = b_grid.flatten()
# p_ref = gdtr_ref(a_grid, b_grid, x_grid)
# lower_threshold_p = 1e-20
# higher_threshold_p = 0.9999999
# mask = (p_ref > lower_threshold_p) & (p_ref < higher_threshold_p)
# a_grid = a_grid[mask]
# x_grid = x_grid[mask]
# b_grid = b_grid[mask]
# p_ref = p_ref[mask]
# gdtria_result = sc.gdtria(p_ref, b_grid, x_grid)

# rel_error = calculate_rel_error(gdtria_result, a_grid)

# np.save('./gdtria_bench_rel_error_PR.npy', rel_error)

rel_error_pr = np.load('./gdtria_bench_rel_error_PR.npy')
rel_error_main = np.load('./gdtria_bench_rel_error_main.npy')
bins = int(np.sqrt(len(rel_error_pr)))
plt.hist(np.log10(rel_error_pr), bins=bins, label="PR/cephes", density=True, cumulative=True, alpha=0.5)
plt.hist(np.log10(rel_error_main), bins=bins, label="main/cdflib", density=True, cumulative=True, alpha=0.5)
plt.legend()
plt.title("Cumulative Distribution of Relative Errors in gdtria")
plt.show()
``` 

</p>

</details>


steppi:
Yeah, this is all `cdflib` does too actually:

https://github.com/scipy/scipy/blob/cb0862c2e927ecc4b7ff1c779ec49ab869352cf0/scipy/special/cdflib.c#L2046-L2052

and

https://github.com/scipy/scipy/blob/cb0862c2e927ecc4b7ff1c779ec49ab869352cf0/scipy/special/cdflib.c#L2076-L2082

after you make sure edge cases are handled correctly, this would work fine. I can help with the infra stuff and I should also submit a PR documenting it.

dschmitz89:
@steppi This is done from my side. Benchmarks demonstrate the improvements (see top post).

Regarding edge case behaviour, I kept the original cdflib choices. If you would like changes there, I can do so but I would not like to invest too much time into it. These functions are rather niche and people should just use `gammaincinv` anyway in my opinion.

Regarding where these functions should be placed: they are not actually simple `xsf` wrappers, so we could put them into another file. I did not do so directly due to the complexities of the build/wrapper system. WDYT?

scipy/special/tests/test_cdflib.py
```diff
@@ -876,3 +876,29 @@ def test_accuracy(self, p, m, k):
...
+    (0, 0, 1, np.nan),
+    (0, np.inf, 0, np.nan),
+    (0, np.inf, 1, np.nan),
+    (np.inf, 0, 0, np.nan),
+    (np.inf, 0, 1, np.nan),
+    (np.inf, np.inf, 0, np.nan),
+    (np.inf, np.inf, 1, np.nan)
+])
+def test_gdtrix_edge_cases(a, b, p, ref):
+    assert_allclose(sp.gdtrix(a, b, p), ref, rtol=0)
```
WarrenWeckesser:
```suggestion
    assert_equal(sp.gdtrix(a, b, p), ref)
```
`rtol=0` makes this the same as `assert_equal()`. `assert_equal()` handles `nan` correctly here.

If you want, you could simplify the test by removing the reference value from the parametrized inputs and hard-code `nan` as the desired value in `assert_equal()`.  In that case, the test should be renamed to something like `test_gdtrix_nan_cases`.  Or just go with the suggested change.

scipy/special/tests/test_cdflib.py
```diff
@@ -876,3 +876,29 @@ def test_accuracy(self, p, m, k):
...
+    (0, 0, np.inf, np.nan),
+    (0, 1, 0, np.nan),
+    (0, 1, np.inf, 0),
+    (np.inf, 0, 0, np.nan),
+    (np.inf, 0, np.inf, np.nan),
+    (np.inf, 1, 0, np.nan),
+    (np.inf, 1, np.inf, np.nan)
+])
+def test_gdtria_edge_cases(p, b, x, ref):
+    assert_allclose(sp.gdtria(p, b, x), ref, rtol=0)
```
WarrenWeckesser:
```suggestion
    assert_equal(sp.gdtria(p, b, x), ref)
```

scipy/special/tests/test_cdflib.py
```diff
@@ -876,3 +876,29 @@ def test_accuracy(self, p, m, k):
...
+    (np.inf, np.inf, 1, np.nan)
+])
+def test_gdtrix_edge_cases(a, b, p, ref):
+    assert_allclose(sp.gdtrix(a, b, p), ref, rtol=0)
+
+@pytest.mark.parametrize("p, b, x, ref", [
+    (0, 0, 0, np.nan),
+    (0, 0, np.inf, np.nan),
+    (0, 1, 0, np.nan),
+    (0, 1, np.inf, 0),
```
WarrenWeckesser, line 897: I think `gdtria(0, 1, inf)` should be `nan`; `a = 0` is not a meaningful limit for the gamma distribution. Note that `gdtr(0, 1, inf)` returns `nan`.  But it looks like returning 0 is consistent with SciPy 1.16.3, so it should be OK to put off rationalizing the special cases in a follow-up.

scipy/special/xsf_wrappers.cpp
```diff
@@ -590,6 +590,28 @@ double xsf_gdtr(double a, double b, double x) { return xsf::gdtr(a, b, x); }
...
+    if (x == 0) {
+        return std::numeric_limits<double>::quiet_NaN();
+    }
+    if ((b == 0) & (p == 0)) {
+        if (std::isinf(x) && (x > 0)) {
+            return std::numeric_limits<double>::quiet_NaN();
+        }
+        return 0.0;
+    }
+    return xsf::gammaincinv(b, p) / x; }
```
WarrenWeckesser: Style nit, since this is a multi-line function body:
```suggestion
    return xsf::gammaincinv(b, p) / x;
}
```

scipy/special/xsf_wrappers.cpp
```diff
@@ -590,6 +590,28 @@ double xsf_gdtr(double a, double b, double x) { return xsf::gdtr(a, b, x); }
...
+
+double special_gdtrix(double a, double b, double p) { 
+    if ((a == 0) && (b == 0)) {
+            return std::numeric_limits<double>::quiet_NaN();
```
WarrenWeckesser: White space:
```suggestion
        return std::numeric_limits<double>::quiet_NaN();
```

scipy/special/xsf_wrappers.cpp
```diff
@@ -590,6 +590,28 @@ double xsf_gdtr(double a, double b, double x) { return xsf::gdtr(a, b, x); }
...
+    }
+    // if a or b is positive infinite, return NaN
+    if ((std::isinf(a) || std::isinf(b)) && (a >= 0 && b >= 0)) {
+        return std::numeric_limits<double>::quiet_NaN();
+    }
+    return xsf::gammaincinv(b, p) / a; }
```
WarrenWeckesser: Style nit:
```suggestion
    return xsf::gammaincinv(b, p) / a;
}
```

WarrenWeckesser (commented):
I made a few suggestions.

This doesn't use @izaid's nice C++-based ufunc generation, but that could be done later.

dschmitz89:
Comments are addressed and CI is green, so merging. I can look into using the new ufunc machinery in a follow up.

Issue #23754, opened 2025-10-10
Tracker: Array API support in `scipy.interpolate` 


ev-br:
This issue tracks progress toward the addition of array-API support to `scipy.interpolate` functionality:

Univariate interpolation
-----------------------------------

- [x] `make_interp_spline`  https://github.com/scipy/scipy/pull/23020

These three use C kernels; the "conversion" is to convert inputs to numpy, compute,
convert back:

- [x] `PPoly`  https://github.com/scipy/scipy/pull/23135
- [x] `BPoly`  https://github.com/scipy/scipy/pull/23135
- [x] `BSpline`  https://github.com/scipy/scipy/pull/23020

Missing niche APIs:
- `BSpline.design_matrix` : blocked on `csr_array`
- `PPoly.roots` : blocked on object arrays
- [x] `BPoly.from_derivatives`: `ndarray | list of arrays` inputs -- done in https://github.com/scipy/scipy/pull/25072/

These are `PPoly` subclasses, same process:

- [x] `CubicSpline`  https://github.com/scipy/scipy/pull/23135
- [x] `PchipInterpolator`  https://github.com/scipy/scipy/pull/23135
- [x] `Akima1DInterpolator`  https://github.com/scipy/scipy/pull/23135
- [x] `CubicHermiteSpline`  https://github.com/scipy/scipy/pull/23135

Floater-Hormann is pure python, but relies on `scipy.linalg`:

- [ ] `FloaterHormannInterpolator` -- in progress in https://github.com/scipy/scipy/pull/25091

These two are lower-prio:

- [ ] `BarycentricInterpolator`
- [ ] `KroghInterpolator`

Multivariate interpolation
-------------------------------------

These classes rely on `scipy.spatial`: Qhull or `KDTree` (NearestNDInterpolator)

- [ ] `LinearNDInterpolator`
- [ ] `NearestNDInterpolator`
- [ ] `CloughTocher2DInterpolator`

`RBFInterpolator` serves as a testbed for the "alternative backends"  idea (cf `spatial.Rotation`):

- [x] `RBFInterpolator`    https://github.com/scipy/scipy/pull/23746
- [ ] `RBFInterpolator(...., neighbors=<integer>)`: needs `KDTree`

These two use compiled kernels:

- [ ] `NdPPoly`
- [x] `NdBSpline`   https://github.com/scipy/scipy/pull/23759

"_legacy" methods are pure python, other methods rely on NdBSpline and `sparse.linalg`:

[...]

PR #23759, opened 2025-10-11
ENH: interpolate: make NdBSpline and RGI Array API compatible

ev-br:
#### Reference issue

towards gh-23754

Make `NdBSpline` array API compatible. The pattern is the same as `BSpline`: convert inputs to numpy, call the compiled implementation, convert back.

One exception is `design_matrix` which returns a `csr_array`, which is numpy-only, so there's nothing we can do until there's array api compatibility in `scipy.sparse`.

### DIFF
--- a/scipy/interpolate/_rgi.py
+++ b/scipy/interpolate/_rgi.py
@@ -294,9 +294,18 @@ class RegularGridInterpolator:
 
         try:
             xp = array_namespace(*points, values)
-        except Exception:
-            # "duck-type" values
-            xp = np_compat
+        except Exception as e:
+            # either "duck-type" values or a user error?
+            xp = array_namespace(*points)
+            try:
+                xp_v = array_namespace(values)
+            except Exception:
+                # "duck-type" values indeed, continue with `xp` as the namespace
+                pass
+            else:
+                # both `points` and `values` are array API objects, check consistency
+                if xp_v != xp:
+                    raise e
 
         self._asarray = xp.asarray
         self.method = method
--- a/scipy/interpolate/_rgi.py
+++ b/scipy/interpolate/_rgi.py
@@ -6,6 +6,8 @@ from types import GenericAlias
 import numpy as np
 
 import scipy.sparse.linalg as ssl
+from scipy._lib._array_api import array_namespace
+from scipy._lib.array_api_compat import numpy as np_compat, is_array_api_obj
 
 from ._interpnd import _ndim_coords_from_arrays
 from ._cubic import PchipInterpolator
@@ -281,17 +283,25 @@ class RegularGridInterpolator:
         if method not in self._ALL_METHODS:
             raise ValueError(f"Method '{method}' is not defined")
         elif method in self._SPLINE_METHODS:
-            self._validate_grid_dimensions(points, method)
+            self._validate_grid_dimensions(points, method)   # NB: uses np.atleast_1d
+
+        try:
+            xp = array_namespace(*points, values)
+        except Exception:
+            # "duck-type" values
+            xp = np_compat
+
+        self._asarray = xp.asarray
         self.method = method
         self._spline = None
         self.bounds_error = bounds_error
-        self.grid, self._descending_dimensions = _check_points(points)
-        self.values = self._check_values(values)
-        self._check_dimensionality(self.grid, self.values)
-        self.fill_value = self._check_fill_value(self.values, fill_value)
+        self._grid, self._descending_dimensions = _check_points(points)
+        self._values = self._check_values(values)
+        self._check_dimensionality(self._grid, self._values)
+        self.fill_value = self._check_fill_value(self._values, fill_value)
         if self._descending_dimensions:
-            self.values = np.flip(values, axis=self._descending_dimensions)
-        if self.method == "pchip" and np.iscomplexobj(self.values):
+            self._values = np.flip(values, axis=self._descending_dimensions)
+        if self.method == "pchip" and np.iscomplexobj(self._values):
             msg = ("`PchipInterpolator` only works with real values. If you are trying "
                    "to use the real components of the passed array, use `np.real` on "
                    "the array before passing to `RegularGridInterpolator`.")
@@ -311,7 +321,7 @@ class RegularGridInterpolator:
         if solver is None:
             solver = ssl.gcrotmk
         spl = make_ndbspl(
-                self.grid, self.values, self._SPLINE_DEGREE_MAP[method],
+                self._grid, self._values, self._SPLINE_DEGREE_MAP[method],
                 solver=solver, **solver_args
               )
         return spl
@@ -323,6 +333,9 @@ class RegularGridInterpolator:
         return _check_points(points)
 
     def _check_values(self, values):
+        if is_array_api_obj(values):
+            values = np.asarray(values)
+
         if not hasattr(values, 'ndim'):
             # allow reasonable duck-typed values
             values = np.asarray(values)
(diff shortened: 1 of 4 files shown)


ev-br:
Added an Array API conversion of `RegularGridInterpolator`. Internals involve compiled code for all methods, so the "conversion" is the same: convert to numpy, compute in a C extension, convert back. An ugly bit is "duck-typed" values, which `"linear"` and `"nearest"` modes historically allow and test for. Thus had to add some minor gymnastics to keep backwards compatibility. 

Might consider deprecating these "duck-typed values" at some point, but anyhow  a) it's separate, and b) we need the array api compatibility during the deprecation period if we deprecate them.

scipy/interpolate/_rgi.py
```diff
@@ -281,17 +290,25 @@ def __init__(self, points, values, method="linear", bounds_error=True,
...
             raise ValueError(f"Method '{method}' is not defined")
         elif method in self._SPLINE_METHODS:
-            self._validate_grid_dimensions(points, method)
+            self._validate_grid_dimensions(points, method)   # NB: uses np.atleast_1d
+
+        try:
+            xp = array_namespace(*points, values)
+        except Exception:
+            # "duck-type" values
+            xp = np_compat
```
steppi: I'm a little concerned about catching the bare `Exception` here and falling back to `np_compat`. The intention is that `values` only needs to quack like a duck, but if someone accidentally mixes array types in `points`, they may be surprised to find everything seems to work, but the NumPy backend is being used.

Might it be fine to instead do

```python
xp = array_namespace(*points)
```
?

All tests pass locally for me still with this change.

steppi (commented):
Besides concerns about the bare `Exception` being caught. I think this looks good @ev-br.

scipy/interpolate/_rgi.py
lucascolley: I would err towards just `array_namespace(*points, values)` and make users adapt, but I agree that this suggestion is an improvement over silently using NumPy.

steppi: Yeah, but users should be given a deprecation period to adapt, and that would be a backwards incompatible change that needs a discussion. Something needs to be done in the meantime. I guess something like

```python
        try:
            xp = array_namespace(*points, values)
        except Exception:
            # "duck-type" values
            xp = array_namespace(*points)
            # raise a deprecation warning here
```

would be the way to go, but could be done in a separate PR, to avoid stalling this one while the discussions play out.

lucascolley: > users should be given a deprecation period to adapt,

This is all hidden behind the experimental env var still, no? Without it, we get `np_compat` back from `array_namespace` anyway.

steppi: OK. If I understand right @lucascolley, you're saying that since `array_namespace` will never error when `SCIPY_ARRAY_API` is unset, directly using `xp = array_namespace(*points, values)` would only cause problems when `SCIPY_ARRAY_API=1`, so this wouldn't cause backwards incompatibility. 

Don't we have to think about what happens when this functionality is no longer hidden behind the experimental var. though? I think that's coming very soon. When that happens, just using `xp = array_namespace(*points, values)` will cause exceptions to be raised for cases where `values` is some odd duck typed thing that are currently supported. @ev-br suggested deprecating support for duck types `values`. It seemed like you were suggesting to immediately move to not supporting them as soon as alt backend support is no longer hidden by the environment variable.

lucascolley: I think it is in the spirit of gh-18286 to deprecate "odd duck typed thing[s]" that we only accidentally support with the env var. Maybe worth checking with @rgommers if we aren't aligned on this.

lucascolley: maybe some concrete examples would be helpful if it seems like my suggestion is too disruptive

EDIT:

> An ugly bit is "duck-typed" values, which "linear" and "nearest" modes historically allow and test for.

could those tests be pointed out?

steppi: Here's an example of one

https://github.com/scipy/scipy/blob/0c4cfab8b27d85785d28ec0e4b8d1ce536276c0e/scipy/interpolate/tests/test_rgi.py#L302-L314

In this test, `MyValue` is defined like this

```python
class MyValue:
    """
    Minimal indexable object
    """

    def __init__(self, shape):
        self.ndim = 2
        self.shape = shape
        self._v = np.arange(np.prod(shape)).reshape(shape)

    def __getitem__(self, idx):
        return self._v[idx]

    def __array_interface__(self):
        return None

    def __array__(self, dtype=None, copy=None):
        raise RuntimeError("No array representation")
```
        
Changing `RegularGridInterpolator` to have direct call to `array_namespace(*points, values)` results in the error

```
                # list, tuple, or arbitrary object
                try:
>                   array = np.asanyarray(array)
                            ^^^^^^^^^^^^^^^^^^^^
E                   ValueError: Invalid __array_interface__ value, must be a dict
```

inside of `array_namespace`. 

On `main`, this error sidestepped inside of `RegularGridInterpolator` here

https://github.com/scipy/scipy/blob/0c4cfab8b27d85785d28ec0e4b8d1ce536276c0e/scipy/interpolate/_rgi.py#L325-L334

we see that if `values` has an `ndim` `attr` then it never has `np.asarray` called on it, avoiding the failure due to invalid `__array_interface__`. 

I wasn't around to know why things are the way they are, but unfortunately this doesn't seem to be something that was accidentally supported that could be removed in the spirit of #18286. This appears to be some deliberately supported quirky behavior that presumably some users are relying on if it made it's way into the tests like this.

lucascolley: Thanks for the detailed explanation Albert! Okay, your suggestion sounds good to me 👍

ev-br: Poking around in `git blame -M -C -C` attributes the "duck-type" values back to commit https://github.com/scipy/scipy/commit/ce0cfff00a3 from 2014, made right after https://github.com/scipy/scipy/pull/3323 which brought the original RGI code for scipy 0.14 (zero fourteen, not a typo).
So it was clearly deliberate at the time.
Now, the landscape might have changed since 2014, and it's hard to tell if anybody is actually using it anymore. Previous experience changing things in RGI tells me that there are users with somewhat unusual constraints (large astronomical data IIRC). 

Fast-forward to 2025, we could probably try deprecating these "duck-type values". If only to see if there are users, and whether we can work with them or have to revert the deprecation. 

That said, I think it's best to
- have the deprecation in a separate PR
- have the deprecation message asking to report to scipy

------------------------------------------------------------------------------------

Now, for `array_namespace(*points, values)` vs `array_namespace(*points)`, as suggested by Albert: the former catches a user error where `points` is a `tuple_of_numpy_arrays` and `values` is a `torch` tensor; the latter IIUC caters for an error where `points=(torch_tensor, numpy_array)`. Is this correct @steppi?

My guess (completely lacking any data, just a pure handwaving) would be that the former is a more common error than the latter, which is why the original PR has the former. The latter is also unpleasant however, so how about checking it twice. The last commit has a dance which I *think* does distinguish between all of
- values are inconsistent with points (error out)
- points themselves are inconsistent (error out)
- points are consistent and values are "duck-typed" (can proceed)

Thoughts?

scipy/interpolate/_rgi.py
```diff
@@ -281,17 +290,34 @@ def __init__(self, points, values, method="linear", bounds_error=True,
...
             raise ValueError(f"Method '{method}' is not defined")
         elif method in self._SPLINE_METHODS:
-            self._validate_grid_dimensions(points, method)
+            self._validate_grid_dimensions(points, method)   # NB: uses np.atleast_1d
+
+        try:
+            xp = array_namespace(*points, values)
+        except Exception as e:
+            # either "duck-type" values or a user error?
+            xp = array_namespace(*points)
```
lucascolley:
```suggestion
            xp = array_namespace(*points) # still forbid mixed namespaces in `points`
```

lucascolley (commented):
this handling looks good to me now with one extra comment, thanks Evgeni!

scipy/interpolate/_rgi.py
ev-br: Thanks, this way it's _obviously better_ (tm). Taken along.

scipy/interpolate/_rgi.py
steppi: Yes, this looks good. I realized laying in bed last night that my suggestion wasn't really satisfactory. I thought of something similar to what you ended up doing, but slightly more complicated because it only allowed duck dtypes `values` when `points` are all NumPy arrays. I don't think the extra complication is necessary, and prefer how you did it here.

steppi (approved):
Looks good. Thanks Evgeni!

PR #23764, opened 2025-10-12
DOC: special: consistent modified Bessel order and fix Hankel function refs

jorenham:
This moves modified Bessel functions of the first kind (`iv`) before those of the second kind (`kv`), and corrects references to the `hankel1` and `hankel2` functions.


steppi (approved):
This looks good to me. Thanks @jorenham. Pro-tip, from now on you can add `[docs only]` to the end of your commit messages to get only the docs CI jobs to run.

PR #23765, opened 2025-10-13
TST: signal: isolate use of alternative backend only to function being tested in signal tests

steppi:
#### Reference issue

Closes #23750

This PR updates tests in `signal` which use the `xp` fixture to always isolate use of the alternative backend to only the function being tested. Many of the tests already did this, but some computed reference values or other values use to validate results with the `xp` backend, with some tests even skipped on some backends due to functions other than the function being tested not being supported. I originally did this work while adding `xp_capabilities` decorators for `signal` functions and using them in the `signal` tests, and realized that this would best be split into a separate PR.

To help implement such isolation, I added a function `xp_copy_to_numpy` which copies `xp` arrays to `np` arrays, to make it more straightforward to allow non-tested functions within a test to use the NumPy backend. There were some `signal` tests which used [`numpy.testing.assert_array_almost_equal_nulp`](https://numpy.org/doc/2.3/reference/generated/numpy.testing.assert_array_almost_equal_nulp.html), and were skipped on alternative backends for this reason. I created `scipy._lib._array_api.xp_assert_close_nulp` which does the strict checks, and then converts to numpy and uses `numpy.testing.assert_array_almost_equal_nulp` internally. These tests probably could have been rewritten to just use `xp_assert_close`, but for now it seems reasonable to preserve the existing behavior as much as possible.

### DIFF
--- a/doc/source/dev/api-dev/array_api.rst
+++ b/doc/source/dev/api-dev/array_api.rst
@@ -331,7 +331,7 @@ ensure that any failures that occur on a backend actually relate to the function
 of interest, and avoids the need to skip backends due to lack of support for
 functions other than ``f``.
 
-To help facillitate such backend isolation, there is a function ``_xp_copy_to_numpy``
+To help facilitate such backend isolation, there is a function ``_xp_copy_to_numpy``
 in ``scipy._lib._array_api`` which can copy an arbitrary ``xp`` array to a NumPy
 array, bypassing any device transfer guards, while preserving dtypes. It is essential
 that this function is only used in tests for functions other than the one being
--- a/doc/source/dev/api-dev/array_api.rst
+++ b/doc/source/dev/api-dev/array_api.rst
@@ -320,6 +320,64 @@ backend, use: ``SCIPY_DEVICE=mps spin test -b torch``.
 Note that there is a GitHub Actions workflow which tests with array-api-strict,
 PyTorch, and JAX on CPU.
 
+Testing Practice
+----------------
+
+It's best if individual tests for a function ``f`` using the ``xp`` fixture restrict
+use of alternative backends to only the function ``f`` being tested. Other functions
+evaluated within a test, for the purpose of producing reference values, inputs,
+round-trip calculations, etc. should instead use the NumPy backend. This helps
+ensure that any failures that occur on a backend actually relate to the function
+of interest, and avoids the need to skip backends due to lack of support for
+functions other than `f` when `f` is supported.
+
+To help facillitate such backend isolation, there is a function ``_xp_copy_to_numpy``
+in ``scipy._lib._array_api`` which can copy an arbitrary ``xp`` array to a NumPy
+array, bypassing any device transfer guards, while preserving dtypes. It is essential
+that this function is only used in tests for the purpose of using the NumPy backend
+for functions other than the function being tested.
+
+When attempting to isolate use of alternative backends to a particular function, one
+must be mindful that PyTorch allows for setting a default dtype, and SciPy is tested
+with both default dtype ``float32`` (set the environment variable ``SCIPY_DEFAULT_DTYPE``
+to ``float32``) and ``float64``. Tests using the ``xp`` fixture often rely on
+``xp.asarray`` producing arrays with the default dtype when list input is given and
+no explicit dtype specified. This means that if a test involves taking input arrays,
+and passing them to a function other than the one being tested in order to produce
+inputs for the function being tested, the following may appear natural to write,
+but will not produce the correct dtype behavior::
+
+  # conjugate pairs give real-coeff num & den
+  z = np.asarray([1j, -1j, 2j, -2j])
+  # shouldn't need elements of pairs to be adjacent
+  p = np.asarray([1+1j, 3-100j, 3+100j, 1-1j])
+  k = 23
+
+  # np.poly should do the right thing, but be explicit about
+  # taking real part
+  b = k * np.poly(z_np).real
+  a = np.poly(p_np).real
+  z, p, b, a = map(xp.asarray, (z, p, b, a))
+
+  bp, ap = zpk2tf(z, p, k)
+
+One should instead construct all inputs as ``xp`` arrays and then copy to
+NumPy arrays in order to ensure the default dtype is respected::
+
+  # conjugate pairs give real-coeff num & den
+  z = xp.asarray([1j, -1j, 2j, -2j])
+  # shouldn't need elements of pairs to be adjacent
+  p = xp.asarray([1+1j, 3-100j, 3+100j, 1-1j])
+  k = 23
+
+  # np.poly should do the right thing, but be explicit about
+  # taking real part
+  b = k * np.poly(_xp_copy_to_numpy(z)).real
+  a = np.poly(_xp_copy_to_numpy(p)).real
+  b, a = map(xp.asarray, (b, a))
+
+  bp, ap = zpk2tf(z, p, k)
+
 
 Testing the JAX JIT compiler
 ----------------------------
--- a/doc/source/dev/api-dev/array_api.rst
+++ b/doc/source/dev/api-dev/array_api.rst
@@ -348,36 +348,47 @@ and passing them to a function other than the one being tested in order to produ
 inputs for the function being tested, the following may appear natural to write
 but would not produce the correct dtype behavior::
 
-  # conjugate pairs give real-coeff num & den
+  # z, p, k will have dtype float64 regardless of the value of
+  # SCIPY_DEFAULT_DTYPE
   z = np.asarray([1j, -1j, 2j, -2j])
-  # shouldn't need elements of pairs to be adjacent
   p = np.asarray([1+1j, 3-100j, 3+100j, 1-1j])
   k = 23
 
-  # np.poly should do the right thing, but be explicit about
-  # taking real part
+  # np.poly will preserve dtype
   b = k * np.poly(z_np).real
   a = np.poly(p_np).real
+  # Input arrays z, p, and reference outputs b, a will all have
+  # dtype float64.
   z, p, b, a = map(xp.asarray, (z, p, b, a))
 
+  # With float64 inputs, the outputs bp and ap will be of dtype
+  # float64.
   bp, ap = zpk2tf(z, p, k)
+  # xp_assert_close checks for matching dtype. Due to the way the
+  # code was written above, zpk2tf is not tested with float32 inputs
+  # when SCIPY_DEFAULT_DTYPE is float32.
+  xp_assert_close(b, bp)
+  xp_assert_close(a, ap)
 
 One should instead construct all inputs as ``xp`` arrays and then copy to
 NumPy arrays in order to ensure the default dtype is respected::
 
-  # conjugate pairs give real-coeff num & den
+  # calls to xp.asarray will respect the default dtype.
   z = xp.asarray([1j, -1j, 2j, -2j])
-  # shouldn't need elements of pairs to be adjacent
   p = xp.asarray([1+1j, 3-100j, 3+100j, 1-1j])
   k = 23
 
-  # np.poly should do the right thing, but be explicit about
-  # taking real part
+  # _xp_copy_to_numpy preserves dtype, as does np.poly.
   b = k * np.poly(_xp_copy_to_numpy(z)).real
   a = np.poly(_xp_copy_to_numpy(p)).real
+  # b and a will have dtype float32
   b, a = map(xp.asarray, (b, a))
 
+  # zpk2tf is tested with float32 inputs when SCIPY_DEFAULT_DTYPE=float32
+  # as intended.
   bp, ap = zpk2tf(z, p, k)
+  xp_assert_close(b, bp)
+  xp_assert_close(a, ap)
 
 
 Testing the JAX JIT compiler
--- a/doc/source/dev/api-dev/array_api.rst
+++ b/doc/source/dev/api-dev/array_api.rst
@@ -362,7 +362,8 @@ but would not produce the correct dtype behavior::
   z, p, b, a = map(xp.asarray, (z, p, b, a))
 
   # With float64 inputs, the outputs bp and ap will be of dtype
-  # float64.
+  # float64. Note that the parameter k is a Python scalar which does
+  # not impact output dtype for NumPy >= 2.0.
   bp, ap = zpk2tf(z, p, k)
   # xp_assert_close checks for matching dtype. Due to the way the
   # code was written above, zpk2tf is not tested with float32 inputs
--- a/doc/source/dev/api-dev/array_api.rst
+++ b/doc/source/dev/api-dev/array_api.rst
@@ -329,15 +329,14 @@ functions evaluated within a test, for the purpose of producing reference values
 inputs, round-trip calculations, etc. should instead use the NumPy backend. This helps
 ensure that any failures that occur on a backend actually relate to the function
 of interest, and avoids the need to skip backends due to lack of support for
-functions other than ``f`` when ``f`` is supported.
+functions other than ``f``.
 
 To help facillitate such backend isolation, there is a function ``_xp_copy_to_numpy``
 in ``scipy._lib._array_api`` which can copy an arbitrary ``xp`` array to a NumPy
 array, bypassing any device transfer guards, while preserving dtypes. It is essential
-that this function is only used in tests and in tests only for the purpose of isolating
-use of alternative backends to only the function being tested. Attempts to copy a
-device array to NumPy outside of tests should fail, because otherwise it can become
-opaque whether a function is working on GPU or not.
+that this function is only used in tests for functions other than the one being
+tested. Attempts to copy a device array to NumPy outside of tests should fail,
+because otherwise it can become opaque whether a function is working on GPU or not.
 
 When attempting to isolate use of alternative backends to a particular function, one
 must be mindful that PyTorch allows for setting a default dtype, and SciPy is tested
--- a/doc/source/dev/api-dev/array_api.rst
+++ b/doc/source/dev/api-dev/array_api.rst
@@ -321,31 +321,33 @@ Note that there is a GitHub Actions workflow which tests with array-api-strict,
 PyTorch, and JAX on CPU.
 
 Testing Practice
-----------------
+````````````````
 
-It's best if individual tests for a function ``f`` using the ``xp`` fixture restrict
-use of alternative backends to only the function ``f`` being tested. Other functions
-evaluated within a test, for the purpose of producing reference values, inputs,
-round-trip calculations, etc. should instead use the NumPy backend. This helps
+It's best if individual tests using the ``xp`` fixture restrict
+use of alternative backends to only a single function ``f`` being tested. Other
+functions evaluated within a test, for the purpose of producing reference values,
+inputs, round-trip calculations, etc. should instead use the NumPy backend. This helps
 ensure that any failures that occur on a backend actually relate to the function
 of interest, and avoids the need to skip backends due to lack of support for
-functions other than `f` when `f` is supported.
+functions other than ``f`` when ``f`` is supported.
 
 To help facillitate such backend isolation, there is a function ``_xp_copy_to_numpy``
 in ``scipy._lib._array_api`` which can copy an arbitrary ``xp`` array to a NumPy
 array, bypassing any device transfer guards, while preserving dtypes. It is essential
-that this function is only used in tests for the purpose of using the NumPy backend
-for functions other than the function being tested.
+that this function is only used in tests and in tests only for the purpose of isolating
+use of alternative backends to only the function being tested. Attempts to copy a
+device array to NumPy outside of tests should fail, because otherwise it can become
+opaque whether a function is working on GPU or not.
 
 When attempting to isolate use of alternative backends to a particular function, one
 must be mindful that PyTorch allows for setting a default dtype, and SciPy is tested
-with both default dtype ``float32`` (set the environment variable ``SCIPY_DEFAULT_DTYPE``
-to ``float32``) and ``float64``. Tests using the ``xp`` fixture often rely on
+with both default dtype ``float32`` and ``float64`` (this is controlled with the
+environment variable ``SCIPY_DEFAULT_DTYPE``). Tests using the ``xp`` fixture rely on
 ``xp.asarray`` producing arrays with the default dtype when list input is given and
-no explicit dtype specified. This means that if a test involves taking input arrays,
+no explicit dtype specified. This means that if a test involves taking input arrays
 and passing them to a function other than the one being tested in order to produce
-inputs for the function being tested, the following may appear natural to write,
-but will not produce the correct dtype behavior::
+inputs for the function being tested, the following may appear natural to write
+but would not produce the correct dtype behavior::
 
   # conjugate pairs give real-coeff num & den
   z = np.asarray([1j, -1j, 2j, -2j])
--- a/doc/source/dev/api-dev/array_api.rst
+++ b/doc/source/dev/api-dev/array_api.rst
@@ -371,7 +371,7 @@ but would not produce the correct dtype behavior::
   xp_assert_close(b, bp)
   xp_assert_close(a, ap)
 
-One should instead construct all inputs as ``xp`` arrays and then copy to
+One could instead construct all inputs as ``xp`` arrays and then copy to
 NumPy arrays in order to ensure the default dtype is respected::
 
   # calls to xp.asarray will respect the default dtype.
--- a/doc/source/dev/api-dev/array_api.rst
+++ b/doc/source/dev/api-dev/array_api.rst
@@ -391,6 +391,13 @@ NumPy arrays in order to ensure the default dtype is respected::
   xp_assert_close(b, bp)
   xp_assert_close(a, ap)
 
+The above suggestions mainly concern the bulk conversion of existing tests to
+work with alternative backends. There may be specific cases where such isolation
+is neither necessary nor desirable. Maintainers have discretion to write and
+accept tests as they see fit, so long as they take care to investigate that use
+of the alternative backend across multiple functions in the same test is sound,
+and will not require unnecessary skips or xfails to be added.
+
 
 Testing the JAX JIT compiler
 ----------------------------
--- a/doc/source/dev/api-dev/array_api.rst
+++ b/doc/source/dev/api-dev/array_api.rst
@@ -323,13 +323,19 @@ PyTorch, and JAX on CPU.
 Testing Practice
 ````````````````
 
-It's best if individual tests using the ``xp`` fixture restrict
-use of alternative backends to only a single function ``f`` being tested. Other
-functions evaluated within a test, for the purpose of producing reference values,
-inputs, round-trip calculations, etc. should instead use the NumPy backend. This helps
-ensure that any failures that occur on a backend actually relate to the function
-of interest, and avoids the need to skip backends due to lack of support for
-functions other than ``f``.
+It's important that for any supported function ``f``, there exist tests using
+the ``xp`` fixture that restrict use of alternative backends to only the function
+``f`` being tested. Other functions evaluated within a test, for the purpose of
+producing reference values, inputs, round-trip calculations, etc. should instead
+use the NumPy backend. This helps ensure that any failures that occur on a backend
+actually relate to the function of interest, and avoids the need to skip backends
+due to lack of support for functions other than ``f``. Property based integration
+tests which check that some invariant holds using the same alternative backend
+across different functions can also have value, giving a window into the general
+health of backend support for a module, but in order to ensure the test suite
+actually reflects the state of backend support for each function, it's vital to
+have tests which isolate use of the alternative backend only to the function being
+tested.
 
 To help facilitate such backend isolation, there is a function ``_xp_copy_to_numpy``
 in ``scipy._lib._array_api`` which can copy an arbitrary ``xp`` array to a NumPy
@@ -390,13 +396,6 @@ NumPy arrays in order to ensure the default dtype is respected::
   bp, ap = zpk2tf(z, p, k)
   xp_assert_close(b, bp)
   xp_assert_close(a, ap)
-
-The above suggestions mainly concern the bulk conversion of existing tests to
-work with alternative backends. There may be specific cases where such isolation
-is neither necessary nor desirable. Maintainers have discretion to write and
-accept tests as they see fit, so long as they take care to investigate that use
-of the alternative backend across multiple functions in the same test is sound,
-and will not require unnecessary skips or xfails to be added.
 
 
 Testing the JAX JIT compiler
--- a/scipy/_lib/_array_api.py
+++ b/scipy/_lib/_array_api.py
@@ -166,7 +166,7 @@ def _xp_copy_to_numpy(x: Array) -> np.ndarray:
     -------
     ndarray
     """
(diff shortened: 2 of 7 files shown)


scipy/signal/tests/test_filter_design.py
```diff
@@ -1103,20 +1104,26 @@ def test_freqz_sos_basic(self, xp):
...
         b, a = ellip(3, 1, 30, (0.2, 0.3), btype='bandpass')
         sos = ellip(3, 1, 30, (0.2, 0.3), btype='bandpass', output='sos')
-        b, a, sos = map(xp.asarray, (b, a, sos))   # XXX until ellip is converted
-
         w, h = freqz(b, a, worN=N)
+
+        w, h, sos = map(xp.asarray, (w, h, sos))
         w2, h2 = freqz_sos(sos, worN=N)
-        xp_assert_equal(w2, w)
+        if not (is_cupy(xp) or is_torch(xp)):
```
ev-br: Minor, trivial, not worth flushing the CI for: could use `xp_assert_close(..., atol=1e-15)` on all backends just as well, IMO.

scipy/signal/tests/test_filter_design.py
```diff
@@ -1598,6 +1604,9 @@ def test_basic(self, xp):
         assert_array_almost_equal(a_bs, xp.asarray([1, 0.18461, 0.17407]), decimal=5)
 
 
+@skip_xp_backends(
+    cpu_only=True, exceptions=["cupy"], reason="not array api agnostic"
+)
```
ev-br, line 1603: Note to self: `bilinear` uses `np.polynomial.Polynomial`.

scipy/_lib/_array_api.py
```diff
@@ -147,6 +147,35 @@ def xp_copy(x: Array, *, xp: ModuleType | None = None) -> Array:
     return _asarray(x, copy=True, xp=xp)
 
 
+def xp_copy_to_numpy(x: Array) -> np.ndarray:
+    """Copies a possibly on device array to a NumPy array.
```
ev-br, line 151: FWIW, I was resisting the urge to use function like this for a while. Which is at least partially responsible for the mess this PR cleans up. So  the time has come, I guess :-). 

Still, it'd be great to add a doc note that this function should be only used as a last resort, only in tests, and almost always to only cover for partially compliant usage in functions under test. (`assert_array_equal_nulp` is an exception, and that function itself should be only very sparingly used if ever IMO) 

Or maybe even consider adding a leading underscore to signal that it's not for general use?

scipy/signal/tests/test_filter_design.py
```diff
@@ -1894,55 +1898,57 @@ def dB(x):
 
 
 @pytest.mark.skipif(DEFAULT_F32, reason="XXX needs figuring out")
+@skip_xp_backends(
+    cpu_only=True, exceptions=["cupy"], reason="not array api agnostic"
```
ev-br: Is the the lack of array api agnosticism due to torch.convolve being cpu-only (if so, could you please add it to the note) or is there more to it?

ev-br (approved):
Let me try double-checking I got the main idea right. 
Consider a test function, which aims to test a public function `func` and also uses a different public function, `other_func`, 

```
def test_func(xp):
    args  = np.asarray(42)   #  np, not xp
    expected = # ... 

    # compute in numpy, convert to xp
    internal_vars_np = other_func(args)  
    internal_vars = xp.asarray(internal_vars_np)

    # use xp-namespaced values below
    result = func(internal_vars)
    xp_assert_close(result, expected)
```

the idea is to evaluate `other_func` with numpy arrays as inputs, and only keep `func` receiving xp-namespaced arrays --- is this correct?

Being more principled in tests is probably a good thing, so +1 from me.

The pattern certainly needs documenting as a recommendation in the array api section of the developer docs. This can be done in either this PR or in a follow-up though.
Ditto for my several small comments below: the main one being we should IMO heavily limit the use of the new `xp_copy_to_numpy` function, both in its docstring and in the to-be-updated developer docs.

scipy/_lib/_array_api.py
steppi: Yeah, `xp_copy_to_numpy` is the devil itself if used outside tests since it just lets you bulldoze through transfer guards. I was thinking of hacking it so that it warns if called outside of a test file, like what is done here https://github.com/scipy/scipy/blob/51dfbcc7b38b251b274cdeb1d6e028a83f3f0cb8/scipy/_lib/tests/test_warnings.py#L56-L59 for ignoring warnings. I think I will go ahead and do that actually, and document this intention better too.

scipy/signal/tests/test_filter_design.py
steppi: I'll check and make the note more specific.

steppi:
> Let me try double-checking I got the main idea right. Consider a test function, which aims to test a public function `func` and also uses a different public function, `other_func`,
> 
> ```
> def test_func(xp):
>     args  = np.asarray(42)   #  np, not xp
>     expected = # ... 
> 
>     # compute in numpy, convert to xp
>     internal_vars_np = other_func(args)  
>     internal_vars = xp.asarray(internal_vars_np)
> 
>     # use xp-namespaced values below
>     result = func(internal_vars)
>     xp_assert_close(result, expected)
> ```
> 
> the idea is to evaluate `other_func` with numpy arrays as inputs, and only keep `func` receiving xp-namespaced arrays --- is this correct?
> 

Yes, you have it right. This way tests will only fail if there's something wrong with the function itself, not other functions that happen to be used in the tests. 

> Being more principled in tests is probably a good thing, so +1 from me.
> 
> The pattern certainly needs documenting as a recommendation in the array api section of the developer docs. This can be done in either this PR or in a follow-up though. Ditto for my several small comments below: the main one being we should IMO heavily limit the use of the new `xp_copy_to_numpy` function, both in its docstring and in the to-be-updated developer docs.

I can update the docs in this PR, and will take steps to ensure `xp_copy_to_numpy` is not used outside of test code.
Thanks for the review!

scipy/_lib/_array_api.py
ev-br: FWIW, I don't think there's a *strong* need to be fancy here. Just documenting it and maybe adding a leading underscore would do it IMO

scipy/_lib/_array_api.py
```diff
@@ -147,6 +147,35 @@ def xp_copy(x: Array, *, xp: ModuleType | None = None) -> Array:
...
+    if is_numpy(xp):
+        return x.copy()
+    if is_cupy(xp):
+        return x.get()
+    if is_torch(xp):
+        return x.cpu().numpy()
+    # Fall back to np.asarray. This will work for jax.numpy and
+    # dask.array. If new backends are added, they may require
+    # explicit handling..
+    return np.asarray(x, copy=True)
```
lucascolley, line 188: sklearn has an additional case here:

https://github.com/scikit-learn/scikit-learn/blob/bde39c1afc8de16ff50ff743b40b89acfd44dd4c/sklearn/utils/_array_api.py#L840-L841

scipy/_lib/_array_api.py
```diff
@@ -147,6 +147,35 @@ def xp_copy(x: Array, *, xp: ModuleType | None = None) -> Array:
...
+    
+    Parameters
+    ----------
+    x : array
+
+    Returns
+    -------
+    ndarray
+    """
+    xp  = array_namespace(x)
```
lucascolley:
```suggestion
    xp = array_namespace(x)
```

scipy/_lib/_array_api.py
steppi: Interesting. I didn't realize `array_api_strict` allowed support for different devices (https://github.com/data-apis/array-api-strict/pull/59). Sure, it's easy enough to throw that in.

steppi:
OK. I've added a comment to `xp_copy_to_numpy` stating that it's doc only and added a leading underscore. I also added a note on testing practices to the array api docs.

There was a failure in `TestGroupDelay` on CuPy. It doesn't fail for me locally, but I noticed that I actually failed to ensure backend isolation for those tests. Instead of loosening the tolerances, I added the backend isolation, and will see if those tests pass. 

I removed the extra space @lucascolley pointed out, and added a case to `_xp_copy_to_numpy` for `array_api_strict` to handle alternative devices. Thanks @lucascolley. 

I still need to track down more specific reasons for needing some xp_backend skips. There's also something else I need to look into that I will make another comment about.

doc/source/dev/api-dev/array_api.rst
```diff
@@ -320,6 +320,65 @@ backend, use: ``SCIPY_DEVICE=mps spin test -b torch``.
...
+
+When attempting to isolate use of alternative backends to a particular function, one
+must be mindful that PyTorch allows for setting a default dtype, and SciPy is tested
+with both default dtype ``float32`` and ``float64`` (this is controlled with the
+environment variable ``SCIPY_DEFAULT_DTYPE``). Tests using the ``xp`` fixture rely on
+``xp.asarray`` producing arrays with the default dtype when list input is given and
+no explicit dtype specified. This means that if a test involves taking input arrays
+and passing them to a function other than the one being tested in order to produce
+inputs for the function being tested, the following may appear natural to write
+but would not produce the correct dtype behavior::
```
steppi, line 355: Before this is merged I want to go through the changes and make sure I actually followed my own advice here.

scipy/signal/tests/test_filter_design.py
steppi: It turns out I was too pessimistic here. This only fails for `bandstop` due to use of `scipy.optimize.fminbound`, the same holds for all other order selection functions.

steppi: Except for `ellipord` because it uses `special.ellipk` and `special.ellipkm1`. These are in CuPy and SciPy can dispatch to CuPy for them, but `ellipord` isn't implemented in such a way to make use of that right now. I can do that in a follow-up.

doc/source/dev/api-dev/array_api.rst
ev-br: I'm probably being dense, but I don't understand the example that follows.

The "bad" version: constructs numpy f64 arrays `z` and `p`, then generates numpy f64 arrays `b` and `a`, converts them to `xp`, so that `z, p, b, a` are f64 xp arrays, which it feeds to `zpk2tf`.

The "good" version (assuming `SCIPY_DEFAULT_DTYPE=float32`): construct xp f32 arrays  `z` and `p`, converts them to numpy in `_xp_copy_to_numpy` so that `z` and `p` are f32 numpy arrays, feeds them to `np.poly` --- I've no idea if that returns f32 or f64 `b` and `a` for f32 inputs without checking  --- and then feeds them to `zpk2tf`. At this stage, I'm really confused as to
- what does this example mean to demonstrate?
- what is the correct behavior we're after, I'm guessing we want to compate `a, b` with `ap, bp`?
- how does the bad version get bad and how does the good version do the right thing?

doc/source/dev/api-dev/array_api.rst
```diff
@@ -320,6 +320,65 @@ backend, use: ``SCIPY_DEVICE=mps spin test -b torch``.
...
+
+  # conjugate pairs give real-coeff num & den
+  z = np.asarray([1j, -1j, 2j, -2j])
+  # shouldn't need elements of pairs to be adjacent
```
ev-br: Not sure these comments are helpful outside of the specific test they are taken from.

doc/source/dev/api-dev/array_api.rst
```diff
@@ -320,6 +320,65 @@ backend, use: ``SCIPY_DEVICE=mps spin test -b torch``.
...
+  p = np.asarray([1+1j, 3-100j, 3+100j, 1-1j])
+  k = 23
+
+  # np.poly should do the right thing, but be explicit about
```
ev-br: "the right thing" relates to something in that specific test, what is it here?

ev-br (commented):
I tried looking through the added documentation example. Sadly that made me only more confused. I'll try to re-read it in the morning.

scipy/signal/tests/test_filter_design.py
ev-br: Ah, fminbound. IIRC it's pure python, so can be array api converted. elliptic filters are a bit of a mess, yes. Both are fine to defer to a follow-up IMO.

steppi:
Thanks @ev-br. I chose a bad example there in the docs for explaining handling default dtype float32. I tried to grab an example from real code, but it contains a lot of extraneous information, and is missing important information (like that we will later check with `xp_assert_close` which will check dtypes). I'll come up with a better one.

steppi:
@ev-br, does this make more sense?

```
  # z, p, k will have dtype float64 regardless of the value of
  # SCIPY_DEFAULT_DTYPE
  z = np.asarray([1j, -1j, 2j, -2j])
  p = np.asarray([1+1j, 3-100j, 3+100j, 1-1j])
  k = 23

  # np.poly will preserve dtype
  b = k * np.poly(z_np).real
  a = np.poly(p_np).real
  # Input arrays z, p, and reference outputs b, a will all have
  # dtype float64.
  z, p, b, a = map(xp.asarray, (z, p, b, a))

  # With float64 inputs, the outputs bp and ap will be of dtype
  # float64.
  bp, ap = zpk2tf(z, p, k)
  # xp_assert_close checks for matching dtype. Due to the way the
  # code was written above, zpk2tf is not tested with float32 inputs
  # when SCIPY_DEFAULT_DTYPE is float32.
  xp_assert_close(b, bp)
  xp_assert_close(a, ap)

One should instead construct all inputs as ``xp`` arrays and then copy to
NumPy arrays in order to ensure the default dtype is respected::

  # calls to xp.asarray will respect the default dtype.
  z = xp.asarray([1j, -1j, 2j, -2j])
  p = xp.asarray([1+1j, 3-100j, 3+100j, 1-1j])
  k = 23

  # _xp_copy_to_numpy preserves dtype, as does np.poly.
  b = k * np.poly(_xp_copy_to_numpy(z)).real
  a = np.poly(_xp_copy_to_numpy(p)).real
  # b and a will have dtype float32
  b, a = map(xp.asarray, (b, a))

  # zpk2tf is tested with float32 inputs when SCIPY_DEFAULT_DTYPE=float32
  # as intended.
  bp, ap = zpk2tf(z, p, k)
  xp_assert_close(b, bp)
  xp_assert_close(a, ap)
```

ev-br:
Much nicer, thanks Albert! Your new comments work wonders. The example is a little long, but I suppose it's great to have a real example from real tests. 

Now that we're at it, what would be the guidance for test inputs generated with `numpy.random`? These are certainly f64, so they apparently have the same problem?

steppi:
> Much nicer, thanks Albert! Your new comments work wonders. The example is a little long, but I suppose it's great to have a real example from real tests.
> 
> Now that we're at it, what would be the guidance for test inputs generated with `numpy.random`? These are certainly f64, so they apparently have the same problem?

Right, good point about `numpy.random`. I knew I'd forgotten something. I guess such inputs just need to be cast to `float32` if needed. 

We could cook up a minimal example, but yeah, I liked the idea of having the example come from a real test. I can still shorten it if you like.

lucascolley:
Once we start seeing boilerplate surrounding rng which can be factored out, I think sharing it on an issue in https://github.com/NeilGirdhar/array-api-rng would be a good idea.

I think that repo was born out of a wider scope than just rng in tests, but IIUC reusable helpers for mixing numpy.random with tests of standard arrays are definitely in scope.

steppi:
> Once we start seeing boilerplate surrounding rng which can be factored out, I think sharing it on an issue in https://github.com/NeilGirdhar/array-api-rng would be a good idea.
> 
> I think that repo was born out of a wider scope than just rng in tests, but IIUC reusable helpers for mixing numpy.random with tests of standard arrays are definitely in scope.

That's a good suggestion. I started adding the boilerplate here, before realizing that it's kind of out-of-scope for this PR. I'll submit an issue over there.

ev-br:
Where do we stand in this PR, folks? Albert, you mentioned you wanted to make second look-through? Somebody feels like reviewing? I think we can merge this one quickly and iterate if needed, to unblock Albert's queue. Unless you're prepared to send a "stack" of PRs Albert and rebase the next ones once they land.

steppi:
> Where do we stand in this PR, folks? Albert, you mentioned you wanted to make second look-through? Somebody feels like reviewing? I think we can merge this one quickly and iterate if needed, to unblock Albert's queue. Unless you're prepared to send a "stack" of PRs Albert and rebase the next ones once they land.

I think it's good to go on my end now. Feel free to merge.

doc/source/dev/api-dev/array_api.rst
```diff
@@ -320,6 +320,77 @@ backend, use: ``SCIPY_DEVICE=mps spin test -b torch``.
...
+  # not impact output dtype for NumPy >= 2.0.
+  bp, ap = zpk2tf(z, p, k)
+  # xp_assert_close checks for matching dtype. Due to the way the
+  # code was written above, zpk2tf is not tested with float32 inputs
+  # when SCIPY_DEFAULT_DTYPE is float32.
+  xp_assert_close(b, bp)
+  xp_assert_close(a, ap)
+
+One should instead construct all inputs as ``xp`` arrays and then copy to
+NumPy arrays in order to ensure the default dtype is respected::
```
lucascolley: I think some tests accomplish similar things using https://data-apis.org/array-api-extra/generated/array_api_extra.default_dtype.html, or its predecessor in SciPy.

Probably fine to leave the text as is, just FYI.

scipy/_lib/_array_api.py
```diff
@@ -147,6 +147,47 @@ def xp_copy(x: Array, *, xp: ModuleType | None = None) -> Array:
     return _asarray(x, copy=True, xp=xp)
 
 
+def _xp_copy_to_numpy(x: Array) -> np.ndarray:
```
lucascolley, line 150: Thinking long-term, I wonder where such a function would belong. Maybe a new `xpx.testing.numpy` namespace.

doc/source/dev/api-dev/array_api.rst
```diff
@@ -320,6 +320,77 @@ backend, use: ``SCIPY_DEVICE=mps spin test -b torch``.
...
+
+It's best if individual tests using the ``xp`` fixture restrict
+use of alternative backends to only a single function ``f`` being tested. Other
+functions evaluated within a test, for the purpose of producing reference values,
+inputs, round-trip calculations, etc. should instead use the NumPy backend. This helps
+ensure that any failures that occur on a backend actually relate to the function
+of interest, and avoids the need to skip backends due to lack of support for
+functions other than ``f``.
+
+To help facillitate such backend isolation, there is a function ``_xp_copy_to_numpy``
```
tylerjereddy:
```suggestion
To help facilitate such backend isolation, there is a function ``_xp_copy_to_numpy``
```

scipy/signal/tests/test_filter_design.py
```diff
@@ -1609,61 +1612,59 @@ def test_exceptions(self):
...
+    @xfail_xp_backends("cupy", reason="https://github.com/cupy/cupy/issues/9404")
     def test_ignore_leading_zeros(self, xp):
         # regression for gh-6606
         # results shouldn't change when leading zeros are added to
         # input numerator or denominator
-        b = [0.14879732743343033]
-        a = [1, 0.54552236880522209, 0.14879732743343033]
-        b, a = map(xp.asarray, (b, a))
+        b = xp.asarray([0.14879732743343033])
+        a = xp.asarray([1, 0.54552236880522209, 0.14879732743343033])
```
tylerjereddy: This was just a cosmetic change, right? I'm reading through your diff and sometimes I see `np`->`xp` coercion changes to respect the "global" dtype initialization, but some of these are just `xp`->`xp`?

scipy/signal/tests/test_filter_design.py
```diff
@@ -1895,73 +1895,75 @@ def dB(x):
...
-        b, a = butter(N, Wn, 'highpass', False)
+        N, Wn = buttord(xp.asarray(wp), ws, rp, rs, False)
+        b, a = butter(N, _xp_copy_to_numpy(Wn), 'highpass', False)
         w, h = freqz(b, a)
-        w /= xp.pi
-        assert xp.all(-rp < dB(h[wp <= w]))
-        assert xp.all(dB(h[w <= ws]) < -rs)
+        w /= np.pi
+        assert np.all(-rp < dB(h[wp <= w]))
+        assert np.all(dB(h[w <= ws]) < -rs)
```
tylerjereddy, line 1928: The waters seem a bit murky here... `butter()` is processing your "actual" array output value from `buttord`, and then `freqz` processes the "actual" output even more, and then you are doing assertions on the transformed "actual" output.

While I think you've sold me on the separation of concerns that comes from generating "expected" values without transiting through SciPy functions that receive `xp` arrays, when it comes to chains of testing logic that are processing the "actual" values before they are asserted on, I'm less convinced that the decision to coerce the control flow chain to NumPy is decisively always the right one. A more convincing guidance might be to avoid using the SciPy functions for "actual" transformations at all here, but that could be quite time consuming to substitute for wrt whatever properties are being verified.

This is also murky because the only real danger we are protecting against seems to be developer "laziness," right? In other words, the main reason for suggesting that we don't transit through SciPy functions not under test with `xp` arrays is that a developer is easily tempted to just mark the test with a skip that isn't needed, but there is no actual problem beyond the need to investigate in those cases, right? That seems to make the calculus a bit messier for me when manipulating the "actual" values. This is probably "ok," but I'm just pointing this out because this is getting quite complex rule-wise even for someone who enjoys writing tests.

scipy/signal/tests/test_filter_design.py
steppi: Yeah, this seems to have been an inadvertent change. What likely happened is I changed it to something else, and then manually changed it back, but in a slightly different way. I can go through the diff and get rid of noise like this.

doc/source/dev/api-dev/array_api.rst
steppi: Changing "should" to "could" in the text above seems like it would be a reasonable way to address this.

scipy/signal/tests/test_filter_design.py
steppi: OK, I see where you are coming from. We don't want to make it too burdensome to write individual tests. My perspective is colored by working on the entire test suit at once, to add `xp_capabilities` decorators and patch problems that come up with alternative backend testing. I came across a fair number of unnecessary skips and xfails in `signal` and essentially hit upon a solution which was convenient for eliminating unnecessary skips at scale. But with the more usual development pattern, where one typically works on one or a few tests at a time, I understand it could be annoying to have to remember all kinds of little rules related to ideal testing for alternative backends. One risk here is that people will just skip writing tests for alternative backends, because they don't want to deal with the hassle of the rules.

I think there's room to allow more flexibility, and the content I added to the docs could be updated to reflect that. I'm still sold on generally preferring nontrivial calculations both upstream and downstream to the call of the function being tested being done with NumPy, but admit that this might not decisively be the best thing to do in all cases. I think something can be done about the ergonomics to make this kind of isolation less painful to enforce. I'll think about it a little more.

steppi: @tylerjereddy. I added this paragraph softening the recommendation. Do you think this improves things?

*The above suggestions mainly concern the bulk conversion of existing tests to
work with alternative backends. There may be specific cases where such isolation
is neither necessary nor desirable. Maintainers have discretion to write and
accept tests as they see fit, so long as they take care to investigate that use
of the alternative backend across multiple functions in the same test is sound,
and will not require unnecessary skips or xfails to be added.*

tylerjereddy: Sure, I'm just a bit more nervous about manipulating "actual" values.

steppi: OK, I think I understand. Does the following make sense? When dealing with tests of a particular function against reference values, they are best thought of as unit tests, and it's natural to isolate use of the alternative backend to only the function of interest. Property based tests like the ones you point out involving filter order selection functions together with filters and `freqz/freqs` really are better thought of as integration tests, and by isolating the alternative backend to only one function being tested, we are forcing these to become less than ideal unit tests.

I think the problem arises because we are trying to adapt an existing test suite where some functions only have property based tests like this, and the result was that skips were added that wouldn't be necessary in a function specific unit test. This doesn't mean we should abandon integration tests for alternative backends though, like this PR suggests and implements. The key thing is that there have to be some tests that function as unit tests and actually test that a particular function works on alternative backends specifically. The underlying issue isn't necessarily about needing to isolate use of the alternative backend to one function for all tests, but rather that there need to be unit tests too, we can't have all property based integration tests.

steppi:
I've further softened the language on isolation of alternative backends. Instead stressing the importance that there exist tests with such isolation, rather than that all tests should have it, and casting the rest of that section as advice to help achieve such isolation, rather than hard requirements about how tests should be written.

I still have no qualms about isolating use of the alternative backend to only one specific function being tested for property based tests that use multiple functions. If we're checking that some invariant holds relating the values produced by various functions, I think it's useful to check that the invariant holds when replacing one particular function in the chain with the `xp` version. This helps when there are differences in the state of backend support across the functions used in a property based tests, but also when there are differences in numerical accuracy. If a property based test fails due to a tolerance not being met, and there's no isolation, the tests don't really tell us where the problem is. With isolation we can look at failures and know right away where the problem is; I ran into situations like this while working on `xp_capabilities` for `signal`, including in the test which originally sparked @tylerjereddy's comment. 

I think there are stylistic issues about writing such tests that could be improved, but I would hope that could be done incrementally and not block this PR. Beyond the annoyance of having to call `_xp_copy_to_numpy(xp.asarray(something))` in order to ensure the default dtype is respected, even for arrays involved in test which are never used on alternative backends, I also found a pattern where I would convert both the desired and actual values in a property based test from numpy arrays to xp arrays, just to get the more strict `xp_assert_close` checks to run. There should be a better pattern for this.

The failures do not appear related, nor were any of the failures from the previous commit surprisingly. I don't think I'd ever seen so many distinct unrelated failures pop up at the same time.

There's still some accidental cosmetic changes I made that I intend to undo. I'll do that now, but otherwise, I think this PR is ready, and hope can be merged. I have two stacked follow-ups coming. One for more consistent dtype behavior for filter functions, and the final one adding `xp_capabilities`.

doc/source/dev/api-dev/array_api.rst
```diff
@@ -320,6 +320,83 @@ backend, use: ``SCIPY_DEVICE=mps spin test -b torch``.
...
+``f`` being tested. Other functions evaluated within a test, for the purpose of
+producing reference values, inputs, round-trip calculations, etc. should instead
+use the NumPy backend. This helps ensure that any failures that occur on a backend
+actually relate to the function of interest, and avoids the need to skip backends
+due to lack of support for functions other than ``f``. Property based integration
+tests which check that some invariant holds using the same alternative backend
+across different functions can also have value, giving a window into the general
+health of backend support for a module, but in order to ensure the test suite
+actually reflects the state of backend support for each function, it's vital to
+have tests which isolate use of the alternative backend only to the function being
```
steppi, line 337:
```suggestion
have some tests which isolate use of the alternative backend only to the function being
```

maybe soften this further

scipy/signal/tests/test_filter_design.py
ev-br: My understanding here is that this is a pragmatic change. To summarize my reasoning:
- this function tests `buttord`, therefore `buttord` receives and returns xp-namespaced arrays,
- a purist "unit test" way would be to make assertions directly on the `buttord` outputs,
- these outputs are hard to interpret and assert on, so instead we transform `buttord` outputs via `butter`, and make interpretable assertions on these transformed values;
- to achieve "isolation" we convert buttord inputs to numpy

What are the alternatives:
- status quo: keep calling `butter` with xp-namespaced objects; this is inconvenient and messy because if `butter` has more limitations than `buttord`, we have to list `butter`'s limitations in `xp_capabilites` of `buttord` itself;
- rework the test to avoid using `butter` and make assertion directly on the outputs of `buttord`. This is way more work, and the result is harder to interpret.

All in all, I believe both alternatives are worse than what this PR does.

tylerjereddy:
I'm taking another look now

tylerjereddy:
Minor points for a follow-up maybe--the new docs in `array_api.rst` have a few undefined variables in the new examples like:

`git grep -E -i "z_np" doc`

`doc/source/dev/api-dev/array_api.rst:  b = k * np.poly(z_np).real`

The `xp_assert_close()` calls all have `actual` and `expected` order reversed I think. It may also be good to clarify what seems to be the main problem with the first/"bad" example--that `z, p, b, a = map(xp.asarray, (z, p, b, a))` is not going to coerce f64 to f32 (i.e., I suppose this only happens if `asarray()` receives non-array input so the default dtype is used). That seems to be the only real issue with that example I think, and is arguably a bit subtle.

Don't push any revision in though.. I'm still going through..

tylerjereddy:
The first example could also be made to work using Lucas' default dtype `xpx` approach for the mapped `asarray` call, I think? So it is less than a line change away from working?

steppi:
> The first example could also be made to work using Lucas' default dtype `xpx` approach for the mapped `asarray` call, I think? So it is less than a line change away from working?

OK, yeah. I agree, that is a better way to do it.

tylerjereddy:
For an eventual follow-up, I suspect the usage of `assert_array_almost_equal` rather than one of its "xp equivalents" for handling xp-array comparisons in `test_fir_filter_design.py` might make sense.

tylerjereddy:
For others to maybe check in a follow-up, the CuPy early `return` in `UpFIRDnCase` made me a little nervous vs. catching problems in a decorated-pytest way (xfail or whatever, maybe I didn't understand why that shim couldn't be handled that way). `test_upfirdn.py` looked a little weird/boiler-plate heavy in general.

Beyond that, I've now read the full diff through and confirmed the CI failures are not related (they're well described in another issue). I went through all review comments as well, and don't think that there are any blockers left. There were 1-2 `xp.asarray` coercions that looked unnecessary, but not worth commenting. Ralf was also basically +1 in the matching issue, for the general concept, so let's move forward, thanks Albert et al.

tylerjereddy:
On balance, I decided not to squash the commits because they were all unique/formatted per our standard, though 47 is a lot.

steppi:
> For an eventual follow-up, I suspect the usage of `assert_array_almost_equal` rather than one of its "xp equivalents" for handling xp-array comparisons in `test_fir_filter_design.py` might make sense.

Sorry, I'm a little dense. Why is that?

There's a comment that says it's a backwards compatible replacement, and for new code we should prefer `xp_assert_close`. 

https://github.com/scipy/scipy/blob/babf6c61effcf5d16da42611576caccbf53080a7/scipy/_lib/_array_api.py#L316-L322C18

```python
def assert_array_almost_equal(actual, desired, decimal=6, *args, **kwds):
    """Backwards compatible replacement. In new code, use xp_assert_close instead.
    """
    rtol, atol = 0, 1.5*10**(-decimal)
    return xp_assert_close(actual, desired,
                           atol=atol, rtol=rtol, check_dtype=False, check_shape=False,
                           *args, **kwds)
```

I admit that my knowledge of this stuff isn't that the level where I can appreciate the difference, and I'm just kind of mindlessly following the recommendations.

tylerjereddy:
Sorry, that's what I mean, we should get rid of `assert_array_almost_equal` for two reasons--it is doc-discouraged and shouldn't be used for xp long term

steppi:
Thanks @tylerjereddy!

tylerjereddy:
I also didn't catch that we're patching over the NumPy version of `assert_array_almost_equal` I guess, also a bit confusing

Issue #19460, opened 2023-11-01
BUG: genextreme.fit not working with method='MM'


MarceloLago-NOAA:
### Describe your issue.

The fitting functions well by default (i.e., with method='MLE') but not with method='MM'.

### Reproducing Code Example

```python
from scipy.stats import genextreme
data= [0.87, 0.87, 1.29, 1.5 , 1.7 , 0.66, 1.5 , 0.5 , 1.  , 1.25, 2.3 ,
       1.03, 2.85, 0.68, 1.74, 1.94, 0.63, 2.04, 1.2 , 0.64, 2.05, 0.97,
       2.81, 1.02, 2.76, 0.86, 1.36, 1.29, 1.68, 0.72, 1.67, 1.15, 3.26,
       0.93, 0.83, 0.91, 0.92, 2.32, 1.12, 3.21, 1.23, 1.22, 1.29, 2.08,
       0.64, 2.83, 2.68, 1.77, 0.69, 1.69, 0.7 , 1.83, 2.25, 1.23, 1.17,
       0.94, 1.22, 0.76, 0.69, 0.48, 1.04, 2.49, 1.38, 1.57, 1.79, 1.59,
       1.3 , 1.54, 1.07, 1.03, 0.76, 2.35, 2.05, 2.02, 2.36, 1.59, 0.97,
       1.63, 1.66, 0.94, 1.45, 1.26, 1.25, 0.68, 2.96, 0.8 , 1.16, 0.82,
       0.64, 0.87, 1.33, 1.28, 1.26, 1.19, 1.24, 1.12, 1.45, 1.03, 1.37,
       1.4 , 1.35, 1.28, 1.04, 1.31, 0.87, 0.96, 2.55, 1.72, 1.05, 1.15,
       1.73, 1.03, 1.53, 2.41, 1.36, 2.08, 0.92, 0.73, 1.56, 1.94, 0.78]
# MLE Method
res1 = genextreme.fit(data) # runs well without error
# MM Method
res2 = genextreme.fit(data, method='MM') # it gives an error
```

### Error message

```shell
ValueError: Method of moments encountered a non-finite distribution moment and cannot continue. Consider trying method='MLE'.
```

### SciPy/NumPy/Python version and system information

```shell
1.11.3 1.24.4 sys.version_info(major=3, minor=10, micro=12, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    detection method: pkgconfig
    found: true
[...]

PR #19475, opened 2023-11-05
ENH: stats.lmoment: add function to calculate sample L-moments

mdhaber:
#### Reference issue
gh-19460

#### What does this implement/fix?
The idea of fitting distributions to data with method of L-moments has been tossed around for a while, and gh-19460 brought it back up again. This PR adds a function to compute sample L-moments, which is an essential step of the method of L-moments.

#### Additional information
No rush here; I wrote the essential part of the code in https://github.com/scipy/scipy/issues/19460#issuecomment-1793547120 so I figured I might as well submit a PR.

This needs an email to the mailing list before merge. ([Done](https://mail.python.org/archives/list/scipy-dev@python.org/thread/N6FZYLSHV6Y3OLJV5KKKNVERAZPLWRPO/) 3/11/2024.)

I can see arguments for the name being `l_moment` instead of `lmoment`. If there is strong support for changing it, that's OK with me.

PR #23766, opened 2025-10-13
ENH: stats.ContinuousDistribution.lmoment: add population L-moments

mdhaber:
Adds `stats.ContinuousDistribution.lmoment` for computing population L-moments.

```python3
import numpy as np
from scipy import stats, special

Gamma = stats.make_distribution(stats.gamma)
a, b = 3, 0.6
X = b * Gamma(a=a)

res = X.lmoment(2)
ref = np.pi**-0.5 * b * special.gamma(a + 0.5) / special.gamma(a)
print(res, ref)  # 0.5624999999999997 0.5625

res = X.lmoment(3, standardized=True)
ref = 6 * special.betainc(a, 2*a, 1/3) - 3
print(res, ref)  # 0.19067215363511616 0.1906721536351177
```
<details>
Reference values from [Hosking 1990](https://academic.oup.com/jrsssb/article/52/1/105/7027905?login=false).

image

As another example, from gh-19460:

```python3
import numpy as np
from scipy import stats, optimize

data = [0.87, 0.87, 1.29, 1.5, 1.7, 0.66, 1.5, 0.5, 1., 1.25, 2.3,
        1.03, 2.85, 0.68, 1.74, 1.94, 0.63, 2.04, 1.2, 0.64, 2.05, 0.97,
        2.81, 1.02, 2.76, 0.86, 1.36, 1.29, 1.68, 0.72, 1.67, 1.15, 3.26,
        0.93, 0.83, 0.91, 0.92, 2.32, 1.12, 3.21, 1.23, 1.22, 1.29, 2.08,
        0.64, 2.83, 2.68, 1.77, 0.69, 1.69, 0.7, 1.83, 2.25, 1.23, 1.17,
        0.94, 1.22, 0.76, 0.69, 0.48, 1.04, 2.49, 1.38, 1.57, 1.79, 1.59,
        1.3, 1.54, 1.07, 1.03, 0.76, 2.35, 2.05, 2.02, 2.36, 1.59, 0.97,
        1.63, 1.66, 0.94, 1.45, 1.26, 1.25, 0.68, 2.96, 0.8, 1.16, 0.82,
        0.64, 0.87, 1.33, 1.28, 1.26, 1.19, 1.24, 1.12, 1.45, 1.03, 1.37,
        1.4, 1.35, 1.28, 1.04, 1.31, 0.87, 0.96, 2.55, 1.72, 1.05, 1.15,
        1.73, 1.03, 1.53, 2.41, 1.36, 2.08, 0.92, 0.73, 1.56, 1.94, 0.78]


GEV = stats.make_distribution(stats.genextreme)

def f(x):
    c, loc, scale = x
    X = scale*GEV(c=c) + loc
    orders = [1, 2, 3]
    population_lmoments = [X.lmoment(i) for i in orders]
    sample_lmoments = stats.lmoment(data, orders, standardize=False)
    return np.linalg.norm(sample_lmoments - population_lmoments)

bounds = [(None, None), (None, None), (0, None)]
res = optimize.minimize(f, [1, 1, 1], bounds=bounds, method='slsqp')
print(res.x)  # [-0.07493655  1.1081755   0.45759972]
```

</details>

#### Additional information
Some observations:

- The `standardize` parameter of `stats.lmoment` (for _sample_ L-moments) determines whether to compute L-moment ratios for order $\geq 3$. It does not affect the results for order $< 3$. Should we follow suit here, or raise for order $< 3$ (current behavior)? (Some might consider $\tau_2$ to be $\lambda_2 / \lambda_1$. Even if we say that `standardize=True` with `order=2` should return that, we have the same question with `order=1`.)
- When a random variable `X` is transformed (e.g. by `order_statistic`), its cache gets cleared. Transformation does not happen in-place, so this shouldn't happen. That can be fixed separately.


scipy/stats/_distribution_infrastructure.py
```diff
@@ -3381,6 +3392,77 @@ def logintegrand(x, order, logcenter, **params):
         return self._quadrature(logintegrand, args=(order, logcenter),
                                 params=params, log=True)
 
+    ### L-Moments
+
+    @_set_invalid_nan_property
+    def lmoment(self, order=1, *, standardized=False, method=None):
```
mdhaber:
```suggestion
    def lmoment(self, order=1, *, standardize=False, method=None):
```
For consistency for `stats.lmoment`. I don't think this is inconsistent with the fact that `ContinuousDistribution.moment` has `kind='standardized'`; one is used a verb, the other as an adjective.

scipy/stats/tests/test_continuous.py
```diff
@@ -1271,6 +1328,10 @@ def moment(self, order, kind='raw', *, a, b):
         assert_allclose(X.mean(method='formula'), X.mean(method='quadrature'))
         assert not X.mean(method='formula') == X.mean(method='quadrature')
 
+        assert_allclose(X.lmoment(method='formula'),
+                        X.lmoment(method='quadrature_icdf'))
+        assert not X.lmoment(method='formula') != X.lmoment(method='quadrature_icdf')
```
mdhaber:
```suggestion
        assert not X.lmoment(method='formula') == X.lmoment(method='quadrature_icdf')
```

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3381,6 +3392,77 @@ def logintegrand(x, order, logcenter, **params):
...
+        min_order = 1 if not standardized else 3
+        order = self._validate_order(order, fname='lmoment', min_order=min_order)
```
mdhaber:
```suggestion
        min_order = 3 if standardized else 1
        fname = 'lmoment` with `standardize=True' if standardize else 'lmoment'
        order = self._validate_order(order, fname=fname, min_order=min_order)
```

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3381,6 +3392,77 @@ def logintegrand(x, order, logcenter, **params):
...
+
+    def _lmoment(self, order, *, standardized, method):
+        methods = self._lmoment_methods if method is None else {method}
+
+        lmoment = self._lmoment_dispatch(order, methods=methods, **self._parameters)
+        if lmoment is None:
+            return None
+
+        if standardized and lmoment is not None:
+            # todo: add general result: if `order==2` and `standardized`, return unity.
```
mdhaber:
```suggestion
```

scipy/stats/_probability_distribution.py
```diff
@@ -336,6 +337,119 @@ def moment(self, order, kind, *, method):
...
+        r"""L-moment or L-moment ratio of positive integer order.
+
+        The L-moment of order :math:`n` of a continuous random variable :math:`X` is:
+
+        .. math::
+
+            \lambda_n(X) = \frac{1}{n} \sum_{k=0}^{n-1} (-1)^{k} {{n-1}\choose k} E[X_{(n-k)}]
+
+        where :math:`X_{(1)}, \dots, X_{(r)}, \dots, X_{(n)}` are the order statistics
+        of an independent sample of size :math:`n`,
```
mdhaber:
```suggestion
        of an independent sample of size :math:`n`.
```

scipy/stats/_probability_distribution.py
```diff
@@ -336,6 +337,119 @@ def moment(self, order, kind, *, method):
...
+
+        The L-moment can also be expresed in terms of the random variable's inverse
```
mdhaber:
```suggestion
        The L-moment can also be expressed in terms of the random variable's inverse
```

scipy/stats/tests/test_continuous.py
```diff
@@ -914,12 +945,22 @@ class Test(ContinuousDistribution):
...
     with pytest.raises(ValueError, match=message):
         Test().moment(np.inf)
 
+    message = ("Argument `order` of `Test.lmoment` must be a "
+               "finite integer greater than or equal to 1.")
+    with pytest.raises(ValueError, match=message):
+        Test().lmoment(0)
+
+    message = ("Argument `order` of `Test.lmoment` must be a "
+               "finite integer greater than or equal to 3.")
```
mdhaber:
```suggestion
    message = ("Argument `order` of `Test.lmoment` with `standardize=True` must be a "
               "finite integer greater than or equal to 3.")
```

scipy/stats/tests/test_continuous.py
```diff
@@ -914,12 +945,22 @@ class Test(ContinuousDistribution):
...
+    with pytest.raises(ValueError, match=message):
+        Test().lmoment(1, standardized=True)
```
mdhaber: Here and throughout.
```suggestion
        Test().lmoment(1, standardize=True)
```

scipy/stats/tests/test_continuous.py
```diff
@@ -1233,6 +1274,16 @@ def moment(self, order, kind='raw', *, a, b):
...
 
+            def lmoment(self, order, *, a, b):
+                s = np.log(b/a)
+                l1 = (b - a) / s
+                l2 = ((s - 2) * b + (s + 2) * a) / s**2
+                l3 = ((s**2 - 6*s + 12) * b - (s**2 + 6*s + 12) * a) / s ** 3
+                l4 = ((s**3 - 12*s**2 + 60*s - 120) * b
+                      + (s**3 + 12*s**2 + 60*s + 120) * a) / s ** 4
+                ls = {1: l1, 2: l2, 3: l3, 4: l4}
+                return ls.get(int(order), None)
```
mdhaber, line 1282: To verify against definition from [here](https://en.wikipedia.org/wiki/L-moment):

image

```python3
from sympy import *
p, a, b, c, x = symbols('p, a, b, c, x', positive=True, real=True)

Q = solve( (log(x) - log(a))/(log(b)- log(a)) - p, x)[0]
l1_ref = simplify(integrate(Q, (p, 0, 1)))
l2_ref = simplify(integrate(Q*(2*p - 1), (p, 0, 1)))
l3_ref = simplify(integrate(Q*(6*p**2 - 6*p + 1), (p, 0, 1)))
l4_ref = simplify(integrate(Q*(20*p**3 -30*p**2 + 12*p - 1), (p, 0, 1)))

s = log(b/a)
l1 = (b - a)/s
simplify(l1 - l1_ref)

l2 = ((s - 2) * b + (s + 2) * a) / s**2
simplify(l2 - l2_ref)

l3 = ((s**2 - 6*s + 12) * b - (s**2 + 6*s + 12) * a) / s ** 3
simplify(l3 - l3_ref)

l4 = ((s**3 - 12*s**2 + 60*s - 120) * b + (s**3 + 12*s**2 + 60*s + 120) * a) / s ** 4
simplify(l4 - l4_ref)
```

mdhaber (commented):
OK, I think this is in good enough shape for another set of eyes. ~~There is a trivial CI failure to address (due to Accelerate's sloppy arithmetic, presumably), and there are a few comments I've made that I'll correct in a bit.~~ Also, even if there were things I'd change, I've followed existing conventions, so let's save those ideas for a separate refactoring PR.

mdhaber:
@jorenham Thought I'd ping you in case you didn't notice this one. Would be nice to add trimming to both the estimator and this method at some point, but it's easier to work incrementally.

jorenham:
> @jorenham Thought I'd ping you in case you didn't notice this one. Would be nice to add trimming to both the estimator and this method at some point, but it's easier to work incrementally.

Ohh awesome! Somehow I completely missed this, but I'll soon have a look.

scipy/stats/tests/test_continuous.py
jorenham: In case you didn't know already, these are the shifted legendre polynomials. So you could also use `special.sh_legendre(r-1, s)` here for the `r`th L-moment.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3381,6 +3392,77 @@ def logintegrand(x, order, logcenter, **params):
...
+        if lmoment is None and 'order_statistics' in methods:
+            lmoment = self._lmoment_from_order_statistics(order, **params)
+
+        if lmoment is None and 'quadrature_icdf' in methods:
+            lmoment = self._lmoment_integrate_icdf(order, **params)
+
+        if lmoment is not None and self.cache_policy != _NO_CACHE:
+            self._lmoment_cache[order] = lmoment
+
+        return lmoment
```
jorenham, line 3442: It's also possible to use quadrature with the CDF (without the PDF).
From https://jorenham.github.io/Lmo/api/low_level/#lmo.theoretical.l_moment_from_cdf:

image

This is for the generalized trimmed L-moments, so for the L-moments you set s=t=0, and the maths will simplify quite a bit. Note that the shifted Jacobi polynomial there isn't  the `eval_sh_jacobi`, but rather `eval_jacobi(r - 2, 1, 1, 2*u - 1)`.

[implementation](https://github.com/jorenham/Lmo/blob/38215eb525eb920e440c3e1363e223720f1d2ccb/lmo/theoretical/_f_to_lm.py#L216-L261)

jorenham: In case you're interested, in those same docs I also describe how to get the generalized trimmed L-moment from the quantile function:

image

(if it wasn't clear yet; the $x(F)$ is just a funky way of denoting the quantile function $F^{-1}(p)$ or $Q(p)$ or something with $p \mapsto F$ substituted)

[docs](https://jorenham.github.io/Lmo/api/low_level/#lmo.theoretical.l_moment_from_ppf), [code](https://github.com/jorenham/Lmo/blob/38215eb525eb920e440c3e1363e223720f1d2ccb/lmo/theoretical/_f_to_lm.py#L388-L413)

Fun fact: I also figured out how you can go the other way around; i.e. L-moments -> PPF ([docs](https://jorenham.github.io/Lmo/api/low_level/#lmo.theoretical.ppf_from_l_moments)). And it actually works too!

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3381,6 +3392,77 @@ def logintegrand(x, order, logcenter, **params):
...
+
+    def _lmoment_general(self, order, **params):
+        return self.mean() if order == 1 else None
+
+    def _lmoment_from_order_statistics(self, order, **params):
+        k = np.arange(order)
+        k = xpx.atleast_nd(k, ndim=self._ndim + 1).T
+        E = order_statistic(self, r=order-k, n=order).mean()
+        bc = special.binom(order-1, k)
+        return np.sum((-1)**k * bc * E, axis=0) / order
```
jorenham, line 3455: If I remember correctly, these coefficients coincide with the inverse of the lower pascal matrix. Perhaps that could speed things up a bit here.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3381,6 +3392,77 @@ def logintegrand(x, order, logcenter, **params):
...
+
+    def _lmoment_integrate_icdf(self, order, **params):
+        def integrand(p, **params):
+            x = self._icdf_dispatch(p, **params)
+            P = special.eval_sh_legendre(order - 1, p)
+            return x * P
+        return self._quadrature(integrand, limits=(0., 1.), params=params)
```
jorenham, line 3462: Yup, this looks correct to me.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -3686,6 +3768,11 @@ def _logccdf2(self, x, y, *, method):
             "Two argument cdf functions are currently only supported for "
             "continuous distributions.")
 
+    def _lmoment(self, order, *, standardize, method):
+        raise NotImplementedError(
+            "L-moments are currently available only "
+            "for continuous distributions.")
```
jorenham, line 3772: A monte-carlo approach could still work, no?

scipy/stats/tests/test_continuous.py
mdhaber: I think I realized that at some point and forgot later. Thanks.

scipy/stats/_distribution_infrastructure.py
```diff
@@ -5297,6 +5396,10 @@ def _moment_central(self, order):
             out += moment * weight
         return out[()]
 
+    def lmoment(self, order=1, *, standardize=False, method=None):
```
jorenham, line 5397: `special.lmoment` has `standardize=True` by default. Any reason for deviating here?

scipy/stats/_new_distributions.py
jorenham, line 1: Maybe for other (future) distributions this table I compiled could help:
https://jorenham.github.io/Lmo/distributions/#l-stats

It's more elaborate that the one in wikipedia at least, and there are also TL-moments and stuff.

scipy/stats/_new_distributions.py
```diff
@@ -198,6 +204,12 @@ def _moment_central_formula(self, order, **kwargs):
     def _moment_standardized_formula(self, order, **kwargs):
         return self._moment_raw_formula(order, **kwargs)
 
+    def _lmoment_formula(self, order, **kwargs):
+        lscale = 1 / np.sqrt(np.pi)
+        lkurtosis = 30*np.arctan(np.sqrt(2))/np.pi - 9
```
jorenham: Since I've seen this twice now, maybe it's an idea to extract this as a constant? If you care about import time, then you could pre-computing it with wolfram alpha or something.

scipy/stats/_new_distributions.py
```diff
@@ -198,6 +204,12 @@ def _moment_central_formula(self, order, **kwargs):
     def _moment_standardized_formula(self, order, **kwargs):
         return self._moment_raw_formula(order, **kwargs)
 
+    def _lmoment_formula(self, order, **kwargs):
+        lscale = 1 / np.sqrt(np.pi)
```
jorenham: See! If L-moments would have been the *standard* (i.e. normalized to have L-scale 1), the `1/sqrt(pi)` disappears form the normal PDF.  And then I wouldn't have had to cheat at my high school maths exams because the normal PDF was too complicated to remember (true story).

scipy/stats/_new_distributions.py
```diff
@@ -417,6 +433,10 @@ def _moment_central_formula(self, order, ab, **kwargs):
 
     _moment_central_formula.orders = [2]  # type: ignore[attr-defined]
 
+    def _lmoment_formula(self, order, *, a, b, ab, **kwargs):
+        lmoments = {1: 0.5*(a + b), 2: ab / 6, 3: 0, 4: 0}
```
jorenham: Fun fact: The reason that every uniform L-moment $r>2$ is 0, is because the uniform quantile function is a polynomial of order 1; so it disappears after taking the derivative twice :)

jorenham (approved):
Look great! Got nothing to complain; just a few suggestions for improvement, and some random L-moment facts (hope you don't mind).

scipy/stats/_distribution_infrastructure.py
mdhaber: I didn't look into it. Just continuous seemed like a good stopping point for one PR.

jorenham: There's no literature of discrete distribution L-moments that I'm aware of, so yea, I doubt that anyone would care :P

scipy/stats/_distribution_infrastructure.py
mdhaber: I asked about this at the top. I believe I was thinking about parity with the `moment` method, but yeah, probably better to do just what the sample `lmoment` function does? Basically `standardize=True` would be the default, but it would only affect moments 3 and above?

scipy/stats/_new_distributions.py
mdhaber: Will do. *Done by using `super()` in `StandardNormal`.*

scipy/stats/_new_distributions.py
mdhaber: Oh thanks, I'll use that.

scipy/stats/_new_distributions.py
mdhaber: Oh then I'll change this to work for higher orders. *Done by returning zeros for all higher l-moments.*

scipy/stats/_distribution_infrastructure.py
mdhaber: We could write a paper.

jorenham: Sure, why not

scipy/stats/_distribution_infrastructure.py
mdhaber: You mean the terms other than `E`?

jorenham: Yea exactly

scipy/stats/_distribution_infrastructure.py
jorenham: Well, it's a bit of a stupid answer, but because dividing the L-scale by the L-scale is 1 :shrug:

mdhaber: Yeah, I just wanted to confirm we're on the same page about what the behavior would be. Not sure why, since it's nothing new. Never mind : ) *Done.*

scipy/stats/tests/test_continuous.py
mdhaber: Oh, I used `eval_sh_legendre` in the code itself for generic l-moment evaluation. The code above is just verifying the formulas for the l-moments of the log-uniform distribution by symbolically integrating with sympy, so I need to spell them out.

scipy/stats/_distribution_infrastructure.py
mdhaber: Geez. I get the sense you did a lot of these derivations yourself, too? How did you get interested?

> It's also possible to use quadrature with the CDF (without the PDF).

Did you mean PPF, or did you really mean PDF, noticing that if PPF isn't defined we fall back to the definition in terms of the order statistic distribution?

Do you know if one formulation tends to be easier to evaluate than the other? I think for most distributions, if it has the CDF, it's not hard to evaluate the PDF, so just avoiding the PDF isn't too important on its own. But if this formula tends to be more reliable numerically, I'd be happy to review a follow-up PR adding it.

That said, the sheer amount you've done in `lmo` makes it a bit demotivating to enhance SciPy's capabilities in this area : )

scipy/stats/_distribution_infrastructure.py
mdhaber: Maybe I don't understand, but the time to evaluate `(-1)**k * bc / order` is very small compared to `E`. 
*Oh, I didn't remember we have an `invpascal` function. Well, this is a bit easier to verify against references, so I think I'll leave it for now, but yeah, maybe something to study and consider for a follow-up. But we'd want to verify that `invpascal` is actually faster.*

mdhaber: I don't think there's a speed boost here:
```python3
import numpy as np
from scipy import special, linalg

def fun(order):
    k = np.arange(order)
    bc = special.binom(order-1, k)
    return (-1)**k * bc / order

%timeit fun(2)
# 6.01 µs ± 71.1 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
%timeit linalg.invpascal(2)
# 5.42 µs ± 86.4 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

%timeit fun(3)
# 7.82 µs ± 2.29 µs per loop (mean ± std. dev. of 7 runs, 100000 loops each)
%timeit linalg.invpascal(3)
# 12.7 µs ± 2.94 µs per loop (mean ± std. dev. of 7 runs, 100000 loops each)

%timeit fun(4)
# 6.22 µs ± 91.4 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
%timeit linalg.invpascal(4)
# 24.9 µs ± 7.46 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

scipy/stats/_distribution_infrastructure.py
jorenham: > Geez. I get the sense you did a lot of these derivations yourself, too? 

Yea, a couple of papers worth even :sweat_smile:.

> How did you get interested?

I was working in quantitative finance for a while, and we were having issues with outliers and were looking for more robust ways to do statistics with. I somehow ended up on the wikipedia page, and I guess that the obscurity of it all just got me curious initially. And then as I kept digging deeper, I kept finding these fascinating patterns (closed form theoretical L-moments, relationship with orthogonal polynomials, the connection between L-moments and the standard deviation, fractional trimmed L-moments, trimmed L-comoments, a generalized method of TL-moments that can outperform max likelihood, etc. etc.) which I'm pretty sure aren't mentioned in the literature yet. And that feeling of "discovering" something truly new is what got me hooked I guess. But yea, then I side-tracked again and wrote scipy-stubs, even though there's still a bunch of papers I should write.

scipy/stats/_distribution_infrastructure.py
jorenham: Ah ok too bad. Well, it was worth I shot I guess :shrug:

jorenham: Ah I see why now; `linalg.invpascal` really optimized, and uses `special.comb` under the hood: https://github.com/scipy/scipy/blob/0c4cfab8b27d85785d28ec0e4b8d1ce536276c0e/scipy/linalg/_special_matrices.py#L864-L876

jorenham: That's probably why I wrote a custom function for it in Lmo then:
https://github.com/jorenham/Lmo/blob/38215eb525eb920e440c3e1363e223720f1d2ccb/lmo/linalg.py#L75-L179

scipy/stats/tests/test_continuous.py
jorenham: makes sense :ok_hand:

scipy/stats/_stats_py.py
```diff
@@ -10406,6 +10406,11 @@ def lmoment(sample, order=None, *, axis=0, sorted=False, standardize=True):
     --------
     moment
 
+    Notes
+    -----
+    SciPy offers only basic capabilities for working with L-moments. For more advanced
+    features, consider the ``lmo`` package [4]_.
```
jorenham, line 10467: Awesome; thanks!

scipy/stats/_distribution_infrastructure.py
jorenham: But yea, the `standardize` might indeed be a bit confusing :thinking:. The term "standardized" L-moment isn't really a thing either, and usually they're called "L-moment ratios", "L-moment coefficients", or specifically the "L-skewness" and "L-kurtosis". 
So maybe something like a dedicated method (`lratio` or `lmoment_ratio` or something) instead of the `standardized` parameter could help? It could even require `order >=3` if that helps explain things.

But now that I think about it; the L-CV (coefficient of variation), usually denoted as $\tau = \lambda_2 / lambda_1$ (so no subscript), is also a thing (and used quite frequently in fields like hydrology actually). So maybe that's could be dubbed as the "2nd L-moment ratio"? Because then you could simply use `lratio(r) = lmoment(r) / lmoment(max(2, r-1))`, which will even work for `r=1` if you define `lmoment(0) = 1` (Lmo does this). Although documentation wise it'd probably be clearer to just list the  L-ratio's: L-location (r=1), L-CV (r=2 / r=1), L-skewness (r=3 / r=2), L-kurtosis (r=4 / r=2), etc. It might sound artificial like this, but if you consider the conventional moment analogues such as the "excess kurtosis" with that -3 hack, then I'd say this is still a lot cleaner :P.
The L-CV iss pretty well-known in the literature BTW. For example, from [Hosking (1990)](https://doi.org/10.1111/j.2517-6161.1990.tb01775.x) (i.e. the paper that coined the term "L-moment"):

> It is often convenient to standardize the higher moments $\lambda_r$ , $r \ge 3$, so that they are independent of the units of measurement of $X$. Define, therefore, the L-moment ratios of $X$ to be the quantities
>
> $$
> \tau_r \equiv \lambda_r/\lambda_2, \quad r=3, 4, \ldots
> $$
>
> It is also possible to define a function of L-moments which is analogous to the
coefficient of variation (CV): this is the *L-CV*, $\tau \equiv \lambda_2/\lambda_1$.

---

TLDR; in order to avoid the somewhat confusing `standardized` parameter; how about a dedicated `lratio` or `lmoment_ratio` method that returns `lmoment(1) if r==1`, `lmoment(2)/lmoment(1) if r==2` (L-CV), and `lmoment(r)/lmoment(2) if r>2`? It would require some documentation, but it's in line with the literature, where the L-CV is often listed as one of the L-moment ratio's.

scipy/stats/_distribution_infrastructure.py
jorenham: Since they're only a thing for continuous univariate distribution, what do you think about move these lmoment methods from `UnivariateDistribution` over to `ContinuousDistribution`? If we ever want to add discrete L-moments, then we can always move it back up to `UnivariateDistribution`.

For scipy-stubs that would make life a bit easier. And I can imagine that it could also save a couple of lines of codes here, e.g. this method wouldn't be needed then.

scipy/stats/_distribution_infrastructure.py
mdhaber: K. `binom` isn't too slow, especially compared to evaluating `E`, so I'll let it be.

scipy/stats/_distribution_infrastructure.py
mdhaber: Documentation for all methods is defined at the superclass level, currently. This would be the first to break that pattern. So if there's a chance of implementing it for `DiscreteDistribution`, I'd like to keep it that way.

scipy/stats/_distribution_infrastructure.py
mdhaber: We discussed this before, though, [here](https://github.com/scipy/scipy/pull/19475#discussion_r1725558899). For better or for worse, it made it in there. It's clearly documented - if a user is looking for l-moment ratios, they will easily find that they can get them with `standardize=True`. Let's stay self-consistent.

j-bowhay:
@mdhaber is there anything else you wanted to do here? Given that Joren has approved and he seems to be an expert on this topic I'd be willing to look at this from a just code/test coverage perspective

mdhaber:
Nope, I'm happy with it. Thanks!

j-bowhay (approved):

PR #23769, opened 2025-10-13
DOC: ``special._specfun.eulerb``: fix cython docstring

jorenham:
Because as far as I'm aware, Euler != Bernoulli (the proof is left as an exercise for the reader).


j-bowhay (approved):
Looks like a previous copy paste error!

PR #23770, opened 2025-10-13
MAINT: clean up some outdated CPython C API idioms

rgommers:
Motivated by the discussion in scipy#19878. With these changes, all Cython code can be built against the Limited API. That just requires adding `-DPy_LIMITED_API=0x030C0000` as a flag to `cython_c_args` in scipy/meson.build

Note that using the Limited API isn't all that useful yet, since it's blocked by lack of support in dependencies other than Cython. But it's a start, and makes it easier to do benchmarking of the functionality implemented in Cython.


tylerjereddy (approved):
When I make the `meson` config change Ralf suggests, I get a failure to build locally on latest `main` branch, and a successful local build if I pull the same build flag into this feature branch.

Considering how small the patch is, and only the one unrelated CI failure, I'm happy to put this in as a tiny step toward the ABI portability business, thanks.

PR #23771, opened 2025-10-14
BUG: odr: prevent segfault with explicit model and job=1 (#23763)

carlini:
#### Reference issue
Closes #23763

ODR now raises an OdrError if job is 1 mod 10 which configures implicit mode, even if model.implicit isn't set to be true. Also adds a corresponding test case.

### DIFF
--- a/scipy/odr/_odrpack.py
+++ b/scipy/odr/_odrpack.py
@@ -799,6 +799,10 @@ class ODR:
             y_s = list(self.data.y.shape)
             if self.model.implicit:
                 raise OdrError("an implicit model cannot use response data")
+            if self.job is not None and (self.job % 10) == 1:
+                raise OdrError("the self.job parameter was set to 1 mod 10;"
+                               " this configures an implicit model, and"
+                               " an implicit model cannot use response data")
         else:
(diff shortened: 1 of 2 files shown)


scipy/odr/_odrpack.py
```diff
@@ -799,6 +799,10 @@ def _check(self):
             y_s = list(self.data.y.shape)
             if self.model.implicit:
                 raise OdrError("an implicit model cannot use response data")
+            if self.job is not None and (self.job % 10) == 1:
+                raise OdrError("the self.job parameter was set to 1 mod 10;"
+                               " this configures an implicit model, and"
+                               " an implicit model cannot use response data")
```
rkern: Here's a more user-facing error message.
```suggestion
                raise OdrError("job parameter requests an implicit model,"
                               " but an explicit model was passed")
```

rkern (approved):
I've tightened up the error message. Other than that, LGTM! Thank you!

rkern:
I believe the test failures and doc build failures are unrelated and being seen in other PRs, so in it goes!

Thank you very much!

PR #23772, opened 2025-10-14
ENH: stats.permutation_test: add array API support

mdhaber:
#### Reference issue
Toward gh-20544

Adds array API support to `stats.permutation_test`.

#### Additional information
Similar to `bootstrap`, `rng` is still a NumPy random number generator, and other permutation indices are generated with NumPy or pure Python. (Would benefit from array API `random`, `permuted`, and `permutation`). The rest of the function has been translated to the array API.

There are a lot of tests that can't be converted because the statistics have not been converted to the array API. I think that's OK - they are a bit overkill. There are already tests that confirm that the behavior is correct.


scipy/stats/_resampling.py
```diff
@@ -1428,7 +1431,7 @@ def batched_perm_generator():
 
 
 def _calculate_null_both(data, statistic, n_permutations, batch,
-                         rng=None):
+                         rng=None, *, xp):
     """
     Calculate null distribution for independent sample tests.
     """
```
mdhaber, line 1448: Lint fix is to remove the next line.

mdhaber:
CI failure with 0e074ce was just CircleCI/Accelerate (all PRs) and a lint issue, fixed in the last commit.

scipy/stats/tests/test_resampling.py
```diff
@@ -1683,20 +1693,21 @@ def test_against_binomtest(self, alternative):
...
         k, n, p = statistic(x), 10, 0.5
         expected = stats.binomtest(k, n, p, alternative=alternative)
 
-        res = stats.permutation_test((x,), statistic, vectorized=True,
-                                     permutation_type='samples',
-                                     n_resamples=np.inf, rng=self.rng,
-                                     alternative=alternative)
-        assert_allclose(res.pvalue, expected.pvalue, rtol=self.rtol)
+        res = stats.permutation_test((xp.asarray(x, dtype=xp.float64),), statistic,
+                                     permutation_type='samples', n_resamples=np.inf,
```
tylerjereddy: looks like this passes with `np.inf` -> `xp.inf`; is that change just pointless or?

scipy/stats/tests/test_resampling.py
```diff
@@ -1589,6 +1596,8 @@ def statistic1d(x, y):
     @pytest.mark.xslow()
     @pytest.mark.parametrize('axis', (-1, 2))
     def test_vectorized_nsamp_ptype_both(self, axis):
+        # statistic only available for NumPy
```
tylerjereddy, line 1631: Is there no value/not worth the hassle to add the explicit array API skip/np-only decorators here and elsewhere so that the lack of capability is "programmatically" recorded in the testsuite outcome?

scipy/stats/tests/test_resampling.py
mdhaber: Oops. I just missed the `np`.
I wouldn't re-run CI for it, but sure. These are all true:
```python3
import math
import numpy as np
import cupy as cp
import jax.numpy as jnp
import torch
import dask.array as da

print(np.inf == math.inf and isinstance(np.inf, float))
print(cp.inf == math.inf and isinstance(cp.inf, float))
print(jnp.inf == math.inf and isinstance(jnp.inf, float))
print(torch.inf == math.inf and isinstance(torch.inf, float))
print(da.inf == math.inf  and isinstance(da.inf, float))
```

scipy/stats/tests/test_resampling.py
mdhaber: I didn't think we were doing that anymore. I don't see value in it, personally. But here's another option:

I added these tests because they gave me a warm and fuzzy feeling, showing that `permutation_test` replicates the  results of basically all the other "exact" tests. But really, they are mostly a vectory lap; they are not all needed as unit tests of `permutation_test`. Rather, they may be more appropriate as unit tests of the *other* function, with `permutation_test` as the reference. I could move them to the *other* test classes; e.g., this one moves to `kruskal`. Then they don't need to be translated to array-API compatible until the *other* function is array-API compatible.

I can also leave them here and add the skips, if you prefer. Note that pretty much [all of these function are getting array API support soon](https://github.com/scipy/scipy/pulls?q=is%3Aopen+is%3Apr+author%3Amdhaber+label%3A%22array+types%22), so we can also just leave them alone for now and come back to translate to array API shortly.

tylerjereddy: Actually, now that I think about it, I don't think I understand why we would need skips in these cases--I thought this was pretty much the exact scenario @steppi had in mind with gh-23765, `_xp_copy_to_numpy`, and isolating `xp` usage to *only* what is directly under test, rather than skipping. Am I misunderstanding that Albert?

>  Note that pretty much all of these function are getting array API support soon,

I'm also not sure this matters--per gh-23765, even if the reference/expected-value/input-value-calculating functions did have array API support, we'd still use NumPy to reduce the likelihood of developers introducing skips for the function under test just because of undesirable behavior for other functions not under test.

I don't know if you saw the discussion/content over there, that's pretty recent. I'm not trying to make this too complicated, but it may be good to clarify on that, since this approach was coded directly into the docs for array API testing in that PR. I believe that would make this fairly straightforward to just convert right now too.

I only got about half way through my review earlier today before getting distracted, sorry...

mdhaber: No, I didn't see that. I guess I like the _option_ to unblock tests of function _A_ by using NumPy on function _B_ . I don't really like a _requirement_ that every existing test of _A_ must be run with all backends, or the idea that we *can't* use `xp` with function _B_ in a test of function _A_ (especially in these cases in which function _A_ accepts function _B_ as an argument and uses it during execution; _B_ is not simply being used to generate reference values).

This process of adding array API support shouldn't be allowed to be sloppy, but it needs to be allowed to be iterative. I think maintainers should be allowed to make decisions about whether it's safe to temporarily skip some tests for non-NumPy backends and come back later - before alternative backends are actually supported (not just experimental) - to reassess what we document as being supported.

I had ChatGPT write a [script](https://chatgpt.com/s/t_68fb0d81bde4819195ab12741a6e5410) to find methods of classes marked with the `make_xp_test_case` decorator that neither accept an `xp` argument or have a `pytest.mark.skip_if(SCIPY_ARRAY_API....)` decorator. In main, it finds the following. 

<details>

```
❌ /Users/matthaberland/Desktop/scipy/scipy/cluster/tests/test_vq.py:
  - TestVq.test__vq_sametype() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestVq.test__vq_invalid_type() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestKMeans.krand_lock() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
❌ /Users/matthaberland/Desktop/scipy/scipy/cluster/tests/test_hierarchy.py:
  - TestIsIsomorphic.test_array_like() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestIsIsomorphic.is_isomorphic_randperm() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestDendrogram.dendrogram_lock() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
❌ /Users/matthaberland/Desktop/scipy/scipy/ndimage/tests/test_filters.py:
  - TestNdimageFilters.test_rank16() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
❌ /Users/matthaberland/Desktop/scipy/scipy/linalg/tests/test_special_matrices.py:
  - TestBlockDiag.test_no_args() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
❌ /Users/matthaberland/Desktop/scipy/scipy/linalg/tests/test_procrustes.py:
  - TestOrthogonalProcrustes.test_orthogonal_procrustes_array_conversion() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
❌ /Users/matthaberland/Desktop/scipy/scipy/optimize/tests/test_chandrupatla.py:
  - TestChandrupatlaMinimize.f() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestFindRoot.f() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
❌ /Users/matthaberland/Desktop/scipy/scipy/optimize/tests/test_bracket.py:
  - TestBracketRoot.f() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestBracketRoot.test_bug_fixes() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestBracketMinimum.init_f() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestBracketMinimum.get_kwargs() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
❌ /Users/matthaberland/Desktop/scipy/scipy/integrate/tests/test_quadpack.py:
  - TestCtypesQuad.setup_method() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestCtypesQuad.test_typical() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestCtypesQuad.test_ctypes_sine() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestCtypesQuad.test_ctypes_variants() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMultivariateCtypesQuad.setup_method() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMultivariateCtypesQuad.test_typical() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMultivariateCtypesQuad.test_indefinite() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMultivariateCtypesQuad.test_threadsafety() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestQuad.test_typical() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestQuad.test_indefinite() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestQuad.test_singular() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestQuad.test_sine_weighted_finite() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestQuad.test_sine_weighted_infinite() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestQuad.test_cosine_weighted_infinite() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestQuad.test_algebraic_log_weight() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestQuad.test_cauchypv_weight() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestQuad.test_b_less_than_a() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestQuad.test_b_less_than_a_2() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestQuad.test_b_less_than_a_3() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestQuad.test_b_less_than_a_full_output() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestQuad.test_b_equals_a() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestQuad.test_complex() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestDblquad.test_double_integral() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestDblquad.test_double_integral2() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestDblquad.test_double_integral3() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestDblquad.test_double_integral_improper() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestTplquad.test_triple_integral() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestTplquad.test_triple_integral_improper() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestNQuad.test_fixed_limits() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestNQuad.test_variable_limits() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestNQuad.test_square_separate_ranges_and_opts() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestNQuad.test_square_aliased_ranges_and_opts() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestNQuad.test_square_separate_fn_ranges_and_opts() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestNQuad.test_square_aliased_fn_ranges_and_opts() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestNQuad.test_matching_quad() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestNQuad.test_matching_dblquad() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestNQuad.test_matching_tplquad() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestNQuad.test_dict_as_opts() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
❌ /Users/matthaberland/Desktop/scipy/scipy/integrate/tests/test__quad_vec.py:
  - TestQuadVec.test_quad_vec_simple() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestQuadVec.test_quad_vec_simple_inf() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestQuadVec.test_quad_vec_args() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestQuadVec.test_quad_vec_pool() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestQuadVec.test_quad_vec_pool_args() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestQuadVec.test_num_eval() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestQuadVec.test_info() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestQuadVec.test_nan_inf() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestQuadVec.test_points() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
❌ /Users/matthaberland/Desktop/scipy/scipy/integrate/tests/test_tanhsinh.py:
  - TestTanhSinh.f1() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestTanhSinh.f2() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestTanhSinh.f3() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestTanhSinh.f4() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestTanhSinh.f5() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestTanhSinh.f6() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestTanhSinh.f7() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestTanhSinh.f8() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestTanhSinh.f9() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestTanhSinh.f10() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestTanhSinh.f11() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestTanhSinh.f12() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestTanhSinh.f13() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestTanhSinh.f14() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestTanhSinh.f15() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestNSum.f1() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestNSum.f2() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestNSum.f3() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
❌ /Users/matthaberland/Desktop/scipy/scipy/integrate/tests/test_cubature.py:
  - TestCubature.test_pass_array_like_not_array() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
❌ /Users/matthaberland/Desktop/scipy/scipy/integrate/tests/test_quadrature.py:
  - TestFixedQuad.test_scalar() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestFixedQuad.test_vector() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestNewtonCotes.test_newton_cotes() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestNewtonCotes.test_newton_cotes2() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestSimpson.test_simpson() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestSimpson.test_simpson_2d_integer_no_x() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestTrapezoid.test_array_like() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestQMCQuad.test_flexible_input() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestLebedev.test_input_validation() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestLebedev.test_quadrature() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestLebedev.test_properties() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
❌ /Users/matthaberland/Desktop/scipy/scipy/constants/tests/test_constants.py:
  - TestConvertTemperature.test_convert_temperature_array_like() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestConvertTemperature.test_convert_temperature_errors() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestLambdaToNu.test_lambda_to_nu_array_like() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestNuToLambda.test_nu_to_lambda_array_like() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
❌ /Users/matthaberland/Desktop/scipy/scipy/special/tests/test_logsumexp.py:
  - TestLogSumExp.test_xp_invalid_input() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestLogSumExp.test_array_like() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestSoftmax.test_softmax_scalar() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestSoftmax.test_softmax_array_like() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestLogSoftmax.test_log_softmax_scalar() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestLogSoftmax.test_log_softmax_array_like() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
❌ /Users/matthaberland/Desktop/scipy/scipy/differentiate/tests/test_differentiate.py:
  - TestDerivative.f() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestDerivative.test_accuracy() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestJacobian.df1() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestJacobian.df2() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestJacobian.df3() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestJacobian.df4() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestJacobian.df5() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestJacobian.rosen() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
❌ /Users/matthaberland/Desktop/scipy/scipy/interpolate/tests/test_bsplines.py:
  - TestInterp.test_non_int_order() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestInterp.test_incompatible_x_y() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestInterp.test_broken_x() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestInterp.test_periodic_points_exception() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestInterp.test_periodic_knots_exception() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestInterp.test_periodic_splev() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestInterp.test_periodic_cubic() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestInterp.test_periodic_full_matrix() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestInterp.test_cubic_deriv_unstable() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestInterp.test_deriv_spec() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestInterp.test_list_input() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestInterp.test_shapes() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestInterp.test_woodbury() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestLSQ.test_lstsq() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestLSQ.test_int_xy() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestLSQ.test_f32_xy() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestLSQ.test_sliced_input() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestLSQ.test_checkfinite() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestLSQ.test_read_only() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestLSQ.test_qr_vs_norm_eq() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestLSQ.test_duplicates() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestSmoothingSpline.test_invalid_input() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestSmoothingSpline.test_compare_with_GCVSPL() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestGenerateKnots.test_split_add_knot() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestGenerateKnots.test_s0_nest() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestGenerateKnots.test_list_input() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestGenerateKnots.test_weights() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestGenerateKnots.test_vs_splrep() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestGenerateKnots.test_s_too_small() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestGenerateKnots.test_zero_weights() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMakeSplrep.test_fitpack_F() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMakeSplrep.test_fitpack_F_with_weights() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMakeSplrep.test_disc_matrix() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMakeSplrep.test_with_knots() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMakeSplrep.test_s_too_small() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMakeSplrep.test_issue_22704() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMakeSplrepPeriodic.test_s0_vs_not() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMakeSplrepPeriodic.test_periodic_with_non_periodic_data() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMakeSplrepPeriodic.test_make_splrep_periodic_m_eq_2_k_eq_1() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMakeSplrepPeriodic.test_fperiodic_basic_fit() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMakeSplprep.test_simple_vs_splprep() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMakeSplprep.test_array_not_list() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMakeSplprep.test_1D() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMakeSplprepPeriodic.test_simple_vs_splprep() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMakeSplprepPeriodic.test_array_not_list() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMakeSplprepPeriodic.test_1D() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
❌ /Users/matthaberland/Desktop/scipy/scipy/interpolate/tests/test_polyint.py:
  - TestCubicSpline.check_correctness() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestCubicSpline.check_all_bc() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestCubicSpline.test_general() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestCubicSpline.test_periodic() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestCubicSpline.test_second_derivative_continuity_gh_11758() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestCubicSpline.test_three_points() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestCubicSpline.test_periodic_three_points_multidim() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestCubicSpline.test_dtypes() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestCubicSpline.test_small_dx() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestCubicSpline.test_incorrect_inputs() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
❌ /Users/matthaberland/Desktop/scipy/scipy/interpolate/tests/test_interpolate.py:
  - TestAkima1DInterpolator.test_eval_3d() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestAkima1DInterpolator.test_linear_interpolant_edge_case_3d() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestAkima1DInterpolator.test_extend() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestAkima1DInterpolator.test_mod_invalid_method() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestAkima1DInterpolator.test_extrapolate_attr() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestAkima1DInterpolator.test_no_overflow() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPolyCommon.test_ctor_c() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPolyCommon.test_shape() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPolyCommon.test_complex_coef() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPoly.test_read_only() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPoly.test_descending() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPoly.test_construct_fast() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPoly.test_vs_alternative_implementations() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPoly.test_from_spline() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPoly.test_derivative_eval() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPoly.test_derivative() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPoly.test_antiderivative_of_constant() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPoly.test_antiderivative_regression_4355() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPoly.test_antiderivative_vs_derivative() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPoly.test_antiderivative_vs_spline() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPoly.test_antiderivative_continuity() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPoly.test_integrate() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPoly.test_integrate_readonly() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPoly.test_integrate_periodic() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPoly.test_roots() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPoly.test_roots_idzero() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPoly.test_roots_all_zero() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPoly.test_roots_repeated() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPoly.test_roots_discont() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPoly.test_roots_random() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPoly.test_roots_croots() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPPoly.test_extrapolate_attr() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestBPoly.test_descending() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestBPoly.test_multi_shape() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestBPoly.test_extrapolate_attr() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestBPolyCalculus.test_deriv_inplace() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestBPolyCalculus.test_der_antider() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestBPolyCalculus.test_antider_ppoly() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestBPolyCalculus.test_antider_continuous() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestBPolyCalculus.test_integrate_extrap() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPolyConversions.test_bp_from_pp_random() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPolyConversions.test_broken_conversions() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
❌ /Users/matthaberland/Desktop/scipy/scipy/stats/tests/test_stats.py:
  - TestPearsonr.test_pearsonr_result_attributes() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPearsonr.test_input_validation() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPearsonr.test_resampling_pvalue() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPearsonr.test_bootstrap_ci() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPearsonr.test_axis01() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPearsonr.test_axis_None() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMode.test_scalar() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMode.test_keepdims_nan_omit() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMode.test_gh16955() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMode.test_gh9955() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMode.test_raise_non_numeric_gh18254() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestGZscore.test_gzscore_masked_array() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMoments.test_nan_policy() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMoments.test_moment_accuracy() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestSkew.stat_fun() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestSkew.test_nan_policy() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestSkew.test_skewness_scalar() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestKurtosis.stat_fun() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestKurtosis.test_kurtosis_nan_policy() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestKurtosis.test_kurtosis_array_scalar() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestKurtosis.test_kurtosis_propagate_nan() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestTTestInd.test_ttest_ind_nan_2nd_arg() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestDescribe.describe_nan_policy_omit_test() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestDescribe.test_describe_result_attributes() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPMean.pmean_reference() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPMean.wpmean_reference() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestPMean.test_infinite_p_gh23111() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestAlexanderGovern.test_compare_dtypes() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestLMoment.test_dtype_iv_non_numeric() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
❌ /Users/matthaberland/Desktop/scipy/scipy/stats/tests/test_correlation.py:
  - TestChatterjeeXi.test_permutation_asymptotic() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
❌ /Users/matthaberland/Desktop/scipy/scipy/stats/tests/test_rank.py:
  - TestRankData.desired_dtype() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestRankData.test_list() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestRankData.test_rankdata_object_string() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestRankData.test_nan_policy_omit_3d() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestRankData.test_nan_policy_2d_axis_none() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestRankData.test_nan_policy_raise() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestRankData.test_nan_policy_propagate() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
❌ /Users/matthaberland/Desktop/scipy/scipy/stats/tests/test_resampling.py:
  - TestBootstrap.test_vectorize_statistic() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestBootstrap.test_vector_valued_statistic_gh17715() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMonteCarloHypothesisTest.test_against_ks_1samp() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMonteCarloHypothesisTest.test_against_normality_tests() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMonteCarloHypothesisTest.test_against_normaltest() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMonteCarloHypothesisTest.test_against_cramervonmises() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMonteCarloHypothesisTest.test_against_anderson() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMonteCarloHypothesisTest.test_p_never_zero() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMonteCarloHypothesisTest.test_against_ttest_ind() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMonteCarloHypothesisTest.test_against_f_oneway() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))
  - TestMonteCarloHypothesisTest.test_finite_precision_statistic() missing 'xp' argument (no skipif(SCIPY_ARRAY_API))

```
</details>

I haven't checked exhaustively, but it seems like it's working pretty well.

Clearly there are some other tests that need to be addressed. Adding a few more here wouldn't make much of a difference, and I'd still say it's a win if the discussion has prompted us to find all these cases when it's important to do so.

Maybe tomorrow I'll realize it's not so bad to add the tests, but a.t.m. I'd prefer to come back and adjust these when the function *B*s get array API support (if it's not a blocker for you. If it is, I'll change it as you see fit - whether that means adding a skip marker or trying to use `_xp_copy_to_numpy`.)

steppi: > Actually, now that I think about it, I don't think I understand why we would need skips in these cases--I thought this was pretty much the exact scenario @steppi had in mind with [gh-23765](https://github.com/scipy/scipy/pull/23765), `_xp_copy_to_numpy`, and isolating `xp` usage to _only_ what is directly under test, rather than skipping. Am I misunderstanding that Albert?

Right @tylerjereddy , I'd say this is the exact scenario. If we wanted these to be unit tests for `permutation_test`, then the function's calculating the test statistics could simulate use of the alternative backend by copying to numpy, doing the calculation,  and then copying back to `xp`.







> No, I didn't see that. I guess I like the _option_ to unblock tests of function _A_ by using NumPy on function _B_ . I don't really like a _requirement_ that every existing test of _A_ must be run with all backends, or the idea that we _can't_ use `xp` with function _B_ in a test of function _A_ (especially in these cases in which function _A_ accepts function _B_ as an argument and uses it during execution; _B_ is not simply being used to generate reference values).

Yeah, fortunately I saw reason on this front and ended up softening the recommendations I added to the docs considerably.

From https://web.archive.org/web/20251024180151/https://scipy.github.io/devdocs/dev/api-dev/array_api.html#testing-practice

*It’s important that for any supported function f, there exist tests using the xp fixture that restrict use of alternative backends to only the function f being tested. Other functions evaluated within a test, for the purpose of producing reference values, inputs, round-trip calculations, etc. should instead use the NumPy backend. This helps ensure that any failures that occur on a backend actually relate to the function of interest, and avoids the need to skip backends due to lack of support for functions other than f. Property based integration tests which check that some invariant holds using the same alternative backend across different functions can also have value, giving a window into the general health of backend support for a module, but in order to ensure the test suite actually reflects the state of backend support for each function, it’s vital to have tests which isolate use of the alternative backend only to the function being tested.*

So the recommendation is that there need to be tests which isolate use of the alternative backend only to the function of interest, but not that all tests are required to do this. This means that when testing a function $f$ which takes a function $g$ as argument, it's valuable to have tests which simulate having $g$ on the alternative backend by copying to numpy, computing, and then copying back to `xp`. I'd see this as a proper unit test for $f$. I think tests which use the alternative backend across functions are valuable too, and better represent how things would be used in practice. These are more like integration tests, and I'm thinking there should be a convention for naming and commenting them which clearly signals intent.

> I didn't think we were doing that anymore. I don't see value in it, personally. But here's another option:

I think current practice is to use `make_xp_test_case` (or `make_xp_pytest_param`) for skips or xfails which are specific to a particular function, and to use the standalone `skip_xp_backends` and `xfail_xp_backends`  for skips which are needed that are specific to a particular test, e.g. tests for cases not supported on a particular backend (some cupy functions have a limit on the dimensionality of supported arrays for instance), tests which hit particular bugs in functions which for the most part do work correctly. 

> Maybe tomorrow I'll realize it's not so bad to add the tests, but a.t.m. I'd prefer to come back and adjust these when the function _B_s get array API support (if it's not a blocker for you. If it is, I'll change it as you see fit - whether that means adding a skip marker or trying to use `_xp_copy_to_numpy`.)

I think it's fine to proceed in an incremental fashion on this if you prefer. I'll leave it to your judgment which tests work better as unit tests which isolate the alternative backend to only `permutation_test` and which work better as integration tests.

I haven't taken a close look at this PR yet but will do it in the next few days if @tylerjereddy doesn't merge first.

scipy/stats/tests/test_resampling.py
```diff
@@ -1683,20 +1693,21 @@ def test_against_binomtest(self, alternative):
...
 
         k, n, p = statistic(x), 10, 0.5
         expected = stats.binomtest(k, n, p, alternative=alternative)
 
-        res = stats.permutation_test((x,), statistic, vectorized=True,
-                                     permutation_type='samples',
-                                     n_resamples=np.inf, rng=self.rng,
-                                     alternative=alternative)
-        assert_allclose(res.pvalue, expected.pvalue, rtol=self.rtol)
+        res = stats.permutation_test((xp.asarray(x, dtype=xp.float64),), statistic,
```
tylerjereddy, line 1723: what happened to the `vectorized=True` argument? it looks like the test still passes with array API active if I leave that in; even if it has no effect, I wonder if we really want to be making simultaneous adjustment on top of the array API test mutations

mdhaber: > I wonder if we really want to be making simultaneous adjustment on top of the array API test mutations

Are there really so many unnecessary adjustments?

I don't remember exactly why `vectorized` disappeared. It could have been inadvertent because it didn't fit on the same line anymore. It could have been intentional for the same reason, with the understanding that `vectorized` is no longer necessary (for several versions) because `permutation_test` inspects `statistic` looking for an `axis` argument.

Fortunately, it doesn't matter much. We can easily add it back.

scipy/stats/_resampling.py
```diff
@@ -1594,28 +1604,35 @@ def _permutation_test_iv(data, statistic, permutation_type, vectorized,
...
     except TypeError:
         raise TypeError(message)
 
-    data = _broadcast_arrays(data, axis)
+    xp = array_namespace(*data)
+
+    if not vectorized:
+        if not is_numpy(xp):
+            message = (f"When using array library {xp.__name__}, `func` must be "
+                       "be vectorized and accept argument `axis`.")
```
tylerjereddy: there's a repetition here with "be be" that I noticed while probing the usage of `vectorized` in testing above

mdhaber:
```suggestion
                       "vectorized and accept argument `axis`.")
```

scipy/stats/tests/test_resampling.py
```diff
@@ -1905,21 +1925,23 @@ def test_batch_generator(self, iterable, batch, expected):
...
+        x = xp.asarray([1., 2., 4., 3.], dtype=xp.float64)
+        y = xp.asarray([2., 4., 6., 8.], dtype=xp.float64)
 
-        def statistic(x, y):
-            return stats.pearsonr(x, y)[0]
+        def statistic(x, y, axis):
+            return stats.pearsonr(x, y, axis=axis)[0]
 
-        res = stats.permutation_test((x, y), statistic, vectorized=False,
+        res = stats.permutation_test((x, y), statistic,
```
tylerjereddy, line 1974: So here `vectorized=False` is removed, and there *are* array API test failures if that argument is preserved. 

Are we not worried about losing coverage of the original `vectorized=False` scenario for NumPy?

scipy/stats/tests/test_resampling.py
tylerjereddy: There's another case where it was removed below and it seemed to matter--it does add to reviewer overhead, especially without domain knowledge.

scipy/stats/_resampling.py
```diff
@@ -1465,38 +1465,42 @@ def _calculate_null_both(data, statistic, n_permutations, batch,
...
+        data_batch = _get_from_last_axis(data, indices, xp=xp)
 
         # Move the permutation axis to the front: we'll concatenate a list
         # of batched statistic values along this zeroth axis to form the
         # null distribution.
-        data_batch = np.moveaxis(data_batch, -2, 0)
-        data_batch = np.split(data_batch, n_obs_ic[:-1], axis=-1)
+        data_batch = xp.moveaxis(data_batch, -2, 0)
+        # data_batch = np.split(data_batch, n_obs_ic[:-1], axis=-1)
+        data_batch = [data_batch[..., i:j] for i, j in zip(n_obs_ic[:-1], n_obs_ic[1:])]
```
tylerjereddy, line 1498: Just asking -- no concerns of performance hits for the NumPy-native codepath?

scipy/stats/_resampling.py
```diff
@@ -1561,24 +1566,29 @@ def _calculate_null_samples(data, statistic, n_permutations, batch,
...
+        # array and list?
+        data = xp.stack(data, axis=0)
+        data = xp_swapaxes(data, 0, -1, xp=xp)
         if n_samples == 1:
-            data = data[0:1]
+            data = data[0:1, ...]
+        data = [data[i, ...] for i in range(data.shape[0])]
         return statistic(*data, axis=axis)
 
+    data = [data[i, ...] for i in range(data.shape[0])]
```
tylerjereddy, line 1595: lots of manual loops -- "ok" for performance?

scipy/stats/_resampling.py
```diff
@@ -1594,28 +1604,35 @@ def _permutation_test_iv(data, statistic, permutation_type, vectorized,
     if vectorized is None:
         vectorized = 'axis' in inspect.signature(statistic).parameters
 
-    if not vectorized:
-        statistic = _vectorize_statistic(statistic)
```
tylerjereddy, line 1610: maybe this is a cleanup that is obvious for `stats` folks I guess--wasn't obvious to me re: array API

scipy/stats/tests/test_resampling.py
mdhaber: I know. I try to keep it straightforward. Looks like I messed up this time.

Does adding at back address it?

scipy/stats/_resampling.py
tylerjereddy: actually, it is just getting migrated down below I guess..

tylerjereddy:
Ok, I put a few more comments. Apart from those I think we're "ok" given Albert's feedback. I don't think we have a clear division with test naming/convention on what is "functional" vs. "unit" for the np-coercion business, but seems like Albert is ok with taking it progressively on that.

scipy/stats/tests/test_resampling.py
tylerjereddy: I think so, for the other case below we can't unless we parametrize NumPy vs. array API I think, but maybe you have others reasons there.

scipy/stats/tests/test_resampling.py
mdhaber: > So here `vectorized=False` is removed,

Removing it was intentional. I added the `axis` argument to the `statistic` because `pearsonr` now *does* support `axis` and because `vectorized=False` isn't supported by backends other than NumPy. We either remove it or lose this test.

>  Are we not worried about losing coverage of the original `vectorized=False` scenario for NumPy?

No, there is:
https://github.com/scipy/scipy/blob/19b2da6d7a0b29cfbe4bd4b8f0e8cdfb1edc9841/scipy/stats/tests/test_resampling.py#L1977

Usually I do a self-review to explain stuff like this; looks like I didn't in this one.

scipy/stats/_resampling.py
mdhaber: Yes, moving it was motivated by the translation. The `vectorized` validation needs to know `xp`, and I had to validate `data` before calling `array_namespace` - otherwise, I think input validation tests that look for the message "`data` must be a tuple..." were failing because `array_namespace` would raise something else.

scipy/stats/_resampling.py
mdhaber: It is probably faster because it avoids the overhead NumPy functions tend to have.

*Indeed, it seems to be:*
```python3
import numpy as np
rng = np.random.default_rng(2438925893453)
data_batch = rng.random((9999, 300))

n_obs_ic = [100, 200, 300]
ref = np.split(data_batch, n_obs_ic[:-1], axis=-1)
# 5.09 μs ± 107 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

n_obs_ic = [0, 100, 200, 300]
res = [data_batch[..., i:j] for i, j in zip(n_obs_ic[:-1], n_obs_ic[1:])]
# 1.16 μs ± 10.9 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

scipy/stats/tests/test_resampling.py
mdhaber: I do; see https://github.com/scipy/scipy/pull/23772#discussion_r2476443908.

scipy/stats/_resampling.py
mdhaber: I think this is the direct translation of what it was doing before. NumPy arrays were iterable so they could do things like:
```
data = rng.random((1000, 3))
def f(*data):
    return data
ref = f(*data)
```

Directly. So it's going to break up the array up into pieces like this one way or the other. Now, we're just explicilty breaking it up before it gets passed into `f`.

It turns out that it is *is* a little slower to break it up explicitly rather than having it happen implicitly like in the code above. But that is pretty unexpected. For instance, I would not have predicted that one could create a tuple out of a NumPy array by calling:
```python3
ref = f(*data)
```
faster than
```python3
res = tuple(data[i, ...] for i in range(data.shape[0]))
```

Regardless, this should still be quite small overhead compared to computing a statistic 9999 times on this amount of data.

mdhaber:
> but seems like Albert is ok with taking it progressively 

Hopefully. The tests that I've left NumPy-only for now involve:
- `ks_2samp`
- `ansari`
- `mannwhitneyu`
- `cramervonmises_2samp`
- `kruskal`
- `wilcoxon`
- `kendalltau`
- `page_trend_test`

I have PRs toward adding array API support for the first five of those already, and I hope to have one for `wilcoxon` soon. We can make those run on the other backends as soon as those PRs get merged. 

Array API support is experimental now, and presumably these will get merged before array API support is officially supported. Unless array API support becomes official in 1.17, I don't think there is too much risk leaving these as NumPy only, especially since it is very unlikely that any one of these will detect a problem in `permutation_test` that the existing tests don't.

scipy/stats/tests/test_resampling.py
mdhaber: Switched to xp.inf.

scipy/stats/tests/test_resampling.py
mdhaber: Added `vectorized=True` back here.

scipy/stats/_resampling.py
mdhaber: Fixed.

scipy/stats/tests/test_resampling.py
```diff
@@ -1551,10 +1556,14 @@ def statistic1d(x, y):
         assert_allclose(res.pvalue, expected.pvalue, rtol=self.rtol)
 
     @pytest.mark.parametrize('alternative', ("less", "greater", "two-sided"))
-    def test_against_mannwhitneyu(self, alternative):
-
+    @pytest.mark.skip_xp_backends('cupy', reason='no kruskal')
+    @pytest.mark.skip_xp_backends('dask.array', reason='no kruskal')
```
mdhaber:
```suggestion
    @pytest.mark.skip_xp_backends('cupy', reason='no mwu')
    @pytest.mark.skip_xp_backends('dask.array', reason='no mwu')
```

scipy/stats/tests/test_resampling.py
```diff
@@ -1565,10 +1574,11 @@ def statistic(x, y, axis):
                                n_resamples=np.inf, alternative=alternative,
```
mdhaber:
```suggestion
                               n_resamples=xp.inf, alternative=alternative,
```

scipy/stats/tests/test_resampling.py
```diff
@@ -1598,45 +1612,40 @@ def test_vectorized_nsamp_ptype_both(self, axis):
...
+        expected_statistic = statistic(x, y, z, axis=axis)
+        expected_pvalue = pvalue(x, y, z, axis=axis)
 
         # Calculate exact and randomized permutation results
-        kwds = {'vectorized': False, 'axis': axis, 'alternative': 'greater',
+        kwds = {'axis': axis, 'alternative': 'greater',
                 'permutation_type': 'independent', 'rng': self.rng}
-        res = permutation_test(data, statistic1d, n_resamples=np.inf, **kwds)
-        res2 = permutation_test(data, statistic1d, n_resamples=1000, **kwds)
+        res = permutation_test(data, statistic, n_resamples=np.inf, **kwds)
```
mdhaber:
```suggestion
        res = permutation_test(data, statistic, n_resamples=xp.inf, **kwds)
```

scipy/stats/tests/test_resampling.py
```diff
@@ -1874,20 +1891,23 @@ def test_against_spearmanr_in_R(self, alternative, expected_pvalue):
...
-            return stats.spearmanr(x, y).statistic
+        y = stats.rankdata(y)
+        def statistic(x, axis):
+            # spearmanr is not array api compatible, but pearsonr is
+            x = stats.rankdata(x, axis=axis)
+            return stats.pearsonr(x, y, axis=axis).statistic
 
-        res = permutation_test((x,), statistic1d, permutation_type='pairings',
+        res = permutation_test((x,), statistic, permutation_type='pairings',
                                n_resamples=np.inf, alternative=alternative)
```
mdhaber:
```suggestion
                               n_resamples=xp.inf, alternative=alternative)
```

scipy/stats/tests/test_resampling.py
```diff
@@ -1488,28 +1493,27 @@ def test_randomized_test_against_exact_pairings(self):
...
-        statistic = _resampling._vectorize_statistic(statistic1d)
+        def statistic(x, axis):
+            return stats.pearsonr(x, y, axis=axis).statistic
 
         kwds = {'vectorized': True, 'permutation_type': 'samples',
                 'batch': 100, 'alternative': alternative, 'rng': rng}
-        res = permutation_test(data, statistic, n_resamples=permutations,
-                               **kwds)
+        res = permutation_test(data, statistic, n_resamples=permutations, **kwds)
         res2 = permutation_test(data, statistic, n_resamples=np.inf, **kwds)
```
mdhaber:
```suggestion
        res2 = permutation_test(data, statistic, n_resamples=xp.inf, **kwds)
```

mdhaber (commented):
Updated the `kruskal` and `mannwhitneyu` tests to run against multiple backend, skipping the ones that are not supported by the statistics.

mdhaber:
`mypy` workflow failure is unrelated. `free-threaded` is likely due to use of `setup_method` to return an RNG. Changing to create RNGs in each test function separately will look like an unrelated change that increases review burden, ~~but it's necessary to address this failure. LMK if this is OK.~~ *Just fixed the one instance that's needed to prevent the observed failure. Can fix the rest in a separate PR with others in `stats`.*

scipy/stats/tests/test_resampling.py
```diff
@@ -1598,45 +1612,40 @@ def test_vectorized_nsamp_ptype_both(self, axis):
...
-        z2 = np.broadcast_to(z, (2, 3, 4))
-        expected_statistic = statistic(x2, y2, z2, axis=axis)
-        expected_pvalue = pvalue(x2, y2, z2, axis=axis)
+        expected_statistic = statistic(x, y, z, axis=axis)
+        expected_pvalue = pvalue(x, y, z, axis=axis)
 
         # Calculate exact and randomized permutation results
-        kwds = {'vectorized': False, 'axis': axis, 'alternative': 'greater',
+        kwds = {'axis': axis, 'alternative': 'greater',
                 'permutation_type': 'independent', 'rng': self.rng}
```
mdhaber:
```suggestion
                'permutation_type': 'independent', 'rng': rng}
```

mdhaber:
I used `_xp_copy_to_numpy` to remove most of the the test skips, but copying to NumPy breaks JAX JIT. Not sure what to do. `pytest.mark.skip_xp_backends` doesn't have a `jax_jit` option like `xp_capabilities`, so does this mean I need to skip those tests with JAX entirely?

steppi:
Sorry for the delay on reviewing this. I'm going to try to get through a lot of review backlog tomorrow and will tackle this first.

> I used `_xp_copy_to_numpy` to remove most of the the test skips, but copying to NumPy breaks JAX JIT. Not sure what to do. `pytest.mark.skip_xp_backends` doesn't have a `jax_jit` option like `xp_capabilities`, so does this mean I need to skip those tests with JAX entirely?

The relevant incantation is `pytest.mark.skip_xp_backends(eager_only=True)`. Should you have been expected to know that? Of course not. Clear documentation on `xp_backends` testing practices is coming soon. Sorry for the hold up on that too.

mdhaber:
I know about `eager_only`. This is not my first rodeo. So actually, I hope you'd be surprised if I didn't know about it. It's not as specific as `jax_jit`, though. Should there not be matching options for `xp_capabilities` and `skip_xp_backends`? And it seems an important thing to call out if using `_xp_copy_to_numpy` inherently prevents testing with lazy backends?

steppi:
> I know about `eager_only`. This is not my first rodeo. So actually, I hope you'd be surprised if I didn't know about it. It's not as specific as `jax_jit`, though. Should there not be matching options for `xp_capabilities` and `skip_xp_backends`? And it seems an important thing to call out if using `_xp_copy_to_numpy` inherently prevents testing with lazy backends?

Sorry. I forget everyone is not as dense as me. I had used `eager_only` before, but had to dig before commenting to find out again what it was called. It's hard for me to remember precisely because it's different from what's used in `xp_capabilities`. I agree completely that the options should match. In this case, `eager_only` is exactly what you would want though. The use of `_xp_copy_to_numpy` here will break things for any lazy backend. 

Copying to NumPy doesn't inherently prevent testing with lazy backends for many kinds of tests. The use cases I had in mind were calculating reference arrays and performing transformations on outputs to check invariants. For those cases, `_xp_copy_to_numpy` is still compatible with lazy backends. The case is in this PR where you have a function with a function argument is trickier though. In this case, I'm thinking it may be preferable to let the function argument use the alternative backend as well and skip otherwise. The important thing would then just to be ensure that `permutation_test` or others like it still have reasonable test coverage.

mdhaber:
> In this case, I'm thinking it may be preferable to let the function argument use the alternative backend as well and skip otherwise. The important thing would then just to be ensure that permutation_test or others like it still have reasonable test coverage.

Which is what I was doing and addressed in https://github.com/scipy/scipy/pull/23772#discussion_r2458960784

> I don't really like a _requirement_ that .. we *can't* use `xp` with function _B_ in a test of function _A_ (especially in these cases in which function _A_ accepts function _B_ as an argument and uses it during execution; _B_ is not simply being used to generate reference values).

But I went ahead and tried to push this along by using `_xp_copy_to_numpy` as suggested. In any case, now we have seen tests passing both ways, so I can keep or revert the changes on a test-by-test basis as the reviewer sees fit.

steppi:
> > In this case, I'm thinking it may be preferable to let the function argument use the alternative backend as well and skip otherwise. The important thing would then just to be ensure that permutation_test or others like it still have reasonable test coverage.
> 
> Which is what I was doing and addressed in [#23772 (comment)](https://github.com/scipy/scipy/pull/23772#discussion_r2458960784)
> 
> > I don't really like a _requirement_ that .. we _can't_ use `xp` with function _B_ in a test of function _A_ (especially in these cases in which function _A_ accepts function _B_ as an argument and uses it during execution; _B_ is not simply being used to generate reference values).
> 
> But I went ahead and tried to push this along by using `_xp_copy_to_numpy` as suggested. In any case, now we have seen tests passing both ways, so I can keep or revert the changes on a test-by-test basis as the reviewer sees fit.

OK. Thinking about it. My preference here would be to not use `_xp_copy_to_numpy` in functions passed as arguments to tested functions. The idea was not to do inessential computations with the alternative backend, things which are not relevant to the behavior of the function being tested, but a function passed into the tested function to be used internally is pretty essential to the functioning of the tested function. The recommendations I made were based on my own experience with `signal`, and should not have been seen as universally applicable. Future docs will be more clear on more general principles that should be followed.

I think the right testing design principle here is for unit tests for something like `permutation_test` to use very simple functions which work across all backends. Like you said before, the other tests for all of the different statistics are more like a victory lap, and not necessary for verifying correct behavior of `permutation_test`. These could be thought of more like integration tests, with skips added as needed.

steppi:
From a correctness standpoint, everything seems good but I'm observing no speedup with CuPy for problems of any size. The bottleneck here seems to be the permutation generation which is done on CPU. Although bootstrap had it's samples generated on CPU, that was a fast operation which contributed a negligible amount to overall runtime. Do we have a policy for adding alt-backend support which offers no real performance benefits over transferring to NumPy and then computing things? I guess the changes here could be seen as an incremental improvement which is necessary for unlocking future GPU acceleration if improvements are made to resampling.

mdhaber:
> the changes here could be seen as an incremental improvement which is necessary for unlocking future GPU acceleration if improvements are made to resampling.

That is the idea. Also, since CuPy has the same rng interface as NumPy, we could use a CuPy rng to do the permutations. ~~I haven't tried it, but it might "just work".~~

steppi:
> That is the idea. Also, since CuPy has the same rng interface as NumPy, we could use a CuPy rng to do the permutations. ~I haven't tried it, but it might "just work".~

OK. I think I'm fine merging it then once the mess my previous recommendation made of the test situation is cleaned up. I think it would be good to add a TODO comment about the bottleneck situation that keeps this from being an efficient GPU implementation.

mdhaber:
BTW did you try vectorized uses with CuPy where the work can actually be done in parallel, or just large non-vectorized problems? If the statistic is sufficiently complicated and there's enough to do in parallel, permutations shouldn't be the bottleneck.

*Update: it doesn't even take many slices to see a speedup. First try:*
```python3
from scipy import stats
import numpy as np
import cupy as cp
rng = np.random.default_rng(85468924398205602)

x = rng.standard_normal(size=(10, 20))
y = rng.standard_normal(size=(10, 21))
def statistic(x, y, axis):
    return stats.ttest_ind(x, y, axis=axis).statistic

%timeit stats.permutation_test((x, y), statistic, axis=-1, permutation_type='independent')
# 158 ms ± 463 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)

x_, y_ = cp.asarray(x), cp.asarray(y)
%timeit  stats.permutation_test((x_, y_), statistic, axis=-1, permutation_type='independent')
# 83.4 ms ± 22.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

*or second try:*
```
x = rng.standard_normal(size=(10, 20))
y = rng.standard_normal(size=(10, 20))
def statistic(x, y, axis):
    return stats.pearsonr(x, y, axis=axis).statistic

%timeit stats.permutation_test((x, y), statistic, axis=-1, permutation_type='pairings')
# 127 ms ± 1.6 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

x_, y_ = cp.asarray(x), cp.asarray(y)
%timeit  stats.permutation_test((x_, y_), statistic, axis=-1, permutation_type='pairings')
# 26.7 ms ± 766 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

*or third try, with only one slice but more resamples*:
```python3
x = rng.standard_normal(size=(20))
y = rng.standard_normal(size=(20))
def statistic(x, y, axis):
    return stats.pearsonr(x, y, axis=axis).statistic

kwargs = dict(axis=-1, n_resamples=999999, permutation_type='pairings')
%timeit stats.permutation_test((x, y), statistic, **kwargs)
# 1.98 s ± 136 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

x_, y_ = cp.asarray(x), cp.asarray(y)
%timeit  stats.permutation_test((x_, y_), statistic, **kwargs)
# 1.04 s ± 11.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

steppi:
Cool, I'm seeing speedup now too. I'd picked bad examples.

mdhaber:
Went back to using `xp` throughout the test where possible and using `_xp_copy_to_numpy` as needed. (It's probably better than skipping the test for non-NumPy backends entirely.) If we'd like to see more use of the native backends:
- gh-23898
- gh-23843 + follow-up array API translation
- gh-23793 + follow-up array API translation
would help make that possible.

We might get some GPU failures, but if not, I'm happy with this.

steppi:
Thanks @mdhaber. This looks like it will be fine after some tolerance adjustments.

steppi (approved):
The failures are unrelated and appear elsewhere as well. I'm happy with this. Sorry for the delay in reviewing.

PR #23775, opened 2025-10-14
ENH: stats.power: add array API support

mdhaber:
#### Reference issue
Toward gh-20544

Adds array API support to `stats.power`.

#### Additional information
~~Should be mostly ready,  but there are a few things I want to take a closer look at.~~
Ready for another pair of eyes.


scipy/stats/_resampling.py
```diff
@@ -1060,42 +1060,51 @@ def _power_iv(rvs, test, n_observations, significance, vectorized,
...
+        raise ValueError("`significance` must contain floats between 0 and 1.")
+
     # Wrap callables to ignore unused keyword arguments
     wrapped_rvs = [_wrap_kwargs(rvs_i) for rvs_i in rvs]
 
     # Broadcast, then ravel nobs/kwarg combinations. In the end,
     # `nobs` and `vals` have shape (# of combinations, number of variables)
-    tmp = np.asarray(np.broadcast_arrays(*n_observations, *vals))
+    # todo: find a better way to do this without combining arrays
+    tmp = xp.stack(xp.broadcast_arrays(*n_observations, *vals))
```
mdhaber, line 1083: I think it's OK to merge with this part as-is, but we should probably track this idea as an enhancement request. These arrays could all be of different types, so we don't want to `stack` them if we can just broadcast and reshape them.

scipy/stats/_resampling.py
```diff
@@ -1060,42 +1060,51 @@ def _power_iv(rvs, test, n_observations, significance, vectorized,
...
         raise TypeError("`kwargs` must be a dictionary that maps keywords to arrays.")
 
     vals = kwargs.values()
     keys = kwargs.keys()
 
+    xp = array_namespace(*n_observations, significance, *vals)
+
+    significance = xp.asarray(significance)
+    if (not xp.isdtype(significance.dtype, "real floating")
+            or xp.min(significance) < 0 or xp.max(significance) > 1):
```
mdhaber, line 1074: Things like this *might* be the only reason why we need `jax_jit=False` and `allow_dask_compute=True`. That may be true of a lot of other `stats` functions, too. I think it's better to leave these as they are in this first pass and take another pass to address these things later.

scipy/stats/_resampling.py
```diff
@@ -1312,33 +1321,36 @@ def power(test, rvs, n_observations, *, significance=0.01, vectorized=None,
...
     # nobs/kwargs arrays were raveled to single axis; unravel
-    pvalues = pvalues.reshape(shape + (-1,))
+    pvalues = xp.reshape(pvalues, shape + (-1,))
     if significance.ndim > 0:
         newdims = tuple(range(significance.ndim, pvalues.ndim + significance.ndim))
-        significance = np.expand_dims(significance, newdims)
-    powers = np.mean(pvalues < significance, axis=-1)
+        significance = xpx.expand_dims(significance, axis=newdims)
+
+    float_dtype = xp_result_type(significance, xp=xp)
```
mdhaber:
```suggestion
    float_dtype = xp_result_type(significance, pvalues, xp=xp)
```

scipy/stats/tests/test_resampling.py
```diff
@@ -1183,95 +1189,109 @@ def test_input_validation(self):
...
-        assert_allclose(res.power, significance, atol=1e-2)
-
-    def test_ttest_1samp_power(self):
+        significance = xp.broadcast_to(significance[:, xp.newaxis], res.power.shape)
+        xp_assert_close(res.power, significance, atol=1e-2)
+
+    @pytest.mark.skip_xp_backends('array_api_strict',
+                                  reason='currently combines integer and float arrays')
+    @pytest.mark.skip_xp_backends(cpu_only=True,
+                                  exceptions=['cupy', 'jax.numpy', 'dask.array'])
```
mdhaber, line 1293: Torch doesn't have T distribution cdf/sf.

scipy/stats/tests/test_resampling.py
```diff
@@ -1183,95 +1189,109 @@ def test_input_validation(self):
...
         res = stats.power(test, rvs, n_observations, significance=significance)
-        significance = np.broadcast_to(significance[:, np.newaxis], res.power.shape)
-        assert_allclose(res.power, significance, atol=1e-2)
-
-    def test_ttest_1samp_power(self):
+        significance = xp.broadcast_to(significance[:, xp.newaxis], res.power.shape)
+        xp_assert_close(res.power, significance, atol=1e-2)
+
+    @pytest.mark.skip_xp_backends('array_api_strict',
+                                  reason='currently combines integer and float arrays')
```
mdhaber, line 1291: https://github.com/scipy/scipy/pull/23775/files#r2433336910

scipy/stats/_resampling.py
lucascolley: yes sounds fine to me

lucascolley (approved):
thanks Matt, LGTM!

PR #23779, opened 2025-10-15
ENH: stats.false_discovery_control: add array API support

mdhaber:
#### Reference issue
Toward gh-20544

Adds array API support to `stats.false_discovery_control`.

#### Additional information
No relevant failures in full CI run on 6c147cd, so I committed the whitespace fix with `[skip ci]`.

### DIFF
--- a/scipy/stats/_morestats.py
+++ b/scipy/stats/_morestats.py
@@ -17,6 +17,8 @@ from scipy._lib._array_api import (
     array_namespace,
     is_marray,
     xp_capabilities,
+    is_numpy,
+    is_jax,
     xp_size,
     xp_vector_norm,
     xp_promote,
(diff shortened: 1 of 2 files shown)


scipy/stats/_morestats.py
```diff
@@ -4650,25 +4655,42 @@ def false_discovery_control(ps, *, axis=0, method='bh'):
...
 
     # Restore original order of axes and data
-    np.put_along_axis(ps, order, values=ps.copy(), axis=-1)
-    ps = np.moveaxis(ps, -1, axis)
+    ps = _reorder_along_axis(ps, order, axis=-1, xp=xp)
+    ps = xp.moveaxis(ps, -1, axis)
 
-    return np.clip(ps, 0, 1)
+    return xp.clip(ps, 0., 1.)
+
```
mdhaber, line 4688:
```suggestion


```

tylerjereddy (approved):
Looks like a reasonable "translation" to array API to me. There are some specialized code paths--pretty sure the ship has sailed on trying to avoid those (in the short term) at this point though.

The one CI failure is not related.

Merging, thanks Matt.

mdhaber:
Thanks @tylerjereddy. If `put_along_axis` makes it into the standard (https://github.com/data-apis/array-api/issues/177), we can probably remove one of those specialized code paths. The other requires the ufunc `accumulate` method, though - it's worth the separate code path for NumPy. Thanks for taking a look!

Issue #23355, opened 2025-07-19
Chatterjee's Xi returns different results based on `numpy` major version


cooperjaXC:
### Describe your issue.

# Issue
Depending on the input data, the coefficient and p-value results of running `scipy.stats.chatterjeexi()` often returns different results based on whether the version of `numpy` installed is >=2.0.0 or <2.0.  Those using `numpy>=2.0` seem to be more correct, but this has not been verified. This is unusual because other correlation functions' coefficients, like Pearson's R and Spearman's Rho, don't have changing coefficients based on numpy's version. And, according to [Chatterjee (2021)](https://doi.org/10.1080/01621459.2020.1758115), this coefficient should not either. 

The _Xi_ coefficient should be between 0-1, but using a proprietary dataset (that I cannot share here) of 14,366 observations, the results using the identical data from the same dataset in the same order are as follows:

> Scipy v1.16.0 with Numpy at v1.26.4 ran Chatterjee's Xi and got a corr=1.2256     w/ P-value=0.0

versus

> Scipy with Numpy at v2.0.0 ran Chatterjee's Xi and got a corr=0.2087*** w/ P-value=0.0

The only variable changed here is the numpy version. The latter with v2.0.0 seems more appropriate, both because it is within the range Chatterjee sets for appropriate calculation of his coefficient, and it lines up a bit better with Spearman's Rho coefficient for the same two variables. 

# Cause
My best guess, based on my limited testing, is that this arises from the following code in scipy.stats._correlation.py:

https://github.com/scipy/scipy/blob/9ae478ddaec6e1a71ca31624199e1ea171835eec/scipy/stats/_correlation.py#L14-L24

The `numpy==2.0.0` [release notes](https://github.com/numpy/numpy/releases/tag/v2.0.0#:~:text=Minor%20changes%20in%20behavior%20of%20sorting%20functions) detail that 

> methods that aren't stable may return slightly different results in 2.0.0 compared to 1.26.x. This includes the default method of numpy.argsort. 

Testing shows that the Xi statistic's computing using numpy's argsort, broadcast_arrays, and take_along_axis all differ based on numpy's version as 1.26.x vs 2.0.0. Because _Xi_ is a rank-order coefficient, my hunch is that the ordering of one or all of these functions differs because of the introduction of `stable` in numpy v2.0 to the `argsort` function. I'm not sophisticated enough to thoroughly test this, though. 

# Notes

[...]

PR #23782, opened 2025-10-15
ENH: speedup quantile by not using stable sort

cakedev0:

This PR adds `stable=False` for the sort in the quantile computation. 
Array API sort is stable by default, but in the quantile computation, the sort doesn't need to be stable. And stable sort is 6-9x slower than normal sort.

#### Reference issue

No issue in this repo.

Related issues: 
- stable sort is slower than sort: https://github.com/scikit-learn/scikit-learn/issues/32283
- https://github.com/data-apis/array-api/issues/976

Related PRs: 
- similar merged PR in scikit-learn for weighted quantile: https://github.com/scikit-learn/scikit-learn/pull/32285
- similar merged PR in numpy for weighted quantile: https://github.com/numpy/numpy/pull/29837

#### Why stable sort is not needed:

The result of the `quantile` function depends only on the multiset of values and their ranks, not on the relative order of tied values. Stability only affects the order among equals; every place the code uses the sorted data, ties collapse to the same numeric outcome anyway. You can get convinced by running this script:

<details>

```Python
import numpy as np
from scipy.stats import quantile

n = 1000
methods = {
    'inverted_cdf', 'averaged_inverted_cdf', 'closest_observation',
    'hazen', 'interpolated_inverted_cdf', 'linear',
    'median_unbiased', 'normal_unbiased', 'weibull',
    'harrell-davis'
}

def assert_quantile_result_invariant_to_suffled_input(x, p, method, nan_policy):
    expected = quantile(x, p, method=method, nan_policy=nan_policy)
    for _ in range(10):
        np.random.shuffle(x)
        actual = quantile(x, p, method=method, nan_policy=nan_policy)
        assert np.array_equal(expected, actual, equal_nan=True)

for method in methods:
    for nan_policy in ["propagate", "omit"]:
        x = np.random.randint(0, n, size=n)  # has some duplicates
        p = np.concat(([0], np.random.rand(100), [1]))
        assert_quantile_result_invariant_to_suffled_input(x, p, method, nan_policy)
        x = x.astype('float64')
        x[np.random.rand(n) < 0.1] = np.nan
        assert_quantile_result_invariant_to_suffled_input(x, p, method, nan_policy)
```
</details>

Also, notice that numpy implementation uses `partition`, which isn't stable (it uses "introselect", see notes [here](https://numpy.org/doc/2.3/reference/generated/numpy.partition.html)):
https://github.com/numpy/numpy/blob/v2.3.0/numpy/lib/_function_base_impl.py#L4839-L4862


mdhaber:
I would just be concerned about reports like gh-23355, in which `stable` affected the results of the calculation. Could that happen here?

cakedev0:
Fair concern, I added section "Why stable sort is not needed" with my arguments in the PR description.

Also, some CI pipelines are failing but I think it's unrelated to this PR, so marking it as ready.

cakedev0:
Hum there is one case for which stability affects the result of `quantile`: if the result can be 0+ or 0-.

- if you use stable sort, 0+ or 0- will be returned only depending on the order of the input, and not the underlying sort implementation
- if you use stable sort, 0+ or 0- will be returned depending on the order of the input, and the underlying sort implementation.

But given than `np.float64(0.) == np.float64(-0.)` is True, I feel this is ok?

mdhaber:
> But given than `np.float64(0.) == np.float64(-0.)` is True, I feel this is ok?

I'm guessing yes. I think I would prioritize the speedup over this distinction. If you can produce an example that would expose the difference, that would be helpful. I will take a closer look at that and the rationale in the top post this afternoon, and hopefully we can get this merged today.

cakedev0:
Here is an example of the current behavior "*0+ or 0- will be returned depending on the order of the input*":

```Python
import numpy as np
from scipy.stats import quantile

x = np.array([-0., -0., 0.])
p = np.linspace(0, 1, num=11)[1:-1]
print(quantile(x, p))
# prints: [-0. -0. -0. -0.  0.  0.  0.  0.  0.]
print(quantile(x[::-1], p))
# prints: [ 0.  0.  0.  0. -0. -0. -0. -0. -0.]
```

Note that in this example, `np.quantile` output is different:
```Python
print(np.quantile(x, p))
# prints: [ 0.  0. -0. -0.  0.  0.  0.  0.  0.]
print(np.quantile(x[::-1], p))
# prints: [ 0.  0.  0.  0.  0.  0.  0. -0. -0.]
```

---
And an example of the impact of stable=True/False: 
```Python
x = np.array([-0.]*20 + [0.])
p = np.linspace(0, 1, num=11)[1:-1]
print(quantile(x, p))
# prints: [-0. -0. -0. -0. -0. -0. -0. -0. -0.] for stable=True
# prints: [ 0.  0.  0. -0.  0. -0.  0.  0.  0.] for stable=False
```

mdhaber:
OK. I think I could live with that. If there were no changes at all, I would merge, but let's be sure that nobody else sees a problem with the zeros changing.

To do that, I'd ask for just one more thing: please post a short summary to the discussion forum describing the change and mentioning the benefit (significant speedup to the function overall) at the risk that the numerical representations of `0` in the output might change *if* the user includes both `0.` and `-0.` in the input. 

If there are no concerns after a week (and no other B.C. concerns identified), I'll go ahead and merge. That still gives plenty of time to get this change in the next release.

mdhaber:
Discussion forum is here: https://discuss.scientific-python.org/c/contributor/scipy/

tylerjereddy (commented):
Looks like there are no `asv` `quantile` benchmarks in our suite at the moment. Drafting a simple one that uses NumPy perhaps predictably doesn't show evidence of improvement:

<details>

```diff
--- a/benchmarks/benchmarks/stats.py
+++ b/benchmarks/benchmarks/stats.py
@@ -814,3 +814,11 @@ class RandomTable(Benchmark):
 
     def time_method(self, method, ntot, ncell):
         self.dist.rvs(1000, method=method, random_state=self.rng)
+
+class Quantile(Benchmark):
+    def setup(self):
+        self.x = np.asarray([[10, 8, 7, 5, 4],
+                             [0, 1, 2, 3, 5]])
+
+    def time_quantile(self):
+        stats.quantile(self.x, 0.5)
```

</details>

 `asv continuous -E virtualenv -e -b "time_quantile" main optim-quantile-unstable-sort`

`BENCHMARKS NOT SIGNIFICANTLY CHANGED.`

`torch.sort()` also seems to default to `stable=False`. `jax.numpy.sort` does default to `True`, so maybe that is one of the cases that might be experiencing the substantial performance improvement here? But the array API support table for SciPy `quantile` has negative for JAX support on both CPU and GPU. So can I ask what is the practical use case where a substantial speedup was observed and how can we reproduce it?

mdhaber:
> doesn't show evidence of improvement

The execution time in that benchmark would be dominated by NumPy function call overhead because the array is so small. Try large `x`, maybe 100000 elements. The computational complexity of `argsort` used in `quantile` should start to show.

```python3
import numpy as np
rng = np.random.default_rng(23849293582854)

# dominated by function call overhead
x = rng.random(10)
%timeit np.argsort(x, stable=False)  # 3.13 µs ± 982 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
%timeit np.argsort(x, stable=True)  # 2.03 µs ± 374 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

# algorithm begins to matter
x = rng.random(100000)
%timeit np.argsort(x, stable=False)  # 4.64 ms ± 120 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
%timeit np.argsort(x, stable=True)  # 13.9 ms ± 1.47 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

cakedev0:
(Note: it's `sort` not `argsort`, the speed-up is even bigger for `sort` than for `argsort`)

cakedev0:
With:
```Python
class Quantile(Benchmark):
    params = [
        [10, 100, 1000, 10_000, 100_000, 1_000_000],
    ]

    def setup(self, n):
        self.x = np.random.rand(n)

    def time_quantile(self, n):
        stats.quantile(self.x, 0.5)
```

You get:

| Change   | Before [7d63c6e0] <main>   | After [4956c95c] <optim-quantile-unstable-sort>   |   Ratio | Benchmark (Parameter)                 |
|----------|----------------------------|---------------------------------------------------|---------|---------------------------------------|
| -        | 884±200μs                  | 418±40μs                                          |    0.47 | stats.Quantile.time_quantile(10000)   |
| -        | 8.15±1ms                   | 1.53±0.09ms                                       |    0.19 | stats.Quantile.time_quantile(100000)  |
| -        | 94.2±20ms                  | 15.1±0.8ms                                        |    0.16 | stats.Quantile.time_quantile(1000000) |

tylerjereddy:
ok, so maybe add at least one benchmark to the suite so we can track it?

cakedev0:
Yes, for sure. Should I open another PR to add the benchmark? And would we merge it before this one so the speed-up is recorded?

cakedev0:
> please post a short summary to the discussion forum describing the change [...]

Done: https://discuss.scientific-python.org/t/proposal-speed-up-scipy-stats-quantile-by-using-an-unstable-sort/2140?u=cakedev0

rkern:
:+1: 

FWIW, IMO signed 0s changing signs is well below the threshold of a "check with the mailing list" behavior change. Unless if we're talking about a special function in the complex domain, and the signed 0s are representing carefully-chosen branch cuts, maintaining the sign of 0s is kind of a crap shoot anyways.

cakedev0:
I think we can merge this PR?

mdhaber:
CI failures are due to gh-23869.
No strong objections on mailing list - there was some discussion about whether stable sort would actually be faster for structured data, but it sounds like having a consistently fast unstable sort is preferable.
Benchmarks on CircleCI show the speedup, especially for larger arrays.
Before:
```
[46.14%] ··· stats.Quantile.time_quantile                                    ok
[46.14%] ··· ========= ========== ==========
             --                  d          
             --------- ---------------------
                size       1         100    
             ========= ========== ==========
               10000    2.97±0ms   2.51±0ms 
               100000   11.0±0ms   7.23±0ms 
              1000000   106±0ms    67.7±0ms 
             ========= ========== ==========
```
After:
```
[46.14%] ··· stats.Quantile.time_quantile                                    ok
[46.14%] ··· ========= ========== ==========
             --                  d          
             --------- ---------------------
                size       1         100    
             ========= ========== ==========
               10000    2.42±0ms   2.31±0ms 
               100000   3.39±0ms   3.19±0ms 
              1000000   15.4±0ms   12.6±0ms 
             ========= ========== ==========
```
so in it goes.

mdhaber (approved):

charris:
That it quite a speedup, I expect it is because SIMD quicksort is used. For plain old C the difference usually runs maybe 30%, and for some hardware/compilers, stable can be a bit faster.

PR #23787, opened 2025-10-15
BUG:optimize:SLSQP: Use double negative in an if condition for NaN handling

ilayn:
#### Reference issue
https://github.com/scikit-hep/pyhf/issues/2593
https://discuss.scientific-python.org/t/minimization-result-success-change-after-slsqp-fortran-to-c-rewrite/2136

As reported in the pyhf issue, the contrapositive condition does not handle NaN the way old Fortran code handles. Hence reverted to the original condition and used the negated value. Since the report was not bare SciPy call, I did not find a good example to add a test for this branch.

@tylerjereddy if we do not have any plans for 1.16.3, no problem, please remove the backport candidate label.

### DIFF
--- a/scipy/optimize/__slsqp.c
+++ b/scipy/optimize/__slsqp.c
@@ -240,7 +240,7 @@ MODE1:
     }
     S->h1 = S->t - S->t0;
 
-    if ((S->h1 > (S->h3 / 10.0)) && (S->line <= 10))
+    if (!((S->h1 <= (S->h3 / 10.0)) || (S->line > 10)))
     {
         S->alpha = fmax(S->h3/(2.0*(S->h3 - S->h1)), alfmin);
         goto LINE_SEARCH;


andyfaff:
From reading the code it seems to do the same thing. Which value was becoming NaN? Did you check if this fixes the downstream issue?

ilayn:
Yes it's `h1` becoming `NaN` and not falling into the block. Because for `NaN` values that's not a contra-positive it is false in both cases which should switch its truth value. So I made the mistake of converting `NOT (False OR False)` to `(False AND True)`. 

Downstream code now runs properly. 

So this 

```python
h1, h3, line = np.nan, -81.5, 1
if (h1 > h3/10.0) and (line <= 10):
     print("Contrapositive")
elif not((h1 <= h3/10.0) or (line > 10)):
     print("Double Negative")

```

prints double negative.

andyfaff:
thank you @ilayn

tylerjereddy:
@ilayn @matthewfeickert what's the severity here? An end user could get an optimization result object that contains a failure result in it, but if they aren't checking that then workflows could silently change without an exception? Or will this error out at least?

I don't mind doing `1.16.3`, but an assessment of the severity might be useful. It looks like `pyhf` is currently blocking all usage of SciPy `1.16.x` in their toolchain if I'm reading correctly, which I suppose is kind of bad.

matthewfeickert:
> what's the severity here? An end user could get an optimization result object that contains a failure result in it, but if they aren't checking that then workflows could silently change without an exception? Or will this error out at least?

@tylerjereddy For pyhf we catch and raise the error, so it is very clear that there's an issue to users (nothing is silently failing and pathologically changing results). Apologies that I didn't take the time in July 2025 when we first noticed this to dig in and raise the issue earlier &mdash; we assumed there was a subtle API changed we missed. To users it is pretty clear that something is wrong, as they might hit this quite quickly in any statistical inference steps that they are performing.

> I don't mind doing `1.16.3`, but an assessment of the severity might be useful. It looks like `pyhf` is currently blocking all usage of SciPy `1.16.x` in their toolchain if I'm reading correctly, which I suppose is kind of bad.

Correct, we currently disallow `v1.16.0` to `v1.16.2` in our codebase (but don't have a release with this yet), and are in the process of getting ready for a release which will at least get these out to users on PyPI (we've already patched these disallows into our conda-forge distributions). We communicated to many of our downstream users in July 2025 that they might need to restrict their SciPy installs with pyhf, and some of the libraries that build on pyhf have already implemented some similar restrictions when all of their tests started failing.

So while a `v1.16.3` release would be useful for us, it seems like:
* This might be a rather niche bug as it seems that we are the only SciPy users to notice it and ask a question in multiple months.
* From pyhf's perspective, it doesn't really matter if this PR's patch gets introduced in `v1.16.3` or `v1.17.0`. If a `v1.16.3` patch release could be made with low effort on the SciPy side that's great, but if you're already in the process of moving towards a `v1.17.0` release in the next week or so and cutting a patch release would slow that work down then that isn't a "problem" from pyhf's side.

tylerjereddy:
Thanks for all the details--for reference, `v1.17.0` would typically not be released until early January 2026.

matthewfeickert:
> For reference, `v1.17.0` would typically not be released until early January 2026.

Ah okay, good to know. In that case, if a v1.16.3 patch release would be viable that would be useful.

PR #23788, opened 2025-10-15
TST: some more test tol bumps

andyfaff:
@mdhaber looks like there's another temperamental test fail on macosx-x86_64/Accelerate. Doesn't happen all the time.

There's also another test flake:

```
_________________ TestDerivative.test_maxiter_callback[numpy] __________________

self = <scipy.differentiate.tests.test_differentiate.TestDerivative object at 0x10d033890>
xp = <module 'scipy._lib.array_api_compat.numpy' from '/Users/runner/work/scipy/scipy/build-install/usr/lib/python3.13/site-packages/scipy/_lib/array_api_compat/numpy/__init__.py'>

    def test_maxiter_callback(self, xp):
        # Test behavior of `maxiter` parameter and `callback` interface
        x = xp.asarray(0.612814, dtype=xp.float64)
        maxiter = 3
    
        def f(x):
            res = special.ndtr(x)
            return res
    
        default_order = 8
        res = derivative(f, x, maxiter=maxiter, tolerances=dict(rtol=1e-15))
        assert not xp.any(res.success)
        assert xp.all(res.nfev == default_order + 1 + (maxiter - 1)*2)
        assert xp.all(res.nit == maxiter)
    
        def callback(res):
            callback.iter += 1
            callback.res = res
            assert hasattr(res, 'x')
            assert float(res.df) not in callback.dfs
            callback.dfs.add(float(res.df))
            assert res.status == eim._EINPROGRESS
            if callback.iter == maxiter:
                raise StopIteration
        callback.iter = -1  # callback called once before first iteration
        callback.res = None
        callback.dfs = set()
    
        res2 = derivative(f, x, callback=callback, tolerances=dict(rtol=1e-15))
        # terminating with callback is identical to terminating due to maxiter
        # (except for `status`)
        for key in res.keys():
            if key == 'status':
                assert res[key] == eim._ECONVERR
                assert res2[key] == eim._ECALLBACK
            else:
>               assert res2[key] == callback.res[key] == res[key]
E               assert np.float64(8.526512829121202e-14) == np.float64(8.171241461241152e-14)

callback   = <function TestDerivative.test_maxiter_callback.<locals>.callback at 0x120e54b80>
default_order = 8
f          = <function TestDerivative.test_maxiter_callback.<locals>.f at 0x120e54a40>
key        = 'error'
maxiter    = 3
res        =      success: False
      status: -2
          df: 0.3306453155026219
       error: 8.171241461241152e-14
         nit: 3
        nfev: 13
           x: 0.612814
res2       =      success: False
      status: -4
          df: 0.3306453155026219
       error: 8.526512829121202e-14
         nit: 3
        nfev: 13
           x: 0.612814
self       = <scipy.differentiate.tests.test_differentiate.TestDerivative object at 0x10d033890>
x          = array(0.612814)
xp         = <module 'scipy._lib.array_api_compat.numpy' from '/Users/runner/work/scipy/scipy/build-install/usr/lib/python3.13/site-packages/scipy/_lib/array_api_compat/numpy/__init__.py'>

scipy/differentiate/tests/test_differentiate.py:296: AssertionError
```

What would you like to do with the differentiate fail?

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -119,7 +119,7 @@ jobs:
     needs: get_commit_message
     if: >
       needs.get_commit_message.outputs.message == 1
-      && (github.repository == 'andyfaff/scipy' || github.repository == '')
+      && (github.repository == 'scipy/scipy' || github.repository == '')
     runs-on: macos-15-intel
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -119,7 +119,7 @@ jobs:
     needs: get_commit_message
     if: >
       needs.get_commit_message.outputs.message == 1
-      && (github.repository == 'scipy/scipy' || github.repository == '')
+      && (github.repository == 'andyfaff/scipy' || github.repository == '')
     runs-on: macos-15-intel
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
--- a/scipy/differentiate/tests/test_differentiate.py
+++ b/scipy/differentiate/tests/test_differentiate.py
@@ -292,6 +292,11 @@ class TestDerivative:
             if key == 'status':
                 assert res[key] == eim._ECONVERR
                 assert res2[key] == eim._ECALLBACK
+            elif key == 'error':
+                # switched from equality check to accommodate
+                # macosx-x86_64/Accelerate
+                xp_assert_close(res2[key], res[key], rtol=1e-14)
+                xp_assert_close(callback.res[key], res[key], rtol=1e-14)
             else:
                 assert res2[key] == callback.res[key] == res[key]
 
--- a/scipy/stats/tests/test_axis_nan_policy.py
+++ b/scipy/stats/tests/test_axis_nan_policy.py
@@ -471,7 +471,8 @@ def _axis_nan_policy_test(hypotest, args, kwds, n_samples, n_outputs, paired,
             res = hypotest(*data1d, *args, nan_policy=nan_policy, **kwds)
         res_1db = unpacker(res)
 
-        assert_allclose(res_1db, res_1da, rtol=1e-15)
+        # changed from 1e-15 solely to appease macosx-x86_64+Accelerate
+        assert_allclose(res_1db, res_1da, rtol=4e-15)
         res_1d[i] = res_1db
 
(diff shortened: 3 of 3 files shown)


mdhaber:
Thanks for doing these. 

This one I need to take a closer look at. Is Accelerate arithmetic non-deterministic between runs on a given machine? (I wouldn't be so surprised given the sporadic failures we've seen, but it sounds like a bad idea...) If not, we need to trace down the difference between these two cases.

andyfaff:
https://github.com/numpy/numpy/issues/29873#issuecomment-3394919165

mdhaber (approved):
I maintain that it is reasonable for a consenting adult to *want* to test for floating point equality. In this case, the test is designed to check that the only difference between the two cases is the criterion used to determine when the iterative algorithm terminates; every Python instruction until that point should be identical[^1]. I guess we just can't do that anymore. 

[^1]: I don't really care whether the outputs are identical or not - give me another way to check whether the arrays are the result of the same set of Python instructions and I'll use that.

mdhaber:
But there is a new failure because Accelerate decided to do something else differently this time.

andyfaff:
I know where you're coming from. I'm as perturbed as you are. I'm taking the pragmatic view that we'll stop supporting macosx-x86_64 in a couple of years and the issue will go away.

Ralf's comment is that we shouldn't necessarily expect bitwise reproducibility for a BLAS calculation, because the library seems to be able to order calculations differently on two runs through. The reproblas link was informative. I guess those things can accumulate over repeated calls.

mdhaber:
Re: https://github.com/scipy/scipy/pull/23788#issuecomment-3408968878, I meant that there was still a test failure here : ) that's why I waited to merge. I'll bump it.

andyfaff:
That's a test fail that I've not seen before, the rest of the test flakes I knew about.

andyfaff:
@mdhaber when you're bumping the tolerance in `scipy/sparse/linalg/_isolve/tests/test_iterative.py::test_show` can you also change the `xp_assert_close` in `test_differentiate` to an atol? i.e.

```
xp_assert_close(res2[key], res[key], atol=1e-14)
xp_assert_close(callback.res[key], res[key], atol=1e-14)
```
This is because the value being compared is very close to zero, the correct thing is to use an atol not an rtol.

mdhaber:
> This is because the value being compared is very close to zero, the correct thing is to use an atol not an rtol.

Unless there is subtractive cancellation or some other reason that the relative tolerance can't be satisfied, relative tolerance is still preferable. But yeah, in this case, the error is estimated as the difference between derivative estimates in successive iterations, so `atol` is OK. I'll change it.

PR #23793, opened 2025-10-16
ENH: stats.ks_2samp: vectorize statistic implementation

mdhaber:
#### Reference issue
Toward gh-14651
Toward gh-20544

This adds an `axis` argument to the `stats.ks_2samp` function itself so the `_axis_nan_policy` decorator does not need to loop over each axis slice for n-d input. Vectorized calls are now used to calculate the test statistic; the p-value calculation still uses `np.vectorize`, ~~but we can probably use boost (ufunc) for the approximate p-value calculation~~, and perhaps the exact p-value calculation can be turned into a ufunc someday. Either way, this paves the way toward array API support,

#### Additional information
The first commit refactors `ks_2samp` to use `stats.rankdata` instead of `np.searchsorted`, providing a small speed boost for large arrays and allowing for vectorization (`np.searchsorted` only works with a 1d input array). The second commit passes `axis`/`keepdims` arguments and such as needed throughout the statistic calculation to vectorize it, providing a speed boost for large batches of arrays. It splits the p-value calculation into a separate function with the `np.vectorize` decorator.

Boost's Kolmogorov-Smirnov distribution is the same as our `kstwobign`. This is different from our `kstwo` distribution, which I assume is more accurate, and the exact calculation in `_attempt_exact_2kssamp`, which I assume is more accurate still.

<details>

```python3
import numpy as np
from scipy import stats
rng = np.random.default_rng(234823489298838)
from scipy.special import _ufuncs as scu
from scipy.stats._stats_py import _attempt_exact_2kssamp

rng = np.random.default_rng(234823489298838)
n1 = 11
n2 = 12
x = rng.random(n1)
y = rng.random(n2)
res = stats.ks_2samp(x, y, method='asymp', alternative='two-sided')

n = n1 + n2
g = np.gcd(n1, n2)
x = res.statistic

m, n = sorted([float(n1), float(n2)], reverse=True)
en = m * n / (m + n)

print(scu._kstwo_sf(x, en),
      stats.kstwo.sf(x, int(round(en))),
      stats.kstwobign.sf(np.sqrt(en)*x),
      _attempt_exact_2kssamp(n1, n2, g, x, 'two-sided')[-1],
      res.pvalue)
```

</details>


scipy/stats/_stats_py.py
```diff
@@ -7981,42 +7983,68 @@ def ks_2samp(data1, data2, alternative='two-sided', method='auto'):
     if alternative not in ['two-sided', 'less', 'greater']:
         raise ValueError(f'Invalid value for alternative: {alternative}')
     MAX_AUTO_N = 10000  # 'auto' will attempt to be exact if n1,n2 <= MAX_AUTO_N
-    if np.ma.is_masked(data1):
-        data1 = data1.compressed()
-    if np.ma.is_masked(data2):
-        data2 = data2.compressed()
```
mdhaber, line 7986: These were unused after gh-19650.

scipy/stats/_stats_py.py
```diff
@@ -7981,42 +7983,68 @@ def ks_2samp(data1, data2, alternative='two-sided', method='auto'):
...
+    d_sign = np.astype(np.where(selector, -1, 1), np.int8)
+    prob = _ks_2samp_prob(d, n1, n2, mode, MAX_AUTO_N, alternative)
+
+    if mode == 'exact' or (max(n1, n2) <= MAX_AUTO_N):
+        g = np.gcd(n1, n2)
+        lcm = (n1 // g) * n2
+        h = np.round(d * lcm)
+        d = h / lcm
+        h_zero = (h == 0)
+        prob[h_zero] = 1.0
```
mdhaber: This code was (and still is) in `_attempt_exact_2kssamp`, and it modifies the test statistic. I moved it out of there because without it, a floating point equality check in `test_location_sign` started failing. I think it just cleans up the test statistic slightly, e.g. nudges from `np.float64(0.09999999999999998)` to `np.float64(0.1)`.

The funny thing is that this code was not running at all when the user passed `mode='asymp'`. In SciPy 1.16.2:
```python3
import numpy as np
from scipy import stats
x = np.arange(10, dtype=np.float64)
y = x.copy()
x[6] = 6.1
stats.ks_2samp(x, y, method='asymp').statistic  # np.float64(0.09999999999999998)
stats.ks_2samp(x, y, method='exact').statistic  # np.float64(0.1)
```
The choice of p-value calculation method shouldn't be changing the test statistic returned to the user, significantly or otherwise. So I suspect this code should either always run or never run. 

If it's OK with the reviewer, I'll just get rid of this here and change the equality check in the test to an `assert_allclose` with a tight tolerance. That will be helpful because `gcd` isn't in the array API.

```suggestion
```

scipy/stats/_stats_py.py
```diff
@@ -7981,42 +7983,68 @@ def ks_2samp(data1, data2, alternative='two-sided', method='auto'):
...
-        data1 = data1.compressed()
-    if np.ma.is_masked(data2):
-        data2 = data2.compressed()
-    data1 = np.sort(data1)
-    data2 = np.sort(data2)
-    n1 = data1.shape[0]
-    n2 = data2.shape[0]
+
+    data1 = np.sort(data1, axis=-1, stable=False)
+    data2 = np.sort(data2, axis=-1, stable=False)
```
mdhaber:
```suggestion
    data1 = np.sort(data1, axis=-1)
    data2 = np.sort(data2, axis=-1)
```
Oops, guess I can't do this until array API kicks in.

scipy/stats/_stats_py.py
```diff
@@ -7981,42 +7983,71 @@ def ks_2samp(data1, data2, alternative='two-sided', method='auto'):
...
-    cdf1 = np.searchsorted(data1, data_all, side='right') / n1
-    cdf2 = np.searchsorted(data2, data_all, side='right') / n2
-    cddiffs = cdf1 - cdf2
+    cdf_counts1 = np.diff(ranks[..., :n1], prepend=1, append=n + 1, axis=-1)
+    cdf_counts2 = np.diff(ranks[..., -n2:], prepend=1, append=n + 1, axis=-1)
+    cdf1_vals, cdf1_counts = np.broadcast_arrays(np.linspace(0, 1, n1 + 1), cdf_counts1)
+    cdf2_vals, cdf2_counts = np.broadcast_arrays(np.linspace(0, 1, n2 + 1), cdf_counts2)
+    cdf1 = np.repeat(np.ravel(cdf1_vals), np.ravel(cdf1_counts), axis=-1)
+    cdf2 = np.repeat(np.ravel(cdf2_vals), np.ravel(cdf2_counts), axis=-1)
+    cddiffs = np.reshape(cdf1 - cdf2, ranks.shape[:-1] + (-1,))
```
mdhaber: I will be submitting a PR with a vectorized `xp_searchsorted`. If this is hard to review, and we can try using that to more directly translate what is here.

mdhaber:
I went ahead and kept the implementation in terms of `_rankdata` rather than switching to `xp_searchsorted` (because it's a lot faster), but I refactored and added code comments to make it a lot easier to follow. I think this is in good shape.

scipy/stats/_stats_py.py
```diff
@@ -7826,9 +7826,9 @@ def ks_2samp(data1, data2, alternative='two-sided', method='auto'):
         Defines the method used for calculating the p-value.
         The following options are available (default is 'auto'):
 
-          * 'auto' : use 'exact' for small size arrays, 'asymp' for large
-          * 'exact' : use exact distribution of test statistic
-          * 'asymp' : use asymptotic distribution of test statistic
+        * 'auto' : use 'exact' for small size arrays, 'asymp' for large
+        * 'exact' : use exact distribution of test statistic
+        * 'asymp' : use asymptotic distribution of test statistic
```
j-bowhay, line 7840: I don't have a strong opinion but for consistency should we manually add docs for `axis` too?

mdhaber: The indentation will be fixed merging main, but yeah, i've added `axis` documentation.

j-bowhay (approved):

PR #23795, opened 2025-10-17
MAINT: stats.iqr: prepare for array API translation

mdhaber:
#### Reference issue
Toward gh-20544

This refactors `stats.iqr` to prepare for array API translation:
- removes cruft (no longer needed due to use of `_axis_nan_policy` decorator
- uses `stats.quantile` instead of `np.percentile`
- strengthens some aspects of tests that seemed missing

#### Additional information
I had to implement the legacy `percentile` methods "lower", "higher", "midpoint", and "nearest" to replace `percentile` with `quantile`. The NumPy documentation about them is not exactly right, so I wrote them differently than described to match the actual behavior.

### DIFF
--- a/scipy/stats/_quantile.py
+++ b/scipy/stats/_quantile.py
@@ -349,7 +349,7 @@ def _quantile_hd(y, p, n, xp):
 def _quantile_bc(y, p, n, method, xp):
     # Methods retained for backward compatibility. NumPy documentation is not
     # quite right about what these methods do: if `p * (n - 1)` is integral,
-    # that is used as the index.
+    # that is used as the index. See numpy/numpy#28910.
     ij = p * (n - 1)
     if method == '_midpoint':
         return (xp.take_along_axis(y, xp.astype(xp.floor(ij), xp.int64), axis=-1)
--- a/scipy/stats/_quantile.py
+++ b/scipy/stats/_quantile.py
@@ -42,7 +42,7 @@ def _quantile_iv(x, p, method, axis, nan_policy, keepdims):
     methods = {'inverted_cdf', 'averaged_inverted_cdf', 'closest_observation',
                'hazen', 'interpolated_inverted_cdf', 'linear',
                'median_unbiased', 'normal_unbiased', 'weibull',
-               'harrell-davis'}
+               'harrell-davis', '_lower', '_midpoint', '_higher', '_nearest'}
     if method not in methods:
         message = f"`method` must be one of {methods}"
         raise ValueError(message)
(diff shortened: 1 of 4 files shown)


scipy/stats/_quantile.py
```diff
@@ -342,3 +344,20 @@ def _quantile_hd(y, p, n, xp):
     w = xpx.at(w, xp.isnan(w)).set(0)
     res = xp.vecdot(w, y, axis=-1)
     return xp.moveaxis(res, 0, -1)
+
+
+def _quantile_bc(y, p, n, method, xp):
+    # Methods retained for backward compatibility. NumPy documentation is not
+    # quite right about what these methods do: if `p * (n - 1)` is integral,
+    # that is used as the index.
```
lucascolley: upstream issue?

scipy/stats/tests/test_quantile.py
```diff
@@ -191,14 +192,15 @@ def test_size_0(self, axis, keepdims, xp):
...
+        ['inverted_cdf', 'averaged_inverted_cdf', 'closest_observation',
+         '_lower', '_higher', '_midpoint', '_nearest'])
     def test_transition(self, method, xp):
         # test that values of discontinuous estimators are correct when
         # p*n + m - 1 is integral.
         if method == 'closest_observation' and np.__version__ < '2.0.1':
             pytest.skip('Bug in np.quantile (numpy/numpy#26656) fixed in 2.0.1')
         x = np.arange(8., dtype=np.float64)
-        p = np.arange(0, 1.0625, 0.0625)
+        p = np.arange(0, 1.03125, 0.03125)
```
lucascolley, line 203: reason for this change?

lucascolley (approved):
thanks Matt, LGTM!

scipy/stats/_quantile.py
mdhaber: Didn't file one. It's only noticeable if you're looking for it.

scipy/stats/tests/test_quantile.py
mdhaber: More points. I wanted to check on the boundaries of the step transitions and at a few points in-between.

scipy/stats/_quantile.py
lucascolley: okay, would be nice but I won't block on it

mdhaber: There was already an issue, numpy/numpy#28910.
```suggestion
    # Methods retained for backward compatibility. NumPy documentation is not
    # quite right about what these methods do: if `p * (n - 1)` is integral,
    # that is used as the index. See numpy/numpy#28910.
```

scipy/stats/_stats_py.py
```diff
@@ -3186,29 +3179,27 @@ def iqr(x, axis=None, rng=(25, 75), scale=1.0, nan_policy='propagate',
...
-                          keepdims=keepdims)
-    out = np.subtract(pct[1], pct[0])
+    pct = stats.quantile(x, rng, axis=-1, method=interpolation, keepdims=True)
+    out = pct[..., 1:2] - pct[..., 0:1]
 
     if scale != 1.0:
         out /= scale
 
-    return out
+    return out[()] if keepdims else np.squeeze(out, axis=axis)[()]
```
mdhaber, line 3202: The `_axis_nan_policy` decorator also takes care of `keepdims`, typically, but it skips lazy backends. I must have written it this way to pass an existing test with JAX. At some point - maybe when we figure out what's happening with Dask - I'll update the decorator to work with JAX, so we'll be able to simplify this.

PR #23796, opened 2025-10-17
TST: still more test tolerance bumps to accommodate Accelerate

mdhaber:
Followup to gh-23788


andyfaff:
fingers crossed that we're finally done. Thank you Matt.

PR #23797, opened 2025-10-17
ENH: sparse: add `copy=False` arg to `astype()` where it could save a copy

### LINKED ISSUE #23778: ENH: Improve performance of converting csc arrays to csr arrays by avoiding copies when indptr and indices are already the correct dtype

Looking at csc_array.tocsr() https://github.com/scipy/scipy/blob/89a2dfbd986b548e8c2ae339684dab597a890de8/scipy/sparse/_csc.py#L44-L65

It
1. Figures out an appropriate dtype
2. Allocates NumPy arrays for the output csr_array
3. Calls the extension function csr_tocsr(), converting self's indptr and indices to the appropriate dtype with astype()

The issue is that by default [ndarray.astype()](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.astype.html) always returns a newly allocated array.

In a specific example with a 15,000 x 15,000 csc array with 100,000 nonzeros I'm seeing 40% of the time being taken up needlessly copying self's indptr and indices.

Call astype() with copy=False, avoiding the copy unless necessary

_No response_

_No response_

dschult:
Fixes #23778 

For a NumPy array, `A.astype(dtype)` makes a copy of `A`. We usually want `A.astype(dtype, copy=False)` even though that is not the default value of the `copy` argument.  Most uses of `astype` already have `copy=False`. A few use the `copy=True` behavior. But most do not. This PR updates the `astype` behavior to save unnecessary copies. 

I believe this kind of change could reduce unnecessary copying of arrays in `sparse.csgraph` and `sparse.linalg` as well, but this PR restricts to the sparse matrix/array structures.


scipy/sparse/_base.py
```diff
@@ -244,8 +244,7 @@ def astype(self, dtype, casting='unsafe', copy=True):
 
         dtype = getdtype(dtype)
         if self.dtype != dtype:
-            return self.tocsr().astype(
-                dtype, casting=casting, copy=copy).asformat(self.format)
+            return self.tocsr().astype(dtype, casting, copy).asformat(self.format)
```
perimosocordiae: I'd prefer to keep the keyword arguments here if possible.

scipy/sparse/_bsr.py
```diff
@@ -331,11 +331,11 @@ def _matmul_sparse(self, other):
         data = np.empty(R*C*bnnz, dtype=upcast(self.dtype,other.dtype))
 
         bsr_matmat(bnnz, M//R, N//C, R, C, n,
-                   self.indptr.astype(idx_dtype),
-                   self.indices.astype(idx_dtype),
+                   self.indptr.astype(idx_dtype, copy=False),
```
perimosocordiae, line 336: Let's assign these to local variables before the maxnnz call above.

scipy/sparse/_bsr.py
```diff
@@ -410,7 +410,7 @@ def tocoo(self, copy=True):
         indptr_diff = np.diff(self.indptr)
         if indptr_diff.dtype.itemsize > np.dtype(np.intp).itemsize:
             # Check for potential overflow
-            indptr_diff_limited = indptr_diff.astype(np.intp)
+            indptr_diff_limited = indptr_diff.astype(np.intp, copy=False)
```
perimosocordiae: This won't have an effect, because we know that the dtype is different in this branch.

scipy/sparse/_lil.py
```diff
@@ -286,7 +286,7 @@ def _mul_scalar(self, other):
             res_dtype = upcast_scalar(self.dtype, other)
 
             new = self.copy()
-            new = new.astype(res_dtype)
+            new = new.astype(res_dtype, copy=False)
```
perimosocordiae: Maybe we could do something fancy here by avoiding the explicit copy when the dtype conversion would copy anyway.

perimosocordiae (changes requested):

scipy/sparse/_lil.py
dschult: Yes -- using `astype(res_dtype)` would work. Then no explicit `self.copy()` is needed.
I could add a comment to make it clear a copy is desired here.  ~~Maybe I'll set the default value `copy=True`
explicitly there for clarity rather than the comment.
`new = new.astype(res_dtype, copy=True)`~~ (comment looked better)

dschult:
I made the recommended changes by @perimosocordiae for the `astype` PR.
~~It has a number of failing tests, but they are all related to a spatial test_rotation failure -- so unrelated.
But I'll try to update once the fix for that is merged.~~ [unrelated failures now fixed in main, so merged here.]

This should be ready to review.

perimosocordiae (approved):

perimosocordiae:
No comments in a month, and changes look good to me. Merged!

PR #23803, opened 2025-10-17
ENH: stats.sigmaclip: add array API support

mdhaber:
#### Reference issue
Toward gh-20544

Adds array API support to `sigmaclip`.

#### Additional information
This one is about as easy as they come. It doesn't support n-d arrays because the size of each slice would change independently each iteration. I will probably take a second pass through these to add `marray` support; and maybe we can support n-d `marray`s at that time.


j-bowhay (approved):

mdhaber:
Thanks @j-bowhay!

PR #23777, opened 2025-10-14
CI: pin Python 3.11.13 in CircleCI job

nickodell:
#### Reference issue

Closes #23642

#### What does this implement/fix?

On 2025-10-15, at 01:24 UTC, CircleCI updated their Python 3.11 image to 3.11.14. Shortly after, our docs job started failing.

The `cimg/python:3.11` image contains a copy of click 8.3.0 located at `/home/circleci/.local/lib/python3.11/site-packages`.

```
(scipy-dev) nodell@scipy-dev:~/scipy$ docker run -it cimg/python:3.11
circleci@43d62fd81758:~/project$ pip show click
Name: click
Version: 8.3.0
Summary: Composable command line interface toolkit
Home-page: 
Author: 
Author-email: 
License-Expression: BSD-3-Clause
Location: /home/circleci/.local/lib/python3.11/site-packages
Requires: 
Required-by: userpath
```

This copy of click is installed by CircleCI when they create the image, as part of installing `pipx`.

For some reason (probably `PYTHONPATH` manipulation) this copy is not uninstalled when we install spin. Spin is not compatible with this version of click, so it breaks the refguide_check and build_docs steps.

Revert this change to make docs builds work again.

#### Additional information

By downgrading to 3.11.13, we would be missing [these security fixes](https://www.python.org/downloads/release/python-31114/).

PR #23805, opened 2025-10-17
MAINT: backports for SciPy 1.16.3

tylerjereddy:
There's a fairly credible request for SciPy `1.16.3` at https://github.com/scipy/scipy/pull/23787#issuecomment-3413210944. That said, let's first test the waters here with a regular flush of the CI to see what kinds of shenanigans we'd have to deal with to make this happen. Also, cc @h-vetinari and @jorenham for planning visibility.

Backports included here (so far):
1. gh-23681
2. gh-23685
3. gh-23777
4. gh-23787
5. gh-23829
6. gh-23821
7. gh-23817
8. gh-23815
9. gh-23864
10. (partial backport only) gh-23811

Backports included prior this PR:
1. gh-23614 (was backported in gh-23615)
2. gh-23629 (was backported in gh-23615)

TODO list:

- [x] assess the situation with regular CI--tractable for a release without too much pain?
- [x] full CI passing, including wheel builds
- [x] update the `1.16.3` release notes/author lists, etc.
- [x] any other backports we need/want to wait a bit for?
- [x] ~should we pin `array-api-strict<2.4.0` in the various dependency specification files?~


tylerjereddy:
The one CI failure with `torch` is gh-23820--since that's just a regex mismatch on an expected error message after the `torch 2.9.0` update this week, I suspect any workaround to get that passing should be fine.

tylerjereddy:
Ok, we're getting closer to a "critical mass" of backports now, we're up to 10 after the latest update. The one labelled "ENH:" is arguably a bug fix for a severe performance regression.

Local testing on ARM Mac seems "ok" (ignoring the usual flakes I get), including array API tests and preview of the new release notes from (local) doc build.

Note that we need `array-api-strict<2.4.0` on the release branch for array API tests to pass, and that was true prior to this release attempt. While that version limitation is already set in CI, it is *not* set in `pyproject.toml` nor in the variety of other files that we can use to pull in/specify deps. I haven't changed that, not sure if it matters a whole lot, but I'll add it to the TODO list above I suppose. I'm somewhat-tempted to avoid touching dep specifications for an optional testing package, though on the other hand it should be harmless/slightly helpful I suppose.

Anyway, let's see if the regular CI is finally green now.

lucascolley:
> I'm somewhat-tempted to avoid touching dep specifications for an optional testing package

I'm perfectly happy with leaving this as-is.

doc/source/release/1.16.3-notes.rst
```diff
@@ -12,12 +12,37 @@ compared to 1.16.2.
...
+    People with a "+" by their names contributed a patch for the first time.
+    This list of names is automatically generated, and may not be fully complete.
 
 
 Issues closed for 1.16.3
 ------------------------
 
+* `#23336 <https://github.com/scipy/scipy/issues/23336>`__: BUG: linalg: wrong handling of args/kwargs in ``_apply_over_batch``
+* `#23469 <https://github.com/scipy/scipy/issues/23469>`__: BUG: huge speed and quality regression for stats.multivariate_normal.cdf
+* `#23642 <https://github.com/scipy/scipy/issues/23642>`__: CI: meson setup fails with "invalid int value: 'Sentinel.UNSET'"
```
jorenham, line 35: I see that xsf has been updated, which means that the fix (https://github.com/scipy/xsf/pull/62) for #23589 is also included:

```suggestion
* `#23589 <https://github.com/scipy/scipy/issues/23589>`__: BUG: ``lqmn(1,1,0)`` and  ``lqmn(1,1,0j)`` return incompatible derivatives
* `#23642 <https://github.com/scipy/scipy/issues/23642>`__: CI: meson setup fails with "invalid int value: 'Sentinel.UNSET'"
```

tylerjereddy: Oh thanks--it didn't get slurped up by the automated release notes list scripts because it was lacking a milestone (I just added the `1.16.3` milestone over there).

That said, this does make me think about something else--I'm pretty sure the author list/name script won't pick you up as an author on the release since your patch was in a different repo.

My initial inclination is to be reasonably inclusive on authorship, maybe I can just add you manually in this case. Is there a feeling that `xsf` is sufficiently independent that folks working on it are happy to get author "credit" over there on its own releases? I don't have a good read on that--I suppose `xsf` has prospects for wide usage separate from SciPy anyway, so maybe?

tylerjereddy:
Of course there are now more CI failures than before. Spending time on pre-release failures is perhaps a bit dubious though I've generally tried to investigate them. I'm guessing the doc build warnings are my fault somehow.

doc/source/release/1.16.3-notes.rst
rgommers: I would say that the primary visibility of authorship is commit counts, and `xsf` has its own (https://github.com/scipy/xsf/graphs/contributors). So I wouldn't do work to treat `xsf` differently from all other libraries we vendor.

jorenham: I was mostly talking about the resolved issue I raised in the scipy repo, actually

tylerjereddy:
For the pre-release CI failure, I backported gh-23864 locally, so that job should fix on next push. For the free-threaded CI errors, I did a diff of a succeeding log for that job (1.16.2 backport PR) vs. the current failing log for that job here (summary of my analysis below the fold).

<details>

- CPython 3.13.7 -> 3.13.9
- Cython 3.2.0a0 -> 3.2.0b1
- NumPy 2.3.3 -> 2.3.4
- spin 0.14 -> 0.15
- hypothesis 6.138.1 -> 6.142.3
- meson 1.9.0 -> 1.9.1

</details>

Of those notable version changes, by far the most suspicious is `Cython 3.2.0a0 -> 3.2.0b1` (released in the last week). I don't think the release branch is the place to be testing Cython alpha/beta release + free threading combination. I suspect this is just a leftover of pulling in pre-release support for free-threading earlier on in the timeline (where there was no stable release option for Cython) and not really intended to be bumped this far. The errors also occur **before** the actual `pytest-run-parallel` test run, so during the "normal" testsuite incantation.

I'm able to reproduce the free-threading errors locally on an ARM Mac with Cython `3.2.0b1` by just running `spin test -j 12`. I can't even compile the maintenance branch with `3.2.0b2`, released 6 hours ago. Cython `3.1.6` seems "ok" locally with the release branch, so I'm going to pin that locally in `linux.yml` with a reference to this comment (`pyproject.toml` already makes the right choice for Cython).

There are still other CI failures to analyze so I won't push up just yet.

tylerjereddy:
For the 32-bit failure, I'll backport the single commit recently merged to deal with that on `main` as noted at https://github.com/scipy/scipy/pull/23811#discussion_r2466995884.

tylerjereddy:
For the CircleCI doc build failure, there are two warnings causing this, and both involve https://www.statsmodels.org/stable/objects.inv not being available for intersphinx. It appears to be available now, so that might be one of the rare cases where "try it again" is actually a sane idea.

I'll make a few more local updates to account for the additional backporting and then push up to see if regular CI turns green.

tylerjereddy:
The "regular" CI is all green now. There are no blocking review comments left. I'm happy with the state of the diff here, so I'll proceed with testing the full wheel build matrix. If that passes I'll assume we're good to go for releasing.

tylerjereddy:
all green, in it goes

tylerjereddy:
Wheel builds started for release proper: https://github.com/scipy/scipy/actions/runs/18863368794

if those also go well, should be a decent chance I can release on Oct. 28th (US time)

tylerjereddy:
@rlucas7 note the "immutability" marker we talked about with the new security setting on releases:

image

It does let me edit the text of the release, but NOT the tag or assets--I suppose that's fine.

PR #23807, opened 2025-10-18
ENH: stats.yeojohnson_llf: add array API support

mdhaber:
#### Reference issue
Toward gh-20544

Adds array API support to `stats.yeojohnson_llf`.

#### Additional information
Also adds `axis`, `nan_policy`, and `keepdims`. It may be easist to review commits separately.

### DIFF
--- a/scipy/stats/_morestats.py
+++ b/scipy/stats/_morestats.py
@@ -1806,9 +1806,9 @@ def yeojohnson_llf(lmb, data, *, axis=0, nan_policy='propagate', keepdims=False)
     # unless necessary.
     kwargs = {}
     if keepdims is not False:
-        kwargs[keepdims] = keepdims
+        kwargs['keepdims'] = keepdims
     if nan_policy != 'propagate':
-        kwargs[nan_policy] = nan_policy
+        kwargs['nan_policy'] = nan_policy
     res = _yeojohnson_llf(data, lmb=lmb, axis=axis, **kwargs)
     return res[()] if res.ndim == 0 else res
 
--- a/scipy/stats/_morestats.py
+++ b/scipy/stats/_morestats.py
@@ -1808,28 +1808,27 @@ def yeojohnson_llf(lmb, data, *, axis=0, nan_policy='propagate', keepdims=False)
         kwargs[keepdims] = keepdims
     if nan_policy != 'propagate':
         kwargs[nan_policy] = nan_policy
-    return _yeojohnson_llf(data, lmb=lmb, axis=axis, keepdims=keepdims,
-                           nan_policy=nan_policy)
+    res = _yeojohnson_llf(data, lmb=lmb, axis=axis, keepdims=keepdims,
+                          nan_policy=nan_policy)
+    return res[()] if res.ndim == 0 else res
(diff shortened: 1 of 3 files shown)


scipy/stats/tests/test_morestats.py
```diff
@@ -2413,24 +2413,29 @@ def test_empty(self):
...
+    def test_2d_input(self, xp):
         x = stats.norm.rvs(size=100, loc=10, random_state=54321)
+        x = xp.asarray(x)
         lmbda = 1
-        llf = stats.yeojohnson_llf(lmbda, x)
-        llf2 = stats.yeojohnson_llf(lmbda, np.vstack([x, x]).T)
-        assert_allclose([llf, llf], llf2, rtol=1e-12)
+        ref = stats.yeojohnson_llf(lmbda, x)
+        res = stats.yeojohnson_llf(lmbda, xp.stack([x, x]).T)
+        assert_allclose(res, ref, rtol=1e-12)
```
mdhaber:
```suggestion
        xp_assert_close(res, xp.stack((ref, ref)), rtol=1e-12)
```

mdhaber:
Looks like CI failures are unrelated. The `boxcox_llf` one is in https://github.com/scipy/scipy/actions/runs/18603167595/job/53046297225, too, for instance.

scipy/stats/_morestats.py
```diff
@@ -1778,24 +1801,33 @@ def yeojohnson_llf(lmb, data):
...
-    data = np.asarray(data)
-    n_samples = data.shape[0]
+    # _axis_nan_policy decorator does not currently support these for lazy arrays.
+    # We want to run tests with lazy backends, so don't pass the arguments explicitly
+    # unless necessary.
+    kwargs = {}
+    if keepdims is not False:
+        kwargs[keepdims] = keepdims
+    if nan_policy != 'propagate':
+        kwargs[nan_policy] = nan_policy
```
j-bowhay: Should the keys not be strings?
```suggestion
        kwargs['keepdims'] = keepdims
    if nan_policy != 'propagate':
        kwargs['nan_policy'] = nan_policy
```

mdhaber:
Good catch, Jake. I'm a little confused as to how it was working when those arguments were not being passed correctly. I should check that out, but hopefully the rest of the PR is ok.

j-bowhay:
> Good catch, Jake. I'm a little confused as to how it was working when those arguments were not being passed correctly. I should check that out, but hopefully the rest of the PR is ok.

It looks like `boxcox_llf` has the same issue too.

mdhaber:
Yeah, I copied it from there.
Yeah I guess we are testing the private functions in `test_axis_nan_policy`, and there aren't tests of `nan_policy` in `test_morestats`, I guess. We can't test the public functions in `test_axis_nan_policy` because they need to accept a `_no_deco` option that skips the decorator. 
This should fix the functionality for `yeojohnson_llf`, but I will fix `boxcox_llf` and the testing of both public functions in an immediate follow-up.

j-bowhay (approved):
The rest looks fine to me!

PR #23815, opened 2025-10-19
ENH: scipy.stats: speed up bivariate normal cdf

### LINKED ISSUE #23469: BUG: huge speed and quality regression for stats.multivariate_normal.cdf
### Describe your issue.

There is a huge speed regression (180x slower) in computing the cdf for the multivariate normal since scipy-1.16.0. The numerical precision is also worse and that lets unit tests in the iminuit package fail that used to pass since the Python-3.9 days.

Tested on MacOS, Linux.

### Reproducing Code Example

```python
import numpy as np
from scipy import stats


def mvnorm(mux, muy, sx, sy, rho):
    C = np.empty((2, 2))
    C[0, 0] = sx**2
    C[0, 1] = C[1, 0] = sx * sy * rho
    C[1, 1] = sy**2
    m = [mux, muy]
    return stats.multivariate_normal(m, C)

def model(xy, mux, muy, sx, sy, rho):
    return mvnorm(mux, muy, sx, sy, rho).cdf(xy.T)

rng = np.random.default_rng(42)
```

```py
%%timeit

model(rng.normal(size=(2, 1000)), 0, 0, 1, 1, 0)
```

v1.15.3: 1.63 ms ± 7.17 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
v1.16.0: 286 ms ± 21.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
v1.16.1: 265 ms ± 5.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

### Error message

None

```shell
1.15.3 2.2.5 sys.version_info(major=3, minor=12, micro=8, releaselevel='final', serial=0)
C:\Users\hansd.Powercat\AppData\Roaming\Python\Python312\site-packages\scipy\__config__.py:154: UserWarning: Install `pyyaml` for better output
{
  "Compilers": {
    "c": {
      "name": "gcc",
      "linker": "ld.bfd",
      "version": "10.3.0",
      "commands": "cc"
    },
    "cython": {
      "name": "cython",
      "linker": "cython",
      "version": "3.0.12",
      "commands": "cython"
    },
    "c++": {
      "name": "gcc",
      "linker": "ld.bfd",
      "version": "10.3.0",
      "commands": "c++"
    },
    "fortran": {
      "name": "gcc",
      "linker": "ld.bfd",
      "version": "10.3.0",
      "commands": "gfortran"
    },
    "pythran": {
      "version": "0.17.0",
      "include directory": "C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pip-build-env-dihuxz0e\\overlay\\Lib\\site-packages/pythran"
    }
  },
  "Machine Information": {
    "host": {
      "cpu": "x86_64",
      "family": "x86_64",
      "endian": "little",
      "system": "windows"
    },
    "build": {
      "cpu": "x86_64",
      "family": "x86_64",
      "endian": "little",
      "system": "windows"
    },
    "cross-compiled": false
  },
  "Build Dependencies": {
    "blas": {
      "name": "scipy-openblas",
      "found": true,
      "version": "0.3.28",
      "include directory": "C:/Users/runneradmin/AppData/Local/Temp/cibw-run-3dk08vlg/cp312-win_amd64/build/venv/Lib/site-packages/scipy_openblas32/include",
      "lib directory": "C:/Users/runneradmin/AppData/Local/Temp/cibw-run-3dk08vlg/cp312-win_amd64/build/venv/Lib/site-packages/scipy_openblas32/lib",
      "openblas configuration": "OpenBLAS 0.3.28 DYNAMIC_ARCH NO_AFFINITY Haswell MAX_THREADS=24",
      "pc file directory": "D:/a/scipy/scipy"
    },
    "lapack": {
      "name": "scipy-openblas",
      "found": true,
      "version": "0.3.28",
      "include directory": "C:/Users/runneradmin/AppData/Local/Temp/cibw-run-3dk08vlg/cp312-win_amd64/build/venv/Lib/site-packages/scipy_openblas32/include",
      "lib directory": "C:/Users/runneradmin/AppData/Local/Temp/cibw-run-3dk08vlg/cp312-win_amd64/build/venv/Lib/site-packages/scipy_openblas32/lib",
      "openblas configuration": "OpenBLAS 0.3.28 DYNAMIC_ARCH NO_AFFINITY Haswell MAX_THREADS=24",
      "pc file directory": "D:/a/scipy/scipy"
    },
    "pybind11": {
      "name": "pybind11",
      "version": "2.13.6",
    }
  },
  "Python Information": {
    "path": "C:\\Users\\runneradmin\\AppData\\Local\\Temp\\cibw-run-3dk08vlg\\cp312-win_amd64\\build\\venv\\Scripts\\python.exe",
    "version": "3.12"
  }
}
```

mdhaber:
#### Reference issue
Closes gh-23469

This PR adds a Python (/Pythran) translation of Alan Genz' Matlab bivariate normal cdf code [`bvn`](https://web.archive.org/web/20200205121851/http://www.math.wsu.edu/faculty/genz/software/matlab/bvn.m) to our `multivariate_normal` implementation.

Using the example from gh-23469 as a benchmark, we get an ~70x speed improvement.
```
# Before: 423 ms ± 1.87 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# After: 6.04 ms ± 24 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
I suspect the part of the difference between that and the 180x difference reported in the issue is that Pythran isn't using BLAS?

Accuracy was checked against SciPy 1.15 multivariate normal with this [script](https://github.com/scipy/scipy/issues/23469#issuecomment-3345238660). I confirmed that the script follows all branches except for the special cases at the top, but I confirmed separately that those execute as expected.

#### Additional information
LMK how I should include the license information or if the text at the top of `_qmvnt.py` is enough.
As a unit test, I only added one case to `test_cdf_vs_cubature` - I'll add what the reviewer deems appropriate.
@HDembinski please consider testing this branch.

### DIFF
--- a/scipy/meson.build
+++ b/scipy/meson.build
@@ -597,6 +597,7 @@ cpp_args_pythran += [
   _cpp_Wno_unused_function,
   _cpp_Wno_unused_variable,
   _cpp_Wno_int_in_bool_context,
+  _cpp_Wno_unused_local_typedefs,
 ]
 
 # Fortran warning flags
--- a/scipy/meson.build
+++ b/scipy/meson.build
@@ -597,7 +597,6 @@ cpp_args_pythran += [
   _cpp_Wno_unused_function,
   _cpp_Wno_unused_variable,
   _cpp_Wno_int_in_bool_context,
-  _cpp_Wno_unused_local_typedefs,
 ]
 
 # Fortran warning flags
--- a/scipy/stats/_stats_pythran.py
+++ b/scipy/stats/_stats_pythran.py
@@ -224,6 +224,8 @@ def np_dot(x, y):
 # function p = bvnu( dh, dk, r )
 #pythran export _bvnu(float64, float64, float64)
 def _bvnu(dh, dk, r):
+    # Adapted from the MATLAB original implementation by Dr. Alan Genz;
+    # see license information in _qmvnt.py
     math_inf, math_pi = np.inf, np.pi
     # if dh ==  inf | dk ==  inf:p = 0;
     if (dh == math_inf) or (dk == math_inf):
--- a/scipy/stats/_stats_pythran.py
+++ b/scipy/stats/_stats_pythran.py
@@ -1,3 +1,4 @@
+import math
 import numpy as np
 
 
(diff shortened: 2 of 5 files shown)


mdhaber:
@serge-sans-paille does this like a warning I can avoid by using `concatenate` differently?
```
In file included from scipy/stats/_stats_pythran.cpython-312-x86_64-linux-gnu.so.p/_stats_pythran.cpp:94:
../../../../../../opt/hostedtoolcache/Python/3.12.12/x64/lib/python3.12/site-packages/pythran/pythonic/numpy/concatenate.hpp:144:11: error: unused type alias 'T' [-Werror,-Wunused-local-typedef]
  144 |     using T =
      |           ^
../../../../../../opt/hostedtoolcache/Python/3.12.12/x64/lib/python3.12/site-packages/pythran/pythonic/numpy/concatenate.hpp:165:11: error: unused type alias 'T' [-Werror,-Wunused-local-typedef]
  165 |     using T = typename E::dtype;
      |           ^
../../../../../../opt/hostedtoolcache/Python/3.12.12/x64/lib/python3.12/site-packages/pythran/pythonic/numpy/concatenate.hpp:183:11: error: unused type alias 'T' [-Werror,-Wunused-local-typedef]
  183 |     using T = typename return_type::dtype;
      |           ^
3 errors generated.
```

rkern:
I'd probably at least have a comment in `_stats_pythran.py` pointing to the license information in `_qmvnt.py`.

To finalize this before merging, I'd probably also delete the MATLAB comments. They were good for review, though, thank you! It looks good to my eye, in terms of faithfully translating the code, but I haven't built or tested it.

serge-sans-paille:
On Sun, Oct 19, 2025 at 11:03:52AM -0700, Matt Haberland wrote:
> ●mdhaber left a comment [1](scipy/scipy#23815)
> 
> ***@***.*** does this like a warning I can avoid in how I use
> concatenate?

Nope, it's an unused local type in Pythran, you can safely ignore it and I'll
submit a patch to fix it.

scipy/stats/_stats_pythran.py
```diff
@@ -209,3 +210,144 @@ def _poisson_binom(k, args, tp):
...
+        h = dh
+        k = dk
+        hk = h*k
+        bvn = 0.
+        # if abs(r) < 0.3      % Gauss Legendre points and weights, n =  6
+        #     w(1:3) = [0.1713244923791705 0.3607615730481384 0.4679139345726904];
+        #     x(1:3) = [0.9324695142031522 0.6612093864662647 0.2386191860831970];
+        if abs(r) < 0.3:
+            w = [0.1713244923791705, 0.3607615730481384, 0.4679139345726904]
+            x = [0.9324695142031522, 0.6612093864662647, 0.2386191860831970]
```
ev-br, line 257: I still think it'd make sense to use `roots_legenendre`,  https://github.com/scipy/scipy/issues/23469#issuecomment-3353487433

I believe the whole extent of the change is 
https://github.com/scipy/scipy/compare/main...ev-br:scipy:mvn_2D?expand=1#diff-b5afc2abf94ff5b2f535a0f4d6a2f09b5838effa4f4686913c61637b624abde0R617-R624

This will also remove `np.concatenate` and its related build warning.

ev-br (approved):
LGTM modulo one suggestion.

scipy/stats/_stats_pythran.py
mdhaber: I don't think `roots_legendre` is available from `pythran`, and I don't want to call it in Python and pass the arrays into Pythran.

serge-sans-paille:
> I don't think roots_legendre is available from pythran.

Correct. Would that make sense for me to add it?

mdhaber:
I don't think so, personally. I think it's fine - preferably, IMO - for this to be a line by line translation of the Matlab code in this PR. ISTM that is the path of least resistance to getting something merged into 1.17.0. But I guess I'm at the mercy of the reviewer, so your call Evgeni.

ev-br:
FWIW, I too don't think pythran needs to know about `roots_legendre`. It's just a way to hardcode the same numbers, only reducing code duplication a little bit. A no-brainer for me, but if it's inconvenient for whatever reason, it's not a blocker, can be done in a follow-up. 

> and I'll submit a patch to fix it.

That'd be great Serge! 

I typed "not a blocker for this PR" and started reaching for the green button, but then went to double-check the CI failures, and... we actually have a build with `-Werror`, which fails the CI at https://github.com/scipy/scipy/actions/runs/18632095961/job/53118247185?pr=23815. So, how hard is it to fix it?

serge-sans-paille:
The fix is in https://github.com/serge-sans-paille/pythran/pull/2365
I can make a pythran patch release if needed, or you can just disable that warning until the next release. What do you prefer?

ev-br:
I think we can add a suppression to https://github.com/scipy/scipy/blob/main/scipy/meson.build#L593

scipy/meson.build
```diff
@@ -597,6 +597,7 @@ cpp_args_pythran += [
   _cpp_Wno_unused_function,
   _cpp_Wno_unused_variable,
   _cpp_Wno_int_in_bool_context,
+  _cpp_Wno_unused_local_typedefs,
```
rgommers: This is a bit of a blunt instrument, I'd rather silence this only for the extension module it's needed in, since that's also a one-liner:

```diff
diff --git a/scipy/stats/meson.build b/scipy/stats/meson.build
index e7bbc86cd8..ea07c252bf 100644
--- a/scipy/stats/meson.build
+++ b/scipy/stats/meson.build
@@ -77,7 +77,7 @@ py3.extension_module('_biasedurn',
 if use_pythran
   py3.extension_module('_stats_pythran',
     pythran_gen.process('_stats_pythran.py'),
-    cpp_args: cpp_args_pythran,
+    cpp_args: [cpp_args_pythran, _cpp_Wno_unused_local_typedefs],
     dependencies: [pythran_dep, np_dep],
     link_args: version_link_args,
     install: true,
```

That will allow us to detect future issues more easily. Also, can you add a comment to that change that the flag can be removed again once we require `pythran >= 0.19.0`?

rgommers (commented):
Thanks for working on this! I just had a look at the build system change and made one suggestion. Glanced over the rest and that all looks good to me.

scipy/stats/_stats_pythran.py
dschmitz89: How about we add a simple comment that shows how to get the numbers using `roots_legendre`?

scipy/meson.build
mdhaber: Done, thanks.

scipy/stats/_stats_pythran.py
mdhaber: Feel free.

dschmitz89:
I added the backport label here as this fixes a severe performance regression that was also noticed by matplotlib for example. WDYT @tylerjereddy for 1.16.3 ?

tylerjereddy:
The `ENH` label gives me some pause, but maybe that's just a "branding issue" if we're fixing a performance regression severe enough to basically be considered a sort of "bug" perhaps.

I'll give it a shot for backporting I guess.

mdhaber:
> if we're fixing a performance regression severe enough to basically be considered a sort of "bug" perhaps.

Yeah, I was thinking of it as a performance enhancement... over a severe performance regression that was reported as a bug. So it can also be considered to be a fix.

HDembinski:
IMHO, a performance regression is a bug, because it breaks the suitability of the library for certain tasks. In this case, it was severe, too, not only slower, but apparently less numerically accurate, too.

PR #23818, opened 2025-10-19
DOC: slsqp parameters-->options [docs only]

andyfaff:
SLSQP minimize documentation doesn't look like other methods. Most of the methods have a "see also" box, then options. SLSQP jumps straight to a parameters section. This PR is trying to make SLSQP look like the rest.

### DIFF
--- a/scipy/optimize/_slsqp_py.py
+++ b/scipy/optimize/_slsqp_py.py
@@ -225,8 +225,8 @@ def _minimize_slsqp(func, x0, args=(), jac=None, bounds=None,
     Minimize a scalar function of one or more variables using Sequential
     Least Squares Programming (SLSQP).
 
-    Parameters
-    ----------
+    Options
+    -------
     ftol : float
         Precision target for the value of f in the stopping criterion. This value
         controls the final accuracy for checking various optimality conditions;


andyfaff:
SLSQP looks like the other minimize methods now.

## After:
Screenshot 2025-10-20 at 10 50 52

## Before: 
Screenshot 2025-10-20 at 11 03 53

mdhaber:
I'll merge it if you want, but since I reviewed the status of things, I should mention:
- Of the methods that have an "Options" section after "See Also", SLSQP is the only one with "Returns", which appears above everything else.
- `trust-constr` is still like `slsqp` was before this PR, with a "Parameters" and "Returns" and without "See Also".
- `powell` has all the "Parameters" and "Returns" of `minimize` followed by "Options".
- I've never understood why `——-` needs to appear at the beginning of the "Options" sections. (I don't expect that to be fixed here... that's infrastructure stuff, I imagine. I know this is all generated by a custom script.)

Did you want to make things consistent in this PR, or make just this change?

andyfaff:
Drat, didn't look at trust-constr. It's probably just better to leave as-is. I'll be biting off too much work to normalise everything.

PR #23811, opened 2025-10-19
TST: whack-a-mole

mdhaber:
#### What does this implement/fix?
Fix/skip tests that have been failing recently.

#### Additional information
See recent [Linux test logs](https://github.com/scipy/scipy/actions/workflows/linux.yml) for examples.

PR #23821, opened 2025-10-20
CI: optimize: allow new Pytorch message

nickodell:
Pytorch 2.9.0 changes the error message on failed broadcasts. Allow the new style to pass tests.

#### Reference issue

Closes #23820

In pytorch/pytorch#160251, broadcast_shapes() was changed to call broadcasting logic in another part of the package. As a consequence, when we upgrade from Pytorch 2.9.0 to Pytorch 2.8.0, the error message changes here from

```
RuntimeError: Shape mismatch: objects cannot be broadcast to a single shape
```

to

```
RuntimeError: Attempting to broadcast a dimension of length 3 at -1! Mismatching argument at index 1 had torch.Size([3]); but expected shape should be broadcastable to [2]
```

Therefore, add a regex that allows either message.

#### Additional information

N/A

### DIFF
--- a/scipy/optimize/tests/test_chandrupatla.py
+++ b/scipy/optimize/tests/test_chandrupatla.py
@@ -850,7 +850,8 @@ class TestFindRoot:
             find_root(func, bracket)
 
         # raised by `np.broadcast, but the traceback is readable IMO
-        message = "...not be broadcast..."  # all messages include this part
+        # all messages include this part
+        message = "(not be broadcast|Attempting to broadcast a dimension of length)"
         with pytest.raises((ValueError, RuntimeError), match=message):
             bracket = xp.asarray([-2, -3]), xp.asarray([3, 4, 5])
             find_root(func, bracket)


nickodell:
cc @tylerjereddy

mdhaber (approved):
LGTM! I'll let CI finish, though.

mdhaber:
CI failure is just one that would be fixed by gh-23811.
@tylerjereddy does this do the trick? I don't think we're seeing gh-23820 in main yet, and I can't check locally ATM.

tylerjereddy (approved):
I confirmed that on x86_64 Linux with `torch` `2.9.0` installed the latest SciPy `main` branch has the single test failure reported in the matching issue via incantation `spin test -j 16 -b all`.

On this branch, the same incantation in the same environment has no remaining test failures. The CI is also passing for CPU/GPU array API tests. On SciPy `main` `pixi` lock files are avoiding the problem scenario for now.

So, I'll merge this, it will be helpful for SciPy `1.16.3` as well. Thanks both.

PR #23822, opened 2025-10-20
ENH: stats.kruskal: vectorize implementation

mdhaber:
#### Reference issue
Toward gh-14651
Toward gh-20544

This vectorizes the `scipy.stats.kruskal` implementation in preparation for translation to the array API.

### DIFF
--- a/scipy/stats/_stats_py.py
+++ b/scipy/stats/_stats_py.py
@@ -8432,7 +8432,7 @@ KruskalResult = namedtuple('KruskalResult', ('statistic', 'pvalue'))
 
 @xp_capabilities(np_only=True)
 @_axis_nan_policy_factory(KruskalResult, n_samples=None)
-def kruskal(*samples, nan_policy='propagate'):
+def kruskal(*samples, nan_policy='propagate', axis=0):
     """Compute the Kruskal-Wallis H-test for independent samples.
 
     The Kruskal-Wallis H-test tests the null hypothesis that the population
@@ -8454,6 +8454,11 @@ def kruskal(*samples, nan_policy='propagate'):
           * 'propagate': returns nan
           * 'raise': throws an error
           * 'omit': performs the calculations ignoring nan values
+    axis : int or tuple of ints, default: None
+        If an int or tuple of ints, the axis or axes of the input along which
+        to compute the statistic. The statistic of each axis-slice (e.g. row)
+        of the input will appear in a corresponding element of the output.
+        If ``None``, the input will be raveled before computing the statistic.
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


scipy/stats/_stats_py.py
```diff
@@ -8498,27 +8503,28 @@ def kruskal(*samples, nan_policy='propagate'):
     KruskalResult(statistic=7.0, pvalue=0.0301973834223185)
 
     """
-    samples = list(map(np.asarray, samples))
```
mdhaber, line 8501: Taken care of by `_axis_nan_policy` decorator.

scipy/stats/_stats_py.py
```diff
@@ -8498,27 +8503,28 @@ def kruskal(*samples, nan_policy='propagate'):
...
-
     num_groups = len(samples)
     if num_groups < 2:
         raise ValueError("Need at least two groups in stats.kruskal()")
 
-    n = np.asarray(list(map(len, samples)))
+    n = np.asarray([sample.shape[-1] for sample in samples])
```
mdhaber: `_axis_nan_policy` also ensures `axis=-1`.

scipy/stats/_stats_py.py
```diff
@@ -8498,27 +8503,28 @@ def kruskal(*samples, nan_policy='propagate'):
...
+    totaln = np.sum(n)
```
mdhaber: `n` is a 1-D array, so I left `axis` argument out to emphasize that `totaln` will be a scalar. Probably should go ahead and cast to `float` now; this will avoid possible mixed float-int issues with array API strict later.
```suggestion
    totaln = float(np.sum(n))
```
TBH, I'll probably leave `n` as a Python list and use `sum` here.

*Yup, this was there to avoid the `test_large_no_samples` failure.*

scipy/stats/_stats_py.py
```diff
@@ -8498,27 +8503,28 @@ def kruskal(*samples, nan_policy='propagate'):
...
 
-    alldata = np.concatenate(samples)
-    ranked = rankdata(alldata)
-    ties = tiecorrect(ranked)
-    if ties == 0:
-        raise ValueError('All numbers are identical in kruskal')
+    if np.any(n) < 1:  # Only needed for `test_axis_nan_policy`
+        raise ValueError("Inputs must not be empty.")
```
mdhaber, line 8514: The `_axis_nan_policy` tests just need this to compare the behavior of the wrapped and unwrapped function. The user will never see this.
```
> stats.kruskal([], [1], axis=-1)
<ipython-input-8-f34d1b36fc2d>:1: SmallSampleWarning: One or more sample arguments is too small; all returned values will be NaN. See documentation for sample size requirements.
  stats.kruskal([], [1], axis=-1)
Out: KruskalResult(statistic=np.float64(nan), pvalue=np.float64(nan))
```

scipy/stats/_stats_py.py
```diff
@@ -8498,27 +8503,28 @@ def kruskal(*samples, nan_policy='propagate'):
...
+
+    alldata = np.concatenate(samples, axis=-1)
+    ranked, t = _rankdata(alldata, method='average', return_ties=True)
+
+    # ties = tiecorrect(ranked)
+    ties = 1 - (t**3 - t).sum(axis=-1) / (totaln**3 - totaln)
```
mdhaber: Compare against
https://github.com/scipy/scipy/blob/de80faf9d3480b9dbb9b888568b64499e0e70c19/scipy/stats/_stats_py.py#L8097-L8102
Note that if `size < 2`, the `_axis_nan_policy` decorator is going to produce NaNs for us, so the special case is not needed.

scipy/stats/_stats_py.py
```diff
@@ -8498,27 +8503,28 @@ def kruskal(*samples, nan_policy='propagate'):
...
 
     # Compute sum^2/n for each group and sum
-    j = np.insert(np.cumsum(n), 0, 0)
```
mdhaber, line 8516: ~~Implementation below avoids inserting a `0` at the begining of `cumsum` because `insert` is not available in the standard and `concatenate` would require creating an array of zeros of the right shape. I suppose the `include_initial` argument of `cumulative_sum` would have worked - but `ssbn` needs to be initialized to an array of the right shape anyway, so might as well refactor as below.~~ Not thinking straight - this is 1d. I will refactor this back.

mdhaber (commented):
Some self-review to aid the reviewer.

scipy/stats/_stats_py.py
```diff
@@ -8498,27 +8503,28 @@ def kruskal(*samples, nan_policy='propagate'):
...
 
-    n = np.asarray(list(map(len, samples)))
+    n = np.asarray([sample.shape[-1] for sample in samples])
+    totaln = np.sum(n)
 
-    alldata = np.concatenate(samples)
-    ranked = rankdata(alldata)
-    ties = tiecorrect(ranked)
-    if ties == 0:
-        raise ValueError('All numbers are identical in kruskal')
```
mdhaber, line 8513: Raising an error depending on values is not going to be good for a vectorized, array API implementation because it won't work for lazy arrays, and it would prevent the function from running if the nonunique values are located in a single slice of a multidimensional array.  This wasn't tested or documented, and without it we just get NaNs in offending slices - an appropriate result. If we're really concerned by the (totally valid) warning with division by zero below, we can use `apply_where`, but i don't think it's necessary, so I haven't done that.

scipy/stats/_stats_py.py
```diff
@@ -8498,27 +8504,24 @@ def kruskal(*samples, nan_policy='propagate'):
...
     """
-    samples = list(map(np.asarray, samples))
-
     num_groups = len(samples)
     if num_groups < 2:
         raise ValueError("Need at least two groups in stats.kruskal()")
 
-    n = np.asarray(list(map(len, samples)))
+    n = [sample.shape[-1] for sample in samples]
+    totaln = sum(n)
```
mdhaber, line 8512: Moved up from much later since it's needed by tie correction. No need to cast as float64 - leave it an integer to avoid overflow.

j-bowhay (approved):

mdhaber:
Thanks for taking a look at this and Yeo-Johnson, Jake!

scipy/stats/_stats_py.py
```diff
@@ -8454,6 +8455,11 @@ def kruskal(*samples, nan_policy='propagate'):
           * 'propagate': returns nan
           * 'raise': throws an error
           * 'omit': performs the calculations ignoring nan values
+    axis : int or tuple of ints, default: None
```
j-bowhay, line 8458: And here

PR #23823, opened 2025-10-20
MAINT: interpolate: mark legacy functions out of scope for Array API

ev-br:
Mark legacy functions as out-of-scope for the Array API, so that the [% listings](https://scipy.github.io/devdocs/dev/api-dev/array_api.html) are less misleading---the tables are filled from `@xp_capabilities` decorated functions themselves. 

Why? Marking a function as _legacy_ basically signals that we no longer want to spend effort improving it. Array API is one way of improving it we are not interested in.

### DIFF
--- a/scipy/interpolate/_fitpack_impl.py
+++ b/scipy/interpolate/_fitpack_impl.py
@@ -34,7 +34,7 @@ from numpy import (atleast_1d, array, ones, zeros, sqrt, ravel, transpose,
 #  f2py-generated version
 from . import _dfitpack as dfitpack
 
-from scipy._lib._array_api import array_namespace, concat_1d
+from scipy._lib._array_api import array_namespace, concat_1d, xp_capabilities
 
 
 dfitpack_int = dfitpack.types.intvar.dtype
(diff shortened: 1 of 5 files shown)


scipy/interpolate/_fitpack_impl.py
```diff
@@ -651,7 +653,7 @@ def bisplev(x, y, tck, dx=0, dy=0):
 
     Notes
     -----
-        See `bisplrep` to generate the `tck` representation.
+    See `bisplrep` to generate the `tck` representation.
```
ev-br, line 656: This change is to prevent inconsistent indentation with the xp-capabilities injected Note.

scipy/interpolate/_fitpack_py.py
```diff
@@ -770,7 +779,7 @@ def splder(tck, n=1):
 
     Parameters
     ----------
-    tck : BSpline instance or tuple
+    tck : BSpline instance or tuple@xp_capabilities(out_of_scope=True)
```
ev-br:
```suggestion
    tck : BSpline instance or tuple
```

j-bowhay (approved):
Thanks, agree that we should not convert these

PR #23824, opened 2025-10-20
DOC: ``special.mathieu_odd_coef``: "even" -> "odd"

jorenham:
Even I thought it was odd to call odd even.


ev-br:
Thanks @jorenham , it was indeed not a little odd :-)

PR #23828, opened 2025-10-20
ENH: spatial.transform.Rotation Add axis argument to `Rotation.mean`

amacati:
Implements `axis` as an argument as discussed in the tracker with @scottshambaugh (see https://github.com/scipy/scipy/issues/22500#issuecomment-3358414549). 

#### Reference issue
Contributes towards  #23391

Adds the axis argument to `Rotation.mean`.

#### Additional information
I'm not super happy with the test coverage we have with the addition of another keyword. There are a lot of combinations, and we only test a few select ones.


scipy/spatial/transform/_rotation_cy.pyx
```diff
@@ -945,9 +945,27 @@ def approx_equal(double[:, :] quat, double[:, :] other, atol = None, bint degree
...
+    if axis == ():
+        return quat
+
+    if axis is None:
+        axis = (0,)
+    if isinstance(axis, int):
+        axis = (axis,)
+    if not isinstance(axis, tuple):  # Must be tuple by now
+        raise ValueError("`axis` must be None, int, or tuple of ints.")
+    axis = tuple(set(x if x >= 0 else x + (quat.ndim - 1) for x in axis))
```
lucascolley: could use `x % (quat.ndim - 1)` here

lucascolley (approved):
LGTM, thanks Martin!

Want to take a look @scottshambaugh ?

scipy/spatial/transform/_rotation_cy.pyx
mschuck-rai: I initially thought about using that, but it results in out-of-bounds axes being casted to valid axes, which is not what we want. E.g. for `quat.ndim = 4` and `axis=(5,)`, we want to raise an error. The modulo operation would turn this into `(2,)`, which does not raise an error.

amacati: _Note to self: Check current account to avoid confusion_

lucascolley: Ah yeah fair, I was thinking about a case in which we perform the bounds check first: https://github.com/data-apis/array-api-extra/blob/main/src/array_api_extra/_delegation.py#L184-L189

amacati: I like this more. We should change the xp and cy implementations to the xpx version.

scipy/spatial/transform/_rotation_xp.py
```diff
@@ -493,18 +496,38 @@ def mean(quat: Array, weights: ArrayLike | None = None) -> Array:
...
-        weights_flat = xp.reshape(weights, (-1,))
-        K = xp.matrix_transpose(weights_flat[..., None] * quat) @ quat
-
+        weighted_quat = weights[..., None, None] * quat_expand
+        K = xp.matrix_transpose(weighted_quat) @ quat_expand
+
+    # Axis logic: For None, we reduce over all axes. For int, we only reduce over that
+    # axis. For tuple, we reduce over all specified axes.
+    if axis is None:
+        axis = tuple(range(quat.ndim - 1))
```
scottshambaugh: `tuple(range(quat.ndim - 1))` gets called 3 times, could you pull it out to a variable?

scipy/spatial/transform/_rotation_xp.py
```diff
@@ -493,18 +496,38 @@ def mean(quat: Array, weights: ArrayLike | None = None) -> Array:
...
-        quat = xp.reshape(quat, (-1, 4))
-        weights_flat = xp.reshape(weights, (-1,))
-        K = xp.matrix_transpose(weights_flat[..., None] * quat) @ quat
-
+        weighted_quat = weights[..., None, None] * quat_expand
+        K = xp.matrix_transpose(weighted_quat) @ quat_expand
+
+    # Axis logic: For None, we reduce over all axes. For int, we only reduce over that
+    # axis. For tuple, we reduce over all specified axes.
+    if axis is None:
```
scottshambaugh: My personal preference is to do all input checking before other calculations are performed, and to move this to the top of the function. But that's just stylistic, so can take or leave it.

scipy/spatial/transform/tests/test_rotation.py
```diff
@@ -1302,6 +1302,28 @@ def test_mean(xp, ndim: int):
         xp_assert_close(r.mean().magnitude(), desired, atol=atol)
 
 
+@make_xp_test_case(Rotation.from_rotvec, Rotation.mean, Rotation.magnitude)
+@pytest.mark.parametrize("ndim", range(1, 5))
+def test_mean_axis(xp, ndim: int):
```
scottshambaugh, line 1326: The expected mean being zero in all these cases is a useful check, but doesn't fully exercise things IMO. Could you add a test with random rotations, that compares a reduction along several axes to a list of axis=None mean of rotations formed from those slices? That will prove out the equivalence to the existing algorithm.

scottshambaugh (commented):
A couple suggestions, please excuse the delay in reviewing!

scipy/spatial/transform/_rotation_xp.py
amacati: Makes sense. I moved the input checking as far up as possible.

scipy/spatial/transform/tests/test_rotation.py
amacati: Added a dedicated test for it. Let me know if you had something else in mind!

scipy/spatial/transform/_rotation_xp.py
```diff
@@ -493,18 +513,23 @@ def mean(quat: Array, weights: ArrayLike | None = None) -> Array:
...
-        quat = xpx.atleast_nd(quat, ndim=2, xp=xp)
-        quat = xp.reshape(quat, (-1, 4))
-        weights_flat = xp.reshape(weights, (-1,))
-        K = xp.matrix_transpose(weights_flat[..., None] * quat) @ quat
-
+        weighted_quat = weights[..., None, None] * quat_expand
+        K = xp.matrix_transpose(weighted_quat) @ quat_expand
+
+    # Move reduction axes to the end
+    keep_axes = tuple(i for i in range(quat.ndim - 1) if i not in axis)
```
scottshambaugh: This can be all_axes too.

scottshambaugh (approved):
New test looks great! Good to go overall.

amacati:
Alright, fixed the last comment. The CI will still fail because of #23869, but the failures are unrelated to this PR and will be fixed in #23874. That is of course, if no new CI failures pop up.

amacati:
CI looks good

PR #23829, opened 2025-10-20
MAINT: patch vendored qhull source for security fix

rlucas7:
#### Reference issue
No open issue-discussed on maintainers forum.

Updates the qhull c library to incorporate changes from a month or two ago, 

#### Additional information
If this is merged relatively easily we might consider adding it to the backports in 
 #-23805 for bugfix version 1.16.3 and subsequent releases.

This passes for me locally


lucascolley (changes requested):
We may not want to do this for unreleased versions of qhull, as the vendored source will then differ from https://github.com/scipy/scipy/blob/main/meson.build#L217

rgommers:
Thanks for digging in to this @rlucas7. This indeed won't work in its current form as @lucascolley said. We only want the minimal patch that applies the upstream fix for the `printf` complaint in https://github.com/scipy/scipy/security/code-scanning/2, as discussed on the core team mailing list.

rlucas7:
> We may not want to do this for unreleased versions of qhull, as the vendored source will then differ from https://github.com/scipy/scipy/blob/main/meson.build#L217

Ah that's where it is, so what's difficult about this is C language things don't have a central package/library distribution mechanism like pypi for python (and most newer programming languages). I was going off of this:

https://github.com/qhull/qhull/releases/tag/v8.1-alpha6

I could change the string in line 217 to reflect this, would that be acceptable? 

FWIW These changes seem to be incremental work towards a 8.1.0 release of qhull and AFAICS should be backwards compatible with the 8.0.2 version and would result is fewer changes for an upgrade to 8.1.0 when that becomes available. 

Also, the sole test failure here is a memory allocation of a matrix that seems unrelated. 

If not, I can put some code into the shell script to pull out the hunk for the args identified by the vuln as the *only change* applied.

rlucas7:
> Thanks for digging in to this @rlucas7. This indeed won't work in its current form as @lucascolley said. We only want the minimal patch that applies the upstream fix for the `printf` complaint in [Too few arguments to formatting function](/scipy/scipy/security/code-scanning/2), as discussed on the core team mailing list.

thanks @rgommers can you (also) take a look at my reply to @lucascolley 's comment and let me know which is preferred? 

I hear the emphasis on *minimal* from your previous reply but am unclear on the release status of the `v8.1-alpha6` version, is the reluctance stemming from the `-alpha6` suffix?

Happy to proceed down whichever path we think is preferrable here to close the issue.

rgommers:
> I hear the emphasis on _minimal_ from your previous reply but am unclear on the release status of the `v8.1-alpha6` version, is the reluctance stemming from the `-alpha6` suffix?

Yes. Not that tag in particular, but essentially we can/should only use proper releases. That has been the case since we introduced the [`-Duse-system-libraries`](https://scipy.github.io/devdocs/building/redistributable_binaries.html#using-system-dependencies-instead-of-vendored-sources) build option, which is a distro packager focused feature. Using some random intermediate commit that we vendor makes it impossible to find any matching version in any distro packaging system. It may silently introduce differences between vendored and system versions of what nominally has the same version number, which is undesirable. So we do what distros also do: carry the minimal number of patches (for build/security or important fixes only) on top of an unchanged vendored copy of an official release.

lucascolley:
So we want to cherry-pick the fix on top of the existing source commit, generate a patch for it, and apply that patch in the vendor script.

rgommers:
> generate a patch for it

I suggested the directory to put that in in my email. A `patches/` directory is similar to, e.g., https://github.com/conda-forge/numpy-feedstock/tree/main/recipe/patches

rlucas7:
> > generate a patch for it
> 
> I suggested the directory to put that in in my email. A `patches/` directory is similar to, e.g., https://github.com/conda-forge/numpy-feedstock/tree/main/recipe/patches

thanks @rgommers I think this does what we want with the latest [d8a57d3](https://github.com/scipy/scipy/pull/23829/commits/d8a57d346282fbb6c265da43aaf426a4d617d334), once we've got this PR in a good state or merged, I can look into a patch for the [conda-forge scipy](https://github.com/conda-forge/scipy-feedstock/tree/main) as @lucascolley had suggested on the other thread. 

The numpy feedstock link will be helpful for cribbing-so thanks for that link too.

lucascolley:
> I can look into a patch for the conda-forge scipy as @lucascolley had suggested on the other thread.

Perhaps, but I was actually suggesting a patch to the qhull feedstock

lucascolley (approved):
I pushed a commit to re-vendor qhull and the patch is indeed applied successfully. LGTM, thanks @rlucas7 !

rgommers (approved):
LGTM too now, thanks @rlucas7 and @lucascolley!

Issue #15943, opened 2022-04-06
scipy.stats.theilslopes fails on small data


rarnaout:
from scipy.stats import theilslopes as ts
ts([0, 1], [0, 1])
"Works fine"
ts([0, 0, 0], [0, 1, 0])
"""Fails:
.pyenv/versions/3.9.2/lib/python3.9/site-packages/numpy/core/_methods.py:189: RuntimeWarning: invalid value encountered in double_scalars
  ret = ret.dtype.type(ret / rcount)
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Input In [9], in <cell line: 1>()
----> 1 ts([0, 1], [0, 0])

File ~/.pyenv/versions/3.9.2/lib/python3.9/site-packages/scipy/stats/_stats_mstats_common.py:344, in theilslopes(y, x, alpha, method)
    342 Ru = min(int(np.round((nt - z*sigma)/2.)), len(slopes)-1)
    343 Rl = max(int(np.round((nt + z*sigma)/2.)) - 1, 0)
--> 344 delta = slopes[[Rl, Ru]]
    345 return medslope, medinter, delta[0], delta[1]

IndexError: index 0 is out of bounds for axis 0 with size 0
"""
The problem appears to be, if you go back to Sen (1968), equation 2.6, which theilslopes uses, is an approximation for large n, but when you use small n like above, sigma (above) can be the sqrt() of a negative value, which is nan, which causes the above. 

I know that when n is small, one can just calculate all the slopes, and 95% confidence intervals make less sense when there's fewer than 20 slopes, but it would be nice for theilslopes not breaking on small n. I suggest not using the approximation but consulting Kendall's U table directly for small n.

Thanks to everyone who's make scipy such a success!

mdhaber:
@rarnaout Thanks for reporting this. I've labeled it as an enhancement request: although the error messages in the current version of the code could be more graceful in these edge cases, they don't return an incorrect result (good), and the failure seems to be due to the equations themselves, which are designed for large samples. (This could also be mentioned a bit more explicitly in the docs rather than relying on the reader to be familiar with reference [1].) 

Essentially, this is asking for an exact method - something we're gradually adding to a lot of functions. PR welcome!

mdhaber:
Interesting, though. I get no warnings or errors in the first example, but a different error message than you show for the second:
```
ts([0, 0, 0], [0, 1, 0])  # Fails with message below
[...]

PR #23830, opened 2025-10-20
ENH: stats.theilslopes: vectorize implementation

mdhaber:
#### Reference issue
Toward gh-14651
Toward gh-20544

This vectorizes the `scipy.stats.theilslopes` implementation in preparation for translation to the array API.

#### Additional information
Ready for review, but at the end, I think we'll want to refactor as follows:
- Commit 1: Split `mstats` and `stats` implementation and tests into separate copies
- Commit 2: Replace `stats` implementation with this code
- Commit 3: Add tests of `stats` implementation against `mstats` implementation (since the tests are not very comprehensive and a lot is changing)


scipy/stats/_stats_mstats_common.py
```diff
@@ -135,55 +142,53 @@ def theilslopes(y, x=None, alpha=0.95, method='separate'):
     if method not in ['joint', 'separate']:
         raise ValueError("method must be either 'joint' or 'separate'."
                          f"'{method}' is invalid.")
-    # We copy both x and y so we can use _find_repeats.
-    y = np.array(y, dtype=float, copy=True).ravel()
```
mdhaber, line 139: default `ravel` behavior handled by `_axis_nan_policy` decorator (default_axis=None).

scipy/stats/_stats_mstats_common.py
```diff
@@ -135,55 +142,53 @@ def theilslopes(y, x=None, alpha=0.95, method='separate'):
...
-    if x is None:
-        x = np.arange(len(y), dtype=float)
-    else:
-        x = np.array(x, dtype=float, copy=True).ravel()
-        if len(x) != len(y):
-            raise ValueError("Array shapes are incompatible for broadcasting.")
-    if len(x) < 2:
-        raise ValueError("`x` and `y` must have length at least 2.")
```
mdhaber, line 147: Decorator handles this, too (`too_small=1`).

scipy/stats/_stats_mstats_common.py
```diff
@@ -135,55 +142,53 @@ def theilslopes(y, x=None, alpha=0.95, method='separate'):
...
+
+    y, x = xp_promote(y, x, force_floating=True, xp=np)
+    x = np.arange(y.shape[-1], dtype=y.dtype) if x is None else x
+    y, x = np.broadcast_arrays(y, x)
 
     # Compute sorted slopes only when deltax > 0
-    deltax = x[:, np.newaxis] - x
-    deltay = y[:, np.newaxis] - y
-    slopes = deltay[deltax > 0] / deltax[deltax > 0]
```
mdhaber, line 152: `deltax` and `deltay` are skew-symmetric with zeros on the diagonal, so `deltay/deltax` is symmetric with `nan`s on the diagonal. The main purpose of the `deltax > 0` mask is to eliminate all the duplicates and `nan`s before taking the median. This PR does same thing using an upper triangular mask over the last two axes. When we translate to array API, the mask has to be the only index, but it's doable with broadcasting the mask and reshaping the result.

We are *also* supposed to ignore any other places where `deltax == 0`; we do that by replacing these elements with `nan` and using `nanmedian`. We can use `quantile` with `nan_policy='omit'` when we translate to array API.

scipy/stats/_stats_mstats_common.py
```diff
@@ -135,55 +141,53 @@ def theilslopes(y, x=None, alpha=0.95, method='separate'):
...
         alpha = 1. - alpha
 
     z = distributions.norm.ppf(alpha / 2.)
     # This implements (2.6) from Sen (1968)
-    _, nxreps = _find_repeats(x)
-    _, nyreps = _find_repeats(y)
-    nt = len(slopes)       # N in Sen (1968)
-    ny = len(y)            # n in Sen (1968)
+    _, nxreps = _stats_py._rankdata(x, method='average', return_ties=True)
+    _, nyreps = _stats_py._rankdata(y, method='average', return_ties=True)
```
mdhaber: The second output of `_find_repeats` is like the counts of `unique_counts`. The second output of `return_ties` is similar, but it is has zeros to keep the array rectangular. Because the tie correction formula has terms like `nxreps * (nxreps-1)`, we get the same result.

scipy/stats/_stats_mstats_common.py
```diff
@@ -135,55 +141,53 @@ def theilslopes(y, x=None, alpha=0.95, method='separate'):
...
+        ny * (ny-1) * (2*ny+5)
+        - np.sum(nxreps * (nxreps-1) * (2*nxreps + 5), axis=-1, keepdims=True)
+        - np.sum(nyreps * (nyreps-1) * (2*nyreps + 5), axis=-1, keepdims=True))
     # Find the confidence interval indices in `slopes`
-    try:
-        sigma = np.sqrt(sigsq)
-        Ru = min(int(np.round((nt - z*sigma)/2.)), len(slopes)-1)
-        Rl = max(int(np.round((nt + z*sigma)/2.)) - 1, 0)
-        delta = slopes[[Rl, Ru]]
-    except (ValueError, IndexError):
```
mdhaber, line 182: One reason for the `try`-`except` was to deal with the possibility that `sigsq` is negative (gh-15943). Now I take care of that explicitly by replacing the results with `nan` where `sigsq < 0`. I don't really see how the `IndexError` would have happened before given the use of `min` and `max`.

scipy/stats/tests/test_axis_nan_policy.py
```diff
@@ -972,7 +972,7 @@ def test_non_broadcastable(hypotest, args, kwds, n_samples, n_outputs, paired,
 
     message = "Array shapes are incompatible for broadcasting."
     with pytest.raises(ValueError, match=message):
-        hypotest(*samples, *args, **kwds)
+        hypotest(*samples, *args, axis=axis, **kwds)
```
mdhaber, line 996: I noticed this was left out of the test.

mdhaber (commented):
Some self-review to aid the reviewer.

scipy/stats/_stats_mstats_common.py
```diff
@@ -135,55 +141,56 @@ def theilslopes(y, x=None, alpha=0.95, method='separate'):
...
+    deltax, deltay = deltax[..., i], deltay[..., i]
+    # with array API, mask must be sole index, so we'll need to broadcast it.
+    # indexing will ravel the array, so we'll need to reshape it after
+    deltax[deltax == 0] = np.nan
+    slopes = deltay / deltax
+    slopes = np.sort(slopes, axis=-1)
+    medslope = np.nanmedian(slopes, axis=-1)
     if method == 'joint':
-        medinter = np.median(y - medslope * x)
+        medinter = np.nanmedian(y - medslope * x, axis=-1)
```
mdhaber: I didn't inject NaNs into `x` or `y` - just into `deltax`.
```suggestion
        medinter = np.median(y - medslope * x, axis=-1)
```

scipy/stats/_stats_mstats_common.py
```diff
@@ -135,55 +141,56 @@ def theilslopes(y, x=None, alpha=0.95, method='separate'):
...
     else:
-        medinter = np.median(y) - medslope * np.median(x)
+        medinter = np.nanmedian(y, axis=-1) - medslope * np.nanmedian(x, axis=-1)
```
mdhaber:
```suggestion
        medinter = np.median(y, axis=-1) - medslope * np.median(x, axis=-1)
```

scipy/stats/_stats_mstats_common.py
mdhaber:
```suggestion
    # we don't actually need ranks, so an enhancement could be to have
    # `rankdata` return only the second output. In the meantime, use the
    # least expensive `method`.
    _, nxreps = _stats_py._rankdata(x, method='min', return_ties=True)
    _, nyreps = _stats_py._rankdata(y, method='min', return_ties=True)
```

mdhaber (commented):
A little more complex than the others, but still manageable.

j-bowhay:
Current changes all look good but I wasn't sure from the description if you were wanting to do more here

mdhaber:
Thanks as usual, Jake. Yes, I need to split the versions and add tests. I wanted to keep this a little simpler to review before I did that. I'll go ahead, preserving history so you can see the changes easily.

scipy/stats/_stats_mstats_common.py
```diff
@@ -135,55 +141,56 @@ def theilslopes(y, x=None, alpha=0.95, method='separate'):
...
-        x = np.arange(len(y), dtype=float)
-    else:
-        x = np.array(x, dtype=float, copy=True).ravel()
-        if len(x) != len(y):
-            raise ValueError("Array shapes are incompatible for broadcasting.")
-    if len(x) < 2:
+
+    y, x = xp_promote(y, x, force_floating=True, xp=np)
+    x = np.arange(y.shape[-1], dtype=y.dtype) if x is None else x
+    y, x = np.broadcast_arrays(y, x)
```
mdhaber: This wouldn't be necessary after the `_axis_nan_policy` decorator, but it might be if `x` was `None`.

mdhaber:
All set. I copy-pasted the vectorized implementation/tests into `_correlation.py`/`test_correlation.py` and restored the original versions. Then I moved the non-vectorized implementation into `_mstats_basic.py` where it is used. Turns out the mstats version has incorrect reference values in a test - and passes. Not going to debug that here. Corrected the reference values in the stats version of the test.

scipy/stats/tests/test_correlation.py
```diff
@@ -224,3 +227,103 @@ def check_nan(res):
...
+    def test_mask(self):
+        # Test for correct masking.
+        y = np.ma.array([0, 1, 100, 1], mask=[False, False, True, False])
+        slope, intercept, lower, upper = stats.theilslopes(y)
+        assert_almost_equal(slope, 0.5)
+        assert_almost_equal(intercept, 0.5)
+
+        slope, intercept, lower, upper = stats.theilslopes(y, method='joint')
+        assert_almost_equal(slope, 0.5)
+        assert_almost_equal(intercept, 0.0)
```
j-bowhay: I know it is just a copy and paste but I feel compelled not to introduce new uses of `assert_almost_equal`
```suggestion
        assert_allclose(slope, 0.5)
        assert_allclose(intercept, 0.5)

        slope, intercept, lower, upper = stats.theilslopes([0, 1, 1], method='joint')
        assert_allclose(slope, 0.5)
        assert_allclose(intercept, 0.0)

        # Test of confidence intervals from example in Sen (1968).
        x = [1, 2, 3, 4, 10, 12, 18]
        y = [9, 15, 19, 20, 45, 55, 78]
        slope, intercept, lower, upper = stats.theilslopes(y, x, 0.07)
        assert_allclose(slope, 4)
        assert_allclose(intercept, 4.0)
        assert_allclose(upper, 4.38, decimal=2)
        assert_allclose(lower, 3.71, decimal=2)

        slope, intercept, lower, upper = stats.theilslopes(y, x, 0.07,
                                                           method='joint')
        assert_allclose(slope, 4)
        assert_allclose(intercept, 6.0)
        assert_allclose(upper, 4.38, decimal=2)
        assert_allclose(lower, 3.71, decimal=2)
    @skip_xp_invalid_arg
    def test_mask(self):
        # Test for correct masking.
        y = np.ma.array([0, 1, 100, 1], mask=[False, False, True, False])
        slope, intercept, lower, upper = stats.theilslopes(y)
        assert_allclose(slope, 0.5)
        assert_allclose(intercept, 0.5)

        slope, intercept, lower, upper = stats.theilslopes(y, method='joint')
        assert_allclose(slope, 0.5)
        assert_allclose(intercept, 0.0)
```

scipy/stats/_correlation.py
```diff
@@ -383,3 +386,179 @@ def spearmanrho(x, y, /, *, alternative='two-sided', method=None, axis=0):
...
+        Confidence degree between 0 and 1. Default is 95% confidence.
+        Note that `alpha` is symmetric around 0.5, i.e. both 0.1 and 0.9 are
+        interpreted as "find the 90% confidence interval".
+    method : {'joint', 'separate'}, optional
+        Method to be used for computing estimate for intercept.
+        Following methods are supported,
+
+        * 'joint': Uses np.median(y - slope * x) as intercept.
+        * 'separate': Uses np.median(y) - slope * np.median(x)
+                      as intercept.
```
j-bowhay:
```suggestion
        * 'joint': Uses ``np.median(y - slope * x)`` as intercept.
        * 'separate': Uses ``np.median(y) - slope * np.median(x)``
                      as intercept.
```
Unrelated I know but since there are other suggestions

scipy/stats/tests/test_correlation.py
```diff
@@ -224,3 +227,103 @@ def check_nan(res):
         with pytest.warns(stats.ConstantInputWarning, match=message):
             res = stats.spearmanrho(y, x)
             check_nan(res)
+
+
+class TestTheilslopes:
+    def test_theilslopes(self):
```
j-bowhay, line 233: It looks like this test is the same as https://github.com/scipy/scipy/blob/eefc1c39a7f20d44444914a464e0745762a0b771/scipy/stats/tests/test_stats.py#L2435-L2466

except the check for invalid method so it would be good to unduplicate

scipy/stats/tests/test_correlation.py
mdhaber: There will be an immediate follow-up converting to `xp_assert_close`. And `assert_allclose` doesn't accept `decimal`, so I can't just accept the recommendation. I'll replace `assert_almost_equal` though.

j-bowhay: Yes good point fine to leave then

mdhaber:
Made the requested adjustments and fixed the lint failures, so should be good to go. Thanks, Jake!

j-bowhay (approved):
Thanks!

PR #23837, opened 2025-10-21
ENH: stats.mood: vectorize implementation

mdhaber:
#### Reference issue
Toward gh-14651
Toward gh-20544

This vectorizes the implementation of `scipy.stats.mood` and prepares for array API translation.

#### Additional information 
It is probably easier to review the commits separately. They are pretty organized. CI failures are unrelated.


tylerjereddy (commented):
If I'm reading the test changes correctly, the new testing is at least as thorough as the original.  I checked the test line coverage on this feature branch locally, and it looks like all the lines changed in the diff are covered.

So, it looks good at first glance, though well outside my wheelhouse.

mdhaber:
Thanks for taking a look, Tyler!

scipy/stats/_morestats.py
```diff
@@ -3293,78 +3293,69 @@ def func(x):
...
 
-    varM = (m * n * (N + 1.0) * (N ** 2 - 4) / 180 -
-            m * n / (180 * N * (N - 1)) * np.sum(
-                t * (t**2 - 1) * (t**2 - 4 + (15 * (N - S - S_i_m1) ** 2))
-            ))
-
-    return ((T - E_0_T) / np.sqrt(varM),)
+    # m, n, N, t, and S are defined in the second paragraph of Mielke pg 312
+    # The only difference is that our `t` has zeros interspersed with the relevant
+    # numbers to keep the array rectangular, but these terms nothing to the sum.
```
lucascolley:
```suggestion
    # numbers to keep the array rectangular, but these terms do not change the sum.
```

scipy/stats/_morestats.py
```diff
@@ -3293,78 +3293,69 @@ def func(x):
...
+    x = np.sort(x, axis=-1)
+    xy = np.concatenate((x, y), axis=-1)
+    i = np.argsort(xy, stable=True, axis=-1)
+    _, a = _stats_py._rankdata(x, method='average', return_ties=True)
+    zeros = np.zeros(a.shape[:-1] + (m,))
+    a = np.concatenate((a, zeros), axis=-1)
+    a = np.take_along_axis(a, i, axis=-1)
+
+    # Mielke pg 312 defines test statistic `T` as the inner product `a` and `phi`
+    T = np.vecdot(a, phi, axis=-1)
```
lucascolley, line 3407: I'm inclined to just trust you on the accuracy here, as this diff is not easy to review. Perhaps instead, I could get you to summarise what has significantly changed from the previous implementation?

mdhaber: Yeah, this was not a translation but a re-implementation from the original paper. It's OK if this needs to a wait for a more thorough review. (I'm not sure anyone is more suited for doing that than anyone else, since it's just carefully following formulas in the paper.) I included the most important notes as code comments, but I'll add some more.

scipy/stats/_morestats.py
mdhaber: Oops, here was the missing word.
```suggestion
    # numbers to keep the array rectangular, but these terms add nothing to the sum.
```
I'll use this comment to explain the format of `t`.

```python3
from array_api_compat import numpy as np
from scipy.stats._stats_py import _rankdata
x = np.asarray([-1, 0, 0, 1, 1, 1, 2, 2, 3])

# r is the ranks of the data. When there are ties, all observations
# with the same value are assigned the average of the ranks that would have
# been assigned had the ties been broken arbitrarily. For instance, the zeros
# would have received ranks 2 and 3 had ties been broaken, so the everage
# is 2.5. This is commonly used in rank-based tests and is called for in the
# reference.
r  # array([1. , 2.5, 2.5, 5. , 5. , 5. , 7.5, 7.5, 9. ])

# `t` is the multiplicity of each value in the data (equivalently, the ranks)
t  # array([1., 2., 0., 3., 0., 0., 2., 0., 1.])
# This format - recording zero at locations corresponding with indices where
# the value would otherwise be repeated - is good because it allows us to
# easily perform common calculations like `sum(t**3 - t)` (found in many tie
# corrrection formulas) in a vectorized way without having a ragged array
# (because the number of unique value could be different for each slice).

# One peculiarity is that the ranks `r` reflect those of the original data
# whereas `t` corresponds with the values in sorted order. This was a matter
# of implementation convenience/efficiency - often we *need* `r` to correspond
# with the order of `x`, but we *don't* need that for `t`, so we don't bother.
rng = np.random.default_rng(22398492834984884)
r, t = _rankdata(rng.permuted(x), method='average', return_ties=True)
r  # array([5. , 2.5, 9. , 1. , 5. , 2.5, 5. , 7.5, 7.5])
t  # array([1., 2., 0., 3., 0., 0., 2., 0., 1.])), same as before
```

scipy/stats/_morestats.py
```diff
@@ -3293,78 +3293,69 @@ def func(x):
...
+    # First equation of "Mood's Squared Rank Test", Mielke pg 313
     E_0_T = n * (N * N - 1) / 12
 
-    varM = (m * n * (N + 1.0) * (N ** 2 - 4) / 180 -
-            m * n / (180 * N * (N - 1)) * np.sum(
-                t * (t**2 - 1) * (t**2 - 4 + (15 * (N - S - S_i_m1) ** 2))
-            ))
-
-    return ((T - E_0_T) / np.sqrt(varM),)
+    # m, n, N, t, and S are defined in the second paragraph of Mielke pg 312
```
mdhaber, line 3357: We'll need the other definitions, too, so:
image

scipy/stats/_morestats.py
```diff
@@ -3293,78 +3293,69 @@ def func(x):
...
+    # The only difference is that our `t` has zeros interspersed with the relevant
+    # numbers to keep the array rectangular, but these terms nothing to the sum.
+    np = array_namespace(x, y, t)  # needed to use `cumulative_sum`
+    S = np.cumulative_sum(t, include_initial=True, axis=-1)
+    S_i, S_i_m1 = S[..., 1:], S[..., :-1]
```
mdhaber, line 3362: See definition of $S$ above - the cumulative sum of ranks  in the pooled sample. The difference is that the $i$th element of the array doesn't necessarily correspond with $S_i$ because there are zeros in `t` that correspond with repeated values. So for the data above, $S$ might look like:
```python3
S = np.cumulative_sum(t, include_initial=True, axis=-1)
S  # array([0., 1., 3., 3., 6., 6., 6., 8., 8., 9.])
```
So $S_0 = 0$, $S_1 = 1$, and $S_2 = 3$, but $S_3 = 6$, not $3$.

Despite all this, our `t`, `S_i` and `S_i_m1` ($S_{i-1}$) still line up as we would want them to.
```python3
t       # array([1., 2., 0., 3., 0., 0., 2., 0., 1.])
S[1:]   # array([1., 3., 3., 6., 6., 6., 8., 8., 9.])
S[:-1]  # array([0., 1., 3., 3., 6., 6., 6., 8., 8.])
```

scipy/stats/_morestats.py
```diff
@@ -3293,78 +3293,69 @@ def func(x):
...
+    # Second equation of "Mood's Squared Rank Test", Mielke pg 313
+    varM = (m * n * (N + 1.0) * (N**2 - 4) / 180
+            - m * n / (180 * N * (N - 1))
+            * np.sum(t * (t ** 2 - 1) * (t ** 2 - 4 + (15 * (N - S_i - S_i_m1) ** 2)),
```
mdhaber, line 3366: This formula can be implemented just as it appears in the paper:
image

because terms where `t` is `0.` aren't included in the sum.

scipy/stats/_morestats.py
```diff
@@ -3293,78 +3293,69 @@ def func(x):
...
-
-    # `T` is equal to a[j] * phi[j], per the first unnumbered equation on
-    # page 312. `phis` is already in the order based on `js`, so we index
-    # into `a` with `js` as well.
-    T = sum(phis * a[js])
-
-    # The approximate statistic
+def _mood_statistic_with_ties(x, y, t, n, m, N):
+    # First equation of "Mood's Squared Rank Test", Mielke pg 313
     E_0_T = n * (N * N - 1) / 12
```
mdhaber: image

scipy/stats/_morestats.py
```diff
@@ -3293,78 +3293,69 @@ def func(x):
...
+    S_i, S_i_m1 = S[..., 1:], S[..., :-1]
+    # Second equation of "Mood's Squared Rank Test", Mielke pg 313
+    varM = (m * n * (N + 1.0) * (N**2 - 4) / 180
+            - m * n / (180 * N * (N - 1))
+            * np.sum(t * (t ** 2 - 1) * (t ** 2 - 4 + (15 * (N - S_i - S_i_m1) ** 2)),
+                     axis=-1))
+
+    # There is a formula for phi in terms of t, S, and Psi(I) = [I - (N+1)/2]^2 is
+    # defined (with a mistake in the location of the ^2) at the beginning of
+    # "Mood's Squared Rank Test", Mielke pg 313. To vectorize this calculation,
```
mdhaber: image
image

where

image

scipy/stats/_morestats.py
```diff
@@ -3293,78 +3293,69 @@ def func(x):
...
+    # let c = (N + 1) / 2, so Psi(I) = I^2 - 2*c*I + c^2. We sum each of these three
+    # parts of Psi separately using formulas for sums from a to b (inclusive) of
+    # terms I^2, I, and 1 where I takes on successive integers.
```
mdhaber: This comment is about one of the most significant changes required to vectorize this: use summation formulas rather than summing with NumPy.

scipy/stats/_morestats.py
```diff
@@ -3293,78 +3293,69 @@ def func(x):
...
+        return (b - a) + 1
+
+    with np.errstate(invalid='ignore', divide='ignore'):
+        sum_I2 = sum_I2(S_i_m1 + 1, S_i)
+        sum_I = sum_I(S_i_m1 + 1, S_i)
+        sum_1 = sum_1(S_i_m1 + 1, S_i)
+        c = (N + 1) / 2
+        phi = (sum_I2 - 2*c*sum_I + sum_1*c**2) / t
+
+    phi[t == 0] = 0  # where t = 0 we get NaNs; eliminate them
```
mdhaber, line 3393: It is not SO bad to compare the implementation of the code above to the math if you assume take `Si` to represent $S_i$ and `S_i_m1` to represent $S_{i-1}$ at face value. The tricky thing is those zeros in `t` and corresponding duplicate elements in `S_i` and `S_i_m1`. But it all works out:
- The unique terms we care about are calculated because `t`, `S_i`, and `S_i_m1` are all aligned as we expect.
- The terms we don't care about turn into NaNs because at these locations, `t == 0` and `S_i == S_i_m1`.
- We replace the NaNs with zeros so they don't bite us later.

scipy/stats/_morestats.py
```diff
@@ -3293,78 +3293,69 @@ def func(x):
...
+
+    # Mielke pg 312 defines `a` as the count of elements in sample `x` for each of the
+    # unique values in the combined sample. The tricky thing is getting these to line
+    # up with the locations of nonzero elements in `t`/`phi`.
```
mdhaber, line 3397: Repeating for convenience:

image

We calculated $\phi$ as `phi`, but it has a lot of terms that don't correspond with the $a_i$. There are two reasons it's tricky:
- First, $phi$ it is computed based on the pooled sample, so the order of the terms is related to the sorted order of the pooled sample, whereas $a_i$ is the multiplicity of terms in only one of the two samples
- Second, `phi` has a bunch of zeros in-between the terms we care about.

Before we take the `vectdot`, we have to line things up.

scipy/stats/_morestats.py
```diff
@@ -3293,78 +3293,69 @@ def func(x):
...
+    x = np.sort(x, axis=-1)
+    xy = np.concatenate((x, y), axis=-1)
+    i = np.argsort(xy, stable=True, axis=-1)
```
mdhaber, line 3400: This finds the locations where the relevant `a_i` need to go. It might be easiest to think through how `rankdata` would have found `t` to begin with: we concatenated the samples, and then rankdata produces `t` that has the multiplicity of each element in a sorted order (see https://github.com/scipy/scipy/pull/23837#discussion_r2491120035). So if `x` starts in a sorted order, and then concatenate `y` (important: _append_ the elements of `y`), and then do a _stable_ argsort, then the indices tell us where the elements of `x` (and therefore the elemnts of `a` below) ended up.

It's probably possible to refactor the code (including code internal to `_rankdata`) so that we don't have to re-calculate these indices. That would be more computationally efficient since we avoid sorting the same data twice, but I don't really want to pick apart `_rankdata` or have it return any other information at this time.

scipy/stats/_morestats.py
```diff
@@ -3293,78 +3293,69 @@ def func(x):
...
+    _, a = _stats_py._rankdata(x, method='average', return_ties=True)
```
mdhaber, line 3401: This (oops - last line only) actually finds the `a_i`s

scipy/stats/_morestats.py
```diff
@@ -3293,78 +3293,69 @@ def func(x):
...
+    zeros = np.zeros(a.shape[:-1] + (m,))
+    a = np.concatenate((a, zeros), axis=-1)
+    a = np.take_along_axis(a, i, axis=-1)
```
mdhaber: This puts them in the order that corresponds with `phi` - along with zeros for the terms we don't care about.

scipy/stats/_morestats.py
```diff
@@ -3293,78 +3293,69 @@ def func(x):
...
+    # parts of Psi separately using formulas for sums from a to b (inclusive) of
+    # terms I^2, I, and 1 where I takes on successive integers.
+    def sum_I2(a, b):
+        return (b * (b + 1) * (2*b + 1) - a * (a - 1) * (2*a - 1)) / 6
+
+    def sum_I(a, b):
+        return (b * (b + 1) - a * (a - 1)) / 2
+
+    def sum_1(a, b):
+        return (b - a) + 1
```
mdhaber, line 3384: This is just the *count* of consecutive integers from `a` to `b`, inclusive.

scipy/stats/_morestats.py
```diff
@@ -3293,78 +3293,69 @@ def func(x):
...
+    # defined (with a mistake in the location of the ^2) at the beginning of
+    # "Mood's Squared Rank Test", Mielke pg 313. To vectorize this calculation,
+    # let c = (N + 1) / 2, so Psi(I) = I^2 - 2*c*I + c^2. We sum each of these three
+    # parts of Psi separately using formulas for sums from a to b (inclusive) of
+    # terms I^2, I, and 1 where I takes on successive integers.
+    def sum_I2(a, b):
+        return (b * (b + 1) * (2*b + 1) - a * (a - 1) * (2*a - 1)) / 6
+
+    def sum_I(a, b):
+        return (b * (b + 1) - a * (a - 1)) / 2
```
mdhaber: The sum of consecutive integers from `a` to `b`, inclusive. The inclusivity is why there's a minus sign instead of a plus in the second term.
```python3
from sympy import *
a, b = symbols('a, b')
x = b * (b + 1) / 2 - a * (a + 1) / 2 + a
y = (b * (b + 1) - a * (a - 1)) / 2
simplify(x - y)
```

scipy/stats/_morestats.py
```diff
@@ -3293,78 +3293,69 @@ def func(x):
...
+                     axis=-1))
+
+    # There is a formula for phi in terms of t, S, and Psi(I) = [I - (N+1)/2]^2 is
+    # defined (with a mistake in the location of the ^2) at the beginning of
+    # "Mood's Squared Rank Test", Mielke pg 313. To vectorize this calculation,
+    # let c = (N + 1) / 2, so Psi(I) = I^2 - 2*c*I + c^2. We sum each of these three
+    # parts of Psi separately using formulas for sums from a to b (inclusive) of
+    # terms I^2, I, and 1 where I takes on successive integers.
+    def sum_I2(a, b):
+        return (b * (b + 1) * (2*b + 1) - a * (a - 1) * (2*a - 1)) / 6
```
mdhaber: [Sum of squares of consecutive integers](https://en.wikipedia.org/wiki/Square_pyramidal_number#Formula) from `a` to `b`, inclusive. 

```python3
from sympy import *
a, b = symbols('a, b')

def sumn2(n):
    return n * (n + 1) * (2 * n + 1) / 6

x = sumn2(b) - sumn2(a) + a**2
y = (b * (b + 1) * (2*b + 1) - a * (a - 1) * (2*a - 1)) / 6
simplify(x - y)
```

I can refactor as `sumn2(b) - sumn2(a) + a**2` if preferred.

mdhaber (commented):
It would be good for someone else to understand how this implements the calculation in a vectorized way. Hopefully these comments help!

scipy/stats/_morestats.py
```diff
@@ -3293,78 +3293,69 @@ def func(x):
...
     E_0_T = n * (N * N - 1) / 12
 
-    varM = (m * n * (N + 1.0) * (N ** 2 - 4) / 180 -
-            m * n / (180 * N * (N - 1)) * np.sum(
-                t * (t**2 - 1) * (t**2 - 4 + (15 * (N - S - S_i_m1) ** 2))
-            ))
-
-    return ((T - E_0_T) / np.sqrt(varM),)
+    # m, n, N, t, and S are defined in the second paragraph of Mielke pg 312
+    # The only difference is that our `t` has zeros interspersed with the relevant
```
lucascolley, line 3358: To be pedantic, I think this isn't strictly true, as `m` and `n` are swapped between our implementation and the paper, right?

scipy/stats/_morestats.py
lucascolley: makes sense, thanks

scipy/stats/_morestats.py
```diff
@@ -3293,78 +3293,69 @@ def func(x):
...
+    S = np.cumulative_sum(t, include_initial=True, axis=-1)
+    S_i, S_i_m1 = S[..., 1:], S[..., :-1]
+    # Second equation of "Mood's Squared Rank Test", Mielke pg 313
+    varM = (m * n * (N + 1.0) * (N**2 - 4) / 180
+            - m * n / (180 * N * (N - 1))
+            * np.sum(t * (t ** 2 - 1) * (t ** 2 - 4 + (15 * (N - S_i - S_i_m1) ** 2)),
+                     axis=-1))
+
+    # There is a formula for phi in terms of t, S, and Psi(I) = [I - (N+1)/2]^2 is
+    # defined (with a mistake in the location of the ^2) at the beginning of
```
lucascolley: correct?

```suggestion
    # There is a formula for phi in terms of t, S, and Psi(I) = [I - (N+1)/2]^2.
    # Psi(I) is defined (with a mistake in the location of the ^2) at the beginning of
```

scipy/stats/_morestats.py
lucascolley: > I can refactor as `sumn2(b) - sumn2(a) + a**2` if preferred.

seems nicer, yes!

lucascolley (approved):
thanks for the detailed comments Matt! I took some time this evening to walk through the paper and implementation with your comments. Happy to merge, just 2 very minor comments.

scipy/stats/_morestats.py
mdhaber: Ah, good catch. Looks like that is left over from Travis' initial implementation. I'll adjust that.

mdhaber: Please check me here. I think I swapped everything, but there is at least one blind spot because the tests w/ ties have $m = n$. I checked a case with ties and $m \neq n$ and get the same results as before.

scipy/stats/_morestats.py
mdhaber: OK, done.

scipy/stats/_morestats.py
```diff
@@ -3408,17 +3408,17 @@ def sum_1(a, b):
 
 
 def _mood_statistic_no_ties(r, m, n, N):
-    rx = r[..., :n]
+    rx = r[..., :m]
     M = np.sum((rx - (N + 1.0) / 2) ** 2, axis=-1)
-    mnM = n * (N * N - 1.0) / 12
+    mnM = m * (N * N - 1.0) / 12
     varM = m * n * (N + 1.0) * (N + 2) * (N - 2) / 180
     return (M - mnM) / sqrt(varM)
```
lucascolley: would it be more consistent to rename `mnM` to `E_0_M`?

Is there a reason for using `M` here, and both `M` and `T` in the variable names in the ties function? They are referring to the same thing, right?

lucascolley (commented):
`m` <-> `n` swap indeed looks correct, thanks!

scipy/stats/_morestats.py
mdhaber: Ask Travis : ) Presumably the original reference used a different notation. I copy-pasted these lines, but didn't touch them otherwise, so they were still using the original variable names. How about I change them during the array API conversion rather than re-running CI here?

lucascolley: Sounds good 👍

My main concern is `mnM`, which is perhaps confusingly not `m*n*M`.

mdhaber: Probably "mean M".

lucascolley: Yes I figured

PR #23841, opened 2025-10-22
BLD: bump wheel build image

andyfaff:
macos-13 images are deprecated

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -37,7 +37,7 @@ jobs:
   get_commit_message:
     name: Get commit message
     runs-on: ubuntu-latest
-    if: github.repository == 'andyfaff/scipy'
+    if: github.repository == 'scipy/scipy'
     outputs:
       message: ${{ steps.commit_message.outputs.message }}
     steps:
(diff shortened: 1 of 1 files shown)


rgommers:
Thanks @andyfaff! The changes LGTM, but for some reason the jobs on `macos-15-intel` are super slow:
- [This log](https://github.com/scipy/scipy/actions/runs/18708785886/job/53352083519?pr=23841) from this PR shows the "Build wheels" step took 1h 13m
- In [the latest job on `main`](https://github.com/scipy/scipy/actions/runs/18711360061/job/53360401013) it took 31m 43s

Did you look into that already by any chance?

andyfaff:
I think it's the container not being powerful enough. The numpy wheels built on `macos-15-intel` also take a long time.

tylerjereddy (approved):
Let's try it I guess. Re: the slowdown, comparing the jobs Ralf mentioned indicates that building the wheel slows down by ~1.4x, and running the tests slows down by ~2.9x (!).

A recent equivalent image/architecture build in the `numpy-release` repo didn't seem too bad when I checked it (https://github.com/numpy/numpy-release/actions/runs/18631393141/job/53118654871), at about 19 minutes total.

Maybe we can smooth it out in a follow-up, though I did see a fair number of open issues open about slowdowns in `macos-15` even from a simple Google search of open tickets.

One question, [the standard runner docs](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories) list 4 cores for macOS + Intel in the regular/free tier--if we actually get those cores (or are they shared via virtualization?) could we try using `xdist` with i.e., 2 procs to see if it helps mitigate the 3x slowdown on tests a bit? Looks like we're currently running a single `pytest` proc. Maybe something to try on a fork..

tylerjereddy:
Also, since it is OpenBLAS-specific... the usual concerns about the N environment variables that screw up concurrency maybe? We do have `threadpoolctl` installed in the two compared jobs/images though.

tylerjereddy:
The NumPy job on same image is using `-n auto` and I can see by the `pytest` output style that more than one proc is being used there vs. just one for us. Maybe worth a try.

andyfaff:
The tests are run by `python -c 'import sys; import scipy; sys.exit(not scipy.test())'` in `tools/wheels/cibw_test_command`. So there's scope for setting environment variables.

rgommers:
Adding `extra_argv=['-n=auto']` there would make sense. I can't remember if we ever had a reason for not doing that, or if it was an oversight.

PR #23849, opened 2025-10-23
ENH: stats.chatterjeexi: add array API support

mdhaber:
#### Reference issue
Toward gh-20544

Adds array API support to `scipy.stats.chatterjeexi`.

#### Additional information
Dask doesn't have `take_along_axis`, and JAX misses some of the input validation niceties from the `_axis_nan_policy` decorator. Otherwise, everything went pretty smoothly.

### DIFF
--- a/scipy/stats/_correlation.py
+++ b/scipy/stats/_correlation.py
@@ -1,6 +1,6 @@
-import numpy as np
+import math
 from scipy import stats
-from scipy._lib._array_api import xp_capabilities
+from scipy._lib._array_api import xp_capabilities, array_namespace, xp_promote
 from scipy.stats._stats_py import _SimpleNormal, SignificanceResult, _get_pvalue
 from scipy.stats._axis_nan_policy import _axis_nan_policy_factory
 
@@ -8,11 +8,7 @@ from scipy.stats._axis_nan_policy import _axis_nan_policy_factory
 __all__ = ['chatterjeexi']
 
 
-# TODO:
-# - Adjust to respect dtype
-
-
-def _xi_statistic(x, y, y_continuous):
+def _xi_statistic(x, y, y_continuous, xp):
     # Compute xi correlation statistic
 
     # `axis=-1` is guaranteed by _axis_nan_policy decorator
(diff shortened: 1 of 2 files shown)


scipy/stats/tests/test_correlation.py
```diff
@@ -1,16 +1,20 @@
 import pytest
 import numpy as np
-from numpy.testing import assert_allclose
 
+from scipy._lib._array_api import make_xp_test_case, xp_default_dtype, is_jax
+from scipy._lib._array_api_no_0d import xp_assert_close
 from scipy import stats
 from scipy.stats._axis_nan_policy import SmallSampleWarning
+import torch
```
mdhaber: Oops. Was testing different default dtype.
```suggestion
```

scipy/stats/tests/test_correlation.py
```diff
@@ -23,16 +27,17 @@ def test_against_R_XICOR(self, case):
...
+        x, y = xp.asarray(x, dtype=dtype), xp.asarray(y, dtype=dtype)
         res = stats.chatterjeexi(x, y, y_continuous=case['y_cont'])
 
-        assert_allclose(res.statistic, case['statistic'])
-        assert_allclose(res.pvalue, case['pvalue'])
+        xp_assert_close(res.statistic, xp.asarray(case['statistic'], dtype=dtype))
+        xp_assert_close(res.pvalue, xp.asarray(case['pvalue'], dtype=dtype))
 
     @pytest.mark.parametrize('y_continuous', (False, True))
     def test_permutation_asymptotic(self, y_continuous):
```
mdhaber, line 44: This can get converted when gh-23772 is merged, but I don't think that needs to hold up this PR.

scipy/stats/tests/test_correlation.py
```diff
@@ -23,16 +26,19 @@ def test_against_R_XICOR(self, case):
         #       0.7620801065573549, 0.31410063302647495, 0.7935620302236199,
         #       0.5423085761365468)
         # xicor(x, y, ties=FALSE, pvalue=TRUE)
-
+        if dtype == 'float32' and np.__version__ < "2":
+            pytest.skip("Scalar dtypes only respected after NEP 50.")
```
tylerjereddy, line 30: I'm always skeptical, but this was indeed needed locally to avoid failures with NumPy 1.26.4 + array API testing active.
Does the full array API testsuite for SciPy actually pass for you locally with 1.26.4? I see some issues with torch and array-api-strict in that configuration, not sure that is surprising though, and not connected to this PR I don't think (I see failures on main as well--maybe a similar shim is needed in more places, though I think also `asarray()` call signature stuff was popping up).

scipy/stats/tests/test_correlation.py
```diff
@@ -51,17 +57,20 @@ def test_permutation_asymptotic(self, y_continuous):
...
             stats.chatterjeexi(x, y[:-1])
 
-        message = '...axis 10 is out of bounds for array...'
-        with pytest.raises(ValueError, match=message):
-            stats.chatterjeexi(x, y, axis=10)
+        if not is_jax(xp):
+            # jax misses out on some input validation from _axis_nan_policy decorator
+            message = '...axis 10 is out of bounds for array...|out of range'
+            with pytest.raises((ValueError, IndexError), match=message):
+                stats.chatterjeexi(x, y, axis=10)
```
tylerjereddy, line 73: this looks good--testing `jax` is retained modulo the exception handling, which I think is not a surprising limitation per previous discussions

scipy/stats/_correlation.py
```diff
@@ -1,69 +1,66 @@
...
     return statistic, r, l
 
 
-def _xi_std(r, l, y_continuous):
+def _xi_std(r, l, y_continuous, xp):
     # Compute asymptotic standard deviation of xi under null hypothesis of independence
 
     # `axis=-1` is guaranteed by _axis_nan_policy decorator
-    n = np.float64(r.shape[-1])
+    n = r.shape[-1]
```
tylerjereddy, line 44: I thought we used to have to shim around `shape` calls, I must be falling behind.

tylerjereddy (approved):
I'm pretty happy with this one. I added some inline comments, but they were mostly just positive things or queries.

Testing the full array API suite locally on ARM Mac with `torch` `2.8.0` installed worked fine on this branch. Testing with NumPy `1.26.4` + array API is problematic, but that's also true on `main` (and I believe not actually surprising, or at least not for cleaning up here).

+1 for merge on my end I think; maybe I'll give it a little time in case actual `stats` folks want to comment

scipy/stats/tests/test_correlation.py
mdhaber: > Does the full array API testsuite for SciPy actually pass for you locally with 1.26.4? 

Not sure, haven't tested. I'm not too concerned - with these stricter tests (`xp_assert_close`) we're just exposing dtypes issues that already existed. I'd rather not fight the pre-NEP 50 promotion rules to get the behavior right with NumPy < 2.0 when we can just leave it alone for a few months until NumPy >= 2.0 is the minimum supported version.

scipy/stats/_correlation.py
mdhaber: Do you mean `xp_size`? I don't think we need anything for `shape`. 
Here I'm just leaving it a Python scalar so it doesn't promote the whole calculation to `float64`.

mdhaber:
Thanks for taking a look, Tyler!

tylerjereddy:
Thanks Matt, we probably don't have the bandwidth to get second approvals on these, and I was already happy a few days ago.

mdhaber:
Thanks, Tyler! I am comfortable with that, too.

PR #23850, opened 2025-10-23
DOC: Improve documentation for the directed Hausdorff distance

star1327p:

Improve formatting and clarity of performance complexity description.


tylerjereddy (approved):
Thanks, I checked the rendered version of the docs, and they do indeed look a bit better here vs. the current devdocs for Hausdorff.

Incidentally, that paper is getting a lot of citations, over 500 now.

Issue #22816, opened 2025-04-08
BUG: stats.spearmanr: incorrectly identifies arrays as constant


jakobhansen-blai:
### Describe your issue.

When `scipy.stats.spearmanr()` checks whether the input array is constant, it appears to only check whether the first two columns of the array are constant. This leads it to issue a `ConstantInputWarning` for nonconstant arrays, and (much worse) fail to calculate any correlations at all, even if there are nonconstant variables in the input.

### Reproducing Code Example

```python
X = np.zeros((10, 4))
X[:,2] = np.arange(10)
X[:,3] = -np.arange(10)
scipy.stats.spearmanr(X)
```

### Error message

```shell
ConstantInputWarning: An input array is constant; the correlation coefficient is not defined.
```

### SciPy/NumPy/Python version and system information

```shell
1.15.1 1.26.4 sys.version_info(major=3, minor=12, micro=8, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    detection method: extraframeworks
    found: true
    include directory: unknown
    lib directory: unknown
    name: Accelerate
    openblas configuration: unknown
    pc file directory: unknown
    version: unknown
  lapack:
    detection method: extraframeworks
    found: true
    include directory: unknown
    lib directory: unknown
    name: Accelerate
    openblas configuration: unknown
    pc file directory: unknown
    version: unknown
  pybind11:
    detection method: config-tool
    include directory: unknown
    name: pybind11
    version: 2.13.6
Compilers:
  c:
    commands: cc
    linker: ld64
    name: clang
    version: 15.0.0
  c++:
    commands: c++
    linker: ld64
    name: clang
    version: 15.0.0
  cython:
    commands: cython
    linker: cython
    name: cython
    version: 3.0.11
  fortran:
    commands: gfortran
    linker: ld64
    name: gcc
    version: 13.3.0
  pythran:
    include directory: ../../../../../../private/var/folders/ml/x_tvfgn50_s7p67dm1ypcqqm0000gn/T/pip-build-env-3iq8ohgb/overlay/lib/python3.12/site-packages/pythran
    version: 0.17.0
Machine Information:
  build:
    cpu: aarch64
    endian: little
    family: aarch64
    system: darwin
  cross-compiled: false
  host:
    cpu: aarch64
    endian: little
    family: aarch64
    system: darwin
Python Information:
[...]

Issue #17746, opened 2023-01-07
ENH: confidence interval for spearman confidence interval


lesshaste:
### Is your feature request related to a problem? Please describe.

Scipy currently doesn’t return a confidence interval for the spearman correlation coefficient.  This is standard in R however, https://search.r-project.org/CRAN/refmans/DescTools/html/SpearmanRho.html .

### Describe the solution you'd like.

An option to compute the confidence interval

mdhaber:
We're gradually working on adding confidence intervals to the statistics that do not already have them. In the meantime, consider `scipy.stats.bootstrap`:
```python3
import numpy as np
from scipy import stats
rng = np.random.default_rng(82468935639010)

x = np.linspace(0, 10)
y = 1.5*x + rng.normal(size=x.shape)

def statistic(x, y):
    return stats.spearmanr(x, y).statistic

res = stats.bootstrap((x, y), statistic, method='bca', paired=True)

print(statistic(x, y), res.confidence_interval)
# 0.9726290516206483 ConfidenceInterval(low=0.9517124639523833, high=0.9884739535066726)
# compared to:
# SpearmanRho(x, y, conf.level=0.95)
#      rho    lwr.ci    upr.ci 
# 0.9726291 0.9520211 0.9844559 
```

lesshaste:
Thank you so much. That code is very helpful.

mdhaber:
@lesshaste if you like gh-18227 and would like that to be added to `spearmanr`, please consider mentioning your support in that PR.

Issue #17089, opened 2022-09-25
ENH: Documentation on test behind p-values of .spearmanr


tstoeger:
### Is your feature request related to a problem? Please describe.

The docstring of [.spearmanr](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.spearmanr.html#scipy.stats.spearmanr) currently does not report on the test which is used to determine the p-value.

Used scipy.stats.spearmanr for a scientific manuscript, and journal editor requested more information about used test.

Reading the code, the specific test is not evident. While it appears based on t distribution, could not identify the full procedure by name. Nor, identify by name the preprocessing scheme between computation of Spearman's rho and comparison against t distribution.

Possibly related to first point under [#8789](https://github.com/scipy/scipy/issues/8789). 

### Describe the solution you'd like.

Docstring stating or citing the test used for p-values.

mdhaber:
I can help you find the information. Would you be willing to submit a PR adding this information to the docstring? Then I could review/merge the PR.

tstoeger:
Thank you for helping me find the information.

Happy to then submit PR.

mdhaber:
Wikipedia mentions the same use of the t distribution and gives citations, but I can't find them online.

![image](https://user-images.githubusercontent.com/6570539/192166531-e6f20b87-b133-4b49-99f9-e6774153b32e.png)

I'll see if I can a more accessible reference.

tstoeger:
Thank you! This is wonderful. Will later this week prepare PR.

mdhaber:
Hi @tstoeger - please ping me when you open a PR so I don't miss it. Thanks!

mdhaber:
@tstoeger just thought I'd check in. Are you still able to submit this PR so we can complete this issue?

tstoeger:
@mdhaber Apologies for the delay. Got caught up by manuscript revision. Submitted PR. Thank you again!

PR #23851, opened 2025-10-23
ENH: stats.spearmanrho: array API compatible substitute for `spearmanr`

### LINKED ISSUE #23238: RFC: stats.spearmanr: interface adjustment to enable array API support
Now that `stats.rankdata` has been translated to the array API standard (gh-20639), it's possible for several other stats functions to follow, including `stats.spearmanr`. 

It would be challenging to match the current interface, though, and I'm not sure it sure it is the best interface for users. To motivate that point, here are a few example invocations. I encourage the reader to try predicting the output shapes (before and/or after reading the [documentation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.spearmanr.html)).

<details>

```python3
import numpy as np
from scipy import stats
x = np.arange(5)
z = np.stack((x, x, x))

### one-arg use ###
print(stats.spearmanr(z[0]).statistic.shape)

print(stats.spearmanr(z[0:1], axis=0).statistic.shape)
print(stats.spearmanr(z[0:1], axis=1).statistic.shape)
print(stats.spearmanr(z[0:1], axis=-1).statistic.shape)

print(stats.spearmanr(z[0:2], axis=0).statistic.shape)
print(stats.spearmanr(z[0:2], axis=1).statistic.shape)

print(stats.spearmanr(z, axis=1).statistic.shape)

### two-arg use ###
print(stats.spearmanr(z[0], z[0]).statistic.shape)
print(stats.spearmanr(z[0:1], z[0:1], axis=1).statistic.shape)
print(stats.spearmanr(z[0:2], z[0:2], axis=1).statistic.shape)
print(stats.spearmanr(z, z, axis=1).statistic.shape)
```

</details>

Now with the answers:

<details>

```python3
### one-arg use ###
print(stats.spearmanr(z[0]).statistic.shape)
# ValueError: `spearmanr` needs at least 2 variables to compare

print(stats.spearmanr(z[0:1], axis=0).statistic.shape)
# AttributeError: 'float' object has no attribute 'shape'
print(stats.spearmanr(z[0:1], axis=1).statistic.shape)
# IndexError: index 1 is out of bounds for axis 0 with size 1
print(stats.spearmanr(z[0:1], axis=-1).statistic.shape)
# IndexError: tuple index out of range

print(stats.spearmanr(z[0:2], axis=0).statistic.shape)
# AttributeError: 'float' object has no attribute 'shape'
print(stats.spearmanr(z[0:2], axis=1).statistic.shape)
# ()

print(stats.spearmanr(z, axis=1).statistic.shape)
# (3, 3)

### two-arg use ###
print(stats.spearmanr(z[0], z[0]).statistic.shape)
# ()
print(stats.spearmanr(z[0:1], z[0:1], axis=1).statistic.shape)
# ()
print(stats.spearmanr(z[0:2], z[0:2], axis=1).statistic.shape)
# (4, 4)
print(stats.spearmanr(z, z, axis=1).statistic.shape)
# (6, 6)
```

</details>

Here is my attempt at interpreting the results:

<details>

- In standard cases, here's what is *supposed* to happen:
  - No single-arg 1-D use
  - Two-arg 1-D is correlation between two univariate samples, a scalar.
  - Single-arg 2-D is the correlation between one multivariate sample and itself, a 2-D array. Alternatively, if one thinks of each variable as a separate sample, it produces all pairwise correlation coefficients between the samples.
  - I don't think two-arg 2-D computes the correlation between two multivariate samples. It is easiest to predict the result by stacking the two samples into one and thinking of the behavior for the single-arg 2-D case.
- Negative `axis` indices just don't work, leading to `IndexError: tuple index out of range`.
- When *any* element of the correlation coefficient matrix is undefined, a Python scalar NaN (rather than an array with NaNs in the right places) is returned. That is why `axis=0` produces a scalar above, and it's Python scalar rather than a NumPy scalar, leading to `'float' object has no attribute 'shape'`. These days, this is typically considered to be a bug both because the shape and `type` are different from what they would be following the usual rules.
- This still leaves exceptions, probably bugs:
  - `stats.spearmanr(z[0:1], z[0:1], axis=1).statistic.shape` and `stats.spearmanr(z[0:2], axis=1).statistic.shape` are `()` instead of `(2, 2)`.
  - `stats.spearmanr(z[0:1], axis=1)` fails, but not for the same reason that `stats.spearmanr(z[0], axis=1)` fails. I would have expected the result to be `[[1]]`.

</details>

Besides confusing users, all the assumptions about what different shapes mean adds to the code's complexity and makes it less flexible.

For instance, consider an implementation of the Spearman statistic like:
```python3
from scipy import stats

def spearmanrho(x, y, *, axis=-1):
    # assumes negative indices or arrays with same `ndim`
    x = stats.rankdata(x, axis=axis)
    y = stats.rankdata(y, axis=axis)
    return stats.pearsonr(x, y, axis=axis).statistic
```
This permits one to do the pairwise correlations above. For instance:
```python3
# stats.spearmanr(z, axis=1)
spearmanrho(z, z[:, None], axis=-1).shape  # (3, 3)
```
As written, this is a little less efficient, because we are not assuming the two arguments to have the same data, but we could always add special support for one-arg 2-D use if we think that's a good idea.

Besides simplicity and predictability, another benefit is that it doesn't *force* the user to calculate all pairwise correlations.
```python3
# currently impossible with stats.spearmanr
spearmanrho(z, z, axis=-1)  # (3,)
```
Of course, this generalizes to any number of dimensions, whereas the existing behavior is for 1-D and 2-D only.

Two other reasons to reconsider the behavior:
- There is no array API standard for `corrcoef`, which might have influenced the behavior of `spearmanr`.
- At this point, `spearmanr` `axis` behavior is inconsistent with all other stats functions, including the other correlation functions.

I see two paths forward that I'd be willing to work toward.
- Leave `spearmanr` alone (declare it "legacy"?) and create a new function `spearmanrho` with the new behavior (initially two-arg only).
- Deprecate (or emit `FutureWarning`s about) two-arg use of `spearmanr`. After two releases, introduce the new two-arg behavior. One-arg use would continue to be restricted to 2-D arrays and would be equivalent to `spearmanrho(z, z[:, None], axis=-1` (when `axis=1` or `axis=-1`; similar for `axis=0`), but it would avoid preprocessing `z` (sorting, de-meaning, etc.) twice. Bugs identified above would be fixed immediately.

LMK what you'd like to see.

Crossref gh-16133.

mdhaber:
#### Reference issue
Closes gh-23238
Toward gh-20544

Adds `scipy.stats.speamanrho`, an array API compatible replacement for `spearmanr`. Other advantages include:

- support for n-d arrays 
- standard broadcasting behavior
- support for `keepdims`
- the ability to compute exact p-values
- does not suffer from gh-22816, IIUC, or any of the old bug `spearmanr` reports I have found - some of which were not completely resolved for `spearmanr` (see below).

#### Additional information
~~Needs a post to the forum.~~ [Forum post](https://discuss.scientific-python.org/t/enh-stats-spearmanrho-array-api-compatible-substitute-for-spearmanr/2154) 10/24/2025.
I'd suggest marking `spearmanr` with the legacy admonition, but I'll leave that to the reviewer.

Re gh-22816, an open bug report:

<details>

I think the equivalent call is:
```python3
import numpy as np
from scipy import stats
X = np.zeros((4, 10))
X[2] = np.arange(10)
X[3] = -np.arange(10)
res = stats.spearmanrho(X[:, np.newaxis], X[np.newaxis, :], axis=-1)
```
We still get a `ConstantInputWarning` - I think that's approprate because there are going to be some NaNs in the output. But it does compute the rest of the correlation coefficients and p-values.
```python3
res.statistic[2:, 2:]
# array([[ 1., -1.],
#        [-1.,  1.]])
res.pvalue[2:, 2:]
# array([[1.06350359e-62, 1.06350359e-62],
       [1.06350359e-62, 1.06350359e-62]])
# array([[1.06350359e-62, 1.06350359e-62],
       [1.06350359e-62, 1.06350359e-62]])
```

</details>

#### Future work

- We addressed gh-17746 with documentation, but I think a `confidence_interval` method could quite easily be added to the function.

I'll take a look at some of the old bug reports/tests for `spearmanr` to make sure there are no regressions.

- [ ] gh-17089 - the reference that was added to close this issue is present in the new function's documentation
- [ ] gh-16148 - not sure if the PR that closed this fully addressed the (documentation) issue for `spearmanr`, and since the documentation of `spearmanrho` was derived from that of `spearmanr`, it might not be fixed in `spearmanrho`. I'll let the reviewer determine whether we need to add more details about the p-value calculation in the text. The asymptotic approximation is pretty standard, and the references are given.
- [ ] gh-13900 - see https://github.com/scipy/scipy/issues/13900#issuecomment-3441228806.
- [ ] gh-13391 - see https://github.com/scipy/scipy/issues/13391#issuecomment-3441266849
- [ ] gh-12458 - see https://github.com/scipy/scipy/issues/12458#issuecomment-3441306932
- [ ] gh-12411 - see https://github.com/scipy/scipy/issues/12411#issuecomment-3441326959
- [ ] gh-9414 - see https://github.com/scipy/scipy/issues/9414#issuecomment-3441340048
- [ ] gh-9396 - see https://github.com/scipy/scipy/issues/9396#issuecomment-3441367429
- [ ] gh-9103 - I'd rather not investigate another cases of incorrect NaN handling. I'll trust that it was fixed for `spearmanr`, and since the `_axis_nan_policy` decorator is used with `spearmanrho`, it's fixed for `spearmanrho`.
- [ ] gh-8585 - this was a reasonable enhancement request, and the same idea has come up several times. The fix is to make the p-value attribute a `property`. I think we should do that *across the board* in SciPy 2.0, when we stop returning unpackable tuple-bunches.
- [ ] gh-8111 - this was fixed a long time ago for `spearmanr`. It is easy to check that `spearmanrho` gets the same (correct) results.
- [ ] gh-7882 - We don't include all the same words in the revised documentation because I found that the part about being accurate only for samples with over 500 observations was unnecessary https://github.com/scipy/scipy/issues/7882#issuecomment-1257519860. We're clear that the default method is an asymptotic approximation, and we document (w/ an example) how the `method` argument can be used to perform an exact test.
- [ ] gh-7714 - See https://github.com/scipy/scipy/issues/7714#issuecomment-3441472911
- [ ] gh-6841 - not investigating any more `nan_policy`-related issues
- [ ] gh-6654 - not investigating any more `nan_policy`-related issues
- [ ] gh-6530 - not investigating any more `nan_policy`-related issues
- [ ] gh-6061 - this was fixed a long time ago for `spearmanr`. It is easy to check that `spearmanrho` gets the same (correct) results.
- [ ] gh-4965 - not surprising that there was an error in the description of the `spearmanr` `axis` parameter given its nonstandard behavior. The behavior of `spearmanrho` is the same as other stats reducing function, and the description of the `axis` parameter is the same as other stats reducing functions.
- [ ] gh-4221 - can't test this. https://github.com/scipy/scipy/issues/4221#issuecomment-417902888 suggests that it was fixed by a fix of `tiecorrect`, but `tiecorrect` is not used directly or indirectly (e.g. via `rankdata`) by `spearmanr` either now or [at that time](https://github.com/behzadnouri/scipy/blob/ae659664b6beb182dc3f18c69f587e1bdf52fbf8/scipy/stats/stats.py).
- [ ] gh-2371 - there are cases with ties in the tests. The correlation coefficient calculation is fixed in `spearmanr` and `spearmanrho`. Additionally, `spearmanrho` can compute exact p-values even when there are ties.
- [ ] gh-1786 - this is patched with a `np.errstate` in `spearmanr`. `spearmanrho` uses the beta distribution via `pearsonr`, which doesn't have this division by zero possibility. This can be confirmed using `x = [1, 2, 3, 4]`, which gives a statistic of `1.0` and p-value of `0.0` without a warning.
- [ ] gh-1627 - neither `spearmanrho` nor `spearmanr` use the formula that is invalid when there are ties
- [ ] gh-1349 - neither `spearmanrho` nor `spearmanr` use the formula that is invalid when there are ties

That was all the issues I saw with `spearmanr` in the title. ~~There are others that mention `spearmanr` in the text, but rather than going through all that, I'll just take a pass through the `spearmanr` tests. If no regression test was added, I guess the problem wasn't very important.~~ *Done.* 

`spearmanrho` seems to pass the important tests of `spearmanr` unmodified, except for those that depend on `spearmanr`'s nonstandard broadcasting or optional second argument behavior. I haven't tried modifying those to get all existing tests passing; not sure if it's needed.


scipy/stats/_correlation.py
```diff
@@ -236,3 +236,145 @@ def chatterjeexi(x, y, *, axis=0, y_continuous=False, method='asymptotic'):
...
+        If axis=None, then both arrays will be raveled.
+        Like other `scipy.stats` functions, `axis` is interpreted after the
+        arrays are broadcasted.
+
+    Returns
+    -------
+    res : SignificanceResult
+        An object containing attributes:
+
+        statistic : float or ndarray (2-D square)
```
jorenham: The "2-D square" is a bit confusing here, because in one of the examples there's a  `res.statistic` that's a vector of size 2. And this also supports >2-D results, no?

scipy/stats/_correlation.py
```diff
@@ -236,3 +236,145 @@ def chatterjeexi(x, y, *, axis=0, y_continuous=False, method='asymptotic'):
...
+            Spearman correlation coefficient
+        pvalue : float
```
jorenham: Judging by the example below, `pvalue` appears to same (d)type and shape as `statistic`, but this "pvalue : float" suggests that it'll always be a scalar.

jorenham: Speaking of scalars; those aren't really a thing in the array api, right? So shouldn't this return a 0-d array then, or are scalars considered equivalent for this purpose?
I suppose that it would be pretty easy to implement it, at least. E.g. by passing `res.statistic` and `res.pvalue` through `xp.asarray` before they're wrapped in a `SignificanceResult`.

Always returning arrays here would also make it easier for static typing, btw 😇. Because now the return type depends on the *shape* of the types, which isn't always statically known (usually it isn't known, and in case of non-numpy arrays it never is).

scipy/stats/_correlation.py
```diff
@@ -236,3 +236,145 @@ def chatterjeexi(x, y, *, axis=0, y_continuous=False, method='asymptotic'):
...
 
     return SignificanceResult(xi, pvalue)
+
+
+@xp_capabilities(cpu_only=True, exceptions=['jax.numpy'],
+    skip_backends=[('dask.array', 'not supported by rankdata (take_along_axis)')]
+)
+@_axis_nan_policy_factory(SignificanceResult, paired=True, n_samples=2,
+                          result_to_tuple=_unpack, n_outputs=2, too_small=1)
+def spearmanrho(x, y, *, alternative='two-sided', method=None, axis=0):
```
jorenham: could `x` and `y` be made positional-only; or would that be incompatible with the array api?

jorenham (commented):
This looks pretty useful, and the implementation is nice and clean 👌🏻

scipy/stats/_correlation.py
mdhaber: Yeah, accidentally copied that from the old one. This supports any dimensionality. *Update: adjusted documentation.*

scipy/stats/_correlation.py
mdhaber: Yeah also copied, but also always an array (or NumPy scalar), not Python float, like pretty much every other stats function.

scipy/stats/_correlation.py
mdhaber: Sure, I don't see why not. That would be *more* consistent with array API - the standard does that a lot. *Done.*

mdhaber:
Thanks @jorenham. Yeah, these are just the result of starting from old documentation and forgetting to update these things. Maybe suggest what you think should be written (since pretty much anything that is not super long is going to be somewhat imprecise). I tried to work towards fixing that (https://github.com/scientific-python/summit-2025/issues/41) but there were some present who disagreed with having a standard, so I put that on hold.

scipy/stats/_correlation.py
mdhaber: Ah and no, `stats` consistently ensures that 0d NumPy arrays are returned as NumPy scalars. "but that's wrong", etc.... Right. I'm doing my fair shair to get NumPy to change that (https://github.com/scientific-python/summit-2025/issues/38, https://github.com/mdhaber/numpy/pull/2); until then, `scipy.stats` is self-consistent and respects the preferred behavior of the array libraries it supports.

jorenham: Ah I see, nothing to do about that then I suppose. FWIW, we've talked about getting rid of scalar types in numpy a couple of times now, and it's definitely on our wishlist. There are no plans yet for how we're going to get there, though, besides starting with making ufuncs more transparent w.r.t. 0-d arrays.

mdhaber: Feel free to share your thoughts at https://github.com/mdhaber/numpy/pull/2, which proposes a NEP. Sebastian has done a fair amount of work exploring the idea, and a few projects have run their test suites against the resulting versions of NumPy. It would be good to revive that discussion.

*Update: adjusted documentation.*

steppi (approved):
I think this looks great. The only suggestion I have is to change the `xp_capabilities` entry for `spearmanr` to `xp_capabilities(out_of_scope=True)`. I'm in favor of merging, but it looks like no one else has responded to the discussion https://discuss.scientific-python.org/t/enh-stats-spearmanrho-array-api-compatible-substitute-for-spearmanr/2154/2.

mdhaber:
> The only suggestion I have is to change the xp_capabilities entry for spearmanr to xp_capabilities(out_of_scope=True)

How about I do that in a separate PR that marks a bunch of functions as out of scope?

> I'm in favor of merging, but it looks like no one else has responded to the discussion

Great. Since there is another approval above, I think that's pretty good lazy consensus. And if this is not backported to 1.17, folks have six more months to look at it. I can ping others to chime in on the mailing list if needed, though.

steppi:
> How about I do that in a separate PR that marks a bunch of functions as out of scope?

Sounds good to me.

> Great. Since there is another approval above, I think that's pretty good lazy consensus. And if this is not backported to 1.17, folks have six more months to look at it. I can ping others to chime in on the mailing list if needed, though.

I think I'm the only active maintainer who's approved so far and would like to see at least one more to establish lazy consensus. I know @jorenham has the same level of trust as a maintainer and could join the team if he wanted to. I think it shouldn't be hard to find someone else on the team to chime in though.

mdhaber:
OK. Since I still have 20 PRs open, I try to spare review requests where possible. @dschmitz89 do you have moment to weigh in? If not, LMK, and I can ask someone else.

jorenham (approved):

steppi:
> OK. Since I still have 20 PRs open, I try to spare review requests where possible. @dschmitz89 do you have moment to weigh in? If not, LMK, and I can ask someone else.

In the meantime I'll try to help reduce the number of  PRs you have open.

jorenham:
> I know @jorenham has the same level of trust as a maintainer

That's great to hear! I'm not a SciPy maintainer though :)
But either way, I gave it another look, and I like what I'm seeing here; nothing to complain about. So for what it's worth, I this looks good :ok_hand:.

steppi:
> > I know @jorenham has the same level of trust as a maintainer
> 
> That's great to hear! I'm not a SciPy maintainer though :) But either way, I gave it another look, and I like what I'm seeing here; nothing to complain about. So for what it's worth, I this looks good 👌.

Right, @mdhaber had said there was another approval, which I assumed was you (I didn't notice you hadn't formally approved yet). I was saying I'd like to not be the only actual maintainer approving an API change, so your approval didn't count on a formality.

ev-br (approved):

steppi:
Thanks @ev-br.

PR #23856, opened 2025-10-24
TST: sparse.linalg.tfqmr: avoid failure due to platform-dependent, intermittent divide by zero

mdhaber:
This attempts to resolve the failure in a test of `tfqmr` in the Accelerate workflow (e.g. [here](https://github.com/scipy/scipy/actions/runs/18784648906)). It seems that `tfqmr` is known to fail in the test case due to algorithmic limitations, so the test is checking for a particular error message. The test fails because there is a divide by zero warning before the desired error occurs. The proposed solution is to avoid failure due to the divide by zero warning; hopefully then the error is emitted as intended. It seems appropriate to silence the warning in just this test case because the problem is only beginning to occur now, and only with Accelerate, and only intermittently. Those with knowledge of `tfqmr` are welcome to investigate further, but we need to keep this from generating false positives in CI.

Note that lack of failure in this CI run does not mean that the problem is resolved since it is intermittent, but if we merge this and it does fail again, at least we'll have more information to go on.

### DIFF
--- a/scipy/sparse/linalg/_isolve/tests/test_iterative.py
+++ b/scipy/sparse/linalg/_isolve/tests/test_iterative.py
@@ -1,6 +1,7 @@
 """ Test functions for the sparse.linalg._isolve module
 """
 
+from contextlib import nullcontext
 import itertools
 import platform
 import pytest
@@ -542,7 +543,10 @@ def test_show(case, capsys):
     def cb(x):
         pass
 
-    x, info = tfqmr(case.A, case.b, callback=cb, show=True)
+    ctx = np.errstate(divide='ignore') if case.name == "nonsymposdef" else nullcontext()
+    with ctx:
+        x, info = tfqmr(case.A, case.b, callback=cb, show=True)
+
     out, err = capsys.readouterr()
 
     if case.name == "sym-nonpd":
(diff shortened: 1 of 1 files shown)


scipy/sparse/linalg/_isolve/tests/test_iterative.py
```diff
@@ -542,7 +543,10 @@ def test_show(case, capsys):
     def cb(x):
         pass
 
-    x, info = tfqmr(case.A, case.b, callback=cb, show=True)
+    ctx = np.errstate(divide='ignore') if case.name == "nonsymposdef" else nullcontext()
```
mdhaber:
```suggestion
    ctx = np.errstate(all='ignore') if case.name == "nonsymposdef" else nullcontext()
```

tylerjereddy:
Given the config of the CI job that you linked, I tried installing the nightly `scipy-1.17.0.dev0-cp313-cp313-macosx_14_0_x86_64.whl` under emulation on ARM Mac inside of a Universal2 CPython `3.13.9` install, and using `pytest-repeat` to see if could replicate the flaky failure with enough repeats: `python -m pytest -n 10 --pyargs scipy.sparse.linalg._isolve.tests.test_iterative::test_show --count=500`, but no luck replicating: `11500 passed, 10 warnings in 26.18s`. The wheel should be using `Accelerate` if `show_config()` is to be believed.

tylerjereddy:
Let's give this a try. I couldn't reproduce it, but suppressing a warning to confirm an error outcome seems ok-ish to me (compared to say, suppressing a warning that still ends up in a "success" result).

PR #23836, opened 2025-10-21
ENH/DEP: sparse.linalg.LinearOperator: n-D batch support

lucascolley:
#### Reference issue
Towards gh-23315.
Closes gh-8444.

https://discuss.scientific-python.org/t/batching-linearoperator-status-and-benchmarks/2199/1

#### What does this implement/fix?

Support for n-dimensional batches of linear operators in `sparse.linalg.LinearOperator`.

#### Deprecation

This PR introduces a `FutureWarning` for calling `{r}matvec` on column vectors. That API does not extend cleanly to batch dimensions, but identical behaviour can be achieved (and extended to batch dimensions) via `{r}matmat`.

#### Future work

Immediate follow-up (me): gh-24562

What I plan to work on:
- array API standard support (including take another look at pydata/sparse)
- supporting these batched operators in the iterative solvers of this submodule

Potential follow-ups that might be nice (but I don't have plans to work on) based on comments here and TODOs in the code:
- taking a look at `_CustomLinearOperator` and the `__init__`/`__new__` dance from the perspective of idiomatic OOP and typing practices in the present day
- performance optimisation when the user's `matvec` supports batching: https://github.com/lucascolley/scipy/blob/53eb4644c3dd8280f7c6dd8294e1c122a1e908b6/scipy/sparse/linalg/_interface.py#L255-L260
- purging `np.matrix` references from the docs and deprecating `np.matrix` support (probably SciPy 2.0 effort)
- some sort of diagram for the User Guide explaining how the various methods interact depending on how classes are instantiated: https://github.com/scipy/scipy/pull/23836#discussion_r2764807854
- API overhaul
	- renaming `matmat` to `matmul`
	- {`.T`, `.mT`, `.H`, `.mH`}, removing {`transpose`,` adjoint`}
	- getting rid of `dot` and `rdot`
	- restricting `*` to scalar-input-only rather than it being a superset of `@`?
- https://github.com/scipy/scipy/pull/24157 would enable some minor simplifications

PR #23860, opened 2025-10-25
TST: sparse.linalg.LinearOperator: add dot-tests

lucascolley:
This PR adds new tests for `LinearOperator`, inspired by similar tests in https://github.com/PyLops/pylops, to verify that forward and adjoint operators match for random data. It extends the concept of 'dot-tests' from PyLops to add API coverage for Python operators (`@` and `*`), `{r}matmat`, and `dot`.

Currently tested are 3 relatively simple linear operators:
- an identity operator for square matrices, which returns the input unmodified, defined by `matvec` and `rmatvec`
- an identity operator capable of handling under/overdetermined systems, with appropriate reshaping, defined by `matvec` and `rmatvec`
- a complex scaling operator for square matrices, defined by `matvec` and `rmatvec`

In this PR it would be good to also cover:
- [ ] an operator defined by `matmat` and `rmatmat`
- [ ] an operator defined via subclassing `LinearOperator`, perhaps implementing `_adjoint` and not `_rmatvec`.

We may want to look at extending coverage for more complex operators in this PR or in a follow-up. Aside from that, this is ready for review!

---

The motivation here is to improve test coverage before gh-23836. Alongside that PR, these tests can be extended to verify correct behaviour in the case of batches too.

### DIFF
--- a/scipy/sparse/linalg/tests/test_interface.py
+++ b/scipy/sparse/linalg/tests/test_interface.py
@@ -4,9 +4,10 @@
 from functools import partial
 from itertools import product
 import operator
+from typing import NamedTuple
 import pytest
 from pytest import raises as assert_raises, warns
-from numpy.testing import assert_, assert_equal
+from numpy.testing import assert_, assert_equal, assert_allclose
 
 import numpy as np
 import scipy.sparse as sparse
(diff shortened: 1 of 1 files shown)


scipy/sparse/linalg/tests/test_interface.py
```diff
@@ -214,6 +215,154 @@ def test_matmul(self):
...
+        x = np.ones(par.nx) + par.imag * np.ones(par.nx)
+        y = Iop * x
+        x1 = Iop.H * y
+    
+        assert_allclose(
+            x[: min(par.ny, par.nx)], y[: min(par.ny, par.nx)]
+        )
+        assert_allclose(
+            x[: min(par.ny, par.nx)], x1[: min(par.ny, par.nx)]
+        )
```
lucascolley: perhaps this could be rolled into `dottest` too?

scipy/sparse/linalg/tests/test_interface.py
```diff
@@ -214,6 +215,154 @@ def test_matmul(self):
...
+        y = Op.matvec(u)  # Op * u
+        x = Op.rmatvec(v)  # Op'* v
+    
+        if getattr(Op, "clinear", True):
+            yy = np.vdot(y, v)  # (Op  * u)' * v
+            xx = np.vdot(u, x)  # u' * (Op' * v)
+        else:
+            # Op is only R-linear, so treat complex numbers as elements of R^2
+            yy = np.dot(y.real, v.real) + np.dot(y.imag, v.imag)
+            xx = np.dot(u.real, x.real) + np.dot(u.imag, x.imag)
```
lucascolley: some of this can probably be simplified to remove handling for cases which don't exist in SciPy

lucascolley:
by the way, what does `par` stand for?

scipy/sparse/linalg/tests/test_interface.py
mrava87: No this is a check specific to our implementation of the identity... I would say passing dot-test is a good indication this is done correctly but this is a more direct check

scipy/sparse/linalg/tests/test_interface.py
mrava87: I guess so

mrava87:
@lucascolley this makes sense to me but I am not a scipy maintainer so you probably want some from the scipy project to see this and get their feedback too 😄 

For `matmat` we really do not check much as right now in PyLops (and I think Scipy) is just a naive for loop over `matvec`, so if matvec passes it is obvious matmat would pass.. but if you plan to create ad-hoc implementations for `matmat` then having something like a "batched dot-test" would start to make more sense

lucascolley:
Thanks! Alright, sounds like tidying this up and extending coverage for more operators is a good idea.

What does `par` stand for, by the way?

> For `matmat` we really do not check much as right now in PyLops (and I think Scipy) is just a naive for loop over `matvec`, so if matvec passes it is obvious matmat would pass.. but if you plan to create ad-hoc implementations for `matmat` then having something like a "batched dot-test" would start to make more sense

Yes, I think having coverage of `matmat` and `rmatmat` would be valuable too, to ensure that my changes are correct when batching is added. `dot` is in the SciPy API too, so we should cover that as well.

scipy/sparse/linalg/tests/test_interface.py
```diff
@@ -214,6 +215,129 @@ def test_matmul(self):
...
+        y = Op.matvec(u)  # Op * u
+        x = Op.rmatvec(v)  # Op'* v
+    
+        yy = np.vdot(y, v)  # (Op  * u)' * v
+        xx = np.vdot(u, x)  # u' * (Op' * v)
+    
+        # evaluate if dot test passed
+        passed = np.isclose(xx, yy, rtol, atol)
+    
+        # verbosity or error raising
```
lucascolley:
```suggestion
```

scipy/sparse/linalg/tests/test_interface.py
```diff
@@ -214,6 +215,129 @@ def test_matmul(self):
...
+        if complexflag:
+            v = v + 1j * np.random.randn(nr).astype(rdtype)
+    
+        y = Op.matvec(u)  # Op * u
+        x = Op.rmatvec(v)  # Op'* v
+    
+        yy = np.vdot(y, v)  # (Op  * u)' * v
+        xx = np.vdot(u, x)  # u' * (Op' * v)
+    
+        # evaluate if dot test passed
```
lucascolley:
```suggestion
```

mrava87:
par simply stands for parameters

lucascolley:
this PR is ready for review! I updated the top-post with a proper PR description.

@mrava87 do you have any suggestions for other linear operators which would be good to define and test here?

scipy/sparse/linalg/tests/test_interface.py
```diff
@@ -214,6 +215,252 @@ def test_matmul(self):
...
+        assert_allclose(X, Op.H @ V)
+        
+        # check dot equivalence
+        assert_allclose(Y, Op.dot(U))
+        assert_allclose(X, Op.H.dot(V))
+
+        YY = np.trace(Y.conj().T @ V)
+        XX = np.trace(U.conj().T @ X)
+
+        passed = np.isclose(XX, YY, rtol, atol)
```
lucascolley: could be that we want to check full equivalence here, not just the trace

scipy/sparse/linalg/tests/test_interface.py
```diff
@@ -214,6 +215,252 @@ def test_matmul(self):
...
+        def scale(x):
+            return (3 + 2j) * x
+        def r_scale(x):
+            return (3 - 2j) * x
+        ScaleOp = interface.LinearOperator(
+            shape=(args.ny, args.nx), dtype=args.dtype, matvec=scale, rmatvec=r_scale
+        )
+        assert self.dottest(ScaleOp, complexflag=args.complex)
+        assert self.matmultest(ScaleOp, complexflag=args.complex)
+
```
lucascolley, line 447: we should include a linear operator defined in terms of solely `{r}matmat` in this PR too

scipy/sparse/linalg/tests/test_interface.py
```diff
@@ -214,6 +215,252 @@ def test_matmul(self):
...
+            raise AssertionError(msg)
+
+        return True
+
+    @pytest.mark.parametrize("args", [args1, args1j])
+    def test_identity_square(self, args):
+        """Simple identity operator on square matrices"""
+        def identity(x):
+            return x
+        Iop = interface.LinearOperator(
```
lucascolley, line 391: it may be good to also test operators defined via subclassing rather than `_CustomLinearOperator.__init__`

scipy/sparse/linalg/tests/test_interface.py
```diff
@@ -214,6 +215,237 @@ def test_matmul(self):
...
+        complex: bool
+        dtype: str
+        
+    args1: OpArgs = OpArgs(11, 11, False, "int32")  # real integer square
+    args2: OpArgs = OpArgs(21, 11, False, "float64")  # overdetermined real
+    args1j: OpArgs = OpArgs(11, 11, True, "complex128")  # square complex
+    args2j: OpArgs = OpArgs(21, 11, True, "complex128")  # overdetermined complex
+    args3: OpArgs = OpArgs(11, 21, False, "float64")  # underdetermined real
+    
+    def dottest(
```
mrava87, line 231: If you plan to use this widely, I would not bury it in a test, rather move it to some utility file

scipy/sparse/linalg/tests/test_interface.py
```diff
@@ -214,6 +215,237 @@ def test_matmul(self):
...
+
+        v = rng.standard_normal(nr).astype(rdtype)
+        if complexflag:
+            v = v + 1j * rng.standard_normal(nr).astype(rdtype)
+
+        op_u = Op.matvec(u)
+        opH_v = Op.rmatvec(v)
+
+        # check operator equivalence
+        assert_allclose(op_u, Op * u)
```
mrava87, line 296: not sure this makes sense as part of the dot test method... I would test this on its own...

mrava87 (commented):
Looks very good to me, but I must say I am not a Scipy developer/maintainer so I am not sure if this follows the standard of their tests (I assume it does) and whether they are happy to add tests on hand-picked operators which do not officially belong to their public API but are created directly in the test suite... 

I guess if the value of doing this is to ensure that the "default" `matmat` and `rmatmat` methods work, makes sense to me 😄

j-bowhay:
PyLops is LGPL licensed, is that compatible with our license as I assume this code is derived from PyLops? I know it is just a test but we distribute tests so I assume it matters?

lucascolley:
> PyLops is LGPL licensed, is that compatible with our license as I assume this code is derived from PyLops? I know it is just a test but we distribute tests so I assume it matters?

Oh great point, I didn't notice that yet.

@mrava87 would you be willing/able to release https://github.com/PyLops/pylops/blob/dev/pylops/utils/dottest.py under a license compatible with BSD-3-Clause?

mrava87:
No, we can’t change PyLops licence. What do you mean for release a single file? Would putting a file in an independent repository work as the same file will still be in the PyLops repo with the LGPL licence. 

Btw, the concept of dot-test is well known, what we do in that file is nothing new; you could just re-implement it yourself with different design choices than the ones we make in PyLops… that would make everything easier 😉

lucascolley:
@mrava87 written permission from all copyright holders of code from which this PR derives is sufficient. I just looked through the commit history of the dottest module, and I'm confident that only your commits are relevant. See e.g. https://stackoverflow.com/a/11455485 're-licensing' section for an explanation. This doesn't affect anyone using the code via PyLops.

There is precedent for this in the project, e.g. https://github.com/scipy/scipy/blob/main/scipy/stats/biasedurn/license.txt.

ilayn:
> @mrava87 written permission from all copyright holders of code from which this PR derives is sufficient. I just looked through the commit history of the dottest module, and I'm confident that only your commits are relevant.

That's not enough. All contributors of the project needs confirmation. But regardless, there is no need for this dance if it can be reimplemented from scratch. We don't need the risk for a task like testing existing code.

You are already in the risk of derived work clause since you already read the code. So let's not come closer to any variant of GPL code.

lucascolley:
as you wish, I'll look at rewriting from scratch soon

lucascolley:
new PR at gh-23969

Issue #19889, opened 2024-01-16
BUG: stats.rankdata: dtype instability


rgommers:
Following up on the issue with `rankdata` in https://github.com/scipy/scipy/pull/19797#issuecomment-1888863426 - the immediate failing test issue is addressed, however the `rankdata` behavior doesn't look good. The dtype of the returned array should not be depending on both the value of the `method` keyword and the values in the array (it's always `float64` if `nan`s are present, and it may be integer if no `nan`s).

The docstring examples show the issue pretty clearly:
```python
>>> rankdata([0, 2, 3, 2])
array([ 1. ,  2.5,  4. ,  2.5])
>>> rankdata([0, 2, 3, 2], method='min')
array([ 1,  2,  4,  2])
>>> rankdata([0, 2, 3, 2], method='max')
array([ 1,  3,  4,  3])
>>> rankdata([0, 2, 3, 2], method='dense')
array([ 1,  2,  3,  2])
>>> rankdata([0, 2, 3, 2], method='ordinal')
array([ 1,  2,  4,  3])
>>> rankdata([[0, 2], [3, 2]]).reshape(2,2)
array([[1. , 2.5],
      [4. , 2.5]])
>>> rankdata([[0, 2, 2], [3, 2, 5]], axis=1)
array([[1. , 2.5, 2.5],
       [2. , 1. , 3. ]])
>>> rankdata([0, 2, 3, np.nan, -2, np.nan], nan_policy="propagate")
array([nan, nan, nan, nan, nan, nan])
>>> rankdata([0, 2, 3, np.nan, -2, np.nan], nan_policy="omit")
array([ 2.,  3.,  4., nan,  1., nan])
```

This looks bad enough that it can be considered a bug I think (and it was definitely buggy until 1.12 when it didn't even adhere to the rules as shown by https://github.com/scipy/scipy/pull/19797#issuecomment-1888863426; that was improved in the recent rewrite in gh-19776). The easiest solution would be to always return a `float64` array.

mdhaber:
> The easiest solution would be to always return a float64 array.

That's fine with me; I was just maintaining the old behavior of returning ints for methods that will always have integer ranks for non-nan inputs. IIRC there were existing tests to enforce that. You want me to open a PR that ensures that the output is always `float64`?

rgommers:
Thanks, that would be helpful. Maybe let's give it a day at least to see if anyone else sees a problem? 

[...]

PR #23862, opened 2025-10-25
ENH: stats.kruskal: add array API support

mdhaber:
#### Reference issue
Toward gh-20544

Adds array API support to `scipy.stats.kruskal`.

#### Additional information
I think when we were converting `rankdata` we respected the original dtype choices, but those were never ideal (gh-19889). It might be worth taking another look at that.


j-bowhay (approved):
LGTM

PR #23863, opened 2025-10-26
ENH: special: add handling of double type for `_gen_harmonic`, `_normalized_gen_harmonic`

### LINKED ISSUE #23678: BUG: stats: custom distribution generated from `zipfian` has problem at support endpoint
### Describe your issue.

If I use `stats.make_distribution(stats.zipfian)` and create an instance of this distribution, the `cdf()` method generates a warning when evaluated at the upper end of the support:

```
In [1]: from scipy import stats

In [2]: Zipfian = stats.make_distribution(stats.zipfian)

In [3]: zdist = Zipfian(a=0.75, n=15)

In [4]: zdist.cdf(15)
/Users/warren/repos/git/forks/scipy/build-install/usr/lib/python3.13/site-packages/scipy/stats/_discrete_distns.py:1426: RuntimeWarning: invalid value encountered in cast
  k = np.floor(k).astype(np.int_)
Out[4]: np.float64(1.0)
```
I was worried that this was caused by https://github.com/scipy/scipy/pull/23537, so I tried the same thing with the commit previous to that change, and got errors instead of a warning:

```
In [1]: import scipy

In [2]: scipy.__version__
Out[2]: '1.17.0.dev0+git20250925.21b945b'

In [3]: from scipy import stats

In [4]: Zipfian = stats.make_distribution(stats.zipfian)

In [5]: zdist = Zipfian(a=0.75, n=15)

In [6]: zdist.cdf(15)
/Users/warren/repos/git/forks/scipy/build-install/usr/lib/python3.13/site-packages/scipy/stats/_discrete_distns.py:1374: RuntimeWarning: All-NaN slice encountered
  n_max = np.nanmax(n)  # loop starts at maximum of all n
Cell In[6], line 1
----> 1 zdist.cdf(15)

File ~/repos/git/forks/scipy/build-install/usr/lib/python3.13/site-packages/scipy/stats/_distribution_infrastructure.py:2631, in UnivariateDistribution.cdf(self, x, y, method)
   2629 def cdf(self, x, y=None, /, *, method=None):
   2630     if y is None:
-> 2631         return self._cdf1(x, method=method)
   2632     else:
   2633         return self._cdf2(x, y, method=method)

File ~/repos/git/forks/scipy/build-install/usr/lib/python3.13/site-packages/scipy/stats/_distribution_infrastructure.py:1022, in _set_invalid_nan.<locals>.filtered(self, x, *args, **kwargs)
   1019     x = np.array(x, dtype=dtype, copy=True)
   1020     x[mask_invalid] = np.nan
-> 1022 res = np.asarray(f(self, x, *args, **kwargs))
   1024 # Ensure that the result is the correct dtype and shape,
   1025 # copying (only once) if necessary.
   1026 res_needs_copy = False

File ~/repos/git/forks/scipy/build-install/usr/lib/python3.13/site-packages/scipy/stats/_distribution_infrastructure.py:2698, in UnivariateDistribution._cdf1(self, x, method)
   2696 @_set_invalid_nan
   2697 def _cdf1(self, x, *, method):
-> 2698     return self._cdf_dispatch(x, method=method, **self._parameters)

File ~/repos/git/forks/scipy/build-install/usr/lib/python3.13/site-packages/scipy/stats/_distribution_infrastructure.py:1160, in _dispatch.<locals>.wrapped(self, method, *args, **kwargs)
   1157         self._method_cache[func_name] = method
   1159 try:
-> 1160     return method(*args, **kwargs)
   1161 except KeyError as e:
   1162     raise NotImplementedError(self._not_implemented) from e

File ~/repos/git/forks/scipy/build-install/usr/lib/python3.13/site-packages/scipy/stats/_discrete_distns.py:1454, in zipfian_gen._cdf(self, k, a, n)
   1452 def _cdf(self, k, a, n):
   1453     k = np.floor(k)
-> 1454     return _gen_harmonic(k, a) / _gen_harmonic(n, a)

File ~/repos/git/forks/scipy/build-install/usr/lib/python3.13/site-packages/scipy/stats/_discrete_distns.py:1387, in _gen_harmonic(n, a)
   1385 """Generalized harmonic number"""
   1386 n, a = np.broadcast_arrays(n, a)
-> 1387 return xpx.apply_where(a > 1, (n, a), _gen_harmonic_gt1, _gen_harmonic_leq1)

File ~/repos/git/forks/scipy/build-install/usr/lib/python3.13/site-packages/scipy/_lib/array_api_extra/_lib/_funcs.py:143, in apply_where(***failed resolving arguments***)
    141     # map_blocks doesn't descend into tuples of Arrays
    142     return xp.map_blocks(_apply_where, cond, f1, f2, fill_value, *args_, xp=meta_xp)
--> 143 return _apply_where(cond, f1, f2, fill_value, *args_, xp=xp)

File ~/repos/git/forks/scipy/build-install/usr/lib/python3.13/site-packages/scipy/_lib/array_api_extra/_lib/_funcs.py:170, in _apply_where(cond, f1, f2, fill_value, xp, *args)
    168 else:
    169     ncond = ~cond
--> 170     temp2 = f2(*(arr[ncond] for arr in args))
    171     dtype = xp.result_type(temp1, temp2)
    172     out = xp.empty_like(cond, dtype=dtype)

File ~/repos/git/forks/scipy/build-install/usr/lib/python3.13/site-packages/scipy/stats/_discrete_distns.py:1377, in _gen_harmonic_leq1(n, a)
   1375 out = np.zeros_like(a, dtype=float)
   1376 # add terms of harmonic series; starting from smallest to avoid roundoff
-> 1377 for i in np.arange(n_max, 0, -1, dtype=float):
   1378     mask = i <= n  # don't add terms after nth
   1379     out[mask] += 1/i**a[mask]

ValueError: arange: cannot compute length
```

<details>
<summary>Details</summary>

```shell
1.17.0.dev0+git20250925.21b945b 2.3.1 sys.version_info(major=3, minor=13, micro=5, releaselevel='final', serial=0)
/Users/warren/repos/git/forks/scipy/build-install/usr/lib/python3.13/site-packages/scipy/__config__.py:164: UserWarning: Install `pyyaml` for better output
{
  "Compilers": {
    "c": {
      "name": "clang",
      "linker": "ld64",
      "version": "18.1.8",
      "commands": "/Users/warren/install/clang-llvm-18.1.8/clang+llvm-18.1.8-arm64-apple-macos11/bin/clang",
      "args": "-isysroot, /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk",
      "linker args": "-L/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/lib, -isysroot, /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk"
    },
    "cython": {
      "name": "cython",
      "linker": "cython",
      "version": "3.1.2",
      "commands": "cython"
    },
    "c++": {
      "name": "clang",
      "linker": "ld64",
      "version": "14.0.3",
      "commands": "c++",
      "args": "-isysroot, /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk",
      "linker args": "-L/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/lib, -isysroot, /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk"
    },
    "fortran": {
      "name": "gcc",
      "linker": "ld64",
      "version": "12.2.0",
      "commands": "gfortran",
      "linker args": "-L/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/lib"
    },
    "pythran": {
      "version": "0.18.0",
      "include directory": "../../../../../py3.13.5/lib/python3.13/site-packages/pythran"
    }
  },
  "Machine Information": {
    "host": {
      "cpu": "aarch64",
      "family": "aarch64",
      "endian": "little",
      "system": "darwin"
    },
    "build": {
      "cpu": "aarch64",
      "family": "aarch64",
      "endian": "little",
      "system": "darwin"
    },
    "cross-compiled": false
  },
  "Build Dependencies": {
    "blas": {
      "name": "openblas",
      "found": true,
      "version": "0.3.28.dev",
      "include directory": "/Users/warren/local/include",
      "lib directory": "/Users/warren/local/lib",
      "openblas configuration": "USE_64BITINT= DYNAMIC_ARCH= DYNAMIC_OLDER= NO_CBLAS= NO_LAPACK= NO_LAPACKE= NO_AFFINITY=1 USE_OPENMP= VORTEX MAX_THREADS=8",
      "pc file directory": "/Users/warren/local/lib/pkgconfig"
    },
    "lapack": {
      "name": "openblas",
      "found": true,
      "version": "0.3.28.dev",
      "include directory": "/Users/warren/local/include",
      "lib directory": "/Users/warren/local/lib",
      "openblas configuration": "USE_64BITINT= DYNAMIC_ARCH= DYNAMIC_OLDER= NO_CBLAS= NO_LAPACK= NO_LAPACKE= NO_AFFINITY=1 USE_OPENMP= VORTEX MAX_THREADS=8",
      "pc file directory": "/Users/warren/local/lib/pkgconfig"
    },
    "pybind11": {
      "name": "pybind11",
      "version": "3.0.1",
    }
  },
  "Python Information": {
    "path": "/Users/warren/py3.13.5/bin/python3.13",
    "version": "3.13"
  }
}
```

</details>

WarrenWeckesser:
The ufuncs _gen_harmonic and _normalized_gen_harmonic now have native loops for type double for the parameters that are normally integers.

This allows removal of the integer cast in the methods of stats.zipfian.

To avoid the possibility of infinite loops and other numerical problems, the `n` parameter of stats.zipfian must now be less than 2**53.

Native loops for `int32` are also added, so these can handle integer arrays on platforms where the default integer type is `int32` without have to upcast to `int64`.

Closes gh-23678.


mdhaber (commented):
Thanks, this is really helpful! I'm not really qualified to give a review of the compiled stuff, but I'll check out the Python things.

WarrenWeckesser:
The one test failure is https://github.com/scipy/scipy/issues/23859.

scipy/stats/tests/test_continuous.py
```diff
@@ -2196,3 +2196,12 @@ def test_inverse(self):
         np.testing.assert_allclose(X.iccdf(p), X0.iccdf(p))
         np.testing.assert_allclose(X.ilogcdf(p), X0.ilogcdf(p))
         np.testing.assert_allclose(X.ilogccdf(p), X0.ilogccdf(p))
+
+
+def test_zipfian_distribution_wrapper():
```
mdhaber, line 2201: This test would fail in SciPy currently and is adequate to protect against an obvious regression.

scipy/special/tests/test_gen_harmonic.py
mdhaber, line 1: This parametrizes all existing tests of `_gen_harmonic` and `_normalized_gen_harmonic` w.r.t. dtype. It also adds tests that the functions return NaN for NaN input. LGTM.

scipy/stats/_discrete_distns.py
mdhaber, line 1: These all look correct. The limitation on `n` is fine. I can see it being documented in the notes of the distribution, but I don't think users should be *too* surprised if they get NaNs when using a distribution with outrageous values of the parameters, and they can let us know if it's a blocker for them. In this case, they can switch over to `zipf` without too much loss in accuracy (depending on `a`).

mdhaber (approved):
Looks OK on the stats side of things. The test failure is indeed unrelated.

WarrenWeckesser:
I moved this PR to *draft* status.  I need to look more closely at how, when and where warnings are generated when an extremely large value of `n` is given.

I'll also add a note in the class docstring about the maximum allowed value of `n`.

WarrenWeckesser:
If this is merged, please squash-merge, and clean up the commit message by keeping just the message from the first commit.  The FIXUP messages are for the benefit of reviewers, and do not need to be included in the squashed commit message.

WarrenWeckesser:
There is a new unrelated test failure:

```
FAILED interpolate/tests/test_bary_rational.py::TestAAA::test_basic_functions[<lambda>-3.5e-12-0]
```

WarrenWeckesser:
Reran the tests with the unrelated failures, all passing now.

scipy/special/tests/test_gen_harmonic.py
```diff
@@ -33,11 +34,12 @@
...
      (75, -1.5, 19811.38815892374)]
 )
-def test_gen_harmonic(n, a, ref):
-    h = _gen_harmonic(n, a)
+def test_gen_harmonic(typ, n, a, ref):
+    h = _gen_harmonic(typ(n), a)
     assert_allclose(h, ref, rtol=5e-15)
 
 
+@pytest.mark.parametrize('typ', [np.int64, np.float64])
```
dschmitz89: What about `int32` and `float32`? I remember these distributions being problematic there. Does the distribution infra upcast this before?

mdhaber: The old infrastructure does upcast - at least most of the time. (Not sure whether it's perfectly consistent.) The new infrastructure itself tries to respects float dtype of input parameters[^1], but IIUC some special functions don't have a real float32 implementation, and that makes sense because sometimes output precision does not scale nicely with working precision, at least not automatically. So if the special function infrastructure either converts back and forth or raises an error given float32 input, that's ok with me.

[^1]: instances have a fixed dtype determined by the shape parameters, and the output dtype of a method is the result dtype of the instance dtype and the method argument dtype. If the distribution doesn't accepts parameters, the instance dtype is currently `float64`. Whenever there is interest, I planned to add a way to control the instance type more explicitly or allow it to be `None` so the output type of methods only depends on method arguments.

WarrenWeckesser: > What about `int32` and `float32`?

Right, they will be upcast to `int64` and `float64`, respectively.  Adding native handling of `int32` is a trivial enhancement for the private ufuncs `_gen_harmonic` and `_normalized_gen_harmonic`, and now I don't see why they shouldn't do it, so I just pushed a change to add support for native `int32`.

Handling `float32` natively would require some additional work, but it would not be impossible.  That might also require some tweaks in the `zipfian` distribution methods to maintain `float64` output even when all floating point inputs are `float32`, because that is the current behavior:

```
In [8]: import scipy

In [9]: scipy.__version__
Out[9]: '1.16.2'

In [10]: from scipy.stats import zipfian

In [11]: k = np.array([1, 2, 4, 8, 10], dtype=np.float32)

In [12]: a = np.float32(0.75)

In [13]: n = np.float32(25)

In [14]: zipfian.cdf(k, a=a, n=n).dtype  # All inputs are float32
Out[14]: dtype('float64')
```
Or we would have to accept a slight backwards incompatible API change and let the methods return `float32` in that case.  At the moment I'm OK with the status quo (i.e. cast to `float64`, consistent with the behavior of many other distributions).

dschmitz89:
Thanks @WarrenWeckesser .

PR #23864, opened 2025-10-26
TST: stats.boxcox_llf: bump test tolerance

mdhaber:
#### Reference issue
Closes gh-23859

Resolves test failure observed in CI by bumping the test tolerance.

#### Additional information
~~Accelerate seems to treat the rules of arithmetic as suggestions. Can we detect the platform in `xp_assert_close` and use that to bump tolerances for Accelerate across the board?~~ Oops, _this_ one isn't Accelerate.

### DIFF
--- a/scipy/stats/tests/test_morestats.py
+++ b/scipy/stats/tests/test_morestats.py
@@ -2025,7 +2025,7 @@ class TestBoxcox_llf:
         # The expected value was computed with mpsci, set mpmath.mp.dps=100
         # expect float64 output for integer input
         xp_assert_close(llf, xp.asarray(-15.32401272869016598, dtype=xp.float64),
-                        rtol=1e-7)
+                        rtol=5e-7)  # bumped tolerance from 1e-7 for Accelerate
 
     def test_axis(self, xp):
         data = xp.asarray([[100, 200], [300, 400]])


scipy/stats/tests/test_morestats.py
```diff
@@ -2025,7 +2025,7 @@ def test_instability_gh20021(self, xp):
         # The expected value was computed with mpsci, set mpmath.mp.dps=100
         # expect float64 output for integer input
         xp_assert_close(llf, xp.asarray(-15.32401272869016598, dtype=xp.float64),
-                        rtol=1e-7)
+                        rtol=5e-7)  # bumped tolerance from 1e-7 for Accelerate
```
mdhaber, line 2028: Should be sufficient to address:
```
Mismatched elements: 1 / 1 (100%)
Max absolute difference: 2.38803211e-06
Max relative difference: 1.55835952e-07
 x: array(-15.32401)
 y: array(-15.324013)
```

mdhaber:
```suggestion
                        rtol=5e-7)
```

dschmitz89:
Oops that merge might have been a bit early, hope accelerate CI comes comes back green.

mdhaber:
Turns out it actually wasn't Accelerate that was failing on this one. It was one of the Prerelease deps workflows, and those passed. Thanks @dschmitz89!

PR #23865, opened 2025-10-26
MAINT/TST: stats.yeojohnson_llf/boxcox_llf: fix/test nan_policy/keepdims of public functions

mdhaber:
#### Reference issue
https://github.com/scipy/scipy/pull/23807#discussion_r2461320633

https://github.com/scipy/scipy/pull/23807#discussion_r2461320633 caught missing quotes that are needed to make `boxcox_llf` and `yeojohnson_llf` respect their `nan_policy` and `keepdims` arguments. The tests didn't catch the problem because they were run against the private functions (called almost immediately by the public function for technical reasons).
This fixes the bug in `boxcox_llf` (already fixed in `yeojohnson_llf` and adjusts the tests so they would have caught the issue.

#### Additional information
`boxcox_llf` bug (ignoring the new `nan_policy` and `keepdims` arguments) present in SciPy 1.16.2. We could backport the fix, but it wasn't a regression because the features were new, and nobody has reported the issue, so I'm not sure if it's critical.

### DIFF
--- a/scipy/stats/_morestats.py
+++ b/scipy/stats/_morestats.py
@@ -996,9 +996,9 @@ def boxcox_llf(lmb, data, *, axis=0, keepdims=False, nan_policy='propagate'):
     # unless necessary.
     kwargs = {}
     if keepdims is not False:
-        kwargs[keepdims] = keepdims
+        kwargs['keepdims'] = keepdims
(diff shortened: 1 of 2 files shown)


j-bowhay (approved):
Thanks for the quick fix!

PR #23871, opened 2025-10-27
DOC: stats: Fix some math markup in the boxcox docstring.

WarrenWeckesser:
The `cases` separator `\\` was missing.


WarrenWeckesser:
[Before](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.boxcox.html) and [after](https://output.circle-artifacts.com/output/job/f350abe7-c2fb-4272-a987-db254effa72f/artifacts/0/html/reference/generated/scipy.stats.boxcox.html) (scroll down to the **Notes** section).

PR #23869, opened 2025-10-26
ENH: spatial.transform.Rotation: Add shape to `random` and `identity`

amacati:
Implements the `shape` kwarg-only argument for `Rotation.random` and `Rotation.identity`.

#### Reference issue
Contributes towards #23391

#### What does this implement/fix?
Adds the shape argument to `Rotation.random` and `Rotation.identity`.

#### Additional information
We should decide what to do about deprecations. In its current form, shape could replace num easily since it can be both a `tuple[int, ...]` as well as an `int`.

PR #23875, opened 2025-10-27
ENH: spatial.transform.RigidTransform: implement `shape` for `identity`

amacati:
Implements the shape kwarg-only argument for `RigidTransform.identity`. This PR mirrors #23869, but does not include a `random` method since there is no `random` method for `RigidTransform`.

#### Reference issue
Contributes towards #23391

Adds the shape argument to `RigidTransform.identity`.

#### Additional information
Same as in #23869, we should talk about deprecating `num` at some point. In its current form, `shape` could replace `num` easily since it can be both a `tuple[int, ...]` as well as an `int`.


scottshambaugh (approved):
Looks good to me!

lucascolley (approved):
thanks both!

PR #23876, opened 2025-10-27
DOC: Add DOI to reference paper in `fractional_matrix_power` of `scipy.linalg`

star1327p:

Add DOI to reference paper in `fractional_matrix_power` of `scipy.linalg`


j-bowhay (approved):
Thanks. In future please try and group these improvements into one pr rather than lots of very small ones

PR #23883, opened 2025-10-28
DOC: forward port 1.16.3 relnotes

tylerjereddy:
* Forward port the SciPy `1.16.3` release notes following the release this morning.

* Note that this patch does not include a version switcher update because we're trying to avoid some infrastructure taxing issues with patch release doc uploads that have effectively no changes as noted at:
https://github.com/scipy/docs.scipy.org/issues/102#issuecomment-3418039816

[docs only]


ev-br:
Thanks Tyler

PR #23884, opened 2025-10-28
REL, MAINT: prepare for SciPy 1.16.4

tylerjereddy:
* Prepare for SciPy `1.16.4`, just in case.

[ci skip] [skip ci]

I suspect another bug fix release in this series would be "emergency only."


rgommers (approved):
Thanks for yet another release Tyler!

PR #23888, opened 2025-10-29
ENH: stats.median_test: vectorize implementation

mdhaber:
#### Reference issue
Toward gh-14651
Toward gh-20544

This vectorizes the implementation of `scipy.stats.median_test` and prepares for array API translation.

#### Additional information 
This one was tricky. First, it relied on use of `stats.chi2_contingency` with 2D tables, but `chi2_contingency` cannot readily be vectorized to work on batches of 2D tables because it already accepts an ND array, interpreting it as a single multiway table. There is the possibility of adding an argument for the user to identify the core dimensionality of the input, but there's no precedent for that. (The closest thing I can think of was that we chose to make `wasserstein_distance_nd` a separate function so that `wassertein_distance` *could* be vectorized in the future. In this case, that ship has sailed.) So the solution here was to write a separate vectorized (private) `_chi2_contingency_2d` function.

Next, this has some very specific data-dependent error messages. Besides the eventual difficulties we'll run into with data-dependent branching on lazy backends, we can already have one slice that is problematic when all the rest are fine, or we could even want to produce different messages for different slices. There is precedent for dealing with this sort of thing - e.g. NaNs in the input produce NaNs in the output rather than raising - but I don't really want to slow down development and code execution for the sake of holding the user's hand when they provide degenerate input. My solution here was to continue raising for 1D input and to just return normally (with NaNs in the appropriate places) for multidimensional input.

The most significant issue is that `median_test` is not strictly a reducing function - the core dimensionality of the `table` output is 2.  The `_axis_nan_policy` decorator and tests were not written with this in mind, and it's not worth generalizing them just for this function, so we need to do some things manually. Besides the extra work, this leads to the question of whether it is worth the trouble of supporting `axis` and `keepdims` arguments. Users are very familiar with reducing functions (output core dimensionality `0`) for which the output shape is the same as the input shape but with element `axis` removed. Functions with core dimensionality `1` - or reducing functions with `keepdims=True` - are also pretty familiar - the output shape is the same as the input shape except possibly that element `axis` may be different - it might be the same (e.g. `rankdata`) or it might depend on the shape of one of the arguments (e.g. `quantile`). But I can't think of any vectorized function that accepts `axis` and has output core dimensionality `2`. If we accept an `axis` argument, then the shape of the `table` output would have additional dimensions inserted at position `axis`. `keepdims` could be applied to the other three outputs, but it wouldn't make sense for the `table` output because the `axis` dimension is not reduced away by default. This seems confusing. My solution here is to effectively support only `axis=-1`; inputs may be multidimensional, but the operation is performed independently on each slice along the last axis. This doesn't stop `axis` and `keepdims` arguments from being added in the future, but that would be a separate effort.

If that all sounds OK for this iteration, great. If not, I'll just table this.


scipy/stats/_morestats.py
```diff
@@ -35,10 +35,10 @@
 from ._fit import FitResult
 from ._stats_py import (_get_pvalue, SignificanceResult,  # noqa:F401
                         _SimpleNormal, _SimpleChi2, _SimpleF)
-from .contingency import chi2_contingency
 from . import distributions
```
mdhaber, line 40: CI failure about this. Not going to fight it here.
```suggestion
from .contingency import chi2_contingency  # noqa:F401
from . import distributions
```

scipy/stats/_morestats.py
```diff
@@ -36,10 +36,10 @@
 from ._fit import FitResult
 from ._stats_py import (_get_pvalue, SignificanceResult,
                         _SimpleNormal, _SimpleChi2, _SimpleF, _demean)
-from .contingency import chi2_contingency
 from . import distributions
```
mdhaber, line 40:
```suggestion
from .contingency import chi2_contingency  # noqa:F401
from . import distributions
```

scipy/stats/contingency.py
```diff
@@ -24,11 +24,12 @@
...
 import math
 import numpy as np
-from ._stats_py import power_divergence, _untabulate
+from ._stats_py import power_divergence, _untabulate, Power_divergenceResult
 from ._relative_risk import relative_risk
 from ._crosstab import crosstab
 from ._odds_ratio import odds_ratio
-from scipy._lib._array_api import xp_capabilities
+from scipy._lib._array_api import (array_namespace, xp_capabilities, xp_result_type,
+                                   xp_size)
```
mdhaber:
```suggestion
                                   xp_size, xp_device)
```

scipy/stats/contingency.py
```diff
@@ -422,6 +423,43 @@ def statistic(table, axis):
                                   alternative='greater', **method)
 
 
+@xp_capabilities(jax_jit=False, allow_dask_compute=True)
+def _chi2_contingency_2d(observed, *, correction=True, lambda_=1, xp=None):
+    # vectorized implementation of chi2_contingency for batches of 2D tables
+    # (chi2_contingency works for ND tables, so it can't be vectorized)
+    xp = array_namespace(observed) if xp is None else xp
+    dtype = xp_result_type(observed.dtype, force_floating=True, xp=xp)
```
mdhaber, line 431:
```suggestion
    dtype = xp_result_type(observed.dtype, force_floating=True, xp=xp)
    device = xp_device(observed)
```

scipy/stats/contingency.py
```diff
@@ -422,6 +423,43 @@ def statistic(table, axis):
...
+    rowsum = xp.sum(observed, axis=-1, keepdims=True)
+    colsum = xp.sum(observed, axis=-2, keepdims=True)
+    tablesum = xp.sum(rowsum, axis=-2, keepdims=True)
+    with np.errstate(invalid='ignore', divide='ignore'):  # NaN is a valid result
+        expected = rowsum * colsum / tablesum
+    dof = math.prod(table_shape) - sum(table_shape) + 1
+    ddof = sum(table_shape) - 2
+    if dof == 0:
+        return (xp.zeros(batch_shape, dtype=dtype)[()],
+                xp.ones(batch_shape, dtype=dtype)[()])
```
mdhaber:
```suggestion
        return (xp.zeros(batch_shape, dtype=dtype, device=device)[()],
                xp.ones(batch_shape, dtype=dtype, device=device)[()])
```

scipy/stats/contingency.py
```diff
@@ -422,6 +423,43 @@ def statistic(table, axis):
...
+    elif correction and dof == 1:
+        diff = expected - observed
+        direction = xp.sign(diff)
+        magnitude = xp.minimum(xp.asarray(0.5, dtype=diff.dtype), xp.abs(diff))
```
mdhaber:
```suggestion
        half = xp.asarray(0.5, dtype=diff.dtype, device=device)
        magnitude = xp.minimum(half, xp.abs(diff))
```

mdhaber (commented):
Good to see the lint check for device propagation.
What is the current practice for functions with multiple input arrays - do we assume they are on the same device, does each function have to invent a check for compatible devices, or is there a function that does it (e.g. have `xp_device` accept multiple arrays and resolve it)?

scipy/stats/contingency.py
```diff
@@ -422,6 +423,45 @@ def statistic(table, axis):
...
+    # (chi2_contingency works for ND tables, so it can't be vectorized)
+    xp = array_namespace(observed) if xp is None else xp
+    dtype = xp_result_type(observed.dtype, force_floating=True, xp=xp)
+    device = xp_device(observed)
+    batch_shape = observed.shape[:-2]
+    table_shape = observed.shape[-2:]
+
+    # Validate the sizes and shapes of the arguments.
+    if xp_size(observed) == 0:
+        nan = xp.full(batch_shape, xp.nan, dtype=dtype)
```
mdhaber:
```suggestion
        nan = xp.full(batch_shape, xp.nan, dtype=dtype, device=device)
```

lucascolley:
> What is the current practice for functions with multiple input arrays - do we assume they are on the same device, does each function have to invent a check for compatible devices, or is there a function that does it (e.g. have `xp_device` accept multiple arrays and resolve it)?

cc @rgommers @j-bowhay

rgommers:
> What is the current practice for functions with multiple input arrays - do we assume they are on the same device, 

Yes, that is the requirement.

> does each function have to invent a check for compatible devices, or is there a function that does it (e.g. have `xp_device` accept multiple arrays and resolve it)?

No need to invent anything. In many cases the current code will already refuse mismatches. And gh-25578 is the last PR in a series to fix device handling, it has better detection + the contributor docs that I hope will answer your questions in more detail.

mdhaber:
I think it's worth reviewing this one. The private function `_chi2_contingency_2d` is already written in terms of the array API, and I can translate `median_test` to the array API when this one is done.

scipy/stats/tests/test_contingency.py
```diff
@@ -269,6 +272,30 @@ def test_resampling_invalid_args(self):
...
+class TestChi2Contingency2d:
+    # test vectorized _chi2_contengency_2d
+    @pytest.mark.parametrize("correction", [False, True])
+    @pytest.mark.parametrize("lambda_", [None, 2])
+    @pytest.mark.parametrize("shape", [(1, 1), (1, 2), (2, 1), (2, 2),
+                                       (3, 2, 2), (3, 3, 4), (3, 4, 5, 6)])
+    def test_against_ref(self, correction, lambda_, shape, xp):
+        rng = np.random.default_rng(735847239109239455)
+        observed = rng.integers(10, size=shape)
+        dtype = xp.float64
```
lucascolley: deliberately using this instead of the default float dtype?

scipy/stats/_morestats.py
```diff
@@ -4305,60 +4307,67 @@ def median_test(*samples, ties='below', correction=True, lambda_=1,
     choice of `ties`.
 
     """
-    if len(samples) < 2:
+    np = array_namespace(*samples)
```
lucascolley: why `np`?

mdhaber: Don't remember. Maybe I thought I was going to to do the translatation at the same time, but decided to hold off? I'll just eliminate it for now.

```suggestion
```

scipy/stats/tests/test_contingency.py
mdhaber: Probably ran into a tolerance issue back then. I'll try parametrizing.

mdhaber: Still trying to compile to test locally. Keep running into problems with the `.ninja_deps` file "ending prematurely" and everything recompiling. Deleting the file fixes the problem (and makes compilation much faster?) so that it doesn't need to rebuild everything on subsequent builds... at least until I check out a new branch.

lucascolley: > Keep running into problems with the .ninja_deps file "ending prematurely" and everything recompiling. 

I was hitting this the other day — nuked my Pixi cache and hit a `git clean -xdf` and the problem went away.

mdhaber: I think it's fine with default dtype or parameterization. I had added the parameterization, then confused myself over what it was doing (just because I'm out of practice) and went back to default for the PR. In any case, it's default dtype now, and I think that's fine.

mdhaber:
Failures are unrelated (gh-25888 and something else).

j-bowhay (approved):
Agree with the approach of not adding `axis` etc

PR #23840, opened 2025-10-22
ENH: stats.levene: vectorize implementation

mdhaber:
#### Reference issue
Toward gh-14651
Toward gh-20544

#### What does this implement/fix?
This vectorizes the implementation of `scipy.stats.levene` and prepares for array API translation.

#### Additional information 
Also adds `_axis_nan_policy` decorator to `trim_mean` so `keepdims` can be used in `levene`.

PR #23889, opened 2025-10-30
ENH: stats.levene: add array API support

mdhaber:
#### Reference issue
Toward gh-20544

Adds array API support to `scipy.stats.levene`.

#### Additional information
This worked out really nicely after gh-23840. The only complication is that I also had to add array API support to `trim_mean`, which I didn't plan on doing at all. But that went very smoothly, too, so might as well.

### DIFF
--- a/scipy/stats/_morestats.py
+++ b/scipy/stats/_morestats.py
@@ -19,6 +19,7 @@ from scipy._lib._array_api import (
     xp_capabilities,
     is_numpy,
     is_jax,
+    is_dask,
     xp_size,
     xp_vector_norm,
     xp_promote,
@@ -3030,7 +3031,7 @@ def bartlett(*samples, axis=0):
 LeveneResult = namedtuple('LeveneResult', ('statistic', 'pvalue'))
 
 
-@xp_capabilities(np_only=True)
+@xp_capabilities()
 @_axis_nan_policy_factory(LeveneResult, n_samples=None)
 def levene(*samples, center='median', proportiontocut=0.05, axis=0):
     r"""Perform Levene test for equal variances.
@@ -3051,7 +3052,7 @@ def levene(*samples, center='median', proportiontocut=0.05, axis=0):
         When `center` is 'trimmed', this gives the proportion of data points
         to cut from each end. (See `scipy.stats.trim_mean`.)
         Default is 0.05.
-    axis : int or tuple of ints, default: None
+    axis : int or tuple of ints, default: 0
         If an int or tuple of ints, the axis or axes of the input along which
         to compute the statistic. The statistic of each axis-slice (e.g. row)
         of the input will appear in a corresponding element of the output.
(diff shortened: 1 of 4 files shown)


scipy/stats/_morestats.py
```diff
@@ -3051,7 +3052,7 @@ def levene(*samples, center='median', proportiontocut=0.05, axis=0):
         When `center` is 'trimmed', this gives the proportion of data points
         to cut from each end. (See `scipy.stats.trim_mean`.)
         Default is 0.05.
-    axis : int or tuple of ints, default: None
+    axis : int or tuple of ints, default: 0
```
mdhaber, line 3055: @j-bowhay fixed this.

scipy/stats/_morestats.py
```diff
@@ -3119,50 +3120,55 @@ def levene(*samples, center='median', proportiontocut=0.05, axis=0):
...
     if k < 2:
-        raise ValueError("Must enter at least two input sample vectors.")
+        raise ValueError("Must provide at least two samples.")
 
     if center == 'median':
 
         def func(x):
-            return np.median(x, axis=-1, keepdims=True)
+            return (xp.median(x, axis=-1, keepdims=True)
+                    if (is_numpy(xp) or is_dask(xp))
```
mdhaber, line 3136: Someone else is welcome to write an `xp_median` and massage this. I use NumPy `median` because it avoids speed regression, and `dask.median` because `quantile` doesn't work with Dask (no `take_along_axis`).

scipy/stats/_morestats.py
```diff
@@ -3119,50 +3120,55 @@ def levene(*samples, center='median', proportiontocut=0.05, axis=0):
...
     dfn = (k - 1.0)
-    denom = dfn * sum(np.sum((Zij - Zbari)**2, axis=-1, keepdims=True)
+    denom = dfn * sum(xp.sum((Zij - Zbari)**2, axis=-1, keepdims=True)
                       for Zij, Zbari in zip(Zijs, Zbaris))
 
     W = numer / denom
-    W = np.squeeze(W, axis=-1)
+    W = xp.squeeze(W, axis=-1)
+    dfn, dfd = xp.asarray(dfn, dtype=W.dtype), xp.asarray(dfd, dtype=W.dtype)
     pval = _get_pvalue(W, _SimpleF(dfn, dfd), 'greater', xp=np)
```
mdhaber:
```python3
import torch as xp
from scipy import special
x = xp.asarray(7., dtype=xp.float32)
special.fdtrc(5., 6., x).dtype  # torch.float64
```
I'd think this should either fail (e.g. `special_gammainc(): argument 'input' (position 1) must be Tensor, not float`) or maintain the dtype of the array argument. Is this intended @steppi?

Anyway, that's why I explicitly convert `dfn` and `dfd` to arrays of the same dtype.

scipy/stats/_stats_py.py
```diff
@@ -3707,11 +3709,13 @@ def trim_mean(a, proportiontocut, axis=0):
...
         raise ValueError("Proportion too big.")
 
-    atmp = np.partition(a, (lowercut, uppercut - 1), axis)
+    atmp = (np.partition(a, (lowercut, uppercut - 1), axis) if is_numpy(xp)
+            else xp.sort(a, axis=axis))
 
     sl = [slice(None)] * atmp.ndim
     sl[axis] = slice(lowercut, uppercut)
-    return np.mean(atmp[tuple(sl)], axis=axis)
+    trimmed = xp_promote(atmp[tuple(sl)], force_floating=True, xp=xp)
```
mdhaber, line 3717: We could promote at the beginning, but I thought some of the above might be a bit faster if the input is integral. Could change it, though.

scipy/stats/tests/test_morestats.py
```diff
@@ -782,69 +782,70 @@ def test_negative_pvalue_gh21152(self, xp):
...
         y = x**3
+        x, x2, y = xp.asarray(x), xp.asarray(x2), xp.asarray(y)
         W1, pval1 = stats.levene(x, y, center='mean')
         W2, pval2 = stats.levene(x2, y, center='median')
-        assert_almost_equal(W1, W2)
-        assert_almost_equal(pval1, pval2)
-
-    def test_bad_keyword(self):
-        x = np.linspace(-1, 1, 21)
-        assert_raises(TypeError, stats.levene, x, x, portiontocut=0.1)
```
mdhaber, line 829: I don't think we need to test for errors emitted by Python.

scipy/stats/tests/test_morestats.py
```diff
@@ -782,69 +782,70 @@ def test_negative_pvalue_gh21152(self, xp):
...
         args = [g1, g2, g3, g4, g5, g6, g7, g8, g9, g10]
+        args = [xp.asarray(arg) for arg in args]
         res = stats.levene(*args)
         attributes = ('statistic', 'pvalue')
-        check_named_results(res, attributes)
-
-    # temporary fix for issue #9252: only accept 1d input
-    def test_1d_input(self):
-        x = np.array([[1, 2], [3, 4]])
-        assert_raises(ValueError, stats.levene, g1, x)
```
mdhaber, line 847: Another reminder that `match` is important: this test continued to pass after `levene` began accepting multidimensional input because `g1` and `x` are not broadcastable. Anyway, this is no longer relevant.

scipy/stats/_morestats.py
```diff
@@ -3030,7 +3031,7 @@ def bartlett(*samples, axis=0):
 LeveneResult = namedtuple('LeveneResult', ('statistic', 'pvalue'))
 
 
-@xp_capabilities(np_only=True)
+@xp_capabilities()
```
mdhaber: Torch GPU fails in CI because it doesn't have `fdtrc`. I imagine JAX will fail, too.
```suggestion
@xp_capabilities(cpu_only=True, exceptions=['cupy'])
```
I still find this a bit indirect - I would prefer to declare that this fails with Torch and JAX on GPU rather than make exceptions to `cpu_only=True`. But @lucascolley, you mentioned during our SDG call that you think that making exceptions is the preferred way?

lucascolley: Yes, I still think this is preferable. What I would really like is to be able to inherit capabilities (or rather limitations) from depended-on functions like `special.fdtrc`. Would be amazing if this could happen automatically, but I wonder whether something like `uses=special.fdtrc` as a parameter to `xp_capabilities` could be implemented.

scipy/stats/_morestats.py
j-bowhay: I guess it doesn't matter since we are using `alternative=greater` but to avoid confusion I think we should update `xp=np`
```suggestion
    pval = _get_pvalue(W, _SimpleF(dfn, dfd), 'greater', xp=xp)
```

j-bowhay (approved):

PR #23881, opened 2025-10-28
ENH: stats.epps_singleton_2samp: vectorize implementation

mdhaber:
#### Reference issue
Toward gh-14651
Toward gh-20544

#### What does this implement/fix?
This vectorizes the implementation of `scipy.stats.epps_singleton_2samp` and prepares for array API translation.

#### Additional information 
This relies on the `_apply_over_batch` decorator to vectorize `xpx.cov`, but I opened data-apis/array-api-extra#502 about vectorizing the implementation itself.

PR #23893, opened 2025-10-30
MAINT: stats.epps_singleton_2samp: NaNs/Infs in input -> NaNs in output

mdhaber:
#### Reference issue
Follow-up to gh-23881

Currently, NaNs in the input of `epps_singleton_2samp` produce NaNs in the output, whereas Infs result in an error. Either the way, the calculation can't produce a valid number (e.g. in the Inf case, we have sines and cosines of infinity), so let's keep the result consistent. Now that the function is vectorized, it is preferable to return normally with NaNs in the output that correspond with affected slices. This PR makes that change.

#### Additional information
@j-bowhay also replaced `axis : None` with `axis : 0` as appropriate in the documentation of `epps_singleton_2samp` and `kruskal`.

### DIFF
--- a/scipy/stats/_hypotests.py
+++ b/scipy/stats/_hypotests.py
@@ -9,7 +9,7 @@ from scipy.optimize import shgo
 from . import distributions
 from ._common import ConfidenceInterval
 from ._continuous_distns import norm
-from scipy._lib._array_api import xp_capabilities, array_namespace, xp_size
+from scipy._lib._array_api import xp_capabilities, array_namespace, xp_size, xp_promote
 from scipy._lib._util import _apply_over_batch
 import scipy._lib.array_api_extra as xpx
 from scipy.special import gamma, kv, gammaln
@@ -29,7 +29,7 @@ Epps_Singleton_2sampResult = namedtuple('Epps_Singleton_2sampResult',
                                         ('statistic', 'pvalue'))
 
 
-# remove when array-api-extra#502 is resolved 
+# remove when array-api-extra#502 is resolved
 @_apply_over_batch(('x', 2))
 def cov(x):
     return xpx.cov(x)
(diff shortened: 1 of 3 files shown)


scipy/stats/tests/test_hypotests.py
```diff
@@ -65,9 +65,22 @@ def test_epps_singleton_size(self):
...
+        w_ref[i] = np.nan
+        p_ref[i] = np.nan
+
+        x[i[0], 0] = np.nan
+        x[i[1], 1] = np.inf
+        y[i[2], 2] = -np.inf
+        w_res, p_res = epps_singleton_2samp(x, y, axis=-1)
+
+        assert_equal(w_ref, w_res)
+        assert_equal(p_ref, p_res)
```
mdhaber:
```suggestion
        assert_allclose(w_res, w_ref)
        assert_allclose(p_res, p_ref)
```

scipy/stats/_hypotests.py
```diff
@@ -112,17 +112,13 @@ def epps_singleton_2samp(x, y, t=(0.4, 0.8), *, axis=0):
     np = array_namespace(x, y)
     # x and y are converted to arrays by the decorator
     # and `axis` is guaranteed to be -1.
-    t = np.asarray(t)
+    x, y = xp_promote(x, y, force_floating=True, xp=np)
+    t = np.asarray(t, dtype=x.dtype)
```
j-bowhay, line 116: Not worth rerunning CI for but if we are coming back here in the near future then
```suggestion
    x, y, t = xp_promote(x, y, t, force_floating=True, xp=np)
```

j-bowhay (approved):

scipy/stats/_hypotests.py
mdhaber: I wrote it this way intentionally because even if `t` is treated as an array internally, I think the user may think of it like a sequence of separate floats. It is 1D and doesn't follow broadcasting rules with the samples. I have similar thoughts about `rng` in `iqr`.

I should double check, but I didn't think we wrote `xp_promote` to accept non-array array-likes, and we don't want to coerce it to an array before dtypes determination because that could upcast everything to the default float.

That said, I am a bit conflicted about it. I'll think some more.

PR #23895, opened 2025-10-30
BUG: linalg: fix MKL ILP64 symbol misnamings

ev-br:

In the ILP64 build,  in addition to 32-bit `_fblas.so` and `_flapack.so`, we build two more extensions, `_fblas_64.so` and `_flapack_64.so`. The latter fails to import due to missing symbols in the underlying MKL libraries: all ILP64 routines should have the `_64_`  suffix, but several names miss the trailing underscore. 

We already work around a missing symbol for the `_fblas_64` extension. Now that I tried to actually import `_flapack_64`, too [1], it's clear that we need to work around several others.

[1] Why importing them? While working on an SVD enhancement, I noticed that `scipy.linalg.lapack.HAS_ILP64` is False while `scipy.linalg.blas.HAS_ILP64` is True, while they should be either both False (in an LP64 build) or both True (for an ILP64 build). 

#### Additional information

The problem has been reported to Intel, so hopefully it'll get fixed in some future MKL version. Meanwhile, the workaround was needed with

```
$ mamba list |grep mkl
# packages in environment at /home/ev-br/.conda/envs/scipy-dev-mkl:
libblas                   3.9.0           31_hfdb39a5_mkl    conda-forge
libcblas                  3.9.0           31_h372d94f_mkl    conda-forge
liblapack                 3.9.0           31_hc41d3b0_mkl    conda-forge
mkl                       2024.2.2            ha957f24_16    conda-forge
mkl-devel                 2024.2.2            ha770c72_16    conda-forge
mkl-include               2024.2.2            ha957f24_16    conda-forge
```

### DIFF
--- a/scipy/__config__.py.in
+++ b/scipy/__config__.py.in
@@ -87,6 +87,7 @@ CONFIG = _cleanup(
                 "lib directory": r"@BLAS_LIBDIR@",
                 "openblas configuration": r"@BLAS_OPENBLAS_CONFIG@",
                 "pc file directory": r"@BLAS_PCFILEDIR@",
+                "has ilp64": bool(r"@BLAS_HAS_ILP64@".lower().replace('false', '')),
             },
             "lapack": {
                 "name": "@LAPACK_NAME@",
@@ -97,6 +98,7 @@ CONFIG = _cleanup(
                 "lib directory": r"@LAPACK_LIBDIR@",
                 "openblas configuration": r"@LAPACK_OPENBLAS_CONFIG@",
                 "pc file directory": r"@LAPACK_PCFILEDIR@",
+                "has ilp64": bool(r"@LAPACK_HAS_ILP64@".lower().replace('false', '')),
             },
             "pybind11": {
                 "name": "@PYBIND11_NAME@",
--- a/scipy/_build_utils/src/_blas64_defines.h
+++ b/scipy/_build_utils/src/_blas64_defines.h
@@ -50,6 +50,12 @@
 // https://community.intel.com/t5/Intel-oneAPI-Math-Kernel-Library/Missing-cspr-64-symbol-in-ILP64-MKL/td-p/1703471
 #ifdef FIX_MKL_2025_ILP64_MISSING_SYMBOL
 #define cspr_64_ cspr_64
+#define sgetc2_64_ sgetc2_64
+#define dgetc2_64_ dgetc2_64
+#define cgetc2_64_ cgetc2_64
+#define zgetc2_64_ zgetc2_64
+#define slasd4_64_ slasd4_64
+#define dlasd4_64_ dlasd4_64
 #endif
 
 #define F_INT npy_int64
--- a/scipy/linalg/blas.py
+++ b/scipy/linalg/blas.py
@@ -227,12 +227,12 @@ try:
 except ImportError:
     _cblas = None
 
-try:
+from scipy.__config__ import CONFIG
+HAS_ILP64 = CONFIG['Build Dependencies']['blas']['has ilp64']
+del CONFIG
+_fblas_64 = None
+if HAS_ILP64:
     from scipy.linalg import _fblas_64
-    HAS_ILP64 = True
-except ImportError:
-    HAS_ILP64 = False
-    _fblas_64 = None
 
 # Expose all functions (only fblas --- cblas is an implementation detail)
 empty_module = None
--- a/scipy/linalg/lapack.py
+++ b/scipy/linalg/lapack.py
@@ -875,12 +875,12 @@ try:
 except ImportError:
     _clapack = None
 
-try:
+from scipy.__config__ import CONFIG
+HAS_ILP64 = CONFIG['Build Dependencies']['lapack']['has ilp64']
+del CONFIG
+_flapack_64 = None
+if HAS_ILP64:
     from scipy.linalg import _flapack_64
-    HAS_ILP64 = True
-except ImportError:
-    HAS_ILP64 = False
-    _flapack_64 = None
 
 
 # Expose all functions (only flapack --- clapack is an implementation detail)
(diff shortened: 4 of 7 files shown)


.github/workflows/linux_blas.yml
```diff
@@ -100,6 +100,9 @@ jobs:
         pkg-config --libs mkl-dynamic-ilp64-seq  # check link flags
         spin build -S-Dblas=mkl -S-Duse-ilp64=true
 
+    - name: Check that the lib imports (fails is some symbols are missing)
+      run: spin run python -P -c"from scipy.linalg import _flapack_64"
```
lucascolley: gh-23700 needs to be updated if this is merged first.

ev-br:
Test failures in `scipy.signal` look unrelated.

.github/workflows/linux_blas.yml
```diff
@@ -100,6 +100,9 @@ jobs:
         pkg-config --libs mkl-dynamic-ilp64-seq  # check link flags
         spin build -S-Dblas=mkl -S-Duse-ilp64=true
 
+    - name: Check that the lib imports (fails is some symbols are missing)
```
tylerjereddy:
```suggestion
    - name: Check that the lib imports (fails if some symbols are missing)
```

.github/workflows/linux_blas.yml
tylerjereddy:
```suggestion
      run: spin run python -P -c "from scipy.linalg import _flapack_64"
```

.github/workflows/linux_blas.yml
```diff
@@ -138,5 +141,8 @@ jobs:
       run: |
         spin build --with-scipy-openblas -S-Duse-ilp64=true
 
+    - name: Check that the lib imports (fails is some symbols are missing)
+      run: spin run python -P -c"from scipy.linalg import _flapack_64"
```
tylerjereddy: similar minor corrections suggested here

tylerjereddy (commented):
Some questions here:

1) First, a bit more superficial--any reason to add the checks in CI like this vs. as unit tests proper under `HAS_ILP64` guards or something like that? If one wanted to verify locally or something, an encoded test might have use if it can be done? I do see a small number of tests with `@pytest.mark.skipif(not HAS_ILP64, reason="64-bit BLAS required")`.
2) Does this also mean that there are no substantive tests that reach into the core functionality that would use those symbols? If a simple import causes the failure, one might guess that basically none of the associated functionality is getting "flushed" by tests. Hard to do linalg checks that leverage those symbols even with `HAS_ILP64` guards?

rgommers:
> First, a bit more superficial--any reason to add the checks in CI like this vs. as unit tests proper under `HAS_ILP64` guards or something like that?

I think this CI test isn't ideal indeed. It could be that Accelerate ILP64 is silently not working as well, for the same reason right now. We have a gap in test coverage that can be fixed properly I think. WDYT about storing the value of `use-ilp64` at build time in the generated `__config__.py`, then at runtime use that to set `HAS_ILP64` rather than this dance:
```python
try:
    from scipy.linalg import _flapack_64
    HAS_ILP64 = True
except ImportError:
```
That import should always succeed unconditionally if `use-ilp64` was true at build time.

ev-br:
`__config__.py` sounds nice, and provides a missing link between build time and run time. grepping the codebase however shows its imports only wrapped in a try-except (https://github.com/scipy/scipy/blob/main/scipy/linalg/tests/test_decomp.py#L38-L41 and similar):

```
try:
    from scipy.__config__ import CONFIG
except ImportError:
    CONFIG = None
```

Is it meant be always available?

rgommers:
It is. That's a left-over from the bad old `distutils` days, see https://github.com/scipy/scipy/blob/8481c64522584eed4cf6fa3fb3aa495000961320/scipy/__init__.py#L47-L53. I think the try-except's can be removed, we shouldn't be importing the unbuilt source tree anymore.

EDIT: although the error message may have to be kept for usability reasons.

ev-br:
Updated to
- remove the CI import check
- add "has ilp64" entries to `__config__.py`, 
- use them instead of wrapping imports in try-except in linalg.{blas,lapack}; 
- while at it, do straightforward cleanups of importing `scipy.__config__` unconditionally in scipy/linalg. The main guard is in `scipy/__init__.py` is in place, and should remain in place.

> Does this also mean that there are no substantive tests that reach into the core functionality that would use those symbols? If a simple import causes the failure, one might guess that basically none of the associated functionality is getting "flushed" by tests. Hard to do linalg checks that leverage those symbols even with HAS_ILP64 guards?

Yes, precisely. ILP64 functionality is undertested and in fact very sparingly used in scipy itself. Several reasons:
- on CI, it's close to impossible with array sizes `N > 2**31` and especially for routines with `O(N**3)` complexity 
- we have only a few tests for large arrays AFAICS ([test_large_vector](https://github.com/scipy/scipy/blob/v1.16.3/scipy/linalg/tests/test_basic.py#L1792) for `norm`, one [svd test](https://github.com/scipy/scipy/blob/5a0029d433a9750f66748cad64d33d2a46cff12b/scipy/linalg/tests/test_decomp.py#L1144) and [one more svd test](https://github.com/scipy/scipy/blob/v1.16.3/scipy/linalg/tests/test_decomp.py#L1220))
- we don't use it much internally, yet. Only a subset of C extensions has been converted, and python-level f2py wrappers need an explicit opt-in. https://github.com/scipy/scipy/issues/23351 tracks both.

.github/workflows/linux_blas.yml
ev-br: The CI import check is gone, so there's no interference with gh-23700 anymore.

.github/workflows/linux_blas.yml
ev-br: Thanks for this and related suggestions. The CI check is gone though, based on popular demand :-).

rgommers:
Testing doesn't have to happen with very large arrays, it's much more important to test the `_64` routines at all, the same way as we test LP64.

rgommers (approved):
LGTM now, thanks @ev-br

ev-br:
Thanks for the reviews.

> test the _64 routines at all, the same way as we test LP64.

There's a concrete suggestion to this effect at
https://github.com/scipy/scipy/issues/23351#issuecomment-3476364853

PR #23897, opened 2025-10-30
DOC: sparse: Add examples sections to three new function doc_strings

dschult:
As discussed in #23534 the doc_strings do not have examples section for 3 new functions: `expand_dims`, `swap_axes`, `permute_dims`. 

    We want the total in #7168 to be monotonically decreasing over time.

This PR adds the examples sections.


WarrenWeckesser:
Thanks Dan!

The names and/or imports need to be fixed. And it might be nice to show `A.shape` before the function is applied to the array, so the reader can clearly see the before and after shapes (and doesn't have to count braces to figure out the input shape).

WarrenWeckesser:
Thanks Dan, merged.

PR #23898, opened 2025-10-31
ENH: stats.wilcoxon: add array API support

mdhaber:
#### Reference issue
Toward gh-20544

Adds array API support to `scipy.stats.wilcoxon`.

### DIFF
--- a/scipy/stats/_wilcoxon.py
+++ b/scipy/stats/_wilcoxon.py
@@ -6,7 +6,7 @@ from . import _morestats
 from ._axis_nan_policy import _broadcast_arrays
 from ._hypotests import _get_wilcoxon_distr
 from scipy._lib._util import _get_nan
-from scipy._lib._array_api import array_namespace, xp_promote, xp_size, xp_copy
+from scipy._lib._array_api import array_namespace, xp_promote, xp_size
 import scipy._lib.array_api_extra as xpx
 
 
--- a/scipy/stats/_wilcoxon.py
+++ b/scipy/stats/_wilcoxon.py
@@ -6,7 +6,7 @@ from . import _morestats
 from ._axis_nan_policy import _broadcast_arrays
 from ._hypotests import _get_wilcoxon_distr
 from scipy._lib._util import _get_nan
-from scipy._lib._array_api import array_namespace, xp_promote, xp_size
+from scipy._lib._array_api import array_namespace, xp_promote, xp_size, xp_copy
 import scipy._lib.array_api_extra as xpx
 
 
@@ -128,6 +128,7 @@ def _wilcoxon_statistic(d, method, zero_method='wilcox', *, xp):
         # Wilcoxon's method for treating zeros was to remove them from
         # the calculation. We do this by replacing 0s with NaNs, which
         # are ignored anyway.
+        d = xp_copy(d)  # this was required for array-api-strict, but it shouldn't be
         d = xpx.at(d)[i_zeros].set(xp.nan)
 
     i_nan = xp.isnan(d)
@@ -252,7 +253,7 @@ def _wilcoxon_nd(x, y=None, zero_method='wilcox', correction=True,
             p = 2 * np.minimum(dist.sf(np.floor(r_plus_np)),
                                dist.cdf(np.ceil(r_plus_np)))
             p = np.clip(p, 0, 1)
-        p = xp.asarray(p)
+        p = xp.asarray(p, dtype=d.dtype)
     else:  # `PermutationMethod` instance (already validated)
         p = stats.permutation_test(
             (d,), lambda d: _wilcoxon_statistic(d, method, zero_method, xp=xp)[0],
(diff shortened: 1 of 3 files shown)


scipy/stats/_wilcoxon.py
```diff
@@ -112,43 +113,43 @@ def _wilcoxon_iv(x, y, zero_method, correction, alternative, method, axis):
...
 
     if zero_method == 'wilcox':
         # Wilcoxon's method for treating zeros was to remove them from
         # the calculation. We do this by replacing 0s with NaNs, which
         # are ignored anyway.
-        if not d.flags['WRITEABLE']:
-            d = d.copy()
-        d[i_zeros] = np.nan
+        d = xp_copy(d)  # this was required for array-api-strict, but it shouldn't be
+        d = xpx.at(d)[i_zeros].set(xp.nan)
```
lucascolley: what's the problem here?

scipy/stats/_wilcoxon.py
```diff
@@ -79,9 +82,7 @@ def _wilcoxon_iv(x, y, zero_method, correction, alternative, method, axis):
         raise ValueError(message)
 
     message = "`x` (and `y`, if provided) must be an array of real numbers."
-    if np.issubdtype(d.dtype, np.integer):
-        d = d.astype(np.float64)
```
lucascolley, line 83: is this handled elsewhere now, or is it a deliberate behaviour change?

scipy/stats/tests/test_morestats.py
```diff
@@ -1490,152 +1503,170 @@ def test_wilcoxon_arg_type(self):
...
-        assert_equal(stat, 0)
-        assert_allclose(p, expected_p, rtol=1e-6)
-
-    def test_onesided(self):
-        # tested against "R version 3.4.1 (2017-06-30)"
+        #   [1] 0.00190419504300439
+        d = xp.asarray([0.1] * 10)
+        expected_stat = xp.asarray(0.0)
+
+        stat, p = stats.wilcoxon(d, method="asymptotic", correction=False, )
```
lucascolley:
```suggestion
        stat, p = stats.wilcoxon(d, method="asymptotic", correction=False)
```

scipy/stats/_wilcoxon.py
mdhaber: My understanding is that `xpx.at` methods should return a modified copy if the array is read only, and I think that's true for backends other than JAX:

> If x is a read-only NumPy array, they are the same as...

It does make a modified copy for a read-only NumPy array here, but it errors with a array-api-strict. Manually copying first solved the problem.

I'll try to create a MRE and file an issue if it helps.

scipy/stats/_wilcoxon.py
mdhaber: We're using `xp_promote` with `force_floating=True` at the top.

scipy/stats/_wilcoxon.py
lucascolley: would indeed be good!

mdhaber: data-apis/array-api-extra#506

scipy/stats/tests/test_morestats.py
mdhaber: 👍 We can commit this before merge, but I'll give CI a break for now.

scipy/stats/_wilcoxon.py
mdhaber:
```suggestion
        # Copy required for array-api-strict. See data-apis/array-api-extra#506.
        d = xpx.at(d)[i_zeros].set(xp.nan, copy=True)
```

scipy/stats/_wilcoxon.py
```diff
@@ -157,17 +158,19 @@ def _wilcoxon_statistic(d, method, zero_method='wilcox'):
...
-        mn -= n_zero * (n_zero + 1.) * 0.25
-        se -= n_zero * (n_zero + 1.) * (2. * n_zero + 1.)
+        n_zero = xp.astype(xp.count_nonzero(i_zeros, axis=-1), dtype)
+        mn = xpx.at(mn)[...].subtract(n_zero * (n_zero + 1.) * 0.25)
+        se = xpx.at(se)[...].subtract(n_zero * (n_zero + 1.) * (2. * n_zero + 1.))
 
         # zeros are not to be included in tie-correction.
         # any tie counts corresponding with zeros are in the 0th column
-        t[i_zeros.any(axis=-1), 0] = 0
+        # t[xp.any(i_zeros, axis=-1), 0] = 0
```
mdhaber, line 167: Can't mix boolean and scalar int indices with array API, so we have to work around it...

scipy/stats/_wilcoxon.py
```diff
@@ -6,6 +6,7 @@
 from ._axis_nan_policy import _broadcast_arrays
 from ._hypotests import _get_wilcoxon_distr
 from scipy._lib._util import _get_nan
+from scipy._lib._array_api import array_namespace, xp_promote, xp_size, xp_copy
```
mdhaber:
```suggestion
from scipy._lib._array_api import array_namespace, xp_promote, xp_size
```

j-bowhay (approved):

PR #23900, opened 2025-10-31
ENH: sparse.linalg.minres: support for complex hermitian matrices

dreivmeister:
#### Reference issue
Addresses #23814

Adds support for complex hermitian matrices A in minres by adjusting the used dot product.

#### Additional information
I am not super sure about taking the real part of e.g. beta1 before applying the sqrt but from what i have seen the imaginary parts a very small so in such cases that wont make a difference.


scipy/sparse/linalg/_isolve/tests/test_minres.py
```diff
@@ -6,83 +6,94 @@
 from pytest import raises as assert_raises
 
 
-def get_sample_problem():
+def get_sample_problem(make_complex=False):
```
tylerjereddy: minor: I suppose this is getting called 8 times now, to produce only two unique values, could decorate with a cache perhaps.

scipy/sparse/linalg/_isolve/tests/test_minres.py
```diff
@@ -6,83 +6,94 @@
...
 
 def test_shift():
-    A, b = get_sample_problem()
-    shift = 0.5
-    shifted_A = A - shift * np.eye(10)
-    x1, info1 = minres(A, b, shift=shift)
-    x2, info2 = minres(shifted_A, b)
-    assert_equal(info1, 0)
-    assert_allclose(x1, x2, rtol=1e-5)
+    for val in [False, True]:
```
tylerjereddy: minor: here and in a few other places in the PR, there is probably a slight preference to use `pytest` parametrization to separate the cases/test feedback

scipy/sparse/linalg/_isolve/minres.py
```diff
@@ -153,25 +154,25 @@ def minres(A, b, x0=None, *, rtol=1e-5, shift=0.0, maxiter=None,
...
         epsa = (s + eps) * eps**(1.0/3.0)
         if z > epsa:
             raise ValueError('non-symmetric matrix')
 
         # see if M is symmetric
         r2 = psolve(y)
-        s = inner(y,y)
-        t = inner(r1,r2)
+        s = dotprod(y,y)
+        t = dotprod(r1,r2)
```
tylerjereddy, line 181: note: these lines don't seem to be covered by tests at the moment when checking locally via `spin test -j 12 --coverage`

Image

lucascolley:
hey @dreivmeister , are you interested in responding to the review comments here?

dreivmeister:
Sorry for the delay!

scipy/sparse/linalg/_isolve/minres.py
lucascolley: now covered 
image

scipy/sparse/linalg/_isolve/tests/test_minres.py
```diff
@@ -2,31 +2,47 @@
...
-    matrix = rng.rand(10, 10)
-    matrix = matrix + matrix.T
+    if make_complex:
+        matrix = rng.rand(10, 10) + 1j * rng.rand(10, 10)
+    else:
+        matrix = rng.rand(10, 10)
+    if make_complex:
+        matrix = matrix + matrix.T.conj()
+    else:
+        matrix = matrix + matrix.T
```
tylerjereddy, line 19: I think we can avoid repeating the same conditional here, right? Simplifying as per the diff below the fold seems to work just fine for testing locally.

<details>

```diff
--- a/scipy/sparse/linalg/_isolve/tests/test_minres.py
+++ b/scipy/sparse/linalg/_isolve/tests/test_minres.py
@@ -13,11 +13,9 @@ def _cached_sample_problem(make_complex: bool):
     rng = np.random.RandomState(1234)
     if make_complex:
         matrix = rng.rand(10, 10) + 1j * rng.rand(10, 10)
-    else:
-        matrix = rng.rand(10, 10)
-    if make_complex:
         matrix = matrix + matrix.T.conj()
     else:
+        matrix = rng.rand(10, 10)
         matrix = matrix + matrix.T
```

</details>

tylerjereddy (commented):
The diff here seems small enough that we might be able to get it across the line on time for 1.17.0 given that it seems we could even use CuPy as a reference check for the same behavior and Lucas seems to have some knowledge of the array API support related to the change.

lucascolley (approved):
@tylerjereddy want another look? After working in this area recently I think I'm happy to put this in for 1.18.0.

tylerjereddy:
> want another look? After working in this area recently I think I'm happy to put this in for 1.18.0.

Ok, I took another look at the diff. I don't think we need type hinting in our test helpers, and the testing setup still looks a bit awkward to me, but apart from that the changes seem pretty clean/sensible, and I suppose there are no hard blockers here. The one CI failure is unrelated -- that's https://github.com/scipy/scipy/issues/25180.

The commit history is a bit messy, I'll probably squash merge then.

Issue #22765, opened 2025-03-31
BUG: scipy.integrate.solve_ivp: solution behavior diverges and converges suddenly


jy021125:
### Describe your issue.

I’m currently trying to solve a stiff ODE problem in Python using the LSODA method from solve_ivp.

The governing equation is u' = Lu. Here, u is defined as complex 2M-vector, and L is complex 2M x 2M matrix and L varies by time. The definition of matrices is complex, but it is well calculated externally, and I only solve the problem with given datas. The main code implementation is as follows. The details of the matrices is too long to show here, but I've checked this is not the problem of interpolation or inversion of matrices.

I’ve encountered various issues during implementation, and the main ones are as follows.

1. When I compute the Euclidean norm of the solution, at some point it suddenly grows and then abruptly drops to zero. I've found the complex part of the solution matrix becomes suddenly zero. And this phenomenon appears at different position by integration rangel. It’s difficult to pinpoint the cause of this phenomenon. Even if I switch the solver to RK45, Radau, or BDF, the same blow-up and drop-to-zero issue still occurs. This even makes the different solution of IVP problem. I think this is a numerical issue.

2. LSODA is supposed to advance quickly through non-stiff regions and reduce its step size in stiff regions, but compared to RK45, it runs nearly 100 times slower. Since RK45 seems to yield a proper solution, I’m wondering whether LSODA is really moving fast enough in the non-stiff portions.

Thank you for your help in advance!

![Image](https://github.com/user-attachments/assets/e12615c7-c6fa-44b0-b9b5-8a6f4484a76b)

![Image](https://github.com/user-attachments/assets/39fcb61f-3872-4d3e-8d58-ff2143e1df03)

### Reproducing Code Example

```python
def compute_matrices(psi):

    Q = M - N * q(psi)
    
    A = ((2 * np.pi) ** 2) * (
        N @ (G_22(psi) @ N + G_23(psi) @ M)
      + M @ (G_23(psi) @ N + G_33(psi) @ M)
    )
    
    B = -2 * np.pi * 1j * chi_r(psi) * (
        N @ (G_22(psi) + q(psi) * G_23(psi))
      + M @ (G_23(psi) + q(psi) * G_33(psi))
    )
    
    C = (
        (-2 * np.pi * 1j) * (
            chi_rr(psi) * (N @ G_22(psi) + M @ G_23(psi))
          + qchi_r_derivative(psi) * (N @ G_23(psi) + M @ G_33(psi))
        )
[...]

Issue #17912, opened 2023-02-02
BUG: Failing tests on multiple architectures when building scipy 1.10.0 for Fedora


psimovec:
### Describe your issue.

I am updating Fedora package for scipy 1.8.3 -> 1.10.0
(We skipped version 1.9, as there were test failures on i686 architecture, as reported in #17213 , and disabling the tests led to another tests failing)
I did 65+ builds, as disabling some tests led to other tests failing directly or being flaky. 

For first build, all of important failed tests and error messages are in [i686-build.log](https://github.com/scipy/scipy/files/10569544/i686-build.log) and [s390x-build.log](https://github.com/scipy/scipy/files/10569516/s390x-build.log).

For rest of the builds - rest of archived logs _( as logs in koji will expire in <14 days)_ : https://drive.google.com/file/d/1op-FgNp1bHHdJB8GAEWowiBrYo5GwDWG/view?usp=sharing _(5.3MB compressed, 943MB uncompressed)_
(check `kojilogs/noarch-<taskID>/<architecture>-*/build.log` in the archive)
You can also check [scipy-notes.txt](https://github.com/scipy/scipy/files/10569480/scipy-notes.txt) to see which tests were disabled at specific moment/in which tasks the tests failed.

### Test failures in first build:
https://koji.fedoraproject.org/koji/taskinfo?taskID=96702545
        =========================== short test summary info ============================
        **all architectures** [s390x-build.log](https://github.com/scipy/scipy/files/10569516/s390x-build.log) or  [i686-build.log](https://github.com/scipy/scipy/files/10569544/i686-build.log)
        FAILED scipy/integrate/tests/test_banded_ode_solvers.py::test_banded_ode_solvers
        FAILED scipy/integrate/tests/test_integrate.py::TestOde::test_vode - Assertio...
        FAILED scipy/integrate/tests/test_integrate.py::TestOde::test_zvode - Asserti...
        FAILED scipy/integrate/tests/test_integrate.py::TestOde::test_concurrent_ok
        FAILED scipy/integrate/tests/test_integrate.py::TestComplexOde::test_vode - A...
        FAILED scipy/integrate/tests/test_integrate.py::TestVODECheckParameterUse::test_no_params
        FAILED scipy/integrate/tests/test_integrate.py::TestVODECheckParameterUse::test_one_scalar_param
        FAILED scipy/integrate/tests/test_integrate.py::TestVODECheckParameterUse::test_two_scalar_params
[...]

Issue #13276, opened 2020-12-21
odeint with empty derivative vector causes segmentation fault


roryyorke:
The crash seems to happen in `xerrwv`, called from `lsoda` (see backtrace below).

`solve_ivp` handles the empty derivative vector case fine (see output below).

I found this while investigating the behaviour of `signal.lsim2` for feedthrough (empty ABC matrix) systems; calling `lsim2` with such a feedthrough system results in this crash.

#### Reproducing code example:

```python
import numpy as np
import scipy
from scipy import integrate

print(f'{np.__version__ = }')
print(f'{scipy.__version__ = }')

def fprime(x, t):
    return -x

def fprime2(t, x):
    return fprime(x, t)

t_eval = np.linspace(0,1)

for x0 in [[1,1], [1], []]:
    print(f'{x0 = }')
    sivp = integrate.solve_ivp(fprime, [t_eval[0],t_eval[-1]], x0, t_eval=t_eval)
    print(f'  {sivp.y.shape = }')
    print(f'  {sivp.t.shape = }')
    
    xout = integrate.odeint(fprime, x0, np.linspace(0, 1))
    print(f'  {xout.shape = }')
```

#### Error message:

Results of above, including a stack trace:
```
(scipydev) rory@rory-latitude:~/projects/scipy/odeint-crash$ python -X dev odeint_crash.py
np.__version__ = '1.19.2'
scipy.__version__ = '1.7.0.dev0+ec8912c'
x0 = [1, 1]
  sivp.y.shape = (2, 50)
  sivp.t.shape = (50,)
  xout.shape = (50, 2)
x0 = [1]
  sivp.y.shape = (1, 50)
  sivp.t.shape = (50,)
  xout.shape = (50, 1)
x0 = []
  sivp.y.shape = (0, 50)
  sivp.t.shape = (50,)
Fatal Python error: Segmentation fault

Current thread 0x00007fa1d76fe740 (most recent call first):
  File "/home/rory/src/scipy/scipy/integrate/odepack.py", line 241 in odeint
  File "odeint_crash.py", line 22 in <module>
Segmentation fault (core dumped)
(scipydev) rory@rory-latitude:~/projects/scipy/odeint-crash$ gdb $(which python) core 
GNU gdb (Ubuntu 9.2-0ubuntu1~20.04) 9.2
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
[...]

PR #23914, opened 2025-11-02
ENH:integrate: Rewrite LSODA in C

### LINKED ISSUE #15940: odeint (lsoda) emits too many warnings that are not filterable.
### Describe your issue.

I have a colleague who is attempting to use `odeint` to integrate a differential equation. This produces a large volume (>100k lines) of warning messages that, whilst real, they would like to filter as they pollute the Jupyter notebooks and are emitted in the Terminal window, etc.
The same noise is emitted in `solve_ivp` if `method='LSODA'` is used.

We have tried redirecting stdout, stderr, filtering Python warnings, etc, to no success. The message is emitted from [lsoda.f](https://github.com/scipy/scipy/blob/1ab82fffd30f3bf936c1118e55617a978f9b7e90/scipy/integrate/odepack/lsoda.f#L1339), which calls [xerrwv](https://github.com/scipy/scipy/blob/1ab82fffd30f3bf936c1118e55617a978f9b7e90/scipy/integrate/odepack/xerrwv.f#L1).

From reading lsoda.f it should be [possible set the logical unit that the messages are printed to, or to suppress warning messages](https://github.com/scipy/scipy/blob/main/scipy/integrate/odepack/lsoda.f#L798). What does logical unit 6 correspond to in Fortran terms?

My preference at the moment would be to sort out where the messages are directed to, followed by addition of a parameter to the module that would turn off the warning messages.

@llimeht

### Reproducing Code Example

```python
import contextlib
import io
import numpy
import numpy as np
from scipy.integrate import odeint


def dxdt(x, t):
    return [ x[0]**3 - 4. * x[1],
             x[0] - x[1] ]


f = io.StringIO()
with contextlib.redirect_stdout(f), contextlib.redirect_stderr(f):
    t = numpy.linspace(0, 8, 1001)
    x0 = np.array([-1.38918542, -7.87846202])
    x = odeint(dxdt, x0, t)
```

### Error message

```shell
lsoda--  warning..internal t (=r1) and h (=r2) are
       such that in the machine, t + h = t on the next step  
       (h = step size). solver will continue anyway
      in above,  r1 =  0.1750743642151D+00   r2 =  0.1257107638210D-16
<repeats ad nauseam>
```

1.8.0 1.21.4 sys.version_info(major=3, minor=9, micro=7, releaselevel='final', serial=0)

ilayn:
#### Reference issue
Towards #18566 

- It is an exceptionally bad code to read or reason about so I had to rewrite certain parts. 
- We are wrapping it twice, once in ODEPACK and once in lsoda.pyf for some unknown historical reason. That is fixed. Now it is the same code wrapped twice to satisfy all interfaces.
- The concurrency issues are now resolved and it is thread-safe, refactored (albeit a bit randomly for my own sanity).
- I nudged one test relative tolerance in test_banded_ode_solvers from 1e-7 to 2e-7 but the absolute difference is 1e-12 so quite small bump.
- I removed the fortran based banded5x5 test which was testing the wrapper code. Since we are going to add asan CI job soon I don't think we need it anymore.

#### Additional information

- We should switch to SUNDIALS to use the modernized versions of these code. 

- Some performance comparison results (ignore the last column, it's LLM masturbation). The important bit is that the Jac and func evals finally match which was not fun to match. 

```

(scipy-dev) ➜  /tmp /home/ilayn/miniforge3/envs/scipy-dev/bin/python3 final_performance_report.py 2>&1
==========================================================================================
                         C LSODA BENCHMARK REPORT
==========================================================================================
Test 1: Robertson Chemical Kinetics (Stiff System)
------------------------------------------------------------------------------------------
Metric                            C LSODA         Fortran        Ratio     Status
------------------------------------------------------------------------------------------
Function evaluations                 4510            4510        1.000x  ✓ PERFECT
Jacobian evaluations                  470             470        1.000x  ✓ PERFECT
Time (median, ms)                   1.018           1.020        0.998x     ✓ GOOD
Max solution error               0.00e+00

==========================================================================================
Test 2: Van der Pol Oscillator (μ=1000, Very Stiff)
------------------------------------------------------------------------------------------
/tmp/final_performance_report.py:147: ODEintWarning: Excess work done on this call (perhaps wrong D...
  sol_c = odeint_c(vanderpol_f_c, y0, t, Dfun=vanderpol_jac_c)
/tmp/final_performance_report.py:157: ODEintWarning: Excess work done on this call (perhaps wrong D...
  sol_f = odeint_fortran(vanderpol_f_f, y0, t, Dfun=vanderpol_jac_f)
Metric                            C LSODA         Fortran        Ratio     Status
------------------------------------------------------------------------------------------
Function evaluations                 6425            6425        1.000x  ✓ PERFECT
Jacobian evaluations                  390             390        1.000x  ✓ PERFECT
Time (median, ms)                   0.898           0.896        1.002x     ✓ GOOD

==========================================================================================
Test 3: Lorenz System (Chaotic, Non-Stiff)
------------------------------------------------------------------------------------------
Metric                            C LSODA         Fortran        Ratio     Status
------------------------------------------------------------------------------------------
Function evaluations                66555           66555        1.000x  ✓ PERFECT
Jacobian evaluations                    0               0        0.000x  ✓ PERFECT
Time (median, ms)                   9.723           9.753        0.997x     ✓ GOOD
Max solution error               0.00e+00

==========================================================================================
SUMMARY
==========================================================================================

✓ All Jacobian evaluations match Fortran exactly (ratio ~1.0)
✓ All function evaluations match Fortran exactly (ratio ~1.0)
✓ Performance overhead < 0.01% across all test cases
✓ Solutions match Fortran LSODA with machine precision

```

- Rewrote `stoda.f` completely. I don't know if I understood what it does but quacks like `stoda` now. 

- We need to do something about legacy interfaces. We can't keep delaying it to SciPy 2.0. This is very difficult to figure out which interface is calling which function. Some are in `_ivp` some are in `LSODA` class some in the `odeint` interface. 

- And way way too much OOP programming on our end. It's very difficult to reason about the code both in tests and in code. We are not touching these codes, almost ever. There is no need for dependency injections or indirections, they can be written purely by functions, even if there is code repetition, we won't be touching them. And it is a pain in the ass to modify a single test,  as all test suites demonstrate via lots of if else cases. 

-------------------

Below are seemingly relevant issues, I'll mark them as I start testing later.

#5335
#7379
#13276
#15860
Closes #15940 
#17912
#22765


ilayn:
The failures are from public private test. But the lsoda is not a module any more so I think this will be a backwards compatible change (not that it should matter). Because there is no way to restore that kind of functionality as a module unless we re-expose the extensions as a separate module. I don't see the point doing that.

steppi:
> The failures are from public private test. But the lsoda is not a module any more so I think this will be a backwards compatible change (not that it should matter). Because there is no way to restore that kind of functionality as a module unless we re-expose the extensions as a separate module. I don't see the point doing that.

Yeah, it looks safe to just take `scipy.integrate.lsoda` out of `test_public_api.py`. Nice work. I spent some time looking into the details of `LSODA` before and will try to find time to take a look at this. Really exciting to see you've managed to translate this one.

ilayn:
Seems like Intel job is having <strike>[an unrelated f2py issue](https://github.com/scipy/scipy/actions/runs/19102961478/job/54581580315?pr=23914#step:10:899).</strike> cython issue. 

Other than that this is good to go.

scipy/integrate/_odepackmodule.c
```diff
@@ -204,172 +193,268 @@ ode_function(F_INT *n, double *t, double *y, double *ydot)
...
+        *n = -1;
+        return;
     }
-    else {
-        nrows = *n;
+
+    // Calculate expected dimensions based on Jacobian type
+    npy_intp expected_nrows, expected_ncols;
+    expected_ncols = *n;
+    if (current_odepack_callback->jac_type == 3) {
```
MatthewFlamm: Should this be 4?  Other sections of the code have it defined that way.  For example current line 613 says

```C
    // jt values: 1=user full, 2=finite-diff full, 4=user banded, 5=finite-diff banded
```

```suggestion
    if (current_odepack_callback->jac_type == 4) {
```

There are a few other lines in this section of code that have 3, and those would also have to be changed.

ilayn: Yes seems so. I'll look into it a bit and push the changes.

ilayn: I pushed some corrections and moved the banded jacobian wrapper to the C extension. `odeint` and `solve_ivp` interface were doing different things so now they are unified in that sense.

ilayn:
If there are no comments, I'd like to merge this tonight since downstream nightlies, there will be some performance concerns or strange behavior.

steppi:
> If there are no comments, I'd like to merge this tonight since downstream nightlies, there will be some performance concerns or strange behavior.

I will take a look today and merge tonight.

scipy/integrate/meson.build
```diff
@@ -111,17 +66,17 @@ py3.extension_module('_test_multivariate',
...
+#   ['tests/banded5x5.f', f2py_gen.process('tests/test_odeint_banded.pyf')],
+#   link_with: [lsoda_lib, mach_lib],
+#   fortran_args: _fflag_Wno_unused_dummy_argument,
+#   link_args: version_link_args,
+#   dependencies: [lapack_lp64_dep, fortranobject_dep],
+#   install: true,
+#   link_language: 'fortran',
+#   subdir: 'scipy/integrate',
+#   install_tag: 'tests'
+# )
```
WarrenWeckesser: Delete this commented out section.

scipy/integrate/src/lsoda.h
```diff
@@ -0,0 +1,85 @@
+#ifndef __LSODA_H
+#define __LSODA_H
```
WarrenWeckesser: Names with two leading underscores (and names with a single underscore followed by a capital letter) are reserved for the compiler and should not be used in user code. Remove the underscores, and maybe follow a guideline that I've seen before (google's?) to use more of the path in the include guard, e.g `INTEGRATE_SRC_LSODA_H`.  Given the context, though, `LSODA_H` is probably fine, too.

WarrenWeckesser (commented):
Thanks for working on this, @ilayn.  I won't be able to do a thorough review, but after a quick skim,  I have just a couple small suggestions.

It might also be nice to reduce the use of the "magic numbers" 1, 2, 3, 4, and 5 that define the Jacobian type by giving them symbolic names, but I see that many of the uses of these magic numbers also have comments explaining what it is, which helps.

WarrenWeckesser:
I forgot to mention that I did run a bunch of examples of the use of `odeint` with this PR (including the example from the `integrate` module docstring that uses `odeint` to solve a banded system that comes from the method of lines for solving a PDE), and I didn't run into any problems. :rocket:

steppi (approved):
I skimmed through `lsoda.c` today and spot checked a selection of things more closely, including `stoda`. Great job from the looks of it. Thanks @WarrenWeckesser for reviewing this too. I'll be happy to merge this once the minor suggestions are addressed.

steppi:
Thanks @ilayn. The failures are unrelated. I did a regular merge because your individual commits and messages look disciplined and worth preserving.

ilayn:
Thanks all for the help.

WarrenWeckesser:
> ... including the example from the integrate module docstring ...

In case anyone is interested: https://github.com/WarrenWeckesser/experiments/tree/main/python/scipy/gray-scott

PR #23915, opened 2025-11-03
BUG: `special`: Fix convergence criterion in `eval_{legendre, gegenbauer}_l`

jorenham:
Calling it a bug might be a overstating this a bit, but these convergence criteria would (almost) never be reached. It shouldn't make (much) difference in the outcomes, but this could help a little bit with performance for large $n$.

An alternative solution would be to just remove these checks.


jorenham:
For those wondering why I was looking at these Cython implementations; it's because I ported them to Rust (yea kinda random): https://github.com/jorenham/xsf-rust/pull/145

lucascolley (approved):
Looks like these originate from https://github.com/scipy/scipy/pull/2475, no obvious indication that the exact check was deliberate.

Any reason for choosing `<` over `<=`?

jorenham:
> Any reason for choosing `<` over `<=`?

consistency with some other checks I've seen around there

lucascolley:
thanks Joren

PR #23917, opened 2025-11-03
DOC: special: Ellipsoidal harmonic functions aren't Lame

jorenham:
I'm sure that whoever wrote this meant no harm (pun intended), but they're the *Lamé* functions, not the *Lame* functions :hand_over_mouth:.

Source: https://dlmf.nist.gov/29

### DIFF
--- a/scipy/special/_ellip_harm.py
+++ b/scipy/special/_ellip_harm.py
@@ -8,8 +8,8 @@ def ellip_harm(h2, k2, n, p, s, signm=1, signn=1):
     r"""
     Ellipsoidal harmonic functions E^p_n(l)
 
-    These are also known as Lame functions of the first kind, and are
-    solutions to the Lame equation:
+    These are also known as Lamé functions of the first kind, and are
+    solutions to the Lamé equation:
 
     .. math:: (s^2 - h^2)(s^2 - k^2)E''(s)
               + s(2s^2 - h^2 - k^2)E'(s) + (a - q s^2)E(s) = 0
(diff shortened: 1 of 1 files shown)


jorenham:
But tbh; I'm not sure whether we should put this in the release highlights, or if a public press release would be more appropriate :thinking:.

lucascolley:
docs build is failing

jorenham:
> docs build is failing

```
py:func reference target not found: statsmodels.stats.multitest.multipletests [ref.func]
```

Does statsmodels requires the ellipsoidal harmonic functions to be lame or something, or is this failure unrelated?

lucascolley:
I'm hesitant to make main red, but I can't fathom how the changes here could be related. The page does indeed exist (https://www.statsmodels.org/stable/generated/statsmodels.stats.multitest.multipletests.html#statsmodels.stats.multitest.multipletests), so let's YOLO this in and follow up if it is somehow the cause of the failure.

Thanks Joren!

lucascolley (approved):

WarrenWeckesser:
Earlier in the full `build docs` log there  this:

```
WARNING: failed to reach any of the inventories with the following issues:
intersphinx inventory 'https://www.statsmodels.org/stable/objects.inv' not readable due to ProtocolError: ('Connection broken: IncompleteRead(147456 bytes read, 502116 more expected)', IncompleteRead(147456 bytes read, 502116 more expected))
```

So it was probably a transient network glitch (or something lame like that ;).

PR #23919, opened 2025-11-03
ENH: linalg/SVD: move the batching loop to C

ev-br:

For `linalg.svd` and `linalg.svdvals`, wrap LAPACK functions directly, instead of a python loop over the f2py wrapped routines. 
The C side uses the same machinery as `inv` and `solve` --- <s>in fact the PR is on top of gh-23362, and will need a rebase once that one lands</s>rebased. 

The end result is that the performance is basically on par with `np.linalg.svd`, _and_ we can choose the lapack driver (`?gesdd` or `?gesvd`). 

The added benchmark shows a reasonable performance gain for deep batches of small arrays. For larger arrays the linear algebra dominates, of course.

```
$ spin bench --compare main -t BatchedBench.time_svd

<snip>

[50.00%] · For scipy commit 89c1b34d <lowlevel_svd> (round 2/2):
[75.00%] ··· linalg.BatchedBench.time_svd                                                                                                                      ok
[75.00%] ··· ================= ============= =============
             --                           module          
             ----------------- ---------------------------
                   shape           scipy         numpy    
             ================= ============= =============
              (10, 10, 10, 2)     255±5μs       247±9μs   
               (100, 10, 10)    1.77±0.01ms   1.74±0.01ms 
               (100, 20, 20)     6.50±0.1ms   6.47±0.02ms 
              (100, 100, 100)     206±4ms      149±0.1ms  
             ================= ============= =============

[75.00%] · For scipy commit 9da62d5e <main> (round 2/2):
[100.00%] ··· linalg.BatchedBench.time_svd                                                                                                                      ok
[100.00%] ··· ================= ============= =============
              --                           module          
              ----------------- ---------------------------
                    shape           scipy         numpy    
              ================= ============= =============
               (10, 10, 10, 2)   2.60±0.02ms     255±10μs  
                (100, 10, 10)    4.27±0.07ms   1.76±0.02ms 
                (100, 20, 20)     9.04±0.1ms   6.68±0.08ms 
               (100, 100, 100)     216±10ms      154±3ms   
              ================= ============= =============

| Change   | Before [9da62d5e] <main>   | After [89c1b34d] <lowlevel_svd>   |   Ratio | Benchmark (Parameter)                                  |
|----------|----------------------------|-----------------------------------|---------|--------------------------------------------------------|
| -        | 9.04±0.1ms                 | 6.50±0.1ms                        |    0.72 | linalg.BatchedBench.time_svd((100, 20, 20), 'scipy')   |
| -        | 4.27±0.07ms                | 1.77±0.01ms                       |    0.41 | linalg.BatchedBench.time_svd((100, 10, 10), 'scipy')   |
| -        | 2.60±0.02ms                | 255±5μs                           |    0.1  | linalg.BatchedBench.time_svd((10, 10, 10, 2), 'scipy') |

PERFORMANCE INCREASED.
```

#### Additional information

A potential follow-up is to add `gejsv`: we have a low-level f2py wrapper, but it is not exposed as a driver for the high-level `linalg.svd`. Drivers for structured matrices can be added with an `assume_a` type keyword (given a structure, convert to a bidiagonal form, then use `?bdsqr` or `?bsdqc`, but all these are potential enhancements. The current PR as is should be backwards compatible.

### DIFF
--- a/scipy/linalg/src/_batched_linalg_module.cc
+++ b/scipy/linalg/src/_batched_linalg_module.cc
@@ -279,16 +279,16 @@ _linalg_svd(PyObject* Py_UNUSED(dummy), PyObject* args) {
 
     switch(typenum) {
         case(NPY_FLOAT32):
-            info = _svd<float>(ap_Am, ap_U, ap_S, ap_Vh, jobz, vec_status);
+            info = _svd<float>(ap_Am, ap_U, ap_S, ap_Vh, jobz, lapack_driver, vec_status);
             break;
         case(NPY_FLOAT64):
-            info = _svd<double>(ap_Am, ap_U, ap_S, ap_Vh, jobz, vec_status);
+            info = _svd<double>(ap_Am, ap_U, ap_S, ap_Vh, jobz, lapack_driver, vec_status);
             break;
         case(NPY_COMPLEX64):
-            info = _svd<npy_complex64>(ap_Am, ap_U, ap_S, ap_Vh, jobz, vec_status);
+            info = _svd<npy_complex64>(ap_Am, ap_U, ap_S, ap_Vh, jobz, lapack_driver, vec_status);
             break;
         case(NPY_COMPLEX128):
-            info = _svd<npy_complex128>(ap_Am, ap_U, ap_S, ap_Vh, jobz, vec_status);
+            info = _svd<npy_complex128>(ap_Am, ap_U, ap_S, ap_Vh, jobz, lapack_driver, vec_status);
             break;
         default:
             PyErr_SetString(PyExc_RuntimeError, "Unknown array type.");
--- a/scipy/linalg/src/_batched_linalg_module.cc
+++ b/scipy/linalg/src/_batched_linalg_module.cc
@@ -1,8 +1,11 @@
 #ifndef _LINALG_INV_H
 #define _LINALG_INV_H
 
+#include <cstring>
+
 #include "_linalg_inv.hh"
 #include "_linalg_solve.hh"
+#include "_linalg_svd.hh"
 #include "_common_array_utils.hh"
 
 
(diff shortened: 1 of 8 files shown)


ev-br:
The CI failure is curious. LAPACK 3.12 documentation for `?gesdd` mentions that a `nan` entry in the input matrix results in `info = -4` (https://www.netlib.org/lapack/explore-html/df/d22/group__gesdd_ga8941e5ff50de36580dae8940015e9cb0.html), and we test for it.

With Accelerate LP64 interface, it appears that
- the f2py interface to `dgesdd` does return `info = -4`,
- the manual wrapping in this PR does not.

ev-br:
https://github.com/scipy/scipy/pull/23919/commits/6d976932305c471d408e4f97793f3a566ef80946 fixes the Accelerate failure. The story is a bit convoluted: on main, LP64 Accelerate is not using correctly mangled symbols. Apparently, Accelerate exports `dgesdd_` along with `dgesdd$NEWLAPACK`. We should only be using the latter, but were using the former because the relevant include variable was not defined. Now, the smoking gun is that `dgesdd_` is broken in an edge case which we happened to have [a test for](https://github.com/scipy/scipy/blob/main/scipy/linalg/tests/test_decomp.py#L1228-L1232). 

The error fixed by https://github.com/scipy/scipy/pull/23919/commits/6d976932305c471d408e4f97793f3a566ef80946 is new in scipy 1.17.0, and potentially affects all users who link to Accelerate in C. As such, we might even consider backporting the commit to 1.17.0 or stash it to 1.17.1.

The remaining CI failure with scipy-openblas64 is also seen in other recent PRs: https://github.com/scipy/scipy/pull/24232, https://github.com/scipy/scipy/pull/24238, https://github.com/scipy/scipy/pull/24225; I'd suspect the [`scipy-openblas64`  release](https://pypi.org/project/scipy-openblas64/#history) of Dec 22nd.

tylerjereddy:
> we might even consider backporting the commit to 1.17.0

I've tentatively added a backport label for the 3-line `meson` patch/commit if you think it is "safe."

ev-br:
FWIW I wouldn't consider this meson patch a must for 1.17.0 specifically. The issue it solves is fairly niche, so it does not need to interfere with your getting the RC across the finish line. The patch can incubate a bit until 1.17.1 (it's not like we'll be able to avoid a point release at some point :-)).

rgommers:
The `meson.build` change looks correct, and should be safe to backport - but it's indeed not important, since there are (almost?) no external users of ILP64. 

The ILP64 job with `scipy-openblas64` is failing, so this isn't mergable yet.

ev-br:
Given that the scipy-openblas64 CI failure happens across all recent PRs, so is unrelated, I'd like to press the green button "soon", in about a week's time. Unless somebody wants to take a look of course.

scipy/linalg/src/_linalg_svd.hh
```diff
@@ -0,0 +1,319 @@
...
+    // --------------------------------------------------------------------
+    for (npy_intp idx = 0; idx < outer_size; idx++) {
+        init_status(slice_status, idx, St::GENERAL);
+
+        // copy the slice to `data` in F order
+        T *slice_ptr = compute_slice_ptr(idx, Am_data, ndim, shape, strides);
+        copy_slice_F(data, slice_ptr, m, n, strides[ndim-2], strides[ndim-1]);
+
+        // SVD the slice
+        gesvd(&jobz, &jobz, &intm, &intn, data, &intm, ptr_S, buf_U, &ldu, buf_Vh, &ldvh, work, &lwork, rwork, &info);
```
rgommers: The `rwork` allocation and this call look incorrect, since `rwork` is only present for the complex `gesvd` routines. See, e.g., [these dgesvd docs](https://netlib.org/lapack/explore-html/d1/d7f/group__gesvd_gac6bd5d4e645049e49bb70691180abf07.html)

Should probably guard with something like `if (type_traits<T>::is_complex)`?

scipy/linalg/src/_linalg_svd.hh
```diff
@@ -0,0 +1,319 @@
...
+            vec_status.push_back(slice_status);
+
+            // cut it short on error in any slice
+            goto done;
+        }
+
+        // copy-and-tranpose U and Vh slices from temp buffers to the output;
+        // S slice is filled in in-place already; 
+        copy_slice_F_to_C(ptr_U, buf_U, u_shape0, u_shape1);
+        copy_slice_F_to_C(ptr_Vh, buf_Vh, vh_shape0, vh_shape1);
```
rgommers: The first arguments in the two lines above can be null pointers. This look like it's missing guarding for `jobz == 'N'`.

rgommers: Same issue for gesvd further down.

scipy/linalg/src/_batched_linalg_module.cc
```diff
@@ -190,6 +193,129 @@ _linalg_solve(PyObject* Py_UNUSED(dummy), PyObject* args) {
...
+            break;
+        default:
+            PyErr_SetString(PyExc_RuntimeError, "Unknown array type.");
+    }
+
+    if (info < 0) {
+        // Either OOM or internal LAPACK error.
+        Py_DECREF(ap_S);
+        Py_DECREF(ap_U);
+        Py_DECREF(ap_Vh);
```
rgommers: These can be NULL for `jobz == 'N'` it looks like. Perhaps test coverage for `'N'` is missing?

scipy/linalg/src/_batched_linalg_module.cc
```diff
@@ -1,8 +1,11 @@
 #ifndef _LINALG_INV_H
 #define _LINALG_INV_H
```
rgommers: Not touched now but it stands out: why is this header guard in a `.cc` file? It shouldn't be here, in the `.hh` file instead.

scipy/linalg/src/_batched_linalg_module.cc
```diff
@@ -190,6 +193,129 @@ _linalg_solve(PyObject* Py_UNUSED(dummy), PyObject* args) {
...
+    if(!dtype_ok || !PyArray_ISALIGNED(ap_Am)) {
+        PyErr_SetString(PyExc_TypeError, "Expected a real or complex array.");
+        return NULL;
+    }
+
+    // Basic checks of array dimensions
+    int ndim = PyArray_NDIM(ap_Am);
+    npy_intp *shape = PyArray_SHAPE(ap_Am);
+    if (ndim < 2) {
+        PyErr_SetString(PyExc_ValueError, "Expected at least a 2D array.");
```
rgommers, line 225: Missing a `return NULL`

scipy/linalg/src/_linalg_svd.hh
```diff
@@ -0,0 +1,319 @@
...
+    T *work = &buf[m*n];
+
+    T *buf_U = NULL;
+    T *buf_Vh = NULL;
+    if (jobz != 'N') {
+        buf_U = &buf[m*n + lwork];
+        buf_Vh = &buf[m*n + lwork + u_shape0*u_shape1];
+    }
+
+    CBLAS_INT *iwork = NULL;
```
rgommers: Unused variable `iwork`

scipy/linalg/src/_common_array_utils.hh
```diff
@@ -494,6 +507,54 @@ GEN_GTCON_CZ(c, npy_complex64, float)
 GEN_GTCON_CZ(z, npy_complex128, double)
 
 
+/* ?GESDD*/
```
rgommers: nit: GESVD

rgommers (changes requested):
Thanks @ev-br. This looks fairly close, but still found a number of issues on an initial pass through this - please have a look.

scipy/linalg/src/_linalg_svd.hh
ev-br: This one is actually correct: https://github.com/scipy/scipy/pull/23919/files#diff-fb7fe08a416c842fbe6608abf2fc85613def3ff03a4b65de37b0bdbbcc9e41dfR511

`gesvd` has four overloads for s-, d-, c-, and z- variants; `s-` and `d-` variants swallow it and `c-` and `z-` variants forward it to the LAPACK function. 

This kind of pattern is used across several routines to paper over this sort of differences (also rwork vs iwork, alphar, alphai vs alpha in ggev etc), exactly to avoid cluttering the algorithmic part with explicit branching (or complex/real sentinels that numpy.linalg uses).

rgommers: Ooh that seems a little ugly? You have the exact same special-casing anyway, just in the header file - which only seems extra confusing and will be unhelpful if `gesvd` is used from other code. Unless I'm missing something, this will be easier to understand if the internal headers we use actually match the BLAS/LAPACK signatures, modulo symbol mangling.

ev-br: We had an offline chat with Ralf about this. To summarize, I'm going to
  - rename the overloaded function to `call_gesvd` to make it more obvious that there's a thin layer on top of LAPACK;
  - spell out the overloads explicitly without C macro trickery
  - add more comments to explain what's going on.

And then ping here.

ev-br: Commit https://github.com/scipy/scipy/pull/23919/commits/6be1521d5859eacf6df947eb903e6c1cf523908b  in the updated PR makes this change. Does it read better to your eyes @rgommers? 
FWIW, I'd think that `call_gesdd` is a readability win indeed; macros or no macros is similar (but I wrote this macro, so am biased).

ev-br:
Thanks for the review Ralf! Updated to address review comments.

scipy/linalg/src/_linalg_svd.hh
ev-br: These two are actually fine, too: `copy_slice_F_to_C` uses loops from zero to `u_shape{0, 1}`, and if `jobz == 'N'`  these are zeros, so loops do not execute. Added a guard anyway, since this was more confusing than strictly necessary.

scipy/linalg/src/_batched_linalg_module.cc
ev-br: Weirdly enough, `'N'` is covered (dropped a debug print, ran the test suite), so looks like the test suite does not catch it. Maybe it's because test functions are short and thus arrays in tests are short-lived? That's unpleasant. Not sure how to defensively catch it (better than, quoting an experienced C/Python dev, "stare at the code really hard")

scipy/linalg/src/_linalg_svd.hh
rgommers: Yes, that looks much clearer, thanks! Resolving.

scipy/linalg/src/_batched_linalg_module.cc
rgommers: Yeah it can be tricky, wish we had better tooling indeed. Fixed now, so resolving.

scipy/linalg/src/_linalg_svd.hh
```diff
@@ -0,0 +1,323 @@
...
+    T *work = &buf[m*n];
+
+    T *buf_U = NULL;
+    T *buf_Vh = NULL;
+    if (jobz != 'N') {
+        buf_U = &buf[m*n + lwork];
+        buf_Vh = &buf[m*n + lwork + u_shape0*u_shape1];
+    }
+
+    real_type *rwork = (real_type *)malloc(5*min_mn*sizeof(real_type));
```
rgommers: It looks like this could use the same `if (type_traits<T>::is_complex)` as higher up to only allocate `rwork` when it's needed.

rgommers (approved):
Thanks @ev-br. The code LGTM now (one comment about avoiding an `rwork` allocation - optional), and the benchmarks show a decent speed-up. This is a lot of code, however batched `svd` is performance-critical and since this closes the gap with `np.linalg.svd` I think it is worth the effort and added code/maintenance indeed.

Feel free to merge, possibly after addressing the one remaining comment.

scipy/linalg/src/_linalg_svd.hh
ev-br: Yeah. This one was deliberate since the allocation here for `?gesvd` is $O(N)$ while `?gesdd` it is $O(N^2)$. So this one can go either way (less branching vs tighter memory footprint), but if it sticks out to a careful reader, fine, let me just add an `if constexpr` and kick the CI before merging.

ev-br:
CI is green, let's land this one. Thanks Ralf for the review.

PR #23921, opened 2025-11-03
BUG:signal: Ensure consistent dtypes for filter functions across alternative backends

steppi:
#### Reference issue

Closes #23747 

This PR fixes some inconsistencies in output dtype for filter functions when using PyTorch with default dtype `float32`. Currently on `main`, `default_dtype=float32` tests are skipped for filter functions, but such tests are enabled in this PR. This PR enforces the current behavior when using the NumPy backend of having outputs always have 64 bit base precision regardless of input `dtype` across backends. The principled thing to do would be to have the output `dtype` respect the input `dtype`, but I have been convinced by @ev-br that it is reasonable to do this incrementally, and start with a pragmatic fix. I have left a note in each relevant docstring informing users that a change in behavior may come in the future:

```python
#    The current behavior is for ``ndarray`` outputs to have 64 bit precision
#    (``float64`` or ``complex128``) regardless of the dtype of `Wn` but
#    outputs may respect the dtype of `Wn` in a future version.
```

In order to more ergonomically control output dtype, I have added a `dtype` keyword to the filter prototype functions (`besselap`, `buttap`, `cheb1ap`, `cheb2ap`, `ellipap`) which controls the base `dtype` for outputs. I am aware the name `dtype` may not be the best name for this kwarg because it does not control the output `dtype`, but rather the base real `dtype` for outputs, which in some cases are real and in some cases are complex. I would be happy to change the name of this kwarg to something which makes this more clear, perhaps `base_dtype`. Technically, I guess the addition of a new kwarg to the filter prototype functions is an enhancement that may require a post on the discourse.

There is another backwards incompatible change. For the filter prototype functions `buttap` etc., I have made the realness or complexness of outputs consistent across filter order. Previously, one could get either real or complex outputs depending on the order `N`. Now, if it is possible to get complex values for any `N`, the output `dtype` will be complex. I think this change is unobtrusive and in line with the general preference to avoid value-dependent output dtype. For cases like `buttap` where the array of zeros `z` is always empty, I have left the output as a real array because an empty set surely has no complex values in it. I suppose this change will also need to go through the discourse. I would be willing to revert it if there are reasonable objections. For filter functions themselves, the `dtype` can still depend on the value of the `btype` `kwarg` since the `"bandstop"` can have complex zeros (one for each zero at infinity of the associated lowpass filter). Of course, we could just have the `dtype` for `z` always complex to avoid this value dependency, but it seems more obtrusive than the existing change, which only impacts the dtypes of empty arrays. Also, I saw a good argument here, https://github.com/numpy/numpy/issues/29000#issuecomment-3414315562, from Sebastian Berg, that return dtype depending on a kwarg with a finite number of string literal options is not so bad, which has me convinced for the moment.

Beyond specifying a `dtype` in the filter prototype functions, making dtypes behave properly also required 

* A change in `zpk2tf` to avoid unnecessary overflow when `k` is a python scalar and the default dtype is `float32`.
* Another change in `zpk2tf` changing `*` to `xp.multiply` to work around PyTorch's difference in promotion rules for `0d` arrays` vs higher dimensional arrays (0d arrays are basically treated like NEP50 Python scalars, and don't impact the result dtype.
* A change in `scipy.signal._polyutils.poly` to avoid returning just a plain Python scalar in some branches, consistently returning an array (or numpy scalar) of appropriate dtype in all branches. There were some places where `xp.asarray` was getting called on the output of `scipy.signal._polyutils.poly`, casting it to the default dtype. Rather than the caller having to worry about this possibility, it seems easier for the function to always consistently return as array.

As mentioned above, `SCIPY_DEFAULT_DTYPE=float32` tests have been enabled for all filter functions. Although the output of the filter functions has `float64` base precision, tolerances still had to be loosened for the `float32` case due to round-off in the inputs. I added explicit float64 base precision dtypes to the reference output arrays.

#### Additional information

If we want to eventually update these filter functions to have output dtype respect the input dtype, it should hopefully be fairly straightforward. Setting the `dtype` in the filter prototype functions to be based on the input dtype, and getting rid of some explicit casts should do it. I think the primary bottleneck is just testing things out to make sure that we actually get useful results when all internal calculations are in float32 precision, and possibly needing to adapt things to lower precision. Then again, I thought this PR would be straightforward, but I found it fairly challenging to get all of the tests passing, even with the simpler behavior of always having `float64` output.

### DIFF
--- a/scipy/signal/_filter_design.py
+++ b/scipy/signal/_filter_design.py
@@ -2833,24 +2833,24 @@ def iirfilter(N, Wn, rp=None, rs=None, btype='band', analog=False,
 
     # Get analog lowpass prototype
     if typefunc == buttap:
-        z, p, k = typefunc(N, xp=xp)
+        z, p, k = typefunc(N, xp=xp, dtype=xp.float64)
     elif typefunc == besselap:
-        z, p, k = typefunc(N, norm=bessel_norms[ftype], xp=xp)
+        z, p, k = typefunc(N, norm=bessel_norms[ftype], xp=xp, dtype=xp.float64)
     elif typefunc == cheb1ap:
         if rp is None:
             raise ValueError("passband ripple (rp) must be provided to "
                              "design a Chebyshev I filter.")
-        z, p, k = typefunc(N, rp, xp=xp)
+        z, p, k = typefunc(N, rp, xp=xp, dtype=xp.float64)
     elif typefunc == cheb2ap:
         if rs is None:
             raise ValueError("stopband attenuation (rs) must be provided to "
                              "design an Chebyshev II filter.")
-        z, p, k = typefunc(N, rs, xp=xp)
+        z, p, k = typefunc(N, rs, xp=xp, dtype=xp.float64)
     elif typefunc == ellipap:
         if rs is None or rp is None:
             raise ValueError("Both rp and rs must be provided to design an "
                              "elliptic filter.")
-        z, p, k = typefunc(N, rp, rs, xp=xp)
+        z, p, k = typefunc(N, rp, rs, xp=xp, dtype=xp.float64)
     else:
         raise NotImplementedError(f"'{ftype}' not implemented in iirfilter.")
 
@@ -4622,7 +4622,7 @@ def ellipord(wp, ws, gpass, gstop, analog=False, fs=None):
     return ord, wn
 
 
-def buttap(N, *, xp=None, device=None):
+def buttap(N, *, xp=None, device=None, dtype=None):
     """Return (z,p,k) for analog prototype of Nth-order Butterworth filter.
 
     The filter will have an angular (e.g., rad/s) cutoff frequency of 1.
@@ -4632,11 +4632,20 @@ def buttap(N, *, xp=None, device=None):
     N : int
         The order of the filter
     %(xp_device_snippet)s
+    dtype : Optional[DType]
+        Base real dtype for returned outputs (complex outputs will have the
+        corresponding complex dtype). ``dtype`` should be for the backend
+        ``xp``. ``None`` is equivalent to the default dtype for``xp``.
+        Default: None.
 
     Returns
     -------
-    z, p, k
-        Poles, zeros and gain for the filter prototype
+    z : ndarray[floating]
+        Zeros of the transfer function. Is always an empty array.
+    p : ndarray[complexfloating]
+        Poles of the transfer function.
+    k : float
+        Gain of the transfer function.
 
     See Also
     --------
@@ -4645,17 +4654,20 @@ def buttap(N, *, xp=None, device=None):
     """
     if xp is None:
         xp = np_compat
+    if dtype is None:
+        dtype = xp_default_dtype(xp)
+
     if abs(int(N)) != N:
         raise ValueError("Filter order must be a nonnegative integer")
-    z = xp.asarray([], device=device)
-    m = xp.arange(-N+1, N, 2, device=device, dtype=xp_default_dtype(xp))
+    z = xp.asarray([], device=device, dtype=dtype)
+    m = xp.arange(-N+1, N, 2, device=device, dtype=dtype)
     # Middle value is 0 to ensure an exactly real pole
     p = -xp.exp(1j * xp.pi * m / (2 * N))
     k = 1
     return z, p, k
 
 
-def cheb1ap(N, rp, *, xp=None, device=None):
+def cheb1ap(N, rp, *, xp=None, device=None, dtype=None):
     """
     Return (z,p,k) for Nth-order Chebyshev type I analog lowpass filter.
 
@@ -4671,11 +4683,20 @@ def cheb1ap(N, rp, *, xp=None, device=None):
     rp: float
         The ripple intensity
     %(xp_device_snippet)s
+    dtype : Optional[DType]
+        Base real dtype for returned outputs (complex outputs will have the
+        corresponding complex dtype). ``dtype`` should be for the backend
+        ``xp``. ``None`` is equivalent to the default dtype for``xp``.
+        Default: None.
 
     Returns
     -------
-    z, p, k
-        Poles, zeros and gain for the filter prototype
+    z : ndarray[floating]
+        Zeros of the transfer function. Is always an empty array.
+    p : ndarray[complexfloating]
+        Poles of the transfer function.
+    k : float
+        Gain of the transfer function.
 
     See Also
     --------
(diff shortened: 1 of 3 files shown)


scipy/signal/_filter_design.py
```diff
@@ -1280,12 +1280,21 @@ def zpk2tf(z, p, k):
...
+    # k is a Python scalar, z has 64-bit precision, and the default
+    # dtype for xp is float32.
+    if not hasattr(k, "dtype"):
+        if isinstance(k, complex):
+            k_dtype = xp.complex128 if xp.real(z).dtype == xp.float64 else xp.complex64
+        else:
+            k_dtype = xp.real(z).dtype
+        k = xp.asarray(k, dtype=k_dtype)
+    else:
+        k = xp.asarray(k)
```
ev-br: FWIW, the need for this dance is scary. `k` in the `zpk` triplet is documented as _float_, so I wonder if we can make it just `k = xp.asarray(k, dtype=xp.float64)` always? With a caveat similar to `Wn` if needs be.

scipy/signal/_filter_design.py
```diff
@@ -4632,11 +4672,20 @@ def buttap(N, *, xp=None, device=None):
     N : int
         The order of the filter
     %(xp_device_snippet)s
+    dtype : Optional[DType]
+        Base real dtype for returned outputs (complex outputs will have the
+        corresponding complex dtype). ``dtype`` should be for the backend
+        ``xp``. ``None`` is equivalent to the default dtype for``xp``.
+        Default: None.
```
ev-br: I've to admit this addition makes me uneasy. We don't have a precedent for this, and this feels like a start of  a rabbit hole.

If the idea is to produce double precision inputs, at least for the time being, would it work to mandate that we always return float64 for reals and complex128 for complex instead?
IOW, this would amount to rolling back the use of `xp_default_dtype(xp)`, cc @j-bowhay

ev-br (commented):
Adding explicit `dtype=xp.float64` in tests certainly is a good change IMO. The fix to `_polyutils`---which is vendored from numpy.polynomial`---is definitely a good change.

Adding `dtype=` keyword to the prototype functions is a bit more questionable, again IMO.
While I realize I'm at least partly responsible for `xp_default_dtype` attempts here, and that the current state in main is inconsistent, and that this PR makes it consistent (which is great), the end result feels a bit too much. Effectively, we will then have three keywords (`xp`, `device`, and `dtype`) for these filter prototype functions.
Sure, this is what it takes to have a function which only accepts scalars and produces arrays, consistently across all backends, devices and single/double precision.
But if it is what it takes to go all the way, do we really need to go all the way? 

One argument for float32 is performance on some devices. Do we expect functions like `buttap` matter for performance though. My gut feeling is that the answer is no, it's not very likely to have a real-world need for `buttap` in a tight loop. The answer is surely different for things which _apply_ filters to potentially large signals, but constructing a filter is something I'd expect to only happen before a tight loop.
Quite likely performance-wise we could have lived with all filter prototype functions returning lists not arrays.

So I believe that for filter prototypes, user convenience matters more than performance. Comparing between
- having to parse the docs for three arguments, and
- having to cast the output of `buttap` to float64 manually once,
the user in me would prefer the second variant.

All that said, my preference is rather weak, and this PR is a definite improvement over the status quo. Therefore, if there are strong opinions in favor of adding the `dtype=` arguments to `buttap` and its ilk, I won't argue further.

steppi:
Thanks @ev-br. I agree with your comments. The explicit `dtype` keyword for prototypes makes it easy to control `dtype` behavior for SciPy developers, and would make it straightforward to have output `dtype` respect input `dtype` for filter functions, (from here it just a matter of testing thoroughly that the precision is reasonable). But, I now realize that this does make the API of the prototype functions unnecessarily complicated for users. I'll get rid of the `dtype` parameter, and change this to have the filter prototype functions also always have `float64`/`complex128` outputs.

steppi:
@ev-br. I've removed the ill conceived `dtype` kwarg for filter prototypes. Now these functions always return `float64` base precision outputs. I managed to make `zpk2tf` respect input dtype without requiring an excessively complex dance.

scipy/signal/_filter_design.py
```diff
@@ -1294,9 +1293,12 @@ def zpk2tf(z, p, k):
...
         for i in range(z.shape[0]):
             k_i = xp.asarray(k[i], dtype=xp.int64)
-            b[i, ...] = k_i * _pu.poly(z[i, ...], xp=xp)
+            b[i, ...] = xp.multiply(k_i, _pu.poly(z[i, ...], xp=xp))
     else:
-        b = k * _pu.poly(z, xp=xp)
+        # Use xp.multiply to work around torch type promotion
+        # non-compliance for operations between 0d and higher
+        # dimensional arrays.
+        b = xp.multiply(k, _pu.poly(z, xp=xp))
```
ev-br, line 1303: Having to write `xp.multiply` instead of a normal binary multiply operator because torch devs have other priorities is infuriating. Nothing we can do about though.

ev-br (approved):
The patch LGTM!

The CI errors are unrelated, and are likely all fixed up in main already. To be on a safe side though, could you please rebase on main to rerun the CI?

My main worry is the `-Werror` failure from pythran. It's either fixed up in https://github.com/scipy/scipy/pull/23815#discussion_r2451603139, or needs a similar suppression. This is for until https://github.com/serge-sans-paille/pythran/pull/2365 is released.

steppi:
Thanks @ev-br. I've rebased on main. Hopefully that clears up CI.

ev-br:
Actually, `zpk2tf` doctest failure looks related. What is the expected dtype of the transfer function's numerator and denominator if zeros and poles arrays are integer? The doctest seems to imply they should be float64, but in this branch

```
In [9]: z = [2,   6]

In [10]: p = [1,   8]

In [11]: k = 5

In [12]: zpk2tf(z, p, k)[0].dtype
Out[12]: dtype('int64')
```

on main, the result is `dtype('float64')`. Sigh.

EDIT: the rest looks unrelated.

steppi:
> Actually, `zpk2tf` doctest failure looks related. What is the expected dtype of the transfer function's numerator and denominator if zeros and poles arrays are integer? The doctest seems to imply they should be float64, but in this branch
> 
> ```
> In [9]: z = [2,   6]
> 
> In [10]: p = [1,   8]
> 
> In [11]: k = 5
> 
> In [12]: zpk2tf(z, p, k)[0].dtype
> Out[12]: dtype('int64')
> ```
> 
> on main, the result is `dtype('float64')`. Sigh.
> 
> EDIT: the rest looks unrelated.

Oops, that looks like something I introduced by mistake in the last set of changes, I assume when I simplified the dance around dtypes.

steppi:
OK, it looks like this last commit should fix it. There was a check before that converted `k` to `float64` if it's an integer `dtype` after converting to array which I accidentally removed when I tried to simplify the `dtype` dance.

ev-br:
The PR LGTM, CI is green modulo unrelated failures, merged. Thank you @steppi!

PR #23667, opened 2025-09-23
ENH: integrate.qmc_quad: add array API support

mdhaber:
#### Reference issue
Toward gh-20930

#### What does this implement/fix?
Adds array API support to `integrate.qmc_quad`.

#### Additional information
Of course, the QMC points are generated as a NumPy array and need to be converted, but there still can be considerable advantage to performing the function evaluations using an alternative backend.

PR #23922, opened 2025-11-04
BUG: optimize: add py314 wrapper for `inspect.signature` of pub callables

### LINKED ISSUE #23913: BUG: optimize: _wrap_callback function can trigger a NameError on Python 3.14 in edge cases due to PEP 649/749
### Describe your issue.

In optimize, scipy/optimize/_optimize.py:_wrap_callback[line 93], `sig = inspect.signature(callback)` is called to check the names in the argument list to decide on how the callback function should be wrapped. 

Because annotations are lazily evaluated in Python 3.14 (PEP 649/749), there is now a scenario where the annotations are undefined or forward referenced at runtime (Example, if the typing is under a TYPE_CHECKING block). `inspect.signature` triggers a NameError, even if the code runs fine otherwise. `_wrap_callback` does not really need the annotation information, so it is a bit unfortunate. 

Before Python 3.14, annotations were either defined or were strings, so this scenario doesn't occur, I think. 

Notes:

Python 3.14 also added the keyword argument `annotation_format` to `inspect.signature`. Specifically, `Format.FORWARDREF` produces the same behavior as before when the annotations are defined at run time, but returns a special `ForwardRef` object in these cases where the annotations are undefined or forward referenced. 

Replacing line 93 with `sig = inspect.signature(callback, annotation_format=inspect.Format.FORWARDREF)` does indeed work for me here.

### Reproducing Code Example

```python
from typing import TYPE_CHECKING

import numpy as np

if TYPE_CHECKING:
    from scipy.optimize import OptimizeResult


def callback(intermediate_result: OptimizeResult) -> None:
    parameters = np.array2string(intermediate_result.x, precision=2)
    print(
        f"\u001b[1K\r{getattr(intermediate_result, 'nit', 0):>7}: {parameters}",
        end="",
    )


def test_function[F: np.floating](x: np.ndarray[tuple[int], np.dtype[F]]) -> float:
    eps = x - np.array([1.0, 2.0, 3.0])
    return float(eps @ eps)


if __name__ == "__main__":
    res = minimize(test_function, np.array([0.0, 0.0, 0.0]), callback=callback)
```

### Error message

```shell
  File "/home/willwz/Work/proj/bogoni/reproducer.py", line 30, in <module>
    res = minimize(test_function, np.array([0.0, 0.0, 0.0]), callback=callback)
  File "/home/willwz/Work/proj/bogoni/.venv/lib/python3.14/site-packages/scipy/optimize/_minimize.py", line 769, in minimize
    callback = _wrap_callback(callback, meth)
  File "/home/willwz/Work/proj/bogoni/.venv/lib/python3.14/site-packages/scipy/optimize/_optimize.py", line 93, in _wrap_callback
    sig = inspect.signature(callback)
  File "/home/willwz/.local/share/uv/python/cpython-3.14.0-linux-x86_64-gnu/lib/python3.14/inspect.py", line 3312, in signature
    return Signature.from_callable(obj, follow_wrapped=follow_wrapped,
           ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                   globals=globals, locals=locals, eval_str=eval_str,
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                   annotation_format=annotation_format)
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/willwz/.local/share/uv/python/cpython-3.14.0-linux-x86_64-gnu/lib/python3.14/inspect.py", line 3027, in from_callable
    return _signature_from_callable(obj, sigcls=cls,
                                    follow_wrapper_chains=follow_wrapped,
                                    globals=globals, locals=locals, eval_str=eval_str,
                                    annotation_format=annotation_format)
  File "/home/willwz/.local/share/uv/python/cpython-3.14.0-linux-x86_64-gnu/lib/python3.14/inspect.py", line 2502, in _signature_from_callable
    return _signature_from_function(sigcls, obj,
                                    skip_bound_arg=skip_bound_arg,
                                    globals=globals, locals=locals, eval_str=eval_str,
                                    annotation_format=annotation_format)
  File "/home/willwz/.local/share/uv/python/cpython-3.14.0-linux-x86_64-gnu/lib/python3.14/inspect.py", line 2325, in _signature_from_function
    annotations = get_annotations(func, globals=globals, locals=locals, eval_str=eval_str,
                                  format=annotation_format)
  File "/home/willwz/.local/share/uv/python/cpython-3.14.0-linux-x86_64-gnu/lib/python3.14/annotationlib.py", line 890, in get_annotations
    ann = _get_dunder_annotations(obj)
  File "/home/willwz/.local/share/uv/python/cpython-3.14.0-linux-x86_64-gnu/lib/python3.14/annotationlib.py", line 1059, in _get_dunder_annotations
    ann = getattr(obj, "__annotations__", None)
  File "/home/willwz/Work/proj/bogoni/reproducer.py", line 12, in __annotate__
    def callback(intermediate_result: OptimizeResult) -> None:
                                      ^^^^^^^^^^^^^^
NameError: name 'OptimizeResult' is not defined
```

```shell
Scipy version: 1.16.3
Numpy version: 2.3.4
Python version: sys.version_info(major=3, minor=14, micro=0, releaselevel='final', serial=0)
Config:
{
  "Compilers": {
    "c": {
      "name": "gcc",
      "linker": "ld.bfd",
      "version": "10.2.1",
      "commands": "cc"
    },
    "cython": {
      "name": "cython",
      "linker": "cython",
      "version": "3.1.6",
      "commands": "cython"
    },
    "c++": {
      "name": "gcc",
      "linker": "ld.bfd",
      "version": "10.2.1",
      "commands": "c++"
    },
    "fortran": {
      "name": "gcc",
      "linker": "ld.bfd",
      "version": "10.2.1",
      "commands": "gfortran"
    },
    "pythran": {
      "version": "0.18.0",
      "include directory": "../../tmp/build-env-hl3_u310/lib/python3.14/site-packages/pythran"
    }
  },
  "Machine Information": {
    "host": {
      "cpu": "x86_64",
      "family": "x86_64",
      "endian": "little",
      "system": "linux"
    },
    "build": {
      "cpu": "x86_64",
      "family": "x86_64",
      "endian": "little",
      "system": "linux"
    },
    "cross-compiled": false
  },
  "Build Dependencies": {
    "blas": {
      "name": "scipy-openblas",
      "found": true,
      "version": "0.3.29.dev",
      "include directory": "/opt/_internal/cpython-3.14.0rc1/lib/python3.14/site-packages/scipy_openblas32/include",
      "lib directory": "/opt/_internal/cpython-3.14.0rc1/lib/python3.14/site-packages/scipy_openblas32/lib",
      "openblas configuration": "OpenBLAS 0.3.29.dev DYNAMIC_ARCH NO_AFFINITY Haswell MAX_THREADS=64",
      "pc file directory": "/project"
    },
    "lapack": {
      "name": "scipy-openblas",
      "found": true,
      "version": "0.3.29.dev",
      "include directory": "/opt/_internal/cpython-3.14.0rc1/lib/python3.14/site-packages/scipy_openblas32/include",
      "lib directory": "/opt/_internal/cpython-3.14.0rc1/lib/python3.14/site-packages/scipy_openblas32/lib",
      "openblas configuration": "OpenBLAS 0.3.29.dev DYNAMIC_ARCH NO_AFFINITY Haswell MAX_THREADS=64",
      "pc file directory": "/project"
    },
    "pybind11": {
      "name": "pybind11",
      "version": "3.0.1",
    }
  },
  "Python Information": {
    "path": "/tmp/build-env-hl3_u310/bin/python",
    "version": "3.14"
  }
}
```

willwiz:
#### Reference issue
Closes gh-23913

Add a wrapper for `inspect.signature` with the keyword parameter `annotation_format=annotationlib.Format.FORWARDREF` if Python version >= 3.14. 

PEP 649/749, lazy evaluation of annotations, allows for scenarios where users may pass callables with annotations that are only defined when type checking or are forward-referenced. `inspect.signature` will raise a NameError in these cases, but scipy does need the annotation information.  Python 3.14 also adds the keyword parameter `annotation_format` to `inspect.signature`. Here, `annotation_format=annotationlib.Format.FORWARDREF` gives the closest behavior to pre-Python 3.14, and will return a `ForwardRef` object. None of the functions affected by this change needs the annotation information, so this should be sufficient to preserve existing functionality. Functions that are internal or for testing should not be affected. 

#### Additional information

Changes:
scipy/optimize/_minimize.py:minimize:741
- inspects user-defined `callback` parameter

scipy/optimize/_optimize.py:_wrap_callback:93
- wraps the `callback` parameter for most algorithms

scipy/optimize/_cobyla_py.py:_minimize_cobyla:249
- skips _wrapped_callback and inspects naked `callback` parameter

scipy/optimize/_basinhopping.py:BasinHoppingRunner._monte_carlo_step:132
- inspects `accept_test` keyword argument from scipy.optimize.basinhopping
- Notes: It checks if the signature matches the exact pattern of `Metropolis`. As long as `inspect.signature` doesn't raise, the actual signature of `accept_test` arguments does not matter, and should match previous functionality. 

{scipy/optimize/_minpack_py.py:curve_fit:904,scipy/optimize/_optimize.py:brute:3674}
- Uses `getfullargspec_no_self` but doesn't need its full functionality. Minor changes to match previous functionality. 

External modules:
The cobyqa submodule (scipy/_lib/cobyqa) also has this issue and skips `_wrap_callback` when called from minimize. 

Side notes on other instances where this can happen in scipy: 

scipy/stats/_resampling.py:_bootstrap_iv:176
- takes callable parameter `statistic` in scipy.stats.bootstrap
scipy/stats/_resampling.py:_monte_carlo_test_iv:724
- takes callable parameter `statistic` in scipy.stats.monte_carlo_test
scipy/stats/_resampling.py:{_wrap_kwargs:1032, _power_iv:1097}
- takes callable parameters (`test`, `rvs`) in scipy.stats.power
scipy/stats/_resampling.py:_permutation_test_iv:1032
- takes callable parameters `statistic` in scipy.stats.permutation_test
scipy/stats/_sensitivity_analysis.py:sobol_indices:618
- keyword `method` can be a user-defined callable.


scipy/_lib/_util.py
```diff
@@ -60,6 +60,28 @@
...
+if sys.version_info >= (3, 14):
+    import annotationlib
+    def wrapped_inspect_signature(callable):
+        """Get a signature object for the passed callable."""
+        return inspect.signature(callable,
+                                 annotation_format=annotationlib.Format.FORWARDREF)
+else:
+    def wrapped_inspect_signature(callable):
+        """Get a signature object for the passed callable."""
+        return inspect.signature(callable)
```
nickodell: For speed / clarity, I would suggest

```suggestion
    wrapped_inspect_signature = inspect.signature
```

scipy/optimize/_minpack_py.py
```diff
@@ -901,8 +902,16 @@ def curve_fit(f, xdata, ydata, p0=None, sigma=None, absolute_sigma=False,
...
+        #
+        # NOTE: This implementation is made to match previous behavior. However,
+        #       inspect.Parameter.POSITIONAL_ONLY parameters would still match
+        #       documented signature, e.g., f(x, a_1, a_2, / ,...), but are
+        #       ignored here. Can trigger misleading behavior.
+        args = [
+            v for v in wrapped_inspect_signature(f).parameters.values()
+            if v.kind in [inspect.Parameter.POSITIONAL_OR_KEYWORD,
+                          inspect.Parameter.KEYWORD_ONLY]
+        ]
```
nickodell: Rather than make this change, can you fix `getfullargspec_no_self()` instead? That has many callers, and likely almost none of them care about annotations. (The name refers to a legacy API, so anything calling this was likely written before annotations existed.)

scipy/optimize/_optimize.py
```diff
@@ -3897,7 +3897,12 @@ def brute(func, ranges, args=(), Ns=20, full_output=0, finish=fmin,
 
     if callable(finish):
         # set up kwargs for `finish` function
-        finish_args = _getfullargspec(finish).args
+        _finish_args = wrapped_inspect_signature(finish)
+        finish_args = [
+            p.name for p in _finish_args.parameters.values()
+            if p.kind in [inspect.Parameter.POSITIONAL_OR_KEYWORD,
+                          inspect.Parameter.KEYWORD_ONLY]
+        ]
```
nickodell: Rather than make this change, can you fix `getfullargspec_no_self()` instead? That has many callers, and likely almost none of them care about annotations. (The name refers to a legacy API, so anything calling this was likely written before annotations existed.)

nickodell (changes requested):
Nice work here. Can you add a test for this behavior?

I would like to see a test which fails on 3.14 if this patch were not applied.

The file `scipy/scipy/optimize/tests/test_optimize.py` is probably a sensible place for it.

Since the test is testing something which is a NameError in a previous version of Python, you might need to write the test such that it selectively defines an import if running on pre-Python 3.14 systems.

e.g.

```
if sys.version < (3, 14):
    from typing import List

def callback(...):  # Uses List in an annotation somehow
    ...
```

This is just an example, you can probably think of a better idea. (Perhaps you could also selectively define the test in an if?)

willwiz:
The suggested changes are integrated. Also added the tests for these issues. 

These tests should be skipped in 3.11 -- or at least they are on my computer. There should be 17 fails on 3.14 without these changes -- `method="tnc"` currently does not implement this type of inspection.

nickodell:
> The suggested changes are integrated. Also added the tests for these issues.

Thanks. I see a few test failures that are caused by issues elsewhere. See #23932

>These tests should be skipped in 3.11 -- or at least they are on my computer. There should be 17 fails on 3.14 without these changes - method="tnc" currently does not implement this type of inspection.

Sorry - by "should" do you mean that they are being skipped right now, or that they ought to be skipped?

willwiz:
> > These tests should be skipped in 3.11 -- or at least they are on my computer. There should be 17 fails on 3.14 without these changes - method="tnc" currently does not implement this type of inspection.
> 
> Sorry - by "should" do you mean that they are being skipped right now, or that they ought to be skipped?

Sorry, I worded that a bit weirdly. 

I marked these tests to be skipped with pytest if the version is < 3.14 with this
```
@pytest.mark.skipif(sys.version_info < (3, 14), reason="Requires PEP 649/749 from Python 3.14+.")
```
However, I only locally tested that these changes behave correctly with spin on Python 3.11 and 3.14. I am assuming that this should be fine for other setups and the tests that are done on CI.

nickodell:
Hi - Sorry for going silent for a few days. I'm still reviewing this.

I have a few TODOs before merging.

- [x] Check that removing annotations did not break docs (One user of getfullargspec_no_self is inside `docs/`.)
   - [x] Spot check
   - [x] Full doc rebuild, then use `diff --recursive` to compare the two documentation version.
- [x] Check that these changes have statement / branch coverage

   - _lib/util.py:545
   - basinhopping.py:133
   - _cobyla_py.py:248
   - _minimize:741
   - _optimize.py:93

- [x] Check that we're correct about the versions of Python where the syntax works and fails.
   - Yes. The test fails on 3.13 without being skipped there.
- [x] Check that attached test fails without the patch.
- [x] This is a fairly subtle change with different behavior on different versions, so I'd like to get CI green before merging.
   - [x] Pythan failure: https://github.com/scipy/scipy/actions/runs/19222125724/job/54941950948?pr=23922#step:7:344 (Issue filed at #23958)
   - [x] Rtools failure: https://github.com/scipy/scipy/actions/runs/19222125699/job/54941947556?pr=23922#step:4:38
   - [x] `ValueError: axes don't match array` #23997
   
   I think that covers all the failures without existing issues. (Boy, this is really the week when all our tooling decided to quit on us.)
- [x] Fix tach error: `[FAIL] scipy/_lib/_util.py:73: Dependency 'annotationlib' is not declared in package 'scipy'.` This looks like it's caused by the new dependency inside _util.py. `tach` needs to be told that this dependency is allowed.

andyfaff:
chocolatey failure works on re-run.

willwiz:
> Hi - Sorry for going silent for a few days. I'm still reviewing this.
> 
> I have a few TODOs before merging.

Not a big deal. Take your time.

willwiz:
In regards to the tach error with `annotationlib`, tach gets its list of std libs from the Python interpreter as 
```python
set(sys.builtin_module_names).union(sys.stdlib_module_names).union(KNOWN_MODULE_SPECIAL_CASES)
```
On my end, locally, tach doesn't flag `annotationlib` if I am using Python 3.14, but does flag it if I'm on Python 3.11. It seems like the logic for handling version-dependent conditional imports is not there. Maybe, it would be more appropriate to add `annotationlib` temporarily to tach.toml under `[external] exclude` until support for Python 3.13 is dropped?

nickodell:
> Maybe, it would be more appropriate to add `annotationlib` temporarily to tach.toml under `[external] exclude` until support for Python 3.13 is dropped?

That seems reasonable to me. Go for it.

But I think you and I have different definitions of temporary. :) We will be dropping support for 3.13 sometime in 2028.

nickodell (approved):
Thanks, looks good to me. As noted [here](https://github.com/scipy/scipy/pull/23922#issuecomment-3509634492), there is a test failure remaining, but it doesn't look very scary to me.

willwiz:
For the error with `test_basinhopping_annotations` on [Linux tests / free-threaded (0) (pull_request)](https://github.com/scipy/scipy/actions/runs/19464637378/job/55696879797?pr=23922):
I can replicate this locally with `spin test -j4 -m full --durations=10`, where the results did not converge within the tolerance. However, this did not happen with any other combinations of 
`spin test -j4 -s optimize --durations=10`
`spin test -j4 -t scipy.optimize --durations=10`
`spin test -j4 -t scipy.optimize.tests.test_optimize --durations=10`
`spin test -j4 -t scipy.optimize.tests.test_optimize::TestAnnotations --durations=10`

I suspect that this is just due to bad luck with how the RNG is advanced on the free-threaded version and this combination of tests. If I explicitly set `seed=1`, this test now converges similarly to other versions. 

I'm not sure about the errors with `pixi`, and I guess the build doc errors on https request to numpy are kind of just bad luck as well?

nickodell:
> I'm not sure about the errors with `pixi`, and

Probably [Cloudflare](https://downdetector.com/status/cloudflare/) related. prefix.pixi.dev is hosted on Cloudflare.

> I guess the build doc errors on https request to numpy are kind of just bad luck as well?

NumPy.org is hosted on GitHub pages and Cloudflare. So that could be the cause of this intersphinx error:

```
WARNING: failed to reach any of the inventories with the following issues:
intersphinx inventory 'https://numpy.org/neps/objects.inv' not fetchable due to <class 'requests.exceptions.HTTPError'>: 500 Server Error: Internal Server Error for url: https://numpy.org/neps/objects.inv
WARNING: failed to reach any of the inventories with the following issues:
intersphinx inventory 'https://numpy.org/doc/stable/objects.inv' not fetchable due to <class 'requests.exceptions.HTTPError'>: 500 Server Error: Internal Server Error for url: https://numpy.org/doc/stable/objects.inv
WARNING: failed to reach any of the inventories with the following issues:
intersphinx inventory 'https://matplotlib.org/stable/objects.inv' not fetchable due to <class 'requests.exceptions.HTTPError'>: 500 Server Error: Internal Server Error for url: https://matplotlib.org/stable/objects.inv
myst v4.0.1: MdParserConfig(commonmark_only=False, gfm_only=False, enable_extensions={'dollarmath', 'l
```

If it can't get the NumPy inventory, then it will not be able to tell if any of the NumPy doc links are correct.

For the [GitHub checkout errors](https://downdetector.com/status/github/), I think something similar is going on there.

willwiz:
I guess the PyTorch error for test_quadrature::TestQMCQuad::basic_test in [Array API / Linux PyTorch/JAX/Dask/xp-strict CPU (pull_request)](https://github.com/scipy/scipy/actions/runs/19478593855/job/55952455071?pr=23922)   was addressed in gh-24023? The values seem to be the same as in https://github.com/scipy/scipy/pull/23667#issuecomment-3543962761.

nickodell:
@willwiz I wouldn't worry about that one - I don't think it's plausibly downstream of this change.

@andyfaff I'm inclined to merge this via 'Squash and merge,' to get rid of some of the merge noise above. Do you want to take more time to review it, or are you happy with it as-is?

andyfaff:
I won't have time to look at it again.

nickodell:
> I won't have time to look at it again.

No worries - thank you for helping to debug the CI issue.

PR #23923, opened 2025-11-04
BUG: sparse: remove spurious CSC to CSR conversions in binops

### LINKED ISSUE #23826: BUG: Performance of `!=` for non-CSR matrices
### Describe your issue.

Warning, this is quite pathological and only came up in my unit testing.

I think the following line:

https://github.com/scipy/scipy/blob/5568abf464c611d2865f6022fb1fd6dc8a10ddb2/scipy/sparse/_base.py#L666

blows up memory on `!=` checks for CSC matrices because the `indptr` can be quite large in that case.  Say you have a `(np.iinfo(np.uint32).max + 1, 20)` matrix - the `indptr` here will only have 20 entries.  But if you compare it to another CSC matrix of the same shape, it will be converted to `csr` and the `indptr` will be `np.iinfo(np.uint32).max + 1` long, which can be quite bad.

I assume this is the case for any matrix that has a big row axis but is still very sparse.

### Reproducing Code Example

```python
import numpy as np
import scipy.sparse as sp

num_minor_axis = np.iinfo(np.uint32).max + 1
minor_axis_index = np.array([num_minor_axis - 1])
major_axis_index = np.array([10])
row_cols = (minor_axis_index, major_axis_index)
shape = [num_minor_axis, 20]
X = sp.csc_array(
    (np.array([10]), row_cols),
    shape=shape,
)
X_2 = X.copy()
X != X_2 # really bad, OOM I think
```

### Error message

```shell
N/A
```

```shell
1.16.2 2.3.4 sys.version_info(major=3, minor=12, micro=3, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    found: true
    include directory: unknown
    lib directory: unknown
    name: Accelerate
    pc file directory: unknown
    version: unknown
  lapack:
    found: true
    include directory: unknown
    lib directory: unknown
    name: Accelerate
    pc file directory: unknown
    version: unknown
  pybind11:
    detection method: config-tool
    include directory: unknown
    name: pybind11
    version: 3.0.1
Compilers:
  c:
    commands: cc
    linker: ld64
    name: clang
    version: 15.0.0
  c++:
    commands: c++
    linker: ld64
    name: clang
    version: 15.0.0
  cython:
    commands: cython
    linker: cython
    name: cython
    version: 3.1.3
  fortran:
    commands: gfortran
    linker: ld64
    name: gcc
    version: 13.4.0
  pythran:
    include directory: ../../../../../../private/var/folders/y6/nj790rtn62lfktb1sh__79hc0000gn/T/build-env-jojwhclh/lib/python3.12/site-packages/pythran
    version: 0.18.0
Machine Information:
  build:
    cpu: aarch64
    endian: little
    family: aarch64
    system: darwin
  cross-compiled: false
  host:
    cpu: aarch64
    endian: little
    family: aarch64
    system: darwin
Python Information:
  path: /private/var/folders/y6/nj790rtn62lfktb1sh__79hc0000gn/T/build-env-jojwhclh/bin/python
  version: '3.12'
```

dschult:
Fixes #23826 

The nD binops changes inadvertently caused CSC inputs to be converted to CSR. This can cause large memory allocations for short but wide arrays. This PR also checks for spurious conversions from CSR to CSC for completeness. No bug exsists for that case. 

This only affects binops which act on two sparse arrays and produce output that is both of sparse type and low density. That is, the operations `lt`, `gt`, `ne`, `minimum` and `maximum` are changed. The operations `add`, `sub` and `mul` already avoid spurious conversions. The operations that don't handle this case well are `truediv`, `le`, `ge`, `eq`; but they have never been able to handle this case well -- the answer itself will use lots of memory. Comparisons `le`, `ge`, `eq` create (after a SparseEfficiencyWarning) a sparse output type that is actually quite dense due to implicit zero values creating `True` results. And `truediv` returns a numpy/dense object with lots of `nan` values.

A workaround for users running into this is to operate on the transposes of the two matrices (which turns them into tall skinny arrays): `(A.T != B.T).T` The transposes change the format, but do not need to allocate any new memory.

 This defect has been in all 1.16 releases. @perimosocordiae and @tylerjereddy: Is this fix sufficient to make a `1.16.4` release? This issue's discussion suggests that short, wide arrays are rarely used, but this bug makes them effectively unusable with `ne`, `gt`, `lt` and `minimum`/`maximum`.

The first commit includes only the new tests. The second commit then fixes the broken ones.


scipy/sparse/tests/test_csc.py
```diff
@@ -96,3 +98,28 @@ def test_argmax_overflow(ax):
         ii, jj = np.max(idx), np.argmax(idx)
 
     assert A[ii, jj] == A[-2, -2]
+
+
+# broken in 1.16: comparison, truediv, maximum, minimum.
+#   truediv becomes dense for floats so has always broken with this kind of shape
+@pytest.mark.timeout(2)  # only slow when broken (when spurious conversion occurs)
+@pytest.mark.parametrize("op", (operator.ne, operator.lt, operator.gt, operator.mul,
+                                operator.mul, operator.add, operator.sub,
```
tylerjereddy: is `operator.mul` duplicated here?

scipy/sparse/tests/test_csc.py
```diff
@@ -96,3 +98,28 @@ def test_argmax_overflow(ax):
         ii, jj = np.max(idx), np.argmax(idx)
 
     assert A[ii, jj] == A[-2, -2]
+
+
+# broken in 1.16: comparison, truediv, maximum, minimum.
+#   truediv becomes dense for floats so has always broken with this kind of shape
+@pytest.mark.timeout(2)  # only slow when broken (when spurious conversion occurs)
```
tylerjereddy, line 110: For reference, when I revert the source patch this test does indeed fail this way, but takes a while--`5 failed, 15 passed in 58.05s` (via `spin test -t scipy/sparse/tests/test_csc.py`).

I tried using `pytest-memray` with `@pytest.mark.limit_memory("200 MB")` instead of the timeout as suggested in the ticket, but it was even slower--indeed, I lost patience and hit Ctrl+C after two tests reported an `M` status (presumably memory failures). I like the idea of failing a test on the basis of its apparent space complexity rather than the timing, but in practice this seems to be very slow rather than quickly terminating the in-progress allocation of mem. If it were really fast it might be worth considering that plugin, but in my hands it was way too sluggish. I suppose adding more fixtures/plugins also creates complexity for free-threading support...

tylerjereddy (commented):
I added a few comments inline.

As for this warranting a `1.16.4` release--I'd like to see a few folks in favor maybe--Ralf, CJ? We just did `1.16.3`, so I'm a little hesitant. I do have a bit of funding to do releases, so if there's a demand for it and it makes sense I'll just do it. We haven't had too many complaints about this one though, right?

scipy/sparse/tests/test_csc.py
dschult: The docs for `limit_memory` say that the process runs to completion and then the decision about whether to fail or not is made based on peak memory usage. That aligns with this timing experiment -- Thanks!
The `timeout` approach stops the process using a process signal, and that takes a minimum of 2 seconds, but could be much longer depending on when  the process next checks for signals.

Luckily, the test is quite fast when the code is correct. It only takes a long time when it is failing. And hopefully it doesn't fail very often.

Thanks @tylerjereddy for your take on the release choice. You are correct that we haven't received reports of trouble other than from @ilan-gold reporting it based on unit tests rather than any users based on an application. But my work with sklearn on converting to sparrays suggests that at least sklearn does use CSC format a fair amount even though csr_matrix appears about 3 times more. The [doc advice in a list of practical tips](https://scikit-learn.org/stable/modules/tree.html) at the end of section 1.10.5 says to use a `csc_matrix` when calling "fit" and `csr_matrix` when calling "predict".  It is also true that this doesn't produce a numerically incorrect result -- it allocates a large amount of memory that might trigger an out of memory error, otherwise it takes a while and results the correct answer. And there is a workaround of using transposes to turn the CSC into a CSR, operating on that and transposing the result.

I'm fine with either choice for milestone depending on library norms and expectations.

tylerjereddy:
@perimosocordiae @rgommers opinions on severity re: release justification?

scipy/sparse/tests/test_csc.py
rgommers: Timing based tests can be very fragile, e.g. if the tests are run under an emulator they can be very slow as well. If you want to keep this check, please limit it to Linux x86-64 only.

rgommers:
> This defect has been in all 1.16 releases. @perimosocordiae and @tylerjereddy: Is this fix sufficient to make a `1.16.4` release? This issue's discussion suggests that short, wide arrays are rarely used, but this bug makes them effectively unusable with `ne`, `gt`, `lt` and `minimum`/`maximum`.

That does seem like the kind of regression that may warrant doing a 1.16.4 release. On the other hand, it's only been noticed last week so after a couple of months since 1.16.0. Which suggests that it's at least not very urgent to do a new bug fix release immediately. I'd leave it on the 1.16.4 milestone and see if anything else shows up. If no more people complain and there are no other reasons to do 1.16.4, it seems fine to not do that release.

scipy/sparse/tests/test_csc.py
dschult: I'm afraid I don't know a good way to run a test on Linux x86-64 only.  The `mark.slow` approach is mentioned in the [continuous integration doc](https://docs.scipy.org/doc/scipy/dev/contributor/continuous_integration.html#failures-due-to-test-duration) as being good for this purpose when you don't need it tested on all types of machines. I tried that and it runs on Windows, Linux and MacOS. Is there a `mark` I can use to run on Linux x86-64 only? Or another way to run tests only on Linux x86_64? `fail-slow` runs to completion before checking the time, so `timeout` seems better -- but I could try fail-slow if that would limit where it runs.

scipy/sparse/tests/test_csc.py
```diff
@@ -96,3 +98,28 @@ def test_argmax_overflow(ax):
...
     assert A[ii, jj] == A[-2, -2]
+
+
+@pytest.mark.slow
+@pytest.mark.timeout(2)  # only slow when broken (when spurious conversion occurs)
+@pytest.mark.parametrize("op", (operator.ne, operator.lt, operator.gt,
+                                operator.add, operator.sub, operator.mul,
+                                # truediv, eq, ge, le make dense output so not tested
+                                lambda x, y: x.minimum(y), lambda x, y: x.maximum(y)))
+def test_compressed_rc_conversion_mixup(op):
```
perimosocordiae, line 115: Let's leave a comment here that refers to the related issue.

scipy/sparse/tests/test_csc.py
perimosocordiae: The combination of slow and timeout here seems practical enough to me. It would be ideal if we had some way to run tests with a custom allocator that proactively fails if we try to malloc too much at once, but unless that exists I'm fine with the current approach.

perimosocordiae (approved):
It's unfortunate that we need to duplicate a bit of the logic for ndim > 2, but overall the change makes sense to me.

tylerjereddy:
> I'm afraid I don't know a good way to run a test on Linux x86-64 only.

Checking in our codebase with `git grep -E -i "platform\.machine.*x86"`:

```
scipy/sparse/linalg/_isolve/tests/test_iterative.py:            platform.machine() not in ['x86_64' 'x86', 'aarch64', 'arm64']):
scipy/stats/tests/test_distributions.py:MACOS_INTEL = (sys.platform == 'darwin') and (platform.machine() == 'x86_64')
```

That second example basically gets you there--swap `linux` in for `darwin` and then use in one of the standard `pytest` `mark` or imperative constructs.

ev-br:
> It would be ideal if we had some way to run tests with a custom allocator that proactively fails if we try to malloc too much at once

Cross-ref https://github.com/scipy/scipy/issues/23396 where such an allocator was wanted, too (nothing to do in this PR though)

dschult:
Thanks for that solution @rgommers.  I'll mark it as slow and also mark it with timeout.
It should be fine to run on only one system type. The difference in time between working and broken is about 400%:  0.16s to 60s. So I think testing with `pytest.mark.slow` should be fine.

In summary, I believe we have decided to go ahead and merge this into main when its ready, leave the 1.16.4 milestone for now and reset that to 1.17 if nothing else happens in the next month.

lucascolley:
thanks all, let's put it in.

PR #23929, opened 2025-11-04
ENH: stats.epps_singleton_2samp: add array API support

mdhaber:
#### Reference issue
Toward gh-20544

Adds array API support to `scipy.stats.epps_singleton_2samp`.

#### Additional information
Updates `array_api_extra` to bring in the new vectorized `cov`.


mdhaber:
Most CI failures are https://github.com/data-apis/array-api-extra/pull/454. I'll wait for that to be resolved before re-running CI.

scipy/stats/_hypotests.py
```diff
@@ -29,13 +29,8 @@
...
 
-# remove when array-api-extra#502 is resolved
-@_apply_over_batch(('x', 2))
-def cov(x):
-    return xpx.cov(x)
-
-
-@xp_capabilities(np_only=True)
+@xp_capabilities(skip_backends=[("dask.array", "lazy -> no _axis_nan_policy")],
+                 jax_jit=False)
```
mdhaber: Looks like this resolves the CI failures, although it's hard to tell with all the noise.
```suggestion
@xp_capabilities(skip_backends=[("dask.array", "lazy -> no _axis_nan_policy"),
                                ("jax.numpy', 'lazy -> no _axis_nan_policy")])
```

lucascolley:
The following rudimentary diff makes the tests pass with NumPy for me:

```diff
diff --git a/scipy/spatial/transform/_rigid_transform.py b/scipy/spatial/transform/_rigid_transform.py
index 4d40b98ea4..e58954fee4 100644
--- a/scipy/spatial/transform/_rigid_transform.py
+++ b/scipy/spatial/transform/_rigid_transform.py
@@ -410,3 +410,3 @@ class RigidTransform:
         if self._single:
-            matrix = xpx.atleast_nd(matrix, ndim=3, xp=xp)
+            matrix = xp.expand_dims(matrix, axis=0)

@@ -993,5 +993,14 @@ class RigidTransform:
         xp = array_namespace(transforms[0].as_matrix())
-        matrix = xp.concat(
-            [xpx.atleast_nd(x.as_matrix(), ndim=3, xp=xp) for x in transforms]
-        )
+        expanded_transforms = []
+        for x in transforms:
+            x = x.as_matrix()
+            if x.ndim == 3:
+                expanded_transforms.append(x)
+            elif x.ndim == 2:
+                expanded_transforms.append(xp.expand_dims(x, axis=0))
+            else:
+                raise ValueError("panic")
+        matrix = xp.concat(expanded_transforms)
         return RigidTransform._from_raw_matrix(matrix, xp, None)
@@ -1936,3 +1945,3 @@ class RigidTransform:
         if tf._single:
-            matrix = xpx.atleast_nd(matrix, ndim=3, xp=xp)
+            matrix = xp.expand_dims(matrix, axis=0)
         tf._matrix = matrix
```

@scottshambaugh @amacati what do you think about merging a polished version of this diff?

Long-term I think we should just enforce that `atleast_nd` adds dimensions to the front, so the existing code would be fine. But right now it is relying on undefined behaviour.

EDIT: context: https://github.com/data-apis/array-api-extra/pull/454#issuecomment-3491388005

amacati:
At first glance this seems incorrect. 

```diff
+            x = x.as_matrix()
+            if x.ndim == 3:
+                expanded_transforms.append(x)
+            elif x.ndim == 2:
+                expanded_transforms.append(xp.expand_dims(x, axis=0))
+            else:
+                raise ValueError("panic")
```
`RigidTransform` can be a batch of multiple leading dimensions. Concatenating several 4D `RigidTransform`s is well-defined, but would error with this loop, right?

lucascolley:
in which case we should add test coverage for that too!

amacati:
Yes, I am surprised that this was not picked up already.

amacati:
Yeah, looking at the concatenate tests, the Nd case is not covered. It would also make sense to introduce an axis argument, but that's maybe something for later, or should be implemented if people actually need it. I will try to expand the test coverage asap, just a bit short of time these days.

mdhaber:
@amacati @lucascolley I merged main here and expected the spatial tests to pass based on the discussion and linked PRs, but I still see failures. In the PR to fix them, could you update `array-api-extra` to make sure tests still pass?

mschuck-rai:
Huh, since when are these tests failing? Does this come with the latest version of `array-api-extra` that fixes `atleast_nd`?

mdhaber:
> since when are these tests failing

Since I tried to update array-api-extra in this PR. That's what motivated the discussion above (https://github.com/scipy/scipy/pull/23929#issuecomment-3488818698).

mdhaber:
I didn't realize array-api-extra was updated, too; I thought it was just https://github.com/scipy/scipy/pull/23937. I'll update that in this PR.

lucascolley:
yeah, `np.atleast_3d` has some quite odd semantics, simply avoiding delegating to NumPy seemed like the most reasonable way forward.

scipy/stats/_hypotests.py
```diff
@@ -125,42 +120,45 @@ def epps_singleton_2samp(x, y, t=(0.4, 0.8), *, axis=0):
...
     # values with a harmless one, and replace results with NaN at the end.
-    i_x = ~np.isfinite(x)
-    i_y = ~np.isfinite(y)
-    x = xpx.at(x)[i_x].set(1)
-    y = xpx.at(y)[i_y].set(1)
-    invalid_result = np.any(i_x, axis=-1) | np.any(i_y, axis=-1)
+    i_x = ~xp.isfinite(x)
+    i_y = ~xp.isfinite(y)
+    # x = xpx.at(x)[i_x].set(1.)  # Can use this when data-apis/array-api-extra#506
+    # y = xpx.at(y)[i_y].set(1.)  # is resolved
```
lucascolley: should we just remove this now?

lucascolley (approved):
LGTM, thanks Matt. I might look at doing an array-api-extra release, it has been a while and there is something nice about bumping to a tagged commit.

scipy/stats/_hypotests.py
lucascolley: if we are green, feel free to resolve with `[skip ci]` and merge however you see fit.

mdhaber: I don't consider that issue to be resolved. array-api-extra introduced `at` to meet a clear need - a simple way to express "mutate when possible and copy otherwise". This is an example of that need not being fully met. That's not necessarily array-api-extra's fault - definitely the standard itself needs to address the larger problem at some point. But I think it's fair to keep some sort of reminder here that copying all data to deal with an exceptional case is not the ideal implementation choice. (Or if it is, then I'm not sure there is *ever* a time to prefer `at.set` over `where`.)

lucascolley: sure, I wonder how to make the comment a bit less cryptic. Maybe just change "when" to "if"? That seems more appropriate given that it isn't entirely clear whether that issue will be resolved for the foreseeable. Could also spell out something about "could avoid copies" for clarity, not sure exactly how to word it.

mdhaber:
```suggestion
    # Ideally we would avoid copying all data here; see
    # discussion in data-apis/array-api-extra#506.
```
?

lucascolley: Yep looks good, thanks

PR #23878, opened 2025-10-28
ENH: stats.wasserstein_distance/energy_distance: vectorize implementations

mdhaber:
#### Reference issue
Toward gh-14651
Toward gh-20544

#### What does this implement/fix?
This vectorizes the implementations of `scipy.stats.wasserstein_distance` and `scipy.stats.energy_distance`. 

#### Additional information 
The `_axis_nan_policy` decorator cannot be applied to these functions (without modification - because weights and corresponding samples are paired, but the two samples are not paired), so we need to do some broadcasting and shape manipulation manually.

Some things that used to raise errors now causes NaNs to appear in the corresponding axis-slices. This is acceptable, as discussed in https://github.com/scipy/scipy/pull/20496#discussion_r1572401090.

Much of the code is array API compatible, but array API compatibility will be tested in a follow-up PR.

Issue #22118, opened 2024-12-18
Consider `x.ndim > 1`  generalizations of spline and polynomial interpolators


ev-br:
(splitting from https://github.com/scipy/scipy/pull/22101#discussion_r1887812794)

In that thread @mdhaber started a discussion on extending spline-based interpolators to allow `x.ndim > 1`. Apparently, it's needed for some scipy.stats applications: https://github.com/scipy/scipy/pull/22101#discussion_r1889224026 gives context.

Let's first sketch the needs and maybe prototype things --- my knee-jerk reaction is that it's best implemented as a layer on top of existing `BSpline`, `PPoly` and `BPoly` classes which only allow `x.ndim == 1`. So, let's sketch the target API.
As a very first step, here's the current API:

- if `x` has shape `(npts,)` and `y` has shape `(npts,) + extra_tpl`, the first axis of `x` and `y` is the interpolation axis, and batch dimensions are _trailing_, (not leading as in ufuncs).
- Evaluating the interpolant an an `xv` array with shape `xv.shape`, the shape of the result is `xv.shape + extra_tpl`.

Demo:

```
In [30]: x = np.arange(5)

In [31]: y = np.ones((5, 4, 5, 6))

In [32]: spl(1).shape
Out[32]: (4, 5, 6)

In [33]: spl([1, 2]).shape
Out[33]: (2, 4, 5, 6)

In [34]: xv = np.asarray([[1, 2, 3], [4, 5, 6]])

In [35]: spl(xv).shape
Out[35]: (2, 3, 4, 5, 6)
``` 

Immediate concerns for if we allow `x.ndim > 1`:
- what is the rule for the result shape?
- in real applications, is there a need for `x` subarrays of different lengths?

mdhaber:
> what is the rule for the result shape?

In the 1D case, `x` and `y` must have the same shape, so in N-D, the two must be broadcastable.

The "shape of the *interpolator*" is the broadcasted shape of `x` and `y` reduced along an `axis`, _keeping_ the dimension of length 1. (Same rule as for any paired-sample statistic, like `wilcoxon`, with `keepdims=True`.)

The shape of the output of `__call__(x)` is the broadcasted shape of the interpolator and the input `x`. Typically, `x` will have the same shape as the interpolator except it will have length `m` along `axis` instead of length `1`.)

> in real applications, is there a need for x subarrays of different lengths?

[...]

PR #23930, opened 2025-11-04
ENH: stats._xp_searchsorted: private vectorized searchsorted

mdhaber:
#### Reference issue
Toward gh-14651
Toward gh-20544

Several stats functions use `searchsorted`, making them difficult to vectorize because `np.searchsorted` only accepts 1D arrays for the first argument (gh-4224). We can often get by reimplementing these functions in terms of `rankdata` and such (e.g. gh-23793, gh-23878), but it would likely be better (easier to write/review in the short term, more performant in the long term) to continue using `searchsorted`. To enable this, this PR implements a (private,) array-API compatible, vectorized version of `searchsorted`. 

#### Additional information
Despite using a plain-python `for` loop to perform bisection,  it's actually faster than `np.searchsorted` for sufficiently large second arguments.
```python3
rng = np.random.default_rng(2348345982435888453)
x = np.sort(rng.random(size=1000000), axis=-1)
y = rng.random(size=100000)
ref = np.searchsorted(x, y)
# 29.8 ms ± 1.06 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
res = _xp_searchsorted(x, y)
# 14.4 ms ± 150 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
np.testing.assert_equal(res, ref)
```

Surprisingly, it may not be faster for batches than looping over slices of the first argument, at least for NumPy. ~~I'll play around with the other backends to see if it might be better to just loop for now.~~ (Can be quite a bit faster - like 150x - than looping for `cupy`. For torch, we could delegate to their version of searchsorted, which is vectorized.) Ultimately, though, I would like to get `np.searchsorted` vectorized and work toward getting the array API standard updated to require vectorization; this is just a way to move forward until that happens.

### DIFF
--- a/scipy/stats/_quantile.py
+++ b/scipy/stats/_quantile.py
@@ -1,3 +1,4 @@
+import math
 import numpy as np
 from scipy.special import betainc
 from scipy._lib._array_api import (
@@ -383,3 +384,35 @@ def _quantile_bc(y, p, n, method, xp):
     elif method == '_nearest':
         k = xp.round(ij)
     return xp.take_along_axis(y, xp.astype(k, xp.int64), axis=-1)
+
+
+
+@xp_capabilities(skip_backends=[("dask.array", "No take_along_axis yet.")])
+def _xp_searchsorted(x, y, *, side='left', xp=None):
+    # Vectorize np.searchsorted. Assumes search is along last axis, which is always
+    # preserved in the output. Does not support zero-length `x`. For side='right',
+    # NaNs in `y` are inserted to the left, in contrast with np.searchsorted.
+    xp = array_namespace(x, y) if xp is None else xp
+    x, y = _broadcast_arrays((x, y), axis=-1, xp=xp)
+
+    a = xp.full(y.shape, 0)
+    n = xp.count_nonzero(~xp.isnan(x), axis=-1, keepdims=True)
+    b = xp.broadcast_to(n, y.shape)
+
+    if side=='right':
+        n_nans = x.shape[-1] - n
+        a, b = a + n_nans, b + n_nans
+        b = xp.where(n > 0, b, b - 1)  # handle all nan case?
+        x, y = -xp.flip(x, axis=-1), -y
+
+    # while xp.any(b - a > 1):
+    # refactored to for loop with ~log2(n) iterations for JAX JIT
+    for i in range(int(math.log2(x.shape[-1])) + 1):
+        c = (a + b) // 2
+        x0 = xp.take_along_axis(x, c, axis=-1)
+        j = x0 >= y
+        b = xp.where(j, c, b)
+        a = xp.where(j, a, c)
+
+    b = xp.where(y <= xp.min(x, axis=-1, keepdims=True), 0, b)
+    return b if side == 'left' else  x.shape[-1] - b
--- a/scipy/stats/_quantile.py
+++ b/scipy/stats/_quantile.py
@@ -415,4 +415,4 @@ def _xp_searchsorted(x, y, *, side='left', xp=None):
         a = xp.where(j, a, c)
 
     b = xp.where(y <= xp.min(x, axis=-1, keepdims=True), 0, b)
-    return b if side == 'left' else  x.shape[-1] - b
+    return b if side == 'left' else x.shape[-1] - b
--- a/scipy/stats/_quantile.py
+++ b/scipy/stats/_quantile.py
@@ -395,7 +395,7 @@ def _xp_searchsorted(x, y, *, side='left', xp=None):
     xp = array_namespace(x, y) if xp is None else xp
     x, y = _broadcast_arrays((x, y), axis=-1, xp=xp)
 
-    a = xp.full(y.shape, 0)
+    a = xp.full(y.shape, 0, device=xp_device(x))
     n = xp.count_nonzero(~xp.isnan(x), axis=-1, keepdims=True)
     b = xp.broadcast_to(n, y.shape)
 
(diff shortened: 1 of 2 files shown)


lucascolley:
> Ultimately, though, I would like to get `np.searchsorted` vectorized and work toward getting the array API standard updated to require vectorization; this is just a way to move forward until that happens.

Would you consider contributing this to array-api-extra in the meantime? We have precedence for similar extensions to existing APIs, like a tuple of axes in https://data-apis.org/array-api-extra/generated/array_api_extra.expand_dims.html.

That doesn't have to block this PR, and may not be preferable depending on the prospects for standardisation. But may be nice.

mdhaber:
I would. It would require a little edge case tweaking to be public, and those don't need to be perfect to suit SciPy's internal needs, so I'd like to start with this.

scipy/stats/tests/test_quantile.py
```diff
@@ -210,3 +211,40 @@ def test_transition(self, method, xp):
...
+        x_ = np.asarray([-np.inf, np.inf])
+        x_ = np.broadcast_to(x_, x.shape[:-1] + (2,))
+        y = rng.permuted(np.concatenate((xl, x, xr, x_), axis=-1), axis=-1)
+        if nans:
+            # all elements of a slice are NaN and side='right',
+            # _xp_searchsorted has a different convention
+            mask = rng.random(shape) < 0.1
+            x[mask] = np.nan
+        x = np.sort(x, axis=-1)
+        x, y = np.astype(x, np.float64), np.astype(y, np.float64)
```
mdhaber: This should be all it takes to resolve CI failures, so I'll wait for something else to commit. Could use `astype` method, but I'd prefer to use something that's in the standard, and I don't want to use `array_namespace`.
```suggestion
        x, y = np.asarray(x, dtype=np.float64), np.asarray(y, dtype=np.float64)
```

scipy/stats/_quantile.py
```diff
@@ -383,3 +384,35 @@ def _quantile_bc(y, p, n, method, xp):
...
+
+@xp_capabilities(skip_backends=[("dask.array", "No take_along_axis yet.")])
+def _xp_searchsorted(x, y, *, side='left', xp=None):
+    # Vectorize np.searchsorted. Assumes search is along last axis, which is always
+    # preserved in the output. Does not support zero-length `x`. For side='right',
+    # NaNs in `y` are inserted to the left, in contrast with np.searchsorted.
+    xp = array_namespace(x, y) if xp is None else xp
+    x, y = _broadcast_arrays((x, y), axis=-1, xp=xp)
+
+    a = xp.full(y.shape, 0)
```
ogrisel: If `x` is allocated on non-CPU device (e.g. PyTorch on MPS device) the subsequent call to `a + n_nans` would fail:

```suggestion
    a = xp.full(y.shape, 0, device=xp_device(x))
```

ogrisel: Note this can also be tested using `array-api-strict` non-default devices such as `array_api_strict.Device("device2")`.

lucascolley: yes we typically test for this for functions in array-api-extra with https://github.com/data-apis/array-api-extra/blob/95291456b7cbbe1adaad6e4cf56f284ff2b3441f/tests/conftest.py#L219-L234, e.g. https://github.com/data-apis/array-api-extra/blob/95291456b7cbbe1adaad6e4cf56f284ff2b3441f/tests/test_funcs.py#L357-L359

scipy/stats/_quantile.py
```diff
@@ -383,3 +384,35 @@ def _quantile_bc(y, p, n, method, xp):
...
+    # refactored to for loop with ~log2(n) iterations for JAX JIT
+    for i in range(int(math.log2(x.shape[-1])) + 1):
+        c = (a + b) // 2
+        x0 = xp.take_along_axis(x, c, axis=-1)
+        j = x0 >= y
+        b = xp.where(j, c, b)
+        a = xp.where(j, a, c)
+
+    b = xp.where(y <= xp.min(x, axis=-1, keepdims=True), 0, b)
+    return b if side == 'left' else  x.shape[-1] - b
```
lucascolley:
```suggestion
    return b if side == 'left' else x.shape[-1] - b
```

lucascolley (approved):
Seems good to me once OIivier's comment is resolved, thanks Matt!

scipy/stats/_quantile.py
mdhaber: Yes, I know,, but i have actually been ignoring that in all my translations lately. It's easier for me to come back and do a whole bunch of corrections like this at once than to keep it in mind every time I create an array. But I can make the adjustment in this PR.

lucascolley: I'm happy either way, perfectly fine to punt on that until it goes into array-api-extra or the standard IMO

mdhaber: OK then, I think I'll wait so we don't need to re-run CI. But I can make the change in gh-23941, since we'll probably need to re-run CI there.

mdhaber: Oh wait we need to re-run CI anyway.

mdhaber:
OK, should be all set here. The failures are (or were, in the case of CircleCI) in `main`.

A useful enhancement would be to see how to change the algorithm `side='right'` rather than negating/reversing the data.

Edge cases I mentioned before are to enforce the same convention for NaNs as NumPy (might be fixed by the idea above) and add support for zero-size input.

I think it would also be useful to put together a simple interpolation function (crossref gh-22118) with this.

@lucascolley if I'm slow to open a PR to `array-api-extra`, just lay down an ultimatum that further review of array API translations and such will depend on that PR. That will motivate me to come back to it : P

lucascolley:
nice to know that you are subject to ultimatums :)

PR #23933, opened 2025-11-05
ENH: spatial.transform: add `axis` argument to `RigidTransform.mean`

scottshambaugh:
#### Reference issue
Builds off of https://github.com/scipy/scipy/pull/23828 and closes out the last item in https://github.com/scipy/scipy/issues/23391.

Adds the `axis` argument to RigidTransform.mean, to match the Rotation implementation.

#### Additional information
@amacati FYI

### DIFF
--- a/scipy/spatial/transform/_rigid_transform_cy.pyx
+++ b/scipy/spatial/transform/_rigid_transform_cy.pyx
@@ -246,11 +246,31 @@ def pow(double[:, :, :] matrix, float n):
 
 @cython.embedsignature(True)
 @cython.boundscheck(False)
-@cython.wraparound(False)
-def mean(double[:, :, :] matrix, weights=None):
+def mean(double[:, :, :] matrix, weights=None, axis=None):
     if matrix.shape[0] == 0:
         raise ValueError("Mean of an empty transform set is undefined.")
-    
+
+    # The Cython path assumes matrix is Nx4x4, so axis has to be None, 0, -1, (0,), (-1,),
+    # or (). The code path is unchanged for any of the options except (), where we
+    # immediately return the matrix
+    if axis == ():
+        return matrix
+
+    if axis is None:
+        axis = (0,)
+    if isinstance(axis, int):
+        axis = (axis,)
+    if not isinstance(axis, tuple):  # Must be tuple by now
+        raise ValueError("`axis` must be None, int, or tuple of ints.")
+    if min(axis) < -1 or max(axis) > 0:
+        raise ValueError(
+            f"axis {axis} is out of bounds for transform with shape "
+            f"{np.asarray(matrix).shape[:-2]}."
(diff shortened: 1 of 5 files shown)


amacati:
Seems like CI is not happy at all, but it's unrelated to this one as far as I can tell. Still, might be good to wait for this to get fixed to make sure we don't overlook anything.

lucascolley:
check out the `Conda & umfpack/scikit-sparse, fast, py3.11/npAny, spin` logs, there seem to be some real failures in there.

It seems like all jobs apart from those which use Pixi are failing, which suggests a new release of a dependency in the past few days is to blame. It will be _so much nicer_ once we have automated reporting on which updates cause failures like this!

amacati:
Is that in the works with pixi?

lucascolley:
> Is that in the works with pixi?

@mdhaber submitted a proposal to the NumFOCUS SDG program which would give us some funding to push that through. I've already started laying groundwork, I commented on my rough plan going forwards at https://github.com/pydata/xarray/pull/10888#issuecomment-3474715493. The relevant discussion to what would help avoid problems like this will likely follow https://github.com/scipy/scipy/issues/23637#issuecomment-3356599227.

amacati:
Nice, sounds like a pretty substantial improvement of the CI pipeline.

lucascolley:
absolutely, fingers crossed for the funding!

scottshambaugh:
Most CI failures cleaned up, the remaining ones look unrelated but should still wait for the builds to complete at least.

amacati:
Can we rerun the CI or have the build issues not been resolved yet?

lucascolley:
> have the build issues not been resolved yet?

we have a new set of issues now 🙃

lucascolley:
down to just a single job failing on main, so should be pretty safe if only that fails here.

scipy/spatial/transform/_rigid_transform_cy.pyx
```diff
@@ -246,11 +246,31 @@ def pow(double[:, :, :] matrix, float n):
 
 @cython.embedsignature(True)
 @cython.boundscheck(False)
-@cython.wraparound(False)
```
lucascolley, line 249: what does the removal of this entail?

lucascolley (approved):
CI is happy!

scipy/spatial/transform/_rigid_transform_cy.pyx
scottshambaugh: It allows us to use the negative indexing in the ValueError below:
`np.asarray(matrix).shape[:-2]`

lucascolley:
@amacati happy to merge?

amacati (approved):
Looks good. Sry for taking so long, wasn't aware the review was still missing.

PR #23934, opened 2025-11-05
DOC: spatial: RigidTransform docs tweaks

### LINKED ISSUE #23879: DOC: RigidTransform.as_components doc is incorrect

In the [docs](https://scipy.github.io/devdocs/reference/generated/scipy.spatial.transform.RigidTransform.as_components.html#scipy.spatial.transform.RigidTransform.as_components), it seems to have added some docs of  `RigidTransform.apply` to `RigidTransform.as_components`.

Also the example (in `as_components`) has an example of `apply` in it

_No response_

_No response_

scottshambaugh:
[docs only]

#### Reference issue
Closes https://github.com/scipy/scipy/issues/23879

Simplifies the wording and examples in `RigidTransform.as_components()`

### DIFF
--- a/scipy/spatial/transform/_rigid_transform.py
+++ b/scipy/spatial/transform/_rigid_transform.py
@@ -1140,7 +1140,7 @@ class RigidTransform:
         returns the rotation corresponding to this rotation matrix
         ``r = Rotation.from_matrix(R)`` and the translation vector ``t``.
 
-        When applying a transform `tf` to a vector `v`, the result is the same
+        When applying a transform ``tf`` to a vector ``v``, the result is the same
         as if the rotation and translation components were applied to the vector
         with the following operation:
         ``tf.apply(v) == translation + rotation.apply(v)``.
--- a/scipy/spatial/transform/_rigid_transform.py
+++ b/scipy/spatial/transform/_rigid_transform.py
@@ -1140,10 +1140,10 @@ class RigidTransform:
         returns the rotation corresponding to this rotation matrix
         ``r = Rotation.from_matrix(R)`` and the translation vector ``t``.
 
-        Take a transform ``tf`` and a vector ``v``. When applying the transform
-        to the vector, the result is the same as if the transform was applied
-        to the vector in the following way:
-        ``tf.apply(v) == translation + rotation.apply(v)``
+        When applying a transform `tf` to a vector `v`, the result is the same
+        as if the rotation and translation components were applied to the vector
+        with the following operation:
+        ``tf.apply(v) == translation + rotation.apply(v)``.
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


scipy/spatial/transform/_rigid_transform.py
```diff
@@ -1140,10 +1140,10 @@ def as_components(self) -> tuple[Array, Rotation]:
...
         ``r = Rotation.from_matrix(R)`` and the translation vector ``t``.
 
-        Take a transform ``tf`` and a vector ``v``. When applying the transform
-        to the vector, the result is the same as if the transform was applied
-        to the vector in the following way:
-        ``tf.apply(v) == translation + rotation.apply(v)``
+        When applying a transform `tf` to a vector `v`, the result is the same
+        as if the rotation and translation components were applied to the vector
+        with the following operation:
+        ``tf.apply(v) == translation + rotation.apply(v)``.
```
haarisr: Personally, the example below seems like it covers this.

Maybe include a `see also` note to point it to the `tf.apply` docs instead of having this doc.

haarisr (commented):
Changes look great!

Thanks for taking the feedback. Just minor comments, feel free to ignore.

scipy/spatial/transform/_rigid_transform.py
scottshambaugh: IMO the sections should stand on their own, so would prefer to keep this information even if a bit redundant. I also think it's important to call this out explicitly, to distinguish from the potential (incorrect) ordering `rotation.apply(translation + v)`.

amacati (approved):
Seems like a minor wording change, but I agree with the consistency of variables and it's more readable. Looks good.

scipy/spatial/transform/_rigid_transform.py
```diff
@@ -1140,10 +1140,10 @@ def as_components(self) -> tuple[Array, Rotation]:
         returns the rotation corresponding to this rotation matrix
         ``r = Rotation.from_matrix(R)`` and the translation vector ``t``.
 
-        Take a transform ``tf`` and a vector ``v``. When applying the transform
-        to the vector, the result is the same as if the transform was applied
-        to the vector in the following way:
-        ``tf.apply(v) == translation + rotation.apply(v)``
+        When applying a transform `tf` to a vector `v`, the result is the same
```
rgommers: minor: you lost a couple of backticks here, code is always typeset with double backticks in ReST. Single backticks are for references that Sphinx turns into links. 

```suggestion
        When applying a transform ``tf`` to a vector ``v``, the result is the same
```

I'll apply that change before merging.

rgommers (approved):
LGTM too, so in it goes. Thanks @scottshambaugh & reviewers!

PR #22894, opened 2025-04-26
ENH: stats.quantile: add `weights` parameter

BrassBull:
#### Reference issue
Closes #22794

#### What does this implement/fix?
Added support for weights parameter for inverted_cdf method

PR #23941, opened 2025-11-06
ENH: stats.quantile: add support for frequency weights

### LINKED ISSUE #22794: ENH: add `weights` param to `stats.quantile` similarly to NumPy

Adding weights support to `scipy.stats.quantile` would allow array API consuming libraries such as scikit-learn not have to implement their own version.

NumPy recently added support for weighted quantile with `method=inverted_cdf`:

- https://github.com/numpy/numpy/pull/24254

This code change could probably be adapted to the scipy code-base while adding array API support along the way.

This could in theory be generalized to `method="averaged_inverted_cdf"` (both in NumPy and SciPy).

For reference, scikit-learn also maintains its own weighted percentile implementation (`_weighted_percentile` is equivalent to `method="inverted_cdf"` and `_averaged_weighted_percentile` is equivalent to `method="averaged_inverted_cdf"`): 

https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/stats.py

The NumPy implementation is closer to SciPy, though.

_No response_

_No response_

mdhaber:
#### Reference issue
Closes gh-22794
Supersedes gh-22894. This PR addresses the concerns expressed in gh-22794, is more general (adds weights support for all `method`s except `'harwell-davis'` rather than just `inverted_cdf`), and include more extensive tests.
https://github.com/data-apis/array-api-extra/pull/494

This adds support for frequency weights to `scipy.stats.quantile`.

Results agree with NumPy for `method='inverted_cdf'`
<details>

```python3
import numpy as np
from scipy import stats
rng = np.random.default_rng(23458924473496)
x = rng.random(100)
weights = rng.random(100)
p = np.linspace(0., 1., 300)
res = stats.quantile(x, p, weights=weights, method='inverted_cdf')
ref = np.quantile(x, p, weights=weights, method='inverted_cdf')
plt.plot(p, res, '-', label='SciPy')
plt.plot(p, res, '--', label='NumPy')
plt.xlabel('p')
plt.ylabel('quantile(p)')
plt.legend()
plt.show()
```

</details>

image

For non-negative integer weights, `stats.quantile(x, p, weights=counts, ...)` is equivalent to `stats.quantile(np.repeat(x, counts), p, ...)`.

<details>

```python3
import numpy as np
from scipy import stats
rng = np.random.default_rng(23458924473496)
x = rng.random(10)
counts = rng.integers(10, size=10)
p = np.linspace(0., 1., 10000)
res = stats.quantile(x, p, weights=counts, method='linear')
ref = stats.quantile(np.repeat(x, counts), p, method='linear')
plt.plot(p, res, '-', label='Using weights')
plt.plot(p, res, '--', label='Using repetitions')
plt.xlabel('p')
plt.ylabel('quantile(p)')
plt.legend()
plt.show()
```

The documentation specifies that weights that do not satisfy the requirements (non-negative integers) are accepted but may not have valid statistical interpretations.

</details>

image

#### Additional information
Support for weights in `scipy.stats` has come up several times before, most notably in gh-6931. As mentioned in https://github.com/scipy/scipy/issues/22794#issuecomment-2784761564, an important discussion point in gh-6931 was the different meanings of "weights" in statistics (e.g., see [here](https://notstatschat.rbind.io/2020/08/04/weights-in-statistics/)), in other statistical software, and possibly within `scipy.stats`. Here are my responses to my own concerns.

I found twelve functions/classes in `scipy.stats` that accept a `weights` argument: `gmean`, `hmean`, `pmean`, `cumfreq`, `relfreq`, `wasserstein_distance_nd`, `wasserstein_distance`, `energy_distance`, `expectile`, `combine_pvalues`, `kde`, `Mixture`. While it's not always documented clearly, it seems that the first nine of these produce valid results when used for frequency weights[^1]. 

<details>

```python3
import numpy as np
from scipy import stats

rng = np.random.default_rng(285494985394583)
nx, ny = 100, 100
x = rng.integers(1, 11, size=nx) / 10.
y = rng.integers(1, 11, size=ny) / 10.
x_vals, x_weights = np.unique_counts(x)
y_vals, y_weights = np.unique_counts(y)

ref = stats.gmean(x)
res = stats.gmean(x_vals, weights=x_weights)
np.testing.assert_allclose(res, ref)

ref = stats.hmean(x)
res = stats.hmean(x_vals, weights=x_weights)
np.testing.assert_allclose(res, ref)

ref = stats.pmean(x, 3)
res = stats.pmean(x_vals, 3, weights=x_weights)
np.testing.assert_allclose(res, ref)

ref = stats.cumfreq(x)
res = stats.cumfreq(x_vals, weights=x_weights)
np.testing.assert_equal(res, ref)

ref = stats.relfreq(x)
res = stats.relfreq(x_vals, weights=x_weights)
# np.testing.assert_equal(res, ref)  # see gh-23931

ref = stats.wasserstein_distance(x, y)
res = stats.wasserstein_distance(x_vals, y_vals, x_weights, y_weights)
np.testing.assert_equal(res, ref)

ref = stats.energy_distance(x, y)
res = stats.energy_distance(x_vals, y_vals, x_weights, y_weights)
np.testing.assert_equal(res, ref)

ref = stats.expectile(x, 0.7)
res = stats.expectile(x_vals, 0.7, weights=x_weights)
np.testing.assert_equal(res, ref)
```

</details>

The `weights` parameter of `combine_pvalues` does _not_ support frequency weights. (I haven't looked into what it *is* doing.)

<details>

 ```python3
ref = stats.combine_pvalues(x[x != 1], method='stouffer')
res = stats.combine_pvalues(x_vals[x_vals != 1], method='stouffer', 
                            weights=x_weights[x_vals != 1])
np.testing.assert_equal(res, ref)  # fails - both statistic and p-value are different
# AssertionError: 
# Items are not equal:
#  ACTUAL: np.float64(0.4734226456871095)
#  DESIRED: np.float64(0.41556093437208813)
```

</details>

I haven't dug into it, but the results of using `weights` and repeating observations affects the smoothing of `gaussian_kde`. I'm not sure whether this is intentional - it might be.

<details>

```
x_ = np.linspace(-1, 2, 300)
ref = stats.gaussian_kde(x)(x_)
res = stats.gaussian_kde(x_vals, weights=x_weights)(x_)
# np.testing.assert_allclose(res, ref)
plt.plot(x_, res, x_, ref)  # affects smoothing
```

image

</details>

`Mixture` only accepts relative weights that sum to one, and they don't really represent weights of _observations_ at all, so I don't think a "frequency weights" interpretation would be valid.

That said, it is pretty common in statistical software (e.g. [SAS, SPSS, Stata](https://stats.oarc.ucla.edu/other/mult-pkg/faq/what-types-of-weights-do-sas-stata-and-spss-support/)) to support different types of weights in different functions, and we are not alone in being imperfect about documenting what kind of weights are implemented[^2]. 

Frequency weights are currently the most widely supported in `scipy.stats`, and they will probably remain the most common in the future because:
- they are conceptually simple,
- it is always _possible_ (although probably not advisable) to add a trivial implementation for counting number-valued weights using `repeats`,
- a more efficient implementation is often easy, and
- correctness is easy to verify with property-based testing.

With all that in mind, I think it's OK to implement frequency weights in `quantile` using the keyword argument `weights`. I don't think it needs to have a more specific name ([Stata](https://www.stata.com/help.cgi?weight) uses `fweights` to distinguish them from `pweights`, `aweights`, and `iweights`.) because we can give the *other* kinds of weights more specific names if/when the need arises. As for `combine_pvalues` and other places where `weights` does NOT refer to frequency weights, we at least need to document how the weights can be interpreted, and we can consider renaming the parameters[^2].

[^1]: In some cases (e.g. sample mean), multiply weight types may lead to the same weighted version of a statistic. I know that these functions support frequency weights, but the results might still be valid for other types of weighting.
[^2]: I'll submit a separate PR to improve this, though.

### CODE REVIEW

File: scipy/stats/_quantile.py [hunk 1]
```diff
@@ -383,3 +432,34 @@ def _quantile_bc(y, p, n, method, xp):
     elif method == '_nearest':
         k = xp.round(ij)
     return xp.take_along_axis(y, xp.astype(k, xp.int64), axis=-1)
+
+
+@xp_capabilities(skip_backends=[("dask.array", "No take_along_axis yet.")])
+def _xp_searchsorted(x, y, *, side='left', xp=None):
```
mdhaber, line 484: This is being reviewed separately in gh-23930.

scipy/stats/_quantile.py
```diff
@@ -325,13 +361,26 @@ def quantile(x, p, *, method='linear', axis=0, nan_policy='propagate', keepdims=
...
+
+    if weights is None:
+        jg = p * n + m
+        jp1 = jg // 1
+        j = jp1 - 1
+    else:
+        cumulative_weights = xp.cumulative_sum(weights, axis=-1)
+        n_int = xp.asarray(n, dtype=xp.int64)
+        n_int = xp.broadcast_to(n_int, cumulative_weights.shape[:-1] + (1,))
+        total_weight = xp.take_along_axis(cumulative_weights, n_int-1, axis=-1)
```
mdhaber, line 415: For `nan_policy='omit'` and `marray`, `n` is an array the is the number of valid entries in a slice; that's why we can't just take the last element. (`marray` tests are skipped right now because I'm running into a mutability issue. I think it's an `marray` issue and the implementation will probably "just work" when that gets resolved because we're already accounting for `n` depending on the slice.)

scipy/stats/tests/test_quantile.py
```diff
@@ -89,23 +96,53 @@ def test_input_validation(self, xp):
         with pytest.raises(ValueError, match=message):
             stats.quantile(x, xp.asarray([0.5, 0.6]), keepdims=False)
 
+    def _get_weights_x_rep(self, x, axis, rng):
```
mdhaber, line 127: The purpose of this function is to generate integral weights and use them to produce an `x_rep` array that is `x` with elements repeated `weights` times. The somewhat tricky thing is to ensure that the weights have the same sum along `axis` so we don't have a ragged `x_rep` array. I thought about using `random_table`... maybe that or another function would help here? It felt very clumsy, but I didn't really try to optimize this.

scipy/stats/_quantile.py
```diff
@@ -233,6 +263,11 @@ def quantile(x, p, *, method='linear', axis=0, nan_policy='propagate', keepdims=
     3. ``closest_observation``: ``m = -1/2`` and
        ``g = 1 - int((index == j) & (j%2 == 1))``
 
+    Note that for methods ``inverted_cdf`` and ``averaged_inverted_cdf``, only the
+    relative proportions of tied observations (and relative weights) affect the
+    results; for all other methods, the total number of observations (and absolute
+    weights) matter.
```
mdhaber, line 287: I was thinking that if I were to include `weights` in an example, this is what I'd show.

I think I'll be asked to include the information  about coverage conditions, etc., that appears in the NumPy documentation, so I'll address that now: I would prefer not to, but I can add references. I don't think the math underlying the weights needs to be spelled out because it's exactly the same math as repeated samples; it's only the implementation that is generalized. Now, that same generalized implementation happens to work for non-integer weights, too, so if there's a reference for valid statistical interpretations of non-integral weights, I'd be interested in including that.

mdhaber (commented):
This relies on `_xp_searchsorted` (gh-23930), so that should merge first. I'd also much prefer to get gh-22644 merged first and resolve the merge conflicts here rather than the other way around. There is also a third PR (yet another ECDF function) that I will post shortly; I'd also like to wait until that's in to merge this.

scipy/stats/tests/test_quantile.py
```diff
@@ -125,7 +125,8 @@ def _get_weights_x_rep(self, x, axis, rng):
     def test_against_numpy(self, method, shape_x, shape_p, axis, weights, xp):
         if weights and method.startswith('_'):
             pytest.skip('`weights=True` not supported by private (legacy) methods.')
-
+        if weights and is_numpy(xp) and xp.__version__ < "2.0":
```
mdhaber: Looks like all failures are either this
```suggestion
        if is_numpy(xp) and xp.__version__ < "2.0":
```
or are (/were) in main.

ogrisel:
> I haven't dug into it, but the results of using weights and repeating observations affects the smoothing of gaussian_kde. I'm not sure whether this is intentional - it might be.

I think this is what caused this PR in scikit-learn to get stalled: https://github.com/scikit-learn/scikit-learn/pull/27971/files#r1745872881 and https://github.com/scikit-learn/scikit-learn/pull/27971/files#r1763478046 (I was worried about not being able to implement weight/repetition equivalence in our scikit-learn estimator).

mdhaber:
@ogrisel thanks - replied there. Let's get that addressed.

ogrisel:
I built scipy with this PR's branch and patched scikit-learn to use this instead of our embedded `sklearn.utils.stats._weighted_percentile` function:

<details>

```python
diff --git a/sklearn/utils/stats.py b/sklearn/utils/stats.py
index 453b0ab122..c35a2b4523 100644
--- a/sklearn/utils/stats.py
+++ b/sklearn/utils/stats.py
@@ -90,7 +90,27 @@ def _weighted_percentile(
         If `array` is 2D and `percentile_rank` is 1D, returns a 2D array
             of shape `(array.shape[1], percentile_rank.shape[0])`
     """
+
+    from scipy.stats import quantile
     xp, _, device = get_namespace_and_device(array)
+
+    array = xp.asarray(array, device=device)
+    if not xp.isdtype(array.dtype, "real floating"):
+        array = xp.asarray(array, dtype=xp.asarray([0.0]).dtype, device=device)
+
+    p = xp.asarray(percentile_rank, dtype=array.dtype, device=device) / 100
+    sample_weight = xp.asarray(sample_weight, dtype=array.dtype, device=device)
+
+    results = quantile(
+        array.T if array.ndim == 2 else array,
+        p=p,
+        axis=-1 if array.ndim == 2 else None,
+        nan_policy="omit",
+        weights=sample_weight.T if sample_weight.ndim == 2 else sample_weight,
+        method="averaged_inverted_cdf" if average else "inverted_cdf",
+    )
+    return results
+
     # `sample_weight` should follow `array` for dtypes
     floating_dtype = _find_matching_floating_dtype(array, xp=xp)
     array = xp.asarray(array, dtype=floating_dtype, device=device)
```

</details>

Then I ran the full scikit-learn tests suite. Once https://github.com/scipy/scipy/pull/23930#discussion_r2503515900 is fixed, the remaining errors seems to be related to the fact that 0 weight data points are not ignored (as they should according to frequency weight semantics).

In particular, the following scikit-learn test fails:

https://github.com/scikit-learn/scikit-learn/blob/c8a5f807a365a069b4d96da1a84318ce6253eab7/sklearn/utils/tests/test_stats.py#L112-L132

There are two other tests failures, but I suspect they stem from the same cause.

EDIT: to make this more directly related to scipy, here is a minimal reproducer. The frequency semantics seem to be violated for the `"(averaged_)inverted_cdf"` methods:

```python
>>> from scipy.stats import quantile
>>> import numpy as np
>>> data = np.array([0, 1, 2, 3, 4, 5, 6])
>>> weights = np.array([0, 0, 1, 1, 0, 1, 0])
>>> quantile(data, p=[0.0, 0.5, 1.0], weights=weights, method="averaged_inverted_cdf")
array([1. , 3. , 5.5])
>>> quantile(data[weights != 0], p=[0.0, 0.5, 1.0], method="averaged_inverted_cdf")
array([2., 3., 5.])
>>> quantile(data, p=[0.0, 0.5, 1.0], weights=weights, method="inverted_cdf")
array([0., 3., 5.])
>>> quantile(data[weights != 0], p=[0.0, 0.5, 1.0], method="inverted_cdf")
array([2., 3., 5.])
```

The `"linear"` method seems to not be impacted by this bug:

```python
>>> quantile(data, p=[0.0, 0.5, 1.0], weights=weights, method="linear")
array([2., 3., 5.])
```

mdhaber:
Hmm there is a test with zero weights that compares against NumPy; surprised it didn't find this. I'll take a look. Thanks for the MRE.

mdhaber:
@ogrisel I thought zero-weights were being tested, but apparently the RNG gods spared me the ~~frustration of debugging~~ opportunity to fix that earlier ; ) The simple fix is to treat them like omitted NaNs - move them to the end of the sorted array and don't include them in the count. I didn't include your test case specifically, but updated an existing test. There's an argument for adding a special test for this, though, since it is not testing all the methods.
I noticed but didn't handle the case of all-zero weights yet. I think that will interpolate between the first two elements along the slice instead of returning NaN like it should. A full slice of NaNs with `nan_policy='omit'` works correctly, so I can generalize that code to fix zero weights, too. However, I'd like to address some other cleanups regarding nan handling at the same time, so if it doesn't break sklearn tests, I'd like to save that for a follow-up rather than complicating this PR.

scipy/stats/tests/test_quantile.py
```diff
@@ -210,3 +253,20 @@ def test_transition(self, method, xp):
...
+        method = 'inverted_cdf'
+        x = rng.random(size=100)
+        weights = rng.random(size=100)
+        if zero_weights:
+            weights[weights < 0.5] = 0
+        p = np.linspace(0., 1., 300)
+        res = stats.quantile(xp.asarray(x, dtype=dtype), xp.asarray(p, dtype=dtype),
+                             method=method, weights=xp.asarray(weights, dtype=dtype))
+        ref = np.quantile(x, p, method=method, weights=weights)
+        xp_assert_close(res, xp.asarray(ref, dtype=dtype))
```
mdhaber, line 313: This also passes, and I can commit if desired.

```suggestion
        xp_assert_close(res, xp.asarray(ref, dtype=dtype))
        
    @pytest.mark.parametrize('method',
         ['inverted_cdf', 'averaged_inverted_cdf', 'closest_observation',
          'hazen', 'interpolated_inverted_cdf', 'linear',
          'median_unbiased', 'normal_unbiased', 'weibull'])
    def test_zero_weights(self, method, xp):
        rng = np.random.default_rng(85468924398205602)
        
        # test 1-D versus eliminating zero-weighted values
        n = 100
        x = xp.asarray(rng.random(size=n))
        p = xp.asarray(rng.random(size=n))
        i_zero = xp.asarray(rng.random(size=n) < 0.1)
        weights = xp.asarray(rng.random(size=n))
        weights = xp.where(i_zero, 0., weights)
        res = stats.quantile(x, p, weights=weights, method=method)
        ref = stats.quantile(x[~i_zero], p, weights=weights[~i_zero], method=method)
        xp_assert_close(res, ref)

        # test multi-D versus `nan_policy='omit'`
        shape = (5, 100)
        x = xp.asarray(rng.random(size=shape))
        p = xp.asarray(rng.random(size=shape))
        i_zero = xp.asarray(rng.random(size=shape) < 0.1)
        weights = xp.asarray(rng.random(size=shape))
        x_nanned = xp.where(i_zero, xp.nan, x)
        weights_zeroed = xp.where(i_zero, 0., weights)
        res = stats.quantile(x, p, weights=weights_zeroed, method=method, axis=-1)
        ref = stats.quantile(x_nanned, p, weights=weights,
                             nan_policy='omit', method=method, axis=-1)
        xp_assert_close(res, ref)

```

scipy/stats/_quantile.py
```diff
@@ -61,8 +71,24 @@ def _quantile_iv(x, p, method, axis, nan_policy, keepdims):
...
+        i_y = xp.argsort(x, axis=axis, stable=False)
+        y = xp.take_along_axis(x, i_y, axis=axis)
+        y, p, weights, i_y, n_zero_weight = _broadcast_arrays(
+            (y, p, weights, i_y, n_zero_weight), axis=axis)
+        weights = xp.take_along_axis(weights, i_y, axis=axis)
+        # When NaNs have zero weight, they shouldn't propagate
+        # data-apis/array-api-extra#506 raises an error here, so use `where`
+        # y = xpx.at(y)[xp.isnan(y) & (weights == 0)].set(1.)
+        if nan_policy == 'propagate':
+            y = xp.where(~xp.isfinite(y) & (weights == 0), 1., y)
```
mdhaber: This might not be needed anymore.

scipy/stats/_quantile.py
```diff
@@ -61,8 +71,24 @@ def _quantile_iv(x, p, method, axis, nan_policy, keepdims):
...
 
-    y = xp.sort(x, axis=axis, stable=False)
-    y, p = _broadcast_arrays((y, p), axis=axis)
+    if weights is None:
+        y = xp.sort(x, axis=axis, stable=False)
+        y, p = _broadcast_arrays((y, p), axis=axis)
+        n_zero_weight = None
+    else:
+        i_zero_weight = (weights == 0)
+        n_zero_weight = xp.count_nonzero(i_zero_weight, axis=axis, keepdims=True)
```
mdhaber, line 84: Hmm we may need to broadcast first depending on what `xpx.at` does when the arrays are of different (but broadcastable) shapes (next line)

mdhaber:
Added more extensive tests of zero-weights. Skipped CI for now - when gh-23930 merges, we can merge main and run again, and hopefully then the sklearn tests will pass, too.

ogrisel:
> then the sklearn tests will pass, too.

@mdhaber I just tried rerunning the scikit-learn tests again with this updated branch, and I found another problem with the `quantile` function corrupting the input array when it holds floating point values (inplace modification):


```python
import numpy as np
from scipy.stats import quantile

a = np.arange(10).astype(np.float64)
w = np.ones(a.shape[0]).astype(np.float64)
w[:3] = 0
quantile(a, 0.5, axis=0, weights=w)
print(a)
```
```
[inf inf inf  3.  4.  5.  6.  7.  8.  9.]
```

mdhaber:
Good catch! I'll fix it today.

mdhaber:
@ogrisel this was new in the recent for zero-weights. Fixed/tested in [cb2ba66](https://github.com/scipy/scipy/pull/23941/commits/cb2ba660a06ac0d8f92a9ad0e3f33bda3eba08c5).

ogrisel:
I confirm that the scikit-learn tests now all pass against this implementation of weighted quantiles 🎉 (except one related to a known bug on the scikit-learn side).

ogrisel:
For the record, I conducted a quick benchmark against the implementation of `averaged_inverted_cdf` we have in scikit-learn and:

- on numpy/CPU the speed is the same;
- on torch/MPS (Apple GPU), the scipy version is a bit faster.

Details: https://gist.github.com/ogrisel/8c645970dcd9478679ad1a7669f25c18

lucascolley:
cc @cakedev0

dschmitz89:
Is this ready to merge from your side @mdhaber ?

mdhaber:
Yup, I think so, once it has gone through a normal review from a SciPy maintainer.

dschmitz89 (changes requested):
Two nitpicks:

* I do not see a validation check that weights are nonnegative
* Could we add a basic sanity check that for equal nonzero weights we get the same results as the unweighted calculation?

mdhaber:
> I do not see a validation check that weights are nonnegative

That was actually intentional. Positive, non-integral values are accepted, too. I documented that: "Values other than finite [non-negative integers] are accepted, but may not have valid statistical interpretations." Part of the reason is to avoid value-dependent input validation that will not be possible on lazy backends. 

I can add the sanity check - although I think that's a very special case of the check of integral weights that is already done, which will already have some `1` weights and checks against an unweighed reference (values are just repeated). Do you still want me to do that?

scipy/stats/_quantile.py
```diff
@@ -14,16 +14,20 @@
...
     if not xp.isdtype(xp.asarray(p).dtype, 'real floating'):
         raise ValueError("`p` must have real floating dtype.")
 
-    x, p = xp_promote(x, p, force_floating=True, xp=xp)
+    if not (weights is None
+            or xp.isdtype(xp.asarray(weights).dtype, ('integral', 'real floating'))):
+        raise ValueError("`weights` must have real dtype.")
+
+    x, p, weights = xp_promote(x, p, weights, force_floating=True, xp=xp)
     p = xp.asarray(p, device=xp_device(x))
```
j-bowhay: Now we call `xp_promote` which in turn calls `xp.asarray` is this line still needed?

dschmitz89 (approved):

scipy/stats/_quantile.py
```diff
@@ -14,17 +14,20 @@
...
     if not xp.isdtype(xp.asarray(p).dtype, 'real floating'):
         raise ValueError("`p` must have real floating dtype.")
 
-    x, p = xp_promote(x, p, force_floating=True, xp=xp)
-    p = xp.asarray(p, device=xp_device(x))
+    if not (weights is None
+            or xp.isdtype(xp.asarray(weights).dtype, ('integral', 'real floating'))):
+        raise ValueError("`weights` must have real dtype.")
+
+    x, p, weights = xp_promote(x, p, weights, force_floating=True, xp=xp)
```
mdhaber, line 30: @j-bowhay I think this is needed because `p` can be a Python scalar, so the device of `x` and `weights` is preserved, but `p` becomes an array on the default device. We could make `xp_promote` smarter in a separate PR.
```suggestion
    x, p, weights = xp_promote(x, p, weights, force_floating=True, xp=xp)
    p = xp.asarray(p, device=xp_device(x))
```

j-bowhay (approved):

PR #23942, opened 2025-11-06
DOC correct "shape" type for "scipy.sparse.random_array"

### LINKED ISSUE #23925: DOC: Invalid documentation of parameter "shape" of the "scipy.sparse. random_array" function

The documentation of [`scipy.sparse.random()`](https://scipy.github.io/devdocs/reference/generated/scipy.sparse.random_array.html) states that the first parameter `shape` can either be an integer or a tuple of integers. However, setting an integer raises an error:

```python
>>> from scipy.sparse import random_array
>>> random_array(100)
TypeError: 'int' object is not iterable
```

because of this line :

https://github.com/scipy/scipy/blob/d419874eb4608dd6a91b02da0517dcf0cba7cf7e/scipy/sparse/_construct.py#L1507

The first argument of [`math.prod()`](https://docs.python.org/3/library/math.html#math.prod) must be an iterable, which is not the case for the [`numpy.prod()`](https://numpy.org/doc/stable/reference/generated/numpy.prod.html) function. I would assume that the numpy function was previously used and then replaced with the math function to handle the case mentioned in the comment.

There are two possible solutions :

1. Changing the documentation and removing `int` as a possible type for the `shape` parameter.
2. Changing the code to handle the case of `shape` being an integer.

_No response_

AdrianRasoOnGit:
#### Reference issue
Closes #23925

This PR changes documentation for the `shape` parameter of `scipy.sparse.random_array`. The previous docstring stated that it could be an `int` or a tuple of ints, when in reality the implementation requires a 2-tuple `(m, n)` and does not really accept a single integer. The docstring now specifies that `shape`must be a tuple of two integers.

### DIFF
--- a/scipy/sparse/_construct.py
+++ b/scipy/sparse/_construct.py
@@ -1405,7 +1405,7 @@ def random_array(shape, *, density=0.01, format='coo', dtype=None,
     Parameters
     ----------
     shape : tuple of int
-        shape of the array as an n-tuple of integers `(m, ..., n)`.
+        shape of the array.
     density : real, optional (default: 0.01)
         density of the generated matrix: density equal to one means a full
         matrix, density of 0 means a matrix with no non-zero items.
--- a/scipy/sparse/_construct.py
+++ b/scipy/sparse/_construct.py
@@ -1404,8 +1404,8 @@ def random_array(shape, *, density=0.01, format='coo', dtype=None,
 
     Parameters
     ----------
-    shape : tuple of two ints
-        shape of the array as a 2-tuple `(m, n)`.
+    shape : tuple of int
+        shape of the array as an n-tuple of integers `(m, ..., n)`.
     density : real, optional (default: 0.01)
         density of the generated matrix: density equal to one means a full
         matrix, density of 0 means a matrix with no non-zero items.
--- a/scipy/sparse/_construct.py
+++ b/scipy/sparse/_construct.py
@@ -1404,8 +1404,8 @@ def random_array(shape, *, density=0.01, format='coo', dtype=None,
 
     Parameters
     ----------
-    shape : int or tuple of ints
-        shape of the array
+    shape : tuple of two ints
+        shape of the array as a 2-tuple `(m, n)`.
     density : real, optional (default: 0.01)
         density of the generated matrix: density equal to one means a full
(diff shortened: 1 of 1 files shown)


WarrenWeckesser:
Thanks @AdrianRasoOnGit, but this isn't quite right.  The function accepts tuples of different lengths, e.g.

```
In [14]: from scipy.sparse import random_array

In [15]: random_array((100,), density=.05)
Out[15]: 
<COOrdinate sparse array of dtype 'float64'
	with 5 stored elements and shape (100,)>

In [16]: random_array((100, 10), density=.05)
Out[16]: 
<COOrdinate sparse array of dtype 'float64'
	with 50 stored elements and shape (100, 10)>

In [17]: random_array((100, 10, 25), density=.05)
Out[17]: 
<COOrdinate sparse array of dtype 'float64'
	with 1250 stored elements and shape (100, 10, 25)>
```

AdrianRasoOnGit:
I noticed the error, sorry. I was fixing it at the moment, I will push now the correction

scipy/sparse/_construct.py
```diff
@@ -1404,8 +1404,8 @@ def random_array(shape, *, density=0.01, format='coo', dtype=None,
 
     Parameters
     ----------
-    shape : int or tuple of ints
-        shape of the array
+    shape : tuple of int
+        shape of the array as an n-tuple of integers `(m, ..., n)`.
```
WarrenWeckesser: `n`-tuple suggests `n` is the length of the tuple, so you shouldn't also use `n` *in* the tuple.  I think it would be fine to leave this line as it was,  or perhaps add `sparse`:  `shape of the sparse array`.

WarrenWeckesser:
@AdrianRasoOnGit, when you push another commit, include the notation '[docs only]' in the commit message.  Since this affects only the docstring of an object, we don't need to run the full CI test suite.

scipy/sparse/_construct.py
AdrianRasoOnGit: Yes, that example is a bit unfortunate, sorry. I'll revert to the previous wording then

AdrianRasoOnGit:
Thanks for the help. I added the notation in this commit, and reverted the line with the example to the original line

WarrenWeckesser (approved):
Looks good, thanks.

Issue #23896, opened 2025-10-30
BUG: batched `linalg.orth` and `linalg.null_space` require ragged arrays


ev-br:
### Describe your issue.

For a single array, the shape of the return values of `linalg.orth` (the orthonormal basic of the range of the input) and `linalg.null_space` (the orthonormal basis of the null space of the input) depends on the rank of the input matrix. Thus both fail for batched inputs where slices have different rank.

### Reproducing Code Example

```python
In [16]: a = np.stack([np.diag([1, 1e-20]), np.diag([1, 1])])

In [17]: linalg.null_space(a)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[17], line 1
----> 1 linalg.null_space(a)

File ~/repos/scipy/scipy/build-install/usr/lib/python3.12/site-packages/scipy/_lib/_util.py:1186, in _apply_over_batch.<locals>.decorator.<locals>.wrapper(*args, **kwargs)
   1184 # Reshape results
   1185 for i, result in enumerate(results):
-> 1186     result = xp.stack(result)
   1187     core_shape = result.shape[1:]
   1188     results[i] = xp.reshape(result, batch_shape + core_shape)

File ~/miniforge3/envs/scipy-dev/lib/python3.12/site-packages/numpy/_core/shape_base.py:460, in stack(arrays, axis, out, dtype, casting)
    458 shapes = {arr.shape for arr in arrays}
    459 if len(shapes) != 1:
--> 460     raise ValueError('all input arrays must have the same shape')
    462 result_ndim = arrays[0].ndim + 1
    463 axis = normalize_axis_index(axis, result_ndim)

ValueError: all input arrays must have the same shape
```

### Error message

```shell
.
```

### SciPy/NumPy/Python version and system information

```shell
scipy >= 1.16
```

ev-br:
Since it's really not fixable, the only options I can think of are to either drop the `@apply_over_batch` decorator from these two functions (it's a backwards compatible break, yes), or shrug it off and keep ignoring their existence (why on earth do we even have these SVD wrappers).

mdhaber:
Interesting. I think it would be acceptable to just document that all elements of a batch need to have the same rank, otherwise the function will error out.

We could also customize the error for these functions. Other options include returning an object array of arrays or a list (of lists) of arrays, although I don't think we should do either.

ilayn:
[...]

PR #23945, opened 2025-11-06
DOC clarify batching behavior in linalg.orth and linalg.null_space [d…

AdrianRasoOnGit:
#### Reference issue
#23896

This PR adds a `Notes` section to both `scipy.linalg.orth` and `scipy.linalg.null_space`, explaining the behavior of batched inputs when the slices have different ranks.

### DIFF
--- a/scipy/linalg/_decomp_svd.py
+++ b/scipy/linalg/_decomp_svd.py
@@ -329,6 +329,15 @@ def orth(A, rcond=None):
         Orthonormal basis for the range of A.
         K = effective rank of A, as determined by rcond
 
+    Notes
+    -----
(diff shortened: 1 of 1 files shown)


j-bowhay:
It's not clear from the original what the consensus was, that should be decided first before merging this

AdrianRasoOnGit:
Apologies, this was a wrong call. I'm closing it then

j-bowhay:
Sorry I didn't mean to imply that we should close this as it might be the approach we decide, more than we should check what the consensus was before merging

AdrianRasoOnGit:
Oh, thanks for the clarification, and sorry for the misunderstanding

PR #23949, opened 2025-11-07
ENH: stats.estimated_cdf: array API function for `plotting_positions`/`percentileofscore`/`ecdf`/`cumfreq`

mdhaber:
#### Reference issue
Toward gh-20544
Toward gh-22194
Addresses the enhancement request contained in gh-21563 (closed by documenting the limitation).

This introduces a function that reproduces much of the functionality of `stats.mstats.plotting_positions`, `stats.percentileofscore`, `stats.ecdf.cdf`, and `stats.cumfreq`. But It is also vectorized like `quantile`, and it currently works with all the backends we test (including `marray`) except for `dask.array` (which is missing the standard `take_along_axis`). It supports the usual options that we expect in `stats` functions: `axis`, `keepdims`, and `nan_policy` (without resorting to loops or masked arrays).

The working name has been `iquantile`, but it's more appropriate to think of the quantile function as the inverse of the CDF, and not the other way around. I think `ogive` would a descriptive name, since a piecewise linear ECDF is referred to as an [ogive](https://en.wikipedia.org/wiki/Ogive_(statistics)).

**Examples**
The ECDF is `ogive` with `method='inverted_cdf'`.
```python3
import numpy as np
from scipy import stats
rng = np.random.default_rng(3249825983458343)
x = rng.standard_normal(1000)
y = rng.standard_normal(100)

method = 'inverted_cdf'
ref = stats.ecdf(x).cdf.evaluate(y)
res = stats.ogive(x, y, method=method)
np.testing.assert_allclose(res, ref)
```

Plotting positions are `ogive` with the data passed as both first and second arguments. (`plotting_positions` allows the user to pass `alpha` and `beta` parameters instead of method names, but we can add that to `quantile` and this function easily.)

```python3
method = 'weibull'
ref = stats.mstats.plotting_positions(x, 0, 0)  # equivalent to 'weibull' 
res = stats.ogive(x, x, method=method)
np.testing.assert_allclose(res, ref)
```

Given gh-23948, I think this function makes it a lot easier to plot a cumulative histogram (smooth or step-wise).
```python3
x_ = np.linspace(-4, 4, 300)
p = stats.ogive(x, x_)
plt.plot(x_, p)
plt.show()
```
image
(Admittedly, `stats.ecdf(x).cdf.plot()` is even easier, but that is discontinuous.)

With plenty of caveats, this function and quantile are inverses (to the extent that the functions are invertible).

```python3
ref = rng.random(100)
q = stats.quantile(x, ref, method='hazen')
res = stats.ogive(x, q, method='hazen')
np.testing.assert_allclose(res, ref)
```

It's very easy to get the edge cases wrong, but I think I've been pretty careful.

#### Additional information
~~The function name need to be changed, the documentation needs to be adapted, and~~ The tests need to be trimmed down. ~~This function relies on `_xp_searchsorted`, so gh-23930 should merge first.~~ Support for weights can be added in a separate PR.

### DIFF
--- a/scipy/stats/_quantile.py
+++ b/scipy/stats/_quantile.py
@@ -552,14 +552,14 @@ def _xp_searchsorted(x, y, *, side='left', xp=None):
 def estimated_cdf(x, y, *, method='linear',
                   axis=0, nan_policy='propagate', keepdims=None):
     """
-    Estimate the cumulative distribution function of the data along the specified axis.
+    Estimate the CDF for the distribution underlying a sample.
 
     Parameters
     ----------
     x : array_like of real numbers
         Data array.
     y : array_like of real numbers
-        Datum or data for which to estimate the cumulative probabilities.
+        Points at which to evaluate the estimated cdf.
         See `axis`, `keepdims`, and the examples for broadcasting behavior.
     method : str, default: 'linear'
         The method to use for estimating the cumulative distribution function.
@@ -577,7 +577,7 @@ def estimated_cdf(x, y, *, method='linear',
 
         See Notes for details.
     axis : int or None, default: 0
-        Axis along which the calculation is performed.
+        Axis along which samples in `x` are given in ND case.
         ``None`` ravels both `x` and `y` before performing the calculation,
         without checking whether the original shapes were compatible.
         As in other `scipy.stats` functions, a positive integer `axis` is resolved
--- a/scipy/stats/_quantile.py
+++ b/scipy/stats/_quantile.py
@@ -7,6 +7,7 @@ from scipy._lib._array_api import (
     array_namespace,
     xp_promote,
     xp_device,
+    xp_size,
     _count_nonmasked,
     is_torch,
     is_lazy_array,
@@ -15,14 +16,31 @@ import scipy._external.array_api_extra as xpx
 from scipy.stats._axis_nan_policy import _broadcast_arrays, _contains_nan
 
 
-def _quantile_iv(x, p, method, axis, nan_policy, keepdims, weights):
+def _quantile_iv(x, p, method, axis, nan_policy, keepdims, weights, fun='quantile'):
     xp = array_namespace(x, p, weights)
+
+    if fun == "quantile":
+        methods = {'inverted_cdf', 'averaged_inverted_cdf', 'closest_observation',
+                   'hazen', 'interpolated_inverted_cdf', 'linear',
+                   'median_unbiased', 'normal_unbiased', 'weibull',
+                   'harrell-davis', '_lower', '_midpoint', '_higher', '_nearest',
+                   'round_outward', 'round_inward', 'round_nearest'}
+        allowed_types = 'real floating'
+        def mask_fun(p): return (p > 1) | (p < 0) | xp.isnan(p)
+        var2_name = 'p'
+        var2_type_msg = '`p` must have real floating dtype.'
+    else:
+        methods = set(_estimated_cdf_methods)
+        allowed_types = ('integral', 'real floating')
+        mask_fun = xp.isnan
+        var2_name = 'y'
+        var2_type_msg = '`y` must have real dtype.'
 
     if not xp.isdtype(xp.asarray(x).dtype, ('integral', 'real floating')):
         raise ValueError("`x` must have real dtype.")
 
-    if not xp.isdtype(xp.asarray(p).dtype, 'real floating'):
-        raise ValueError("`p` must have real floating dtype.")
+    if not xp.isdtype(xp.asarray(p).dtype, allowed_types):
+        raise ValueError(var2_type_msg)
 
     if not (weights is None
             or xp.isdtype(xp.asarray(weights).dtype, ('integral', 'real floating'))):
@@ -46,11 +64,6 @@ def _quantile_iv(x, p, method, axis, nan_policy, keepdims, weights):
         raise ValueError(message)
     axis = int(axis)
 
-    methods = {'inverted_cdf', 'averaged_inverted_cdf', 'closest_observation',
-               'hazen', 'interpolated_inverted_cdf', 'linear',
-               'median_unbiased', 'normal_unbiased', 'weibull',
-               'harrell-davis', '_lower', '_midpoint', '_higher', '_nearest',
-               'round_outward', 'round_inward', 'round_nearest'}
     if method not in methods:
         message = f"`method` must be one of {methods}"
         raise ValueError(message)
@@ -70,7 +83,7 @@ def _quantile_iv(x, p, method, axis, nan_policy, keepdims, weights):
     # If data has length zero along `axis`, the result will be an array of NaNs just
     # as if the data had length 1 along axis and were filled with NaNs. This is treated
     # naturally below whether `nan_policy` is `'propagate'` or `'omit'`.
-    if x.shape[axis] == 0:
+    if x.shape[axis] == 0 and fun == 'quantile':
         shape = list(x.shape)
         shape[axis] = 1
         x = xp.full(shape, xp.nan, dtype=dtype, device=xp_device(x))
@@ -91,7 +104,8 @@ def _quantile_iv(x, p, method, axis, nan_policy, keepdims, weights):
             (y, p, weights, i_y, n_zero_weight), axis=axis)
 
     if (keepdims is False) and (p.shape[axis] != 1):
-        message = "`keepdims` may be False only if the length of `p` along `axis` is 1."
+        message = ("`keepdims` may be False only if the length of "
+                   f"`{var2_name}` along `axis` is 1.")
         raise ValueError(message)
     keepdims = (p.shape[axis] != 1) if keepdims is None else keepdims
 
@@ -104,6 +118,7 @@ def _quantile_iv(x, p, method, axis, nan_policy, keepdims, weights):
     n = _count_nonmasked(y, -1, xp=xp, keepdims=True)
     n = n if n_zero_weight is None else n - n_zero_weight
 
+    nan_out = None
     if is_lazy_array(y) or contains_nans:
         nans = xp.isnan(y)
 
@@ -127,14 +142,20 @@ def _quantile_iv(x, p, method, axis, nan_policy, keepdims, weights):
 
     n = xp.asarray(n, dtype=dtype, device=xp_device(y))
 
-    p_mask = (p > 1) | (p < 0) | xp.isnan(p)
-    p = xp.where(p_mask, 0.5, p)
+    # apparently xpx.at is accepting nan_out as a mask even though it doesn't have the
+    # same number of dimensions as `y`, yet it still appears to work correctly?
+    # should refactor for clarity, and p_mask that gets returned here should probably
+    # get a different name - `nan_out` would be more appropriate!
+    p_mask = mask_fun(p) if nan_out is None else mask_fun(p) | nan_out
+    if is_lazy_array(p_mask) or xp.any(p_mask):
+        p = xp.where(p_mask, 0.5, p)  # these get NaN-ed out at the end
 
     return (y, p, method, axis, nan_policy, keepdims,
             n, axis_none, ndim, p_mask, weights, xp)
 
 
-@xp_capabilities(skip_backends=[("dask.array", "No take_along_axis yet.")])
+@xp_capabilities(skip_backends=[("dask.array", "No take_along_axis yet.")],
+                 marray=True)
 def quantile(x, p, *, method='linear', axis=0, nan_policy='propagate', keepdims=None,
              weights=None):
     """
@@ -227,6 +248,7 @@ def quantile(x, p, *, method='linear', axis=0, nan_policy='propagate', keepdims=
     See Also
     --------
     numpy.quantile
+    estimated_cdf
     :ref:`outliers`
 
     Notes
@@ -381,6 +403,10 @@ def quantile(x, p, *, method='linear', axis=0, nan_policy='propagate', keepdims=
     else:  # method.startswith('round'):
         res = _quantile_winsor(y, p, n, method, xp)
 
+    return _post_quantile(res, p_mask, axis, axis_none, ndim, keepdims, xp)
+
+
+def _post_quantile(res, p_mask, axis, axis_none, ndim, keepdims, xp):
     res = xpx.at(res, p_mask).set(xp.nan)
 
     # Reshape per axis/keepdims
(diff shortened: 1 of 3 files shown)


scipy/stats/tests/test_quantile.py
```diff
@@ -410,3 +415,342 @@ def test_nd(self, side, ties, shape, nans_x, infs_x, xp):
...
+        message = "If specified, `keepdims` must be True or False."
+        with pytest.raises(ValueError, match=message):
+            stats.ogive(x, y, keepdims=42)
+
+        message = "`keepdims` may be False only if the length of `y` along `axis` is 1."
+        with pytest.raises(ValueError, match=message):
+            stats.ogive(x, xp.asarray([0.5, 0.6]), keepdims=False)
+
+    @pytest.mark.parametrize('method', _ogive_methods_list)
+    @pytest.mark.parametrize('dtype', [None, 'float32', 'float64'])
```
mdhaber: The simplest way to speed up the tests without much loss of coverage is to remove `dtype` parameterization here. I think it's sufficient to `parametrize` over `dtype` in `test_against_reference` only.

scipy/stats/_quantile.py
```diff
@@ -520,3 +544,251 @@ def _xp_searchsorted(x, y, *, side='left', xp=None):
...
+
+    >>> x = np.stack((np.arange(0, 11), np.arange(10, 21)))
+    >>> stats.ogive(x, [[2.5], [17.5]], axis=-1, keepdims=True)
+    array([[0.25],
+           [0.75]])
+
+    Calculate the empirical distribution function at many points for each of two
+    samples.
+
+    >>> rng = np.random.default_rng(6110515095)
```
mdhaber, line 718: To address CI failure:
```suggestion
    >>> import matplotlib.pyplot as plt
    >>> rng = np.random.default_rng(6110515095)
```

mdhaber (commented):
The Array API failure in the last run was just https://github.com/scipy/scipy/pull/24039#issuecomment-3602457415. The doc failure was a missing import in an example; fixed in the last commit. Second to last commit just removes a parameterization from a test, and there are no syntax errors (since the test still passes locally), so I've just run CI with `[docs only]`.

steppi:
The function itself looks good to me and seems useful. I'm not so sure about the name. I have not come across the term "Ogive" and it wasn't clear to me even how to pronounce it ( "OH-jyve" apparently).  It seems like it is more often used to refer to the graphical plot itself, despite also sometimes referring to an empirical CDF, and also seems to be rarely used in the literature. I'm +1 for adding this but think we can come up with a better name.

I'd also like to see the documentation fleshed out a little more before merging and can help with that. Is there a relevant Scientific Python discourse thread yet?

mdhaber:
~~There is not a post on the forum.~~ Can you merge gh-24707 so I can resolve merge conflicts here (once), then I'll post?
>  but think we can come up with a better name

I didn't love the name "ogive" either, but it is the result of a lot of searching. "ecdf" would be the obvious name, but it's taken. LLMs steered me away from "iquantile" but I do think it's the most descriptive. Best of luck!

*Update: went ahead and posted [here](https://discuss.scientific-python.org/t/scipy-stats-new-ecdf-function-adjustment-to-quantile-test-dtype-declare-ranksums-legacy/2266).*

steppi:
> There is not a post on the forum. Can you merge [gh-24707](https://github.com/scipy/scipy/pull/24707) so I can resolve merge conflicts here (once), then I'll post?

Sure, I'll try to get the review for #24707 done today. I like `iquantile` better than `ogive` but hope to come up with something better.

steppi:
I think Empirical CDF typically refers to the familiar step function, not just any data-driven estimate of the CDF, so `ecdf` doesn't seem like the right name either even if it were available. Under the expectation the `ecdf` is a step function, the line in the documentation saying this computes the empirical distribution function also seems inaccurate since:

```rst
Only the continuous methods are available at this time.
```

If we offer `quantile`'s method 1, then this would for computing the `ecdf` though. Are there any difficulties in adding it or is it for consistency with the term Ogive which [Wolfram MathWorld](https://mathworld.wolfram.com/Ogive.html), at least, defines as any continuous cumulative frequency curve?  Since the `ecdf` is usually understood to be a step function and it seems ojive is usually understood to be continuous, I don't think either name fits.

I think `iquantile` or `inv_quantile` are fine as names. I know it seems kind of odd, since the quantile function is the inverse of the cdf, so it's like naming it `inverse inverse cdf`, but I think it makes sense for the sake of discoverability since the name expresses what the function does: it inverts `scipy.stats.quantile` with respect to `p`.

mdhaber:
Yeah, we can split hairs in a lot of ways. 

- "Empirical CDF" _typically_ refers to a very specific step function (`quantile` with `inverted_cdf`). That's why in the top post I compared `ogive` with `scipy.stats.ecdf` and wrote "The ECDF is `ogive` with `method='inverted_cdf'`.".
- "ogive" _typically_ seems to refer to a continuous curve.
- "inverse" - unless used in a more or less standard phrase like inverse CDF - is is _technically_ appropriate only when a function is invertible. (We've even seen arguments against `icdf`/"inverse CDF", _despite_ it being a pretty standard term.)

The original name of this function was `iquantile`, hence the feature branch name. After much consideration, I guessed that `ogive` was the least offensive term, considering that [Wikipedia's](https://en.wikipedia.org/wiki/Ogive_(statistics)) definition is even more broad than MathWorld's:

> In statistics, an ogive ... can refer to .. any empirical cumulative distribution function.

With this definition, the name `ogive` and use of the term "empirical distribution function" in the docs seems fine, no? In any case, I doubt that as many people get their kicks policing the term "ogive" as "inverse".

But if you still prefer naming it `iquantile` and describing as "inverse quantile" instead of "empirical CDF", I will change it, because I think it's the most accurate and discoverable, too. I just think it's just important to admit that none of these names will match everyone's strictest definition; if we go with one of these, we can only guess which is going to be the least worst / subjective best.

steppi:
> considering that [Wikipedia's](https://en.wikipedia.org/wiki/Ogive_(statistics)) definition is even more broad than MathWorld's:

My impression is that that Wikipedia article is of low quality and the content does not accurately reflect what is in the cited sources. Point taken that `inverse_quantile` and shorthands for it are also bad names because it is not a mathematical inverse. How do you feel about `estimated_cdf``? That seems to sum up what the function does pretty well.

scipy/stats/_quantile.py
```diff
@@ -519,3 +543,265 @@ def _xp_searchsorted(x, y, *, side='left', xp=None):
...
+        p = 1 / (n - 1) * (j +  (     y - z[j])
+                              / (z[j+1] - z[j]))
+
+    where the index ``j`` is that of the largest element of ``z`` that does not exceed
+    ``y``, and ``n`` is the number of elements in the sample. Note that if ``y`` is an
+    element of ``z``, then ``j`` is the index such that ``y = z[j]``, and the formula
+    simplifies to the intuitive ``j / (n - 1)``. The full formula linearly interpolates
+    between ``j / (n - 1)`` and ``(j + 1) / (n - 1)``.
+
+    This is a special case of the more general:
```
mdhaber:
```suggestion
    This is a special case of the more general::
```

mdhaber:
I don't think the names involving "inverse" are bad. They are descriptive and understandable; they just don't adhere to the strict mathematical definition of inverse. That happens sometimes with names. 

I've seen ["generalized inverse"](https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://people.math.ethz.ch/~embrecht/ftp/generalized_inverse_springer&ved=2ahUKEwif_NCe5KOTAxX8HEQIHVdANAQQFnoECD0QAQ&usg=AOvVaw2amZ7IW89RgIZuX6obTfP0) used to refer to inverse-like things in statistics. We could name the function `iquantile` and clarify in the text that it is really a generalized inverse. Or we point to the default "linear" method, which really is an inverse of the corresponding   method of `quantile` (for continuous data).

"Estimated CDF" is also descriptive. It brings up an asymmetry, though - why is `quantile` not "estimated quantile"?

"Plotting positions" is maybe the term most closely related to this family of formulas, although usually it is only evaluated at the observations themselves (like `scipy.stats.mstats.plotting_positions`). The `ecdfPlot` function of R's EnvStats package has an option to use different values of a plotting position constant (a special case of the two-parameter family).

steppi:
> "Estimated CDF" is also descriptive. It brings up an asymmetry, though - why is `quantile` not "estimated quantile"?

It's common to refer to sample quantiles as just quantiles, and sample $p^{\mathrm{th}}$ quantiles for fixed $p$ are commonly used in isolation as descriptive statistics. The estimated cdf at a fixed point `x` is not so commonly used, and most applications involving the empirical cdf involve its global properties. Having `quantile` and `estimated_cdf` would follow typical language use in statistics I think.

mdhaber:
OK; changed to `estimated_cdf`.
Failure appears to be unrelated, but I don't see it in other PRs. If it continues, I'll just make it an `xp_assert_close`.

steppi:
Thanks @mdhaber. I'm happy with the name now. I think the implementation looks good. I'm working on some suggestions for the docstring and should have those out soon.

mdhaber:
OK, if it sounds good, I'll go ahead and squash the existing commits, and you can put one on top?

steppi:
> OK, if it sounds good, I'll go ahead and squash the existing commits, and you can put one on top?

That works for me. Here is what I have in case you want to take a look at it. 

```rst
    Vectorized estimation of cumulative distribution functions.

    Parameters
    ----------
    x : array_like of real numbers
        Array of samples for which cdf estimates are sought. `x` may be
        multidimensional, with separate samples along a specified `axis`.
    y : array_like of real numbers
        Points at which to evaluate the estimated cdf(s).
        See `axis`, `keepdims`, and the examples for broadcasting behavior.
    method : str, default: 'linear'
        The method to use for estimating the cumulative distribution function.
        The available options, numbered as they appear in [1]_, are:

        1. 'inverted_cdf' (AKA *the* empirical cumulative distribution function)
        2. 'averaged_inverted_cdf'
        3. 'closest_observation'
        4. 'interpolated_inverted_cdf'
        5. 'hazen'
        6. 'weibull'
        7. 'linear'  (default)
        8. 'median_unbiased'
        9. 'normal_unbiased'

        See Notes for details.
    axis : int or None, default: 0
        Axis of `x` along which samples are given.


```

Let me know if any of this is wrong, because that would mean I don't understand things completely yet.

steppi:
This seems like it will work well together with `marray`, since then it will be easy to work with samples in `x` of different lengths.

mdhaber:
> This seems like it will work well together with marray

Yes - it does support `MArray`, at least with masked values in the sample. `quantile` and this one might need an adjustment/tests to support masked values for the second argument, since the mask would just propagate.

I don't see anything that's wrong. Maybe comment on the rationale for the changes - is the main idea to emphasize that it supports separate samples in a single, vectorized call? Sure, that's true, and with MArray, they can effectively be different lengths.

If that's the intent, might make it a separate PR since the same could be said of `quantile`.

steppi:
> I don't see anything that's wrong

Sorry, I was just making an offhand positive comment about how things are shaping up. I thought for a moment, "It's cool that we can have arrays of multiple samples, but the samples aren't always the same size. Oh, but wait, one could just mask out 'missing' sample values with marray. This is pretty cool."

mdhaber:
Shoot, I forgot to push `main` here before doing the force push. That would have made GH's "Compare" a lot easier to read. But in case it helps, you can see that the files changed here are identical to those in the [backup branch](https://github.com/mdhaber/scipy/tree/iquantile_backup) that has the full history.

mdhaber:
Oh, yeah. I mean, that's another way I've thought about describing MArray. I'm not sure if it's as flexible as something like Awkward Array, but yeah, you can think of it as supporting ragged arrays.

steppi:
Oh, I see, you were referring to my earlier suggested doc updates. I'll post inline comments with the rationale for the suggested changes.

mdhaber:
> Let me know if any of this is wrong, because that would mean I don't understand things completely yet.

Yeah I meant there was nothing wrong with the suggested doc updates.

I was a little confused about what might be wrong with those suggestions, since they seemed innocuous to me. Pretty much every function in `stats` is vectorized at this point, so pretty much everything can be thought of as supporting separate samples. And pretty much everything supports MArray, so those samples can effectively have different lengths. The only difference here and in `quantile` (compared to other functions) is that instead of them being purely reducing gufuncs, these are implemented with signatures like `(i),(j)->(j)` to avoid sorting the first argument for each element of the second argument. (Ideally, we'd allow this sort of thing in some other functions like `ttest_1samp`, although there's much less of a performance benefit there.)

scipy/stats/_quantile.py
```diff
@@ -519,3 +545,270 @@ def _xp_searchsorted(x, y, *, side='left', xp=None):
...
     out = xp.astype(out, xp_default_int, copy=False)
     return out
+
+
+@xp_capabilities(skip_backends=[("dask.array", "No take_along_axis yet.")],
+                 marray=True)
+def estimated_cdf(x, y, *, method='linear',
+                  axis=0, nan_policy='propagate', keepdims=None):
+    """
+    Estimate the cumulative distribution function of the data along the specified axis.
```
steppi:
```suggestion
    Vectorized estimation of cumulative distribution functions.
```

I found the original kind of confusing due to the compression that was necessary to get it under the line length limit. 

Some points of confusion for me, the shorthand of saying "the cumulative distribution of the data" instead of "the cumulative distribution of the distribution underlying the data". Also, the suggestion that there is just one estimated cumulative distribution function here, when actually, this produces estimates of multiple cdfs underlying multiple samples.

Rather than trying to fit in an unambiguous description of what's going here, the suggestion is just meant as a generic indicator of what the function does.

scipy/stats/_quantile.py
```diff
@@ -519,3 +545,270 @@ def _xp_searchsorted(x, y, *, side='left', xp=None):
...
+
+    Parameters
+    ----------
+    x : array_like of real numbers
+        Data array.
+    y : array_like of real numbers
+        Datum or data for which to estimate the cumulative probabilities.
```
steppi:
```suggestion
  x : array_like of real numbers
        Array containing a sample for which a cdf estimate is sought or
        separate samples for which cdf estimates are sought (`x` may be
        multidimensional, with the separate samples along a specified `axis`).
    y : array_like of real numbers
        Points at which to evaluate the estimated cdf(s).
        See `axis`, `keepdims`, and the examples for broadcasting behavior.
```

I found the original a little unclear because of the overloading of the term data. We have a data array `x` and then data `y` for which to estimate cumulative probabilities. It suggests to me that the points in `y` might have to overlap with the points in `x`, but the values in `y` need not be existing data points, they can be any points.

scipy/stats/_quantile.py
```diff
@@ -519,3 +545,270 @@ def _xp_searchsorted(x, y, *, side='left', xp=None):
...
+        4. 'interpolated_inverted_cdf'
+        5. 'hazen'
+        6. 'weibull'
+        7. 'linear'  (default)
+        8. 'median_unbiased'
+        9. 'normal_unbiased'
+
+        See Notes for details.
+    axis : int or None, default: 0
+        Axis along which the calculation is performed.
```
steppi:
```suggestion
        Axis of `x` along which samples are given.
```

Just replacing the generic boilerplate language with more specific language.

steppi (commented):
I've made comments with the rationale behind the suggested docstring edits. The reason behind making them is because I found the docstring hard to understand at first. The docstring for `quantile` makes sense to me, and this evolved from it, but the the context is different enough that it became confusing to me.

steppi:
> I was a little confused about what might be wrong with those suggestions

Ah, sometimes I just have so many different things I'm looking at that I worry I start talking nonsense without realizing it. Looking again, I think there's nothing wrong with them.

scipy/stats/_quantile.py
mdhaber: Description of `y` is fine, and I get the intent, but in the first sentence:

> Array of samples for which cdf estimates are sought.

Is "samples" plural to focus on the multidimensional case? I think the first sentence should describe the typical case, which is probably 1D array. 

If by "samples" you meant "observations", please make the substitution to avoid the ambiguity (e.g. with "samples" in the second sentence".

steppi: Yeah, it was plural to focus on the multidimensional case. Having "along the given axis" in the first line of the original docstring lead me to think of the multidimensional case first. I'll change it to first describe the 1D case and then the ND case.

steppi: I've updated the suggestion.

scipy/stats/_quantile.py
mdhaber: > I found the original kind of confusing due to the compression that was necessary to get it under the line length limit.

I agree that "of the data" should be "of the distribution underlying the data", and that wouldn't fit well in the limit. (We could have a huge PR just cleaning that sort of thing up throughout `scipy.stats`.)

> this produces estimates of multiple cdfs underlying multiple samples.

While pretty much every statistic and hypothesis test in ScPy is vectorized, we usually focus the documentation on the simple case. I don't think it the vectorization of *this* function is particularly important to emphasize in the one-line summary, so I would prefer "Estimate the cumulative distribution function from a sample."[^1] or something. I shouldn't have included "along the specified axis" before. I suppose it happened because I was adapting the documentation of `quantile`, but I should have simplified that instead of carrying it over here.

[^1]: note that it's intentionally "from a sample", not "of a sample". We leave it implicit that it's **of** an underlying distribution.

steppi: I see, that makes sense. I think most of my confusion came from having "along the given axis" in the first line coupled with it not being clear to me what that means in the rest of the docstring. My suggestions centered around trying to bring the same emphasis to the ND case that "along the given axis" suggested to me, while clearing up the points that had me confused.

steppi:
```suggestion
    Estimate the CDF for the distribution underlying a sample.
```

New suggestion, in light of new info here (https://github.com/scipy/scipy/pull/23949#discussion_r2949533910). I had the wrong impression of the importance of the ND case due to the phrase "along the given axis". If it isn't important to include that, then we can just fit underlying the data `x`.

scipy/stats/_quantile.py
steppi:
```suggestion
    y : array_like of real numbers
        Points at which to evaluate the estimated cdf.
```

scipy/stats/_quantile.py
steppi:
```suggestion
        Axis along which samples in `x` are given in ND case.
```

steppi (commented):
Here's a new batch of suggestions @mdhaber. I agree with your pattern of emphasizing the 1D case. Sorry if this discussion has come off as pedantic.

scipy/stats/_quantile.py
mdhaber: We're [not supposed to](https://numpydoc.readthedocs.io/en/latest/format.html#short-summary) include variable names in the one-line summary.

image

mdhaber:
OK, so ignore the old suggestions and add the new ones?

scipy/stats/_quantile.py
steppi: Oh, good to know. They appear all over `special` (https://docs.scipy.org/doc/scipy/reference/special.html) and that's what I'm most familiar with. I guess `special` is...

steppi:
> OK, so ignore the old suggestions and add the new ones?

Yeah. I tried to hide the old suggestions as outdated. Not sure if that succeeded.

scipy/stats/_quantile.py
steppi: I've updated the suggestion to not use `x`.

steppi (approved):
I think this looks good to go, pending agreement on the latest docs suggestions.

steppi:
Thanks @mdhaber. This looks good to me and I think the silence after the discourse post can be taken as implied consensus. I made sure to do a regular merge to keep the main work and the docs refinements separate.

mdhaber:
Thanks for the review! We're getting pretty close to gh-22194. I think the only thing left is Maritz-Jarrett standard error / confidence intervals for quantiles (https://github.com/scipy/scipy/issues/22194#issuecomment-4014528168 and gh-24757).

scipy/stats/tests/test_quantile.py
```diff
@@ -428,3 +433,342 @@ def test_nd(self, side, ties, shape, nans_x, infs_x, xp):
...
+         ([-np.inf, -1, 0, 1, np.inf], [-np.inf, -2, -1, 0, 1, 2, np.inf],
+          [0.25, 0.25, 0.25, 0.5 , 0.75, 0.75, np.nan], {'method':'linear'}),
+         ([-np.inf, -1, 0, 1, np.inf], [-np.inf, -2, -1, 0, 1, 2, np.inf],
+          [0.2, 0.2, 0.4, 0.6, 0.8, 0.8, 1.], {'method': 'inverted_cdf'}),
+        ])
+    def test_edge_cases(self, x, y, ref, kwargs, xp):
+        if kwargs.get('method', None) == 'weibull' and is_torch(xp):
+            pytest.skip('data-apis/array-api-compat#360')
+        if kwargs.get('axis', None) == -1 and is_cupy(xp):
+            pytest.skip('Fails; need to investigate.')
```
mdhaber, line 752: Investigated. CuPy raises `ZeroDivisionError` when sorting multidimensional array along axis of length zero; see cupy/cupy#9816.

PR #23951, opened 2025-11-08
DOC: interpolate: add tutorial on how to fill an image from its boundary

DWesl:
#### Reference issue

Related to #21910
Replaces #23285

This tutorial documents ways to estimate the interior of an image given only values on the boundary.  Much of that consists of reducing the input or requested points, to avoid NaNs or to reduce computational time.

The simplex tolerance section depends on #23737, and should probably be commented out until that is merged.

#### Additional information

The original query was about estimating the background illumination on a microscope slide.

[docs only]


doc/source/tutorial/interpolate/2D_fill_from_border.rst
```diff
@@ -0,0 +1,408 @@
...
+
+That is, a sine wave with three full periods on each side, identical
+on opposite sides of the square.
+
+------------------
+Analytic solutions
+------------------
+
+With this simple boundary condition, we can find a few analytic
+solutions as a baseline.
```
ev-br, line 42: I'm not entirely sure what is the problem these are solutions for. You mentioned estimating the background above, maybe expand it a little? This whole story started from a problem, maybe discuss it a bit? Even if you start from a real-world case and then quickly turn into a simplified artificial example.

doc/source/tutorial/interpolate/2D_fill_from_border.rst
```diff
@@ -0,0 +1,408 @@
...
+   >>> plot_and_check_filled_image(values, "Laplace polynomial")
+   Laplace polynomial: Mean error: -3.55e-17 Error std: 0.663
+
+--------------
+Other packages
+--------------
+
+There are packages that will solve Laplace equation with given boundary conditions.
+
+Scikit-Image has an ``inpaint_biharmonic`` function that will solve
```
ev-br: Would be nice to turn it into a link to the scikit-image function. While at at --- scikit-image has several inpainting functions, and you only mention the biharmonic one here. Would be nice to know the reason (this tags to the "what is the problem we're solving" question above).

doc/source/tutorial/interpolate/2D_fill_from_border.rst
```diff
@@ -0,0 +1,408 @@
...
+also fill an image from the boundary conditions, in a manner related
+to the Laplace Equation.
+
+
+Statsmodels offers a regression interface that hides the linear
+algebra and offers additional diagnostics.
```
ev-br: Again, would be nice to have a link to a specific statsmodels function(s), and some reasoning to which statsmodels functionality I as a reader should be looking at and why.

ev-br (commented):
Looks nice on a cursory glance!
Will do a more careful read-through later. Meanwhile, it would be nice to understand and see
- how does this relate to other tutorial pages, where should this be linked from
- (for a non-expert like me) what is the problem(s) this solves? Is there a way to have expand the motivation / what are the requirements / what are the constraints?
- is there a way to relate the material in this tutorial to a real-world problem (e.g. the one that started https://github.com/scipy/scipy/issues/21910, was it something about microscopy images?

rgommers:
This looks like a nice addition to me. The doc build in CI can be fixed by adding an entry for the new file in `doc/source/tutorial/interpolate.rst`.

doc/source/tutorial/interpolate/2D_fill_from_border.rst
DWesl: Scikit-image has two that I know of: inpaint_biharmonic to fill a region, and rolling_ball to estimate background from nearby points.  The former is designed for nearly this exact setup, though it works a bit slowly, The latter is designed for many small features rather than one big one as this implicitly assumes: the example in the documentation is a bad scan of a printed page.

DWesl:
> Looks nice on a cursory glance! Will do a more careful read-through later. Meanwhile, it would be nice to understand and see
> 
>   * how does this relate to other tutorial pages, where should this be linked from

Not terribly well.  There's a suggestion to add it to `tutorials/interpolate.rst`.  NumPy's [Diátaxis](https://diataxis.fr/) approach would probably put this as a how-to, but SciPy doesn't seem to have adopted that.

>   * (for a non-expert like me) what is the problem(s) this solves? Is there a way to have expand the motivation / what are the requirements / what are the constraints?
> 

I suspect I need to think about that more clearly.  The contrived example is the kind I'm most familiar with, but anyone with that kind of problem will likely come across it in the context of solving a Laplace equation and will go for Laplace solvers before SciPy.  

>   * is there a way to relate the material in this tutorial to a real-world problem (e.g. the one that started #21910, was it something about microscopy images?

Yes, estimation of uneven background illumination to isolate the nearly-transparent specimen from effects due to the positioning of the lamp or thin-film interference off the water between the slides.

I suppose we could find a CC-BY stock image of a microscope slide, or ask if the person who posted #21910 has an image they'd be willing to let us downscale and use.  I could also try to apply a point-source illumination model to a few ellipses that look vaguely like a cell.

ev-br:
We can adopt a Diataxis if that helps (not sure what that means but no matter, the structure of scipy.interpolate tutorials is not set in stone).

> The contrived example is the kind I'm most familiar with, but anyone with _that kind of problem_ will likely come across it in the context of solving a Laplace equation and will go for Laplace solvers before SciPy.

(emphasis mine) My actual question is what is that kind of problem? Just reading what's written leaves me (a non-expert!) very confused: if we're talking about the boundary value problem for the Laplace equation, then it's a problem with a theoretically unique solution (provided some conditions are met etc). If it's _estimating the background_, then is there a specific model for the background, or what are the constraints for the estimation?
(Or does the Laplace BVP appear as an example where we just know the correct solution).

DWesl:
> We can adopt a Diataxis if that helps (not sure what that means but no matter, the structure of scipy.interpolate tutorials is not set in stone).

The relevant bit is division into tutorial, how-to, explanation, and reference, based on whether they're designed to help acquire or apply knowledge and another axis I can't summarize as easily, but discussion of whether SciPy should adopt [NEP 44](https://www.numpy.org/neps/nep-0044-restructuring-numpy-docs) should probably be left for the mailing list.

> > I suspect I need to think about that more clearly. The contrived example is the kind I'm most familiar with, but anyone with _that kind of problem_ will likely come across it in the context of solving a Laplace equation and will go for Laplace solvers before SciPy.
> 
> (emphasis mine) My actual question is what is that kind of problem? Just reading what's written leaves me (a non-expert!) very confused:  if we're talking about the boundary value problem for the Laplace equation, then it's a problem with a theoretically unique solution (provided some conditions are met etc). 

That's the bit I need to think about.  My first reaction on seeing "I have values near the edges, and want values near the center", especially with the rectangular domain, is to treat is at a Laplace problem.  The theoretical guarantee that a solution exists and is unique is great ... if the Laplace equation describes what they want.  Laplace solutions tend to be *very* smooth, which might be what any given user wants, or might not be.

> If it's _estimating the background_, then is there a specific model for the background, or what are the constraints for the estimation? (Or does the Laplace BVP appear as an example where we just know the correct solution).

For the general case, I can't tell.  Part of what I was trying to do was to show that you need additional constraints to get a unique solution, and show some of those constraints

If the Laplace equation is a poor match to their problem, my next conceptualization of the problem is "I'm trying to predict a value at these points based on these other points", which is a regression problem, and polynomial regression is reasonably straightforward conceptually (I can also write those examples using `optimize.curve_fit` instead of rolling my own linear regression.

For the microscope illumination problem, I could find an equation for diffuse reflection of a point source from a plane toward a point observer and try to fit the five parameters, but that feels like a better fit for a microscopy package than for SciPy.

The lack of an actual problem statement is part of why I made #23285 first.  Elaborating on why `interpolate.griddata` is a weird choice for whatever problem results in information about the edges but not the middle and a need for the middle is why I made this PR.

YuzeHao2023 (approved):
lgtm

doc/source/tutorial/interpolate/2D_fill_from_border.rst
```diff
@@ -0,0 +1,420 @@
...
+==============================
+Filling an image from a border
+==============================
+
+Suppose we have values for a function only on the border of a region,
+either to estimate the background of an image from the parts not
+covered by the sample, or as part of a boundary-value problem.  As a
+concrete example, consider the estimation of the (possibly uneven)
+background illumination on a microscope slide from the edge of the
+image, away from the specimen.
```
ev-br: It would be helpful to expand this first introductory paragraph. As far as I understand the story:

- Suppose we know the values of a function at a boundary of a region, and we want to extend the function into its interior. 
- For example, we want to estimate the background of a photographic image with potentially uneven illumination. 
- In general, we not have a model for the background, thus we reach to interpolation.
- Below we work out a toy illustrative example, which has an analytic solution: a BVP problem with a regular boundary.

Does this sound about right? If not --- and there seems to be more to the story --- please feel free to flow it as you see fit.

ev-br (commented):
> discussion of whether SciPy should adopt NEP 44 should probably be left for the mailing list.

Agreed. A clearer split would be nice indeed, needs a champion :-). Whether to start from a ML or from reorganizing existing tutorial pages for one of submodules... Up to that champion, I suppose. 

> division into tutorial, how-to, explanation, and reference, based on whether they're designed to help acquire or apply knowledge

So what you're adding is a how-to, it seems. Which is great! So it's somewhat similar to the set of working examples in https://docs.scipy.org/doc/scipy/tutorial/interpolate/extrapolation_examples.html 
(written without relation to NEP 44 or anything diataxis though).

ev-br:
Now that gh-22831 is in, it'd be if you have to finish the documentation enhancement @DWesl :-).

DWesl:
Python and Matplotlib are disagreeing with each other, so I'm not able to check the plots, but I should be able to rework the text.

ev-br:
`/home/circleci/repo/doc/source/tutorial/interpolate/2D_fill_from_border.rst: WARNING: document isn't included in any toctree` is real, could you please fix it? I'm interested in seeing the plots, too!

EDIT: since this is a doc-only change, can add `[docs only]` to the commit message to only run the doc-related CI.

DWesl:
The RBFInterpolator examples differ in the mean error by an order of magnitude in most cases from what I got with Python 3.9, though not in a consistent direction.  I should probably check the changelog for the last few years to see why.

I'm trying to get Python 3.12 importing SciPy and Matplotlib.

doc/source/tutorial/interpolate/2D_fill_from_border.rst
```diff
@@ -0,0 +1,456 @@
...
+   ...             scaled_grid_x[:, 0], scaled_grid_y[0, :],
+   ...             coeffs.reshape(num_coeffs, num_coeffs) * coeff_multiplier[:, :]
+   ...         )
+   ...         plot_and_check_filled_image(values, f"{poly_family:s} degree {num_coeffs - 1:d}")
+   legendre degree 7: Mean error: -3.2e-16 Error std: 0.41
+   chebyshev degree 7: Mean error: 6.8e-17 Error std: 0.41
+   legendre degree 11: Mean error: 3.4e-16 Error std: 0.024
+   chebyshev degree 11: Mean error: -2.6e-16 Error std: 0.024
+   legendre degree 15: Mean error: 1.4e-16 Error std: 0.0003
+   chebyshev degree 15: Mean error: 7.4e-17 Error std: 0.0003
```
ev-br: drive-by comment: our doctesting machinery will switch to vanilla doctest because of mixing text and numbers. Were it only numbers (specifically, anything which can be np.asarray-ed), the check would have went via np.allclose, thus obviating the need to mess around with fp formats, precision and all that. Which seems to be what CircleCI complains about.

DWesl: Does a rank-3 NumPy array count for that?  The failure is complaining of differences on the order of $10^{-15}$.
The only reason I can think that it wouldn't is the `<BLANKLINE>` tags that continue the example.

ev-br: Indeed, blank lines trip doctests. So either remove them manually in the docstring examples, or https://github.com/scipy/scipy_doctest/pull/216. The former is a hack of course, so thank you for the latter!

DWesl:
It's now flagging only one difference, of order $10^{-5}$.  

It's passing locally, so I'm not sure what's going on: the first thing that comes to mind is Netlib Lapack locally versus OpenBLAS Lapack on GHA doing something with the RBF calculations, but the inputs are probably double precision and that seems large.  I could increase the number of points done directly by the RBF interpolator, but this part already takes longer than the other parts of the example.  

This difference also only appears for the quintic kernel of the RBF interpolator, which I suppose would be less well-conditioned than the other kernels.  Is there a way to increase the absolute tolerance for only one element of one array?
I could also drop the quintic kernel from the example.

ev-br (approved):
Sorry for the delay and thank you for the write-up!
This now reads great, and shows multiple tricks and tips. I wish this existed when I started looking at 2D interpolators in SciPy.

The only thing which stands between me and the green button is 20 commits, which is a bit much. Is there some history you'd want to preserve @DWesl? If so could you please squash it a bit. Otherwise, I'll squash-merge.

DWesl:
More commits than I expected are merges, but most of the rest are fixes for typos, syntax errors, and related issues.  How that wound up divided into commits isn't particularly useful.  The big change I can think of is the shift from string outputs to arrays for scipy-doctest, but I'm not sure I actually got all of that in a single commit.

I added a closing section with a few notes, and added a few section headers under multilinear regression to round out the tutorial.

Thanks for your help in shaping this into something useful for people who aren't me.

ev-br:
Squash-merged, thank you @DWesl , much appreciated!

PR #23956, opened 2025-11-10
MAINT: add `xp-capabilities` to `BSpline`

ev-br:
I was sure `BSpline` already has `xp_capabilities` but apparently not. Thus add it. While at it, unskip a test which does not need to be skipped anymore.

### DIFF
--- a/scipy/interpolate/tests/test_bsplines.py
+++ b/scipy/interpolate/tests/test_bsplines.py
@@ -46,7 +46,7 @@ from scipy.interpolate import _dierckx
 skip_xp_backends = pytest.mark.skip_xp_backends
 
 
-@skip_xp_backends(cpu_only=True)
+@make_xp_test_case(BSpline)
 class TestBSpline:
 
     def test_ctor(self, xp):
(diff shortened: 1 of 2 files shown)


scipy/interpolate/tests/test_bsplines.py
```diff
@@ -130,23 +130,18 @@ def test_degree_1(self, xp):
...
         bspl = BSpline(t, c, k)
 
-        xx = np.linspace(-1., 2., 10)
+        xx = xp.linspace(-1., 2., 10)
         xp_assert_close(bp(xx, extrapolate=True),
                         bspl(xx, extrapolate=True), atol=1e-14)
 
-        t, c = map(np.asarray, (t, c))
-        xp_assert_close(splev(xx, (t, c, k)),
-                        bspl(xx), atol=1e-14)
```
lucascolley, line 148: we don't want this any more?

ev-br: Yes, we test this elsewhere.
 
The purpose of this test is to construct a specific knot vector so that the BSpline result is equivalent to a BPoly result. And then, as a second assertion, it was checking equivalence of a BSpline and splev evaluations. The latter we check elsewhere, e.g. https://github.com/scipy/scipy/blob/main/scipy/interpolate/tests/test_bsplines.py#L1025

j-bowhay (approved):
CI failures look unrelated, thanks

PR #23955, opened 2025-11-10
DOC: list classes in Array API tables

ev-br:
Tweak the Array API table listings to include classes (and exclude warnings and exceptions: we want RegularGridInterpolator and do not want LinAlgWarning).

PR #23957, opened 2025-11-10
Revert DOC: list classes in Array API tables

mdhaber:
Reverts scipy/scipy#23955

I'm seeing some issues in the full run after gh-23955, e.g. [here](https://github.com/scipy/scipy/actions/runs/19236447321/job/54987570074?pr=23810). Thought I'd open this as a quick fix unless there is another easy patch.


ev-br:
Sure let's revert it right away then

PR #23959, opened 2025-11-11
ENH: stats.bws_test: vectorize and add array API support

mdhaber:
#### Reference issue
Toward gh-20544

Vectorizes and adds array API support to `scipy.stats.bws_test`.

#### Additional information
First commit vectorizes, second translates to array API.
Can't easily apply `_axis_nan_policy` decorator to this one because it returns the permutation distribution of the statistic.


scipy/stats/tests/test_hypotests.py
```diff
@@ -2071,22 +2075,24 @@ def test_bws_input_validation(self):
...
+            pytest.skip("Scalar dtypes only respected after NEP 50.")
+        dtype = xp_default_dtype(xp) if dtype is None else getattr(xp, dtype)
+        x = xp.asarray([1, 2, 3, 4, 6, 7, 8], dtype=dtype)
+        y = xp.asarray([5, 9, 10, 11, 12, 13, 14], dtype=dtype)
         res = stats.bws_test(x, y, alternative='two-sided')
-        assert_allclose(res.statistic, 5.132, atol=1e-3)
-        assert_equal(res.pvalue, 10/3432)
-
+        xp_assert_close(res.statistic, xp.asarray(5.132, dtype=dtype), atol=1e-3)
+        xp_assert_equal(res.pvalue, xp.asarray(10/3432, dtype=dtype))
```
mdhaber:
```suggestion
        xp_assert_close(res.pvalue, xp.asarray(10/3432, dtype=dtype))
```

mdhaber: Also, add NaN propagation test.

scipy/stats/_bws_test.py
```diff
@@ -167,13 +194,13 @@ def bws_test(x, y, *, alternative="two-sided", method=None):
...
 
-    x, y, alternative, method = _bws_input_validation(x, y, alternative,
-                                                      method)
-    bws_statistic = partial(_bws_statistic, alternative=alternative)
+    x, y, alternative, axis, method, xp = _bws_input_validation(x, y, alternative,
+                                                                axis, method)
+    bws_statistic = partial(_bws_statistic, alternative=alternative, xp=xp)
 
     permutation_alternative = 'less' if alternative == 'less' else 'greater'
     res = stats.permutation_test((x, y), bws_statistic,
```
mdhaber, line 202: `bws_test` does whatever `permutation_test` does with NaNs and empty input. I noticed that `permutation_test` didn't have a test with NaN input and the behavior was not ideal[^1], so I fixed it and added a test in this PR.

[^1]: The p-value was `0` for an exact test or `1 / (n_resamples + 1)` for a randomized permutation test because zero elements of the null distribution are more extreme than NaN... but the same could be said of any stats function. We shouldn't return a numerical p-value with a NaN statistic.

scipy/stats/_bws_test.py
```diff
@@ -93,6 +117,9 @@ def bws_test(x, y, *, alternative="two-sided", method=None):
         Under a more restrictive set of assumptions, the alternative hypotheses
         can be expressed in terms of the locations of the distributions;
         see [2] section 5.1.
+    axis : int or None, default: 0
```
j-bowhay, line 120: The docs on line 102 still state `1-d arrays of samples.`

scipy/stats/tests/test_hypotests.py
```diff
@@ -2071,22 +2075,24 @@ def test_bws_input_validation(self):
...
-
-    def test_against_published_reference(self):
+    @pytest.mark.parametrize('dtype', [None, 'float32', 'float64'])
+    def test_against_published_reference(self, dtype, xp):
         # Test against Example 2 in bws_test Reference [1], pg 9
         # https://link.springer.com/content/pdf/10.1007/BF02762032.pdf
-        x = [1, 2, 3, 4, 6, 7, 8]
-        y = [5, 9, 10, 11, 12, 13, 14]
+        if is_numpy(xp) and xp.__version__ < "2.0" and dtype == 'float32':
+            pytest.skip("Scalar dtypes only respected after NEP 50.")
```
j-bowhay:
```suggestion
```
I think we now have 2.0.0 as the min version

j-bowhay (changes requested):

mdhaber:
Thanks, Jake!

j-bowhay (approved):
Thanks!

Issue #23510, opened 2025-08-26
BUG: NdBSpline segfaults for too high derivative orders


ev-br:
### Describe your issue.

`NdBSpline` evaluations segfault if the requested derivative order is too high. We should probably filter out clearly nonsense inputs and return zeros instead.

### Reproducing Code Example

```python
Apply 

$ git diff
diff --git a/scipy/interpolate/tests/test_bsplines.py b/scipy/interpolate/tests/test_bsplines.py
index 87dd7e5a03..ea56c7ac86 100644
--- a/scipy/interpolate/tests/test_bsplines.py
+++ b/scipy/interpolate/tests/test_bsplines.py
@@ -2583,6 +2583,8 @@ class TestNdBSpline:
             (0, 0): lambda x, y: x**3 * (y**2 + 2*y),
         }
 
+        breakpoint()
+
         for nu, expected_fn in test_cases.items():
             expected_vals = [expected_fn(x, y) for x, y in xi]



and


$ spin test scipy/interpolate/tests/test_bsplines.py::TestNdBSpline::test_2D_derivative 
Invoking `build` prior to running tests:
$ meson compile -j 4 -C build
INFO: autodetecting backend as ninja
INFO: calculating backend command to run: /home/br/miniforge3/envs/scipy-dev/bin/ninja -C /home/br/repos/scipy/scipy/build -j 4
ninja: Entering directory `/home/br/repos/scipy/scipy/build'
[2/338] Generating subprojects/highs/src/HConfig.h with a custom command
$ meson install --only-changed -C build --destdir ../build-install --tags=runtime,python-runtime,tests,devel
$ export PYTHONPATH="/home/br/repos/scipy/scipy/build-install/usr/lib/python3.12/site-packages"
$ /home/br/miniforge3/envs/scipy-dev/bin/python3.12 -P -c 'import scipy'
$ cd /home/br/repos/scipy/scipy/build-install/usr/lib/python3.12/site-packages
$ /home/br/miniforge3/envs/scipy-dev/bin/python3.12 -P -m pytest -m 'not slow' scipy/interpolate/tests/test_bsplines.py::TestNdBSpline::test_2D_derivative
=========================================================================================== test session starts ===========================================================================================
platform linux -- Python 3.12.0, pytest-8.4.1, pluggy-1.6.0
[...]

PR #23962, opened 2025-11-11
ENH: interpolate: Python bivariate spline fitter `_regrid`

czgdp1807:
#### Reference issue

This work was prototyped and iterated in my fork, with early discussion and review captured in PR https://github.com/czgdp1807/scipy/pull/22. That thread documents the initial design choices, preliminary benchmarks, and API shape. Feedback gathered there informed the current structure, naming, and documentation included in this PR.

**Summary**

This PR adds a new implementation of the classic REGRID workflow in Python + C++ combination: ``_regrid`` plus a small helper ``_ndbspline_call_like_bivariate``. It fits a 2-D tensor-product B-spline surface to gridded data using separable 1-D FITPACK/Dierckx kernels (design matrices, banded QR, and back-substitution). The solver returns an ``NdBSpline``, and the convenience wrapper lets users call it with the same ergonomics as ``RectBivariateSpline.__call__``.
The implementation mirrors FITPACK's mathematics, including the same 1/p penalty convention and the classic residual-guided knot-growth strategy, while making each step explicit and well-documented in Python for maintainability and extensibility.

**Files changed**

- `scipy/interpolate/_regrid.py` (new): Core implementation containing detailed module-level documentation of the algorithm, mathematics, and design choices. Implements the two-pass separable solve strategy using 1-D FITPACK kernels.
- `scipy/interpolate/_fitpack_repro.py`: Minor enhancement to the `root_rati` helper function, adding `maxit` as a configurable parameter with default value `MAXIT` for more flexible control over the iteration limit.
- `benchmarks/benchmarks/interpolate.py`: New `RectBivariateSplineVsRegridPython` benchmark class comparing fit time of `RectBivariateSpline` against the new `regrid` across three grid sizes and two smoothing parameter values.

**Motivation**

- Provide a Fortran-free, readable, and maintainable path for bivariate spline fitting that matches the behavior and API expectations of ``RectBivariateSpline``.
- Expose the algorithm’s moving parts (knot growth, residual projection, p-search, banded solves) for easier debugging, benchmarking, and future optimization.

**What’s included**

- `_regrid(x, y, z, *, bbox=None, kx=3, ky=3, s=0.0, maxit=50)` (**Private API**)
Private function (currently in `scipy.interpolate._regrid`) that validates inputs, performs adaptive knot growth, and optionally searches for a smoothing parameter `p` so that `fp(p) ~ s`. Returns an `NdBSpline`. To use: `from scipy.interpolate._regrid import _regrid`.
- `_ndbspline_call_like_bivariate(ndbs, x, y, dx=0, dy=0, grid=True)` (**Private API**)
Private helper function (in `scipy.interpolate._regrid`) providing a thin wrapper so an `NdBSpline` object can be evaluated with the same signature pattern and shape semantics as `RectBivariateSpline.__call__`.
- Clear docstrings and a module-level narrative:
    - FITPACK-consistent 1/p penalty scaling.
    - `p == −1` is a sentinel for the interpolatory limit (`p = inf`), i.e. no penalty rows are stacked.
    - Residual projection guides data-aligned knot insertion.
    - Separable two-pass solve: x-direction then y-direction, both via banded QR and backsolve.

**Design notes**

- **Mathematics**: minimize |A c − z|^2 + (1/p)|D c|^2 with augmented systems `[A; D/p]`. Identical to FITPACK’s convention.
- **Separable composition**: build the 2-D fit from 1-D banded operators (`data_matrix`, `disc`, `qr_reduce`, `fpback`) instead of a single monolithic subroutine.
- **Knot growth**: project residuals into row/column spans, then add knots where residual is high (respecting storage and interpolation caps). When both axes are at minimal size, record `fp0` for bracketing the `p` search.
- **p search**: use `g(p) = fp(p) − s`, with `p = inf` represented by `p == −1`. Smaller p -> stronger smoothing because the penalty weight is 1/p.
- **Parametrized iteration limit**: the `root_rati` helper in `_fitpack_repro.py` now accepts `maxit` as a parameter (with default `MAXIT`), providing fine-grained control for the root-finding step in the p-search.
- **Evaluation path**: convert the packed banded matrices to CSR only for residual evaluation because packed×dense multiplication is not implemented; CSR supports standard `@`.

**API compatibility**

- Input/shape/monotonicity rules align with `RectBivariateSpline`.
- The returned object is an `NdBSpline`. With `_ndbspline_call_like_bivariate`, users can evaluate it with the same calling pattern as `RectBivariateSpline.__call__`.
- The fitter passes the same test suite that `RectBivariateSpline` passes for the covered scenarios.

**Performance summary**

- **Interpolation-like (very small s, e.g., s ~ 1e−12)**: the new path is faster than `RectBivariateSpline` on tested sizes.

|   nx |   ny |   kx |   ky |     s |   SciPy fit (s) — median |   Custom fit (s) — median | Custom/SciPy   |
|------|------|------|------|-------|--------------------------|---------------------------|----------------|
|  256 |  512 |    3 |    3 | 1e-12 |                 0.226999 |                  0.119361 | 0.53x          |
|  512 | 1024 |    3 |    3 | 1e-12 |                 1.03052  |                  0.66797  | 0.65x          |
| 1024 | 1024 |    3 |    3 | 1e-12 |                 1.82459  |                  1.51854  | 0.83x          |

- **Smoothing (finite s)**: current implementation is slower than `RectBivariateSpline`, especially as grids grow. Overhead sources include repeated CSR evaluations for residuals, Python-level control flow, and the generic root search around `p`.

|   nx |   ny |   kx |   ky |   s |   SciPy fit (s) — median |   Custom fit (s) — median | Custom/SciPy   |
|------|------|------|------|-----|--------------------------|---------------------------|----------------|
|  256 |  512 |    3 |    3 |   3 |                 0.193432 |                  0.308117 | 1.59x          |
|  512 | 1024 |    3 |    3 |   3 |                 0.627185 |                  1.4102   | 2.25x          |
| 1024 | 1024 |    3 |    3 |   3 |                 1.03742  |                  2.93738  | 2.83x          |

|   nx |   ny |   kx |   ky |   s |   SciPy fit (s) — median |   Custom fit (s) — median | Custom/SciPy   |
|------|------|------|------|-----|--------------------------|---------------------------|----------------|
|   64 |   64 |    3 |    3 | 0.2 |                 0.002661 |                  0.007046 | 2.65x          |
|  128 |  128 |    3 |    3 | 0.2 |                 0.012084 |                  0.021517 | 1.78x          |
|  256 |  256 |    3 |    3 | 0.2 |                 0.042626 |                  0.065456 | 1.54x          |
|  512 |  512 |    3 |    3 | 0.2 |                 0.400949 |                  0.738505 | 1.84x          |
| 1024 | 1024 |    3 |    3 | 0.2 |                 1.32275  |                  3.71267  | 2.81x          |

**Correctness**

- The implementation reproduces `RectBivariateSpline` behavior across the same tests.
- Mathematical choices (1/p scaling, `p == −1` sentinel, residual-driven knot growth) are consistent with FITPACK’s REGRID.

**Future plans**

`_regrid` is presently a private API. We intend to replace the existing implementation used under the hood in `RectBivariateSpline` with the one added here. Currently it is not exposed directly; one would need to import from `scipy.interpolate._regrid`.

#### Additional information

I’ve posted a community update on the new Python-based REGRID implementation (`_regrid`) here: https://discuss.scientific-python.org/t/enh-interpolate-python-bivariate-spline-fitter-regrid-python/2219. It summarizes the design, FITPACK compatibility, knot-growth strategy, smoothing search, and usage notes. Feedback from SciPy users is welcome.

cc: @rgommers @ev-br @melissawm

### DIFF
--- a/scipy/interpolate/__init__.py
+++ b/scipy/interpolate/__init__.py
@@ -142,8 +142,6 @@ functional.
    :toctree: generated/
 
    RectBivariateSpline
-   regrid_python
-   ndbspline_call_like_bivariate
    RectSphereBivariateSpline
 
 **For unstructured data (OOP interface):**
@@ -210,8 +208,6 @@ from ._bsplines import *
 from ._fitpack_repro import generate_knots, make_splrep, make_splprep
 
 from ._pade import *
-
-from ._regrid_python import regrid_python, ndbspline_call_like_bivariate
 
 from ._rgi import *
 
--- a/scipy/interpolate/__init__.py
+++ b/scipy/interpolate/__init__.py
@@ -142,6 +142,7 @@ functional.
    :toctree: generated/
 
    RectBivariateSpline
+   regrid_python
    RectSphereBivariateSpline
 
 **For unstructured data (OOP interface):**
@@ -208,6 +209,8 @@ from ._bsplines import *
 from ._fitpack_repro import generate_knots, make_splrep, make_splprep
 
 from ._pade import *
+
+from ._regrid_python import regrid_python, ndbspline_call_like_bivariate
 
 from ._rgi import *
 
--- a/scipy/interpolate/__init__.py
+++ b/scipy/interpolate/__init__.py
@@ -143,6 +143,7 @@ functional.
 
    RectBivariateSpline
    regrid_python
+   ndbspline_call_like_bivariate
    RectSphereBivariateSpline
 
 **For unstructured data (OOP interface):**
--- a/scipy/interpolate/_regrid.py
+++ b/scipy/interpolate/_regrid.py
@@ -1,0 +1,1110 @@
+"""
+Regrid (2-D smoothing B-splines via separable 1-D FITPACK kernels)
+==================================================================
+
+1) Overview
+-----------
+This module fits a bivariate tensor-product B-spline surface to gridded data
+`Z[i, j]` sampled on strictly increasing coordinates `x[i]`, `y[j]`. It mirrors
+the adaptive spirit of FITPACK's REGRID/fpgrre (knot growth + smoothing
+parameter search to meet a target residual `s`) while keeping the control flow
+explicit in Python and returning a 2-D `NdBSpline`.
+
+**Key conventions**
+- **Penalty scaling is 1/p** - *same as FITPACK*. The augmented system stacks
+  `(D / p)` under the data matrix. Smaller `p` -> stronger smoothing; larger `p`
+  -> weaker smoothing (approaching interpolation).
+- `p == -1` is used **as a sentinel for `p = ∞`** (interpolatory limit): when
+  seen, the solver omits penalty rows entirely.
+
+2) Mathematics (and how this differs from FITPACK's REGRID)
+-----------------------------------------------------------
+Let `A_x`, `A_y` be banded 1-D design matrices and `D_x`, `D_y` the banded
+1-D roughness (difference) matrices returned by FITPACK/Dierckx's 1-D APIs
+(`data_matrix`, `disc`). The 2-D smoothing objective is
+
+    minimize  ||A c - z||^2 + (1/p) ||D c||^2,            (1)
+
+implemented by the **augmented** least squares system
+
+    [ A ] c ~ [ z ]
+    [D/p]     [ 0 ].
+
+**Same as FITPACK:** Equation (1) uses the *1/p* convention.
+**Different in this module:** We realize the 2-D problem by composing **1-D**
+banded operators in two separable passes (x then y), instead of calling a
+monolithic 2-D routine. This makes the mathematics transparent while producing
+the same normal equations structure that REGRID targets internally.
+
+3) Residual energy: definition and use
+--------------------------------------
+After solving on the current knots (initially at the interpolatory limit,
+`p = ∞`), we evaluate the surface and form
+
+    R = Z - Zhat,                         fp = sum(R[i, j]^2).
+
+We then **project** residual energy to knot spans along each axis:
+
+- Row energy  `row_energy[i] = sum(R[i, j]^2, j)`  -> accumulated into `fpintx`.
+- Column energy `col_energy[j] = sum(R[i, j]^2, i)` -> accumulated into `fpinty`.
+
+These per-span energies guide **adaptive knot insertion**: the algorithm picks
+high-energy spans (skipping zero-length ones) and inserts data-aligned knots
+(e.g., median sample within the span), with simple batch sizing and headroom
+limits. This is conceptually the same idea as REGRID's `fpint` arrays; here it
+is implemented explicitly and vectorized in Python for clarity.
+
+4) Solver used here vs. FITPACK's REGRID
+----------------------------------------
+**This module (separable 1-D composition):**
+- Uses **1-D FITPACK/Dierckx kernels** (`data_matrix`, `disc`, `qr_reduce`,
+  `fpback`) to solve 2-D via two passes:
+  1) augment/QR/backsolve along **x**, producing an intermediate;
+  2) augment/QR/backsolve along **y**, producing the coefficient grid `C`.
+- The augmented rows are stacked as `[A; D/p]` (1/p scaling, **same as FITPACK**).
+- If `fp > s` after knot growth, performs a **scalar search in `p`** using a
+  ratio-of-roots routine; `p == -1` is treated as `p = ∞` for the interpolatory
+  reference without penalty rows.
+
+**FITPACK REGRID (monolithic 2-D routine):**
+- Implements the same mathematical objective and **1/p** penalty scaling,
+  but inside a specialized, fully 2-D Fortran routine with in-place Givens/QR
+  and a rational update for `p`.
+- Handles residual partitioning, knot insertion, and `p` updates internally.
+
+**Practical difference:** We build the 2-D solve from **1-D building blocks**,
+which keeps each stage observable/testable and uses the same low-level kernels
+as FITPACK, but without relying on the single monolithic REGRID entry point.
+
+5) Execution flow (who calls whom)
+----------------------------------
+
+**Top-level**
+1. `_regrid`
+   - Validates shapes/monotonicity and normalizes `bbox`.
+   - Dispatches to `_regrid_fitpack(...)`.
+
+**Core driver**
+2. `_regrid_fitpack`
+   - Clips inputs to `bbox` via `_apply_bbox_grid` -> `(x_fit, y_fit, Z_fit)`.
+   - If `s == 0` (interpolation path):
+     - Build initial not-a-knot vectors: `tx = _not_a_knot(x_fit, kx)`,
+       `ty = _not_a_knot(y_fit, ky)`.
+     - Build design matrices: `(Ax, Ay, Q) = build_design_matrices(...)`.
+     - Solve once at `p = -1` (inf): `C0, fp = _solve_2d_fitpack(...)`.
+     - Return `return_NdBSpline(...)`.
+   - Else (`s > 0`, smoothing with adaptive knots):
+     - Initialize clamped no-interior-knot vectors with
+       `_initialise_knots` (for x and y).
+     - **Knot-growth loop** (up to `len(x_fit)+len(y_fit)` iterations):
+       1) Build design matrices: `(Ax, Ay, Q) = build_design_matrices(...)`.
+       2) Interpolatory reference solve (`p = -1`): `C0, fp = _solve_2d_fitpack(...)`.
+       3) If both `tx`, `ty` are at minimal size, record `fp0 = fp`
+          (used to bracket the p-search).
+       4) If `fp < s`: **stop growth** and proceed to p-search / finalize.
+       5) Compute residuals for knot insertion:
+          - Convert packed to CSR once per axis for evaluation:
+            `_Ax = Ax.tocsr(...)`, `_Ay = Ay.tocsr(...)`
+          - Evaluate `Z0 = _Ax @ C0 @ _Ay.T`, residual `R = Z_fit - Z0`.
+       6) Insert knots on alternating axes using energy heuristics:
+          - If last axis was `y`, grow `tx` with
+            `_add_knots(..., residuals=np.sum(R**2, axis=1), ...)`.
+          - Else, grow `ty` with `_add_knots(..., residuals=np.sum(R**2, axis=0), ...)`.
+          - Update `fpold`, `nplus{ x|y }`, `last_axis`.
+     - If growth ended with both axes still minimal: return `return_NdBSpline(...)`.
+     - **Finite-p smoothing (if needed):**
+       - Build 1-D penalty operators: `(Drx, _, ncx) = disc(tx, kx)`,
+         `(Dry, _, ncy) = disc(ty, ky)`, wrap as `PackedMatrix`.
+       - Rebuild design matrices on the final `(tx, ty)`.
+       - Run `_p_search_hit_s(...)` to find `p*` such that `fp(p*) ~ s`:
+         - Internally constructs `F(...)` which evaluates `fp(p)`
+           by calling `_solve_2d_fitpack(...)`.
+         - Uses `root_rati` on `g(p) = fp(p) - s` with the interpolatory
+           reference at `p = inf` and `fp0` bracket.
+       - Return `return_NdBSpline(...)`.
+
+**Separable solver (used by both interpolation and p-search)**
+3. `_solve_2d_fitpack`
+   - Forms augmented banded systems via `_stack_augmented_fitpack`:
+     - X-side: `Ax_aug = [Ax; Dx/p]` if `p != -1`, else just `Ax`.
+     - Y-side: `Ay_aug = [Ay; Dy/p]` if `p != -1`, else just `Ay`.
+     - Pads RHS `Q` with zeros when penalties are stacked.
+   - **Stage X**: `_dierckx.qr_reduce(Ax_aug, ...)` then
+     `_dierckx.fpback(...)` -> intermediate `T`.
+   - Transpose `T` to feed Y solves column-wise; pad if needed.
+   - **Stage Y**: `_dierckx.qr_reduce(Ay_aug, ...)` then
+     `_dierckx.fpback(...)` -> coefficients `C`.
+   - Build CSR design matrices once: `_Ax = Ax.tocsr(...)`, `_Ay = Ay.tocsr(...)`.
+   - Evaluate `zhat = _Ax @ C.T @ _Ay.T`, compute `fp = ||Z_fit - zhat||^2`.
+   - Return `C.T` (in `(nx_coef, ny_coef)` layout) and `fp`.
+
+**Helpers**
+- `_apply_bbox_grid(...)` - slices to `(x_fit, y_fit, Z_fit)`.
+- `build_design_matrices(...)` - wraps `_dierckx.data_matrix` and
+  returns `PackedMatrix` wrappers + `Q`.
+- `_initialise_knots(...)`, `_add_knots(...)` - FITPACK-style knot bookkeeping/growth.
+- `disc(...)` - 1-D roughness operators (packed band form).
+- `F` + `root_rati` - maps `p -> fp(p)` and finds `p*` with `fp(p*) ~ s`.
+- `return_NdBSpline(...)` - packs `(tx, ty, C)` into an `NdBSpline`.
+"""
+import numpy as np
+from scipy.interpolate._ndbspline import NdBSpline
+from scipy.interpolate._fitpack_repro import (
+    root_rati, disc, add_knot, _not_a_knot)
+from . import _dierckx
+from scipy.sparse import csr_array
+from scipy._lib._util import _validate_int
+
+def _ndbspline_call_like_bivariate(ndbs, x, y, dx=0, dy=0, grid=True):
+    """
+    Evaluate a 2D `NdBSpline` like a classical bivariate API.
+
+    Parameters
+    ----------
+    ndbs : NdBSpline
+        A 2D spline object (``len(ndbs.t) == 2``).
+    x, y : array_like
+        Sample locations. If ``grid=True``, these must be 1-D strictly
+        increasing vectors. If ``grid=False``, they can be broadcastable
+        arrays of the same shape.
+    dx, dy : int, optional
+        Derivative orders along `x` and `y` respectively, by default 0.
+    grid : bool, optional
+        If True, evaluate on the cartesian product of `x` and `y`;
+        otherwise treat `(x, y)` as paired coordinates, by default True.
+
+    Returns
+    -------
+    ndarray or (ndarray, dict)
+        Evaluated values with shape:
+        - ``(len(x), len(y), ...)`` if ``grid=True``.
+        - ``x.shape + ...`` if ``grid=False``.
+
+    Raises
+    ------
+    ValueError
+        If `ndbs` is not 2D, derivatives are negative, or monotonicity checks fail.
+
+    Notes
+    -----
+    This is a thin convenience wrapper around ``NdBSpline.__call__`` with input
+    validation and optional profiling.
+    """
+    if len(ndbs.t) != 2:
+        raise ValueError("ndbs must be a 2D NdBSpline (len(t) == 2).")
+
+    dx = _validate_int(dx, 'dx')
+    dy = _validate_int(dy, 'dy')
+    if dx < 0 or dy < 0:
+        raise ValueError("order of derivative must be positive or zero")
+
+    trailing = ndbs.c.shape[2:]
+    x = np.asarray(x)
+    y = np.asarray(y)
+
+    if grid:
+        if x.size == 0 or y.size == 0:
+            vals = np.zeros((x.size, y.size) + trailing, dtype=ndbs.c.dtype)
+            return vals
+
+        if (x.size >= 2) and (not np.all(np.diff(x) >= 0.0)):
+            raise ValueError("x must be strictly increasing when `grid` is True")
+        if (y.size >= 2) and (not np.all(np.diff(y) >= 0.0)):
+            raise ValueError("y must be strictly increasing when `grid` is True")
+
+        X, Y = np.meshgrid(x, y, indexing="ij")
+        xi = np.stack((X, Y), axis=-1)  # (len(x), len(y), 2)
+
+        vals = ndbs(xi, nu=(dx, dy), extrapolate=ndbs.extrapolate)
+
+        return vals
+    else:
+        if x.shape != y.shape:
+            x, y = np.broadcast_arrays(x, y)
+
+        if x.size == 0:
+            return np.zeros(x.shape + trailing, dtype=ndbs.c.dtype)
+        xi = np.stack((x.ravel(), y.ravel()), axis=-1)
+        vals = ndbs(xi, nu=(dx, dy), extrapolate=ndbs.extrapolate)
+        return vals.reshape(x.shape + trailing)
+
+
+def return_NdBSpline(fp, tck, degrees):
+    """
+    Build a 2D ``NdBSpline`` from knot vectors and a coefficient grid.
+
+    Parameters
+    ----------
+    fp : float
+        Residual sum of squares of the produced fit (kept for upstream use).
+    tck : tuple
+        Tuple ``(tx, ty, C)`` where ``tx``, ``ty`` are knot vectors and ``C``
+        is a coefficient array with shape ``(nx - kx - 1, ny - ky - 1)`` or
+        a compatible shape that can be reshaped to that.
+    degrees : tuple of int
+        Degrees ``(kx, ky)`` along x and y.
+
+    Returns
+    -------
+    NdBSpline
+        The constructed 2D spline.
+
+    Notes
+    -----
+    Only repacks the coefficient grid; ``fp`` is not used internally here.
+    """
+    nx, ny = len(tck[0]), len(tck[1])
+    kx, ky = degrees
+    c = tck[2].reshape(nx - kx - 1, ny - ky - 1)
+    return NdBSpline((tck[0], tck[1]), c, degrees)
+
+
+class PackedMatrix:
+    """A simplified CSR format for when non-zeros in each row are consecutive.
+
+    Assuming that each row of an `(m, nc)` matrix 1) only has `nz` non-zeros, and
+    2) these non-zeros are consecutive, we only store an `(m, nz)` matrix of
+    non-zeros and a 1D array of row offsets. This way, a row `i` of the original
+    matrix A is ``A[i, offset[i]: offset[i] + nz]``.
+
+    """
+    def __init__(self, a, offset, nc):
+        self.a = a
+        self.offset = offset
+        self.nc = nc
+
+    @property
+    def shape(self):
+        return self.a.shape[0], self.nc
+
+    def todense(self):
+        out = np.zeros(self.shape)
+        nelem = self.a.shape[1]
+        for i in range(out.shape[0]):
+            nel = min(self.nc - self.offset[i], nelem)
+            out[i, self.offset[i]:self.offset[i] + nel] = self.a[i, :nel]
+        return out
+
+    def tocsr(self, k, m, len_t):
+        # Inlined from https://github.com/scipy/scipy/blob/v1.16.2/scipy/interpolate/_bsplines.py#L455-L466
+        data = self.a.ravel()
+
+        # Convert from per-row offsets to the CSR indices/indptr format
+        indices = np.repeat(self.offset, k+1).reshape(-1, k+1)
+        indices = indices + np.arange(k+1, dtype=self.offset.dtype)
+        indices = indices.ravel()
+
+        indptr = np.arange(0, (m + 1) * (k + 1), k + 1,
+                           dtype=self.offset.dtype)
+
+        return csr_array(
+            (data, indices, indptr),
+            shape=(m, len_t - k - 1)
+    )
+
+
+def _stack_augmented_fitpack(A, D, nc, k, p):
+    """
+    Builds augmented banded matrix.
+
+    Parameters
+    ----------
+    A : PackedMatrix
+        Banded data/design matrix for one axis (from `_dierckx.data_matrix`).
+    D : PackedMatrix
+        Banded roughness (difference) penalty matrix for the same axis
+        (from `disc`).
+    nc : int
+        Number of top (data) rows from `A` to include.
+    k : int
+        Spline degree (used only for sizing in the current implementation).
+    p : float
+        Smoothing parameter. The effective penalization term is scaled as **1/p**:
+        larger `p` means *less* smoothing (approaching interpolation).
+        If `p == -1`, it signals *p -> inf*, i.e. a pure interpolatory system
+        with **no** penalty rows appended.
+
+    Returns
+    -------
+    AA : ndarray
+        Augmented banded matrix with `A` stacked over `(D / p)` when `p != -1`.
+    offset : ndarray
+        Concatenated band offsets for the augmented matrix.
+    nc : int
+        Returned unchanged for downstream convenience.
+    """
+    if p == -1:
+        return A.a.copy(), A.offset.copy(), nc
+
+    nz = k + 1
+    AA = np.zeros((nc + D.shape[0], k + 2), dtype=float)
+    AA[:nc, :nz] = A.a[:nc, :]
+    AA[nc:, :] = D.a / p
+    offset = np.concatenate((A.offset, D.offset))
+    return AA, offset, nc
+
+def _solve_2d_fitpack(Ax, Ay, Q, p,
+                      kx, tx, x_x,
+                      ky, ty, x_y, z,
+                      Dx=None, Dy=None):
+    """
+    Solve the 2-D tensor-product spline system using separable banded QR.
+
+    ================================================================
+    Mathematical model (step by step, plain text)
+    ================================================================
+
+    Shapes:
+        Z      : (mx, my)  -> original data
+        Ax, Ay : design matrices for x and y
+        Dx, Dy : roughness penalty matrices for x and y
+        C      : (nx, ny)  -> spline coefficients to solve for
+
+    Surface approximation:
+        Zhat = Ax * C * Ay^T
+
+    Objective (smoothing formulation):
+        minimize ||Ax*C*Ay^T - Z||^2 + (1/p)*(||Dx*C||^2 + ||C*Dy^T||^2)
+
+    In practice (FITPACK-style separable approach), we solve this in two stages:
+
+    --------------------------------------------------------
+    Stage 1 (x-direction solve for all y-columns together):
+    --------------------------------------------------------
+
+        For each column of Z:
+            minimize ||Ax*T - Z||^2 + (1/p)*||Dx*T||^2
+
+        This is equivalent to the augmented least-squares system:
+            [Ax]       [Z]
+            [Dx/p] * T = [0]
+
+        i.e.  minimize || [Ax; Dx/p]*T - [Z; 0] ||^2
+
+        The solution T is obtained by QR reduction and back-substitution.
+
+    --------------------------------------------------------
+    Stage 2 (y-direction solve using transposed result):
+    --------------------------------------------------------
+        Now treat T^T as the new RHS for the y-direction:
+            minimize ||Ay*C^T - T^T||^2 + (1/p)*||Dy*C^T||^2
+
+        Equivalent to augmented system:
+            [Ay]       [T^T]
+            [Dy/p] * C^T = [0]
+
+        i.e.  minimize || [Ay; Dy/p]*C^T - [T^T; 0] ||^2
+
+        Solving this gives C^T (then transposed back to C).
+
+    --------------------------------------------------------
+    Interpolation limit:
+    --------------------------------------------------------
+        If p == -1, penalties are omitted (Dx, Dy are not stacked).
+        The solver behaves as a near-interpolating system.
+
+    --------------------------------------------------------
+    Residual computation:
+    --------------------------------------------------------
+        Zhat = Ax * C * Ay^T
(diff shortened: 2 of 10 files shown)


scipy/interpolate/_regrid_python.py
```diff
@@ -0,0 +1,1110 @@
+#!/usr/bin/env python3
```
j-bowhay:
```suggestion
```

scipy/interpolate/__init__.py
```diff
@@ -142,6 +142,7 @@
    :toctree: generated/
 
    RectBivariateSpline
+   regrid_python
```
j-bowhay: I'm not sure we should include "python" in the name of a public api

scipy/interpolate/tests/test_rgi.py
```diff
@@ -873,6 +881,14 @@ def test_spline_2d_outofbounds(self):
         expected[2:4] = 999.99
         assert_array_almost_equal(actual, expected)
 
+        expected_custom = ndbspline_call_like_bivariate(
+            lut_custom, xi[:, 0], xi[:, 1], grid=False)
+        expected_custom[2:4] = 999.99
+        assert_array_almost_equal(actual, expected_custom)
```
j-bowhay: New tests should use `assert_allclose` as per the NumPy docs

scipy/interpolate/_regrid_python.py
```diff
@@ -0,0 +1,1110 @@
...
+
+    """
+    def __init__(self, a, offset, nc):
+        self.a = a
+        self.offset = offset
+        self.nc = nc
+
+        assert a.ndim == 2
+        assert offset.ndim == 1
+        assert a.shape[0] == offset.shape[0]
```
j-bowhay: Plain asserts shouldn't be used in user facing code, if these are important they should be proper exceptions

j-bowhay (commented):
Some drive-by comments

czgdp1807:
@j-bowhay Thanks for the reviews. I am waiting for more reviews. Then I will address them collectively.

scipy/interpolate/__init__.py
czgdp1807: Correct. Let's rename to just `regrid`? Or `rect_bivariate_spline`? What do you suggest @ev-br @rgommers?

rgommers: `regrid` is too general I'd think. I haven't followed along very closely, so I can't say if `rect_bivariate_spline` is an appropriate name here or not.

ev-br: >  if `rect_bivariate_spline` is an appropriate name here or not.

For this function, it is.

`regrid` is the name of the FITPACK routine (and the scattered analog is `surfit`) 

WHat we probably should think of at this stage is how it generalizes. For 2D FITPACK splines, we currently have (https://docs.scipy.org/doc/scipy/reference/interpolate.html#id1):

RectBivariateSpline    : this PR is a replacement
RectSphereBivariateSpline

SmoothBivariateSpline
SmoothSphereBivariateSpline

The `Smooth*` variants are for the scattered interpolation, as opposed gridded. The scattered non-sphere version are also available as `bisplrep`. 
There are also `LSQ*` counterparts for `Smooth*`, but those I belive can be folded together into a single function/class.

So, we can have `rect_bivariate_spline` and `scattered_bivariate_spline`, but then the spherical versions are awkward: `rect_sphere_bivariate_spline`?

ev-br:
I believe the algorithmic content of this PR is a faithful rewrite of the FITPACK logic behind `RectBivariateSpline`. The knot selection logic is the same, and the linear algebra is equivalent. The main difference is that this PR is modular and IMO way more readable than the Fortran original.

scipy/interpolate/__init__.py
```diff
@@ -209,6 +210,8 @@
 
 from ._pade import *
 
+from ._regrid_python import regrid_python, ndbspline_call_like_bivariate
```
j-bowhay: It is not clear if `ndbspline_call_like_bivariate` is meant to be public as it hasn't been added to the toc tree but has been imported here

scipy/interpolate/_regrid_python.py
```diff
@@ -0,0 +1,1110 @@
...
+import numpy as np
+from scipy.interpolate._ndbspline import NdBSpline
+from scipy.interpolate._fitpack_repro import (
+    root_rati, disc, add_knot, _not_a_knot)
+from . import _dierckx
+from scipy.sparse import csr_array
+
+def ndbspline_call_like_bivariate(ndbs, x, y, dx=0, dy=0, grid=True):
+    """
+    Evaluate a 2D ``NdBSpline`` like a classical bivariate API.
```
j-bowhay:
```suggestion
    Evaluate a 2D `NdBSpline` like a classical bivariate API.
```

scipy/interpolate/_regrid_python.py
```diff
@@ -0,0 +1,1110 @@
...
+    Parameters
+    ----------
+    ndbs : NdBSpline
+        A 2D spline object (``len(ndbs.t) == 2``).
+    x, y : array_like
+        Sample locations. If ``grid=True``, these must be 1-D strictly
+        increasing vectors. If ``grid=False``, they can be broadcastable
+        arrays of the same shape.
+    dx, dy : int, optional
+        Derivative orders along x and y respectively, by default 0.
```
j-bowhay:
```suggestion
        Derivative orders along `x` and `y` respectively, by default 0.
```

scipy/interpolate/_regrid_python.py
```diff
@@ -0,0 +1,1110 @@
...
+    grid : bool, optional
+        If True, evaluate on the cartesian product of ``x`` and ``y``;
+        otherwise treat ``(x, y)`` as paired coordinates, by default True.
```
j-bowhay:
```suggestion
        If True, evaluate on the cartesian product of `x` and `y`;
        otherwise treat ``(x, y)`` as paired coordinates, by default True.
```

scipy/interpolate/_regrid_python.py
```diff
@@ -0,0 +1,1110 @@
...
+    -------
+    ndarray or (ndarray, dict)
+        Evaluated values with shape:
+        - ``(len(x), len(y), ...)`` if ``grid=True``.
+        - ``x.shape + ...`` if ``grid=False``.
+
+    Raises
+    ------
+    ValueError
+        If ``ndbs`` is not 2D, derivatives are negative, or monotonicity checks fail.
```
j-bowhay:
```suggestion
        If `ndbs` is not 2D, derivatives are negative, or monotonicity checks fail.
```

scipy/interpolate/_regrid_python.py
```diff
@@ -0,0 +1,1110 @@
...
+    if not (isinstance(dx, int | np.integer) and isinstance(dy, int | np.integer)):
+        raise ValueError("dx and dy must be integers.")
+    if dx < 0 or dy < 0:
+        raise ValueError("order of derivative must be positive or zero")
+
+    trailing = ndbs.c.shape[2:]
+
+    if grid:
+        x = np.asarray(x)
+        y = np.asarray(y)
```
j-bowhay: Can this be moved out of the if statement since it is the first thing that appears in both branches

scipy/interpolate/_regrid_python.py
```diff
@@ -0,0 +1,1110 @@
...
+        If ``ndbs`` is not 2D, derivatives are negative, or monotonicity checks fail.
+
+    Notes
+    -----
+    This is a thin convenience wrapper around ``NdBSpline.__call__`` with input
+    validation and optional profiling.
+    """
+    if len(ndbs.t) != 2:
+        raise ValueError("ndbs must be a 2D NdBSpline (len(t) == 2).")
+    if not (isinstance(dx, int | np.integer) and isinstance(dy, int | np.integer)):
```
j-bowhay: https://github.com/scipy/scipy/blob/602e4427f6738c7209dc63b21da7d2057b0d444a/scipy/_lib/_util.py#L459

scipy/interpolate/_regrid.py
```diff
@@ -0,0 +1,1110 @@
...
+
+        if x.shape != y.shape:
+            x, y = np.broadcast_arrays(x, y)
+
+        if x.size == 0:
+            return np.zeros(x.shape + trailing, dtype=ndbs.c.dtype)
+        xi = np.stack((x.ravel(), y.ravel()), axis=-1)
+        vals = ndbs(xi, nu=(dx, dy), extrapolate=ndbs.extrapolate)
+        return vals.reshape(x.shape + trailing)
+
```
j-bowhay, line 230:
```suggestion


```

scipy/interpolate/_regrid.py
```diff
@@ -0,0 +1,1110 @@
...
+    Notes
+    -----
+    Only repacks the coefficient grid; ``fp`` is not used internally here.
+    """
+    nx, ny = len(tck[0]), len(tck[1])
+    kx, ky = degrees
+    c = tck[2].reshape(nx - kx - 1, ny - ky - 1)
+    return NdBSpline((tck[0], tck[1]), c, degrees)
+
```
j-bowhay, line 260:
```suggestion


```

scipy/interpolate/_regrid.py
```diff
@@ -0,0 +1,1110 @@
...
+        indices = indices.ravel()
+
+        indptr = np.arange(0, (m + 1) * (k + 1), k + 1,
+                           dtype=self.offset.dtype)
+
+        return csr_array(
+            (data, indices, indptr),
+            shape=(m, len_t - k - 1)
+    )
+
```
j-bowhay, line 304:
```suggestion


```

scipy/interpolate/_regrid_python.py
```diff
@@ -0,0 +1,1110 @@
...
+        Returned unchanged for downstream convenience.
+    """
+    if p == -1:
+        return A.a.copy(), A.offset.copy(), nc
+
+    nz = k + 1
+    AA = np.zeros((nc + D.shape[0], k + 2), dtype=float)
+    AA[:nc, :nz] = A.a[:nc, :]
+    AA[nc:, :] = D.a / p
+    offset = np.r_[A.offset, D.offset]
```
j-bowhay: Can we use the `np.concatinate` rather than this historical heirloom

scipy/interpolate/_regrid.py
```diff
@@ -0,0 +1,1110 @@
...
+    bracket = ((0.0, fp0 - s), (np.inf, fpms))
+    ftol = max(s * tol_rel, 1e-12)
+
+    r = root_rati(g, p_init, bracket, ftol, maxit=maxit)
+    p_star = r.root
+    fp_star = fp_at(p_star)
+    C_star = fp_at.C
+
+    return p_star, C_star, fp_star
+
```
j-bowhay, line 684:
```suggestion


```

scipy/interpolate/_regrid.py
```diff
@@ -0,0 +1,1110 @@
...
+    if not (xb < xe and yb < ye):
+        raise ValueError("bbox must satisfy xb < xe and yb < ye")
+
+    ix = np.where((x >= xb) & (x <= xe))[0]
+    iy = np.where((y >= yb) & (y <= ye))[0]
+    if ix.size == 0 or iy.size == 0:
+        raise ValueError("bbox excludes all samples in x or y.")
+
+    return x[ix], y[iy], Z[np.ix_(ix, iy)], np.s_[ix], np.s_[iy]
+
```
j-bowhay, line 724:
```suggestion


```

scipy/interpolate/_regrid_python.py
```diff
@@ -0,0 +1,1110 @@
...
+
+    Ax, offset_x, nc_x = _dierckx.data_matrix(x, tx, kx, w_x)
+    Ay, offset_y, nc_y = _dierckx.data_matrix(y, ty, ky, w_y)
+    Q = z.copy()
+
+    return (PackedMatrix(Ax, offset_x, nc_x),
+            PackedMatrix(Ay, offset_y, nc_y),
+            Q)
+
+TOL = 0.001
```
j-bowhay: Can this be moved closer to where it is used

scipy/interpolate/_regrid.py
```diff
@@ -0,0 +1,1110 @@
...
+        # c  if n=nest we cannot increase the number of knots because of
+        # c  the storage capacity limitation.
+        # https://github.com/scipy/scipy/blob/v1.16.2/scipy/interpolate/fitpack/fpregr.f#L280
+        # https://github.com/scipy/scipy/blob/v1.16.2/scipy/interpolate/fitpack/fpregr.f#L294
+        if n >= nest:
+            return t, nplus
+
+    return t, nplus
+
```
j-bowhay, line 910:
```suggestion


```

scipy/interpolate/_regrid.py
```diff
@@ -0,0 +1,1110 @@
...
+            raise ValueError(f"`nest` too small: {nest = } < 2*(k+1) = {2*(k+1)}.")
+
+    nmin = 2*(k + 1)    # the number of knots for an LSQ polynomial approximation
+    nmax = m + k + 1  # the number of knots for the spline interpolation
+
+    # start from no internal knots
+    t = np.asarray([xb]*(k+1) + [xe]*(k+1))
+
+    return t, nest, nmin, nmax
+
```
j-bowhay, line 789:
```suggestion


```

scipy/interpolate/_regrid_python.py
```diff
@@ -0,0 +1,1110 @@
...
+    Public interface for 2-D smoothing B-spline fitting (1/p penalty form).
+
+    Parameters
+    ----------
+    x, y : array_like
+        Strictly increasing 1-D coordinate vectors.
+    z : array_like, shape (len(x), len(y))
+        Data grid.
+    bbox : sequence of 4 scalars
+        Optional bounding box `(xb, xe, yb, ye)`; use `None` entries to disable.
```
j-bowhay:
```suggestion
        Optional bounding box ``(xb, xe, yb, ye)``; use ``None`` entries to disable.
```

scipy/interpolate/_regrid_python.py
```diff
@@ -0,0 +1,1110 @@
...
+    (Ax, Ay, Q) = _build_design_matrices(
+        x_fit, y_fit, Z, tx, ty, kx, ky)
+    _, C_sm, fp_sm = _p_search_hit_s(Ax, Drx, Ay, Dry, Q,
+                                     kx, tx, x_fit, ky,
+                                     ty, y_fit, Z_fit, s,
+                                     fp0, maxit=maxit, p_init=p)
+    return return_NdBSpline(fp_sm, (tx, ty, C_sm), (kx, ky))
+
+
+def regrid_python(x, y, z, *, bbox=[None]*4, kx=3, ky=3, s=0.0, maxit=50):
```
j-bowhay: Mutable objects should generally be avoided as default arguments

scipy/interpolate/_regrid_python.py
```diff
@@ -0,0 +1,1110 @@
...
+    Parameters
+    ----------
+    x, y : array_like
+        Strictly increasing 1-D coordinate vectors.
+    z : array_like, shape (len(x), len(y))
+        Data grid.
+    bbox : sequence of 4 scalars
+        Optional bounding box `(xb, xe, yb, ye)`; use `None` entries to disable.
+    kx, ky : int, optional
+        Spline degrees along x and y, default cubic (3).
```
j-bowhay:
```suggestion
        Spline degrees along `x` and `y`, default cubic (3).
```

scipy/interpolate/_regrid_python.py
```diff
@@ -0,0 +1,1110 @@
...
+    s : float, optional
+        Target smoothing residual (`fp` target). Must satisfy `s >= 0`.
```
j-bowhay: What is fp?

j-bowhay:
```suggestion
        Target smoothing residual (`fp` target). Must satisfy ``s >= 0``.
```

scipy/interpolate/_regrid_python.py
```diff
@@ -0,0 +1,1110 @@
...
+
+    Returns
+    -------
+    NdBSpline
+        Fitted bivariate spline surface.
+
+    Notes
+    -----
+    This validates input, enforces monotonicity, and calls
+    `_regrid_python_fitpack`.  The fit obeys the 1/p-penalty convention used
```
j-bowhay: This is an implementation detail so should not be in the public docs

scipy/interpolate/_regrid.py
```diff
@@ -0,0 +1,1110 @@
...
+    NdBSpline
+        Fitted bivariate spline surface.
+
+    Notes
+    -----
+    This validates input, enforces monotonicity, and calls
+    `_regrid_python_fitpack`.  The fit obeys the 1/p-penalty convention used
+    throughout this module, and `p == -1` is treated as the interpolatory
+    (infinite-p) case.
+    """
```
j-bowhay, line 1085: New functions should have an example section as per #7168

scipy/interpolate/_regrid.py
```diff
@@ -0,0 +1,1110 @@
...
+
+    x = np.asarray(x, dtype=float)
+    y = np.asarray(y, dtype=float)
+    z = np.asarray(z, float)
+    bbox = np.ravel(bbox)
```
j-bowhay, line 1092: Why do we need to ravel this input?

scipy/interpolate/_regrid_python.py
```diff
@@ -0,0 +1,1110 @@
...
+
+    if not np.all(np.diff(x) > 0.0):
+        raise ValueError("x must be strictly increasing")
+    if not np.all(np.diff(y) > 0.0):
+        raise ValueError("y must be strictly increasing")
+    if x.size != z.shape[0]:
+        raise ValueError("x dimension of z must have same number of elements as x")
+    if y.size != z.shape[1]:
+        raise ValueError("y dimension of z must have same number of elements as y")
+    if s is not None and not (s >= 0.0):
```
j-bowhay: Why is the None check needed?

scipy/interpolate/tests/test_fitpack2.py
```diff
@@ -924,18 +925,90 @@ def test_array_like_input(self):
...
 class TestRectBivariateSpline:
-    def test_defaults(self):
+    @pytest.mark.parametrize("k", [3, 4])
+    def test_defaults(self, k):
         x = array([1,2,3,4,5])
         y = array([1,2,3,4,5])
         z = array([[1,2,1,2,1],[1,2,1,2,1],[1,2,3,2,1],[1,2,2,2,1],[1,2,1,2,1]])
-        lut = RectBivariateSpline(x,y,z)
+
+        lut = RectBivariateSpline(x,y,z,kx=k,ky=k)
```
j-bowhay: Please follow PEP8 recommendations on whitespace

scipy/interpolate/tests/test_fitpack2.py
```diff
@@ -924,18 +925,90 @@ def test_array_like_input(self):
...
+        custom = regrid_python(x, y, z, kx=k, ky=k)
+        custom_vals = ndbspline_call_like_bivariate(custom, xi, yi)
+        # Compare interpolated values
+        assert_array_almost_equal(custom_vals, ref_vals)
+
+    @pytest.mark.parametrize("k", [3, 4])
+    def test_midpoints(self, k):
+        # Midpoint interpolation
+        x = np.array([i for i in range(1, k + 2)])
+        y = np.array([i for i in range(1, k + 2)])
```
j-bowhay: `np.arange`?

scipy/interpolate/tests/test_fitpack2.py
```diff
@@ -1070,73 +1192,129 @@ def test_partial_derivative_order_too_large(self):
 
         assert (lut_ndbspline.derivative([4, 1]).c == 0.0).all()
 
-    def test_broadcast(self):
+        lut = RectBivariateSpline(x, y, z, kx=k, ky=k)
+        with assert_raises(ValueError):
+            lut.partial_derivative(4, 1)
+
+        lut_custom = regrid_python(x, y, z, kx=k, ky=k)
+        # TODO: Should be exactly equal to 0.0, bug in NdBSpline.derivative
```
j-bowhay: Is this a blocking todo?

scipy/interpolate/tests/test_fitpack2.py
```diff
@@ -1070,73 +1192,129 @@ def test_partial_derivative_order_too_large(self):
...
+        lut = RectBivariateSpline(x,y,z,kx=k,ky=k)
         xp_assert_close(lut(x, y), lut(x[:,None], y[None,:], grid=False))
 
+        # Broadcasting doesn't make sense with the interface of NdBSpline
+        # lut_custom = regrid_python(x,y,z,kx=k,ky=k)
+        # xp_assert_close(
+        #     ndbspline_call_like_bivariate(lut_custom, x, y),
+        #     ndbspline_call_like_bivariate(lut_custom, [
+        #         (xp, yp) for xp, yp in zip(x[:,None], y[None,:])])
+        # )
```
j-bowhay: remove?

j-bowhay (commented):
Thanks again, quite an undertaking! There are a few places where this still feels a little rough around the edges so it would be nice for things to be polished up a bit before merging

scipy/interpolate/__init__.py
czgdp1807: > For this function, it is.

I see. So we can go with `rect_bivariate_spline`?



> but then the spherical versions are awkward: `rect_sphere_bivariate_spline`?

Any specific reason for not using `rect_sphere_bivariate_spline` for the replacement of `RectBivariateSpline`?

scipy/interpolate/_regrid_python.py
czgdp1807: I am using the same API as `RectBivariateSpline`. See, 

https://github.com/scipy/scipy/blob/27e4a454db8a476c12e7ef93d39b6225d73fd38e/scipy/interpolate/_fitpack2.py#L1620

Should I change `bbox=None` and then handle it inside the function body? @ev-br What do you say?

scipy/interpolate/_regrid_python.py
czgdp1807: `fp` is the residual. That's why written `Target smoothing residual` and `fp` target. It's the residual. Do you want me to make it more explicit?

scipy/interpolate/_regrid.py
czgdp1807: No strong reasons. Just doing `RectBivariateSpline` is doing,

https://github.com/scipy/scipy/blob/27e4a454db8a476c12e7ef93d39b6225d73fd38e/scipy/interpolate/_fitpack2.py#L1621

scipy/interpolate/_regrid_python.py
czgdp1807: No strong reason. Taken from,

https://github.com/scipy/scipy/blob/27e4a454db8a476c12e7ef93d39b6225d73fd38e/scipy/interpolate/_fitpack2.py#L1635-L1636

scipy/interpolate/tests/test_fitpack2.py
czgdp1807: No not a blocking TODO. Its an issue with `NdBSpline.derivative`. Shouldn't block this PR.

czgdp1807:
@j-bowhay I have addressed your reviews and responded to your questions.

Had to resolve those conversations which don't need further discussion (the page was getting heavy to load otherwise).

scipy/interpolate/tests/test_fitpack2.py
ev-br: Is it https://github.com/scipy/scipy/issues/23510 or something else? Would be nice to fix, in a separate PR

scipy/interpolate/__init__.py
ev-br: User-facing interface might need some more thought. In my (somewhat rough) ideas, I think we should have a reasonable new function and a documentation guide for porting from RectBivariateSpline to the new framework. 
Then `ndbspline_call_like_bivariate` is needed internally for tests and its simplified version ends up in the docs as a recipe.

However, if we want to actually remove the Fortran implementations, we likely also want to replace RectBivariateSpline implementations with the new ones: keep the API exactly the same, and replace the internals (cf the ODR thread on Discourse).

By all means let's discuss the strategy (and I'm all ears for suggestions!)

However no matter how we play it, I don't see `ndbspline_call_like_bivariate` becoming a part of public SciPy API, really.

scipy/interpolate/_regrid_python.py
ev-br: I'd suggest to keep the default to `s=0`, document it as a float, call `s = float(s)` once  and forget about None.

scipy/interpolate/_regrid_python.py
ev-br: +1 for  `if bbox is None: bbox = [None]*4`.

scipy/interpolate/_regrid_python.py
```diff
@@ -0,0 +1,1104 @@
...
+
+    Returns
+    -------
+    NdBSpline
+        Fitted bivariate spline surface.
+    """
+
+    x = np.asarray(x, dtype=float)
+    y = np.asarray(y, dtype=float)
+    z = np.asarray(z, float)
```
ev-br: Nit
```suggestion
    z = np.asarray(z, dtype=float)
```

scipy/interpolate/tests/test_fitpack2.py
czgdp1807: I think so. I think `NdBSpline` keeps evaluating derivatives. A minor check like if derivative order is higher than the degree of the polynomial, then answer should be 0 polynomial directly. 

Can I take up #23510 in a separate PR?

scipy/interpolate/__init__.py
czgdp1807: > I don't see `ndbspline_call_like_bivariate` becoming a part of public SciPy API, really.

Makes sense. Will remove it from the public API.



> In my (somewhat rough) ideas, I think we should have a reasonable new function and a documentation guide for porting from RectBivariateSpline to the new framework.

Makes sense. What about adding an option in `RectBivariateSpline.__init__(..., use_fitpack=True)`. When we pass `RectBivariateSpline(..., use_fitpack=False)` then our `_regrid_python.regrid_python` API will be called. Also, in `RectBivariateSpline.__call__` we will check whether the object is `NdBSpline`. If so then we would call `ndbspline_call_like_bivariate`. So nothing will be exposed to the user. :-)

@ev-br What do you say?

ev-br: Sorry for the delay @czgdp1807 . Please no `use_fitpack={True,False}`. 
Let's add a constructor creating an NdBSpline instance + see what it takes to make `RectBivariateSpline` use it under the hood (add a `self._ndbspline` attribute, and use it instead of FITPACK calls).

czgdp1807: I see. Then I think for `self._ndbspline` we would call our `_regrid_python.regrid_python`.

> add a `self._ndbspline` attribute, and use it instead of FITPACK calls

So we are planning to replace the FITPACK calls with `_regrid_python` logic via `RectBivariateSpline._ndbspline`. Am I right? It would be a big change, API wise it would still be the same, however the implementation would be entirely replaced. Please let me know that's what we are looking at.

ev-br: Yes, it's a bit of a change. We now realized we want to get rid of Fortran in the 1.18 timeframe. Therefore, I'd split the work in two steps (two and a half, in fact) :
1. make `regrid_python` logic private, concentrate on getting it across the finish line
2. replace FITPACK calls with `self._ndbspline`
3. consider exposing `regrid_python logic` as a public function 

Steps 2 and 3 can be done in a follow-up PR.

czgdp1807: > 1. make `regrid_python` logic private, concentrate on getting it across the finish line

I see. `regrid_python` is already in a private module `_regrid_python`. Its not exposed in SciPy's public API. 

There aren't any pending reviews as of now. Please let me know if something needs to be addressed for merging this PR. Thank you.

czgdp1807:
To merge this PR, I think following 2 needs decision,

- [x] https://github.com/scipy/scipy/pull/23962#discussion_r2569807093 - I would privatise the module completely by removing `regrid_python` function from the public namespace of `scipy.interpolate`. If one wants to use `regrid_python`, they will have to do, `from scipy.interpolate._regrid_python import regrid_python` - https://github.com/scipy/scipy/pull/23962/commits/59d81369ec501dd27ae0b394536e980469f7d495
- [ ] https://github.com/scipy/scipy/pull/23962#discussion_r2526911060 - Doesn't seem necessary currently, since we will be privatising the `regrid_python` function. However, I think going for `rect_bivariate_spline` name seems okay for now, since we will not be exposing it directly in `scipy.interpolate` namespace. Please let me know if you want to rename before we merge this PR.

cc: @ev-br

czgdp1807:
@ev-br Any plans to merge this?

> Hi. Considering, SciPy main branch is Fortran free, do we still plan to merge - https://github.com/scipy/scipy/pull/23962?

ilayn:
> Hi. Considering, SciPy main branch is Fortran free, do we still plan to merge - https://github.com/scipy/scipy/pull/23962?

Definitely, please don't take the translation work as a competitor against this work, I just removed the Fortran-ness of the codebase. The FITPACK code still reads like a tongue-twister independent of the implementation language. Your work and anything that modernizes/cleans up FITPACK code is much appreciated.

czgdp1807:
Sure. I will resolve the conflicts here, re-run the benchmarks with the latest `main` branch. Let's merge it soon. It has been complete since 1 month.

cc: @ev-br

j-bowhay (changes requested):
I am not sure this is ready to merge, it still isn't clear what is public interface and what isn't

czgdp1807:
As far as I remember, `regrid_python` is present as a private API. We intend to replace the existing implementation used under the hood in `RectBivariateSpline` with the one I added. Currently it is not exposed. Would ask @ev-br to confirm.

j-bowhay:
> As far as I remember, `regrid_python` is present as a private API. We intend to replace the existing implementation used under the hood in `RectBivariateSpline` with the one I added. Currently it is not exposed. Would ask @ev-br to confirm.

That makes sense, can you update the pr description as that was the source of my confusion.

czgdp1807:
Updated the PR description, benchmarks. Rebased on top of `main` branch. This is good to go.

@ev-br @j-bowhay

czgdp1807:
Heads up: I’ve posted a community update on the new Python-based REGRID implementation (regrid_python) here: https://discuss.scientific-python.org/t/enh-interpolate-python-bivariate-spline-fitter-regrid-python/2219. It summarizes the design, FITPACK compatibility, knot-growth strategy, smoothing search, and usage notes. Feedback from SciPy users is welcome -- please add thoughts there or in this PR.

j-bowhay (dismissed):
Some of these tests are prohibitively slow, on my machine
```
 pytest scipy/interpolate/tests/test_fitpack2.py
meson-python: building scipy: /home/up19056/.local/share/mamba/envs/scipy-dev/bin/ninja
[2/356] Generating subprojects/highs/highs/HConfig.h with a custom command
==================================================================== test session starts =====================================================================
platform linux -- Python 3.13.3, pytest-8.3.5, pluggy-1.5.0
rootdir: /home/up19056/development/scipy
configfile: pytest.ini
plugins: hypothesis-6.131.15, anyio-4.9.0, timeout-2.4.0, scipy_doctest-2.0.1, cov-6.1.1, xdist-3.6.1
collected 98 items

scipy/interpolate/tests/test_fitpack2.py ..................................................................................................            [100%]

=============================================================== 98 passed in 840.72s (0:14:00) ===============================================================
```
I haven't looked at whether it is just a difficult test or if it hints at a deficiency of the algorithm/implementation.

czgdp1807:
> I haven't looked at whether it is just a difficult test or if it hints at a deficiency of the algorithm/implementation.

The `test_spline_large_2d_maxit` is the slowest. This test is intentionally demanding to ensure robustness under challenging conditions. Here's a breakdown of the execution time on my Apple Silicon:

- **RectBivariateSpline portion**: ~7.29s
- **New implementations (regrid_python + ndbspline_call_like_bivariate)**: ~44.25s

The test's duration is a result of three factors:

1. **Computationally intensive problem**: The test uses a 1000×1700 data grid with iterative p searching (maxit=25), which is inherently expensive.
2. **Multiple implementations verified within the same test**: The test validates both the `RectBivariateSpline` and new implementations against the same challenging dataset.
3. **Intentional difficulty**: The test is designed to stress-test edge cases and ensure numerical stability in demanding scenarios.

While the new implementations are slower on this particular test configuration, the PR description demonstrates they outperform `RectBivariateSpline` in typical use cases. This test is specifically designed to exercise the algorithms under extreme conditions, making runtime secondary to correctness and stability.

**Heads Up** - I will mark `test_spline_large_2d_maxit` as `xslow`. Also I will add asv benchmarks for `regrid_python` since it is able to beat `RectBivariateSpline` in typical use cases. I hope that would work @j-bowhay. 

cc: @ev-br

czgdp1807:
Latest updates made to this PR, 

1. **Marked `test_spline_large_2d_maxit` as `xslow`** (refer, https://github.com/scipy/scipy/pull/23962/commits/801706e3f69b05b5a4cb2eb8734b4236f1eabf7e)

The test `test_spline_large_2d_maxit` has been marked with the `xslow` decorator to indicate it's a very slow test.

2. **Added Benchmark: `RectBivariateSplineVsRegridPython`** (refer, https://github.com/scipy/scipy/pull/23962/commits/c330ac66f005babc9e6f77793363ee59f75f6046)

New performance benchmark comparing `RectBivariateSpline` vs `scipy.interpolate._regrid_python.regrid_python` fit time across various grid sizes and smoothing parameters.

**Results**

**`time_rect_bivariate_spline`:**

| size | s=1e-12 | s=3.0 |
|------|---------|-------|
| (256, 512) | 326±0.5ms | 282±1ms |
| (512, 512) | 893±8ms | 460±7ms |
| (512, 1024) | 1.47±0s | 887±0.9ms |

**`time_regrid_python`:**

| size | s=1e-12 | s=3.0 |
|------|---------|-------|
| (256, 512) | 151±2ms | 438±1ms |
| (512, 512) | 332±1ms | 835±3ms |
| (512, 1024) | 899±2ms | 2.01±0s |

**Key Observations**

- **Interpolation-like regime (s=1e-12):** `regrid_python` is faster (1.6x to 2.7x faster), showing consistent speedup across all tested sizes
- **Smoothing regime (s=3.0):** `regrid_python` is slower than `RectBivariateSpline` (1.55x to 2.27x slower).

cc: @j-bowhay @ev-br

czgdp1807:
@j-bowhay Kindly review [my response](https://github.com/scipy/scipy/pull/23962#issuecomment-3794289727) to [your comment](https://github.com/scipy/scipy/pull/23962#pullrequestreview-3670151961) regarding slow tests. Also I have [taken two steps](https://github.com/scipy/scipy/pull/23962#issuecomment-3819716829) after taking your comment into consideration. Please let me know if they are sufficient.

Also a technical summary from my investigation:

- For `s > 0`, `fpregr` does **not** jump to interpolatory knots by default. It starts from minimal knots (polynomial LSQ), grows knots only as needed, and stops early when `fp < s`, then runs the `p`-search.
- The `_regrid_python` loop involves Python-level function calls each iteration, whereas `fpregr` runs the loop in compiled code.
- Always forcing interpolatory knots before `p`-search would be incorrect for cases where `fp < s` early; it also defeats the intended "minimal knots + early exit" behavior. That said, for very large grids (and large `s`) the "jump to interpolatory knots + p‑search" path can be faster in practice because it avoids many Python‑level iterations, even if it is not generally valid as a replacement strategy.

**Conclusion:** to close the performance gap for `s >> 0`, the biggest win is moving the knot-growth loop (residual partitioning + `add_knot`) into compiled code. Small Python-side optimizations (buffer reuse, fewer allocations, avoiding CSR conversion) may help but won't match `fpregr` speed.

**Recommendation:** it may be best to merge `_regrid_python` to `main` in its current state and follow up with optimization work afterwards. For `s >> 0` we’re already within ~3x of `fpregr` for most use cases.

tylerjereddy:
Reviewers haven't really been active here for a few months--unless a reviewer champions moving this across the line I'm hesitant to merge this ~a week before we branch for `1.18.0`, so I'm going to bump the milestone.

ev-br:
I think we could merge this for 1.18 as a private API and iterate. While how exactly to make it public needs some more discussion/deliberation,  the actual "meat" of this PR is IMO solid, and there is no advantage to keeping it in an unmerged PR. Merging it as a non-public API will not significantly increase the maintenance overhead (and if anything shows up, I'm volunteering to handle it), and it will if anything unblock further experimentation and polishing. 

To help further experimentation, I think Gagan was planning to write a blog post at some point (?), also we could potentially mention this in the "other changes" section of the release notes---not that we often use it for provisional, experimental APIs, so I'm just throwing a suggestion for the RM to consider.

Two small things before merging @czgdp1807 : 
- could you please prune the git history a bit;
-  could you rename `regrid_python` to just `_regrid`---both the module name and the function name.

tylerjereddy:
Got it, thanks, just switch the milestone back when you merge.

czgdp1807:
@ev-br I have addressed both the requirements, 

> prune the git history a bit

and,

> rename `regrid_python` to just `_regrid`---both the module name and the function name.

Please let me know if anything else to be done.

czgdp1807:
The [following test failure](https://github.com/scipy/scipy/actions/runs/25679529876/job/75386888029?pr=23962) is unrelated to this PR (different module),

```
PARALLEL FAILED scipy/special/tests/test_sf_error.py::test_errstate - TypeError: expected string or bytes-like object, got 'SpecialFunctionError'
= 63724 passed, 18916 skipped, 11643 deselected, 134 xfailed, 1 error in 1580.25s (0:26:20) =
```

This PR is ready for merge.

cc: @ev-br @rgommers

ev-br (approved):
LGTM, let's land this now and figure the details of how to make it public in the 1.19 timeframe.
Thanks @czgdp1807 , it's a huge piece of work!

Issue #23967, opened 2025-11-11
ENH: stats.pointbiserialr: replace implementation


mdhaber:
The point-biserial correlation is a special case of Pearson's correlation when one of the variables is dichotomous. Due to this simplification, there is a shortcut for computing the coefficient itself, and the p-value of a test of no correlation can be computed exactly.

Per the discussion in gh-627, gh-5166 removed a special-case implementation in favor of calling `stats.pearsonr`. I'd suggest that if we're going to have the function in the namespace, we should offer a shortcut implementation with an exact p-value.

PR #22155, opened 2024-12-21
ENH: stats: add axis/nan_policy/keepdims/etc. support to correlation tests

mdhaber:
#### Reference issue
Toward gh-14651

#### What does this implement/fix?
This adds add axis/nan_policy/keepdims/etc. support to most correlation functions. Exceptions (right now) are `pearsonr` because its return object stores `x` and `y` (complicating things) and `spearmanr` because its current `axis` implementation is inconsistent with all other stats functions.

PR #23968, opened 2025-11-11
ENH: stats.pointbiserialr: add array API support

mdhaber:
#### Reference issue
Toward gh-20544

Adds array API support to `scipy.stats.pointbiserialr`.

#### Additional information
I'd suggest we only merge this if we're going to address gh-23967. Otherwise, let's leave this function NumPy-only and phase it out.

### DIFF
--- a/scipy/stats/_stats_py.py
+++ b/scipy/stats/_stats_py.py
@@ -5454,11 +5454,11 @@ def spearmanr(a, b=None, axis=0, nan_policy='propagate',
         return res
 
 
-@xp_capabilities(np_only=True)
+@xp_capabilities(cpu_only=True, exceptions=['cupy', 'jax.numpy'])
 @_axis_nan_policy_factory(_pack_CorrelationResult, n_samples=2,
                           result_to_tuple=_unpack_CorrelationResult, paired=True,
                           too_small=1, n_outputs=3)
-def pointbiserialr(x, y):
+def pointbiserialr(x, y, *, axis=0):
     r"""Calculate a point biserial correlation coefficient and its p-value.
 
     The point biserial correlation is used to measure the relationship
@@ -5476,6 +5476,9 @@ def pointbiserialr(x, y):
         Input array.
     y : array_like
         Input array.
+    axis : int or None, default
+        Axis along which to perform the calculation. Default is 0.
+        If None, ravel both arrays before performing the calculation.
 
     Returns
     -------
@@ -5546,7 +5549,7 @@ def pointbiserialr(x, y):
            [ 0.8660254,  1.       ]])
 
     """
-    rpb, prob = pearsonr(x, y)
+    rpb, prob = pearsonr(x, y, axis=axis)
     # create result object with alias for backward compatibility
     res = SignificanceResult(rpb, prob)
     res.correlation = rpb
--- a/scipy/stats/_stats_py.py
+++ b/scipy/stats/_stats_py.py
@@ -5476,7 +5476,7 @@ def pointbiserialr(x, y, *, axis=0):
         Input array.
     y : array_like
         Input array.
-    axis : int or None, default
+    axis : int or None, default: 0
         Axis along which to perform the calculation. Default is 0.
         If None, ravel both arrays before performing the calculation.
 
--- a/scipy/stats/tests/test_stats.py
+++ b/scipy/stats/tests/test_stats.py
@@ -6523,21 +6523,23 @@ class TestJarqueBera:
         xp_assert_close(res.pvalue, resT.pvalue)
 
 
-def test_pointbiserial():
-    # same as mstats test except for the nan
-    # Test data: https://web.archive.org/web/20060504220742/https://support.sas.com/ctx/samples/index.jsp?sid=490&tab=output
-    x = [1,0,1,1,1,1,0,1,0,0,0,1,1,0,0,0,1,1,1,0,0,0,0,0,0,0,0,1,0,
-         0,0,0,0,1]
-    y = [14.8,13.8,12.4,10.1,7.1,6.1,5.8,4.6,4.3,3.5,3.3,3.2,3.0,
-         2.8,2.8,2.5,2.4,2.3,2.1,1.7,1.7,1.5,1.3,1.3,1.2,1.2,1.1,
-         0.8,0.7,0.6,0.5,0.2,0.2,0.1]
-    assert_almost_equal(stats.pointbiserialr(x, y)[0], 0.36149, 5)
+@make_xp_test_case(stats.pointbiserialr)
+class TestPointBiserialR:
+    def test_pointbiserial(self, xp):
+        # same as mstats test except for the nan
+        # Test data: https://web.archive.org/web/20060504220742/https://support.sas.com/ctx/samples/index.jsp?sid=490&tab=output
+        x = [1,0,1,1,1,1,0,1,0,0,0,1,1,0,0,0,1,1,1,0,0,0,0,0,0,0,0,1,0,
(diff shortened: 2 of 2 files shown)


scipy/stats/tests/test_stats.py
```diff
@@ -6523,21 +6523,23 @@ def test_axis(self, xp):
...
-         2.8,2.8,2.5,2.4,2.3,2.1,1.7,1.7,1.5,1.3,1.3,1.2,1.2,1.1,
-         0.8,0.7,0.6,0.5,0.2,0.2,0.1]
-    assert_almost_equal(stats.pointbiserialr(x, y)[0], 0.36149, 5)
+@make_xp_test_case(stats.pointbiserialr)
+class TestPointBiserialR:
+    def test_pointbiserial(self, xp):
+        # same as mstats test except for the nan
+        # Test data: https://web.archive.org/web/20060504220742/https://support.sas.com/ctx/samples/index.jsp?sid=490&tab=output
+        x = [1,0,1,1,1,1,0,1,0,0,0,1,1,0,0,0,1,1,1,0,0,0,0,0,0,0,0,1,0,
+             0,0,0,0,1.]
```
tylerjereddy, line 6532: just a note for other reviewers since it is subtle--the `1` -> `1.` transformation at the end of this list is used to avoid a failure for `array_api_strict` type promotion rules

scipy/stats/_stats_py.py
```diff
@@ -5476,6 +5476,9 @@ def pointbiserialr(x, y):
         Input array.
     y : array_like
         Input array.
+    axis : int or None, default
```
tylerjereddy: purge out "dangling" `default` here since it is specified below?

scipy/stats/_stats_py.py
```diff
@@ -5454,11 +5454,11 @@ def spearmanr(a, b=None, axis=0, nan_policy='propagate',
         return res
 
 
-@xp_capabilities(np_only=True)
+@xp_capabilities(cpu_only=True, exceptions=['cupy', 'jax.numpy'])
 @_axis_nan_policy_factory(_pack_CorrelationResult, n_samples=2,
                           result_to_tuple=_unpack_CorrelationResult, paired=True,
                           too_small=1, n_outputs=3)
-def pointbiserialr(x, y):
+def pointbiserialr(x, y, *, axis=0):
```
tylerjereddy, line 5461: Running the full array API testsuite for `stats` passes just fine on this branch without the `axis` shim. A simple enough change that no more testing is worth adding for that?

tylerjereddy (commented):
I added some minor comments--this looks pretty straightforward.

Not commenting on the cross-linked discussion, I'll leave that to `stats` regulars.

Checked that the CI failures are all clearly unrelated.

scipy/stats/_stats_py.py
mdhaber: It doesn't actually get [rendered ](https://output.circle-artifacts.com/output/job/c820ad0f-1900-4c1a-aa8e-73d5b469eb42/artifacts/0/html/reference/generated/scipy.stats.pointbiserialr.html) like that
image

but yes, let's:
```suggestion
    axis : int or None, default: 0
```

scipy/stats/_stats_py.py
mdhaber: It does get tested - the `_axis_nan_policy` decorator sees that `pointbiserialr` accepts `axis`, and so it passes it `axis` rather than assuming it is a 1-D implementation that needs to be wrapped. And the `_axis_nan_policy` tests are already testing the `axis` argument of the public function.

tylerjereddy:
> It does get tested 

Can you elaborate on that a bit more. For example, we could do a release where there is an `axis` argument, and then we could delete that argument, do another release, and the testsuite would just silently allow this? Or I'm still not understanding? I can see not wanting to add test shims manually every time you add an `axis` argument somewhere, but I'm a little worried about that.

mdhaber:
Here is the signature of the function in SciPy 1.16.3:
https://github.com/scipy/scipy/blob/b9105ccc2237f57acb1060202cd77f6dd264fb34/scipy/stats/_stats_py.py#L5443-L5446
 
Yet here is the corresponding documentation from the website:
image

And here is some example code and the output in Colab:
```python3
import numpy as np
import scipy
from scipy import stats
rng = np.random.default_rng()
x, y = rng.random((2, 3, 10))
print(scipy.__version__)
stats.pointbiserialr(x, y, axis=-1)
# 1.16.3
# SignificanceResult(statistic=array([ 0.15764425, -0.47154727,  0.47500216]), pvalue=array([0.6635962 , 0.16887332, 0.1653416 ]))
```

The `_axis_nan_policy_factory` decorator added in gh-22155 is doing the work of adding the additional arguments and behavior. The behaviors of these features have been tested in `test_axis_nan_policy.py` since then, too.

So yes, you can add or remove the explicit `axis` argument from the function, and the tests will still pass because the feature will still be there. The decorator implements it by looping in Python over 1-D slices.

The advantage of adding the argument explicitly and passing it to `pearsonr` is that `pearsonr` is actually implemented for N-d arrays with vectorized array API calls, so processing arrays with many slices should be faster. This code should give you the same result with or without the `axis` argument in the signature and the call to `pearsonr`:
```python3
import numpy as np
import scipy
from scipy import stats
rng = np.random.default_rng(4715472715764425)
x, y = rng.random((2, 10000, 10))
# res = stats.pointbiserialr(x, y, axis=-1)
%timeit stats.pointbiserialr(x, y, axis=-1)
# with `axis` in signature / call to `pearsonr` (PR)
# 10 ms ± 421 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
# without `axis` in signature / call to `pearsonr` (main)
# 5.22 s ± 1.28 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
but it is much faster with `axis` appearing explicitly.

tylerjereddy:
very helpful, thanks

mdhaber (commented):
@tylerjereddy if this looks good to you, let's go ahead and merge it. I think there is a good case to be made for special-casing `pointbiserialr` as suggested in gh-23967 rather than deprecating. 

The CI failure does not seem to be related.

tylerjereddy:
>  let's go ahead and merge it

Ok, I read it over a bit more. The issue mentioned actually has no discussion, just your original comment about specializing the function.

It is a bit annoying that I can't find evidence of a test that fails before any of the source changes applied (even if you update the decorator and keep the rest of the function unchanged, the array API-leveraging testsuite still seems happy). I guess it is already compatible basically.

I suppose the patch is small and folks have had 3 months to review it by now, and I also don't see any evidence of a change in numerical behavior--the adjusted test is just enforcing the same numerical values it seems, so risk seems fairly low.

tylerjereddy:
thanks

mdhaber:
>  I guess it is already compatible basically

Yeah, the only thing this changed is 1) that `pointbiserialr` is now tested against all the backends and 2) vectorized calls are now faster because I pass the `axis` argument through to `pearsonr` (which is now natively vectorized) rather than relying on the `_axis_nan_policy` decorator to loop over slices.

```python3
import numpy as np
from scipy import stats
rng = np.random.default_rng(8524430873)
x, y = rng.random((2, 1000, 1000))
%timeit stats.pointbiserialr(x, y, axis=-1)
# (main before PR) 249 ms ± 1.21 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# (main after PR) 20.4 ms ± 535 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

mdhaber:
Thanks for reviewing!

PR #23972, opened 2025-11-12
ENH: stats.quantile_test: vectorize_implementation

mdhaber:
#### Reference issue
Toward gh-20544

Vectorizes the implementation of `scipy.stats.quantile_test`.

#### Additional information
This vectorizes `quantile_test` similarly to how `quantile`, `_xp_searchsorted`, and the proposed `ogive` (gh-23949) are vectorized. The `_axis_nan_policy` decorator and tests don't cover this case, so we have to add quite a few new tests. I'll try to condense what is here to avoid duplication, though.

~~I think we'll want to add a `confidence_interval` method to `quantile`. There are several methods available, and one of them is the ones produced by the `confidence_interval` method of `quantile_test`. They are non-parametric and somewhat conservative, but they are valid for any `method` of `quantile` and IIRC the one-sided CIs are as tight as non-parametric CIs can be in this case.~~ *We can't add a method to result of `quantile` because the result of `quantile` is an array, not an extensible object.*


scipy/stats/_stats_py.py
```diff
@@ -8982,6 +8984,15 @@ class QuantileTestResult:
...
     _x : np.ndarray = field(repr=False)
     _p : float = field(repr=False)
+    _statistic : float = field(repr=False)
+    _statistic_type : int = field(repr=False)
+    _pvalue : float = field(repr=False)
+    _axis : int = field(repr=False)
+    _axis_none : bool = field(repr=False)
+    _keepdims : bool = field(repr=False)
+    _ndim : int = field(repr=False)
+    _nan_out : bool = field(repr=False)
```
mdhaber, line 9006: The `confidence_interval` method draws the endpoints of the confidence intervals from the original data, so we store `_statistic`, `_statistic_type`, and `_pvalue` without any dimensions reduced away to keep the dimensions lined up with the data stored in `_x`. It might be possible to just reshape the `statistic`, `statistic_type`, and `pvalue` attributes as needed so we don't need to store copies of the data, but that is an enhancement that can come later. This PR is complex enough; let's focus on correctness. The rest of the attributes get used, too:
- `_keepdims` is needed to remember whether we need to reduce dimensions away
- `_axis_none` remembers whether the user passed `axis=None`, which means that the data was raveled
- `_ndim` is to remember the original number of dimensions in case the user passed `axis=None` and `keepdims=True`
- `_axis` to make sure we move the axes as they should be if `keepdims=True`
- `_nan_out` remembers where to inject NaNs in the result

scipy/stats/_stats_py.py
```diff
@@ -9018,64 +9029,136 @@ def confidence_interval(self, confidence_level=0.95):
...
+        n = x.shape[-1]
         bd = stats.binom(n, p)
+        shape = p.shape
 
         if confidence_level <= 0 or confidence_level >= 1:
             message = "`confidence_level` must be a number between 0 and 1."
             raise ValueError(message)
 
-        low_index = np.nan
-        high_index = np.nan
```
mdhaber, line 9043: These were unused.

scipy/stats/_stats_py.py
```diff
@@ -9018,64 +9029,136 @@ def confidence_interval(self, confidence_level=0.95):
...
-
-        if alternative == 'less':
+        if n == 0:
+            zeros = np.zeros(p.shape, dtype=x.dtype)
+            low, high = zeros, zeros
```
mdhaber, line 9055: `n` is the length of the data along `axis`. When it's zero, these low and high endpoints (like all return values) will get nan-ed out, so set them to anything of the correct shape and dtype for now.

scipy/stats/_stats_py.py
```diff
@@ -9018,64 +9029,136 @@ def confidence_interval(self, confidence_level=0.95):
...
+        elif alternative == 'less':
             p = 1 - confidence_level
-            low = -np.inf
-            high_index = int(bd.isf(p))
-            high = x[high_index] if high_index < n else np.nan
+            low = np.full(shape, -np.inf)
+            high_index = bd.isf(p).astype(int)
+            valid_index = high_index < n
+            high_index[~valid_index] = 0
+            high = np.where(valid_index, np.take_along_axis(x, high_index), np.nan)
```
mdhaber: `take_along_axis` doesn't like invalid indices. If the index is out of bounds, the result is supposed to be NaN, so we use an arbitrary valid index and NaN-out the result.

scipy/stats/_stats_py.py
```diff
@@ -9018,64 +9029,136 @@ def confidence_interval(self, confidence_level=0.95):
...
+            valid_index = low_index >= 0
+            low_index[~valid_index] = 0
+            low = np.where(valid_index, np.take_along_axis(x, low_index), np.nan)
+            high_index = bd.isf(p).astype(int)
+            valid_index = high_index < n
+            high_index[~valid_index] = 0
+            high = np.where(valid_index, np.take_along_axis(x, high_index), np.nan)
+
+        args = self._axis, self._axis_none, self._keepdims, self._ndim, self._nan_out
+        low = _quantile_test_postprocess(low, *args)
```
mdhaber, line 9086: This takes care of reducing away dimensions if `_keepdims=False`, adding dimensions back if `axis=None` and `keepdims=True` was passed, and NaNing-out values based on the nan-pattern of the input (or if the working axis. has zero length).

scipy/stats/_stats_py.py
```diff
@@ -9018,64 +9029,136 @@ def confidence_interval(self, confidence_level=0.95):
...
 
 
-def quantile_test_iv(x, q, p, alternative):
+def quantile_test_iv(x, q, p, alternative, axis, keepdims):
 
     x = np.atleast_1d(x)
-    message = '`x` must be a one-dimensional array of numbers.'
-    if x.ndim != 1 or not np.issubdtype(x.dtype, np.number):
+    message = '`x` must be an array of numbers.'
+    if not np.issubdtype(x.dtype, np.number):
```
mdhaber, line 9095: We'll look more specifically for real numbers when we translate to array API.

scipy/stats/_stats_py.py
```diff
@@ -9018,64 +9029,136 @@ def confidence_interval(self, confidence_level=0.95):
...
     q = np.array(q)[()]
-    message = "`q` must be a scalar."
-    if q.ndim != 0 or not np.issubdtype(q.dtype, np.number):
+    message = "`q` must be a scalar or array of numbers."
+    if not np.issubdtype(q.dtype, np.number):
         raise ValueError(message)
 
     p = np.array(p)[()]
-    message = "`p` must be a float strictly between 0 and 1."
-    if p.ndim != 0 or p >= 1 or p <= 0:
```
mdhaber, line 9079: Where `p >= 1` or `p <= 0`, results get replaced with NaN rather than raising. That way, a single bad entry doesn't cause the function to error out when there are a lot of valid entries to be processed.

Should add a test for these.

scipy/stats/_stats_py.py
```diff
@@ -9353,35 +9436,42 @@ def quantile_test(x, *, q=0, p=0.5, alternative='two-sided'):
...
     # "We will use two test statistics in this test. Let T1 equal "
     # "the number of observations less than or equal to x*, and "
     # "let T2 equal the number of observations less than x*."
-    T1 = np.count_nonzero(X <= x_star)
-    T2 = np.count_nonzero(X < x_star)
+    if X.shape[-1] > 0:
+        T1 = _xp_searchsorted(X, x_star, side='right')
+        T2 = _xp_searchsorted(X, x_star, side='left')
+    else:
+        nan_out = np.ones_like(nan_out)
```
mdhaber, line 9487: Should move this to input validation.

scipy/stats/_stats_py.py
```diff
@@ -9018,64 +9029,136 @@ def confidence_interval(self, confidence_level=0.95):
...
+        p = np.asarray(p, copy=True)
+        p = xpx.at(p, nan_out).set(0.5)
+
+    return x, q, p, alternative, axis, keepdims, axis_none, ndim, nan_out
+
+
+def _quantile_test_postprocess(res, axis, axis_none, keepdims, ndim, nan_out):
+    # Reshape per axis/keepdims
+
+    res[nan_out] = -1 if np.issubdtype(res.dtype, np.integer) else np.nan
```
mdhaber, line 9160: The `statistic` of this function is an order statistic of the data, so it is an integer. This means we can't actually NaN-out the result; instead, we set it to a sentinel value that can never be valid (-1).  Do you agree that's the best we can do? We could raise a `FutureWarning` about making the output floating point, but I wanted to be able to do something now.

`stastic_type` always takes on value `1` or `2` in normal operation, so `-1` is also an appropriate sentinel.

Need to document the part about the sentinel value.

scipy/stats/tests/test_stats.py
```diff
@@ -8704,6 +8724,159 @@ def test_match_conover_examples(self):
...
+        if axis == 0:
+            res_statistic, res_pvalue = res_statistic.T, res_pvalue.T
+            res_low, res_high = res_low.T, res_high.T
+
+        assert_allclose(res_statistic, ref_statistic)
+        assert_allclose(res_pvalue, ref_pvalue)
+        assert_allclose(res_low, res_low)
+        assert_allclose(res_low, ref_low)
+
+    def test_zero_size(self):
```
mdhaber, line 8980: These and the test below are pretty verbose, but a lot of the bulk is due to the fact that there are five separate outputs with two different dtypes, so we can't very easily loop over them. Still, I'll work on condensing them where I can.

scipy/stats/tests/test_stats.py
```diff
@@ -8704,6 +8724,159 @@ def test_match_conover_examples(self):
...
+        x = rng.random(x_shape)
+        q = rng.random(qp_shape)
+        p = q + rng.random(qp_shape) * 1e-2
+
+        # case 1: p/q is size zero.
+        qp_zero = np.empty((0, *qp_shape))
+        out = qp_zero
+        res = stats.quantile_test(x, q=qp_zero, p=qp_zero, axis=-1)
+        ci = res.confidence_interval()
+        xp_assert_equal(res.statistic, np.astype(out, np.int64))
```
mdhaber: I used `assert_close`/`assert_equal` elsewhere... no harm done in accidentally using `xp_assert_`s here because that will be the next step anyway.

mdhaber (commented):
Some self-review to help the reviewer.

scipy/stats/_stats_py.py
```diff
@@ -9018,64 +9029,136 @@ def confidence_interval(self, confidence_level=0.95):
...
     message = f"`alternative` must be one of {alternatives}"
     if alternative not in alternatives:
         raise ValueError(message)
 
-    return x, q, p, alternative
+    if keepdims not in {None, True, False}:
+        message = "If specified, `keepdims` must be True or False."
+        raise ValueError(message)
+
+    x = np.sort(x, axis=axis, stable=False)
```
mdhaber: `stable` not in NumPy < 2.0. We can add it back when this is translated to array API.
```suggestion
    x = np.sort(x, axis=axis)
```

scipy/stats/tests/test_stats.py
steppi: I think I see how this could be refactored, but I'm fine with leaving it like this for now. Perfect is the enemy of the good enough.

scipy/stats/_stats_py.py
steppi: I think the principled thing to do if starting from scratch would be to make the output a floating point value. I think you had convinced me at some point to use a floating point dtype for all integers appearing in stats, to get access to `NaN` and graceful overflow. My vote is to do the `FutureWarning` thing. If you want to use a sentinel value in the meantime, let's document that this is a temporary solution until the output dtype is changed.

scipy/stats/_stats_py.py
```diff
@@ -9076,64 +9086,138 @@ def confidence_interval(self, confidence_level=0.95):
...
+
+    if not keepdims:
+        res = np.squeeze(res, axis=axis)
+    return res[()]
 
 
 @xp_capabilities(np_only=True)
-def quantile_test(x, *, q=0, p=0.5, alternative='two-sided'):
+def quantile_test(x, *, q=0, p=0.5, alternative='two-sided', axis=0, keepdims=None):
     r"""
```
steppi, line 9176: I think it's important that the changes here are reflected in the documentation. Did you plan to do that in a follow-up?

steppi (commented):
I think the implementation looks good enough to merge now and iterate on. Same for the tests. For the question of sentinel values for the statistic, my vote is for moving to floating point dtypes for all integer-valued things appearing in stats, but I think it's probably fine to have a temporary solution in the meantime so long as its well-documented as such. I'd like to see the changes here reflected in the documentation and think this shouldn't be too much trouble to do.

scipy/stats/_stats_py.py
mdhaber: Interesting - from the published documentation:

Image

even though in the notes we document that the statistics are the counts, not the proportions.

So there is some argument to be made that the statistic should be changed reflect the documentation. 

And actually, a `FutureWarning` strategy wouldn't really work here, because there would be no way for the user to silence it.

So I think we will treat this like `rankdata`: https://github.com/scipy/scipy/pull/24420. 

But let's do it in a separate PR. I won't document the sentinel value now, since the documentation is already in an inconsistent state. In immediate followup, I'll make everything consistent, post on the forum and (when the PR merges) include the change in the appropriate section of the release notes.

scipy/stats/_stats_py.py
mdhaber: Nope, done here in the next commit.

scipy/stats/_stats_py.py
mdhaber: I'm not sure why I wrote that. `nan_out` is defined in the input validation. It doesn't make sense to me now that I should return `T1` and `T2` from the input validation function. Besides, when this function gets `marray` support, `X.shape[-1]` is going to be replaced with `_count_nonmasked`, and that can be different for each slice - some will get valid values, others will get NaNs. I'll leave this alone.

scipy/stats/_stats_py.py
mdhaber: Added a test with other input validation tests.

mdhaber (commented):
> I think the implementation looks good enough to merge now and iterate on. 

What about the iteration is there to iterate on?  Certainly the sentinels, but is there something else you had in mind?

scipy/stats/tests/test_stats.py
```diff
@@ -8815,6 +8815,13 @@ def test_quantile_test_iv(self):
         with pytest.raises(ValueError, match=message):
             stats.quantile_test(x, p=2)
 
+        # should p == 0. / p == 1. be valid?
+        p = np.asarray([-1., 0., 1., 1.5, np.nan])
+        res = stats.quantile_test(x, p=p)
+        assert_equal(res.statistic, np.full_like(p, -1, dtype=int))
+        assert_equal(res.statistic_type, np.full_like(p, -1, dtype=int))
```
mdhaber, line 8822: Probably didn't even need to bother with dtype since `assert_equal` doesn't check it; should have left it out since we won't want these to be integers and `int` isn't array-API compatible anyway. Oh well.

scipy/stats/_stats_py.py
```diff
@@ -9211,8 +9319,8 @@ def quantile_test(x, *, q=0, p=0.5, alternative='two-sided'):
     data, so the value is clipped into the interval :math:`[0, 1]`.
 
     The approach for confidence intervals is attributed to Thompson [2]_ and
-    later proven to be applicable to any set of i.i.d. samples [3]_. The
-    computation is based on the observation that the probability of a quantile
+    later proven to be applicable to any i.i.d. sample [3]_. The
+    procedure is based on the observation that the probability of a quantile
```
mdhaber: I was thinking about more changes, but didn't go with them, so `procedure` is fine.
```suggestion
    computation is based on the observation that the probability of a quantile
```

scipy/stats/_stats_py.py
```diff
@@ -9239,25 +9347,25 @@ def quantile_test(x, *, q=0, p=0.5, alternative='two-sided'):
     Two-sided confidence intervals are not guaranteed to be optimal; i.e.,
     there may exist a tighter interval that may contain the quantile of
     interest with probability larger than the confidence level.
-    Without further assumption on the samples (e.g., the nature of the
+    Without further assumption on the sample (e.g., the nature of the
     underlying distribution), the one-sided intervals are optimally tight.
```
mdhaber: "e.g." is wrong here. If anything, it should be "i.e.". Moreover,
```suggestion
    Without further assumption on the nature of the
    underlying distribution, the one-sided intervals are optimally tight.
```

scipy/stats/_stats_py.py
```diff
@@ -9119,6 +9203,30 @@ def quantile_test(x, *, q=0, p=0.5, alternative='two-sided'):
           than `q`.
         * 'greater': the quantile associated with the probability `p` is
           greater than `q`.
+    axis : int or None, default: 0
+        Axis along which the quantiles are computed.
```
mdhaber:
```suggestion
        Axis of `x` along which the test is performed.
```

scipy/stats/_stats_py.py
```diff
@@ -9119,6 +9203,30 @@ def quantile_test(x, *, q=0, p=0.5, alternative='two-sided'):
...
+        ``None`` ravels `x`, `q`, and `p` before performing the calculation,
+        without checking whether the original shapes were compatible.
+        As in other `scipy.stats` functions, a positive integer `axis` is resolved
+        after prepending 1s to the shape of `x`, `q`, and `p` as needed until the arrays
+        have the same dimensionality. When providing `x`, `q`, and `p` with different
+        dimensionality, consider using negative `axis` integers for clarity.
+    keepdims : bool, optional
+        Consider the case in which `x` is 1-D and `p` and `q` are scalars: the quantile
+        is a reducing statistic, and the default behavior is to return a scalar.
```
mdhaber:
```suggestion
        Consider the case in which `x` is 1-D and `p` and `q` are scalars: the test
        computes a reducing statistic, and the default behavior is to return a scalar.
```

scipy/stats/_stats_py.py
steppi: That sounds like a good plan.

steppi:
> > I think the implementation looks good enough to merge now and iterate on.
> 
> What about the iteration is there to iterate on? Certainly the sentinels, but is there something else you had in mind?

You had actually suggested some iterations earlier, such as here https://github.com/scipy/scipy/pull/23972#discussion_r2517091866. I was just commenting that I agree things like this aren't necessary to merge.

steppi (approved):
Thanks @mdhaber. I think this looks good to go now. I like the uncommitted docs suggestions you've made, so I'll wait until you commit those to merge.

mdhaber:
Thanks @steppi. My plan to provide the functionality of [`stats.mstats.mjci`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.mstats.mjci.html)/[`mquantiles_cimj`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.mstats.mquantiles_cimj.html) in `scipy.stats` is to add a Maritz-Jarrett `method` to `quantile_test`. 

- `statistic_type` would take on a third value
- I would probably not add a public `standard_error` attribute, because it wouldn't be available for the default `method`. Technically that would be needed for a drop-in replacement of `mjci`, but I figure it *could* be determined by the user from the confidence interval if needed.
- It is not common to perform a hypothesis test corresponding with the Maritz-Jarrett standard error, but it seems perfectly valid since using it + normality assumption to generate a confidence interval is standard. TBH, the original request for `quantile_test` was to just add a function that computed the confidence interval; we decided to *also* make it a hypothesis test to fit better with the rest of `stats`, so I think it's worth doing the same thing here.

If that sounds good, I'll proceed after translating this to the array API.

steppi:
> Thanks @steppi. My plan to provide the functionality of [`stats.mstats.mjci`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.mstats.mjci.html)/[`mquantiles_cimj`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.mstats.mquantiles_cimj.html) in `scipy.stats` is to add a Maritz-Jarrett `method` to `quantile_test`.
> 
>     * `statistic_type` would take on a third value
> 
>     * I would probably not add a public `standard_error` attribute, because it wouldn't be available for the default `method`. Technically that would be needed for a drop-in replacement of `mjci`, but I figure it _could_ be determined by the user from the confidence interval if needed.
> 
>     * It is not common to perform a hypothesis test corresponding with the Maritz-Jarrett standard error, but it seems perfectly valid since using it + normality assumption to generate a confidence interval is standard. TBH, the original request for `quantile_test` was to just add a function that computed the confidence interval; we decided to _also_ make it a hypothesis test to fit better with the rest of `stats`, so I think it's worth doing the same thing here.
> 
> 
> If that sounds good, I'll proceed after translating this to the array API.

That all sounds good to me.

PR #23979, opened 2025-11-12
BUG: special: fix `comb(n, 0, repetition=True)` returning `0` instead of `1`

### LINKED ISSUE #23867: BUG: `special.comb(0, 0, repetition=False) == 0` , not 1
### Describe your issue.

Just like how "0 choose 0" == 1, we also have that "0 multichoose 0" == 1 (since we're actually talking about [multiset coefficients](https://en.wikipedia.org/wiki/Multiset#Counting_multisets)). But `special.comb` does not support negative `n` (i.e. it isn't the [Roman coefficient](https://mathworld.wolfram.com/RomanCoefficient.html), as requested in #15213). And since `repetition=True` simply returns `comb(n + k - 1, k)`, it will return `comb(0 + 0 - 1, 0) == comb(-1, 0)`, which evaluates to 0. It's not difficult to see why that's wrong, and wolfram alpha confirms that it's indeed 1: https://www.wolframalpha.com/input?i=0+multichoose+0

Here's the full matrix (calculated using a Wolfram language notebook) with `n` as rows and `k` as columns for $\binom{n+k-1}{k}$:

$$
\left(
\begin{array}{cccccc}
 1 & 0 & 0 & 0 & 0 & 0 \\
 1 & 1 & 1 & 1 & 1 & 1 \\
 1 & 2 & 3 & 4 & 5 & 6 \\
 1 & 3 & 6 & 10 & 15 & 21 \\
 1 & 4 & 10 & 20 & 35 & 56 \\
 1 & 5 & 15 & 35 & 70 & 126 \\
\end{array}
\right)
$$

### Reproducing Code Example

```python
>>> from scipy.special import comb
>>> comb(0, 0, exact=True, repetition=False)
1
>>> comb(0, 0, exact=True, repetition=True)
0
>>> comb(0, 0, exact=False, repetition=False)
np.float64(1.0)
>>> comb(0, 0, exact=False, repetition=True)
np.float64(0.0)
```

ssam18:
Fixes #23867

## Description
This PR fixes a bug where  incorrectly returns 0 instead of 1 for any n >= 0.

## Root Cause
When , the implementation transforms the multiset coefficient into a binomial coefficient via:
```python
comb(N, k, repetition=True) → comb(N + k - 1, k)
```

For the edge case where `k=0`, this becomes:
```python
comb(n, 0, repetition=True) → comb(n - 1, 0)
```

And specifically for `n=0`:
```python
comb(0, 0, repetition=True) → comb(-1, 0) → 0  # Incorrect!
```

The function returns 0 because the existing logic treats any case with `N < 0` as 0.

## Solution
Mathematically, C(n, 0) = 1 for any n >= 0, since there is exactly one way to choose zero elements from any set. This fix adds a special case check within the `if repetition:` block to handle `k=0` before applying the transformation.

The fix handles both:
- `exact=True`: returns integer 1
- `exact=False`: returns float 1.0

## Testing
Added test `test_comb_repetition_k_zero` to verify:
- `comb(0, 0, repetition=True)` returns 1
- `comb(n, 0, repetition=True)` returns 1 for various n >= 0
- Both exact and non-exact modes work correctly


scipy/special/_basic.py
```diff
@@ -2528,6 +2528,15 @@ def comb(N, k, *, exact=False, repetition=False):
...
     if repetition:
+        # Special case: C(n, 0) with repetition = 1 for n >= 0
+        # Without this check, comb(0, 0, repetition=True) would compute
+        # comb(-1, 0) which incorrectly returns 0
+        if exact:
+            if int(k) == k and k == 0 and int(N) == N and N >= 0:
+                return 1
+        else:
+            if np.isscalar(k) and k == 0 and np.isscalar(N) and N >= 0:
+                return np.float64(1.0)
```
jorenham: For array-like input, you could use the same `cond` trick as below at line 2550.

scipy/special/_basic.py
```diff
@@ -2528,6 +2528,15 @@ def comb(N, k, *, exact=False, repetition=False):
 
     """
     if repetition:
+        # Special case: C(n, 0) with repetition = 1 for n >= 0
+        # Without this check, comb(0, 0, repetition=True) would compute
+        # comb(-1, 0) which incorrectly returns 0
+        if exact:
+            if int(k) == k and k == 0 and int(N) == N and N >= 0:
+                return 1
```
jorenham: The `int(k) == k` check is redundant

```suggestion
            if k == 0 and int(N) == N and N >= 0:
                return 1
```

scipy/special/tests/test_basic.py
```diff
@@ -1561,6 +1561,15 @@ def test_comb_exact_non_int_error(self):
...
 
+    def test_comb_repetition_k_zero(self):
+        # Regression test for gh-23867
+        # C(n, 0) should always be 1 for n >= 0, regardless of repetition
+        assert_equal(special.comb(0, 0, exact=True, repetition=True), 1)
+        assert_equal(special.comb(0, 0, exact=False, repetition=True), 1.0)
+        assert_equal(special.comb(5, 0, exact=True, repetition=True), 1)
+        assert_equal(special.comb(5, 0, exact=False, repetition=True), 1.0)
+        assert_equal(special.comb(10, 0, exact=True, repetition=True), 1)
+        assert_equal(special.comb(10, 0, exact=False, repetition=True), 1.0)
```
jorenham: It'd be cleaner to use `@pytest.mark.parametrize` (twice)

scipy/special/_basic.py
ssam18: make sense. Thank you Jorenham. Just now pushed the changes as per your review comment.

scipy/special/tests/test_basic.py
jorenham, line 1: this could also use a quick little test for array-like input and `exact=False`

ssam18: yeah, you are right. Let me add that quickly.

scipy/special/tests/test_basic.py
```diff
@@ -1561,6 +1561,20 @@ def test_comb_exact_non_int_error(self):
         with pytest.raises(ValueError, match=msg):
             special.comb(3, 4.4, exact=True)
 
+    @pytest.mark.parametrize('N', [0, 5, 10])
+    @pytest.mark.parametrize('exact', [True, False])
+    def test_comb_repetition_k_zero(self, N, exact):
+        # Regression test for gh-23867
+        # C(n, 0) should always be 1 for n >= 0, regardless of repetition
+        expected = 1 if exact else 1.0
```
tylerjereddy: This caught my eye while reading through the diff--I was skeptical that the current assertion would distinguish between `1` and `1.0` and indeed the tests still pass with that removed:

```diff
--- a/scipy/special/tests/test_basic.py
+++ b/scipy/special/tests/test_basic.py
@@ -1566,8 +1566,7 @@ class TestCombinatorics:
     def test_comb_repetition_k_zero(self, N, exact):
         # Regression test for gh-23867
         # C(n, 0) should always be 1 for n >= 0, regardless of repetition
-        expected = 1 if exact else 1.0
-        assert_equal(special.comb(N, 0, exact=exact, repetition=True), expected)
+        assert_equal(special.comb(N, 0, exact=exact, repetition=True), 1)
```

I suppose the distinction is relevant to some degree, given the docstring and what Joren does with the `comb` overloads at `scipy-stubs/special/_basic.pyi`. I don't know that we actually need to worry about that particular detail in this set of tests, but if we do I suppose an `isinstance` check might suit for distinguishing between the exact and not-exact return types/algorithms.

Maybe wait for Joren's input on that since I'm just doing a quick drive by review.

jorenham: Yea that's right; `int`, `float` and `np.float64` are all mutually incompatible types, so replacing one the other is technically a breaking change.

As Tyler noted, `assert_equal` doesn't compare the types, just the values. And since `1 == 1.0 == np.float64(1.0)` is true, this test isn't checking what you seem to want it to. That is, that `exact=True` returns a `1` of type `int`, and that `exact=False` returns a `1.0` of type `np.float64`. That'd require an additional `type(...) is int` assertion, or something similar.

But even with the additional type checks, this test should still pass with flying colors. Because judging by your code, you seem to have already taken these subtleties into consideration. So the purpose of testing it is mostly for preventing regressions.

> the `comb` overloads at `scipy-stubs/special/_basic.pyi`

for reference: https://github.com/scipy/scipy-stubs/blob/ff259fc1c7bce6e0eb2ddf694972c26640101ee9/scipy-stubs/special/_basic.pyi#L380-L389

jorenham (approved):

scipy/special/tests/test_basic.py
```diff
@@ -1561,6 +1561,20 @@ def test_comb_exact_non_int_error(self):
...
+        assert_equal(special.comb(N, 0, exact=exact, repetition=True), expected)
```
lucascolley:
```suggestion
        actual = special.comb(N, 0, exact=exact, repetition=True)
        assert actual == 1
        assert type(actual) is int if exact else np.float64
```

SamareshSingh:
@lucascolley @tylerjereddy @steppi Please let me know if this PR looks good! I am happy to help and contribute further in this repo!

tylerjereddy:
Looks good enough to merge before the branch point now, thanks all.

PR #23981, opened 2025-11-13
ENH: stats._xp_searchsorted: simplification, edge case handling, and performance improvements

mdhaber:
#### Reference issue
Follow-up to gh-23930

Each commit makes an enhancement to `_xp_searchsorted` as described in the commit message.

#### Additional information
I also tested whether the loop should be a `for` loop for JAX (to allow JIT) and a `while` loop for other backends. It turns out that checking the `while` loop condition is sufficiently expensive in some cases that it would have been preferable to do more iterations of a `for` loop. Since this is just trying to pave the way toward implementation in the standard, it doesn't have to be super fast, so I'll let it be. 

@lucascolley thought you'd be interested in these if this is to go in `array-api-extra`.
In a PR to `array-api-extra`, can I just copy paste [`_broadcast_arrays`](https://github.com/scipy/scipy/blob/c6f2976c88cb5419e2f8fdc1efb00fbc49387b92/scipy/stats/_axis_nan_policy.py#L45-L123) and [`_apply_over_batch`](https://github.com/scipy/scipy/blob/c6f2976c88cb5419e2f8fdc1efb00fbc49387b92/scipy/_lib/_util.py#L1098), too? It would be possible to trim them down, and I probably will to some extent, but if I know I have the option of leaving them as they are, I won't be as reluctant to start.


scipy/stats/_quantile.py
```diff
@@ -386,33 +387,38 @@ def _quantile_bc(y, p, n, method, xp):
...
 
 
-
 @xp_capabilities(skip_backends=[("dask.array", "No take_along_axis yet.")])
 def _xp_searchsorted(x, y, *, side='left', xp=None):
-    # Vectorize np.searchsorted. Assumes search is along last axis, which is always
-    # preserved in the output. Does not support zero-length `x`. For side='right',
-    # NaNs in `y` are inserted to the left, in contrast with np.searchsorted.
+    # Vectorize np.searchsorted. Assumes search is along last axis, which is preserved
+    # in the output unless x is 1d and y is 0d.
```
mdhaber: What do you think about this? 
`take_along_axis` avoids this problem - and the ambiguity of positive `axis` - by requiring both inputs to have the same `ndim`. Since this function produces indices, it might make sense to follow suit. But that wouldn't work with the current behavior, for which the output shape is that of the second argument.

scipy/stats/_quantile.py
```diff
@@ -386,33 +387,38 @@ def _quantile_bc(y, p, n, method, xp):
...
-        j = x0 >= y
+        j = compare(y, x0)
         b = xp.where(j, c, b)
         a = xp.where(j, a, c)
 
-    b = xp.where(y <= xp.min(x, axis=-1, keepdims=True), 0, b)
-    return b if side == 'left' else x.shape[-1] - b
+    out = xp.where(compare(y, xp.min(x, axis=-1, keepdims=True)), 0, b)
+    out = xp.where(xp.isnan(y), x.shape[-1], out) if side == 'right' else out
+    return out
```
mdhaber, line 429: The CI failures all have to do with dtype of the output. Should it be the default int type of the platform or should we specify? If the latter, should we use the default int type throughout the calculation and convert (if necessary) at the end?

lucascolley:
> In a PR to `array-api-extra`, can I just copy paste [`_broadcast_arrays`](https://github.com/scipy/scipy/blob/c6f2976c88cb5419e2f8fdc1efb00fbc49387b92/scipy/stats/_axis_nan_policy.py#L45-L123) and [`_apply_over_batch`](https://github.com/scipy/scipy/blob/c6f2976c88cb5419e2f8fdc1efb00fbc49387b92/scipy/_lib/_util.py#L1098), too? It would be possible to trim them down, and I probably will to some extent, but if I know I have the option of leaving them as they are, I won't be as reluctant to start.

In principle seems fine 👍 we do already have a public `xpx.broadcast_shapes` so it would be good to extend that rather than duplicate, for the common functionality.

scipy/stats/_quantile.py
lucascolley: What changes would be needed for ensuring the default int type of the platform?

scipy/stats/_quantile.py
lucascolley: Hmm not sure. I suppose https://github.com/data-apis/array-api/issues/967 would be another exception to the same `ndim` rule.

lucascolley (approved):

scipy/stats/_quantile.py
mdhaber: I think the implementation could be left alone - I don't specify the dtype anywhere. I would just need to ensure that the reference array, which is calculated with NumPy, gets the default integer dtype of the backend.

scipy/stats/_quantile.py
mdhaber: Most of the time I don't like the same `ndim` requirement of `take_along_axis`; I wish it would just broadcast. So maybe we take the position that `take_along_axis` *should* work with arrays of different dimensionality (at least when `axis` is negative), and the shape of the output should be the shape of the second input when possible and the broadcasted shape of the two inputs - ignoring the length of the last axis - otherwise. I think - and please check me - that the only case in which this is *not* the same as the broadcasted shape is that of 1D first argument and 0d second argument. But maybe that way of documenting the output shape is more palatable because it doesn't sound so much like a weird special case.

So:
> Assumes search is along last axis, which is preserved in the output unless x is 1d and y is 0d.

becomes:
> Assumes search is along last axis. The shape of the output is that of `y` if possible and `broadcast_shapes(x, y, axis=-1)` otherwise.

where adding an `axis` argument to `xpx.broadcast_shapes` is what I need to do to avoid duplication with `_broadcast_shapes`.

scipy/stats/_quantile.py
lucascolley: in which case that sounds reasonable to me

scipy/stats/_quantile.py
```diff
@@ -386,33 +387,38 @@ def _quantile_bc(y, p, n, method, xp):
     return xp.take_along_axis(y, xp.astype(k, xp.int64), axis=-1)
 
 
-
 @xp_capabilities(skip_backends=[("dask.array", "No take_along_axis yet.")])
 def _xp_searchsorted(x, y, *, side='left', xp=None):
-    # Vectorize np.searchsorted. Assumes search is along last axis, which is always
-    # preserved in the output. Does not support zero-length `x`. For side='right',
-    # NaNs in `y` are inserted to the left, in contrast with np.searchsorted.
+    # Vectorize np.searchsorted. Assumes search is along last axis, which is preserved
```
lucascolley:
```suggestion
    # Vectorized xp.searchsorted. Assumes search is along last axis, which is preserved
```

scipy/stats/_quantile.py
lucascolley: I think that sounds right, yes.

scipy/stats/_quantile.py
mdhaber: That didn't work. The function is defaulting to `int32`, but `xp.asarray(1).dtype` produces an `int64`. I'll try to reproduce locally, but if I can't, I might skip the dtype checking. I think we should let JAX default do its thing and use int32 if it wants to. It's unfortunate that it seems inconsistent, but it doesn't look like that's the fault of the function.

mdhaber: It was that `jax.numpy.searchsorted`, which this PR delegates to for 1D input, produces an `int32` regardless of JAX's default integer dtype. I added an `astype` to ensure that JAX respects its own default int dtype.

lucascolley: is there an upstream issue? I can open one if not.

lucascolley (approved):
will merge if CI is happy, thanks Matt.

scipy/stats/_quantile.py
```diff
@@ -7,7 +7,9 @@
     array_namespace,
     xp_promote,
     xp_device,
-    _length_nonmasked
+    _length_nonmasked,
+    is_torch,
+    is_jax,
```
mdhaber:
```suggestion
```
`[lint only]`

mdhaber:
Why the merge + force push? I couldn't tell what change, so I reverted to what I had, fixed the CI failure, and pushed back. The only failure with commit [1a146d0](https://github.com/scipy/scipy/pull/23981/commits/1a146d0f5f6e1cfe40b3a72e1e71541ba48faa3f) was lint, so I fixed that, too, and ran CI with `lint only`.

I'm happy with this now. Something to think about before moving to array API extra, though, is that if we delegate to the backends for 1D input, the 1D test cases are only testing the behavior of the backends' `searchsorted` functions. We should have some way of running tests with those lengths along the working axis against the actual `xp_searchsorted` code.

PR #23920, opened 2025-11-03
DEV: update `CODEOWNERS`

lucascolley:
- add myself for `CODEOWNERS`, `*pixi*`, `lint*`, `*array_api*`
- add @rgommers @thalassemia for `scipy/_build_utils/`
- add @rkern for `scipy/odr/`
- add @DietBru for `scipy/signal/`
- add @dschult for `scipy/sparse/`
- remove code owners from `scipy/sparse/csgraph/` and `scipy/sparse/linalg/`
- remove @person142 following move to emeritus
- adjust for deleted/renamed directories
- reorder for a candidate sensible precedence:
	- `.github/workflows` does not dwarf `*pixi*` or `lint*`
	- `*array_api*` does not dwarf submodules
	- submodules do not dwarf `meson*`

feedback welcome :)

PR #23984, opened 2025-11-14
DEV: update CODEOWNERS

dschult:
Add Dan Schult as codeowner for /scipy/sparse

After checking with @perimosocordiae about what this entails I would like to request, as a follow-up to #23920, to be added as a code owner for scipy/sparse. It should help me keep better track of issues/PRs in that space.


lucascolley (approved):
great, thanks for following-up Dan!

PR #23994, opened 2025-11-15
DOC/BLD: Clarify `yum` usage in RHEL

jaimergp:
#### Reference issue

N/A.

Small clarification in the documentation: RHEL 8+ uses `dnf`, not `yum`.

#### Additional information

cc @rgommers. This was mentioned by @tiran as part of the PEP 804 feedback. See https://github.com/jaimergp/peps/pull/9.

### DIFF
--- a/doc/source/building/index.rst
+++ b/doc/source/building/index.rst
@@ -69,7 +69,7 @@ your system.
         advantage that new dependencies or updates to required versions are
         handled by the package managers.
 
-      .. tab-item:: Fedora
+      .. tab-item:: Fedora / RHEL & CentOS 8+
 
         To install SciPy build requirements, you can do::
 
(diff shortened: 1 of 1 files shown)


tylerjereddy (approved):
I checked the updated tabs in the rendered docs and they seem fine. A quick internet search will confirm the change in "preferred"/default package manager as stated here.

That said, multiple sources clearly indicate that one can generally count on `yum` being symlinked to `dnf` for the moment, so this probably isn't urgent. But may as well update I suppose--presumably `yum` will be removed eventually on newer versions.

tylerjereddy:
CI failures obviously not related to the small doc change.

rgommers:
Thanks for the update @jaimergp. We can clean this up further I think, no need to keep CentOS/RHEL <8 instructions around, those versions are from <2019.

Issue #23932, opened 2025-11-05
CI: mypy and clang jobs failing with "unused function" error during compilation


mdhaber:
### Describe your issue.

Most CI workflows are failing with an ImportError. See e.g. gh-23877 and gh-23772.

### Reproducing Code Example

```python
-
```

### Error message

```shell
2025-11-05T16:26:29.0775694Z _______________ ERROR collecting stats/tests/test_variation.py ________________
2025-11-05T16:26:29.0776965Z ImportError while importing test module 'C:\hostedtoolcache\windows\Python\3.11.9\x64\Lib\site-packages\scipy\stats\tests\test_variation.py'.
2025-11-05T16:26:29.0777951Z Hint: make sure your test modules/packages have valid Python names.
2025-11-05T16:26:29.0778383Z Traceback:
2025-11-05T16:26:29.0778843Z C:\hostedtoolcache\windows\Python\3.11.9\x64\Lib\importlib\__init__.py:126: in import_module
2025-11-05T16:26:29.0779569Z     return _bootstrap._gcd_import(name[level:], package, level)
2025-11-05T16:26:29.0780050Z            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2025-11-05T16:26:29.0780631Z C:\hostedtoolcache\windows\Python\3.11.9\x64\Lib\site-packages\scipy\stats\__init__.py:631: in <module>
2025-11-05T16:26:29.0781236Z     from ._multicomp import *
2025-11-05T16:26:29.0781830Z C:\hostedtoolcache\windows\Python\3.11.9\x64\Lib\site-packages\scipy\stats\_multicomp.py:11: in <module>
2025-11-05T16:26:29.0782473Z     from scipy.stats._qmc import check_random_state
2025-11-05T16:26:29.0783066Z C:\hostedtoolcache\windows\Python\3.11.9\x64\Lib\site-packages\scipy\stats\_qmc.py:26: in <module>
2025-11-05T16:26:29.0783693Z     from scipy.sparse.csgraph import minimum_spanning_tree
2025-11-05T16:26:29.0784421Z C:\hostedtoolcache\windows\Python\3.11.9\x64\Lib\site-packages\scipy\sparse\csgraph\__init__.py:201: in <module>
2025-11-05T16:26:29.0785165Z     from ._reordering import reverse_cuthill_mckee, structural_rank
[...]

PR #23952, opened 2025-11-09
MAINT: stats: silence the "unused function" build noise from Cython

ev-br:
Silence noise from building `stats/qmc_cy` with -Werror:

```
FAILED: [code=1] scipy/stats/_qmc_cy.cpython-312-x86_64-linux-gnu.so.p/meson-generated__qmc_cy.cpp.o 
clang++-17 -Iscipy/stats/_qmc_cy.cpython-312-x86_64-linux-gnu.so.p -Iscipy/stats -I../scipy/stats -I../../../../../../opt/hostedtoolcache/Python/3.12.12/x64/lib/python3.12/site-packages/numpy/_core/include -I/opt/hostedtoolcache/Python/3.12.12/x64/include/python3.12 -fvisibility=hidden -fvisibility-inlines-hidden -fdiagnostics-color=always -DNDEBUG -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Werror -std=c++17 -O3 -fPIC -pthread -DNPY_NO_DEPRECATED_API=NPY_1_9_API_VERSION -DCYTHON_CCOMPLEX=0 -MD -MQ scipy/stats/_qmc_cy.cpython-312-x86_64-linux-gnu.so.p/meson-generated__qmc_cy.cpp.o -MF scipy/stats/_qmc_cy.cpython-312-x86_64-linux-gnu.so.p/meson-generated__qmc_cy.cpp.o.d -o scipy/stats/_qmc_cy.cpython-312-x86_64-linux-gnu.so.p/meson-generated__qmc_cy.cpp.o -c scipy/stats/_qmc_cy.cpython-312-x86_64-linux-gnu.so.p/_qmc_cy.cpp
scipy/stats/_qmc_cy.cpython-312-x86_64-linux-gnu.so.p/_qmc_cy.cpp:13536:1: error: unused function '__pyx_memoryview_copy_new_contig' [-Werror,-Wunused-function]
 13536 | __pyx_memoryview_copy_new_contig(const __Pyx_memviewslice *from_mvs,
       | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

observed in https://github.com/scipy/scipy/pull/23921 and elsewhere.

PR #24003, opened 2025-11-16
MAINT: stats: re-enable the "unused function" warning in qmc and ckdtree

nickodell:
#### Reference issue

In order to fix issue #23932, PR #23952 silenced an "unused function" warning originating from Cython incorrectly detecting a dependency of some C code.

Cython has fixed this upstream in https://github.com/cython/cython/pull/7293, and the fix is included in Cython 3.2.1, so we don't need to ignore this any more. Re-enable the warning to catch future issues.


lucascolley (approved):
LGTM, thanks Nick! Anything more to take it out of draft?

Issue #23997, opened 2025-11-16
CI: special/stats: failures for `sf_error` and `MatrixT` on dev deps job


lucascolley:
```
FAILED scipy/special/tests/test_sf_error.py::test_seterr - SystemError: <built-in function len> returned a result with an exception set
FAILED scipy/special/tests/test_sf_error.py::test_sf_error_special_refcount - SystemError: <ufunc 'ndtri'> returned a result with an exception set
FAILED scipy/special/tests/test_sf_error.py::test_errstate - SystemError: <class 'Failed'> returned a result with an exception set
FAILED scipy/special/tests/test_sf_error.py::test_errstate_all_but_one - SystemError: <ufunc 'spence'> returned a result with an exception set
FAILED scipy/stats/tests/test_multivariate.py::TestMatrixT::test_array_input - ValueError: axes don't match array
FAILED scipy/stats/tests/test_multivariate.py::TestMatrixT::test_moments - ValueError: axes don't match array
FAILED scipy/stats/tests/test_multivariate.py::TestMatrixT::test_pdf_against_julia - ValueError: substring not found
FAILED scipy/stats/tests/test_multivariate.py::TestMatrixT::test_pdf_against_mathematica - ValueError: substring not found
FAILED scipy/stats/tests/test_multivariate.py::TestMatrixT::test_samples - ValueError: axes don't match array
FAILED scipy/stats/tests/test_multivariate.py::TestMatrixT::test_against_multivariate_t[row] - ValueError: substring not found
FAILED scipy/stats/tests/test_multivariate.py::TestMatrixT::test_against_multivariate_t[col] - ValueError: axes don't match array
= 11 failed, 67892 passed, 7339 skipped, 203 xfailed, 22 xpassed in 406.98s (0:06:46) =
```

example log: https://github.com/scipy/scipy/actions/runs/19393070829/job/55489304653

cc @steppi

lucascolley:
x-ref https://github.com/numpy/numpy/pull/23513#issuecomment-3538488751 for the stats failures, likely candidate for the upstream shift there

WarrenWeckesser:
For the `sf_error` issue, git bisect points to this numpy commit (see https://github.com/numpy/numpy/pull/29819):

```
a2d99b745e397e9e3ea0cf8a4033b7290bb3be3b is the first bad commit
commit a2d99b745e397e9e3ea0cf8a4033b7290bb3be3b
[...]

PR #24004, opened 2025-11-16
CI: use stable numpy release in free-threading job

rgommers:
This was an overdue change anyway, and also makes the job green again by avoiding the set of failures that's due to changes in `numpy`'s main branch visible on the pre-release job right now (see gh-23997).

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -480,10 +480,7 @@ jobs:
 
     - name: Install Python dependencies
       run: |
-        # Needs nightly numpy build until the 2.3.x release for Python 3.14 support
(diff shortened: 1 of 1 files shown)


.github/workflows/linux.yml
ev-br, line 474: Optional:
```suggestion
        python-version: '3.14t'
```

rgommers: The suggestion can't be applied for some reason, and it doesn't make a difference, so let's go ahead and merge this to make the job green?

ev-br:
The FT job is green, let's land it right away indeed. Thanks Ralf.

PR #24005, opened 2025-11-16
BUG: optimize: fix refcycle in `ScalarFunction`

andyfaff:
partial fix for #23080.

There is a very subtle refcycle in `ScalarFunction` that prevented multiprocessing resources being cleaned up, which triggered a "Too many open files" error.

### DIFF
--- a/scipy/optimize/_differentiable_functions.py
+++ b/scipy/optimize/_differentiable_functions.py
@@ -77,19 +77,29 @@ class _ScalarHessWrapper:
             self.nhev += 1
 
             if sps.issparse(self.H):
-                self._hess_func = self._sparse_callable
+                self._hess_func = "sparse_callable"
                 self.H = sps.csr_array(self.H)
             elif isinstance(self.H, LinearOperator):
-                self._hess_func = self._linearoperator_callable
+                self._hess_func = "linearoperator_callable"
             else:
                 # dense
-                self._hess_func = self._dense_callable
+                self._hess_func = "dense_callable"
                 self.H = np.atleast_2d(np.asarray(self.H))
         elif hess in FD_METHODS:
-                self._hess_func = self._fd_hess
+                self._hess_func = "fd_hess"
 
     def __call__(self, x, f0=None, **kwds):
-        return self._hess_func(np.copy(x), f0=f0)
+        match self._hess_func:
+            case "sparse_callable":
+                _h = self._sparse_callable
+            case "linearoperator_callable":
+                _h = self._linearoperator_callable
+            case "dense_callable":
+                _h = self._dense_callable
+            case "fd_hess":
+                _h = self._fd_hess
+
+        return _h(np.copy(x), f0=f0)
 
     def _fd_hess(self, x, f0=None, **kwds):
         self.H, dct = approx_derivative(
@@ -114,7 +124,6 @@ class _ScalarHessWrapper:
         self.nhev += 1
         self.H = self.hess(x, *self.args)
         return self.H
-
 
 class ScalarFunction:
     """Scalar function and its derivatives.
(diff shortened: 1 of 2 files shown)


andyfaff:
~First we add the test, then we add the fix.~

Added test and fix together by mistake.

andyfaff:
If there's going to be a 1.16.4 this should be backported.

scipy/optimize/_differentiable_functions.py
```diff
@@ -77,19 +77,29 @@ def __init__(
...
+            case "sparse_callable":
+                _h = self._sparse_callable
+            case "linearoperator_callable":
+                _h = self._linearoperator_callable
+            case "dense_callable":
+                _h = self._dense_callable
+            case "fd_hess":
+                _h = self._fd_hess
+
+        return _h(np.copy(x), f0=f0)
```
nickodell, line 102: This approach seems reasonable to me.

scipy/optimize/tests/test_optimize.py
```diff
@@ -3479,3 +3481,15 @@ def test_equal_bounds(self, workers, method):
...
+@pytest.mark.fail_slow(15)
+def test_multiprocessing_too_many_open_files_23080():
+    # https://github.com/scipy/scipy/issues/23080
+    rng = np.random.default_rng()
+    # should be enough fits to trigger the issue, which was due to a refcycle in
+    # ScalarFunction
+    for i in range(10):
+        x0 = rng.uniform(-20, 20, size=3)
+        with multiprocessing.Pool() as p:
+            _minimize_bfgs(rosen, x0, workers=p.map)
```
nickodell: This test will only fail if the user has a limit on open files set low enough. What would you think about an assert_deallocated test instead?

https://github.com/scipy/scipy/blob/d33e485e0d899296b2cf72f38f5280ca20617a1a/scipy/optimize/tests/test_differentiable_functions.py#L992

andyfaff: I've just had a look into this and couldn't get it to fail like it ought to:

i.e. whilst the following code illustrates the refcycle (no patch applied):
```
import multiprocessing
import numpy as np
from scipy.optimize import rosen
from scipy.optimize._differentiable_functions import ScalarFunction
import refcycle
import gc


def hess(x, *args):
	return 1


if __name__ == "__main__":
    rng = np.random.default_rng()
    
    for i in range(5):
        x0 = rng.uniform(-20, 20, size=3)
        with multiprocessing.Pool() as p:
            sf = ScalarFunction(rosen, x0, (), '2-point', hess, workers=2)
            sf.fun(x0+0.1)
            sf.grad(x0+0.01)
        graph = refcycle.garbage()
        print(graph)
        graph.export_image(f'garbage-{i}.svg')
        del graph
```

the following doesn't fail.
```python
from scipy._lib._gcutils import assert_deallocated
from scipy.optimize._differentiable_functions import ScalarFunction

with assert_deallocated(ScalarFunction, rosen, x0, (), '2-point', hess, workers=2) as f:
    f.fun(x0+0.1)
    f.grad(x0+0.01)
    del f
```

nickodell: Strictly speaking, it's not that the ScalarFunction isn't getting cleaned up, but one of its members, ScalarHessWrapper.

So if we use assert_deallocated on the `f._wrapped_hess` attribute, we can detect if it's getting cleaned up.

Alternatively, we can check that the Pool gets cleaned up after being used inside ScalarFunction. The Pool is what we really care about, after all.

I wrote a version of that here: https://github.com/scipy/scipy/compare/main...nickodell:scipy:njo-test-scalarhesswrapper

This correctly fails without the fix.

andyfaff: Using your much better test in a new commit. I verified that the test failed on main.

andyfaff:
Test fails are unrelated. I verified that the fix doesn't create refcycles by the new test, and by a manual script from the linked issue.

PR #24006, opened 2025-11-17
MAINT: removes a reference cycle in `optimize.differential_evolution`

andyfaff:
Related to #23080.
See cycle in https://github.com/scipy/scipy/issues/23080#issuecomment-3539620010

Script used to detect cycle:
```python
import multiprocessing
import numpy as np
from scipy.optimize import rosen, differential_evolution
import refcycle
import gc


if __name__ == "__main__":
    rng = np.random.default_rng()
    
    for i in range(10):
        x0 = rng.uniform(-20, 20, size=6)
        differential_evolution(rosen, [(-20, 20)]*6, polish=False)

        graph = refcycle.garbage()
        print(graph)
        print(len(graph))
        graph.export_image(f'garbage{i}.svg')
        del graph
        gc.collect()
```

### DIFF
--- a/scipy/optimize/_differentialevolution.py
+++ b/scipy/optimize/_differentialevolution.py
@@ -796,6 +796,7 @@ class DifferentialEvolutionSolver:
                     'currenttobest1exp': '_currenttobest1',
                     'best2exp': '_best2',
                     'rand2exp': '_rand2'}
+    __combined = _binomial | _exponential
 
     __init_error_msg = ("The population initialization method must be one of "
                         "'latinhypercube' or 'random', or an array of shape "
@@ -813,9 +814,9 @@ class DifferentialEvolutionSolver:
             # a callable strategy is going to be stored in self.strategy anyway
             pass
         elif strategy in self._binomial:
-            self.mutation_func = getattr(self, self._binomial[strategy])
+            pass
         elif strategy in self._exponential:
-            self.mutation_func = getattr(self, self._exponential[strategy])
+            pass
         else:
             raise ValueError("Please select a valid mutation strategy")
         self.strategy = strategy
@@ -1027,6 +1028,10 @@ class DifferentialEvolutionSolver:
         # rather than repeatedly creating it in _select_samples.
         self._random_population_index = np.arange(self.num_population_members)
         self.disp = disp
+
+    @property
+    def mutation_func(self):
+        return getattr(self, self.__combined[self.strategy])
 
     def init_population_lhs(self):
         """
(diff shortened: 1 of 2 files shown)


scipy/optimize/_differentialevolution.py
```diff
@@ -813,9 +814,9 @@ def __init__(self, func, bounds, args=(),
             # a callable strategy is going to be stored in self.strategy anyway
             pass
         elif strategy in self._binomial:
-            self.mutation_func = getattr(self, self._binomial[strategy])
+            pass
         elif strategy in self._exponential:
-            self.mutation_func = getattr(self, self._exponential[strategy])
+            pass
         else:
             raise ValueError("Please select a valid mutation strategy")
```
nickodell: Can you combine these three branches?

e.g. replace

```
        elif strategy in self._binomial:
            pass
        elif strategy in self._exponential:
            pass
        else:
            raise ValueError("Please select a valid mutation strategy")
```

with (not tested)

```
        elif strategy not in self.__combined:
            raise ValueError("Please select a valid mutation strategy")
```

nickodell (changes requested):

andyfaff:
BTW I did a cursory check for performance by using `%timeit` on a simple example. Timing didn't seem to change.
```
import multiprocessing
import numpy as np
from scipy.optimize import rosen, differential_evolution
rng = np.random.default_rng()
x0 = rng.uniform(-20, 20, size=6)
%timeit differential_evolution(rosen, [(-20, 20)]*6, polish=False)
```

nickodell (approved):

andyfaff:
The underlying issue would've been around since this code was first written.

tylerjereddy (approved):
I think the relevant CI jobs are green now (I checked the full array API suite locally on this branch via `spin test -j 12 -b all` on ARM Mac with i.e., `torch` installed, and it was fine).

The CI failures are unrelated and from gh-23997.

Mostly I'm happy to trust that Nick is pretty thorough with the refcycle analysis so that approval seems sufficient for correctness here.

scipy/optimize/tests/test__differential_evolution.py
```diff
@@ -1751,3 +1752,7 @@ def custom_strategy_fn(candidate, population, rng=None):
                 bounds,
                 strategy=custom_strategy_fn
             )
+
+    def test_reference_cycles(self):
+        with assert_deallocated(DifferentialEvolutionSolver, rosen, [(0, 10)]*2):
+            pass
```
tylerjereddy, line 1758: ah, looks like you pushed this in just as I was about to merge.

Anyway, confirming this fails before and passes after the source patch is applied for me locally on ARM Mac, so even better.

I suspect we *should* now let the CI flush through again though, since reference cycling testing can be tricky?

andyfaff:
I don't think we need to wait for the GPU task to go through.

PR #24008, opened 2025-11-17
BUG: fix interpreter crashes in special functions related to binomial distribution

### LINKED ISSUE #5503: Binomial CDF Errors
The following command correctly returns around 0.5.
scipy.stats.binom.cdf(3000000, 10000000, .3)

The probability that a binomial RV with p=0.3 is less than or equal to 3,000,000 in 10,000,000 draws is around 0.5.

The same should be true for X<=30,000,000 in n=100,000,000 draws.
scipy.stats.binom.cdf(30000000, 100000000, .3)
should return around 0.5, but it returns 0.57036.

I also tested scipy.special.betainc, since that's what is used for calculating Binomial's CDF, and it exhibits the same problem.

dschmitz89:
#### Reference issue
Closes https://github.com/scipy/scipy/issues/5503

With the recent boost update, we now get accurate CDF values for up to 10 billion trials. That must be enough to close https://github.com/scipy/scipy/issues/5503 as we cannot and should not try to achieve arbitrary precision levels wherever possible. I added the 10 billion trial test case to the already long list of tests for #5503 to guard against regressions.

While at it, I noticed that for even higher `n` the Python interpreter crashes due to an uncaught Boost error: `libc++abi: terminating due to uncaught exception of type std::out_of_range: basic_string`. As a quick fix, I added a simple error handling in the wrapper that can make it into the next release.

#### Additional information
We should improve our boost wrappers in a way that all functions use some `try/catch (...)` handling. Hopefully I get to that in the coming weeks.


scipy/special/tests/test_boost_ufuncs.py
```diff
@@ -62,3 +62,12 @@ def test_landau():
...
     _ = scu._ncx2_pdf(30, 1e307, 16)
+
+@pytest.mark.parametrize("k, n, p", [
+    (3*10**18, 10*10**18, 0.3),
+    (10**200, 10**210, 0.5),
+])
+def test_extreme_inputs(k, n, p):
+    # extreme inputs caused C++ exceptions 
+    # resulting in Python interpreter crashes
+    scu._binom_cdf(k, n, p)
```
tylerjereddy: I can reproduce the CI complaints on x86_64 Linux when I build this branch from source with latest NumPy and then downgrade to NumPy `1.26.4` before running this test.

I can make it pass with some type coercions on `1.26.4`,

```diff
--- a/scipy/special/tests/test_boost_ufuncs.py
+++ b/scipy/special/tests/test_boost_ufuncs.py
@@ -70,4 +70,4 @@ def test_gh22956():
 def test_extreme_inputs(k, n, p):
     # extreme inputs caused C++ exceptions 
     # resulting in Python interpreter crashes
-    scu._binom_cdf(k, n, p)
+    scu._binom_cdf(np.float64(k), np.float64(n), p)
```

Or, without any source changes at all when using `NPY_PROMOTION_STATE=weak` with the older NumPy. I think Matt has basically been skipping cases that require shims for older NumPy due to NEP50, so maybe we should do the same here using i.e., `numpy.lib.NumpyVersion`?

Circling back to the source patch itself--I confirmed that locally I see a Python interpreter crash on x86_64 Linux when I revert it and run your new test.

dschmitz89: Promoting to double precision seems fine here to me.

dschmitz89:
Turns out that only the binomial CDF was affected fortunately, not the negative binomial distribution (asserted by test). Nevertheless, we need to add proper error handling for everything we send to C++ land in future. Let me know if anything else is required.

dschmitz89:
CC @jzmaddock for visibility: I discovered that boost throws a `libc++abi: terminating due to uncaught exception of type std::out_of_range: basic_string` for extreme inputs in the binomial distribution CDF. We are catching the error now on the python side. Might not need fixing in boost itself though, it is a valid C++ exception.

jzmaddock:
> I discovered that boost throws a libc++abi: terminating due to uncaught exception of type std::out_of_range: basic_string for extreme inputs in the binomial distribution CDF. We are catching the error now on the python side. Might not need fixing in boost itself though, it is a valid C++ exception.

That seems strange to me (and sounds like a bug): we shouldn't be using never mind throwing from basic_string.  Do you have a test case?

tylerjereddy:
I don't quite follow the testing changes that were made in the most recent round of revisions--I no longer see an interpreter crash when I revert your source patch on Mac or on x86_64 Linux, whereas I was certainly seeing it when I reverted last night.

Even at high repeat count the test now seems stable without the patch: `spin test -t scipy/special/tests/test_boost_ufuncs.py -- --count=90`

I don't mind if the `stats`/`special` folks want to merge anyway though, just pointing that out from my crude mechanical perspective.

dschmitz89:
> I don't quite follow the testing changes that were made in the most recent round of revisions--I no longer see an interpreter crash when I revert your source patch on Mac or on x86_64 Linux, whereas I was certainly seeing it when I reverted last night.
> 
> Even at high repeat count the test now seems stable without the patch: `spin test -t scipy/special/tests/test_boost_ufuncs.py -- --count=90`
> 
> I don't mind if the `stats`/`special` folks want to merge anyway though, just pointing that out from my crude mechanical perspective.

Thanks for cross checking. You are correct, I fooled myself. Turns out the binomial SF, PPF and ISF were indeed all affected. Interestingly, the issue only occurs for very special inputs somehow. Will push the fix once the Github outage is over.

dschmitz89:
@jzmaddock Actually, boost reports the issue correctly, just our wrapping mechanism was flawed. The following reproducer gives `Error in function boost::math::ibeta: Series evaluation exceeded 1000000 iterations, giving up now`.

<details><summary>Reproducer</summary>
<p>

```c++
#include <iostream>
#include <exception>
#include "boost/math/distributions.hpp"

int main() {
    try {
        // Parameters that cause the exception
        double k = 3e18;
        double n = 10e18;
        double p = 0.3;
        
        std::cout << "Testing Boost binomial CDF with k=" << k 
                  << ", n=" << n << ", p=" << p << std::endl;
        
        // Create binomial distribution
        boost::math::binomial_distribution<double> dist(n, p);
        
        // This should trigger the std::out_of_range exception
        double result = boost::math::cdf(dist, k);
        
        std::cout << "Result: " << result << std::endl;
        
    } catch (const std::out_of_range& e) {
        std::cerr << "std::out_of_range exception caught: " << e.what() << std::endl;
        return 1;
    } catch (const std::exception& e) {
        std::cerr << "Exception caught: " << e.what() << std::endl;
        return 1;
    } catch (...) {
        std::cerr << "Unknown exception caught" << std::endl;
        return 1;
    }
    
    return 0;
}
``` 

</p>
</details> 

Do you think it is sensible to increase the maximum number of iterations via the policy?

lucascolley (approved):
I can reproduce Fatal Python errors from both tests on main, and the tests pass on this branch. The fix looks good to me, and the extra stats cases can only be a good thing. Thanks @dschmitz89 ! Is this ready to go in?

dschmitz89:
> I can reproduce Fatal Python errors from both tests on main, and the tests pass on this branch. The fix looks good to me, and the extra stats cases can only be a good thing. Thanks @dschmitz89 ! Is this ready to go in?

It is.

PR #24010, opened 2025-11-17
ENH: stats: added marginal probability distributions

### LINKED ISSUE #23901: ENH: stats: Implement marginal distributions for multivariate probability distributions

It doesn't appear that there is an easy way to plot the marginal distribution for multivariate probability distributions. For example, the scipy.stats.dirichlet distribution has a pdf method but no method for the marginal pdf of a given parameter. This could be done easily for the dirichlet and multivariate_normal distributions, but I'm not sure about others. 

A method which returns the marginal distribution for a multivariate probability distribution for a given parameter. 

_No response_

_No response_

JacobHass8:
Only added marginal distributions for stats.multivariate_normal, stats.multivariate_t, and stats.matrix_normal. Closes gh-23901.


mdhaber (commented):
I'm still not sure if I want to be the one to review this, but just a few comments that any reviewer would need to mention:
- It looks like these methods assume the distribution is frozen. I don't think we have any other methods that only work for frozen distributions. As unfortunate as the frozen/unfrozen distinction is, I'd assume this method should follow precedent. If you disagree, that's something that needs to be discussed.
- the multivariate normal implementation needs to work when the covariance is specified as an instance of [`Covariance`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.Covariance.html).
- input validation needs to be tested.

If I were the one to review this, I'd also suggest:
- restricting this first PR to one of the distributions. It's easier to work out the kinks on something small before doing 3x as much work.
- writing a generic marginal distribution function/class using `nquad` or `cubature` and using it in tests rather than including the integration code and loops in every test. I'd want to see a case in which the marginal distribution is also multivariate. If I do decide to review this, I would probably write this reference class for you to review and use in your tests.

*Update: here is a function to evaluate the marginal PDF of a multivariate distribution.*
<details>

```python3
import numpy as np
from scipy import stats, integrate

def marginal_pdf(X, X_ndim, dimensions, x):
    i_marginalize = np.ones(X_ndim, dtype=bool)
    i_marginalize[dimensions] = False

    def g(z):
        y = np.empty((z.shape[0], x.shape[0], X_ndim))
        y[..., i_marginalize] = z[:, np.newaxis, :]
        y[..., ~i_marginalize] = x
        return X.pdf(y)

    inf = np.full(X_ndim - len(dimensions), np.inf)
    return integrate.cubature(g, -inf, inf).estimate

rng = np.random.default_rng(413911473)
mean = rng.standard_normal(3)
A = rng.standard_normal((3, 3))
cov = np.dot(A, A.transpose())

X = stats.multivariate_normal(mean, cov)  # frozen distribution
X_ndim = 3  # dimensionality of frozen distribution
dimensions = [1, 2]  # (non-negative) indices of dimensions being retained
x = np.asarray([[1.5, 2.5], [2.5, 2.5], [3.5, 2.5]])  # points at which to evaluate marginal PDF

Y = X.marginal(dimensions)
ref = marginal_pdf(X, X_ndim, dimensions, x)
res = Y.pdf(x)
np.testing.assert_allclose(ref, res)
```

</details>

Also, in the future, please keep your fork named SciPy and change the name of the *branch* from `main` to some unique string that is related to what you're doing, e.g. `gh23901` or `multivariate_normal_marginal`. This will make it easier for the reviewer to check out your branch.

JacobHass8:
Thanks for the great feedback. Perhaps, I'll restrict to only modifying the `stats.multivariate_normal` distribution then! I'll make a couple more commits to address your feedback.

JacobHass8:
> I'm still not sure if I want to be the one to review this, but just a few comments that any reviewer would need to mention:
> 
>     * It looks like these methods assume the distribution is frozen. I don't think we have any other methods that only work for frozen distributions. As unfortunate as the frozen/unfrozen distinction is, I'd assume this method should follow precedent. If you disagree, that's something that needs to be discussed.
> 
>     * the multivariate normal implementation needs to work when the covariance is specified as an instance of [`Covariance`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.Covariance.html).
> 
>     * input validation needs to be tested.
> 
> 
> If I were the one to review this, I'd also suggest:
> 
>     * restricting this first PR to one of the distributions. It's easier to work out the kinks on something small before doing 3x as much work.
> 
>     * writing a generic marginal distribution function/class using `nquad` or `cubature` and using it in tests rather than including the integration code and loops in every test. I'd want to see a case in which the marginal distribution is also multivariate. If I do decide to review this, I would probably write this reference class for you to review and use in your tests.
> 
> 
> _Update: here is a function to evaluate the marginal PDF of a multivariate distribution._
> 
> Also, in the future, please keep your fork named SciPy and change the name of the _branch_ from `main` to some unique string that is related to what you're doing, e.g. `gh23901` or `multivariate_normal_marginal`. This will make it easier for the reviewer to check out your branch.

Thanks again for the comments. I've made all the necessary changes and I think it's ready for review. I wasn't sure how to handle the github side of things and made several commits in response to your comments. Is this normal, or should pull requests only be a single commit?

mdhaber:
Several commits is fine - it's encouraged if each does a self-contained step and has a meaningful commit message (like these, it seems). And I also much prefer this to rewriting the history and force-pushing, especially after review has started, so I can easily see what has changed. The only thing is that it's best to save pushing these commits for when you need CI to run. It doesn't matter so much in this PR - since it's your first, a maintainer has to approve CI runs - but in the future, each push will start CI, which takes a lot of computational resources. (See [here](https://docs.scipy.org/doc/scipy/dev/contributor/continuous_integration.html#skipping) if you want to push but don't need to run CI.)

Thanks for the changes - I think those will really help this along. I think I will take a more detailed look shortly, but just a few more suggestions you are welcome to think about in the meantime: 
- Consider using `pytest.mark.parametrize` to test distinct cases, for instance, use with unfrozen vs frozen distribution, distribution covariance specified as an array vs `Covariance` object, and marginalizing out one vs a few variables. 
- I like that you generated the distribution parameters at random. I think it would be great if you generated the test points at random too. (There was no significance to the points I choose in my example code, and if I were being more careful I would have generated them at random, too.)
- Please use the [context form of `pytest.raises` with a `match` keyword argument](https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions). This is preferred in modern tests because it's more readable and more specific. (I've caught many bugged error messages that went undetected because they used the old-school function form without a `match`.)
- Are there other edge cases to test? (e.g. what happens when all variables are marginalized out? When none are? Perhaps these are already covered - if so, I'll figure it out when I review more closely, and no need to respond.)
- There are many flavors of `Covariance` objects - is it possible/useful (in some cases, anyway) for the marginal distribution to preserve the type that is used with the original distribution?

Thanks!

JacobHass8:
> Thanks for the changes - I think those will really help this along. I think I will take a more detailed look shortly, but just a few more suggestions you are welcome to think about in the meantime:
> 
>     * Consider using `pytest.mark.parametrize` to test distinct cases, for instance, use with unfrozen vs frozen distribution, distribution covariance specified as an array vs `Covariance` object, and marginalizing out one vs a few variables.

I'm still getting familiar with `pytest`. This is a great feature. 

>     * I like that you generated the distribution parameters at random. I think it would be great if you generated the test points at random too. (There was no significance to the points I choose in my example code, and if I were being more careful I would have generated them at random, too.)

That's an easy enough fix. 
 
>     * Please use the [context form of `pytest.raises` with a `match` keyword argument](https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions). This is preferred in modern tests because it's more readable and more specific. (I've caught many bugged error messages that went undetected because they used the old-school function form without a `match`.)

Good idea. I've added that.

>     * Are there other edge cases to test? (e.g. what happens when all variables are marginalized out? When none are? Perhaps these are already covered - if so, I'll figure it out when I review more closely, and no need to respond.)

When no variables are marginalized (all dimensions are retained), my function returns the original `multivariate_normal_frozen` object. I'm not sure how to handle if all variables are marginalized. I think the best solution would be to throw an error, but I could also return 1 since you're integrating over a probability distribution. 

>     * There are many flavors of `Covariance` objects - is it possible/useful (in some cases, anyway) for the marginal distribution to preserve the type that is used with the original distribution?

I'm not exactly sure. It doesn't seem like the `Covariance` object type is saved when instantiating a `matrix_normal_frozen` object. I'm using the `_process_parameters` function so this is consistent with how `matrix_normal_frozen` handles `Covariance` object types.

mdhaber:
> I think the best solution would be to throw an error, but I could also return 1 since you're integrating over a probability distribution.

I think that the "correct" thing would be to return a 0d multivariate normal distribution, but since we don't have an implementation for it and adding it would be a waste of time, I'm happy with an error.

> It doesn't seem like the Covariance object type is saved when instantiating a `matrix_normal_frozen` object.

`multivariate_normal_frozen` objects store the original covariance object, so you can distinguish between them; for example:

```python3
import numpy as np
from scipy import stats
cov1 = np.diag([1, 2])
cov2 = stats.Covariance.from_diagonal([1, 2])
cov3 = stats.Covariance.from_precision(np.diag([1, 0.5]))
X1 = stats.multivariate_normal(cov=cov1)
X2 = stats.multivariate_normal(cov=cov2)
X3 = stats.multivariate_normal(cov=cov3)
X1.cov_object, X2.cov_object, X3.cov_object
# (<scipy.stats._covariance.CovViaPSD at 0x7ad3248887d0>,
# <scipy.stats._covariance.CovViaDiagonal at 0x7ad324888f50>,
# <scipy.stats._covariance.CovViaPrecision at 0x7ad324aaf8c0>)
```

An implementation strategy would be for each subclass of `Covariance` to implement a method for returning an object that represents the covariance of a marginal distribution.

Let's not worry about that in this PR, but I'll suggest a comment to add to the code about this enhancement idea.

JacobHass8:
> > I think the best solution would be to throw an error, but I could also return 1 since you're integrating over a probability distribution.
> 
> I think that the "correct" thing would be to return a 0d multivariate normal distribution, but since we don't have an implementation for it and adding it would be a waste of time, I'm happy with an error.

I've added a check to throw an error when trying to marginalize all dimensions.

scipy/stats/_multivariate.py
```diff
@@ -877,6 +877,53 @@ def fit(self, x, fix_mean=None, fix_cov=None):
...
+            raise ValueError(msg)
+
+        if not np.issubdtype(dims.dtype, np.integer):
+            msg = ("Elements of `dimensions` must be integers - the indices "
+                   "of the marginal variables being retained.")
+            raise ValueError(msg)
+
+        original_dims = dims.copy()
+
+        dims[dims < 0] = n + dims[dims < 0]
```
mdhaber: This can mutate user input. If gh-9932 has the same issue, let's fix that at the same time. TBH, can we factor out some of this input validation into a function and share it between the two classes? Probably the same stuff is going to be needed for `multivariate_t`, anyway.

scipy/stats/tests/test_multivariate.py
```diff
@@ -1226,6 +1237,59 @@ def test_fit_fix_cov(self):
...
+
+    def test_marginal_dimensions(self):
+        rng = np.random.default_rng(413911473)
+        mean = rng.standard_normal(3)
+        A = rng.standard_normal((3, 3))
+        cov = np.dot(A, A.transpose())        
+
+        X = multivariate_normal(mean, cov)
+
+        with pytest.raises(ValueError, match=r"Dimensions \[3\] are invalid .*"):
```
mdhaber: Nitpick, but if these are consistently ordered - especially if they are ordered as they are in the input validation - it's easier to see what is and isn't checked.

scipy/stats/tests/test_multivariate.py
```diff
@@ -1226,6 +1237,59 @@ def test_fit_fix_cov(self):
...
+    @pytest.mark.parametrize('X_ndim', [3])
+    @pytest.mark.parametrize('dimensions', [[1], 
+                                            [0, 1]])
+    @pytest.mark.parametrize('frozen', [True, False])
+    @pytest.mark.parametrize('cov_object', [True, False])
+    def test_marginal_distribution(self, X_ndim, dimensions, frozen, cov_object):
+        rng = np.random.default_rng(413911473)
+        mean = rng.standard_normal(X_ndim)
+        A = rng.standard_normal((X_ndim, X_ndim))
+        cov = np.dot(A, A.transpose())
```
mdhaber: Not necessary, but some would prefer:
```suggestion
        cov = A @ A.T
```
here and below.

scipy/stats/tests/test_multivariate.py
```diff
@@ -1428,8 +1492,7 @@ def test_samples(self):
               [2.59428444080606, 5.79987854490876]]]
         )
         assert_allclose(actual, expected)
-
-
+
```
mdhaber:
```suggestion


```
Two blank lines between top-level functions/classes is correct. We also don't want whitespace on an otherwise blank line.

scipy/stats/tests/test_multivariate.py
```diff
@@ -3733,7 +3796,6 @@ def test_logpdf_df_inf_gh19930(self):
         assert_allclose(Y.logpdf(x), X.logpdf(x))
         assert_allclose(Y.pdf(x), X.pdf(x))
```
mdhaber, line 3820:
```suggestion


```

scipy/stats/tests/test_multivariate.py
```diff
@@ -1226,6 +1237,59 @@ def test_fit_fix_cov(self):
...
+        mean = rng.standard_normal(3)
+        A = rng.standard_normal((3, 3))
+        cov = np.dot(A, A.transpose())        
+
+        X = multivariate_normal(mean, cov)
+
+        with pytest.raises(ValueError, match=r"Dimensions \[3\] are invalid .*"):
+            X.marginal(3)
+
+        with pytest.raises(ValueError, match=r"All elements of `dimensions` must be unique."):
```
mdhaber: Many of these lines exceed the character limit. Please run `spin lint` and resolve style checks.

scipy/stats/tests/test_multivariate.py
```diff
@@ -1226,6 +1237,59 @@ def test_fit_fix_cov(self):
                                                      ).sum()
         assert logp_perturbed < logp_fix
 
+    @pytest.mark.parametrize('X_ndim', [3])
+    @pytest.mark.parametrize('dimensions', [[1], 
+                                            [0, 1]])
```
mdhaber: Valid negative indices are not currently testsed, so how about:
```suggestion
                                            [-1, 1]])
```
Then again, some tests fail when I make that change locally. So does `[2, 1]`, but `[0, -1]` and `[1, 2]` pass. Sorting the dimensions after standardizing them to be positive fixes the failing cases, but I don't think that's necessarily what we want to do. Please check what happens when `dimensions` are out of order in `gaussian_kde` - does that re-order the dimensions of the returned object in a consistent/usable way? If so, we can consider it a "feature", preserve that behavior here, and adjust the reference implementation to follow suit.

Also would be fine with me to keep this on the same line.

scipy/stats/tests/test_multivariate.py
```diff
@@ -1226,6 +1237,59 @@ def test_fit_fix_cov(self):
...
+            Y = X.marginal(dimensions)
+            res = Y.pdf(x)
+        else: 
+            Y = multivariate_normal.marginal(dimensions, mean, cov)
+            res = Y.pdf(x)
+
+        ref = marginal_pdf(X, X_ndim, dimensions, x)
+        assert_allclose(ref, res)
+
+    def test_marginal_dimensions(self):
```
mdhaber: A common name for this sort of test might be `test_input_validation`. Another test, `test_special_cases`, might be good for testing e.g. all dimensions retained / none marginalized out.

mdhaber (commented):
This looks pretty close.

scipy/stats/_multivariate.py
JacobHass8: What line can mutate the user input? Is it lines 895 and 896 where we set the mean? I think having an input validation function would be smart. I can take a shot at writing a function `_marginal_dimension_validation` that takes in the `dimensions` parameter and number of degrees of freedom for the multivariate distribution

mdhaber: `dims[dims < 0] = n + dims[dims < 0]` mutates `dims` if `dims` was passed as an array by the user.

JacobHass8: Oh, I assumed that `np.atleast_1d` copied the array (https://numpy.org/doc/2.2/reference/generated/numpy.atleast_1d.html). I'll just make a copy of `dimensions`.

mdhaber: "Copies are made only if necessary."

```python3
import numpy as np
x = np.asarray([1, 2, 3])
x2 = np.atleast_1d(x)
x2[0] = 100
x  # array([100,   2,   3])
```

scipy/stats/tests/test_multivariate.py
JacobHass8: I've done a little testing and when the dimensions aren't sorted, the mean and covariance matrix are rearranged based on the order of the `dimension` array. I think this is the expected behavior and a "feature" as you noted.  The reason the tests are failing is the `marginal_pdf` function I'm comparing the result to does not rearrange the mean or covariance matrix. I'll work to fix that.

JacobHass8:
> This looks pretty close.

I've addressed all of the review comments. The `marginal` function and tests look a lot clearer now. Let me know what you think.

mdhaber:
@jorenham I think you mentioned on the ~~mailing list~~forum that you were interested in reviewing this one?

jorenham:
Yea that's right; thanks for the ping @mdhaber

scipy/stats/_multivariate.py
```diff
@@ -122,6 +122,51 @@ def _pinv_1d(v, eps=1e-5):
...
+        msg = ("Elements of `dimensions` must be integers - the indices "
+                "of the marginal variables being retained.")
+        raise ValueError(msg)
+
+    original_dims = np.copy(dims)
+
+    dims[dims < 0] = multivariate_dims + dims[dims < 0]
+
+    if len(np.unique(dims)) != len(dims):
+        msg = ("All elements of `dimensions` must be unique.")
```
jorenham: nitpick:

```suggestion
        msg = "All elements of `dimensions` must be unique."
```

scipy/stats/_multivariate.py
```diff
@@ -122,6 +122,51 @@ def _pinv_1d(v, eps=1e-5):
...
+    Returns
+    -------
+    dims : ndarray
+        Array of indices to marginalize
+    """
+    dims = np.copy(dimensions)
+    dims = np.atleast_1d(dims)
+
+    if len(dims) == 0:
+        msg = ("Cannot marginalize all dimensions.")
```
jorenham: nitpick:

```suggestion
        msg = "Cannot marginalize all dimensions."
```

scipy/stats/_multivariate.py
```diff
@@ -122,6 +122,51 @@ def _pinv_1d(v, eps=1e-5):
...
+        raise ValueError(msg)
+
+    if not np.issubdtype(dims.dtype, np.integer):
+        msg = ("Elements of `dimensions` must be integers - the indices "
+                "of the marginal variables being retained.")
```
jorenham: nitpick:

```suggestion
        msg = ("Elements of `dimensions` must be integers - the indices "
               "of the marginal variables being retained.")
```

scipy/stats/_multivariate.py
```diff
@@ -122,6 +122,51 @@ def _pinv_1d(v, eps=1e-5):
...
+        raise ValueError(msg)
+
+    original_dims = np.copy(dims)
+
+    dims[dims < 0] = multivariate_dims + dims[dims < 0]
```
jorenham: nitpick:

```suggestion
    dims[dims < 0] += multivariate_dims
```

scipy/stats/_multivariate.py
```diff
@@ -877,6 +922,29 @@ def fit(self, x, fix_mean=None, fix_cov=None):
...
 
+    def marginal(self, dimensions, mean=None, cov=1, allow_singular=False):
+        """Return a marginal multivariate normal distribution
+
+        Parameters
+        ----------
+        dimensions : int or 1-d array_like
+            The dimensions of the multivariate distribution corresponding
+            with the marginal variables, that is, the indices of the dimensions
+            that are being retained. The other dimensions are marginalized out.
```
jorenham, line 934: The `mean`, `cov`, and `allow_singular` parameters aren't documented

scipy/stats/_multivariate.py
```diff
@@ -877,6 +922,29 @@ def fit(self, x, fix_mean=None, fix_cov=None):
             cov = centered_data.T @ centered_data / n_vectors
         return mean, cov
 
+    def marginal(self, dimensions, mean=None, cov=1, allow_singular=False):
```
jorenham: How about making the parameters with default keyword-only? It's a lot easier to do this now than to change it later.

```suggestion
    def marginal(self, dimensions, *, mean=None, cov=1, allow_singular=False):
```

jorenham: I had to read the `dimensions` docs before I was able to understand what it did. If instead it would've been named `axis`, I think I would have immediately understood its purpose. So maybe we should change its name?

scipy/stats/_multivariate.py
```diff
@@ -877,6 +922,29 @@ def fit(self, x, fix_mean=None, fix_cov=None):
...
+        """Return a marginal multivariate normal distribution
```
jorenham: Hmm I don't think I understand. I thought that marginal distributions were always univariate, maybe even by definition?

scipy/stats/tests/test_multivariate.py
```diff
@@ -294,6 +294,24 @@ def _sample_orthonormal_matrix(n):
     u, s, v = scipy.linalg.svd(M)
     return u
 
+def marginal_pdf(X, X_ndim, dimensions, x):
+    # First sort input data based on order of dimensions
```
jorenham: Did you forget to add the comment for the second step?

scipy/stats/tests/test_multivariate.py
```diff
@@ -1226,6 +1243,71 @@ def test_fit_fix_cov(self):
                                                      ).sum()
         assert logp_perturbed < logp_fix
 
+    @pytest.mark.parametrize('X_ndim', [3])
+    @pytest.mark.parametrize('dimensions', [[1], 
+                                            [-1, 1]])
```
jorenham: nitpick

```suggestion
    @pytest.mark.parametrize('dimensions', [[1], [-1, 1]])
```

scipy/stats/_multivariate.py
```diff
@@ -122,6 +122,51 @@ def _pinv_1d(v, eps=1e-5):
...
+    dims[dims < 0] = multivariate_dims + dims[dims < 0]
+
+    if len(np.unique(dims)) != len(dims):
+        msg = ("All elements of `dimensions` must be unique.")
+        raise ValueError(msg)
+
+    i_invalid = (dims < 0) | (dims >= multivariate_dims)
+    if np.any(i_invalid):
+        msg = (f"Dimensions {original_dims[i_invalid]} are invalid "
+                f"for a distribution in {multivariate_dims} dimensions.")
```
jorenham: nitpick:

```suggestion
        msg = (f"Dimensions {original_dims[i_invalid]} are invalid "
               f"for a distribution in {multivariate_dims} dimensions.")
```

scipy/stats/tests/test_multivariate.py
```diff
@@ -294,6 +294,24 @@ def _sample_orthonormal_matrix(n):
     u, s, v = scipy.linalg.svd(M)
     return u
 
+def marginal_pdf(X, X_ndim, dimensions, x):
+    # First sort input data based on order of dimensions
+    dimensions = np.asarray(dimensions)
+    dimensions[dimensions < 0] = X_ndim + dimensions[dimensions < 0]
```
jorenham: nitpick:

```suggestion
    dimensions[dimensions < 0] += X_ndim
```

scipy/stats/tests/test_multivariate.py
```diff
@@ -294,6 +294,24 @@ def _sample_orthonormal_matrix(n):
...
+    i_marginalize[dimensions] = False
+
+    def g(z):
+        y = np.empty((z.shape[0], x.shape[0], X_ndim))
+        y[..., i_marginalize] = z[:, np.newaxis, :]
+        y[..., ~i_marginalize] = x
+        return X.pdf(y)
+
+    inf = np.full(X_ndim - len(dimensions), np.inf)
+    return cubature(g, -inf, inf).estimate
```
jorenham, line 318: this could do with a quick explanation in a comment

scipy/stats/tests/test_multivariate.py
```diff
@@ -1226,6 +1243,71 @@ def test_fit_fix_cov(self):
...
+            res = Y.pdf(x)
+
+        ref = marginal_pdf(X, X_ndim, dimensions, x)
+        assert_allclose(ref, res)
+
+    def test_marginal_input_validation(self):
+        rng = np.random.default_rng(413911473)
+        mean = rng.standard_normal(3)
+        A = rng.standard_normal((3, 3))
+        cov = np.dot(A, A.transpose())
```
jorenham:
```suggestion
        cov = A @ A.T
```

scipy/stats/_multivariate.py
```diff
@@ -877,6 +922,29 @@ def fit(self, x, fix_mean=None, fix_cov=None):
...
+            An object representing the marginal distribution.
+        """
+        params = self._process_parameters(mean, cov, allow_singular)
+        n, mean, cov_object = params
+        dims = _validate_marginal_input(dimensions, n)
+
+        mean = mean[dims]
+        cov = cov_object.covariance[np.ix_(dims, dims)]
+        
+        return multivariate_normal_frozen(mean, cov, allow_singular)
```
jorenham:
```suggestion
        return multivariate_normal_frozen(mean, cov, allow_singular)
```

scipy/stats/_multivariate.py
```diff
@@ -122,6 +122,51 @@ def _pinv_1d(v, eps=1e-5):
     return np.array([0 if abs(x) <= eps else 1/x for x in v], dtype=float)
 
 
+def _validate_marginal_input(dimensions, multivariate_dims):
+    """Determine if input dimensions can be marginalized.
```
jorenham: nitpick:
```suggestion
    """Determine if input dimensions can be marginalized.
```

scipy/stats/_multivariate.py
```diff
@@ -122,6 +122,51 @@ def _pinv_1d(v, eps=1e-5):
     return np.array([0 if abs(x) <= eps else 1/x for x in v], dtype=float)
 
 
+def _validate_marginal_input(dimensions, multivariate_dims):
+    """Determine if input dimensions can be marginalized. 
+
+    Parameters
+    ----------
+    dimensions : float, ndarray
+        Input dimensions to be marginalized
```
jorenham: nitpick:
```suggestion
        Input dimensions to be marginalized
```

scipy/stats/_multivariate.py
```diff
@@ -122,6 +122,51 @@ def _pinv_1d(v, eps=1e-5):
...
 
 
+def _validate_marginal_input(dimensions, multivariate_dims):
+    """Determine if input dimensions can be marginalized. 
+
+    Parameters
+    ----------
+    dimensions : float, ndarray
+        Input dimensions to be marginalized 
+
```
jorenham: nitpick:

```suggestion
```

scipy/stats/tests/test_multivariate.py
```diff
@@ -1226,6 +1243,71 @@ def test_fit_fix_cov(self):
...
+        
+        msg = r"Elements of `dimensions` must be integers."
+        with pytest.raises(ValueError, match=msg):
+            X.marginal([1.1, 2.0])
+    
+    def test_special_cases(self):
+        rng = np.random.default_rng(413911473)
+        mean = rng.standard_normal(3)
+        A = rng.standard_normal((3, 3))
+        cov = np.dot(A, A.transpose())
```
jorenham: nitpick:
```suggestion
        cov = A @ A.T
```

jorenham: Also, since this is the third time I've seen this bit of rng code, extracting it into a helper function might be a good idea.

scipy/stats/tests/test_multivariate.py
```diff
@@ -1226,6 +1243,71 @@ def test_fit_fix_cov(self):
...
+            X.marginal([2, -1])
+
+        with pytest.raises(ValueError, match=msg):
+            X.marginal([[0, 1]])
+        
+        msg = r"Elements of `dimensions` must be integers."
+        with pytest.raises(ValueError, match=msg):
+            X.marginal([1.1, 2.0])
+    
+    def test_special_cases(self):
```
jorenham: It'd help to make this test name a bit more specific

scipy/stats/_multivariate.py
```diff
@@ -1448,7 +1519,6 @@ def entropy(self):
         return self._dist._entropy(self.dims, self.rowpsd.log_pdet,
                                    self.colpsd.log_pdet)
 
-
```
jorenham, line 1451: no need to touch this

jorenham (commented):
Thanks for this! I've left some comments, most of them nitpicks.

scipy/stats/_multivariate.py
JacobHass8: No, a marginal distribution describes a subset of random variables. However, a marginal distribution is usually used to describe a single random variable rather than a subset. We're going with the subset definition since this is what was done in `stats.gaussian_kde`.

scipy/stats/_multivariate.py
JacobHass8: The `dimensions` naming comes from the `stats.gaussian_kde` convention. See #9932.

scipy/stats/tests/test_multivariate.py
mdhaber: What do you have in mind? I wrote this to compute reference values. It computes them according to the definition of a marginal distribution.

scipy/stats/tests/test_multivariate.py
mdhaber: I suggested this test name. If there were multiple special cases to parameterize over, it would have been fine. If we're only going to test one thing, it can be made more specific.

scipy/stats/_multivariate.py
mdhaber: Both of these things are determined by precedent. 

I'm not aware of any other multivariate distribution methods for which state parameters are keyword only. If you can find such a precedent, I guess we can change it - the multivariate distributions are pretty inconsistent anyway. But if not, please change it back and add the gripe to gh-16278.

~~As for `dimensions`: If these were the dimensions to integrate out, I think `axis` would have been a fine name.~~[^1] Throughout `stats`, `axis` identifies dimensions that a reducing operation is performed along, and they are removed from the output. Here, these `dimensions` are the only ones that are preserved; all the others are reduced away by integration. There may be other perspectives that would make `axis` seem like a better name, but to avoid confusion, we used a different name.

[^1]: Actually it still would have been a bad name. `axis` is used to refer to an axis (integer in `range(ndim)`) of an array. Here, we expect a list of integers, and more importantly, the meaning of these integers is more closely associated with the *indices* of an array (integers in `range(len(mean))`, say) than the *axes* of an array. And that's all besides the fact that we're talking about an object that represents a multivariate distribution, not an array.

scipy/stats/tests/test_multivariate.py
jorenham: > What do you have in mind?

I don't know that definition by heart, so it would help to informally describe that definition, or otherwise link to it.

scipy/stats/tests/test_multivariate.py
jorenham: `test_multivariate_special_cases`?

scipy/stats/_multivariate.py
jorenham: Mathematically it makes complete sense to call it `dimensions`. It's just that in numpy, the term "dimensions" is only ever used for `ndim`, as in "the number of". So my initial assumption was that `dimensions` was something along those lines. But in actually describes the *indices* of the dimensions. At least in NumPy, those have always been referred to as "axes". 
And I don't think that these semantic differences outweigh the semantic overlap. So, at least for me, it's (a lot) more confusing when called `dimensions` and not `axis`/`axes`.
Also, in NumPy the `axis` semantics aren't as consistent as you might think. For example, `axis=None` can mean completely different things, and some functions accept passing a tuple to `axis`, while others don't. So I don't think that renaming it as `axis` would be more confusing than that.

scipy/stats/tests/test_multivariate.py
mdhaber: In the file `test_multivariate` and in a class `TestMultivariateNormal`, I am not sure that would add much information. I could see `test_marginal_special_cases`, since it's right after `test_marginal_distribution` and `test_marginal_input_validation`.

scipy/stats/_multivariate.py
mdhaber: Well, NumPy is welcome to do that. But I'd prefer for `scipy.stats` to be consistent with itself where possible.

scipy/stats/tests/test_multivariate.py
jorenham: `test_marginal_special_cases` sounds good :ok_hand:

scipy/stats/_multivariate.py
JacobHass8: I think `%(_mvn_doc_default_callparams)s` should address this issue.

scipy/stats/_multivariate.py
jorenham: There are more functions in `scipy.stats` that use `axis` then there are that use `dimensions`. So at the very least it wouldn't be *in*consistent.

mdhaber: I don't see how that follows.

The name `axis` consistently refers to the same concept throughout `scipy.stats`. It is an integer in `range(ndim - 1)` used to refer to an axis of a array with `ndim` dimensions. 

The parameter we're discussing is entirely different. It is not even referring to an aspect of an array, really - it's referring to an object that represents a multivariate distribution. But when it is used with an array, it will be used as an *index along a particular axis* (e.g. the last axis of `mean`).

I can't get behind the idea of repurposing the name `axis` for something with an entirely different meaning. I have a hard time believing that a user would immediately understand its purpose had we used that name instead, especially if the definition of a marginal distribution was not already understood (https://github.com/scipy/scipy/pull/24010#discussion_r2551100476, https://github.com/scipy/scipy/pull/24010#discussion_r2553280679).

In the original PR, [I suggested `marginal_variables`](https://github.com/scipy/scipy/pull/9932#issuecomment-1211375216). To help move the PR along, I [compromised on `dimensions`](https://github.com/scipy/scipy/pull/9932#discussion_r943227172). So I can see a case for renaming, and you're welcome to open an issue about that. Just not to `axis`.

In the meantime, I'm going to consider this resolved, since this PR is just following precedent.

mdhaber:
With the exception of the controversial https://github.com/scipy/scipy/pull/24010#discussion_r2575417139, looks like this has incorportated all the feedback. Thanks @JacobHass8! And thanks @jorenham for the review.

jorenham:
Thanks @JacobHass8 @mdhaber !

JacobHass8:
Thanks for the review @mdhaber and @jorenham!

JacobHass8:
Could I open another PR for `stats.multivariate_t`? I think that most of the changes would be the same.

PR #24011, opened 2025-11-18
ENH: stats: Efficient random sampling for the Zipfian distribution.

WarrenWeckesser:
Implement a rejection algorithm for sampling from the Zipfian distribution.

For background:

* Some notes on the implementation: https://github.com/WarrenWeckesser/experiments/tree/main/python/numpy/random-cython/zipfian/docs
* Some notes and scripts on validating the implementation: https://github.com/WarrenWeckesser/experiments/tree/main/python/scipy/test-zipfian-distr
* The handling of the arguments in the Cython function `zipfian_rvs()` (broadcasting for arrays, separate handling of scalar parameters) is patterned after similar methods in the `Generator` class in `numpy.random`. See, for example, the implementation of `Generator.binomial()` in [`_generator.pyx`](https://github.com/numpy/numpy/blob/main/numpy/random/_generator.pyx).

Closes gh-23487.

### DIFF
--- a/scipy/stats/_discrete_distns.py
+++ b/scipy/stats/_discrete_distns.py
@@ -23,6 +23,7 @@ from ._biasedurn import (_PyFishersNCHypergeometric,
                          _PyWalleniusNCHypergeometric,
                          _PyStochasticLib3)
 from ._stats_pythran import _poisson_binom
+from ._zipfian_rvs import zipfian_rvs as _zipfian_rvs
 
 
 class binom_gen(rv_discrete):
@@ -1469,6 +1470,21 @@ class zipfian_gen(rv_discrete):
               - 3*Hna1**4) / mu2n**2
         g2 -= 3
         return mu1, mu2, g1, g2
+
+    def _rvs(self, a, n, size=None, random_state=None):
+        # This method should only be called by the public function rvs()
+        # with random_state set to an instance of np.random.RandomState or
+        # np.random.Generator, and both of those have a _bit_generator
+        # attribute, so despite the default value shown in the signature,
+        # we can safely access that attribute.
+        bitgen = random_state._bit_generator
+        a_f64 = a.astype(np.float64)
+        # _argcheck() & rvs() will have ensured that all the values in `n`
+        # can be cast safely to int64. The actual dtype of `n` might be some
+        # other integer or floating point type.
+        n_int64 = n.astype(np.int64)
+
+        return _zipfian_rvs(bitgen, a=a_f64, n=n_int64, size=size)
 
 
 zipfian = zipfian_gen(a=1, name='zipfian', longname='A Zipfian')
--- a/scipy/stats/_zipfian_rvs.pyx
+++ b/scipy/stats/_zipfian_rvs.pyx
@@ -1,0 +1,301 @@
+#cython: language_level=3
+"""
+This module defines the function zipfian_rvs(bitgen_t gen, a, n, size=None)
+that generates variates from the Zipfian distribution (scipy.stats.zipfian)
+using a rejection method.
+"""
+
+# C standard library...
+from libc.math cimport trunc, pow
+from libc.stdint cimport int64_t
+
+# Cython...
+cimport cython
+
+# CPython...
+from cpython.pycapsule cimport PyCapsule_IsValid, PyCapsule_GetPointer
+from cpython cimport PyFloat_AsDouble
+
+# NumPy...
+import numpy as np
+cimport numpy as cnp
+from numpy.random cimport bitgen_t
+
+# SciPy Cython API from scipy.special...
+from scipy.special.cython_special cimport boxcox, inv_boxcox
+
+cnp.import_array()
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+cdef verify_equal_shapes(cnp.broadcast it, cnp.ndarray out):
+    # Check that the shape of the iterator `it` is the same as that of
+    # the ndarray `out`.
+
+    cdef bint valid_shapes = True
+    cdef cnp.npy_intp i
+    cdef cnp.npy_intp it_ndim = cnp.PyArray_MultiIter_NDIM(it)
+    cdef cnp.npy_intp out_ndim = cnp.PyArray_NDIM(out)
+
+    if it_ndim != out_ndim:
+        valid_shapes = False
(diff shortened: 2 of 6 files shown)


charris:
What is the basic idea here? If the maximum value can be set, it is possible to speed things up many thousands of times, but I had the impression you already did that...

WarrenWeckesser:
> What is the basic idea here?

I have some notes about it here: https://github.com/WarrenWeckesser/experiments/tree/main/python/numpy/random-cython/zipfian/docs

> If the maximum value can be set, it is possible to speed things up many thousands of times, ...

I may have missed something obvious--what did you have in mind?  Don't forget that the upper bound `n` can be large; a million or so would not be unusual for someone modeling word distribution in a body of text.

Here's a timing example from this PR:

```
In [1]: from scipy.stats import zipfian

In [2]: a = 0.5

In [3]: n = 10000000

In [4]: %timeit zipfian.rvs(a, n)
14.3 μs ± 22.4 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [5]: %timeit zipfian.rvs(a, n, size=1000)
64.3 μs ± 74 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

The `.rvs()` method has a lot of overhead.  This is amortized in the second example, where the average time per variate is about 64 nanoseconds.  (Without the overhead the actual time per variate is about 50 ns on my machine.)

charris:
The reason (IIRC) it is so slow is that the large values are so very rare that all random samples are rejected. I note that some implementations have an argument of where to truncate the series and recompute the distribution. Note that there was an issue long ago in NumPy where I demonstrated that.

That said, I'm not sure we are talking about the same thing, so that is why I was asking.

EDIT: NumPy was using an upper bound of the max value of int64. It was a long time ago, so I might not recall exactly.

WarrenWeckesser:
To be clear, this PR is not replacing the use of NumPy's `zipf` method with our own implementation.

SciPy's `zipfian` distribution is not the same as the distribution implemented in NumPy's `zipf` method.  Well, being pedantic, NumPy's `zipf` is effectively a truncation of the ideal distribution, as you noted, but the intent of it is to model the case where the support is all positive integers.  NumPy's `zipf` requires $a > 1$, because the infinite series that defines the normalization constant in that case does not converge for $a \le 1$. In SciPy we have [`scipy.stats.zipf`](https://scipy.github.io/devdocs/reference/generated/scipy.stats.zipf.html) for that, and it uses the `zipf` method from NumPy to generate random variates.

[`zipfian`](https://scipy.github.io/devdocs/reference/generated/scipy.stats.zipfian.html) is the (explicitly) truncated case, where the parameter $n \ge 1$ defines the upper bound of the support.  Our implementation in SciPy requires $a \ge 0$. When $a = 0$, the distribution becomes the uniform distribution.

The slowness issue fixed by this PR is discussed in https://github.com/scipy/scipy/issues/23487.

WarrenWeckesser:
Out of draft, ready to review.

I think everyone is busy finishing up the existing 1.17 milestone PRs and issues, so I suspect this won't get much attention for a while.

If this does eventually get merged, the new tests should probably be marked XSLOW before merging.  The time for each set of parameters in those tests would be reasonable as a slow test, but the net result of running all the parameter cases puts it into XSLOW territory.

WarrenWeckesser:
I pushed an update that removed the shims that were required to build with NumPy 1.26.

dschmitz89 (commented):
I had a look at your notes about the algorithm. This was very helpful, thanks for that. The approach is correct and the rejection sampler will likely never fail. The performance improvement is also great.

Before formally requesting changes I would like to understand two things:
* do we need to handle the reshaping of the samples in Cython or could we do this later in NumPy?
* Do we need the additional tests? The generic distribtuion tests should ideally already check if we generate samples from the distribution we claims to do

chrisb83:
Hi Warren,

thanks for looking into this topic, it is a nice application of the rejection method. 

Note that we have a general method to sample from discrete distributions (`scipy.stats.sampling.DiscreteAliasUrn`), maybe it is helpful for testing:

```

from scipy.stats.sampling import DiscreteAliasUrn
import numpy as np
from scipy import stats
import timeit

n, a = 400, 1.05

urng = np.random.default_rng()

ps = [j**(-a) for j in range(1, n+1)]
gen = DiscreteAliasUrn(ps, domain=(1,n+1), random_state=urng)

# test speed against current approach
N = int(1e3)

t = timeit.timeit(lambda: gen.rvs(size=N), number=10)
print(f"Average UNURAN: {t/10:.4f}s over 10 runs")

gen2 = stats.zipfian(a, n)
t = timeit.timeit(lambda: gen2.rvs(size=N), number=10)
print(f"Average scipy: {t/10:.4f}s over 10 runs")

Average UNURAN: 0.0001s over 10 runs
Average scipy: 0.6977s over 10 runs
```

WarrenWeckesser:
@chrisb83, thanks for taking a look.

One reason this has been languishing for so long is exactly the issue you brought up.  I did some preliminary tests and found that the aliasing method was very effective, and went down a bit of a rabbit hole related to this.  I suspect there is a large range of values of `a` and `n` where the aliasing method could be used as the main algorithm behind `zipfian.rvs()`. I still need to map out where the rejection method wins.

And thanks @dschmitz89 for taking a look back in January.  I haven't forgotten about it!

chrisb83:
I am not sure if it will be possible to integrate the aliasing method into the existing setup. It has a small setup cost of a few milliseconds that is not too relevant if you run it once and then generate a large sample. But if you want to generate small samples, this will dominate the runtime. Worst case is N different parameters with one sample each, then you run the setup N times. So even if you conclude that aliasing it better for large samples, say for some parameters in an internval (a,b), then it would be better to use the rejection method in this scenario. 

another complication is that in case of frozen distributions, you would need to store the generator that you create with `rng = DiscreteAliasUrn(pv, random_state=urng)` for a given probability vector somehow for re-use.

that is what I remember from my consideration when I implemented #[FastGeneratorInversion](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.sampling.FastGeneratorInversion.html#scipy.stats.sampling.FastGeneratorInversion) which only supports scalar parameters.

So that could be an argument to proceed with the current implemetation based on the rejection method only which should already a massive speed.up

PR #23987, opened 2025-11-14
CI: reduce cache usage in GitHub Actions

rgommers:
The primary change is using release builds, which are 3x-4x smaller than debug builds, hence adding `--release` shrinks the cache usage on GitHub Actions a lot where `ccache` is used.
    
Caching the downloaded packages from PyPI isn't worth it given the 10 GB org-wide limit.

xref https://github.com/scipy/scipy/issues/21491#issuecomment-3314757541 for context

Note: there is more to gain along these lines. Using `--release` in conda/pixi envs is complicated by the compiler activation issue (it adds `-O2` which clashes with `--release`). This is better left alone until conda-forge finally fixes this issue, since there's some recent motion there.

PR #24016, opened 2025-11-18
CI: run cache-heavy OneAPI job on main every 2 days to optimize cache usage

rgommers:
This will ensure that the GitHub cache is populated better, so CI jobs don't save their own cache per job. PRs from forks can use cache entries created by a previous run on the `main` branch, but not those created by another PR.

This particular job is by far the most cache space-hungry one, with one entry of 3.1 GB and another of 670 MB.

See gh-21491 for more details.

Also reduce the frequency of wheel build runs from daily to every other day. It seems unnecessary to run daily, that's a very heavy CI load nowadays for no real reason. The NumPy nightlies refresh every 4 days for example. Every 2 days is frequent enough.

EDIT: note that I changed only one job on purpose, to verify that it really works. Some of the other jobs with high cache space usage are being optimized in gh-23987.

### DIFF
--- a/.github/workflows/linux_intel_oneAPI.yml
+++ b/.github/workflows/linux_intel_oneAPI.yml
@@ -8,6 +8,15 @@ on:
     branches:
       - main
       - maintenance/**
+  schedule:
+  #        ┌───────────── minute (0 - 59)
+  #        │  ┌───────────── hour (0 - 23)
+  #        │  │  ┌───────────── day of the month (1 - 31)
+  #        │  │  │  ┌───────────── month (1 - 12 or JAN-DEC)
+  #        │  │  │  │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
+  #        │  │  │  │ │
+  - cron: "9  9 2/2 * *"
+  push:
 
 permissions:
    contents: read  # to fetch code (actions/checkout)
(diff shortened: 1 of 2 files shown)


lucascolley (approved):
LGTM! Any reason for the difference between `2/2` and `*/2`?

rgommers:
> Any reason for the difference between `2/2` and `*/2`?

Running the wheel build jobs on odd days and the regular CI jobs on even days, to avoid congestion.

lucascolley:
smart!

.github/workflows/linux_intel_oneAPI.yml
```diff
@@ -8,6 +8,15 @@ on:
...
       - maintenance/**
+  schedule:
+  #        ┌───────────── minute (0 - 59)
+  #        │  ┌───────────── hour (0 - 23)
+  #        │  │  ┌───────────── day of the month (1 - 31)
+  #        │  │  │  ┌───────────── month (1 - 12 or JAN-DEC)
+  #        │  │  │  │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
+  #        │  │  │  │ │
+  - cron: "9  9 2/2 * *"
+  push:
```
rgommers, line 19: Oops copy-paste issue here, one spurious line. I have no idea why this didn't turn CI red on this PR, but the job is broken now - will fix asap.

rgommers: Cleaned up in commit abadaf6532

PR #24018, opened 2025-11-18
ENH: stats.binomtest: vectorize implementation

mdhaber:
#### Reference issue
Toward gh-20544

This vectorizes `scipy.stats.binomtest` in preparation for translation to the array API.

#### Additional information
~~I'm wondering whether we should keep the existing implementation as-is for 1D scalars and only transition to this implementation for array input? There is actually a lot less overhead than I feared (maybe 4x execution time for scalar input), but it does away with the explicit error messages and just returns NaNs for invalid input.~~ No.


scipy/stats/_binomtest.py
```diff
@@ -115,48 +115,24 @@ def proportion_ci(self, confidence_level=0.95, method='exact'):
...
+    alpha = ((1 - confidence_level) / 2 if alternative == 'two-sided'
+             else 1 - confidence_level)
+
+    plow = (np.zeros(k.shape, dtype=np.float64) if alternative == 'less' else
+            find_root(lambda p, k, n: binom.sf(k-1, n, p) - alpha, init, args=args).x)
+    phigh = (np.ones(k.shape, dtype=np.float64) if alternative == 'greater' else
+             find_root(lambda p, k, n: binom.cdf(k, n, p) - alpha, init, args=args).x)
+
+    plow = np.where(k == 0, 0.0, plow)
+    phigh = np.where(k == n, 1.0, phigh)
```
mdhaber, line 139: Could use `at.set` here with array API. Probably should test these explicitly.

Also need to make sure these and all other attributes are scalars for scalar input.

scipy/stats/_binomtest.py
```diff
@@ -115,48 +115,24 @@ def proportion_ci(self, confidence_level=0.95, method='exact'):
...
-    except RuntimeError:
-        raise RuntimeError('numerical solver failed to converge when '
-                           'computing the confidence limits') from None
-    except ValueError as exc:
-        raise ValueError('brentq raised a ValueError; report this to the '
-                         'SciPy developers') from exc
-    return p
-
-
 def _binom_exact_conf_int(k, n, confidence_level, alternative):
```
mdhaber, line 118: `confidence_level` could easily be an array, if desired, but I don't suppose anyone is asking for it.

scipy/stats/_binomtest.py
```diff
@@ -286,52 +254,61 @@ def binomtest(k, n, p=0.5, alternative='two-sided'):
...
-    if not (0 <= p <= 1):
-        raise ValueError(f"p ({p}) must be in range [0,1]")
+    k, n, p = np.broadcast_arrays(k, n, p)
+    k_valid = (k >= 0) & (k <= n) & (k == np.floor(k))
+    n_valid = (n >= 1) & (n == np.floor(n))
+    p_valid = (p >= 0) & (p <= 1)
+    valid = k_valid & n_valid & p_valid
+    k = np.where(valid, k, np.nan)
+    n = np.where(valid, n, np.nan)
+    p = np.where(valid, p, np.nan)
```
mdhaber, line 268: Unfortunately, `k` and `n` are public attributes of the return object. This makes it unclear whether it's OK to NaN these out when there is, say, a problem with `p`. 

To keep it as simple as possible, I NaN-out all of `k`, `n`, `p` wherever there is a problem with one of them because the NaNs propagate naturally through the calculations. We can do whatever we want; it just might require more effort.

mdhaber:
The lint failure is a `tach` complaint about using `optimize.elementwise.find_root` in stats. We're going to want other elementwise solve/minimization capabilities in functions like `expectile` and `boxcox_normmax`, so I think that's a fair dependency to add. 

*Update: pushed with `[lint only]` since all other tests were passing.*

tach.toml
```diff
@@ -164,6 +164,7 @@ depends_on = [
     "scipy.stats.distributions",
     "scipy.stats.mstats",
     "scipy.stats.qmc",
+    "scipy.optimize.elementwise",
```
lucascolley: _extreme nit_ in case you push another commit, I would keep the list alphabetical 😅

mdhaber:
I came back to this today, and I think it is pretty valuable. It can provide a 200x speedup vs looping over array arguments, and it prepares for array API translation, which could offer additional speedups with other backends. Turns out it's not necessarily slower than the original implementation, even for the 0d case - it depends on the input. The results for the statistic and p-value are identical, and the confidence interval limits are very close. Here's some code for generating test cases and comparing the times:

```python3
import numpy as np
from scipy import stats
rng = np.random.default_rng(324892349814829284)

def gen_test(size):
    p0 = rng.random(size)
    n = rng.integers(1, 1000, size=size)
    k = np.round(p0 * n).astype(int)
    p = rng.random(size)
    return k, n, p

args = gen_test(1000)
binomtest_vec = np.vectorize(stats.binomtest)
%timeit binomtest_vec(*args)  # main: 423 ms ± 1.88 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
%timeit stats.binomtest(*args)  # PR: 2.65 ms ± 11.7 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

For a quick and easy comparison of the values produced between the two branches, I find it convenient to print the arrays to the console:
```python3
@np.vectorize
def binomtest_vec(*args):
    res = stats.binomtest(*args)
    ci = res.proportion_ci()
    return res.statistic, res.pvalue, ci.low, ci.high

statistic, pvalue, low, high = binomtest_vec(*args)
print(f"statistic = np.asarray({statistic.tolist()})")
print(f"pvalue = np.asarray({pvalue.tolist()})")
print(f"low = np.asarray({low.tolist()})")
print(f"high = np.asarray({high.tolist()})")
```

Since it's been a few months, I'll do another self-review, and potentially propose on the forum that we call that sufficient for merge if nobody else is interested.

scipy/stats/_binomtest.py
```diff
@@ -175,28 +151,17 @@ def _binom_wilson_conf_int(k, n, confidence_level, alternative, correction):
...
-            dlo = (1 + z*sqrt(z**2 - 2 - 1/n + 4*p*(n*q + 1))) / denom
-            lo = center - dlo
-        if alternative == 'greater' or k == n:
-            hi = 1.0
-        else:
-            dhi = (1 + z*sqrt(z**2 + 2 - 1/n + 4*p*(n*q - 1))) / denom
-            hi = center + dhi
+        with np.errstate(divide='ignore', invalid='ignore'):
+            dlo = (1 + z*np.sqrt(z**2 - 2 - 1/n + 4*p*(n * q + 1)))/denom
+            dhi = (1 + z*np.sqrt(z**2 + 2 - 1/n + 4*p*(n * q - 1)))/denom
```
mdhaber:
```suggestion
            dlo = (1 + z*np.sqrt(z**2 - 2 - 1/n + 4*p*(n*q + 1)))/denom
            dhi = (1 + z*np.sqrt(z**2 + 2 - 1/n + 4*p*(n*q - 1)))/denom
```

scipy/stats/_binomtest.py
```diff
@@ -175,28 +151,17 @@ def _binom_wilson_conf_int(k, n, confidence_level, alternative, correction):
...
-        delta = z/denom * sqrt(4*n*p*q + z**2)
-        if alternative == 'less' or k == 0:
-            lo = 0.0
-        else:
-            lo = center - delta
-        if alternative == 'greater' or k == n:
-            hi = 1.0
-        else:
-            hi = center + delta
+        delta = z / denom * np.sqrt(4 * n * p * q + z ** 2)
```
mdhaber:
```suggestion
        delta = z / denom * np.sqrt(4*n*p*q + z**2)
```

scipy/stats/_binomtest.py
mdhaber: This is fine for now. If a user requests to NaN-out only the values that were invalid, I think that would be a compatible change.

scipy/stats/_binomtest.py
```diff
@@ -286,52 +254,61 @@ def binomtest(k, n, p=0.5, alternative='two-sided'):
...
             # we need to add a 1 to account for the 0 index.
             # For comparing this with old behavior, see
             # tst_binary_srch_for_binom_tst method in test_morestats.
             y = ix + 1
-            pval = binom.cdf(y-1, n, p) + binom.sf(k-1, n, p)
+            pval = B.cdf(y-1) + B.sf(k-1)
+            return pval
 
-        pval = min(1.0, pval)
+        pval = xpx.apply_where(k < p * n, (d, k, p, n), k_lt_pn,  k_gte_pn)
```
mdhaber:
```suggestion
        pval = xpx.apply_where(k < p*n, (d, k, p, n), k_lt_pn,  k_gte_pn)
```
Probably should check the  `k == p*n` is tested.

scipy/stats/_binomtest.py
```diff
@@ -286,52 +254,61 @@ def binomtest(k, n, p=0.5, alternative='two-sided'):
...
+        pval = np.minimum(1.0, pval)
 
+    statistic = np.where(valid, k/n, np.nan)
```
mdhaber, line 312:
```suggestion
    statistic = k / n
```
This should be fine because `k` and `n` are already `np.nan` where `~valid`.

scipy/stats/tests/test_morestats.py
```diff
@@ -1088,6 +1064,42 @@ def test_boost_overflow_raises(self):
...
+        ref_k, ref_n, ref_statistic, ref_pvalue, ci_low, ci_high = binomtest_1d(
+            k, n, p, alternative)
+        assert_allclose(res.k, ref_k)
+        assert_allclose(res.n, ref_n)
+        assert_allclose(res.statistic, ref_statistic)
+        assert_allclose(res.pvalue, ref_pvalue)
+        assert_allclose(ci.low, ci_low)
+        assert_allclose(ci.high, ci_high)
+
+
```
mdhaber, line 1254:
```suggestion
```

scipy/stats/_binomtest.py
```diff
@@ -115,48 +115,24 @@ def proportion_ci(self, confidence_level=0.95, method='exact'):
...
-        else:
-            plow = _findp(lambda p: binom.sf(k-1, n, p) - alpha)
-        phigh = 1.0
+    init = (0.0, 1.0)
+    args = (k, n)
+    alpha = ((1 - confidence_level) / 2 if alternative == 'two-sided'
+             else 1 - confidence_level)
+
+    plow = (np.zeros(k.shape, dtype=np.float64) if alternative == 'less' else
+            find_root(lambda p, k, n: binom.sf(k-1, n, p) - alpha, init, args=args).x)
```
mdhaber: maybe use `_SimpleBinom` here.

j-bowhay (approved):
I had looked over this previously but didn't press approve as I hadn't looked into the impact of this slow down for the 1D case. Since it sounds like this isn't an issue I'm happy to merge

mdhaber:
Great! OK, I think I'll make the changes I suggested above and re-run CI, then we should be good to go.

scipy/stats/_binomtest.py
mdhaber: Went ahead and used the private methods of `binom` directly. At this point, we shouldn't need any of the input validation/standardization done by `_SimpleBinomial`.

scipy/stats/_binomtest.py
mdhaber: >  Probably should test these explicitly.

`test_confidence_interval_exact_k0` and `test_confidence_interval_exact_k_is_n` seem to do just that.

> Also need to make sure these and all other attributes are scalars for scalar input.

Done.

scipy/stats/_binomtest.py
mdhaber: Indeed, `test_estimate_equals_hypothesized_prop`.

scipy/stats/_binomtest.py
mdhaber: `pval = np.where(valid, pval, np.nan)` is needed, so I'll keep this as-is for visual consistency and to make it more obvious what the result will be when parameters are invalid.

tach.toml
mdhaber: Actually it's already been added in the right place in `main`, so I can just remove this.

PR #23666, opened 2025-09-23
DOC/DEV: Document `scipy.signal`'s Array API caveats.

DietBru:
This is a contribution towards resolving issue #23539 by pointing to the code decorating `scipy.signal` calls. 

I am not sure if I caught all relevant information. The  `Caveats` section should be the explanation to point to when needing to skip unit tests due to CuPy or JAX issues.

What is not included are guidelines on how to consistently mark skipped unit tests, i.e., how to word the `reason` parameter to make it easier to unmark, once the backends caught up.

PR #24024, opened 2025-11-19
ENH/TST: signal: Add xp_capabilities for scipy.signal

steppi:

This PR adds `xp_capabillities` for public functions in `scipy.signal` and makes use of them in tests through `make_xp_test_case` and `make_xp_pytest_param` where needed. This follows up on #23765 and #23921 . #23765 expanded alternative backend test coverage in `signal` by isolating the use of the alternative backend only to the function being tested, removing skips that were included due to lack of support in functions not being tested by a particular test. #23921 updated filter functions to ensure consistent output dtypes across backends. The work done in those two PRs and the work done in this PR was originally all done on a single branch, but for ease of review I stacked the work into two PRs. The interactive rebasing and merge conflict resolution I needed to do to create this PR was nontrivial; I'm fairly confident everything was done correctly, but it may be worth keeping an eye out for quirks that may have popped. Everything worked on prior to the stacking of this work into three separate PRs, and the subsequent merge conflict resolution after the first two PRs were merged, has been squashed into the single commit "ENH/TST: Add xp_capabilities for scipy.signal and use in tests". Incidentally, I figured out how to squash commits through intermediate merges of main which was something that had vexed attempts to clean up history in some stats PRs before. The process involves creating temporary branches, and use of `git merge --squash`. 

Now for the meat of the PR itself. Except for the functions in `scipy.signal.windows`, `xp_capabilities` are applied in `scipy/signal/_support_alternative_backends.py`, in the loop where delegation/dispatch is set up. The default capabilities for such a function is that everything is supported, and one must add capabilities entries to a `dict` called `capabilities_overrides` to override the default. Perhaps it makes more sense to have the default capabilities be `np_only`, I'm not sure. There were a number of public functions which are included in the dispatch loop but have no alternative backend tests in the entire codebase. I've set things up so that these are all marked as `np_only`. It doesn't make sense to me to mark something as supported if it's not tested. It will be a natural follow-up to enable tests for these functions. There are quite a few, 43 of them. I identified these functions in a dumb way by grepping through the test suite. It seems like it may be a good idea to add a test for whether there exist functions and classes in the public API with `xp_capabilities` entries other than `np_only` but no tests using the `xp` fixture.

There are transfer guard annoyances with JAX, where tests on GPU can seem like they pass even if they didn't because there's no protection against transferring from device to cpu.  I started out by checking for this situation carefully, but ended up just following the heuristic that if the jit isn't supported, then it's probably not actually working on GPU. This reminds me that I think it's supposed to be up to me to help push on this transfer guard issue.

I've used the `extra_note` option for `xp_capabilities` that @ev-br added in #23746 to add notes about specific limitations for some backends. It's doubtful that I was exhaustive in listing the limitations, but I think I caught everything that had a corresponding `skip_xp_backends` in the test suite.

I worry that I've put an almost pathological amount of care into documenting the capabilities of the functions correctly in this PR. All tests pass for me locally, but it's possible that I might have still got something wrong. I don't think it's worth checking this over any more though. In my opinion it's more than good enough to ship.

#### Additional information
@DietBru. #23666 and it's follow-up got merged after I started working on this. I was very happily surprised when I first looked at the page for `signal` array api support in my browser after merging `main` recently. Thanks for working on that.


steppi:
The docs failure is puzzling

```
/home/circleci/repo/build-install/usr/lib/python3.11/site-packages/scipy/signal/_signaltools.py:docstring of scipy.signal._signaltools.detrend:48: WARNING: py:func reference target not found: numpy.polynomial.polynomial.Polynomial.fit [ref.func]
```

What's changed in the docstring is the insertion of the capabilities note. I'm not sure why that would cause this warning to suddenly appear though.

steppi:
It looks like https://github.com/scipy/scipy/blob/main/scipy/_lib/_docscrape.py, used by `xp_capabilities` to update the docs is causing cross references to get mangled. Inside of `xp_capabilities`, it changes

```
    See Also
    --------
    numpy.polynomial.polynomial.Polynomial.fit: Create least squares fit polynomial.
```

to

```
See Also
--------

:func:`numpy.polynomial.polynomial.Polynomial.fit`
    Create least squares fit polynomial.
```

which works fine for references within SciPy but not for intersphinx references.

scipy/signal/tests/test_bsplines.py
```diff
@@ -13,14 +14,16 @@
 skip_xp_backends = pytest.mark.skip_xp_backends
 xfail_xp_backends = pytest.mark.xfail_xp_backends
 
+lazy_xp_modules = [signal]
```
ev-br, line 17: Could you please explain what this is? I'm guessing it's something about testing jax/dask support; some guidance would be helpful going forward on when this needs to be added, what this does and what it interacts with.

scipy/signal/tests/test_bsplines.py
```diff
@@ -153,7 +158,8 @@ def test_qspline1d(self, xp):
             signal.qspline1d(xp.asarray([1., 2, 3, 4, 5], dtype=xp.float64)), q1d0
         )
 
-    @skip_xp_backends(cpu_only=True)
+    @xfail_xp_backends("cupy", reason="https://github.com/cupy/cupy/pull/9484")
```
ev-br, line 161: Could you comment on when a skip/xfails goes to xp_capabilities and when it's a decorator to the test. I'm guessing the latter is for when the there's something limiting in the test itself?

Note I'm not suggesting to change anything here, the implementatation is solid as is.

scipy/signal/tests/test_bsplines.py
```diff
@@ -153,7 +158,8 @@ def test_qspline1d(self, xp):
...
+    @make_xp_test_case(signal.cspline1d_eval)
```
ev-br, line 162: The test also uses  `cspline1d` --- does it mean that it should be `make_xp_test_case([signal.cspline1d_eval, signal.cspline1d])`? Not that it matters here specifically, the question is more for my understanding of the game.

scipy/signal/tests/test_cont2discrete.py
```diff
@@ -11,16 +12,19 @@
 # Author: Jeffrey Armstrong <jeff@approximatrix.com>
 # March 29, 2011
 
+skip_xp_backends = pytest.mark.skip_xp_backends
 
+
```
ev-br, line 18: does this test module need `lazy_xp_modules = [signal]`, and if it does not then what tells it from, say, `test_bsplines.py`

scipy/signal/tests/test_cont2discrete.py
```diff
@@ -11,16 +12,19 @@
...
 class TestC2D:
-    def test_zoh(self):
-        ac = np.eye(2, dtype=np.float64)
-        bc = np.full((2, 1), 0.5, dtype=np.float64)
-        cc = np.array([[0.75, 1.0], [1.0, 1.0], [1.0, 0.25]])
-        dc = np.array([[0.0], [0.0], [-0.33]])
-
-        ad_truth = 1.648721270700128 * np.eye(2)
-        bd_truth = np.full((2, 1), 0.324360635350064)
+    def test_zoh(self, xp):
```
ev-br, line 21: Ah, so you also converted tests modules / test for functions which were not converted previously. Wow.

scipy/signal/tests/test_cont2discrete.py
```diff
@@ -228,10 +232,11 @@ def test_zerospolesgain(self):
 
         assert_array_almost_equal(zeros_d, zeros)
         assert_array_almost_equal(polls_d, poles)
-        assert_almost_equal(k_d, k)
-        assert_almost_equal(dt_requested, dt)
+        np.testing.assert_almost_equal(k_d, k)
```
ev-br: Note in passing: why? `assert_array_almost_equal` just above comes from `_array_api`. Not that it matters much at this stage --- the function this tests has not been  not converted IIUC, and tests will likely need one more pass anyway.
Fine to keep as is, just wondering if there's a deep reason for this change.

scipy/signal/tests/test_filter_design.py
```diff
@@ -240,14 +237,14 @@ def test_identity(self, xp):
             assert isinstance(a, np.ndarray)
 
     @skip_xp_backends("dask.array", reason="https://github.com/dask/dask/issues/11883")
-    @xfail_xp_backends("cupy", reason="inaccurate")
```
ev-br, line 243: So to check my understanding: the CuPY xfail got moved to `_support_alternative_backends` centrally because it's a limitation of the function itself, but the dask skip is sensitive to the specific testing scenario, hence it stays here with the test.

scipy/signal/tests/test_filter_design.py
```diff
@@ -1643,6 +1645,9 @@ def test_basic(self, xp):
         xp_assert_close_nulp(b_z, xp.asarray(b_zref))
         xp_assert_close_nulp(a_z, xp.asarray(a_zref))
 
+        xp_assert_close_nulp(b_z, xp.asarray(b_zref))
+        xp_assert_close_nulp(a_z, xp.asarray(a_zref))
```
ev-br: Duplicate with two lines just above.

scipy/signal/tests/test_filter_design.py
```diff
@@ -1876,14 +1887,16 @@ def test_output_type(self):
                 assert isinstance(p, np.ndarray)
 
     @pytest.mark.parametrize(
-        'func', [buttap, besselap,
-                 lambda N, xp: cheb1ap(N, 1, xp=xp),
-                 lambda N, xp: cheb2ap(N, 20, xp=xp),
-                 lambda N, xp: ellipap(N, 1, 20, xp=xp)
-                ],
+        'base_func,func', [
+            make_xp_pytest_param(buttap, buttap),
```
ev-br, line 1888: I'm sure there's a deep reason for `make_xp_test_param` requiring two arguments. Not sure what it is though.

scipy/signal/tests/test_filter_design.py
```diff
@@ -1985,17 +2002,18 @@ def test_analog(self, xp):
...
-        n, Wn = buttord(1, xp.asarray(550/450), 1, 26, analog=True)
+        n, Wn = buttord(1., xp.asarray(550/450), 1, 26, analog=True)
         assert n == 19
         xp_assert_close(
-            Wn, xp.asarray(1.0361980524629517), rtol=1e-15, check_0d=False
+            Wn, xp.asarray(1.0361980524629517), rtol=4*eps, check_0d=False
         )
 
-        assert buttord(1, xp.asarray(1.2), 1, 80, analog=True)[0] == 55
+        assert buttord(1., xp.asarray(1.2), 1, 80, analog=True)[0] == 55
```
ev-br: For my education: is this int->float change required, or is it from some other refactoring? 
No need to change it either way.

scipy/signal/tests/test_filter_design.py
```diff
@@ -2129,9 +2152,9 @@ def test_analog(self, xp):
         assert np.all(dB(h[w <= ws]) < -rs + 0.1)
 
         assert N == 4
-        assert math.isclose(Wn, 700.0, rel_tol=1e-15)
-
-        assert cheb1ord(1, 1.2, 1, 80, analog=True)[0] == 17
+        math.isclose(Wn, 700.0, rel_tol=1e-15)
```
ev-br:
```suggestion
        assert math.isclose(Wn, 700.0, rel_tol=1e-15)
```

scipy/signal/tests/test_filter_design.py
```diff
@@ -2147,7 +2170,8 @@ def test_fs_param(self, xp):
         assert np.all(dB(h[ws <= w]) < -rs + 0.1)
 
         assert N == 8
-        assert math.isclose(Wn, 4800.0, rel_tol=1e-15)
+        math.isclose(Wn, 4800.0, rel_tol=1e-15)
```
ev-br:
```suggestion
        assert math.isclose(Wn, 4800.0, rel_tol=1e-15)
```

scipy/signal/tests/test_filter_design.py
```diff
@@ -2279,7 +2305,8 @@ def test_fs_param(self, xp):
         assert np.all(dB(h[w <= ws]) < -rs + 0.1)
 
         assert N == 9
-        assert math.isclose(Wn, 103.4874609145164, rel_tol=1e-15)
+        math.isclose(Wn, 103.4874609145164, rel_tol=1e-15)
```
ev-br:
```suggestion
        assert math.isclose(Wn, 103.4874609145164, rel_tol=1e-15)
```

scipy/signal/tests/test_filter_design.py
```diff
@@ -4739,32 +4788,42 @@ def test_fs_validation(self):
...
+            xp_assert_close(_sort_cmplx(p, xp=xp), _sort_cmplx(xp.conj(p), xp=xp))
+            assert complex(k).imag == 0
+
+            b, a = iirfilter(N, xp.asarray(1.1), 1, 20, 'low', analog=True,
+                             ftype=ftype, output='ba')
+            if is_numpy(xp):
+                assert issubclass(b.dtype.type, np.floating)
+                assert issubclass(a.dtype.type, np.floating)
+            else:
+                assert xp.isdtype(b.dtype, ('real floating', 'complex floating'))
```
ev-br: Is `is_numpy(xp)` needed because `xp` is a "bare" numpy, so can be numpy 1.26 which does not have `isdtype`? Could either use a comment to this effect, or `isdtype = array_namespace(b).isdtype` or something like this.

scipy/signal/tests/test_fir_filter_design.py
```diff
@@ -557,7 +570,7 @@ def test_gh_23266(self, xp):
         remez(21, bands, desired, weight=weight)
 
 
-@skip_xp_backends(cpu_only=True, reason="lstsq", exceptions=["cupy"])
```
ev-br, line 560: Note to self: can reuse the pattern from `savgol_filter` for `lstsq` in `firls`.

scipy/signal/tests/test_signaltools.py
```diff
@@ -3190,24 +3194,27 @@ def test_auto_n(self, xp):
...
         sfreq = 100.
         n = 1000
-        t = np.arange(n) / sfreq
+        t = xp.arange(n) / sfreq
         # will alias for decimations (>= 15)
-        x = np.sqrt(2. / n) * np.sin(2 * np.pi * (sfreq / 30.) * t)
-        xp_assert_close(np.linalg.norm(x), 1., rtol=1e-3)
+        x = xp.asarray(xp.sqrt(2. / n) * xp.sin(2 * xp.pi * (sfreq / 30.) * t))
+        # Use xp.sqrt(x.dot(x)) instead of xp.linalg.vector_norm(x) because
+        # linear algebra extension is not universally available.
```
ev-br, line 3201: Uhm, this is because of `dask`, right? I'm pretty sure having a sensibly complete linalg extension is a prerequisite for support in scipy. Going forward I'd just skip whatever tests fail with dask. 
No need to change anything though.

scipy/signal/tests/test_signaltools.py
```diff
@@ -3190,24 +3194,27 @@ def test_auto_n(self, xp):
...
-        x = signal.decimate(np.ones(10_000, dtype=np.float32), 10)
-        assert not any(np.isnan(x))
+        x = signal.decimate(xp.ones(10_000, dtype=xp.float32), 10)
+        assert not any(xp.isnan(x))
 
     def test_float16_upcast(self, xp):
         # float16 must be upcast to float64
-        x = signal.decimate(np.ones(100, dtype=np.float16), 10)
-        assert x.dtype.type == np.float64
+        x = signal.decimate(xp.ones(100, dtype=xp.float16), 10)
```
ev-br: Nit: `xp.float16` is not a thing. 
So this only works because of `np_only=True` _somewhere_. If you're updating the PR for anything else, I think it's more self-descriptive to keep explicit `np.float16` and decorate the test itself with `skip_xp_backends(np_only)`.

scipy/signal/tests/test_signaltools.py
```diff
@@ -3217,16 +3224,17 @@ def test_complex_iir_dlti(self, xp):
         # sample rate [Hz]
         fs = 1e3
 
-        z, p, k = signal.butter(2, 2*np.pi*fwidth/2, output='zpk', fs=fs)
-        z = z.astype(complex) * np.exp(2j * np.pi * fcentre/fs)
-        p = p.astype(complex) * np.exp(2j * np.pi * fcentre/fs)
+        z, p, k = signal.butter(2, xp.asarray(2*xp.pi*fwidth/2),
+                                output='zpk', fs=fs)
+        z = z.astype(complex) * xp.exp(xp.asarray(2j * xp.pi * fcentre/fs))
```
ev-br, line 3229: `z.astype` only works because of `np_only=True`. OK to keep as is for now, this will need an update anyway, when converting `dlti`.

scipy/signal/tests/test_signaltools.py
```diff
@@ -3746,20 +3753,22 @@ def test_nyquist(self):
         xp_assert_close(z6, np.zeros(6, dtype=z6.dtype), atol=1e-12)
         xp_assert_close(z6_res, x6.astype(z6.dtype), atol=1e-12)
 
-@skip_xp_backends(np_only=True)
+
 class TestPartialFractionExpansion:
```
ev-br, line 3757: This I believe is safe to declare out of scope for Array API. Can go either way, so fine to keep designated as _numpy_only_, too.

Converting this test (and everything around residue/residuez) was annoying, I'm sure; since you already did the work, it's better as is.

scipy/signal/_signaltools.py
```diff
@@ -2692,7 +2692,7 @@ def hilbert2(x, N=None, axes=(-2, -1)):
         if N <= 0:
             raise ValueError("N must be positive.")
         N = (N, N)
-    elif len(N) != 2 or xp.any(xp.asarray(N) <= 0):
+    elif len(N) != 2 or np.any(np.asarray(N) <= 0):
```
steppi, line 2695: Changing `xp` to `np` here avoids an unnecessary dask compute.

scipy/signal/tests/test_filter_design.py
```diff
@@ -1643,6 +1645,9 @@ def test_basic(self, xp):
         xp_assert_close_nulp(b_z, xp.asarray(b_zref))
         xp_assert_close_nulp(a_z, xp.asarray(a_zref))
 
+        xp_assert_close_nulp(b_z, xp.asarray(b_zref))
+        xp_assert_close_nulp(a_z, xp.asarray(a_zref))
+
```
steppi: Oops. This looks like something that went wrong when I was resolving merge conflicts.

scipy/signal/tests/test_filter_design.py
```diff
@@ -1912,8 +1926,9 @@ def test_lowpass(self, xp):
         assert np.all(dB(h[ws <= w]) < -rs)
 
         assert N == 16
+        eps = xp.finfo(Wn).eps
         xp_assert_close(Wn,
-                        xp.asarray(2.0002776782743284e-01), rtol=1e-15, check_0d=False)
+                        xp.asarray(2.0002776782743284e-01), rtol=4*eps, check_0d=False)
```
steppi: Oh no. There should be no use of `eps` for `rtol` any longer. Anywhere this appears is due to a mistake in merge conflict resolution.

steppi (commented):
I did a quick self-review. It looks like I did in fact make mistakes resolving the merge conflicts, but nothing too serious.

scipy/signal/_support_alternative_backends.py
```diff
@@ -73,5 +372,11 @@ def wrapper(*args, **kwds):
     else:
         f = bare_obj
 
+    if callable(f) and hasattr(f, "__name__"):
```
ev-br: Could you double-check that this works correctly with classes? There are several in signal, `ShortTimeFFT`, `lti`, `dlti`. 
In between classes becoming functions under dask (cannot find the issue right now), and https://github.com/scipy/scipy/pull/23960 (which we had to revert), I would really like to make sure we don't keep paper over classes.

Speaking of which, should `lti`, `dlti` et al get decorated by `xp_capabilities(np_only)`, too?

ev-br (commented):
This is great @steppi!

I've several small comments below, mostly usage questions, trying to tap into your having figured out this woefully underdocumented machinery.
Would be great to actually document it, but it's certainly for a follow-up if you're planning to have one.
None of my comments below are blocking.

scipy/signal/tests/test_bsplines.py
steppi: This seems to be required for tests involving functions wrapped with `lazy_xp_function` to work properly when  functions are referred to by names qualified by a module, e.g. `signal.spline_filter`. I'd never looked into the details before, it was just an incantation I'd seen others use, but it is used in `patch_xp_modules` from `array_api_extra.testing`, see:

https://github.com/data-apis/array-api-extra/blob/20593693e22bcfc0fddee27fcacc91abc8c8dbf7/src/array_api_extra/testing.py#L273-L285

It's not required when you import functions directly from the module, so wouldn't be needed if we'd done `from scipy.signal import spline_filter`.

scipy/signal/tests/test_bsplines.py
steppi: My understanding is something goes into `xp_capabilities` when it's a limitation of the function itself, and the `skip` or `xfail` goes on the test when the problem either relates to the test, or is due to a specific case which is unsupported or doesn't work for a function which is generally supported. For example, the bug above only happens when you hit an edge case in CuPy, but the function is otherwise supported.

scipy/signal/tests/test_bsplines.py
steppi: Good catch. It either needs to be `make_xp_test_case([signal.cspline1d_eval, signal.cspline1d])`, or `signal.cspline1d` should get the `_xp_copy_to_numpy` treatment. Here I think it's the latter, since this is meant as a consistency check for `cspline1d_eval`.

scipy/signal/tests/test_cont2discrete.py
steppi: The only reason it doesn't need `lazy_xp_modules` is because all functions used in this test are imported from `signal`. We would need `lazy_xp_modules` if there was an import `import scipy.signal as signal` and then we used `signal.c2d` somewhere, or something like that.

scipy/signal/tests/test_cont2discrete.py
steppi: Yeah, if it seemed easy enough to convert, I just did it, but there's a whole list of over 40 functions with tests that still need to be converted, but which all probably work with CuPy.

scipy/signal/tests/test_cont2discrete.py
steppi: I'll check. I think `assert_array_almost_equal` failed here on some backends because `k_d` and `k` aren't `xp` arrays, they are just floats, The pattern I saw elsewhere that other people used was to use `math.isclose` in this case, so for consistency I should probably change to that. Just something I missed due to the scale of the changes I think.

scipy/signal/tests/test_filter_design.py
steppi: Yes, I think that's right.

scipy/signal/tests/test_filter_design.py
steppi: Yeah, I noticed that too. It was a mistake I made when resolving the merge conflicts, which got kind of hairy.

scipy/signal/tests/test_filter_design.py
steppi: It's more of a lazy reason, where I just wanted to make the minimal changes needed to get this working. The first argument of `make_xp_pytest_param` is always the function whose `xp_capabilities` will be used to make the pytest params, but in this test, some of the functions are tested through specializations defined with `lambda`. This could have been refactored to avoid this by using a pattern like

```python
 func,args, [
            make_xp_pytest_param(buttap, []),
            make_xp_pytest_param(besselap, []),
            make_xp_pytest_param(cheb1ap, [1]),
            make_xp_pytest_param(cheb2ap, [20]),
            make_xp_pytest_param(ellipap, [1, 20]),
        ],
        ids=['butter', 'bessel', 'cheb1', 'cheb2', 'ellip']
    )
```

and then calling `func(7, *args, xp=xp)` inside of the test body. Your call on what you think is better.

scipy/signal/tests/test_filter_design.py
steppi: I'll check. It may have been an accidental change.

scipy/signal/tests/test_filter_design.py
steppi: Good catch.

scipy/signal/tests/test_filter_design.py
steppi: Yes, I think that's exactly right. I like the `isdtype =` suggestion.

scipy/signal/tests/test_signaltools.py
steppi: Good point. This probably wasn't worth spending time.

scipy/signal/tests/test_signaltools.py
steppi: Good point. This was a result of mindless conversion. I'll fix.

scipy/signal/_support_alternative_backends.py
steppi: Good catch. When I started I was still ignoring classes. I'll fix this.

steppi:
Thanks for the careful review @ev-br! Your suggestions are very helpful.

scipy/signal/tests/test_signaltools.py
ev-br: Great that it works now anyhow.

scipy/signal/tests/test_bsplines.py
ev-br: Wait, what. So `import foo; foo.bar(x)` is no longer the same as `from foo import bar; bar(x)`?

scipy/signal/tests/test_filter_design.py
ev-br: Thanks, that's clever!
Now that I stared long enough at the docstring example of `make_xp_pytest_param`, it makes sense. 

Some more centralized documentation for various moving parts would be nice, but there's no need to mix this into this PR.

scipy/signal/tests/test_bsplines.py
steppi: Yeah, kind of, but only with respect to the behavior of `patch_lazy_xp_functions`. This is what the docstring of `patch_lazy_xp_functions` has to say about it

```
    Test lazy execution of functions tagged with :func:`lazy_xp_function`.

    If ``xp==jax.numpy``, search for all functions which have been tagged with
    :func:`lazy_xp_function` in the globals of the module that defines the current test,
    as well as in the ``lazy_xp_modules`` list in the globals of the same module,
    and wrap them with :func:`jax.jit`. Unwrap them at the end of the test.
```

I assume it's implemented this way to make it easier to avoid applying `jax.jit` or whatever to functions from separate modules that aren't being tested by the current test. In principle, it should be possible to search for all functions that have been tagged with `lazy_xp_function` in the globals of the current module, and globals of public submodules or something, and so on recursively.

scipy/signal/tests/test_filter_design.py
steppi: Agreed about documentation. The next PR I plan to submit is for exactly that.

scipy/signal/tests/test_cont2discrete.py
steppi: OK, so this does actually work with CuPy, and I changed it because `assert_array_almost_equal` was failing because the two values being compared are Python floats and `assert_array_almost_equal` expects them both to be `xp` arrays. I'm going to change to using `assert math.isclose` to compare scalars, following what I saw being done for tests of `k` for functions returning a `z`, `p`, `k` representation.

scipy/signal/tests/test_signaltools.py
steppi: OK, so the reality was that this test does pass with CuPy, the only supported alternative backend for `signal.decimate`. CuPy, PyTorch, and JAX all support `float16`, half precision is just not required by the standard. I'm leaning towards thinking that it's fine to test things that go beyond the standard if they are supported and were deemed useful enough to make it into the existing test suite. I suspect such cases would be rare enough that it wouldn't be burdensome to add skips where necessary.

scipy/signal/tests/test_filter_design.py
steppi: This seems to have been an accidental change, the tests still pass when I undo it.

scipy/signal/_support_alternative_backends.py
steppi: I changed it to 

```python
    if not isinstance(f, types.ModuleType):
        capabilities = capabilities_overrides.get(
            obj_name, get_default_capabilities(obj_name, delegator)
        )
        f = capabilities(f)
```

which seems to be the minimal restriction needed. I added entries to `capabilities_overrides` for `dlti` and `lti`, but they aren't actually needed because everything in the public API not appearing in `_delegators.py` by default is decorated with `np_only=True`.

steppi:
I think that should cover all of the review comments.

scipy/signal/tests/test_bsplines.py
lucascolley: yes, see the notes section of https://data-apis.org/array-api-extra/generated/array_api_extra.testing.lazy_xp_function.html

ev-br:
I think it's about time we come up with some pytest markings for  _slow under certain backends_. 
Locally, the test run across backends takes a whopping 10 minutes, as compared to under 20 sec for a "regular" test run. Memory requirements are also spiking---not anywhere near swapping but nonetheless. 

That's not new in this PR though, it's the same on main.

```
$ spin test -s signal -b all -v -- --durations 20

... snip... 

================================= slowest 20 durations =================================
16.18s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_fir_filter_design.py::TestMinimumPhase::test_hilbert[array_api_strict]
14.82s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_filter_design.py::TestButter::test_basic[jax.numpy]
13.65s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_splines.py::TestSymIIR::test_symiir2[dask.array--1.0-float32]
13.58s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_splines.py::TestSymIIR::test_symiir2[dask.array--1.0-float64]
13.09s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_splines.py::TestSymIIR::test_symiir2[dask.array-0.7-float32]
12.42s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_splines.py::TestSymIIR::test_symiir2[dask.array-0.7-float64]
12.34s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_splines.py::TestSymIIR::test_symiir2[dask.array-0.25-float64]
11.91s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_splines.py::TestSymIIR::test_symiir2[dask.array-0.0075-float64]
11.72s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_splines.py::TestSymIIR::test_symiir2[dask.array-0.5-float64]
11.65s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_splines.py::TestSymIIR::test_symiir2[dask.array-0.0075-float32]
11.37s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_splines.py::TestSymIIR::test_symiir2[dask.array-0.25-float32]
11.30s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_splines.py::TestSymIIR::test_symiir2[dask.array-0.5-float32]
10.81s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_splines.py::TestSymIIR::test_symiir1[dask.array--1.0-float64]
10.71s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_splines.py::TestSymIIR::test_symiir1[dask.array-0.0075-complex64]
10.64s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_fir_filter_design.py::TestMinimumPhase::test_hilbert[torch]
10.58s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_splines.py::TestSymIIR::test_symiir1[dask.array--1.0-complex64]
10.14s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_splines.py::TestSymIIR::test_symiir1[dask.array--1.0-float32]
9.97s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_splines.py::TestSymIIR::test_symiir1[dask.array-0.7-float32]
9.97s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_splines.py::TestSymIIR::test_symiir1[dask.array-0.0075-float32]
9.95s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_savitzky_golay.py::test_sg_coeffs_compare[jax.numpy]
= 10255 passed, 3219 skipped, 51736 deselected, 1199 xfailed, 10 xpassed in 717.70s (0:11:57) =
```

A "regular", numpy-only test run:

```
$ spin test -s signal  -- --durations 20

... snip ... 

1.99s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_signaltools.py::test_filtfilt_gust[numpy]
1.50s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_filter_design.py::TestBessel::test_fs_param
0.45s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_signaltools.py::test_choose_conv_method[numpy]
0.25s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_filter_design.py::TestEllip::test_fs_param
0.22s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_filter_design.py::TestCheby2::test_fs_param
0.21s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_filter_design.py::TestCheby1::test_fs_param
0.21s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_filter_design.py::TestButter::test_fs_param
0.20s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_fir_filter_design.py::Testfirwin_2d::test_circular_symmetry
0.16s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_signaltools.py::TestConvolve::test_convolve_method[numpy]
0.16s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_fir_filter_design.py::TestMinimumPhase::test_hilbert[numpy]
0.13s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_spectral.py::TestSTFT::test_roundtrip_complex[spectrum]
0.13s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_ltisys.py::TestPlacePoles::test_complex
0.13s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_filter_design.py::TestFreqz_sos::test_freqz_sos_against_mp[numpy]
0.13s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_spectral.py::TestSTFT::test_roundtrip_complex[psd]
0.12s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_czt.py::test_1D
0.12s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_short_time_fft.py::test_roundtrip_windows[hann-1024-256-255]
0.10s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_filter_design.py::TestIIRFilter::test_symmetry[numpy-bessel]
0.10s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_short_time_fft.py::test_energy_conservation[7.0-4-256-1337]
0.10s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_short_time_fft.py::test_energy_conservation[23.0-4-256-1337]
0.10s call     build-install/usr/lib/python3.12/site-packages/scipy/signal/tests/test_signaltools.py::TestFFTConvolve::test_valid_mode_complex_axes[numpy-axes3]
============================================= 3607 passed, 3 skipped, 10348 deselected, 5 xfailed, 4 xpassed in 18.82s =============================================
```

scipy/signal/tests/test_cont2discrete.py
```diff
@@ -30,21 +35,21 @@ def test_zoh(self):
         assert_array_almost_equal(bd_truth, bd)
         assert_array_almost_equal(cc, cd)
         assert_array_almost_equal(dc, dd)
-        assert_almost_equal(dt_requested, dt)
+        np.testing.assert_almost_equal(dt_requested, dt)
```
ev-br, line 38: A few more `np.testing` leftovers slipped past---not surprising at all given the size of the patch. I'll send a follow-up PR.

ev-br (approved):
LGTM now. It's been a herculean effort @steppi, much appreciated!

Two large items---the test times and documenting the `lazy_xp_function` esoteria are best addressed in a follow-up; if we find small glitches, we'll address them in follow-ups, too, given that it's already a 1 KLOC patch.

steppi:
Thanks @ev-br! I'll work on the documentation next. I've been thinking about adding something like `slow_xp_backends` and `xslow_xp_backends` too; I agree we should get to that fairly soon.

PR #13288, opened 2020-12-23
DOC: stats: discrete distribution shape parameter restrictions 

mdhaber:
#### Reference issue
gh-13283

#### What does this implement/fix?
While reviewing gh-13283, I noticed that some discrete distribution parameter restrictions were not documented. This corrects the ones I noticed, and also tries to make the stylization (e.g. math vs teletype) a bit more consistent. (I can't claim to have fixed all issues, as what should be done is sometimes ambiguous unless we spell out the relationship between the math symbols and code parameters.)

PR #24025, opened 2025-11-19
DOC: stats: add equations for Cramer von Mises functions

scottshambaugh:
[docs only]

Adds equations for the one-sample and two-sample Cramer-von Mises statistics. Previously it was a little ambiguous whether the sample-size-weighted or unweighted values were being calculated.

There are a lot of functions in stats that could use similar clarification, but this is the one I was working with today. :)

#### Additional information
Here's code showing the equivalence to a naive implementation that matches the equations:

```python
import numpy as np
from scipy.stats import cramervonmises, cramervonmises_2samp, rankdata, norm

def cvm_1sample_naive(x, cdf):
    x = np.sort(np.asarray(x))
    n = len(x)
    F = cdf(x)
    i = np.arange(1, n+1)
    return 1/(12*n) + np.sum(((2*i - 1)/(2*n) - F)**2)

def cvm_2sample_naive(x, y):
    n, m = len(x), len(y)

    z = np.concatenate([x, y])
    r = rankdata(z, method="average")
    rx = r[:n]
    ry = r[n:]

    rx_sorted = rx[np.argsort(x)]
    ry_sorted = ry[np.argsort(y)]
    i = np.arange(1, n+1)
    j = np.arange(1, m+1)

    U = n * np.sum((rx_sorted - i)**2) + m * np.sum((ry_sorted - j)**2)
    return U / (n*m*(n+m)) - (4*n*m - 1)/(6*(n+m))

rng = np.random.default_rng(0)

x = rng.normal(size=200)
s1 = cramervonmises(x, norm.cdf).statistic
n1 = cvm_1sample_naive(x, norm.cdf)

x = rng.normal(size=50)
y = rng.normal(size=40)
s2 = cramervonmises_2samp(x, y).statistic
n2 = cvm_2sample_naive(x, y)

errors = (s1 - n1, s2 - n2)
print((s1, n1, errors[0]))
print((s2, n2, errors[1]))
```

```
(0.04009744261640295, 0.04009744261640295, 0.0)
(0.05531481481481393, 0.05531481481481393, 0.0)
```

### DIFF
--- a/scipy/stats/_hypotests.py
+++ b/scipy/stats/_hypotests.py
@@ -534,7 +534,7 @@ def _cvm_result_to_tuple(res, _):
 @_axis_nan_policy_factory(CramerVonMisesResult, n_samples=1, too_small=1,
                           result_to_tuple=_cvm_result_to_tuple)
 def cramervonmises(rvs, cdf, args=(), *, axis=0):
-    """Perform the one-sample Cramér-von Mises test for goodness of fit.
+    r"""Perform the one-sample Cramér-von Mises test for goodness of fit.
 
     This performs a test of the goodness of fit of a cumulative distribution
     function (cdf) :math:`F` compared to the empirical distribution function
@@ -542,6 +542,13 @@ def cramervonmises(rvs, cdf, args=(), *, axis=0):
     assumed to be independent and identically distributed ([1]_).
     The null hypothesis is that the :math:`X_i` have cumulative distribution
     :math:`F`.
+
+    The test statistic :math:`T` is defined as in [1]_, where :math:`\omega^2`
+    is the Cramér-von Mises criterion and :math:`x_i` are the observed values.
+
+    .. math::
+        T = n\omega^2 =
+        \frac{1}{12n} + \sum_{i=1}^n \left[ \frac{2i-1}{2n} - F(x_i) \right]^2
 
     Parameters
     ----------
@@ -565,7 +572,7 @@ def cramervonmises(rvs, cdf, args=(), *, axis=0):
     -------
     res : object with attributes
         statistic : float
-            Cramér-von Mises statistic.
+            Cramér-von Mises statistic :math:`T`.
         pvalue : float
             The p-value.
 
@@ -1629,12 +1636,26 @@ def _pval_cvm_2samp_asymptotic(t, N, nx, ny, k, *, xp):
 @_axis_nan_policy_factory(CramerVonMisesResult, n_samples=2, too_small=1,
                           result_to_tuple=_cvm_result_to_tuple)
 def cramervonmises_2samp(x, y, method='auto', *, axis=0):
-    """Perform the two-sample Cramér-von Mises test for goodness of fit.
+    r"""Perform the two-sample Cramér-von Mises test for goodness of fit.
 
     This is the two-sample version of the Cramér-von Mises test ([1]_):
     for two independent samples :math:`X_1, ..., X_n` and
     :math:`Y_1, ..., Y_m`, the null hypothesis is that the samples
     come from the same (unspecified) continuous distribution.
+
+    The test statistic :math:`T` is defined as in [1]_:
+
+    .. math::
+        T = \frac{nm}{n+m}\omega^2 =
+        \frac{U}{n m (n+m)} - \frac{4 m n - 1}{6(m+n)}
+
+    where :math:`U` is defined as below, and :math:`\omega^2` is the Cramér-von
+    Mises criterion. The function :math:`r(\cdot)` here denotes the rank of the
+    observed values :math:`x_i` and :math:`y_j` within the pooled sample of size
+    :math:`n + m`, with ties assigned mid-rank values:
+
+    .. math::
+        U = n \sum_{i=1}^n (r(x_i)-i)^2 + m \sum_{j=1}^m (r(y_j)-j)^2
 
     Parameters
     ----------
@@ -1657,7 +1678,7 @@ def cramervonmises_2samp(x, y, method='auto', *, axis=0):
     -------
     res : object with attributes
         statistic : float
-            Cramér-von Mises statistic.
+            Cramér-von Mises statistic :math:`T`.
         pvalue : float
             The p-value.
 


scipy/stats/_hypotests.py
```diff
@@ -531,6 +531,12 @@ def cramervonmises(rvs, cdf, args=()):
     The null hypothesis is that the :math:`X_i` have cumulative distribution
     :math:`F`.
 
+    The statistic `T` is weighted by the sample size `n`, and is defined as:
```
mdhaber: Single backticks (without explicit [role](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#roles)) can be used for parameters, functions, and classes; double backticks are for (other, general) code. I think this is supposed to be math, so you need to specify the `:math:` role:
```suggestion
    The statistic :math:`T` is weighted by the sample size :math:`n`, and is defined as:
```

(Here and below)

mdhaber: Could you point to the part of the references where a distinction is made between versions of the statistic that is/is not weighted by the sample size?

scottshambaugh: Thanks, math tags added.

I took out the bit about being weighted by sample size. The factor of `n` is the difference between the criterion and the test statistic, but it looks like the Cramer von Mises criterion is not commonly used.

scottshambaugh:
Rendered equations look good:

image

image

scipy/stats/_hypotests.py
```diff
@@ -548,7 +554,7 @@ def cramervonmises(rvs, cdf, args=()):
     -------
     res : object with attributes
         statistic : float
-            Cramér-von Mises statistic.
+            Cramér-von Mises statistic `T`.
```
mdhaber:
```suggestion
            Cramér-von Mises statistic :math:`T`.
```

scipy/stats/_hypotests.py
```diff
@@ -1634,7 +1653,7 @@ def cramervonmises_2samp(x, y, method='auto', *, axis=0):
...
```
mdhaber:
```suggestion
            Cramér-von Mises statistic :math:`T`.
```

scipy/stats/_hypotests.py
```diff
@@ -1606,13 +1612,26 @@ def _pval_cvm_2samp_asymptotic(t, N, nx, ny, k, *, xp):
...
 def cramervonmises_2samp(x, y, method='auto', *, axis=0):
-    """Perform the two-sample Cramér-von Mises test for goodness of fit.
+    r"""Perform the two-sample Cramér-von Mises test for goodness of fit.
 
     This is the two-sample version of the Cramér-von Mises test ([1]_):
     for two independent samples :math:`X_1, ..., X_n` and
     :math:`Y_1, ..., Y_m`, the null hypothesis is that the samples
     come from the same (unspecified) continuous distribution.
 
+    The statistic :math:`T` is defined as in [2]_:
```
mdhaber: I don't see the statistic labeled as $T$ in [2]. [2] does define a $U$, but it is quite different from what is here, which would be confusing. Since this is actually what is on Wikipedia, which is more accessible than [2], how about:
```suggestion
    The statistic :math:`T` is defined as in [1]_:
```

mdhaber (commented):
> There are a lot of functions in stats that could use similar clarification,

The first reference for both functions is the Wikipedia page, and in the end, all we're doing here is repeating the formulas that are on Wikipedia. That might be OK, but I think it would have been sufficient to clarify "Cramér-von Mises statistic $T$." to avoid confusion with the "criterion" $\omega^2$. An alternative would be to spend the time improving choices of references and, where applicable, the referenced Wikipedia pages. Note that there has been a lot of resistance to inclusion of LaTeX in documentation in the past, and that is probably why you don't find formulas like this already.

scottshambaugh:
> Note that there has been a lot of resistance to inclusion of LaTeX in documentation in the past

Interesting, I wasn't aware of the history there. My personal take is that wikipedia is a mutable reference so relying on variable names to line up isn't reliable, but I'm happy to align with the project conventions.

mdhaber:
I'm happy to include it here. That's just to explain why it's not all over stats. But for instance, you can see in the numpydoc style guide that the [Notes](https://numpydoc.readthedocs.io/en/latest/format.html#notes) section "may include mathematical equations", suggesting that the rest can't. There are also a lot of comments like https://github.com/scipy/scipy/pull/13288#issuecomment-751297769 to be found in the repo.
Yeah, Wikipedia is not the ideal reference; it's better when there is an accessible/free, reliable alternative.

scipy/stats/_hypotests.py
```diff
@@ -1630,13 +1636,26 @@ def _pval_cvm_2samp_asymptotic(t, N, nx, ny, k, *, xp):
...
     :math:`Y_1, ..., Y_m`, the null hypothesis is that the samples
     come from the same (unspecified) continuous distribution.
 
+    The statistic :math:`T` is defined as in [1]_:
+
+    .. math::
+        T = \frac{nm}{n+m}\omega^2 =
+        \frac{U}{n m (n+m)} - \frac{4 m n - 1}{6(m+n)}
+
+    where :math:`U` is defined as below. The function :math:`r()` here denotes
```
mdhaber: I think this is a bit more standard, but feel free to ignore.
```suggestion
    where :math:`U` is defined as below. The function :math:`r(\cdot)` here denotes
```

scipy/stats/_hypotests.py
```diff
@@ -544,6 +544,12 @@ def cramervonmises(rvs, cdf, args=(), *, axis=0):
     The null hypothesis is that the :math:`X_i` have cumulative distribution
     :math:`F`.
 
+    The test statistic :math:`T` is defined as:
+
+    .. math::
+        T = n\omega^2 =
+        \frac{1}{12n} + \sum_{i=1}^n \left[ \frac{2i-1}{2n} - F(x_i) \right]^2
```
mdhaber, line 551: One could say that $\omega$ and $x_i$ are undefined.

scipy/stats/_hypotests.py
```diff
@@ -1630,13 +1636,26 @@ def _pval_cvm_2samp_asymptotic(t, N, nx, ny, k, *, xp):
...
+    .. math::
+        T = \frac{nm}{n+m}\omega^2 =
+        \frac{U}{n m (n+m)} - \frac{4 m n - 1}{6(m+n)}
+
+    where :math:`U` is defined as below. The function :math:`r()` here denotes
+    the rank of the element within the pooled sample of size :math:`n + m`, with
+    ties assigned mid-rank values:
+
+    .. math::
+        U = n \sum_{i=1}^n (r(X_i)-i)^2 + m \sum_{j=1}^m (r(Y_j)-j)^2
```
mdhaber: I commented above that $x_i$ was undefined. That was assuming that $X_i$ is supposed to represent a random variable and $x_i$ is our observation of that random variable, an element in the sample. From that perspective, I'd think that $X_i$ and $Y_j$ here should be $x_i$ and $y_j$, which would also need to be defined.

mdhaber (commented):
I'm tempted to merge this anyway, but I think these things are worth noting. If nothing else, it explains why _I_ wouldn't want to sign up for a larger project at this time - it seems simple, but it takes some time to dot the $i$ s and cross the $t$ s.

Let me know if I should let it go or if you'd like to adjust these things.

scottshambaugh:
Tightened things up a bit per your suggestions!

mdhaber:
Thanks. Note that I can't easily see what changed due to the rebase + force push. Since there is rarely a reason not to squash merge these days, I'd recommend merge + regular push after review begins.

mdhaber (approved):
LGTM. I'll merge when CI finishes and I can see the latest render of the equations (although it seems those are mostly unchanged since last time, so they should be fine). Thanks @scottshambaugh.

PR #24029, opened 2025-11-19
ENH: stats.mood: add array API support

mdhaber:
#### Reference issue
Toward gh-20544

Adds array API support to `scipy.stats.mood`.

### DIFF
--- a/scipy/stats/tests/test_morestats.py
+++ b/scipy/stats/tests/test_morestats.py
@@ -1226,11 +1226,10 @@ class TestMood:
             title "Data for above results";
           run;
         """
-        dtype = xp.float64
-        x, y = xp.asarray(x, dtype=dtype), xp.asarray(y, dtype=dtype)
+        x, y = xp.asarray(x.tolist()), xp.asarray(y.tolist())
         statistic, pvalue = stats.mood(x, y, alternative=alternative)
-        xp_assert_close(statistic, xp.asarray(stat_expect, dtype=dtype), atol=1e-16)
-        xp_assert_close(pvalue, xp.asarray(p_expect, dtype=dtype), atol=1e-16)
+        xp_assert_close(statistic, xp.asarray(stat_expect), atol=1e-16)
+        xp_assert_close(pvalue, xp.asarray(p_expect), atol=1e-16)
 
     @pytest.mark.parametrize("dtype", [None, 'float32', 'float64'])
     @pytest.mark.parametrize("alternative, expected",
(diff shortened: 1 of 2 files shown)


scipy/stats/tests/test_morestats.py
```diff
@@ -1225,37 +1226,48 @@ class X;
             title "Data for above results";
           run;
         """
+        dtype = xp.float64
```
tylerjereddy: Since you test with "default dtypes" below, I was somewhat curious why this was being pinned to a single dtype here. Maybe this best reflects a match to what the original test did.

FWIW, the test seems to pass with array API locally even if you parametrize this to try "float32" and "float64".

mdhaber: The only thing we can't do is ignore dtypes entirely - the `x` and `y` arrays are delivered as `float64` yet the reference values arrive as Python scalars, so if we don't do something, there will be a type mismatch when the backend has a default dtype of `float32`. I chose this way of resolving it, but we can just as well remove the dtypes and convert `x` and `y` to lists before passing them to `asarray`, or we can parametrize. I parametrized other tests below, so the function itself is respecting dtypes.

If you prefer:
```suggestion
```

scipy/stats/tests/test_morestats.py
```diff
@@ -1225,37 +1226,48 @@ class X;
...
+        x, y = xp.asarray(x, dtype=dtype), xp.asarray(y, dtype=dtype)
```
mdhaber:
```suggestion
        x, y = xp.asarray(x.tolist()), xp.asarray(y.tolist())
```

scipy/stats/tests/test_morestats.py
```diff
@@ -1225,37 +1226,48 @@ class X;
...
         statistic, pvalue = stats.mood(x, y, alternative=alternative)
-        assert_allclose(stat_expect, statistic, atol=1e-16)
-        assert_allclose(p_expect, pvalue, atol=1e-16)
+        xp_assert_close(statistic, xp.asarray(stat_expect, dtype=dtype), atol=1e-16)
+        xp_assert_close(pvalue, xp.asarray(p_expect, dtype=dtype), atol=1e-16)
```
mdhaber:
```suggestion
        xp_assert_close(statistic, xp.asarray(stat_expect), atol=1e-16)
        xp_assert_close(pvalue, xp.asarray(p_expect), atol=1e-16)
```

j-bowhay (approved):
LGTM

Issue #23950, opened 2025-11-07
BUG: stats.anderson(dist=’norm’) tabulated critical values don’t match significance levels


tommie979:
### Describe your issue.

stats.anderson(dist=’norm’) tabulated critical values don’t match significance levels
For the Anderson-Darling test, the tabulated critical values are based on (Stephens, 1974) . These critical values are widely used and are also found on the wiki for the AD test  ([Anderson–Darling test - Wikipedia](https://en.wikipedia.org/wiki/Anderson%E2%80%93Darling_test), table below note 4). When generating normally distributed data, the probabilities of false alarm are almost always smaller than the associated significance levels (This is especially true for significance levels 10-1%). This issue was also mentioned in the comments  on issue scipy #10737 by ```mdhaber```. A solution would be to update the critical values and modified AD test statistic as mentioned on p.123, table 4.7 (D’Agostino, 1986) .

I have provided a script that compares the critical values and modified AD test statistics from both sources and in addition does a Monte Carlo estimation of the critical values. We see the estimated critical values are much closer to those of D’Agostino (1986). I have only considered the normal distribution here. Other critical values for other distributions such as exponential, gumbel, logistic, and Weibull come from different sources, but it may prove useful to check these in a similar fashion.

References:

D’Agostino, Ralph B. (1986). “Tests for the Normal Distribution”. In: Goodness-of-Fit Techniques. Ed. by Ralph B. D’Agostino and Michael A. Stephens. New York: Marcel Dekker, pp. 122–128. ISBN: 0-8247-7487-6.

Stephens, M. A. (1974). “EDF Statistics for Goodness of Fit and Some Comparisons”. In: Journal of the American Statistical Association 69.347, pp. 730–737. DOI: 10.1080/01621459.1974.10480196.

The example code for 100000 chunks of size 100 took around 27s to run on my machine and prints:

```
Using Stephens, M.A. (1974) critical values and modified statistic:
Chunks that fail the 15% Anderson-Darling test: 15.03 % (-0.03 difference)
Chunks that fail the 10% Anderson-Darling test: 9.64 % (0.36 difference)
Chunks that fail the 5% Anderson-Darling test: 4.64 % (0.36 difference)
Chunks that fail the 2.5% Anderson-Darling test: 2.21 % (0.29 difference)
Chunks that fail the 1% Anderson-Darling test: 0.84 % (0.16 difference)
Using D'Agostino (1986) critical values and modified statistic:
[...]

PR #24030, opened 2025-11-20
ENH: stats.anderson: add `method` parameter

mdhaber:
#### Reference issue
Addresses `anderson` part of https://github.com/scipy/scipy/issues/23950#issuecomment-3529544183.

This adds a `method` parameter to `scipy.stats.anderson` to allow the user to choose how to compute a p-value. When `method` is not specified, it warns that the user must opt into a p-value calculation method explicitly, and when `method` is specified, it returns a different result object with attributes `statistic` and `pvalue`. See https://github.com/scipy/scipy/issues/23950#issuecomment-3529544183 for discussion.

#### Additional information
~~Marking as draft until I've come back to take a look, but @nickodell @tommie979 you're welcome to take an early look.~~ This is in pretty good shape; I should just investigate `weibull_min` a little.

### DIFF
--- a/scipy/stats/_morestats.py
+++ b/scipy/stats/_morestats.py
@@ -2229,7 +2229,7 @@ def anderson(x, dist='norm', *, method=None):
         If `method` is an instance of `MonteCarloMetod`, the p-value is computed using
         `scipy.stats.monte_carlo_test` with the provided configuration options and other
         appropriate settings.
-        
+
         .. versionadded:: 1.17.0
             If `method` is not specified, `anderson` will emit a ``FutureWarning``
             specifying that the user must opt into a p-value calculation method.
@@ -2256,7 +2256,7 @@ def anderson(x, dist='norm', *, method=None):
         fit_result : `~scipy.stats._result_classes.FitResult`
             An object containing the results of fitting the distribution to
             the data.
-            
+
         If `method` is provided, this is an object with the following attributes:
 
         statistic : float
@@ -2264,7 +2264,7 @@ def anderson(x, dist='norm', *, method=None):
         pvalue: float
             The p-value corresponding with the test statistic, calculated according to
             the specified `method`.
-            
+
         .. deprecated :: 1.17.0
             The tuple-unpacking behavior of the return object and attributes
             ``critical_values``, ``significance_level``, and ``fit_result`` are
@@ -2452,14 +2452,14 @@ def anderson(x, dist='norm', *, method=None):
         elif isinstance(method, stats.MonteCarloMethod):
             method = method._asdict()
             if method.pop('rvs', False):
-                message = ("The `rvs` attribute of a `MonteCarloMethod` object passed "
-                           "as the `method` parameter of `scipy.stats.anderson` is "
-                           "ignored.")
+                message = ("The `rvs` attribute of a `MonteCarloMethod` object "
+                           "passed as the `method` parameter of `scipy.stats.anderson` "
+                           "is ignored.")
                 warnings.warn(message, UserWarning, stacklevel=2)
             if method.pop('batch', False):
-                message = ("The `batch` attribute of a `MonteCarloMethod` object passed "
-                           "as the `method` parameter of `scipy.stats.anderson` is "
-                           "ignored.")
+                message = ("The `batch` attribute of a `MonteCarloMethod` object "
+                           "passed as the `method` parameter of `scipy.stats.anderson` "
+                           "is ignored.")
                 warnings.warn(message, UserWarning, stacklevel=2)
             method['n_mc_samples'] = method.pop('n_resamples')
 
@@ -2469,7 +2469,6 @@ def anderson(x, dist='norm', *, method=None):
 
             dist = getattr(stats, dist)
             res = stats.goodness_of_fit(dist, x, statistic='ad', **kwargs, **method)
-            print(res.statistic, res.pvalue)
             pvalue = res.pvalue
         else:
             message = ("`method` must be either 'interpolate' or an instance of"
--- a/scipy/stats/_morestats.py
+++ b/scipy/stats/_morestats.py
@@ -2242,6 +2242,14 @@ def anderson(x, dist='norm', *, method=None):
     Returns
     -------
     result : AndersonResult
+        If `method` is provided, this is an object with the following attributes:
+
+        statistic : float
+            The Anderson-Darling test statistic.
+        pvalue: float
+            The p-value corresponding with the test statistic, calculated according to
+            the specified `method`.
+
         If `method` is unspecified, this is an object with the following attributes:
 
         statistic : float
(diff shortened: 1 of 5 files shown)


scipy/stats/_morestats.py
```diff
@@ -2211,11 +2222,27 @@ def anderson(x, dist='norm'):
         The type of distribution to test against.  The default is 'norm'.
         The names 'extreme1', 'gumbel_l' and 'gumbel' are synonyms for the
         same distribution.
+    method : str or instance of `MonteCarloMethod`
+        Defines the method used to compute the p-value.
+        If `method` is ``"interpolated"``, the p-value is interpolated from
+        pre-calculated tables.
+        If `method` is an instance of `MonteCarloMetod`, the p-value is computed using
```
mdhaber:
```suggestion
        If `method` is an instance of `MonteCarloMethod`, the p-value is computed using
```

scipy/stats/tests/test_morestats.py
```diff
@@ -396,6 +397,88 @@ def test_anderson_weibull_As(self):
...
+        method = stats.MonteCarloMethod(rng=np.random.default_rng(23495984828))
+        res = stats.anderson([1, 2, 3, 4, 5], 'norm', method=method)
+        assert res.pvalue != ref.pvalue  # different random state -> different p-value
+
+    @pytest.mark.parametrize('dist_name, seed',
+        [('norm', 4202165767275),
+         ('expon', 9094400417269),
+         pytest.param('logistic', 3776634590070, marks=pytest.mark.xslow),
+         pytest.param('gumbel_l', 7966588969335, marks=pytest.mark.xslow),
+         pytest.param('gumbel_r', 1886450383828, marks=pytest.mark.xslow)])
```
mdhaber, line 444: Tried to add `weibull_min` here but I'm not sure it's working. Will investigate further.

scipy/stats/_morestats.py
```diff
@@ -2192,8 +2192,19 @@ def dnllf(params):
...
 
+_anderson_warning_message = (
+"""As of SciPy 1.17, users must choose a p-value calculation method by providing the
+`method` parameter. `method='interpolate'` interpolates the p-value from pre-calculated
+tables; `method` may also be an instance of `MonteCarloMethod` to approximate the
+p-value via Monte Carlo simulation. When `method` is specified, the result object will
+include a `pvalue` attribute and not attributes `critical_value`, `significance_level`,
+or `fit_result`. Beginning in 1.19.0, these other attributes will no longer be
+available, and a p-value will always be computed according to one of the available
+`method` options.""".replace('\n', ''))
```
mdhaber:
```suggestion
`method` options.""".replace('\n', ' '))
```

scipy/stats/tests/test_morestats.py
mdhaber: With `weibull_min`, the results of Monte Carlo simulation and tabulated values can agree, but it depends on the data. For the cases in which they disagree, I don't think there's anything wrong with how it's implemented now in `anderson` or even in `goodness_of_fit`. Ultimately, the problem is that `rv_continuous.fit` converges to values of the parameters that are inconsistent with the data (e.g. `loc` greater than the minimum of the data) without warning or error. In this case, the AD statistic is infinite (infinite disagreement between data and distribution CDF), and so these infinities end up in the null distribution. Consequently, the p-value returned by `goodness_of_fit` tends to be inflated (conservative).

Here is some example data that triggers the problem:
<details>

```python3
import numpy as np
from scipy import stats
x = [1.0607969176105423, 0.7563616035154184, 0.9733315310979942, 0.843163866274454, 0.9042875943469282, 0.6721504043579607, 1.231839969987856, 1.1213643342727713, 0.896359928853492, 0.724510085963644, 1.176209727329505, 0.9583033976574415, 0.6664495656675816, 0.9850260667145947, 0.539631766998343, 0.9900945808551318, 0.79655449430418, 0.9578997265085347, 0.7690034421284264, 0.8573532711453653, 0.8824637181642769, 0.8328378288153777, 0.7238429430863027, 0.7174785862638946, 0.7554988587111076, 0.8320872956031995, 0.7039544092287224, 1.3211158691587954, 0.7725398348986974, 0.8593009091028263, 0.7644391412960172, 1.0227547699513122, 1.1305821846516872, 0.990432287377093, 0.7878093371868877, 0.875800789090309, 0.681306267860599, 1.4086114454927998, 0.9744642186352348, 0.935361043427666, 0.8565027284314886, 0.7749878194330573, 0.5643198504744864, 0.9537180031559482, 0.2502179231576509, 1.3747794415622552, 0.7879793581394539, 0.5485266231617405, 0.8573829234059458, 1.12945753745936]
params = stats.weibull_min.fit(x)
print(np.min(x), params[1])  # the minimum of the data is less than `loc`
print(stats.weibull_min.nnlf(params, x))  # infinite negative log-likelihood (impossible)
```

</details>

I don't think there's anything that needs to be done in this PR about it. In a separate PR, `weibull_min.fit` could be improved or `rv_continuous.fit` could raise when the parameters are inconsistent with the data.

scipy/stats/tests/test_resampling.py
```diff
@@ -1022,22 +1022,19 @@ def statistic1d(x):
...
-        # statistic. `i` used to index this list.
+    @pytest.mark.parametrize('dist_name', ['norm', 'logistic'])
+    @pytest.mark.parametrize('target_statistic', [0.6, 0.7, 0.8])
+    def test_against_anderson(self, dist_name, target_statistic):
+        # test that monte_carlo_test can reproduce results of `anderson`.
 
-        # find the skewness for which the sample statistic matches one of the
-        # critical values provided by `stats.anderson`
+        # find the skewness for which the sample statistic is within the range of
+        # values tubulated by `anderson`
```
nickodell, line 1031:
```suggestion
        # values tabulated by `anderson`
```

steppi (approved):
I think this seems like a reasonable plan for getting anderson and anderson_ksamp to have outputs more consistent with other hypothesis tests and am support of merging this PR and https://github.com/scipy/scipy/pull/24031. I posted on the discourse discussion (https://discuss.scientific-python.org/t/anderson-anderson-ksamp-improvements-and-return-object-changes/2186/2) that I would like to merge this and https://github.com/scipy/scipy/pull/24031 within 5 days if no one objects.

nickodell (approved):

scipy/stats/tests/test_resampling.py
steppi: I think I'll merge this PR as planned tomorrow even if Matt doesn't accept this suggestion. Since it's just a comment and not user facing, I'm fine with "tubulated". I see it kind of like the flaw left in a Persian rug.

scipy/stats/_morestats.py
```diff
@@ -2211,11 +2222,35 @@ def anderson(x, dist='norm'):
...
         same distribution.
+    method : str or instance of `MonteCarloMethod`
+        Defines the method used to compute the p-value.
+        If `method` is ``"interpolated"``, the p-value is interpolated from
+        pre-calculated tables.
+        If `method` is an instance of `MonteCarloMethod`, the p-value is computed using
+        `scipy.stats.monte_carlo_test` with the provided configuration options and other
+        appropriate settings.
+
+        .. versionadded:: 1.17.0
```
j-bowhay, line 2233: 1.17 has now been branch so this will either need to be backported or the version updated

scipy/stats/tests/test_resampling.py
mdhaber: Fixing in gh-26166.

mdhaber: NM, leaving it in : )

PR #24031, opened 2025-11-20
ENH: stats.anderson_ksamp: add `variant` parameter

mdhaber:
#### Reference issue
Addresses `anderson_ksamp` part of https://github.com/scipy/scipy/issues/23950#issuecomment-3529544183.

This adds a `variant` parameter to `scipy.stats.anderson` to allow the user to choose which of the *three* versions of the test are used. (Previously, only two were available via the `midrank` parameter.) When `variant` is not specified, it warns that the user must use the `variant` parameter explicitly, and when `variant` is specified, it returns a different result object with attributes `statistic` and `pvalue` only. See https://github.com/scipy/scipy/issues/23950#issuecomment-3529544183 for discussion.


steppi (approved):
I think this seems like a reasonable plan for getting `anderson` and `anderson_ksamp`  to have outputs more consistent with other hypothesis tests and am support of merging this PR and https://github.com/scipy/scipy/pull/24030. I posted on the discourse discussion (https://discuss.scientific-python.org/t/anderson-anderson-ksamp-improvements-and-return-object-changes/2186/2) that I would like to merge this and #24030 within 5 days if no one objects.

steppi:
@mdhaber, I'll merge this after the conflicts are resolved.

scipy/stats/_morestats.py
```diff
@@ -2593,6 +2639,10 @@ def anderson_ksamp(samples, midrank=True, *, method=None):
         critical_values : array
             The critical values for significance levels 25%, 10%, 5%, 2.5%, 1%,
             0.5%, 0.1%.
+
+            .. deprecated::1.17.0
```
j-bowhay, line 2643: As per https://github.com/scipy/scipy/pull/24030, I think we either need to update the version or backport this

mdhaber: Since it involves a deprecation, it would be nice to backport both, but I won't push too hard. @tylerjereddy is this doable? I can suggest updates to the release notes.

tylerjereddy: Ok, I put the backport label on the other one. It is doable but a bit frowned upon to backport enhancement PRs.

It does increase the likelihood that I'd need to do an RC3 (probably still a low chance, but would increase it). That said, I have a little funding to do releases, so if this helps the team it may be a good use of bandwidth.

PR #22070, opened 2024-12-13
ENH: `_lib`: JAX support (non-jitted)

crusaderky:
- Bump array-api-compat to 1.10 (needed for https://github.com/data-apis/array-api-compat/pull/205)
- Bump array-api-extra to 0.5.0 (needed for https://github.com/data-apis/array-api-extra/pull/53)
- Support JAX in `_lib` and all its unit tests. This doesn't include `@jax.jit` yet (read comments below).

PR #24034, opened 2025-11-21
ENH: stats: enable _axis_nan_policy decorator for JAX; remove test skips

mdhaber:
#### Reference issue
-

JAX arrays were not allowed to pass through the `_axis_nan_policy` decorator. Consequently, many tests with JAX involving use of `axis`, `keepdims`, zero-length slices, etc., had to be skipped.

This PR allows JAX arrays through the decorator but avoids data-dependent branching with lazy arrays. This allows us to remove many of these skips.

#### Additional information
I will take another pass in a separate PR to see where we can remove `jax_jit=False` from `xp_capabilities`.

I am confused about the following:
- `is_lazy_array` returns `True` for JAX arrays even when `jax_jit=False` (and therefore JAX arrays are actually eager). This forces an unfortunate choice - we have some functions that work with `jax_jit=True` only if we skip data-dependent branching using `is_lazy_array`, but then this breaks functions that work with `jax_jit=False` only if we have the data-dependent branching. I understand that JAX without JIT is not really JAX, but getting tests passing with `jax_jit=False` is a useful stepping stone on the way to `jax_jit=True`. Can we make `is_lazy_array` more accurate?
- We have `eager_only=True` in `pytest.mark.skip_xp_backends`, but not `jax_jit=False`. How can we get the more specific behavior of `jax_jit=False` with the `skip_xp_backends` decorator?


scipy/stats/tests/test_stats.py
```diff
@@ -3525,16 +3528,21 @@ def test_scale(self, xp):
...
         # # Bad scale
         message = "foobar not a valid scale for `iqr`"
         with pytest.raises(ValueError, match=message):
             stats.iqr(x, scale='foobar')
 
+    @pytest.mark.skip_xp_backends("jax.numpy", reason="lazy -> no nan_policy")
+    def test_scale_nan_policy_omit(self, xp):
+        x = xp.reshape(xp.arange(15.0), (3, 5))
+        x = xpx.at(x)[1, 2].set(xp.nan)
+        nan = xp.asarray(xp.nan)
```
mdhaber:
```suggestion
```

scipy/stats/_axis_nan_policy.py
```diff
@@ -566,16 +566,17 @@ def hypotest_fun_out(*samples, **kwds):
                     # Behave as though there are no NaNs (even if there are)
                     contains_nan = [False] * len(samples)
 
+                any_contains_nan = not is_jax(xp) and any(contains_nan)
```
steppi: This could use an explanatory comment to make this easier to follow for someone who isn't aware of the context.

steppi, line 576: The `not is_jax` seems like it could use an explanatory comment. It might seem a little mysterious to someone who isn't aware of the broader context. Do I understand correctly that  `nan_policy` is effectively ignored for the JAX backend, and the results end up being whatever happens with JAX? If so, that seems like reasonable behavior. I recommend documenting it with the `extra_note` field of `xp_capabilities` which was added recently. See here for an example:

https://github.com/scipy/scipy/blob/d1af6247ea2c79aadf86cf26beb3af07a0a2488e/scipy/interpolate/_rbfinterp.py#L57-L60

scipy/stats/_stats_py.py
```diff
@@ -4061,7 +4061,7 @@ class AlexanderGovernResult:
     pvalue: float
 
 
-@xp_capabilities()
+@xp_capabilities(jax_jit=False)
```
steppi, line 4070: For all these cases where `jax_jit=False` had to be added to `xp_capabilities`, did these functions originally work with the `jit` before this PR, or is it just that `jax_jit=False` was never added because all relevant tests were skipped?

steppi (commented):
If I understand everything correctly here, I think this looks like a good change. I completely agree that it would be good if `skip_xp_backends` could get an explicit `jax_jit` kwarg. I can look into making that happen. Before thinking about changing `is_lazy_array` to actually check if a JAX array is a tracer or a concrete array, I guess we should discuss whether it's OK to branch based on whether something is concrete or not, since the documentation of `is_lazy_array` seems to suggest this is bad practice. I don't think I know enough yet to make a call on this, but if we decide it's fine in some cases to take different code paths for concrete arrays vs tracers, it seems reasonable to have a cautious API like `is_lazy_array` is now, as well as an unsafe one that we try to use carefully. Perhaps this could be done with a `kwarg` to `is_lazy_array`?

ev-br:
> changing is_lazy_array to actually check if a JAX array is a tracer or a concrete array, I guess we should discuss whether it's OK to branch based on whether something is concrete or not, since the documentation of is_lazy_array seems to suggest this is bad practice. 

cc @crusaderky, you had a discussion with Jax devs about telling if an array is material or abstract, or am I hallucinating like an llm

rgommers:
That was discussed, pretty extensively in a couple of places (e.g., see https://github.com/scipy/scipy/pull/22070#discussion_r1883971831). Please do not spend time trying to distinguish between JAX eager and jit, it's not possible with public API and it is undesirable.

scipy/stats/_axis_nan_policy.py
rgommers: > Do I understand correctly that `nan_policy` is effectively ignored for the JAX backend, and the results end up being whatever happens with JAX? If so, that seems like reasonable behavior.

I don't quite understand this. An error seems much more logical than an incorrect answer.

mdhaber:
>  Please do not spend time trying to distinguish between JAX eager and jit, it's not possible with public API and it is undesirable.

OK, then can you address the concern?

SciPy has some heavily-used code-paths that want to do data-dependent branching, e.g. [`_contains_nan`](https://github.com/scipy/scipy/blob/783b39b7d77364368c33f50f5bc19033d7e44878/scipy/_lib/_util.py#L841). We have some functions that work with `jax_jit=True` only if we skip the data-dependent branching using `is_lazy_array`. This breaks features of functions that work with `jax_jit=False` only if we do the data-dependent branching. What is the prefered solution? E.g., `_contains_nan` get an extra argument that controls whether JAX skips the data-dependent logic? Where is SciPy headed w.r.t. data-based input validation of arrays?

scipy/stats/_axis_nan_policy.py
mdhaber: `nan_policy='raise'` means "raise an error if there is a NaN in the array". My understanding is that this is not possible if the array is lazy. I suppose you mean that JAX should always raise an error if `nan_policy='raise'` is specified, regardless of whether the array contains NaNs or not.

We can do that, it's just going to require a lot of test skips. Sometimes, especially with old test, tests of `nan_policy` options are not written separately, and it may require a fair amount of refactoring to untangle things.

> An error seems much more logical than an incorrect answer.

The answer would not be incorrect; it would either be correct (if no NaNs are present) or NaN (if NaNs are present, since the default behavior is to propagate NaNs).

rgommers:
> This breaks features of functions that work with `jax_jit=False` only if we do the data-dependent branching. What is the prefered solution?

Just letting it raise an exception is fine. There is no point putting in extra work or code paths specifically to make something work with JAX eager. JAX eager is really slow and its main purpose is debugging/development so that when the JIT is turned on things work (and with high performance). 

In some cases it _might_ be possible to make functionality that is inherently value-dependent work with `jax.jit`, but I think that will require using compiled code that handles the if-else branches on those data values, and that the JIT machinery treats as a single call without tracing away one of the branches. But that's a lot of complexity and we haven't done this yet. It's fine to keep the test skips for the time being I'd say.

> Where is SciPy headed w.r.t. data-based input validation of arrays?

For regular input validation we just skip it for JAX as a whole. Regular input validation is "emit a warning or raise an exception in case of ...". `nan_policy` isn't that though, it actually changes the answer based on the values - that's quite different and shouldn't be skipped.

scipy/stats/_axis_nan_policy.py
rgommers: Oh in that case, propagating may be fine. 

If `'omit'` would actually yield a different numerical value for some code, that'd be more problematic.

scipy/stats/_stats_py.py
mdhaber: The `alexandergovern` and `ttest_ind_from_stats` `jax_jit=False` might not be necessary after all; they are now passing with JIT locally. Perhaps there were changes between when I added `jax_jit=False` and now that made it work again; I'm not sure what was going on.

The `ttest_ind` `jax_jit=False` is necessary because `ttest_ind`'s code for unpacking the result object as a tuple has some code that raises during tracing. That code only runs now that we allow the decorator to do its thing.

scipy/stats/_axis_nan_policy.py
mdhaber: > If `'omit'` would actually yield a different numerical value for some code, that'd be more problematic.

Right, but If the user passes `nan_policy='omit'`, then the function would either properly omit the NaNs or ignore the provided `nan_policy` and propagate NaNs.

This assumes that when NaNs are propagated through the code, the result is NaN. That is *usually* true, but not always. That is why the decorator looks out for NaNs even when `nan_policy='propagate'` and *ensures* that the output is NaN if there are any in the input. But currently, *all* non-numpy backends - not just JAX - skip that step for multidimensional arrays, and that shortcoming is unrelated to this change.

So maybe the resolution here is to:
- comment about the change as requested,
- document the (known) issues related to `nan_policy` for non-NumPy backends,
- fix those issues in future PRs.

> Do I understand correctly that nan_policy is effectively ignored for the JAX backend, and the results end up being whatever happens with JAX?

Right. I think the only *new* possibility of getting an incorrect result is 
- if the user opts into the experimental `ARRAY_API_STRICT=1` AND
- if the call involves only 1D JAX arrays AND 
- if one of those arrays has a NaN, AND
- if the NaN does not propagate to the output naturally, AND
- if the user specifies `nan_policy` explicitly,
they will now get an incorrect result rather than seeing an error that `nan_policy` is not supported. 

Proportionally, this is a small addition to the set of cases in which NaNs can conveivably cause an incorrect result with non-NumPy backends - in `main` it could happen if a non-NumPy array is multidimensional, or it could happen with 1D JAX arrays if the user doesn't specify `nan_policy` explicitly. Moreover, the set of cases in which NaNs can conceivably cause an incorrect result with non-NumPy backends is already quite small - most functions propagate NaNs to the output naturally. So yeah, I think it's fair to document that currently, JAX ignores `nan_policy` entirely, and that NaNs might propagate naturally, or they might not. `extra_note` seems like a good place for that information.

If we agree that it's OK for JAX arrays to propagate NaNs with `nan_policy='omit'` and `nan_policy='raise'` rather than raising an error about not supporting `nan_policy`, I think we can ensure (in a follow-up) that JAX *always* does the extra work to ensure that NaNs propagate to the output in a JIT-compatible way. It's just not the default for eager arrays because it wastes time.

mdhaber:
> Just letting it raise an exception is fine. There is no point putting in extra work or code paths specifically to make something work with JAX eager. JAX eager is really slow and its main purpose is debugging/development so that when the JIT is turned on things work (and with high performance).

My point is that letting it raise is not fine - raising an exception interferes with using JAX eager mode for debugging.

> `nan_policy` isn't that though, it actually changes the answer based on the values - that's quite different and shouldn't be skipped.

`nan_policy='raise'` *is* that, so it sounds like it's OK to treat that just like `nan_policy='propagate'`, instead.
`nan_policy='propagate'` is typically already respected, even for JAX.
`nan_policy='omit'` is typically ignored and behaves like `nan_policy='propagate'` for JAX.

We can work on improving JAX support for `nan_policy='propagate'` and `nan_policy='omit'` in a JIT-compatible way. In the meantime, it would be nice to be able to use eager mode for debugging without a bunch of test refactoring/skipping.

rgommers:
> raising an exception interferes with using JAX eager mode for debugging.

The debugging that is useful _is_ for making JAX work in JIT mode (you are better off just using `numpy` otherwise). You are arguing from the premise that it's good to make JAX eager work by changing code, even though that doesn't help for `jax.jit`. This is not a useful goal.

scipy/stats/_axis_nan_policy.py
```diff
@@ -437,7 +437,7 @@ def axis_nan_policy_wrapper(*args, _no_deco=False, **kwds):
             else:
                 temp = args[0]
 
-            if is_lazy_array(temp):
+            if is_dask(array_namespace(temp)):
```
rgommers, line 440: This change just goes in the wrong direction in my opinion. It makes the error with `jax.jit` less clear, and makes it harder to either add support for another kind of lazy array or to drop Dask support.

steppi:
Thanks @rgommers for weighing in. Since the intention of JAX eager mode is for debugging something which one wants to run in jit mode, do I understand correctly that no effort should be spent on making things work with JAX eager mode that are not supported in jit mode? If it just naturally happens that a function which converts to NumPy and back works with JAX eager mode on CPU, that can be documented as supported in `xp_capabilities` with `jax_jit=False`, but beyond that, if special code paths are needed to make something work in eager mode, there's no point in doing that? That sounds right to me.

I guess this PR should be shelved? It seems like it may not actually that important to remove these skips.

mdhaber:
> The debugging that is useful is for making JAX work in JIT mode (you are better off just using numpy otherwise)

There is other debugging to be done other than dealing with JIT, e.g. immutability and lack of special functions. 

> I guess this PR should be shelved?

Please let's not be too hasty. _I_ am knee deep in trying to add support for all array backends in `stats` - JAX JIT included. This is part of my process. I think others are welcome to their opinion about what is useful - hopefully I am, too? 

If there is concern about other effects of this PR, such as allowing extra tests to be executed with JAX in eager mode and therefore requiring more CI time, or this adds new possibilities for incorrect outputs (in extreme cases when using an experimental backend), I can try to address those. In any case, please state whether you want your comments to be considered blocking here. I usually just fret and pour time into addressing these sorts of objections, but I'd rather not spend the rest of the weekend here if I can avoid it.

scipy/stats/_axis_nan_policy.py
mdhaber: > It makes the error with `jax.jit` less clear

Do most functions that don't work with `jax.jit` emit a clear error message? IIUC, we are in a transition period, and the edges w.r.t. to JIT are not all polished at this moment. We will polish them before the experimental backends are announced as fully supported, but in the meantime, we can make progress incrementally.

> and makes it harder to either add support for another kind of lazy array or to drop Dask support.

I'm not seeing seeing that. If you have a new kind of lazy array that works more like JAX, it will just work without changes to this code. If you have a new kind of lazy array that works more like Dask, it can be added to this condition.
If you want to drop Dask support, this line is just unnecessary and can be removed. It will not get in the way otherwise.
These don't seem hard. And while we're dealing in hypotheticals, I would be the one dealing with these scenarios. I am comfortable with this burden.

steppi:
> > I guess this PR should be shelved?
> 
> Please let's not be too hasty. _I_ am knee deep in trying to add support for all array backends in `stats`. This is part of my process. I think others are welcome to their opinion about what is useful - hopefully I am, too?
> 
> If there is a real concern about what this PR does, such as allowing extra tests to be executed with JAX in eager mode and therefore requiring more CI time, or this adds new possibilities for incorrect outputs (in extreme cases when using an experimental backend), I can try to address those. Also please state whether you want your comments to be considered blocking here. I usually just fret and pour time into addressing these sorts of objections, but I'd rather not if I can avoid it.

I wouldn't consider my comments blocking if @rgommers thought it was OK to merge as is. I'm not confident enough to give an approval on this one though, even if the behavior is documented better. I interpreted Ralf's comments as potentially implying this PR should be shelved and was just trying to get confirmation if that's the case. I'd like to hold off on looking at this PR for a bit; I don't really have bandwidth to spend much time looking at details at this level yet, and am still focused on general infra and the larger picture. Is this blocking for any of your planned work? I know you have a large queue of array API related PRs in need of review, and can spend a little time each week chipping away at these, but will try to focus on things that don't require decision making beyond what I feel can accomplish given the amount of time and knowledge I currently have.

mdhaber:
> I interpreted Ralf's comments as potentially implying this PR should be shelved and was just trying to get confirmation if that's the case.

Right, I'm mostly asking for that to be made explicit if that's the case. Hopefully not.

> Is this blocking for any of your planned work?

No. It will mean that I will ignore JAX, instinctively using adding `jax.numpy` to `skip_backends`, on all remaining array API PRs. I will not use `at.set` in translations, etc., and any form of special considerations for JAX, JIT-related or otherwise, will have to be considered from scratch at some later time rather than me getting us partway there now.

j-bowhay:
@rgommers is there anything that can be done to move this forward. Reading the above discussion I'm not sure I follow why this can't be used as a stepping stone towards Jax JIT support

rgommers:
Please consider my comments non-blocking. I don't think this change is all that useful nor is there much receptiveness to review feedback, but this PR is also not actively harmful (or if so, in quite minor ways), so if you want to merge it go ahead. 

> Please do not spend time trying to distinguish between JAX eager and jit, it's not possible with public API and it is undesirable.

For the record, this is the essential comment that would be a hard blocker (but that's not happening in this PR, it was only a suggestion here).

scipy/stats/_axis_nan_policy.py
mdhaber, line 576:
```suggestion
                # To give JAX most benefits of the `_axis_nan_policy` decorator without
                # value-dependent branching, the decorator will always treat JAX arrays
                # as if there are no NaNs. Typically, this means that NaNs will
                # multidimensional naturally. However, there may be cases in which this
                # doesn't produce the desired result, so for now, JAX users should
                # treat all functions as though there is no explicit`nan_policy`.
                # Future work tracked in gh-14651.
                any_contains_nan = not is_jax(xp) and any(contains_nan)
```

mdhaber:
So as not to spend more time on this, I'll go ahead and merge. Future work with `nan_policy` and non-NumPy backends is tracked in gh-14651.

PR #24036, opened 2025-11-21
CI: run Windows OneAPI job on main every 2 days to prefetch cache

nickodell:
#### Reference issue

See #21491

PR #24016 moved the Linux Intel OneAPI job to run on main. This successfully ran today. This reduced the number of cache entries seen for the Linux job - there is now one when there were 20 Intel ones, half of which were Linux OneAPI, a few days ago.

The Windows job is the [next biggest user of cache](https://github.com/scipy/scipy/actions/caches?query=sort%3Asize-desc+key%3Ainstall-https), so do the same thing to it.

### DIFF
--- a/.github/workflows/windows_intel_oneAPI.yml
+++ b/.github/workflows/windows_intel_oneAPI.yml
@@ -9,6 +9,14 @@ on:
     branches:
       - main
       - maintenance/**
+  schedule:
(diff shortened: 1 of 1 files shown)


rgommers (approved):
LGTM, thanks @nickodell. And yes, cache usage is starting to look good, it's at 9.7 GB right now which is right at the limit.

PR #24037, opened 2025-11-21
ENH: Return value from spsolve matches input RHS shape and sparsity

broesler:
The existing `scipy.sparse.linalg.spsolve` function returns a 1D array when the right-hand side `b` is provided as a 2D column vector (shape `(N, 1)`). This behavior is inconsistent with `numpy.linalg.solve` and `scipy.linalg.solve`, as well as the `solve` method of the SuperLU object from `scipy.sparse.linalg.splu`, all of which return a 2D array (shape `(N, 1)`) in that case. 

This PR modifies `spsolve` to return a 2D array when `b` is 2D. In addition, it now returns a sparse array when `b` is sparse, and a dense array when `b` is dense. This behavior differs from `SuperLU.solve`, but is more intuitive since it is consistent with the input type.

These changes also apply to the `use_umfpack=True` option of `spsolve`.

#### Setup
```python
>>> import numpy as np
>>> from scipy import linalg as la
>>> from scipy.sparse import linalg as spla
>>> from scipy import sparse
>>> A = sparse.random_array((5, 5), density=0.8, format='csc')
>>> A.setdiag(A.diagonal() + 5)  # ensure non-singular
>>> expect_x = np.arange(1, 6)
>>> b = A @ expect_x
>>> type(b)
=== numpy.ndarray
>>> b.shape
=== (5,)
>>> b = b.reshape((-1, 1))
>>> b.shape
=== (5, 1)
```

#### Existing Behavior
```python
>>> # Dense solve
>>> x_np = np.linalg.solve(A.toarray(), b)
>>> x_la = la.solve(A.toarray(), b)
>>> x_sp = spla.spsolve(A, b)
>>> x_np.shape
=== (5, 1)
>>> x_la.shape
=== (5, 1)
>>> x_sp.shape
=== (5,)
>>> type(x_sp)
=== numpy.ndarray
>>> # Sparse solve
>>> b_sp = sparse.csc_array(b)
>>> x_sp = spla.spsolve(A, b_sp, use_umfpack=False)
>>> x_sp.shape
=== (5,)
>>> type(x_sp)
=== numpy.ndarray
>>> x_sp = spla.spsolve(A, b_sp, use_umfpack=True)
>>> x_sp.shape
=== (5,)
>>> type(x_sp)
=== numpy.ndarray
```

#### New Behavior
```python
>>> x_np = np.linalg.solve(A.toarray(), b)
>>> x_la = la.solve(A.toarray(), b)
>>> x_sp = spla.spsolve(A, b)
>>> x_np.shape
=== (5, 1)
>>> x_la.shape
=== (5, 1)
>>> x_sp.shape
=== (5, 1)
>>> type(x_sp)
=== numpy.ndarray
>>> # Sparse solve
>>> b_sp = sparse.csc_array(b)
>>> x_sp = spla.spsolve(A, b_sp, use_umfpack=False)
>>> x_sp.shape
=== (5, 1)
>>> type(x_sp)
=== scipy.sparse._array.csc_array
>>> x_sp = spla.spsolve(A, b_sp, use_umfpack=True)
>>> x_sp.shape
=== (5, 1)
>>> type(x_sp)
=== scipy.sparse._array.csc_array
```

#### Existing SuperLU behavior (unchanged)
```python
>>> lu = spla.splu(A)  # SuperLU factorization
>>> # Dense solve
>>> x_lu = lu.solve(b)
>>> x_lu.shape
=== (5, 1)
>>> type(x_lu)
=== numpy.ndarray
>>> # Sparse solve
>>> lu.solve(b_sp)
Cell In[17], line 1
----> 1 lu.solve(b_sp)

TypeError: argument 1 must be numpy.ndarray, not csc_array
>>>
```

#### Existing scikit-umfpack behavior (unchanged)
```python
>>> from scikits import umfpack
>>> f = umfpack.UmfpackLU(sparse.csc_matrix(A))  # umfpack fails on csc_array
>>> x_umf = f.solve(b)
>>> x_umf.shape
=== (5, 1)
>>> type(x_umf)
=== numpy.ndarray
>>> x_umf_sp = f.solve(b_sp)
>>> x_umf_sp.shape
=== (5, 1)
>>> type(x_umf_sp)
=== numpy.ndarray
```

### DIFF
--- a/scipy/sparse/linalg/_dsolve/linsolve.py
+++ b/scipy/sparse/linalg/_dsolve/linsolve.py
@@ -223,11 +223,11 @@ def spsolve(A, b, permc_spec=None, use_umfpack=True):
         warn('spsolve requires A be CSC or CSR matrix format',
              SparseEfficiencyWarning, stacklevel=2)
 
-    # b is a vector only if b have shape (n,) or (n, 1)
+    # b is a vector only if b have shape (n,)
     b_is_sparse = issparse(b)
     if not b_is_sparse:
         b = asarray(b)
-    b_is_vector = ((b.ndim == 1) or (b.ndim == 2 and b.shape[1] == 1))
+    b_is_vector = b.ndim == 1
 
     # sum duplicates for non-canonical format
     A.sum_duplicates()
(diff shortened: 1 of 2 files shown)


j-bowhay:
I am not sure this makes sense since the solution will be dense

dschult:
I'm pretty sure the superLU solver returns a dense vector because that's what we should expect -- most often, solving a sparse system will give a dense result.

As for returning 1D vector solutions when the input rhs vector is 2D, I can see how you might want to mimic the behavior of dense solvers like `nump.linalg` and `scipy.linalg`. But the sparse world has been inherently 2D for a long time. If the solution is a vector, someone had to decide whether to return it as a 1D, or a 2D column. They chose a 1D. We could argue about that choice. But it is going to be tricky for users who have written their code assuming the result will be 1D to go back now and change it to handle a 2D column. We have to think about backward compatibility. If we expect people to make a switch -- then what is the advantage they get from making that switch?

I would say anyone who wants this new behavior can write: `spsolve(A, b).reshape(b.shape)`
I guess after this change, folks wanting the old behavior would change `spsolve(A, b)` to: `spsolve(A, b.toarray().ravel())`.

Do you have a motivating case for needing the result to match the shape of `b`? 
And can you think of a way that would allow old code to "just work" if we change the return shape of the solver?

broesler:
I'll concede that the result for a sparse RHS tends to be dense, although I am working on an analysis of the problems in the [SuiteSparse Matrix Collection](https://sparse.tamu.edu) to see how true this is in "real" problems. NB: there does not appear to be metadata on the RHS or solutions to fully-defined problems, so it may be more tedious than is worth it to download and solve each one.

A counter-argument on sparsity is that `spsolve` returns a sparse result when `b` is a sparse matrix (*i.e.* `b.shape[1] > 1`). It seems inconsistent that a 1-column matrix should be treated differently.

I don't have a particular motivating example about the shape other than my own experience getting into sparse computations and being surprised that the shape of sparse output did not match that of the dense versions. It is also surprising that the shape of the `spsolve` output is different from that of both `SuperLU.solve` and `scikit-umfpack` solve that are used for the underlying computation in `spsolve`. (I have updated the PR description to show the `scikit-umfpack` behavior.)

The `scikit-umfpack` [source](https://github.com/scikit-umfpack/scikit-umfpack/blob/d39202e14e75afc8892fa8edefdce775b8aafda8/scikits/umfpack/interface.py#L63C4-L69C17) also notes that the `spsolve` behavior is a "quirk":

```python
def spsolve(A, b):
    # ... docstring ...
    x = UmfpackLU(A).solve(b)

    if b.ndim == 2 and b.shape[1] == 1:
        # compatibility with scipy.sparse.spsolve quirk
        return x.ravel()
    else:
        return x
```

To cite the Gilbert 1992 paper describing MATLAB's sparse design philosophy:

> The value of the result of an operation does not depend on the storage class of the operands, although the storage class of the result may.[^0]

[0]: Gilbert, J. R., Moler, C., & Schreiber, R. (1992). Sparse matrices in MATLAB: Design and implementation. *SIAM journal on matrix analysis and applications*, 13(1), 333-356.

"value" here meaning the mathematical result of `Ax = b` where `A` is N x N and `b` is N x 1 is `x` that is N x 1, and "storage class" referring to sparse vs. dense.

I definitely understand the backwards compatibility concern. Maybe the shape change is a change that gets bumped to a "SciPy v2.0" update?

PR #24039, opened 2025-11-21
DOC/TST: fft: mark array api coverage with `xp_capabilities`

j-bowhay:
Marks all the functions in `fft` with `xp_capabilities` so that they show up in the compatibility table. Also includes a few testing cleanups/improvements.


lucascolley (commented):
thanks Jake!

1. do we need to be using `lazy_xp_modules` to enable JIT tests?
2. what about other test classes like `TestFFTShift` and those in `test_basic.py`?

j-bowhay:
> thanks Jake!
> 
> 1. do we need to be using `lazy_xp_modules` to enable JIT tests?

Quite possibly! I can never remember how this work...

> 2. what about other test classes like `TestFFTShift` and those in `test_basic.py`?

Sure I can add `make_xp_test_case` to these if you like.

lucascolley:
yeah, that is what gh-24024 is doing

scipy/fft/tests/test_basic.py
```diff
@@ -234,16 +253,18 @@ def _check_axes(self, op, xp):
...
 
+    @make_xp_test_case(fft.hfftn, fft.ihfftn)
     @pytest.mark.parametrize("op", [fft.hfftn, fft.ihfftn])
     def test_axes_non_standard(self, op, xp):
         self._check_axes(op, xp)
 
-    @pytest.mark.parametrize("op", [fft.fftn, fft.ifftn,
-                                    fft.rfftn, fft.irfftn])
+    @make_xp_test_case(fft.fftn, fft.ifftn, fft.rfftn, fft.irfftn)
+    @pytest.mark.parametrize("op", [fft.fftn, fft.ifftn, fft.rfftn, fft.irfftn])
```
ev-br: One thing I can't help noticing is that @steppi used `make_xp_test_param` to parametrize over array api capable functions: https://github.com/scipy/scipy/pull/24024#discussion_r2546403287

It'd be best to converge on a uniform way to handle these, whichever it is (and I don't have a foggiest as to which one).

j-bowhay: Ah thanks I did not know about this!

j-bowhay:
@lucascolley do you understand these failures?
```
FAILED scipy/fft/tests/test_basic.py::TestFFT::test_axes_standard[jax.numpy-fftn] - ValueError: Unknown FFT function: <uarray multimethod 'fftn'>
FAILED scipy/fft/tests/test_basic.py::TestFFT::test_axes_standard[jax.numpy-ifftn] - ValueError: Unknown FFT function: <uarray multimethod 'ifftn'>
FAILED scipy/fft/tests/test_basic.py::TestFFT::test_axes_standard[jax.numpy-rfftn] - ValueError: Unknown FFT function: <uarray multimethod 'rfftn'>
FAILED scipy/fft/tests/test_basic.py::TestFFT::test_axes_standard[jax.numpy-irfftn] - ValueError: Unknown FFT function: <uarray multimethod 'irfftn'>
FAILED scipy/fft/tests/test_basic.py::TestFFT::test_axes_non_standard[jax.numpy-hfftn] - ValueError: Unknown FFT function: <uarray multimethod 'hfftn'>
FAILED scipy/fft/tests/test_basic.py::TestFFT::test_axes_non_standard[jax.numpy-ihfftn] - ValueError: Unknown FFT function: <uarray multimethod 'ihfftn'>
FAILED scipy/fft/tests/test_basic.py::TestFFT::test_axes_subset_with_shape_standard[jax.numpy-fftn] - ValueError: Unknown FFT function: <uarray multimethod 'fftn'>
FAILED scipy/fft/tests/test_basic.py::TestFFT::test_axes_subset_with_shape_standard[jax.numpy-ifftn] - ValueError: Unknown FFT function: <uarray multimethod 'ifftn'>
FAILED scipy/fft/tests/test_basic.py::TestFFT::test_axes_subset_with_shape_standard[jax.numpy-rfftn] - ValueError: Unknown FFT function: <uarray multimethod 'rfftn'>
FAILED scipy/fft/tests/test_basic.py::TestFFT::test_axes_subset_with_shape_standard[jax.numpy-irfftn] - ValueError: Unknown FFT function: <uarray multimethod 'irfftn'>
FAILED scipy/fft/tests/test_basic.py::TestFFT::test_axes_subset_with_shape_non_standard[jax.numpy-fft2] - ValueError: Unknown FFT function: <uarray multimethod 'fft2'>
FAILED scipy/fft/tests/test_basic.py::TestFFT::test_axes_subset_with_shape_non_standard[jax.numpy-ifft2] - ValueError: Unknown FFT function: <uarray multimethod 'ifft2'>
FAILED scipy/fft/tests/test_basic.py::TestFFT::test_axes_subset_with_shape_non_standard[jax.numpy-rfft2] - ValueError: Unknown FFT function: <uarray multimethod 'rfft2'>
FAILED scipy/fft/tests/test_basic.py::TestFFT::test_axes_subset_with_shape_non_standard[jax.numpy-irfft2] - ValueError: Unknown FFT function: <uarray multimethod 'irfft2'>
FAILED scipy/fft/tests/test_basic.py::TestFFT::test_axes_subset_with_shape_non_standard[jax.numpy-hfft2] - ValueError: Unknown FFT function: <uarray multimethod 'hfft2'>
FAILED scipy/fft/tests/test_basic.py::TestFFT::test_axes_subset_with_shape_non_standard[jax.numpy-ihfft2] - ValueError: Unknown FFT function: <uarray multimethod 'ihfft2'>
FAILED scipy/fft/tests/test_basic.py::TestFFT::test_axes_subset_with_shape_non_standard[jax.numpy-hfftn] - ValueError: Unknown FFT function: <uarray multimethod 'hfftn'>
FAILED scipy/fft/tests/test_basic.py::TestFFT::test_axes_subset_with_shape_non_standard[jax.numpy-ihfftn] - ValueError: Unknown FFT function: <uarray multimethod 'ihfftn'>
```
tbh I find the structure of `fft` quite difficult to understand

lucascolley:
> @lucascolley do you understand these failures?

Yes. The reason it fails only for JAX (testing JIT) is due to the monkey-patching that occurs from `lazy_xp_function` via `make_xp_test_case`:

```
scipy on 🎋 xpfft [🔨] is 📦 v1.17.0.dev0 via 🐍 via 🧚 v0.59.0
❯ pixi run test-strict -t scipy.fft.tests.test_basic "--" -k test_axes_non_standard
...
> /Users/lucascolley/ghq/github.com/rgommers/pixi-dev-scipystack/scipy/scipy/build-cpu-install/usr/lib/python3.13/site-packages/scipy/fft/tests/test_basic.py(23)get_expected_input_dtype()
-> breakpoint()
(Pdb) func
<uarray multimethod 'hfftn'>
(Pdb) func is fft.hfftn
True

scipy on 🎋 xpfft [🔨] is 📦 v1.17.0.dev0 via 🐍 via 🧚 v0.59.0
❯ pixi run test-jax -t scipy.fft.tests.test_basic "--" -k test_axes_non_standard
...
> /Users/lucascolley/ghq/github.com/rgommers/pixi-dev-scipystack/scipy/scipy/build-cpu-install/usr/lib/python3.13/site-packages/scipy/fft/tests/test_basic.py(23)get_expected_input_dtype()
-> breakpoint()
(Pdb) func
<uarray multimethod 'hfftn'>
(Pdb) func is fft.hfftn
False
```

I think we just need to make https://github.com/scipy/scipy/blob/02d9a63b04b5a4c7bedd8c95e33dd6a8c20f1854/scipy/fft/tests/test_basic.py#L20-L32 robust to the monkey-patching, e.g. by using the name of the function instead of object identity.

scipy/fft/tests/test_real_transforms.py
```diff
@@ -190,53 +191,50 @@ def test_orthogonalize_noop(func, type, norm, xp):
...
+@make_xp_test_case(dct, dst)
 @pytest.mark.parametrize("norm", ["backward", "ortho", "forward"])
 @pytest.mark.parametrize("func", [dct, dst])
 def test_orthogonalize_dcst2(func, norm, xp):
     x = xp.asarray(np.random.rand(100))
     y1 = func(x, type=2, norm=norm, orthogonalize=True)
     y2 = func(x, type=2, norm=norm, orthogonalize=False)
 
-    y2[0 if func == dct else -1] /= SQRT_2
+    xpx.at(y2, 0 if func == dct else -1).divide(SQRT_2)
```
j-bowhay:
```suggestion
    xpx.at(y2, 0 if func.__name__ == "dct" else -1).divide(SQRT_2)
```

scipy/fft/tests/test_real_transforms.py
```diff
@@ -190,53 +191,50 @@ def test_orthogonalize_noop(func, type, norm, xp):
...
-                  reason='jax arrays do not support item assignment')
+@make_xp_test_case(dct, dst)
 @skip_xp_backends(cpu_only=True)
 @pytest.mark.parametrize("norm", ["backward", "ortho", "forward"])
 @pytest.mark.parametrize("func", [dct, dst])
 def test_orthogonalize_dcst3(func, norm, xp):
     x = xp.asarray(np.random.rand(100))
     x2 = xp_copy(x, xp=xp)
-    x2[0 if func == dct else -1] *= SQRT_2
+    xpx.at(x2, 0 if func == dct else -1).multiply(SQRT_2)
```
j-bowhay:
```suggestion
    xpx.at(x2, 0 if func.__name__ == "dct" else -1).multiply(SQRT_2)
```

j-bowhay:
I'm sure it's probably not too complicated but the final Jax failures are completely illuding me so far

lucascolley:
The tolerance violations look particularly egregious for `dct1`, I suspect a JAX bug

j-bowhay:
> The tolerance violations look particularly egregious for `dct1`, I suspect a JAX bug

In the interests of getting this merged for 1.17 is it fine to make these functions which are failing as not supporting Jax along with a to-do?

steppi:
> > The tolerance violations look particularly egregious for `dct1`, I suspect a JAX bug
> 
> In the interests of getting this merged for 1.17 is it fine to make these functions which are failing as not supporting Jax along with a to-do?

Yes. If something doesn't work and the problem appears nontrivial, then I think it would be beyond the scope of this PR to fix it anyway.

I've started reviewing this and things look good. One thing to look out for is that JAX tests can pass on GPU even if something isn't actually supported because JAX doesn't raise on conversion like the other backends do.  I've found that generally, if something raises with the jit enabled, it probably isn't supported on the GPU either, but that may not always be true. I'll take a look for such cases as I review further and will finish today. As an aside, I got involved after `uarray`, used for dispatch in `fft`, was discarded. I spent a little time looking into it's history, that's an interesting lost byway,on the way to the Array API standard.

lucascolley:
>  that's an interesting lost byway,on the way to the Array API standard

hopefully on the way to gh-23493 too IMO

scipy/fft/_basic.py
```diff
@@ -271,6 +275,7 @@ def ifft(x, n=None, axis=-1, norm=None, overwrite_x=False, workers=None, *,
     return (Dispatchable(x, np.ndarray),)
 
 
+@xp_capabilities()
```
j-bowhay:
```suggestion
@xp_capabilities(allow_dask_compute=1)
```

j-bowhay:
CI is finally green!

steppi (approved):
I'm going to go ahead and merge this, because I think the contents are mostly correct, it is an improvement on the current state of things, and I don't want a repeat of the situation with `signal` where I spent weeks dotting i's and crossing t's. This is going to have to be revisited though. I find the current state of alternative backend support in `fft` deeply confusing.

Take for instance `idct`. It is available in both CuPy and JAX but calling `idct` on a CuPy array fails, while even JAX jit tests for `idct` seem to pass. Nevertheless, I've determined that calls to `idct` on both CuPy and JAX both go through, [scipy.fft._realtransforms_backend.idct](https://github.com/scipy/scipy/blob/277da4104025fc7b53fb9e129698488009362394/scipy/fft/_realtransforms_backend.py#L48-L51), which converts to NumPy. This is where CuPy fails due to a device transfer. JAX succeeds on GPU because we haven't sorted the transfer guard issue yet. More bizarrely, JAX tests for `idct` and other real transforms appear to succeed with the JIT enabled even though the functions are internally converting to NumPy and back. This should not be happening, and I think is a sign that there is something wrong with our assumptions around how the JIT tests should work.

To add to the confusion, `fft` appears to be using `uarray` and CuPy defines `__ua_convert__` and `__ua_function__`, and decorates `idct` and other real transforms with a decorator which should allow these to get picked up by `uarray`s dispatch. But no such dispatch is occurring.  For functions in `fft` where SciPy currently is successful in dispatching to CuPy, this appears to not be due to `uarray` but because of explicit conditional dispatch like below

https://github.com/scipy/scipy/blob/277da4104025fc7b53fb9e129698488009362394/scipy/fft/_basic_backend.py#L27-L49

See https://github.com/cupy/cupy/blob/720192ff7957605fce798120e150fb0c269a9584/cupyx/scipy/fft/_fft.py#L36-L69

The current state of alternative backend support in `scipy.fft`, mixing `uarray` with explicit dispatch, and where some things do and don't work for mysterious reasons, is a bit of a horrifying byzantine labyrinth. In the interest of finishing the addition of `xp_capabilities` for all modules within the next couple of weeks, I will let this go for now and will revisit later.

j-bowhay:
Thanks @steppi, I completely agree with your assessment I found this incredibly confusing to work on so it would great if we could simplify things.

lucascolley:
Dask CI is now failing on main, see e.g. https://productionresultssa15.blob.core.windows.net/actions-results/939c1bba-0ea1-40c9-81f8-7b53ebf7e59f/workflow-job-run-c146b8b5-0667-5168-a5b7-3614a75e47cf/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-12-02T15%3A00%3A49Z&sig=4zTA068kjN0RW1JkXYdI9h2QBO%2FgniBARt41k%2FvHLDk%3D&ske=2025-12-03T01%3A17%3A07Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-12-02T13%3A17%3A07Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2025-12-02T14%3A50%3A44Z&sv=2025-11-05, seems related to this PR, although it looks like it passed here.

Maybe both this and some other change (like bumping array-api-extra?) have jointly caused the failures?

lucascolley:
the failures all seem to be where other modules like `signal` and `ndimage` call `fft` functions.

@steppi perhaps this is related to what we were discussing when I mentioned that it would be nice for `xp_capabilities` to be able to "inherit" from subroutine functions? I'm not sure why it wouldn't have failed on this PR, though.

steppi:
> the failures all seem to be where other modules like `signal` and `ndimage` call `fft` functions.
> 
> @steppi perhaps this is related to what we were discussing when I mentioned that it would be nice for `xp_capabilities` to be able to "inherit" from subroutine functions? I'm not sure why it wouldn't have failed on this PR, though.

The failures are odd, but maybe not that surprising given how complicated and brittle the situation appears to be in `fft`. When reviewing this PR, I actually couldn't understand why these `real_transforms` tests which are now failing with `dask` were passing with `dask` in the first place, but chalked it up to me not being able to understand the infra due to its complexity. For now, we should just add skips. If I understand right, having `xp_capabilities` inherit from dependencies sounds like it may be complicated. You can have different chains of dependencies depending on backend, due to situations like the following:

`f` is not directly available in backend libraries `X` and `Y`.  `f` depends on `g` which depends on `h` and `f` is api agnostic modulo its dependence on `g`.  `g` is array api agnostic modulo dependence on `h` which is not array api agnostic.  `g` is available in library `X` and can be dispatched to directly, but `h` is not available in `X`. `g` is not available in library `Y` but `h` is. The end result is that  `f` can be supported for libraries `X` and `Y`. For library `X`, `f` depends only on `g`, but for library `Y`, `f` depends transitively on `g` and `h`. This isn't just academic, there actually was a situation like this in `signal` if I recall correctly.

I think the root cause here  is the complexity and seemingly brittle nature of the lazy testing setup combined with the opacity that the use of `uarray` brings to this situation. Looking further into it, I really don't think these tests should have been passing in the first place. Perhaps the update to  `array_api_extra` came with a bug fix that made the tests fail like they are supposed to?

I can submit a PR adding the necessary skips.

lucascolley:
> `f` is not directly available in backend libraries `X` and `Y`. `f` depends on `g` which depends on `h` and `f` is api agnostic modulo its dependence on `g`. `g` is array api agnostic modulo dependence on `h` which is not array api agnostic. `g` is available in library `X` and can be dispatched to directly, but `h` is not available in `X`. `g` is not available in library `Y` but `h` is. The end result is that `f` can be supported for libraries `X` and `Y`. For library `X`, `f` depends only on `g`, but for library `Y`, `f` depends transitively on `g` and `h`. This isn't just academic, there actually was a situation like this in `signal` if I recall correctly.

yeah good point, that is a mess we probably don't want to get stuck in

> I can submit a PR adding the necessary skips.

great thanks!

PR #24049, opened 2025-11-22
MAINT: speed up `xp_promote` and `xp_result_type` for big torch tensors

### LINKED ISSUE #23670: BUG: scipy.special.logsumexp seems slower for version 1.16
### Describe your issue.

When running scipy.special.logsumexp on a google colab notebook with the T4 GPU with cuda, the runtimes are slower with scipy versions 1.16.2 and 1.16.2 as compared to scipy version 1.15.3. Here are the runtimes that I noted:
```console
Scipy version 1.16.2: avg time taken for logsumexp = 0.24944710731506348

Scipy version 1.16.1: avg time taken for logsumexp = 0.27803874015808105

Scipy version 1.15.3: avg time taken for logsumexp = 0.02717113494873047
```

This causes a significant slowdown when introducing array API support for a feature which calls this function multiple times. For reference please check https://github.com/scikit-learn/scikit-learn/pull/32246

### Reproducing Code Example

```python
import scipy
import torch as xp
from scipy.special import logsumexp
from time import time


pred = xp.rand((100000, 500), dtype=xp.float64, device="cuda")
timings = []
for _ in range(10):
  start = time()
  ouput = logsumexp(pred, axis=1)
  end = time()
  timings.append(end - start)

avg_time_taken = sum(timings) / 10
print(f"Scipy version {scipy.__version__}: avg time taken for logsumexp = {end - start}")
```

### Error message

```shell
No errors
```

```shell
This issue is in comparison with different versions of scipy on a google colab notebook with the T4 GPU

I checked with the following scipy versions:
scipy 1.16.2
scipy 1.16.1
scipy 1.15.3

numpy version 2.0.2

sys.version_info(major=3, minor=12, micro=11, releaselevel='final', serial=0)
```

mdhaber:
#### Reference issue
Closes gh-23670

In `xp_promote` and `xp_result_type`, we are using `np.iterable` to determine whether objects should be treated as arrays for the sake of dtype assessment. `np.iterable` - and just calling `iter` - is really slow for PyTorch tensors, so short-circuit if the object is a tensor.

#### Additional information
Please commit fixes directly to this branch.

### DIFF
--- a/scipy/_lib/_array_api.py
+++ b/scipy/_lib/_array_api.py
@@ -488,8 +488,8 @@ def xp_result_type(*args, force_floating=False, xp):
     Typically, this function will be called shortly after `array_namespace`
     on a subset of the arguments passed to `array_namespace`.
     """
-    args = [(_asarray(arg, subok=True, xp=xp) if np.iterable(arg) else arg)
-            for arg in args]
+    args = [(_asarray(arg, subok=True, xp=xp) if is_torch_array(arg) or np.iterable(arg)
+            else arg) for arg in args]
     args_not_none = [arg for arg in args if arg is not None]
     if force_floating:
         args_not_none.append(1.0)
@@ -539,8 +539,8 @@ def xp_promote(*args, broadcast=False, force_floating=False, xp):
     if not args:
         return args
 
-    args = [(_asarray(arg, subok=True, xp=xp) if np.iterable(arg) else arg)
-            for arg in args]  # solely to prevent double conversion of iterable to array
+    args = [(_asarray(arg, subok=True, xp=xp) if is_torch_array(arg) or np.iterable(arg)
+            else arg) for arg in args]  # solely to prevent double conversion of iterable to array
 
     dtype = xp_result_type(*args, force_floating=force_floating, xp=xp)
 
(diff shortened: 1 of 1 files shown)


scipy/_lib/_array_api.py
```diff
@@ -539,8 +539,8 @@ def xp_promote(*args, broadcast=False, force_floating=False, xp):
     if not args:
         return args
 
-    args = [(_asarray(arg, subok=True, xp=xp) if np.iterable(arg) else arg)
-            for arg in args]  # solely to prevent double conversion of iterable to array
+    args = [(_asarray(arg, subok=True, xp=xp) if is_torch_array(arg) or np.iterable(arg)
+            else arg) for arg in args]  # solely to prevent double conversion of iterable to array
```
mdhaber:
```suggestion
            else arg) for arg in args]  # prevent double conversion of iterable to array
```

lucascolley (approved):
thanks Matt!

Part of me would prefer special casing for NumPy scalars than PyTorch arrays to achieve the same thing, but not worth splitting hairs.

mdhaber:
I tried other things, and tests started failing because the assumptions that led to my plan above weren't met. At some point, `array_namespace` started letting lists and tuples through, and now we run some tests with lists and tuples as input even when `SCIPY_ARRAY_API=1`. I didn't want to make this a big project, so I did the simple thing. 

Using `iter`/`np.iterable` to define what "array-like" really means could be considered a hack in the first place, but I don't know what the clean solution would be since "array-like" is poorly defined. If we don't feel the need to replace that, I don't think it's wrong to special case for tensors because ultimately they're doing something surprising if calling `iter` on them takes so long.

If we do want to replace that, I think we'd special case on Python scalars, not NumPy scalars. That could work.

PR #22971, opened 2025-05-11
MAINT: fix skellam distribution tests

dschmitz89:
#### Reference issue
Closes #22956

#### What does this implement/fix?
This should help with the recent discrete distribution updates. I do not understand the context fully though. Locally, I still see failures for other distributions.

Long term, it would be better to have a unified wrapping mechanism for boost functions that handle C++ exceptions generally. Let's fight that battle another day though.

PR #24054, opened 2025-11-24
MAINT: special: streamlined & safe wrappers for private boost functions

dschmitz89:
#### Reference issue
We had already two issues where C++ exceptions in Boost could crash the python interpreter ([here](https://github.com/scipy/scipy/pull/22971) and [here](https://github.com/scipy/scipy/pull/24008)). I am quite sure there are more cases that we just have not encountered yet. The underlying reason is simply that not all C++/boost exceptions are caught in the wrapper function.

Adds unified error handling of boost functions that are used in `stats`. Wrappers for `special` need a more elaborate treatment due to `sf_error` in that module. It essentially boils down to:

```c++
// Simple wrapper to avoid C++ exceptions from crashing Python
// Should be used for all functions that do not have special error handling
template<typename Real, typename Callable>
Real error_handler(Callable&& func) {
    try {
        return func();
    } catch (const std::underflow_error&) {
        /* Boost underflow - return 0 */
        return Real(0);
    } catch (const std::overflow_error&) {
        /* Boost overflow - return infinity */
        return std::numeric_limits<Real>::infinity();
    } catch (...) {
        /* Boost was unable to produce a result. */
        return std::numeric_limits<Real>::quiet_NaN();
    }
}
``` 

Additional changes to be more idiomatic regarding C++ templates:
* replace `INFINITY` with `std::numeric_limits<Real>::infinity()`
* replace `NAN` with `std::numeric_limits<Real>::quiet_NaN()`
* replace duplicated wrappers for single and double precision with a templated one where the error handling is applied to avoid code duplication

### DIFF
--- a/scipy/special/boost_special_functions.h
+++ b/scipy/special/boost_special_functions.h
@@ -66,6 +66,24 @@ boost::math::policies::user_overflow_error(const char* function, const char* mes
     return 0;
 }
 
+// Simple wrapper to avoid C++ exceptions from crashing Python
+// Should be used for all functions that do not have special error handling
+template<typename Real, typename Callable>
+Real error_handler(Callable&& func) {
+    try {
+        return func();
+    } catch (const std::underflow_error&) {
+        /* Boost underflow - return 0 */
+        return Real(0);
+    } catch (const std::overflow_error&) {
+        /* Boost overflow - return infinity */
+        return std::numeric_limits<Real>::infinity();
+    } catch (...) {
+        /* Boost was unable to produce a result. */
+        return std::numeric_limits<Real>::quiet_NaN();
+    }
+}
+
 template<typename Real>
 static inline
 Real ibeta_wrap(Real a, Real b, Real x)
@@ -73,12 +91,12 @@ Real ibeta_wrap(Real a, Real b, Real x)
     Real y;
 
     if (std::isnan(a) || std::isnan(b) || std::isnan(x)) {
-        return NAN;
+        return std::numeric_limits<Real>::quiet_NaN();
     }
 
     if ((a < 0) || (b < 0) || (x < 0) || (x > 1)) {
         sf_error("betainc", SF_ERROR_DOMAIN, NULL);
-        return NAN;
+        return std::numeric_limits<Real>::quiet_NaN();
     }
 
     /* In limiting cases, SciPy treats `betainc` as a two parameter family
@@ -95,7 +113,7 @@ Real ibeta_wrap(Real a, Real b, Real x)
 	 * Bernoulli(t / (t + 1)). The a = 0, b = 0 case is thus indeterminate.
 	 * A similar statement can be made for the limit (a, b) -> (inf, inf).
 	 */
-        return NAN;
+        return std::numeric_limits<Real>::quiet_NaN();
     }
 
     if ((a == 0) || std::isinf(b)) {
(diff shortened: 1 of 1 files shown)


scipy/special/boost_special_functions.h
```diff
@@ -66,19 +66,37 @@ boost::math::policies::user_overflow_error(const char* function, const char* mes
...
+template<typename Real, typename Callable>
+Real error_handler(Callable&& func) {
+    try {
+        return func();
+    } catch (const std::underflow_error&) {
+        /* Boost underflow - return 0 */
+        return Real(0);
+    } catch (const std::overflow_error&) {
+        /* Boost overflow - return infinity */
+        return std::numeric_limits<Real>::infinity();
```
ev-br, line 80: How does boost signal overflows to a negative infinity?

ev-br (commented):
Could you say a few words why we need/want a handler different from `sf_error`?
Naively it feels like we should be moving towards having the same handlers across scipy, so what's needed is a connector between whatever boost is doing and `sf_error`?
(I'm sure it's been discussed already, am happy to be pointed to a previous discussion)

scipy/special/boost_special_functions.h
dschmitz89: Good point, will investigate.

dschmitz89:
> Could you say a few words why we need/want a handler different from `sf_error`? Naively it feels like we should be moving towards having the same handlers across scipy, so what's needed is a connector between whatever boost is doing and `sf_error`? (I'm sure it's been discussed already, am happy to be pointed to a previous discussion)

AFAIK, there has been no previous discussion. My understanding is that `sf_error` is only applicable for public functions in `special`. For use in `stats`, we wrap a LOT of boost functions that are not part of the public special API though. Even if we made those wrappers aware of `sf_error`, it would not be obvious to users that it can be used here. Also, I am pretty sure that the vast majority of users (especially stats, not special users) is not even aware of `sf_error` and how to change the error handling.

That's why I opted for the simple error handler here. Its main purpose it to avoid crashes, not to give the most sophisticated error handling possibilities. Currently, these functions do not have any error handling at all. For use in `special`, we can create another error handler or extend this one to include `sf_error`.

dschmitz89:
It essentially boils down to:
> 
> ```c++
> // Simple wrapper to avoid C++ exceptions from crashing Python
> // Should be used for all functions that do not have special error handling
> template<typename Real, typename Callable>
> Real error_handler(Callable&& func) {
>     try {
>         return func();
>     } catch (const std::underflow_error&) {
>         /* Boost underflow - return 0 */
>         return Real(0);
>     } catch (const std::overflow_error&) {
>         /* Boost overflow - return infinity */
>         return std::numeric_limits<Real>::infinity();
>     } catch (...) {
>         /* Boost was unable to produce a result. */
>         return std::numeric_limits<Real>::quiet_NaN();
>     }
> }
> ``` 

@mborland @jzmaddock Would you have time to comment on this? I am trying to establish an error handler to catch all possible exceptions from boost to avoid crashing python (we should always return at least `NaN`). Questions from our side are for example: 
* Is it possible to catch negative overflow?
* Should other exceptions be caught?

jzmaddock:
There's also:

`std::domain_error`:  error in input range.
`boost::math::evaluation_error`: don't know how to evaluate.
`boost::math::rounding_error`: result of a rounding operation (to integer) does not fit in the target type (generally an internal logic error).

But these will be caught by the `catch(...)` clause.  They are all derived from `std::runtime_error`.

And no, you can't tell the sign of an overflow from an `overflow_error`.  Generally, as long as you have a "do not throw" policy in place you should see none of these exceptions, and if you do it's a bug!

mborland:
Like John said with the correct policy you won't get any of them. In the past I believe you guys used ([also here](https://github.com/boostorg/math/blob/develop/test/scipy_issue_15101.cpp)):

```
#define BOOST_MATH_DOMAIN_ERROR_POLICY ignore_error
#define BOOST_MATH_OVERFLOW_ERROR_POLICY ignore_error
#define BOOST_MATH_PROMOTE_DOUBLE_POLICY false
```

dschmitz89:
Thanks @jzmaddock @mborland . The culprit seems to be our policy which includes special handling of `evaluation_error`: see [here](https://github.com/scipy/scipy/blob/main/scipy/special/boost_special_functions.h#L23) and [here](https://github.com/scipy/scipy/blob/d292d3221bb7f2172a67fbe377083edd81e4abfe/scipy/special/boost_special_functions.h#L36). Calling the binomial CDF with this policy alone was not enough to ignore errors, that's why I added some more try/catch in https://github.com/scipy/scipy/pull/24008 .

PR #24055, opened 2025-11-24
ENH: sparse.linalg.spsolve: reduce memory footprint and increase speed

broesler:
This PR makes a substantial improvement to the peak memory usage and a reduction in run-time of `sparse.linalg.spsolve` when solving systems with many sparse right-hand sides. The original implementation looped over each column of the input, converted it to a dense column, and computed the full coordinates of the output with calls to `np.full`. The new implementation uses the `sparse.csc_array` format to avoid those full array creations.

This PR also introduces the `batch_size` kwarg that can be used to control the number of columns solved at one time. The `SuperLU` solver can handle dense matrix inputs natively at the C level, so instead of using a python loop over each column of the sparse RHS matrix input to `spsolve`, we can convert, *e.g.*, 100 columns at a time to dense and pass them to `SuperLU`. The input `batch_size=1` retains the original `spsolve` behavior. The `batch_size` argument does not apply to `use_umfpack=True` cases, because the low-level `scikit-umfpack` routines used in `spsolve` cannot handle matrix inputs.

## Additional information
[Benchmark results](https://broesler.github.io/scipy/#summarylist?sort=0&dir=asc) show a 35-50% reduction in peak memory use, with no change in run time for `batch_size=1`.

#### Peak memory use vs commit, `batch_size=1`
image

#### Runtime vs commit, `batch_size=1`
image

Note that the benchmarks across commits are only valid for `batch_size=1`, since that option was introduced in the latest commit.

### Default Batch Size
I chose the default batch size to optimize runtime and memory use. Plots of peak memory use and timing shows minimums for `batch_size=10` on my local machine. Peak memory use is not significantly changed vs. `batch_size=1`, but the benchmark shows a 50% reduction in runtime by using `batch_size=10`.

#### Peak memory use vs batch_size
image

#### Runtime vs commit vs batch_size
image

Memory and Runtime tables filtered to show only the `use_umfpack=False` results (SuperLU):

#### Peak Memory Usage (`peakmem_solve`)
*Lower is better. `batch_size=10` uses the least memory.*

| Nsq | K | Density | Batch=1 (Baseline) | Batch=10 | Batch=100 | Batch=5000 (Full) |
|---:|---:|---:|---:|---:|---:|---:|
| 5000 | 2000 | 0.01 | 331M | **318M** | 385M | 550M |
| 5000 | 2000 | 0.1 | 358M | **349M** | 413M | 606M |
| 5000 | 2000 | 0.5 | 501M | **487M** | 541M | 641M |

#### Execution Time (`time_solve`)
*Lower is better. `batch_size=10` is consistently the fastest.*

| Nsq | K | Density | Batch=1 (Baseline) | Batch=10 | Batch=100 | Batch=5000 (Full) |
|---:|---:|---:|---:|---:|---:|---:|
| 5000 | 2000 | 0.01 | 810±40ms | **391±3ms** | 441±8ms | 498±8ms |
| 5000 | 2000 | 0.1 | 843±40ms | **395±20ms** | 437±10ms | 491±4ms |
| 5000 | 2000 | 0.5 | 864±60ms | **423±30ms** | 458±5ms | 530±20ms |

### DIFF
--- a/scipy/sparse/linalg/_dsolve/linsolve.py
+++ b/scipy/sparse/linalg/_dsolve/linsolve.py
@@ -1,5 +1,6 @@
 from warnings import warn, catch_warnings, simplefilter
 
+from numbers import Integral
 import numpy as np
 from numpy import asarray
 from scipy.sparse import (issparse, SparseEfficiencyWarning,
@@ -257,6 +258,9 @@ def spsolve(A, b, permc_spec=None, use_umfpack=True, rhs_batch_size=10):
         useUmfpack.u = not noScikit
 
     use_umfpack = use_umfpack and useUmfpack.u
+
+    if not isinstance(rhs_batch_size, Integral) or rhs_batch_size <= 0:
+        raise ValueError("rhs_batch_size must be a positive integer")
 
     if b_is_vector and use_umfpack:
         if b_is_sparse:
--- a/scipy/sparse/linalg/_dsolve/linsolve.py
+++ b/scipy/sparse/linalg/_dsolve/linsolve.py
@@ -163,6 +163,8 @@ def spsolve(A, b, permc_spec=None, use_umfpack=True, rhs_batch_size=10):
         ``use_umfpack=False``, since the low-level scikit-umfpack routines do
         not support multiple right-hand sides. In that case, ``rhs_batch_size=1``.
 
+        .. versionadded:: 1.18.0
+
     Returns
     -------
     x : ndarray or sparse array or matrix
--- a/scipy/sparse/linalg/_dsolve/linsolve.py
+++ b/scipy/sparse/linalg/_dsolve/linsolve.py
@@ -132,7 +132,7 @@ def _get_umf_family(A):
 
     return family, A_new
 
-def spsolve(A, b, permc_spec=None, use_umfpack=True, rhs_batch_size=10):
+def spsolve(A, b, permc_spec=None, use_umfpack=True, *, rhs_batch_size=10):
     """Solve the sparse linear system Ax=b, where b may be a vector or a matrix.
 
     Parameters
--- a/scipy/sparse/linalg/_dsolve/linsolve.py
+++ b/scipy/sparse/linalg/_dsolve/linsolve.py
@@ -1,6 +1,5 @@
 from warnings import warn, catch_warnings, simplefilter
 
-from numbers import Integral
 import numpy as np
 from numpy import asarray
 from scipy.sparse import (issparse, SparseEfficiencyWarning,
@@ -8,6 +7,7 @@ from scipy.sparse import (issparse, SparseEfficiencyWarning,
 from scipy.sparse._sputils import (is_pydata_spmatrix, convert_pydata_sparse_to_scipy,
                                    safely_cast_index_arrays)
 from scipy.linalg import LinAlgError
+from scipy._lib._util import _validate_int
 import copy
 import threading
 
@@ -261,8 +261,7 @@ def spsolve(A, b, permc_spec=None, use_umfpack=True, *, rhs_batch_size=10):
 
     use_umfpack = use_umfpack and useUmfpack.u
 
-    if not isinstance(rhs_batch_size, Integral) or rhs_batch_size <= 0:
-        raise ValueError("rhs_batch_size must be a positive integer")
+    _validate_int(rhs_batch_size, "rhs_batch_size", minimum=1)
 
     if b_is_vector and use_umfpack:
         if b_is_sparse:
--- a/scipy/sparse/linalg/_dsolve/linsolve.py
+++ b/scipy/sparse/linalg/_dsolve/linsolve.py
@@ -131,7 +131,7 @@ def _get_umf_family(A):
 
     return family, A_new
 
-def spsolve(A, b, permc_spec=None, use_umfpack=True, batch_size=10):
+def spsolve(A, b, permc_spec=None, use_umfpack=True, rhs_batch_size=10):
     """Solve the sparse linear system Ax=b, where b may be a vector or a matrix.
 
     Parameters
@@ -154,13 +154,13 @@ def spsolve(A, b, permc_spec=None, use_umfpack=True, batch_size=10):
         if True (default) then use UMFPACK for the solution [3]_, [4]_, [5]_,
         [6]_ . This is only referenced if b is a vector and
         ``scikits.umfpack`` is installed.
-    batch_size : int, optional
+    rhs_batch_size : int, optional
         If ``b`` is a 2D sparse array, this parameter controls the number of
         columns to be solved simultaneously. A larger number will increase
         memory consumption by converting more columns at a time to dense
         arrays, but may improve runtime. This option only applies when
         ``use_umfpack=False``, since the low-level scikit-umfpack routines do
-        not support multiple right-hand sides. In that case, ``batch_size=1``.
+        not support multiple right-hand sides. In that case, ``rhs_batch_size=1``.
 
     Returns
     -------
@@ -315,18 +315,18 @@ def spsolve(A, b, permc_spec=None, use_umfpack=True, batch_size=10):
                      SparseEfficiencyWarning, stacklevel=2)
                 b = csc_array(b)
 
-            if use_umfpack and batch_size > 1:
-                batch_size = 1
+            if use_umfpack and rhs_batch_size > 1:
+                rhs_batch_size = 1
 
             # Solve in batches to reduce memory consumption
             K = b.shape[1]
             x_blocks = []
 
             # Pre-allocate arrays to avoid repeated allocations
-            b_batch = np.empty((N, min(batch_size, K)), dtype=b.dtype, order="F")
+            b_batch = np.empty((N, min(rhs_batch_size, K)), dtype=b.dtype, order="F")
 
-            for k in range(0, K, batch_size):
-                batch_end = min(k + batch_size, K)
+            for k in range(0, K, rhs_batch_size):
+                batch_end = min(k + rhs_batch_size, K)
                 width = batch_end - k
                 # Convert sparse to dense in the buffer
                 b_view = b_batch[:, :width]
--- a/scipy/sparse/linalg/_dsolve/linsolve.py
+++ b/scipy/sparse/linalg/_dsolve/linsolve.py
@@ -132,7 +132,7 @@ def _get_umf_family(A):
 
     return family, A_new
 
-def spsolve(A, b, permc_spec=None, use_umfpack=True, *, rhs_batch_size=10):
+def spsolve(A, b, permc_spec=None, use_umfpack=True, *, rhs_block_size=10):
     """Solve the sparse linear system Ax=b, where b may be a vector or a matrix.
 
     Parameters
@@ -155,13 +155,13 @@ def spsolve(A, b, permc_spec=None, use_umfpack=True, *, rhs_batch_size=10):
         if True (default) then use UMFPACK for the solution [3]_, [4]_, [5]_,
         [6]_ . This is only referenced if b is a vector and
         ``scikits.umfpack`` is installed.
-    rhs_batch_size : int, optional
+    rhs_block_size : int, optional
         If ``b`` is a 2D sparse array, this parameter controls the number of
         columns to be solved simultaneously. A larger number will increase
         memory consumption by converting more columns at a time to dense
         arrays, but may improve runtime. This option only applies when
         ``use_umfpack=False``, since the low-level scikit-umfpack routines do
-        not support multiple right-hand sides. In that case, ``rhs_batch_size=1``.
+        not support multiple right-hand sides. In that case, ``rhs_block_size=1``.
 
         .. versionadded:: 1.18.0
 
@@ -228,6 +228,11 @@ def spsolve(A, b, permc_spec=None, use_umfpack=True, *, rhs_batch_size=10):
     A = convert_pydata_sparse_to_scipy(A)
     b = convert_pydata_sparse_to_scipy(b)
 
+    # TODO generalize spsolve to use batched arrays, like:
+    # A.shape == (2, 3, N, N)
+    # b.shape == (2, 3, N, K) =>
+    # x.shape == (2, 3, N, K)
+
     if not (issparse(A) and A.format in ("csc", "csr")):
         A = csc_array(A)
         warn('spsolve requires A be CSC or CSR matrix format',
@@ -261,7 +266,7 @@ def spsolve(A, b, permc_spec=None, use_umfpack=True, *, rhs_batch_size=10):
 
     use_umfpack = use_umfpack and useUmfpack.u
 
-    _validate_int(rhs_batch_size, "rhs_batch_size", minimum=1)
+    _validate_int(rhs_block_size, "rhs_block_size", minimum=1)
 
     if b_is_vector and use_umfpack:
         if b_is_sparse:
@@ -320,22 +325,22 @@ def spsolve(A, b, permc_spec=None, use_umfpack=True, *, rhs_batch_size=10):
                      SparseEfficiencyWarning, stacklevel=2)
                 b = csc_array(b)
 
-            if use_umfpack and rhs_batch_size > 1:
-                rhs_batch_size = 1
+            if use_umfpack and rhs_block_size > 1:
+                rhs_block_size = 1
 
-            # Solve in batches to reduce memory consumption
+            # Solve in blocks to reduce memory consumption
             K = b.shape[1]
             x_blocks = []
 
             # Pre-allocate arrays to avoid repeated allocations
-            b_batch = np.empty((N, min(rhs_batch_size, K)), dtype=b.dtype, order="F")
+            b_block = np.empty((N, min(rhs_block_size, K)), dtype=b.dtype, order="F")
 
-            for k in range(0, K, rhs_batch_size):
-                batch_end = min(k + rhs_batch_size, K)
-                width = batch_end - k
+            for k in range(0, K, rhs_block_size):
+                block_end = min(k + rhs_block_size, K)
+                width = block_end - k
                 # Convert sparse to dense in the buffer
-                b_view = b_batch[:, :width]
-                b[:, k:batch_end].toarray(out=b_view)
+                b_view = b_block[:, :width]
+                b[:, k:block_end].toarray(out=b_view)
                 # Solve the linear systems
                 x_dense = Afactsolve(b_view)
                 x_blocks.append(csc_array(x_dense, dtype=b.dtype))
--- a/scipy/sparse/linalg/_dsolve/linsolve.py
+++ b/scipy/sparse/linalg/_dsolve/linsolve.py
@@ -131,7 +131,7 @@ def _get_umf_family(A):
 
     return family, A_new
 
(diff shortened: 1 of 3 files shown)


lucascolley (commented):
Cool work! Just a quick comment on the parameter name: perhaps it should be specific to the RHS in some way? In case there are ever future developments which allowed you to pass multiple LHS matrices too. E.g. I'm working on that for the iterative solvers now. 

cc @dschmitz89 also

dschmitz89:
LGTM but I cannot comment on the details as I have never done batched linear solves.

broesler:
I will make the kwarg name change and have one other tweak to pre-allocate memory before the loop that I will test before pushing.

broesler:
The pre-allocation of the `b_batch` array marginally improves peak memory and timing:

### Peak Memory
| Nsq | K    | Density | Batch | Before [0b1b25dd] | After [452ef954] | Ratio |
|-----|------|---------|--------|--------|--------|--------|
| 500 | 5003 | 0.01 | 1    | 127M | 127M | 1.00 |
| 500 | 5003 | 0.01 | 10   | 127M | 127M | 1.00 |
| 500 | 5003 | 0.01 | 100  | 131M | 128M | 0.98 |
| 500 | 5003 | 0.01 | 6000 | 212M | 211M | 0.99 |
| 500 | 5003 | 0.10 | 1    | 138M | 137M | 1.00 |
| 500 | 5003 | 0.10 | 10   | 132M | 132M | 1.00 |
| 500 | 5003 | 0.10 | 100  | 133M | 133M | 0.99 |
| 500 | 5003 | 0.10 | 6000 | 214M | 213M | 1.00 |
| 500 | 5003 | 0.50 | 1    | 160M | 160M | 1.00 |
| 500 | 5003 | 0.50 | 10   | 154M | 154M | 1.00 |
| 500 | 5003 | 0.50 | 100  | 159M | 154M | 0.97 |
| 500 | 5003 | 0.50 | 6000 | 240M | 235M | 0.98 |


### Timing
| Nsq | K    | Density | Batch | Before  [0b1b25dd] | After [452ef954] | Ratio |
|-----|------|---------|--------|---------------|---------------|--------|
| 500 | 5003 | 0.01 | 1    | 404 ± 1 ms     | 414 ± 0.8 ms   | 1.02 |
| 500 | 5003 | 0.01 | 10   | 97.1 ± 0.5 ms  | 89.4 ± 0.5 ms  | 0.92 |
| 500 | 5003 | 0.01 | 100  | 81.2 ± 3 ms    | 81.8 ± 4 ms    | 1.01 |
| 500 | 5003 | 0.01 | 6000 | 77.6 ± 1 ms    | 75.5 ± 2 ms    | 0.97 |
| 500 | 5003 | 0.10 | 1    | 406 ± 0.7 ms   | 416 ± 0.4 ms   | 1.02 |
| 500 | 5003 | 0.10 | 10   | 96.4 ± 0.5 ms  | 88.6 ± 0.5 ms  | 0.92 |
| 500 | 5003 | 0.10 | 100  | 83.5 ± 3 ms    | 82.5 ± 4 ms    | 0.99 |
| 500 | 5003 | 0.10 | 6000 | 77.6 ± 2 ms    | 75.7 ± 2 ms    | 0.98 |
| 500 | 5003 | 0.50 | 1    | 411 ± 0.6 ms   | 419 ± 0.7 ms   | 1.02 |
| 500 | 5003 | 0.50 | 10   | 98.5 ± 0.4 ms  | 89.9 ± 0.2 ms  | 0.91 |
| 500 | 5003 | 0.50 | 100  | 86.2 ± 4 ms    | 84.5 ± 4 ms    | 0.98 |
| 500 | 5003 | 0.50 | 6000 | 80.2 ± 1 ms    | 76.1 ± 1 ms    | 0.95 |

I tweaked the benchmark to have a greater ratio of number of RHS to number of rows. The original results above still hold.

It turns out the bottleneck is actually the creation of a temporary `csc_array` for the slice, before converting to dense (whether or not in a pre-allocated buffer). It takes ~2-3x longer to create the slice than to convert the slice to dense (using `%timeit` tests). It would be neat if `csc_array` had a `.dense_slice()` method that avoided the intermediate array object, but that is for another PR.

broesler:
I just rebased onto main and it fixed the remaining failing checks. Let me know if there is anything else that needs to be done before merging.

broesler:
Just checking in on this PR. Is there anything else that needs to be done before merging?

scipy/sparse/linalg/_dsolve/linsolve.py
```diff
@@ -131,7 +132,7 @@ def _get_umf_family(A):
 
     return family, A_new
 
-def spsolve(A, b, permc_spec=None, use_umfpack=True):
+def spsolve(A, b, permc_spec=None, use_umfpack=True, rhs_batch_size=10):
```
j-bowhay:
```suggestion
def spsolve(A, b, permc_spec=None, use_umfpack=True, *, rhs_batch_size=10):
```

scipy/sparse/linalg/_dsolve/linsolve.py
```diff
@@ -251,6 +261,9 @@ def spsolve(A, b, permc_spec=None, use_umfpack=True):
 
     use_umfpack = use_umfpack and useUmfpack.u
 
+    if not isinstance(rhs_batch_size, Integral) or rhs_batch_size <= 0:
```
j-bowhay: https://github.com/scipy/scipy/blob/b5d412d4e3e3edb210807fef2740dd3e6920806e/scipy/_lib/_util.py#L478

scipy/sparse/linalg/_dsolve/tests/test_linsolve.py
```diff
@@ -449,6 +451,29 @@ def test_bug_8278(self):
...
+        rng = np.random.default_rng(56)
+        N = 5
+        A = eye_array(N, dtype=float, format='csc')
+        s = rng.random(N)
+        x_true = np.outer(s, 1 + np.arange(K))  # (N, K)
+        b = A @ x_true
+        x = spsolve(A, b, rhs_batch_size=rhs_batch_size, use_umfpack=False)
+        if K == 1:
+            x_true = x_true.ravel()
+        assert_array_almost_equal(x, x_true)
```
j-bowhay: Please can new tests use `assert_allclose` instead

scipy/sparse/linalg/_dsolve/linsolve.py
broesler: Thanks! I didn't realize `_validate_int` existed.

scipy/sparse/linalg/_dsolve/linsolve.py
```diff
@@ -300,25 +312,30 @@ def spsolve(A, b, permc_spec=None, use_umfpack=True):
...
-                col_segs.append(np.full(segment_length, j, dtype=int))
-                data_segs.append(np.asarray(xj[w], dtype=A.dtype))
-            sparse_data = np.concatenate(data_segs)
-            idx_dtype = get_index_dtype(maxval=max(b.shape))
-            sparse_row = np.concatenate(row_segs, dtype=idx_dtype)
-            sparse_col = np.concatenate(col_segs, dtype=idx_dtype)
-            x = A.__class__((sparse_data, (sparse_row, sparse_col)),
-                           shape=b.shape, dtype=A.dtype)
+            if use_umfpack and rhs_batch_size > 1:
+                rhs_batch_size = 1
```
lucascolley: any reason for not just raising a `ValueError` in this case?

scipy/sparse/linalg/_dsolve/linsolve.py
```diff
@@ -154,6 +155,15 @@ def spsolve(A, b, permc_spec=None, use_umfpack=True):
         if True (default) then use UMFPACK for the solution [3]_, [4]_, [5]_,
         [6]_ . This is only referenced if b is a vector and
         ``scikits.umfpack`` is installed.
+    rhs_batch_size : int, optional
```
lucascolley: could we add a batched solve under `Examples`?

lucascolley: is it not possible/desirable to generalise this to multi-dimensional batch shapes?

lucascolley (commented):
thanks @broesler, a few cursory comments.

By the way, would you happen to have any interest in batched iterative solvers? If so, you may be interested in reading https://discuss.scientific-python.org/t/batching-linearoperator-status-and-benchmarks/2199.

scipy/sparse/linalg/_dsolve/linsolve.py
broesler: I didn't raise so that the default value of `rhs_batch_size=10` wouldn't break a bunch of code and make users put `rhs_batch_size=1` everywhere if they install `scikit-umfpack`. We could issue a warning so that users are aware of the limitation in that case.

As an aside, I have a (draft) PR for integrating v0.5.0 of `scikit-sparse`, which supports passing a 2D RHS to UMFPACK, and obviates the need for this check in `spsolve`.

scipy/sparse/linalg/_dsolve/linsolve.py
broesler: It is certainly possible to generalize to multi-dimensional batch shapes. I had started with 0D/1D vector -> 2D matrix as the most typical and easiest fix, but I will take a look at your discussion forum post and see how we can incorporate that into the direct solver.

dschmitz89:
> The pre-allocation of the `b_batch` array marginally improves peak memory and timing:
> 
> ### Peak Memory
> Nsq	K	Density	Batch	Before [[0b1b25d](https://github.com/scipy/scipy/commit/0b1b25dd58d0cc46a125ef48a0294466b587a26e)]	After [[452ef95](https://github.com/scipy/scipy/commit/452ef9547afa0d2617376f5addf021cd8c9117fe)]	Ratio
> 500	5003	0.01	1	127M	127M	1.00
> 500	5003	0.01	10	127M	127M	1.00
> 500	5003	0.01	100	131M	128M	0.98
> 500	5003	0.01	6000	212M	211M	0.99
> 500	5003	0.10	1	138M	137M	1.00
> 500	5003	0.10	10	132M	132M	1.00
> 500	5003	0.10	100	133M	133M	0.99
> 500	5003	0.10	6000	214M	213M	1.00
> 500	5003	0.50	1	160M	160M	1.00
> 500	5003	0.50	10	154M	154M	1.00
> 500	5003	0.50	100	159M	154M	0.97
> 500	5003	0.50	6000	240M	235M	0.98
> ### Timing
> Nsq	K	Density	Batch	Before [[0b1b25d](https://github.com/scipy/scipy/commit/0b1b25dd58d0cc46a125ef48a0294466b587a26e)]	After [[452ef95](https://github.com/scipy/scipy/commit/452ef9547afa0d2617376f5addf021cd8c9117fe)]	Ratio
> 500	5003	0.01	1	404 ± 1 ms	414 ± 0.8 ms	1.02
> 500	5003	0.01	10	97.1 ± 0.5 ms	89.4 ± 0.5 ms	0.92
> 500	5003	0.01	100	81.2 ± 3 ms	81.8 ± 4 ms	1.01
> 500	5003	0.01	6000	77.6 ± 1 ms	75.5 ± 2 ms	0.97
> 500	5003	0.10	1	406 ± 0.7 ms	416 ± 0.4 ms	1.02
> 500	5003	0.10	10	96.4 ± 0.5 ms	88.6 ± 0.5 ms	0.92
> 500	5003	0.10	100	83.5 ± 3 ms	82.5 ± 4 ms	0.99
> 500	5003	0.10	6000	77.6 ± 2 ms	75.7 ± 2 ms	0.98
> 500	5003	0.50	1	411 ± 0.6 ms	419 ± 0.7 ms	1.02
> 500	5003	0.50	10	98.5 ± 0.4 ms	89.9 ± 0.2 ms	0.91
> 500	5003	0.50	100	86.2 ± 4 ms	84.5 ± 4 ms	0.98
> 500	5003	0.50	6000	80.2 ± 1 ms	76.1 ± 1 ms	0.95
> I tweaked the benchmark to have a greater ratio of number of RHS to number of rows. The original results above still hold.
> 
> It turns out the bottleneck is actually the creation of a temporary `csc_array` for the slice, before converting to dense (whether or not in a pre-allocated buffer). It takes ~2-3x longer to create the slice than to convert the slice to dense (using `%timeit` tests). It would be neat if `csc_array` had a `.dense_slice()` method that avoided the intermediate array object, but that is for another PR.

The benchmarks do not show a clear performance improvement here. Do we need more benchmarks for the case of multiple right sides or batching?

broesler:
> Cool work! Just a quick comment on the parameter name: perhaps it should be specific to the RHS in some way? In case there are ever future developments which allowed you to pass multiple LHS matrices too. E.g. I'm working on that for the iterative solvers now.
> 
> cc @dschmitz89 also

As I read more into the [Batched Linear Solvers](https://docs.scipy.org/doc/scipy/tutorial/linalg_batch.html#linalg-batch), I realized `rhs_batch_size` isn't really the right name to use, since the "core" dimension should be considered 2, and the "batch" dimension is anything prepended to that (see the last 2 paragraphs of the article).

In this PR, the `rhs_batch_size` parameter just refers to the number of columns of the RHS that are converted to a dense matrix at a time, not the "batch" dimension as used in the rest of Numpy/Scipy. I will push a change to `rhs_block_size` instead.

broesler:
> 
> The benchmarks do not show a clear performance improvement here. Do we need more benchmarks for the case of multiple right sides or batching?

Just to show a quick experiment:

```python
[50]>>> b
[51]===
<Compressed Sparse Column sparse array of dtype 'float64'
	with 24215 stored elements and shape (484, 5003)>
[51]>>> k = 3  # arbitrary
[52]>>> rhs_block_size = 100  # arbitrary
[53]>>> N
[54]=== 484
[54]>>> K
[55]=== 5003
[55]>>> block_end = min(k + rhs_block_size, K)
[56]>>> width = block_end - k
[57]>>> b_view = b_block[:, :width]  # get a slice view
[58]>>> %timeit b[:, k:block_end]  # slice the CSC array
64.9 μs ± 2.21 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
[59]>>> %timeit bd = b[:, k:block_end].toarray()  # slice to new array
87.6 μs ± 8.56 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
[60]>>> %timeit b[:, k:block_end].toarray(out=b_view)  # slice to pre-alloc array
129 μs ± 58.2 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
[61]>>> bs = b[:, k:block_end]  # store slice of b
[62]>>> %timeit ba = bs.toarray()  # convert to new array
13.9 μs ± 783 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
[63]>>> %timeit bs.toarray(out=b_view)  # convert to dense in pre-allocated array
30.4 μs ± 3.39 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
[64]>>> bs
[65]===
<Compressed Sparse Column sparse array of dtype 'float64'
	with 478 stored elements and shape (484, 100)>
[65]>>> b_block = np.empty((N, min(rhs_block_size, K)), dtype=b.dtype, order="C")
[66]>>> b_view = b_block[:, :width]  # get a slice view
[67]>>> %timeit bs.toarray(out=b_view)  # convert to dense in pre-allocated array
117 μs ± 16.2 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

Slicing a `csc_array` is 2-4x slower than converting that slice to a dense array.
Converting the sliced array into a pre-allocated array is 2x slower than converting to a *new* dense array. I am not sure why that would be the case, since an allocation shouldn't have to happen in the former. I also tried `b_block` in "C" order, but that is even slower than in "F" order (as expected).

I am fine with reverting back to the code that just creates a new dense array on each loop instead of using the pre-allocated block, but am curious if anyone has an idea as to why that is so much faster than using a pre-allocated array?

broesler:
The difference gets a bit closer as the density of `b` increases, but the same holds that converting a `csc_array` to a newly-allocated dense array is faster than using a pre-allocated block. Slicing the array is still 2-3x slower than the conversion to dense.

```python
[86]>>> density = 0.5
[87]>>> b = sparse.random_array((N, K), density=density, format="csc", dtype=float)
[88]>>> b_block = np.empty((N, min(rhs_block_size, K)), dtype=b.dtype, order="F")
[89]>>> b_view = b_block[:, :width]  # get a slice view
[90]>>> bs = b[:, k:block_end]  # slice the CSC array
[91]>>> %timeit bs = b[:, k:block_end]  # slice the CSC array
200 μs ± 9.61 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
[92]>>> %timeit bd = bs.toarray()  # convert to dense in new array
70.9 μs ± 2.92 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
[93]>>> %timeit bs.toarray(out=b_view)  # convert to dense in pre-allocated array
86.8 μs ± 2.68 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

broesler:
> am curious if anyone has an idea as to why that is so much faster than using a pre-allocated array?

Did some digging, and [this line](https://github.com/scipy/scipy/blob/b2b7944064af15bd87e781216c2dfda7400ecd2d/scipy/sparse/_base.py#L1621) is the culprit:

```python
def _process_toarray_args(self, order, out):
    if out is not None:
        if order is not None:
            raise ValueError('order cannot be specified if out '
                             'is not None')
        if out.shape != self.shape or out.dtype != self.dtype:
            raise ValueError('out array must be same dtype and shape as '
                             'sparse array')
        out[...] = 0.
        return out
    else:
        return np.zeros(self.shape, dtype=self.dtype, order=order)
```

The assignment of every element to 0.0 is slower than just allocating a new array.

lucascolley:
> > am curious if anyone has an idea as to why that is so much faster than using a pre-allocated array?
> 
> Did some digging, and [this line](https://github.com/scipy/scipy/blob/b2b7944064af15bd87e781216c2dfda7400ecd2d/scipy/sparse/_base.py#L1621) is the culprit:
> 
> ```python
> def _process_toarray_args(self, order, out):
>     if out is not None:
>         if order is not None:
>             raise ValueError('order cannot be specified if out '
>                              'is not None')
>         if out.shape != self.shape or out.dtype != self.dtype:
>             raise ValueError('out array must be same dtype and shape as '
>                              'sparse array')
>         out[...] = 0.
>         return out
>     else:
>         return np.zeros(self.shape, dtype=self.dtype, order=order)
> ```
> 
> The assignment of every element to 0.0 is slower than just allocating a new array.

Interesting, nice digging! FYI @perimosocordiae @dschult, looks like this before your time (gh-229, haven't seen <4 digits in an issue/PR number in a while!) but maybe you have an idea for improving this.

broesler:
I think the avenue to improve is the relatively slow slicing operation before converting to dense. I haven't gone through every detail of the slicing code on `csc_array` yet, but  it seems like maybe we could implement a `dense_column_slice` method that directly allocates and populates the dense array instead of creating the intermediate `csc_array` object. I will take a look and see if there are any ways to improve the process.

tylerjereddy:
I'm not comfortable merging this right before branching -- if it was ready a domain expert should have merged it since the last comment on Feb. 19th. In the meanwhile, I'm going to bump the milestone -- I need to get a move on for branching.

PR #24056, opened 2025-11-24
ENH: spatial: factor out `from_matrix` orthogonalization for speedup on known orthogonal matrices

scottshambaugh:
This factors out matrix orthogonalization when generating quaternions from matrices into a private function. This allows us to skip some moderately expensive checks (determinants and matmuls) when working with internal matrices that we know are already orthogonal, and should provide a speed boost. These checks were added in https://github.com/scipy/scipy/pull/22418

This also now skips singular value decomposition in `Rotation.from_matrix()` on eager backends when applicable.

Note that the algorithm to convert the matrix to quaternion still does include a small-angle orthogonalization which corrects for round-off error. See the reference from the docs: [F. Landis Markley, "Unit Quaternion from Rotation Matrix", Journal of guidance, control, and dynamics vol. 31.2, pp. 440-442, 2008.](https://arc.aiaa.org/doi/abs/10.2514/1.31730)

Originally suggested by @amacati in this comment: https://github.com/scipy/scipy/pull/23718#discussion_r2440576137

### DIFF
--- a/scipy/spatial/transform/_rotation_cy.pyx
+++ b/scipy/spatial/transform/_rotation_cy.pyx
@@ -503,12 +503,26 @@ def from_matrix(matrix):
     # algorithm described in [3]_. This will also apply another
     # orthogonalization step to correct for any small errors in the matrices
     # that skipped the SVD step above.
+    if is_single:
+        return _from_matrix_orthogonal(matrix[0])
+    return _from_matrix_orthogonal(matrix)
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def _from_matrix_orthogonal(matrix):
+    is_single = False
+    if matrix.shape == (3, 3):
+        matrix = matrix[np.newaxis, :, :]
+        is_single = True
+
     cdef double[:, :, :] cmatrix
     cmatrix = matrix
     cdef Py_ssize_t num_rotations = cmatrix.shape[0]
     cdef Py_ssize_t i, j, k
     cdef double[:] decision = _empty1(4)
     cdef int choice
+    cdef int ind
 
     cdef double[:, :] quat = _empty2(num_rotations, 4)
 
@@ -1258,8 +1272,8 @@ def align_vectors(a, b, weights=None, bint return_sensitivity=False):
         with np.errstate(divide='ignore', invalid='ignore'):
             sensitivity = np.mean(weights) / zeta * (
                     kappa * np.eye(3) + np.dot(B, B.T))
-        return from_matrix(C), rssd, sensitivity
-    return from_matrix(C), rssd, None
+        return _from_matrix_orthogonal(C), rssd, sensitivity
+    return _from_matrix_orthogonal(C), rssd, None
 
 
 @cython.embedsignature(True)
--- a/scipy/spatial/transform/_rotation_xp.py
+++ b/scipy/spatial/transform/_rotation_xp.py
@@ -71,6 +71,14 @@ def from_matrix(matrix: Array) -> Array:
     orthogonal_matrix = U @ Vt
     matrix = xp.where(is_orthogonal, matrix, orthogonal_matrix)
 
+    return _from_matrix_orthogonal(matrix)
+
+
+def _from_matrix_orthogonal(matrix: Array) -> Array:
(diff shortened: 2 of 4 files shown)


scottshambaugh:
CI failure is unrelated

scipy/spatial/transform/_rotation_xp.py
```diff
@@ -71,6 +71,14 @@ def from_matrix(matrix: Array) -> Array:
     orthogonal_matrix = U @ Vt
     matrix = xp.where(is_orthogonal, matrix, orthogonal_matrix)
```
amacati: While we are at it, should we make the SVD optional for eager frameworks?

scipy/spatial/transform/_rotation_xp.py
```diff
@@ -759,7 +767,7 @@ def _align_vectors(a: Array, b: Array, weights: Array) -> tuple[Array, Array, Ar
     kappa = s[..., 0] * s[..., 1] + s[..., 1] * s[..., 2] + s[..., 2] * s[..., 0]
     eye = xp.eye(3, dtype=a.dtype, device=device)
     sensitivity = xp.mean(weights) / zeta * (kappa * eye + B @ B.mT)
-    q_opt = from_matrix(C)
+    q_opt = _from_matrix_orthogonal(C)
```
amacati, line 775: Given that u and vh are the results of the SVD, I assume C is guaranteed to be orthonormal?

scipy/spatial/transform/_rotation_xp.py
```diff
@@ -71,6 +71,14 @@ def from_matrix(matrix: Array) -> Array:
     orthogonal_matrix = U @ Vt
     matrix = xp.where(is_orthogonal, matrix, orthogonal_matrix)
 
+    return _from_matrix_orthogonal(matrix)
+
+
+def _from_matrix_orthogonal(matrix: Array) -> Array:
```
amacati, line 82: Should this be orthonormal (orthogonal and normal columns/rows)?

scipy/spatial/transform/_rotation_cy.pyx
```diff
@@ -503,12 +503,26 @@ def from_matrix(matrix):
...
     # orthogonalization step to correct for any small errors in the matrices
     # that skipped the SVD step above.
+    if is_single:
+        return _from_matrix_orthogonal(matrix[0])
+    return _from_matrix_orthogonal(matrix)
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def _from_matrix_orthogonal(matrix):
```
amacati, line 513: Should this be orthonormal (orthogonal and normal columns/rows)?

amacati (commented):
Thanks for working on this! Have you checked by how much this increases the speed of affected functions?

I have some smaller questions, but looks pretty good already.

scipy/spatial/transform/_rotation_xp.py
scottshambaugh: It is!

scipy/spatial/transform/_rotation_xp.py
scottshambaugh: Some discussion on orthogonal vs orthonormal here, I'll keep it as-is for now: https://github.com/scipy/scipy/pull/22418#discussion_r1933465042

lucascolley: my linear algebra course at university also told me that "orthogonal matrices" have normal rows

scipy/spatial/transform/_rotation_xp.py
scottshambaugh: Sure thing, added that change in a second commit

scottshambaugh:
Addressed your comments, I don't have benchmarking set up locally so can't test the speedup unfortunately.

scipy/spatial/transform/_rotation_xp.py
amacati: We made the distinction at my university, but both are fine. In this context, it's clear what was meant anyways.

amacati (approved):
LGTM, thanks Scott!

lucascolley (approved):
thanks both!

PR #24059, opened 2025-11-25
CI: revert pre-release NumPy and no isolation from np2 release days

### LINKED ISSUE #19826: MAINT: follow-ups to gh-19724 (pytest pins, temporary CI changes, special ufunc type coercion)
Some reminders after merging gh-19724, if only for myself...:

- [ ] the hard pin of `pytest` version used by `cibuildwheel` should probably be reverted once the `pytest` 8 release candidate issues for Windows + Python < `3.11.0` + NumPy pre-release are resolved
- [ ] we should try to figure out if there's an issue on our end for the above `pytest` RC + Windows + NumPy pre-release + Python < `3.11.0` matter, or if we just need to wait for a fix from https://github.com/pytest-dev/pytest/issues/9765 (or some combination of fixes on both ends)
- [ ] when NumPy `2.0.0` is out, we can of course restore build isolation and turn off the pre-release package pull-ins for both the `sdist` CI job that was modified in above PR, and for our regular wheel builds
- [ ] the `special` `ufunc` input `long` type shims may require a revisit (see comment from Albert in above PR), but were effectively considered "ok" in the short term; need to exercise caution

lucascolley:
closes gh-19826

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -132,13 +132,12 @@ jobs:
 
       - name: pip-packages
         run: |
-          python -m pip install build delvewheel cython pybind11 meson-python meson ninja pytest pytest-xdist pytest-timeout pooch hypothesis
-          python -m pip install --pre --upgrade --timeout=60 -i https://pypi.anaconda.org/scientific-python-nightly-wheels/simple numpy
+          python -m pip install build delvewheel numpy cython pybind11 meson-python meson ninja pytest pytest-xdist pytest-timeout pooch hypothesis
 
       - name: Build
         shell: bash
         run: |
-          python -m build --no-isolation -x -Csetup-args="-Duse-pythran=false"
+          python -m build -Csetup-args="-Duse-pythran=false"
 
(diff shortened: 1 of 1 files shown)


tylerjereddy (approved):
LGTM, thanks Lucas!

I can't remember why 2024 Tyler added the `-x`, we usually want to see the full accounting of failures in the CI, but the reversion seems right indeed.

tylerjereddy:
Completely unrelated, but what the heck is this (I had to pick another email address--I've never seen that before...):

image

lucascolley:
> Completely unrelated, but what the heck is this (I had to pick another email address--I've never seen that before...):
> 
> image

Do you use the refined GitHub browser extension?

I'm not sure if it is buggy behaviour from that or from GitHub itself, but I frequently see a message like that when the interface seems confused between squash merge and regular merge. E.g. sometimes it appears like I'm squashing and yet it says the commit will be authored by me rather than the PR author — then somewhere further down the chain it realises the problem and says invalid email.

Could be a completely unrelated issue though, or my diagnosis might be incorrect...

Issue #23033, opened 2025-05-22
Some public sub-submodules still expose private objects


rgommers:
This is a follow-up to gh-14360. That issues treated all top-level public modules, however some public API one level deeper was not taken along. One `sparse.csgraph` and `sparse.linalg` were included in the tracker in the issue description. We have a fair amount of other (more niche) subsubmodules that are public. 

From https://github.com/scipy/scipy/issues/14360#issuecomment-2898421901:
```python
from scipy.cluster.vq import xpx, np, check_random_state
from scipy.cluster.hierarchy import xp_capabilities, array_namespace, is_dask
from scipy.io.wavfile import warnings, struct, sys
from scipy.linalg.blas import functools
from scipy.linalg.lapack import np
from scipy.linalg.interpolative import np
from scipy.spatial.distance import np
from scipy.stats.contingency import np
```

The list of modules considered public and tested as such is here:
https://github.com/scipy/scipy/blob/02c41130f6ca23257e805ea61ec493d4f1b4978b/scipy/_lib/tests/test_public_api.py#L33

The ones with nesting level 3 deep cleaning up. Leaked namespaces that aren't from SciPy itself `(sys`, `np`) can just be deleted. If private scipy-specific functions/objects are exposed and could potentially be in use, they may need to be deprecated first. The `__all__` dicts of each module should also be inspected for completeness.

lucascolley:
bumping per https://github.com/scipy/scipy/pull/24060#issuecomment-3583669465

lucascolley:
subsumed by gh-24322

PR #24060, opened 2025-11-25
DEP/MAINT: cluster: deprecate `vq.py_vq` and tidy public API

lucascolley:
Towards gh-24322.

---

This PR:
- deprecates the unintentionally public (but perhaps used) `cluster.vq.py_vq`
- restructures the `cluster` directory to cleanly separate files under the `cluster.vq` and `cluster.hierarchy` namespaces
	- this includes splitting the contents of `cluster/meson.build`
- introduces `cluster.vq.__init__.py` and `cluster.hierarchy.__init__.py` which define the public API for each namespace
	- this includes moving the contents of the previous `vq.py` and `hierarchy.py` files to new `..._impl.py` files
		- note that the `_hierarchy` and `_vq` module names are unavailable without renaming the Cython extension modules
- makes `cluster.vq.ClusterError` and `cluster.hierarchy.ClusterWarning` officially public by adding docstrings
	- maybe these should be exposed in the `cluster` namespace instead, https://github.com/scipy/scipy/pull/24060#discussion_r2675944165
- makes the unintentionally public but trivial attribute `cluster.hierarchy.int_floor` officially private by prefixing its name with an `_`


ev-br (commented):
`import numpy as _np  # noqa: ICN001` looks really weird. Could it work to define a dunder all instead?

lucascolley:
> Could it work to define a dunder all instead?

I assumed that that wouldn't work, but great if so! How should I check?

j-bowhay:
Most of these do actually have a `__all__`! This only defines what is imported when you use a wildcard not what you can import fullstop. I think one pattern that would work is (for example) is to move all the code from `vq.py` to `_vq.py` and then in `vq.py` you just have `from ._vq import *`. This is basically the pattern top level modules use in their `__init__.py`.

lucascolley (commented):
@j-bowhay @ev-br what do you think of this approach that I have demonstrated here for `cluster`? It seems like quite a nice refactor to me. I have identified the undocumented public attributes which I have found.

scipy/cluster/hierarchy/__init__.py
```diff
@@ -0,0 +1,114 @@
...
+    ClusterNode, DisjointSet, average, centroid, complete, cophenet, correspond,
+    cut_tree, dendrogram, fcluster, fclusterdata, from_mlab_linkage, inconsistent,
+    is_isomorphic, is_monotonic, is_valid_im, is_valid_linkage, leaders,
+    leaves_list, linkage, maxRstat, maxdists, maxinconsts, median, num_obs_linkage,
+    optimal_leaf_ordering, set_link_color_palette, single, to_mlab_linkage, to_tree,
+    ward, weighted
+)
+
+# TODO: deprecate this unintentionally public attribute
+from ._hierarchy_impl import ClusterWarning
```
j-bowhay: I actually think we'd probably want this in the public api anyway

scipy/cluster/vq/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+image of interest was of an ocean, we would expect many 24-bit
+blues to be represented by 8-bit codes. If it was an image of a
+human face, more flesh-tone colors would be represented in the
+code book.
+
+"""
+from ._vq_impl import kmeans, kmeans2, vq, whiten
+
+# TODO: deprecate these unintentionally public attributes
+from ._vq_impl import ClusterError, py_vq
```
j-bowhay: and probably `ClusterError` too

lucascolley:
makes sense! How would we go about deprecating `py_vq` with this structure @j-bowhay ?

j-bowhay:
> makes sense! How would we go about deprecating `py_vq` with this structure @j-bowhay ?

Something along the lines of
```
def __getattr__(name):
    if name == "deprecated thing":
        # raise warning
    try:
        return globals()[name]
    except KeyError:
        raise AttributeError(
            f"Module 'scipy.vq' has no attribute '{name}'"
        )
```

lucascolley:
any existing examples of that or is this new?

j-bowhay:
https://github.com/scipy/scipy/blob/fc8d4e7f85acf8d85730e235993677eb67ac3c9f/scipy/sparse/__init__.py#L336-L350

https://github.com/scipy/scipy/blob/fc8d4e7f85acf8d85730e235993677eb67ac3c9f/scipy/sparse/sparsetools.py#L10-L17

lucascolley:
perhaps we make this PR an exemplar covering cluster, then convert gh-23033 into a tracker issue for the rest of the nested public modules, milestoned for 1.18.0

lucascolley:
@j-bowhay IIUC this covers the case of

```python
from scipy.cluster import vq
vq.py_vq(...)
```

but what about this?

```python
from scipy.cluster.vq import py_vq
py_vq(...)
```

scipy/cluster/hierarchy/__init__.py
```diff
@@ -0,0 +1,116 @@
...
+   :toctree: generated/
+
+   DisjointSet -- data structure for incremental connectivity queries
+   
+Warnings:
+    
+.. autosummary::
+   :toctree: generated/
+
+   ClusterWarning
```
lucascolley, line 97: it's weird that we have `cluster.hierarchy.ClusterWarning` and `cluster.vq.ClusterError`, rather than both belonging to the `cluster` namespace. But I think exposing them there is probably out of scope for this PR.

or maybe not

lucascolley:
gentle ping @j-bowhay on https://github.com/scipy/scipy/pull/24060#issuecomment-3589955901

lucascolley:
I've updated the top-post with a full description of the changes in this diff. Please let me know if anything should be split out into a separate PR.

j-bowhay:
> ```python
> vq.py_vq
> ```

Hmm something odd is happening:
```
In [1]: from scipy.cluster.vq import py_vq
meson-python: building scipy: /home/up19056/.local/share/mamba/envs/scipy-dev/bin/ninja
[2/356] Generating subprojects/highs/highs/HConfig.h with a custom command
> /home/up19056/development/scipy/scipy/cluster/vq/__init__.py(89)__getattr__()
-> breakpoint()
(Pdb) pp name
'__wrapped__'
(Pdb)
```

lucascolley:
the `__getattr__` seems to be bypassed for me?

```
>>> from scipy.cluster.vq import py_vq
>>> from scipy.cluster.vq import foo
> c:\users\lucas\ghq\github.com\scipy\scipy\build-install\lib\site-packages\scipy\cluster\vq\__init__.py(89)__getattr__()
-> breakpoint()
(Pdb) name
'foo'
```

lucascolley:
ah I think we need to intercept `__dict__` too

scipy/cluster/vq/__init__.py
```diff
@@ -0,0 +1,99 @@
...
+
+def __getattr__(name):
+    if name == "py_vq":
+        msg = (
+            "`scipy.cluster.vq.py_vq` was unintentionally public,"
+            " use `scipy.cluster.vq.vq` instead."
+        )
+        warnings.warn(msg, DeprecationWarning, stacklevel=2)
+        from ._vq_impl import py_vq
+        return py_vq
```
lucascolley: I think checking the name before we import the function is the only robust way of ensuring the warning is emitted @j-bowhay

tach.toml
```diff
@@ -24,11 +24,11 @@ depends_on = ["scipy._lib", "scipy.cluster.hierarchy", "scipy.cluster.vq"]
 
 [[modules]]
 path = "scipy.cluster.hierarchy"
-depends_on = ["scipy._lib", "scipy.cluster", "scipy.spatial.distance"]
+depends_on = ["scipy._lib", "scipy.spatial.distance"]
 
 [[modules]]
 path = "scipy.cluster.vq"
-depends_on = ["scipy._lib", "scipy.cluster", "scipy.spatial.distance"]
+depends_on = ["scipy._lib", "scipy.spatial.distance"]
```
lucascolley, line 35: nice confirmation :)

scipy/_lib/tests/test_warnings.py
```diff
@@ -40,8 +40,14 @@ def visit_Call(self, node):
...
-            # get first argument of the `args` node of the filter call
-            match node.args[0]:
+            action = None
+            # get the action argument
+            for kw in node.keywords:
+                if kw.arg == "action":
+                    action = kw.value
+            action = action or node.args[0]
+            
+            match action:
```
lucascolley: this wasn't robust to specifying `action` as a kwarg

scipy/cluster/vq/_vq_impl.py
```diff
@@ -276,6 +212,13 @@ def py_vq(obs, code_book, check_finite=True):
     return code, min_dist
 
 
+_py_vq_dep_msg = (
+    "`scipy.cluster.vq.py_vq` was unintentionally public,"
+    " use `scipy.cluster.vq.vq` instead."
+)
+py_vq = _deprecated(_py_vq_dep_msg)(_py_vq)
```
lucascolley, line 219: @j-bowhay now that we have worked through the `getattr` business, I think we are probably better off with a more traditional 'warn-at-call-time' deprecation. I don't think we need to be too worried about code being broken by unused imports?

refguide-check was assuming that deprecated things warn at call-time

pixi.toml
```diff
@@ -243,7 +243,7 @@ scipy-stubs = "*"
 ### Documentation ###
 
 [feature.docs.dependencies]
-sphinx = "*"
+sphinx = "<8.2.0"
```
lucascolley, line 251: `refguide-check` needs some love to bump to recent sphinx versions

EDIT: done in gh-24336

pixi.toml
```diff
@@ -257,6 +257,7 @@ jupyterlite-sphinx = "*"
 jupyterlite-pyodide-kernel = "*"
 scipy-doctest = "*"
 linkify-it-py = "*"
+make = ">=4.4.1,<5"
```
lucascolley, line 265: windows

scipy/cluster/vq/_vq_impl.py
j-bowhay: Sounds reasonable

lucascolley:
Ready from my side

scipy/cluster/vq/_vq_impl.py
```diff
@@ -1,87 +1,23 @@
...
+from scipy._lib.deprecation import _deprecated
 from scipy._lib import array_api_extra as xpx
 from scipy.spatial.distance import cdist
 
 from . import _vq
 
-__docformat__ = 'restructuredtext'
-
-__all__ = ['whiten', 'vq', 'kmeans', 'kmeans2']
+__all__ = ['ClusterError', 'kmeans', 'kmeans2', 'py_vq', 'vq', 'whiten']
```
lucascolley, line 14: maybe we want to include things like `_py_vq, _krandinit, _vq` also as they are imported by tests

j-bowhay: I think this is still fine right, you can still import them just not by a wildcard import?

lucascolley: yes I don't think there is a meaningful difference at runtime, I suppose I am not clear on the intended semantics of `__all__` for private modules

scipy/cluster/vq/_vq_impl.py
j-bowhay: Should we say that we will remove py_vq in 1.20?

lucascolley: sure if that is your preference 👍

scipy/cluster/vq/_vq_impl.py
```diff
@@ -276,6 +212,13 @@ def py_vq(obs, code_book, check_finite=True):
     return code, min_dist
 
 
+_py_vq_dep_msg = (
+    "`scipy.cluster.vq.py_vq` was unintentionally public,"
+    " use `scipy.cluster.vq.vq` instead."
```
j-bowhay:
```suggestion
    "`scipy.cluster.vq.py_vq` was unintentionally public, "
    "and will be removed in SciPy 1.20.0, use `scipy.cluster.vq.vq` instead."
```

j-bowhay (approved):
Thanks!

scipy/cluster/vq/_vq_impl.py
```diff
@@ -221,7 +157,7 @@ def vq(obs, code_book, check_finite=True):
     return py_vq(obs, code_book, check_finite=False)
```
hawkinsp, line 157: I suspect this code should be calling _py_vq not py_vq, since otherwise a DeprecationWarning is issued.

j-bowhay: Oops, thanks!

PR #24063, opened 2025-11-26
TST: linalg/solve: reenable and fix the "pos def" solve test

ev-br:
#### Reference issue

splt off from https://github.com/scipy/scipy/pull/23547#discussion_r2558547821

Add a test for pos.def solve, make sure that the input matrix is clearly positive definite.


tylerjereddy (approved):
I asked for this to be split off--I don't see a problem with these changes since we're just checking for consistency with NumPy `solve` and for consistency whether `assume_a` is specified or not.

Since no algorithmic source code is changed, only a few missing test cases restored, this seems relatively safe to merge. Even if the underlying algorithmic code isn't perfect, having a test we could tune if issues are discovered makes sense rather than having no tests that flush these `assume_a` cases.

tylerjereddy:
thanks Evgeni

PR #24064, opened 2025-11-26
DOC: add documentation about sanitizers

FFY00:
#### Reference issue

#23612

This adds documentation about sanitizer, with a focus on the address sanitizer, whose usage is currently only documented on the bug tracker, and hard to follow.

It also documents a couple common issues with obscure error messages when trying to setup and use the address sanitizer.

#### Additional information

It recommends a manual environment setup, but as mentioned on the bug tracker, using a tool like pixi might provide a better DX.

### DIFF
--- a/doc/source/dev/contributor/sanitizers.rst
+++ b/doc/source/dev/contributor/sanitizers.rst
@@ -130,6 +130,30 @@ This may occur if the ``asan-ignore.txt`` file passed to
 You can verify this in the Meson log.
 
 
+THe ASAN runtime failing to intercept libc++
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In some cases, such as when loading an extension module that links against
+libc++, the ASAN runtime has some trouble intercepting symbols like
+``__cxa_throw`` to hook into them. This then leads to the interceptor handling
+code to fail.
+
+.. code-block::
+
+    AddressSanitizer: CHECK failed: asan_interceptors.cpp:463 "((__interception::real___cxa_throw)) != (0)" (0x0, 0x0) (tid=77864)
+
+The best solution would be linking the extension modules against an instrumented
+build of libc++, however this can make the execution environment extremely hard
+to setup.
+
+However, a commonly viable work-around is to have the linker load libc++ before
+anything else.
+
+.. code-block::
+
+    $ export LD_PRELOAD=/usr/lib/libstdc++.so
+
+
 Incompatible Runtimes
 ~~~~~~~~~~~~~~~~~~~~~
 
--- a/doc/source/dev/contributor/sanitizers.rst
+++ b/doc/source/dev/contributor/sanitizers.rst
@@ -1,0 +1,161 @@
+.. _scipy-sanitizers:
+
+==========
+Sanitizers
+==========
+
+Sanitizers are compiler features that allow us to add run-time checks to detect
+multiple types of bugs. They usually consist of an instrumentation module, which
+adds error detection mechanisms to the generated code, and a supporting run-time
+library.
+
+There are multiple kinds of sanitizers, with support varying between compilers,
+but the most common is the address sanitizer — often abbreviated to "ASAN" —
+which checks for memory bugs, such as buffer overflows.
+
+
+Side-effects
+------------
+
+Enabling sanitizers will almost always result in very significant slowdowns
+and size increases to the generated objects, due to the added code
+instrumentation. It can also make the build workflow substantially more complex.
+
+Moreover, the resulting artifacts often require a carefully setup execution
+environment, as well as a substantially more complex build workflow, especially
+when multiple components are used together (eg. the Python interpreter and
+extension modules). Even in the correct execution environment, the run-time
+operation of the sanitizers has some sharp edges, as they're intended to be a
+developer tool.
+
+Consequently, sanitizers are generally only suitable to be used during
+development. Even so, in larger projects like SciPy, it's usually best not to
+use them as a part of your general development workflow. Rather, only to
+validate code when necessary.
+
+
+:ref:`CI <continuous-integration>` Coverage
+-------------------------------------------
+
+.. list-table::
+
+    * - Type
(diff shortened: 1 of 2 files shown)


doc/source/dev/contributor/sanitizers.rst
```diff
@@ -0,0 +1,161 @@
...
+recommended:
+
+#. Set ``ASAN_OPTIONS=detect_leaks=0:symbolize=1:strict_init_order=true:allocator_may_return_null=1:use_sigaltstack=0``
+#. Compile CPython_ with ``--with-address-sanitizer`` and use it to setup a test environment
+#. Build NumPy_ with the ``setup-args`` config setting set to ``-Db_sanitize=address``
+    * Eg. ``pip install numpy --no-binary numpy -Csetup-args="-Db_sanitize=address" -v``
+#. In the project root directory, set the compiler flags as follows:
+    * ``CFLAGS="-fsanitize=address -fno-omit-frame-pointer -fsanitize-ignorelist=$(pwd)/tools/asan-ignore.txt"``
+    * ``CXXFLAGS="-fsanitize=address -fno-omit-frame-pointer -fsanitize-ignorelist=$(pwd)/tools/asan-ignore.txt"``
+#. Build SciPy using ``spin build --clean``
```
lucascolley, line 72: I think these options should instead go here? https://github.com/scipy/scipy/blob/fc8d4e7f85acf8d85730e235993677eb67ac3c9f/.spin/cmds.py#L120-L121

```suggestion
#. Build SciPy using ``spin build --clean --asan``
```

lucascolley (changes requested):

doc/source/dev/contributor/sanitizers.rst
lucascolley: this seems to do the trick for me:

```python
if asan:
    meson_args = meson_args + ('-Db_sanitize=address,undefined', )
    build_dir = os.path.abspath(kwargs['build_dir'])
    root = Path(build_dir).parent
    asan_ignore_file = os.path.join(root, 'tools', 'asan-ignore.txt')
    compiler_args = f'-fno-omit-frame-pointer -fsanitize-ignorelist={asan_ignore_file}'
    meson_args = meson_args + (f'-Dc_args={compiler_args}', )
    meson_args = meson_args + (f'-Dcpp_args={compiler_args}', )
```

lucascolley: Meson's `-b_sanitize` actually seems to hit problems with conda-forge's gfortran here:

```
ERROR: Linker /Users/lucascolley/ghq/github.com/rgommers/pixi-dev-scipystack/scipy/scipy/.pixi/envs/asan/bin/arm64-apple-darwin20.0.0-gfortran does not support sanitizer arguments ['-fsanitize=address']
```

just putting everything in `c_args` and `cpp_args` seems fine

EDIT: maybe we want `c_link_args` etc. too

EDIT2: this was already noted at https://github.com/scipy/scipy/blob/main/.github/workflows/macos.yml#L203

lucascolley (commented):
> using a tool like pixi might provide a better DX

Okay, you got me... this should really be as simple as `pixi run test-asan`!!

After being nerd-sniped for the day, I raise you gh-24066.

doc/source/dev/contributor/sanitizers.rst
FFY00: We could possibly manually remove the `b_sanitize` option for Fortran in Meson.

FFY00:
Let's see in which shape we get the pixi PR, then we can update this based on that. I would like to at least keep the common problems section for folks setting up their own environment, as the errors are really non-obvious to fix.

doc/source/dev/contributor/sanitizers.rst
```diff
@@ -0,0 +1,185 @@
+.. _scipy-sanitizers:
+
+==========
+Sanitizers
+==========
+
+Sanitizers are compiler features that allow us to add run-time checks to detect
+multiple types of bugs. They usually consist of an instrumentation module, which
```
rgommers, line 8: I'd add "in compiled code" or "in C/C++ code"

doc/source/dev/contributor/sanitizers.rst
rgommers: I'm not sure you can do that, since `b_sanitize` is a `meson setup` argument and not specific to a language.

doc/source/dev/contributor/sanitizers.rst
```diff
@@ -0,0 +1,185 @@
...
+.. _scipy-asan:
+
+Address Sanitizer (abbv. "ASAN")
+================================
+
+As mentioned above, the address sanitizer helps detect memory bugs.
+
+To setup the LLVM address sanitizer in SciPy, the following workflow is
+recommended:
+
```
rgommers, line 64: Linking to https://numpy.org/devdocs/dev/development_advanced_debugging.html#compiler-sanitizers may also be useful.

doc/source/dev/contributor/sanitizers.rst
```diff
@@ -0,0 +1,185 @@
...
+You can verify this by running ``numpy-config``. If it fails with the address
+sanitizer reporting a memory leak, like below, set the ``ASAN_OPTIONS``
+environment as specified in the workflow instructions.
+
+.. code-block:: sh
+
+    $ numpy-config
+    (...)
+    SUMMARY: AddressSanitizer: 3027 byte(s) leaked in 98 allocation(s).
```
rgommers, line 117: Hmm that is odd. There are TSan and UBSan suppression files in the NumPy repo, but ASan is clean in CI:
https://github.com/numpy/numpy/blob/09aa74efbe1ae421079e860c6e775f34814b1bf0/.github/workflows/compiler_sanitizers.yml#L76-L77

@ngoldbaum may be keen to see these perhaps

rgommers (commented):
Thanks @FFY00! This is quite useful, and the troubleshooting tips are valuable even if we do end up achieving a very polished build setup with `pixi`.

A few comments, but this looks close to ready after a first read.

PR #24067, opened 2025-11-26
DEV: add test-coverage and test-xslow tasks

lucascolley:
Towards gh-23848

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -15,8 +15,8 @@ features = ["build-deps", "build-task", "win-openblas"]
 solve-group = "default"
 
 [environments.test]
-# tasks: test
-features = ["run-deps", "test-deps", "test-task", "win-openblas"]
+# tasks: test, test-coverage, test-xslow
+features = ["run-deps", "test-deps", "test-tasks", "win-openblas"]
 solve-group = "default"
 
 [environments.docs]
@@ -181,6 +181,7 @@ description = "Build SciPy (default settings)"
 [feature.test-deps.dependencies]
 pytest = "*"
 hypothesis = "*"
+pytest-cov = "*"
 pytest-timeout = "*"
 pytest-xdist = "*"
 threadpoolctl = "*"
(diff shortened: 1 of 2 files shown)


pixi.toml
```diff
@@ -189,11 +190,22 @@ mpmath = "*"
...
-[feature.test-task.tasks.test]
+[feature.test-tasks.tasks.test]
 cmd = "spin test --no-build"
 depends-on = "build"
 description = "Test SciPy (default settings)"
 
+[feature.test-tasks.tasks.test-coverage]
+cmd = "spin test --no-build --coverage"
+depends-on = "build"
+description = "Test SciPy with coverage"
```
tylerjereddy, line 201: I spot checked `pixi run test-coverage` and `pixi run test-coverage -j 12` in a clean checkout on x86_64 Linux and everything worked as expected.

The rest of the diff looks fine to me. I've not been following the broader plans we have for `pixi`, but I use it enough now to be comfortable merging this I think (larger structural changes maybe not though).

tylerjereddy (approved):

tylerjereddy:
thanks Lucas

lucascolley:
thanks for the review Tyler, and great to hear you have started using Pixi!

WarrenWeckesser:
Is there a plan or roadmap for dealing with what appears to be an overlap between your work with `pixi` and with `spin`?  I would have expected the option to run the XSLOW tests to be incorporated into `spin test` somehow.

lucascolley:
> Is there a plan or roadmap for dealing with what appears to be an overlap between your work with `pixi` and with `spin`? I would have expected the option to run the XSLOW tests to be incorporated into `spin test` somehow.

Seems good to me to move things into `spin` where it makes sense. I only use Pixi so I haven't seen the need to put in the extra effort to figure out where things should ideally go.

tylerjereddy:
Yeah, there could be some debate on which abstractions are the right ones. Even with `spin`, for some stuff like benchmarking, I often end up back at the native tool for some things with particularly tricky/specific incantations.

Still, a few basic `pixi` tasks struck me as fairly harmless given how convenient it is to get from empty checkout to working incantation there, and easy to ignore for non-pixi devs.

lucascolley:
In general I think one-line spin invocations are probably fine to keep at the Pixi level. If we start scripting-proper then that should probably be moved to Python instead of TOML.

Issue #19462, opened 2023-11-02
Vet SciPy for use of `long` as the default integer


seberg:
We just merged https://github.com/numpy/numpy/pull/24224 changing the default integer in NumPy.  This changes the definition of `np.int_` to intp, which only matters on windows, though.

There are three things that might be good to ensure here:
1. In some cases `np.int_` is probably used as long, and needs to be `np.dtype("long").type` (compatible version.
2. I think I saw a cython use `np.int_t`, which is ill defined unfortunately. It needs to be either `npy_long` (because it is hard to create a compatible name).  Or `intp` (with potential casting if required).  Worst case, you need to support both, which would be annoying I admit (fused type).
3. It probably doesn't matter much, but I think many special functions are defined with long loops.  In principle it may make sense to make them int or int64 loops to ensure 64bit integers can match them.

It is very likely that this will break something on windows at this moment, the Cython issue should break compilation.

tylerjereddy:
Some of this is dealt with in gh-19466, now merged, but there are a few broader points above so leaving this open for now.

tylerjereddy:
I think we fixed most of the issues here for `1.13.0`, in large part by kicking the can down the road with `np.dtype("long")`.

That said, there is a point about `special` function loops types above that I'm not tracking, and we have lots of C++ rewrite activity happening in `special` now, so I'll bump the milestone in case that particular point remains relevant.

ev-br:
Would be great to act on the remaining `np.dtype("long")` uses:

```
$ git grep 'np.dtype("long")'
scipy/interpolate/_rbfinterp.py:    out = np.zeros((nmonos, ndim), dtype=np.dtype("long"))
scipy/linalg/_decomp_schur.py:        a1 = asarray(a, dtype=np.dtype("long"))
scipy/optimize/_trlib/_trlib.pyx:        self.iwork = np.zeros([iwork_size], dtype=np.dtype("long"))
scipy/optimize/_trlib/_trlib.pyx:                               dtype=np.dtype("long"))
scipy/special/_basic.py:            dt = np.dtype("long")
scipy/special/_lambertw.py:    k = np.asarray(k, dtype=np.dtype("long"))
scipy/special/_spherical_bessel.py:    n = np.asarray(n, dtype=np.dtype("long"))
scipy/special/_spherical_bessel.py:    n = np.asarray(n, dtype=np.dtype("long"))
[...]

PR #24070, opened 2025-11-27
MAINT: interpolate/RBF: use fixed-width ints with pythran

ev-br:
#### Reference issue

a part of gh-19462

Use fixed-width integer dtypes in the python/pythran interface, instead of `np.dtype("long")`  in the python side and `int[:]` in the pythran side.

#### Additional information

The whole `np.dtype("long")` thing is a remnant of the NumPy 2.0 transition. There's no reason to keep using it if we can help it.


tylerjereddy (approved):
CI seems happy, and cleaning up the old shim seems useful, thanks.

PR #24071, opened 2025-11-27
DEV: add `bench` task

lucascolley:


pixi.toml
```diff
@@ -29,6 +29,11 @@ solve-group = "default"
 features = ["run-deps", "test-deps", "ipython"]
 solve-group = "default"
 
+[environments.bench]
+# tasks: bench
+features = ["run-deps", "test-deps", "build-deps", "bench"]
```
tylerjereddy: What's the motivation for adding `test-deps` here? I was able to run `pixi run bench` just fine without it locally on x86-64 Linux starting from a clean checkout.

lucascolley: ah good point, I think we can remove them
```suggestion
features = ["run-deps", "build-deps", "bench"]
```

tylerjereddy:
thanks

PR #24073, opened 2025-11-27
ENH: sparse.sparray: add utility functions

### LINKED ISSUE #15849: ENH: Easy transformation betwean `spmatrix` and `_sparray`

All the constructor methods in `scipy.sparse` (e.g. `kron`, `eye`, etc) are not yet transferred to the new array interface. I really like the numpy compatible element-wise multiplication, but the usage of those new arrays is really cumbersome since you have to manually change every single format. I would suggest a simple switcher method to simply switch between  both formats.

A straight forward solution would be something like
```
def tosparray(matrix):
    if matrix._is_array == True: 
        return matrix
    if matrix.format == 'coo':
        return coo_array(matrix)
    if matrix.format == 'csr':
        return csr_array(matrix)
    # and so forth
    
def tospmatrix(array):
    if array._is_array == False: 
        return array
    if array.format == 'coo':
        return coo_matrix(array)
    if array.format == 'csr':
        return csr_matrix(array)
    # and so forth
```
This would make the interplay between those two formats easier and especially the new arrays way more accessible.

Of course, this could (and probably) should be done in a more flexible way than simply `if` statements, so that it ideally works for all `spmatrix` e.g. with something like
```
def tosparray(matrix):
    class TMP(_sparray, type(matrix)):
        pass    
    return TMP(matrix)
```
Unfortunatly, the result would be a `_sparray` of type `TMP` and not e.g. of type `coo_array` which is a problem e.g. for `isspmatrix_coo`. But i think a quite easy solution is from this point not to far away.
    

_No response_

_No response_

mfeuerle:
#### Reference issue
Closes #15849

I felt that the new `scipy.sparse.sparray` lacked some utility support compared to the now deprecated `scipy.sparse.spmatrix`, in particular, there were only functions to check for `spmatrix` such as `isspmatrix` and `isspmatrix_{format}`. I extended that to also include `issparray` and `issparray_{format}` to check for the new classes. In addition to the already existing `issparse` function i also added `issparse_{format}` to check for a given format, independent of the array or matrix type.

Further, it was quite hard to switch to the new `sparray` format when developing new code, because most of the existing libraries are still based on the deprecated `spmatrix` format. Thus, when getting a sparse object `A` from a library, one had to first check if it is an `spmatrix` and if so figure out the format to then manually call `A = sparray_{format}(A)`. Thus i also added to new methods to `_spbase` that allow now to conveniently switch between these two formats by calling  `A = A.tosparray(copy=?)` or `A = A.tospmatrix(copy=?)`.

#### Additional information

I fully documented them and wrote pytests for all new functions, should be ready to go.

### DIFF
--- a/scipy/sparse/__init__.py
+++ b/scipy/sparse/__init__.py
@@ -19,7 +19,8 @@ SciPy 2-D sparse array package for numeric data.
    This package is switching to an array interface, compatible with
    NumPy arrays, from the older matrix interface.  We recommend that
    you use the array objects (`bsr_array`, `coo_array`, etc.) for
-   all new work.
+   all new work. You can check if an object is a sparse array or 
+   sparse matrix using `issparray` and `isspmatrix`, respectively.
 
    When using the array interface, please note that:
 
@@ -117,8 +118,14 @@ Identifying sparse arrays
 .. autosummary::
    :toctree: generated/
 
-   issparse - Check if the argument is a sparse object (array or matrix).
    issparray
+   issparray_csc
+   issparray_csr
+   issparray_bsr
+   issparray_lil
+   issparray_dok
+   issparray_coo
+   issparray_dia
 
 
 Sparse matrix classes
@@ -158,7 +165,6 @@ Identifying sparse matrices
 .. autosummary::
    :toctree: generated/
 
-   issparse
    isspmatrix
    isspmatrix_csc
    isspmatrix_csr
@@ -167,6 +173,23 @@ Identifying sparse matrices
    isspmatrix_dok
    isspmatrix_coo
    isspmatrix_dia
+   
+
+Identifying sparse objects (arrays and matrices)
+------------------------------------------------
+
+
+.. autosummary::
+   :toctree: generated/
+
+   issparse
+   issparse_csc
+   issparse_csr
+   issparse_bsr
+   issparse_lil
+   issparse_dok
+   issparse_coo
+   issparse_dia
 
 
 Warnings
--- a/scipy/sparse/__init__.py
+++ b/scipy/sparse/__init__.py
@@ -118,6 +118,7 @@ Identifying sparse arrays
    :toctree: generated/
 
    issparse - Check if the argument is a sparse object (array or matrix).
+   issparray
 
 
 Sparse matrix classes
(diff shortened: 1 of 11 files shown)


dschult:
Hi and thanks for the recommendations -- 
The sparray interface is designed for a world after deprecation of the spmatrix interface. 

So `issparse` is the way to check for sparray. Right now it also returns true for spmatrix, but we decided that `issparse(A) and not isspmatrix(A)` was not very cumbersome during the transition and would be easy to adjust upon removal of spmatrix.  We are aiming for `issparse` to be the function to check for sparse arrays, so we'd like to avoid adding other functions like `issparray` to do that.

Checking for format is done using an attribute rather than a function. That is, the old `isspmatrix_csr(A)` is replaced by `issparse(A) and A.format == "csr"`. This allows more flexible format criteria such as `A.format in ('csr', 'csc')`. And that means we chose to not provide functions like `issparse_csr`. I see that it is tempting to add them because they are in the spmatrix api. But we decided the better long term api we are aiming for would use more flexible attribute searches rather than functions that often ended up with more computation, e.g. `isspamtrx_csr(A) or isspmatrix_csc(A)` which duplicates the checks for sparse. 

The convenience methods `tosparray` and `tospmatrix` were decided against because they encourage code that would have to be changed when spmatrix is removed. Converting to a given format of sparray can just be e.g. `csr_array(A)`.  If you want to have the same (unknown) format in sparray form, use `getattr(scipy.sparse, A.format + "_array")(A)`.  It seems that most code knows what format it wants, so e.g. `csr_array(A)` works well (or put it inside `if isspmatrix(A) or A.format != "csr":` if you prefer).

I hope this explains our reasoning (as I understand it) for the relevant choices of API. The idea is to aim for our desired long term api and make it easy for people to switch without adding new method/function surface aimed at manipulating both types. I agree with you that additional methods would be desirable if we planned to keep both around. But I think the best way forward is to write the sparray code and then add easily identifiable and removable code that handles the spmatrix cases. Comments welcome! :)

mfeuerle:
@dschult I can see that argumentation, in the long run, that is most likely the better approach. I actually like the new way for checking the format quite a lot (maybe that should be added as a small note to the documentation?). I guess i just close the PR?

dschult:
Yes -- you are right that checking the format should be added to the documentation. Thanks!
(I'll close this PR but you can still comment on it and we can reopen if needed.)

PR #24075, opened 2025-11-27
DEV/BENCH: add `flamegraph` Pixi task for linux

lucascolley:
image

image

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -39,6 +39,17 @@ solve-group = "default"
 features = ["run-deps", "build-deps", "bench-common", "bench"]
 solve-group = "default"
 
+[environments.build-py313]
+# tasks: build-py313
+features = ["build-deps", "build-py313", "py313"]
+solve-group = "py313"
+
+[environments.profile]
+# tasks: flamegraph, py-spy
+# py313 as py-spy does not support py314 at the time of writing
+features = ["build-deps", "run-deps", "profile", "py313"]
+solve-group = "py313"
+
 [environments.build-debug]
 # tasks: build-debug
 features = ["build-deps", "build-debug"]
@@ -637,7 +648,56 @@ depends-on = "build-system-libs"
 description = "Test SciPy with system libs"
 
 
+### Profiling ###
+
+[feature.profile]
+platforms = ["linux-64", "osx-arm64"]
+
+[feature.profile.dependencies]
+py-spy = "*"
+flamegraph-pl = "*"
+
+[feature.profile.pypi-dependencies]
+flamegraph.git = "https://github.com/nickodell/python-flamegraph.git"
+
+[feature.profile.tasks._log_flamegraph]
+cmd = "spin run --build-dir=build-py313 'cd tools/profiling && python -m flamegraph -o {{ log_name }}.log {{ script_name }}.py'"
+args = ["script_name", "log_name"]
+depends-on = "build-py313"
+
+[feature.profile.tasks.flamegraph]
+cmd = "flamegraph {{ log_name }}.log > {{ log_name }}.svg"
+args = ["script_name", "log_name"]
+cwd = "tools/profiling"
+depends-on = [{ task = "_log_flamegraph", args = ["{{ script_name }}", "{{ log_name }}"]}]
+description = "Generate a flamegraph {{ log_name }}.svg from tools/profiling/{{ script_name }}.py"
+
+[feature.profile.tasks.py-spy]
+cmd = "spin run --build-dir=build-py313 'cd tools/profiling && py-spy record --native -o {{ log_name }}.svg -- python {{ script_name }}.py'"
+args = ["script_name", "log_name"]
+depends-on = "build-py313"
+description = "Generate a py-spy flamegraph {{ log_name }}.svg from tools/profiling/{{ script_name }}.py"
+
+[feature.profile.target.osx.tasks.py-spy]
+# `--native` is unsupported on osx, and `sudo` is required
+cmd = "sudo spin run --build-dir=build-py313 'cd tools/profiling && py-spy record -o {{ log_name }}.svg -- python {{ script_name }}.py'"
+args = ["script_name", "log_name"]
+depends-on = "build-py313"
+description = "WARNING: REQUIRES `sudo` — Generate a py-spy flamegraph {{ log_name }}.svg from tools/profiling/{{ script_name }}.py"
+
+
 ### Alternative python versions ###
+
+[feature.py313]
+platforms = ["linux-64", "osx-arm64"]
+
+[feature.py313.dependencies]
+python = "3.13.*"
+
+[feature.build-py313.tasks.build-py313]
+cmd = "spin build --build-dir=build-py313 --setup-args=-Dblas=blas --setup-args=-Dlapack=lapack --setup-args=-Duse-g77-abi=true"
+env = { CC = "ccache $CC", CXX = "ccache $CXX", FC = "ccache $FC" }
+description = "Build SciPy (with Python 3.13)"
 
 [feature.py312]
 platforms = ["linux-64", "osx-arm64", "win-64"]
--- a/tools/profiling/README.md
+++ b/tools/profiling/README.md
@@ -1,0 +1,35 @@
+# Flame profiling for SciPy
+
+## Usage
+
+To generate a [flame graph](https://www.brendangregg.com/flamegraphs.html) with https://github.com/nickodell/python-flamegraph,
+use:
+
+```console
+pixi run flamegraph script_name log_name
+```
+
+where `script_name` and `log_name` are such that
+the script you want to profile exists at `tools/profiling/script_name.py`,
+and `tools/profiling/log_name.log` and `tools/profiling/log_name.svg` will be generated.
+
+To generate a flame graph with py-spy, which can also profile native extensions on Linux, use:
+
+```console
+pixi run py-spy script_name log_name
+```
+
+> [!CAUTION]
+> `py-spy` requires `sudo` on macOS machines.
+> You may wish to establish trust in the [the codebase](https://github.com/benfred/py-spy)
+> before use.
+
+where `script_name` and `log_name` are such that
+the script you want to profile exists at `tools/profiling/script_name.py`,
+and `tools/profiling/log_name.log` and `tools/profiling/log_name.svg` will be generated.
+
+## Opportunities for future development
+
+- write this up in the developer documentation
+- support for line profiling
+- support for Windows?
--- a/pixi.lock
+++ b/pixi.lock
@@ -2093,6 +2093,191 @@ environments:
       - conda: https://prefix.dev/conda-forge/noarch/zipp-3.23.0-pyhcf101f3_1.conda
       - conda: https://prefix.dev/conda-forge/osx-arm64/zlib-1.3.1-h8359307_2.conda
       - conda: https://prefix.dev/conda-forge/osx-arm64/zstd-1.5.7-hbf9d68e_6.conda
+  build-py313:
+    channels:
+    - url: https://prefix.dev/conda-forge/
+    options:
+      pypi-prerelease-mode: if-necessary-or-explicit
+    packages:
+      linux-64:
+      - conda: https://prefix.dev/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2
+      - conda: https://prefix.dev/conda-forge/linux-64/_openmp_mutex-4.5-2_gnu.tar.bz2
+      - conda: https://prefix.dev/conda-forge/noarch/beniget-0.4.2.post1-pyhd8ed1ab_1.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/binutils-2.45-default_h4852527_105.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/binutils_impl_linux-64-2.45-default_hfdba357_105.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/binutils_linux-64-2.45-default_h4852527_105.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/blas-devel-3.11.0-5_h1ea3ea9_openblas.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/bzip2-1.0.8-hda65f42_8.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/c-compiler-1.11.0-h4d9bdce_0.conda
+      - conda: https://prefix.dev/conda-forge/noarch/ca-certificates-2026.1.4-hbd8a1cb_0.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/ccache-4.12.2-hedf47ba_0.conda
+      - conda: https://prefix.dev/conda-forge/noarch/click-8.3.1-pyh8f84b5b_1.conda
+      - conda: https://prefix.dev/conda-forge/noarch/colorlog-6.10.1-pyh707e725_0.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/compilers-1.11.0-ha770c72_0.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/conda-gcc-specs-14.3.0-he8ccf15_16.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/cxx-compiler-1.11.0-hfcd1e18_0.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/cython-3.2.4-py313hc80a56d_0.conda
+      - conda: https://prefix.dev/conda-forge/noarch/decorator-5.2.1-pyhd8ed1ab_0.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/fortran-compiler-1.11.0-h9bea470_0.conda
+      - conda: https://prefix.dev/conda-forge/noarch/gast-0.6.0-pyhd8ed1ab_0.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/gcc-14.3.0-h0dff253_16.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/gcc_impl_linux-64-14.3.0-he8b2097_16.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/gcc_linux-64-14.3.0-h298d278_17.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/gfortran-14.3.0-h76987e4_16.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/gfortran_impl_linux-64-14.3.0-h1a219da_16.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/gfortran_linux-64-14.3.0-h9ce9316_17.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/gxx-14.3.0-h76987e4_16.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/gxx_impl_linux-64-14.3.0-h2185e75_16.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/gxx_linux-64-14.3.0-h310e576_17.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/icu-78.2-h33c6efd_0.conda
+      - conda: https://prefix.dev/conda-forge/noarch/importlib-metadata-8.7.0-pyhe01879c_1.conda
+      - conda: https://prefix.dev/conda-forge/noarch/kernel-headers_linux-64-4.18.0-he073ed8_9.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/ld_impl_linux-64-2.45-default_hbd61a6d_105.conda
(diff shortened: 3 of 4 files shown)


pixi.toml
```diff
@@ -243,6 +248,29 @@ depends-on = "build"
...
+args = ["script_name", "log_name"]
+inputs = ["{{ script_name }}.py"]
+outputs = ["{{ log_name }}.log"]
+cwd = "tools"
+# XXX: PYTHONPATH will change when Python version changes
+env.PYTHONPATH = "$PIXI_PROJECT_ROOT/build-install/usr/lib/python3.14/site-packages"
+depends-on = "build"
+
+[feature.profile.tasks.flamegraph]
+cmd = "./flamegraph.pl {{ log_name }}.log > {{ log_name }}.svg"
```
lucascolley: can we package `flamegraph.pl` for conda-forge? Is it already there?

pixi.toml
```diff
@@ -243,6 +248,29 @@ depends-on = "build"
...
+flamegraph.git = "https://github.com/nickodell/python-flamegraph.git"
+
+[feature.profile.tasks._log_flamegraph]
+cmd = "python -m flamegraph -o {{ log_name }}.log {{ script_name }}.py"
+args = ["script_name", "log_name"]
+inputs = ["{{ script_name }}.py"]
+outputs = ["{{ log_name }}.log"]
+cwd = "tools"
+# XXX: PYTHONPATH will change when Python version changes
+env.PYTHONPATH = "$PIXI_PROJECT_ROOT/build-install/usr/lib/python3.14/site-packages"
```
lucascolley: this is obviously a hack

pixi.toml
```diff
@@ -29,6 +29,11 @@ solve-group = "default"
 features = ["run-deps", "test-deps", "ipython"]
 solve-group = "default"
 
+[environments.profile]
+# tasks: profile-shell
```
lucascolley:
```suggestion
# tasks: flamegraph
```

pixi.toml
nickodell: Side note: this is a hack that I find myself doing constantly. e.g. if I want to use `strace`, but I don't want to `strace` spin. Same for `gdb`. Would be really nice if `spin python` had a way to insert a 'trace command', which would be a command pre-pended to the Python command it executes.

ev-br: `$ spin run strace python -P ....`?

lucascolley: TIL `spin run`!

pixi.toml
lucascolley: https://github.com/conda-forge/staged-recipes/pull/31581

lucascolley: @nickodell I packaged `flamegraph-pl` (the version corresponding to the tagged release on GitHub) for conda-forge.

Maybe you would be interested in contributing `python-flamegraph`?

lucascolley (commented):
I now have `pixi run flamegraph bench_tfqmr tfqmr_flame` working, at least to some extent, generating 

Image

Is that about the best we can expect out of `python-flamegraph` @nickodell ?

For venturing into native profiling, I have struggled to get `py-spy` working on my linux box. With `sudo`, I run into a whole host of path-related problems, but judging by the py-spy readme, `sudo` shouldn't be necessary on linux? Without it, I see errors like `Error: Failed to find python version from target process`.

ev-br:
> For venturing into native profiling, I have struggled to get py-spy working on my linux box. With sudo, I run into a whole host of path-related problems, but judging by the py-spy readme, sudo shouldn't be necessary on linux?

FWIW, working in a mamba env on linux
 

```
$ py-spy
py-spy: command not found
$ pip install py-spy
Collecting py-spy
  Using cached py_spy-0.4.1-py2.py3-none-manylinux_2_5_x86_64.manylinux1_x86_64.whl.metadata (510 bytes)
Using cached py_spy-0.4.1-py2.py3-none-manylinux_2_5_x86_64.manylinux1_x86_64.whl (2.8 MB)
Installing collected packages: py-spy
Successfully installed py-spy-0.4.1

$ cat x.py
for _ in range(1_000_000):
   pass
$ py-spy record --native python x.py 
py-spy> Sampling process 100 times a second. Press Control-C to exit.


py-spy> Stopped sampling because process exited
py-spy> Wrote flamegraph data to 'python-2025-12-12T14:32:29+01:00.svg'. Samples: 2 Errors: 0

$ ll python*svg
-rw-rw-r-- 1 br br  18802 дек 12 14:32 python-2025-12-12T14:32:29+01:00.svg
```

lucascolley:
@ev-br and does this work inside `spin run`?

EDIT: in particular with the script added in this PR

ev-br:
> and does this work inside spin run?

Apparently, yes:

```
$ git co pr/24075
M	subprojects/xsf
Branch 'pr/24075' set up to track remote branch 'pr/24075' from 'upstream'.
Switched to a new branch 'pr/24075'
(scipy-dev) br@gonzales:~/repos/scipy/scipy$ git submodule update --init
Submodule path 'subprojects/xsf': checked out 'dd14bf826356f3a365d4735060b82c943072d4ff'
(scipy-dev) br@gonzales:~/repos/scipy/scipy$ spin run py-spy record --native tools/profiling/bench_tfqmr.py 
$ meson compile -j 4 -C build
Error: Failed to create process 'tools/profiling/bench_tfqmr.py': Permission denied (os error 13)

# argh, forgot `python`
(scipy-dev) br@gonzales:~/repos/scipy/scipy$ spin run py-spy record --native python tools/profiling/bench_tfqmr.py 
$ meson compile -j 4 -C build
py-spy> Sampling process 100 times a second. Press Control-C to exit.


py-spy> Stopped sampling because process exited
py-spy> Wrote flamegraph data to 'python-2025-12-12T15:00:31+01:00.svg'. Samples: 1047 Errors: 0
```

lucascolley:
how frustrating...

```console
scipy on 🎋 profile is 📦 v1.18.0.dev0 via 🐍 v3.14.0 via 🧚 v0.60.0 (profile)
❯ spin run py-spy record --native python tools/profiling/bench_tfqmr.py
$ meson compile -j 6 -C build
Error: Failed to find python version from target process
```

lucascolley:
so something in this function seems to work with mamba but not Pixi: https://github.com/benfred/py-spy/blob/5e14e32f8d43be38cf538cd104531494f508c6d4/src/python_process_info.rs#L296-L364

ev-br:
I'd try to downgrade python first (my local envs above are on 3.12). If that doesn't help, then yes, it's something about pixi

EDIT: https://github.com/benfred/py-spy/blob/5e14e32f8d43be38cf538cd104531494f508c6d4/src/python_interpreters.rs#L15

lucascolley:
oh yeah, py-spy doesn't support py314 yet :melting_face: thanks Evgeni!

lucascolley:
niceeeeee

```console
scipy on 🎋 profile is 📦 v1.18.0.dev0 via 🐍 via 🧚 v0.60.0
❯ pixi run py-spy bench_tfqmr tfqmr_spy
```

image

lucascolley (commented):
added a README. Would love to get this in, looking to use it on other PRs!

lucascolley:
GPU job failure fixed in gh-24119

pixi.toml
```diff
@@ -34,6 +34,17 @@ solve-group = "default"
 features = ["run-deps", "build-deps", "bench"]
 solve-group = "default"
 
+[environments.build-py313]
```
ev-br: Hardcoding python versions looks odd TBH.

I'mm not a pixi user though, so might be missing how things are done in the pixi-land.

pixi.toml
```diff
@@ -568,8 +579,57 @@ depends-on = "build-system-libs"
...
+
+[feature.profile]
+platforms = ["linux-64", "osx-arm64"]
+
+[feature.profile.dependencies]
+py-spy = "*"
+flamegraph-pl = "*"
+
+[feature.profile.pypi-dependencies]
+flamegraph.git = "https://github.com/nickodell/python-flamegraph.git"
```
ev-br: Is this something to move to live under the scipy GH org?

pixi.toml
```diff
@@ -568,8 +579,57 @@ depends-on = "build-system-libs"
...
+
+[feature.profile.tasks.py-spy]
+cmd = "spin run --build-dir=build-py313 'cd tools/profiling && py-spy record --native -o {{ log_name }}.svg -- python {{ script_name }}.py'"
+args = ["script_name", "log_name"]
+depends-on = "build-py313"
+description = "Generate a py-spy flamegraph {{ log_name }}.svg from tools/profiling/{{ script_name }}.py"
+
+[feature.profile.target.osx.tasks.py-spy]
+# `--native` is unsupported on osx, and `sudo` is required
+cmd = "sudo spin run --build-dir=build-py313 'cd tools/profiling && py-spy record -o {{ log_name }}.svg -- python {{ script_name }}.py'"
```
ev-br: Again, I'm very naive re: ways of the pixi, just noting that `sudo` caught me off guard here. I'd expect to be calling `sudo something` explicitly, and would be _very_ worried if some random command starts asking me to type the password.

ev-br: Also `pixi` calling `spin`? Fine.

tools/profiling/README.md
```diff
@@ -0,0 +1,35 @@
+# Flame profiling for SciPy
```
ev-br, line 1: Would be nice to add equivalent `spin` commands here, too.

ev-br (commented):
A couple of comments, with a _grumpy old man_ hat on.

pixi.toml
lucascolley: The problem is py-spy's metadata, it falsely claims to support py314. Indeed this wouldn't be needed if py-spy pulled in py313 by default.

pixi.toml
lucascolley: > I'd expect to be calling sudo something explicitly, and would be very worried if some random command starts asking me to type the password.

Hence the warning in the task description (seen on the CLI when invoked) and README. Maybe this would work as `sudo pixi`, but I figured we are better off scoping sudo as tightly as possible, no?

tools/profiling/README.md
lucascolley: > Would be nice to add equivalent `spin` commands here, too.

'Would be nice' or blocking?

> with a _grumpy old man_ hat on.

With my grumpy hat on, I would rather not work on things that I don't see being useful!

pixi.toml
lucascolley: I opened https://github.com/conda-forge/py-spy-feedstock/issues/39

pixi.toml
rgommers: A little confusing indeed. I've used `py-spy` before and never needed `sudo`. The way it's used here, that should be macOS-specific: https://github.com/benfred/py-spy?tab=readme-ov-file#when-do-you-need-to-run-as-sudo. 

I think this can be moved into `.spin/cmds.py` and have Python code with `if <darwin>` for the invocation, to make it better on Linux.

Creating `spin py-spy` and calling that from this `pixi.toml` seems desirable anyway.

lucascolley: > The way it's used here, that should be macOS-specific

indeed, hence the `target.osx` here

> I think this can be moved into `.spin/cmds.py` and have Python code with `if <darwin>` for the invocation, to make it better on Linux.

what is the story with `sudo` in that case?

rgommers: > indeed, hence the `target.osx` here

Ah yes

> what is the story with `sudo` in that case?

The same I think, when using `subprocess.run("py-spy ...)` on Linux and `subprocess.run("sudo py-spy ...)` on macOS? 

It just aligns with our design to have all dev steps as `spin` commands so they work everywhere given the right environment, and `pixi` being an as-thin-as-possible shell.

lucascolley: > It just aligns with our design to have all dev steps as `spin` commands so they work everywhere given the right environment, and `pixi` being an as-thin-as-possible shell.

Right, I'm fine with that as the end goal, I just don't want to work on it myself until there are compelling use-cases from developers unable to use Pixi. I think my time is better spent enhancing the Pixi workflow. IOW others are welcome to push to this branch if that is blocking.

nickodell:
> Without it, I see errors like `Error: Failed to find python version from target process`.

That's probably caused by YAMA. On Ubuntu, for example, they have this extra kernel security feature where a process is not permitted to ptrace a process unless you have a debugger relationship with that process.

I usually disable this by changing the file  `/etc/sysctl.d/10-ptrace.conf`. This is safer than running everything you intend to trace as root. See docs here: https://www.kernel.org/doc/html/latest/admin-guide/LSM/Yama.html#ptrace-scope

lucascolley:
> That's probably caused by YAMA

sadly it was something more mundane, https://github.com/benfred/py-spy/issues/750#issuecomment-3479913889

ev-br:
>> Would be nice to add equivalent spin commands here, too.

> 'Would be nice' or blocking?

Not blocking. 
I'm not approving this PR though either. It has enough things which look strange to my eyes:
- python version dependent command groups
- elevating permissions inside commands
- I don't immediately see what is the problem it solves w.r.t. user typing the same commands (py-spy does not work on 3.14? OK, use 3.13; OS X requires sudo? OK, type `$ sudo spin ....`; what does `$ spin py-spy ...` command buys w.r.t. `spin run py-spy python ...`?)

FWIW, if pixi / MacOS users (I'm neither) see the need and consider these acceptable, great. I just don't have enough context to make an educated decision.

> until there are compelling use-cases from developers unable to use Pixi. ... others are welcome to push to this branch if that is blocking.

Again FWIW, and not to start a discussion: this sounds a bit divisive. 
It's great that you're excited about pixi, and it's perfectly fine to limit the scope of a PR.
That said, enhancements of general-purpose workflows call for general-purpose solutions IMO, and fragmenting the DX experience, on purpose, feels undesirable. 
My 2p anyway.

lucascolley:
> I don't immediately see what is the problem it solves w.r.t. user typing the same commands (py-spy does not work on 3.14? OK, use 3.13; OS X requires sudo? OK, type $ sudo spin ....; what does $ spin py-spy ... command buys w.r.t. spin run py-spy python ...?)

I can see why you would have these questions as a non-Pixi user, since you are used to the convoluted way of doing things. Precisely the point is that one *shouldn't have to* think about manual management of Python versions when one wants to generate a flamegraph. One *shouldn't have to* remember (or more importantly learn / dig through docs for) how to perform a build with a non-default Python version and point to that build. One *shouldn't have to* remember that the correct invocation is `spin run --build-dir=build-py313 'cd tools/profiling && py-spy record --native -o {{ log_name }}.svg -- python {{ script_name }}.py'`, and one shouldn't have to type that out in full each time.

And for the cherry on top, when I say that I have generated this svg with `pixi run ...`, you should be able to execute the same command on your machine and get similar results. Such a story is *a lot* less nice when we have to accommodate for disparate custom env setups.

These points are already strongly compelling for me, and I can only imagine more-so for a less experienced developer. If you still can't see what problem it solves, I can only suggest putting yourself in the shoes of a less experienced developer, or even better, trying out Pixi yourself!

It's great if people *do* want to learn how to do these things — then they can perform esoteric work that we don't have reason to make reproducible via Pixi, and they can help us extend the Pixi manifest further. They can also choose not to use Pixi at all!

Just keep in mind that the target audience here is not maintainers who are already deeply familiar with building SciPy and have already used tools like py-spy. Rather it is the broader category of 'people who might want to flame-profile some SciPy code'.

ev-br:
> perform a build with a non-default Python version and point to that build.
> ...
> `
spin run --build-dir=build-py313 'cd tools/profiling && py-spy record --native -o {{ log_name }}.svg -- python {{ script_name }}.py'
`

This stanza is completely self-inflicted: No one has to type this _unless_ they want to use incompatible versions of python and py-spy.

So now I understand that one of the goals here is to hide the python version from the user? So if I only have python 3.14 , will your `pixi` invocation silently install a lower version, too?
If so, I believe this is not a good DX. Feel free to chalk me as being an old dog etc :-).

As a practical suggestion, a detailed description would be helpful of what the PR adds, what are alternatives and decisions you make here. While you don't have me sold on it, it might be more helpful to those who will actually decide on the PR.

lucascolley:
> So now I understand that one of the goals here is to hide the python version from the user? So if I only have python 3.14 , will your pixi invocation silently install a lower version, too?
> If so, I believe this is not a good DX. Feel free to chalk me as being an old dog etc :-).

The Python version is not hidden at all, simply `pixi list python` to get the version used in the default  env, or `pixi list --environment=profile python` to get the version used in the environment added here.

The version will also be printed in the meson build log as usual.

> if I only have python 3.14 , will your pixi invocation silently install a lower version, too?

I'm not sure what this means. If you mean "will Pixi silently install another Python version *globally*?", the answer is no: Pixi will install the requested environments as they are detailed in the lock file *locally*, under `.pixi` of your clone.

I assume the first phrase means the situation where you have Python 3.14 installed globally on your system? This is irrelevant to the Pixi setup. One of the greatest strengths of Pixi is that it is not beholden to whatever global setup might be causing an "it works on my machine" moment.

If you instead mean invoking Pixi from within a venv or mamba environment, that is just misuse.

lucascolley:
> This stanza is completely self-inflicted: No one has to type this unless they want to use incompatible versions of python and py-spy.

On the contrary, the hypothetical developer does not even want to think about Python version. How did you come to this conclusion? With `pixi run build`, the developer never even needs to install Python manually!

The conflict comes from the fact that our default build environment uses the latest minor Python release, which seems sensible?

I-Al-Istannen:
Probably not completely relevant, but also note that unless you have a libunwind with https://github.com/libunwind/libunwind/pull/926, you won't be able to get native symbols for astrals standalone python binaries (e.g. obtained from `uv`). If you rely on pixi only it should use python from conda-forge, which doesn't have this problem afaik. But you already seem to have done some symbol debugging so I thought it would be nice to mention one more :)

lucascolley:
if this PR doesn't find a reviewer before Python 3.15 is released, we may want to investigate interfacing with https://docs.python.org/3.15/whatsnew/3.15.html#tachyon-high-frequency-statistical-sampling-profiler instead at that point.

rgommers:
I'm willing to have another look here. One question though: why do we need two? Can we just use `py-spy` - it's much more mature and with more features; installing a v0.1 package straight from GitHub doesn't feel quite right.

lucascolley:
> One question though: why do we need two? Can we just use py-spy

@nickodell thoughts?

nickodell:
> > One question though: why do we need two? Can we just use py-spy
> 
> @nickodell thoughts?

You probably don't need both. python-flamegraph has two advantages:

* Low overhead. Since it does not need to pause and resume the process under test using signals, the performance impact is smaller.
* No use of ptrace / debugging interfaces. Since python-flamegraph is a native thread inside the same process as the thing you're testing, it does not need ptrace. Useful in container set ups.

py-spy probably beats it everywhere else.

 * The flip side of the low-overhead part: because python-flamegraph is a Python thread, it won't run when the GIL is held, which leads to a bias toward recording more time spent in functions which release the GIL.
 * Documentation documentation documentation.
 * py-spy can profile native call frames. python-flamegraph cannot.
 * py-spy can profile processes that it didn't start. python-flamegraph cannot.

I-Al-Istannen:
py-spy has an option to create samples without pausing the process by the way: `--nonblocking`. This increases the error rate as the stack changes while collecting the sample, but might be good enough

pixi.toml
lucascolley: the latest py-spy release added py314 support

tylerjereddy:
I think I'll bump the milestone here since this hasn't been active in a month or two and it is a developer tool thing that can wait 11-12 days until after we branch for the next release.

lucascolley:
I think I'll take a look at replacing this with a PR for Tachyon once Python 3.15 is released

lucascolley:
I've reworked this PR to use the Tachyon sampler available in the stdlib at `profiling.sampling` from Python 3.15. Leaving out macOS for now, due to admin permissions considerations. Tested on native linux-64 and in a linux-aarch64 container, but could do with some more battle testing.

image

Compared to py-spy, we don't get symbols in native frames (just a single `<native>` layer), but we could think about adding py-spy in a future PR for that use-case.

Would love if people could try this out with (e.g.) `pixi run flamegraph bench_tfqmr tfqmr_flame` !

cc @nickodell @ev-br @rgommers @fbourgey @j-bowhay

ev-br:
Tachyon is great! 
Not sure where the balance is in hardcoding specific profiling tech though. Adding more switches/pixie subcommands for various tech (tachyon, py-spy, samply) is probably too much of a whack-a-mole. So I don't have any intelligent suggestions beyond "great if at least one person finds it useful".

ikrommyd:
Sorry for jumping in, I saw this thread at the scientific python discord. If you don't know this already, tachyon's big brother chronon is coming out within 2026 (if you watch the end of the talk at PyCon US: https://youtu.be/f1x4X83CDSA?t=1523) and it will be able to profile native frames and have a lot more cool stuff on top of what tachyon can already do. It would be nice to support that.

lucascolley:
> Sorry for jumping in, I saw this thread at the scientific python discord. If you don't know this already, tachyon's big brother chronon is coming out within 2026 (if you watch the end of the talk at PyCon US: https://youtu.be/f1x4X83CDSA?t=1523) and it will be able to profile native frames and have a lot more cool stuff on top of what tachyon can already do. It would be nice to support that.

oh awesome, thanks @ikrommyd I didn't know about that. Yes, +100 on including that once available!

PR #24076, opened 2025-11-28
BUG: Return zero for out-of-range derivative orders in `NdBSpline`; add C++ guard for excessive derivative `m` in `_deBoor_D`

### LINKED ISSUE #23510: BUG: NdBSpline segfaults for too high derivative orders

czgdp1807:
#### Reference issue

Closes https://github.com/scipy/scipy/issues/23510

**Summary**

This PR introduces two complementary improvements to derivative handling in NdBSpline and the underlying FITPACK-style C++ routine _deBoor_D:

1. **Python-side safeguard** — When users request derivative orders `nu` larger than the maximum supported `(k+1)` for any dimension, `NdBSpline.__call__` now returns a zero array of the correct broadcasted shape instead of raising or producing undefined behaviour.

2. **C++-side validation** — `_deBoor_D` now explicitly checks that the derivative order m does not exceed `k+1`, raising a clear `runtime_error` if violated. This prevents undefined memory usage inside the derivative recursion loop.

Together, these changes make the derivative behaviour consistent, safer, and aligned with standard B-spline mathematical properties.

#### Additional information

@ev-br

### DIFF
--- a/scipy/interpolate/_ndbspline.py
+++ b/scipy/interpolate/_ndbspline.py
@@ -180,6 +180,10 @@ class NdBSpline:
         if xi_shape[-1] != ndim:
             raise ValueError(f"Shapes: xi.shape={xi_shape} and ndim={ndim}")
 
+        if any(nu > np.asarray(self.k, dtype=np.int64) + 1):
+            return np.zeros(xi_shape[:-1] + self._c.shape[ndim:],
+                            dtype=self._c.dtype)
+
         # complex -> double
         was_complex = self._c.dtype.kind == 'c'
         cc = self._c
--- a/scipy/interpolate/_ndbspline.py
+++ b/scipy/interpolate/_ndbspline.py
@@ -180,10 +180,6 @@ class NdBSpline:
         if xi_shape[-1] != ndim:
             raise ValueError(f"Shapes: xi.shape={xi_shape} and ndim={ndim}")
 
-        if any(nu > np.asarray(self.k, dtype=np.int64) + 1):
-            return np.zeros(xi_shape[:-1] + self._c.shape[ndim:],
-                            dtype=self._c.dtype)
-
         # complex -> double
         was_complex = self._c.dtype.kind == 'c'
         cc = self._c
--- a/scipy/interpolate/src/__fitpack.cc
+++ b/scipy/interpolate/src/__fitpack.cc
@@ -15,18 +15,6 @@ namespace fitpack{
 void
 _deBoor_D(const double *t, double x, int k, int ell, int m, double *result) {
     /*
-     * If m > k+1, the B-spline derivative is identically zero.
-     * Avoid the de Boor recursion (which assumes m <= k+1) and
-     * return a zero-filled result directly.
-     */
-    if( m > k + 1 ) {
-        for( size_t i = 0; i < 2*k + 2; i++ ) {
-            result[i] = 0.0;
-        }
-        return ;
-    }
-
-    /*
      * On completion the result array stores
      * the k+1 non-zero values of beta^(m)_i,k(x):  for i=ell, ell-1, ell-2, ell-k.
      * Where t[ell] <= x < t[ell+1].
@@ -1217,7 +1205,18 @@ _evaluate_spline(
         else {
             // Evaluate (k+1) b-splines which are non-zero on the interval.
             // on return, first k+1 elements of work are B_{m-k},..., B_{m}
-            _deBoor_D(t.data, xval, k, interval, nu, wrk);
+            /*
+             * If nu > k+1, the B-spline derivative is identically zero.
+             * Avoid the de Boor recursion (which assumes m <= k+1) and
+             * return a zero-filled result directly.
+             */
+            if( nu > k + 1 ) {
+                for( size_t i = 0; i < 2*k + 2; i++ ) {
+                    wrk[i] = 0.0;
+                }
+            } else {
+                _deBoor_D(t.data, xval, k, interval, nu, wrk);
+            }
 
             // Form linear combinations
             for (int64_t jp=0; jp < m; jp++) {
@@ -1387,7 +1386,18 @@ _evaluate_ndbspline(const double *xi_ptr, int64_t npts, int64_t ndim,  // xi, sh
             }
 
             // compute non-zero b-splines at this value of xd in dimension d
-            _deBoor_D(td, xd, kd, i_d, nu(d), wrk.data());
+            /*
+             * If nu(d) > kd+1, the B-spline derivative is identically zero.
+             * Avoid the de Boor recursion (which assumes m <= k+1) and
+             * return a zero-filled result directly.
+             */
+            if( nu(d) > kd + 1 ) {
+                for( size_t i = 0; i < wrk.size(); i++ ) {
+                    wrk[i] = 0.0;
+                }
+            } else {
+                _deBoor_D(td, xd, kd, i_d, nu(d), wrk.data());
+            }
 
             for (int s=0; s < kd + 1; s++) {
                 b(d, s) = wrk[s];
--- a/scipy/interpolate/src/__fitpack.cc
+++ b/scipy/interpolate/src/__fitpack.cc
@@ -55,6 +55,12 @@ _deBoor_D(const double *t, double x, int k, int ell, int m, double *result) {
      * Now do m "derivative" recursions
      * to convert the values of beta into the mth derivative
      */
+    if (m > k + 1) {
+        throw std::runtime_error(
+            "Requested derivative order m exceeds the maximum allowed (k+1). "
+            "Cannot compute derivative of order m for spline of order k."
+        );
+    }
     for (j = k - m + 1; j <= k; j++) {
         memcpy(hh, h, j*sizeof(double));
         h[0] = 0.0;
--- a/scipy/interpolate/src/__fitpack.cc
+++ b/scipy/interpolate/src/__fitpack.cc
@@ -55,12 +55,6 @@ _deBoor_D(const double *t, double x, int k, int ell, int m, double *result) {
      * Now do m "derivative" recursions
      * to convert the values of beta into the mth derivative
      */
-    if (m > k + 1) {
-        throw std::runtime_error(
-            "Requested derivative order m exceeds the maximum allowed (k+1). "
-            "Cannot compute derivative of order m for spline of order k."
-        );
-    }
     for (j = k - m + 1; j <= k; j++) {
         memcpy(hh, h, j*sizeof(double));
         h[0] = 0.0;
@@ -1211,6 +1205,12 @@ _evaluate_spline(
         else {
             // Evaluate (k+1) b-splines which are non-zero on the interval.
             // on return, first k+1 elements of work are B_{m-k},..., B_{m}
+            if (nu > k + 1) {
+                throw std::runtime_error(
+                    "Requested derivative order nu exceeds the maximum allowed (k+1). "
+                    "Cannot compute derivative of order nu for spline of order k."
+                );
+            }
             _deBoor_D(t.data, xval, k, interval, nu, wrk);
 
             // Form linear combinations
@@ -1381,6 +1381,12 @@ _evaluate_ndbspline(const double *xi_ptr, int64_t npts, int64_t ndim,  // xi, sh
             }
 
             // compute non-zero b-splines at this value of xd in dimension d
+            if (nu(d) > kd + 1) {
+                throw std::runtime_error(
+                    "Requested derivative order nu(d) exceeds the maximum allowed (k+1). "
+                    "Cannot compute derivative of order nu(d) for spline of order k."
+                );
+            }
             _deBoor_D(td, xd, kd, i_d, nu(d), wrk.data());
 
             for (int s=0; s < kd + 1; s++) {
--- a/scipy/interpolate/src/__fitpack.cc
+++ b/scipy/interpolate/src/__fitpack.cc
@@ -1191,6 +1191,20 @@ _evaluate_spline(
     auto xp = ConstRealArray1D(xp_ptr, s);
     auto out = RealArray2D(out_ptr, s, m);
 
+    /*
+     * If nu > k+1, the B-spline derivative is identically zero.
+     * Avoid the de Boor recursion (which assumes nu <= k+1) and
+     * return a zero-filled result directly.
+     */
+    if( nu > k + 1 ) {
+        for( int64_t i = 0; i < s; i++ ) {
+            for( int64_t j = 0; j < m; j++ ) {
+                out(i, j) = 0.0;
+            }
+        }
+        return ;
+    }
+
     int64_t interval = k;
     for(int64_t ip=0; ip < s; ip++) {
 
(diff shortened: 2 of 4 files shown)


scipy/interpolate/src/__fitpack.cc
```diff
@@ -55,6 +55,12 @@ _deBoor_D(const double *t, double x, int k, int ell, int m, double *result) {
      * Now do m "derivative" recursions
      * to convert the values of beta into the mth derivative
      */
+    if (m > k + 1) {
+        throw std::runtime_error(
+            "Requested derivative order m exceeds the maximum allowed (k+1). "
+            "Cannot compute derivative of order m for spline of order k."
+        );
+    }
```
ev-br: While raising here is correct, this effectively puts the check inside a hot loop over the data points: typically we call evaluations for arrays of data points.
Thus, could you please remove the `if` check from `_deBoor_D`, and add it to call sites?
Grepping shows three relevant places:

```
$ git grep -n deBoor_D
scipy/interpolate/src/__fitpack.cc:16:_deBoor_D(const double *t, double x, int k, int ell, int m, double *result) {
scipy/interpolate/src/__fitpack.cc:166:        _deBoor_D(t.data, xval, k, ind, 0, wrk);
scipy/interpolate/src/__fitpack.cc:253:        _deBoor_D(t.data, xval, k, ind, 0, wrk);
scipy/interpolate/src/__fitpack.cc:858:        _deBoor_D(t.data, xval, k, ind, 0, wrk.data());
scipy/interpolate/src/__fitpack.cc:1208:            _deBoor_D(t.data, xval, k, interval, nu, wrk);    # evaluate_spline
scipy/interpolate/src/__fitpack.cc:1246:        _deBoor_D(t.data, xval, k, left, 0, wrk);
scipy/interpolate/src/__fitpack.cc:1291:        _deBoor_D(t.data, xval, k, left, 0, wrk);
scipy/interpolate/src/__fitpack.cc:1378:            _deBoor_D(td, xd, kd, i_d, nu(d), wrk.data());   # evaluate_ndbspline
scipy/interpolate/src/__fitpack.cc:1498:            _deBoor_D(td, xd, kd, i_d, 0, wrk.data());
scipy/interpolate/src/__fitpack.h:116:_deBoor_D(const double *t, double x, int k, int ell, int m, double *result);
scipy/interpolate/src/_dierckxmodule.cc:1049:    fitpack::_deBoor_D(        # eval_all_bspl
```

The last one, `eval_all_bspline` is special in that it's directly user facing, so it should just return the correct answer straight from C instead of raising an error.

ev-br (changes requested):

scipy/interpolate/src/__fitpack.cc
czgdp1807: > The last one, `eval_all_bspline` is special in that it's directly user facing, so it should just return the correct answer straight from C instead of raising an error.

I have added the check here as well. It is not supposed to raise error but if it does then it means the input itself was wrong. Should we change the `std::runtime_error` to `ValueError` in this case because it is directly user facing? What do you say?

ev-br: >  Should we change the std::runtime_error to ValueError in this case because it is directly user facing? 

An uncaught C++ exception will hard crash the interpreter. Therefore, in 99.9999% cases, you catch a C++ exception and convert it to a Python error---this is similar to what cython's `except+` does:

```
try {
     .....
}
catch (std::runtime_exception& e) {
   PyErr_SetString(PyExc_RuntimeError, e.what())
}
catch (std::invalid_argument& e) {
   PyErr_SetString(PyExc_ValueError, e.what())
}
catch (std::exception& e) {
   PyErr_SetString(PyExc_RuntimeError, e.what()
}
// the length of the chain depends on how granular you need to be with Python exceptions
```

This however should be used sparingly. Only use it
- if C++ forces it on you, some C++ construction can throw
- something impossible happened. you covered all bases you could think of, but an internal invariant in your code has been violated, and you want to ask a user to file a bug report for you. Then the message should include "please report it to SciPy developers" or some such.

---------------------

For a user attempting to compute a derivative of the order `m > k+1`, I think the best course of action is to return zeros _before_ doing computations. If the check is far away from the computation, then the second reason might apply. As in, "this really should have been handled earlier, and this function should not have seen this value of m; therefore something is wrong with the code flow and I as a scipy developer need to fix it".

czgdp1807: > Therefore, in 99.9999% cases, you catch a C++ exception and convert it to a Python error

~I don't think we need that here. I am already raising the error before the call to `_deBoor_D`. However, instead of C++ `std::runtime_error`, I would just raise a `Py_ValueError` and return `NULL`.~

Well the other way is to set the return value to an array of `0s`, instead of raising an error.

> should just return the correct answer straight from C instead of raising an error.

**Seems like you wanted to say the same thing in https://github.com/scipy/scipy/pull/24076#discussion_r2573566552. I missed the last line in your comment. Apologies.**

scipy/interpolate/tests/test_bsplines.py
```diff
@@ -2172,6 +2172,14 @@ def test_fpback(self):
 
         xp_assert_close(cc, c, atol=1e-14)
 
+    def test_evaluate_all_bspl(self):
+        n = 10
+        x, _, t, k = self._get_xyt(n)
+        for xval in x:
+            assert (_dierckx.evaluate_all_bspl(t, k, xval, n, k + 2) == 0.0).all()
+        for xval in x:
+            assert (_dierckx.evaluate_all_bspl(t, k, xval, n, 2*k) == 0.0).all()
```
j-bowhay: Nit, these loops can be combined and we should use `xp_assert_close` for better error reporting

czgdp1807: Yes. Correct. Here exact `0.0` should be there because requested derivative order is larger than `degree + 1`. That's why I used exact `==` to make sure, the special case is handled correctly.

j-bowhay: Sorry I meant `xp_assert_equal`

czgdp1807:
This PR is ready to go.

czgdp1807:
@rgommers https://github.com/scipy/scipy/actions/runs/19830862191/job/56816306975?pr=24076 is failing consistently. This is `Windows tests (MSVC + ifx + OpenBLAS) / py3.11`. `conda` is giving HTTP error on this job, otherwise it's working fine with every other job. Should I try fixing this (initial idea may be use, `pip` instead of `conda` on Windows?)

scipy/interpolate/src/__fitpack.cc
```diff
@@ -1205,6 +1205,12 @@ _evaluate_spline(
         else {
             // Evaluate (k+1) b-splines which are non-zero on the interval.
             // on return, first k+1 elements of work are B_{m-k},..., B_{m}
+            if (nu > k + 1) {
+                throw std::runtime_error(
+                    "Requested derivative order nu exceeds the maximum allowed (k+1). "
+                    "Cannot compute derivative of order nu for spline of order k."
+                );
+            }
```
ev-br: This is not desired on two counts:

1. If you evaluate a spline on an array of 100500 points, this runs the check 100500 times, instead of just once.

2. Evaluating a too-high derivative of a cubic spline should return zeros, not give a RuntimeError:

```
In [1]: from scipy.interpolate import make_interp_spline

In [2]: import numpy as np

In [3]: spl = make_interp_spline(np.arange(8), np.arange(8))

In [4]: spl.k
Out[4]: 3

In [5]: spl(1)
Out[5]: array(1.)

In [6]: spl(1, nu=8)
RuntimeError                              Traceback (most recent call last)
Cell In[6], line 1
----> 1 spl(1, nu=8)

File ~/repos/scipy/scipy/build-install/usr/lib/python3.12/site-packages/scipy/interpolate/_bsplines.py:564, in BSpline.__call__(self, x, nu, extrapolate)
    561 cc = cc.reshape(cc.shape[0], -1)
    563 # heavy lifting: actually perform the evaluations
--> 564 out = _dierckx.evaluate_spline(self._t, cc, self.k, x, nu, extrapolate)
    566 if is_complex:
    567     out = out.view(complex)

RuntimeError: Requested derivative order nu exceeds the maximum allowed (k+1). Cannot compute derivative of order nu for spline of order k.

```

ev-br (changes requested):
Please move the new checks out of hot loops, and make it return zeros instead of raising an error.

It looks like the whole `std::throw runtime_error` does not bring much here, let's remove it.

scipy/interpolate/src/__fitpack.cc
czgdp1807: > It looks like the whole `std::throw runtime_error` does not bring much here, let's remove it.

Well yes. My preliminary idea was simple, just to not run the following loop at all,

https://github.com/scipy/scipy/blob/7d117511aded88e49ee22a8d4a54adf439bafe51/scipy/interpolate/src/__fitpack.cc#L58-L73

Instead set the full `result` to `0.0` and return from `_deBoor_D`.

Setting an array to 0.0s in C is simple.

scipy/interpolate/_ndbspline.py
```diff
@@ -180,6 +180,10 @@ def __call__(self, xi, *, nu=None, extrapolate=None):
         if xi_shape[-1] != ndim:
             raise ValueError(f"Shapes: xi.shape={xi_shape} and ndim={ndim}")
 
+        if any(nu > np.asarray(self.k, dtype=np.int64) + 1):
+            return np.zeros(xi_shape[:-1] + self._c.shape[ndim:],
+                            dtype=self._c.dtype)
+
```
czgdp1807: This change is not needed now, after I updated `_deBoor_D` to return `0`s when `m > k + 1`. See the diff of `scipy/interpolate/src/__fitpack.cc`. We can remove this if @ev-br agrees.

ev-br: This still has the O(num_data_points) cost. Lets' please leave `_deBoor_D` as it is in main, and guard the call sites with `m>0`.

czgdp1807: Sure. Will do once I come back. Will remove this Pythonic change as well. That should do the job I hope.

rgommers:
> @rgommers https://github.com/scipy/scipy/actions/runs/19830862191/job/56816306975?pr=24076 is failing consistently.

Nothing to do there it looks like, that was just a temporary hiccup - the latest runs passed.

Regarding the "install into venv" failure: I recommend you build locally with `-Werror` (e.g., append `-Werror` to CFLAGS/CXXFLAGS), or pass `--werror` to `meson setup`).

scipy/interpolate/src/__fitpack.cc
ev-br: The `if (nu > k + 1)` check is still performed for each data point. So the item 1. above is not resolved yet.

ev-br (changes requested):

scipy/interpolate/src/__fitpack.cc
czgdp1807: I see your point. There are already two loops in `_deBoor_D` which always execute and the condition check in them is executed in each iteration. Specifically the first loop executes for, `k - nu` times and the second loop executes for `nu` times. So overall, `k` steps. The check, `if (nu > k + 1)` adds `1` to it, so `k + 1` if the check would fail and `1` if the check would pass. So an overhead of `m`. I see, while typing this comment I got it.

https://github.com/scipy/scipy/blob/4d6c4ac61322a053bab930475ec18c2993e66ed4/scipy/interpolate/src/__fitpack.cc#L12-L74

To fix this, for `_evaluate_spline` I will keep the check at the entry point,

https://github.com/scipy/scipy/blob/4d6c4ac61322a053bab930475ec18c2993e66ed4/scipy/interpolate/src/__fitpack.cc#L1192-L1194

For, `py_evaluate_all_bspl`, there is no loop and since its user facing so we can't do anything there other than returning 0s when `if (nu > k + 1)`.

For, `_evaluate_ndbspline`, we can set `out` to 0s for all the data points if any of `nu(d)` is greater than `kd + 1`. We won't run the loop over the data points,

https://github.com/scipy/scipy/blob/4d6c4ac61322a053bab930475ec18c2993e66ed4/scipy/interpolate/src/__fitpack.cc#L1341-L1343

This should avoid the `O(num_data_points)` overhead wherever possible.

czgdp1807:
Addressed as per https://github.com/scipy/scipy/pull/24076#discussion_r2583793946

ev-br (approved):
LGTM now, thanks @czgdp1807

czgdp1807:
Thanks for your patience @ev-br. There was a bit of confusion involved in this PR. Earlier, we started off with raising errors. Then we went for returning 0s straight from C. And then I was lagging behind in figuring out the reasoning behind extra `O(num_data_points)` (which you already pointed out before I typed https://github.com/scipy/scipy/pull/24076#discussion_r2583793946). Anyways all good now. The right solution went in. <3.

PR #24079, opened 2025-11-29
ENH: migrate `special.bdtrin` to boost

dschmitz89:
#### Reference issue
Towards #21966 and #7168

Replace cdflib's bdtrin by boost's `binomial_distribution<>::find_minimum_number_of_trials`.

#### Additional information
**Benchmark**
bdtrin_benchmark

<details><summary>Plot generation</summary>
<p>

```python
from mpmath import mp
import numpy as np


# mp.dps = 30  # set decimal places for mpmath

# @np.vectorize
# def _binomial_cdf(k, n, p):
#     k, n, p = mp.mpf(k), mp.mpf(n), mp.mpf(p)
#     if k <= 0:
#         return mp.mpf(0)
#     elif k >= n:
#         return mp.mpf(1)

#     onemp = mp.fsub(1, p, exact=True)
#     return float(mp.betainc(n - k, k + 1, x2=onemp, regularized=True))


# rng = np.random.default_rng(123456876475)

# number_of_samples = 10000
# # Generate n values (e.g., between 2 and 100)
# n = rng.integers(2, 3000, size=number_of_samples)
# # For each n, generate k in [0, n-1]
# k = np.array([rng.integers(0, ni) for ni in n])
# n = n.astype(np.float64)
# k = k.astype(np.float64)
# p = rng.uniform(0, 1, size=number_of_samples)


# # Compute reference values using mpmath
# reference_values = _binomial_cdf(k, n, p)

# mask = (reference_values > 1e-120) & (reference_values < 0.999999)
# # Compute bdtrin values using scipy
# bdtrin_values = sc.bdtrin(k, reference_values, p)
# # Calculate relative errors
# relative_errors = np.abs(bdtrin_values - n) / np.abs(n)
# relative_errors = relative_errors[mask]
# relative_errors = np.clip(relative_errors, np.finfo(float).eps, 1)

#np.save('bench_bdtrin_rel_error_pr.npy', relative_errors)

pr = np.load('bench_bdtrin_rel_error_pr.npy')
main = np.load('bench_bdtrin_rel_error_main.npy')

print(np.all(pr == main))
plt.hist(np.log10(pr), bins=int(np.sqrt(len(pr))), density=True, cumulative=True, alpha=0.5, label="PR")
plt.hist(np.log10(main), bins=int(np.sqrt(len(main))), density=True, cumulative=True, alpha=0.5, label="main")
plt.title("Cumulative Distribution of Relative Errors in bdtrin")
plt.legend()
plt.show()
``` 

</p>
</details>


scipy/special/functions.json
```diff
@@ -1021,4 +1023,4 @@
             "hypergeom_skewness_double": "ddd->d"
         }
     }
-}
+}
```
dschmitz89:
```suggestion
}

```

Issue #18703, opened 2023-06-19
DEP: deprecate positional arguments for signatures being changed in 1.12 cycle


h-vetinari:
Following https://docs.scipy.org/doc/scipy/dev/missing-bits.html#required-keyword-names, we should use the removal of keywords as an opportunity to turn relevant arguments in the signature into keyword-only arguments.

This needs a separate deprecation from the one that just warns about removing a given kwarg (although if the kwarg being deprecated is early in the signature, it might end up being equivalent), to warn users that feed kwargs positionally that they need to pass them as keywords going forward.

The mechanism for emitting a good deprecation warning on this is not so straight-forward, but scikit-learn came up with a good [helper](https://github.com/scikit-learn/scikit-learn/blob/1.2.2/sklearn/utils/validation.py#L42) for their own effort in this direction that we can vendor here.

This is aimed particularly at the functions being touched in https://github.com/scipy/scipy/pull/18624, but is in force for all function signatures where kwargs are being scheduled for removal (or otherwise changed).

More context: https://github.com/scipy/scipy/issues/14714

tupui:
I am all for this initiative 👍 Thank you for the proposal.

h-vetinari:
For consistency of the functions in `scipy.signal`, do we want to extend the deprecation of positional use of kwargs to all windows? #18932 did `firls, firwin, firwin2, remez`, remaining are:
```
freqz, kaiser_beta, kaiser_atten, kaiserord, minimum_phase
```

This is the same approach we chose for the sparse methods as well (split between #18934 & #18943). WDYT @tupui @j-bowhay? CC @larsoner @ilayn as code owners.

larsoner:
Hmm... looking at just one example https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.minimum_phase.html it seems like the `method` shouldn't be required to be kwarg-only, I think for these make sense:
```
minimum_phase(h, 'homomorphic')
minimum_phase(h, 'hilbert')
```
Maybe `n_fft` (which follows the `method='homomorphic'`) should be required.

I'd be slightly inclined to take a case-by-case basis for these functions, but if that wasn't how the others have been done or it would create an ugly within-(sub)module inconsistency to go case-by-case I think it's okay to deprecate positional use.

h-vetinari:
Thanks for the comment, that's a fair point of course, and all (families of) functions are different in terms of specifics and especially consistency (intra-module and with other related functions).

[...]

PR #24080, opened 2025-11-29
DEP: signal.lombscargle: deprecate positional args, warn on `precenter=False`

lucascolley:
Towards gh-18703, follow-up to gh-23476.

- take the opportunity to deprecate passing `normalize` positionally
- warn also on `precenter=False`, not just `precenter=True`

FYI @adammj @DietBru @tylerjereddy

### DIFF
--- a/scipy/signal/_spectral_py.py
+++ b/scipy/signal/_spectral_py.py
@@ -3,6 +3,7 @@
 import numpy as np
 import numpy.typing as npt
 from scipy import fft as sp_fft
+from scipy._lib.deprecation import _deprecate_positional_args
 from . import _signaltools
 from ._short_time_fft import ShortTimeFFT, FFT_MODE_TYPE
 from .windows import get_window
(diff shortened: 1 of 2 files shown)


scipy/signal/_spectral_py.py
```diff
@@ -3,6 +3,7 @@
 import numpy as np
 import numpy.typing as npt
 from scipy import fft as sp_fft
+from scipy._lib.deprecation import _deprecate_positional_args
```
j-bowhay:
```suggestion
from scipy._lib.deprecation import _deprecate_positional_args, _NoValue
```

scipy/signal/_spectral_py.py
```diff
@@ -15,13 +16,14 @@
...
 
+@_deprecate_positional_args(version="1.19.0")
 def lombscargle(
     x: npt.ArrayLike,
     y: npt.ArrayLike,
     freqs: npt.ArrayLike,
-    precenter: bool = False,
-    normalize: bool | Literal["power", "normalize", "amplitude"] = False,
     *,
+    precenter: bool | None = None,
```
j-bowhay:
```suggestion
    precenter: bool = _NoValue,
```

scipy/signal/_spectral_py.py
```diff
@@ -255,13 +257,15 @@ def lombscargle(
     weights = weights * (1.0 / weights.sum())
 
     # if requested, perform precenter
-    if precenter:
+    if precenter is not None:
```
j-bowhay:
```suggestion
    if precenter is not _NoValue:
```

j-bowhay (commented):
I think this is ever so slightly more bulletproof

adammj:
Good catch on precenter=False

tylerjereddy (approved):
Looks a bit better yeah, thanks, I'll squash merge I think.

PR #24081, opened 2025-11-29
DEV: auto apply `needs-release-note-decision` label to ENH/DEP PRs

### LINKED ISSUE #24051: RFC: Should we have a 'needs-release-note-decision' tag?
### Issue with current approach:

One thing that I think could be improved about our current release notes process is to improve the coverage of changes which get release notes.

Right now, we use the 'needs-release-notes' tag to indicate changes which ought to be mentioned in the release notes, but have not yet been written.

This system has two problems:

1. It is hard to tell the difference between two kinds of PRs: one where a maintainer has actively thought about the question of whether the PR ought to get a release note, and decided not to; and a second kind where no maintainer has thought about it. It is hard to understand the intent behind a lack of action.
2. Some uses of this tag are on open PRs that are not likely to be merged in this release. I interpret this to mean, "if this PR were to be merged, it would need a release note." However, we don't need to make this decision until we decide we are actually including the code. A PR may require substantial revision to provide backward compatibility, and this may affect whether it needs a release note.

### Idea

I propose three changes:

1. Introduce a new tag, called 'needs-release-note-decision'. When a PR is tagged with this tag, a maintainer should do one of three things: 
   1. Remove the tag, as no release note needs to be written.
   2. Remove the tag, and apply 'needs-release-note,' because the change needs a release note.
   3. Remove the tag, and write a release note.
2. Write a GH action which automatically applies 'needs-release-note-decision' to every PR which is merged at the time of merge. I believe this can be done using the `actions/labeller` action in response to the `pull_request_target.closed` event.
3. Remove the tag 'needs-release-note' from every PR which is open. This would become a tag that is only used for merged PRs.

### Justification for change

This would improve the current process in the following ways:

 * When a maintainer merges a PR, and it clearly needs no decision, they can remove the automatically applied 'needs-release-note-decision' tag. This is a small amount of work which improves the process of writing the release notes, as the next maintainer to look at this can see the tag being removed, and know that this PR intentionally has no release notes.
 * A maintainer looking for PRs which ought to have 'needs-release-note' can find them by looking at the 'needs-release-note-decision' to find PRs where no decision has been made.

See also [Slack discussion](https://scipy-community.slack.com/archives/C03JL2YGCNR/p1763698679701519)

nickodell:
#### Reference issue

Fixes #24051

In order to get better coverage of release notes, and to keep track of which PRs we have made an explicit decision on whether to write a release note, apply the 'needs-release-note-decision' whenever a PR with a title matching ENH or DEP is merged. 

#### Testing notes

I have tested the following scenarios:

1. Merging a PR causes the label to be applied
2. Closing a PR without merging does not cause the label to be applied.
3. This action does not run in response to issues being closed.

#### Additional information

1. This action lacks the permissions to create labels. Therefore, the label must exist before the action can apply the label. If this label does not exist, this action will not fail, but it will not apply the 'needs-release-note-decision' label either. New labels can be created at https://github.com/scipy/scipy/labels.

Proposed text for new label:

>Label name: needs-release-note-decision
>
>Mouseover text: A maintainer should decide if this change requires a release note.

2. zizmor analysis of the new action file:

```
🌈 zizmor v1.17.0
 INFO audit: zizmor: 🌈 completed .github/workflows/label-pr-on-close.yml
error[dangerous-triggers]: use of fundamentally insecure workflow trigger
 --> .github/workflows/label-pr-on-close.yml:2:1
  |
2 | / on:
3 | |   pull_request_target:
4 | |     types: [closed]
  | |___________________^ pull_request_target is almost always used insecurely
  |
  = note: audit confidence → Medium

4 findings (3 suppressed): 0 informational, 0 low, 0 medium, 1 high
```

I tried replacing `pull_request_target` with `pull_request` as a trigger, but could not get this to work.


lucascolley (approved):
thanks Nick, looks great to me. Let's hold off on merging this until Tyler is finished with the 1.17.0 release

nickodell:
> thanks Nick, looks great to me. Let's hold off on merging this until Tyler is finished with the 1.17.0 release

Makes sense. We should avoid changing the process too much in the middle of a release.

nickodell:
> thanks Nick, looks great to me. Let's hold off on merging this until Tyler is finished with the 1.17.0 release

Now that 1.17.0 is out, can this be merged?

nickodell:
I was trying to figure out why github actions applied the "needs-release-note-decision" label here, when it is only supposed to do that for ENH / DEP PRs. Then I realized: I put ENH/DEP in the title. 😆

PR #24083, opened 2025-11-30
ENH: stats.f_oneway: add array API support

mdhaber:
#### Reference issue
Toward gh-20544

Adds array API support to `scipy.stats.f_oneway`.

#### Additional information
Commits are pretty clean and worth reviewing separately.

### DIFF
--- a/scipy/stats/_stats_py.py
+++ b/scipy/stats/_stats_py.py
@@ -71,6 +71,7 @@ from scipy._lib._array_api import (
     is_dask,
     is_numpy,
     is_cupy,
+    is_marray,
     xp_size,
     xp_vector_norm,
     xp_promote,
@@ -3903,7 +3904,7 @@ def f_oneway(*samples, axis=0, equal_var=True):
 
     # axis is guaranteed to be -1 by the _axis_nan_policy decorator
     alldata = xp.concat(samples, axis=-1)
-    bign = alldata.shape[-1]
+    bign = _length_nonmasked(alldata, axis=-1, xp=xp)
 
     # Check if the inputs are too small (for testing _axis_nan_policy decorator)
     if _f_oneway_is_too_small(samples):
@@ -3919,7 +3920,7 @@ def f_oneway(*samples, axis=0, equal_var=True):
     # It is True if the values within the groups along the axis slice are
     # identical. In the typical case where each input array is 1-d, is_const is
     # a 1-d array with length num_groups.
-    is_const = xp.concat([xp.all(sample[..., :1] == sample, axis=-1, keepdims=True)
+    is_const = xp.concat([xp.all(xp.diff(sample, axis=-1) == 0, axis=-1, keepdims=True)
                           for sample in samples], axis=-1)
 
     # all_const is a boolean array with shape (...) (see previous comment).
@@ -3929,7 +3930,7 @@ def f_oneway(*samples, axis=0, equal_var=True):
 
     # all_same_const is True if all the values in the groups along the axis=0
     # slice are the same (e.g. [[3, 3, 3], [3, 3, 3, 3], [3, 3, 3]]).
-    all_same_const = xp.all(alldata[..., :1] == alldata, axis=-1)
+    all_same_const = xp.all(xp.diff(alldata, axis=-1) == 0, axis=-1)
 
     if not isinstance(equal_var, bool):
         raise TypeError("Expected a boolean value for 'equal_var'")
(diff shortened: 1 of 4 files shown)


scipy/stats/_stats_py.py
```diff
@@ -3937,25 +3919,17 @@ def f_oneway(*samples, axis=0, equal_var=True):
...
+                          for sample in samples], axis=-1)
 
     # all_const is a boolean array with shape (...) (see previous comment).
     # It is True if the values within each group along the axis slice are
     # the same (e.g. [[3, 3, 3], [5, 5, 5, 5], [4, 4, 4]]).
-    all_const = is_const.all(axis=axis)
-    if all_const.any():
-        msg = ("Each of the input arrays is constant; "
-               "the F statistic is not defined or infinite")
-        warnings.warn(stats.ConstantInputWarning(msg), stacklevel=2)
```
mdhaber, line 3954: Avoiding backend-dependent warnings. Code produces a correct result (`nan` for undefined and `inf` for infinite); no need to add noise.

scipy/stats/_stats_py.py
```diff
@@ -3765,7 +3746,7 @@ def _f_oneway_is_too_small(samples, kwargs=None, axis=-1):
     return False
 
 
-@xp_capabilities(np_only=True)
+@xp_capabilities(jax_jit=False)
```
mdhaber: To fix CI failure:
```suggestion
@xp_capabilities(jax_jit=False, cpu_only=True, exceptions=['cupy'])
```
`fdtrc` is not native to all backends.

I'll wait for other suggestions before committing.

scipy/stats/_stats_py.py
```diff
@@ -3965,17 +3940,17 @@ def f_oneway(*samples, axis=0, equal_var=True):
...
         # improves numerical stability.
-        offset = alldata.mean(axis=axis, keepdims=True)
+        offset = xp.mean(alldata, axis=-1, keepdims=True)
         alldata = alldata - offset
 
-        normalized_ss = _square_of_sums(alldata, axis=axis) / bign
+        normalized_ss = xp.sum(alldata, axis=-1)**2. / bign
 
-        sstot = _sum_of_squares(alldata, axis=axis) - normalized_ss
+        sstot = xp.vecdot(alldata, alldata, axis=-1) - normalized_ss
```
j-bowhay, line 3948: Not that it really matter but is there any reason for doing `xp.sum(alldata, axis=-1)**2` in some place but `xp.vecdot(alldata, alldata, axis=-1)` in others?

scipy/stats/_stats_py.py
```diff
@@ -3987,38 +3962,43 @@ def f_oneway(*samples, axis=0, equal_var=True):
...
         # "As a particular case $y_t$ may be the means ... of samples
-        y_t = np.asarray([np.mean(sample, axis=axis) for sample in samples])
+        y_t = xp.stack([xp.mean(sample, axis=-1) for sample in samples])
         # "... of $n_t$ observations..."
-        n_t = np.asarray([sample.shape[axis] for sample in samples])
-        n_t = np.reshape(n_t, (-1,) + (1,) * (y_t.ndim - 1))
+        if is_marray(xp):
+            n_t = xp.stack([_length_nonmasked(sample, axis=-1, xp=xp)
+                            for sample in samples])
+            n_t = xp.asarray(n_t, dtype=n_t.dtype)
```
j-bowhay, line 3976: How come we don't require the same reshaping (3979) of `n_t` for masked arrays

scipy/stats/_stats_py.py
mdhaber: The former is the square of a sum, and the latter is the sum of squares.

j-bowhay: Oops yes indeed, too many "s"s for me in  `_square_of_sums` and `_sum_of_squares`!

scipy/stats/_stats_py.py
mdhaber: Because `_length_nonmasked` already returns an array representing the length of each slice - it already has the dimensions that are appended in 3979.

j-bowhay: That makes sense, thanks. Is `n_t = xp.asarray(n_t, dtype=n_t.dtype)` necessary? I thought `xp.stack` would already be an array?

mdhaber: The other PRs involving hypothesis tests with an arbitrary number of samples have avoided this sort of thing by keeping the statistics taken on each sample as a list (not an array) and using built-In Python `sum` and such rather than array library `xp.sum` along axis `0`. This may be faster when there is only a handful of samples. It was going to be a bit more complex to refactor like that here, so I didn't think I should do that, but it might be something to consider for a follow-up.

mdhaber: Re: https://github.com/scipy/scipy/pull/24083#discussion_r2581362293, the dtype conversion is necessary to appease `array-api-strict`, and also unlike most (all?) real libraries, `array-api-strict` doesn't allow `dtype` to be passed to `stack`.

j-bowhay (approved):
LGTM

PR #24085, opened 2025-11-30
DOC: linalg: edit the doc note: we no longer import numpy linalg func…

ev-br:
…tions

Maybe we did before, but we do not anymore.

[doc only]


mdhaber:
This one branched from an outdated main, so updating may fix the doc build failure.

ev-br:
> updating may fix the doc build failure.

Done.

j-bowhay (approved):

PR #24086, opened 2025-11-30
DOC: Improve the formatting of references in `scipy.stats`

star1327p:

Improve the formatting of references in `scipy.stats`


mdhaber:
The references sections of the documentation are certainly in need of work! Rather than picking several aspects of a few references, can we discuss a scope and make the change systematically?

For instance, maybe it would be helpful to discuss what references *should* look like. For instance:
- What style (e.g. MLA) should we use?
- For each reference type (e.g. journal), what fields (e.g. title) should be included?
- What fields should be omitted? For instance, journal volume and issue numbers don't seem very helpful when there is no ambiguity about the article, such as when there is a DOI or stable URL.

That may be a bit more than you're interested in tackling right now. No problem - we can keep the scope smaller, but still think about a bigger picture:
- Should we be including the access date in references?
- Should DOIs be included using the `:doi:` role or as a `doi.org` URL?

Either way, after making some decisions, PRs would:
- Implement the recommendations consistently throughout a module or throughout the entire library (depending on the scope).
- Add a tool that helps us keep things consistent automatically.
- Document the recommendations so other contributors have something to follow.

How does that sound? If it sounds good, what would you like the scope of this PR to be?

star1327p:
@mdhaber Thank you for your thoughtful comments!

My recommendation is to have a `.bib` file for all references, and use LaTeX compilation to generate the bibliography. The main advantage is to keep all references consistent without worrying about the formatting. However, this would require major changes to the scipy structure, and I don't think I have enough understanding to do so.

Regarding the questions:

> Should we be including the access date in references?  

I think including the access date is necessary only when the reference does not contain its published date.

> Should DOIs be included using the :doi: role or as a doi.org URL?

I prefer the :doi: role that can link directly to the URL.

mdhaber:
Great. Using a BiBTeX and such to generate references would be amazing, but you're right, that would be very ambitious. We can start smaller.

> I prefer the :doi: role that can link directly to the URL.

OK, since there are already a few DOI-related changes here, what do you think about focusing the PR on that? Instead of singling out a few functions in `stats`, it wouldn't be too difficult to fix _all_ the issues of this class throughout the documentation.

The simplest thing would be to just search (e.g. `grep`) for `doi`, maybe excluding existing uses of `:doi:` and `{doi}`, and fixing appearances as necessary. For example, this sort of search would find `doi:https://doi.org/10.3102/10769986001002113` as you did in the PR, but rather than changing it to `https://doi.org/10.3102/10769986001002113`, you would change it to:
```
:doi:`10.3102/10769986001002113`
```

I would be more interested in reviewing a PR like this. One reason is that it would provide a more complete solution to a known issue. Also, for my brain, at least, it's a lot easier to look for and fix all the occurences of a given issue in one go rather than making some fixes of a few different types. So the ratio of the benefit to the effort seems a lot more favorable.

The next step would be to document this recommendation - but actually, it's already in the contributor guide:
https://docs.scipy.org/doc/scipy/dev/contributor/rendering_documentation.html#dois-in-references

And to really solve the problem once and for all, it would be great to have a tool that looks for incorrect uses of `doi` in docstrings. It could be something that is run manually everyone once in a while, or - if false-positives are very rare - it could become part of CI to enforce the recommendation. The same sort of thing could be done with arXiv links, apparently. If you're interested in working on that, we can figure out where it would go.

star1327p:
@mdhaber Thank you for the detailed instructions again. This time I replaced many `https://doi.org/` links with the `:doi:` notation. I modified only the Python-related files, and I did not touch any C++ files to avoid complications.

scipy/spatial/transform/_rotation_cy.pyx
```diff
@@ -786,7 +786,7 @@ def as_euler(
     # directly by Rotation. See:
     # Bernardes E, Viollet S (2022) Quaternion to Euler angles conversion: A
     # direct, general and computationally efficient method.
-    # PLoS ONE 17(11): e0276302. https://doi.org/10.1371/journal.pone.0276302
+    # PLoS ONE 17(11): e0276302. :doi:`10.1371/journal.pone.0276302`.
```
mdhaber: Let's not change the comment; the `:doi:` role only works when rendered with Sphinx.

scipy/integrate/tests/test_cubature.py
```diff
@@ -28,7 +28,7 @@
 #   A.C. Genz, A.A. Malik, Remarks on algorithm 006: An adaptive algorithm for
 #   numerical integration over an N-dimensional rectangular region, Journal of
 #   Computational and Applied Mathematics, Volume 6, Issue 4, 1980, Pages 295-302,
-#   ISSN 0377-0427, https://doi.org/10.1016/0771-050X(80)90039-X.
+#   ISSN 0377-0427, :doi:`10.1016/0771-050X(80)90039-X`.
```
mdhaber:
```suggestion
#   ISSN 0377-0427, https://doi.org/10.1016/0771-050X(80)90039-X.
```

scipy/special/_add_newdocs.py
```diff
@@ -4445,7 +4438,7 @@ def add_newdoc(name, doc):
            http://netlib.org/amos/
     .. [2] Donald E. Amos, "Algorithm 644: A portable package for Bessel
            functions of a complex argument and nonnegative order", ACM
-           TOMS Vol. 12 Issue 3, Sept. 1986, p. 265
+           TOMS Vol. 12 Issue 3, Sept. 1986, p. 265.
```
mdhaber, line 4441: I won't ask for these to be undone, but for future reference, it would be easier to review if the PR consisted *only* of the stated scope of the PR - the `:doi:` role adjustments, in this case.

mdhaber (commented):
Thanks @star1327p. Is this comprehensive, or do you think there are other instances that might need to be changed later?  

And it sounds like the search was only for `https://doi.org/` - not other strings involving `doi` that might benefit from adjustment? It would be nice to look for other related mistakes, e.g.
https://github.com/scipy/scipy/blob/783b39b7d77364368c33f50f5bc19033d7e44878/scipy/linalg/_basic.py#L1860

image

I can't recomend a specific query - it depends on how you choose to search, but I think there are ways of looking for all appearances of `doi` without too many false positives: for instance, search only the Reference sections of docstrings using [docscrape](https://github.com/scipy/scipy/blob/main/scipy/_lib/_docscrape.py), or search for `doi` followed by non-alphabetic characters.

star1327p:
@mdhaber I don't think any improvised search can be comprehensive. 

However, I performed another search on `. doi:` to correct this pattern, because the `doi` mark should start and end with `:`.

In this manual change, I am very conservative on what needs to be revised and what does not. The top priority is not to break the scipy compilation workflow.

> Thanks @star1327p. Is this comprehensive, or do you think there are other instances that might need to be changed later?
> 
> And it sounds like the search was only for `https://doi.org/` - not other strings involving `doi` that might benefit from adjustment? It would be nice to look for other related mistakes.

mdhaber:
> I don't think any improvised search can be comprehensive.

I trust most search tools will bring up all occurences of "doi" if that is the search query. If all of those are manually inspected, that would be comprehensive in my book.

I used my IDE to search `py`, `rst`, and `md` files for "doi" as a whole word. After fixing the issues I found there, I went back and searched for the regular expression `(?<!:)doi(?!:)`. Another pass might actually review appearances of `:doi:` to make sure that the DOI is enclosed with beginning and ending backticks.

I'm aware of a few mistakes that are not fixed in my most recent commit. If you find them, I think we can be confident that this is "comprehensive". (*Update: fixed two of them already, since I didn't mean to introduce mistakes.*)

> In this manual change, I am very conservative on what needs to be revised and what does not. The top priority is not to break the scipy compilation workflow.

Fortunately, you can build the documentation locally to check for that, and as a last line of defense, there is continuous integration. A PR that fixes only Python docstrings is unlikely to break the scipy compilation workflow, but in any case, I wouldn't merge a PR that did. So I would recommend trying to be thorough first, and if we *do* run into problems (like sphinx build errors) we can debug them before merging.

scipy/ndimage/_filters.py
```diff
@@ -1650,7 +1650,7 @@ def minimum_filter1d(input, size, axis=-1, output=None,
 
     References
     ----------
-    .. [1] http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.42.2777
+    .. [1] http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.42.2777.
```
mdhaber, line 1653: I put the period here and below just to mark these as something to review - this URL (http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.42.2777) loads a page that looks like:
image

which is not a valid reference. So I think it would be nice in follow-up work to fix this reference. Maybe it meant https://doi.org/10.1109/78.542446 or https://doi.org/10.1109/ISCAS.1996.540338, which introduced a "MAXLIST" algorithm?

*Never mind, I think this can just be considered another incomplete/broken reference. There are probably hundreds or thousands of these, so this one is not special. Something to help address this sort of problem would be a broken link detector. The second reference link is broken, too. The wayback machine has it [here](https://web.archive.org/web/20190331105834/http://richardhartersworld.com/cri/2001/slidingmin.html). Unclear whether the algorithm is directly related to MINLIST/MAXLIST.*

scipy/optimize/_tstutils.py
```diff
@@ -34,7 +34,7 @@
 #  [1] Alefeld, G. E. and Potra, F. A. and Shi, Yixun,
 #      "Algorithm 748: Enclosing Zeros of Continuous Functions",
 #      ACM Trans. Math. Softw. Volume 221(1995)
-#       doi = {10.1145/210089.210111},
+#      https://dl.acm.org/doi/10.1145/210089.210111
```
mdhaber, line 37: Looks like this was in BiBTeX style? In a comment, a URL seems easier to work with.

scipy/sparse/linalg/_dsolve/linsolve.py
```diff
@@ -60,12 +60,12 @@ def use_solver(**kwargs):
...
            multifrontal method with a column pre-ordering strategy, ACM
            Trans. on Mathematical Software, 30(2), 2004, pp. 196--199.
-           https://dl.acm.org/doi/abs/10.1145/992200.992206
+           :doi:`10.1145/992200.992206`.
 
     .. [2] T. A. Davis, A column pre-ordering strategy for the
            unsymmetric-pattern multifrontal method, ACM Trans.
            on Mathematical Software, 30(2), 2004, pp. 165--195.
-           https://dl.acm.org/doi/abs/10.1145/992200.992205
+           :doi:`https://doi.org/10.1145/992200.992205`.
```
mdhaber: Oops.
```suggestion
           :doi:`10.1145/992200.992205`.
```

scipy/stats/_stats_py.py
```diff
@@ -3837,7 +3837,7 @@ def f_oneway(*samples, axis=0, equal_var=True):
 
     .. [4] B. L. Welch, "On the Comparison of Several Mean Values:
            An Alternative Approach", Biometrika, vol. 38, no. 3/4,
-           pp. 330-336, 1951. https://doi.org/10.2307/2332579
+           pp. 330-336, 1951. :doi:`doi.org/10.2307/2332579`.
```
mdhaber: Oops again.
```suggestion
           pp. 330-336, 1951. :doi:`10.2307/2332579`.
```

scipy/optimize/_zeros_py.py
```diff
@@ -1429,7 +1429,7 @@ def toms748(f, a, b, args=(), k=1,
        Alefeld, G. E. and Potra, F. A. and Shi, Yixun,
        *Algorithm 748: Enclosing Zeros of Continuous Functions*,
        ACM Trans. Math. Softw. Volume 221(1995)
-       doi = {10.1145/210089.210111}
+       https://doi.org/10.1145/210089.210111
```
mdhaber, line 1432:
```suggestion
       :doi:`10.1145/210089.210111`.
```
Oops. This was the third mistake I found in my second pass. I guess I fixed this incorrectly.

Anyway, this is the last one I'm aware of, but it's worth taking one more pass to see if you can find anything else.

scipy/special/_add_newdocs.py
```diff
@@ -1604,8 +1604,7 @@ def add_newdoc(name, doc):
            https://dlmf.nist.gov/19.16.E6
     .. [2] B. C. Carlson, "Numerical computation of real or complex elliptic
            integrals," Numer. Algorithm, vol. 10, no. 1, pp. 13-26, 1995.
-           https://arxiv.org/abs/math/9409227
-           https://doi.org/10.1007/BF02198293
+           :doi:`10.1007/BF02198293`. https://arxiv.org/abs/math/9409227
```
mdhaber, line 1607: Note all the other references were ending with `.` . That's not the point of this PR, so it's OK with me to leave it alone; just thought I'd make the note.

mdhaber (approved):
One more fixup of my own mistake to commit, but otherwise this LGTM, and I'm confident that this is reasonably comprehensive. 

As follow-up work, I think it would be good to do the same for arXiv and add an automated check for both.

fancidev:
> My recommendation is to have a `.bib` file for all references, and use LaTeX compilation to generate the bibliography. The main advantage is to keep all references consistent without worrying about the formatting. However, this would require major changes to the scipy structure, and I don't think I have enough understanding to do so.

There appears to be a sphinx extension, [sohinxcontrib-bibtex](https://sphinxcontrib-bibtex.readthedocs.io/en/latest/), that claims to do exactly that, though I have not tried it out myself!

PR #24091, opened 2025-12-01
MAINT: remove old numpy warning filters from pytest.ini

mattip:
#### Reference issue

Related to numpy/numpy#30340

NumPy has removed `numpy.distutils`, so it should no longer warn. Also remove older `numpy.core` warning. There is no harm in having the filters, but it would be nice to not carry baggage.

#### Additional information

There are more harmless warning filters here that probably can be removed.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -20,8 +20,6 @@ filterwarnings =
     ignore:'environmentfilter' is renamed to 'pass_environment'
     ignore:'contextfunction' is renamed to 'pass_context'
     ignore:.*The distutils.* is deprecated.*:DeprecationWarning
-    ignore:\s*.*numpy.distutils.*:DeprecationWarning
-    ignore:.*`numpy.core` has been made officially private.*:DeprecationWarning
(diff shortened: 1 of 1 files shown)


mattip:
Hmm, actually scipy is building for NumPy 1.26, so maybe this PR is too early?

tylerjereddy:
I haven't branched for `maintenance/1.17.x` yet if that's what you mean. I'm not sure on our plans for dropping 1.26.x after branching yet, I guess we'll see what Ralf thinks.

tylerjereddy (approved):
This should be "ok" to do now--I did a quick `git grep` of the code base for those two things, and I don't see any remaining problems.

We do have one warning filter decorating a unit test for `numpy.core` in `test_support_alternative_backends.py` -- I'm actually not sure we can remove that particular one because it may invoke backends that are still doing the old thing I suppose.

The one CI failure doesn't look related.

tylerjereddy:
thanks Matti

PR #24092, opened 2025-12-01
ENH: spatial.transform: Add `assume_valid` argument to `Rotation.from_matrix`

### LINKED ISSUE #24089: ENH: spatial.transform: Add `normalize` argument to `Rotation.from_matrix`

Some `from_` methods in `Rotation` have requirements on the input, e.g. the quaternion in `from_quat` must be unit norm. We enforce this by projections where necessary. This is computationally expensive, and oftentimes users may be able to guarantee that the input is already valid.

### Adding the `normalize` keyword to more representations

`from_quat` already has a `normalize` keyword, but skipping the normalization matters most for rotation matrices, which require an SVD to orthogonalize. We already have a skip path in place from https://github.com/scipy/scipy/pull/24056. I propose we expose that to users for significant speed-ups. 

The performance benefit can be seen in the benchmarks from https://github.com/scipy/scipy/issues/23391#issue-3265938509, where the difference in `Rotation.from_matrix` between our version and the jax native stems from the fact that jax does not normalize.

@lucascolley @scottshambaugh

amacati:
This PR adds an `assume_valid` keyword-only argument to `Rotation.from_matrix` which optionally disables the SVD projection. Users that can guarantee their matrix input is valid can save considerable time by skipping the SVD on eager and especially lazy frameworks. This should close the gap to the jax native performance seen in the benchmarks from #23391.

It also optimizes the existing projection to perform the SVD only where necessary on eager frameworks (suggested in https://github.com/scipy/scipy/issues/24089#issuecomment-3597166650).

#### Example
```python
import numpy as np

# This is valid and significantly faster
r = R.from_matrix(np.eye(3), assume_valid=True)  

# This is invalid, but still works. Using assume_valid=True puts the responsibility
# on the user to ensure the inputs fulfill all conditions of rotation matrices
# The default remains False, of course, and assume_valid is a keyword-only argument
r_invalid = R.from_matrix(np.random.normal((3, 3)), assume_valid=True)
```

#### Implementation
The only required change is that we skip checks and projections and directly take the `_from_matrix_orthogonal` code path introduced in https://github.com/scipy/scipy/pull/24056.

#### Reference issue
Closes #24089

#### Correction
I said earlier that we already have a `normalize` kwarg for `from_quat`. Turns out I misremembered, we only have it for `Rotation.__init__()`. Should we add this to `from_quat`? Contrary to the rather expensive SVD, getting rid of the norm is a minor optimization, but it still has its use cases, e.g. when users want to transfer rotations between frameworks with

```python
import torch
import jax.numpy as jnp

rot_torch = R.from_quat(torch.ones((3, 5, 4)))
rot_jax = R.from_quat(jnp.from_dlpack(rot_torch.as_quat()), assume_valid=True)
```
As previously mentioned, this is a minor optimization. If any, it should go into a separate PR to not mix things up.

### DIFF
--- a/scipy/spatial/transform/_rotation.py
+++ b/scipy/spatial/transform/_rotation.py
@@ -596,7 +596,7 @@ class Rotation:
         matrix = _promote(matrix, xp=xp)
         # Resulting quat will have 1 less dimension than matrix
         backend = select_backend(xp, cython_compatible=matrix.ndim < 4)
-        quat = backend.from_matrix(matrix, normalize=normalize)
+        quat = backend.from_matrix(matrix, assume_valid=assume_valid)
         return Rotation._from_raw_quat(quat, xp=xp, backend=backend)
 
     @staticmethod
--- a/scipy/spatial/transform/_rotation.py
+++ b/scipy/spatial/transform/_rotation.py
@@ -477,7 +477,7 @@ class Rotation:
     @xp_capabilities(
         skip_backends=[("dask.array", "missing linalg.cross/det functions")]
     )
-    def from_matrix(matrix: ArrayLike, *, normalize: bool = True) -> Rotation:
+    def from_matrix(matrix: ArrayLike, *, assume_valid: bool = False) -> Rotation:
         """Initialize from rotation matrix.
 
         Rotations in 3 dimensions can be represented with 3 x 3 orthogonal
@@ -492,11 +492,12 @@ class Rotation:
         matrix : array_like, shape (..., 3, 3)
             A single matrix or an ND array of matrices, where the last two dimensions
             contain the rotation matrices.
-        normalize : bool, optional
-            Must be True unless users can guarantee the input is a valid rotation
+        assume_valid : bool, optional
+            Must be False unless users can guarantee the input is a valid rotation
             matrix, i.e. it is orthogonal, rows and columns have unit norm and the
-            determinant is 1. Setting this to False without ensuring these properties
-            is unsafe behavior. Default is True.
+            determinant is 1. Setting this to True without ensuring these properties
+            is unsafe behavior. If True, normalization steps are skipped, which can
+            improve runtime performance. Default is False.
 
         Returns
         -------
--- a/scipy/spatial/transform/_rotation.py
+++ b/scipy/spatial/transform/_rotation.py
@@ -477,7 +477,7 @@ class Rotation:
     @xp_capabilities(
         skip_backends=[("dask.array", "missing linalg.cross/det functions")]
     )
-    def from_matrix(matrix: ArrayLike) -> Rotation:
+    def from_matrix(matrix: ArrayLike, *, normalize: bool = True) -> Rotation:
         """Initialize from rotation matrix.
 
         Rotations in 3 dimensions can be represented with 3 x 3 orthogonal
@@ -492,6 +492,10 @@ class Rotation:
         matrix : array_like, shape (..., 3, 3)
             A single matrix or an ND array of matrices, where the last two dimensions
             contain the rotation matrices.
+        normalize : bool, optional
+            If True, then the given matrices are orthogonalized via Singular Value
+            Decomposition (SVD) to form a valid rotation matrix before the conversion.
+            Default is True.
 
         Returns
         -------
@@ -590,7 +594,7 @@ class Rotation:
         matrix = _promote(matrix, xp=xp)
         # Resulting quat will have 1 less dimension than matrix
         backend = select_backend(xp, cython_compatible=matrix.ndim < 4)
-        quat = backend.from_matrix(matrix)
+        quat = backend.from_matrix(matrix, normalize=normalize)
         return Rotation._from_raw_quat(quat, xp=xp, backend=backend)
 
     @staticmethod
--- a/scipy/spatial/transform/_rotation.py
+++ b/scipy/spatial/transform/_rotation.py
@@ -496,8 +496,9 @@ class Rotation:
             Must be False unless users can guarantee the input is a valid rotation
             matrix, i.e. it is orthogonal, rows and columns have unit norm and the
             determinant is 1. Setting this to True without ensuring these properties
-            is unsafe behavior. If True, normalization steps are skipped, which can
-            improve runtime performance. Default is False.
+            is unsafe and will silently lead to incorrect results. If True,
+            normalization steps are skipped, which can improve runtime performance.
+            Default is False.
 
         Returns
         -------
--- a/scipy/spatial/transform/_rotation.py
+++ b/scipy/spatial/transform/_rotation.py
@@ -493,9 +493,10 @@ class Rotation:
             A single matrix or an ND array of matrices, where the last two dimensions
             contain the rotation matrices.
         normalize : bool, optional
-            If True, then the given matrices are orthogonalized via Singular Value
-            Decomposition (SVD) to form a valid rotation matrix before the conversion.
-            Default is True.
+            Must be True unless users can guarantee the input is a valid rotation
+            matrix, i.e. it is orthogonal, rows and columns have unit norm and the
+            determinant is 1. Setting this to False without ensuring these properties
+            is unsafe behavior. Default is True.
 
         Returns
         -------
--- a/scipy/spatial/transform/_rotation_cy.pyx
+++ b/scipy/spatial/transform/_rotation_cy.pyx
@@ -458,7 +458,7 @@ def from_euler(seq, angles, bint degrees=False):
 @cython.embedsignature(True)
 @cython.boundscheck(False)
 @cython.wraparound(False)
-def from_matrix(matrix):
+def from_matrix(matrix, bint normalize=True):
     cdef int ind
     is_single = False
     matrix = np.array(matrix, dtype=float)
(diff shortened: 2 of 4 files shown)


scipy/spatial/transform/_rotation.py
```diff
@@ -492,6 +492,10 @@ def from_matrix(matrix: ArrayLike) -> Rotation:
         matrix : array_like, shape (..., 3, 3)
             A single matrix or an ND array of matrices, where the last two dimensions
             contain the rotation matrices.
+        normalize : bool, optional
+            If True, then the given matrices are orthogonalized via Singular Value
```
scottshambaugh: I'd reword this to something more along the lines of cautioning users to only make this False if the matrix is known orthogonal, as it's unsafe behavior otherwise. Also with the confusion in terminology we've had, should probably specify orthogonal and _unit_ norm.

scottshambaugh (commented):
If we don't currently have a similar normalize flag in the public API, then I'm a little more hesitant to introduce one. It opens up the possibility of undefined/unsafe behavior to users.

amacati:
Understandable. To be exact though, we do expose the unnormalized variant, but it's implicit: Users can create a `Rotation` directly from its `__init__` function to skip the normalization in `from_quat`. `__init__` does take the `normalize` keyword. 

For matrices there is no such option using the public API, but it's arguably more needed since the performance impact, especially on lazy implementations, is pretty large.

So I'm in favor of adding it, but agree that we should make it clear in the docs that the flag is strictly intended for users that can guarantee that the input is valid.

scipy/spatial/transform/_rotation.py
amacati: I agree and updated the description correspondingly. Let me know if you like the current wording better.

amacati:
Seems like CI failures are unrelated, but the logs are too huge for a thorough assessment. Dask is broken in several modules (fft, ndimage, signal).

amacati:
@lucascolley Any idea what's up with the CI?

lucascolley:
the same failures are observed in gh-24076

amacati:
They've just run again with the same failure mode though, so it seems like this is not a temporary hiccup, or am I missing something? 

@scottshambaugh could you review the changes, and we hold off with merging until the CI is resolved? As things stand, it is too easy to miss a test failure in those logs, but it would be great if this was merged before 1.17rc1 gets finalized. That is, of course, as long as you and @lucascolley are fine with the docs.

scipy/spatial/transform/_rotation.py
scottshambaugh: Looks great!

scottshambaugh:
Code looks good, I'm still iffy on exposing an unsafe interface though. We call out in the docs that `__init__` should never be used directly, so it's not really something in the API right now. Would like to hear another maintainer weight in on this.

amacati:
> Code looks good, I'm still iffy on exposing an unsafe interface though. We call out in the docs that `__init__` should never be used directly, so it's not really something in the API right now. Would like to hear another maintainer weight in on this.

@crusaderky @ev-br Do you have an opinion on this? I see the point, but feel like there are a lot of valid use cases where we would like to have this feature (e.g. interop with a simulator like MuJoCo that guarantees the input is a valid matrix, transformation matrices published in ROS, outputs from neural networks that already implement the SVD layer to guarantee predictions are valid and want to avoid a double SVD). 

The reason I'm advocating for it is that I have encountered these situations in my group and got asked about the feature by other researchers that need performance and can guarantee these properties, and the current workaround is to use the raw `rotation_xp` backend. Using the backend is not part of the public API, significantly more error-prone, and people that need the functionality don't know under which conditions their workaround is valid.

lucascolley:
I don't think the interface being 'unsafe' is inherently problematic as long as it is clearly opt-in. Consider the `assume_a` kwarg found across `linalg` at places like https://scipy.github.io/devdocs/reference/generated/scipy.linalg.inv.html.

I suppose we have missed the boat on calling it `assume_valid`, given the kwarg for `__init__`? Regardless, I am +1 on including this for 1.17.0 in some form.

amacati:
As Scott said, we recommend to not use init. So while `assume_valid` is inconsistent with the internal naming scheme, it would not conflict with any public API.

lucascolley:
> while `assume_valid` is inconsistent with the internal naming scheme, it would not conflict with any public API.

in which case I am +0.5 for this as-is and +1 for renaming to `assume_valid`

amacati:
Changed everything to `assume_valid`. I agree that the wording is stronger and makes it clearer that users must guarantee the properties.

scipy/spatial/transform/_rotation.py
```diff
@@ -492,6 +492,12 @@ def from_matrix(matrix: ArrayLike) -> Rotation:
         matrix : array_like, shape (..., 3, 3)
             A single matrix or an ND array of matrices, where the last two dimensions
             contain the rotation matrices.
+        assume_valid : bool, optional
+            Must be False unless users can guarantee the input is a valid rotation
+            matrix, i.e. it is orthogonal, rows and columns have unit norm and the
+            determinant is 1. Setting this to True without ensuring these properties
```
ev-br, line 498: It might be obvious for users (which I'm not): which norm is it? 

Also, if I'm doing the normalization myself: is there any target tolerance on how close the determinant is to 1 in floating-point? IOW, if I normalize it myself and it's 1e-10 off, am I in trouble? 1e-5? what are the failure modes, is there some diagnostics, what do I need to watch out for?

scipy/spatial/transform/_rotation.py
```diff
@@ -492,6 +492,12 @@ def from_matrix(matrix: ArrayLike) -> Rotation:
...
+            is unsafe behavior. If True, normalization steps are skipped, which can
```
ev-br: Consider strengthening the wording. As is, it's a single word, "unsafe", in a middle of a paragraph. Reading it for the first time I can't help wondering what it means: crashes, segfaults, silently wrong answers, or something else? Some readers can imagine the Rust-popularized meaning, which is I suppose not what it meant here.

ev-br (commented):
>  @ev-br Do you have an opinion on this? I see the point, but feel like there are a lot of valid use cases where we would like to have this feature (e.g. interop with a simulator like MuJoCo that guarantees the input is a valid matrix, transformation matrices published in ROS, outputs from neural networks that already implement the SVD layer to guarantee predictions are valid and want to avoid a double SVD).

Presumably there is a demonstrably large performance impact, via some existing or new benchmark?

There are precedents elsewhere in scipy:
- linalg functions have a `check_finite` keyword to toggle checking for nans (which costs O(N**2))
- [`PPoly.construct_fast`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.PPoly.construct_fast.html) is the same as `__init__` only skips all checks. This is a public function, but it's really mainly for internal uses, where we do know the inputs are valid and have been already checked.

For `PPoly`, there is only a single "normal" constructor, so it's natural to provide an alternative, "unsafe" one. 
Here IIUC the story is different, and it's probably not reasonable to give each `from_???` method an "unsafe" companion. So I suppose a kwarg is OK, provided it's flagged very clearly as being only for those who know what they are doing

amacati:
> Presumably there is a demonstrably large performance impact, via some existing or new benchmark?

Yes. We recover the performance of the native JAX implementation with this option (see the original benchmark for `from_matrix` [here](https://github.com/scipy/scipy/issues/23391)). At the moment, lazy frameworks **always** do an SVD. For eager frameworks such as torch and numpy, we still avoid the orthogonality check, so for large batches that's saving us from a Nx3x3 @ Nx3x3 matrix multiplication.

scipy/spatial/transform/_rotation.py
amacati: I think this should be indeed obvious for users of the module. Rotation matrices have unit L2 norm for rows and columns (unit vectors). 

The determinant is either 1 of -1 by construction if the columns and rows are orthogonal and normal, so this is covered via the condition on the rows and columns. At the moment, we check if `R.T @ R` is equal to `xp.eye(3)` up to 1e-12 precision. I am not sure if we should specify this though. In my opinion, users either have normalized the matrix up to machine precision, or the matrix cannot be considered valid.

scipy/spatial/transform/_rotation.py
amacati: I considered the `must be False unless users can guarantee` to already be part of the warning. Isn't that strong enough to signal that using `assume_valid` without guaranteeing the conditions is a bad idea?

The clarification on unsafe is probably a good idea. How would you formulate this? This is not related to the Rust-meaning (everything is still well defined from a program perspective), but the results can be different.

ev-br: Sure it's not Rust meaning :-). 

> but the results can be different.

Different how? linalg `check_finite` description <s>promises terrible retributions</s> [mentions](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.svd.html) _crashes, non-termination if the inputs do contain infinities or NaNs_ . From your comment it sounds like the the failure mode is silently getting
wrong answers? If so, I'd mention it explicitly. [1]

[1] if not known, I'd intentionally feed it with an invalid matrix and run a few experiments.

ev-br:
> so for large batches that's saving us from a Nx3x3 @ Nx3x3 matrix multiplication.

Great. I was hinting at adding a benchmark :-). As it's not immediately clear --- maybe it is for those deep in the know of the module? --- what is the relative cost of this matmul to other work.
Alternatively, consider adding a discussion to the user guide, with an informal `%timeit` demonstration.

None of my comments are blocking though, take them as suggestions.

amacati:
> Great. I was hinting at adding a benchmark :-). As it's not immediately clear --- maybe it is for those deep in the know of the module? --- what is the relative cost of this matmul to other work.

I know, but I will have to dust off my benchmark suite first. A coarse data point: With 1e7 matrices, jax goes from 1.17 seconds to 0.68s. This benchmark was not run by me though, so take this with a grain of salt. I would expect the improvement to be more significant.

The user guide lives in the docs I assume?

ev-br:
> The user guide lives in the docs I assume?

Yup, https://docs.scipy.org/doc/scipy/tutorial/spatial.html 
Now that I actually looked at it, I does not have anything on Rotations at all (hint hint).

> At the moment, lazy frameworks always do an SVD. For eager frameworks such as torch and numpy, we still avoid the orthogonality check

I suppose there is a _very_ deep reason for the divergence. In general we want to make the behavior agnostic to the array provider, and only reluctantly allow deviations if there is absolutely no reasonable way to avoid it.

amacati:
> I suppose there is a _very_ deep reason for the divergence. In general we want to make the behavior agnostic to the array provider, and only reluctantly allow deviations if there is absolutely no reasonable way to avoid it.

The behaviour is the same, but the computation graph must take all paths. So we must compute both versions and then select the results based on what happens in the eager version. We essentially do an SVD for all matrices and then throw away the ones that were not needed. But the computation must happen either way.

scipy/spatial/transform/_rotation.py
amacati: Makes sense, will double-check this and include the results in an update

amacati:
Just to finish the discussion and satisfy @ev-br 's unquenchable thirst for benchmarks and hard numbers:

image

I think this makes a pretty compelling argument as for why we should expose this (with a stern warning) to researchers that work intensely with batched transforms.

With the latest clarification of the docs (and if the CI passes ofc) I'd be good for merging on my side.

amacati:
Something major is broken with the CI again, the logs of Array API are exploding. This is very likely not related to the PR, but maybe we should wait for this to be fixed and have a clean run before we merge as long as it still makes it into 1.17.

amacati:
@lucascolley CI looks good

scottshambaugh (approved):
Nice benchmarks! LGTM

lucascolley (approved):
thanks all for the effort here! 🚀

PR #24094, opened 2025-12-02
TST: fft: `allow_dask_compute` for all `uarray`-using funcs

steppi:

This PR adds `allow_dask_compute=True` to the `xp_capabilities` of all functions in `fft` which rely on dispatch with `uarray` in attempt to fix the test failures noted by @lucascolley here: https://github.com/scipy/scipy/pull/24039#issuecomment-3602457415. For mysterious reasons, these failures did not occur in the CI #24039 but are now occurring on main. I found they did not occur on `main` for my current local dev environment, even with the bump to `array_api_extra`, but I was able to reproduce locally after creating a fresh `venv` with the same Python version used in CI (Python 3.13.9) and with a fresh install of all dependencies. The Python version did not change in CI, so Python version appears to be unrelated to why the tests suddenly started failing. It appears to be due to some change in dependencies or other but I did not investigate further.

A vexing thing is that, contrary to@lucascolley's suspicion about this relating to functions which depend on offending functions in `fft`, I've found the failures actually occur in tests for which there is no dependence on `fft` whatsoever, for instance `test_romb` in `integrate`

```
2025-12-02T14:36:09.6294603Z FAILED scipy/integrate/tests/test_quadrature.py::TestRomb::test_romb[dask.array] - AssertionError: Called `dask.compute()` or `dask.persist()` 1 times, but no calls are allowed. Set `lazy_xp_function(ifft, allow_dask_compute=1)` to allow for more (but note that this will harm performance).
```

and many more places. My guess is this is some kind of bizarre thread safety issue in the `lazy_xp_function` machinery causing failures to bleed through into other tests. I do not know why this is happening, my gut tells me it probably has something to do with the interaction of `uarray` and the `lazy_xp_function` machinery. I've added `allow_dask_compute=True` to every `xp_capabilities` for functions in `fft` which use `uarray` and this makes the tests pass for me locally. I've used `allow_dask_compute=True` instead of `allow_dask_compute=1` because I've found that the number of dask computes used can be flaky: I was actually seeing some failures locally where 2 dask computes occurred, possibly due to bleeding between threads.

### DIFF
--- a/scipy/fft/_basic.py
+++ b/scipy/fft/_basic.py
@@ -22,7 +22,7 @@ def _dispatch(func):
     return generate_multimethod(func, _x_replacer, domain="numpy.scipy.fft")
 
 
-@xp_capabilities()
+@xp_capabilities(allow_dask_compute=True)
 @_dispatch
 def fft(x, n=None, axis=-1, norm=None, overwrite_x=False, workers=None, *,
         plan=None):
@@ -168,7 +168,7 @@ def fft(x, n=None, axis=-1, norm=None, overwrite_x=False, workers=None, *,
     return (Dispatchable(x, np.ndarray),)
 
 
-@xp_capabilities()
+@xp_capabilities(allow_dask_compute=True)
 @_dispatch
 def ifft(x, n=None, axis=-1, norm=None, overwrite_x=False, workers=None, *,
          plan=None):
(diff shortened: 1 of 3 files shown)


lucascolley (approved):
thanks Albert! Let's put this in to unblock CI.

PR #24095, opened 2025-12-03
DOC: differential_evolution - fix typo in custom strategy example

jonaspleyer:
[skip ci]

- This should be self-explanatory
- A tuple must be indexed with brackets, not braces.


lucascolley (approved):
thanks!

Issue #23931, opened 2025-11-05
BUG: stats.relfreq: relative frequencies do not sum to 1 when `weights` is used


mdhaber:
### Describe your issue.

The relative frequencies of `relfreq` are supposed to sum to 1, but they don't when `weights` is used. 

Looks like this is because:

https://github.com/scipy/scipy/blob/ab936e4839119b51b682f80dec68e0978b8b7646/scipy/stats/_stats_py.py#L2496

normalizes by `a` unconditionally, whereas it should normalize by the sum of the weights if `weights` is used.

### Reproducing Code Example

```python
import numpy as np
from scipy import stats
rng = np.random.default_rng(285494985394583)
x = rng.integers(1, 11, size=100) / 10.
vals, weights = np.unique_counts(x)

ref = stats.relfreq(x)
res = stats.relfreq(vals, weights=weights)
res.frequency.sum()  # 10.0
np.testing.assert_equal(res, ref)
```

### Error message

```shell
Mismatched elements: 10 / 10 (100%)
Max absolute difference among violations: 1.26
Max relative difference among violations: 9.
 ACTUAL: array([1.1, 0.9, 0.6, 1.4, 1.1, 1. , 0.5, 1. , 1. , 1.4])
 DESIRED: array([0.11, 0.09, 0.06, 0.14, 0.11, 0.1 , 0.05, 0.1 , 0.1 , 0.14])
```

### SciPy/NumPy/Python version and system information

```shell
main
```

AdrianRasoOnGit:
Hi! I'd like to work on this, if that's okay

josef-pkt:
I remember a long, long time ago, something similar to this could be used to compute the expectation of the `weights` function with respect to the frequency distribution.
So not normalizing by weights was useful.

The current behavior would not necessarily be a BUG.

(might have been a similar numpy function, and I have not used any of these in a long time.)

mdhaber:
I'm happy to reclassify as DOC if we decide that's what needs to change. Without context, though, it seems like the behavior disagrees with the name/description of the function. In retrospect, perhaps a more descriptive name would have been `relfreq_when_all_weights_are_one_otherwise_not_relfreq` if that is the desired behavior. Or maybe better to just to use use `histogram` directly.

vaishnavirao24:
i just want to confirm one thing
when the weights are used with relfreq, is the expected behavior that the result should sum to 1 (which is  normalize by weights.sum()), or is the current behavior intentional?

[...]

Issue #23868, opened 2025-10-26
BUG: stats: docstrings of several discrete distr. methods show unsupported 'scale' parameter


WarrenWeckesser:
### Describe your issue.

The parameter `scale` is included in the docstrings of these discrete distribution methods, but in fact, `scale` is not accepted by these methods: `entropy`, `interval`, `mean`, `median`, `moment`, `stats`, `std`, `support`, `var`.

### Reproducing Code Example

```python
In [34]: from scipy.stats import binom

In [35]: binom.var(10, 0.25, scale=0.5)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[35], line 1
----> 1 binom.var(10, 0.25, scale=0.5)

File ~/repos/git/forks/scipy/build-install/usr/lib/python3.13/site-packages/scipy/stats/_distn_infrastructure.py:1446, in rv_generic.var(self, *args, **kwds)
   1427 """Variance of the distribution.
   1428 
   1429 Parameters
   (...)   1443 
   1444 """
   1445 kwds['moments'] = 'v'
-> 1446 res = self.stats(*args, **kwds)
   1447 if isinstance(res, ndarray) and res.ndim == 0:
   1448     return res[()]

File ~/repos/git/forks/scipy/build-install/usr/lib/python3.13/site-packages/scipy/stats/_distn_infrastructure.py:1174, in rv_generic.stats(self, *args, **kwds)
   1148 def stats(self, *args, **kwds):
   1149     """Some statistics of the given RV.
   1150 
   1151     Parameters
   (...)   1172 
   1173     """
-> 1174     args, loc, scale, moments = self._parse_args_stats(*args, **kwds)
   1175     # scale = 1 by construction for discrete RVs
   1176     loc, scale = map(asarray, (loc, scale))

TypeError: _parse_args_stats() got an unexpected keyword argument 'scale'
```

<details>

<summary>
SciPy/NumPy/Python version and system information
</summary>

```shell
1.17.0.dev0+git20251026.2161a68 2.3.4 sys.version_info(major=3, minor=13, micro=7, releaselevel='final', serial=0)
Build Dependencies:
  blas:
    detection method: pkgconfig
    found: true
    include directory: /Users/warren/local/include
    lib directory: /Users/warren/local/lib
    name: openblas
    openblas configuration: USE_64BITINT= DYNAMIC_ARCH= DYNAMIC_OLDER= NO_CBLAS= NO_LAPACK=
[...]

PR #24097, opened 2025-12-03
ENH: stats: added marginal function to `stats.multivariate_t`

JacobHass8:
Adds the `marginal` function to `stats.multivariate_t` to return a marginal probability distribution. See issue gh-23901.


scipy/stats/_multivariate.py
```diff
@@ -5502,6 +5502,34 @@ def _process_parameters(self, loc, shape, df):
...
+        marginal_multivariate_t : multivariate_t_frozen
+            An object representing the marginal t.
+        
+        Notes
+        -----
+        %(_mvt_doc_frozen_callparams_note)s
+        """
+        params = self._process_parameters(loc, shape, df)
+        n, loc, shape, df = params
+        dims = _validate_marginal_input(dimensions, n)
```
mdhaber, line 5529: Great that everything could be re-used and the rest of the implementation is so simple.

scipy/stats/tests/test_multivariate.py
```diff
@@ -3718,6 +3718,71 @@ def test_entropy_1d(self, df):
     #     return -integrate.dblquad(integrand, -limit, limit, -limit, limit,
     #                               args=(mvt, ))[0]
 
+    @pytest.mark.parametrize('X_ndim', [3])
+    @pytest.mark.parametrize('dimensions', [[1], [-1, 1]])
+    @pytest.mark.parametrize('frozen', [True, False])
+    @pytest.mark.parametrize('df', [0.1, 1, 2])
+    def test_marginal_distribution(self, X_ndim, dimensions, frozen, df):
```
mdhaber: These look pretty similar to the multivariate normal tests. Could you factor those into a new class and `parametrize` them like `(dist, [stats.multivariate_normal, stats.multivariate_t])` to reuse that code? Alternatively, `TestMultivariateNormal` and `TestMultivariateT` class can inherit from that class, and you could set an instance variable to run the test with a different distribution.

JacobHass8: I think this would be easy for the `test_marginal_input_validation` and `test_marginal_special_cases` functions. Here's a code snippet of what I've come up with. 
```python
class TestMarginal:
    @pytest.mark.parametrize('dist', [multivariate_normal, multivariate_t])
    def test_marginal_input_validation(self, dist):
        rng = np.random.default_rng(413911473)
        mean = rng.standard_normal(3)
        A = rng.standard_normal((3, 3))
        cov = A @ A.T

        X = dist(mean, cov)

        msg = r"Dimensions \[3\] are invalid .*"
        with pytest.raises(ValueError, match=msg):
            X.marginal(3)

        with pytest.raises(ValueError, match=msg):
            X.marginal([0, 1, 2, 3])

        msg = r"All elements of `dimensions` must be unique."
        with pytest.raises(ValueError, match=msg):
            X.marginal([2, -1])

        with pytest.raises(ValueError, match=msg):
            X.marginal([[0, 1]])
        
        msg = r"Elements of `dimensions` must be integers."
        with pytest.raises(ValueError, match=msg):
            X.marginal([1.1, 2.0])
    
    @pytest.mark.parametrize('dist', [multivariate_normal, multivariate_t])
    def test_marginal_special_cases(self, dist):
        rng = np.random.default_rng(413911473)
        mean = rng.standard_normal(3)
        A = rng.standard_normal((3, 3))
        cov = A @ A.T

        X = dist(mean, cov)
        
        msg = r"Cannot marginalize all dimensions."
        with pytest.raises(ValueError, match=msg):
            X.marginal([])
```

JacobHass8: I'm not sure how to write the `test_marginal_distribution` function though. The main issue is that `multivariate_normal` and `multivariate_t` have slightly different parameters to test. For the `multivariate_normal` function, we test if `cov` is passed as a matrix or `covariance` object. For the `multivariate_t` function, we have the additional parameter `df` which we pass. I'm not sure how to build those into a single generic test. Here's what I have for a function below.  
```python
@pytest.mark.parametrize('dist', [multivariate_normal, multivariate_t])
@pytest.mark.parametrize('X_ndim', [3])
@pytest.mark.parametrize('dimensions', [[1], [-1, 1]])
@pytest.mark.parametrize('frozen', [True, False])
def test_marginal_distribution(self, dist, X_ndim, dimensions, frozen):
      rng = np.random.default_rng(413911473)
      mean = rng.standard_normal(X_ndim)
      A = rng.standard_normal((X_ndim, X_ndim))
      cov = A @ A.T

      # number of points at which to evaluate marginal PDF
      x = np.random.standard_normal((4, len(dimensions)))
      X = dist(mean, cov)
      
      if frozen:
          Y = X.marginal(dimensions)
          res = Y.pdf(x)
      else: 
          Y = dist.marginal(dimensions, mean, cov)
          res = Y.pdf(x)

      ref = marginal_pdf(X, X_ndim, dimensions, x)
      assert_allclose(ref, res)
```

mdhaber: OK, you can parametrize like:
```python3
@pytest.mark.parametrize('dist, kwargs', [(multivariate_normal, {}), 
                                          (multivariate_t, {'df': 10.})])
```
and then in the test:
```python3
if cov_object and dist == 'multivariate_t':
    pytest.skip('`multivariate_t` does not accept a `Covariance` object')
```
At some point we wanted to add support for `Covariance` to `multivariate_t`, at which point we could remove the skip.

JacobHass8: I went ahead and made the changes you suggested. It feels wrong to have so many `@pytest.mark.parametrize` decorators. I also deleted the old tests in the `TestMultivariateNormal` and `TestMultivariateT` classes.

scipy/stats/_multivariate.py
```diff
@@ -5502,6 +5502,34 @@ def _process_parameters(self, loc, shape, df):
...
+
+        loc = loc[dims]
+        shape = shape[np.ix_(dims, dims)]
+
+        return multivariate_t_frozen(loc, shape, df, allow_singular)
```
mdhaber, line 5535:
```suggestion


```

scipy/stats/_multivariate.py
```diff
@@ -1059,7 +1059,7 @@ def entropy(self):
         return 0.5 * (rank * (_LOG_2PI + 1) + log_pdet)
 
     def marginal(self, dimensions):
-        return self._dist.marginal(dimensions, self.mean, 
+        return self._dist.marginal(dimensions, self.mean,
                                    self.cov_object, self.allow_singular)
```
mdhaber, line 1063: Looks like previously, `self.allow_singular` was stored as an attribute in `__init__` just for fun? It doesn't appear in any of the other methods, even though it could have been passed to the other methods. I guess it's fine to fix it for just this method.

scipy/stats/tests/test_multivariate.py
```diff
@@ -1247,40 +1247,48 @@ def test_fit_fix_cov(self):
                                                      ).sum()
         assert logp_perturbed < logp_fix
 
+
+class TestMarginal:
```
mdhaber, line 1251: I moved this back to reduce the diff in this PR. If you want to refactor later, we can do that, but it's easier to review a minimal diff first, then separately I can review a copy-paste to somewhere else.

scipy/stats/tests/test_multivariate.py
```diff
@@ -1247,40 +1247,48 @@ def test_fit_fix_cov(self):
...
+    @pytest.mark.parametrize('dist,kwargs', [(multivariate_normal, {}),
+                                             (multivariate_t, {'df': 4})])
```
mdhaber, line 1253: I removed some of these - unless there's a particular reason to have three different `df`? Are there different code paths or something that could affect the behavior of `marginal`?

mdhaber (commented):
This looks pretty good! Did you ever find a source for the marginal distribution of a many-variable multivariate t?

> It feels wrong to have so many @pytest.mark.parametrize decorators. 

Sorry to hear that, but an extra few lines of parameterization looks a lot better to me than to repeat 70 lines of code. If you had to add a lot of extra logic inside the function to support it, I might agree with you, but it really fits quite naturally. Besides being easier to review, it ensures that the tests and the behavior they check are consistent when they should be.

scipy/stats/tests/test_multivariate.py
JacobHass8: Do you mean the individual tests in the `TestMultivariateNormal` class? That makes sense.

scipy/stats/_multivariate.py
JacobHass8: Yes, it doesn't seem like `allow_singular` is used for `multivariate_normal` or `multivariate_t`. However, `allow_singular` wasn't saved as an attribute in `multivariate_t` so I needed to add that.

scipy/stats/tests/test_multivariate.py
JacobHass8: I was only able to find a reference for the marginal distribution of the `multivariate_t` distribution for df > 2. That's why I wanted to test df = 1 and df<1. Here's the reference https://www.researchgate.net/publication/228613894_A_short_review_of_multivariate_t-distribution

mdhaber: I see. That requirement doesn't seem specific to the marginal distribution result, so I think mostly we'd want to check whether the other methods of the class are valid at all for `df <= 2`. That doesn't need to be done in this PR.

mdhaber: I think this reference is helpful:
https://web.ics.purdue.edu/~jltobias/BayesClass/full_semester_lecture_notes/distributions.pdf?utm_source=chatgpt.com

It notes what's going on with $\nu \leq 2$, and it doesn't make any restrictions on $\nu$ when discussing the marginal distribution.

JacobHass8: Oh, so the variance of the distribution is undefined for $\nu \leq 2$. That makes more sense. Maybe we don't need these checks then.

JacobHass8:
> Sorry to hear that, but an extra few lines of parameterization looks a lot better to me than to repeat 70 lines of code. If you had to add a lot of extra logic inside the function to support it, I might agree with you, but it really fits quite naturally. Besides being easier to review, it ensures that the tests and the behavior they check are consistent when they should be.

Yes, there doesn't seem to be a better alternative. I was mainly hesitant because there is extra logic in the function because of the parameterization. 

Are there any more changes I should fix before a review?

mdhaber:
> I was mainly hesitant because there is extra logic in the function because of the parameterization.

But even this small amount will go away when we add `Covariance` support to `multivariate_t`. Please fill free to pick up the torch: gh-20015.

> Yes, there doesn't seem to be a better alternative. 

Well, there is something that can reduce use of `parameterize` slightly. In https://github.com/scipy/scipy/pull/24097#discussion_r2586262598, I suggested that you could create a new class like:

```python3
class TestMultivariateNormalT:
```
and define the marginal tests there. Instead of parameterizing them w.r.t. argument `dist`, they rely on an attribute `self.dist`.

Then your existing `TestMultivariateNormal` and `TestMultivariateT` are subclasses with attribute `dist`, e.g.:
```python3
class TestMultivariateNormal(TestMultivariateNormalT)
    self.dist = stats.multivariate_normal
....
class TestMultivariateT(TestMultivariateNormalT)
    self.dist = stats.multivariate_t
```
so they inherit the test.

You still need the logic because `multivariate_t` doesn't support `Covariance` objects, but you reduce the parameterization a little. 

I don't mind it the way it is now, though. I'm not sure if the other way is "better", so I wouldn't recommend changing it.

mdhaber (approved):
I think this is fine - it just needs `marginal` added manually to the list of methods, like I did for `multivariate_normal`:
https://github.com/scipy/scipy/blob/67ab3c3196f7cfef7cef483d5bef7c9617355695/scipy/stats/_multivariate.py#L366-L367

Because right now, take a look at the preview [documentation](https://output.circle-artifacts.com/output/job/d1f8e4d9-5ec1-4926-b211-4fe8da4b22f5/artifacts/0/html/reference/generated/scipy.stats.multivariate_t.html#scipy.stats.multivariate_t) - the `marginal` method isn't shown.

Please include `[docs only]` in your commit message to run only the documentation (and lint) workflow.

---

After this PR, what did you want to do next?

JacobHass8:
> > I was mainly hesitant because there is extra logic in the function because of the parameterization.
> 
> But even this small amount will go away when we add `Covariance` support to `multivariate_t`. Please fill free to pick up the torch: [gh-20015](https://github.com/scipy/scipy/pull/20015).

Oh, I'll look into that. I'm not familiar with the `Covariance` object, but I can figure that out. 

> I don't mind it the way it is now, though. I'm not sure if the other way is "better", so I wouldn't recommend changing it.

Yeah, I think I just need to accept it accomplishes what we want in the most compact way. It will look nicer if `Covariance` support is added too.

JacobHass8:
> I think this is fine - it just needs `marginal` added manually to the list of methods, like I did for `multivariate_normal`:

I need to remember to also add the functions to the list of methods. Thanks for the check. 
 
> Please include `[docs only]` in your commit message to run only the documentation (and lint) workflow.

Shoot, I forgot to add this. I'm still getting used to the `scipy` workflow. What resources do these checks take up and does this clog up the `scipy` workflow? I'm not familiar with Github checks. 

> After this PR, what did you want to do next?

I'm not exactly sure because adding marginal distributions to other distributions would likely require new multivariate classes. I could try adding `Covariance` support for the `multivariate_t` distribution. I also saw #24086 and would be interested in adding .bib support or a check for missing `:doi:` usage. Is there anything pressing that you would suggest?

mdhaber:
> What resources do these checks take up and does this clog up the scipy workflow? I'm not familiar with Github checks.

GHA is mostly free (except maybe for GPU), and I don't think we often run into limits on concurrent workflows, but still, we appreciate that these resources are provided for free, and we try to respect the privilege by not using them unnecessarily.

> I could try adding `Covariance` support for the `multivariate_t` distribution. 

That would be great - or maybe you could start getting familiar with them  by *preserving* the `Covariance` type when using `marginal` on a `multivariate_normal` distribution. LMK if you run into trouble.

> I also saw https://github.com/scipy/scipy/pull/24086 and would be interested in adding .bib support or a check for missing :doi: usage.

Re: `:doi:` usage, please coordinate with the author to avoid duplicating effort. 
I think the BiBTeX idea is interesting, but as noted, it would be a major change. I would be interested in reviewing a summary of existing tools for this purpose. For instance, I'd love to see a proof of concept for how `sphinxcontrib_bibtex` might work with SciPy.

<details>

*Note about the following: this just a guess about the opinion of others. Personally, I would prefer a single BiBTeX source of truth, but I think it would be tough to get approval to remove formatted references from the committed docstrings.*

My guess is that however the system works, the `References` section of the docstrings should be committed with the formatted reference, not as BiBTeX or some intermediate representation. We would want `help <feature>` to show the references section much like it does now, without some sort of doc-build-time or import-time dynamic modification/injection of text. Does that make sense? So I can see there being a _separate_ BiBTeX representation of references that gets committed, too, and a tool would keep that in sync with the reference section of the docstrings, but I am guessing that it would be a tough sell if the references section were not actually populated as they are now.

</details>

> Is there anything pressing that you would suggest?

Here are some things that are mostly self-contained.

- It would be helpful for someone else to take a look at the status of gh-23622. After reviewing what was attempted before and the trouble we ran into, does the latest solution look convincing?
- If you have experience with compiled code, gh-21958. This is not something I would have the expertise to review, but if the solution is reasonably simple, we can find someone to take a look at a PR. Same with gh-23172, gh-22577, gh-21463, gh-23390.
- If you are interested, gh-21370. I would also not want to review that one, but perhaps the author of the issue would (and they are a maintainer, so they could merge). I would like to see it fixed.
- It would be great to have someone take a look at gh-22816 and gh-23238 and propose marking the old function as "legacy" (won't fix, see gh-17597) or actually deprecate it in light of gh-23851 merging.
- gh-23868
- Read gh-23931 carefully, without the assumption that "relative frequency" is equivalent to "frequencies sum to one" in all contexts. Is this a bug, or is there merit to https://github.com/scipy/scipy/issues/23931#issuecomment-3493669258?
- Assess whether gh-19769 has a simple, obvious fix. If so, it would be great for that issue to get resolved. If not, investigate the function further and whether it belongs in SciPy, considering the history of how that function got added (see git blame and review [mailing list archives](https://mail.python.org/archives/list/scipy-dev@python.org/)) that [hyppo](https://github.com/neurodata/hyppo) exists and is maintained.
- Assess whether there is anything SciPy can do about gh-23177 and if not, whether it has been fixed upstream yet.

Issue #8064, opened 2017-10-22
linalg.solve test crashes on Windows


dalito:
Running the scipy 1.0.0rc2 tests crashes Python 3.6.3-64bit. 

### Reproducing code example:
```
import scipy
scipy.test(verbose=2)
```

Error message:
```
::TestSolve::test_assume_a_keyword <- ..\..\dev_amd64\Python36\lib\site-packages\scipy\linalg\tests\test_basic.py PASSED
::TestSolve::test_all_type_size_routine_combinations <- ..\..\dev_amd64\Python36\lib\site-packages\scipy\linalg\tests\test_basic.py
<<Python interpreter crashes>>
````

Scipy/Numpy/Python version information:
```
scipy 1.0.0rc2 
numpy 1.13.3 
Python 3.6.3-64bit
Windows 10 Pro
```
Also #7500 reports an error in this test but not an interpreter crash.

**Comment by xoviat** [comment]
Is the version that you built yourself or the version that's from pypi?

**Comment by xoviat** [comment]
If it's the version that you built yourself you probably built openblas incorrectly and didn't even know it. Try to reproduce using the pypi version.

**Comment by dalito** [comment]
The crash happens with both the version from pypi and with my own build.

This morning I tried to reproduce the crash on two other machines. I did not get the segfault (all machines had Python 3.6.3-64bit from python.org, one has Win 7 Pro German the other Win10 Pro German). Both the scipy from pypi and my own build were OK on these two other machines.

On the machine where I observed the segfault, I also completely removed the Python 3.6 installation and reinstalled. The crash is reproducible. When I looked for differeces between the machines, I found out that the installed versions of VC redistributables differ. Maybe this is the cause of the problem.

Since I can reproduce the segfault only on one machine, this issue can probably be closed.

**Comment by pv** [comment]
What are the VC redistributable versions, and can you reproduce it on other machines if you downgrade (if that is possible to do)?

**Comment by dalito** [comment]
On the machine which segfaults:
![vcredistributables_when-segfaulting_s](https://user-images.githubusercontent.com/2648874/31861058-eb24c384-b725-11e7-9bce-b9b271092d62.png)
There is no VC redistributable from VS2017. I will no update to the latest versions and test again.

**Comment by dalito** [comment]
Success. After applying[ Microsoft Visual C++ 2015 Redistributable Update 3 RC](https://www.microsoft.com/en-us/download/details.aspx?id=52685) (to version 14.0.24123) the scipy tests complete fine.
```
[...]

PR #24098, opened 2025-12-03
TST: linalg: unskip a test

ev-br:
The test was problematic, and was permanently skipped in 2018 for scipy 1.2.0 (cf https://github.com/scipy/scipy/issues/8064)
The problematic versions were MSVC <2015 and we are at 2019 now.
Implementation of `solve` changed a little, too. Eyeballing the test, there is nothing immediately wrong with it, so hopefully it just fixed itself?

Unskip the test and trivially refactor it so that if it segfaults again, it's easier to see which mode/dtype does.
The refactor is clearer under "hide whitespace changes" view mode.

[wheel build]

### DIFF
--- a/scipy/linalg/tests/test_basic.py
+++ b/scipy/linalg/tests/test_basic.py
@@ -838,55 +838,41 @@ class TestSolve:
     def test_assume_a_keyword(self):
         assert_raises(ValueError, solve, 1, 1, assume_a='zxcv')
 
-    @pytest.mark.skip(reason="Failure on OS X (gh-7500), "
-                             "crash on Windows (gh-8064)")
-    def test_all_type_size_routine_combinations(self):
+    @pytest.mark.parametrize("size", [10, 100])
+    @pytest.mark.parametrize("assume_a", ['gen', 'sym', 'pos', 'her'])
+    @pytest.mark.parametrize(
+        "dtype", [np.float32, np.float64, np.complex64, np.complex128]
+    )
+    def test_all_type_size_routine_combinations(self, size, dtype, assume_a):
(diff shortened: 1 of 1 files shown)


mdhaber:
The array API failure looks like https://github.com/scipy/scipy/pull/24039#issuecomment-3602457415. I would have expected the Array API workflow to merge with `main` before running tests, which should pick up gh-24094, but maybe a similar fix is needed in `linalg`?

ev-br:
The Array API failures were in `fft` indeed. I rebased on main to see if this fixes it. The key run is the `[wheel build]` battery of runs, which aims to probe the full range of supported platforms (or close to the full range). 
The wheel builds passed in the previous push, https://github.com/scipy/scipy/actions/runs/19909411246, and I rerun them just to be on a safe side.
If it passes again, I suppose it's up to @tylerjereddy whether he wants to take it into 1.17.0rc1 (and potentially risk having to skip it again like gh-8064) or prefers to keep it to after branching.

One reason I think it makes sense to enable this test now is that the wrapping machinery for `solve` has changed in 1.17 (was a python decorator loop over f2py wrapped LAPACK calls previously, and is a C loop with no f2py in 1.17), and I'd be willing to debug it if it shows problems in the rc phase.

tylerjereddy (approved):
> I'd be willing to debug it if it shows problems in the rc phase.

Sounds good, and activating a previously-disabled test is of course otherwise pretty safe apart from the potential for the annoyance of disabling/fixing it again.

It sounds like there is a decent motivation to turn it back on so let's try it. The full wheel build matrix is green.

PR #24101, opened 2025-12-04
ENH: Add support for scikit-sparse v0.5.0

broesler:
This PR adds support for [scikit-sparse v0.5.0](https://github.com/scikit-sparse/scikit-sparse), namely, the `cholmod` and `umfpack` submodules. The `cholmod` module in scikit-sparse v0.4.16 has been largely rewritten to remove bugs and run more efficiently with a Cython interface. The scikit-sparse `umfpack` submodule is new as of v0.5.0, and replaces the [scikits-umfpack](https://github.com/scikit-umfpack/scikit-umfpack) project, which is largely dormant. See [this SciPy forum discussion](https://discuss.scientific-python.org/t/improve-interface-to-suitesparse-umfpack/2138) for more details.

The only external API change in SciPy is that the `sparse.linalg.use_solver` function has been removed along with the global variable `useUmfpack` used in `sparse.linalg._dsolve`. The `use_umfpack` kwarg alone controls the use of UMFPACK. This PR keeps the existing behavior of using the UMFPACK factorization/solver if scikit-sparse is installed, but silently defaulting to SuperLU factorization/solver otherwise.

Additional changes:
* `sparse.linalg.spsolve` raises `LinAlgError` for a singular matrix, regardless of the RHS input type (sparse vs. dense, 1D vs. 2D). Previously, a dense RHS would *not* raise, but return a vector of `np.nan`, whereas a sparse RHS would raise `RuntimeError` when `use_umfpack=False`.
* The `sparse.linalg.MatrixRankWarning` class has been removed.

#### Additional information
This PR will remain a draft until the scikit-sparse v0.5.0 changes are all merged into master. The development branch is available [on my fork](https://github.com/broesler/scikit-sparse).

### DIFF
--- a/scipy/interpolate/_ndbspline.py
+++ b/scipy/interpolate/_ndbspline.py
@@ -507,7 +507,9 @@ def make_ndbspl(points, values, k=3, *, solver=ssl.gcrotmk, **solver_args):
     vals_shape = (prod(v_shape[:ndim]), prod(v_shape[ndim:]))
     vals = values.reshape(vals_shape)
 
-    if solver != ssl.spsolve:
+    if solver == ssl.spsolve:
+        matr = matr.tocsc()  # spsolve requires CSC format
+    else:
         solver = functools.partial(_iter_solve, solver=solver)
         if "atol" not in solver_args:
             # avoid a DeprecationWarning, grumble grumble
--- a/scipy/optimize/_linprog_doc.py
+++ b/scipy/optimize/_linprog_doc.py
@@ -1002,7 +1002,7 @@ def _linprog_ip_doc(c, A_ub=None, b_ub=None, A_eq=None, b_eq=None,
     1. ``sksparse.cholmod.CholeskyFactor.solve`` (if scikit-sparse and
        SuiteSparse are installed)
 
-    2. ``scipy.sparse.linalg.factorized`` (if scikit-umfpack and SuiteSparse
+    2. ``sksparse.umfpack.UMFFactor.solve`` (if scikit-sparse and SuiteSparse
        are installed)
 
     3. ``scipy.sparse.linalg.splu`` (which uses SuperLU distributed with SciPy)
--- a/scipy/optimize/_linprog_doc.py
+++ b/scipy/optimize/_linprog_doc.py
@@ -999,8 +999,8 @@ def _linprog_ip_doc(c, A_ub=None, b_ub=None, A_eq=None, b_eq=None,
 
     For sparse problems:
 
-    1. ``sksparse.cholmod.cholesky`` (if scikit-sparse and SuiteSparse are
-       installed)
+    1. ``sksparse.cholmod.CholeskyFactor.solve`` (if scikit-sparse and
+       SuiteSparse are installed)
 
     2. ``scipy.sparse.linalg.factorized`` (if scikit-umfpack and SuiteSparse
        are installed)
--- a/scipy/optimize/_linprog_ip.py
+++ b/scipy/optimize/_linprog_ip.py
@@ -28,11 +28,7 @@ from ._linprog_util import _postsolve
 has_umfpack = True
 has_cholmod = True
 try:
-    import sksparse  # noqa: F401
-    from sksparse.cholmod import cholesky as cholmod  # noqa: F401
-    from sksparse.cholmod import analyze as cholmod_analyze
-    from sksparse.cholmod import CholmodTypeConversionWarning
-    from warnings import catch_warnings
+    from sksparse import cholmod
 except ImportError:
     has_cholmod = False
 try:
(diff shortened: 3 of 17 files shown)


broesler:
# Points of Discussion
The new `sksparse.umfpack` module only accepts CSC format matrices, which is
consistent with SuperLU (via `scipy.sparse.linalg.splu`). Non-CSC inputs are
converted to CSC and issue a `SparseEfficiencyWarning`. The existing
`scikits.umfpack` module accepts CSR format matrices without warning, but its
implementation is buggy:

* The factorization routine just converts to CSC internally with `mtx.tocsc()`
  and *recomputes* the factorization to return L, U, P, Q, R.
* If a `scikits.umfpack.UmfpackLU` object is created with a CSR input, then any
  of the L, U, P, Q, R matrices are accessed, the object's `solve` method will
  raise a `ValueError` because the internal factorization has been converted to
  CSC format independently of the input matrix CSR format.
* The solve routine uses `autoTranspose=True` to switch the system from
  A x = b to A.T x = b, but raises a `RuntimeError` if the input is CSR and the
  system requested is A.T.conj() (*i.e.* `trans='H'` in SuperLU terms), because
  the CSR format is a non-conjugate transpose.

It is possible to implement support for CSR format in `sksparse.umfpack`, but it
creates 2x the code paths for most operations, which puts a large burden on
maintenance and testing. It also means that the row scaling matrix output
becomes a *column* scaling matrix if the input is in CSR format, which would
likely be unexpected and confusing to the user.

## Performance Testing of SuperLU with CSR input
The consequence of this behavior is that `spsolve` only accepts CSR inputs
without warning when the RHS is dense and scikit-umfpack is not installed. In
this case, the existing `spsolve` directly calls the `_superlu.gssv` function
with an appropriate flag to indicate that the input is in CSR format. I was
curious about this code path, so I did some performance testing to see how it
compares to using `splu(A).solve(b)`. 

### CSC Input Matrix

superlu_solve_perf_CSC

* Minor peak memory increase when using `splu` with `b` shape (N, 1) since
  a temporary `SuperLU` object is created.
* `gssv` is faster for matrices up to size ~ 30 x 30, but runtime and memory use
  are essentially identical for larger matrices.

### CSR Input Matrix

superlu_solve_perf_CSR

* Essentially no difference in peak memory use for any scenario.
* `gssv` is substantially slower than `splu` for all but the smallest matrices.
  The difference becomes more pronounced as the number of RHS vectors increases.
  
[Full testing script](https://github.com/broesler/scipy_dev_scripts/blob/main/superlu_solve_perf.py).

## Conclusions
I have chosen to optimize for large matrices with multiple RHS
vectors, since that is where sparse solvers are most useful. Therefore, I
have removed the `_superlu.gssv` call from `spsolve`, and replaced it with
`splu(A).solve(b)`. This means that `spsolve` now only accepts CSC format inputs,
regardless of whether UMFPACK or SuperLU is used as the underlying solver.

## Downstream Changes
This change led to some minor changes in tests to filter the
`SparseEfficiencyWarning` where CSR inputs were used, or to change the inputs to
CSC format where appropriate.

There are two changes to other parts of the code base as a result:
1. In `scipy/interpolate/_ndbspline.py`, I have added a check if the solver is
   `spsolve`, and if so, convert the input matrix to CSC format before calling
   the solver.
2. In `scipy/optimize/_remove_redundancy.py`, I have replaced a call to
   `spsolve(B.transpose(), e)` with `spsolve(B, e, trans='T')`.
   Since both `splu(A).solve` and `umfpack.UMFFactor.solve` accept a `trans`
   argument, we can now accept `trans` directly to `spsolve`. In this particular 
   use case, `B` is created in CSC format for column slicing, so its transpose
   is in CSR format. By passing `trans='T'`, we can avoid the explicit transpose
   and conversion to CSC format before calling `spsolve`.

**Note**: `_superlu.gssv` accepts an [options dict](https://portal.nersc.gov/project/sparse/superlu/superlu_code_html/dgssv_8c.html) which may contain a `Trans` option,
but it does not seem like the function
respects that option. SciPy's wrapper seems to pass that option through to
SuperLU, but the SuperLU library itself does not seem to support that option in
`gssv`.

These changes fix all of the `SparseEfficiencyWarning` issues in the SciPy test
suite (with or without `sksparse.umfpack` installed). I am open to any feedback
on either the behavioral or implementation changes.

broesler:
I have opened a [PR for v0.5.0](https://github.com/scikit-sparse/scikit-sparse/pull/152) on the scikit-sparse project. If any of the reviewers here have comments on the updated API, feel free to join that discussion. I am now a maintainer of that project and will merge the changes next week if there are no objections.

tylerjereddy:
This PR has no review comments from SciPy core developers despite being open for almost six months.

It looks like there is support for this, but perhaps there hasn't been enough recent activity to make it safe to merge this just days before branching for the next release.

Perhaps I'll bump the milestone so it doesn't fall off the radar, apologies it has taken a while for us to get to reviewing this.

broesler:
Thanks, @tylerjereddy. I am cleaning up some things in scikit-sparse v0.5.0 and can then get this PR updated and convert to being ready for review.

lucascolley:
hey @broesler, just wondering if there is any update here?

broesler:
I am waiting on review for the [conda-forge PR](https://github.com/conda-forge/scikit-sparse-feedstock/pull/55) of Scikit-Sparse v0.5.0, and then will convert this to a reviewable PR.

broesler:
This PR incorporates changes from #24055 and #24037. Based on the discussion from #24037, I am fine with reverting to the old behavior regarding size (N, 1) RHS, but the current behavior in this PR is consistent with the SuperLU and Scikit-Sparse packages.

lucascolley:
> The only external API change in SciPy is that the `sparse.linalg.use_solver` function has been removed

Has this been discussed yet at all? Typically we put public API removals through a deprecation cycle rather than executing them immediately.

broesler:
> > The only external API change in SciPy is that the `sparse.linalg.use_solver` function has been removed
> 
> Has this been discussed yet at all? Typically we put public API removals through a deprecation cycle rather than executing them immediately.

This deprecation has not been discussed, but I'd be happy to make the changes to deprecate it. Is there specific developer documentation on the proper way to do that? Or a recent example to which I can refer?

lucascolley:
> > > The only external API change in SciPy is that the `sparse.linalg.use_solver` function has been removed
> 
> > 
> 
> > Has this been discussed yet at all? Typically we put public API removals through a deprecation cycle rather than executing them immediately.
> 
> 
> 
> This deprecation has not been discussed, but I'd be happy to make the changes to deprecate it. Is there specific developer documentation on the proper way to do that? Or a recent example to which I can refer?

For very simple cases, a proposal like https://discuss.scientific-python.org/t/deprecate-stats-tiecorrect/2438/1 suffices. For more complex/impactful cases, we're looking for a proposal more akin to this: https://discuss.scientific-python.org/t/dep-depcrecate-lti-classes-in-signal-module/2432/3.

This is definitely closer to the former case, but it would be good to explain how you assess the impact of the removal and what the transition guidance would be to users. If the transition is easy enough, we may be happy to make a hard break rather than going through a deprecation cycle, but that needs discussion.

More info at https://scipy.github.io/devdocs/dev/core-dev/index.html#deprecations.

Issue #19024, opened 2023-08-07
BUG: #18481 made BFGS minimization less accurate


jakevdp:
### Describe your issue.

#18481 appears to have made BFGS minimization less accurate. We noticed this in the JAX upstream nightly tests, which began failing after #18481 was merged: https://github.com/google/jax/actions/workflows/upstream-nightly.yaml

I suspect the dtype changes in `approx_derivative` are to blame.

### Reproducing Code Example

```python
import numpy as np
import scipy

def func(p):
    x, y = p
    return (x ** 2 + y - 11.) ** 2 + (x + y ** 2 - 7.) ** 2

x0 = np.ones(2, dtype='float32')
result = scipy.optimize.minimize(func, x0, method='BFGS')

print(f"{scipy.__version__=}")
print(result)
np.testing.assert_allclose(result.x, [3, 2])
```

### Error message

Output before #18481:
```
scipy.__version__='1.11.1'
  message: Optimization terminated successfully.
  success: True
   status: 0
      fun: 2.288511533827582e-15
        x: [ 3.000e+00  2.000e+00]
      nit: 8
      jac: [-1.907e-08  2.738e-08]
 hess_inv: [[ 1.612e-02 -9.549e-03]
            [-9.549e-03  3.517e-02]]
     nfev: 39
     njev: 13
```

Output after #18481:
```
scipy.__version__='1.12.0.dev0+1440.64be173'
  message: Desired error not necessarily achieved due to precision loss.
  success: False
   status: 2
      fun: 9.628441344453949e-06
        x: [ 3.000e+00  2.000e+00]
      nit: 6
      jac: [ 6.255e-04  2.977e-04]
 hess_inv: [[ 1.615e-02 -9.191e-03]
            [-9.191e-03  3.561e-02]]
     nfev: 291
     njev: 70
```
```pytb
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-1-d421a22eaf1d> in <cell line: 13>()
     11 result = scipy.optimize.minimize(func, x0, method='BFGS')
     12 print(result)
---> 13 np.testing.assert_allclose(result.x, [3, 2])

/usr/lib/python3.10/contextlib.py in inner(*args, **kwds)
     77         def inner(*args, **kwds):
     78             with self._recreate_cm():
---> 79                 return func(*args, **kwds)
     80         return inner
     81 

[...]

PR #18481, opened 2023-05-17
MAINT: differentiable fns respect float width.  Closes #15602

andyfaff:
`ScalarFunction` and `VectorFunction` now send the same floating point type that was used to construct the differentiable function to the user supplied function.

`approx_derivative` is also updated to send the appropriate floating point type to the function to be differentiated.

`approx_derivative` (and by extension `ScalarFunction.grad`/`VectorFunction.jac`) still currently return`np.float64`. I'm not sure of the ramifications of changing the output of `approx_derivative` to match the most precise floating point type of the input and output.
For example `approx_derivative` could be called with `x0.dtype == np.float32`, but `fun(x0).dtype` could be `np.float64`, so it seemed natural to me to keep `np.float64` as the output. But If both input and output are `np.float32` then the return of `approx_derivative` could possibly be changed to `np.float32`. I'm not sure what effect this would have on the downstream consumers of `approx_derivative`.

PR #24104, opened 2025-12-05
BUG: optimize: prevent float32 overflow in BFGS line search

krishneetRAJ:
**Reference Issue: **
Fixes #23944 

**What it Fixes: **
This PR addresses a float32 overflow issue in the DCSRCH line search used by BFGS. This leads to non-finite step values and causing the optimization to fail almost immediately. 

-Uses float64 to avoid overflow, then casts back to float32
-Maintains dtype behavior
-Adds a regression test ensuring BFGS works correctly with float32 inputs

**Additional information:**
Some issues that have been referenced when solving this issue:
https://github.com/scipy/scipy/issues/19024, https://github.com/scipy/scipy/pull/18481, #22865

### DIFF
--- a/scipy/optimize/tests/test_optimize.py
+++ b/scipy/optimize/tests/test_optimize.py
@@ -34,10 +34,9 @@ from scipy.optimize._root import ROOT_METHODS
 from scipy.optimize._root_scalar import ROOT_SCALAR_METHODS
 from scipy.optimize._qap import QUADRATIC_ASSIGNMENT_METHODS
 from scipy.optimize._differentiable_functions import ScalarFunction, FD_METHODS
-from scipy.optimize._optimize import (
-    MemoizeJac, show_options, OptimizeResult, _minimize_bfgs
-)
+from scipy.optimize._optimize import MemoizeJac, show_options, OptimizeResult
 from scipy.optimize import rosen, rosen_der, rosen_hess
+from scipy.optimize._optimize import _minimize_bfgs
 
 from scipy.sparse import (coo_matrix, csc_matrix, csr_matrix, coo_array,
                           csr_array, csc_array)
@@ -3482,7 +3481,6 @@ class TestWorkers:
         assert res.success
         assert_allclose(res.x[1], 2.0)
 
-
 # Tests for PEP 649/749 style annotations in callback and objective functions
 if sys.version_info < (3, 14):
     from typing import Any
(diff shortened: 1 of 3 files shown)


scipy/optimize/tests/test_optimize.py
```diff
@@ -3482,122 +3481,26 @@ def test_equal_bounds(self, workers, method):
...
+    # Use a more reasonable starting point for float32 precision
+    x0 = np.array([3.0, 3.0], dtype=np.float32)
+    res = _minimize_bfgs(rosenbrock, x0, jac=rosenbrock_grad, maxiter=10000, gtol=1e-3)
+
+    # It should not produce overflow warnings or type promotions
+    assert np.all(np.isfinite(res.x)), "Optimization produced non-finite values"
+    assert res.x.dtype == np.float32, "Result dtype should remain float32"
+    # We expect convergence near [1, 1] (within float32 precision)
+    # With float32, we can't expect as tight convergence as float64
+    assert np.allclose(res.x, [1.0, 1.0], atol=0.01) or res.success
```
tylerjereddy, line 3625: `assert_allclose` is generally preferable because of the testing-related feedback it provides

scipy/optimize/tests/test_optimize.py
```diff
@@ -3482,122 +3481,26 @@ def test_equal_bounds(self, workers, method):
...
-
-def test_multiprocessing_too_many_open_files_23080():
-    # https://github.com/scipy/scipy/issues/23080
-    x0 = np.array([0.9, 0.9])
-    # check that ScalarHessWrapper doesn't keep pool object alive
-    with assert_deallocated(multiprocessing.Pool, 2) as pool_obj:
-        with pool_obj as p:
-            _minimize_bfgs(rosen, x0, workers=p.map)
-        del p
-        del pool_obj
```
tylerjereddy, line 3603: I'm a bit confused--why are regression tests unrelated to the issue at hand being deleted?

krishneetRAJ: Yes, sorry about that. I accidentally deleted it; it should be fixed now.

nickodell:
I would suggest closing this PR as a duplicate of #22865.

The reason for this is that I think that #22865 is a technically simpler change, and unlocks an additional optimization.

Specifically, #22865 also solves the problem of overflow for float32 inputs, but it does so by adjusting stpmax for the specific types involved. This is simpler than trying to do the calculation in 64 bit for part of the calculation, then clipping and down-casting.

This PR also makes a number of unnecessary changes, like changing a calculation from being done using Python scalars (in 64 bit) to using NumPy scalars of float64 type. This is slower and more complex than the current code.

